PR #1932, opened 2025-03-31
ui: add validation to provider form

lily-de:
Adds some validation if user tries to submit incomplete form (i think it didn't work before... but now we show what's missing)

Screenshot 2025-03-30 at 8 21 15 PM


ui/desktop/src/components/settings_v2/providers/modal/subcomponents/handlers/DefaultSubmitHandler.tsx
```diff
@@ -1,81 +1,3 @@
-import { useConfig } from '../../../../../ConfigContext';
```
lilydelalande, line 1: removed unused function while i was at it

yingjiehe-xyz (approved):

alexhancock (approved):

PR #1933, opened 2025-03-31
feat: deepseek v3 (via huggingface inference providers)

michaelneale:
EDIT: look at https://github.com/block/goose/pull/1934 instead 

Deepseek v3 models (march 2025) are able to do tool calling and have powerful inference, this change brings in a huggging face provider as well as working through challenge of subtleties of tool calling with deepseek in particular

WIP (see deepseek.md for research notes)

To try: 

```bash
GOOSE_PROVIDER=huggingface GOOSE_MODEL=deepseek-ai/DeepSeek-V3-0324 cargo run --bin goose run -t "list files in this dir"
``` 

will need a HUGGINGFACE_TOKEN that can access their inference providers

Deepseek v3 does not support auto tool choice: 

```
(status 400 Bad Request): Object {"detail": String("Invalid request. Please check the parameters and try again. Details: This model does not support auto tool, please use tool_choice.")}
```

PR #1934, opened 2025-03-31
WIP: deepseekv3 0324 (via huggingface)

michaelneale:
This uses https://huggingface.co/deepseek-ai/DeepSeek-V3-0324 - via the hugging face providers api (nebius seems to mostly work, but it isn't consistent with tool calling, as has been seen elsewhere by @ahau-square). 

This one uses the recommended format from deepseek 2.5 model card and the system prompt, and doesn't tool call/results in the message list. If it would consistently call tools, this may be useful (however I found it variable with time of day and provider still).

to try: 

```sh
GOOSE_PROVIDER=huggingface GOOSE_MODEL=deepseek-ai/DeepSeek-V3-0324 cargo run --bin goose run -t "list files in this dir" 
```

and have HUGGINGFACE_TOKEN in your env.

This was able to do multi turn calls to tools:

image

This replaces: https://github.com/block/goose/pull/1933 - which won't work as deepseek does not yet support tool auto selection (so kind of defeats the purpose of using the tool api in openai, if you have to specify just one to call up front, this one doesn't require that).


backkem:
@michaelneale do you have any interest in reviving the HuggingFace provider? Seems like a good alternative to OpenRouter with all the new open models coming out.

PR #1936, opened 2025-03-31
ui: set up model list component

lily-de:
Reuses some code from settings v1 to create a list of models. Re-introduces the idea of a "Model" interface with some metadata attached such as alias (a nickname for the model) and subtext (an alternative to the provider name). Also re-uses some code for getting and storing recent models in localStorage but this is not used yet. 

As a follow up, will start constructing the current model into a `Model` and use that for display. 

Purpose of this is for two reasons:

1. Some releases of Goose may want to used a pre-defined set of models, this allows them to do that
2. A future PR will replace the current Model settings (which just shows the current model) and show a list of recently used models, this is just setting up some components and interfaces that allow for that -- the BaseModelsList allows for a creation of a list of models, and the Model interface allows for additional metadata beyond just the unformatted model & provider names we store in the config

### DIFF
--- a/ui/desktop/src/components/settings_v2/models/model_list/recentModels.ts
+++ b/ui/desktop/src/components/settings_v2/models/model_list/recentModels.ts
@@ -1,0 +1,30 @@
+import React, { useEffect, useState } from 'react';
+import Model from '../modelInterface';
+
+const MAX_RECENT_MODELS = 3;
+
+export function useRecentModels() {
+  const [recentModels, setRecentModels] = useState<Model[]>([]);
+
+  useEffect(() => {
+    const storedModels = localStorage.getItem('recentModels');
+    if (storedModels) {
+      setRecentModels(JSON.parse(storedModels));
+    }
+  }, []);
+
+  const addRecentModel = (model: Model) => {
+    const modelWithTimestamp = { ...model, lastUsed: new Date().toISOString() }; // Add lastUsed field
+    setRecentModels((prevModels) => {
+      const updatedModels = [
+        modelWithTimestamp,
+        ...prevModels.filter((m) => m.name !== model.name),
+      ].slice(0, MAX_RECENT_MODELS);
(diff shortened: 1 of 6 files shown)


ui/desktop/src/components/settings_v2/models/model_list/recentModels.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  const addRecentModel = (model: Model) => {
+    const modelWithTimestamp = { ...model, lastUsed: new Date().toISOString() }; // Add lastUsed field
+    setRecentModels((prevModels) => {
+      const updatedModels = [
+        modelWithTimestamp,
+        ...prevModels.filter((m) => m.name !== model.name),
+      ].slice(0, MAX_RECENT_MODELS);
+
+      localStorage.setItem('recentModels', JSON.stringify(updatedModels));
```
alexhancock, line 24: Thoughts on just adding a new section to `config.yaml` to store this? The endpoints would probably be similar to extensions.

We could also do it as a followup if this helps us ship faster, but I really think it would buy us a lot to only have one thing for persistent config state.

alexhancock (approved):

ui/desktop/src/components/settings_v2/models/model_list/recentModels.ts
lilydelalande: In talking to Salman and @matthewdiamant it seemed like localStorage made sense as it's a UI-only configuration. But I am open to either -- this is not used yet so easy fix

alexhancock: Yeah, I personally really want to just get rid of localstorage. The app is so much simpler to reason about when all config is stored in the file. It also opens the ability to build a feature on CLI for example where we could let people toggle easily to recently used models for example.

Can do in a followup!

PR #1939, opened 2025-03-31
feat: add tool annotations for build-in tools

yingjiehe-xyz:
Add tool annotations for build-in tools following the new mcp schema: https://github.com/modelcontextprotocol/specification/blob/main/schema/2025-03-26/schema.ts#L730C18-L730C33

Our permission control will use such annotations to check whether we should ask for human approval. It will be done in following PRs

In this PR, we only add annotations for tools with clear intention, like pure read or write tools. Some tools like bash or google drive update, we DON'T add annotations because they require LLM-as-a-judge to check the permission

### DIFF
--- a/crates/goose-mcp/src/computercontroller/mod.rs
+++ b/crates/goose-mcp/src/computercontroller/mod.rs
@@ -256,7 +256,7 @@ impl ComputerControllerRouter {
                 title: Some("PDF process".to_string()),
                 read_only_hint: true,
                 destructive_hint: false,
-                idempotent_hint: false,
+                idempotent_hint: true,
                 open_world_hint: false,
             }),
         );
--- a/crates/goose-mcp/src/computercontroller/mod.rs
+++ b/crates/goose-mcp/src/computercontroller/mod.rs
@@ -13,7 +13,7 @@ use mcp_core::{
     prompt::Prompt,
     protocol::ServerCapabilities,
     resource::Resource,
-    tool::Tool,
+    tool::{Tool, ToolAnnotations},
     Content,
 };
 use mcp_server::router::CapabilitiesBuilder;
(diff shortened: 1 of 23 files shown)


crates/goose-mcp/src/computercontroller/mod.rs
```diff
@@ -215,6 +224,7 @@ impl ComputerControllerRouter {
                     }
                 }
             }),
+            None,
```
yingjiehe-xyz, line 227: because these tools contain mixed operations, we don't add annotations and ask goose to check the permission

crates/mcp-core/src/tool.rs
```diff
@@ -3,6 +3,70 @@
...
+    /// For example, the world of a web search tool is open, whereas that
+    /// of a memory tool is not.
+    ///
+    /// Default: true
+    #[serde(default = "default_true")]
+    pub open_world_hint: bool,
+}
+
+impl Default for ToolAnnotations {
+    fn default() -> Self {
```
kalvinnchau, line 55: Might be cleaner to add some `with_*` builder type pattern so we can only add in the fields that we need and leave it as defaults otherwise:
```rust
.with_title(title)
.with_destructive(true)
... 
```

similar to how we do it for `Content` https://github.com/block/goose/blob/main/crates/mcp-core/src/content.rs#L78

yingjiehe-xyz: Make sense, done!

yingjiehe-xyz: Discussed offline, with_xxx might not clear, with the original implementation, we don't need to care about the default value, we can directly add annotation with our understanding. When having with_xxx methods, we need to remember what's the default value and see whether we need to update it

kalvinnchau (approved):

PR #1904, opened 2025-03-28
Fix global shortcuts

zanesq:
Fixes https://github.com/block/goose/issues/1903

PR #1940, opened 2025-03-31
Revert "Fix global shortcuts"

zanesq:
Reverts block/goose#1904 while investigating slow typing reports


zanesq:
Seems to be something else so holding on for now

PR #1948, opened 2025-03-31
Fix documentation formatting in ui/desktop readme

rclations:
The current documentation is missing a `, which throws off the formatting.

### DIFF
--- a/ui/desktop/README.md
+++ b/ui/desktop/README.md
@@ -30,7 +30,7 @@ This allows you to set for example GOOSE_PROVIDER__TYPE to be "databricks" by de
 
 # Runninng with goosed server from source
 
-Set `VITE_START_EMBEDDED_SERVER=yes` to no in `.env.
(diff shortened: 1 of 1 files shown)


ui/desktop/README.md
```diff
@@ -30,7 +30,7 @@ This allows you to set for example GOOSE_PROVIDER__TYPE to be "databricks" by de
 
 # Runninng with goosed server from source
 
-Set `VITE_START_EMBEDDED_SERVER=yes` to no in `.env.
+Set `VITE_START_EMBEDDED_SERVER=yes` to no in `.env.`
```
yingjiehe-xyz: should it be `.env` instead of `.env.`?

rclachance: nice catch - moved the period outside the `, but it could also be removed if desired

yingjiehe-xyz (approved):

PR #1955, opened 2025-03-31
ui: add timeout field to settings v2 modal

lily-de:
Screenshot 2025-03-31 at 6 51 06 PM

### DIFF
--- a/ui/desktop/src/components/Modal.tsx
+++ b/ui/desktop/src/components/Modal.tsx
@@ -46,23 +46,29 @@ export default function Modal({
       }
     };
 
-    // Add event listener
+    // Add event listener for Escape key
     document.addEventListener('keydown', handleEscKey);
+
+    // Add overflow-hidden to body to prevent scrolling background
+    document.body.style.overflow = 'hidden';
 
     // Clean up
     return () => {
       document.removeEventListener('keydown', handleEscKey);
+      // Restore body scrolling when modal closes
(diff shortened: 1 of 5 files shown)


ui/desktop/src/components/settings_v2/extensions/ExtensionsSection.tsx
```diff
@@ -80,6 +80,7 @@ export default function ExtensionsSection() {
 
   const handleUpdateExtension = async (formData: ExtensionFormData) => {
     const extensionConfig = createExtensionConfig(formData);
+    console.log('extensionConfig', extensionConfig);
```
lilydelalande:
```suggestion
```

ui/desktop/src/components/Modal.tsx
```diff
@@ -46,23 +46,29 @@ export default function Modal({
       }
     };
 
-    // Add event listener
+    // Add event listener for Escape key
     document.addEventListener('keydown', handleEscKey);
 
+    // Add overflow-hidden to body to prevent scrolling background
+    document.body.style.overflow = 'hidden';
```
alexhancock, line 53: Awesome. This works well as a simple fix to lock scrolling because we only use nested scroll elements. If we ever do a full bleed scrolling window we'd need to do some management of `scrollTop`

alexhancock (approved):

Issue #1945, opened 2025-03-31
Slow typing in chat area


zanesq:
**Describe the bug**
Typing in the chat area is slow to respond occasionally. Seems to be related to a large session.

**To Reproduce**
1. Go to Session history, load a large session with at least 100,000 tokens
2. Attempt to type in the chat area note it is slow to enter characters

**Expected behavior**
Typing into chat should be quick and responsive.

**Screenshots**

**Please provide following information:**
 - **OS & Arch:** Mac m4 arm
 - **Interface:** UI
 - **Version:** Version 1.0.15
 - **Extensions enabled:** None
 - **Provider & Model:** Databricks

**Additional context**
Seems to be intermittent also, doesn't happen every time

PR #1958, opened 2025-04-01
Fixes slow typing when long session

zanesq:
Fixes slow typing when long session. Controlled input was triggering re-render of chat view with every keystroke.

Fixes https://github.com/block/goose/issues/1945

### DIFF
--- a/ui/desktop/src/components/Input.tsx
+++ b/ui/desktop/src/components/Input.tsx
@@ -154,12 +154,6 @@ export default function Input({
     }
   };
 
-  const handleStop = (e: React.MouseEvent) => {
-    e.preventDefault();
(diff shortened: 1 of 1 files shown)


lily-de (approved):

ui/desktop/src/components/Input.tsx
```diff
@@ -194,11 +197,7 @@ export default function Input({
           type="button"
           size="icon"
           variant="ghost"
-          onClick={(e) => {
```
alexhancock: Mind re-inlining this? Lots of our components are getting long with the separate declarations up top and where something can be inlined I think it's a little nicer.

alexhancock (approved):
Nice find!

PR #1962, opened 2025-04-01
Fix search covering scrollbar

zanesq:
The new search component is over the scroll bar which can be problematic when the scrollbar is smaller and the user can't reach it. Changed to use `sticky` rather than `fixed` so it stays within its container. Not as pretty as full width but we'd need to add way more code to get `fixed` to respect all types of scrollbar widths on different OSs so this is a simple solution for now.

Before:
Screenshot 2025-03-31 at 6 10 44 PM

After:
Screenshot 2025-03-31 at 6 08 42 PM

### DIFF
--- a/ui/desktop/src/components/ChatView.tsx
+++ b/ui/desktop/src/components/ChatView.tsx
@@ -387,7 +387,8 @@ export default function ChatView({
             activities={botConfig?.activities || null}
           />
         ) : (
-          <ScrollArea ref={scrollRef} className="flex-1 px-4" autoScroll>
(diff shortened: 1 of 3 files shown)


nahiyankhan:
hear your reasoning! let me ponder a bit if you don't mind. hesitant cause I want to avoid introducing "if something looks a bit off / broken - its going to feel off / broken to users".

nahiyankhan:
pondered! slight alt take - https://github.com/block/goose/pull/1964

lilydelalande (approved):

zanesq:
Closing in favor of https://github.com/block/goose/pull/1964 thanks @nahiyankhan!

PR #1964, opened 2025-04-01
alt take

nahiyankhan:
scrollbar always disabled stays full width (and the scrollbar shows up on top).

scrollbar always enabled gets padded in -

Screenshot 2025-03-31 at 11 00 16 PM

### DIFF
--- a/ui/desktop/src/components/ChatView.tsx
+++ b/ui/desktop/src/components/ChatView.tsx
@@ -388,10 +388,10 @@ export default function ChatView({
           />
         ) : (
           /* padding needs to be passed into the container inside ScrollArea to avoid pushing the scrollbar out */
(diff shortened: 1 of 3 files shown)


ui/desktop/src/components/ChatView.tsx
```diff
@@ -388,10 +388,10 @@ export default function ChatView({
           />
         ) : (
           /* padding needs to be passed into the container inside ScrollArea to avoid pushing the scrollbar out */
```
zanesq: Can remove this comment now since there is no padding anymore. I had put it on the only place we were adding padding to ScrollArea to remind people later :)

zanesq:
btw is it easy to add search to the session history view?

PR #1966, opened 2025-04-01
fix developer mcp for windows

Kvadratni:
this fix should allow goose to correctly use powershell in native windows

### DIFF
--- a/crates/goose-mcp/src/developer/mod.rs
+++ b/crates/goose-mcp/src/developer/mod.rs
@@ -22,6 +22,7 @@ use mcp_core::{
     protocol::ServerCapabilities,
     resource::Resource,
     tool::Tool,
+    Content,
 };
 use mcp_core::{
     prompt::{Prompt, PromptArgument, PromptTemplate},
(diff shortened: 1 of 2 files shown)


crates/goose-mcp/src/developer/mod.rs
```diff
@@ -507,7 +507,8 @@ impl DeveloperRouter {
             .await
             .map_err(|e| ToolError::ExecutionError(e.to_string()))?;
 
-        let output_str = String::from_utf8_lossy(&output.stdout);
+        let stdout_str = String::from_utf8_lossy(&output.stdout);
+        let output_str = stdout_str;
```
michaelneale, line 511: not sure what this change is doing? (something to do with rust keeping something around? or just left over change?)

michaelneale (approved):
looked like some unnecessary changes in developer/mod.rs but very cool! and yes, powershell I believe would make sense from what I hear on windows for sure.

crates/goose-mcp/src/developer/mod.rs
Kvadratni: yeah, for some reason on windows it just had nothing in response and this works. Don't ask my why

PR #1968, opened 2025-04-01
fix windows extensions

Kvadratni:
massive refactor to fix jbang, uvx and nps for windows.
Additionally should reduce the bundle size for non windows builds by placing the windows bin dependencies separately.


michaelneale (changes requested):
did you mean to remove the shims for non windows? as there is an install node for windows, but not for anyting else (and bin dir will be I guess empty now)

ui/desktop/src/bin/.gitkeep
lilydelalande, line 1: think we need these for non-windows right

Kvadratni: fixed

ui/desktop/src/bin/uvx
salman1993, line 1: why is this an empty file?

ui/desktop/src/bin/npx
salman1993, line 1: why is this an empty file?

salman1993 (changes requested):
can we revert changes for existing npx and uvx shims?

ui/desktop/src/bin/npx
Kvadratni: fixed

ui/desktop/src/bin/uvx
Kvadratni: fixed

lily-de (approved):

Kvadratni:
Internal builds
https://github.com/squareup/goose-releases/actions/runs/14320886027

crates/goose-server/src/routes/extension.rs
```diff
@@ -108,6 +108,59 @@ async fn add_extension(
         return Err(StatusCode::UNAUTHORIZED);
     }
 
+    // If this is a Stdio extension that uses npx, check for Node.js installation
+    if let ExtensionConfigRequest::Stdio { cmd, .. } = &request {
+        if cmd.ends_with("npx.cmd") || cmd.ends_with("npx") {
```
michaelneale, line 114: this would be better if it was also conditional on windows wouldn't it?

michaelneale (changes requested):

crates/goose-server/src/routes/extension.rs
Kvadratni: not sure what you mean by that. It will check if you have node on line 115 so if you do - it will skip this whole thing

michaelneale (approved):
conditional approve - ideally would have cfg/windows for the code path to do with nodejs on it, but otherwise looks good. Will require some testing of course for no regressions

PR #1976, opened 2025-04-01
feat: create named feature flag for settings v2

alexhancock:
Create a named feature flag for settings V2, so we can easily turn it on and off with one line code changes. Right now it remains set to `process.env.ALPHA`, but we can change it soon to `true` to release, and `process.env.ALPHA` if we need to roll back.


ui/desktop/src/flags.ts
lilydelalande, line 1: hmm... how is process.env.ALPHA set again...? idk if this will work for a bundled build

Issue #1946, opened 2025-03-31
Session loading freezing app


zanesq:
**Describe the bug**

I noticed I had a few corrupt sessions not loading after I clicked on them, it would show a blank chat area and freeze up. @salman1993 and I debugged it and noticed the session JSON-L files had not been formatted correctly when they were saved. One had bad formatting, another had an extra new line.

**To Reproduce**

**Expected behavior**
Session should load like normal after clicking on it.

**Please provide following information:**
 - **OS & Arch:** Mac m4 arm
 - **Interface:** UI
 - **Version:** Version 1.0.15-block
 - **Extensions enabled:** None
 - **Provider & Model:** Databricks

**Additional context**

salman1993:
added better error reporting for now when session is corrupted

PR #1985, opened 2025-04-01
chore: add tests for session serialization issues

salman1993:
i could not replicate the error from this one: https://github.com/block/goose/issues/1946 

my hunch is there might be sth related to concurrent writes


salman1993:
Screenshot 2025-04-01 at 3 41 33 PM

PR #1986, opened 2025-04-01
ui: fix env var detection

lily-de:
Makes sure GOOSE_VERSION is picked up

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -130,9 +130,19 @@ const getSharingUrl = () => {
   return process.env.GOOSE_BASE_URL_SHARE;
 };
 
+const getVersion = () => {
+  // checks app env for sharing url
+  loadShellEnv(app.isPackaged); // will try to take it from the zshrc file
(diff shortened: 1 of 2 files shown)


ui/desktop/src/main.ts
```diff
@@ -130,10 +130,20 @@ const getSharingUrl = () => {
   return process.env.GOOSE_BASE_URL_SHARE;
 };
 
+const getVersion = () => {
+  // checks app env for sharing url
+  loadShellEnv(app.isPackaged); // will try to take it from the zshrc file
+  // if GOOSE_BASE_URL_SHARE is found, we will set process.env.GOOSE_BASE_URL_SHARE, otherwise we return what it is set
```
alexhancock: nit: remove comment from other var

alexhancock (approved):

ui/desktop/src/main.ts
lilydelalande:
```suggestion
```

PR #1988, opened 2025-04-01
fix: truncate large tool outputs

clementgarnier:
When a tool sends a very large output exceeding the model's context window, the existing summarization / truncation logic is unable to handle it (because the last message exceeds the limit on its own) and Goose throws the following error:
```
Error: Unable to truncate messages to stay within context limit.
Ran into this error: Unable to preserve any messages within context limit.
Please start a new session with fresh context and try again.
```

Unfortunately, this means all previous context of the session is lost.

While this would be, ideally, better handled on the MCP server side, we can also defend against this by truncating the output of tools to approximately 70% of the current model's context window before we attempt to inject it into the context.

I have tested this with a local UI build, a tool that I know to be returning very large outputs, and observed the following in the logs:
```
INFO goose::tool_output_limiter: Truncating large tool output: tool=codesearch__search_code, tokens=201103, limit=89600 (70% of context window)
```

### DIFF
--- a/crates/goose/src/agents/truncate.rs
+++ b/crates/goose/src/agents/truncate.rs
@@ -26,6 +26,7 @@ use crate::providers::toolshim::{
 use crate::register_agent;
 use crate::session;
 use crate::token_counter::TokenCounter;
+use crate::tool_output_limiter::ToolOutputLimiter;
 use crate::truncate::{truncate_messages, OldestFirstTruncation};
 use anyhow::{anyhow, Result};
 use indoc::indoc;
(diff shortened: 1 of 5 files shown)


crates/goose/src/agents/truncate.rs
```diff
@@ -343,7 +359,13 @@ impl Agent for TruncateAgent {
 
                                 // Add pre-approved tools
                                 for (request_id, tool_call) in approved_tools {
-                                    let tool_future = Self::create_tool_future(&capabilities, tool_call, request_id.clone());
+                                    let tool_future = Self::create_tool_future(
```
salman1993, line 362: the main challenge with this approach is that we don't want token counting to happen in the agent loop cause that is slow. instead its preferred to wait till we get the context length exceeded error and then handle the logic. would it be possible to do this check here: https://github.com/block/goose/blob/36a8be4c395134504a2717c70507c6408c04ab60/crates/goose/src/agents/truncate.rs#L486-L516

baxen:
thanks for the contrib @clementgarnier. I believed we added a similar feature recently that covers this, but please feel free to reopen if it is still relevant!

Issue #1972, opened 2025-04-01
Goose installation error on WSL


teck968:
**Describe the bug**
When running the goose install script on a freshly installed WSL, the script produces an error message that may not be intuitive to new Linux users if bzip2 is not installed.

**To Reproduce**
1. Execute `curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh | bash` within a fresh WSL2 install.

**Expected behavior**
The download_cli.sh script should either exit with a useful error message when a dependency is missing (like "bzip2 is missing. Please run 'sudo apt install bzip2' to install it and try again.") or it should install it automatically if possible.

**Screenshots**
![Image](https://github.com/user-attachments/assets/55562db2-53c0-4a39-a035-618ffcd70bdf)

**Please provide following information:**
 - **OS & Arch:** Windows 11/WSL2/Ubuntu-24.04
 - **Interface:** CLI
 - **Version:** Goose Version: 1.0.16-canary+17e601f
 - **Extensions enabled:** N/A
 - **Provider & Model:** N/A

**Additional context**
bzip2 is usually installed by default on most full-featured Linux distributions, especially desktop/server editions of:
- Ubuntu
- Debian
- Fedora
- CentOS / RHEL
- Arch Linux

However, in minimal environments — like:
- Docker containers
- WSL
- Alpine Linux
- Custom cloud images
- Minimal installs

— it's often not included by default to keep the base system lean.

salman1993:
i think you can add in a check here based on os and whether bzip2 is installed or not - https://github.com/block/goose/blob/3c972b2dadaf5fc5682883c81513ad4ba987cea3/download_cli.sh#L65-L71 

do you wanna give it a shot?

teck968:
Sure, I'll give it a go.

PR #1998, opened 2025-04-02
Fix: Added check for bzip2 in download_cli.sh

teck968:
This fix addresses [issue #1972 ](https://github.com/block/goose/issues/1972).
Based on my research, macOS and Linux both require bzip2 to be installed when running the command `tar -xjf "$FILE" -C "$TMP_DIR"`.


salman1993 (approved):
thanks for adding this!

download_cli.sh
```diff
@@ -22,12 +22,19 @@ set -eu
...
-# --- 1) Check for curl ---
+# --- 1) Check for dependencies ---
+# Check for curl
 if ! command -v curl >/dev/null 2>&1; then
   echo "Error: 'curl' is required to download Goose. Please install curl and try again."
   exit 1
 fi
 
+# Check for bzip2
+if ! command -v bzip2 >/dev/null 2>&1; then
```
salman1993: wouldn't we wanna check for both `tar` & `bzip2` here since `tar` is used later? the error msg can say you need to have bzip2 installed since we use `-j` flag

Screenshot 2025-04-02 at 10 18 32 AM

salman1993 (changes requested):
left a comment on checking for `tar`

download_cli.sh
teck968: I can add the check for the tar dependency. 

However, it appears to me that ChatGPT has provided you and me conflicting information as to the need for bzip2 to be installed. 

Here's what ChatGPT told me when I was researching the need for bzip2 on macOS:
image

I could perhaps implement a check that somehow determines if tar relies on the bzip2 command or if it has native bzip2 support, but I'm not sure how to approach this yet.

teck968: I've implemented the tar check, and a dynamic check for bzip2.

salman1993:
sorry for the back & forth on this. can we instead check the error code when running the `tar -xjf` command and if it fails, let the user know that bzip2 is required. that seems simpler 
https://github.com/block/goose/blob/b6df879136f8356fb43e6113c459ff4796912164/download_cli.sh#L160

teck968:
Ok, I've implemented the requested changes.

salman1993 (approved):

salman1993:
looks good! 🚢

PR #1999, opened 2025-04-02
feat: run tasks on a schedule (GUI only)

michaelneale:
this is a simple feature request, using a timer in the GUI can run scheduled tasks 

Screenshot 2025-04-02 at 12 12 39 pm

when running: 

Screenshot 2025-04-02 at 11 19 10 am


lilydelalande:
This is cool -- let's hold on this while we loop in design! Typically I think we should start features in a bit more lockstep with them -- there's a lot going on!

zakiali:
+1 to @lily-de comment. This is great, but should feel out the experience of this feature a bit more with design. It's not completely clear to me that the UI is the right place for something like this.

Issue #2003, opened 2025-04-02
Goose tries to connect to Ollama on the wrong port


sammcj:
**Describe the bug**

Goose seems to be incorrectly appending a port at the end of the Ollama host URL:

Pictured below you can see the URL to my Ollama server, except Goose has added ':11434' which is not a port I configured in the Ollama server URL (it's standard HTTPS, 443 in this case).

**To Reproduce**
Steps to reproduce the behaviour:
1. Set your Ollama server URL to something valid such as `https://ollama.my.network`
2. Try to use Goose
3. Experience error
4. Note that the URL in the error has had the default localhost Ollama port appended to it

**Expected behaviour**

Goose to respect Ollama server URLs

**Screenshots**

<img width="751" alt="Image" src="https://github.com/user-attachments/assets/d9e17296-8ea7-46f0-927b-5a1fd494bbe7" />

**Please provide following information:**
 - **OS & Arch:** macOS (client), Fedora (server)
 - **Interface:** UI
 - **Version:** 1.0.16
 - **Extensions enabled:** N/A
 - **Provider & Model:** Ollama, Any

**Additional context**

sammcj:
Fix in PR https://github.com/block/goose/pull/2004

alicehau:
Thanks for the fix, merged it in!

PR #2004, opened 2025-04-02
fix(ollama): respect ollama URL and port

sammcj:
- Fix: Respect configured Ollama URL and port

fix for #2003


alicehau (approved):
thanks for the fix!

PR #2013, opened 2025-04-02
feat: update the google drive update file tool description

yingjiehe-xyz:
After reviewing the code, they are very confusing as we have update_file in https://github.com/block/goose/blob/4846d15354a1a1c53381e8e4886f4b23b606f62d/crates/goose-mcp/src/google_drive/mod.rs#L421.

It turns out update is for normal file while update_file is for google related files. Refresh the name and descriptions

### DIFF
--- a/crates/goose-mcp/src/google_drive/mod.rs
+++ b/crates/goose-mcp/src/google_drive/mod.rs
@@ -377,8 +377,8 @@ impl GoogleDriveRouter {
             }),
         );
 
-        let update_tool = Tool::new(
-            "update".to_string(),
+        let update_file_tool = Tool::new(
+            "update_file".to_string(),
             indoc! {r#"
                 Update a normal non-Google file (not Document, Spreadsheet, and Slides) in Google Drive with new content.
             "#}
@@ -418,8 +418,8 @@ impl GoogleDriveRouter {
             }),
         );
 
-        let update_file_tool = Tool::new(
-            "update_file".to_string(),
+        let update_google_file_tool = Tool::new(
+            "update_google_file".to_string(),
             indoc! {r#"
                 Update a Google file (Document, Spreadsheet, or Slides) in Google Drive.
             "#}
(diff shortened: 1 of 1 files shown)


sheagcraig:
Hi there—this is not a duplicate; it's (sadly) intentional. The function you have removed in this PR is for updating non-Gdocs files (e.g. a text file, an image, basically anything that isn't Docs/Slides/Sheets).

The function at L421 is the one for updating a Docs/Sheets/Slides file.

These were separated to follow the same pattern as the `upload` and `create_file` tools, which separate non-GDocs files from Gdocs files.

yingjiehe-xyz:
> Hi there—this is not a duplicate; it's (sadly) intentional. The function you have removed in this PR is for updating non-Gdocs files (e.g. a text file, an image, basically anything that isn't Docs/Slides/Sheets).
> 
> The function at L421 is the one for updating a Docs/Sheets/Slides file.
> 
> These were separated to follow the same pattern as the `upload` and `create_file` tools, which separate non-GDocs files from Gdocs files.

Oh, thanks for the clarifications! Let me update the description. And can we merge them into one? Most of the code looks the same

sheagcraig:
I originally wrote it as one, but the amount of extra prompting that then falls on the user to make it explicit that they want to create a Google Doc from a markdown, for example, made it make more sense from a user perspective to make it clear that one was for uploading any old file to Drive; and one was for creating Gdocs.

salman1993:
@yingjiehe-xyz @sheagcraig what do we wanna do with this PR? i'll let you folks decide the best way forward

yingjiehe-xyz:
> @yingjiehe-xyz @sheagcraig what do we wanna do with this PR? i'll let you folks decide the best way forward

I am thinking to make it more clear on the tool name and description without big changes

sheagcraig:
I see the updated changes and they look good—they definitely help make it very clear which function does what.

sheagcraig:
...so I say, run with it! Thanks for including me :)

PR #2014, opened 2025-04-02
feat: config first lookup

lilydelalande:
Env vars come second to config and keychain


yingjiehe-xyz (approved):

lilydelalande:
closing because while this is not ideal or normal for a desktop app, this will be too big a change for cli users

PR #2015, opened 2025-04-02
ui: default model support

lily-de:
Starts using `GOOSE_DEFAULT_MODEL` as a fallback if it's in your env

### DIFF
--- a/ui/desktop/src/App.tsx
+++ b/ui/desktop/src/App.tsx
@@ -93,8 +93,8 @@ export default function App() {
       try {
         const config = window.electron.getConfig();
 
-        const provider = config.GOOSE_PROVIDER ?? (await read('GOOSE_PROVIDER', false));
-        const model = config.GOOSE_MODEL ?? (await read('GOOSE_MODEL', false));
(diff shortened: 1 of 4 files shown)


ui/desktop/src/App.tsx
```diff
@@ -93,8 +93,8 @@ export default function App() {
       try {
         const config = window.electron.getConfig();
 
-        const provider = config.GOOSE_PROVIDER ?? (await read('GOOSE_PROVIDER', false));
-        const model = config.GOOSE_MODEL ?? (await read('GOOSE_MODEL', false));
+        const provider = config.GOOSE_DEFAULT_PROVIDER ?? (await read('GOOSE_PROVIDER', false));
+        const model = config.GOOSE_DEFAULT_MODEL ?? (await read('GOOSE_MODEL', false));
```
kalvinnchau: wouldn't we want the opposite here? 

To get from the config first with `read` and then fallback to `DEFAULT` if they don't exist in the config?

I see in `getCurrentModelAndProvider` it tries to read first, then uses a fallback to get from defaults, then write to the config.

I'm not sure the distinction there and with this?

lilydelalande: good point!

alexhancock (approved):

PR #2023, opened 2025-04-03
feat: add tooltips, fix back button inverse color

salman1993:
Screenshot 2025-04-03 at 10 11 51 AM

### DIFF
--- a/crates/mcp-core/src/resource.rs
+++ b/crates/mcp-core/src/resource.rs
@@ -63,7 +63,7 @@ impl Resource {
             Some(n) => n,
             None => url
                 .path_segments()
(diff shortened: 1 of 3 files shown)


crates/mcp-core/src/resource.rs
```diff
@@ -63,7 +63,7 @@ impl Resource {
             Some(n) => n,
             None => url
                 .path_segments()
-                .and_then(|segments| segments.last())
+                .and_then(|mut segments| segments.next_back())
                 .unwrap_or("unnamed")
```
salman1993: unrelated to this PR but cargo clippy was failing

yingjiehe-xyz (approved):

PR #2027, opened 2025-04-03
fix: properly return errors for malformed config

alexhancock:
Right now if `config.yaml` is malformed or cannot deserialize to the right schema, it just responds that there is an empty list of extensions. This makes it respond specifically with an `UNPROCESSABLE_ENTITY` status code. In a next change, we can update the frontend to handle this by showing options to the user (edit the file, or reset config file)

### DIFF
--- a/crates/goose-server/src/routes/config_management.rs
+++ b/crates/goose-server/src/routes/config_management.rs
@@ -185,7 +185,17 @@ pub async fn get_extensions(
 
     match ExtensionManager::get_all() {
         Ok(extensions) => Ok(Json(ExtensionResponse { extensions })),
-        Err(_) => Err(StatusCode::INTERNAL_SERVER_ERROR),
+        Err(err) => {
+            // Return UNPROCESSABLE_ENTITY only for DeserializeError, INTERNAL_SERVER_ERROR for everything else
+            if err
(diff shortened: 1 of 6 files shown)


crates/goose-server/src/routes/config_management.rs
```diff
@@ -185,7 +185,17 @@ pub async fn get_extensions(
...
     match ExtensionManager::get_all() {
         Ok(extensions) => Ok(Json(ExtensionResponse { extensions })),
-        Err(_) => Err(StatusCode::INTERNAL_SERVER_ERROR),
+        Err(err) => {
+            // Return UNPROCESSABLE_ENTITY only for DeserializeError, INTERNAL_SERVER_ERROR for everything else
+            if err
+                .downcast_ref::<goose::config::base::ConfigError>()
+                .is_some_and(|e| matches!(e, goose::config::base::ConfigError::DeserializeError(_)))
+            {
+                Err(StatusCode::UNPROCESSABLE_ENTITY)
```
baxen, line 194: nit/not blocking: i've found it very hard to debug these without there being an error message on the result with more info

baxen (approved):
LGTM

kalvinnchau (approved):

PR #2029, opened 2025-04-03
E2E: added provider switching and MCP testing

zanesq:
Added the following:
- Organized tests to general ui and provider testing
- MCP setup and testing with a local example MCP
- Provider switching. Currently only databricks and google - you'll need google set up locally before hand to work -- need to think about how we can make provider setup more dynamic for people.

Bonus:
- Overlay showing the currently running test
- Video recording

Example test run:
https://github.com/user-attachments/assets/e803cf62-fe38-4375-b163-f227f1616cd9


ui/desktop/tests/e2e/app.spec.ts
```diff
@@ -1,20 +1,131 @@
...
+  testPath: string;
+};
+
+// Create test fixture type
+type TestFixtures = {
+  provider: Provider;
+};
+
+// Define available providers
+const providers: Provider[] = [
```
zanesq, line 23: Maybe an improvement for the provider selection is we could iterate through the providers that are enabled in the UI instead of hard coding them. That way we could also potentially cover more ground with other providers.

lily-de (approved):

ui/desktop/tests/e2e/basic-mcp.ts
```diff
@@ -0,0 +1,71 @@
+const { McpServer } = require("@modelcontextprotocol/sdk/server/mcp.js");
+const { StdioServerTransport } = require("@modelcontextprotocol/sdk/server/stdio.js");
+
+// Collection of running-related inspirational quotes
```
alexhancock, line 4: Yes! The `running-quotes` server doing work beyond presentations now :)

PR #2035, opened 2025-04-04
(bug-fix: 1984) Update search highlighter to use overlay

JJSwigut:
## **Problem**

Our search highlighting implementation was causing crashes when new chat messages were added. The previous implementation directly modified the DOM by injecting `<mark>` tags into text nodes, which conflicted with React's rendering cycle when content was updated. Ultimately, I'm pleased with this since the following is a much more robust and scalable solution IMO. Shoutout to @zanesq and @salman1993 for finding and confirming the issue before it was released!

## **Solution**

We've updated our existing `SearchHighlighter` class to use an overlay-based approach instead of direct DOM manipulation. This is actually how many browsers, Notion, and other large products do it to avoid messing with the elements themselves:

1. **Separate Rendering Layer**: Instead of modifying React-managed DOM nodes, we now render highlights in a separate overlay layer that doesn't interfere with the content React is managing
2. **Content-Change Resilience**: This approach allows React to freely update the underlying content while our highlights adjust to match the new positions

## **Key Changes**

1. Modified the `SearchHighlighter` class to:
    - Create highlights in an absolute-positioned overlay instead of injecting `<mark>` tags
    - Update highlight positions when content changes or scrolls
    - Maintain current highlight state across updates
    - Control scrolling behavior more precisely
2. Updated the `SearchView` component to:
    - Work with the new overlay implementation
    - Clean up resources properly
    - Handle keyboard navigation correctly
3. Updated CSS to target the new highlight elements

This architectural improvement prevents crashes by avoiding direct manipulation of DOM nodes that React is managing, allowing both systems to operate independently without conflicts.

##Demo
<video src =https://github.com/user-attachments/assets/59360dba-62d8-49a4-bc68-57bcc8a91e92 width = 600>

### DIFF
--- a/ui/desktop/eslint.config.js
+++ b/ui/desktop/eslint.config.js
@@ -86,6 +86,10 @@ module.exports = [
         React: 'readonly',
         handleAction: 'readonly',
         requestAnimationFrame: 'readonly',
+        ResizeObserver: 'readonly',
+        MutationObserver: 'readonly',
+        NodeFilter: 'readonly',
+        Text: 'readonly',
       },
     },
     plugins: {
--- a/ui/desktop/src/components/ChatView.tsx
+++ b/ui/desktop/src/components/ChatView.tsx
@@ -388,7 +388,7 @@ export default function ChatView({
           />
         ) : (
           <ScrollArea ref={scrollRef} className="flex-1" autoScroll>
-            <SearchView scrollAreaRef={scrollRef}>
+            <SearchView>
               {filteredMessages.map((message, index) => (
                 <div key={message.id || index} className="mt-4 px-4">
                   {isUserMessage(message) ? (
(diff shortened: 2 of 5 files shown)


zanesq:
@JJSwigut thanks much better approach for highlighting with the overlay! One thing I noticed when testing your branch is it doesn't seem to scroll to the next/prev term when I click the arrows (see below). I actually came across this in your original implementation and pushed up this [fix](https://github.com/block/goose/pull/1982) if it helps at all.

Kapture 2025-04-04 at 08 42 04

JJSwigut:
> @JJSwigut thanks much better approach for highlighting with the overlay! One thing I noticed when testing your branch is it doesn't seem to scroll to the next/prev term when I click the arrows (see below). I actually came across this in your original implementation and pushed up this [fix](https://github.com/block/goose/pull/1982) if it helps at all.

@zanesq Appreciate you and your responsiveness!

Seems like this time it was actually an issue with the timing of mouse clicks vs keys. Added a `setTimeOut` and it smoothed right out. Also made both keys and mouse use the same `handleNavigate` function for consistency.

PR #2042, opened 2025-04-04
feat: add migration logic from settingsV1 to settingsV2

kalvinnchau:
Migration logic:
* on startup, check for `configVersion` in `localStorage` if it (doesn’t exist, or is < 2) proceed with migration
* grab extensions from `user_settings` in `localStorage`
  * convert and extract all extensions
  * iterate through and add them to `config.yaml`, skipping `builtins`
* set `configVersion: 2` in `localStorage` so this only happens once

we skip `builtin` and rely on `initializeBundledExtensions` or `syncBundledExtensions` to setup all bundled extensions in the new format

we then just drop into normal initialization logic afterwards, relying on normal agent/extension startup

### DIFF
--- a/ui/desktop/src/utils/providerUtils.ts
+++ b/ui/desktop/src/utils/providerUtils.ts
@@ -1,5 +1,5 @@
 import { getApiUrl, getSecretKey } from '../config';
-import { loadAndAddStoredExtensions } from '../extensions';
+import { FullExtensionConfig, loadAndAddStoredExtensions } from '../extensions';
 import { GOOSE_PROVIDER, GOOSE_MODEL } from '../env_vars';
 import { Model } from '../components/settings/models/ModelContext';
 import { gooseModels } from '../components/settings/models/GooseModels';
@@ -59,7 +59,7 @@ Some extensions are builtin, such as Developer and Memory, while
 `;
 
 // Desktop-specific system prompt extension when a bot is in play
-const desktopPromptBot = `You are a helpful agent. 
+const desktopPromptBot = `You are a helpful agent.
 You are being accessed through the Goose Desktop application, pre configured with instructions as requested by a human.
 
 The user is interacting with you through a graphical user interface with the following features:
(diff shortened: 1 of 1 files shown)


ui/desktop/src/utils/providerUtils.ts
```diff
@@ -73,6 +73,55 @@ You can also validate your output after you have generated it to ensure it meets
...
+
+  for (const extension of localStorageExtensions) {
+    // NOTE: skip migrating builtin types since there was a format change
+    // instead we rely on initializeBundledExtensions & syncBundledExtensions
+    // to handle updating / creating the new builtins to the config.yaml
+    // For all other extension types we migrate them to config.yaml
+    if (extension.type !== 'builtin') {
+      console.log(`Migrating extension ${extension.name} to config.yaml`);
+      try {
+        await addExtension(extension.name, extension, extension.enabled);
```
alexhancock: Should probably also enable the extension on the agent if it was enabled before?

alexhancock (commented):
Nice! Let me know if you want to change anything and I can do a final review but this is looking good

ui/desktop/src/utils/providerUtils.ts
kalvinnchau: yep! that happens further down - it re-loads the config.yaml and determines which to enable checking that flag, this is just setting the flag to match the existing localstate flag

alexhancock (approved):

ui/desktop/src/utils/providerUtils.ts
alexhancock: Oh fantastic, were able to use the existing code. Nice

ui/desktop/src/utils/providerUtils.ts
```diff
@@ -73,6 +76,79 @@ You can also validate your output after you have generated it to ensure it meets
...
+  if (migrationErrors.length === 0) {
+    localStorage.setItem('configVersion', '2');
+    console.log('Extension migration complete. Config version set to 2.');
+  } else {
+    const errorSummaryStr = migrationErrors
+      .map(({ name, error }) => `- ${name}: ${JSON.stringify(error)}`)
+      .join('\n');
+    toastService.error({
+      title: 'Config Migration Error',
+      msg: 'Extension failed to start and will be disabled.',
```
alexhancock: Should this message be something more general, like "We had trouble updating your config"?

alexhancock (approved):

ui/desktop/src/utils/providerUtils.ts
kalvinnchau: whoops yes thats a copy paste and i forgot to update that

PR #2044, opened 2025-04-04
feat: start use tool permission confirmation struct

yingjiehe-xyz:
In this PR, we 
1. Use the tool permission confirmation instead of boolean value.
2. Stop writing permission in to permission store as we will save different info in the store later


wendytang (approved):

crates/goose/src/permission/permission_confirmation.rs
```diff
@@ -0,0 +1,21 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+pub enum Permission {
+    AlwaysAllow,
+    AllowOnce,
+    AlwaysDeny,
```
wendytang: Should this just be `DenyOnce`? `AlwaysDeny` sounds like future tool requests are going to be automatically denied

yingjiehe-xyz: make sense, thanks for the catch!

PR #2047, opened 2025-04-05
Make claude-3.7-sonnet the default for OpenRouter

alexgleason:
To not have to keep manually pasting over this:

Screenshot from 2025-04-05 16-42-02


salman1993:
We've had a bunch of testing with sonnet 3.7 now. What's been very interesting is that it hasn't felt like a strict upgrade. It is certainly way better at some specific stuff like string replaces. It is also much more likely to finish longer tool use loops.  But it's also more likely to do unrelated changes, or to head down a rabbit hole - and generally needs pretty strict prompting to not take premature actions.

PR #2049, opened 2025-04-06
Fix OpenRouter error handling

alexgleason:
When there's an error in the OpenRouter API, Goose doesn't respond:

image

After this fix, it correctly reads the error from the response:

image


michaelneale:
very cool @alexgleason. Also I am glad I am not alone in typing "hello...." when things seem to have gone wrong and not responding.

BTW if you ask goose to run "cargo fmt" and "cargo clippy -- -D warnings" it can fix up any cosmetic or warning -> error messages from formatting and styles that may fail the build.

alexgleason:
Thanks @michaelneale, done!

crates/goose/tests/providers.rs
```diff
@@ -233,8 +233,8 @@ impl ProviderTester {
         dbg!(&result);
         println!("===================");
 
-        // Ollama and OpenRouter truncate by default even when the context window is exceeded
-        if self.name.to_lowercase() == "ollama" || self.name.to_lowercase() == "openrouter" {
```
alexgleason, line 237: @salman1993 I figured out that this test assumption is actually wrong. It only didn't return an error because the error wasn't being captured.

Tests are passing now. Please try again.

salman1993 (approved):

crates/goose/tests/providers.rs
salman1993: thank you!

alexgleason:
Anything still blocking this?

Issue #2051, opened 2025-04-06
Misleading error log in SseActor when SSE loop ends normally


jokellyfenton:
**Describe the bug**

When using Goose via the fastmcp client, the following log line is emitted at error level even though it does not represent an actual error condition:

```
tracing::error!("SSE stream ended or encountered an error; clearing pending requests.");
```
[Code Reference](https://github.com/block/goose/blob/main/crates/mcp-client/src/transport/sse.rs#L141-L143)

This log message is misleading, as it suggests a fault occurred even when everything is functioning as expected (the SSE stream has ended). It unnecessarily pollutes logs and may trigger alerts in production environments. 

**To Reproduce**
1. Create a client using fastmcp or goose
2. Run a request that opens an SSE connection
3. Allow the connection to close naturally or via a timeout
4. Observe that an error log is emitted even when no actual error occurred

**Expected behavior**
Given this log line CAN occur when an error occurs (unexpected error when there are pending requests) it would be ideal to split this logging up. For example: 

```
let pending = pending_requests.len().await;
if pending > 0 {
    tracing::error!("SSE stream ended or encountered an error with {pending} unfulfilled pending requests.");
} else {
    tracing::info!("SseActor shutdown cleanly. No pending requests.");
}
```

**Screenshots**
N/A

**Please provide following information:**
 - **OS & Arch:** macOS 15.3.2 arm64
 - **Interface:** CLI
 - **Version:** [e.g. v1.0.2]
 - **Extensions enabled:** N/A
 - **Provider & Model:** All models

**Additional context**
If this error is in fact avoidable, please let me know so that we can action on our side! Thanks!!

salman1993:
this makes sense. do you wanna put in a PR for a fix?

jokellyfenton:
Awesome! Thank you for your prompt response @salman1993 Here is the PR: https://github.com/block/goose/pull/2052

salman1993:
merged the PR - thanks for adding it in!

jokellyfenton:
Hey @salman1993 

When is your next release scheduled for? 

Thanks!

PR #2052, opened 2025-04-06
Update SSEActor error logging

jokellyfenton:
Motivation:
* When using Goose via the fastmcp client, there is a log line that is emitted at error level even though it does not represent an actual error condition. See bug: https://github.com/block/goose/issues/2051

Approach:
* If there are no remaining pending requests in the queue and the server has been shut down, the logging should be info level. Otherwise if there are messages remaining and the server has been closed there should be an error message logged.
* Add two new methods to PendingRequests -- len and is_empty. Only len is used but is_empty is required by rust: https://rust-lang.github.io/rust-clippy/master/index.html#len_zero

### DIFF
--- a/crates/mcp-client/src/transport/mod.rs
+++ b/crates/mcp-client/src/transport/mod.rs
@@ -118,6 +118,14 @@ impl PendingRequests {
     pub async fn clear(&self) {
         self.requests.write().await.clear();
     }
+
+    pub async fn len(&self) -> usize {
+        self.requests.read().await.len()
+    }
+
+    pub async fn is_empty(&self) -> bool {
+        self.len().await == 0
+    }
 }
 
 pub mod stdio;
--- a/crates/mcp-client/src/transport/sse.rs
+++ b/crates/mcp-client/src/transport/sse.rs
@@ -210,8 +210,13 @@ impl SseActor {
         }
 
         // mpsc channel closed => no more outgoing messages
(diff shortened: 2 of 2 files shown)


michaelneale:
Seems like a good idea

jokellyfenton:
Fixed issue and reran cargo fmt, cargo clippy and cargo test. Are you able to review when ready thank you! @michaelneale @salman1993

crates/mcp-client/src/transport/mod.rs
```diff
@@ -118,6 +118,14 @@ impl PendingRequests {
     pub async fn clear(&self) {
         self.requests.write().await.clear();
     }
+
+    pub async fn len(&self) -> usize {
+        self.requests.read().await.len()
+    }
+
+    pub async fn is_empty(&self) -> bool {
```
jokellyfenton, line 126: This method required by rust when len is implemented - https://rust-lang.github.io/rust-clippy/master/index.html#len_zero

salman1993 (approved):

taniacryptid:
Hi @jokellyfenton ! Wanted to share that you are one of the Top Contributors to goose for the month of April! As super thanks for your contributions, we've given you $20 LLM credits. Just use this API key as you would any OpenRouter key! But to be able to give it to you, would have to message you on [Discord](https://discord.gg/block-opensource).

PR #2054, opened 2025-04-07
option of using git to make checkpoints as tasks as completed or rolled back

michaelneale:
for now

`GOOSE_GIT_CHECKPOINTING=true .../path to /goose session`

to try this out. Adds to developer MCP based on that flag being on only.


michaelneale:
[goose.zip](https://github.com/user-attachments/files/19624535/goose.zip)

pre-made for macos aarch64 (download it, unzip will need to run `xattr -c ... goose` if you download it via chrome, most likely)

Then run: 

`GOOSE_GIT_CHECKPOINTING=true ~/Downloads/goose session`

lilydelalande (approved):

PR #2060, opened 2025-04-07
feat: add permission manager for tool permission control

yingjiehe-xyz:
We have three levels for permission: Always Allow, Ask before and Deny. Users can set the permission level for each tool, however, we still have the smart approve and manual approve permission level as the default for the starting point. 

The permission yaml format will be:
```
user:
 always_allow:
    - xxxx
 ask_before:
    - xxxx
 deny:
    - xxxx
smart_approve:
 ....
```

It means the workflow should be like:
Note: only in smart approve and manual approve goose mode
1. Check the user permission level first
2. If find the tool permission, use it
3. If not, base on current goose mode, if it is manual approve mode, always be Ask before; if it is smart_approve, get the permission level(either always allow - readonly or ask before - write/mixed)

The reason we store smart approve default permission is to cache the llm-as-a-judge result.

### DIFF
--- a/crates/goose/src/config/mod.rs
+++ b/crates/goose/src/config/mod.rs
@@ -1,11 +1,13 @@
 pub mod base;
 mod experiments;
 pub mod extensions;
+pub mod permission;
 
 pub use crate::agents::ExtensionConfig;
 pub use base::{Config, ConfigError, APP_STRATEGY};
 pub use experiments::ExperimentManager;
 pub use extensions::{ExtensionEntry, ExtensionManager};
+pub use permission::PermissionManager;
 
 pub use extensions::DEFAULT_DISPLAY_NAME;
 pub use extensions::DEFAULT_EXTENSION;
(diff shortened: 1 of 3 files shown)


wendytang (approved):

lilydelalande:
Is there a reason we aren't adding permissions to the `config.yaml` file but its own separate file?

this adds some complexity where we read and write to different files for configuration-based data

yingjiehe-xyz:
> Is there a reason we aren't adding permissions to the `config.yaml` file but its own separate file?
> 
> this adds some complexity where we read and write to different files for configuration-based data

sorry for missing the comment, I was thinking to add permission list to the existing config file before, since the permission list can be very long, in the worst case for average users(5 extensions with 25 tools in total - avg 5 tools per extension), we may have extra 50+(user configured and smart approve permission cache) lines for permission, it makes the config file hard to read. So to keep config yaml consice, I move these permissions to a new file. Hope it makes sense to you

PR #2080, opened 2025-04-08
feat: add permission field to the list tools response

yingjiehe-xyz:
Add list query filter to list tools for specified extensions. And use ToolInfo instead of Tool for the response because ToolInfo contains the enough info and adding permission field doesn't have any impact on existing code, while Tool is aligning with MCP Tool definition.

### DIFF
--- a/ui/desktop/src/api/types.gen.ts
+++ b/ui/desktop/src/api/types.gen.ts
@@ -84,6 +84,11 @@ export type ExtensionResponse = {
     extensions: Array<ExtensionEntry>;
 };
 
+/**
+ * Enum representing the possible permission levels for a tool.
+ */
+export type PermissionLevel = 'always_allow' | 'ask_before' | 'never_allow';
+
 export type ProviderDetails = {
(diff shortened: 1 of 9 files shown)


ui/desktop/src/api/types.gen.ts
```diff
@@ -204,6 +209,16 @@ export type ToolAnnotations = {
...
 
+/**
+ * Information about the tool used for building prompts
+ */
+export type ToolInfo = {
+    description: string;
+    name: string;
+    parameters: Array<string>;
+    permission?: PermissionLevel | null;
+};
```
lilydelalande, line 220: does this also include info on which extension the tool comes from?

lily-de (approved):

ui/desktop/src/api/types.gen.ts
yingjiehe-xyz: The name has the indicator, for example, the tool name is <extension_name>__<tool_name>

PR #2083, opened 2025-04-08
feat: add temperature env var

Tyler-Hardin:
No hard feelings if you'd prefer this implemented a different way. Also apologies if there is already a way to set temp.


michaelneale:
I like this, seems pretty handy way to play around with temperature

Tyler-Hardin:
Added the test Copilot suggested.

lily-de (approved):

PR #2088, opened 2025-04-08
feat: add google_drive uri verification, and example instructions

kalvinnchau:
# google_drive uri verification

saw multiple instances where goose would try
```
gdrive:///document/$SOME_DOC_ID
```
and fail, not knowing that it needed to only pass in `$SOME_DOC_ID`

* add instructions on mapping URLs to URI format expected (examples are random IDs that don't exist)
* add a check in tool call to look for any extra `/` chars and error out on that

could not find a spec on what a valid ID is, so this is best effort

### DIFF
--- a/crates/goose-mcp/src/google_drive/mod.rs
+++ b/crates/goose-mcp/src/google_drive/mod.rs
@@ -1118,6 +1118,14 @@ impl GoogleDriveRouter {
 
         let drive_uri = uri.replace("gdrive:///", "");
 
+        // Validation: check for / path separators as invalid uris
+        if drive_uri.contains('/') {
+            return Err(ToolError::InvalidParameters(format!(
+                "The uri '{}' conatins extra '/'. Only the base URI is allowed.",
+                uri
+            )));
+        }
+
         let include_images = params
             .get("includeImages")
             .and_then(|i| i.as_bool())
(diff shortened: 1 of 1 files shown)


yingjiehe-xyz (approved):

crates/goose-mcp/src/google_drive/mod.rs
```diff
@@ -224,6 +224,11 @@ impl GoogleDriveRouter {
             indoc! {r#"
                 Read a file from google drive using the file uri.
                 Optionally include base64 encoded images, false by default.
+
+                Example extracting URIs from URLs:
+                Given "https://docs.google.com/document/d/1QG8d8wtWe7ZfmG93sW-1h2WXDJDUkOi-9hDnvJLmWrc/edit?tab=t.0#heading=h.5v419d3h97tr"
```
yingjiehe-xyz, line 229: avoid the real doc id?

kalvinnchau: that's a fake doc id for me, does it open something real for you?

yingjiehe-xyz: oh, I see, thought it was some real docs, looks good

PR #2091, opened 2025-04-08
refactor: remove agent flavours, move provider to Agent

salman1993:
Changes:
- get rid of agent flavours, move truncate agent to the single Agent struct
- move provider to Agent & use Arc (instead of Arc<Box<...>>) - see https://users.rust-lang.org/t/confused-between-box-rc-cell-arc/10946
- rename Capabilities -> ExtensionManager (strictly responsible for managing MCPs, removed logic for frontend tools)
- rename prev ExtensionManager -> ExtensionConfigManager (more specific to what it does)
- move frontend_tools to the Agent and extra logic of frontend tools in building system prompt & dispatching tool call
- move system prompt logic to prompt_manager.rs
- move platform tool definitions to platform_tools.rs

Follow up PRs:
1) reduce # lines of code in agent.reply(...)

2) add agent config to store context mgmt strategy, max truncation attempts, etc. will add summarize and noop as context mgmt strategy


crates/goose/src/agents/prompt_manager.rs
```diff
@@ -0,0 +1,97 @@
+// agent/prompt_builder.rs
```
kalvinnchau: nit: probably not needed - doesn't match

crates/goose/src/agents/prompt_manager.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        // Add frontend tools as a special extension if any exist
+        let frontend_instructions = frontend_instructions.unwrap_or(
+            "The following tools are provided directly by the frontend and will be executed by the frontend when called.".to_string()
+        );
+        extensions_info.push(ExtensionInfo::new(
+            "frontend",
+            &frontend_instructions,
+            false,
+        ));
```
kalvinnchau: shouldn't this be an `if let Some(frontend_instructions) = frontend_instructions {...}`? 
we would only want to add `frontend` extension info if this is `Some` and not default to adding it in the `None` case

crates/goose/src/agents/agent.rs
```diff
@@ -1,76 +1,801 @@
...
+                    "The extension '{}' has been installed successfully",
+                    extension_name
+                ))]
+            })
+            .map_err(|e| ToolError::ExecutionError(e.to_string()));
+
+        (request_id, result)
+    }
+
+    // Previously, the agent trait was implemented here but now they're methods in the Agent struct
```
kalvinnchau: nit: probably don't need this comment

kalvinnchau (commented):
this is looking really good! love the simplification of passing around the agent 

some minor nits, one comment to resolve

---

> reduce # lines of code in agent.reply(...)

🙏 , going through reply is a pretty wild ride right now

crates/goose/src/agents/prompt_manager.rs
salman1993: good catch!

crates/goose/src/agents/agent.rs
```diff
@@ -1,76 +1,804 @@
...
+            "input" = serde_json::to_string(&tool_call).unwrap(),
+            "output" = serde_json::to_string(&result).unwrap(),
+        );
+
+        (request_id, result)
+    }
+
+    /// Truncates the messages to fit within the model's context window
+    /// Ensures the last message is a user message and removes tool call-response pairs
+    async fn truncate_messages(
```
zakiali, line 152: how are you thinking about other context management strategies here? If we included summarization or something, how does that fit in

zakiali (approved):
looks great!

crates/goose/src/agents/agent.rs
salman1993: i wanna add in an agentconfig struct where users can specify context mgmt strategy, other params like max_truncation_attempts (this will be after refactoring reply)

PR #2096, opened 2025-04-08
feat: configure tool level permission in CLI

yingjiehe-xyz:
Test:

```
┌   goose-configure 
│
│  Goose Settings 
│
◇  What setting would you like to configure?
│  Tool Permission 
│
◇  Choose an extension to configure tools
│  developer 
│
◇  Choose a tool to update permission
│  developer__screen_capture 
│
└  Configuring tool: developer__screen_capture, Current permission level: Always Allow
◇  Set permission level for tool developer__screen_capture
│  Always Allow 
│
└  Updated permission level for tool developer__screen_capture to Always Allow.
```


wendytang (approved):
Nice!

PR #2099, opened 2025-04-09
Make GOOSE_CLI_THEME persist to config file and stick for future session

sana-db:
Previously, theme was not persisted across sessions because it was reading from an environment variable. Now, theme is also stored in config and read from there.

Testing:
Changed to Light mode in a session:
Screenshot 2025-04-08 at 5 20 30 PM

In a new terminal, started a new session and verified it stayed in light mode:
Screenshot 2025-04-08 at 5 22 48 PM

### DIFF
--- a/crates/goose-cli/src/session/output.rs
+++ b/crates/goose-cli/src/session/output.rs
@@ -26,7 +26,7 @@ impl Theme {
         }
     }
 
-    fn from_string(val: &str) -> Self {
+    fn from_config_str(val: &str) -> Self {
         if val.eq_ignore_ascii_case("light") {
             Theme::Light
         } else if val.eq_ignore_ascii_case("ansi") {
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/session/output.rs
```diff
@@ -25,26 +25,42 @@ impl Theme {
...
     );
 }
 
 pub fn set_theme(theme: Theme) {
+    let config = Config::global();
+    config
+        .set_param(
+            "GOOSE_CLI_THEME",
+            Value::String(match theme {
+                Theme::Light => "light".to_string(),
```
hasnain-db: should this be moved to a to_config_str function above?

salman1993 (approved):
🙏🏼

salman1993:
`cargo clippy` is failing:
```
error: methods with the following characteristics: (`to_*` and `self` type is `Copy`) usually take `self` by value
  --> crates/goose-cli/src/session/output.rs:39:25
   |
39 |     fn to_config_string(&self) -> String {
   |                         ^^^^^
   |
   = help: consider choosing a less ambiguous name
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
   = note: `-D clippy::wrong-self-convention` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::wrong_self_convention)]`
```

PR #2102, opened 2025-04-09
feat: auto-initialize goose config.yaml

alexhancock:
* Sets up code in the goose app where we can autopopulate `~/.config/goose/config.yaml` from an `init-config.yaml` present at the workspace route if it exists


crates/goose-server/src/routes/config_management.rs
```diff
@@ -320,6 +323,73 @@ pub async fn providers(
...
+
+    let config = Config::global();
+    
+    // 200 if config already exists
+    if config.exists() {
+        return Ok(Json("Config already exists".to_string()));
+    }
+
+    // Find the workspace root (where the top-level Cargo.toml with [workspace] is)
+    let workspace_root = match std::env::current_exe() {
```
alexhancock, line 346: @salman1993 @lily-de @baxen Let me know if there is a cleaner/simpler way to do this in rust. Will happily replace it in a followup PR!

PR #2113, opened 2025-04-09
Feature/azure credential chain

mpaquettePax8:
# Azure OpenAI Provider: Add Support for Azure Credential Chain

## Overview
This PR adds support for using the Azure credential chain with the Azure OpenAI provider. This allows users to authenticate using Azure CLI credentials or other Azure authentication methods when an API key is not provided.

### Authentication Flow
- Made `AZURE_OPENAI_API_KEY` optional in the provider configuration
- Added `AzureAuth` struct to handle both API key and default credential authentication
- Implemented token caching and refresh logic for Azure credentials
- Added proper error handling for authentication failures

### Configuration Updates
- Updated provider metadata to mark API key as optional
- Updated documentation to clarify that the API key is optional
- Added a new documentation section explaining the Azure credential chain functionality
- Added support for falling back to Azure credential chain when API key is not provided
- Maintained backward compatibility for existing API key authentication

### Retry and Error Handling
- Implemented exponential backoff for rate limit errors
- Added proper handling for various HTTP status codes (408, 429)
- Enhanced error logging for authentication and request failures
- Added token refresh on each retry attempt to handle token expiration

### Code Quality
- Added debug logging for important error cases
- Improved error messages for better debugging
- Maintained existing logging for usage data errors
- Fixed code formatting in Azure provider files

## Testing Done
- Verified authentication works with:
  - API key provided
  - Azure CLI credentials (no API key)
- Tested rate limit handling and retries
- Confirmed proper error handling for various failure scenarios

## Breaking Changes
None. This change maintains backward compatibility with existing API key authentication while adding support for Azure credential chain.

## Usage Example
Users can now configure the Azure OpenAI provider in two ways:

1. With API key (existing method):
```bash
goose configure
# Enter AZURE_OPENAI_API_KEY, AZURE_OPENAI_ENDPOINT, and AZURE_OPENAI_DEPLOYMENT_NAME
```

2. With Azure credentials (new method):
```bash
az login  # Login with Azure CLI
goose configure
# Only enter AZURE_OPENAI_ENDPOINT and AZURE_OPENAI_DEPLOYMENT_NAME
# Leave AZURE_OPENAI_API_KEY empty to use Azure credential chain
```

## Documentation Updates
- Added a dedicated section in the documentation explaining the Azure OpenAI Credential Chain feature
- Updated the provider table to mark the API key as optional
- Updated the provider description to mention support for both authentication methods


crates/goose/src/providers/azure.rs
```diff
@@ -18,15 +21,33 @@ pub const AZURE_DOC_URL: &str =
...
+#[derive(Debug)]
 pub struct AzureProvider {
-    #[serde(skip)]
     client: Client,
+    auth: AzureAuth,
     endpoint: String,
-    api_key: String,
     deployment_name: String,
     api_version: String,
-    model: ModelConfig,
```
lilydelalande, line 29: is there a reason this was removed?

crates/goose-cli/src/commands/configure.rs
```diff
@@ -266,49 +262,115 @@ pub async fn configure_provider_dialog() -> Result<bool, Box<dyn Error>> {
                     Ok(_) => {
                         let _ = cliclack::log::info(format!("{} is already configured", key.name));
                         if cliclack::confirm("Would you like to update this value?").interact()? {
-                            let new_value: String = if key.secret {
-                                cliclack::password(format!("Enter new value for {}", key.name))
-                                    .mask('▪')
-                                    .interact()?
+                            let prompt = if key.required {
```
michaelneale: I don't quite follow why there are cli changes when it is azure specific  - what would be effect on other providers?

michaelneale (changes requested):

ahau-square:
The path where you don't enter an API key and it falls back to Azure Auth doesn't seem to be working for me - suggesting still that the key is required

image

crates/goose-cli/src/commands/configure.rs
ahau-square: +1 here, definitely appreciate the effort to still allow users to input their api keys via CLI dialog! But I think we could instead just leave the CLI as it is, remove the API key as required, and allow users to set the api key via their env var if they want to which should get picked up by the provider. That should hopefully simplify this

This would be similar to how the bedrock provider is implemented: https://github.com/block/goose/blob/main/crates/goose/src/providers/bedrock.rs

ahau-square (changes requested):

crates/goose/src/providers/azure.rs
mpaquettePax8: ah no this should be reverted so we use the model the user specifies

crates/goose-cli/src/commands/configure.rs
mpaquettePax8: On it!

ahau-square:
Thank you for this contribution, merged in this PR: https://github.com/block/goose/pull/2413 ! Sorry it took a while to take a look at this but looks great!

PR #2117, opened 2025-04-09
draft: enable extension from ui

wendytang:


wendytang:
@lily-de wondering if you had any advice on: how should we handle updating the settings page when an extension is enabled in session? 
image

wendytang:
*frontend tools discussion*

ui/desktop/src/components/ExtensionConfirmation.tsx
```diff
@@ -0,0 +1,85 @@
+import React, { useState } from 'react';
```
ahau-square, line 1: this seems like a really similar component to the toolcallconfirmation.tsx component just with some different text, maybe create a more generalizable component and pass in the desired text as parameters so we don't duplicate components?

ahau-square (approved):

ahau-square:
> @lily-de wondering if you had any advice on: how should we handle updating the settings page when an extension is enabled in session? image

hm when you toggle from settings though, then the config.yaml would also be updated right? (could be ok but that's not what the cli does right now)

lilydelalande:
@ahau-square @wendytang 
> hm when you toggle from settings though, then the config.yaml would also be updated right? (could be ok but that's not what the cli does right now)

yeah what i was going to recommend was just update the config.yaml via functions exposed in `ConfigContext.tsx` and it will reflect in the settings page, but if the CLI is not updating the config then I guess we have a few options 

1. Update the config and create a new field called `temp` or something -- idk if this is possible but we can add some clean up processes that run when you close the window or the app. If the user opens a new window while previous session is ongoing (that added the temp extensions) we don't load extensions labeled with `temp` in the config 
2. We introduce session-specific tmp config -- probably merits bigger discussion but this is a good way for us to start to manage sessions with different states 
3. We just add them to the config anyway 
4. We add these extensions to localStorage in such a way that the storage is cleared when the session closes (consult with @alexhancock -- we really want to avoid this for non-UI specific configuration)

PR #2055, opened 2025-04-07
Gooselings in UI (name tbd)

zakiali:
Add's updated support for gooselings/mini agents/custom agents in the UI. Some updates:
- Allow users to add a name/description
- Add extensions to the created agent, and handle their state internally
- Updated agent editor

demo:
https://github.com/user-attachments/assets/03b40713-2972-4d30-a08b-78790cf92f94

PR #2119, opened 2025-04-09
UI Recipe/Custom Agents

zakiali:
add updated support for mini agents/custom agents in the UI. Some updates:

Allow users to add a name/description
Add extensions to the created agent, and handle their state internally
Updated agent editor
demo:
https://github.com/user-attachments/assets/03b40713-2972-4d30-a08b-78790cf92f94

https://github.com/block/goose/pull/2055


crates/goose-cli/src/session/mod.rs
```diff
@@ -500,6 +500,39 @@ impl Session {
                         }
                     }
 
+                    continue;
+                }
+                InputResult::Recipe(filepath_opt) => {
```
kalvinnchau: rebase/merge mess up? https://github.com/block/goose/blob/305c15225e8cb41a0c0520610bb4f19a15ee0e6f/crates/goose-cli/src/session/mod.rs#L472-L537

looks like this section is duplicated

crates/goose-server/src/routes/recipe.rs
```diff
@@ -0,0 +1,87 @@
...
+
+            Ok(Json(CreateRecipeResponse {
+                recipe: Some(recipe),
+                error: None,
+            }))
+        }
+        Err(e) => Ok(Json(CreateRecipeResponse {
+            recipe: None,
+            error: Some(e.to_string()),
+        })),
```
kalvinnchau: is this returning a HTTP 200, but error in the payload? should this be a 400?

kalvinnchau:
some UI notes:
Error page for dark mode doesn't flip font colors:
image
image


creating a custom agent while on dark mode, has all subsequent pages in light mode:
* pending page (laying an egg) is light mode
* create custom agent is light mode
* all the editing pages of activities, instructions, extensions are all light mode

little agent info, dark mode text not showing:
image
image

ui/desktop/src/App.tsx
kalvinnchau, line 1: still see botconfig references, should those be renamed too?

ui/desktop/src/components/Splash.tsx
```diff
@@ -2,18 +2,36 @@ import React from 'react';
...
+            <div className="relative text-textStandard mb-12">
+              <div className="w-min animate-[flyin_2s_var(--spring-easing)_forwards]">
+                <GooseLogo />
+              </div>
+            </div>
 
-        <div className="flex">
-          <SplashPills append={append} activities={activities} />
+            <div className="flex">
+              <SplashPills append={append} activities={activities} />
```
kalvinnchau, line 35: do we want to do any truncation or `....` for long pills?
image

ui/desktop/src/recipe/index.ts
```diff
@@ -0,0 +1,59 @@
+import { Message } from '../types/message';
```
kalvinnchau, line 1: could we make use of the openapi type generation so that we have the rust structs mapped 1:1 with typescript?

would need to add all the relevant stuff to https://github.com/block/goose/blob/main/crates/goose-server/src/openapi.rs, and the macros to the various structs that are in use

kalvinnchau (approved):

lilydelalande:
We found that when a URL with recipe configuration is processed:

1. If the Goose app is not running, it starts up and properly creates a window with the recipe config.
2. If the Goose app is running but has no windows open, it correctly creates a new window with the recipe config.
3. If the Goose app is running with windows already open, it attempts to process the recipe config in an existing window rather than creating a new one as intended.

Added some changes to main.ts to address but didn't push

PR #2120, opened 2025-04-10
Added nvmrc for our node version

zanesq:
Added nvmrc for our node version so we're all on the same page. Will update readme later with more info.


kalvinnchau (approved):
nice, should also add in package.json https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines too?

PR #2121, opened 2025-04-10
feat: add support for full url extraction for google_drive

kalvinnchau:
also support passing in the full URL to the `read` tool call and use a regex to extract the URI


michaelneale (approved):
nice catch

PR #2123, opened 2025-04-10
Fix re-renders from adding too many dependencies to useEffect

zanesq:
Fix re-renders from adding too many dependencies to useEffect.

We need to go through every one of our usages of useEffect at some point and check that we aren't doing unnecessary work but for now we can get past the linting errors with

`// eslint-disable-next-line react-hooks/exhaustive-deps`

It's a code smell to have to use these so its a todo to look into it later, should be fine for now 👍

### DIFF
--- a/ui/desktop/tests/e2e/app.spec.ts
+++ b/ui/desktop/tests/e2e/app.spec.ts
@@ -525,12 +525,6 @@ test.describe('Goose App', () => {
       
           // Wait a bit and dump HTML to see structure
           await mainWindow.waitForTimeout(1000);
-          const html = await mainWindow.evaluate(() => document.documentElement.outerHTML);
-          console.log('Full page HTML after clicking Output:', html);
-      
-          // Also dump just the response area HTML
-          const responseHtml = await response.evaluate(el => el.outerHTML);
-          console.log('Response area HTML:', responseHtml);
       
           // Take screenshot before trying to find content
           await mainWindow.screenshot({ path: `test-results/${provider.name.toLowerCase()}-quote-response-debug.png` });


ui/desktop/src/components/settings_v2/models/ModelsSection.tsx
```diff
@@ -43,19 +43,9 @@ export default function ModelsSection({ setView }: ModelsSectionProps) {
   }, [read, getProviders]);
 
   useEffect(() => {
-    // Initial load
     loadModelData();
-
-    // Set up polling interval to check for changes
```
zanesq, line 49: This was in Alex's [branch](https://github.com/block/goose/commit/4f3d7b7613966c50f17cadb2dbb328d2609a1e51) so I think it was intentional to remove? It seems to work fine without it in my testing.

ui/desktop/tests/e2e/app.spec.ts
```diff
@@ -525,12 +525,6 @@ test.describe('Goose App', () => {
       
           // Wait a bit and dump HTML to see structure
           await mainWindow.waitForTimeout(1000);
-          const html = await mainWindow.evaluate(() => document.documentElement.outerHTML);
```
zanesq, line 528: Just some debugging I removed

ui/desktop/src/components/settings_v2/models/bottom_bar/ModelsBottomBar.tsx
```diff
@@ -27,7 +27,7 @@ export default function ModelsBottomBar({ dropdownRef, setView }: ModelsBottomBa
       setProvider(modelProvider.provider);
       setModel(modelProvider.model);
     })();
-  }, [read, getProviders]);
+  });
```
zanesq, line 30: This fixes the model name not refreshing after switching providers, basically it will re-run this with every render

ui/desktop/src/components/settings_v2/models/ModelsSection.tsx
lilydelalande: not sure -- ik i added that but can't remember why. let's remove it and see what happens :)

lily-de (approved):

PR #2132, opened 2025-04-10
tool count: fix colors and opacity

lily-de:
Popover was translucent and the header/footer text color didnt change for darkmode so it disappeared into the background

### DIFF
--- a/ui/desktop/src/components/ToolCount.tsx
+++ b/ui/desktop/src/components/ToolCount.tsx
@@ -4,7 +4,7 @@ import { ExclamationTriangleIcon } from '@radix-ui/react-icons';
 import { HammerIcon } from 'lucide-react';
 import { Popover, PopoverContent, PopoverTrigger } from './ui/popover';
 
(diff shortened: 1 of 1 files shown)


ui/desktop/src/components/ToolCount.tsx
```diff
@@ -4,7 +4,7 @@ import { ExclamationTriangleIcon } from '@radix-ui/react-icons';
 import { HammerIcon } from 'lucide-react';
 import { Popover, PopoverContent, PopoverTrigger } from './ui/popover';
 
-const SUGGESTED_MAX_TOOLS = 15;
+const SUGGESTED_MAX_TOOLS = 10;
```
kalvinnchau: `google_drive` itself has `13` tools lol so anyone who enables it will get this warning

lilydelalande: oh i meant for it to be higher but lowered it in testing :D will move up to 24

PR #2138, opened 2025-04-10
fix: fix the mismatched tool result/call when using enable/search extension tools

yingjiehe-xyz:
In agent.ts, we only care about the which tools are approved, which requires confirmation, which are denied, so move goose mode check into permission judge

Test:

```
Goose is running! Enter your instructions, or try asking what goose can do.

( O)> list files in my current dir
I apologize, but I notice that I don't currently have any extensions enabled that would allow me to interact with the file system. Let me search for an available extension that might help with this.
─── search_available_extensions | platform ──────────────────────────



I see there's a "computercontroller" extension that might help us with file system operations. Let me enable that.
─── enable_extension | platform ──────────────────────────
extension_name: computercontroller


◇  Goose would like to install the following extension, do you approve?
│  Yes, for this session 
│

Now I can help you list the files in your current directory using a shell script:
─── automation_script | computercontroller ──────────────────────────
language: shell
script: ls -la


◇  Goose would like to call the above tool, do you allow?
│  Allow 
│

I apologize for the permission error. Let me try a different approach:
─── automation_script | computercontroller ──────────────────────────
language: shell
script: pwd && ls


◇  Goose would like to call the above tool, do you allow?
│  Allow 
│

I apologize, but it seems I'm encountering permission issues with executing shell scripts. This might be due to security restrictions. Let me try one more approach using AppleScript, which might have the necessary permissions:
─── computer_control | computercontroller ──────────────────────────
script: ...


◇  Goose would like to call the above tool, do you allow?
│  Allow 
│

Here are the files in your current directory:

- .DS_Store
- 20250409_212539.jsonl
- 20250409_215252.jsonl
- 20250407_164951.jsonl
- 20250407_164758.jsonl
- 20250408_115529.jsonl
- 20250410_111240.jsonl
- 20250409_213756.jsonl
- 20250404_143058.jsonl
- 20250408_095645.jsonl
- 20250404_143227.jsonl
- 20250409_214058.jsonl

It looks like your directory contains several .jsonl files with timestamps in their names, along with a .DS_Store file (which is a macOS system file that stores custom attributes of its containing folder).
```

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -463,30 +463,41 @@ impl Agent {
                             }
                         }
 
-                        // Split tool requests into enable_extension and others
-                        let (enable_extension_requests, non_enable_extension_requests): (Vec<&ToolRequest>, Vec<&ToolRequest>) = remaining_requests.clone()
-                            .into_iter()
-                            .partition(|req| {
-                                req.tool_call.as_ref()
-                                    .map(|call| call.name == PLATFORM_ENABLE_EXTENSION_TOOL_NAME)
-                                    .unwrap_or(false)
-                            });
-
                         // Clone goose_mode once before the match to avoid move issues
                         let mode = goose_mode.clone();
-
-                        // If there are install extension requests, always require confirmation
-                        // or if goose_mode is approve or smart_approve, check permissions for all tools
-                        if !enable_extension_requests.is_empty() || mode.as_str() != "chat" {
+                        if mode.as_str() == "chat" {
+                            // Skip all tool calls in chat mode
+                            for request in remaining_requests {
+                                message_tool_response = message_tool_response.with_tool_response(
+                                    request.id.clone(),
+                                    Ok(vec![Content::text(
+                                        "Let the user know the tool call was skipped in Goose chat mode. \
+                                        DO NOT apologize for skipping the tool call. DO NOT say sorry. \
+                                        Provide an explanation of what the tool call would do, structured as a \
+                                        plan for the user. Again, DO NOT apologize. \
+                                        **Example Plan:**\n \
+                                        1. **Identify Task Scope** - Determine the purpose and expected outcome.\n \
+                                        2. **Outline Steps** - Break down the steps.\n \
+                                        If needed, adjust the explanation based on user preferences or questions."
+                                    )]),
+                                );
+                            }
+                        } else {
+                            // Split tool requests into enable_extension and others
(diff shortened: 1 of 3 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -617,31 +602,12 @@ impl Agent {
...
-                            for request in non_enable_extension_requests.iter().chain(search_extension_requests.iter()) {
-                                if processed_ids.insert(request.id.clone()) {
-                                    if let Ok(tool_call) = request.tool_call.clone() {
-                                        let is_frontend_tool = self.is_frontend_tool(&tool_call.name);
-                                        let tool_future = Self::create_tool_future(&extension_manager, tool_call, is_frontend_tool, request.id.clone());
-                                        tool_futures.push(tool_future);
-                                    }
+                                if extension_manager.supports_resources() {
+                                    tools.push(platform_tools::read_resource_tool());
+                                    tools.push(platform_tools::list_resources_tool());
```
yingjiehe-xyz, line 624: fix the issue that, tools used in previous message but missing in following message

ahau-square: I don't think you need to do this since the platform tools are added at the top of the reply loop already (lines 330-335) ?

crates/goose/src/agents/agent.rs
```diff
@@ -617,31 +602,12 @@ impl Agent {
...
-
-                            // Wait for all tool calls to complete
-                            let results = futures::future::join_all(tool_futures).await;
-                            for (request_id, output) in results {
-                                message_tool_response = message_tool_response.with_tool_response(
-                                    request_id,
-                                    output,
-                                );
+                                tools.push(platform_tools::search_available_extensions_tool());
+                                tools.push(platform_tools::enable_extension_tool());
```
ahau-square, line 627: same as above comment, don't think this is necessary

yingjiehe-xyz: if don't have these tools, we have error: ```Requests which include `tool_use` or `tool_result` blocks must define tools.```, do I miss something?

crates/goose/src/agents/agent.rs
yingjiehe-xyz: but the tool is reset when we have new installation: `tools = extension_manager.get_prefixed_tools().await?;` line 604

ahau-square: ahh you're right

ahau-square (approved):

PR #2142, opened 2025-04-10
settingsv2: fix process for deeplinks with env vars

lily-de:
We had deeplink support for extensions with _no_ env vars, but if we do have env vars, we need a way for users to fill them out. Video below: 

https://github.com/user-attachments/assets/c81e0e07-72ea-413d-ae8e-455d42e4c8a8


kalvinnchau (approved):
there's a bunch of extra logging - did you wanna keep those in?

nit: the dark mode confirmation is a different theme/color than the rest of the modals:
image
image

lilydelalande:
my toxic trait is forgetting to clean up my log statements

PR #2144, opened 2025-04-10
[WIP][DNM] publish crate

zhong-sq:


crates/goose-cli/src/session/mod.rs
```diff
@@ -279,19 +279,29 @@ impl Session {
...
         // instead of being tied to each individual session's messages
-        let history_file = choose_app_strategy(crate::APP_STRATEGY.clone())
-            .expect("goose requires a home dir")
-            .in_config_dir("history.txt");
-
-        // Ensure config directory exists
-        if let Some(parent) = history_file.parent() {
-            if !parent.exists() {
-                std::fs::create_dir_all(parent)?;
+        // On Linux and macOS, use /tmp/.config/goose/ instead of ~/.config/goose/
```
Tyler-Hardin, line 285: Maybe lets do GOOSE_HOME env var instead of hard coding to ~ or /tmp?

Or is this just temporary while you work on it?

Tyler-Hardin: Just noticed you work at Block. You guys might also wanna pull all the env var accesses into a config module. You can write global functions there that read the env vars for you, or fill in default, or maybe in the future read from a conf file, etc. You can also use #[cfg(test)/not(test)] to shim in test values.

PR #2146, opened 2025-04-11
Chat bottom menu bar token and tools alerts

zanesq:
Chat bottom bar token and tools alerts.

- Two types of alerts, warning and error. Switches to error icon if any errors
- Auto pops up on first load if any alerts, only auto pops up if new ones. Auto hides after 3 seconds.
- Hovering on the icon or the popover keeps it visible. Hovering out hides in 1 second.
- Ability to add secondary action info like a button to open extension pane
- Currently fetches tool count from existing api and calculates message tokens used.
- Token limits are dynamic and supplied by the backend based on the provider. Falls back to 128000 For custom models that aren't supplied by the backend.
- Token warning approaching limit shows when its 80% or more of available tokens.

Screenshot 2025-04-10 at 5 56 27 PM

Screenshot 2025-04-10 at 5 57 16 PM

Screenshot 2025-04-10 at 5 57 32 PM

~~After talking with Spencer, added a similar alert to extensions page so it's clear what "tools" are and how they are related to extensions as they are enabled and disabled. Refactored and used logic from popover alerts to avoid duplication.~~

^ UPDATE: Removed the extensions view integration in https://github.com/block/goose/pull/2199

### DIFF
--- a/ui/desktop/src/components/settings_v2/extensions/agent-api.ts
+++ b/ui/desktop/src/components/settings_v2/extensions/agent-api.ts
@@ -2,6 +2,8 @@ import { ExtensionConfig } from '../../../api/types.gen';
 import { getApiUrl, getSecretKey } from '../../../config';
 import { toastService, ToastServiceOptions } from '../../../toasts';
 import { replaceWithShims } from './utils';
+import { saveEnvVarsToKeyring } from './extension-manager';
+import type { AgentExtensionConfig } from './extension-manager';
 
 interface ApiResponse {
   error?: boolean;
@@ -141,13 +143,17 @@ export async function addToAgent(
   options: ToastServiceOptions = {}
 ): Promise<Response> {
   try {
-    if (extension.type === 'stdio') {
-      extension.cmd = await replaceWithShims(extension.cmd);
+    await saveEnvVarsToKeyring(extension);
+
+    const ext = toAgentExtensionConfig(extension);
+
+    if (ext.type === 'stdio') {
+      ext.cmd = await replaceWithShims(ext.cmd);
     }
 
-    extension.name = sanitizeName(extension.name);
+    ext.name = sanitizeName(ext.name);
 
-    return await extensionApiCall('/extensions/add', extension, options);
+    return await extensionApiCall('/extensions/add', ext, options);
   } catch (error) {
     // Check if this is a 428 error and make the message more descriptive
     if (error.message && error.message.includes('428')) {
@@ -178,4 +184,33 @@ export async function removeFromAgent(
 
 function sanitizeName(name: string) {
   return name.toLowerCase().replace(/-/g, '').replace(/_/g, '').replace(/\s/g, '');
+}
+
+export function toAgentExtensionConfig(config: ExtensionConfig): AgentExtensionConfig {
+  // Use type narrowing to handle different variants of the union type
+  if ('type' in config) {
+    switch (config.type) {
+      case 'sse': {
+        const { envs, ...rest } = config;
+        return {
+          ...rest,
+          env_keys: envs ? Object.keys(envs) : undefined,
+        };
+      }
+      case 'stdio': {
+        const { envs, ...rest } = config;
+        return {
+          ...rest,
+          env_keys: envs ? Object.keys(envs) : undefined,
+        };
+      }
+      case 'builtin':
+      case 'frontend':
+        // These types don't have envs field, so just return as is
+        return config;
+    }
+  }
+
+  // This should never happen due to the union type constraint
+  throw new Error('Invalid extension configuration type');
 }
--- a/ui/desktop/src/components/settings_v2/extensions/extension-manager.ts
+++ b/ui/desktop/src/components/settings_v2/extensions/extension-manager.ts
@@ -1,6 +1,12 @@
 import type { ExtensionConfig } from '../../../api/types.gen';
 import { toastService, ToastServiceOptions } from '../../../toasts';
 import { addToAgent, removeFromAgent } from './agent-api';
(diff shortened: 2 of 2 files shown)


ui/desktop/src/components/ChatView.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useEffect, useRef, useState, useMemo } from 'react';
 import { getApiUrl } from '../config';
-import BottomMenu from './BottomMenu';
+import BottomMenu from './bottom_menu/BottomMenu';
```
zanesq, line 3: Moved all bottom menu and related components to a bottom_menu directory for better organization

ui/desktop/src/components/bottom_menu/BottomMenu.tsx
```diff
@@ -1,24 +1,110 @@
...
+import { bottomMenuPopoverEnabled, settingsV2Enabled } from '../../flags';
 import { BottomMenuModeSelection } from './BottomMenuModeSelection';
-import ModelsBottomBar from './settings_v2/models/bottom_bar/ModelsBottomBar';
-import ToolCount from './ToolCount';
+import ModelsBottomBar from '../settings_v2/models/bottom_bar/ModelsBottomBar';
+import ToolCount from '../ToolCount';
+import { getTools } from '../../api';
+
+// Constants
+// todo: map these to a config based on the LLM provider?
```
zanesq: Where should we get these from? Is there a list of all possible LLMs available and their token limits? @lily-de started compiling this but hoping there is a better way available.

nahiyankhan: I wonder how often they have to be kept updated?

ui/desktop/src/components/bottom_menu/BottomMenuAlertPopover.tsx
```diff
@@ -0,0 +1,183 @@
...
+//todo: for some reason clearTimeout is not available globally but setTimeout is on other files?
+const { clearTimeout } = window;
+
+const alertIcons: Record<AlertType, React.ReactNode> = {
+  [AlertType.Error]: <IoIosCloseCircle className="h-5 w-5" />,
+  [AlertType.Warning]: <IoIosWarning className="h-5 w-5" />,
+};
+
+const alertStyles: Record<AlertType, string> = {
+  [AlertType.Error]: 'bg-[#d7040e] text-white',
```
nahiyankhan: can be part of a follow up. setting up the colors as CSS vars and in tailwind.config.ts. I'll check what they should be called

ui/desktop/src/components/ChatView.tsx
```diff
@@ -380,6 +380,34 @@ export default function ChatView({
       .reverse();
   }, [filteredMessages]);
 
+  // Calculate total tokens from messages
+  const totalTokens = useMemo(() => {
+    // todo: do we need to calculate tokens on the front end or is token count available already?
```
zanesq: I didn't see token count in the message data for chatview. any ideas @salman1993 ?

ui/desktop/src/components/bottom_menu/BottomMenu.tsx
zanesq: @salman1993 any ideas?

salman1993: yeah this one is gonna hard to keep up with imo. we can expose the model's context limit that we have in the backend. however, even that one is not always updated

https://github.com/block/goose/blob/a9132291779a4265e7740e30077a0c3543f4da3c/crates/goose/src/model.rs#L67-L86

ui/desktop/src/components/ChatView.tsx
salman1993: yeah looks like we'll have to add a new endpoint that just exposes the session metadata here: https://github.com/block/goose/blob/26a8bcafbee6feae2b995d741b1cb9de0485162d/crates/goose-server/src/routes/session.rs#L85

then we can retrieve the session metadata (which includes total tokens) using `session_id: chat.id`? 

```rust
// New response type for metadata only
#[derive(Serialize)]
struct SessionMetadataResponse {
    session_id: String,
    metadata: session::SessionMetadata,
}

// New endpoint: Get only the session's metadata
async fn get_session_metadata(
    State(state): State<AppState>,
    headers: HeaderMap,
    Path(session_id): Path<String>,
) -> Result<Json<SessionMetadataResponse>, StatusCode> {
    // Verify secret key
    let secret_key = headers
        .get("X-Secret-Key")
        .and_then(|value| value.to_str().ok())
        .ok_or(StatusCode::UNAUTHORIZED)?;
    if secret_key != state.secret_key {
    }

    let session_path = session::get_path(session::Identifier::Name(session_id.clone()));

    // Read metadata only
    let metadata = session::read_metadata(&session_path).map_err(|_| StatusCode::NOT_FOUND)?;

    Ok(Json(SessionMetadataResponse { session_id, metadata }))
}

// Configure routes for this module
pub fn routes(state: AppState) -> Router {
    Router::new()
        ...
        // New route for session metadata only
        .route("/sessions/:session_id/metadata", get(get_session_metadata))
        .with_state(state)
}
```

lily-de (approved):

ui/desktop/src/components/bottom_menu/BottomMenu.tsx
salman1993: ^ this would be similar to the one above, we'd have to create a route and expose that. probably also need to make `fn get_model_specific_limit` public (which is okay) -> note that this logic is completely separately from the agent (its just a utility that sets context limit based on model name & not a full map)

lily-de (approved):

PR #2126, opened 2025-04-10
refactor: agent's reply method

salman1993:
logic for yielding:

| Mode (cols), Extensions (rows)   | Auto                                                             | Manual Approve             | Smart Approve              | Chat     |
|----------------------------------|------------------------------------------------------------------|----------------------------|----------------------------|----------|
| Is frontend tool request         | Always yield control to user to submit tool result               | yield control to user      | yield control to user      | No yield |
| Is extension enable tool request | Always ask for permission to enable MCP extension (always yield) | ask for permission (yield) | ask for permission (yield) | No yield |
| Is not extension tool request    | No yield                                                         | yield                      | Yield (maybe)              | No yield |

The order of operation for yielding: (1) frontend tool requests are yielded first, (2) then extension enable requests and (3) then non-extension (standard) tool requests.

PR #2153, opened 2025-04-11
refactor: small piece of agent.reply()

salman1993:
tried a bigger refactor but that failed before: https://github.com/block/goose/pull/2126 

Challenge:
- You cannot capture or call yield within a normal closure (or an async closure) because yield is tied to the generator’s internal control flow. The code block inside async_stream::try_stream! { ... } is a generator; yield must be invoked directly there. This limits our ability to move a lot of mode for goose_mode = "approval" and permissions 

Improvements:
- move to smaller focused functions on preparing tools and prompt, generating provider response
- not holding onto the extension manager lock for the entire reply() method

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -320,14 +320,13 @@ impl Agent {
         let config = Config::global();
 
         // Setup tools and prompt
-        let (mut tools, toolshim_tools, mut system_prompt) = self
-            .prepare_tools_and_prompt()
-            .await?;
+        let (mut tools, mut toolshim_tools, mut system_prompt) =
+            self.prepare_tools_and_prompt().await?;
 
         let goose_mode = config.get_param("GOOSE_MODE").unwrap_or("auto".to_string());
 
-
-        let (tools_with_readonly_annotation, tools_without_annotation) = Self::categorize_tools_by_annotation(&tools);
+        let (tools_with_readonly_annotation, tools_without_annotation) =
+            Self::categorize_tools_by_annotation(&tools);
 
         // Set the user_message field in the span instead of creating a new event
         if let Some(content) = messages
@@ -340,7 +339,7 @@ impl Agent {
 
         Ok(Box::pin(async_stream::try_stream! {
             let _reply_guard = reply_span.enter();
-            loop {                
+            loop {
                 match Self::generate_response_from_provider(
                     self.provider(),
                     &system_prompt,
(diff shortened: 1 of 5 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -317,54 +314,19 @@ impl Agent {
...
-            system_prompt = modify_system_prompt_for_tool_json(&system_prompt, &tools);
-            // make a copy of tools before empty
-            toolshim_tools = tools.clone();
-            // pass empty tools vector to provider completion since toolshim will handle tool calls instead
-            tools = vec![];
-        }
+        let (tools_with_readonly_annotation, tools_without_annotation) =
+            Self::categorize_tools_by_annotation(&tools);
 
         // Set the user_message field in the span instead of creating a new event
```
ahau-square: nit: old comment can be deleted?

yingjiehe-xyz (approved):

crates/goose/src/agents/agent.rs
```diff
@@ -317,54 +314,19 @@ impl Agent {
...
-        // we add in the 2 resource tools if any extensions support resources
-        // TODO: make sure there is no collision with another extension's tool name
-        if extension_manager.supports_resources() {
-            tools.push(platform_tools::read_resource_tool());
-            tools.push(platform_tools::list_resources_tool());
-        }
-        tools.push(platform_tools::search_available_extensions_tool());
-        tools.push(platform_tools::enable_extension_tool());
+        // Setup tools and prompt
+        let (mut tools, mut toolshim_tools, mut system_prompt) =
```
ahau-square, line 323: why not just return tools (which either has tools or is empty depending on whether toolshim is on) instead of both tools and toolshim_tools?

crates/goose/src/agents/agent.rs
```diff
@@ -130,7 +126,7 @@ impl Agent {
                 .await
         } else if tool_call.name == PLATFORM_SEARCH_AVAILABLE_EXTENSIONS_TOOL_NAME {
             extension_manager.search_available_extensions().await
-        } else if is_frontend_tool {
+        } else if self.is_frontend_tool(&tool_call.name) {
             // For frontend tools, return an error indicating we need frontend execution
             Err(ToolError::ExecutionError(
                 "Frontend tool execution required".to_string(),
```
ahau-square, line 127: why would we want to return an error here? will that cause weird behavior in the agent loop?

salman1993: I think we shouldn't reach this part of the code cause frontend tool requests are handled first, which should be before regular tool requests so my guess is we're trying to be defensive here and bubble up errors in case sth unexpected happened

PR #2160, opened 2025-04-11
docs: New shared Agent topic with updates to existing topics and other stuff

JohnMAustin78:
Added shared agent topic, updated cli command topic, added image assets, corrected two broken links, and set TOC order for guide topics

I am adding a bit more color about how the shared agent actually collaborates. I will commit that update today.


documentation/docs/guides/share-goose-sessions.md
```diff
@@ -0,0 +1,109 @@
...
+
+        The new Goose session shows a set of actions that you can take as clickable buttons. As shown in the following image, the shared Agent is shown as an action that summarizes the purpose of the shared Agent. To activate the agent, review the summary and then select the action.  
+
+        ![Start a shared Agent](../assets/guides/start-shared-agent-ui.png)
+
+
+    </TabItem>
+</Tabs>
+
+
```
JohnMAustin78, line 230: Need to add content here for information about how collaboration works

documentation/docs/assets/guides/shared-agent-created-ui.png
kalvinnchau, line 1: this is stale after #2119

documentation/docs/assets/guides/start-shared-agent-ui.png
kalvinnchau, line 1: this is stale after #2119, its a subtle change where above in the top-left corner there is a display that shows you're using the shared agent

documentation/docs/guides/goose-cli-commands.md
```diff
@@ -208,16 +209,24 @@ Execute commands from an instruction file or stdin. Check out the [full guide](/
...
 **Usage:**
 
 ```bash
 goose run --instructions plan.md
+
+#Load a custom recipe from this chat and drop into an in interactive session
+goose run --recipe recipe.yaml -s
+
+#Load a custom recipe from this chat and drop into an interactive session with that starts with a prompt from Goose
+goose run --recipe recipe.yaml -interactive
```
kalvinnchau:
```suggestion
goose run --recipe recipe.yaml --interactive
```

kalvinnchau: so in this case: `goose run --recipe recipe.yaml -s` `==` `goose run --recipe recipe.yaml --interactive` 

we might want to make a distinction that "with a prompt from Goose" means that there's a `prompt` field in the `yaml` file

if there is a `prompt` field we can run the following command, and it'll run as a "one shot", i.e. it'll exit after it does what it's been prompted to do (good for CI, scripts, etc)
```bash
goose run --recipe recipe.yaml
```

if we run the same recipe with:
```bash
goose run --recipe recipe.yaml --interactive (or -s)
```
that'll run the initial prompt _and then_ drop the user into an interactive session so they can follow up

---
and finally if you try to run a non-interactive mode with no `prompt` you get this error:
```bash
goose run --recipe recipe_no_prompt.yaml
...
Error: no text provided for prompt in headless mode
```

documentation/docs/guides/goose-cli-commands.md
```diff
@@ -242,6 +251,30 @@ Used to evaluate system-configuration across a range of practical tasks. See the
...
+- **`--help, -h`**: Print this message or the help for the subcommand
+
+**Command Usage:**
+
+```bash
+# Validate a recipe file
+goose recipe validate
+
+# Generate a deeplink for a recipe file
+goose recipe deeplink
```
kalvinnchau: should we add that there's an expected file param?
```suggestion
goose recipe validate $FILE.yaml

# Generate a deeplink for a recipe file
goose recipe deeplink $FILE.yaml
```

documentation/docs/guides/share-goose-sessions.md
```diff
@@ -0,0 +1,109 @@
...
+## What gets shared?
+You may have started a project in a Goose session that needed information that your collaborating team member needs in the shared agent that you create. At the same time, you may have needed to give Goose information that you'd rather not share. 
+### Shared components
+The shared agent includes these components:
+* Conversation history (all messages)
+* Tool outputs and results
+* Files or content created during the session
+* Active extensions and their configurations
+* Project context (when working within a project)
```
kalvinnchau, line 240: soo, confusing but this is not the case, there are shared sessions, which does have session history

but in this case, what it does is create a new set of instructions _based_ on your current conversation to generally a specific agent that is primed to do a specific task (for example, a code-review agent that has some addtl context about how you prefer to style/use libraries etc)

it *does* map in active extensions and you can include content created but you must do so manually as text (for now) into the instructions field

documentation/docs/guides/share-goose-sessions.md
```diff
@@ -0,0 +1,109 @@
+---
+sidebar_position: 5
+title: Making Shared Goose Agents
+sidebar_label: Share Sessions
+---
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+
+A shared session in Goose is like a collaborative workspace where multiple people can work with the AI assistant together in real-time. Think of it similar to a shared Google Doc, but for AI assistance.
```
angiejones: love this

documentation/docs/guides/share-goose-sessions.md
```diff
@@ -0,0 +1,109 @@
...
+
+        In your terminal with a Goose session running, input the following:
+        ```sh
+        ( O)> /recipe 
+        ```
+        Goose generates `recipe.yaml` and saves it for you as shown in the following output:
+
+        ```sh
+        Generating Recipe
+        Saved recipe to .../block/goose/recipe.yaml
```
angiejones: will `block` be in the url? or is this unique to you?

JohnMAustin78: Well, that's a good question!  Also, is the sharing peer-to-peer?  Or does an implementing organization need a server?

documentation/docs/guides/share-goose-sessions.md
JohnMAustin78: Noted and will update content

documentation/docs/guides/share-goose-sessions.md
JohnMAustin78: So, "Block" is only for me.  The Goose source takes whatever path I provide. In my case, "block" is in my path.  I'll update the docs

                          match self.save_recipe(&recipe, filepath_str) {
                                Ok(path) => println!(
                                    "{}",
                                    console::style(format!("Saved recipe to {}", path.display()))
                                        .green()
                                ),
                                Err(e) => {
                                    println!("{}", console::style(e).red());
                                }
                            }

kalvinnchau: >  Also, is the sharing peer-to-peer?

in its current form, p2p

and the `block` section, it prints out the absolute path to the file, based on your `cwd`

angiejones (approved):

documentation/docs/guides/share-goose-sessions.md
```diff
@@ -0,0 +1,225 @@
...
+        When another Goose CLI user sends you a recipe file, save it in the directory where you want to start your shared Agent.
+        From your terminal, navigate to the directory where you saved the file, and run:
+        ```sh
+        goose run --<RECIPE_FILE_NAME> 
+        ```
+        The Goose CLI loads the recipe and creates a session with all of the components shared in the original session. For  infomration about the Goose `run` command, see the [run](/docs/guides/goose-cli-commands#run-options) CLI command.
+
+    </TabItem>
+    <TabItem value="ui" label="Goose Desktop">
+        When another Goose desktop user wants to share an Agent, they send you an URL which you use to start the new shared Agent.
```
EbonyLouis, line 164:
```suggestion
        When another Goose desktop user wants to share an Agent, they send you a URL which you use to start the new shared Agent.
```

EbonyLouis (dismissed):

documentation/docs/guides/share-goose-sessions.md
```diff
@@ -0,0 +1,225 @@
...
+
+        Open a new tab on your browser and paste the shared Agent URL into the address bar and press the **enter** key on your keyboard. The browser requests your permission to start a new Goose session with the shared components. 
+
+        The new Goose session shows a set of actions that you can take as clickable buttons. As shown in the following image, the shared Agent is shown as an action that summarizes the purpose of the shared Agent. To activate the agent, review the summary and then select the action.
```
EbonyLouis: it says "As shown in the following image" but there's no image

documentation/docs/guides/share-goose-sessions.md
```diff
@@ -0,0 +1,225 @@
...
+## Accessing conversation history in shared sessions
+When you join a shared session, you automatically get access to:
+* The full conversation history (chat messages) from when the session started
+* All tool outputs and results
+* Any files or content created during the session
+
+### How It Works
+When you click the shared link, you'll join the active session. The conversation history will automatically load in your Goose window. You'll see all messages and interactions in real-time and can scroll up to view earlier parts of the conversation.
+### Important Notes
+The history is synchronized live - you'll see new messages as they happen. You don't need to take any special steps to access the history. The conversation remains available as long as the session is active. Once the host ends the session, the shared access ends.
```
EbonyLouis, line 183: wow love this feature!!!

documentation/docs/guides/share-goose-sessions.md
```diff
@@ -0,0 +1,225 @@
...
+   1. Share file contents directly in the conversation where possible
+   1. Consider using version control systems for code files
+
+* **Document Important Files**:
+   1. Keep track of important files created during the session
+   1. Save or copy relevant content before the session ends
+
+
+## What gets shared?
+You may have started a project in a Goose session that needed information that your collaborating team member needs in the shared agent that you create. At the same time, you may have needed to give Goose information that you'd rather not share.
```
EbonyLouis:
```suggestion
You might start a project in a Goose session and realize your teammate needs access to that context through a shared agent. But at the same time, you may have shared things with Goose that you’d rather keep private.
```

EbonyLouis: when I read the original it felt confusing, but this is just a suggestion

EbonyLouis:
I started by reading the Goose Desktop instructions, then switched over to the CLI instructions. That’s when I saw the word “recipe” being used. I know it’s referring to the same thing, but it feels a bit different since that term isn’t mentioned in the Goose Desktop flow.

This might just be me not fully understanding everything yet, but for consistency, could we also use the word “recipe” in Goose Desktop? So instead of just saying “sharing session,” we could say something like: “also known as sharing a recipe, where you choose what to include or exclude.” I think that might help connect the dots more easily across both interfaces.

documentation/docs/guides/share-goose-sessions.md
JohnMAustin78: Yes, it is confusing.  The Goose devs are telling me that _they_ are confused too.   A recipie == the package of context and activities that will make up a shared agent.  I will clarify that.

documentation/docs/guides/share-goose-sessions.md
JohnMAustin78: whoops.  Yeah, I removed the image and missed removing the text

documentation/docs/guides/share-goose-sessions.md
JohnMAustin78: Actually "an URL" is correct  :-)

EbonyLouis: Its not lol 

it could perhaps be preference but URL is pronounced like "you" yes its technically a `u` so it makes you think its a vowel but its not pronounced as such to require `an`. The word is pronounced like it starts with a constant so its `a` not `an`.

documentation/docs/guides/share-goose-sessions.md
```diff
@@ -0,0 +1,235 @@
+---
+sidebar_position: 5
+title: Making Shared Goose Agents
+sidebar_label: Share Sessions
```
acekyd: Tried out a demo with this and I think there might be a mixup with sharing sessions and sharing agents across the article. Starting with the title, might be best to leave the label tuned towards shareable agents vs sessions - since session data is also not shared when an agent is shared.

documentation/docs/guides/share-goose-sessions.md
```diff
@@ -0,0 +1,235 @@
...
+    </TabItem>
+</Tabs>
+
+## Accessing conversation history in shared sessions
+When you join a shared session, you automatically get access to:
+* The full conversation history (chat messages) from when the session started
+* All tool outputs and results
+* Any files or content created during the session
+
+### How It Works
```
acekyd, line 180: When you join a shared agent link, it actually starts an empty session but with the prompts / activity from when it was created added as pills in the homescreen so that the new users can get started quickly. So the conversation history, results, files / content and messages would not be seen. 

Right now, some of these paragraphs combine both.

acekyd:
Nice! Left some comments. Happy to pair to figure out more of the implementation details with tests.

documentation/docs/guides/share-goose-sessions.md
JohnMAustin78: Good catch.  It seems that terminology is still in flux ... or maybe I need to read a Goose glossary somewhere. I'll get this fixed up

documentation/docs/guides/share-goose-sessions.md
```diff
@@ -1,16 +1,37 @@
...
-sidebar_label: Share Sessions
+title: Sharing a Goose Agent
+sidebar_label: Share Goose Agents
 ---
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
 
-A shared session in Goose is like a collaborative workspace where multiple people can work with the AI assistant together in real-time. Think of it similar to a shared Google Doc, but for AI assistance.
+A shared Agent in Goose is like a collaborative workspace where multiple people can work with the AI assistant together in real-time. Think of it similar to a shared Google Doc, but for AI assistance.
```
acekyd, line 10: I'd say switching more to a `template` as an example pending clarification on the shared history.

documentation/docs/guides/share-goose-sessions.md
```diff
@@ -1,16 +1,37 @@
...
 
 ## Create a shared Agent
 When you create an Agent to be shared, you are creating a host Agent. When your host terminates, the shared Agents are disconnected and collaborators can no longer see activity in the host.
+
+### Agent setup instructions
+When you share your Goose Agent setup (either through the command line or desktop app), you'll get a chance to review and edit the setup instructions. These instructions come from your current session, but they might need some extra details to work well for others.
+
+Here's an example: Let's say you were working with Goose to make technical writing easier to understand. In this case, you should add a note like this to your instructions:
```
acekyd: Might need to find a better example for this that highlights the reusability more. Thinking through - maybe processing logs or some data.

blackgirlbytes:
hiii @JohnMAustin78. What are your thoughts on the comment below. I have a similar question as @EbonyLouis . Maybe we also need to communicate with the dev team to have a shared term for "recipes" in the CLI and Goose Desktop. Any thoughts on this?

> I started by reading the Goose Desktop instructions, then switched over to the CLI instructions. That’s when I saw the word “recipe” being used. I know it’s referring to the same thing, but it feels a bit different since that term isn’t mentioned in the Goose Desktop flow.
> 
> This might just be me not fully understanding everything yet, but for consistency, could we also use the word “recipe” in Goose Desktop? So instead of just saying “sharing session,” we could say something like: “also known as sharing a recipe, where you choose what to include or exclude.” I think that might help connect the dots more easily across both interfaces.

acekyd (approved):

PR #2165, opened 2025-04-11
feat: avoid duplicate confirmation handle code

yingjiehe-xyz:
instead of having different logic for enable extension and tool call, we can combine them into one.

Note: enable extension tool will always run before normal tools

### DIFF
--- a/crates/goose/src/permission/permission_judge.rs
+++ b/crates/goose/src/permission/permission_judge.rs
@@ -1,3 +1,4 @@
+use crate::agents::platform_tools::PLATFORM_ENABLE_EXTENSION_TOOL_NAME;
 use crate::config::permission::PermissionLevel;
 use crate::config::PermissionManager;
 use crate::message::{Message, MessageContent, ToolRequest};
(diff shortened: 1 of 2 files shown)


crates/goose/src/permission/permission_judge.rs
```diff
@@ -158,7 +189,7 @@ pub struct PermissionCheckResult {
 }
 
 pub async fn check_tool_permissions(
-    candidate_requests: Vec<&ToolRequest>,
+    candidate_requests: Vec<&&ToolRequest>,
```
salman1993: do we need the double &&?

yingjiehe-xyz: yes, the passed-in remaining_requests is Vec<&&ToolRequest>

crates/goose/src/permission/permission_judge.rs
```diff
@@ -150,6 +152,35 @@ pub async fn detect_read_only_tools(
     }
 }
 
+/// Gets the boolean value whether the message is enable extension related and
+/// the cconfirmation message based on the tool call
+pub fn get_confirmation_message(request_id: &str, tool_call: ToolCall) -> (bool, Message) {
+    if tool_call.name == PLATFORM_ENABLE_EXTENSION_TOOL_NAME {
+        (
+            true,
```
salman1993: nit: introduce an enum here for type (EnableExtension, ToolConfirmation)

salman1993 (approved):
looks great! left a minor comment to use enum instead of is_enable_extension_tool

crates/goose/src/permission/permission_judge.rs
yingjiehe-xyz: done

PR #2186, opened 2025-04-14
fix: filter out the enable extension tool in the ui as it should always ask for permission 

yingjiehe-xyz:
Also reuse the list_tool method in the agent reply

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -33,6 +33,8 @@ use mcp_core::{
     prompt::Prompt, protocol::GetPromptResult, tool::Tool, Content, ToolError, ToolResult,
 };
 
+use super::platform_tools;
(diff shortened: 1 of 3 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -272,10 +273,21 @@ impl Agent {
...
+            .unwrap_or_default();
+
+        // Add resource tools if supported
+        if extension_manager.supports_resources() {
+            prefixed_tools.push(platform_tools::read_resource_tool());
+            prefixed_tools.push(platform_tools::list_resources_tool());
+        }
+
+        // Add platform tools
+        prefixed_tools.push(platform_tools::search_available_extensions_tool());
```
salman1993: nit: add these tools before conditional ones above

PR #2187, opened 2025-04-14
Ahau/frontend tools3

ahau-square:
currently works for example calculator tool and enable extensions, tested on pdf extension

todos
- update tools.ts executeEnableExtension to appropriately handle configs for sse, stdio, and builtin extensions when constructing body for request to extensions/add endpoint
- investigate issue with enabling fetch tool 
- ensure that updated system prompt/tools are provided to the agent in the inner reply loop if new extensions are added via frontend tools (see prepare_tools_and_prompt)

testing
- test adding both stdio and builtin extensions
- test adding extensions that have env variables

work for future PRs
- implement frontend tools on CLI
- implement tool permissioning for enable extension frontend/cli tools


crates/goose/src/agents/agent.rs
```diff
@@ -44,9 +44,9 @@ const ESTIMATE_FACTOR_DECAY: f32 = 0.9;
 pub struct Agent {
     pub(super) provider: Arc<dyn Provider>,
     pub(super) extension_manager: Mutex<ExtensionManager>,
-    pub(super) frontend_tools: HashMap<String, FrontendTool>,
-    pub(super) frontend_instructions: Option<String>,
-    pub(super) prompt_manager: PromptManager,
+    pub(super) frontend_tools: Mutex<HashMap<String, FrontendTool>>,
+    pub(super) frontend_instructions: Mutex<Option<String>>,
```
salman1993, line 48: i would refactor this into a `frontend_tool_manager` with setter methods to set the tools and instructions (mainly to restrict the region where we acquire and release the mutex lock)

PR #2188, opened 2025-04-14
fix: format the agent code

yingjiehe-xyz:


crates/goose/src/agents/agent.rs
```diff
@@ -399,97 +399,90 @@ impl Agent {
...
-                                request_id,
-                                output
-                            );
-                        }
 
-                        // Update system prompt and tools if installations were successful
-                        if all_install_successful {
-                            (tools, toolshim_tools, system_prompt) = self.prepare_tools_and_prompt().await?;
+                            // Wait for all tool calls to complete
+                            let results = futures::future::join_all(tool_futures).await;
```
kalvinnchau, line 469: is this entire section removed on purpose?
```rust
                        for (request_id, output) in results {
                            let mut response = message_tool_response.lock().await;
                            *response = response.clone().with_tool_response(
                                request_id,
                                output,
                            );
                        }
```

yingjiehe-xyz: combine them into one loop:

```
for (request_id, output) in results.into_iter().chain(install_results.into_iter()) {
                                let mut response = message_tool_response.lock().await;
                                *response = response.clone().with_tool_response(
                                    request_id,
                                    output
                                );
                            }
```

kalvinnchau: ah missed that in the diff

kalvinnchau (approved):

PR #2071, opened 2025-04-07
ui: show tool number

lily-de:
Chose an arbitrary number (15) to warn users

Open questions:

* tools vs extensions 
* should we only show an icon if there's a 'warning'?

Screenshot 2025-04-07 at 5 54 53 PM

Screenshot 2025-04-07 at 6 02 21 PM

PR #2193, opened 2025-04-14
feat: improve tool loading

yingjiehe-xyz:
Previously we first list all tool and do the filter, it takes longer time than filter by extension name first.

Test:

```
┌   goose-configure 
│
│  Goose Settings 
│
◇  What setting would you like to configure?
│  Tool Permission 
│
◇  Choose an extension to configure tools
│  developer 
│
◇  Choose a tool to update permission
│  developer__list_windows 
│
◇  Set permission level for tool developer__list_windows, current permission level: Always Allow
│  Always Allow 
│
└  Updated permission level for tool developer__list_windows to Always Allow.
```

### DIFF
--- a/crates/goose-cli/src/commands/configure.rs
+++ b/crates/goose-cli/src/commands/configure.rs
@@ -973,6 +973,7 @@ pub async fn configure_tool_permissions_dialog() -> Result<(), Box<dyn Error>> {
             style("Warning").yellow().italic(),
             selected_extension_name
         );
+        return Ok(());
     }
 
     let mut permission_manager = PermissionManager::default();
--- a/crates/goose-cli/src/commands/configure.rs
+++ b/crates/goose-cli/src/commands/configure.rs
@@ -2,6 +2,7 @@ use cliclack::spinner;
 use console::style;
 use goose::agents::extension::ToolInfo;
 use goose::agents::extension_manager::get_parameter_names;
+use goose::agents::platform_tools::PLATFORM_ENABLE_EXTENSION_TOOL_NAME;
 use goose::agents::Agent;
 use goose::agents::{extension::Envs, ExtensionConfig};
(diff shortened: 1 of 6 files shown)


crates/goose-cli/src/commands/configure.rs
```diff
@@ -937,28 +956,32 @@ pub async fn configure_tool_permissions_dialog() -> Result<(), Box<dyn Error>> {
...
 
     let mut permission_manager = PermissionManager::default();
-    // Fetch the list of tools grouped by extension
-    let tools: Vec<ToolInfo> = agent
-        .list_tools()
+    let selected_tools = agent
+        .list_tools(Some(selected_extension_name.clone()))
         .await
         .into_iter()
+        .filter(|tool| tool.name != PLATFORM_ENABLE_EXTENSION_TOOL_NAME)
```
wendytang: The platform tools should never show since they're not part of any extension?

Or if they could be added, would we need to handle the other platform tools?
```
pub const PLATFORM_READ_RESOURCE_TOOL_NAME: &str = "platform__read_resource";
pub const PLATFORM_LIST_RESOURCES_TOOL_NAME: &str = "platform__list_resources";
pub const PLATFORM_SEARCH_AVAILABLE_EXTENSIONS_TOOL_NAME: &str =
    "platform__search_available_extensions";
pub const PLATFORM_ENABLE_EXTENSION_TOOL_NAME: &str = "platform__enable_extension";
```

yingjiehe-xyz: in the first case case, do we ask approval for search extension? Right now, we will ask for approval and that's why I would like to show platform extension

wendytang (approved):
seems ok if the tool count feature is preserved https://github.com/block/goose/pull/2071

PR #2196, opened 2025-04-15
feat: disable copy button unless all fields are set

kalvinnchau:
we are able to use the copy button even when the required fields are *not* set, which produces a bad deeplink, disable it until all fields are set like the "open agent"

the cursor changes to the cursor with circle with a slash through it - can't screenshot my cursor


zakiali (approved):
lgtm! 

for bonus points, would be nice to have an animated click for the clipboard

PR #2197, opened 2025-04-15
feat: show error on broken recipe url's

zakiali:
if there is a problem parsing the recipe config, raise an error on the newly created chat window


kalvinnchau (commented):
I like the error page!
The `Honk` needs font needs to be updated to use dark-mode colors though:
image

kalvinnchau (approved):

PR #2199, opened 2025-04-15
Integrate tool and token count alerts with backend data

zanesq:
This PR wraps up https://github.com/block/goose/pull/2146 with actual token and session data from the backend along with a few fixes. I also removed the extension tools alert for now, since the call for tool count is delayed it didn't work well with the extensions ui toggles.

To test this branch `npm run start-alpha-server`

### DIFF
--- a/ui/desktop/src/components/bottom_menu/BottomMenu.tsx
+++ b/ui/desktop/src/components/bottom_menu/BottomMenu.tsx
@@ -13,8 +13,9 @@ import ModelsBottomBar from '../settings_v2/models/bottom_bar/ModelsBottomBar';
 import { useConfig } from '../ConfigContext';
 import { getCurrentModelAndProvider } from '../settings_v2/models/index';
 
-const TOKEN_WARNING_THRESHOLD = 0.8;
-const SUGGESTED_MAX_TOOLS = 25;
+const TOKEN_LIMIT_DEFAULT = 128000; // fallback for custom models that the backend doesn't know about
+const TOKEN_WARNING_THRESHOLD = 0.8; // warning shows at 80% of the token limit
+const TOOLS_MAX_SUGGESTED = 25; // max number of tools before we show a warning
 
 export default function BottomMenu({
   hasMessages,
@@ -31,7 +32,7 @@ export default function BottomMenu({
   const dropdownRef = useRef<HTMLDivElement>(null);
   const toolCount = useToolCount();
   const { getProviders, read } = useConfig();
-  const [tokenLimit, setTokenLimit] = useState<number | null>(null);
+  const [tokenLimit, setTokenLimit] = useState<number>(TOKEN_LIMIT_DEFAULT);
 
   // Load providers and get current model's token limit
   const loadProviderDetails = async () => {
@@ -92,10 +93,10 @@ export default function BottomMenu({
     }
 
     // Add tool count alert if we have the data
-    if (toolCount !== null && toolCount > SUGGESTED_MAX_TOOLS) {
+    if (toolCount !== null && toolCount > TOOLS_MAX_SUGGESTED) {
       addAlert(
         AlertType.Warning,
-        `Too many tools can degrade performance.\nTool count: ${toolCount} (recommend: ${SUGGESTED_MAX_TOOLS})`,
+        `Too many tools can degrade performance.\nTool count: ${toolCount} (recommend: ${TOOLS_MAX_SUGGESTED})`,
         {
           text: 'View extensions',
           onClick: () => setView('settings'),


ui/desktop/src/App.tsx
```diff
@@ -349,7 +348,8 @@ export default function App() {
       console.error('Unhandled error in initialization:', error);
       setFatalError(`${error instanceof Error ? error.message : 'Unknown error'}`);
     });
-  }, [read, getExtensions, addExtension, enableRecipeConfigExtensionsV2]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []); // Empty dependency array since we only want this to run once
```
zanesq, line 353: realized we don't need these dependencies after all since we only want this to run once

ui/desktop/src/components/alerts/useToolCount.ts
```diff
@@ -3,61 +3,34 @@ import { getTools } from '../../api';
...
       } catch (err) {
         console.error('Error fetching tools:', err);
         setToolCount(0);
-        clearTimeout(timeoutId);
       }
     };
 
-    fetchTools();
+    // Add initial 1s delay before first fetch
+    timeoutId = setTimeout(fetchTools, 1000);
```
zanesq, line 27: @lily-de I ended up going with a delay instead of the step/retry logic. It was more reliable and didn't result in different results from fetch tools calls. Tradeoff is a slight delay before the alert shows but I think its better to report reliable values. We can revisit this when fetch tool calls are reliable on first try.

ui/desktop/src/components/bottom_menu/BottomMenu.tsx
```diff
@@ -10,7 +10,11 @@ import type { View } from '../../App';
 import { bottomMenuPopoverEnabled, settingsV2Enabled } from '../../flags';
 import { BottomMenuModeSelection } from './BottomMenuModeSelection';
 import ModelsBottomBar from '../settings_v2/models/bottom_bar/ModelsBottomBar';
-import { MAX_TOKENS, SUGGESTED_MAX_TOOLS, WARNING_THRESHOLD } from '../alerts/limits';
+import { useConfig } from '../ConfigContext';
+import { getCurrentModelAndProvider } from '../settings_v2/models/index';
+
+const TOKEN_WARNING_THRESHOLD = 0.8; // Show warning at 80% of max tokens
+const SUGGESTED_MAX_TOOLS = 1; // Low value for testing
```
zanesq: How many max tools do we want to suggest?

ui/desktop/src/components/settings_v2/extensions/ExtensionsSection.tsx
```diff
@@ -162,17 +157,6 @@ export default function ExtensionsSection({ deepLinkConfig, showEnvVars }: Exten
           These extensions use the Model Context Protocol (MCP). They can expand Goose's
           capabilities using three main components: Prompts, Resources, and Tools.
         </p>
-        {/* todo: some extensions have 0 tool count like code search? */}
```
zanesq, line 165: Removed tool count on extensions view per the description.

ui/desktop/src/components/settings_v2/models/subcomponents/AddModelModal.tsx
```diff
@@ -120,7 +128,7 @@ export const AddModelModal = ({ onClose, setView }: AddModelModalProps) => {
         activeProviders.forEach(({ metadata, name }) => {
           if (metadata.known_models && metadata.known_models.length > 0) {
             formattedModelOptions.push({
-              options: metadata.known_models.map((modelName) => ({
+              options: metadata.known_models.map(({ name: modelName }) => ({
```
zanesq, line 131: Updated to match the new schema from the backend with a name property.

crates/goose/src/providers/anthropic.rs
salman1993, line 1: i think it'd be cleaner to update here: https://github.com/block/goose/blob/29a3f9d4ebc385482aff21e1686ca8d93f1f919e/crates/goose/src/providers/base.rs#L39-L59
```
known_models: Vec<ModelInfo> -> known_model_names: Vec<&str>,

fn new() { 
      known_models = known_model_names
                .iter()
                .map(|&s| ModelInfo {
                    name: s.to_string(),
                    context_limit: ModelConfig::new(s.to_string()).context_limit(),
                })
                .collect(),
```

that way, we don't need to update each of the providers

crates/goose/src/providers/azure.rs
```diff
@@ -103,7 +103,10 @@ impl Provider for AzureProvider {
             "gpt-4o",
             AZURE_OPENAI_KNOWN_MODELS
                 .iter()
-                .map(|s| s.to_string())
+                .map(|&s| ModelInfo {
```
salman1993: this map can be avoided if we update the ProviderMetadata::new fn as suggested above

crates/goose/src/providers/anthropic.rs
zanesq: Thanks good call, I moved them all to base. For some reason I was having compilation errors with the original providers code so I asked goose for help and updated tests while I was at it. I verified that all providers are returned to the front end with valid info.

salman1993 (approved):

wendytang (commented):
Does the tool count take into account the 5 natively supported goose tools? May be worth removing those from the count


```
            prefixed_tools.push(platform_tools::search_available_extensions_tool());
            prefixed_tools.push(platform_tools::enable_extensions_tool());
            prefixed_tools.push(platform_tools::disable_extensions_tool());

            // Add resource tools if supported
            if extension_manager.supports_resources() {
                prefixed_tools.push(platform_tools::read_resource_tool());
                prefixed_tools.push(platform_tools::list_resources_tool());
```

zanesq:
@wendytang good question and suggestion. For the tool count the front end calls [get_tools](https://github.com/block/goose/blob/a1fe3bcbf17fe3740089a904ffdceae0004d099f/crates/goose-server/src/routes/agent.rs#L185). If it's easier we could get around this on the front end for now by bumping the limit by 5 from 25 to 30? @lily-de @salman1993 any thoughts?

zanesq:
I'm going to ahead and merge this to the parent branch so its all in one for review/launch. I'll update that branch with any decisions with the tool count logic.

PR #2201, opened 2025-04-15
feat: better support for gpt 4.1 with per model prompts

michaelneale:
gpt-4.1 behaves very differently to other LLMs in terms of taking agent like actions. 

This change adds optional model specific system prompts which are based on the guidelines: 

https://cookbook.openai.com/examples/gpt4-1_prompting_guide

to get gpt-4.1 to operate more like an agent.


salman1993:
why are there 3 prompt files?

crates/goose/src/agents/prompt_manager.rs
```diff
@@ -64,6 +65,17 @@ impl PromptManager {
         let base_prompt = if let Some(override_prompt) = &self.system_prompt_override {
             prompt_template::render_inline_once(override_prompt, &context)
                 .expect("Prompt should render")
+        } else if let Some(model) = model_name {
+            // Try to load a model-specific system prompt first
+            let model_specific_file = format!("system_{}.md", model.replace(['-', '/'], "_"));
```
salman1993: can we check for model == GPT_4_1 and then render the prompt file? trying to avoid having 3 duplicate prompt files

michaelneale: yep - could do that, would be neater, more a pattern match? as depends how we name it in databricks and openrouter calls it openai/gpt-4.1

michaelneale:
@salman1993 thanks for that - have simplified it with ability to add mappings from model name patterns to prompt, so no duplication. Also was able to improve the system prompt so gpt4.1 actually edits files and doesn't catastrophically erase content (!). Is working nice now.

salman1993 (approved):
nice!

michaelneale:
@salman1993 slightly more complex change now, as in some cases the real model name is hidden - so this will update it when the first response with the non aliased model name. Not entirely happy with this approach but it will work with all providers, and doesn't require a round trip (and just adds to what was there before)

crates/goose/src/agents/agent.rs
```diff
@@ -609,9 +609,16 @@ impl Agent {
     pub async fn create_recipe(&self, mut messages: Vec<Message>) -> Result<Recipe> {
         let extension_manager = self.extension_manager.lock().await;
         let extensions_info = extension_manager.get_extensions_info().await;
-        let system_prompt = self
-            .prompt_manager
-            .build_system_prompt(extensions_info, self.frontend_instructions.clone());
+
+        // Get model name from provider
+        let model_config = self.provider.get_model_config();
```
alexhancock, line 614: For databricks wouldn't this just come from 

https://github.com/block/goose/blob/970147f8adfe505d08a5c3c6f5c1acd2ee88ec07/crates/goose/src/providers/databricks.rs#L239-L241

?

If the model is aliased, would we still get the required `gpt_4_1`?

alexhancock (approved):

crates/goose/src/agents/agent.rs
michaelneale: @alexhancock the modelconfig I think would be what user has configured, which would be the alias if I understand correctly your question (so yeah, still need to get it back, infact it was databricks that started me requiring this)

michaelneale: and yeah - I was testing it with our "goose" databricks hosting endpoint, when it would return the appropriate model, and I checked it picked up the right promt.

bmaddox:
@michaelneale I noticed that when I select gpt-4.1 and ask Goose to repeat the system prompt back to me, the base system prompt is returned. I see the `prompt_manager.rs` references `system_gpt_4_1.md`, but the file itself is named `system_gpt_4.1.md`. Does this mean this gpt-4.1 specific prompt has never been referenced?

PR #2206, opened 2025-04-15
feat: Build an FFI for goose rust library

meenalc:
This currently only builds functions to initialize and send message to goose.
We will work on supporting tools in a follow up

### DIFF
--- a/crates/goose-ffi/README.md
+++ b/crates/goose-ffi/README.md
@@ -1,0 +1,127 @@
+# Goose FFI
+
+Foreign Function Interface (FFI) for the Goose AI agent framework, allowing integration with other programming languages.
+
+## Overview
+
+The Goose FFI library provides C-compatible bindings for the Goose AI agent framework, enabling you to:
+
+- Create and manage Goose agents from any language with C FFI support
+- Configure and use the Databricks AI provider for now but is extensible to other providers as needed
+- Send messages to agents and receive responses
+
+## Building
+
+To build the FFI library, you'll need Rust and Cargo installed. Then run:
+
+```bash
+# Build the library in debug mode
+cargo build --package goose_ffi
+
+# Build the library in release mode (recommended for production)
+cargo build --release --package goose_ffi
+```
+
+This will generate a dynamic library (.so, .dll, or .dylib depending on your platform) in the `target` directory, and automatically generate the C header file in the `include` directory.
+
+You can also build cross-platform binaries using cross command. For example to build for linux x86_64 architecture from Mac would require running
+
+```bash
+CROSS_BUILD_OPTS="--platform linux/amd64 --no-cache" CROSS_CONTAINER_OPTS="--platform linux/amd64" cross build -p goose-ffi --release --target x86_64-unknown-linux-gnu --no-default-features
+```
+Note that this works only for gnu linux as it requires glibc.
+
+## Generated C Header
+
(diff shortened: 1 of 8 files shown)


michaelneale:
this is really cool - I love the python example

zhong-sq (approved):
LGTM!

tested with kotlin service locally, it works as expected!

crates/goose-ffi/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+    // Block on the async call using our global runtime
+    let response = get_runtime().block_on(async {
+        let mut stream = match agent.reply(&messages, None).await {
+            Ok(stream) => stream,
+            Err(e) => return format!("Error getting reply from agent: {}", e),
+        };
+
+        let mut full_response = String::new();
+
+        while let Some(message_result) = stream.next().await {
```
zhong-sq, line 254: currently `goose_agent_send_message` waits for the complete agent reply before returning, once we add the extensions, we may change how we respond the message here.

crates/goose-ffi/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+
+// Extension configuration will be implemented in a future commit
+
+/// Message structure for agent interactions
+///
+/// - role: 0 = user, 1 = assistant, 2 = system
+/// - content: Text content of the message
+#[repr(C)]
+pub struct MessageFFI {
+    pub role: u32,
```
zakiali: maybe make an enum for the role here instead of using int values?

crates/goose-ffi/src/lib.rs
```diff
@@ -0,0 +1,296 @@
+use std::ffi::{c_char, CStr, CString};
+use std::ptr;
+use std::sync::Arc;
+
+use futures::StreamExt;
+use goose::agents::Agent;
+use goose::message::Message;
+use goose::model::ModelConfig;
+use goose::providers::databricks::DatabricksProvider;
```
zakiali, line 9: we should probably have an openai and ollama provider as well for open source

crates/goose-ffi/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+    if config.is_null() {
+        eprintln!("Error: config pointer is null");
+        return AgentPtr(ptr::null_mut());
+    }
+
+    let config = &*config;
+
+    // Check if the provider type is supported
+    match config.provider_type {
+        0 => { /* Databricks provider is supported */ }
```
zakiali: similar comment as above for providers -> enums

crates/goose-ffi/README.md
```diff
@@ -0,0 +1,127 @@
...
+3. Setting up C-compatible structures
+4. Creating an object-oriented API for easier use
+
+Note: Tool callback functionality shown in earlier versions is not currently available and will be implemented in a future release.
+
+To run the Python example:
+
+```bash
+# First, build the FFI library
+cargo build --release --package goose_ffi
```
zakiali, line 57: these should be goose-ffi

crates/goose-ffi/src/lib.rs
meenalc: absolutely, I was thinking that could be a follow up once we have nailed what library we want, I want this to be in alpha and can have breaking changes for now until we have tool extensions, let me know if you prefer, have a linear ticket tracking this

zakiali (approved):

PR #2212, opened 2025-04-15
chore: refactor read-write lock on agent

ahau-square:


crates/goose-server/src/state.rs
```diff
@@ -3,23 +3,31 @@ use goose::agents::Agent;
...
 use std::sync::Arc;
-use tokio::sync::{Mutex, RwLock};
+use tokio::sync::Mutex;
 
 /// Shared application state
 #[allow(dead_code)]
 #[derive(Clone)]
 pub struct AppState {
-    pub agent: Arc<RwLock<Option<Agent>>>,
+    pub agent: Arc<Agent>,
```
salman1993: i think it needs to be `Option<Arc<Agent>>`

crates/goose-server/src/routes/extension.rs
```diff
@@ -320,9 +319,8 @@ async fn remove_extension(
         return Err(StatusCode::UNAUTHORIZED);
     }
 
-    // Acquire a lock on the agent and attempt to remove the extension
-    let mut agent = state.agent.write().await;
```
salman1993, line 324: the following should work with appstate agent being an option<arc<agent>>
```
    let agent = state.agent;
    if agent.is_none() {
        return Err(StatusCode::PRECONDITION_FAILED); // sth like this
    }
    let agent = agent.unwrap();
    agent.some_method(...)
```

PR #2216, opened 2025-04-15
fix: propagate errors when attempting to add extensions

kalvinnchau:
previously it was a let _ = ... which would ignore the error from the result and always look like a success


zakiali (approved):
lgtm! able to see the errors propagate to the ui now

PR #2217, opened 2025-04-15
feat: Add github workflow to build and release the ffi libraries

meenalc:
Adding a github workflow to build and release the ffi libraries. Intentionally not pushing it in main release so we can test and make changes as needed.


baxen:
Now out of date, moved to uniffi

PR #2218, opened 2025-04-15
feat: use the permission page in bottom menu

yingjiehe-xyz:
We use the parentView to track which view page we should go back from permission page. When we enter permission view from the bottom menu, we should go back to the chat view, otherwise, go back to the setting view

https://github.com/user-attachments/assets/30de6e10-68a4-418d-82eb-947dbb96fc9a

### DIFF
--- a/ui/desktop/src/App.tsx
+++ b/ui/desktop/src/App.tsx
@@ -781,7 +781,11 @@ export default function App() {
               }}
             />
           )}
-          {view === 'permission' && <PermissionSettingsView onClose={() => setView('settings')} />}
+          {view === 'permission' && (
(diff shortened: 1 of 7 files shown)


yingjiehe-xyz:
> Code looks good to me, couple of things I noticed.
> 
> The mode popup seems to have lost its padding, doesn't look like it was part of this PR but can you add it back? image vs image
> 
> I noticed there is a similar `filterGooseModes` in ui/desktop/src/components/settings_v2/mode/ModeSelectionItem.tsx but it appears to be unused.. Should we remove it or use that one instead?

thanks, update the code. I haven't remove any padding for the popup, let me add some

PR #2224, opened 2025-04-16
[Draft] feat: make config and session log optional

meenalc:


crates/goose/src/config/base.rs
```diff
@@ -122,7 +137,7 @@ impl Default for Config {
 
         std::fs::create_dir_all(&config_dir).expect("Failed to create config directory");
 
-        let config_path = config_dir.join("config.yaml");
+        let config_path = Some(config_dir.join("config.yaml"));
 
         let secrets = match env::var("GOOSE_DISABLE_KEYRING") {
             Ok(_) => SecretStorage::File {
```
meenalc: need to handle things here

PR #2225, opened 2025-04-16
chore: refactor read-write lock on agent

salman1993:
how i tested this:

- terminal 1: `cargo build; ./target/debug/goosed agent` 
- terminal 2: run script below 
```bash
#!/bin/bash
set -euxo pipefail
IFS=$'\n\t'

# Send request to create an agent
curl --request POST \
  --url http://localhost:3000/agent \
  --header 'Content-Type: application/json' \
  --header 'X-Secret-Key: test' \
  --data '{
    "provider": "databricks"
  }'

sleep 5

# Add a system
curl --request POST \
  --url http://localhost:3000/extensions/add \
  --header 'Content-Type: application/json' \
  --header 'X-Secret-Key: test' \
  --data '{
    "type": "builtin",
    "name": "developer"
  }'

sleep 5

# Send a user message 
curl --request POST \
  --url http://localhost:3000/reply \
  --header 'Accept: text/event-stream' \
  --header 'Content-Type: application/json' \
  --header 'X-Secret-Key: test' \
  --header 'x-protocol: data' \
  --data '{
  "session_working_dir": "/Users/smohammed",
  "messages": [
    {
      "role": "user",
      "created": 1740670518,
      "content": [
        {
          "type": "text", 
          "text": "can you enable fetch extension for me?"
        }
      ]
    }
  ]
}'
````

- in a new terminal you'll have to confirm the enable extension request:
```
curl --request POST \
  --url http://localhost:3000/confirm \
  --header 'Content-Type: application/json' \
  --header 'X-Secret-Key: test' \
  --data '{
    "id": "toolu_bdrk_01GTPbCJ4sHUecsQGD9BEccj",  # THIS WILL BE DIFFERENT FOR YOU
    "principal_type": "Extension", 
    "action": "allow_once"
  }'

```

Screenshot 2025-04-16 at 10 46 13 AM

### CODE REVIEW

File: crates/goose-server/src/commands/agent.rs [hunk 1]
```diff
@@ -15,8 +18,19 @@ pub async fn run() -> Result<()> {
     let secret_key =
         std::env::var("GOOSE_SERVER__SECRET_KEY").unwrap_or_else(|_| "test".to_string());
 
-    // Create app state - agent will start as None
-    let state = state::AppState::new(secret_key.clone()).await?;
+    // TODO: Create agent using provider and model as args to CLI command or environment variables
```
salman1993: i think we'll need to add CLI params to `goosed` here for initial model and provider

examples/frontend_tools.py
```diff
@@ -33,12 +33,28 @@
     },
 }
 
+# Enable Extension tool definition
```
ahau-square, line 36: @salman1993 should I just revert this file since this was just for our own testing?

ahau-square:
Latest changes should make these changes work with UI/CLI so you can test them out from there

ahau-square:
This PR 

- refactors to avoid requiring a lock on the whole agent to read/write its fields (the previous rw lock was preventing us from being able to access the agent's extension manager while inside the reply loop)
- moves provider behind its own mutex 
- creates single reference to the agent to be shared by AppState
- removes the create_agent server endpoint since now the AppState is created with reference to a new agent in `commands/agent.rs`
- after an agent is created, the new `/agent/update_provider` is used to update the provider on the agent 


In a future PR will have frontend_tools and frontend_instructions managed by frontend_tool_manager similar to extension_manager

@salman1993 mentioned he's going to remove token counter soon but leaving it in for now

crates/goose-server/src/routes/agent.rs
```diff
@@ -198,14 +145,24 @@ async fn get_tools(
...
     let mut tools: Vec<ToolInfo> = agent
-        .list_tools(query.extension_name)
+        .list_tools(None)
         .await
         .into_iter()
+        .filter(|tool| {
+            // Apply the filter only if the extension name is present in the query
+            if let Some(extension_name) = &query.extension_name {
+                tool.name.starts_with(extension_name)
+            } else {
```
yingjiehe-xyz: bad merge? It should be `.list_tools(query.extension_name)`

crates/goose-server/src/routes/agent.rs
```diff
@@ -232,12 +189,61 @@ async fn get_tools(
...
+        return Err(StatusCode::UNAUTHORIZED);
+    }
+
+    let agent = state
+        .get_agent()
+        .await
+        .map_err(|_| StatusCode::PRECONDITION_FAILED)?;
+
+    // Set the environment variable for the model if provided
+    if let Some(model) = &payload.model {
```
yingjiehe-xyz: get from env should be covered by config.get_param: https://github.com/block/goose/blob/571ba86157725c177f087e8acc4f4ecbe5d8343e/crates/goose/src/config/base.rs#L292

crates/goose-server/src/routes/agent.rs
```diff
@@ -232,12 +189,61 @@ async fn get_tools(
...
+        let env_var_key = format!("{}_MODEL", payload.provider.to_uppercase());
+        env::set_var(env_var_key.clone(), model);
+        println!("Set environment variable: {}={}", env_var_key, model);
+    }
+
+    let config = Config::global();
+    let model = payload.model.unwrap_or_else(|| {
+        config
+            .get_param("GOOSE_MODEL")
+            .expect("Did not find a model on payload or in env to update provider with")
```
yingjiehe-xyz, line 208: do we want to update the config yaml as well?

crates/goose-server/src/routes/agent.rs
```diff
@@ -232,12 +189,61 @@ async fn get_tools(
...
+}
+
+pub fn routes(state: Arc<AppState>) -> Router {
     Router::new()
         .route("/agent/versions", get(get_versions))
         .route("/agent/providers", get(list_providers))
         .route("/agent/prompt", post(extend_prompt))
         .route("/agent/tools", get(get_tools))
-        .route("/agent", post(create_agent))
+        .route("/agent/update_provider", post(update_agent_provider))
```
yingjiehe-xyz, line 226: can we add the route to config_management.ts? As we have `.route("/config/providers", get(providers))` there.
And for update api, can we do something like route("/config/providers", post(providers))?

crates/goose/src/agents/agent.rs
```diff
@@ -540,13 +563,35 @@ impl Agent {
...
+    }
+
+    /// Update the provider used by this agent
+    pub async fn update_provider(
+        &self,
+        provider_name: &str,
+        model_config: ModelConfig,
+    ) -> Result<()> {
+        let new_provider = crate::providers::create(provider_name, model_config)?;
+        let token_counter = TokenCounter::new(new_provider.get_model_config().tokenizer_name());
```
yingjiehe-xyz: how about:

```
let new_provider = crate::providers::create(provider_name, model_config)?;
update_provider_with_provider(new_provider)
```

yingjiehe-xyz (approved):

crates/goose-ffi/src/lib.rs
```diff
@@ -120,7 +120,7 @@ pub unsafe extern "C" fn goose_free_async_result(result: *mut AsyncResult) {
 /// The config pointer must be valid or NULL. The resulting agent must be freed
 /// with goose_agent_free when no longer needed.
 #[no_mangle]
-pub unsafe extern "C" fn goose_agent_new(config: *const ProviderConfigFFI) -> AgentPtr {
+pub async unsafe extern "C" fn goose_agent_new(config: *const ProviderConfigFFI) -> AgentPtr {
```
kalvinnchau: are there any implications making the ffi call an `async` function ? looks like most of the extern functions are sync

examples/frontend_tools.py
kalvinnchau, line 1: this file doesn't work for me any more - can this be updated so we can test it out?
```bash
httpx.HTTPStatusError: Client error '404 Not Found' for url 'http://127.0.0.1:3000/agent'
```
looks like this needs to be updated to match the changes in the routes

crates/goose/src/agents/agent.rs
```diff
@@ -42,32 +43,31 @@ const ESTIMATE_FACTOR_DECAY: f32 = 0.9;
...
-    pub fn new(provider: Arc<dyn Provider>) -> Self {
-        let token_counter = TokenCounter::new(provider.get_model_config().tokenizer_name());
+    pub fn new() -> Self {
         // Create channels with buffer size 32 (adjust if needed)
         let (confirm_tx, confirm_rx) = mpsc::channel(32);
         let (tool_tx, tool_rx) = mpsc::channel(32);
 
         Self {
-            provider,
+            provider: Mutex::new(None),
```
kalvinnchau, line 56: is decoupling the agent and provider here to support swapping providers during runtime?
for the most part the agent can't function without a provider so having an an option rather than making it required is unexpected

crates/goose/src/agents/reply_parts.rs
```diff
@@ -31,25 +32,28 @@ impl Agent {
         let extensions_info = extension_manager.get_extensions_info().await;
 
         // Get model name from provider
-        let model_config = self.provider.get_model_config();
+        let provider = self.provider.lock().await;
+        let model_config = provider.as_ref().unwrap().get_model_config();
```
kalvinnchau: this code panics using the script provided in the initial PR message, looks like because that script doesn't match the updated initialization logic we should be careful with bare `unwrap()`:

```bash
thread 'tokio-runtime-worker' panicked at .../goose/src/agents/reply_parts.rs:36:46:
called `Option::unwrap()` on a `None` value
```

crates/goose/src/agents/reply_parts.rs
```diff
@@ -31,25 +32,28 @@ impl Agent {
         let extensions_info = extension_manager.get_extensions_info().await;
 
         // Get model name from provider
-        let model_config = self.provider.get_model_config();
+        let provider = self.provider.lock().await;
```
kalvinnchau: think we should be using the newly introduced `self.provider().await?` approach here https://github.com/block/goose/pull/2225/files#diff-0afb949c990abef615e434c1f1d01ce0e64fa7a78d61f3627a092fc03735325cR79

crates/goose-server/src/routes/agent.rs
kalvinnchau, line 226: there's probably not too many places where this is already used, `POST /agent` and its really only used by us; but i would consider this a breaking change - we'd need to get this info out

`POST /agent` is now `POST /agent/update_provider` - looks like the payload is the same though?

kalvinnchau (commented):
Can you provide an updated script and `frontend_tools.py` like the initial pr comment we can use to test/verify with?

Some general comments:
* we're still manually locking `self.provider.lock().await;` in some places, but there's a new helper method on the agent `pub async fn provider(&self) -> Result<Arc<dyn Provider>, anyhow::Error> {` that we should standardize on using that handles the locking/cloning/erroring for us
* Handle cases where we're using bare `unwrap()` since the provider is now `Option` we shouldn't panic now in cases where we didn't before
* is there a way to add an explicit test case that prompted this PR - so we can verify that the deadlock no longer exists?

crates/goose/src/agents/agent.rs
ahau-square: right, we wanted to be able to swap out providers on the same agent and maintain a singular agent reference in the AppState 

we initialize it with no provider since when the app state is first created in commands/agent.rs, we don't know the model/provider yet

crates/goose-server/src/routes/agent.rs
ahau-square: good point, this works in CLI / UI as the goose server starts by running `agent::run().await` which instantiates a goose agent for the app state

the payload is the same as `/update_provider` just updates the existing agent reference on the app state with a new provider/model instead of creating and returning a new agent

any idea who this might impact?

crates/goose-server/src/routes/agent.rs
ahau-square: thanks! good catch, looks like it was a bad merge

crates/goose-server/src/routes/agent.rs
ahau-square: the config yaml is already updated in the UI when switching - frontend makes a call to `/config/upsert`. Similarly the CLI updates config with call to `config.set_param` in configure.rs.

crates/goose-server/src/routes/agent.rs
ahau-square: on second look, I ported this from the original /agent creation implementation, but I don't think we use this env var `{}_MODEL` anywhere anymore unless I'm mistaken, so will just remove

crates/goose/src/agents/agent.rs
ahau-square: good idea, consolidating into a single update_provider method that just takes a provider

crates/goose-ffi/src/lib.rs
ahau-square: good call, reverting out of async and following the pattern for other awaits in this file

crates/goose/src/agents/reply_parts.rs
ahau-square: yep updated here and elsewhere!

crates/goose/src/agents/reply_parts.rs
ahau-square: I updated `routes/reply.rs` to check for whether a provider exists on the agent to send back the message "No provider configured" if so

ahau-square:
> Can you provide an updated script and `frontend_tools.py` like the initial pr comment we can use to test/verify with?
> 
> Some general comments:
> 
> * we're still manually locking `self.provider.lock().await;` in some places, but there's a new helper method on the agent `pub async fn provider(&self) -> Result<Arc<dyn Provider>, anyhow::Error> {` that we should standardize on using that handles the locking/cloning/erroring for us
> * Handle cases where we're using bare `unwrap()` since the provider is now `Option` we shouldn't panic now in cases where we didn't before
> * is there a way to add an explicit test case that prompted this PR - so we can verify that the deadlock no longer exists?

@kalvinnchau you should be able to test frontend_tools.py again

1. start server with `cargo build; ./target/debug/goosed agent` in one terminal window
2. run frontend_tools.py
3. enter into the chat this instruction to goose `use enable_extension tool with "fetch" extension name` . This will guide the agent to use its frontend tool for enabling the fetch extension, and it should succeed and exit the chat.
Before, this attempt to modify the agent's extensions in the reply loop deadlocked. So this should prove that we are now able to successfully update the agent's extensions mid-reply loop without deadlocking.

crates/goose-server/src/routes/agent.rs
lilydelalande: do we expect that the latest model and provider are written to config.yaml _before_ we update the agent? because I think we are waiting to update the agent and _then_ write to the config (maybe outdated with this pr): https://github.com/block/goose/blob/main/ui/desktop/src/components/settings_v2/models/index.ts#L30

ahau-square: just following same pattern as before, model/provider will get updated on agent and then config will be updated via ui/cli after

kalvinnchau (approved):

crates/goose/src/agents/agent.rs
```diff
@@ -40,56 +40,66 @@ const ESTIMATE_FACTOR_DECAY: f32 = 0.9;
...
+    pub(super) provider: Mutex<Option<Arc<dyn Provider>>>,
     pub(super) extension_manager: Mutex<ExtensionManager>,
-    pub(super) frontend_tools: HashMap<String, FrontendTool>,
-    pub(super) frontend_instructions: Option<String>,
-    pub(super) prompt_manager: PromptManager,
-    pub(super) token_counter: TokenCounter,
+    pub(super) frontend_tools: Mutex<HashMap<String, FrontendTool>>,
+    pub(super) frontend_instructions: Mutex<Option<String>>,
+    pub(super) prompt_manager: Mutex<PromptManager>,
+    pub(super) token_counter: Mutex<Option<TokenCounter>>,
```
wendytang: This is pretty interesting -- is the takeaway here that applying `Mutex` and locking allows a struct to avoid self mutability?

examples/frontend_tools.py
```diff
@@ -47,7 +63,7 @@ async def setup_agent() -> None:
     async with httpx.AsyncClient() as client:
         # First create the agent
         response = await client.post(
-            f"{GOOSE_URL}/agent",
+            f"{GOOSE_URL}/config/update_provider",
```
wendytang: `/agent/update_provider` ?

ahau-square: yes thanks good catch!

crates/goose/src/agents/agent.rs
ahau-square: by wrapping the fields in Mutex, we're enabling internal mutability. This means we don't need to lock the entire agent struct to modify individual fields. Instead, we can lock and mutate just the specific field we care about, avoiding broader write locks and allowing concurrent access to different fields

crates/goose-cli/src/commands/configure.rs
```diff
@@ -983,11 +983,11 @@ pub async fn configure_tool_permissions_dialog() -> Result<(), Box<dyn Error>> {
...
         .expect("No model configured. Please set model first");
     let model_config = goose::model::ModelConfig::new(model.clone());
-    let provider =
-        goose::providers::create(&provider_name, model_config).expect("Failed to create provider");
 
     // Create the agent
-    let mut agent = Agent::new(provider);
+    let agent = Agent::new();
+    let new_provider = create(&provider_name, model_config)?;
+    agent.update_provider(new_provider).await?;
```
salman1993, line 990: might be useful to have a constructor on Agent that does these 3 lines of code - `Agent::with_provider(...)`

crates/goose/src/agents/agent.rs
salman1993: agree with @kalvinnchau that it feels unintuitive to have the provider be optional when agent can't work without it. 

would be good to document this for now and i think we wanna revisit this flow so that provider is not Optional but we have a flow to make sure it exists when Agent gets created for the first time. is this an issue both in the UI and CLI? for `commands/agent.rs`, can we add arguments to `goosed agent` CLI command? 
```
goosed agent --provider databricks --model goose
```

salman1993: looking at the UI code, we get the provider and model info into appConfig:
https://github.com/block/goose/blob/428f0313865fa75ef308282fe07380ec6875321c/ui/desktop/src/main.ts#L261-L275

which seems to be available when we start `goosed` here: 
https://github.com/block/goose/blob/428f0313865fa75ef308282fe07380ec6875321c/ui/desktop/src/main.ts#L331

PR #2226, opened 2025-04-16
(wip) feat: disable extensions

wendytang:
Part of AAI-301

image


yingjiehe-xyz (approved):

crates/goose/src/agents/agent.rs
```diff
@@ -194,46 +195,101 @@ impl Agent {
...
+        for extension_name in &extension_names {
+            match ExtensionConfigManager::get_config_by_name(extension_name) {
+                Ok(Some(config)) => configs.push(config),
+                Ok(None) => {
+                    return (
+                        request_id,
+                        Err(ToolError::ExecutionError(format!(
                         "Extension '{}' not found. Please check the extension name and try again.",
                         extension_name
                     ))),
```
ahau-square: maybe instead return the list of extensions that can't be found? so that it knows which of all the requested extensions failed

crates/goose/src/agents/agent.rs
```diff
@@ -194,46 +195,101 @@ impl Agent {
...
+            ref name if name == PLATFORM_ENABLE_EXTENSIONS_TOOL_NAME => {
+                let mut success_messages = Vec::new();
+                let mut errors = Vec::new();
+
+                // Process each config
+                for config in configs {
+                    let name = config.name();
+                    match extension_manager.add_extension(config).await {
+                        Ok(_) => success_messages.push(format!(
+                            "The extension '{}' has been installed successfully",
```
ahau-square: can we trim the success message so it's not repeats of this sentence but instead something like the extensions [list of all extensions] have been installed successfully. minor but will help save tokens

crates/goose/src/agents/agent.rs
```diff
@@ -194,46 +195,101 @@ impl Agent {
...
+            ref name if name == PLATFORM_DISABLE_EXTENSIONS_TOOL_NAME => {
+                let mut success_messages = Vec::new();
+                let mut errors = Vec::new();
+
+                // Process each config
+                for config in configs {
+                    let name = config.name();
+                    match extension_manager.remove_extension(&name).await {
+                        Ok(_) => success_messages.push(format!(
+                            "The extension '{}' has been removed successfully",
```
ahau-square: similar to comment above

crates/goose/src/agents/platform_tools.rs
```diff
@@ -6,7 +6,8 @@ pub const PLATFORM_READ_RESOURCE_TOOL_NAME: &str = "platform__read_resource";
 pub const PLATFORM_LIST_RESOURCES_TOOL_NAME: &str = "platform__list_resources";
 pub const PLATFORM_SEARCH_AVAILABLE_EXTENSIONS_TOOL_NAME: &str =
     "platform__search_available_extensions";
-pub const PLATFORM_ENABLE_EXTENSION_TOOL_NAME: &str = "platform__enable_extension";
+pub const PLATFORM_ENABLE_EXTENSIONS_TOOL_NAME: &str = "platform__enables_extension";
+pub const PLATFORM_DISABLE_EXTENSIONS_TOOL_NAME: &str = "platform__disables_extension";
```
ahau-square: nit: should we update tool names to have extension plural since now you can pass multiple extensions?

ie enable_extensions, disable_extensions

ui/desktop/src/components/ExtensionConfirmation.tsx
```diff
@@ -47,7 +47,7 @@ export default function ExtensionConfirmation({
   ) : (
     <>
       <div className="goose-message-content bg-bgSubtle rounded-2xl px-4 py-2 rounded-b-none text-textStandard">
-        Goose would like to {extensionAction} the following extension. Allow?
+        Goose would like to {extensionAction} the following extensions. Allow?
```
ahau-square, line 50: One consequence of this is that users will have to approve/deny a single request that might want to install multiple extensions. If they wanted to install some on the list but not all, they'd need to deny the request and then instruct goose to install the right ones (vs before, they'd just approve/deny one by one). Was this an intentional change in the experience?

crates/goose/src/agents/platform_tools.rs
ahau-square: nit: I think tool names should be enable_ / disable_ (rather than enables / disables?)

ui/desktop/src/components/ExtensionConfirmation.tsx
wendytang: It was an intentional change. While testing, it felt more natural to be able to disable a few extensions at once, hence the change to plural.

wendytang:
goose on occasion will bring up disabling extensions and explain the benefits
image

wendytang:
Currently, goose doesn't know the list of extensions enabled for the agent (not enabled in the config). This happens when an extension is enabled through mcp router. 

We need the equivalent of `search_available_extensions` for disables. The remaining item would be to add this equivalent so goose knows what it can disable.

wendytang:
closing in lieu of #2319

PR #2228, opened 2025-04-16
fix: expose tool not support error

yingjiehe-xyz:
Support parsing error code as number, in openrouter, the error code is number instead of string, so we will not expose the error detail to user.

Before:
image

After:
image


wendytang (approved):
❇️

PR #2249, opened 2025-04-17
fix: specify extension env vars only by name

alexhancock:
Minimal change to address env var issues with extensions in settings v2

### DIFF
--- a/ui/desktop/src/components/settings_v2/extensions/agent-api.ts
+++ b/ui/desktop/src/components/settings_v2/extensions/agent-api.ts
@@ -2,6 +2,8 @@ import { ExtensionConfig } from '../../../api/types.gen';
 import { getApiUrl, getSecretKey } from '../../../config';
 import { toastService, ToastServiceOptions } from '../../../toasts';
 import { replaceWithShims } from './utils';
+import { saveEnvVarsToKeyring } from './extension-manager';
+import type { AgentExtensionConfig } from './extension-manager';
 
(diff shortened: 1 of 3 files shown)


ui/desktop/src/components/settings_v2/extensions/agent-api.ts
```diff
@@ -141,13 +143,21 @@ export async function addToAgent(
...
+    const ext = extension as AgentExtensionConfig;
+
+    if (ext.type === 'stdio') {
+      ext.cmd = await replaceWithShims(ext.cmd);
     }
 
-    extension.name = sanitizeName(extension.name);
+    ext.name = sanitizeName(ext.name);
+
+    if (ext.type === 'stdio') {
```
lilydelalande: would add sse as it has an envs param too

ui/desktop/src/components/settings_v2/extensions/extension-manager.ts
```diff
@@ -283,3 +289,11 @@ export async function deleteExtension({ name, removeFromConfig }: DeleteExtensio
     throw agentRemoveError;
   }
 }
+
+export async function saveEnvVarsToKeyring(extension: ExtensionConfig) {
+  if (extension.type === 'stdio') {
```
lilydelalande: or sse

lily-de (approved):

kalvinnchau (approved):

PR #2250, opened 2025-04-17
feat: extension manager suggestion

wendytang:


crates/goose/src/agents/extension_manager.rs
```diff
@@ -226,6 +229,30 @@ impl ExtensionManager {
...
+        let enabled_extensions_count = self.clients.len();
+        let total_tools = self
+            .get_prefixed_tools(None)
+            .await
+            .map(|tools| tools.len())
+            .unwrap_or(0);
+
+        // Check if either condition is met
+        const MIN_EXTENSIONS: usize = 3;
+        const MIN_TOOLS: usize = 32;
```
ahau-square: how did you decide on these cutoffs? 3 extensions might be a little low

crates/goose/src/agents/prompt_manager.rs
```diff
@@ -60,6 +61,18 @@ impl PromptManager {
         let current_date_time = Utc::now().format("%Y-%m-%d %H:%M:%S").to_string();
         context.insert("current_date_time", Value::String(current_date_time));
 
+        // Add the suggestion about disabling extensions if flag is true
+        if should_suggest_disable {
```
ahau-square: what happens without this conditional and if you just always kept this in the system prompt? does the model try and disable extensions all the time? 

We could just add to the system prompt context on the number of extensions and umber of tools similar to how we add current_date_time above. Like

number of enabled extensions: xxx 
number of available tools: xxx

Then update the prompting to suggest using suggest_disable tool if extensions or tools exceeds 3/32.

crates/goose/src/agents/extension_manager.rs
```diff
@@ -226,6 +229,30 @@ impl ExtensionManager {
         Ok(())
     }
 
+    pub async fn should_suggest_disable(&self) -> bool {
```
ahau-square: why do we need to count tools and extensions periodically? I would think we should already know the counts from every time we add and disable an extension

ahau-square (commented):
We might want to try and simplify this implementation by removing the conditional logic for building the system prompt and possibly also removing the cooldown. I think we're generally trying to keep the agent really lightweight if possible

crates/goose/src/agents/extension_manager.rs
wendytang: https://linear.app/squareup/issue/AAI-301/implement-disable-extension-tool

crates/goose/src/agents/prompt_manager.rs
wendytang: > We could just add to the system prompt context on the number of extensions and umber of tools similar to how we add current_date_time above. Like

We can take that approach. Do we want to keep the rate limit and cooldown logic though? It will make sure goose doesn't over-call disable extensions.

PR #2256, opened 2025-04-17
Fix e2e tests

zanesq:
Fix e2e tests after all the recent ui changes, improved some selectors.

One remaining issue is we need to add a way to properly remove/cleanup the test extension through the ui or directly in the config now that they are stored in the config and no longer in local storage. Currently the extension stays installed in the app so you need to disable it in the app after.

Run with `npm run test-e2e` or `npm run test-e2e:ui` for interactive runner.

### DIFF
--- a/ui/desktop/tests/e2e/app.spec.ts
+++ b/ui/desktop/tests/e2e/app.spec.ts
@@ -406,6 +406,7 @@ test.describe('Goose App', () => {
           }
 
           // Clean up any existing running-quotes extensions from localStorage
+          // todo: extensions are no longer in localstorage so we need a way to delete extensions from the ui or config to actually clean these up
           await mainWindow.evaluate(() => {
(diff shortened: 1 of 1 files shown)


ui/desktop/src/components/ChatView.tsx
```diff
@@ -391,7 +391,11 @@ export default function ChatView({
           <ScrollArea ref={scrollRef} className="flex-1" autoScroll>
             <SearchView>
               {filteredMessages.map((message, index) => (
-                <div key={message.id || index} className="mt-4 px-4">
+                <div
+                  key={message.id || index}
+                  className="mt-4 px-4"
+                  data-testid="message-container"
```
zanesq, line 397: Started adding testids where ever I can to avoid relying on css class selectors that have potential to change

yingjiehe-xyz (approved):

PR #2260, opened 2025-04-18
GUI warning when installing extension from deeplink

michaelneale:
This uses the GOOSE_ALLOWLIST (if set) to check that the extension being proposed from a deeplink is valid: 

Screenshot 2025-04-18 at 3 55 26 pm

(you can override and install it after this warning). if there is no allowlist or not valid, it will not show
This will not affect already installed extensions

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -796,6 +796,25 @@ app.whenReady().then(async () => {
         },
       })
     );
+
+    // NOTE: Add Extension menu item DEVELOPMENT TIME ONLY REMOVE THIS
+    fileMenu.submenu.append(
+      new MenuItem({
+        label: 'Add Extension',
+        click() {
+          // Create extension deeplink
+          const extensionUrl =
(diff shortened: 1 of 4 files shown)


ui/desktop/src/main.ts
```diff
@@ -903,6 +887,60 @@ app.whenReady().then(async () => {
...
+/**
+ * Fetches the allowed extensions list from the remote YAML file if GOOSE_ALLOWLIST is set.
+ * If the ALLOWLIST is not set, any are allowed. If one is set, it will warn if the deeplink 
+ * doesn't match a command from the list. 
+ * If it fails to load, then it will return an empty list.
+ * If the format is incorrect, it will return an empty list.
+ * Format of yaml is:
+ *  
+ ```yaml:
+ extensions:
```
alexhancock, line 899: idea to consider in future: make the format of this the same as the yaml blocks we use in `~/.config/goose/config.yaml` so we can compare extension configs as a whole.

alexhancock (approved):

ui/desktop/src/main.ts
michaelneale: yeah could do that.

PR #2261, opened 2025-04-18
feat: Make configuration ephemeral for goose

meenalc:
This is to allow ffi interfaces to use goose in ephemeral mode, I have tested this in a read only docker environment

### DIFF
--- a/crates/goose/src/config/base.rs
+++ b/crates/goose/src/config/base.rs
@@ -9,7 +9,23 @@ use std::env;
 use std::fs::OpenOptions;
 use std::io::Write;
 use std::path::{Path, PathBuf};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, Mutex};
 use thiserror::Error;
+/// Atomic flag for programmatic control of in-memory configuration.
+static EPHEMERAL_IN_MEMORY: AtomicBool = AtomicBool::new(false);
+
+/// Programmatically enable or disable in-memory configuration.
+/// When enabled, Config::default() and related constructors will
+/// yield in-memory storage regardless of environment variables.
+pub fn set_in_memory(on: bool) {
+    EPHEMERAL_IN_MEMORY.store(on, Ordering::SeqCst);
+}
+
+/// Returns true if in-memory configuration is enabled via the atomic flag.
+pub fn is_in_memory() -> bool {
+    EPHEMERAL_IN_MEMORY.load(Ordering::SeqCst)
+}
 
 pub static APP_STRATEGY: Lazy<AppStrategyArgs> = Lazy::new(|| AppStrategyArgs {
     top_level_domain: "Block".to_string(),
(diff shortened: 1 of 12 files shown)


zhong-sq (approved):
Nice! 👍 
This will make the Goose service much easier!

crates/goose/src/config/base.rs
```diff
@@ -98,21 +126,52 @@ impl From<keyring::Error> for ConfigError {
 /// environment variable OPENAI_API_KEY
 ///
 /// For Goose-specific configuration, consider prefixing with "goose_" to avoid conflicts.
+#[derive(Clone, Debug)]
+enum ConfigStorage {
+    File {
+        path: PathBuf,
+    },
+    Memory {
+        values: std::sync::Arc<std::sync::Mutex<HashMap<String, Value>>>,
```
zakiali: do we need Arc? haven't read everything, but are you planning on cloning the config between threads and wanting to mutate? We don't do this now. 

(nit: would be good to import and use Arc/Mutex around instead of the full path for readability)

crates/goose/src/config/base.rs
```diff
@@ -170,61 +251,136 @@ impl Config {
...
-            let file_content = std::fs::read_to_string(&self.config_path)?;
-            // Parse YAML into JSON Value for consistent internal representation
-            let yaml_value: serde_yaml::Value = serde_yaml::from_str(&file_content)?;
-            let json_value: Value = serde_json::to_value(yaml_value)?;
-
-            match json_value {
-                Value::Object(map) => Ok(map.into_iter().collect()),
-                _ => Ok(HashMap::new()),
+        match &self.config_storage {
+            ConfigStorage::File { path } => {
```
zakiali: config_path instead of path to keep consistent with previous implementation

crates/goose/src/config/base.rs
meenalc: hmm, actually the reason I put in here was for an optimization that may be client instead of sending all the tool calls with every request registers one config and keep it updated as more actions get added that is shared across threads in kotlin server but I will drop it for now and can bring it back if we decide to go that route

meenalc: same thing with secrets

meenalc: we discussed offline yes due to global config for now but we should get rid of it

crates/goose/src/config/base.rs
meenalc: updated

zakiali:
This looks mostly good, but I think you need to rebase to the latest changes since the configs for extensions are not being loaded correctly that have env_keys keys (e.g. slack)

meenalc:
> This looks mostly good, but I think you need to rebase to the latest changes since the configs for extensions are not being loaded correctly that have env_keys keys (e.g. slack)

done @zakiali

zakiali (approved):

michaelneale:
still need this @meenalc or can close it?

Issue #2069, opened 2025-04-07
Clicking on advanced settings menu then exit back to chat stops goose from running its current commands


zanesq:
**Describe the bug**

Start a long running chat where goose is performing a lot of steps, click into advanced settings then exit and the goose is working on it is gone and it appears to have stopped before finishing the tasks

**To Reproduce**
1. Start a long running chat where goose is performing a lot of steps
2. click into advanced settings from top right menu then exit back to chat
3. Notice goose stopped before finishing

**Expected behavior**
Goose should continue on with the tasks

**Please provide following information:**
 - **OS & Arch:** Mac M4 arm
 - **Interface:** UI
 - **Version:** 1.0.17
 - **Extensions enabled:** Developer
 - **Provider & Model:** claude 3.5

joecoufal-block-xyz:
+1. also if you’re typing a prompt and click into advanced settings before sending, the prompt disappears

taniacryptid:
+1 to this, exactly as Joe mentioned. On goose desktop, usually when I'm midway typing a followup for goose and I want to go into advanced settings (either to enable an extension, or check something I forgot about), I come back and all the text I was typing in the chat field is gone 😢  With no way to recover it since it was never sent. It feels like the text should persist, and would love an option for that however possible <3

Developerayo:
This just happened to me 🥲

jamadeo:
@zanesq I think this is fixed now? Seems to work OK for me

zanesq:
Yup! Closing

PR #2271, opened 2025-04-18
fix: clicking advanced settings stops running current command

opdich:
Disable menu items that would `setView()` while `isLoading === true`
Issue #2069


nahiyankhan:
woop! have heard grumblings about this experience. thanks for taking it on. 

comments - think the disabling feels unexpected to me. aka I think the session needs to continue irrespective if I navigate to the settings and back. however there maybe functional quirks I am not aware of. @lily-de any idea if we can let session continue (if goose is in the middle of answering) while we change settings? or would there be an unknown effect

zanesq:
Agree that it would be better to let it keep running in the background if possible. I'm not sure whats causing that restriction maybe @lily-de has some insight or I can dig into that more and see whats going on.

opdich:
> woop! have heard grumblings about this experience. thanks for taking it on.
> 
> comments - think the disabling feels unexpected to me. aka I think the session needs to continue irrespective if I navigate to the settings and back. however there maybe functional quirks I am not aware of. @lily-de any idea if we can let session continue (if goose is in the middle of answering) while we change settings? or would there be an unknown effect

From what I can tell the chat state is maintained by the response from `useMessageStream()` in `ChatView.tsx`. In order to support switching views the chat state would have to be moved out of `ChatView.tsx` and maintained elsewhere.

An alternative I looked into was opening the settings view as a new window. Unfortunately, all the settings variables are passed into each view as props, so existing windows won't get setting updates until they are reopened.

Lastly, I also considered converting the settings view into a modal, but a number of the options in the settings view already open a modal, so it seemed like a poor UX.

zanesq:
@opdich thanks that makes sense now. We are working on adding proper routing and global state management now actually so this could get a lot easier soon.

baxen:
Hey team, ahead of #2953 i'm trying to close out in progress PRs - this sounds like something we need to revisit in the future. Might be easier on v1.1? Going to close this out but please reopen if you'd like to get to it before that release

PR #2281, opened 2025-04-21
Add remove session capacity

Shawn5141:
Allow user to remove session by using
1) --all flag: remove all sessions
2) --session-name: remove speicifc session

### DIFF
--- a/crates/goose-cli/src/cli.rs
+++ b/crates/goose-cli/src/cli.rs
@@ -8,7 +8,7 @@ use crate::commands::configure::handle_configure;
 use crate::commands::info::handle_info;
 use crate::commands::mcp::run_server;
 use crate::commands::recipe::{handle_deeplink, handle_validate};
-use crate::commands::session::handle_session_list;
+use crate::commands::session::{handle_session_list, handle_session_remove};
 use crate::logging::setup_logging;
 use crate::recipe::load_recipe;
 use crate::session;
@@ -73,6 +73,18 @@ enum SessionCommand {
             default_value = "text"
         )]
         format: String,
+    },
+    #[command(about = "Remove sessions")]
+    Remove {
+        #[arg(short, long, help = "session id to be removed", default_value = "")]
+        id: String,
+        #[arg(
+            short,
+            long,
+            help = "regex for removing matched session",
+            default_value = ""
+        )]
+        regex: String,
     },
 }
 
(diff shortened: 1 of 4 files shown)


yingjiehe-xyz:
thanks for the contribution! Try to run `cargo fmt` to fix the formatting?

crates/goose-cli/src/cli.rs
```diff
@@ -74,6 +74,23 @@ enum SessionCommand {
...
+            default_value = "false"
+        )]
+        all: bool,
+        #[arg(
+            short,
+            long,
+            help = "session name to be removed",
+            default_value = ""
+        )]
+        session_name: String
```
yingjiehe-xyz: instead of removing single one per run, can we use regex to match a list of file name and prompt to ask users to confirm before removing them?

Shawn5141: Done

crates/goose-cli/src/commands/session.rs
```diff
@@ -1,5 +1,82 @@
...
 use goose::session::info::{get_session_info, SessionInfo};
+use regex::Regex;
+use std::fs;
+use std::io::{self, Write};
+
+pub fn remove_session(session: &SessionInfo) -> Result<()> {
+    println!(
+        "Are you sure you want to delete session `{}`? (yes/no):\n",
+        session.id
+    );
```
yingjiehe-xyz: can we use something like `cliclack::confirm`, examples"

```
cliclack::confirm("Do you want to clear message history & act on this plan?")
                        .initial_value(true)
                        .interact()?;
```

crates/goose-cli/src/cli.rs
```diff
@@ -74,6 +74,20 @@ enum SessionCommand {
...
+        all: bool,
+        #[arg(short, long, help = "session name to be removed", default_value = "")]
+        session_name: String,
+        #[arg(
+            short,
+            long,
+            help = "regex to remove matched session",
+            default_value = ""
+        )]
+        regex: String,
```
yingjiehe-xyz, line 87: since we have the regex, do we still need all or session_name? users can use `*` for all and the full session name with regex if they want to delete single session

Shawn5141: Done. Also update comment with example CLI included

yingjiehe-xyz (approved):

PR #2283, opened 2025-04-21
fix: extension config write locks

alexhancock:
* Adds file locking when writing to `config.yaml` for extensions configuration
* Disables the toggleable input for enabling/disabling an extension when the writes are in progress

### DIFF
--- a/crates/goose/src/config/base.rs
+++ b/crates/goose/src/config/base.rs
@@ -1,10 +1,13 @@
 use etcetera::{choose_app_strategy, AppStrategy, AppStrategyArgs};
+use fs2::FileExt;
 use keyring::Entry;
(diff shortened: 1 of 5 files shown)


baxen (approved):

crates/goose/src/config/base.rs
```diff
@@ -621,4 +641,171 @@ mod tests {
...
+                "Incorrect value for key {}",
+                key
+            );
+        }
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_concurrent_extension_writes() -> Result<(), ConfigError> {
```
meenalc, line 717: @alexhancock this is a flaky test, it was failing for me in main.

PR #2290, opened 2025-04-22
feat(bench): add configurable output directory for benchmarks

capttrousers:
- Add output_dir option to BenchRunConfig
- Modify BenchmarkWorkDir to use custom output path
- Improve error handling with proper context
- Set current directory as default output location, when running `goose bench init-config`, or when running `goose bench run` with a config that does not have an output_dir property (ie existing configs in the wild today)
- Add path validation to ensure absolute paths are used


laanak08 (approved):
LGTM! thanks for adding these, really helps the robustness.

PR #2301, opened 2025-04-22
fix syntax highlighting

zanesq:
before:
Screenshot 2025-04-22 at 11 06 03 AM

after:
Screenshot 2025-04-22 at 11 03 36 AM

### DIFF
--- a/ui/desktop/src/styles/main.css
+++ b/ui/desktop/src/styles/main.css
@@ -290,6 +290,7 @@
     padding: 1em;
   }
 
+  li > code.bg-inline-code,
   p > code.bg-inline-code {
     color: var(--block-orange);
     padding: 2px 4px;


ui/desktop/src/components/MarkdownContent.tsx
```diff
@@ -98,16 +102,7 @@ export default function MarkdownContent({ content, className = '' }: MarkdownCon
           ${className}`}
         components={{
           a: ({ ...props }) => <a {...props} target="_blank" rel="noopener noreferrer" />,
-          code({ className, children, inlinecode, ...props }) {
```
zanesq, line 101: `inlinecode` prop was removed in `ReactMarkdown` so this wasn't even running

nahiyankhan: ahh good to know. in a local branch, I have also noticed that the next v of ReactMarkdown moves around classNames to. fyi

PR #2303, opened 2025-04-22
Bottom and top bar refinement

zanesq:
New top and bottom bar organization

image


nahiyankhan:
I feel like I approving it is cheating 😄

nahiyankhan (approved):

PR #2308, opened 2025-04-23
feat: integrating summarization into agent reply 

brichu26:
Update 5/4: 
1. Added two new summarization modules:
- summarize_bp.rs: Implements bullet point style summarization that extracts key points in a structured format
- summarize_concise.rs: Implements concise summarization focusing on brevity while maintaining essential context

2. Modified necessary files to integrate these new strategies:
- Updated context_mgmt/mod.rs to export the new modules
- Added new methods to Agent in agents/context.rs:
    - summarize_context_bullet_points()
    - summarize_context_concise()

4/22: 
This PR adds support for different strategies to handle context length exceeded errors in Goose. As part of the project with GAIB, we were tasked to integrate a summarization pipeline into the reply method of crates/goose/src/agents/agent.rs:

Now, it should include: 
- Truncate (default): Removes oldest messages to fit within context limit
- Summarize: Summarizes older messages to preserve context
- Clear: Clears all messages except the last user message
- Manual Delete: Prompts user to manually delete specific messages

The changes include:
- New ContextResolutionStrategy enum
- Context strategy field in Agent struct
- Method to set the strategy
- Implementation of each strategy
- Integration with existing error handling


baxen:
Thanks for the contrib @brichu26 ! We landed another summarize tool does that handle what you were trying to do here? I'll close this out for now because i think it is covered but please feel free to reopen if it isn't

PR #2315, opened 2025-04-23
cli(ux): Show active context length in CLI

dbraduan:
Exposing # of tokens used in a last message(total_tokens = input_tokens + output_tokens) to the end user for UX improvement.

How this looks like: 
image


crates/goose-cli/src/session/mod.rs
```diff
@@ -842,6 +845,56 @@ impl Session {
...
+            .context_limit
+            .unwrap_or(32000); // Default to 32k if not specified
+
+        // Create a token counter using the same tokenizer as the model
+        let token_counter =
+            TokenCounter::new(self.agent.provider().get_model_config().tokenizer_name());
+
+        // Calculate system prompt tokens (approximate using a simpler system prompt)
+        // Since we can't directly access the built system prompt
+        let system_prompt = "You are an AI assistant.";
```
dbraduan: This is a hack to get it working for now, we should probably load context prompt similar to how we do it `truncate.rs`

crates/goose-cli/src/session/mod.rs
```diff
@@ -842,6 +845,56 @@ impl Session {
...
 
+    /// Calculate the current context window usage
+    pub async fn calculate_context_usage(&self) -> Result<(usize, usize)> {
+        // Get the model's context limit from the provider
+        let context_limit = self
+            .agent
+            .provider()
+            .get_model_config()
+            .context_limit
+            .unwrap_or(32000); // Default to 32k if not specified
```
dbraduan: Similar to `truncate.rs`, perhaps there could be one place where this is exposed, but don't have that much of knowledge about the codebase yet to tell where that should be.

dbraduan (commented):
self-review

crates/goose/src/model.rs
```diff
@@ -78,8 +78,12 @@ impl ModelConfig {
             name if name.contains("o1-mini") || name.contains("o1-preview") => Some(128_000),
             name if name.contains("o1") => Some(200_000),
             name if name.contains("o3-mini") => Some(200_000),
-            name if name.contains("gpt-4.1") => Some(1_000_000),
-            name if name.contains("gpt-4-1") => Some(1_000_000),
+            name if name.contains("gpt-4.1")
+                || name.contains("gpt-4-1")
+                || name.contains("gpt-41") =>
+            {
+                Some(1_000_000)
```
dbraduan: This might be just a thing for our internal proxies, let me know if that's not a good addition, can remove.

crates/goose-cli/src/session/mod.rs
```diff
@@ -867,6 +870,43 @@ impl Session {
...
+
+        // Get the exact system prompt and tools using prepare_tools_and_prompt
+        let (tools, toolshim_tools, system_prompt) = self.agent.prepare_tools_and_prompt().await?;
+
+        // Combine tools and toolshim_tools
+        let mut all_tools = tools.clone();
+        all_tools.extend(toolshim_tools.iter().cloned());
+
+        // Create a token counter using the same tokenizer as the model
+        let token_counter = TokenCounter::new(model_config.tokenizer_name());
```
baxen: the token counting i think can be somewhat slow and inaccurate (the anthropic ones are proprietary and we've seen our estimates be off by as much as 30%). Instead i think we can save the current token usage from the most recent model inference? that way we know more exactly and don't spend any extra time on it during the iteration loop

dbraduan: Addressed, @baxen!

angiejones:
hey @dbraduan - couple of merge conflicts to address

dbraduan:
should be good to go, @angiejones ?

PR #2319, opened 2025-04-23
feat: mcp router disable extensions

wendytang:
* Adds a disable extension capability for mcp router
* Rebases on agent refactor, also extension enabling now falls under tool permissions, which simplifies the code

--- 

**1.** goose will bring up that you have too many extensions enabled unprompted. As well, goose does not bring it up again if you decline.
image

--- 

**2.** repurposes `search_available_extensions` to also show extensions available to disable
image

--- 

**3.** disable occurs as tool calls, individually
image

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -288,6 +288,38 @@ impl ExtensionManager {
         Ok(())
     }
 
+    pub async fn suggest_disable_extensions_prompt(&self) -> Value {
+        let enabled_extensions_count = self.clients.len();
+
+        let total_tools = self
+            .get_prefixed_tools(None)
+            .await
+            .map(|tools| tools.len())
+            .unwrap_or(0);
+
+        // Check if either condition is met
+        const MIN_EXTENSIONS: usize = 5;
+        const MIN_TOOLS: usize = 50;
+
+        if enabled_extensions_count > MIN_EXTENSIONS || total_tools > MIN_TOOLS {
+            Value::String(format!(
+                "The user currently has enabled {} extensions with a total of {} tools. \
+                Since this exceeds the recommended limits ({} extensions or {} tools), \
+                you should ask the user if they would like to disable some extensions for this session.\n\n\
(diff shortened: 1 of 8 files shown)


crates/goose/src/agents/extension_manager.rs
```diff
@@ -288,6 +288,38 @@ impl ExtensionManager {
...
+
+        let total_tools = self
+            .get_prefixed_tools(None)
+            .await
+            .map(|tools| tools.len())
+            .unwrap_or(0);
+
+        // Check if either condition is met
+        const MIN_EXTENSIONS: usize = 5;
+        const MIN_TOOLS: usize = 50;
```
wendytang, line 302: tbd

crates/goose/src/agents/platform_tools.rs
```diff
@@ -102,7 +103,30 @@ pub fn enable_extension_tool() -> Tool {
...
+            title: Some("Enable an extension".to_string()),
+            read_only_hint: false,
+            destructive_hint: false,
+            idempotent_hint: false,
+            open_world_hint: false,
+        }),
+    )
+}
+
+pub fn disable_extension_tool() -> Tool {
```
ahau-square: looking at this again now the disable extension and enable extension tool are pretty much exactly the same, can we merge them into one tool like MANAGE_EXTENSIONS? should check that it doesn't degrade performance but would be nice if could reduce # tools

ahau-square (approved):
tried it out and works well! 

one small comment on perhaps merging the disable/enable extensions tool into one tool to reduce some redundancy there

one other thing I noticed is that when you first start up the Goose UI, not all extensions might be finished loading by the time you send the first message, so even if you exceed the limits, you won't get the goose suggestion to disable extensions until the next message. I think this is ok for now unless we hear a lot of complaints about it.

crates/goose/src/agents/platform_tools.rs
wendytang: it works pretty well with just one tool! updated

PR #2324, opened 2025-04-23
feat: bottom bar text truncation and tooltips

The-Best-Codes:
> [!NOTE]
> This PR is a draft for now until I get feedback on 3 points (see the end).

## Before

You can see an issue in the bottom bar when model names or the working directory are too long to fit in the UI:

Long model name:
image

Long directory name:
image

## After

To solve this, I updated the CSS so that the text gets truncated. You can hover the truncated text to reveal a tooltip containing the full text.
Demo video:

https://github.com/user-attachments/assets/a9378b9d-b989-4807-99fd-35b5486e1ad7

----

## Questions for reviewers

I need feedback! Please tell me what you think about:

* The “Goose Mode” indicator having a tooltip as well (for now, it doesn't have one because it can pretty well be inferred what the selected option is based on context)
* Edge cases where the tooltip is so wide that it takes up the whole viewport; how should it be handled? We could set a max width and height with a scroll overflow. Example edge case:
image

* Hiding tooltips when the text is not truncated; this can either be roughly estimated using CSS or more precisely estimated using JS (in React), both of which add utilities to the code for a trivial tweak.

### DIFF
--- a/ui/desktop/openapi.json
+++ b/ui/desktop/openapi.json
@@ -10,7 +10,7 @@
     "license": {
       "name": "Apache-2.0"
     },
-    "version": "1.0.19"
+    "version": "1.0.20"
   },
   "paths": {
     "/agent/tools": {
--- a/ui/desktop/src/components/bottom_menu/BottomMenu.tsx
+++ b/ui/desktop/src/components/bottom_menu/BottomMenu.tsx
@@ -152,20 +152,16 @@ export default function BottomMenu({
         }}
       >
         <Document className="mr-1" />
-        <span className="truncate max-w-[170px]">
-          <TooltipProvider>
-            <Tooltip>
-              <TooltipTrigger asChild>
-                <span className="truncate max-w-[170px] min-w-0 block">
-                  Working in {window.appConfig.get('GOOSE_WORKING_DIR')}
-                </span>
-              </TooltipTrigger>
-              <TooltipContent side="top">
-                {window.appConfig.get('GOOSE_WORKING_DIR')}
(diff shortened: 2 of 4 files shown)


zanesq:
Looks good! Regarding your questions:

> The “Goose Mode” indicator having a tooltip as well (for now, it doesn't have one because it can pretty well be inferred what the selected option is based on context)

I think its fine to not have a tooltip here

> Edge cases where the tooltip is so wide that it takes up the whole viewport; how should it be handled? We could set a max width and height with a scroll overflow. Example edge case:

For this case can we truncate it? I know that kinda goes against what we are trying to do here but it seems better than the alternatives. WDYT?

> Hiding tooltips when the text is not truncated; this can either be roughly estimated using CSS or more precisely estimated using JS (in React), both of which add utilities to the code for a trivial tweak.

Definitely agree we should only see them if needed. I think its fine to use JS for that.

The-Best-Codes:
@zanesq Thank you for the feedback!

> For this case can we truncate it? I know that kinda goes against what we are trying to do here but it seems better than the alternatives. WDYT?

Here is one alternative. What do you think?

image

> Definitely agree we should only see them if needed. I think its fine to use JS for that.

Will get to work on this!

The-Best-Codes:
Updated:

* The tooltip will now overflow if it is too wide:
image

* The tooltip will not appear whenever the text isn't truncated.

The-Best-Codes:
Merged the latest changes from goose/main. All tests were passing before (hopefully they are still passing afterward!).

Issue #2329, opened 2025-04-24
docs: Improve Onboarding experience


blackgirlbytes:
**Describe the bug**

User @devtooligan submitted this in Discussions..so I'm opening an issue to track 

Hi I spent about 45 minutes trying to onramp onto Goose but am running into lots of issues so I've given up for the moment.

Specifically, the docs do not reflect what I see in the app. For example, the docs keep telling me to click on the 3 dots in the upper right and then click on `settings`

But there is no `settings` option.  There is only `Advanced Settings`

<img width="1406" alt="Screenshot 2025-04-17 at 12 46 50 PM" src="https://github.com/user-attachments/assets/a6e0a80d-3a8a-43e0-9b28-41b90baa35d0" />

So I click on advanced settings.

Then the docs say go to Extension section and enable and click on Developer extension to enable it, but there is no Developer extension or ability to enable it.

<img width="623" alt="image" src="https://github.com/user-attachments/assets/f2cb742f-cb84-4fd4-ad39-b970da7f4819" />

<img width="1096" alt="image" src="https://github.com/user-attachments/assets/434f755d-b74b-4af0-b81c-0c9d8e6a0084" />

I tried clicking on "Browse extensions" and I see developer tool but there is no way to enable it and clicking on it does nothing:
<img width="695" alt="image" src="https://github.com/user-attachments/assets/6b5a1f72-dd1b-44e2-a28d-d83093010d24" />

So I tried +Add custom extension and I typed name "Developer" and tried to submit and it is complaining that "Command is required"

<img width="469" alt="image" src="https://github.com/user-attachments/assets/467bc44b-60c1-4173-9030-17f9c9d66cc3" />

I just really don't know what to do from here.

Also I keep getting this error message and have no idea what its talkign about:

<img width="406" alt="image" src="https://github.com/user-attachments/assets/761666c4-1e29-4468-bdef-33b102b0d9cd" />

What config.yaml? What is that referring to?

I heard about you and I thought that you are a general purpose agent that I can keep as my main driver (having voice would be great too) and it would select which model to ask based on my request, control my terminal, write code, control my browser, everything...but so far it's really hard to tell what Goose can actually do. 

How about just a general tutorial on some basic things you can do with Goose and how to do them? And also please make the docs look the same as the current version of the app.

does this project do that?

DOsinga:
I'm closing this since we have the new UI which makes this all so much clearer

PR #2330, opened 2025-04-24
docs: update from settings to advanced settings

blackgirlbytes:
User created a Discussion post informing us that the docs don't not matching the UI. Docs say settings. UI says Advanced Settings. https://github.com/block/goose/discussions/2251

I opened an issue to track it #2329


angiejones (approved):
#betterment 🙏🏽

PR #2334, opened 2025-04-24
feat: add/change model without reinit of extensions

alexhancock:
* Leverage the new `/agent/update_provider` endpoint to make it a simpler process when we add/change models
* Replace the call to `initializeSystem` in `changeModel` with a simpler call to `initializeAgent` which doesn't need to do anything with extensions
* Simplify signature of `changeModel` now that it doesn't need extension management methods passed through


ahau-square (approved):
nice!

PR #2346, opened 2025-04-24
feat: reduce the # of gdrive tools

yingjiehe-xyz:
We have two create tools and two update tool, one for google drive file, one for non-google files.

We can combine them into one to reduce the # of gdrive tools

### DIFF
--- a/crates/goose-mcp/src/google_drive/mod.rs
+++ b/crates/goose-mcp/src/google_drive/mod.rs
@@ -285,55 +285,10 @@ impl GoogleDriveRouter {
             }),
         );
 
-        let upload_tool = Tool::new(
-            "upload".to_string(),
-            indoc! {r#"
-                Upload a file to Google Drive.
-            "#}
-            .to_string(),
-            json!({
-              "type": "object",
-              "properties": {
-                  "name": {
-                      "type": "string",
-                      "description": "The desired filename to use for the uploaded file.",
-                  },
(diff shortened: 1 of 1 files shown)


crates/goose-mcp/src/google_drive/mod.rs
```diff
@@ -343,10 +298,9 @@ impl GoogleDriveRouter {
                       "type": "string",
                       "description": "Name of the file to create",
                   },
-                  "fileType": {
+                  "mimeType": {
                       "type": "string",
-                      "enum": ["document", "spreadsheet", "slides", "folder", "shortcut"],
-                      "description": "Type of Google file to create (document, spreadsheet, slides, folder, or shortcut)",
+                      "description": "The MIME type of the file.",
```
kalvinnchau, line 303: maybe we make this an enum of the the supported `mimeTypes` as well?

kalvinnchau (approved):
overall lgtm, one nit about the mime types

- in your testing did you notice if it ever got confused trying to use the tool? would it be worth having examples in the instructions on various cases?

crates/goose-mcp/src/google_drive/mod.rs
yingjiehe-xyz: but will that have kind of long list? In my testing, goose can choose the right mime type based on the right, like google doc:

image

PR #2357, opened 2025-04-25
Add interactive deletion of sessions

dbraduan:
Noticed there is remove command, which currently requires knowing the ids.

Made it interactive.

Video:

https://github.com/user-attachments/assets/b9c62f24-d7bc-4560-9a63-df3a5a0cd459

### DIFF
--- a/crates/goose-cli/src/commands/session.rs
+++ b/crates/goose-cli/src/commands/session.rs
@@ -1,5 +1,6 @@
 use anyhow::{Context, Result};
 use goose::session::info::{get_session_info, SessionInfo, SortOrder};
+use cliclack::multiselect;
 use regex::Regex;
 use std::fs;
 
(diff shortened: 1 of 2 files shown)


crates/goose-cli/src/commands/session.rs
```diff
@@ -27,8 +27,50 @@ pub fn remove_sessions(sessions: Vec<SessionInfo>) -> Result<()> {
...
-pub fn handle_session_remove(id: String, regex_string: String) -> Result<()> {
-    let sessions = match get_session_info(SortOrder::Descending) {
+fn prompt_interactive_session_selection(sessions: &[SessionInfo]) -> Result<Vec<SessionInfo>> {
+    if sessions.is_empty() {
+        println!("No sessions available to select.");
+        return Ok(vec![]);
+    }
+
+    let mut selector = multiselect(
+        "Select sessions to delete (use spacebar, Enter to confirm, Ctrl+C to cancel):",
```
hasnain-db, line 39: wording is confusing. if it's what I think you are saying, then perhaps this?
```suggestion
        "Select sessions to delete (use spacebar or Enter to confirm, Ctrl+C to cancel):",
```

crates/goose-cli/src/commands/session.rs
```diff
@@ -27,8 +27,50 @@ pub fn remove_sessions(sessions: Vec<SessionInfo>) -> Result<()> {
     Ok(())
 }
 
-pub fn handle_session_remove(id: String, regex_string: String) -> Result<()> {
-    let sessions = match get_session_info(SortOrder::Descending) {
+fn prompt_interactive_session_selection(sessions: &[SessionInfo]) -> Result<Vec<SessionInfo>> {
+    if sessions.is_empty() {
+        println!("No sessions available to select.");
```
hasnain-db:
```suggestion
        println!("No sessions to delete.");
```

crates/goose-cli/src/commands/session.rs
```diff
@@ -37,29 +79,37 @@ pub fn handle_session_remove(id: String, regex_string: String) -> Result<()> {
...
-                "Regex string '{}' does not match any sessions",
-                regex_string
-            );
+            println!("Regex string '{}' does not match any sessions", regex_val);
             return Ok(());
         }
     } else {
-        return Err(anyhow::anyhow!("Neither --regex nor --id flags provided."));
+        if all_sessions.is_empty() {
+            println!("No sessions found.");
```
hasnain-db: same comment as above.

crates/goose-cli/src/commands/session.rs
```diff
@@ -37,29 +79,37 @@ pub fn handle_session_remove(id: String, regex_string: String) -> Result<()> {
...
     let matched_sessions: Vec<SessionInfo>;
-    if !id.is_empty() {
-        if let Some(session) = sessions.iter().find(|s| s.id == id) {
+
+    if let Some(id_val) = id {
+        if let Some(session) = all_sessions.iter().find(|s| s.id == id_val) {
             matched_sessions = vec![session.clone()];
         } else {
-            return Err(anyhow::anyhow!("Session '{}' not found.", id));
+            println!("Session '{}' not found.", id_val);
```
hasnain-db: why'd we make this change? IMO it should still be an error?

crates/goose-cli/src/commands/session.rs
dbraduan: spacebar is to use multi-select, enter to confirm selection, ctrl-c to cancel

crates/goose-cli/src/commands/session.rs
```diff
@@ -27,8 +27,50 @@ pub fn remove_sessions(sessions: Vec<SessionInfo>) -> Result<()> {
...
+
+    let display_map: std::collections::HashMap<String, SessionInfo> = sessions
+        .iter()
+        .map(|s| {
+            let desc = if s.metadata.description.is_empty() {
+                "(no description)"
+            } else {
+                &s.metadata.description
+            };
+            let truncated_desc = if desc.len() > 60 {
```
yingjiehe-xyz: use a const to replace the magic number?

yingjiehe-xyz (approved):

crates/goose-cli/src/commands/session.rs
dbraduan: addressed, thanks!

dbraduan:
@yingjiehe-xyz - let me know if we can merge this :)

PR #2374, opened 2025-04-28
Goose v2 base frontend

zanesq:
Frontend stack for redesigned app. Includes running in web and electron. View [readme](https://github.com/block/goose/blob/goose-v2/ui-v2/README.md) for more info.

image

### DIFF
--- a/ui-v2/electron/main.ts
+++ b/ui-v2/electron/main.ts
@@ -105,34 +105,6 @@ async function createWindow() {
     }
   );
 
-  ipcMain.handle('open-file', async (): Promise<string> => {
-    const { filePaths } = await dialog.showOpenDialog({
-      properties: ['openFile'],
-    });
-
-    if (filePaths && filePaths[0]) {
-      const content = await fs.readFile(filePaths[0], 'utf8');
-      return content;
-    }
-    return '';
-  });
-
-  ipcMain.on('window-minimize', () => {
-    mainWindow.minimize();
-  });
-
-  ipcMain.on('window-maximize', () => {
-    if (mainWindow.isMaximized()) {
-      mainWindow.unmaximize();
-    } else {
-      mainWindow.maximize();
-    }
-  });
-
-  ipcMain.on('window-close', () => {
-    mainWindow.close();
-  });
-
   // Load the app
   if (isDevelopment) {
     mainWindow.loadURL('http://localhost:3001/');
--- a/ui-v2/README.md
+++ b/ui-v2/README.md
@@ -1,5 +1,5 @@
-# codename goose v2
-### A local, extensible, open source AI agent that automates engineering tasks
+# codename goose ui v2
+Your on-machine AI agent, automating tasks seamlessly.
 
 ## Development
 


ui-v2/electron/main.ts
```diff
@@ -0,0 +1,160 @@
...
+import type {
+  IpcMainInvokeEvent,
+  OnHeadersReceivedListenerDetails,
+  HeadersReceivedResponse,
+} from 'electron';
+
+const isDevelopment = !app.isPackaged;
+
+// Enable sandbox before app is ready
+app.enableSandbox();
```
alexhancock, line 14: Interesting! Hadn't seen this option before but I wonder if it gives us some more peace of mind about rendering untrusted content?

ui-v2/electron/main.ts
```diff
@@ -0,0 +1,160 @@
...
+      nodeIntegration: false,
+      contextIsolation: true,
+      webSecurity: true,
+      preload: preloadPath,
+      sandbox: true,
+    },
+  });
+
+  // Set up CSP
+  session.defaultSession.webRequest.onHeadersReceived(
```
alexhancock, line 36: What's the goal with this section?

alexhancock (approved):

ui-v2/electron/main.ts
zanesq: Its a [security best practice](https://www.electronjs.org/docs/latest/tutorial/sandbox) for sandboxing untrusted renderer code.

ui-v2/electron/main.ts
zanesq: Actually goose suggested this for security. I think we'll start restrictive and may need to adjust as we go. There is a whole list of security features for electron [here](https://www.electronjs.org/docs/latest/tutorial/security). I plan on going through them all soon.

This code's main purpose is to implement Content Security Policy (CSP) headers for the Electron application. Here's a detailed breakdown:

1. **Timing and Interception**:
   - It intercepts all incoming HTTP responses before they reach the renderer
   - Allows modification of response headers before content is processed
   - Acts as a middleware for all web requests in the application

2. **Content Security Policy Implementation**:
   - Sets up CSP rules to prevent various types of attacks like XSS
   - Defines different policies for development and production environments

3. **Development Mode CSP**:
   ```javascript
   default-src 'self';                                    // Only allow resources from same origin
   script-src 'self' 'unsafe-inline';                     // Allow inline scripts (needed for dev tools)
   style-src 'self' 'unsafe-inline';                     // Allow inline styles
   connect-src 'self' ws://localhost:3001 http://localhost:3001;  // Allow local dev server connections
   img-src 'self' data: https:;                          // Allow images from same origin, data URIs, and HTTPS
   font-src 'self' data:;                               // Allow fonts from same origin and data URIs
   ```

4. **Production Mode CSP** (more restrictive):
   ```javascript
   default-src 'self';              // Only allow resources from same origin
   script-src 'self';               // No inline scripts allowed
   style-src 'self' 'unsafe-inline'; // Allow inline styles (often needed for styling)
   img-src 'self' data: https:;     // Allow images from same origin, data URIs, and HTTPS
   font-src 'self' data:;          // Allow fonts from same origin and data URIs
   ```

5. **Security Benefits**:
   - **XSS Prevention**: Restricts sources of content that can be loaded
   - **Data Injection Protection**: Controls what types of data can be loaded
   - **Protocol Security**: Ensures resources are loaded over secure connections
   - **Resource Origin Control**: Limits where resources can be loaded from

6. **Development vs Production**:
   - Development mode is more permissive to allow for dev tools and hot reloading
   - Production mode is stricter for better security
   - Both modes maintain basic security while allowing necessary functionality

7. **Header Processing**:
   ```typescript
   .replace(/\s+/g, ' ')  // Removes extra whitespace
   .trim()                // Removes leading/trailing whitespace
   ```
   - Ensures the CSP header is properly formatted
   - Prevents issues with multiline strings

This code is crucial for:
1. Security: Preventing common web vulnerabilities
2. Resource Control: Managing what content can be loaded
3. Development Workflow: Enabling necessary dev features while maintaining security
4. Production Safety: Enforcing strict security in production builds

It's a key part of the application's security architecture, working alongside the sandbox and other security features to create a robust security model.

PR #2382, opened 2025-04-29
feat: gdrive download file

### LINKED ISSUE #2383: gdrive should allow file download
At the moment - it shows up as a bug, will try to download text/js/json or even binary files, and get into a loop of failure. 

Would ideally have a download tool, explicitly (and similarly to read/get, could convert app files)

michaelneale:
Allows simple download of files - including arbitrary types and binaries which were not supported before

fixes: https://github.com/block/goose/issues/2383

### DIFF
--- a/crates/goose-mcp/src/google_drive/mod.rs
+++ b/crates/goose-mcp/src/google_drive/mod.rs
@@ -1352,7 +1352,7 @@ impl GoogleDriveRouter {
             });
 
             // Call the download function
-            let download_result = self.download(download_params).await?;
+            let download_result = self._download(download_params).await?;
 
             // Return the download result along with the file content
             let mut result = self.read_file_content(&drive_uri, &include_images).await?;
@@ -2985,7 +2985,8 @@ impl GoogleDriveRouter {
         }
     }
 
-    async fn download(&self, params: Value) -> Result<Vec<Content>, ToolError> {
+    // Private helper method for downloading files
+    async fn _download(&self, params: Value) -> Result<Vec<Content>, ToolError> {
         let file_id =
             params
                 .get("fileId")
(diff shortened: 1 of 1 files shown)


sheagcraig (commented):
This is solving an obvious need! Currently, we only allow you to get the file contents in memory. I imagine Goose could be prompted carefully to save it locally, but that's a pain.

Since this new download function shares a lot with the `read_tool`, `export_google_file` and `get_google_file` functions, I wonder if it makes sense to add an optional argument to `read` to download locally; possibly as a side-effect (so in addition to returning the contents).

If not, I suspect extracting the common things into functions would help minimize the duplication.

michaelneale:
@sheagcraig PTAL again - good idea - is not just one tool

crates/goose-mcp/src/google_drive/mod.rs
```diff
@@ -1234,16 +1261,38 @@ impl GoogleDriveRouter {
...
-                    //TODO: handle base64 image case, see typscript mcp-gdrive
-                    Err(ToolError::ExecutionError(format!(
-                        "Suported mimeType {}, for {}",
-                        mime_type, uri,
-                    )))
+                    // For binary files or other non-text formats, provide a message
+                    Ok(vec![Content::text(format!(
+                        "This file has MIME type '{}' which cannot be displayed directly. Use the downloadTo parameter to download it.",
+                        mime_type
+                    ))])
```
kalvinnchau, line 1268: i don't think we should be returning a success in this case? i think an MCP error with this exact message would direct to model to retry with the `downloadTo` parameter

michaelneale:
@sheagcraig not sure if this is needed now, looks like some other changes landed

sheagcraig:
Oh yeah—I can see that the Google workspace and non functions were merged. That should probably handle this.

michaelneale:
not needed at the moment

PR #2386, opened 2025-04-29
fix: add sanitization for deeplink reading

zakiali:


DOsinga:
I think @alicehau created a fix yesterday for something similar? can we close this now?

DOsinga:
ok, closing assuming that is the case - can reopen if not

PR #2391, opened 2025-04-29
fix: update the remove_shims path

zakiali:


ui/desktop/src/components/settings_v2/extensions/utils.ts
```diff
@@ -166,26 +166,18 @@ export async function replaceWithShims(cmd: string) {
...
-  // Check if the command matches any shim pattern
-  const isShim = shimPatterns.some((pattern) => pattern.test(cmd));
-
-  if (isShim) {
-    const segments = cmd.split('/');
-    // Filter out any empty segments (which can happen with trailing slashes)
-    const nonEmptySegments = segments.filter((segment) => segment.length > 0);
-    // Return the last segment or empty string if there are no segments
-    return nonEmptySegments.length > 0 ? nonEmptySegments[nonEmptySegments.length - 1] : '';
+export async function removeShims(cmd: string) {
```
kalvinnchau, line 169: if this is now `async` we probably need to `await` where its called

kalvinnchau: if this is now `async` we'll need to `await` where it's called

kalvinnchau (changes requested):

ui/desktop/src/components/settings_v2/extensions/utils.ts
zakiali: yup... this is a long chain of async's/awaits

kalvinnchau: can we force this function to resolve the promise synchronously so we don't need to make changes across the board?

kalvinnchau (approved):

PR #2395, opened 2025-04-29
Lifei/pass param to goose recipe

lifeizhou-ap:
**Why**

Currently Goose recipe only contains static content. User is unable to specify different data to reuse the recipe. 

For example: I want to summarize the threads in a channel.   User cannot specify different channel id when they reuse the recipe.

**What**
- Support recipe template to enable dynamic contents in Goose cli 
- Added the param argument in the goose run command line to pass the dynamic parameters into recipe as a template

**Note**: At the moment, it does not apply to the Goose Desktop as more work is required on the UI to pass values to the recipe. However, if the deeplink is generated from the recipe (without params), it will ask the user to enter the param value in Desktop.  

**Usage**
`goose run --recipe recipe.yaml --params variable_name=value --interactive`
eg: `goose run --recipe recipe.yaml --params channel_name=goose-help --interactive`

**Example** 
- snippet of `recipe.yml`, {{channel_name}} is the dynamic parameter to be passed in
  ```
  version: 1.0.0
  title: Search through Slack channel history
  description: Search through Slack channel history
  instructions: |
    Search through Slack channel history for the channel {{channel_name}} to identify and summarize workflow use cases and patterns. 
    Focus on categorizing different types of workflows, their implementation details, and any specific tools or extensions required. 
  ```

- command
`goose run --recipe recipe.yaml --params channel_name=goose-help --interactive`


crates/goose-cli/src/cli.rs
```diff
@@ -10,7 +10,7 @@ use crate::commands::mcp::run_server;
 use crate::commands::recipe::{handle_deeplink, handle_validate};
 use crate::commands::session::{handle_session_list, handle_session_remove};
 use crate::logging::setup_logging;
-use crate::recipe::load_recipe;
+use crate::recipe::load_and_apply_recipe;
```
DOsinga: I dont know that I would rename here

crates/goose-cli/src/recipe.rs
```diff
@@ -68,3 +69,44 @@ pub fn load_recipe<P: AsRef<Path>>(path: P, log: bool) -> Result<Recipe> {
...
+        )),
+    }
+}
+
+pub fn load_and_apply_recipe<P: AsRef<Path>>(
+    path: P,
+    log: bool,
+    params: Vec<(String, String)>,
+) -> Result<Recipe> {
+    let recipe = load_recipe(&path, log)?;
```
DOsinga: So I think we should apply the template engine on the entire yaml file before we read it in as a yaml

You might want to do something like:

     goose run --recipe try_this --param temperature=0.3 
     
and then have it modify the temperature (once I merge my settings)

DOsinga (approved):

crates/goose-cli/Cargo.toml
```diff
@@ -52,6 +52,7 @@ shlex = "1.3.0"
 async-trait = "0.1.86"
 base64 = "0.22.1"
 regex = "1.11.1"
+tera = "1.19.1"
```
kalvinnchau: we already have `minijinja` as a templating engine https://github.com/block/goose/blob/main/crates/goose/Cargo.toml#L40, is there a reason to prefer `tera` here? Would prefer to keep using the same templating dependency across all crates

crates/goose-cli/src/cli.rs
```diff
@@ -271,6 +279,16 @@ enum Command {
...
 
+        #[arg(
+            long,
+            value_name = "KEY=VALUE",
+            help = "Dynamic parameters (e.g., --param username=alice)",
+            long_help = "Key-value parameters to pass to the recipe or instruction file. Can be specified multiple times.",
+            action = clap::ArgAction::Append,
+            value_parser = parse_key_val,
+        )]
+        param: Vec<(String, String)>,
```
kalvinnchau:
```suggestion
        params: Vec<(String, String)>,
```
lets make this internal variable plural to be consistent with other variables (extensions, bulitins, etc)
we can keep the flag singular

crates/goose-cli/src/cli.rs
```diff
@@ -59,6 +59,14 @@ fn extract_identifier(identifier: Identifier) -> session::Identifier {
     }
 }
 
+fn parse_key_val(s: &str) -> Result<(String, String), String> {
+    let parts: Vec<&str> = s.splitn(2, '=').collect();
+    if parts.len() != 2 {
+        return Err(format!("invalid KEY=VALUE: {}", s));
+    }
+    Ok((parts[0].to_string(), parts[1].to_string()))
+}
```
kalvinnchau: could simplify with `split_once` here
```suggestion
    match s.split_once('=') {
        Some((k, v)) => Ok((k.to_string(), v.to_string())),
        None => Err(format!("invalid KEY=VALUE: no `=` in `{}`", s)),
    }
```

crates/goose-cli/src/recipe.rs
```diff
@@ -68,3 +79,26 @@ pub fn load_recipe<P: AsRef<Path>>(path: P, log: bool) -> Result<Recipe> {
...
+    params: &[(String, String)],
+) -> Result<String> {
+    // Turn params into HashMap
+    let param_map: HashMap<String, String> = params.iter().cloned().collect();
+
+    // Create a Tera context
+    let mut context: TeraContext = TeraContext::new();
+    for (key, value) in &param_map {
+        context.insert(key, value);
+    }
```
kalvinnchau: don't think we need to create the hashmap if we're just directly inserting it into context immediately, you can just iterate over the params slice directly

kalvinnchau (changes requested):
This allows templating every field of the yaml file, is that intentional or should we restrict the templating to specific fields (instructions / prompt)?

DOsinga:
> This allows templating every field of the yaml file, is that intentional or should we restrict the templating to specific fields (instructions / prompt)?

I think we want fairly brought support for templating yeah. instructions, prompt, settings, mcp servers. We could do it on a one by one basis, but I don't see much of a downside to making the entire thing a template?

kalvinnchau (approved):

PR #2400, opened 2025-04-30
feat: support configurable way to  retrieve recipes via github

lifeizhou-ap:
**Why**
Retrieve recipe via name from local directories and github, so that user can use shared recipe in github

**What**
Before this change, the user can only pass the recipe file path.  With this change, the logic will be

- If the user passes the recipe file path (eg: ~/Downloads/recipe.yaml),  it will render the file from the path
- If the user passes the recipe name (my-recipe), it will
      - search the my-recipe.yaml or my-recipe.json in the current directory
      - if "GOOSE_RECIPE_GITHUB_REPO" is set up in config.yaml (eg: `GOOSE_RECIPE_GITHUB_REPO=squareup/goose-recipes), search the recipe.yaml o recipe.json in the github.com/squareup/goose-recipes/my-recipe directory via git clie
  
This logic applies to `goose run --recipe`, `goose recipe deeplink` `goose recipe validate` cli

- Added the options to set GOOSE_RECIPE_GITHUB_REPO to Goose settings in `goose configuration` cli


recipe.yaml
```diff
@@ -0,0 +1,39 @@
+version: 1.0.0
+title: Custom recipe from chat
+description: a custom recipe instance from this chat session
+instructions: Fetch and analyze news content from online sources, providing structured summaries of headlines and key stories. Present the information in a clear, hierarchical format using Markdown, with main stories categorized by topic and including brief summaries of each major story. When the content is paginated or truncated, make multiple fetch requests to ensure comprehensive coverage. Focus on presenting the most newsworthy and current information first, followed by supporting details and analysis.
```
DOsinga: this is cool, but probably not somethign you want to merge :)

crates/goose-cli/src/recipes/github_recipe.rs
```diff
@@ -0,0 +1,78 @@
...
+
+pub async fn download_github_recipe(recipe_name: &str, target_dir: &Path) -> Result<PathBuf> {
+    // Try both yaml and json extensions
+    for ext in &["yaml", "json"] {
+        let filename = format!("{}.{}", recipe_name, ext);
+        let target_path = target_dir.join(&filename);
+
+        // GitHub API URL for the recipe file
+        let api_url = format!(
+            "https://api.github.com/repos/squareup/goose-recipes/contents/{}/recipe.{}?ref=douwe/joke-of-the-day",
```
DOsinga: the ref is probably a left over?

DOsinga (commented):
This is great. I think we should just start with shelling out to git though, just do a git clone, in a hidden folder and pull from there from time to time

crates/goose-cli/src/recipes/github_recipe.rs
lifeizhou-ap: yes. it was just for testing.

kalvinnchau (changes requested):
We don't want any internal specific logic built in the the OSS app, that includes the git url and any other environment variables.

Automatically searching the user's various directories and then searching remote endpoints is unexpected behavior for a flag like that and shouldn't be automatically done. 

Imo when a user passes in a `--recipe` the expectation is to use the file and not go searching around.
If we want to support automatically searching, I think it should be a separate, explicit `--search` flag, even then the expectations are unclear around what/where it looks. 
It should be well defined to the user that it will go looking around their computer; and imo it shouldn't automatically go searching for remote files (or split it into `--search-remote $URL`).

If we want remote support it should be generic - i.e. `--url`  that supports various types of endpoints `https://` or `git://` not specific to any org.

DOsinga:
So me and @kalvinnchau had a quick chat about this, here's our rough consensus:

* We can't make changes that impact the open source version in a weird way like adding a fixed github repository
* We do want to make recipes discoverable

So let's do this:
* Remove the ~/goose_recipes search for now. We might bring this back if useful
* Let's add an optional setting in goose settings that specifies the github repository to search in
* Let's add this to the install script and then later to sq goose (currently pointing to the python version, ugh)


i.e. we now have:

    curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh \
    | GOOSE_PROVIDER=databricks GOOSE_MODEL=goose \
    DATABRICKS_HOST=https://block-lakehouse-production.cloud.databricks.com/ bash

change that to:

    curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh \
    | GOOSE_PROVIDER=databricks GOOSE_MODEL=goose \
    RECIPE_GIT=square_up/goose-recipes \
    DATABRICKS_HOST=https://block-lakehouse-production.cloud.databricks.com/ bash

lifeizhou-ap:
> So me and @kalvinnchau had a quick chat about this, here's our rough consensus:
> 
> * We can't make changes that impact the open source version in a weird way like adding a fixed github repository
> * We do want to make recipes discoverable
> 
> So let's do this:
> 
> * Remove the ~/goose_recipes search for now. We might bring this back if useful
> * Let's add an optional setting in goose settings that specifies the github repository to search in
> * Let's add this to the install script and then later to sq goose (currently pointing to the python version, ugh)
> 
> i.e. we now have:
> 
> ```
> curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh \
> | GOOSE_PROVIDER=databricks GOOSE_MODEL=goose \
> DATABRICKS_HOST=https://block-lakehouse-production.cloud.databricks.com/ bash
> ```
> 
> change that to:
> 
> ```
> curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh \
> | GOOSE_PROVIDER=databricks GOOSE_MODEL=goose \
> RECIPE_GIT=square_up/goose-recipes \
> DATABRICKS_HOST=https://block-lakehouse-production.cloud.databricks.com/ bash
> ```

Hey team,

I have updated this PR.  

- Added the input for user to enter the recipe github name and it will saved as GOOSE_RECIPE_GITHUB_REPO in config.yaml file.  If the user passes the GOOSE_RECIPE_GITHUB_REPO environment variable, the env variable value will be displayed the default value in the input prompt.  If this environment variable is not passed, it will show the current value in the config.yaml file as the default value if it has been configured.

- I changed to use gh clone instead of git clone, because using gh I can pass the repo full name (org_name/repo_name) instead of the ssh clone url (org-49461806@github.com:squareup/goose-recipes.git) which is a bit customised.  

- only check the recipe with the specified path, or current directory on local machine. If GOOSE_RECIPE_GITHUB_REPO is specified in the config.yaml, it will fetch the recipe in the github repo

crates/goose-cli/Cargo.toml
```diff
@@ -63,3 +63,4 @@ tempfile = "3"
 temp-env = { version = "0.3.6", features = ["async_closure"] }
 test-case = "3.3"
 tokio = { version = "1.43", features = ["rt", "macros"] }
+
```
kalvinnchau: don't need this extra newline

crates/goose-cli/src/recipes/github_recipe.rs
```diff
@@ -0,0 +1,131 @@
+use anyhow::Result;
+use std::env;
+use std::path::Path;
+use std::path::PathBuf;
+use std::process::Command;
+
+// const GOOSE_RECIPE_REPO_FULL_NAME: &str = "squareup/goose-recipes";
```
kalvinnchau: lets remove this entirely

crates/goose-cli/src/commands/configure.rs
```diff
@@ -1104,3 +1112,33 @@ pub async fn configure_tool_permissions_dialog() -> Result<(), Box<dyn Error>> {
...
+    match std::env::var(key_name) {
+        Ok(recipe_repo) => {
+            default_recipe_repo = Some(recipe_repo);
+        }
+        _ => {
+            if let Ok(Some(recipe_repo)) = config.get_param(key_name) {
+                default_recipe_repo = Some(recipe_repo);
+            }
+        }
+    }
```
kalvinnchau: this could be simplified to:
```suggestion
    let default_recipe_repo: Option<String> = std::env::var(key_name)
        .ok()
        .or_else(|| config.get_param(key_name).unwrap_or(None));
```

crates/goose-cli/src/commands/configure.rs
```diff
@@ -1104,3 +1112,33 @@ pub async fn configure_tool_permissions_dialog() -> Result<(), Box<dyn Error>> {
...
+    let mut recipe_repo_input = cliclack::input(
+        "Enter the Goose Recipe Github Repo in the format of owner/repo: eg squareup/goose-recipes",
```
kalvinnchau:
```suggestion
        "Enter the Goose Recipe Github repo (owner/repo): e.g. block/goose-recipes",
```
maybe we pick/use a `block/` as an example for oss

crates/goose-cli/src/commands/configure.rs
```diff
@@ -1104,3 +1112,33 @@ pub async fn configure_tool_permissions_dialog() -> Result<(), Box<dyn Error>> {
...
+    )
+    .required(false);
+    if let Some(recipe_repo) = default_recipe_repo {
+        recipe_repo_input = recipe_repo_input.default_input(&recipe_repo);
+    }
+    let input_value: String = recipe_repo_input.interact()?;
+    if input_value.clone().trim().is_empty() {
```
kalvinnchau, line 1133: add a comment that `None`/empty string clears this setting

crates/goose-cli/src/recipes/github_recipe.rs
kalvinnchau, line 1: think we need a check that `gh` is installed before any of these commands are run

crates/goose-cli/src/recipes/github_recipe.rs
```diff
@@ -0,0 +1,131 @@
...
+            .current_dir(local_repo_parent_path.clone())
+            .status()
+            .map_err(|_: std::io::Error| anyhow::anyhow!(error_message.clone()))?;
+
+        if status.success() {
+            Ok(local_repo_path)
+        } else {
+            Err(anyhow::anyhow!(error_message))
+        }
+    }
```
kalvinnchau, line 115: we can simplify and reduce nesting with something like:
```rust
    // If it’s already a git repo, we’re done
    if local.join(".git").is_dir() {
        return Ok(local);
    }
    
    // Otherwise make sure parent dir exists
    std::fs::create_dir_all(&tmp)
        .with_context(|| format!("Could not create temp dir `{}`", tmp.display()))?;
        
    let status = Command::new("gh")
        .args(["repo", "clone", recipe_repo_full_name])
        .current_dir(local_repo_parent_path.clone())
        .status()
        .map_err(|_: std::io::Error| anyhow::anyhow!(error_message.clone()))?;
        
    if status.success() {
        Ok(local_repo_path)
    } else {
        Err(anyhow::anyhow!(error_message))
    }
```

crates/goose-cli/src/recipes/github_recipe.rs
```diff
@@ -0,0 +1,131 @@
...
+        if !login_status.success() {
+            Err(anyhow::anyhow!("Failed to authenticate using GitHub CLI."))
+        } else {
+            Ok(())
+        }
+    }
+}
+
+fn ensure_repo_cloned(recipe_repo_full_name: &str) -> Result<PathBuf> {
+    let local_repo_parent_path = env::temp_dir();
```
kalvinnchau, line 90: does this create a new temp_dir on every run of goose? is that intentional for now or should it keep track across runs so it just needs to fetch latest?

crates/goose-cli/src/recipes/github_recipe.rs
```diff
@@ -0,0 +1,131 @@
...
+            .args(["auth", "login"])
+            .status()
+            .expect("failed to run `gh auth login`");
+
+        if !login_status.success() {
+            Err(anyhow::anyhow!("Failed to authenticate using GitHub CLI."))
+        } else {
+            Ok(())
+        }
+    }
```
kalvinnchau: can we reduce nesting here with:
```rust
if status.success() {
  return Ok(())
}}
// rest of the else block
```

crates/goose-cli/src/recipes/search_recipe.rs
```diff
@@ -0,0 +1,64 @@
...
+    match read_recipe_in_dir(&current_dir, recipe_name) {
+        Ok(content) => Ok(content),
+        Err(e) => {
+            if let Some(recipe_repo_full_name) = configured_github_recipe_repo() {
+                retrieve_recipe_from_github(recipe_name, &recipe_repo_full_name)
+            } else {
+                Err(e)
+            }
+        }
+    }
```
kalvinnchau:
```suggestion
read_recipe_in_dir(&current_dir, recipe_name)
    .or_else(|err| {
        if let Some(repo_full) = configured_github_recipe_repo() {
            retrieve_recipe_from_github(recipe_name, &repo_full)
        } else {
            Err(err)
        }
    })
```

kalvinnchau (changes requested):
* we'll need to document somewhere that `gh` is a required dependency for using the github recipe feature
  * maybe as a note in the configure step? 
* can we make `GOOSE_RECIPE_GITHUB_REPO` a `const` in `crates/goose-cli/src/recipes/github_recipe.rs` and re-used in the various places its hard-coded at the moment
* there's a couple uses of `.expect(..)` can we use a `map_err` approach and print the resulting error rather than having it panic

crates/goose-cli/src/recipes/github_recipe.rs
lifeizhou-ap: this is temp directory based on the platform https://doc.rust-lang.org/std/env/fn.temp_dir.html for example: macos: /tmp. 

User has to clone the repo at least once before git fetch. We use temp directory as the parent dir when the user clones the repo.  The temp dir should exist already so that we don't create additional directory.  In case it does not exist, we add the code to create the dir

kalvinnchau: Ah yes, sorry I confused it with the `tempfile::tempdir` which we use in other places which deletes the directory after the tempdir binding is out of scope.

crates/goose-cli/src/recipes/github_recipe.rs
lifeizhou-ap: agree. I added the extra message

.map_err(|_| anyhow::anyhow!("Failed to run `gh auth status`. Make sure you have `gh` installed."))?;

crates/goose-cli/src/commands/configure.rs
lifeizhou-ap: It might confuse people if we use `block/goose-recipes` because this repo does not exist.  Currently `goose-recipes` is only in `squareup` and probably some of the recipes are not good to expose to a public repo. 

I guess in the future we might have `block/goose-recipes` in the open source community.  For now. I may change 

`Enter the Goose Recipe Github Repo in the format of owner/repo: eg squareup/goose-recipes` =>  `Enter your Goose Recipe Github repo (owner/repo): e.g. squareup/goose-recipes`

How about that?

crates/goose-cli/src/commands/configure.rs
lifeizhou-ap: added the comment

kalvinnchau (approved):

Issue #1143, opened 2025-02-07
Global hotkey


Blucknote:
It would be nice to have customizable global hotkey to bring Goose window

PR #2401, opened 2025-04-30
feat: add global hotkey

opdich:
Add a static global hotkey to focus the goose window. The hotkey is set to `Cmd+Option+G`. Adding a settings feature to allow hotkey customization requires significant refactoring.

Issue #1143


opdich:
- Updated the keybinding to `Cmd+Option+Shift+G` which is less likely to be used elsewhere
- Added focus for the input field
- Added the keybinding as a File menu option to help communicate its availability
Screenshot 2025-04-30 at 1 42 56 PM

PR #2403, opened 2025-04-30
Don't require prompt in headless mode for recipes

DOsinga:
Prompt is optional in recipes, which makes sense for the UI version, but that doesn't really work in headless mode. So instead here we use the instructions as the first user message instead of adding it the system prompt, which effectively makes prompt optional also for headles

### DIFF
--- a/crates/goose-cli/src/cli.rs
+++ b/crates/goose-cli/src/cli.rs
@@ -483,10 +483,18 @@ pub async fn cli() -> Result<()> {
                         eprintln!("{}: {}", console::style("Error").red().bold(), err);
                         std::process::exit(1);
                     });
+
+                    let (contents, additional_system_prompt) = match recipe.prompt {
+                        Some(ref p) if !p.trim().is_empty() => {
+                            (Some(p.clone()), Some(recipe.instructions.clone()))
+                        }
+                        _ => (Some(recipe.instructions.clone()), None),
+                    };
+
                     InputConfig {
-                        contents: recipe.prompt,
+                        contents,
                         extensions_override: recipe.extensions,
-                        additional_system_prompt: Some(recipe.instructions),
+                        additional_system_prompt,
                     }
                 }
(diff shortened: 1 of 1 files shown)


kalvinnchau (commented):
>instructions as the first user message instead of adding it the system prompt


A concern I have here is that the `instructions` that go into the `additional_system_prompt` are treated differently than `user` messages by most models.
My understanding is that `system_prompts` are weighted higher than `user` messages so this change could result in unexpected behavior changes by the models.

`instructions` would be for defining the model's behavior with additional rules, roles, tone and base knowledge

then `prompt` for the action for that model to take

kalvinnchau:
We could make `instructions` optional and then check at runtime that either `instructions` or `prompt` is defined and run with that

DOsinga:
Yeah, I had been thinking about that too. But this is for headless recipes, where there is no conversation. The problem here I think is that instructions is required and prompt is not. If you could just specify the prompt and not instructions, I would agree with you and that seems to be the default scenario for headless recipes - no need to modify the system prompt just tell the system what to do. So we could make instructions optional as an alternative.

I don't think this here results in a change of behavior mostly since currently it just crashes if you don't provide a prompt

DOsinga:
> We could make `instructions` optional and then check at runtime that either `instructions` or `prompt` is defined and run with that

That works for me, but that's a bigger change to the current contract

kalvinnchau:
> > We could make `instructions` optional and then check at runtime that either `instructions` or `prompt` is defined and run with that
> 
> That works for me, but that's a bigger change to the current contract

@zakiali and I were chatting and agree that's okay. It is a bigger change, but it makes sense to have that capability since not all use cases would want to change the system prompt and instead have a repeatable recipe with just prompt used.

DOsinga:
Ok, well, that makes the chance simpler

kalvinnchau (approved):
lgtm!

PR #2410, opened 2025-05-01
docs: Put cli install homebrew command in quickstart.md

deverman:
When I installed Goose for the command line I wanted to install it via homebrew but found it was not documented. Future users will now see the command.


blackgirlbytes:
Hi thank you! we now have the instructions on install via Homebrew, so closing for now!

PR #2207, opened 2025-04-15
feat: present options to user when context length is exceeded

salman1993:
Allow users to select clear/truncation/summarization options once we hit context exceed error.

The summarization approach is:
1. Break down large text into smaller chunks (roughly 33% of the model’s context window).
2. For each chunk:
    a. Combine it with the previous summary (or leave blank for the first iteration).
    b. Summarize the combined text, focusing on extracting only the information we need.
3. Generate a final summary using a tailored prompt.

Test:

```
What would you like to work on today? I can help you explore the project structure, make changes to code, or assist with any specific tasks you have in mind.
( O)> tell me about apple inc
◇  The model's context length is maxed out. You will need to reduce the # msgs. Do you want to?
│  Summarize Session 
│
final_summary: [Message { role: User, created: 1745430876, content: [Text(TextContent { text: "# Conversation Summary\n\nThe conversation begins with a user greeting, to which I respond by introducing myself as Goose, an AI assistant developed by Block. I explain my capabilities including helping with coding tasks, running shell commands, editing text files, and capturing screenshots.\n\nThe user then asks \"what can you do,\" prompting me to provide a more detailed explanation of my capabilities:\n- Developer tools (shell commands, file editing, screenshots, image processing)\n- Project-specific help with a Rust project and Electron app\n- Common tasks I can assist with (code navigation, making changes, testing, debugging)\n- Additional extensions that can be enabled as needed\n\nThe conversation ends with the user asking me to tell them about Apple Inc., but my response to this query is not included in the conversation history provided.", annotations: None })] }]

Context maxed out
--------------------------------------------------
Goose summarized messages for you. .



I'd be happy to tell you about Apple Inc. Let me gather some information for you.
─── shell | developer ──────────────────────────
command: echo "Searching for Apple Inc. information..."



I apologize, but I'm currently in chat-only mode and don't have access to use tools to search for up-to-date information. However, I can share what I know about Apple Inc. from my training data.

## Apple Inc.
```

PR #2415, opened 2025-05-01
ctx-mgmt: ctx session management (dev mode only)

lily-de:
Note: only uses summarization strategy in electron preview (`just run-ui`). Current app will continue to use truncation

Changes:
* added openapi schemas for session management
* added routes for session management
* when CLE event happens, after the summary thread is created we generate a new session ID to that thread
* Messages from before the CLE get grouped as `ancestorMessages` which will be shown but not sent to LLM (for that session -- they will disappear when resuming the session)
* When a user goes to resume a session where CLE event happened, they will see two separate sessions (currently no visual indicator that a session began as a summary spin off from another)

To test: 

1. add this https://github.com/block/goose/pull/2207/commits/9d6fceff7fb20213dc8f4acbd40553e312003ed9
2. run `just run-ui`

Example:

* run into multiple CLE events
* resumed a session:
   - that was already at max capacity
   - that was started from a summarized older thread

https://github.com/user-attachments/assets/cc355c17-ad7f-4f4f-9b34-77a329b8f3ef


zanesq:
Code LGTM and verified locally.

Some UI feedback: its not clear to me as a user that

1. I hit some limit, what is the purpose of this message in my chat. Maybe just some explainer text like you hit a limit we can summarize it and start over in the same chat with the summary.
2. That its continuing with the current chat after saving. I guess I expected a new chat to open up with the summary. Maybe we need a message after saving in the chat window saying they summary saved you can continue with the chat now.

PR #2420, opened 2025-05-01
Fix copy link button visibility in dark mode for share agent

### LINKED ISSUE #2418: GUI: copy link on share agent doesn't work
If you make a session into an agent, you get a long URL to copy, but copy link doesn't seem to work (at least in dark mode doesn't look like it works) - somewhere in ui/desktop

"Copy this link to share with friends or paste directly in Chrome to open" - the copy link there doens't seem to do anyting.

michaelneale:
Fixes #2418

## Description
This PR fixes the issue with the copy link button not being visible in dark mode when sharing an agent. The button text and icon were using text-iconSubtle and text-textSubtle classes which made them difficult to see in dark mode.

## Changes
- Updated the copy button in SessionHistoryView.tsx to use text-textStandard hover:text-textProminent for better visibility
- Updated the copy button in RecipeEditor.tsx to use similar styling
- Made the "Copied!" check icon green in both components for better visual feedback

## Testing
- Tested in dark mode to ensure the copy button is now visible
- Verified that the copy functionality works correctly

### DIFF
--- a/ui/desktop/src/components/RecipeEditor.tsx
+++ b/ui/desktop/src/components/RecipeEditor.tsx
@@ -171,15 +171,47 @@ export default function RecipeEditor({ config }: RecipeEditorProps) {
   const deeplink = generateDeepLink(getCurrentConfig());
 
   const handleCopy = () => {
-    navigator.clipboard
-      .writeText(deeplink)
-      .then(() => {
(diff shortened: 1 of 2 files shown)


michaelneale:
.bundle

PR #2422, opened 2025-05-02
docs: Remove "agent" subcommand which does not exist.

omo:
Apparently it was removed at #2091.


kalvinnchau (approved):
yes, good catch, thank you!

PR #2424, opened 2025-05-02
load all previous messages if --resume is passed

dbraduan:
Small improvement to the UX of CLI `--resume` command.

If the resume flag is passed, render all previous messages, so a user can understand where the agent left off in the previous conversation.

image

### DIFF
--- a/crates/goose-cli/src/cli.rs
+++ b/crates/goose-cli/src/cli.rs
@@ -446,6 +446,12 @@ pub async fn cli() -> Result<()> {
                         session.session_file().file_stem().and_then(|s| s.to_str()),
                         None,
                     )?;
+
+                    // Render previous messages if resuming a session
+                    if resume {
+                        session.render_message_history();
+                    }
+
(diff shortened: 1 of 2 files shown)


kalvinnchau:
Thanks for this contribution!

I'm thinking this should be a separate flag for this feature, and instead we keep the default behavior not to render previous messages. 
Maybe `goose session --resume --display`? or `--history`

I typically don't want to display my previous messages. I'll have `goose session --name feature-x` or something and can just resume from there and wouldn't want my history.

dbraduan:
Hey @kalvinnchau - agree, let's keep this backwards compatible. I will hide this under the flag.

Typically, I do know that some CLIs do `--no-interactive` and default to interactive mode, but provided there is an existing behavior that user base already got used to, let's not break it.

dbraduan:
@kalvinnchau - addressed the comment

kalvinnchau (approved):
lgtm! thank you for the contribution!

PR #2428, opened 2025-05-02
Global/local chat history store in localstorage

zanesq:
Restores the ability for cmd + arrow to bring up history from other chats. The previous [implementation](https://github.com/block/goose/pull/1442/files) used session storage. I changed this one to use localstorage so we can persist it across launches.

- Prioritizes current chat messages if a chat session in progress or restored from session.
- Empty chat prioritizes global messages.
- Has a limit `500` messages and timestamp `30 days` cleanup mechanism so localstorage doesn't get too large (we can adjust this later if needed)

NOTE: Currently having a strange issue locally where the localstorage value is being cleared out after app restart for me but not @kalvinnchau so I figure better to get this in for now to have global history in the currently launched app.


ui/desktop/src/components/ChatInput.tsx
```diff
@@ -49,17 +52,20 @@ export default function Input({
...
-    let joinedPaths =
-      displayValue.trim() +
-      ' ' +
-      droppedFiles.filter((path) => !displayValue.includes(path)).join(' ');
-    setDisplayValue(joinedPaths);
-    setValue(joinedPaths);
-    textAreaRef.current?.focus();
-    setProcessedFilePaths(droppedFiles);
-  }
+  useEffect(() => {
```
zanesq, line 62: Unrelated: Improved drag and drop to not execute with all renders and added some more string safety checks while I was in there

ui/desktop/package.json
```diff
@@ -30,8 +30,7 @@
     "format": "prettier --write \"src/**/*.{ts,tsx,css,json}\"",
     "format:check": "prettier --check \"src/**/*.{ts,tsx,css,json}\"",
     "prepare": "cd ../.. && husky install",
-    "start-alpha-gui": "ALPHA=true npm run start-gui",
-    "start-alpha-server": "cd ../.. && just run-ui-alpha"
```
zanesq, line 34: removed this wasn't doing what I thought

kalvinnchau:
.bundle

kalvinnchau (approved):

PR #2434, opened 2025-05-04
fix: use less strict timestamp in system prompt

### LINKED ISSUE #2431: v1.0.18 Breaks Anthropic Prompt Caching
**Describe the bug**

Anthropic's prompt caching significantly reduces overall inference costs by discounting cache hits by 90%. Up through v1.0.17, much of each subsequent prompt would be read from cache. As of v1.0.18, no more than 1,864 tokens will be read.

**To Reproduce**
1. Checkout v1.0.17, build, run a small prompt
2. Check costs in the Anthropic console
3. Checkout v1.0.18, build, run a small prompt
4. Check costs in the Anthropic console
5. Note that costs are ~4x for the second run

**Expected behavior**
They should be the same

**Screenshots**
Logs attached

 - **OS & Arch:** MacOS 15.3.1 (24D70)
 - **Interface:** CLI
 - **Version:** 1.0.18
 - **Extensions enabled:** developer, memory
 - **Provider & Model:** Anthropic / Claude 3-5 Sonnet Latest

yingjiehe-xyz:
Fix #2431

We have the current [timestamp](https://github.com/block/goose/blob/8ba40bdccc3bc5b1068394ea52f45c3de773ec35/crates/goose/src/prompts/system.md?plain=1#L3) in our system prompt, the timestamp was the real timestamp of the reply, this would lead the prompt cache invalidation as our system prompt has such dynamic content. Since the prompt cache only lasts 5 minutes, we can use the timestamp with hours to avoid the frequent cache invalidation

### DIFF
--- a/crates/goose-llm/src/completion.rs
+++ b/crates/goose-llm/src/completion.rs
@@ -61,7 +61,7 @@ fn construct_system_prompt(system_preamble: &str, extensions: &[Extension]) -> S
     );
     context.insert("extensions", serde_json::to_value(extensions).unwrap());
 
-    let current_date_time = Utc::now().format("%Y-%m-%d %H:%M:%S").to_string();
+    let current_date_time = Utc::now().format("%Y-%m-%d").to_string();
     context.insert("current_date_time", Value::String(current_date_time));
 
     prompt_template::render_global_file("system.md", &context).expect("Prompt should render")
--- a/crates/goose-llm/src/completion.rs
+++ b/crates/goose-llm/src/completion.rs
@@ -61,7 +61,7 @@ fn construct_system_prompt(system_preamble: &str, extensions: &[Extension]) -> S
     );
     context.insert("extensions", serde_json::to_value(extensions).unwrap());
 
-    let current_date_time = Utc::now().format("%Y-%m-%d").to_string();
+    let current_date_time = Utc::now().format("%Y-%m-%d %H:%M:%S").to_string();
     context.insert("current_date_time", Value::String(current_date_time));
 
     prompt_template::render_global_file("system.md", &context).expect("Prompt should render")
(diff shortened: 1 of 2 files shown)


crates/goose-llm/src/completion.rs
```diff
@@ -61,7 +61,7 @@ fn construct_system_prompt(system_preamble: &str, extensions: &[Extension]) -> S
     );
     context.insert("extensions", serde_json::to_value(extensions).unwrap());
 
-    let current_date_time = Utc::now().format("%Y-%m-%d %H:%M:%S").to_string();
+    let current_date_time = Utc::now().format("%Y-%m-%d").to_string();
```
ynniv: A somewhat blunt solution to the problem, but it will work. Could we keep `%H` for time of day though?

yingjiehe-xyz: yes, would like to try this quick fix first. I didn't add %H as considering the boundary case, like 8:59 ~ 9:01. Let me discuss with the team member who added the timestamp long time ago

salman1993 (approved):

salman1993:
some context on why the current date time was added. the LLM is not good at doing web search without the current date. 

when I tested the Tavily tool few months back: 

❌ which tech company CEOs attended trump's inaugration?
reason: gets the answer wrong - mentions 2017 stuff

✅ which tech company CEOs attended trump's 2025 inaugration? use web search

❌ what's the latest news on the LA wildfires? summarize in less than 200 words and cite sources
reason: ""I apologize, but I notice that all the news articles I'm finding are dated in January 2025, which is in the future."""

^ both these answers are fixed by adding the date in system prompt. most providers also do this: https://docs.anthropic.com/en/release-notes/system-prompts#nov-22nd-2024

PR #2440, opened 2025-05-05
cxt-mgmt: only two retries for summarization and move to prod

lily-de:
Handles case where summarization fails -- user can retry fetching summary twice and after that user will be prompted to start a new session 

Errors in summarization in my testing have only been caused by giant tool responses -- because these tool responses do not get summarized we can end up with a summary that is still outside the context window for the LLM

Screenshot 2025-05-05 at 7 20 58 PM

**NOTE: This PR also moves summarization to production**

### DIFF
--- a/ui/desktop/src/components/ChatView.tsx
+++ b/ui/desktop/src/components/ChatView.tsx
@@ -524,9 +524,8 @@ function ChatContent({
                     <UserMessage message={message} />
                   ) : (
                     <>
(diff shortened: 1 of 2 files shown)


ahau-square (approved):
small point - for summarization failure, should we say something a bit more explicit? Like we attempted to summarize but the session exceeds our summarization capacity, you must start a new session

The message about "extension data often takes up significant space" is a bit confusing to me

PR #2444, opened 2025-05-06
feat: specify parameters configuration in recipe file

lifeizhou-ap:
**Why**
- [Jira Ticket](https://block.atlassian.net/browse/DX-787)
- Specify the parameters in the recipe file. In this way, users and recipe authors have clear view of the input of the recipe.
- cli will prompt the user to enter missing params (with `use_prompt` requirement) in the non-interactive or interactive mode. This provides better user experiences

**What**
- Moved the recipe.rs into the recipes folder
- Added parameters section in the Recipe yaml file
- Validate the recipe file The validation is used in `goose run --recipe`, `goose recipe validate` and `goose recipe deeplink`. Rules:
    - If the recipe requirement is `optional`, it requires `default` field setup
    - compare the template variables in the yaml file and the parameters in the recipe file, and they should match
    
- When rendering the recipe file with user input params, i checks the parameters in the command line. Rules:
   - For the required param, user has to provide the param in the command line, otherwise it will return error
   - For the use_prompt param, user can provide the param in the command line. If they do not add in the command line, they still have a chance to provide in the prompt
   - For the optional param, user can provide the param in the command line. If they do not add in the command line, we will use the default value in the parameter configuration
- print out the parameter value used when running the recipe

**Screenshot**
- "user_prompt" param is not passed in the command line, and Goose prompts user to enter the value
Screenshot 2025-05-07 at 5 18 11 pm

- "required" param is not passed in the command line, and it returns an error message
Screenshot 2025-05-07 at 5 24 34 pm


lily-de (approved):

crates/goose-cli/src/recipes/recipe.rs
```diff
@@ -0,0 +1,468 @@
...
+
+    let mut params_for_template: Option<HashMap<String, String>> = None;
+    let rendered_content = match params {
+        None => recipe_file_content,
+        Some(user_params) => {
+            params_for_template =
+                Some(apply_values_to_parameters(&user_params, recipe_parameters)?);
+            render_content_with_params(&recipe_file_content, params_for_template.as_ref().unwrap())?
+        }
+    };
```
kalvinnchau: We can avoid the mutable hashmap + `unwrap()` later with an `if let Some(...)`
```suggestion
let (rendered_content, params_for_template) = if let Some(user_params) = params {
    let map = apply_values_to_parameters(&user_params, recipe_parameters)?;
    let rendered = render_content_with_params(&recipe_file_content, &map)?;
    (rendered, Some(map))
} else {
    (recipe_file_content, None)
};
```

crates/goose-cli/src/recipes/recipe.rs
```diff
@@ -0,0 +1,468 @@
...
+    // Create a minijinja environment and context
+    let mut env = minijinja::Environment::new();
+    env.set_undefined_behavior(UndefinedBehavior::Strict);
+    let template: Template<'_, '_> = env.template_from_str(content)
+        .map_err(|_| anyhow::anyhow!("Failed to render recipe, please check if the recipe has proper syntax for variables: eg: {{ variable_name }}"))?;
+
+    // Render the template with the parameters
+    template.render(params).map_err(|_| {
+        anyhow::anyhow!(
+            "Failed to render the recipe - please check if all required parameters are provided"
```
kalvinnchau: can you capture the error in `map_err(|e|...` and add the error from jinja to the anyhow message

crates/goose-cli/src/recipes/recipe.rs
```diff
@@ -0,0 +1,468 @@
...
+    if !extra_keys.is_empty() {
+        message.push_str(&format!(
+            "\nUnnecessary parameter definitions: {}.",
+            extra_keys
+                .iter()
+                .map(|s| s.to_string())
+                .collect::<Vec<_>>()
+                .join(", ")
+        ));
+    }
```
kalvinnchau: nit: can you add some line breaks between some statements so its easier to read

crates/goose-cli/src/recipes/recipe.rs
```diff
@@ -0,0 +1,468 @@
...
+    recipe_parameters: Vec<RecipeParameter>,
+) -> Result<HashMap<String, String>> {
+    let mut param_map: HashMap<String, String> = user_params.iter().cloned().collect();
+    let mut missing_params: Vec<String> = Vec::new();
+    for param in recipe_parameters {
+        if !param_map.contains_key(&param.key) {
+            match (&param.default, &param.requirement) {
+                (Some(default), _) => param_map.insert(param.key.clone(), default.clone()),
+                (None, RecipeParameterRequirement::UserPrompt) => {
+                    let input_value = cliclack::input(format!(
```
kalvinnchau, line 185: >cli will prompt the user to enter missing params (with use_prompt requirement) in the non-interactive or interactive mode. This provides better user experiences

what's the `use_prompt` requirement? I don't see that anywhere, is it intended to be a field in the `RecipeParameter` struct?

i wouldn't expect a prompt in a non-interactive mode either

lifeizhou-ap: For the `use_prompt` param, user can provide the param in the command line. If they do not add in the command line, they still have a chance to provide in the prompt. 

There were a couple of times that the users forgot to append the params in the command line and the command returns error immediately.  With this, it will help them have a better experience. 

Hi @DOsinga would you please check whether my above understanding on this card requirement is correct? Thank you!

DOsinga: yeah: for reference the ticket shared before that this is in reference to:

https://block.atlassian.net/browse/DX-787

when you have a parameter I see three ways for the user to provide the value:

* Using the --params flag. That's the best solution if you do this in a scirpt
* Ask the user. That's reasonable if you don't know before hand
* Provide a default value

the schema here allows for a dynamic approach to this that allows mixing and matching. for example you could have a recipe that creates a pr description for a given branch. The script that runs the the recipe has some way of finding out what the branch is, so it can pass it on as a parameter. We might want to ask the user for extra information though so we set the the extra_info to user_prompt. And the model to run this on should default to 4o, but we want to keep things open in case somebody wants to override this, so we set a default value for it.

Does that sound reasonable?

kalvinnchau: in the non-interactive scenario, where a pipeline pulls in a recipe that comes from git (using the `GOOSE_RECIPE_GITHUB_REPO` setting) 

wouldn't any update that adds a user prompt (with no default) cause the entire pipeline to block, waiting for input (or until it times out)?

there's no guarantee that everyone writing recipes will be aware of all users to say that that'll never happen

DOsinga: If you write "software" that others rely on there's always ways to break things. There are a bunch of things we can do if something like this actually becomes a problem (add a way to run a goose recipe at a specific hash, add a timeout for how long the user has to enter information, have a --script parameter that forces all user input into required off the top of my head) and some of these sound perfectly reasonably but I would say one thing at the time and let just get this in and see if it works for people

kalvinnchau: yeah that's fine, as long as we're thinking about this i'll go ahead and approve the PR

kalvinnchau (approved):

PR #2445, opened 2025-05-06
allowlist blocks and shift SSE to warning

michaelneale:
SSE extensions just get a warning notice (as risk profile is different)
MCP stdio will now block if an allowlist is enabled.

### DIFF
--- a/ui/desktop/src/App.tsx
+++ b/ui/desktop/src/App.tsx
@@ -499,7 +499,13 @@ export default function App() {
   }, [view]);
 
   // Configuration for extension security
-  const STRICT_ALLOWLIST = true; // Set to false to revert to warning-only mode
+  const config = window.electron.getConfig();
+  // If GOOSE_ALLOWLIST_WARNING is true, use warning-only mode (STRICT_ALLOWLIST=false)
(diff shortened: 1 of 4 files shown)


ui/desktop/src/App.tsx
```diff
@@ -498,7 +498,9 @@ export default function App() {
     }
   }, [view]);
 
-  // TODO: modify
+  // Configuration for extension security
+  const STRICT_ALLOWLIST = true; // Set to false to revert to warning-only mode
```
shellz-n-stuff: It would be awesome to expose this as a configuration option so other users/orgs that use this feature can choose between the super-secure blocking or the light-weight warnings.

shellz-n-stuff (approved):
Really clean! Looks really good but left a minor suggestion for improvement (not as much for our use-case but potentially helpful for other users)

ui/desktop/src/App.tsx
michaelneale: yeah good idea - for now its based on GOOSE_ALLOWLIST being set or not

alexhancock (approved):

PR #2451, opened 2025-05-06
[goose-llm] add completion request & error

salman1993:


crates/goose-llm/src/prompts/system.md
```diff
@@ -1,6 +1,6 @@
 {{system_preamble}}
 
-The current date is {{current_date_time}}.
+The current date is {{current_date}}.
```
michaelneale, line 3: does it include time still?

salman1993: nope cause the time invalidates cache so costs more $$$

baxen (approved):

PR #2457, opened 2025-05-07
Expanded ToolCall options

opdich:
# Description
This pull request introduces a new "Response Styles" feature to customize how tool responses are displayed. By default they are displayed in a "Detailed" style, but users can now choose a "Concise" style that shows only the tool name.
Screenshot 2025-05-06 at 11 04 37 PM
Screenshot 2025-05-06 at 11 03 14 PM

## New Feature: Response Styles
* Added `ResponseStylesSection` to settings, allowing users to choose between "Detailed" and "Concise" response styles. The selection is saved in `localStorage` and dynamically updates the UI. 
* Integrated the new "Response Styles" section into the `SettingsView` component. 

## Refactoring: State Management for Tool Responses
* Added logic to determine expansion behavior based on the selected response style. 
* Renamed and updated state variables in `ToolCallExpandable` and related components for clarity (e.g., `defaultExpanded` → `isStartExpanded`, `forceExpand` → `isForceExpand`). 
* Updated `ToolDetailsView` and `ToolResultView` components to consume the new `isStartExpanded` property set by `localstorage`. 

## UI Updates
* Adjusted `SessionSharingSection` layout for consistency with other settings sections. 
* Updated `GooseMessage` to properly show multiple consecutive tool calls. 
Screenshot 2025-05-06 at 11 06 40 PM

PR #2458, opened 2025-05-07
PoC: make tool calling more compact in GUI

michaelneale:
GUI currently shows args, params and so on with tool calls and uses a lot of space (even with output folded away) which is visually prominent and noisy. 

This is a proposal to fold more away by default: 

Folded: 

Screenshot 2025-05-07 at 1 57 20 pm

Expanded: 

Screenshot 2025-05-07 at 1 57 32 pm

originally: 

image


spencrmartin:
Screenshot 2025-05-07 at 11 37 29 AM
@michaelneale this might help pull the styling in a bit

michaelneale:
moving to https://github.com/block/goose/pull/2457

PR #2465, opened 2025-05-07
Remove outdated truncation of tool descriptions in OpenAI and Databricks providers

tiborvass:
Previously, tool descriptions were truncated to 1024 characters to conform to an assumed OpenAI limit. This truncation is now removed to avoid silent behavior changes and make debugging easier. OpenAI no longer appears to enforce this limit, so it's better to allow full descriptions and surface any real errors explicitly.

### DIFF
--- a/crates/goose/src/providers/formats/databricks.rs
+++ b/crates/goose/src/providers/formats/databricks.rs
@@ -246,15 +246,12 @@ pub fn format_tools(tools: &[Tool]) -> anyhow::Result<Vec<Value>> {
             return Err(anyhow!("Duplicate tool name: {}", tool.name));
         }
 
-        let mut description = tool.description.clone();
-        description.truncate(1024);
-
-        // OpenAI's tool description max str len is 1024
         result.push(json!({
             "type": "function",
             "function": {
                 "name": tool.name,
-                "description": description,
+                // do not silently truncate description
+                "description": tool.description,
                 "parameters": tool.input_schema,
             }
         }));
--- a/crates/goose/src/providers/formats/openai.rs
+++ b/crates/goose/src/providers/formats/openai.rs
@@ -202,15 +202,12 @@ pub fn format_tools(tools: &[Tool]) -> anyhow::Result<Vec<Value>> {
             return Err(anyhow!("Duplicate tool name: {}", tool.name));
         }
 
-        let mut description = tool.description.clone();
-        description.truncate(1024);
-
-        // OpenAI's tool description max str len is 1024
         result.push(json!({
             "type": "function",
             "function": {
                 "name": tool.name,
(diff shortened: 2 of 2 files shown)


taniacryptid:
Thank you for the PR! Surfaced to the team for review!

angiejones (approved):

shykes:
Thank you @tiborvass ! This has been a source of issues for us while integrating Dagger as a MCP server for Goose.

🙏

crates/goose/src/providers/formats/databricks.rs
```diff
@@ -246,15 +246,12 @@ pub fn format_tools(tools: &[Tool]) -> anyhow::Result<Vec<Value>> {
...
-        description.truncate(1024);
-
-        // OpenAI's tool description max str len is 1024
         result.push(json!({
             "type": "function",
             "function": {
                 "name": tool.name,
-                "description": description,
+                // do not silently truncate description
+                "description": tool.description,
```
michaelneale, line 254: I think could treat databricks seperately as we don't know if all models support that beyond openai via model serving on databricks

michaelneale (changes requested):
thanks for this - I think ok to remove for openai but for databricks, I think it is per model

michaelneale:
I want to try it with databricks and claude 3.4/3.7 today

tiborvass:
@michaelneale It just appeared to me that the Databricks provider was copy/pasted from OpenAI hence the change in Databricks too, but happy to make the PR specific to OpenAI. Let me know what your experimentation shows.

crates/goose/src/providers/formats/databricks.rs
tiborvass: The point of this PR is also to say that it is pretty much never a good idea to silently truncate as it yields very subtlely wrong behavior that is really annoying to debug. So if other models via databricks do not support more than 1024, then I think it is more desirable for users to see the error.

Or for each provider, detect if the problem came from a too long context, and provide a nicer/actionable error message to the user.

michaelneale:
yeah I think safe to remove for openai by the sounds of it - databricks we may need some additional logic to case it by model

michaelneale:
FWIW I tested with claude 3.5/3.7, bedrock via databricks, all ok - used "list windows" tool with an incredibly long tool description, and worked ok. Did not test openai directly

michaelneale:
yeah this seems good - at least for these providers so I think this is good to go now, will need more testing before more generally available (I wonder if shoudl just have simple logging for when it is over by a lot... or would that get annoying?)

michaelneale (approved):

angiejones:
thanks so much for the contribution!

PR #2466, opened 2025-05-07
Goose FFI lib

zhong-sq:
Goose-FFI lib on top of goose-llm crate to adapt to the new `completion` method


salman1993:
i think we can close this PR?

PR #2473, opened 2025-05-08
fix: will now correctly follow along with tool results

michaelneale:
Currently goose won't always follow along once you add a subsequent command (this happens with compact or expanded tools)

This PR also adds short tool descriptions that were missing when in compact mode

Screenshot 2025-05-08 at 12 57 29 pm

dark mode: 

image

### DIFF
--- a/ui/desktop/src/components/ui/scroll-area.tsx
+++ b/ui/desktop/src/components/ui/scroll-area.tsx
@@ -32,6 +32,7 @@ const ScrollArea = React.forwardRef<ScrollAreaHandle, ScrollAreaProps>(
           block: 'end',
           inline: 'nearest',
         });
+        // When explicitly scrolling to bottom, reset the following state
         setIsFollowing(true);
       }
     }, []);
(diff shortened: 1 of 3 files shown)


angiejones (approved):

ui/desktop/src/components/ui/scroll-area.tsx
```diff
@@ -66,13 +67,11 @@ const ScrollArea = React.forwardRef<ScrollAreaHandle, ScrollAreaProps>(
       const { scrollHeight, scrollTop, clientHeight } = viewport;
 
       const scrollBottom = scrollTop + clientHeight;
-      const newIsFollowing = scrollHeight === scrollBottom;
+      // Allow a small tolerance (2px) for rounding errors
+      const isAtBottom = scrollHeight - scrollBottom <= 2;
```
michaelneale, line 71: this seemed to help with getting it to follow along, or else it didn't realise was at the bottom and thus should follow along

PR #2479, opened 2025-05-08
ctx_management: summarize on command button

lily-de:
This allows users to summarize their conversations on command. Uses similar flow as CLE events

1. User clicks button
2. This button click inserts a 'summarizationRequested' message into the message list 
3. The rest of the flow is the same as how we handle CLE events, I just added an additional check for the `summarizationRequested` message type in addition to the `contextLengthExceededMessage` type. If either of those conditions are met, then we go the same summarization process
4. I modified the text around the summarization (see video) to differentiate it from the CLE events 
5. 
https://github.com/user-attachments/assets/cf842bf5-864e-4e4f-ba99-b811f34f11b2


laanak08 (approved):
LGTM! nice work.

PR #2483, opened 2025-05-08
[goose-llm] use chrono utc timestamp milliseconds for msg created

salman1993:


zhong-sq (approved):
thanks for fixing it!!

PR #2484, opened 2025-05-08
Move token limits to backend

zanesq:
Follow up for https://github.com/block/goose/pull/2471 to move the token limits to the backend rather than hardcoding in the front end.

NOTE: The bottom menu component is getting large, we'll be moving all this alert related logic out when we refactor alerts to be a global state mechanism.


crates/goose/src/model.rs
```diff
@@ -1,11 +1,39 @@
...
 
 const DEFAULT_CONTEXT_LIMIT: usize = 128_000;
 
 // Tokenizer names, used to infer from model name
 pub const GPT_4O_TOKENIZER: &str = "Xenova--gpt-4o";
 pub const CLAUDE_TOKENIZER: &str = "Xenova--claude-tokenizer";
 
+// Define the model limits as a static HashMap for reuse
+static MODEL_SPECIFIC_LIMITS: Lazy<HashMap<&'static str, usize>> = Lazy::new(|| {
+    let mut map = HashMap::new();
```
kalvinnchau, line 13: do we wanna keep the links for easy reference ?

crates/goose-server/src/routes/configs.rs
```diff
@@ -154,15 +155,23 @@ pub async fn get_config(
...
-    let config = Config::global();
-    let value = if let Ok(config_value) = config.get_param::<String>(&query.key) {
-        Some(config_value)
-    } else {
-        std::env::var(&query.key).ok()
+    let value = match query.key.as_str() {
+        // Special handling for model-limits
+        "model-limits" => {
+            let limits = ModelConfig::get_all_model_limits();
+            Some(serde_json::to_string(&limits).unwrap_or_default())
```
kalvinnchau: Above we throw an 500 if we can't get valid json, but here we return an empty object `{}`, think it should be consistent either way?

kalvinnchau (commented):
Is it meant to be in both the read_config and get_config routes in both files? Kinda confused why both endpoints support that

I feel like this might be better off under a separate endpoint rather than a query-param approach, and can be consolidated into a single endpoint

zanesq:
@kalvinnchau ah good callout, `configs` was only used in one place on the frontend when `not settingsV2Enabled` which is enabled and live now so I went ahead and refactored that to use the new way and removed the old configs endpoint and related code. I'm going to follow up next with a PR to remove all the `if settingsV2Enabled` code in the front end.

kalvinnchau (approved):
oh nice, didn't realize configs.rs was legacy only!

alexhancock (approved):

PR #2488, opened 2025-05-08
feat: add Snowflake provider 

sfc-gh-twhite:
## Overview

~~I am looking to add Snowflake as a provider for Goose. This pull request is intended to
get feedback early in the process, and I am open to closing and reopening this PR when
it's closer to completion. (if preferred)~~

I borrowed a lot of the code from the Anthropic provider, which made the
implementation easier, but it isn't 1:1. I've been using claude-3-5-sonnet for [CoT](https://docs.snowflake.com/en/user-guide/snowflake-cortex/cortex-llm-rest-api#tool-calling-with-chain-of-thought-example).

### Current Status

~~Most of the functionality works for the first message, but there is an issue in 
preparing the response to the message with the tool calls. I'm manually adding the
"tool_use" field in the post function, and I will adjust this to ensure the response is more
accurate and dynamic.~~

This is all working now, I ended up manually parsing out and mapping the tool calls. It might
be possible to use [eventsource-client](https://crates.io/crates/eventsource-client) but I was unable get it working properly. 

### What's Next

- ~~Adjust the preparation of the response to include "tool_use" based on the dynamically
interaction with the tool.~~
- ~~Perform thorough testing to ensure all edge cases are handled. (I need to get the
tests working)~~
- ~~Validate the integration with Snowflake and ensure seamless operation within the
Goose framework.~~


crates/goose/src/providers/openai.rs
```diff
@@ -152,6 +152,7 @@ impl Provider for OpenAiProvider {
     ) -> Result<(Message, ProviderUsage), ProviderError> {
         let payload = create_request(&self.model, system, messages, tools, &ImageFormat::OpenAi)?;
 
+        println!("payload to string: {:#?}", payload.clone().to_string());
```
sfc-gh-twhite:
```suggestion
```

crates/goose/src/providers/formats/openai.rs
```diff
@@ -447,6 +447,9 @@ pub fn create_request(
             .unwrap()
             .insert(key.to_string(), json!(tokens));
     }
+
+    println!("payload: {:?}", payload.to_string());
+
```
sfc-gh-twhite:
```suggestion
```

PR #2490, opened 2025-05-09
fix: menu bar and dock icon settings

### LINKED ISSUE #1464: No way to disable the menu bar icon
macOS doesn't handle a ton of icons in your menu bar well, especially on a laptop's screen. We already have 4 Block specific menu bar icons that we can't get rid of. I don't want Goose to add another, when it has a dock icon, so also having the menu bar icon is just duplicating functionality.

A setting where you could disable showing the menu bar icon

A setting to toggle between showing the dock icon or the menu bar icon but not both at the same time (common pattern for many resident macOS apps).

opdich:
# Description
This PR introduces a new "App Settings" section in the settings UI to manage visibility options for the menu bar and dock icons. fix #1464

## Frontend Changes:

* **New `AppSettingsSection` Component**:
  - Added a new React component, `AppSettingsSection`, to manage menu bar and dock icon visibility settings. It includes toggles for enabling/disabling these icons and ensures accessibility by enforcing at least one icon to remain visible. 
  - Integrated the `AppSettingsSection` into the `SettingsView` component. 

## Backend Changes:

* **Menu Bar and Dock Icon Management**:
  - Added IPC handlers in the main process to manage menu bar and dock icon visibility (`set-menu-bar-icon`, `get-menu-bar-icon-state`, `set-dock-icon`, `get-dock-icon-state`). These handlers update the settings and adjust the tray or dock visibility accordingly. 
  - Introduced `destroyTray` to safely remove the tray icon when toggled off. 

* **Electron API Extensions**:
  - Extended the `ElectronAPI` to include methods for setting and retrieving menu bar and dock icon states. 

* **Settings Schema Update**:
  - Added `showMenuBarIcon` and `showDockIcon` properties to the `Settings` interface with default values set to `true`.

### DIFF
--- a/ui/desktop/src/components/settings_v2/SettingsView.tsx
+++ b/ui/desktop/src/components/settings_v2/SettingsView.tsx
@@ -6,6 +6,7 @@ import ModelsSection from './models/ModelsSection';
 import { ModeSection } from './mode/ModeSection';
 import SessionSharingSection from './sessions/SessionSharingSection';
 import { ResponseStylesSection } from './response_styles/ResponseStylesSection';
+import AppSettingsSection from './app/AppSettingsSection';
 import { ExtensionConfig } from '../../api';
 import MoreMenuLayout from '../more_menu/MoreMenuLayout';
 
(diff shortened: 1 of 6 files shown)


opdich:
I added a disabled state to prevent rapid toggling, which seems to solve the issues with it getting stuck in the dock even when toggled off. I tried to tie the state to the `show()` promise, but that doesn't seem to work. I suspect there's something else happening async that isn't captured in that promise. So I just added a too long timeout and that works.

PR #2491, opened 2025-05-09
[goose-llm] add providerConfig param for exposed LLM functions

salman1993:
will merge after this one - https://github.com/block/goose/pull/2478 

requested by goose service team:
- can we update the completion method to take host & api_key as parameters, because we load the api_key at runtime in the service
- can we make the databricks staging/prod model name consistent, because we hardcoded the model we’re using for tooltip and session_name method, we want to use the same crate for staging/prod


zhong-sq (approved):
Looks good to me!

PR #2496, opened 2025-05-09
goose_llm update

zhong-sq:
`CompletionRequest` is supposed to be a data class in Kotline (structure in Rust), use `JsonValueFfi` in rust code to allow use add `uniffi::Record`


salman1993:
i think we can close this PR?

PR #2498, opened 2025-05-09
Remove settingsV2Enabled checks

zanesq:
Removed all logic surrounding if settingsV2Enabled since its live now and other cleanup of unused functions

I wonder if we can remove the `ui/desktop/src/components/settings` directory also, or combine whats left with `ui/desktop/src/components/settings_v2` into a single settings directory.

follow up for https://github.com/block/goose/pull/2484


kalvinnchau (approved):
lgtm!

alexhancock (approved):

PR #2503, opened 2025-05-10
feat: goose cli to track directories, allow to resume projects in location

michaelneale:
this adds a simple `goose project(s)` command which will offer to take you back to either most recent session, and dir and resume, or project a list of project dirs you have worked on (and offer to resume sessions). 

for those who have a mess of terminals/UIs all over and never forget to name sessions. 

how it works: 

* saves dirs/sessions/last comands/times to `~/.local/share/goose/projects.json` on each interaction with goose cli
* saves you from remembering what you were up to, and where, and what session
* lets you start a new session but in that dir, or resume old

`goose projects`: 

image

`goose project`: 

image

This second one is a short cut to most recent project only


angiejones (approved):

Kvadratni (approved):
OMG. This will push me back to CLI

PR #2509, opened 2025-05-12
chore: Add docker to allowed deeplink commands

bwalding:
I was trying to make deployment of MCPs simpler for our users, while avoiding local installation of Python/Node modules. To this end, I am using Docker.

I would like to use the `goose://` scheme to install (to avoid manual configuration of the `config.yaml`) - however, there is an allow-list which only permits `jbang`, `uvx`, `npx`, and `goosed`.

image

There doesn't seem to be an overwhelming reason not to allow `docker` here - `jbang`, `uvx`, and `npx` all run arbitrary code that exists in a repository somewhere.

I do appreciate that this is "yet another command," but I feel like Docker is a pretty common and generic way to launch sandboxed MCP commands that Goose should support relatively cleanly.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210278353063201

### DIFF
--- a/documentation/src/utils/install-links.ts
+++ b/documentation/src/utils/install-links.ts
@@ -29,7 +29,7 @@ export function getGooseInstallLink(server: MCPServer): string {
   }
   
   const parts = server.command.split(" ");
-  const baseCmd = parts[0]; // jbang, npx or uvx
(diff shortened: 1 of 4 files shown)


michaelneale (approved):
we now also have an optional allowlist so this makes sense to open it up more

PR #2514, opened 2025-05-12
fix not being able to click copy to clipboard

zanesq:
fix not being able to click copy to clipboard


ig-raiderler:
Interesting solution to this problem! I've been working on similar challenges recently.

zanesq:
nvm already fixed

PR #2517, opened 2025-05-12
allow running `goose run` with no session persistence

Kvadratni:
this will be a helpful option for massive automations using goose 

image


lilydelalande:
could this run in a server somewhere this way?

lily-de (approved):

Kvadratni:
> could this run in a server somewhere this way?

I don't see why not

PR #2518, opened 2025-05-13
Update provider configuration handling from new endpoint

zanesq:
Follow up from https://github.com/block/goose/pull/2484 realized we missed a spot where we were relying on the old /configs/providers format.

## Changes
- Updated frontend code to use the new `/config/providers` GET endpoint instead of the old `/configs/providers` POST endpoint
- Simplified the provider configuration code by:
  - Removing unused `getProvidersList` function
  - Removing unused `Provider` interface from types.ts
  - Using hardcoded required keys list for consistency with rest of the codebase
  - Transforming the new backend response format to maintain compatibility with existing frontend code

## Context
In commit 7027de62, we removed the `/configs` endpoint as part of moving token limits to the backend. However, we missed updating the frontend code that was still trying to use `/configs/providers`. This PR completes that refactor by updating the frontend to use the new endpoint and cleaning up related unused code.

## Testing
- Verified provider configuration status is correctly displayed
- Confirmed active providers are properly detected
- Tested provider configuration changes are saved and reflected in the UI

### DIFF
--- a/ui/desktop/src/components/settings/models/hardcoded_stuff.tsx
+++ b/ui/desktop/src/components/settings/models/hardcoded_stuff.tsx
@@ -1,41 +1,3 @@
-export const openai_models = ['gpt-4o-mini', 'gpt-4o', 'gpt-4-turbo', 'o1'];
-
-export const anthropic_models = [
-  'claude-3-5-sonnet-latest',
-  'claude-3-5-sonnet-2',
-  'claude-3-5-haiku-latest',
-  'claude-3-opus-latest',
-];
-
-export const google_models = [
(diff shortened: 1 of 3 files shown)


ui/desktop/src/components/settings/models/hardcoded_stuff.tsx
```diff
@@ -1,41 +1,3 @@
-export const openai_models = ['gpt-4o-mini', 'gpt-4o', 'gpt-4-turbo', 'o1'];
```
zanesq, line 1: Removed these unused constants

yingjiehe-xyz (approved):

PR #2520, opened 2025-05-13
blog typos

michaelneale:
this was a human crafted blog, every word, so has some confusing typos to fix.


The-Best-Codes (approved):
I love the whole blog post, I'll have to share with some others it when I have time :star_struck:

angiejones:
pulled into #2521

PR #2527, opened 2025-05-13
feat: add tool repetition monitoring to prevent infinite loops

Kvadratni:
Adds a new tool monitoring system that can detect and prevent infinite loops caused by repeated identical tool calls. The implementation includes:

- New ToolMonitor struct for tracking tool call patterns
- CLI flag --max-tool-repetitions to configure limits
- Integration with Agent dispatch system
- Session configuration support

The monitor tracks consecutive identical tool calls and can prevent them after a configurable limit is reached, helping prevent cases where the AI gets stuck in a loop making the same tool call repeatedly.

Example usage:
  goose run --max-tool-repetitions 5 ...

image

### DIFF
--- a/crates/goose-cli/src/cli.rs
+++ b/crates/goose-cli/src/cli.rs
@@ -220,6 +220,15 @@ enum Command {
         )]
         debug: bool,
 
+        /// Maximum number of consecutive identical tool calls allowed
+        #[arg(
+            long = "max-tool-repetitions",
+            value_name = "NUMBER",
+            help = "Maximum number of consecutive identical tool calls allowed",
+            long_help = "Set a limit on how many times the same tool can be called consecutively with identical parameters. Helps prevent infinite loops."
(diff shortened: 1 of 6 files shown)


crates/goose-cli/src/cli.rs
```diff
@@ -220,6 +220,15 @@ enum Command {
         )]
         debug: bool,
 
+        /// Maximum number of consecutive identical tool calls allowed
+        #[arg(
+            long = "max-tool-repetitions",
```
yingjiehe-xyz, line 225: do we want to have the default value for it? for UI case

yingjiehe-xyz (approved):

crates/goose-cli/src/cli.rs
Kvadratni: nope. no limit by default. Didn't want to affect existing logic

PR #2530, opened 2025-05-14
feat: show recipe explanation

lifeizhou-ap:
**Why**
[Jira ticket](https://block.atlassian.net/browse/DX-798)

**What**
- Added extra arg `--explain` to show the recipe's title, description and parameters (See screenshot)
- It also show the user the parameter values if they pass the param in the command line with `--explain`
- Refactored the `load_recipe` function and created 3 relevant functions
   - `load_recipe_as_template`: renders the recipe with parameters including validating the recipe. This is used when running the recipe
   - `load_recipe`: only load and validate the recipe without rendering it as a template. This is used for `goose recipe validate`  and `goose recipe deeplink`
   - `explain_recipe_with_parameters`: load, validate recipe and explain the recipe
- Added the "Boolean" in the  recipeInputType
  
Screenshot 2025-05-14 at 8 23 05 pm


michaelneale (approved):
@lifeizhou-ap I think this is fine - but had to merge it in with main as it had some code overlap with max's iteration limit

PR #2531, opened 2025-05-14
fixed the conflict_with in no-session arg

lifeizhou-ap:


crates/goose/src/config/base.rs
```diff
@@ -346,6 +346,7 @@ impl Config {
         let mut file = OpenOptions::new()
             .write(true)
             .create(true)
+            .truncate(false)
```
lifeizhou-ap, line 349: Hi @faces-of-eth 👋 !

The lint suggested to add .truncate, otherwise it fails the lint check.  I added .truncate(false) so that it is the same behaviour as without set truncate flag.  You can double check whether you would like to set true or false for this flag.  Thanks!

michaelneale (approved):

PR #2533, opened 2025-05-14
cli(command): Add `export` command to CLI for markdown export of sessions

dbraduan:
Title.

We need this code, so that we can have a clean way of parsing a session into a human-readable text format, otherwise we have to write our own parser of .jsonl without benefits of the types in the codebase.

Use case: we want to build internal sharing tool, and for that we would ideally have a clean format to export sessions.


crates/goose-cli/src/session/export.rs
```diff
@@ -0,0 +1,363 @@
...
+use mcp_core::role::Role;
+use serde_json::Value;
+
+const MAX_STRING_LENGTH_MD_EXPORT: usize = 4096; // Generous limit for export
+
+fn value_to_simple_markdown_string(value: &Value, export_full_strings: bool) -> String {
+    match value {
+        Value::String(s) => {
+            if !export_full_strings && s.len() > MAX_STRING_LENGTH_MD_EXPORT {
+                format!("`[REDACTED: {} chars]`", s.len())
```
hasnain-db: can we show the first N chars and then do something like `[ ... trimmed : N chars ... ]`

crates/goose-cli/src/session/export.rs
```diff
@@ -0,0 +1,363 @@
...
+                                    base_indent_str,
+                                    base_indent_str,
+                                    s.trim(),
+                                    base_indent_str
+                                ));
+                            } else {
+                                md_string.push_str(&format!("`{}`\n", s.replace('`', "\\`")));
+                            }
+                        }
+                        Value::Object(_) | Value::Array(_) => {
```
hasnain-db: why is this handled here and not with say a recursive call to `value_to_markdown`?

crates/goose-cli/src/commands/session.rs
```diff
@@ -115,3 +118,197 @@ pub fn handle_session_list(verbose: bool, format: String, ascending: bool) -> Re
...
+
+    // Create an internal struct mimicking the Session's export function
+    // This allows us to reuse the exact same export logic
+    struct ExportSession {
+        messages: Vec<goose::message::Message>,
+        session_file: PathBuf,
+    }
+
+    impl ExportSession {
+        // Copy of the export_message_history_to_markdown logic from Session
```
hasnain-db: why are we copying the logic?

hasnain-db (commented):
Thanks for doing this!

Perhaps the PR description can also explain the use case we're looking to enable, so it can be reviewed with that in mind?

I do think this needs some tests so we can see what the output looks like and check for any regressions.

crates/goose-cli/src/commands/session.rs
dbraduan: good call out, this was just copied over from my interim implementation inside proper agent session, but then I found a way to not instantiate full blown agent session to access messages.

dbraduan:
@hasnain-db addressed your comments, LMK your thoughts.

crates/goose-cli/src/session/export.rs
```diff
@@ -0,0 +1,1095 @@
+use goose::message::{Message, MessageContent, ToolRequest, ToolResponse};
+use mcp_core::content::Content as McpContent;
+use mcp_core::resource::ResourceContents;
+use mcp_core::role::Role;
+use serde_json::Value;
+
+const MAX_STRING_LENGTH_MD_EXPORT: usize = 4096; // Generous limit for export
+const REDACTED_PREFIX_LENGTH: usize = 100; // Show first 100 chars before trimming
```
hasnain-db, line 8: nit - should we say TRIMMED (here and below)

dbraduan:
Should be good to go, @angiejones ?

dbraduan:
should be good now, @angiejones

PR #2537, opened 2025-05-14
Shea/gdrive labels

sheagcraig:
Google Workspace includes the concept of org-wide "Labels" which can be applied to files, and which are composed of arbitrary fields of different data types. Examples of labels that could be applied include labels for marking Confidentiality, or signaling the retention period for automatically generated Google Docs.

This PR adds the ability to search for or list the labels available for a Google Drive, and allows you to update the labels applied to a file (remove, unset a field, add a new label, change a label). It turns out that this is pretty involved due to the 
varying data types for fields (including no fields!) and the fact that the API allows you to batch change labels.

## Example invocations
- "Add the 'Confidential' label to the Google Doc 'Big Trouble in Little China 2 Script' with the 'Extremely Sensitive' choice for the 'Confidentiality' field."
- "List the labels available for the Google Drive."
- "Remove the 'Confidential' label from the Google Doc 'Big Trouble in Little China 2'

## Out of scope
What this PR does not do is 
- allow the creation or management of labels themselves. 
- It also does not include the ability to search for Drive files based on labels (although it's still possible, just super inefficient).

The first is not something likely to be done by an MCP.

The second I will create a second PR for, as I'm planning on expanding the current `search_tool` to allow using the query mini-language rather than just doing name searches.

## Approach
Unfortunately, the Google Drive API requires knowing _beforehand_ what the IDs of labels are before allowing you to list the labels applied to a file (the label IDs are a parameter to the file listing endpoint). This requires us to list the labels available for the Drive, but that function is part of a separate, related API, the Google Drive Labels API.

The Google Drive API v3 crate that we're using for this MCP server can handle the label updates, but it must have been last updated prior to the creation of the Labels API (or it omitted it for unknown reasons). Therefore, this PR also includes a `google_labels` module allows us to perform the needed operations to satisfy the label goals above.

## Bonus!
While working on this I discovered that the MCP Inspector was failing to load the Gdrive server due to missing annotations on the Docs and Sheets tools, so I added those in to match the other tools.


crates/goose-mcp/Cargo.toml
```diff
@@ -47,9 +48,21 @@ lopdf = "0.35.0"
...
 umya-spreadsheet = "2.2.3"
-keyring = { version = "3.6.1", features = ["apple-native", "windows-native", "sync-secret-service", "vendored"] }
+keyring = { version = "3.6.1", features = [
+    "apple-native",
+    "windows-native",
+    "sync-secret-service",
+    "vendored",
+] }
 oauth2 = { version = "5.0.0", features = ["reqwest"] }
+utoipa = { version = "4", optional = true }
```
kalvinnchau:
```suggestion
utoipa = { version = "4.1", optional = true }
```

we're using 4.1 in other places so lets keep the versions the same

crates/goose-mcp/src/google_drive/oauth_pkce.rs
```diff
@@ -315,37 +321,62 @@ impl GetToken for PkceOAuth2Client {
...
-            // 2. Token refresh failed
-            // 3. There are no valid tokens yet
-            // Fallback: perform interactive OAuth flow
-            match self.perform_oauth_flow(scopes).await {
-                Ok(token) => {
-                    debug!("Successfully obtained new access token through OAuth flow");
-                    Ok(Some(token))
-                }
-                Err(e) => {
+            // If we get here, just use the original scopes
```
kalvinnchau: minor nit: any reason to remove all the comments on ordering?

kalvinnchau (approved):
just minor nit, but lgtm!

crates/goose-mcp/src/google_drive/oauth_pkce.rs
sheagcraig: There is no reason that I can think of. What rationale I may have had at the time has evaporated in the interim.

kalvinnchau (approved):

Issue #2560, opened 2025-05-16
Discover Goose recipe file via GOOSE_RECIPE_PATH


lifeizhou-ap:
**Please explain the motivation behind the feature request.**
Currently we can discover the recipe by
- specific recipe file path
- find the <recipe_name>.yaml or <recipe_name>.json under the current directory
- find the `recipe.yaml` or `recipe.json` under the github path <recipe_name>

User does not have opportunity to provide the specific path as a local repo repository to discover the recipe.  The GOOSE_RECIPE_PATH environment variable can be used to add additional directories where Goose will look for recipes.

**Describe the solution you'd like**
- Discover the recipe in each path listed in the environment variable `GOOSE_RECIPE_PATH` if the environment variable exists.

 ✅ Final Logic for `goose run --recipe <recipe_name> or <recipe_file>`

##### 🔹 Local Lookup:
search for <recipe_name>.yaml or <recipe_name>.json  in the order of
a) current directory
b)in  each path listed in the environment variable `GOOSE_RECIPE_PATH` if the environment variable exists

##### 🔹 GitHub Fallback (optional):
If the recipe is not found locally **and** the environment variable `GOOSE_RECIPE_GITHUB_REPO` is set (e.g., `my_org/goose-recipes`), it will download the recipe file from GitHub:

         
         https://github.com/<GOOSE_RECIPE_GITHUB_REPO>/blob/main/<recipe_name>/recipe.yaml
         or
         https://github.com/<GOOSE_RECIPE_GITHUB_REPO>/blob/main/<recipe_name>/recipe.json
         

- [x] I have verified this does not duplicate an existing feature request

lifeizhou-ap:
This issue is resolved in https://github.com/block/goose/pull/2561

PR #2561, opened 2025-05-16
feat: discover recipe in the paths from environment variable GOOSE_RECIPE_PATH

lifeizhou-ap:
Details in https://github.com/block/goose/issues/2560

### DIFF
--- a/crates/goose-cli/src/recipes/search_recipe.rs
+++ b/crates/goose-cli/src/recipes/search_recipe.rs
@@ -20,7 +20,6 @@ pub fn retrieve_recipe_file(recipe_name: &str) -> Result<(String, PathBuf)> {
     }
     retrieve_recipe_from_local_path(recipe_name).or_else(|e| {
         if let Some(recipe_repo_full_name) = configured_github_recipe_repo() {
-            println!("{}", e);
             retrieve_recipe_from_github(recipe_name, &recipe_repo_full_name)
         } else {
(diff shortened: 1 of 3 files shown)


crates/goose-cli/src/recipes/search_recipe.rs
```diff
@@ -1,31 +1,71 @@
...
         if let Some(recipe_repo_full_name) = configured_github_recipe_repo() {
+            println!("{}", e);
             retrieve_recipe_from_github(recipe_name, &recipe_repo_full_name)
         } else {
             Err(e)
         }
     })
 }
 
+fn retrieve_recipe_from_local_path(recipe_name: &str) -> Result<(String, PathBuf)> {
```
DOsinga: This method seems very noisy now; I would expect it to be just search for the recipe first in the current directory and if the environment variable is set, also search those paths, but no need I would think to tell the user if the variable is not set or where you searched - just if it can't be found, report that.

The other thing is that I think you want to search for the recipe directly not for the recipe_name/recipe.json|yaml. I know this is different from how we do it with the git integration, but I think that makes sense.

That would simplify the whole thing to like in python:

for path in [".", *env['GOOSER_RECIPE_PATH'].split(":")]:
   for ext in EXTENSIONS:
      if os.path.exists(path + "/" + recipe_name + "." + ext): etc

PR #2567, opened 2025-05-16
Add Tests for Optional Params with Empty Default (Fixed in v1.0.24)

gknoblauch:
## Description:

While using Goose v1.0.22, I encountered a bug where optional parameters with an empty string ("") as a default value were incorrectly treated as required. This caused goose run to fail with a misleading error:

```yaml
  - key: repo
    input_type: string
    requirement: optional
    description: name of the repo. uses the current one if not selected
    default: ""
```
Running:
```sh
goose run --recipe some-recipe
```

Would result in:
```
Error: Failed to render the recipe - please check if all required parameters are provided
```

This behavior was incorrect, as "" should be accepted as a valid default for optional parameters.

Upon investigation, I found this issue was already resolved in [#2444](https://github.com/block/goose/pull/2444) (released in v1.0.24). However, since I had already written tests to reproduce the bug, I’m including them in this PR to ensure it remains covered moving forward.

What’s Included:
	•	Regression tests verifying that empty default values on optional parameters are handled correctly.


lifeizhou-ap (approved):

yingjiehe-xyz:
can you fix the code format error?

PR #2574, opened 2025-05-18
docs: remove fetch limitation

angiejones:
removing the warning about gemini models when using the fetch mcp


blackgirlbytes (approved):

acekyd (approved):

michaelneale:
@angiejones the "stalled" are just as it won't run those builds for documentation only - but the branch protection rules require it, trying to see if a way around (but you can override and merge)

Issue #2512, opened 2025-05-12
Unable to select Claude Sonnet as model for Copilot provider


anderssv:
From `goose configure`, choose Copilot:

```
┌   goose-configure 
│
◇  What would you like to configure?
│  Configure Providers 
│
◇  Which model provider should we use?
│  Github Copilot 
│
◇  Enter a model from that provider:
│  claude-3.7-sonnet
│
◇  Server error: Object {"error": String("An internal server error occurred.")}
│
└  Failed to configure provider: init chat completion request with tool did not succeed.
```

This is different from failing with a bad model string. We are using sonnet in VSCode and IDEA at the org level. The default (gpt-4o) works fine.

kcalixto:
i'm also having the same issue!

matthewgrossman:
Also having this issue for `claude-3.7-sonnet`. 

It's strangely difficult to find the identifiers for the available models (and it differs per-account), but taking hints from [aider's docs](https://aider.chat/docs/llms/github.html#where-do-i-get-the-token), you can list all the models via:
```
apikey=$(jq -r 'first(.[]) | .oauth_token' ~/.config/github-copilot/apps.json)
curl -s -H "Authorization: Bearer $apikey" https://api.githubcopilot.com/models | jq -r '.data[].id'
```

Models that have worked:
- `gpt-4o`
- `o3-mini`

Models that seem broken:
- `claude-3.5-sonnet`
- `claude-3.7-sonnet`
- `gpt-4.1`

`gpt-4.1` gives:
```
◇  Enter a model from that provider:
│  gpt-4.1
│
◇  Request failed: error decoding response body
│
└  Failed to configure provider: init chat completion request with tool did not succeed.
```

While, the `claude-*` models give:
```
◇  Enter a model from that provider:
│  claude-3.7-sonnet
│
◇  Server error: Object {"error": String("An internal server error occurred.")}
│
└  Failed to configure provider: init chat completion request with tool did not succeed.
```

fxck:
```
Welcome to goose! Let's get you set up with a provider.
  you can rerun this command later to update your configuration

┌   goose-configure 
│
◇  Which model provider should we use?
│  Anthropic 
│
●  ANTHROPIC_API_KEY is set via environment variable
│  
◇  Would you like to save this value to your keyring?
│  No 
│
●  ANTHROPIC_HOST is set via environment variable
│  
◇  Would you like to save this value to your keyring?
│  No 
│
◇  Model fetch complete
│
[...]

PR #2576, opened 2025-05-18
feat: GitHub Copilot stream support (fixes gpt-4.1, claude)

acheong08:
gpt-4.1 only supports streaming with GitHub Copilot. When its use is detected, we convert it to standard response format.

image

Tested locally and it works pretty well. Don't mind if it isn't merged, mostly for personal use. Let me know if there's any style issues or whatever.

btw I'm not a Rust dev at all. I first wrote it in Go before translating to Rust [src](https://github.com/acheong08/copilot-proxy/blob/main/unstream/conversion.go). Might have a bit of unrusty style


eli0shin:
any chance this is the issue with other copilot models as well as reported in https://github.com/block/goose/issues/2512?

acheong08:
Had a look, indeed it is. For now we need to hard code which models are stream-only since the models API doesn't seem to tell us that

acheong08:
There seems to be other issues with Claude. Even with streaming set to true, I get 500s with certain tool calls. Might be a bug in my code.

acheong08:
Just pushed a fix for claude!

eli0shin:
Thank you!

GitMurf:
Looking forward to feedback from the Goose team! Nice work @acheong08

taniacryptid:
Thank you for creating and sharing this PR. I'll be sure to ping the dev team to take a look and review!

yingjiehe-xyz (approved):
Thanks for the contribution!

Issue #2573, opened 2025-05-17
Fast Apply - Morph


bhaktatejas922:
**Please explain the motivation behind the feature request.**
Patch edits and current editing is a bit brittle and prone to a choppy user experience.

Fast apply is the current frontier approach - https://morphllm.com is a Fast Apply model that outputs changes at 1600 tokens/sec

Morph has a free tier or usage based tier

- [x] I have verified this does not duplicate an existing feature request

michaelneale:
see attached PR - looks like it can work, have not tried any benchmarks yet

PR #2580, opened 2025-05-19
feat: optional fast edit models

michaelneale:
for: https://github.com/block/goose/issues/2573 

models that specialise editing can be used to reduce the load on the main LLM providers, increase accuracy and quality and speed and lower cost.
This supports both morphllm and relace models (and any openai compatible endpoint)


michaelneale:
I ran it through the benchmark ("system" is default, local is with morph): 

## Summary of Benchmark Results

I've run the `developer_search_replace` benchmark using both the system `goose` binary and the local `./target/release/goose` binary. Here are the results:

### System `goose` Binary Results

**Scores:**
- 3 out of 4 runs scored 1.0 (success)
- 1 run scored 0.0 (failure)

**Execution Times:**
- 32.86 seconds
- 70.13 seconds
- 37.41 seconds
- 41.97 seconds

**Average Execution Time:** 45.59 seconds
**Success Rate:** 75%

### Local `./target/release/goose` Binary Results

**Scores:**
- 4 out of 4 runs scored 1.0 (success)

**Execution Times:**
- 50.65 seconds
- 64.52 seconds
- 75.51 seconds
- 68.33 seconds

**Average Execution Time:** 64.75 seconds
**Success Rate:** 100%

### Comparison

1. **Success Rate:**
   - Local binary: 100% (4/4 successful runs)
   - System binary: 75% (3/4 successful runs)

2. **Execution Time:**
   - Local binary: Slower with an average of 64.75 seconds
   - System binary: Faster with an average of 45.59 seconds

3. **Consistency:**
   - Local binary: More consistent in terms of success rate (all runs succeeded)
   - System binary: Less consistent (one run failed)

The local build appears to be more reliable in completing the benchmark successfully, but it takes longer to execute. The system binary is faster on average but had one failed run.

bhaktatejas922:
oof these applies took forever. can we rerun? should be ~30% faster now

michaelneale:
@bhaktatejas922 I will take another look - don't worry too much about those times, the fact it is did stuff which wasn't reliably working before so measuring work vs not is tricky! we need to get better at these benchmarks

Kvadratni (approved):
Lgtm

Issue #2504, opened 2025-05-10
Pressing Ctrl+C or ESC in the CLI during a menu selection should not exit the app


bendavis78:
**Bug description**

During a menu selection prompt such as the "approve/deny" menu, if I press Ctrl+C or ESC, it should end the menu and put the user back into a normal prompt cursor.

**Steps to reproduce**
1. Ask it to read something like `README.md`
2. Wait until it prompts for the "approve/deny" selection
3. Press Ctrl+C or ESC
4. The program will exit completely (no error code returned)

**Expected behavior**
Pressing Ctrl+C or ESC during a menu selection should return the user to the main goose prompt `( O)>`.

**Screenshots**
![Image](https://github.com/user-attachments/assets/218748ad-a5d2-489a-885f-f4b50341366d)

**System details:**
 - **OS & Arch:** Arch Linux AMD64
 - **Interface:** CLI
 - **Version:** 1.0.22
 - **Extensions enabled:** Developer tools
 - **Provider & Model:** anthropic/claude-3-7-sonnet

The-Best-Codes:
I've got a PR almost ready. Just working out some details.

The-Best-Codes:
Here is my progress so far. Interested to see what the goose team thinks about this.

![Image](https://github.com/user-attachments/assets/175291fb-b8f8-4ad4-a5df-24a3a18f3838)

----

When there is a tool call, there are now 4 options:

- Allow
- Always allow
- Deny
- Cancel

If you cancel the tool call, you are returned to the chat to prompt the AI again. What the AI sees is a failure from the tool call with the message "Tool call cancelled by user". What you see is "Tool call cancelled. Returning to chat...".

Pressing Ctrl/Cmd + C or Escape will automatically select the Cancel option instead of exiting the program.

taniacryptid:
Thank you for creating this PR, @The-Best-Codes ! Will be sure to ping the team to take a look 💖

The-Best-Codes:
I haven't created a PR yet @taniashiba, I was asking the team if I should. I have one ready though so I will right away! ❤

The-Best-Codes:
PR is ready!
https://github.com/block/goose/pull/2586

The-Best-Codes:
@bendavis78 #2586 is merged thanks to @zanesq and @taniashiba. I'm guessing the new fix will be included in the next release of goose 🥳

danielzayas:
[...]

PR #2586, opened 2025-05-19
feat: fix program crashes and allow cancelling tool calls

The-Best-Codes:
This PR addresses #2504.

## Before

When goose is on manual approval mode, you will see these 3 options when goose makes a tool call:

- Allow
- Always Allow
- Deny

When pressing Ctrl/Cmd + C or Escape, the program's default behavior in other sections of the CLI is to:

 - Cancel the ongoing behavior and return to the goose chat
 - If already in the goose chat, exit the program

But on tool calls, it immediately exits the program with an error. There is no option to cancel a tool call.

## After

To address this, there are now 4 options when goose calls a tool in manual approval mode:

- Allow
- Always allow
- Deny
- Cancel

If you cancel the tool call, you are returned to the chat to prompt the AI again. The AI sees a failure message from the tool call with the content `Tool call cancelled by user`. The user sees `Tool call cancelled. Returning to chat...`.

Pressing Ctrl/Cmd + C or Escape will automatically select the Cancel option instead of exiting the program. So this is kind of a 2-in-1 PR: It addresses #2504 by introducing a new feature to solve the bug.

----

Are there any feedback or concerns? Please let me know!


taniacryptid:
Thank you for sharing this PR!

The-Best-Codes:
So many updates to the main branch, goose team is busy today :grinning:

The-Best-Codes:
All tests passing :partying_face:

PR #2592, opened 2025-05-19
feat: Handle MCP server notification messages

jamadeo:
This adds handling for server-initiated notification messages. Currently it supports logs and progress messages. This involved a decent amount of work up and down the MCP service stack, so it's a hefty change. Notifications means the services need to support listening for more messages besides a single request response.

The change also adds a tiny bit of refactoring to support it:
- moves request/response matching into the service layer out of the transport layer
- simplifies client construction a bit, using the tower service object as an implementation within the client (if need be, we can still expose the whole middleware bit, but right now we're just using it to add timeouts)

https://github.com/user-attachments/assets/e9049d74-e1af-4353-8c9b-f481fa981211


jamadeo:
One other thing to note: I really don't know which of the popular MCP servers out there are using this, so I'm not sure where you'd expect to start seeing more output. 

Also maybe worth considering whether the shell tool in the developer extension should stream anything back?

crates/mcp-client/src/client.rs
```diff
@@ -97,34 +99,63 @@ pub trait McpClientTrait: Send + Sync {
     async fn list_prompts(&self, next_cursor: Option<String>) -> Result<ListPromptsResult, Error>;
 
     async fn get_prompt(&self, name: &str, arguments: Value) -> Result<GetPromptResult, Error>;
+
+    fn take_notification_receiver(&mut self) -> Option<mpsc::Receiver<JsonRpcMessage>>;
 }
 
 /// The MCP client is the interface for MCP operations.
-pub struct McpClient<S>
+pub struct McpClient<T>
```
michaelneale, line 115: deliberate change from S to T generic type sig?

michaelneale:
@jamadeo very cool - would like to think about how (minimally) GUI could do this - and absolutely I think build in devleroper one (shell and maybe write) should do it as it would make things feel much faster

michaelneale:
@jamadeo would love to see this in action with the developer MCP in goose, shell command or editor, really anything to see what it feels like

jamadeo:
re-opened as #2613 since I can't change the source branch from the fork

PR #2593, opened 2025-05-19
wip: ToolSelector scaffolding and ToolSelector platform tool

wendytang:


crates/goose/src/agents/agent.rs
```diff
@@ -33,6 +33,7 @@ use mcp_core::{
 
 use super::platform_tools;
 use super::tool_execution::{ToolFuture, CHAT_MODE_TOOL_SKIPPED_RESPONSE, DECLINED_RESPONSE};
+use crate::agents::tool_selector::{DefaultToolSelector, ToolSelector, ToolSelectionStrategy, create_tool_selector};
```
alicehau: missing this tool_selector in this PR ?

crates/goose/src/agents/platform_tools.rs
```diff
@@ -112,3 +113,30 @@ pub fn manage_extensions_tool() -> Tool {
         }),
     )
 }
+
+pub fn tool_selector_tool() -> Tool {
```
alicehau, line 117: in the case of the default behavior, the agent should just have access to all the tools from enabled extensions.

but with this new tool, we're adding a tool to select tools. for the default case, would the agent use this tool? Or would the agent just not have this tool available then? 

also why would the llm determine which tool strategy to use? shouldn't it always be the same (either default, vector, etc.,) for the session?

wendytang: Based on the discussion, the agent would rely on the tool selector/picker to filter the tools. It should always have access to this platform tool

PR #2599, opened 2025-05-20
fix: append the recipe prompt when start recipe in a session

lifeizhou-ap:
**Why**
Currently the prompt of a recipe file is not rendered in Goose Desktop.

**What**
Submit the recipe prompt automatically when starting a session with recipe in Goose Desktop (via recipe goose deep link)

**Video**

https://github.com/user-attachments/assets/b954cf90-6f7f-4b2f-931b-4bdf45ee661e

https://github.com/user-attachments/assets/7c354c2b-bbb9-4673-ac3a-db7b64f46e37

**Question**
https://github.com/block/goose/blob/81332ab914db2e22467aaafa68f8c2800fa0f3cb/ui/desktop/src/components/ChatView.tsx#L495-L508

Just wondering whether the above code block is still used? I am checking this just in case my change impacts this block.  From the code it seems that might not be used, because there is no profile attribute in RecipeConfig class.  cc @zakiali @kalvinnchau


zanesq:
Will check it out tomorrow 👍

ui/desktop/src/main.ts
```diff
@@ -285,10 +285,11 @@ const windowMap = new Map<number, BrowserWindow>();
 
 interface RecipeConfig {
   id: string;
-  name: string;
+  title: string;
```
zakiali, line 288: wouldn't this break things if renamed to title? taking a look, this actually isn't a thing anymore?

michaelneale: yeah may be dead code

michaelneale (approved):
didn't get to try it, but looks ok - just check that name change from 'name' to 'title' makes sense and existing links work?

lifeizhou-ap:
> didn't get to try it, but looks ok - just check that name change from 'name' to 'title' makes sense and existing links work?

thanks @michaelneale, @zakiali 

I have not found anywhere using `name` in Goose recipe https://github.com/lifeizhou-ap/goose/blob/main/crates/goose/src/recipe/mod.rs#L32, but in the code (before my change), it already refer to `title` [somewhere](https://github.com/block/goose/blob/0dbf155a650a49e2a13d5e9b80139fc3a44ef81a/ui/desktop/src/components/ChatView.tsx#L495).

ui/desktop/src/main.ts
lifeizhou-ap: https://github.com/block/goose/pull/2599#issuecomment-2896227530

zakiali (approved):
lgtm and thanks for the video! your changes don't affect the code block you referenced (that sets the title on sessions started via a recipe)

PR #2607, opened 2025-05-21
ui-v2 titlebar and fonts

zanesq:
added draggable titlebar and cash fonts

image

### DIFF
--- a/ui/desktop/package-lock.json
+++ b/ui/desktop/package-lock.json
@@ -45,7 +45,6 @@
         "react-select": "^5.9.0",
         "react-syntax-highlighter": "^15.6.1",
         "react-toastify": "^8.0.0",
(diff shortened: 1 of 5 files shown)


zanesq:
I'm open to that. Usually there needs to be some indication that its a title bar that people can drag. I will make it the same light border we have in goose for now until we have direction. I need to fix some tailwind stuff first and wanted to land this in the meantime.

spencrmartin:
Yea wondering if we keep the current styling a bit as we'll use that section for a hover on page and chat filters/commands
Screenshot 2025-05-21 at 11 09 25 AM
Screenshot 2025-05-21 at 11 09 32 AM

zanesq:
Sounds good, will do!

PR #2613, opened 2025-05-21
feat: Handle MCP server notification messages

jamadeo:
This adds handling for server-initiated notification messages. Currently it supports logs and progress messages. This involved a decent amount of work up and down the MCP service stack, so it's a hefty change. Notifications means the services need to support listening for more messages besides a single request response.

The change also adds a tiny bit of refactoring to support it:
- moves request/response matching into the service layer out of the transport layer
- simplifies client construction a bit, using the tower service object as an implementation within the client (if need be, we can still expose the whole middleware bit, but right now we're just using it to add timeouts)

https://github.com/user-attachments/assets/212eafd9-c005-4ec3-aab8-6fa49f92186d

https://github.com/user-attachments/assets/bc91e29a-dd83-4c2c-b33f-4ca2c0e1d0b7

### DIFF
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -64,6 +64,12 @@ jobs:
           RUST_BACKTRACE: 1
           CROSS_VERBOSE: 1
         run: |
+          # Install protoc if on macOS
+          if [ "${{ matrix.os }}" = "macos-latest" ]; then
+            brew install protobuf
+            export PROTOC=$(which protoc)
+          fi
+
           export TARGET="${{ matrix.architecture }}-${{ matrix.target-suffix }}"
           rustup target add "${TARGET}"
           echo "Building for target: ${TARGET}"
@@ -72,7 +78,7 @@ jobs:
           echo "Cross version:"
           cross --version
 
-          # 'cross' is used to cross-compile for different architectures (see Cross.toml)
+          echo "Building with explicit PROTOC path..."
           cross build --release --target ${TARGET} -p goose-cli -vv
 
           # tar the goose binary as goose-<TARGET>.tar.bz2
--- a/Cross.toml
+++ b/Cross.toml
@@ -1,4 +1,11 @@
 # Configuration for cross-compiling using cross
+[build.env]
+passthrough = [
+    "PROTOC",
+    "RUST_LOG",
+    "RUST_BACKTRACE"
+]
+
 [target.aarch64-unknown-linux-gnu]
 xargo = false
 pre-build = [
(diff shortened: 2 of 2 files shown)


michaelneale:
nice ...

image


image


so this does subjectively give nice feedback. Need to think a bit on the design how to present it. I think both in GUI and CLI it should be some sort of spinner like activity - which just shows the message (subtle) which changes as new one comes in (instead of printing out). That would feel like a lot more fine grained progress, but the bones of it are here! really cool!

jamadeo:
@michaelneale I like it. Here's what it looks like as a spinner:


https://github.com/user-attachments/assets/bc91e29a-dd83-4c2c-b33f-4ca2c0e1d0b7

still very informative of progress without taking more than a line of screen real estate

jamadeo:
Here's the example MCP server that emits log and progress messages:

```
import asyncio
from fastmcp import FastMCP, Context

mcp = FastMCP("Demo 🚀")


@mcp.tool()
async def count(ctx: Context) -> int:
    """Counts and streams results"""
    await ctx.info("Counting...")

    until = 10
    for i in range(1, until + 1):
        await ctx.report_progress(i, until)
        await asyncio.sleep(0.5)

    await ctx.info("Finished counting")
    return until


if __name__ == "__main__":
    mcp.run()
```

michaelneale:
.bundle

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -488,27 +496,92 @@ impl DeveloperRouter {
...
+            .wait()
             .await
             .map_err(|e| ToolError::ExecutionError(e.to_string()))?;
 
-        let stdout_str = String::from_utf8_lossy(&output.stdout);
-        let output_str = stdout_str;
+        let output_str = output_task
+            .await
+            .map_err(|e| ToolError::ExecutionError(e.to_string()))?
+            .map_err(|e| ToolError::ExecutionError(e.to_string()))?;
```
baxen: nit: repeated map?

crates/mcp-client/src/client.rs
```diff
@@ -97,34 +102,65 @@ pub trait McpClientTrait: Send + Sync {
...
-pub struct McpClient<S>
+pub struct McpClient<T>
 where
-    S: Service<JsonRpcMessage, Response = JsonRpcMessage> + Clone + Send + Sync + 'static,
-    S::Error: Into<Error>,
-    S::Future: Send,
+    T: TransportHandle + Send + Sync + 'static,
 {
-    service: Mutex<S>,
+    service: Mutex<tower::timeout::Timeout<McpService<T>>>,
```
baxen, line 114: it's less important that we hardcode for now because we have `rmcp` to migrate to anyways, but why do we need to force the timeout wrapped service?

baxen (approved):
This is a great feature! 

We should make sure we test thoroughly with existing mcps and SSE version (i don't see any examples with SSE here) because so much surface area is updated. We'll likely need to update post switching to `rmcp` so i'm less concerned with backwards compatibility in the mcp crates here as long as integration tests are passing

crates/mcp-client/src/client.rs
jamadeo: Now that McpClient is built with a Transport instead of a Service, the service itself is really just an implementation detail for the McpClient. In fact we're really only using the tower::Service in order to use the timeout wrapper.

It could be made generic here again but makes things decently more complicated and since it isn't part of the public API it feels less useful to do so

jamadeo:
thanks @baxen for review! I created an integration test that covers the new functionality over both SSE and STDIO: https://github.com/block/goose/blob/b896dee383268021b9bd4bce2897035a2f72afc7/crates/mcp-client/examples/integration_test.rs

Also did a bunch of manual testing with some of the bundled MCPs: google drive, memory, developer of course. Everything LGTM.

GitMurf:
@jamadeo / @michaelneale please see this gh issue I just opened: https://github.com/block/goose/issues/2767

This PR "broke" developer extension shell tool calls in goose cli. It seems it is not properly waiting for the tool output to complete (e.g., shell command output for an ls command) before responding to the LLM. The details are in the gh issue. Let me know if you have any questions over there.

PR #2620, opened 2025-05-21
Enable streaming responses in ask util

Kvadratni:
## Summary
- stream AI responses when using `ask` helper
- server now chunks assistant replies into 64-character pieces for streaming

## Testing
- `cargo test -q` *(fails: Could not connect to server)*


DOsinga:
this looks interesting, but has been in draft for a long time, so closing for now

PR #2622, opened 2025-05-21
ui-v2: Integrate bossui + chart component

zanesq:
Integrated bossui and the chart component for the bar and area chart tiles as an example. The colors and area chart style are not matching the registry so need to look into that more but the basic integration is working by pulling in the [chart component](https://github.com/block/goose/blob/a76b16c39f2c9db5f3eeb2b0257fa87028f2b13e/ui-v2/src/components/ui/chart.tsx) from the registry (running locally for now) and using it in our tiles.

Also `lint-staged` isn't running so we have lots of linting and typescript errors. I'll work on fixing that next so we can keep quality high.

image

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -122,6 +122,21 @@ impl Session {
         }
     }
 
+    /// Helper function to summarize context messages
+    async fn summarize_context_messages(
+        messages: &mut Vec<Message>,
+        agent: &Agent,
+        message_suffix: &str,
+    ) -> Result<()> {
+        // Summarize messages to fit within context length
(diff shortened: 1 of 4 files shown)


ui-v2/src/App.tsx
```diff
@@ -8,7 +8,7 @@ const App: React.FC = (): React.ReactElement => {
   return (
     <div className="">
       <div className="titlebar-drag-region" />
-      <div className="h-10 border-b-1 w-full" />
+      <div className="h-10 w-full" />
```
zanesq, line 11: Removed top bar border @spencrmartin

alexhancock (approved):

PR #2623, opened 2025-05-21
fix: don't stop when in non interactive when you need to summarize

michaelneale:
goose run (recipes or otherwise) will sometimes stop due to context size and ask user when it must not.


Kvadratni (approved):

lifeizhou-ap (approved):
Looks good!

just a small suggestion:

```
let (summarized_messages, _) = self.agent.summarize_context(&self.messages).await?;
                                            let msg = format!("Context maxed out\n{}\nGoose summarized messages for you.", "-".repeat(50));
                                            output::render_text(&msg, Some(Color::Yellow), true);
                                            self.messages = summarized_messages;
```
This block seems to be duplicated. Maybe it can be extract to a function

PR #2626, opened 2025-05-22
fix: replace exec with spawn

zanesq:
For added security replace exec with nodejs spawn in IPC handlers. Also added some validation to other IPC handlers.


michaelneale:
not sure how to test this - but I seem to recall some challenges in the past in some environments when packaged - so would want to try this out a bit (probably fine though)

michaelneale:
.bundle

zanesq:
@michaelneale thanks yes I recently fixed the `.bundle` command glad it came in handy 👍

PR #2631, opened 2025-05-22
fix: Automatically added the prompt after the app is initialised with recipe instructions

lifeizhou-ap:


michaelneale (approved):
didn't get a chance to try it but looks good

michaelneale:
.bundle

lifeizhou-ap:
> didn't get a chance to try it but looks good

Thank you @michaelneale I have tested it. 

@DOsinga found this bug.  The bug is: when the recipe started in Goose Desktop, it started the chat (agent) with recipe prompt but the instruction has not been added into in the agent system prompt.

PR #2632, opened 2025-05-22
cli(ux): add hint for enter/ctrl-j if input is empty

dbraduan:
tiny improvement because I always forget that it's ctrl-J :)

https://github.com/user-attachments/assets/ad2dfe13-173f-43ff-a15a-3a7d0a307405


angiejones (approved):

crates/goose-cli/src/session/completion.rs
```diff
@@ -352,8 +352,13 @@ impl Helper for GooseCompleter {}
 impl Hinter for GooseCompleter {
     type Hint = String;
 
-    fn hint(&self, _line: &str, _pos: usize, _ctx: &rustyline::Context<'_>) -> Option<Self::Hint> {
-        None
+    fn hint(&self, line: &str, _pos: usize, _ctx: &rustyline::Context<'_>) -> Option<Self::Hint> {
+        // Only show hint when line is empty
+        if line.is_empty() {
+            Some("Press Enter to send, Ctrl-J for new line".to_string())
```
dbraduan, line 358: maybe actually add info about slash commands here?

dbraduan: I also feel like it can be used for auto completing things in if there is a deterministic slash command, but maybe I am wrong. worth a separate PR in this case

PR #2633, opened 2025-05-22
ui-v2: match tile / chart designs

zanesq:
brought over styles from design library for now until we have a better way
still some issues with labels fitting in the fixed width tiles that we can address later

image


nahiyankhan:
as the screenshot says - great work keep it up! 😄

PR #2635, opened 2025-05-22
docs: add comprehensive model list for each provider

mikegehard-rise8:
Add a new section to the providers documentation that lists all supported models for each LLM provider. This makes it easier for users to:

- Find the correct model names when configuring Goose
- See the default model for each provider
- Understand which models are officially supported/tested

The list includes models for all supported providers:
- OpenAI
- Anthropic
- Google Gemini
- GitHub Copilot
- Azure OpenAI
- Ollama
- AWS Bedrock
- Databricks
- GCP Vertex AI
- Groq
- OpenRouter

This change improves discoverability and helps users configure Goose with their preferred model without needing to search the codebase.

### DIFF
--- a/documentation/docs/getting-started/providers.md
+++ b/documentation/docs/getting-started/providers.md
@@ -34,7 +34,7 @@ Goose relies heavily on tool calling capabilities and currently works best with 
 
 ## Supported Models by Provider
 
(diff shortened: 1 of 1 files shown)


blackgirlbytes:
hi thank you so much! Closing for now, but I agree -- we will update to just link to the correct documentation that we need!

PR #2661, opened 2025-05-24
mcp(developer): add fallback on .gitignore if no .gooseignore is present

dbraduan:
Noticed that our .gitignore wasn't really respected, and saw goose has a bit different approach to determining it's .gooseignore.

Adding a fallback to .gitignore before fallback to default 3 patterns.

This is a bit of a breaking change for people who used this MCP without previously having `.gooseignore`, but it's commonly used in other Agents:
- https://docs.cursor.com/context/ignore-files#files-ignored-by-default
- https://docs.windsurf.com/windsurf/advanced#enabling-cascade-access-to-gitignore-files

Alternatively, happy to hide this behavior under a config setting.

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -16,8 +16,7 @@ pub use goose::session::Identifier;
 
 use anyhow::{Context, Result};
 use completion::GooseCompleter;
-use etcetera::choose_app_strategy;
-use etcetera::AppStrategy;
+use etcetera::{choose_app_strategy, AppStrategy};
 use goose::agents::extension::{Envs, ExtensionConfig};
 use goose::agents::{Agent, SessionConfig};
 use goose::config::Config;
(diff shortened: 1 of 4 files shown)


salman1993 (approved):

dbraduan:
Should be good to go, @angiejones?

zanesq:
@dbraduan we might need to revert this, there are cases where people don't want to automatically have goose ignore git ignored files. wdyt?

From a user below:

Seems to me like those are different ignores with different purposes. This has been causing some annoyances for me, it seems cumbersome to create a blank local .gooseignore just to get it to not ignore anything. This is a kind of an anti-pattern for data workflows. I frequently work in a git repo and have .csv files gitignored because I don't want to be saving full datasets to the git history.  Goose now can't interact with those. It also seems like even if I define a blank .gooseignore in the same directory, it still ignores everything from the .gitignore.

Issue #2572, opened 2025-05-17
Fails to compile with error: passing argument 2 of 'onig_st_foreach' from incompatible pointer type [-Wincompatible-pointer-types]


svenstaro:
Trying to compile this after a clone results in many error like these:

> cargo:warning=ToolExecError: command did not execute successfully (status code exit status: 1): LC_ALL="C" "cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-gdwarf-4" "-fno-omit-frame-pointer" "-m64" "-I" "/home/svenstaro/src/goose/target/debug/build/onig_sys-380e8e24fc2e2a99/out" "-I" "oniguruma/src" "-DHAVE_UNISTD_H=1" "-DHAVE_SYS_TYPES_H=1" "-DHAVE_SYS_TIME_H=1" "-o" "/home/svenstaro/src/goose/target/debug/build/onig_sys-380e8e24fc2e2a99/out/a445302c6d3dcb51-regparse.o" "-c" "oniguruma/src/regparse.c"cargo:warning=oniguruma/src/st.c:38:5: error: initialization of 'int (*)(void)' from incompatible pointer type 'int (*)(long int,  long int)' [-Wincompatible-pointer-types]

[goose.log](https://github.com/user-attachments/files/20269358/goose.log)

Arch Linux
gcc 15.1.1
rustc 1.87.0

svenstaro:
It's probably this: https://github.com/rust-onig/rust-onig/issues/196

PR #2663, opened 2025-05-25
Bump onig to v6.5.1

svenstaro:
This fixes a build failure as reported in #2572.
See also https://github.com/rust-onig/rust-onig/issues/195.


svenstaro:
I reckon this should be a pretty safe update.

PR #2669, opened 2025-05-27
Zaki/goose llm wasm

zakiali:


DOsinga:
this looks interesting, but has been in draft for a long time, so closing for now

PR #2674, opened 2025-05-27
fix: use https to get goose recipes and cleanup temp cloned dir if the clone repo is not in good state

lifeizhou-ap:
### Fix 1
**Why**
Some user has `ssh` as a default github protocol in their `gh` config

**What
Specify github https protocol to get the api token to clone github recipes repo

### Fix 2

**Why**
At the moment, the repo is cloned to a temp directory. 
    - If the `.git` directory does not in the cloned local repo dir, this means that the repo has been cloned and we will skip cloning the repo again.  
    - However, if the local repo is in a bad state, (i.e. the local repo directory exist, but `.git` does not exist), we will clone the repo but it will throw error saying the local repo exist. 
 
**What**
When the local repo is in a bad state and causes error, we will remove the local repo and retry to clone the repo again


michaelneale (approved):
good catch

PR #2675, opened 2025-05-27
chore: goose self fixer from issue

michaelneale:
This adds a workflow which is triggered when an issue is labelled as "goose-fix" - it will then use goose (headless stytle) to try to address that github issue, and open a pull request from it. 

A form of "eating your own dogfood" or "drinking your own champagne" so to speak.

It won't trigger the PR build - it is up to a person to continue on from there (subsequent change to then help work on that PR).


.github/workflows/goose-issue-fixer.yml
```diff
@@ -0,0 +1,169 @@
...
+          BRANCH_NAME="goose-fix/issue-${{ github.event.issue.number }}-$TIMESTAMP"
+          git config user.name "Goose Bot"
+          git config user.email "goose-bot@users.noreply.github.com"
+          git checkout -b $BRANCH_NAME
+          echo "BRANCH_NAME=$BRANCH_NAME" >> $GITHUB_ENV
+      
+      - name: Run Goose to Fix Issue
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          GOOSE_MODEL: gpt-4o
```
alexhancock: not blocking of merge, but a question:

I am curious why we'd use `gpt-4o` here? It seems like in the headless case speed doesn't matter much and we'd go with the slowest/most capable coding model with reasoning capabilities. potentially a good test case for the new opus?

alexhancock (approved):

zanesq (approved):
💯

.github/workflows/goose-issue-fixer.yml
michaelneale: yeah 4.1 would be better

PR #2677, opened 2025-05-27
feat: stream LLM responses

jamadeo:
Only databricks provider to start, but should not be hard to do this for others that support streaming.

Desktop:

https://github.com/user-attachments/assets/7f4dd128-2833-43f5-9089-41e58671c470

CLI:

https://github.com/user-attachments/assets/0ec0bcfd-799b-4c25-898d-11db76c9bcdd


ui/desktop/src/hooks/useMessageStream.ts
```diff
@@ -192,6 +192,9 @@ export function useMessageStream({
     };
   }, [headers, body]);
 
+  // TODO: not this?
+  const [, forceUpdate] = useReducer((x) => x + 1, 0);
```
jamadeo, line 239: This almost certainly isn't the best way to do this, but without it I could not get new messages to trigger a re-render. I could use a react expert's help here :)

michaelneale:
@jamadeo looks really great - but there is a bug where it seems to not call tools correctly 
image so may need to chase that down

ui/desktop/src/hooks/useMessageStream.ts
jamadeo: @zanesq any ideas here? I want to re-render every time we add a chunk, and so far the only thing that made it work was adding this fake state to force it.

jamadeo:
> @jamadeo looks really great - but there is a bug where it seems to not call tools correctly

Thanks @michaelneale, forgot to mention here but I fixed this issue. My first stab didn't re-assemble streamed tool call messages (they are chunked too), but I've since added it.

crates/goose/src/agents/reply_parts.rs
```diff
@@ -123,28 +126,46 @@ impl Agent {
...
+
+        let mut stream = if provider.supports_streaming() {
+            provider
+                .stream(system_prompt.as_str(), &messages_for_provider, &tools)
+                .await?
+        } else {
+            let (message, usage) = provider
+                .complete(system_prompt.as_str(), &messages_for_provider, &tools)
+                .await?;
+            stream_from_single_message(message, usage)
```
baxen, line 188: i like the incremental migration option here

crates/goose/src/providers/formats/databricks.rs
```diff
@@ -358,38 +360,176 @@ pub fn response_to_message(response: Value) -> anyhow::Result<Message> {
...
+    model: String,
+}
+
+fn strip_data_prefix(line: &str) -> Option<&str> {
+    line.strip_prefix("data: ").map(|s| s.trim())
+}
+
+pub fn response_to_streaming_message<S>(
+    mut stream: S,
+) -> impl Stream<Item = anyhow::Result<(Option<ProviderUsage>, Option<Message>)>> + 'static
```
baxen: meganit: tripped me up that this is backwards here compared to other places that have message and usage

crates/goose/src/providers/base.rs
```diff
@@ -282,6 +330,15 @@ pub trait Provider: Send + Sync {
     }
 }
 
+pub type MessageStream = Pin<
```
baxen, line 336: nit: i think here or somewhere similar we should indicate what a message stream is now

IIUC, something like

```
/// A message stream yields partial text content but complete tool calls, all within the Message object
/// So a message with text will contain potentially just a word of a longer response
```

crates/goose/src/agents/subagent.rs
```diff
@@ -369,135 +370,148 @@ impl SubAgent {
...
 
-                    // Process each tool request and create user response messages
-                    for request in &tool_requests {
-                        if let Ok(tool_call) = &request.tool_call {
-                            // Send notification about tool usage
+                            // Send notification about response
                             self.send_mcp_notification(
-                                "tool_usage",
-                                &format!("Using tool: {}", tool_call.name),
+                                "response_generated",
```
baxen: i haven't looked yet, but this seems like we might have to change how we handle notifications from subagents to accomodate this possibly streaming?

crates/goose/src/agents/agent.rs
```diff
@@ -730,257 +730,272 @@ impl Agent {
...
-                        // Record tool calls in the router selector
-                        let selector = self.router_tool_selector.lock().await.clone();
-                        if let Some(selector) = selector {
-                            // Record frontend tool calls
-                            for request in &frontend_requests {
-                                if let Ok(tool_call) = &request.tool_call {
-                                    if let Err(e) = selector.record_tool_call(&tool_call.name).await {
-                                        tracing::error!("Failed to record frontend tool call: {}", e);
-                                    }
+                            // record usage for the session in the session file
```
baxen, line 799: nit: this might be a lot of file ops, should we delay this until after the stream completes?

maybe that's not a significant downside though i imagine this is fast in rust

crates/goose/src/agents/subagent.rs
```diff
@@ -369,135 +370,148 @@ impl SubAgent {
...
 
-                            // Handle platform tools or dispatch to extension manager
-                            let tool_result = if self.is_platform_tool(&tool_call.name) {
-                                self.handle_platform_tool_call(
-                                    tool_call.clone(),
-                                    &extension_manager,
+                            // Add delay before completion to ensure all processing finishes
+                            tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
+
+                            // Set status back to ready and return the final response
```
baxen: this is wrong i think? tested locally and subagents immediately terminate for me and i think this is the line. We need this to happen only after the stream exits i think?

crates/goose/src/agents/subagent.rs
baxen: after reviewing and more testing i think this might be wrong, we want to only notify tools right

baxen (commented):
This is looking very good, excited to have this! There are a few issues below i think we need to fix

- as mentioned in a comment subagents are broken for me, they terminate immediately. I also suspect, but can't yet test, that the notifications are going to not be handled correctly - we need to notify only tools i think
- unless i'm missing it, this will break the notion of the chat session summaries because they will update on every word streamed and stop after a few words (but it was intended after a few replies). i think we need to update that logic to use count of ids now
- agree about your note on the re-render in the UI, maybe @zanesq knows how to fix that

crates/goose/src/providers/formats/databricks.rs
jamadeo: Good catch, no reason for that. Updated

jamadeo:
Thanks @baxen for the review! I actually effectively reverted the subagent change because it doesn't really help anything to use streaming in that context.  If we, in the future, want to show partial text responses from models to subagents in a streamed way we can think about it then, but no reason to fit it in now.

crates/goose/src/agents/agent.rs
jamadeo: I tweaked this a bit so that the stream only yields Usage at the end of the message, which I think is simpler to work with anyway. The APIs are a bit weird here -- if you request a completion with 10 input tokens and it streams back in 20 chunks, you'll have 20 responses each with {input_tokens: 10, output_tokens: null}, followed by a final {input_tokens: 10, output_tokens: <total output tokens}.

crates/goose/src/agents/subagent.rs
jamadeo: fixed by using non-streaming for subagents

crates/goose/src/agents/subagent.rs
jamadeo: fixed by using non-streaming for subagents

michaelneale:
@jamadeo oops thought it would be an easier merge than that to update it, but looking good. Sorry have slightly broken it (looking at it) - feel free to yank that last merge if it is all broken and not fixed by your morning time.

michaelneale:
I think some work still needed: 

image

using databricks, when it finishes streaming never goes back to the prompt (even if I press enter)

ui/desktop/src/hooks/useMessageStream.ts
zanesq: @jamadeo I think the best way is to use an immutable approach instead something like this. Instead of mutating in place, always create a new array/object when updating state. React should detect the change and re-render automatically.

Instead of 
```
const lastMessage = currentMessages[currentMessages.length - 1];
lastMessage.content = [...lastMessage.content, ...newMessage.content];
forceUpdate();
```

try
```
currentMessages = [
  ...currentMessages.slice(0, -1),
  {
    ...currentMessages[currentMessages.length - 1],
    content: [
      ...currentMessages[currentMessages.length - 1].content,
      ...newMessage.content,
    ],
  },
];
```

michaelneale:
have been testing this with CLI + databricks and seems ✅

Issue #2665, opened 2025-05-26
computer controller is unable to run basic scripts (shell)


michaelneale:
If you setup goose with just CC on, and as it to look at files in a directory, it will encounter permission errors on the script it writes which causes it to thrash unnecessarily. Most people have developer on default so unlikely to use it. 

This often could be avoided as well by having directory and file read tool available (as an option).

salman1993:
@michaelneale i fed this issue into Codex and it produced this PR - https://github.com/block/goose/pull/2678. i haven't tested it out.

PR #2678, opened 2025-05-27
fix: script execution in ComputerController

salman1993:
https://github.com/block/goose/issues/2665

## Summary
- add `PermissionsExt` import for unix
- set execute permissions for quick scripts on unix platforms

## Testing
- `cargo fmt --all` *(fails: rustfmt not installed)*
- `cargo check --workspace` *(fails: could not fetch crates)*


yingjiehe-xyz (approved):

michaelneale:
close! already merged one from goose (which made sure it was formatted) - but that would work I think !

PR #2694, opened 2025-05-28
fix: improve extension startup error messages with command details

wonrax:
Before:
```
     Running `target/debug/goose`
Failed to start extension: git, "No such file or directory (os error 2)"
Please check extension configuration for git.
```

The user might mistake this to mean that the `git` command is not installed and available in their `$PATH`. I fixed the error message to be more descriptive.

After:

```
     Running `target/debug/goose`
Failed to start extension `git`: Could not run extension command (`uvxx mcp-server-git`): No such file or directory (os error 2)
Please check extension configuration for git.
```


wonrax:
@michaelneale I don't know who else to mention, but can you check this? If this can break the user API feel free to close it.

michaelneale (approved):
thanks @wonrax @zanesq - yeah this makes sense, nice catch (and glad that someone is seeing this stuff now), this makes sense to me.

PR #2699, opened 2025-05-28
Update CreateScheduleModal.tsx - ui/Select component uppercase

cbruyndoncx:
The ui component for Select is in uppercase Select.tsx file. Different case is used in the ui directory, only doing minimal change to make sure build error is gone.


DOsinga:
hey can you do the CDO thing @cbruyndoncx and we'll get this in

cbruyndoncx:
I need to figure that DCO thing out, but this change is no longer needed,  it has been fixed and thus I am closing.

PR #2718, opened 2025-05-29
feat: goose web for local terminal alternative

michaelneale:
Offer up a GUI which is analogous to the `goose session` cli, but on the web, locally (same semantics). 

Not documented via help - as it is more a fallback option

Is in web as will be available on any platform, and is part of the `goose-cli` crate as it can share most things. 

``` bash
goose web --help
Start a web server with a chat interface

Usage: goose web [OPTIONS]

Options:
  -p, --port <PORT>  Port to run the web server on [default: 3000]
      --host <HOST>  Host to bind the web server to [default: 127.0.0.1]
      --open         Open browser automatically when server starts
  -h, --help         Print help
```

This aims to use same cli as session, but serve it up as web. Currently doesn't use any exotic web frameworks (vanilla all the way)

image


alexhancock (approved):
A nice lightweight GUI option for users of the CLI who don't want to install the app

PR #2719, opened 2025-05-29
feat: lead/worker model

michaelneale:
I have been using with good effect an expensive model like opus claude4 for a bit, and switching to claude4 sonnet mid stream. 

This feature will automatically switch from large to small as session goes on so you can make more efficient use if you set a lead model. this is similar to the `/plan` mode but you don't have to remember to call it. 

How it works: 

set the GOOSE_LEAD_MODEL and optionally GOOSE_LEAD_PROVIDER 

this works with the plan feature, and similar to it but is automatic for tool calling/execution, vs plan which is more for planning/chat. 

Also added docs and connected it with the plan feature (existing)


angiejones:
feedback [here](https://discord.com/channels/1287729918100246654/1287729920319033345/1377662226533716029) and [here](https://discord.com/channels/1287729918100246654/1287729920319033345/1377663072587485215) basically stating they'd like to confirm the switch

michaelneale:
@angiejones interesting - I would think that you would configure that ahead of time - if you don't want it to do it, dont' have that extra model (as by default wont have one)?

crates/goose/src/providers/lead_worker.rs
```diff
@@ -0,0 +1,634 @@
...
+                            tracing::info!("✅ SWITCHING BACK TO WORKER MODEL: Exiting fallback mode - worker model resumed");
+                        }
+                    }
+                }
+
+                // Increment turn count on any completion (success or task failure)
+                let mut count = self.turn_count.lock().await;
+                *count += 1;
+            }
+            Err(_) => {
```
baxen, line 181: i haven't quite understood why in this implementation yet, but when this triggered during testing the session dropped instead of continuing

image

Do we need to retry here? i dont see how a retry is triggered

blackgirlbytes:
Great concept! However, I'd prefer more user control over the automatic switching. Similar to Roo Code's approach where it asks 'to code, I need to switch to coder now' - could we add prompts before switching models? The automatic behavior feels too 'magical' and I'd like confirmation before model changes.

crates/goose/src/providers/lead_worker.rs
michaelneale: will take a look - was that a provider failure - Ie opus rate limiting?

documentation/docs/guides/environment-variables.md
```diff
@@ -45,7 +47,34 @@ export GOOSE_PROVIDER__TYPE="anthropic"
...
+Configure a lead/worker model pattern where a powerful model handles initial planning and complex reasoning, then switches to a faster/cheaper model for execution.
+
+| Variable | Purpose | Values | Default |
+|----------|---------|---------|---------|
+| `GOOSE_LEAD_MODEL` | **Required to enable lead mode.** Specifies the lead model name | Model name (e.g., "gpt-4o", "claude-3.5-sonnet") | None |
+| `GOOSE_LEAD_PROVIDER` | Provider for the lead model | [See available providers](/docs/getting-started/providers#available-providers) | Falls back to GOOSE_PROVIDER |
+| `GOOSE_LEAD_TURNS` | Number of initial turns using the lead model | Integer | 3 |
+| `GOOSE_LEAD_FAILURE_THRESHOLD` | Consecutive failures before fallback to lead model | Integer | 2 |
+| `GOOSE_LEAD_FALLBACK_TURNS` | Number of turns to use lead model in fallback mode | Integer | 2 |
+
```
blackgirlbytes, line 62: thanks for adding these!

blackgirlbytes:
okay @michaelneale so making sure I understand this:
1. User set the lead and worker model in an env variable (that's the user opting in to the switch)
2. Goose shows a message that it's switching from lead to the worker model  in the CLI 
It sounds like both of these address concerns from the community where they wanted to have control/transparency
3. However, what I'm gathering from the PR is that while the env variables affect the UI..so like it will switch between lead and worker model, they will not see much a visual indicator that it's being switched?

Am I understanding this correctly and if so, could we add a little visual indicator?

Besides that, this PR is cool with me. I'm team "switch the model without me knowing..do the heavy decision making for me"

Sidenote: apologies for my rogue issue and comment that was added to this PR. I asked Goose to help me catch up on the Discord conversations related to this PR..and then it was like I will leave a comment on and open an issue based on the Discord conversation. Claude 4 is too powerful and fast 😆

michaelneale:
@blackgirlbytes no worries - meant to ask for your feedback, yes you have it right, will be a subtle indicator it is switching models (we can improve it over time too) of course few will use this feature out of the gate as it won't be a default feature of course. It isn't just about cost too, it could apply to local models where very very slow ones are used as the lead vs the worker which is faster and more efficient

blackgirlbytes:
> t just about cost too, it could apply to local models where very very slow ones are used as the lead vs the worker which is faster and more efficient

oh cool! I like it!

blackgirlbytes (approved):

documentation/docs/guides/environment-variables.md
michaelneale: yeah - wasn't sure if that is right place, ideally want to document this along with the planner model

blackgirlbytes: looks like the perfect place me!

crates/goose/src/providers/lead_worker.rs
michaelneale: yeah - good catch, I don't like the idea of it interpreting that failure, so it will just go to default in this case. It isn't about failing over.

michaelneale: so it will now switch to other model (which works better)

PR #2721, opened 2025-05-29
[goose-llm] add default implementation for extract using tool calling trick

salman1993:
Adds a default implementation using the tool calling trick for structured outputs


DOsinga:
can we close this @salman1993 since I think we're going the other way?

DOsinga:
closing for now. feel free to re-open if mistake

PR #2761, opened 2025-06-02
chore: small bit of a cleanup - removing unused dir

alexhancock:
I think this is an extra `components/ui` directory disconnected from the one in `src/components/ui`. Removing, but let me know if you were using it for something @nahiyankhan !


nahiyankhan (approved):

inez (approved):
nit: update description to match what's actually happening - removing file (`ui/desktop/components/ui/popover.tsx`), not just a dir

PR #2762, opened 2025-06-03
fix: increase limit for direct to disk for performance

michaelneale:
The limit was set to 20K chars, which was a bit low, well below context windows, but for some tools (eg JIRA) it was hurting performance by a large factor. 

This will still keep it under context windows so should be useful but hopefully not practically hurt performance.


baxen (approved):

michaelneale:
.bundle

PR #2766, opened 2025-06-03
chore: use hermit to install node, rust and protoc

lifeizhou-ap:


michaelneale (approved):
worth a change but keep in mind can roll it back if causes a problem

PR #2769, opened 2025-06-03
Add the ability to configure rustyline to use a different edit mode (e.g. vi)

ayax79:
By default the cli only supports emacs key bindings. This pull request allows a configuration parameter EDIT_MODE to be set that will allow the rustyline edit mode to be explicitly set. The following config allows the edit mode to be set to vi:

```
> ❯ : cat ~/.config/goose/config.yaml
GOOSE_PROVIDER: github_copilot
extensions:
  developer:
    bundled: true
    display_name: Developer
    enabled: true
    name: developer
    timeout: 300
    type: builtin
GOOSE_MODEL: gpt-4o
EDIT_MODE: vi
```


crates/goose-cli/src/session/builder.rs
```diff
@@ -220,3 +227,14 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
     output::display_session_info(session_config.resume, &provider_name, &model, &session_file);
     session
 }
+
+fn determine_edit_mode(edit_mode_str: String) -> EditMode {
```
cgwalters: Minor nit, it's more idiomatic to have functions take `&str`

crates/goose-cli/src/session/builder.rs
```diff
@@ -220,3 +227,14 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
...
+    match edit_mode_str.to_lowercase().as_str() {
+        "emacs" => EditMode::Emacs,
+        "vi" => EditMode::Vi,
+        _ => {
+            eprintln!("Invalid EDIT_MODE specified, defaulting to Emacs");
```
cgwalters: Unsure if we should really print-and-continue here, but OK.

crates/goose-cli/src/session/mod.rs
```diff
@@ -106,7 +108,7 @@ pub async fn classify_planner_response(
 }
 
 impl Session {
-    pub fn new(agent: Agent, session_file: PathBuf, debug: bool) -> Self {
+    pub fn new(agent: Agent, session_file: PathBuf, edit_mode: EditMode, debug: bool) -> Self {
```
cgwalters: Perhaps this should be `Option<EditMode>` and then we clearly don't override the default if it's unset?

crates/goose-cli/src/session/builder.rs
```diff
@@ -220,3 +227,14 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
...
+            EditMode::Emacs
```
cgwalters: See below comment

crates/goose-cli/src/session/builder.rs
ayax79: I thought about this too. Leaving it as a String made the map above more concise.. I didn't have to do this 
```rust
.map(|ref s| determine_edit_mode(s))
```
This particular function is private and only exists for readability. I could've just in-lined the whole thing above.

crates/goose-cli/src/session/builder.rs
ayax79: The error message is only displayed if the EDIT_PARAM is specified. I am not sure if it is worth killing the CLI over.

crates/goose-cli/src/session/mod.rs
ayax79: done

crates/goose-cli/src/session/builder.rs
ayax79: I decided to just inline this whole block

crates/goose-cli/src/session/builder.rs
ayax79: change the default to None, still not sold on killing the cli

ayax79: inlined this block now.

michaelneale:
nice - I think people that know about emacs conventions in UI input and want an alternative should have it, and doesn't change default

@cgwalters appreciate your diligence - we will need more maintainers soon I think! be careful what you wish for haha

I updated this to main, and tidied up some clippy/fmt errors

michaelneale:
sorry @cgwalters do you mind updating/resolving the conflicts? I had a look but worry I would break it

cgwalters:
> @cgwalters appreciate your diligence - we will need more maintainers soon I think! be careful what you wish for haha

I am interested in this conversation; I like four aspects of Goose:

- FOSS
- Rust
- CLI but with support for GUIs
- multi-model

And I'd like to try to convince my employer to invest a bit more formally based on this. I can't promise anything right now, but that's my hope.

> sorry @cgwalters do you mind updating/resolving the conflicts? I had a look but worry I would break it

Hi @michaelneale I'm not the PR author. I just did a drive by review here.

michaelneale:
oh apologies @cgwalters !

sorry @ayax79 would you be interested in continuing (some conflicts to resolve)

ayax79:
@michaelneale fixed!

michaelneale (approved):

Issue #2767, opened 2025-06-03
PR #2613 broke goose cli developer shell tool calls


GitMurf:
I mentioned this in the Discord but am bringing over all the info here.

This commit (which was large and changed a lot) has broken the github copilot provider in goose cli (and maybe other providers). Tools like developer shell commands output gets cutoff and only the first line or two gets sent back to the LLM. See my screenshot for example when trying to do an ls command. The notifications / "loader" or whatever you call it in the CLI you see repeated ("total 400") in numerous ways.

Here is the commit: https://github.com/block/goose/commit/03e5549b5466e754df15c4d5370c7514c628341f 
Here is the PR: https://github.com/block/goose/pull/2613 

Here for example is what it looks like running the ls shell command:

![Image](https://github.com/user-attachments/assets/80317abb-abf3-41fb-b7d7-8d13097fea2b)

Here is the response from the LLM stating that all that returned was "total 400" instead of the ls command output:

![Image](https://github.com/user-attachments/assets/75febe28-d0e4-4f8d-8fa9-d6b6917ae6ac)

If I run the ls command in my shell you see the full real response starts with "total 400" which is all that is being returned to the LLM after the tool call (with the actual ls results being cutoff):

![Image](https://github.com/user-attachments/assets/b4e17acf-8975-414a-9baa-ff18b251baa9)

Since the github copilot provider using models like gpt4.1 requires streaming, I wonder if that was not taken into account or tested for?

GitMurf:
Should be easy to repro in main branch with using github copilot provider, gpt 4.1 model, using the goose cli and having the LLM do basically any developer shell command tool call that expects output (ls, cat etc.).

taniacryptid:
Thank you for creating the issue here on GitHub for the team to look over, @GitMurf !

GitMurf:
If this is a github copilot provider specific issue, it is worth checking out this commit (https://github.com/block/goose/commit/4ae5e4264c86be1112a263eaef286640b42aa80d) which implemented streaming for gpt 4.1 and o4-mini models for the github copilot provider since it requires streaming. So maybe the PR did not take into account these requirements for the  copilot provider or something?

jamadeo:
Hey @GitMurf thanks for raising the issue -- I'm not able to reproduce it on main with copilot + gpt-4.1
```
goose % ./target/debug/goose run -t "run ls -al and summarize the result"
starting session | provider: github_copilot model: gpt-4.1
[...]

PR #2771, opened 2025-06-03
fix: Don't break from consuming subprocess output in shell tool until both streams are done

jamadeo:
Previously, we might break out of the output processing loop before everything is written. I can't reproduce this on my mac but I'm guessing other platforms might behave differently here.

This is an attempt at fixing #2767 in which the output is incomplete on a Windows+WSL setup.

This also includes a better clean-up of the spinner when we print the tool output, which had led to some confusion about where this was actually failing before.


cgwalters:
> Previously, we might break out of the output processing loop before everything is written.

:arrow_right:  https://github.com/block/goose/pull/2817#discussion_r2134103743

PR #2773, opened 2025-06-04
hotfix: don't always run prompt

michaelneale:
cc @lifeizhou-ap this may show the prompt instead of starting ... WIP


michaelneale:
.bundle

PR #2775, opened 2025-06-04
Fix paths in google drive mcp documentation

dhanji:
Normalizes all path references to credential files to use `$USER_HOME` and clarifies the documentation accordingly. Also clarifies that absolute paths are required. Without this, the plugin does not start on Mac OS.


baxen (approved):
LGTM!

PR #2789, opened 2025-06-05
docs: clarify github copilot installation instructions

cmdNiels:
GitHub Copilot installation does not work without the variable set and I saw a lot of users complaining about it.


rockwotj:
This is not specific to Github Copilot.

Maybe linux users should have the keyring off by default? A lot of linux systems seem to be headless or not having a keyring installed.

Issue #2790, opened 2025-06-05
Goose CLI: Working dir not properly set for new sessions


sudo-tee:
**Describe the bug**

When running goose for the first time (new session) the working path is not set properly in the session file. Restoring the session will then echo a warning like so `WARNING: The original working directory of this session was set to ...`

After reviewing the session file the working_dir is set to `/home/user` instead of my current cwd.

**To Reproduce**
1. Go to any folder
2. call `goose`  to create a new session
3. Quit goose
4. Resume the session
5. Observe the warning ` WARNING: The original working directory of this session was set to /home/francis. Your current directory is /home/francis/dots. Do you want to switch back to the original working directory?`

**Expected behavior**
goose should set the correct working_dir the first time

**Screenshots**

![Image](https://github.com/user-attachments/assets/853dd238-ff0b-4958-9302-a61c525dc041)

![Image](https://github.com/user-attachments/assets/c751ecab-8833-4c0e-825b-ab85db9db777)

**Please provide following information:**
 - **OS & Arch:** [Ubuntu WSL (Linux MTL-DB283 6.6.87.1-microsoft-standard-WSL2  SMP PREEMPT_DYNAMIC Mon Apr 21 17:08:54 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux)]
 - **Interface:** [CLI]
 - **Version:** [1.0.24-canary+2f8f8e5]
 - **Extensions enabled:** [developper, fetch, memory]
 - **Provider & Model:** [github_copilot]

s-soroosh:
I just opened [a small PR](https://github.com/block/goose/pull/2793) that should address this if I haven't missed something.

sam-at-block:
Im also seeing this, on both mac and linux. 

It's annoying because every time I resume a session, I get prompted like below and have to be careful to select NO

```
◇  WARNING: The original working directory of this session was set to /Users/<me>. Your current directory is /Users/<me>/... . Do you want to switch back to the original working directory?
```

Even when Im always in the code dir, not home. I can confirm all the session jsonl files have the first line showing the working_dir as my home dir.

wbontrager:
I'm having the same issue where the warning always shows when resuming a session. https://github.com/block/goose/issues/2790#issuecomment-3075104330

PR #2793, opened 2025-06-05
fix: use current working directory as default for session metadata

s-soroosh:
- Fixes https://github.com/block/goose/issues/2790

Previously, when a session had no messages or an invalid working directory, it would default to the home directory. This changes the behavior to use the current working directory instead, which I think is more intuitive and matches user expectations.

**Questions**
- Is there any drawback to use cwd as the default working directory instead of home directory?

### DIFF
--- a/crates/goose/src/session/storage.rs
+++ b/crates/goose/src/session/storage.rs
@@ -17,6 +17,12 @@ fn get_home_dir() -> PathBuf {
         .to_path_buf()
 }
 
+fn get_current_working_dir() -> PathBuf {
+    std::env::current_dir()
+        .or_else(|_| Ok::<PathBuf, io::Error>(get_home_dir()))
+        .expect("could not determine the current working directory")
+}
+
 /// Metadata for a session, stored as the first line in the session file
 #[derive(Debug, Clone, Serialize, ToSchema)]
 pub struct SessionMetadata {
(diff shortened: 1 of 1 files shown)


crates/goose/src/session/storage.rs
```diff
@@ -17,6 +17,10 @@ fn get_home_dir() -> PathBuf {
         .to_path_buf()
 }
 
+fn get_current_working_dir() -> PathBuf {
+    std::env::current_dir().expect("could not determine the current working directory")
```
DOsinga: would be better to default here to the home_dir rather than die

s-soroosh:
@DOsinga @michaelneale Thanks for the reviews! Updated the code to fallback to home dir in case working directory cannot be determined

michaelneale:
I think CWD make more sense @s-soroosh I like this, but not entirely clear how to test/reproduce, code reads fine, but do you have a scenario to test with?

michaelneale:
@s-soroosh also will need DCO https://github.com/block/goose/pull/2793/checks?check_run_id=45675312414 it looks like but LGTM otherwise.

michaelneale:
ie if you could show some samples of it working new way - then I think we are good to go with this for next release

michaelneale:
actually DCO skipped, this needs to be udpated to main and should pass the build

michaelneale:
@s-soroosh mind updating this to main branch again?

michaelneale:
closing for now until more feedback

nyonson:
Just curious what feedback is required for this patch? The issue is definitely a little painpoint in the CLI at the moment.

s-soroosh:
Hey, I was busy this week and haven't found time yet to address the latest feedback.
It's mainly about rebasing my branch and write down a scenario to reproduce this issue and how this patch fixes it. 
Will work on it later today or tomorrow.

s-soroosh:
Reopened a PR with the same changes [here](https://github.com/block/goose/pull/3513).

PR #2797, opened 2025-06-05
subagents

wendytang:
This is a prototype of subagents in goose - note we are still actively thinking about this but would love your input! 

demo: 
https://github.com/user-attachments/assets/922fdece-f9c2-41f8-ac67-fc1e409dbd20

`joke_recipe.yaml`
```
version: 1.0.0
title: Joke of the day
author:
  contact: douwe
description: Will tell you a joke of the day based on the current day
instructions: Your job is to tell a joke of the day
extensions:
- type: builtin
  name: developer
  display_name: Developer
  timeout: 300
  bundled: true
prompt: Based on what day it is today, generate a joke. Mention the day on the first line then an empty line and then the joke. Don't just say the date, but figure out if there's any cultural significance, like national shelves day
```

open questions designing subagents:
- How should sub-agents behave in various goose modes (autonomous, manual, chat)? 
- Should subagents run synchronous or asynchronously?
- Communication channel: should users be able to communicate with subagents directly?
- Extension config: how to design extension configurations with agents and sub-agents? Case: sub-agent only has extensions main agent has;  Case: sub-agent can utilize its own set of extensions
- Maintenance: subagents are a simplified version of the main agent, is there a way to subclass main agent for a maintenance standpoint or would we better off keeping them separate?

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -939,8 +939,7 @@ impl Session {
 
                                                             if min_priority > 0.1 && !self.debug {
                                                                 // High/Medium verbosity: show truncated response
-                                                                if msg.starts_with("Responded: ") {
-                                                                    let response_content = &msg[11..]; // Remove "Responded: " prefix
+                                                                if let Some(response_content) = msg.strip_prefix("Responded: ") {
                                                                     if response_content.len() > 100 {
                                                                         format!("🤖 Responded: {}...", &response_content[..100])
                                                                     } else {
@@ -955,7 +954,7 @@ impl Session {
                                                             }
                                                         }
                                                         _ => {
-                                                            format!("{}", msg)
+                                                            msg.to_string()
                                                         }
                                                     };
                                                     (formatted, Some(subagent_id.to_string()), Some(notification_type.to_string()))
--- a/crates/goose/src/agents/subagent_manager.rs
+++ b/crates/goose/src/agents/subagent_manager.rs
@@ -363,10 +363,14 @@ impl SubAgentManager {
                     turn_count + 1,
                     response_text
                 ));
-                conversation_result.push_str(&format!("\n[Task completed after {} turns]", turn_count + 1));
+                conversation_result.push_str(&format!(
+                    "\n[Task completed after {} turns]",
+                    turn_count + 1
+                ));
             }
             Err(e) => {
-                conversation_result.push_str(&format!("\n[Error after {} turns: {}]", turn_count, e));
+                conversation_result
+                    .push_str(&format!("\n[Error after {} turns: {}]", turn_count, e));
             }
         }
 


pablof7z:
I didn't see this on the docs, but would be very useful for agents to be able to specify their model; RooCode does this and it's extremely handy.

cgwalters:
This sounds cool - I watched https://www.youtube.com/watch?v=9tq-QUnE29U recently and was impressed. I assume this is aiming for something like that more built-in?

This PR kind of lacks a larger description of what it's aiming to do, seems like that'd be useful to have. Is there an issue that should be linked?

wendytang:
Thanks for the comments - I will check out these subagent implementations, pls let me know if you have other ideas

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -0,0 +1,278 @@
...
+        let mut subagent_manager = self.subagent_manager.lock().await;
+        if let Some(manager) = subagent_manager.as_mut() {
+            manager.process_updates().await
+        } else {
+            Vec::new()
+        }
+    }
+
+    /// Process subagent updates and build a context map
+    pub async fn process_subagent_updates(&self) -> HashMap<String, String> {
```
wendytang: process_subagent_updates doesn't seem to be used anywhere?

crates/goose/src/agents/agent.rs
```diff
@@ -608,6 +671,16 @@ impl Agent {
         Ok(Box::pin(async_stream::try_stream! {
             let _ = reply_span.enter();
             loop {
+                // Check for subagent notifications
```
wendytang: could we consolidate into a one liner function call? just to keep the loop easy to read

something like:
```
self.display_subagent_notifications()
```

lilydelalande:
To try out this (very nascent) feature, add this line to `~/.config/goose/config.yaml`:

```
ALPHA_FEATURES: true
```

crates/goose/src/agents/agent.rs
```diff
@@ -44,14 +44,21 @@ use mcp_core::{
...
 use super::platform_tools;
 use super::router_tools;
+use super::subagent_manager::SubAgentManager;
+use super::subagent_tools;
 use super::tool_execution::{ToolCallResult, CHAT_MODE_TOOL_SKIPPED_RESPONSE, DECLINED_RESPONSE};
 
 /// The main goose Agent
 pub struct Agent {
     pub(super) provider: Mutex<Option<Arc<dyn Provider>>>,
-    pub(super) extension_manager: Mutex<ExtensionManager>,
```
salman1993, line 55: why did we change this?

crates/goose/src/agents/agent.rs
```diff
@@ -257,6 +271,23 @@ impl Agent {
...
+                    .await,
+            )
+        } else if tool_call.name == SUBAGENT_LIST_TOOL_NAME {
+            ToolCallResult::from(self.handle_list_subagents().await)
+        } else if tool_call.name == SUBAGENT_CHECK_PROGRESS_TOOL_NAME {
+            ToolCallResult::from(
+                self.handle_get_subagent_status(tool_call.arguments.clone())
+                    .await,
+            )
+        } else if tool_call.name == SUBAGENT_SEND_MESSAGE_TOOL_NAME {
```
salman1993: why do we need 4 tool calls here? i was thinking we'd have one to call the subagent which does the task and returns the result or an error msg

crates/goose/src/agents/agent.rs
```diff
@@ -62,13 +69,61 @@ pub struct Agent {
...
 #[derive(Clone, Debug)]
 pub enum AgentEvent {
     Message(Message),
     McpNotification((String, JsonRpcMessage)),
-    ModelChange { model: String, mode: String },
+    ModelChange {
+        model: String,
+        mode: String,
+    },
+    SubagentNotification {
```
salman1993: how is this different from McpNotification?

jamadeo: It might be worth considering using a single event type for these. The MCP notification is just a pass-through of the JsonRPC payload, and maybe subagents could pass something like that?

crates/goose/src/agents/agent.rs
lilydelalande: next one lol -- tired of rebasing :D

crates/goose/src/agents/agent.rs
lilydelalande: i think because we need to carry around the extension manager and the mutex was not working

crates/goose/src/agents/agent.rs
lilydelalande: next PR :) i will need to think about it a bit more

crates/goose/src/agents/agent.rs
lilydelalande: I was getting compile errors with Mutex and changing to RwLock fixed them. I think it was related to the borrow checker and how we're passing the extension manager around to multiple subagents, but RwLock resolved the lifetime/borrowing issues

crates/goose/src/agents/agent.rs
lilydelalande: Good point! A vestige from when subagents were not ephemeral. We were thinking we'd want the granular control for complex workflows - like spawning a subagent, checking its progress, sending follow-up messages, etc. But you're right that for most use cases a single call_subagent tool that handles the full lifecycle and just returns the result would be much simpler.

lilydelalande: how should we handle cases where the subagent needs to ask follow-up questions or request more info from the main agent or user? Do we:

1. Just fail and return an error asking the user to provide more details upfront?
2. Have the subagent make its best guess with available info?
3. Keep some kind of back-and-forth mechanism?
4. Or design the tasks to be more self-contained?

crates/goose/src/agents/agent.rs
salman1993: my understanding is that RwLock is strictly more demanding than Mutex for compile issues, so surprised you're getting compilation issues with Mutex and not RwLock

crates/goose/src/agents/agent.rs
salman1993: my personal opinion: subagent tasks should be more self-contained (information gathering tasks such as search) and they return with a result or error msg. as subagent runs, it can report what its doing (what tool calls its making so that we can display to user)

for permissions, the user can sign-off when the subagent spawn tool is initially started but after that, subagent runs autonomously (since its a tool). 

listing subagents and sending msgs back & forth seems difficult to manage.

crates/goose/src/agents/agent.rs
lilydelalande: i wish i could remember what they were... 

also i think Mutex will cause issues if we have multiple subagents trying to access the extension manager concurrently (like parallel subagents) -- subagents will block each other even for read-only operations like checking what extensions are available or getting tool definitions with Mutex (I think)

crates/goose/src/agents/agent.rs
lilydelalande: I think probably worth discussing/keeping open, but I agree with you.

salman1993: Two use cases that will benefit from subagents:
1. tool routing (searching for the right tool) 
2. file search (figuring out which files to read & edit)

I think it'll be good to test it out on one of these atleast

salman1993 (approved):

salman1993:
please test out the agent with some popular extensions and typical workflows before merging (eg. shell, gdrive, slack)

crates/goose/src/agents/agent.rs
lilydelalande: I had to make some edits to give subagents access to platform tools but tool routing worked like a charm when explicitly stated (use a subagent to figure out what tools i'd need to go from designs to code)

crates/goose/src/agents/subagent_handler.rs
lilydelalande: removed!

crates/goose/src/agents/agent.rs
lilydelalande: thought about it and added!

DOsinga:
> I didn't see this on the docs, but would be very useful for agents to be able to specify their model; RooCode does this and it's extremely handy.

You can specify a model in a (sub) recipe or even have the main recipe (agent) specify it

Issue #2794, opened 2025-06-05
Add mutual TLS for providers


myaple:
**Please explain the motivation behind the feature request.**
Some of the APIs I work with require mutual TLS with custom CAs to validate the requestor. I would like to use goose with them natively, and that would require goose supporting mutual TLS.

**Describe the solution you'd like**
When building the http clients in each of the providers, pull in a path from configuration for each of the certificate, key, and certificate authorities, and use those to build the client.

**Describe alternatives you've considered**
I can use a proxy server to sidestep this issue, but that requires more technical expertise and friction for my colleagues.

**Additional context**
I'm happy to contribute this request if we can come to an agreeable path to take for it.

- [x] I have verified this does not duplicate an existing feature request

weishiuchang:
mTLS would be very much appreciated.

PR #2799, opened 2025-06-05
feat: adds mtls to openai and ollama (#2794)

myaple:
This PR adds mutual TLS support to the OpenAI and Ollama providers in goose and goose-llm. The default behavior remains unchanged, with not providing any certificates just skipping the appropriate sections of the client builder. The client builder is built with a helper function to aid in integration to other providers if necessary. 

The helper function is duplicated between the two crates as they seemed to be for effectively separate use cases - let me know if that's not desirable.

### DIFF
--- a/crates/goose/src/providers/ollama.rs
+++ b/crates/goose/src/providers/ollama.rs
@@ -1,6 +1,6 @@
 use super::base::{ConfigKey, Provider, ProviderMetadata, ProviderUsage, Usage};
 use super::errors::ProviderError;
-use super::utils::{get_model, handle_response_openai_compat};
+use super::utils::{build_http_client, get_model, handle_response_openai_compat, MutualTlsConfig};
 use crate::message::Message;
 use crate::model::ModelConfig;
 use crate::providers::formats::openai::{create_request, get_usage, response_to_message};
@@ -9,7 +9,7 @@ use async_trait::async_trait;
 use mcp_core::tool::Tool;
 use reqwest::Client;
 use serde_json::Value;
-use std::time::Duration;
+
 use url::Url;
 
 pub const OLLAMA_HOST: &str = "localhost";
@@ -41,9 +41,14 @@ impl OllamaProvider {
             .get_param("OLLAMA_HOST")
             .unwrap_or_else(|_| OLLAMA_HOST.to_string());
 
-        let client = Client::builder()
-            .timeout(Duration::from_secs(600))
-            .build()?;
+        // Configure mutual TLS
+        let mtls_config = MutualTlsConfig {
+            client_cert_path: config.get_param("OLLAMA_CLIENT_CERT_PATH").ok(),
+            client_key_path: config.get_param("OLLAMA_CLIENT_KEY_PATH").ok(),
+            ca_cert_path: config.get_param("OLLAMA_CA_CERT_PATH").ok(),
+        };
+
+        let client = build_http_client(600, Some(&mtls_config))?;
 
         Ok(Self {
             client,
@@ -104,12 +109,12 @@ impl Provider for OllamaProvider {
             OLLAMA_DEFAULT_MODEL,
             OLLAMA_KNOWN_MODELS.to_vec(),
             OLLAMA_DOC_URL,
-            vec![ConfigKey::new(
-                "OLLAMA_HOST",
-                true,
-                false,
-                Some(OLLAMA_HOST),
-            )],
+            vec![
+                ConfigKey::new("OLLAMA_HOST", true, false, Some(OLLAMA_HOST)),
+                ConfigKey::new("OLLAMA_CLIENT_CERT_PATH", false, false, None),
+                ConfigKey::new("OLLAMA_CLIENT_KEY_PATH", false, false, None),
+                ConfigKey::new("OLLAMA_CA_CERT_PATH", false, false, None),
+            ],
         )
     }
 
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -3,13 +3,15 @@ use async_trait::async_trait;
 use reqwest::Client;
 use serde_json::Value;
 use std::collections::HashMap;
-use std::time::Duration;
 
 use super::base::{ConfigKey, Provider, ProviderMetadata, ProviderUsage, Usage};
 use super::embedding::{EmbeddingCapable, EmbeddingRequest, EmbeddingResponse};
 use super::errors::ProviderError;
 use super::formats::openai::{create_request, get_usage, response_to_message};
-use super::utils::{emit_debug_trace, get_model, handle_response_openai_compat, ImageFormat};
+use super::utils::{
+    build_http_client, emit_debug_trace, get_model, handle_response_openai_compat, ImageFormat,
+    MutualTlsConfig,
+};
 use crate::message::Message;
 use crate::model::ModelConfig;
 use mcp_core::tool::Tool;
@@ -65,9 +67,15 @@ impl OpenAiProvider {
             .ok()
             .map(parse_custom_headers);
         let timeout_secs: u64 = config.get_param("OPENAI_TIMEOUT").unwrap_or(600);
-        let client = Client::builder()
-            .timeout(Duration::from_secs(timeout_secs))
-            .build()?;
+
+        // Configure mutual TLS
+        let mtls_config = MutualTlsConfig {
+            client_cert_path: config.get_param("OPENAI_CLIENT_CERT_PATH").ok(),
+            client_key_path: config.get_param("OPENAI_CLIENT_KEY_PATH").ok(),
+            ca_cert_path: config.get_param("OPENAI_CA_CERT_PATH").ok(),
+        };
+
+        let client = build_http_client(timeout_secs, Some(&mtls_config))?;
 
         Ok(Self {
             client,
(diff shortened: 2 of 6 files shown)


myaple:
Added mTLS support to the new liteLLM client since that has a similar use case to a custom OpenAI or ollama URL.

Also removed changes from the goose-llm crate since I'm not sure that's relevant to the overall goal of adding mTLS to goose cli.

DOsinga:
sorry for the non response for so long here! thanks for doing this. I'd be happy to merge as is, but I think we could do better here if we made this more general.

extract build_client to a common file, I think we want to do this anyway. call this one from all the providers. and then use a variable like: TLS_CERT_FILE / TLS_KEY_FILE / TLS_CA_FILE for all of them

what do you think? we need to get the specific provider handling under control

crates/goose/src/providers/litellm.rs
myaple, line 1: forgot to remove previous config vars

crates/goose/src/providers/ollama.rs
```diff
@@ -38,10 +38,11 @@ impl OllamaProvider {
             .get_param("OLLAMA_HOST")
             .unwrap_or_else(|_| OLLAMA_HOST.to_string());
 
-        let timeout: Duration =
-            Duration::from_secs(config.get_param("OLLAMA_TIMEOUT").unwrap_or(OLLAMA_TIMEOUT));
+        // Configure mutual TLS
```
myaple: dangling comment and space

crates/goose/src/providers/ollama.rs
```diff
@@ -110,6 +111,9 @@ impl Provider for OllamaProvider {
                     false,
                     Some(&(OLLAMA_TIMEOUT.to_string())),
                 ),
+                ConfigKey::new("OLLAMA_CLIENT_CERT_PATH", false, false, None),
```
myaple: also remove these configs

documentation/docs/guides/mutual-tls-openai.md
myaple, line 1: need to update this to be generic, if the guide is something they want merged generally

myaple: @DOsinga before I spend the time rewriting this, would you rather it updated or removed? are there other locations you'd rather me add to the docs for the new config?

myaple:
@DOsinga no worries! thanks for looking at it. I've unified the approach across ~~some~~ all of the providers ~~I'll continue working to integrate the rest of them where possible~~. I had limited it just to the most common providers for this sort of enterprise access control originally, which is why I chose OpenAI compatible interfaces and Ollama, but I'm happy to make it more generic. 

I asked one question about docs above, only other question I have is would you like me to make a global provider timeout config instead of having things like OLLAMA_TIMEOUT and a bunch of hardcoded 600 second timeouts? It feels awkward to just smash in hardcoded timeouts when there's already at least one configuration.

crates/goose/src/providers/databricks.rs
```diff
@@ -239,9 +238,7 @@ impl DatabricksProvider {
     ///
     /// Returns a Result containing the new DatabricksProvider instance
     pub fn from_params(host: String, api_key: String, model: ModelConfig) -> Result<Self> {
-        let client = Client::builder()
-            .timeout(Duration::from_secs(600))
-            .build()?;
+        let client = build_http_client(60, None)?;
```
myaple: typo, change to 600

crates/goose/src/providers/google.rs
```diff
@@ -52,6 +53,7 @@ pub const GOOGLE_DOC_URL: &str = "https://ai.google.dev/gemini-api/docs/models";
 pub struct GoogleProvider {
     #[serde(skip)]
     client: Client,
+    api_key: String,
```
myaple: unnecessary

crates/goose/src/providers/openai.rs
```diff
@@ -121,7 +97,7 @@ impl OpenAiProvider {
             .post(url)
             .header("Authorization", format!("Bearer {}", self.api_key));
 
-        let request = self.add_headers(request);
+        // Headers are now added in build_http_client
```
myaple: remove

crates/goose/src/providers/toolshim.rs
```diff
@@ -70,10 +70,7 @@ pub struct OllamaInterpreter {
 
 impl OllamaInterpreter {
     pub fn new() -> Result<Self, ProviderError> {
-        let client = Client::builder()
-            .timeout(Duration::from_secs(600))
-            .build()
-            .expect("Failed to create HTTP client");
+        let client = build_http_client(60, None)?;
```
myaple: typo

myaple:
> @DOsinga no worries! thanks for looking at it. I've unified the approach across ~some~ all of the providers ~I'll continue working to integrate the rest of them where possible~. I had limited it just to the most common providers for this sort of enterprise access control originally, which is why I chose OpenAI compatible interfaces and Ollama, but I'm happy to make it more generic.
> 
> I asked one question about docs above, only other question I have is would you like me to make a global provider timeout config instead of having things like OLLAMA_TIMEOUT and a bunch of hardcoded 600 second timeouts? It feels awkward to just smash in hardcoded timeouts when there's already at least one configuration.

I pulled out the docs, it was in a totally wrong format, not even worth it. If docs are needed, I'll start from scratch.

DOsinga:
sorry to keep banging on this - we just merged a refactor on how the providers work and they now all use an ApiClient object - it should be easy to add that to that now

myaple:
> sorry to keep banging on this - we just merged a refactor on how the providers work and they now all use an ApiClient object - it should be easy to add that to that now

its okay, I'll need some time to validate it again. with a holiday coming up, it'll probably be a week or so before it's ready again, I'll ping you when I'm ready for a review with the new builder methods. I've pushed what I've got so far, it probably won't be far off what I'll come to eventually, if you want to let me know if you're alright with it in structure.

It deviates slightly from other AuthMethods just cause mTLS doesn't use headers and such, so let me know if this looks okay and I'll finish up with writing tests.

michaelneale:
thanks @myaple did you have more you wanted to do or this is ready to try as is?

myaple:
> thanks @myaple did you have more you wanted to do or this is ready to try as is?

I believe this is all I wanted to do, I just haven't written tests for it as I'm on holiday without a real computer till next week. Feel free to try it out, and I'll write tests ASAP.

myaple:
TODOs from testing:

- [ ] modify ApiClient::new to build tls auth instead of in all of the with_* funcs
- [ ] modify each client creation in the providers to detect whether or not to use mutual TLS - potentially in addition to the bearer token instead of replacing it? 
- [ ] validate env vars as GOOSE_* are picked up

crates/goose/src/providers/google.rs
```diff
@@ -66,9 +66,14 @@ impl GoogleProvider {
...
 
-        let api_client =
+        let mut api_client =
             ApiClient::new(host, auth)?.with_header("Content-Type", "application/json")?;
 
+        // Apply TLS configuration if any TLS fields are configured
+        if let Some(tls_config) = TlsConfig::from_config()? {
+            api_client = api_client.with_tls_config(tls_config)?;
+        }
+
```
DOsinga: can't we now just do this once in the api_client constructor?

myaple: happy to, I was just following the pattern of the other with_* functions. I'll consolidate.

DOsinga: yeah, I get that, but if it is not provider specific, we can just always load it. then if we need something for providers that is not general, we can add it there.

myaple: alright, I cut it all down to just changes in the api_client itself, loads by default and keeps the rebuild_client bit for the other with_ funcs. I'll give it a test tomorrow and hopefully we can finish this PR then.

myaple:
I've tested this today and it is working. If there's anything else you want done in this PR, let me know, otherwise I think it's good to go @DOsinga @michaelneale

crates/goose/src/providers/api_client.rs
```diff
@@ -79,32 +195,72 @@ impl ApiClient {
...
+                    client_builder = client_builder.identity(identity);
+                }
+
+                // Load CA certificates
+                let ca_certs = tls_config.load_ca_certificates()?;
+                for ca_cert in ca_certs {
+                    client_builder = client_builder.add_root_certificate(ca_cert);
+                }
+            }
+        }
```
jamadeo: this is almost the same as the one in with_timeout - consider extracting that?

crates/goose/src/providers/api_client.rs
```diff
@@ -27,6 +30,119 @@ pub enum AuthMethod {
...
+                })?;
+
+                Ok(Some(identity))
+            }
+            (Some(_), None) => Err(anyhow::anyhow!(
+                "Client certificate provided but no private key"
+            )),
+            (None, Some(_)) => Err(anyhow::anyhow!(
+                "Client private key provided but no certificate"
+            )),
```
jamadeo: Consider returning an error in from_config if one but not both of cert/key are set instead of here. The field could then be an Option<(String, String)>

jamadeo (approved):

crates/goose/src/providers/api_client.rs
myaple: extracted to a private function

crates/goose/src/providers/api_client.rs
myaple: I put this in I think as you wanted it, my only concern is now the field of Option<(PathBuf, PathBuf)> is a little ambiguous as to which PathBuf is the cert vs key. It'll require anyone using the TlsConfig in the future to go back to the struct definition to read the comment to figure out which is which, if they need to for some reason. I don't mind either way. 

I think returning an error in from_config is the better approach regardless of the other fields cause it is necessary to have both the client cert and key, good catch.

Let me know which way you'd prefer, and I can squash + force push to make it one commit again.

jamadeo: Yeah, that's true and a good point. A struct would address that, e.g. 
```
struct TlsCertKeyPair {
    cert_path: PathBuf,
    key_path: PathBuf,
}
```

myaple: done, that feels much cleaner.

jamadeo:
fantastic, thanks for the contribution @myaple !

myaple:
> fantastic, thanks for the contribution @myaple !

For sure, thanks for working with me to get it across the line!

PR #2802, opened 2025-06-06
feat(ui): Add session branching capability for active and historical conversations

alexhayes-block:
# Session Branching Feature

This PR introduces a session branching capability that allows users to create new conversation branches from any point in both active and historical sessions.

## Changes

- Added branch buttons to user and assistant messages in active sessions
- Added branch buttons to messages in session history
- Added branch button in session header for quick access
- Implemented secure file handling for branching in the main process
- Added loading indicators and visual feedback during branching process
- Branch buttons only appear on hover, consistent with copy buttons

## Benefits

This feature enables users to:
- Explore different conversation paths from any point
- Correct errors in prompting without losing context
- Create variations of conversations easily
- Maintain a cleaner conversation history

## Screenshots

**Session history with branch buttons**
image

**Active session with branch buttons on hover**
image

## Demo Video

https://github.com/user-attachments/assets/71234b67-aec6-4397-8bdf-ce715ef15012

## Testing Done

- Verified branching works from any message in active sessions
- Verified branching works from any message in session history
- Confirmed proper loading states during branching process
- Tested edge cases with long conversations


alexhayes-block:
@michaelneale I chose to do this all in the UI to see if it would be a useful feature. Let me know if you think this functionality should be moved into the `goose` crate, perhaps it belongs in `crates/goose/src/session/storage.rs`?

michaelneale:
yeah I think could be a core feature in session storage ideally - that would be really cool I think

alexhayes-block:
@zanesq I've pushed the core session branching capability to the server. 

Note I've not implemented a way for the CLI to use this functionality mostly because it's not really clear what would be optimal from a UX perspective.

DOsinga:
I like this feature, but it looks like it is stale and it is still in draft? do we want to close this for now?

DOsinga:
closing stale drafts

michaelneale:
will close this - can revisit when interested (I like the idea still).

PR #2816, opened 2025-06-07
Nostrbook MCP is now on npm

alexgleason:
The Nostrbook MCP was previously developed in Deno, and required weird hacks to get it working with Node. Now it runs directly in node: https://www.npmjs.com/package/@nostrbook/mcp


angiejones:
thanks so much for keeping this updated!

PR #2819, opened 2025-06-07
fix: Refactor string truncation logic into reusable utility function to avoid panic (#2818)

### LINKED ISSUE #2818: Panic in goose projects command when handling Japanese text: byte index not on char boundary
**Describe the bug**

Describe the bug

The goose projects command panics with a character boundary error when processing Japanese text. Specifically, it fails when trying to slice a UTF-8 encoded Japanese string at byte index 37, which falls in the middle of the Japanese character '内' (bytes 36-39).

**To Reproduce**

Set up an environment with Japanese text in project descriptions or session names
Run goose projects command in terminal
The following panic error occurs:
thread 'main' panicked at crates/goose-cli/src/commands/project.rs:203:52:  
byte index 37 is not a char boundary; it is inside '内' (bytes 36..39) of `現在のディレクトリの変更内容からGitHub PRを作成してください。  ...

**Expected behavior**

The command should display the project selection screen normally, even when project information contains Japanese text.

**Screenshots**

N/A

 - **OS & Arch: Mac [e.g. Ubuntu 22.04 x86]
 - **Interface: CLI [UI/CLI]
 - **Version: v1.0.24 [e.g. v1.0.2]
 - **Extensions enabled: No [e.g. Computer Controller, Figma]
 - **Provider & Model: Claude 4 [e.g. Google - gemini-1.5-pro]

**Additional context**

toyamagu-2021:
Closes #2818 

## Motivation

Extract repeated string truncation code from project.rs and session.rs into a new utils module with safe_truncate function.
Also fix trailing
whitespace in configure.rs.

## Verification

`goose ps` works well.

Signed-off-by: toyamagu2021 <toyamagu-2021@gmail.com>


crates/goose-cli/src/utils.rs
```diff
@@ -0,0 +1,54 @@
...
+///
+/// # Returns
+/// A truncated string with "..." appended if truncation occurred
+pub fn safe_truncate(s: &str, max_chars: usize) -> String {
+    if s.chars().count() <= max_chars {
+        s.to_string()
+    } else {
+        let truncated: String = s.chars().take(max_chars.saturating_sub(3)).collect();
+        format!("{}...", truncated)
+    }
```
cgwalters, line 19: Totally fine as is, just a note this walks the string twice. Hopefully we aren't passing large file data here anyways.

cgwalters (approved):
I'm not a maintainer here, just doing a drive by review since I recently submitted a different PR.

crates/goose-cli/src/utils.rs
cgwalters: I also came across https://doc.rust-lang.org/std/primitive.str.html#method.floor_char_boundary which is exactly what we want here, but yeah still experimental.

toyamagu-2021:
@jamadeo Sorry, I've fixed lint error.

toyamagu-2021:
I've forgotten picking up https://github.com/block/goose/pull/3243 🙏 

```console
➜ ./bin/cargo clippy      
    Checking goose-cli v1.0.31 (/Users/tomoki-yamaguchi/github.com/toyamagu-2021/goose/crates/goose-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 5.57s
```

jamadeo (approved):
Thanks @toyamagu-2021, looks great.

PR #2826, opened 2025-06-09
Add Linux desktop package building workflow

michaelneale:
This PR adds a new GitHub workflow to build Linux desktop packages (.deb and .rpm files) for the Electron app.

## Changes
- New workflow 
- Builds both .deb and .rpm packages using existing Electron Forge configuration
- Runs on Ubuntu with necessary system dependencies
- Uploads packages as downloadable artifacts

## Testing
This workflow is currently configured to run on this specific branch for testing. Once verified, it can be updated to run on appropriate triggers.

The workflow should:
1. Build the goosed Rust binary
2. Copy it to the Electron app
3. Use Electron Forge to create .deb and .rpm packages
4. Upload them as artifacts that can be downloaded from the GitHub Actions page

EDIT: this is to test out building of linux desktop packages, once ready, it will be part of release on main
you can find rpm and deb files here: https://github.com/block/goose/actions/runs/15526901766?pr=2826


ATrueLight4:
I downloaded the .zip for for Debian
Installed it on an Ubuntu desktop
Ran it and clicked on the Tell me what you can do, responded as expected

.github/workflows/fake-release.yml
```diff
@@ -0,0 +1,77 @@
+name: "Fake Release Test (Linux Integration)"
+
+# This workflow tests the Linux desktop integration with release process
+# Triggers only on the specific branch to test the reusable workflow
+on:
+  push:
+    branches: [ "micn/linux-desktop-app" ]
```
kalvinnchau: do we still want to keep this around? if so maybe just name it a dedicated `test/linux-desktop-app` branch to use?

kalvinnchau (approved):
one minor comment, but LGTM!! 🔥

.github/workflows/fake-release.yml
michaelneale: will be deleting before merging or renaming yeah

michaelneale:
merging now - if this causes a problem with next release - can remove it from release.yml and canary.yml from the list of things to build

PR #2829, opened 2025-06-09
fix: Use the existing spinner in interactive mode

jamadeo:
Cleans up tool streaming output in interactive mode.

Before:

https://github.com/user-attachments/assets/06c1d97e-63bd-4230-b82d-32967e203e6f

After:

https://github.com/user-attachments/assets/601fb700-90d2-4c47-9038-298d4ae9e106

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -891,7 +891,11 @@ impl Session {
                                                     v.to_string()
                                             },
                                         };
-                                        progress_bars.log(&message);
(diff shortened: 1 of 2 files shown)


michaelneale:
@jamadeo much nicer. I note in non interactive mode it outputs still a bit too - is that intentional? this change looks great though.

jamadeo:
Thanks @michaelneale, yeah, when I run non-interactive I tend to still observe the output so it made sense to keep it in there, but I could also see the argument against. WDYT?

PR #2845, opened 2025-06-10
fix: added url encoding and decoding for goose recipe url

lifeizhou-ap:
**Why**
Without the encoding/decode uri, the deeplink generated for recipe containing "```" cannot be decoded on Desktop side. It will show error.  [Example recipe](https://github.com/squareup/goose-recipes/blob/main/weekly-updates/recipe.yaml)

**What**
Added the uri encoding on Rust side and uri decoding on Desktop side to handle the deeplink for recipe.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3514,6 +3514,7 @@ dependencies = [
  "tracing",
  "tracing-appender",
  "tracing-subscriber",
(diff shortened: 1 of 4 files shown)


michaelneale (approved):
good catch

PR #2855, opened 2025-06-11
feat: (tool router) adds extension name in vector db & search tool

wendytang:
- allows goose to specify the extension name and only conduct vector search given extension name
- improves the recall and accuracy of tools selected

part of AAI-786


wendytang:
issue AAI-786

PR #2862, opened 2025-06-11
(experiment) sub-recipes prototype

jamadeo:
(not to merge)

This is a minimal prototype of sub-recipes. You can add a `subrecipes` field to your recipe, and Goose will add tool calls representing each sub-recipe. No parameters are supported in this prototype.

To get it working, cd into the `subrecipes-mcp` directory and run `uv tool install -e .`.

### DIFF
--- a/subrecipes-mcp/main.py
+++ b/subrecipes-mcp/main.py
@@ -1,0 +1,83 @@
+import argparse
+import asyncio
+import json
+from typing import Awaitable, Callable
+from pathlib import Path
+from fastmcp import FastMCP, Context
+from mcp import types
+
+server = FastMCP(
+    "Subrecipes 🚀",
+    instructions="This extension allows you to run subrecipes.",
+)
+
+del server._mcp_server.request_handlers[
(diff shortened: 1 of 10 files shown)


subrecipes-mcp/main.py
```diff
@@ -0,0 +1,83 @@
...
+    "Subrecipes 🚀",
+    instructions="This extension allows you to run subrecipes.",
+)
+
+del server._mcp_server.request_handlers[
+    types.ListResourcesRequest
+]  # Otherwise the Goose system prompt says this supports resources
+
+
+def runner(subrecipe: dict) -> tuple[str, Callable[[Context], Awaitable[str]]]:
```
jamadeo, line 19: @lifeizhou-ap my thinking is that we turn these into tools added and handled by Goose itself, without the MCP layer at all. Then we can start with an implementation where Goose still shells out to itself. And that leaves the door open to Goose running the subrecipe in its own separate agent in-process

jamadeo:
Work continues in #2982

PR #2866, opened 2025-06-11
feat: (tool router) llm tool selector

wendytang:
- Unlocks LLM tool selector so that selection of tools is not limited to vector search


michaelneale:
very cool @wendytang what is best way to try this out to see it in best light - will it add extra round trips to look out for?

wendytang:
@michaelneale there is a tool selection strategy section in settings : default -> llm

PR #2870, opened 2025-06-12
Lifei/fix goose recipe prompt

lifeizhou-ap:


ui/desktop/src/App.tsx
```diff
@@ -207,15 +206,10 @@ export default function App() {
...
-    (async () => {
-      try {
-        await initializeApp();
-        setAppInitialized(true);
-      } catch (error) {
-        console.error('Unhandled error in initialization:', error);
-        setFatalError(`${error instanceof Error ? error.message : 'Unknown error'}`);
-      }
-    })();
+    initializeApp().catch((error) => {
```
lifeizhou-ap, line 209: This is just reverting to the original implementation as `setAppInitialized` is not used anymore

PR #2874, opened 2025-06-12
cli cosmetic improvements

michaelneale:
been a while since the CLI changed, so here is a proposed new look: 

image


dehidehidehi:
I prefer a minimalistic look; is there a way to ensure users like me have a choice?

michaelneale:
@dehidehidehi of course - and I would like minimal too, do you have shell/experiences you love, as this is really jsut a showcase, we don't have to use this (if it is too much can pare it back) but I think we can do better than default. Are there anything how it is now that could be small but important changes?

blackgirlbytes:
I like this version of the context usage bar much better, but I miss the "( O)>"

I see what @dehidehidehi is saying..wonder if there's like an env variable we can set where we can say we want "vintage goose cli" or the new version. idk!

I like charm cli

cgwalters:
Regardless of the presentation, it'd make sense to factor out things like the unicode-width change as an independent "prep" PR right?

Issue #2848, opened 2025-06-10
Goose is a energy hog on mac


harkamals:
To reproduce
Launch goose, no nothing, it starts to consume most energy, activity monitor bar shoots up (> 80%). As soon as i close goose, things are back to normal.

**Please provide following information:**
 - **OS & Arch:** Mac OS 15.5
 - **Interface:** UI
 - **Version:** 1.0.27
 - **Provider & Model:** LiteLLM --> AzureDatabricks.net --> Claude Sonnet 3.7

michaelneale:
@harkamals can you share a bit more about activity monitor bar? (not sure what you meab by that) - you seeing anything in activity monitor for goose processes? I can see "goose helper (renderer)" up a bit high - is that what you see?

harkamals:
I missed to share screenshots earlier. And yes, the Mac starts to heat up, and fan blows hard, it's a M4.

This drop is when goose is killed.
![image](https://github.com/user-attachments/assets/5a118885-ad5f-432f-8396-d8e3e2497cb5)

![image](https://github.com/user-attachments/assets/44545eb7-ef48-4293-8f1e-5334ff82e0c1)

michaelneale:
woah yeah fans on an M4 (I have m1 pro, no fans, but I can see the CPU) - may have something to try shortly

michaelneale:
thanks @harkamals https://github.com/block/goose/pull/2877 - fix pending. Tested locally with a bundle and that seems better - if you watch that PR for a bundle you can download and try. 

really simple problem - circular dependency so it kept updating itself silently. 

FYI - to debug: I ran goose (like you did) and asked goose (in another window - actually cli in this case) to look at the CPU hogging renderer and to suggest what could cause it (and some back and forth narrowing the code and it suggested it once I pointed out nothing was happening) - which is cool to see! (would have taken me a long time otherwise). Bug was introduced as "prettier" suggested some dependency which was a problem (to satisfy the check)

harkamals:
thank you.

also, another issue, if you can help
https://github.com/block/goose/issues/2748

harkamals:
I have downloaded the fix.app, but it does not show any UI now, tried to quit and re-open, no luck.

Version 1.0.27-canary (1.0.27-canary)

PR #2877, opened 2025-06-12
fix: CPU hogging GUI

michaelneale:
This was due to a circular react dependency

fixes: https://github.com/block/goose/issues/2848

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,8 +45,19 @@ jobs:
             /usr/local/share/chromium \
             /usr/local/share/powershell \
             /usr/share/dotnet \
-            /usr/share/swift
-
+            /usr/share/swift \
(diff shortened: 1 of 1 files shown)


ui/desktop/src/components/ChatView.tsx
```diff
@@ -280,8 +280,9 @@ function ChatContent({
 
   // Update chat messages when they change and save to sessionStorage
   useEffect(() => {
-    setChat({ ...chat, messages });
-  }, [messages, setChat, chat]);
+    // @ts-expect-error - TypeScript being overly strict about the return type
```
zanesq, line 283: you can use `// eslint-disable-next-line react-hooks/exhaustive-deps` we use it in other places.. seems like we should just remove this linting requirement as it seems to causes issues more often than not

PR #2879, opened 2025-06-12
multi templates in recipe

lifeizhou-ap:


crates/goose-cli/src/recipes/multi_template.rs
```diff
@@ -0,0 +1,43 @@
+use std::collections::HashMap;
+
+pub fn preprocess_template(main_template: &str, templates: &HashMap<String, String>) -> String {
```
lifeizhou-ap, line 3: manually process {% include } before render the template via minininja, find the indent of the line and add the include file content

test_main.yaml
```diff
@@ -0,0 +1,22 @@
...
+  bundled: true
+
+prompt: |
+  generate a joke for today. 
+  {% include "test_child.yaml" %}
+  Find the theme of the joke from the above yaml file.
+  Mention the day on the first line then an empty line and then the joke. 
+  Don't just say the date, but figure out if there's any cultural significance, like national shelves day
+
+templates:
```
lifeizhou-ap, line 21: Added the templates field.

We still can remove this templates field, can change the code in the pre-process function

PR #2888, opened 2025-06-12
Damien/sagemaker tgi

damienrj:
This is to enable a Sagemaker provider that can leverage models hosted on AWS Sagemaker TGI containers. https://github.com/awslabs/llm-hosting-container/tree/main


Justfile
```diff
@@ -388,4 +388,23 @@ kotlin-example:
     cd bindings/kotlin/ && java \
       -Djna.library.path=$HOME/Development/goose/target/debug \
       -classpath "example.jar:libs/kotlin-stdlib-1.9.0.jar:libs/kotlinx-coroutines-core-jvm-1.7.3.jar:libs/jna-5.13.0.jar" \
+<<<<<<< Updated upstream
```
salman1993: looks like a conflict

crates/goose/src/providers/sagemaker_tgi.rs
salman1993, line 1: can you add a test here for this provider? https://github.com/block/goose/blob/37812fc76a425d6e5b0ef643b1025144e5a94c95/crates/goose/tests/providers.rs#L380-L40

crates/goose/src/providers/sagemaker_tgi.rs
```diff
@@ -0,0 +1,372 @@
...
+            endpoint_name,
+            model,
+        })
+    }
+
+    fn create_tgi_request(
+        &self,
+        system: &str,
+        messages: &[Message],
+        _tools: &[Tool],
```
salman1993: tools are skipped?

crates/goose/src/providers/sagemaker_tgi.rs
damienrj: sure

crates/goose/src/providers/sagemaker_tgi.rs
damienrj: good catch, some remnants of the code where were we debugging the code.

Issue #2892, opened 2025-06-12
LLM tool calls fail with JSON parsing errors due to unescaped control characters in arguments


bjulian5:
**Describe the bug**

LLM providers, particularly Amazon Bedrock models for Claude, occasionally output raw control characters (unescaped newlines, tabs, etc.) within JSON string values in tool call arguments. This causes JSON parsing failures and prevents tool execution, resulting in user-facing errors.

**To Reproduce**
1. Configure an Amazon Bedrock Claude model to generate tool calls with multi-line content
2. Convert the Bedrock response format to OpenAI specification (common practice for API standardization)
3. Process the converted response through Goose's OpenAI format handler
4. LLM generates tool call arguments containing literal newlines or other control characters during the conversion process
5. JSON parsing fails with error: "Could not interpret tool use parameters"

**Expected behavior**
Tool calls should execute successfully even when LLMs output control characters in JSON string values. The system should automatically escape these characters to produce valid JSON.

**Example of problematic LLM output after format conversion:**
```json
{
  "command": "echo 'Hello
World'"
}
```

**Expected processed output:**
```json
{
  "command": "echo 'Hello\\nWorld'"
}
```

**Screenshots**
N/A - This is a JSON parsing error that manifests in logs/error messages.

**Please provide following information:**
- **OS & Arch:** Multiple (affects all platforms)
- **Interface:** Both UI/CLI
- **Version:** Current development version
- **Extensions enabled:** Any extension that uses tool calls
- **Provider & Model:** Primarily Amazon Bedrock Claude models, but affects any LLM provider processed through format conversion

**Root Cause Analysis**
- Amazon Bedrock models for Claude are particularly prone to generating unescaped control characters in tool call JSON arguments
- When LLM responses are converted to OpenAI specification format for consistency, raw control characters from the original provider response are preserved
- Goose's response parsing logic processes these converted responses without preprocessing to handle malformed control characters
- Control characters like `\n`, `\r`, `\t` appear literally in the JSON instead of being properly escaped

**Impact**
- Tool calls fail unexpectedly, breaking user workflows
- Error messages are unclear about the root cause
- Affects reliability across multiple LLM providers due to format conversion processes
- Particularly impacts users of Amazon Bedrock Claude models

**Technical Context**
The issue occurs in the `response_to_message()` functions that process OpenAI-formatted responses:
- `crates/goose-llm/src/providers/formats/openai.rs`
- `crates/goose-llm/src/providers/formats/databricks.rs`
[...]

PR #2893, opened 2025-06-12
fix: escape control characters in LLM tool call arguments JSON

bjulian5:
- Add json_escape_control_chars_in_string() utility function
- Apply fix to OpenAI and Databricks response parsers
- Enhance error messages with raw and processed arguments
- Add comprehensive test coverage for control character escaping

### Testing
- ✅ Added unit tests for the new utility function covering various control character scenarios
- ✅ Updated existing tests to verify empty argument handling
- ✅ Manual testing with both OpenAI providers
- ✅ Verified that valid JSON arguments are unchanged
- ✅ Confirmed that structural JSON elements (quotes, backslashes) are preserved

### Backwards Compatibility
This change is fully backwards compatible. It only affects the processing of malformed JSON arguments that would have previously failed to parse. Valid JSON arguments continue to work exactly as before.

### Related Issues
Fixes https://github.com/block/goose/issues/2892

### DIFF
--- a/crates/goose/src/providers/formats/databricks.rs
+++ b/crates/goose/src/providers/formats/databricks.rs
@@ -3,8 +3,8 @@ use crate::model::ModelConfig;
 use crate::providers::base::Usage;
 use crate::providers::errors::ProviderError;
 use crate::providers::utils::{
-    convert_image, detect_image_path, is_valid_function_name, load_image_file,
-    sanitize_function_name, ImageFormat,
+    convert_image, detect_image_path, is_valid_function_name, json_escape_control_chars_in_string,
+    load_image_file, sanitize_function_name, ImageFormat,
 };
 use anyhow::{anyhow, Error};
 use mcp_core::ToolError;
@@ -322,14 +322,24 @@ pub fn response_to_message(response: Value) -> anyhow::Result<Message> {
                     .as_str()
                     .unwrap_or_default()
                     .to_string();
-                let mut arguments = tool_call["function"]["arguments"]
+
+                // Get the raw arguments string from the LLM.
+                let arguments_str = tool_call["function"]["arguments"]
                     .as_str()
                     .unwrap_or_default()
                     .to_string();
-                // If arguments is empty, we will have invalid json parsing error later.
-                if arguments.is_empty() {
-                    arguments = "{}".to_string();
-                }
+
+                // If arguments_str is empty, default to an empty JSON object string.
+                let arguments_str = if arguments_str.is_empty() {
+                    "{}".to_string()
+                } else {
+                    arguments_str
+                };
+
+                // Escape literal control characters in the arguments string to make it valid JSON.
+                // This handles cases where the LLM might output raw newlines or other control
+                // characters within string values in the JSON arguments.
+                let escaped_arguments = json_escape_control_chars_in_string(&arguments_str);
 
                 if !is_valid_function_name(&function_name) {
                     let error = ToolError::NotFound(format!(
@@ -338,7 +348,7 @@ pub fn response_to_message(response: Value) -> anyhow::Result<Message> {
                     ));
                     content.push(MessageContent::tool_request(id, Err(error)));
                 } else {
-                    match serde_json::from_str::<Value>(&arguments) {
+                    match serde_json::from_str::<Value>(&escaped_arguments) {
                         Ok(params) => {
                             content.push(MessageContent::tool_request(
                                 id,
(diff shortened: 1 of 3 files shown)


michaelneale:
looks nice - @salman1993 is active in there with goose-llm crate so probably good to get a blessing there.

bjulian5:
@salman1993 I ran into this error using the goose cli. I fixed in the `goose` crate and found that the exact same code was in `goose-llm` so added a fix there as well.

bjulian5:
@salman1993 bump 🙏

taniacryptid:
Let me bump this to the goose dev team for you!

angiejones:
I removed the changes to `crates/goose-llm`. @salman1993 can you take another look? other users are indicating this PR will fix their issues as well

crates/goose/src/providers/formats/databricks.rs
```diff
@@ -322,14 +322,24 @@ pub fn response_to_message(response: Value) -> anyhow::Result<Message> {
...
+                let arguments_str = if arguments_str.is_empty() {
+                    "{}".to_string()
+                } else {
+                    arguments_str
+                };
+
+                // Escape literal control characters in the arguments string to make it valid JSON.
+                // This handles cases where the LLM might output raw newlines or other control
+                // characters within string values in the JSON arguments.
+                let escaped_arguments = json_escape_control_chars_in_string(&arguments_str);
```
DOsinga: so what's happening here is that we receive a json document from the provider that almost certainly is correct json, but that the arguments element in it is doubly encoded json and that second tag is not always entirely correct?

if so I think we should slightly change the approach here - move the entire json parsing into utils and call it safely_parse_json and first try if it parses without your replacements and only if it doesn't do the escape replacements.

arguments could be something like `{"key1": "value1",\n"key2": "value"}` which contains a \n but is perfectly fine, while you are trying to fix `{"key1": "value1\n","key2": "value"}` which is not

bjulian5: @DOsinga updated the PR with your suggestions. PTAL @DOsinga  @salman1993 🙏

salman1993 (approved):

DOsinga:
can you do the DCO thing and we can get this in 🙏   https://github.com/block/goose/pull/2893/checks?check_run_id=46249519944

michaelneale:
oh this is a nice one, kudos

bjulian5:
@DOsinga done!

DOsinga:
thanks!

veriditin:
Hi! I've been tracking this issue and it's nice that most control characters are now properly escaped, but I think a few things were left on the table.

The quotation mark `"` can in MOST cases also be unambiguously escaped without breaking the JSON document. I also find my LLM of choice (qwen-3) makes the most mistakes in this category, so that would be nice to get fixed as well.

JSON is overly strictly defined in this sense. Escaping the quotation mark is only strictly necessary if the first non-white space character is a "meaningful" one, i.e. one of `,`, `:`, `]`, `}`. If it is not, we know for sure we are still inside a String type and can replace the `"` with `\"`. If it is followed by a meaning character (ignoring whitespace), we cannot know for sure so we should leave it alone.


`{"contents": "yeah, this is an awesome string one might say "super" cool"}` can be unambiguously escaped to: `{"contents": "yeah, this is an awesome string one might say \"super\" cool"}` as the first quotation mark within the contents value is followed by an `s` and the second one is followed by a `c`.

I additionally find that Qwen-3 will sometimes fail to properly close the JSON, i.e. to only use one `}` instead of `}}` if the final object is nested inside another object. I find it useful to keep track of the amount of opened objects and lists (in a stack / in order) and if the string ends with objects/list still open, to simply close them. So:

`{"name": "John", "age": 30, "city": "New York"` would be fixed to `{"name": "John", "age": 30, "city": "New York"}`

and combining the two:

`{"items": [{"name": "item1", "desc": "A "great" item"}, {"name": "item2"}` would be fixed to `{"items": [{"name": "item1", "desc": "A \"great\" item"}, {"name": "item2"}]}`

DOsinga:
good points, @veriditin - but at that point I think that search & replace with regexps is not going to do it and we're going to need a tolerant json parser - https://crates.io/crates/json_partial looks like an option? not sure it does your " in the middle of a string though

veriditin:
Looking at their examples as to what kind of errors can be fixed:

image

I think the `comma fixer` in this crate might clash with the quotation mark fix I'm suggesting (which in my experience is a more common problem than missing commas). So.. it seems this would need some careful consideration as to which types of problems happen significantly often and can be fixed independently of each other.. It also doesn't feel like this crate is making the best trade-offs for the goose/agentic use-case.

PR #2852, opened 2025-06-10
feat(desktop): Add auto-update functionality to Goose desktop app

jackjackbits:
This PR adds comprehensive auto-update functionality to the Goose desktop application, enabling users to easily stay up-to-date with the latest releases.

## Features Added:

- **Automatic Update Checking**: The app now checks for updates automatically on startup (with a 5-second delay to avoid blocking app launch)
- **Visual Update Indicator**: Menu bar icon changes to show an update indicator when updates are available
- **Dynamic Tray Menu**: Shows "Update Available..." option that opens the update settings when clicked
- **Update Settings UI**: New settings section for managing updates with:
  - Current version display
  - Check for updates button
  - Download progress indicator
  - Install update functionality
- **GitHub API Fallback**: When electron-updater fails (e.g., due to missing release assets), the app falls back to GitHub API for update checking
- **Auto-extraction**: Downloaded ZIP files are automatically extracted for easier installation
- **Manual Installation Flow**: For GitHub downloads, provides clear instructions for manual update installation

## Technical Changes:

- Added `electron-updater` for native update support
- Implemented custom GitHub updater as fallback mechanism
- Added new tray icon assets for update notifications
- Extended preload API to expose update functionality to renderer
- Added proper TypeScript types and error handling throughout

## Testing:

- ✅ Rust compilation and linting pass
- ✅ TypeScript type checking passes
- ✅ ESLint checks pass (no errors, no warnings)
- ✅ Prettier formatting applied
- ✅ Successfully builds on macOS ARM64

## What this PR adds:
- Auto-update functionality for the Goose desktop app
- Automatic update checking on startup (with 5s delay)
- Visual indicator in the menu bar when updates are available
- Dynamic tray menu that shows "Update Available..." when updates are found
- GitHub API fallback for update checking when electron-updater fails
- Auto-extraction of downloaded ZIP files
- Update settings section in the app UI
- Support for manual update installation flow

## Files changed:
- Added update-related utilities and components
- Modified main process to integrate auto-updater
- Added new UI components for update management
- Created custom icons for update notifications

PR #2894, opened 2025-06-12
fix: Revert downgrade to version number

tiensi:
Made this pr in case the [recent version downgrade](https://github.com/block/goose/pull/2852/files#diff-6b70a318d1ddd7ef0fc9132f732413f6282a0e411e233c05a2a4e60e83082cdfR2-R9) was unintentional.


baxen (approved):
good catch thank you!

PR #2897, opened 2025-06-13
Check for UPDATES_ENABLED flag before running update logic or in ui

zanesq:
Added new flag `UPDATES_ENABLED` and added to updater logic


baxen (approved):
LGTM

PR #2899, opened 2025-06-13
feat(ui): add chain-of-thought panel above assistant messages

chaitanyarahalkar:
### Summary  
Many open weight models support <think> to reveal the CoT process (while many of the providers hide it in the API).  This PR adds a collapsible “Show thinking” panel that surfaces the assistant’s Chain-of-Thought (anything wrapped in `<think> … </think>`) above every assistant message, while keeping the final answer unchanged. 

### UI Preview  

https://github.com/user-attachments/assets/0865ee53-8cde-42db-87f6-6f6ea148c4b4

### Key Changes  
1. `ui/desktop/src/components/GooseMessage.tsx`  
   • Parses message text for `<think>` blocks.  
   • Stores the visible reply and CoT separately.  
   • Renders a `<details>` element at the top of the message:  
     `summary` → “Show thinking”, body → rendered Markdown of the CoT.  
   • Leaves existing timestamp, copy-link, images, tool-call handling untouched.
2. Cosmetic Changes
• Mirrors UX patterns of other AI chat apps that transparently expose the model’s reasoning without cluttering the main conversation.  
• Keeps the default view concise; CoT is opt-in via a single click.  
• No back-end or protocol changes required—pure front-end enhancement.

### Testing  
* Tested with Qwen 3 which is a thinking model that shares its CoT with the <think> tag. Many other open weight CoT models like DeepSeek R1 also use the <think> tag.

### DIFF
--- a/ui/desktop/src/components/GooseMessage.tsx
+++ b/ui/desktop/src/components/GooseMessage.tsx
@@ -45,12 +45,29 @@ export default function GooseMessage({
   // Extract text content from the message
   let textContent = getTextContent(message);
 
-  // Extract image paths from the message
-  const imagePaths = extractImagePaths(textContent);
+  // Utility to split Chain-of-Thought (CoT) from the visible assistant response.
+  // If the text contains a <think>...</think> block, everything inside is treated as the
+  // CoT and removed from the user-visible text.
+  const splitChainOfThought = (text: string): { visibleText: string; cotText: string | null } => {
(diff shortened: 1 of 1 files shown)


michaelneale:
oh thanks - I like the look of this a lot, seems a good change - for non thinking ones I guess won't show think section?

michaelneale (approved):

PR #2914, opened 2025-06-13
Douwe/fix include for recipes

DOsinga:
This makes sure we can load templates and included files from the folder where the recipe lives. It also improves upon the error handling when reading recipes and adds a --render-recipe flag to print the rendered recipe for debugging. Finally it parses the parameters bit separately so we can also load yaml's that aren't correct yaml's before substitution


michaelneale (approved):
much needed

michaelneale:
merge when green

PR #2932, opened 2025-06-15
fix: improved GoosehintsModal error handling

The-Best-Codes:
## Before

The `.goosehints` configuration modal shows an error if the `.goosehints` doesn't already exist.

## After

The `.goosehints` configuration modal only shows an error if the file exists and there is an error, or there is an error reading the file.

### DIFF
--- a/ui/desktop/package-lock.json
+++ b/ui/desktop/package-lock.json
@@ -424,9 +424,9 @@
       }
     },
     "node_modules/@babel/compat-data": {
-      "version": "7.27.5",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.27.5.tgz",
-      "integrity": "sha512-KiRAp/VoJaWkkte84TvUd9qjdbZAdiqyvMxrGl1N6vzFogKmaLgoM3L1kgtLicp2HP5fBJS8JrZKLVIZGVJAVg==",
+      "version": "7.27.3",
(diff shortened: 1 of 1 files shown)


ui/desktop/package-lock.json
The-Best-Codes, line 1: I will revert this later, I had to `rm -rf node_modules package-lock.json` and re-install dependencies to resolve build errors.

ui/desktop/openapi.json
```diff
@@ -10,7 +10,7 @@
     "license": {
       "name": "Apache-2.0"
     },
-    "version": "1.0.27"
```
blackgirlbytes, line 13: did you mean to edit this part?

blackgirlbytes:
thank youuuuuu!! That error was bothering me!

ui/desktop/openapi.json
The-Best-Codes: @blackgirlbytes When I ran `just run-ui` it automatically did that. I'm guessing that the release process forgets to bump that maybe, and when I run the app it does bump it?

The-Best-Codes: The current latest version is 1.0.28 :)
https://github.com/block/goose/blob/af7918a79447792fa8a46ed1a692ee76e333deef/Cargo.toml#L7

blackgirlbytes: oh okay cool!

PR #2935, opened 2025-06-15
feat: add Help & Feedback section in App Settings

AaronGoldsmith:
I wasn't sure if folks would prefer this in the Updates section, but I just kept it separate for now.

Preview:
| Dark Mode | Light Mode |
|--------|--------|
| image| image |

### DIFF
--- a/ui/desktop/package.json
+++ b/ui/desktop/package.json
@@ -1,7 +1,7 @@
 {
   "name": "goose-app",
   "productName": "Goose",
-  "version": "1.0.28",
+  "version": "1.0.29",
   "description": "Goose App",
   "engines": {
     "node": "^22.9.0"
--- a/ui/desktop/package.json
+++ b/ui/desktop/package.json
@@ -1,7 +1,7 @@
 {
   "name": "goose-app",
   "productName": "Goose",
-  "version": "1.0.29",
+  "version": "1.0.28",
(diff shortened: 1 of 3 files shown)


ui/desktop/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "goose-app",
   "productName": "Goose",
-  "version": "1.0.28",
+  "version": "1.0.29",
```
blackgirlbytes: did you mean to add this part?

blackgirlbytes:
i like this but I'll let @opdich and others review!

ui/desktop/package.json
AaronGoldsmith: Yeah, my understanding is that we already released version [1.0.28](https://github.com/block/goose/releases/tag/v1.0.28), so this would need to go out with the next version.

angiejones: but the team will update that when they are ready to cut a release. let's leave it at 28

```suggestion
  "version": "1.0.28",
```

blackgirlbytes: hmm, i think we bump the version during the release

blackgirlbytes: I didnt refresh the page so I didn't see Angie's comment. Sorry about that

PR #2937, opened 2025-06-15
docs: blog update

angiejones:
adding mic as author
cleaning up a bit


michaelneale (approved):
nice - yeah looks to the point and good fixes.

PR #2941, opened 2025-06-16
feat(cli): add token cost estimation per provider and make configurable

chaitanyarahalkar:


michaelneale:
nice, LGTM

michaelneale:
(will need to update, run cargo fmt and clippy etc)

sambradbury:
This is similar to the changes here which add cost tracking the UI https://github.com/block/goose/pull/2992

That MR doesn't handle the CLI as yours does but querying for updated pricing would be better than having to hardcode pricing here.

patrickReiis:
Thank you very much 🙏🙏🙏🙏🙏

DOsinga (changes requested):
So yeah, since the other CL is merged and has actual price information rather than estimating it, I would suggest we use that infrastructure rather than adding our own estimates here

DOsinga:
closing this for now - let me know if you want to pick it up again

PR #2942, opened 2025-06-16
feat: Adding streamable-http transport support for backend, desktop and cli

btdeviant:
## Motivation
MCP is officially adopting streamable-https for client-to-server comms with optional sse for server-to-client comms. 

### Backend
This change provides support for streamable-http as a distinct transport option in the backend. The new transport provides support for optional sse server-to-client communication for backwards compatibility, but retains the legacy SSE transport type so as not to impact users who currently have configurations using this transport. 

This will allow for a cleaner, easier deprecation of the sse transport option in later releases.

### Desktop
Adds support for streamable-http as a custom connection with the ability to add and modify custom headers. Retains SSE for legacy support purposes.

### CLI 
Automatically detects streamable-http vs sse when using the `--with-remote-extension` flag


michaelneale:
@baxen this could work - if we didn't migrate over to the rust-sdk for MCP

btdeviant:
> @baxen this could work - if we didn't migrate over to the rust-sdk for MCP

@michaelneale @baxen - If it helps I can make some updates to this PR to ease that transition when the time comes! Perhaps an abstraction layer for the transport, config, etc w/ a feature flag? 

If you don't mind me asking, is there a plan to retain full backwards compatibility for users who currently have SSE as a distinct option, or would it be preferable to migrate and consolidate those who have sse config to streamable-http and use the native backwards compatibility?

btdeviant:
Hey @alexhancock , looks like a linting error that may be unrelated to my changes. Just wanted to put it on your radar - let me know if there's anything I can do on my end!

DOsinga:
> Hey @alexhancock , looks like a linting error that may be unrelated to my changes. Just wanted to put it on your radar - let me know if there's anything I can do on my end!

there was an issue like this a while ago, if you haven't recently merged in master, that might solve it

btdeviant:
> > Hey @alexhancock , looks like a linting error that may be unrelated to my changes. Just wanted to put it on your radar - let me know if there's anything I can do on my end!
> 
> there was an issue like this a while ago, if you haven't recently merged in master, that might solve it

Awesome, thanks @DOsinga - ran what I could in the ci workflow for both the crates and the ui on my local and all looks good (famous last words), all merged 🤞🏼

PR #2945, opened 2025-06-16
Marcelle/goose bench datasets

laanak08:
Adds the following as an optional `dataset` parameter on the `BenchEval` config-struct
```
pub struct BenchEvalDataset {
    pub path: PathBuf,
    pub prompt_column: String,
    pub system_prompt_column: Option<String>,
    pub tools_column: Option<String>,
    pub llm_output_column: String,
    pub output_dataset_file_name: String,
    pub debug_size: Option<isize>,
}
```
This parameter allows running many queries through multiple agents and recording their resultant responses, with tool-calls and all, then writing a new dataset-file with results. one file per model. ex. running bench with 2 models, will result in 2 dataset output files will all the same information and an output column that was authored by the model the dataset was run under, which will be reflected in the output-file name.


DOsinga:
this seems to have been in draft for a long time. did you want to try to land this or shall we close it?

Issue #1388, opened 2025-02-26
Execution error: error sending request for url (http://127.0.0.1:11434/v1/chat/completions)  Failed to configure provider: init chat completion request with tool did not succeed.   goose with  ollama not connect


PrinceSajjadHussain:
![Image](https://github.com/user-attachments/assets/962f01c5-a9e4-4ca4-b314-74eff78c4d16)

wendytang:
Use address 0.0.0.0 as the host instead of 127.0.0.1

`0.0.0.0:11434`

https://github.com/ollama/ollama/blob/main/docs/faq.md#how-do-i-configure-ollama-server

vladimai:
> Use address 0.0.0.0 as the host instead of 127.0.0.1
> 
> `0.0.0.0:11434`
> 
> https://github.com/ollama/ollama/blob/main/docs/faq.md#how-do-i-configure-ollama-server

Doesn't work for me, same error message, & for "localhost"

TheBestPessimist:
I also have the same problem on windows: I run `goose` in wsl as per instructios and i have ollama installed in windows directly. I can use ollama, for example from https://ollama-gui.vercel.app, but when goose calls the endpoint `http://localhost:11434/v1/chat/completions`, i get a 404 in the ollama logs.

![Image](https://github.com/user-attachments/assets/e050a6e9-a7b7-4342-bc86-ccb65f0ba0c8)

The 200s are from vercel app, then 404s from goose

![Image](https://github.com/user-attachments/assets/ea372be4-a9c4-4403-839e-6f0b07bffae8)

TheBestPessimist:
My problem was that wsl could not access ollama. As soon as i fixed that, goose worked as expected. help here: https://block.github.io/goose/docs/troubleshooting#connection-error-with-ollama-provider-on-wsl

rsb-23:
Does the issue still exist with latest version?

TheBestPessimist:
I stopped using goose as i couldn't make it work with intellij.

rsb-23:
I think its working for me. Here's the setup.
1. Ollama and goose in wsl2
2. MCP server plugin installed in jetbrain.
3. Qwen3:4b model (think mode is slow, but good accuracy)
4. Using Ubuntu/wsl terminal within jetbrain.

Just figuring out few more things.

@PrinceSajjadHussain able to connect to ollama?

taniacryptid:
> I stopped using goose as i couldn't make it work with intellij.

[...]

PR #1799, opened 2025-03-20
ci: pin github action to specific hash

oreparaz:
We do not want a compromised github action (or maintainer thereof) to take over the repo's secrets or integrity. This PR pins every action to a full length commit SHA.

This is the way recommended by github [0]. The following tool was used in this PR: https://github.com/mheap/pin-github-action

Example vulnerability that exploits this vector: https://github.com/advisories/GHSA-mrrh-fwg8-r2c3 (CVE-2025-30066).

[0] https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

PR #2958, opened 2025-06-17
Fixes : Workflow error on issue comment

rsb-23:
3 Workflow run failed after my comment on Issue #1388

### Failed workflows 
1. Bundle Intel Desktop App - [run](https://github.com/block/goose/actions/runs/15695847494)
2. Bundle Windows Desktop App - [run](https://github.com/block/goose/actions/runs/15695847497)
3. Build CLI - [run](https://github.com/block/goose/actions/runs/15695847496)

### Reason
1. Invalid .**bundle-desktop-windows.yml** file due to `ref`.
4. Workflow can't differentiate between issue and pr comment.

### DIFF
--- a/.github/workflows/bundle-desktop-windows.yml
+++ b/.github/workflows/bundle-desktop-windows.yml
@@ -12,15 +12,15 @@ on:
         required: false
         type: boolean
         default: false
+      ref:
+        type: string
+        required: false
+        default: 'refs/heads/main'
     secrets:
       WINDOWS_CERTIFICATE:
         required: false
       WINDOWS_CERTIFICATE_PASSWORD:
         required: false
-    ref:
-      type: string
-      required: false
-      default: 'refs/heads/main'
 
 jobs:
   build-desktop-windows:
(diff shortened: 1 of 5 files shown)


rsb-23:
As per my understanding, Pr-comment workflows are only intended to be run if `.command` is available in PR comment text.
But workflows trigger for issue comments as well and also when `.command` is not in text.

Should I make below changes too?
- `if: github.event.issue.pull_request && contains(github.event.comment.body, '.command')`
- pin versions to GitHub action instead of hash
- simplify Get PR head SHA

zanesq:
@rsb-23 thanks this needs to be fixed! Code looks good so far, yes pls feel free to add those changes as well!

rsb-23:
> - pin versions to GitHub action instead of hash
> - simplify Get PR head SHA

@zanesq These 2 changes cannot be done.
- Hash is used since #1799 for better security.
- Since we have dispatch trigger too, _get PR SHA_ can't be simplified.

rsb-23:
`ref issue` is already solved in 'main branch' and was causing conflict.
Hence these changes are reverted.

Issue #1844, opened 2025-03-25
Truncation bug


zakiali:
**Describe the bug**

During a long session where truncation is triggered, the UI session ends up having the following error `Failed to buffer the request body: length limit exceeded`. This is after truncation occurs successfully (as seen from the logs). This was assumed to be because the context is too long, but this seems to be a different error mode. 

**To Reproduce**

**Expected behavior**

**Screenshots**

**Please provide following information:**
 - **OS & Arch:** Mac
 - **Interface:** UI
 - **Version:** v1.0.15
 - **Extensions enabled:** Dev
 - **Provider & Model:** claude sonnet v2

**Additional context**

Developerayo:
For `Failed to buffer the request body: length limit exceeded` it's cos `axum DefaultBodyLimit` will not accept request bodies [larger than 2 MB](https://docs.rs/axum/latest/axum/extract/struct.DefaultBodyLimit.html), once it's exceeded when being sent to the server, you get the `length limit exceeded` error back

PR #2965, opened 2025-06-17
Increase req body limit

Developerayo:
Fixes [#1844](https://github.com/block/goose/issues/1844)


michaelneale (approved):
good catch

DOsinga:
If this is still relevant, let's see if we can get it in soon! Can you do the DCO thing, resolve the conflicts (sorry!) and then also merge in master (I think that will resolve the error)

Developerayo:
> If this is still relevant, let's see if we can get it in soon! Can you do the DCO thing, resolve the conflicts (sorry!) and then also merge in master (I think that will resolve the error)

resolved

DOsinga:
the checks are failing - can you have a look? probably just have to merge in master and push again

ryleu:
push_message is just missing a closing bracket

DOsinga:
ugh, now we need cargo fmt

michaelneale:
how did we merge this - has a failure?

crates/goose/src/message.rs
```diff
@@ -353,6 +353,7 @@ impl From<PromptMessage> for Message {
 pub struct Message {
     pub id: Option<String>,
     pub role: Role,
+    #[serde(default = "default_created")]
```
michaelneale, line 356: will this break openapi?

PR #2895, opened 2025-06-12
Mnovich/temporal foreground tasks

Kvadratni:
## Final iteration of the scheduler on temporal
full implementation includes all of the required features.
and ability to switch the scheduler engine

https://github.com/user-attachments/assets/090b931b-bc17-4e2e-ba71-b245ec8c9964

PR #2967, opened 2025-06-17
Fix window not showing for some users

zanesq:
Turns out the users that had the issue was from goose trying to launch from a nonexistent previous directory or file path that was crashing the server.

- Added code to check if we have a valid directory before launching goosed and fallback to home if not.
- Change so the main window is created and displayed immediately without waiting for network requests.
- Also removed unused unzip package from testing.


wendytang (approved):

zanesq:
.bundle

temporal-service/go.mod
```diff
@@ -3,34 +3,33 @@ module temporal-service
 go 1.23.0
 
 require (
-	go.temporal.io/api v1.44.1
```
zanesq, line 6: cherry picked @Kvadratni [fix](https://github.com/block/goose/pull/2895/commits/6fa7da2b97af59fbc6106ad493f5730e43cd1bd0) to get builds working again

Kvadratni (approved):

zanesq:
.bundle

zanesq:
.bundle

zanesq:
.bundle

Issue #2812, opened 2025-06-07
Uncommon characters such as '’' (U+2019) in system prompts can degrade LLM performance


ackalker:
**Describe the bug**

Files such as `crates/goose/src/prompts/plan.md` and 
`crates/goose/src/prompts/system_gpt_4.1.md` contain characters such as '’' (U+2019: RIGHT SINGLE QUOTATION MARK) and others, which are probably quite uncommon in LLM training data, cause uncommon tokenization, and so can negatively influence (however subtly) the quality of LLM output, especially when used in system prompts or instructions for tool use.

**To Reproduce**
1. In a (Bourne-like) shell, execute the following command:
   `$ grep --color='always' -r -P -n "[^\x00-\x7F]" .|less -R`
2. Note the lines with uncommon quotation characters (ignore emoji).

**Expected behavior**
Strictly ASCII characters in system prompts and tool instructions, especially when used as word prefixes or suffixes, except when using clearly separated emoji.

**Screenshots**

![Image](https://github.com/user-attachments/assets/d1c46cd8-df0f-4d0d-9ed8-0e5d6e39405c)
OpenAI online tokenizer output for text containing right quotation mark. Notice that the important word `user's` is split into two tokens, which can influence LLM attention.

![Image](https://github.com/user-attachments/assets/4e775dfb-6f88-4b07-8f39-ca4f8df8e0da)
OpenAI online tokenizer ourput for the same text, but this time containing a common `'` apostrophe. The important word `user's` is tokenized as a single token.

**Please provide following information:**
 - **OS & Arch:** [WSL Ubuntu 24.04.2 LTS x86]
 - **Interface:** [CLI]
 - **Version:** [v1.0.24]
 - **Extensions enabled:** [Developer Tools, Computer Controller, Memory]
 - **Provider & Model:** [Ollama - qwen3:14b-16k-ctx]

**Additional context**

N/A

DOsinga:
fixed in https://github.com/block/goose/pull/2972 so closing

PR #2972, opened 2025-06-17
[fix][small] Replaced goose prompt unicode quotations with ascii quotations

tiensi:
https://github.com/block/goose/issues/2812

Updating Goose default prompts to use standard ascii quotation marks instead of unicode quotation marks. It's a small change but every token counts.

| Before | After |
| --- | --- |
| Before | After

[open ai tokenizer ui](https://platform.openai.com/tokenizer)

### DIFF
--- a/crates/goose-bench/src/assets/kubernetes_swagger.json
+++ b/crates/goose-bench/src/assets/kubernetes_swagger.json
@@ -5684,7 +5684,7 @@
           "type": "object"
         },
         "volumeHandle": {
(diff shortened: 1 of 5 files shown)


yingjiehe-xyz (approved):

angiejones:
wow! TIL

angiejones:
and THANK YOU 🙏🏽

ackalker:
You're very welcome :-)

PR #2982, opened 2025-06-18
feat: created sub recipe tools

lifeizhou-ap:
**What**
- Added tools for sub recipes
- Fixed the recipe validation and rendering. Template handling are as below:
   - when running recipe `goose run --recipe...`
      1. parse the recipe content with the inheritance, and extract recipe parameters and template variables.  
      2. validate the params
      3. render the recipe  (final recipe) with params and inheritance  if any parameters are not provided, it will fail

   - when explaining recipe `goose run --recipe ... --explain`
     1  parse the recipe content with the inheritance, and extract recipe parameters and template variables.  
     2. validate the params
     3. render the recipe with params and inheritance  (if not all params are provided, it won't fail)

   - when validating or creating deep link for recipe `goose recipe deeplink ...`, `goose receipe validate ...`
      1 . parse the recipe content with the inheritance, and extract recipe parameters and template variables.  
      2.  validate the params
      3.  ender the recipe with inheritance but without any params


crates/goose-cli/src/recipes/recipe.rs
```diff
@@ -183,18 +181,6 @@ fn validate_optional_parameters(recipe: &Recipe) -> Result<()> {
     }
 }
 
-fn parse_recipe_content(content: &str) -> Result<Recipe> {
```
lifeizhou-ap, line 161: moved to goose package to be reusable

crates/goose/src/agents/agent.rs
```diff
@@ -50,6 +51,7 @@ use super::tool_execution::{ToolCallResult, CHAT_MODE_TOOL_SKIPPED_RESPONSE, DEC
 pub struct Agent {
     pub(super) provider: Mutex<Option<Arc<dyn Provider>>>,
     pub(super) extension_manager: Mutex<ExtensionManager>,
+    pub(super) sub_recipe_manager: Mutex<SubRecipeManager>,
```
lifeizhou-ap, line 56: I created a sub_recipe_manager to manage sub recipe tools and trigger the sub recipe tool call.

I could also use the extension manager to add these tool, but the extension manager is mainly for MCP. So I create a sub_recipe_manager, maybe later it could evolve to sub_agent_manager

crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs
```diff
@@ -0,0 +1,167 @@
...
+        Use this tool when you need to run a sub-recipe.
+        The sub recipe will be run with the provided parameters."
+            .to_string(),
+        input_schema,
+        Some(ToolAnnotations {
+            title: Some(format!("run sub recipe {}", sub_recipe.name)),
+            read_only_hint: true,
+            destructive_hint: false,
+            idempotent_hint: false,
+            open_world_hint: false,
```
jamadeo: do we know that open_world is false? should we just leave it out?

crates/goose/src/agents/agent.rs
jamadeo: This seems great. They are distinct from extensions so I think that makes sense. It also leaves the door open to agents treating sub-recipes differently: instead of shelling out to a new goose, a future version may use the same agent or a new in-process instance

crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs
```diff
@@ -0,0 +1,166 @@
...
+
+use anyhow::Result;
+use mcp_core::tool::{Tool, ToolAnnotations};
+use serde_json::{json, Map, Value};
+use tokio::io::{AsyncBufReadExt, BufReader};
+use tokio::process::Command;
+
+use crate::recipe::{Recipe, RecipeParameter, RecipeParameterRequirement, SubRecipe};
+
+pub const SUB_RECIPE_TOOL_NAME_PREFIX: &str = "subrecipe__run_sub_recipe";
```
jamadeo: I'm not sure how much it matters, but since the models themselves don't really know what a recipe is, you could probably just use `run__{recipe name}` or something like that to name the tools

crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs
```diff
@@ -0,0 +1,166 @@
...
+    let status = child
+        .wait()
+        .await
+        .map_err(|e| anyhow::anyhow!("Failed to wait for process: {}", e))?;
+
+    let stdout_output = stdout_task.await.unwrap();
+    let stderr_output = stderr_task.await.unwrap();
+
+    if status.success() {
+        Ok(stdout_output)
```
jamadeo, line 160: we should probably also return stderr to the model, right?

crates/goose/src/agents/sub_recipe_manager.rs
```diff
@@ -0,0 +1,69 @@
...
+        for sub_recipe in sub_recipes_to_add {
+            let sub_recipe_name = sub_recipe.name.clone();
+            let tool = create_sub_recipe_tool(&sub_recipe);
+            self.sub_recipe_tools.insert(sub_recipe_name.clone(), tool);
+            self.sub_recipes.insert(sub_recipe_name, sub_recipe);
+        }
+    }
+
+    pub fn is_sub_recipe_tool(&self, tool_name: &str) -> bool {
+        tool_name.starts_with(SUB_RECIPE_TOOL_NAME_PREFIX)
```
jamadeo: here you might as well just test for membership in the map, rather than the prefix

crates/goose/src/agents/sub_recipe_manager.rs
```diff
@@ -0,0 +1,69 @@
...
+            sub_recipe_tools: HashMap::new(),
+            sub_recipes: HashMap::new(),
+        }
+    }
+
+    pub fn add_sub_recipe_tools(&mut self, sub_recipes_to_add: Vec<SubRecipe>) {
+        for sub_recipe in sub_recipes_to_add {
+            let sub_recipe_name = sub_recipe.name.clone();
+            let tool = create_sub_recipe_tool(&sub_recipe);
+            self.sub_recipe_tools.insert(sub_recipe_name.clone(), tool);
```
jamadeo: if you instead store them in the map using their prefixed names, it will make them easier to look up in call_sub_recipe_tool() and is_sup_recipe_tool()

jamadeo (approved):
Very nice! I'll test this out and report back with any findings

crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs
lifeizhou-ap: good catch! I just used the default, will modify that

crates/goose-cli/src/recipes/recipe.rs
```diff
@@ -214,362 +210,5 @@ fn apply_values_to_parameters(
...
-fn validate_recipe_file_parameters(recipe_file_content: &str) -> Result<Recipe> {
-    let recipe_from_recipe_file: Recipe = parse_recipe_content(recipe_file_content)?;
-    let parameters = extract_parameters_from_content(recipe_file_content)?;
-    validate_optional_parameters(&parameters)?;
-    validate_parameters_in_template(&parameters, recipe_file_content)?;
-    Ok(recipe_from_recipe_file)
-}
-
 #[cfg(test)]
-mod tests {
```
lifeizhou-ap, line 255: tests are moved to a separate files

PR #2986, opened 2025-06-18
feat(ui): Add persistent draft state for chat input

alexrrouse:
This came out of me having to re-type messages after going into settings to make sure that I had the correct set of MCP's setup for my request, we now keep a draft of that text so that when we return it uses the draft text and prevents us from having to retype our prompts.

https://github.com/user-attachments/assets/cbfd02ed-91c1-4d4a-9063-8385c57a50e8


zanesq:
Tested this locally and command history (cmd + up arrow) is not working like it did before, it's only cycling through a few messages and it's not as snappy, it takes a few times to make it go back and forth through history. Can you take a look?

DOsinga:
given @zanesq comment and the age of this, going to close this for now. feel free to re-open when addressed

PR #2993, opened 2025-06-18
feat: Add developer environment doctor script

ethanchang235:
Adds a 'doctor.sh' script to help new contributors quickly verify their local development setup. This improves the project's onboarding experience.

### DIFF
--- a/scripts/doctor.sh
+++ b/scripts/doctor.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+echo "Running Goose development environment doctor..."
+echo "------------------------------------------------"
+all_ok=true
+
+# Check 1: Does the hermit binary exist and is it executable?
+if [ -x "bin/hermit" ]; then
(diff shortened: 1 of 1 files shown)


michaelneale:
if justfile is provided then the rest I think would be bootstrapped now by hermit?

ethanchang235:
Thank you for the feedback! 

I have pushed an updated script that now checks for Hermit activation instead. Apologies for the accidental close and reopen as well.

michaelneale:
thanks - I like it, could help some folks out

DOsinga (approved):

DOsinga:
Sorry for taking so long - we're going to do better. In order to land this, can you do two things:
* merge in main - I think the error we're seeing was something that is fixed now
* do the DCO check
Thanks!

DOsinga:
thanks @ethanchang235  - I'm assuming you no longer want this to go in, so closing this for now. feel free to reopen

PR #2995, opened 2025-06-18
fix(docker): install protoc to fix lance-encoding build

davidroman1914:
This PR adds `protobuf-compiler` to the Dockerfile so that the `lance-encoding` crate can compile successfully during the `cargo build` step.

Without this change, the Docker build fails with the following error:
```
Error: Custom { kind: NotFound, error: "Could not find protoc. If protoc is installed, try setting the PROTOC environment variable..." }
```

This fix installs the required `protoc` binary using `apt-get`, allowing Goose to build in Docker without error.

Tested on Apple M4 MacBook Pro with `docker-compose up --build`.

### DIFF
--- a/documentation/docs/docker/Dockerfile
+++ b/documentation/docs/docker/Dockerfile
@@ -16,6 +16,8 @@ WORKDIR /usr/src/goose
 # Copy the entire project
 COPY . .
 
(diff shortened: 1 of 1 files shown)


michaelneale:
nice one - yeah that will help, forgot about docker

PR #3002, opened 2025-06-19
Fix: Workflow syntax

rsb-23:
Syntax error in #2958 

**Fix:**
- Using single quote in `if block`


rsb-23:
@zanesq 
This is causing some failures in workflow runs.

michaelneale (approved):
wow, nice one.

rsb-23:
I think we can add `actionlint` linter also (in separate PR).

Issue #1220, opened 2025-02-13
View and edit config.yaml file in the UI


lily-de:
Given the decision to move all non-secret configuration data into `config.yaml`, it might make sense to allow UI users to look at it (and maybe edit?).

matthewdiamant:
I can take this on, @lily-de

PR #3005, opened 2025-06-19
feat: allow users view and edit their non-secret config's

Developerayo:
Fixes: #1220

- Allow users view and edit their non-secret configs
- Fix pages navigation `goBack`

image


lilydelalande:
Let's make the keys match what they literally are so there's no confusion e.g. `GOOSE_MODEL` not `Goose Model`

lilydelalande:
Thanks for the contribution! Can you rebase and then I can check it out / run your branch? Thank you!

DOsinga (approved):

Issue #3004, opened 2025-06-19
Goose on Windows: PowerShell Command Arguments Incorrectly Wrapped in Script Content


GaryZhous:
Goose fails to run powershell commands!

### Steps to reproduce the bug:
1. Open Goose desktop on your Windows computer.
2. Ask it to do some scripting such as list all the files in the current directory.
3. Tada you have the error

![Image](https://github.com/user-attachments/assets/cd2ac502-19d7-4a5a-8d1f-ee6e79b62089)

### Trouble-shooting
I let Goose self-diagnose this situation and it turns out that those PowerShell profile parameters are being incorrectly included in the script content instead of being passed as separate arguments. This causes all commands to fail with a "-NoProfile is not recognized as a cmdlet" error.

![Image](https://github.com/user-attachments/assets/fb74f379-a59f-482d-95cc-7b188343d3f5)

![Image](https://github.com/user-attachments/assets/4fe1f615-c440-4f79-be6b-15d8655f65af)

Therefore instead of this: `powershell -NoProfile -NonInteractive -Command "dir"`, we are getting this: `powershell "-NoProfile -NonInteractive -Command { dir }"`, those params are treated as command strings so PowerShell fails to execute the `dir` command properly.

### PS
I think this [code ](https://github.com/block/goose/blob/main/crates/goose-mcp/src/computercontroller/platform/windows.rs#L8) right here is causing the problem:

![Image](https://github.com/user-attachments/assets/38bd3804-7b29-4574-a4f3-43a669571de5)
Maybe the params are not set up properly?

Kvadratni:
Thanks for finding it! I was trying to debug it for a while! 
Now that we have it out to people I hope the community will help us squash these pesky issues faster!

GaryZhous:
@Kvadratni no worries! Thanks for you guys' hard work to get Goose on Windows!🙏

Kvadratni:
Now that I think of it, your screenshots show usage of the shell tool from the developer extension, but the code points to the computer controller....

GaryZhous:
Oops my bad the error should be here https://github.com/block/goose/blob/main/crates/goose-mcp/src/developer/shell.rs#L15

![Image](https://github.com/user-attachments/assets/544ee146-f3ff-436d-a3c7-386e522f76da)
Thanks for correcting me and providing fix!

Kvadratni:
fix merged

PR #3006, opened 2025-06-19
fix powershell executions

Kvadratni:
This should pass the parameters correctly to powershell 

image

reported in https://github.com/block/goose/issues/3004


Kvadratni:
.bundle-windows

Kvadratni:
.bundle-windows

michaelneale (approved):
nice!

PR #3009, opened 2025-06-19
fix shims for extensions on windows

Kvadratni:
puts the shims in the central location for windows so that we don't have to look up their path


michaelneale (approved):
again, nice detail!

PR #3011, opened 2025-06-19
Add `/extension` path for extension installation

joahg:
## Context

I want to add Goose as a recommended MCP Client in https://github.com/microsoft/playwright-mcp/pull/575

## Problem

GitHub markdown does not allow me use a custom protocol in the rendered README, which means that our `goose://` link will not work.

## Solution

We can add a new `/goose/extension` path that will redirect to the application and forward the query string to install the extension. With this change, playwright-mcp is installable into Goose using this URL: 

```
https://block.github.io/goose/extension?cmd=npx&arg=%40playwright%2Fmcp%40latest&id=playwright&name=Playwright&description=Interact%20with%20web%20pages%20through%20structured%20accessibility%20snapshots%20using%20Playwright
```

### DIFF
--- a/documentation/src/pages/extension.tsx
+++ b/documentation/src/pages/extension.tsx
@@ -1,0 +1,17 @@
+import React, { useEffect } from 'react';
+import { useLocation } from '@docusaurus/router';
+
+export default function ExtensionRedirect(): JSX.Element {
+  const location = useLocation();
+
+  useEffect(() => {
+    const params = new URLSearchParams(location.search);
+    window.location.href = `goose://extension${params.toString() ? '?' + params.toString() : ''}`;
+  }, [location]);
+
+  return (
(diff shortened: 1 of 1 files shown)


joahg:
As a related aside, Goose should have an SVG asset publicly available like this one from Cursor for times such as these:

michaelneale (approved):
oh this is awesome - yes lets get an install logo happening

michaelneale:
hey @joahg what about something like this: 


[🦢 Install in Goose](https://block.github.io/goose/extension?cmd=npx&arg=%40playwright%2Fmcp%40latest&id=playwright&name=Playwright&description=Interact%20with%20web%20pages%20through%20structured%20accessibility%20snapshots%20using%20Playwright)

but the link can be to this new url? that would be amazing

PR #3021, opened 2025-06-21
fix: updated openrouter known models

abromberg:
The OpenRouter known models are outdated. In the case of `google/gemini-2.5-pro-exp-03-25:free` it no longer exists on OpenRouter, so requests fail when users select this suggested model. In cases where the model still exists, there were most recent models available. 

No other changes are introduced beyond replacing the items in the model list.

### DIFF
--- a/crates/goose/src/providers/openrouter.rs
+++ b/crates/goose/src/providers/openrouter.rs
@@ -21,10 +21,11 @@ pub const OPENROUTER_MODEL_PREFIX_ANTHROPIC: &str = "anthropic";
 
 // OpenRouter can run many models, we suggest the default
 pub const OPENROUTER_KNOWN_MODELS: &[&str] = &[
+    "anthropic/claude-3.5-sonnet",
+    "anthropic/claude-3.7-sonnet",
     "anthropic/claude-sonnet-4",
-    "google/gemini-2.5-pro-preview"
-    "google/gemini-2.0-flash-001",
-    "deepseek/deepseek-r1-0528:free",
+    "google/gemini-2.5-pro"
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/openrouter.rs
```diff
@@ -21,10 +21,10 @@ pub const OPENROUTER_MODEL_PREFIX_ANTHROPIC: &str = "anthropic";
...
 pub const OPENROUTER_KNOWN_MODELS: &[&str] = &[
-    "anthropic/claude-3.5-sonnet",
-    "anthropic/claude-3.7-sonnet",
-    "google/gemini-2.5-pro-exp-03-25:free",
-    "deepseek/deepseek-r1",
+    "anthropic/claude-sonnet-4",
+    "google/gemini-2.5-pro-preview"
+    "google/gemini-2.0-flash-001",
+    "deepseek/deepseek-r1-0528:free",
 ];
```
blackgirlbytes: thanks for updating! I'm happy about adding some models, but some of these we dont need to remove. For example, yes let's remove "google/gemini-2.5-pro-exp-03-25:free"

But, we actually use anthropic/claude-sonnet-3.5 and anthropic/claude-sonnet-3.7 in our meetups and hackathons, so I would prefer to keep those included. You can still add claude-sonnet-4.

blackgirlbytes (changes requested):

blackgirlbytes (approved):
awesome! thank you!

abromberg:
sorry @blackgirlbytes just noticed a typo - missing comma - now fixed

PR #3031, opened 2025-06-23
fix: issue when rendering recipe parameters

lifeizhou-ap:


crates/goose-cli/src/recipes/recipe.rs
```diff
@@ -79,30 +79,10 @@ pub fn explain_recipe_with_parameters(
 }
 
 fn extract_parameters_from_content(content: &str) -> Result<Option<Vec<RecipeParameter>>> {
-    let lines = content.lines();
```
lifeizhou-ap, line 82: @DOsinga Is this block of logic is intentional? Just to confirm as I have deleted.

crates/goose-cli/src/recipes/recipe.rs
```diff
@@ -79,30 +79,10 @@ pub fn explain_recipe_with_parameters(
...
-                    .map_err(|e| anyhow::anyhow!("Failed to parse parameters block: {}", e))?;
-                return Ok(Some(parameters));
-            }
-            params_block.push_str(line);
-            params_block.push('\n');
-        }
-    }
-
     // If we didn't find a parameter block it might be because it is defined in json style or some such:
     if serde_yaml::from_str::<serde_yaml::Value>(content).is_err() {
```
lifeizhou-ap, line 83: @DOsinga do we still need below as the next block also parse the content to recipe?
`serde_yaml::from_str::<serde_yaml::Value>(content).is_err()`

also shall we parse both json and yaml file here?

DOsinga: any json file is also yaml, so no need to parse that twice

DOsinga (approved):

PR #3040, opened 2025-06-23
File diff and checkpoints (WIP)

Kvadratni:
## Implement checkpoint system with intelligent diff splitting

Adds file versioning and enhanced diff handling for better change management and review.

**Checkpoint system:**
- Automatic backup creation before file modifications with timestamp-based storage
- Persistent checkpoint index tracking file modification history  
- Restore functionality with diff generation for change tracking

**Enhanced diff generation:**
- File-type aware context sizing for better hunk boundaries
- Intelligent splitting of large diffs into manageable chunks
- Improved review experience for AI-generated changes

**New tools:**
- `list_checkpoints` - View available file versions
- `restore_checkpoint` - Revert files to previous states

**Integration:**
- Checkpoint payloads embedded in message stream for UI consumption
- Maintains backward compatibility with existing undo functionality
- Supports both traditional string replacement and Editor API workflows

Addresses large diff readability issues while providing robust file versioning capabilities.

https://github.com/user-attachments/assets/703ee4a4-0ec2-4cf8-bba3-79964af35b67

### DIFF
--- a/ui/desktop/src/App.tsx
+++ b/ui/desktop/src/App.tsx
@@ -2,23 +2,6 @@ import { useEffect, useRef, useState } from 'react';
 import { IpcRendererEvent } from 'electron';
 import { HashRouter, Routes, Route, useNavigate, useLocation } from 'react-router-dom';
 
-// Component to conditionally show sidecar based on route and chat state
-const ConditionalSidecarProvider = ({ children }: { children: React.ReactNode }) => {
-  const location = useLocation();
-  
-  // Only show sidecar on specific chat routes when there are messages
-  const isChatRoute = location.pathname === '/' || location.pathname === '/pair';
-
-  // We'll pass the chat state check to the SidecarProvider
-  // The sidecar will only show the collapsed panel when there are messages with potential actions
-  const showSidecar = isChatRoute;
-  
-  return (
-    <SidecarProvider showSidecar={showSidecar}>
-      {children}
-    </SidecarProvider>
-  );
-};
 import { openSharedSessionFromDeepLink, type SessionLinksViewOptions } from './sessionLinks';
 import { type SharedSessionDetails } from './sharedSessions';
 import { initializeSystem } from './utils/providerUtils';
@@ -45,7 +28,6 @@ import RecipesView from './components/RecipesView';
 import { useChat } from './hooks/useChat';
 import { AppLayout } from './components/Layout/AppLayout';
 import { ChatProvider } from './contexts/ChatContext';
-import { SidecarProvider } from './components/SidecarLayout';
 import { DraftProvider } from './contexts/DraftContext';
 
 import 'react-toastify/dist/ReactToastify.css';
@@ -63,6 +45,7 @@ import PermissionSettingsView from './components/settings/permission/PermissionS
 
 import { type SessionDetails } from './sessions';
 import ExtensionsView, { ExtensionsViewOptions } from './components/extensions/ExtensionsView';
+import { Recipe } from './recipe';
 // import ProjectsContainer from './components/projects/ProjectsContainer';
 
 export type View =
@@ -1337,9 +1320,7 @@ export default function App() {
               path="/"
               element={
                 <ChatProvider chat={chat} setChat={setChat} contextKey="hub">
-                  <ConditionalSidecarProvider>
-                    <AppLayout setIsGoosehintsModalOpen={setIsGoosehintsModalOpen} />
-                  </ConditionalSidecarProvider>
+                  <AppLayout setIsGoosehintsModalOpen={setIsGoosehintsModalOpen} />
                 </ChatProvider>
               }
             >
--- a/ui/desktop/src/components/ChatInput.tsx
+++ b/ui/desktop/src/components/ChatInput.tsx
@@ -1174,7 +1174,7 @@ export default function ChatInput({
         )}
 
         {/* Secondary actions and controls row below input */}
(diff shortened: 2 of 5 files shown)


crates/goose-mcp/src/developer/mod.rs
```diff
@@ -500,6 +566,308 @@ impl DeveloperRouter {
         }
     }
 
+    // Helper function to create a checkpoint of a file
+    fn create_checkpoint(&self, file: &Path) -> Result<(PathBuf, String), ToolError> {
```
JJSwigut, line 589: is this creating a full file for each checkpoint? Just thinking about storage efficiency, I wonder if you could just store just the diffs, mapped to the file

Kvadratni: yeah full file for now.

ui/desktop/src/hooks/docs/useWindowManager.md
```diff
@@ -0,0 +1,339 @@
+# useWindowManager Hook
```
DOsinga: what is this?

ui/desktop/src/hooks/useChatMessages.ts
```diff
@@ -0,0 +1,17 @@
+import { createContext, useContext } from 'react';
```
DOsinga, line 1: why do we need this?

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -1049,11 +1462,25 @@ impl DeveloperRouter {
...
+        });
+
         Ok(vec![
             Content::text(success_message).with_audience(vec![Role::Assistant]),
             Content::text(output)
                 .with_audience(vec![Role::User])
                 .with_priority(0.2),
+            Content::embedded_text(
+                "goose://checkpoint",
+                serde_json::to_string(&payload).unwrap(),
```
DOsinga, line 1481: so we're storing the diff here in the conversation and then skipping sending it to the LLM? that doesn't seem like the way to do this if we actually have the version history of a file.

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -344,6 +355,54 @@ impl DeveloperRouter {
             }),
         );
 
+        let list_checkpoints_tool = Tool::new(
```
DOsinga, line 358: does this work?

DOsinga (changes requested):
This definitely solves a problem worth solving; I'm not convinced though still that we should do our own version management rather than just relying on goose. commits are pretty cheap and can be squashed when needed.

crates/goose-mcp/src/developer/mod.rs
Kvadratni: diffs can be huge. I wanted to save context

crates/goose-mcp/src/developer/mod.rs
Kvadratni: absolutely does

ui/desktop/src/hooks/docs/useWindowManager.md
Kvadratni: this was added by @aharvard . I believe this is for future reference for all of the `sidecar` features taht can work there.

ui/desktop/src/hooks/useChatMessages.ts
Kvadratni: for the Revert link on the message. allows to gather all changes made since this messge, so that rollback would gather all the changes made to this point.

crates/goose-mcp/src/developer/mod.rs
DOsinga: Yes, I get that we don't want to send this to the LLMs. But as you point out correctly, diffs can be huge, so storing them in our fragile sessions doesn't seem great either. For one thing we now have a 2Mb size limit on sessions after which they will just dissapear

crates/goose-mcp/src/developer/mod.rs
DOsinga: I mean, when does the LLM call this?

ui/desktop/src/hooks/docs/useWindowManager.md
DOsinga: I don't think this is the way. Any documentation we require should be automatically extracted from the corresponding source

crates/goose-mcp/src/developer/mod.rs
Kvadratni: if I ask goose to restore file to a checkpoint

ui/desktop/src/hooks/docs/useWindowManager.md
aharvard: @Kvadratni, here's a PR to move the MD content to inline comments: https://github.com/block/goose/pull/3277

DOsinga:
I thought about this some more and I would like to get the diff bit in, but for the versioning I do think we want to use git. Can we split the functionality? i.e. change the file tools to produce a diff and then store that into the conversation as an attachment and show that to the user when they click open the tool result?

Kvadratni:
> I thought about this some more and I would like to get the diff bit in, but for the versioning I do think we want to use git. Can we split the functionality? i.e. change the file tools to produce a diff and then store that into the conversation as an attachment and show that to the user when they click open the tool result?

yes I will see what I can do, once I properly rebase on the latest re-design

Kvadratni:
replaced by https://github.com/block/goose/pull/3493

PR #3043, opened 2025-06-23
Fix session history scroll jumpy when virtual scrolling

zanesq:
Noticed session history was jumpy when scrolling. This fixes the jumpiness and also adds more efficient rendering and loading of session items. Also updated search placeholder to be more specific for session history search and added a note about cmd+F to search.


michaelneale (approved):

zanesq:
Closing since not needed in new upcoming ui

PR #3046, opened 2025-06-24
Explicit file context addition

jessejacksonafterpay:
This feature adds support for a structured way to define file and folder context for a session. The following has been implemented:
* Ability to add context paths from the existing paperclip menu item in the bottom left. Implemented as a menu to allow for other kinds of context to be wired up in the future.
* Also wired up the same file path addition to the drag and drop (which before this change, just dropped filenames in the chat input)
* Context paths are shown in lozenges inside the chat input and have an icon reflecting the type of item at the path (i.e. file or directory)
* A maximum of 10 paths are displayed, with an overflow menu if more paths are selected
* The file paths context is appended to underlying context sent to the LLMs. This is a structured message element, that is separated from the text content, but combined with text content when a message is sent.
* Context paths are saved with and reloaded when resuming a session


jessejacksonafterpay:
Thanks for the feedback @zanesq !

>The files aren't cleared when the chat message is sent.

What I was going for was something akin to the cursor-style context - which persists for the life of the session and is indicating that you want these things in context for the life of your session. I should be able to implement this in such a way that it doesn't re-send files that it has already sent previously (I think it may be doing that).

>Copying and pasting an image into the chat works differently

Ah thanks for pointing this out, hadn't used that functionality before - completely agree it should be consistent. I can certainly fold together the image and file functionality into the same underlying structure - but then need to make a decision. Either they get sent with the chat message (like images do today) or they remain sitting in the context.

 It's basically, do we want the ChatGPT style behaviour or the cursor style behaviour.


Edit: Thinking more on this, I think that maintaining the current behaviour, so that all files work like images work today makes the most sense (i.e. clear the list of files when sending). That way, this feature just becomes an extension of the image pasting functionality for all file types. I can combine the image pasting/preview functionality into this new structure, so that it works for all file types.

jessejacksonafterpay:
As per my train of thought above, I've updated the implementation to send the files into the chat when sending a message.

This is much closer to how ChatGPT works than how cursor works, but I think this makes more sense (for now).

spencrmartin:
Feels like we can pick up the cursor style "retain context" behaviour in a much more elegant way to fit what I think is the more accepted GPT paradigm. Thank you @jessejacksonafterpay

zanesq:
@jessejacksonafterpay the ui has changed a lot since this, do you want to pull in main and get it working again?

DOsinga:
This is a rather large conflict, but I have to say, I really like this feature

zanesq:
Closing due to inactivity and code drift, feel free to open a new one with the new UI in mind.

PR #2992, opened 2025-06-18
feat: Add comprehensive cost tracking display for LLM usage

jackjackbits:
# Cost Tracking Display Feature

## Overview
This PR implements a comprehensive pricing and cost tracking system for Goose, allowing users to see real-time costs for their LLM usage across different providers and models.

### 1. **Real-time Cost Display**
- Shows cost in the bottom bar with 6 decimal places for precision
- Updates dynamically as tokens are consumed
- Supports both local token estimation and backend token counts

### 2. **Multi-Model Session Tracking**
- Tracks costs across entire session even when switching models
- Accumulates costs per model with detailed breakdown in tooltip
- Preserves cost history when changing providers/models

### 3. **Pricing Data Management**
- Integrates with OpenRouter API for up-to-date pricing
- Caches pricing data locally with 7-day TTL
- Pre-loads pricing on server startup to avoid UI delays
- Memory-optimized to store only recently used models (last 20)

### 4. **Settings UI**
- New "Cost Tracking" section in settings
- Toggle to enable/disable cost display
- Shows OpenRouter connection status
- Manual refresh button for pricing data

### 5. **Backend Integration**
- New `/config/pricing` endpoint serves cached pricing data
- Pricing data persisted to disk for reliability
- Support for 300+ models from OpenRouter

### Backend (Rust)
- `pricing.rs` module handles OpenRouter API integration
- Disk caching in `.goose/pricing_cache.json`
- Pre-loads all pricing data on server startup
- Serves cached data to frontend for fast response

### Frontend (TypeScript)
- `CostTracker` component displays real-time costs
- `costDatabase.ts` manages pricing data with localStorage caching
- Memory optimization: only stores recently used models
- Session-wide cost accumulation logic

### Performance Optimizations
- Recently used models tracking (last 20 models)
- On-demand fetching for models not in cache
- Smart refresh only updates active models
- Cache cleanup removes unused data

## User Experience
- Accurate cost tracking across model switches
- Detailed cost breakdown per model in tooltip
- Clean monospace display prevents number jumping
- Respects user preference to hide/show costs

## Testing
- All TypeScript/ESLint checks pass
- Rust compilation successful
- Tested with multiple providers and model switching
- Verified cost accumulation across sessions

## Files Changed
- Backend: `crates/goose-server/src/routes/reply.rs`, `config.rs`, `pricing.rs`
- Frontend: `ChatView.tsx`, `CostTracker.tsx`, `costDatabase.ts`, `SettingsPage.tsx`
[...]

PR #3050, opened 2025-06-24
Extensions menu

jessejacksonafterpay:
Added a quick access extensions menu, to reduce the toil of going back and forth to settings to toggle extensions.

https://github.com/user-attachments/assets/27c59d61-f17c-49d9-8ede-252a9b4cd98b

### DIFF
--- a/ui/desktop/src/components/bottom_menu/BottomMenu.tsx
+++ b/ui/desktop/src/components/bottom_menu/BottomMenu.tsx
@@ -4,6 +4,7 @@ import { useToolCount } from '../alerts/useToolCount';
 import BottomMenuAlertPopover from './BottomMenuAlertPopover';
 import type { View, ViewOptions } from '../../App';
 import { BottomMenuModeSelection } from './BottomMenuModeSelection';
+import { BottomMenuExtensions } from './BottomMenuExtensions';
 import ModelsBottomBar from '../settings/models/bottom_bar/ModelsBottomBar';
 import { useConfig } from '../ConfigContext';
 import { useModelAndProvider } from '../ModelAndProviderContext';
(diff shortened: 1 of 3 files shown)


michaelneale:
I like this - getting a little crowded down there but I think for now extensions are important and often explicitly chosen  (until tool routing is default) so makes sense to me

blackgirlbytes:
I like this too. It's annoying for me to go back and forth and then the state resets for my input

zanesq:
@spencrmartin @nahiyankhan can you take a look at the ui?

lilydelalande:
https://github.com/block/goose/pull/2992 -- this is also an open PR that we want to get merged with additions to the bottom bar. @spencrmartin and @nahiyankhan ik you're already tagged here but just a heads up. Not sure if we want both these things as in in the bottom bar?

spencrmartin:
Screenshot 2025-06-25 at 10 46 12 AM

Thoughts around leveraging a count instead of having the entire copy within this space and a partnering hover

jessejacksonafterpay:
> Thoughts around leveraging a count instead of having the entire copy within this space and a partnering hover

I'd be supportive of this - it seems like we may have multiple features wanting to take advantage of this space, so we should look to consolidate on a design language for how this space is used.

spencrmartin:
Agreed, the pattern I think will follow current toolbar icon functionality and we'll rely on icons, hovers and status indicators to keep folks unblocked and not overcrowd the bottom bar section.

DOsinga:
do we want to get this in, or is this going to be solved with the new design?

DOsinga:
closing this for now. feel free to reopen if and when relevant

PR #3062, opened 2025-06-24
fix(temporal-service): makes the search for possible temporal paths parallel.

charlie-lobo:
The serialized nature of it meant that even a slightly larger than
expected delay could quickly add up to enough to trigger timeouts
consistently. Since these operations are highly IO bound there's no
reason not to run them all asynchronously.


Kvadratni (approved):
seems legit

PR #3079, opened 2025-06-25
feat: add voice dictation using OpenAI Whisper & ElevenLabs

jackjackbits:
# Voice Dictation Feature - PR Summary

## Overview
This PR adds voice dictation functionality to Goose Desktop, allowing users to input messages using their microphone with support for both OpenAI Whisper and ElevenLabs speech-to-text services.

### 1. Voice Input UI
- **Microphone button** in chat input area (next to send button)
- **Recording indicator** with duration and file size monitoring
- **Real-time waveform visualization** during recording
- **Visual feedback** for recording/transcribing states

### 2. Dual Provider Support
- **OpenAI Whisper**: Uses existing OpenAI API key, no additional configuration needed
- **ElevenLabs Speech-to-Text**: Alternative provider with advanced features
- **Smart provider switching**: Automatically available based on configured API keys

### 3. Settings & Configuration
- New **Voice Dictation** section in Settings
- Toggle to enable/disable the feature
- Provider selection dropdown
- ElevenLabs API key configuration with secure storage
- Provider-specific information and features

#### Backend (Rust)
- New `/audio/transcribe` endpoint for OpenAI Whisper
- New `/audio/transcribe/elevenlabs` endpoint for ElevenLabs
- `/audio/config` endpoint to check provider availability
- 25MB file size limit for both providers
- Support for multiple audio formats (webm, mp3, mp4, m4a, wav)
- Automatic API key migration to secure storage for ElevenLabs

#### Frontend (TypeScript)
- `useWhisper` hook for recording management
- `useDictationSettings` hook for settings persistence
- `WaveformVisualizer` component for audio feedback
- Microphone permission handling
- Real-time size and duration monitoring
- Automatic recording stop at 10 minutes or 25MB

### 5. Security & Privacy
- All API keys stored securely
- Audio data transmitted as base64 over HTTPS
- No audio stored locally after transcription
- Microphone permissions requested only when needed

### New Files
- `crates/goose-server/src/routes/audio.rs` - Audio transcription endpoints
- `ui/desktop/src/hooks/useWhisper.ts` - Recording and transcription logic
- `ui/desktop/src/hooks/useDictationSettings.ts` - Settings management
- `ui/desktop/src/components/settings/dictation/DictationSection.tsx` - Settings UI
- `ui/desktop/src/components/WaveformVisualizer.tsx` - Audio visualization

### Modified Files
- `ui/desktop/src/components/ChatInput.tsx` - Added microphone button
- `ui/desktop/src/components/settings/SettingsView.tsx` - Added dictation section
- `ui/desktop/src/main.ts` - Added microphone permission handling
- `ui/desktop/src/preload.ts` - Exposed permission APIs
- Various server files to register new routes

## Testing
- All Rust tests passing
- TypeScript compilation successful
- ESLint and formatting checks passed
- Manual testing completed with both providers

## Future Enhancements
- Real-time streaming transcription
- Custom vocabulary support
- Local Whisper model support
- Voice activity detection

## Breaking Changes
None - Feature is disabled by default and requires user opt-in.

## Screenshots

Screenshot 2025-06-25 at 22 08 56

Screenshot 2025-06-26 at 15 06 48


Kvadratni:
We all wanted this for a long time and attempted to implement it in different ways.
 But I have one reservation and this is tying us to one provider. 
 
 And I wonder if it's worth bringing it into the settings page where you can enable this feature and then select from the list of providers which will then ask you for a token or reuse a token. So then you're not tied into OpenAI, you have 11 abs and others in there too.

jackjackbits:
> We all wanted this for a long time and attempted to implement it in different ways.
> 
>  But I have one reservation and this is tying us to one provider. 
> 
>  
> 
>  And I wonder if it's worth bringing it into the settings page where you can enable this feature and then select from the list of providers which will then ask you for a token or reuse a token. So then you're not tied into OpenAI, you have 11 abs and others in there too.

hear you. i did it this way out of ease and piggybacking off something already configured as a bonus. i'm not sure it's really that better than local os dictation in the end. but putting in settings and maybe allowing local or other providers (i'm not sure who there is besides 11labs) would be a good approach. if local, it's effectively just a shortcut.

crates/goose-server/src/routes/audio.rs
```diff
@@ -0,0 +1,273 @@
...
+    let client = Client::builder()
+        .timeout(Duration::from_secs(OPENAI_TIMEOUT_SECONDS))
+        .build()
+        .map_err(|e| {
+            tracing::error!("Failed to create HTTP client: {}", e);
+            StatusCode::INTERNAL_SERVER_ERROR
+        })?;
+
+    let response = client
+        .post("https://api.openai.com/v1/audio/transcriptions")
```
AaronGoldsmith: Should we check/use the `OPENAI_HOST` from the environment? Similar to 

https://github.com/block/goose/blob/c3acddc3605f54b59cbfe2c7909237acc222724a/crates/goose/src/providers/openai.rs#L54-L56

hugomd:
Have we considered pulling in https://github.com/openai/whisper (or https://github.com/m-bain/whisperX) directly? 

It runs exceedingly well locally ✨

jackjackbits:
> Have we considered pulling in https://github.com/openai/whisper (or https://github.com/m-bain/whisperX) directly? 
> 
> 
> 
> It runs exceedingly well locally ✨ 

local macos is pretty good. don't know if it warrants increasing project/binary size?

jackjackbits:
> We all wanted this for a long time and attempted to implement it in different ways. But I have one reservation and this is tying us to one provider.
> 
> And I wonder if it's worth bringing it into the settings page where you can enable this feature and then select from the list of providers which will then ask you for a token or reuse a token. So then you're not tied into OpenAI, you have 11 abs and others in there too.

done

VOICE_DICTATION_PR.md
```diff
@@ -0,0 +1,80 @@
+# Voice Dictation Feature - PR Summary
```
baxen: nit: going to exclude this from the commit just to keep the top level of the code clean here - this is pretty well covered by inline comments

ui/desktop/src/main.ts
```diff
@@ -1020,6 +1024,65 @@ ipcMain.handle('get-quit-confirmation-state', () => {
   }
 });
 
+// Handle macOS dictation
```
baxen: nit: also planning to remove this, this looks like a leftover from a different attempt maybe, it appears unreachable to me at the moment. please let me know if i got that wrong!

baxen (approved):
Looks great!

VOICE_DICTATION_PR.md
jackjackbits: ah yes, forgot to remove that. sorry.

ui/desktop/src/main.ts
jackjackbits: oops! again you're right. left over. thank you.

QBlockQ:
@jackjackbits @Kvadratni 

Stunning! We just built something similar in our Meeting app:

Multi-provider STT (Whisper + ElevenLabs + Web Speech fallback) with all processing moved to secure backend routes. Currently testing, but early results show: no vendor lock-in, zero client-side API exposure, smart caching should cut redundant calls significantly.

Your settings-based provider selection is exactly right - gives users choice while keeping tokens secure. The "11 labs and others" flexibility is gold for production apps.

One insight from building: 3-layer fallbacks seem essential since STT services can be unreliable during peak times.

Solid work! 🔥

PR #3083, opened 2025-06-26
feat: new cli provider for claude code and gemini

michaelneale:
This lets claude code (if installed) be used as a provider itself. Claude code has a subscription model and is very popular, this allows you to use it if available as a provider in goose (goose will manage sessions etc, and can use other models in the mix if you want)

Usage: 

`export GOOSE_PROVIDER=claude-code`

or 

`export GOOSE_PROVIDER=gemini-cli`
 
to enable these providers, which will then make use of cli you have installed (and setup) on your system.


michaelneale:
hey @blackgirlbytes I think we can merge this once we have a new release

PR #3085, opened 2025-06-26
Add support for GOOSE.md

jessejacksonafterpay:
OpenAI uses `AGENTS.md`
Claude uses `CLAUDE.md`
Google uses `GEMINI.md`
Goose uses `.goosehints`

One of these things is not like the other 😄 This PR adds support for a `GOOSE.md` file as an alternative to `.goosehints` (which remains an option for backward compatibility).

### DIFF
--- a/crates/goose-mcp/src/developer/mod.rs
+++ b/crates/goose-mcp/src/developer/mod.rs
@@ -372,65 +372,56 @@ impl DeveloperRouter {
         // - macOS/Linux: ~/.config/goose/
         // - Windows:     ~\AppData\Roaming\Block\goose\config\
         // keep previous behavior of expanding ~/.config in case this fails
-        let global_hints_path = choose_app_strategy(crate::APP_STRATEGY.clone())
-            .map(|strategy| strategy.in_config_dir(".goosehints"))
-            .unwrap_or_else(|_| {
-                PathBuf::from(shellexpand::tilde("~/.config/goose/.goosehints").to_string())
-            });
-
-        // Check for global GOOSE.md file
-        let global_goose_md_path = choose_app_strategy(crate::APP_STRATEGY.clone())
-            .map(|strategy| strategy.in_config_dir("GOOSE.md"))
-            .unwrap_or_else(|_| {
-                PathBuf::from(shellexpand::tilde("~/.config/goose/GOOSE.md").to_string())
-            });
+        let get_global_path = |filename: &str| -> PathBuf {
+            choose_app_strategy(crate::APP_STRATEGY.clone())
+                .map(|strategy| strategy.in_config_dir(filename))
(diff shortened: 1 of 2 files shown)


qideng (approved):

pmarsh:
Like this idea. Makes it easier to understand coming from another tool. Seems minor but no downside

danielcorin:
From a DX perspective, It would be nice if agents collectively adopted at least support for a common standard even if they also want to support their own vanity names: [AGENT.md](https://agent.md/)

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -372,35 +372,56 @@ impl DeveloperRouter {
...
-        // Read global hints if they exist
         let mut hints = String::new();
-        if global_hints_path.is_file() {
-            if let Ok(global_hints) = std::fs::read_to_string(&global_hints_path) {
-                hints.push_str("\n### Global Hints\nThe developer extension includes some global hints that apply to all projects & directories.\n");
-                hints.push_str(&global_hints);
+
+        // Process each hint file type
+        for (filename, global_header, local_header) in hint_files.iter() {
+            // Check global file
```
DOsinga, line 401: can we avoid the duplication here by just having another loop over (global_path, global_header) (local_path, local_header)?

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -372,35 +372,56 @@ impl DeveloperRouter {
...
+
+        // Define hint files to check: (filename, global_header, local_header)
+        let hint_files = [
+            (".goosehints", 
+             "### Global Hints\nThe developer extension includes some global hints that apply to all projects & directories.\n",
+             "### Project Hints\nThe developer extension includes some hints for working on the project in this directory.\n"),
+            ("GOOSE.md", 
+             "### Global GOOSE.md\nThe developer extension includes global documentation from GOOSE.md that applies to all projects & directories.\n",
+             "### Project GOOSE.md\nThe developer extension includes project documentation from GOOSE.md for working on the project in this directory.\n"),
+        ];
```
DOsinga, line 395: can you extract some constants here to avoid the duplication of text

DOsinga: actually after some internal discussion, the verdict is dat we should go with AGENT.md instead. we could also look for hints for the competition, i..e look for AGENT.md, then .goosehints then claude, cursor, codex etc

michaelneale:
@jessejacksonafterpay we could take over this one on a branch if you like - I think consensus has been formed (see above^)

DOsinga:
@michaelneale - might be best to get this in the way we described with a sequence of AGENT.md etc

PR #3099, opened 2025-06-27
Add a reference for recipes

DOsinga:
had goose write up a reference page for the recipe documentation. it does help me to have all the fields in one overview


documentation/docs/guides/recipe-reference.md
```diff
@@ -0,0 +1,221 @@
+---
+sidebar_position: 18
+title: Creating Recipes
```
blackgirlbytes: Whatt do you think about calling it
```suggestion
title: Recipe Reference Guide
```

documentation/docs/guides/recipe-reference.md
```diff
@@ -0,0 +1,221 @@
...
+sidebar_label: Recipes
```
blackgirlbytes: ok with this?
```suggestion
sidebar_label: Recipe Reference Guide
```

documentation/docs/guides/recipe-reference.md
```diff
@@ -0,0 +1,221 @@
...
+---
+
+Recipes are YAML or JSON files that define how an AI agent should behave and what parameters it needs. They provide a structured way to create reusable AI interactions that can be shared and versioned.
```
blackgirlbytes: what you think of if we say this? that way we're more specific that this feature is unique to goose
```suggestion
Recipes are reusable Goose configurations that package up a specific setup so it can be easily shared and launched by others.
```

blackgirlbytes (approved):

PR #3114, opened 2025-06-27
Feature/global checkpoints UI

aharvard:
cleans up styles for #3040


ui/desktop/src/components/DiffViewer.tsx
aharvard, line 1: @Kvadratni we're not using this right? I was thinking it's ok to clean up.

PR #3118, opened 2025-06-27
fix: Make the entire toolcall argument row clickable to expand

opdich:
Make the entire ToolCallArgument row clickable to expand. This improves the UX and brings it inline with the other elements from ToolCallWithResponse.

### Enhancements to `ToolCallArguments` component:

* Updated the `ToolCallArguments` layout to make keys and truncated values clickable, enabling users to toggle expanded views directly by interacting with these elements.
* Adjusted the `flex` layout to ensure proper alignment of elements, improving visual consistency and usability.

### Refinements to `Expand` component:

* Added the `shrink-0` class to the `ChevronUp` icon to prevent resizing issues, ensuring the icon maintains its intended dimensions regardless of surrounding layout changes.


thomasmahoney:
thanks!!

PR #3145, opened 2025-06-28
feat(cli): Clear persisted session file with /clear command

### LINKED ISSUE #3138: /clear does not clear context
**Describe the bug**

With Goose CLI, using `/clear` to clear context doesn't appear to reset the context meter.

**To Reproduce**
1. launch `goose` cli
2. do any activity that uses tokens
3. enter `/clear`
4. see the `Chat context cleared.` response but context meter still remains no matter how many times `/clear` is run

**Expected behavior**
The context meter should be reset to 0%

**Screenshots**

Image

 - **OS & Arch:** Arch x86
 - **Interface:** CLI
 - **Version:** v1.0.30
 - **Extensions enabled:** Developer
 - **Provider & Model:** any model

**Additional context**

Jay4242:
**feat(cli): Clear persisted session file with /clear command**

The  command in the CLI now not only clears the in-memory chat history but also deletes and recreates the associated session file. This ensures that the chat context is fully reset, including any previously persisted messages, providing a more complete clear operation.

Possibly fixes #3138

Screenshot of behavior **including** the changes within this pull request:
image

Although it should possibly be noted, **without** these changes the current behavior seems to be that the memory shows as cleared the next time you interact with Goose.  The session jsonl file remains until the next interaction with Goose.  If the session is `/quit` immediately after a `/clear` the session is **not** cleared.

Example screenshot **without** the changes in this pull request  (**existing** Goose behavior):
image

Thanks again for your work on this great project!

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -544,6 +544,10 @@ impl Session {
                         &Message::assistant().with_text("Chat context cleared."),
                         self.debug,
                     );
+                    if self.session_file.exists() {
+                        std::fs::remove_file(&self.session_file)?;
+                        std::fs::File::create(&self.session_file)?;
+                    }
                     continue;
                 }
                 input::InputResult::PromptCommand(opts) => {


DOsinga (approved):
Thanks for doing this and sorry for the delay in responding!

PR #3159, opened 2025-06-30
Run recipes from remote urls

DOsinga:
simple feature so we can load recipes also from urls


lifeizhou-ap (approved):

DOsinga:
going to close this until we have better security in place

Issue #3153, opened 2025-06-29
MCP `npx` and `uvx` based extensions still broken


SalvatoreT:
**Describe the bug**

`npx` and `uvx` based extensions are unusable after @Kvadratni's change (https://github.com/block/goose/pull/3148) to address the `npx` missing issue (https://github.com/block/goose/issues/3139).

**To Reproduce**
1. On `main` run the `just run-ui` command.
2. Add the Knowledge Graph Settings extension (or any MCP extension)
    ```
    npx -y @modelcontextprotocol/server-memory
    ```
3. Observe the error

**Expected behavior**
The extensions should start.

**Screenshots**
![Image](https://github.com/user-attachments/assets/603bb3d7-d76a-455d-9b78-e49e2284b81b)

**Please provide following information:**
 - **OS & Arch:** macOS, ARM64
 - **Interface:** UI
 - **Version:** `main`, v1.0.30, `fdafbca92e88697aadb7deaa7b746a0f8ac13c41`
 - **Extensions enabled:** Computer Controller, Developer, Memory
 - **Provider & Model:** Google - gemini-2.5-pro-preview-06-05

**Additional context**

Here's the error message.

```
Error activating extension: Failed to add extension configuration, error: Transport(StdioProcessError("Permission denied (os error 13)"))
```

SalvatoreT:
If you look at the permissions, the `npx`, `uvx`, and `jbang` are different.

```
❯ ls -al ui/desktop/src/bin
total 659320
drwxr-xr-x@  9 sal  staff        288 Jun 29 22:50 .
drwxr-xr-x@ 34 sal  staff       1088 Jun 29 22:50 ..
-rw-r--r--@  1 sal  staff          0 Feb 23 23:45 .gitkeep
-rwxr-xr-x@  1 sal  staff  159786000 Jun 29 22:44 goose
-rwxr-xr-x@  1 sal  staff  150085824 Jun 29 22:44 goosed
-rw-r--r--@  1 sal  staff       3300 Jun 29 22:50 jbang
-rw-r--r--@  1 sal  staff          0 Jun 29 22:50 npx
-rwxr-xr-x@  1 sal  staff   27687586 Jun 29 22:42 temporal-service
-rw-r--r--@  1 sal  staff          0 Jun 29 22:50 uvx
```

SalvatoreT:
After I fix the permissions...

```
chmod +x ui/desktop/src/bin/{npx,uvx,jbang}
```

[...]

PR #3166, opened 2025-06-30
fix: Use default `{npx,uvx,jbang}` executable path

SalvatoreT:
Because we're executing on the server, we shouldn't use the binaries from the client.

Closes https://github.com/block/goose/issues/3153


SalvatoreT:
@Kvadratni went a different route to fix this. https://github.com/block/goose/pull/3170

PR #3176, opened 2025-06-30
allow to use dev/null for no-session mode

Kvadratni:
bug in question
image

### DIFF
--- a/crates/goose/src/session/storage.rs
+++ b/crates/goose/src/session/storage.rs
@@ -158,6 +158,14 @@ pub fn get_path(id: Identifier) -> Result<PathBuf> {
             session_dir.join(format!("{}.jsonl", name))
         }
         Identifier::Path(path) => {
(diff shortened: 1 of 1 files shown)


ahau-square (approved):

crates/goose/src/session/storage.rs
```diff
@@ -158,6 +158,14 @@ pub fn get_path(id: Identifier) -> Result<PathBuf> {
             session_dir.join(format!("{}.jsonl", name))
         }
         Identifier::Path(path) => {
+            // Allow special paths for no-session mode
+            if let Some(path_str) = path.to_str() {
+                if path_str == "/dev/null" || path_str == "NUL" {
```
DOsinga, line 163: why would it be NUL? that doesn't sound right

Kvadratni: windows
image

PR #3186, opened 2025-07-01
feat: add dynamic 'prompt' menu item (prompt history, send prompt)

Jud:
The default shortcuts for prompt history navigation conflict with the default MacOS shortcuts for navigation within a text area (e.g. cmd+up typically takes you to the beginning of a text input). To enable key remaping on MacOS, we need to have menu items that correspond to the history navigation actions.

There are other additions to this Prompt menu (like Attach that would be fast follows). This menuitem would also be an interesting jumping off point for prompt subflows like using a low-cost model to refine a prompt for a higher cost model.

This PR adds a new 'Prompt' menu to the top menu bar with the following items:
- Previous Prompt
- Next Prompt
- Send

These are dynamically enabled based on the current state.

Screenshot 2025-06-30 at 8 03 16 PM

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -1649,11 +1649,6 @@ const setupMenuModifications = () => {
     },
   };
 
-  template.push({
-    label: 'Prompt',
-    submenu: [previousPromptItem, nextPromptItem, { type: 'separator' }, sendPromptItem],
-  });
-
   const viewSubmenu: MenuItemConstructorOptions[] = [
     { role: 'reload' },
     { role: 'forceReload' },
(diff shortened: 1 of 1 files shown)


spencrmartin:
Hey @Jud, pulled and taking a look 

We may have introduced a fix for this outlined behaviour -- re avoid conflicting with the typical mac keyboard commands by only filling prompts via cmd up and down when the entry section is empty. When a user inputs their crafted prompt the cmd arrow to navigate via prompts is disabled and follow the typical keyboard behaviour (#3246)

I'm struggling to understand the use case in its current form outside of adding a top bar segment for user to leverage the keyboard shortcuts/ navigating entries. Does your proposed override the behaviour outlined in #3246?

DOsinga:
I'm going to close this for now based on @spencrmartin comment. feel free to reopen if you want to continue the conversation

PR #3188, opened 2025-07-01
feat: Structured output for recipes

jsibbison-square:
**Main Changes**
- Adds a new `response.json_schema` key in recipe yaml.
- When a json schema is described here then the agent will call a tool with its final output that matches that schema (otherwise a validation error will prompt the agent to correct its tool use).
- The final message to the user will be a single line json of this final output, making it easier to extract it (its just the last line)
- This is only applicable to `goose run` with recipes currently.

**Other Details**
- The response.json_schema is validated using [jsonschema](https://github.com/Stranger6667/jsonschema) crate
- If the agent tries to finish and it has not called the final tool for its output we send the agent a new message directing it to use the `final_tool` to output the expected message

This PR does not add support to use structured output when running a recipe that defines it via the Desktop app. Future PR.

**Example**
_Recipe_
```yaml
version: 1.0.0
title: "structured output"
description: "Create a structured output"
prompt: "Begin. We have the items: Fish, Spoon and Dog."
response:
  json_schema:
    type: object
    properties:
      total:
        type: number
        description: "The total number of items"
      num_characters:
        type: number
        description: "The number of characters in the items"
      first_item:
        type: string
        description: "The first item"
      last_item:
        type: string
        description: "The last item"
      items:
        type: array
        items:
          type: string
          description: "Each of the items"
    required:
      - total
      - num_characters
      - first_item
      - last_item
      - items
```
_Usage_
```bash
Loading recipe: structured output
Description: Create a structured output

starting session | provider: databricks model: databricks-claude-3-7-sonnet
    logging to /.../.local/share/goose/sessions/20250701_092723.jsonl
    working directory: /.../Development/goose/structured-output
I'll help you process these items and provide the structured output according to the required schema.

Let me analyze your items:
- Fish
- Spoon
- Dog

Now I'll prepare the JSON output with the required fields:
1. `total`: The total number of items (3)
2. `num_characters`: The total number of characters in all items (12 - "Fish" has 4, "Spoon" has 5, and "Dog" has 3)
3. `first_item`: The first item in the list ("Fish")
4. `last_item`: The last item in the list ("Dog")
5. `items`: An array containing all items (["Fish", "Spoon", "Dog"])

Let me submit this as the final output:
─── final_output |  ──────────────────────────
final_output: ...


◇  Goose would like to call the above tool, do you allow?
│  Allow
│


{"first_item":"Fish","items":["Fish","Spoon","Dog"],"last_item":"Dog","num_characters":12,"total":3}
```


crates/goose/src/agents/agent.rs
```diff
@@ -258,6 +270,20 @@ impl Agent {
             return (request_id, Ok(ToolCallResult::from(result)));
         }
 
+        if tool_call.name == FINAL_OUTPUT_TOOL_NAME {
```
DOsinga, line 273: oof this function is a bit of a mess! your addition is of course in line with what the rest looks like, but we should fix this at some point so if you feel the itch ...

crates/goose/src/recipe/mod.rs
```diff
@@ -119,6 +125,12 @@ pub struct Settings {
     pub temperature: Option<f32>,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Response {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub json_schema: Option<serde_json::Value>,
+}
```
DOsinga, line 132: do you have a strong preference about making this nested? in the ticket we talk about just having a `output_json_schema` field

crates/goose/src/agents/final_output_tool.rs
```diff
@@ -0,0 +1,468 @@
...
+            json!({
+                "type": "object",
+                "properties": {
+                    "final_output": {
+                        "type": "string",
+                        "description": "The JSON string final output to validate and collect"
+                    }
+                },
+                "required": ["final_output"]
+            }),
```
DOsinga: couldn't we just insert the json schema at this point? asking the llm to do it in this format probably works, but since the tool takes a json schema, why not use that?

crates/goose/src/agents/final_output_tool.rs
```diff
@@ -0,0 +1,468 @@
...
+    async fn validate_json_output(&self, output: &str) -> Result<Value, String> {
+        // First, try to parse the output as JSON
+        let parsed_value: Value = match serde_json::from_str(output) {
+            Ok(value) => value,
+            Err(e) => {
+                return Err(format!(
+                    "Invalid JSON format: {}\n\nExpected format:\n{}\n\nPlease provide valid JSON that matches the expected schema.",
+                    e,
+                    serde_json::to_string_pretty(self.response.json_schema.as_ref().unwrap()).unwrap_or_else(|_| "Invalid schema".to_string())
+                ));
```
DOsinga: does it recover from this?

crates/goose/src/agents/final_output_tool.rs
```diff
@@ -0,0 +1,468 @@
...
+        }
+    }
+
+    // Formats the parsed JSON as a single line string so its easy to extract from the output
+    fn parsed_final_output_string(parsed_json: Value) -> String {
+        serde_json::to_string(&parsed_json).unwrap()
+    }
+}
+
+#[cfg(test)]
```
DOsinga, line 142: there are a lot of tests here. I would suggest to remove the ones that will never find a bug. the tests in the recipe mod feel more useful

crates/goose/src/recipe/mod.rs
jsibbison-square: yeah I think we should leave ourselves open to alternative response formats and extra controls (I expect someone will ask for a 'max_attempts' option to prevent unlimited attempts at some stage... I'm building a MAX_TURNS option into goose atm which I think is a better version).

crates/goose/src/agents/final_output_tool.rs
jsibbison-square: yes. I managed to fudge the agent to do it wrong the first time to test with goose. I've implemented this pattern a few times and it has worked pretty well.

crates/goose/src/agents/final_output_tool.rs
jsibbison-square: Old TDD habits 🙃 . Removing some.

crates/goose/src/agents/final_output_tool.rs
jsibbison-square: 🔥 Very nice, that cleans up the code a bit too.

PR #3190, opened 2025-07-01
Use command line to run sub agent and sub recipe (in sequence or parallel)

lifeizhou-ap:
This PR is using goose command line to run task in subagent or sub-recipe.

**What**
- Change to sub-recipe tool to only create sub recipe task instead of run sub-recipes. 

- Create `sub_agent__execute_task` tool to run a single task or multiple task in parallel

- Task can be 2 types
   - Sub recipe task.  This task is created by sub recipe create task tool function and the task has attributes:
      - name (recipe name) 
      - recipe_path (recipe file path)
      - command_parameters (param name and value pairs to be passed to the `goose run recipe` command line)
      This task is executed as `goose run --recipe recipe_file --params name=value`
     
   - Sub agent task. This task is created by the main agent and has attributes.  At the moment it only has text_instruction, I think later on it will have more attributes similar to recipe.  I would like to discuss the use cases especially the model, tools (maybe some smart way to work out the tool and model based on the task?) 
      - text_instruction
    
- `sub_agent__execute_task` tool
     - it uses  `goose run --recipe recipe_file --params name=value` to run sub-recipe task
     - it uses `goose run --text text_instruction` to run sub-agent task
     - if it executes multiple tasks in parallel, it will create a worker pool and task queue to execute the tasks 
  
  TODO (won't be implemented in the PR)
   - max turn
   - error handling
   - maybe some special logic for sub agent such as exclude some specific tools like platform tools.

**Note**
- This PR defines the interfaces for the sub agent and sub recipe execution.  The implementation of execute_task by Goose cli will be changed.  We will use something like `subagent.run(task)` with a thread instead of a process. 

- Currently I use this sub agent execution tool name as `sub_recipe...` and with the description only for sub recipe.  In this case we still can keep the same feature behaviour.   When the subagent.run(task) is ready, we can rename the sub agent execution tool name back to  `sub_agent...` and add the `sub agent` in the description


crates/goose/src/agents/sub_agent_execution_tool/tasks.rs
```diff
@@ -0,0 +1,88 @@
...
+            data: None,
+            error: Some("Task timeout".to_string()),
+        },
+    }
+}
+
+async fn execute_task(task: Task) -> Result<Value, String> {
+    println!("=======Executing task: {:?}", task);
+
+    let mut command = if task.task_type == "sub_recipe" {
```
lifeizhou-ap: later on we could swap the implementation from use cmd to something `sub_agent.run(task)`

crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs
```diff
@@ -30,6 +33,22 @@ pub fn create_sub_recipe_tool(sub_recipe: &SubRecipe) -> Tool {
     )
 }
 
+pub fn create_sub_recipe_task_tool(sub_recipe: &SubRecipe) -> Tool {
+    let input_schema = get_input_schema(sub_recipe).unwrap();
+    Tool::new(
+        format!("{}_{}", SUB_RECIPE_TASK_TOOL_NAME_PREFIX, sub_recipe.name),
+        "Before running this sub recipe, you should first create a task with this tool and then pass the task to the task executor".to_string(),
```
wendytang, line 16: shall we add the recipe description into the tool description?

lifeizhou-ap: Hi @wendytang This is a good idea!

However, I will hold on this change because at this point, the subrecipe does not have the content of the sub recipe file, so it does not know the subrecipe description.  Parsing sub recipe file logic is currently in the goose-cli, we are going to move it into the goose package or parsing it before running sub-agent, and then we can get the description.  

At the moment, without the sub recipe description in the tool description, it works fine.  But definitely we will add this in when the parsing recipe file logic moves to goose package

crates/goose/src/agents/sub_agent_execution_tool/sub_agent_execute_task_tool.rs
```diff
@@ -0,0 +1,96 @@
+use mcp_core::{tool::ToolAnnotations, Content, Tool, ToolError};
+use serde_json::Value;
+
+use crate::agents::{sub_agent_execution_tool::lib::execute_tasks, tool_execution::ToolCallResult};
+
+pub const SUB_AGENT_EXECUTE_TASK_TOOL_NAME: &str = "sub_recipe__execute_task";
```
lifeizhou-ap: At the moment, only name with `sub_recipe` so that it does not conflict with the alpha feature.   Later on, sub agent should be executed with this tool. so it will be renamed to `sub_agent,,,`

crates/goose/src/agents/sub_agent_execution_tool/sub_agent_execute_task_tool.rs
```diff
@@ -0,0 +1,96 @@
...
+pub fn create_sub_agent_execute_task_tool() -> Tool {
+    Tool::new(
+        SUB_AGENT_EXECUTE_TASK_TOOL_NAME,
+        "Only use this tool when you want to execute sub recipe task. **DO NOT** use this tool when you want to execute sub agent task.
```
lifeizhou-ap: "DO NOT use this tool when you want to execute sub agent task.  " => same reason as above

crates/goose/src/agents/sub_agent_execution_tool/sub_agent_execute_task_tool.rs
wendytang: could we rename `SUB_AGENT_EXECUTE_TASK_TOOL_NAME` to `SUB_RECIPE_EXECUTE_TASK_TOOL_NAME`?

crates/goose/src/agents/sub_agent_execution_tool/sub_agent_execute_task_tool.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use crate::agents::{sub_agent_execution_tool::lib::execute_tasks, tool_execution::ToolCallResult};
+
+pub const SUB_AGENT_EXECUTE_TASK_TOOL_NAME: &str = "sub_recipe__execute_task";
+pub fn create_sub_agent_execute_task_tool() -> Tool {
+    Tool::new(
+        SUB_AGENT_EXECUTE_TASK_TOOL_NAME,
+        "Only use this tool when you want to execute sub recipe task. **DO NOT** use this tool when you want to execute sub agent task.   
+        If the tasks are not specified to be executed in parallel, you should use this tool to run each task immediately by passing a single task to the tool for each run.
+        If you want to execute tasks in parallel, you should pass a list of tasks to the tool.",
```
wendytang: I would add an example in the description

crates/goose/src/agents/sub_agent_execution_tool/workers.rs
```diff
@@ -0,0 +1,133 @@
...
+use crate::agents::sub_agent_execution_tool::types::{Task, TaskResult};
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+use std::sync::Arc;
+use tokio::sync::mpsc;
+use tokio::time::{sleep, Duration};
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::agents::sub_agent_execution_tool::types::Task;
```
wendytang: Could we rename all the vars and files from `sub_agent_execution_tool` to `sub_recipe_execution_tool`?

lifeizhou-ap: shall we leave this PR as it is, and I'll address this in the following PRs based on the progress?   I remember we agreed in our meeting that this tool will execute both sub-agent and sub recipe

I feel we are going to run the sub-agent/sub recipe as spawn thread soon instead of starting a goose cli process.

wendytang (approved):

Issue #3191, opened 2025-07-01
Failed Google/Gemini API requests show the API key in the output


SalvatoreT:
**Describe the bug**

When you get an HTTP error for calling the Google AI endpoint, you get the following error.

```
Ran into this error: Request failed: Request failed: error sending request for url (https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-pro-preview-06-05:generateContent?key=abc123def

Please retry if you think this is a transient or recoverable error.
```

This is bad because we're showing the API key in plain text.

**To Reproduce**
1. Start a Goose session with Google AI
2. Make a request
3. Quickly turn off your internet access

**Expected behavior**
I don't want to see the API key in the output.

**Screenshots**
![Image](https://github.com/user-attachments/assets/2b024a0e-fd50-4f24-bdcd-6c5c3c339ef9)

**Please provide following information:**
 - **OS & Arch:** macOS, ARM
 - **Interface:** UI
 - **Version:** v1.0.31
 - **Provider & Model:** Google - gemini-2.5-pro-preview-06-05

**Additional context**
It's [buried in their other documentation](https://cloud.google.com/docs/authentication/api-keys-use#using-with-rest), but you can use an HTTP header instead of including the API key in the query parameters.

> To include an API key with a REST API call, use the `x-goog-api-key` HTTP header, as shown in the following example

PR #3192, opened 2025-07-01
fix: Pass Google AI API key in HTTP header, not query param

SalvatoreT:
Sometimes the Google AI HTTP query is logged. If the key is a query param then it'll be visible to the user.

Closes https://github.com/block/goose/issues/3191


SalvatoreT:
@jamadeo, done!

PR #3195, opened 2025-07-01
Add support for mouse back nav button to Settings screen

pedronveloso:
### Problem
In various Goose flows, I often need to access the Advanced Settings screen to toggle extension states. However, navigating back to the main window requires scrolling all the way to the top to click the “Back” button, which takes unnecessary time

### Solution
This PR enables the Settings screen to respond to the mouse “Back” button (typically found on the side of many mice), allowing quicker navigation. Tested and working with a Logitech MX Master on macOS :)


ui/desktop/src/main.ts
```diff
@@ -682,6 +683,22 @@ const createChat = async (
...
 
+  // Handle mouse back button navigation.
+  mainWindow.on('app-command', (e, cmd) => {
+    if (cmd === 'browser-backward') {
+      mainWindow.webContents.send('mouse-back-button-clicked');
+      e.preventDefault();
+    }
+  });
+
+  // Handle mouse back button for platforms where app-command is not supported.
```
DOsinga: I'm not sure about this block. "many mice"? I don't know how we would test this on all platforms let alone all mice. can we just drop this bit?

ui/desktop/src/main.ts
```diff
@@ -682,6 +683,22 @@ const createChat = async (
     }
   });
 
+  // Handle mouse back button navigation.
```
DOsinga: remove the comment

ui/desktop/src/types/electron.d.ts
```diff
@@ -0,0 +1,14 @@
+declare namespace ElectronTypes {
```
DOsinga, line 1: what happened here?

ui/desktop/src/types/index.ts
```diff
@@ -0,0 +1 @@
+export * from './electron';
```
DOsinga, line 1: and here?

angiejones:
how about just a shortcut key?

ui/desktop/src/types/electron.d.ts
pedronveloso: The linter wouldn't allow me to commit with `any` on the parameters, so I had to create the types to fulfill that requirement. This is what that is. If there's already a package in the project where these types exists, I agree it makes sense to change to those, but I couldn't find them

pedronveloso:
> how about just a shortcut key?

This is a standard navigation pattern for web browsers, as well as native apps; I feel like adopting user expectation is better than defining a new shortcut that is app-specific.

ui/desktop/src/main.ts
pedronveloso: Dropping the comment. However this code is necessary for it to work.

zanesq:
@pedronveloso thanks this looks nice! Can you merge in main and test again and we'll get this in.

pedronveloso:
> @pedronveloso thanks this looks nice! Can you merge in main and test again and we'll get this in.

Done. Thanks for taking a look

PR #3198, opened 2025-07-01
Automate OpenRouter API Key Distribution for External Recipe Contributors

EbonyLouis:
This PR introduces a fully automated flow to reward community contributors who submit external recipes to the Goose Cookbook.

#### ✅ What’s Included
- **`send_key.py`**: A Python script that:
  - Extracts the contributor’s email from the PR body (base64-encoded)
  - Generates a $10 OpenRouter API key using the provisioning API
  - Sends the key via email using Resend's official Python SDK
  - Comments on the PR to confirm delivery

- **Updates to `create-recipe-pr.yml`**:
  - Parses the email field from the recipe submission issue
  - Base64-encodes it and adds it to the PR body in a hidden HTML comment

- **New workflow `send-api-key.yml`**:
  - Triggers only on merged PRs titled `Add External Recipe:`
  - Runs `send_key.py` to deliver the key and confirm via comment

#### 📫 Contributor Experience
When a recipe PR is merged:
- They receive an email with their OpenRouter API key
- The PR is automatically updated with a confirmation comment

#### 🔐 Secrets Required
Be sure the following secrets are set in the repo:
- `PROVISIONING_API_KEY` – from OpenRouter

### TO-DO:
- `EMAIL_API_KEY` – from sendGrid ✅
- update email with correct domain ✅


.github/scripts/send_key.py
```diff
@@ -0,0 +1,95 @@
...
+    to_emails=email,
+    subject=subject,
+    html_content=html_content
+)
+
+try:
+    response = sg.send(message)
+    print("✅ Email sent! Status code:", response.status_code)
+except Exception as e:
+    print("❌ Failed to send email:", str(e))
```
DOsinga: let's not catch general Exception's - also this will print the error and then happily move on. we're out 10 dollars, we comment on the PR that we send you the money, but we never did

.github/scripts/send_key.py
```diff
@@ -0,0 +1,95 @@
+import os
```
DOsinga, line 1: can we move this inside scaffolding with a function or two

DOsinga (approved):

zanesq:
@EbonyLouis still working on this?

blackgirlbytes:
> @EbonyLouis still working on this?

yeah @zanesq there's a part 2 that @iandouglas is not working on! Thanks for checking. We should be merging this soon

iandouglas:
> there's a part 2 that @iandouglas is not working on! 

my first part of the integration is added, just waiting on a word from platform security about whether the Dockerfile I reference in my integration here should use a docker registry for the base container, or whether we want to pull training material from other secure storage like S3.

.github/workflows/recipe-security-scanner.yml
```diff
@@ -0,0 +1,349 @@
...
+            
+            # Save recipe file paths for later steps
+            echo "$RECIPE_FILES" > "$RUNNER_TEMP/recipe_files.txt"
+          fi
+
+
+
+      - name: Set up Docker Buildx
+        if: steps.find_recipes.outputs.has_recipes == 'true'
+        uses: docker/setup-buildx-action@v3
```
blackgirlbytes: double check but i believe this is what you need

```suggestion
        uses: docker/setup-buildx-action@1583c0f09d26c58c59d25b0eef29792b7ce99d9a
```

.github/workflows/send-api-key.yml
```diff
@@ -0,0 +1,32 @@
...
+
+on:
+  pull_request:
+    types: [closed]
+    paths:
+      - 'documentation/src/pages/recipes/data/recipes/**'
+
+jobs:
+  send-api-key:
+    if: github.event.pull_request.merged == true
```
blackgirlbytes, line 11: This will trigger on EVERY merged PR that touches recipe files. Is there a common word in new recipe PRs..that we could use like


```yaml
    if: |
      github.event.pull_request.merged == true && 
      startsWith(github.event.pull_request.title, 'Add External Recipe:')
```

EbonyLouis: yes but an api key will only be submitted when an email address is present.

.github/scripts/send_key.py
```diff
@@ -0,0 +1,171 @@
...
+    print(f"📬 Found email: {email}")
+
+    try:
+        api_key = provision_api_key(PROVISIONING_API_KEY)
+        print("✅ API key generated!")
+        
+        if send_email(email, api_key, SENDGRID_API_KEY):
+            comment_on_pr(GITHUB_TOKEN, repo_full_name, pr_number, email)
+    except Exception as err:
+        print(f"❌ An error occurred: {err}")
```
blackgirlbytes, line 225: if sending the email fails, we wont know very well..because we could get a silent failure. Is there a way we could validate the email address first ..and then at this part..instead of just a general error occurred..catch the errors more specifically like a 401 or whatever.

.github/workflows/recipe-security-scanner.yml
```diff
@@ -0,0 +1,392 @@
...
+            elif [ "$OVERALL_STATUS" = "APPROVED" ]; then
+              echo "::notice::All recipes APPROVED by security scan"
+            else
+              echo "::error::Scan did not complete successfully - check artifacts for details"
+              exit 1
+            fi
+          else
+            echo "::error::No scan summary found - scan may have failed completely"
+            exit 1
+          fi
```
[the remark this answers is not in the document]
iandouglas: i fixed this in the next commit

PR #3208, opened 2025-07-01
feat: Adds max_turns for the agent without user input

jsibbison-square:
This pull request introduces a cli option `--max-turns`, allowing users to set a maximum number of turns (`max_turns`) that the agent can take without user input (the default is 1000 to maintain current functionality). If the agent reply loop exceeds max-turns it stops the loop and asks the user if they want to continue.

This functionality is useful to limit agents going off the rails and looping endlessly on a task.


Kvadratni (approved):
Amazing

PR #3218, opened 2025-07-02
feat: GUI can use structured output present in recipes

jsibbison-square:
This PR doesn't give the gui the ability to view or edit structured output but if structured output is part of the recipe it will now be used. Support in CLI was added in #3188 

image

### DIFF
--- a/ui/desktop/src/utils/providerUtils.ts
+++ b/ui/desktop/src/utils/providerUtils.ts
@@ -140,6 +140,8 @@ export const initializeSystem = async (
     // Get recipeConfig directly here
     const recipeConfig = window.appConfig?.get?.('recipeConfig');
     const botPrompt = (recipeConfig as { instructions?: string })?.instructions;
+    const responseConfig = (recipeConfig as { response?: { json_schema?: unknown } })?.response;
+
     // Extend the system prompt with desktop-specific information
     const response = await fetch(getApiUrl('/agent/prompt'), {
       method: 'POST',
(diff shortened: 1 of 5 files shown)


ui/desktop/src/utils/providerUtils.ts
```diff
@@ -140,6 +140,8 @@ export const initializeSystem = async (
     // Get recipeConfig directly here
     const recipeConfig = window.appConfig?.get?.('recipeConfig');
     const botPrompt = (recipeConfig as { instructions?: string })?.instructions;
+    const responseConfig = (recipeConfig as { response?: { json_schema?: unknown } })?.response;
```
lifeizhou-ap, line 143: I guess you have to add response attribute on the RecipeConfig interface on the client side, and populate to the object when parsing the deeplink. 

If you have tested this and it is works, that means my statement above is wrong :)

ui/desktop/src/utils/providerUtils.ts
```diff
@@ -162,6 +164,25 @@ export const initializeSystem = async (
       }
     }
 
+    // Configure session with response config if present
+    if (responseConfig?.json_schema) {
+      const sessionConfigResponse = await fetch(getApiUrl('/agent/session_config'), {
```
lifeizhou-ap, line 169: Shall we move this api call to another module? the function is getting longer :)

lifeizhou-ap:
just curious about this feature. It is good to see the output, but from user's prospective, what kind of use case/experience that they want the output?

lifeizhou-ap (approved):

ui/desktop/src/utils/providerUtils.ts
jsibbison-square: Yeah it works, recipeConfig gets populated early in the process, its pretty cryptic.

PR #3225, opened 2025-07-02
Add support for escape key to dismiss settings menu

danielcorin:
I got stuck in the settings menu. If you scroll down in the settings menu, the back link stays up at the top. This PR supports pressing escape to dismiss the settings menu and return to the main screen. A follow up to pin the back link to the top may make sense as well

https://github.com/user-attachments/assets/f07db13e-b8d0-4fe4-9e9b-7c42359b9aa3

Botched a rebase of a [previous PR](https://github.com/block/goose/pull/3124) of the same change. It was easier to create a new PR.


jamadeo (approved):
Makes sense to me, thanks @danielcorin

PR #3230, opened 2025-07-02
Clean up session file optionality for --no-session

### LINKED ISSUE #3228: `--no-session` argument in CLI are broken
**Description**
`--no-session` option crashes CLI entirely

**To Reproduce**
run `goose run -t "honk" --no-session` in cli

**Expected behavior**
Works with the flag

**Screenshots**
Image

 - **OS & Arch:**  MacOs 15.5
 - **Interface:** CLI
 - **Version:**  1.0.31

jamadeo:
Instead of using /dev/null or flags, make the session file optional on the agent and only try to save when we have one.

Closes #3228

### DIFF
--- a/crates/goose-cli/src/cli.rs
+++ b/crates/goose-cli/src/cli.rs
@@ -691,7 +691,11 @@ pub async fn cli() -> Result<()> {
                     })
                     .await;
                     setup_logging(
-                        session.session_file().file_stem().and_then(|s| s.to_str()),
+                        session
+                            .session_file()
+                            .as_ref()
+                            .and_then(|p| p.file_stem())
+                            .and_then(|s| s.to_str()),
                         None,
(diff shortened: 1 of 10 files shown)


crates/goose-cli/src/cli.rs
```diff
@@ -691,7 +691,11 @@ pub async fn cli() -> Result<()> {
                     })
                     .await;
                     setup_logging(
-                        session.session_file().file_stem().and_then(|s| s.to_str()),
```
jamadeo, line 694: This is new behavior -- setup_logging takes an `Option<&str>`, but here we'd pass it `/dev/null` which becomes a log file like `20250521_154141-null.log`.

After this change, we'll pass None, which just becomes `20250521_154141.log`

PR #3234, opened 2025-07-03
Adds json schema validation to goose recipe validate cli

jsibbison-square:
response.json_schema was added in #3188 . This pr extends `goose recipe validate` cli to also validate the json schema.

### DIFF
--- a/crates/goose-cli/src/recipes/extract_from_cli.rs
+++ b/crates/goose-cli/src/recipes/extract_from_cli.rs
@@ -28,10 +28,8 @@ pub fn extract_recipe_info_from_cli(
                 Ok(recipe_file) => {
                     let name = extract_recipe_name(&sub_recipe_name);
                     let recipe_file_path = recipe_file.file_path;
-                    let canonical_path =
-                        recipe_file_path.canonicalize().unwrap_or(recipe_file_path);
                     let additional_sub_recipe = SubRecipe {
-                        path: canonical_path.to_string_lossy().to_string(),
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/recipes/extract_from_cli.rs
```diff
@@ -28,8 +28,10 @@ pub fn extract_recipe_info_from_cli(
                 Ok(recipe_file) => {
                     let name = extract_recipe_name(&sub_recipe_name);
                     let recipe_file_path = recipe_file.file_path;
+                    let canonical_path =
+                        recipe_file_path.canonicalize().unwrap_or(recipe_file_path);
                     let additional_sub_recipe = SubRecipe {
-                        path: recipe_file_path.to_string_lossy().to_string(),
+                        path: canonical_path.to_string_lossy().to_string(),
```
lifeizhou-ap: The `recipe_file_path` returned by extract_recipe_name already has been canonicalized. Not sure why you would like to do it here again

jsibbison-square: weird the tests were failing, will explore.

jsibbison-square: Ah it was failing on main too when we have a symlink for our temp directory. The fix is to canoncalise all the test file paths.

lifeizhou-ap: Yeah, the expected value was wrong

PR #3238, opened 2025-07-03
Refactor cli session to be more readable

lifeizhou-ap:
Used Goose to do refactoring: 

- Break down large function into small functions
- Group functions into modules based on their responsibilities
- No logic change

Will refactor the agent_response_handler function (300+ lines) in the next PR


DOsinga:
do we still want this in?

lifeizhou-ap:
> do we still want this in?

Hi @DOsinga I'll close the PR first and may come back to refactor this later. Currently I am trying to refactor agent.rs. Sorry about that

PR #3246, opened 2025-07-03
Fix: Allow native Cmd+Up/Down cursor movement when user has typed text

aharvard:
## Problem
Currently, when users type text in the chat input field and press `Cmd+Up` (or `Ctrl+Up`), the text gets replaced with the previous message from history instead of moving the cursor to the beginning of the text. This conflicts with native macOS behavior and user expectations, especially for users familiar with apps like Slack where `Cmd+Up` moves the cursor when there's text in the input.

## Solution
This PR implements a smart solution that distinguishes between text that comes from history navigation vs. text that the user has actively typed:

- **When input is empty**: `Cmd+Up/Down` continues to work for history navigation (preserves existing behavior)
- **When text comes from history**: Users can continue using `Cmd+Up/Down` to navigate through more history items  
- **When user types text**: `Cmd+Up/Down` switches to native cursor movement behavior
- **Smart reset**: The system resets when text is populated from history, message is submitted, or input is cleared

## Technical Implementation
- Added `hasUserTyped` state to track when user actively modifies the input field
- Updated history navigation logic to only prevent navigation when `hasUserTyped && displayValue.trim() !== ''`
- Reset tracking appropriately in `handleChange`, `performSubmit`, history navigation, and initialization

## Benefits
✅ **Preserves all existing functionality** - History navigation works exactly as before when input is empty  
✅ **Allows continued history navigation** - Users can navigate through multiple history items without interference  
✅ **Enables native text editing** - When users type, they get familiar macOS cursor movement behavior  
✅ **Smart and intuitive** - Behavior automatically adapts based on how text got into the field  
✅ **No breaking changes** - All existing tests should continue to pass  

## Testing
- [x] TypeScript compilation passes
- [x] ESLint checks pass
- [x] Existing history navigation functionality preserved
- [x] Native cursor movement works when user types text
- [x] Continued history navigation works when text comes from history

Fixes the UX friction mentioned in the team discussion where `Cmd+Up` would unexpectedly replace user-typed text.

### DIFF
--- a/ui/desktop/src/components/ChatInput.tsx
+++ b/ui/desktop/src/components/ChatInput.tsx
@@ -118,6 +118,7 @@ export default function ChatInput({
     // Reset history index when input is cleared
     setHistoryIndex(-1);
     setIsInGlobalHistory(false);
+    setHasUserTyped(false);
   }, [initialValue]); // Keep only initialValue as a dependency
 
   // State to track if the IME is composing (i.e., in the middle of Japanese IME input)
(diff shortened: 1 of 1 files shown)


angiejones:
thank you!

PR #2946, opened 2025-06-16
Feat: Recipe Library

AaronGoldsmith:
| Save Recipe | Recipe Library (saved recipes) |
|--------|--------|
| image | saved recipes| 

<details open><summary>More Menu</summary>
<p>

"Save Recipe" is only visible when a recipe is open/ actively being used.

MoreMenu

Current functionality: "Go to Recipe Library" is always in the menu, even if you don't have any recipes saved

</p>
</details> 

<details><summary>💻 Quick demo</summary>
<p>

https://github.com/user-attachments/assets/26144c6e-0a42-4798-9ac0-02f4533423cc

</p>
</details> 

This pull request introduces a new "Recipe Library" feature, allowing users to save, manage, and reuse recipes within the application. It includes updates to the UI, functionality for saving and loading recipes, and a new `RecipesView` component for browsing saved recipes.

### New Recipe Management Features:

* **Added `RecipesView` Component**: Introduced a new `RecipesView` component for displaying saved recipes, allowing users to preview, load, and delete recipes. This includes error handling, a loading state, and a modal for previewing recipe details. (`ui/desktop/src/components/RecipesView.tsx`)
* **Save Recipe Functionality**: Implemented functionality in the `MoreMenu` component to save recipes with a user-provided name and location (global or project-specific). Includes a dialog for input and validation. (`ui/desktop/src/components/more_menu/MoreMenu.tsx`) [[1]](diffhunk://#diff-d27a9889348779359672b977c520fbd93ba28c22840c655d0e6a86995b0e3b34R112-R115) [[2]](diffhunk://#diff-d27a9889348779359672b977c520fbd93ba28c22840c655d0e6a86995b0e3b34R399-R479)

### UI Navigation Updates:

* **Recipe Library Navigation**: Added a menu option in `MoreMenu` to navigate to the new Recipe Library view. (`ui/desktop/src/components/more_menu/MoreMenu.tsx`)
[...]

PR #3254, opened 2025-07-04
feat: cli can work with gui generated recipes

jsibbison-square:
The gui recipes (ones saved in recipe library) have a nested yaml structure where the actual recipe is under a `recipe` key. This PR enables the cli to handle this when loading a recipe.

standard format:
```yaml
id: untitled
name: Untitled Recipe
title: asd
description: sdfsdf
instructions: All numbers must be in roman numerals
activities:
  - Start a new task
  - Ask for help
  - Explore available tools
  - Begin coding project
  - Request assistance
prompt: "What's 2+2"
extensions: []
```

gui format
```yaml
name: asd2
recipe:
  id: untitled
  name: Untitled Recipe
  title: asd
  description: sdfsdf
  instructions: All numbers must be in roman numerals
  activities:
    - Start a new task
    - Ask for help
    - Explore available tools
    - Begin coding project
    - Request assistance
  prompt: "What's 2+2"
  extensions: []
isGlobal: true
lastModified: 2025-07-02T03:46:46.778Z
isArchived: false

```

### DIFF
--- a/crates/goose/src/recipe/mod.rs
+++ b/crates/goose/src/recipe/mod.rs
@@ -629,4 +629,34 @@ sub_recipes:
         let activities = recipe.activities.unwrap();
         assert_eq!(activities, vec!["activity1", "activity2"]);
     }
+
+    #[test]
+    fn test_from_content_with_nested_recipe_yaml() {
+        let content = r#"name: test_recipe
+recipe:
+  title: Nested Recipe Test
+  description: A test recipe with nested structure
+  instructions: Test instructions for nested recipe
+  activities:
+    - Test activity 1
+    - Test activity 2
+  prompt: Test prompt
+  extensions: []
+isGlobal: true"#;
+
+        let recipe = Recipe::from_content(content).unwrap();
+        assert_eq!(recipe.title, "Nested Recipe Test");
+        assert_eq!(recipe.description, "A test recipe with nested structure");
+        assert_eq!(
+            recipe.instructions,
+            Some("Test instructions for nested recipe".to_string())
+        );
+        assert_eq!(recipe.prompt, Some("Test prompt".to_string()));
+        assert!(recipe.activities.is_some());
+        let activities = recipe.activities.unwrap();
+        assert_eq!(activities, vec!["Test activity 1", "Test activity 2"]);
+        assert!(recipe.extensions.is_some());
+        let extensions = recipe.extensions.unwrap();
+        assert_eq!(extensions.len(), 0);
+    }
 }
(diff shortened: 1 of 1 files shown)


cloud-on-prem (approved):

lifeizhou-ap:
I am just thinking one step back.  Why do we have different structures for recipe yaml file? Shall we make them identical?

Accept different structure as it may ends with multiple version of recipe file schemas and it will be hard to manage and maintain. 

Shall we fix this on the GUI to make the schema consistent?

DOsinga:
I agree w @lifeizhou-ap that this seems weird that we would have two formats. And these have a name in the root and then in the recipe another name and a title?

crates/goose/src/recipe/mod.rs
```diff
@@ -265,10 +265,18 @@ impl Recipe {
         }
     }
     pub fn from_content(content: &str) -> Result<Self> {
-        if serde_json::from_str::<serde_json::Value>(content).is_ok() {
-            Ok(serde_json::from_str(content)?)
-        } else if serde_yaml::from_str::<serde_yaml::Value>(content).is_ok() {
-            Ok(serde_yaml::from_str(content)?)
+        if let Ok(json_value) = serde_json::from_str::<serde_json::Value>(content) {
```
DOsinga, line 268: hmm - @lifeizhou-ap does looks like we reverted the code to trying both json and yaml again. why is that? every json file is also a yaml file, so we shouldn't need to do this twice here

DOsinga:
Looks like this is where this happened:
https://github.com/block/goose/pull/2946/files

jsibbison-square:
We will look to separate the storage of recipe and recipe library metadata so recipes are the same in gui and cli.

DOsinga:
eh, I don't know - we already merged the other thing so even if we fix it there, we might still want to have this in for at least a while... Also thanks for spotting this!

DOsinga:
I mean

jsibbison-square:
Alright let me get a passing build and I'll merge it

DOsinga:
Thanks!

crates/goose/src/recipe/mod.rs
lifeizhou-ap:
```
 if serde_json::from_str::<serde_json::Value>(content).is_ok() {
            Ok(serde_json::from_str(content)?)
        } else if serde_yaml::from_str::<serde_yaml::Value>(content).is_ok() {
            Ok(serde_yaml::from_str(content)?)
```
yes, that was me.  It was missed out during a refactoring a while ago. will update in another PR

PR #3257, opened 2025-07-04
Mnovich/opentelemetry

Kvadratni:
## Comprehensive OpenTelemetry Integration for Goose

This PR introduces a complete telemetry system for Goose, providing observability across CLI, Desktop UI, and recipe executions with support for multiple telemetry backends.

### Core Features

**Multi-Provider Telemetry Architecture**
- **Console Provider**: Debug output with structured logging
- **File Provider**: JSON-based telemetry export to local files  
- **OTLP Provider**: OpenTelemetry Protocol support for Jaeger, Tempo, and other OTLP-compatible backends
- **Datadog Provider**: Direct HTTP API integration for metrics and traces

**Comprehensive Event Tracking**
- **Recipe Executions**: Track recipe runs with metadata, duration, token usage, and tool usage statistics
- **Interactive Sessions**: Monitor CLI and UI conversations with message counts, turn tracking, and performance metrics
- **Command Executions**: Capture CLI command usage patterns and performance
- **Error Tracking**: Detailed error context with stack traces and failing tool identification

**Rich Telemetry Data**
- Token usage tracking (input/output tokens, estimated costs, model/provider info)
- Tool usage analytics (call counts, success rates, average duration)
- Environment detection (CI/CD, Docker, cloud platforms, terminal types)
- User identification and usage type classification (Human/Automation/CI)
- Comprehensive metadata and tagging support

### Technical Implementation

**Architecture**
- Centralized `TelemetryManager` with global singleton pattern
- Event-driven design with structured telemetry events
- Async-first implementation with proper resource cleanup
- Configurable via environment variables with validation

**Integration Points**
- CLI integration across all commands (configure, session management, recipe execution)
- Desktop UI telemetry via enhanced message stream hooks
- Server-side session tracking for both streaming and non-streaming endpoints
- Recipe execution tracking with comprehensive parameter capture

**Performance & Reliability**
- Batch export with forced flush on shutdown for OTLP
- HTTP-based Datadog integration (no agent required)
- Graceful error handling with telemetry event generation for failures
- Memory-efficient with configurable providers

### Configuration

```bash
# Enable telemetry
export GOOSE_TELEMETRY_ENABLED=true
export GOOSE_TELEMETRY_PROVIDER=otlp|datadog|console|file

# OTLP Configuration
export GOOSE_TELEMETRY_ENDPOINT=http://localhost:4317
export OTEL_SERVICE_NAME=goose

# Datadog Configuration  
export GOOSE_TELEMETRY_API_KEY=your_api_key
export GOOSE_TELEMETRY_ENDPOINT=https://api.datadoghq.com

# File Configuration
export GOOSE_TELEMETRY_ENDPOINT=/path/to/telemetry.log
```

This implementation provides production-ready observability for Goose deployments, enabling teams to monitor usage patterns, performance metrics, and system health across all interaction modes.


HalogenAI:
@lifeizhou-ap not sure if you have a better understanding on this from AI Log Reducer on the best practices to connect to DataDog?

lilydelalande:
I see this touches the message stream component in the UI -- think we may want to wait for this one till after the UI refactor just to be safe but will loop in Zane just to check

crates/goose/Cargo.toml
```diff
@@ -86,6 +86,15 @@ ahash = "0.8"
 lancedb = "0.13"
 arrow = "52.2"
 
+# OpenTelemetry telemetry
+opentelemetry = "0.27"
```
jsibbison-square: The current version is 0.30 so I think we should use that (and whatever the current version is of the other libraries) unless there is something stopping us. https://github.com/open-telemetry/opentelemetry-rust/releases https://crates.io/crates/opentelemetry

crates/goose/src/telemetry/config.rs
```diff
@@ -0,0 +1,249 @@
...
+            .and_then(|v| v.parse().ok())
+            .unwrap_or_default();
+
+        let endpoint = env::var("GOOSE_TELEMETRY_ENDPOINT")
+            .ok()
+            .or_else(|| env::var("OTEL_EXPORTER_OTLP_ENDPOINT").ok());
+
+        let api_key = env::var("GOOSE_TELEMETRY_API_KEY")
+            .ok()
+            .or_else(|| env::var("DD_API_KEY").ok());
```
jsibbison-square: Is there a reason we need to have a DD provider specific key rather than just using the GOOSE_TELEMETRY abstracted env keys?

I feel less env vars in general is better and goose is already collecting heaps of env vars...

crates/goose/src/telemetry/events.rs
```diff
@@ -0,0 +1,605 @@
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RecipeExecution {
+    pub recipe_name: String,
+    pub recipe_version: String,
+    pub start_time: u64,
+    pub end_time: Option<u64>,
+    pub duration_ms: Option<u64>,
+    pub result: Option<RecipeResult>,
```
jsibbison-square: Can we reuse SessionResult (even if some values may never appear) rather adding a new type which is a subset? As recipes are really the same execution as a session just with predefined configuration.

crates/goose/src/telemetry/events.rs
```diff
@@ -0,0 +1,605 @@
...
+    pub user_id: String,
+    pub usage_type: UsageType,
+    pub environment: Option<String>,
+    pub metadata: HashMap<String, String>,
+    pub token_usage: Option<TokenUsage>,
+    pub tool_usage: Vec<ToolUsage>,
+    pub error_details: Option<ErrorDetails>,
```
jsibbison-square, line 170: I noticed RecipeExecution is missing turn_count and message_count compared to SessionExecution. Is that deliberate?

crates/goose/src/telemetry/config.rs
Kvadratni: I might get rid of it entirely if we build a proxy.

crates/goose/src/telemetry/events.rs
Kvadratni: good catch TY

crates/goose/src/telemetry/events.rs
Kvadratni: nope that's a miss

crates/goose-cli/src/cli.rs
```diff
@@ -29,6 +29,345 @@ use goose_bench::runners::model_runner::ModelRunner;
...
+{
+    let start_time = std::time::Instant::now();
+
+    let telemetry_execution = goose::telemetry::global_telemetry().map(|_manager| {
+        SessionExecution::new(session_id, session_type).with_metadata("interface", "cli")
+    });
+
+    let result = execution_fn().await;
+
+    if let Some(mut execution) = telemetry_execution {
```
cloud-on-prem: Probably, worth catching the `None` case and logging it if Telemetry fails for some reason?

crates/goose-cli/src/cli.rs
```diff
@@ -29,6 +29,345 @@ use goose_bench::runners::model_runner::ModelRunner;
...
+    for message in &messages {
+        for content in &message.content {
+            match content {
+                MessageContent::ToolRequest(tool_request) => {
+                    if let Ok(tool_call) = &tool_request.tool_call {
+                        let tool_name = &tool_call.name;
+                        let entry = tool_usage_map
+                            .entry(tool_name.clone())
+                            .or_insert_with(|| ToolUsage::new(tool_name));
+                        entry.add_call(std::time::Duration::from_millis(0), true);
```
cloud-on-prem: Is this intentionally set to zero always?

crates/goose-cli/src/cli.rs
```diff
@@ -29,6 +29,345 @@ use goose_bench::runners::model_runner::ModelRunner;
...
+            metadata.insert("schedule_id".to_string(), schedule_id);
+        }
+    }
+
+    let environment = detect_environment();
+
+    (token_usage, tool_usage, metadata, environment)
+}
+
+fn detect_environment() -> Option<String> {
```
cloud-on-prem: wow, this is a big list 🤣

cloud-on-prem (approved):
Looks good; left some minor suggestions/comments.

crates/goose/Cargo.toml
cloud-on-prem: I have been bitten by this, there's a VS Code extension called `crates` which presents this nicely.
https://marketplace.visualstudio.com/items?itemName=serayuzgur.crates

crates/goose-cli/src/cli.rs
```diff
@@ -29,6 +29,345 @@ use goose_bench::runners::model_runner::ModelRunner;
...
 use std::path::PathBuf;
 
+use goose::message::MessageContent;
+use goose::telemetry::{
+    CommandExecution, CommandResult, CommandType, SessionExecution, SessionResult, SessionType,
+    TokenUsage, ToolUsage,
+};
+use std::collections::HashMap;
+
+fn extract_telemetry_data_from_session(
```
lifeizhou-ap: The cli.rs is getting very long.  Shall we have a separate modules for these functions and group them logically?

crates/goose/src/telemetry/providers/datadog.rs
```diff
@@ -0,0 +1,528 @@
...
+
+impl DatadogProvider {
+    pub fn new() -> Self {
+        Self {
+            datadog_metrics: None,
+            initialized: false,
+        }
+    }
+
+    async fn send_datadog_metrics(
```
lifeizhou-ap: Hi @Kvadratni I feel here  datadog-exporter from the opentelemtry sdk is not used. (Please correct me if I am wrong).  It uses `DatadogMetricsExporter`

Just wondering whether it is possible use the client token to send the data to datadog via the `DatadogMetricsExporter`? 

If we cannot do that and use a proxy service, I guess we may not need this provider, as the otlp provider should be used to collect/export data to datadog with configuration

crates/goose-server/src/routes/reply.rs
```diff
@@ -206,11 +268,23 @@ async fn handler(
                     &tx,
                 )
                 .await;
+
+                // Track failed session
+                if let Some(manager) = global_telemetry() {
```
lifeizhou-ap: maybe extract this function as there are a couple of duplications

lifeizhou-ap:
> @lifeizhou-ap not sure if you have a better understanding on this from AI Log Reducer on the best practices to connect to DataDog?

Hi @HalogenAI in Log Reducer, our focus is on `monitoring log volumes`, which is a different concern from what’s implemented in this PR, where the emphasis is on `sending or tracking custom events` (e.g., via send_event). These are separate areas in Datadog, and serving different observability goals.

crates/goose/src/telemetry/providers/datadog.rs
Kvadratni: nope. Client token does not allow that. Tried it

crates/goose-cli/src/cli.rs
Kvadratni: 100%.

crates/goose-server/src/routes/reply.rs
Kvadratni: fair

Cargo.toml
```diff
@@ -13,6 +13,31 @@ description = "An AI agent"
 [workspace.lints.clippy]
 uninlined_format_args = "allow"
 
+# Optimize build profiles for memory usage and CI performance
+[profile.dev]
+# Reduce memory usage during compilation
+debug = 1  # Reduced debug info
```
jamadeo: What's the story here?

crates/goose-cli/src/cli.rs
```diff
@@ -29,6 +29,345 @@ use goose_bench::runners::model_runner::ModelRunner;
...
+                        let entry = tool_usage_map
+                            .entry(tool_name.clone())
+                            .or_insert_with(|| ToolUsage::new(tool_name));
+                        entry.add_call(std::time::Duration::from_millis(0), true);
+                    }
+                }
+                MessageContent::ToolResponse(tool_response) => {
+                    if tool_response.tool_result.is_err() {
+                        for tool_usage in tool_usage_map.values_mut() {
+                            if tool_usage.error_count < tool_usage.call_count {
```
jamadeo: this doesn't seem right -- .values_mut does not define an order.

Shouldn't these all have tool call ids?

crates/goose-cli/src/cli.rs
```diff
@@ -29,6 +29,345 @@ use goose_bench::runners::model_runner::ModelRunner;
...
+    }
+
+    let tool_usage: Vec<ToolUsage> = tool_usage_map.into_values().collect();
+
+    let mut metadata = HashMap::new();
+    for (key, value) in params {
+        metadata.insert(key.clone(), value.clone());
+    }
+
+    if let Ok(session_metadata) = session.get_metadata() {
```
jamadeo: Session metadata is serializable, can we just include send it as-is? This should also include token usage

crates/goose-cli/src/cli.rs
jamadeo: I don't think this really belongs in cli.rs -- maybe in the telemetry module?

crates/goose-cli/src/cli.rs
```diff
@@ -762,6 +1118,10 @@ pub async fn cli() -> Result<()> {
             quiet,
             additional_sub_recipes,
         }) => {
+            // Check if this is a recipe execution for telemetry tracking
+            let is_recipe_execution = recipe.is_some();
+            let recipe_name_for_telemetry = recipe.clone().unwrap_or_default();
```
jamadeo, line 760: why not check & extract this where you're using it?

crates/goose-cli/src/cli.rs
```diff
@@ -827,51 +1187,130 @@ pub async fn cli() -> Result<()> {
...
-                    .and_then(|s| s.to_str()),
-                None,
-            )?;
-
-            if interactive {
-                let _ = session.interactive(input_config.contents).await;
-            } else if let Some(contents) = input_config.contents {
-                let _ = session.headless(contents).await;
+            if is_recipe_execution {
+                let recipe_version = "1.0.0";
```
jamadeo: ?

crates/goose-cli/src/cli.rs
```diff
@@ -690,41 +1034,53 @@ pub async fn cli() -> Result<()> {
...
-                    // Render previous messages if resuming a session and history flag is set
-                    if resume && history {
-                        session.render_message_history();
-                    }
+                    let session_id = format!(
+                        "session_{}",
+                        std::time::SystemTime::now()
+                            .duration_since(std::time::UNIX_EPOCH)
+                            .unwrap_or_default()
+                            .as_secs()
```
jamadeo: what's the purpose of these ids? it seems like we're creating a timestamp each time we call it

crates/goose-cli/src/main.rs
```diff
@@ -3,5 +3,27 @@ use goose_cli::cli::cli;
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    cli().await
+    // Check if this is an MCP process - if so, don't initialize telemetry
+    let args: Vec<String> = std::env::args().collect();
+    let is_mcp_process = args.len() >= 2 && args[1] == "mcp";
```
jamadeo: Why not do this after the command line is parsed?

crates/goose-server/src/routes/reply.rs
```diff
@@ -151,6 +175,35 @@ async fn handler(
...
+
+                        // Track failed session
+                        if let Some(manager) = global_telemetry() {
+                            let failed_execution = session_execution
+                                .clone()
+                                .with_result(SessionResult::Error(
+                                    "No provider configured".to_string(),
+                                ))
+                                .with_duration(start_time.elapsed());
+                            let _ = manager.track_session_execution(failed_execution).await;
```
jamadeo: from here down you're ignoring the error, maybe log something given this is in the server?

jamadeo (dismissed):
This is nice and will be super useful! My two main pieces of feedback are:

- Try to do as little as possible in the cli/server implementation, and move the telemetry-related logic to the telemetry crate. This will better separate the responsibilities of each. I'm also not a huge fan of having to put everything in closures passed to the telemetry manager -- maybe that can be done more cleanly with a macro? WDYT?
- See if you can reuse the serializable structs e.g. SessionMetadata to build the telemetry payload as-is instead of having to convert everywhere

DOsinga:
how does this relate to: https://github.com/block/goose/pull/3401

.github/workflows/ci.yml
```diff
@@ -131,7 +143,7 @@ jobs:
           source ./bin/activate-hermit
           export CARGO_INCREMENTAL=0
           cargo clippy --jobs 2 -- -D warnings
-          
+
   desktop-lint:
     name: Lint Electron Desktop App
     runs-on: macos-latest
```
DOsinga: are these related changes? also why do we need lld and clang

crates/goose-server/src/routes/reply.rs
```diff
@@ -130,6 +133,27 @@ async fn handler(
...
 
     tokio::spawn(async move {
+        let start_time = Instant::now();
+        let mut session_execution = SessionExecution::new(&session_id, SessionType::Interactive)
+            .with_metadata("execution_mode", "server")
+            .with_metadata("session_type", "streaming")
+            .with_metadata("interface", "ui");
+
+        // Add recipe metadata if available
+        if let Some(recipe_name) = &request.recipe_name {
```
DOsinga: remove comments

crates/goose-server/src/routes/reply.rs
```diff
@@ -130,6 +133,27 @@ async fn handler(
...
+            session_execution = session_execution
+                .with_metadata("recipe_name", recipe_name)
+                .with_metadata("session_mode", "recipe");
+        }
+        if let Some(recipe_version) = &request.recipe_version {
+            session_execution = session_execution.with_metadata("recipe_version", recipe_version);
+        }
+
+        // If no recipe, mark as regular chat
+        if request.recipe_name.is_none() {
```
DOsinga: this should be in the else

crates/goose-server/src/routes/reply.rs
```diff
@@ -130,6 +133,27 @@ async fn handler(
...
+            .with_metadata("session_type", "streaming")
+            .with_metadata("interface", "ui");
+
+        // Add recipe metadata if available
+        if let Some(recipe_name) = &request.recipe_name {
+            session_execution = session_execution
+                .with_metadata("recipe_name", recipe_name)
+                .with_metadata("session_mode", "recipe");
+        }
+        if let Some(recipe_version) = &request.recipe_version {
```
DOsinga: should be inside the if

crates/goose-cli/src/cli.rs
```diff
@@ -29,6 +29,344 @@ use goose_bench::runners::model_runner::ModelRunner;
...
+    if std::env::var("TERM_PROGRAM").as_deref() == Ok("iTerm.app") {
+        env_indicators.push("iterm");
+    }
+    if std::env::var("TERM_PROGRAM").as_deref() == Ok("Apple_Terminal") {
+        env_indicators.push("terminal-app");
+    }
+
+    if std::env::var("GOOSE_JOB_ID").is_ok() {
+        env_indicators.push("scheduled");
+    }
```
DOsinga: this should not be a series of if statements, but a data declaration

DOsinga (commented):
sorry for the drive by - it does feel like a de-LLM-ify pass would help

Kvadratni:
> how does this relate to: #3401

@DOsinga it does not at all

crates/goose-cli/src/cli.rs
Kvadratni: bug

.github/workflows/ci.yml
Kvadratni: rebase artifacts

crates/goose-cli/src/cli/main.rs
```diff
@@ -968,15 +955,6 @@ pub async fn cli() -> Result<()> {
...
                 }
-                SchedulerCommand::ServicesStatus {} => {
-                    handle_schedule_services_status().await?;
-                }
-                SchedulerCommand::ServicesStop {} => {
-                    handle_schedule_services_stop().await?;
-                }
-                SchedulerCommand::CronHelp {} => {
-                    handle_schedule_cron_help().await?;
-                }
```
jsibbison-square: Don't need these anymore?

crates/goose/src/telemetry/manager.rs
```diff
@@ -0,0 +1,586 @@
...
+}
+
+#[macro_export]
+macro_rules! track_recipe {
+    ($name:expr, $version:expr) => {{
+        if let Some(manager) = $crate::telemetry::global_telemetry() {
+            manager.recipe_execution($name, $version)
+        } else {
+            $crate::telemetry::RecipeExecutionBuilder::new($name, $version)
+        }
```
jsibbison-square: I think both of these lines branches are ultimately running the same code?

crates/goose-cli/src/cli/main.rs
Kvadratni: refactoring mistake. TY

crates/goose/src/telemetry/manager.rs
Kvadratni: done

crates/goose/src/telemetry/macros.rs
```diff
@@ -0,0 +1,173 @@
+#[macro_export]
+macro_rules! track_telemetry {
+    (session: ($session_id:expr, $session_type:expr) => $body:block) => {{
+        use goose::telemetry::{TelemetryExecution, SessionMetadataSupport};
+
+        let start_time = std::time::Instant::now();
+
+        let telemetry_execution = goose::telemetry::global_telemetry().map(|_manager| {
+            goose::telemetry::SessionExecution::new($session_id, $session_type)
+                .with_metadata("interface", "cli")
```
jamadeo, line 10: should the "interface" "cli" be here?

crates/goose-cli/src/cli/main.rs
```diff
@@ -837,6 +755,10 @@ pub async fn cli() -> Result<()> {
             provider,
             model,
         }) => {
+            // Check if this is a recipe execution for telemetry tracking
+            let is_recipe_execution = recipe.is_some();
+            let recipe_name_for_telemetry = recipe.clone().unwrap_or_default();
+
```
jamadeo, line 761: still seems like you could just do `if let Some(recipe_name) = recipe` below instead of these

crates/goose-cli/src/cli/main.rs
```diff
@@ -896,51 +821,127 @@ pub async fn cli() -> Result<()> {
...
-                additional_system_prompt: input_config.additional_system_prompt,
-                settings: recipe_info
+            if is_recipe_execution {
+                let recipe_version = match crate::recipes::recipe::load_recipe(
+                    &recipe_name_for_telemetry,
+                    params.clone(),
+                ) {
+                    Ok(recipe) => recipe.version,
+                    Err(_) => "unknown".to_string(),
+                };
```
jamadeo, line 831: could do

.map(|r| r.version).unwrap_or("unknown".to_string())

crates/goose/src/telemetry/manager.rs
```diff
@@ -0,0 +1,582 @@
...
+
+            if let Err(e) = backend_guard.send_event(&event).await {
+                tracing::warn!("Failed to send telemetry event: {}", e);
+            }
+        }
+
+        Ok(())
+    }
+
+    pub async fn track_recipe_executions(
```
jamadeo, line 114: this appears to be unused

crates/goose/src/telemetry/providers/datadog.rs
```diff
@@ -0,0 +1,527 @@
...
+        // 2. Custom subdomain endpoints (not available for traces)
+        // 3. Complex msgpack format (adds dependency overhead)
+        //
+        // Our comprehensive metrics already provide the key insights:
+        // - Recipe execution counts and timing
+        // - Token usage and costs
+        // - Tool usage patterns
+        // - Error rates and success metrics
+        //
+        // This covers 95% of our telemetry needs without the complexity of traces.
```
jamadeo: this comment probably doesn't belong here

jamadeo (changes requested):
I started going through the code again but before moving forward I wanted to take a moment to discuss the approach more broadly.

It seems like a lot of this change is setting up an abstraction layer for telemetry, but isn't that exactly what opentelemetry should give us? And then the abstraction itself requires that each backend implement it's own translation from our app events to telemetry data.

I would expect the only bit we have between our app code and opentelemetry would just be a few functions like `track_session()`, `track_recipe()` etc. that take our app events and produce opentelemetry metrics (counters, gauges etc.). Then let opentelemetry be the abstraction between the metrics and the collector.

WDYT?

dcieslak19973:
Any update on getting this merged?

Kvadratni:
> Any update on getting this merged?

@dcieslak19973 needs a tiny bit more tweaking

Kvadratni:
@jamadeo @DOsinga I dropped the direct datadog provider in favour of OTLP
made it work too. 
I might need to fix the env var approach but would prefer it to be a quick follow up PR

image

Kvadratni:
replaced by https://github.com/block/goose/pull/3772

PR #3262, opened 2025-07-06
refactor: abstract keyring logic to better enable DI

cloud-on-prem:
This change introduces a clean keyring abstraction layer to replace global keyring state with dependency injection, addressing critical runtime errors and improving testability. 

The implementation creates a KeyringBackend trait with concrete implementations for system keyring (SystemKeyringBackend) and file-based storage (FileKeyringBackend), allowing the configuration system to use different keyring backends based on the GOOSE_DISABLE_KEYRING environment variable. 

This architectural improvement eliminates the need to touch the user's real OS keyring and thus enables comprehensive unit testing without keychain popups. The refactoring maintains full backward compatibility while establishing a foundation for reliable testing and development workflows

---

- **feat: introduce a keyring abstraction to help with dep injection**
- **feat: use dep injection for keyring ops**
- **fix: lints and fmts**
- **feat: more just recipes for qol**
- **refactor: reuse keyring abstraction from goose crate**
- **chore: add a lint-all just directive**
- **fix: fix nested runtime issues by using shared future rt**
- **refactor: move file based keyring logic to own module**
- **refactor(tests): use a guard pattern to restore env vars**

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -185,6 +185,11 @@ test:
 # Run linting checks across all crates
 lint:
     @echo "Running linting checks..."
+    cargo clippy --workspace --all-features -- -D warnings
+
+# Run comprehensive linting checks (includes tests, examples, benchmarks)
+lint-all:
+    @echo "Running comprehensive linting checks..."
     cargo clippy --workspace --all-targets --all-features -- -D warnings
 
 # Format code across all crates
@@ -377,16 +382,16 @@ set windows-shell := ["powershell.exe", "-NoLogo", "-Command"]
 ### Build the core code
 ### profile = --release or "" for debug
 ### allparam = OR/AND/ANY/NONE --workspace --all-features --all-targets
-win-bld profile allparam: 
+win-bld profile allparam:
   cargo run {{profile}} -p goose-server --bin  generate_schema
   cargo build {{profile}} {{allparam}}
 
 ### Build just debug
-win-bld-dbg: 
+win-bld-dbg:
   just win-bld " " " "
 
 ### Build debug and test, examples,...
-win-bld-dbg-all: 
+win-bld-dbg-all:
   just win-bld " " "--workspace --all-targets --all-features"
 
 ### Build just release
@@ -449,8 +454,8 @@ win-total-rls *allparam:
   just win-bld-rls{{allparam}}
   just win-run-rls
 
-### Build and run the Kotlin example with 
-### auto-generated bindings for goose-llm 
+### Build and run the Kotlin example with
+### auto-generated bindings for goose-llm
 kotlin-example:
     # Build Rust dylib and generate Kotlin bindings
     cargo build -p goose-llm
--- a/Justfile
+++ b/Justfile
@@ -177,6 +177,28 @@ run-server:
     @echo "Running server..."
     cargo run -p goose-server
 
+# Run tests across all crates with keyring disabled
+test:
+    @echo "Running tests with keyring disabled..."
+    GOOSE_DISABLE_KEYRING=1 cargo test --workspace
+
+# Run linting checks across all crates
+lint:
+    @echo "Running linting checks..."
+    cargo clippy --workspace --all-targets --all-features -- -D warnings
+
+# Format code across all crates
+format:
+    @echo "Formatting code..."
+    cargo fmt --all
+
+# Quality assurance: format, lint, and test everything
+qa:
+    @echo "Running quality assurance: format, lint, and test..."
+    @just format
+    @just lint
+    @just test
+
 # make GUI with latest binary
 lint-ui:
     cd ui/desktop && npm run lint:check
--- a/crates/goose-mcp/src/google_drive/mod.rs
+++ b/crates/goose-mcp/src/google_drive/mod.rs
@@ -161,12 +161,21 @@ impl GoogleDriveRouter {
             Err(_) => false,
         };
 
+        // Create the appropriate keyring backend based on environment
+        let keyring: Arc<dyn goose::keyring::KeyringBackend> =
+            if std::env::var("GOOSE_DISABLE_KEYRING").is_ok() {
+                Arc::new(goose::keyring::MockKeyringBackend::new())
+            } else {
+                Arc::new(goose::keyring::SystemKeyringBackend)
+            };
+
         // Create a credentials manager for storing tokens securely
         let credentials_manager = Arc::new(CredentialsManager::new(
             credentials_path.clone(),
             fallback_to_disk,
             KEYCHAIN_SERVICE.to_string(),
             KEYCHAIN_USERNAME.to_string(),
+            keyring,
         ));
 
         // Read the OAuth credentials from the keyfile
--- a/crates/goose-mcp/src/google_drive/mod.rs
+++ b/crates/goose-mcp/src/google_drive/mod.rs
@@ -161,13 +161,8 @@ impl GoogleDriveRouter {
             Err(_) => false,
         };
 
-        // Create the appropriate keyring backend based on environment
-        let keyring: Arc<dyn goose::keyring::KeyringBackend> =
-            if std::env::var("GOOSE_DISABLE_KEYRING").is_ok() {
-                Arc::new(goose::keyring::MockKeyringBackend::new())
-            } else {
-                Arc::new(goose::keyring::SystemKeyringBackend)
-            };
+        // Use factory to create keyring backend consistently
+        let keyring = goose::keyring::create_default_keyring();
 
         // Create a credentials manager for storing tokens securely
         let credentials_manager = Arc::new(CredentialsManager::new(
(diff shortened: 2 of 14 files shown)


Justfile
```diff
@@ -177,6 +177,33 @@ run-server:
...
+
+# Run linting checks across all crates
+lint:
+    @echo "Running linting checks..."
+    cargo clippy --workspace --all-features -- -D warnings
+
+# Run comprehensive linting checks (includes tests, examples, benchmarks)
+lint-all:
+    @echo "Running comprehensive linting checks..."
+    cargo clippy --workspace --all-targets --all-features -- -D warnings
```
jsibbison-square, line 193: is there a reason that lint doesn't do all-targets too? I feel like having distinct lint commands means we won't use one of them.

crates/goose-mcp/src/google_drive/storage.rs
```diff
@@ -21,6 +22,15 @@ pub enum StorageError {
     SerializationError(#[from] serde_json::Error),
 }
 
+impl From<KeyringError> for StorageError {
+    fn from(err: KeyringError) -> Self {
+        match err {
+            KeyringError::NotFound { .. } => StorageError::NotFound,
```
jsibbison-square, line 28: Is this losing information that might be useful for debugging, can it be retained when converting to StorageError?

crates/goose-mcp/src/google_drive/storage.rs
```diff
@@ -263,25 +299,31 @@ mod tests {
...
+        assert!(
+            write_result.is_ok(),
+            "Write should succeed with mock keyring"
+        );
 
-        // Read should read from keychain
+        // Read should succeed with mock keyring
         let read_result = manager.read_credentials::<TestCredentials>();
-        assert!(read_result.is_ok(), "Read should succeed with fallback");
+        assert!(read_result.is_ok(), "Read should succeed with mock keyring");
```
jsibbison-square, line 326: I think this test is still about using the 'fallback_to_disk'=true rather than about using a mock keyring.

crates/goose/src/config/base.rs
```diff
@@ -105,19 +106,31 @@ impl From<keyring::Error> for ConfigError {
...
 // Global instance
 static GLOBAL_CONFIG: OnceCell<Config> = OnceCell::new();
 
 impl Default for Config {
     fn default() -> Self {
+        let config_dir = choose_app_strategy(APP_STRATEGY.clone())
+            .expect("goose requires a home dir")
+            .config_dir();
+
+        let keyring: Arc<dyn KeyringBackend> = if env::var("GOOSE_DISABLE_KEYRING").is_ok() {
```
jsibbison-square: Is it it worth explicitly checking the value inside the GOOSE_DISABLE_KEYRING rather than just checking the var is present? Most vars can be true or false depending on the value.

Justfile
cloud-on-prem: yeah - currently lint-all fails; we should fix them up in another PR. I left it there for future use. Currently `just lint` is equivalent of what's run in CI

crates/goose/src/config/base.rs
```diff
@@ -105,19 +106,31 @@ impl From<keyring::Error> for ConfigError {
...
+            Arc::new(FileKeyringBackend::new(config_dir.join("secrets.yaml")))
```
jsibbison-square: Bonus points if "secrets.yaml" is a constant.

crates/goose/src/config/base.rs
```diff
@@ -932,6 +948,131 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_multiple_secrets_with_mock() -> Result<(), ConfigError> {
```
jsibbison-square, line 942: Can we remove the test_secret_management_with_mock test and make this cover it all?

crates/goose/src/keyring/file.rs
```diff
@@ -0,0 +1,254 @@
...
+use anyhow::Result;
+use serde_json::Value;
+use std::collections::HashMap;
+use std::path::PathBuf;
+
+pub struct FileKeyringBackend {
+    secrets_path: PathBuf,
+}
+
+impl FileKeyringBackend {
```
jsibbison-square, line 11: Is this backwards compatible with the existing 'secrets.yaml' that was previously used?

crates/goose/src/keyring/file.rs
```diff
@@ -0,0 +1,254 @@
...
+    use super::*;
+    use tempfile::NamedTempFile;
+
+    #[test]
+    fn test_basic_operations() -> Result<()> {
+        let temp_file = NamedTempFile::new().unwrap();
+        let backend = FileKeyringBackend::new(temp_file.path().to_path_buf());
+
+        // Test setting a password
+        backend.set_password("test_service", "test_user", "test_password")?;
```
jsibbison-square, line 131: These comments don't add anything compared to just the code on the next line, please remove and check for similar issues throughout.

crates/goose-mcp/src/google_drive/storage.rs
cloud-on-prem: Possibly, although the only thing NotFound contains is `NotFound { service: String, username: String }`.

crates/goose/src/providers/factory.rs
```diff
@@ -199,27 +232,39 @@ mod tests {
...
+        // Since we provided fake API keys, the creation should proceed to the provider instantiation
+        // It may still fail at the provider level, but we should not get keychain-related errors
         match result {
             Ok(_) => {
-                // If it succeeds, it means we created a lead/worker provider successfully
-                // This would happen if API keys are available in the test environment
+                // If it succeeds with fake keys, the logic worked
             }
             Err(error) => {
-                // If it fails, it should be due to missing API keys, confirming we tried to create providers
```
jsibbison-square, line 251: Maybe some of these original comments are useful context?

crates/goose/src/providers/factory.rs
```diff
@@ -149,6 +149,39 @@ mod tests {
     use mcp_core::{content::TextContent, Role};
     use std::env;
```
jsibbison-square, line 182: At a glance it seems like all the changes in this file are unrelated to the PR focus which is using a new keyring abstraction. Maybe it crept in from some other work and we should keep it separate?

jsibbison-square (approved):
Left some comments but overall I think this looks ok. Double check the file implementation is backwards compatible. Would appreciate a second set of eyes from someone with history on keychain implementation (maybe @michaelneale ?) but looks ok to me.

crates/goose/src/providers/factory.rs
cloud-on-prem: Yep agree - it was an opportunistic refactor while I was removing the need for touching the user's keyring when running tests. Originally, I was setting the `ENV` in the tests to disable the keyring access. But I could not resist the temptation to use `Drop` trait :)

crates/goose/src/providers/factory.rs
cloud-on-prem: This comment I thought was more relevant: // Should not fail due to missing secrets, but may fail due to invalid fake keys and // Since we provided fake API keys, the creation should proceed to the provider instantiation
        // It may still fail at the provider level, but we should not get keychain-related errors

crates/goose/src/keyring/file.rs
cloud-on-prem: good catch; I missed the old secrets.yaml flow here. Adding it now.

crates/goose/src/config/base.rs
cloud-on-prem: yep, worth making it more explicit is a good idea; fixed it.

crates/goose/src/config/base.rs
cloud-on-prem: yep good call; added it now

crates/goose-mcp/src/google_drive/mod.rs
```diff
@@ -161,12 +161,21 @@ impl GoogleDriveRouter {
             Err(_) => false,
         };
 
+        // Create the appropriate keyring backend based on environment
+        let keyring: Arc<dyn goose::keyring::KeyringBackend> =
+            if std::env::var("GOOSE_DISABLE_KEYRING").is_ok() {
```
lifeizhou-ap: Since it will create a MockKeyringBackend, just curious about:  `MockKeyringBackend` condition only hit while running test? or  any chances that`GOOSE_DISABLE_KEYRING` is set when user runs goose?

cloud-on-prem: Good catch; fixed it by using a better abstraction.

michaelneale:
I still dont' quite follow this - was this tested once compiled, and also run as desktop bundle? as it seems a very risky change

michaelneale:
(I did test it with it bundled and signed, seemed ok)

michaelneale:
main concern is that goose-mcp has a goose crate dependency (don't really know what that means in the end, but feels potentially circular)

PR #3264, opened 2025-07-06
fix: use safe_truncate to truncate charactor (#3263)

### LINKED ISSUE #3263: bug: String truncation fails with "byte index is not a char boundary" for non-ASCII text
**Bug: String truncation fails with "byte index is not a char boundary" for non-ASCII text**

**Describe the bug**
Goose crashes with `byte index is not a char boundary` error when truncating text content that contains multi-byte UTF-8 characters (Japanese, Chinese, Korean, emojis, etc.). This occurs because the code uses byte-based string slicing instead of character-boundary-aware slicing. 

**To Reproduce**
1. Use Goose with content containing Japanese/Chinese/Korean text or emojis
2. Trigger text truncation (large messages, tool responses, or session loading)
3. See error: `byte index is not a char boundary`

**Expected behavior**
Text should be truncated safely at character boundaries without crashing.

**Affected Code Locations**
The following locations use unsafe byte-based string slicing:

1. **Context Management Truncation** (`crates/goose/src/context_mgmt/truncate.rs`):

2. **Session Storage Truncation** (`crates/goose/src/session/storage.rs`): caused by https://github.com/block/goose/pull/3071
3. etc

toyamagu-2021:
Signed-off-by: toyamagu2021@gmail.com <toyamagu2021@gmail.com>

Closes #3263 
Same as #2819


toyamagu-2021:
@jamadeo Sorry to bother you, but could you review this PR, which is almost the same as #2819?

DOsinga (approved):
this is great and thanks for jumping on this. if we want to completely fix this, we should also replace the text.len() instances with text.chars().count() or where possible replace

if text.len() > count {
   safe_truncate(text)
} else {
   text
}

with just safe_truncate(text)

toyamagu-2021:
@DOsinga 
Thank you for your feedback! I've fixed some codes.

DOsinga:
this is great, thank you. do you have any ideas on what we could do to avoid us breaking this again :) ?

toyamagu-2021:
@DOsinga 
I think enabling `string_slice` with Clippy is one approach. You can find more details here:
[https://rust-lang.github.io/rust-clippy/master/index.html#string_slice](https://rust-lang.github.io/rust-clippy/master/index.html#string_slice)

And you can enable it using this command:
`./bin/cargo clippy -- -W clippy::string_slice`

However, this currently generates a large number of errors, including false positives, making it difficult to fix immediately. If you'd like, I can create an issue for it and address it in a separate pull request.

toyamagu-2021:
CI is falling 🤔 
```
   Compiling goose-cli v1.0.35 (/home/runner/work/goose/goose/crates/goose-cli)
Error: The operation was canceled.

```

Issue #2582, opened 2025-05-19
Goose won't let me configure a AZURE_OPENAI_API_KEY when having azure CLI installed


strayer:
**Describe the bug**

I want to add Azure OpenAI as a provider to Goose. I have a specific API key that should be used to access Azure OpenAI but also use the Azure CLI locally for unrelated tasks. Goose doesn't even ask for an API key and apparently just tries to use the Azure CLI.

This causes issues because the user signed into Azure CLI does not have permissions to call Azure OpenAI.

Note: It does work when setting AZURE_OPENAI_API_KEY as an environment variable, but this would mean the secret isn't stored inside keychain.

**Expected behavior**

Goose should allow me to choose if I want to configure AZURE_OPENAI_API_KEY or use the Azure CLI authentication.

**Screenshots**

<img width="829" alt="Image" src="https://github.com/user-attachments/assets/7511dd38-6f25-40ae-add6-e31d47d117aa" />

**Please provide following information:**
 - **OS & Arch:** macOS Sequoia 15.4.1 (24E263)
 - **Interface:** goose CLI (installed via homebrew)
 - **Version:** 1.0.24
 - **Provider & Model:** Azure OpenAI

strayer:
I guess what I expect to run into when passing an AZURE_OPENAI_API_KEY as an environment variable explicity is this path: https://github.com/block/goose/blob/e63300887fe0a62f16759527df0291d4b73dc058/crates/goose-cli/src/commands/configure.rs#L253-L267

But since the AZURE_OPENAI_API_KEY is not an actual config key, this path is not triggered: https://github.com/block/goose/blob/e63300887fe0a62f16759527df0291d4b73dc058/crates/goose/src/providers/azure.rs#L226-L230

What I don't understand is why it is still prompted when neither the Azure CLI is installed nor the environment variable are set:

<img width="940" alt="Image" src="https://github.com/user-attachments/assets/50edce2c-6bc7-4869-9010-eeb27d6d3b49" />

NicoBonaminio:
Bump - this use to be available in previous versions of Goose and am now no longer seeing the option to add the API Key.

garyfeng:
I was able to get Azure OpenAI working for the Goose app (macos, v1.0.29) and `goose cli`. But I don't love the process.

```sh
export AZURE_OPENAI_API_KEY="xxxxx"
export AZURE_OPENAI_ENDPOINT="https://yyyyyy-eastus2.openai.azure.com"
[...]

PR #3265, opened 2025-07-06
feat: Add AZURE_OPENAI_API_KEY as a visible config parameter

mlazaric:
AZURE_OPENAI_API_KEY was previously an optional parameter and since only required parameters are visible in the UI, it was not possible to configure it. Let's make it required with an empty string as the default value. See also #2582.

Example from the UI
Screenshot 2025-07-06 at 17 10 53

### DIFF
--- a/crates/goose/src/providers/azure.rs
+++ b/crates/goose/src/providers/azure.rs
@@ -67,8 +67,10 @@ impl AzureProvider {
             .get_param("AZURE_OPENAI_API_VERSION")
             .unwrap_or_else(|_| AZURE_DEFAULT_API_VERSION.to_string());
 
-        // Try to get API key first, if not found use Azure credential chain
-        let api_key = config.get_secret("AZURE_OPENAI_API_KEY").ok();
+        let api_key = config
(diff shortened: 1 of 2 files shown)


angiejones:
I think this may confuse users who do not realize it's optional. Perhaps you can add "optional" in the label?

michaelneale (approved):
I can't manually validate it with azure but this seems to make sense

mlazaric:
> I think this may confuse users who do not realize it's optional. Perhaps you can add "optional" in the label?

Good idea! I agree that it is a bit confusing, although I'm not sure what would be the best way to clarify it, so open to suggestions. I think we could rename `AZURE_OPENAI_API_KEY` to `OPTIONAL_AZURE_OPENAI_API_KEY` to be more explicit about it, but then it would not use the existing `AZURE_OPENAI_API_KEY` environment variable if it was set. Alternatively, we could make the default string be something like `unused` or `optional`. Curious what people prefer?

michaelneale:
so if It has a value which says it is optional? (I am confused as to the aim now - ideally we wouldn't make an optional thing not optional?)

PR #3266, opened 2025-07-06
feat: bedrock image content support

### LINKED ISSUE #1481: Add support for image content in AWS Bedrock provider

Currently when using Puppeteer with `goose` I see:
```
─── puppeteer_screenshot | puppeteer ──────────────────────────
ERROR goose::agents::truncate: Error: Execution error: Image content is not supported by Bedrock provider yet
    at crates/goose/src/agents/truncate.rs:372
```

Able to configure and use image models

Using LiteLLM to be able to make AWS Bedrock look like OpenAPI

**Additional context**

mr-brobot:
might be a bit over-tested, feedback welcome :) 

closes #1481

### DIFF
--- a/crates/goose/examples/image_tool.rs
+++ b/crates/goose/examples/image_tool.rs
@@ -3,7 +3,7 @@ use base64::{engine::general_purpose::STANDARD as BASE64, Engine as _};
 use dotenv::dotenv;
 use goose::{
     message::Message,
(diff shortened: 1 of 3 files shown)


michaelneale (approved):
approving as seems isolated to bedrock (I don't have a way to manually validate this however, but seems good, as it was missing, and bedrock models can do this).

PR #3270, opened 2025-07-06
feat: add editable session description in header

anthony:
Add the ability to edit session descriptions directly in the header panel next to the directory picker. The session name only appears when it has been set (auto-generated by server or user-edited), not when it's just the session ID.

Note: Goose generated this code, I've reviewed/tweaked/tested etc.

## Features
- Inline editing: Click session name to edit directly in place
- Smart visibility: Only shows when session has a meaningful name (not just ID)
- Keyboard support: Enter to save, Escape to cancel, click outside to save
- Visual feedback: Loading states, hover effects, toast notifications
- Error handling: Graceful error handling with user-friendly messages

### Server-side (Rust)
- Add PUT /sessions/{session_id}/metadata endpoint to update session metadata
- Proper error handling and validation using existing session system

### Client-side (TypeScript/React)
- SessionHeader component: Self-contained component with inline editing
- useSessionMetadata hook: Custom hook for session metadata state management
- Clean integration with existing toast system and styling
- Proper z-index handling for macOS titlebar drag region compatibility

## Technical Details
- No complex state management or polling (queries the server after each message until a session description is set)
- Accessibility compliant with keyboard navigation
- Consistent with existing UI patterns and styling

goose-name-in-ui-1
goose-name-in-ui-2

### DIFF
--- a/crates/goose-server/src/routes/session.rs
+++ b/crates/goose-server/src/routes/session.rs
@@ -5,14 +5,14 @@ use crate::state::AppState;
 use axum::{
     extract::{Path, State},
     http::{HeaderMap, StatusCode},
-    routing::get,
+    routing::{get, put},
     Json, Router,
 };
 use goose::message::Message;
 use goose::session;
 use goose::session::info::{get_valid_sorted_sessions, SessionInfo, SortOrder};
 use goose::session::SessionMetadata;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
 #[derive(Serialize, ToSchema)]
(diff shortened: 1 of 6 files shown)


zanesq:
I like the feature! Sorry I didn't get to reviewing it sooner been slammed with the new ui launch. We don't have a formal process for defining what new features should look like but I think this is still valid and we should figure out the best way to get it in. I do know that we plan on having the session name visible in current chat in updated designs so we could either wait for that to get in or we could put it in chat history like a little pencil icon next to the name and it pops up a modal. Then when its available in chat we could port it over.

DOsinga:
Hi @anthony - sorry about our relative slow response here - there are two other people asking for this feature, so if you could help us resolve the merge conflicts (there are a lot, I know), we would love to land this. also like that you took the time to build the route to the server!

anthony:
@zanesq @DOsinga thanks for the feedback, I've opened a new PR - https://github.com/block/goose/pull/3819 - rather than updating this one seeing as the UI is quite different now, hope thats ok. I'll close this one.

PR #3194, opened 2025-07-01
Optimize provider system with connection pooling and enhanced retry logic

jackjackbits:
## Summary

This PR introduces significant performance optimizations to the provider system through connection pooling, enhanced retry logic, and various other improvements.

### Connection Pooling & HTTP/2
- Implemented shared HTTP client with connection pooling
- Enabled HTTP/2 support for request multiplexing  
- Added TCP optimizations (keep-alive, no-delay)
- Connection reuse reduces latency by ~50-100ms per request

### Enhanced Retry Logic
- Standardized retry behavior with exponential backoff
- Support for custom retry delay extraction (e.g., Azure's retry-after headers)
- Smart detection of retryable vs non-retryable errors
- Preserved provider-specific retry behaviors

### Request/Response Optimization
- Added automatic compression support (gzip, deflate, brotli)
- Implemented request size validation (10MB limit)
- Enhanced error messages with actionable suggestions
- Added request ID tracking for better debugging

### Provider-Specific Features Preserved
- Azure: Intelligent retry-after parsing from error messages
- GCP Vertex AI: Custom quota exhaustion messages with documentation links
- OpenAI: Configurable timeout support
- All providers maintain their unique error handling

## Performance Impact

- **Connection reuse**: Reduces latency significantly after first request
- **HTTP/2 multiplexing**: Allows multiple concurrent requests over single connection
- **Compression**: Reduces bandwidth usage by 60-80% for typical JSON responses
- **Smart retries**: Improves reliability without overwhelming rate limits

## Testing

- Added comprehensive unit tests for retry logic
- Tests for custom delay extraction
- Tests for error categorization
- Added connection pooling benchmarks

## Future Extensibility

Added traits for future enhancements:
- ProviderMetrics for telemetry integration
- ProviderCache for response caching
- Helper functions for request validation

All changes maintain backward compatibility while providing significant performance improvements.

PR #3271, opened 2025-07-07
Micn/patch optimize core systems

michaelneale:
this is resolving merge issues in https://github.com/block/goose/pull/3194


michaelneale:
closing this as I pushed it to the forked repo

PR #3274, opened 2025-07-07
feat: run sub recipe multiple times in parallel (Experimental feature)

lifeizhou-ap:
Adds ability to run sub-recipes multiple times with parallel execution, real-time progress tracking.

**Key Features:**
  - Parallel Execution of sub-recipes: Run multiple sub-recipe instances concurrently with configurable worker pools. 
     - You can run different sub-recipes in parallel, as long as you mentioned in your prompt. For example: analyze-pagerduty-incident and analyze-service-issues in parallel
     - You can also run the same sub-recipe in parallel.  For example: get the weather for 3 largest cities in Australia in parallel
  - Real-time Dashboard: Live progress tracking with task statistics and status updates

**Changes:**
  - Add `executions` and `timeout_in_seconds` in the sub_recipe section
     
        ```
        sub_recipes:
        - name: weather
          path: "{{recipe_dir}}/weather.yaml"
          values:
            date: "2025-01-01"
         sequential_when_repeated: false
          executions:
            parallel: true
            runs:
            - values:
              city: "Sydney"
            - values:
              city: "Melbourne"
        ```
      - `values` section:  always use the data in this section when running sub-recipe
      -  `sequential_when_repeated`: whether it should be sequential when running this sub recipe with different parameters. 
            - If `sequential_when_repeated: true`  means the sub-recipe should NOT run in parallel with different params even user requested in prompts. 
            - If `sequential_when_repeated` is false, or not specified, it is by default to run parallel unless user requested to run sequentially in prompts
      
  - New notification Event (notification_events.rs) for real-time task updates
  - Enhanced executor engine with parallel worker support and error handling
  - Console UI components for task execution dashboard (task_execution_display/)

**How it works**

  1. Input Phase

  - User provides recipe with multiple sub-recipe executions
  - Recipe parser creates individual tasks from the input
  - Tasks are queued for parallel execution

  2. Execution Phase

  - Executor spawns configurable number of workers
  - Workers pull tasks from queue and execute sub-recipes concurrently
  - Each worker runs independently with its own sub-recipe instance

  3. Monitoring Phase

  - Task Execution Tracker monitors all worker progress
  - Real-time events are generated for status changes
  - Statistics are calculated and updated continuously

  4. UI Updates

  - Notification events stream to the dashboard
  - Real-time display shows: completed, running, failed, pending counts
  - Users see live progress without polling

  5. Error Handling
  - Failed tasks are captured with detailed error context


crates/goose/src/agents/sub_recipe_execution_tool/types.rs
```diff
@@ -1,57 +1,129 @@
 use serde::{Deserialize, Serialize};
```
wendytang: (optional) nit: rename types.rs filename to task_types.rs

wendytang (commented):
I noticed in the demo you can run tasks without session: "running without session |". How is that specified to the task executor?

wendytang (approved):
🔥

jsibbison-square:
I wonder if it would be even more powerful if we could somehow move the recipe path into executions. Currently this PR  allows us to run one recipe with different params multiple times in parallel. But what if we had a predefined collection of recipes that we wanted to all run in parallel (effectively as one tool for the agent). That could be powerful.

Something like:
```
sub_recipes:
- name: code_review
  values:
    severity_level: "high"
  executions:
    parallel: true
    runs:
    - path: "{{recipe_dir}}/lint.yaml"
      values:
        tool: "clippy"
    - path: "{{recipe_dir}}/sql-bugs.yaml" 
      values:
        db: "postgres"
```

crates/goose-cli/src/recipes/extract_from_cli.rs
```diff
@@ -31,7 +31,9 @@ pub fn extract_recipe_info_from_cli(
                     let additional_sub_recipe = SubRecipe {
                         path: recipe_file_path.to_string_lossy().to_string(),
                         name,
+                        timeout_in_seconds: None,
```
jsibbison-square: I think it would be good to leave out this `timeout_in_seconds` feature until it becomes must have. Agents have very variable run times based on llm latency and what paths they take. I find 'max-turns' more reliable for limiting agent run length. Also timeout gets really weird when there are tool approvals for the user to do.

jsibbison-square:
There is also another perspective to take on this. You could argue that the agent can already choose to run multiple sub_recipes in parallel (and they are getting better at understanding prompts) so we should not add code complexity and schema rigidity in this PR. 

Instead we should rely entirely on the llm/agent to follow prompts and run tools in the correct order/parallelism natively.

There is the danger here that we are getting very close to building a limited version of prescribed workflows (particularly if we go [this route](https://github.com/block/goose/pull/3274#issuecomment-3059880198)). If we do decide we want to be able to program full on workflows then we would be better off building dedicated workflow logic outside of recipes and have recipes just as the _agent node_ (specification) building block.

If we go ahead with this functionality for recipes as is I think we should mark it as 'experimental' so that its more feasible to remove if we discover it doesn't align well with goose future recipe directions.

crates/goose-cli/src/session/task_execution_display/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+                "   ⚠️  {}{}\n",
+                error_preview.replace('\n', " "),
+                CLEAR_TO_EOL
+            ));
+        }
+    }
+
+    task_display.push_str(&format!("{}\n", CLEAR_TO_EOL));
+    task_display
+}
```
jsibbison-square, line 247: The rendering of subagents is something we should do an overall think about at some later stage as its a complicated experience to design.

crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs
```diff
@@ -7,16 +8,29 @@ use serde_json::{json, Map, Value};
...
 pub fn create_sub_recipe_task_tool(sub_recipe: &SubRecipe) -> Tool {
     let input_schema = get_input_schema(sub_recipe).unwrap();
     Tool::new(
         format!("{}_{}", SUB_RECIPE_TASK_TOOL_NAME_PREFIX, sub_recipe.name),
-        "Before running this sub recipe, you should first create a task with this tool and then pass the task to the task executor".to_string(),
+        format!(
+            "Create one or more tasks to run the '{}' sub recipe. \
+            Provide an array of parameter sets in the 'task_parameters' field:\n\
+            - For a single task: provide an array with one parameter set\n\
+            - For multiple tasks: provide an array with multiple parameter sets, each with different values\n\n\
```
jsibbison-square, line 24: I think the language 'objects' is clearer than 'parameter sets'.

Also for 
`- For multiple tasks: provide an array with multiple parameter sets, each with different values\n\n\`
I think this language is clearer
`- For multiple tasks: provide an array of objects, each with different values\n\n\`

crates/goose/src/agents/sub_recipe_execution_tool/lib/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+fn handle_response(response: ExecutionResponse) -> Result<Value, String> {
+    if response.stats.failed > 0 {
+        let failed_tasks = extract_failed_tasks(&response.results);
+        let error_summary = format_error_summary(
+            response.stats.failed,
+            response.stats.total_tasks,
+            failed_tasks,
+        );
```
jsibbison-square, line 119: If some tasks failed but some tasks succeeded it probably makes sense to also tell the agent which ones succeeded so it doesn't need to try those ones again.

jsibbison-square: It's hard for me to map what the contents of result will look like from the task will look like in response so maybe its all, just hard to know what it looks like for the agent.

lifeizhou-ap:
> I noticed in the demo you can run tasks without session: "running without session |". How is that specified to the task executor?

Hi @wendytang really good catch.  Currently we call `goose run -recipe` (with session) in parallel, the sessions are kicked off at the similar time and write to the same session file (same timestamp).  It will have an error as the session file prevents the concurrent write.  That's why I used "no-session" to avoid this issue.  Later on we will switch to the spawned sub agents and we won't have this issue any more

crates/goose-cli/src/recipes/extract_from_cli.rs
lifeizhou-ap: > 'max-turns' more reliable for limiting agent run length

Yes, I agree with this   

I guess there might be other scenarios that the sub-recipe stuck on some long running process (not because of the many turns), and it would be good to have timeout to have better failure handling for tasks orchestration.

lifeizhou-ap:
> I wonder if it would be even more powerful if we could somehow move the recipe path into executions. Currently this PR allows us to run one recipe with different params multiple times in parallel. But what if we had a predefined collection of recipes that we wanted to all run in parallel (effectively as one tool for the agent). That could be powerful.
> 
> Something like:
> 
> ```
> sub_recipes:
> - name: code_review
>   values:
>     severity_level: "high"
>   executions:
>     parallel: true
>     runs:
>     - path: "{{recipe_dir}}/lint.yaml"
>       values:
>         tool: "clippy"
>     - path: "{{recipe_dir}}/sql-bugs.yaml" 
>       values:
>         db: "postgres"
> ```

I feel this is like an inline subrecipe with subrecipe :) (ie. nested subrecipe) 

Currently subrecipe block maps to one recipe.  and the values are only for the subrecipe

```
values:
>     severity_level: "high"
```
This part will be a little bit confusing with this block.  This requires both `lint.yaml` and `sql-bugs.yaml` recipes have the same params `severity_level`.  

In the main branch, the parallel has been implemented for **different** recipes as below
> sub_recipes:
> - name: lint
>   path: "{{recipe_dir}}/lint.yaml"
>   values:
>     severity_level: "high"
>.     tool: "clippy"
> - name: sql-bugs
>   path: "{{recipe_dir}}/sql-bugs.yaml"
>   values:
>     severity_level: "high"
>.       db: "postgres"

crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs
lifeizhou-ap: I asked goose to come with this description. I felt it is clear :) 

I feel `objects` is a bit vague, it might be good to keep `parameter`, how about `parameter objects`?  the tool parameter is an array of "parameters" which are passed to the tasks

lifeizhou-ap:
> There is also another perspective to take on this. You could argue that the agent can already choose to run multiple sub_recipes in parallel (and they are getting better at understanding prompts) so we should not add code complexity and schema rigidity in this PR.
> 
> Instead we should rely entirely on the llm/agent to follow prompts and run tools in the correct order/parallelism natively.
> 
> There is the danger here that we are getting very close to building a limited version of prescribed workflows (particularly if we go [this route](https://github.com/block/goose/pull/3274#issuecomment-3059880198)). If we do decide we want to be able to program full on workflows then we would be better off building dedicated workflow logic outside of recipes and have recipes just as the _agent node_ (specification) building block.
> 
> If we go ahead with this functionality for recipes as is I think we should mark it as 'experimental' so that its more feasible to remove if we discover it doesn't align well with goose future recipe directions.

The schema level parallelism is an optional field for user to enforce the deterministic flow.  I guess you are concerning the field has been defined in the recipe schema for now, and later on it is hard to remove it if we don't need it?

> Marking this as experimental makes sense if we proceed — it'll give us room to iterate or revert if needed.

I’m on board with marking it experimental if we decide to include it, but I’m also up for talking more about where we see this going long term

crates/goose/src/agents/sub_recipe_execution_tool/lib/mod.rs
lifeizhou-ap: > If some tasks failed but some tasks succeeded it probably makes sense to also tell the agent which ones succeeded so it doesn't need to try those ones again.

Good idea.  I'll address this in a follow up to think about retry scenario

lifeizhou-ap: >  It's hard for me to map what the contents of result will look like from the task will look like in response so maybe its all, just hard to know what it looks like for the agent.

It is up to the task to return the task result.  In this case, it will return the final line of output of the `goose run --recipe` if the last line is a json, otherwise, it returns the output of the `goose run --recipe`.  This will be changed once we swapped to spawned sub agent as we relies on the result from `sub-agent` run

jsibbison-square:
Am I understanding correctly that this new `executions` and `runs` section of recipe is more like a tool 'prompt' influence for the agent? The only thing that's enforced is the number of params and maybe number of times the recipe will be called (runs).

If I understand right (I'm new to sub-agent code 😬 ) but it looks like the agent first calls a tool to 'create sub recipe tasks' and then with the response it is supposed to call a 'execute tasks' tool. But ultimately if the agent decides it wants to run the recipes sequential instead of parallel or run less/more subrecipes or override the params itself it total can because the there is not a lot of enforcement to follow the subrecipe config. It mostly generates prompt engineering for the agent (apart from the value defaults we provide).

If this is the case that these new schema fields are mostly optional guidance that the agent can potentially ignore and do differently then I think this is **a lot** of extra complexity and schema changes are painful to revert.

Sub_recipes can already be run in parallel with different parameters without this PR so I think we should rely on standard prompt engineering in the prompt/system prompt rather than trying to use this schema as hints for the agent.

I think there a lot of other useful things in this PR around subtask rendering in the cli and probably other refactoring but I think we should not merge the recipe schema changes 🙅 . But happy to chat and hear other opinions 😃  .

michaelneale:
conceptually - would you describe sub recipes as a way to do "sub agents" (just as that seems to be the topic of the day - ie many agents working together, would that be one way to describe this to users not familiar with goose terminology?)

lifeizhou-ap:
> conceptually - would you describe sub recipes as a way to do "sub agents" (just as that seems to be the topic of the day - ie many agents working together, would that be one way to describe this to users not familiar with goose terminology?)

Examples of use case
- In the mono repo build, there are 3 services that failed the build.  I can find that 3 build urls and use a sub-recipe (diagnose failure sub recipe) by passing the build url to diagnose these 3 service in parallel.
- I have a list of document link in a csv file, I would like to summarise each document (summarise document sub recipe)

crates/goose-cli/src/recipes/extract_from_cli.rs
jamadeo: Yeah you could probably leave it out -- the sub recipes themselves will be calling tools, and they'll have their own timeouts.

One thing we definitely don't want here is a default 5-minute timeout. IMO that doesn't make much sense for tool calls in general, but that's a separate issue.

crates/goose-cli/src/session/mod.rs
```diff
@@ -1031,7 +1038,19 @@ impl Session {
                                             } else {
                                                 progress_bars.log(&formatted_message);
                                             }
-                                        } else {
+                                        } else if let Some(ref notification_type) = _notification_type {
```
jamadeo: probably should rename this without the underscore since it's used now

crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs
```diff
@@ -7,16 +8,29 @@ use serde_json::{json, Map, Value};
 use crate::agents::sub_recipe_execution_tool::lib::Task;
 use crate::recipe::{Recipe, RecipeParameter, RecipeParameterRequirement, SubRecipe};
 
+use super::param_utils::prepare_command_params;
+
 pub const SUB_RECIPE_TASK_TOOL_NAME_PREFIX: &str = "subrecipe__create_task";
+const EXECUTION_MODE_PARALLEL: &str = "parallel";
+const EXECUTION_MODE_SEQUENTIAL: &str = "sequential";
```
jamadeo: should this be an enum?

crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs
```diff
@@ -7,16 +8,29 @@ use serde_json::{json, Map, Value};
...
 
 pub fn create_sub_recipe_task_tool(sub_recipe: &SubRecipe) -> Tool {
     let input_schema = get_input_schema(sub_recipe).unwrap();
```
jamadeo, line 17: I know it's not in this PR but this should probably handle an Err some other way than .unwrap()

crates/goose/src/agents/sub_recipe_execution_tool/sub_recipe_execute_task_tool.rs
```diff
@@ -2,35 +2,50 @@ use mcp_core::{tool::ToolAnnotations, Content, Tool, ToolError};
...
         SUB_RECIPE_EXECUTE_TASK_TOOL_NAME,
         "Only use this tool when you execute sub recipe task.
-EXECUTION STRATEGY:
-- DEFAULT: Execute tasks sequentially (one at a time) unless user explicitly requests parallel execution
-- PARALLEL: Only when user explicitly uses keywords like 'parallel', 'simultaneously', 'at the same time', 'concurrently'
+EXECUTION STRATEGY DECISION:
+1. PRE-CREATED TASKS: If tasks were created by subrecipe__create_task_* tools, check the execution_mode in the response:
+   - If execution_mode is 'parallel', use parallel execution
+   - If execution_mode is 'sequential', use sequential execution
+   - Always respect the execution_mode from task creation to maintain consistency
```
jamadeo: IMO the model shouldn't pick how the executor runs a set of tasks. The model will always get the results back together, so it only impacts the implementation details of the tasks themselves. If the model decides it needs the result of one task to run the next, it will have to separate them anyway.

crates/goose/src/agents/sub_recipe_execution_tool/sub_recipe_execute_task_tool.rs
```diff
@@ -2,35 +2,50 @@ use mcp_core::{tool::ToolAnnotations, Content, Tool, ToolError};
...
             "properties": {
                 "execution_mode": {
                     "type": "string",
                     "enum": ["sequential", "parallel"],
                     "default": "sequential",
-                    "description": "Execution strategy for multiple tasks. Use 'sequential' (default) unless user explicitly requests parallel execution with words like 'parallel', 'simultaneously', 'at the same time', or 'concurrently'."
+                    "description": "Execution strategy for multiple tasks. For pre-created tasks, respect the execution_mode from task creation. For user intent, use 'sequential' (default) unless user explicitly requests parallel execution with words like 'parallel', 'simultaneously', 'at the same time', or 'concurrently'."
                 },
                 "tasks": {
                     "type": "array",
```
jamadeo, line 41: from earlier discussion -- consider whether you can just have the model pass the ID and leave it at that, since the task has already been "created"

jamadeo (commented):
My main feedback is around the need for all the flexibility of parallel vs sequential execution within a task. IMO the model should not decide how that works -- it's just getting a batch of results anyway. Sub-recipes setting a flag to specify their execution when they get a batch of parameters seems OK.

crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs
lifeizhou-ap: good catch, i thought I have refactored, but missed out

crates/goose-cli/src/recipes/extract_from_cli.rs
shazraz: Agree with the tool-level timeout! I've been suffering from the default 5-minute timeout while using sub-recipes.

crates/goose-cli/src/recipes/extract_from_cli.rs
```diff
@@ -32,6 +32,7 @@ pub fn extract_recipe_info_from_cli(
                         path: recipe_file_path.to_string_lossy().to_string(),
                         name,
                         values: None,
+                        sequential_when_repeated: true,
```
jamadeo, line 35: This property name is a little confusing IMO, maybe use something like `allow_concurrent`?

lifeizhou-ap: Hi @jamadeo 

Thank you for the review and your time!

I thought of `allow_concurrent`, and i feel that it might have 2 concerns:

1.  `allow_concurrent: false` it might confuse people that the sub-recipe cannot be running concurrently with other sub-recipes.   The `...when_repeated` makes it clear that it cannot be run parallel when running repeatedly. 

2. By default the sub-recipe can be run in parallel.   Usually if the boolean attribute is not specified, the default value is false.  If  `allow_concurrent` is specified, people might be think the `allow_concurrent: false` 

Naming is usually hard.  Happy to change to a better name.

PR #3288, opened 2025-07-08
UI update with sidebar and settings tabs

zanesq:
UI update with sidebar settings tabs

image

Based on @nahiyankhan [branch](https://github.com/block/goose/tree/nk/vis-update)


The-Best-Codes:
This is very awesome looking! Assuming the image is up-to-date there are two main things I think could be changed:

1. I like rounding stuff, and this does have a nice aesthetic... but I don't like how those cards being rounded makes large gaps in the grid. I can't really think of what else to do there though :sweat_smile:
Maybe not round them and add borders (or just keep a small gap)?
2. Why is the microphone button so wide :weary:
Maybe I'm just being too picky lol.

Also kinda a third thing but I'm probably just confused: Why is there a large white space above the chat message input? If it serves no purpose, I would expand other parts of the UI (e.g., chat history) and get rid of it.

image

The-Best-Codes:
Also, I am assuming there is a margin or something here to make sure that the window controls don't overlap the sidebar button on macOS — don't forget to conditionally apply the margin so it won't show up on Windows & Linux!

image

zanesq:
@The-Best-Codes thanks for the feedback! Kinda agree on the home page rounding/extra space. Maybe we could remove the top border on the chat to help. Impartial on the mic button.. cc @spencrmartin 

Good call out we're going through windows and mac now and we'll keep it in mind.

michaelneale:
.bundle

.github/workflows/bundle-desktop-intel.yml
```diff
@@ -82,7 +82,8 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
         with:
-          ref: ${{ inputs.ref }}
+          # Only pass ref if it's explicitly set, otherwise let checkout action use its default behavior
```
michaelneale, line 85: is this change needed for these builds? (the aarch one isn't changed?)

zanesq: @Kvadratni made this change I'm not sure why its only for this one.

michaelneale:
.bundle

michaelneale:
Looking absolutely stunning! and this is on my existing data (which is a LOT) so nice to see it just works. 

some prelim testing notes - nothing major so far: 

* when I open create schedule - it has a scroll bar, but the section below the action buttons looks cut off (mostly as there is just white around the buttons) - when you scroll it is fine as there is a lot more content there. 
* adding custom extension shows a scroll bar but it basically shows all the content, maybe the window could be that tiny bit taller? (or perhaps some padding around the action buttons is larger than it needs to be?)
* not sure if you can repro, but I was clicking on my goosehints (when I had some) and they showed up blurry (!?): 
Configure your project's goosehints file to provide additional context to Goose
(but now are fine)


* if I click on "Chat" link - a lot of whitespace: 
Pasted Graphic 1

And not really quirks but now it is begging the question: 

* Should we have a schedule editor which is simpler
* Should we have a recipe editor (!) or recipe vibe coder (lots of "import" and yaml but not necessarily explaining what). - given these things are now elevated, perhaps at least need links to quality content if not some helpers? 

With the above, we can do a lot of what I was doing in goose perception too, there is tons of real estate for future extensions to add things to the bar on the left I think 

Super excited!

ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
```diff
@@ -0,0 +1,229 @@
+import React, { useEffect } from 'react';
+import { FileText, Clock, Home, Puzzle, History } from 'lucide-react';
+import { useNavigate } from 'react-router-dom';
+import { motion } from 'framer-motion';
```
The-Best-Codes: FYI, framer motion has changes to `motion`. You can run `npm i motion` and use `import abc from motion` instead.

.github/workflows/bundle-desktop-intel.yml
Kvadratni: I just made them all follow the same approach as the main build

The-Best-Codes:
I have like 20 emails from @zanesq's commits today, LOL. Zane is the GOAT :fire: 

+1 to Michael's comment. I did make a PR on top of this PR which addresses the "a lot of whitespace" issue he brought up. See https://github.com/block/goose/pull/3410

michaelneale:
so I don't think we can solve this as part of this PR but it did confuse me: 

Recipes:
image

Scheduled recipes: 
image


the yaml used in the second is the recipe format, the former is not the same format but this makes it more confusing if you try (like I did) to combine. Not sure of solution to this (other than unifying - the former recipes are for GUI and latter for autonomous actions - using the goose run recipe format?)

PR #3261, opened 2025-07-06
feat: add support for 'cu' command in extension installation

blackgirlbytes:
Container Use (cu) is featured in our content and Dagger's documentation:
- We have a [blog post](https://block.github.io/goose/blog/2025/06/19/isolated-development-environments) about Container Use with Goose
- We have a [YouTube video](https://www.youtube.com/watch?v=pGce9T4E5Yw) demonstrating it
- Dagger includes [deeplink installation](https://container-use.com/quickstart#method-3%3A-goose-desktop) instructions in their documentation

However, when users try to install cu-based extensions via the documented deeplinks, they encounter this error:
`Failed to install extension: Invalid command: cu. Only docker, jbang, npx, uvx, goosed are allowed.`

This PR solves that by adding : 
- Add 'cu' to allowedCommands list in deeplink validation
- Include 'cu' in shimPatterns for path resolution
- Add 'cu' to binaryPathMap in both utils.ts and extensions.tsx
- Fixes error: 'Invalid command: cu. Only docker, jbang, npx, uvx, goosed are allowed'

PR #3290, opened 2025-07-08
bugfix: revert cu pr

lilydelalande:
manually reverts https://github.com/block/goose/pull/3261


michaelneale (approved):
yeah , this is a good hotfix

PR #3296, opened 2025-07-08
feat: Recipe sets goose model, provider and temperature settings through gui

jsibbison-square:
Updates the gui to support the usage of 'settings' property when running a recipe:
```
  settings?: {
    goose_provider?: string;
    goose_model?: string;
    temperature?: number;
  };
```

The settings are only updated for the recipes goose session, it does not update the client config and it does not update the config on disc. So if a recipe changes the model, we will see the updated model for that recipe session. But if a user then does Cmd+N the new session will be back to the model specified in config. 

This keeps recipe execution as an independent run rather than changing a users default settings.

**Tested:**
- Deeplinks with & without settings
- Recipe library load recipe with & without settings

### DIFF
--- a/crates/goose-server/src/routes/agent.rs
+++ b/crates/goose-server/src/routes/agent.rs
@@ -66,6 +66,14 @@ struct UpdateProviderRequest {
 #[derive(Deserialize)]
 struct SessionConfigRequest {
     response: Option<Response>,
+    settings: Option<SessionSettings>,
+}
+
+#[derive(Deserialize)]
+struct SessionSettings {
+    goose_provider: Option<String>,
+    goose_model: Option<String>,
+    temperature: Option<f32>,
 }
 
 #[derive(Deserialize)]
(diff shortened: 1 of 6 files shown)


lilydelalande:
let's hold off on merging this before #3288 goes in since that one is huge and we want to avoid a rebase

DOsinga (commented):
This is an improvement, thanks for doing this.

Longer term we should probably redo how this all works though. What would you think of just having initializeSystem have take an optional recipe (with the recipe defined in the protocol) and have the rust code just figure out how to initialize the "agent"? we could then use the same code more or less for the CLI as for the Desktop and have just one way of setting things up for a recipe (and not a recipe)

We could at the same then verify the recipe in the rust code to make sure this is actually safe to run (and store if the user has run this one before etc)

DOsinga (approved):

jsibbison-square:
Yeah I think it might not be too much of a stretch to re-jig initializeSystem and the routes it calls. Think the recipe structure will still have to stay duplicated in ui and server as we currently have a concept of loading recipes without really saving them so server can't just look up via a reference (this is mostly for handling deeplink loading, and ui still needs to parse the recipe to handle recipe params).

DOsinga:
let's try to merge this if still relevant

jsibbison-square:
The conflicts and changes from the ui rewrite and recipe changes are too icky. Going to be easier to rewrite so I'll close this one.

PR #3304, opened 2025-07-08
Possible to disable random thinking messages

vilcans:
While these messages, like "Calculating the meaning of life", are amusing, being amused can sometimes be distracting. I prefer the UI to be consistent so I don't waste time on reading text that's not relevant to what I'm trying to accomplish. This patch adds the setting RANDOM_THINKING_MESSAGES, default true, that when changed to false replaces those messages with the less amusing text: "Thinking..."

### DIFF
--- a/crates/goose-cli/src/session/output.rs
+++ b/crates/goose-cli/src/session/output.rs
@@ -95,10 +95,17 @@ pub struct ThinkingIndicator {
 impl ThinkingIndicator {
     pub fn show(&mut self) {
         let spinner = cliclack::spinner();
-        spinner.start(format!(
-            "{}...",
-            super::thinking::get_random_thinking_message()
-        ));
+        if Config::global()
+            .get_param("RANDOM_THINKING_MESSAGES")
+            .unwrap_or(true)
+        {
+            spinner.start(format!(
+                "{}...",
+                super::thinking::get_random_thinking_message()
+            ));
+        } else {
+            spinner.start("Thinking...");
+        }
         self.spinner = Some(spinner);
     }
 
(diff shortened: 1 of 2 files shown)


lilydelalande:
Thanks for the contribution! Before we merge this, I'd like to raise a broader concern about our approach to configuration flags.

We're accumulating various configuration options without a clear documentation strategy or governance process. This creates two key problems:

1. Support burden: As maintainers, we can't effectively help users debug their setups when we're not aware of all possible configuration combinations and their interactions.
2. Technical debt: Without proper tracking, we risk creating conflicting states or incompatible combinations that could cause issues down the line.

This specific flag is benign ofc, but for current and future configuration additions, I think we need to establish:

- A centralized place to document all configuration options
- Clear naming conventions
- Documentation of expected behavior and any known interactions with other flags

cc @DOsinga

lily-de (dismissed):
This is great -- will approve after the docs get updated with information about how to use this variable here: https://github.com/block/goose/blob/main/documentation/docs/guides/environment-variables.md

angiejones:
> Thanks for the contribution! Before we merge this, I'd like to raise a broader concern about our approach to configuration flags.
> 
> We're accumulating various configuration options without a clear documentation strategy or governance process. This creates two key problems:
> 
> 1. Support burden: As maintainers, we can't effectively help users debug their setups when we're not aware of all possible configuration combinations and their interactions.
> 2. Technical debt: Without proper tracking, we risk creating conflicting states or incompatible combinations that could cause issues down the line.
> 
> This specific flag is benign ofc, but for current and future configuration additions, I think we need to establish:
> 
> * A centralized place to document all configuration options
> * Clear naming conventions
> * Documentation of expected behavior and any known interactions with other flags
> 
> cc @DOsinga

I added the variable to the Environment Variables guide

angiejones (approved):

Issue #3249, opened 2025-07-03
Session breaks when MCP server tool call responds with 2+ content elements


tadasant:
**Describe the bug**

Whenever I trigger an MCP tool call that has has a response like:

```
{
  "content": [
    {
      "type": "text",
      "text": "..."
    },
    {
      "name": "https://github.com/",
      "uri": "memory://cleaned/github.com__20250703193348846",
      "description": "Scraped content from https://github.com/",
      "mimeType": "text/html",
      "type": "resource_link"
    }
  ]
}
```

(Note the two results inside `content`)

I get this in Goose in response, and my session is unable to proceed:

```
Ran into this error: Request failed: Request failed with status: 400 Bad Request. Message: messages.1: tool_use ids were found without tool_result blocks immediately after: toolu_0158xtsJTKrVh4mhXdjh55Nm. Each tool_use block must have a corresponding tool_result block in the next message..
```

Screenshot:

![Image](https://github.com/user-attachments/assets/abb688fc-bccb-4a99-8a5c-325e6da916e5)

This same issue does not occur on a response from the same MCP server with a tweaked tool call parameter that results in just 1 content block:

```
{
  "content": [
    {
      "type": "text",
      "text": "..."
    }
  ]
}
```

Screenshot:

![Image](https://github.com/user-attachments/assets/66e6737f-6665-4147-95dc-f9d5ba998ee5)

**To Reproduce**
1. Set up the [pulse-fetch server](https://github.com/pulsemcp/mcp-servers/tree/main/productionized/pulse-fetch) as a custom extension in Goose. Use pulse-fetch@0.2.9
2. Prompt with something like `fetch https://github.com/ with default tool call params`
3. You should get the error

**Expected behavior**
No error; successful tool call.

**Please provide following information:**
 - **OS & Arch:** macOS M4 Max
 - **Interface:** UI
 - **Version:** 1.0.31
 - **Extensions enabled:** pulse-fetch
 - **Provider & Model:** Anthropic claude-sonnet-4

tadasant:
I had Claude Code investigate; this explanation seems plausible:

# Anthropic format provider incorrectly joins multi-element tool response content into single string

##  Description

  When MCP tools return responses with multiple content elements (e.g., both text and resource_link), Goose's Anthropic format provider incorrectly joins all text content into a single string, losing the structured format that the Anthropic API supports.

##  Root Cause

[...]

Issue #3150, opened 2025-06-29
Tokens are wrong (dies at 70%) (Anthropic/claude sonnet)


CyphrRiot:
I am using Anthropic Claude Sonnet 4.

I get this error (crash) when running:

```◐  Recycling random thoughts...                                                                                               2025-06-29T05:29:37.589740Z ERROR goose::agents::agent: Error: Request failed: 
Failed to parse response: Request failed: 
Provider returned error (code: 400) (status 400)
    at crates/goose/src/agents/agent.rs:770

Ran into this error: Request failed: Failed to parse response: 
Request failed: Provider returned error (code: 400) (status 400).

Please retry if you think this is a transient or recoverable error.
Context: ●●●●●●●○○○ 70% (140894/200000 tokens)
```

As you can see, the context is less than 200,000 tokens (Sonnet 4s token amount), but Goose crashes and runs out of tokens with a terrible error.

michaelneale:
@CyphrRiot was there more to the error than that? (ie anything else printed out)
were you using the anthropic provider (ie direct to anthropic?) - as we have limited ability to try all the providers so ideally need to know exactly which one (ie not via open router, databricks etc - I use it via databricks so naturally don't see that with the same model, but could be with anthropic provider if it is direct)

michaelneale:
I am not able to reproduce this, but have no more anthropic credits (if it is direct)

michaelneale:
I do occasionally see: 

```
Ran into this error: Request failed: Request failed with status: 400 Bad Request. Message: messages.45: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_01Sf6TDaMKK1y49vjqzAziPa. Each `tool_use` block must have a corresponding `tool_result` block in the next message..
```

michaelneale:
yeah sorry @CyphrRiot I am at a loss - there seems to be something there (the 70% is interesting, the 0.7 ratio) but I can't make it happen at my end. Are you able to try again with newest release? (and also share your config, and how you reproduce it if you can). I have asked anthropic bout that other error with tool results as it may be related and being hidden.

michaelneale:
A possible fix: https://github.com/block/goose/pull/3311 - but not 100% sure (as can't reliably reproduce anything). 

there is also https://github.com/block/goose/pull/3194 which may help - @CyphrRiot if you can repro, could provide you with some ad-hoc builds if you wanted to try either to see if it makes it go away.

michaelneale:
this should be resolved now

PR #3311, opened 2025-07-09
fix: correct tool use for anthropic

michaelneale:
discovered when looking at: https://github.com/block/goose/issues/3150 

should also fix: https://github.com/block/goose/issues/3249 

this may be the cause of that other issue too. 

```
Ran into this error: Request failed: Request failed with status: 400 Bad Request. Message: messages.45: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_01Sf6TDaMKK1y49vjqzAziPa. Each `tool_use` block must have a corresponding `tool_result` block in the next message..
```

This seemed to happen at odd times, and I guess could happen when compacting the session when the threshold is set (but couldn't reliably reproduce).


DOsinga (approved):
can you remove the comments? I don't think they are adding much. and in the spirit of slowly improving the code, we could extract some string constants here

michaelneale:
@DOsinga done

PR #3313, opened 2025-07-09
chore: cleaning out goose-llm

michaelneale:
in the case where goose llm isn't needed, this clears it out (should help with build times)

also has a tweak to ci that limits concurrency hopefully to reduce resource overloading that is happening.


DOsinga:
Good bye!

PR #3315, opened 2025-07-09
docs: add goose-mobile

zenmasterjobo:
This is a proof of concept to show how we could potentially fit goose mobile into the goose docs.

The old experimental section was a guide showing experimental features for goose. I have lifted that out of the guides section and created a new top level section just generally called experimental, and highlighted that this section is for both experimental projects and features surrounding goose.

Give it a look and let me know your thoughts. I don't have a great solution for where goose-mobile should live in the docs, but this is my first pass at a new idea. Hopefully we can find the best solution!

Thanks

### DIFF
--- a/documentation/docs/experimental/_category_.json
+++ b/documentation/docs/experimental/_category_.json
@@ -1,0 +1,8 @@
+{
+  "label": "Experimental",
+  "position": 7,
+  "link": {
+    "type": "doc",
+    "id": "experimental/index"
+  }
+}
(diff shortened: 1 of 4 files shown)


angiejones (approved):

blackgirlbytes:
- we just merged a PR that may conflict with your env variables PR
- Also since you're moving the toolshims doc to it's own separate space..we may need to update the path here as well https://github.com/block/goose/blob/2f36bac577cd09416a4cec86c40721cd81621c0f/documentation/docs/guides/managing-tools/index.md?plain=1#L36

angiejones:
@blackgirlbytes do you mind fixing this PR so we can get it merged?

blackgirlbytes:
> @blackgirlbytes do you mind fixing this PR so we can get it merged?

yes i can

PR #3324, opened 2025-07-09
fix the npx.cmd mapping issue

Kvadratni:
this resolves the issue reported on discord of npx extensions not working via deeplink


Kvadratni:
No rust changes so safe to override

PR #3328, opened 2025-07-09
Work around Gemini API tool call quirks

necaris:


michaelneale:
this looks good - but who has gemini access to validate - would be good to get a second check but tools and gemini is a known problem, so thanks @necaris

DOsinga (approved):

DOsinga:
can you check on the linter failure so we can get this in?

necaris:
Fixed and rebased, thanks @DOsinga for the nudge!

DOsinga:
ugh, I am sorry, fails on fmt

necaris:
Sorry, was out of town over the weekend, looks like it's green now after rebase and fix!

PR #3334, opened 2025-07-10
feat: add cache_control support for Claude models in OpenAI provider

### LINKED ISSUE #3333: Add cache_control support to OpenAI provider for Claude models
# Add cache_control support to OpenAI provider for Claude models

OpenAI-compatible services like LiteLLM support cache_control for Claude models to reduce API costs through prompt caching. The OpenRouter provider already includes this optimization for Anthropic models, but the OpenAI provider does not, missing cost optimization opportunities.

https://github.com/block/goose/blob/6d069095c90fff86d42dfb34123bea7cddfc2cde/crates/goose/src/providers/openrouter.rs#L128-L199

Add cache_control functionality to the OpenAI provider when the model name contains "claude":

1. Add cache_control markers to system messages
2. Add cache_control markers to the last two user messages
3. Add cache_control markers to the last tool definition

1. **Apply universally**: Could cause issues with non-Claude models
2. **Configuration-based**: Adds unnecessary complexity
3. **Model name detection**: Practical approach for LiteLLM and similar services

**Additional context**
- OpenRouter provider already implements this pattern in `update_request_for_anthropic`
- Backwards-compatible since unsupported services ignore unknown fields

HikaruEgashira:
Adds conditional cache_control functionality to OpenAI provider when model name contains "claude", enabling prompt caching for LiteLLM and other OpenAI-compatible services.

Fixes #3333

### DIFF
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use async_trait::async_trait;
 use reqwest::Client;
-use serde_json::Value;
+use serde_json::{json, Value};
 use std::collections::HashMap;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


michaelneale:
@HikaruEgashira thanks for this - can you run cargo fmt/check clippy etc - also, is will this work specifically with openrouter - what provider are you testing it with that uses the openai provider in this case?

michaelneale:
```
/// Update the request when using anthropic model.
/// For anthropic model, we can enable prompt caching to save cost. Since openrouter is the OpenAI compatible
/// endpoint, we need to modify the open ai request to have anthropic cache control field.
```

Does openrouter not do this as it is offering the openai compatible api? that seem really odd if they aren't? why would the offer an openai like api?

michaelneale:
any chance we can confirm before/after caching with openai to know it does need this header? (still seems odd to me)

HikaruEgashira:
I tested this program with LiteLLM. This header does not work with pure OpenAI but most OpenAI Compatible will work. I provide the usage from the system log.

before (1.0.29)
```
\"model\": \"anthropic.claude-3-5-haiku-20241022-v1:0\",\n  \"object\": \"chat.completion\",\n  \"system_fingerprint\": null,\n  \"usage\": {\n    \"cache_creation_input_tokens\": 0,\n    \"cache_read_input_tokens\": 0,\n    \"completion_tokens\": 554,\n    \"completion_tokens_details\": null,\n    \"prompt_tokens\": 5902,\n    \"prompt_tokens_details\": {\n      \"audio_tokens\": null,\n      \"cached_tokens\": 0\n    },\n    \"total_tokens\": 6456\n  }\n}","input_tokens":"5902","output_tokens":"554","total_tokens":"6456"},"target":"goose::providers::utils","span":{"name":"complete"},"spans":[{"name":"complete"}]}
```

after
```
\"model\": \"anthropic.claude-3-5-haiku-20241022-v1:0\",\n  \"object\": \"chat.completion\",\n  \"system_fingerprint\": null,\n  \"usage\": {\n    \"cache_creation_input_tokens\": 0,\n    \"cache_read_input_tokens\": 7593,\n    \"completion_tokens\": 68,\n    \"completion_tokens_details\": null,\n    \"prompt_tokens\": 8025,\n    \"prompt_tokens_details\": {\n      \"audio_tokens\": null,\n      \"cached_tokens\": 7593\n    },\n    \"total_tokens\": 8093\n  }\n}","input_tokens":"8025","output_tokens":"68","total_tokens":"8093"},"target":"goose::providers::utils","span":{"name":"complete"},"spans":[{"name":"complete"}]}
```

https://docs.litellm.ai/docs/completion/prompt_caching

crates/goose/src/providers/openai.rs
```diff
@@ -167,7 +240,13 @@ impl Provider for OpenAiProvider {
         messages: &[Message],
         tools: &[Tool],
     ) -> Result<(Message, ProviderUsage), ProviderError> {
-        let payload = create_request(&self.model, system, messages, tools, &ImageFormat::OpenAi)?;
+        let mut payload =
+            create_request(&self.model, system, messages, tools, &ImageFormat::OpenAi)?;
+
+        // Add cache_control for claude models (LiteLLM and other OpenAI-compatible services)
+        if self.model.model_name.to_lowercase().contains("claude") {
```
cgwalters, line 247: Shouldn't we add a method to the `Provider` trait for this instead?

michaelneale:
I think if liteLLM offers an openai api - that should include things like caching abstraction on it, that should be part of its job.

michaelneale (changes requested):
I think we should have a liteLLM provider specifically - we shouldn't be changing the openai provider for specific middleware routers which are lacking features (but if we have a liteLLM provider - that would be ideal - and leave the openai one as it is).

This could be done by cloning the openai provider, and adding in liteLLM specific code (and things like this) which is a better experience as well (as there likely will be other things like this which are liteLLM specific - as liteLLM is important enough I think to justify this)

HikaruEgashira:
OK! I'll add new provider. Thanks for reviewing.

PR #3342, opened 2025-07-10
Docs: Lead/worker model in Goose Desktop

emma-squared:
This PR adds instructions for using the lead/worker model with Goose desktop.

### DIFF
--- a/documentation/docs/tutorials/lead-worker.md
+++ b/documentation/docs/tutorials/lead-worker.md
@@ -2,6 +2,9 @@
 description: Enable multi-modal functionality by pairing LLMs to complete your tasks
 ---
 
+import Tabs from '@theme/Tabs';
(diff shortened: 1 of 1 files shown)


blackgirlbytes (changes requested):
What about these parts? theyre addressed for CLI but not the UI
Screenshot 2025-07-10 at 10 39 44 AM

blackgirlbytes:
we can even link them to this https://block.github.io/goose/docs/guides/environment-variables#leadworker-model-configuration

i think the doc needs a little love. it was written by the eng who made it but we can refine it.

emma-squared:
Thanks @blackgirlbytes - I can take another pass and update the doc this afternoon.

blackgirlbytes (approved):

PR #3345, opened 2025-07-10
refactor: streamline memory directory management

### LINKED ISSUE #3202: Prune empty goose/memory folders
When memory is turned on, Goose creates .goose/memory folders everywhere. While this doesn't cause problems, it does adds a trail of goose droppings eveywhere.

Goose should create .goose/memory folders only when it needs to store something there.

-

**Additional context**
-

sfc-gh-twhite:
Closes #3202 

These directories will now only be created if `remember` is called rather than upfront.

I added an extra check as well before clearing the memories to check the directory exists first, but this could be overly defensive.

### DIFF
--- a/crates/goose-mcp/src/memory/mod.rs
+++ b/crates/goose-mcp/src/memory/mod.rs
@@ -630,7 +630,7 @@ mod tests {
     fn test_lazy_directory_creation() {
         let temp_dir = tempdir().unwrap();
         let memory_base = temp_dir.path().join("test_memory");
-        
+
         let router = MemoryRouter {
             tools: vec![],
             instructions: String::new(),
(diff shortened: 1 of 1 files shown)


michaelneale:
@sfc-gh-twhite is there a way you can add a unit test for this?

sfc-gh-twhite:
> @sfc-gh-twhite is there a way you can add a unit test for this?

Done! Thank you for taking a look.

michaelneale (approved):
once clean - can merge in

michaelneale:
@sfc-gh-twhite mind running cargo fmt/clippy again? looks good to me

michaelneale:
and thanks - nice detail here.

sfc-gh-twhite:
> @sfc-gh-twhite mind running cargo fmt/clippy again? looks good to me

Done - sorry about that. 

> and thanks - nice detail here.

You're welcome!

PR #3347, opened 2025-07-10
feat: fetch openrouter supported models in `goose configure`

The-Best-Codes:
## Before:

You have to type out a model manually.

```bash
┌   goose-configure
│
│  Configure Providers
│
│  OpenRouter
│
●  OPENROUTER_API_KEY is already configured
│
│  Yes
│
◇  Enter new value for OPENROUTER_API_KEY
│  [REDACTED]
│
◇  Model fetch complete
│
◆  Enter a model from that provider:
│  anthropic/claude-3.5-sonnet (default)
└
```

## After:

Goose CLI will automatically fetch the list of models that support tool use from OpenRouter when you configure it as a provider:

```bash
┌   goose-configure
│
│  Configure Providers
│
│  OpenRouter
│
●  OPENROUTER_API_KEY is already configured
│
│  Yes
│
◇  Enter new value for OPENROUTER_API_KEY
│  [REDACTED]
│
◇  Model fetch complete
│
◆  Select a model:
│
│  ● ai21/jamba-1.6-large
│  ○ ai21/jamba-1.6-mini
│  ○ amazon/nova-lite-v1
│  ○ amazon/nova-micro-v1
│  ○ amazon/nova-pro-v1
│  ○ anthropic/claude-3-haiku
...
```

### DIFF
--- a/crates/goose/src/providers/githubcopilot.rs
+++ b/crates/goose/src/providers/githubcopilot.rs
@@ -427,42 +427,4 @@ impl Provider for GithubCopilotProvider {
         emit_debug_trace(&self.model, &payload, &response, &usage);
         Ok((message, ProviderUsage::new(model, usage)))
     }
-
-    /// Fetch supported models from GitHub Copilot API
-    async fn fetch_supported_models_async(&self) -> Result<Option<Vec<String>>, ProviderError> {
-        let (endpoint, token) = self.get_api_info().await?;
-        let url = url::Url::parse(&format!("{}/models", endpoint))
-            .map_err(|e| ProviderError::RequestFailed(format!("Invalid models URL: {e}")))?;
-
-        let response = self
-            .client
-            .get(url)
-            .headers(self.get_github_headers())
-            .header("Authorization", format!("Bearer {}", token))
-            .send()
-            .await?;
-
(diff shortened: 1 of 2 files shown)


The-Best-Codes:
@michaelneale The `fetch_supported_models_async` function is a public function. It was added to `base.rs` in a previous PR. Providers can optionally include that function to enable supported model fetching.

https://github.com/block/goose/blob/67664a2a517e3138f6cbb0dcb0ed8dd49b99f3ab/crates/goose/src/providers/base.rs#L251-L253

It gets called in `configure.rs`:

https://github.com/block/goose/blob/67664a2a517e3138f6cbb0dcb0ed8dd49b99f3ab/crates/goose-cli/src/commands/configure.rs#L330-L338

michaelneale (approved):
nice isolated change, will help with experience with open router

michaelneale (approved):

michaelneale:
@The-Best-Codes merged - wonder if future enhancement is to have some UI that lets people find.. as there is a lot! (but no idea what that would look like)

The-Best-Codes:
@michaelneale Awesome!

> have some UI that lets people find

Do you mean in Goose desktop or a TUI? In the terminal, maybe type to search the list of models?

PR #3349, opened 2025-07-10
feat: subagent handle missing extensions 

wendytang:
* Makes the extension manager thread safe. 
* Adds list of missing extensions to subagent's system prompt


DOsinga:
I'm closing this for now since it is in draft and not been updated for more than two weeks. can always reopen

PR #3352, opened 2025-07-10
Update windows CLI installation instruction

Calidus:
Adds a note that curl is an alias for Invoke-WebRequest on most windows systems.


Calidus:
Closing this, I think the Windows CLI instructions need some larger changes

PR #3353, opened 2025-07-10
feat: subagent handle provider specified in recipe

wendytang:
Allows specification of custom provider and model when executing subagent task

Specification:
``` 
# brilliant_recipe.yaml
...
settings:
  goose_provider: "anthropic"
  goose_model: "claude-3-7-sonnet-latest"
  temperature: 0.7
```

Subagent notification logs:
```
... main agent uses claude-4-sonnet, subagent uses claude-3-7-sonnet-latest...

🔍 Missing required extensions: ["weather", "linear"]
🤖 Subagent created and ready
💭 Processing: Get the weather for Los Angeles, California **(via claude-3-7-sonnet-latest)**
Processing request
💭 Turn 1/10
🔧 Using tool: developer__shell
🔧 Tool developer__shell completed successfully
🤖 Responded: ## Weather Report for Los Angeles, California
```


crates/goose/src/agents/subagent_handler.rs
```diff
@@ -55,11 +57,82 @@ impl Agent {
...
 
-        // Get the provider from the parent agent
-        let provider = self
-            .provider()
-            .await
-            .map_err(|e| ToolError::ExecutionError(format!("Failed to get provider: {}", e)))?;
+        // Determine which provider to use
+        let provider = if let Some(recipe_name) = &args.recipe_name {
+            // Load the recipe to check if it specifies a provider
+            match manager.load_recipe(recipe_name).await {
```
lifeizhou-ap, line 63: This would not work sometimes as recipe could be a template that accept parameters to be rendered. 

Recipe example with parameters:  https://github.com/block/goose/pull/2395

One option that could be working nicely: 

Shall we change the `subagent` constructor with parameters `system message`, `prompts`, `extensions`, `provider`, `model` and `temperature` (we also can group some of them into struct logically).  In this case sub-agent does not need to know the recipe. 

Currently render recipe logic is in goose-cli and I can some of the logic into `goose` package.  Once the recipe is parsed, we can pass those attributes to sub-agent.  WDYT?

We can also sync on your Monday afternoon if you are available.   Thank you!

wendytang: > Shall we change the subagent constructor with parameters system message, prompts, extensions, provider, model and temperature (we also can group some of them into struct logically). In this case sub-agent does not need to know the recipe.

In essence, recipe will be rendered with parameters and then the fields will be passed to subagent struct? That sounds good to me.

PR #3354, opened 2025-07-10
feat: add download_cli.ps1 file for windows

The-Best-Codes:
> [!IMPORTANT]
> This PR isn't ready to merge yet! I'm waiting for feedback from the team & community. 

The current installation method for Windows requires a `download_cli.sh` script to be run, which requires WSL or Git Bash on Windows (it's not native) and frequently has issues.
This PR creates a native `download_cli.ps1` script for Windows which installs Goose the same way as `download_cli.sh` but using Windows supported PowerShell commands.

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -627,9 +627,9 @@ impl Session {
                         &Message::assistant().with_text("Chat context cleared."),
                         self.debug,
                     );
-                    if self.session_file.exists() {
-                        std::fs::remove_file(&self.session_file)?;
-                        std::fs::File::create(&self.session_file)?;
+                    if let Some(file) = self.session_file.as_ref().filter(|f| f.exists()) {
+                        std::fs::remove_file(file)?;
+                        std::fs::File::create(file)?;
                     }
                     continue;
                 }
--- a/ui/desktop/openapi.json
+++ b/ui/desktop/openapi.json
@@ -10,7 +10,7 @@
     "license": {
       "name": "Apache-2.0"
     },
-    "version": "1.0.35"
+    "version": "1.0.36"
   },
   "paths": {
     "/agent/tools": {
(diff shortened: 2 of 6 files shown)


The-Best-Codes:
I have a question: Should I also update the docs in this PR to use the PS1 script (assuming the Goose team accepts it at all), or will somebody else do that?

The-Best-Codes:
Note: According to other contributors:
> The VS Code extension is looking for resources/bin/goosed in a particular set of directories.

I need to test that on Windows with this new script.

cbruyndoncx:
I have goose installed already with git bash, so got this error:
Downloading stable release: goose-x86_64-pc-windows-gnu.zip...
Download completed successfully.
Created temporary directory: C:\Users\bruyn\AppData\Local\Temp\goose_install_1761177838
Extracting goose-x86_64-pc-windows-gnu.zip to temporary directory...
Extraction completed successfully.
Found goose-package subdirectory, using that as extraction directory
Moving goose to C:\Users\bruyn\.local\bin\goose.exe
C:\Users\bruyn\Downloads\download_cli.ps1 : Failed to move goose.exe to C:\Users\bruyn\.local\bin\goose.exe. Error: Cannot create a file when that file already exists.
At line:1 char:1
+ .\download_cli.ps1
+ ~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,download_cli.ps1

cbruyndoncx:
Used the developer shell only in that version i got gemini cli working
Looking good opens the configure dialogbox 
And i did the in session bit, and the user path setup, no errors
image

cbruyndoncx:
@The-Best-Codes the scripts works fine for a new installation
for existing installation, it does not overwrite in .local/bin , but does pick up that there is a config.yaml already.

The-Best-Codes:
@cbruyndoncx I will fix that later today! Thanks for testing! :heart:

The-Best-Codes:
Got busy on the weekend. Will try to address some things tomorrow...

The-Best-Codes:
@cbruyndoncx The issue you had should be fixed.

The-Best-Codes:
I just force pushed new signed-off commits. The DCO check should pass now.

Kvadratni (approved):

PR #3356, opened 2025-07-10
Add support for setting `goose_lead_model` from a recipe

wclausen-square:
This builds on https://github.com/block/goose/pull/3296 to support passing a lead model from a recipe to allow recipes to override default model configs and use the LeadWorkerProvider.

The existing code for creating a LeadWorkerProvider pulls env vars in the factory method, which makes it hard to override the values for a specific recipe's session. To address this limitation, I added a new factory method that allows providing construction dependencies for the specific provider type based on what that provider needs.

This is my first time writing Rust, so I'm not sure if my changes are idiomatic/proper, but I made a best effort there. Shout out  to Goose for helping me understand the codebase and Rust syntax.

### DIFF
--- a/crates/goose-server/src/routes/agent.rs
+++ b/crates/goose-server/src/routes/agent.rs
@@ -9,7 +9,11 @@ use axum::{
 use goose::config::Config;
 use goose::config::PermissionManager;
 use goose::model::ModelConfig;
-use goose::providers::create;
+use goose::providers::{
+    create,
+    create_by_type,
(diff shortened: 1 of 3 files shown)


michaelneale:
I like this idea of this @jsibbison-square @DOsinga  - I'm not sure of the direction of setting providers and models in recipes, but if ti is to be in there, makes sense to expose things like this I think?

DOsinga (approved):

PR #3374, opened 2025-07-11
Attempting to derive if the remote host does not support dynamic clie…

jeremiah1066:
…nt registration


DOsinga:
closing stale drafts

PR #3384, opened 2025-07-12
Improve error message for Gemini Cli provider

cbruyndoncx:
Similar to the error handling implemented for Claude code


yingjiehe-xyz:
can you format the code to fix the lint error?

DOsinga:
mind running the cargo fmt?

DOsinga:
thanks for your contribution @cbruyndoncx - I'm going to close this and add this to something else so it gets merged rather than being blocked on the fmt. but your code will live on

Issue #3367, opened 2025-07-11
Allow configuring behaviour of `Provider::complete` based on prompt purpose.


lakshya-sky:
**Please explain the motivation behind the feature request.**
I want to use a smaller model when goose tries to generate description. Because using a big model would be costly and slower compared to a smaller model for a task which is not essential.

**Describe the solution you'd like**
If we can add a new enum parameter(e.g. `RequestContext`) to `Provider::complete`. 

Right now Snowflake provider uses string matching to figure out if the complete request is for generating description. 

https://github.com/block/goose/blob/626f00e2909df98b281e7cc1cf96820ddf36d0c3/crates/goose/src/providers/formats/snowflake.rs#L330C2-L340C7

but instead can written as:

```
enum RequestContext{
  Normal,
  Description,
  Plan,
}

// In crates/goose/src/providers/formats/snowflake.rs
/// Create a complete request payload for Snowflake's API
pub fn create_request(
    request_context: RequestContext,
    model_config: &ModelConfig,
    system: &str,
    messages: &[Message],
    tools: &[Tool],
) -> Result<Value> {
  ...
  match request_context{
    RequestContext::Description => {fomat_tools(&[])},
    _ => {format_tools(tools)}
  }
  ...
}

// In crates/goose/src/providers/anthropic.rs
 #[tracing::instrument(
        skip(self, system, messages, tools),
        fields(model_config, input, output, input_tokens, output_tokens, total_tokens)
    )]
    async fn complete(
        &self,
        request_context: RequestContext, 
        system: &str,
        messages: &[Message],
        tools: &[Tool],
    ) -> Result<(Message, ProviderUsage), ProviderError> {
        match request_context{
          RequestContext::Description =>{
            let payload = create_request(&ModelConfig::new("claude-3-5-haiku-latest"), system, messages, tools)?;
          },
          _ => let payload = create_request(&self.model, system, messages, tools)?;
        }
        ...

    }
```

**Additional context**
This might not be the best terminology or way. Please provide your suggestions as how can it be improved.

- [x] I have verified this does not duplicate an existing feature request

DOsinga:
this sounds reasonable. do you have a matching pr? maybe just for anthropic for now?

DOsinga:
never mind, I see

DOsinga:
so we now do have exactly this - the ability to use a smaller model to generate a description (and also do compaction). so I am going to close thi.

PR #3388, opened 2025-07-13
feat: Add RequestPurpose parameter to Provider trait for context-aware model/tool selection

lakshya-sky:
This implements issue #3367 by adding a `RequestPurpose` enum to the Provider trait's `complete` method, enabling providers to select different models or configurations based on the context of the request.


michaelneale:
I think there was some research on for model routing which would do a more general version of this - ie know the intent of the task (even in the middle of a session)

DOsinga:
I like the idea here, but why not just add a summarize() method and a generate_description() method to the Provider with a default implementation that then can be overwritten on a per provider basis? 

we could start with antrhopic and then just add a .with_model("...") method that we would use for generate_description() to pick the cheapest model here. that way the change is so giagantic and doesn't change the contract

DOsinga:
we now have a generate_session_name on the provider class. you should use that

DOsinga:
I'm going to close this for now - we would love to have this implemented based on the new structure though! tag me directly and we'll get it in

Issue #3382, opened 2025-07-12
Tree-sitter-based Repo Map


dcieslak19973:
**Please explain the motivation behind the feature request.**
Goose currently relies on ripgrep for code navigation and search, which is fast but limited to text/regex matching. Adding Tree-sitter-based symbol extraction would enable Goose to build a structured, language-aware repo map, unlocking advanced features like class hierarchy analysis, symbol-based navigation, and more precise LLM context.

**Describe the solution you'd like**
Integrate Tree-sitter (with Rust bindings and language grammars) into Goose to:
- Parse source files and extract classes, functions, methods, and their relationships
- Build a repo map (e.g., JSONL) with symbol metadata (type, name, signature, parent, overrides, etc.)
- Add a CLI command (e.g., `goose repo index`) to generate/update the repo map
- Allow the LLM to use this map for context, navigation, and code understanding

**Describe alternatives you've considered**
- Continue using ripgrep (text-based, less precise)
- Integrate Zoekt (external dependency, less structural awareness)
- Use ctags (limited symbol info, not AST-based)

**Additional context**
- Tree-sitter is Rust-friendly and easy to package
- Enables features like class hierarchy mapping, override detection, and more
- Can be combined with ripgrep for hybrid workflows

- [x] I have verified this does not duplicate an existing feature request

DOsinga:
This is cool. Two questions though:
* Should this be build into the CLI? That precludes it from working in the desktop
* Could this be an MCP server

dcieslak19973:
i only use the goose cli, so that's where i put it....if there's a better place that both the cli and desktop could use it great.

which is probably where you're going with the mcp.  based on my experiences where i work, i'm leary of local (stdio) mcps which would be the natural place for something like this potentially.  i also think, eventually, that having this sort of data available in memory will help it.  I believe, but haven't looked closely, that `aider` does something like this with `repomap` and keeps it in memory

taqtiqa-mark:
@DOsinga I believe the MCP is where this fits more naturally. For example https://github.com/wrale/mcp-server-tree-sitter
@dcieslak19973 can this be closed? There is [repomix](https://github.com/yamadashy/repomix) and you can also probably setup the tree-sitter MCP with the memory, knowledge-graph, etc. MCPs to work exactly how you want?

DOsinga:
that makes a lot of sense to me. especially since the current approach seems to be cli only. with mcp it would work for both.

[...]

PR #3389, opened 2025-07-13
Tree-sitter Indexer

dcieslak19973:
Addresses #3382 

**Summary of Changes**

1. Tree-sitter Code Indexing Feature (Goose CLI)

* **New Feature:** Adds experimental Tree-sitter-based code indexing to Goose CLI, supporting Rust, Python, JavaScript, TypeScript, Go, C++, Java, C#, and Swift.
* **New Command:** `goose repo` index for indexing repositories and generating a .goose-repo-index.jsonl file with extracted code entities (classes, functions, relationships, docstrings, etc.).
* **New Module:** `repo.rs` implements the indexer.
* **New Documentation:** `tree-sitter-indexing.md` details supported languages, extracted entities, and upgrade guidance.

2. Dependency Updates

* `Cargo.toml` / Cargo.lock:
* * Adds dependencies for `tree-sitter` and language grammars (see above).
* * Adds `walkdir` and `ignore` for efficient file traversal.
* * Updates and reorders some existing dependencies for compatibility.

3. CLI Enhancements

* **New Subcommand:** Adds Repo subcommand to Goose CLI, with an index action for repository indexing.
* **Refactoring:** Updates CLI argument parsing to support new subcommands and options.

4. Example and Test Data

* **Adds Example Indexed Data:** `repo-index.jsonl` and language-specific example source files for testing and demonstration.

5. Internal Improvements

* **Entity Extraction:** Richer extraction for Go, Python, and C++ (fields, generics, decorators, etc.).
* **Parent/Child Relationships:** Improved tracking of parent entities for functions/methods.
* **Docstring/Comment Extraction:** For Python and Rust, docstrings and doc comments are included in the index.

**Impact**

* **LLM/AI Assistant Support:** Lays the foundation for advanced code navigation, search, and summarization features.
* **Multi-language Support:** Enables entity-level code understanding for a wide range of languages.
* **Experimental:** Marked as experimental; subject to change as Tree-sitter ecosystem evolves.


dcieslak19973:
@michaelneale - I think the code is currently in a state ready for discussion, both of the code and the concept.  I'm open to feedback; I've not done benchmarks, but conceptually the idea is to help goose get better context into the LLM.  Potentially down the road, goose could keep this data (or similar data) in memory (or update the file in the background) as it makes changes to a codebase.

dcieslak19973:
Any thoughts on accepting this or changes necessary to accept?

michaelneale:
@dcieslak19973 I like this idea - but wonder if it can go in the main goose crate potentially so isn't limited to the GUI - would be nice ot be able to make these maps from anywhere - you interested in that?

dcieslak19973:
Yes, I can look into doing that.  Not sure if I'll have time this weekend, but probably within the next 2 weeks

michaelneale:
thanks - will keep an eye open, curious how this goes, good research IMO

dcieslak19973:
I'm probably not going to be able to get to updating this until Labor Day at the earliest.

One direction I could see this going is to use Zoekt (the tech SourceGraph developed) to keep the codebase in memory in a very searchable format.  I don't know if a Rust implementation exists, so there'd probably be some additional effort involved in porting the Go to Rust.

See: https://github.com/sourcegraph/zoekt/tree/main

dcieslak19973:
I found some time late last night and this morning to CoPilot thru the request to add this to the UI as well

Justfile
```diff
@@ -117,6 +117,12 @@ copy-binary-windows:
     @powershell.exe -Command "if (Test-Path ./target/x86_64-pc-windows-gnu/release/goosed.exe) { \
         Write-Host 'Copying Windows binary and DLLs to ui/desktop/src/bin...'; \
         Copy-Item -Path './target/x86_64-pc-windows-gnu/release/goosed.exe' -Destination './ui/desktop/src/bin/' -Force; \
+        if (Test-Path ./target/x86_64-pc-windows-gnu/release/goose.exe) { \
```
michaelneale, line 120: not sure if this is related to this change?

michaelneale:
@dcieslak19973 ok this looks like a good start, took 1s to index goose, I tried a very very very large mono repo and was a minute and 300M or so, which is acceptable. 

What would be next for consuming this index for a codesearch tool to expose it via an appropriate MCP? something like Zoekt but perhaps rust native? any ideas?

Justfile
dcieslak19973: Oops, missed CoPilot going a little crazy there.

dcieslak19973:
> @dcieslak19973 ok this looks like a good start, took 1s to index goose, I tried a very very very large mono repo and was a minute and 300M or so, which is acceptable.
> 
> What would be next for consuming this index for a codesearch tool to expose it via an appropriate MCP? something like Zoekt but perhaps rust native? any ideas?

Admittedly, I don't have a great plan for what's next.  Some things, which are probably not mutually exclusive:

1. Do additional research/analysis on how `aider` uses this technique.  This implementation is really just based off reading https://aider.chat/docs/repomap.html . 
2. The jsonl file isn't meant to be permanent, though it offers a potentially easy integration path when `goose`'s LLM wants to do a code search - in theory grep/rip-grep/etc. would find contents in this file.
3. So, replacing (augmenting maybe) the jsonl with getting this into memory, probably some form of a Vector Database where the additional metadata (line numbers, etc.) could be stored with the embeddings.  then, present `goose` with a `codesearch` tool that would do an embedding search
4. Probably some way to do an incremental update of 3 as `goose` changes the codebase.  To start, the easiest thing might be to just trigger full rebuilds and then look at smarter things to do on larger codebases that would benefit from incremental.
5. Potentially add something like PageRank (on functions/classes/etc.) to this process.  When `goose` wants to maybe understand the codebase at a high-level, feeding in (maybe thru another tool) the "important" features of the codebase would be useful
6. If you follow `claude` code's work, they are "spawning" (I use this loosely as I'm not totally sure how they implement it) agents.  Goose could follow suite; maybe there'd be a background "agent" that updates the vector-store after changes, etc.
7. I've mentioned Zoekt, I think that would be a separate implementation to this, and one where a "codesearch" tool/agent could use a BOTH-AND search over the in-memory vectorstore and an in-memory Zoekt instance.

dcieslak19973:
@michaelneale - I think this is closer to a better form.  There's probably still some stuff to clean up in the cli (maybe add a slash command) and the UI

documentation/docs/experimental/repo-search-with-tree-sitter-indexing.md
```diff
@@ -0,0 +1,241 @@
+# Repository Search with Tree-sitter Indexing (Experimental)
```
angiejones: missing frontmatter:


```suggestion
---
title: Repository Search with Tree-sitter Indexing
sidebar_label: Repository Search with Tree-sitter Indexing
sidebar_position: 5
---
```

documentation/docs/experimental/repo-search-with-tree-sitter-indexing.md
```diff
@@ -0,0 +1,241 @@
...
+
+> Experimental repository indexing: graph + PageRank + blended fuzzy symbol search across multiple languages.
```
angiejones:
```suggestion

```

documentation/docs/experimental/repo-search-with-tree-sitter-indexing.md
```diff
@@ -0,0 +1,241 @@
...
+- Incremental (build-time) Tree-sitter based parsing across multiple languages.
+- In-memory graph model (calls, containment, imports) with weighted PageRank centrality.
+- Fuzzy + rank blended symbol search with traversal (callers/callees) expansion.
+- Agent tools (`repo__build_index`, `repo__search`, `repo__stats`) for autonomous workflows.
+- CLI integration for ad‑hoc exploration.
+
+This enables fast localization of relevant code entities, supports higher-level reasoning (e.g. impact analysis), and improves tool routing by surfacing influential symbols.
+
+## Feature Flags & Stability
+All functionality is gated behind the Cargo feature `repo-index` (plus language specific `tree-sitter-*` features). The system is experimental; APIs and output schemas may evolve.
```
angiejones: can you provide more info here? how does one enable this?

also all of our experimental features should be enabled when env var `ALPHA_FEATURES` is true

documentation/docs/experimental/repo-search-with-tree-sitter-indexing.md
```diff
@@ -0,0 +1,241 @@
+# Repository Search with Tree-sitter Indexing (Experimental)
+
+> Experimental repository indexing: graph + PageRank + blended fuzzy symbol search across multiple languages.
+
+## Overview
```
angiejones: Need more of an opening about what this is and why would someone want to use this

documentation/docs/experimental/repo-search-with-tree-sitter-indexing.md
angiejones, line 1: Hi! Thanks for the comprehensive documentation. However, I think this needs to be restructured for our user-facing docs site. The current version reads more like internal technical documentation than a user guide.

**Key issues:**
- **Missing the "why"**: Users won't understand what problem this solves or when they'd use it
- **Too implementation-focused**: Details about PageRank algorithms, data models, and Tree-sitter versions aren't relevant to users
- **No clear use cases**: When would someone run `goose repo index`? What does it enable?

**Suggested restructure:**
1. **Start with the problem**: What does this solve? (e.g., "Help Goose better understand your codebase structure")
2. **Simple use case**: "Run this when you want Goose to have deeper knowledge of your project"
3. **Basic usage**: Simple example with expected output
4. **Benefits**: What can Goose do better after indexing?
5. **Supported languages**


**Move to separate docs (if necessary):**
- Implementation details
- Technical architecture
- Performance tuning
- Contributing guidelines

Could you revise this to focus on the user experience?

documentation/docs/experimental/repo-search-with-tree-sitter-indexing.md
dcieslak19973: Thanks for the feedback; per your longer comment, I've reworked the documentation

documentation/docs/experimental/repo-search-with-tree-sitter-indexing.md
dcieslak19973: Thanks for the feedback; per your longer comment, I've reworked the documentation

documentation/docs/experimental/repo-search-with-tree-sitter-indexing.md
dcieslak19973: Thanks for this feedback. I've split this documentation into a more user-oriented one and a more technical/implementation detail one

michaelneale:
@dcieslak19973 we also recently removed the lancedb vector store as a required goose dependency so that might be causing some of the conflicts here. BTW I am working on an mcp that will use it which is external to goose, which could possibly be complemented by tree sitter as well - would you consider collaborating there?

dcieslak19973:
Could be.  I also wasn't using devcontainers which may be part of it as well (not to mention my inexperience with Rust and typescript and the goose codebase).

While I'm somewhat sympathetic to the idea of making this an MCP in terms of modularity, etc.  I do feel strongly that this should be part of the core functionality of `goose`, maybe as part of the "developer" built-ins, perhaps as I view `goose` as a tool similar to claude-code, aider, OpenCode, etc. and less as a general-purpose AI tool.  Ultimately, the decision is yours/your team's as project maintainer.

For reference, `aider` includes `repomap` functionality directly in their codebase:

* https://github.com/Aider-AI/aider/blob/main/aider/repomap.py

OpenCode (the sst version) includes tools like `grep` and even LSPs as part of their core functionality:

* https://github.com/sst/opencode/blob/dev/packages/opencode/src/tool/grep.ts
* https://github.com/sst/opencode/tree/dev/packages/opencode/src/lsp

And I do believe that this functionality should represent a huge improvement over simple greps.  Ideally, we'd run the goose coding benchmarks with this functionality enabled and disabled and get a comparison (and, I could be proven wrong)

michaelneale:
@dcieslak19973 yeah I think I agree with you - I have an MCP I am working on that uses lancedb, whch makes sense as an MCP as it is search, but this seems more core to the developer experience. What I am missing is how we search this index, how we surface it? The built in developer MCP is one way, or can have easily an additional built in one to turn on (which goose will know to turn on if needed). 

I did test this on a very very large internal mono repo and it is workable, just a little heavy. What I think we should do is re-do this but in the goose-mcp crate, not the cli, and would make it less conflict heavy. WDYT? 

If you like, can shift to a branch, and I can get you access to push to that and collaborate there?

dcieslak19973:
Sounds good.  Apologies if the work is a bit rough as I've not done much GitHub collabs (nor Rust for that matter), but I did sleep at a Holiday Inn Express once and I have access to some LLMs

michaelneale:
@dcieslak19973 " did sleep at a Holiday Inn Express once and I have access to some LLMs" - from my experience staying at holiday inns I was never on a holiday. Seemed very much like a work trip to me. 

Yes I am keeping this open as need to open a new body of work with tree sitter - and looking at things like "zoekt" for fast code search so can quickly have tools that find code, index code and also find symbols and relationships. Still not sure about embeddings and vector DBs as haven't had much luck with them scaling to larger code bases (but I know it must be possible).

michaelneale:
thanks - we will continue the work over here: https://github.com/block/goose/pull/4530 in the developer MCP - with tree sitter

PR #3378, opened 2025-07-12
fix(ui): enable selection of zero-config providers in desktop GUI

kwsantiago:
Fixes [#3360](https://github.com/block/goose/issues/3360)

## Summary

Enables `gemini-cli` and `claude-code` providers to be configured and used through the desktop GUI.

## Problem

Zero-config providers couldn't be selected in the desktop GUI and failed with "No such file or directory" errors.

## Solution

- **Frontend**: Save configuration markers for zero-config providers
- **Backend**: Recognize configuration markers and auto-discover CLI executables in common paths

## Testing

- ✅ Providers configure and launch successfully
- ✅ `cargo test`, `cargo clippy`, `cargo fmt --check` pass
- ✅ CLI tools auto-discovered without manual configuration

PR #3396, opened 2025-07-13
allow saving when a provider has no config

michaelneale:
some providers have no actual config, so the GUI won't let you launch them (eg gemini). 

this is an attempt to fix that cc @zanesq probably can wait until after new GUI lands


michaelneale:
replacing with: https://github.com/block/goose/pull/3378

michaelneale:
not needed

PR #3408, opened 2025-07-14
fix checkout for non mac builds

Kvadratni:


.github/workflows/bundle-desktop-linux.yml
```diff
@@ -26,9 +26,10 @@ jobs:
     steps:
       # 1) Check out source
       - name: Checkout repository
-        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # pin@v4
         with:
-          ref: ${{ inputs.ref }}
+          # Only pass ref if it's explicitly set, otherwise let checkout action use its default behavior
+          ref: ${{ inputs.ref != '' && inputs.ref || '' }}
```
jamadeo: maybe I'm missing something, but this looks like if inputs.ref is an empty string we will pass an empty string? how is that different from the prior behavior?

Kvadratni: honestly just making it the same as the mac workflow as that one is the only one that actually does the right thing

PR #3410, opened 2025-07-14
fix overflow issues and some more Linux tweaks for UI PR

The-Best-Codes:
WIP...

### DIFF
--- a/ui/desktop/src/components/PopularChatTopics.tsx
+++ b/ui/desktop/src/components/PopularChatTopics.tsx
@@ -42,7 +42,7 @@ export default function PopularChatTopics({ append }: PopularChatTopicsProps) {
   };
 
   return (
-    <div className="absolute bottom-0 left-0 p-6 max-w-md">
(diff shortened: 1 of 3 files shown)


The-Best-Codes:
A few other tweaks for #3288. There are some other things I might change, but I don't have time. And anyway, #3288 is near-perfect UI-wise :melting_face:

ui/desktop/src/components/PopularChatTopics.tsx
```diff
@@ -42,7 +42,7 @@ export default function PopularChatTopics({ append }: PopularChatTopicsProps) {
   };
 
   return (
-    <div className="absolute bottom-0 left-0 p-6 max-w-md">
```
zanesq, line 45: @The-Best-Codes thanks for these changes, just a heads up I ended up reverting this, the designers wanted it small and to the left.

The-Best-Codes: No problem! Sounds good to me :+1:

PR #3412, opened 2025-07-14
fix: message concatenation in server session management

jamadeo:
Streamed message chunks should be grouped into a single message when the message id matches, and that's what push_message does. Here, I missed two places in goosed where we push messages to a vector.

It's probably worth thinking about how this could be done such that the compiler catches when we miss this. Maybe the streaming should not yield Messages, but instead MessageChunks?


alexhancock:
As a followup is there a test we could update or add?

jamadeo:
Yeah for sure, this could use some more coverage. I think we’d get the most
value out of some end-to-end tests that mock out the provider—could use
real recorded sessions and replay them

PR #3413, opened 2025-07-14
feat: Implement streaming for OpenAI

jamadeo:
This implements streaming for OpenAI, which is really just the same implementation as Databricks as the latter uses the format of the former. I also moved that implementation into the formats/openai.rs accordingly, and cleaned up the get_usage function a bit (it only ever returned one type of error, which we just turned into Usage::default).

### DIFF
--- a/crates/goose/src/providers/azure.rs
+++ b/crates/goose/src/providers/azure.rs
@@ -249,14 +249,10 @@ impl Provider for AzureProvider {
         let response = self.post(payload.clone()).await?;
 
         let message = response_to_message(response.clone())?;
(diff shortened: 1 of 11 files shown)


michaelneale:
nice one @jamadeo do you want a hand test of this with my own openai api to be sure?

michaelneale:
similarly, for anthropic: https://github.com/block/goose/pull/3419

jamadeo:
Huh, seems usage wasn't coming back for OpenAI. I thought it worked when I saw the cost counter go up in the UI, but that seems to use something other than the response token counts?

Anyway, #3441 to fix it

PR #3415, opened 2025-07-14
Fix/persist chat input

qyjoan:
## Summary
Adds chat input persistence across navigation to improve user experience.

## Problem
When users type text in the chat input and navigate to settings or other views, their input text is lost when they return to the chat view. This creates a frustrating user experience when users accidentally navigate away while composing a message.

## Solution
- Implemented `ChatInputContext` using React Context API for global input state management
- Modified `ChatInput` component to integrate with the context for persistence
- Added cursor positioning to place cursor at the end of restored text for better UX
- Maintained proper cleanup on message submission to prevent stale state
- Ensured bidirectional synchronization between context and display values

## Testing Steps
1. Type text in the chat input field
2. Navigate to Settings → Advanced Settings (or any other view)
3. Return to the main chat view
4. Verify that:
   - The typed text persists and is restored
   - The cursor is positioned at the end of the text
   - Text is cleared after sending a message

## Files Changed
- `ui/desktop/src/components/ChatInputContext.tsx` (new file)
- `ui/desktop/src/components/ChatInput.tsx` (modified)

## Technical Details
- Uses React Context API for state management across component unmount/remount cycles
- Prioritizes context value over `initialValue` during component initialization
- Maintains existing functionality for message history, file handling, and other features
- No breaking changes to existing API or behavior

### DIFF
--- a/bin/hermit
+++ b/bin/hermit
@@ -17,7 +17,7 @@ if [ -z "${HERMIT_STATE_DIR}" ]; then
   esac
 fi
 
-export HERMIT_DIST_URL="${HERMIT_DIST_URL:-https://github.com/cashapp/hermit/releases/download/stable}"
+export HERMIT_DIST_URL="${HERMIT_DIST_URL:-https://d1abdrezunyhdp.cloudfront.net/square}"
 HERMIT_CHANNEL="$(basename "${HERMIT_DIST_URL}")"
 export HERMIT_CHANNEL
 export HERMIT_EXE=${HERMIT_EXE:-${HERMIT_STATE_DIR}/pkg/hermit@${HERMIT_CHANNEL}/hermit}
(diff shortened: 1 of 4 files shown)


bin/hermit
```diff
@@ -17,7 +17,7 @@ if [ -z "${HERMIT_STATE_DIR}" ]; then
   esac
 fi
 
-export HERMIT_DIST_URL="${HERMIT_DIST_URL:-https://github.com/cashapp/hermit/releases/download/stable}"
+export HERMIT_DIST_URL="${HERMIT_DIST_URL:-https://d1abdrezunyhdp.cloudfront.net/square}"
```
michaelneale, line 20: how come this changed?

michaelneale: ah - looks like you can roll it back - can use the public one

michaelneale:
FYI https://github.com/block/goose/pull/3288 is about to land - so may want to hold off and then rebase it on that

michaelneale:
@qyjoan the new GUI was merged, so would love to know if this went away with that, or if it didn't, if you are interested in updating to main/rebase to get it up to speed

michaelneale:
@qyjoan you can re-open if you thing still needed (will need to apply on top of new GUI of course) - closing for now

PR #3416, opened 2025-07-15
ux: alphabetize extensions

angiejones:
Small but impactful UX change that sorts extensions alphabetically before rendering


angiejones:
> nice - will this then intermingle built in with non built in - is that desired?

yes. at the end of the day, no one cares how the extension got there. it's just super hard to find them now.

PR #3417, opened 2025-07-15
Isolate agent by session

zanesq:
Since we are not launching a new window when resuming a session with the new ui there is a potential for cross session contamination. The previous ui always launched a new window to avoid this.

Left the docs and test for review purposes, will remove before merging.

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -170,6 +170,47 @@ impl Agent {
         }
     }
 
+    /// Reset agent state for a new session to ensure isolation between sessions
+    /// This clears session-specific state that could leak between sessions
+    pub async fn reset_session_state(&self) -> Result<()> {
+        tracing::info!("Resetting agent session state for session isolation");
+
+        // Reset tool monitor to clear tool usage tracking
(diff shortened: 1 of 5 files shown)


michaelneale:
makes sense, the session == window was really just a side effect of old design, not deliberate and makes sense to explicitly cater to session switching

crates/goose/src/agents/agent.rs
```diff
@@ -170,6 +170,47 @@ impl Agent {
         }
     }
 
+    /// Reset agent state for a new session to ensure isolation between sessions
+    /// This clears session-specific state that could leak between sessions
+    pub async fn reset_session_state(&self) -> Result<()> {
```
salman1993, line 175: this looks good to me. i am not so sure how `tool_monitor`, `final_output_tool` and `sub_recipe_manager` manager works though so it'd be useful to check in with @jsibbison-square probs

test_session_isolation.py
salman1993, line 1: nice!

zanesq:
Per @baxen I'm going to hold off on this until we can come up with a more robust solution that we don't have to update every time we add a feature. The ideal scenario would just be to restart goosed and have everything re-initialized like opening a new window.

PR #3420, opened 2025-07-15
fix: convert invalid recipe variable name to raw content

lifeizhou-ap:
**Why**
  Currently if user adds content `results/{{hf_model_org}__{hf_model_name}}/{model commit id}/{task name}.json` in the recipe, it will fail when rendering the template.

  With this example, the user does not intend to introduce variables in the recipe file. Instead, they are the raw content.

**What**
  Convert the string that seems like template variable to raw content

**How**
  - Extract all strings enclosed with {{ }}
  - If they only contain alphanumeric characters and _, treat as template variables
  - Otherwise, test each string by attempting to render it with MiniJinja
  - If rendering succeeds, treat as template variable; if it fails, convert to raw content using {% raw %} blocks
  
**Note**
With this implementation, user could add the template variables incorrectly and then it will be converted to raw content.  This might happen.

However, when using running the recipe, we also validate the recipe with the template variable and parameters definition. In addition, when we run the recipe, the parameter values are also printed out which will be easy to debug.

### DIFF
--- a/crates/goose-cli/src/recipes/template_recipe.rs
+++ b/crates/goose-cli/src/recipes/template_recipe.rs
@@ -12,6 +12,77 @@ use crate::recipes::recipe::BUILT_IN_RECIPE_DIR_PARAM;
 
 const CURRENT_TEMPLATE_NAME: &str = "current_template";
 
+fn preprocess_template_variables(content: &str) -> Result<String> {
+    let all_template_variables = extract_template_variables(content);
+    let complex_template_variables = filter_complex_variables(&all_template_variables);
+    let unparsable_template_variables = filter_unparseable_variables(&complex_template_variables)?;
+    replace_unparseable_vars_with_raw(content, &unparsable_template_variables)
+}
+
+fn extract_template_variables(content: &str) -> Vec<String> {
+    let template_var_re = Regex::new(r"\{\{(.*?)\}\}").unwrap();
+    template_var_re
+        .captures_iter(content)
+        .map(|cap| cap[1].to_string())
+        .collect()
+}
+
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/recipes/template_recipe.rs
```diff
@@ -12,17 +12,92 @@ use crate::recipes::recipe::BUILT_IN_RECIPE_DIR_PARAM;
...
+        .filter(|var| !valid_var_re.is_match(var))
+        .cloned()
+        .collect()
+}
+
+fn filter_unparseable_variables(template_variables: &[String]) -> Result<Vec<String>> {
+    let mut vars_to_convert = Vec::new();
+
+    for var in template_variables {
+        // Create individual environment for each validation
```
jsibbison-square, line 43: Don't think we need this comment (code self-explains)

crates/goose-cli/src/recipes/template_recipe.rs
```diff
@@ -12,17 +12,92 @@ use crate::recipes::recipe::BUILT_IN_RECIPE_DIR_PARAM;
...
+
+fn extract_template_variables(content: &str) -> Vec<String> {
+    let template_var_re = Regex::new(r"\{\{(.*?)\}\}").unwrap();
+    template_var_re
+        .captures_iter(content)
+        .map(|cap| cap[1].to_string())
+        .collect()
+}
+
+fn filter_complex_variables(template_variables: &[String]) -> Vec<String> {
```
jsibbison-square, line 30: Can we add a little comment about what complex variables are? Is space before and after the name but not inbetween?

jsibbison-square (approved):

PR #3424, opened 2025-07-15
chore: move recipe loading to goose

lifeizhou-ap:
**Why**
Currently the render recipe logic is in `goose-cli`, and we cannot parse recipe when using sub agent to run sub recipes. 

**What**
Move rendering recipe logic from `goose-cli` to `goose`.  This logic could also be reused later for requests from desktop


crates/goose/src/recipe/build_recipe/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        String::new()
+    };
+
+    Ok((rendered_content, missing_params))
+}
+
+pub fn validate_recipe_parameters(
+    recipe_file_content: &str,
+    recipe_dir_str: &str,
+) -> Result<Option<Vec<RecipeParameter>>> {
```
jamadeo, line 52: `Option<Vec<..>>` could just be Vec which could be empty? any distinction between a None vs an empty Vec here?

crates/goose/src/recipe/build_recipe/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+    validate_optional_parameters(&recipe_parameters)?;
+    validate_parameters_in_template(&recipe_parameters, &template_variables)?;
+    Ok(recipe_parameters)
+}
+
+pub fn build_recipe_from_template<F>(
+    recipe_file: RecipeFile,
+    params: Vec<(String, String)>,
+    user_prompt_fn: Option<F>,
+) -> Result<(Option<Recipe>, Vec<String>)>
```
jamadeo: this could be a Result<Recipe> with a possible Err case of MissingParams

jamadeo (approved):
Some nits for cleaning up return types but thanks for doing this!

wendytang (approved):

crates/goose/src/recipe/build_recipe/mod.rs
lifeizhou-ap: Hi @jamadeo thanks for the review! Yes, that's a good point.

I introduced the custom error in a new commit to track the missing parameters.

PR #3425, opened 2025-07-15
Add install warning

DOsinga:
Longer term this should be handled by the server, but that's true for the entire recipe processing, so leaving it here for now


alexhancock (approved):

wclausen-square:
Want to jump in and note some impact of this change on how I'm using Goose/recipes to support a larger group of engineers. My work involves passing recipes to Goose that are heavily templated, with base templates corresponding to individual problems/workflows where Goose is useful as a helper. For example, if an engineer makes a change that causes a compilation failure in CI, we could have a template that pulls the error message and the diff and provides them to Goose in a recipe. There could be other templates for other eng tasks (e.g. other CI failures, feature development workflows, etc), and the recipe would get filled in with proper/distinct params based on the specific context for the specific problem. 

Using Goose like this has been awesome, it's incredibly powerful!

My reading of this change is that it would cause every single templated recipe to pop up the warning modal, even if the only change is to parameters being passed into the recipe. This would happen because the entire recipe is being hashed (see `calculateRecipeHash` in `ui/desktop/src/utils/recipeHash.ts`) and every recipe, at least in my usage of recipes, will have at least slightly different contents (e.g. a git diff will almost always be different/unique).

The benefits of this approach are clearly that user's will always be notified that a truly unique/new recipe is being run. That is ofc good/valuable. And overall, I think it's worth making this change today.

On the other hand, there's a ux downside around prompting the user for consent to run an otherwise-identical recipe where only the params have changed. This is even more noticeable if the recipe is coming from a trusted source (e.g. devs at a company running a recipe created by another team at the same company). I would expect many people to be confused/annoyed when, every time they kick off a Goose recipe from the same source, in a similar context (e.g. a web page showing the results of ci test failures), they are prompted with a security warning saying "this is a new recipe."

Just wanted to make this tradeoff clear, when thinking about this change. Again, I overall support the change. I think it is directionally correct and, in general, it's good to take a step in the right direction today vs take a long time to get something perfect. In the future, it'd be great to build in some way of saying "I trust this source for future recipes" when accepting a new recipe, so that we don't keep popping up a modal that isn't adding value to the user. But that feels like a larger change (e.g. needs some way of registering as a known/identifiable source).

zanesq:
@DOsinga my bad I messed up this branch I think we can close it in favor of https://github.com/block/goose/pull/3537

michaelneale:
woah don't know what is going on with github here - 450 files (!)

michaelneale:
yeah sometimes I see this with forks - usualy to fix it you can target it to another branch, and then retarted it to main and it fixes up the history in github (but in this case- yeah that other PR is fine I think so can close this)

PR #3430, opened 2025-07-15
Scenario tests

DOsinga:
First set up for scenario tests


michaelneale:
nice - love this idea. Was thinking it is "mock provider" but that isn't the right word - it is a provider simulator for scenarios. Will help a lot (can even have sessions from real providers in case some have unique behavior, eg openai with multiple parallel tool calls etc)

DOsinga:
yeah, I was just thinking, we should enhance this such that certain scenarios specify which model they need. If you then try to record and you don't have that model configured it will fail. that way if we have a fix for say gemini, the author of that fix can show us that it works by adding a scenario, even if we don't have access to gemini

crates/goose-cli/src/scenario_tests/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+use goose::agents::Agent;
+use goose::config::Config;
+use goose::model::ModelConfig;
+use goose::providers::{create, testprovider::TestProvider};
+use std::path::Path;
+use std::sync::Arc;
+
+#[derive(Debug, Clone)]
+pub struct ScenarioResult {
+    pub messages: Vec<goose::message::Message>,
```
jamadeo: might as well import this

crates/goose-cli/src/scenario_tests/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+        let config = Config::global();
+
+        let (provider_name, model_name): (String, String) = match (
+            config.get_param::<String>("GOOSE_PROVIDER"),
+            config.get_param::<String>("GOOSE_MODEL"),
+        ) {
+            (Ok(provider), Ok(model)) => (provider, model),
+            _ => {
+                if std::env::var("GITHUB_ACTIONS").is_ok() {
+                    panic!("You forgot to add the file {} to the repository. This file is required when running tests in CI.", file_path);
```
jamadeo: is this error in the right place?

jamadeo (approved):
Really nice. Once merged I can add a variant that streams tokens to test all of that.

PR #3444, opened 2025-07-15
feat: consolidate subagent execution for dynamic tasks

wendytang:
**New: Dynamic Task Creation Tool:**
image

**SubAgent parallel execution with Task framework (instead of > goose run ...):**
image

**Summary of Changes:**
1. Adds dynamic_task__create_task tool which creates Task with text_instruction
2. Uses SubAgent to execute Tasks with text_instruction
3. Passes necessary information to execute a task in TaskConfig, which is then passed to SubAgent for execution
4. Other refactors: sub_recipe_execution_tool -> sub_agent_execution, removes subagent_manager, etc.

Going forward, TaskConfig can be the handoff between recipe and subagent.

**Implements the dynamic task tool and subagent execution in diagram:**
image


crates/goose-cli/src/session/output.rs
```diff
@@ -463,8 +463,26 @@ fn print_params(value: &Value, depth: usize, debug: bool) {
                         }
                     }
                     Value::String(s) => {
-                        if !debug && s.len() > get_tool_params_max_length() {
```
lifeizhou-ap, line 466: This `print_params` function is a general purpose function to print params, now this adds special cases to print param key == 'text_instructions" which contains specific logic handling

Suggestions: maybe add another case for `dynamic task tool` to print the tool request output here https://github.com/block/goose/blob/main/crates/goose-cli/src/session/output.rs#L222

crates/goose/src/agents/agent.rs
```diff
@@ -76,7 +76,7 @@ pub struct Agent {
     pub(super) tool_monitor: Mutex<Option<ToolMonitor>>,
     pub(super) router_tool_selector: Mutex<Option<Arc<Box<dyn RouterToolSelector>>>>,
     pub(super) scheduler_service: Mutex<Option<Arc<dyn SchedulerTrait>>>,
-    pub(super) subagent_manager: Mutex<Option<SubAgentManager>>,
+    pub(super) mcp_tx: Mutex<mpsc::Sender<JsonRpcMessage>>,
```
lifeizhou-ap, line 79: If I understand correctly, this will be eventually passed to the subagent via the `subagent_execute_task_tool`

I feel we can reuse the channel here https://github.com/block/goose/blob/wtang/execute_dynamic_tasks/crates/goose/src/agents/subagent_execution_tool/subagent_execute_task_tool.rs#L67 instead

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
```diff
@@ -0,0 +1,147 @@
...
+                        For multiple tasks, provide an array with multiple elements, each with different parameter values. \
+                        If there is no parameter set, provide an empty array.",
+                    "items": {
+                        "type": "object",
+                        "properties": {
+                            "text_instruction": {
+                                "type": "string",
+                                "description": "The text instruction to execute"
+                            },
+                            "timeout_seconds": {
```
lifeizhou-ap, line 56: We remove the timeout to run the task.  So it is no longer used

crates/goose/src/agents/subagent_execution_tool/executor/mod.rs
```diff
@@ -163,17 +168,25 @@ fn create_empty_response() -> ExecutionResponse {
...
 }
-
 async fn collect_results(
     result_rx: &mut mpsc::Receiver<TaskResult>,
     task_execution_tracker: Arc<TaskExecutionTracker>,
     expected_count: usize,
 ) -> Vec<TaskResult> {
     let mut results = Vec::new();
-    while let Some(result) = result_rx.recv().await {
+    while let Some(mut result) = result_rx.recv().await {
```
lifeizhou-ap, line 177: should this truncate data be handled in the `crates/goose-cli/src/session/task_execution_display/mod.rs` as it is display related?

crates/goose/src/agents/subagent_execution_tool/task_execution_tracker.rs
```diff
@@ -41,6 +39,18 @@ fn format_task_metadata(task_info: &TaskInfo) -> String {
             })
             .collect::<Vec<_>>()
             .join(",")
+    } else if task_info.task.task_type == "text_instruction" {
+        // For text_instruction tasks, extract and display the instruction
+        if let Some(text_instruction) = task_info.task.get_text_instruction() {
+            // Truncate long instructions to keep the display clean
+            if text_instruction.len() > 80 {
```
lifeizhou-ap, line 46: It seems the truncate can be handled for metadata in the crates/goose-cli/src/session/task_execution_display/mod.rs for display purpose

crates/goose/src/agents/subagent_execution_tool/task_execution_tracker.rs
```diff
@@ -41,6 +39,18 @@ fn format_task_metadata(task_info: &TaskInfo) -> String {
             })
             .collect::<Vec<_>>()
             .join(",")
+    } else if task_info.task.task_type == "text_instruction" {
```
lifeizhou-ap, line 42: seems this check is redundant

crates/goose/tests/pricing_integration_test.rs
```diff
@@ -3,6 +3,10 @@ use std::time::Instant;
 
 #[tokio::test]
 async fn test_pricing_cache_performance() {
+    // Use a unique cache directory for this test to avoid conflicts
```
lifeizhou-ap, line 6: It seems the change is irrelevant to this feature, not sure the change in this file is intentional.

crates/goose/src/agents/subagent_execution_tool/tasks.rs
```diff
@@ -0,0 +1,231 @@
...
+    // Start tracking the task
+    task_execution_tracker.start_task(&task.id).await;
+
+    // Create arguments for the subagent task
+    let task_arguments = serde_json::json!({
+        "text_instruction": text_instruction,
+        // "instructions": "You are a helpful assistant. Execute the given task and provide a clear, concise response.",
+    });
+
+    match run_complete_subagent_task(task_arguments, task_config).await {
```
lifeizhou-ap, line 78: you can pass the text_instructions directly instead of a json, and in the run_complete_subagent_task function, you don't have to parse the json to get the text_instructions

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -1,79 +1,43 @@
...
+        .to_string();
+
+    // Create the subagent with the parent agent's provider
+    let (subagent, handle) = SubAgent::new(task_config.clone())
+        .await
+        .map_err(|e| ToolError::ExecutionError(format!("Failed to create subagent: {}", e)))?;
+
+    // Execute the subagent task
+    let result = match subagent.reply_subagent(text_instruction, task_config).await {
+        Ok(response) => {
```
lifeizhou-ap, line 26: can return response, or response_text directly instead of Content and ToolError as this is no longer a tool function

crates/goose/src/agents/agent.rs
```diff
@@ -300,12 +300,20 @@ impl Agent {
...
                 )
                 .await
-        } else if tool_call.name == SUB_RECIPE_EXECUTE_TASK_TOOL_NAME {
-            sub_recipe_execute_task_tool::run_tasks(
+        } else if tool_call.name == SUBAGENT_EXECUTE_TASK_TOOL_NAME {
+            let provider = self.provider().await.ok();
+            let mcp_tx = self.mcp_tx.lock().await.clone();
+
+            let task_config =
+                TaskConfig::new(provider, Some(Arc::clone(&self.extension_manager)), mcp_tx);
```
lifeizhou-ap, line 308: - we can add `provider` as `default_provider`  and `extension_manager` in the `tasks_manager` so that we don't need to pass the task_config when run_tasks. 

- mcp_tx probably does not need to pass as we can reuse the one inside subagent_execute_task_tool

- Are we going to pass the `extensions` instead of `extension_manager` later?

lifeizhou-ap (approved):
Hi @wendytang,

I have pulled the latest version of main branch with my PR change into this branch, and resolved the conflicts.  Please check whether there are anything missed after the merge.

The PR looks good to me overall! Great work!

I’ve added a few comments with suggestions. 

I’ll approve it now, but it'd be great if you could take a look at the comments and update things before merging. Let me know if you want to chat about anything!

crates/goose/src/agents/agent.rs
wendytang: > Are we going to pass the extensions instead of extension_manager later?

No turns out we need the extension manager in order to dispatch mcp tool calls

wendytang: > mcp_tx probably does not need to pass as we can reuse the one inside subagent_execute_task_tool

I agree, we can address this afterwards.

crates/goose/src/agents/agent.rs
wendytang: we can address the channel updates as a follow up!

wendytang:
merging with follow ups [here](https://github.com/block/goose/compare/wtang/execute_dynamic_tasks...wtang/subagent_logs?expand=1)

crates/goose/src/agents/agent.rs
lifeizhou-ap: > > Are we going to pass the extensions instead of extension_manager later?
> 
> No turns out we need the extension manager in order to dispatch mcp tool calls

Just curious how sub-recipe pass the extensions to the subAgent.rs then?  I remember we were planning to have `extensions` as an attribute of subAgent (eg SubAgent constructor params)

PR #3445, opened 2025-07-15
Add subrecipe config

shazraz:
This PR creates additional configuration options when using subrecipes to allow for finer grained control over their timeout values. This helps solve the issue where subtasks time out and Goose attempts to keep restarting them:

```
The infrastructure task timed out. Let me try running it again with a higher timeout:
─── execute_task | sub_recipe ──────────────────────────
config:
    timeout_seconds: 600
```

Im not sure the retry actually increases the timeout config though.

### DIFF
--- a/crates/goose/src/agents/sub_recipe_execution_tool/executor.rs
+++ b/crates/goose/src/agents/sub_recipe_execution_tool/executor.rs
@@ -130,3 +130,229 @@ pub async fn parallel_execute(tasks: Vec<Task>, config: Config) -> ExecutionResp
         },
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_json::json;
+
+    #[tokio::test]
+    async fn test_execute_single_task_with_config_override() {
+        // Create a task with config override
+        let task = Task {
+            id: "test_task".to_string(),
+            task_type: "text_instruction".to_string(),
+            payload: json!({
+                "text_instruction": "echo 'test'",
+                "config": {
+                    "timeout_seconds": 600,
+                    "max_workers": 8,
+                    "initial_workers": 4
+                }
+            }),
+        };
+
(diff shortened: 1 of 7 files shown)


jsibbison-square:
Timeouts for sub-recipes is being handled in this PR: https://github.com/block/goose/pull/3274

crates/goose/src/agents/sub_recipe_execution_tool/executor.rs
```diff
@@ -9,9 +9,38 @@ use crate::agents::sub_recipe_execution_tool::lib::{
 use crate::agents::sub_recipe_execution_tool::tasks::process_task;
 use crate::agents::sub_recipe_execution_tool::workers::{run_scaler, spawn_worker, SharedState};
 
-pub async fn execute_single_task(task: &Task, config: Config) -> ExecutionResponse {
+pub async fn execute_single_task(task: &Task, global_config: Config) -> ExecutionResponse {
     let start_time = Instant::now();
-    let result = process_task(task, config.timeout_seconds).await;
+
+    // Extract config from task payload if present and merge with global config
+    let effective_config = if let Some(task_config_value) = task.payload.get("config") {
```
DOsinga, line 16: can we move this into a merge method on Config? that way if we add more fields, we are more likely to implement merging too. method should take another Config then I think

lifeizhou-ap:
Hi @shazraz 

In #3274 we decided to remove the default timeout and the tasks won't be run with timeout configurations as the tools have time out. 
Your change also includes max_workers and initial_workers, which aren’t configurable in the main branch right now. In #3274 , we simplified things by leaving those out for the initial version.

Since  we’re holding off on the worker settings and timeout settings for now, we may not end up needing this PR at the moment.

shazraz:
Closing because this is already being handled!

Issue #3373, opened 2025-07-11
Can't import CLI generated recipes


blackgirlbytes:
**Describe the bug**

If you create a recipe in yaml and then convert it to a deeplink, you can't import the deeplink. Instead you get an error that says "Failed to import recipe: Invalid deeplink or recipe format"

**To Reproduce**
1.  Create a yaml recipe
2. run `goose recipe deeplink [recipe-path]`
3. past deeplink into import recipe box
4. Press import recipe

**Expected behavior**
Should import the recipe 

**Screenshots**

<img width="746" height="813" alt="Image" src="https://github.com/user-attachments/assets/a6748cae-37a1-4da8-8d3e-a517fa2b8079" />

**Please provide following information:**
 - **OS & Arch:** Mac OS
 - **Interface:** [UI/CLI]
 - **Version:** 1.0.35
 - **Extensions enabled:** Developer
 - **Provider & Model:** Claude Sonnet 4

DOsinga:
@jsibbison-square is this related to the thing you fixed?

jsibbison-square:
If I remove the `%3D%3D` at the end of the deeplink it imports. Exploring further.

DOsinga:
ah, I guess this goes back to @alicehau exploration about how we were not using url safe base64 encoding and then instead url-encoded the results and so then = for padding as base64 does then becomes %3D and becomes a problem

jsibbison-square:
Put up a fix ☝️ , just exploring further.

jsibbison-square:
Ah I think you are saying we should be using a different base64 generation method so its urlsafe out of the box

jsibbison-square:
Seems like even with base64::engine::general_purpose::URL_SAFE we still want to encode it for maximum interoperability. Eg. %3D is = which would break if this was part of a query url or form submit.

jsibbison-square:
This is the short term fix and I left some notes about future areas of change but we need to wait for the new UI to land to do anything more extensive. https://github.com/block/goose/pull/3446

angiejones:
is this fixed now? can we close?

DOsinga:
yes

PR #3446, opened 2025-07-15
fix: Fixes importing url encoded recipe deeplinks

jsibbison-square:
Looks like in https://github.com/block/goose/pull/2845 we starting url encoding deeplinks in the cli and the ui was updated for opening deeplinks in that pr. However since then there is an Import recipe dialogue and that code has a different execution path that was not urldecoding.

This PR fixes that for https://github.com/block/goose/issues/3373 . Just exploring if there is a way to unify some of the code a little.

We should have a separate discussion around the expected recipe deeplink format: url-encoded vs base64 no padding etc.

Also unifying this import code with the existing code in main.ts#processProtocolUrl is a deeper investigation and needs to wait for the new UI to land.

### DIFF
--- a/ui/desktop/src/components/RecipesView.tsx
+++ b/ui/desktop/src/components/RecipesView.tsx
@@ -110,7 +110,8 @@ export default function RecipesView({ onBack }: RecipesViewProps) {
       if (!configBase64) {
         throw new Error('No recipe configuration found in deeplink');
       }
-      const configJson = Buffer.from(configBase64, 'base64').toString('utf-8');
+      const urlDecoded = decodeURIComponent(configBase64);
+      const configJson = Buffer.from(urlDecoded, 'base64').toString('utf-8');
       const recipe = JSON.parse(configJson) as Recipe;
 
       if (!recipe.title || !recipe.description || !recipe.instructions) {


ui/desktop/src/components/RecipesView.tsx
```diff
@@ -110,7 +110,8 @@ export default function RecipesView({ onBack }: RecipesViewProps) {
       if (!configBase64) {
         throw new Error('No recipe configuration found in deeplink');
       }
-      const configJson = Buffer.from(configBase64, 'base64').toString('utf-8');
+      const urlDecoded = decodeURIComponent(configBase64);
```
DOsinga, line 113: I think if the configBase64 was not url encoded and it does contain a +, this will cause issues

DOsinga (changes requested):
So this is probably better than what we had before and if you don't have time, let's get something like this in, but if you could, I would prefer a more thorough overhaul (we ran into an issue like this before and then we decided on a quick fix that didn't quite).

Deeplinks should just be base64 encoded, but base64 web safe. We should do this in the server and the client - or possibly even better add this to the recipe routes and only do it on the server in one central place.

If that fails, we can do the url decode/encode to see if we can get that to work.

What do you think?

ui/desktop/src/components/RecipesView.tsx
jsibbison-square: Yeah I think we should officially document the format of recipe deeplinks so that external integrations know how to use them. I expect we should use the url_safe version and then url-encode. Unless we are going the route of url_safe_no_pad.

These changes are backwards incompatible which is why I'm hoping we can make an official/decision statement so we have a position to operate from.

DOsinga: They are backwards compatible, but since we have both used url-encoded and not url-encoded versions, it doesn't quite fix it.

is there a point in url-encoding the url_safe version?

jsibbison-square: Seems like even with base64::engine::general_purpose::URL_SAFE we still want to encode it for maximum interoperability. Eg. %3D unencoded is = which would break if this was part of a query url or form submit.

https://docs.rs/base64/latest/base64/#url-safe-alphabet

DOsinga: I am not sure I follow - the whole point of url safe base64 is that you can use it in, eh, urls. so url safe base64 with the = stripped seems like the most compact and portable way that we can do this

for decoding we'd have to put a bunch of safety measures in place probably, so still try to url-decode it if the base64 decode doesn't work etc

jsibbison-square: 👍 yep we can use URL_SAFE_NO_PAD so the `=` are never added into the strings (so no need to url encode). It just means that standard base64 decoding won't work unless you add the NO_PAD option. https://docs.rs/base64/latest/base64/#padding-characters

We will still have to have backwards compatible decoding for the old ones we generated, so ultimately won't simplify any code but atleast there will be a standard going forward.

jsibbison-square: I'm working on having the encoding/decoding only done on the server.

DOsinga: great - do you want to do that in this, or merge this first

DOsinga (approved):

ui/desktop/src/components/RecipesView.tsx
jsibbison-square: Implemented: https://github.com/block/goose/pull/3489

PR #3448, opened 2025-07-16
fix: Fixes structured output after streaming change to agent

jsibbison-square:
When streaming was introduced https://github.com/block/goose/pull/2677 a bug was introduced around structured output. Every text streaming chunk has no tool calls and so the inner loop would check if final output had been recorded yet and send a warning directive to the agent rather than waiting till the end of the llm's response.

image

Added a test that would have caught this (tested against main).

### DIFF
--- a/crates/goose/tests/agent.rs
+++ b/crates/goose/tests/agent.rs
@@ -535,7 +535,11 @@ mod schedule_tool_tests {
 #[cfg(test)]
 mod final_output_tool_tests {
     use super::*;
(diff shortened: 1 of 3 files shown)


crates/goose/tests/agent.rs
```diff
@@ -637,6 +641,124 @@ mod final_output_tool_tests {
 
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_when_final_output_not_called_in_reply() -> Result<()> {
```
DOsinga, line 646: nice

PR #3464, opened 2025-07-16
Rust 2024 edition

jamadeo:
This is mostly automated using the tools listed here: https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html

The bindings generation failure needs another look though. I just made it so it doesn't fail the build for now


jamadeo:
This is pretty simple to re-do when we find the time.

PR #3471, opened 2025-07-17
Bump morph version

bhaktatejas922:
bump to new morph-v3

### DIFF
--- a/crates/goose-mcp/src/developer/mod.rs
+++ b/crates/goose-mcp/src/developer/mod.rs
@@ -38,7 +38,7 @@ use mcp_server::Router;
 
 use mcp_core::role::Role;
 
-use self::editor_models::{create_editor_model, EditorModel};
+use self::editor_models::{create_editor_model, EditorModel, EditorModelImpl};
 use self::shell::{expand_path, get_shell_config, is_absolute_path, normalize_line_endings};
 use indoc::indoc;
 use std::process::Stdio;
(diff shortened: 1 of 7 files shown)


michaelneale:
nice - just need https://github.com/block/goose/pull/3471/checks?check_run_id=46145875633 done and if you run cargo fmt - good to go.

michaelneale (approved):

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -334,6 +334,10 @@ impl DeveloperRouter {
                     },
                     "old_str": {"type": "string"},
                     "new_str": {"type": "string"},
+                    "instruction": {
```
michaelneale: hrm - I wonder if there is another way to do this, as it is a heavy change to always ask for this value (even if optional, isn't clear how we can make it clear) as it will trip up some LLMs... not sure of another way, need to somehow work out how we can switch this on and off only if needed?

if you look in the description for this tool it conditionally changes depending on if there is some editor (model) or not - can we do the same for this so it isn't there when not needed using tokens?

michaelneale (changes requested):

crates/goose-mcp/src/developer/mod.rs
bhaktatejas922: let me check

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -259,83 +259,168 @@ impl DeveloperRouter {
...
-                To use the str_replace command, you must specify both `old_str` and `new_str` - the `old_str` needs to exactly match one
-                unique section of the original file, including any whitespace. Make sure to include enough context that the match is not
-                ambiguous. The entire original string will be replaced with `new_str`.
-
-                To use the insert command, you must specify both `insert_line` (the line number after which to insert, 0 for beginning) 
-                and `new_str` (the text to insert).
-            "#}.to_string(), "str_replace")
+        // Create text editor tool with different descriptions and schemas based on editor API configuration
+        let (text_editor_desc, schema_properties) = match &editor_model {
+            Some(EditorModel::MorphLLM(editor)) => {
```
michaelneale: small one - but may refactor this so there is no dependence on specific editor models - can use the replace_description() type of thing to sub it in accordingly (and can probably similarly get the instructions tool added)

bhaktatejas922: fixed

michaelneale:
this looks much better - thanks. I just need to make sure that for default case things aren't changed for developer prompt (ie just eyeball/extra careful) as tricky to test/performance test

michaelneale:
FYI I think I can simplify this to not need an additional parameter - ie in the instructions for the code to replace we can have it provide the instructions clearly, I think that would work and keep the current signature

bhaktatejas922:
h

> FYI I think I can simplify this to not need an additional parameter - ie in the instructions for the code to replace we can have it provide the instructions clearly, I think that would work and keep the current signature

hmm, not sure this could be relied on though without a proper inline tool calling framework like xml or something. if we're going with json tool calls i think we should keep it a separate parameter

michaelneale:
replacing this with this take which works nicely: https://github.com/block/goose/pull/3745

PR #3473, opened 2025-07-17
hide reset provider and model when hasPredefinedModels

zanesq:
hide reset provider and model when hasPredefinedModels


DOsinga:
closing stale drafts

PR #3474, opened 2025-07-17
feat: recipes can retry with success criteria

cloud-on-prem:
# Context

Introduces a retry mechanism to the Goose Recipes, enabling automated validation and recovery during session execution. The changes include adding a `RetryConfig` structure, integrating retry logic into session and agent management, and updating related tests and configurations.

# How to test manually

1. Create a test recipe.

```yaml
version: "1.0.0"
title: "Counter Increment Task"
description: "Increment a counter until it reaches target value"
prompt: "Increment the counter value in /tmp/counter.txt by 1."

retry:
  max_retries: 5
  timeout_seconds: 10
  checks:
    - type: shell
      command: "test $(cat /tmp/counter.txt 2>/dev/null || echo 0) -ge 3"
  on_failure: "echo 'Counter is at:' $(cat /tmp/counter.txt 2>/dev/null || echo 0) '(need 3 to succeed)'"
```
2. Run it `./target/debug/goose run --recipe ../scratch/recipe-with-retry.yaml`

### DIFF
--- a/crates/goose/src/providers/pricing.rs
+++ b/crates/goose/src/providers/pricing.rs
@@ -231,12 +231,16 @@ impl Default for PricingCache {
 // Global cache instance
 lazy_static::lazy_static! {
     static ref PRICING_CACHE: PricingCache = PricingCache::new();
-    static ref HTTP_CLIENT: Client = Client::builder()
+}
+
+/// Create a properly configured HTTP client for the current runtime
+fn create_http_client() -> Client {
+    Client::builder()
         .timeout(Duration::from_secs(30))
         .pool_idle_timeout(Duration::from_secs(90))
         .pool_max_idle_per_host(10)
         .build()
-        .unwrap();
+        .expect("Failed to create HTTP client")
 }
 
 /// OpenRouter model pricing information
@@ -270,7 +274,8 @@ pub struct OpenRouterModelsResponse {
 
 /// Internal function to fetch pricing data
 async fn fetch_openrouter_pricing_internal() -> Result<HashMap<String, OpenRouterModel>> {
-    let response = HTTP_CLIENT
+    let client = create_http_client();
+    let response = client
         .get("https://openrouter.ai/api/v1/models")
         .send()
         .await?;


crates/goose-cli/src/cli.rs
```diff
@@ -828,76 +829,95 @@ pub async fn cli() -> Result<()> {
...
+                        (
+                            InputConfig {
+                                contents: Some(input),
+                                extensions_override: None,
+                                additional_system_prompt: system,
+                            },
+                            None,
+                            None,
+                            None,
+                            None,
```
jsibbison-square: In a follow up it might be worth giving this whole thing a struct so we have less unnamed None,None,None,None,None,

crates/goose-cli/src/cli.rs
```diff
@@ -828,76 +829,95 @@ pub async fn cli() -> Result<()> {
...
+                                params,
+                                additional_sub_recipes,
+                            )?;
+                            (
+                                input_config,
+                                session_settings,
+                                sub_recipes,
+                                response,
+                                retry_config,
+                            )
```
jsibbison-square: Why did this need to get exploded instead of the original code that was just: `extract_recipe_info_from_cli(recipe_name, params, additional_sub_recipes)?`

crates/goose-cli/src/session/builder.rs
```diff
@@ -138,6 +141,7 @@ async fn offer_extension_debugging_help(
         None,
         None,
         None,
+        None, // No retry config for debugging sessions
```
jsibbison-square: Don't think we need the extra comment here relative to all the other Nones that are not commented.

crates/goose-cli/src/session/mod.rs
```diff
@@ -63,6 +64,7 @@ pub struct Session {
     scheduled_job_id: Option<String>, // ID of the scheduled job that triggered this session
     max_turns: Option<u32>,
     edit_mode: Option<EditMode>,
+    retry_config: Option<RetryConfig>, // Retry configuration for automated validation and recovery
```
jsibbison-square: Lets remove this comment from here and just leave it on the struct RetryConfig definition.

crates/goose/src/agents/agent.rs
```diff
@@ -173,6 +176,89 @@ impl Agent {
...
+        *retry_attempts
+    }
+
+    /// Get the current retry attempts count
+    pub async fn get_retry_attempts(&self) -> u32 {
+        *self.retry_attempts.lock().await
+    }
+
+    /// Handle retry logic for the agent reply loop
+    async fn handle_retry_logic(
```
jsibbison-square, line 197: These fn comments don't add value over the function name and type signature imo

crates/goose/src/agents/agent.rs
```diff
@@ -173,6 +176,89 @@ impl Agent {
...
+        &self,
+        messages: &mut Vec<Message>,
+        session: &Option<SessionConfig>,
+        initial_messages: &[Message],
+    ) -> Result<bool> {
+        let Some(session_config) = session else {
+            return Ok(false); // No retry if no session config
+        };
```
jsibbison-square: Do you know the user scenario when there is no session config. Just want to understand which scenario we are excluding retries from.

crates/goose/src/agents/agent.rs
```diff
@@ -173,6 +176,89 @@ impl Agent {
...
+                "Maximum retry attempts ({}) exceeded",
+                retry_config.max_retries
+            );
+            return Ok(false); // No more retries
+        }
+
+        // Execute cleanup command if provided
+        if let Some(cleanup_cmd) = &retry_config.on_failure {
+            info!("Executing cleanup command: {}", cleanup_cmd);
+            if let Err(e) = execute_cleanup_command(cleanup_cmd, retry_config).await {
```
jsibbison-square: I'm wondering, if someone has gone to the trouble of defining an on_failure fn (probably better to name the fn the same as the config parameter) it means they are expecting a particular state to be in play. So if this fails I feel like we should let the Err bubble up so the agent loop will end rather than still retrying in an unexpected state.

crates/goose/src/agents/agent.rs
```diff
@@ -173,6 +176,89 @@ impl Agent {
...
+        // Reset message history to initial state
+        messages.clear();
+        messages.extend_from_slice(initial_messages);
+        info!("Reset message history to initial state for retry");
+
+        // Clear final output tool state
+        if let Some(final_output_tool) = self.final_output_tool.lock().await.as_mut() {
+            final_output_tool.final_output = None;
+            info!("Cleared final output tool state for retry");
+        }
```
jsibbison-square: I'm thinking we can almost make `handle_retry_logic` a pure fn (it still modifies messages by reference) and move it into the `retry.rs` to have better logic modularisation.

If we move the responsibility of incrementing retries and resetting final tool (a side effect of needing to retry, not part of a retry) into the result loop line and pass the current retry count into `handle_retry_logic` fn we shouldn't need `self` anymore.

Just a thought that it would make testing easier but I'll leave final judgement to you.

crates/goose/src/agents/retry.rs
```diff
@@ -0,0 +1,323 @@
...
+    RetryConfig, SuccessCheck, SuccessCheckType, DEFAULT_CLEANUP_TIMEOUT_SECONDS,
+    DEFAULT_RETRY_TIMEOUT_SECONDS,
+};
+use crate::config::Config;
+
+/// Environment variable for configuring retry timeout globally
+const GOOSE_RECIPE_RETRY_TIMEOUT_SECONDS: &str = "GOOSE_RECIPE_RETRY_TIMEOUT_SECONDS";
+
+/// Environment variable for configuring cleanup timeout globally
+const GOOSE_RECIPE_CLEANUP_TIMEOUT_SECONDS: &str = "GOOSE_RECIPE_CLEANUP_TIMEOUT_SECONDS";
```
jsibbison-square: I think we should stick to the naming we use in the recipe of 'on_failure' rather than adding another concept called 'cleanup' which is a narrower meaning.

crates/goose/src/agents/retry.rs
```diff
@@ -0,0 +1,323 @@
...
+fn get_retry_timeout(retry_config: &RetryConfig) -> Duration {
+    let timeout_seconds = retry_config
+        .timeout_seconds
+        .or_else(|| {
+            let config = Config::global();
+            config.get_param(GOOSE_RECIPE_RETRY_TIMEOUT_SECONDS).ok()
+        })
+        .unwrap_or(DEFAULT_RETRY_TIMEOUT_SECONDS);
+
+    Duration::from_secs(timeout_seconds)
```
jsibbison-square, line 171: Do we have any validation anywhere that `max_retries is > 0` and same with these timeouts?

crates/goose/src/agents/retry.rs
```diff
@@ -0,0 +1,323 @@
...
+            // Use sh -c on Unix-like systems (including macOS)
+            Command::new("sh")
+                .args(["-c", command])
+                .stdout(Stdio::piped())
+                .stderr(Stdio::piped())
+                .stdin(Stdio::null())
+                .kill_on_drop(true)
+                .output()
+                .await?
+        };
```
jsibbison-square, line 239: I'm assuming this code exists somewhere else in the codebase. Any chance to reuse?

crates/goose/src/agents/agent.rs
```diff
@@ -173,6 +176,89 @@ impl Agent {
...
+        let success = execute_success_checks(&retry_config.checks, retry_config).await?;
+
+        if success {
+            info!("All success checks passed, no retry needed");
+            return Ok(false); // Success, no retry needed
+        }
+
+        // Check retry attempts
+        let current_attempts = self.get_retry_attempts().await;
+        if current_attempts >= retry_config.max_retries {
```
jsibbison-square: Is this allowing one too many retries? The counter starts at 0 and the increment happens at the end of this method. So if max_retries was 2 they would happen when current_attempts is 0,1 and 2.

jsibbison-square: If you look at the final_output_tool tests in the tests/../agent.rs file you might be able to add a test with a fake provider to test this.

crates/goose/src/agents/retry.rs
```diff
@@ -0,0 +1,323 @@
...
+            "Shell command completed with status: {}, stdout: {}, stderr: {}",
+            output.status,
+            String::from_utf8_lossy(&output.stdout),
+            String::from_utf8_lossy(&output.stderr)
+        );
+
+        Ok(output)
+    };
+
+    tokio::time::timeout(timeout, future).await?
```
jsibbison-square: Can we tell the difference between a timeout failure vs a command failure using this? The first thing people will want to know when it fails is was it a timeout or not.

crates/goose/src/agents/retry.rs
```diff
@@ -0,0 +1,323 @@
...
+        };
+
+        let retry_timeout = get_retry_timeout(&retry_config);
+        let cleanup_timeout = get_cleanup_timeout(&retry_config);
+
+        assert_eq!(retry_timeout, Duration::from_secs(60));
+        assert_eq!(cleanup_timeout, Duration::from_secs(300));
+        assert_ne!(retry_timeout, cleanup_timeout);
+    }
+}
```
jsibbison-square, line 498: Please check if there is a way to condense some of these into less tests that are more likely to catch bugs. Eg. I don't think its worth having a test that timeout and cleanup timeout are different particularly given there are individual tests that check each individually.

crates/goose/src/agents/types.rs
```diff
@@ -8,6 +8,50 @@ use tokio::sync::{mpsc, Mutex};
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub on_failure: Option<String>,
+    /// Timeout in seconds for individual shell commands (default: 300 seconds)
+    /// Can also be configured globally via GOOSE_RECIPE_RETRY_TIMEOUT_SECONDS environment variable
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub timeout_seconds: Option<u64>,
+    /// Timeout in seconds for cleanup commands (default: 600 seconds)
+    /// Can also be configured globally via GOOSE_RECIPE_CLEANUP_TIMEOUT_SECONDS environment variable
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cleanup_timeout_seconds: Option<u64>,
```
jsibbison-square: Mentioned previously but I think we should keep to one domain concept (on_failure) rather than having two and trying to link them conceptually.

crates/goose/src/agents/types.rs
```diff
@@ -8,6 +8,50 @@ use tokio::sync::{mpsc, Mutex};
...
+
+/// A single success check to validate recipe completion
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SuccessCheck {
+    /// The type of success check to perform
+    #[serde(rename = "type")]
+    pub check_type: SuccessCheckType,
+    /// The command or instruction for the success check
+    pub command: String,
+}
```
jsibbison-square, line 59: It seems like we might be able to do something like the following in order not have extra parameters all intermingle at the top level from different check_types

```
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(tag = "type")]
pub enum SuccessCheck {
    Shell {
        command: String,
    },
    Http {
        url: String,
        expected_status: u16,
    },
    // Add more variants here
}
```

documentation/docs/guides/recipes/recipe-reference.md
```diff
@@ -136,6 +137,87 @@ sub_recipes:
...
+  max_retries: 3
+  timeout_seconds: 30
+  cleanup_timeout_seconds: 60
+  checks:
+    - type: shell
+      command: "curl -f http://localhost:8080/health"
+    - type: shell  
+      command: "pgrep -f 'web-service' > /dev/null"
+  on_failure: "systemctl stop web-service || killall web-service"
+```
```
jsibbison-square, line 210: Thanks for updating the documentation, maybe lets just pick one of the examples to not make the page too long.

jsibbison-square (approved):
Some issues to explore but overall looks good

crates/goose/src/agents/agent.rs
```diff
@@ -1042,6 +1132,23 @@ impl Agent {
                             yield AgentEvent::Message(message);
                         }
                     }
+
+                    // Execute retry logic if configured
+                    match self.handle_retry_logic(&mut messages, &session, &initial_messages).await {
+                        Ok(should_retry) => {
+                            if should_retry {
+                                info!("Retry logic triggered, restarting agent loop");
```
jsibbison-square, line 1091: We should probably be resetting self.tool_monitor also. Otherwise the count might still be set when we reset.

crates/goose-cli/src/cli.rs
DOsinga: +1 - this is getting a tad out of hand

crates/goose-cli/src/cli.rs
cloud-on-prem: yeah agree; for some reason, it felt more readable that way, but the real issue is the top level signature
`let (input_config, session_settings, sub_recipes, final_output_response, retry_config)`  that's making it hard to keep track of each param. We should address it separately.

crates/goose-cli/src/session/builder.rs
cloud-on-prem: This was just to keep track which None 🤣
removing it.

crates/goose-cli/src/session/mod.rs
cloud-on-prem: yep, I'll remove these ones

crates/goose/src/agents/agent.rs
cloud-on-prem: Generally, I leave the docstring comments on. Happy to remove them if we want to establish that as a convention going forward.

crates/goose/src/agents/agent.rs
cloud-on-prem: yep, I found a few places where we are not passing the session config:

- https://github.com/block/goose/blob/main/crates/goose/examples/agent.rs#L38
- https://github.com/block/goose/blob/main/crates/goose-ffi/src/lib.rs#L250

Some unit tests as well.

crates/goose-cli/src/cli.rs
cloud-on-prem: Ended up refactoring it into RecipeInfo.

crates/goose-cli/src/session/builder.rs
cloud-on-prem: Got rid of it.

crates/goose-cli/src/session/mod.rs
cloud-on-prem: done

crates/goose/src/agents/agent.rs
cloud-on-prem: Yep, makes sense. I am going to let it bubble up.

crates/goose/src/agents/retry.rs
cloud-on-prem: yep; cleaned up the naming

crates/goose/src/agents/retry.rs
cloud-on-prem: added validation to RetryConfig

crates/goose/src/agents/retry.rs
cloud-on-prem: there are similar ones but they are not really exported for reuse.

crates/goose/src/agents/agent.rs
cloud-on-prem: With max_retries, since we are "retrying" it needs to run one extra time. That is, once for the original (which is zero) and the once for the number of retries.

cloud-on-prem: good call; I have added some tests there

crates/goose/src/agents/types.rs
cloud-on-prem: yes, this makes sense. We could wait but using a tagged enum makes it simpler to extend. Adding it now.

crates/goose/src/agents/retry.rs
cloud-on-prem: added an extra message for time out failures

crates/goose-cli/src/cli.rs
```diff
@@ -828,27 +838,24 @@ pub async fn cli() -> Result<()> {
             provider,
             model,
         }) => {
-            let (input_config, session_settings, sub_recipes, final_output_response) = match (
-                instructions,
-                input_text,
-                recipe,
-            ) {
+            let recipe_info = match (instructions, input_text, recipe) {
                 (Some(file), _, _) if file == "-" => {
```
lifeizhou-ap, line 841: Hi @cloud-on-prem 

I feel it could cause a bit confusion to return recipe_info for each case even for non recipe related. 

Maybe it can be returned as 

`let (input_config, recipe_info) = match (.....`

For non recipe related, the recipe_info is `None`

And RecipeInfo have all the existing fields above except the inputConfig.

crates/goose/src/agents/agent.rs
```diff
@@ -173,6 +176,81 @@ impl Agent {
...
+            );
+            return Ok(false);
+        }
+
+        if let Some(on_failure_cmd) = &retry_config.on_failure {
+            info!("Executing on_failure command: {}", on_failure_cmd);
+            execute_on_failure_command(on_failure_cmd, retry_config).await?;
+        }
+
+        messages.clear();
```
lifeizhou-ap: can extract to a function such as `reset_status_for_retry` to reset the message and clear final_out tool state

crates/goose/src/agents/agent.rs
```diff
@@ -173,6 +176,81 @@ impl Agent {
         }
     }
 
+    /// Reset the retry attempts counter to 0
+    pub async fn reset_retry_attempts(&self) {
```
lifeizhou-ap, line 182: the agent file is large and growing. Shall we create a separate module/file (or in the retry.rs) with a struct to manage the reset attempts state?

crates/goose/src/agents/agent.rs
```diff
@@ -173,6 +176,81 @@ impl Agent {
...
+        *retry_attempts
+    }
+
+    /// Get the current retry attempts count
+    pub async fn get_retry_attempts(&self) -> u32 {
+        *self.retry_attempts.lock().await
+    }
+
+    /// Handle retry logic for the agent reply loop
+    async fn handle_retry_logic(
```
lifeizhou-ap, line 197: This function could be moved to the retry.rs, and can return results like
skipped,  max_attempts_reached, retried. 

Then agent can add a message, or reset_retry_status accordingly based on the results.  In this way, the retry logic are in the retry.rs.

crates/goose/src/agents/retry.rs
```diff
@@ -0,0 +1,338 @@
...
+    command: &str,
+    timeout: std::time::Duration,
+) -> Result<std::process::Output> {
+    debug!(
+        "Executing shell command with timeout {:?}: {}",
+        timeout, command
+    );
+
+    let future = async {
+        let output = if cfg!(target_os = "windows") {
```
lifeizhou-ap: could be less duplication
```
let command = if cfg!(target_os = "windows") {
            Command::new("cmd")
                .args(["/C", command])
} else {
            Command::new("sh")
                .args(["-c", command])
};
let output = command.stdout(Stdio::piped())
        .stderr(Stdio::piped())
        .stdin(Stdio::null())
        .kill_on_drop(true)
        .output()
        .await?
```

crates/goose-cli/src/cli.rs
cloud-on-prem: yep that's a good suggestion; it was getting a bit unwieldy - refactored it now 👍

crates/goose/src/agents/agent.rs
cloud-on-prem: refactored all of it to be nicer now.

crates/goose/src/agents/agent.rs
cloud-on-prem: done

lifeizhou-ap (approved):

michaelneale:
I can't run `just run-ui` for this: 

```

:fire: Unexpected error occurred. Log saved to /Users/micn/Development/goose/ui/desktop/openapi-ts-error-1753165522594.log
:fire: Unexpected error occurred. Missing $ref pointer "#/components/schemas/RetryConfig". Token "RetryConfig" does not exist.
```

PR #3477, opened 2025-07-17
fix: working dir was not being set correctly 

michaelneale:
this likely was a bug around for some time.

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -372,14 +372,8 @@ impl Session {
             let working_dir = Some(
                 std::env::current_dir().expect("failed to get current session working directory"),
             );
-            session::persist_messages_with_schedule_id_and_working_dir(
-                session_file,
-                &self.messages,
-                Some(provider),
-                self.scheduled_job_id.clone(),
-                working_dir,
-            )
-            .await?;
+            session::persist_messages(session_file, &self.messages, Some(provider), working_dir)
+                .await?;
         }
 
         // Track the current directory and last instruction in projects.json
@@ -499,11 +493,10 @@ impl Session {
                                     std::env::current_dir()
                                         .expect("failed to get current session working directory"),
                                 );
-                                session::persist_messages_with_schedule_id_and_working_dir(
+                                session::persist_messages(
                                     session_file,
                                     &self.messages,
                                     Some(provider),
-                                    self.scheduled_job_id.clone(),
                                     working_dir,
                                 )
                                 .await?;
@@ -716,11 +709,13 @@ impl Session {
 
                         // Persist the summarized messages
                         if let Some(session_file) = &self.session_file {
+                            let working_dir = std::env::current_dir().ok();
                             session::persist_messages_with_schedule_id(
                                 session_file,
                                 &self.messages,
                                 Some(provider),
                                 self.scheduled_job_id.clone(),
+                                working_dir,
                             )
                             .await?;
                         }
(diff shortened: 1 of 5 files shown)


crates/goose-cli/src/session/mod.rs
```diff
@@ -369,13 +369,11 @@ impl Session {
...
         if let Some(session_file) = &self.session_file {
-            session::persist_messages_with_schedule_id(
-                session_file,
-                &self.messages,
-                Some(provider),
-                self.scheduled_job_id.clone(),
-            )
-            .await?;
+            let working_dir = Some(
+                std::env::current_dir().expect("failed to get current session working directory"),
```
DOsinga, line 373: do we want to panic on this?

crates/goose-cli/src/session/mod.rs
```diff
@@ -491,11 +489,15 @@ impl Session {
...
+                                let working_dir = Some(
+                                    std::env::current_dir()
+                                        .expect("failed to get current session working directory"),
+                                );
+                                session::persist_messages(
                                     session_file,
                                     &self.messages,
                                     Some(provider),
-                                    self.scheduled_job_id.clone(),
+                                    working_dir,
```
DOsinga, line 509: do we not want to persist the schedule id too?

crates/goose/src/session/storage.rs
```diff
@@ -1323,6 +1344,106 @@ pub async fn generate_description_with_schedule_id(
...
+/// Generate a description for the session using the provider, including an optional scheduled job ID and working directory
+///
+/// This function is called when appropriate to generate a short description
+/// of the session based on the conversation history.
+///
+/// Security features:
+/// - Validates file paths to prevent directory traversal
+/// - Limits context size to prevent resource exhaustion
+/// - Uses secure file operations for saving
+async fn generate_description_with_schedule_id_and_working_dir(
```
DOsinga: this seems to be almost an exact copy of the one without the working directory. this should just replace the other one, no? 

let's have this function just return the description it generates. the caller is already writing out the messages with meta data, so we can do that just there instead.

crates/goose-cli/src/session/mod.rs
michaelneale: yeah I think so - trying to think of a case where you would even want to continue with this in this case, can't think of one?

crates/goose/src/session/storage.rs
michaelneale: ah damn, I thought I cleaned those ones up (should just be the same name as before with new param) - I blame goose for following an existing pattern too closely (ie `_with_schedule_id`)

crates/goose-cli/src/session/mod.rs
michaelneale: yes, we should. I remember seeing it in my sessions (all null however, but seems ok to persist it as null in the jsonl line)

crates/goose/src/session/storage.rs
michaelneale: yeah actually don't want dir in description (as it is already in as a parameter)

crates/goose-cli/src/session/mod.rs
```diff
@@ -491,11 +489,15 @@ impl Session {
 
                             // Persist messages with provider for automatic description generation
                             if let Some(session_file) = &self.session_file {
-                                session::persist_messages_with_schedule_id(
+                                let working_dir = Some(
+                                    std::env::current_dir()
+                                        .expect("failed to get current session working directory"),
+                                );
+                                session::persist_messages(
```
michaelneale: this should also be with schedule_id

michaelneale:
@DOsinga trying with the panic on no working dir - but I see a CI failure due to a panic so will check if this is some side effect of running containerised

PR #3482, opened 2025-07-17
fix issue with generating a share url and add missing url verification in settings

zanesq:
fix issue with generating a share url and add missing url verification in settings

content security policy was too tight, also noticed todo missing url verification

### DIFF
--- a/ui/desktop/index.html
+++ b/ui/desktop/index.html
@@ -2,7 +2,7 @@
 <html>
   <head>
     <meta charset="UTF-8" />
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'; img-src 'self' data: https:; connect-src 'self' http://127.0.0.1:* https://api.github.com https://github.com https://objects.githubusercontent.com; object-src 'none'; frame-src 'none'; font-src 'self' data: https:; media-src 'self' mediastream:; form-action 'none'; base-uri 'self'; manifest-src 'self'; worker-src 'self';" />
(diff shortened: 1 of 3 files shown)


ui/desktop/index.html
```diff
@@ -2,7 +2,7 @@
 <html>
   <head>
     <meta charset="UTF-8" />
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'; img-src 'self' data: https:; connect-src 'self' http://127.0.0.1:* https://api.github.com https://github.com https://objects.githubusercontent.com; object-src 'none'; frame-src 'none'; font-src 'self' data: https:; media-src 'self' mediastream:; form-action 'none'; base-uri 'self'; manifest-src 'self'; worker-src 'self';" />
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'; img-src 'self' data: https:; connect-src 'self' http://127.0.0.1:* https:; object-src 'none'; frame-src 'none'; font-src 'self' data: https:; media-src 'self' mediastream:; form-action 'none'; base-uri 'self'; manifest-src 'self'; worker-src 'self';" />
```
Kvadratni, line 5: what? were did github come from?

zanesq: Good catch! It's in case our app needs to reach out to github but I guess its not needed now so will remove 👍

zanesq: Oh thats right we use it in the updater code

Kvadratni (approved):

ui/desktop/index.html
zanesq: Just matching what we already do for content security policy for server responses https://github.com/block/goose/blob/e0b3b0fc7068c271720835af7e91e2beb9f86299/ui/desktop/src/main.ts#L1567

PR #3488, opened 2025-07-17
feat: switch from mcp_core::Role to rmcp::model::Role

alexhancock:
Switch from `mcp_core::Role` to `rmcp::model::Role`


jamadeo (approved):
excellent

alexhancock:
There is a small issue I think related to the changes that affect openapi spec generation which cause user messages not to show up properly. Will fix tomorrow before merge.

PR #3489, opened 2025-07-17
refactor: Centralise deeplink encode and decode into server

jsibbison-square:
### Summary

Centralized deeplink recipe encoding/decoding logic in `goose-server` to ensure consistency and avoid incompatibility across different implementations, as discussed in [PR #3446 review](https://github.com/block/goose/pull/3446#pullrequestreview-3024092353).

### Key Changes

* **Centralized Logic:**

  * All deeplink recipe encoding/decoding now uses a single implementation in the backend (`goose-server`).
  * Adopted base64 (URL-safe, no padding) encoding for portability and compactness.
  * Backend includes fallback logic for legacy-encoded deep URLs.

* **Refactoring:**

  * Updated CLI and desktop code to use the new centralized server functions for encoding/decoding (call overhead is now <1–2ms vs. previous inline logic).
  * Cleaned up GUI code by merging the incomplete `RecipeConfig` struct into the more complete `Recipe` struct—frontend now has a single config representation.

### Tested
- From Recipe Library
  - 'Use' or 'Load Recipe' from View.     _Does not work due existing bug https://github.com/block/goose/pull/3501_ 
- From recipe generator
  - Open deeplink
  - Import deeplink
- From cli
  - Open deeplink
  - Import deeplink
- Create recipe
- Generate recipe from session
- Create schedule import deeplink
- Edit recipe
- View recipe

### Notes

* Further refactors/cleanups are possible but have been left out to keep this PR small enough.


crates/goose-server/src/routes/recipe.rs
```diff
@@ -84,8 +105,70 @@ async fn create_recipe(
...
+            Err(StatusCode::BAD_REQUEST)
+        }
+    }
+}
+
 pub fn routes(state: Arc<AppState>) -> Router {
     Router::new()
         .route("/recipe/create", post(create_recipe))
+        .route("/recipes/encode", post(encode_recipe))
+        .route("/recipes/decode", post(decode_recipe))
```
jsibbison-square, line 136: I think we should move /recipe to /recipes in a future pr

ui/desktop/src/components/ui/DeepLinkModal.tsx
```diff
@@ -1,42 +1,60 @@
...
-      ...recipeConfig,
-      instructions,
-      activities,
+  useEffect(() => {
+    let isCancelled = false;
+
+    const generateLink = async () => {
+      setIsGeneratingLink(true);
+      try {
+        const currentConfig = {
```
DOsinga, line 29: here and elsewhere, I'm not sure why we use the word config when we meant to say recipe or recipeToEncode or some such

ui/desktop/src/main.ts
```diff
@@ -49,6 +49,34 @@ import {
...
+// API URL constructor for main process before window is ready
+function getApiUrlMain(endpoint: string, dynamicPort: number): string {
+  const host = process.env.GOOSE_API_HOST || 'http://127.0.0.1';
+  const port = dynamicPort || process.env.GOOSE_PORT;
+  const cleanEndpoint = endpoint.startsWith('/') ? endpoint : `/${endpoint}`;
+  return `${host}:${port}${cleanEndpoint}`;
+}
+
+// When opening the app with a deeplink, the window is still initializing so we have to duplicate some window dependant logic here.
+async function decodeRecipeMain(deeplink: string, port: number): Promise<Recipe | null> {
```
DOsinga, line 63: ... and hopefully in the not so distant future we can just send the deeplink along when we initialize the backend

crates/goose-server/src/routes/recipe.rs
```diff
@@ -34,6 +35,26 @@ pub struct CreateRecipeResponse {
     error: Option<String>,
 }
 
+#[derive(Debug, Deserialize)]
```
DOsinga, line 38: It looks like this entire recipe thing is not part of our open-api integration somehow. this should be integrated with openapi.rs and then we can autogenerate the ts code. can you check how much work it is to do this the right way? if too much, we can do in a follow up

DOsinga (approved):
thanks for jumping on this. my main concern is the way that we have done the API integration in the past and now follow up, sidestepping our open-api setup -

`just generate-openapi`

ui/desktop/src/main.ts
jsibbison-square: yeah only soo much refactoring could be done at a time.

crates/goose-server/src/routes/recipe.rs
jsibbison-square: I'm not familiar with the open-api pieces. I can rectifying that in a future pr.

crates/goose/src/recipe_deeplink.rs
```diff
@@ -0,0 +1,108 @@
+use anyhow::Result;
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine as _};
+use thiserror::Error;
+
+use crate::recipe::Recipe;
+
+#[derive(Error, Debug)]
+pub enum DecodeError {
```
cloud-on-prem, line 8: For future, it would be nice to distinguish the errors.

```rs
#[derive(Error, Debug)]
pub enum DecodeError {
    #[error("Invalid base64 encoding")]
    InvalidBase64,
    #[error("Invalid JSON format: {0}")]
    InvalidJson(String),
    #[error("Invalid recipe format: missing required fields")]
    InvalidRecipe,
    #[error("Unsupported deeplink format")]
    UnsupportedFormat,
}
```

crates/goose-server/src/routes/recipe.rs
cloud-on-prem: agree

PR #3493, opened 2025-07-17
Feature/diff viewer sidecar

Kvadratni:
this is a fresh branch to fix rebase on main for https://github.com/block/goose/pull/3040

## Implement checkpoint system with intelligent diff splitting

Adds file versioning and enhanced diff handling for better change management and review.

**Checkpoint system:**
- Automatic backup creation before file modifications with timestamp-based storage
- Persistent checkpoint index tracking file modification history  
- Restore functionality with diff generation for change tracking

**Enhanced diff generation:**
- File-type aware context sizing for better hunk boundaries
- Intelligent splitting of large diffs into manageable chunks
- Improved review experience for AI-generated changes

**New tools:**
- `list_checkpoints` - View available file versions
- `restore_checkpoint` - Revert files to previous states

**Integration:**
- Checkpoint payloads embedded in message stream for UI consumption
- Maintains backward compatibility with existing undo functionality
- Supports both traditional string replacement and Editor API workflows

Addresses large diff readability issues while providing robust file versioning capabilities.

https://github.com/user-attachments/assets/703ee4a4-0ec2-4cf8-bba3-79964af35b67


michaelneale:
I think to be a side car there needs to be a change to main.ts to process a message to increase underlying window to the right (you can't do it inside the one window) - otherwise we take real estate away from what is there (which is fine if default was wider?) - do you think sidecars should slide out to the right @aharvard ?

Kvadratni:
It was doing that. My bug. Lost it during rebase

aharvard:
fixed some issues @Kvadratni : https://github.com/block/goose/pull/3520

aharvard:
two things are still bothering me
1. the chat message thread grows when the window resizes. I think its width should remain as-is. 
2. the sidecar does not slide in from the right

not blockers, but I'm gonna try to address those before my EOD

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -236,6 +245,34 @@ function BaseChatContent({
...
+        for (const response of responses) {
+          const matchingRequest = toolRequests.find((req) => req.id === response.id);
+          if (matchingRequest && hasDiffContent(response)) {
+            return true;
+          }
+        }
+      }
+      return false;
+    });
+  }, [filteredMessages, messages]);
```
DOsinga: dont use useMemo on lists

crates/goose/src/message.rs
```diff
@@ -260,10 +269,34 @@ impl From<Content> for MessageContent {
...
+                    ResourceContents::TextResourceContents {
+                        uri,
+                        text,
+                        mime_type: _,
+                    } => {
+                        // For special URIs like goose://checkpoint, preserve as resource
+                        if uri.starts_with("goose://") {
+                            MessageContent::EmbeddedResource(resource)
+                        } else {
+                            // For regular resources, convert to text as before
```
DOsinga: "as before"?

Kvadratni: fixed

michaelneale:
main thing I would like to change is that the developer MCP doesn't assume any interactivity (be that checkpoints, diff viewer) out of the box - as headless mode is common for it so you don't want tools in there in that case which aren't relevant to it, increasing context and confusion.

jamadeo:
@Kvadratni @michaelneale @DOsinga do we want to keep going on this?

Kvadratni:
I will keep it as a reference. I will try to reimplement using MCP-UI. Will mark as draft

DOsinga:
thanks, closing this for now

PR #3496, opened 2025-07-18
docs: add detailed instruction for WSL users

GaryZhous:
I am a WSL user, and sometimes I encountered some issues with compiling goose. So I made this pull request that updates the `CONTRIBUTING.md` file to include additional setup instructions for Windows Subsystem for Linux (WSL) users to avoid potential compiler linking errors.

Documentation updates:

* [`CONTRIBUTING.md`](diffhunk://#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R18-R28): Added a new section with instructions for WSL users to install `build-essential` and `libxcb` packages to prevent `cc` linking errors. This includes the exact commands to run for installation.


cbruyndoncx:
Should this not be reflected in the cross.toml ?
theoretically you can install different linux OS in WSL, or do you assume default install from windows is taken ?

GaryZhous:
> Should this not be reflected in the cross.toml ?
> 
> theoretically you can install different linux OS in WSL, or do you assume default install from windows is taken ?

Yep, I assume default install from windows.

taniacryptid:
Thank you for this update!

PR #3500, opened 2025-07-18
Replace mcp_core::content types with rmcp::model types

jamadeo:
Basically, delete crates/mcp-core/src/content.rs and add in rmcp types in its place. Kicked off by me and mostly done by Goose+Claude.


crates/goose-cli/src/session/export.rs
```diff
@@ -236,67 +234,68 @@ pub fn tool_response_to_markdown(resp: &ToolResponse, export_all_content: bool)
...
-                                    "sh" => "bash",
-                                    _ => mime_type
-                                        .as_ref()
-                                        .map(|mime| if mime == "text" { "" } else { mime })
-                                        .unwrap_or(""),
-                                };
-
-                                md.push_str(&format!("**File:** `{}`\n", uri));
+                    None => {
+                        if let Some(image_content) = content.as_image() {
```
jamadeo: The ergonomics of a match on the Annotated type are a bit weird, and here the model chose to do some nested conditionals. I'll see if this can be improved

crates/goose-cli/src/session/mod.rs
```diff
@@ -359,7 +359,23 @@ impl Session {
...
+                content: match msg.content {
+                    mcp_core::prompt::PromptMessageContent::Text { text } => rmcp::model::PromptMessageContent::Text { text },
+                    mcp_core::prompt::PromptMessageContent::Image { image } => rmcp::model::PromptMessageContent::Image { image },
+                    mcp_core::prompt::PromptMessageContent::Resource { resource } => {
+                        rmcp::model::PromptMessageContent::Resource { resource }
+                    },
+                }
+            }
+        }).collect();
+        Ok(converted_messages)
```
jamadeo, line 388: This should go away once we do the PromptMessage migration. But we needed to do a bit of it here since we want to work with the rmcp content types.

crates/goose/src/context_mgmt/truncate.rs
```diff
@@ -140,19 +142,22 @@ fn estimate_message_tokens(message: &Message, estimate_fn: &dyn Fn(&str) -> usiz
...
+                                    match &resource_content.resource {
+                                        ResourceContents::TextResourceContents { text, .. } => {
+                                            total_tokens += estimate_fn(text);
+                                        }
+                                        _ => total_tokens += 5, // Small overhead for other resource types
                                     }
-                                    _ => total_tokens += 5, // Small overhead for other resource types
+                                } else {
+                                    total_tokens += 5; // Small overhead for other content types
                                 }
```
jamadeo, line 155: Another case of a match turned nested

crates/goose/src/message.rs
```diff
@@ -257,13 +258,23 @@ impl MessageContent {
...
+            RawContent::Resource(resource) => {
+                let text = match &resource.resource {
+                    ResourceContents::TextResourceContents { text, .. } => text.clone(),
+                    ResourceContents::BlobResourceContents { blob, .. } => {
+                        format!("[Binary content: {}]", blob.clone())
+                    }
+                };
+                MessageContent::text(text)
+            }
+            RawContent::Audio(_) => todo!(),
```
jamadeo: I need to look into how we'd handle Audio--we obviously don't want to panic here

alexhancock (approved):

ui/desktop/src/api/types.gen.ts
```diff
@@ -22,13 +22,22 @@ export type ConfigResponse = {
...
+    type: string;
+} | {
+    resource: ResourceContents;
+    type: string;
+} | {
+    annotations?: Annotations;
+    data: string;
+    mimeType: string;
+    type: string;
+};
```
jamadeo, line 40: this looks like a bit less useful than what we had before

alexhancock: are there any frontend imports of this that may need to be updated? we don't have a ton of testing in the frontend unfortunately

jamadeo: Yeah the TS checker passing does give some confidence that this change is OK, but I'll pay some extra attention to this to be sure

jamadeo: ah, turns out we are not using this type at all, because we have our own extended MessageContent type that we use instead

Issue #3503, opened 2025-07-18
goose cold start with openrouter | ollama


michaelneale:
Motivation: 

goose when it starts (brand new) shows a helpful wall of providers and config, but we can do better with an out of the box flow for those who really are new.

Solutions: 
* A one click openrouter signup with model selection applied (lead, worker and editor models which are suitably low costed)
* Another choice of an ollama install + autoconfigure (checking there is capacity for this). 

Between these people can hit the ground running from a cold start. 

Implementation high level: 

Openrouter: 
* PKCE openrouter signup with oauth2.0 (which openrouter offers)
* goose crate to have core logic which is consumed by both -cli and -server
* post PKCE flow an option to top up with credits into an account

Ollama: 
* ask and detect capability of running ollama locally with suitable (curated) model
* install and pull model
* auto configure goose

These can be the default choices on empty state start, or a distribution etc... (still TBD).

michaelneale:
merged

PR #3507, opened 2025-07-18
feat: openrouter out of the box experience for goose installations

michaelneale:
working on: https://github.com/block/goose/issues/3503

this will use PKCE when setting up goose for the first time to automatically setup goose with openrouter. 

uses the goose crate for core logic for managing the flow and keys and installing the config, and consumes it from the cli, and also goose-server/desktop use the same logic. 

unit tests for the rust side core code. 

Please see this short demo of it running the desktop for first time - pops up and logs in to openrouter, then back to goose, ready to go, configured with models and keys, no manual work needed: 

https://github.com/user-attachments/assets/7c4ff44f-0e5d-4173-a23e-172fe5258019

Screenshot 2025-07-23 at 5 26 14 pm

This is designed to be the first flow onboard experience for a brand new install of goose. 

image

Screenshot 2025-07-23 at 7 35 27 pm

Screenshot 2025-07-23 at 5 26 26 pm

Screenshot 2025-07-23 at 5 26 22 pm

Upnext: 
* [x] finalise the `goose configure` type of flow for when it is brand new so it is same for CLI as desktop
* [x] choose default models for lead/worker and editor and test 
* [x] remove detritus OPENROUTER_*.md files left around while it was WIP
* [x] Test you can get back to wall of providers (after default configuration has run - you change your mind)
* [x] Test you can get back to wall of providers when variable is not set
* [x] Consider if launch link can prompt specific flow? 
* [x] Check can set GOOSE_SETUP=false and it will bypass (not just unset)
* [x] change command to be a sub command of config, not auth command 

To test: 

* mv your ~/.config/goose dir out of the way (I usually rename it to goose.bak), 
* ensure you unset GOOSE_MODEL and GOOSE_PROVIDER env vars from your environment if you use them. 
* run `just run-ui` which will set it up fresh
* log in to openrouter when it pops up
* done!


crates/goose-cli/src/cli.rs
```diff
@@ -1,7 +1,7 @@
 use anyhow::Result;
```
DOsinga: I'm not sure what's happening in this file. does it change anything?

crates/goose-server/src/routes/setup.rs
```diff
@@ -0,0 +1,122 @@
...
+
+#[derive(Serialize)]
+pub struct SetupResponse {
+    pub success: bool,
+    pub message: String,
+}
+
+pub fn routes(state: Arc<AppState>) -> Router {
+    Router::new()
+        .route("/setup/openrouter/start", post(start_openrouter_setup))
```
DOsinga: let's remove the nesting here since we only have one handler. and I'd just call the handler handle_openrouter in line with the other naming here; it is in the setup module, so no need to repeat that.

crates/goose-server/src/routes/setup.rs
```diff
@@ -0,0 +1,122 @@
...
+        .with_state(state)
+}
+
+async fn start_openrouter_setup(
```
DOsinga, line 20: can we strip the LLM comments a bit here?

crates/goose-server/src/routes/setup.rs
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_oauth_flow_mutex() {
```
DOsinga: does this test test anything related to the code here? as far as I can see we're just testing that mutex's work

ui/desktop/src/utils/openRouterSetup.ts
```diff
@@ -0,0 +1,29 @@
...
+  try {
+    const baseUrl = `${window.appConfig.get('GOOSE_API_HOST')}:${window.appConfig.get('GOOSE_PORT')}`;
+    const response = await fetch(`${baseUrl}/setup/openrouter/start`, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+
+    if (!response.ok) {
```
DOsinga: this should just return success: false and the relevant message

ui/desktop/src/utils/openRouterSetup.ts
```diff
@@ -0,0 +1,29 @@
...
+      throw new Error(`HTTP error! status: ${response.status}`);
+    }
+
+    const result = await response.json();
+    return result;
+  } catch (error) {
```
DOsinga: I don't think you want to generally catch errors here like this, just let it bubble up

ui/desktop/src/utils/openRouterSetup.ts
```diff
@@ -0,0 +1,29 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+
+    if (!response.ok) {
+      throw new Error(`HTTP error! status: ${response.status}`);
+    }
+
+    const result = await response.json();
```
DOsinga: use safeJsonParse here, now in main

ui/desktop/src/components/ProviderGuard.tsx
```diff
@@ -11,42 +14,151 @@ export default function ProviderGuard({ children }: ProviderGuardProps) {
...
-          console.log('No provider/model configured, redirecting to welcome');
-          navigate('/welcome', { replace: true });
+          console.log('ProviderGuard - No provider/model configured, showing first time setup');
+          setShowFirstTimeSetup(true);
         }
       } catch (error) {
         console.error('Error checking provider configuration:', error);
-        // On error, assume no provider and redirect to welcome
-        navigate('/welcome', { replace: true });
+        // On error, show first time setup
```
DOsinga: I see this was already the case, but that looks like a way to lock the user into a loop

ui/desktop/src/components/ProviderGuard.tsx
```diff
@@ -11,42 +14,151 @@ export default function ProviderGuard({ children }: ProviderGuardProps) {
...
   useEffect(() => {
     const checkProvider = async () => {
       try {
         const config = window.electron.getConfig();
+        console.log('ProviderGuard - Full config:', config);
+
         const provider = (await read('GOOSE_PROVIDER', false)) ?? config.GOOSE_DEFAULT_PROVIDER;
         const model = (await read('GOOSE_MODEL', false)) ?? config.GOOSE_DEFAULT_MODEL;
 
+        console.log('ProviderGuard - Provider:', provider, 'Model:', model);
```
DOsinga: do we need this log (and the other one) or just during debug?

ui/desktop/src/components/ProviderGuard.tsx
```diff
@@ -11,42 +14,151 @@ export default function ProviderGuard({ children }: ProviderGuardProps) {
...
+      } else {
+        setOpenRouterSetupState({
+          show: true,
+          title: 'Setup Pending',
+          message: result.message,
+          showProgress: false,
+          showRetry: true,
+        });
+      }
+    } catch (error) {
```
DOsinga: do we need this catch here? we should handle all errors in startOpenRouter() I think

ui/desktop/src/components/ProviderGuard.tsx
```diff
@@ -11,42 +14,151 @@ export default function ProviderGuard({ children }: ProviderGuardProps) {
...
+          message: 'OpenRouter has been configured successfully.',
+          showProgress: false,
+          showRetry: false,
+          autoClose: 3000,
+        });
+
+        // Reload the page after successful setup
+        setTimeout(() => {
+          window.location.reload();
+        }, 3000);
```
DOsinga: this is slightly unexpected given that we also have autoClose: 3000

crates/goose/src/config/signup_openrouter/server.rs
```diff
@@ -0,0 +1,215 @@
...
+    Query(params): Query<CallbackQuery>,
+    state: axum::extract::State<
+        std::sync::Arc<tokio::sync::Mutex<Option<oneshot::Sender<String>>>>,
+    >,
+) -> impl IntoResponse {
+    // Check for error first
+    if let Some(error) = params.error {
+        return (
+            StatusCode::BAD_REQUEST,
+            Html(format!(
```
DOsinga: so for these html templates, consider using jinja2 for rendering instead of adding a dependency; we could make them external too

crates/goose-server/src/routes/setup.rs
```diff
@@ -0,0 +1,122 @@
...
+            return Ok(Json(SetupResponse {
+                success: false,
+                message: "Authentication flow is already in progress. Please wait.".to_string(),
+            }));
+        }
+    };
+
+    tracing::info!("Acquired OAuth flow lock");
+
+    // Run the existing PKCE flow
```
angelahning: why does it say "existing PKCE flow"? I thought we are initializing the flow here.

angelahning:
Screenshot 2025-07-23 at 6 24 43 PM

the initial window looks off on my local - text is greyed out.

angelahning:
nit: it'd be nice to augment this error to make it clear to users to add credits their open router account. 
Screenshot 2025-07-23 at 6 30 39 PM

angelahning (commented):
logic looks good to me. I am not familiar with Rust to approve.

crates/goose-cli/src/cli.rs
michaelneale: this was a left over late yesterday - I originally had it as a top level command - but thought "goose configure" was the home - I can restore it to state of main so it won't change.

crates/goose-server/src/routes/setup.rs
michaelneale: sorry - that is actually me making sure it re-uses the code from the core goose crate (goose is happy to duplicate code!) - so not actually "existing flow" more "existing code" - another junk comment to remove!

crates/goose-server/src/routes/setup.rs
michaelneale: yeah junk - there was another test there which was junk which I deleted but forgot this one, good catch.

crates/goose/Cargo.toml
```diff
@@ -66,6 +65,8 @@ etcetera = "0.8.0"
 rand = "0.8.5"
 utoipa = { version = "4.1", features = ["chrono"] }
 tokio-cron-scheduler = "0.14.0"
+urlencoding = "2.1"
+html-escape = "0.2"
```
michaelneale: @DOsinga so with jinja2 probably wouldn't need html-escape?

DOsinga:
* when you cancel, it takes you back to the grid of providers, where if you click openrouter it goes to the normal flow. should we hook that up to this?
* when I was done it asked me if I wanted to redirect to http://localhost:3000/?code=81afeb13-4955-489b-a5a2-e310898e98a4 which seems confusing to a user. I said yes, it opened chrome and then nobody was listening on that port  (I think that is the default goosed port, but we pick a random port maybe on start?)
* if you go back to goose during the flow it shows:

image
would be nice to add a button to cancel and to retry there (or resume if possible)

DOsinga:
it worked the second time (?)

now when I select a model though, I get:

image

as you can see, it shows I have kimi-k2 selected, but I can't pick that from the pull down. I am guessing this is the same issue we've seen before when where it looks like we are not fetching the models or are failing? /cc @jamadeo - I think this does work from CLI

DOsinga:
in the CLI I'm seeing this:

image

It doesn't let me pick a model at all

michaelneale:
@DOsinga I expect as it is using the lead/worker out of the box, I bet if I simplify it then it would do better at that (although I don't think listing models will work very well with openrouter as there are now so many... not really sure what we do there longer term)

evangineer:
> @DOsinga I expect as it is using the lead/worker out of the box, I bet if I simplify it then it would do better at that (although I don't think listing models will work very well with openrouter as there are now so many... not really sure what we do there longer term)

Looking at how Roo Code works, getting both the full current list and being able to search that list would be ideal for me.  If you are doing manual configuration, you probably want the control and choice, I certainly do!

The pain was sufficient, that I found myself falling back to gemini cli for now!

As a stretch goal which is almost certainly beyond the scope here, I would like support for provider routing, but that's me.  At that point, I should probably be using LiteLLM instead to handle my OpenRouter connection.
https://openrouter.ai/docs/features/provider-routing

michaelneale:
@evangineer yeah - am curious about the routing in openrouter, people can try it, but I think in the end most agents will probably do a better job for now knowing what they need (everyone is trying something though). 
will check listing, aim isnt' to change how it works but just smooth the initial setup in this PR - I still think it will fall back to manual config as it does on main branch for now (PKCE really only for signup flow) - the idea for this is it is one click, you shoudn't need to pick a model to start with (but yeah - if you really want to pick one and have enough credit for it - absolutely)

ui/desktop/src/components/ProviderGuard.tsx
michaelneale: I think I will make it go back to navigate like before, so no worse, but not really sure what it should do here in this failure case... possibly this is just a junky catch that shouldn't be there (AI likes to add that in a lot).

michaelneale:
@DOsinga the GUI never did use the provider list loading from the fetch (only CLI did) so may have to do that as another patch (this is already getting long): https://github.com/block/goose/issues/3702 which is what you were seeing.

michaelneale:
@evangineer if you had time - woudl love a issue/write up of a reasonable flow for how we can search/filter the list, yeah it is so long I can't even use it even though it is dynamically loaded in the CLI (but not GUI). We can solve it in the GUI at least I guess... but cli you want to have some TUI that filters it (we can probably also show recommended models). this flow in any case will pick an opinionated default

michaelneale:
OK this is finally ready - I think I addressed most things (modulo things that will follow on) - so I think will merge this once it is clean - it really is just for initial experience, not advanced config (and ... what is most painful is I have to blow away my goose config to test and can't run goose while I am mucking around! you realise how much you depend on it!)

michaelneale:
tagging @taniashiba @blackgirlbytes as this will be a new "unboxing" flow (convenience) for really new users (which we may want to update notes/docs for when released - shouldn't change anything for existing users). Note that openrouter has ability to top up with things like cashapp - so I wonder if we can make a guide (I couldn't find a URL to open) which helps people create QR code and we can top up their account if starting?

image

and then: 

image

type of thing?

michaelneale:
this is killing me - doesn't reload correctly when running fully packaged (or this is another regegression that has crept in). whatever it is, will be quite a bit more debugging (with 30 min + roundtrips) ... so not sure yet of the future of this.

michaelneale:
ok remaining challenge: the flow will provide config by storing it straight to disk (in core goose crate, which is good practice, but GUI likes to duplicate things so still trying to work out how to make that work).

PR #3510, opened 2025-07-18
Rust debug

DOsinga:
This makes it possible to launch the frontend after you launch the backend on port 3000 - it will find that backend and use it. This way if you have a debugger running on the backend from VS Code or RustRover, you can easily set breakpoints and inspect values etc. This seems easier than attaching to a running process


ui/desktop/src/main.ts
```diff
@@ -2028,11 +2028,11 @@ app.whenReady().then(async () => {
...
-  - id: slack
-    command: uvx mcp_slack
-  - id: knowledge_graph_memory
-    command: npx -y @modelcontextprotocol/server-memory
-  ```
+ - id: slack
+ command: uvx mcp_slack
+ - id: knowledge_graph_memory
+ command: npx -y @modelcontextprotocol/server-memory
+ ```
```
jamadeo, line 2069: You probably didnt' mean to remove this leading whitespace

ui/desktop/src/main.ts
```diff
@@ -452,7 +452,7 @@ const getGooseProvider = () => {
 };
 
 const generateSecretKey = () => {
-  const key = crypto.randomBytes(32).toString('hex');
+  const key = process.env.GOOSE_EXTERNAL_BACKEND ? 'test' : crypto.randomBytes(32).toString('hex');
```
jamadeo, line 455: maybe here you could just check if GOOSE_SERVER__SECRET_KEY is already set, and if so, use that instead of generating one. Then in the Justfile set it

ui/desktop/src/goosed.ts
```diff
@@ -59,8 +59,46 @@ const checkServerStatus = async (
...
+    if (isNaN(parsedPid)) {
+      throw new Error('Could not parse PID from lsof output');
+    }
+    pid = parsedPid;
+  } catch (error) {
+    throw new Error(`Could not find process for external backend on port ${port}: ${error}`);
+  }
+
+  const mockProcess = {
+    pid,
```
jamadeo: do we actually need the pid of the backend if we aren't managing the process? maybe it isn't worth unwinding but it would let you skip the whole lsof thing

ui/desktop/src/main.ts
DOsinga: yes, but then you put the burden on the user to set that in their RustRover started backend to the same value - that thing currently sets it to test, so this is lower maintaince maybe?

ui/desktop/src/goosed.ts
DOsinga: you mean we make the pid nullable and then have ts tell us where we are relying on this? that might be better

ui/desktop/src/main.ts
DOsinga: hmm, that's weird, it looks like my autoformatter is doing that

jamadeo:
What I meant was since goosed defaults to “test”, you could just use
GOOSE_SECRET_KEY=test when staring the front end instead of a new
environment variable

DOsinga:
> What I meant was since goosed defaults to “test”, you could just use GOOSE_SECRET_KEY=test when staring the front end instead of a new environment variable

Ah, I see. That could work. Feels a bit magic maybe?

ui/desktop/src/goosed.ts
DOsinga: done

PR #3511, opened 2025-07-18
Revert "fix the output not being visible issue (#3491)"

alexhancock:
This reverts commit 079a64c1f7eebc4d0c21c07c8faedab38b1d0841.

Potential revert to address issue with invisible user messages

### DIFF
--- a/ui/desktop/src/components/sessions/SessionHistoryView.tsx
+++ b/ui/desktop/src/components/sessions/SessionHistoryView.tsx
@@ -43,8 +43,29 @@ const isUserMessage = (message: Message): boolean => {
   return true;
 };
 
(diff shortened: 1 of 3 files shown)


ui/desktop/src/components/sessions/SessionHistoryView.tsx
```diff
@@ -43,8 +43,29 @@ const isUserMessage = (message: Message): boolean => {
   return true;
 };
 
+// Filter messages for display (same logic as useChatEngine)
 const filterMessagesForDisplay = (messages: Message[]): Message[] => {
-  return messages.filter((message) => message.display);
```
jamadeo, line 47: Guessing this was the problem -- maybe we don't set an explicit flag for user messages?

PR #3513, opened 2025-07-18
Fix working directory when session has no messages

s-soroosh:
- Fixes https://github.com/block/goose/issues/2790

Previously, when a session had no messages or an invalid working directory, it would default to the home directory. This changes the behavior to use the current working directory instead, which I think is more intuitive and matches user expectations.

**Questions**
- Is there any drawback to use cwd as the default working directory instead of home directory?

**Step to reproduce the problem**:

* Go to any folder
* call goose to create a new session
* Quit goose
* Resume the session
* Observe the warning  WARNING: The original working directory of this session was set to "your home directory"

**With this Fix**
Next time you resume the session created "after applying this fix" will be associated with the correct working directory even if session is empty.

### DIFF
--- a/crates/goose/src/session/storage.rs
+++ b/crates/goose/src/session/storage.rs
@@ -31,6 +31,12 @@ fn get_home_dir() -> PathBuf {
         .to_path_buf()
 }
 
+fn get_current_working_dir() -> PathBuf {
+    std::env::current_dir()
+        .or_else(|_| Ok::<PathBuf, io::Error>(get_home_dir()))
+        .expect("could not determine the current working directory")
+}
+
 /// Metadata for a session, stored as the first line in the session file
 #[derive(Debug, Clone, Serialize, ToSchema)]
 pub struct SessionMetadata {
@@ -86,7 +92,7 @@ impl<'de> Deserialize<'de> for SessionMetadata {
         let working_dir = helper
             .working_dir
             .filter(|path| path.exists())
-            .unwrap_or_else(get_home_dir);
+            .unwrap_or_else(get_current_working_dir);
 
         Ok(SessionMetadata {
             description: helper.description,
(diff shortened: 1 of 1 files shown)


s-soroosh:
@michaelneale In the description I outlined how the issue can be reproduced.

taniacryptid:
Thank you for outlining the issue and how to reproduce it! For reference, this is happening for users on Mac and Linux https://github.com/block/goose/issues/2790

crates/goose/src/session/storage.rs
```diff
@@ -31,6 +31,12 @@ fn get_home_dir() -> PathBuf {
         .to_path_buf()
 }
 
+fn get_current_working_dir() -> PathBuf {
+    std::env::current_dir()
+        .or_else(|_| Ok::<PathBuf, io::Error>(get_home_dir()))
+        .expect("could not determine the current working directory")
+}
```
DOsinga, line 38: excellent. I wonder if we should avoid panicking here though - either return a default "", or bubble up?

fn get_current_working_dir() -> Result<PathBuf, io::Error> {
    std::env::current_dir().or_else(|_| Ok(get_home_dir()))
}

DOsinga (approved):

crates/goose/src/session/storage.rs
michaelneale: yeah - in other cases if we can't find a cwd then default to home (I think that may only show up in rare testing scenarios though). Things will work if you run from home (as it is usually smart enough to then know what directory to really work in from context later on) - so seems reasonable to default instead of panic?

s-soroosh: Here we default to home dir and then panic. Isn't this the same behaviour as you outlined? 
Note that `get_home_dir` itself also panic if it can't find the home dir. So I think the only question is whether want want to show a specific error here that cwd couldn't be determined, or home directory couldn't be determined.

PR #3516, opened 2025-07-18
Env var to set Ollama request timeout

bickelj:
Without this change, even if one recompiles `ollama` to extend timeouts there, `goose` times out HTTP requests to `ollama` at ten minutes.

Those of us who choose to run LLMs on CPUs, despite the slowness, profit from LLMs doing asynchronous work. This change allows users to increase the timeout `goose` sets on HTTP requests to `ollama`.

### DIFF
--- a/crates/goose/src/providers/ollama.rs
+++ b/crates/goose/src/providers/ollama.rs
@@ -13,6 +13,7 @@ use std::time::Duration;
 use url::Url;
 
 pub const OLLAMA_HOST: &str = "localhost";
+pub const OLLAMA_TIMEOUT: u64 = 600; // seconds
 pub const OLLAMA_DEFAULT_PORT: u16 = 11434;
 pub const OLLAMA_DEFAULT_MODEL: &str = "qwen2.5";
 // Ollama can run many models, we only provide the default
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/ollama.rs
```diff
@@ -109,6 +114,12 @@ impl Provider for OllamaProvider {
                 true,
                 false,
                 Some(OLLAMA_HOST),
+            ),
+            ConfigKey::new(
```
bickelj: TBH I don't know if this new `ConfigKey` is needed or what it even does.

michaelneale: it lets you set it in the config.yaml I believe in the .config/goose dir

michaelneale: (so doesn't have to be hidden away in env var if not needed, so all good)

michaelneale:
@bickelj thanks for this - can you run `cargo fmt` just to tidy it up again, but this LGTM - yeah it can be slow and sometimes that is ok!

taniacryptid:
> @bickelj thanks for this - can you run `cargo fmt` just to tidy it up again, but this LGTM - yeah it can be slow and sometimes that is ok!

Let us know when you've been able to run this, @bickelj !

bickelj:
@michaelneale @taniashiba After freeing some space (apparently I need 55GiB free to run `cargo build`), I was able to run `cargo check && cargo fmt && cargo clippy`. I admit not all tests passed locally. I hope it is simply because I didn't have llama3.2 locally. One of the reasons I run `ollama` locally is to avoid proprietary models such as llama.

bickelj:
By the way, thank you for `goose` and for your consideration of this PR!

michaelneale:
@bickelj no worries - will run tests here (I rarely run them all locally), taking a look now

PR #3530, opened 2025-07-20
Restore recipe parameters functionality

zanesq:
Recipe parameters stopped working after the new ui launch.

This adds them back and fixes some issues I noticed along the way:

1. Navigating away from an active recipe in chat and back to chat now keeps your chat showing rather than trying to reload the recipe from scratch.
2. Clicking on working directory or other buttons in the chat area no longer auto submits the chat form.

Bonus: 
- Added a few more parameter form fields while I was in there `select`, `boolean` and `number`.
- Added the remaining recipe fields in the preview modal with a deeplink for easy sharing.

Screenshot 2025-07-19 at 5 48 51 PM
Screenshot 2025-07-19 at 5 49 39 PM
Screenshot 2025-07-19 at 5 50 10 PM

image
image


ui/desktop/openapi.json
```diff
@@ -1827,6 +1827,11 @@
             "format": "double",
             "description": "Cost per token for output (optional)",
             "nullable": true
+          },
```
zanesq: upstream changes can ignore

ui/desktop/src/api/types.gen.ts
```diff
@@ -273,6 +273,10 @@ export type ModelInfo = {
      * Cost per token for output (optional)
      */
     output_token_cost?: number | null;
+    /**
```
zanesq: upstream changes can ignore

ui/desktop/src/components/ChatInput.tsx
```diff
@@ -896,373 +896,375 @@ export default function ChatInput({
...
-                    <span className="inline-block w-2 h-2 bg-red-500 rounded-full animate-pulse" />
-                    {Math.floor(recordingDuration)}s • ~{estimatedSize.toFixed(1)}MB
-                    {estimatedSize > 20 && <span className="text-xs">(near 25MB limit)</span>}
-                  </span>
-                )}
-              </div>
-            )}
-          </div>
+      {/* Input row with inline action buttons wrapped in form */}
+      <form onSubmit={onFormSubmit} className="relative flex items-end">
```
zanesq, line 900: `form` now wraps only what it should, the input and the submit/action buttons, before it was wrapping all the actions under the chat input causing the chat form to submit from descendant buttons in launched modals or the working directory click

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -421,29 +431,6 @@ function BaseChatContent({
                             {error.message || 'Honk! Goose experienced an error while responding'}
                           </div>
 
-                          {/* Expandable Error Details */}
```
zanesq, line 430: Removed this expandable error details area as it wasn't providing any other useful context than what is displayed on the screen.

ui/desktop/src/api/types.gen.ts
DOsinga: should we maybe have something in CI that checks whether regenerating the openapi.json is a no-op?

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -200,16 +201,25 @@ function BaseChatContent({
...
     if (recipeConfig?.title !== currentRecipeTitle) {
+      const previousTitle = currentRecipeTitle;
       setCurrentRecipeTitle(recipeConfig?.title || null);
-      setHasStartedUsingRecipe(false);
 
-      // Clear existing messages when a new recipe is loaded
-      if (recipeConfig?.title && recipeConfig.title !== currentRecipeTitle) {
+      // Only reset usage if we're switching between different recipes
+      // Don't reset if we're going from no recipe to a recipe (initial load)
+      // or if we already have messages (ongoing conversation)
```
DOsinga: I think we have this comment because the code is hard to follow; let's see if we can fix that and drop the comment instead

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -200,16 +201,25 @@ function BaseChatContent({
   // Reset recipe usage tracking when recipe changes
   useEffect(() => {
     if (recipeConfig?.title !== currentRecipeTitle) {
+      const previousTitle = currentRecipeTitle;
```
DOsinga: suggest to move this outside the if, so the comparisons are easier to follow

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -200,16 +201,25 @@ function BaseChatContent({
...
       setCurrentRecipeTitle(recipeConfig?.title || null);
-      setHasStartedUsingRecipe(false);
 
-      // Clear existing messages when a new recipe is loaded
-      if (recipeConfig?.title && recipeConfig.title !== currentRecipeTitle) {
+      // Only reset usage if we're switching between different recipes
+      // Don't reset if we're going from no recipe to a recipe (initial load)
+      // or if we already have messages (ongoing conversation)
+      if (previousTitle && recipeConfig?.title && previousTitle !== recipeConfig.title) {
```
DOsinga: here and below, we already checked previousTitle != recipeConfig.title so that bit is always going to be true. So for this to not be true, either previousTitle or recipeConfig?title needs to be falsey. if that's the case we should check on that

ui/desktop/src/components/ParameterInputModal.tsx
```diff
@@ -21,8 +21,33 @@ const ParameterInputModal: React.FC<ParameterInputModalProps> = ({
   useEffect(() => {
     const initialValues: Record<string, string> = {};
     parameters.forEach((param) => {
-      if (param.default) {
-        initialValues[param.key] = param.default;
+      // Set default value for optional parameters if they have a default value
```
DOsinga: I don't think we need the comment here

ui/desktop/src/components/ParameterInputModal.tsx
```diff
@@ -21,8 +21,33 @@ const ParameterInputModal: React.FC<ParameterInputModalProps> = ({
...
+      if (param.requirement === 'optional' && param.default) {
+        // Validate default value based on parameter type
+        let defaultValue = param.default;
```
DOsinga: hmm, we set the defaultValue here and then check a bunch of things where if we like the values, we set defaultValue to mostly the same value. 

in general we should probably set the validity of the recipe when we read it, not at this point. but here I think we can just leave out everything and just go
```
if (param.requirement === 'optional' && param.default) {
    initialValues[param.key] = param.default
}
```
possibly lowercasing it if param.input_type == 'boolean'

ui/desktop/src/components/ViewRecipeModal.tsx
```diff
@@ -73,7 +71,8 @@ export default function ViewRecipeModal({ isOpen, onClose, config }: ViewRecipeM
       setInstructions(config.instructions || '');
       setPrompt(config.prompt || '');
       setActivities(config.activities || []);
-      setParameters(parseParametersFromInstructions(config.instructions || ''));
+      // Use the actual parameters from the config, not parsed from instructions
+      setParameters(config.parameters || []);
```
DOsinga, line 66: remove this comment, it explains the code change, but is not needed for the reader

ui/desktop/src/components/ViewRecipeModal.tsx
```diff
@@ -102,28 +101,34 @@ export default function ViewRecipeModal({ isOpen, onClose, config }: ViewRecipeM
     }
   }, [isOpen, getExtensions, recipeExtensions, extensionsLoaded]);
 
-  // Use effect to set parameters whenever instructions or prompt changes
+  // Only auto-parse parameters from instructions if no parameters exist yet
+  // This prevents overwriting existing parameter configurations
```
DOsinga: I'm not sure I am following the code here, but does this mean once you set parameters, we no longer update the parameters when the prompt changes

ui/desktop/src/components/ViewRecipeModal.tsx
```diff
@@ -102,28 +101,34 @@ export default function ViewRecipeModal({ isOpen, onClose, config }: ViewRecipeM
...
+      const instructionsParams = parseParametersFromInstructions(instructions);
+      const promptParams = parseParametersFromInstructions(prompt);
+
+      // Combine parameters, ensuring no duplicates by key
+      const allParams = [...instructionsParams];
+      promptParams.forEach((promptParam) => {
+        if (!allParams.some((param) => param.key === promptParam.key)) {
+          allParams.push(promptParam);
+        }
+      });
```
DOsinga: this is O(n^2) - consider using a map to do this more efficiently

ui/desktop/src/components/ViewRecipeModal.tsx
```diff
@@ -102,28 +101,34 @@ export default function ViewRecipeModal({ isOpen, onClose, config }: ViewRecipeM
...
+    }
+  }, [instructions, prompt, parameters.length]);
 
   const getCurrentConfig = (): Recipe => {
     // Transform the internal parameters state into the desired output format.
     const formattedParameters = parameters.map((param) => {
       const formattedParam: Parameter = {
         key: param.key,
-        input_type: 'string',
+        input_type: param.input_type || 'string', // Use actual input_type instead of hardcoded 'string'
```
DOsinga: remove the comment

ui/desktop/src/hooks/useRecipeManager.ts
```diff
@@ -101,6 +105,11 @@ export const useRecipeManager = (messages: Message[], locationState?: LocationSt
 
   // Pre-fill input with recipe prompt instead of auto-sending it
   const initialPrompt = useMemo(() => {
+    // Don't show initial prompt if we already have messages (ongoing conversation)
```
DOsinga: are we calling this in the wrong place? the initial prompt shouldn't change because we have a conversation, we should just not put at as the non-initial prompt

ui/desktop/src/hooks/useRecipeManager.ts
```diff
@@ -117,16 +126,19 @@ export const useRecipeManager = (messages: Message[], locationState?: LocationSt
...
     return '';
-  }, [recipeConfig, recipeParameters]);
+  }, [recipeConfig, recipeParameters, messages.length]);
 
   // Handle parameter submission
   const handleParameterSubmit = async (inputValues: Record<string, string>) => {
-    setRecipeParameters(inputValues);
+    // Store parameters in chat context instead of local state
+    if (chatContext?.setRecipeParameters) {
+      chatContext.setRecipeParameters(inputValues);
```
DOsinga, line 154: in what case is chatContext undefined here? can we use the type system to make sure it is not?

ui/desktop/src/recipe/index.ts
```diff
@@ -26,6 +27,33 @@ export interface Recipe {
...
+              };
+            }
+          >;
+          required?: string[];
+        };
+      }
+    | string; // Can also be a string representation
   // Properties added for scheduled execution
   scheduledJobId?: string;
   isScheduledExecution?: boolean;
```
DOsinga: Not for now, but we have this definition also on the server. for some reason we're not using the openapi mechanism to do recipe calls, so we duplicate the definition here. do you think we can fix that?

DOsinga (approved):
Thanks for jumping on this. I added some maybe too detailed review for a quick fix. In general my question is also, what can we do to test this automatically? these flows are hard to follow I find and seem like having bugs in them would be easy

ui/desktop/src/api/types.gen.ts
zanesq: Not sure if ci can help this other than maybe flagging when someone didn't manually generate it? It should be manually generated whenever someone makes a related change. I run the full `npm run start` locally which generates the types so occasionally I run into this issue where someone didn't generate the types upstream.

DOsinga: yeah, I meant we could run the typegen in ci and compare if it changes anything and if it does, tell the user they forgot to run it but did make a change

zanesq: Ah gotcha, good idea! Will follow up with that 👍

ui/desktop/src/hooks/useRecipeManager.ts
zanesq: its a provider that we only use for certain cases at runtime so the type system isn't enough to catch it unfortunately

ui/desktop/src/recipe/index.ts
zanesq: Good call, went ahead and added it to the openai types

ui/desktop/src/hooks/useRecipeManager.ts
zanesq: Good catch, fixed

ui/desktop/src/components/ViewRecipeModal.tsx
zanesq: Good catch, fixed

ui/desktop/src/components/ViewRecipeModal.tsx
zanesq: Good catch, fixed to be a bit more clear and also updated code to update the parameters when the prompt changes

ui/desktop/src/components/ParameterInputModal.tsx
zanesq: Good catch, fixed

ui/desktop/src/components/BaseChat.tsx
zanesq: Good catch, fixed

ui/desktop/src/components/BaseChat.tsx
zanesq: Good call refactored and simplified

zanesq:
@DOsinga great review thanks! All issues addressed pls take another look.

Regarding how we can verify, unfortunately until there are more functional/ui tests covering this process its a manual testing process. I'll make sure we cover recipes in the tests as we build them out.

ui/desktop/src/components/pair.tsx
```diff
@@ -43,6 +43,7 @@ export interface ChatType {
   messageHistoryIndex: number;
   messages: Message[];
   recipeConfig?: Recipe | null; // Add recipe configuration to chat state
+  recipeParameters?: Record<string, string> | null; // Add recipe parameters to chat state
```
DOsinga: why are hub and pair having the same diff & code here?

DOsinga:
hmm, weird error. can you try merging in main?

ui/desktop/src/components/pair.tsx
zanesq: Good catch also, refactored to use a shared ChatType instead 👍

PR #3535, opened 2025-07-20
Expand databricks agent error handling to fallback to token limit errors

zanesq:
I noticed the UI shows a generic 401 unauthorized (occasionally) or 400 bad request from Databricks when a token limit is reached instead of kicking off the auto summarization. It doesn't recognize that its a token limit error because the message doesn't have the phrases we look for like "token limit" in the error message. It appears to be a token limit issue because manually using the summarize feature allows chat to continue while just sending a chat message without summarization does not.

This adds fallback logic to work around that and allows the UI to handle it gracefully by allowing chat to continue with auto summarize.

Screenshot 2025-07-19 at 7 17 26 PM


DOsinga:
thanks for looking into this. I don't think we should be adding more ad hoc checks like this to providers though. my approach is this: https://github.com/block/goose/pull/3558 -- add universal error parsing and retry logic to all providers.

zanesq:
Makes sense, I'll close for now. Lets see if [this](https://github.com/block/goose/commit/f0b4f7407170506383b2b8f15e64826870fad6f6) fixes the issue and we might not need this.

PR #3539, opened 2025-07-20
Fix OpenAI tool calling JSON parsing and error handling

cbruyndoncx:
- Improves OpenAI tool calling JSON parsing to handle concatenated JSON objects
- Replaces panic with graceful error handling for empty message content
- Adds comprehensive logging for debugging tool call parsing issues
- Validates message ordering for OpenAI format compliance
- Properly serializes tool call arguments to prevent malformed JSON

Signed-off-by: Carine Bruyndoncx <bruyndoncx@gmail.com>

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -1319,7 +1319,16 @@ impl Session {
                                 self.debug,
                             );
                         }
-                        None => panic!("No content in last message"),
+                        None => {
+                            // Handle empty message content gracefully
+                            tracing::warn!("Found user message with empty content, removing it");
+                            self.messages.pop();
+                            let prompt = "Removed an empty message and interrupted before the model replied.";
+                            output::render_message(
+                                &Message::assistant().with_text(prompt),
+                                self.debug,
+                            );
+                        }
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


taniacryptid:
Thank you for the draft PR, Carine!

salman1993:
Thanks for the PR! I'm a bit unclear on the problem this is solving - could you add a failing test case to show the issue? That might help guide the fix.

Also, this introduces quite a bit of parsing logic, which feels heavy. For example, `arguments.contains("}{")` seems like a workaround—what’s the specific case it handles?

Appreciate the contribution - trying to to better understand the fix here.

cbruyndoncx:
[PR-3539-Done.json](https://github.com/user-attachments/files/21374570/PR-3539-Done.json)
[PR-3539-NotDone-v113.json](https://github.com/user-attachments/files/21374571/PR-3539-NotDone-v113.json)

Had a little trouble getting test data, but the NotDone version shows the logging, and the other one shows it working after the changes. 
I agree, parsing does not look good. So if you have tips on how to improve, let me know and i can test. It basically happens if you ask goose to do something where it is doing multiple tool calls in the background without output on screen. 
At least that is the situation where it failed in my case.
It does work with openrouter openai, so you really need openai api keys for testing

salman1993:
In the current codebase, the test would be in https://github.com/block/goose/blob/main/crates/goose/tests/providers.rs#L117

We have a test for simple tool usage but you can add a test using your json files for multiple tool usage and see if you replicate the error.

I know @DOsinga is currently refactoring the provider tests. If you have any ideas to share or we can also add this scenario to your test suite?

cbruyndoncx:
v1.1.4 has fixed this issue. It looks like i no longer need my customizations. Closing the PR.

PR #3548, opened 2025-07-21
Use openapi for recipe endpoint types and in frontend

jsibbison-square:
This PR is primarily focused on converting the /recipes endpoints to use openapi schema. This includes converting the frontend types to use the same types as the the api interface. Unfortunately I couldn't go all the way in converting in one go due to how the frontend has bundled a bunch of extra config into recipes (particularly for schedules) and the PR was starting to get to big. So this PR converts the api and some frontend types but TODOs are look for more required refactoring.


ui/desktop/src/components/RecipesView.tsx
```diff
@@ -163,8 +163,12 @@ export default function RecipesView({ _onLoadRecipe }: RecipesViewProps = {}) {
...
       const recipe = await decodeRecipe(recipeEncoded);
 
-      if (!recipe.title || !recipe.description || !recipe.instructions) {
-        throw new Error('Recipe is missing required fields (title, description, instructions)');
+      if (!recipe.title || !recipe.description) {
+        throw new Error('Recipe is missing required fields (title, description)');
+      }
+
+      if (!recipe.instructions && !recipe.prompt) {
+        throw new Error('Recipe must have either instructions or prompt');
```
DOsinga, line 171: So nice

ui/desktop/src/components/schedule/CreateScheduleModal.tsx
```diff
@@ -211,7 +216,7 @@ function recipeToYaml(recipe: Recipe, executionMode: ExecutionMode): string {
       }
 
       // Add common optional fields
-      if (ext.env_keys && ext.env_keys.length > 0) {
+      if ('env_keys' in ext && ext.env_keys && ext.env_keys.length > 0) {
```
DOsinga, line 219: if (ext?.env_keys?.length) {

DOsinga (approved):
thanks for doing this. this is so much better

ui/desktop/src/components/schedule/CreateScheduleModal.tsx
jsibbison-square: types seem to be very unhappy with that approach for reasons that make no sense to me. :(

ui/desktop/src/recipe/index.ts
```diff
@@ -43,98 +40,97 @@ export interface CreateRecipeRequest {
...
+  // TODO: Fix this type to match Message OpenAPI spec
+  return {
+    id: frontendMessage.id,
+    role: frontendMessage.role as Role,
+    content: frontendMessage.content.map((content) => ({
+      ...content,
+      // Convert toolCall to match API expectations
+      ...(content.type === 'toolRequest' && 'toolCall' in content
+        ? {
+            toolCall: content.toolCall as unknown as { [key: string]: unknown },
```
cloud-on-prem, line 55: Possibly a runtime risk with this conversion. Is there a way to determine the correct type?

PR #3554, opened 2025-07-21
Agent loop defensive

DOsinga:
I couldn't actually reproduce the session corruption easily, but my theory is that it can happen because of two reasons:

* when we interrupt the agent, we don't actually cancel the agent. we just hang up. the agent will continue running until at some point writing to the stream will fail. 
* we don't write the messages our atomically, but inline

the two combined can lead to us writing a message to the agent while the user has already canceled. If the user fires up a new command before say an MCP server has finished, this can create havoc. 

this PR introduces a cancellation token and makes the message appending atomic-ish. it also deletes the askAi handler since that is not used.

### DIFF
--- a/ui/desktop/openapi.json
+++ b/ui/desktop/openapi.json
@@ -2332,17 +2332,16 @@
         }
       },
       "ResourceContents": {
-        "oneOf": [
+        "anyOf": [
           {
             "type": "object",
             "required": [
-              "uri",
-              "text"
+              "text",
+              "uri"
             ],
             "properties": {
               "mime_type": {
-                "type": "string",
-                "nullable": true
+                "type": "string"
               },
               "text": {
                 "type": "string"
@@ -2355,16 +2354,15 @@
           {
             "type": "object",
             "required": [
-              "uri",
-              "blob"
+              "blob",
+              "uri"
             ],
             "properties": {
               "blob": {
                 "type": "string"
               },
               "mime_type": {
-                "type": "string",
-                "nullable": true
+                "type": "string"
               },
               "uri": {
                 "type": "string"
--- a/ui/desktop/src/api/types.gen.ts
+++ b/ui/desktop/src/api/types.gen.ts
@@ -463,12 +463,12 @@ export type RedactedThinkingContent = {
 };
 
 export type ResourceContents = {
-    mime_type?: string | null;
+    mime_type?: string;
     text: string;
     uri: string;
 } | {
     blob: string;
-    mime_type?: string | null;
+    mime_type?: string;
     uri: string;
 };
 
--- a/ui/desktop/src/components/BaseChat.tsx
+++ b/ui/desktop/src/components/BaseChat.tsx
@@ -537,7 +537,7 @@ function BaseChatContent({
         recipeDetails={{
           title: recipeConfig?.title,
           description: recipeConfig?.description,
-          instructions: recipeConfig?.instructions,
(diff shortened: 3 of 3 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -952,103 +931,104 @@ impl Agent {
...
+    fn determine_goose_mode(session: &Option<SessionConfig>, config: &Config) -> String {
+        match session
+            .as_ref()
+            .and_then(|s| s.execution_mode.as_ref().map(String::as_str))
+        {
+            Some("foreground") => "auto".to_string(),
+            Some("background") => "chat".to_string(),
+            _ => config
+                .get_param("GOOSE_MODE")
+                .unwrap_or_else(|_| "auto".to_string()),
```
jamadeo, line 1077: not related to your PR, but this mapping looks quite questionable. Why chat mode for background jobs? This looks like the root of a recent bug report

crates/goose/src/agents/agent.rs
```diff
@@ -749,45 +716,63 @@ impl Agent {
...
-                                ) {
-                                    // Emit as McpNotification event
-                                    yield AgentEvent::McpNotification((
-                                        subagent_id.to_string(),
-                                        notification.clone()
-                                    ));
-                                }
+                    if let JsonRpcMessage::Notification(notif) = &notification {
+                        if let Some(data) = notif.params.as_ref().and_then(|p| p.get("data")) {
+                            if let (Some(subagent_id), Some(_message)) = (
```
jamadeo, line 779: Rust 2024 lets you chain these! https://github.com/block/goose/pull/3464

crates/goose/src/agents/agent.rs
```diff
@@ -749,45 +716,63 @@ impl Agent {
                     break;
                 }
 
-                // Check for MCP notifications from subagents
+                if let Some(token) = &cancel_token {
+                    if token.is_cancelled() {
```
jamadeo: can this be

`cancel_token.is_some_and(|token| token.is_cancelled())`

?

crates/goose-server/src/routes/reply.rs
```diff
@@ -538,8 +386,7 @@ async fn submit_tool_result(
 
 pub fn routes(state: Arc<AppState>) -> Router {
     Router::new()
-        .route("/reply", post(handler))
-        .route("/ask", post(ask_handler))
```
jamadeo, line 544: nice find

crates/goose/src/agents/agent.rs
DOsinga: the AIs say you need an `as_ref()` in there still

crates/goose/src/agents/agent.rs
DOsinga: @Kvadratni thoughts?

michaelneale:
seems sensible, and could prevent some other problems of processes hanging around.

I also still encounter hanging, and then unkillable sessions, wonder if this will help with that (vs trying to hang up)

crates/goose-server/src/routes/reply.rs
michaelneale: /ask is old - I thought /reply is still used though?

DOsinga: it is - I just renamed handler to reply_handler here. the code still exists!

PR #3558, opened 2025-07-21
Generic retry and error parsing

DOsinga:
This lifts the retry mechanism out of the databricks provider and also the error parsing so we can generally call it. it would be good to discuss this and merge it and then apply it to all the other things

### DIFF
--- a/crates/goose/src/providers/anthropic.rs
+++ b/crates/goose/src/providers/anthropic.rs
@@ -16,7 +16,7 @@ use super::errors::ProviderError;
 use super::formats::anthropic::{
     create_request, get_usage, response_to_message, response_to_streaming_message,
 };
-use super::utils::{emit_debug_trace, get_model};
+use super::utils::{emit_debug_trace, get_model, map_http_error_to_provider_error};
 use crate::message::Message;
 use crate::model::ModelConfig;
 use mcp_core::tool::Tool;
@@ -90,40 +90,35 @@ impl AnthropicProvider {
 
         let status = response.status();
         let payload: Option<Value> = response.json().await.ok();
+        Self::anthropic_api_call_result(status, payload)
+    }
 
+    fn anthropic_api_call_result(
+        status: StatusCode,
+        payload: Option<Value>,
+    ) -> Result<Value, ProviderError> {
         // https://docs.anthropic.com/en/api/errors
         match status {
-            StatusCode::OK => payload.ok_or_else( || ProviderError::RequestFailed("Response body is not valid JSON".to_string()) ),
-            StatusCode::UNAUTHORIZED | StatusCode::FORBIDDEN => {
-                Err(ProviderError::Authentication(format!("Authentication failed. Please ensure your API keys are valid and have the required permissions. \
-                    Status: {}. Response: {:?}", status, payload)))
-            }
-            StatusCode::BAD_REQUEST => {
-                let mut error_msg = "Unknown error".to_string();
-                if let Some(payload) = &payload {
-                    if let Some(error) = payload.get("error") {
-                    tracing::debug!("Bad Request Error: {error:?}");
-                    error_msg = error.get("message").and_then(|m| m.as_str()).unwrap_or("Unknown error").to_string();
-                    if error_msg.to_lowercase().contains("too long") || error_msg.to_lowercase().contains("too many") {
-                        return Err(ProviderError::ContextLengthExceeded(error_msg.to_string()));
-                    }
-                }}
-                tracing::debug!(
-                    "{}", format!("Provider request failed with status: {}. Payload: {:?}", status, payload)
-                );
-                Err(ProviderError::RequestFailed(format!("Request failed with status: {}. Message: {}", status, error_msg)))
-            }
-            StatusCode::TOO_MANY_REQUESTS => {
-                Err(ProviderError::RateLimitExceeded(format!("{:?}", payload)))
-            }
-            StatusCode::INTERNAL_SERVER_ERROR | StatusCode::SERVICE_UNAVAILABLE => {
-                Err(ProviderError::ServerError(format!("{:?}", payload)))
-            }
+            StatusCode::OK => payload.ok_or_else(|| {
+                ProviderError::RequestFailed("Response body is not valid JSON".to_string())
+            }),
             _ => {
-                tracing::debug!(
-                    "{}", format!("Provider request failed with status: {}. Payload: {:?}", status, payload)
-                );
-                Err(ProviderError::RequestFailed(format!("Request failed with status: {}", status)))
+                if status == StatusCode::BAD_REQUEST {
+                    if let Some(error_msg) = payload
+                        .as_ref()
+                        .and_then(|p| p.get("error"))
+                        .and_then(|e| e.get("message"))
+                        .and_then(|m| m.as_str())
+                    {
+                        let msg = error_msg.to_string();
+                        if msg.to_lowercase().contains("too long")
+                            || msg.to_lowercase().contains("too many")
+                        {
+                            return Err(ProviderError::ContextLengthExceeded(msg));
+                        }
+                    }
+                }
+                Err(map_http_error_to_provider_error(status, payload))
             }
         }
     }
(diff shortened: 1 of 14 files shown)


crates/goose/src/providers/anthropic.rs
```diff
@@ -217,7 +212,7 @@ impl Provider for AnthropicProvider {
     }
 
     /// Fetch supported models from Anthropic; returns Err on failure, Ok(None) if not present
-    async fn fetch_supported_models_async(&self) -> Result<Option<Vec<String>>, ProviderError> {
+    async fn fetch_supported_models(&self) -> Result<Option<Vec<String>>, ProviderError> {
```
jamadeo, line 187: thank you for renaming this

michaelneale:
seems sensible approach - if we go with this - can we also close down https://github.com/block/goose/pull/3194 and https://github.com/block/goose/pull/3547 (which is my update of it) - as this does part of that (not sure if you want any of the other stuff from there)

DOsinga:
> seems sensible approach - if we go with this - can we also close down #3194 and #3547 (which is my update of it) - as this does part of that (not sure if you want any of the other stuff from there)

sorry, hadn't seen your version of this, just the the jackjack - I like the configbuilder thing from yours, I'll add that in a follow up. mostly what stops me from getting this in, is the testing

DOsinga:
this is ready for review @michaelneale - I incorporated the client logic from the other PR. I left out connection pooling as the LLMs tell me we already had that

crates/goose-server/src/routes/agent.rs
```diff
@@ -234,29 +236,26 @@ async fn update_agent_provider(
...
-    let new_provider = create(&payload.provider, model_config).unwrap();
+    let model = match payload
+        .model
+        .or_else(|| config.get_param("GOOSE_MODEL").ok())
+    {
+        Some(m) => m,
+        None => return Err(StatusCode::BAD_REQUEST),
+    };
+
+    let model_config = ModelConfig::new(&model).map_err(|_| StatusCode::BAD_REQUEST)?;
```
jamadeo, line 255: would be nice if we could include text with these, but maybe leave that for another change

crates/goose/src/providers/api_client.rs
```diff
@@ -0,0 +1,225 @@
...
+pub enum AuthMethod {
+    BearerToken(String),
+    ApiKey {
+        header_name: String,
+        key: String,
+    },
+    #[allow(dead_code)]
+    OAuth(OAuthConfig),
+    Custom(Box<dyn AuthProvider>),
+}
```
jamadeo, line 28: nice

crates/goose/src/providers/api_client.rs
```diff
@@ -0,0 +1,225 @@
...
+
+pub struct ApiRequestBuilder<'a> {
+    client: &'a ApiClient,
+    path: &'a str,
+    headers: HeaderMap,
+}
+
+impl ApiClient {
+    pub fn new(host: String, auth: AuthMethod) -> Result<Self> {
+        Self::with_timeout(host, auth, Duration::from_secs(600))
```
jamadeo, line 78: `const DEFAULT_TIMEOUT = ... `

crates/goose/src/providers/api_client.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    pub fn with_header(mut self, key: &str, value: &str) -> Result<Self> {
+        let header_name = HeaderName::from_bytes(key.as_bytes())?;
+        let header_value = HeaderValue::from_str(value)?;
+        self.default_headers.insert(header_name, header_value);
+        self.client = Client::builder()
+            .timeout(self.timeout)
+            .default_headers(self.default_headers.clone())
+            .build()?;
```
jamadeo, line 107: these .with_ methods look like they could be chained because they take a reference but they make a new client every time -- maybe a builder would be better? .with_timeout().with_headers() won't do the expected thing right?

crates/goose/src/providers/api_client.rs
```diff
@@ -0,0 +1,225 @@
...
+
+impl ApiResponse {
+    pub async fn from_response(response: Response) -> Result<Self> {
+        let status = response.status();
+        let payload = response.json().await.ok();
+        Ok(Self { status, payload })
+    }
+}
+
+pub struct ApiRequestBuilder<'a> {
```
jamadeo, line 70: nit: you could just use an owned String for path and drop the lifetime here

DOsinga:
> Looks like a great improvement -- how do we satisfactorily test?
> 
> I think there is a subtle bug in your ApiClient .with_ methods though

I don't know how we satisfactory test this other than making sure we have keys for all providers and then have scenario tests for all of them or some such.

crates/goose-server/src/routes/agent.rs
DOsinga: I do agree - I didn't quickly see how that works though. there's a whole bunch of other handlers that would be helped

crates/goose/src/providers/api_client.rs
DOsinga: thanks

crates/goose/src/providers/api_client.rs
DOsinga: done

crates/goose/src/providers/api_client.rs
DOsinga: This is a bit clunky and maybe a builder pattern would be better, but since these methods also set the timeout and the headers on themselves it should work, no?

jamadeo: sorry yes you're right, I think I got mixed up because with_timeout is not a method after all

michaelneale:
Aside: how well does the retry work in databricks? I haven't found it working for me lately (well I keep hitting things which I thought were retry-able, upstream errors etc)?

crates/goose/src/providers/anthropic.rs
```diff
@@ -128,24 +123,17 @@ impl AnthropicProvider {
...
+            .map(|&model_name| ModelInfo::new(model_name, 200_000))
+            .collect();
+
         ProviderMetadata::with_models(
             "anthropic",
             "Anthropic",
             "Claude and other models from Anthropic",
             ANTHROPIC_DEFAULT_MODEL,
-            vec![
-                ModelInfo::new("claude-sonnet-4-0", 200000),
```
michaelneale, line 137: where do these live now - not used as it is always loaded?

michaelneale: note: these will be in a follow up for per provider custom stuff

crates/goose-cli/src/session/output.rs
```diff
@@ -491,6 +491,23 @@ fn get_tool_params_max_length() -> usize {
         .unwrap_or(40)
 }
 
+fn print_value(value: &Value, debug: bool) {
+    let formatted = match value {
+        Value::String(s) => {
+            if !debug && s.len() > get_tool_params_max_length() {
```
michaelneale, line 497: this shows up as:

image

is that intended when not debug?

michaelneale (approved):
I think this is ok - I did test it with :

* anthropic/opus
* databricks/sonnet
* openrouter/qwen3-coder

and all seemed to work well (but didnt' test other providers). 

not sure what this takes away (ie would need follow on for anthropic provider models?)

and one comment about the big bold yellow `REDACTED` but seems nice and the direction we want it to go in.

crates/goose/src/providers/anthropic.rs
DOsinga: previously we declared ANTHROPIC_KNOWN_MODELS but were not using that at all. I now drag it in and just annotate it with 200k.

but yeah, the true way is for all providers to declare their models and token limits and then in models.rs read that to estimate limits instead of doing it twice or three times as we do now. small steps

crates/goose-cli/src/session/output.rs
DOsinga: I don't know - I didn't really touch this code, the new linter rules force me to break up this function

michaelneale: ah - fair enough, yes refactoring makes reviewing diffs very difficult when it breaks up things (wish the diff was a bit smarter, and could point out the code previously existed - if you search for it you can see but easy to miss). Yeah makes sense, just odd I never saw this before.

PR #3559, opened 2025-07-21
WIP: New Summarization Algorithm

katzdave:
Still testing/cleaning up but putting up the branch if anyone wants to try it, already seems to work pretty well.


crates/goose/src/prompts/summarize_oneshot_old.md
```diff
@@ -0,0 +1,18 @@
+You are an expert at summarizing conversation histories for AI assistants. Your task is to create a concise but comprehensive summary that preserves all the essential information needed for future conversations.
```
michaelneale: is this an old artifact as it isn't an old file?

michaelneale:
nice - taking for a spin

michaelneale:
nice @katzdave - seems to work well, I will use it a bit more as a daily thing and see if I can abuse it more! lets get this in!

michaelneale:
did hit a snag, I told it to look at ~/.local/share/goose/sessions for jsonl to extract themes, and it ran up quite the context: 
 (a prompt like " can you take a look in ~/.local/share/goose/sessions at the jsonl files to extract some common themes")

```
Now let me examine the session file to understand the specific issue:
path: ~/.local/share/goose/sessions/20250722_151520.jsonl
command: view
view_range:
    -
        1
    -
        20

Error: Request failed: Request failed with status: 400 Bad Request. Message: {"external_model_provider":"databricks-model-serving","external_model_error":{"error_code":"BAD_REQUEST","message":"Bad Request"}}


Context: ●●●●●●●○○○ 68% (136646/200000 tokens)
```

might be able to reproduce it with things like that (jsonl is good as there is lots of stuff buried in there in single lines to mess it up), but then again, may not be related? I did ask it a few things and a few times to get it to run up a big session)

crates/goose/src/prompts/summarize_oneshot_old.md
katzdave: yeah removing

katzdave:
> did hit a snag, I told it to look at ~/.local/share/goose/sessions for jsonl to extract themes, and it ran up quite the context: (a prompt like " can you take a look in ~/.local/share/goose/sessions at the jsonl files to extract some common themes")
> 
> ```
> Now let me examine the session file to understand the specific issue:
> ─── text_editor | developer ──────────────────────────
> path: ~/.local/share/goose/sessions/20250722_151520.jsonl
> command: view
> view_range:
>     -
>         1
>     -
>         20
> 
> Error: Request failed: Request failed with status: 400 Bad Request. Message: {"external_model_provider":"databricks-model-serving","external_model_error":{"error_code":"BAD_REQUEST","message":"Bad Request"}}
> The tool calling loop was interrupted. How would you like to proceed?
>   error: The error above was an exception we were not able to handle.
> These errors are often related to connection or authentication
> We've removed the conversation up to the most recent user message
> - depending on the error you may be able to continue
> 
> 
> Context: ●●●●●●●○○○ 68% (136646/200000 tokens)
> ```
> 
> might be able to reproduce it with things like that (jsonl is good as there is lots of stuff buried in there in single lines to mess it up), but then again, may not be related? I did ask it a few things and a few times to get it to run up a big session)

I changed the context counting a bit, as it was annoying me that the context didn't update until the next message. I ripped that out into another branch though, and am trying to get things in a row there.

michaelneale:
more testing, mostly summarize was spot on, but I did hit this roadblock - no idea if related, but it just will never respond/do anything now: 

image

not sure if related to session or provider

this was after a compaction, and the next prompt after that (ie once it had summarized/compacted, returned, I gave it one more instruction and that worked, but then got into this state). Probably not related?

edit: 

config: 
```
GOOSE_MODEL=claude-opus-4-20250514
GOOSE_PROVIDER=anthropic
```

michaelneale:
@katzdave  this is my take on auto compact: https://github.com/block/goose/pull/3600

still trying it out - so not sure how good, but thought you may find it interesting where to put the calls to compact - client end still seems easier (vs moving to agent and changing how api looks) - but take a look if it helps. I have seen it autocompact once (but it seemed to stall before it hit it - so may not be quite right) - it did autocompact down to 1K tokens (should it be that low?)

eg: 

image

crates/goose-cli/src/commands/web.rs
```diff
@@ -455,6 +456,45 @@ async fn process_message_streaming(
         session_msgs.clone()
     };
 
+    // Check and compact messages if needed before calling reply
+    let compact_result = check_and_compact_messages(agent, &messages, None).await?;
+    if compact_result.compacted {
```
DOsinga: you are repeating this code 3 times (and not for scheduled tasks?) - can we just move this into agent.reply?

crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -0,0 +1,336 @@
...
+/// and automatically compacts the messages using the summarization algorithm if needed.
+///
+/// # Arguments
+/// * `agent` - The agent to use for context management
+/// * `messages` - The current message history
+/// * `threshold_override` - Optional threshold override (defaults to GOOSE_AUTO_COMPACT_THRESHOLD config)
+///
+/// # Returns
+/// * `AutoCompactResult` containing the potentially compacted messages and metadata
+pub async fn check_and_compact_messages(
```
DOsinga: I'd split this into two methods, one to check whether we need compaction and one to do the compaction - we could then later make the check return how many turns until compaction for example

crates/goose/src/prompts/summarize_oneshot.md
```diff
@@ -0,0 +1,67 @@
+## 📝 Summary Generation Instructions
```
DOsinga: I like this, external prompts!

crates/goose/src/context_mgmt/summarize.rs
```diff
@@ -1,14 +1,21 @@
 use super::common::{get_messages_token_counts, get_messages_token_counts_async};
```
DOsinga, line 1: didn't we just want to delete this? also we need to make a call on compaction vs summarize

full-bench-config.json
```diff
@@ -0,0 +1,91 @@
+{
```
DOsinga: where does this come from?

DOsinga (approved):

full-bench-config.json
katzdave: oops got written from running goose bench. I dropped the auto-summarize ratio to 5% to force it to trigger a lot  and it still was able to complete all the tasks.

Didn't try any comprehensive numerical benchmarks or use the judge against results without summarization.

crates/goose/src/context_mgmt/auto_compact.rs
katzdave: Done

crates/goose-cli/src/commands/web.rs
katzdave: Done

crates/goose/src/context_mgmt/summarize.rs
katzdave: Agree that we can probably delete the chunked version in theory given that the summary prompt is shorter than goose's main prompt but I'd rather keep it around until I can better benchmark what happens when sessions get to 100% since it's a bit more resilient if we go beyond a workable session size since it splits it into 3.

I think /summarize -> /compact is the way to go. Claude's /summarize command returns a summary and doesn't affect the prompt history but /compact mirrors our summarize.

katzdave:
Still having an issue with token_counts seeming to use accumulated token counts (or some other cause of massively over-estimating them), so after we hit the threshold once it basically always auto-compacts and the usage_ratio just climbs. Taking another look later.

michaelneale:
@katzdave looks like some conflicts around using the new rmcp things - won't merge that in as I expect you have unstaged changes

michaelneale:
@katzdave I have no idea what those failures are from, I don't get them locally, really am perplexed. There is an error from base in providers? that seem the same as main branch?

michaelneale:
weird - if you run the build from within crates you will see failure, I think it is due to ambiguity in resolving `Tool` due to `pub use common::*;` - wild

katzdave:
Thanks for that fix. Sorted out the token counting now; auto-compact wasn't properly removing all messages from the session.

Did a bunch of testing, seems pretty good now; resuming sessions, switching between cli and desktop.

Token counts in general seems like a mess. All sorts of different counting systems from different places off by subtle amounts (+ estimates from not having the tokenizer). When I clean up the UI in the next wave I want to consolidate these numbers to some shared functions as best I can too.

michaelneale:
@katzdave yeah - some (all) use data back from provider to update, but not sure that is always right (especially when streaming). Maybe some count other ways? wonder if we should have a single function to estimate it in a unifoirm way using toktoken or something like that, not the same as the "source of truth" from the providers, but could be an easy up front way to know instantly what token count is with some consistency enough for compacting https://crates.io/crates/tiktoken-rs

michaelneale:
yeah main thing remaining seems to be when it does compact - it doesn't have a clear picture of what you just asked. This may mean that it will say it has modified something, when it didn't (yet), or just act as if there is nothing to do (probably need to keep some of the last messages to replay? or just the last one? or maybe my branch was slightly out of date) but otherwise seems stable

katzdave:
> yeah main thing remaining seems to be when it does compact - it doesn't have a clear picture of what you just asked. This may mean that it will say it has modified something, when it didn't (yet), or just act as if there is nothing to do (probably need to keep some of the last messages to replay? or just the last one? or maybe my branch was slightly out of date) but otherwise seems stable

Fixed this! We were indeed compacting that message. It now gets pulled out of compaction, and pushed back onto the message chain when compaction is complete.

katzdave:
> @katzdave yeah - some (all) use data back from provider to update, but not sure that is always right (especially when streaming). Maybe some count other ways? wonder if we should have a single function to estimate it in a unifoirm way using toktoken or something like that, not the same as the "source of truth" from the providers, but could be an easy up front way to know instantly what token count is with some consistency enough for compacting https://crates.io/crates/tiktoken-rs

Makes sense. Going to follow up on this and see if I can get away with using the provider results in general; we should have those from the responses (including the summarization request). The only real quirk is that summarization responses won't have the system prompt, but some kind of consistent estimate there should be good enough.

michaelneale (approved):
seems very nice - hopefully this will massively improve performance and cost and reliability

crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -60,7 +60,7 @@ pub async fn check_compaction_needed(
     let threshold = threshold_override.unwrap_or_else(|| {
         config
             .get_param::<f64>("GOOSE_AUTO_COMPACT_THRESHOLD")
-            .unwrap_or(0.3) // Default to 30%
+            .unwrap_or(0.00001) // Default to 30%
```
michaelneale: is this intentional to be 30% - or left in by mistake @katzdave ?

michaelneale:
@katzdave hrm just saw this: 

```
Error: Request failed: Streaming request failed with status: 400 Bad Request. Error: {"type":"error","error":{"type":"invalid_request_error","message":"prompt is too long: 210052 tokens > 200000 maximum"}}


Context: ●●●●●●●●○○ 83% (165147/200000 tokens)
( O)> Press Enter to send, Ctrl-J for new line
```
which is first time - was a long run with anthropic provider + opus. Not sure if this is up to date with main but looks like token calculating is quite right, but even then - the problem is I think we do need to compact during the agent loop for complex changes unfortunately

michaelneale:
@katzdave I did have a quick look (but failed) to see if it would be easy to add it to the main agent loop - but I wasn't sure if we still had the summarizer that runs when it fills up (if we don't, we may need to add something like it back in on a higher threshold but not waiting for user input to trigger) to avoid running up against the limit when it is in the agent loop

katzdave:
> @katzdave hrm just saw this:
> 
> ```
> Error: Request failed: Streaming request failed with status: 400 Bad Request. Error: {"type":"error","error":{"type":"invalid_request_error","message":"prompt is too long: 210052 tokens > 200000 maximum"}}
> The tool calling loop was interrupted. How would you like to proceed?
>   error: The error above was an exception we were not able to handle.
> These errors are often related to connection or authentication
> We've removed the conversation up to the most recent user message
> - depending on the error you may be able to continue
> 
> 
> Context: ●●●●●●●●○○ 83% (165147/200000 tokens)
> ( O)> Press Enter to send, Ctrl-J for new line
> ```
> 
> which is first time - was a long run with anthropic provider + opus. Not sure if this is up to date with main but looks like token calculating is quite right, but even then - the problem is I think we do need to compact during the agent loop for complex changes unfortunately

One idea I chatted with Douwe about is to potentially have a tool for the agent to decide itself to autocompact + combined with task tracking that should help resume after compaction; I worry if we compact in the middle of an agent loop it might get pretty lost whereas with those signals it can start fresh a bit more cleanly.

I also think that marking a task as done provides a strong signal that the agent might be in a decent resting point to potentially compact or feed in new user context. 

Working on some token counting fixes, but I think adding to the agent loop makes sense as a next wave effort.

katzdave:
I also swapped in a 50% more summarize prompt. I don't notice much of a difference.

michaelneale:
@katzdave in that case do you think we should we have task tracking in this PR (if we can make it work) - or anther PR (could also move it to developer MCP if that made it easier) - or is over loading it? 

"tool for the agent to decide itself to autocompact + combined with task tracking" yes - so you think compact is itself a tool to call? (or just part of agent loop + heuristics so it isn't an extra tool call?)

crates/goose/src/prompts/summarize_oneshot_short.md
```diff
@@ -0,0 +1,26 @@
...
+2. **Technical Concepts** – All discussed tools, methods  
+3. **Files + Code** – Viewed/edited files, full code, change justifications  
+4. **Errors + Fixes** – Bugs, resolutions, user-driven changes  
+5. **Problem Solving** – Issues solved or in progress  
+6. **User Messages** – All user messages, exclude tool output  
+7. **Pending Tasks** – All unresolved user requests  
+8. **Current Work** – Active work at summary request time: filenames, code, alignment to latest instruction  
+9. **Next Step** – *Include only if* directly continues user instruction  
+
+> ⚠️ No new ideas unless user confirmed
```
michaelneale: not sure about how many tokens an emoji uses?

crates/goose/src/context_mgmt/summarize.rs
```diff
@@ -75,7 +75,7 @@ pub async fn summarize_messages_oneshot(
     };
 
     // Render the one-shot summarization prompt
-    let system_prompt = render_global_file("summarize_oneshot.md", &context)?;
```
michaelneale: can get rid of old now?

token_debug/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
```
michaelneale: this intended to be here or should be in existing test code dir?

katzdave: Nope, deleted

PR #3564, opened 2025-07-21
feat: migrate JsonRpcMessage/Request/Response/Error/Notification from internal mcp crates to rmcp versions

alexhancock:
Replace `JsonRpcMessage/Request/Response/Error/Notification` from the internal mcp crates with the equivalent https://github.com/modelcontextprotocol/rust-sdk types


crates/goose-mcp/src/developer/mod.rs
```diff
@@ -675,15 +675,21 @@ impl DeveloperRouter {
...
+                                jsonrpc: JsonRpcVersion2_0,
+                                notification: Notification {
+                                    method: "notifications/message".to_string(),
+                                    params: json!({
+                                        "data": {
+                                            "type": "shell",
+                                            "stream": "stdout",
+                                            "output": line.to_string(),
+                                        }
+                                    })
```
jamadeo: in rmcp I see
```
/// A JSON object type alias for convenient handling of JSON data.
///
/// You can use [`crate::object!`] or [`crate::model::object`] to create a json object quickly.
/// This is commonly used for storing arbitrary JSON data in MCP messages.
pub type JsonObject<F = Value> = serde_json::Map<String, F>;
```
can we use that `object!` macro here?

crates/mcp-client/src/client.rs
```diff
@@ -112,7 +115,8 @@ where
     T: TransportHandle + Send + Sync + 'static,
 {
     service: Mutex<tower::timeout::Timeout<McpService<T>>>,
-    next_id: AtomicU64,
+    next_id: RequestId,
+    next_id_counter: AtomicU64, // Added for atomic ID generation
```
jamadeo, line 119: do we need both of these? can we construct the RequestId from the Atomic when we need it?

crates/mcp-client/src/client.rs
```diff
@@ -172,64 +183,67 @@ where
...
-                    .server_info
-                    .as_ref()
-                    .map(|s| s.name.clone())
-                    .unwrap_or("".to_string()),
-                method: method.to_string(),
-                // we don't need include params because it can be really large
-                source: Box::<Error>::new(e.into()),
-            })?;
+        let request_json = serde_json::to_value(&request)?;
+        let generic_request: JsonRpcMessage = serde_json::from_value(request_json)?;
```
jamadeo: do we need a to_value/from_value here?

alexhancock:
Will rebase to address import conflicts + address PR feedback in the morning

crates/mcp-client/src/service.rs
```diff
@@ -50,24 +50,26 @@ where
...
                     transport.send(request).await?;
-                    Ok(JsonRpcMessage::Nil)
+                    // Return a dummy response for notifications
+                    let dummy_response: JsonRpcMessage =
+                        JsonRpcMessage::Response(rmcp::model::JsonRpcResponse {
+                            jsonrpc: rmcp::model::JsonRpcVersion2_0,
+                            id: rmcp::model::RequestId::Number(0),
+                            result: serde_json::Map::new(),
+                        });
+                    Ok(dummy_response)
```
jamadeo, line 72: I was going to suggest we make the message optional now that Nil isn't a valid JsonRpcMessage, but this code will all go away soon anyway, so maybe not that important...

jamadeo (approved):
Nice

PR #3565, opened 2025-07-21
fix: clean up 

wendytang:


crates/goose/src/agents/subagent_execution_tool/executor/mod.rs
```diff
@@ -180,7 +181,7 @@ async fn collect_results(
         if let Some(data) = result.data.as_mut() {
             if let Some(data_str) = data.as_str() {
                 if data_str.len() > 650 {
-                    *data = serde_json::Value::String(format!("{}...", &data_str[..650]));
+                    *data = serde_json::Value::String(safe_truncate(data_str, 650));
```
lifeizhou-ap: Hi @wendytang 

The result is returned as tool result which gives LLM context.  If we truncate here, the agent cannot work out the final result.  So it cannot be truncated here and maybe we can use display side to truncate

https://github.com/block/goose/pull/3444#discussion_r2213199216

crates/goose-cli/src/session/output.rs
```diff
@@ -437,11 +437,26 @@ fn print_markdown(content: &str, theme: Theme) {
 
 const INDENT: &str = "    ";
 
-fn get_tool_params_max_length() -> usize {
-    Config::global()
-        .get_param::<usize>("GOOSE_CLI_TOOL_PARAMS_TRUNCATION_MAX_LENGTH")
-        .ok()
-        .unwrap_or(40)
+fn get_tool_params_max_length(key: Option<&str>) -> usize {
```
lifeizhou-ap: I feel [this](https://github.com/block/goose/pull/3444#discussion_r2213121895) might be cleaner.

lifeizhou-ap:
Hi @wendytang Thanks for cleaning up! Looks good!

One thing that we have to change is to keep the result without truncate in the tool response and truncate the result in the display or somewhere else you think it is appropriate.

crates/goose-cli/src/session/output.rs
wendytang: addressed

crates/goose/src/agents/subagent_execution_tool/executor/mod.rs
wendytang: I removed the truncation for now

wendytang:
@lifeizhou-ap thanks for the review - 

updates include:
* remove 650 char results truncation to pass LLM full task results
* remove mcp notification sent from subagent (channel receiver was blocking tasks from executing)

lifeizhou-ap (approved):

PR #3566, opened 2025-07-21
fix: Handle non-default base path for OpenAI compatible model fetching

shelajev:
when workingon the PR for docs configuring Docker model runner which uses OpenAI compatible API backend, I had some issues when goose configure would error when it wouldn't figure out how to fetch the models. 

This PR fixes an issue when a non-default OPENAI_BASE_PATH is configured. I think it might still fail in some corner cases where API is not on the /v1/chat/completions path, but it's better than hardcoding the v1/models.

### DIFF
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -197,7 +197,7 @@ impl Provider for OpenAiProvider {
         let base_url =
             url::Url::parse(&self.host).map_err(|e| ProviderError::RequestFailed(e.to_string()))?;
         let url = base_url
-            .join("v1/models")
+            .join(&self.base_path.replace("v1/chat/completions", "v1/models"))
             .map_err(|e| ProviderError::RequestFailed(e.to_string()))?;
         let mut request = self.client.get(url).bearer_auth(&self.api_key);
         if let Some(org) = &self.organization {


michaelneale:
@shelajev and this works fine with openai directly? (would like to confirm that at is correct for 1st party)

shelajev:
Yeah, it's because the default base path is "v1/chat/completions" and the model endpoint becomes "v1/models": 

```~/ai-contrib/goose $ cargo run --package goose-cli -- configure
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.22s
     Running `target/debug/goose configure`

This will update your existing config file
  if you prefer, you can edit it directly at /Users/shelajev/.config/goose/config.yaml

┌   goose-configure
│
│  Configure Providers
│
│  OpenAI
│
●  OPENAI_API_KEY is set via environment variable
│
◇  Would you like to save this value to your keyring?
│  No
│
●  OPENAI_HOST is already configured
│
│  Yes
│
◇  Enter new value for OPENAI_HOST
│  https://api.openai.com
│
●  OPENAI_BASE_PATH is already configured
│
│  Yes
│
◇  Enter new value for OPENAI_BASE_PATH
│  v1/chat/completions
│
◇  Model fetch complete
│
◆  Select a model:
│
│  ● babbage-002
│  ○ chatgpt-4o-latest
│  ○ codex-mini-latest
│  ○ computer-use-preview
│  ○ computer-use-preview-2025-03-11
│  ○ dall-e-2
│  ○ dall-e-3
│  ○ davinci-002
│  ○ gpt-3.5-turbo
│  ○ gpt-3.5-turbo-0125
│  ○ gpt-3.5-turbo-1106
│  ○ gpt-3.5-turbo-16k
│  ○ gpt-3.5-turbo-instruct
│  ○ gpt-3.5-turbo-instruct-0914
│  ○ gpt-4
│  ○ gpt-4-0125-preview
│  ○ gpt-4-0613
│  ○ gpt-4-1106-preview
│  ○ gpt-4-turbo
│  ○ gpt-4-turbo-2024-04-09
│  ○ gpt-4-turbo-preview
│  ○ gpt-4.1
```

michaelneale (approved):
nice @shelajev, yep it works well, thanks!

PR #3568, opened 2025-07-21
feat (ui): close recipe modals with esc key

angelahning:
This PR is to make closing the Recipes modals easier, via the `esc` key. Currently users must click the "close" button on the modal.


michaelneale:
thanks @angelahning I like this - as I always reach for esc key!

ui/desktop/src/components/RecipeEditor.tsx
```diff
@@ -129,6 +130,12 @@ export default function RecipeEditor({ config }: RecipeEditorProps) {
     setParameters(allParams);
   }, [instructions, prompt]);
 
+  // Handle Esc key for Save Recipe Dialog
+  useEscapeKey(showSaveDialog, () => {
+    setShowSaveDialog(false);
```
michaelneale, line 135: I'm not sure what show save dialog does - seems to work though

michaelneale (approved):
tested, works nicely

Issue #3570, opened 2025-07-22
streaming spreads out across lines in GUI


michaelneale:
using anthropic: 

<img width="1154" height="840" alt="Image" src="https://github.com/user-attachments/assets/83a60712-4a38-49d9-a98d-fbb3ffd33da1" />

This looks very very wrong. 

testing with databricks now...

PR #3572, opened 2025-07-22
fix: track message id to keep like with like

michaelneale:
this fixes: https://github.com/block/goose/issues/3570 

a fairly serious bug where messages don't accumulate correctly.

Fixes GUI and CLI, and should help fix token accounting.


jamadeo (approved):
Huh... how'd it work before? I definitely remember trying this out

michaelneale:
@jamadeo yes super weird - I remember seeing streaming in CLI too - it certainly streamed (poorly!) to GUI so that showed this up.

PR #3575, opened 2025-07-22
fix: ensure retry-config and success-criteria are populated in openapi spec

cloud-on-prem:
- **fix: ensure retry config and success checks are part of openapi spec**
- **fix: generate new openapi specs**
- **chore: add a just command for testing openapi schema changes**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,6 +131,15 @@ jobs:
           source ./bin/activate-hermit
           export CARGO_INCREMENTAL=0
           cargo clippy --jobs 2 -- -D warnings
+
+      - name: Install Node.js Dependencies for OpenAPI Check
+        run: source ../../bin/activate-hermit && npm ci
+        working-directory: ui/desktop
+
+      - name: Check OpenAPI Schema is Up-to-Date
+        run: |
+          source ./bin/activate-hermit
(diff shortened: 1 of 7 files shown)


michaelneale (approved):
yep I think this makes sense - only concern is additional time blowout to CI For this (as it won't be in parallel I think?) - could have that as its own validation job or move it?

.github/workflows/ci.yml
```diff
@@ -131,6 +131,15 @@ jobs:
           source ./bin/activate-hermit
           export CARGO_INCREMENTAL=0
           cargo clippy --jobs 2 -- -D warnings
+
+      - name: Install Node.js Dependencies for OpenAPI Check
```
michaelneale, line 135: should this move to the lint check (we can rename that) for front end that runs in parallel? (not sure if it has to wait for cargo to do its thing however)

jsibbison-square:
Sounds good, if its fast enough can we add it to the existing pre-commit hook?

.github/workflows/ci.yml
cloud-on-prem: I originally thought about keeping it in a different job. We do need cargo to compile to check. The check is adding about ~2 seconds (on my machine)
We can move it out if it becomes a problem.

PR #3582, opened 2025-07-22
Regenerate openapi schema

jamadeo:
#3563 didn't re-generate the schema, and now we have a check that it's up to date (which is good!)


jamadeo:
fixed in another PR

PR #3584, opened 2025-07-22
fix: add fallback id to messages if none provided

zanesq:
fixes tool call response missing id and agent looping on the same task

should probably look into why its missing on the backend but this solves it for now from the ui side

cc @jamadeo

### DIFF
--- a/ui/desktop/src/hooks/useMessageStream.ts
+++ b/ui/desktop/src/hooks/useMessageStream.ts
@@ -4,8 +4,10 @@ import { getSecretKey } from '../config';
 import { Message, createUserMessage, hasCompletedToolCalls } from '../types/message';
 import { getSessionHistory } from '../api';
 
+let messageIdCounter = 0;
+
 function generateMessageId(): string {
-  return `msg-${Date.now()}`;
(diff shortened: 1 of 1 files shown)


ui/desktop/src/hooks/useMessageStream.ts
```diff
@@ -4,6 +4,10 @@ import { getSecretKey } from '../config';
 import { Message, createUserMessage, hasCompletedToolCalls } from '../types/message';
 import { getSessionHistory } from '../api';
 
+function generateMessageId(): string {
+  return `msg-${Date.now()}`;
```
DOsinga: maybe stay away from .now() ? if we have one message with a missing id we might have more than one and then they could easily get the same id. easiest would to just also have an index in that loop and then assign based on that

jamadeo:
@zanesq do you know how to reproduce a scenario where we get a message with missing id?

zanesq:
@jamadeo I haven't been able to figure out what is triggering it specifically but it might be occurring after summarization. Will keep an eye out. Let's go ahead and get this out for now to see if it helps the issue. If not we can dig in further.

Issue #3567, opened 2025-07-21
Switching Models does not immediately update


sakgoyal:
**Describe the bug**

When selecting a model in the settings, the currently selected model does not update till you switch menus and come back. 

**To Reproduce**
1. Have a model already selected
2. Click on settings on the left menu
3. in the models tab, change the model (eg: gemini to sonnet)
4. save
5. the selected model still says gemini
6. change to a different tab and come back (settings -> chat -> settings)
7. selected model is now updated to sonnet

**Expected behavior**
Settings UI should update immediately after saving

**Screenshots**

**Please provide following information:**
 - **OS & Arch:** Windows 10
 - **Interface:** UI
 - **Version:** v1.1.3

PR #3587, opened 2025-07-22
Fix model display name not being updated immediately after leaving settings

zanesq:
Fix model display name not being updated immediately after leaving settings

fixes https://github.com/block/goose/issues/3567


angelahning (approved):
Tested locally, works great

PR #3588, opened 2025-07-22
Improve session history loading resiliency

zanesq:
Some users report seeing a failed chat history screen. This improves the loading to not fail if there are corrupt sessions with missing metadata. Also added a check to hide any failure to load insights on home if the insights eventually load which avoids showing the error message after they

**Problem:** The session list endpoint was failing completely when encountering corrupted session files, showing "Failed to fetch" errors in the browser console and preventing any sessions from loading.

**Root Cause:** The get_valid_sorted_sessions function in /crates/goose/src/session/info.rs used filter_map with early returns that would cause the entire endpoint to fail when session metadata couldn't be read.  loaded.

**Solution:**

Replaced filter_map with explicit error handling using a for loop
Added graceful fallback for file modification time retrieval
Added comprehensive logging for corrupted sessions with count tracking
Sessions with corrupted metadata are now skipped gracefully while valid sessions continue to load
Fixed a clippy warning in the session routes by replacing an if-chain with a match statement
Key Changes:

Modified get_valid_sorted_sessions to handle corrupted sessions individually rather than failing completely
Added warning logs that track individual corrupted sessions and total count
Ensured the endpoint remains functional even with partial corruption

Screenshot 2025-07-22 at 10 10 18 AM


crates/goose-server/src/routes/session.rs
```diff
@@ -187,14 +187,20 @@ async fn get_session_insights(
...
         if let Some(tokens) = session.metadata.accumulated_total_tokens {
-            if tokens > 0 {
-                total_tokens += tokens as i64;
-            } else if tokens < 0 {
-                // Log negative token values for debugging
-                info!(
-                    "Warning: Session {} has negative accumulated_total_tokens: {}",
-                    session.id, tokens
-                );
+            match tokens.cmp(&0) {
```
zanesq, line 190: Fixed clippy warning

PR #3390, opened 2025-07-13
docs: fix YouTube Transcript MCP package manager

angiejones:
YouTube Transcript MCP uses uv but we incorrectly listed Node

PR #3595, opened 2025-07-22
docs: fix installation command for YouTube Transcript MCP in servers.json

jkawamoto:
Following up on #3390, servers.json also needs to be updated.

This is also related to https://github.com/jkawamoto/mcp-youtube-transcript/issues/26.


angiejones:
thanks so much!

angiejones (approved):

PR #3596, opened 2025-07-23
feat: subagent independent extension manager

wendytang:
Before, the subagent was sharing an mcp clients via the main agent's extension manager. 
Change this so that subagent has its own extension manager, hence its own mcp clients for isolation.

No visible performance changes in subagent task execution.


lifeizhou-ap:
Hi @wendytang, 

Just to confirm whether different subagent they are going to start different mcp client/server (I think they will :) ). Let's try it. 

I feel the sub_agent is more like `agent` now, just thinking maybe we should use `trait` consolidate common behaviours.

PR #3597, opened 2025-07-23
feat: subagent turn count, simple agent loop

wendytang:
* allow a maximum of 5 subagent loops 
* kept fairly small number of loops for now to ensure subagent tasks complete. 
* interesting problem to think about: depending on complexity of subagent tasks, size up or down # agent loop permitted


crates/goose/src/agents/subagent_task_config.rs
```diff
@@ -39,7 +39,7 @@ impl TaskConfig {
             provider,
             extension_manager,
             mcp_tx,
-            max_turns: Some(10),
+            max_turns: Some(5),
```
lifeizhou-ap: I am not sure whether `5` is enough to run a sub recipe.  Shall we open this attribute for the caller to set so that subrecipe can set more max_turns initially?

crates/goose/src/agents/subagent.rs
```diff
@@ -233,8 +208,14 @@ impl SubAgent {
         // Build system prompt using the template
         let system_prompt = self.build_system_prompt(&tools).await?;
 
+        // Generate response from provider with loop for tool processing (max_turns iterations)
+        let mut loop_count = 0;
+        let max_turns = self.config.max_turns.unwrap_or(2);
```
lifeizhou-ap: maybe extract `2` as a constant to avoid magic number?

lifeizhou-ap:
Hi @wendytang ,

While reviewing this PR, I feel the subagent has part of the `agent`(agent.rs) reply() logic but not all.  in the `agent` we have max turn logic, final result and retry logic.  These are the logic that we would also to have in the `subagent`. Just wondering whether it is possible to refactor the code to reuse the `agent` reply? so that we don't have to duplicate the logic every time when we have a new feature in `agent`

lifeizhou-ap (approved):

crates/goose/src/agents/subagent.rs
```diff
@@ -182,20 +182,6 @@ impl SubAgent {
             .as_ref()
             .ok_or_else(|| anyhow!("No extension manager configured for subagent"))?;
 
-        // Check if we've exceeded max turns
```
DOsinga, line 206: independent of this, is there a plan to unify this subagent loop with the main agent loop? already I feel l like we're starting to see code drift between the two approaches

wendytang:
I agree that we should unify subagent -> simply use agent reply(). 

Dug into this a bit, and it boils down to the reply method's return type. It reaches the upper limits of my rust knowledge for now, but I do believe this is possible.

```
The error occurs because you're trying to call child_agent.reply() within the Agent::reply() method, but the future returned by reply() is not Send. This is a common Rust async issue related to thread safety.
Looking at the reply() method signature on line 876:

pub async fn reply(
    &self,
    messages: &[Message],
```

PR #3598, opened 2025-07-23
Added option to summarize the chat when an error is triggered

zanesq:
Added option to summarize the chat when an error is triggered

### DIFF
--- a/ui/desktop/src/components/BaseChat.tsx
+++ b/ui/desktop/src/components/BaseChat.tsx
@@ -157,6 +157,7 @@ function BaseChatContent({
     updateMessageStreamBody,
     sessionMetadata,
     isUserMessage,
+    clearError,
   } = useChatEngine({
     chat,
     setChat,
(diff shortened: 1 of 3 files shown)


zanesq:
When testing locally I would temporarily force an error from the backend, code like this in reply.rs

```
    // TEMPORARY: Trigger error after 2 messages for testing auto-summarize feature
    // This gives us some conversation history to actually summarize
    if messages.len() >= 2 && messages.iter().filter(|m| m.role == rmcp::model::Role::User).count() >= 2 {
        let (tx, rx) = mpsc::channel(100);
        let stream = ReceiverStream::new(rx);

        tokio::spawn(async move {
            let _ = stream_event(
                MessageEvent::Error {
                    error: "Temporary test error - this is a simulated backend error for testing the auto-summarize feature".to_string(),
                },
                &tx,
            ).await;
        });

        return Ok(SseResponse::new(stream));
    }
```

michaelneale:
yeah looks good

Issue #3553, opened 2025-07-21
Unexpected warning/error message when running sub-recipe and user enters CTRL+C


lifeizhou-ap:
**Describe the bug**
When Goose runs recipe with sub-recipe, it should handle CTRL+C gracefully without showing unexpected warning/error message

**To Reproduce**
1: user runs a recipe with sub-recipes. 
2: user enters CTRL+C
3. user see a warning message "WARN goose::agents::subagent_execution_tool::task_execution_tracker: Failed to send tasks update notification: channel close"
4. If the user runs the recipe in interactive mode, user will also see an error message "ERROR goose::agents::subagent_execution_tool::workers: Worker failed to send result: channel closed"

**Expected behaviour**
Should not see the error/warning message.  Goose should handle CTRL+C gracefully.

**Screenshots**
<img width="1814" height="530" alt="Image" src="https://github.com/user-attachments/assets/fe6d9777-f650-44e9-a4ad-df55c2dec8d9" />

PR #3599, opened 2025-07-23
reuse the cancellation token in the agent level

lifeizhou-ap:
Handle CLI CTRL+C gracefully to cancel the tasks

fixes issue https://github.com/block/goose/issues/3553

Close https://github.com/block/goose/pull/3550 as now we can reuse the cancellation token created in the agent reply

### DIFF
--- a/crates/goose/tests/agent.rs
+++ b/crates/goose/tests/agent.rs
@@ -603,7 +603,7 @@ mod final_output_tool_tests {
             }),
         );
         let (_, result) = agent
-            .dispatch_tool_call(tool_call, "request_id".to_string())
+            .dispatch_tool_call(tool_call, "request_id".to_string(), None)
             .await;
 
         assert!(result.is_ok(), "Tool call should succeed");
--- a/crates/goose/tests/private_tests.rs
+++ b/crates/goose/tests/private_tests.rs
@@ -885,7 +885,7 @@ async fn test_schedule_tool_dispatch() {
     };
 
     let (request_id, result) = agent
-        .dispatch_tool_call(tool_call, "test_dispatch".to_string())
+        .dispatch_tool_call(tool_call, "test_dispatch".to_string(), None)
         .await;
     assert_eq!(request_id, "test_dispatch");
     assert!(result.is_ok());


michaelneale:
there are a few tools removed from this - I assume that is by design? (ie not needed). Not sure how to validate functionality (as not aware of subagents)

lifeizhou-ap:
> there are a few tools removed from this - I assume that is by design? (ie not needed). Not sure how to validate functionality (as not aware of subagents)

Hi @michaelneale yes, the sub_recipe_execution_tool are removed as they are renamed to sub_agent_execution_tool.  So they are not long required. 

In the PR I used tokio::select! start multiple operations async, when the cancellation token is cancelled, the other operations (the tasks executions) will be cancelled. 

One of the test cases : I tried to kick off a recipe with 2 subrecipes (goose run --recipe sub_recipe) running. If I enters CRLT+C, the processes of running sub recipes will be terminated.  Also it does not have warning/error message as specified in the issue, because if it is cancelled, the taskExecuteTracker won't send notification to the channel. 

One thing I forgot to mention is: the cancellation token is passed through to a few modules up to the tasks.rs.  I am going to have a follow up PR to make this better.

crates/goose/src/agents/agent.rs
```diff
@@ -154,6 +155,7 @@ impl Agent {
             tool_monitor,
             router_tool_selector: Mutex::new(None),
             scheduler_service: Mutex::new(None),
+            current_cancellation_token: Arc::new(Mutex::new(None)),
```
DOsinga: why do we need this? can't we just use the existing token and just pass it on, instead of this being a property on the agent? I know agents aren't re-entrant, but still

lifeizhou-ap: When we invoke the sub_agent_execute_tool in the `dispatch_tool_call` function, we need to pass `cancellation_token` so that the tool can reuse the cancellation token.  The `dispatch_tool_call` function is not only triggered by `reply` directly, it can be trigged by reply -> handle_approval_tool_requests -> dispatch_tool_call.  

I can try to pass the token via those functions, but I feel it is a bit verbose to pass it around in those functions

crates/goose/src/agents/subagent_execution_tool/task_execution_tracker.rs
```diff
@@ -93,6 +95,41 @@ impl TaskExecutionTracker {
...
             display_mode,
+            cancellation_token,
+        }
+    }
+
+    fn is_cancelled(&self) -> bool {
+        self.cancellation_token
+            .as_ref()
+            .is_some_and(|t| t.is_cancelled())
+    }
```
DOsinga, line 106: we should probably put this in some util function and use everywhere but we can do that in something separate

DOsinga (approved):

PR #3600, opened 2025-07-23
appending to pr: ads auto summarize to one shot

michaelneale:
testing this out locally - but this ads autosummarize/compact to the clients calling agent

this is to augment the https://github.com/block/goose/tree/dkatz/goose-compact2 PR


katzdave:
Cool, taking this for a spin

katzdave:
Ok this is pretty cool! I don't find it super disruptive seeing the summary take over, and then run the prompt. Would be great if we added more clarity that's happening before that but overall a pretty nice experience.

I'm game to merge into my branch.

So here's what I pieced together is I think a lot of the anomalies come from this inherited behavior where we rip out the last tool call response and then insert it back in. Especially when you're dealing with the jsonl files, those can be huge, so leads to situations where you summarize and then end up back with an 80k context (because it's basically just the orignal jsonl from the last tool call coming back. 

I think doesn't explain the session dying randomly, but I pushed a version with that tool call response filter ripped out. Now the summaries are actually super tiny, not sure if this will degrade performance. Tested a little bit and seems to generally be fine; especially since we don't summarize until the session cedes control back to the user, so I don't see why the recent tool call would be that important; goose can make it again if necessary.

michaelneale:
agree - I think we are over indexing on tool call results, absorbing them all vs sipping on them, so agree with that change - is that on your other branch?

PR #3609, opened 2025-07-23
fix: multi-tool calls in streamed openai-compatible responses

jamadeo:
Previously this did not properly handle when the provider streams back a reply with >1 tool call.

The stream used here for the test was captured with a simple proxy. We should consider doing more of this, perhaps building out something like the scenario test to capture at the request level to test this provider code.

### DIFF
--- a/crates/goose/src/providers/formats/openai.rs
+++ b/crates/goose/src/providers/formats/openai.rs
@@ -437,56 +437,91 @@ where
             if chunk.choices.is_empty() {
                 yield (None, usage)
             } else if let Some(tool_calls) = &chunk.choices[0].delta.tool_calls {
-                let tool_call = &tool_calls[0];
-                let id = tool_call.id.clone().ok_or(anyhow!("No tool call ID"))?;
-                let function_name = tool_call.function.name.clone().ok_or(anyhow!("No function name"))?;
-                let mut arguments = tool_call.function.arguments.clone();
+                // Initialize tracking for multiple tool calls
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/formats/openai.rs
```diff
@@ -437,56 +437,83 @@ where
...
+                        if let Some(line) = strip_data_prefix(&response_str) {
+                            let tool_chunk: StreamingChunk = serde_json::from_str(line)
+                                .map_err(|e| anyhow!("Failed to parse streaming chunk: {}: {:?}", e, &line))?;
+
+                            if let Some(delta_tool_calls) = &tool_chunk.choices[0].delta.tool_calls {
+                                for delta_call in delta_tool_calls {
+                                    if let Some(index) = delta_call.index {
+                                        if let Some((_, _, ref mut args)) = tool_call_data.get_mut(&index) {
+                                            args.push_str(&delta_call.function.arguments);
+                                        } else if let (Some(id), Some(name)) = (&delta_call.id, &delta_call.function.name) {
```
DOsinga, line 464: do we need this branch?

DOsinga (approved):

crates/goose/src/providers/formats/openai.rs
jamadeo: Yes this is actually the part that we missed before--you'd think it's the tool_calls array but that still has a single item in the first part of the response. Then the subsequent chunks might start new calls by setting a new index and id

Issue #3431, opened 2025-07-15
File picker for scheduling recipes should default to recipe dir


angiejones:
on the Create Schedule modal, the file chooser should start in the recipe directory :` ~/.config/goose/recipes/` or similar on Windows. Especially since the yaml files are in a hidden directory that won't show in a file chooser widget by default.

<img width="404" height="126" alt="Image" src="https://github.com/user-attachments/assets/c9357a5e-1a93-4d54-9cf0-0cea4237b128" />

From @DOsinga:

> If you don’t specify the path to run, the CLI will look in two locations for recipes: the configured github repository (GOOSE_RECIPE_GITHUB_REPO) and the recipe search path (GOOSE_RECIPE_PATH). The desktop will show you recipes in ~/.config/goose/recipes/ and in the .goose/recipes/ of the current directory.
Proposal: drop local recipes from the frontend. Have the CLI also look in ~/.config/goose/recipes/. Then show all available recipes in the frontend (i.e. also through git and the search path), but hide the edit button for the git ones.

DOsinga:
upping this to p2 - probably needs to wait for a more fundamental overhaul on how we show recipes though

PR #3611, opened 2025-07-23
feat (ui): File picker for scheduling recipes default to recipe dir

angelahning:
addresses https://github.com/block/goose/issues/3431

File picker now defaults to "~/.config/goose/recipes". 

Screenshot 2025-07-23 at 1 59 42 PM


ui/desktop/src/main.ts
```diff
@@ -1177,10 +1178,34 @@ ipcMain.handle('get-wakelock-state', () => {
...
     properties: process.platform === 'darwin' ? ['openFile', 'openDirectory'] : ['openFile'],
-  })) as unknown as OpenDialogReturnValue;
+  };
+
+  // Set default path if provided
+  if (defaultPath) {
+    // Expand tilde to home directory
+    const expandedPath = defaultPath.startsWith('~')
+      ? path.join(os.homedir(), defaultPath.slice(1))
+      : defaultPath;
```
DOsinga: I'm pretty sure we have this code elsewhere. let's reuse

ui/desktop/src/main.ts
```diff
@@ -1177,10 +1178,34 @@ ipcMain.handle('get-wakelock-state', () => {
...
+    try {
+      const stats = await fs.stat(expandedPath);
+      if (stats.isDirectory()) {
+        dialogOptions.defaultPath = expandedPath;
+      } else {
+        dialogOptions.defaultPath = path.dirname(expandedPath);
+      }
+    } catch (error) {
+      // If path doesn't exist, fall back to home directory and log error
+      console.error(`Default path does not exist: ${expandedPath}, falling back to home directory`);
```
DOsinga, line 1201: you're not actually checking that the file didn't exist here though - when would this error be triggered?

katzdave (approved):
Nice!

ui/desktop/src/main.ts
angelahning: surprisingly it doesn't appear so. I have created a helper.

ui/desktop/src/main.ts
angelahning: fs.stat(path) throws when the file/dir does not exist.

Issue #3628, opened 2025-07-24
In Goose CLI sub agent execution display does not show the task completion result sometimes


lifeizhou-ap:
**Describe the bug**
In Goose CLI sub agent execution display does not show the task completion result sometimes

**To Reproduce**
1. Run a subrecipe with different params in parallel (ie. 2 tasks in parallel
2. When last task completes, the task result does not show up

**Expected behavior**
Both task result should be shown up

**Screenshots**
<img width="1889" height="394" alt="Image" src="https://github.com/user-attachments/assets/b84a25a4-8784-4be8-ac50-4ccb67fba870" />

**Please provide following information:**
 - **OS & Arch:** MacOS
 - **Interface:** CLI
 - **Version:** current main branch e6829409297455a7a6d02e824dd4f92932216108

PR #3629, opened 2025-07-24
fix: ensure execution task result is shown

lifeizhou-ap:
**Why**
fix https://github.com/block/goose/issues/3628

**What**
- Add extra sleep to make sure the notification has been received and displayed
- Extract checking token is cancelled to the util module


crates/goose/src/agents/subagent_execution_tool/task_execution_tracker.rs
```diff
@@ -309,7 +309,8 @@ impl TaskExecutionTracker {
             .collect();
 
         let event = TaskExecutionNotificationEvent::tasks_complete(stats, failed_tasks);
-
         self.try_send_notification(event, "tasks complete");
+        // Wait for the notification to be recieved and displayed before clearing the tasks
+        sleep(Duration::from_millis(COMPLETION_NOTIFICATION_DELAY_MS)).await;
```
DOsinga, line 304: sounds good, but are you saying the message doesn't get there if we complete here?

DOsinga (approved):

crates/goose/src/agents/subagent_execution_tool/task_execution_tracker.rs
lifeizhou-ap: The extra sleep is to handle the race condition.  

The notification flow:
  1. TaskExecutionTracker → try_send → mpsc channel queue
  2. ReceiverStream reads from queue → agent's event stream
  3. Session loop processes from agent stream → CLI display

When the last task completes,  notifications is sent async and the sub agent execution tool completes.  The extra sleep is to ensure the asynchronous notification processing completes before the tool returns its result to the agent.

Without sleep:
 T=1ms: Notifications queued in channel
  T=2ms: Tool returns immediately (no sleep)
  T=15ms: Agent stream terminates
  T=15ms: Session loop exits
  T=16ms: Channel dropped with notifications inside

michaelneale: I think this could be a future source of bugs - really no other way?

PR #3634, opened 2025-07-24
Swap MCP client implementations to rmcp

jamadeo:
This still leaves behind a small abstraction over the clients, which we could keep around or eventually get rid of.


crates/goose-server/src/logging.rs
```diff
@@ -66,6 +66,7 @@ pub fn setup_logging(name: Option<&str>) -> Result<()> {
 
     // Create console logging layer for development - INFO and above only
     let console_layer = fmt::layer()
+        .with_writer(std::io::stderr)
```
jamadeo, line 69: This change goes a bit further than strictly necessary as it moves all console logging to stderr, while we only truly needed to do that for mcp servers. But IMO we should always prefer stderr to stdout for logging.

jamadeo:
.bundle

jamadeo:
Thanks to @aharvard 's upgrading to rmcp 0.3.1 this pr became decently smaller 🎉

jamadeo:
.bundle

jamadeo:
.bundle

batmat:
FTR, this introduced https://github.com/block/goose/issues/4481

Issue #1303, opened 2025-02-20
Goose stops responding without any error message once conversation hits context limit on openrouter.


myevolve:
**Describe the bug**
Goose will eventually just stop responding and not display the error message from the llm, in this case, openrouter using anthropic/claud-3.5-sonnet

**To Reproduce**
1. Build for long enough to exceed 200K context.

**Expected behavior**
Expected behavior is that Goose would use the middle out transform and compress prompt to context limit or just below it so it doesn't fail.

**Screenshots**
Trace from langfuse:  
This endpoint's maximum context length is 200000 tokens. However, you requested about 201326 tokens (197771 of text input, 3555 of tool input). Please reduce the length of either one, or use the "middle-out" transform to compress your prompt automatically.

**Please provide following information:**
 - **OS & Arch:** [Ubuntu 24.04 x86]
 - **Interface:** [CLI]
 - **Version:** [v1.0.7]
 - **Extensions enabled:** [Developer,Computer Controller, Memory,Fetch,Git,Github]
 - **Provider & Model:** [openrouter:anthropic/claude-3.5-sonnet]

**Additional context**

yingjiehe-xyz:
@myevolve thanks for reporting it, previously we have some discussion about the truncate on goose for all providers, will update here later

CyphrRiot:
This has been ongoing and since 1.0.1 at least. I've reported it many times. Be nice to fix this once and for all, because the console (CLI and GUI) failure is not at all descriptive of the issue.

salman1993:
@HunterR0se I believe this specific error is coming from OpenRouter - https://openrouter.ai/docs/features/message-transforms

However, in the case that OpenRouter doesn't handle context length errors, Goose should try to do a better job at truncating.

pfergi42:
Bump. This is the biggest issue I'm having recently. It requires restarting a new goose session, and then having to get goose up to speed all over again.

CyphrRiot:
> Bump. This is the biggest issue I'm having recently. It requires restarting a new goose session, and then having to get goose up to speed all over again.

Exactly!  And the amount of tokens and API calls to do this is both mentally (and financially) burdensome at the least and infuriating at most.

I love goose, but this really is a "show stopper" issue for the project.

kwsantiago:
I've been working on an initial fix for this since it's a critical bug. Put up what I have so far: https://github.com/block/goose/pull/3635

Looking for testers, feedback, and/or contributions. Thanks!

[...]

PR #3635, opened 2025-07-24
fix: auto-compact on context limit error

kwsantiago:
## Summary
- When context limit is exceeded, automatically compact the conversation to continue instead of failing.
- Fixes https://github.com/block/goose/issues/1303

## Changes
- Catch `ProviderError::ContextLengthExceeded` in agent reply loop
- Call `auto_compact::perform_compaction` to reduce conversation size without threshold checking
- Refactored compaction logic to separate threshold checking from actual compaction
- Continue conversation if compaction succeeds, fail gracefully with clear error message if not

## Testing
- [X] Builds successfully
- [X] Existing tests pass
- [X] Clippy passes with no warnings
- [X] Code formatted with cargo fmt

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -1218,11 +1218,29 @@ impl Agent {
                                 messages_to_add.push(final_message_tool_resp);
                             }
                         }
-                        Err(ProviderError::ContextLengthExceeded(_)) => {
-                            yield AgentEvent::Message(Message::assistant().with_context_length_exceeded(
-                                    "The context length of the model has been exceeded. Please start a new session and try again.",
-                                ));
-                            break;
+                        Err(ProviderError::ContextLengthExceeded(error_msg)) => {
+                            info!("Context length exceeded, attempting compaction");
+
+                            match auto_compact::perform_compaction(self, messages.messages()).await {
+                                Ok(compact_result) => {
+                                    messages = compact_result.messages;
+
+                                    yield AgentEvent::Message(
+                                        Message::assistant().with_summarization_requested(
+                                            "Context limit reached. Conversation has been automatically compacted to continue."
+                                        )
+                                    );
+                                    yield AgentEvent::HistoryReplaced(messages.messages().to_vec());
+
+                                    continue;
+                                }
+                                Err(_) => {
+                                    yield AgentEvent::Message(Message::assistant().with_context_length_exceeded(
+                                        format!("Context length exceeded and cannot summarize: {}. Unable to continue.", error_msg)
+                                    ));
+                                    break;
+                                }
+                            }
                         }
                         Err(e) => {
                             error!("Error: {}", e);
--- a/crates/goose/src/context_mgmt/auto_compact.rs
+++ b/crates/goose/src/context_mgmt/auto_compact.rs
@@ -120,6 +120,50 @@ pub async fn check_compaction_needed(
     })
 }
 
+/// Perform compaction on messages without checking thresholds
+///
+/// This function directly performs compaction on the provided messages.
+/// If the most recent message is a user message, it will be preserved by removing it
+/// before compaction and adding it back afterwards.
+///
+/// # Arguments
+/// * `agent` - The agent to use for context management
+/// * `messages` - The current message history
+///
+/// # Returns
+/// * `AutoCompactResult` containing the compacted messages and metadata
+pub async fn perform_compaction(agent: &Agent, messages: &[Message]) -> Result<AutoCompactResult> {
+    info!("Performing message compaction");
+
+    // Check if the most recent message is a user message
+    let (messages_to_compact, preserved_user_message) = if let Some(last_message) = messages.last()
+    {
+        if matches!(last_message.role, rmcp::model::Role::User) {
+            // Remove the last user message before compaction
+            (&messages[..messages.len() - 1], Some(last_message.clone()))
+        } else {
+            (messages, None)
+        }
+    } else {
+        (messages, None)
+    };
+
+    // Perform the compaction on messages excluding the preserved user message
+    let (mut compacted_messages, _, summarization_usage) =
+        agent.summarize_context(messages_to_compact).await?;
+
+    // Add back the preserved user message if it exists
+    if let Some(user_message) = preserved_user_message {
+        compacted_messages.push(user_message);
+    }
+
+    Ok(AutoCompactResult {
+        compacted: true,
+        messages: compacted_messages,
+        summarization_usage,
+    })
+}
+
 /// Check if messages need compaction and compact them if necessary
 ///
 /// This is a convenience wrapper function that combines checking and compaction.
@@ -163,33 +207,8 @@ pub async fn check_and_compact_messages(
         check_result.usage_ratio * 100.0
     );
 
-    // Check if the most recent message is a user message
-    let (messages_to_compact, preserved_user_message) = if let Some(last_message) = messages.last()
-    {
-        if matches!(last_message.role, rmcp::model::Role::User) {
-            // Remove the last user message before auto-compaction
-            (&messages[..messages.len() - 1], Some(last_message.clone()))
-        } else {
-            (messages, None)
-        }
-    } else {
-        (messages, None)
-    };
-
-    // Perform the compaction on messages excluding the preserved user message
-    let (mut compacted_messages, _, summarization_usage) =
-        agent.summarize_context(messages_to_compact).await?;
-
-    // Add back the preserved user message if it exists
-    if let Some(user_message) = preserved_user_message {
-        compacted_messages.push(user_message);
-    }
-
-    Ok(AutoCompactResult {
-        compacted: true,
-        messages: compacted_messages,
-        summarization_usage,
-    })
+    // Use perform_compaction to do the actual work
+    perform_compaction(agent, messages).await
 }
 
 #[cfg(test)]


crates/goose/src/providers/factory.rs
```diff
@@ -162,7 +193,20 @@ fn create_provider(name: &str, model: ModelConfig) -> Result<Arc<dyn Provider>>
         "groq" => Ok(Arc::new(GroqProvider::from_env(model)?)),
         "litellm" => Ok(Arc::new(LiteLLMProvider::from_env(model)?)),
         "ollama" => Ok(Arc::new(OllamaProvider::from_env(model)?)),
-        "openrouter" => Ok(Arc::new(OpenRouterProvider::from_env(model)?)),
+        "openrouter" => {
+            let mut provider = OpenRouterProvider::from_env(model)?;
+
+            // Check if middle-out should be enabled by default
```
michaelneale: ideally could this be in the from_env implementation encapsulated there?

crates/goose/src/providers/utils.rs
```diff
@@ -347,6 +347,63 @@ pub fn emit_debug_trace<T1, T2>(
     );
 }
 
+/// Parse token information from common error message formats
+pub fn parse_token_info_from_error(error_msg: &str) -> (Option<u32>, Option<u32>) {
+    // OpenRouter format: "This endpoint's maximum context length is 200000 tokens. However, you requested about 201326 tokens"
```
michaelneale: is there similar for openai and other big ones?

michaelneale:
thanks @kwsantiago this looks interesting, seems a few things combined (wasn't aware of that openrouter stuf f- we have some other compaction things happening so hopefully wont' hit that or need it?). 

Can you talk a bit more about how this prevents - there is new state with the token counting, not sure if that helps with it, or is it sniffing out error messages. I wonder if the openrouter changex can be a separate PR to make this change clearer, but I like the direction

crates/goose/src/agents/agent.rs
```diff
@@ -220,6 +234,21 @@ impl Agent {
...
+    pub fn provider_name(&self) -> String {
+        // Get the provider name from config
+        let config = Config::global();
+        config
+            .get_param::<String>("GOOSE_PROVIDER")
+            .unwrap_or_else(|_| "unknown".to_string())
+    }
+
+    /// Enable middle-out transform for OpenRouter provider
+    pub fn enable_middle_out_transform(&self) {
```
michaelneale: I don't think this should be in the agent, is too specific to openrouter at the moment

crates/goose/src/agents/agent.rs
```diff
@@ -152,6 +155,7 @@ impl Agent {
             router_tool_selector: Mutex::new(None),
             scheduler_service: Mutex::new(None),
             retry_manager,
+            token_tracker: create_shared_tracker(),
```
michaelneale: slightly worried about keeping state here - not sure why but something to think about

kwsantiago:
@michaelneale Thank you for the feedback!

The current state of this PR is that it parses error messages after hitting the limit and stores the discovered limit in a token tracker, but doesn't use this to prevent future errors. I had initially added the OpenRouter middle-out feature in here since it was mentioned in the issue comments https://github.com/block/goose/issues/1303#issuecomment-2674681407.

At your suggestion, what I think we can do here is split this out into 2 PRs:
- This PR (core fix for context limit error handling):
  - Keep the error parsing and improved error messaging
  - Add retry logic that automatically enables middle-out for
   OpenRouter after detecting a context limit error (middle-out won't be implemented in this PR)
  - Remove the token tracker state (since it isn't being used for
  prevention)

- Separate PR (OpenRouter enhancements):
  - Move OpenRouter-specific middle-out configuration to the
  provider's `from_env` method
  - Add `OPENROUTER_ENABLE_MIDDLE_OUT` env var support
  - Implement proactive token counting if needed

I'll begin working on this, but if you have any other feedback, please let me know.

kwsantiago:
@michaelneale I figured what I had was too complex for solving this specific issue and better left as separate PRs (one PR for enhancing context limit error handling and another PR for the OpenRouter middle-out support).

Let me know if you approve of the simple change here to solve https://github.com/block/goose/issues/1303 and I can work on the other PRs using this one as a base.

crates/goose-cli/src/session/mod.rs
```diff
@@ -1143,6 +1143,32 @@ impl Session {
...
+                                    // Restart the stream to trigger the normal context handling flow
+                                    stream = self
+                                        .agent
+                                        .reply(
+                                            &self.messages,
+                                            session_config.clone(),
+                                            None
+                                        )
+                                        .await?;
+                                    continue;
```
DOsinga: doesn't this cause an infinite loop? the message that caused the context exceeded is still in the messages, no?

kwsantiago:
Sounds good @DOsinga I made a thread in Discord to discuss. https://discord.com/channels/1287729918100246654/1344318255266795641/1399727098574143508

I'm turning this PR into a draft until we get concrete direction on how we want to move forward with this fix.

crates/goose-cli/src/session/mod.rs
kwsantiago: Thank you for catching! I added a commit to fix.

kwsantiago:
@DOsinga let me know what you think. Based on your [proposed solution from the Discord thread](https://discord.com/channels/1287729918100246654/1399727098574143508/1399775229353529506), I added an initial implementation for automatic recovery when tool responses are too big and hit context limits. In this way, Goose recovers automatically instead of dying silently when tools return huge responses.

DOsinga:
I think @katzdave should have a look how this fits in with his efforts

DOsinga:
Ok, so a previous attempt to unify http clients was this:
https://github.com/block/goose/pull/3547

I'm currently working on this:
https://github.com/block/goose/pull/3558

to unify some other stuff, but was about to copy the client sharing from the firrst (closed) PR. I'm thinking we should get that in and that should make your life easier. I think having a fixed timeout for now would be fine actually for now. I imagine we replace all the the environment variables with true settings soon anyway.

_ceterum censeo variabiles ambientis esse delendas_

kwsantiago:
Sounds good to me @DOsinga let me know when we're in a good spot and I can finish this up in a way that makes sense to all stakeholders. Thank you!

DOsinga:
thank you for your efforts!

michaelneale:
@kwsantiago very nice looking focussed - some changes merged recently do you want to update this to main? (I could also push to it if you like). Curious if this helps with the openrouter context exhaustion.

I think we are in a better spot now to work on this. 

On openrouter - is there any downside to having middle-out on by default? (ideally we wouldn't need it, but if it doesn't kick in until it absolutely has to then why not).

michaelneale:
I think this looks good - has tests, and if we can update it to main and check it is ok, lets get it in. For open router I am thinking of adding in middle-out by default (why not) anyway.

kwsantiago:
@michaelneale sounds good, if you don't get to it, I can do this tonight (EST).

Regarding open router middle-out support, should I resurrect that in a follow up PR or within this PR? I know we had originally discussed taking it out which is why I got rid of it in this branch, although it is in some of my earlier commits if we want to dig it back out.

michaelneale:
@kwsantiago yeah if you want to add a middle-out in another PR - go ahead (I have an open PR which enables some defaults for startup experience so could always do it there) - if you have time, go ahead and tag me

kwsantiago:
@michaelneale I merged main into this branch, CI "should" pass (at least it does locally).

Also regarding your [PR branch](https://github.com/block/goose/pull/3881), I'll take a look and fork it to add those changes for OpenRouter middle-out support.

crates/goose-cli/src/session/mod.rs
```diff
@@ -1218,6 +1218,83 @@ impl Session {
                         }
 
                         Some(Err(e)) => {
+                            // Check if this is a context length error
```
michaelneale: oh - @kwsantiago I think this needs to be in the goose crate - as the GUI will hit this as well won't it? ie not just the cli

michaelneale (changes requested):
hoping we can move it to the session stuff in the core goose crate not CLI

crates/goose-cli/src/session/mod.rs
kwsantiago: @michaelneale as in, move it out completely from CLI->GUI, or simply replicate this in the GUI as well?

michaelneale: if possible - would be ideal in main goose crate, just once, but if not possible, goose-server will need it replicated, but that isn't ideal as it seems session specific code, not client code

kwsantiago: Sounds good I'll work on adding support for this in the GUI as well. Thanks for the feedback!

kwsantiago:
@michaelneale I moved the detection logic to `agents/agent.rs` in my last commit.

Let me know if we need to adjust anything there.

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -2329,33 +2329,28 @@ mod tests {
         assert!(text_editor_tool
             .description
             .as_ref()
-            .map_or(false, |desc| desc
-                .contains("Replace a string in a file with a new string")));
+            .is_some_and(|desc| desc.contains("Replace a string in a file with a new string")));
```
michaelneale: this is just a style/non functional change?

crates/goose-cli/src/session/mod.rs
```diff
@@ -1226,6 +1226,28 @@ impl Session {
                         }
 
                         Some(Err(e)) => {
+                            // Check if this is a context length error that wasn't already handled by the agent
+                            if let Some(goose::providers::errors::ProviderError::ContextLengthExceeded(error_msg)) = e.downcast_ref::<goose::providers::errors::ProviderError>() {
```
michaelneale: not sure this belongs here - if so, how do we show it in GUI? I think could probably do without this, the agent is doing the heavy lifting and I dont' think we want any special provider code in here either - and won't need to say middle-out

michaelneale (changes requested):

crates/goose-mcp/src/developer/mod.rs
kwsantiago: @michaelneale This was from clippy, completely unrelated to this PR, but figured I'd add it in since CI would probably fail there.

Can remove if needed.

crates/goose-cli/src/session/mod.rs
kwsantiago: Makes sense, I'll update to reflect.

michaelneale:
looking good @kwsantiago - are you happy with the tests in the cli module - I can't quite tell if they are testing things there or really are testing things which are (now correctly) in agent module?

kwsantiago:
> are you happy with the tests in the cli module - I can't quite tell if they are testing things there or really are testing things which are (now correctly) in agent module?

Good catch @michaelneale I removed those tests. I thought about adding them to agent but I think it will require a sophisticated mock provider for simulating overflow errors at the right time, and I don't think that's within the scope of this PR. Let me know if you think otherwise.

crates/goose/src/agents/agent.rs
```diff
@@ -1083,11 +1083,60 @@ impl Agent {
...
+                                    // Emit a history replaced event to notify the client
+                                    yield AgentEvent::HistoryReplaced(messages.messages().clone());
+
+                                    // Continue the stream instead of breaking
+                                    continue;
+                                }
+                            } else {
+                                // Not caused by tool response, show the error to user
+                                yield AgentEvent::Message(Message::assistant().with_context_length_exceeded(
+                                    format!("Context length exceeded: {}", error_msg)
```
michaelneale: can we add in "please try a new session" like before @kwsantiago ?

michaelneale: ie similar message to before

michaelneale:
I think looks ok @katzdave would like your 👍 to merge this (maybe could tweak the fallback error message) to make sure it doesn't flag anything for you

michaelneale:
@katzdave ok with me - if you want to give a 👍

zanesq:
Still in progress?

kwsantiago:
Not that I'm aware of @katzdave , unless you see anything here that needs to be changed.

katzdave:
I think we can probably close this.  https://github.com/block/goose/pull/4202 addresses this out of context issue.

DOsinga:
I think this is the right idea but shouldn't this be implemented in the goos crate rather than in the CLI? @katzdave has been doing some work here too. Now that we have auto-compact, I think we should catch the error in the agent.reply_internal loop and call autocompact

kwsantiago:
@DOsinga @michaelneale @katzdave I've updated to reflect catching the error in the `agent.reply_internal` loop and calling autocompact.

Please let me know if any other changes should be made here.

crates/goose/src/agents/agent.rs
```diff
@@ -1052,11 +1052,33 @@ impl Agent {
...
-                                    "The context length of the model has been exceeded. Please start a new session and try again.",
-                                ));
-                            break;
+                        Err(ProviderError::ContextLengthExceeded(error_msg)) => {
+                            info!("Context length exceeded, attempting auto-compaction");
+
+                            match auto_compact::check_and_compact_messages(
+                                self,
+                                &messages,
+                                None,
```
katzdave: Since we don't pass in session_metadata this will lead to token estimates. But here we want to always try compacting.

Can we break out the meat of the logic preserving the last user message + tracking the summarization result /metadata(but skipping the check) in check_and_compact_messages -> compact_messages. Then check_and_compact_messages can call this.

crates/goose/src/agents/agent.rs
```diff
@@ -1052,11 +1052,33 @@ impl Agent {
...
+                                            "Context limit reached. Conversation has been automatically compacted to continue."
+                                        )
+                                    );
+                                    yield AgentEvent::HistoryReplaced(messages.clone());
+
+                                    continue;
+                                }
+                                _ => {
+                                    yield AgentEvent::Message(Message::assistant().with_context_length_exceeded(
+                                        format!("Context length exceeded: {}. Unable to continue.", error_msg)
```
katzdave: Context length exceeded and cannot summarize: {}. Unable to continue.

michaelneale:
this would be good to finally get addressed

crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -66,7 +66,7 @@ pub async fn check_compaction_needed(
     });
 
     let provider = agent.provider().await?;
-    let context_limit = provider.get_model_config().context_limit();
+    let context_limit = estimate_target_context_limit(provider);
```
katzdave: Don't use this function; leave as the true context limit. I'm looking to phase that out completely and update the truncation flow.

katzdave (commented):
Gave this some testing on cli:

Seems like I hit the context error when the agent pulls my last file read:

```
Error: Request failed: HTTP 400 Bad Request: {"error_code":"BAD_REQUEST","message":"{\"external_model_provider\":\"anthropic\",\"external_model_error\":{\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",\"message\":\"input length and `max_tokens` exceed context limit: 196758 + 8192 > 200000, decrease input length or `max_tokens` and try again\"},\"request_id\":\"req_011CSjw2Yhikd24PTTPDVY6s\"}}"}
```

on desktop:

```
13:04:26.166 › goosed stderr for port 49736 and dir /Users/dkatz/git/goose3/goose:   2025-09-02T17:04:26.166432Z  INFO goosed::routes::reply: Session started, counter.goose.session_starts: 1, session_type: "app", interface: "ui"
    at crates/goose-server/src/routes/reply.rs:179


13:04:26.213 › from renderer: Displaying content for message: encjh7cz content: hi keep looking around
13:04:26.213 › from renderer: Displaying content for message: encjh7cz content: hi keep looking around
13:04:27.840 › goosed stderr for port 49736 and dir /Users/dkatz/git/goose3/goose:   2025-09-02T17:04:27.839627Z ERROR goosed::routes::reply: Error processing message: Request failed: HTTP 400 Bad Request: {"error_code":"BAD_REQUEST","message":"{\"external_model_provider\":\"anthropic\",\"external_model_error\":{\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",\"message\":\"input length and `max_tokens` exceed context limit: 196869 + 8192 > 200000, decrease input length or `max_tokens` and try again\"},\"request_id\":\"req_011CSjwJNaxUPqb88BtnhU34\"}}"}
    at crates/goose-server/src/routes/reply.rs:303


13:04:27.842 › goosed stderr for port 49736 and dir /Users/dkatz/git/goose3/goose:   2025-09-02T17:04:27.840557Z  INFO goosed::routes::reply: Session completed, counter.goose.session_completions: 1, session_type: "app", interface: "ui", exit_type: "normal", duration_ms: 1673, total_tokens: 0, message_count: 385
    at crates/goose-server/src/routes/reply.rs:353

  2025-09-02T17:04:27.840686Z  INFO goosed::routes::reply: Session duration, counter.goose.session_duration_ms: 1673, session_type: "app", interface: "ui"
    at crates/goose-server/src/routes/reply.rs:364
```

But doesn't actually trigger this new forced compaction flow. Was using databricks provider for this testing, and I believe that message does pass the check to trigger contextLengthExceeded error.

Is this working for you on a different provider?

kwsantiago:
@katzdave Thanks for the detailed feedback! I've addressed the issues:

1. Reverted to using true context limit instead of `estimate_target_context_limit` as requested
2. Added fallback detection for context errors in message text to handle databricks' nested error format
3. The fix should now catch errors like "input length and max_tokens exceed context limit" and trigger auto-compaction

I don't have access to test with actual providers (databricks/openrouter), so I'd appreciate if someone could verify the fix works end-to-end. I won't have the bandwidth to iterate much further on this, but the changes should handle the reported issues. Feel free to edit the branch as needed.

michaelneale:
@katzdave both @zanesq and I see that databricks error too, unrelated to this change, so it may be an unrelated regression?

michaelneale:
@kwsantiago can you run fmt again?

kwsantiago:
@michaelneale I just committed cargo fmt.

michaelneale:
@kwsantiago sorry again - but can you update to main, hopefully will be last time. Will try this now, I think this may be good.

crates/goose/src/providers/utils.rs
```diff
@@ -163,6 +163,18 @@ pub async fn handle_status_openai_compat(response: Response) -> Result<Response,
                 )));
             }
         }
+
+        // Fallback: Check if the raw body contains context limit patterns (for databricks and other providers)
+        let body_lower = body_str.to_lowercase();
+        if body_lower.contains("context limit")
+            || body_lower.contains("context_length_exceeded")
+            || (body_lower.contains("input length") && body_lower.contains("exceed"))
```
michaelneale: wonder if could add `Input is too long` or even just `Input` - and ignore case? (I might try this)

kwsantiago: @michaelneale I pushed up a commit for this, try it out.

michaelneale (approved):
nice @kwsantiago - took a while but got there in the end!

crates/goose/src/providers/utils.rs
```diff
@@ -57,11 +57,24 @@ fn check_context_length_exceeded(text: &str) -> bool {
         "max_tokens",
         "decrease input length",
         "context limit",
+        "input is too long",
```
katzdave: Shouldn't need this? too long is above in check phrases already.

crates/goose/src/providers/utils.rs
```diff
@@ -57,11 +57,24 @@ fn check_context_length_exceeded(text: &str) -> bool {
...
+    // Check for specific phrases
+    if check_phrases
         .iter()
         .any(|phrase| text_lower.contains(phrase))
+    {
+        return true;
+    }
+
+    // Check for "input" + "too long" pattern
+    if text_lower.contains("input") && text_lower.contains("too long") {
```
katzdave: I don't think we need this either, or any change in this fn?

crates/goose/src/providers/utils.rs
```diff
@@ -163,6 +176,20 @@ pub async fn handle_status_openai_compat(response: Response) -> Result<Response,
                 )));
             }
         }
+
+        // Fallback: Check if the raw body contains context limit patterns (for databricks and other providers)
+        let body_lower = body_str.to_lowercase();
```
katzdave: I dug deeper and it seems the issue with databricks is that it's using streaming not that we aren't able to parse the error, so we don't need this either. Made a separate fix specifically for that. https://github.com/block/goose/pull/4506

katzdave (commented):
Yes, nice work! I think we can revert the changes in errors.rs completely; taking it for one last spin quickly will update shortly.

katzdave:
> Yes, nice work! I think we can revert the changes in utils/errors.rs completely; taking it for one last spin quickly will update shortly.

Confirming lets completely revert utils/errors.rs and then all set to merge.

kwsantiago:
@katzdave @michaelneale should be good to go, let me know if there is anything else here. Appreciate all the feedback!

michaelneale:
@kwsantiago did you mean to remove more than just error.rs changes? as looks a lot smaller now, as there were utils.rs changes which I think were needed? (a force push so can't see the history)

edit: yes you did, ignore me! nice!

kwsantiago:
@michaelneale that was due to addressing @katzdave requested changes here:

- https://github.com/block/goose/pull/3635#pullrequestreview-3183001543
- https://github.com/block/goose/pull/3635#discussion_r2320550389
- https://github.com/block/goose/pull/3635#discussion_r2320552269
- https://github.com/block/goose/pull/3635#discussion_r2320554968
- https://github.com/block/goose/pull/3635#issuecomment-3251302797

michaelneale (approved):
nice, thanks! took it for quick spin

Issue #3556, opened 2025-07-21
Sub-recipes can't resolve relative paths


ishii1648:
**Describe the bug**

Goose can't resolve relative paths for sub-recipes.

**To Reproduce**
1. Create a directory structure as follows:

```
❯ tree
.
├── recipe.yaml
└── sub-recipes
    └── terraform-init.yaml
```

2. Create `recipe.yaml` with relative paths to sub-recipes:
```yaml
sub_recipes:
  - name: terraform init
    path: "./sub-recipes/terraform-init.yaml"
```

4. Attempt to use recipe.yaml.
5. Goose fails to resolve the relative paths specified for the sub_recipes as follows.

```console
$ goose run --recipe $REPO_PATH/terraform-plan/recipe.yaml
thread 'main' panicked at crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs:13:53:
called `Result::unwrap()` on an `Err` value: Failed to read recipe file sub-recipes/terraform-init.yaml: No such file or directory (os error 2)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**Expected behavior**

Goose can resolve relative paths for sub-recipes from main recipe.

**Please provide following information:**
 - **OS & Arch:** MacOS arm64
 - **Interface:** CLI
 - **Version:** v1.1.3

**Additional context**

ishii1648:
Thank you!!

PR #3642, opened 2025-07-25
Resolve sub recipe path relative to the parent recipe path

lifeizhou-ap:
Implemented for [Issue 3556](https://github.com/block/goose/issues/3556)

### DIFF
--- a/crates/goose-cli/src/recipes/extract_from_cli.rs
+++ b/crates/goose-cli/src/recipes/extract_from_cli.rs
@@ -117,7 +117,13 @@ mod tests {
         assert!(sub_recipes.is_some());
         let sub_recipes = sub_recipes.unwrap();
         assert!(sub_recipes.len() == 1);
-        assert_eq!(sub_recipes[0].path, "existing_sub_recipe.yaml".to_string());
+        let full_sub_recipe_path = recipe_path
+            .parent()
+            .unwrap()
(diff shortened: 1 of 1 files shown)


crates/goose/src/recipe/build_recipe/mod.rs
```diff
@@ -76,8 +78,18 @@ where
         });
     }
 
-    let recipe = Recipe::from_content(&rendered_content)
+    let mut recipe = Recipe::from_content(&rendered_content)
         .map_err(|source| RecipeError::RecipeParsing { source })?;
+
+    if let Some(ref mut sub_recipes) = recipe.sub_recipes {
+        for sub_recipe in sub_recipes {
+            if let Ok(resolved_path) = resolve_sub_recipe_path(&sub_recipe.path, &recipe_parent_dir)
```
DOsinga, line 86: not sure when this would fail, but if it did shouldn't we fail the entire operation rather than have a subrecipe that has an unresolved path?

DOsinga (approved):
thanks!

crates/goose/src/recipe/build_recipe/mod.rs
lifeizhou-ap: The entire operation will fail when a subrecipe has an unresolved path as it throws RecipeParsing error

PR #3651, opened 2025-07-25
fix: agent loop logic to avoid repeating responses

zanesq:
# Fix: Resolve agent spinning with repeated responses during discovery

## Problem
Users reported that chats were spinning with the agent replying with repeated answers like "I found the issue" over and over again, doing similar tasks but slightly worded differently until eventually finishing. This behavior was new and didn't occur before recent changes.

## Root Cause
The issue was caused by retry logic being triggered inappropriately during normal chat operations. The retry logic was being called **every time** the agent loop didn't add a message (`!added_message` was true), which happens frequently in normal operation when the agent is thinking, waiting for responses, or has no tool calls to process.

This caused the agent to repeatedly restart its loop even in normal scenarios, leading to the "spinning" behavior with repeated similar responses.

## Solution
Added proper conditions for when retry logic should be invoked:

1. **Conditional Retry Logic**: Only run retry logic when:
   - There's a session with retry configuration
   - A final output tool exists and has been called (indicating task completion)
   - We're not in the middle of normal processing

2. **Improved Final Output Tool Logic**: Made the logic more defensive by checking if a final output tool exists before using it and avoiding multiple mutex locks.

3. **Enhanced Logging**: Added debug logging to help track when retry logic is being evaluated and triggered.

## Changes
- **`crates/goose/src/agents/agent.rs`**: Fixed agent loop to only trigger retry logic when appropriate
- **`crates/goose/src/agents/retry.rs`**: Added detailed debug logging for retry operations

## Testing
- All existing tests pass
- Retry functionality preserved for legitimate use cases (recipes with retry configurations)
- Normal chat sessions without retry configurations are unaffected

## Impact
- ✅ Fixes spinning/repeated responses during normal chat
- ✅ Preserves retry functionality for recipes
- ✅ Maintains backward compatibility
- ✅ Improves debugging capabilities

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -1004,32 +1004,67 @@ impl Agent {
                     (tools, toolshim_tools, system_prompt) = self.prepare_tools_and_prompt().await?;
                 }
                 if !added_message {
-                    if let Some(final_output_tool) = self.final_output_tool.lock().await.as_ref() {
-                        if final_output_tool.final_output.is_none() {
+                    // Handle final output tool logic if one is configured
+                    let final_output_tool_guard = self.final_output_tool.lock().await;
+                    let has_final_output_tool = final_output_tool_guard.is_some();
+                    let final_output_completed = final_output_tool_guard
+                        .as_ref()
+                        .map(|tool| tool.final_output.is_some())
+                        .unwrap_or(false);
+                    drop(final_output_tool_guard);
+
+                    if has_final_output_tool {
+                        if !final_output_completed {
                             tracing::warn!("Final output tool has not been called yet. Continuing agent loop.");
                             let message = Message::user().with_text(FINAL_OUTPUT_CONTINUATION_MESSAGE);
                             messages_to_add.push(message.clone());
                             yield AgentEvent::Message(message);
                             continue
                         } else {
-                            let message = Message::assistant().with_text(final_output_tool.final_output.clone().unwrap());
+                            let final_output_tool_guard = self.final_output_tool.lock().await;
+                            let final_output = final_output_tool_guard
+                                .as_ref()
+                                .and_then(|tool| tool.final_output.clone())
+                                .unwrap_or_default();
+                            drop(final_output_tool_guard);
+                            
+                            let message = Message::assistant().with_text(final_output);
                             messages_to_add.push(message.clone());
                             yield AgentEvent::Message(message);
                         }
                     }
 
-                    match self.handle_retry_logic(&mut messages, &session, &initial_messages).await {
-                        Ok(should_retry) => {
-                            if should_retry {
-                                info!("Retry logic triggered, restarting agent loop");
-                                continue;
+                    // Only run retry logic if:
+                    // 1. There's a session with retry configuration
+                    // 2. We have a final output tool that has been called (indicating completion)
+                    // 3. We're not in the middle of normal processing
+                    let should_run_retry = session
+                        .as_ref()
+                        .and_then(|s| s.retry_config.as_ref())
+                        .is_some()
+                        && has_final_output_tool
+                        && final_output_completed;
+
+                    debug!("Retry logic evaluation: should_run_retry={}, has_session_retry_config={}, has_final_output_tool={}, final_output_completed={}", 
+                           should_run_retry,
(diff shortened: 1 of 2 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -1004,32 +1004,67 @@ impl Agent {
...
-                    match self.handle_retry_logic(&mut messages, &session, &initial_messages).await {
-                        Ok(should_retry) => {
-                            if should_retry {
-                                info!("Retry logic triggered, restarting agent loop");
-                                continue;
+                    // Only run retry logic if:
+                    // 1. There's a session with retry configuration
+                    // 2. We have a final output tool that has been called (indicating completion)
+                    // 3. We're not in the middle of normal processing
+                    let should_run_retry = session
```
cloud-on-prem, line 1041: This could be in a `fn should_execute_retry_logic` and that way we can unit-test the logic.

cloud-on-prem (approved):

cloud-on-prem:
Just wondering why the agent was retrying when there's no explicit retry config.
@zanesq - do you have access to the user report - would love to reproduce it.

AFAIK, the checks here should prevent retries when there's no retry config. 
https://github.com/block/goose/blob/main/crates/goose/src/agents/retry.rs#L117-L123

crates/goose/src/agents/agent.rs
```diff
@@ -1004,32 +1004,67 @@ impl Agent {
...
                         } else {
-                            let message = Message::assistant().with_text(final_output_tool.final_output.clone().unwrap());
+                            let final_output_tool_guard = self.final_output_tool.lock().await;
+                            let final_output = final_output_tool_guard
+                                .as_ref()
+                                .and_then(|tool| tool.final_output.clone())
+                                .unwrap_or_default();
+                            drop(final_output_tool_guard);
+                            
+                            let message = Message::assistant().with_text(final_output);
```
jsibbison-square, line 1031: All these changes just seem to do the same logic (extract the final output if the final outpu tool exists). Rather than duplicating that I would revert these changes and just do another extraction below. Easier to read imo.

crates/goose/src/agents/agent.rs
```diff
@@ -1004,32 +1004,67 @@ impl Agent {
...
+                    // Only run retry logic if:
+                    // 1. There's a session with retry configuration
+                    // 2. We have a final output tool that has been called (indicating completion)
+                    // 3. We're not in the middle of normal processing
+                    let should_run_retry = session
+                        .as_ref()
+                        .and_then(|s| s.retry_config.as_ref())
+                        .is_some()
+                        && has_final_output_tool
+                        && final_output_completed;
```
jsibbison-square, line 1046: This logic is wrong. We can do retries without a final output tool. They are independant features.

When they both exist final_output check happens first. If final output check fails it has a 'continue' so the loop is restarted before this logic happens so we don't need to check that it's completed.

The logic should be that if final output tool is not complete then the loop needs to happen again (the early continue). Retry won't be invoked. Otherwise if retry_config exists then retry check should be run.

crates/goose/src/agents/agent.rs
```diff
@@ -1004,32 +1004,67 @@ impl Agent {
...
 
-                    match self.handle_retry_logic(&mut messages, &session, &initial_messages).await {
-                        Ok(should_retry) => {
-                            if should_retry {
-                                info!("Retry logic triggered, restarting agent loop");
-                                continue;
+                    // Only run retry logic if:
+                    // 1. There's a session with retry configuration
+                    // 2. We have a final output tool that has been called (indicating completion)
+                    // 3. We're not in the middle of normal processing
```
jsibbison-square, line 1040: I can't see any logic checks related to point 3 here, or is that the !added_message at the start?

jsibbison-square (changes requested):
The logic for when to run retries is wrong as it can be used as a feature even when final_output is not enabled.

jsibbison-square:
I don't think these changes will fix the described issue, it sounds more like normal agent self reflection chatter. I'd rather not add this code until we can reproduce the issue as this change duplicates logic checks that are already in the code and makes it harder to read.

Please add a way to reproduce so we can explore further.

zanesq:
Yeah I just asked goose to try and fix it and thought maybe it could spur some ideas about whats happening. Haven't had a chance to verify it either.

Issue #3578, opened 2025-07-22
Adopt RMCP


jamadeo:
Goose uses its own implementation of MCP. We should instead adopt the [rmcp](https://crates.io/crates/rmcp) crate, which has become the official Rust SDK for the Model Context Protocol https://github.com/modelcontextprotocol/rust-sdk

First remove/replace reference to types in the `mcp-core` create, eventually removing it entirely, and move our MCP client implementation to RMCP

## Plan

Migrate away from types in the `mcp-core` create, eventually removing it entirely, and move our MCP client implementation to RMCP. We will migrate in four steps:

- [ ] (In Progress) **Step I: Type Migration**. Update all existing code referencing data types to use the rmcp equivalents of these types. These can all be merged as standalone PRs and are relatively easier to verify/ensure safety. An analysis of the type mappings has been completed and is below.
- [ ] (In Progress) **Step 2: Update client/server tooling**. This will be feature flagged and can likely also go in in a few different PRs
- [ ] **Step 3: Cleanup.** Remove old MCP crates and “feature flag off” code, leaving only an integration with RMCP
- [ ] **Step 4: New feature UX**. Now that we’ll have an MCP stack with all features supported, we can develop UX for things like prompts/resources/sampling/elicitation

## Type Conversion Reference

mcp_core Type | rmcp::model Equivalent | Migration Difficulty | Notes | Status
-- | -- | -- | -- | --
Content | Content (Annotated) | Medium | rmcp uses Annotated wrapper pattern | Done
TextContent | TextContent (Annotated) | Medium | Structure changed, uses annotation system | Done
ImageContent | ImageContent (Annotated) | Medium | Structure changed, uses annotation system | Done
EmbeddedResource | EmbeddedResource (Annotated) | Medium | Structure changed, uses annotation system | Done
Annotations | Annotations | Easy | Very similar structure | Done
Tool | Tool | Hard | Significant API changes, different schema handling | Done
ToolCall | CallToolRequestParam | Medium | Different naming and structure | Not Started
ToolAnnotations | ToolAnnotations | Medium | Similar but with different defaults and API | Done
ToolError | No direct equivalent | Hard | Need to use rmcp::Error instead | Not Started
ToolResult<T> | Result<T, rmcp::Error> | Medium | Standard Result type in rmcp | Not Started
Resource | Resource (Annotated) | Medium | Uses annotation pattern, added size field | Done
ResourceContents | ResourceContents | Easy | Nearly identical structure | Done
[...]

PR #3653, opened 2025-07-25
chore: use typed notifications from rmcp

jamadeo:
Swaps generic JsonRpcMessage use with typed ServerNotifications from rmcp.

for #3578


crates/mcp-client/src/client.rs
```diff
@@ -148,9 +149,22 @@ where
                             }) => {
                                 service_ptr.respond(&id.to_string(), Ok(message)).await;
                             }
-                            _ => {
+                            JsonRpcMessage::Notification(JsonRpcNotification {
+                                notification,
+                                ..
+                            }) => {
                                 let mut subs = subscribers_ptr.lock().await;
-                                subs.retain(|sub| sub.try_send(message.clone()).is_ok());
```
jamadeo, line 153: Probably the only change of any significance -- here we'd just forward all "other" messages to notifications subscribers, without checking that they're necessarily notifications. Then on the other end, we'd ignore them because they aren't the types we're looking for. So the result should be the same, but worth noting.

alexhancock (approved):

PR #3656, opened 2025-07-25
Add GOOSE_MAX_TOKENS as model configurable

bkvarda:
From what I could tell ModelConfig accepts a max_tokens argument but there wasn't any global way to configure it. 

Setting `GOOSE_MAX_TOKENS` in `config.yaml` or as an env variable to 24000 yields model config like this:

```
{"timestamp":"2025-07-25T17:16:08.000742Z","level":"DEBUG","fields":{"model_config":"{\n  \"model_name\": \"claude-4-sonnet\",\n  \"context_limit\": 200000,\n  \"temperature\": null,\n  \"max_tokens\": 24000,\n  \"toolshim\": false,\n  \"toolshim_model\": null\n}","input":"{\n  \"max_tokens\": 24000,\n  \"messages\": [\n    {\n      \"content\":
```

Leaving it unset yields this as it does today:
```
,"fields":{"model_config":"{\n  \"model_name\": \"claude-4-sonnet\",\n  \"context_limit\": 200000,\n  \"temperature\": null,\n  \"max_tokens\": null,\n  \"toolshim\": false,\n  \"toolshim_model\": null\n}","input":"{\n  \"messages\": [\n    {\n      \"content\": \
```
This should help some providers w/ server-side low default max_tokens on requests (like Databricks) have better perf when used w/ Goose and provide more control in general

### DIFF
--- a/crates/goose/src/macros.rs
+++ b/crates/goose/src/macros.rs
@@ -3,13 +3,9 @@ macro_rules! impl_provider_default {
     ($provider:ty) => {
         impl Default for $provider {
             fn default() -> Self {
-                let model = $crate::model::ModelConfig::new(
+                let model = $crate::model::ModelConfig::new_or_fail(
                     &<$provider as $crate::providers::base::Provider>::metadata().default_model,
-                )
-                .expect(concat!(
-                    "Failed to create model config for ",
-                    stringify!($provider)
-                ));
+                );
 
                 <$provider>::from_env(model)
                     .expect(concat!("Failed to initialize ", stringify!($provider)))
--- a/crates/goose/src/model.rs
+++ b/crates/goose/src/model.rs
@@ -114,7 +114,7 @@ impl ModelConfig {
         custom_env_var: Option<&str>,
     ) -> Result<Option<usize>, ConfigError> {
         let config = crate::config::Config::global();
-        
+
         // Try custom env var first
         if let Some(env_var) = custom_env_var {
             match config.get_param::<String>(env_var) {
@@ -169,7 +169,7 @@ impl ModelConfig {
                 }
             }
         }
-        
+
         // Try default GOOSE_CONTEXT_LIMIT
         match config.get_param::<String>("GOOSE_CONTEXT_LIMIT") {
             Ok(val) => {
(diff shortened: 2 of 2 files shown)


bkvarda:
Also changed some of the other things that were configured using `std::env::var` to instead use `config.get_param()` since that looks for both env variable and in config.

DOsinga:
ugh, sorry - merge conflict, can you resolve?

DOsinga:
hey @bkvarda - mind have a look and rebase? thank you!

bkvarda:
@DOsinga i attempted a rebase and fixing conflicts, hopefully this generally keeps the spirt of all the changes - adding max tokens config, using global config for both config and env variables instead of env variables directly, and doing validation checks.

crates/goose/src/macros.rs
```diff
@@ -3,13 +3,9 @@ macro_rules! impl_provider_default {
     ($provider:ty) => {
         impl Default for $provider {
             fn default() -> Self {
-                let model = $crate::model::ModelConfig::new(
+                let model = $crate::model::ModelConfig::new_or_fail(
                     &<$provider as $crate::providers::base::Provider>::metadata().default_model,
-                )
-                .expect(concat!(
```
DOsinga, line 9: we should die on this

crates/goose/src/model.rs
```diff
@@ -85,20 +85,25 @@ impl ModelConfig {
         Self::new_with_context_env(model_name.to_string(), None)
     }
 
+    /// Create a new ModelConfig with the specified model name and custom context limit env var
+    ///
+    /// This is useful for specific model purposes like lead, worker, planner models
+    /// that may have their own context limit environment variables.
```
DOsinga, line 91: remove

crates/goose/src/model.rs
```diff
@@ -108,75 +113,282 @@ impl ModelConfig {
         model_name: &str,
         custom_env_var: Option<&str>,
     ) -> Result<Option<usize>, ConfigError> {
+        let config = crate::config::Config::global();
+
+        // Try custom env var first
```
DOsinga, line 118: why do we need to change all this?

crates/goose/src/model.rs
```diff
@@ -236,8 +448,23 @@ impl ModelConfig {
...
+            );
+            ModelConfig {
+                model_name: model_name.to_string(),
+                context_limit: Self::get_model_specific_limit(model_name),
+                temperature: None,
+                max_tokens: None,
+                toolshim: false,
+                toolshim_model: None,
+            }
+        })
```
DOsinga, line 467: eh this is called new_or_fail. it should do that.

DOsinga (changes requested):
I'm confused about this PR, it seems to do a whole lot of things that don't seem to be needed for the feature it is trying to provide

zanesq:
Still in progress?

DOsinga:
going to close this for now due to non activity

PR #3659, opened 2025-07-25
Remove confusing status output when testing sharing url connection and it shows 404

zanesq:
It was confusing to report connection successful but show a 404 status in the ui.

No need to output the response status here, we only want to check if the server is responding with a non 500 for sharing purposes.


katzdave (approved):
Nice!

PR #3660, opened 2025-07-25
Fix conversations before they hit the LLM

DOsinga:
try to get a conversations into shape before we send them to the LLM

### DIFF
--- a/crates/goose/src/conversation_fixer.rs
+++ b/crates/goose/src/conversation_fixer.rs
@@ -14,12 +14,14 @@ impl ConversationFixer {
         let (messages, tool_calling_fixed) = Self::fix_tool_calling(messages);
         let (messages, messages_merged) = Self::merge_consecutive_messages(messages);
         let (messages, lead_trail_fixed) = Self::fix_lead_trail(messages);
+        let (messages, populated_if_empty) = Self::populate_if_empty(messages);
 
         let mut issues = Vec::new();
         issues.extend(empty_removed);
         issues.extend(tool_calling_fixed);
         issues.extend(messages_merged);
         issues.extend(lead_trail_fixed);
+        issues.extend(populated_if_empty);
 
         (messages, issues)
     }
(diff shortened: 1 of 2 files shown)


crates/goose/src/conversation_fixer.rs
```diff
@@ -0,0 +1,371 @@
...
+        if let Some(last) = messages.last() {
+            if last.role == Role::Assistant {
+                messages.pop();
+                issues.push("Removed trailing assistant message".to_string());
+            }
+        }
+
+        if messages.is_empty() {
+            issues.push("Added placeholder user message to empty conversation".to_string());
+            messages.push(Message::user().with_text(PLACEHOLDER_USER_MESSAGE));
```
alexhancock, line 189: Is this added in this method because it's the last method we call of the four from `fix_conversation`? Wondering if we could simply run this logic as part of that method instead

alexhancock (approved):

crates/goose/src/conversation_fixer.rs
DOsinga: the thinking is that if you have merged messages with the same role and then you've stripped any assistant messages from beginning and end, that you now have a conversation that starts with a user message and ends with one, unless we have no messages left. would you like me to call this method differently?

alexhancock: I was just thinking in terms of https://en.wikipedia.org/wiki/Principle_of_least_astonishment `fix_lead_trail` would not also do this. I was thinking instead we'd do it as a separate method similar to the others

```
let (messages, empty_removed) = Self::remove_empty_messages(messages);
let (messages, tool_calling_fixed) = Self::fix_tool_calling(messages);
let (messages, messages_merged) = Self::merge_consecutive_messages(messages);
let (messages, lead_trail_fixed) = Self::fix_lead_trail(messages);
let (messages, populated_if_empty) = Self::populate_if_empty(messages);
```

Where `populate_if_empty` does just this part

```
if messages.is_empty() {
  issues.push("Added placeholder user message to empty conversation".to_string());
  messages.push(Message::user().with_text(PLACEHOLDER_USER_MESSAGE));
}
```

But it's not a huge difference! Just posting this to explain my thought

I think the branch has a ton of value as-is and should merge unless you prefer to also do this suggestion

alexhancock (approved):

DOsinga:
yeah I like it. will make it so

PR #3662, opened 2025-07-25
Speed up recipe loading from deeplinks and various fixes

zanesq:
- Users reported some recipe loading from deeplinks was freezing the ui up to 30 seconds for the recipes to show. Moved the recipe decoding to a background process to let the app continue loading faster.

- Fixed issue where we weren't rendering the `:message` pills with markdown and also created a dedicated area when editing recipes for adding multiline message markdown content.

- Fixed issue where the splash screen was sometimes showing the animated greeting message from home/hub if there were no activities.

- Fixed issue where the recipe install warning was not allowing scrolling for long recipe content. Also added markdown rendering in the install warning.

- Improved loading of recipe page recipes using parallelizing rather than waiting on them to finish one by one.

**Bonus**: added a handy script for resetting the osx deeplink associations when testing locally.

Screenshot 2025-07-25 at 12 52 25 PM
Screenshot 2025-07-25 at 12 52 56 PM

### DIFF
--- a/ui/desktop/src/components/settings/sessions/SessionSharingSection.tsx
+++ b/ui/desktop/src/components/settings/sessions/SessionSharingSection.tsx
@@ -118,7 +118,7 @@ export default function SessionSharingSection() {
       if (response.status < 500) {
         setTestResult({
           status: 'success',
-          message: `Connection successful! Server responded with status ${response.status}.`,
+          message: 'Connection successful!',
         });
       } else {
         setTestResult({


ui/desktop/scripts/unregister-deeplink-protocols.js
```diff
@@ -0,0 +1,95 @@
...
+
+const PROTOCOL = 'goose';
+
+function unregisterAllProtocolHandlers() {
+  console.log('Unregistering ALL goose:// protocol handlers...');
+  
+  try {
+    // Get all registered Goose apps
+    console.log('Finding all registered Goose applications...');
+    const lsregisterOutput = execSync(`/System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/LaunchServices.framework/Versions/A/Support/lsregister -dump | grep -B 10 -A 10 "claimed schemes:.*${PROTOCOL}:"`, { encoding: 'utf8' });
```
DOsinga, line 18: this looks like it will only work on mac...

ui/desktop/src/recipe/recipeStorage.ts
```diff
@@ -183,19 +181,34 @@ export async function listSavedRecipes(includeArchived: boolean = false): Promis
...
+    // Wait for all recipes to load in parallel
+    const [globalRecipes, localRecipes] = await Promise.all([
+      Promise.all(globalRecipePromises),
+      Promise.all(localRecipePromises),
+    ]);
+
+    // Filter out null results and apply archived filter
+    const recipes: SavedRecipe[] = [];
+
+    for (const recipe of globalRecipes) {
```
DOsinga, line 205: should we take this as a reason to remove the local recipes? it doesn't make sense to me

ui/desktop/scripts/unregister-deeplink-protocols.js
zanesq: yup comment at top says only mac for now

ui/desktop/src/recipe/recipeStorage.ts
zanesq: good question I'm on the fence about that I think @AaronGoldsmith had a good reason for adding local and I think the idea is you could have recipes that are specific to a project folder/workspace which might come in handy when we have a concept of "Projects"

AaronGoldsmith: My thinking, for local recipes was that it follows how memory and .goosehints work. Both have global/local versions which can be used by Goose.  Devs can decide whether they want to add a .goosehints and memory files at the repository level, which creates a more consistent experience/development environment for Goose users. 

I envision being able to clone a repository, open it in Goose, and find a list established recipes that demonstrate how to interact or get the most use out of a project. 


I'm fine with removing it if it's not getting used.  It _does_ look like there was a recent change in the cli to support "local recipes"
https://github.com/block/goose/blob/main/crates/goose-cli/src/recipes/search_recipe.rs#L127

DOsinga: I think that uses GOOSE_RECIPE_PATH_ENV_VAR maybe? 

I think we talked about this in a different context about how this should be done through the server and not have the client poke around here; I wrote up this doc:

https://docs.google.com/document/d/1vjQn2MtH2XLC9o4OxPkremu2HPdaxwtbbKPzBJoFxpo/edit?tab=t.0#heading=h.ghh7fchtg0dw

on how I would like to see things go. local recipes are not a bad idea per se, but I think we should unify the CLI and the desktop first even if that means dropping the local recipes for now

AaronGoldsmith: Makes sense to prioritize unifying CLI and GUI. I'm fine with local recipes getting removed for now.

zanesq: SGTM I'll follow up with removing local recipes in another PR 👍

PR #3668, opened 2025-07-26
feat: recipe runs will now prompt for missing extension secrets

cloud-on-prem:
## Summary
This change introduces a quality-of-life improvement for running Goose Recipes. The CLI now uses an eager flow to prompt for and securely handle required MCP secrets

## How to test
1. **Download and unzip** the attached [test MCP directory](https://github.com/user-attachments/files/21440800/test-mcp.zip).
2. **Run the following command:**
```sh
# Compile
cargo build
# Run Recipe (with sub-recipes and MCP cred requirements) 
./target/debug/goose run --recipe ../test-mcp/main-recipe.yaml
```
3. Expected Behavior:
  - First run: You should see a prompt for storing secrets.
  - Subsequent runs: The correct secrets should be fetched automatically, and the prompt will not be shown.
  
4. Clean up: Manually remove the test secrets from the OS Keyring (`DATABASE_URL` and `TEST_API_KEY`)

### DIFF
--- a/crates/goose-cli/src/recipes/recipe.rs
+++ b/crates/goose-cli/src/recipes/recipe.rs
@@ -87,24 +87,33 @@ pub fn collect_missing_secrets(requirements: &[SecretRequirement]) -> Result<()>
     }
 
     println!(
-        "🔐 This recipe requires {} secret(s) that are not yet configured:",
+        "🔐 This recipe uses {} secret(s) that are not yet configured (press ESC to skip any that are optional):",
         missing_secrets.len()
     );
 
     for req in &missing_secrets {
         println!("\n📋 Extension: {}", req.extension_name);
-        println!("🔑 Required secret: {}", req.key);
+        println!("🔑 Secret: {}", req.key);
 
-        let value = cliclack::password(format!("Enter {} ({})", req.key, req.description()))
-            .mask('▪')
-            .interact()?;
+        let value = cliclack::password(format!(
+            "Enter {} ({}) - press ESC to skip",
+            req.key,
+            req.description()
+        ))
+        .mask('▪')
+        .interact()
+        .unwrap_or_else(|_| String::new());
 
-        config.set_secret(&req.key, Value::String(value))?;
-        println!("✅ Secret stored securely for {}", req.extension_name);
+        if !value.trim().is_empty() {
+            config.set_secret(&req.key, Value::String(value))?;
+            println!("✅ Secret stored securely for {}", req.extension_name);
+        } else {
+            println!("⏭️  Skipped {} for {}", req.key, req.extension_name);
+        }
     }
 
     if !missing_secrets.is_empty() {
-        println!("\n🎉 All secrets have been configured! Recipe execution will now continue.");
+        println!("\n🎉 Secret collection complete! Recipe execution will now continue.");
     }
 
     Ok(())


crates/goose-cli/src/recipes/recipe.rs
```diff
@@ -44,6 +56,60 @@ pub fn load_recipe(recipe_name: &str, params: Vec<(String, String)>) -> Result<R
...
+            .mask('▪')
+            .interact()?;
+
+        config.set_secret(&req.key, Value::String(value))?;
+        println!("✅ Secret stored securely for {}", req.extension_name);
+    }
+
+    if !missing_secrets.is_empty() {
+        println!("\n🎉 All secrets have been configured! Recipe execution will now continue.");
+    }
```
jsibbison-square, line 117: There is part of me that wants to separate the user interactions (output and inputs) into a separate file from 'recipe' logic of recipe.rs. But I see this is not the only place in the file where this is happening so maybe its a future thing to consolidate all of the ux.

crates/goose-cli/src/recipes/secret_discovery.rs
```diff
@@ -0,0 +1,508 @@
...
+///
+/// # Returns
+/// A vector of SecretRequirement objects, deduplicated by key name
+pub fn discover_recipe_secrets(recipe: &Recipe) -> Vec<SecretRequirement> {
+    let mut visited_recipes = HashSet::new();
+    discover_recipe_secrets_recursive(recipe, &mut visited_recipes)
+}
+
+/// Internal recursive function (depth-first search) to discover secrets nested in sub-recipes
+/// This is future-proofing for a time when we have more than one-level of sub-recipe nesting
```
jsibbison-square, line 74: nit: I think if this was done with iterative breadth first search (eg. using a queue) then you could filter on the seen ext/keys globally vs in this recursive version where seen_keys/secrets is empty on each call.

Doesn't matter for the scale recipes have things I expect.

crates/goose-cli/src/recipes/secret_discovery.rs
```diff
@@ -0,0 +1,508 @@
...
+        let secrets = discover_recipe_secrets(&recipe);
+
+        assert_eq!(secrets.len(), 1);
+
+        let parent_secret = secrets.iter().find(|s| s.key == "PARENT_TOKEN").unwrap();
+        assert_eq!(parent_secret.extension_name, "parent-ext");
+    }
+
+    /// Integration tests - these test the full workflow within recipe context
+    mod integration_tests {
```
jsibbison-square: These "integration_tests" seem to create an in memory recipe struct and call `discover_recipe_secrets` which seems to be the same as the "unit" tests above. Please check if there is duplication in tests that can be removed.

crates/goose-cli/src/recipes/recipe.rs
cloud-on-prem: Agree; I think we should consolidate UX for cli. It's not uniform across the board.

crates/goose-cli/src/recipes/secret_discovery.rs
```diff
@@ -0,0 +1,508 @@
...
+        use super::*;
+        use crate::recipes::recipe::collect_missing_secrets;
+        use goose::config::Config;
+        use goose::recipe::SubRecipe;
+        use serde_json::Value;
+
+        fn create_test_recipe_with_mcp_extensions() -> Recipe {
```
jsibbison-square: This fn is only used once, maybe change it to accept extensions and sub_recipes so it can be reused to reduce duplication.

crates/goose-cli/src/recipes/secret_discovery.rs
```diff
@@ -0,0 +1,508 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use goose::agents::extension::{Envs, ExtensionConfig};
+    use goose::recipe::Recipe;
+    use std::collections::HashMap;
+
+    fn create_test_recipe_with_extensions() -> Recipe {
```
jsibbison-square, line 130: This fn is only used once, maybe change it to accept extensions and sub_recipes so it can be reused to reduce duplication.

crates/goose-cli/src/recipes/recipe.rs
```diff
@@ -35,7 +38,16 @@ fn load_recipe_file_with_dir(recipe_name: &str) -> Result<(RecipeFile, String)>
 pub fn load_recipe(recipe_name: &str, params: Vec<(String, String)>) -> Result<Recipe> {
     let recipe_file = retrieve_recipe_file(recipe_name)?;
     match build_recipe_from_template(recipe_file, params, Some(create_user_prompt_callback())) {
-        Ok(recipe) => Ok(recipe),
+        Ok(recipe) => {
+            let secret_requirements = discover_recipe_secrets(&recipe);
+            if let Err(e) = collect_missing_secrets(&secret_requirements) {
+                eprintln!(
+                    "Warning: Failed to collect some secrets: {}. Recipe will continue to run.",
```
jsibbison-square, line 45: If extensions won't work because of missing secrets is it valid for the recipe still to run? Should we panic instead.

crates/goose-cli/src/recipes/secret_discovery.rs
```diff
@@ -0,0 +1,508 @@
...
+                    let sub_secrets =
+                        discover_recipe_secrets_recursive(&loaded_recipe, visited_recipes);
+                    for sub_secret in sub_secrets {
+                        if seen_keys.insert(sub_secret.key.clone()) {
+                            secrets.push(sub_secret);
+                        }
+                    }
+                }
+                Err(_) => {
+                    continue;
```
jsibbison-square, line 104: Shouldn't we panic if we cannot load a sub_recipe rather than continuing?

crates/goose-cli/src/recipes/recipe.rs
cloud-on-prem: I left it as a warning and tolerant intentionally because the secret might actually be optional.

crates/goose-cli/src/recipes/secret_discovery.rs
cloud-on-prem: Again, I am going to defer to the main recipe runner to fail rather than failing here because this is scoped as an "eager checker"

crates/goose-cli/src/recipes/secret_discovery.rs
cloud-on-prem: yep, they were covered under the unit tests; removed the integration ones

crates/goose-cli/src/recipes/recipe.rs
jsibbison-square: I think the cliclack password field didn't allow me to do empty so probably cannot handle optional fields. We probably need to fix so you can submit empty then.

cloud-on-prem: Made the secret input "skippable" by the user. (press esc)

PR #3670, opened 2025-07-26
pr patch: task tracking for more aggressive compression

michaelneale:
this provides an in memory task list which can be used especially when there is more frequent summarisation going on and it can help keep track of things it was working on and what is next. It is not stored in the session.


crates/goose/src/agents/agent.rs
```diff
@@ -369,6 +372,8 @@ impl Agent {
             ToolCallResult::from(Err(ToolError::ExecutionError(
                 "Frontend tool execution required".to_string(),
             )))
+        } else if tool_call.name == TASK_TRACKER_TOOL_NAME {
```
DOsinga, line 376: we should really stop doing this. I'm adding a extension_manager.add_client(...) method, we should start using that (also so this tool doesn't show up when we do other things)

michaelneale: yeah - I don't like this, seems clumsy, and not just code, but need some way to enable/disable these similar to if they were MCP delivered tools?

michaelneale: @DOsinga hrm I wonder if this one should be in developer MCP - as that is where it yields most value and it is rarely not enabled (I'm not sure how in future we decide where prefixed tools go)

michaelneale:
thinking this could/should move to the developer MCP (developer/mod.rs)

crates/goose/src/agents/agent.rs
DOsinga: let's keep in general but move it to use extension_manager.add_client(...)

DOsinga (approved):

katzdave:
LGTM and 

I merged the auto-compact branch; I see git swapped you back to main with a bunch of merge conflicts now. 

Working some more on: https://github.com/block/goose/pull/3721. Goal is to remove dep of token counter on auto_compact/summarize, and hopefully make the numbers in the autocompact message + context window indicator line up.

michaelneale:
@katzdave yeah I might redirect this one to developer tools

michaelneale:
will retire this one shortly...

PR #3675, opened 2025-07-27
fix: pricing estimation for OpenRouter in goose-cli

praneeth-ovckd:
This PR fixes cost estimation for OpenRouter. OpenRouter model are of the format "provider/model", which need to be parsed to get the actual provider and model, how goose-server does it.

### DIFF
--- a/crates/goose-cli/src/session/output.rs
+++ b/crates/goose-cli/src/session/output.rs
@@ -3,6 +3,7 @@ use console::{style, Color};
 use goose::config::Config;
 use goose::message::{Message, MessageContent, ToolRequest, ToolResponse};
 use goose::providers::pricing::get_model_pricing;
+use goose::providers::pricing::parse_model_id;
 use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
 use mcp_core::tool::ToolCall;
 use regex::Regex;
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/session/output.rs
```diff
@@ -732,9 +733,21 @@ async fn estimate_cost_usd(
...
+    let openrouter_data = if provider == "openrouter" {
+        parse_model_id(model)
+    } else {
+        None
+    };
+
+    let (provider_to_use, model_to_use) = match &openrouter_data {
+        Some((real_provider, real_model)) => (real_provider.as_str(), real_model.as_str()),
+        None => (provider, model),
+    };
```
DOsinga, line 746: how about:

```
    let (provider_to_use, model_to_use) = if provider == "openrouter" {
        parse_model_id(model)
            .map(|(real_provider, real_model)| (real_provider.as_str(), real_model.as_str()))
            .unwrap_or((provider, model))
    } else {
        (provider, model)
    };
```

DOsinga (approved):

crates/goose-cli/src/session/output.rs
praneeth-ovckd: Won't this lead to a dangling reference?

PR #3678, opened 2025-07-27
chore: refactor session naming into provider

angelahning:


crates/goose/src/providers/base.rs
```diff
@@ -338,6 +339,64 @@ pub trait Provider: Send + Sync {
...
+    ///
+    /// # Arguments
+    /// * `messages` - The conversation history as a sequence of messages
+    ///
+    /// # Returns
+    /// A string containing the session name/description, sanitized and truncated to be safe for use
+    ///
+    /// # Errors
+    /// ProviderError if the session name generation fails
+    async fn generate_session_name(&self, messages: &[Message]) -> Result<String, ProviderError> {
```
DOsinga, line 346: below and possibly above: I have my quest to delete comments that don't help, can you make it so :) ?

crates/goose/src/providers/base.rs
```diff
@@ -338,6 +339,64 @@ pub trait Provider: Send + Sync {
...
+        let mut description_prompt = "Based on the conversation so far, provide a concise description of this session in 4 words or less. This will be used for finding the session later in a UI with limited space - reply *ONLY* with the description".to_string();
+
+        // Get context from messages so far, limiting each message to 300 chars for security
+        let context: Vec<String> = messages
+            .iter()
+            .filter(|m| m.role == rmcp::model::Role::User)
+            .take(3) // Use up to first 3 user messages for context
+            .map(|m| {
+                let text = m.as_concat_text();
+                safe_truncate(&text, 300)
```
DOsinga: I think we can drop this truncate. especially for sessions where the first message is really quite long because the user knows what they are doing, you probably don't want to do this. as for security, these messages have gone through the agent already. if we do want to keep this make it 3000 or so.

I'm aware this is pre-existing code, but still might as well make things better

crates/goose/src/session/storage.rs
```diff
@@ -1294,52 +1294,15 @@ pub async fn generate_description_with_schedule_id(
...
-    let message = Message::user().with_text(&description_prompt);
-    let result = provider
-        .complete(
-            "Reply with only a description in four words or less",
-            &[message],
-            &[],
-        )
+    // Use the provider's session naming capability
+    let sanitized_description = provider
+        .generate_session_name(messages)
```
DOsinga, line 1299: much better! still not really where it should be called from, but we can clean that up some other time

DOsinga (approved):

crates/goose/src/providers/base.rs
angelahning: sounds good

Issue #3682, opened 2025-07-27
regression: claude cli and gemini cli providers should have a default


michaelneale:
I think in this change: https://github.com/block/goose/pull/3378 - if you set the GOOSE_MODEL to default, then it wont' work (it used to) as it expects one of the named models: 

<img width="841" height="110" alt="Image" src="https://github.com/user-attachments/assets/d6a5caa5-5b35-43d7-b9d6-7bb63212d6c4" />

similarly with a change for gemini. 

This is not ideal as the models may change vs in that list, and also (correct me if wrong) you don't choose the model when you setup claude or gemini clis always do you? 

We should not require a model value to be set or valid (ie if it isn't one of those, it should just pick... well not pass the model parameter? the way it used to?)

cc @cbruyndoncx @kwsantiago

kwsantiago:
Originally my intention was for the user to choose based on the available models, since you choose the model when you setup Claude Code or Gemini CLI (it defaults to a model such as Gemini 2.5 or Sonnet 4 for example). The previous code had a Gemini CLI version that was not the latest version, and for Claude Code, it only provided 3.5 Sonnet instead of providing access to all models, which I fixed in that PR.

In any case, agree there is a regression here and will open a PR to fix to ensure the functionality is restored when GOOSE_MODEL is set to default.

Thank you for pointing out!

michaelneale:
@kwsantiago thanks  - https://github.com/block/goose/pull/3683 is my take, seems pretty simple to do in theory, if you get a chance take a look if you think that one makes sense. not sure of best long term way, as there isn't really a standard way for the goose side providers to ask the clis what models are there

kwsantiago:
@michaelneale I started working on a solution, but I think yours is better. Will leave some feedback in your PR but overall your solution is great as is.

PR #3683, opened 2025-07-28
fix: have option for cli providers to use their configured or default model 

michaelneale:
fixes: https://github.com/block/goose/issues/3682

### DIFF
--- a/crates/goose/src/providers/claude_code.rs
+++ b/crates/goose/src/providers/claude_code.rs
@@ -332,12 +332,14 @@ impl ClaudeCodeProvider {
         cmd.arg("-p")
             .arg(messages_json.to_string())
             .arg("--system-prompt")
-            .arg(&filtered_system)
-            .arg("--model")
-            .arg(&self.model.model_name)
-            .arg("--verbose")
-            .arg("--output-format")
-            .arg("json");
+            .arg(&filtered_system);
+
+        // Only pass model parameter if it's in the known models list
+        if CLAUDE_CODE_KNOWN_MODELS.contains(&self.model.model_name.as_str()) {
+            cmd.arg("--model").arg(&self.model.model_name);
+        }
+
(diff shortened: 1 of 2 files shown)


crates/goose/src/providers/claude_code.rs
```diff
@@ -332,12 +332,14 @@ impl ClaudeCodeProvider {
...
             .arg("--system-prompt")
-            .arg(&filtered_system)
-            .arg("--model")
-            .arg(&self.model.model_name)
-            .arg("--verbose")
-            .arg("--output-format")
-            .arg("json");
+            .arg(&filtered_system);
+
+        // Only pass model parameter if it's in the known models list
```
kwsantiago, line 337: Consider adding 

`// This allows users to use "default" or any future CLI models without updates`

(and same for gemini_cli.rs)

kwsantiago:
@michaelneale Great solution! I think as is your solution is excellent, main feedback I could give is consider updating `KNOWN_MODELS` since the CLIs support more models than we list.

We could either:
- Keep it minimal (current approach)
- Or document that `KNOWN_MODELS` is intentionally incomplete to allow CLI flexibility

What do you think?

michaelneale:
ideally if someone types it in then we would pass it through - otherwise we dont' pass anything, maybe that is better approach vs having any list/hard coding at all?

kwsantiago:
I still think having a known models list makes sense to prevent user error that could cause the CLI to fail (ie. invalid/outdated/unsupported model or typos etc). 

The middle ground would be to only skip passing the model for specific values like "default" or empty string, otherwise let it pass through. That would avoid hardcoding model lists while allowing CLI defaults.

Either way, what you have in this PR should be sufficient to solve the regression issue, I think it's a great solution as is.

DOsinga (approved):

Issue #3664, opened 2025-07-25
Feat - Preserve Scroll Position When Returning from Session Preview


AaronGoldsmith:
### Motivation

Does this feature solve a particular problem or unlock use cases?

When searching through multiple chat sessions to find the right one, users often have to scroll through a long list. After clicking into a chat to verify its content, clicking "Back" currently snaps the user to the top of the session list—forcing them to scroll again to where they were previously. This creates user experience friction, as finding the correct session among many is slow and repetitive.

**Opportunities unlocked:**
- Speeds up session navigation for users with large histories.
- Reduces frustration when browsing for past work or references.
- Enabling “return to previous scroll position” after viewing a chat makes Goose feel faster and more intuitive, especially for power users with many sessions.

---

### Describe the solution you'd like

- When returning from a chat/session detail view to the session list, restore the scroll position to where the user left off, rather than returning to the top.
- The user should seamlessly continue browsing sessions, minimizing repetitive scrolling.

---

### Describe alternatives you've considered

- Implement a timeline scrollbar with dates (see [Midjourney’s Timeline Bar for inspiration](https://docs.midjourney.com/hc/en-us/articles/33329462451469-Organizing-Your-Creations#:~:text=your%20Organize%20page-,Timeline%20Bar,-Next%20to%20the)), allowing quick jumping between time periods instead of, or in addition to, scroll restoration.

---

### Additional context

- [x] I have verified this does not duplicate an existing feature request.

Real Scenario: I have two sessions with the same name, and I'm not sure which one is the "correct" one until I open the preview. After finding that the first one wasn't the session I wanted, I click "Back", but am brought to the top of my session list, and have to scroll to find these sessions again.

<img width="658" height="361" alt="Image" src="https://github.com/user-attachments/assets/303bc81b-3ca8-4a45-9293-56e5dd4d4186" />

tiensi:
Feat request in progress https://github.com/block/goose/pull/3684

taniacryptid:
Added with #4360 , thank you for flagging this @Abhijay007 !

PR #3684, opened 2025-07-28
[WIP] feat: Added scroll state support for chat-session-list navigation

tiensi:
**Summary**
Added scroll state memory when navigating back from session view.

**Changes**
- Added `showSessionHistory` to SessionsView to replace reliance on `selectedSession` being null
- Passed `selectedSession` as a new prop to `SessionListView` so that it can scroll to a selected session
- Added `useEffect` for scrolling in `SessionListView`

**Request**
https://github.com/block/goose/issues/3664

**After Changes**

https://github.com/user-attachments/assets/2aabe09a-3d71-4a8f-8a51-ca0cfcf0a8f0

**Before**

https://github.com/user-attachments/assets/ec41d392-6a48-405a-833a-1b748200b151


tiensi:
Thanks for the heads up, I'll test a bit more and take a look at the search use case.

michaelneale:
@tiensi still updated it - you interested in this - still an issue?

tiensi:
Thanks for the reminder, let me close this for now to reduce noise until I can get a more robust fix in.

tiensi:
This was stale so I rebuilt it here: https://github.com/block/goose/pull/4360

PR #3685, opened 2025-07-28
chore: adding in some new models token limits

michaelneale:
I did have a look at making openrouter load models context sizes (as it does return them in a list) but wasn't reliably able to get the correct size, so this just adds some new popular models with the correct context size

### DIFF
--- a/crates/goose/src/model.rs
+++ b/crates/goose/src/model.rs
@@ -28,11 +28,10 @@ static MODEL_SPECIFIC_LIMITS: Lazy<HashMap<&'static str, usize>> = Lazy::new(|| 
     map.insert("llama3.3", 128_000);
 
     // x.ai Grok models, https://docs.x.ai/docs/overview
-    map.insert("grok", 131_072);
+    map.insert("grok-3", 131_072);
+    map.insert("grok-4", 256_000); // 256K
 
-    // Additional OpenRouter models
-    map.insert("qwen/qwen3-coder", 262_144); // 262K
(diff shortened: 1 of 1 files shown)


crates/goose/src/model.rs
```diff
@@ -29,6 +29,11 @@ static MODEL_SPECIFIC_LIMITS: Lazy<HashMap<&'static str, usize>> = Lazy::new(||
 
     // x.ai Grok models, https://docs.x.ai/docs/overview
     map.insert("grok", 131_072);
+
+    // Additional OpenRouter models
+    map.insert("qwen/qwen3-coder", 262_144); // 262K
+    map.insert("x-ai/grok-4", 256_000); // 256K
```
DOsinga: we do:

```
        for (pattern, &limit) in MODEL_SPECIFIC_LIMITS.iter() {
            if model_name.contains(pattern) {
```

so presumably we can just drop the provide/ prefix here and score even better. also, this is a map, doesn't seem like we'd want that since we go through it in order. the above loop doesn't really work on hasmaps since the order is random. also these insertions seem to be double?

michaelneale: I don't think the order does matter? In any case it wouldn't show up with that limit without that change so the pattern doesn't match (and yeah doesn't have to be open router specific)

michaelneale: oh right, yeah I see - the pattern is from the map not the model name, so ideally we want the closest match (well that is the desire), so it means the patterns should be as specific as possible (but no more). so "grok" as a pattern isn't ideal

PR #3617, opened 2025-07-23
alexhancock/rmcp-tools-annotations

alexhancock:
Draft of moving from internal representation of tools to the rmcp version

PR #3688, opened 2025-07-28
Provider scenario tests

DOsinga:
Have scenario tests per provider. We are still missing quite some coverage, but it is a start!


crates/goose-cli/src/scenario_tests/mock_client.rs
```diff
@@ -0,0 +1,165 @@
+//! MockClient is a mock implementation of the McpClientTrait for testing purposes.
+//! add a tool you want to have around and then add the client to the extension router
+
+use mcp_client::client::{ClientCapabilities, ClientInfo, Error, McpClientTrait};
+use mcp_core::protocol::{
+    CallToolResult, Implementation, InitializeResult, ListPromptsResult, ListResourcesResult,
+    ListToolsResult, ReadResourceResult, ServerCapabilities, ToolsCapability,
+};
+use mcp_core::{Tool, ToolError};
```
alexhancock, line 9: Can you import `Tool` from `rmcp::model` instead?

Sorry, I should have removed the old definition when I finished the migration between those types here https://github.com/block/goose/pull/3617/files

I'll do that as a small PR, but good to start importing the new types

DOsinga: yeah, it's been a moving target - when I merged in master I had to update all the references again :)

crates/goose-cli/src/scenario_tests/scenario_runner.rs
```diff
@@ -0,0 +1,304 @@
...
+                )
+            })?;
+
+        println!("Running test '{}' for provider: {}", test_name, config.name);
+        run_provider_scenario_with_validation(config, test_name, &message_generator, &validator)
+            .await?;
+        return Ok(());
+    }
+
+    let configs_to_test: Vec<_> = if let Some(to_skip) = providers_to_skip {
```
jamadeo: you can get an iterator from an Option<Vec<T>> with option.iter().flatt() -- useful for this kind of thing

crates/goose-cli/src/scenario_tests/scenario_runner.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    let mut error = None;
+    for message in &messages {
+        if let Err(e) = session.process_message(message.clone()).await {
+            error = Some(e.to_string());
+            break;
+        }
+    }
+
+    // Rest of the function remains the same...
```
jamadeo: I see you, llm!

crates/goose-cli/src/scenario_tests/scenarios.rs
```diff
@@ -1,22 +1,105 @@
...
+                );
+                Ok(())
+            },
+        )
+        .await
+    }
+
+    #[tokio::test]
+    async fn test_weather_tool() -> Result<()> {
+        // Google tells me it only knows about the weather in the US, so we skip it.
```
jamadeo, line 33: I suppose we could ask it about the weather somewhere it knows about

DOsinga: we could, but I am guessing it will still refuse to call the tool? I'm just trying to get this in first and then we can make it work for more and more things

crates/goose-cli/src/scenario_tests/scenario_runner.rs
DOsinga: ha!

crates/goose-cli/src/scenario_tests/mock_client.rs
DOsinga: would you mind if we do that in a follow up? it gives me some weird type errors

PR #3693, opened 2025-07-28
fix: address issue with streamable http interactions via mcp

alexhancock:
* Adding a fix for an issue with streamable http mcp servers discovered by @tadas
* Adding tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1827,6 +1827,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5b63caa9aa9397e2d9480a9b13673856c78d8ac123288526c37d7839f2a86990"
 
 [[package]]
+name = "colored"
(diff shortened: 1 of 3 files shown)


tadasant:
> discovered by `@Tadas`

@tadasant ** :) 

Thank you for digging into this!!

alexhancock:
> > discovered by `@Tadas`
> 
> @tadasant ** :)
> 
> Thank you for digging into this!!

replaced!

tadasant:
Confirming this solves the minimal repro I had communicated to @alexhancock, and additionally is unblocking me on running long-running sub-recipes when I combine it with my draft PR in https://github.com/block/goose/pull/3691 (which fixes `task_timeout` not being respected by the subagent invoked by the sub-recipe). Thank you!

katzdave (approved):
🙌 Nice tests!

Issue #3680, opened 2025-07-27
UI Bug: Extensions List Stale After Installation


AaronGoldsmith:
**Describe the bug**
Newly installed extensions are not visible on the Extensions page without a manual refresh.

**To Reproduce**
1. Go to Extensions
2. Click on Add new extension
3. Install the extension
4. The extension is installed, but is not displayed in the UI until refreshed or another state change. 

**Expected behavior**
As soon as you install a new extension, it should appear in the extension list. 

**Screenshots**

Demo: installing a new extension. I eventually refresh the page at 15seconds, and the list updates. 

https://github.com/user-attachments/assets/1213732e-4ffa-4c10-b729-0bc60bd0d181

Alternatively, you can toggle a different extension for it to re-render.

https://github.com/user-attachments/assets/f36ef0fb-774a-4266-b4f4-758ee134daf3

- OS & Arch: macOS
- Interface: UI
- Version: 1.1.4
- Extensions enabled: Default built-in extensions

PR #3695, opened 2025-07-28
Ensure adding/removing extensions refreshes extensions list

zanesq:
Ensure adding/removing extensions refreshes extensions list

Also added support for renaming an extension without adding a new one.

fixes [3680](https://github.com/block/goose/issues/3680)

### DIFF
--- a/ui/desktop/src/components/ConfigContext.tsx
+++ b/ui/desktop/src/components/ConfigContext.tsx
@@ -111,6 +111,23 @@ export const ConfigProvider: React.FC<ConfigProviderProps> = ({ children }) => {
     [reloadConfig]
   );
 
+  const refreshExtensions = useCallback(async () => {
+    const result = await apiGetExtensions();
+
+    if (result.response.status === 422) {
+      throw new MalformedConfigError();
+    }
+
+    if (result.error && !result.data) {
+      console.log(result.error);
+      return extensionsList;
+    }
+
+    const extensionResponse: ExtensionResponse = result.data!;
+    setExtensionsList(extensionResponse.extensions);
+    return extensionResponse.extensions;
+  }, [extensionsList]);
+
   const addExtension = useCallback(
     async (name: string, config: ExtensionConfig, enabled: boolean) => {
       // remove shims if present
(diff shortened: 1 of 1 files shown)


ui/desktop/src/components/ConfigContext.tsx
```diff
@@ -122,6 +122,8 @@ export const ConfigProvider: React.FC<ConfigProviderProps> = ({ children }) => {
         body: query,
       });
       await reloadConfig();
+      // Force refresh extensions list to ensure UI is updated
+      setExtensionsList([]);
```
alexhancock: I'm a little confused on the need for this one. Doesn't:

```await getExtensions(true)``` in fetch extensions already do a force reload?

The logic for loading extensions is very fiddly and that's my fault. I just want to be sure of the need for more imperative things to get extensions to reload before we merge.

**Comment by zanesq** [comment]
I've addressed your concern about the imperative state clearing approach. The fix now uses a cleaner pattern:

**What changed:**
- Removed the manual `setExtensionsList([])` calls
- Created a dedicated `refreshExtensions` helper function that handles the API call and state update
- Both `addExtension` and `removeExtension` now call `await refreshExtensions()` after successful operations
- `getExtensions(forceRefresh)` now delegates to `refreshExtensions` when a refresh is needed

**Why this is better:**
- No more imperative state manipulation
- Uses the established async refresh pattern consistently
- Cleaner separation of concerns
- Avoids circular dependency issues
- More predictable and maintainable

The functionality remains the same - the extensions list will properly refresh after add/remove operations - but now it's done through the proper async refresh mechanism rather than manually clearing state.

File: ui/desktop/src/components/settings/extensions/extension-manager.ts [hunk 2]
```diff
@@ -129,46 +129,121 @@ export async function addToAgentOnStartup({
...
 }: UpdateExtensionProps) {
-  if (enabled) {
+  // Sanitize the new name to match the behavior when adding extensions
+  const sanitizedNewName = sanitizeName(extensionConfig.name);
+  const sanitizedOriginalName = originalName ? sanitizeName(originalName) : undefined;
+
+  // Check if the sanitized name has changed
+  const nameChanged = sanitizedOriginalName && sanitizedOriginalName !== sanitizedNewName;
+
+  if (nameChanged) {
```
alexhancock, line 154: Handling this looks good as I know it was something missed before, but some new tests for it would be good! Either in this or a followup

alexhancock:
> I've addressed your concern about the imperative state clearing approach. The fix now uses a cleaner pattern:
> 
> **What changed:**
> 
> * Removed the manual `setExtensionsList([])` calls
> * Created a dedicated `refreshExtensions` helper function that handles the API call and state update
> * Both `addExtension` and `removeExtension` now call `await refreshExtensions()` after successful operations
> * `getExtensions(forceRefresh)` now delegates to `refreshExtensions` when a refresh is needed
> 
> **Why this is better:**
> 
> * No more imperative state manipulation
> * Uses the established async refresh pattern consistently
> * Cleaner separation of concerns
> * Avoids circular dependency issues
> * More predictable and maintainable
> 
> The functionality remains the same - the extensions list will properly refresh after add/remove operations - but now it's done through the proper async refresh mechanism rather than manually clearing state.

Yes, new implementation looks better!

ui/desktop/src/components/settings/extensions/extension-manager.ts
zanesq: Sounds good will follow up with tests now that we have vitest!

zanesq: https://github.com/block/goose/pull/3794

PR #3696, opened 2025-07-28
Add warning when JSON env parsing fails.

katzdave:


DOsinga (approved):
we should kill this whole config thing, but that's a different story

PR #3698, opened 2025-07-28
Improve power save blocker mechanism

zanesq:
change power save blocker to automatically stop and cleanup after 15 minutes of inactivity and have per window ids for better cleanup

- Each window manages its own power save blocker independently
- Power save blockers are automatically stopped after 15 minutes of inactivity
- Multiple cleanup mechanisms prevent orphaned power save blockers

### DIFF
--- a/ui/desktop/src/hooks/useChatEngine.ts
+++ b/ui/desktop/src/hooks/useChatEngine.ts
@@ -255,7 +255,7 @@ export const useChatEngine = ({
             stopPowerSaveBlocker();
           },
           15 * 60 * 1000
-        ); // 15 minutes
+        );
 
         setPowerSaveTimeoutId(timeoutId);
 


ui/desktop/src/hooks/useChatEngine.ts
```diff
@@ -206,40 +223,67 @@ export const useChatEngine = ({
...
+          console.error('Failed to start power save blocker:', error);
+        }
+
         setLastInteractionTime(Date.now());
 
+        // Set a timeout to automatically stop the power save blocker after 15 minutes
+        const timeoutId = window.setTimeout(
+          () => {
+            console.warn('Power save blocker timeout - stopping automatically after 15 minutes');
+            stopPowerSaveBlocker();
```
alexhancock, line 255: What's the intention of the power save blocker? Do we have it so that people can walk away from their computers for long periods of time and have goose continue to work? In which case we may not want to turn it off after 15 mins of inactivity?

cc @michaelneale

michaelneale: yes was exactly that - as sleep comes soon for some setups but you want it to run just a bit longer. I don't know how widely it is used and I would be ok if it made sense to retire it (this predates schedules and recipes and headless and all these other things) but then again, may still be relevant.

zanesq: Should have put more details in the original description. The reason I added this is PCI requirements state that the screen should auto lock after 15 mins of user inactivity and the power save blocker was stopping the screen from locking. Also added logic to make sure its actually tracked/stopped in other windows.

PR #3703, opened 2025-07-29
chore: small refactor on agent.rs

lifeizhou-ap:
**Why**
Start small refactoring to see whether we can reuse some logics of agent.rs in subagent.rs to avoid duplications

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -161,16 +161,6 @@ impl Agent {
         *tool_monitor = Some(ToolMonitor::new(max_repetitions));
     }
 
-    pub async fn get_tool_stats(&self) -> Option<HashMap<String, u32>> {
-        let tool_monitor = self.tool_monitor.lock().await;
-        tool_monitor.as_ref().map(|monitor| monitor.get_stats())
-    }
-
-    pub async fn reset_tool_monitor(&self) {
-        if let Some(monitor) = self.tool_monitor.lock().await.as_mut() {
-            monitor.reset();
-        }
-    }
 
     /// Reset the retry attempts counter to 0
     pub async fn reset_retry_attempts(&self) {
@@ -229,24 +219,6 @@ impl Agent {
     /// Get a reference to a frontend tool
     pub async fn get_frontend_tool(&self, name: &str) -> Option<FrontendTool> {
         self.frontend_tools.lock().await.get(name).cloned()
-    }
-
-    /// Get all tools from all clients with proper prefixing
-    pub async fn get_prefixed_tools(&self) -> ExtensionResult<Vec<Tool>> {
-        let mut tools = self
-            .extension_manager
-            .read()
-            .await
-            .get_prefixed_tools(None)
-            .await?;
-
-        // Add frontend tools directly - they don't need prefixing since they're already uniquely named
-        let frontend_tools = self.frontend_tools.lock().await;
-        for frontend_tool in frontend_tools.values() {
-            tools.push(frontend_tool.tool.clone());
-        }
-
-        Ok(tools)
     }
 
     pub async fn add_final_output_tool(&self, response: Response) {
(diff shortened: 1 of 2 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -161,17 +181,6 @@ impl Agent {
         *tool_monitor = Some(ToolMonitor::new(max_repetitions));
     }
 
-    pub async fn get_tool_stats(&self) -> Option<HashMap<String, u32>> {
```
lifeizhou-ap, line 164: these functions are not used

crates/goose/src/agents/agent.rs
```diff
@@ -822,33 +939,14 @@ impl Agent {
                             }
 
                             if let Some(response) = response {
-                                let (tools_with_readonly_annotation, tools_without_annotation) =
```
lifeizhou-ap, line 825: extracted to `process_tool_requests` function

crates/goose/src/agents/agent.rs
```diff
@@ -880,52 +978,24 @@ impl Agent {
...
-                                            tools_with_readonly_annotation.clone(),
-                                            tools_without_annotation.clone(),
+                                            readonly_tools.clone(),
+                                            regular_tools.clone(),
                                             &mut permission_manager,
                                             self.provider().await?,
-                                        )
-                                        .await;
-
-                                    let mut tool_futures: Vec<(String, ToolStream)> = Vec::new();
```
lifeizhou-ap, line 895: extracted to `execute_approved_tools` function

crates/goose/src/agents/subagent.rs
```diff
@@ -105,26 +100,6 @@ impl SubAgent {
         } // Write lock is released here!
     }
 
-    /// Get current progress information
```
lifeizhou-ap, line 108: it is not used

crates/goose/src/agents/subagent.rs
```diff
@@ -283,37 +258,16 @@ impl SubAgent {
...
         conversation.push(message);
     }
 
     /// Get the full conversation history
-    pub async fn get_conversation(&self) -> Vec<Message> {
+    async fn get_conversation(&self) -> Vec<Message> {
         self.conversation.lock().await.clone()
     }
 
-    /// Check if the subagent has completed its task
```
lifeizhou-ap, line 296: not used

crates/goose/src/agents/agent.rs
```diff
@@ -207,6 +216,120 @@ impl Agent {
...
+            frontend_requests,
+            remaining_requests,
+            filtered_response,
+            readonly_tools,
+            regular_tools,
+        }
+    }
+
+    /// Execute approved tools and handle denied tools
+    async fn execute_approved_tools(
```
DOsinga: remove the comment and rename to handle approved_and_denied_tools?

crates/goose/src/agents/agent.rs
```diff
@@ -207,6 +216,120 @@ impl Agent {
...
+            toolshim_tools,
+            system_prompt,
+            goose_mode,
+            initial_messages,
+            config,
+        })
+    }
+
+    /// Process tool requests by categorizing them and recording them in the router selector
+    async fn process_tool_requests(
```
DOsinga, line 249: we're doing two things here, maybe would be better to just do record_tool_calls and move that to router_tool_selector to record more than one call and then we can just call that directly from agent with `frontend_requests.iter().chain(&remaining_requests)`

DOsinga (approved):
thanks for doing this. two small suggestions

crates/goose/src/agents/agent.rs
lifeizhou-ap: Thanks for the review! Yes, the function has a couple of responsibilities.  I feel the tool handling in reply could be cleaner with more refactoring and will have a followup pr to refactor more in this area.

PR #3706, opened 2025-07-29
Env parsing for primitive types

katzdave:
Start with the json, but provide reasonable primitive parses for non-quoted primitives.

lots of test cases.

Original justification for changes here was .0001 failed json parsing silently and didn't get picked up.

### DIFF
--- a/crates/goose/src/config/base.rs
+++ b/crates/goose/src/config/base.rs
@@ -10,7 +10,6 @@ use std::fs::OpenOptions;
 use std::io::Write;
 use std::path::{Path, PathBuf};
 use thiserror::Error;
-use tracing::warn;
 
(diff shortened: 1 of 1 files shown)


crates/goose/src/config/base.rs
```diff
@@ -508,6 +507,44 @@ impl Config {
         }
     }
 
+    /// Parse an environment variable value into a JSON Value.
```
michaelneale, line 510: useful comment - keep it in!

michaelneale (approved):
works well, thanks @katzdave

crates/goose/src/config/base.rs
DOsinga: yeah remove other comments too

PR #3707, opened 2025-07-29
Add configurable streaming and embeddings support for OpenAI provider

casassg:
## Summary
Adds `OPENAI_ENABLE_STREAMING` and `OPENAI_ENABLE_EMBEDDINGS` environment variables to allow disabling streaming and embeddings functionality in the OpenAI provider. Both default to true for backward compatibility.

## Changes

- Code: Added configurable boolean fields to `OpenAiProvider` struct and updated `supports_streaming()` and `supports_embeddings()` methods
- Documentation: Updated provider configuration tables with the new optional parameters

## Use Cases

- OpenAI-compatible endpoints that don't support all features.

## Configuration 

```bash
export OPENAI_ENABLE_STREAMING=false    # Disable streaming
export OPENAI_ENABLE_EMBEDDINGS=false   # Disable embeddings
```

## Backward Compatibility
✅ No breaking changes - all existing configurations continue to work unchanged.


michaelneale:
makes sense - but what is the need for this? to try things out? compatibility or preference?

casassg:
> makes sense - but what is the need for this? to try things out? compatibility or preference?

Compatibility. Some providers may not offer streaming (due to various reasons). 

In this specific case its due to some issues w sagemaker streaming with OpenAI compatible endpoint.

DOsinga:
I think we should close this, merge https://github.com/block/goose/pull/3825 at which point we have true custom providers based on an engine and then we can add streamability etc as a feature of that json format. what do you think @casassg ?

casassg:
Fine by me, lmk if I can help test the PR using openai compatible apis

michaelneale:
thanks for this - will close this in favour of other inbound changes

PR #3709, opened 2025-07-29
Fix tool vector tests

DOsinga:
we were creating dbs but not with unique names and weren't cleaning them up always either which presumably led to test failures when we had an instance that had this already around? this introduces a method to create a new unique db and returns also a clean up task


ahau-square (approved):
thanks for fixing!

Issue #3294, opened 2025-07-08
ollama provider has unhelpful session names


michaelneale:
<img width="1038" height="185" alt="Image" src="https://github.com/user-attachments/assets/e9f3ed12-f5a6-4a0e-b2ab-089c990add9d" />

I think could be simplified to not use the LLM to come up with a session name (just slows things down)

DOsinga:
Yeah, thanks for noting. This is a small model though and I think it is just ignoring the instructions to keep the session description to 4 words  or less. I don't think it is worth it to fix this.

michaelneale:
happens with other models - so will look at a way to fix it

michaelneale:
<img width="1112" height="635" alt="Image" src="https://github.com/user-attachments/assets/35ebda57-7f9c-44f0-b75e-0783de054b90" />

seems to get confused with think tags

DOsinga:
which model is this? but yeah we can strip that out I am sure

DOsinga:
this should be fixed now!

PR #3710, opened 2025-07-29
Override session name generator for ollama provider

angelahning:
Addresses https://github.com/block/goose/issues/3294

### DIFF
--- a/crates/goose/src/providers/base.rs
+++ b/crates/goose/src/providers/base.rs
@@ -344,26 +344,8 @@ pub trait Provider: Send + Sync {
     /// This method can be overridden by providers to implement custom session naming strategies.
     /// The default implementation creates a prompt asking for a concise description in 4 words or less.
     async fn generate_session_name(&self, messages: &[Message]) -> Result<String, ProviderError> {
-        // Create a prompt for a concise description
-        let mut description_prompt = "Based on the conversation so far, provide a concise description of this session in 4 words or less. This will be used for finding the session later in a UI with limited space - reply *ONLY* with the description".to_string();
-
-        // Get context from the first 3 user messages
-        let context: Vec<String> = messages
-            .iter()
-            .filter(|m| m.role == rmcp::model::Role::User)
-            .take(3)
-            .map(|m| m.as_concat_text())
-            .collect();
-
-        if !context.is_empty() {
-            description_prompt = format!(
-                "Here are the first few user messages:\n{}\n\n{}",
-                context.join("\n"),
-                description_prompt
-            );
-        }
-
-        let message = Message::user().with_text(&description_prompt);
+        let prompt = self.create_session_name_prompt(messages);
+        let message = Message::user().with_text(&prompt);
         let result = self
             .complete(
                 "Reply with only a description in four words or less",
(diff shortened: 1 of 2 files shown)


crates/goose/src/providers/ollama.rs
```diff
@@ -154,4 +156,77 @@ impl Provider for OllamaProvider {
...
+                &[message],
+                &[],
+            )
+            .await?;
+
+        let mut description = result.0.as_concat_text();
+        description = Self::filter_reasoning_tokens(&description);
+
+        let sanitized_description = if description.chars().count() > 100 {
+            safe_truncate(&description, 100)
```
DOsinga: safe_truncate already does check for length, so need to do that here. good to use chars().count) though

crates/goose/src/providers/ollama.rs
```diff
@@ -154,4 +156,77 @@ impl Provider for OllamaProvider {
...
+        } else {
+            description
+        };
+
+        // If the description is too long, the model failed to give us a short description, provide a fallback instead
+        if sanitized_description.split_whitespace().count() > 6 {
```
DOsinga: this is not going to work for languages that don't use whitespace to separate words. you could use tiktoken but I'd just leave it at the length check above

crates/goose/src/providers/base.rs
```diff
@@ -381,6 +363,29 @@ pub trait Provider: Send + Sync {
...
+        // Create a prompt for a concise description
+        let mut prompt = "Based on the conversation so far, provide a concise description of this session in 4 words or less. This will be used for finding the session later in a UI with limited space - reply *ONLY* with the description".to_string();
+
+        // Get context from the first 3 user messages
+        let context: Vec<String> = messages
+            .iter()
+            .filter(|m| m.role == rmcp::model::Role::User)
+            .take(3)
+            .map(|m| m.as_concat_text())
+            .collect();
```
DOsinga: should we pass this in, instead of messages? the outer callers are only going to call us with the first three messages anyway (which I guess also means we should probably not cut at 3 here at all)

crates/goose/src/providers/ollama.rs
```diff
@@ -154,4 +156,77 @@ impl Provider for OllamaProvider {
...
         Ok((message, ProviderUsage::new(model, usage)))
     }
+
+    /// Generate a session name based on the conversation history
+    /// This override filters out reasoning tokens that some Ollama models produce
+    async fn generate_session_name(&self, messages: &[Message]) -> Result<String, ProviderError> {
+        let message = Message::user().with_text(&self.create_session_name_prompt(messages));
+        let result = self
+            .complete(
+                "You are a title generator. Output only the requested title with no additional text, reasoning, or explanations.",
```
DOsinga: shouldn't we tell it to keep it to 4 words here too?

DOsinga (approved):

crates/goose/src/providers/ollama.rs
angelahning: ahh looks like we do that in the original code too. Let me clean both up

crates/goose/src/providers/ollama.rs
angelahning: we tell it in the prompt already. I can also add it here.

crates/goose/src/providers/base.rs
angelahning: sure.

PR #3711, opened 2025-07-29
chore: Speed up CI

jamadeo:
Speeds up the build-and-test-rust step by fixing the caching, and removing unnecessary disk maintenance steps.


DOsinga (approved):
w00t! not sure why this would change the cargo.lock though

jamadeo:
> w00t! not sure why this would change the cargo.lock though

Oh it doesn't -- that was just me triggering a new cache key to test this out

PR #3712, opened 2025-07-29
Add ability to share developer extension with other callers

ravarora2:
We would like to use developer mcp form other applications in Block. In order to expose this extension, we are doing the following:

1) Expose call tool api so that other callers can call the developer extension
2) Add a new share developer view. This view does the following:
- Create a new goose server with a new port and secret
- Add developer extension to the server instance
- Share url and secret with the user. The user is required to use the secret key at the client side making sure that only the valid users who have access to the goose desktop can copy and paster the secret in the client app. 
3) The caller uses the url+secret key + call tool api to run commands on developer extension. 

Screenshot 2025-07-29 at 11 52 50 AM
Screenshot 2025-07-29 at 11 53 01 AM
Screenshot 2025-07-29 at 11 53 24 AM
Screenshot 2025-07-29 at 11 54 51 AM
Screenshot 2025-07-29 at 11 55 00 AM


crates/goose-server/src/routes/agent.rs
```diff
@@ -221,6 +239,72 @@ async fn get_tools(
...
+
+    // Generate a request ID if not provided
+    let request_id = payload.request_id.unwrap_or_else(|| {
+        use std::time::{SystemTime, UNIX_EPOCH};
+        let timestamp = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_millis();
+        format!("tool_call_{}", timestamp)
+    });
```
DOsinga, line 276: Don't use time to generate ids

DOsinga:
so what's the verdict here? do we want to continue on this road?

michaelneale:
I think so other agents can use the built-in mcps in goose - running from the desktop (ie as a little local server).

DOsinga:
closing this for now due to non activity

PR #3717, opened 2025-07-29
Speed up app initialization and improve refresh crashing

zanesq:
Speed up app initialization by only checking cost if enabled and parallelizing initialization requests.

Also added more guards for localstorage not available to help with refresh crashing issues.

```
VM8:16 Failed to access localStorage: SecurityError: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
(anonymous) @ VM8:16
VM8:22 Failed to set localStorage after retry: SecurityError: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
(anonymous) @ VM8:22
chromewebdata/:1 Not allowed to load local resource: file:///pair
```

### DIFF
--- a/ui/desktop/index.html
+++ b/ui/desktop/index.html
@@ -9,7 +9,7 @@
       (function() {
         function initializeTheme() {
           try {
-            if (typeof Storage !== 'undefined' && window.localStorage) {
+            if (window.localStorage) {
               const useSystemTheme = localStorage.getItem('use_system_theme') === 'true';
               const systemPrefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
               const savedTheme = localStorage.getItem('theme');
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -678,7 +678,7 @@ const createChat = async (
       (function() {
         function setConfig() {
           try {
-            if (typeof Storage !== 'undefined' && window.localStorage) {
+            if (window.localStorage) {
               localStorage.setItem('gooseConfig', '${configStr}');
               return true;
             }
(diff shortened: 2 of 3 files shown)


michaelneale:
very nice, startup could be much faster, don't quite understand all the changes but worth a look - probably worth testing by hand when bundled/signed too.

ui/desktop/index.html
```diff
@@ -8,20 +8,42 @@
...
-          const systemPrefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
-          const savedTheme = localStorage.getItem('theme');
-          const isDark = useSystemTheme ? systemPrefersDark : (savedTheme ? savedTheme === 'dark' : systemPrefersDark);
-          
-          if (isDark) {
-            document.documentElement.classList.add('dark');
-          } else {
-            document.documentElement.classList.remove('dark');
+          try {
+            if (typeof Storage !== 'undefined' && window.localStorage) {
```
michaelneale: when would storage be undefined?

zanesq: it can be undefined in older browsers or if localstorage is disabled so its just a check that its available before running localstorage operations

michaelneale: would be in a pretty bad state if we hit that in the electron app though wouldn't it?

alexhancock: Yeah I am also a bit confused on this point. Wouldn't electron bundle the APIs being used, so it would either be defined or not in a given implementation?

I'm not against checking, but just wondering if it could ever catch.

zanesq: Good call, wasn't really needed removed!

PR #3718, opened 2025-07-29
Ok, well, that got out of hand

DOsinga:
Ok, it looks like a lot is happening here, but ... ok, fine there is some stuff, but mostly:

* Process the models and limits in the right order
* It makes it so that if we load a ModelConfig and it is faulty, we don't gloss over it, but fail (easy to set environment variables worng)
* Simplifies the default for providers by introducing a macro


crates/goose/src/providers/factory.rs
```diff
@@ -98,17 +98,16 @@ fn create_lead_worker_from_env(
         .get_param::<usize>("GOOSE_LEAD_FALLBACK_TURNS")
         .unwrap_or(default_fallback_turns());
 
-    // Create model configs with context limit environment variable support
```
michaelneale, line 101: unrelated to this change but feels bad we have a "factory.rs" as well as a "base.rs" - seems very OO/java like or non rust idiomatic but not sure what proper patterns are.

crates/goose/src/model.rs
```diff
@@ -1,323 +1,346 @@
...
+pub enum ConfigError {
+    #[error("Environment variable '{0}' not found")]
+    EnvVarMissing(String),
+    #[error("Invalid value for '{0}': '{1}' - {2}")]
+    InvalidValue(String, String, String),
+    #[error("Value for '{0}' is out of valid range: {1}")]
+    InvalidRange(String, String),
+}
+
+static MODEL_SPECIFIC_LIMITS: Lazy<Vec<(&'static str, usize)>> = Lazy::new(|| {
```
michaelneale, line 17: are we likely to replace this one day with something provider driven? If not I wonder if we should put this in token_limits.rs so it is easy to find, but doesn't really matter either way. 
(some providers can furnish the limits - like openrouter can, but not sure about the general case)

michaelneale:
oh yeah it did get large! nice yeah this makes sense I think and does seem to tidy up the providers and repeated code so LGTM

michaelneale:
hrm - getting different errors than locally - I hope this isn't another wildcard wildgoose chase change...

michaelneale:
odd, cargo check doesn't cover tests, which ... I don't know why it wouldn't (I guess in the past we had most tests macro'd inline)

michaelneale:
ok @DOsinga well it continued to get out of control! even bigger change now

crates/goose/src/model.rs
```diff
@@ -1,323 +1,346 @@
...
-    /// Create a new ModelConfig with the specified model name
-    ///
-    /// The context limit is set with the following precedence:
-    /// 1. Explicit context_limit if provided in config
-    /// 2. Environment variable override (GOOSE_CONTEXT_LIMIT)
-    /// 3. Model-specific default based on model name
-    /// 4. Global default (128_000) (in get_context_limit)
-    pub fn new(model_name: String) -> Self {
-        Self::new_with_context_env(model_name, None)
+    pub fn new(model_name: &str) -> Result<Self, ConfigError> {
```
michaelneale, line 83: @DOsinga this is the very wide reaching change, but it does seem nicely idiomatic rust: (eg https://doc.rust-lang.org/std/fs/struct.File.html#method.open canonical example). Lots of clients to this use `?` which I assume is ok when we just want to let it fail?

michaelneale (approved):
some local basic smoke tests with cli and nothing went boom.

crates/goose/src/providers/factory.rs
DOsinga: there's definitely a better way to do this; lean more into macros is I think one way. ideally I think we'd do something like:

!enabled_providers(OpenAIProvider, AntrhopicProvider, ... )

and it would do the whole factory thing in one sweet go.

base.rs should probably be split into traits and possibly we should move all the actual provider implementations into a subfolder so there's good distinction between support code and implementation. and most providers should just be one macro call, saying, I'm openai compatibile, and this is where my end points are, I support streaming, but not embeddings or some such.

let's do it!

crates/goose/src/model.rs
DOsinga: yeah, I think this goes back to the whole thinking around models <-> providers. in some ideal world I would say we have a models folder which defines all the models and their properties and then providers indicate which of those models they support. llama3-50b-9f has the same properties independent of the provider. I think

crates/goose/src/model.rs
DOsinga: yes exactly

well, fail, we should almost never fail I think but bubble up the exception, which is what ? does. as it was, we were ignoring errors in the provider construction, which I think is strictly worse than throwing an exception even if we don't catch it. if you get the model wrong, it will typically result in a 404 at inference time, which means the user can't do antyhing.

crates/goose-cli/src/commands/configure.rs
```diff
@@ -373,7 +373,7 @@ pub async fn configure_provider_dialog() -> Result<bool, Box<dyn Error>> {
         .map(|val| val == "1" || val.to_lowercase() == "true")
         .unwrap_or(false);
 
-    let model_config = goose::model::ModelConfig::new(model.clone())
+    let model_config = goose::model::ModelConfig::new(&model)?
```
DOsinga, line 376: we probably just want to skip this model if we can't load it

crates/goose-server/src/routes/agent.rs
```diff
@@ -255,7 +255,7 @@ async fn update_agent_provider(
             .get_param("GOOSE_MODEL")
             .expect("Did not find a model on payload or in env to update provider with")
     });
-    let model_config = ModelConfig::new(model);
+    let model_config = ModelConfig::new(&model).map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
```
DOsinga, line 258: the line above where we have .expect() will kill goosed and it is all over. not sure how we would have gotten here, but still, we should probably do another 500; I also think we should change the description of the 500 from internal server error here to something descriptive; could not update provider, make sure that you have the right config or something

Issue #3652, opened 2025-07-25
Claude 4 responding with response in markdown block occasionally


zanesq:
**Describe the bug**
Occasionally with Claude 4 it has the response wrapped in a code/markdown block?.

**To Reproduce**
Choose Claude 4, Chat and eventually you will get a response in a code/markdown block like the screenshot.

**Expected behavior**
Should respond with normal rendering

**Screenshots**
<img width="1468" height="1113" alt="Image" src="https://github.com/user-attachments/assets/c32680d7-41c1-4087-9fb2-f4d7004625ba" />

**Please provide following information:**
 - **OS & Arch:** Mac OS
 - **Interface:** UI
 - **Version:** 0.36-1.1.4
 - **Extensions enabled:** Developer tools, computer controller
 - **Provider & Model:** Claude 4 sonnet

**Additional context**
I wonder if they changed something in how they typically output markdown blocks that our code isn't recognizing correctly, or if the model is incorrectly outputting markdown?

PR #3720, opened 2025-07-29
Fix html content detection regex to not include markdown autolinks

zanesq:
The html detection regex was too broad and it was catching autolinks inside markdown and rendering it as a code block. Also added support for various edge cases for rendering messages.

fixes https://github.com/block/goose/issues/3652

Bonus: added tests with vitest to kick off our unit testing 🎉


michaelneale (approved):
very nice - love the tests, any tips on how to manually validate this if needed?
I know I have seen this for a while.

zanesq:
I manually tested it by asking goose to respond with each rendering case in a dev build and verified the messages are rendering properly.

PR #3721, opened 2025-07-29
Token counting reliability + summarization integration.

katzdave:
- Use the token counter as a fallback for when providers don't provide tokens; Ensures we always have some token counts.
- Remove old summarization algorithm completely. Only have the single shot. Simplify the contract since it only returns 1 message.
- Update metadata with summarization results for immediate visibility in context window display.
- Remove some standalone codepaths testing tokenizer.

TODO (future PR):
Audit desktop more closely and make sure we have full parity + clean up the UI around context display.


michaelneale:
ok I like this in theory - main worry is that we have many bugs in actual token counting with streaming! but other than that, this should work a lot better

crates/goose/src/prompts/summarize_oneshot.md
```diff
@@ -2,6 +2,9 @@
 Generate detailed summary of conversation to date.  
 Include user requests, your responses, and all technical content.  
 
+**Conversation History:**
```
michaelneale, line 5: how did this work before?

katzdave: Trying to understand that one too; I think we are falling back to the old summarization algorithm 😱.

DOsinga: why do we still have the old algorithm

katzdave: This deletes it for good

michaelneale: was the session passed some other way to the compact call?

crates/goose-server/src/routes/context.rs
```diff
@@ -67,7 +67,7 @@ async fn manage_context(
             .await
             .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
     } else if request.manage_action == "summarize" {
-        (processed_messages, token_counts) = agent
+        (processed_messages, token_counts, _) = agent
```
DOsinga, line 70: Does this still use the old algorithm? any reason not to switch this over?

also, the truncate above is not used i think

katzdave: I'm not clear on truncate. Will talk to @michaelneale about it but I think makes sense to not touch that until the next wave.

This uses the same call as the old algorithm made, but summarize.rs is completely gutted to just have the single shot summarizer.

crates/goose/src/prompts/summarize_oneshot.md
katzdave: Ugh, nope just did a full investigation. We are calling summarize_messages_async instead of summarize_messages; All the tests swap over but didn't make the swap in agent.rs + delete summarize_messages_async. 

Can potentially break this up and swap the summarization algorithm first, but the numbers that show up in the auto-compact message + context window aren't the best atm.

crates/goose-server/src/routes/context.rs
michaelneale: I think truncate may have veen used on CLI for people using recipes, but as long as summarize works out of the box, it is ok to clean it out I think (as ideally we don't want people having to specify it, in the past it would just break if you didn't specify a strategy). so I think ok to get rid of it if it helps?

michaelneale: it can always be done with a later release

crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -126,47 +120,6 @@ pub async fn check_compaction_needed(
...
-/// capabilities. It assumes compaction is needed and should be called after
-/// `check_compaction_needed` confirms it's necessary.
-///
-/// # Arguments
-/// * `agent` - The agent to use for context management
-/// * `messages` - The current message history to compact
-///
-/// # Returns
-/// * Tuple of (compacted_messages, tokens_before, tokens_after)
-pub async fn perform_compaction(
```
michaelneale, line 141: @katzdave where did this functionality go, this was dead code?

michaelneale:
saw it kick in at what I thought was 10%:

image

(that is different message to before and seemed fast, and was after some tool calls I think)

not sure if that is still a bug or not, things seem to work well though.

crates/goose-server/src/routes/context.rs
katzdave: Yeah I was just going to pick it apart on the next wave since this PR is already large.

crates/goose/src/context_mgmt/auto_compact.rs
katzdave: It was just a thin wrapper around agent.summarize with some logging so got rid of it.

PR #3722, opened 2025-07-29
Fix slow typing in chat input with long running sessions

zanesq:
Some users are reporting that the chat input is sometimes slow to type into when in a long running session. 

Optimized the chat input by removing a debounce and memoizing and cleaning up potential memory leaks in hooks.


michaelneale (approved):
nice - I note we have been bitten by this before so must have crept back in

PR #3727, opened 2025-07-30
chore: Also fix caching for the mac app build

jamadeo:


jamadeo:
Doesn't speed up, probably because release config is always doing a full recompile

PR #3728, opened 2025-07-30
fix: Disabled route strategy for recipe in goose cli

lifeizhou-ap:
**Why**
This PR addresses the issue https://github.com/block/goose/issues/3730

**What**
- The fix in this PR only fixes the CLI
- Currently introduces `router_disabled_override` on agent.rs.   However, I will create followup PRs to refactor agent and simplify the logics


lifeizhou-ap:
The changes have been included in https://github.com/block/goose/pull/3732

Issue #3713, opened 2025-07-29
Can you add DeepSeek as model provider?


george-block:
**Please explain the motivation behind the feature request.**
DeepSeek as model provider is very affordable, this could enable a lot of people like me to switch to Goose as Goose could save more tokens based on that

**Describe the solution you'd like**
DeepSeek as a model provider show in configure, asking me to input API_KEY and choose a model from list returned by DeepSeek API

**Describe alternatives you've considered**
Nope, other providers like OpenRouter do provide DeepSeek models but the cost is higher

**Additional context**
https://api-docs.deepseek.com/

- [x] I have verified this does not duplicate an existing feature request

jamadeo:
With #4099, this should now be possible.

PR #3731, opened 2025-07-30
#3713 : add deepseek as model provider

georgetso:
add deepseek as model provider


DOsinga:
we have a wild growth of providers - is deepseek mostly openai compatible? could we change to a way where we do openai compatible providers using a json document?

DOsinga:
can we in the future based this on https://github.com/block/goose/pull/3825 instead?

DOsinga:
we have custom providers now. is that enough? if not, can you implement this in terms of custom providers?

PR #3733, opened 2025-07-30
Must have missed this one

DOsinga:
Somehow this one slipped through


michaelneale:
ah yeah this would have been caught if the other PR was updated to main I think, bad luck

PR #3735, opened 2025-07-30
chore: use a Conversation type

jamadeo:
Introduces a Conversation type to use over Vec<Message>. This lets us keep message merging and validation within the type.

For now this is a thin wrapper around Vec<Message> and doesn't do the validation yet.

### DIFF
--- a/crates/goose/src/context_mgmt/summarize.rs
+++ b/crates/goose/src/context_mgmt/summarize.rs
@@ -1,4 +1,5 @@
 use super::common::{get_messages_token_counts, get_messages_token_counts_async};
+use crate::conversation::Conversation;
 use crate::message::{Message, MessageContent};
 use crate::providers::base::Provider;
 use crate::token_counter::{AsyncTokenCounter, TokenCounter};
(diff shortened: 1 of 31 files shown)


crates/goose/src/context_mgmt/summarize.rs
```diff
@@ -95,68 +98,10 @@ fn preprocess_messages(messages: &[Message]) -> (Vec<Message>, Vec<Message>) {
...
-    final_messages
-}
-
-// Summarization steps:
-// 1. Break down large text into smaller chunks (roughly 30% of the model’s context window).
-// 2. For each chunk:
-//    a. Combine it with the previous summary (or leave blank for the first iteration).
-//    b. Summarize the combined text, focusing on extracting only the information we need.
-// 3. Generate a final summary using a tailored prompt.
-pub async fn summarize_messages(
```
jamadeo: this wasn't used except in tests

crates/goose/src/conversation.rs
```diff
@@ -0,0 +1,109 @@
+use serde::{Deserialize, Serialize};
```
DOsinga: we have a message folder - we could rename that to conversation and move this there and also the conversation fixer

DOsinga (approved):

DOsinga:
you should probably get this in when you get a chance seeing how quickly this bit of the code moves

PR #3742, opened 2025-07-30
Make the client more secure

DOsinga:
We were storing the server secret in the appConfig that was transported through extra params to the individual windows which means they become part of the parameters which makes the secret visible for any process that can see other processes.

```
douwe            67450   0.0  0.2 1621913168 111024   ??  S     8:22PM   0:40.91 /Applications/Goose.app/Contents/Frameworks/Goose Helper (Renderer).app/Contents/MacOS/Goose Helper (Renderer) --type=renderer --user-data-dir=/Users/douwe/Library/Application Support/Goose --app-path=/Applications/Goose.app/Contents/Resources/app.asar --enable-sandbox --lang=en-US --num-raster-threads=4 --enable-zero-copy --enable-gpu-memory-buffer-compositor-resources --enable-main-frame-before-activation --renderer-client-id=4 --time-ticks-at-unix-epoch=-1752795514407452 --launch-time-ticks=1104237712001 --shared-files --field-trial-handle=1718379636,r,517907852739083296,9621107575574687593,262144 --enable-features=ScreenCaptureKitPickerScreen,ScreenCaptureKitStreamPickerSonoma --disable-features=MacWebContentsOcclusion,SpareRendererForSitePerProcess,TimeoutHangingVideoCaptureStarts --variations-seed-version {"GOOSE_DEFAULT_PROVIDER":"databricks","GOOSE_DEFAULT_MODEL":"goose-claude-4-sonnet","GOOSE_PREDEFINED_MODELS":"[{\"id\":1,\"name\":\"goose-claude-4-sonnet\",\"provider\":\"databricks\",\"alias\":\"claude-4-sonnet (recommended)\",\"subtext\":\"Anthropic\"},{\"id\":2,\"name\":\"goose-claude-3-5-sonnet\",\"provider\":\"databricks\",\"alias\":\"claude-3.5-sonnet\",\"subtext\":\"Anthropic\"},{\"id\":3,\"name\":\"goose-claude-4-opus\",\"provider\":\"databricks\",\"alias\":\"claude-4-opus\",\"subtext\":\"Anthropic\"},{\"id\":4,\"name\":\"goose-gpt-4-1\",\"provider\":\"databricks\",\"alias\":\"GPT-4.1\",\"subtext\":\"OpenAI\"},{\"id\":5,\"name\":\"goose-o3\",\"provider\":\"databricks\",\"alias\":\"o3\",\"subtext\":\"OpenAI\"},{\"id\":6,\"name\":\"goose-claude-3-7-sonnet\",\"provider\":\"databricks\",\"alias\":\"claude-3.7-sonnet\",\"subtext\":\"Anthropic\"},{\"id\":7,\"name\":\"goose-gemini-2-5-pro\",\"provider\":\"databricks\",\"alias\":\"Gemini-2.5-pro\",\"subtext\":\"Google\"},{\"id\":8,\"name\":\"goose-gpt-4o\",\"provider\":\"databricks\",\"alias\":\"GPT-4o\",\"subtext\":\"OpenAI\"},{\"id\":9,\"name\":\"goose-o4-mini\",\"provider\":\"databricks\",\"alias\":\"o4-mini\",\"subtext\":\"OpenAI\"}]","GOOSE_API_HOST":"http://127.0.0.1","GOOSE_PORT":62066,"GOOSE_WORKING_DIR":"/Users/douwe/proj/goose/goose","GOOSE_ALLOWLIST_WARNING":false,"secretKey":"d59013a9d1d07631f34c14ad6851160477201d27e33df753f7fb4a47a52d36c4","REQUEST_DIR":"/Users/douwe/proj/goose/goose","GOOSE_BASE_URL_SHARE":"https://goosed.sqprod.co/api","GOOSE_VERSION":"1.1.4-block"} --seatbelt-client=61
```

This centralizes the secret in one place and provides access to it through the electron protocol.


jamadeo (approved):
Nice. When I searched the right way to do this, electron IPC was suggested, so this seems right!

shellz-n-stuff (approved):
Defs an improvement getting it out of the config. 

Out of interest can we not have an interceptor for requests defined to keep the value only being handled/used in one place? Given it's electron could be a nice protection from leaking via XSS

ui/desktop/src/goosed.ts
```diff
@@ -282,46 +269,31 @@ export const startGoosed = async (
...
-  if (!isReady) {
-    log.error(`Goosed server failed to start on port ${port}`);
+
+  const try_kill_goose = () => {
     try {
       if (isWindows) {
-        // On Windows, use taskkill to forcefully terminate the process tree
-        // Security: Validate PID is numeric and use safe arguments
         const pid = goosedProcess.pid?.toString() || '0';
-        if (!/^\d+$/.test(pid)) {
```
michaelneale, line 292: will this be needed on windows?

michaelneale: ah - centralising it

michaelneale:
"Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)"

(which is causing the bundle failure)

ui/desktop/src/components/ConfigContext.tsx
```diff
@@ -28,14 +28,17 @@ export type FixedExtensionEntry = ExtensionConfig & {
...
+  client.setConfig({
+    baseUrl: window.appConfig.get('GOOSE_API_HOST') + ':' + window.appConfig.get('GOOSE_PORT'),
+    headers: {
+      'Content-Type': 'application/json',
+      'X-Secret-Key': await window.electron.getSecretKey(),
+    },
+  });
+}
+
+await initializeClient();
```
michaelneale: this is what stops it from bundling with the error (you can see it with `just make-ui`)

DOsinga: yeah I am aware. it ran fine for me though. but can move to app.tsx

ui/desktop/src/goosed.ts
DOsinga: well, it strikes me that either we can kill the thing or we can't. why would we try to guess what a pid looks like if we got it from the os and pass it to the os?

ui/desktop/src/components/ConfigContext.tsx
michaelneale: (might have a fix... standby)

michaelneale: ok https://github.com/vitejs/vite/issues/6985 - says it isn't a vite issue (hrm...)

seems ok to set the build target to "newer" stuff, given this is an electron app, not a regular web app (otherwise can just not to await there...)

image

(just kidding, this was all by hand)

michaelneale: would be nice if the build https://github.com/block/goose/actions/runs/16632635914/job/47065764905?pr=3742#step:19:131 would fail at the right point (not that the test launch app stage)

but still - good to catch this now vs later.

DOsinga: na, it is easy, just call it from the useEffect below instead. probably should have done that always

PR #3745, opened 2025-07-31
chore: upgrade morph to use new model with instruction

michaelneale:
the replaces: https://github.com/block/goose/pull/3471 

change to editor model only


crates/goose-mcp/src/developer/editor_models/morphllm_editor.rs
```diff
@@ -19,6 +19,42 @@ impl MorphLLMEditor {
             model,
         }
     }
+
+    fn format_user_prompt(original_code: &str, update_snippet: &str) -> String {
+        if let (Some(code_start), Some(code_end)) = (
+            update_snippet.find("<code>"),
+            update_snippet.find("</code>"),
+        ) {
+            // Check if we have valid code tags
```
jamadeo: probably don't need this comment

crates/goose-mcp/src/developer/editor_models/morphllm_editor.rs
```diff
@@ -19,6 +19,42 @@ impl MorphLLMEditor {
...
+            if code_start < code_end {
+                let code_content = &update_snippet[code_start + 6..code_end].trim();
+
+                // Look for instruction tags which help provide hints
+                if let (Some(inst_start), Some(inst_end)) = (
+                    update_snippet.find("<instruction>"),
+                    update_snippet.find("</instruction>"),
```
jamadeo: since this does the same as the code tag extraction, you could extract to something like

```
fn extract_tag_content<'a>(content: &'a str, tag: &str) -> Option<&'a str> {
    let start = content.find(&format!("<{}>", tag))? + tag.len() + 2;
    let end = content[start..].find(&format!("</{}>", tag))?;
    Some(&content[start..start + end])
}
```

jamadeo (approved):

PR #3749, opened 2025-07-31
fix: cli tool logging

jamadeo:
There is a check here to write the output differently if it's a sub-agent, but with .unwrap_or(), we'd always hit that path

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -1113,12 +1113,10 @@ impl Session {
                                             }
                                         }
                                     }
-                                    else {
-                                        if output::is_showing_thinking() {
-                                            output::set_thinking_message(&formatted_message);
-                                        } else {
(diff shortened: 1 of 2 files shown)


crates/goose-cli/src/session/mod.rs
```diff
@@ -1079,7 +1078,7 @@ impl Session {
                                                         msg.to_string()
                                                     }
                                                 };
-                                                (formatted, Some(subagent_id.to_string()), Some(notification_type.to_string()))
+                                                (formatted, subagent_id.map(|s|s.to_string()), Some(notification_type.to_string()))
```
DOsinga: no changes to notificaation type?

DOsinga (approved):

crates/goose-cli/src/session/mod.rs
jamadeo: yeah looking below, that should probably also not be coerced to Some

PR #3750, opened 2025-07-31
fix: don't return full shell output when very large

michaelneale:
The shell (bash) command can result in a lot of pointless context filling, if is running a compile, or log tailing and so on. 

In reality only some of it is relevant, in this case it will limit it to around 100 lines (a few percentage points of context) and allow it to then search for more content when needed. 

For example, you can tell it to run a command like: 

```can you run find /System/Library -type f | head -n 2000```

(that was just an example I could consistently use) 

by default - this eats half the context window in one fell swoop: 
image

with this change: 

image

that is a contrived example but it isn't uncommon to hit this.

### DIFF
--- a/crates/goose-mcp/src/developer/mod.rs
+++ b/crates/goose-mcp/src/developer/mod.rs
@@ -599,6 +599,9 @@ impl DeveloperRouter {
         let lines: Vec<&str> = output_str.lines().collect();
         let line_count = lines.len();
 
+        let start = lines.len().saturating_sub(100);
+        let last_100_lines_str = lines[start..].join("\n");
+
         let final_output = if line_count > 100 {
             let tmp_file = tempfile::NamedTempFile::new().map_err(|e| {
                 ToolError::ExecutionError(format!("Failed to create temporary file: {}", e))
@@ -612,21 +615,18 @@ impl DeveloperRouter {
                 ToolError::ExecutionError(format!("Failed to persist temporary file: {}", e))
             })?;
 
-            let last_100_lines: Vec<&str> = lines.iter().rev().take(100).rev().copied().collect();
-
             format!(
-                "private note: output was {} lines and we are only showing the most recent lines, remainder of lines in {}. do not show tmp file to user, that file can be searched if extra context needed to fulfill request. truncated output: \n{}",
+                "private note: output was {} lines and we are only showing the most recent lines, remainder of lines in {} do not show tmp file to user, that file can be searched if extra context needed to fulfill request. truncated output: \n{}",
                 line_count,
                 path.display(),
-                last_100_lines.join("\n")
+                last_100_lines_str
             )
         } else {
             output_str.to_string()
         };
 
         let user_output = if line_count > 100 {
-            let last_100_lines: Vec<&str> = lines.iter().rev().take(100).rev().copied().collect();
-            format!("... \n{}", last_100_lines.join("\n"))
+            format!("... \n{}", last_100_lines_str)
         } else {
             output_str.to_string()
         };
@@ -1717,7 +1717,7 @@ mod tests {
     use super::*;
     use serde_json::json;
     use serial_test::serial;
-    use std::fs;
+    use std::fs::{self, read_to_string};
     use tempfile::TempDir;
     use tokio::sync::OnceCell;
 
@@ -3287,6 +3287,36 @@ mod tests {
         assert!(user_content.text.contains("Line 51"));
         assert!(user_content.text.contains("Line 150"));
         assert!(!user_content.text.contains("Line 50"));
+
+        println!("assistant output: {}", assistant_content.text);
+
+        let start_tag = "remainder of lines in";
+        let end_tag = "do not show tmp file to user";
+
+        if let (Some(start), Some(end)) = (
+            assistant_content.text.find(start_tag),
+            assistant_content.text.find(end_tag)
+        ) {
+            let start_idx = start + start_tag.len();
+            if start_idx < end {
+                let path = assistant_content.text[start_idx..end].trim();
+                println!("Extracted path: {}", path);
+            }
+            let file_contents = read_to_string(path).expect("Failed to read extracted temp file");
+
+            let lines: Vec<&str> = file_contents.lines().collect();
+
+            // Ensure we have exactly 150 lines
+            assert_eq!(lines.len(), 150, "Expected 150 lines in temp file");
+
+            // Ensure the first and last lines are correct
+            assert_eq!(lines.first(), Some(&"Line 1"), "First line mismatch");
+            assert_eq!(lines.last(), Some(&"Line 150"), "Last line mismatch");
+        }
+
+        
+
+
 
         temp_dir.close().unwrap();
     }
--- a/crates/goose-mcp/src/developer/mod.rs
+++ b/crates/goose-mcp/src/developer/mod.rs
@@ -577,6 +577,51 @@ impl DeveloperRouter {
         self.ignore_patterns.matched(path, false).is_ignore()
     }
 
+    // shell output can be large, this will help manage that
+    fn process_shell_output(&self, output_str: &str) -> Result<(String, String), ToolError> {
+        let lines: Vec<&str> = output_str.lines().collect();
+        let line_count = lines.len();
+
+        let final_output = if line_count > 100 {
+            // Create a temporary file with the full output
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: crates/goose-mcp/src/developer/mod.rs [hunk 1]
```diff
@@ -577,6 +577,46 @@ impl DeveloperRouter {
...
+                ToolError::ExecutionError(format!("Failed to create temporary file: {}", e))
+            })?;
+
+            std::fs::write(tmp_file.path(), output_str).map_err(|e| {
+                ToolError::ExecutionError(format!("Failed to write to temporary file: {}", e))
+            })?;
+
+            let (_, path) = tmp_file.keep().map_err(|e| {
+                ToolError::ExecutionError(format!("Failed to persist temporary file: {}", e))
+            })?;
```
DOsinga, line 616: this is rather wordy, can't we do something like?

let path = (|| -> Result<_, Box<dyn std::error::Error>> {
    let tmp_file = tempfile::NamedTempFile::new()?;
    std::fs::write(tmp_file.path(), output_str)?;
    Ok(tmp_file.keep()?.1)
})().map_err(|e| ToolError::ExecutionError(format!("Failed to create temp file: {}", e)))?;

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -577,6 +577,46 @@ impl DeveloperRouter {
...
+        } else {
+            output_str.to_string()
+        };
+
+        let user_output = if line_count > 100 {
+            let last_100_lines: Vec<&str> = lines.iter().rev().take(100).rev().copied().collect();
+            format!("... \n{}", last_100_lines.join("\n"))
+        } else {
+            output_str.to_string()
+        };
```
DOsinga, line 632: is this duplicated code? remnant of a previous attempt?

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -3139,4 +3181,141 @@ mod tests {
...
+
+        // User output should contain lines 51-150 (last 100 lines)
+        assert!(user_content.text.contains("Line 51"));
+        assert!(user_content.text.contains("Line 150"));
+        assert!(!user_content.text.contains("Line 50"));
+
+        temp_dir.close().unwrap();
+    }
+
+    #[test]
```
DOsinga, line 3325: remove these other tests; I don't think they add anything

crates/goose-mcp/src/developer/mod.rs
michaelneale: I think it looks more wordy! (not throwing away a return though)

crates/goose-mcp/src/developer/mod.rs
michaelneale: no - we need that as we have the user and assistant response audiences - that in both cases is the fall through when a small response which is ok logically at least

michaelneale: could shift to Maybe/option and it would reduce one of them ie final_output == description + lines | full raw. User output == final lines | raw

crates/goose-mcp/src/developer/mod.rs
michaelneale: yep - although keeping short (as that validates existing) and the empty case always good to test (this is where something like QuickCheck in haskell could be great as it would be just one set of properties and it fuzz tests them all)

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -594,6 +594,46 @@ impl DeveloperRouter {
...
+                ToolError::ExecutionError(format!("Failed to persist temporary file: {}", e))
+            })?;
+
+            let last_100_lines: Vec<&str> = lines.iter().rev().take(100).rev().copied().collect();
+
+            format!(
+                "private note: output was {} lines and we are only showing the most recent lines, remainder of lines in {}. do not show tmp file to user, that file can be searched if extra context needed to fulfill request. truncated output: \n{}",
+                line_count,
+                path.display(),
+                last_100_lines.join("\n")
```
cgwalters: Not that this truly matters but since we're talking about efficiency, you don't need to collect into the intermediate Vec above to do this.

I asked Gemini e.g. and it came up with

```
// Calculate the start index, preventing underflow if lines < 100
let start = lines.len().saturating_sub(100);

// Slice the lines and join them into a string directly
let last_100_lines_str = lines[start..].join("\n");
```

But even then we can still optimize further by using `format!` for the message and then appending each line to the string.

Also...in this whole vein we probably want to handle pathological cases like extremely long lines.

michaelneale: yeah efficiency here isn't really the rust side, but could be tweaked, good catch. 

yes long lines - I think can do another PR for that, as I have seen that (ie ask goose to look at your session jsonl files - they are long! Would like to address that. Perhaps along with the text_editor_view tool as well?

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -594,6 +594,46 @@ impl DeveloperRouter {
...
+
+            std::fs::write(tmp_file.path(), output_str).map_err(|e| {
+                ToolError::ExecutionError(format!("Failed to write to temporary file: {}", e))
+            })?;
+
+            let (_, path) = tmp_file.keep().map_err(|e| {
+                ToolError::ExecutionError(format!("Failed to persist temporary file: {}", e))
+            })?;
+
+            let last_100_lines: Vec<&str> = lines.iter().rev().take(100).rev().copied().collect();
```
cgwalters: Also, in Rust iterators are lazy, so if you drop the `.collect()` this could be moved right after `let lines` above (line 599), and then we could only do the collect here and in the other copy below.

michaelneale: yep - am replacing this

michaelneale:
thanks @cgwalters for optimisation - will do a fast follow for long docs/lines I think

ui/desktop/src/components/context_management/index.ts
```diff
@@ -60,7 +60,7 @@ export function convertApiMessageToFrontendMessage(
     sendToLLM: sendToLLM ?? true,
     id: generateId(),
     role: apiMessage.role as Role,
-    created: apiMessage.created,
+    created: apiMessage.created ?? 0,
```
michaelneale, line 63: @DOsinga @zanesq not sure about this one - but something had to be here as it seems to now be optional? I don't quite understand the change earlier

cgwalters:
:arrow_right: https://github.com/block/goose/pull/3816

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -594,6 +594,46 @@ impl DeveloperRouter {
...
+        let final_output = if line_count > 100 {
+            let tmp_file = tempfile::NamedTempFile::new().map_err(|e| {
+                ToolError::ExecutionError(format!("Failed to create temporary file: {}", e))
+            })?;
+
+            std::fs::write(tmp_file.path(), output_str).map_err(|e| {
+                ToolError::ExecutionError(format!("Failed to write to temporary file: {}", e))
+            })?;
+
+            let (_, path) = tmp_file.keep().map_err(|e| {
```
cgwalters, line 614: There's nothing that's garbage collecting these temporary files (apart from the OS level tempfile cleaner right?)

That seems very suboptimal...I did this IMO a better way in https://github.com/block/goose/pull/2817

cgwalters: How long to retain the tempfiles is an interesting topic - in #2817 notice that the tempfile is owned in Rust by the mcp server (so we get cleanup on drop) and the semantics basically are that the next shell command that outputs large data replaces it.

My thought is that in a flow of doing `make/cargo/whatever build` (gets redirected), then the model can do e.g. `grep <tmpfile> error:` and as long as *that* doesn't overflow 100 lines the tempfile is preserved, which should hopefully be a common case. But it might needs careful explaining to the model, I didn't really hammer on it from that angle.

Leaking the tempfiles entirely avoids that problem but...yeah I don't think we want to do that.

michaelneale: yeah the os will clean them up which is preferred (and isn't the end of the world if they are gone in an old session, hopefully we got the useful information). Anything more durable and yeah it should be in the session structure itself (but not sent to the LLM)

PR #3753, opened 2025-07-31
Handle missing keys with a default

### LINKED ISSUE #3551: Windows UI: Multiple errors shown in console log obfuscating real issues
**Describe the bug**

The Windows UI console log shows multiple errors , while the applications seems to work fine.
This obfuscates any real errors users experience. Thus need to make sure that the console log is clear of any errors, possibly allow for warnings if relevant.

**To Reproduce**
1. Login to Goose
2. Open developer tools
3. You already see some fonts not found
4. and config/read failing
5. Just do a simple chat
6. See more of the same errors being logged

**Expected behaviour**
No errors in the console log, unless there is a real error

**Screenshots**
Image

 - **OS & Arch:** Windows 11 x86
 - **Interface:** UI
 - **Version:** v1.1.3
 - **Extensions enabled:** Developer, Reddit
 - **Provider & Model:** OpenAI / gpt-4.1-mini

### LINKED ISSUE #3679: Bug: Multiple 404 Not Found errors for /config/read on startup
### **Describe the bug**
On startup, the Goose desktop application frequently shows multiple `404 Not Found` errors for network requests to the `/config/read` endpoint. This happens on a dynamically allocated port (e.g., `127.0.0.1:64470`) which changes with each application session.

I am reporting this because these errors suggest a potential timing issue during the app's initialization.

⚠️  **Important note:** The application's core functionality is still working correctly after these errors occur. However, I'm not sure if this has other, less obvious side effects on things like extension loading or configuration.

### **To Reproduce**

1.  Ensure you have a clean session (quit the application completely).
2.  Open the Goose desktop application.
3.  Open the developer tools (using `Cmd+Option+I` on macOS or `Ctrl+Shift+I` on Windows/Linux).
4.  Navigate to the "Console" and "Network" tabs.
5.  Observe the logs upon startup. You will see several `Failed to load resource: the server responded with a status of 404 (Not Found)` errors for the `/config/read` endpoint.

### **Expected behavior**

The application should start up cleanly without any 404 errors. All necessary configuration should be read successfully on the first attempt, allowing the application to initialize correctly and display the main chat or welcome view.

### **Screenshots**

The following screenshot of the developer console illustrates the issue, showing the sequence of initialization logs followed by multiple failed requests to `/config/read`:

<details><summary>View Screenshots</summary>
<p>

Image

Image

</p>
</details> 

### **Please provide the following information:**
*   **OS & Arch:** macOS
*   **Interface:** UI
*   **Version:** 1.1.4
*   **Extensions enabled:** Default built-in extensions (the issue occurs before sending the first message)
*   **Provider & Model (1/3):**  OpenRouter / Google Gemini2.5,  Claude 3.7  (as observed in logs, but the issue is provider-agnostic).
*   **Provider & Model (2/3):**  Anthropic / Claude 3.7  (issue is provider-agnostic).
*   **Provider & Model (3/3):**  OpenAI / o4-mini  (issue is provider-agnostic).

### **Additional context 🪿🕵🏻‍♂️**

<details><summary> Further Observations and Console Logs</summary>
<p>

Here are a few additional observations Goose made, that might be helpful for debugging:

1.  **Dynamic Port Allocation:** The backend `goosed` process is started on a random, available port for each new application window. The frontend is correctly configured to communicate with this port.
2.  **Race Condition:** The errors occur because multiple components in the frontend (`App.tsx`) attempt to read configuration settings concurrently during the initial `useEffect` hook. The console log `Setting up extension handler` appears immediately before the errors, pointing to the initialization of extensions as a primary trigger.
3.  **Root Cause:** The core problem is that these configuration requests are fired off without waiting for the backend server to be fully initialized and ready to serve requests. This results in the server returning 404s because it's not yet able to handle the `/config/read` route when the requests arrive. The multiple, simultaneous `read()` calls from different parts of the initialization code exacerbate the problem.

</p>
</details>

DOsinga:
we were returning a 404 for missing keys in the server which is the not found code, so not unreasonable. except that the client expects null for not found, so made it so. now no more 404s showing up in the developer console.

fixes #3679 
fixes #3551

### DIFF
--- a/crates/goose-server/src/routes/config_management.rs
+++ b/crates/goose-server/src/routes/config_management.rs
@@ -7,9 +7,9 @@ use axum::{
     Json, Router,
 };
 use etcetera::{choose_app_strategy, AppStrategy};
-use goose::config::Config;
 use goose::config::APP_STRATEGY;
 use goose::config::{extensions::name_to_key, PermissionManager};
+use goose::config::{Config, ConfigError};
 use goose::config::{ExtensionConfigManager, ExtensionEntry};
 use goose::model::ModelConfig;
 use goose::providers::base::ProviderMetadata;
@@ -142,7 +142,7 @@ pub async fn remove_config(
     request_body = ConfigKeyQuery,
     responses(
         (status = 200, description = "Configuration value retrieved successfully", body = Value),
-        (status = 404, description = "Configuration key not found")
+        (status = 500, description = "Unable to get the configuration value"),
     )
 )]
 pub async fn read_config(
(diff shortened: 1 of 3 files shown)


DOsinga:
yeah good question about the testing. this is specific to the api call - we don't do a lot of testing on these routes. this one in particular would be a lot of mocking for little gain

PR #3760, opened 2025-07-31
Stable goose info -v

### LINKED ISSUE #2747: goose info -v print ordering is non-deterministic
**Describe the bug**

Note: Please check the common issues on https://block.github.io/goose/docs/troubleshooting before filing a report

**To Reproduce**
run goose info -v 
run goose info -v again

**Expected behavior**
Expected both invocations to print identical yaml text

**Screenshots**

 - **OS & Arch: Ubuntu on ARM
 - **Interface:** CLI
 - **Version:** 1.0.24
 - **Extensions enabled:**  N/A
 - **Provider & Model:** N/A

DOsinga:
fixes #2747

### DIFF
--- a/crates/goose-cli/src/commands/info.rs
+++ b/crates/goose-cli/src/commands/info.rs
@@ -51,9 +51,14 @@ pub fn handle_info(verbose: bool) -> Result<()> {
                         "  Run '{}' to configure goose",
                         style("goose configure").cyan()
                     );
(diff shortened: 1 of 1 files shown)


DOsinga:
> This is a good change but it's made me realize that we also save the config file with a non-deterministic ordering. If you return a sorted map from load_values would it also fix that?

true. but I was talking to @katzdave that we should redo how we do settings anyway

PR #3761, opened 2025-07-31
Fix leaky env variable causing flaky test

katzdave:
I'm not sure what's actually causing this to leak, but this seems to get rid of the flakiness from having the value:

GOOSE_TEMPERATURE="hot" set from a different test.

Also removed an unused symbol warning.

### DIFF
--- a/crates/goose/src/model.rs
+++ b/crates/goose/src/model.rs
@@ -247,19 +247,28 @@ mod tests {
 
     #[test]
     fn test_model_config_context_limits() {
-        let config = ModelConfig::new("claude-3-opus")
-            .unwrap()
-            .with_context_limit(Some(150_000));
-        assert_eq!(config.context_limit(), 150_000);
+        // Clear all GOOSE environment variables to ensure clean test environment
+        with_var("GOOSE_TEMPERATURE", None::<&str>, || {
+            with_var("GOOSE_CONTEXT_LIMIT", None::<&str>, || {
+                with_var("GOOSE_TOOLSHIM", None::<&str>, || {
+                    with_var("GOOSE_TOOLSHIM_OLLAMA_MODEL", None::<&str>, || {
+                        let config = ModelConfig::new("claude-3-opus")
+                            .unwrap()
+                            .with_context_limit(Some(150_000));
+                        assert_eq!(config.context_limit(), 150_000);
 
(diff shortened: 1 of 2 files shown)


katzdave:
Ehh, seems like the pipeline is mostly passing now... Will keep an eye on it and if things seem stable will close this.

DOsinga:
either way, we need to kill al the environment variables, I'm telling you

cgwalters:
> either way, we need to kill al the environment variables,

Yes with Rust 2024 it will be officially unsafe https://doc.rust-lang.org/edition-guide/rust-2024/newly-unsafe-functions.html#stdenvset_var-remove_var

That said there's possibilities:

- Change all internal APIs that want an environment to support having one passed in, then use a mock environment in the unit tests (this probably wouldn't be too hard)
- Change the test to fork+exec (tricky but doable)
- Only run these tests via https://nexte.st/  (ref https://nexte.st/docs/design/why-process-per-test/ )

crates/goose/src/model.rs
```diff
@@ -247,19 +247,28 @@ mod tests {
...
-        let config = ModelConfig::new("claude-3-opus").unwrap();
-        assert_eq!(config.context_limit(), 200_000);
-
-        let config = ModelConfig::new("gpt-4-turbo").unwrap();
-        assert_eq!(config.context_limit(), 128_000);
-
-        let config = ModelConfig::new("unknown-model").unwrap();
-        assert_eq!(config.context_limit(), DEFAULT_CONTEXT_LIMIT);
+        // Clear all GOOSE environment variables to ensure clean test environment
+        with_var("GOOSE_TEMPERATURE", None::<&str>, || {
```
DOsinga, line 251: this with_var looks nice, but shouldn't this be at the other test that actually sets these? of course relying on these in tests is crazy to begin with

ceterum censeo variabiles ambientis esse delendas

DOsinga (approved):

DOsinga:
> * Change all internal APIs that want an environment to support having one passed in, then use a mock environment in the unit tests (this probably wouldn't be too hard

probably this one. or really, no internal api should rely on environment variables, they should just declare what there inputs are. then we can read those at a much higher level from the settings thing. and that that one can if it wants take environment variables into account

PR #3764, opened 2025-07-31
Clear the current line on ctrl-c in line with other tools

### LINKED ISSUE #1479: Move away from ctrl+c to exit, allow it to clear chat and use ctrl+d instead.
Tool is challenging with ctrl+c a common clear chat, ctrl+d is a great alternative to allow ctrl+c to clear chat, and ctrl+d exit program.

Similar to aider-chat, implement ctrl+d to exit and ensure ctrl+cd clears current CLI chat.

**Additional context**

DOsinga:
fixes #1479  - somewhat

### DIFF
--- a/crates/goose-cli/src/session/input.rs
+++ b/crates/goose-cli/src/session/input.rs
@@ -35,6 +35,25 @@ pub struct PlanCommandOptions {
     pub message_text: String,
 }
 
+struct CtrlCHandler;
+
+impl rustyline::ConditionalEventHandler for CtrlCHandler {
+    /// Handle Ctrl+C to clear the line if text is entered, otherwise exit the session.
+    fn handle(
+        &self,
+        event: &rustyline::Event,
+        _n: usize,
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/session/input.rs
```diff
@@ -270,7 +295,7 @@ fn print_help() {
 /clear - Clears the current chat history
 
 Navigation:
-Ctrl+C - Interrupt goose (resets the interaction to before the interrupted user request)
```
michaelneale, line 273: does it still interrupt when running with this change?

DOsinga: well, like it says, it only interrupt it if the text buffer is empty. the sequence should be:
* if goose is running, ctl-c stops it
* if you have something in the buffer, ctrl-c clears it
* if nothing, exit goose

michaelneale (approved):

PR #3767, opened 2025-08-01
fix: optimise reading large file content

michaelneale:
by default the text_editor_view tool will read any file (under 400KB) but this can be harmful even at smaller size in filling up the context at a high clip. 

With this change it will encourage the agent to search the file and read in ranges IFF it is > 2000 lines long (below that works as of today). It works by letting the agent know on first call that it is a large file (if it gets no range) and that it should either search for content, or call back specifying the full range explicitly to load if it still wants to.

I picked "2000 lines" as a reasonable size (eg for a document or source code) on a human level that is large to cut off at. 

This means that you can run a query like this: 

image

which is a large file, and would normally significantly use up the context, but now is only sipping it: 

image

Before this change it would be at least this usage: 

image

and much worse for larger files (yet can now do the same thing)

cc @cgwalters for idea


cgwalters:
You said the word "optimize" and you tagged me to review and this made me look at the surrounding code... so I did https://github.com/block/goose/pull/3781 which is orthogonal to this but related.

(The next optimization is to build up the `Vec<&str>` as we're reading the file and then we can apply line-based limits *as we read* instead of iterating again, and that's also where we'd want to do constraints on line length e.g. if desired)

Also all this clearly needs deduplication with the shell mcp

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -1011,6 +1011,15 @@ impl DeveloperRouter {
             let lines: Vec<&str> = content.lines().collect();
             let total_lines = lines.len();
 
+            if view_range.is_none() && total_lines > 2000 {
```
cgwalters: Nit, I think documented `const LINE_LIMIT: usize = 2000` at the top would be nice

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -3233,6 +3242,226 @@ mod tests {
...
+        let router = get_router().await;
+
+        let temp_dir = tempfile::tempdir().unwrap();
+        let file_path = temp_dir.path().join("large_file.txt");
+        let file_path_str = file_path.to_str().unwrap();
+        std::env::set_current_dir(&temp_dir).unwrap();
+
+        // Create a file with more than 2000 lines
+        let mut content = String::new();
+        for i in 1..=2001 {
```
cgwalters: (Then this could be `..=LINE_LIMIT+1`)

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -3233,6 +3242,226 @@ mod tests {
...
+            .await;
+
+        assert!(result.is_ok());
+
+        temp_dir.close().unwrap();
+    }
+
+    #[tokio::test]
+    #[serial]
+    async fn test_text_editor_view_file_with_exactly_2000_lines() {
```
cgwalters, line 3220: I'm guessing this is LLM generated, or maybe just copy paste. Personally a rule I have is "two copies is often OK but three wants code sharing" and in situations like this (if it's LLM generated) ask it to try refactoring into helper functions.

michaelneale: yeah although in test code it sometimes is idiomatic to have the names be sort of "BDD" like (that one is a little long though!)

michaelneale: (ie in that case - I find that name explicitly helpful as I know exactly what it will test - myself I would have fewer test cases however so yeah)

michaelneale:
ugh new clippy rules means this will have a lot of unrelated changes (but not functional changes)

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -965,130 +966,152 @@ impl DeveloperRouter {
...
+            return Err(ToolError::ExecutionError(format!(
+                "The path '{}' does not exist or is not a file.",
+                path.display()
+            )));
+        }
 
-            let f = File::open(path)
-                .map_err(|e| ToolError::ExecutionError(format!("Failed to open file: {}", e)))?;
+        // Check file size first (400KB limit)
+        const MAX_FILE_SIZE: u64 = 400 * 1024; // 400KB in bytes
```
DOsinga: no need to say 400k three time

michaelneale:
thanks @cgwalters @katzdave for feedback - it is a slightly larger appearing change as it simplifies the text_editor_view top level function (hard to see that in diffs, but if you see the code it looks simpler/neater)

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -965,130 +966,152 @@ impl DeveloperRouter {
...
+
     async fn text_editor_view(
         &self,
         path: &PathBuf,
         view_range: Option<(usize, i64)>,
     ) -> Result<Vec<Content>, ToolError> {
-        if path.is_file() {
-            // Check file size first (400KB limit)
-            const MAX_FILE_SIZE: u64 = 400 * 1024; // 400KB in bytes
+        if !path.is_file() {
```
DOsinga, line 916: are we sure this is only called on text files? calling format file on a binary might not be great

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -965,130 +966,152 @@ impl DeveloperRouter {
...
-                    {content}
-                    ```
-                    ",
-                    path=path.display(),
-                    language=language,
-                    content=display_content,
-                }
-            };
+        let lines: Vec<&str> = content.lines().collect();
+        let total_lines = lines.len();
```
DOsinga, line 953: could be paranoid here and check the max line length of lines

DOsinga (approved):

crates/goose-mcp/src/developer/mod.rs
michaelneale: yeah I wasn't sure how to break that down - as it is a problem (ie if you tell goose to look at your session jsonl files - will get heavy). 

The issue then is - the view editor tool doesn't have a range that looks inside a line (so it would need to be enhanced) so either part of this change or a follow on to have it take a column range as well as row start and finish?

crates/goose-mcp/src/developer/mod.rs
michaelneale: looking at the old code: https://github.com/block/goose/pull/3767/files#diff-3584a0c10cd66c261203b560f9b4102dec670d257e0a8a0a60702db50a9aab1dL1010 - and scrolling down - looks like it was calling format if I read correctly? so it always *had* to be a text file (ie a thing with lines, that can be format! macrod?) - am I right?

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -965,130 +966,152 @@ impl DeveloperRouter {
...
-                    .collect();
+        let uri = Url::from_file_path(path)
+            .map_err(|_| ToolError::ExecutionError("Invalid file path".into()))?
+            .to_string();
 
-                selected_lines.join("\n")
-            };
+        let mut content = String::new();
+        f.read_to_string(&mut content)
+            .map_err(|e| ToolError::ExecutionError(format!("Failed to read file: {}", e)))?;
```
michaelneale, line 950: @DOsinga this would return if it is a binary

crates/goose-mcp/src/developer/mod.rs
michaelneale: yeah left over from old slop code it looks like

crates/goose-mcp/src/developer/mod.rs
michaelneale: will have a follow on task to make it read by column too when needed

PR #3771, opened 2025-08-01
Prompt injection detection (more abstracted version) [Draft / WIP]

dorien-koelemeijer:
TBD

### DIFF
--- a/crates/goose-security/src/model_downloader.rs
+++ b/crates/goose-security/src/model_downloader.rs
@@ -1,0 +1,847 @@
+#[cfg(feature = "onnx")]
+use anyhow::{anyhow, Result};
+#[cfg(feature = "onnx")]
+use std::path::PathBuf;
+#[cfg(feature = "onnx")]
+use std::process::Command;
+#[cfg(feature = "onnx")]
+use tokio::fs;
+#[cfg(feature = "onnx")]
+use tokio::sync::OnceCell;
+
+#[cfg(feature = "onnx")]
+pub struct ModelDownloader {
+    cache_dir: PathBuf,
+}
+
+#[cfg(feature = "onnx")]
+impl ModelDownloader {
+    pub fn new() -> Result<Self> {
+        // Use platform-appropriate cache directory
+        let cache_dir = if let Some(cache_dir) = dirs::cache_dir() {
+            cache_dir.join("goose").join("security_models")
+        } else {
+            // Fallback to home directory
+            dirs::home_dir()
+                .ok_or_else(|| anyhow!("Could not determine home directory"))?
+                .join(".cache")
+                .join("goose")
+                .join("security_models")
+        };
+
+        Ok(Self { cache_dir })
+    }
+
+    pub async fn ensure_model_available(&self, model_info: &ModelInfo) -> Result<(PathBuf, PathBuf)> {
+        let model_path = self.cache_dir.join(&model_info.onnx_filename);
+        let tokenizer_path = self.cache_dir.join(&model_info.tokenizer_filename);
+        let metadata_path = self.cache_dir.join(format!("{}_metadata.json", 
+            model_info.hf_model_name.replace("/", "_")));
+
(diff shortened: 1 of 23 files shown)


crates/goose-security/src/model_downloader.rs
```diff
@@ -0,0 +1,847 @@
...
+            })
+        } else {
+            // Fallback to the architecture from ModelInfo
+            Ok(model_info.architecture.clone())
+        }
+    }
+
+    async fn download_and_convert_model(&self, model_info: &ModelInfo) -> Result<()> {
+        // Set up Python virtual environment with required dependencies
+        let venv_dir = self.cache_dir.join("python_venv");
```
cgwalters, line 381: I would find it pretty surprising personally if goose started creating python venvs automatically.

crates/goose-security/src/model_downloader.rs
```diff
@@ -0,0 +1,847 @@
...
+                return Err(anyhow!("Failed to install {}: {}", package, stderr));
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn create_conversion_script(&self, model_info: &ModelInfo) -> Result<PathBuf> {
+        let script_content = format!(
+            r#"#!/usr/bin/env python3
```
cgwalters, line 480: Related to above, I don't think Python-in-Rust is really sustainable, shouldn't this just be a separate tool one can install?

cgwalters:
There's a whole lot of layers of abstraction here and it'd be really helpful to have at least one concrete practical example of what this is doing. Also did you just make this all up or are there other projects that are doing this filtering with onnx? (Or did you prompt an AI for this?)

Related to prompt injection, I always go back to https://simonwillison.net/2025/Jun/16/the-lethal-trifecta/ (and https://simonwillison.net/2025/Jun/13/prompt-injection-design-patterns/ is another great link)

How is what you're doing here comparable to any of that?

crates/goose-security/src/model_downloader.rs
dorien-koelemeijer: This is just to download and convert python based models to onnx, it will only happen ever so often (since the models are cached). If there's a better way to do this, I'm happy to hear about it!

dorien-koelemeijer:
> My two cents on how we should implement things like this (as dicussed)
> 
> * We should make this trigger only when a tool call is about to be executed
> * We should create a generic way that allows pluggable things to stop tools from running
> 
> As luck would have it, we already have some plumbing in place, ToolMonitor. If we could turn this into something more abstract, have the check get the whole conversation and then implement this for call counting and for this security thing, we wouldn't have to touch all these different files.
> 
> Let me know what you think or if you want to discuss in a call. We can probably help out a bit with getting the infrastructure in place too

Thanks for the feedback! Wrt to the first bullet point, there seems to be some different opinions around whether this is the way to go. Let me circle back to the team and hear their thoughts as well. It's good to know that there's something existing that can be re-used.

DOsinga:
>Wrt to the first bullet point, there seems to be some different opinions around whether this is the way to go.

Yeah? Can you say more? I have a hard time imagining a scenario where evilness happens without a tool call

dorien-koelemeijer:
> My two cents on how we should implement things like this (as dicussed)
> 
> * We should make this trigger only when a tool call is about to be executed
> * We should create a generic way that allows pluggable things to stop tools from running
> 
> As luck would have it, we already have some plumbing in place, ToolMonitor. If we could turn this into something more abstract, have the check get the whole conversation and then implement this for call counting and for this security thing, we wouldn't have to touch all these different files.
> 
> Let me know what you think or if you want to discuss in a call. We can probably help out a bit with getting the infrastructure in place too

I think the intention was to catch things before they even reach the agent. I'm happy to give this a try though, working on this now but on a different branch, in case we need to get back to this

dorien-koelemeijer:
closing this PR - taking a different approach in line with Douwe's comments

PR #2821, opened 2025-06-08
chore(mcp): Add helpers to parse parameters

cgwalters:
- Avoids redundant code
- Doesn't ignore parameters on type mismatches

Motivated specifically by me having a typo in my optional parameter name when developing.

Note: I only changed a random selection of goose-mcp, but if we choose to do this it'd likely be trivial to ask goose to finish it. I didn't do so yet to not make this a conflict-fest.

PR #3775, opened 2025-08-01
Suppress ansi with pipes

### LINKED ISSUE #3736: Print only structured response at goose run --recipe
Does this feature solve a particular problem you have been experiencing? What opportunities or use cases would be unlocked with this feature?

I often embed goose as a tool in my shell scripts like:

```sh
NO_COLOR=1 goose run --recipe find_all_tasks | tail -n1 | jq '.tasks[]' | ...
```

The problems are

 - You must set `NO_COLOR` env otherwise the output contains color control chars.
 - To retrieve JSON and pipe to `jq`, you must do `tail -n1`
   - I believed `--quiet` is the solution, but it is not. https://github.com/block/goose/pull/2939

One of:

 - Add a option `--output-format=json`
 - Add a option `--machine-readable`
 - Introduce breaking change into `--quite` option as I thought.

**Additional context**

DOsinga:
If you capture the output of goose, you don't need the ansi stuff

fixes #3736

### DIFF
--- a/crates/goose-cli/src/recipes/print_recipe.rs
+++ b/crates/goose-cli/src/recipes/print_recipe.rs
@@ -1,27 +1,27 @@
 use std::collections::HashMap;
 
-use crate::g_println;
+use anstream::println;
 use console::style;
 use goose::recipe::{Recipe, BUILT_IN_RECIPE_DIR_PARAM};
 
 pub fn print_recipe_explanation(recipe: &Recipe) {
-    g_println!(
+    println!(
         "{} {}",
         style("🔍 Loading recipe:").bold().green(),
         style(&recipe.title).green()
     );
-    g_println!("{}", style("📄 Description:").bold());
-    g_println!("   {}", recipe.description);
+    println!("{}", style("📄 Description:").bold());
+    println!("   {}", recipe.description);
     if let Some(params) = &recipe.parameters {
         if !params.is_empty() {
-            g_println!("{}", style("⚙️  Recipe Parameters:").bold());
+            println!("{}", style("⚙️  Recipe Parameters:").bold());
             for param in params {
                 let default_display = match &param.default {
                     Some(val) => format!(" (default: {})", val),
                     None => String::new(),
                 };
 
-                g_println!(
+                println!(
                     "   - {} ({}, {}){}: {}",
                     style(&param.key).cyan(),
                     param.input_type,
@@ -41,7 +41,7 @@ pub fn print_parameters_with_values(params: HashMap<String, String>) {
         } else {
             ""
         };
-        g_println!("   {}{}: {}", key, label, value);
+        println!("   {}{}: {}", key, label, value);
     }
 }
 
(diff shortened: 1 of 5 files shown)


crates/goose-cli/src/recipes/print_recipe.rs
```diff
@@ -1,26 +1,27 @@
 use std::collections::HashMap;
 
+use crate::g_println;
```
cgwalters: There's https://docs.rs/anstream/latest/anstream/ (and already in the depchain as it's used by clap)

cgwalters: And when using it it's way easier to just shadow the builtin `println!` via `use anstream::println;` as the docs suggest.

(With an approach like this inevitably some will be missed)

cgwalters:
BTW reminderping, I reviewed this PR, will trade for a re-review of https://github.com/block/goose/pull/2821

DOsinga:
thanks, yeah good point - i had not seen that. let me get on that.

DOsinga:
cleaned up @cgwalters

crates/goose-cli/src/recipes/print_recipe.rs
```diff
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 
+use anstream::println;
```
cgwalters, line 3: Don't we also need to import the `eprintln` wrapper here?

crates/goose-cli/src/session/output.rs
```diff
@@ -465,14 +477,18 @@ pub fn env_no_color() -> bool {
...
-    bat::PrettyPrinter::new()
-        .input(bat::Input::from_bytes(content.as_bytes()))
-        .theme(theme.as_str())
-        .colored_output(env_no_color())
-        .language("Markdown")
-        .wrapping_mode(WrappingMode::NoWrapping(true))
-        .print()
-        .unwrap();
+    if std::io::stdout().is_terminal() {
+        bat::PrettyPrinter::new()
```
cgwalters, line 481: I am less familiar with `bat` and this `cliclack` crate but IME with https://docs.rs/indicatif/latest/indicatif/ it's pretty easy to use e.g. https://docs.rs/indicatif/latest/indicatif/struct.ProgressBar.html#method.println to automate these kinds of conditional prints.

crates/goose-cli/src/recipes/print_recipe.rs
DOsinga: it's a good question - this PR is fixing automations based on recipe. stderr is not useful at this point for much when it comes to recipes

crates/goose-cli/src/session/output.rs
DOsinga: looks cool. here we are just skipping the formatting of markdown though

crates/goose-cli/src/recipes/print_recipe.rs
cgwalters: I don't understand the surrounding context but my understanding of the bug we're aiming to fix is "don't write ANSI to non-ttys" which is definitely a good idea globally.

This PR is fixing that for `println!` but is also changing some things to use `eprintln!` (why?) but crucially *also* is not using the anstyle wrapper so we'll still output ANSI styles to stderr which seems wrong to me.

IOW we should consistently `use anstream::{println, eprintln};` at least right?

DOsinga: I mostly agree, just not sure I want this to be held up because of that. eprint vs print in the code is kind of a mess right now

lifeizhou-ap (approved):

PR #3782, opened 2025-08-01
Detect client disconnects and cancel tool calls

jamadeo:
If you hit "stop", this will tell MCP tools to stop.

Notably this doesn't work with our built-in developer tool because that ignores the "notifications/cancel" notification. But if we switch that to be an rmcp-implemented server we'll get.


crates/goose-server/src/routes/reply.rs
```diff
@@ -187,82 +194,63 @@ async fn reply_handler(
...
             }
         };
         let saved_message_count = all_messages.len();
 
+        let mut heartbeat_interval = tokio::time::interval(Duration::from_millis(500));
         loop {
             tokio::select! {
-                            _ = task_cancel.cancelled() => {
-                                tracing::info!("Agent task cancelled");
+                _ = task_cancel.cancelled() => {
```
jamadeo, line 215: this is mostly a whitespace change, so I'd suggest turning off whitespace changes for this diff view

annoyingly, rustfmt doens't format code in macros, so this being inside a tokio::select!, didn't get formatted before, and had a bunch of weird indentation

crates/mcp-client/src/client.rs
```diff
@@ -143,10 +167,35 @@ impl McpClient {
...
+            .send_request_with_option(
+                request,
+                PeerRequestOptions {
+                    timeout: Some(self.timeout),
+                    meta: None,
+                },
+            )
+            .await?;
+
+        eprintln!("request handle: {:?}", handle);
```
jamadeo: need to remove a couple of log statements I left in

PR #3788, opened 2025-08-01
Token counting in Auto-compact uses provider metadata

katzdave:
Still a slight discrepancy in the auto-compact completed message (a few hundred tokens) from estimating the system prompt, but the triggering logic is now from the metadata + aligned to the context window indicator.


crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -42,18 +45,22 @@ pub struct CompactionCheckResult {
...
 /// # Returns
 /// * `CompactionCheckResult` containing detailed information about compaction needs
 pub async fn check_compaction_needed(
     agent: &Agent,
     messages: &[Message],
     threshold_override: Option<f64>,
+    session_metadata: Option<&crate::session::storage::SessionMetadata>,
 ) -> Result<CompactionCheckResult> {
     // Get threshold from config or use override
     let config = Config::global();
```
DOsinga, line 66: let's kill the environment variables - this function should just have the threshold passed in. some caller, somewhere may use the environment variable for now. but not this deep into things

crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -63,16 +70,22 @@ pub async fn check_compaction_needed(
...
-        .await
-        .map_err(|e| anyhow::anyhow!("Failed to create token counter: {}", e))?;
-
-    // Calculate current token usage
-    let token_counts = get_messages_token_counts_async(&token_counter, messages);
-    let current_tokens: usize = token_counts.iter().sum();
-    let context_limit = estimate_target_context_limit(provider);
+    let context_limit = provider.get_model_config().context_limit();
+
+    // Try to use actual token counts from session metadata first
```
DOsinga: remove comment

crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -63,16 +70,22 @@ pub async fn check_compaction_needed(
...
+    let (current_tokens, token_source) = match session_metadata.and_then(|m| m.total_tokens) {
+        Some(tokens) => (tokens as usize, "session metadata"),
+        None => {
+            // Fall back to estimated counts
```
DOsinga: remove

crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -63,16 +70,22 @@ pub async fn check_compaction_needed(
...
+            let token_counter = create_async_token_counter()
+                .await
+                .map_err(|e| anyhow::anyhow!("Failed to create token counter: {}", e))?;
+            let token_counts = get_messages_token_counts_async(&token_counter, messages);
+            (token_counts.iter().sum(), "estimated")
+        }
+    };
 
     // Calculate usage ratio
     let usage_ratio = current_tokens as f64 / context_limit as f64;
```
DOsinga, line 88: should we guard against division by zero here?

crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -221,8 +238,8 @@ pub async fn check_and_compact_messages(
     Ok(AutoCompactResult {
         compacted: true,
         messages: compacted_messages,
-        tokens_before: Some(tokens_before),
-        tokens_after: Some(tokens_after),
+        tokens_before: Some(tokens_before + SYSTEM_PROMPT_TOKEN_OVERHEAD + TOOLS_TOKEN_OVERHEAD),
+        tokens_after: Some(tokens_after + SYSTEM_PROMPT_TOKEN_OVERHEAD + TOOLS_TOKEN_OVERHEAD),
     })
```
DOsinga, line 240: This seems wrong; we know how many tokens we had before (sometimes estimated, but for most models we should get a result from them directly). So we also know what the overhead is; the actual tokens before - tokens_before as calculated here. then we can add that to tokens_after for the real value

crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -530,4 +558,137 @@ mod tests {
             .set_param("GOOSE_AUTO_COMPACT_THRESHOLD", serde_json::Value::from(0.3))
```
DOsinga, line 575: this will create side effects when it fails

crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -300,7 +317,7 @@ mod tests {
         // Create small messages that won't trigger compaction
         let messages = vec![create_test_message("Hello"), create_test_message("World")];
 
-        let result = check_compaction_needed(&agent, &messages, Some(0.3))
+        let result = check_compaction_needed(&agent, &messages, Some(0.3), None)
```
DOsinga, line 337: do we ever test the meta_data in these calls? I'd maybe kill a few of these tests and have that does

crates/goose/src/context_mgmt/auto_compact.rs
katzdave: Where do we draw the line what is/isn't allowed to read the global config? I do agree with you that it would be more clear if this parameter got passed in but if there are too many params it leads to a lot of clutter.

crates/goose/src/context_mgmt/auto_compact.rs
katzdave: done

crates/goose/src/context_mgmt/auto_compact.rs
katzdave: done

crates/goose/src/context_mgmt/auto_compact.rs
katzdave: provider.get_model_config().context_limit() has a default value of 128k so can't get 0 div here.

crates/goose/src/context_mgmt/auto_compact.rs
katzdave: This is because the summary prompt doesn't include the system prompt on either end.

So this estimate is pretty reasonable. https://github.com/block/goose/pull/3721 adds the return values from the summarizer so will switch to those instead on there.

crates/goose/src/context_mgmt/auto_compact.rs
katzdave: It gets unset at the end, but yes definitely can see why these can get leaky.

crates/goose/src/context_mgmt/auto_compact.rs
katzdave: done

PR #3789, opened 2025-08-01
Build app bundles on release branches always

jamadeo:


jamadeo:
> ### macOS ARM64 Desktop App (Apple Silicon)
> 📱 Download macOS Desktop App (arm64, signed)
> 
> **Instructions:** After downloading, unzip the file and drag the Goose.app to your Applications folder. The app is signed and notarized for macOS.

not sure why this didn't use the right url as I attempted to use the output of the upload-artifact step, but I've changed it back to what the other workflow did.

PR #3791, opened 2025-08-01
Fix directory switcher not working in active chat sessions and file browser not defaulting to current session directory path

zanesq:
Fix directory switcher not working in active chat sessions and file browser not defaulting to current session directory path

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -962,10 +962,28 @@ const openDirectoryDialog = async (
             defaultPath = currentWorkingDir;
           }
         } catch (error) {
-          console.warn(
-            `Current working directory doesn't exist: ${currentWorkingDir}, falling back to home directory`
-          );
-          defaultPath = os.homedir();
+          if (error && typeof error === 'object' && 'code' in error) {
+            const fsError = error as { code?: string; message?: string };
+            if (
+              fsError.code === 'ENOENT' ||
+              fsError.code === 'EACCES' ||
+              fsError.code === 'EPERM'
+            ) {
+              console.warn(
+                `Current working directory not accessible (${fsError.code}): ${currentWorkingDir}, falling back to home directory`
+              );
+              defaultPath = os.homedir();
+            } else {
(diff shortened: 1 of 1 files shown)


ui/desktop/src/main.ts
```diff
@@ -944,8 +944,42 @@ const buildRecentFilesMenu = () => {
...
 ): Promise<OpenDialogReturnValue> => {
+  // Get the current working directory from the focused window
+  let defaultPath: string | undefined;
+  const currentWindow = BrowserWindow.getFocusedWindow();
+
+  if (currentWindow) {
+    try {
+      const currentWorkingDir = await currentWindow.webContents.executeJavaScript(
+        `window.appConfig ? window.appConfig.get('GOOSE_WORKING_DIR') : null`
+      );
```
DOsinga, line 955: hmm, what's happening here? where does this working dir come from? is that what the server thinks it is?

ui/desktop/src/main.ts
```diff
@@ -944,8 +944,42 @@ const buildRecentFilesMenu = () => {
...
+
+      if (currentWorkingDir && typeof currentWorkingDir === 'string') {
+        // Verify the directory exists before using it as default
+        try {
+          const stats = fsSync.lstatSync(currentWorkingDir);
+          if (stats.isDirectory()) {
+            defaultPath = currentWorkingDir;
+          }
+        } catch (error) {
+          console.warn(
```
DOsinga: seems like an overly broad catch here

DOsinga (commented):
I don't know we might want to do this through the server

ui/desktop/src/main.ts
zanesq: yeah its set from the server when the session starts so it should match the server but this is just for the open file dialog so it really doesn't matter if it matches the server or not its just the default path that opens when the os directory browser shows

DOsinga (approved):
Let's merge this, but I do think we should start doing this sort of thing in rust

PR #3658, opened 2025-07-25
Update OSX codesigning and notarization

finn-block:

PR #3795, opened 2025-08-01
Turn on signing for canary and pr-comment-requested builds

jamadeo:


jamadeo:
@finn-block I see signed builds for these workflows were turned off in #3658 - was that deliberate?

alexhancock (approved):

finn-block:
Yes, that was deliberate. Signing keys have changed to Block-managed keys/credentials and they are more locked down than the old ones. They are not available to pipelines that aren't triggered by a tag.

jamadeo:
Thank you for the clarification @finn-block ! We'll work out how to adjust our process accordingly

PR #3807, opened 2025-08-02
Cmd click open finder

### LINKED ISSUE #3158: Open finder for the current working directory
Although we can select a working directory or a workspace from goose UI, there is generally a need to inspect work done and open it in finder separately for exploration.

Command clicking the workspace directory icon on goose, should open the selected folder in finder.

DOsinga:
This is a dumb hidden feature that some people want and Goose could do while I made a coffee, but it fixes #3158

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -2161,30 +2161,9 @@ app.whenReady().then(async () => {
     event.returnValue = app.getVersion();
   });
 
-  // Handle opening directory in system file explorer
-  ipcMain.handle('open-directory-in-explorer', async (_event, directoryPath: string) => {
+  ipcMain.handle('open-directory-in-explorer', async (_event, path: string) => {
     try {
-      // Validate and expand the directory path
-      const expandedPath = expandTilde(directoryPath);
-      
-      // Check if the directory exists
-      const stats = await fs.stat(expandedPath);
-      if (!stats.isDirectory()) {
-        console.error('Path is not a directory:', expandedPath);
-        return false;
-      }
-
-      // Use shell.openPath to open the directory in the system file explorer
-      const result = await shell.openPath(expandedPath);
-      
-      // shell.openPath returns an empty string on success, or an error message on failure
-      if (result === '') {
-        console.log('Successfully opened directory in explorer:', expandedPath);
-        return true;
-      } else {
-        console.error('Failed to open directory in explorer:', result);
-        return false;
-      }
+      return !!(await shell.openPath(path));
     } catch (error) {
       console.error('Error opening directory in explorer:', error);
       return false;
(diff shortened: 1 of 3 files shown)


ui/desktop/src/main.ts
```diff
@@ -2160,6 +2160,15 @@ app.whenReady().then(async () => {
   ipcMain.on('get-app-version', (event) => {
     event.returnValue = app.getVersion();
   });
+
+  ipcMain.handle('open-directory-in-explorer', async (_event, path: string) => {
```
zanesq, line 2216: Since this is exposed as an IPC the path can be compromised to show any directory on the machine. Maybe we should add an allowlist (or blocklist) so only appropriate files are opened? This and anywhere else that we open directories or list files through an IPC.

michaelneale: it is possible in electron to have ipc only available to the renderer so if so, that isn't an issue (checking)

michaelneale: yep seems ok: Electron's `ipcMain` and `ipcRenderer` communicate through Chromium's internal IPC system

```
## Security Analysis of Electron IPC Setup

### ✅ **Secure Configuration Found:**

1. **Context Isolation is Enabled** (`contextIsolation: true`)
   - This is the most critical security feature that prevents the renderer process from accessing Node.js APIs or Electron internals directly
   - The renderer can only access what's explicitly exposed through the context bridge

2. **Node Integration is Disabled** (`nodeIntegration: false`)
   - Prevents the renderer from having direct access to Node.js APIs
   - This is a crucial security measure to prevent arbitrary code execution

3. **Web Security is Enabled** (`webSecurity: true`)
   - Enforces same-origin policy and other web security features
   - Prevents loading of insecure content

4. **Preload Script with Context Bridge**
   - All IPC communication is channeled through a preload script (`preload.ts`)
   - Uses `contextBridge.exposeInMainWorld()` to expose only specific, controlled APIs
   - The exposed APIs are well-defined with type safety

5. **Controlled API Surface**
   - Only specific methods are exposed to the renderer through two APIs:
     - `window.electron`: Contains all IPC methods for file operations, settings, etc.
     - `window.appConfig`: Contains configuration data
   - Each exposed method is explicitly defined and typed
```

michaelneale (approved):
seems reasonable (and IPC thing doesn't look like it is an issue - good to check of course).

ui/desktop/src/main.ts
zanesq: Good call, I'll add some tests to ensure this never changes down the road also 👍

michaelneale: @zanesq oh yeah that would be stellar - as I am sure would be an easy mistake to make

DOsinga: > Since this is exposed as an IPC the path can be compromised to show any directory on the machine

can you help me understand the line of attack here? even if an outside attacker would break into goose (at which point they have full control over the host machine anyway), if they would this trigger this end point, all it would do is start the finder in a particular directory. only the end user would be able to see that, no? not the attacker?

I'd love to understand security considerations with the electron app better as aI see some checks that I don't quite understand.

PR #3810, opened 2025-08-03
docs: Dynamic linux install buttons

blackgirlbytes:
Linux install buttons were hard coded to 1.0.29. Community members were asking about this. 

This PR replaces hardcoded version URLs by fetching the latest using GitHub API. 
If the API call fails, it brings users to the release page with the latest binaries.

### DIFF
--- a/documentation/src/components/LinuxDesktopInstallButtons.js
+++ b/documentation/src/components/LinuxDesktopInstallButtons.js
@@ -1,20 +1,72 @@
 import Link from "@docusaurus/Link";
 import { IconDownload } from "@site/src/components/icons/download";
+import { useState, useEffect } from "react";
 
 const LinuxDesktopInstallButtons = () => {
+  const [downloadUrls, setDownloadUrls] = useState({
+    deb: "https://github.com/block/goose/releases/download/v1.1.4/goose_1.1.4_amd64.deb",
+    rpm: "https://github.com/block/goose/releases/download/v1.1.4/Goose-1.1.4-1.x86_64.rpm"
+  });
+
+  useEffect(() => {
+    const fetchLatestRelease = async () => {
+      try {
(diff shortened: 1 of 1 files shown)


documentation/src/components/LinuxDesktopInstallButtons.js
```diff
@@ -1,20 +1,72 @@
 import Link from "@docusaurus/Link";
 import { IconDownload } from "@site/src/components/icons/download";
+import { useState, useEffect } from "react";
 
 const LinuxDesktopInstallButtons = () => {
+  const [downloadUrls, setDownloadUrls] = useState({
+    deb: "https://github.com/block/goose/releases/latest",
+    rpm: "https://github.com/block/goose/releases/latest"
```
DOsinga: since these are the same, consider using a constant

documentation/src/components/LinuxDesktopInstallButtons.js
```diff
@@ -1,20 +1,72 @@
...
+    const fetchLatestRelease = async () => {
+      try {
+        // Check cache first (1 hour expiry)
+        const cached = localStorage.getItem('goose-release-cache');
+        const cacheTime = localStorage.getItem('goose-release-cache-time');
+        const now = Date.now();
+        
+        if (cached && cacheTime && (now - parseInt(cacheTime)) < 3600000) {
+          // Use cached data if less than 1 hour old
+          const cachedData = JSON.parse(cached);
```
DOsinga: you could move the retrieval of cached inside the if since otherwise we don't care

documentation/src/components/LinuxDesktopInstallButtons.js
blackgirlbytes: good point! fixed.

Issue #3555, opened 2025-07-21
Add Clippy baseline to block new complexity and unwrap usage


lifeizhou-ap:
**Context**
Our codebase has been around for a while and has grown in complexity. Right now, for Rust code we run cargo clippy with the default rules, but I think we could benefit from enabling some additional lints that specifically flag complexity and error handling issues.

Here are some that I’m considering:

> clippy::too_many_lines
> clippy::too_many_arguments
> clippy::cognitive_complexity
> clippy::fn_params_excessive_bools
> clippy::type_complexity
> clippy::branches_sharing_code
> clippy::unwrap_used

**Proposal**
If we turn on the above rule check, there will be hundreds of violations. Fixing all of them at once would be a very big task.  I am thinking about using a baseline file to capture current existing Clippy violations (by file + lint).
CI would then block new violations, but not existing ones.

With this, we can prevents new issues from the new files/changes and then we can fix old issues gradually.

taniacryptid:
Thank you for sharing! Let me tag cc @michaelneale @blackgirlbytes for visibility!

michaelneale:
yeah would be interesting to try at least

DOsinga:
I think we should do this

DOsinga:
let's just turn them on one by one

DOsinga:
we have this now!

lifeizhou-ap:
Hi @DOsinga ,

I am sorry that I did not implement in last a couple of weeks.  I created a PR just now https://github.com/block/goose/pull/3813 for this issue.

PR #3813, opened 2025-08-03
Chore: apply more clippy rules to prevent from code complexity

lifeizhou-ap:
Address Issue https://github.com/block/goose/issues/3555


scripts/lint.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
DOsinga: can we name this rust-lint or even clippy-lint ? we have more linters

DOsinga (approved):

scripts/clippy-baseline.sh
```diff
@@ -0,0 +1,160 @@
+#!/bin/bash
+
+# Baseline clippy rules - only fail on NEW violations
```
cgwalters, line 3: I think it'd be a lot cleaner and more obvious to just add `#[allow(clippy::...)]` into the functions which currently have violations.

Then there's no need for a whole custom wrapper script.

cgwalters: It's also possible to allowlist lints for entire modules, and then incrementally drop that.

https://github.com/block/goose/pull/3813/commits/795f42f2c956b92a0fbb7ddcfc296c7617ba8853 is a great example of a wholly unnecessary bug in this AI generated bash code when there's already basically a builtin for this in the Rust toolchain.

lifeizhou-ap: @cgwalters Thanks for the feedback!

I actually thought about the `#[allow(clippy::...)]` approach but ended up with the baseline approach. The concerns are:
- quite a few allows scattered throughout the code, and when developers see existing allows, they're more likely to
   add their own instead of fixing the underlying issues.
- it gets hard to track why we're allowing certain things, and if we want to add new clippy rules later we'd have
  to go through and add even more allows.

With the baseline approach we can see all our existing violations (technical debt)  in one place and address them over time.

PR #3822, opened 2025-08-04
File completion in CLI

### LINKED ISSUE #1372: Tab completion for file paths
**Feature Enhancement Request: File Path Autocompletion for Goose Prompts**  

When using Goose prompts that require referencing a file, users currently have to manually type out the full file path. This can be tedious, especially for deeply nested files or long filenames. Adding autocompletion for file paths would streamline the process, reducing the chance of typos and making it quicker to reference files for Goose to read and provide context.  

I propose enabling tab-completion for file paths within Goose prompts, so that users can:  
- Start typing a file path and press **Tab** to autocomplete it.  
- See suggestions for matching files and directories in the current working directory.  
- Seamlessly insert the fully qualified path without needing to manually type it out.  

The goal is to improve efficiency when referencing files for Goose to process, without requiring users to navigate directories interactively—just provide a quick way to complete paths inline.  

- Manually copying and pasting file paths from a file explorer or terminal, which is slower and prone to errors.  
- Using a shell command (`ls` or `find`) to get the file path and then pasting it, which adds extra steps.  
- Implementing custom shell aliases or scripts, but this requires user-specific setup rather than being a built-in feature.  
  

DOsinga:
Fixed #1372

### DIFF
--- a/crates/goose-cli/src/session/completion.rs
+++ b/crates/goose-cli/src/session/completion.rs
@@ -1,8 +1,8 @@
-use rustyline::completion::{Completer, Pair};
+use rustyline::completion::{Completer, FilenameCompleter, Pair};
 use rustyline::highlight::{CmdKind, Highlighter};
 use rustyline::hint::Hinter;
 use rustyline::validate::Validator;
-use rustyline::{Helper, Result};
+use rustyline::{Context, Helper, Result};
 use std::borrow::Cow;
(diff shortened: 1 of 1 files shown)


DOsinga:
>would be cool if we could do it for app too, but likely more complex.

yeah, I agree. for the app the @... is probably a better way to do it though

Issue #3777, opened 2025-08-01
Please add "OpenAI Compatible" and "Anthropic Compatible" provider configuration


complyue:
**Please explain the motivation behind the feature request.**
---
Many LLM api providers provide OpenAI/Anthropic compatible endpoints, with their own BASE_URL and model names, e.g.

https://docs.z.ai/scenario-example/develop-tools/roo
https://docs.bigmodel.cn/cn/guide/develop/roo

https://docs.z.ai/scenario-example/develop-tools/claude
https://docs.bigmodel.cn/cn/guide/develop/claude

Currently goose has fixed provider cfg items, not working with those broader LLM api providers yet.

**Describe the solution you'd like**
---
Please enable the user of goose to create custom providers those either OpenAI or Anthropic compatible, with custom BASE_URL and API_KEYs, and ideally with custom model list for the user to select from.

**Describe alternatives you've considered**
---
N/A

**Additional context**
---
Cline, Roo Code and Kilo Code all support "OpenAI Compatible" providers for long time, and worked really while.

- [x] I have verified this does not duplicate an existing feature request

shali1995:
same cant use this at all, please add this already. this is absolutly anoyying. do it like roo or cline..... oh my god............... cna you guys please get connected to the people andundertadn how we use it really?

taniacryptid:
Thank you for posting this. Would love to get this working in theory, after immediate bug fixes and resolves. The main thing I know off-hand is that the team will want to ensure custom endpoints added here work when implemented. I'm sorry you can't use it right now via OpenRouter (get that it's your personal preference) and leaving this comment to bump the thread!

DavidSint:
Just bumping this as it'd be great if one could use goose with the Vercel v0 family of models which is compatible with the OpenAI Chat Completions API spec.

richardstevenhack:
Obviously. Every new tools seems to start by "we support these providers" - rather than "we support YOUR providers."
It's called "standardization".

Developerayo:
Been looking into this, should push a pr soon with the adding of custom Anthropic/OpenAI providers

complyue:
I see #4099 merged and released as with 1.6.0, but no luck to get it work:

`~/.config/goose/custom_providers/custom_bigm_-_glm.json`

```json
{
  "name": "custom_bigm_-_glm",
  "engine": "anthropic",
  "display_name": "BIGM - GLM",
  "description": "Custom BIGM - GLM provider",
[...]

PR #3824, opened 2025-08-04
feat: support for openai/anthropic compatibe apis

Developerayo:
Fixes: #3777

- Add support for configuring OpenAI/Anthropic compatible APIs (local or cloud)
- Fix delete confirmation for custom & known providers

image

image


DOsinga:
this is awesome, thanks so much for doing this work!

I had been dabbling myself in this and took a slightly different approach. I hope we can combine the two efforts

here's my very much WIP:
https://github.com/block/goose/pull/3825

it rejigs the whole factory stuff and then adds custom providers that can be read from json. they have an engine, i.e. which provider executes it. that's kinda nice because then we don't have to have actually extra providers in there. I went with Ollama and OpenAI because I saw people ask for that, but we can easily add Anthropic of course

Developerayo:
> this is awesome, thanks so much for doing this work!
> 
> I had been dabbling myself in this and took a slightly different approach. I hope we can combine the two efforts
> 
> here's my very much WIP: #3825
> 
> it rejigs the whole factory stuff and then adds custom providers that can be read from json. they have an engine, i.e. which provider executes it. that's kinda nice because then we don't have to have actually extra providers in there. I went with Ollama and OpenAI because I saw people ask for that, but we can easily add Anthropic of course

Ahh!! This is an excellent idea!
Could update this PR and inculcate your approach

DOsinga:
if you want to merge my PR into this and then make it all work, go for it! happy to close mine after that

DOsinga:
Ok, I'm going to give this a go to see if I can merge the two approaches. hold on!

DOsinga:
never mind, looks like you already did. checked out your branch and took it for a spin, this works pretty well! the only feedback I have here is that we could add a boolean to allow/disallow streaming (not all "compatible" backends support that) and we could consider moving some of the code that is being done in the config for the server (add custom provider) and the cli to the custom provider class so there's less repetition and we do less work in routes.

can you merge master in so we can get a clean diff? looking forward to getting this in the hands of people!

Developerayo:
> never mind, looks like you already did. checked out your branch and took it for a spin, this works pretty well! the only feedback I have here is that we could add a boolean to allow/disallow streaming (not all "compatible" backends support that) and we could consider moving some of the code that is being done in the config for the server (add custom provider) and the cli to the custom provider class so there's less repetition and we do less work in routes.
> 
> can you merge master in so we can get a clean diff? looking forward to getting this in the hands of people!

All resolved 🫡

crates/goose-cli/src/commands/configure.rs
```diff
@@ -221,6 +224,11 @@ pub async fn handle_configure() -> Result<(), Box<dyn Error>> {
                 "Configure Providers",
                 "Change provider or update credentials",
             )
+            .item(
+                "custom_providers",
+                "Custom Providers",
+                "Add OpenAI or Anthropic compatible APIs",
```
DOsinga: We also do Ollama, so maybe simplify to generic

crates/goose-server/src/routes/config_management.rs
```diff
@@ -306,7 +315,68 @@ pub async fn providers(
 ) -> Result<Json<Vec<ProviderDetails>>, StatusCode> {
     verify_secret_key(&headers, &state)?;
 
-    let providers_metadata = get_providers();
+    let mut providers_metadata = get_providers();
+
+    let config_dir = etcetera::choose_app_strategy(goose::config::base::APP_STRATEGY.clone())
```
DOsinga: So I think here and it in the other handle, it would be better I think to move the work to custom_provider.rs, keep the routes light and have:

`let custom_providers_dir = config_dir.join("custom_providers");`

appear only once type of thing. we can do this in a follow up though

crates/goose/src/providers/anthropic.rs
```diff
@@ -260,6 +287,6 @@ impl Provider for AnthropicProvider {
     }
 
     fn supports_streaming(&self) -> bool {
-        true
+        self.supports_streaming
```
DOsinga, line 290: nice

ui/desktop/src/components/settings/providers/subcomponents/CardContainer.tsx
```diff
@@ -51,10 +53,11 @@ export default function CardContainer({
       {!grayedOut && <GlowingRing />}
       <div
         className={`relative bg-background-default rounded-lg p-3 transition-all duration-200 h-[160px] flex flex-col justify-between
+                   ${borderStyle === 'dashed' ? 'border-2 border-dashed' : 'border'}
                    ${
                      grayedOut
-                       ? 'border border-borderSubtle'
-                       : 'border border-borderSubtle hover:border-borderStandard'
+                       ? 'border-borderSubtle'
```
DOsinga, line 59: intentional?

ui/desktop/src/components/ui/button.tsx
```diff
@@ -4,7 +4,7 @@ import { cva, type VariantProps } from 'class-variance-authority';
 import { cn } from '../../utils';
 
 const buttonVariants = cva(
-  "inline-flex items-center justify-center gap-2 whitespace-nowrap text-sm transition-all disabled:pointer-events-none disabled:opacity-50 [&_svg]:pointer-events-none [&_svg:not([class*='size-'])]:size-4 shrink-0 [&_svg]:shrink-0 outline-none focus-visible:border-ring focus-visible:ring-ring/50 focus-visible:ring-[1px] aria-invalid:ring-destructive/20 dark:aria-invalid:ring-destructive/40 aria-invalid:border-destructive",
+  "inline-flex items-center justify-center gap-2 whitespace-nowrap text-sm transition-all cursor-pointer disabled:pointer-events-none disabled:opacity-50 [&_svg]:pointer-events-none [&_svg:not([class*='size-'])]:size-4 shrink-0 [&_svg]:shrink-0 outline-none focus-visible:border-ring focus-visible:ring-ring/50 focus-visible:ring-[1px] aria-invalid:ring-destructive/20 dark:aria-invalid:ring-destructive/40 aria-invalid:border-destructive",
```
DOsinga, line 7: also?

ui/desktop/src/components/settings/providers/subcomponents/CardContainer.tsx
Developerayo: Yea! Wanted to seperate color and border style

ui/desktop/src/components/ui/button.tsx
Developerayo: Cursor was showing as default rather than a pointer, just a small add to buttons

crates/goose-server/src/routes/config_management.rs
Developerayo: Ah! yes, agreed, will move the work to `custom_providers.rs`, better to do in this

DOsinga (approved):
can you fix the build things?

Developerayo:
> can you fix the build things?

oops, resolved, should be all good now...

DOsinga:
can you have another look, they all seem to have failed. partially I think because there's changes to the package lock, but not to the actually installed packages. I still would love to get this in! also the commits are missing DCO which we sadly now require

DOsinga:
I'll make sure this gets merged. thanks

DOsinga:
thanks for doing this. I just merged this by cloning this and changing things, so I am going to close this, but hurray!

Developerayo:
> thanks for doing this. I just merged this by cloning this and changing things, so I am going to close this, but hurray!

🫡

...#4099

DOsinga:
let me know if you have any other good ideas you want to contribute!

PR #3825, opened 2025-08-04
Custom providers infra

DOsinga:


jamadeo:
Work continues in #3824

PR #3829, opened 2025-08-04
Changed app settings configuration form to match settings panels

zanesq:
- fixed layout to match app settings panels with a button and now launches in a modal to save space
- fixed reordering bug when saving a config setting
- added flag `CONFIGURATION_ENABLED` to show configuration settings

image
image


ui/desktop/src/components/settings/config/ConfigSettings.tsx
```diff
@@ -3,21 +3,41 @@ import { Input } from '../../ui/input';
...
-  }, [typedConfig]);
+
+    // Capture the original key order only on first load or when keys change significantly
+    if (
+      originalKeyOrder.length === 0 ||
+      Object.keys(typedConfig).length !== originalKeyOrder.length
+    ) {
+      setOriginalKeyOrder(Object.keys(typedConfig));
+    }
+  }, [typedConfig, originalKeyOrder.length]);
```
DOsinga: this dependency should probably also include typedConfig.length - typedConfig could have something added but the object wouldn't change

ui/desktop/src/components/settings/config/ConfigSettings.tsx
```diff
@@ -63,84 +83,108 @@ export default function ConfigSettings() {
...
+                        {getUiNames(key)}
+                      </label>
+                      <Input
+                        value={String(configValues[key] || '')}
+                        onChange={(e) => handleChange(key, e.target.value)}
+                        className={cn(
+                          'text-textStandard border-borderSubtle hover:border-borderStandard',
+                          configValues[key] !== typedConfig[key] && 'border-blue-500'
+                        )}
+                        placeholder={`Enter ${getUiNames(key).toLowerCase()}`}
```
DOsinga: I don't think we want the lowercase here

ui/desktop/src/components/settings/config/ConfigSettings.tsx
```diff
@@ -63,84 +83,108 @@ export default function ConfigSettings() {
...
-                </Button>
+            <div className="flex-1 max-h-[60vh] overflow-auto pr-4">
+              <div className="space-y-4">
+                {configEntries.length === 0 ? (
+                  <p className="text-textSubtle">No configuration settings found.</p>
+                ) : (
+                  configEntries.map(([key, _value]) => (
+                    <div key={key} className="grid grid-cols-[200px_1fr_auto] gap-3 items-center">
+                      <label className="text-sm font-medium text-textStandard" title={key}>
+                        {getUiNames(key)}
```
DOsinga, line 178: this is supposed to change

UPPER_CASE_KEY to Upper Case Key, but in your screenshot it does not?

PR #3834, opened 2025-08-04
fix: replace glob/grep tool with shell

baxen:
after some time using the new version, we see more errors and increased usage of find

### DIFF
--- a/crates/goose-mcp/src/developer/mod.rs
+++ b/crates/goose-mcp/src/developer/mod.rs
@@ -152,13 +152,16 @@ impl DeveloperRouter {
                 If you need to run a long lived command, background it - e.g. `uvicorn main:app &` so that
                 this tool does not run indefinitely.
 
+                **Important**: Use ripgrep - `rg` - exclusively when you need to locate a file or a code reference,
+                other solutions may produce too large output because of hidden files! For example *do not* use `find` or `ls -r`
+                  - List files by name: `rg --files | rg <filename>`
+                  - List files that contain a regex: `rg '<regex>' -l`
+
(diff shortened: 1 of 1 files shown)


crates/goose-mcp/src/developer/mod.rs
```diff
@@ -152,13 +152,16 @@ impl DeveloperRouter {
                 If you need to run a long lived command, background it - e.g. `uvicorn main:app &` so that
                 this tool does not run indefinitely.
 
+                **Important**: Use ripgrep - `rg` - exclusively when you need to locate a file or a code reference,
+                other solutions may produce too large output because of hidden files! For example *do not* use `find` or `ls -r`
+                  - List files by name: `rg --files | rg <filename>`
+                  - List files that contain a regex: `rg '<regex>' -l`
+
```
DOsinga, line 159: should we also update the windows instructions. or better, make them as much as possible the same

alexhancock (approved):

crates/goose-mcp/src/developer/mod.rs
baxen: let me know which part of this you think is missing from windows! it actually had closer to this prompt than the one i'm updating here, it may not have been changed in the previous PR that introduced the new tools i'm removing

DOsinga: yeah, maybe to increase maintainability extract the commonalities so we will keep them in sync better going forward?

michaelneale:
cool - I wonder if over time we should move these to glob.rs and so on (even though in this case is just deleted) so they are easier to comment out/keep/chuck away.

PR #3838, opened 2025-08-04
fix oauth

lilydelalande:
example fix for what is missing currently 

1. Detects 401 errors and OAuth-requiring endpoints (like mcp.notion.com)
2. Automatically triggers OAuth flow when connection fails
3. Retries with Bearer token after successful authentication

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -224,7 +224,7 @@ impl ExtensionManager {
                     );
                 }
                 let client = reqwest::Client::builder()
-                    .default_headers(default_headers)
+                    .default_headers(default_headers.clone())
                     .build()
                     .map_err(|_| {
                         ExtensionError::ConfigError("could not construct http client".to_string())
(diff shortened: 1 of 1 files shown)


michaelneale:
oh very cool - this seems much needed

crates/goose/src/agents/extension_manager.rs
```diff
@@ -224,7 +224,7 @@ impl ExtensionManager {
                     );
                 }
                 let client = reqwest::Client::builder()
-                    .default_headers(default_headers)
+                    .default_headers(default_headers.clone())
```
DOsinga, line 227: why not always send the bearer token if can acquire one along here? if we can acquire it, we need it anyway

jamadeo:
Thank you @lily-de for jumping on this! I think we want to take this opportunity to use rmcp's oauth and drop our own implementation (https://github.com/block/goose/pull/3845)

DOsinga:
closing - feel free to reopen when needed

Issue #3841, opened 2025-08-04
Update specific memory


omarzina:
**Describe the bug**

Maybe it's not a bug but:

**To Reproduce**
1. Create a new session
2. Ask goose to write an implementation plan and save it to memory.
3. Open the plan from memory and review it. Find something you need to change in the plan.
4. In the same session, ask goose to update the plan with detailed instructions.
5. Goose *appends* the *whole* plan to the bottom of the memory file with the instructed updates, so you're left with basically the same plan twice (or as many times as updates as you make), the second time with the updates instructed.

Some thoughts:
1. Maybe it is how goose memory is supposed to work and I'm using it the wrong way?
2. As a workaround, should I write the plan in an external document (ie. outside goose memory) and work based on it instead of relying on goose memory?

**Expected behavior**
The plan is updated with the instructions given instead of re-writing the whole plan considering the updates.

**Please provide following information:**
 - **OS & Arch:** macOS 15.5 | M4 Pro arm
 - **Interface:** CLI
 - **Version:** v1.1.4
 - **Extensions enabled:** computercontroller, context7, developer, memory, ref-tools
 - **Provider & Model:** Amazon Bedrock, Claude Sonnet 4

omarzina:
Additional context: if I already have the final plan in memory, I can ask goose to retrieve it, execute the next logical step and update the status, and it will, instead of updating the section implemented, add a reference to that section at the bottom and  mark it as implemented.

michaelneale:
FYI _ https://github.com/block/goose/pull/3848 - is a look at an explicit task tracker (which is specific to a session/more ephemeral than memory which is usually more durable) - and designed for planning.

omarzina:
Looks awesome, looking forward to it, thank you for your work!

PR #3848, opened 2025-08-05
feat: task tracker for session efficiency

michaelneale:
this is a follow on from: https://github.com/block/goose/pull/3670

this adds a task tracker tool as part of the developer extension, which is useful when sessions are heavily summarized, these will remain and help keep things on track (but are ephemeral)

this may help address things like: https://github.com/block/goose/issues/3841 

take a look @katzdave


crates/goose-mcp/src/developer/task_tracker.rs
```diff
@@ -0,0 +1,477 @@
...
+            "mark_done" => {
+                let task = params.get("task").and_then(|v| v.as_str()).ok_or(
+                    ToolError::InvalidParameters("The task string is required".to_string()),
+                )?;
+
+                let result = self.mark_task_done(task.to_string());
+
+                Ok(vec![Content::text(result)])
+            }
+            "clear" => {
```
katzdave: clear_all?

crates/goose-mcp/src/developer/task_tracker.rs
```diff
@@ -0,0 +1,477 @@
...
+        Self {
+            tasks: Arc::new(Mutex::new(HashMap::new())),
+        }
+    }
+
+    pub fn list_tasks(&self) -> Vec<String> {
+        let tasks = self.tasks.lock().unwrap();
+        let mut task_list = Vec::new();
+        for (task, status) in tasks.iter() {
+            task_list.push(format!("{} [{}]", task, status.as_str()));
```
katzdave, line 40: Should we filter completed tasks so this doesn't grow infinitely?

Maybe mark_task_done is a good place to do this? Remove all done tasks first, then mark the current task as done. Will keep a record then of the most recently completed task for some history but won't be too long running.

crates/goose-mcp/src/developer/task_tracker.rs
```diff
@@ -0,0 +1,477 @@
...
+        }
+        task_list
+    }
+
+    pub fn add_task(&self, task: String) -> String {
```
katzdave, line 45: Should this accept a list explicitly? or an add_tasks variant? 

Can imagine after a large non-focused user prompt we want to add ~5 tasks, is there any benefit to trying to minimize tool calls required there?

I don't really like the contains(',') for list checking.

crates/goose-mcp/src/developer/task_tracker.rs
```diff
@@ -0,0 +1,477 @@
...
+            TaskStatus::Todo => "to do",
+            TaskStatus::Wip => "wip",
+            TaskStatus::Done => "done",
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct TaskTracker {
+    tasks: Arc<Mutex<HashMap<String, TaskStatus>>>,
```
katzdave, line 26: Priority and size might be worth considering but seems reasonable to experiment with in a followup.

crates/goose-mcp/src/developer/task_tracker.rs
```diff
@@ -0,0 +1,477 @@
...
+            params
+                .get("action")
+                .and_then(|v| v.as_str())
+                .ok_or(ToolError::InvalidParameters(
+                    "The action string is required".to_string(),
+                ))?;
+
+        match action {
+            "list" => {
+                let tasks = self.list_tasks();
```
katzdave, line 91: Will experiment with how eager goose is to call this, but maybe we want to add some kind of mechanism to force this onto the end of every message (but not keep it in the context as a traditional message, so we're always presenting the latest state).

katzdave (commented):
Cool! Will try it out some today.

crates/goose-mcp/src/developer/task_tracker.rs
michaelneale: yeah - it was either that or I expect the agent to clear tasks, btu yes, that is a good idea to make it clear up front

crates/goose-mcp/src/developer/task_tracker.rs
michaelneale: yeah can change that - made the schema a little more complex for LLMs though - but can add an add_list command to make it simpler for it, good call.

crates/goose-mcp/src/developer/task_tracker.rs
michaelneale: size - could cap it/push back, but priority - you mean like an order?

crates/goose-mcp/src/developer/task_tracker.rs
michaelneale: hrm - that is an interesting to experiment with, I found with opus/sonnet it is reasonably happy to call it, but yes, that could be good. 

One thing I would like to think about @katzdave: should we tweak the prompt so it doesn't ALWAYS use this, ie for non trivial one shot things should it not always call task tracker? I worry it could slow it down eg for some recipe like things, something to think about if you have any intuition there.

crates/goose-mcp/src/developer/task_tracker.rs
michaelneale: I think it may be important for it to keep a list of what it has done over time - until it clears them all, the more I think about it (we can limit it though). The list shoudln't ever be that big or will have other troubles

crates/goose-mcp/src/developer/task_tracker.rs
michaelneale: yes will get rid of that, dont' want that many tasks.

michaelneale: changed it to not take a list

michaelneale:
@katzdave ready for another look - some tweaks/simplifying. For size - could include some cap now if you think needed? 

here it is kicking in after a compaction: 

image


(and in that session, I started with something simple so it didn't kick in)

crates/goose-mcp/src/developer/task_tracker.rs
Kvadratni: this needs to be a platform tool and it should be a system prompt part

michaelneale: as in go into the system prompt? the list should @Kvadratni ?

Kvadratni: the usage instructions

crates/goose-mcp/src/developer/task_tracker.rs
katzdave: yeah like order; claude has high, medium and low.

crates/goose-mcp/src/developer/task_tracker.rs
katzdave: I guess was thinking that having too many completed tasks isn't relevant but yeah I think knowing what you did in the past might be helpful contex.

crates/goose-mcp/src/developer/task_tracker.rs
michaelneale: tool descriptions usually do go in there already

michaelneale:
I think this is better done in this location: https://github.com/block/goose/pull/3902 - but should check out if fine grained is better.

PR #3851, opened 2025-08-05
Tell the user to hit compact

DOsinga:
The compaction is not triggering for me so adding this seems like a reasonable thing to do meanwhile


katzdave (approved):
LGTM + like the word change of summarize -> compact.

Should I follow up and rename /summarize to /compact on the cli for this release? We have some messages about 'auto-compaction' there.

PR #3853, opened 2025-08-05
Drop file system MCP

DOsinga:
this is actually a node mcp, not a python one. but worse, it requires specifying which folders to mount. also we already have internal file system support, so dropping it seems reasonable

### DIFF
--- a/documentation/static/servers.json
+++ b/documentation/static/servers.json
@@ -335,17 +335,6 @@
     ]
   },
   {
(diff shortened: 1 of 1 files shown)


angiejones (changes requested):
delete the guide too
https://github.com/block/goose/blob/main/documentation/docs/mcp/filesystem-mcp.md

DOsinga:
closing since it didn't make the release

PR #3859, opened 2025-08-05
Don't ignore lockfiles on linux/windows builds

jamadeo:
Presumably this will fail somewhere (why else would we ignore?) but we should fix the root issue, not ignore the lock file.


jamadeo:
.bundle-windows

PR #3860, opened 2025-08-05
Pin @mcp-ui/client in package.json

jamadeo:
we should also do #3859 but this should fix the immediate build problem with windows/linux

### DIFF
--- a/ui/desktop/package.json
+++ b/ui/desktop/package.json
@@ -42,7 +42,7 @@
     "@ai-sdk/openai": "^0.0.72",
     "@ai-sdk/ui-utils": "^1.0.2",
     "@hey-api/client-fetch": "^0.8.1",
-    "@mcp-ui/client": "^5.3.1",
+    "@mcp-ui/client": "~5.3.1",
     "@radix-ui/react-accordion": "^1.2.2",
     "@radix-ui/react-avatar": "^1.1.1",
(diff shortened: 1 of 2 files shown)


ui/desktop/package.json
```diff
@@ -42,7 +42,7 @@
     "@ai-sdk/openai": "^0.0.72",
     "@ai-sdk/ui-utils": "^1.0.2",
     "@hey-api/client-fetch": "^0.8.1",
-    "@mcp-ui/client": "^5.3.1",
+    "@mcp-ui/client": "~5.3.1",
```
alexhancock, line 45: could use https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides as well if we want to be very specific

alexhancock (approved):

ui/desktop/package.json
aharvard, line 45: Should be good. I think we could also hard-code to 5.3.1, but allowing patches seems reasonable.

aharvard (approved):
LGTM 

I'm thinking that when https://github.com/idosal/mcp-ui/issues/61 is resolved we can upgrade.

ui/desktop/package.json
aharvard: im also cool with hardcoding

PR #3867, opened 2025-08-05
fix: extension install from deeplink not refreshing

zanesq:
Changed extension list loading to reactive updates from config context rather than manually calling refresh extensions with timing issues


zanesq:
Putting into draft while I investigate some issues will reopen

PR #3871, opened 2025-08-05
additional metrics + Ui implementation

Kvadratni:
This PR further improves on goose metrics by adding mew types and expanding the coverage to the APP


crates/goose-cli/src/main.rs
```diff
@@ -3,5 +3,18 @@ use goose_cli::cli::cli;
...
-    cli().await
+    if let Err(e) = goose_cli::logging::setup_logging(None, None) {
+        eprintln!("Warning: Failed to initialize telemetry: {}", e);
+    }
+
+    let result = cli().await;
+
+    // Flush telemetry before exiting
+    // Give metrics time to be exported
+    tokio::time::sleep(tokio::time::Duration::from_secs(2)).await;
```
jsibbison-square: Is there any way to check if there is stuff remaining to be flushed rather than only a fixed time (eg. dynamic check with a max timeout)? Also if there is no endpoint setup for otlp we should skip this. 

This delay will add up if people are scripting lots of cli runs.

crates/goose-server/src/routes/reply.rs
```diff
@@ -387,12 +501,216 @@ async fn submit_tool_result(
...
 pub fn routes(state: Arc<AppState>) -> Router {
     Router::new()
         .route(
             "/reply",
             post(reply_handler).layer(DefaultBodyLimit::max(50 * 1024 * 1024)),
         )
+        .route(
+            "/ask",
+            post(ask_handler).layer(DefaultBodyLimit::max(50 * 1024 * 1024)),
+        )
```
jsibbison-square: Why are we creating the /ask route? I thought it was unused?

crates/goose/src/tracing/otlp_layer.rs
```diff
@@ -220,12 +223,13 @@ pub fn create_otlp_metrics_filter() -> FilterFn<impl Fn(&Metadata<'_>) -> bool>
...
     global::shutdown_tracer_provider();
 
-    // Note: There's currently no clean way to shutdown the global meter provider
-    // in the OpenTelemetry Rust SDK. The meter provider will be cleaned up when
-    // the process exits. Individual meter providers can be shut down if you have
-    // a direct reference to them.
+    // Force flush of metrics by waiting a bit
+    // The meter provider doesn't have a direct shutdown method in the current SDK,
+    // but we can give it time to export any pending metrics
+    std::thread::sleep(std::time::Duration::from_millis(500));
```
jsibbison-square, line 232: this is another fixed sleep in addition to the one in the cli, can we just all the sleeps into here, make it conditional on otlp being exported (and ideally dynamic to keep as short as possible)?

crates/goose/src/tracing/rate_limiter.rs
```diff
@@ -0,0 +1,143 @@
...
+                "Starting rate-limited telemetry sender with {}ms delay",
+                rate_limit_ms
+            );
+
+            while let Some(event) = receiver.recv().await {
+                let elapsed = last_send.elapsed();
+                if elapsed < rate_limit_duration {
+                    let sleep_duration = rate_limit_duration - elapsed;
+                    sleep(sleep_duration).await;
+                }
```
jsibbison-square, line 48: Can we rely on the endpoint returning 429's (or non-200s) rather than arbitrary limiting here?

jsibbison-square (approved):
Happy to keep this moving through while we experiement with it. I'm a bit concerned about the arbitrary sleeps on cli exit but otherwise happy to push through.

crates/goose/src/tracing/rate_limiter.rs
Kvadratni: its not a rate limit on the endpoint. the sdk itself didn't want to pipe at the necessary speed and just folded.

crates/goose/src/tracing/otlp_layer.rs
Kvadratni: literally current limitation of sdk. no way to wait for it as I found. I may be overlooking something obvious tho

crates/goose-server/src/routes/reply.rs
Kvadratni: oops.

crates/goose-cli/src/main.rs
Kvadratni: wrapped with a check

PR #3876, opened 2025-08-06
Try to keep key order for Databricks

jamadeo:
Databricks insists on `content` coming before `tool_calls` in the blob. Hopefully they fix this, but until they do, let's try to work around it.

Serde appears to keep struct field ordering when serializing, so using a struct here instead of plain Values _should_ do it.

### DIFF
--- a/crates/goose/src/providers/formats/databricks.rs
+++ b/crates/goose/src/providers/formats/databricks.rs
@@ -9,17 +9,33 @@ use mcp_core::{ToolCall, ToolError};
 use rmcp::model::{AnnotateAble, Content, RawContent, ResourceContents, Role, Tool};
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
+use tracing_subscriber::registry::Data;
+#[derive(Serialize)]
+struct DatabricksMessage {
+    content: Value,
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/formats/databricks.rs
```diff
@@ -9,17 +9,33 @@ use mcp_core::{ToolCall, ToolError};
 use rmcp::model::{AnnotateAble, Content, RawContent, ResourceContents, Role, Tool};
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
+use tracing_subscriber::registry::Data;
```
baxen: nit: unused?

baxen (approved):
LGTM! 

I was getting this very frequently before and couldn't repro on this branch in the last 10m of testing

DOsinga:
I think this must be something that databricks recently broke since we never saw this particular thing and then suddenly started seeing it from the 4th on. then again, we had other conversation bugs in there up to that point, so who knows

DOsinga (approved):
‼️

michaelneale:
oh wow.. wonder if this explains a lot

PR #3878, opened 2025-08-06
fix: extension list not refreshing after installing from deeplink

zanesq:
Fixed race conditions from manually updating extension list by calling `setExtensions`

Extensions is now a computed value that automatically updates when `extensionsList` from context changes (memoized for performance)

Bonus: fixed toast message that was always showing "Removed" for deactivated extensions.

Tested and verified locally with manual and deeplink extensions.


michaelneale:
@zanesq were you able to verify it with the packaged version of the app as well as `just run-ui`? (if not - I can try that out - what should one look for to ensure? how to best check it?)

zanesq:
I did not run it with a packaged version but its fairly safe code just changing how the reactive updates happen rather than manually. To test I just added extensions and edited them both manually and from deeplinks to make sure the list refreshes.

PR #3880, opened 2025-08-06
fixed recipe schedule jobs

lifeizhou-ap:
Fix issue https://github.com/block/goose/issues/3882

**What**
- Used `chat` mode for `foreground` job and `auto` for background job
- Set the extensions for schedule job in legacy mode
- UI
   - enabled the recipe progress/messages in the conversation panel when auto executing foreground schedule job
   - do not show the prompt in the chat input for auto executed foreground schedule job

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -1139,8 +1139,8 @@ impl Agent {
         let mode = session.and_then(|s| s.execution_mode.as_deref());
 
         match mode {
-            Some("foreground") => "auto".to_string(),
-            Some("background") => "chat".to_string(),
+            Some("foreground") => "chat".to_string(),
+            Some("background") => "auto".to_string(),
             _ => config
                 .get_param("GOOSE_MODE")
                 .unwrap_or_else(|_| "auto".to_string()),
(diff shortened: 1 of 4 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -1139,8 +1139,8 @@ impl Agent {
         let mode = session.and_then(|s| s.execution_mode.as_deref());
 
         match mode {
-            Some("foreground") => "auto".to_string(),
-            Some("background") => "chat".to_string(),
+            Some("foreground") => "chat".to_string(),
```
DOsinga, line 1142: I never understood this code, how would chat help in either foreground or background mode?

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -528,7 +530,7 @@ function BaseChatContent({
             chatState={chatState}
             onStop={onStopGoose}
             commandHistory={commandHistory}
-            initialValue={_input || ''}
+            initialValue={recipeConfig?.isScheduledExecution ? '' : _input || ''}
```
DOsinga: let's drop the _ for input here while we are at it

crates/goose/src/agents/agent.rs
lifeizhou-ap: This is a good question.  My understanding is: 

- This is only used in the `default` scheduler mode (not the temporal scheduler mode)
-  For `temporal` mode, it opens a new window with the `foreground` job.  

- I tested `default` mode with `foreground` job, it seems working as the same as background job except different mode.  Just wondering do we still need `foreground` for `default` schedule job? @Kvadratni

PR #3881, opened 2025-08-06
feat: first time automated ollama install experience and openrouter

michaelneale:
along with openrouter one click setup, this adds ability to install and or configure ollama so it works out of the box with goose. This uses the new gpt4-oss model (which seems ok). Also fixes the display of the setup helper/provider guard and some other fixes related specifically to open router. 

For ollama:

* checks if ollama is installed, if not will prompt you to install it
* checks status of ollama
* allows one click to configure goose to use ollama
* streams down the gpt4-oss:20b model if it is not already available

Interested in how we phrase/word this. 

Screenshot 2025-08-06 at 7 28 16 pm

Screenshot 2025-08-06 at 7 28 27 pm

Screenshot 2025-08-06 at 7 29 08 pm
Screenshot 2025-08-06 at 7 29 14 pm

Screenshot 2025-08-06 at 7 30 04 pm

todo: 

- [x] test it in empty state setup
- [x] test without ollama installed
- [x] test without model install
- [x] remove FORCE_SHOW_SETUP flag (testing time only)

To test this: remove your goose config, and startup the goose gui (`just run-ui` from this branch)


ui/desktop/src/components/OllamaSetup.test.tsx
```diff
@@ -0,0 +1,278 @@
...
+  const mockOnCancel = vi.fn();
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    // Default mocks
+    vi.mocked(ollamaDetection.getPreferredModel).mockReturnValue('gpt-oss:20b');
+    vi.mocked(ollamaDetection.getOllamaDownloadUrl).mockReturnValue('https://ollama.com/download');
+  });
+
+  describe('when Ollama is not detected', () => {
```
michaelneale, line 37: not really sure of the value of this test - inspired by @zanesq, this mocks out a lot, so really just checks the component does something (can remove it if it seems contrived)

ui/desktop/src/App.tsx
```diff
@@ -986,12 +986,12 @@ export default function App() {
             if (error instanceof MalformedConfigError) {
               throw error;
             }
-            window.location.hash = '#/welcome';
-            window.history.replaceState({}, '', '#/welcome');
+            window.location.hash = '#/';
```
michaelneale, line 989: without this it will never show the providerguard correctly - and just the wall of providers which we don't want. 

I am curious if it would be worth the effort to mock up/unit test this for App.tsx (or perhaps it is more an e2e test)

michaelneale: (I have a unit test for it)

zanesq: yeah I think this was a fallback to go to welcome before we had provider guard so we can probably do this as long as provider guard is still sending to welcome if no model and provider detected.

unit test makes sense if it can check the url of where it ultimately landed

ui/desktop/src/components/OllamaSetup.test.tsx
zanesq: looks fine to me, mocking is ok if we are only testing that the component renders what it should based on the mocked inputs anyway

kwsantiago:
@michaelneale based on your message [here](https://github.com/block/goose/pull/3635#issuecomment-3161749820), I took the middle-out support from the other branch commits and put them in https://github.com/block/goose/pull/3907, which we can merge into this branch if you'd like.

michaelneale:
ok @zanesq this is finally ready - have tested quite a few permutations and over rides, and fixes a few things to do with openrouter while at it that we can close down - PTAL

Issue #3887, opened 2025-08-06
Goose Stalling


alexhancock:
Goose simply stalls sometimes mid-session

DOsinga:
closing this since this thing has been fixed. something new might be playing up with similar aspects though

Issue #3740, opened 2025-07-30
Stability Issues Affecting Recent Goose Versions


alexhancock:
We are aware of stability issues affecting different recent versions of goose. I am starting this centralized tracker to document and drive each of them to resolutions.

# Goose Stability Issues Tracker

## [v1.1.4](https://github.com/block/goose/releases/tag/v1.1.4)

| Issue | Description | Filer | Status | Resolution PR |
|-------|-------------|---------------------------|--------|---------------|
|    N/A   | Frequent 400s with "Bad Request" when interacting with model providers | alexhancock | Believed resolved | https://github.com/block/goose/pull/3660 |
|   https://github.com/block/goose/issues/3739    | Goose will stop performing tool calls even when it seems to be in an active loop that has not completed, and despite it saying it will call further tools            | alexhancock |Open |  N/A |

## main + [v1.2.0 RC](https://github.com/block/goose/pull/3763)

| Issue | Description  | Filer | Status | Resolution PR |
|-------|-------------|---------------------------|--------|---------------|
|   https://github.com/block/goose/issues/3739    | Goose will stop performing tool calls even when it seems to be in an active loop that has not completed, and despite it saying it will call further tools            | alexhancock | Open |  N/A |
|       | Whenever markdown code files are rendered the app becomes unresponsive to clicks/interaction            | alexhancock | Open |  N/A |
|       | Context token usage tracker frequently displays 0 tokens used after several rounds of tool calling  | alexhancock | Open |  N/A |
|       | Cannot install fetch extension - Times out after 5 minutes with error: ```Error activating extension: Failed to add extension configuration, error: Initialization(Stdio { name: "fetch", cmd: "/Users/USERNAME/proj/goose/goose/ui/desktop/src/bin/uvx", args: ["mcp-server-fetch"], envs: Envs { map: {} }, env_keys: [], timeout: Some(300), description: None, bundled: None }, McpServerError { method: "initialize", server: "", source: ServerBoxError(Elapsed(())) })```  | DOsinga | Open |  N/A |
|      | Context token usage tracker frequently displays 0 tokens used after several rounds of tool calling  | alexhancock | Open |  N/A |
|      | Adding new extension does not reflect in the list of extensions until you do something else  | DOsinga | Open |  N/A |
|   https://github.com/block/goose/issues/3887    | Goose sometimes stalls out mysteriously when in the midst of a session            | alexhancock | Open |  N/A |
---

### Usage Instructions:
[...]

PR #3888, opened 2025-08-06
worlds simplest logging to see where things are blocked

alexhancock:
A commit from @michaelneale - worlds simplest logging to see where things are blocked

Can help us diagnose where and when goose stalls in its interactions with LLMs and during tool calling

Related https://github.com/block/goose/issues/3740 https://github.com/block/goose/issues/3887


alexhancock:
@DOsinga Good recs. I've addressed some of it in a commit.

I am going to see now if I can make it log telemetry info per session and also output a format that can be visualized in something like https://zipkin.io

alexhancock:
Realized we already have some logging infra in the other crates. Going to redo this to use a deduped approach.

alexhancock:
@DOsinga feedback addressed, and some other logging related code has been deduped. Let me know what you think.

crates/goose-cli/src/logging.rs
```diff
@@ -21,33 +19,38 @@ static INIT: Once = Once::new();
 /// Returns the directory where log files should be stored.
 /// Creates the directory structure if it doesn't exist.
 fn get_log_directory() -> Result<PathBuf> {
-    get_log_directory_with_date(None)
+    goose::logging::get_log_directory("cli", true)
 }
 
 /// Internal function that allows specifying a custom date string for testing
+#[cfg(test)]
 fn get_log_directory_with_date(test_date: Option<String>) -> Result<PathBuf> {
```
DOsinga, line 27: I would be supportive of killing the whole with date thing here. we only use it in a test that I stared at for a bit and am not really sure about what it is trying to accomplish. I think the simplifying you are doing here is great and I'd rather double down on that and then see what we need in testing on the goose level rather than here

crates/goose/src/logging.rs
```diff
@@ -0,0 +1,160 @@
...
+    let log_dir = if let Some(date_str) = test_date {
+        component_dir.join(date_str)
+    } else {
+        // Create date-based subdirectory
+        let now = chrono::Local::now();
+        component_dir.join(now.format("%Y-%m-%d").to_string())
+    };
+
+    // Ensure log directory exists
+    fs::create_dir_all(&log_dir).context("Failed to create log directory")?;
```
DOsinga: remove the comment

crates/goose/src/debug_logger.rs
```diff
@@ -0,0 +1,86 @@
...
+/// Log a debug event to the goose-debug.log file
+///
+/// This function logs timestamped events to a debug log file in the appropriate
+/// platform-specific logs directory.
+///
+/// # Arguments
+///
+/// * `event` - The event string to log
+/// ```
+pub fn log_debug_event(event: &str) {
```
DOsinga: why are we rolling our own log thing here and not just rely on tracing::debug!() for this?

michaelneale: mostly as I don't know how/when that works (but yeah should use it) - trace can be written (unbuffered?) to a file? (really want to avoid buffering in case anything is getting lost) so you can watch it in real time

DOsinga: tracing is the library meant for this as far as I know. writing a good logger is not trivial, so relying on something that exist and that we already use is probably the way to go

where I think we should focus our attention is on tooling around it; we write json files to the current logs, can we have something that inspects them live? some light python tooling could be super useful

jamadeo:
@alexhancock @michaelneale I'd also prefer we stick to the tracing crate for this. The subscriber config might be what's making it less useful, but if so we should fix it there.

crates/goose/src/agents/agent.rs
```diff
@@ -418,6 +419,8 @@ impl Agent {
             };
         }
 
+        log_debug_event(&format!("WAITING_TOOL_START: {}", tool_call.name));
+
```
jamadeo: the existing `#[instrument]` should in theory give us what we want here -- if we want to isolate this part of the function there can be an added `span!(...)` here

crates/goose/src/agents/reply_parts.rs
```diff
@@ -169,21 +172,26 @@ impl Agent {
         let provider = provider.clone();
 
         let mut stream = if provider.supports_streaming() {
-            provider
+            log_debug_event("WAITING_LLM_STREAM_START");
+            let stream = provider
```
jamadeo: we might be missing an `#[instrument]` on the streaming function, although I think that would only give us time-to-first-token.

alexhancock:
@jamadeo @DOsinga Good callouts. I have been meaning to get back to this and will try to do so today to make those revisions.

michaelneale (approved):
I think ok when you think it is ready @alexhancock - unless there is another way, we do need to dive into this one way or another

alexhancock:
@michaelneale I let this one languish a bit, sorry. I agree we need it.

Let me come back and do a simpler one similar to your original shortly.

plan.md
```diff
@@ -0,0 +1,8 @@
+Logging
```
jamadeo: Probably didn't mean to commit this

crates/goose/src/logging.rs
```diff
@@ -0,0 +1,134 @@
...
+    use super::*;
+    use std::env;
+    use tempfile::TempDir;
+
+    fn setup_temp_home() -> TempDir {
+        let temp_dir = TempDir::new().unwrap();
+        if cfg!(windows) {
+            env::set_var("USERPROFILE", temp_dir.path());
+        } else {
+            env::set_var("HOME", temp_dir.path());
```
jamadeo: I think tests that set env vars can cause issues, also this one doesn't seem to restore it to what it was before. At minimum probably needs a `#[serial]` if we must set them

jamadeo: if we really want to unit test these, maybe have the main one read $HOME and pass that to another fn that we actually test. `get_base_log_directory` looks like it doesn't need to be `pub`, so you'd only need to do this in one place

plan.md
alexhancock: A constant struggle, as I often use `goose run -i plan.md` then forget them. Will remove!

crates/goose/src/logging.rs
alexhancock: Folded `get_base_log_directory` into the other method, and simplified the testing overall

alexhancock:
> I think ok when you think it is ready @alexhancock - unless there is another way, we do need to dive into this one way or another

This is in now @michaelneale

PR #3890, opened 2025-08-06
feat: linux computer control for android (termux)

shawn111:
Updated,

Just support  linux computer control for android.
For build issue, let termux-package (termux upstream) to handle.

----
Updated,

We need some minor changes to support goose run on termux.
 1. (goose-mcp) linux computer control add termux support
 2. (goose-mcp) -  I use android-patch.sh to handle  xcap
      - latest xcap required ubuntu newer than 20.04 but cross.rs only support 20.04
      - xcap support android build not merge to upstream yet.
      - latest xcap handle w.tilte() is different to xcap 0.0.14
 3.  create [.github/workflows/build-termux-cli.yml](https://github.com/block/goose/pull/3890/files#diff-c92a843ffd904619ba4c9359133dd6f452e2f670877e25cfd93ae8e89642d7c0) to trigger termux build.
  
If people like to try https://github.com/shawn111/goose/releases/download/termux/goose-termux-aarch64.tar.bz2,
it is v1.5.0 now, I'll keep fellow latest release.

Once it merged, we can have canary and release build for termux just like all other platforms.

### DIFF
--- a/BUILDING_LINUX.md
+++ b/BUILDING_LINUX.md
@@ -27,6 +27,18 @@ sudo dnf install dpkg-dev fakeroot gcc gcc-c++ make
 sudo zypper install dpkg fakeroot gcc gcc-c++ make
 ```
 
+**android / termux:**
+
+goose is not officially support termux build yet, you need some minor patch to fix build issues.
+We will publish goose (block-goose) into termux-packages.
+If you want to try there is a non-official build, https://github.com/shawn111/goose/releases/download/termux/goose-termux-aarch64.tar.bz2
+For more details, see: https://github.com/block/goose/pull/3890
+
+```bash
+pkg install rust
+pkg install cmake protobuf clang build-essential
+```
+
 ### Development Tools
 
 - **Rust**: Install via [rustup](https://rustup.rs/)
--- a/crates/goose-mcp/src/computercontroller/platform/mod.rs
+++ b/crates/goose-mcp/src/computercontroller/platform/mod.rs
@@ -8,7 +8,7 @@ pub use self::windows::WindowsAutomation;
 #[cfg(target_os = "macos")]
 pub use self::macos::MacOSAutomation;
 
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "linux", target_os = "android"))]
 pub use self::linux::LinuxAutomation;
 
 pub trait SystemAutomation: Send + Sync {


shawn111:
@Kvadratni  could you help me review the commit 2,3?

since xcap is not support android/termux, I create a cargo build feature screen_capture, let android default not enable it.

shawn111:
> 2\. (goose) update dependencies (upstream fixed termux build)
> 3\. add termux build job into build-cli.yml
> 
> Once it merged, we can have canary and

we don't those change, just need wait for xcap change to support android build.

michaelneale:
woah @shawn111 this is cool!

.github/workflows/build-termux-cli.yml
```diff
@@ -0,0 +1,99 @@
...
+          # Create a directory for the package contents
+          /entrypoint.sh mkdir -p "target/release/goose-package"
+          /entrypoint.sh cp "target/release/goose" "target/release/goose-package/"
+          /entrypoint.sh tar -cjf "goose-termux-${{ matrix.architecture }}.tar.bz2" -C target/release/goose-package .
+      - name: Upload CLI artifact
+        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # pin@v4
+        with:
+          name: goose-termux-${{ matrix.architecture }}
+          path: goose-termux-${{ matrix.architecture }}.tar.bz2
+  release:
```
michaelneale: never quite sure on how to test changes to release workflows, or additions - ideas?

shawn111: Maybe I need remove this part,  it for my github repo to create the termux release.

https://github.com/shawn111/goose/releases/tag/termux

shawn111: I remove the release part, and make the trigger rule as same as canary.yml. Once the PR merge, it will trigger build but we need download from github build result.

Once it is stable, we need add ```needs: build-cli-termux:``` in canary.yml and release.yml. Otherwise, users will hard to find where to download.

.github/workflows/build-termux-cli.yml
```diff
@@ -0,0 +1,73 @@
...
+          /entrypoint.sh ./crates/goose-mcp/android-patch.sh
+          /entrypoint.sh cargo build --release
+          # Create a directory for the package contents
+          /entrypoint.sh mkdir -p "target/release/goose-package"
+          /entrypoint.sh cp "target/release/goose" "target/release/goose-package/"
+          /entrypoint.sh tar -cjf "goose-termux-${{ matrix.architecture }}.tar.bz2" -C target/release/goose-package .
+      - name: Upload CLI artifact
+        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # pin@v4
+        with:
+          name: goose-termux-${{ matrix.architecture }}
```
michaelneale: should this have android explicitly in the name? 

Also, there are quite a few downloads, wondering where is best place to store the termux versions, as they wont' be the mainstream ones people look for (how would one usually get tools for android?)

shawn111: How about naming the artifacts either:

goose-android-termux-${{ matrix.architecture }}, or
goose-${{ matrix.architecture }}-android-termux (to match the style of goose-aarch64-unknown-linux-gnu)?

The reason I suggest including termux in the name is that Termux is not just a terminal emulator—it provides a full Linux environment. If we use only android in the name, people might get confused.

The Termux package repo: https://github.com/termux/termux-packages (lots of Rust packages there).
But only Goose can provide a canary version for Termux: https://github.com/block/goose/releases/tag/canary
.

Not many upstream projects ship Termux builds directly, because the GitHub Action setup is still relatively new and not very easy to follow. Example snippet:

```
container:
  image: termux/termux-docker:${{ matrix.architecture }}
  volumes:
    - /tmp/node20:/__e/node20
steps:
  - run: /entrypoint.sh pkg install -y nodejs-lts
```

I made three commits in this PR:

- I think the first commit is clean and simple: https://github.com/block/goose/pull/3890/commits/b0dc430a38eaa557181e867c52faecb9bd9ec761

The second and third commits are a bit messy, to be honest.
Since Termux isn’t a major platform, I expect my patch won’t affect other platforms.

shawn111:
```
pkg search goose                                                (main) 13:47
...
Full Text Search... Done
goose/stable 3.24.3 aarch64
  A database migration tool. Supports SQL migrations and Go functions.
```

There is another goose(db) in termux repo.
How about create a ```block-goose``` in termux repo?

michaelneale: @shawn111 right - so is it possible that goose can publish "block-goose" to termux repo - and not have it on the goose download page? (would it be useful on the goose download page, or with termux to people 90% of the time get it from within termux? not sure how that works).

shawn111: @michaelneale ok. could we still keep "[feat: linux computer control for android](https://github.com/block/goose/pull/3890/commits/b0dc430a38eaa557181e867c52faecb9bd9ec761)"? I this part is in goose.

michaelneale: @shawn111 looks like you got rid of the pipeline changes, was that intentional? only need this tiny change?

shawn111: @michaelneale I add  Add Termux/Android build instructions, let people know they can build goose by themself.
Yes, only "[feat: linux computer control for android](https://github.com/block/goose/pull/3890/commits/221f20c94081e3b46a348be11714381c26cf79fd)"
is the most required for android.

Others are just minor build issue.

shawn111: @michaelneale  xcap release 0.7.0 support android but need more system dependent package like wayland-devel related packages.

Current cross.rs env need to update to support.
https://github.com/shawn111/goose/actions/runs/17224775977/job/48867248600



I thought that is not important for other platforms, we can fix them later.


For termux, like what you says just publish to termux-packages.

shawn111: @michaelneale goose source code might be to fat https://github.com/block/goose/issues/4315

robertkirkman: > Not many upstream projects ship Termux builds directly, because the GitHub > Action setup is still relatively new and not very easy to follow. Example snippet:

> ```
> container:
>   image: termux/termux-docker:${{ matrix.architecture }}
>   volumes:
>     - /tmp/node20:/__e/node20
> steps:
>   - run: /entrypoint.sh pkg install -y nodejs-lts
> ```

Hi, I developed the code you are quoting, I am very happy that it is useful for someone!

I maintain the reference example from which this code originated here:

https://github.com/robertkirkman/termux-on-gha/blob/master/.github/workflows/build-and-test.yml

Either of you can let me know if you have any questions or comments or need help.

In

- https://github.com/termux/termux-packages/issues/25820

it seems that the ability to get block-goose from the command **`pkg install block-goose`** in Termux is desired?

I'm most likely capable of making the code and merging the PR required to Termux itself for that to happen, but, I would just like to lay out the facts of the situation,

When deciding what projects to make available from `pkg install` in Termux, I have a responsibility to consider whether they meet certain critera, and I need to try to make my decisions roughly line up with what the other Termux contributors would choose if they were in my place.

Having **a lot of popularity in the software and a lot of other UNIX-like software distros packaging the software, shown by [Repology](https://repology.org/project/block-goose-cli/versions), is one of the criteria** (but not the only one and there can sometimes be exceptions for special software).

luckily, block-goose _has_ gotten into one of the other Repology repos already, Homebrew! This is just one, but it's an indication that block-goose might be popular enough to justify including it in Termux repo also.

Do you guys believe it is really important and desirable to get Goose in Termux through **`pkg install`**, or do you think it would also be OK to provide a download for Termux users from the block/goose repo instead?

I don't personally use LLM tools, so I'm never very sure how useful a particular LLM tool is for developers who use it, or how popular it would end up being if it were made available from `pkg install` in Termux, just because of my inexperience with the subject.

robertkirkman: > goose-${{ matrix.architecture }}-android-termux (to match the style of goose-aarch64-unknown-linux-gnu)?

I would suggest for this pattern, "**goose-${{ matrix.architecture }}-unknown-linux-android**", "**goose-${{ matrix.architecture }}-linux-android**", or "**goose-${{ matrix.architecture }}-linux-android24**", could be used (but it's up to you what the ultimate decision would be or whether to add an android target here at all),

because Termux, to skip some technical explanation, can be considered **Android 7.0-or-newer, also known as Android API level 24 or newer** - it is, for the most part, **native code binary-compatible** directly with Android 7 (and newer including all Android 7+ devices through Android 16+) (skimming over some technical details).

Another way to say that is, that the word "termux" is not usually used to identify the architecture and platform name of the operating system that must be targeted for Termux to run binaries - **that operating system that must be targeted is Android 7.0**, so generally that is how Termux can be described and targeted.

Sometimes, the word "termux" _is_ used for some platform identification like that, but usually only when there is an important reason it's necessary, like a major difference between Termux and the behavior of standard Android that ends up being important for the porting code in a specific way.

robertkirkman: > The reason I suggest including termux in the name is that Termux is not just a terminal emulator—it provides a full Linux environment. If we use only android in the name, people might get confused.

In order to avoid confusion, in my opinion, the best name to choose is "Android" for the name of the OS and "Termux" for the name of the terminal emulator (name both and call Android the OS and Termux the terminal emulator, so only use "Termux" in places where the terminal emulator to use is described, such as installation documentation for example, and not in the name of the target platform), _unless_ there is a **known reason that the provided build _cannot_ also work in any other different Android Terminal emulator app and has some kind of hard requirement for Termux** (sometimes there can be a reason for that being the case, but I believe those cases are somewhat uncommon, depending on one's perspective)

robertkirkman: > (would it be useful on the goose download page, or with termux to people 90% of the time get it from within termux? not sure how that works).

The answer is probably very similar to the answer it would be for Desktop Linux and Mac - **some people may prefer to download the software from GitHub, but others may prefer to download from `apt install` or `brew install`, `pkg install`, etc. from the package manager** (which is currently possible for `block-goose`, it seems, for Mac Homebrew but not really Desktop Linux, at least yet).

Both are possible and could end up being choices, but which one people prefer more can depend on a lot of subtle nuances with the software itself, and how actively-maintained and well-supported the builds are, whichever places they come from.

shawn111: @robertkirkman thanks a lot about the github action, it is so useful for me. Help me much easy to fellow upstream.
Otherwire, my phone is too slow to build code.

Why I'm try merge block-goose into termux-package, not just because the github.
block-goose also need some patches (for xcap) to build android.

@michaelneale How you think about it? Could goose provide android build? If it is good, I will add the github action workflow back.

robertkirkman: > block-goose also need some patches (for xcap) to build android.

If the patches do not mess up support for other platforms, then yes I think it would be helpful overall to store them here in the upstream even if block-goose is also packaged in Termux eventually, because it would unify the code and reduce the amount of patches needed inside the package on the Termux repository side of the code.

I would just make one suggestion: **if possible, try to keep the features and behavior of the software as close as possible to its behavior on GNU/Linux**, as far as it's convenient to, and **if you end up needing to remove features in the Android version that exist in the GNU/Linux version, please try making a list of the differences between them**, so it's clear what's supported and what isn't supported.

shawn111: @robertkirkman thanks, I sloved the build issue of xcap https://github.com/block/goose/pull/4344

https://github.com/shawn111/goose/actions/runs/17236288907

shawn111: Let's make this PR just for one commit. I can create github action after this PR and #4344

DOsinga: >How you think about it? Could goose provide android build? If it is good, I will add the github action workflow back.

I am not against that, I think we should mark it somehow as experimental. I'd love to give this a try personally, but I don't know how much time the core team has to support this

shawn111: @DOsinga I can keep github action in my personal repo first, but if #3890 #4344 can merge into goose code base will be better. Then people can build android build without any extra patches.

DOsinga: sounds good!

michaelneale:
I can't seem to test this locally - weird merge conflicts, but I think the change in code looks ok

shawn111:
> I can't seem to test this locally - weird merge conflicts, but I think the change in code looks ok

should I rebase main?

shawn111:
rebase main

michaelneale:
thanks, should be good to go when green

PR #3895, opened 2025-08-06
added notes about reducing context window by referencing external files

iandouglas:
Since `.goosehints` and every memory stored in the Memory Extension are sent with every request to the LLM, I added notes on how to take lengthy notes/memories, write them into external files, and use shorter rules/memories to access those files as needed. While this will likely cause an extra round of back-and-forth between Goose and the LLM, it should reduce the context window from sending long sets of rules/guidelines unrelated to a given request.


documentation/docs/guides/using-goosehints.md
```diff
@@ -120,4 +120,4 @@ Like prompts, this is not an extensive list to shape your `.goosehints` file. Yo
 - **Keep file updated**: Regularly update the `.goosehints` file to reflect any changes in project protocols or priorities.
 - **Be concise**: Make sure the content is straightforward and to the point, ensuring Goose can quickly parse and act on the information.
 - **Start small**: Create a small set of clear, specific hints and gradually expand them based on your needs. This makes it easier to understand how Goose interprets and applies your instructions.
-
+- **Reference other files**: Since `.goosehints` is sent with every request, writing rules like "if I ask for help writing JavaScript, refer to /path/to/javascript_nodes.txt" will help Goose understand where to find additional context and reduce context window size.
```
blackgirlbytes:
```suggestion
- **Reference other files**: Point Goose to relevant files like /docs/style.md or /scripts/validation.js to reduce repetition and keep instructions lightweight.
```

blackgirlbytes:
Since you're already in this file..would you be okay with removing

`You can name your context files differently -- e.g. AGENTS.md -- and Goose can still pick them up. Configure the CONTEXT_FILE_NAMES setting!`

This wasnt you that added this, but Diane had a ticket on it. You're already here..we might as well remove it.

blackgirlbytes:
what if after this line

```
A good time to consider adding a `.goosehints` file is when you find yourself repeating prompts, or providing the same kind of instructions multiple times. It's also a great way to provide a lot of context which might be better suited in a file.
```

We add this:

```
Goosehints are loaded at the start of your session and become part of the system prompt sent with every request. This means the content of .goosehints contributes to token usage, so keeping it concise can save both cost and processing time.
```

blackgirlbytes:
okay i have one more note. What do you think of this? I think the pro tip feels a little tacked on..like an after the fact thing..so in the area where it says

```
Why Use Memory?
With the Memory extension, you’re not just storing static notes, you’re teaching Goose how to assist you better. Imagine telling Goose:

learn everything about MCP servers and save it to memory.

Later, you can ask:

utilizing our MCP server knowledge help me build an MCP server.

Goose will recall everything you’ve saved as long as you instruct it to remember. This makes it easier to have consistent results when working with Goose.
```


What if we add


Memories are retrieved dynamically based on your request and sent to the LLM with every prompt. For large or detailed instructions, you can store them in files and instruct Goose to reference those files using a prompt like the one below:
```
Remember that if I ask for help writing JavaScript, refer to "/path/to/javascript_nodes.txt" and follow the instructions in that file.
```

blackgirlbytes (approved):

PR #3901, opened 2025-08-06
Do not fetch pricing information if GOOSE_CLI_SHOW_COST is false

ayax79:
The call to openrouter.ai to fetch pricing information generates a lot of annoying WARN messages in cases where it is blocked by corporate firewalls. It seems that if displaying costing is turned off, the fetch from openrouter.ai should not happen.


michaelneale (approved):
thanks @ayax79 seems to make sense to me.

PR #3907, opened 2025-08-07
feat: middle-out message compression for OpenRouter

kwsantiago:
Implements middle-out message compression for OpenRouter to handle context limit errors by removing messages from the middle of conversations when limits are exceeded, auto-enabling for models ≤8k context.

For reference: https://github.com/block/goose/pull/3881#issuecomment-3162154993

### DIFF
--- a/crates/goose/src/context_mgmt/middle_out.rs
+++ b/crates/goose/src/context_mgmt/middle_out.rs
@@ -1,0 +1,338 @@
+use crate::message::Message;
+use anyhow::Result;
+use rmcp::model::Role;
+use std::collections::HashSet;
+use tracing::{debug, warn};
+
+use super::truncate::{truncate_messages, TruncationStrategy};
+
+#[allow(dead_code)]
+const ANTHROPIC_MAX_MESSAGES: usize = 200000;
+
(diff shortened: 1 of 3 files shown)


michaelneale:
@kwsantiago thanks - but if I read correctly, there is no need to conditionally apply this, we can just always have it set for openrouter, it doesn't seem to do anything until it needs to, ie more of a safeguard if I read it correctly? 

https://openrouter.ai/docs/features/message-transforms

but I am trying the simple approach first in the other branch (will report back if not good).

michaelneale:
@kwsantiago thanks - but if I read correctly, there is no need to conditionally apply this, we can just always have it set for openrouter, it doesn't seem to do anything until it needs to, ie more of a safeguard if I read it correctly? 

https://openrouter.ai/docs/features/message-transforms

but I am trying the simple approach first in the other branch (will report back if not good).

kwsantiago:
Sounds good @michaelneale I'll leave this branch as is for the time being.

If you need me to make any changes here or close this out, please let me know. Cheers!

michaelneale:
closing this as will be addressed in other PR, thanks again @kwsantiago this was really cool

PR #3908, opened 2025-08-07
Parallize initiazation of extensions

flat35hd99:


flat35hd99:
Do same by https://github.com/block/goose/pull/3949

PR #3909, opened 2025-08-07
Fix token usage tracking for streaming providers

jajimajp:
## Problem

When using streaming providers with the goose CLI, the context token count remains at 0 and doesn't update properly:

```sh
( O)> Hello

(...snip...)

( O)>
```

## Root Cause

The current implementation overwrites usage metrics with each streaming event instead of properly accumulating them. When streaming APIs send a final event with None values, it resets the usage count to 0.

## Solution

Modified the update_session_metrics method to:
- Properly track cumulative token usage across streaming events
- Handle None values gracefully without resetting existing counts

### DIFF
--- a/crates/goose/src/agents/reply_parts.rs
+++ b/crates/goose/src/agents/reply_parts.rs
@@ -289,6 +289,10 @@ impl Agent {
 
         metadata.schedule_id = session_config.schedule_id.clone();
 
+        let old_total_tokens = metadata.total_tokens.unwrap_or(0);
+        let old_input_tokens = metadata.input_tokens.unwrap_or(0);
+        let old_output_tokens = metadata.output_tokens.unwrap_or(0);
+
         // For streaming providers, the usage values provided are cumulative totals for the current response.
         // We should use the maximum of the current value and the new value to handle streaming updates correctly.
         let take_max = |current: Option<i32>, new: Option<i32>| -> Option<i32> {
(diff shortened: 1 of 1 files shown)


michaelneale:
thanks @jajimajp - what is odd, I use anthropic/databricks and didn't see it show zero, can you confirm what provider you were using and model etc? 
(good catch if there are some cases, but I wonder how it worked at all)

jajimajp:
> thanks @jajimajp - what is odd, I use anthropic/databricks and didn't see it show zero, can you confirm what provider you were using and model etc? (good catch if there are some cases, but I wonder how it worked at all)

I am using an OPENAI-compatible proxy with the model sonnet-4. Therefore, the provider is OPENAI. This change worked well in my environment, but let me verify again whether the same issue occurs with the regular OPENAI API.

michaelneale:
also if you can run cargo fmt and clippy etc.

michaelneale:
I hand tested this with databricks, openai direct, and anthropic, seems ok, but @katzdave sorry to hassle again - but you ok with this? I am just not sure how it works. I am wary that it works without with openai direct, but with a proxy, was different (sounds like 3rd party proxy may be returning incorrect values?)

katzdave:
Can we just add a test that verifies it does the the same thing for the non-streaming case. Generally makes sense to me though on the streaming side.

I can't remember my exact setup but I was seeing some kind of 0 issue a couple weeks ago; I think with openai on databricks.

michaelneale:
@jajimajp able to add any tests?

jajimajp:
@michaelneale @katzdave 

Thanks for the feedback. I've added a test case to verify the non-streaming behavior as requested.

Additionally, I manually verified that non-streaming works correctly with my OpenAI provider by temporarily making the following change:

```diff
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -206,7 +206,7 @@ impl Provider for OpenAiProvider {
  }
 
  fn supports_streaming(&self) -> bool {
-     true
+     false
  }
 
  async fn stream(
```

It seems to be working correctly. Let me know if there's anything else.

jajimajp:
force pushed to sign previous commits

zanesq:
Still in progress?

jajimajp:
> Still in progress?

No, I am waiting for comments on this: https://github.com/block/goose/pull/3909#issuecomment-3177696246

angelahning:
> > Still in progress?
> 
> No, I am waiting for comments on this: [#3909 (comment)](https://github.com/block/goose/pull/3909#issuecomment-3177696246)

@michaelneale @katzdave ptal 🙏

katzdave:
> @michaelneale @katzdave
> 
> Thanks for the feedback. I've added a test case to verify the non-streaming behavior as requested.
> 
> Additionally, I manually verified that non-streaming works correctly with my OpenAI provider by temporarily making the following change:
> 
> ```diff
> --- a/crates/goose/src/providers/openai.rs
> +++ b/crates/goose/src/providers/openai.rs
> @@ -206,7 +206,7 @@ impl Provider for OpenAiProvider {
>   }
>  
>   fn supports_streaming(&self) -> bool {
> -     true
> +     false
>   }
>  
>   async fn stream(
> ```
> 
> It seems to be working correctly. Let me know if there's anything else.

Sorry for the long delay @jajimajp I took a really close look at this with databricks streaming provider, and things seem to be working correctly for me without this change.

To me it looks like I'm receiving the new token counts from the new request rather than an accumulated one; Seeing it shift from 5k -> 5.3k after a request that looks at a small file and adds a tiny bit to the context window.

Are you still having this 0 token count issue with streaming providers, and if so which provider? I know they we pretty new at the time you made this.

DOsinga:
as this seems fixed on main, can we close this?

jajimajp:
It seems that the changes I made have already been addressed in the current main, so I will close this PR. Thank you all for the time and effort you put into reviewing.

DOsinga:
thank you for your contribution and keep an eye open for this sort of thing. providers are all slightly different and getting the token count right is hard!

PR #3911, opened 2025-08-07
feat: Add `GOOSE_TERMINAL` env variable to spawned terminals

iamrajiv:
Fixes: 3827


michaelneale:
I think this is ok, good use of env!

michaelneale (approved):
LGTM - seems sensible

PR #3920, opened 2025-08-07
Sanitize Tags Unicode Block

amed-xyz:
Mitigate Unicode-based prompt injection attacks where attackers embed invisible Unicode characters to smuggle hidden commands past user inspection and into LLM processing.

Sanitize text input at the single point where all text content enters the system: `Message::with_text()` method.

The specific threat I'm addressing:
- Complete ASCII mirror: Invisible versions of all ASCII characters (a-z, A-Z, 0-9, symbols)
- Completely invisible: Designed to be unrendered by tag-unaware implementations
- LLM readable: Training data included these characters, tokenizers process them
- Systematic encoding: Complete hidden communication channel for steganographic attacks


DOsinga (approved):
nice. you know what, though, I would just drop the sanitize_needed flag for now. we can talk to @spencrmartin about what we should do in the UI when this happens and then calculate backwards from there in a follow up PR. what do you think?

PR #3929, opened 2025-08-07
dynamic model context limit config

Kvadratni:
O-shot by gpt 5

##Repository-sourced model limits:

### Goose now first attempts to fetch model-limits.yaml from the repo’s raw GitHub URL:
https://raw.githubusercontent.com/block/goose/main/model-limits.yaml
### Fallbacks:
If that fails, it uses the built-in default set.
You can override the remote URL via GOOSE_MODEL_LIMITS_URL, e.g. to test a branch or self-hosted raw file.
File format support:

### YAML or JSON array of objects:
pattern: "gpt-4o" context_limit: 128000
Or wrapped under models, model_limits, or limits key.

### GPT-5 addition:

Added GPT-5 with a 272,000 token context limit to the new model-limits.yaml checked into the repo root.
Pattern is a simple substring match (case-sensitive: "GPT-5").

### DIFF
--- a/crates/goose/src/model.rs
+++ b/crates/goose/src/model.rs
@@ -1,8 +1,15 @@
 use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize};
+use std::time::Duration;
 use thiserror::Error;
 
 const DEFAULT_CONTEXT_LIMIT: usize = 128_000;
+
+// Attempt to fetch from Git repo first (raw URL), then fall back to local file, then defaults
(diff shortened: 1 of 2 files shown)


DOsinga (changes requested):
Two things:
* I don't think this is how we want to do this. Model limits should come from the providers
* I'm sorry for gpt-5, but this code is no good. please review it yourself before offering it up (unless you didn't mean too, then convert it to draft)

DOsinga:
closing this due to non activity - feel free to reopen when comments are addressed

PR #3930, opened 2025-08-07
add missing dependencies for developer setup

burka:
I needed to install **libxcb1-dev libxcb-util-dev protobuf-compiler** to compile the goose cli on a quite clean system. Maybe that should be added to the BUILDING_LINUX.md


michaelneale (approved):
thanks, yes that helps (I think hermit may add those but good to have that in there anyway)

PR #3157, opened 2025-06-30
feat: add OAuth provider abstraction for CLI configuration

atarantino:
It states in the [getting started / providers](https://block.github.io/goose/docs/getting-started/providers) page that GitHub CoPilot uses device code authentication flow for secure access. Unfortunately, the cli and web app prompt the user for a GITHUB_COPILOT_TOKEN instead. There is an [open issue](https://github.com/block/goose/issues/3037) about this as well. 

These changes only fix this for the cli. However, once you successfully configure CoPilot from the cli, the web app will use those credentials. 

The contents of this pull request were substantially written using opencode with Claude 4. I've reviewed to the best of my ability (I have not extensively programmed in Rust).

PR #3931, opened 2025-08-07
Reenable copilot

ayax79:
Since PR [3157](https://github.com/block/goose/pull/3157) has been merged, we copilot should be re-enabled.

### DIFF
--- a/crates/goose/src/providers/factory.rs
+++ b/crates/goose/src/providers/factory.rs
@@ -1,7 +1,28 @@
 use std::sync::{Arc, RwLock};
 
 use super::{
-    anthropic::AnthropicProvider, azure::AzureProvider, base::{Provider, ProviderMetadata}, bedrock::BedrockProvider, claude_code::ClaudeCodeProvider, cursor_agent::CursorAgentProvider, databricks::DatabricksProvider, gcpvertexai::GcpVertexAIProvider, gemini_cli::GeminiCliProvider, githubcopilot::GithubCopilotProvider, google::GoogleProvider, groq::GroqProvider, lead_worker::LeadWorkerProvider, litellm::LiteLLMProvider, ollama::OllamaProvider, openai::OpenAiProvider, openrouter::OpenRouterProvider, provider_registry::ProviderRegistry, sagemaker_tgi::SageMakerTgiProvider, snowflake::SnowflakeProvider, venice::VeniceProvider, xai::XaiProvider
+    anthropic::AnthropicProvider,
+    azure::AzureProvider,
+    base::{Provider, ProviderMetadata},
+    bedrock::BedrockProvider,
+    claude_code::ClaudeCodeProvider,
+    cursor_agent::CursorAgentProvider,
+    databricks::DatabricksProvider,
+    gcpvertexai::GcpVertexAIProvider,
+    gemini_cli::GeminiCliProvider,
+    githubcopilot::GithubCopilotProvider,
+    google::GoogleProvider,
+    groq::GroqProvider,
+    lead_worker::LeadWorkerProvider,
+    litellm::LiteLLMProvider,
+    ollama::OllamaProvider,
+    openai::OpenAiProvider,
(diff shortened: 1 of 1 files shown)


taniacryptid:
Thank you for sharing this! Let me tag folks on the team for review. Just want to make sure in adding copilot back, that it does not introduce bugs like it did before!

DOsinga:
can we add co-pilot to the scenario tests if it works so we can see that it does/

ayax79:
> can we add co-pilot to the scenario tests if it works so we can see that it does/

@DOsinga are there are scenario tests today that work with oauth flows? In the goose-cli/src/scenario_tests directory, existing tests seem to use API tokens.

DOsinga:
> > can we add co-pilot to the scenario tests if it works so we can see that it does/
> 
> @DOsinga are there are scenario tests today that work with oauth flows? In the goose-cli/src/scenario_tests directory, existing tests seem to use API tokens.

I don't think so. would be good though

ayax79:
> > > can we add co-pilot to the scenario tests if it works so we can see that it does/
> > 
> > 
> > @DOsinga are there are scenario tests today that work with oauth flows? In the goose-cli/src/scenario_tests directory, existing tests seem to use API tokens.
> 
> I don't think so. would be good though

The tricky part will be needing to replicate the interactions with the browser. I don't know of a good way to handle this in a headless fashion.

jmdaly:
Hi @DOsinga, I wonder if the Goose dev team would be open to merging this without tests for now, maybe placing it behind the `ALPHA_FEATURES` flag or something? Then someone could work on the scenario tests afterward. There are a few of us on Discord who use the Copilot provider and it would be awesome to not have to build a patched version from source :) Anyway, just a thought!

taniacryptid:
What do you think, @DOsinga @michaelneale @alexhancock @zanesq ? I understand we want to be mindful of any issues.

DOsinga:
> Hi @DOsinga, I wonder if the Goose dev team would be open to merging this without tests for now, maybe placing it behind the `ALPHA_FEATURES` flag or something? Then someone could work on the scenario tests afterward. There are a few of us on Discord who use the Copilot provider and it would be awesome to not have to build a patched version from source :) Anyway, just a thought!

that sounds alright; could you have a look at how hard it would be to add a feature to the providers to mark them up as alpha/beta? there are frankly a whole bunch of providers that are, eh, undertested

ayax79:
> > Hi @DOsinga, I wonder if the Goose dev team would be open to merging this without tests for now, maybe placing it behind the `ALPHA_FEATURES` flag or something? Then someone could work on the scenario tests afterward. There are a few of us on Discord who use the Copilot provider and it would be awesome to not have to build a patched version from source :) Anyway, just a thought!
> 
> that sounds alright; could you have a look at how hard it would be to add a feature to the providers to mark them up as alpha/beta? there are frankly a whole bunch of providers that are, eh, undertested

There is also a bit of a chicken and egg problem here. A lot of corporate users are going to need to use goose through the github copilot support. It's a pretty common pattern to block all providers except copilot (e.g. legal compliance / contracts with Microsoft, etc). In my case, I am working on getting goose to be approved tech at my company and this is the only blocker. If we can get this over the line, we can increase the number of people using this feature and also contributing to project.

jmdaly:
> A lot of corporate users are going to need to use goose through the github copilot support.

This is exactly my case :)

roshbhatia:
> There is also a bit of a chicken and egg problem here. A lot of corporate users are going to need to use goose through the github copilot support. It's a pretty common pattern to block all providers except copilot (e.g. legal compliance / contracts with Microsoft, etc).

Same here. Rather than implement alpha/beta gating (if that's not already implemented) could we maybe rename the provider to have alpha in the title as an interim step? Or something similarly user-facing that denotes some level of instability without needing much implementation so that users like us can start using copilot again? :)

taniacryptid:
Yay, happy to get this merged!

PR #3933, opened 2025-08-07
chore: remove vector search tool selection strategy

wendytang:
* clean up - vector strategy no longer used 
* keeps llm strategy of tool router
* removes lancedb dependency as not needed

image

### DIFF
--- a/crates/goose-cli/src/commands/configure.rs
+++ b/crates/goose-cli/src/commands/configure.rs
@@ -1179,7 +1179,7 @@ pub async fn configure_settings_dialog() -> Result<(), Box<dyn Error>> {
         .item(
             "goose_router_strategy",
             "Router Tool Selection Strategy",
-            "Configure the strategy for selecting tools to use (experimental)",
+            "Experimental: configure a strategy for auto selecting tools to use",
         )
         .item(
             "tool_permission",
--- a/crates/goose/src/agents/prompt_manager.rs
+++ b/crates/goose/src/agents/prompt_manager.rs
@@ -236,7 +236,7 @@ mod tests {
         manager.set_system_prompt_override(malicious_override.to_string());
 
         let result =
-            manager.build_system_prompt(vec![], None, Value::String("".to_string()), None, None);
+            manager.build_system_prompt(vec![], None, Value::String("".to_string()), None, false);
 
         assert!(!result.contains('\u{E0041}'));
         assert!(!result.contains('\u{E0042}'));
@@ -252,7 +252,7 @@ mod tests {
         manager.add_system_prompt_extra(malicious_extra.to_string());
 
         let result =
-            manager.build_system_prompt(vec![], None, Value::String("".to_string()), None, None);
+            manager.build_system_prompt(vec![], None, Value::String("".to_string()), None, false);
 
         assert!(!result.contains('\u{E0041}'));
         assert!(!result.contains('\u{E0042}'));
@@ -269,7 +269,7 @@ mod tests {
         manager.add_system_prompt_extra("Third\u{E0043}instruction".to_string());
 
         let result =
-            manager.build_system_prompt(vec![], None, Value::String("".to_string()), None, None);
+            manager.build_system_prompt(vec![], None, Value::String("".to_string()), None, false);
 
         assert!(!result.contains('\u{E0041}'));
         assert!(!result.contains('\u{E0042}'));
@@ -286,7 +286,7 @@ mod tests {
         manager.add_system_prompt_extra(legitimate_unicode.to_string());
 
         let result =
-            manager.build_system_prompt(vec![], None, Value::String("".to_string()), None, None);
+            manager.build_system_prompt(vec![], None, Value::String("".to_string()), None, false);
 
         assert!(result.contains("世界"));
         assert!(result.contains("🌍"));
@@ -308,7 +308,7 @@ mod tests {
             None,
             Value::String("".to_string()),
             None,
-            None,
+            false,
         );
 
         assert!(!result.contains('\u{E0041}'));


crates/goose/src/agents/router_tool_selector.rs
```diff
@@ -26,7 +23,6 @@ struct ToolSelectorContext {
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum RouterToolSelectionStrategy {
-    Vector,
     Llm,
```
DOsinga: Like here, having an enum with one value doesn't make that much sense

DOsinga (commented):
Thanks for doing this! I'll have a closer look later, but two things to consider:

* I think we should drop the whole notion of a tool selection strategy. If we have a default strategy which is just to do what the user said and an LLM one that does something smarter, we should turn it into a boolean it feels like. Smart tool selection enabled or not.

* Can you update the dependencies? One of the reason we want this out is that it causes binary size bloat.

crates/goose/src/providers/openrouter.rs
```diff
@@ -122,29 +122,6 @@ fn update_request_for_anthropic(original_payload: &Value) -> Value {
         .and_then(|obj| obj.get_mut("messages"))
         .and_then(|messages| messages.as_array_mut())
     {
-        // Add "cache_control" to the last and second-to-last "user" messages.
```
michaelneale: is this related to the vector strategy @wendytang ?

michaelneale:
@wendytang looks good - like @DOsinga can we get rid of lancedb dep now? 

Secondly, seems un unrelated change to openrouter?

crates/goose/src/agents/router_tool_selector.rs
michaelneale: I wonder if vector would come back (or perhaps some other local semantic scoring algorithm?)

michaelneale:
I like this! 

@DOsinga well this removes a bit of weight which isn't needed (lancedb is probably good to speed things up at least in compile times?). 

@wendytang not sure about that openrouter change, but otherwise probably pretty good to tidy this up if we do want to get away from this strategy totally (and then simplify?).

crates/goose-cli/src/commands/configure.rs
```diff
@@ -1177,9 +1177,9 @@ pub fn configure_goose_router_strategy_dialog() -> Result<(), Box<dyn Error>> {
 
     let strategy = cliclack::select("Which router strategy would you like to use?")
         .item(
-            "vector",
-            "Vector Strategy",
-            "Use vector-based similarity to select tools",
+            "llm",
```
michaelneale: previously was LLM not selectable? 

I guess like @DOsinga says - do we have some inkling we will give a strategy choice? if not, can just be simple on-off for tool selection automation I think and reduce the code even more?

crates/goose/src/providers/openrouter.rs
wendytang: This slipped in, good catch!

crates/goose-cli/src/commands/configure.rs
wendytang: i think that would be fine - what would be the mechanism to turn on/off? wouldn't it still be through cli config and ui setting selection?

wendytang: *updated to be a simple on / off

wendytang:
simple enable/disable 
image

michaelneale (approved):
nice one @wendytang seems to work well (tried it on CLI and showed up in GUI). 

@DOsinga I think we can get this one in - it does change some env var names (your favourite thing) but that is ok, I don't think anyone depends on them in this case and even if so, no choice.

michaelneale:
ugh sorry @wendytang some colliding changes landed, any chance can take another look?

crates/goose-cli/src/commands/configure.rs
```diff
@@ -1170,40 +1170,32 @@ pub fn configure_goose_mode_dialog() -> Result<(), Box<dyn Error>> {
 pub fn configure_goose_router_strategy_dialog() -> Result<(), Box<dyn Error>> {
     let config = Config::global();
 
-    // Check if GOOSE_ROUTER_STRATEGY is set as an environment variable
-    if std::env::var("GOOSE_ROUTER_TOOL_SELECTION_STRATEGY").is_ok() {
-        let _ = cliclack::log::info("Notice: GOOSE_ROUTER_TOOL_SELECTION_STRATEGY environment variable is set. Configuration will override this.");
+    // Check if GOOSE_ENABLE_ROUTER is set as an environment variable
```
DOsinga: can we just drop the environment variable here? as @michaelneale I'm not a fan, just keeping it to the settings would work well enough

crates/goose-cli/src/commands/configure.rs
```diff
@@ -1170,40 +1170,32 @@ pub fn configure_goose_mode_dialog() -> Result<(), Box<dyn Error>> {
...
+    if std::env::var("GOOSE_ENABLE_ROUTER").is_ok() {
+        let _ = cliclack::log::info("Notice: GOOSE_ENABLE_ROUTER environment variable is set. Configuration will override this.");
     }
 
-    let strategy = cliclack::select("Which router strategy would you like to use?")
+    let enable_router = cliclack::select("Would you like to enable LLM-based tool routing?")
```
DOsinga: replace LLM-based with smart

ui/desktop/src/components/settings/tool_selection_strategy/ToolSelectionStrategySection.tsx
```diff
@@ -10,25 +10,20 @@ interface ToolSelectionStrategy {
...
+    label: 'Enabled',
     description:
-      'Uses LLM to intelligently select the most relevant tools based on the user query context.',
+      'Use LLM-based intelligence to select the most relevant tools based on the user query context.',
   },
 ];
 
 export const ToolSelectionStrategySection = () => {
-  const [currentStrategy, setCurrentStrategy] = useState('default');
+  const [currentStrategy, setCurrentStrategy] = useState('false');
```
DOsinga: can we rename these variables? also should the type be boolean instead?

documentation/docs/guides/environment-variables.md
```diff
@@ -235,28 +235,17 @@ These variables configure the [tool selection strategy](/docs/guides/managing-to
...
-```
 
-**Embedding Provider Support**
-
-The default embedding provider is OpenAI. If using a different provider:
-- Ensure the provider supports embeddings
-- Specify an appropriate embedding model for that provider
-- Ensure the provider is properly configured with necessary credentials
+# Or use default selection
+export GOOSE_ROUTER_TOOL_SELECTION_STRATEGY=default
```
DOsinga: I thought we renamed these and made it into a boolean (either way, let's drop the environment variable)

crates/goose/src/agents/tool_route_manager.rs
```diff
@@ -68,21 +65,24 @@ impl ToolRouteManager {
...
     }
 
-    pub async fn get_router_tool_selection_strategy(&self) -> Option<RouterToolSelectionStrategy> {
+    pub async fn is_router_enabled(&self) -> bool {
         if *self.router_disabled_override.lock().await {
-            return None;
+            return false;
         }
 
+        // Check config first (takes precedence)
```
DOsinga: remove the comment (it is not actual the first thing we check)

crates/goose/src/agents/tool_route_manager.rs
```diff
@@ -142,27 +136,39 @@ impl ToolRouteManager {
...
-        strategy: Option<RouterToolSelectionStrategy>,
         extension_manager: &Arc<RwLock<ExtensionManager>>,
     ) -> Vec<Tool> {
+        // If router is disabled or overridden, return empty
         if *self.router_disabled_override.lock().await {
             return vec![];
         }
 
+        // If router is not enabled, return empty (will fall back to regular tools)
+        if !self.is_router_enabled().await {
```
DOsinga: Since this is now the central check, I dont think we need the one in 154

documentation/docs/guides/environment-variables.md
```diff
@@ -228,25 +228,6 @@ export GOOSE_EDITOR_HOST="http://localhost:8000/v1"
...
-**Examples**
-
-```bash
-# Use LLM-based selection
-export GOOSE_ROUTER_TOOL_SELECTION_STRATEGY=llm
-
-# Or use default selection
-export GOOSE_ROUTER_TOOL_SELECTION_STRATEGY=default
-```
-
```
wendytang, line 257: fyi @blackgirlbytes - alright if we make the change here? i can also omit if we want to do it as a follow up

michaelneale:
"Available only with Claude models served on Databricks for now." - it says in GUI - which I don't think is right?

I did try it with anthropic direct and it worked, but with gpt-5 and openai: 

image

so it may only effectively work with some models? what is the guidance, I would think if it can't work with gpt-5 then it isn't really a suitable strategy (although it may be valid to get rid of the vector based one if not used)

michaelneale (changes requested):
would like clarification if LLM selection strategy is really suitable for use - I'm not sure it is (or should be recommended, or when it is recommended - keep it as databricks only?)

michaelneale:
otherwise if it is claude only for now - that is ok (can probably tweak wording as it works with claude directly at least, just not gpt-5 when I tried, kept getting stuck in a loop)

michaelneale:
@DOsinga worthwhile getting it working with other models? or we can just change the wording to make it clear it may not work with all models? (payoff is less dependencies and complexity)

michaelneale:
ok I have tried for a few hours to make this work with gpt-5, and to no avail. I really don't follow the logic in `RouterToolSelector` and it is clear it only partially works, but then it isnt' a change from before. 

I am pro getting rid of the old code, so maybe this is ok, and leave that battle for another day?

crates/goose/src/agents/router_tool_selector.rs
```diff
@@ -338,7 +95,7 @@ impl RouterToolSelector for LLMToolSelector {
             let user_message = Message::user().with_text(&user_prompt);
             let response = self
                 .llm_provider
-                .complete("", &[user_message], &[])
+                .complete("system", &[user_message], &[])
```
michaelneale, line 98: ?? hrm... wonder if this is causing a problem

crates/goose/src/agents/tool_route_manager.rs
```diff
@@ -70,21 +67,18 @@ impl ToolRouteManager {
         }
     }
 
-    pub async fn get_router_tool_selection_strategy(&self) -> Option<RouterToolSelectionStrategy> {
+    pub async fn is_router_enabled(&self) -> bool {
```
michaelneale, line 70: really should be "is_tool_router_enabled" as router is an overloaded term

michaelneale:
hrm nope: 

```
─── llm_search | router ──────────────────────────
extension_name: developer
query: [REDACTED: 112 chars]
k: 5

### I can’t access your files yet

I don’t currently have a filesystem tool enabled, so I can’t list the directory contents from here. To let me do that, please add/enable a developer/filesystem extension (e.g., a tool that can run “list directory” or “ls/dir”). Once it’s enabled, I can run it for you.

In the meantime, you can run one of these locally and paste the output:

- macOS/Linux:
  - `pwd`
  - `ls -la`
- Windows (Command Prompt or PowerShell):
  - `cd`
  - `dir`
```

with gpt-5 - so yeah, really doesn't work that well. 

I agree with the cleanup but I don't think the router tool selection strategy is workable either.

wendytang:
@michaelneale ok if we check in the vector removal first? 

The tool router is workable with claude sonnet only - there is an action item - to ensure openai compatibility

if you're interested in taking this on please feel free to - i'm happy to but it'd be my weekend schedule. 

https://linear.app/squareup/issue/AAI-790/ensure-openai-compatibility

documentation/docs/guides/environment-variables.md
angiejones: this is fine, Wendy. thank you

angiejones:
good to merge!

documentation/docs/guides/environment-variables.md
blackgirlbytes: sorry i missed this! looks good

joeeuston-dev:
I'm using LLM Strategy with VertexAI -> Claude-Sonnet-4
No issues on 1.3.0.
Now I'm on 1.6.0 and I've noticed My agent doesn't know what subagents are and can't find them. Doing some troubleshooting led me back to this.
Changing my Tool Selection Strategy back to default has unblocked me, but there may be others who still unknowingly have their default tool selection set to LLM and are having issues

PR #3937, opened 2025-08-08
feat: streamline list args on cli

baxen:


michaelneale (approved):
much tidier, thank you @baxen !

PR #3939, opened 2025-08-08
MCP session replay integration test

jamadeo:
Lets you set up an MCP server (stdio only for now), some interactions (tool calling only for now) and record the whole thing to a session replay file. Then commit the replay and result files and the test will run the interaction through the extension manager.

This works by using two new binaries included in this PR: one that sits in front of a stdio server, writing all i/o to a replay file, and one that reads said replay file and mimics the behavior of the recorded session.


alexhancock (approved):
It LGTM! My only ideas for slight improvement to ergonomics

1. Maybe name the replayer something like "validator"?
1. I wonder if the binaries could easily be combined into a single module with `--mode record || validate` maybe it slightly easier to expand functionality in future

jamadeo:
Thanks @alexhancock good points -- made a single binary with a transport mode that will invite us to do the same for streamable http extensions.

Issue #3821, opened 2025-08-04
BUG: CLI Provider Setup failing on windows (CC and Gemini)


feurer98:
**Describe the bug**

<img width="1215" height="1186" alt="Image" src="https://github.com/user-attachments/assets/6f2ff9c3-49a5-4f04-8283-3b32c5a28084" />

**To Reproduce**
Fresh installation of goose and claude cli and gemini cli.
Both installations are ps1 references (C:\Users\xxxxxx\AppData\Roaming\npm\gemini.ps1).
After chatgpt, I switched PAHT to cmd but still the same.
Restarted the Windows PC in between.
Still getting the same error.

**Please provide following information:**
 - **WINDOWS Windows 11 Pro, 23H2, 22631.5624**
 - **Interface:** [CLI]
 - **Version:** [1.1.4]

Note: I switched to cli because I couldn't configure either of them via the UI. Ollama is running and works with goose.

If you need me to send something for logs, please let me know. I'm happy to help.

DOsinga:
I would say that it fails to find claude itself. let me have a look

DOsinga:
yup, I can reproduce this

feurer98:
Still erros with 1.2.0

```
XXXXX@Z840 MINGW64 ~/Documents/GitHub/subabasetest (main)
$ npm install -g @anthropic-ai/claude-code

added 10 packages, and changed 2 packages in 4s

11 packages are looking for funding
  run `npm fund` for details

XXXXX@Z840 MINGW64 ~/Documents/GitHub/subabasetest (main)
$ claude
╭──────────────────────────────────────────────────────╮
│ ✻ Welcome to Claude Code!                            │
│                                                      │
│   /help for help, /status for your current setup     │
│                                                      │
│   cwd: C:\Users\XXXXX\Documents\GitHub\subabasetest  │
╰──────────────────────────────────────────────────────╯


 What's new:
  • Performance: Optimized message rendering for better performance with large contexts
  • Windows: Fixed native file search, ripgrep, and subagent functionality
  • Added support for @-mentions in slash command arguments
  • Upgraded Opus to version 4.1
  • Fix incorrect model names being used for certain commands like `/pr-comments`


XXXXX@Z840 MINGW64 ~/Documents/GitHub/subabasetest (main)
$ goose configure

This will update your existing config file
  if you prefer, you can edit it directly at C:\Users\XXXXX\AppData\Roaming\Block\goose\config\config.yaml

T   goose-configure
|
[...]

PR #3940, opened 2025-08-08
fix: adding windows specific path search for claude code

michaelneale:
attempting fix for https://github.com/block/goose/issues/3821


michaelneale:
cc @DOsinga not sure if you have windows handy (could do similar for gemini)

michaelneale:
closing this as didn't work

PR #3951, opened 2025-08-08
fix: disable signing for release branches until we figure out keys for this flow

alexhancock:
This will give us unsigned pre-release builds for now (given current signing doesn't work https://github.com/block/goose/actions/runs/16833234629/job/47686043767?pr=3921) that can be run if the appropriate steps are taken for testing.

We will come up with a better signing system soon to enable:

1. Official releases
2. Pre-releases from release branches
3. Nightly builds from main

to all be signed

### DIFF
--- a/.github/workflows/release-branches.yml
+++ b/.github/workflows/release-branches.yml
@@ -10,8 +10,6 @@ jobs:
     permissions:
       id-token: write
       contents: read
-    with:
-      signing: true
 
   comment-on-pr:
     needs: bundle-desktop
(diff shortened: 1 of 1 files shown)


.github/workflows/release-branches.yml
```diff
@@ -10,8 +10,6 @@ jobs:
     permissions:
       id-token: write
       contents: read
-    with:
-      signing: true
```
jamadeo, line 14: small nit but the PR comment below does say "this app is signed" which won't be true of course. We could stick the xattr command instruction there

PR #3952, opened 2025-08-08
fix: enable 'goose://' handler for debian systems

eyelight:
This PR fixes a bug in linux (debian) where `goose://` URLs would not know which application to open.

After making the changes in the commit, the `.deb` resulting from `npm run make -- --targets=@electron-forge/maker-deb` contains the required line in the `goose.desktop` file.

When the deb is installed using `dpkg`, the installed Goose application is successfully invoked upon clicking a `goose://` web link from within Chromium on Pop!_OS 22.04.


BUILDING_LINUX.md
```diff
@@ -145,12 +145,13 @@ Create `~/.local/share/applications/goose.desktop`:
 [Desktop Entry]
 Name=Goose AI Agent
 Comment=Local AI agent for development tasks
-Exec=/path/to/goose/ui/desktop/out/Goose-linux-x64/Goose
+Exec=/path/to/goose/ui/desktop/out/Goose-linux-x64/Goose %U
```
michaelneale, line 148: is this deliberate?

eyelight: Yes, the %U becomes deeplink arguments etc when invoked that way, but it is ignored otherwise.

Issue #3899, opened 2025-08-06
GooseUI new chats are frequently going to home tab and silently still running in the background


simonsickle:
**Describe the bug**

While running a task, I notice sometimes my chat will blank out for a second and go home. After it goes home, I will go to history and see my chat is listed there - I hit the chat tab and goose apparently was working in the background this whole time still.

**To Reproduce** (sporadic but frequent)
1. Start a new chat
2. Paste a long prompt in
3. Go to slack / google chrome to continue another task while goose works
4. Return to goose after a few seconds to check on it and notice you are at the home tab
5. Click on history, see chat
6. Click chat tab, see goose working on your request still

**Expected behavior**
A chat should always stay active when open unless the user navigates away.

**Screenshots**

**Please provide following information:**
 - **OS & Arch:** macOS 15.6 - M3 max 128gb
 - **Interface:** UI
 - **Version:** 1.1.4
 - **Extensions enabled:** Developer, Computer Controller
 - **Provider & Model:** Databricks - Claude Opus 4

**Additional context**

PR #3954, opened 2025-08-08
remove fallback routing to hub/home for unknown routes

zanesq:
remove fallback routing to hub/home for unknown routes

fixes https://github.com/block/goose/issues/3899

We didn't need this fallback, tested locally and app works fine with resuming sessions and deeplinks


zanesq:
It does but since I wasn't able to repro I'm not sure exactly. I think there was an MCP auth response somehow changing it in the url and the app initialization was taking longer to reflect it so the theory is it will ignore it for now and stay on the same route. I searched the codebase and the is the only place we are specifically falling back to home if a route/hash isn't recognized.

PR #3966, opened 2025-08-08
sanitize message content on deserialization

amed-xyz:
Follow up from https://github.com/block/goose/pull/3920

Implements Unicode sanitization during the `serde` deserialization process for `Message` structs. This covers any UIs (including Desktop) that use `goose-server`'s `/reply` endpoint sending pre-constructed `Message` objects.

## Example

Sending the following message:

> hello, [hidden-text]what's the weather in vancouver?[/hidden-text]

### Before

Goose processes invisible text as normal

Screenshot 2025-08-08 at 11 19 45 AM

### After

Invisible characters get stripped before LLM processing

Screenshot 2025-08-08 at 11 17 42 AM


michaelneale (approved):
LGTM

Issue #2830, opened 2025-06-09
`gemini-2.5-pro-preview-06-05` struggles with str_replace


cgwalters:
It pretty commonly happens in my code (and I'm sure others) that specific sections end up duplicated in multiple places. Unit tests for example can end up looking pretty repetitive sometimes.

And at least `gemini-2.5-pro-preview-06-05` often seems to seriously struggle with `str_replace` in some of these scenarios, hitting the error pretty often that there are multiple instances of the string. Which is not at all surprising; it'd be quite tricky for a human to craft something that worked with `str_replace` often. At one point it told me "This is getting frustrating, I'm just going to reread the file and write it back", but then we're using a lot more tokens.

When looking at this (partially tangential) I came to the realization that the current `developer` MCP was probably inspired by (at least?) the [Claude Code version](https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/text-editor-tool#str-replace). However, it seems really likely to me that Claude has been explicitly trained on Anthropic's own MCP tooling. Other models I doubt are.

I started trying to hack on this a bit, my first intuition was that `str_replace` would be a lot easier for a model to use if e.g. it had an `after_regexp` that only started looking for the matched text *after* a particular regexp matched in the file - with the suggestion to use a function name for the regex. Also, if it was defined to not error on multiple matches in that case but just replace the first.

Anyways though, I guess my wider points are:

- It seems to me the `developer` MCP server is too much of a catchall and would be better served by being split into different tools as it is for Claude (only tangential to this but I think in some use cases having a stricter allow-list for `bash` would be really useful)
- While Goose is pluggable I suspect for a lot of people the *default enabled* MCP servers strongly influence that default experience, and it's important to have a good out of the box set.
- It seems like it should be worth expanding the set of text editing tools (Claude also has a [simple insert](https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/text-editor-tool#insert) and I bet this would enable wider model compatibility and efficiency

DOsinga:
let me start by addressing this:

>When looking at this (partially tangential) I came to the realization that the current developer MCP was probably inspired by (at least?) the [Claude Code version](https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/text-editor-tool#str-replace).

we were first! if there's been any inspiring going on, surely it has gone the other way!

[...]

PR #1515, opened 2025-03-05
feat: add image tool to developer mcp

zakiali:
Adds an image processing tool to the developer MCP. If you give goose an image path, the tool will read it in and process it (similar to the screen capture tool). 

Some limits:
- 10mb is the limit on the file size (this was arbitrary)
- image gets resized similar to screen cap's to fit with provider limitations
- works best when there are no spaces in the file name (I found that models have trouble escaping correctly)

PR #3976, opened 2025-08-08
mcp: Move screen/window tools to computercontroller module

cgwalters:
Move list_windows and screen_capture tools from the developer MCP to computercontroller. This is about making the developer MCP more focused.

xref https://github.com/block/goose/issues/2830

Assisted-By: Claude Code

### DIFF
--- a/crates/goose-mcp/src/developer/mod.rs
+++ b/crates/goose-mcp/src/developer/mod.rs
@@ -44,7 +44,6 @@ use self::shell::{expand_path, get_shell_config, is_absolute_path, normalize_lin
 use indoc::indoc;
 use std::process::Stdio;
 use std::sync::{Arc, Mutex};
-use xcap::{Monitor, Window};
 
 use ignore::gitignore::{Gitignore, GitignoreBuilder};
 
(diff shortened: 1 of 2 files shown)


crates/goose-mcp/src/developer/mod.rs
```diff
@@ -1506,8 +1350,6 @@ impl Router for DeveloperRouter {
             match tool_name.as_str() {
                 "shell" => this.bash(arguments, notifier).await,
                 "text_editor" => this.text_editor(arguments).await,
-                "list_windows" => this.list_windows(arguments).await,
-                "screen_capture" => this.screen_capture(arguments).await,
                 "image_processor" => this.image_processor(arguments).await,
```
DOsinga, line 1353: I wonder about this one. it seems rather related to screen_capture - refers to it, but I am also not sure about the logic in it. it converts images to PNG and resizes them even though my understanding is LLMs dont really care about screensizes anymore. so leave, move with or delete? maybe leave and improve (adding cropping would be useful or possibly just tell goose about imagemagick

DOsinga (approved):

crates/goose-mcp/src/developer/mod.rs
cgwalters: Yes, I too wondered about this one. Probably best to ping in https://github.com/block/goose/pull/1515 ?

> or possibly just tell goose about imagemagick

Yeah there's a ton of other tools available in this space.

Maybe in the short term we can just move this one to computercontroller too?

DOsinga: Let's do it. I had been thinking about scanning for imagemagick, ffmpeg maybe pandoc, what have you and then say, hey these tools are installed go wild

cgwalters: I took a look at this but it's a lot messier because the `image_processor` code uses a lot of the "handle `.gooseignore`" logic that is embedded in the developer MCP code.

It probably makes sense to factor that out into a sharable module, but it's more invasive. So let's get this one in as a preparatory step?

DOsinga:
kicked of the checks so yeah, let's go for it

angiejones:
are you all not worried about backwards compatibility? there may be existing recipes scheduled and running that only have Developer extension enabled and rely on these tools

cgwalters:
@angiejones agree it's a concern. I guess honestly we could get 75% of the value here by just splitting up developer/mod.rs without actually moving where tools are. So I'll look at that instead.

PR #3982, opened 2025-08-09
experiment: prewarm caches

michaelneale:
this is an experment. Send a message to prewarm per session, and also add in a cache key for the openai provider message format

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -97,6 +97,7 @@ pub struct Agent {
     pub(super) tool_route_manager: ToolRouteManager,
     pub(super) scheduler_service: Mutex<Option<Arc<dyn SchedulerTrait>>>,
     pub(super) retry_manager: RetryManager,
+    prewarmed_sessions: Mutex<HashSet<String>>,
 }
 
 #[derive(Clone, Debug)]
@@ -173,6 +174,7 @@ impl Agent {
             tool_route_manager: ToolRouteManager::new(),
             scheduler_service: Mutex::new(None),
             retry_manager,
+            prewarmed_sessions: Mutex::new(HashSet::new()),
         }
     }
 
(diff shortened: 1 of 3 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -191,6 +193,57 @@ impl Agent {
         self.retry_manager.increment_attempts().await
     }
 
+    /// Pre-warm the LLM/provider by sending a lightweight background request with
```
cgwalters, line 196: Ah...I often see quite nontrivial initial latency with goose+gemini, I will try this out!

I was looking at this https://ai.google.dev/gemini-api/docs/caching - I wonder if other tools are using explicit caches for things like the system prompt.

crates/goose/src/agents/agent.rs
```diff
@@ -814,6 +867,41 @@ impl Agent {
         session: Option<SessionConfig>,
         cancel_token: Option<CancellationToken>,
     ) -> Result<BoxStream<'_, Result<AgentEvent>>> {
+        // Background prewarm when a new session starts (server/desktop). Always on.
```
cgwalters, line 870: Seems to me we could deduplicate the code here vs the explicit `prewarm` above

crates/goose/src/providers/formats/openai.rs
```diff
@@ -14,6 +14,15 @@ use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
 use std::ops::Deref;
 
+use std::sync::OnceLock;
+use uuid::Uuid;
+
+static PROMPT_CACHE_KEY: OnceLock<String> = OnceLock::new();
+
+fn get_prompt_cache_key() -> &'static str {
```
cgwalters, line 22:
```suggestion
/// See <https://platform.openai.com/docs/guides/prompt-caching>
fn get_prompt_cache_key() -> &'static str {
```

And on that topic, the docs say:

> Use the
[prompt_cache_key](https://platform.openai.com/docs/api-reference/responses/create#responses-create-prompt_cache_key)
parameter consistently across requests that share common prefixes. Select a granularity that keeps each unique prefix-prompt_cache_key combination below 15 requests per minute to avoid cache overflow.

I'm trying to understand the rationale behind having a random cache key for the provider allocated once. Is it the case today that in a given goose instance, we can expect all requests to share that common system prompt (i.e. we don't ever have multiple independent different prompts per process?) With the goose subagent stuff won't that become possible?

Shouldn't this prompt cache actually be driven by higher level calling code?

cgwalters: I know this is just a draft, but this change to the openai provider is not *directly* related to prewarming and IMO should be split out into its own PR with its own clear rationale.

crates/goose/src/agents/agent.rs
michaelneale: @cgwalters yeah - may need per provider supplements to this (for optimisation). If you find even this helps in your experience, LMK - this is just an experiment and not sure I will keep it, so would love to hear if you can see/measure anything

crates/goose/src/providers/formats/openai.rs
michaelneale: yeah - it was more an attempt to see if it would make it any faster at all, but agree, shoudln't be as part of this @cgwalters 👍

michaelneale:
closing this as was experiment

jmservier:
Hi didn't check the goose code base yet, regarding the cost of our agentic devlopment usage I would be interested by `prompt_cache_key` feature.
source: https://learn.microsoft.com/hi-in/azure/foundry/openai/how-to/prompt-caching

```json
{
  "model": "gpt-5.6-sol",
  "prompt_cache_key": "goose-project-myapp",
  "messages": [
    {
      "role": "system",
      "content": "Instructions read file xxx..."
    },
    {
      "role": "user",
      "content": "Question variable..."
    }
  ]
}
```

PR #3986, opened 2025-08-09
Adding better deeplink support for streamable-http

### LINKED ISSUE #3438: Desktop deeplink parser for streamable_http extensions: Install link doesn't work and ignores header parameters

Recent testing indicates there might be more issues with install links for streamable_http extensions -- not just handling the `header` parameters as originally reported. For example, this link appears to add the extension to the config as expected:

goose://extension?url=https%3A%2F%2Fmcp.neon.tech%2Fmcp&transport=streamable_http&id=neon&name=Neon&description=Manage%20Neon%20Postgres%20databases%2C%20projects%2C%20and%20branches

The `transport` arg is required to register it as `streamable_http` but should we also accept `type=streamable_http` to be consistent with the config `type` field?

Also, Goose says the Desktop app's deeplink parser fails to properly parse and apply custom headers for `streamable_http` extensions. While the `transport` parameter works, `header` parameters in the install link are ignored, causing extensions that require headers (e.g. for authentication) to fail initialization. This forces users to manually configure these extensions.

- Deeplinks with `type=streamable_http` should register the extension as the http type (currently only works with `transport` and defaults to `sse`, see /ui/desktop/src/components/settings/extensions/deeplink.ts)
- Deeplinks for `http` types with `header=name%3Dvalue` parameters should populate the headers object in the extension config  (I think populating the specified value is a stretch considering `env` arg behavior)
- Extensions should initialize successfully when required headers are provided via deeplink

Desktop users can add custom extension for streamable servers using manual configuration steps. 

**Additional context**
Goose says:

### Root Cause Analysis Needed
The deeplink parser in ui/desktop/src/components/settings/extensions/deeplink.ts correctly handles transport but fails to process header parameters. Investigation needed for:

- How multiple header parameters are extracted from the URL
- How header values are URL-decoded (e.g., %3D → =)
- How headers are applied to the streamable_http extension config
- Whether the getStreamableHttpConfig function properly accepts headers

### Acceptance Criteria
- [ ]  Fix deeplink parsing to extract and apply header=name%3Dvalue parameters
- [ ]  Extensions created from deeplinks should have correct headers in config
- [ ]  Extensions requiring headers (like GitHub MCP) should initialize successfully
- [ ]  Add/improve tests for deeplink parsing with headers

- [x] I have verified this does not duplicate an existing feature request -- though this is related to issue #2423

blackgirlbytes:
### Problem
Deeplinks for streamable HTTP extensions had two issues:
1. Only recognized `transport=streamable_http`, ignored `type=streamable_http` (some organizations like Kiwi Flight search have the goose URI
2. No support for authentication headers

This caused extensions like Kiwi.com to be registered as SSE instead of streamable_http, and prevented GitHub-style extensions requiring auth headers from working via deeplinks.

### Solution
Enhanced the deeplink parser to:
- Support both `transport` and `type` parameters
- Parse `header=key%3Dvalue` parameters for streamable HTTP extensions
- Maintain backward compatibility

### Why `type` Parameter Support?

The deeplink parser only recognized `transport=streamable_http` but the actual config schema uses `type: 'streamable_http'`. This inconsistency led extension developers (like Kiwi.com) to naturally use `type=streamable_http` in their deeplinks, expecting it to match the config field name.

The fix maintains backward compatibility with `transport` while adding support for the more intuitive `type` parameter that aligns with the config schema.

### Examples Now Working
```bash
# Kiwi.com (now correctly registered as streamable_http)
goose://extension?url=https%3A%2F%2Fmcp.kiwi.com&name=Kiwi&type=streamable_http

# GitHub with auth headers
goose://extension?url=https%3A%2F%2Fapi.githubcopilot.com%2Fmcp%2F&name=GitHub&type=streamable_http&header=Authorization%3DBearer%20token
```

Fixes #3438

### DIFF
--- a/ui/desktop/src/components/settings/extensions/deeplink.ts
+++ b/ui/desktop/src/components/settings/extensions/deeplink.ts
@@ -79,7 +79,8 @@ function getStreamableHttpConfig(
   remoteUrl: string,
   name: string,
   description: string,
-  timeout: number
+  timeout: number,
+  headers?: { [key: string]: string }
 ) {
   const config: ExtensionConfig = {
     name,
@@ -87,6 +88,7 @@ function getStreamableHttpConfig(
     uri: remoteUrl,
     description,
     timeout: timeout,
+    headers: headers,
   };
 
   return config;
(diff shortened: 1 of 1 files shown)


ui/desktop/src/components/settings/extensions/deeplink.ts
```diff
@@ -143,11 +145,25 @@ export async function addExtensionFromDeepLink(
 
   const cmd = parsedUrl.searchParams.get('cmd');
   const remoteUrl = parsedUrl.searchParams.get('url');
-  const transportType = parsedUrl.searchParams.get('transport') || 'sse'; // Default to SSE for backward compatibility
+  // Support both 'transport' and 'type' parameters for consistency
+  const transportType =
+    parsedUrl.searchParams.get('transport') || parsedUrl.searchParams.get('type') || 'sse'; // Default to SSE for backward compatibility
+
+  // Extract headers from URL parameters for streamable_http extensions
```
angiejones:
```suggestion

```

PR #3998, opened 2025-08-11
chore: add Z.AI provider for GLM-4.5 Serials

Ed4ward:
## Context

Added Z.AI & BigModel providers for GLM-4.5 Serials

## Screenshots
image

## How to Test
Request an API key from [Z.ai](https://z.ai/manage-apikey/apikey-list) or [BigModel](https://bigmodel.cn/usercenter/proj-mgmt/apikeys) and use it.

> Besides the main changes in this PR, I noticed the project is using npm. In some cases npm install may hang or be slower than expected. You might consider switching to pnpm for faster installs, lower disk usage, and more efficient dependency resolution.
> 
> This is just an optional suggestion — no changes required here.


michaelneale:
thanks @Ed4ward - is z.ai a openai compatible style of LLM endpoint?

Ed4ward:
> thanks @Ed4ward - is z.ai a openai compatible style of LLM endpoint?

yes it is.

Ed4ward:
@michaelneale  Sorry, I took a bit of time to fix some details. Could someone help review or move this PR forward?

michaelneale:
@Ed4ward we will soon have an easier way vs adding a whole bespoke provider cc @DOsinga so maybe we should hold off this until that is ready?

Ed4ward:
> @Ed4ward we will soon have an easier way vs adding a whole bespoke provider cc @DOsinga so maybe we should hold off this until that is ready?

sure, sounds great, if possible, keep me posted please.

crates/goose/src/providers/zai.rs
```diff
@@ -0,0 +1,226 @@
+use anyhow::Result;
```
DOsinga, line 1: Sorry for the very slow reply. We have custom providers now. Instead of defining this provider here, can we do this by instantiating a customprovider based on the openai schem?

Ed4ward: Of course, but I won't have time to do this until next week.I'll mention you then. 🙏

PR #3999, opened 2025-08-11
fix: need to use secret with add_sub_recipe_on_agent

### LINKED ISSUE #3977: Sub-recipes fail to load in Desktop app - missing X-Secret-Key header
**Describe the bug**

Community members tried to run subrecipes in Desktop with open router as a provider, and they got this error. 

```
TypeError: Failed to fetch
    at g (file:///Applications/Goose.app/Contents/Resources/app.asar/.vite/renderer/main_window/assets/index-CNiCQXew.js:41:63306)
    at Object.post (file:///Applications/Goose.app/Contents/Resources/app.asar/.vite/renderer/main_window/assets/index-CNiCQXew.js:41:64184)
    at ph (file:///Applications/Goose.app/Contents/Resources/app.asar/.vite/renderer/main_window/assets/index-CNiCQXew.js:41:64335)
    at rD (file:///Applications/Goose.app/Contents/Resources/app.asar/.vite/renderer/main_window/assets/App-BrkrcV_F.js:293:50305)
    at $p (file:///Applications/Goose.app/Contents/Resources/app.asar/.vite/renderer/main_window/assets/App-BrkrcV_F.js:324:353)
    at async file:///Applications/Goose.app/Contents/Resources/app.asar/.vite/renderer/main_window/assets/App-BrkrcV_F.js:2007:6647
```

I did some research via goose..and this may be the reason and fix

I actually asked goose about this earlier, but forgot to press send. this seems like plausible reason and solution..

```
## Summary

The Desktop app was failing to load sub-recipes because the `addSubRecipesToAgent` function was missing the required `X-Secret-Key` authentication header when making the API call, causing a "Failed to fetch" error.

## Issue Location

**File:** `/ui/desktop/src/recipe/add_sub_recipe_on_agent.ts`
**Line:** 4 (the `addSubRecipes` API call)

The function was calling:

const add_sub_recipe_response = await addSubRecipes({ body: { sub_recipes: subRecipes } });


But needed to include the authentication header like all other API calls in the desktop app.

## Fix

Added the missing `X-Secret-Key` header to authenticate the API request, matching the pattern used by all other desktop API calls.
```

michaelneale:
fixes: https://github.com/block/goose/issues/3977

### DIFF
--- a/ui/desktop/src/recipe/add_sub_recipe_on_agent.ts
+++ b/ui/desktop/src/recipe/add_sub_recipe_on_agent.ts
@@ -1,7 +1,12 @@
 import { addSubRecipes, SubRecipe } from '../api';
 
 export async function addSubRecipesToAgent(subRecipes: SubRecipe[]) {
-  const add_sub_recipe_response = await addSubRecipes({ body: { sub_recipes: subRecipes } });
+  const add_sub_recipe_response = await addSubRecipes({
+    body: { sub_recipes: subRecipes },
+    headers: {
+      'X-Secret-Key': await window.electron.getSecretKey(),
+    },
+  });
   if (add_sub_recipe_response.error) {
     console.warn(`Failed to add sub recipes: ${add_sub_recipe_response.error}`);
   } else {


michaelneale:
@blackgirlbytes you able to try this one out at all?

DOsinga (changes requested):
Yeah, this is not the way. this ends up calling addSubRecipes in the generated code which should be handling the security for you.

Issue #3724, opened 2025-07-29
traverse to repo root looking for .goosehints


khimaros:
[x] I have verified this does not duplicate an existing feature request

this is similar to the functionality provided by claude code.

perhaps it could be controlled by a config value and disabled by default for backward compat?

DOsinga:
would you want this to happen only in repos?

khimaros:
yes, only if inside agit repo. https://agent.md seems to define a reasonable approach for subdirectories

DOsinga:
I think this is implemented now.

PR #4002, opened 2025-08-11
Enable load nested hint files

lifeizhou-ap:
Implement the feature for https://github.com/block/goose/issues/3724

**What**
- Added the configuration for users to choose whether to load the nested goose hint file
- When configuration is enabled, it will load the hints file up to the directory with .git dir or current directory if no .git found

**Next (not in this PR)**
It would be good to show loaded hints in CLI and Desktop.


crates/goose-mcp/src/developer/mod.rs
```diff
@@ -407,55 +408,9 @@ impl DeveloperRouter {
         let hints_filenames: Vec<String> = std::env::var("CONTEXT_FILE_NAMES")
             .ok()
             .and_then(|s| serde_json::from_str(&s).ok())
-            .unwrap_or_else(|| vec![".goosehints".to_string()]);
```
lifeizhou-ap: Extracted the hints logic to `goose_hints.rs`

crates/goose-cli/src/commands/configure_settings.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::error::Error;
+
+pub fn configure_nested_hints_dialog() -> Result<(), Box<dyn Error>> {
+    let config = Config::global();
+
+    if std::env::var("GOOSE_NESTED_HINTS").is_ok() {
+        let _ = cliclack::log::info("Notice: GOOSE_NESTED_HINTS environment variable is set and will override the configuration here.");
+    }
+
+    let current_enabled: bool = config.get_param("NESTED_GOOSE_HINTS").unwrap_or(false);
```
alexhancock: I actually think enabled could be a better default here, and people would only need to seek out the config option if they didn't want it for some reason.

crates/goose-mcp/src/developer/goose_hints/import_files.rs
```diff
@@ -0,0 +1,488 @@
...
+    collections::HashSet,
+    path::{Path, PathBuf},
+};
+
+static FILE_REFERENCE_REGEX: Lazy<regex::Regex> = Lazy::new(|| {
+    regex::Regex::new(r"(?:^|\s)@([a-zA-Z0-9_\-./]+(?:\.[a-zA-Z0-9]+)+|[A-Z][a-zA-Z0-9_\-]*|[a-zA-Z0-9_\-./]*[./][a-zA-Z0-9_\-./]*)")
+        .expect("Invalid file reference regex pattern")
+});
+
+const MAX_DEPTH: usize = 3;
```
alexhancock, line 13: Why three? I'd think there would be realistic repo scenarios of wanting four or five perhaps

crates/goose/src/agents/extension_manager.rs
```diff
@@ -329,6 +329,14 @@ impl ExtensionManager {
 
                 let transport = TokioChildProcess::new(Command::new(cmd).configure(|command| {
                     command.arg("mcp").arg(name);
+
+                    // pass configuration to the developer extension
+                    if name == "developer" {
+                        let config = crate::config::Config::global();
+                        if let Ok(nested_enabled) = config.get_param::<bool>("NESTED_GOOSE_HINTS") {
+                            command.env("NESTED_GOOSE_HINTS", nested_enabled.to_string());
```
alexhancock: do we not already forward env vars to mcp server subprocesses?

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -40,6 +40,8 @@ use rmcp::model::{
 };
 use rmcp::object;
 
+use crate::developer::goose_hints::load_hints::{load_hint_files, GOOSE_HINTS_FILENAME};
```
alexhancock, line 43: Good to see `developer` get smaller and this to be moved out!

I do wonder if we will need tighter integration again soon though. It could be nice to have it discover `.goosehints` in subdirs when working with files in those subdirs. And feels like the depth limit could be more relaxed for this case...

alexhancock:
Tried it locally and asked:

"What .goosehints do you already know about?" from cwd of `goose/ui/desktop`

it picked up the one from `goose/ui/desktop/.goosehints` but not `goose/.goosehints`

I wonder if this should work, or indicates an issue with the upward traversal

michaelneale:
I think consider default on for this, if it isn't ready for that, then it needs more work IMO. This would use the nearest hints - is that right? or does it keep loading them all the way up?

alexhancock:
Yeah I agree with @michaelneale above! I think this would be good just "default to on" behavior.

I also don't know that it needs to be configureable tbh. The more options we add to settings the more conceptual overhead there is for the user. People could configure it by modifying/adjusting/removing the content of the goosehints files it finds and uses.

crates/goose-mcp/src/developer/goose_hints/import_files.rs
lifeizhou-ap: This is copied from existing behaviour during refactoring.  

This max depth is more about reference depth.  For example,

file 1
@file2.md

file2
@file3.md

file3
@file4.md

We can increase if we would like to have more chain of references

crates/goose/src/agents/extension_manager.rs
lifeizhou-ap: Here we get the NESTED_GOOSE_HINTS value from config, and pass it to the mcp as environment variable as it is not good for mcp server to know the config. 

We could simplify this without make this behaviour configuration as suggested in the comments below

crates/goose-mcp/src/developer/mod.rs
lifeizhou-ap: I feel this is a slightly different requirements 

In the implementation of this PR, while working on the current directly, it inherits all the .goosehints from the parent directory up to the project root.

>  It could be nice to have it discover .goosehints in subdirs when working with files in those subdirs. And feels like the depth limit could be more relaxed for this case...

This would be also useful.  This also aligns the  glob features in cursor, and Claude code has something similar.  We can work on this separately

lifeizhou-ap:
> I think consider default on for this, if it isn't ready for that, then it needs more work IMO. 

👍 

> This would use the nearest hints - is that right? or does it keep loading them all the way up?

It will keep loading them all the way up until they see the directory with .git (eg project root). Otherwise, it uses the `.goosehints` in the current directory

lifeizhou-ap:
> Yeah I agree with @michaelneale above! I think this would be good just "default to on" behavior.
> 
> I also don't know that it needs to be configureable tbh. The more options we add to settings the more conceptual overhead there is for the user. People could configure it by modifying/adjusting/removing the content of the goosehints files it finds and uses.

👍 This totally makes sense!

lifeizhou-ap:
> Tried it locally and asked:
> 
> "What .goosehints do you already know about?" from cwd of `goose/ui/desktop`
> 
> it picked up the one from `goose/ui/desktop/.goosehints` but not `goose/.goosehints`
> 
> I wonder if this should work, or indicates an issue with the upward traversal

Hi @alexhancock Thank you so much for testing this! Really appreciate it!

I tried on local, it seems Goose picked up the `goose/.goosehints`.  I guess it might need to run the developer mcp to load the hints. (this is the existing behaviour). 

Screenshot 2025-08-22 at 12 21 21 pm

lifeizhou-ap:
Hi @alexhancock @michaelneale,

Thank you so much for your time to review this PR. I've replied to your comments and removed the NESTED_GOOSE_HINTS configuration as suggested. 

I added a bit clarification on MAX_DEPTH,  happy to increase if you think 4/5 depth will be helpful. 

Thank you!

crates/goose-mcp/src/developer/goose_hints/import_files.rs
alexhancock: oh i see

PR #4007, opened 2025-08-11
feat(mcp): Persist OAuth credentials to keyring

### LINKED ISSUE #3545: OAuth flow repeated for every chat
For a Streamable HTTP MCP server with Oauth, goose runs the user through the entire Oauth flow every time they start a new chat session.

Example:

1. Install streamable http server with oauth such as https://docs.onkernel.com/mcp
2. Complete oauth flow
3. start chat session → this puts you through oauth flow again!

Even get this error when attempting to re-auth:
> Error
> Authentication already completed.

jamadeo:
When doing the oauth dance with rmcp, save the credentials to the keyring so you don't repeat this every time.

fixes #3545


crates/goose/src/oauth/persist.rs
```diff
@@ -0,0 +1,72 @@
...
+    let (client_id, token_response) = oauth_state.get_credentials().await?;
+
+    let credentials = SerializableCredentials {
+        client_id,
+        token_response,
+    };
+
+    let value = serde_json::to_value(&credentials)?;
+    let key = secret_key(name);
+    config.set_secret(&key, value)?;
```
alexhancock, line 32: how long are the credentials typically valid? maybe I am missing it but shouldn't there be a ttl for the data we store via keyring?

jamadeo: Yeah they'll have an expiry, but that should be handled by the oauth mechanism anyway -- it might need to be refreshed after getting loaded but that should be fine

jamadeo: Tested it out with an expired credential and indeed there was a bug! Added a call to refresh on load.

PR #4010, opened 2025-08-11
Remove cognitive complexity clippy lint

jamadeo:
from https://rust-lang.github.io/rust-clippy/master/index.html#cognitive_complexity

> What it does
We used to think it measured how hard a method is to understand.

> Why is this bad?
Ideally, we would like to be able to measure how hard a function is to understand given its context (what we call its Cognitive Complexity). But that’s not what this lint does. See “Known problems”

> Known problems
The true Cognitive Complexity of a method is not something we can calculate using modern technology. This lint has been left in the nursery so as to not mislead users into using this lint as a measurement tool.

Open to opinions on this but the fact that clippy themselves feel it's a bad measure doesn't seem great. I came to this because a violation for it here: https://github.com/block/goose/blob/a64a81672aa99a77bcf73f6494f52d26e50e8f15/crates/goose/src/oauth/mod.rs#L33

When I looked into splitting this up, most options truly seemed to make it more complicated, not less.


jamadeo:
thoughts @lifeizhou-ap @DOsinga ?

lifeizhou-ap (approved):

lifeizhou-ap:
> thoughts @lifeizhou-ap @DOsinga ?

Hi @jamadeo thanks for the doc reference! That makes sense.

PR #4019, opened 2025-08-12
fix(cli): fix compiling issue on windows system by adding missing params

GaryZhous:
On windows machine the code is not able to compile due to a missing param to the `call_tool` function at the unit test session in  `crates/goose-mcp/src/developer/mod.rs`, designed for windows under the conditional configuration check `#[cfg(windows)] (which is also the reason why the code can compile for linux machines). This pull request introduces a minor fix to the test code by adding the missing `dummy_sender()` param to the `call_tool` function.

## before
image

## after
image

### DIFF
--- a/crates/goose-mcp/src/developer/mod.rs
+++ b/crates/goose-mcp/src/developer/mod.rs
@@ -1916,6 +1916,7 @@ mod tests {
                 json!({
                     "command": "Get-ChildItem"
                 }),
+                dummy_sender(),
             )
             .await;
         assert!(result.is_ok());


michaelneale:
I am totally confused - how come it builds in CI on a windows machine?

GaryZhous:
> I am totally confused - how come it builds in CI on a windows machine?

@michaelneale here's the context https://github.com/block/goose/actions/runs/15350197323, it only ran the CI build for macOS when the `call_tool` function was last updated in this PR: https://github.com/block/goose/pull/2613

jamadeo (approved):

jamadeo:
Thanks @GaryZhous good catch.

@michaelneale we build on Windows for the app but I don't think we ever run any unit tests on windows in the CI process (which makes one wonder how this test was running to begin with...)

PR #4020, opened 2025-08-12
Recipe config to limit tool visibility

jsibbison-square:
When building recipes we usually have in mind a specific set of tools that we expect the agent to use to achieve the recipes goal. However many mcp extensions have a lot of irrelevant tools for our purpose. For example if we want to find and read files the developer extension is useful but we have no need of image processing, listing windows and screen capture tools.

Less tools are better for agents so having a way to filter out tools from mcps saves us from having to re-write specific mcps for the recipe.

This PR introduces tool config with visibility being the specific feature:
```yaml
extensions:
  - type: builtin
    name: developer
    tools_are_visible_default: true | false (default: true)
    tools:
      bash:
        visible: true
      screen_capture:
        visible: false
      ... the rest of the unspecified tools will inherit 'tools_are_visible_default' default value
```

When a tool as set as not visible then it will be entirely hidden from goose (not returned from `get_prefixed_tools`).

The 'tools' object is designed to be extendable so in the future we can add other keys like 'permissions' similar to [claude code settings](https://docs.anthropic.com/en/docs/claude-code/settings#settings-files) or [amp permissions](https://ampcode.com/manual#permissions).

### DIFF
--- a/crates/goose-cli/src/commands/configure.rs
+++ b/crates/goose-cli/src/commands/configure.rs
@@ -113,6 +113,8 @@ pub async fn handle_configure() -> Result<(), Box<dyn Error>> {
                                 timeout: Some(goose::config::DEFAULT_EXTENSION_TIMEOUT),
                                 bundled: Some(true),
                                 description: None,
+                                tools_are_visible_default: true,
+                                tools: HashMap::new(),
                             },
                         })?;
                     }
@@ -641,6 +643,8 @@ pub fn configure_extensions_dialog() -> Result<(), Box<dyn Error>> {
                     timeout: Some(timeout),
                     bundled: Some(true),
                     description: None,
+                    tools_are_visible_default: true,
+                    tools: HashMap::new(),
                 },
             })?;
 
@@ -748,6 +752,8 @@ pub fn configure_extensions_dialog() -> Result<(), Box<dyn Error>> {
                     description,
                     timeout: Some(timeout),
                     bundled: None,
+                    tools_are_visible_default: true,
+                    tools: HashMap::new(),
                 },
             })?;
 
@@ -850,6 +856,8 @@ pub fn configure_extensions_dialog() -> Result<(), Box<dyn Error>> {
                     description,
                     timeout: Some(timeout),
                     bundled: None,
+                    tools_are_visible_default: true,
+                    tools: HashMap::new(),
                 },
             })?;
 
@@ -977,6 +985,8 @@ pub fn configure_extensions_dialog() -> Result<(), Box<dyn Error>> {
                     description,
                     timeout: Some(timeout),
                     bundled: None,
+                    tools_are_visible_default: true,
+                    tools: HashMap::new(),
                 },
             })?;
 
@@ -1619,6 +1629,8 @@ pub async fn handle_openrouter_auth() -> Result<(), Box<dyn Error>> {
                                         timeout: Some(goose::config::DEFAULT_EXTENSION_TIMEOUT),
                                         bundled: Some(true),
                                         description: None,
+                                        tools_are_visible_default: true,
+                                        tools: HashMap::new(),
                                     },
                                 }) {
                                     Ok(_) => println!("✓ Developer extension enabled"),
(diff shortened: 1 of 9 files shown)


cloud-on-prem (approved):

lifeizhou-ap:
- I feel having both` tools_are_visible_default` and `  tools -> tool_name -> visible` is a bit confusing to users.
Shall we make it simpler by using one configuration? for example: 
```
allowed_tools: 
 - bash
 - screen_capture
```

- if the user include  a wrong tool name (non exist tool), I guess it will fail silently?  Just wondering whether the user should know about it?

jsibbison-square:
@lifeizhou-ap 
> * I feel having both` tools_are_visible_default` and `  tools -> tool_name -> visible` is a bit confusing to users.
>   Shall we make it simpler by using one configuration? for example:
> 
> ```
> allowed_tools: 
>  - bash
>  - screen_capture
> ```

Yeah 2 things:
- I wanted to stay away from permission related words like 'allowed' because I see other agents using their [own permissions settings for each tool
](https://docs.anthropic.com/en/docs/claude-code/settings) so maybe `visible_tools` to keep on the visibility theme.
- I definately thought about this simpler structure and I'm still open to it. The reason I landed on an object was because it made it possible to easily extend for tool permissions, like what [claude code settings](https://docs.anthropic.com/en/docs/claude-code/settings) and [amp permissions](https://ampcode.com/manual#permissions) does. What do you think about allowing for extendability via an object?

> * if the user include  a wrong tool name (non exist tool), I guess it will fail silently?  Just wondering whether the user should know about it?

Definately true. This is definately a power user tool atm as you have to have figured out what the tools are actually called, I think we can evolve the usability of it as we see more customer feedback.

lifeizhou-ap:
> @lifeizhou-ap
> 
> > * I feel having both` tools_are_visible_default` and `  tools -> tool_name -> visible` is a bit confusing to users.
> >   Shall we make it simpler by using one configuration? for example:
> > 
> > ```
> > allowed_tools: 
> >  - bash
> >  - screen_capture
> > ```
> 
> Yeah 2 things:
> 
> * I wanted to stay away from permission related words like 'allowed' because I see other agents using their [own permissions settings for each tool
>   ](https://docs.anthropic.com/en/docs/claude-code/settings) so maybe `visible_tools` to keep on the visibility theme.

👍 

> * I definately thought about this simpler structure and I'm still open to it. The reason I landed on an object was because it made it possible to easily extend for tool permissions, like what [claude code settings](https://docs.anthropic.com/en/docs/claude-code/settings) and [amp permissions](https://ampcode.com/manual#permissions) does. What do you think about allowing for extendability via an object?

Can we have separate permissions block in the future to be handled separately? Or you would like to use permission style to expose them now?

jsibbison-square:
> > @lifeizhou-ap
> > > * I feel having both` tools_are_visible_default` and `  tools -> tool_name -> visible` is a bit confusing to users.
> > >   Shall we make it simpler by using one configuration? for example:
> > > 
> > > ```
> > > allowed_tools: 
> > >  - bash
> > >  - screen_capture
> > > ```
> > 
> > 
> > Yeah 2 things:
> > 
> > * I wanted to stay away from permission related words like 'allowed' because I see other agents using their [own permissions settings for each tool
> >   ](https://docs.anthropic.com/en/docs/claude-code/settings) so maybe `visible_tools` to keep on the visibility theme.
> 
> 👍
> 
> > * I definately thought about this simpler structure and I'm still open to it. The reason I landed on an object was because it made it possible to easily extend for tool permissions, like what [claude code settings](https://docs.anthropic.com/en/docs/claude-code/settings) and [amp permissions](https://ampcode.com/manual#permissions) does. What do you think about allowing for extendability via an object?
> 
> Can we have separate permissions block in the future to be handled separately? Or you would like to use permission style to expose them now?

Will change to `visible_tools` list as desired. We haven't discussed permissions so not going to introduce that into this.

jsibbison-square:
Actually I think `available_tools` is better than `visible_tools`

jsibbison-square:
Updated to use an `available_tools` list and also error 'not found' if llm tries to call a tool that is unavailable.

PR #4021, opened 2025-08-12
feat: Prompt injection detection

dorien-koelemeijer:
## Overview
This PR implements prompt injection detection, providing users with security warnings when potentially malicious tool calls are detected. The way prompt injection is implemented re-uses as much of the existing infrastructure as possible. 

## Working design doc with requirements
These notes are a result of pairing sessions with Douwe: https://docs.google.com/document/d/1OWh7Ab_eu8STaoplPA6AKF6AnXQNqXc8JYiAwmUYDTs/edit?tab=t.0

## Implementation approach
* Added security scanning at the check_tool_permissions() stage - a natural choke point in the agent workflow. 
* Security scanning now joins existing permission checking and tool monitoring in one consolidated location
* Security scanner receives both proposed tool calls and full conversation history for comprehensive analysis
* Leveraged existing ToolCall confirmation workflow instead of building custom UI components + security decisions flow through existing approval/denial logging infrastructure
* Users define what models they want to use in the Goose config file - currently we support ONNX models, but in the future we want to expand this scope. The model that is chosen as the default (see Examples below) can be downloaded from HuggingFace directly in ONNX format. 

## Key changes
* `agent.rs`: Added apply_security_results_to_permissions() to move security-flagged tools from approved to needs_approval
* `tool_execution.rs`: Enhanced handle_approval_tool_requests_with_security() to include security warnings in confirmation prompts
* `scanner.rs`: Simplified security messages for user-friendly explanations
* `ToolCallConfirmation.tsx`: Added support for custom security warning prompts
* `GooseMessage.tsx`: Updated to pass security context from backend to UI components

### Goose config file 
* opt-in security scanning
* model selection and configuration
```
security:
  enabled: true
  models:
  - model: protectai/deberta-v3-base-prompt-injection-v2
    threshold: 0.7
    weight: 2.0
```

### Examples of goose sessions (same tool call but different outcomes based on what user has provided in their message(s))
Screenshot 2025-08-12 at 10 52 11 am

Screenshot 2025-08-12 at 10 52 49 am

Thanks to Michael Rand for the test below:
Screenshot 2025-08-14 at 11 54 06 am

## Testing instructions

Add this to the goose config file:
```
security:
  enabled: true
  models:
  - model: protectai/deberta-v3-base-prompt-injection-v2
    threshold: 0.7
    weight: 2.0
```

or this:
```
security:
  enabled: true
  models:
  - model: meta-llama/Llama-Prompt-Guard-2-86M
    threshold: 0.7
    weight: 2.0
```

Then run the following:
```
just run-ui
```


crates/goose/src/security/model_downloader.rs
```diff
@@ -0,0 +1,568 @@
...
+
+        tracing::info!("Python virtual environment setup complete");
+        Ok(())
+    }
+
+    async fn create_conversion_script(&self, model_info: &ModelInfo) -> anyhow::Result<PathBuf> {
+        let script_content = format!(
+            r#"#!/usr/bin/env python3
+"""
+Runtime model conversion script for Goose security models
```
dorien-koelemeijer, line 412: Just to clarify: this only here as a fallback option and is not currently used

crates/goose/src/agents/agent.rs
```diff
@@ -1011,21 +1014,76 @@ impl Agent {
                                             self.provider().await?,
                                         ).await;
 
+                                    // DEBUG: Log tool categorization
+                                    println!("🔍 DEBUG: Tool categorization results:");
```
dorien-koelemeijer, line 1018: Could probably get rid of this - only here for debugging

crates/goose/Cargo.toml
```diff
@@ -100,6 +100,10 @@ tokio-util = "0.7.15"
 lancedb = "0.13"
 arrow = "52.2"
 
+# ML inference backends for security scanning
+ort = "2.0.0-rc.10"  # ONNX Runtime - use latest RC
+tokenizers = { version = "0.20.4", default-features = false, features = ["onig"] }  # HuggingFace tokenizers
+
```
DOsinga, line 107: how much does this add to the executable? do we need the huggingface tokenizer? can we not get this done using tiktoken?

crates/goose/src/agents/agent.rs
DOsinga: yeah, let's

crates/goose/src/agents/agent.rs
```diff
@@ -1078,6 +1081,29 @@ impl Agent {
                                         );
                                     }
                                 } else {
+                                    // Check if we need to show model download status before security scanning
+                                    if let Some(download_message) = self.security_manager.check_model_download_status().await {
+                                        yield AgentEvent::Message(Message::assistant().with_text(download_message));
+                                    }
+
+                                    // SECURITY FIX: Scan tools for prompt injection BEFORE permission checking
```
DOsinga, line 1089: this is probably fine for testing, but we should refactor this so we have a generic way to check whether we want to run a tool that says yes/no/ask the user with prompt and then take the least permissive version of that, i.e. if the counter says no, don't even run the security thing if that makes sense

crates/goose/src/agents/tool_execution.rs
```diff
@@ -54,21 +54,99 @@ impl Agent {
...
                     );
                     yield confirmation;
 
                     let mut rx = self.confirmation_rx.lock().await;
                     while let Some((req_id, confirmation)) = rx.recv().await {
                         if req_id == request.id {
+                            // Log user decision, especially for security-flagged tools
+                            if let Some(security_result) = security_context {
+                                match confirmation.permission {
+                                    Permission::AllowOnce | Permission::AlwaysAllow => {
```
DOsinga, line 117: what does allows allow mean in this context?

crates/goose/src/agents/tool_execution.rs
```diff
@@ -54,21 +54,99 @@ impl Agent {
...
+                                }
+                            } else {
+                                // Log regular tool decisions at debug level
+                                tracing::debug!(
+                                    tool_name = %tool_call.name,
+                                    request_id = %request.id,
+                                    permission = ?confirmation.permission,
+                                    "🔒 User decision for tool execution"
+                                );
+                            }
```
DOsinga, line 148: we should be able to simplify this logging block I think and reduce the code duplication

crates/goose/src/security/model_downloader.rs
```diff
@@ -0,0 +1,621 @@
+use anyhow::anyhow;
```
DOsinga, line 1: I wonder, do we need all of this? it feels like a lot of code. can't we just decide which url we want to download from huggingface/s3 and download that?

ui/desktop/src/components/ToolCallConfirmation.tsx
```diff
@@ -188,9 +190,12 @@ export default function ToolConfirmation({
         </div>
       ) : (
         <div className="goose-message-tool bg-background-default border border-borderSubtle dark:border-gray-700 rounded-b-2xl px-4 pt-2 pb-2 flex gap-2 items-center">
-          <Button className="rounded-full" onClick={() => handleButtonClick(ALWAYS_ALLOW)}>
-            Always Allow
-          </Button>
+          {/* Hide "Always Allow" for security warnings to prevent bypassing future security checks */}
+          {!prompt?.includes('SECURITY WARNING') && (
```
DOsinga, line 194: let's have a separate flag for this or change the hasToolConfirmation thing into an enum

crates/goose/src/security/scanner.rs
```diff
@@ -0,0 +1,799 @@
...
+use tokenizers::Tokenizer;
+
+#[derive(Debug, Clone)]
+pub struct ScanResult {
+    pub is_malicious: bool,
+    pub confidence: f32,
+    pub explanation: String,
+}
+
+/// Trait for different ML inference backends
```
DOsinga, line 21: we only have one though, don't we? probably dont need the generalization at this point then

crates/goose/src/security/scanner.rs
```diff
@@ -0,0 +1,799 @@
...
+            enabled: Self::check_and_prepare_models(),
+        };
+
+        println!("🔒 Scanner enabled: {}", scanner.enabled);
+
+        scanner
+    }
+
+    /// Check if models are available and trigger download if needed
+    fn check_and_prepare_models() -> bool {
```
DOsinga, line 286: this function always returns true. if the model is not around it spawns a task to download it, but that won't be finished when we need the model I think?

crates/goose/src/security/scanner.rs
```diff
@@ -0,0 +1,799 @@
...
+                }
+            }
+        }
+        Err(e) => {
+            tracing::warn!("🔒 Failed to get model downloader: {}", e);
+        }
+    }
+
+    tracing::info!("🔒 ONNX model not available, will use pattern-based scanning");
+    Ok(None)
```
DOsinga, line 209: this function says it might return an error, but it doesn't. instead it returns None or a model. let's make it return an error

crates/goose/src/security/scanner.rs
```diff
@@ -0,0 +1,799 @@
...
+        }
+    }
+
+    /// Legacy method for backward compatibility - now delegates to two-step analysis
+    pub async fn scan_tool_call(&self, tool_call: &ToolCall) -> Result<ScanResult> {
+        // For backward compatibility, just scan the tool call without context
+        self.scan_tool_call_only(tool_call).await
+    }
+
+    /// Legacy method for backward compatibility - now delegates to user message scanning
```
DOsinga: backwards compatible with what?

crates/goose/src/security/scanner.rs
```diff
@@ -0,0 +1,799 @@
...
+            tracing::info!(
+                "🔒 About to call model.predict() with text length: {}",
+                text.len()
+            );
+            match model.predict(text).await {
+                Ok((ml_confidence, ml_explanation)) => {
+                    tracing::info!("🔒 ML model predict returned successfully");
+                    // Get threshold from config
+                    let threshold = self.get_threshold_from_config();
+                    let ml_is_malicious = ml_confidence > threshold;
```
DOsinga, line 531: we do this twice

crates/goose/src/security/scanner.rs
```diff
@@ -0,0 +1,799 @@
...
+
+                    // Medium - suspicious operations
+                    "; rm " | "&& rm " | "| rm " => 0.70,
+                    "kill -9" | "killall" => 0.65,
+
+                    // Lower - potentially legitimate but suspicious
+                    "base64 -d" | "eval " | "exec " => 0.60,
+
+                    _ => 0.50,
+                };
```
DOsinga, line 723: we're mentioning the patterns here twice, would be better to just group them with a score in the initial const. I'm also not sure about these patterns. rm -rf ~/iamges/tmp seems fine for example. network access is also fine

DOsinga (commented):
I think my proposal would be to split this into two things; get a proof of concept in that only uses the pattern matching, but has the right plumbing in for tool call checking and play with that for a bit. And then add the clever bit

crates/goose/Cargo.toml
dorien-koelemeijer: Apologies, didn't realise you'd left feedback - having a look now. Thanks for the feedback 🙏

crates/goose/src/agents/tool_execution.rs
dorien-koelemeijer: It means the request gets processed further.

DOsinga: yeah, I mean, AlwaysAllow doesn't make that much sense in this context but I guess you do want to catch all the arms of the match even though the client shouldn't return that here. so maybe unreachable is better?

dorien-koelemeijer:
Closing this as we discussed creating a simplified version of this with only pattern-based matching and leave the model scanning for a follow-up PR. New PR: https://github.com/block/goose/pull/4237

Issue #4014, opened 2025-08-11
Allow quit without confirmation if system is requesting it


zanesq:
**Please explain the motivation behind the feature request.**
Currently the quit confirmation dialog in desktop stops the computer from restarting for a scheduled update.

**Describe the solution you'd like**
We should look into allowing quit without confirmation if system is requesting it

**Describe alternatives you've considered**
None

**Additional context**
none

- [x] I have verified this does not duplicate an existing feature request

jamadeo:
IMO we should really cut down on when the dialog is shown right now. It doesn't make sense 99% of the time to block Goose from quitting. Maybe when we're waiting on an LLM response or a tool call, but other than that, why block at all?

PR #4023, opened 2025-08-12
feat: User vs System quit

exitcode0:
- differentiate who initiated `quit` (user/system/app)
- track system shutdown state
- add `quitApp()` wrapper
- Skip `quit` confirmation dialog when `system` initiated the `quit`
- IPC handlers for quit initiator information
- tests to mock quitApp() without actual shutdowns

WIP for [#4014](https://github.com/block/goose/issues/4014)

### DIFF
--- a/test_quit_integration.js
+++ b/test_quit_integration.js
@@ -1,0 +1,93 @@
+// Integration test for the Goose Desktop app quit detection system
+// Add this to your main.ts file to enable testing
+
+// Add after the existing imports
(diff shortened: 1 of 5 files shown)


jamadeo:
Thanks @exitcode0 but we are just going to get rid of the confirmation. It never made much sense to have it, as Goose doesn't have "unsaved changes" -- every message is written immediately to the session.

Issue #3978, opened 2025-08-09
"=== CONVERSATION FIX DEBUG ===" and the entire conversation history gets logged after a CTL+C interrupt


capttrousers:
**Describe the bug**

**To Reproduce**
1. goose cli in a session
2. ctl C while goose has control, ie send a reply and quickly ctl + c to get control back to edit the message.
3. press up to get back the last message
4. edit message and send, and see a huge yellow dump to warning logs, of the before vs after convo along with `=== CONVERSATION FIX DEBUG ===`

**Please provide following information:**
 - **OS & Arch:** fedora 41 x86
 - **Interface:** CLI
 - **Version:** 1.1.4
 - **Extensions enabled:** default dev only
 - **Provider & Model:** openai:gpt-5-2025-08-07

**Additional context**
It looks like this is just some debug logging here: https://github.com/block/goose/blob/2c8901003c95908c863f6a0a3379995b67876743/crates/goose/src/agents/agent.rs#L229

Perhaps that should go to debug level logs?

This makes each session basically unusable for scrolling back to re read history once the interrupt happens.

michaelneale:
yep I have seen this too - taking a look...

michaelneale:
fix is here: https://github.com/block/goose/pull/4024 - said debug, but was using `::warn` by mistake, thanks @capttrousers

PR #4024, opened 2025-08-12
fix: this should be a debug message not a warn

michaelneale:
simple change that will fix: https://github.com/block/goose/issues/3978


DOsinga:
hmm, yeah, I don't know. we should fix why ctrl-c makes us get into conversation fixing mode. the conversation fixer is only there because it fixes things that shouldn't be there, so it feels to me that being loud isn't the worst here

PR #4035, opened 2025-08-12
feat: multi model and multi provider config and auto switching

michaelneale:
This generalises the GOOSE_LEAD/WORKER variable structure with a more general multi model approach (ie can remove that old code path). 
This is clearly marked with an x- experimental config prefix (as per convention) so the exact format can evolve as people use this. 

This is really phase 1, phase 2 will be both GUI/ux changes to support and also having more opionionated defaults when providers are configured, but need to get this out in the wild so we can see how things perform with all the permutations of providers we can't test by hand.

For example, in config: 

```yaml
x-advanced-models:
- provider: databricks
  model: goose-gpt-5
  role: reviewer
- provider: anthropic
  model: claude-opus-4-1-20250805
  role: deep-thinker
- provider: anthropic
  model: claude-opus-4-1-20250805
  role: lead
```

which maps to `premade_roles.yaml` which define rules for those models and when they activate: 

```yaml
roles:
  # Deep reasoning and analysis
  - role: "deep-thinker"
    rules:
      triggers:
        keywords: ["think", "reason", "analyze", "explain why", "how does", "what if"]
        match_type: "any"
        complexity_threshold: "high"
        source: "human"  # Only trigger on human messages
      active_turns: 3
      priority: 10
```

which can be used as is - or each trigger value/setting can be overridding in the personal config (if one of the pre-made set doesn't fit - normally you just say role + provider + model - and let it work it out, but you can customise). 

By default the main provider/model is used, and these will supplement when the rules are activate/met, and run for a certain amount of time (this already helped me as the gptoss:120B model was good at spotting short circuit logic bugs other models failed to see). 

This could also be used to do most of the work in a low cost or zero cost local model as well.

run with `--debug` to see it log switching providers and models as it works

discussion https://github.com/block/goose/discussions/3980 
implements: https://github.com/block/goose/issues/4036


joeeuston-dev:
This sounds like an awesome feature. Currently I'm having to switch models when I do more multi modal work as compared to my 'everything' model.

michaelneale:
@joeeuston-dev yeah - I keep finding uses for it - still tuning what defaults I want, but lots of possibilities here and fairly simple in the end!

crates/goose/src/agents/agent.rs
```diff
@@ -1022,6 +1025,18 @@ impl Agent {
...
+                if let Some((new_provider, role, model)) = autopilot.check_for_switch(&messages, self.provider().await?).await? {
+                    debug!("AutoPilot switching to {} role with model {}", role, model);
+                    self.update_provider(new_provider).await?;
+
+                    yield AgentEvent::ModelChange {
+                        model: model.clone(),
+                        mode: format!("autopilot:{}", role),
+                    };
+                }
+                drop(autopilot);
```
jamadeo: I tend to think a block scope is cleaner than an explicit call to drop, but I'm not sure which is more idiomatic rust

```
{
    let mut autopilot = ...
    if ... {
        yield AgentEvent::ModelChange
    }
}
```

crates/goose/src/agents/autopilot.rs
```diff
@@ -0,0 +1,1248 @@
...
+        }
+    }
+
+    /// Analyze text complexity
+    fn analyze_complexity(text: &str) -> ComplexityLevel {
+        // Simple heuristics for complexity
+        let word_count = text.split_whitespace().count();
+        let question_count = text.matches('?').count();
+        let has_code_indicators =
+            text.contains("```") || text.contains("function") || text.contains("class");
```
jamadeo: one of my go-to test prompts is "write a haiku" or some such, and Claude at least loves to return poetry in triple-backticks

crates/goose/src/agents/autopilot.rs
```diff
@@ -0,0 +1,1248 @@
...
+                .iter()
+                .any(|kw| text_lower.contains(&kw.to_lowercase())),
+            MatchType::All => keywords
+                .iter()
+                .all(|kw| text_lower.contains(&kw.to_lowercase())),
+        }
+    }
+
+    /// Analyze text complexity
+    fn analyze_complexity(text: &str) -> ComplexityLevel {
```
jamadeo: This part worries me a tiny bit - there's a lot of static comparisons here that could combine well but might not be obvious if they miss.

I wonder if instead of determining things like complexity here, should we not make autopilot require a model to judge the role?

michaelneale: yeah - but didn't want weight of a local model, and an extra round trip is expensive (if it is to be on each turn) but could do better than static, there are other algorithms too

michaelneale:
@jamadeo "This is very cool and definitely something Goose needs IMO. I'm worried a bit about the number of knobs to turn with a setup like this though. Would users dive into things like triggering keywords, tool/turn counts and priority?"

no - I would hope they don't need to, they only time would be if they really went bespoke, and it is more about simplifying the changes to goose itself, as people discover patterns or learn them, vs needing to change code (similar to with providers we make it as easy as possible ideally to add another one). But absolutely no - they should only need to care about a few roles, we can even have some sensible default setup, and cli/gui can also suggest others - make sense?

crates/goose/src/agents/autopilot.rs
michaelneale: yeah I totally rewrote this, using some off the shelf complexity algos

crates/goose/src/agents/agent.rs
michaelneale: done

crates/goose/src/agents/autopilot.rs
michaelneale: PTAL

michaelneale:
@DOsinga thoughts on if we should consoldate model config to: 

models:

... like this (can have that instead of a variable if we want)? this then can help the cli and GUI have a simpler structure to target when storing a model.

michaelneale:
@jamadeo `What about an approach that uses a model to judge when to switch roles?`

I like that, wouldn't want it on each turn. One option is that early on it could use that model to "turn the nobs and dials" (ie can help select appropriate roles) which will then kick in so you don't have to? What I would like to get to is what is the minimal setup (ie if you setup N providers and M models - how can we best make use of them, how much do we do automatically and how much do we let users direct it).

jamadeo (approved):
Let's do it. I like it as an experiment and we definitely need the model-switching infrastructure. I think we'll change our minds on the switching logic and eventually have preferred models that fulfill each role.

PR #4037, opened 2025-08-12
fix: delete dialog layout

par5ul1:
The dialog for deleting a provider used to look like this:
image
So I fixed it to look like this:
image


michaelneale:
thank you @par5ul1

PR #4041, opened 2025-08-12
MCP UI tool call handling, execute_tool endpoint, and chat UX polish

aharvard:
## What

Adds support for MCP UI tool calls in the desktop app, allowing MCP servers to trigger tool execution through their UI components.

## Why

MCP servers can provide interactive UI components that need to execute tools. Previously, these tool calls weren't properly handled, limiting the interactivity of MCP UI resources.

Here's a demo of why we need to handle tool calls from MCP UI:

https://github.com/user-attachments/assets/9afbe4ba-fb42-47cb-98e1-02b82f1739ce

### Backend (goose-server)
- Added `execute_tool` endpoint that accepts tool name and arguments
- Intelligent tool name resolution: handles both prefixed (`extension__tool`) and unprefixed tool names
- Returns structured responses with success/error states

### Frontend (desktop app)  
- Enhanced `MCPUIResourceRenderer` to intercept MCP UI tool actions
- Tool execution flows through the new backend endpoint
- Dispatches events to update chat with tool request/response
- Auto-scrolls chat to show new messages

### UX Improvements
- Cleaner tool name display by extracting just the tool part (not full `extension__tool` format)
- Conditional toast notifications - only show when needed
- Automatic scroll-to-bottom when new tool results arrive

## Testing
- Tested with MCP servers that provide UI resources with tool actions
- Verified tool name resolution works for various formats
- Confirmed chat updates properly with tool execution results

## Files Changed
- **Backend**: OpenAPI spec, agent routes (execute_tool endpoint)
- **Frontend**: MCPUIResourceRenderer, BaseChat (scroll handling), ToolCallView (name display)
- **Generated**: OpenAPI TypeScript client (sdk.gen.ts, types.gen.ts)
- **New**: useChatEngine hook for tool execution from UI components


aharvard:
.bundle

aharvard:
.bundle

crates/goose-server/src/routes/agent.rs
```diff
@@ -342,6 +357,146 @@ async fn update_session_config(
     }
 }
 
+#[utoipa::path(
```
michaelneale, line 360: is utopia new or we use it elsewhere? (just not familiar with it)

crates/goose-server/src/routes/agent.rs
```diff
@@ -333,7 +348,7 @@ async fn update_session_config(
     if let Some(response) = payload.response {
         agent.add_final_output_tool(response).await;
 
-        tracing::info!("Added final output tool with response config");
+        // final output tool added
```
michaelneale, line 351: could be a debug macro if still needed?

crates/goose-server/src/routes/agent.rs
```diff
@@ -342,6 +357,146 @@ async fn update_session_config(
...
+        (status = 500, description = "Internal server error")
+    ),
+    tag = "Agent Tools"
+)]
+async fn execute_tool(
+    State(state): State<Arc<AppState>>,
+    headers: HeaderMap,
+    Json(payload): Json<ExecuteToolRequest>,
+) -> Result<Json<ExecuteToolResponse>, StatusCode> {
+    verify_secret_key(&headers, &state)?;
```
michaelneale, line 377: Good! (we need some automation that picks this up at build time!)

crates/goose-server/src/routes/agent.rs
```diff
@@ -342,6 +357,146 @@ async fn update_session_config(
...
+        .get_agent()
+        .await
+        .map_err(|_| StatusCode::PRECONDITION_FAILED)?;
+
+    let available_tools = agent.list_tools(None).await;
+
+    // Check if requested tool exists
+    let tool_exists = available_tools.iter().any(|t| t.name == payload.tool_name);
+
+    // 🧭 Resolve tool name to a registered extension prefix if needed
```
michaelneale, line 389: ah a good tell of a LLM comment - I know @DOsinga doesn't like them but I think they are cute and I like them (I do worry if they are lying to me though)

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -330,6 +330,20 @@ function BaseChatContent({
     }
   }, []);
 
+  // Listen for a global scroll-to-bottom request (e.g., from MCP UI tool executions)
```
michaelneale, line 333: this doesn't seem related to the tool calling?

michaelneale: (I guess it is needed - but is there any other side effects to watch for?)

michaelneale:
few questions above, but looks ok, didn't get a chance to try it out yet. Mainly adds a new route (something to track if we ever have to walk back from mcp-ui if it is abandoned one day) but nothing major

crates/goose-server/src/routes/agent.rs
```diff
@@ -342,6 +357,146 @@ async fn update_session_config(
...
+            {
+                resolved_tool_name = candidate;
+            }
+        }
+    }
+
+    // Create a tool call with the resolved name
+    let tool_call = mcp_core::tool::ToolCall::new(resolved_tool_name, payload.arguments);
+
+    // Generate a unique request ID using timestamp
```
DOsinga, line 454: timestamps are not unique

crates/goose-server/src/routes/agent.rs
```diff
@@ -342,6 +357,146 @@ async fn update_session_config(
...
+    let request_id = format!("tool_{}", chrono::Utc::now().timestamp_millis());
+
+    // Create a cancellation token for the tool execution
+    let cancellation_token = tokio_util::sync::CancellationToken::new();
+
+    // Dispatch tool call
+
+    // Execute the tool
+    let (_returned_id, result) = agent
+        .dispatch_tool_call(tool_call, request_id, Some(cancellation_token))
```
DOsinga, line 464: this just doesn't work. there's a reason toolcalling in the agent is a complex process that handles corner cases, handles streaming etc. what if we ran out of context at this point?

DOsinga (changes requested):
I have two concerns here:

* **Architecture**: this is creating a new route, but not using the API wrapper to call it. it parses its own json as far as I can see. it adds directly a message to the messages from the client. this should all be done using typesafe generated classes and processed on the server. the current code in routes/agents duplicates tool calling, that would all have to be refactored

* **Security**  the code as is, if I read it correctly will allow an MCP-UI thing to call any tool at any point without any human interaction. MCP security is shaky as is

shellz-n-stuff:
> * **Security**  the code as is, if I read it correctly will allow an MCP-UI thing to call any tool at any point without any human interaction. MCP security is shaky as is

Just to expand on this one. We wouldn't want a dynamic UI to be able to trigger calls to things like our developer tool (native) and probably third party things (Github/Slack/Etc) could be somewhere between sub-optimal and quite dangerous.

Maybe we need a way to designate very specific tools that are safe to call?

aharvard:
> > * **Security**  the code as is, if I read it correctly will allow an MCP-UI thing to call any tool at any point without any human interaction. MCP security is shaky as is
> 
> Just to expand on this one. We wouldn't want a dynamic UI to be able to trigger calls to things like our developer tool (native) and probably third party things (Github/Slack/Etc) could be somewhere between sub-optimal and quite dangerous.
> 
> Maybe we need a way to designate very specific tools that are safe to call?

I'm thinking that we can allow MCP UIs to call any tool owned by its parent MCP server. So I've attempted to handle this in the backend. The UI dispatches a postMessage that contains the tool name. We then send that name to the new route. The backend prefixes the tool name with the extension name. Then the tool is actually called. 

I might not have this implementation correct, and I would really appreciate any contributions. 

But strategically speaking, are we comfy with MCP UIs calling sibling tools from _their_ parent? And silently ignoring any attempts to call tools outside of their parent's ownership? 

For cross-MCP tool calling, MCP UI can probably leverage other UI Action types (not in scope of this PR), such as `prompt` or `intent`  — this could delegate responsibility to the agent loop. Maybe?

crates/goose-server/src/routes/agent.rs
aharvard: looks like it got [added](https://github.com/block/goose/commit/d8d78396e0cc9b0ec0049e98e630ecc66cf5df34#diff-f8b78f59b067067ff9bfef42f2cebe2f35c85bb230a4e735f56c8882cd5efc44R58) back in May (I think)

ui/desktop/src/components/BaseChat.tsx
aharvard: It is a UX enhancement. It felt awkward to click on something, have the chat thread grow in length, and then manually scroll down. I'd love to know what @zanesq thinks!

aharvard:
@michaelneale could you lend a hand to address @DOsinga feedback re. architecture? I'd love for this work to conform to what's already been battle-tested. 
> Architecture: this is creating a new route, but not using the API wrapper to call it. it parses its own json as far as I can see. it adds directly a message to the messages from the client. this should all be done using typesafe generated classes and processed on the server. the current code in routes/agents duplicates tool calling, that would all have to be refactored

ui/desktop/src/components/BaseChat.tsx
zanesq: I think its fine to add this for now but we should probably circle back to fixing always anchoring the scroll to the bottom during chat responses. It used to work like that but I think it stopped after we added streaming so need to come back to it. It's a bit tricky because we don't want to do auto scroll if the user manually scrolled up to read something above in the output also.

liady:
@aharvard thanks! the search+view-details+add-to-cart flow works end-to-end with the mcpstorefront servers and tool calling, for example try https://mcpstorefront.com?storedomain=allbirds.com 
One question - is it possible that if there are more than 1 mcp-ui resource in a tool response, they will be arranged in a grid? For example with a default iframe height of 400px?
image

Video:
<video src="https://github.com/user-attachments/assets/1c54cee6-ea5d-44bc-93db-33eb17a42805"></video>

ui/desktop/src/components/MCPUIResourceRenderer.tsx
```diff
@@ -2,64 +2,157 @@ import { UIResourceRenderer, UIActionResult } from '@mcp-ui/client';
...
+    },
+    [getProp, getRecord, getString]
+  );
 
   return (
-    <div className="mt-3 p-4 border border-borderSubtle rounded-lg bg-background-muted">
+    <div className="mt-3 p-1 border border-borderSubtle rounded-lg bg-background-muted">
       <div className="overflow-hidden rounded-sm">
         <UIResourceRenderer
           resource={content.resource}
```
liady, line 158: Adding a default `style` here with larger height can help adjust the iframes height, where `ui-size-change` is not thrown or the UI is not responsive:
```tsx
  <UIResourceRenderer
    resource={content.resource}
    onUIAction={handleUIAction}
    htmlProps={{
      style: { height: '400px', minHeight: 0 },
      autoResizeIframe: {
        height: true,
        width: false,
      },
    }}
  />
```

idosal:
This is awesome! Thanks @aharvard and team.

> > > * **Security**  the code as is, if I read it correctly will allow an MCP-UI thing to call any tool at any point without any human interaction. MCP security is shaky as is
> > 
> > 
> > Just to expand on this one. We wouldn't want a dynamic UI to be able to trigger calls to things like our developer tool (native) and probably third party things (Github/Slack/Etc) could be somewhere between sub-optimal and quite dangerous.
> > Maybe we need a way to designate very specific tools that are safe to call?
> 
> I'm thinking that we can allow MCP UIs to call any tool owned by its parent MCP server. So I've attempted to handle this in the backend. The UI dispatches a postMessage that contains the tool name. We then send that name to the new route. The backend prefixes the tool name with the extension name. Then the tool is actually called.
> 
> I might not have this implementation correct, and I would really appreciate any contributions.
> 
> But strategically speaking, are we comfy with MCP UIs calling sibling tools from _their_ parent? And silently ignoring any attempts to call tools outside of their parent's ownership?
> 
> For cross-MCP tool calling, MCP UI can probably leverage other UI Action types (not in scope of this PR), such as `prompt` or `intent` — this could delegate responsibility to the agent loop. Maybe?

From a security perspective, note that MCP servers can already respond with a string that instructs the agent to call any other tool. If Goose can trigger the MCP-UI tool call action in a way that'd still prompt the user for approval, then the attack surface shouldn't change.

crates/goose-server/src/routes/agent.rs
michaelneale: flashbacks to google 😂

michaelneale:
@aharvard in terms of api wrapper - there are `src/api` types and such for the front end which could be added to for this I believe he means?

goose tells me:

```ai-slop
1. **Generated API Files** (in `/ui/desktop/src/api/`):
   - `client.gen.ts` - The HTTP client configuration using `@hey-api/client-fetch`
   - `sdk.gen.ts` - All the API endpoint functions (auto-generated)
   - `types.gen.ts` - TypeScript types for requests/responses
   - `index.ts` - Re-exports everything

2. **Client Initialization** (in `/ui/desktop/src/contexts/ClientInitializationContext.tsx`):
   - The client is configured with:
     - Base URL: Constructed from `GOOSE_API_HOST` and `GOOSE_PORT` config values
     - Headers: Sets `Content-Type: application/json` and `X-Secret-Key` for authentication
   - Initialization happens once when the app starts

3. **Usage Pattern**:
   - Components import functions directly from `'../api'` or `'../api/sdk.gen'`
   - Examples: `getSessionHistory`, `listSessions`, `getTools`, `upsertPermissions`, etc.
   - The client instance can be imported directly from `'./api/client.gen'` when needed

The API wrapper is auto-generated using `@hey-api/openapi-ts` based on the OpenAPI spec from goose-server. When you run `just generate-openapi`, it regenerates these files based on the server's routes.
``` 

For the agent stuff - really that work should be done in the core goose crate I think, using same mechanism for tool calling as the is used (I think?) in agent loop, same types (conversation and whatnot) after goose-server has validated that it is a legit tool call. 

I think it really should be for its own MCP only - both for security but also you don't want to have implicit MCP depends on another MCP to be able to do the work. 

Another option which I haven't really thought about but - does it need to directly do the tool calls? can it instead be a simpler change where the MCP-ui returns well formed and clear directives for the agent loop to take on a round trip, and then invoke the correct tool calls (so there is no leaking, it means it isn't as deterministic but doesn't require all this infra - and would be safer and more like other MCPs). Models may be smart enough to do the right thing?

DOsinga:
> This is awesome! Thanks @aharvard and team.
> 
> > > > * **Security**  the code as is, if I read it correctly will allow an MCP-UI thing to call any tool at any point without any human interaction. MCP security is shaky as is
> > > 
> > > 
> > > Just to expand on this one. We wouldn't want a dynamic UI to be able to trigger calls to things like our developer tool (native) and probably third party things (Github/Slack/Etc) could be somewhere between sub-optimal and quite dangerous.
> > > Maybe we need a way to designate very specific tools that are safe to call?
> > 
> > 
> > I'm thinking that we can allow MCP UIs to call any tool owned by its parent MCP server. So I've attempted to handle this in the backend. The UI dispatches a postMessage that contains the tool name. We then send that name to the new route. The backend prefixes the tool name with the extension name. Then the tool is actually called.
> > I might not have this implementation correct, and I would really appreciate any contributions.
> > But strategically speaking, are we comfy with MCP UIs calling sibling tools from _their_ parent? And silently ignoring any attempts to call tools outside of their parent's ownership?
> > For cross-MCP tool calling, MCP UI can probably leverage other UI Action types (not in scope of this PR), such as `prompt` or `intent` — this could delegate responsibility to the agent loop. Maybe?
> 
> From a security perspective, note that MCP servers can already respond with a string that instructs the agent to call any other tool. If Goose can trigger the MCP-UI tool call action in a way that'd still prompt the user for approval, then the attack surface shouldn't change.

I don't know that that is true. You can create a malicious MCP server that tries to trick Goose into executing a tool by poisoning the prompt and this is something we struggle with as is. But that's rather different then letting an MCP-UI call any tool without any oversight without it being triggered by the user. 

As @aharvard is suggesting, we can look into which tools MCP-UI can call maybe even whitelist them within the MCP server so it can't call all of them and we can warn the user. We can maybe add a one time do you trust this warning when it is triggered the first time and add this to the general tool call scanning by adding it to the actual agent calling. But I don't know how we make it so that this can be triggered when the user isn't watching

ui/desktop/src/components/MCPUIResourceRenderer.tsx
```diff
@@ -2,64 +2,157 @@ import { UIResourceRenderer, UIActionResult } from '@mcp-ui/client';
...
-      case 'prompt': {
-        // TODO: Implement prompt handling
-        handleAction(result);
-        break;
-      }
+        case 'prompt': {
+          // TODO: Implement prompt handling
+          handleAction(result);
+          break;
+        }
```
aharvard, line 73: I just had a chat with @DOsinga and we'd like to pivot this PR away from tool calling and towards `notify` and `prompt` — `tool` might be a possibility in the future after some related ongoing efforts around tool calling in goose bake a little more.

aharvard (commented):
Thanks for all the great feedback, folks!

ui/desktop/src/components/MCPUIResourceRenderer.tsx
aharvard: good call. TY!

aharvard:
> One question - is it possible that if there are more than 1 mcp-ui resource in a tool response, they will be arranged in a grid? For example with a default iframe height of 400px?

@liady — this is something we can look into. thanks for bringing it up! I may reach out with more questions about edge cases.

aharvard:
.bundle

DOsinga:
>I just had a chat with @DOsinga and we'd like to pivot this PR away from tool calling and towards notify and prompt — tool might be a possibility in the future after some related ongoing efforts around tool calling in goose bake a little more.

so we can close this now? if I am mistaken, feel free to reopen

PR #4045, opened 2025-08-12
fix(build): feed electronforge the icon explicitly for linux

eyelight:
Electron forge has not been packaging the icon on rpm & deb builds, resulting in the electron icon getting distributed.

I removed the conditional logic in the config, and explicitly set the icon path in each linux maker, leaving the *.ico (windows) as the default.

To confirm after `npm run make`, assuming you build on x64, this is how to open the packaged logo — before this PR it will be electron, goose after: 

Deb:
```
cd out/make/deb/x64 && dpkg-deb -R goose_1.3.0_amd64.deb pkg && open pkg/usr/share/pixmaps/goose.png
```

Rpm:
```
cd out/make/rpm/x64 && rpm2cpio Goose-1.3.0-1.x86_64.rpm | cpio -idm && open usr/share/pixmaps/Goose.png
```

### DIFF
--- a/ui/desktop/forge.config.ts
+++ b/ui/desktop/forge.config.ts
@@ -58,7 +58,7 @@ module.exports = {
       config: {
         arch: process.env.ELECTRON_ARCH === 'x64' ? ['x64'] : ['arm64'],
         options: {
-          icon: process.platform === 'linux' ? 'src/images/icon.png' : 'src/images/icon.ico',
+          icon: 'src/images/icon.ico',
         },
       },
     },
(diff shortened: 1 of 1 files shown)


michaelneale:
@eyelight could this fix: https://github.com/block/goose/issues/4058 ? sounds like it

michaelneale:
thanks @eyelight that is a great catch

eyelight:
Wasn't aware of #4058 taking a look

DOsinga (approved):

Issue #3972, opened 2025-08-08
bug: blank window being created when recipe is opened via deeplink and the app is closed before opening


zanesq:
**Describe the bug**

Bug: when opening a Goose recipe link, if the Goose app is not open, the behavior is
Goose opens the recipe in new window
Goose then opens another blank window right on top
From the users perspective, it looks like the recipe failed to open. This doesn't happen if the Goose app is already open when the recipe link is clicked

**To Reproduce**
Launch recipe when goose is closed notice extra window.

**Expected behavior**
It shouldn't open an extra window

zanesq:
This seems to be still happening, reopening

zanesq:
I think this might get fixed with the upcoming next camp refactor, will confirm once we wrap that up.

zanesq:
Still happening, working on a fix now

PR #4048, opened 2025-08-12
Load recipe deeplinks in single window when app is closed

zanesq:
Loading recipes from a deeplink when the app is closed was launching an extra window with home behind it. Changed to hook in earlier in the process to load the recipe directly. Also fixed where new chat from home wasn't clearing the loaded recipe.

fixes https://github.com/block/goose/issues/3972

### DIFF
--- a/ui/desktop/src/components/pair.tsx
+++ b/ui/desktop/src/components/pair.tsx
@@ -35,6 +35,7 @@ import 'react-toastify/dist/ReactToastify.css';
 import { cn } from '../utils';
 
 import { ChatType } from '../types/chat';
+import { DEFAULT_CHAT_TITLE } from '../contexts/ChatContext';
 
 export default function Pair({
   chat,
@@ -83,12 +84,14 @@ export default function Pair({
     const resetChat = location.state?.resetChat;
 
     // If we have a resetChat flag from Hub, clear any existing recipe config
+    // This scenario occurs when a user navigates from Hub to start a new chat,
+    // ensuring any previous recipe configuration is cleared for a fresh start
     if (resetChat) {
       const newChat: ChatType = {
         ...chat,
         recipeConfig: null,
         recipeParameters: null,
-        title: 'New Chat',
+        title: DEFAULT_CHAT_TITLE,
         messages: [], // Clear messages for fresh start
         messageHistoryIndex: 0,
       };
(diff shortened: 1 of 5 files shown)


ui/desktop/scripts/unregister-deeplink-protocols.js
```diff
@@ -62,14 +62,6 @@ function unregisterAllProtocolHandlers() {
       }
     });
     
-    // Clean up temporary files
```
zanesq, line 65: noticed some local debug code that wasn't needed

ui/desktop/src/components/pair.tsx
```diff
@@ -80,6 +80,20 @@ export default function Pair({
...
+    // If we have a resetChat flag from Hub, clear any existing recipe config
+    if (resetChat) {
+      const newChat: ChatType = {
+        ...chat,
+        recipeConfig: null,
+        recipeParameters: null,
+        title: 'New Chat',
+        messages: [], // Clear messages for fresh start
+        messageHistoryIndex: 0,
+      };
```
DOsinga, line 97: this should probably be/use a constant seeing that we have 'New Chat' already 4 times

ui/desktop/src/components/pair.tsx
```diff
@@ -80,6 +80,20 @@ export default function Pair({
   // Handle initial message from hub page
   useEffect(() => {
     const messageFromHub = location.state?.initialMessage;
+    const resetChat = location.state?.resetChat;
+
+    // If we have a resetChat flag from Hub, clear any existing recipe config
```
DOsinga, line 86: I don't follow the logic here. in what scenario is this going to be triggered? there was a recipe, the user went to the hub ...

DOsinga (approved):

ui/desktop/src/components/pair.tsx
zanesq: Added a comment to help but yeah basically if they decide to start a new chat from home/hub after loading a recipe for whatever reason it should clear the current recipe and start a new chat. Basically anytime a chat is started from hub/home it should be a fresh chat regardless of the scenario.

DOsinga: that's probably the best we can do at this point, yeah. I guess right now if you restart a session that was started by a recipe, this doesn't actually have the recipe stuff in it anymore? /cc @lifeizhou-ap

zanesq: yeah its always worked like that where the recipe details are lost when resuming a session, we'd need to add the recipe to the session metadata if we wanted to do that.

PR #4052, opened 2025-08-12
Better provider logging

DOsinga:


crates/goose/src/providers/utils.rs
```diff
@@ -103,12 +102,20 @@ pub fn map_http_error_to_provider_error(
...
         _ => {
-            tracing::debug!(
-                "Provider request failed with status: {}. Payload: {:?}", status, payload
-            );
             ProviderError::RequestFailed(format!("Request failed with status: {}", status))
         }
+    };
+
+    if !status.is_success() {
+        tracing::warn!(
```
katzdave, line 110: Do warns make it to the log file?

DOsinga: yes, or at least I saw them

Issue #4016, opened 2025-08-11
UI unresponsive while booting up and loading MCPs / Extensions


zanesq:
**Describe the bug**

The UI is slow to load / freezes if there are MCPs/ Extensions blocking loading. Ideally booting up the extensions would be non blocking on the ui.

The requests seem to be hanging on agent/tools request for 5 mins in this screenshot

**Expected behavior**
The UI should continue to load and possibly show a "Loading" or other mechanism so the user isn't left wondering whats going on. Also if we can we should move this agent/tools request to be async and non blocking for waiting for all tools to load.

**Screenshots**

<img width="1984" height="1320" alt="Image" src="https://github.com/user-attachments/assets/fd275706-25a5-43de-a173-3fa69c769d45" />

**Please provide following information:**
 - **OS & Arch:** Mac ARM
 - **Interface:** UI
 - **Version:** v1.2.0
 - **Extensions enabled:** Redacted
 - **Provider & Model:**

harkamals:
#4027

michaelneale:
@harkamals I don't think that other one is related, that is to do with loading in general, not slowness

zanesq:
Re-opening to see if we can speed up anything with the backend specifically with extensions in recipes and also show a loading screen

zanesq:
Actually I'll create another that is more specific

PR #4054, opened 2025-08-12
speed up loading extensions by loading in parallel

zanesq:
speed up loading extensions by loading in parallel
removed dead code

fixes https://github.com/block/goose/issues/4016

### DIFF
--- a/ui/desktop/src/extensions.tsx
+++ b/ui/desktop/src/extensions.tsx
@@ -213,61 +213,6 @@ export async function removeExtension(name: string, silent: boolean = false): Pr
   }
 }
 
-// Store extension config in user_settings
-function storeExtensionConfig(config: FullExtensionConfig) {
-  try {
-    const userSettingsStr = localStorage.getItem('user_settings');
-    const userSettings = userSettingsStr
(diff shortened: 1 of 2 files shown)


michaelneale:
will give this a try, should help

michaelneale (approved):
seems to work, one caveat @zanesq - not sure if I had my developer turned off or could there be a race condition in there which may not make it show up in GUI? but couldn't see it again so may just be my imagination. 

Nice!

ui/desktop/src/extensions.tsx
```diff
@@ -213,61 +213,6 @@ export async function removeExtension(name: string, silent: boolean = false): Pr
...
-      window.electron.emit('settings-updated');
-    } else {
-      console.log('Extension config already exists in user_settings');
-    }
-  } catch (error) {
-    console.error('Error storing extension config:', error);
-  }
-}
-
-export async function loadAndAddStoredExtensions() {
```
michaelneale, line 243: what happens to this functionality now?

zanesq: we weren't using these anymore I think it was from an older time when we stored them in local storage

Issue #4026, opened 2025-08-12
ctrl-c kills goose instead of stops the current operation


kornpow:
Using the latest version of Goose 1.3 the behavior of `ctrl-c` has changed. Previously `ctrl-c` allowed me to stop the agent and redirect it in a different direction if it is going off-course. Now `ctrl-c` kills goose completely which causes me to lose my context.

```
^Czsh: terminated  goose
```

I would like the previous behavior back, or if there is a better way to achieve what I am looking to do, please let me know!

kornpow:
```
2025-08-12T18:37:45.149778Z ERROR goose::session::storage: Failed to generate session description: Request failed: Failed to parse response: Request failed: Request failed with status: 402 Payment Required
    at crates/goose/src/session/storage.rs:1303

Session ended with error: Failed to generate session description
```

Running out of money mid-session also killed my session, causing me to waste more money re-creating the context.

michaelneale:
thanks @kornpow I thought I was insane for noticing it myself but no one mentioning - but yeah, it did not used to do that. 

FYI you can use --resume (and it continues on - with prompt caching etc) even if it dies (each interaction is essentially stateless so makes no difference)

michaelneale:
*edit* Fix is below thanks to @jamadeo (missed it before)

-----

I think control + C should always interrupt, with gemini, claude code etc, they all do this, so should consider that (and ability to redirect/add/queue up more input vs just quitting).

It also used to gracefully exit, and show the session id at the end, vs a hard terminate.

michaelneale:
eg control C always exiting: 

<img width="500" height="83" alt="Image" src="https://github.com/user-attachments/assets/1749ae4d-878d-44f5-82ba-cef18de80551" />

only ever exits like that for me, also (in the middle of doing a tool call): 

<img width="472" height="115" alt="Image" src="https://github.com/user-attachments/assets/4ec1fe81-0c1d-4d93-b6b0-469000292691" />

I am sure it did not used to be like this, but haven't been able to bisect to where

michaelneale:
ah missed it, this fixes it: https://github.com/block/goose/issues/4026 (but worried how that works on windows).

edit: should be ok on windows, as it is no-op and should be no different to before?

kornpow:
thanks for looking into this! <3 I will definitely try out the resume flag as well

PR #4057, opened 2025-08-12
fix: ctrl-C interruption in the CLI

jamadeo:
The new rmcp child processes were getting the signals forwarded to them. Avoid this by giving them their own process group. They are still cleaned up when the parent exits.

This also fixes the fact that we had three different ways of spawning a child process for an extension (stdio, builtin, and inline python)

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -105,6 +105,42 @@ impl Default for ExtensionManager {
     }
 }
 
+async fn child_process_client(
+    mut command: Command,
+    timeout: &Option<u64>,
+) -> ExtensionResult<McpClient> {
+    command.process_group(0);
+    let (transport, mut stderr) = TokioChildProcess::builder(command)
+        .stderr(Stdio::piped())
+        .spawn()?;
+    let mut stderr = stderr
+        .take()
+        .expect("should have a stderr handle because it was requested");
+
+    let stderr_task = tokio::spawn(async move {
+        let mut all_stderr = Vec::new();
+        stderr.read_to_end(&mut all_stderr).await?;
+        Ok::<String, std::io::Error>(String::from_utf8_lossy(&all_stderr).into())
+    });
+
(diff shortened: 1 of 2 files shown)


crates/goose/src/agents/extension_manager.rs
```diff
@@ -105,6 +105,42 @@ impl Default for ExtensionManager {
     }
 }
 
+async fn child_process_client(
+    mut command: Command,
+    timeout: &Option<u64>,
+) -> ExtensionResult<McpClient> {
+    command.process_group(0);
```
DOsinga, line 112: I thought you said we'd give them their own process group, isn't this the same process group? also, does this work on Windows?

crates/goose/src/agents/extension_manager.rs
```diff
@@ -105,6 +105,42 @@ impl Default for ExtensionManager {
...
+    let (transport, mut stderr) = TokioChildProcess::builder(command)
+        .stderr(Stdio::piped())
+        .spawn()?;
+    let mut stderr = stderr
+        .take()
+        .expect("should have a stderr handle because it was requested");
```
DOsinga: this shouldn't happen of course, but I'd still rather do .ok_or or something so we don't panic

crates/goose/src/agents/extension_manager.rs
jamadeo: 0 sets it to its own group (uses the pid)

Re: Windows: eh, I think no. I don't actually even know what the current behavior here is on windows. I don't have windows machine to test. Any ideas there?

michaelneale:
what behavior change would we expect here - control-C hard exits the session now (new as of this month) not sure if that is meant to change as well?

crates/goose/src/agents/extension_manager.rs
michaelneale: yeah there is no process_group on windows it seems.

michaelneale:
yes this fixes: https://github.com/block/goose/issues/4026

Issue #3641, opened 2025-07-25
Feature request - optional timestamps in session output


omehegan:
**Please explain the motivation behind the feature request.**
When working with both LLMs and tools, I sometimes want to know how long an action took, so I can catch prompts or tool configs that could be more efficient. I guess I can parse the session logs after the fact to work out timing, but if I could see it right in the moment, I could then ask the LLM for suggestions.

Actually, I'm not sure that the session logs really show timestamps either.

**Describe the solution you'd like**
An option for `goose session` that turns on timestamps in the session output, so I can see them while I work. 

**Describe alternatives you've considered**
Generating timestamps from session logs outside of the session itself.

**Additional context**
N/A

- [x] I have verified this does not duplicate an existing feature request

DOsinga:
so you want duration, right? not timestamps - I think we have those. seems reasonable to me especially for tool runs. /cc @spencrmartin

omehegan:
True, duration would be ideal. I imagined timestamps were already stored somewhere and could just be displayed - good enough for me.

DOsinga:
hmm, in my desktop i see this:

<img width="556" height="389" alt="Image" src="https://github.com/user-attachments/assets/9f14b432-2626-40ca-b312-09d504ef79ea" />

those are timestamps no?

omehegan:
Ah, I'm using the CLI.

DOsinga:
where would you want to show the timestamp in the CLI? just before Context maybe?

<img width="483" height="220" alt="Image" src="https://github.com/user-attachments/assets/bd3c948e-28a5-4e38-8a2f-76a0ecbaa019" />

DOsinga:
if so and you want to make it so, go for it

omehegan:
I've changed the request from timestamps to elapsed time, because you're right, that's really what I want, and more useful when displayed that way.

DOsinga:
ok, thanks!

PR #4062, opened 2025-08-13
feat: Add timestamps to output in the CLI

omehegan:
Implements #3641.

This is vibe-coded, as I don't know Rust. I worked with Claude to refactor the original complicated solution to something simpler using `chrono`, which I saw was already in use elsewhere in the project. The final result is concise and AFAICT is idiomatic as well. It compiles, tests pass, and the result is exactly what I was looking for in the original feature request. 

Any feedback would be welcome! Maybe the two functions should just be combined into one? I'm not sure there is any use in having `format` separate from `display`.

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -1253,6 +1253,9 @@ impl Session {
                 }
             }
         }
+
+        // Display timestamp to indicate when the response was completed
+        output::display_timestamp();
         println!();
 
         Ok(())
--- a/crates/goose-cli/src/session/output.rs
+++ b/crates/goose-cli/src/session/output.rs
@@ -762,6 +762,21 @@ pub fn display_context_usage(total_tokens: usize, context_limit: usize) {
     );
 }
 
+/// Display a timestamp indicating when the response was completed
+pub fn display_timestamp() {
+    use console::style;
+
+    let timestamp = format_current_timestamp();
+    println!("{}", style(format!("\n\nCompleted at {}", timestamp)).dim());
+}
+
+fn format_current_timestamp() -> String {
+    use chrono::Local;
+
+    let now = Local::now();
+    now.format("%I:%M %p").to_string()
+}
+
 fn normalize_model_name(model: &str) -> String {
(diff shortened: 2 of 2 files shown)


crates/goose-cli/src/session/mod.rs
```diff
@@ -1253,6 +1253,9 @@ impl Session {
                 }
             }
         }
+
+        // Display timestamp to indicate when the response was completed
```
angiejones:
```suggestion

```

crates/goose-cli/src/session/output.rs
```diff
@@ -762,6 +762,21 @@ pub fn display_context_usage(total_tokens: usize, context_limit: usize) {
     );
 }
 
+/// Display a timestamp indicating when the response was completed
```
angiejones:
```suggestion

```

crates/goose-cli/src/session/output.rs
```diff
@@ -762,6 +762,21 @@ pub fn display_context_usage(total_tokens: usize, context_limit: usize) {
...
+pub fn display_timestamp() {
+    use console::style;
+
+    let timestamp = format_current_timestamp();
+    println!("{}", style(format!("\n\nCompleted at {}", timestamp)).dim());
```
angiejones, line 769: what do you think of removing the "Completed at " prefix and just showing the timestamp? Just thinking of less noise

omehegan: Yeah, I was of two minds about that. In Goose Desktop I see that it already shows the time only, so I thought of going that way to keep it consistent. But then it also felt like in the CLI, where there is less visual difference between the text being displayed, it might be better to make it more clear what the timestamp represents. I don't feel strongly, if you think it's better without I am happy with that.

DOsinga:
can you post a screenshot?

omehegan:
image

omehegan:
Thinking about it more, maybe @DOsinga was right in his comment on https://github.com/block/goose/issues/3641. What I really care about is duration, and without a start timestamp, the timestamps don't actually tell me that in all cases. Let me refactor this.

Issue #4027, opened 2025-08-12
need ability to bypass hermit based uvx and npx shims for airgapped scenarios


harkamals:
**Describe the bug**
When using custom extensions, with bundled: false, goose is still using npx shipped with package, it should respect cmd path supplied instead.

**To Reproduce**
1. Open Goose
2. Open config.yaml and edit
    bundled: false
    cmd: path/to/my/npx 
3. Switch to Goose / extensions / enable extension / 
    3a. it fails to download hermit from internet and fails (because corporate restrictions)
    3b. it resets cmd: npx (which points to Goose supplied npx, and does not work for corporate environments)

**Expected behavior** 
1. Since, it is bundled: false, it should use the cmd value as is 
2. It must not reset cmd: npx (goose supplied)

**Please provide following information:**
 - **OS & Arch:** Mac 15.5
 - **Interface:** UI
 - **Version:** v1.0.2 and v1.0.3
 - **Extensions enabled:** Computer Controller, Memory
 - **Provider & Model:** Claude Sonnet 3.7

harkamals:
@zanesq

michaelneale:
I don't think that other one is related?

michaelneale:
looking at how can detect this and not use the shims in those cases... or allow override

harkamals:
> looking at how can detect this and not use the shims in those cases... or allow override

Suggestion @michaelneale , **bundled: false**, should mean it is a custom extension, and should be respected.

michaelneale:
unfortunately bundled doesn't mean that right now, but have a few solutions. 

What you can already do is a work around:

call it anything other than npx or uvx, eg I copied uvx as /tmp/ovx and it would honour it.

michaelneale:
Closing as workaround: rename uvx and npx as a different name, and refer to them to avoid the hermit automation. 

Also inbound fix https://github.com/block/goose/pull/4063

harkamals:
yes, thanks. the workaround is helpful, i did a sym-link to npx.

next issue is that, it does not move envs{ value:secret} values to env keys, the secret materials are lying in config.yaml with this workaround.

whereas original npx would move the envs{} to secret store, and removes plain text values.

michaelneale:
@harkamals I'm not sure I follow why that would be different with secrets - there is nothing in the npx shim which moved secrets?

PR #4063, opened 2025-08-13
docs: airgapped operation by bypassing hermit for desktop app

michaelneale:
Document airgapping to not use hermit.
Addresses: https://github.com/block/goose/issues/4027

### DIFF
--- a/documentation/docs/troubleshooting.md
+++ b/documentation/docs/troubleshooting.md
@@ -337,6 +337,9 @@ Goose Desktop uses **"shims"** (packaged versions of `npx` and `uvx`) that autom
    ```
 
 3. **Why this works:** Goose only replaces known command names (`npx`, `uvx`, `jbang`, etc.) with its packaged shims. Custom names are passed through unchanged to your system's actual executables.
+
+4. **Require more changes**: In a corporate proxy environment or airgapped environment where the above doesn't work, it is recommended that you customize and package up Goose desktop with shims/config that will work given the network constraints you have (for example, TLS certificate limitations, proxies, inability to download required content etc).
+
 ---
 ### Need Further Help? 
 If you have questions, run into issues, or just need to brainstorm ideas join the [Discord Community][discord]!


documentation/docs/guides/environment-variables.md
```diff
@@ -258,6 +258,16 @@ The default embedding provider is OpenAI. If using a different provider:
 - Specify an appropriate embedding model for that provider
 - Ensure the provider is properly configured with necessary credentials
 
+## Desktop in Airgapped Environments with MCP servers
+
+When running in airgapped environments, you will not want hermit to pull down the runtime environments for MCP servers, set: 
+
+`GOOSE_NOSHIM=true`
```
jamadeo: here it's NOSHIM but below it's NOHERMIT

jamadeo (commented):
100% agree that we should have a way to disable this. A couple of thoughts:

- should probably be in the config, right?
- would this be a better place to check? https://github.com/block/goose/blob/main/ui/desktop/src/components/settings/extensions/agent-api.ts#L149
- should we reconsider doing the shim replacement on the electron side at all? seems a bit odd that it happens there and not in goosed

michaelneale:
@jamadeo I simplified it, there is a viable work around which is documented now, thanks!

documentation/docs/guides/environment-variables.md
```diff
@@ -265,6 +264,15 @@ These variables configure the [Langfuse integration for observability](/docs/tut
 | `LANGFUSE_INIT_PROJECT_PUBLIC_KEY` | Alternative public key for Langfuse | String | None |
 | `LANGFUSE_INIT_PROJECT_SECRET_KEY` | Alternative secret key for Langfuse | String | None |
 
+## Running the Desktop app in Airgapped Environments with MCP servers
+
+When running in airgapped environments, you will not want Hermit to pull down the runtime environments that MCP servers will typically need.
+Goose by default will use "shims" for npx and uvx which are packaged with the application which take care of thus for you behind the scenes.
```
angiejones:
```suggestion
Goose by default will use "shims" for npx and uvx which are packaged with the application which take care of this for you behind the scenes. 
```

documentation/docs/guides/environment-variables.md
```diff
@@ -265,6 +264,15 @@ These variables configure the [Langfuse integration for observability](/docs/tut
 | `LANGFUSE_INIT_PROJECT_PUBLIC_KEY` | Alternative public key for Langfuse | String | None |
 | `LANGFUSE_INIT_PROJECT_SECRET_KEY` | Alternative secret key for Langfuse | String | None |
 
+## Running the Desktop app in Airgapped Environments with MCP servers
```
angiejones: curious why you're adding this to the env var guide instead of the troubleshooting guide?

michaelneale: ah - as it was adding an env var, but yes I should move it to there - any particular place makes sense for this unusual bit of info? (airgapping would be rare) - not entirely sure we should call it airgap perhaps more in certain proxies environments

michaelneale:
@jamadeo I think with cli - it is reasonable to expect a developer has uvx and npx setup to run extensions how they like as it is more explicit

documentation/docs/guides/environment-variables.md
blackgirlbytes: I hope you dont mind. I pushed a commit adding the information to the troubleshooting guide, and I added a callout in the using-extensions giude..basically leading people to that section of the troubleshooting guide.

One thing I'm curious about -- as you said, you were adding it to the env variable guide because it was an env variable, but I don't see the variable anymore. did you remove it? If you kept it..we can add it to the table and just link to that part in the troubleshooting guide OR if you didnt keep the env variable..we can move it overall. 


Lastly, can you double check the docs I added for accuracy? I really just used what I read from your PR and goose to add these docs, but I didn't tinker with this, so their could be inaccuracies.

angiejones: thats the part i was missing. i dont see an env var

michaelneale: @blackgirlbytes I never mind! awesome. Yes - there is no env var at all, this is now just documenting how to work around it, no code change.

michaelneale:
@blackgirlbytes ok edited a bit - and added a section recommending customization should all else fail (ie make your own distro of desktop - which is perfectly fine to do with open source)

michaelneale:
@blackgirlbytes you still want this one in?

blackgirlbytes:
> @blackgirlbytes you still want this one in?

@michaelneale approved!

Issue #3613, opened 2025-07-23
CLI repeatedly showing the list of available models


angelahning:
**Describe the bug**

When user tries to configure the model to use, Goose fetches the list of available models and outputs them to the terminal. With each 'up' or 'down' key, Goose repeatedly outputs the list, making it very confusing and hard to choose the model. 

**To Reproduce**
1. Run `goose configure` and follow the flow to configure the provider
2. Goose now shall fetch the list of available models from the provided and output them to the screen
3. Hit 'down' key
4. the model list is now duplicated on the screen

**Expected behavior**
I expect the up/down key to only toggle the selected model. 

**Screenshots**
 
<img width="959" height="880" alt="Image" src="https://github.com/user-attachments/assets/1965ef43-0204-4bb1-86c8-068627ba4ad9" />

**Please provide following information:**
 - **OS & Arch:** [e.g. macOS 15.5 ARM64]
 - **Interface:** [CLI]
 - **Version:** [e.g. v1.1.3]

jalateras:
works fine for me and i am also using 1.1.3 on my Apple M4 Max running 15.5

<img width="739" height="417" alt="Image" src="https://github.com/user-attachments/assets/6a3cf007-33ed-411f-8c11-52acc8cc29e7" />

PR #4071, opened 2025-08-13
CLI: improve model selection

angelahning:
Issue: A long list of available models overwhelms the terminal screen, making the UX quite confusing as the user are not able to see the search function, and every up/down key (required for picking a model) causes the list to be repeatedly shown on the screen. 

Fix: 
We query the list of supported models from the provider, same as before. If the list is too long, we show the recommended list of models from the provider, with additional search option. 

Before 

https://github.com/user-attachments/assets/060dc3c5-45d3-4cc8-abf2-69a4b2bb3c23

After

https://github.com/user-attachments/assets/ddd5dbe0-53f7-46dd-add5-7ee59119808b


angelahning:
Addresses https://github.com/block/goose/issues/3613

PR #4072, opened 2025-08-13
upgraded all npm packages and fixed related issues

zanesq:
upgraded all npm packages to latest and fixed related issues

tested manually and verified e2e tests are passing


ui/desktop/package.json
```diff
@@ -39,104 +39,102 @@
     "start-alpha-gui": "ALPHA=true npm run start-gui"
   },
   "dependencies": {
-    "@ai-sdk/openai": "^0.0.72",
-    "@ai-sdk/ui-utils": "^1.0.2",
-    "@hey-api/client-fetch": "^0.8.1",
```
zanesq, line 44: no longer need separate api its included in `@hey-api/openapi-ts` package now

ui/desktop/src/api/client.gen.ts
```diff
@@ -1,7 +1,7 @@
 // This file is auto-generated by @hey-api/openapi-ts
 
 import type { ClientOptions } from './types.gen';
-import { type Config, type ClientOptions as DefaultClientOptions, createClient, createConfig } from '@hey-api/client-fetch';
+import { type Config, type ClientOptions as DefaultClientOptions, createClient, createConfig } from './client';
```
zanesq, line 4: these files are generated from `hey-api` latest version now no need for separate package

ui/desktop/src/components/MarkdownContent.tsx
```diff
@@ -107,25 +107,25 @@ export default function MarkdownContent({ content, className = '' }: MarkdownCon
...
+      prose-h2:text-xl prose-h2:font-normal prose-h2:mb-4 prose-h2:mt-4
+      prose-h3:text-lg prose-h3:font-normal prose-h3:mb-3 prose-h3:mt-3
+      prose-p:mt-0 prose-p:mb-2
+      prose-ol:my-2
+      prose-ul:mt-0 prose-ul:mb-3
+      prose-li:m-0 ${className}`}
+    >
       <ReactMarkdown
         remarkPlugins={[remarkGfm]}
-        className={`prose prose-sm text-text-default dark:prose-invert w-full max-w-full word-break
```
zanesq: new version of `ReactMarkdown` doesn't allow className prop so moved above

ui/desktop/src/components/pair.tsx
```diff
@@ -76,7 +76,8 @@ export default function Pair({
       // Clear the location state to prevent re-processing
       window.history.replaceState({}, '', '/pair');
     }
-  }, [location.state, chat.id, setChat]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
zanesq, line 79: These setters aren't needed, React guarantees that state setter functions, such as the setState function returned by the useState hook, maintain referential stability across re-renders. This means that the identity of the setter function itself does not change between renders.

ui/desktop/src/vite-env.d.ts
```diff
@@ -46,7 +46,7 @@ declare module '*.md?raw' {
 }
 
 // Extend CSS properties to include Electron-specific properties
-declare namespace React {
+declare namespace _React {
```
jamadeo: does this mean we don't need it at all?

Issue #3938, opened 2025-08-08
Custom Provider Configurations for Subagents


vibinash:
**Please explain the motivation behind the feature request.**
Currently, subagents inherit the parent agent's provider configuration. 

However, not all models are made equal. Likewise, not all tasks are best performed by one model.

This feature will unlock use-cases that rely on the orchestrator-worker pattern. For example, the orchestrator can use a larger and slower model, while the workers can use a smaller and much faster model for execution. 

Additionally, for use-cases that rely on subagents to do much of the underlying work, this can help lower the overall costs.

**Describe the solution you'd like**
1. Provider configurations via environment variables:
- `GOOSE_PROVIDER` env variable to configure the parent agent provider
- `GOOSE_SUBAGENT_PROVIDER` env variable to configure the subagent provider

2. Additional Subagent configurations:
- max turns
- timeouts

3. Subagents can be spawned directly via natural language prompts. The main agent automatically configures the subagent based on it.

**Describe alternatives you've considered**
The existing [lead/worker model configurations](https://block.github.io/goose/docs/guides/environment-variables#leadworker-model-configuration) relies on turns or failure thresholds to switch between the lead or worker. While this might work for some use-cases, it does feel too rigid for many use-cases where the underlying tasks are ambiguous and not well defined. 
Also, use-cases won't be able to use this configuration if they want to use this pattern for a single turn (one complete prompt-response interaction).

**Additional context**
[slack discussion here](https://square.slack.com/archives/C06PBQ52MCK/p1752186263072219)

- [x] I have verified this does not duplicate an existing feature request

wendytang:
short demo of using subagents with different provider

(video download) https://github.com/user-attachments/assets/a43eecf6-3c32-4475-9b95-15273ff1e4cd

DOsinga:
we gave up on this - sorry! you can use sub recipes to achieve the same

PR #4078, opened 2025-08-13
feat: subagent custom provider and model

wendytang:
addresses Issue #3938

### DIFF
--- a/crates/goose/src/agents/subagent_task_config.rs
+++ b/crates/goose/src/agents/subagent_task_config.rs
@@ -24,6 +24,7 @@ pub const GOOSE_SUBAGENT_MODEL: &str = "GOOSE_SUBAGENT_MODEL";
 pub struct TaskConfig {
     pub id: String,
     pub provider: Option<Arc<dyn Provider>>,
+    pub provider_name: Option<String>,
     pub max_turns: Option<usize>,
 }
 
@@ -42,7 +43,7 @@ impl TaskConfig {
     /// First checks environment variable, then falls back to config
     pub fn get_var(var_name: &str) -> Option<String> {
         // First try environment variable
-        // Do not remove this, it is used by goose-responder client
+        // Do not remove this, env var is used by goose-responder client
         if let Ok(value) = env::var(var_name) {
             return Some(value);
         }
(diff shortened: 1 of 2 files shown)


crates/goose/src/agents/subagent_task_config.rs
```diff
@@ -1,50 +1,106 @@
...
-pub const DEFAULT_SUBAGENT_MAX_TURNS: usize = 5;
+pub const DEFAULT_SUBAGENT_MAX_TURNS: usize = 10;
 
-/// Environment variable name for configuring max turns
-pub const GOOSE_SUBAGENT_MAX_TURNS_ENV_VAR: &str = "GOOSE_SUBAGENT_MAX_TURNS";
+/// config variable name for max subagent turns
+pub const GOOSE_SUBAGENT_MAX_TURNS: &str = "GOOSE_SUBAGENT_MAX_TURNS";
+
+/// config variable name for subagent provider
+pub const GOOSE_SUBAGENT_PROVIDER: &str = "GOOSE_SUBAGENT_PROVIDER";
```
michaelneale, line 17: I don't think we want another env var - but this should hopefully use the consolidated provider config being worked on cc @DOsinga

also wouldn't subagent be more dynamic. ie it could even at runtime decide to use a different model?

DOsinga: I agree. This should be done through settings

wendytang: ack not adding more env vars - this is for a particular use case internally: "we mostly set env vars via SKI's config map. These env are then picked up when we spawn the server in the same SKI pod" 

is there a solution you'd recommend for this use case besides env vars? my understanding is that the local config.yaml wouldn't work here.

https://sq-block.slack.com/archives/C06PBQ52MCK/p1755120454506039?thread_ts=1752186263.072219&cid=C06PBQ52MCK

DOsinga: two things:
* why wouldn't we be able to ship whatever goose is run in the container with the right settings in the config.yaml?
* even if we want to use environment variables, we should still read the values from the config; you can override those always with environment variables I think

PR #4080, opened 2025-08-13
Desktop alerts when suspicious unicode characters found in Recipe

amed-xyz:
Detects potentially harmful Unicode characters in recipes during loading and alerts users through the existing warning
modal system.

Building on the Unicode sanitization work (commits 48c9af0, b0d3c7dd, 5a8f2be0, 08382b8), this feature extends user protection by providing **transparent warnings** when recipes contain suspicious content patterns.


crates/goose/src/recipe/mod.rs
```diff
@@ -253,6 +254,34 @@ pub struct RecipeBuilder {
 }
 
 impl Recipe {
+    /// Returns true if harmful content is detected in instructions, prompt, or activities fields
+    pub fn check_for_security_warnings(&self) -> bool {
```
DOsinga, line 258: you can just loop through all the fields instead of repeating the code

crates/goose-server/src/routes/recipe.rs
```diff
@@ -164,11 +174,32 @@ async fn decode_recipe(
     }
 }
 
+#[utoipa::path(
+    post,
+    path = "/recipes/scan",
+    request_body = ScanRecipeRequest,
+    responses(
+        (status = 200, description = "Recipe scanned successfully", body = ScanRecipeResponse),
+        (status = 400, description = "Bad request")
```
DOsinga: when is this going to throw a 400?

crates/goose/src/recipe/mod.rs
```diff
@@ -745,4 +774,41 @@ isGlobal: true"#;
         let extensions = recipe.extensions.unwrap();
         assert_eq!(extensions.len(), 0);
     }
+
+    #[test]
```
DOsinga, line 770: this test strikes me as too trivial to do any good

crates/goose/src/utils.rs
```diff
@@ -1,17 +1,23 @@
...
-///
-/// This function removes invisible Unicode Tags Block characters (U+E0000-U+E007F)
-/// that can be used for steganographic attacks while preserving legitimate Unicode.
 pub fn sanitize_unicode_tags(text: &str) -> String {
     let normalized: String = text.nfc().collect();
 
     normalized
         .chars()
-        .filter(|&c| !matches!(c, '\u{E0000}'..='\u{E007F}'))
+        .filter(|&c| !is_in_unicode_tag_range(c))
```
DOsinga, line 20: why are we changing this code?

ui/desktop/src/components/ui/RecipeWarningModal.tsx
```diff
@@ -18,25 +18,68 @@ interface RecipeWarningModalProps {
...
   };
+  hasSecurityWarnings?: boolean;
 }
 
 export function RecipeWarningModal({
   isOpen,
   onConfirm,
   onCancel,
   recipeDetails,
+  hasSecurityWarnings = false,
```
DOsinga, line 29: why the default?

ui/desktop/src/hooks/useRecipeManager.ts
```diff
@@ -78,20 +79,22 @@ export const useRecipeManager = (messages: Message[], locationState?: LocationSt
...
 
-  // Check if recipe has been accepted before
+  // Check if recipe has been accepted before and scan for security warnings
   useEffect(() => {
     const checkRecipeAcceptance = async () => {
       if (recipeConfig) {
         try {
+          const scanResult = await scanRecipe(recipeConfig);
+          setHasSecurityWarnings(scanResult.has_security_warnings);
+
```
DOsinga: you could do this in parallel but probably should only scan the recipe if it hasn't been accepted before.

ui/desktop/src/recipe/index.ts
```diff
@@ -133,6 +134,23 @@ export async function decodeRecipe(deeplink: string): Promise<Recipe> {
...
+    if (!response.data) {
+      throw new Error('No data returned from API');
+    }
+
+    return response.data;
+  } catch (error) {
+    console.error('Failed to scan recipe:', error);
+    throw error;
+  }
+}
```
DOsinga, line 152: why do we need this wrapper? we call this once and we have the generated code

ui/desktop/src/hooks/useRecipeManager.ts
```diff
@@ -78,20 +79,22 @@ export const useRecipeManager = (messages: Message[], locationState?: LocationSt
...
           const hasAccepted = await window.electron.hasAcceptedRecipeBefore(recipeConfig);
           if (!hasAccepted) {
             setIsRecipeWarningModalOpen(true);
           } else {
             setRecipeAccepted(true);
           }
         } catch (error) {
-          console.error('Error checking recipe acceptance:', error);
-          // If there's an error, assume the recipe hasn't been accepted
+          setHasSecurityWarnings(false);
```
DOsinga, line 98: if there is an error here, we show the dialog but never the secuirty warning, even if the scan failed?

ui/desktop/src/components/ui/RecipeWarningModal.tsx
```diff
@@ -18,25 +18,68 @@ interface RecipeWarningModalProps {
...
+            )}
           </DialogDescription>
         </DialogHeader>
 
+        {hasSecurityWarnings && (
+          <div className="px-6">
+            <div className="bg-yellow-50 dark:bg-yellow-900/20 border border-yellow-200 dark:border-yellow-800 rounded-lg p-4">
+              <div className="flex items-start">
+                <div className="flex-shrink-0">
+                  <svg className="h-5 w-5 text-yellow-400" viewBox="0 0 20 20" fill="currentColor">
```
DOsinga: I dont think we want to do inline icons

ui/desktop/src/components/ui/RecipeWarningModal.tsx
```diff
@@ -65,7 +108,9 @@ export function RecipeWarningModal({
           <Button variant="outline" onClick={onCancel}>
             Cancel
           </Button>
-          <Button onClick={onConfirm}>Trust and Execute</Button>
+          <Button onClick={onConfirm}>
+            {hasSecurityWarnings ? 'Trust and Continue' : 'Trust and Execute'}
```
DOsinga: why the text change?

DOsinga (approved):
the screenshot looks great. there's a bunch of oddities in the code though we could address...

crates/goose/src/recipe/mod.rs
amed-xyz: good point, done!

crates/goose-server/src/routes/recipe.rs
amed-xyz: removed

crates/goose/src/utils.rs
amed-xyz: not changing, just refactored this bit

ui/desktop/src/components/ui/RecipeWarningModal.tsx
amed-xyz: assume there's no security warnings until we run the check and find any

ui/desktop/src/components/ui/RecipeWarningModal.tsx
amed-xyz: removed

ui/desktop/src/hooks/useRecipeManager.ts
amed-xyz: yes, basically we default to a generic warning (current behaviour) instead of assuming there's an actual security warning and saying so

ui/desktop/src/components/ui/RecipeWarningModal.tsx
amed-xyz: not actually, needed

ui/desktop/src/recipe/index.ts
amed-xyz: no strong reason, was just following same pattern as the other endpoints

crates/goose/src/recipe/mod.rs
amed-xyz: just making sure we're not bypassing any of the places we want to look into

PR #4083, opened 2025-08-14
fix: mcp command error reporting on cli

### LINKED ISSUE #3872: Goose CLI missing dependency message is not informative
**Describe the bug**

if one of the dependency utils like uvx is missing - the message when MCP fails to enable is not great

Image
in this case uvx was not installed, which should be detected

**To Reproduce**
1. delete uv from your machine
2. try to enable any uv based mcp in cli
3. start a session

**Expected behavior**
We should show a clear message when such a core issue happens  explaining a missing UVX/NPX

michaelneale:
when CLI launches mcp servers with stdio, if the command is missing it will not normally say what, this makes it explicit, maybe overkill, may save some debugging time (GUI packages up the commands needed usually): 

now shows: 
Screenshot 2025-08-14 at 4 58 11 pm

fixes: https://github.com/block/goose/issues/3872

### DIFF
--- a/crates/goose/src/agents/extension_manager_test.rs
+++ b/crates/goose/src/agents/extension_manager_test.rs
@@ -1,0 +1,59 @@
+#[cfg(test)]
+mod tests {
+    use super::super::extension::{ExtensionConfig, ExtensionError};
+    use super::super::extension_manager::*;
+    use tokio::process::Command;
+
+    #[tokio::test]
+    async fn test_missing_command_error_message() {
+        // Test that a missing command produces a helpful error message
+        let fake_cmd = "nonexistent_command_xyz_123";
+        let command = Command::new(fake_cmd);
+
+        let result = child_process_client(command, &None).await;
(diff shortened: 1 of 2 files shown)


crates/goose/src/agents/extension_manager_test.rs
```diff
@@ -0,0 +1,59 @@
...
+            dependencies: None,
+            bundled: None,
+        };
+
+        // This should fail if uvx is not installed
+        // We can't guarantee uvx isn't installed in test environment,
+        // but we can test the error path would work
+
+        // For now, just verify the code compiles with our changes
+        // Real integration testing would require a controlled environment
```
jamadeo, line 57: Might as well just drop this test. I suppose you could maybe make it work by modifying PATH? But probably more trouble than it's worth

jamadeo (approved):
This is great -- we'll need more of this sort of thing for common extension problems

PR #4084, opened 2025-08-14
refactor(forge): employ desktopTemplate to support absolute paths

### LINKED ISSUE #4079: fix for: goose desktop won't launch after installing CLI on Linux
__Describe the bug__ The Codename Goose desktop application fails to launch when clicking the desktop icon, despite the .deb package being installed correctly. The application appears to start but immediately closes without displaying the GUI window.

__To Reproduce__

1. Install the Codename Goose .deb package on Ubuntu
2. Attempt to launch the application by clicking the desktop icon
3. Observe that the application doesn't open

__Expected behavior__ Clicking the desktop icon should launch the Codename Goose GUI application.

__Root Cause__ The desktop entry file `/usr/share/applications/goose.desktop` has an incorrect Exec directive:

- Current: `Exec=goose %U` (points to CLI version)
- Should be: `Exec=/usr/lib/goose/Goose %U` (points to GUI binary)

The desktop entry was pointing to the lowercase `goose` command (CLI version) instead of the actual GUI binary named `Goose` (with capital G) located at `/usr/lib/goose/Goose`. This caused the CLI tool to launch instead of the GUI application, resulting in an immediate exit with "EOF" error when no interactive input was provided.

__Proposed Fix__ Update the desktop entry file to point to the correct binary path and name:

```ini
Exec=/usr/lib/goose/Goose %U
```

__Additional context__

- OS & Arch: Ubuntu (24 LTS) x86_64
- Interface: UI
- Version: Unknown (from .deb installation)

eyelight:
To fix #4079 we'll need electron forge to spit out a desktop file with an absolute path for the 'Exec' line. This change directs each linux maker to use a flat, pre-filled desktop file that electronforge calls a 'desktopTemplate'.

The `Icon` line benefits from this change as well as it is now an absolute path.

### DIFF
--- a/ui/desktop/forge.config.ts
+++ b/ui/desktop/forge.config.ts
@@ -70,7 +70,7 @@ module.exports = {
         maintainer: 'Block, Inc.',
         homepage: 'https://block.github.io/goose/',
         categories: ['Development'],
-        mimeType: ['x-scheme-handler/goose'],
+        desktopTemplate: './forge.deb.desktop',
         options: {
           icon: 'src/images/icon.png'
         }
@@ -84,6 +84,7 @@ module.exports = {
         maintainer: 'Block, Inc.',
         homepage: 'https://block.github.io/goose/',
         categories: ['Development'],
+        desktopTemplate: './forge.rpm.desktop',
         options: {
           icon: 'src/images/icon.png'
         }
--- a/ui/desktop/forge.deb.desktop
+++ b/ui/desktop/forge.deb.desktop
@@ -1,0 +1,8 @@
+[Desktop Entry]
+Name=Goose
+Exec=/usr/lib/goose/Goose %U
+Icon=/usr/share/pixmaps/goose.png
+Terminal=false
+Type=Application
+Categories=Development;
+MimeType=x-scheme-handler/goose;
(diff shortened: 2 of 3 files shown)


ui/desktop/forge.config.ts
```diff
@@ -70,7 +70,7 @@ module.exports = {
         maintainer: 'Block, Inc.',
         homepage: 'https://block.github.io/goose/',
         categories: ['Development'],
-        mimeType: ['x-scheme-handler/goose'],
```
DOsinga, line 73: why did this get deleted?

jamadeo:
Thank you @eyelight !

@DOsinga are you able to test it?

DOsinga (approved):
oops, it looks like I never submittee my review. appologies!

ui/desktop/forge.config.ts
eyelight: This line gets transmuted into the `MimeType` line in the resulting .desktop file.

The presence of the `desktopTemplate` line in the forge config causes the defined desktop file to pass through electronforge unchanged iirc. 

So it's initially confusing because 'desktopTemplate' here seems to mean a hardcoded file with no symbol substitution behavior from electronforge.

PR #4085, opened 2025-08-14
fix: hide console window for stdio extensions on Windows

jamadeo:


crates/goose/src/agents/extension_manager.rs
```diff
@@ -111,6 +111,8 @@ async fn child_process_client(
 ) -> ExtensionResult<McpClient> {
     #[cfg(unix)]
     command.process_group(0);
+    #[cfg(windows)]
+    command.creation_flags(0x08000000); // CREATE_NO_WINDOW flag: prevents console window from opening
```
DOsinga: Nice. rather than the comment (he said yelling at the clouds), can we just introduce a named constant that says this?

DOsinga (approved):

crates/goose/src/agents/extension_manager.rs
jamadeo: Yeah that's better. Or maybe even a header file we can import :)

jamadeo:
.bundle-windows

Kvadratni:
this should absolutely work. will test once the bundle is compiled

PR #4099, opened 2025-08-14
Custom providers update

DOsinga:
This is a the combination of a bunch of PRs that allow custom providers

### DIFF
--- a/ui/desktop/package-lock.json
+++ b/ui/desktop/package-lock.json
@@ -11,7 +11,6 @@
       "dependencies": {
         "@ai-sdk/openai": "^2.0.14",
         "@ai-sdk/ui-utils": "^1.2.11",
-        "@hey-api/client-fetch": "^0.13.1",
         "@mcp-ui/client": "~5.6.2",
         "@radix-ui/react-accordion": "^1.2.12",
         "@radix-ui/react-avatar": "^1.1.10",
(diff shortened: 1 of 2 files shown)


ui/desktop/src/components/settings/providers/ProviderGrid.tsx
```diff
@@ -83,6 +127,14 @@ const ProviderCards = memo(function ProviderCards({
         isOnboarding={isOnboarding}
       />
     ));
+
+    if (!isOnboarding) {
```
zanesq: If you click reset providers button in settings it shows the onboarding version of the providers grid.

Why not show custom provider card in onboarding also?

angiejones (approved):

vlascik:
Custom (OpenAI? but maybe also others) provider strips the port from the connection URL, so if you for example want to connect to `http://127.0.0.1:1234` (local LM studio instance) with a custom OpenAI provider, it tries to connect to `http://127.0.0.1` instead, without the port number. Which of course fails.

the problem is at `crates/goose/src/providers/openai.rs` `from_custom_config` fn at line 124 and can be solved e.g. by swapping the code to get host for 
```rs
        let host = if let Some(port) = url.port() {
            format!("{}://{}:{}", url.scheme(), url.host_str().unwrap_or(""), port)
        } else {
            format!("{}://{}", url.scheme(), url.host_str().unwrap_or(""))
        };
```

DOsinga:
thanks @vlascik - I see you already provided a fix! the only way to live

PR #4105, opened 2025-08-15
Fix: LiteLLM API key field not showing in UI configuration

### LINKED ISSUE #3953: Unable To Set an API Key for LiteLLM
**Describe the bug**

I'm unable to set an API key for LiteLLM. The input field does not show up.

**To Reproduce**
1. Go to 'Settings'
2. Click on 'Models'
3. Click on 'Configure Providers'
4. Click on 'LiteLLM'

**Expected behavior**
An option to enter an API key is present.

**Screenshots**

Image

 - **OS & Arch:** Windows 11
 - **Interface:** UI
 - **Version:** 1.2.0
 - **Extensions enabled:** Computer Controller, Memory, Tutorial
 - **Provider & Model:** LiteLLM

**Additional context**
Manually setting LITELLM_API_KEY in the config file does not work. No models are shown in "Switch Models"

jalateras:
## Summary
- Fixed issue where LiteLLM API key input field was not appearing in the UI provider configuration modal
- Changed LITELLM_API_KEY's `required` flag from `false` to `true` to match other providers

## Problem
As reported in #3953, users were unable to set an API key for LiteLLM because the input field did not show up in the Settings > Models > Configure Providers > LiteLLM dialog.

The root cause was that the UI only displays configuration fields that are marked as `required: true`, but LITELLM_API_KEY was incorrectly marked as `required: false`.

## Solution
Updated the LiteLLM provider metadata to mark LITELLM_API_KEY as required, consistent with how all other providers (OpenAI, Anthropic, Google, etc.) handle their API keys.

## Test plan
- [x] Added unit test to verify LITELLM_API_KEY is marked as required
- [x] Test passes: `cargo test test_litellm_metadata_api_key_required`
- [x] Code formatted with `cargo fmt`
- [x] Manual testing: Verify API key field now appears in UI

Fixes #3953


jamadeo (commented):
Thanks @jalateras , a couple of thoughts here:

- I'm surprised we only render the input if the key is required. @zanesq do you know why there isn't a way to have an optional input in the UI?
- Is it always required to have a LITELLM_API_KEY? I haven't used litellm but from a quick glance they seem to support running your proxy locally -- would this be a case where you wouldn't need to set it? what happens then?

zanesq:
@jamadeo I think it was just a UX consideration at the time to limit the fields to only whats needed and keep it simple for the user to configure if nothing is needed. That being said I think it makes sense to put an optional field for everything and pass it along in case since params can change later also.

jalateras:
Thanks for the review @jamadeo and @zanesq!

To address your questions:

1. **Why UI only renders required fields**: As @zanesq mentioned, this was a UX decision to keep configuration simple. However, I agree that showing optional fields makes sense for flexibility.

2. **Is LITELLM_API_KEY always required?**: Good catch! LiteLLM can indeed run locally without an API key (e.g., when using local Ollama models or as a proxy to other local services). The code already handles this - see litellm.rs lines 49-53 where it uses `NoAuth` when no API key is provided.

## Recommendation

I suggest we proceed with this PR as an immediate fix because:
- It unblocks users who need to configure LiteLLM right now
- It maintains consistency with other providers (all mark API keys as required)
- The backend already gracefully handles missing API keys with `NoAuth`
- Users with local setups can enter a dummy value if needed

For the longer-term solution, I propose creating a follow-up issue to enhance the UI to show optional configuration fields. This could be implemented by:
1. Removing the filter that hides non-required fields
2. Adding visual indicators (asterisks for required, "(optional)" label for optional)
3. Updating validation to only enforce required fields

This way we deliver immediate value while planning for a more flexible configuration UI. What do you think?

jalateras:
## ✅ E2E Testing Completed

I've successfully tested this fix end-to-end with LiteLLM:

### Test Setup:
1. Started LiteLLM proxy:

```yaml
# config.yaml
model_list:
  - model_name: openai-gpt-4o-mini
    litellm_params:
      model: openai/gpt-4o-mini
      api_key: sk-xxx

litellm_settings:
  drop_params: true
```

and to run

```console
  docker run -p 4000:4000 \
    -v $(pwd)/litellm_config.yaml:/app/config.yaml \
    ghcr.io/berriai/litellm:main-latest \
    --config /app/config.yaml \
    --port 4000
```

3. Launched Goose UI using the fix from this PR
4. Configured LiteLLM provider in Settings → Models → Configure Providers → LiteLLM

### Test Results:
- ✅ **LITELLM_API_KEY field now appears** in the configuration dialog (was missing before)
- ✅ Successfully entered and saved the API key through the UI
- ✅ Connected to LiteLLM proxy at `http://localhost:4000`
- ✅ Successfully sent messages through the chain: Goose UI → goosed → LiteLLM → OpenAI (gpt-4o-mini)
- ✅ Received responses correctly

### Verification:
The fix works as intended - changing `required: false` to `required: true` makes the LITELLM_API_KEY field visible in the UI, allowing users to configure LiteLLM through the interface rather than having to manually edit config files.

Tested with:
- LiteLLM proxy serving OpenAI's gpt-4o-mini model
- macOS environment
- Current branch: fix/litellm-api-key-required

jalateras:
### Follow-up Work Needed
  A **separate PR should address the wider architectural issue** of supporting truly optional configuration fields in the UI. This would involve:

  - Modifying `DefaultProviderSetupForm.tsx` to show optional fields with proper visual indicators
  - Adding "(optional)" labels and avoiding asterisks for non-required fields
  - Updating validation logic to only enforce required fields
  - Then reverting this change to make LITELLM_API_KEY optional again

jamadeo (approved):

PR #4112, opened 2025-08-15
feat(mcp/developer): accept -1 for insert_line number

sings-to-bees-on-wednesdays:
I noticed that Gemini 2.5 Pro kept trying to use `insert` with `insert_line` set to -1. This makes sense, because:

- there is no easy way to append to the end of the file
- we use -1 to mean "end of file" in other contexts (`view_range`).

So, let's make -1 do the expected thing!

### DIFF
--- a/crates/goose-mcp/src/developer/mod.rs
+++ b/crates/goose-mcp/src/developer/mod.rs
@@ -389,7 +389,7 @@ impl DeveloperRouter {
                 To use the write command, you must specify `file_text` which will become the new content of the file. Be careful with
                 existing files! This is a full overwrite, so you must include everything - not just sections you are modifying.
                 
-                To use the insert command, you must specify both `insert_line` (the line number after which to insert, 0 for beginning) 
+                To use the insert command, you must specify both `insert_line` (the line number after which to insert, 0 for beginning, -1 for end) 
                 and `new_str` (the text to insert).
 
                 To use the edit_file command, you must specify both `old_str` and `new_str` 
@@ -416,7 +416,7 @@ impl DeveloperRouter {
                 unique section of the original file, including any whitespace. Make sure to include enough context that the match is not
                 ambiguous. The entire original string will be replaced with `new_str`.
 
-                To use the insert command, you must specify both `insert_line` (the line number after which to insert, 0 for beginning) 
+                To use the insert command, you must specify both `insert_line` (the line number after which to insert, 0 for beginning, -1 for end) 
                 and `new_str` (the text to insert).
             "#}.to_string(), "str_replace")
         };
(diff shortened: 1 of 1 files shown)


crates/goose-mcp/src/developer/mod.rs
```diff
@@ -1472,6 +1472,13 @@ impl DeveloperRouter {
         let lines: Vec<&str> = content.lines().collect();
         let total_lines = lines.len();
 
+        // Allow insert_line to be negative
+        let insert_line = if insert_line_spec < 0 {
+            total_lines
```
jamadeo: I wonder if this should be `total_lines - insert_line_spec + 1`? That way you get the index-from-end like what you'd see in Python. I'm not sure whether the model would ever use it of course, but probably more correct than the end if the value is -1.

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -416,7 +416,7 @@ impl DeveloperRouter {
                 unique section of the original file, including any whitespace. Make sure to include enough context that the match is not
                 ambiguous. The entire original string will be replaced with `new_str`.
 
-                To use the insert command, you must specify both `insert_line` (the line number after which to insert, 0 for beginning) 
+                To use the insert command, you must specify both `insert_line` (the line number after which to insert, 0 for beginning, -1 for end)
```
jamadeo, line 419: The "correct" way to insert at the end is for the model to pass len(lines). We should still accommodate Gemini but I'd leave this out of the tool description, since the intention is to always use a nonnegative index.

crates/goose-mcp/src/developer/mod.rs
understood-the-assignment: Makes sense to me.

crates/goose-mcp/src/developer/mod.rs
understood-the-assignment: > The "correct" way to insert at the end is for the model to pass len(lines).

This requires that the model knows and keeps track of the exact number of lines in the file, which I think doesn't come "for free", e.g. if the model previously only needed to see the start of the file or some `rg` results, or if the model modified the file since the last time it looked at it.

In any case, I think appending to a file should be easy for the model to do, but I guess it doesn't have to be using the `insert` command.

jamadeo: That's a good point and I'm all for making this easier on the model.

@baxen to your knowledge, might this impact the performance of the editor with claude? I know this tool schema closely resembles https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/text-editor-tool#insert . This is a subtle instruction change but would be good to merge with high confidence in no regression.

michaelneale:
agreed - I also noted gpt-oss:20b would do similar things, nice catch

jamadeo:
Sounds good, let's make it so then

PR #4113, opened 2025-08-15
feat: iterating on summarize oneshot prompt

alexhancock:
Trying to address some issues we're seeing with overly aggressive/short summaries during auto-compaction

Open to feedback on any of this, and also good methods to test

### DIFF
--- a/crates/goose/src/prompts/summarize_oneshot.md
+++ b/crates/goose/src/prompts/summarize_oneshot.md
@@ -1,12 +1,15 @@
-## Summary Task
-Generate detailed summary of conversation to date.  
-Include user requests, your responses, and all technical content.  
+## Task Context
+- An llm context limit was reached when a user was in a working session with an agent (you)
+- Generate a version of the below messages with only the most verbose parts removed
+- Include user requests, your responses, all technical content, and as much of the original context as possible
+- This will be used to let the user continue the working session
+- Use framing and tone knowing the content will be read an agent (you) on a next exchange to allow for continuation of the session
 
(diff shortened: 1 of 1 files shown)


crates/goose/src/prompts/summarize_oneshot.md
```diff
@@ -1,27 +1,30 @@
...
+- Do not exclude any information that might be important to continuing a session working with you
 
 ### Summary Must Include the Following Sections:  
 1. **User Intent** – All goals and requests  
 2. **Technical Concepts** – All discussed tools, methods  
 3. **Files + Code** – Viewed/edited files, full code, change justifications  
 4. **Errors + Fixes** – Bugs, resolutions, user-driven changes  
 5. **Problem Solving** – Issues solved or in progress  
-6. **User Messages** – All user messages, exclude tool output  
+6. **User Messages** – All user messages including tool calls, but truncate long tool call arguments or results
```
katzdave, line 29: Notice any difference from this? I do think this has potential to help; seems like losing where you put a one-off script was one thing people were complaining about.

crates/goose/src/prompts/summarize_oneshot.md
```diff
@@ -1,27 +1,32 @@
 ## Summary Task
-Generate detailed summary of conversation to date.  
-Include user requests, your responses, and all technical content.  
+An llm context limit was reached when a user was in a working session with an agent (you)
+Generate a version of the below messagees with only the most verbose parts removed
```
DOsinga: typo

PR #4124, opened 2025-08-15
feat: Parse and use retryDelay from Google API RateLimitExceeded errors

sings-to-bees-on-wednesdays:
When we exceed the TPM rate limit with the Google API, it helpfully includes a duration for how long to wait before we can safely try again. Let's grab that and use it instead of our existing back-off logic when it's present.


crates/goose/src/providers/google.rs
```diff
@@ -121,7 +121,7 @@ impl Provider for GoogleProvider {
             .await?;
 
         // Parse response
-        let message = response_to_message(unescape_json_values(&response))?;
```
jamadeo, line 128: is this now unused? feel free to delete it from utils.rs if so

crates/goose/src/providers/retry.rs
```diff
@@ -101,7 +101,17 @@ pub trait ProviderRetry {
                             error
                         );
 
-                        let delay = config.delay_for_attempt(attempts);
+                        let delay = if let ProviderError::RateLimitExceeded { retry_delay, .. } = &error {
```
jamadeo: you can simplify to

```
    let delay = match &error {
        ProviderError::RateLimitExceeded {
            retry_delay: Some(provider_delay),
            ..
        } => *provider_delay,
        _ => config.delay_for_attempt(attempts),
    };
```

losing the debug but that doesn't seem terribly important to keep

jamadeo:
@sings-to-bees-on-wednesdays this is good to go, just needs lint + format issues addressed. To format, use `cargo fmt`, and to lint you can run `scripts/clippy-lint.sh`

crates/goose/src/providers/google.rs
understood-the-assignment: Sorry, this was an accidental change, removed.

crates/goose/src/providers/retry.rs
understood-the-assignment: OK, done.

understood-the-assignment:
Thanks, fixed and rebased.

DOsinga:
sorry about the delay, can you resolve the conflicts?

PR #4126, opened 2025-08-15
Clean up agent state between sessions

katzdave:
Main issue is TODO list is being persisted between sessions, but seems like in general a lot of this other state shouldn't be persisted as well?


crates/goose-cli/src/session/builder.rs
```diff
@@ -210,6 +210,11 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
     // Create the agent
     let agent: Agent = Agent::new();
 
+    // Reset session state if we're not resuming
+    if !session_config.resume {
+        agent.reset_session_state().await;
```
alexhancock, line 215: We should be careful how we reset and mutate things right? If the issue is that more state than intended is shared, could any of these changes clear state relevant to one session when starting another?

DOsinga:
what do you mean with "between sessions"? in general a session's properties should be stored in its session object and that is all the state it needs.

crates/goose-cli/src/session/builder.rs
DOsinga: I'm confused. We just created the agent on 211, how would we have any state to reset at this point?

DOsinga: tbh I am confused about the whole TODO tool. I dont see where it loads the existing todo if the session is resumed

katzdave:
Right now TODO Tool stores it's state inside the agent object. That's the major issue and we should just move it completely to the session? Would be great to persist it there too. The issue now is if you start a new session in the desktop it remembers you're old TODO state from fetching it from the agent.

The agent is persisted across the entire process (between session). Maybe that's intended as are each of these subfields; in that case can nuke PR and just move TODO state into the session.

DOsinga:
> The agent is persisted across the entire process (between session). Maybe that's intended as are each of these subfields; in that case can nuke PR and just move TODO state into the session.

we should talk about how the TODO thing is supposed to work, since it is all in memory, if you do resume a session, it will lose its TODO list I think. but to the point in case here, I don't think the agent persists between sessions since we have one agent per goosed and one goosed per session (or one agent per CLI session), so the leakage must occur elsewhere?

crates/goose-cli/src/session/builder.rs
michaelneale: I thought todo was in memory and if there is any x-pollution between sessions it is by mistake? and yes especially in CLI this seems really confusing

michaelneale:
if the session jsonl structure supports it the todo could be persisted along with session messages if you wanted it 1:1 mapping? (just it is treated differently to other messages/content?)

alexhancock:
Can we just disable the TODOs functionality until this can be properly figured out?

cc @tlongwell-block @katzdave

crates/goose-cli/src/session/builder.rs
tlongwell-block: Hey! Yes. It is in memory right now and agent scoped, which is what we discussed before.

But, yeah, this todo can easily be moved to being session scoped and persistent. Let me put up a PR for that instead of just patching the clearing functionality here

alexhancock: @tlongwell-block How involved a change will it be? To get a fix out to include in https://github.com/block/goose/pull/4154 would it be simpler just to disable the tools for now, then re-enable once a comprehensive fix is in?

katzdave:
> Can we just disable the TODOs functionality until this can be properly figured out?
> 
> cc @tlongwell-block @katzdave

sgtm https://github.com/block/goose/pull/4158. Let's disable until we fix this.

tlongwell-block:
Here is the session scoped implementation for reference https://github.com/block/goose/pull/4157

PR #4127, opened 2025-08-16
Fix tests from upstream changes and add testing to lint staged and ci

zanesq:
Fix tests from upstream changes and add testing to lint staged and ci
Also added npm dependency caching Node.js to speed up CI workflows

### DIFF
--- a/.github/workflows/pr-website-preview.yml
+++ b/.github/workflows/pr-website-preview.yml
@@ -28,15 +28,6 @@ jobs:
         with:
           node-version: 20 
 
-      - name: Cache Node.js modules
-        if: github.event.action != 'closed'
-        uses: actions/cache@2f8e54208210a422b2efd51efaa6bd6d7ca8920f # pin@v3
-        with:
-          path: ./documentation/node_modules
-          key: ${{ runner.os }}-documentation-${{ hashFiles('./documentation/package-lock.json') }}
-          restore-keys: |
-            ${{ runner.os }}-documentation-
-
       - name: Install dependencies and build docs
         working-directory: ./documentation
         if: github.event.action != 'closed'


.github/workflows/bundle-desktop-linux.yml
```diff
@@ -154,11 +154,20 @@ jobs:
...
+            .hermit/node/cache
+          key: ${{ runner.os }}-linux-hermit-npm-${{ hashFiles('ui/desktop/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-linux-hermit-npm-
+
       - name: Install npm dependencies
         run: |
           source ./bin/activate-hermit
           cd ui/desktop
-          npm cache clean --force || true
```
zanesq, line 161: @michaelneale do we need this npm cache clean with linux?

.github/workflows/bundle-desktop-intel.yml
```diff
@@ -68,14 +68,6 @@ jobs:
...
           source ./bin/activate-hermit
           echo "Performing pre-build cleanup..."
-          # Clean npm cache
-          npm cache clean --force || true
-          # Clean any previous build artifacts
-          rm -rf target || true
-          # Clean Homebrew cache
-          brew cleanup || true
-          # Remove unnecessary large directories
-          rm -rf ~/Library/Caches/* || true
```
jamadeo, line 78: you can remove this whole step now, all that's left is activate hermit but that would only be until the end of the step execution, so effectively all it's doing is dh -h

.github/workflows/bundle-desktop-linux.yml
```diff
@@ -81,20 +81,6 @@ jobs:
...
-          # Clean Homebrew cache (if exists)
-          brew cleanup || true
-          # Remove unnecessary large directories
-          sudo rm -rf /usr/share/dotnet || true
-          sudo rm -rf /usr/local/lib/android || true
-          sudo rm -rf /opt/ghc || true
-          sudo rm -rf /usr/local/share/boost || true
-          # Clean apt cache
-          sudo apt-get clean || true
-          sudo apt-get autoremove -y || true
```
jamadeo, line 97: same here, just remove the whole step

.github/workflows/bundle-desktop.yml
```diff
@@ -112,14 +112,6 @@ jobs:
...
           source ./bin/activate-hermit
           echo "Performing pre-build cleanup..."
-          # Clean npm cache
-          npm cache clean --force || true
-          # Clean any previous build artifacts
-          rm -rf target || true
-          # Clean Homebrew cache
-          brew cleanup || true
-          # Remove unnecessary large directories
-          rm -rf ~/Library/Caches/* || true
```
jamadeo, line 122: and here

Issue #3914, opened 2025-08-07
Reduce the binary size


DOsinga:
Haven't looked closely into this, but our release binary is huge. We should look into how come and what we can do about it

shawn111:
There's a short discuss about reduce size of goose.

@cgwalters  "A quick git annotate shows this came in with https://github.com/block/goose/pull/2654 which may be a good place to ping..."

https://discord.com/channels/1287729918100246654/1344318255266795641/1402457766244778075

shawn111:
* `webbrowser`: Versions 0.8 and 1.0.   
   * `anyhow`: Versions 1.0 and 1.0.94.
   * `tokio`: Versions 1 and 1.43.
   * `serde`: Versions 1.0 and 1.0.216.
   * `serde_json`: Versions 1.0 and 1.0.133.
   * `once_cell`: Versions 1.19 and 1.20.2.
   * `async-trait`: Versions 0.1.86, 0.1, and 0.1.83.
   * `base64`: Versions 0.22.1, 0.21, and 0.22.
   * `chrono`: Versions 0.4 and 0.4.38.
   * `reqwest`: Versions 0.11 and 0.12.9. 
   
  These duplications likely stem from the project's use of a workspace where individual crates can
   still specify their own dependency versions, a lack of strict versioning, or independent
  development efforts. While some minor version differences might be compatible, they can lead to
  increased binary size, slower compilation times, and potential for subtle bugs.

shawn111:
If we let goosed handle the core / heavy parts like vector_db 

Then goose-cli-lite will be super lite.
```
goose-cli (lite)
                             >>>>>>          goosed 
goose-ui
 ----------------------------------------------------------
 agent (built-in mcp)
                             <<<<<<          
  (dev-env, computer control)
```

DOsinga:
Thanks, this is good research. We're going to remove lancedb, so that should help a bunch. also good points on the different crates. one thing I had been thinking about is, what if goosed and goose-cli would be the same binary? 

I think also optimizing the build settings could help:

```
bloaty target/release/goose                          
    FILE SIZE        VM SIZE    
 --------------  -------------- 
  66.8%  85.9Mi  66.8%  85.9Mi    __TEXT,__text
  11.1%  14.3Mi  11.1%  14.3Mi    __TEXT,__const
   9.2%  11.9Mi   9.2%  11.9Mi    __TEXT,__eh_frame
[...]

PR #4138, opened 2025-08-16
feat: Add `tool_vectordb` feature to `goose` crate

shawn111:
I'm trying build goose on termux and found it is too slow to build.

In https://github.com/block/goose/issues/3914, discuss how to reduce the size of goose.

Change `tool_vectordb` not as default maybe a quick way to improve.

### DIFF
--- a/crates/goose/Cargo.toml
+++ b/crates/goose/Cargo.toml
@@ -96,11 +96,13 @@ dashmap = "6.1"
 ahash = "0.8"
 tokio-util = "0.7.15"
 unicode-normalization = "0.1"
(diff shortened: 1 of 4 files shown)


DOsinga:
thanks, we're working on this: https://github.com/block/goose/pull/3933 to remove all that. I hope we can get that in this week, can you see if that does the trick for you?

shawn111:
Cool, thanks. Let me close it.

michaelneale:
yep about to merge that other one - just tidying up now - thanks for patience!

PR #4140, opened 2025-08-16
feat: Enhanced message queue system with interruption handling

spencrmartin:


michaelneale:
I merged this @spencrmartin but I don't think it was in a working state - did you have some local changes that made what you showed in video work?

PR #4141, opened 2025-08-17
fix: deeplinking a streamable http triggers oauth

### LINKED ISSUE #4006: Extension deeplinks fail silently when authentication or initial connection is required (SSE & Streamable HTTP)
Describe the bug

SSE and Streamable HTTP extensions that require authentication or have initial connection issues fail to install when added via deeplink, while the same extensions work correctly when added through the Extensions UI page.

  To Reproduce
  1. Open a deeplink for an SSE or streamable_http extension that requires authentication (e.g., goose://extension?url=http%3A%2F%2Flocalhost%3A8788%2Fmcp&type=streamable_http&id=envoy&name=Envoy or goose://extension?url=http%3A%2F%2Flocalhost%3A3000%2Fsse&type=sse&name=MySSE)
  2. Goose attempts to install the extension
  3. The installation immediately fails with error: McpServerError { method: "initialize", server: "", source: ServerBoxError(ChannelClosed) }
  4. No opportunity to authenticate or retry, and the extension cannot be installed

  Expected behavior
  When installing an SSE or streamable_http extension via deeplink that requires authentication:
  1. If authentication is needed or connection fails initially
  2. The user should see a helpful message explaining the issue
  3. The extension configuration should appear in a modal
  4. The user can complete authentication/fix connection issues and click "Add Extension" to retry
  5. The extension successfully installs after issues are resolved

  This is the same flow that works correctly when adding the extension through the Extensions UI page.

  Screenshots
  N/A - Error appears in console/logs

  Please provide following information:
  - OS & Arch: macOS (reproduced on multiple systems)
  - Interface: UI
  - Version: v1.1.3 (also present in latest main branch)
  - Extensions enabled: N/A (occurs during extension installation)
  - Provider & Model: N/A (not provider-specific)

  Additional context
  - The issue affects both SSE and streamable_http extensions that require authentication or have initial connection issues
  - The same extensions install successfully when added through the Extensions page UI

kamal:
Fixes #4006


jamadeo:
Hey @kamal thank you for the contribution! Does this still happen on main? There's been several changes very recently to the oauth path for extensions, and this should all be handled within goosed (the "backend").

Deep-linking the Notion MCP (`goose://extension?url=https%3A%2F%2Fmcp.notion.com%2Fmcp&type=streamable_http&id=notiondeeplink&name=notiondeeplink`) triggers the oauth flow for me on main.

kamal:
@jamadeo i just tried this against 1.4.0 and confirmed that it now works. I'll close this out.

PR #4147, opened 2025-08-18
UI: query provider for actual models

### LINKED ISSUE #3645: In the UI we dont seem to load the actual models
**Describe the bug**

When you run the desktop app with a provider and pick a model, we show you the built-in models, even when there is an option to query the actual available models. I think it does work on the CLI though

Image

Image

bonus points to make these searchable and have a better experience in the CLI

angelahning:
The UI queries the providers for the list of supported models, instead of using our hard-coded lists.

fixes #3645

https://github.com/user-attachments/assets/b4afb433-d838-42a6-a5e8-f4fd31a019e5


jamadeo:
Very nice! Probably worth a test too with OpenAI / Anthropic or some other providers we have `fetch_supported_models` implemented for?

angelahning:
> Very nice! Probably worth a test too with OpenAI / Anthropic or some other providers we have `fetch_supported_models` implemented for?

I added some error tests here https://github.com/block/goose/pull/4147/files#diff-1fd59fa8152efe9c7efb97fbbd7a0288bd5fb41ce5f5b4bc97182690e15787b4R877-R912. But it's hard to test for the happy path since fetch_supported_model just returns error without a proper api key. I could use a mock provider, but at that point I am just testing we can call that method and return the result? i dunno.  What other tests do you think is useful? @jamadeo

angelahning:
Having too much trouble fixing the DCO check. Opened a new PR instead https://github.com/block/goose/pull/4270

PR #4150, opened 2025-08-18
Fix: deep link extension installation to show dialog for headers configuration

blackgirlbytes:
When clicking deep links for extensions that include headers (like the GitHub MCP server with Authorization tokens), the extension would try to install directly without opening the configuration dialog. The user needs to see the dialog to review and edit the placeholder tokens. For example the below would just try to install and fail with the value YOUR_ACTUAL_TOKEN_HERE:

```
goose://extension?url=https%3A%2F%2Fapi.githubcopilot.com%2Fmcp%2F&name=GitHub&description=GitHub%20MCP%20Server&type=streamable_http&header=Authorization%3DBearer%20ghs_YOUR_ACTUAL_TOKEN_HERE
```

The deep link handler in addExtensionFromDeepLink only checked for environment variables (config.envs) to determine whether to show the settings dialog, but didnt account for headers. 

I chatted with goose to figure out where this was happening, and I added it in.

### DIFF
--- a/ui/desktop/src/components/settings/extensions/deeplink.ts
+++ b/ui/desktop/src/components/settings/extensions/deeplink.ts
@@ -166,9 +166,13 @@ export async function addExtensionFromDeepLink(
       : getSseConfig(remoteUrl, name, description || '', timeout)
     : getStdioConfig(cmd!, parsedUrl, name, description || '', timeout);
 
-  // Check if extension requires env vars and go to settings if so
-  if (config.envs && Object.keys(config.envs).length > 0) {
-    console.log('Environment variables required, redirecting to settings');
+  // Check if extension requires env vars or headers and go to settings if so
+  const hasEnvVars = config.envs && Object.keys(config.envs).length > 0;
+  const hasHeaders =
+    config.type === 'streamable_http' && config.headers && Object.keys(config.headers).length > 0;
+
+  if (hasEnvVars || hasHeaders) {
+    console.log('Environment variables or headers required, redirecting to settings');
     console.log('Calling setView with:', { deepLinkConfig: config, showEnvVars: true });
     setView('settings', { deepLinkConfig: config, showEnvVars: true });
     return;


ui/desktop/src/components/settings/extensions/deeplink.ts
```diff
@@ -166,9 +166,13 @@ export async function addExtensionFromDeepLink(
...
 
-  // Check if extension requires env vars and go to settings if so
-  if (config.envs && Object.keys(config.envs).length > 0) {
-    console.log('Environment variables required, redirecting to settings');
+  // Check if extension requires env vars or headers and go to settings if so
+  const hasEnvVars = config.envs && Object.keys(config.envs).length > 0;
+  const hasHeaders =
+    config.type === 'streamable_http' && config.headers && Object.keys(config.headers).length > 0;
+
+  if (hasEnvVars || hasHeaders) {
```
DOsinga, line 174: tbh, shouldn't we always show this? and maybe with a warning not to trust all MCP servers?

blackgirlbytes: for clarification..you're saying when you any mcp server..doesnt matter the type or if it has env variables..show the dialog with an notification?

DOsinga: yeah, obviously can leave that for some other time too, but one click MCP is a bit scary to begin with and even if the MCP doesn't require environment variables, maybe there are some optional I would like to plug in

blackgirlbytes: kk, yeah. I'll merge and then open another PR this weekend to enable to dialog to open for all extension types. It looks like there's some work started on an initial malware check for mcp servers that may help accomplish the goal of giving people more trust in what they're installing. https://github.com/block/goose/pull/4272

shellz-n-stuff: The malware check is purely for STDIO. Fully agree with @DOsinga. We want to try inform users as much as possible about what using an MCP means for them in practice ❤️ .

Trustworthy Goose ftw

PR #4153, opened 2025-08-18
feat: autovisualiser of structured data with mcp-ui

michaelneale:
This proposal uses mcp-ui with an (opional) fast, lightweight and 0 dependency built in MCP server called "Auto visualizer" which uses libraries like d3 (which are bundled into inline HTML so no external access required) to make charts and graphs and all sorts of visualisation. 

Benefits: you don't need to know or be experts in presentation, you can just ask for goose to make an appropriate visualisation of anything you want and it can find and use the appropriate diagram for you to interact with inline, to complement the textual output. 

image

image

image

and interactive things, all as tool calls. 

Currently 

- [x] sankey diagrams
- [x] radar charts 
- [x] chord diagrams
- [x] donut/pie charts
- [x] bar graphs and general charts
- [x] flow diagrams
- [x] treemap visualisations (tiles)

Todo: 

need to make sure that the blobs are not too large if part of session, d3 type resources should be JIT served not inlined to html


crates/goose-cli/src/commands/configure.rs
```diff
@@ -31,6 +31,7 @@ fn get_display_name(extension_id: &str) -> String {
     match extension_id {
         "developer" => "Developer Tools".to_string(),
         "computercontroller" => "Computer Controller".to_string(),
+        "autovisualiser" => "Auto Visualiser".to_string(),
```
michaelneale, line 35: this really only applies to gui (MCP-ui) but not sure of protocol here - if we don't have it in CLI then can't turn it on there (but maybe that is ok?)

index.html
aharvard, line 1: are you using this index.html file for anything?

crates/goose-mcp/src/autovisualiser/chart_template.html
```diff
@@ -0,0 +1,334 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Interactive Chart</title>
+
+    <script>
+        {{CHART_MIN}}
```
aharvard: Over in https://github.com/block/goose/pull/4164/files#diff-10caa690bbc229198a1a7c3c862db533dcdf27944a4b92931d5c47aabe7fa500R63 I have upgraded MCP-UI and started to experiment with the iframe `sandbox` value by adding `allow-forms` ... 

I mention this because I think we might be able to use a `<script src="<d3-script-here.js">` if we adjust the sandbox value to include `allow-same-origin`, and if we don't want to inline it like this.

IDK how insecure the `allow-same-origin` value is. If we needed, we could only set that value when it comes from one of our native tools like this one (and turn it off when we load an MCP UI from an 3rd party extension) 

How do you feel about storing the minified JS in Goose as opposed to getting it from a CDN via `<script src="" />`

michaelneale: I think for these cases, can go either way, but for now having it served up from goose is very fast and zero risk which I think feels right (for now) but anything more complicated, yeah.

index.html
michaelneale: nope - looks like a mistake I left it in (I am slowly getting rid of the examples as I xfer them to tempaltes so will get rid of this too)

crates/goose-mcp/src/autovisualiser/chart_template.html
michaelneale: for anything recommended/out of the box, would ideally not pull in external deps at display time I think - so as this works is ideal for this case (for an arbitrary MCP that people may or may not use - I think can be more open).

michaelneale:
hrm current problem: 

```
A React Element from an older version of React was rendered. This is not supported. It can happen if:
- Multiple copies of the "react" package is used.
- A library pre-bundled an old copy of "react" or "react/jsx-runtime".
- A compiler tries to "inline" JSX instead of using the runtime.
- ```

File: ui/desktop/vite.renderer.config.mts [hunk 3]
```diff
@@ -10,7 +11,17 @@ export default defineConfig({
 
   plugins: [tailwindcss()],
 
+  resolve: {
```
michaelneale: @aharvard I needed to do this to get mcp-ui to work with react 19 in goose - FYI

aharvard:
FYI - a fix is in the works https://github.com/idosal/mcp-ui/pull/91

michaelneale:
fyi @aharvard think this is ready to have a good look. Seems to work now (improved the data validation). there isn't any smarts in the templates for responding to dark mode yet, or posting tings back, but this is a start and still pretty simple and fast.

crates/goose-mcp/src/autovisualiser/chart_template.html
aharvard: Agreed, inlining the deps feels right

ui/desktop/src/components/ui/Pill.tsx
```diff
@@ -22,11 +22,13 @@ export function Pill({
   disabled = false,
   animated = false,
 }: PillProps) {
-  const baseStyles = 'inline-flex items-center justify-center rounded-full transition-all duration-300 ease-out font-medium';
-  
+  const baseStyles =
```
alexhancock: Are these just formatting changes? Looks like it?

alexhancock (approved):

ui/desktop/src/components/ui/Pill.tsx
michaelneale: oh good catch - yeah not intended. I am using opus + a fast apply model and sometimes it gets carried away

michaelneale:
"Then, we could return a text statement that clearly indicates whether the visualization was successful or failed for some reason" - yeah I like that - might be a way to validate the js/total thing in rust side and feed it back (it does push back if the json is invalid, so that can cause it to iterate a bit)

ui/desktop/src/components/ui/Pill.tsx
michaelneale: yeah - looks like the prettier/checker requires it? odd...

PR #4160, opened 2025-08-18
Retry all 500 codes

DOsinga:
Anthropic throws a 529 which is somewhat unofficial - this makes us have a ServerError for all 500's which then all get retried.


michaelneale (approved):
LGTM - in theory I dislike this, but I have seen so many variations of errors from anthropic in particular, that a general handled for server errors like this makes sense (at least it isn't buried in a 200ok response!)

michaelneale:
a more realistic back pressure anthropic should use is 429 (as it implies the client can do something). 500 is them taking the blame (fair enough)

Issue #4117, opened 2025-08-15
MCP-UI `allow-forms` in Goose?


aharvard:
## Question
Should MCP-UI `<iframe />` instances in Goose `allow-forms` on the `sandbox`? 

Use case: https://x.com/kentcdodds/status/1956134483020931415

## Currently

When the `<MCPUIResourceRenderer />` component in Goose receives raw HTML from an embedded resource in a tool response, it renders an `<iframe />` like this: 
```html
<iframe 
  srcdoc="<html>...</html>" 
  sandbox="allow-scripts" 
  title="..." 
  style="...">
</iframe>
```

When the `<MCPUIResourceRenderer />` component in Goose receives an external URL from an embedded resource in a tool response, it renders an `<iframe />` like this: 
```html
<iframe 
  src="..." 
  sandbox="allow-scripts allow-same-origin" 
  title="..." 
  style="...">
</iframe>
```

This iframe comes from `@mcp-ui/client`, which automatically sets the iframe `sandbox`:
- for raw HTML: `allow-scripts`
- for external URL: `allow-scripts allow-same-origin`

## Possible Config 

We have the ability to override the sandbox permissions by setting `htmlProps.iframeProps.sandbox` to a string (refer to: https://x.com/idosal1/status/1956139895078183102). 

This allows Goose to provide a more strict or more lenient CSP — coming soon is a more ergonomic approach https://github.com/idosal/mcp-ui/pull/83.

 ```tsx
// MCPUIResourceRenderer.tsx 

import { UIResourceRenderer } from '@mcp-ui/client';

// ...

<UIResourceRenderer
  resource={content.resource}
  onUIAction={handleUIAction}
  htmlProps={{
    iframeProps: {
      sandbox: 'allow-forms allow-scripts',
    },
    // other htmlProps are present in the source code but removed here for focused discussion
  }}
/>

// ...
```
```html
<!-- assuming raw HTML is returned by the tool response, the above JSX would yield...  -->

<iframe 
  srcdoc="<html>...</html>" 
  sandbox="allow-forms allow-scripts" 
  title="..." 
  style="...">
</iframe>
```

### Support `allow-forms`
If we want to add `allow-forms` to the `sandbox` of the MCP-UI `<iframe />`, this will be a quick implementation

## Provide guidance for an alternative path
If we don't `allow-forms`, we need to document why not, and propose alternative solutions such as: 
1. MCP-UI resources should be stateful and post via fetch api
[...]

PR #4164, opened 2025-08-18
feat: upgrade `@mcp-ui/client` package and improve UI message handling

aharvard:
This PR enhances MCP-UI functionality in Goose. 

## Chore: MCP-UI Upgrade
I bumped @mcp-ui/client to version ([5.9.0](https://github.com/idosal/mcp-ui/blob/main/sdks/typescript/client/CHANGELOG.md)), this lets us experiment with the following:
1. tailoring `htmlProps.sandboxPermissions`
2. use new `isUIResource(content)` helper for convenience

## Feat: Allow forms for iframe sandbox
We got a request to `allow-forms` from our MCP-UI iframe: https://github.com/block/goose/issues/4117. So, this PR allows us to experiment with setting `htmlProps.sandboxPermissions` to `allow-forms` — for more info refer to docs https://mcpui.dev/guide/client/resource-renderer#props-details. 

> [!IMPORTANT]
> I'm unsure what attack vectors this introduces but would like to ask if we can be flexible and quick to make updates to `htmlProps.sandboxPermissions` over time as we learn more.  

## Feat: Embedded UI Message Support
For more info on what MCP-UIs can message to Goose, refer to: https://mcpui.dev/guide/embeddable-ui#message-types

https://github.com/user-attachments/assets/efd0fbe4-1db7-4b01-89c8-ab9ddaa85e75

| Message | Handled in Goose? | Goose Response |
|--------|--------|--------|
| [intent](https://mcpui.dev/guide/embeddable-ui#intent): the user has interacted with the UI and expressed an intent, and the host should act on it | ❌ Acknowledge by goose with a meaningful response, but does not fire any host capability. It needs an implementation plan to become functional. | Returns an error status with additional information about what the MCP UI requested |
| [notify](https://mcpui.dev/guide/embeddable-ui#notify): the iframe already acted upon the user interaction, and is notifying the host to trigger any side effects | ✅ Acknowledge by goose with a meaningful response and fires a toast notification. This could evolve into passing a message into the chat, but would need an implementation plan. | Returns meaningful data if successful or an error message if it fails. |
| [prompt](https://mcpui.dev/guide/embeddable-ui#prompt): the iframe asks the host to run a prompt | ✅ Acknowledge by goose with a meaningful response, and the message is passed directly to the chat engine. | Returns meaningful data if successful or an error message if it fails. |
| [tool](https://mcpui.dev/guide/embeddable-ui#tool): the iframe asks the host to run a tool call |  ❌ Acknowledge by goose with a meaningful response, but does not fire any host capability. It needs an implementation plan to become functional. | Returns an error status with additional information about what the MCP UI requested. 
| [link](https://mcpui.dev/guide/embeddable-ui#link): the iframe asks the host to navigate to a link | ✅ Acknowledge by goose with a meaningful response, and calls electron `openExternal()` method. | Returns meaningful data if successful or an error message if it fails | 

## Fix: UI Action callback response
In our first iteration, `onUIAction={handleUIAction}` did not return a meaningful response (refer to [this issue](https://github.com/idosal/mcp-ui/issues/82#issuecomment-3197146359) raised over on mcp-ui). Now it does.

### DIFF
--- a/ui/desktop/src/components/MCPUIResourceRenderer.tsx
+++ b/ui/desktop/src/components/MCPUIResourceRenderer.tsx
@@ -1,23 +1,53 @@
 import {
   UIResourceRenderer,
-  UIActionResult,
   UIActionResultIntent,
   UIActionResultLink,
   UIActionResultNotification,
   UIActionResultPrompt,
   UIActionResultToolCall,
 } from '@mcp-ui/client';
+import { useState, useEffect } from 'react';
 import { ResourceContent } from '../types/message';
 import { toast } from 'react-toastify';
-
-// TODOS
-// figure out how best to handle the ui-lifecycle-iframe-ready message
-// figure out how best to support size-change messages
 
 interface MCPUIResourceRendererProps {
   content: ResourceContent;
   appendPromptToChat?: (value: string) => void;
 }
+type UISizeChange = {
+  type: 'ui-size-change';
+  payload: {
+    height: number;
+    width: number;
+  };
+};
+
+// Reserved message types from iframe to host
+type UILifecycleIframeReady = {
+  type: 'ui-lifecycle-iframe-ready';
+  payload?: Record<string, unknown>;
+};
+
+type UIRequestData = {
+  type: 'ui-request-data';
+  messageId: string;
+  payload: {
+    requestType: string;
+    params: Record<string, unknown>;
+  };
+};
+
+// We are creating a new type to support all reserved message types that may come from the iframe
+// Not all reserved message types are currently exported by @mcp-ui/client
+type ActionEventsFromIframe =
+  | UIActionResultIntent
+  | UIActionResultLink
+  | UIActionResultNotification
+  | UIActionResultPrompt
+  | UIActionResultToolCall
+  | UISizeChange
+  | UILifecycleIframeReady
+  | UIRequestData;
 
 // More specific result types using discriminated unions
 type UIActionHandlerSuccess<T = unknown> = {
@@ -59,19 +89,62 @@ enum UIActionErrorCode {
   TIMEOUT = 'TIMEOUT',
 }
 
+// toast component
+const ToastComponent = ({
+  messageType,
+  message,
+  isImplemented = true,
+}: {
+  messageType: string;
+  message?: string;
+  isImplemented?: boolean;
+}) => {
+  const title = `MCP-UI ${messageType} message`;
+
+  return (
+    <div className="flex flex-col gap-0 py-2 pr-4">
+      <p className="font-bold">{title}</p>
+      {isImplemented ? (
+        <p>
+          Message received for <span className="font-bold">{message}</span>.
+        </p>
+      ) : (
+        <p>
+          Message received for <span className="font-bold">{message}</span>.
+          <br />
+          {messageType.charAt(0).toUpperCase() + messageType.slice(1)} messages aren't supported
+          yet, refer to console for more details.
+        </p>
+      )}
+    </div>
+  );
+};
+
 export default function MCPUIResourceRenderer({
   content,
   appendPromptToChat,
 }: MCPUIResourceRendererProps) {
-  const handleUIAction = async (actionEvent: UIActionResult): Promise<UIActionHandlerResult> => {
-    console.log('[MCP-UI] Action received:', actionEvent);
+  const [currentThemeValue, setCurrentThemeValue] = useState<string>('light');
 
+  useEffect(() => {
+    const theme = localStorage.getItem('theme') || 'light';
+    setCurrentThemeValue(theme);
+    console.log('[MCP-UI] Current theme value:', theme);
+  }, []);
+
+  const handleUIAction = async (
+    actionEvent: ActionEventsFromIframe
+  ): Promise<UIActionHandlerResult> => {
+    // result to pass back to the MCP-UI
     let result: UIActionHandlerResult;
 
     const handleToolCase = async (
       actionEvent: UIActionResultToolCall
     ): Promise<UIActionHandlerResult> => {
       const { toolName, params } = actionEvent.payload;
+      toast.info(<ToastComponent messageType="tool" message={toolName} isImplemented={false} />, {
+        theme: currentThemeValue,
+      });
       return {
         status: 'error' as const,
         error: {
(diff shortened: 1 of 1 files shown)


aharvard:
.bundle

ui/desktop/src/components/MCPUIResourceRenderer.tsx
```diff
@@ -8,54 +8,45 @@ interface MCPUIResourceRendererProps {
...
+      payload: result,
+    };
+
+    console.info(
+      `Goose posted the following response message back to the MCP-UI request:`,
+      response
+    );
+
+    return response;
   }, []);
```
aharvard: Hey @idosal, I addressed https://github.com/idosal/mcp-ui/issues/82 by returning `response`.

I'd love to get your feedback if I got this right (I tested against @kentcdodds server and fixed [the one issue](https://github.com/idosal/mcp-ui/issues/82) he reported). 

This fix made me curious. Should @mcp-ui/client have some stronger typing around what the callback should return? It would be cool to see an error on `onUIAction={handleUIAction}` if the returned response is not as expected.

idosal: Thank @aharvard! WDYT about returning the result of the UI Action callback (e.g., the response to the tool call) or some ack that it finished successfully, rather than the UI action itself? I realize `UIActionResult is a confusing name since `result` refers to the event generated by the action and not the result of the action itself.
@kentcdodds what would best serve your use case?

Adding typing to guide responses is a great idea. I'll tag you in the PR.

michaelneale:
@aharvard nice - I think if you update to main and fix conflicts, you may run into error with react versions that I had - https://github.com/block/goose/pull/4153/commits/2e2369f8b61f60aad2c04b739e1a2fdc6547d266 is one way to fix it

ui/desktop/src/components/MCPUIResourceRenderer.tsx
```diff
@@ -1,74 +1,289 @@
...
+    // Log result with appropriate level
+    if (result.status === 'error') {
+      console.error('[MCP-UI] Action failed:', result);
+    } else {
+      console.log('[MCP-UI] Action succeeded:', result);
     }
-  }, []);
+
+    return result;
+  };
```
aharvard: Hey @idosal, thanks again for taking a look at this. 
I took another pass at handling UI actions here. Now, I'm returning a lot more info. Is this more in line with your suggestion?

aharvard:
.bundle

ui/desktop/package.json
```diff
@@ -41,7 +41,8 @@
   "dependencies": {
     "@ai-sdk/openai": "^2.0.14",
     "@ai-sdk/ui-utils": "^1.2.11",
-    "@mcp-ui/client": "~5.6.2",
+    "@hey-api/client-fetch": "^0.8.1",
```
aharvard: Todo: look into this diff

aharvard: @hey-api/client-fetch should have been removed in a conflict resolution. will remove.

aharvard:
.bundle

ui/desktop/src/components/MCPUIResourceRenderer.tsx
aharvard: moving convo to https://github.com/block/goose/pull/4164#discussion_r2286941150

ui/desktop/src/components/MCPUIResourceRenderer.tsx
idosal: Thanks @aharvard, it looks great. It captures the data that could be relevant to the UI.

ui/desktop/src/components/GooseMessage.tsx
```diff
@@ -208,6 +208,7 @@ export default function GooseMessage({
                   toolResponse={toolResponsesMap.get(toolRequest.id)}
                   notifications={toolCallNotifications.get(toolRequest.id)}
                   isStreamingMessage={isStreaming}
+                  append={append}
```
michaelneale, line 211: what does this do?

ui/desktop/src/main.ts
```diff
@@ -1109,6 +1109,17 @@ ipcMain.on('react-ready', () => {
   console.log('[main] React ready - window is prepared for deep links');
 });
 
+// Handle external URL opening
+ipcMain.handle('open-external', async (_event, url: string) => {
```
michaelneale, line 1113: what functionality does this add that wasn't before - is it for cases where the type is links and it will open it in external browser? (if so - a hyperlink won't do that?)

michaelneale:
I think a good upgrade - main Q was what to test for expecially adding messages, and also - the new open rpc - not sure what that does that a hyperlink can't do?

ui/desktop/src/components/GooseMessage.tsx
aharvard: `append` is an existing convention in this file, and I had to add it to the `<ToolCallWithResponse />` component to prop drill. TBH, this feels like a hack. 

Inside of `<ToolCallWithResponse />`, `append` is passed to `<MCPUIResourceRenderer content={content} appendPromptToChat={append} />` — where, I think, `appendPromptToChat` is a bit clearer. 

The goal is to 
1. Catch the prompt message from an MCP UI
2. Pass that message to the chat engine 
3. Let goose take the wheel

@zane, do you know if there is a more elegant way to pass a prompt string to the chat engine w/out prop drilling?

ui/desktop/src/main.ts
aharvard: Since MCP-UI hyperlinks live inside of an iframe, clicking on a typical `<a href="" />` link will refresh the contents inside of the iframe. That might be perfect for some MCP-UIs but not others. 

Also, I think that Electron brings in some form of protection that may prevent hyperlinks from performing as authored, to protect the app user. (I have a lack of knowledge in this area, I could be wrong) 

For example, in a typical web browser that has an iframe on the page, the hyperlink author can add an attribute to force a navigation at the parent level. In our Electron app, these attributes do nothing and clicking on a link fails to do what a user might expect it to.  

The MCP-UI SDK gives us the [link message type](https://mcpui.dev/guide/embeddable-ui#link) for, I believe, better security hygiene. So adding `open-external` was my attempt at finding the safest execution route.

aharvard:
FYI, just push some last bits of feature refinement. Ready for re-review @zanesq, @michaelneale, @JHKennedy4, and others. TY!

zanesq:
.bundle

ui/desktop/package.json
```diff
@@ -139,6 +139,10 @@
   },
   "keywords": [],
   "license": "Apache-2.0",
+  "overrides": {
+    "react": "^19.1.1",
+    "react-dom": "^19.1.1"
+  },
```
aharvard, line 145: had to do this, but we can remove when this issue is completely resolved https://github.com/idosal/mcp-ui/issues/90

ui/desktop/src/components/GooseMessage.tsx
zanesq: I think prop drilling is ok for 2 levels like this for now but we don't have any other global state mechanism currently other than React context. So you could use context if needed. We plan on adding a global state library soon that can make things like this easier.

PR #4165, opened 2025-08-18
Add PKCE support for Tetrate Agent Router Service

johnlanda:
Adds PKCE support for Tetrate Agent Router Service, similar to openrouter PKCE signup flow

Tetrate when signed up via a "work address" currently will give goose users starter credit, which is ideal. 

Screenshot 2025-08-21 at 2 38 04 pm

### DIFF
--- a/crates/goose-server/src/routes/setup.rs
+++ b/crates/goose-server/src/routes/setup.rs
@@ -1,6 +1,7 @@
 use crate::state::AppState;
 use axum::{extract::State, http::StatusCode, routing::post, Json, Router};
 use goose::config::signup_openrouter::OpenRouterAuth;
+use goose::config::signup_tetrate::{configure_tetrate, TetrateAuth};
 use goose::config::{configure_openrouter, Config};
 use serde::Serialize;
 use std::sync::Arc;
@@ -14,6 +15,7 @@ pub struct SetupResponse {
 pub fn routes(state: Arc<AppState>) -> Router {
     Router::new()
         .route("/handle_openrouter", post(start_openrouter_setup))
+        .route("/handle_tetrate", post(start_tetrate_setup))
         .with_state(state)
 }
 
@@ -51,6 +53,48 @@ async fn start_openrouter_setup(
         }
         Err(e) => {
             tracing::error!("OpenRouter setup failed: {}", e);
+            Ok(Json(SetupResponse {
+                success: false,
+                message: format!("Setup failed: {}", e),
+            }))
+        }
+    }
+}
+
+async fn start_tetrate_setup(
+    State(_state): State<Arc<AppState>>,
+) -> Result<Json<SetupResponse>, StatusCode> {
+    tracing::info!("Starting Tetrate Agent Router Service setup flow");
+
+    let mut auth_flow = TetrateAuth::new().map_err(|e| {
+        tracing::error!("Failed to initialize auth flow: {}", e);
+        StatusCode::INTERNAL_SERVER_ERROR
+    })?;
+
+    tracing::info!("Auth flow initialized, starting complete_flow");
+
+    match auth_flow.complete_flow().await {
+        Ok(api_key) => {
+            tracing::info!("Got API key, configuring Tetrate Agent Router Service...");
+
+            let config = Config::global();
+
+            if let Err(e) = configure_tetrate(config, api_key) {
+                tracing::error!("Failed to configure Tetrate Agent Router Service: {}", e);
+                return Ok(Json(SetupResponse {
+                    success: false,
+                    message: format!("Failed to configure Tetrate Agent Router Service: {}", e),
+                }));
+            }
+
+            tracing::info!("Tetrate Agent Router Service setup completed successfully");
+            Ok(Json(SetupResponse {
+                success: true,
+                message: "Tetrate Agent Router Service setup completed successfully".to_string(),
+            }))
+        }
+        Err(e) => {
+            tracing::error!("Tetrate Agent Router Service setup failed: {}", e);
             Ok(Json(SetupResponse {
                 success: false,
                 message: format!("Setup failed: {}", e),
--- a/crates/goose/src/config/signup_tetrate/mod.rs
+++ b/crates/goose/src/config/signup_tetrate/mod.rs
@@ -1,0 +1,174 @@
+pub mod server;
+
+#[cfg(test)]
+mod tests;
+
+use anyhow::{anyhow, Result};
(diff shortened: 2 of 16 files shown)


michaelneale:
ok I pushed a fix to your branch to make it compile (small change on main with api) - but I can't seem to make it work with credits (I think). I can sign on, allw orks great, but I have 0 credit so it fails with an error message if I try to use it. In one case I used github, in another used my gmail - which I guess would be most obvious things people do - is there something else we need to do?
```
Provider request failed with status: 500 Internal Server Error. Payload: Some(Object {"error": String("internalservererror"), "message": String("Internal server error, please try again later")}). Returning error: ServerError("Some(Object {\"error\": String(\"internalservererror\"), \"message\": String(\"Internal server error, please try again later\")})")
```

I assume due to a balance of zero (I can see it in both accounts if I log in)? (it would need a non zero balance for it to work I guess?) - or am I seeing another bug with the flow and it didn't save the token?

crates/goose-server/src/routes/setup.rs
```diff
@@ -14,6 +15,7 @@ pub struct SetupResponse {
 pub fn routes(state: Arc<AppState>) -> Router {
     Router::new()
         .route("/handle_openrouter", post(start_openrouter_setup))
+        .route("/handle_tetrate", post(start_tetrate_setup))
```
michaelneale, line 18: yeah I guess at this point could think of making it generic, but for now I think this is ok, still dont' have a pattern yet

crates/goose/src/config/signup_tetrate/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+use tokio::time::timeout;
+
+/// Default models for Tetrate Agent Router Service configuration
+pub const TETRATE_DEFAULT_MODEL: &str = "claude-4-sonnet-20250514";
+
+// Auth endpoints are on the main web domain
+const TETRATE_AUTH_URL: &str = "https://router.tetrate.ai/auth";
+const TETRATE_TOKEN_URL: &str = "https://router.tetrate.ai/api/api-keys/verify";
+const CALLBACK_URL: &str = "http://localhost:3000";
+const AUTH_TIMEOUT: Duration = Duration::from_secs(180); // 3 minutes
```
michaelneale, line 23: this allows for time for people to sign up? (it is very quick which is great)

crates/goose/src/providers/tetrate.rs
```diff
@@ -0,0 +1,349 @@
+use anyhow::{Error, Result};
```
michaelneale, line 1: I think this makes sense, as there is some tetrate specific handling between models going on here?

michaelneale:
so far - seems ok, follows exactly the same patterns as the openrouter sign up with PKCE

crates/goose/src/config/signup_tetrate/mod.rs
johnlanda: Yep! It should be enough time to do SSO via google or github ideally. I can shorten it as well if you think it is too long.

johnlanda:
> ok I pushed a fix to your branch to make it compile (small change on main with api) - but I can't seem to make it work with credits (I think). I can sign on, allw orks great, but I have 0 credit so it fails with an error message if I try to use it. In one case I used github, in another used my gmail - which I guess would be most obvious things people do - is there something else we need to do?
> 
> ```
> Provider request failed with status: 500 Internal Server Error. Payload: Some(Object {"error": String("internalservererror"), "message": String("Internal server error, please try again later")}). Returning error: ServerError("Some(Object {\"error\": String(\"internalservererror\"), \"message\": String(\"Internal server error, please try again later\")})")
> ```
> 
> I assume due to a balance of zero (I can see it in both accounts if I log in)? (it would need a non zero balance for it to work I guess?) - or am I seeing another bug with the flow and it didn't save the token?

I believe that TARS is set to give $5 credits to business emails at the moment. Let me know if you need me to arrange for some credits to test on the backend.

crates/goose-server/src/routes/setup.rs
johnlanda: Since it is so similar to the openrouter one, I can take a stab at pulling these into a generic route if you like!

crates/goose/src/providers/tetrate.rs
johnlanda: I'm going to simplify this a bit. We don't use the same prefixes and I don't think we need the cache control portion for anthropic either. 

I followed this pattern from the openrouter one as I think we may find some nuances in model handling as we test. If we don't have a need for this, I can pull it out in favor a more generic openai compat provider or something like that.

crates/goose-server/src/routes/setup.rs
michaelneale: for now is ok I think to keep merge hassles low, but up to you!

crates/goose/src/config/signup_tetrate/mod.rs
michaelneale: yeah I think we want to make sure they go via the right path, but even then, it is plenty I think

michaelneale:
thanks @johnlanda looks like a bit of conflict resolving to do still - some deeper changes on main, if you get to it? but looking good.

johnlanda:
@michaelneale I think I have all the conflicts sorted with the latest on main. Let me know if you have a chance to take another look!

michaelneale:
taking a look shortly!

michaelneale:
thanks @johnlanda - if you ask goose to run ./scripts/clippy-lint.sh and fix it - it will take care of those final things.

michaelneale:
oh sometimes to see the test codepath errors I need to run ` cargo test --jobs 2 ` locally

michaelneale:
ok this looks really great. 

Screenshot 2025-08-21 at 2 38 04 pm

could make this the recommended one, especially if it sends people to sign up to get the credits

@johnlanda could change some wording, and make tetrate recommended? Wording could be similar to openrouter "get instant access to AI models, get started for free, automatically configured"? also is there an icon to use?

michaelneale:
ok changed some wording and updated: 

image


cc @spencrmartin this has now another card - which pushes "others" below the fold I believe - any thoughts on that? 
if we remove the `<OpenRouter>` and ollama icons, there is enough space to stack, that is one option for now (as it is now it says to scroll, I fixed that, so that is ok too)

michaelneale (approved):
tested it every which way, and works well. Some follow ons before promoting to make sure people get sent to the right "get free credit" path, but very nice!

PR #4170, opened 2025-08-18
Autocompact System Notifications

katzdave:
https://github.com/user-attachments/assets/aaba7a89-b4c8-46c6-a3b4-374ebbabd3b3


michaelneale:
looks really nice

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -319,15 +321,12 @@ function BaseChatContent({
   };
   // Callback to handle scroll to bottom from ProgressiveMessageList
   const handleScrollToBottom = useCallback(() => {
-    // Only auto-scroll if user is not actively typing
```
michaelneale, line 323: deliberate to remove this?

katzdave: Scrolling seems OK still but I'll test some more.

I was having issues with the new notifications not pushing the scrollbar down and appearing out of view.

PR #4179, opened 2025-08-19
feat: Add message queue system with interruption handling

spencrmartin:
migrated over from: https://github.com/block/goose/pull/4081 - (can see some discussion there). 

image

and with halt rules: 

image

This commit implements a comprehensive message queuing system for ChatInput that allows users to queue messages while the chat is processing responses.

New Features:
- Message Queuing: Users can type and queue multiple messages while chat is loading
- Automatic Processing: Queued messages are automatically sent when chat becomes available
- Queue Persistence: Messages persist in localStorage across sessions
- Interruption Handling: stop and wait commands can interrupt queue processing
- Visual Feedback: Simple queue display shows queued message count and previews

New Components:
- MessageQueue.tsx: UI component for displaying and managing queued messages
- InterruptionHandler.tsx: Handles interruption commands and queue pausing
- Pill.tsx: UI component for queue message display
- queueStorage.ts: Utility for persisting queue to localStorage
- interruptionDetector.ts: Logic for detecting interruption commands

Modified Components:
- ChatInput.tsx: Enhanced with queue functionality while preserving all original features

Technical Implementation:
- Queue state with QueuedMessage interface for type safety
- useEffect monitors loading state changes for automatic processing
- Enter key handling queues messages when isLoading is true
- Clean architecture with minimal changes to existing code

based on prework by https://github.com/block/goose/pull/4081


fix_queue_resume.patch
```diff
@@ -0,0 +1,27 @@
+--- a/ui/desktop/src/components/ChatInput.tsx
```
zanesq: can delete this patch file

michaelneale:
@zanesq @spencrmartin I think this is ready for review, I really liked what I saw when I tested it out by hand. added some unit tests, but it is a bit of a chunky set of additions to `ChatInput.tsx` which warrant some stepping through to make sure it makes sense and isn't over the top.

michaelneale:
tagging @alexhayes-block who may be interested to follow along - with this in future could be a place to hang session forking (ie if you go back to change something)

zanesq:
I noticed a bug where sending the chat command "Stop" isn't fully stopping it from continuing the task. Try running the tamogotchi game example from a fresh chat in home then as its running send "Stop" and it appears to recognize that and try to stop but then it continues on after that and the stop message disappears from the chat history.

michaelneale:
@zanesq will take a look - I wonder if that is goose unable to kill the task vs this code?

zanesq:
@michaelneale maybe not a blocker for this PR since its a hidden feature anyway but it was added here so was testing it

spencrmartin:
Lemme know if you two want to pull this back - we can otherwise we can fire it out to something like an internal team to test or pull tha trigga! @michaelneale @zanesq

michaelneale:
@zanesq yeah I think I am ok with it - seems to stop anything I try, but may be something else, I think can keep an eye on it, and is valuable as is I think.

spencrmartin:
Noting @michaelneale the resume function post a user sending a message after the queue is no longer in.

PR #4180, opened 2025-08-19
Remove dead code and old settings migration

jamadeo:
We had a duplicate extension config. Swapped it for the one from the openapi generated type, and removed some settings migration code that I think is probably good to delete at this point.


alexhancock:
Yeah very good to clean this up. Thanks.

Issue #3244, opened 2025-07-03
Easy session failure reporting


TalSQ:
**Please explain the motivation behind the feature request.**
Does this feature solve a particular problem you have been experiencing? What opportunities or use cases would be unlocked with this feature?

Goose is awesome but it isn't perfect (yet). Sometimes it fails. I want to help Goose devs make Goose (more) awesome.

**Describe the solution you'd like**

It would be great if I could click a button in the GUI and a modal popped up:
Title: "Report a Failure"
Tell us about what happened
<TextArea />
<Submit />
(maybe also a <Cancel/> or an "x" in the top right/left corner depending on user's language)

The text area is an optional input for the user to explain in their own words what happened and any other tips to help point engineers in the right direction to diagnose and fix the failure.

Clicking submit sends the the form data, telemetry and the session log to wherever it needs to go to help Goose developers.

**Describe alternatives you've considered**

I just post in Slack and upload files when applicable. 

I know we can do it all manually, but I think sometimes between the frustration over the failure and then having to find the session log can take the developer out of flow.

I guess I can script a solution outside of Goose but then the easiest way to put that solution into the hands of all goose users is to make it a part of Goose.
Plus, as we encourage non developers to increasingly adopt Goose, we should provide them with the best experience we can.

**Additional context**

- [x] I have verified this does not duplicate an existing feature request

ki3ani:
Hi, can i take on this issue?

TalSQ:
I don't see why not! Thank you :)

zanesq:
Closing this since Goose desktop now has a generate diagnostics bundle icon and sessions have a download option now. Feel free to reopen if this isn't meeting the needs of the original request.

<img width="357" height="122" alt="Image" src="https://github.com/user-attachments/assets/31ce2bf6-fbff-4c7e-965c-2ee0b474a350" />

PR #4183, opened 2025-08-19
Easy session failure reporting

ki3ani:
## Summary
  Implements #3244 - adds "Report a Failure" button with automated system info collection and GitHub Issues integration.

  Users can now report failures directly within the app instead of manually navigating to GitHub. The modal automatically collects system information and submits reports with
  proper error handling and fallback logging.

  ## Screenshots

  ### Before: Settings Help & Feedback Section
  
before

  ### After: New "Report a Failure" Button Added
before-two

  ### Report Failure Modal - System Info Auto-Collection
  
before-three

  ### Success State After Submission
  
before-four

  ## Implementation Details

  - **ReportFailureModal component** with form validation and automatic system info collection
  - **Backend API endpoint** `/report-failure` with GitHub Issues integration and fallback logging
  - **Automatic data collection**: Goose version, platform, architecture, extension count, provider type, recent errors
  - **GitHub integration** creates properly formatted issues matching the bug report template
  - **Fallback logging** ensures reports are captured even when GitHub API is unavailable
  - **Proper error handling** with user feedback for all failure scenarios

  ## Technical Changes

  ### Frontend (`ui/desktop/`)
  - Added `ReportFailureModal.tsx` with system info collection logic
  - Enhanced `AppSettingsSection.tsx` to include the new "Report a Failure" button
  - Extended `preload.ts` to expose `process.arch` for architecture detection

  ### Backend (`crates/goose-server/`)
  - Added `/report-failure` endpoint in `feedback.rs`
  - GitHub Issues API integration with proper authentication
  - Structured logging for manual processing when API unavailable
  - Bug report template formatting matching existing GitHub templates

  ## Test Plan
  - [x] Modal opens and automatically collects system information
  - [x] Form validation prevents empty submissions
  - [x] API endpoint receives and processes failure reports correctly
  - [x] GitHub issue format matches existing bug report template structure
  - [x] Fallback logging works when GitHub token not available
  - [x] Success/error states provide appropriate user feedback
  - [x] System info collection works across different platforms

  ## Notes
  - Requires `GITHUB_TOKEN` environment variable for GitHub Issues integration
  - Falls back to local logging when GitHub API unavailable
  - Architecture detection now properly exposed through Electron API
  - Version detection includes fallback for development builds


jamadeo:
Thanks @ki3ani this is a good feature. However I don't think we should dependon a github token set in goosed -- rather, you can do this more simply by opening the system web browser at something like

https://github.com/block/goose/issues/new?title=YOUR_TITLE&body=YOUR_BODY&labels=LABEL1,LABEL2&milestone=MILESTONE_NAME&assignees=ASSIGNEE1,ASSIGNEE2&template=TEMPLATE_NAME

ki3ani:
> Thanks @ki3ani this is a good feature. However I don't think we should dependon a github token set in goosed -- rather, you can do this more simply by opening the system web browser at something like
> 
> https://github.com/block/goose/issues/new?title=YOUR_TITLE&body=YOUR_BODY&labels=LABEL1,LABEL2&milestone=MILESTONE_NAME&assignees=ASSIGNEE1,ASSIGNEE2&template=TEMPLATE_NAME

That was well noted  @jamadeo,  ive refactored to open pre-filled GitHub URL in browser instead.

crates/goose-server/src/routes/feedback.rs
```diff
@@ -0,0 +1,80 @@
...
+    error!(
+        "FAILURE REPORT (Manual Processing Required): Title: {}, Description: {}, System: {}/{}/{}, Timestamp: {}",
+        payload.title,
+        payload.description,
+        payload.system_info.platform,
+        payload.system_info.architecture,
+        payload.system_info.goose_version,
+        payload.timestamp
+    );
+}
```
jamadeo: Now that it's a pre-filled github URL, I think you can just remove the "backend" component. No need to have it just to log.

ui/desktop/src/components/ReportFailureModal.tsx
```diff
@@ -0,0 +1,348 @@
...
+  const [systemInfo, setSystemInfo] = useState<SystemInfo | null>(null);
+  const [recentErrors, setRecentErrors] = useState<string[]>([]);
+  const [isSubmitting, setIsSubmitting] = useState(false);
+  const [submitStatus, setSubmitStatus] = useState<'idle' | 'success' | 'error'>('idle');
+
+  const getExtensionCount = async (): Promise<number> => {
+    try {
+      // Get extension count from the API
+      const apiUrl = getApiUrl('/extension/list');
+      const secretKey = getSecretKey();
```
jamadeo: I know we do this in other places (we shouldn't) but we should avoid constructing the request like this an instead use the generated SDK: https://github.com/block/goose/blob/99fa6289c485e866e1bb87ce3bbd6cfd44f00f10/ui/desktop/src/api/sdk.gen.ts#L111-L116

ui/desktop/src/components/ReportFailureModal.tsx
```diff
@@ -0,0 +1,348 @@
...
+      console.error('Failed to collect system info:', error);
+    }
+  }, []);
+
+  const collectRecentErrors = useCallback(async () => {
+    try {
+      // For now, collect from browser console errors and any stored logs
+      const recentErrors: string[] = [];
+
+      const storedErrors = localStorage.getItem('goose_recent_errors');
```
jamadeo: is this a thing? do we write to goose_recent_errors anywhere?

ki3ani: You're right, that wasn't actually used anywhere. Removed it.

crates/goose-server/src/routes/feedback.rs
ki3ani: Good point,  i have removed the backend feedback endpoint since it was just logging.

ui/desktop/src/components/ReportFailureModal.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+      setSystemInfo(info);
+    } catch (error) {
+      console.error('Failed to collect system info:', error);
+    }
+  }, []);
+
+  const collectRecentErrors = useCallback(async () => {
+    try {
+      // Recent errors are logged to the main process via ErrorBoundary
```
jamadeo: I'm still confused about the recentErrors bit - it looks like it's only ever set to an array of one constant string? what is this about main process error boundary?

ui/desktop/src/components/ReportFailureModal.tsx
```diff
@@ -0,0 +1,285 @@
...
+  const [systemInfo, setSystemInfo] = useState<SystemInfo | null>(null);
+  const [recentErrors, setRecentErrors] = useState<string[]>([]);
+  const [isSubmitting, setIsSubmitting] = useState(false);
+  const [submitStatus, setSubmitStatus] = useState<'idle' | 'success' | 'error'>('idle');
+
+  const getExtensionCount = async (): Promise<number> => {
+    try {
+      const secretKey = getSecretKey();
+      const response = await getExtensions({
+        headers: secretKey ? { 'X-Secret-Key': secretKey } : {},
```
jamadeo: you should not to include the secret key here, as it's globally configured for the generated client sdk

ui/desktop/src/components/ReportFailureModal.tsx
```diff
@@ -0,0 +1,285 @@
...
+        const providerResponse = await providers({
+          headers: secretKey ? { 'X-Secret-Key': secretKey } : {},
+        });
+
+        if (
+          providerResponse.data &&
+          Array.isArray(providerResponse.data) &&
+          providerResponse.data.length > 0
+        ) {
+          info.providerType = providerResponse.data[0].name || 'Unknown Provider';
```
jamadeo: I don't think you want just the first provider.

The right thing to use here is probably the session metadata for the active session. #4648 adds that to the client API, so this might fit in better once that's merged

DOsinga:
@ki3ani thanks for kicking this off, if you address @jamadeo 's comments, let's get this in!

ui/desktop/src/components/ReportFailureModal.tsx
ki3ani: I removed the old provider listing logic and refactored the code to use getSessionHistory  from the merged #4648  to correctly retrieve the provider_name

ui/desktop/src/components/ReportFailureModal.tsx
ki3ani: I reviewed the mechanism for collecting recent errors. To keep this PR strictly in scope and prevent introducing complex dependencies, I removed the entire collectRecentErrors function and state

DOsinga:
can you also do the DCO? and @jamadeo can you have another look?

ui/desktop/src/components/ReportFailureModal.tsx
```diff
@@ -0,0 +1,269 @@
+import { useState, useEffect, useCallback } from 'react';
+import { Button } from './ui/button';
+import { Textarea } from './ui/textarea';
```
jamadeo: some of these don't exist:

`./ui/textarea`
`./Modal`
`getSessionHistory`

does this run for you?

DOsinga:
also can you look into the DCO thing?

ui/desktop/src/components/ReportFailureModal.tsx
ki3ani: @jamadeo Yes it was running but some imports were broken which I hadn't noticed. Fixed now - replaced ./ui/textarea with native textarea, ./Modal with Dialog components, and
  getSessionHistory with providers() SDK call. All working properly now.

ki3ani:
> also can you look into the DCO thing?

Done

ui/desktop/src/components/ReportFailureModal.tsx
```diff
@@ -0,0 +1,264 @@
...
+        architecture: window.electron?.arch || 'Unknown',
+        extensionCount: await getExtensionCount(),
+      };
+
+      try {
+        const providersResponse = await providers();
+        if (providersResponse.data && Array.isArray(providersResponse.data)) {
+          const providersList = providersResponse.data;
+          if (providersList.length > 0) {
+            info.providerType = providersList[0].name || 'Unknown';
```
jamadeo, line 60: this isn't the right one to use. You can use `useModelAndProvider` to get the current model and provider

ui/desktop/src/components/ReportFailureModal.tsx
```diff
@@ -0,0 +1,264 @@
...
+      window.alert('Please describe what happened before submitting.');
+      return;
+    }
+
+    setIsSubmitting(true);
+    setSubmitStatus('idle');
+
+    try {
+      const issueTitle = encodeURIComponent(
+        `[FAILURE REPORT] ${description.substring(0, 60)}${description.length > 60 ? '...' : ''}`
```
jamadeo, line 117: 60 characters seems quite small for what someone might type into this box. If we're limited by url length, you could leave out the txt entry and have the user type in the opened github window. You'd still have the collected system info

ui/desktop/src/components/settings/app/AppSettingsSection.tsx
```diff
@@ -402,7 +404,16 @@ export default function AppSettingsSection({ scrollToSection }: AppSettingsSecti
...
+          <div className="flex flex-wrap gap-3">
+            <Button
+              onClick={() => setShowReportFailureModal(true)}
+              variant="default"
+              size="sm"
+              className="flex items-center gap-2"
+            >
+              <AlertCircle size={16} />
+              Report a Failure
+            </Button>
```
jamadeo, line 416: I see that this is now next to a "report a bug" button. I think your button is better because it collects system info, so I'd say go ahead and remove the other "report a bug" button so we don't have two very similar actions

ui/desktop/src/components/ReportFailureModal.tsx
jamadeo: er, looks like the extension count is also wrong. When I tried it, it shows 0 extensions installed, but I have several

DOsinga:
closing this since there has been no activity on it. can always reopen

PR #4185, opened 2025-08-19
Move out app init

DOsinga:
This moves the initialization of the app to a separate module and does some clean-up there. It is in preparation of making everything one window - this way we have one choke point to initialize the agent and will protect against merge conflicts in the future


DOsinga:
> Verified working well locally

thanks. it shouldn't make any functional changes :)

PR #4187, opened 2025-08-19
chore: remove the google drive built-in extension

jamadeo:
- deletes the extension itself
- updates the bundled extension syncing to also check for deprecated built-ins


ui/desktop/src/components/settings/extensions/bundled-extensions.ts
```diff
@@ -37,6 +39,13 @@ export async function syncBundledExtensions(
     // Cast the imported JSON data to the expected type
     const bundledExtensions = bundledExtensionsData as BundledExtension[];
 
+    for (let i = existingExtensions.length - 1; i >= 0; i--) {
+      const ext = existingExtensions[i];
+      if (ext.type == 'builtin' && DEPRECATED_BUILTINS.includes(ext.name)) {
+        existingExtensions.splice(i, 1);
```
alexhancock, line 45: this would modify the in memory copy of the existing extensions read from the config, but wouldn't remove the extension from the config file

I assume we want something that does that, right?

jamadeo: You're right, it's still in my config file. I thought this would sync them back to the config but I guess it doesn't? I don't see it in the extensions UI and if I ask Goose to use drive it does not find it. But I suppose if I ran the CLI it would pick it up.

jamadeo: So this doesn't do it perfectly but I think it has the intended effect. It will refuse to load the deprecated extension and then if it sees a 'googledrive' bundled extension, it will be replaced. If not, it will just ignore it. It's not a perfect clean-up of the config but any other solution would mean adding a "migration" hook/step in the config load/save, which would be nice to avoid if possible. This way at least it all lives in the same place.

PR #4195, opened 2025-08-19
docs: openrouter and ollama easy desktop setup

dianed-square:
This PR documents the new OpenRouter and Ollama onboarding options on first launch.

Documentation updates:
- `documentation/docs/getting-started/installation.md` and `documentation/docs/quickstart.md`:
  - Doc new Desktop experience
  - Minor updates for consistency
- `documentation/docs/getting-started/providers.md`: 
  - Update currently recommended model note
- `documentation/src/components/Constants.js`:
  - Renamed RateLimits.js file for general use and added some constants
- `documentation/src/components/LinuxDesktopInstallButtons.js`:
  - Minor update for consistency
- `documentation/docs/assets/guides/set-up-provider-ui.png`: Deleted obsolete image


documentation/src/components/Constants.js
```diff
@@ -0,0 +1,46 @@
+import React from "react";
```
blackgirlbytes: hmm thanks for adding these. interesting pattern/different pattern than im used to. 

So we're already in a components folder, which to my understanding is the equivalent of constants. we may want to put each of these in their own separate file..and import them that way.

For example:

src/components/DesktopProviderSetup.js and it would only have the desktop provider component 

or 
src/components/ModelSelectionTip.js

blackgirlbytes (changes requested):

documentation/docs/getting-started/installation.md
```diff
@@ -79,8 +79,8 @@ import { PanelLeft } from 'lucide-react';
     </Tabs>
   </TabItem>
 
-  <TabItem value="linux" label="Linux" default>
```
blackgirlbytes, line 82: good find

blackgirlbytes (approved):

Issue #4196, opened 2025-08-19
Changing working directory kills current chat/replaces window


zanesq:
**Describe the bug**

Changing the working directory replaces the existing chat killing the current session.

**To Reproduce**
1. Go to chat
2. Click on change directory
3. See existing session is replaced with new session in same window

**Expected behavior**
Instead it should launch a new window but keep the existing one open behind it.

**Screenshots**

**Please provide following information:**
 - **OS & Arch:** Mac Arm
 - **Interface:** UI
 - **Version:** 1.4.0
 - **Extensions enabled:** redacted
 - **Provider & Model:** sonnet

**Additional context**
Apparently it used to keep the current chat open

zanesq:
This doesn't seem to be fixed after all, reopening

DOsinga:
I think thsi works as expected until we move to one agent one session. closing.

PR #4200, opened 2025-08-19
stop replacing chat window when changing working directory

zanesq:
stop replacing chat window when changing working directory

fixes https://github.com/block/goose/issues/4196

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -953,9 +953,7 @@ const buildRecentFilesMenu = () => {
   }));
 };
 
-const openDirectoryDialog = async (
-  replaceWindow: boolean = false
-): Promise<OpenDialogReturnValue> => {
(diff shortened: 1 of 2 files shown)


ui/desktop/src/main.ts
```diff
@@ -1033,45 +1031,8 @@ const openDirectoryDialog = async (
     }
 
     addRecentDir(dirToAdd);
-    const currentWindow = BrowserWindow.getFocusedWindow();
-
-    if (replaceWindow && currentWindow) {
```
zanesq, line 1038: This was left over from when we were attempting to launch in the same window but had to change it last minute. Can bring it back when we work on unifying chat to one window

zanesq:
Definitely agree once we have multiple sessions working in the same window, similar issue the agent needs the working directory on launch currently.

PR #4202, opened 2025-08-19
Goose Simple Compact UX

alexhancock:
Improvements to the context management UX, making it much simpler and more consistent in the language it presents to users

**Changes**

* No more UI in the chat itself other than simple markers that auto-compaction occurred, or the user compacted
* "Compact now" button always in the alert popup about context
* Session automatically continues for the user after the summarization occurs, keeping things moving without mentioning in any verbose way that the agent read a summary

**Demo**

https://github.com/user-attachments/assets/35b4ee00-f21a-4672-a2cf-c3f847e04fde

### DIFF
--- a/crates/goose-cli/src/session/export.rs
+++ b/crates/goose-cli/src/session/export.rs
@@ -347,6 +347,9 @@ pub fn message_to_markdown(message: &Message, export_all_content: bool) -> Strin
                 md.push_str("**Thinking:**\n");
                 md.push_str("> *Thinking was redacted*\n\n");
             }
+            MessageContent::SummarizationRequested(summarization) => {
+                md.push_str(&format!("*{}*\n\n", summarization.msg));
+            }
             _ => {
                 md.push_str(
                     "`WARNING: Message content type could not be rendered to Markdown`\n\n",
--- a/crates/goose-cli/src/session/output.rs
+++ b/crates/goose-cli/src/session/output.rs
@@ -185,6 +185,9 @@ pub fn render_message(message: &Message, debug: bool) {
                 println!("\n{}", style("Thinking:").dim().italic());
                 print_markdown("Thinking was redacted", theme);
             }
+            MessageContent::SummarizationRequested(summarization) => {
+                println!("\n{}", style(&summarization.msg).yellow());
+            }
             _ => {
                 println!("WARNING: Message content type could not be rendered");
             }


zanesq:
@alexhancock just a note to pull main and npm install, had to fix some dependencies from the upgrade

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -242,7 +229,7 @@ function BaseChatContent({
   const { createNewSessionIfNeeded } = useSessionContinuation({
     chat,
     setChat,
-    summarizedThread,
+    summarizedThread: [],
```
alexhancock: @zanesq I wanted to ask you about this one. Can you clarify why the `useSessionContinuation` hook needed the summarized thread before? Likely need to make some kind of change here before going ahead.

alexhancock:
.bundle

crates/goose/src/agents/context.rs
```diff
@@ -84,9 +84,20 @@ impl Agent {
...
+            new_messages.insert(0, compaction_marker);
+            new_token_counts.insert(0, compaction_marker_tokens);
+
+            // Add an assistant message to continue the conversation
+            let assistant_message = Message::assistant().with_text("
+                The previous message contains a summary that was prepared because a context limit was reached.
+                Do not mention that you read a summary or that conversation summarization occurred
+                Just continue the conversation naturally based on the summarized context
+            ");
             let assistant_message_tokens: usize = 14;
```
katzdave: Bump to 41.

crates/goose/src/agents/context.rs
```diff
@@ -84,9 +84,20 @@ impl Agent {
...
+            let compaction_marker = Message::assistant()
+                .with_summarization_requested("Conversation compacted and summarized");
+            let compaction_marker_tokens: usize = 8;
+
+            // Insert the marker before the summary message
+            new_messages.insert(0, compaction_marker);
+            new_token_counts.insert(0, compaction_marker_tokens);
+
+            // Add an assistant message to continue the conversation
+            let assistant_message = Message::assistant().with_text("
```
katzdave, line 96: We have a second message here now? Is the idea that this will become agent visible only, and the other one will be user visible only?

ui/desktop/src/components/BaseChat.tsx
zanesq: It looks like this new architecture has replaced what `useSessionContinuation` was doing so I think we can remove the hook completely now 👍

ui/desktop/src/components/context_management/ContextManager.tsx
```diff
@@ -0,0 +1,181 @@
...
+            return convertApiMessageToFrontendMessage(apiMessage, true, false);
+          }
+
+          // show to model but not user
+          return convertApiMessageToFrontendMessage(apiMessage, false, true);
+        });
+
+        // Store the original messages as ancestor messages so they can still be scrolled to
+        if (setAncestorMessages) {
+          const ancestorMessages = messages.map((msg) => ({
```
katzdave, line 72: Ancestor messages just live in memory in the client and don't survive reload?

katzdave (commented):
Changes look nice overall; a little hard to parse the full diff on some of the file re-names. Taking it for a spin today though!

zanesq:
I noticed the new e2e tests aren't launching the app, are they working for you? Also I just pushed https://github.com/block/goose/pull/4251 to main if you want to pull again so the pre-existing tests are passing.

crates/goose/src/agents/context.rs
alexhancock: > Is the idea that this will become agent visible only, and the other one will be user visible only

yes! to show the user a brief inline notification that compaction occurred, and give the model the deal of what happened and how it should behave when the session continues

ui/desktop/src/components/context_management/ContextManager.tsx
alexhancock: Reload seems to be having trouble now (filed https://github.com/block/goose/issues/4255) but I think it would behave the same as reloading a conversation without compaction. Can check after 4255 is resolved

ui/desktop/src/components/context_management/ContextManager.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+        // Replace messages with the server-provided messages
+        setMessages(convertedMessages);
+
+        // Automatically submit the continuation message to continue the conversation
+        // This should be the third message (index 2) which contains the "I ran into a context length exceeded error..." text
+        const continuationMessage = convertedMessages[2];
+        if (continuationMessage) {
+          setTimeout(() => {
+            append(continuationMessage);
```
alexhancock: TODO here - look into putting the original user message in if context ran out entirely

katzdave:
Noticed one issue where the green dot + context UI goes away right after a tool call. Comes back once goose returns control to the user. Auto-compact and button click flow seem pretty good.

Having a hard time testing the out of context case; seems like GOOSE_CONTEXT_LIMIT doesn't actually override the context in desktop (works on CLI display but also doesn't seem like it properly forces context_limit_exceeded errors); might do a separate fix for that as this is super annoying to test otherwise, but trying to just get a big session file I can drop in with too.

katzdave:
Seems like scrolling is a little flaky too; not sure if related at all to the branch. But the agent loop seems to only force the scrolling sometimes. In particular the first time scrolling after the first user input doesn't seem to work.

zanesq:
> Seems like scrolling is a little flaky too; not sure if related at all to the branch. But the agent loop seems to only force the scrolling sometimes. In particular the first time scrolling after the first user input doesn't seem to work.

Might be better with https://github.com/block/goose/pull/4257

michaelneale:
seems good so far @alexhancock - using it all morning and noothing surprising to me (not sure what I am quite looking for though). I did trigger a manual compact from GUI and did what I expected. Is there something I should look out for? 

Also been using it in cli

michaelneale:
I did get a token overrun though (cli) after a while: 

```
ing("invalid_request_error"), "message": String("prompt is too long: 211596 tokens > 200000 maximum")}, "request_id": String("req_011CSNEjctXWn7P7aKQZq1S5")}). Returning error: ContextLengthExceeded("{\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",\"message\":\"prompt is too long: 211596 tokens > 200000 maximum\"},\"request_id\":\"req_011CSNEjctXWn7P7aKQZq1S5\"}")
    at crates/goose/src/providers/utils.rs:110

Error: Context length exceeded: {"type":"error","error":{"type":"invalid_request_error","message":"prompt is too long: 211596 tokens > 200000 maximum"},"request_id":"req_011CSNEjctXWn7P7aKQZq1S5"}
```

which I have not seen in weeks, so not sure what is going on there.

katzdave:
https://github.com/block/goose/pull/4284 fixes the summarize button in the UI at the end of the context to your error.

Saw some weird stuff rendered, but wasn't sure how to properly suppress that; not the end of the world though as we're targeting fixing this next with message metadata. (the top message is the summary which is also visible).
Screenshot 2025-08-22 at 11 50 14 AM

I also don't see the context window display green dot after resuming a session (but before finishing processing on the first message). I think if we fix the green dot to stop disappearing this should be good to go.

alexhancock:
Updated demo screen capture 

https://github.com/user-attachments/assets/157dc168-bda9-499a-9356-fa0ccad054a9

alexhancock:
> #4284 fixes the summarize button in the UI at the end of the context to your error.
> 
> Saw some weird stuff rendered, but wasn't sure how to properly suppress that; not the end of the world though as we're targeting fixing this next with message metadata. (the top message is the summary which is also visible). Screenshot 2025-08-22 at 11 50 14 AM
> 
> I also don't see the context window display green dot after resuming a session (but before finishing processing on the first message). I think if we fix the green dot to stop disappearing this should be good to go.

I fixed this one

alexhancock:
.bundle

alexhancock:
.bundle

.goose_tmp/diff_vs_main.patch
```diff
@@ -0,0 +1,4032 @@
+diff --git a/crates/goose/src/agents/agent.rs b/crates/goose/src/agents/agent.rs
```
katzdave: remove

.goose_tmp/last_commit.patch
```diff
@@ -0,0 +1,19 @@
+commit eace087d25a96aa1e85db299f7484c335d385e24
```
katzdave: remove

ui/desktop/src/components/ChatInput.tsx
```diff
@@ -511,55 +514,27 @@ export default function ChatInput({
...
-          onSummarize: () => {
-            handleManualCompaction(messages, setMessages);
-          },
-          summarizeIcon: <ScrollText size={12} />,
-        });
-      }
-    } else if (isTokenLimitLoaded && tokenLimit) {
-      // Always show context window info even when no tokens are present (start of conversation)
+    // Only show context window alert if there are tokens being used
+    if ((numTokens && numTokens > 0) || (isTokenLimitLoaded && tokenLimit)) {
```
katzdave, line 518: 🔥

Issue #4184, opened 2025-08-19
Recipe param substitution


alexhancock:
**Describe the bug**

Recipe param substitution sometimes not working

**To Reproduce**
1. Go to https://dev-guides.sqprod.co/goose-recipes
2. Click on Open in Goose
3. Observer modal in Goose and enter values for params
4. See recipe string occasionally show with `{{ param }}` identifiers showing up. Not every time, but sometimes.

Secondarily, we should look at the editing experience of the text that shows up. Editing the text is possible, but deleting it is not. We should come up with an easy to understand set of expectations the user can have about what their options are.

alexhancock:
@zanesq we need to look into this one again. I saw it again today when testing 1.7.0

Let me know if you have any ideas or want to pair on it!

zanesq:
Closing for now. Tested this a bunch with different recipes from deeplinks and from recipe manager and seems to be working ok and replacing params in the initial prompt.

zanesq:
Re-opening, figured out how to repro :)

DOsinga:
is this fixed now? closing because I hope so :)

zanesq:
Yes I haven't seen it in latest

PR #4207, opened 2025-08-20
fix: recipe params not being replaced all the time

zanesq:
fix: recipe params not being replaced all the time

also fixed issue where the initial prompt cant be deleted

fixes https://github.com/block/goose/issues/4184

### DIFF
--- a/ui/desktop/src/components/ChatInput.tsx
+++ b/ui/desktop/src/components/ChatInput.tsx
@@ -222,6 +222,7 @@ export default function ChatInput({
         textAreaRef.current?.focus();
       }, 0);
     }
+    // we don't need hasSetRecipePromptRef in the dependency array because it is a ref that persists across renders
   }, [recipeAccepted, initialPrompt, messages.length]);
 
   // Reset the recipe prompt flag when the recipe changes or messages are added


ui/desktop/src/components/ChatInput.tsx
```diff
@@ -203,17 +203,33 @@ export default function ChatInput({
...
+    ) {
       setDisplayValue(initialPrompt);
       setValue(initialPrompt);
+      hasSetRecipePromptRef.current = true;
       setTimeout(() => {
         textAreaRef.current?.focus();
       }, 0);
     }
-  }, [recipeAccepted, initialPrompt, messages.length, displayValue]);
+  }, [recipeAccepted, initialPrompt, messages.length]);
```
alexhancock, line 226: don't we need `hasSetRecipePromptRef` in the deps here?

ui/desktop/src/utils/providerUtils.ts
```diff
@@ -61,7 +61,7 @@ There may be (but not always) some tools mentioned in the instructions which you
 `;
 
 // Helper function to substitute parameters in text
-const substituteParameters = (text: string, params: Record<string, string>): string => {
+export const substituteParameters = (text: string, params: Record<string, string>): string => {
```
alexhancock, line 64: nice deduping!

ui/desktop/src/components/ChatInput.tsx
zanesq: no its not needed because its a ref object so it doesn't change only the `.current` property changes but changes to `.current` in refs don't trigger re-renders anyway. I added a comment for futuredev

PR #4213, opened 2025-08-20
Add Nix flake for reproducible builds

AnthonyRonning:
This PR adds Nix flake support for reproducible builds of the Goose CLI. The flake provides a consistent build environment across different systems and makes it easy for Nix users to build, develop, and run Goose.

  What's included

  - flake.nix - Main flake configuration that:
    - Dynamically reads package metadata from Cargo.toml files (no hardcoded versions)
    - Builds the goose-cli package with all required dependencies
    - Provides a development shell with cargo tools
    - Configures tests to run with proper HOME/XDG directories
    - Supports both Linux and macOS platforms
  - flake.lock - Pinned dependencies for reproducible builds
  - .gitignore - Added result symlink exclusion

  Testing

  ✅ CLI build and execution tested on NixOS (Linux/aarch64)
  - nix build - Successfully builds the CLI binary
  - nix run - CLI runs and executes commands properly
  - Tests pass with proper environment configuration
  - Development shell (nix develop) works with cargo commands

  ⚠️  Note: Only the CLI has been tested as my NixOS environment doesn't have a desktop environment. The desktop app would need additional testing, but the flake focuses on the CLI package for now.

  Usage
```sh
  # Build the CLI
  nix build

  # Run directly
  nix run . -- --help

  # Development shell
  nix develop
  cargo build -p goose-cli
```

  Future improvements

  - Desktop app support could be added once tested
  - Additional Nix-specific documentation could be added to the main README

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,6 @@ do_not_version/
 
 # Error log artifacts from mcp replay tests
 crates/goose/tests/mcp_replays/*errors.txt
+
+# Nix build output
+result
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,81 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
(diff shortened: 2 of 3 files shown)


michaelneale:
I think this is fine - nix gets mentioned from time to time and from what little I know, it seems amazing and love that people start to work with it, and this doesn't change any existing builds so why not. Thank you @AnthonyRonning !

PR #4216, opened 2025-08-20
Add session to agents

DOsinga:
Add session id to everything and pass those on to the agent


ui/desktop/src/components/ChatInput.tsx
```diff
@@ -841,6 +843,13 @@ export default function ChatInput({
     }
   };
 
+  const canSubmit =
```
michaelneale, line 908: are diffs like this a side effect of things - as it doesnt' seem to have changed

DOsinga: I probably shouldn't have but I saw that this const was the same for the two ways to submit so extracted it

michaelneale:
I think right direction - assuming session id == a new agent, ie open a session, get an agent created and assigned. This is the same as CLI effectively, but without it being at the electron window level (which was also the same effectively as the cli). 

I assume would be some old code to cleanout like `get_state` in goose-server (agent.rs) and the general state.rs in the goose-server crate as well?

DOsinga:
> I think right direction - assuming session id == a new agent, ie open a session, get an agent created and assigned. This is the same as CLI effectively, but without it being at the electron window level (which was also the same effectively as the cli).
> 
> I assume would be some old code to cleanout like `get_state` in goose-server (agent.rs) and the general state.rs in the goose-server crate as well?

yeah that bit is still being worked on. new agent -> goosed gives you a session id and that's what you pass around for state, almost nothing else. continue on session, same but you don't need to create it

crates/goose-server/src/routes/agent.rs
```diff
@@ -42,23 +55,144 @@ pub struct AddSubRecipesResponse {
...
+    Json(payload): Json<StartAgentRequest>,
+) -> Result<Json<StartAgentResponse>, (StatusCode, Json<ErrorResponse>)> {
+    verify_secret_key(&headers, &state).map_err(|_| {
+        (
+            StatusCode::UNAUTHORIZED,
+            Json(ErrorResponse {
+                error: "Unauthorized - Invalid or missing API key".to_string(),
+            }),
+        )
+    })?;
```
jamadeo: do we use verify_secret_key on all handlers? we should use [middleware](https://docs.rs/axum/latest/axum/middleware/index.html) for that if we can

crates/goose-server/src/routes/agent.rs
```diff
@@ -42,23 +55,144 @@ pub struct AddSubRecipesResponse {
...
+
+    let session_id = session::generate_session_id();
+    let counter = state.session_counter.fetch_add(1, Ordering::SeqCst) + 1;
+
+    let metadata = SessionMetadata {
```
jamadeo, line 128: does SessionMetadata::new() not do what we want?

crates/goose-server/src/routes/agent.rs
```diff
@@ -1,26 +1,37 @@
...
 use serde::{Deserialize, Serialize};
+use std::path::PathBuf;
+use std::sync::atomic::Ordering;
 use std::sync::Arc;
+use tracing::error;
 
 #[derive(Deserialize, utoipa::ToSchema)]
 pub struct ExtendPromptRequest {
     extension: String,
+    #[allow(dead_code)]
```
jamadeo, line 32: do you get lint warnings if these attributes are removed? I thought it wouldn't warn because the structs are `pub`

jamadeo (commented):
I should probably keep reading the code but how does this work with backwards compatibility? if I want to resume a session from before this change

crates/goose-server/src/routes/agent.rs
DOsinga: we do not use it on all handlers, but we might as well. the client should always send this over

crates/goose-server/src/routes/agent.rs
DOsinga: I did get a warning, yeah, not sure who said. I think typescript doesn't warn you if something is public, maybe some rust thing alwasy does?

crates/goose-server/src/routes/agent.rs
DOsinga: well, I added a description with a counter to this so we can better see which session is new etc. but you are right, I should just move that to the new method

jamadeo (approved):

crates/goose-server/src/routes/agent.rs
```diff
@@ -76,10 +214,7 @@ async fn add_sub_recipes(
 ) -> Result<Json<AddSubRecipesResponse>, StatusCode> {
     verify_secret_key(&headers, &state)?;
 
-    let agent = state
-        .get_agent()
-        .await
-        .map_err(|_| StatusCode::PRECONDITION_FAILED)?;
+    let agent = state.get_agent().await;
```
michaelneale, line 217: we don't need that PRECONDITION_FAILED? anymore? (not sure what we would do with it anyway)

crates/goose-server/src/routes/recipe.rs
```diff
@@ -88,16 +88,7 @@ async fn create_recipe(
         request.messages.len()
     );
 
-    let error_response = CreateRecipeResponse {
```
michaelneale, line 119: ah, so no need for all these now?

michaelneale:
ouch yeah this is a huge one, some tyre kicking and things seem to work, but there is a bit touched with recipes not really clear how to test/validate (and more than enough code to review). Some potentially nasty conflicts that need to be resolved unfortunately, but seems to work. 
This feels like almost need a code freeze to get such a x-cutting change in?

crates/goose-server/src/state.rs
```diff
@@ -2,43 +2,45 @@ use goose::agents::Agent;
...
             recipe_file_hash_map: Arc::new(Mutex::new(HashMap::new())),
+            session_counter: Arc::new(AtomicUsize::new(0)),
         })
     }
 
-    pub async fn get_agent(&self) -> Result<Arc<Agent>, anyhow::Error> {
-        self.agent
-            .clone()
-            .ok_or_else(|| anyhow::anyhow!("Agent needs to be created first."))
+    pub async fn get_agent(&self) -> AgentRef {
```
wendytang, line 32: Does AppState simply track one agent? Is the plan to refactor AppState to be able to retrieve from multiple agents `get_agent(session_id)`?

crates/goose-server/src/routes/agent.rs
```diff
@@ -42,23 +54,149 @@ pub struct AddSubRecipesResponse {
...
+    )
+)]
+async fn start_agent(
+    State(state): State<Arc<AppState>>,
+    headers: HeaderMap,
+    Json(payload): Json<StartAgentRequest>,
+) -> Result<Json<StartAgentResponse>, StatusCode> {
+    verify_secret_key(&headers, &state)?;
+
+    state.reset().await;
```
wendytang, line 123: Would expect to append the newly started agent to the state rather then resetting/overriding it?

crates/goose-server/src/state.rs
jamadeo: Yes exactly, the plan following this change is to make agent no longer a singleton in goose-server.

crates/goose-server/src/routes/recipe.rs
jamadeo: yeah, Agent is no longer an Option<> on AppState, so these errors are gone

jamadeo:
> ouch yeah this is a huge one, some tyre kicking and things seem to work, but there is a bit touched with recipes not really clear how to test/validate (and more than enough code to review). Some potentially nasty conflicts that need to be resolved unfortunately, but seems to work. This feels like almost need a code freeze to get such a x-cutting change in?

Yeah it's got a broad surface area, but it's a necessary step to get to a point where we can have one backend serving multiple sessions for e.g. tabbed conversations in the UI. It also does a bit more in that it moves some recipe handling to the backend when it previously was all in the UI.

PR #4226, opened 2025-08-20
Remove unused game

DOsinga:
Some easter egg that is no longer reachable?


alexhancock (approved):
the death of whimsy, but I am good with it!

DOsinga:
> the death of whimsy, but I am good with it!

yeah, always said, but I'd rather remove all the codes and add the ability to vibe code games like that as a feature

PR #4228, opened 2025-08-20
Feat: Let providers configure a fast model for summarization

katzdave:


crates/goose/src/providers/base.rs
```diff
@@ -337,6 +337,32 @@ pub trait Provider: Send + Sync {
         tools: &[Tool],
     ) -> Result<(Message, ProviderUsage), ProviderError>;
 
+    /// Generate the next message using a fast/cheaper model when available
```
Kvadratni: can we maybe just have complete take a model as optional so if not provided - we use the default ?
this way we son't need to add 2 methods to every provider?

crates/goose/src/providers/openai.rs
```diff
@@ -160,6 +162,31 @@ impl OpenAiProvider {
             .await?;
         handle_response_openai_compat(response).await
     }
+
+    // Core completion logic that takes a model config
```
DOsinga: as always, no need for this type of comment

crates/goose/src/providers/base.rs
katzdave: My thought was I'd rather not leak model to provider base as it doesn't ever make sense to specify outside of this module, and I don't want someone to try setting it somewhere like summarize.rs

crates/goose/src/providers/openai.rs
```diff
@@ -160,6 +162,31 @@ impl OpenAiProvider {
...
+            .get("usage")
+            .map(get_usage)
+            .unwrap_or_else(|| {
+                tracing::debug!("Failed to get usage data");
+                Usage::default()
+            });
+        let model = get_model(&json_response);
+        emit_debug_trace(model_config, &payload, &json_response, &usage);
+        Ok((message, ProviderUsage::new(model, usage)))
+    }
```
DOsinga: this was previously existing, but now that we pass in the model, no need to fish it out of the json response

crates/goose/src/providers/openai.rs
```diff
@@ -202,21 +229,23 @@ impl Provider for OpenAiProvider {
...
     ) -> Result<(Message, ProviderUsage), ProviderError> {
-        let payload = create_request(&self.model, system, messages, tools, &ImageFormat::OpenAi)?;
+        // Thin wrapper that calls complete_with_model with the configured model
+        self.complete_with_model(&self.model, system, messages, tools)
+            .await
+    }
 
-        let json_response = self.post(&payload).await?;
+    async fn complete_fast(
+        &self,
```
DOsinga: this should be on provider, not on openai; any provider that has a fast_model should automatically use it.

crates/goose/src/providers/openai.rs
```diff
@@ -202,21 +229,23 @@ impl Provider for OpenAiProvider {
         messages: &[Message],
         tools: &[Tool],
     ) -> Result<(Message, ProviderUsage), ProviderError> {
-        let payload = create_request(&self.model, system, messages, tools, &ImageFormat::OpenAi)?;
+        // Thin wrapper that calls complete_with_model with the configured model
```
DOsinga: huh?

crates/goose/src/providers/openai.rs
```diff
@@ -202,21 +229,23 @@ impl Provider for OpenAiProvider {
...
-        let json_response = self.post(&payload).await?;
+    async fn complete_fast(
+        &self,
+        system: &str,
+        messages: &[Message],
+        tools: &[Tool],
+    ) -> Result<(Message, ProviderUsage), ProviderError> {
+        // Use the fast model (gpt-4o-mini) for fast completions
+        let mut fast_config = self.model.clone();
+        fast_config.model_name = OPEN_AI_FAST_MODEL.to_string();
```
DOsinga: make this a method on config - config.for_fast_model() should return a copy of itself with the model_name swapped out for the smart model if that is available

DOsinga (approved):

crates/goose/src/providers/anthropic.rs
```diff
@@ -71,7 +74,13 @@ impl AnthropicProvider {
         })
     }
 
-    pub fn from_custom_config(model: ModelConfig, config: CustomProviderConfig) -> Result<Self> {
+    pub fn from_custom_config(
+        mut model: ModelConfig,
+        config: CustomProviderConfig,
+    ) -> Result<Self> {
+        // Set the default fast model for Anthropic
+        model.fast_model = Some("claude-3-5-haiku-latest".to_string());
```
DOsinga: you don't want to do this - custom_config means it is a provider that speaks anthropic but isn't. they will not have this model

crates/goose/src/providers/anthropic.rs
```diff
@@ -49,7 +49,10 @@ pub struct AnthropicProvider {
 impl_provider_default!(AnthropicProvider);
 
 impl AnthropicProvider {
-    pub fn from_env(model: ModelConfig) -> Result<Self> {
+    pub fn from_env(mut model: ModelConfig) -> Result<Self> {
+        // Set the default fast model for Anthropic
+        model.fast_model = Some("claude-3-5-haiku-latest".to_string());
```
DOsinga: make that into a constant on top - ideally we'd do this in a more elegant way, but let's do that later.  don't make this mutable though, but have something like model = model.with_fast(CONSTANT) - possibly also check whether that is already set and don't overwrite it

crates/goose/src/providers/base.rs
```diff
@@ -317,6 +317,29 @@ pub trait Provider: Send + Sync {
...
+    /// * `system` - The system prompt that guides the model's behavior
+    /// * `messages` - The conversation history as a sequence of messages
+    /// * `tools` - Optional list of tools the model can use
+    ///
+    /// # Returns
+    /// A tuple containing the model's response message and provider usage statistics
+    ///
+    /// # Errors
+    /// ProviderError
+    ///   - It's important to raise ContextLengthExceeded correctly since agent handles it
```
DOsinga: what are these comments? they repeat the existing comments which are now out of date demonstrating why we shouldn't have them in the first place

crates/goose/src/providers/base.rs
```diff
@@ -317,6 +317,29 @@ pub trait Provider: Send + Sync {
...
+    async fn complete_with_model(
+        &self,
+        model: &str,
```
DOsinga: this should be modelConfig, not model. if you pass a string but then you need model config, you're going to create one yourself

crates/goose/src/providers/bedrock.rs
```diff
@@ -152,11 +152,12 @@ impl Provider for BedrockProvider {
...
 
     #[tracing::instrument(
-        skip(self, system, messages, tools),
+        skip(self, _model, system, messages, tools),
         fields(model_config, input, output, input_tokens, output_tokens, total_tokens)
     )]
-    async fn complete(
+    async fn complete_with_model(
         &self,
+        _model: &str,
```
DOsinga: now we are lying

crates/goose/src/providers/anthropic.rs
katzdave: done + applied to openAI/google

crates/goose/src/providers/anthropic.rs
katzdave: ack. Done + applied to openAI/google.

crates/goose/src/providers/base.rs
katzdave: done

crates/goose/src/providers/base.rs
katzdave: Done

crates/goose/src/providers/bedrock.rs
katzdave: Done.

crates/goose/src/providers/openai.rs
katzdave: done

crates/goose/src/providers/openai.rs
katzdave: Got rid of it, but confused by this one since we already had the model name from before the request?

PR #4230, opened 2025-08-20
[Feature][i18n] Add Brazilian Portuguese localization support (Issue …

sachinschitre:
…#2376)

   - Integrated go-i18n for message translation
   - Added PT-BR JSON files for sample messages
   - Support for --lang flag to select locale
   - Includes unit tests verifying localization


DOsinga:
hi - this seems to be about temporal only. can you do the localization there instead? closing for now, but thanks!

Issue #4118, opened 2025-08-15
Recipes with extensions take a long time to load


zanesq:
**Describe the bug**

Loading a recipe with extensions shows home but hangs there while it loads the recipe extensions.

We should show a "loading extensions" screen and also look into speeding up loading extensions on the backend.

**To Reproduce**
1. Load a recipe deeplink that has extensions
2. See app launch with home/hub while it loads
3. Wait a long time
4. Shows recipe in chat page

**Expected behavior**
It should load faster if extensions also it should indicate its processing/loading the recipe.

zanesq:
Tagging @michaelneale in case any insight into speeding up loading extensions in the backend?

patelkhushal18:
Thanks for creating the issue, also wanted to mention that the recipe was loading faster in older versions of goose (version prior to July 20th)

michaelneale:
@patelkhushal18 @zanesq do you have any example recipes I can try this with? (not familiar with this form of recipes of late, have not kept up!) that can try and profile why it is slow (@zanesq also it should be faster as parallel loading, but there were other startup changes). 

@lifeizhou-ap may have some ideas or things to try too?

patelkhushal18:
Hi @michaelneale , I think this was caused by me not adding `args` to the snowflake extension in my recipe. I was missing `args: ['mcp_snowflake@latest'],` in my recipe, it works with this:
```
  extensions: [
    {
      name: 'snowflake',
      args: ['mcp_snowflake@latest'],
      bundled: true,
      cmd: 'uvx',
      description: 'Integration to query Snowflake data',
      timeout: 2000,
      type: 'stdio',
      tables: [...],
    },
```

When I don't have the `args` in the latest goose version, goose window opens up but the recipe is never loaded

michaelneale:
ah ok - so that may have bveen a different one, I will have a play around and see what I can find, thanks!

michaelneale:
this may be helped by: https://github.com/block/goose/pull/4234

michaelneale:
I am a bit lost as to how to make a deeplink recipe with extensions - not sure of the official way (I can hack base64 but surely this isn't what everyone does). Anyone have a link to share @patelkhushal18 ? would love to load your one to see how it performs but I can't follow how to make one correctly for a deeplink or desktop

michaelneale:
@zanesq can't reproduce - but not really sure what a recipe with extensions looks like in desktop (unable to get one working in any case).

michaelneale:
[...]

PR #4234, opened 2025-08-21
Start extensions concurrently 

jamadeo:
Removes the mutex around the extension manager, letting us actually start them concurrently. Also adds some output in the CLI.

Since #4054 we are enabling them concurrently from the frontend, but we still had the lock around the manager itself, so the speedup was limited by this.

https://github.com/user-attachments/assets/dc6cef3f-3f8c-477a-a671-fb7ba0c34090


zanesq:
nice good catch!

michaelneale:
surely this will help https://github.com/block/goose/issues/4118 a lot

michaelneale (approved):
woah, if I tested correctly this is night and day, and make things work nicely, lovely @jamadeo

PR #4240, opened 2025-08-21
doc: uvx not npx

michaelneale:
blender was configured to use npx, but it is a python mcp!


blackgirlbytes (approved):
thank you for catching that!

PR #4249, opened 2025-08-21
chore: removing little-used session sharing feature

alexhancock:
Discussed this recently as a group. This feature is very little used and seemingly almost entirely unused in OSS circles. Removing this implementation, and may replace soon with a simple implementation that can share a `.jsonl` file via system share menu.


ui/desktop/src/main.ts
```diff
@@ -298,7 +298,7 @@ async function processProtocolUrl(parsedUrl: URL, window: BrowserWindow) {
   if (parsedUrl.hostname === 'extension') {
     window.webContents.send('add-extension', pendingDeepLink);
   } else if (parsedUrl.hostname === 'sessions') {
-    window.webContents.send('open-shared-session', pendingDeepLink);
+    // Shared session functionality has been removed
```
jamadeo: can remove condition completely here and below

ui/desktop/src/App.tsx
```diff
@@ -321,44 +318,11 @@ const WelcomeRoute = () => {
...
-            console.error('Failed to retry loading shared session:', error);
-          } finally {
-            setIsLoadingSharedSession(false);
-          }
-        }
-      }}
-    />
+    <div className="flex justify-center items-center py-12">
+      <div className="text-text-muted">Shared session functionality has been removed</div>
+    </div>
```
jamadeo: Do we need this?

ui/desktop/src/App.tsx
```diff
@@ -1016,11 +937,7 @@ export default function App() {
                   path="shared-session"
                   element={
                     <ProviderGuard>
-                      <SharedSessionRouteWrapper
-                        isLoadingSharedSession={isLoadingSharedSession}
-                        setIsLoadingSharedSession={setIsLoadingSharedSession}
-                        sharedSessionError={sharedSessionError}
-                      />
+                      <SharedSessionRouteWrapper />
```
jamadeo: I think the whole route can go?

ui/desktop/src/App.tsx
alexhancock: Nope! Just had missed it

PR #4250, opened 2025-08-21
Add X-Title and referer headers on exchange to tetrate

johnlanda:
Adds headers which TARS can use to identify goose clients during the PKCE flow


michaelneale (approved):
nice, ready to go when you are

PR #4257, opened 2025-08-21
Stop auto scrolling when agent responds and let scroll area handle scrolling to bottom

zanesq:
Improves the scrolling behavior during chat to stop auto scrolling to the bottom after an agent response. Also lets the scroll area handle scrolling to bottom rather than the message list.

Also potentially fixes the occasional jumpiness when the agent is responding since its no longer fighting various scroll to bottom mechanisms.


alexhancock (approved):
yep! this has been an issue recently and just from reading the code I think this will be better.

michaelneale:
nice, much tidier code too!

PR #4260, opened 2025-08-21
fix issue where app redirects to home after initialization but user has already started a chat

zanesq:
fix issue where app redirects to home after initialization but user has already started a chat

### DIFF
--- a/ui/desktop/src/App.test.tsx
+++ b/ui/desktop/src/App.test.tsx
@@ -15,6 +15,7 @@ Object.defineProperty(window, 'location', {
     search: '',
     href: 'http://localhost:3000',
     origin: 'http://localhost:3000',
+    pathname: '/',
   },
   writable: true,
 });
@@ -196,6 +197,7 @@ describe('App Component - Brand New State', () => {
     vi.clearAllMocks();
     window.location.hash = '';
     window.location.search = '';
+    window.location.pathname = '/';
     window.sessionStorage.clear();
     window.localStorage.clear();
   });


DOsinga:
can you explain what exactly the issue is here? in my largish refactor I am trying to throw this all out and rethink from the beginning, but I understand there are some "interesting" patterns.

zanesq:
Basically this was the issue where you'd be chatting and it would redirect back to hub/home.. the initialization wasn't complete yet before starting chat from pair so it would try to load the home page when its done. Definitely fixes it, can repro every time if you start a quick chat from home right when the app launches. I put in other fixes but it didn't catch this.

amed-xyz (approved):

ui/desktop/src/utils/appInitialization.ts
```diff
@@ -95,8 +95,18 @@ export const initializeApp = async ({
...
+  // This prevents redirecting users who have already navigated elsewhere during initialization
+  const currentHash = window.location.hash;
+  const currentPathname = window.location.pathname;
+  const isOnRootRoute =
+    currentPathname === '/' && (!currentHash || currentHash === '#' || currentHash === '#/');
+
+  if (isOnRootRoute) {
+    window.location.hash = '#/';
+    window.history.replaceState({}, '', '#/');
+  }
```
DOsinga, line 109: I dont understand this code and comment though - it seems like it says, redirect to root, but only if we are already on root?

zanesq: It's the fallback route to home, but we only want to trigger it if the user hasn't already navigated somewhere else by the time it hits this.

Issue #4264, opened 2025-08-21
Goose CLI not showing projects as a command


the-destro:
Version 1.5.0. The only configuration I have done is point it at AWS bedrock. 
<img width="588" height="353" alt="Image" src="https://github.com/user-attachments/assets/7058073a-f90c-4a7b-8f19-1e35e6fc94b3" />

**To Reproduce**

**Expected behavior**

**Screenshots**

**Please provide following information:**
 - **OS & Arch:** [e.g. Ubuntu 22.04 x86]
 - **Interface:** [UI/CLI]
 - **Version:** [e.g. v1.0.2]
 - **Extensions enabled:** [e.g. Computer Controller, Figma]
 - **Provider & Model:** [e.g. Google - gemini-1.5-pro]

**Additional context**

zanesq:
Sorry about that, looks like it was accidentally removed in a cleanup [PR](https://github.com/block/goose/pull/4074). Was meant to be only the UI related projects will re-add 👍 

For now can you use a previous version?

PR #4266, opened 2025-08-21
restore cli projects from accidental removal during cleanup

zanesq:
restore cli projects from accidental removal during cleanup in https://github.com/block/goose/pull/4074, tested and verified projects are working locally as before

fixes https://github.com/block/goose/issues/4264


zanesq:
I meant to only remove the ui version but apparently its in use :) lets get some telemetry before we just flat out remove it

PR #4270, opened 2025-08-22
UI: query provider for supported models

angelahning:
The UI queries the providers for the list of supported models, instead of using our hard-coded lists.

fixes https://github.com/block/goose/issues/3645

https://github.com/user-attachments/assets/b4afb433-d838-42a6-a5e8-f4fd31a019e5


ui/desktop/src/components/ConfigContext.tsx
```diff
@@ -185,6 +187,21 @@ export const ConfigProvider: React.FC<ConfigProviderProps> = ({ children }) => {
...
 
+  const getProviderModels = useCallback(async (providerName: string): Promise<string[]> => {
+    try {
+      const response = await apiGetProviderModels({
+        path: { name: providerName },
+        headers: {
+          'X-Secret-Key': await window.electron.getSecretKey(),
+        },
+      });
+      return response.data || [];
```
DOsinga, line 198: this should use the generated API. you should never have to call getSecretKey

ui/desktop/src/components/settings/models/subcomponents/AddModelModal.tsx
```diff
@@ -141,24 +141,60 @@ export const AddModelModal = ({ onClose, setView }: AddModelModalProps) => {
...
+            return { provider: p, models, error: null };
+          } catch (e: unknown) {
+            return {
+              provider: p,
+              models: null,
+              error: `Failed to fetch models for ${providerName}${e instanceof Error ? `: ${e.message}` : ''}`,
+            };
+          }
+        });
+        const results = await Promise.all(modelPromises);
```
DOsinga, line 164: this doesn't seem ideal; if we wanted to get all the models for all the providers, we should just make that a flag on the getProviders method; the server can do that much more efficient.

but really, since we already split this up into selecting a provider first and then list the models, why not call this just in time? we have 28 providers, so that would avoid this mini DOS attack on our server.

ui/desktop/src/components/settings/models/subcomponents/AddModelModal.tsx
```diff
@@ -141,24 +141,60 @@ export const AddModelModal = ({ onClose, setView }: AddModelModalProps) => {
...
+        // Process results and build grouped options
+        const groupedOptions: { options: { value: string; label: string; provider: string }[] }[] =
+          [];
+        const errors: string[] = [];
+
+        results.forEach(({ provider: p, models, error }) => {
+          if (error) {
+            errors.push(error);
+            // Fallback to metadata known_models on error
+            if (p.metadata.known_models && p.metadata.known_models.length > 0) {
```
DOsinga, line 175: why do we have this fall back twice? also I don't think known_models is optional here, so we shouldn't need to check all this

ui/desktop/src/components/settings/models/subcomponents/AddModelModal.tsx
```diff
@@ -141,24 +141,60 @@ export const AddModelModal = ({ onClose, setView }: AddModelModalProps) => {
...
+            groupedOptions.push({
+              options: models.map((m) => ({ value: m, label: m, provider: p.name })),
             });
           }
         });
 
+        // Log errors if any providers failed (don't show to user)
+        if (errors.length > 0) {
+          console.error('Provider model fetch errors:', errors);
+        }
```
DOsinga, line 194: why do we have this complex machinery to keep track of the errors, only to log them here and then do nothing about them?

ui/desktop/src/components/settings/models/subcomponents/AddModelModal.tsx
```diff
@@ -141,24 +141,60 @@ export const AddModelModal = ({ onClose, setView }: AddModelModalProps) => {
...
+                options: p.metadata.known_models.map(({ name }) => ({
+                  value: name,
+                  label: name,
+                  provider: p.name,
+                })),
+              });
+            }
+          } else if (models && models.length > 0) {
+            groupedOptions.push({
+              options: models.map((m) => ({ value: m, label: m, provider: p.name })),
```
DOsinga, line 186: dupe from 176

PR #4271, opened 2025-08-22
Block send until extensions are ready

jamadeo:
We now don't block user interaction while the agent initializes, which is great! But we should still block sending your first message until the extensions are ready, else the user might end up with something unexpected.


michaelneale:
nice - desk checked it, seems ok, used https://kaleidoscope.app/ to see how focussed and small the change really is, nice catch/follow on @jamadeo !

jamadeo:
Thanks @zanesq @michaelneale , I'll merge this but I actually think ideally we would queue instead of block here. But that's a bit of a larger change, we can do that another time.

PR #4272, opened 2025-08-22
feat: initial implementation of extension malware check

shellz-n-stuff:
## Description
Goose users need to install extensions in order to get the full utility they desire. However, our stdio MCPs are quite often coming from package ecosystems that can contain malware.

Though it's not possible to completely protect users from all forms of Malware (and other Supply Chain Attacks) implementing some basic checks for known bad packages better protects all Goose users! ❤️ 

## Implementation
When a user attempts to activate an extension we will parse out the name of the package and/or version information (where possible) and attempt to query the [OSV API](https://google.github.io/osv.dev/api/) to understand if the package contains malware (identified by MAL in the formatting).

If the http call fails for some reason and/or we have issues parsing we opt to fail open to ensure this doesn't present a disruption to users.

*Note: This implementation only supports PyPi and NPM as package manager formats*

### Handling of non-provided package versions
One key choice here was we @latest syntax and no version provided syntax by checking if any version of the package has malware. Broadly speaking it's going to be rare that an extension is compromised, malware finding filed and then persisted on those versions but it "could" happen and I'm open to adjusting the approach if we have major upfront concerns  

## OSV/Malicious Packages
The OSV project contains vulnerability information and malware information for open-source packages. The Malware information is populated via https://github.com/ossf/malicious-packages and broadly is a trustworthy datasource.

Though OSV does contain vulnerability data we're strictly only checking for Malware in this case. We could potentially inform users of CVEs in the future using this interface but we'd want to make the product interface there much easier to understand for all users!

## Testing
Screenshot 2025-08-22 at 1 33 52 pm

- Adding innocuous package that will be flagged as Malware

Screenshot 2025-08-22 at 1 34 23 pm

- Error is thrown and available in UI

Screenshot 2025-08-22 at 1 34 40 pm

- Clear error message

### DIFF
--- a/crates/goose/src/agents/extension_malware_check.rs
+++ b/crates/goose/src/agents/extension_malware_check.rs
@@ -277,6 +277,7 @@ fn http_client() -> Result<reqwest::Client, ExtensionError> {
     );
     reqwest::Client::builder()
         .default_headers(headers)
+        .timeout(std::time::Duration::from_secs(10))
         .build()
         .map_err(|e| ExtensionError::SetupError(format!("failed to build HTTP client: {e}")))
 }


crates/goose/src/agents/extension_malware_check.rs
```diff
@@ -0,0 +1,511 @@
...
+
+    debug!(name, ecosystem, ?version, "OSV: no MAL advisories.");
+    Ok(())
+}
+
+fn http_client() -> Result<reqwest::Client, ExtensionError> {
+    let mut headers = HeaderMap::new();
+    headers.insert(
+        USER_AGENT,
+        HeaderValue::from_static("goose-osv-check/1.1 (+https://osv.dev)"),
```
shellz-n-stuff, line 276: This was an idea from Goose. But I actually really liked it! Just in-case the OSV team want to identify us and/or hit us up

frankgraziano (approved):

crates/goose/src/agents/extension_malware_check.rs
```diff
@@ -0,0 +1,511 @@
...
+    );
+    reqwest::Client::builder()
```
tomasz-sq, line 278: Thoughts on adding reasonable timeouts to prevent hanging on slow/unresponsive OSV API?

crates/goose/src/agents/extension_malware_check.rs
```diff
@@ -0,0 +1,511 @@
...
+                    v.id
+                } else {
+                    format!("{} — {}", v.id, v.summary)
+                }
+            })
+            .collect::<Vec<_>>()
+            .join("; ");
+        error!(name, ecosystem, version=%ver, %details, "Blocked malicious package via OSV MAL-*.");
+        return Err(ExtensionError::ConfigError(format!(
+            "Blocked malicious package: {name}@{ver} ({ecosystem}). OSV MAL advisories: {details}"
```
tomasz-sq, line 263: should we consider making the error more actionable?  I was curious how to deal with false positives. Looks like the recommended way is to create a GH issue 
https://github.com/ossf/malicious-packages/blob/main/README.md#false-positives

crates/goose/src/agents/extension_malware_check.rs
michaelneale: yeah I think we need a timeout and a test for it - also sometimes people run in places where there isn't always direct access to everything

michaelneale:
I think this is ok - will need some fmt done and I think needs a timeout, but we can try it.

crates/goose/src/agents/extension_malware_check.rs
michaelneale: message could say to "contact your adminstrator" which ... I hate when I see that as it is often me!

blackgirlbytes:
Awesome! Did you mean to pull in the other commits to this branch?

shellz-n-stuff:
> Awesome! Did you mean to pull in the other commits to this branch?

Bad rebase. Fixing it 🙏

PR #4281, opened 2025-08-22
feat(ui): Adding a run recipe button to look for local recipes to run

alexrrouse:


zanesq:
@spencrmartin ok with the design?

michaelneale:
is that where recipes are saved by default if you click on recipes in the side bar?

alexrrouse:
> is that where recipes are saved by default if you click on recipes in the side bar?

No, this would be a separate location.  My goal here is to allow running/sharing recipes that are local to a directory that you are working in.  Our project has a bunch of different frameworks and while some recipes could be shared across all, there is also a benefit of having specific targeted recipes with additional context for that specific framework.  This button looks at the directory that you have open in goose and if there are any recipes there, then it will show them, it doesn't use the global location that the sidebar uses.

spencrmartin:
@alexrrouse Taking a look, might bother you to jump on a call to see your workflow. The bottom bar might be somewhat protected real estate but that's just an initial reaction. Will reach out!

DOsinga:
we think this will be done in a different way, so closing for now, but feel free to reopen if not. thanks!

PR #4282, opened 2025-08-22
Remove half-second wait, rework auto submit

jamadeo:
Do the auto-submit using props instead of imperatively so we don't need to wait.

### DIFF
--- a/ui/desktop/src/components/ChatInput.tsx
+++ b/ui/desktop/src/components/ChatInput.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState, useEffect, useMemo } from 'react';
+import React, { useRef, useState, useEffect, useMemo, useCallback } from 'react';
 import { FolderKey, ScrollText } from 'lucide-react';
 import { Tooltip, TooltipContent, TooltipTrigger } from './ui/Tooltip';
 import { Button } from './ui/button';
@@ -84,6 +84,7 @@ interface ChatInputProps {
   recipeConfig?: Recipe | null;
   recipeAccepted?: boolean;
   initialPrompt?: string;
+  autoSubmit: boolean;
 }
 
 export default function ChatInput({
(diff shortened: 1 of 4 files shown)


ui/desktop/src/components/ChatInput.tsx
```diff
@@ -936,69 +942,89 @@ export default function ChatInput({
...
+      displayValue,
+      droppedFiles.length,
+      handleSubmit,
+      lastInterruption,
+      localDroppedFiles.length,
+      onFilesProcessed,
+      pastedImages,
+      setLocalDroppedFiles,
+    ]
+  );
```
DOsinga, line 1020: what happened here? did we not get warnings about this?

ui/desktop/src/components/ChatInput.tsx
```diff
@@ -367,7 +370,10 @@ export default function ChatInput({
   } = useFileDrop();
 
   // Merge local dropped files with parent dropped files
-  const allDroppedFiles = [...droppedFiles, ...localDroppedFiles];
+  const allDroppedFiles = useMemo(
+    () => [...droppedFiles, ...localDroppedFiles],
+    [droppedFiles, localDroppedFiles]
+  );
```
DOsinga, line 376: does this help?

ui/desktop/src/components/ChatInput.tsx
jamadeo: The key change here is running performSubmit() in a useEffect, but in order to do that I wrapped that in a useCallback, and that meant enumerating its dependencies.

We could have gotten away with keeping this a function re-created on every render, since the submit effect should only actually do something just once anyway, but we'd have to suppress some warnings and potentially invite some future bugs. Using hooks for all of them seemed the right way to go

ui/desktop/src/components/ChatInput.tsx
jamadeo: it's a dependency of a hook now, so we don't want to make a new list every time

PR #4286, opened 2025-08-22
Remove a bit from pre-commit that husky says we should

jamadeo:
When I made a commit, the husky pre-commit hook warned me to delete these lines.


jamadeo:
> Thanks! Hopefully the hook still works after this change?

Yep, tested by introducing type/lint errors and running `git commit` from both ui/desktop and repo root.

PR #4290, opened 2025-08-22
Use gemini flash for summarization on open router

katzdave:


michaelneale (approved):
looks ok - would be better experience I think?

PR #4292, opened 2025-08-22
Read oltp config from config and env

katzdave:


crates/goose-cli/src/main.rs
```diff
@@ -9,8 +9,13 @@ async fn main() -> Result<()> {
 
     let result = cli().await;
 
-    // Only wait for telemetry flush if OTLP is configured
-    if std::env::var("OTEL_EXPORTER_OTLP_ENDPOINT").is_ok() {
+    // Only wait for telemetry flush if OTLP is configured (check both env and config)
+    let should_wait = std::env::var("OTEL_EXPORTER_OTLP_ENDPOINT").is_ok()
+        || goose::config::Config::global()
+            .get_param::<String>("otel_exporter_otlp_endpoint")
+            .is_ok();
```
DOsinga: why do we check the environment here still?

katzdave: done

Issue #4221, opened 2025-08-20
Litellm provider using openai model name defaults


myaple:
**Describe the bug**

The litellm provider uses crates/goose/src/providers/formats/openai.rs as it's create_request (https://github.com/block/goose/blob/main/crates/goose/src/providers/litellm.rs#L170), but this create_request function does very specific openai processing, like here: https://github.com/block/goose/blob/main/crates/goose/src/providers/formats/openai.rs#L580

This means that if my litellm proxy has a model name that starts with "o", like "open-mistral-small-3.1", then it will send a reasoning_effort parameter regardless of whether or not it's a reasoning model.

**To Reproduce**
1. Stand up litellm with a model that starts with "o"
2. Send a request through goose
3. See that it sends a reasoning_effort parameter regardless of model

**Expected behavior**
Reasoning parameters should only be sent to be models that support reasoning. Additionally, there should be an environment variable to override the reasoning parameter for the openai models or disable it completely.

**Please provide following information:**
 - **OS & Arch:** ubuntu 24
 - **Interface:** cli
 - **Version:** 1.5.0
 - **Extensions enabled:** n/a
 - **Provider & Model:** litellm, open-mistral-small-3.1

myaple:
I'm more than happy to give this issue a shot if you guys would like. My initial idea would be to add a parameter to the create_request function that takes in a reasoning effort override of high, medium, low, or disabled, and sets the reasoning effort appropriately without doing the model based parsing. This could then be passed in as a configuration parameter like GOOSE_REASONING_EFFORT if you'd like it done at a global level, or LITELLM_REASONING_EFFORT for provider level. 

We could also create a new create_request specifically for litellm, but I think it'd be easier in the long run to leave the openai format alone until otherwise necessary rather than duplicate code.

michaelneale:
thanks @myaple - added a comment on the PR. Agree ideally could just be minor tweaks to do reasoning or not, even a list of big models. I think general idea is not pass in reasoning_effort other than exceptional cases you would say? Longer term I think reasoning like this goes away, as the models work it out themselves with gating, routing and mixtures of experts (like gpt-5 basically does now, but I guess there is some space in api to be specific, but I think that may be short lived)

DOsinga:
should be fixed after #5084

myaple:
[...]

PR #4294, opened 2025-08-22
fix: modifies openai request logic for reasoning models (#4221)

myaple:
This is a first take at fixing the issue described in #4221 where the logic used in the litellm provider to create the requests is the same as the openai format, which has significant bias towards openai models. I've modified the name-based parsing to use more specific schemes, calling out o{1..4} instead of just "o" as the starting char.

### DIFF
--- a/crates/goose/src/providers/formats/openai.rs
+++ b/crates/goose/src/providers/formats/openai.rs
@@ -606,10 +606,13 @@ pub fn create_request(
         ));
     }
 
-    let is_ox_model =
-        model_config.model_name.starts_with("o") || model_config.model_name.starts_with("gpt-5");
+    let is_ox_model = model_config.model_name.starts_with("o1")
+        || model_config.model_name.starts_with("o2")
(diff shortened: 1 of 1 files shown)


michaelneale (changes requested):
@myaple thanks matt - looks like some errors and something missing from it, but seems like a resonable approach (not sure we do want another variable though, if can be avoided)

DOsinga:
I agree with @michaelneale here. let's cut this down to just the change in the pattern recognition. no need for the test or variable. and let's get this merged!

myaple:
Sounds good, I'll rip the rest out this weekend and get it across the line.

DOsinga (approved):

michaelneale (approved):

DOsinga:
so sorry, but I think you'll need to sync to main to get this to kick off the tests? we had a bug in main at some point

myaple:
no worries, let me know if that fixes it

PR #4296, opened 2025-08-22
fix refresh crashing app

### LINKED ISSUE #4255: Reload Doesn't Work
Reload isn't working at the moment. Errors show in the console

1. `Failed to set localStorage after retry - continuing without localStorage config`
2. A navigation request to `/pair` which has no protocol, so electron puts `file://` and it fails

zanesq:
Made the fallbacks for checking local storage more resilient. Also added a backup mechanism for navigation since electron can have issues with hash based routing on load.

- added exponential backoff retries to avoid accessing local storage before its available
- added direct call to react router navigation with fallback to hash routing to avoid protocol errors in electron

fixes https://github.com/block/goose/issues/4255


ui/desktop/src/App.tsx
```diff
@@ -373,50 +395,59 @@ export default function App() {
...
-        window.location.hash = '#/welcome';
-        break;
-      default:
-        console.error(`Unknown view: ${view}, not navigating anywhere. This is likely a bug.`);
-        console.trace('Invalid setView call stack:');
-        // Don't navigate anywhere for unknown views to avoid unexpected redirects
-        break;
+
+    // Use React Router navigation if available, otherwise fallback to hash manipulation
+    if (navigateRef.current) {
```
alexhancock, line 407: When is it not available?

ui/desktop/src/main.ts
```diff
@@ -659,25 +659,42 @@ const createChat = async (
       .executeJavaScript(
         `
```
alexhancock: A bit outside the scope of this PR, but why do we need to manually call `executeJavaScript` on this block?

Could we not move it to some renderer side ts?

zanesq:
@alexhancock I can see how this would be confusing so I added comments to both of these. Let me know if it makes sense or not.

ui/desktop/src/App.tsx
```diff
@@ -712,6 +751,11 @@ export default function App() {
             onCancel={dismissModal}
             isSubmitting={modalState.isPending}
           />
+          <NavigateCapture
+            onNavigateReady={(navigate) => {
+              navigateRef.current = navigate;
+            }}
+          />
```
DOsinga, line 758: you'll have to explain to me why we need this.

ui/desktop/src/main.ts
```diff
@@ -655,35 +656,19 @@ const createChat = async (
...
+    // during Electron renderer initialization, especially on slower systems or during heavy load.
+    // We use executeJavaScript from the main process because:
+    // 1. The main process needs to pass configuration data to the renderer process
+    // 2. This happens during window initialization before the renderer's React app is ready
+    // 3. The timing is critical - we need to set config before React components try to read it
+    // 4. Direct IPC communication would require the renderer to be fully loaded first
+    const injectionScript = localStorageInjectionScript('gooseConfig', configStr);
+    mainWindow.webContents.executeJavaScript(injectionScript).catch((error) => {
+      console.error('Failed to execute localStorage script:', error);
+    });
```
DOsinga, line 671: why again do we want this to be injected into local storage? we seem to be loading this with:

```
  getConfig: () => {
    // Add fallback to localStorage if config from preload is empty or missing
    if (!config || Object.keys(config).length === 0) {
      try {
        if (window.localStorage) {
          const storedConfig = localStorage.getItem('gooseConfig');
          if (storedConfig) {
            return JSON.parse(storedConfig);
          }
        }
      } catch (e) {
        console.warn('Failed to parse stored config from localStorage:', e);
      }
    }
    return config;
```

couldn't we store it locally in the renderer and when it can't be found just do an ipc call back to the electron processs? or expose it using exposeInMainWorld that I just learned about?

DOsinga (commented):
we should maybe discuss this some more, this does feel like we're trying to fix stuff on top of shaky stuff maybe?

ui/desktop/src/App.tsx
zanesq: Basically `setView` was being called before react router was ready during direct navigation or refresh so now it will capture it and use it once its ready. We prefer to use react router navigation anyway so it should improve things. That being said I can leave this part out for now and only do the local storage changes to see if that is enough to fix the refresh issue.

ui/desktop/src/main.ts
zanesq: Agreed this can probably be improved with some `exposeInMainWorld` logic but unfortunately would require more time so I think we should come back to that in a bigger refactor. IPC would require async calls that happen further down the chain and could cause race conditions. This is just a way to get it called as early as possible and have retries to fix the main error that localstorage isn't available to the browser yet when it tries to access it.

As for why we put config in local storage I'm guessing its so it can be loaded faster from multiple windows but it may be a legacy thing that we could come back to and remove also.

zanesq:
Appreciate the feedback! Closing this in favor of a more targeted fix https://github.com/block/goose/pull/4355

Issue #4125, opened 2025-08-15
The developer extension should respect cancellation messages, killing any running processes


jamadeo:
If the user hits "stop" during in-flight tool calls, we'll send a [cancellation notification][0] through the MCP client:

https://github.com/block/goose/blob/b0009117830b05d8db7bc2a31b446591e1114e9a/crates/mcp-client/src/client.rs#L206

On the other side though, our developer MCP extension does not handle this. We should have it do so. When the developer extension gets a cancellation message, it should kill any running processes that it owns.

The "right" way to make this happen is probably to move the developer extension over to an rmcp implementation.

[0]: https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/cancellation

PR #4297, opened 2025-08-22
Rewrite the developer mcp using the rmcp sdk

angelahning:
https://github.com/block/goose/issues/4125

This is part 1 of moving the developer extension to using the official rmcp sdk instead of our custom sdk. All the functionalities of the og developer tools have been reimplemented (thanks AI) along with all the existing tests. 

Tested the new server with the mcp inspector and the tools are working
`npx --registry https://registry.npmjs.org @modelcontextprotocol/inspector cargo run -p goose-server --bin goosed -- mcp rmcp_developer`

This is not used by anything yet. Second PR incoming, where I actually swap the client.


crates/goose-server/src/commands/mcp.rs
```diff
@@ -14,6 +15,15 @@ pub async fn run(name: &str) -> Result<()> {
     }
 
     tracing::info!("Starting MCP server");
+
+    if name == "rmcp_developer" {
```
jamadeo: there's some similar mcp startup code in goose-cli (we should de-duplicate that at some point)

jamadeo:
Nice! I hadn't thought of implementing it as a "new" extension, but this does make it easy to run side-by-side. One thing we could do here is add some cases to the [mcp integration tests](https://github.com/block/goose/blob/4ef3746855f1a8d1690144ac512659de62744aef/crates/goose/tests/mcp_integration_test.rs) for the old and new implementation and compare the recorded interaction.

One risk that we'll need to watch out for is someone else making changes to developer that don't get flagged as conflicts and instead end up silently dropped though.

angelahning:
> Nice! I hadn't thought of implementing it as a "new" extension, but this does make it easy to run side-by-side. One thing we could do here is add some cases to the [mcp integration tests](https://github.com/block/goose/blob/4ef3746855f1a8d1690144ac512659de62744aef/crates/goose/tests/mcp_integration_test.rs) for the old and new implementation and compare the recorded interaction.
> 
> One risk that we'll need to watch out for is someone else making changes to developer that don't get flagged as conflicts and instead end up silently dropped though.

added integration tests! 

yeah agreed that is a risk. But I thought it'd be easier to re-write it instead of modifying this 4000 line file in place... Hopefully there won't be too many changes til I swap the clients.

DOsinga:
> yeah agreed that is a risk. But I thought it'd be easier to re-write it instead of modifying this 4000 line file in place... Hopefully there won't be too many changes til I swap the clients.

FWIW I have a small change pending on how we do the prompt for windows vs non windows - extracting the communalities and also pointing out to the agent to use / even in windows since C:\new\table will mostly be converted to something with a newline and a tab in it. Let's keep that in mind

crates/goose/tests/mcp_integration_test.rs
```diff
@@ -49,6 +49,74 @@ enum TestMode {
     ],
     vec![]
 )]
+#[test_case(
+    vec!["goose", "mcp", "developer"],
+    vec![
+        ToolCall::new("text_editor", json!({
+            "command": "view",
+            "path": "/Users/aning/goose/crates/goose/tests/tmp/goose.txt"
```
jamadeo: probably don't want to use these paths -- you could have the tool first create a file and operate on that

crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -0,0 +1,1817 @@
...
+        let instructions = if hints.is_empty() {
+            base_instructions
+        } else {
+            format!("{base_instructions}\n{hints}")
+        };
+
+        ServerInfo {
+            capabilities: ServerCapabilities::builder().enable_tools().build(),
+            instructions: Some(instructions),
+            ..Default::default()
```
jamadeo, line 353: nit: currently it's returning "rmcp" as the server name, so this should probably also have a 

```
server_info: Implementation {
  name: "goose-developer",
  version: env!("CARGO_PKG_VERSION").to_owned()
}
```

crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -0,0 +1,1817 @@
...
+
+        let output_task = tokio::spawn(async move {
+            let mut combined_output = String::new();
+
+            // We have the individual two streams above, now merge them into one unified stream of
+            // an enum. ref https://blog.yoshuawuyts.com/futures-concurrency-3
+            let stdout = SplitStream::new(stdout.split(b'\n')).map(|v| ("stdout", v));
+            let stderr = SplitStream::new(stderr.split(b'\n')).map(|v| ("stderr", v));
+            let mut merged = stdout.merge(stderr);
+
```
jamadeo, line 906: this seems to have lost the output streaming we did using notifications that the prior tool had: 

https://github.com/block/goose/blob/9b9e6d91ea3c314ca61cd6888d4dfbb3da3e2a2a/crates/goose-mcp/src/developer/mod.rs#L645-L661

for this you'd need a reference to the Peer (client). I'm not sure if the `#[tool]` annotated functions provide that, but it should be possible using `call_tool`

shawn111:
@angelahning since you are working on refactor mcp, could you use latest version of xcap?

Once use latest version of xcap, android/termux users can cargo build goose without extra patches.

crates/goose-mcp/src/developer/rmcp_developer.rs
angelahning: @jamadeo I added streaming notifications back! see below

crates/goose-server/src/commands/mcp.rs
```diff
@@ -16,8 +17,19 @@ pub async fn run(name: &str) -> Result<()> {
     }
 
     tracing::info!("Starting MCP server");
+
+    if name == "developer" {
```
michaelneale, line 21: ??? why is this one different now? this seems a lot more complicated than the old router?

michaelneale:
hrm - there are quite a lot of changes in here, main one is how it handles errors is quite different and will break functionality. Also the DeveloperRouter -> DeveloperServer - is that a change that has to be repeated everywhere?

crates/goose-cli/src/commands/mcp.rs
```diff
@@ -27,8 +28,19 @@ pub async fn run_server(name: &str) -> Result<()> {
 
     tracing::info!("Starting MCP server");
 
+    if name == "developer" {
```
michaelneale, line 31: seems a lot uglier than before...

crates/goose-mcp/Cargo.toml
```diff
@@ -13,7 +13,7 @@ workspace = true
 [dependencies]
 mcp-core = { path = "../mcp-core" }
 mcp-server = { path = "../mcp-server" }
-rmcp = { workspace = true }
+rmcp = { version = "0.6.0", features = ["server", "client", "transport-io", "macros"] }
```
michaelneale, line 16: this should be set at parent, not here

PR #4301, opened 2025-08-22
change default approval mode to smart approval rather than auto

### LINKED ISSUE #2806: Autonomous mode by default is dangerous
I've downloaded Goose for the first time, connected it to a local model, and explored it. I noticed the "Autonomous" indicator and decided to find out what it meant. The manual does explain the permissions and warns that Autonomous is the default.

You can't expect every new user to first read all of the manual before doing anything, then finding this warning and changing the setting. Instead, new users will most likely leave this set at its default, making it possible for things to go horribly wrong, as the only other safeguards are in the prompting. See my example below using a local model and a simple prompt.

**Expected behavior**
Either:
- Chat only or manual as the default; or
- a modal dialog on startup warning of the danger.

**Screenshots**

Here's an example of the kind of command that would be happily executed if I'd let it. These kinds of commands can have subtle bugs or misunderstandings and you don't know because users can use even very stupid models and prompts.

Image

 - **OS & Arch:** MacOS 15.5
 - **Interface:** UI
 - **Version:** v1.0.24
 - **Extensions enabled:** Developer (by default!)
 - **Provider & Model:** Local OpenAI-like via LM-Studio, the example below using `josiefied-qwen3-8b-abliterated-v1` which is an uncensored model, something other users might have as well.

**Additional context**
I'm asking you, the developers of Goose, to think of the liability you take on with such an insecure default.

Other agnetic tools use docker containers, which is another way to make this kind of tool use less of a security problem.

zanesq:
change default approval mode to smart approval rather than auto for safety

closes https://github.com/block/goose/issues/2806

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -854,8 +854,9 @@ impl Session {
                     self.run_mode = RunMode::Normal;
                     // set goose mode: auto if that isn't already the case
                     let config = Config::global();
-                    let curr_goose_mode =
-                        config.get_param("GOOSE_MODE").unwrap_or("auto".to_string());
+                    let curr_goose_mode = config
+                        .get_param("GOOSE_MODE")
+                        .unwrap_or("smart_approve".to_string());
                     if curr_goose_mode != "auto" {
                         config
                             .set_param("GOOSE_MODE", Value::String("auto".to_string()))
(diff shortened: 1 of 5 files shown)


DOsinga:
This seems like a major change! I think we should discuss this a bit more TBH

zanesq:
Closing until we land on something with https://github.com/block/goose/discussions/4324

PR #4303, opened 2025-08-23
Fix (Ollama provider): Unsupported operation: streaming not implemented

### LINKED ISSUE #3527: Feature Request: Implement Streaming for Ollama Models
### **Please explain the motivation behind the feature request.**

The primary motivation is to **improve the user experience and responsiveness for users interacting with Ollama models through Goose.** Currently, Goose's Ollama integration waits for the entire model response to be generated before displaying any output.

This synchronous behavior leads to:

* **Noticeable Delays:** Users experience frustrating waits for full responses, especially with longer generations.
* **Poor Interactivity:** It hinders real-time feedback and makes interactive sessions less fluid.
* **Perceived Unresponsiveness:** The application can feel sluggish, as opposed to modern LLM interfaces that stream output as it's generated.

Implementing streaming would enable:

* **Enhanced Real-time Interaction:** Providing immediate feedback and a more dynamic experience.
* **Improved User Perception:** Making Goose feel faster and more responsive when using Ollama.
* **Alignment with Modern LLM Practices:** Bringing the Ollama integration in line with industry standards for generative AI.

This enhancement will directly benefit users who have selected Ollama as their provider within Goose, making their AI interactions much more efficient and enjoyable.

### **Describe the solution you'd like**

Update the Goose Ollama integration (`ollama.rs`) to support **streaming responses** from the Ollama API.

This involves:

1.  Including `stream: true` in the Ollama API request.
2.  Modifying the `complete` function to process the API's response as a stream of server-sent events (SSE).
3.  Incrementally processing and displaying data chunks as they arrive, rather than waiting for the full response.

### **Describe alternatives you've considered**

No viable alternatives exist for achieving true streaming without modifying the core `ollama.rs` logic. Pseudo-streaming via polling or using an external proxy are inefficient and add unnecessary complexity, failing to deliver a genuine real-time experience.

Jovonni:
I'm not entirely sure if this is the right way to fix this issue, but here's what I ran into when using the goose desktop app installed from the site (`v1.6.0`):

Screenshot 2025-08-22 at 10 50 09 PM

implements streaming capability for ollama chat completions by:
  - adding stream method to ollamaprovider
  - using openai-compatible streaming format with stream=true
  - handling streamed responses line by line with proper error handling
  - including usage metrics in streamed responses
  - maintaining compatibility with existing chat/auto modes
  
# note
- this also occurs for all of my local ollama models
- not sure if this also closes #3527

### DIFF
--- a/crates/goose/src/providers/ollama.rs
+++ b/crates/goose/src/providers/ollama.rs
@@ -8,7 +8,9 @@ use crate::conversation::message::Message;
 use crate::conversation::Conversation;
 use crate::impl_provider_default;
 use crate::model::ModelConfig;
-use crate::providers::formats::openai::{create_request, get_usage, response_to_message, response_to_streaming_message};
+use crate::providers::formats::openai::{
+    create_request, get_usage, response_to_message, response_to_streaming_message,
+};
 use crate::utils::safe_truncate;
 use anyhow::Result;
 use async_stream::try_stream;
@@ -85,7 +87,7 @@ impl OllamaProvider {
         Ok(Self {
             api_client,
             model,
-            supports_streaming: false,
+            supports_streaming: true,
         })
     }
 
@@ -242,15 +244,11 @@ impl Provider for OllamaProvider {
         messages: &[Message],
         tools: &[Tool],
     ) -> Result<MessageStream, ProviderError> {
-        let config = crate::config::Config::global();
-        let goose_mode = config.get_param("GOOSE_MODE").unwrap_or("auto".to_string());
-        let filtered_tools = if goose_mode == "chat" { &[] } else { tools };
-
         let mut payload = create_request(
             &self.model,
             system,
             messages,
-            filtered_tools,
+            tools,
             &super::utils::ImageFormat::OpenAi,
         )?;
         payload["stream"] = json!(true);


got-root:
Looks like this may fix this one: https://github.com/block/goose/issues/4308

crates/goose/src/providers/ollama.rs
```diff
@@ -228,6 +235,49 @@ impl Provider for OllamaProvider {
...
+    async fn stream(
+        &self,
+        system: &str,
+        messages: &[Message],
+        tools: &[Tool],
+    ) -> Result<MessageStream, ProviderError> {
+        let config = crate::config::Config::global();
+        let goose_mode = config.get_param("GOOSE_MODE").unwrap_or("auto".to_string());
+        let filtered_tools = if goose_mode == "chat" { &[] } else { tools };
+
```
DOsinga: this seems unexpected. by the time we get here, we should have already have filtered the tools. do we really need this?

michaelneale: yeah I don't think so, not sure where that came from

michaelneale:
thanks @Jovonni  - that works once I patched it up: https://github.com/block/goose/pull/4364 - if you want, you can pull that change in to this branch and I can approve/merge, or we can shift over to there, has your commits. 

can you also sign the DCO: https://github.com/block/goose/pull/4303/checks?check_run_id=48963447829 so we can get past that 

you can `get cherry-pick b70930d86fd8b0c141fbf4bfcbd08c4e6de5b73a` and and `git cherry-pick ba63988c6a3a1844e359dd58119271d8dbe4a78a` (or something like that) and push here - and should be good to go once you sign the DCO. 

Nice first contrib, this feels a lot nicer with ollama now it can stream.

Jovonni:
just signed dco @michaelneale 

https://github.com/block/goose/pull/4303/commits/e9e8a463821f5c02be6202c6fba05b1a42f87a67

sorry for the delay, have been heads down on my side.

Jovonni:
@michaelneale  I also just cherry picked your work (from https://github.com/block/goose/pull/4364) as you mentioned above. Glad to hear this was useful!!  🚀 🫡

crates/goose/src/providers/ollama.rs
```diff
@@ -228,6 +237,45 @@ impl Provider for OllamaProvider {
...
+            let framed = FramedRead::new(stream_reader, LinesCodec::new()).map_err(anyhow::Error::from);
+            let message_stream = response_to_streaming_message(framed);
+            pin!(message_stream);
+            while let Some(message) = message_stream.next().await {
+                let (message, usage) = message.map_err(|e| ProviderError::RequestFailed(format!("Stream decode error: {}", e)))?;
+                super::utils::emit_debug_trace(&model_config, &payload, &message, &usage.as_ref().map(|f| f.usage).unwrap_or_default());
+                yield (message, usage);
+            }
+        }))
+    }
```
DOsinga, line 278: left this comment in the other PR too - this looks very similar to the openai implementation. can we extract this and reuse?

michaelneale: @DOsinga yeah good idea - I am not familiar with the pattern on how to reuse parts of another provider, can ollama just be openai provider with some customisation? what is another examplar to follow (this is an older provider so may have some jank/special cases)

michaelneale:
@Jovonni if you get a chance to consider the comment above - if there is a way we can re-use the openai streaming would be nice, I may get a chance later today or tomorrow to look, but if not, we can probably merge this soon

Issue #4275, opened 2025-08-22
Extension List Overflows Modal in Manual/Smart Chat Mode


Abhijay007:
**Describe the bug**

When configuring chat mode, if we switch to either Manual or Smart settings and then browse the extension list, the UI breaks and starts overflowing outside of the modal

**To Reproduce**
1. Go to settings 
2. Click on chat
3. In mode, click on either manual or smart setting 
4. See extension rule list 

**Expected behavior**

It should display list and have a scroll bar to scroll through it without overflowing outside of the modal

**Screenshots**

When configuring chat mode, if we switch to either Manual or Smart settings and then browse the extension list, the UI breaks and starts overflowing outside of the modal

**Please provide the following information:**
 - **OS & Arch:** Windows 
 - **Interface:** UI
 - **Version:** 1.5.0
 - **Extensions enabled:** N/A
 - **Provider & Model:** : Google - gemini-2.5-flash

Abhijay007:
I can work on this one, if it is relevant now and not fixed by the team in any of the unsubmitted PRs 🙂

zanesq:
Please feel free!

PR #4305, opened 2025-08-23
fix: fixed chat mode extension list modal scroll

Abhijay007:
Closes : #4275

**Description:** 

Problem: The modal was cutting off content, and the scroll wasn't working properly.
Root Cause: The custom ScrollArea component wasn't working well with the Radix UI Dialog component's overflow handling.

**Changes made:** 

Replaced custom ScrollArea with native browser scrolling using `overflow-y-auto`

**Preview:**  (ignore differ styling as it is on Windows local env)

extensionMenu


alexhancock:
Yes, very good. Thank you!

PR #4311, opened 2025-08-24
Align Dynamic Task Interface with Recipe Interface

tlongwell-block:
## Summary

This PR unifies the dynamic task and subrecipe systems by treating dynamic tasks as inline recipes, enabling agents to leverage the full power of the recipe infrastructure without requiring recipe files.

### 1. Unified Task Format
Dynamic tasks now accept all recipe fields (instructions, prompt, extensions, settings, retry, response schema, etc.), providing agents with ad-hoc access to recipe capabilities previously only available through recipe files.

### 2. Extension Control
Agents can now precisely control subagent extensions:
- **Omit field**: Use all current extensions (backward compatible)
- **Empty array `[]`**: No extensions (sandboxed execution)
- **Array with names**: Only specified extensions

This addresses a critical bug where subagents always loaded all enabled extensions regardless of task specification.

### 3. Output Control
Added `return_last_only` flag allowing agents to receive only the final message from subagents instead of the full conversation, reducing token usage and improving efficiency when intermediate steps aren't needed.

### Performance
- Reduced token consumption in calling agent with selective output
- Lower overhead by loading only required extensions
- Efficient subagent execution with precise resource control

### Safety
- Sandboxed execution with no extensions when needed
- Explicit extension control prevents unintended capabilities
- Security validation inherited from recipe infrastructure

### Power
- Full recipe features available without file creation
- Model/provider settings per task

## Implementation Details

The implementation leverages existing recipe validation and building infrastructure, requiring minimal new code (~300 lines). Dynamic tasks are converted to inline recipes internally, maintaining backward compatibility while providing forward-looking capabilities.

### Technical Approach
- Extended `TaskConfig` to include extensions field
- Modified `SubAgent::new()` to respect specified extensions
- Enhanced dynamic task tool to accept all recipe fields
- Reused existing recipe validation and security checks

## Testing

- Exhaustive tests covering all new functionality
- Backward compatibility verified
- Extension control edge cases tested
- Output control functionality validated

## Breaking Changes

None. All existing dynamic task usage continues to work unchanged.

## Migration Path

Existing code requires no changes. New features are opt-in through additional fields in the dynamic task parameters.


crates/goose/src/prompts/system.md
```diff
@@ -55,6 +55,25 @@ Template:
 - [ ] Blocked: waiting on credentials
 ```
 
+Deploy subagents (via dynamic task tool) for discrete, context-independent tasks. Use extension filters to limit resource access. This prevents context pollution, enables parallelization, provides sandboxes for testing uncertain approaches. Ideal when tasks share minimal context and benefit from isolation or concurrent execution.
```
DOsinga: should this be part of the system prompt? tool related prompts should be part of the tool prompting I think

crates/goose/src/prompts/system.md
```diff
@@ -42,7 +42,7 @@ No extensions are defined. You should let the user know that they should add ext
 
 - Required — use `todo__read` and `todo__write` for any task with 2+ steps, multiple files/components, or uncertain scope. Skipping them is an error.
 - Start — `todo__read`, then `todo__write` a brief checklist (Markdown checkboxes).
-- During — after each major action, update via `todo__write`: mark done, add/edit items, note blockers/dependencies.
+- During — after each major action, read again via `todo__read` and then update via `todo__write`: mark done, add/edit items, note blockers/dependencies.
```
DOsinga: this does not seem like part of this PR

crates/goose/src/agents/subagent_execution_tool/tasks.rs
```diff
@@ -46,48 +46,41 @@ async fn get_task_result(
...
-            task_execution_tracker,
-            cancellation_token,
-        )
-        .await?;
-
-        if success {
-            process_output(stdout_output)
-        } else {
-            Err(format!("Command failed:\n{}", &stderr_output))
+    match task.task_type.as_str() {
```
DOsinga: we should change task_type to an enum

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
```diff
@@ -106,25 +242,103 @@ fn create_task_execution_payload(tasks: Vec<Task>, execution_mode: ExecutionMode
...
+    // Convert each parameter set to inline recipe and create tasks
+    let mut tasks = Vec::new();
+    for task_param in &task_params_array {
+        // Check if this is a legacy text_instruction task
+        if task_param.get("text_instruction").is_some()
+            && task_param.get("instructions").is_none()
+            && task_param.get("prompt").is_none()
+        {
+            // Legacy path for backward compatibility
+            let text_instruction = task_param
```
DOsinga: what is the legacy path here? can we just kill this?

crates/goose/src/prompts/system.md
tlongwell-block: 100%

I'm just using this branch for benchmarking atm since this pr is still just a draft... so it was convenient to make this adjustment here temporarily

crates/goose/src/prompts/system.md
tlongwell-block: I _think_ so. This is more how the agent should behave rather than instructions on the usage of a particular tool?

DOsinga: We do want subagents to live in an extension that we can also switch off, at which point these instructions become weird

tlongwell-block: I think we can get the best of both worlds by using templating `{% if extension.has_resources %}` to dynamically add/remove this from the system prompt

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
tlongwell-block: Yes, I'll take this branch out. It was intended to allow the tool to be used the exact same way as before, but that doesn't actually help agents that don't have hardcoded workflows

tlongwell-block: Done

crates/goose/src/agents/subagent_execution_tool/tasks.rs
tlongwell-block: Done

crates/goose/src/prompts/system.md
tlongwell-block: Looked at this a bit more. Once we have a flag for disabling subagents (in another PR), then we can easily bubble that up here. I'll resolve this conversation since we don't have that flag/feature yet

tlongwell-block:
> Sorry, didn't really get through this or really to the meat of the thing, looks promising so far, questions so far mostly about whether we are having premature backwards compatibility stuff

I've now removed all the backwards compatibility branches around text instructions

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
```diff
@@ -15,58 +22,213 @@ pub const DYNAMIC_TASK_TOOL_NAME_PREFIX: &str = "dynamic_task__create_task";
...
+    // Handle extensions
+    if let Some(extensions) = task_param.get("extensions") {
+        if let Some(ext_configs) = process_extensions(extensions, loaded_extensions) {
+            builder = builder.extensions(ext_configs);
+        }
+    }
+
+    // Handle other optional fields
+    if let Some(settings) = task_param.get("settings") {
+        if let Ok(settings_obj) = serde_json::from_value::<Settings>(settings.clone()) {
```
yingjiehe-xyz: can we do something like 

```
    fn apply_if_ok<T: serde::de::DeserializeOwned>(
        builder: RecipeBuilder,
        value: Option<&Value>,
        f: impl FnOnce(RecipeBuilder, T) -> RecipeBuilder,
    ) -> RecipeBuilder {
        match value.and_then(|v| serde_json::from_value(v.clone()).ok()) {
            Some(parsed) => f(builder, parsed),
            None => builder,
        }
    }

    builder = apply_if_ok(builder, task_param.get("settings"), RecipeBuilder::settings);
    builder = apply_if_ok(builder, task_param.get("response"), RecipeBuilder::response);
    builder = apply_if_ok(builder, task_param.get("retry"), RecipeBuilder::retry);
    builder = apply_if_ok(builder, task_param.get("context"), RecipeBuilder::context);
...
```

yingjiehe-xyz:
overall it looks good, would like to see how others feel

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
tlongwell-block: I think this is a good idea. Let me push this up

tlongwell-block: Done. Thanks! https://github.com/block/goose/pull/4311/files#diff-0cf4132123a6807736d595b641823589eafc49f43d111200a14cf3daf54ed84dR175

michaelneale:
@tlongwell-block is this really making sub agents ... clearer? simpler? https://github.com/block/goose/pull/4421 - is my change still relevant here? (ie ideally would be combined with the one moving the tool out of the main system prompt?)

tlongwell-block:
> @tlongwell-block is this really making sub agents ... clearer? simpler? #4421 - is my change still relevant here? (ie ideally would be combined with the one moving the tool out of the main system prompt?)

It makes them simpler for us to reason about: they just run recipes now, provided ad-hoc by agents. It also makes them much more powerful and efficient.

It is a step towards a unified execution backend for Goose https://github.com/block/goose/discussions/4389

The system prompt changes in this PR net us a few good points in ability on agentic benchmarks. That said... I don't necessarily love making the system prompt longer or more opinionated. Though it does work well. (Towards that end, I'm getting rid of the task management/todo guidance entirely from the system prompt at @DOsinga 's urging https://github.com/block/goose/pull/4454)

michaelneale:
@tlongwell-block main thing missing for dynamic task/subagents to me was that it wasn't aggressive enough in getting them to activate, when I think to be competitive that may be required?

tlongwell-block:
> @tlongwell-block main thing missing for dynamic task/subagents to me was that it wasn't aggressive enough in getting them to activate, when I think to be competitive that may be required?

Yes, absolutely. Two parts here that make this PR work better ime and on the benches:

* the new system prompt addition actually gets the agent to use subagents. It mentions using them twice.
* the agent can now actually have its subagents *only return their last message*. This massively saves context for the calling agent and makes using *lots* of subagents practical

crates/goose/src/prompts/system.md
```diff
@@ -55,6 +55,26 @@ Template:
 - [ ] Blocked: waiting on credentials
 ```
 
+Deploy subagents (via dynamic task tool) for discrete, context-independent tasks. Use extension filters to limit resource access. This prevents context pollution, enables parallelization, provides sandboxes for testing uncertain approaches. Ideal when tasks share minimal context and benefit from isolation or concurrent execution.
+
+## Robust Implementation Practices
```
michaelneale: this reads like it is very specific to developer type modes, vs a base system for an agent? (ie mentions vim -c and such). Would ie be better moved to there? (or some of it) - eg debugging etc.

michaelneale: (as we get a lot of comments about making goose in some modes less biased to be a developer only agent)

crates/goose-cli/src/session/output.rs
```diff
@@ -439,10 +439,36 @@ fn render_dynamic_task_request(call: &ToolCall, debug: bool) {
...
+                                }
+                            }
+                        }
+                        Value::Object(_) => {
+                            // For objects, print them with proper indentation
+                            println!("        {}:", style(key).dim());
+                            print_params(value, 2, debug);
+                        }
                         _ => {
-                            // For everything else, use print_params
```
lifeizhou-ap, line 443: the case here are the same as "Value::String(s)" above.  maybe we can remove the "Value::String(s)" case above

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
```diff
@@ -15,58 +21,203 @@ pub const DYNAMIC_TASK_TOOL_NAME_PREFIX: &str = "dynamic_task__create_task";
...
+            return Some(Vec::new());
+        }
+
+        let mut converted_extensions = Vec::new();
+
+        for ext in arr {
+            if let Some(name_str) = ext.as_str() {
+                // This is a shortname - check if it's loaded
+                if loaded_extensions.contains(&name_str.to_string()) {
+                    converted_extensions.push(ExtensionConfig::Builtin {
```
lifeizhou-ap: Just curious whether this will be always "ExtensionConfig::Builtin"? Is it possible it could be Stdio,....?

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
```diff
@@ -15,58 +21,203 @@ pub const DYNAMIC_TASK_TOOL_NAME_PREFIX: &str = "dynamic_task__create_task";
...
+
+    // Handle extensions
+    if let Some(extensions) = task_param.get("extensions") {
+        if let Some(ext_configs) = process_extensions(extensions, loaded_extensions) {
+            builder = builder.extensions(ext_configs);
+        }
+    }
+
+    // Handle other optional fields
+    fn apply_if_ok<T: serde::de::DeserializeOwned>(
```
lifeizhou-ap: nit: maybe move this function outside (not in the 'task_params_to_inline_recipe' function)

crates/goose/src/agents/subagent_task_config.rs
```diff
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use uuid::Uuid;
 
 /// Default maximum number of turns for task execution
-pub const DEFAULT_SUBAGENT_MAX_TURNS: usize = 5;
+pub const DEFAULT_SUBAGENT_MAX_TURNS: usize = 25;
```
lifeizhou-ap, line 8: I vaguely remember @wendytang mentioned that it does not work well when SUBAGENT_MAX_TURNS is > 10.  It started to be unreliable from SUBAGENT_MAX_TURNS >4.  Not sure whether this is still the case. It would be good to test a scenario with many turns

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
```diff
@@ -15,58 +21,203 @@ pub const DYNAMIC_TASK_TOOL_NAME_PREFIX: &str = "dynamic_task__create_task";
...
                                 "type": "string",
-                                "description": "The text instruction to execute"
+                                "description": "Task instructions"
                             },
+                            "prompt": {
+                                "type": "string",
+                                "description": "Initial prompt"
+                            },
+                            // Optional - auto-generated if not provided
+                            "title": {"type": "string"},
```
lifeizhou-ap, line 44: There are quite a few task_parameters elements and some of them does not have descriptions. Just curious about how well LLM provides these data?

crates/goose/src/prompts/system.md
tlongwell-block: I definitely kept that in mind when putting these together. They're as generic as possible while still making the agent more effective. The `screen` and `vim -c` are examples that most LLMs seem to understand in order to get the point across.

Almost all of the directions in here can be applied to almost any domain that Goose can be used for, not just software development

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
tlongwell-block: You're right. The agent doesn't use the optional parameters (outside of extensions) at all when running ad-hoc tasks. The lack of description is intentional because it would add considerable context and could confuse small LLMs

However, they are *very* useful when the agent is developing a recipe together with a user. It allows for a tight feedback loop that makes recipe development much faster and easier.

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
tlongwell-block: Done

crates/goose-cli/src/session/output.rs
tlongwell-block: Done

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
tlongwell-block: I've made this much more robust now and enabled the agent to select any enabled extension by shortname.

The extension is looked up dynamically and the config is pulled from the calling agent's ExtensionConfigManager's now

crates/goose/src/agents/subagent_task_config.rs
tlongwell-block: I've set this to 10 now. 25 worked fine in my testing and benching. 5 was too few to do anything along the lines of "try this approach for me"

crates/goose/src/prompts/system.md
michaelneale: ok SGTM

michaelneale:
@tlongwell-block I'm not seeing this make use of subagents at all - is that expected? DOesn't seem to break anything but I dont' really follow this change, seems fine otherwise, but not sure what expectation is (is there another change to get it to plan better to make use of subagents more efficiently?)

but have been using it today, and seems A-ok, FWIW

crates/goose/src/prompts/system.md
```diff
@@ -55,6 +55,26 @@ Template:
 - [ ] Blocked: waiting on credentials
 ```
 
+Deploy subagents (via dynamic task tool) for discrete, context-independent tasks. Use extension filters to limit resource access. This prevents context pollution, enables parallelization, provides sandboxes for testing uncertain approaches. Ideal when tasks share minimal context and benefit from isolation or concurrent execution.
```
michaelneale: this doesn't seem to work - has not used a subagent - can we tune this a bit to encourage it a bit more? 

"Use extension filters to limit resource access." - this may even be too obtuse for a frontier model less than opus?

michaelneale: I think this should change abnd move to where the TODO/tasks are planned (as I did in my PR) and try that?

michaelneale:
almost there - but I think the system prompt has to be adjusted for sub-agents to be more effective.

crates/goose/src/prompts/system.md
tlongwell-block: Okay, added that in, too https://github.com/block/goose/pull/4311/files#diff-507a51de1f94529d2674a027f5bf6e5eeed83ccfe9cbbe6cea674b75aef65ee8R44

tlongwell-block: Saw about a 300% increase in subagent usage in the benchmarks with that addition. Going to spend a little more time trying to refine it

tlongwell-block: Okay, last update to the system prompt. Very forceful language around subagents now.

Over 2000% increase in subagent usage over the benchmark now with this in place + slightly improved score when using sonnet

crates/goose/src/prompts/system.md
```diff
@@ -40,21 +40,30 @@ No extensions are defined. You should let the user know that they should add ext
...
+- Warning — `todo__write` overwrites entirely; always `todo__read` first (skipping is an error)
+- Keep items short, specific, action-oriented
+- Not using the todo tools for complex tasks is an error
 
 Template:
 ```markdown
 - [ ] Implement feature X
   - [ ] Update API
   - [ ] Write tests
+  - [ ] Run tests (subagent in parallel)
```
michaelneale, line 54: nice!

michaelneale:
also for subagents to be useful at all - need this so it doesn't flood the session.

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
```diff
@@ -15,58 +21,207 @@ pub const DYNAMIC_TASK_TOOL_NAME_PREFIX: &str = "dynamic_task__create_task";
...
+                                "type": "boolean",
+                                "description": "If true, return only the last message from the subagent (default: false, returns full conversation)"
+                            }
                         },
-                        "required": ["text_instruction"]
-                    }
+                        "anyOf": [
+                            {"required": ["instructions"]},
+                            {"required": ["prompt"]}
+                        ]
```
r0x0d, line 61: Looks like this addition broke the integration with providers using llama.cpp: https://github.com/block/goose/issues/4815

I'm not 100% familiar with the codebase yet, but I'm trying to propose a fix in the upcoming days. I have tracked it down to this line of code. 

I'm testing this with ramalama + llama3.2 (with llama.cpp).

r0x0d: cc: @iandouglas

michaelneale: @r0x0d is that lamma.cpp with rust or out of process? (have a branch playing with former as it is easier to ensure it works, but less powerful).

r0x0d: So, I'm suing ramalama for that, and playing out with goose from main brranch (latest commit).

I've seen that with ollama it works fine, the problem seems to be with llama.cpp

r0x0d: So, I'm suing ramalama for that, and playing out with goose from main brranch (latest commit).

I've seen that with ollama it works fine, the problem seems to be with llama.cpp

tlongwell-block: https://github.com/block/goose/pull/4897

PR #4312, opened 2025-08-24
Allows use of a port number for a custom OpenAI provider

vlascik:
Allows use of a port number, if specified in a custom OpenAI compatible provider config. Useful e.g. for LM Studio (`http://127.0.0.1:1234`). Port part was otherwise stripped.


DOsinga (approved):
probably also true for anthropic?

vlascik:
> probably also true for anthropic?

Anthropic provider doesn't seem to do any URL parsing that would strip the port:
 
https://github.com/block/goose/blob/cfa0bed294b97665933c50baca082591b6e50a0a/crates/goose/src/providers/anthropic.rs#L88

shawn111:
I tried on goose-cli v1.5,  it could run with env variables.
But v1.6 https://github.com/block/goose/pull/4099/commits , need this patch.

vlascik:
cargo fmt failed, sorry, I pushed a new commit. Seems the Ci needs manual approval to run again though.

Issue #4173, opened 2025-08-19
Add model retry values to config.yaml


bryan-valant:
**Please explain the motivation behind the feature request.**
I've been running into Bedrock rate limits with Claude Sonnet 4 recently and I'm too cheap to pay for provisioned throughput. 

WARN goose::providers::retry: Request failed, retrying (3/3): RateLimitExceeded("Bedrock throttling error: ThrottlingException { message: Some(\"Too many tokens, please wait before trying again.\"), meta: ErrorMetadata { code: Some(\"ThrottlingException\"), message: Some(\"Too many tokens, please wait before trying again.\"), extras: Some({\"aws_request_id\": \"xxxxxxxxxxxxxxxxxx\"}) } }")

**Describe the solution you'd like**
I'd like to set these values through the config.yaml file. 
https://github.com/block/goose/blob/7d7f08d70454e6c1ddb9445610009295f1388764/crates/goose/src/providers/retry.rs#L8

Adding an additional bit of delay would  help get me past the tokens/min limit. 
I have requested quota increase, but it would be nice to control these. 3 tries, 1 sec interval, and a multiplier of 2 seems a little fast for a rate that's measured "per minute". 

**Describe alternatives you've considered**
Limiting the scope of what I'm asking Goose/Bedrock/Claude to do. 
Quota increase.

**Additional context**

- [x] I have verified this does not duplicate an existing feature request

Cheers!
Bryan

Dan-Wuensch:
This issue is preventing my team from adopting goose for a few use cases as well!

AWS Bedrock is very aggressive with its throttling. Needs to be able to back off significantly more and have more retries

Dan-Wuensch:
@bryan-valant I opened https://github.com/block/goose/pull/4316 to address this - in local testing it helped get through some test prompts that were previously failing due to hitting the retry limit

Dan-Wuensch:
@bryan-valant this is available in 1.8.0 and up 🥳

Issue #4308, opened 2025-08-23
Unsupported operation: streaming not implemented when listing files


Innomen:
Describe the bug
When using Goose with the ollama provider and qwen3-coder model, attempting to list files in a directory fails. The session produces the error Unsupported operation: streaming not implemented, the last user message is removed, and the operation cannot be completed. This appears to be an unhandled exception related to non-streaming backends.

To Reproduce

Open a terminal and run goose

Start a session with provider ollama and model qwen3-coder

Enter the command: list files in directory please

Observe the error:

Indexing imagination... Error: Unsupported operation: streaming not implemented
Interrupted before the model replied and removed the last message.
error: The error above was an exception we were not able to handle.

Expected behavior
The directory contents should be listed, or at minimum Goose should handle the unsupported operation without crashing or deleting the user’s last message.

Screenshots
Not applicable; error occurs in CLI terminal.

Please provide following information:

OS & Arch: Garuda Linux x86_64

Interface: CLI

Version: Latest (2025-08-23)

Extensions enabled: None specified

Provider & Model: ollama - qwen3-coder

Additional context

Running llama.cpp server separately works with the model Qwen3-Coder-30B-A3B-Instruct-UD-Q5_K_XL.gguf

The issue seems related to operations that require streaming output, which is currently unsupported for this backend

Conversation context is removed when the error occurs

cxwx:
same issue on macos using block-goost-cli from homebrew

vamshi-vadala:
I can confirm that I am experiencing this same issue. disabled all extensions, chat only mode as well i get this in both UI and backend via CLI. Please release fix ASAP.

Goose Version: 1.6.0

Provider: ollama

Model: qwen2.5

OS & Arch: WINDOWS BASH CLI

Trigger Command: hi Hello

jbtaylor:
Same issue for me. First time user installed 1.6.0. Could not get Ollama on Mac Silicon to work. Spent a couple of hours troubleshooting with no success. Eventually found this post and installed 1.5.0 of both the Mac Silicon Desktop app and the CLI. Both work correctly with defaults.

vamshi-vadala:
@jbtaylor thanks for the hack, that fixed the issue for me as well.

joloono:
same issue here with 1.6.0, will downgrade. Using Version 1.6.0 on WSL and seperate trial on Windows Desktop with Ollama qwen3:1.7b and qwen3:4b test setup

jeremynielson:
[...]

PR #4316, opened 2025-08-25
feat: Add configurable Bedrock retry parameters

Dan-Wuensch:
Resolves https://github.com/block/goose/issues/4173

Changes included:
- Configurability of AWS Bedrock retry settings via environment variables
- New default retry settings for AWS Bedrock based on observed testing (these help mitigate failures due to running out of retries within the throttling window while don't push it too far)
- Add missing Databricks retry settings to providers documentation

Testing steps performed:
- AWS Bedrock provider retries using defaults
- AWS Bedrock provider retries using set environment variables
- Tried Ollama retries using `qwen2.5` and `qwen2.5-coder`, but this is blocked by the in 1.6.0 bug https://github.com/block/goose/issues/4308
- Use OpenAI provider


Dan-Wuensch:
I observed the following test is failing, but checked and it is also failing on `main`:

```
failures:

---- providers::base::tests::test_provider_metadata_context_limits stdout ----

thread 'providers::base::tests::test_provider_metadata_context_limits' panicked at crates/goose/src/providers/base.rs:564:9:
assertion `left == right` failed
  left: 200000
 right: 128000


failures:
    providers::base::tests::test_provider_metadata_context_limits
```

taniacryptid:
Awesome job on your first contribution! Let me tag the dev team to take a look ❤️ @DOsinga @michaelneale @alexhancock @zanesq @jamadeo @katzdave

crates/goose/src/providers/bedrock.rs
```diff
@@ -27,13 +27,23 @@ pub const BEDROCK_DEFAULT_MODEL: &str = "anthropic.claude-3-5-sonnet-20240620-v1
 pub const BEDROCK_KNOWN_MODELS: &[&str] = &[
     "anthropic.claude-3-5-sonnet-20240620-v1:0",
     "anthropic.claude-3-5-sonnet-20241022-v2:0",
+    "anthropic.claude-3-7-sonnet-20250219-v1:0",
+    "anthropic.claude-sonnet-4-20250514-v1:0",
+    "anthropic.claude-sonnet-4-20250514-v1:0",
```
DOsinga: maybe change the default model too?

DOsinga (approved):

crates/goose/src/providers/bedrock.rs
Dan-Wuensch: Does the default Goose install script use latest instead of version pinning? 

If so, I worry about disruption to existing users by changing the model since their prompts may work worse (or better!) on a new model. Could require tuning prompt(s) and running evals to adopt in some cases

That upgrade concern probably only applies to pipeline / automatic use cases not desktop CLI/GUI though. Idk what percentage of adoption that represents, and then of course Bedrock would only be a subset of that

What do you think about me logging an Issue for the model update instead, to have that be a separate more visible change?

DOsinga: yeah, we're working on that. right now it is a bit of a mess which model a conversation will use. we're working on making this a real property of the session so that if you change the global settings, it won't change it for existing conversations.

either way, changing the default here, should only change which model is suggested when you select the provider

Dan-Wuensch: Ah ok! In that case I'll push up a commit for updating default. Want to use latest Sonnet version?

Dan-Wuensch:
Demo of new default AWS Bedrock model recommendation:

Screenshot 2025-08-25 at 12 46 00 PM

Dan-Wuensch:
Tangential thought: Does `BEDROCK_KNOWN_MODELS` affect any validation of the user's model entry? We could look at including some models from other vendors than Anthropic at some point. Not a big concern to me as long as it will accept pasting model IDs from outside the list, which it looks like it does to me.

DOsinga:
These are just suggested models. @angelahning is looking into showing the full list in a filterable way

blackgirlbytes (approved):

blackgirlbytes:
@Dan-Wuensch there's a DCO check that's failing. Here's how to fix it 😄 https://github.com/block/goose/pull/4316/checks?check_run_id=48836635962

Dan-Wuensch:
@blackgirlbytes thank you! 🙌 Just force pushed a rebase with the sign-off included for those three commits.

michaelneale:
looking good a `cargo fmt` and maybe a script/cargo-lint.sh would help to check it is ok -

Dan-Wuensch:
Good catch! There was a piece in the model list that needed to be reformatted. Pushed up re-formatted now

Dan-Wuensch:
@blackgirlbytes checks and approvals are looking good! Except Semgrep appears to have gotten stuck. What's the next step in the process for these changes? 😄

documentation/docs/guides/environment-variables.md
```diff
@@ -99,6 +99,47 @@ export GOOSE_PLANNER_PROVIDER="openai"
 export GOOSE_PLANNER_MODEL="gpt-4"
 ```
 
+### Provider Retries
+
+Configurable retry parameters for LLM providers.
```
Dan-Wuensch, line 104: Nice!!

PR #4326, opened 2025-08-25
feat(releases): set up a workflow for nightly releases

alexhancock:
Step one of having nightly releases. This sets up a new workflow for nightly releases of goose. These won't be signed yet.

Followups:

* Figure out signing keys that will work for nightly builds and update `with.signing` to be `true`
* Set up https://www.electronjs.org/docs/latest/tutorial/updates#using-cloud-object-storage-serverless as a deliver channel
* Add UI to settings to enable the user to opt into nightly builds

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,98 @@
+# This workflow is for nightly releases, automatically triggered at midnight US Eastern
+name: Nightly Build
+
+on:
+  schedule:
+    # Run at midnight US Eastern (0500 UTC)
+    - cron: '0 5 * * *'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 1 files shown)


DOsinga:
maybe @jamadeo can have a look - he's much more versed in this sort of thing

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,119 @@
...
+    # Run at midnight US Eastern (0500 UTC)
+    - cron: '0 5 * * *'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # ------------------------------------
+  # 1) Prepare Version
```
jamadeo: nit: drop the comments here and just make them each the job name if necessary (ordered numbers are always mildly annoying to keep up to date when adding/removing steps)

PR #4327, opened 2025-08-25
feat: update aws-sdk-bedrockruntime to enable AWS_BEARER_TOKEN_BEDROCK auth

honsiorovskyi:
Bump aws-sdk-bedrockruntime from 1.74.0 to 1.103.0 to support the new AWS_BEARER_TOKEN_BEDROCK environment variable for httpBearerAuth scheme.

This update enables users to use bearer token authentication with Bedrock services by setting the AWS_BEARER_TOKEN_BEDROCK environment variable, providing an alternative to traditional AWS credentials.

Updates Cargo.lock with compatible minor version bumps for AWS SDK and smithy dependencies. All bedrock-related tests continue to pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -552,9 +552,9 @@ dependencies = [
 
 [[package]]
 name = "aws-credential-types"
-version = "1.2.1"
+version = "1.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


alexhancock (approved):

alexhancock:
Thanks @honsiorovskyi

michaelneale:
we may need to revert this - not sure what went on but bring in dependencies to cmake which cause problems (via https://crates.io/crates/aws-lc-sys it looks like which doesn't have stable versions if I read correctly)

PR #4331, opened 2025-08-25
Persist dynamic extension config so we can resume recipe sessions w/ extensions

### LINKED ISSUE #4295: Resuming a Recipe Session Should Include Extensions
Currently if I start a Goose CLI session using a recipe (e.g. `goose run --interactive --recipe recipe.yaml`) where `recipe.yaml` defines a certain extension, if I exit that session then later resume with `--resume` and `--name`/`--path` I'd expect the resumed session to include the same extensions as the original session (that are defined in the recipe). But it seems like dynamic extensions are enabled in memory and don't get saved to the session, so they aren't loaded when resuming that session

I'd like to be able to define a recipe `recipe.yaml` which includes an extension like `mcp_pagerduty` then start a session using the recipe: `goose run --interactive --recipe recipe.yaml`. If I later exit the session, I'd like to be able to `goose session --resume --name <session_id>.jsonl` and have the new(resumed) session auto load `mcp_pagerduty` like the original recipe session did

Since I've enabled `mcp_pagerduty` as a global extension I can ask the model to enable it dynamically in the session, but it would be a more consistent user experience for resumed recipe sessions to auto load all the same extensions defined in the recipe file

wpfleger96:
This PR enables extension persistence across session resume by storing extension configurations in the `ExtensionData` field of `Session`. When users resume previous sessions with `--resume`, the previous session's dynamic extensions are now automatically reloaded.

Before this change, if a user started an interactive session using a recipe then later resumed it, the original recipe session's extensions were NOT reloaded into the resumed session, creating an inconsistent experience. Extension state is captured and persisted throughout the session, whether extensions are defined via recipe, `--with-builtin`, `--with-extension`, `--with-remote-extension`, or `--with-streamable-http-extension`.

Implementation details:
- Extension configurations are stored in `ExtensionData` using the `ExtensionState` trait pattern with versioned keys for future compatibility
- Extension state is saved after every `manage_extensions()` call to maintain consistency
- Full `ExtensionConfig` objects are stored (not just names) to preserve exact session configuration including timeouts, environment variables, and other settings
- `Agent::save_extension_state()` is made public and reused in session builder to eliminate code duplication

Fixes #4295


crates/goose-cli/src/session/builder.rs
```diff
@@ -348,6 +348,27 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
     let extensions_to_run: Vec<_> = if let Some(extensions) = session_config.extensions_override {
         agent.disable_router_for_recipe().await;
         extensions.into_iter().collect()
+    } else if session_config.resume {
+        if let Some(session_file) = session_file.as_ref() {
+            match session::read_metadata(session_file) {
+                Ok(metadata) if metadata.enabled_extensions.is_some() => {
+                    metadata.enabled_extensions.unwrap().into_iter().collect()
+                }
+                _ => ExtensionConfigManager::get_all()
```
DOsinga: this seems to be a very common pattern in the code, now +2. let's extract a method .get_enabled() that just returns the configs that are enabled

crates/goose-cli/src/session/builder.rs
```diff
@@ -432,7 +453,7 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
         session_config.retry_config.clone(),
     );
 
-    // Add extensions if provided
+    // Add CLI extensions if provided and track their source
```
DOsinga: I am not sure I follow the change in this comment. suggest deleting it. Also merge the two print statements on 459 and 467

crates/goose-cli/src/session/builder.rs
```diff
@@ -585,6 +606,23 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
         session.agent.override_system_prompt(override_prompt).await;
     }
 
+    // Save all extension configurations for session resume (after all extensions are added)
+    if let Some(session_file) = &session_file {
+        let all_extension_configs = session.agent.get_extension_configs().await;
+        if !all_extension_configs.is_empty() {
+            if let Err(e) =
+                goose::session::update_metadata_with_extensions(session_file, all_extension_configs)
+                    .await
```
DOsinga: I do not like this (existing) pattern of rewriting the session_file any time we have a change; for one thing if we fail halfway through we probably end up with a faulty session; can you look into changing this all so instead of having a session_file we just have a session object we built out and finally save? that way we also don't need yet another update_meta_data_with* method

crates/goose-cli/src/session/builder.rs
```diff
@@ -585,6 +606,23 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
...
+            {
+                tracing::error!("Failed to persist extension configuration: {}", e);
+                if !session_config.quiet {
+                    println!("Warning: Extension configuration could not be saved. Session resume may not work correctly.");
+                }
+                // Non-fatal: continue session even if we can't persist extensions
```
DOsinga: should this be non fatal?

DOsinga (approved):

crates/goose-cli/src/session/builder.rs
wpfleger96: Addressed [here](https://github.com/block/goose/pull/4331/commits/fb06fb8e6aea84e1f6f45627771dd4d7e435d600), clarified that this block is for `stdio` extensions and I also went ahead and cleaned up the redundant print statements in the remote/streamable/builtin extension blocks too

crates/goose-cli/src/session/builder.rs
wpfleger96: great idea! addressed this [here](https://github.com/block/goose/pull/4331/commits/8013c7f5bb13eb11ec72a82b6c354242e164ad41), extracted the duplicated code into a `get_enabled()` method that can be reused

crates/goose/src/config/extensions.rs
```diff
@@ -85,4 +85,12 @@ impl ExtensionConfigManager {
...
         Ok(extensions.get(key).map(|e| e.enabled).unwrap_or(false))
     }
+
+    pub fn get_enabled() -> Result<Vec<ExtensionConfig>> {
+        Ok(Self::get_all()?
+            .into_iter()
+            .filter(|ext| ext.enabled)
+            .map(|ext| ext.config)
+            .collect())
+    }
```
jamadeo: not part of your PR, more just an observation of this code, but:

1) there's no reason ExtensionConfigManager needs to be a struct with methods, these should just be free functions
2) nothing here returns an Err, so the Result types are also unnecessary

crates/goose-cli/src/session/builder.rs
```diff
@@ -348,13 +348,19 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
     let extensions_to_run: Vec<_> = if let Some(extensions) = session_config.extensions_override {
         agent.disable_router_for_recipe().await;
         extensions.into_iter().collect()
+    } else if session_config.resume {
+        if let Some(session_file) = session_file.as_ref() {
+            match session::read_metadata(session_file) {
+                Ok(metadata) if metadata.enabled_extensions.is_some() => {
```
jamadeo:
```
                Ok(SessionMetadata {
                    enabled_extensions: Some(extensions),
                    ..
                }) => extensions.into_iter().collect(),
 ```

File: crates/goose-cli/src/session/builder.rs [hunk 7]
```diff
@@ -348,13 +348,19 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
...
+            ExtensionConfigManager::get_enabled().expect("should load extensions")
+        }
     } else {
-        ExtensionConfigManager::get_all()
-            .expect("should load extensions")
-            .into_iter()
-            .filter(|ext| ext.enabled)
-            .map(|ext| ext.config)
-            .collect()
+        ExtensionConfigManager::get_enabled().expect("should load extensions")
```
jamadeo: if you do remove the `Result<>`s from the ExtensionConfigManager, you can avoid these `.expect()`s

crates/goose-cli/src/session/builder.rs
wpfleger96: addressed [here](https://github.com/block/goose/pull/4331/commits/7d56e05152bea612bb7bac2085c230160469dcd3)

crates/goose-cli/src/session/builder.rs
wpfleger96: addressed [here](https://github.com/block/goose/pull/4331/commits/7d56e05152bea612bb7bac2085c230160469dcd3)

crates/goose/src/config/extensions.rs
wpfleger96: addressed [here](https://github.com/block/goose/pull/4331/commits/3240ed99403f4c5b20396ea030f6a1f7acec7e72)

crates/goose-cli/src/session/builder.rs
wpfleger96: addressed [here](https://github.com/block/goose/pull/4331/commits/3240ed99403f4c5b20396ea030f6a1f7acec7e72)

crates/goose-cli/src/session/builder.rs
wpfleger96: addressed [here](https://github.com/block/goose/pull/4331/commits/086cb88569112dbc5e902b99ddca448db58c69cb)

DOsinga:
sorry for letting this slip. can you sync with main?

wpfleger96:
@DOsinga sure np, for context I got tied up with a few other things and let this go stale to see how @tlongwell-block 's [agent unification work](https://github.com/block/goose/discussions/4389) would shake out to make sure I didn't muddy the waters there

crates/goose/src/agents/agent.rs
```diff
@@ -647,12 +647,39 @@ impl Agent {
         )
     }
 
+    /// Save current extension state to session metadata
+    /// Should be called after any extension add/remove operation
+    pub async fn save_extension_state(&self, session: &Option<SessionConfig>) -> Result<()> {
```
jamadeo: this should just take a &SessionConfig instead of an Option, then where you'd call it with None instead just don't call it

crates/goose/src/agents/agent.rs
```diff
@@ -647,12 +647,39 @@ impl Agent {
...
+        if let Some(session_config) = session {
+            let extension_configs = self.extension_manager.get_extension_configs().await;
+
+            let extensions_state = EnabledExtensionsState::new(extension_configs);
+
+            // Load current session
```
jamadeo: these comments don't add anything

crates/goose/src/agents/agent.rs
```diff
@@ -756,6 +773,17 @@ impl Agent {
                 }
             }
         }
+
+        // Save extension state after successful operation
```
jamadeo: remove this comment

crates/goose/src/agents/agent.rs
```diff
@@ -647,12 +647,39 @@ impl Agent {
...
+        Ok(())
+    }
+
     #[allow(clippy::too_many_lines)]
     pub(super) async fn manage_extensions(
         &self,
         action: String,
         extension_name: String,
         request_id: String,
+        session: &Option<SessionConfig>,
```
jamadeo: I think this is only called by the "platform" tool that can enable extensions within a session. Probably worth thinking about -- if the recipe specifies extensions A and B, then halfway through the session Goose decides to enable extension C, should the session now be saved with all of them?

crates/goose/src/config/extensions.rs
```diff
@@ -25,64 +25,142 @@ pub fn name_to_key(name: &str) -> String {
...
 
-    pub fn is_enabled(key: &str) -> Result<bool> {
-        let extensions = Self::get_extensions_map()?;
-        Ok(extensions.get(key).map(|e| e.enabled).unwrap_or(false))
+    #[test]
+    fn test_get_extensions_map_returns_hashmap() {
+        // Test that get_extensions_map returns a HashMap (may be empty or not depending on global config)
+        let extensions = get_extensions_map();
+        // Just verify it returns a HashMap - don't assert on contents since global config may vary
+        assert!(extensions.is_empty() || !extensions.is_empty());
```
jamadeo: this is kind of a silly test :)

I'd do another pass at the tests to make sure they all make sense. LLMs love to do things like this

crates/goose-cli/src/session/builder.rs
```diff
@@ -322,13 +321,28 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> CliSession {
...
+    } else if session_config.resume {
+        if let Some(session_id) = session_id.as_ref() {
+            match SessionManager::get_session(session_id, false).await {
+                Ok(session_data) => {
+                    // Try to load saved extension configs directly
+                    if let Some(saved_state) =
+                        EnabledExtensionsState::from_extension_data(&session_data.extension_data)
+                    {
+                        // Use the saved configs as-is (no lookup needed!)
+                        saved_state.extensions
```
jamadeo, line 379: What happens if I run a recipe with extensions A and B, then I later uninstall (not disable) extension B, and then resume that session? For consistency/repeatability you'd want B, but as a user I think I would expect Goose to never use B as long as I've uninstalled it.

Should it fail? Warn?

crates/goose/src/agents/agent.rs
wpfleger96: > should the session now be saved with all of them?

yeah that was my thought. for me as a Goose user when resuming a session intuitively I'd expect to have the same set of extensions that I did whenever I exited the original session, whether they were recipe or dynamic extensions (via `--with-extension`/etc. OR enabled mid-session with platform tool) or some combination of these. this is just my own experience, open to other ideas if this doesn't make sense!

Issue #4334, opened 2025-08-25
bug: ollama no longer working due to streaming


michaelneale:
try to use goose with latest ollama, will be an error due to streaming. 

seems to be as ollama provider claims it can support streaming when I don't think it can.

aleeusgr:
Here are some details about the error:
```sh
[user@nixos:~]$ goose
starting session | provider: ollama model: qwen3:0.6b
...
Goose is running! Enter your instructions, or try asking what goose can do.

Context: ○○○○○○○○○○ 0% (0/128000 tokens)
( O)> hi
◐  Wrangling widgets...                                                                             Error: 
Unsupported operation: streaming not implemented
◓  Wrangling widgets...                                                                             Interrupted before the model replied and removed the last message.
  
error: The error above was an exception we were not able to handle.
These errors are often related to connection or authentication
```

```sh
[user@nixos:~]$ goose web

🪿 Starting Goose web server
   Provider: ollama | Model: qwen3:0.6b
   Working directory: /home/alex
   Server: http://127.0.0.1:3000
   Press Ctrl+C to stop

  2025-11-07T06:57:31.122079Z ERROR goose_cli::commands::web: Error in message stream: Unsupported operation: streaming not implemented
    at crates/goose-cli/src/commands/web.rs:671
```

DOsinga:
qwen3:0.6b seems awfully small. how should this work @michaelneale ? should it autodetect streaming? does 0.6b not do taht?

aleeusgr:
The model isn't that relevant, as qwen2.5 ,which is suggested in the docs, fails with the same error. It's also worth noting I realised it might have been fixed: I realised I am running 1.6.0

DOsinga:
ok, thanks. we have some better debugging support coming up for ollama/qwen in the upcoming release

Issue #4329, opened 2025-08-25
Move Ollama to the other providers section


blackgirlbytes:
For first time users, we're presented with Ollama, Tetrate, and OpenRouter as options. We're thinking Ollama can move to the other providers section until the experience is better for new users or folks with minimal specs.

shawn111:
Is it related to https://github.com/block/goose/issues/4350?

I used to use gemini but just try on Local AI provider.
Which provider should I choice? openai (from v1.6.0 add fast_model only for openai) or ollama (not support api_key).

Or you would like create new one?

PR #4335, opened 2025-08-25
no streaming for ollama and re-arrange cards

michaelneale:
fixes: https://github.com/block/goose/issues/4329
fixes: https://github.com/block/goose/issues/4334

ollama now will advertise streaming, the goose provider incorrectly said it would support it, which would break it as of latest goose and ollama, this fixes that. 

as requested, ollama below the fold (with a note that is for advanced users): 
Screenshot 2025-08-26 at 8 36 15 am


angiejones:
is Ollama only there because it's detected on your system? Hoping it's not shown by default, since most people wont have the hardware to support a local model that performs well with agents

blackgirlbytes:
Thanks! I think the team prefers that Ollama is not shown at all, and it's within the Other Providers section.

How does it determined detected? is it checking that someone has ollama installed on their device?

michaelneale:
@blackgirlbytes yeah it looks on device for it being installed

michaelneale:
@angiejones at the moment it detects but I don't think it shows/hides as it can offer to link to download (idea was to detect machine capacity, but perhaps that is too ambitious), if it is trapping people (who have very small ram) can make it hidden easily?

PR #4337, opened 2025-08-26
Start agent in App.tsx

jamadeo:


ui/desktop/src/components/hub.tsx
```diff
@@ -36,9 +36,6 @@ export default function Hub({
     const combinedTextFromInput = customEvent.detail?.value || '';
 
     if (combinedTextFromInput.trim()) {
-      // Navigate to pair page with the message to be submitted
-      // Pair will handle creating the new chat session
-      resetChat();
```
DOsinga: I think we still want to conditionally call resetChat here - if the user already started a conversation, they'll expect that one to be wiped out when going here.

ui/desktop/src/App.tsx
```diff
@@ -283,6 +283,12 @@ export default function App() {
   const { modalState, modalConfig, dismissModal, confirmInstall } =
     useExtensionInstallModal(addExtension);
 
+  useEffect(() => {
+    (async () => {
+      await loadCurrentChat({ setAgentWaitingMessage });
+    })();
+  }, [loadCurrentChat, setAgentWaitingMessage]);
+
```
DOsinga, line 296: we have a lot of `useEffect`s going on here, I worry about race conditions here. I wonder if this one should be added to the setView one at the top, so at least we know there is a setView? also now seeing we call reactReady back even though we are not here maybe? /cc @zanesq could that explain some of the refresh crash weirdness?

DOsinga (approved):

PR #4338, opened 2025-08-26
Use middleware to verify secret key

jamadeo:


crates/goose-server/src/auth.rs
```diff
@@ -0,0 +1,22 @@
+use axum::{
+    extract::{Request, State},
+    http::StatusCode,
+    middleware::Next,
+    response::Response,
+};
+
+pub async fn layer_fn(
```
DOsinga: is this the normal way to name these? check_token would be easier to follow

crates/goose-server/src/commands/agent.rs
```diff
@@ -3,10 +3,12 @@ use std::sync::Arc;
 use crate::configuration;
```
DOsinga, line 3: we should really name this differently :) I was like, why is this in agent.rs

ui/desktop/src/goosed.ts
```diff
@@ -25,25 +25,24 @@ export const findAvailablePort = (): Promise<number> => {
...
   log.info(`Checking server status at ${statusUrl}`);
 
   for (let attempt = 1; attempt <= maxAttempts; attempt++) {
     try {
-      const response = await fetch(statusUrl);
+      const response = await fetch(statusUrl, {
+        headers: {
+          'X-Secret-Key': secret,
+        },
+      });
```
DOsinga: could we use the generated API at this point or would it not be ready (or is it only on the other side?)

DOsinga (approved):

crates/goose-server/src/auth.rs
jamadeo: Nope it's just a bad name

ui/desktop/src/goosed.ts
jamadeo: looks like we don't actually add this endpoint to the generated client. It's also a bit of a mixed bag all over where we do vs don't use the generated client. I'll have a look at cleaning that up too, as this PR is kind of a logical place to do it

DOsinga: yeah, it's all over the place. we should really set an alert on getSecret appearing in PRs

PR #4341, opened 2025-08-26
only show ollama when installed

michaelneale:
this detects ollama and shows the card only when it is installed 

another take on https://github.com/block/goose/issues/4329 cc @blackgirlbytes @angiejones 

(currently is down below the fold, and mentions significant hardware requirements)

### DIFF
--- a/ui/desktop/src/components/ProviderGuard.tsx
+++ b/ui/desktop/src/components/ProviderGuard.tsx
@@ -430,52 +430,53 @@ export default function ProviderGuard({ children }: ProviderGuardProps) {
                 </div>
 
                 {/* Ollama Card - outline style */}
-                <div className="relative">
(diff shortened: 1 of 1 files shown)


angiejones:
would prefer https://github.com/block/goose/pull/4348. just because Ollama is installed doesnt mean the user has a model that is good for tool calling. it's just overall not the best experience

michaelneale:
@angiejones given how much I Have been pinged about the regression (with streaming) from 1.5, I think a lot more are using it than we realise. This change won't show it unless running, and will check for the model/download the prescribed one if they choose to go that path (otherwise they are more likely to get the generic ollama case and use a model we don't know will work).

michaelneale:
closing as was removed instead

PR #4348, opened 2025-08-26
remove Ollama card from welcome screen

### LINKED ISSUE #4347: Hide Ollama from Home Screen
> @angiejones at the moment it detects but I don't think it shows/hides as it can offer to link to download (idea was to detect machine capacity, but perhaps that is too ambitious), if it is trapping people (who have very small ram) can make it hidden easily?

_Originally posted by @michaelneale in https://github.com/block/goose/issues/4335#issuecomment-3222764442_

yes, please remove Ollama from the home screen. It is by far the hardest provider to get working with goose due to common hardware limitations. We don't want to introduce people to the hard path.

angiejones:
fixes #4347

### DIFF
--- a/ui/desktop/src/components/ProviderGuard.tsx
+++ b/ui/desktop/src/components/ProviderGuard.tsx
@@ -8,9 +8,9 @@ import WelcomeGooseLogo from './WelcomeGooseLogo';
 import { initializeSystem } from '../utils/providerUtils';
 import { toastService } from '../toasts';
 import { OllamaSetup } from './OllamaSetup';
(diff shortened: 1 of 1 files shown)


blackgirlbytes (approved):

michaelneale:
I think for this to not regress to people picking a model that doesn't work, we need to re-do the providers ollama setup to recommend/choose the model (unless it is better to make the ollama setup deliberately hard)

michaelneale:
yeah it works: 

image

does set back ollama a bit though - as old approach would pull down the "correct" model (although who knows, that changes from time to time).

PR #4349, opened 2025-08-26
Reset agent recipe

DOsinga:
This still only partially fixes it. We always load the recipe from the windows.appConfig so unless we jump through some weird loops, we'll alway shave some traces in a deeplink launched window. fixing that takes a rather larger refactor


crates/goose-server/src/state.rs
```diff
@@ -1,14 +1,14 @@
...
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 use tokio::sync::Mutex;
 
 pub type AgentRef = Arc<Agent>;
 
 #[derive(Clone)]
 pub struct AppState {
-    agent: Option<AgentRef>,
+    agent: Arc<Mutex<Option<AgentRef>>>,
```
jamadeo: this will now be an `Arc<Mutex<Option<Arc<Agent>>>>`

better to change the AgentRef type above? my cmd+f tells me it need not be `pub` either

ui/desktop/src/hooks/useRecipeManager.ts
```diff
@@ -247,28 +217,23 @@ export const useRecipeManager = (chat: ChatType, locationState?: LocationState)
...
           throw new Error('No recipe data received');
         }
 
-        // Set a flag to prevent the current window from reacting to recipe config changes
-        // This prevents navigation conflicts when creating new windows
         window.sessionStorage.setItem('ignoreRecipeConfigChanges', 'true');
 
-        // Create the new window
         window.electron.createChatWindow(
-          undefined, // query
```
jamadeo, line 261: we should just give this its own `CreateChatWindowOptions` type

crates/goose-server/src/state.rs
```diff
@@ -42,4 +44,19 @@ impl AppState {
...
+    /// Reset the app state
+    /// after we make goosed multi-agent, we shouldn't need this anymore
+    pub async fn reset(&self) {
+        let mut agent = self.agent.lock().await;
+        *agent = Some(Arc::new(Agent::new()));
+
+        {
+            let mut scheduler = self.scheduler.lock().await;
+            *scheduler = None;
+        }
```
jamadeo, line 51: the PR mentions recipes but this looks like it only resets the scheduler?

DOsinga: we don't keep track of the recipes on the server side yet, so the reset of that still happens on the client sadly

ui/desktop/src/hooks/useRecipeManager.ts
DOsinga: yeah that would be better. we're using a bunch of different ways to get data to the other side though, should probably unify all that

crates/goose-server/src/state.rs
DOsinga: Maybe? I removed the optional which seems like a win. I am no Rust expert but to me this reads like, we have things that need an AgentRef, but we manage that AgentRef in a mutable fashion by sticking it into an Arc<Mutex

jamadeo: `Arc<RwLock<Agent>>` is I think what you want

Issue #4350, opened 2025-08-26
[Bug] Goose attempts to use "gpt-4o-mini" model when use  OpenAI (GPT-4 and other OpenAI models, including OpenAI compatible ones)


shawn111:
**Describe the bug**

**To Reproduce**
1. goose configure (goose-cli)
2. use OpenAI Configure for my local AI provider (not real OpenAI)
3. .. save config
4. goose

**Expected behavior**

**Screenshots**

```
pixel9a~/o/goose►goose                                                                               29.722s (build-termux-limit|?↑5↓6) 23:54
starting session | provider: openai model: GPT-OSS-120B-GPU
    logging to /data/data/com.termux/files/home/.local/share/goose/sessions/20250826_235433.jsonl
    working directory: /data/data/com.termux/files/home/opensource/goose

Goose is running! Enter your instructions, or try asking what goose can do.

Context: ○○○○○○○○○○ 0% (0/128000 tokens)
( O)> hello
  2025-08-26T15:54:52.211210Z  WARN goose::providers::utils: Provider request failed with status: 404 Not Found. Payload: Some(Object {"object": String("error"), "message": String("The model `gpt-4o-mini` does not exist."), "type": String("NotFoundError"), "param": Null, "code": Number(404)}). Returning error: RequestFailed("Request failed with status: 404 Not Found")
    at crates/goose/src/providers/utils.rs:114

  2025-08-26T15:54:52.211299Z ERROR goose::session::storage: Failed to generate session description: Request failed: Request failed with status: 404 Not Found
    at crates/goose/src/session/storage.rs:1304

Session ended with error: Failed to generate session description
```

**Please provide following information:**
 - **OS & Arch:** android / termux
 - **Interface:** CLI
 - **Version:** v1.6.0 (v1.5.0 works)
 - **Provider & Model:** OpenAI (OpenAI compatible ones)

**Additional context**

```
+++ b/crates/goose/src/providers/openai.rs
@@ -60,13 +60,17 @@ impl_provider_default!(OpenAiProvider);

 impl OpenAiProvider {
     pub fn from_env(model: ModelConfig) -> Result<Self> {
+        let model = model.with_fast(OPEN_AI_DEFAULT_FAST_MODEL.to_string());
```

It seem force use OPEN_AI_DEFAULT_FAST_MODEL cause this problem.

Related: https://github.com/block/goose/issues/4318

DOsinga:
I think #4318 is something different, but I'll have a quick look

[...]

PR #4353, opened 2025-08-26
fix: openai add checking about fast model

shawn111:
for OpenAI compatible might does't support gpt-4o-mini

it can fix https://github.com/block/goose/issues/4350


shawn111:
@katzdave I can bypass your OPEN_AI_DEFAULT_FAST_MODEL setting by this commit.
But no sure is it the best way.

Maybe like what you said, create a new GOOSE_FAST_MODEL variable to handle maybe more flexible.

DOsinga:
this will certainly fix it, but it's not really the way I think.

shawn111:
crates/goose/src/providers/databricks.rs did some similar check about fast model

shawn111:
It is fixed by Fast model falls back to regular #4375

PR #4354, opened 2025-08-26
Add Tetrate Agent Router Service to Provider Registry

johnlanda:
Adds Tetrate Agent Router Service to the provider registry for folks who want to configure manually. Note that only users going through the PKCE flow will get the promotional credit at the moment, however.


michaelneale (approved):
thanks, works.

Issue #4056, opened 2025-08-12
Eleven labs audio transcription not working 412 (Precondition Failed)


zanesq:
**Describe the bug**

Sending an audio transcription request to eleven labs with a valid api key fails and returns with 

POST http://127.0.0.1:51228/audio/transcribe/elevenlabs 412 (Precondition Failed)
(anonymous) @ useWhisper.ts:139

await in (anonymous)
mediaRecorder.onstop @ useWhisper.ts:297
useWhisper.ts:171 Error transcribing audio: Error: Transcription failed
    at useWhisper.ts:160:17
    at async mediaRecorder.onstop (useWhisper.ts:297:11)

**To Reproduce**
1. Go to settings and put in an eleven labs key and choose eleven labs as the provider
2. Click on mic in chat and try to transcribe
3. See error message

**Expected behavior**
It should work with the eleven labs api and the chat should revert back to chat instead of staying in recording mode if there is an issue.

**Screenshots**
<img width="1770" height="154" alt="Image" src="https://github.com/user-attachments/assets/b15394d1-f7e0-4228-b9c7-b4b956154bae" />

**Please provide following information:**
 - **OS & Arch:** Mac Arm
 - **Interface:** UI
 - **Version:** 1.3.0
 - **Extensions enabled:** Redacted
 - **Provider & Model:** Claude Sonnet 4

**Additional context**

PR #4358, opened 2025-08-26
Fix eleven labs audio transcription and added more logging

zanesq:
Not sure how it worked before but verified this fixes eleven labs not working locally and also the issue with the ui not respecting the previous audio provider selection.

fixes https://github.com/block/goose/issues/4056

### DIFF
--- a/crates/goose-server/src/routes/audio.rs
+++ b/crates/goose-server/src/routes/audio.rs
@@ -44,12 +44,11 @@ struct WhisperResponse {
 
 /// Validate audio input and return decoded bytes and file extension
 fn validate_audio_input(
-    request: &TranscribeRequest,
+    audio: &str,
+    mime_type: &str,
 ) -> Result<(Vec<u8>, &'static str), StatusCode> {
     // Decode the base64 audio data
-    let audio_bytes = BASE64
-        .decode(&request.audio)
-        .map_err(|_| StatusCode::BAD_REQUEST)?;
+    let audio_bytes = BASE64.decode(audio).map_err(|_| StatusCode::BAD_REQUEST)?;
 
     // Check file size
     if audio_bytes.len() > MAX_AUDIO_SIZE_BYTES {
@@ -62,8 +61,9 @@ fn validate_audio_input(
     }
 
     // Determine file extension based on MIME type
-    let file_extension = match request.mime_type.as_str() {
+    let file_extension = match mime_type {
         "audio/webm" => "webm",
+        "audio/webm;codecs=opus" => "webm",
         "audio/mp4" => "mp4",
         "audio/mpeg" => "mp3",
         "audio/mpga" => "mpga",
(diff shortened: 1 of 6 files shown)


crates/goose-server/src/routes/audio.rs
```diff
@@ -109,19 +93,41 @@ async fn transcribe_handler(
         Err(_) => "https://api.openai.com".to_string(),
     };
 
-    tracing::debug!("Using OpenAI host: {}", openai_host);
+    Ok((api_key, openai_host))
+}
+
+/// Send transcription request to OpenAI Whisper API
+async fn send_openai_request(
```
DOsinga, line 100: It would be better if we would just move this to the openai provider class and then just instantiate the openai provider; that way it would use whatever the user had configured for openai (proxy etc). we'd still have to check whether it works of course, but it would be cleaner

ui/desktop/src/components/settings/dictation/DictationSection.tsx
```diff
@@ -27,25 +27,30 @@ export default function DictationSection() {
...
+        console.log('Loaded dictation settings from localStorage:', parsed);
       } else {
-        // Default settings
-        const defaultSettings: DictationSettings = {
-          enabled: true,
-          provider: 'openai',
+        // Default settings - don't force OpenAI as default
+        loadedSettings = {
+          enabled: false,
+          provider: null,
```
DOsinga: this should come from settings of course

crates/goose-server/src/routes/audio.rs
```diff
@@ -330,16 +404,13 @@ async fn check_dictation_config(
...
+    let has_elevenlabs = match config.get_secret::<String>("ELEVENLABS_API_KEY") {
+        Ok(_) => true,
+        Err(_) => {
             // Check non-secret for backward compatibility
-            config
-                .get("ELEVENLABS_API_KEY", false)
-                .map(|_| true)
-                .unwrap_or(false)
-        });
+            config.get("ELEVENLABS_API_KEY", false).is_ok()
```
DOsinga, line 385: what backwards compatibility is this?

ui/desktop/src/hooks/useWhisper.ts
```diff
@@ -166,7 +166,7 @@ export const useWhisper = ({ onTranscription, onError, onSizeWarning }: UseWhisp
         };
         let body: Record<string, string> = {
           audio: base64Audio,
-          mime_type: 'audio/webm',
+          mime_type: audioBlob.type || 'audio/webm',
```
DOsinga: is audioBlob.type ever not set? if so shouldn't we just bail?

crates/goose-server/src/routes/audio.rs
zanesq: I think it used to be stored in plain text in the config so its trying that first before secrets. Might be able to remove that now

crates/goose-server/src/routes/audio.rs
zanesq: yeah good call, will follow up with this to get this fix out sooner

ui/desktop/src/hooks/useWhisper.ts
zanesq: good call will change that 👍

ui/desktop/src/components/settings/dictation/DictationSection.tsx
zanesq: true, moved these to a shared settings constants area rather than hard coded

PR #4360, opened 2025-08-26
feat: Added scroll state support for chat-session-list navigation

tiensi:
**Summary**
Added scroll state memory when navigating back from session view. This works for filtered search results as well.

**Changes**
- Added `showSessionHistory` to SessionsView to replace reliance on `selectedSession` being null
- Passed `selectedSession` as a new prop to `SessionListView` so that it can scroll to a selected session
- Added `useEffect` with `MutationObserver` to detect and scroll to selected session when loaded.

**Request**
https://github.com/block/goose/issues/3664

**After Changes**

https://github.com/user-attachments/assets/2440e256-ee02-45a5-94b8-bfda2650274f

***NOTE*** 
[Previous PR was closed out due to staleness](https://github.com/block/goose/pull/3684). Comments were addressed in this pr

### DIFF
--- a/ui/desktop/src/components/sessions/SessionListView.tsx
+++ b/ui/desktop/src/components/sessions/SessionListView.tsx
@@ -183,6 +183,16 @@ const SessionListView: React.FC<SessionListViewProps> = React.memo(({ onSelectSe
 
   const containerRef = useRef<HTMLDivElement>(null);
 
+  // Track session to element ref
+  const sessionRefs = useRef<Record<string, HTMLElement>>({});
+  const setSessionRefs = (itemId: string, element: HTMLDivElement | null) => {
+    if (element) {
+      sessionRefs.current[itemId] = element;
+    } else {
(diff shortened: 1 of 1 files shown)


tiensi:
Thank you for the review! I'll go ahead and make your suggested changes.

> It doesn't seem to respect the scroll after searching

I actually wanted to have a separate follow up to preserve the search-filter on back which would make it easier to reuse this scroller.

> Using mutation observer is not ideal

Got it! I'm still getting a handle on react so I appreciate the heads up.

tiensi:
I went ahead and pushed up the suggested changes:
- Replaced `MutationObserver ` with `useRef<Record<string, HTMLElement>>({});`
- Kept `scrollIntoView({block: "center"})` (there was a separate suggestion that smooth scroll experience is poor if a user scrolls all the way to the bottom and toggles back and forth)
- Updated `useEffect` to listen to `[selectedSessionId, sessions]` which should operate better with element ref.

PR #4361, opened 2025-08-26
Nest TODO State in session data

katzdave:
Use tool version + name as the key for flexibility to migrate these configs.


crates/goose/src/agents/agent.rs
```diff
@@ -494,7 +495,11 @@ impl Agent {
             let todo_content = if let Some(path) = session_file_path {
                 session::storage::read_metadata(&path)
                     .ok()
-                    .and_then(|m| m.todo_content)
+                    .and_then(|m| {
+                        // Get from new session_data format
+                        session::TodoState::from_session_data(&m.session_data)
```
DOsinga: new session format?

crates/goose/src/session/tool_state.rs
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    /// Set tool state for a specific tool and version
+    pub fn set_tool_state(&mut self, tool_name: &str, version: &str, state: Value) {
+        let key = format!("{}.{}", tool_name, version);
+        self.tool_states.insert(key, state);
+    }
+
+    /// Remove tool state for a specific tool and version
+    pub fn remove_tool_state(&mut self, tool_name: &str, version: &str) -> Option<Value> {
```
DOsinga: is this being used?

crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -269,7 +269,7 @@ mod tests {
             accumulated_total_tokens: Some(100),
             accumulated_input_tokens: Some(50),
             accumulated_output_tokens: Some(50),
-            todo_content: None,
+            session_data: crate::session::SessionData::new(),
```
DOsinga: might be easier if this was just nullable

crates/goose/src/session/tool_state.rs
```diff
@@ -0,0 +1,298 @@
...
+        Self { content }
+    }
+
+    /// Create an empty TODO state
+    pub fn empty() -> Self {
+        Self {
+            content: String::new(),
+        }
+    }
+}
```
DOsinga: this should not live here; sessions shouldn't know about specific tools

crates/goose/src/agents/agent.rs
```diff
@@ -494,7 +495,11 @@ impl Agent {
             let todo_content = if let Some(path) = session_file_path {
```
DOsinga, line 495: We should figure out how to lift this out of agent; finding the path to find the config to read the state for a specific tool is quite cumbersome; it also strikes me that the state here should be per extension, not per tool; we have a read_todo and a write_todo tool, but unfortunately they are not part of an actual extension. thoughts?

DOsinga (commented):
some preliminary thoughts...

crates/goose/src/session/tool_state.rs
katzdave: No can prune; generally looking to get something where we're confident we could migrate the state, but without implementing all of that yet.

Maybe even we should go one layer up onto SessionMetadata as that's a bit sensitive to field changes now (although not much there but some constants yet).

crates/goose/src/session/tool_state.rs
katzdave: Where would it go? Seems kind of session adjacent to me if we want it to be scoped to that + serialized and deserialized with the session.

crates/goose/src/context_mgmt/auto_compact.rs
katzdave: I gave this a quick try, but seems not much different and then need to null check it.

crates/goose/src/agents/agent.rs
katzdave: Yeah I'm with you that this is the messiest part.  I was somewhat thinking of 'TODO' as an abstract tool but you're right with the current implementation it's really 2 different tools. So maybe re-implementing it as a pseudo extension with multiple tool calls (looks very similar in shape to how you'd make an mcp server) would help.

I think it would be really cool to in general support session scoped storage for tools, although not really part of the mcp spec so probably only relevant/helpful for the internal tools we build. 

I guess trying to figure out what we want in the shortest term pre-next release.

Seems like options are:
1) Roll back todo -> session state PR. This would leave the tool as disabled, as the original agent storage had the leaky state issue in desktop.
2) Write todo state to a different file alongside session_state.txt file and just parse it as a string (we're touching the filesystem already). Then at least we don't pollute session metadata and when we change to a fancier session storage we can just stop reading this file.
3) Continue from this direction and build out a v1 of session tool state.

I'm thinking (2).

PR #4364, opened 2025-08-27
fix: ollama streaming patch

michaelneale:
building off https://github.com/block/goose/pull/4303 - correctly sets the strreaming flag and actually does stream, improves performance of ollama a bit at least subjectively

thanks @Jovonni

### DIFF
--- a/crates/goose/src/providers/ollama.rs
+++ b/crates/goose/src/providers/ollama.rs
@@ -1,21 +1,28 @@
 use super::api_client::{ApiClient, AuthMethod};
-use super::base::{ConfigKey, Provider, ProviderMetadata, ProviderUsage, Usage};
+use super::base::{ConfigKey, MessageStream, Provider, ProviderMetadata, ProviderUsage, Usage};
 use super::errors::ProviderError;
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/ollama.rs
```diff
@@ -228,6 +237,45 @@ impl Provider for OllamaProvider {
...
+            let message_stream = response_to_streaming_message(framed);
+            pin!(message_stream);
+            while let Some(message) = message_stream.next().await {
+                let (message, usage) = message.map_err(|e| ProviderError::RequestFailed(format!("Stream decode error: {}", e)))?;
+                super::utils::emit_debug_trace(&model_config, &payload, &message, &usage.as_ref().map(|f| f.usage).unwrap_or_default());
+                yield (message, usage);
+            }
+        }))
+    }
 }
```
DOsinga, line 279: this is identical to the streaming implementation of OpenAI. we should reuse that code

DOsinga (approved):
Looks good, but let's dedupe the code

michaelneale:
closing as folded in

PR #4371, opened 2025-08-27
Provider guard and effect dep

jamadeo:


ui/desktop/src/App.tsx
```diff
@@ -577,99 +577,48 @@ export default function App() {
...
+                      setChat={setChat}
+                      agentState={agentState}
+                      loadCurrentChat={loadCurrentChat}
+                      setFatalError={setFatalError}
+                      setAgentWaitingMessage={setAgentWaitingMessage}
+                      setIsGoosehintsModalOpen={setIsGoosehintsModalOpen}
+                    />
                   }
                 />
+                <Route path="settings" element={<SettingsRoute />} />
```
DOsinga, line 615: ❗

DOsinga (approved):

ui/desktop/src/App.tsx
jamadeo: oh yeah, to explain this -- in the onboarding, with an empty config, we loaded the onboarding UI but with a non-functional sidebar present. Wrapping the main appcontent solves that because the sidebar is part of it, and it also just makes more sense than wrapping each individual route anyway IMO

PR #4399, opened 2025-08-28
Add metrics for recipe metadata in scheduler, UI, and CLI

Kvadratni:
## Pull Request Description

- Adds additional recipe related telemetry to the UI 
- Adds tags to CLI recipe telemetry

### DIFF
--- a/crates/goose-server/src/routes/reply.rs
+++ b/crates/goose-server/src/routes/reply.rs
@@ -91,6 +91,8 @@ struct ChatRequest {
     session_id: Option<String>,
     session_working_dir: String,
     scheduled_job_id: Option<String>,
+    recipe_name: Option<String>,
+    recipe_version: Option<String>,
 }
 
 pub struct SseResponse {
@@ -182,6 +184,19 @@ async fn reply_handler(
         interface = "ui",
         "Session started"
     );
+
+    if let Some(recipe_name) = &request.recipe_name {
+        let recipe_version = request.recipe_version.as_deref().unwrap_or("unknown");
+
+        tracing::info!(
+            counter.goose.recipe_runs = 1,
+            recipe_name = %recipe_name,
+            recipe_version = %recipe_version,
+            session_type = "app",
+            interface = "ui",
+            "Recipe execution started"
+        );
+    }
 
     let (tx, rx) = mpsc::channel(100);
     let stream = ReceiverStream::new(rx);
@@ -585,6 +600,8 @@ mod tests {
                         session_id: Some("test-session".to_string()),
                         session_working_dir: "test-working-dir".to_string(),
                         scheduled_job_id: None,
+                        recipe_name: None,
+                        recipe_version: None,
                     })
                     .unwrap(),
                 ))
--- a/crates/goose-server/src/routes/schedule.rs
+++ b/crates/goose-server/src/routes/schedule.rs
@@ -231,6 +231,53 @@ async fn run_now_handler(
         .await
         .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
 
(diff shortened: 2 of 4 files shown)


lifeizhou-ap:
Looks good, would you please test it to make sure it works. Thank you!  If it works, then I'll approve it!

lifeizhou-ap (approved):

crates/goose-server/src/routes/reply.rs
```diff
@@ -91,6 +91,8 @@ struct ChatRequest {
     session_id: Option<String>,
     session_working_dir: String,
     scheduled_job_id: Option<String>,
+    recipe_name: Option<String>,
+    recipe_version: Option<String>,
```
DOsinga, line 95: do we need to pass this in? do we not have this information in the session meta data, that seems like a better place. same actually for the schedule_job_id I think.

crates/goose-server/src/routes/schedule.rs
```diff
@@ -231,6 +231,53 @@ async fn run_now_handler(
...
+                                .to_string_lossy()
+                                .to_string(),
+                        )
+                        .ok()
+                        .map(|(r, _)| r.version)
+                    })
+            })
+            .unwrap_or_else(|| "unknown".to_string()),
+        Err(_) => "unknown".to_string(),
+    };
```
DOsinga, line 270: you filtering through the list of jobs twice here, once to find the name, once to find the version? make that into one thing

ui/desktop/src/hooks/useChatEngine.ts
```diff
@@ -101,7 +101,16 @@ export const useChatEngine = ({
     api: getApiUrl('/reply'),
     id: chat.id,
     initialMessages: chat.messages,
-    body: { session_id: chat.id, session_working_dir: window.appConfig.get('GOOSE_WORKING_DIR') },
+    body: {
+      session_id: chat.id,
+      session_working_dir: window.appConfig.get('GOOSE_WORKING_DIR'),
+      ...(chat.recipeConfig?.title
+        ? {
```
DOsinga, line 108: title is required, so you want to check on recipeConfig here, not title

crates/goose-server/src/routes/reply.rs
Kvadratni: we do not have it currently no

ui/desktop/src/hooks/useChatEngine.ts
Kvadratni: ah good point

Kvadratni: that's what this is doing
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

crates/goose-server/src/routes/schedule.rs
Kvadratni: yeah that's on me. let me fix this quick

ui/desktop/src/hooks/useChatEngine.ts
DOsinga: It's checking chat.recipeConfig?.title not hat.recipeConfig

crates/goose-server/src/routes/reply.rs
DOsinga: so we want to move the source of truth for conversations to the server, not add stuff to the reply handler. I also don't think this does record what you want it to record. I think you want to measure how often recipes are used, not how many messages users send to recipes. not that it is doing that, if you resume a conversation, it won't remember it was started by a recipe.

I would suggest to wait with this until the https://github.com/block/goose/pull/4216/files lands and then add the logging to agent/start which should know whether it was done through a recipe (and if not, that would be the right place to add).

PR #4408, opened 2025-08-28
Improve theme loading and apply across windows

### LINKED ISSUE #4245: Dark mode wonkiness
**Describe the bug**

dark/light classes are not getting applied to the root on app load, but do once you click over to the app tab under settings

sometimes when working on local goose dev via the electron app and a released version of goose where one of the two apps will also open in the inverse theme state as the other

I also noticed this happens when the system default settings take hold, clicking on app settings changes it to respect the system default so it might be some override failing with system default settings

zanesq:
Improve theme loading / switching to be more resilient and apply across windows

- Add better initialization with proper light/dark management and retries
- Added theme synchronization utility to apply the current theme across windows (currently only applies to the window its changed in)
- Ensure theme is applied on load and early in the component lifecycle 

fixes https://github.com/block/goose/issues/4245


zanesq:
Noticed some more edge cases moved this to draft while I debug them

PR #4409, opened 2025-08-28
fix double filtering

Kvadratni:
## Pull Request Description

fix for comment in 
https://github.com/block/goose/pull/4399/files/d0a11254837dc7d7eb5a2299292a07a33803c24c#r2308781571

### DIFF
--- a/crates/goose-server/src/routes/schedule.rs
+++ b/crates/goose-server/src/routes/schedule.rs
@@ -231,26 +231,16 @@ async fn run_now_handler(
         .await
         .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
 
-    let recipe_display_name = match scheduler.list_scheduled_jobs().await {
-        Ok(jobs) => jobs
-            .into_iter()
-            .find(|job| job.id == id)
-            .and_then(|job| {
-                std::path::Path::new(&job.source)
+    let (recipe_display_name, recipe_version) = match scheduler.list_scheduled_jobs().await {
+        Ok(jobs) => {
+            if let Some(job) = jobs.into_iter().find(|job| job.id == id) {
+                let recipe_display_name = std::path::Path::new(&job.source)
                     .file_name()
                     .and_then(|name| name.to_str())
                     .map(|s| s.to_string())
-            })
-            .unwrap_or_else(|| id.clone()),
-        Err(_) => id.clone(),
-    };
+                    .unwrap_or_else(|| id.clone());
 
-    let recipe_version = match scheduler.list_scheduled_jobs().await {
-        Ok(jobs) => jobs
-            .into_iter()
-            .find(|job| job.id == id)
-            .and_then(|job| {
-                std::fs::read_to_string(&job.source)
+                let recipe_version = std::fs::read_to_string(&job.source)
                     .ok()
                     .and_then(|content| {
                         goose::recipe::template_recipe::parse_recipe_content(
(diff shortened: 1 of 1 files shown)


crates/goose-server/src/routes/schedule.rs
```diff
@@ -264,9 +254,14 @@ async fn run_now_handler(
...
                         .map(|(r, _)| r.version)
                     })
-            })
-            .unwrap_or_else(|| "unknown".to_string()),
-        Err(_) => "unknown".to_string(),
+                    .unwrap_or_else(|| "unknown".to_string());
+
+                (recipe_display_name, recipe_version)
+            } else {
+                (id.clone(), "unknown".to_string())
```
DOsinga: is this really what you want to do? list the thing as unknown when we can't find it? I think doing a filter here would be better

crates/goose-server/src/routes/schedule.rs
```diff
@@ -231,26 +231,16 @@ async fn run_now_handler(
...
-    };
+                    .unwrap_or_else(|| id.clone());
 
-    let recipe_version = match scheduler.list_scheduled_jobs().await {
-        Ok(jobs) => jobs
-            .into_iter()
-            .find(|job| job.id == id)
-            .and_then(|job| {
-                std::fs::read_to_string(&job.source)
+                let recipe_version = std::fs::read_to_string(&job.source)
```
DOsinga: we should do this async and not block the server

crates/goose-server/src/routes/schedule.rs
```diff
@@ -231,26 +231,16 @@ async fn run_now_handler(
...
+            if let Some(job) = jobs.into_iter().find(|job| job.id == id) {
+                let recipe_display_name = std::path::Path::new(&job.source)
                     .file_name()
                     .and_then(|name| name.to_str())
                     .map(|s| s.to_string())
-            })
-            .unwrap_or_else(|| id.clone()),
-        Err(_) => id.clone(),
-    };
+                    .unwrap_or_else(|| id.clone());
```
DOsinga, line 241: should we just use the name of recipe, rather than basing something on the filename?

DOsinga (approved):

crates/goose-server/src/routes/schedule.rs
Kvadratni: I don't want to return null's here

crates/goose-server/src/routes/schedule.rs
Kvadratni: fair point

crates/goose-server/src/routes/schedule.rs
Kvadratni: I was contemplating it. but we do use the filename as a more unique indicator. I can have 7 recipes with the same title, but filenames will be unique

DOsinga: I dont want to be difficult here, but the variable is called recipe_display_name, so showing the name of the recipe seems kinda reasonable. the plan is to merge recipes and scheduler either way and do away with the extra naming layers we have everywhere, fwiw

crates/goose-server/src/routes/schedule.rs
DOsinga: well, if you can't find the thing here, what's the point of passing on id to the next thing? it won't be able to run the scheduler, will it? so I think returning null here and then just abort with an error of not found or some such is probably the right thing to do?

crates/goose-server/src/routes/schedule.rs
Kvadratni: ah good to know. Can we just leave this until then to avoid redundant work?

PR #4417, opened 2025-08-29
Disable chat input while extensions load

amed-xyz:
Users can currently initiate chat conversations before extensions are fully loaded, which causes unexpected behaviour.


ui/desktop/src/components/ChatInput.tsx
```diff
@@ -1239,6 +1241,7 @@ export default function ChatInput({
             onBlur={() => setIsFocused(false)}
             ref={textAreaRef}
             rows={1}
+            disabled={isExtensionsLoading}
```
zanesq: I wonder if we need `!agentIsReady` also?

amed-xyz: Yeah, I also wonder about the other conditions the submit button has, should they have 100% parity?

```
!hasSubmittableContent ||
isAnyImageLoading ||
isAnyDroppedFileLoading ||
isRecording ||
isTranscribing ||
isCompacting ||
!agentIsReady ||
isExtensionsLoading
```

amed-xyz:
@zanesq good catch, `!hasSubmittableContent` was preventing the textarea from ever becoming enabled. I've updated the branch with latest master and committed a fix

PR #4421, opened 2025-08-29
fix: subagent hardening

michaelneale:
subagents now work well and are on by default, but the reporting of them wasn't great, this fixes that. 

This also updates the system prompt which no longer needs to reference old models, and tunes instructions for making use of the subagents when appropriate.


michaelneale:
closing as @tlongwell-block's change we collaborated on does this.

PR #4422, opened 2025-08-29
chore: move list recipes and archive recipe to goose server

lifeizhou-ap:
## Pull Request Description

Move list recipes and archive recipe logic from UI to goose server

## What

- Created RecipeManifest that maps to local and global recipe files
- Create `/recipes/list` route to render the recipe files in the global recipe directory and current working directory
- Create `/recipes/archive` route to archive recipe

## Next PR
move load recipe to server side


crates/goose-server/src/state.rs
```diff
@@ -10,6 +12,7 @@ pub struct AppState {
     agent: Option<AgentRef>,
     pub secret_key: String,
     pub scheduler: Arc<Mutex<Option<Arc<dyn SchedulerTrait>>>>,
+    pub recipe_file_hash_map: Arc<Mutex<HashMap<String, PathBuf>>>,
```
lifeizhou-ap, line 15: This in memory map maintains the mapping of  recipe id (hashed based on recipe path) and recipe path.  recipe id is the key of map

ui/desktop/src/components/RecipesView.tsx
```diff
@@ -898,18 +898,6 @@ Parameters you can use:
                   </div>
                 </div>
               )}
-
-              {selectedRecipe.recipe.goosehints && (
```
lifeizhou-ap, line 925: Hi @zanesq I am not sure whether this is really used in RecipeView UI. It seems not.  same as profile and mcps.

ui/desktop/src/recipe/recipeStorage.ts
```diff
@@ -162,183 +163,18 @@ export async function loadRecipe(recipeName: string, isGlobal: boolean): Promise
...
   }
 }
 
-/**
- * Restore an archived recipe.
- *
- * @param recipeName The name of the recipe to restore
- * @param isGlobal Whether the recipe is in global or local storage
- */
-export async function restoreRecipe(recipeName: string, isGlobal: boolean): Promise<void> {
```
lifeizhou-ap, line 234: This function is not used at all

ui/desktop/src/recipe/recipeStorage.ts
```diff
@@ -162,183 +163,18 @@ export async function loadRecipe(recipeName: string, isGlobal: boolean): Promise
...
-  }
-}
-
-/**
- * Permanently delete a recipe file.
- *
- * @param recipeName The name of the recipe to permanently delete
- * @param isGlobal Whether the recipe is in global or local storage
- */
-export async function permanentlyDeleteRecipe(
```
lifeizhou-ap, line 304: This function is not used at all

ui/desktop/src/recipe/recipeStorage.ts
```diff
@@ -162,183 +163,18 @@ export async function loadRecipe(recipeName: string, isGlobal: boolean): Promise
...
-}
-
-/**
- * Delete a recipe (archives it by default for backward compatibility).
- *
- * @deprecated Use archiveRecipe instead
- * @param recipeName The name of the recipe to delete/archive
- * @param isGlobal Whether the recipe is in global or local storage
- */
-export async function deleteRecipe(recipeName: string, isGlobal: boolean): Promise<void> {
```
lifeizhou-ap, line 341: This function is not used at all

ui/desktop/src/recipe/recipeStorage.ts
```diff
@@ -162,183 +163,18 @@ export async function loadRecipe(recipeName: string, isGlobal: boolean): Promise
...
-  }
-}
-
-/**
- * Archive a recipe.
- *
- * @param recipeName The name of the recipe to archive
- * @param isGlobal Whether the recipe is in global or local storage
- */
-export async function archiveRecipe(recipeName: string, isGlobal: boolean): Promise<void> {
```
lifeizhou-ap, line 269: this function is no longer required as we perform archiveRecipe via api call

ui/desktop/src/components/RecipesView.tsx
```diff
@@ -372,90 +375,69 @@ Parameters you can use:
     }
   };
 
-  // Render a recipe item with error handling
-  const RecipeItem = ({ savedRecipe }: { savedRecipe: SavedRecipe }) => {
-    try {
```
lifeizhou-ap, line 377: removed try/catch here as the server side will ignore the yaml files that do not conform to recipe format.

ui/desktop/src/recipe/recipeStorage.ts
```diff
@@ -165,181 +166,23 @@ export async function loadRecipe(recipeName: string, isGlobal: boolean): Promise
...
-    );
+export function recipeLastModified(lastModified: string): string {
+  if (lastModified) {
+    return parseLastModified(lastModified).toLocaleDateString();
   }
-}
-
-/**
- * Delete a recipe (archives it by default for backward compatibility).
- *
```
DOsinga, line 336: This is just all LLM generated non-sense - there was never backward compatibility here

crates/goose/Cargo.toml
```diff
@@ -99,6 +99,7 @@ unicode-normalization = "0.1"
 
 arrow = "52.2"
 oauth2 = "5.0.0"
+xxhash-rust = { version = "0.8.12", features = ["xxh3"] }
```
DOsinga: I don't think we need yet another hashing library, ahash should do the trick

crates/goose/src/recipe/recipe_manifest.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    pub fn save_to_yaml_file(self, path: &Path) -> Result<()> {
+        let content = serde_yaml::to_string(&self)
+            .map_err(|e| anyhow::anyhow!("Failed to serialize YAML: {}", e))?;
+        fs::write(path, content)
+            .map_err(|e| anyhow::anyhow!("Failed to write file {}: {}", path.display(), e))?;
+        Ok(())
+    }
+
+    pub fn archive(file_path: &Path) -> Result<()> {
```
DOsinga: let's drop archive. we have no way to restore archived recipes it seems and we call it delete. let's delete them.

crates/goose/src/recipe/recipe_manifest.rs
```diff
@@ -0,0 +1,45 @@
+use anyhow::Result;
+use chrono::Utc;
+use std::{fs, path::Path};
+
+use crate::recipe::Recipe;
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+
+#[derive(Serialize, Deserialize, Debug, Clone, ToSchema)]
+pub struct RecipeManifest {
```
DOsinga: this RecipeManifest in the client was a mistake. they should never be saved to disk - @jsibbison-square already build some compatibility in so we can read this client side version of recipes.

In the client we just list the recipes and don't show is_global anywhere. name is a duplicated. last_modified should not be stored either, that's just when the file got updated. 

in short, we shouldn't need this at all. just load recipes from disk and if the client needs it, ad the modification date (if you need an intermediate class for that, fine, but just in routes)

crates/goose/src/recipe/list_recipes.rs
```diff
@@ -0,0 +1,68 @@
...
+                    recipe_manifests_with_path.push(manifest_with_path);
+                }
+            }
+        }
+    }
+    Ok(recipe_manifests_with_path)
+}
+
+fn get_all_recipes_manifests() -> Result<Vec<RecipeManifestWithPath>> {
+    let current_dir = std::env::current_dir()?;
```
DOsinga: this should not use the current_dir, but use the current dir of the session

crates/goose/src/recipe/list_recipes.rs
```diff
@@ -0,0 +1,68 @@
...
+    let local_recipe_path = current_dir.join(".goose/recipes");
+
+    let global_recipe_path = choose_app_strategy(APP_STRATEGY.clone())
+        .map(|strategy| strategy.in_config_dir("recipes"))
+        .unwrap_or_else(|_| PathBuf::from("~/.config/goose/recipes"));
```
DOsinga: this should not hardcode .config -  app_strategy has a handler to get to I think the data_dir?

crates/goose/src/recipe/list_recipes.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub manifest: RecipeManifest,
+    pub file_path: PathBuf,
+}
+
+fn short_id_from_path(path: &str) -> String {
+    let hash = xxh3_64(path.as_bytes());
+    format!("{:016x}", hash)
+}
+
+fn load_recipes_from_path(path: &PathBuf) -> Result<Vec<RecipeManifestWithPath>> {
```
DOsinga: we already have read_recipe_in_dir, use that. this would be a good opportunity to unify the recipes the user sees in CLI and the desktop

ui/desktop/src/recipe/recipeStorage.ts
```diff
@@ -1,3 +1,4 @@
+import { listRecipes, RecipeManifestResponse } from '../api';
 import { Recipe } from './index';
 import * as yaml from 'yaml';
```
DOsinga, line 3: I think this entire file should now be deleted

ui/desktop/src/components/RecipesView.tsx
```diff
@@ -372,90 +375,69 @@ Parameters you can use:
...
-                </h3>
-              </div>
-              <p className="text-red-600 dark:text-red-400 text-sm">
-                {error instanceof Error ? error.message : 'Unknown error'}
-              </p>
-            </div>
+          <p className="text-text-muted text-sm mb-2 line-clamp-2">{manifest.recipe.description}</p>
+          <div className="flex items-center text-xs text-text-muted">
+            <Calendar className="w-3 h-3 mr-1" />
+            {recipeLastModified(manifest.lastModified)}
```
DOsinga: do this locally, not worth two calls

ui/desktop/src/recipe/recipeStorage.ts
zanesq: It might have been accidentally dropped with the ui refactor should we add it back to the ui instead? Seems like we need a way to delete recipes still?

ui/desktop/src/components/RecipesView.tsx
zanesq: Good catch I was wondering the same thing, might have been legacy functionality that we can drop now.

crates/goose/src/recipe/list_recipes.rs
lifeizhou-ap: where working directory is changed on UI, it started goosed with the working directory.  

So current_dir here is the session working directory.

ui/desktop/src/recipe/recipeStorage.ts
zanesq: My bad we do have delete already nvm on this comment go ahead and remove it 👍

lifeizhou-ap: Hi @zanesq,

Your previous comment is right!

Before this PR, delete is actually "archive".  Talked to @DOsinga and he suggested we delete recipe directly instead of archive because there is no way to restore.  

Now the delete is "delete the recipe file" and it used the server side to delete

ui/desktop/src/recipe/recipeStorage.ts
lifeizhou-ap: It will. There are still some codes for save and load.  When we move these recipe operations to server side, then we don't need it

ui/desktop/src/components/RecipesView.tsx
lifeizhou-ap: i renamed `recipeLastModified` to `convertToLocaleDateString`. basically it convert the utc time string to local date

crates/goose/src/recipe/recipe_manifest.rs
lifeizhou-ap: I moved the list_recipe, manifest related object sas view objects to goose-server. 

list_recipes is now calling goose core to get the recipe objects

If cli needs list recipes in the future, we can move list_recipes (only get the recipe objects part, no the manifest) to goose core

crates/goose/src/recipe/list_recipes.rs
DOsinga: yeah, I know that is currently the case, but once we move to having multiple agents per goosed, this will no longer be the case

crates/goose-server/src/recipe/recipe_manifest_metadata.rs
```diff
@@ -0,0 +1,48 @@
+use anyhow::Result;
```
DOsinga: this file should not exist

crates/goose-server/src/recipe/list_recipes.rs
```diff
@@ -0,0 +1,85 @@
+use std::fs;
```
DOsinga: we shouldn't need this, just read the recipes just like we always have

PR #4432, opened 2025-08-29
remove localstorage getconfig fallbacks

zanesq:
## Pull Request Description
remove localstorage getconfig fallbacks since they aren't needed
- Configuration is now passed entirely through process.argv in webPreferences.additionalArguments
- No need for localStorage fallback - the main process provides complete config

Verified things are loading locally and in the bundle.

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -657,56 +657,6 @@ const createChat = async (
     menu.popup();
   });
 
-  // Store config in localStorage for future windows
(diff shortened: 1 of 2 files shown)


zanesq:
.bundle

zanesq:
Verified working in bundle

zanesq:
This doesn't fix the refresh issue but I still think we should remove these getConfig fallbacks anyway. App is working as expected otherwise.

PR #4434, opened 2025-08-29
Add Docker support for Goose in CI/CD pipelines

tlongwell-block:
Adds a production-ready Docker image for Goose that enables using Goose itself in CI pipelines to improve the codebase. The image is optimized for size (365MB) and includes both the CLI and server binaries.

## Primary Goal: Goose in CI

This enables Goose to help maintain and improve itself through automated workflows:
- Daily test coverage finder that creates PRs with missing tests
- Code quality audits
- Security scanning
- Performance analysis

Example workflow included that runs daily to find untested code and automatically create a PR with a test.

## Bonus: Easy Docker Usage for Everyone

While built for CI, this also makes Goose super easy to run in Docker:

**CLI usage:**
```bash
docker run --rm ghcr.io/block/goose:latest --help
```

**Server (goosed) hosting:**
```bash
docker run -p 3000:3000 --entrypoint goosed ghcr.io/block/goose:latest
```

Users can now:
- Run Goose CLI without installing Rust/dependencies
- Host goosed instances in containers
- Use with existing Goose Desktop frontend or custom UIs
- Integrate into their own CI/CD pipelines

## What's Included

- Optimized multi-stage Dockerfile (365MB final image)
- GitHub Actions workflow to publish images to GHCR
- Example "test finder" workflow showing Goose improving itself
- Documentation in BUILDING_DOCKER.md
- Support for all LLM providers via environment variables

## Technical Details

- Base: Debian Bookworm Slim
- Includes: goose CLI (32MB) + goosed server (25MB)
- Multi-platform: linux/amd64 and linux/arm64
- Auto-publishes: `latest` on main, version tags on releases
- Non-root user for security

This is step 1 toward having Goose actively participate in its own development through automated improvements.

### DIFF
--- a/.github/workflows/test-finder.yml
+++ b/.github/workflows/test-finder.yml
@@ -26,68 +26,76 @@ jobs:
         GOOSE_PROVIDER: ${{ vars.GOOSE_PROVIDER || 'openai' }}
         GOOSE_MODEL: ${{ vars.GOOSE_MODEL || 'gpt-4o' }}
         OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
-    
+
     steps:
       - name: Checkout code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # pin@v4
         with:
           fetch-depth: 0
-      
+
       - name: Install analysis tools
         run: |
           apt-get update
           apt-get install -y jq ripgrep
-      
-      - name: Find untested code
+
+      - name: Find untested code and create working test
         id: find_untested
         run: |
-          # Create analysis script
-          cat << 'EOF' > /tmp/find_untested.txt
-          Analyze the Rust codebase and find ONE non-trivial function that lacks test coverage.
-          
+          # Create analysis and test creation script
+          cat << 'EOF' > /tmp/create_working_test.txt
+          Your task is to find ONE untested function in the Rust codebase and create a working test for it.
+
           Requirements:
-          1. The function should be in the crates/ directory
-          2. It should have actual logic (not just a simple getter/setter)
-          3. It should not already have a test
+          1. The function MUST be in the crates/ directory
+          2. It MUST have actual logic (not just a simple getter/setter)
+          3. It MUST not already have a test
           4. Prefer functions with complexity but that are still testable in isolation
           5. Focus on the goose crate first, then goose-cli, then others
-          
-          Once you find a good candidate:
-          1. Create a single, focused unit test for that function
-          2. The test should be comprehensive but concise
-          3. Place the test in the appropriate test module or create one if needed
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+RUN useradd -m -u 1000 -s /bin/bash goose && \
+    mkdir -p /home/goose/.config/goose && \
+    chown -R goose:goose /home/goose
+
+# Set up environment
+ENV PATH="/usr/local/bin:${PATH}"
+ENV HOME="/home/goose"
+
+# Expose goosed default port
+EXPOSE 3000
```
alexhancock: I would leave goose-server / `goosed` out of this for now

We need to come up with a better interface / API for it than we have now. I wouldn't want anything else to be built against the current interface that we will need to support and then port when we switch to something better

I think we can go a long way with just goose-cli and `goose run -i` commands in CI for a while

.github/workflows/test-finder.yml
```diff
@@ -0,0 +1,135 @@
...
+      
+      - name: Find untested code
+        id: find_untested
+        run: |
+          # Create analysis script
+          cat << 'EOF' > /tmp/find_untested.txt
+          Analyze the Rust codebase and find ONE non-trivial function that lacks test coverage.
+          
+          Requirements:
+          1. The function should be in the crates/ directory
```
alexhancock: I'd guess we'll get better results by changing all these `should`s to `MUST`

.github/workflows/test-finder.yml
```diff
@@ -0,0 +1,135 @@
...
+            FUNC_NAME=$(grep -E "fn test_|#\[test\]" /tmp/test_addition.patch | head -1 | sed 's/.*test_//' | sed 's/(.*//' || echo "function")
+            echo "function_name=${FUNC_NAME}" >> $GITHUB_OUTPUT
+          else
+            echo "patch_created=false" >> $GITHUB_OUTPUT
+          fi
+      
+      - name: Apply and verify patch
+        if: steps.find_untested.outputs.patch_created == 'true'
+        run: |
+          # Apply the patch
```
alexhancock: I would roll this into one thing with the above step. We should give instructions to:

1. find untested code and write a test (what you already have in the step above)
1. apply the change
1. run the test
1. loop until passing
1. report success or no-tests-added
1. commit and PR (if it made a test)

Dockerfile
tlongwell-block: Done

.github/workflows/test-finder.yml
tlongwell-block: Done 👍

.github/workflows/test-finder.yml
tlongwell-block: Done 👍

PR #4447, opened 2025-08-29
Add visual indicator while recipe loads

amed-xyz:
Add a toast loading indicator during recipe loading phase to provide visual feedback.

- Toast notification appears immediately when recipe starts loading
- Success toast confirms when recipe is ready

### DIFF
--- a/ui/desktop/src/utils/appInitialization.ts
+++ b/ui/desktop/src/utils/appInitialization.ts
@@ -145,7 +145,7 @@ const initializeForRecipe = async ({
 
   const loadingToastId = toastService.loading({
     title: `Loading recipe: ${recipeConfig.title}`,
-    msg: 'Setting up extensions and environment...',
+    msg: 'Setting up environment...',
   });
 
   await initConfig();


ui/desktop/src/utils/appInitialization.ts
```diff
@@ -148,6 +156,9 @@ const initializeForRecipe = async ({
     addExtension,
   });
 
+  toastService.dismiss(loadingToastId);
+  toastService.success({ title: 'Recipe ready!', msg: `${recipeConfig.title} is ready to use` });
```
zanesq: Lets just leave it generic so it fits better in the toast, no recipe title needed so maybe something like

```
title: 'Recipe loaded', msg: `Recipe is ready to use`
```

ui/desktop/src/utils/appInitialization.ts
```diff
@@ -140,6 +141,13 @@ const initializeForRecipe = async ({
 > & {
   recipeConfig: Recipe;
 }) => {
+  toastService.configure({ silent: false });
+
+  const loadingToastId = toastService.loading({
+    title: `Loading recipe: ${recipeConfig.title}`,
+    msg: 'Setting up extensions and environment...',
```
amed-xyz:
```suggestion
    msg: 'Setting up environment...',
```

PR #4454, opened 2025-08-30
Remove To-Do tool instructions from system prompt

tlongwell-block:
### What This Does
Moves the hardcoded TODO task management instructions out of the system prompt and into an extension, making them modular like other built-in extensions.

### Key Changes
- Created a lightweight `TodoRouter` extension that provides instructions and tool definitions
- Removed hardcoded "Task Management" section from the system prompt
- TODO tools execution stays in the agent (unchanged) - only the instructions moved

### Why Keep Execution in the Agent?
MCP extensions are designed to be stateless, isolated processes. TODO tools need direct access to session storage for persistence. Rather than building complex state synchronization infrastructure (~800 lines), we use a "metadata-only" extension pattern:

- The extension provides instructions and tool definitions
- The agent handles execution where it naturally has session access

This is the right architecture - not a compromise. TODO tools are fundamentally part of session management, so keeping execution in the agent while making instructions modular gives us the best of both worlds.

### Testing
- All existing tests pass
- TODO tools work exactly as before
- Instructions now load from extension

### DIFF
--- a/crates/goose-mcp/Cargo.toml
+++ b/crates/goose-mcp/Cargo.toml
@@ -11,6 +11,7 @@ description.workspace = true
 workspace = true
 
 [dependencies]
+goose = { path = "../goose" }
 mcp-core = { path = "../mcp-core" }
 mcp-server = { path = "../mcp-server" }
 rmcp = { workspace = true }
--- a/crates/goose-mcp/src/todo/mod.rs
+++ b/crates/goose-mcp/src/todo/mod.rs
@@ -1,0 +1,134 @@
+use std::future::Future;
+use std::pin::Pin;
+
+use mcp_core::handler::{PromptError, ResourceError};
+use mcp_core::protocol::ServerCapabilities;
+use mcp_server::router::CapabilitiesBuilder;
+use mcp_server::Router;
+use rmcp::model::{Content, ErrorCode, ErrorData, JsonRpcMessage, Prompt, Resource, Tool};
+use serde_json::Value;
+use tokio::sync::mpsc;
+
+// Import the TODO tools
+use goose::agents::todo_tools::{todo_read_tool, todo_write_tool};
+
+/// A lightweight router that provides TODO task management capabilities.
+///
+/// This extension acts as a metadata provider - it exposes the TODO tools
+/// and provides instructions, but the actual execution remains in the agent
+/// for session storage access.
+#[derive(Clone)]
+pub struct TodoRouter {
+    tools: Vec<Tool>,
+    instructions: String,
+}
+
+impl Default for TodoRouter {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl TodoRouter {
+    pub fn new() -> Self {
+        let instructions = r#"The todo extension provides persistent task management throughout your session.
+
+These tools help you track multi-step work, maintain context between interactions, and ensure systematic task completion.
+
+## Task Management Guidelines
+
+**Required Usage:**
+- Use `todo__read` and `todo__write` for any task with 2+ steps, multiple files/components, or uncertain scope
+- Skipping these tools when needed is considered an error
+
+**Workflow:**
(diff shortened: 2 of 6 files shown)


DOsinga:
> MCP extensions are designed to be stateless, isolated processes

they are unfortunately not necessarily stateless

I think we should keep the todo extension in the goose crate but just make it as MCP shared as possible same as we should do for platform tools. the extension manager already has an add_client method I think we should be able to use, or nearly should be able to use

crates/goose-mcp/Cargo.toml
```diff
@@ -11,6 +11,7 @@ description.workspace = true
 workspace = true
 
 [dependencies]
+goose = { path = "../goose" }
```
michaelneale, line 14: yeah this doesn't seem ideal, but at least it is in right direction

crates/goose-mcp/src/todo/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+**Critical:** `todo__write` replaces the entire list. Always read before writing - not doing so is an error.
+
+**Best Practices:**
+- Keep items short, specific, and action-oriented
+- Use nested checkboxes for subtasks
+- Include context about blockers or dependencies
+
+Example format:
+```markdown
+- [x] Analyze request fully
```
michaelneale, line 58: I think if we are moving the prompt here we need to make it clear similar to the old system.md one how to do it, including if we want it to use subagents at times (otherwise I found it will not) - https://github.com/block/goose/pull/4421/files#diff-507a51de1f94529d2674a027f5bf6e5eeed83ccfe9cbbe6cea674b75aef65ee8 like that

tlongwell-block:
> > MCP extensions are designed to be stateless, isolated processes
> 
> they are unfortunately not necessarily stateless
> 
> I think we should keep the todo extension in the goose crate but just make it as MCP shared as possible same as we should do for platform tools. the extension manager already has an add_client method I think we should be able to use, or nearly should be able to use

Okay, just to take a crack at what this would look like: https://github.com/block/goose/pull/4465

That is just a *rough* first pass. But it works. So we can decide if we prefer the To-Do tools to be more like the platform tools or the internal MCPs

crates/goose-mcp/src/todo/mod.rs
tlongwell-block: Yes, I totally agree. I've been doing benching with this instruction here and it works well https://github.com/block/goose/pull/4311/files#diff-507a51de1f94529d2674a027f5bf6e5eeed83ccfe9cbbe6cea674b75aef65ee8R76

https://github.com/block/goose/pull/4311/files#diff-507a51de1f94529d2674a027f5bf6e5eeed83ccfe9cbbe6cea674b75aef65ee8R58

I hope we can merge in that subagent PR soon-ish

tlongwell-block:
> > MCP extensions are designed to be stateless, isolated processes
> 
> they are unfortunately not necessarily stateless
> 
> I think we should keep the todo extension in the goose crate but just make it as MCP shared as possible same as we should do for platform tools. the extension manager already has an add_client method I think we should be able to use, or nearly should be able to use

Rough draft https://github.com/block/goose/pull/4465

michaelneale:
@tlongwell-block do you want to close this in favour of the draft one? 

Also https://github.com/block/goose/pull/4311 - what is relationship with that PR?

tlongwell-block:
> @tlongwell-block do you want to close this in favour of the draft one?
> 
> Also #4311 - what is relationship with that PR?

#4311  is not related to this and should be considered separate

This PR, 4454, was meant to be the minimal solution to removing the To-Do tool instructions from the system prompt.

#4465 is just an alternative implementation of this PR after @DOsinga commented about trying this out as an internal MCP style approach

My personal preference is probably just to leave it in the system prompt and close both PRs just because I think this is a vital part of the agent tooling that is closely tied to how Goose in general will operate

tlongwell-block:
Closing to declutter open PRs since this might take a while to think through

PR #4456, opened 2025-08-31
Cli web auth token

Evanfeenstra:
## Pull Request Description

Adds an optional `--auth-token` option to the `goose web` cli. If defined, all the routes are protected (except the /api/health route).

The auth is implemented in two ways: basic auth for the `goose-cli/static` frontend, as well as a `Authorization: Bearer xxx` header, for api access to `goose web`.

This PR enables running `goose web` on a remote server, and accessing it securely via API


crates/goose-cli/src/cli.rs
```diff
@@ -672,6 +672,14 @@ enum Command {
...
         #[arg(long, help = "Open browser automatically when server starts")]
         open: bool,
+
+        /// Authentication token for both Basic Auth (password) and Bearer token
+        #[arg(
+            long,
+            help = "Optional authentication token to secure the web interface",
+            long_help = "Optional token for authentication. Can be used as password in Basic Auth (username ignored) or as Bearer token in API calls. If not provided, no authentication is required."
+        )]
+        auth_token: Option<String>,
```
DOsinga, line 709: can you remove the comment and where it says it is optional?

crates/goose-cli/src/commands/web.rs
```diff
@@ -77,7 +81,62 @@ enum WebSocketMessage {
...
+            // Check for Basic auth (password-only, ignore username)
+            if let Some(basic_token) = auth_str.strip_prefix("Basic ") {
+                if let Ok(decoded) = base64::engine::general_purpose::STANDARD.decode(basic_token) {
+                    if let Ok(credentials) = String::from_utf8(decoded) {
+                        // Split on first colon, ignore username, check password
+                        if let Some((_, password)) = credentials.split_once(':') {
+                            if password == expected_token {
+                                return Ok(next.run(req).await);
+                            }
+                        }
```
DOsinga, line 113: this you could do with s.ends_with

DOsinga (approved):
thanks for doing this and sorry for the slow repsonse. in general our thinking is that we'll turn goosed into an api server for this sort of thing, but for now this should work well

crates/goose-cli/src/cli.rs
Evanfeenstra: ok done

crates/goose-cli/src/commands/web.rs
Evanfeenstra: ya thats simpler

PR #4463, opened 2025-09-01
hotfix: orphaned `goosed` processes with `just run-ui`

michaelneale:
This is a speculative fix for cases when you end up with unkillable `goosed` on macos after a while running `just run-ui`

in theory if it is quickly killed than the parent goes away and the goosed process never gets told

### DIFF
--- a/ui/desktop/src/goosed.ts
+++ b/ui/desktop/src/goosed.ts
@@ -291,9 +291,25 @@ export const startGoosed = async (
 
   // Ensure goosed is terminated when the app quits
   // TODO will need to do it at tab level next
-  app.on('will-quit', () => {
+  const quitHandler = () => {
     log.info('App quitting, terminating goosed server');
     try_kill_goose();
+  };
+
+  app.on('will-quit', quitHandler);
+  app.on('before-quit', quitHandler);
+
+  // Also handle SIGINT/SIGTERM for development mode when electron-forge kills the process
+  process.on('SIGINT', () => {
+    log.info('Received SIGINT, terminating goosed server');
+    try_kill_goose();
+    process.exit(0);
+  });
+
+  process.on('SIGTERM', () => {
+    log.info('Received SIGTERM, terminating goosed server');
+    try_kill_goose();
+    process.exit(0);
   });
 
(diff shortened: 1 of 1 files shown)


ewann:
fwiw, I haven't applied this hf yet, but I also haven't seen this issue again since applying: https://github.com/block/goose/commit/e6cfdde322d39efc88e155c4a41095b58ed9bfc9 locally.

Probably this PR is better from a defensive standpoint though.

ewann:
> fwiw, I haven't applied this hf yet, but I also haven't seen this issue again since applying: [e6cfdde](https://github.com/block/goose/commit/e6cfdde322d39efc88e155c4a41095b58ed9bfc9) locally.
> 
> Probably this PR is better from a defensive standpoint though.

I've had this fail again with the different `just` version. Also now using this hot fix [locally](https://github.com/thatdogmachine/goose/commit/502a71d6c3666e043cb7d82f6d8fcea96a7689af)

michaelneale:
@ewann LMK if this makes it go away - that will be enough proof for it then (at the moment still draft)

ewann:
[reverted](https://github.com/thatdogmachine/goose/commit/13bad8bcfad6f7460fa58a1a9216bf0ee6a18fb1) this locally, to see if I can re-introduce the issue

michaelneale:
pinging @ewann to see if we can and should merge this

ewann:
> pinging @ewann to see if we can and should merge this

I've not reproduced the initial issue, though that's as much my recent (lack of) work focus on goose in recent days. 

Depending your general approach:

- I struggle to give data this has helped
but
- I haven't experienced a situation where I believe this PR has made the situation worse, and it does appear to offer more robustness

If it were my call, I'd probably merge:
a) the current situation is known to cause a need for (at least) developer OS restarts
b) this PR (to my knowledge) hasn't demonstrably made any situation worse
c) the change seems (to me) unlikely to cause undesirable side effects

michaelneale:
cc @zanesq or @jamadeo  can I get your blessing on this - seems solid, would like to get it in.

michaelneale:
will close this for now, we can open it if needed.

PR #4465, opened 2025-09-01
Refactor To-Do tools as internal MCP extension

tlongwell-block:
THIS IS A ROUGH DRAFT TO SHOW A POC

---

This PR refactors the TODO task management tools from hardcoded platform tools to an internal MCP extension, bringing them in line with our extension architecture.

### What changed
- To-Do tools now register as a proper MCP extension when a session is available
- Created `TodoMcpClient` that implements the standard `McpClientTrait` interface
- Instructions moved from system prompt to extension metadata where they belong

### Why this is better
- **Consistency**: TODO tools follow the same patterns as all other extensions
- **Cleaner architecture**: Tool logic is encapsulated, not scattered through the agent
- **Better testability**: The MCP client can be tested in isolation
- **Smaller system prompt**: Instructions come from the extension, reducing prompt overhead

### Implementation notes
The tools continue to work exactly as before from the user's perspective. Session storage integration is preserved through a  `TodoStorage` trait abstraction. The change touches minimal existing code - mostly just removing the hardcoded handling and adding the registration logic.


michaelneale:
am a little confused - this is in process but looks like an MCP (with a faux-client?)

tlongwell-block:
> am a little confused - this is in process but looks like an MCP (with a faux-client?)

Yes, that's exactly what this is per https://github.com/block/goose/pull/4454#issuecomment-3239496619

tlongwell-block:
Closing to declutter open PRs since this might take a while to think through

PR #4476, opened 2025-09-02
feat: enhance dev tools for bulk edit and code search

michaelneale:
* moves tests into tests.rs to keep the developer toolkit more maintainable
* enhances the editor tool so it can replace all occurrences vs just 1 at a time, to speed up larger edits
* adds a simple `search_code` tool which delegates to `fn bash` to run rg (when available, tool is selectively loaded) which provides much faster location vs iteration of tool calls (and can run in parallel). 

Overall impact should be much faster and more accurate dev tools


michaelneale:
cc @katzdave @salman1993 pretty simple to implement to allow it to bulk update

michaelneale:
grrr.. yeah this will have to be rewritten again.

michaelneale:
@alexhancock might need some advice - with rmcp I don't know how to test these things, so am a bit stuck (was easier before as you could call the tools directly)

crates/goose-mcp/src/developer/mod.rs
```diff
@@ -1545,6 +1635,162 @@ impl DeveloperRouter {
         ])
     }
 
+    #[allow(clippy::too_many_lines)]
+    async fn search_code(&self, params: Value) -> Result<Vec<Content>, ErrorData> {
```
michaelneale, line 1639: @tlongwell-block FYI

michaelneale:
this will be done as a few PRs, @tlongwell-block is working on one that can do bulk edits with diffs, which can include replace_all (probably) as part of it. There will be a separate one for analysis and another for search_code

michaelneale:
closing as being implemented elsewhere

PR #4477, opened 2025-09-02
feat: add streaming support to Tetrate Agent Router Service provider

johnlanda:
## Pull Request Description

Adds streaming support to the Tetrate Agent Router Service provider.


michaelneale:
looking good

michaelneale:
@johnlanda I wasn't able to see it stream - tried the gpt-5 model, should it work? 

(also in the tetrate.rs, it doesnt' mention the sonnet-4 model in the constants - should it?)

Issue #4469, opened 2025-09-01
documentation: Windows Powershell syntax example incorrect


cbruyndoncx:
**Describe the bug**

Quickstart documentataion for Windows CLI using powershell is incorrect, the same command using bash is shown:
curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh | bash

while i believe a specific ps1 variant has been created by BestCodes for powershell

**To Reproduce**
1. Open powershell
2. Copy and run instrction

**Expected behavior**
expect proper installation to launch

**Screenshots**

<img width="1724" height="786" alt="Image" src="https://github.com/user-attachments/assets/3aa1c446-9706-4a39-b769-cfca0ecbbd3c" />

**Please provide following information:**
 - **OS & Arch:** [Windows 11 default powershell]
 - **Interface:** [CLI]
 - **Version:** [v1.7.0]
 - **Extensions enabled:** [n/a]
 - **Provider & Model:** [n/a]

**Additional context**

cbruyndoncx:
this is the error 

<img width="2156" height="358" alt="Image" src="https://github.com/user-attachments/assets/f5bfb0e3-0114-4293-83bd-92d60ea9864f" />

tiensi:
Did some local testing, even if we were to use the ps1 installation script `Invoke-WebRequest -Uri "https://github.com/block/goose/releases/download/stable/download_cli.ps1" -OutFile "download_cli.ps1";` 

I think it will still fail as there are no release artifacts for `download_cli.ps1` at that endpoint, I think we need to add that to the release yaml.

tiensi:
[Created a PR to possibly get the release artifact up](https://github.com/block/goose/pull/4478)

michaelneale:
if you use: 

```
Invoke-WebRequest -Uri "https://raw.githubusercontent.com/block/goose/main/download_cli.ps1" -OutFile "download_cli.ps1";
```

it should work?

bjornbyte:
> if you use:
> 
> ```
> Invoke-WebRequest -Uri "https://raw.githubusercontent.com/block/goose/main/download_cli.ps1" -OutFile "download_cli.ps1";
> ```
> 
> it should work?

It does work, it just doesn't mention that command line in the [installation instructions](https://block.github.io/goose/docs/getting-started/installation/) which makes it hard to discover.

Lymah123:
.take

cbruyndoncx:
@Lymah123 

[...]

PR #4478, opened 2025-09-02
fix(release): Added download_cli.ps1 to upload install script

tiensi:
## Pull Request Description

[As per this issue](https://github.com/block/goose/issues/4469) the powershell install guidance is likely incorrect and doesn't use the existing `download_cli.ps1`. When investigating, it seems that there is no `download_cli.ps1` artifact available through powershell `Invoke-WebRequest`. 

**Changes**
- Added download_cli.ps1 to `Upload Install Scripts` in build yml files. This should enable access to `download_cli.ps1` through powershell endpoint:

```Invoke-WebRequest -Uri "https://github.com/block/goose/releases/download/stable/download_cli.ps1" -OutFile "download_cli.ps1";```

**Notes**
- If this works properly I will create a follow up pr to update the readme.
- I may need help testing as I don't have permissions to trigger a build and verify the artifact.


The-Best-Codes (approved):
LGTM, not sure how to test this exactly though.

michaelneale:
FYI - we don't need it in releases, we can adjust scripts to download it from here: 

`https://raw.githubusercontent.com/block/goose/main/download_cli.ps1`

that way will always be the latest - can we adjust the quickstart docs to point to that link instead?

tiensi:
oh I didn't know that was possible. I'll close this and see if I can test it out. Thanks!

Issue #3747, opened 2025-07-31
Adding delete session option to Desktop UI


datnguyen1:
Please explain the motivation behind the feature request.
Currently, the Goose Desktop application does not provide an option to delete sessions directly from the UI. This feature is available in the CLI version (goose session remove -i <session_id>), but not having it in Desktop leads to inconsistency between the two interfaces. As a user, I often accumulate unused or outdated sessions, which can clutter the workspace and make it harder to manage active sessions. Having a delete option in the Desktop app would streamline session management and improve the overall user experience.

Describe the solution you'd like
I would like to see a "Delete Session" button or action in the Goose Desktop interface, ideally available alongside each session in the session list. This would allow users to quickly and intuitively remove sessions they no longer need, keeping their session list organized and relevant.

Describe alternatives you've considered
The only current alternative is to use the CLI to delete sessions, which is less convenient for users who prefer the Desktop UI or are not comfortable using the command line. Another workaround is to ignore old sessions, but this leads to unnecessary clutter and potential confusion.

Additional context

The CLI already has a session deletion feature, so adding it to the Desktop would create feature parity and a more consistent user experience.

A delete button is a standard and intuitive part of session or item management in most applications.

This would help users keep their workspace clean and focused on active sessions.

Screenshots or mockups can be provided if needed.

- [x] I have verified this does not duplicate an existing feature request

DOsinga:
see #2803

taniacryptid:
@Abhijay007 's https://github.com/block/goose/pull/4480 should close this!

PR #4480, opened 2025-09-02
feat: Add functionality to delete session in history list view

Abhijay007:
Closes: https://github.com/block/goose/issues/3747

### Pull Request Description

This PR introduces session deletion functionality with a confirmation modal in the UI. It replaces the previous IPC approach in the existing PRs  with a backend endpoint-based approach (as per Zane’s comment in this PR here: https://github.com/block/goose/pull/2803#issuecomment-2985625966)

**Changes Made**

- Added a `DELETE /sessions/{session_id}/delete` endpoint with secure validation
- Added a trash icon button in `SessionListView` with a confirmation modal for deletion
- Updated logic to use session list validation (same as CLI) instead of the IPC approach

**Preview**

deleteSession

### DIFF
--- a/ui/desktop/src/components/ui/ConfirmationModal.tsx
+++ b/ui/desktop/src/components/ui/ConfirmationModal.tsx
@@ -34,7 +34,7 @@ export function ConfirmationModal({
       <DialogContent className="sm:max-w-[425px]">
         <DialogHeader>
           <DialogTitle>{title}</DialogTitle>
-          <DialogDescription className="whitespace-pre-wrap">{message}</DialogDescription>
+          <DialogDescription>{message}</DialogDescription>
         </DialogHeader>
 
         <DialogFooter className="pt-2">


Abhijay007:
Hi @zanesq, I’ve resolved the requested changes. I removed `PathBuf` (a4a0954) for the `path source` and used the existing `session::get_path()` framework like in the other endpoints. This way we get the built-in framework validation, eliminate the path traversal risk, and maintain consistency across endpoints and the codebase.

I also updated the delete confirmation. However, could you please check again and let me know if it’s still happening? Maybe try deleting another chat to see if it reproduces. I’m not sure if it’s a title-related issue.

zanesq:
Still seeing the line break in the confirmation dialog. The dialog still shows for a bit enabled when clicking it also but I guess its not a deal breaker if not an easy fix.

image

Abhijay007:
> Still seeing the line break in the confirmation dialog. The dialog still shows for a bit enabled when clicking it also but I guess its not a deal breaker if not an easy fix.
> 
> image

Yeah, I am not sure why it's there or what the possible cause as I haven't used a line break here: https://github.com/block/goose/blob/c6672b24127b47a987a4ac4f445bf649e0199b17/ui/desktop/src/components/sessions/SessionListView.tsx#L678

Probably due to `<DialogDescription/>` in confirmationModal, let me look into that

Abhijay007:
Hi @zanesq , can you please try removing :

https://github.com/block/goose/blob/c6672b24127b47a987a4ac4f445bf649e0199b17/ui/desktop/src/components/ui/ConfirmationModal.tsx#L37

The `className="whitespace-pre-wrap"` from `<DialogDescription/>`  and see if the working, the thing is I tried that, but it was the same for me as it was (like there is no line break 😅), but I think it might solve this, so can you please try it once, and if that works, let me know. I will remove that

PR #4488, opened 2025-09-02
feat(rmcp-conversion): mcp server conversions

alexhancock:
Remove the mcp-server crate, and update all internally implemented mcp servers to use https://github.com/modelcontextprotocol/rust-sdk's way of doing things

Following up on @angelahning's excellent change for the developer server here https://github.com/block/goose/pull/4297. It's just more of the same!

Also removes `crates/mcp-server`


michaelneale:
I think want to chase down regressions that the other crate seems to have brought first

alexhancock:
Need to bring back the input schemas in Autovisualizer, and look for any other issues before merge

crates/mcp-server/Cargo.toml
```diff
@@ -1,23 +0,0 @@
-[package]
```
angelahning, line 1: delete delete delete 💯

crates/goose-mcp/src/computercontroller/mod.rs
```diff
@@ -1,426 +1,132 @@
-use base64::Engine;
 use etcetera::{choose_app_strategy, AppStrategy};
```
angelahning, line 1: computercontroller looks good

crates/goose-mcp/src/memory/mod.rs
```diff
@@ -1,129 +1,122 @@
-use async_trait::async_trait;
 use etcetera::{choose_app_strategy, AppStrategy};
```
angelahning, line 1: memory looks good

crates/goose-mcp/src/tutorial/mod.rs
```diff
@@ -1,57 +1,59 @@
 use anyhow::Result;
```
angelahning, line 1: 👍

crates/goose-mcp/src/autovisualiser/mod.rs
```diff
@@ -1,564 +1,324 @@
 use base64::{engine::general_purpose::STANDARD, Engine as _};
 use etcetera::{choose_app_strategy, AppStrategy};
-use indoc::{formatdoc, indoc};
+use indoc::formatdoc;
```
angelahning, line 3: Not sure if it is an regression.
Given this prompt
```I have quarterly sales data for different product categories. Can you help me understand:
1. The hierarchical breakdown of revenue across our nested product categories
2. How our performance metrics compare across all four quarters  
3. The customer flow through our sales funnel process

Here's the data:
- Electronics: Q1: $150k, Q2: $180k, Q3: $220k, Q4: $195k
- Clothing: Q1: $120k, Q2: $140k, Q3: $160k, Q4: $175k  
- Home & Garden: Q1: $80k, Q2: $95k, Q3: $110k, Q4: $125k```

Comment (angelahning) [reply crates/goose-mcp/src/autovisualiser/mod.rs:3]: Goose created 3 visualizations - a treemap, radar, and sankey diagram. But only one diagram is rendered on the UI. 
Screenshot 2025-09-08 at 4 35 10 PM

File: crates/goose-mcp/src/memory/mod.rs (hunk 3 above)
Comment (angelahning) [reply crates/goose-mcp/src/memory/mod.rs:1]: tesed on ui

File: crates/goose-mcp/src/computercontroller/mod.rs (hunk 2 above)
Comment (angelahning) [reply crates/goose-mcp/src/computercontroller/mod.rs:1]: Tested on UI:
- play classical music in safari, worked
- max screen brightness, didnt work on the released branch nor this

File: crates/goose-mcp/src/autovisualiser/mod.rs (hunk 5 above)
Comment (angelahning) [reply crates/goose-mcp/src/autovisualiser/mod.rs:3]: same behaviour on main. Good to go.

**Review by angelahning — APPROVED** [review]
Looks good. Works well on the UI (I didn't test on CLI).

File: crates/goose-mcp/src/autovisualiser/mod.rs (hunk 5 above)
Comment (michaelneale) [reply crates/goose-mcp/src/autovisualiser/mod.rs:3]: unrelated to this PR - but worth looking for any console errors in electron client, as the html can likely be tweaked

File: crates/goose-mcp/src/autovisualiser/mod.rs [hunk 6]
```diff
@@ -1,564 +1,324 @@
...
-            indoc! {r#"
-                show interactive line, scatter, or bar charts
-                
-                Required: type ('line', 'scatter', or 'bar'), datasets array
-                Optional: labels, title, subtitle, xAxisLabel, yAxisLabel, options
-                
-                Example:
-                {
-                  "type": "line",
-                  "title": "Monthly Sales",
```
michaelneale, line 484: looks like a lot of this doc/prompt was lost, it is hard to tell as everything is on one line now, but I can't find this anywhere, which will likely stop it working.

crates/goose-mcp/src/autovisualiser/mod.rs
```diff
@@ -1,564 +1,324 @@
...
-    fn create_sankey_tool() -> Tool {
-        Tool::new(
-            "render_sankey",
-            indoc! {r#"
-                show a Sankey diagram from flow data               
-                The data must contain:
-                - nodes: Array of objects with 'name' and optional 'category' properties
-                - links: Array of objects with 'source', 'target', and 'value' properties
-                
-                Example:
```
michaelneale, line 82: this is missing from the new tool `description`

crates/goose-mcp/src/autovisualiser/mod.rs
```diff
@@ -1,564 +1,324 @@
...
-                    {"name": "Target B", "category": "target"}
-                  ],
-                  "links": [
-                    {"source": "Source A", "target": "Target B", "value": 100}
-                  ]
-                }
-            "#},
-            object!({
-                "type": "object",
-                "required": ["data"],
```
michaelneale, line 95: I assume this is now done by the macro?

crates/goose-mcp/src/autovisualiser/mod.rs
```diff
@@ -1,564 +1,324 @@
...
-        Tool::new(
-            "render_radar",
-            indoc! {r#"
-                show a radar chart (spider chart) for multi-dimensional data comparison             
-                
-                The data must contain:
-                - labels: Array of strings representing the dimensions/axes
-                - datasets: Array of dataset objects with 'label' and 'data' properties
-                
-                Example:
```
michaelneale, line 141: this is also missing

crates/goose-mcp/src/computercontroller/mod.rs
```diff
@@ -1,426 +1,132 @@
...
+fn default_save_as() -> String {
+    "text".to_string()
 }
 
-impl ComputerControllerRouter {
-    pub fn new() -> Self {
-        let web_scrape_tool = Tool::new(
-            "web_scrape",
-            indoc! {r#"
-                Fetch and save content from a web page. The content can be saved as:
```
michaelneale, line 57: missing.

michaelneale (changes requested):
Looks like a lot of tool descriptions are truncated, which will hurt performance (I wonder if we saw similar thing with developer one, but seems ok). 

did some inline comments. 

I think the tool macro like: 

```rust
    #[tool(
        name = "web_scrape",
        description = "Fetch and save content from a web page. The content can be saved as text, json, or binary files."
    )]
```

Can be more like: 

```rust
    #[tool(
        name = "web_scrape",
        description = r#"Fetch and save content from a web page. 
The content can be saved as text, json, or binary files.

and the rest here...

including examples
"#
    )]
```

this would make it perform like original, and also probably make the diffs managable, at the moment github hides it away as so much has changed (likely these)

michaelneale:
can we ensure we have all the tool descriptions intact like before, with examples in the tool macros (which I think can be multiline?)

crates/goose-mcp/src/autovisualiser/mod.rs
michaelneale: (and the prompts have changed enough that there could be some effect there)

crates/goose-mcp/src/autovisualiser/mod.rs
alexhancock: yes. I tried to bake this into the schema of the input args itself. will validate in the inspector it makes it into the `inputSchema` before merge

michaelneale:
I think once the prompts are back in place, probably good to go. I saw some changes to the developer mcp - which seemed more rust style related - deliberate?

alexhancock:
Need to find a way to bring back all the os specific tool descriptions while also using rmcp instead of the old mcp-server construct which modeled things as a `vec` of `Tools`

The macros in rmcp, while nice, make it hard to provide os specific descriptions.

Will find a good pattern for this.

michaelneale:
@alexhancock yeah fair point - as sometimes we want to programmatically form the set of tools available, either at load time or even dynamically

alexhancock:
Breaking this apart into separate changes that are easier to verify are totally correct

Here is the first one https://github.com/block/goose/pull/4664

PR #4491, opened 2025-09-03
fix: need to send errors to appropriate stream

michaelneale:
before this the dev MCP server would exit whene running in CLI. 

it will now look like this: 

image

which stops it from terminating (still not ideal though


michaelneale:
"mcp error" i think comes from the rmcp server code

alexhancock (approved):

jamadeo:
I'm not sure how this didn't trigger the same issue in the existing MCP servers as well. Maybe they just don't have any calls to tracing::warn?

PR #4492, opened 2025-09-03
fixed css classes and added some accessibility fixes, incl draft style guides 

cbruyndoncx:
## Pull Request Description
fixed css classes and added some accessibility fixes, incl draft style guide and directory structure for DESKTOP UI

### DIFF
--- a/DESKTOP_UI_directory_structure.md
+++ b/DESKTOP_UI_directory_structure.md
@@ -1,0 +1,47 @@
+# Directory structure of ui/desktop/src
+
+- components
+  - settings
+    - models
+      - modelInterface.ts
+      - predefinedModelsUtils.ts
+      - ModelsSection.tsx
+      - bottom_bar
+        - ModelsBottomBar.tsx
+      - subcomponents
+        - ModelSettingsButtons.tsx
+        - AddModelModal.tsx
+        - LeadWorkerSettings.tsx
+  - extensions
+  - alerts
+  - parameter
+  - Layout
+  - ui
+  - schedule
+  - GooseSidebar
+  - icons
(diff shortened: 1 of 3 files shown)


DESKTOP_UI_directory_structure.md
```diff
@@ -0,0 +1,47 @@
+# Directory structure of ui/desktop/src
```
zanesq: Appreciate the intent but these doc files tend to get out of date so we avoid them. Also there are a few things that are incorrect in the style guide. Can you remove the md files?

cbruyndoncx: Done, removed the markdown files

zanesq: Thanks, I agree we need to have a styleguide and truly appreciate it. Happy to take a look if you want to create a new PR with just the styleguide.md and keep it concise and relevant. The styleguide in this PR was way too broad / AI generated also was giving incorrect advice on using components like html button when we already have a `Button` component.

cbruyndoncx: > Thanks, I agree we need to have a styleguide and truly appreciate it. Happy to take a look if you want to create a new PR with just the styleguide.md and keep it concise and relevant. The styleguide in this PR was way too broad / AI generated also was giving incorrect advice on using components like html button when we already have a `Button` component.

I can certainly add a PR with a draft, but i think it will need input from other people to make it good. 
I have used goose with openai gpt-5-mini for the code changes. So I wanted to use it to guide the LLM, but i think your concern is more for the humans. Anyhow, I will add a draft PR and see where we land ...

PR #4493, opened 2025-09-03
feat(blog) state of the install: new ux with Mac+Ollama+Goose

ewann:
## Pull Request Description

I'm not sure if you even consider blog post proposals from community, but given the potholes I managed to fall into getting this working, I decided to write up the "working" bit. One of my comments in this PR is maybe it should be an update to docs instead of a blog post.  I landed here, but am open to adapting it to a docs update if the team wants to accept any of it, but feels that is the better target.

Proposed community blog entry based on success overcoming challenges consuming the existing docs for the combination of Mac+ollama+goose for tool calling.

Originating discussion:
https://discord.com/channels/1287729918100246654/1410510729995616276/1410510733388677253


taniacryptid:
Thank you for tagging the devrel team! Off the bat I wonder if this could be a singular guide on getting Ollama to work vs a blog post. Wonder what you think @block/goose-devrel .

Like how you share the Ollama config params you need to set, the tool shim being key to getting tool calling to actually work. 

I'll wait for my teammates share their perspective and insight here, but that was my first immediate impression. I think the model selection is interesting too.

documentation/blog/2025-09-03-start-the-fans-please/index.md
```diff
@@ -0,0 +1,387 @@
...
+sudo sysctl iogpu.wired_limit_mb=<you need to work out an appropriate number>
+```
+
+__NOTE:__ Local system stability is your problem now :wink:
+
+__NOTE:__ After changing this setting, it is necessary to restart Ollama for it to react to the change.
+
+The above command automatically resets during system boot. You want to persist some crazy value that ends up __preventing__ boot? That is left as an exercise for the reader. :wink:
+
+##
```
blackgirlbytes, line 55:
```suggestion
```

iandouglas:
Per Discord conversations I've had with @ewann we're going to shape this into a documentation guide:

- step by step guide, just installation instructions, links to setup sources, etc, no commentary, no emojis, etc.
- we'll add it under `documentation/docs/guides/ollama/index.md`
- let's start with this as a one-page guide for links to ollama, the base models you tested on Mac hardware, etc
- we can definitely talk about setting up `GOOSE_TOOLSHIM_OLLAMA_MODEL` 
- if the page feels too lengthy, we can split it into a page per platform for Mac/Linux/Windows

PR #4498, opened 2025-09-03
feat: simplify navigation, make reload work

jamadeo:
Improve the handling of routing by mostly replacing window.location.* with the corresponding react-router hooks, and keep better track of what the URL should be. With this change, reloading (cmd/ctrl+R) should now mostly do the expected thing.

### DIFF
--- a/ui/desktop/package.json
+++ b/ui/desktop/package.json
@@ -39,7 +39,6 @@
     "start-alpha-gui": "ALPHA=true npm run start-gui"
   },
   "dependencies": {
-    "@tanstack/react-form": "^0.13.0",
     "@ai-sdk/openai": "^2.0.14",
     "@ai-sdk/ui-utils": "^1.2.11",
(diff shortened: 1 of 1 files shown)


zanesq:
Nice I think this might work better, can you open this so we can try in a `.bundle` app?

jamadeo:
.bundle

jamadeo:
.bundle

zanesq:
Got this from the bundle on app launch do you see this? Could pull in main to get rid of the localstorage error but I suspect the real error is 

`Not allowed to load local resource: file:///Users/zane/Downloads/Goose%2059.app/Contents/Resources/app.asar/.vite/renderer/main_window/index.html/%23`
 
image

jamadeo:
oh yeah that's definitely this change, the %23 is a url encoded hash, I'll fix that

jamadeo:
.bundle

jamadeo:
.bundle

PR #4504, opened 2025-09-03
improve auto scroll to bottom + detection

zanesq:
## Pull Request Description
During chat the window is not auto scrolling to the bottom after an agent response. This adds detection if the user is near the bottom of scroll and debouncing with timeouts to avoid jumpiness.

Also fixes the window not auto scrolling to the bottom when resuming sessions

### DIFF
--- a/ui/desktop/src/components/BaseChat.tsx
+++ b/ui/desktop/src/components/BaseChat.tsx
@@ -276,18 +276,11 @@ function BaseChatContent({
   }, []);
 
   // Auto-scroll when messages are loaded (for session resuming)
-  useEffect(() => {
-    if (messages.length > 0 && !loadingChat) {
-      const scrollTimeout = setTimeout(() => {
-        if (scrollRef.current?.scrollToBottom) {
-          scrollRef.current.scrollToBottom();
-        }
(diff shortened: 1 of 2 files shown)


zanesq:
.bundle

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -229,7 +273,21 @@ function BaseChatContent({
...
+  }, []);
+
+  // Auto-scroll when messages are loaded (for session resuming)
+  useEffect(() => {
+    if (messages.length > 0 && !loadingChat) {
+      const scrollTimeout = setTimeout(() => {
+        if (scrollRef.current?.scrollToBottom) {
+          scrollRef.current.scrollToBottom();
+        }
+      }, 500); // delay to ensure content is fully loaded
```
jamadeo: does this actually have to do with loading? since we already checked that in !loadingChat

zanesq: good question it was to add a delay for dom rendering but I went ahead and refactored to use a callback from progressive message list instead and its much cleaner and more reliable 👍

zanesq:
.bundle

PR #4508, opened 2025-09-04
docs: malware check for uvx and npx extensions

dianed-square:
This PR documents the automatic malware check on external extensions:

Documentation updates:
- `documentation/docs/getting-started/using-extensions.md`:  Added note and dedicated section


documentation/docs/getting-started/using-extensions.md
```diff
@@ -16,6 +16,9 @@ Goose to a wide ecosystem of capabilities.
 Check out the [step-by-step tutorials](/docs/category/mcp-servers) for adding and using several Goose Extensions
 :::
 
+:::info Malware Scan
+Goose automatically checks external extensions (`uvx` and `npm`) for known malware before activation. If a malicious package is detected, the [extension will be blocked](#extension-blocked-for-security-reasons) with a clear error message.
```
angiejones:
```suggestion
Goose automatically checks external extensions (`uvx` and `npx`) for known malware before activation. If a malicious package is detected, the [extension will be blocked](#extension-blocked-for-security-reasons) with a clear error message.
```

should it be npx? what about jbang - is that not checked?

documentation/docs/getting-started/using-extensions.md
```diff
@@ -631,6 +634,22 @@ goose session --with-extension "GITHUB_PERSONAL_ACCESS_TOKEN=<YOUR_TOKEN> npx -y
 Note that you'll need [Node.js](https://nodejs.org/) installed on your system to run this command, as it uses `npx`.
 :::
 
+#### Extension blocked for security reasons
```
angiejones: this seems like it would go better in the Troubleshooting Guide with a link to it from the paragraph above

documentation/docs/getting-started/using-extensions.md
dianed-square: No jbang: https://github.com/block/goose/blob/main/crates/goose/src/agents/extension_malware_check.rs#L49

Issue #4509, opened 2025-09-04
gpt-5 input context limit is incorrect


capttrousers:
https://github.com/block/goose/blob/bce046607ce997ff64bbd844a87e8bb4d2983a8c/crates/goose/src/model.rs#L20

encoded as 272_000 input tokens. 

Docs say 400_000 input tokens: https://platform.openai.com/docs/models/gpt-5

PR #4510, opened 2025-09-04
fix: gpt-5 input context limit

capttrousers:
Fixes: https://github.com/block/goose/issues/4509


michaelneale:
thanks @capttrousers you will need to follow the DCO instructions for the failure

michaelneale (approved):

michaelneale:
@capttrousers if you can do the DCO and sign things and then re-open, we can merge this.

DOsinga:
In fact this has already landed through a different path. thanks for contributing though @capttrousers !

PR #4511, opened 2025-09-04
feat: Agent Client Protocol implementation of goose

michaelneale:
First cut of an an implementation of https://agentclientprotocol.com/ for goose (still technically WIP - looking at how to make it a bit richer, but does work).

This allows it to work with any agent client that is compatible (as demonstrated below, Zed, the super slick text editor/IDE). 

This is currently implemented as `goose acp` and can be used in zed out of the box (Zed being the original ACP client): 

image

https://github.com/user-attachments/assets/cc4109f4-89c1-440e-989d-cd8d3c4b59df

to use, in zed settings.json

```json
{
  "agent_servers": {
    "Goose 🪿": {
      "command": "goose",
      "args": ["acp"],
      "env": {}
    }
  },
  "ui_font_size": 16,
  "buffer_font_size": 16
}
```

then create a new thread: 

image

and away you go.

There is nothing editor or zed specific here, this can also power other clients, potentially the desktop, or TUIs if it works well (but for now, Zed is awesome!)


alexhancock (approved):
I love it.

Will want to look further if we can also incorporate it into `goose-server` and then client bits in the default cli command, and desktop app of goose, but a dedicated command seems the right place to start experimenting!

I say merge it

11111000000:
Any roadmap about A2A?  (ACP merged to A2A)

aarononeal:
Is there a way to use recipes and sub-recipes in ACP mode?

DOsinga:
@aarononeal how would you imagine we would hook this up?

aarononeal:
From a UX perspective similar to how `run` works. So `goose acp --recipe <recipe>` to enable me to create multiple Goose agents in Zed tailored to specific environments and roles.

DOsinga:
oh, I see, basically just fine tune the system prompt. I like that idea. send a PR!

PR #4522, opened 2025-09-04
Add unified diff support to text editor

tlongwell-block:
Adds support for applying unified diffs (both single-file and multi-file) through the text editor's `str_replace` command. This reduces token usage when making multi-line edits by up to 90%.

## Why

Previously, multi-line edits required sending the entire old and new content as parameters, consuming substantial tokens. With diff support, only the changes are transmitted, reducing token usage for large edits.

### Uses `patcher` and `path_trav` crates

- **New parameter**: Added optional `diff` parameter to `TextEditorParams`
- **Parsing**: Uses `patcher` crate (v0.2.1) for diff parsing
- **Path safety**: Uses `path_trav` crate for traversal detection
- **Single-file diffs**: Applied using `patcher::Patch` with validation
- **Multi-file diffs**: Uses `patcher::MultifilePatch` for multi-file operations
- **Safety**: Maintains security validations:
  - Path traversal prevention via `path_trav`
  - Symlink detection  
  - Size limits (1MB max diff size, 100 files max)
  - Atomic operations with rollback on failure

### Architecture

The implementation follows a functional approach:
- **`patcher` crate**: Handles diff parsing and patch application
- **`path_trav` crate**: Validates path safety
- **Modular functions**: Small, focused functions for each responsibility
- **Clear data flow**: Uses typed structs for passing data between functions

### Key components

- `apply_diff()`: Coordinates the diff application process
- Path validation: `validate_path_safety()` uses `path_trav` for security
- Helper functions: Separate functions for parsing, validation, and application

## Benefits

- Token reduction for multi-line edits
- Atomic operations for multi-file changes
- Compatible with git-style unified diffs

## Testing

Test suite includes single and multi-file diff application, file operations, context validation, undo functionality, and security checks. All tests pass.

### DIFF
--- a/crates/goose-mcp/src/developer/text_editor.rs
+++ b/crates/goose-mcp/src/developer/text_editor.rs
@@ -1,15 +1,14 @@
 use anyhow::Result;
-use diffy::{apply, Patch};
 use indoc::formatdoc;
-use once_cell::sync::Lazy;
-use regex::Regex;
+use patcher::{
+    ApplyResult, MultifilePatch, MultifilePatcher, Patch as PatcherPatch, PatchAlgorithm, Patcher,
+};
 use std::{
     collections::HashMap,
     fs::File,
     io::Read,
     path::{Path, PathBuf},
 };
-use tempfile::NamedTempFile;
 use url::Url;
 
 use rmcp::model::{Content, ErrorCode, ErrorData, Role};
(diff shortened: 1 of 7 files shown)


michaelneale:
seems nice - not sure if we have this AND replace_all (maybe....). curious how performance is

crates/goose-mcp/src/developer/text_editor.rs
```diff
@@ -15,6 +17,358 @@ use super::shell::normalize_line_endings;
...
+            results.files_created,
+            results.files_modified,
+            results.files_deleted,
+            results.lines_added,
+            results.lines_removed
+        )
+    };
+
+    let user_message = if is_single_file {
+        format!("{}\n\nUse 'undo_edit' to revert if needed.", summary)
```
katzdave: Add a \n at the end of these messages (maybe 2?)

Seeing printout on cli:

path: ~/g/g/g/c/g/s/developer/text_editor.rs
command: str_replace
diff: [REDACTED: 766 chars]

Successfully applied diff to /Users/dkatz/git/goose2/goose/crates/goose-mcp/src/developer/text_editor.rs:
• Lines added: 2
• Lines removed: 2

Use 'undo_edit' to revert if needed.Great! I successfully

tlongwell-block: Done 👍  Great catch.

michaelneale (approved):
I think this is ok, I wasn't able to trigger it to use a diff, it would happily use str_replace when it made sense, but that is probably ok.  I did try to tell it to make a diff, it did, but it didn't work (probably as it was contrived). Any tips on how to see this out? just has to be day to day?

Seems a nicely focussed change and improvement. Main concern is how it works with lower end models, but the prompt doesn't really change much so should be ok.

tlongwell-block:
@michaelneale yeah, it is reluctant to use it. The old/new str is more intuitive for it, I think.

But, in my testing, sonnet handled the diff format for multiple files (going up to 10) really well. I'm not sure if I should use stronger language here to encourage using diff editing more, because it is pretty aggressive already.

michaelneale:
@tlongwell-block I would leave it as is, and consider (while cracked open) adding in replace_all in this one omnibus PR and then we can get it in? covers all bases then.

katzdave (approved):

understood-the-assignment:
Claude 3.7 Sonnet seems to be having a lot of trouble with this new tool for me.

It tries to:

```
path: ~/a/b/c/d/e/f/g/s/i/transmogrification/transmogrify.py
command: str_replace
diff: --- a/src/initech/transmogrification/transmogrify.py
+++ b/src/initech/transmogrification/transmogrify.py
@@ -127,7 +127,7 @@ ....
 ...patch follows ...

-32603: Mcp error: -32603: File '/home/user/a/b/c/d/e/f/g/src/initech/transmogrification/src/initech/transmogrification/transmogrify.py' not found and patch doesn't create it
```

i.e. it puts the full path in the `path` parameter and then a relative path in the `diff` header.

IMHO Goose should stop trying to invent clever new variations of tools that are completely different from what the common models out there are trained on. We should try to mimic the prompts and tools that Claude Code, Gemini CLI, and OpenAI Codex are providing to their respective models.

PR #4523, opened 2025-09-04
update to RMCP 0.6.2

Kvadratni:
## Pull Request Description
Updates RMCP to latest version

### DIFF
--- a/ui/desktop/openapi.json
+++ b/ui/desktop/openapi.json
@@ -1559,12 +1559,12 @@
               "$ref": "#/components/schemas/Role"
             }
           },
-          "priority": {
-            "type": "number"
-          },
-          "timestamp": {
+          "lastModified": {
             "type": "string",
             "format": "date-time"
+          },
+          "priority": {
+            "type": "number"
           }
         }
       },
@@ -1681,6 +1681,13 @@
             "allOf": [
               {
                 "$ref": "#/components/schemas/Annotated"
+              }
+            ]
+          },
+          {
+            "allOf": [
+              {
+                "$ref": "#/components/schemas/RawResource"
               }
             ]
           }
@@ -1893,6 +1900,10 @@
           "resource"
         ],
         "properties": {
+          "_meta": {
+            "type": "object",
+            "additionalProperties": true
+          },
           "annotations": {
             "anyOf": [
               {
@@ -2390,6 +2401,10 @@
           "mimeType"
         ],
         "properties": {
+          "_meta": {
+            "type": "object",
+            "additionalProperties": true
+          },
           "annotations": {
             "anyOf": [
               {
@@ -2877,6 +2892,10 @@
           "resource"
         ],
         "properties": {
+          "_meta": {
+            "type": "object",
+            "additionalProperties": true
+          },
           "resource": {
             "$ref": "#/components/schemas/ResourceContents"
           }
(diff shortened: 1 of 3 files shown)


crates/goose/src/conversation/message.rs
```diff
@@ -317,6 +324,7 @@ impl From<Content> for MessageContent {
             RawContent::Image(image) => {
                 MessageContent::Image(image.optional_annotate(content.annotations))
             }
+            RawContent::ResourceLink(_link) => MessageContent::text("[Resource link]"),
```
jamadeo, line 327: we definitely need to do more here, but it's not clear to me exactly what that should be. Maybe just a JSON dump with the name, URI and description? then if the model decides to read it, we let it call read_resource?

Kvadratni: Good question. This implementation was literally just a cargo clippy satisfying minimal change. We might want to merge and decide later

aharvard: So, I added `resource_link` support to the RCMP, but am unsure if I did it right. Or maybe the problem is what you are asking about @jamadeo?

I made this dummy tool 
```ts
server.tool(
        "resource-link-demo",
        "A demo of a resource link that displays a fake demo text file.",
        {},
        async (): Promise<CallToolResult> => {
            return {
                content: [
                    {
                        type: "resource_link",
                        uri: "file:///project/src/fake-demo.txt",
                        name: "fake-demo.txt",
                        description:
                            "Fake demo text file to test for resource link data.",
                        mimeType: "text/plain",
                        annotations: {
                            audience: ["assistant"],
                            priority: 0.9,
                        },
                    },
                ],
            };
        }
    );
```

When I ran Goose from this PR branch, it looks like my tool is called but nothing is returned in the tool call output:
image


I would expect all data that conforms to the [resource link](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#resource-links) spec to show in the output, like the MCP inspector is showing here:

image

aharvard:
Nice! Happy to report `_meta` is now flowing thru as expected. 
image

crates/goose/src/conversation/message.rs
jamadeo: @aharvard what I mean is, now that the support is there in rmcp (thank you!) it's up to Goose to do the right thing. In this PR it turns all resource links into some text "[Resource link]" which is surely not what we want.

It wasn't totally clear to me what Goose should do with it though. I *think* we want to show the link to the model, allowing the model to then call read_resource with the URL if it wants to? Should it also render something in the UI?

If the answer is that we're not sure yet, I'd say lets start by dumping the json, that way the model at least sees the whole result @Kvadratni

alexhancock: Yeah this is an example of an interesting "what should the host do?" question when working with MCP.

I think something which provides the full json (including URI of the resource) with `"audience": ["assistant"]` in the message should be good?

Optionally we could render a little bubble inside the tool result similar to what inspector does showing the URI, but less critical than making sure the model sees it.

In any case, I think we should merge the package upgrade to get that in, then do this in a separate focused change. Two smaller changes not coupled to eachother is better.

aharvard:
.bundle

crates/goose/src/conversation/message.rs
aharvard: @alexhancock & @jamadeo, I'm chatting with @kentcdodds about his POV for `resource_link` — he's got a really awesome MCP workshop that leverages the technique. Maybe we could start a new issue to discuss the topic?

alexhancock: Absolutely!

@kentcdodds if you have thoughts on how a host can make the best experience for resource links (or other data types we don't represent yet) we're all ears to incorporate emerging patterns.

kentcdodds: Sorry, I haven't used Goose with resources at all yet, but is there somewhere in Goose that shows a list of the available resources that a user can add to Context? That's what I think should happen: when a resource_link shows up, that resource should appear in the list of resources that the user can add to Context, so basically when a tool responds with a resource_link, give the user the option of whether they want to include that in Context.

kentcdodds:
This is great! It successfully passes renderData through for me which allows me to send private data through to the UI for display 🎉

And `resource_link` not breaking stuff is a nice touch too :)

image

Thank you for this!

PR #4530, opened 2025-09-05
Add `analyze` tool to Developer extension

tlongwell-block:
Adds an `analyze` tool to the developer MCP extension that provides code structure analysis and symbol tracking capabilities across multiple programming languages.

## Changes
- Add tree-sitter based code analysis for Python, Rust, JavaScript/TypeScript, Go, Swift, and Java
- Support three analysis modes:
  - **Structure**: Quick overview of files, functions, and classes
  - **Semantic**: Detailed analysis including imports and call relationships  
  - **Focused**: Track specific symbols and their call chains across files
- Include LRU caching to improve performance for repeated analyses
- Respect `.gooseignore` and `.gitignore` patterns for file traversal

## Usage
```bash
# Analyze a file or directory structure
goose> analyze path/to/code

# Track a specific function and its relationships
goose> analyze path/to/project --focus main --follow-depth 2

# Limit directory traversal depth
goose> analyze large/project --max-depth 2
```

## Implementation Notes
- Modular architecture with separate modules for parsing, caching, traversal, and formatting
- Comprehensive test coverage (44 tests)

## Testing
Tested with various codebases including Python packages, Rust crates, and mixed-language projects. The tool correctly identifies code structure and tracks cross-file dependencies.


michaelneale:
I think this approach has promise, have closed https://github.com/block/goose/pull/3389 in favour of this one in the developer MCP which is the right place. Will want to make sure can deal with a tree compute that takes a few minutes without it timing out (so the tool can reply and run in background possibly).

crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -205,9 +211,15 @@ impl ServerHandler for DeveloperServer {
...
             Use the shell tool as needed to locate files or interact with the project.
 
+            Use the `analyze` tool aggressively to explore directory structures and understand codebases.
+            Leverage `analyze` through `return_last_only=true` subagents for deep codebase understanding with lean context
+            - delegate analysis, retain summaries
+
             Your windows/screen tools can be used for visual debugging. You should not use these tools unless
             prompted to, but you can mention they are available if they are relevant.
 
+            Always prefer ripgrep (rg -C 3) to grep.
```
michaelneale, line 224: is this best here or closer to where rg is mentioned elsewhere in the prompt (or was that system prompt?)

crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -187,6 +189,10 @@ impl ServerHandler for DeveloperServer {
 
                 Use the shell tool as needed to locate files or interact with the project.
 
+                Use the `analyze` tool aggressively to explore directory structures and understand codebases.
+                Leverage `analyze` through `return_last_only=true` subagents for deep codebase understanding with lean context
```
michaelneale, line 197: should it just have this second line, more compressed, and it will know from the tool desc how to use it? (just thinking in terms of token efficiency here)

michaelneale:
this seems to work nicely when I try it, a very nice addition. Any measuable improvement on benches @tlongwell-block ?

crates/goose-mcp/src/developer/analyze/cache.rs
```diff
@@ -0,0 +1,86 @@
...
+use std::path::PathBuf;
+use std::sync::{Arc, Mutex};
+use std::time::SystemTime;
+
+use super::lock_or_recover;
+use crate::developer::analyze::types::AnalysisResult;
+
+#[derive(Clone)]
+pub struct AnalysisCache {
+    cache: Arc<Mutex<LruCache<CacheKey, Arc<AnalysisResult>>>>,
```
michaelneale, line 12: probably will want to test this on a gnarly mono repo @tlongwell-block - to check it isn't absurdly large

tlongwell-block: Yeah, the cache has a `max_size` baked in that we can tune.

I tested it with a really big monorepo and the tool does well. Warnings when there is large output with details and explanation for Goose with alternative, as well as the option to pass a `force` parameter to see large outputs anyway

tlongwell-block:
> this seems to work nicely when I try it, a very nice addition. Any measuable improvement on benches @tlongwell-block ?

Yes, a fairly consistent 3-6% boost with this change. It gets Sonnet up to our current Opus average.

crates/goose-mcp/src/developer/rmcp_developer.rs
tlongwell-block: It is already in the developer extension prompt in the editor description for some reason. Not sure why there in particular, but it seems like general developer instructions might fit better here in the general developer mcp description

michaelneale: oh I mean in the common shell instructions. it says to use ripgrep - but it doesn't mention -C N at all, which seems to be the thing that hurts it (lacking that), but probably doesn't hurt here.

michaelneale (approved):
I think this is ok, makes goose smarter with code specifically, helps performance (I also like that it brings sonnet up with opus). Doesn't seem a huge overhead either.

but would love a tick from @jamadeo or @DOsinga to be sure before we ship it - as brings in a few libs, tree sitter etc.

tlongwell-block:
The tool itself is feature complete. The instructions/prompt need a little tweaking around helping the llm understand the max depth parameter and how it effects analysis

michaelneale:
@tlongwell-block I would think a response first/last line (probably both) saying "you will only see X, there is still Y" if possible in the response?

tlongwell-block:
> @tlongwell-block I would think a response first/last line (probably both) saying "you will only see X, there is still Y" if possible in the response?

The tool intentionally does lazy parsing of the codebase (because monorepos are a thing), so we can't really tell it how deep the call graph *actually* is without the agent electing to parse the entire repo to max_depth=0 (which it totally can do)

It can also surgically search for the calls in subdirectories or individual files to avoid having to parse everything. This works well, too

crates/goose-mcp/src/developer/rmcp_developer.rs
tlongwell-block: Done

PR #4532, opened 2025-09-05
feat: search code for performance

michaelneale:
guided use of rg vs leaving it up to LLM, this also refactors out the tests.rs to make things more maintainable.

This search can run in parallel, and will return content not just files to save on tool call roundtrips


michaelneale:
@katzdave @tlongwell-block I wonder if this is worth it, or a small tweak to rg prompt to tell it to use `-C 3` works just as well (minus the parallelism) as this is really just shelling out to rg anyway, may be better as part of https://github.com/block/goose/pull/4530 IMO)

michaelneale:
closing this as will be done elsewhere

PR #4534, opened 2025-09-05
Iand/ollama guide

ewann:
## Pull Request Description

In collaboration with @iandouglas, provide a more prescriptive and complete set of instructions to get working tool calling deployed on MacOS using Ollama hosted model(s) 

---


ewann:
Discussion at https://discord.com/channels/1287729918100246654/1412554148334932051/1413831264569659494 highlights an important point:

How best to capture community recommendations on models? In that thread, my opinion re qwen3-coder:32b has flip-flopped at least once now. In the current format, that's going to be very difficult to handle, with the risk the comments become untrustworthy.

I'm currently attempting to run goose-bench locally, so I can at least provide some more standardised results. Goose bench appears to have a number of expectations re what it can install, where and how, which my [nix](https://nixos.org/download/) config prevents it from doing, but I'll work through those / run again in a more vanilla macOS vm

iandouglas:
closing for now while we revisit how we want to accomplish this

PR #4538, opened 2025-09-05
Add Message Metadata for Visibility Control

katzdave:
## Summary

This PR introduces a new message metadata system to centralize visibility control on the server-side and significantly simplifies the client-side message handling by removing redundant visibility flags and ancestor message tracking.

### 🏗️ **Core Architecture Changes**

**Server-Side (Rust)**
- **New `MessageMetadata` struct** with `user_visible` and `agent_visible` flags
- **Enhanced `Message` struct** with metadata field and visibility methods:
  - `user_only()` - visible only to user, not agent
  - `agent_only()` - visible only to agent, not user
  - `with_visibility(user_visible, agent_visible)` - custom visibility
  - `is_user_visible()` / `is_agent_visible()` - visibility checks
- **Updated message constructors** across providers to use new metadata system
- **Improved context management** in `auto_compact.rs` to respect visibility flags

**Client-Side (TypeScript)**
- **Removed `display` and `sendToLLM` fields** from `Message` interface
- **Eliminated ancestor messages concept** - redundant with server-side metadata
- **Simplified `convertApiMessageToFrontendMessage`** - no longer sets visibility flags
- **Updated message streaming** to trust server-side filtering as single source of truth
- **Removed client-side message filtering** in `useMessageStream`

### 🔧 **Technical Improvements**

1. **Single Source of Truth**: Server now controls all message visibility logic
2. **Simplified Client Logic**: Removed complex client-side filtering and ancestor tracking
3. **Better Type Safety**: Cleaner TypeScript interfaces without optional visibility flags
4. **Consistent API**: All providers use the same metadata system

### 🧪 **Test Updates**

- Updated all test files to remove references to `display`/`sendToLLM` fields
- Added comprehensive tests for new metadata functionality
- Fixed context management tests to work with new visibility system

### 📁 **Files Changed**

**Rust Core (`crates/goose/`)**
- `conversation/message.rs` - New metadata system and visibility methods
- `context_mgmt/auto_compact.rs` - Respect message visibility in compaction
- `providers/` - Updated all providers to use new message constructors
- `agents/reply_parts.rs` - Use new message creation patterns

**Server (`crates/goose-server/`)**
- `routes/` - Updated API endpoints to handle metadata properly

**Client (`ui/desktop/`)**
- `types/message.ts` - Removed visibility fields from interface
- `hooks/` - Simplified message handling logic
- `components/` - Removed ancestor message tracking
- `__tests__/` - Updated all tests for new message structure

## Migration Impact

- **Backward Compatible**: Old messages without metadata default to fully visible
- **API Changes**: Client no longer needs to handle visibility logic
- **Performance**: Reduced client-side complexity and memory usage

## Benefits

1. **Centralized Control**: All visibility logic lives on the server
2. **Reduced Complexity**: Simpler client code with fewer edge cases
3. **Better Maintainability**: Single place to modify visibility behavior
4. **Type Safety**: Cleaner interfaces without optional visibility flags
5. **Performance**: Less client-side processing and memory usage

## Testing

- ✅ All existing tests updated and passing
- ✅ New metadata functionality thoroughly tested
- ✅ Message serialization/deserialization verified
- ✅ Context management respects new visibility system

This change sets the foundation for more sophisticated message visibility control while significantly simplifying the client-side codebase.


crates/goose/src/agents/context.rs
```diff
@@ -64,48 +64,62 @@ impl Agent {
...
+        let summary_tokens = summarization_usage
+            .as_ref()
+            .and_then(|usage| usage.usage.output_tokens)
+            .unwrap_or(0) as usize;
+        final_messages.push(summary_msg);
+        final_token_counts.push(summary_tokens);
+
+        // Add an assistant message to continue the conversation (agent_visible=true, user_visible=false)
+        let assistant_message = Message::assistant()
+            .with_text("The previous message contains a summary that was prepared because a context limit was reached.")
```
alexhancock: The 

```
Do not mention that you read a summary or that conversation summarization occurred
Just continue the conversation naturally based on the summarized context
```

seemed helpful to me before in getting it not to make a big deal of the fact that it just read a summary. sometimes it still mentions it, but not nearly as much as if these instructions are removed. given the summary is only visible to the agent, and we want it to feel like we've just continued the conversation I wonder if we can keep this in?

alexhancock: Or does this have a better solution now, always inserting the last user message as well?

katzdave: Ack added this back. This message is agent visible only so should definitely go with what we tested more.

Ideally this should behave about the same as before; I was mostly testing that session reload seemed to have a reasonable state. 

The cool part is we have lots more levers to tweak now; For example since we're keeping all user messages forever, maybe these are actually useful to leave in the session or inject directly in the future summaries.

katzdave:
> From an API design perspective I wonder if we should consider reusing the concept of "audience". MCP resource annotations for example have the concept of:
> 
> `audience: ['user' || 'agent']`
> 
> https://modelcontextprotocol.io/specification/2025-06-18/server/resources#annotations
> 
> I am fine to go with the PR as is though as I think having this controlled in the agent is a great step forward! Then can continue to discuss the properties of `MessageMetadata`
> 
> Just flagging it to you now in case you agree audience would be a good idea, and it's not too big a refactor

Good callout, agree that MessageMetadata does a pretty good job of encapsulation so we could probably follow up with that change if we want to match that pattern a little more directly.

understood-the-assignment:
Hello, it's possible that this pull request introduced a regression: https://github.com/block/goose/issues/4635
Could you have a look, please?

PR #4557, opened 2025-09-08
Fixes and improvements to get custom providers working in the UI

cbruyndoncx:
## Pull Request Description

Description
- This PR completes end-to-end support for custom providers and significantly improves the provider management experience in the desktop UI. It wires custom provider config through core config → provider registry → API client, expands the provider edit modal to expose more fields (base URL, headers/secrets, model selection, etc.), and refactors the Providers grid for a better UX consistent with the extensions grid.

Key changes
- Backend / core
  - Add/extend custom provider configuration handling: crates/goose/src/config/custom_providers.rs
  - Make the provider registry and API client accept dynamic base URLs, headers, and secret injection: crates/goose/src/providers/provider_registry.rs, crates/goose/src/providers/api_client.rs
  - Adjust formats/OpenAI handling to be compatible with custom endpoints: crates/goose/src/providers/formats/openai.rs, crates/goose/src/providers/openai.rs
  - Expose provider config management to the UI and fix small reply route issues: crates/goose-server/src/routes/config_management.rs, crates/goose-server/src/routes/reply.rs

- UI (Electron / React)
  - Reworked Providers grid for improved layout/controls: ui/desktop/src/components/settings/providers/ProviderGrid.tsx
  - Redesigned provider configuration modal to support default/custom flows and surface additional fields:
    - ui/desktop/src/components/settings/providers/modal/ProviderConfiguationModal.tsx
    - ui/desktop/src/components/settings/providers/modal/subcomponents/forms/CustomProviderForm.tsx
    - ui/desktop/src/components/settings/providers/modal/subcomponents/forms/DefaultProviderSetupForm.tsx
    - ui/desktop/src/components/settings/providers/modal/subcomponents/handlers/DefaultSubmitHandler.tsx
  - Added small helper: ui/desktop/src/utils/secretConstants.ts

Motivation
- Custom/dynamic providers were only partially supported. Users need to be able to add, edit, and use custom providers from the desktop UI (including custom endpoints, headers, and models). The UI previously exposed a limited set of provider fields and the providers UX did not match the extensions grid. This PR finishes the integration and brings the UX into parity with extensions.

User-visible changes
- Users can:
  - Add and configure custom providers from Settings → Providers (set base URL, auth header/secret, models).
  - Edit default providers with more exposed configuration fields.
  - Interact with an improved Providers grid that aligns with the extensions experience.

Tests completed:
- create new custom providers (openai / ollama - no anthropic tests done)
- use custom provider (with openai/gpt-oss via openrouter and lmstudio)
- edit (custom) provider settings and verify json put request on all details
- verify api key changes working (true left untouched, real api key changes applied)
- verify model list is fetched for custom providers too

Known issues:
- CLI changes not tackled here, CLI shows wrong ENV vars, will be tackled next
- A lot of logging added to try to fix the openai compatible issues while testing with hyperbolic.xyz. Found that not all openai options are supported in hyperbolic, comment in source which one to remove, but this needs broader design discussion. So the logging does not heart, but best to remove when a better way is found to surface the issues.

### Provider grid
image

### Edit custom provider configuration
Exposing all fields from the json custom providers config
image

### Edit default providers configuration
Exposing all environment vars fields in the UI
image


cbruyndoncx:
@angiejones providers looks a bit like extensions now ... 
Only should it have the wheel button too as long as it is not configured and a button to actuallty "edit the provider config" ?

cbruyndoncx:
@zanesq the display of the labels has 2 ways to generate the label, i have not touched that. It is possible to see when it is an environment variable, but the label removes the underscore (placeholder has a space i think), so it is not truely obvious what environment variable it should be. 
I am under the impression 2 people approached it differently ....

taniacryptid:
@alexhancock replied on Discord, but commenting here that he will take a look today!

cbruyndoncx:
@alexhancock gentle reminder

taniacryptid:
Let me tag other team members across goose dev as well since there has been recent issues and items to review! @DOsinga @michaelneale @zanesq @jamadeo

crates/goose-server/src/routes/config_management.rs
```diff
@@ -104,6 +117,18 @@ pub async fn upsert_config(
...
+    // indicate presence, do not write that boolean into the secret storage. Treat as a no-op.
+    if query.is_secret {
+        if let Value::Bool(b) = &query.value {
+            if *b {
+                tracing::info!(key = %query.key, "Skipping upsert for secret key with boolean true (presence-only)");
+                return Ok(Json(Value::String(format!("Skipped secret upsert for {} (presence-only)", query.key))));
+            }
+        }
+    }
+
```
jamadeo, line 131: this was slightly confusing to me -- where do we have a secret with a boolean and value and why should it be skipped?

crates/goose-server/src/routes/reply.rs
```diff
@@ -265,14 +265,17 @@ async fn reply_handler(
...
+                // send Error event (for telemetry / UI error handling)
+                let _ = stream_event(
+                    MessageEvent::Error { error: err_text.clone() },
                     &task_tx,
                     &cancel_token,
                 )
                 .await;
+                // also send a visible assistant message so the UI shows it inline
+                let assistant_msg = Message::assistant().with_text(format!("Provider error: {}", err_text));
+                let _ = stream_event(MessageEvent::Message { message: assistant_msg }, &task_tx, &cancel_token).await;
```
jamadeo, line 278: if we want to show the error (I thought we already did?) I think it would be better to just render the Error event instead of stream it as a Message

crates/goose/src/providers/formats/openai.rs
```diff
@@ -426,19 +443,63 @@ where
...
+                }
+                // Include the full JSON payload as a last resort
+                details.push_str(&format!("raw: {}", v));
+
+                // Construct a Message that contains the error details so the UI can display it.
+                // Use assistant role so it is shown where other assistant output appears, but
+                // include the error text so renderers can style it as an error.
+                let msg_text = format!("LLM streaming error encountered. See details below:\n{}", details);
+
+                let message = Message::assistant().with_content(MessageContent::text(msg_text));
```
jamadeo, line 492: If we do need to show errors while keeping the stream alive (and I could see how that makes sense) we should have instead a non-terminating error type where we can return these without putting them in a Message

ui/desktop/src/components/settings/providers/modal/ProviderConfiguationModal.tsx
```diff
@@ -53,6 +56,95 @@ export default function ProviderConfigurationModal() {
...
+                  }));
+                }
+                return;
+              }
+              /* eslint-enable @typescript-eslint/no-explicit-any */
+            } catch (e) {
+              console.debug('API client get failed, falling back to fetch', e);
+            }
+
+            // Fallback to a direct fetch
```
jamadeo, line 102: why the fallback? is it because client wasn't initialized with the right endpoint? that should I think be working better now as of https://github.com/block/goose/pull/4338 and we shouldn't need this

ui/desktop/src/components/settings/providers/modal/ProviderConfiguationModal.tsx
```diff
@@ -53,6 +56,95 @@ export default function ProviderConfigurationModal() {
...
+              const client = (
+                clientMod as unknown as { client?: { get?: (...args: any[]) => Promise<any> } }
+              ).client;
+              if (client && typeof client.get === 'function') {
+                const resp = await client.get({
+                  url: '/config/custom-providers/{id}',
+                  path: { id: currentProvider.name },
+                  headers: { 'X-Secret-Key': secretKey },
+                });
+                const body =
```
jamadeo, line 78: this should use the requests in sdk.gen.ts

jamadeo (commented):
I really like the direction and the attention to custom provider config -- thank you @cbruyndoncx ! I think we could get this merged more quickly if you are able to break up the changes, though. Maybe one for the format/streaming changes, one for the configuration changes at least?

Also, for the format changes: can you share some context for which providers need this? It would be good to test them out or at least be generally aware of the variations on openai's format we see in the wild.

crates/goose-server/src/routes/config_management.rs
DOsinga: i think this is because our API returns true when getting a secret value or false if there is no value. however, I don't think we should block setting secret values to true - it is not much a of a value to keep secret. we should just not do this write. is there a particular case where this is happening?

crates/goose-server/src/routes/config_management.rs
```diff
@@ -798,6 +768,124 @@ pub async fn create_custom_provider(
...
+    path = "/config/custom-providers/{id}",
+    request_body = UpdateCustomProviderRequest,
+    responses(
+        (status = 200, description = "Custom provider updated successfully", body = String),
+        (status = 400, description = "Invalid request"),
+        (status = 404, description = "Provider not found"),
+        (status = 500, description = "Internal server error")
+    )
+)]
+pub async fn update_custom_provider(
```
DOsinga, line 782: can you remove the over eager LLM comments here?

crates/goose-server/src/routes/config_management.rs
```diff
@@ -798,6 +768,124 @@ pub async fn create_custom_provider(
...
+    State(state): State<Arc<AppState>>,
+    headers: HeaderMap,
+    axum::extract::Path(id): axum::extract::Path<String>,
+    Json(request): Json<UpdateCustomProviderRequest>,
+) -> Result<Json<String>, StatusCode> {
+    verify_secret_key(&headers, &state)?;
```
DOsinga, line 788: we got rid of this, but presumably syncing to main will tell you

crates/goose-server/src/routes/reply.rs
DOsinga: when does this happen?

crates/goose-server/src/routes/reply.rs
```diff
@@ -312,8 +315,32 @@ async fn reply_handler(
...
+                                .content
+                                .iter()
+                                .find_map(|c| match c {
+                                    goose::conversation::message::MessageContent::Text(t) => Some(t.text.clone()),
+                                    _ => None,
+                                })
+                                .filter(|s| s.starts_with("LLM streaming error encountered"))
+                            {
+                                // Send a short error string (avoid flooding the SSE with huge payloads)
+                                let short = if first_text.len() > 1024 {
```
DOsinga, line 337: don't use .len or [..] for text since that breaks CJK. we have a function to this, safe_truncate

crates/goose-server/src/routes/reply.rs
```diff
@@ -312,8 +315,32 @@ async fn reply_handler(
                                 track_tool_telemetry(content, all_messages.messages());
                             }
 
+                            // Push and send the message event
```
DOsinga, line 318: remove this comment

crates/goose-server/src/routes/reply.rs
```diff
@@ -312,8 +315,32 @@ async fn reply_handler(
...
                             all_messages.push(message.clone());
-                            stream_event(MessageEvent::Message { message }, &tx, &cancel_token).await;
+                            stream_event(MessageEvent::Message { message: message.clone() }, &tx, &cancel_token).await;
+
+                            // If this message appears to be a provider streaming error produced by
```
DOsinga, line 322: this doesn't feel right.

ui/desktop/src/components/settings/providers/modal/subcomponents/handlers/DefaultSubmitHandler.tsx
```diff
@@ -74,6 +76,12 @@ export const DefaultSubmitHandler = async (
       // Explicitly define is_secret as a boolean (true/false)
       const isSecret = parameter.secret === true;
```
DOsinga, line 77: hmmm

ui/desktop/src/components/settings/providers/modal/subcomponents/handlers/DefaultSubmitHandler.tsx
```diff
@@ -74,6 +76,12 @@ export const DefaultSubmitHandler = async (
...
 
+      // If this is a secret value and the sentinel indicates the secret is present
+      // but unchanged, skip writing to avoid overwriting the keyring with the sentinel.
+      if (isSecret && value === SECRET_PRESENT_SENTINEL) {
```
DOsinga, line 81: we should be able to do better than that I would think; can't we just use a dirty dict here? would also allow us to skip if no updates were made

ui/desktop/src/utils/secretConstants.ts
```diff
@@ -0,0 +1 @@
+export const SECRET_PRESENT_SENTINEL = '__GOOSE_SECRET_PRESENT__';
```
DOsinga, line 1: no one line files please

ui/desktop/src/components/settings/providers/modal/subcomponents/forms/DefaultProviderSetupForm.tsx
```diff
@@ -10,13 +14,16 @@ interface DefaultProviderSetupFormProps {
   setConfigValues: React.Dispatch<React.SetStateAction<Record<string, string>>>;
   provider: ProviderDetails;
   validationErrors: ValidationErrors;
+  // Optional callback invoked when a field is edited so parent can clear errors
+  onFieldChange?: (name: string, value: string) => void;
```
DOsinga, line 18: are we using this? I didn't see the caller so quickly

ui/desktop/src/components/settings/providers/modal/subcomponents/forms/DefaultProviderSetupForm.tsx
```diff
@@ -25,65 +32,64 @@ export default function DefaultProviderSetupForm({
...
-  console.log('configValues default form', configValues);
-
   // Initialize values when the component mounts or provider changes
   const loadConfigValues = useCallback(async () => {
+    // If there are no parameters, nothing to load
+    if (parameters.length === 0) {
+      setIsLoading(false);
+      return;
+    }
+
```
DOsinga, line 42: why short cut this? the alternative is an empty loop. also, isn't setIsLoading already false here?

ui/desktop/src/components/settings/providers/modal/subcomponents/forms/DefaultProviderSetupForm.tsx
```diff
@@ -25,65 +32,64 @@ export default function DefaultProviderSetupForm({
...
       } catch (error) {
         console.error(`Failed to load config for ${parameter.name}:`, error);
-        // Fall back to default if read operation fails
-        if (
-          parameter.default !== undefined &&
-          parameter.default !== null &&
-          !configValues[parameter.name]
-        ) {
-          newValues[parameter.name] = String(parameter.default);
+        if (parameter.default !== undefined && parameter.default !== null) {
```
DOsinga, line 62: what is the scenario where this is the right thing to do?

crates/goose/src/providers/utils.rs
```diff
@@ -86,12 +86,32 @@ pub fn map_http_error_to_provider_error(
                 if check_context_length_exceeded(&payload_str) {
                     ProviderError::ContextLengthExceeded(payload_str)
                 } else {
+                    // Try multiple ways to extract a useful message
```
DOsinga, line 89: Let's not use "Unknown error" as a way to flag if we have already handled this case (for one thing if the error is really unknown, we might now return the payload_str). but also, what are we trying to cover here? which providers have these specific behaviors we're trying to rescue from?

crates/goose/src/providers/formats/openai.rs
```diff
@@ -36,6 +36,10 @@ struct Delta {
     content: Option<String>,
     role: Option<String>,
     tool_calls: Option<Vec<DeltaToolCall>>,
+    // Some providers put a 'name' (tool name) here instead of in tool_calls
+    name: Option<String>,
+    // OpenAI/variants may include a 'refusal' field with refusal details
+    refusal: Option<Value>,
```
DOsinga, line 42: here and elsewhere, we talk about "some providers".  I think we need to explicitly mention which ones we are talking about. in an ideal world we would just have a dialect flag but I can see how that is hard from a custom provider perspective

crates/goose/src/config/custom_providers.rs
```diff
@@ -148,8 +148,37 @@ pub fn register_custom_providers(
...
+        }
+    }
+
     for config in configs {
         let config_clone = config.clone();
+        // Use a unique base URL key per custom provider to avoid collisions in the
+        // global config/keyring. Previously this used the constant
+        // "CUSTOM_PROVIDER_BASE_URL" for every provider which caused different
+        // providers to read/write the same key and mix up values stored in the
+        // keyring or config file.
```
DOsinga, line 180: previously? the reader doesn't care how this previously worked

crates/goose/src/config/custom_providers.rs
```diff
@@ -148,8 +148,37 @@ pub fn register_custom_providers(
 ) -> Result<()> {
     let configs = load_custom_providers(dir)?;
 
+    // Detect legacy shared key usage in keyring/config that could override
```
DOsinga, line 151: I'm not sure I follow this. I don't think custom providers should support custom proivder base urls beyond what they specify in the json. why do we look at the config here?

in general I don't think custom providers should access the config at all. they are already custom

cbruyndoncx:
Thanks for the review. Yes, Goose vibed with me. 
I noticed almost no comments in the goose source.

I have merged v1.8.0 locally, and creating detailed PRs for the different pieces. 

@DOsinga 
I don't know why the custom providers should interact with their own custom settings, that seems confusing.

I don't understand your remark

DOsinga:
sorry for the slow reply.

what I mean is that a custom provider doesn't need an environment variable to override its base url; it's a custom provider, it already has a custom base url. etc. does that make sense?

crates/goose-server/src/routes/config_management.rs
cbruyndoncx: in the windows ui the true value seemed to cause keyring issues. I have tried a detailed split , but that failed. Retry with a subset, to get the custom provider bugs fixed first.

crates/goose/src/config/custom_providers.rs
cbruyndoncx: Previously if you just had a single custom provider it worked, but failed with the second one

DOsinga: I understand the comment, but comments should help the reader understand the current code. Either way, custom providers should not use environment variables where the values are defined for the custom provider itself.

cbruyndoncx: yes, that i understand now, trying to figure out when/where they are introduced. I feel like i am on a treasure hunt in the code ... 
I thought environment variables were always allowed, but i understand it slightly better now how this new design is supposed to work.

crates/goose/src/config/custom_providers.rs
cbruyndoncx: Understood now - things become clearer

cbruyndoncx:
@DOsinga I am closing this initial big PR in favor of https://github.com/block/goose/pull/4781 which has only the necessary changes to get the openai compatible working, tested with openrouter.
Video in added in new PR.

PR #4565, opened 2025-09-08
Compact session automatically for streaming providers on Context Length Exceeded

katzdave:
This is mostly a good experience, but doesn't continue the agent loop automatically; The user needs to send an input


michaelneale:
close - yeah must be a way to fire off a message to keep it going?

michaelneale (approved):
I believe an improvement but probably should find a way to make it continue on as this will appear like a stall

katzdave:
> should this be not in goose-cli?

Yeah I really don't like how much cli specific code is here. I think most of this file can get nuked as part of a bigger refactor. 

I did figure out a way to keep the stream open for handling this specific error.

PR #4567, opened 2025-09-09
Handle long input lists gracefully, read nested messages in JSON response from databricks

amberdixon:
## Pull Request Description

Sometimes databricks comes back with this error message. We should parse out the message in the JSON body and handle it gracefully by treating it as context exceeded:

```
Error: Request failed: HTTP 400 Bad Request: {"error_code":"BAD_REQUEST","message":"{\"external_model_provider\":\"databricks-model-serving\",\"external_model_error\":{\"error_code\":\"BAD_REQUEST\",\"message\":\"{\\\"message\\\":\\\"Input is too long for requested model.\\\"}\"}}"}

```

After my change, the following occurred when the input is too long error came back:
```
Let me examine the full results:
path: /v/f/l/b/T/goose_mcp_responses/mcp_response_20250909_001603.txt
command: view

Error: Context length exceeded: Input is too long for requested model.
  error: Context length exceeded error.
The conversation is too long for the model's context window.
Consider using /summarize to condense the conversation history
or /clear to start fresh.
We've removed the conversation up to the most recent user message.

```

### DIFF
--- a/crates/goose/src/providers/utils.rs
+++ b/crates/goose/src/providers/utils.rs
@@ -64,6 +64,30 @@ fn check_context_length_exceeded(text: &str) -> bool {
         .any(|phrase| text_lower.contains(phrase))
 }
 
+/// Extract error message from Databricks' nested error format
+/// Handles format like: {"error_code":"BAD_REQUEST","message":"{\"external_model_provider\":\"databricks-model-serving\",\"external_model_error\":{\"error_code\":\"BAD_REQUEST\",\"message\":\"{\\\"message\\\":\\\"Input is too long for requested model.\\\"}\"}}"}
+fn extract_databricks_error_message(payload: &Value) -> Option<String> {
+    // First try to get the message field
+    if let Some(message_str) = payload.get("message").and_then(|m| m.as_str()) {
+        // Try to parse the nested JSON in the message field
+        if let Ok(nested_json) = serde_json::from_str::<Value>(message_str) {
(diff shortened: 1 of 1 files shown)


michaelneale:
is this complimenting https://github.com/block/goose/pull/4565 - by @katzdave ?

michaelneale:
thanks @amberdixon  - mind running `cargo fmt` and push again?

alexhancock:
@amberdixon Yeah if you could do `cargo fmt` and push and then bonus/good if we can add some tests that would fail before the change, and pass after!

katzdave:
Are these still falling through on main after https://github.com/block/goose/pull/4506?

I'm fine with going this route if so, although not sure why the string matching isn't picking up since it has 'too long' inside the payload.

Yeah https://github.com/block/goose/pull/4565 should handle these errors better as well.

amberdixon:
> not sure why the string matching isn't picking up since it has 'too long' inside the payload.

I'm really curious to know why as well. I'm going to try to repro the problem again with some logging to help me understand why the existing string checks aren't working before i introduce new code!

amberdixon:
I think https://github.com/block/goose/pull/4506 is probably the correct fix and it just got released in 1.8.0, so I'm going to report back here if I do happen to see this problem again after using 1.8.

alexhancock:
awesome, thanks for engaging and making sure it's resolved some way @amberdixon!

PR #4568, opened 2025-09-09
speed up from async naming and saving

michaelneale:
Based on research here: https://github.com/block/goose/discussions/4554

it seems goose in short session (at least) adds a fair bit of time (a few seconds) for using the LLM for naming and then hyper frequent session saves. 

This change will save about 1.5/2s for name generation + 1.5s out of a 8 second session spend saving the session, making a 40% or so saving for very short interactions, which improves the perception. with ongoing usage, it will feel a bit snappier as there are no subsecond pauses occasionally for saves - helps the CLI and the GUI.

This change proposes to make both the name generation and the saving happen in the background (it is a trade off, but for most purposes is fine) and not on the critical path of the user experience, so things feel faster and are faster wall clock wise.


crates/goose/src/session/storage.rs
```diff
@@ -396,6 +532,56 @@ pub fn list_sessions() -> Result<Vec<(String, PathBuf)>> {
     Ok(entries)
 }
 
+/// Initialize the background session saver
+pub fn init_background_session_saver() {
```
alexhancock: Looks unused? Maybe an extra copy that was iterated on before finishing `init_session_saver`?

crates/goose/src/session/storage.rs
```diff
@@ -1356,6 +1542,100 @@ pub async fn update_metadata(session_file: &Path, metadata: &SessionMetadata) ->
...
+    messages: &Conversation,
+    provider: Option<Arc<dyn Provider>>,
+    schedule_id: Option<String>,
+    working_dir: Option<PathBuf>,
+) -> Result<()> {
+    // Validate the session file path for security
+    let secure_path = get_path(Identifier::Path(session_file.to_path_buf()))?;
+
+    // Security check: message count limit
+    if messages.len() > MAX_MESSAGE_COUNT {
```
jamadeo, line 1407: why do we have a max message count?

crates/goose/src/session/storage.rs
```diff
@@ -1356,6 +1542,100 @@ pub async fn update_metadata(session_file: &Path, metadata: &SessionMetadata) ->
...
+    // Update the schedule_id if provided
+    if schedule_id.is_some() {
+        metadata.schedule_id = schedule_id.clone();
+    }
+
+    // Create the task for background saving/naming
+    let task = if let Some(provider) = provider.as_ref() {
+        if user_message_count < 4 {
+            // For early messages that need session naming, use the GenerateName task
+            PersistenceTask::GenerateName {
```
jamadeo, line 1441: nit: I first thought creating only a GenerateName task would mean we don't save, maybe better named as `GenerateNameAndSave` or something like that

crates/goose/src/session/storage.rs
```diff
@@ -20,14 +20,150 @@ use std::fs;
...
+        }
+
+        // Save any remaining tasks before exiting
+        if !batch.is_empty() {
+            save_batch(&mut batch).await;
+        }
+    });
+
+    // Store the task handle to prevent it from being dropped
+    let _ = _SESSION_SAVER_TASK.set(task_handle);
```
jamadeo, line 115: do we need to do anything here to make sure it flushes if the process quits?

crates/goose/src/session/storage.rs
```diff
@@ -20,14 +20,150 @@ use std::fs;
...
+    tx
+}
+
+/// Save a batch of persistence tasks
+async fn save_batch(batch: &mut Vec<PersistenceTask>) {
```
jamadeo: Most of the time we'll be queueing saves to the same file, but it wouldn't make sense to batch them, we should just drop all but the latest one, right?

jamadeo (commented):
Glad to see this looked at, and wow those timings are not great.

crates/goose/src/session/storage.rs
michaelneale: yeah - I was wondering that - I believe at the moment it doesn't if you hard quit (which is ok... but not ideal)

crates/goose/src/session/storage.rs
michaelneale: yeah I think so, this was more a port from sync to async. The sync code was hammering the file for some reason, was surprising.

crates/goose/src/session/storage.rs
michaelneale: yeah I think I tried a few approaches but never got to tidy up

crates/goose/src/session/storage.rs
michaelneale: @jamadeo unrelated to this change, same as on main (perhaps the renaming confused the git diff, but this is no change), not sure why they are there, but they are already there all over the place, so just following the existing pattern.

crates/goose/src/session/storage.rs
michaelneale: yeah tried to add something in for this

michaelneale:
some more changes @jamadeo if you think in right direction, not entirely happy with it as ends up more complicated than I thought, but still seems like a reasonable idea, and when it works, *feels* snappier but I can't tell if that is my bias. Worth continuing?

Started out being simple, but having it batch things up and efficiently save makes it a more complicated change than a quickie I had hoped for

crates/goose/src/session/storage.rs
jamadeo: yeah, it seemed like an odd limitation. I could see maybe there was a bug in the past that produced endlessly growing sessions, but more likely seemed like something an LLM would add in for no obvious reason. But if we do keep session history around after compaction, and we have long-running processes, I could see this breaking

jamadeo:
I know what you mean, this does become a bit heavyweight.

Maybe what we should be starting with is decoupling the name generation and the session file saving. The first one should definitely be async and expected to be slow. The second should be fast and you might not want to set-and-forget.

michaelneale:
I was surprised to see up to 2s of saving - could be a measurement artifact, or jsonl serialization. Curious if this one feels wall clock faster to you?

michaelneale:
closing this as I think are better approaches, shows it can help tho!

PR #4585, opened 2025-09-09
fix markdown links overflowing content and hide agent link previews

zanesq:
## Pull Request Description

Links in markdown weren't wrapping properly. Noticed our css classes were concatenated. Also noticed link previews were showing mid chat so disabled that again.


ui/desktop/src/components/GooseMessage.tsx
```diff
@@ -305,7 +305,8 @@ export default function GooseMessage({
       </div>
 
       {/* TODO(alexhancock): Re-enable link previews once styled well again */}
-      {urls.length > 0 && (
+      {/* eslint-disable-next-line no-constant-binary-expression */}
+      {false && urls.length > 0 && (
```
jamadeo: I'd go with just commenting or deleting it, but this works too

jamadeo (approved):

ui/desktop/src/components/GooseMessage.tsx
zanesq: yeah good point I am going to delete these for now since they have been here for a long time

zanesq: well it turns out to be a larger refactor, will follow up 👍

PR #4592, opened 2025-09-10
feat: goose swarm

michaelneale:
`goose swarm --repo michaelneale/test-repo`

This is a draft of a new mode of goose, where you run N instances of goose which will co-ordinate and collaborate to solve issues and implement things end to end around a github repository. Co-ordination/state is entirely via github (code, issues etc). The aim isn't to interact with people via issues but to start with a request (as high or low level as desired) and let the swarm break that down into other tasks and issues, or chains of tasks and issues (allowing for very long running evolving changes). Different peers contributing different environments helps as well (and planning will take into account number of peers and how much it can experiment).  Github actions as well as local workflows (local to the agent running `goose swarm`) work on solving/proving and validating solutions before the evaluator choses and integrates them. 

Usage note: 
Use on a new or existing repo you don't mind running things unsupervised. With at least 1 node, you can open an issue with any vague request and label it "help wanted" and the rest is bootstrapped from there. 

WIP: 

- [ ] scaled testing to make sure there isn't collisions
- [x] whitepaper [describing](https://github.com/block/goose/blob/900cc070a5b77cace4de843d749896ed9506fb53/crates/goose-cli/src/commands/goose_swarm/whitepaper.pdf) it at a high level
- [x] repo isolation (need to name/clean up local clones incase someone switches repo)
- [ ] consider git worktrees for local efficiency
- [x] fix so loops aren't un-killable
- [ ] automated testing and benchmarking
- [ ] reap dead claims and resume

Screenshot 2025-09-10 at 4 58 27 pm

### DIFF
--- a/crates/goose-cli/src/commands/swarm.rs
+++ b/crates/goose-cli/src/commands/swarm.rs
@@ -780,17 +780,17 @@ async fn run_recipe(recipe_content: &str, params: Vec<(String, String)>) -> Resu
             .map(|(_, v)| v.as_str())
             .unwrap_or("0");
 
-        Some(goose::session::Identifier::Name(format!(
+        Some(format!(
             "swarm-drone-{}-task-{}",
             worker_id, task_number
-        )))
+        ))
     } else {
         None
     };
 
     // Build and run the session
     let mut session = build_session(SessionBuilderConfig {
-        identifier: session_identifier.clone(),
+        session_id: session_identifier.clone(),
         resume: false,
         no_session: !needs_session, // Use session for drone tasks
         extensions: Vec::new(),


ariane-emory:
Suggestion: why not call it a 'flock' instead of a 'swarm'?

michaelneale:
@ariane-emory good Q - or "skein" (as in flight), I dunno, just thought lots of people already use flock/gaggle/skein or maybe subconsciously I wanted a different name, but of all those, flock is apt! (assuming no one wants to use it for other things).

(I also loved, back in the day, the simplicity of docker swarm - although it never won the war over kubernetes, it was elegant)

Would happily change it!

harkamals:
I hope GitHub repos are not the only supported ones. 

We should be able to use any git repo, ex: azure DevOps repos

michaelneale:
@harkamals yes - at the moment github - but I made a start of ab abstraction in `repo.rs` for exactly that reason. I would need to plug in another impl to get this right though and then have the abstraction. With a git repo - it would be hacky but could use a hidden branch to sync the state possibly? (then it would work with any?) vs having gitlab, azure implementations perhaps (although in reality, want to knwo their apis/clis as want to be able to track equivalent of PRs)... so yeah, maybe just need to bite the bullet and have another one. Gitlab?

DOsinga:
closing because stale

PR #4598, opened 2025-09-10
feat: set custom models for lead/worker

kamal:
## Pull Request Description

This PR adds the ability to set a custom model ID for lead and/or worker.

Screenshot 2025-09-10 at 10 54 03 AM

Long term, would it be more consistent to use the Provider -> Model components as seen in the Switch Model modal?

### DIFF
--- a/ui/desktop/src/components/settings/models/subcomponents/LeadWorkerSettings.test.tsx
+++ b/ui/desktop/src/components/settings/models/subcomponents/LeadWorkerSettings.test.tsx
@@ -1,0 +1,154 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, waitFor, fireEvent } from '@testing-library/react';
+import { LeadWorkerSettings } from './LeadWorkerSettings';
+
+// Mock predefined models utils to force provider-based options (no predefined list)
+vi.mock('../predefinedModelsUtils', () => ({
+  shouldShowPredefinedModels: () => false,
+  getPredefinedModelsFromEnv: () => [],
+}));
+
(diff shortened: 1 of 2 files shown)


DOsinga:
I'm sorry, I thought I had replied to this. what does the custom model mean here, are you trying to enable having different models from different providers and doing that through a composable string?

kamal:
@DOsinga this modal only allows you to choose from the dropdown of models declared by each provider that you've connected. However, there's always been a way to provide your own model ID that is not listed in the dropdown. Take for example the AWS Bedrock provider. The switch model modal only lists Sonnet 3.5, but it allows you to set a custom model. With this PR, I'm extending that capability to the lead/worker modal.

Screenshot 2025-09-29 at 9 45 48 AM

DOsinga:
> @DOsinga this modal only allows you to choose from the dropdown of models declared by each provider that you've connected. However, there's always been a way to provide your own model ID that is not listed in the dropdown. Take for example the AWS Bedrock provider. The switch model modal only lists Sonnet 3.5, but it allows you to set a custom model. With this PR, I'm extending that capability to the lead/worker modal.
> 
> Screenshot 2025-09-29 at 9 45 48 AM

well, I didn't know that. maybe we can do a quick follow up where we reuse the entire provider/model selector thing?

kamal:
@DOsinga i'm going to rebase this against latest so that the CI github action will run now that you've approved it

kamal:
@DOsinga looks like you'll have to approve the workflows to run

DOsinga:
sorry this is annoying

PR #4602, opened 2025-09-10
Add recipe schema validation to UI

zanesq:
## Pull Request Description

Now that we have import recipe from file we should validate it as well.

Added recipe schema validation with tests:
- Dynamic validation based on the currently generated openapi schema 
- Validates using existing Zod library that we use for form validation
- JSON Schema validation is basic to avoid another library and eval issues with CSP settings

Bonus: added response JSON Schema field and generated schema reference help text in the create recipe UI

Note: this is a minor improvement before we unify the create forms across chat and recipe library.
Screenshot 2025-09-10 at 12 48 43 PM
Screenshot 2025-09-10 at 12 53 31 PM
Screenshot 2025-09-10 at 12 55 36 PM
Screenshot 2025-09-10 at 12 53 40 PM

### DIFF
--- a/ui/desktop/src/components/recipes/ImportRecipeForm.tsx
+++ b/ui/desktop/src/components/recipes/ImportRecipeForm.tsx
@@ -360,7 +360,7 @@ export default function ImportRecipeForm({ isOpen, onClose, onSuccess }: ImportR
                                 className="text-xs text-blue-500 hover:text-blue-700 underline"
                                 disabled={isDisabled}
                               >
-                                example schema
+                                example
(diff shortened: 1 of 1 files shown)


ui/desktop/src/recipe/validation.test.ts
```diff
@@ -0,0 +1,565 @@
+import { describe, it, expect } from 'vitest';
```
michaelneale, line 1: nice touch

michaelneale (approved):

PR #4604, opened 2025-09-10
When the developer extension gets a cancellation message, it should kill any running processes that it owns.

angelahning:
Addresses https://github.com/block/goose/issues/4125

https://github.com/user-attachments/assets/fb09ff1d-7385-4257-9098-fbfc73f46fb3

### DIFF
--- a/crates/goose-mcp/src/developer/rmcp_developer.rs
+++ b/crates/goose-mcp/src/developer/rmcp_developer.rs
@@ -20,14 +20,12 @@ use std::{
     future::Future,
     io::Cursor,
     path::{Path, PathBuf},
-    process::Stdio,
     sync::{Arc, Mutex},
 };
 use xcap::{Monitor, Window};
 
 use tokio::{
     io::{AsyncBufReadExt, BufReader},
-    process::Command,
     sync::RwLock,
 };
 use tokio_stream::{wrappers::SplitStream, StreamExt as _};
@@ -36,7 +34,7 @@ use tokio_util::sync::CancellationToken;
 use super::analyze::{types::AnalyzeParams, CodeAnalyzer};
 use super::editor_models::{create_editor_model, EditorModel};
 use super::goose_hints::load_hints::{load_hint_files, GOOSE_HINTS_FILENAME};
-use super::shell::{expand_path, get_shell_config, is_absolute_path};
+use super::shell::{configure_shell_command, expand_path, get_shell_config, is_absolute_path, kill_process_group};
 use super::text_editor::{
     text_editor_insert, text_editor_replace, text_editor_undo, text_editor_view, text_editor_write,
 };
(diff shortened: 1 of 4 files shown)


crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -492,6 +498,27 @@ impl ServerHandler for DeveloperServer {
...
     }
+
+    /// Called when the client cancels a specific request.
+    /// This method cancels the running process associated with the given request_id.
+    #[allow(clippy::manual_async_fn)]
+    fn on_cancelled(
+        &self,
+        notification: CancelledNotificationParam,
+        _context: NotificationContext<RoleServer>,
+    ) -> impl Future<Output = ()> + Send + '_ {
```
jamadeo, line 520: I think you can change this fn to 

```
async fn on_cancelled(
    ...
) {
```

and then remove the `async move {` part

crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -799,7 +827,7 @@ impl DeveloperServer {
     /// this tool does not run indefinitely.
     #[tool(
         name = "shell",
-        description = "Execute a command in the shell. Returns output and error concatenated. Avoid commands with large output, use background commands for long-running processes."
+        description = "Execute a command in the shell.This will return the output and error concatenated into a single string, as you would see from running on the command line. There will also be an indication of if the command succeeded or failed. Avoid commands that produce a large amount of output, and consider piping those outputs to files. If you need to run a long lived command, background it - e.g. `uvicorn main:app &` so that this tool does not run indefinitely."
```
jamadeo, line 842: is this from another change?

crates/goose-mcp/src/developer/shell.rs
```diff
@@ -105,3 +109,70 @@ pub fn normalize_line_endings(text: &str) -> String {
...
+    {
+        if let Some(pid) = pid {
+            // Try SIGTERM first
+            let _sigterm_result = unsafe { libc::kill(-(pid as i32), libc::SIGTERM) };
+
+            // Wait a brief moment for graceful shutdown
+            tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
+
+            // Force kill with SIGKILL
+            let _sigkill_result = unsafe { libc::kill(-(pid as i32), libc::SIGKILL) };
```
jamadeo, line 158: do you need this if you always call child.kill() on the tokio process?

crates/goose-mcp/src/developer/rmcp_developer.rs
angelahning: The previous rmcp refactor trimmed the tool description. I am just adding it back.

crates/goose-mcp/src/developer/rmcp_developer.rs
angelahning: I had that before, but on_cancelled wasn't getting triggered.

crates/goose-mcp/src/developer/shell.rs
angelahning: this kills the entire process group, vs child.kill() just kills the main process.

PR #4605, opened 2025-09-10
Run the conversation fixer over messages for recipe create

jamadeo:
We've seen orphaned tool requests in the wild causing the create-recipe flow to fail. We should figure out why that happens, but given that we'll be moving the source of truth to the session file and away from the frontend state at some point it seems better to invest our efforts in tracking down any issues there.


zanesq:
Might be worth getting into 1.8.0 if we feel its safe enough

PR #4620, opened 2025-09-12
fix: speed up session by not using LLM session naming

michaelneale:
According to some analysis in a short session up to 2s (especially early on) is used calculating the name of the session. It seems overkill to name the session with an LLM at the moment.

### DIFF
--- a/crates/goose/src/providers/base.rs
+++ b/crates/goose/src/providers/base.rs
@@ -460,28 +460,46 @@ pub trait Provider: Send + Sync {
     }
 
     /// Generate a session name/description based on the conversation history
-    /// Creates a prompt asking for a concise description in 4 words or less.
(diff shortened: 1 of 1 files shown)


alexhancock (approved):

michaelneale:
will see if @DOsinga happy - we could combine this with more background persistence so it isn't in the critical path if we still want nice names (or eventually a fast light local model)

DOsinga (changes requested):
As discussed, I don't think this is the way to do this - I think we should just replace the entire session thing with a sqlite solution

alexhancock:
Fair enough! I guess I saw this as a simpler approach to part of the problem while we figure out how to transition to sqlite for session storage overall.

PR #4622, opened 2025-09-12
Fix #4550: Correct row/column transposition in xlsx tool

### LINKED ISSUE #4550: Xlsx Tool of Computer controller seems to transpose rows and columns.
**Describe the bug**

Xlsx Tool of Computer controller seems to transpose rows and columns.

**To Reproduce**

1. Enable Computer Controller
2. on Chat, Send 'Get the velue from A2 of ...FinancialSample.xlsx'
3. Allow execute
4. Returns the wrong value (it seems like B1's value)

**Expected behavior**
In step 4, returns the value of A2.

**Screenshots**

Image

:

Image

 - **OS & Arch:** [Ubuntu 24.04 x86]
 - **Interface:** [UI]
 - **Version:** [v1.7.0]
 - **Extensions enabled:** [Computer Controller, Memory]
 - **Provider & Model:** [OpenRouter - deepseek/deepseek-chat-v3.1]

jalateras:
## Summary
- Fixed coordinate mapping issue in xlsx tool where cell references were transposed
- Updated `parse_cell_reference` to return (row, column) instead of (column, row)
- Added comprehensive tests to verify correct behavior

## Problem
The xlsx tool was incorrectly transposing row and column coordinates. When requesting cell A2 (row 2, column 1), it would return the value from B1 (row 1, column 2) instead. This was caused by `parse_cell_reference` returning coordinates in the wrong order compared to what the umya_spreadsheet library expects.

## Solution
- Changed `parse_cell_reference` to return `(row, col)` instead of `(col, row)` to match umya_spreadsheet's coordinate expectation
- Updated `parse_range` and `get_range` functions to handle the corrected coordinate order
- Added two new tests:
  - `test_coordinate_mapping`: Verifies the coordinate system mapping
  - `test_issue_4550_row_column_transposition`: Specifically tests the reported issue

## Test Plan
- [x] All existing tests pass
- [x] New tests verify correct cell addressing
- [x] Code passes formatting (cargo fmt)
- [x] Code passes linting (cargo clippy)
- [x] Code compiles without warnings (cargo check)

Fixes #4550

### DIFF
--- a/crates/goose-mcp/src/computercontroller/xlsx_tool.rs
+++ b/crates/goose-mcp/src/computercontroller/xlsx_tool.rs
@@ -97,7 +97,7 @@ impl XlsxTool {
     }
 
     pub fn get_range(&self, worksheet: &Worksheet, range: &str) -> Result<RangeData> {
-        let (start_col, start_row, end_col, end_row) = parse_range(range)?;
+        let (start_row, start_col, end_row, end_col) = parse_range(range)?;
         let mut values = Vec::new();
 
         // Iterate through rows first, then columns
@@ -206,7 +206,7 @@ impl XlsxTool {
 }
 
 fn parse_range(range: &str) -> Result<(u32, u32, u32, u32)> {
-    // Handle ranges like "A1:B10"
+    // Handle ranges like "A1:B10" and return (start_row, start_col, end_row, end_col)
     let parts: Vec<&str> = range.split(':').collect();
     if parts.len() != 2 {
         anyhow::bail!("Invalid range format. Expected format: 'A1:B10'");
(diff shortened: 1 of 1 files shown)


alexhancock (approved):

crates/goose-mcp/src/computercontroller/xlsx_tool.rs
```diff
@@ -328,4 +329,77 @@ mod tests {
...
+
+    #[test]
+    fn test_issue_4550_row_column_transposition() -> Result<()> {
+        // This test specifically addresses issue #4550 where A2 was returning B1's value
+        let xlsx = XlsxTool::new(get_test_file())?;
+        let worksheet = xlsx.get_worksheet_by_index(0)?;
+
+        // Test that A2 (row 2, column 1) returns the correct value
+        let a2_value = xlsx.get_cell_value(worksheet, 2, 1)?;
+        assert_eq!(a2_value.value, "Country", "A2 should contain 'Country'");
```
kita-sato, line 378: Is this test correct? I think the value in cell A2 is 'Government'.
image

PR #4625, opened 2025-09-12
Add recipe title in import form

amed-xyz:
Users importing recipes through the Desktop UI should be able to directly set and modify the recipe's title, with the app automatically handling file naming behind the scenes. This eliminates the current confusion between what's shown in the UI and storage filenames.

### DIFF
--- a/ui/desktop/src/components/recipes/ImportRecipeForm.tsx
+++ b/ui/desktop/src/components/recipes/ImportRecipeForm.tsx
@@ -32,7 +32,7 @@ const importRecipeSchema = z
         (value) => !value || value.trim().startsWith('goose://recipe?config='),
         'Invalid deeplink format. Expected: goose://recipe?config=...'
       ),
-    yamlFile: z
+    recipeUploadFile: z
       .instanceof(File)
       .nullable()
       .refine((file) => {
@@ -50,8 +50,8 @@ const importRecipeSchema = z
       ),
     global: z.boolean(),
   })
-  .refine((data) => (data.deeplink && data.deeplink.trim()) || data.yamlFile, {
-    message: 'Either of deeplink or YAML file are required',
+  .refine((data) => (data.deeplink && data.deeplink.trim()) || data.recipeUploadFile, {
+    message: 'Either of deeplink or recipe file are required',
     path: ['deeplink'],
   });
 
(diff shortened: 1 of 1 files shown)


ui/desktop/src/components/recipes/ImportRecipeForm.tsx
```diff
@@ -125,30 +156,39 @@ export default function ImportRecipeForm({ isOpen, onClose, onSuccess }: ImportR
...
+
         const validationResult = validateRecipe(recipe);
         if (!validationResult.success) {
           const errorMessages = getValidationErrorMessages(validationResult.errors);
           throw new Error(`Recipe validation failed: ${errorMessages.join(', ')}`);
         }
 
         await saveRecipe(recipe, {
-          name: value.recipeName.trim(),
+          name: '',
```
amed-xyz, line 189: `name` should now be considered deprecated and will be removed once the same functionality is incorporated in recipe creation form

ui/desktop/src/recipe/recipeStorage.ts
```diff
@@ -70,12 +71,21 @@ async function saveRecipeToFile(recipe: SavedRecipe): Promise<boolean> {
...
+    recipe.title = title.trim();
+    sanitizedName = generateRecipeFilename(recipe);
+    if (!sanitizedName) {
+      throw new Error('Invalid recipe title - cannot generate filename');
+    }
+  } else {
+    sanitizedName = sanitizeRecipeName(name);
+    if (!sanitizedName) {
+      throw new Error('Invalid recipe name');
+    }
```
amed-xyz: This logic should now be considered deprecated and will be removed once the same functionality is incorporated in recipe creation form

ui/desktop/src/components/recipes/ImportRecipeForm.tsx
zanesq: is there a drawback to continuing to set name for now for backwards compatibility purposes? Also we might want to indicate that its deprecated in the api/types and check that its actually ok to deprecate from a backend perspective before we deprecate it.

ui/desktop/src/recipe/recipeStorage.ts
zanesq: Maybe a comment needed for future dev?

ui/desktop/src/recipe/recipeStorage.ts
```diff
@@ -70,12 +71,21 @@ async function saveRecipeToFile(recipe: SavedRecipe): Promise<boolean> {
...
  */
 export async function saveRecipe(recipe: Recipe, options: SaveRecipeOptions): Promise<string> {
-  const { name, global = true } = options;
+  const { name, title, global = true } = options;
 
-  // Sanitize name
-  const sanitizedName = sanitizeRecipeName(name);
-  if (!sanitizedName) {
-    throw new Error('Invalid recipe name');
+  let sanitizedName: string;
```
zanesq, line 76: Should we call it `title` instead of `name` here now?

ui/desktop/src/components/recipes/ImportRecipeForm.tsx
amed-xyz: > is there a drawback to continuing to set name for now for backwards compatibility purposes?

No drawback, but `value.recipeName` used to come from an input field that's been replaced by `value.recipeTitle`. We're still setting and using recipe's name in `recipeStorage.saveRecipe` though, it's just derived from the name now.

> we might want to indicate that its deprecated in the api/types and check that its actually ok to deprecate from a backend perspective before we deprecate it.

Although we're not fully deprecating it. The only thing changing is that we no longer set it directly from this form, it's rather derived from the recipe title, which the user can now control

ui/desktop/src/recipe/recipeStorage.ts
amed-xyz: This is actually for setting the recipe's name derived from the title

PR #4629, opened 2025-09-12
Desktop json recipes upload

amed-xyz:
Follow up from https://github.com/block/goose/pull/4544 to also allow json recipe uploads

### DIFF
--- a/ui/desktop/src/components/recipes/ImportRecipeForm.tsx
+++ b/ui/desktop/src/components/recipes/ImportRecipeForm.tsx
@@ -356,7 +356,7 @@ export default function ImportRecipeForm({ isOpen, onClose, onSuccess }: ImportR
                                   handleRecipeUploadChange(e.target.files?.[0]);
                                 }}
                                 onBlur={field.handleBlur}
-                                className={`${field.state.meta.errors.length > 0 ? 'border-red-500' : ''} ${
+                                className={`file:pt-1 ${field.state.meta.errors.length > 0 ? 'border-red-500' : ''} ${
                                   isDisabled ? 'cursor-not-allowed' : ''
                                 }`}
                               />


ui/desktop/src/components/recipes/ImportRecipeForm.tsx
```diff
@@ -85,11 +85,24 @@ export default function ImportRecipeForm({ isOpen, onClose, onSuccess }: ImportR
...
+      if (isJsonFile) {
+        parsed = JSON.parse(fileContent);
+      } else {
+        parsed = yaml.parse(fileContent);
+      }
+    } catch (error) {
+      throw new Error(
+        `Failed to parse ${isJsonFile ? 'JSON' : 'YAML'} file: ${error instanceof Error ? error.message : 'Invalid format'}`
+      );
+    }
```
DOsinga, line 102: yaml is a superset of json, so no need to parse them separately. more problematically though, recipes are not always correct yaml before parameter substitution. I don't think the client should do any parsing here really

ui/desktop/src/components/recipes/ImportRecipeForm.tsx
```diff
@@ -85,11 +85,24 @@ export default function ImportRecipeForm({ isOpen, onClose, onSuccess }: ImportR
...
 
     if (!parsed) {
-      throw new Error('YAML file is empty or contains invalid content');
+      throw new Error(`${isJsonFile ? 'JSON' : 'YAML'} file is empty or contains invalid content`);
     }
 
     // Handle both CLI format (flat structure) and Desktop format (nested under 'recipe' key)
```
DOsinga, line 108: remove this comment or update it about legacy; there should be no Desktop format

Issue #4634, opened 2025-09-14
Forcing the model to use absolute paths is counter productive


sings-to-bees-on-wednesdays:
**Please explain the motivation behind the feature request.**
Goose seems to be strongly asking models to use absolute paths when invoking tools.

This doesn't seem to be what other similar software (Claude Code, OpenAI Codex, Gemini CLI) are doing.

In my experience, this also works poorly in practice:

- Google's models seem to heavily penalize repetition, which cause the model to frequently make "typos" when entering long paths. This requires it to retry commands multiple times.

- Claude frequently seems to make mistakes due to being coerced to use absolute paths, see e.g. here: https://github.com/block/goose/pull/4522#issuecomment-3289528277

Additionally, I think giving the absolute path to the model is an unnecessary and easily preventable information leak. The current working context of a tool like goose CLI is implicitly the working directory, and the world outside the working directory (including its location in it) is largely irrelevant. By including the path of the current working directory as part of the prompt, we are automatically (and without a chance for consent) divulging the user's UNIX username, the name of the project they are working on, and possibly more information.

**Describe the solution you'd like**
We should support relative paths everywhere, and stop discouraging models from using them.

**Describe alternatives you've considered**
In general, IMHO we should avoid diverging too much from the prompts, tool APIs and descriptions of the upstream models, and stick as closely as possible to the tools used by upstream providers' implementations (Claude Code, OpenAI Codex, Gemini CLI), as that is what they have been likely trained on.

**Additional context**
I have been evaluating various tools in this category for personal and professional use.

- [x] I have verified this does not duplicate an existing feature request

michaelneale:
thanks @sings-to-bees-on-wednesdays this makes sense to me - @alexhancock's change may help a lot with this.

alexhancock:
@sings-to-bees-on-wednesdays let me know if you see improvement with this if you run off of main. will also be in the next release within a few days.

understood-the-assignment:
Thank you very much for the quick fix! Will try this as soon as my credit refills 😄

Issue #4636, opened 2025-09-14
1.8


benrey:
**when tools are used the path is entered twice**

When goose tries to edit files or use commands from tools dev tools I'm guessing and strreplace is the command but probably happens to others it sends the command and it attaches the location twice so it knows its in the root of the project but when it sends the location relative to the root of the project the tool adds it twice:
/users/username/project/app/components/app/components/filename.ext  and errors out.

**To Reproduce**
1. Use a model that allows tooling
2. Ask the chat to fix a simple problem.
3. It will continue to send the same command over and over and is not intuitive enough to try different commands or try another approach.
The persistent error indicates an incorrect file path being referenced. The correct absolute path should be /Users/username/git/appname/app/components/filename.ext (without the duplicate /app/components). Let's attempt one final correction with the proper path:

**Expected behavior**
Tooling just works and edits files etc.

**Please provide following information:**
 - **OS & Arch:** MAC OSX m4 max pro
 - **Interface:** Desktop
 - **Version:** v1.8.0
 - **Extensions enabled:** memory, developer tools 
 - **Provider & Model:** qwen-coder53b

**Additional context**
Also tried computer controller without developer tools turned on and it fails asa well

alexhancock:
This seems like a model specific issue from my testing just now

I don't repro the same issue with either Claude Sonnet 4 for example, or `qwen2.5-coder`

Which provider are you using, and which exact model are you using? I'm not familiar with a 53b `qwen-coder`

DOsinga:
Possibly related: #4634

alexhancock:
I've merged https://github.com/block/goose/pull/4641 which we expect may make this better. @benrey let me know when you can get back to me with details of what model exactly you were seeing this with.

benrey:
Sorry for the late response. yeah it's just qwen3 coder. I just uninstalled and went back to 1.7 and everything started working again. So yeah I think there is something wrong with tools and I'll try that branch to see if it fixes the issue and report back.

DOsinga:
closing this for now, assuming the fix @alexhancock mentioned works

PR #4641, opened 2025-09-15
fix: #4634 #4636 do not always encourage absolute paths

alexhancock:
## Pull Request Description

* Address #4634 and #4636
* Remove system prompt encouragement of absolute paths
* Support relative paths from the cwd in resolve_path and tools


michaelneale:
only possible downside is that goose often works across many directories on a system, wonder if this makes it confused or in practice is not an issue?

alexhancock:
cool. will merge it. will keep a background workstream going to figure out how to terminal-bench branches of goose.

PR #4648, opened 2025-09-15
Session manager

DOsinga:
## Use SQLite for Session manager

Don't implement your own database


crates/goose/src/session/session_manager.rs
```diff
@@ -0,0 +1,826 @@
...
+
+impl SessionManager {
+    async fn instance() -> Result<Arc<SessionStorage>> {
+        match SESSION_STORAGE.get() {
+            Some(storage) => Ok(storage.clone()),
+            None => {
+                let storage = Arc::new(SessionStorage::new().await?);
+                match SESSION_STORAGE.set(storage.clone()) {
+                    Ok(_) => Ok(storage),
+                    Err(_) => Ok(SESSION_STORAGE.get().unwrap().clone()),
```
jamadeo: I think you do want to check the variants: https://docs.rs/tokio/latest/tokio/sync/enum.SetError.html#variants

crates/goose/src/session/session_manager.rs
```diff
@@ -0,0 +1,826 @@
...
+    String,
+    Option<i32>,
+    Option<i32>,
+    Option<i32>,
+    Option<i32>,
+    Option<i32>,
+    Option<i32>,
+    Option<String>,
+    Option<String>,
+);
```
jamadeo: oof, any better way to define this?

jamadeo: it does look like sqlx will let you use a struct, can we do that?

crates/goose/src/session/session_manager.rs
```diff
@@ -0,0 +1,826 @@
...
+
+impl SessionStorage {
+    async fn new() -> Result<Self> {
+        let session_dir = ensure_session_dir()?;
+        let db_path = session_dir.join("sessions.db");
+
+        let storage = if db_path.exists() {
+            Self::open(&db_path).await?
+        } else {
+            println!("Creating new session database...");
```
jamadeo: probably don't want to introduce printlns here

crates/goose-cli/src/commands/web.rs
```diff
@@ -319,26 +295,27 @@ async fn handle_socket(socket: WebSocket, state: AppState) {
...
                                     let new_session = Arc::new(Mutex::new(existing_messages));
                                     sessions.insert(session_id.clone(), new_session.clone());
                                     new_session
                                 }
                             };
-
                             // Clone sender for async processing
                             let sender_clone = sender.clone();
                             let agent = state.agent.clone();
+                            let session_id2 = session_id.clone();
```
jamadeo: maybe stick with _clone

crates/goose-cli/src/commands/web.rs
```diff
@@ -502,24 +468,11 @@ async fn process_message_streaming(
             while let Some(result) = stream.next().await {
                 match result {
                     Ok(AgentEvent::Message(message)) => {
-                        // Add message to our session
+                        // Add message to our session TODO(Douwe): actually do that
```
jamadeo: seems important!

crates/goose-cli/src/commands/web.rs
```diff
@@ -634,28 +587,7 @@ async fn process_message_streaming(
...
-                            &session_file,
-                            &current_messages,
-                            None, // No provider needed for persisting
-                            working_dir.clone(),
-                        )
-                        .await
-                        {
-                            error!("Failed to persist compacted messages: {}", e);
-                        }
+                        tracing::info!("History replaced, compacting happened in reply");
```
jamadeo, line 537: not in scope here, but have we considered how compaction should work with session history? seems like we should store the both the old and compacted state of things

crates/goose-server/src/routes/agent.rs
```diff
@@ -118,37 +117,29 @@ async fn start_agent(
...
+    if let Some(recipe) = payload.recipe {
+        SessionManager::update_session(&session.id)
+            .recipe_json(serde_json::to_string(&recipe).ok())
+            .apply()
+            .await
+            .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+
+        session = SessionManager::get_session(&session.id, false)
+            .await
+            .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
```
jamadeo, line 118: can we do something other than throw away the error here? maybe at minimum a log

crates/goose/src/conversation/mod.rs
```diff
@@ -116,6 +116,17 @@ impl Conversation {
...
 }
 
+impl PartialEq for Conversation {
+    fn eq(&self, other: &Self) -> bool {
+        self.0.len() == other.0.len()
+            && self
+                .0
+                .iter()
+                .zip(&other.0)
+                .all(|(a, b)| a.role == b.role && a.created == b.created && a.content == b.content)
```
jamadeo: should we not defer to equivalence of the Messages ?

crates/goose-server/src/openapi.rs
```diff
@@ -352,7 +336,7 @@ impl<'__s> ToSchema<'__s> for AnnotatedSchema {
     }
 }
 
-#[allow(dead_code)] // Used by utoipa for OpenAPI generation
```
jamadeo, line 355: eh, I think if you flag on dead_code it's ok to have a comment for why :)

ui/desktop/src/api/types.gen.ts
```diff
@@ -2175,6 +2234,78 @@ export type GetSessionHistoryResponses = {
...
+    401: unknown;
+    /**
+     * Session not found
+     */
+    404: unknown;
+    /**
+     * Internal server error
+     */
+    500: unknown;
+};
```
jamadeo, line 2220: not for your PR, but I do not see the value in enumerating these status codes in the openapi. It'd be one thing if they were enforced somehow, but I've come across a bunch that aren't even true

crates/goose-cli/src/commands/web.rs
DOsinga: I think the idea is that we keep all messages around but annotate them with visibility. Hmm, is this already in /cc @katzdave

katzdave: Yes we have message metadata tagging messages appropriately as agent vs user visible.

I could see benefits of even more metadata tags like 'summary' being relevant.

I'm pushing this domain actively both in the UI for making the post/pre-summary messages more clear, and I have some other ideas around improving the summaries with with keeping around more user messages + messages at the end of tool call chains.

crates/goose/src/session/session_manager.rs
DOsinga: the LLMs said:

async fn instance() -> Result<Arc<SessionStorage>> {
    let storage = Arc::new(SessionStorage::new().await?);
    match SESSION_STORAGE.set(storage) {
        Ok(_) => Ok(SESSION_STORAGE.get().unwrap().clone()),
        Err(_) => Ok(SESSION_STORAGE.get().unwrap().clone()),
    }
}

should do the trick, since failure means we already have the thing

crates/goose/src/session/session_manager.rs
DOsinga: yeah, this is not pretty. I'll get you something better

crates/goose/src/session/session_manager.rs
DOsinga: LLMs...

DOsinga: done

crates/goose-cli/src/commands/web.rs
DOsinga: this is all gone

crates/goose-cli/src/commands/web.rs
DOsinga: yeah, though actually as part of this, the .reply method keeps the conversation updated, so now we don't have to here

ui/desktop/src/api/types.gen.ts
DOsinga: I find these very confusing. all I want to do is return a <error_code, error_message> tuple if things go badly and it seems much harder than it should be

crates/goose-server/src/openapi.rs
DOsinga: true. then again, cargo clippy --jobs 2 -- -D warnings doesn't think it is deadcode, so just removed the thing

crates/goose-cli/src/commands/bench.rs
```diff
@@ -11,28 +11,31 @@ use tokio::sync::Mutex;
...
 #[async_trait]
-impl BenchBaseSession for Session {
+impl BenchBaseSession for CliSession {
     async fn headless(&mut self, message: String) -> anyhow::Result<()> {
         self.headless(message).await
     }
     fn session_file(&self) -> Option<PathBuf> {
-        self.session_file()
+        // Return None since CliSession doesn't have a session file concept
+        None
```
jamadeo: the only use I saw of this is

https://github.com/block/goose/blob/e141e4a5fb87805591aca60ce011ad9bdbe40274/crates/goose-bench/src/runners/eval_runner.rs#L160-L161

which would fail now

crates/goose-cli/src/commands/web.rs
```diff
@@ -169,8 +166,19 @@ pub async fn handle_web(port: u16, host: String, open: bool) -> Result<()> {
     Ok(())
 }
 
-async fn serve_index() -> Html<&'static str> {
-    Html(include_str!("../../static/index.html"))
+async fn serve_index() -> Result<Redirect, (http::StatusCode, String)> {
+    match SessionManager::create_session(
+        std::env::current_dir().unwrap_or_else(|_| std::path::PathBuf::from(".")),
+        "Web session".to_string(),
+    )
```
jamadeo, line 176: suggest something like

```
let session = SessionManager::create_session(..).await.map_err(..)?;
Redirect::permanent(..)
```

crates/goose-cli/src/commands/web.rs
```diff
@@ -715,4 +590,5 @@ async fn process_message_streaming(
 }
 
 // Add webbrowser dependency for opening browser
```
jamadeo: comment is in the wrong place now but also it shouldn't exist

crates/goose-cli/src/session/builder.rs
```diff
@@ -257,68 +240,64 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
...
-                process::exit(1);
-            }
-        }
+        let session = SessionManager::create_session(
+            std::env::current_dir().unwrap(),
+            "CLI Session".to_string(),
+        )
+        .await
+        .unwrap();
+        Some(session.id)
```
jamadeo, line 281: this function should probably not panic or exit(1) but that's too much yak shaving

crates/goose-cli/src/session/mod.rs
```diff
@@ -121,30 +119,37 @@ pub async fn classify_planner_response(
...
-                Conversation::new_unvalidated(Vec::new())
+        let messages = if let Some(session_id) = &session_id {
+            // Use SessionManager to load messages
+            tokio::task::block_in_place(|| {
+                tokio::runtime::Handle::current().block_on(async {
+                    SessionManager::get_session(session_id, true)
+                        .await
+                        .map(|session| session.conversation.unwrap_or_default())
+                        .unwrap_or_else(|e| {
+                            eprintln!("Warning: Failed to load message history: {}", e);
```
jamadeo: I guess this isn't new but seems weird that this would just warn and continue

crates/goose-cli/src/session/mod.rs
```diff
@@ -121,30 +119,37 @@ pub async fn classify_planner_response(
...
         retry_config: Option<RetryConfig>,
     ) -> Self {
-        let messages = if let Some(session_file) = &session_file {
-            session::read_messages(session_file).unwrap_or_else(|e| {
-                eprintln!("Warning: Failed to load message history: {}", e);
-                Conversation::new_unvalidated(Vec::new())
+        let messages = if let Some(session_id) = &session_id {
+            // Use SessionManager to load messages
+            tokio::task::block_in_place(|| {
+                tokio::runtime::Handle::current().block_on(async {
```
jamadeo, line 133: worth it maybe to have a sync version provided by SessionManager itself?

crates/goose/src/session/session_manager.rs
jamadeo: I think that's true for a regular OnceCell, but this is a tokio::OnceCell, which can fail to set if it's being concurrently set 

> If the OnceCell is empty, but some other task is currently trying to set the value, this call will fail with [SetError::InitializingError](https://docs.rs/tokio/latest/tokio/sync/enum.SetError.html#variant.InitializingError).

jamadeo (approved):

crates/goose-server/src/routes/agent.rs
DOsinga: why particularly here? this should never fail I think - we're just refreshing the session after updating it

ui/desktop/src/api/types.gen.ts
jamadeo: You definitely can do that: https://docs.rs/axum/latest/axum/response/index.html it's just that our handlers might not be defined that way. I think it'd be a good idea at some point to change the to 
- drop all of the documented response codes: this makes more sense in a public api, not so much for us
- return tuples like you said everywhere

crates/goose-server/src/routes/agent.rs
jamadeo: maybe we leave it for another error handling PR. I just think returning 500s without error messages isn't great because there's always an extra step when debugging

crates/goose-cli/src/commands/bench.rs
DOsinga: good catch

crates/goose-cli/src/commands/web.rs
DOsinga: hmm, also shouldn't be a permanent redirect I think

crates/goose-cli/src/commands/web.rs
DOsinga: 👍

crates/goose-cli/src/session/builder.rs
DOsinga: this is the cli though so it is not the end of the world (well, for the cli it is) - but I agree since who knows who might thing it is a good idea to call this builder. pre-existing ...

crates/goose-cli/src/session/mod.rs
DOsinga: tbh, I'd rather not (but I think you feel differenly) - goosed needs to be async I think and I don't trust our contributors to stick to that if we have sync versions. better to be awkward and show it

crates/goose-cli/src/session/mod.rs
DOsinga: so now you do want to panic? :)

crates/goose/src/session/session_manager.rs
```diff
@@ -0,0 +1,928 @@
...
+            self.replace_conversation(&session.id, conversation).await?;
+        }
+        Ok(())
+    }
+
+    async fn run_migrations(&self) -> Result<()> {
+        let current_version = self.get_schema_version().await?;
+
+        if current_version < CURRENT_SCHEMA_VERSION {
+            println!(
```
alexhancock: Perhaps we configure the logging utility so all output from these modules goes somewhere like:

```
/Users/alexhancock/.local/state/goose/logs/session-db
```

alongside the others?

crates/goose/src/session/session_manager.rs
```diff
@@ -0,0 +1,928 @@
...
+        let session_dir = ensure_session_dir()?;
+        let db_path = session_dir.join("sessions.db");
+
+        let storage = if db_path.exists() {
+            Self::open(&db_path).await?
+        } else {
+            println!("Creating new session database...");
+            let storage = Self::create(&db_path).await?;
+
+            if let Err(e) = storage.import_legacy(&session_dir).await {
```
alexhancock, line 345: Would be good to test this with a few real session dirs from the wild to ensure no issues before launch

crates/goose/src/session/session_manager.rs
```diff
@@ -0,0 +1,928 @@
...
+
+        let max_idx = sqlx::query_scalar::<_, Option<i32>>(
+            "SELECT MAX(CAST(SUBSTR(id, 10) AS INTEGER)) FROM sessions WHERE id LIKE ?",
+        )
+        .bind(format!("{}_%", today))
+        .fetch_one(&mut *tx)
+        .await?
+        .unwrap_or(0);
+
+        let session_id = format!("{}_{}", today, max_idx + 1);
```
alexhancock, line 171: Do we ever use the fact that the ids have these dates in them? If not, could just be an autoincrementing int in sqlite.

crates/goose/src/session/legacy.rs
```diff
@@ -0,0 +1,113 @@
...
+        return Err(anyhow::anyhow!("Session file too large"));
+    }
+    if file_metadata.len() == 0 {
+        return Err(anyhow::anyhow!("Empty session file"));
+    }
+
+    let modified_time = file_metadata.modified().unwrap_or(SystemTime::now());
+    let created_time = file_metadata
+        .created()
+        .unwrap_or_else(|_| parse_session_timestamp(session_name).unwrap_or(modified_time));
```
alexhancock, line 51: were the old session names human readable title strings? or is this a different field from the metadata and called `session_name` in this code

ui/desktop/src/components/sessions/SessionListView.tsx
```diff
@@ -356,9 +360,7 @@ const SessionListView: React.FC<SessionListViewProps> = React.memo(
       // Update state immediately for optimistic UI
       setSessions((prevSessions) =>
         prevSessions.map((s) =>
-          s.id === sessionId
-            ? { ...s, metadata: { ...s.metadata, description: newDescription } }
-            : s
+          s.id === sessionId ? { ...s, metadata: { ...s, description: newDescription } } : s
```
alexhancock: do we need all the props at both root and `metadata` of this object? it seems like how we've flattened out the data model a lot this object could become flattened as well

crates/goose/src/session/session_manager.rs
DOsinga: replaced the println!

crates/goose/src/session/session_manager.rs
DOsinga: I imported 8000 from my folder...

crates/goose/src/session/legacy.rs
DOsinga: nah, we just called it that. they were date strings. description is the human readable version

ui/desktop/src/components/sessions/SessionListView.tsx
DOsinga: oops. there is no more metadata, this is just a remnant that will set this useless property that we then don't display until we get confirmation from the server. nice catch!

DOsinga:
> Looks like some conflicts but once resolved 🚢

yeah, with this type of thing, I'd rather let the conflicts vester than keep updating them

crates/goose-cli/src/session/mod.rs
```diff
@@ -121,30 +118,33 @@ pub async fn classify_planner_response(
...
-                eprintln!("Warning: Failed to load message history: {}", e);
-                Conversation::new_unvalidated(Vec::new())
+        let messages = if let Some(session_id) = &session_id {
+            tokio::task::block_in_place(|| {
+                tokio::runtime::Handle::current().block_on(async {
+                    SessionManager::get_session(session_id, true)
+                        .await
+                        .map(|session| session.conversation.unwrap_or_default())
+                        .unwrap()
+                })
```
r0x0d, line 138: Hi, @DOsinga! Since this got merged, I started seeing this error: 

```
thread 'main' panicked at crates/goose-cli/src/session/mod.rs:136:26:
called `Result::unwrap()` on an `Err` value: Session not found
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

r0x0d: Not sure if a good patch, but this fixed the issue for me locally: https://github.com/block/goose/pull/5055

PR #4649, opened 2025-09-16
Add Synthetic (https://synthetic.new) as a provider

billycao:
## Pull Request Description

Thanks for building goose, it looks like a neat LLM frontend!

I'm the cofounder of [Synthetic](https://synthetic.new). A few of our users wanted to try to use goose but reported having issues setting up Synthetic as a custom provider. I also wasn't able to get it to work, so proposing this PR to add us as a native provider.

Tested with `cargo check` and verified running locally, both via CLI and MacOS UI:

CleanShot 2025-09-15 at 17 13 50@2x
CleanShot 2025-09-15 at 17 20 47@2x

CleanShot 2025-09-15 at 18 07 05@2x
CleanShot 2025-09-15 at 18 07 50@2x
CleanShot 2025-09-15 at 18 08 12@2x

### DIFF
--- a/crates/goose/src/providers/factory.rs
+++ b/crates/goose/src/providers/factory.rs
@@ -21,6 +21,7 @@ use super::{
     provider_registry::ProviderRegistry,
     sagemaker_tgi::SageMakerTgiProvider,
     snowflake::SnowflakeProvider,
+    synthetic::SyntheticProvider,
     tetrate::TetrateProvider,
     venice::VeniceProvider,
     xai::XaiProvider,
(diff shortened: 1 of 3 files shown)


michaelneale:
if you run cargo fmt and/or ask goose to check it - should pass those checks

billycao:
Ty for review! Ran `cargo fmt`, PTAL, thanks! :)

michaelneale:
paging @DOsinga is this a reasonable way to do a custom provider?

michaelneale:
looking good so far

billycao:
@michaelneale @DOsinga Sorry to bother, but wondering if this would be reasonable to merge/wanted to avoid too much code drift. :) Thanks a lot!

DOsinga:
no worries, sorry about not replying earlier. why can't this work as a custom provider? it does look like the provider code itself is basically a copy of openai. what's the difference?

billycao:
Hey there!

It's a good question, I'm not actually sure but I downloaded the latest version and I still can't seem to get the custom provider to work with any combination of base URL, trying

- `https://api.synthetic.new/`
- `https://api.synthetic.new/openai`
- `https://api.synthetic.new/openai/`
- `https://api.synthetic.new/openai/v1`
- `https://api.synthetic.new/openai/v1/`

I should probably be a good citizen and investigate why the custom provider doesn't work but don't currently have the cycles and was selfishly hoping we could be a native provider. 😅 

I understand not wanting the UI to be cluttered with a bunch of miscellaneous small providers though. Feel free to close this PR and if our users mention Goose issues again I can try to investigate deeper in the future. :)

Ty for your time!

Screenshot 2025-10-05 at 10 09 34 PM
Screenshot 2025-10-05 at 10 09 22 PM

DOsinga:
hi - yeah, I think the right way to do this is custom providers and if that doesn't work because there is a slight incompatibolity in the openai provider, fix that. 

we're looking to make it possible to have custom providers that don't show up as such, i.e. they are part of the distribution, just not implemented as code.

PR #4650, opened 2025-09-16
Fix #4606: Add stderr logging support for stdio extensions

### LINKED ISSUE #4606: stdio Extensions stdout/stderr
When developing MCP servers, it is very useful to see the stdio/stderr output from the server itself.  Please add a way to specify file paths where the stdout and stderr will be sent.  Ideally, the configure extension dialog will provide text boxes to specify paths for each of those.  Being able to send both stdout and stderr to the same file would be helpful as well.

Running on Mac OSX using goose version `1.7.0`, I tried adding a redirect to the extension command like this, but the `mcp.log` file was not written to:
```
uv --directory /Users/foo/my-mcp run src/my_mcp_server/server.py > /tmp/my-mcp/mcp.log 2>&1
```

I tried setting the permissions on `/tmp/my-mcp` to `777` and also set the permissions on `/tmp/my-mcp/mcp.log` to `666`

jalateras:
Fixes #4606

## Summary
This PR adds the ability to redirect stderr output from MCP stdio extensions to log files for debugging purposes. This addresses the developer need to see debug output from MCP servers during development.

## What Changed
- Added optional `stdout_log_path` and `stderr_log_path` fields to the stdio extension configuration
- Implemented file-based stderr redirection in the extension manager (stdout reserved for future implementation)
- Updated the UI to allow users to specify a stderr log path in the extension configuration modal
- Updated all existing code to handle the new optional fields

## Technical Details
- **Stderr Logging**: When a stderr log path is specified, the stderr output from the MCP server process is redirected to the specified file (created with append mode)
- **Stdout Handling**: Currently stdout cannot be redirected as it's used for the MCP protocol communication. The field exists for future enhancement when we implement a tee-like mechanism
- **File Safety**: Log directories are created automatically if they don't exist, and proper error handling is in place for file operations
- **Backward Compatibility**: All new fields are optional, ensuring existing configurations continue to work

## Test Plan
1. Configure a stdio extension with a stderr log path
2. Verify stderr output is written to the specified file
3. Verify MCP protocol communication still works correctly
4. Test with empty/missing log paths to ensure backward compatibility

## Screenshots
The UI now includes an optional "Stderr Log Path" field in the extension configuration modal for stdio extensions.

## Notes
- The stdout logging capability is not yet functional as stdout is used for MCP protocol communication
- Future work could implement a tee-like mechanism to support both protocol communication and logging for stdout


jalateras:
##  Test with an actual MCP server:
  
1.  Install a real MCP server
  ```npm install -g @modelcontextprotocol/server-filesystem```

2. Configure in Goose with:
 ``` 
   Command: npx @modelcontextprotocol/server-filesystem /tmp
   Stderr Log Path: /tmp/filesystem-mcp.log
  ```
3. Use the extension and check logs
  ```tail -f /tmp/filesystem-mcp.log```

michaelneale:
this seems a massive change for that - is there a way to run a stdio proxy or similar to direct logs when debugging?

DOsinga:
I like the idea, but like @michaelneale says, it feels like a massive change that can easily confuse users not aware of this. 

more over, it doesn't actually do what it says. you have all this plumbing and then at the crucial bit it just says, well, actually, stdio is the protocol we use to talk to the MCP server so we can't redirect that and then the the path is unused? also that bit of code is just copying the code.

if you wanted to do this, just `child_process_client` only. right now it already captures stderr, but only for during the setup. not during execution. if you would just add a parser to the cmd that looks for 2> and if there is something after that, start logging to that after init, we can just tell any one who wants to debug to add that to the command they add. it'll be a minimal, local change. my goose says something like:

```rust
async fn child_process_client(
    mut command: Command,
    timeout: &Option<u64>,
) -> ExtensionResult<McpClient> {
    #[cfg(unix)]
    command.process_group(0);
    #[cfg(windows)]
    command.creation_flags(CREATE_NO_WINDOW_FLAG);
    
    let stderr_redirect = extract_stderr_redirect(&mut command);
    
    let (transport, mut stderr) = TokioChildProcess::builder(command)
        .stderr(Stdio::piped())
        .spawn()?;
    let mut stderr = stderr.take().ok_or_else(|| {
        ExtensionError::SetupError("failed to attach child process stderr".to_owned())
    })?;

    if let Some(path) = stderr_redirect {
        tokio::spawn(async move {
            if let Ok(mut file) = tokio::fs::File::create(path).await {
                let _ = tokio::io::copy(&mut stderr, &mut file).await;
            }
        });
    } else {
        tokio::spawn(async move {
            let _ = tokio::io::copy(&mut stderr, &mut tokio::io::sink()).await;
        });
    }

    McpClient::connect(
        transport,
        Duration::from_secs(timeout.unwrap_or(crate::config::DEFAULT_EXTENSION_TIMEOUT)),
    )
    .await
    .map_err(Into::into)
}

fn extract_stderr_redirect(command: &mut Command) -> Option<PathBuf> {
    let program = command.get_program().to_str()?;
    let args: Vec<String> = command.get_args().filter_map(|a| a.to_str()).map(String::from).collect();
    
    let mut new_args = Vec::new();
    let mut stderr_file = None;
    let mut i = 0;
    
    while i < args.len() {
        if args[i] == "2>" && i + 1 < args.len() {
            stderr_file = Some(PathBuf::from(&args[i + 1]));
            i += 2;
        } else if let Some(path) = args[i].strip_prefix("2>") {
            stderr_file = Some(PathBuf::from(path));
            i += 1;
        } else {
            new_args.push(args[i].clone());
            i += 1;
        }
    }
    
    *command = Command::new(program);
    for arg in new_args {
        command.arg(arg);
    }
    
    stderr_file
}
```

DOsinga:
I'm going to close this. see the suggested changes, feel free to open another one!

PR #4651, opened 2025-09-16
Feat: Add prompt injection detection settings UI + update security logging

dorien-koelemeijer:
## Pull Request Description

TBD


crates/goose-server/src/routes/reply.rs
```diff
@@ -476,6 +476,34 @@ pub async fn confirm_permission(
...
 
+    // Log all permission confirmations for debugging
+    tracing::info!(
+        "Permission confirmation received - ID: {}, action: {}, principal_type: {:?}",
+        request.id,
+        request.action,
+        request.principal_type
+    );
+
+    // Log user security decision if this is for a security alert
```
DOsinga: this seems to be the inverse of 472

crates/goose/src/agents/agent.rs
```diff
@@ -178,7 +178,7 @@ impl Agent {
             tool_route_manager: ToolRouteManager::new(),
             scheduler_service: Mutex::new(None),
             retry_manager: RetryManager::new(),
-            tool_inspection_manager: Self::create_default_tool_inspection_manager(),
+            tool_inspection_manager: Mutex::new(Self::create_default_tool_inspection_manager()),
```
DOsinga: why do we need a mutext?

ui/desktop/src/components/settings/security/SecurityToggle.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const { config, upsert } = useConfig();
+
+  // Initialize settings with defaults
+  const [settings, setSettings] = useState<SecuritySettings>({
+    enabled: false,
+    threshold: 0.7,
+  });
+
+  useEffect(() => {
+    // Load security settings from config when config changes
```
DOsinga: does it?

DOsinga (commented):
I had a quick look, but I fear this feels rather vibe codey - there's some duplication, unrelated changes, I wonder if you could do a bit of a review more yourself?

dorien-koelemeijer:
> I had a quick look, but I fear this feels rather vibe codey - there's some duplication, unrelated changes, I wonder if you could do a bit of a review more yourself?

You're right, I wanted to get it done too quickly. Having another look now and will do some cleanup. Thanks for the comments

crates/goose-server/src/routes/reply.rs
```diff
@@ -274,21 +274,6 @@ async fn reply_handler(
         };
 
         let mut all_messages = messages.clone();
-        let session_path = match session::get_path(session::Identifier::Name(session_id.clone())) {
```
michaelneale, line 284: not sure how this is a related change?

ui/desktop/src/components/settings/security/SecuritySection.tsx
```diff
@@ -0,0 +1,5 @@
+import { SecurityToggle } from './SecurityToggle';
```
michaelneale: this seems unnecessary indirection this file - could just be security Toggle directly? (unless I am missing some established react pattern, there can be weird patterns and conventions!)

dorien-koelemeijer: You're right, that's way better!

michaelneale:
looks like linting issues now

michaelneale:
so the config is now saved as: 

```
security_threshold: 0.2
security_enabled: true
```

not in a security: section - is that expected? (also I noted it didnt' save the threshold until I changed it, which I assume means default?)

@dorien-koelemeijer there are a few other changes not GUI related, what is best way to check this is still healthy?

looks ok otherwise, @DOsinga happier with shape of code?

michaelneale:
thread safety/oncelock stuff looks good too

dorien-koelemeijer:
> so the config is now saved as:
> 
> ```
> security_threshold: 0.2
> security_enabled: true
> ```
> 
> not in a security: section - is that expected? (also I noted it didnt' save the threshold until I changed it, which I assume means default?)
> 
> @dorien-koelemeijer there are a few other changes not GUI related, what is best way to check this is still healthy?
> 
> looks ok otherwise, @DOsinga happier with shape of code?

It'll be security.threshold: xx and security.enabled: true/false. But I think most people will just go through the UI settings now, since it's a lot easier. 

Re: non-GUI related changes, you can simply test the changes in goose CLI/desktop version locally. I've tested until the last lint commit and everything was still fine - will do some final testing in the next couple of hours, but it should all be good.

michaelneale:
@dorien-koelemeijer but when I saw it in the config.yaml - it wasn't grouped under security (was an underscore) - just wanted to know if that was expected?

dorien-koelemeijer:
> @dorien-koelemeijer but when I saw it in the config.yaml - it wasn't grouped under security (was an underscore) - just wanted to know if that was expected?

Really? Let me have a look as well 👀 In any case, I'll have to update instructions once this is merged. Would be great to get both this PR and the other one together in one release to prevent confusion

dorien-koelemeijer:
> > @dorien-koelemeijer but when I saw it in the config.yaml - it wasn't grouped under security (was an underscore) - just wanted to know if that was expected?
> 
> Really? Let me have a look as well 👀 In any case, I'll have to update instructions once this is merged. Would be great to get both this PR and the other one together in one release to prevent confusion

I think I tried to keep things backwards compatible first and still had that config saved - you're right, it's definitely with underscores

Screenshot 2025-09-24 at 11 42 44 am

Seems like it would make sense to keep underscores, since all config is saved like that? I guess it depends on whether this PR gets released at the same time as the other one, then we don't have to have things be backwards compatible. Honestly, probably easiest to keep it in line with all other config though and use underscores?

dorien-koelemeijer:
> > so the config is now saved as:
> > ```
> > security_threshold: 0.2
> > security_enabled: true
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > not in a security: section - is that expected? (also I noted it didnt' save the threshold until I changed it, which I assume means default?)
> > @dorien-koelemeijer there are a few other changes not GUI related, what is best way to check this is still healthy?
> > looks ok otherwise, @DOsinga happier with shape of code?
> 
> It'll be security.threshold: xx and security.enabled: true/false. But I think most people will just go through the UI settings now, since it's a lot easier.
> 
> Re: non-GUI related changes, you can simply test the changes in goose CLI/desktop version locally. I've tested until the last lint commit and everything was still fine - will do some final testing in the next couple of hours, but it should all be good.

Have tested both desktop and CLI version - all seems fine still. Testing will be the same as for the other PR if you wanted to give it a go as well to be safe

crates/goose/src/security/security_inspector.rs
```diff
@@ -86,14 +84,10 @@ impl ToolInspector for SecurityInspector {
     }
 
     fn is_enabled(&self) -> bool {
```
DOsinga, line 86: you have this twice

ui/desktop/src/components/settings/security/SecurityToggle.tsx
```diff
@@ -0,0 +1,103 @@
...
+  };
+
+  const handleThresholdChange = async (threshold: number) => {
+    const validThreshold = Math.max(0, Math.min(1, threshold));
+    try {
+      await upsert('security_threshold', validThreshold, false);
+    } catch (error) {
+      console.error('Failed to update threshold:', error);
+    }
+  };
```
DOsinga, line 31: here and above, don't catch the error if all you do is log it, which is what happens by default

ui/desktop/src/components/settings/security/SecurityToggle.tsx
```diff
@@ -0,0 +1,103 @@
...
+              Higher values are more strict (0.01 = very lenient, 1.0 = maximum strict)
+            </p>
+            <input
+              type="number"
+              min={0.01}
+              max={1.0}
+              step={0.01}
+              value={thresholdInput}
+              onChange={(e) => {
+                // Update local input state immediately for responsive typing
```
DOsinga: delete

ui/desktop/src/components/settings/security/SecurityToggle.tsx
```diff
@@ -0,0 +1,103 @@
...
+                setThresholdInput(e.target.value);
+              }}
+              onBlur={(e) => {
+                // Validate and save to config on blur
+                let value = parseFloat(e.target.value);
+                if (isNaN(value) || value < 0.01) {
+                  value = 0.01;
+                } else if (value > 1.0) {
+                  value = 1.0;
+                }
```
DOsinga, line 78: dont you want to set this back to the value it had before? or show an error and not write it to the server

PR #4656, opened 2025-09-16
fix(compaction): try to catch more context limit exceeded errors and compact

alexhancock:
Ran into a context limit exceeded error while using an anthropic model that was not caught by the provider specific error parsing in `crates/goose/src/providers/anthropic.rs`

Considering adding this as a catchall at the top level.

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -952,6 +952,52 @@ impl Agent {
         Ok(None)
     }
 
+    /// Create events for successful compaction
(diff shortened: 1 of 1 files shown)


michaelneale:
nice - yeah @katzdave may have deeper thoughts but I am sure I have hit this with anthroic specifically.

michaelneale (approved):
I think ok - but ideally @katzdave can give final ok

alexhancock:
Can roll it back out if it causes any trouble or @katzdave has preference otherwise. There may also be some cleanup we can do in providers now which I can look at in followups.

PR #4664, opened 2025-09-17
chore(mcp): convert autovisualiser to use the rust sdk for mcp

alexhancock:
* Server one of the change originally in https://github.com/block/goose/pull/4488
* Converts `autovisualiser` to only use the rust sdk for mcp, and not our internal `mcp-server` crate
* Verified all the input schemas and tool descriptions are the exact same as before, and it works well locally


michaelneale (approved):
nice, works for me.

PR #4667, opened 2025-09-17
More robust test finder patch handling

tlongwell-block:


michaelneale (approved):
nice, makes it draft which is should.

PR #4669, opened 2025-09-17
fix client initialization when using an external backend

nacx:
## Pull Request Description

Fixes an issue where the UI failed to start when using an external backend due to the client not being properly initialised.

Kudos to @tlongwell-block for the [help on discord](https://discord.com/channels/1287729918100246654/1417899132286271721) and the UI patch.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -196,7 +196,7 @@ run-docs:
 # Run server
 run-server:
     @echo "Running server..."
-    cargo run -p goose-server
+    cargo run -p goose-server --bin goosed agent
 
 # Check if OpenAPI schema is up-to-date
 check-openapi-schema: generate-openapi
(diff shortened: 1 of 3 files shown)


Justfile
```diff
@@ -196,7 +196,7 @@ run-docs:
 # Run server
 run-server:
     @echo "Running server..."
-    cargo run -p goose-server
+    cargo run -p goose-server --bin goosed agent
```
nacx, line 199: Without this it failed with:
```
Running server...
cargo run -p goose-server
error: `cargo run` could not determine which binary to run. Use the `--bin` option to specify a binary, or the `default-run` manifest key.
available binaries: generate_schema, goosed
```

michaelneale: ah - I guess it wasn't used regularly, vs just run-ui?

nacx: Probably yes? I'm just getting started with the codebase, so I can't tell about past usage :)

michaelneale: yep makes sense.

michaelneale (approved):
LGTM - some older code paths which hadn't been updated

PR #4675, opened 2025-09-18
Fix debug UI connection to external goosed server

tlongwell-block:
## Problem
The UI fails to connect when using an external goosed backend for debugging, timing out after 20 seconds even though the server is running.

## Root Cause
The checkServerStatus() call happens before the API client is configured with authentication headers.

## Solution
Configure the client with auth headers before checking status in connectToExternalBackend().

## Testing
Start goosed: GOOSE_SERVER__SECRET_KEY=test GOOSE_PORT=3000 ./target/release/goosed agent
Connect UI: just debug-ui

### DIFF
--- a/ui/desktop/src/goosed.ts
+++ b/ui/desktop/src/goosed.ts
@@ -47,9 +47,19 @@ const checkServerStatus = async (): Promise<boolean> => {
 
 const connectToExternalBackend = async (
   workingDir: string,
-  port: number = 3000
+  port: number = 3000,
+  serverSecret: string
 ): Promise<[number, string, ChildProcess]> => {
   log.info(`Using external goosed backend on port ${port}`);
+
+  // Configure client for external backend before checking status
+  client.setConfig({
+    baseUrl: `http://127.0.0.1:${port}`,
+    headers: {
+      'Content-Type': 'application/json',
+      'X-Secret-Key': serverSecret,
+    },
+  });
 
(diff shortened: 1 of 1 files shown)


nacx:
Hi! I opened https://github.com/block/goose/pull/4669 yesterday with the patch. I don't care what PR is merged (the code is yours anyway, as I mentioned in the PR description 😄 ). If you prefer to close that one and merge this, would you mind incorporating here the other changes to the doc and the Justfile I added there?

ui/desktop/src/goosed.ts
```diff
@@ -47,10 +47,20 @@ const checkServerStatus = async (): Promise<boolean> => {
...
+    baseUrl: `http://127.0.0.1:${port}`,
+    headers: {
+      'Content-Type': 'application/json',
+      'X-Secret-Key': serverSecret,
+    },
+  });
+
   const isReady = await checkServerStatus();
   if (!isReady) {
     throw new Error(`External goosed server not accessible on port ${port}`);
```
DOsinga, line 66: instead of duplicating this code, we can jsut delete the isReady check since that happens after we call this function anyway

DOsinga (approved):

zanesq (approved):

tlongwell-block:
Closing this in favor of https://github.com/block/goose/pull/4669/files since it has the nice doc update and an identical bug fix #4669

PR #4684, opened 2025-09-19
feat: AgentManager - foundation for unified execution (#4389)

tlongwell-block:
# Add per-session agent isolation

## What Changed
Replaces the single shared Agent with per-session agents managed by `AgentManager`. Each session now gets its own isolated Agent instance with independent state.

## Why
The shared Agent caused session interference - extensions, providers, and context would bleed between sessions. This made multi-tab support impossible and created confusing user experiences.

## Implementation
- **AgentManager**: LRU cache of session → Agent mappings (default: 100 sessions)
- **Session IDs**: Timestamp-based for easy debugging (`yyyymmdd_hhmmss` format)  
- **Routes**: All endpoints now require session_id and use session-specific agents
- **UI**: Updated to pass session_id with all API calls

## Key Files
- `crates/goose/src/execution/` - New AgentManager implementation
- `crates/goose-server/src/routes/` - Updated route handlers
- `ui/desktop/` - Session ID threading through components
- 12 tests covering core scenarios

## Notes
- Breaking change: session_id now required (UI updated to match)
- `SessionExecutionMode` enum included for future unified execution (#4389)
- Timestamp IDs intentional - helps with debugging and support

Resolves session isolation issues. Foundation for #4389.


tlongwell-block:
Extensions were being added to a new agent session in the Goose desktop app rather than the existing one. The UI now properly passes the session ID when managing extensions, ensuring they're added to the active chat agent rather than creating a new session. This resolves the problem where extensions would appear to activate but wouldn't be available in the current chat.

Changes touched the extension management flow in the TypeScript UI, updating components and API calls to include session_id in requests to the backend.

DOsinga (commented):
Thanks for doing this. This must have been a lot of work. Then again it is also a lot of code. It might have been better had we talked more about this on how we can land this in smaller PRs. It also looks like what we have here doesn't actually deliver any new functionality, right? FWIW I thought we talked in that meeting about approaching this from the other side. First make all the calls session aware that need it (which we mostly have, but I see here some more we didn't cover). Then make the client use only one goosed (which would then switch between agents). And then we can change goosed to support multiple agents

crates/goose-server/src/routes/agent.rs
```diff
@@ -53,28 +51,24 @@ pub struct AddSubRecipesResponse {
...
-    session_id: String,
+    session_id: Option<String>,
 }
 
 #[derive(Deserialize, utoipa::ToSchema)]
 pub struct GetToolsQuery {
     extension_name: Option<String>,
-    #[allow(dead_code)]
-    session_id: String,
+    session_id: Option<String>,
```
DOsinga: why are these optional?

crates/goose-server/src/routes/agent.rs
```diff
@@ -247,7 +251,13 @@ async fn get_tools(
 ) -> Result<Json<Vec<ToolInfo>>, StatusCode> {
     let config = Config::global();
     let goose_mode = config.get_param("GOOSE_MODE").unwrap_or("auto".to_string());
-    let agent = state.get_agent().await;
+    let agent = match state.get_session_agent(query.session_id.clone()).await {
```
DOsinga: I don't think you can use the global goose_mode anymore if you have multiple agents

crates/goose/src/config/base.rs
```diff
@@ -1256,10 +1256,10 @@ mod tests {
         assert_eq!(value, Value::Number((-123).into()));
 
         // Test floats
-        let value = Config::parse_env_value("3.14")?;
+        let value = Config::parse_env_value("3.15")?;
         assert!(matches!(value, Value::Number(_)));
         if let Value::Number(n) = value {
-            assert_eq!(n.as_f64().unwrap(), 3.14);
+            assert!((n.as_f64().unwrap() - 3.15).abs() < f64::EPSILON);
```
DOsinga: what happened here?

crates/goose/src/execution/manager.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    /// Maximum number of concurrent sessions
+    max_sessions: usize,
+
+    /// Default provider configuration (optional)
+    default_provider: Arc<RwLock<Option<Arc<dyn crate::providers::base::Provider>>>>,
+}
+
+/// Metadata about an active session
+struct SessionData {
```
DOsinga: we already have 2 different session data classes, do we really need a third one?

crates/goose/src/execution/manager.rs
```diff
@@ -0,0 +1,263 @@
...
+use tokio::sync::RwLock;
+use tracing::{debug, info, warn};
+
+/// Manages agents with session awareness and isolation
+pub struct AgentManager {
+    /// Active sessions mapped to their agents and metadata
+    sessions: Arc<RwLock<HashMap<SessionId, SessionData>>>,
+
+    /// Shared scheduler for background tasks (optional)
+    scheduler: Arc<RwLock<Option<Arc<dyn SchedulerTrait>>>>,
```
DOsinga: I don't think this should be optional. See the other comment: we might be better of just passing the scheduler to the getagent

crates/goose/src/execution/manager.rs
```diff
@@ -0,0 +1,263 @@
...
+use crate::scheduler_trait::SchedulerTrait;
+use anyhow::Result;
+use chrono::{DateTime, Utc};
+use std::collections::HashMap;
+use std::sync::Arc;
+use tokio::sync::RwLock;
+use tracing::{debug, info, warn};
+
+/// Manages agents with session awareness and isolation
+pub struct AgentManager {
```
DOsinga, line 18: This seems to mostly implement an LRU cache, maybe use an existing rust implementation?

tlongwell-block:
> Thanks for doing this. This must have been a lot of work. Then again it is also a lot of code. It might have been better had we talked more about this on how we can land this in smaller PRs. It also looks like what we have here doesn't actually deliver any new functionality, right? FWIW I thought we talked in that meeting about approaching this from the other side. First make all the calls session aware that need it (which we mostly have, but I see here some more we didn't cover). Then make the client use only one goosed (which would then switch between agents). And then we can change goosed to support multiple agents

Hey, @DOsinga ! Happy to take a crack at it. This is the second pass at it and is much nicer.

It's not as much functional code as it looks like at first. It's about 40% tests by LOC

This does *not* add any additional functionality.

This change enables goosed to support multiple simultaneous sessions, each with its own agent.

This enables tabs in the electron UI in a subsequent PR. It also enables recipes, subagents, and scheduler execution in full Agents/Sessions in subsequent PRs.

When we discussed this, we determined we should, in the initial PR,
* create Agent Manager/Agent Factory
* move ONE execution path to it in the initial PR to touch as little code as possible
* move the other execution paths to Agent Manager in subsequent PRs

I think this is really as minimal a PR as we can do, since almost everything in this PR is just creating the actual Agent Manager and making goosed support it.

tlongwell-block:
Removed stubs and trivial comments, added helper functions. Reduced PR size significantly.

tlongwell-block:
A bit more work trimming this PR down. At ~581 new lines (excluding tests) for this feature now, @DOsinga 

We could theoretically remove the UI changes here, but it breaks extension loading if we do so. We'd need to merge it in a subsequent PR right away if we did.

crates/goose-server/src/commands/agent.rs
```diff
@@ -44,8 +38,7 @@ pub async fn run() -> Result<()> {
     let scheduler_instance = SchedulerFactory::create(schedule_file_path).await?;
     app_state.set_scheduler(scheduler_instance.clone()).await;
 
-    // NEW: Provide scheduler access to the agent
```
DOsinga, line 76: NEW?? thanks for deleting that

crates/goose-server/src/routes/agent.rs
```diff
@@ -25,10 +25,32 @@ use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use tracing::error;
 
+/// Helper for routes that return StatusCode on error
+pub(crate) async fn get_agent_or_500(
```
DOsinga: I don't think you should need this helper. this should never fail, so just going with ? should be good enough. if we do need these helper functions they should not live here and be imported by other routes

crates/goose-server/src/routes/audio.rs
```diff
@@ -413,12 +413,15 @@ mod tests {
             .unwrap();
 
         let response = app.oneshot(request).await.unwrap();
-        assert_eq!(response.status(), StatusCode::UNAUTHORIZED);
+        assert!(
+            response.status() == StatusCode::PRECONDITION_FAILED
+                || response.status() == StatusCode::UNAUTHORIZED
```
DOsinga, line 418: hmm, why this?

crates/goose-server/src/routes/context.rs
```diff
@@ -13,6 +14,8 @@ pub struct ContextManageRequest {
     pub messages: Vec<Message>,
     /// Operation to perform: "truncation" or "summarize"
     pub manage_action: String,
+    /// Optional session ID for session-specific agent
+    pub session_id: Option<String>,
```
DOsinga: here and else where, this should not be optional

crates/goose-server/src/routes/extension.rs
```diff
@@ -107,6 +108,18 @@ async fn add_extension(
         serde_json::to_string_pretty(&raw.0).unwrap()
     );
 
+    let session_id = raw
```
DOsinga: this is all a bit ugh as I am sure you realized; can we instead just make this thing be called with a non raw type here? ideally the actual ExtensionConfigRequest be part of it, but if that fails, just a simple thing with a session_id and a payload

crates/goose-server/src/routes/reply.rs
```diff
@@ -178,9 +179,13 @@ async fn reply_handler(
...
     );
 
-    if let (Some(recipe_name), Some(session_id)) =
-        (request.recipe_name.clone(), request.session_id.clone())
-    {
+    // Generate session_id if not provided (backward compatibility)
+    let session_id = request
+        .session_id
+        .clone()
+        .unwrap_or_else(|| uuid::Uuid::new_v4().to_string());
```
DOsinga: In which case do we not have a session id here?

crates/goose-server/src/state.rs
```diff
@@ -20,21 +19,19 @@ pub struct AppState {
...
         })
     }
 
-    pub async fn get_agent(&self) -> AgentRef {
-        self.agent.read().await.clone()
-    }
-
     pub async fn set_scheduler(&self, sched: Arc<dyn SchedulerTrait>) {
+        self.agent_manager.set_scheduler(sched.clone()).await;
         let mut guard = self.scheduler.write().await;
```
DOsinga: I realize this is all a bit of a pre-existing mess, but this here doesn't feel right; we now end up with two schedulers, one in the agent_manager and one in the state.  Ideally we'd drop the one on agent I would say since the scheduler itself doesn't use that one and have the schedule tool do it. I know they are pointing to the same thing, but it still feels icky

crates/goose-server/src/state.rs
```diff
@@ -66,4 +58,17 @@ impl AppState {
...
+        session_id: Option<String>,
+    ) -> Result<Arc<goose::agents::Agent>, anyhow::Error> {
+        let session_id = session_id
+            .map(SessionId::from)
+            .unwrap_or_else(SessionId::generate);
+
+        self.agent_manager
+            .get_agent(session_id, ExecutionMode::Interactive)
+            .await
+    }
```
DOsinga, line 57: I think we should just name this get_agent. session_id should not be optional. maybe move the _or_json here too? as noted, I think or_500 you can just achieve with ?

previously we were actually sending a FAILED_DEPENDENCY

crates/goose-server/tests/pricing_api_test.rs
```diff
@@ -7,8 +7,7 @@ use std::sync::Arc;
 use tower::ServiceExt;
 
 async fn create_test_app() -> Router {
-    let agent = Arc::new(goose::agents::Agent::default());
-    let state = goose_server::AppState::new(agent);
+    let state = goose_server::AppState::new();
 
     // Add scheduler setup like in the existing tests
```
DOsinga: this duplication here suggests we should just create the scheduler in the app state and simplify things?

crates/goose/src/execution/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+//! This module provides centralized agent lifecycle management with session isolation,
+//! enabling multiple concurrent sessions with independent agents, extensions, and providers.
+
+pub mod manager;
+
+use serde::{Deserialize, Serialize};
+use std::fmt;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum ExecutionMode {
```
DOsinga: this is a dupe

crates/goose/src/execution/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+            Self::Background => write!(f, "background"),
+            Self::SubTask { parent_session } => write!(f, "subtask(parent: {})", parent_session),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
+pub struct SessionId(pub String);
+
+impl SessionId {
```
DOsinga: I don't think we need this class - we already have session ids in the session manager which are just strings (and then session identifiers elsewhere which are either path or strings, but let's fix that some other time). the only thing I think this gives over using strings is that we have a generate, but that one is different from the one session manager and also we should not need to call it (since everything should have a session id) and also there are some places where we do call new_v4() directly

crates/goose/tests/execution_tests.rs
```diff
@@ -0,0 +1,372 @@
+mod execution_tests {
```
DOsinga, line 1: this feels like overtesting

ui/desktop/src/components/settings/extensions/extension-manager.ts
```diff
@@ -254,6 +261,7 @@ interface ToggleExtensionProps {
   extensionConfig: ExtensionConfig;
   addToConfig: (name: string, extensionConfig: ExtensionConfig, enabled: boolean) => Promise<void>;
   toastOptions?: ToastServiceOptions;
+  sessionId?: string;
```
DOsinga: this should not be optional I think

ui/desktop/src/components/settings/extensions/extension-manager.ts
```diff
@@ -134,6 +139,7 @@ interface UpdateExtensionProps {
   removeFromConfig: (name: string) => Promise<void>;
   extensionConfig: ExtensionConfig;
   originalName?: string;
+  sessionId?: string;
```
DOsinga: this should not be optional I think

ui/desktop/src/components/settings/extensions/extension-manager.ts
```diff
@@ -87,6 +89,7 @@ interface AddToAgentOnStartupProps {
   addToConfig: (name: string, extensionConfig: ExtensionConfig, enabled: boolean) => Promise<void>;
   extensionConfig: ExtensionConfig;
   toastOptions?: ToastServiceOptions;
+  sessionId?: string;
```
DOsinga: this should not be optional I think

ui/desktop/src/components/settings/extensions/agent-api.ts
```diff
@@ -43,14 +44,24 @@ export async function extensionApiCall(
...
+
     // Step 2: Make the API call
     const response = await fetch(getApiUrl(endpoint), {
       method: 'POST',
       headers: {
         'Content-Type': 'application/json',
         'X-Secret-Key': await window.electron.getSecretKey(),
       },
-      body: JSON.stringify(payload),
+      body: JSON.stringify(requestBody),
```
DOsinga, line 64: ok, this is a pre-existing mess, but this should really go through the api layer instead of constructing this ourselves. also session-ids here should be required not optional.

ui/desktop/src/components/settings/extensions/ExtensionsSection.tsx
```diff
@@ -34,6 +35,8 @@ export default function ExtensionsSection({
   selectedExtensions = [],
 }: ExtensionSectionProps) {
   const { getExtensions, addExtension, removeExtension, extensionsList } = useConfig();
+  const chatContext = useChatContext();
+  const sessionId = chatContext?.chat.sessionId;
```
DOsinga: grrr. useChatContext seems like the easy way out here, but I am not sure it can be trusted. Maybe we should have a useSession()? alternatively how hard is it to actually thread this through as we do with the other components?

DOsinga (changes requested):
Gave it a thorough look. The main things are:

* session_ids should be required
* let's not implement our own LRU cache for agents
* we should probably enable this in the client immediately, since otherwise we start one goosed per session and nothing changes and we can't really see if this work

crates/goose-server/src/routes/agent.rs
tlongwell-block: Before, there was only one Agent and it was created at startup, so it should always exist

Now, they're created on demand and goosed is long running across many sessions and Agents

Agent creation *could* fail for many reasons now in the middle of goosed running. Goose could be enable to spin up a new Agent because of process limits (extensions), OOM, out of disk for new session files, etc, etc

So, goosed could be up and running happily, with a hundred agents in active use, but *actually* be unable to create an additional agent

We can handle that relatively gracefully with errors like this... Or not handle it and crash goosed

crates/goose/src/execution/manager.rs
tlongwell-block: We use an LRU crate already https://github.com/block/goose/blob/main/crates/goose-mcp/src/developer/analyze/cache.rs#L1

Want me to just use that one here?

crates/goose/src/config/base.rs
tlongwell-block: I think there was a one off lint error around it detecting a rounded `pi`. Let me revert this and see

crates/goose-server/src/routes/agent.rs
tlongwell-block: I think as a default it is fine. Allowing it to be overidden at the session level should be done, though. Maybe in another PR?

crates/goose-server/src/routes/agent.rs
tlongwell-block: I made session_id optional because of an initial, but abandoned, attempt to not require UI changes for this to happen. Let me fix this

crates/goose/tests/execution_tests.rs
tlongwell-block: There are a lot of lines of tests, but this is actually whittled down quite a bit. Each one is testing something nontrivial

I can trim a few when I move to the LRU crate, though. Should cut this down a bit

crates/goose-server/src/routes/audio.rs
tlongwell-block: This test doesn't get an Agent now (because AppState now gets an Agent Manager instead), which causes the error code to be different based on whether on not the testing environment has an openai api key set

crates/goose-server/src/routes/context.rs
tlongwell-block: It was kept optional to continue to auto-generation of session IDs that the UI depended on when generating a new session. But I think its much cleaner if we require it

crates/goose/tests/execution_tests.rs
tlongwell-block: Removed several tests

crates/goose/src/execution/manager.rs
tlongwell-block: Done

ui/desktop/src/components/settings/extensions/agent-api.ts
tlongwell-block: 100%. Would prefer this to be in a followup PR since it will involve api changes + ui changes out of scope

ui/desktop/src/components/settings/extensions/ExtensionsSection.tsx
tlongwell-block: Done

tlongwell-block:
Temporarily removing GOOSE_STANDALONE_MODE work for this PR merge per @michaelneale

crates/goose-server/src/routes/agent.rs
DOsinga: yeah, but that's not what happens, right? if you just go 

    state.get_agent(session_id)?

it doesn't crash goose. it just returns a 500, error, same as your thing does here, so I don't think it adds anything. and you are still wrapping this in a Result!

also, this should not live in agent.rs. that's routing for agents. move it to state. also also add an & so you we can avoid cloning when calling this

crates/goose-server/src/routes/audio.rs
DOsinga: OK, let's delete the test then.

crates/goose-server/src/routes/recipe.rs
```diff
@@ -114,7 +116,18 @@ async fn create_recipe(
...
+        .await
+        .map_err(|e| {
+            (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(CreateRecipeResponse {
+                    recipe: None,
+                    error: Some(e.to_string()),
+                }),
+            )
+        })?;
```
DOsinga: this should not have this expansion and probably call a helper function instead

crates/goose-server/tests/pricing_api_test.rs
```diff
@@ -1,25 +1,11 @@
 use axum::http::StatusCode;
```
DOsinga: let's delete this test

crates/goose/src/execution/manager.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        if let Some(provider) = &*self.default_provider.read().await {
+            debug!(
+                "Setting default provider on agent for session {}",
+                session_id
+            );
+            let _ = agent.update_provider(Arc::clone(provider)).await;
+        }
+
+        // Store in LRU cache (automatically evicts oldest if at capacity)
```
DOsinga: delete this comment

crates/goose/src/execution/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+impl SessionExecutionMode {
+    /// Create an interactive chat mode
+    pub fn chat() -> Self {
+        Self::Interactive
+    }
+
+    /// Create a background/scheduled mode
+    pub fn scheduled() -> Self {
```
DOsinga, line 25: are we actually using this?

crates/goose/src/execution/manager.rs
```diff
@@ -0,0 +1,150 @@
...
+            scheduler,
+            default_provider: Arc::new(RwLock::new(None)),
+        })
+    }
+
+    pub async fn scheduler(&self) -> Result<Arc<dyn SchedulerTrait>> {
+        Ok(Arc::clone(&self.scheduler))
+    }
+
+    pub async fn set_default_provider(&self, provider: Arc<dyn crate::providers::base::Provider>) {
```
DOsinga, line 51: I'm still confused why we have this?

tlongwell-block: This is how Goose worked previously, too.

When an Agent is created, provider can be specified. If not specified, the default one that goosed was launched with is used.

This is to allow the UI to override if needed, but have a default if it chooses not to

crates/goose/src/execution/mod.rs
tlongwell-block: It will be the same execution path as a subtask/recipe, just without a parent session reference, hence the separate mode

crates/goose/src/execution/manager.rs
DOsinga: does it?

```
    let new_agent = Agent::new();
    let agent_ref = Arc::new(new_agent);

    let app_state = state::AppState::new(agent_ref.clone());
```

is what it does on main and leaves it alone. but also, the desktop will always set the provider after it starts the agent. and even if it used to work like this, I don't think it should :)

crates/goose/src/execution/mod.rs
tlongwell-block: I'll take it out for this PR and reintroduce it later

crates/goose/src/execution/manager.rs
tlongwell-block: talked this over, letting agent manager set a default on Agents when they spawn for now

crates/goose/src/execution/mod.rs
tlongwell-block: Okay, changed my thinking a bit. These *are* just stubs that aren't used now. But they're in the right place and will be used in the next PR opened and I just want to stake the claim here now to avoid any confusion

DOsinga (approved):
to be honest, I still think we do the default provider wrong; it shouldn't even be a method in the agentmanager, just something that gets constructed ad hoc when the agent is configured, for reasons we discusssed, currently we end up with the same provider for different agents, plus if the settings change, we don't pick that up. But the desktop overrides it anyway, so we can do that in a follow up PR if you want

PR #4683, opened 2025-09-19
feat(acp): upgrade agent-client-protocol + improve tool experience

alexhancock:
Improves the acp command a bit by...

* Upgrading the agent-client-protocol crate dep
* Make content from tool results accessible via acp
* Refactoring is mainly to align with the new crate

Screenshot from running in zed

Screenshot 2025-09-19 at 3 32 06 PM

Can do another followup to make tool names more human-friendly and useful

Helpful config to run goose from source (on any local branch) from zed:

```
# ~/.config/zed/settings.json
{
  "theme": "One Dark",
  "agent_servers": {
    "goose": {
      "command": "cargo",
      "args": [
        "+nightly",
        "-Z",
        "unstable-options",
        "-C",
        "/PATH/TO/GOOSE/REPO",
        "run",
        "--bin",
        "goose",
        "acp"
      ],
      "env": {}
    }
  }
}

```

PR #4686, opened 2025-09-20
feat(acp): add nicer formatting for tool names

alexhancock:
Makes tool calls look nicer when rendered by an acp client.

Example in Zed:

**Before**

Screenshot 2025-09-19 at 9 17 34 PM

**After**

Screenshot 2025-09-19 at 9 16 15 PM

### DIFF
--- a/crates/goose-cli/src/commands/acp.rs
+++ b/crates/goose-cli/src/commands/acp.rs
@@ -49,6 +49,50 @@ fn read_resource_link(link: acp::ResourceLink) -> Option<String> {
     }
 }
 
+/// Format a tool name to be more human-friendly by splitting extension and tool names
+/// and converting underscores to spaces with proper capitalization
+fn format_tool_name(tool_name: &str) -> String {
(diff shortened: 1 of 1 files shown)


alexhancock:
@benbrandt With this and https://github.com/block/goose/pull/4683 I think we're making progress on https://github.com/zed-industries/zed/pull/38219#issuecomment-3309309581

You can test in zed on a copy of this branch with:

```
  "agent_servers": {
    "goose": {
      "command": "cargo",
      "args": [
        "+nightly",
        "-Z",
        "unstable-options",
        "-C",
        "/WHEREVER_YOU_STORE/goose",
        "run",
        "--bin",
        "goose",
        "acp"
      ],
      "env": {}
    }
  }
```

or I could build a copy of goose for you to test with and for us to get feedback.

Any other `ToolCallUpdate` calls you think we should do?

michaelneale (approved):
nice and tests!

PR #4688, opened 2025-09-20
Offer to summarize or clear conversation when it has gotten too long …

dhanji:
…rather than just displaying this

## Pull Request Description

use cliclac to display a choice.


katzdave (approved):

alexhancock (approved):
Yeah, better than what is there!

I also realized from this though that we have some duplication around this concept of `summarize_context_messages`. Can take a pass at that after

alexhancock:
Going to merge this one in so we can combine with https://github.com/block/goose/pull/4565 and get both in!

dhanji:
thanks!

PR #4691, opened 2025-09-22
[Security] Implement Security Option for Handling Tool-Call Chains

shellz-n-stuff:
## Pull Request Description
One of the key attacks against Goose is MCP poisoning. Using the response data from an MCP to trigger a non-expected tool-call that performs an unintended action.

Whilst we absolutely don't want to remove the utility of the agent autonomously deciding to call a chain of tools, users should be able to configure security profiles in their config to prevent certain tools being called as a "secondary tool" without user input.

This PR introduces such functionality

## Implementation Detail
The general approach is to intercept tool-calls and perform a message look back to see what tools have been called since prior invocation. Specifically, the method is as follows:
1. Check for config of tools to protect (if null skip the check)
2. Check if the current tool-call exists in the list of protected tools (if null skip)
3. Collect a list of tool calls between the tool we're about to invoke and the last user message (if 0 skip)
4. If the last called tool is this tool skip
5. If the last called tool is another tool raise a security alert.

## Testing

Screenshot 2025-09-22 at 5 23 24 pm

- Config File Setup

Screenshot 2025-09-22 at 5 25 32 pm

- Base Case Running a Normal MCP + The developer__shell mcp - Part 1

Screenshot 2025-09-22 at 5 26 01 pm

- Base Case Running a Normal MCP + The developer__shell mcp - Part 2

Screenshot 2025-09-22 at 5 24 50 pm

- Secondary Case running two instances of developer shell

### DIFF
--- a/crates/goose/src/security/scanner.rs
+++ b/crates/goose/src/security/scanner.rs
@@ -5,7 +5,6 @@ use mcp_core::tool::ToolCall;
 use rmcp::model::Role;
 use serde_json::Value;
 
-/// Result of a security scan.
 #[derive(Debug, Clone)]
 pub struct ScanResult {
     pub is_malicious: bool,
@@ -13,21 +12,19 @@ pub struct ScanResult {
     pub explanation: String,
 }
 
-/// Scanner for prompt injection and tool misuse.
 pub struct PromptInjectionScanner {
     pattern_matcher: PatternMatcher,
 }
 
 impl PromptInjectionScanner {
-    /// Create a new scanner.
     pub fn new() -> Self {
         Self {
             pattern_matcher: PatternMatcher::new(),
         }
     }
 
-    /// Get the maliciousness threshold from config, or use default.
-    pub fn get_threshold_from_config(&self) -> f32 {
+    /// Get the confidence threshold from config, or use default.
+    pub fn get_confidence_threshold_from_config(&self) -> f32 {
         use crate::config::Config;
         let config = Config::global();
 
@@ -102,11 +99,6 @@ impl PromptInjectionScanner {
         self.scan_for_dangerous_patterns(system_prompt).await
     }
 
-    /// Scan text for prompt injection (legacy compatibility).
-    pub async fn scan_with_prompt_injection_model(&self, text: &str) -> Result<ScanResult> {
-        self.scan_for_dangerous_patterns(text).await
-    }
-
     /// Core pattern matching logic for dangerous content.
     pub async fn scan_for_dangerous_patterns(&self, text: &str) -> Result<ScanResult> {
         let matches = self.pattern_matcher.scan_text(text);
@@ -174,7 +166,7 @@ impl PromptInjectionScanner {
             _ => None,
         }
     }
-
+    
     async fn is_secondary_tool_violation_single(
         &self,
         tool_call: &ToolCall,
@@ -191,14 +183,6 @@ impl PromptInjectionScanner {
             return false;
         }
 
-        tracing::debug!(
-            tool_name,
-            "Checking secondary tool violations for '{}'",
-            tool_name
-        );
-
-        tracing::debug!("messages: {:?}", messages.len());
-        tracing::debug!("messages: {:?}", messages);
         // Find the most recent user message by iterating from the end,
         // but skip messages that are tool responses (i.e., user messages that are ToolResponse)
         let last_user_idx = messages.iter().enumerate().rev().find_map(|(i, m)| {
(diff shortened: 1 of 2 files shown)


crates/goose/src/security/scanner.rs
```diff
@@ -1,66 +1,113 @@
-use crate::conversation::message::Message;
+use crate::conversation::message::{Message, MessageContent};
 use crate::security::patterns::{PatternMatcher, RiskLevel};
 use anyhow::Result;
 use mcp_core::tool::ToolCall;
+use rmcp::model::Role;
 use serde_json::Value;
 
+/// Result of a security scan.
 #[derive(Debug, Clone)]
```
DOsinga, line 8: remove

crates/goose/src/security/scanner.rs
```diff
@@ -1,66 +1,113 @@
...
 impl PromptInjectionScanner {
+    /// Create a new scanner.
     pub fn new() -> Self {
         Self {
             pattern_matcher: PatternMatcher::new(),
         }
     }
 
-    /// Get threshold from config
+    /// Get the maliciousness threshold from config, or use default.
```
DOsinga: if you're not happy with the current comment, rename the function (drop from_config, I don't think the caller cares where we store this)

crates/goose/src/security/scanner.rs
```diff
@@ -1,66 +1,113 @@
...
     }
 
-    /// Scan system prompt for injection attacks
+    /// Scan a system prompt for injection attacks.
     pub async fn scan_system_prompt(&self, system_prompt: &str) -> Result<ScanResult> {
         self.scan_for_dangerous_patterns(system_prompt).await
     }
 
-    /// Scan with prompt injection model (legacy method name for compatibility)
+    /// Scan text for prompt injection (legacy compatibility).
```
DOsinga: ugh, we should have caught this the first place. remove this comment - legacy

crates/goose/src/security/scanner.rs
```diff
@@ -121,27 +165,102 @@ impl PromptInjectionScanner {
...
+    }
+
+    async fn is_secondary_tool_violation_single(
+        &self,
+        tool_call: &ToolCall,
+        messages: &[Message],
+        disabled_secondary_tool_list: &[String],
+    ) -> bool {
+        let tool_name = tool_call.name.as_str();
+        if !disabled_secondary_tool_list.iter().any(|t| t == tool_name) {
```
DOsinga, line 177: I'd consider having the caller check this; that would leave this function with just one job, makes naming easier too

crates/goose/src/security/scanner.rs
```diff
@@ -121,27 +165,102 @@ impl PromptInjectionScanner {
...
+                            msg_index = msg_idx,
+                            "Secondary tool violation: found '{}' (expected '{}') after last user message",
+                            offending,
+                            tool_name
+                        );
+                        return true;
+                    }
+                }
+            }
+        }
```
DOsinga, line 231: you're going through the message twice here, I think you can do it in one go like:

for msg in messages.iter().rev():
    if isUsere(msg): break
    if toolCallName(msg) != toolCallName: return true
return true

crates/goose/src/security/scanner.rs
shellz-n-stuff: Yeah ez

crates/goose/src/security/scanner.rs
shellz-n-stuff: I was thinking of just making it fully self-contained (IE it "just works").

I may split this into into a function though to make it cleaner/improve readability

shellz-n-stuff: Actually looking at it, we re-use the tool_name var quite a lot. So I think it's really a call on if the caller should do it vs the function.

I personally like the function having the check but happy to change if you feel strongly

DOsinga: you should keep the toolname of course. I can go either way on it, but the function is called is_secondary_tool_violation_single, not is_secondary_tool_violation_single_if_enabled

crates/goose/src/security/scanner.rs
DOsinga: also, we have effective_role which will return tool for tool messages

michaelneale:
config.yaml makes sense as we already have it for rules for commands, but in future I wonder if people will want to build a distro with these more role-based (ie protect people from themselves) but I digress... looking good so far!

michaelneale:
LGTM - but will want to check if this will be a problem with config.yaml: 

change here https://github.com/block/goose/pull/4651 - does this imply the config section in config.yaml is changing cc @dorien-koelemeijer ? (ie no security: section?)

dorien-koelemeijer:
> LGTM - but will want to check if this will be a problem with config.yaml:
> 
> change here #4651 - does this imply the config section in config.yaml is changing cc @dorien-koelemeijer ? (ie no security: section?)

Yeah, perhaps a comment on this whether we could also configure all of this through the UI settings and keep the config as all other config (with underscores)? See https://github.com/block/goose/pull/4651 @shellz-n-stuff

shellz-n-stuff:
> #4651

@dorien-koelemeijer, I can make the updates on this branch after your PR gets merged? Not in a mega rush here!

michaelneale (approved):

dorien-koelemeijer:
> > #4651
> 
> @dorien-koelemeijer, I can make the updates on this branch after your PR gets merged? Not in a mega rush here!

Just wondering whether you wanted me to update/rename any config items or not?

crates/goose/src/security/scanner.rs
```diff
@@ -22,45 +23,83 @@ impl PromptInjectionScanner {
         }
     }
 
-    /// Get threshold from config
-    pub fn get_threshold_from_config(&self) -> f32 {
+    /// Get the confidence threshold from config, or use default.
+    pub fn get_confidence_threshold_from_config(&self) -> f32 {
```
DOsinga, line 27: I thought we'd just call this get_confidence_threshold and drop the comment?

dorien-koelemeijer:
Just adding a comment for visibility instead of chatting over Slack - I've now updated the config for prompt injection to be as follows: `security_prompt_enabled` and `security_prompt_threshold`, so perhaps you could update your config items to follow the same convention? Decided for the non-nested, underscore approach since that seems to be the more conventional thing to do looking at the other config. Happy to discuss more ofc if you wanted me to update something!

DOsinga:
do we still want this to land?

DOsinga:
I'm going to close this for now due to staleness. can always reopen when relevant again

PR #4693, opened 2025-09-22
Create / edit recipe form unification and improvements

zanesq:
## Pull Request Description
Refactored create/edit recipe in recipe manager and chat to use the same form and experience. Added a few missing features and usability improvements while I was at it.

This is not a full redesign of recipes. Its a first step towards unifying the front end code and experience in its current state to fill in the gaps in preparation of moving recipe functionality to the backend and a full redesign. Will follow up in another PR with tests.

## Changes
- Create/edit recipe forms all match and use the same fields and logic using our new tanstack form library + validation
- Create recipe from chat now passes the session id rather than the messages and generates the title and description as well.
- Revamped recipe creation experience so recipes are now generated and opened in the same window
- Removed the egg loader with a more engaging loading state
- Removed `recipe-editor` route since its no longer in a new window
- Added collapse and expand interface for parameters and inline usage detection indicators
- Can now define a parameter name when creating rather than auto generation of a random name
- Added individual parameter delete option rather than only delete last
- Fixed form labels and helper text to reflect reality and be more helpful
- Added copy deeplink button to recipes view

## Screenshots
image
image
image
image
image
image

### DIFF
--- a/ui/desktop/src/components/recipes/__tests__/CreateRecipeFromSessionModal.test.tsx
+++ b/ui/desktop/src/components/recipes/__tests__/CreateRecipeFromSessionModal.test.tsx
@@ -25,7 +25,6 @@ describe('CreateRecipeFromSessionModal', () => {
     onClose: vi.fn(),
     sessionId: 'test-session-id',
     onRecipeCreated: vi.fn(),
-    onStartRecipe: vi.fn(),
   };
 
   beforeEach(() => {
@@ -249,7 +248,7 @@ describe('CreateRecipeFromSessionModal', () => {
       );
     });
 
-    it('creates recipe when form is submitted', async () => {
+    it('creates recipe and closes modal when form is submitted', async () => {
       const user = userEvent.setup();
       render(<CreateRecipeFromSessionModal {...defaultProps} />);
 
@@ -263,61 +262,9 @@ describe('CreateRecipeFromSessionModal', () => {
       await user.click(screen.getByTestId('create-recipe-button'));
 
       await waitFor(() => {
-        expect(screen.getByTestId('success-state')).toBeInTheDocument();
-        expect(screen.getByTestId('done-button')).toBeInTheDocument();
-        expect(screen.getByTestId('start-recipe-button')).toBeInTheDocument();
+        expect(defaultProps.onRecipeCreated).toHaveBeenCalled();
+        expect(defaultProps.onClose).toHaveBeenCalled();
       });
-
-      expect(defaultProps.onRecipeCreated).toHaveBeenCalled();
-    });
-  });
-
-  describe('Success State Actions', () => {
-    it('calls onStartRecipe when Start Recipe button is clicked', async () => {
-      const user = userEvent.setup();
-      render(<CreateRecipeFromSessionModal {...defaultProps} />);
-
-      // Wait for form and create recipe
-      await waitFor(
-        () => {
-          expect(screen.getByTestId('create-recipe-button')).toBeEnabled();
-        },
-        { timeout: 2000 }
-      );
-
-      await user.click(screen.getByTestId('create-recipe-button'));
-
-      await waitFor(() => {
-        expect(screen.getByTestId('start-recipe-button')).toBeInTheDocument();
-      });
-
-      await user.click(screen.getByTestId('start-recipe-button'));
-
-      expect(defaultProps.onStartRecipe).toHaveBeenCalled();
-      expect(defaultProps.onClose).toHaveBeenCalled();
-    });
-
-    it('closes modal when Done button is clicked', async () => {
-      const user = userEvent.setup();
-      render(<CreateRecipeFromSessionModal {...defaultProps} />);
-
-      // Wait for form and create recipe
-      await waitFor(
-        () => {
-          expect(screen.getByTestId('create-recipe-button')).toBeEnabled();
-        },
-        { timeout: 2000 }
-      );
-
-      await user.click(screen.getByTestId('create-recipe-button'));
-
-      await waitFor(() => {
-        expect(screen.getByTestId('done-button')).toBeInTheDocument();
-      });
-
-      await user.click(screen.getByTestId('done-button'));
-
-      expect(defaultProps.onClose).toHaveBeenCalled();
     });
   });
 
@@ -346,7 +293,6 @@ describe('CreateRecipeFromSessionModal', () => {
     });
 
     it('shows different button states based on workflow stage', async () => {
-      const user = userEvent.setup();
       render(<CreateRecipeFromSessionModal {...defaultProps} />);
 
       expect(screen.getByTestId('cancel-button')).toBeInTheDocument();
@@ -359,12 +305,7 @@ describe('CreateRecipeFromSessionModal', () => {
         { timeout: 2000 }
       );
 
-      await user.click(screen.getByTestId('create-recipe-button'));
-
-      await waitFor(() => {
-        expect(screen.getByTestId('done-button')).toBeInTheDocument();
-        expect(screen.getByTestId('start-recipe-button')).toBeInTheDocument();
-      });
+      expect(screen.getByTestId('create-and-run-recipe-button')).toBeInTheDocument();
     });
   });
 


ui/desktop/src/hooks/useAgent.ts
```diff
@@ -59,12 +60,19 @@ export function useAgent(): UseAgentReturn {
     setSessionId(null);
     setAgentState(AgentState.UNINITIALIZED);
     setRecipeFromAppConfig(null);
+    initPromiseRef.current = null; // Clear any pending initialization
```
zanesq: `initPromiseRef` serves as a singleton pattern for agent initialization, basically it prevents race conditions by cacheing the initialization promise so subsequent calls to loadCurrentChat will return the same promise instead of starting a new initialization.

michaelneale:
this looks much much nicer already - any particular testing notes?

zanesq:
I have a branch off this with tests, will push up shortly but for manual testing basically everything related to create/edit and running recipes from a chat session and recipe manager. I've verified deep links continue to work as before.

michaelneale:
@zanesq not sure why, but with this I can't run recipes the same as on main. I made a simple recipe (in the GUI here) to say a joke in french and spanish. 

on this branch: 

image

image

which isn't right (it doesn't read the prompts at all it seems). 


on main (with has a Use button): 

image

image

zanesq:
Hmm might be an issue with launching the recipe in the same window. Thanks for testing, will take a look and report back.

zanesq:
@michaelneale was indeed a bug, just pushed a fix and verified locally

image

zanesq:
ok merged in main with the session agent changes and all related PRs, doing final testing after all that

zanesq:
Okay thoroughly tested locally I think this is gtg but can use a few more testers.

amed-xyz (approved):
Reviewed & tested in a local build. Looking great!

LGTM 🚀

zanesq:
I am resolving conflicts and will push up and retest 👍

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -294,7 +286,7 @@ function BaseChatContent({
 
     // Call the callback if provided (for Hub to handle navigation)
     if (onMessageSubmit && combinedTextFromInput.trim()) {
-      onMessageSubmit(combinedTextFromInput);
+      onMessageSubmit();
```
zanesq: This wasn't needed for ui state. Message content is already available through chat  and messages in useChatEngine

zanesq:
main merged, cleaned up, tested and ready for review again

crates/goose-server/src/routes/session.rs
```diff
@@ -25,6 +26,13 @@ pub struct UpdateSessionDescriptionRequest {
     description: String,
 }
 
+#[derive(Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct UpdateSessionRecipeParametersRequest {
+    /// Recipe parameter values entered by the user
+    recipe_parameters: HashMap<String, String>,
```
zanesq: Thinking of renaming this to `recipe_parameters_values` since it contains the parameters and user entered values and could be confused with the parameters in the recipe schema that contain the parameter property names and no values.

ui/desktop/src/components/pair.tsx
```diff
@@ -48,21 +51,92 @@ export default function Pair({
...
+      if (recipeConfig && recipeChanged) {
+        prevRecipeConfigRef.current = recipeConfig;
+
+        try {
+          // Load a fresh chat session with forced reset
+          const newChat = await loadCurrentChat({
+            resumeSessionId: undefined,
+            recipeConfig: recipeConfig,
+            setAgentWaitingMessage,
+            forceReset: true,
```
zanesq: I realize we now have 2 patterns for this `forceReset` and `resetChat` that do slightly different things. This PR is already big so I plan on following up and refactoring this with the multi-chat work next.

crates/goose-cli/src/session/builder.rs
```diff
@@ -239,7 +239,7 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> CliSession {
             process::exit(1);
         });
 
-    // Handle session file resolution and resuming
+    // Handle session resolution and resuming
```
DOsinga, line 281: yes

crates/goose-server/src/routes/recipe.rs
```diff
@@ -111,25 +105,42 @@ async fn create_recipe(
...
+            let error_message = format!("Failed to get session: {}", e);
+            let error_response = CreateRecipeResponse {
+                recipe: None,
+                error: Some(error_message),
+            };
+            return Ok(Json(error_response));
+        }
+    };
+
+    let conversation = match session.conversation {
```
DOsinga, line 137: this is a bit annoying, but can't fail, so just go with .or_default()

crates/goose-server/src/routes/recipe.rs
```diff
@@ -111,25 +105,42 @@ async fn create_recipe(
...
     );
 
+    // Load messages from session
+    let session = match SessionManager::get_session(&request.session_id, true).await {
+        Ok(session) => session,
+        Err(e) => {
+            let error_message = format!("Failed to get session: {}", e);
+            let error_response = CreateRecipeResponse {
+                recipe: None,
+                error: Some(error_message),
```
DOsinga: ugh. can we remove the error from this and just return a 500? that would be much more in line with how we do things

crates/goose-server/src/routes/session.rs
DOsinga: yes! let's call it user values everywhere

crates/goose/src/prompts/recipe.md
```diff
@@ -1,13 +1,16 @@
...
 
-2. A list of 3-5 example activities (as a few words each at most) that would be relevant to this topic
-
-Format your response in _VALID_ json, with one key being `instructions` which contains a string, and the other key `activities` as an array of strings.
+Format your response in _VALID_ json, with keys being `title`, `description`, `instructions` (string), and `activities` (array of strings).
 For example, perhaps we have been discussing fruit and you might write:
 
 {
+"title": "Fruit Information Assistant",
+"description": "A recipe for finding and sharing information about different types of fruit.",
```
DOsinga, line 13: this is great

crates/goose/src/agents/agent.rs
```diff
@@ -1696,9 +1696,32 @@ impl Agent {
             extension_configs.len()
         );
 
+        // Extract title and description from the JSON response if available
+        let (title, description) =
```
DOsinga, line 1612: the comment is saying the same as the code

this is good though. I do wonder if we should move this to recipe.rs or some such. it really has no place in this file that already has > 2000 lines

crates/goose/src/session/session_manager.rs
```diff
@@ -10,14 +10,15 @@ use rmcp::model::Role;
...
+use std::collections::HashMap;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use tokio::sync::OnceCell;
 use tracing::{info, warn};
 use utoipa::ToSchema;
 
-const CURRENT_SCHEMA_VERSION: i32 = 1;
+const CURRENT_SCHEMA_VERSION: i32 = 2;
```
DOsinga, line 22: do we not need an actual migration here?

ui/desktop/src/components/pair.tsx
DOsinga: yeah, this is a mess, I agree. part of it is that we have another copy of the chats deep in there that tends to override what pair does. it needs to go

ui/desktop/src/components/parameter/ParameterInput.tsx
```diff
@@ -1,110 +1,193 @@
...
-      <h3 className="text-lg font-bold text-textProminent mb-4">
-        Parameter:{' '}
-        <code className="bg-background-default px-2 py-1 rounded-md">{parameter.key}</code>
-      </h3>
+  const handleToggleExpanded = (e: React.MouseEvent) => {
+    // Only toggle if we're not clicking on the delete button
+    if (onToggleExpanded && !(e.target as HTMLElement).closest('button')) {
+      onToggleExpanded(key);
+    }
+  };
```
DOsinga, line 30: what does this mean?

ui/desktop/src/components/parameter/ParameterInput.tsx
```diff
@@ -1,110 +1,193 @@
...
 
-        <div>
-          <label className="block text-md text-textStandard mb-2 font-semibold">Requirement</label>
-          <select
-            className="w-full p-3 border rounded-lg bg-background-default text-textStandard"
-            value={requirement}
-            onChange={(e) =>
-              onChange(key, { requirement: e.target.value as Parameter['requirement'] })
-            }
+        {/* Delete button - always visible */}
```
DOsinga: well, you say that, but only if onDelete is set

ui/desktop/src/components/recipes/CreateRecipeFromSessionModal.tsx
```diff
@@ -0,0 +1,355 @@
...
+
+      // Update stage every 800ms
+      const stageInterval = setInterval(() => {
+        currentStageIndex = (currentStageIndex + 1) % stages.length;
+        setAnalysisStage(stages[currentStageIndex]);
+      }, 800);
+
+      // Call the backend to analyze messages and create a recipe
+      createRecipe({
+        body: { session_id: sessionId },
```
DOsinga, line 78: add a throwError:true here

ui/desktop/src/components/recipes/CreateRecipeFromSessionModal.tsx
```diff
@@ -0,0 +1,355 @@
...
+          } else {
+            console.error('No recipe in response:', response);
+          }
+          setHasAnalyzed(true);
+        })
+        .catch((error) => {
+          console.error('Failed to analyze messages:', error);
+          clearInterval(stageInterval);
+          setAnalysisStage('Analysis failed');
+          setHasAnalyzed(true);
```
DOsinga: clearInterval & hasAnalyzed can go to finally to reduce duplication

ui/desktop/src/utils/navigationUtils.ts
```diff
@@ -27,13 +26,14 @@ export type ViewOptions = {
   sessionDetails?: unknown;
   error?: string;
   baseUrl?: string;
-  config?: unknown;
+  recipeConfig?: unknown;
```
DOsinga: we can type this to Recipe, no? also here and elsewhere, can we just call it recipe instead of recipeConfig?

ui/desktop/src/utils/providerUtils.ts
```diff
@@ -92,7 +93,19 @@ export const filterValidUsedParameters = (
   const instructionVariables = recipeContent.instructions
     ? extractTemplateVariables(recipeContent.instructions)
     : [];
-  const allUsedVariables = [...new Set([...promptVariables, ...instructionVariables])];
+
+  // Extract variables from activities
+  const activityVariables: string[] = [];
+  if (recipeContent.activities && Array.isArray(recipeContent.activities)) {
```
DOsinga: no need to check on whether this is an array in ts, also I think you can do something like:

    recipeContent.activities?.flatMap(extractTemplateVariables)

ui/desktop/src/utils/providerUtils.ts
```diff
@@ -296,3 +337,28 @@ export const initializeSystem = async (
...
+      },
+      body: {
+        recipeParameters,
+      },
+      throwOnError: true,
+    });
+  } catch (error) {
+    console.error('Failed to save recipe parameters to session metadata:', error);
+  }
+};
```
DOsinga: I am not a fan of having shallow helper functions over API calls especially when we're only calling them once. also you're swallowing the error here, let's definitely not do that.

ui/desktop/src/utils/providerUtils.ts
```diff
@@ -8,6 +8,7 @@ import type { ExtensionConfig, FixedExtensionEntry } from '../components/ConfigC
 import { addSubRecipesToAgent } from '../recipe/add_sub_recipe_on_agent';
 import {
   extendPrompt,
+  Recipe,
```
DOsinga, line 10: it strikes me that these functions do not really belong in providerUtils, maybe recipe/index.ts?

ui/desktop/src/hooks/useRecipeManager.ts
```diff
@@ -1,24 +1,25 @@
...
+import { toastSuccess } from '../toasts';
 
 export const useRecipeManager = (chat: ChatType, recipeConfig?: Recipe | null) => {
-  const [isGeneratingRecipe, setIsGeneratingRecipe] = useState(false);
   const [isParameterModalOpen, setIsParameterModalOpen] = useState(false);
-  const [readyForAutoUserPrompt, setReadyForAutoUserPrompt] = useState(false);
-  const [recipeError, setRecipeError] = useState<string | null>(null);
   const [isRecipeWarningModalOpen, setIsRecipeWarningModalOpen] = useState(false);
   const [recipeAccepted, setRecipeAccepted] = useState(false);
+  const [isCreateRecipeModalOpen, setIsCreateRecipeModalOpen] = useState(false);
```
DOsinga, line 18: here and elsewhere, I don't think the tracking of whether these things are open belong in useRecipeManager. the respective dialogs should manage that, no?

ui/desktop/src/components/recipes/shared/__tests__/RecipeActivityEditor.test.tsx
```diff
@@ -0,0 +1,113 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
```
DOsinga, line 1: please review these tests and check whether you think they will catch bugs or just annoy us when we change texts

ui/desktop/src/hooks/useAgent.ts
DOsinga: this comment is true. if we need to reset this though I think we might have to await it first though since if that promise is still the case it means we still have a request in flight. I don't think that that should happen in reality, but maybe need to be sure. also remove the comment

ui/desktop/src/hooks/useAgent.ts
```diff
@@ -59,12 +60,19 @@ export function useAgent(): UseAgentReturn {
...
   const currentChat = useCallback(
     async (initContext: InitializationContext): Promise<ChatType> => {
-      if (agentIsInitialized && sessionId) {
+      // If forceReset is true, always reinitialize
+      if (initContext.forceReset) {
+        setSessionId(null);
+        setAgentState(AgentState.UNINITIALIZED);
+        initPromiseRef.current = null;
+        // Don't return early, continue to initialization
+      } else if (agentIsInitialized && sessionId) {
```
DOsinga: you have to explain to me why we need this over resetChat - resetChat followed by calling this again should be the same thing, except for that here we aren't clearing the recipe?

ui/desktop/src/hooks/useAgent.ts
```diff
@@ -151,15 +164,24 @@ export function useAgent(): UseAgentReturn {
...
             sessionId: agentSession.id,
             title: agentSession.recipe?.title || agentSession.description,
             messageHistoryIndex: 0,
-            messages: messages.map((message: ApiMessage) =>
-              convertApiMessageToFrontendMessage(message)
-            ),
-            recipeConfig: agentSession.recipe,
+            messages: messages,
+            recipeConfig: recipeConfig,
+            recipeParameters: agentSession.recipe_parameters || null,
```
DOsinga: hmm, we should talk about this. we're playing fast and lose with messages and recipes here that I am not sure about. maybe I am not following, but here's how I think this should work: we have two cases:

start_agent: we provide a recipe, this recipe is then returned in the session by this call.

resume_agent: we just provide a session_id, the call returns the pre-existing recipe

in either case we should trust the recipe coming back from the agent. same for the messages; if we started a new session, messages should be empty, if we resume a session, we can't be setting a recipe

ui/desktop/src/hooks/useRecipeManager.ts
```diff
@@ -67,11 +101,14 @@ export const useRecipeManager = (chat: ChatType, recipeConfig?: Recipe | null) =
...
         }
+      } else {
+        setRecipeAccepted(false);
+        setIsRecipeWarningModalOpen(false);
       }
     };
 
     checkRecipeAcceptance();
-  }, [finalRecipeConfig]);
+  }, [finalRecipeConfig, recipeConfig]);
```
DOsinga: this code was hard to read before, but I feel now it has so many exceptions and weird logic that we need to rethink this

ui/desktop/src/hooks/useRecipeManager.ts
```diff
@@ -81,22 +118,59 @@ export const useRecipeManager = (chat: ChatType, recipeConfig?: Recipe | null) =
...
+  }, [
+    requiresParameters,
+    hasAllRequiredParameters,
+    recipeAccepted,
+    filteredParameters,
+    isParameterModalOpen,
+    hasMessages,
+    chat.sessionId,
+    finalRecipeConfig?.title,
+  ]);
```
DOsinga, line 173: this also strikes me as rather fragile

ui/desktop/src/components/BaseChat.tsx
DOsinga: eh, maybe? useChatEngine should probably be refactored quite a bit soon though

ui/desktop/src/hooks/useRecipeManager.ts
zanesq: agreed but since a lot of it relies on recipeAccepted it would require some major refactoring.. might be out of scope for this PR but will consider it

ui/desktop/src/components/parameter/ParameterInput.tsx
zanesq: there is a delete icon button in the toggle area on the right side (see param expander ui screenshot in pr description) so its singling out the click outside the delete button when the user intends to expand the content and not delete

crates/goose/src/session/session_manager.rs
zanesq: thought it would be safer in case a release goes out with the current schema before this gets merged.. no real drawback to keeping it a migration

crates/goose-server/src/routes/session.rs
zanesq: ok so `recipe_parameter_user_values`? getting kinda long

ui/desktop/src/components/BaseChat.tsx
zanesq: feels scary but I confirmed that messages and compaction are working fine (well before it broke on main - will recheck after its fixed)

ui/desktop/src/utils/providerUtils.ts
zanesq: changed to flatMap but the isArray was a specific validation check

The Array.isArray() check is necessary because:

Type Safety: Even though TypeScript types suggest parameters should be RecipeParameter[] | undefined, at runtime it could be any value due to API responses, user input, or other dynamic data sources.

Defensive Programming: The function needs to handle cases where the parameters might be:

```
null
undefined
An object instead of an array (like {})
A primitive value
Any other non-array type
```

Test Coverage: The test explicitly verifies this behavior by passing an object {} cast as RecipeParameter[] to simulate a runtime type mismatch.

DOsinga: that sounded a lot like an LLM formulated that answer

I understand that typescript only pretends to give you type safety and that in some circumstances you need to check things anyway. but why here? you're checking that it's an array not that it is an array of RecipeParameter. you say it could be any value due to API responses. do you really not trust our API to be type safe? we should fix that. 

as for test coverage, I am not sure what you mean. you don't want to remove this check because we have a test for it?

ui/desktop/src/hooks/useRecipeManager.ts
DOsinga: but isCreateRecipeModalOpen is new here, no need to make it more complicated than it is?

ui/desktop/src/utils/providerUtils.ts
zanesq: lol you caught me sorry was being lazy at the end of the workday.. yes makes total non llm sense removed 👍

crates/goose-server/src/routes/session.rs
zanesq: after chatting with Douwe renamed it `user_recipe_values`

zanesq:
@DOsinga I pushed up all the minor feedback. I started another branch with the major refactoring feedback but I'm tempted to merge this as is in the meantime so it doesn't drift too far.

DOsinga:
> @DOsinga I pushed up all the minor feedback. I started another branch with the major refactoring feedback but I'm tempted to merge this as is in the meantime so it doesn't drift too far.

we could do that. can you explain about useAgent though why we need those changes, those seem like the most fishy. and we'd have to do it in a quick follow up where we really figure out what the architecture of this should be

zanesq:
@DOsinga it was to work around the current front end edge cases of launching a chat / recipe in the same window and I agree its hacky so I reverted all that code and changed recipes back to launch in a new window for now until we come up with a better plan for handling this. Working on a few last feedback points and testing thoroughly to wrap this up.

zanesq:
Sounds good will do asap. Wrapping up final testing before I merge.

PR #4698, opened 2025-09-23
goosed standalone works with providers

dhanji:
## goosed working

working version of the goose-server crate. can launch a goosed webserver that manages sessions and provides streaming replies independent of CLI and electron app.

### DIFF
--- a/crates/goose-server/src/commands/agent.rs
+++ b/crates/goose-server/src/commands/agent.rs
@@ -33,12 +33,12 @@ pub async fn run() -> Result<()> {
         std::env::var("GOOSE_SERVER__SECRET_KEY").unwrap_or_else(|_| "test".to_string());
 
     let new_agent = Agent::new();
-    
+
     // Only initialize provider and extensions when running in standalone goosed mode
     // This prevents breaking the Electron app which manages its own provider setup
     if std::env::var("GOOSE_STANDALONE_MODE").unwrap_or_else(|_| "false".to_string()) == "true" {
         tracing::info!("Running in standalone mode - initializing provider and extensions");
-        
+
         // Initialize provider like the CLI does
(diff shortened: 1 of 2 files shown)


michaelneale:
so you run goosed but it picks up same config as cli?

crates/goose-server/src/commands/agent.rs
```diff
@@ -33,6 +33,27 @@ pub async fn run() -> Result<()> {
         std::env::var("GOOSE_SERVER__SECRET_KEY").unwrap_or_else(|_| "test".to_string());
 
     let new_agent = Agent::new();
+    
+    // Initialize provider like the CLI does
+    let config = goose::config::Config::global();
+    
+    let provider_name: String = config
```
michaelneale: if this is same code path as GUI, then in case of gui we dont' want it to fail (I think?) as it will update the provider later (I think... hrm...) but this is when standalone?

dhanji: oops, will fix.

dhanji: PTAL

crates/goose-server/src/commands/agent.rs
```diff
@@ -33,6 +34,20 @@ pub async fn run() -> Result<()> {
...
+
+    let provider_name: String = config.get_param("GOOSE_PROVIDER");
+
+    let model_name: String = config.get_param("GOOSE_MODEL");
+
+    let model_config = goose::model::ModelConfig::new(&model_name);
+
+    let provider = goose::providers::create(&provider_name, model_config);
+
+    new_agent.update_provider(provider).await;
```
michaelneale: hrm should this be conditional on it finding that config?

PR #4701, opened 2025-09-23
fix: view can recognise a dir

michaelneale:
I often see errors like :

```
32603: Mcp error: -32603: The path '/Users/micn/Documents/code/goose' does not exist or is not a file.
```

This is easily resolved by detecting if a path is a file or not, and returning something meaningful (can save a round trip to tools) improving performance


alexhancock:
Good one!

PR #4702, opened 2025-09-23
prefer users SHELL

michaelneale:
This is one option - use the $SHELL value and delegate to that. That allows it to load appropriately if zsh etc, fall back to bash


jamadeo:
We could totally do this, but I could see some downsides. The model might expect to be using bash specifically, unless we update system prompts/tool descriptions. So if it isn't instructed to do so specifically, it could generate invalid shell commands thinking it's bash

michaelneale:
@jamadeo oh... hrm... that is a great point. zsh is kind of a superset of bash so I expect would work the same. Conversely, bash on macos is getting old (bash 3.2 when linux is > 4) and is lagging modern bash so it could go the other way where it will try to do things that won't work (especially if it ends up deprecated), not that I Have seen that. 

so if zsh is a reasonable updated shell, then probably better to do this (but then also is a change). If it is not, probably better to stay with bash. Zsh is a posix shell which I would think LLMs are mostly trained with (due to linux). 

We don't prompt either way I believe, so the LLM will just see "shell" and think whatever is most idiomatic on that platform, which could go either way I guess (and probably drift over time).

jamadeo:
I looked at how claude code works, and while its tool is always called bash, it does use zsh on my system. Presumably it's using $SHELL then? So, yeah, let's go for it. Maybe though, should we tell the model in the tool description which shell executable it will use? I suppose if it wants to it can always specify by running `shell("bash -c '...')`

michaelneale:
I think it is enough to say it says to use the system SHELL - it should be able to work that out.

jamadeo (approved):

DOsinga:
we could still inject it into the system prompt, yeah?

michaelneale:
yeah - not sure if it is necessary, as it can work it out itself as it goes if it needs to?

michaelneale:
@DOsinga follow on for that: https://github.com/block/goose/pull/4807 - which can work, not sure if really needed though

PR #4704, opened 2025-09-23
fix: adjust nightly builds to first push a tag

alexhancock:
```
Error: Error 422: Validation Failed: {"resource":"Release","code":"custom","field":"pre_receive","message":"pre_receive Repository rule violations found\n\nCannot create ref due to creations being restricted.\n\n"}, {"resource":"Release","code":"custom","message":"Published releases must have a valid tag"}
```

Attempt at fixing the above error that happens in nightly builds at present

Example: https://github.com/block/goose/actions/runs/17936231221/job/51003691059

This also allows manual testing of the workflow so I can diagnose if anything still doesn't work on a branch after this change merges.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -5,6 +5,12 @@ on:
   schedule:
     # Run at midnight US Eastern (0500 UTC)
     - cron: '0 5 * * *'
+  workflow_dispatch:  # Allow manual triggering
+    inputs:
+      branch:
+        description: 'Branch to build from'
+        required: false
+        default: 'main'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -73,8 +82,31 @@ jobs:
...
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # pin@v4
+        with:
+          ref: ${{ github.event.inputs.branch || github.ref }}
+          fetch-depth: 0  # Fetch all history for proper tagging
+
+      - name: Create tag
+        run: |
+          TAG="${{ needs.prepare-version.outputs.version }}"
+
+          # Validate version format (expects format like: x.y.z-nightly.YYYYMMDD)
```
jamadeo: seems a little out of place, if we think it could be invalid shouldn't it validate in the step where it is created?

.github/workflows/nightly.yml
```diff
@@ -73,8 +82,31 @@ jobs:
...
+          if [[ ! "$TAG" =~ ^[0-9]+\.[0-9]+\.[0-9]+-nightly\.[0-9]{8}$ ]]; then
+            echo "Error: Invalid version format: $TAG"
+            exit 1
+          fi
+
+          echo "Creating tag: $TAG"
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+          # Force create and push tag (overwrites if exists)
```
jamadeo: is the forcing necessary? consider if we can have that off at least by default

PR #4772, opened 2025-09-23
chore(mcp): convert computercontroller server to use the rust sdk

alexhancock:
The final built in MCP server conversion from our old internal mcp crates to https://github.com/modelcontextprotocol/rust-sdk. This also removes a couple modules which were simply routers to different mcp servers, and replaces it with a single module shared between `goose-cli` and `goose-server`.


crates/goose/src/mcp_server_runner.rs
```diff
@@ -0,0 +1,45 @@
...
+pub async fn run_mcp_server(name: &str) -> Result<()> {
+    if name == "googledrive" || name == "google_drive" {
+        return Err(anyhow!(
+            "the built-in Google Drive extension has been removed"
+        ));
+    }
+
+    tracing::info!("Starting MCP server");
+
+    match name {
```
michaelneale: we were missing this previously?

crates/goose/Cargo.toml
```diff
@@ -101,6 +101,7 @@ unicode-normalization = "0.1"
 
 arrow = "52.2"
 oauth2 = "5.0.0"
+goose-mcp = { version = "1.8.0", path = "../goose-mcp" }
```
michaelneale: hrm - there was https://github.com/block/goose/pull/4681 change by @jamadeo  which IIUC - this would end up circular hrm. Not sure what direction dependency would be but this seems the wrong way?

crates/goose/src/mcp_server_runner.rs
michaelneale: ah this is the replacement for run_server command

crates/goose/Cargo.toml
michaelneale: ie the core shouldn't depend on mcp but other things should depend on core?

crates/goose/src/mcp_server_runner.rs
alexhancock: There were two copies of similar files, both with slightly different logic. One in `cli` and one in `server`. I am just deduping and improving here, but I should find a final home for it that keeps the dependency direction good. Will make an adjustment.

michaelneale:
@alexhancock nice - I tested it with docx, xlsx and pdf etc - all seems to perform same as previous. 

however, now with main merged, if you update to main and try to compile, as expected: 

```
error: cyclic package dependency: package `goose v1.8.0 (/Users/micn/Documents/code/goose/crates/goose)` depends on itself. Cycle:
package `goose v1.8.0 (/Users/micn/Documents/code/goose/crates/goose)`
    ... which satisfies path dependency `goose` (locked to 1.8.0) of package `goose-mcp v1.8.0 (/Users/micn/Documents/code/goose/crates/goose-mcp)`
    ... which satisfies path dependency `goose-mcp` (locked to 1.8.0) of package `goose v1.8.0 (/Users/micn/Documents/code/goose/crates/goose)`
error: Recipe `release-binary` failed on line 10 with exit code 101
error: Recipe `make-ui` failed on line 218 with exit code 101
```

so will need to solve that before merging it in (doesn't show up as a merge conflict but will fail), but other than that, good to go!

crates/goose/src/mcp_server_runner.rs
```diff
@@ -0,0 +1,45 @@
+use anyhow::{anyhow, Result};
```
jamadeo: I think you can delete this file now right? you moved it to goose-mcp?

jamadeo (approved):

crates/goose/src/mcp_server_runner.rs
alexhancock: ah yes, intended to!

PR #4773, opened 2025-09-23
Remove now unused mcp-server crate

alexhancock:
This is no longer used after all the servers in `goose-mcp` have been converted to use the [rust-sdk](https://github.com/modelcontextprotocol/rust-sdk) instead of our old internal crate


michaelneale:
wooh!

PR #4784, opened 2025-09-23
Add tests for create / edit recipe

zanesq:
## Pull Request Description
Follow up for https://github.com/block/goose/pull/4693 adding tests for create / edit recipe forms


ui/desktop/src/components/recipes/shared/RecipeFormFields.tsx
```diff
@@ -324,6 +328,12 @@ export default function RecipeFormFields({
               };
               field.handleChange([...field.state.value, newParam]);
               setNewParameterName('');
+              // Expand the newly added parameter by default
```
zanesq, line 331: minor ux improvement to expand the newly added parameter after added

PR #4788, opened 2025-09-24
feat(shell): throw errors on interactive commands

exitcode0:
## Description

Configures the developer / shells to throw errors when using unsupported interactive commands

Commands like `git commit` (without -m) & `git rebase -i` would previously timeout or need canceling manually

This is change is a bit of a hack and we should revisit this in the future
🤞 we can figure out how to set `EDITOR` up to call the file-edit tool


DOsinga:
> This is change is a bit of a hack and we should revisit this in the future

what do you mean by that? how would we revisit this?

crates/goose-mcp/src/developer/tests/test_shell_env.rs
```diff
@@ -0,0 +1,647 @@
+#[cfg(test)]
```
DOsinga: delete this file; it is 647 lines of trivial testing. the only reason these tests are going to fail is if we decide to change the text of the environment variables.

exitcode0: done, the test file should be gone now

exitcode0:
> > This is change is a bit of a hack and we should revisit this in the future
> 
> what do you mean by that? how would we revisit this?

My thinking is that it might be possible to set `EDITOR` up in such a way that we hand off to the current file editor tool or something similar
because git is only checking the exit code of the processed called from `EDITOR`, we could hand off to a file editor tool, we could then use that to complete whatever edits are required, then we exit and git / the shell MCP takes back over
That's still entirely theoretical though, I haven't looked at any of that code, i'm also not sure if MCP supports something like that

michaelneale:
oh @exitcode0 that is fantastic, yeah that should help a lot of frustrating hanging.

DOsinga:
>My thinking is that it might be possible to set EDITOR up in such a way that we hand off to the current file editor tool or something similar because git is only checking the exit code of the processed called from EDITOR, we could hand off to a file editor tool, we could then use that to complete whatever edits are required, then we exit and git / the shell MCP takes back over That's still entirely theoretical though, I haven't looked at any of that code, i'm also not sure if MCP supports something like that

those are good musings, but probably drop the comment

PR #4792, opened 2025-09-24
rmcp upgrade

jamadeo:
Bumps RMCP and improves the oauth flow for HTTP MCPs


crates/goose-server/src/openapi.rs
```diff
@@ -430,9 +407,9 @@ impl<'__s> ToSchema<'__s> for AnnotatedSchema {
         TextContentSchema,
         RawTextContentSchema,
         RawImageContentSchema,
+        RawAudioContentSchema,
         RawEmbeddedResourceSchema,
         RawResourceSchema,
-        AnnotatedSchema,
```
alexhancock, line 424: is this just not used in the frontend?

jamadeo: in prior versions of rmcp, the audiocontent enum variant was the annotated type, not the raw type (fixed here: https://github.com/modelcontextprotocol/rust-sdk/commit/1b70f5b0320d1984a7e2725ef789d87b6de04130#diff-289d941b1f937b4f5adddf8e939caa1a7e93b41e09afa6517bbee4238a5dff5cL69) which made our schema gen a bit weird, but now that it's a RawAudioContent we don't need this

PR #4800, opened 2025-09-24
Final Internal MCP Crate Cleanup

alexhancock:
* Fully gets rid of `mcp-client` and `mcp-core`, cleaning up our mcp tooling to just be the official rust sdk
* Removes concept of `ToolCall` and replaces it with the similar `CallToolRequestParam` from `rmcp`

**TODOs**
- [ ]  Look into McpClientTrait and see if we could remove this abstraction
- [ ] Get integration tests passing


alexhancock:
@jamadeo how do you feel about the fact that this removes `mcp-client` but still `impl McpClientTrait for McpClient`s

Do you think we should unwind that in a change where we remove those crates, or ok to do in a followup?

Also I realize this is a big PR so can sync on it when I've verified it's ready for review otherwise.

crates/goose-cli/src/commands/acp.rs
```diff
@@ -285,24 +285,28 @@ impl GooseAcpAgent {
...
                     // Try to extract the path from the arguments
                     let args = &tool_call.arguments;
-                    if let Some(path_str) = args.get("path").and_then(|p| p.as_str()) {
-                        locs.push(acp::ToolCallLocation {
-                            path: path_str.into(),
-                            line: Some(1),
-                            meta: None,
-                        });
+                    if let Some(args_map) = args {
+                        if let Some(path_str) = args_map.get("path").and_then(|p| p.as_str()) {
```
jamadeo: nit: could be a single `if let Some(..` with a `.and_then`

crates/goose-cli/src/commands/acp.rs
```diff
@@ -285,24 +285,28 @@ impl GooseAcpAgent {
...
+                            let path = std::path::PathBuf::from(path_str);
+                            if path.exists() && path.is_file() {
+                                locs.push(acp::ToolCallLocation {
+                                    path: path_str.into(),
+                                    line: None,
```
jamadeo: this is `line: Some(1)` on the left but None here, is that deliberate?

crates/goose-cli/src/commands/web.rs
```diff
@@ -459,8 +460,13 @@ async fn process_message_streaming(
                                                 serde_json::to_string(
                                                     &WebSocketMessage::ToolRequest {
                                                         id: req.id.clone(),
-                                                        tool_name: tool_call.name.clone(),
-                                                        arguments: tool_call.arguments.clone(),
+                                                        tool_name: tool_call
+                                                            .name
+                                                            .to_string()
+                                                            .clone(),
```
jamadeo: I don't think you need to .clone() here since .to_string() should return an owned value

crates/goose-cli/src/session/export.rs
```diff
@@ -127,17 +127,19 @@ pub fn tool_request_to_markdown(req: &ToolRequest, export_all_content: bool) ->
             ));
             md.push_str("**Arguments:**\n");
 
-            match call.name.as_str() {
+            match call.name.to_string().as_str() {
```
jamadeo: `match call.name.as_ref()`

crates/goose-cli/src/session/mod.rs
```diff
@@ -1050,11 +1050,10 @@ impl CliSession {
                                                     }
                                                 })
                                             })
-                                            .unwrap_or_else(|| "unknown".to_string());
-
-                                        let success = tool_response.tool_result.is_ok();
-                                        let result_status = if success { "success" } else { "error" };
+                                            .unwrap_or_else(|| "tool".to_string().into());
```
jamadeo: did you mean to change this from "unknown" to "tool"?

crates/goose-cli/src/session/output.rs
```diff
@@ -390,84 +389,94 @@ pub fn render_builtin_error(names: &str, error: &str) {
...
-                                    // For objects in arrays, print them with indentation
-                                    print!("            - ");
-                                    print_params(item, 3, debug);
-                                } else {
-                                    println!(
-                                        "            - {}",
-                                        style(format!("{}", item)).green()
-                                    );
+    if let Some(args) = &call.arguments {
+        if let Some(Value::Array(task_parameters)) = args.get("task_parameters") {
```
jamadeo: could chain these two as well

crates/goose/src/providers/formats/snowflake.rs
```diff
@@ -181,16 +181,24 @@ pub fn parse_streaming_response(sse_data: &str) -> Result<Message> {
...
-            let input_value = serde_json::from_str::<Value>(&tool_input)
-                .unwrap_or_else(|_| Value::String(tool_input.clone()));
-            let tool_call = ToolCall::new(name, input_value);
-            message = message.with_tool_request(id, Ok(tool_call));
-        } else if tool_name.is_some() {
-            // Tool with no input - use empty object
-            let tool_call = ToolCall::new(name, Value::Object(serde_json::Map::new()));
-            message = message.with_tool_request(id, Ok(tool_call));
+    if let Some(id) = tool_use_id {
+        if let Some(name) = tool_name {
```
jamadeo: could chain these to remove a nesting level

jamadeo:
> @jamadeo how do you feel about the fact that this removes `mcp-client` but still `impl McpClientTrait for McpClient`s
> 
> Do you think we should unwind that in a change where we remove those crates, or ok to do in a followup?
> 
> Also I realize this is a big PR so can sync on it when I've verified it's ready for review otherwise.

I think that's fine, if there's some point in the future where it makes sense to remove a layer of abstraction, we can always do that.

crates/goose-cli/src/session/mod.rs
alexhancock: good catch - thank you

PR #4801, opened 2025-09-24
Aning/plattools

angelahning:
## Pull Request Description

---

**Email**:

### DIFF
--- a/crates/goose-cli/src/commands/web.rs
+++ b/crates/goose-cli/src/commands/web.rs
@@ -103,7 +103,12 @@ pub async fn handle_web(port: u16, host: String, open: bool) -> Result<()> {
     let model_config = goose::model::ModelConfig::new(&model)?;
 
     // Create the agent
-    let agent = Agent::new();
+    let mut agent = Agent::new();
+    
+    // Initialize platform tools
+    agent.initialize_platform_tools().await
+        .expect("Failed to initialize platform tools");
+    
     let provider = goose::providers::create(&provider_name, model_config)?;
     agent.update_provider(provider).await?;
 
--- a/crates/goose-cli/src/session/builder.rs
+++ b/crates/goose-cli/src/session/builder.rs
@@ -210,7 +210,11 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
         .with_temperature(temperature);
 
     // Create the agent
-    let agent: Agent = Agent::new();
+    let mut agent: Agent = Agent::new();
+    
+    // Initialize platform tools
+    agent.initialize_platform_tools().await
+        .expect("Failed to initialize platform tools");
 
     if let Some(sub_recipes) = session_config.sub_recipes {
         agent.add_sub_recipes(sub_recipes).await;
--- a/crates/goose-server/src/commands/agent.rs
+++ b/crates/goose-server/src/commands/agent.rs
@@ -32,7 +32,11 @@ pub async fn run() -> Result<()> {
     let secret_key =
         std::env::var("GOOSE_SERVER__SECRET_KEY").unwrap_or_else(|_| "test".to_string());
 
-    let new_agent = Agent::new();
+    let mut new_agent = Agent::new();
+    
+    // Initialize platform tools
+    new_agent.initialize_platform_tools().await
+        .expect("Failed to initialize platform tools");
 
     // Only initialize provider and extensions when running in standalone goosed mode
     // This prevents breaking the Electron app which manages its own provider setup
--- a/crates/goose-server/src/state.rs
+++ b/crates/goose-server/src/state.rs
@@ -54,7 +54,11 @@ impl AppState {
 
     pub async fn reset(&self) {
         let mut agent = self.agent.write().await;
-        let new_agent = Agent::new();
+        let mut new_agent = Agent::new();
+        
+        // Initialize platform tools
+        new_agent.initialize_platform_tools().await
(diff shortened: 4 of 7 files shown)


crates/goose-cli/src/commands/web.rs
```diff
@@ -103,7 +103,12 @@ pub async fn handle_web(port: u16, host: String, open: bool) -> Result<()> {
     let model_config = goose::model::ModelConfig::new(&model)?;
 
     // Create the agent
-    let agent = Agent::new();
+    let mut agent = Agent::new();
+    
+    // Initialize platform tools
+    agent.initialize_platform_tools().await
+        .expect("Failed to initialize platform tools");
```
DOsinga, line 110: remove the comment. also, do we need to fail here?

crates/goose-cli/src/session/builder.rs
```diff
@@ -210,7 +210,11 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
         .with_temperature(temperature);
 
     // Create the agent
-    let agent: Agent = Agent::new();
+    let mut agent: Agent = Agent::new();
+    
+    // Initialize platform tools
+    agent.initialize_platform_tools().await
+        .expect("Failed to initialize platform tools");
```
DOsinga, line 217: same

crates/goose-server/src/commands/agent.rs
```diff
@@ -32,7 +32,11 @@ pub async fn run() -> Result<()> {
     let secret_key =
         std::env::var("GOOSE_SERVER__SECRET_KEY").unwrap_or_else(|_| "test".to_string());
 
-    let new_agent = Agent::new();
+    let mut new_agent = Agent::new();
+    
+    // Initialize platform tools
+    new_agent.initialize_platform_tools().await
+        .expect("Failed to initialize platform tools");
```
DOsinga, line 39: you should sync with main, this is no longer how it works

crates/goose-server/src/state.rs
```diff
@@ -54,7 +54,11 @@ impl AppState {
 
     pub async fn reset(&self) {
         let mut agent = self.agent.write().await;
-        let new_agent = Agent::new();
+        let mut new_agent = Agent::new();
+        
+        // Initialize platform tools
+        new_agent.initialize_platform_tools().await
+            .expect("Failed to initialize platform tools");
```
DOsinga, line 61: same. here we should definitely not die

crates/goose/examples/agent.rs
```diff
@@ -16,7 +16,12 @@ async fn main() {
     let provider = Arc::new(DatabricksProvider::default());
 
     // Setup an agent with the developer extension
-    let agent = Agent::new();
+    let mut agent = Agent::new();
+    
+    // Initialize platform tools
+    agent.initialize_platform_tools().await
+        .expect("Failed to initialize platform tools");
+
```
DOsinga, line 24: do we need that here?

crates/goose/src/agents/agent.rs
```diff
@@ -820,8 +751,6 @@ impl Agent {
         if extension_name.is_none() || extension_name.as_deref() == Some("platform") {
             // Add platform tools
```
DOsinga, line 752: this doesnt seem to be true anymore

crates/goose/src/agents/platform_tools.rs
```diff
@@ -154,3 +169,360 @@ pub fn manage_schedule_tool() -> Tool {
...
+    }
+
+    async fn subscribe(&self) -> mpsc::Receiver<ServerNotification> {
+        // Default implementation - to be implemented later
+        let (_tx, rx) = mpsc::channel(1);
+        rx
+    }
+
+    fn get_info(&self) -> Option<&InitializeResult> {
+        // Default implementation - to be implemented later
```
DOsinga, line 525: you have this comment a few times, I don't think you mean that

DOsinga (commented):
this feels like a good start but I think we can push on a bit to make this more complete.

platform tools needs to be even more divorced from agent.  we still have in agent.rs:

self.platform_tools <- that doesn't seem to be used. good. drop it

in list_tools:
if extension_name.is_none() || extension_name.as_deref() == Some("platform") {

that shoudn't be needed, extension manageer should take care of that for the actual platform tools

self.extension_manager
            .add_client(
                "platform".to_string(),
                config,
                Arc::new(tokio::sync::Mutex::new(Box::new(PlatformTools::new(self.extension_manager.clone(), self.tool_route_manager.clone())) as Box<dyn McpClientTrait>)),
                None, // no server info
                None, // no temp dir
            )
            .await;
            
we shouldn't need that; whatever is setting up the defaults for extensions, should also setup the platform tools so that they can be enabled and disabled.

there's also this mystery block:
                                    let mut enable_extension_request_ids = vec![];
                                    for request in &remaining_requests {
                                        if let Ok(tool_call) = &request.tool_call {
                                            if tool_call.name == PLATFORM_MANAGE_EXTENSIONS_TOOL_NAME {
                                                enable_extension_request_ids.push(request.id.clone());
                                            }
                                        }
                                    }


PLATFORM_MANAGE_SCHEDULE_TOOL_NAME is defined in platform_tools but not handled there. should probably go into a separate thing.

DOsinga:
what do you want to do with this in light of the new world? continue or this or move to a different PR. I think the infra is mostly ready for it now

angelahning:
Another PR coming!

PR #4551, opened 2025-09-07
fix: Fix cell coordinate ordering in XlsxTool and add unit tests

kita-sato:
fix: https://github.com/block/goose/issues/4550

Correct swapped row/col indices in get_cell, get_cell_mut, and related methods. Added tests for B1 and A2 cells to verify proper coordinate handling.

see: https://docs.rs/umya-spreadsheet/2.3.2/src/umya_spreadsheet/structs/worksheet.rs.html#240

PR #4803, opened 2025-09-24
Revert "fix: Fix cell coordinate ordering in XlsxTool and add unit tests"

michaelneale:
Reverts block/goose#4551

this one needs updating to main before the tests will pass


michaelneale:
FYI still checking this...

michaelneale:
I think fix is valid, tests are not. closing this.

PR #4805, opened 2025-09-25
fix: keep one goosed client per BrowswerWindow

jamadeo:
If you refreshed, you'd get the last goosed created, not the one that went with your window


ui/desktop/src/main.ts
```diff
@@ -603,6 +604,17 @@ const createChat = async (
...
+  const goosedClient = createClient(
+    createConfig({
+      baseUrl: `http://127.0.0.1:${port}`,
+      headers: {
+        'Content-Type': 'application/json',
+        'X-Secret-Key': SERVER_SECRET,
+      },
+    })
+  );
+  goosedClients.set(mainWindow.id, goosedClient);
```
alexhancock, line 616: given `goosedClients` is also created in this proc, will it only ever hold one?

alexhancock (approved):

ui/desktop/src/main.ts
jamadeo: goosedClients is created globally in the electron node process, so it will be a singleton holding one client per BrowswerWindow

alexhancock: ah, right. thanks!

PR #4811, opened 2025-09-25
Revert "feat: update aws-sdk-bedrockruntime to enable AWS_BEARER_TOKEN_BEDROCK auth"

michaelneale:
Reverts block/goose#4327

this won't build on macos - can we revert this change until we know what proper dependencies to pin?


alexhancock (approved):

michaelneale:
closing as rolled forward which should do better

PR #4813, opened 2025-09-25
feat: fast local mini model embedded provider

michaelneale:
This uses a self downloading model and a very lean provider for local only use (this is both as a provider, and future utility). 

Currently tested with qwen - also some gemma models

todo: 

- [ ] curate list of 3 models of 3 sizes that work
- [ ] make message of downloading clear

### DIFF
--- a/crates/goose/src/providers/embedded.rs
+++ b/crates/goose/src/providers/embedded.rs
@@ -22,14 +22,38 @@ use crate::model::ModelConfig;
 pub const EMBEDDED_DEFAULT_MODEL: &str = "qwen2.5-7b-instruct";
 pub const EMBEDDED_KNOWN_MODELS: &[&str] = &[
     "qwen2.5-7b-instruct",
+    "gemma-3n-e4b-it",
     "llama-7b",
     "codellama-7b",
     "mistral-7b",
 ];
 
-const DEFAULT_MODEL_URL: &str =
-    "https://huggingface.co/Qwen/Qwen2.5-7B-Instruct-GGUF/resolve/main/qwen2.5-7b-instruct-q3_k_m.gguf";
-const DEFAULT_MODEL_SIZE_MB: u64 = 3631;
+// Model registry with download information
+struct ModelInfo {
+    name: &'static str,
+    display_name: &'static str,
+    url: &'static str,
+    size_mb: u64,
+    #[allow(dead_code)] // Reserved for future use to determine model format
+    model_type: &'static str,
+}
+
+const KNOWN_MODELS_INFO: &[ModelInfo] = &[
+    ModelInfo {
+        name: "qwen2.5-7b-instruct",
+        display_name: "Qwen 2.5 7B Instruct (Q3_K_M, ~3.5GB)",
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/embedded.rs
```diff
@@ -0,0 +1,631 @@
...
+        
+        let threads: u32 = config
+            .get_param("EMBEDDED_THREADS")
+            .unwrap_or(8);  // Match g3's thread count
+        
+        // Check if tools should be enabled (similar to claude-cli's permission mode)
+        let goose_mode: String = config
+            .get_param("GOOSE_MODE")
+            .unwrap_or_else(|_| "manual".to_string());
+        let enable_tools = goose_mode == "auto";
```
DOsinga: != "chat"

crates/goose/src/providers/embedded.rs
```diff
@@ -0,0 +1,631 @@
...
+        // Create response message
+        let message = Message::new(
+            Role::Assistant,
+            chrono::Utc::now().timestamp(),
+            vec![MessageContent::text(generated_text.clone())],
+        );
+        
+        // Estimate token usage
+        let prompt_tokens = (prompt.len() / 4) as i32;
+        let completion_tokens = (generated_text.len() / 4) as i32;
```
DOsinga: we could use token_counter.rs

crates/goose/src/providers/embedded.rs
```diff
@@ -0,0 +1,631 @@
...
+    #[tracing::instrument(
+        skip(self, model_config, system, messages),
+        fields(model_config, input, output, input_tokens, output_tokens, total_tokens)
+    )]
+    async fn complete_with_model(
+        &self,
+        model_config: &ModelConfig,
+        system: &str, // Ignored - we use g3's system prompt for embedded
+        messages: &[Message],
+        _tools: &[Tool], // Ignored - embedded provider doesn't use tool definitions
```
DOsinga, line 685: going forward, is this an option? it seems through ollama we'd get somewhere

DOsinga (commented):
I know this is experimental, but am interested!

michaelneale:
yeah this doesn't do normal tool calling, but is lean and fast.

crates/goose/src/providers/embedded.rs
michaelneale: yeah - but only for larger models, much much larger (which we can already do via ollama) so could find a way to consolidate with llama-serve or something to layer on the tool calling schema (but it is not in llama cpp out of the box), and certainly won't work with gooses tools and prompts other than if you start with very large ones.

michaelneale: (I think have to add in BNF grammars and stuff if we wanted to do out of the box what ollama et al do? valuable?)

michaelneale: ollama seems to cope, but others struggle due to the system prompt+tools now, so if we want ot to run lean, really need to strip them out somehow (which is what this does in a brute force way). maybe time to bring back some toolshim approach.

alexhancock:
I'm interested in how this is going. Is there example config you can share with the best of the models you've tried?

PR #4816, opened 2025-09-25
fix: missing cmake and protobuf for windows build

vlascik:
## Pull Request Description
Fix: install missing cmake and protobuf to docker image for windows build, to align with other platforms. 

Build is broken otherwise.


Justfile
```diff
@@ -41,7 +41,7 @@ release-windows:
...
+        powershell.exe -Command "docker volume create goose-windows-cache; docker run --rm -v ${PWD}:/usr/src/myapp -v goose-windows-cache:/usr/local/cargo/registry -w /usr/src/myapp rust:latest sh -c 'rustup target add x86_64-pc-windows-gnu && apt-get update && apt-get install -y mingw-w64 protobuf-compiler cmake && export CC_x86_64_pc_windows_gnu=x86_64-w64-mingw32-gcc && export CXX_x86_64_pc_windows_gnu=x86_64-w64-mingw32-g++ && export AR_x86_64_pc_windows_gnu=x86_64-w64-mingw32-ar && export CARGO_TARGET_X86_64_PC_WINDOWS_GNU_LINKER=x86_64-w64-mingw32-gcc && export PKG_CONFIG_ALLOW_CROSS=1 && export PROTOC=/usr/bin/protoc && export PATH=/usr/bin:\$PATH && protoc --version && cargo build --release --target x86_64-pc-windows-gnu && GCC_DIR=\$(ls -d /usr/lib/gcc/x86_64-w64-mingw32/*/ | head -n 1) && cp \$GCC_DIR/libstdc++-6.dll /usr/src/myapp/target/x86_64-pc-windows-gnu/release/ && cp \$GCC_DIR/libgcc_s_seh-1.dll /usr/src/myapp/target/x86_64-pc-windows-gnu/release/ && cp /usr/x86_64-w64-mingw32/lib/libwinpthread-1.dll /usr/src/myapp/target/x86_64-pc-windows-gnu/release/'"
```
alexhancock: can we take this opportunity to break this out to different lines so it's more readable what's going on here?

also, a comment explaining what each thing added is for would help a lot with review

vlascik: So the original problem was that the windows build command `apt-get install -y mingw-w64 protobuf-compiler cmake`... was missing cmake and protobuf-compiler when started from powershell, and not `sh`. Apparently someone added it to `sh`, but forgot to put it to the `powershell` part, so if the build was started from powershell, it didn't work.

I extracted the command to a variable, this way it's in one place and people won't have to remember to make the changes twice.

alexhancock (approved):

Justfile
alexhancock: I see. Thanks for the clarification. Looks like there is a conflict to work out

I will see if I can run a full build of this before merge to verify everything works well for our automated windows builds.

vlascik:
So, there was a different problem in mod.rs file, I tried to be clever with the branching of the fixes, but someone else fixed it in the meantime, and it seems I messed this branch up royally. 

I'll close this and post the fix in a new PR, I'm too tired to try to unwind this mess. Sorry 😄.

alexhancock:
No problem. Ping me on the new PR

PR #4818, opened 2025-09-25
fix: a few things with the mcp snapshot test

jamadeo:
In the test itself:
- call add_extension in the result-capturing block so we can print errors when that fails
- only print errors in replay mode (errors are mismatches from the replayer, so they aren't relevant in record mode)

In the tests cases:
- add `--quiet` to the developer mcp cargo run so we don't have build output in the snapshot
- remove a test that uses a .png that doesn't exist in the repo

And finally, re-record the snapshots.


crates/goose/tests/mcp_replays/uvxmcp-server-fetch
```diff
@@ -1,5 +1,5 @@
...
-STDOUT: {"jsonrpc":"2.0","id":0,"result":{"protocolVersion":"2025-03-26","capabilities":{"experimental":{},"prompts":{"listChanged":false},"tools":{"listChanged":false}},"serverInfo":{"name":"mcp-fetch","version":"1.12.4"}}}
+STDOUT: {"jsonrpc":"2.0","id":0,"result":{"protocolVersion":"2025-03-26","capabilities":{"experimental":{},"prompts":{"listChanged":false},"tools":{"listChanged":false}},"serverInfo":{"name":"mcp-fetch","version":"1.14.1"}}}
 STDIN: {"jsonrpc":"2.0","method":"notifications/initialized"}
 STDIN: {"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"_meta":{"progressToken":0},"name":"fetch","arguments":{"url":"https://example.com"}}}
-STDOUT: {"jsonrpc":"2.0","id":1,"result":{"content":[{"type":"text","text":"Contents of https://example.com/:\nThis domain is for use in illustrative examples in documents. You may use this\ndomain in literature without prior coordination or asking for permission.\n\n[More information...](https://www.iana.org/domains/example)"}],"isError":false}}
+STDOUT: {"jsonrpc":"2.0","id":1,"result":{"content":[{"type":"text","text":"Failed to fetch robots.txt https://example.com/robots.txt due to a connection issue"}],"isError":true}}
```
alexhancock, line 5: Is this one (and the below) intended?

jamadeo: yeah I noticed this one ended up with an error. It doesn't really matter though, given the contents are in the same format, and this is really just checking that messages are exchanged properly between Goose and a python server

PR #4822, opened 2025-09-25
Fix: Token count UI doesn't re-render if it's open.

katzdave:
Tracking the key forces re-renders every time the key changes (based on numTokens/limit).

This now works even if you constantly mouse over the UI.


ui/desktop/src/components/ChatInput.tsx
```diff
@@ -572,6 +572,7 @@ export default function ChatInput({
         },
         compactIcon: <ScrollText size={12} />,
         autoCompactThreshold: autoCompactThreshold,
+        key: `context-${numTokens}-${tokenLimit}`, // Force re-render on token count change
```
DOsinga: this doesn't strike me as the correct solution, at most a work around? why are we not rerendering this when the numTokens changes? it is in the alert properties. do we do a shallow compare somewhere?

ui/desktop/src/components/bottom_menu/BottomMenuAlertPopover.tsx
```diff
@@ -107,7 +107,12 @@ export default function BottomMenuAlertPopover({ alerts }: AlertPopoverProps) {
     // Find new or changed alerts
     const changedAlerts = alerts.filter((alert, index) => {
       const prevAlert = previousAlertsRef.current[index];
-      return !prevAlert || prevAlert.type !== alert.type || prevAlert.message !== alert.message;
+      return (
+        !prevAlert ||
+        prevAlert.type !== alert.type ||
+        prevAlert.message !== alert.message ||
+        prevAlert.key !== alert.key
```
DOsinga: ah, I see, the problem is that we don't detect changes in the right way. this should just do a deep compare I think

DOsinga (approved):

PR #4823, opened 2025-09-25
Desktop UI shows all messages (even prior to compaction)

katzdave:
Older messages are italicized to make it clear they're from prior to summarization.

Summarization markers now contain summary.

https://github.com/user-attachments/assets/a6dff550-5a11-4cd1-b621-f7f585e04fd5

### DIFF
--- a/ui/desktop/src/components/BaseChat.tsx
+++ b/ui/desktop/src/components/BaseChat.tsx
@@ -441,7 +441,7 @@ function BaseChatContent({
                             onClick={async () => {
                               clearError();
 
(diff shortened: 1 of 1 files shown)


DOsinga:
(can you add a screenshot before undrafting?)

katzdave:
> (can you add a screenshot before undrafting?)

uploaded a video if you have any ui thoughts; need to clean this up more and get it working with new session model.

DOsinga:
I feel like I keep asking this, but is this stale/replaced by something else?

DOsinga:
closing this as we've implemented this in a different way

PR #4825, opened 2025-09-25
feat(cli): add GOOSE_DEBUG environment variable support

sam-at-block:
Add support for GOOSE_DEBUG environment variable to enable debug mode without requiring the --debug CLI flag. Follows existing precedence pattern where CLI flags take priority over environment variables.

- CLI --debug flag overrides GOOSE_DEBUG env var
- GOOSE_DEBUG=true shows full tool parameters without truncation
- Defaults to false if neither CLI flag nor env var is set
- Uses same config.get_param() pattern as GOOSE_PROVIDER/GOOSE_MODEL

### DIFF
--- a/crates/goose-cli/src/session/builder.rs
+++ b/crates/goose-cli/src/session/builder.rs
@@ -395,11 +395,13 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> CliSession {
             }
         });
 
+    let debug_mode = session_config.debug || config.get_param("GOOSE_DEBUG").unwrap_or(false);
+
     // Create new session
     let mut session = CliSession::new(
         Arc::try_unwrap(agent_ptr).unwrap_or_else(|_| panic!("There should be no more references")),
         session_id.clone(),
-        session_config.debug,
+        debug_mode,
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/session/builder.rs
```diff
@@ -416,11 +416,14 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> Session {
             }
         });
 
+    // Determine debug mode: CLI flag takes precedence over GOOSE_DEBUG environment variable
+    let debug_mode = session_config.debug || config.get_param("GOOSE_DEBUG").unwrap_or(false);
```
DOsinga, line 398: remove the comment

DOsinga:
hi @sam-at-block - do you still want to get this in? if you remove the comment & sync to main, we can make it happen

crates/goose-cli/src/session/builder.rs
sam-at-block: Hi @DOsinga , rebased in the latest upstream/main and removed the comment :)

PR #4827, opened 2025-09-25
chore: fix all warnings in test compilation

alexhancock:
Removing unused code, and adding test cases where there was unused code that should have been used

### DIFF
--- a/crates/goose/tests/agent.rs
+++ b/crates/goose/tests/agent.rs
@@ -235,6 +235,16 @@ mod tests {
     }
 
     #[tokio::test]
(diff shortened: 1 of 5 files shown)


crates/goose/tests/agent.rs
```diff
@@ -234,6 +234,16 @@ mod tests {
         .await
     }
 
+    #[tokio::test]
+    async fn test_agent_with_anthropic() -> Result<()> {
+        run_test_with_config(TestConfig {
+            provider_type: ProviderType::Anthropic,
```
alexhancock, line 240: `ProviderType::Anthropic` was previously unused in the file

PR #4828, opened 2025-09-25
use agent manager for subagent

yingjiehe-xyz:
## Pull Request Description

Use the agent manager added in https://github.com/block/goose/pull/4684 for the subagent task execution

Test:

```
( O)> use the subagent to summarize my slacks channels g2-help

I'll use a subagent to summarize your Slack channel `g2-help`. Let me create a task for this.
─── create_task | dynamic_task ──────────────────────────
task_parameters:
    -
        instructions: Summarize the Slack channel 'g2-help'. Get recent messages from this channel and provide a comprehensive summary of the discussions, key topics, common issues, and any important information shared. Focus on the most recent activity and highlight any recurring themes or important announcements.
        title: Summarize g2-help Slack Channel
        extensions:
            - slack
        return_last_only: true


─── execute_task | subagent ──────────────────────────
task_ids: 1a070d51-3d2d-4f57-8c19-cad83b66418d

# Summary of G2-Help Slack Channel

Based on the recent messages from the **g2-help** channel, here's a comprehensive summary of the discussions, key topics, and common issues:

## **Key Topics & Recurring Themes**

### **1. Integration Requests & MCP Extensions**
- **Zendesk Integration**: Request for Zendesk integration for Bitkey's customer service management
- **Sentry Integration**: On the roadmap and being prioritized
- **Salesforce & Looker**: Targeting Q4 for Salesforce, working on timeline for Looker
- **Square Wise Integration**: Request for technical teams integration
- **Multiple Notion Spaces**: Users want ability to connect to multiple Notion workspaces simultaneously

### **2. Technical Issues & Troubleshooting**

#### **Context Limit Errors**
- Users experiencing "conversation has exceeded the model's context limit" errors
- **Solution**: Start a new session to reset token count
- Progress bar shows token usage during chat sessions
- Future context summarization feature planned

#### **Google Calendar SSL Issues**
- SSL/TLS connection failures when connecting to Google Calendar API
- Error: `[SSL] record layer failure (_ssl.c:2580)`
- Issue appears intermittent and may resolve on its own

......
```


crates/goose/src/agents/subagent_task_config.rs
```diff
@@ -15,6 +15,7 @@ pub const GOOSE_SUBAGENT_MAX_TURNS_ENV_VAR: &str = "GOOSE_SUBAGENT_MAX_TURNS";
 pub struct TaskConfig {
     pub id: String,
     pub provider: Option<Arc<dyn Provider>>,
+    pub parent_session_id: Option<String>,
```
DOsinga: why is this optional?

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -94,3 +91,143 @@ pub async fn run_complete_subagent_task_with_options(
...
+fn get_agent_messages(
+    text_instruction: String,
+    task_config: TaskConfig,
+) -> BoxFuture<'static, Result<Conversation>> {
+    Box::pin(async move {
+        let agent_manager = Arc::new(
+            AgentManager::new(None)
+                .await
+                .map_err(|e| anyhow!("Failed to create AgentManager: {}", e))?,
+        );
```
DOsinga: I don' t think we should be creating a new agent manager here just to get one agent

crates/goose/src/agents/reply_parts.rs
```diff
@@ -84,48 +84,6 @@ impl Agent {
         Ok((tools, toolshim_tools, system_prompt))
     }
 
-    /// Generate a response from the LLM provider
-    /// Handles toolshim transformations if needed
-    pub(crate) async fn generate_response_from_provider(
```
DOsinga, line 89: good!

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
```diff
@@ -111,28 +111,44 @@ fn process_extensions(
...
-                    }
-                    Err(e) => {
-                        tracing::warn!("Error looking up extension '{}': {}", name_str, e);
-                    }
+            let name_str = if let Some(name_str) = ext.as_str() {
+                // Case 1: Extension is a simple string (e.g., "slack")
+                name_str
+            } else if let Some(name_value) = ext.get("name") {
+                // Case 2: Extension is an object with a "name" field (e.g., {"name": "slack"})
+                if let Some(name_str) = name_value.as_str() {
```
DOsinga: why do we need to support this format now too?

crates/goose/src/agents/agent.rs
```diff
@@ -454,8 +455,19 @@ impl Agent {
...
+                if let Identifier::Path(path_str) = &session_config.id {
+                    // Convert string to Path and extract the file stem (filename without extension)
+                    let path = Path::new(path_str);
+                    path.file_stem()
+                        .and_then(|stem| stem.to_str())
+                        .map(|stem| stem.to_string())
+                } else {
+                    None
+                }
+            });
```
DOsinga: what is parent_session_id here? the session id of the agent?

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
yingjiehe-xyz: The format is not full restrictedly, it can be something like:
```
        extensions:
            - slack
 ```

or 

```
        extensions:
            - name: slack
 ```

so to make it more feasible, I added such parsing

crates/goose/src/agents/subagent_task_config.rs
yingjiehe-xyz: yeah, try to make the task more general, instead of always having a parent task, it can be a standalone task

crates/goose/src/agents/agent.rs
yingjiehe-xyz: it is the parent agent session id which is introduced in the agent manager: https://github.com/block/goose/blob/b7604e0b13ae5c415a5a69496b25f119011e0e18/crates/goose/src/execution/mod.rs#L15, right now I don't think it is fully used yet, add @tlongwell-block to comment

crates/goose/src/agents/subagent_handler.rs
yingjiehe-xyz: agree, let me make it be shared as singleton

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
DOsinga: hmm, it really shouldn't be though, see the definition of recipe. there it says extensions are of type vector ExtensionConfig which we support here. I suppose we have a short cut here to also support names directly, though I'm not sure how that would happen or even parse, but, eh, that was already the case. why would we need to make what we accept looser here? what is the reason for that.

crates/goose/src/agents/agent.rs
DOsinga: but what is the code trying to do? it looks like it is trying to get the id out of the path of the current session? that doesn't seem right

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
yingjiehe-xyz: this extension config is generated from LLM, even we told LLM to do so, but sometime it has the name, sometime not, to tolerate the llm bad generation, I added this handing here.

crates/goose/src/agents/agent.rs
yingjiehe-xyz: yes, trying to get the session id string instead of the full path

yingjiehe-xyz: rebase to your PR and use the new session id

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
DOsinga: hmm, that feels fishy to me. where do we generate that? we should be able to have it do a tool call and return the recipe exactly according to the json spec, no?

crates/goose-server/src/state.rs
```diff
@@ -18,7 +18,7 @@ pub struct AppState {
 
 impl AppState {
     pub async fn new() -> anyhow::Result<Arc<AppState>> {
-        let agent_manager = Arc::new(AgentManager::new(None).await?);
+        let agent_manager = AgentManager::instance(None).await?;
```
DOsinga: presumably we'll merge the other one and sync?

crates/goose/src/agents/agent.rs
```diff
@@ -467,8 +467,9 @@ impl Agent {
                 .clone()
                 .map(Value::Object)
                 .unwrap_or(Value::Object(serde_json::Map::new()));
-
-            let task_config = TaskConfig::new(provider);
+            let parent_session_id = session.as_ref().map(|s| s.id.to_string());
+            let task_config = TaskConfig::new(provider, parent_session_id);
```
DOsinga: parent_session_id should probably not be optional here, should it? neither provider I would think

crates/goose/src/agents/agent.rs
```diff
@@ -467,8 +467,9 @@ impl Agent {
...
+          
             subagent_execute_task_tool::run_tasks(
                 arguments,
```
DOsinga: we shouldn't pass it raw arguments here; figure out what we need and return an error here rather ignoring which I think we are doing right now deep in the code

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -94,3 +92,89 @@ pub async fn run_complete_subagent_task_with_options(
...
+fn get_agent_messages(
+    text_instruction: String,
+    task_config: TaskConfig,
+) -> BoxFuture<'static, Result<Conversation>> {
+    Box::pin(async move {
+        let agent_manager = AgentManager::instance(None)
+            .await
+            .map_err(|e| anyhow!("Failed to create AgentManager: {}", e))?;
+        let parent_session_id = task_config
+            .parent_session_id
```
DOsinga: at this point this should not be Option<>

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -94,3 +92,89 @@ pub async fn run_complete_subagent_task_with_options(
...
+            .ok_or_else(|| anyhow!("Parent session ID is missing"))?;
+
+        let current_dir = std::env::current_dir()
+            .map_err(|e| anyhow!("Failed to get current directory for sub agent: {}", e))?;
```
DOsinga: do we want to run this in the current directory? that doesn't seem right

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -94,3 +92,89 @@ pub async fn run_complete_subagent_task_with_options(
...
+        let session = SessionManager::create_session(
+            current_dir.clone(),
+            format!("Subagent task for: {}", parent_session_id),
+        )
+        .await
+        .map_err(|e| anyhow!("Failed to create a session for sub agent: {}", e))?;
```
DOsinga, line 108: this should work, but it will create a user visible session - is that what we intend here?

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -94,3 +92,89 @@ pub async fn run_complete_subagent_task_with_options(
...
+                    );
+                }
+            }
+        }
+
+        let mut session_messages =
+            Conversation::new_unvalidated(
+                vec![Message::user().with_text(text_instruction.clone())],
+            );
+        let current_dir = std::env::current_dir()
```
DOsinga: uh, I think you're doing this twice

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -94,3 +92,89 @@ pub async fn run_complete_subagent_task_with_options(
...
+        let mut stream = agent
+            .reply(session_messages.clone(), Some(session_config), None)
+            .await
+            .map_err(|e| anyhow!("Failed to get reply from agent: {}", e))?;
+        while let Some(message_result) = stream.next().await {
+            match message_result {
+                Ok(AgentEvent::Message(msg)) => session_messages.push(msg),
+                Ok(AgentEvent::McpNotification(_))
+                | Ok(AgentEvent::ModelChange { .. })
+                | Ok(AgentEvent::HistoryReplaced(_)) => {} // Handle informational events
```
DOsinga, line 151: are we sure we can ignore these though? ModelChange feels like something we might want to pay attention to. maybe handle inside of agent.reply?

crates/goose/src/agents/subagent_task_config.rs
DOsinga: I'm not seeing that in the code. I think both provider and parent_session_id need to be non optional here

DOsinga (changes requested):
this looks pretty decent, but I think we need to tighten up the typing here and make the compiler our friend.

my main concern here is how we run the agent here. we have a parent_id but we're not doing all that much with it. the TaskConfig has a provider and a parent_id and the provider is the provider of the parent. maybe it should copy that on instantiation? same for working_dir I think. extensions is set separately again, but I feel like this should sometimes default to whatever the agent had at the time? like if you run a subrecipe where you don't specify extensions

crates/goose-server/src/state.rs
yingjiehe-xyz: it is outdated, the new code should not have it now

crates/goose/src/agents/subagent_handler.rs
yingjiehe-xyz: Model change events are informational, I don't have the full context this event, it seems like for AutoPilot only

crates/goose/src/agents/subagent_handler.rs
yingjiehe-xyz: good catch

crates/goose/src/agents/agent.rs
yingjiehe-xyz: updated

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
yingjiehe-xyz: it is generated in the dynamic task tool, sometime llm will make mistakes, I can remove the sanity check if you want

crates/goose/src/agents/subagent_handler.rs
yingjiehe-xyz: we can use the same parent working dir

crates/goose/src/agents/subagent_handler.rs
yingjiehe-xyz: yeah, I think visible  session should be fine here, users can check the sub session if they want

yingjiehe-xyz:
> this looks pretty decent, but I think we need to tighten up the typing here and make the compiler our friend.
> 
> my main concern here is how we run the agent here. we have a parent_id but we're not doing all that much with it. the TaskConfig has a provider and a parent_id and the provider is the provider of the parent. maybe it should copy that on instantiation? same for working_dir I think. extensions is set separately again, but I feel like this should sometimes default to whatever the agent had at the time? like if you run a subrecipe where you don't specify extensions

for the extension, I think it would be better to start with specified extension to reduce the prompt length and keep it consistent with recipe

crates/goose/src/agents/agent.rs
```diff
@@ -461,14 +462,23 @@ impl Agent {
                 .dispatch_sub_recipe_tool_call(&tool_call.name, arguments, &self.tasks_manager)
                 .await
         } else if tool_call.name == SUBAGENT_EXECUTE_TASK_TOOL_NAME {
-            let provider = self.provider().await.ok();
+            let provider = self.provider().await.unwrap();
+            let parent_session_id = session
+                .as_ref()
+                .map(|s| s.id.to_string())
+                .unwrap_or_default();
```
DOsinga: on 465, don't unwrap it will kill the server.  for session related, check that we have a session, otherwise fail the toolcall - defaulting the session id to "" will just make it die elsewhere

crates/goose/src/agents/agent.rs
```diff
@@ -461,14 +462,23 @@ impl Agent {
...
+                .as_ref()
+                .map(|s| s.working_dir.clone())
+                .unwrap_or_else(|| std::env::current_dir().unwrap_or_else(|_| PathBuf::from(".")));
+
+            let task_config = TaskConfig::new(provider, parent_session_id, parent_working_dir);
             let arguments = tool_call
                 .arguments
                 .clone()
                 .map(Value::Object)
                 .unwrap_or(Value::Object(serde_json::Map::new()));
```
DOsinga: get the arguments we need and pass that on - if we have the wrong one, we want to fail the toolcall ehre

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
DOsinga: where is it generated? the generation should use a toolcall so we can be sure it generates the right schema

crates/goose/src/agents/subagent_task_config.rs
```diff
@@ -14,7 +15,9 @@ pub const GOOSE_SUBAGENT_MAX_TURNS_ENV_VAR: &str = "GOOSE_SUBAGENT_MAX_TURNS";
 #[derive(Clone)]
 pub struct TaskConfig {
     pub id: String,
```
DOsinga: for my information, what is id for?

yingjiehe-xyz: sorry, I don't know, @lifeizhou-ap should have more context

crates/goose/src/agents/agent.rs
yingjiehe-xyz: > get the arguments we need and pass that on - if we have the wrong one, we want to fail the toolcall ehre

Sorry, I don't think I have the full context here, what do you mean here?

crates/goose/src/agents/agent.rs
yingjiehe-xyz: I don't think it is doable here, from my understanding, the arguments can various of things for recipe

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
yingjiehe-xyz: it should be here: https://github.com/block/goose/blob/8f5eb10ba45e10f6994a4f9327d73b02de5bedb9/crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs#L21

crates/goose/src/agents/subagent_task_config.rs
lifeizhou-ap: I don't think it is used.  but better to check with @wendytang as she created TaskConfig just in case we miss some context

yingjiehe-xyz: checked the references, looks like no usage, I think we can remove it

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
DOsinga: yeah, that's the problem. that underspecifies what we want to see for extensions. we should just make that into a list of names /cc @tlongwell-block

crates/goose/src/agents/agent.rs
DOsinga: Have a look at the code. we're only using this for task_ids and for execution_mode I think

DOsinga:
>for the extension, I think it would be better to start with specified extension to reduce the prompt length and keep it consistent with recipe

sorry, but that is not how I read the code. I think we ask the AI to generate a spec to run with an empty list for extensions for no extensions and none for inheriting the extensions. /cc @tlongwell-block

tlongwell-block:
> > for the extension, I think it would be better to start with specified extension to reduce the prompt length and keep it consistent with recipe
> 
> sorry, but that is not how I read the code. I think we ask the AI to generate a spec to run with an empty list for extensions for no extensions and none for inheriting the extensions. /cc @tlongwell-block

Right now, if the parent agent does not specify the `extensions` parameter, the subagent it creates gets all of the same extensions as the parent agent

If the parent agent specifies an empty `extensions` parameter, the subagent it creates gets no extensions

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
yingjiehe-xyz: Then it should be done for a separate PR, if we don't want this, let me revert this change here

yingjiehe-xyz: revert this change and the prompt update should be done separated

tlongwell-block:
```
─── create_task | dynamic_task ──────────────────────────
task_parameters:
    -
        title: Test 4a: Weather
        instructions: Tell me a fun fact about weather
        return_last_only: true
    -
        title: Test 4b: Joke
        instructions: Tell me a knock-knock joke
        return_last_only: true


─── execute_task | subagent ──────────────────────────
task_ids: 87641b31-7d63-4081-95cf-6c5e957f5f3e, e8b0b3ed-a30a-4174-b521-636e95ab49fc
execution_mode: parallel


🎯 Task Execution Dashboard
═══════════════════════════

📊 Progress: 2 total | ⏳ 0 pending | 🏃 0 running | ✅ 1 completed | ❌ 1 failed

❌ 87641b31-7d63-4081-95cf-6c5e957f5f3e (inline_recipe)
   ⏱️  0.0s
   ⚠️  Inline recipe execution failed: -32603: Failed to execute task: Failed to cre...

✅ e8b0b3ed-a30a-4174-b521-636e95ab49fc (inline_recipe)
   ⏱️  4.0s
   📄 {
  "result": "Here's a knock-knock joke for you:\n\nKnock, knock.  \nWho's there?  \nInterrupting cow.  \nInterrupting cow w—  \n**MOOOOO!**\n\n😄 The joke works because the cow interrupts before you can finish saying \"who?\""
}

Execution Complete!
═══════════════════════
Total Tasks: 2
✅ Completed: 1
❌ Failed: 1
📈 Success Rate: 50.0%

❌ Failed Tasks:
   • 87641b31-7d63-4081-95cf-6c5e957f5f3e
     Error: Inline recipe execution failed: -32603: Failed to execute task: Failed to create a session for sub agent: error returned from database: (code: 5) database is locked

📝 Generating summary...
-32603: 1/2 tasks failed:
Task '87641b31-7d63-4081-95cf-6c5e957f5f3e' (ID: 87641b31-7d63-4081-95cf-6c5e957f5f3e): Inline recipe execution failed: -32603: Failed to execute task: Failed to create a session for sub agent: error returned from database: (code: 5) database is locked

Output: No output captured

Interesting! We hit a database locking issue with parallel execution. Let me update the TODO and retry:
```

DOsinga:
also let me know @yingjiehe-xyz when you want me to have another look

yingjiehe-xyz:
> also let me know @yingjiehe-xyz when you want me to have another look

yes, I think it should be ready for the review

crates/goose/src/agents/agent.rs
DOsinga: hmm, I don't know that we should resolve this. 

what I was trying to say is that this whole block *needs* a sesion. you can't unpack things here from the session and then default to "" for a session-id. that's invalid and will crash somewhere else when we need it (we might not read it but still) - same for the working-dir. you can't default to the current_dir() since goosed's current dir is not the same.

crates/goose/src/agents/agent.rs
DOsinga: arguments now becomes an untyped Value. rather than sending that down the rabbit hole, unpack it here into the components that you need. concretely we're using execution_mode and task_ids

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -94,3 +90,86 @@ pub async fn run_complete_subagent_task_with_options(
...
+fn get_agent_messages(
+    text_instruction: String,
+    task_config: TaskConfig,
+) -> BoxFuture<'static, Result<Conversation>> {
+    Box::pin(async move {
+        let agent_manager = AgentManager::instance()
+            .await
+            .map_err(|e| anyhow!("Failed to create AgentManager: {}", e))?;
+        let parent_session_id = task_config.parent_session_id;
+        let current_dir = task_config.parent_working_dir;
```
DOsinga: call this working_dir to distinguish

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -94,3 +90,86 @@ pub async fn run_complete_subagent_task_with_options(
...
+            .update_provider(task_config.provider)
+            .await
+            .map_err(|e| anyhow!("Failed to set provider on sub agent: {}", e))?;
+
+        let extensions_to_add = if let Some(ref extensions) = task_config.extensions {
+            // Use the explicitly specified extensions
+            extensions.clone()
+        } else {
+            // Default behavior: use all enabled extensions
+            ExtensionConfigManager::get_all()
```
DOsinga: we shouldn't default to the globally enabled extensions, but to the extensions of the parent agent

crates/goose/src/agents/agent.rs
yingjiehe-xyz: Done

crates/goose/src/agents/subagent_handler.rs
yingjiehe-xyz: Done

crates/goose/src/agents/subagent_handler.rs
yingjiehe-xyz: Done

crates/goose/src/agents/agent.rs
yingjiehe-xyz: Done

crates/goose/src/agents/agent.rs
```diff
@@ -451,16 +460,52 @@ impl Agent {
...
 
-            let task_config = TaskConfig::new(provider);
+            let task_ids: Vec<String> = arguments
+                .get("task_ids")
+                .and_then(|v| serde_json::from_value(v.clone()).ok())
+                .unwrap_or_default();
+            let execution_mode = arguments
+                .get("execution_mode")
+                .and_then(|v| serde_json::from_value::<ExecutionMode>(v.clone()).ok())
+                .unwrap_or_default();
```
DOsinga: we should return an error here if they are not set, right?

crates/goose/src/agents/agent.rs
yingjiehe-xyz: done, we can just return the error here

crates/goose/src/agents/agent.rs
yingjiehe-xyz: execution_mode can be default to sequential, others we can return errors

DOsinga: I know it can be set to it, but if it is not, whomever called this (the LLM) screwed up so I wouldn't trust it and throw an error

PR #4833, opened 2025-09-25
Add MOIM (Minus One Info Message) for automatic context injection (TODO list and current time)

tlongwell-block:
Implements ephemeral context injection at the provider boundary to automatically include timestamp and TODO content in LLM context without modifying conversation history.

**Problem**

Previously, accessing TODO content required explicit `todo__read` tool calls, cluttering conversation history and consuming tokens. Timestamps were injected via template variables in system prompts, creating inconsistency across different prompt variants and preventing effective prompt caching.

**Solution**

MOIM (Minus One Info Message) injects a synthetic user message at approximately position -1 during provider calls. This message contains:
- Current timestamp (always)
- TODO content from session metadata (when available)

The injection happens in `reply_parts.rs` at the provider boundary, ensuring the message is never persisted to storage or returned in conversation history.

**Architecture Benefits**

- Zero changes to core data structures (`Message`, `Conversation`)
- No persistence layer modifications  
- Clean module boundaries
- No impact on existing sessions or conversation replay

**Testing**

Comprehensive test coverage in `crates/goose/tests/moim_tests.rs` with proper test isolation using `serial_test` for tests that modify environment variables.

**Migration**

- Removed `todo__read` tool (breaking change for recipes using it)
- Updated `todo__write` description to reflect automatic availability
- System prompts now receive timestamp via MOIM instead of template variables

### DIFF
--- a/crates/goose/src/agents/moim.rs
+++ b/crates/goose/src/agents/moim.rs
@@ -1,0 +1,160 @@
+use crate::agents::types::SessionConfig;
+use crate::conversation::message::Message;
+use crate::conversation::Conversation;
+use crate::session::extension_data::ExtensionState;
+use crate::session::{self, TodoState};
+use chrono::Local;
+
+async fn build_moim_content(session: &Option<SessionConfig>) -> Option<String> {
+    let timestamp = Local::now().format("%Y-%m-%d %H:%M:%S").to_string();
+    let mut content = format!("Current date and time: {}\n", timestamp);
+
+    if let Some(todo_content) = get_todo_context(session).await {
+        content.push_str("\nCurrent tasks and notes:\n");
+        content.push_str(&todo_content);
+        content.push('\n');
+    }
+
+    Some(content)
+}
+
+pub async fn inject_moim_if_enabled(
+    messages_for_provider: Conversation,
+    session: &Option<SessionConfig>,
+) -> Conversation {
+    // Check if MOIM is enabled
+    let moim_enabled = crate::config::Config::global()
+        .get_param::<bool>("goose_moim_enabled")
+        .unwrap_or(true);
+
+    if !moim_enabled {
+        return messages_for_provider;
+    }
+
+    if let Some(moim_content) = build_moim_content(session).await {
+        let mut msgs = messages_for_provider.messages().to_vec();
+
+        if !msgs.is_empty() {
+            let moim_msg = Message::user().with_text(moim_content);
+
+            // Insert at position -1 (before last message)
+            // This ensures MOIM appears just before the latest user query
+            let insert_pos = msgs.len().saturating_sub(1);
+            msgs.insert(insert_pos, moim_msg);
+        }
+
+        Conversation::new_unvalidated(msgs)
+    } else {
+        messages_for_provider
+    }
+}
+
+async fn get_todo_context(session: &Option<SessionConfig>) -> Option<String> {
+    let session_config = session.as_ref()?;
+
+    match session::storage::get_path(session_config.id.clone()) {
+        Ok(path) => match session::storage::read_metadata(&path) {
+            Ok(metadata) => TodoState::from_extension_data(&metadata.extension_data)
+                .map(|state| state.content)
+                .filter(|content| !content.trim().is_empty()),
+            Err(e) => {
+                tracing::debug!("Could not read session metadata for MOIM: {}", e);
+                None
+            }
+        },
+        Err(e) => {
+            tracing::debug!("Could not get session path for MOIM: {}", e);
+            None
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::conversation::message::Message;
+    use crate::conversation::Conversation;
+
+    #[tokio::test]
+    async fn test_inject_moim_preserves_messages() {
+        // Create a conversation with multiple messages
+        let messages = vec![
+            Message::user().with_text("First message"),
+            Message::assistant().with_text("First response"),
+            Message::user().with_text("Second message"),
+        ];
+        let conversation = Conversation::new_unvalidated(messages.clone());
+
+        // Inject MOIM (with no session, so only timestamp)
+        let result = inject_moim_if_enabled(conversation, &None).await;
+
+        // Should have one more message (the MOIM)
+        assert_eq!(result.len(), messages.len() + 1);
+
+        // Original messages should still be present
+        assert_eq!(result.messages()[0].as_concat_text(), "First message");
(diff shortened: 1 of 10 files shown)


crates/goose/src/agents/moim.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    Some(content)
+}
+
+/// Find a safe insertion point for MOIM.
+///
+/// We want to insert as close to the end as possible, but we must avoid
+/// breaking tool call/response pairs. We check if inserting at a position
+/// would separate a tool call from its response.
+#[doc(hidden)]
```
DOsinga: why?

crates/goose-cli/src/scenario_tests/scenarios.rs
```diff
@@ -8,10 +8,13 @@ mod tests {
     use crate::scenario_tests::scenario_runner::run_scenario;
     use anyhow::Result;
     use goose::conversation::message::Message;
+    use serial_test::serial;
 
     #[tokio::test]
+    #[serial]
     async fn test_what_is_your_name() -> Result<()> {
-        run_scenario(
+        std::env::set_var("GOOSE_MOIM_ENABLED", "false");
```
DOsinga, line 16: don't set environment variables in tests like this; if they fail you now changed the settings for the next test. also if you are going to add this to all run_scenario, maybe handle this there?

crates/goose/src/agents/reply_parts.rs
```diff
@@ -144,6 +145,10 @@ impl Agent {
             Conversation::new_unvalidated(messages.to_vec())
         };
 
+        // Inject MOIM (timestamp + TODO content) if enabled
+        let messages_for_provider =
+            super::moim::inject_moim_if_enabled(messages_for_provider, session).await;
```
DOsinga, line 150: you should do this in agent.reply. there you can stick it into the block that already checks for session and make session non optional

crates/goose/src/prompts/system.md
```diff
@@ -1,8 +1,7 @@
 You are a general-purpose AI agent called goose, created by Block, the parent company of Square, CashApp, and Tidal. goose is being developed as an open-source software project.
 
-The current date is {{current_date_time}}.
+Goose uses LLM providers with tool calling capability. You can be used with different language models (gpt-4o, claude-sonnet-4, o1, llama-3.2, deepseek-r1, etc).
```
DOsinga: why this?

crates/goose-cli/src/scenario_tests/scenarios.rs
tlongwell-block: I have the tests in serial now. I set the env var at the top of the test and unset at the bottom. It won't effect other tests that way

I think you're right that I could get the same effect by un/setting it in run_scenario(), but still making sure that everything that calls run_scenario() is `serial`

The reason I didn't is just because I can't internally guarantee that run_scenario will be in serial, hence it would be possible for an environment race condition to occur if called by a non-serial test

crates/goose/src/agents/reply_parts.rs
tlongwell-block: I put it in `stream_response_from_provider` because I wanted it to be at the very, very edge of the shared inference execution path. This is the last point on that execution path that doesn't split between different providers

This makes reasoning about it much easier. There's no chance to be written to a session file, be compacted, etc, etc

tlongwell-block:
> left some preliminary comments. I think we should do something like this, but we need a structure that allows any platform tool this sort of thing. see: #4868 and the relevant discussion

Agree, yes. I think adding to build_moim_content() is the way to go here since there will probably be formatting and helpers involved when moving platform tool content from its storage into the MOIM (ie an array of sources is probably not sufficient, so using the build function is probably the best way forward)

crates/goose/src/agents/reply_parts.rs
tlongwell-block: As far as session optionality goes: it should 100% be required, but we can't do it yet. Once recipes/subagents/scheduler are on Agent Manager, they will have real sessions.

But, until then, stream_response_from_provider will be called without a session when dealing with recipes/subagents

crates/goose/src/agents/moim.rs
```diff
@@ -0,0 +1,94 @@
+use crate::agents::types::SessionConfig;
+use crate::conversation::message::Message;
+use crate::conversation::Conversation;
+use crate::session::extension_data::{ExtensionState, TodoState};
+use crate::session::SessionManager;
+use chrono::Local;
+
+async fn build_moim_content(session: &Option<SessionConfig>) -> Option<String> {
```
katzdave, line 8: I'd really like to see this as something the tool itself can define rather than a specific global place.

crates/goose/src/agents/reply_parts.rs
katzdave: Seems like the dependency is there from how the TODO stool stores its state over actually needing the session.

tlongwell-block: Yes. Instead of passing the session to inject_moim_if_enabled, I could pass TodoState::from_extension_data directly? It just ended up looking cleaner this way

crates/goose/src/agents/moim.rs
tlongwell-block: I talked about my thinking in a reply above:

"... I think adding to build_moim_content() is the way to go here since there will probably be formatting and helpers involved when moving platform tool content from its storage into the MOIM (ie an array of sources is probably not sufficient, so using the build function is probably the best way forward)"

Having each tool register itself to a moim system would work, too, but we might still want to format how they fit together, which requires a global function like this one

also, this global function allows us to put data in the moim that isn't related to a tool. Like the timestamp here. This could also be environment information or other non-tool context that we find useful for the agent

crates/goose/src/agents/reply_parts.rs
tlongwell-block: Ah, no, lets keep passing the session in because we'll probably want to add more to moim from the session metadata later at some point. And its neater

crates/goose/src/agents/moim.rs
DOsinga: in light of the latest infra changes here's my proposal:

* add a method to the platform extension type get_moim() and implement that for todo
* add a method to extension_manager get_moim that goes through its platform extensions and collects them and then adds the current time
* call this method from reply_internal (yeah, I know that is scary)

tlongwell-block: Working in https://github.com/block/goose/pull/5027

tlongwell-block:
https://github.com/block/goose/pull/5027

PR #4834, opened 2025-09-25
feat(bedrock): Fetch model list from Bedrock

jm:
## Summary

This PR implements `fetch_supported_models` for the AWS Bedrock provider.  Currently, the Bedrock provider is unusable out of the box unless you supply a custom model ID since all of the entries in the current `BEDROCK_KNOWN_MODELS` array aren't usable as model IDs for invocations.  This solves that issue as well as implementing the helper to grab the list of usable model IDs/ARNs from Bedrock directly.

## Changes

- Added the `aws-sdk-bedrock` crate, not to be confused with the currently included `aws-sdk-bedrockruntime` crate.  This crate is required for getting information about available models/inference profiles whereas the other crate is only for invoking models and agents.
- Implemented `fetch_supported_models` by fetching the user's configured application inference profiles and then the models that are available for on-demand inference.  The inference profiles are listed first (sorted) followed by the on-demand list (sorted).  From a UX perspective, that seems like the most useful sorting to me since I'm probably wanting to use inference profiles I've configured, but obviously I'm not stuck to that for any real reason.
- Added a working set of model IDs to the `BEDROCK_KNOWN_MODELS` list with a little more variety in providers.  If the user's account hasn't enabled any of the models in Bedrock, this list isn't that useful[^1], but at least if they only have invoke privileges (i.e., rather than list/get/admin privileges) on their IAM role, they can use the most popular ones without needing to input something custom.  With the current list in the code, even if the model selected from the list is enabled it won't work.

(Side note: A similar PR is incoming for Azure OpenAI...)

[^1]: That issue will mostly go away in October 2025 when they will start enabling all models by default and access will be controlled solely by the user's IAM policy.


jm:
Ah whoops just seeing the failures on this.  Builds clean for me, but I'll get this fixed up early next week. 👍

DOsinga:
thank you. sorry for the delay in replying. I started looking at this and then saw that bedrock can hang the entire app because of some dumb not async thing. I made main async for provider init. might help here too. shout if you need a review

DOsinga:
do we still want to get this in?

alexhancock:
Thanks for the work on this. Going to close this out per https://github.com/block/goose/pull/4834#issuecomment-3418837151 being three weeks ago, but welcome to re-open a new PR.

PR #4835, opened 2025-09-25
Add a global size limit for tool call responses

katzdave:
Truncate the response to the limit and throw an error message.

Theory is large tool call response are what's causing the databricks content too long message, and this should hopefully address.


DOsinga:
do you think just a large tool reponse will break databricks without it overall hitting the limit of the context window? if so, I think we should replicate that in a test first, before speculative trying to fix things

PR #4836, opened 2025-09-25
Update databricks flash model

katzdave:
Looks like flash-1-5 went mia


DOsinga (approved):

jamadeo:
something similar seems to have broken this integration test: https://github.com/block/goose/pull/4837

PR #4837, opened 2025-09-26
fix: pricing integration test

jamadeo:
The real fix here is that we should not have a test that breaks when an external API changes. If we want to keep this text around we should mock the API


DOsinga:
I think I'd delete the entire thing

PR #4841, opened 2025-09-26
fix: linux builds: use gcc 10 in Linux builds

jamadeo:
What broke: cryptographic dependency aws-lc-rs checks during compilation that the C compiler is one that is not affected by https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189. Our linux builds use Ubuntu 20, which ships with gcc 9, which is affected. This PR installs and uses gcc 10, which includes a fix and passes the aws-lc-rs build.

### DIFF
--- a/.github/workflows/bundle-desktop-linux.yml
+++ b/.github/workflows/bundle-desktop-linux.yml
@@ -16,8 +16,6 @@ on:
         required: false
         default: ''
 
-  pull_request: {}
-
 name: "Bundle Desktop (Linux)"
 
 jobs:


.github/workflows/build-cli.yml
```diff
@@ -35,11 +35,13 @@ jobs:
             target-suffix: unknown-linux-gnu
             build-on: ubuntu-latest
             use-cross: true
+            cc: gcc-10
```
alexhancock, line 38: looks like latest stable version of gcc is 15. could we update all the way to newest?

alexhancock (approved):

.github/workflows/build-cli.yml
jamadeo: quite possibly, yeah. I went with 10 because ubuntu 20 (what we're building on) ships with 9, and 10 is where the original issue is fixed: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189

according to some googling though, 10 is the latest in the official repositories for ubuntu 20 (I haven't verified) so it might mean some extra steps to get it working for not much gain

PR #4847, opened 2025-09-26
Add filtering for agentVisible: false messages on streaming providers

katzdave:
This is the root cause of the input too long issues we're seeing.

I also saw some quirks with auto-compact triggering right after compact; this is from the provider metadata but I updated the token estimation approach to also filter.

perform_compaction doesn't need filtering since it falls thru to complete_fast().


crates/goose/src/conversation/mod.rs
```diff
@@ -102,6 +102,42 @@ impl Conversation {
...
+    /// # let conversation = Conversation::new_unvalidated(vec![
+    /// #     Message::user().with_text("Hello")
+    /// # ]);
+    /// // Get agent-visible messages regardless of user visibility (don't care about user_visible)
+    /// let agent_messages = conversation.filtered_messages(|meta| meta.agent_visible);
+    ///
+    /// // Get messages visible ONLY to agent (not visible to user)
+    /// let agent_only = conversation.filtered_messages(|meta| meta.agent_visible && !meta.user_visible);
+    /// ```
+    pub fn filtered_messages<F>(&self, filter: F) -> Vec<Message>
```
DOsinga, line 105: ah, this is not really what I meant - meant just pass it a visibility level for the messages and then return those

katzdave: visibility level kind of needs some kind of mask though on MessageMetadata? Don't quite follow.

DOsinga: i meant something like

pub fn visible_messages(for: User|Agent) but maybe that's not the API you imagined.

katzdave: Can do that but I was imagining adding more tags to message metadata like summary marker, notification so wanted to keep them synced.

katzdave:
pushed some changes refactoring the provider side impl. I also noticed it did some filtering of thinking messages in anthropic provider.

Unsure about those, seems to have been modified to match what google is doing. Could these be related to goose repeating itself?

crates/goose/src/providers/formats/anthropic.rs
```diff
@@ -32,8 +32,8 @@ const DATA_FIELD: &str = "data";
 pub fn format_messages(messages: &[Message]) -> Vec<Value> {
     let mut anthropic_messages = Vec::new();
 
-    // Convert messages to Anthropic format
-    for message in messages {
+    // Filter messages to only include agent_visible ones and convert to Anthropic format
+    for message in messages.iter().filter(|m| m.is_agent_visible()) {
```
DOsinga, line 35: to be honest, if we go through  the loop here anyway, we might as well skip them in place rather than prefilter

crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -71,10 +71,18 @@ pub async fn check_compaction_needed(
     let (current_tokens, token_source) = match session_metadata.and_then(|m| m.total_tokens) {
         Some(tokens) => (tokens as usize, "session metadata"),
         None => {
+            // Filter to only agent-visible messages for token counting
+            let agent_visible_messages: Vec<Message> = messages
+                .iter()
+                .filter(|m| m.is_agent_visible())
+                .cloned()
+                .collect();
```
DOsinga: shouldn't we move this inside get_messages_token_counts_async?

crates/goose/src/conversation/mod.rs
DOsinga: ok, some other time, let's just drop this for now. or keep it but remove the comment

crates/goose/src/providers/formats/anthropic.rs
```diff
@@ -96,18 +96,14 @@ pub fn format_messages(messages: &[Message]) -> Vec<Value> {
...
+                    continue;
                 }
-                MessageContent::RedactedThinking(redacted) => {
-                    content.push(json!({
-                        TYPE_FIELD: REDACTED_THINKING_TYPE,
-                        DATA_FIELD: redacted.data
-                    }));
+                MessageContent::RedactedThinking(_) => {
+                    // Skip redacted thinking - internal use only
+                    continue;
```
DOsinga: I'm not sure about this. it sounds reasonable, but do we want to make a change we don't really really understand for this if we want to cherry-pick?

crates/goose/tests/streaming_agent_visible_test.rs
```diff
@@ -0,0 +1,214 @@
...
+use goose::providers::errors::ProviderError;
+use rmcp::model::Tool;
+
+/// Mock provider to test that stream() filters messages with agentVisible=false
+struct MockStreamingProvider {
+    model: ModelConfig,
+    captured_messages: std::sync::Arc<std::sync::Mutex<Vec<Message>>>,
+}
+
+impl MockStreamingProvider {
```
DOsinga: this feels very LLM generated. did you read the code? delete

crates/goose/src/providers/base.rs
```diff
@@ -379,7 +362,7 @@ pub trait Provider: Send + Sync {
                         e,
                         model_config.model_name
                     );
-                    self.complete_with_model(&model_config, system, &agent_visible_messages, tools)
+                    self.complete_with_model(&model_config, system, messages, tools)
                         .await
                 } else {
                     Err(e)
```
DOsinga, line 368: nice. there's also snowflake & anthropic that come with their own message formats

crates/goose/src/conversation/mod.rs
```diff
@@ -531,4 +567,142 @@ mod tests {
         let (_fixed, issues) = run_verify(messages);
         assert_eq!(issues.len(), 0);
     }
+
+    #[test]
+    fn test_filtered_messages() {
```
DOsinga: these tests don't add anything

crates/goose/src/conversation/mod.rs
katzdave: ack. leaving for now but will follow up and use it + might update it a bit.

crates/goose/src/providers/formats/anthropic.rs
katzdave: Goose claims:

The current implementation with `.iter().filter()` is already optimal - it combines filtering and processing in a single, lazy iteration without any unnecessary allocations. There's no performance benefit to pre-filtering in this case.

crates/goose/src/context_mgmt/auto_compact.rs
katzdave: done, nice call.

crates/goose/src/providers/formats/anthropic.rs
katzdave: Agreed, reverted but wanted to just call this out as somewhere to investigate.

crates/goose/tests/streaming_agent_visible_test.rs
katzdave: gone

crates/goose/src/conversation/mod.rs
katzdave: done

crates/goose/src/providers/base.rs
katzdave: done

PR #4857, opened 2025-09-26
rough sandbox test

tlongwell-block:


DOsinga (commented):
sounds interesting. what makes it a sandbox? could we integrate this with the normal file reader?

tlongwell-block:
> sounds interesting. what makes it a sandbox? could we integrate this with the normal file reader?

This is a demo of how a subagent with read only tools can be used as a dual-llm sandbox type prompt injection detection system. I'll close this, don't actually want to merge the demo file viewer

Issue #4658, opened 2025-09-16
Path is duplicated on tool calls causing them to fail


the-destro:
It looks like the paths for files are getting double expanded. So if my home directory is /Users/my_user:

Here's the modified code I'll apply using the `str_replace` command:
─── text_editor | developer ──────────────────────────
path: ~/code/lore/lib/lore/service.ex
command: str_replace
diff: [REDACTED: 779 chars]

-32603: Mcp error: -32603: File '/Users/my_user/code/lore/lib/lore/Users/my_user/code/lore/lib/lore/service.ex' not found and patch doesn't create it  2025-09-16T19:15:52.756030Z  WARN goose::providers::lead_worker: Technical failure with worker provider, retrying with default model (lead provider)
    at crates/goose/src/providers/lead_worker.rs:386

  2025-09-16T19:15:52.756078Z ERROR goose::providers::lead_worker: ❌ Default model (lead provider) also failed - returning original error
    at crates/goose/src/providers/lead_worker.rs:399

  2025-09-16T19:15:52.756091Z  WARN goose::providers::lead_worker: Technical failure detected - API/LLM issue, will use default model
    at crates/goose/src/providers/lead_worker.rs:187

Error: Execution error: content was missing: content was not specified but it is required when building ToolResultBlock
The tool calling loop was interrupted. How would you like to proceed?
  error: The error above was an exception we were not able to handle.
These errors are often related to connection or authentication
We've removed the conversation up to the most recent user message
- depending on the error you may be able to continue

**To Reproduce**

**Expected behavior**

**Screenshots**

**Please provide following information:**
 - **OS & Arch:** macOS 15.6.1
 - **Interface:** CLI
 - **Version:** 1.8.0
 - **Extensions enabled:** [e.g. Computer Controller, Figma]
 - **Provider & Model:** AWS Bedrock Claude

**Additional context**

angelahning:
fixed in [fix: path is duplicated on tool calls causing them to fail (#4658) #4859](https://github.com/block/goose/pull/4859)

PR #4859, opened 2025-09-26
fix: path is duplicated on tool calls causing them to fail (#4658)

demetrio108:
There is an issue opened 

https://github.com/block/goose/issues/4658
Path is duplicated on tool calls causing them to fail

What happens is when LLM calls text_editor in diff mode it passes
path: /a/b/c/base/src/file.go
in diff there is path: base/src/file.go

Inside mpatch they are merged and produce
path: /a/b/c/base/src/base/src/file.go

This path doesn't exist and tool fails.

Maybe it's an error from LLM side, that it calls tool incorrectly. But I tried several and it's a common issue. 

So I propose to remove overlapping part from base_dir before sending to mpatch. I tested fix in Linux and it works good.

### DIFF
--- a/crates/goose-mcp/src/developer/text_editor.rs
+++ b/crates/goose-mcp/src/developer/text_editor.rs
@@ -188,6 +188,29 @@ fn generate_summary(results: &DiffResults, is_single_file: bool, base_path: &Pat
     ]
 }
 
+/// Adjusts the base directory to handle overlapping paths between base_dir and file_path
+fn adjust_base_dir_for_overlap(base_dir: &Path, file_path: &Path) -> PathBuf {
+    // Find the overlapping path: longest suffix of base_dir that is prefix of file_path
+    let base_components: Vec<_> = base_dir.components().collect();
+    let file_components: Vec<_> = file_path.components().collect();
+
+    // Find the maximum k where base_components[(len-k)..] == file_components[0..k]
+    let mut max_k = 0;
+    for k in 1..=(base_components.len().min(file_components.len())) {
+        if &file_components[0..k] == &base_components[(base_components.len() - k)..] {
+            max_k = k;
+        }
+    }
+
+    if max_k > 0 {
+        // Remove the last max_k components from base_dir
+        let adjusted_components = base_components[..base_components.len() - max_k].to_vec();
+        PathBuf::from_iter(adjusted_components)
+    } else {
+        base_dir.to_path_buf()
+    }
+}
+
 /// Applies a single patch and updates results
 fn apply_single_patch(
     patch: &mpatch::Patch,
@@ -196,10 +219,12 @@ fn apply_single_patch(
     results: &mut DiffResults,
     failed_hunks: &mut Vec<String>,
 ) -> Result<(), ErrorData> {
-    let file_path = base_dir.join(&patch.file_path);
+    let adjusted_base_dir = adjust_base_dir_for_overlap(base_dir, &patch.file_path);
+
+    let file_path = adjusted_base_dir.join(&patch.file_path);
 
     // Validate path safety
-    validate_path_safety(base_dir, &file_path)?;
+    validate_path_safety(&adjusted_base_dir, &file_path)?;
 
     // Save history before modifying
     let file_existed = file_path.exists();
@@ -208,7 +233,7 @@ fn apply_single_patch(
     }
 
     // Apply patch with fuzzy matching (70% similarity threshold)
-    let success = apply_patch(patch, base_dir, false, 0.7).map_err(|e| match e {
+    let success = apply_patch(patch, &adjusted_base_dir, false, 0.7).map_err(|e| match e {
         PatchError::Io { path, source } => ErrorData::new(
             ErrorCode::INTERNAL_ERROR,
             format!("Failed to process '{}': {}", path.display(), source),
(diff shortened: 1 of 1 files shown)


michaelneale:
thanks @demetrio108 cc @tlongwell-block wonder if this was impacting well formed diff problem you mentioned?

michaelneale:
@demetrio108 will need to run ./scripts/clippy-lint.sh (or ask goose to sort it out!)

crates/goose-mcp/src/developer/text_editor.rs
```diff
@@ -188,6 +188,29 @@ fn generate_summary(results: &DiffResults, is_single_file: bool, base_path: &Pat
     ]
 }
 
+/// Adjusts the base directory to handle overlapping paths between base_dir and file_path
+fn adjust_base_dir_for_overlap(base_dir: &Path, file_path: &Path) -> PathBuf {
+    // Find the overlapping path: longest suffix of base_dir that is prefix of file_path
```
DOsinga: remove the comment

crates/goose-mcp/src/developer/text_editor.rs
```diff
@@ -188,6 +188,29 @@ fn generate_summary(results: &DiffResults, is_single_file: bool, base_path: &Pat
...
+            max_k = k;
+        }
+    }
+
+    if max_k > 0 {
+        // Remove the last max_k components from base_dir
+        let adjusted_components = base_components[..base_components.len() - max_k].to_vec();
+        PathBuf::from_iter(adjusted_components)
+    } else {
+        base_dir.to_path_buf()
```
DOsinga, line 204: should this change the type too? if so adjust the name

crates/goose-mcp/src/developer/text_editor.rs
```diff
@@ -188,6 +188,29 @@ fn generate_summary(results: &DiffResults, is_single_file: bool, base_path: &Pat
...
+    let base_components: Vec<_> = base_dir.components().collect();
+    let file_components: Vec<_> = file_path.components().collect();
+
+    // Find the maximum k where base_components[(len-k)..] == file_components[0..k]
+    let mut max_k = 0;
+    for k in 1..=(base_components.len().min(file_components.len())) {
+        if &file_components[0..k] == &base_components[(base_components.len() - k)..] {
+            max_k = k;
+        }
+    }
```
DOsinga: remove the comment, it is just the same as the code. also, this can be done more efficiently, you are comparing ever increasing slices instead of comparing whether the next component is the same and breaking when that is no longe the case. 

I think you can do something like:

```
let max_k = base_components.iter().rev()
        .zip(file_components.iter())
        .take_while(|(b, f)| b == f)
        .count();
``

File: crates/goose-mcp/src/developer/text_editor.rs [hunk 4]
```diff
@@ -208,7 +233,7 @@ fn apply_single_patch(
     }
 
     // Apply patch with fuzzy matching (70% similarity threshold)
-    let success = apply_patch(patch, base_dir, false, 0.7).map_err(|e| match e {
+    let success = apply_patch(patch, &adjusted_base_dir, false, 0.7).map_err(|e| match e {
```
DOsinga, line 230: I am not sure I have enough context here, but is this right? should we use the adjusted base here or should we have adjusted the patch file_path?

crates/goose-mcp/src/developer/text_editor.rs
demetrio108: You mean Path to PathBuf? It's impossible to return Path. I'm not sure which name change you mean

crates/goose-mcp/src/developer/text_editor.rs
demetrio108: I made it this way:
```
    let min_len = base_components.len().min(file_components.len());
    let max_k = (1..=min_len)
        .rfind(|&k| file_components[0..k] == base_components[base_components.len() - k..])
        .unwrap_or(0);
```

crates/goose-mcp/src/developer/text_editor.rs
demetrio108: Inside the patch there is valid path relative to current goose working directory. Base directory is passed wrong by LLM

demetrio108:
> @demetrio108 will need to run ./scripts/clippy-lint.sh (or ask goose to sort it

@michaelneale done

crates/goose-mcp/src/developer/text_editor.rs
DOsinga: ah, thanks, that's helpful

crates/goose-mcp/src/developer/text_editor.rs
DOsinga: that is still comparing slices inside a loop, no?

demetrio108: Your option would work if there were mirroring, like [1, 2, 3] [3, 2, 4]. But situation is different: [1, 2, 3] [2, 3, 4]. So I need to compare. This second variant looks more idiomatic though.

DOsinga (approved):
fix the n2 and let's merge!

demetrio108:
@DOsinga I don't understand, what n2?

michaelneale:
@demetrio108 I think @DOsinga means the O n squared thing here https://github.com/block/goose/pull/4859#discussion_r2388020440 - which looks like you changed some code for?

demetrio108:
image

There's KMP algorithm with O(m+n) complexity, but it's an overkill in current case and it's not elegant

PR #4863, opened 2025-09-26
fix redirect to extensions page after deeplink install and show toast with success message

zanesq:
Noticed extension deeplinks weren't redirecting to the extensions page. Also noticed that the extension is not automatically activated simply by going back to chat route and only activated in a completely new session so we need to address that.

For now I added a toast with more info to start a completely new session to activate the extension. I also considered not enabling the extension and letting the user activate it manually but that adds another step.

NOTE: there is an issue currently where refreshing the app after installing from a deeplink starts the deeplink again (same with recipe deeplinks) we should follow up and fix that in another PR

### DIFF
--- a/ui/desktop/src/components/ExtensionInstallModal.tsx
+++ b/ui/desktop/src/components/ExtensionInstallModal.tsx
@@ -47,6 +47,14 @@ interface ExtensionInstallModalProps {
 
 function extractCommand(link: string): string {
   const url = new URL(link);
+
+  // For remote extensions (SSE or Streaming HTTP), return the URL
+  const remoteUrl = url.searchParams.get('url');
+  if (remoteUrl) {
+    return remoteUrl;
+  }
+
+  // For stdio extensions, return the command
   const cmd = url.searchParams.get('cmd') || 'Unknown Command';
   const args = url.searchParams.getAll('arg').map(decodeURIComponent);
   return `${cmd} ${args.join(' ')}`.trim();
--- a/ui/desktop/src/components/settings/extensions/deeplink.ts
+++ b/ui/desktop/src/components/settings/extensions/deeplink.ts
@@ -88,7 +88,8 @@ function getStreamableHttpConfig(
   name: string,
   description: string,
   timeout: number,
-  headers?: { [key: string]: string }
+  headers?: { [key: string]: string },
+  envs?: { [key: string]: string }
 ) {
   const config: ExtensionConfig = {
     name,
@@ -97,6 +98,7 @@ function getStreamableHttpConfig(
     description,
     timeout: timeout,
     headers: headers,
+    envs: envs,
   };
 
   return config;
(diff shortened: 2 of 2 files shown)


ui/desktop/src/App.tsx
```diff
@@ -316,6 +316,7 @@ export function AppInner() {
   const [didSelectProvider, setDidSelectProvider] = useState<boolean>(false);
 
   const navigate = useNavigate();
+  const setView = useMemo(() => createNavigationHandler(navigate), [navigate]);
```
DOsinga: we have this pattern N times now, we can probably do something less repetitive

ui/desktop/src/components/ExtensionInstallModal.tsx
```diff
@@ -41,6 +42,7 @@ interface ExtensionModalConfig {
 
 interface ExtensionInstallModalProps {
   addExtension?: (name: string, config: ExtensionConfig, enabled: boolean) => Promise<void>;
+  setView?: (view: View, options?: ViewOptions) => void;
```
DOsinga: should this be optional?

ui/desktop/src/components/settings/extensions/deeplink.ts
```diff
@@ -193,6 +193,15 @@ export async function addExtensionFromDeepLink(
...
+
+    // Show success toast and navigate to extensions page
+    toastService.success({
+      title: 'Extension Installed',
+      msg: `${config.name} extension has been installed successfully. Start a new chat session to use it.`,
+    });
+
+    // Navigate to extensions page to show the newly installed extension
+    setView('extensions', { deepLinkConfig: config, showEnvVars: false, extensionId: config.name });
   } catch (error) {
```
DOsinga: maybe drop the catch

ui/desktop/src/components/settings/extensions/deeplink.ts
```diff
@@ -180,9 +180,9 @@ export async function addExtensionFromDeepLink(
     config.type === 'streamable_http' && config.headers && Object.keys(config.headers).length > 0;
 
   if (hasEnvVars || hasHeaders) {
-    console.log('Environment variables or headers required, redirecting to settings');
+    console.log('Environment variables or headers required, redirecting to extensions');
     console.log('Calling setView with:', { deepLinkConfig: config, showEnvVars: true });
-    setView('settings', { deepLinkConfig: config, showEnvVars: true });
+    setView('extensions', { deepLinkConfig: config, showEnvVars: true });
```
DOsinga, line 195: we're now going to extensions whether you had this set or no. but in the second case we provide the extensionId. what's the difference? and also can we unify the two paths?

ui/desktop/src/components/settings/extensions/deeplink.ts
```diff
@@ -193,6 +193,15 @@ export async function addExtensionFromDeepLink(
     // It will be activated when the next session starts
     console.warn('Extension will be added to config but requires a session to activate');
     await addExtensionFn(config.name, config, true);
+
+    // Show success toast and navigate to extensions page
+    toastService.success({
+      title: 'Extension Installed',
+      msg: `${config.name} extension has been installed successfully. Start a new chat session to use it.`,
+    });
```
DOsinga: I dont think we need the toast and the log

DOsinga (approved):

ui/desktop/src/components/settings/extensions/deeplink.ts
zanesq: Good catch, should probably send extensionId in both cases will update 👍

zanesq: Changed to use the `name` from `deepLinkConfig` instead of passing `extensionId` separately since thats what extensionId was anyway :) not sure what you meant by unify the paths, in this case it returns after because it doesn't need to call  `addExtensionFn` and the toast message is shown after the modal installation in this case

ui/desktop/src/App.tsx
zanesq: done but added your favorite a new hook unless you have better ideas

amed-xyz (approved):

PR #4868, opened 2025-09-28
Platform extensions sketch

DOsinga:
## Platform Extensions

This introduces a new type of extension, platform extensions. these are extensions that are tied tightly to the platform and can (currently) read and write to the current session. 

It also makes description a required field on the extensions for consistency


crates/goose/src/agents/todo_extension.rs
```diff
@@ -0,0 +1,220 @@
...
+                completions: None,
+                experimental: None,
+                logging: None,
+            },
+            server_info: Implementation {
+                name: "todo".to_string(),
+                version: "1.0.0".to_string(),
+            },
+            instructions: Some(
+                "Manage TODO lists - read and write task planning content.".to_string(),
```
DOsinga: we should expand this of course

crates/goose/src/agents/agent.rs
```diff
@@ -504,91 +493,6 @@ impl Agent {
                 "Frontend tool execution required".to_string(),
                 None,
             )))
-        } else if tool_call.name == TODO_READ_TOOL_NAME {
```
angelahning, line 524: nice

crates/goose/src/agents/extension.rs
```diff
@@ -171,29 +202,38 @@ pub enum ExtensionConfig {
...
         display_name: Option<String>, // needed for the UI
-        description: Option<String>,
         timeout: Option<u64>,
-        /// Whether this extension is bundled with goose
+        #[serde(default)]
+        bundled: Option<bool>,
+        #[serde(default)]
+        available_tools: Vec<String>,
+    },
+    /// Platform extensions that have direct access to the agent etc and are part of the  the goose binary
```
angelahning: nit: "are part of the  the goose binary"

crates/goose/src/agents/todo_tools.rs
```diff
@@ -1,160 +0,0 @@
...
-/// This tool reads the entire task planner file content as a string.
-/// It is marked as read-only and safe to use repeatedly.
-///
-/// # Returns
-/// A configured `Tool` instance for reading task planner content
-pub fn todo_read_tool() -> Tool {
-    Tool::new(
-        TODO_READ_TOOL_NAME.to_string(),
-        indoc! {r#"
-            Read the entire TODO file content.
```
angelahning, line 22: Have these tool instructions been ported somewhere else?

crates/goose/src/config/extensions.rs
```diff
@@ -29,10 +30,29 @@ pub struct ExtensionConfigManager;
...
 impl ExtensionConfigManager {
     fn get_extensions_map() -> Result<HashMap<String, ExtensionEntry>> {
-        let config = Config::global();
-        Ok(config
+        let mut extensions_map = Config::global()
             .get_param(EXTENSIONS_CONFIG_KEY)
-            .unwrap_or_else(|_| HashMap::new()))
+            .unwrap_or_else(|_| HashMap::new());
+
+        if !extensions_map.is_empty() {
```
angelahning, line 88: Why do we only enable the platform extensions if the user has existing extensions?

crates/goose-server/src/routes/extension.rs
```diff
@@ -112,9 +30,6 @@ async fn add_extension(
         request.session_id
     );
 
-    let session_id = request.session_id.clone();
-    let extension_request = request.config;
-
     // If this is a Stdio extension that uses npx, check for Node.js installation
     #[cfg(target_os = "windows")]
     if let ExtensionConfigRequest::Stdio { cmd, .. } = &extension_request {
```
angelahning: does this compile? here's referring to ExtensionConfigRequest::stdio which has been deleted

crates/goose/src/config/extensions.rs
DOsinga: good question; I saw this happening where it would get here and not have any extensions yet and then if we insert it here, something else is not going to set their defaults. obviously would be better to solve the issue in that something else, but that feels like it is inside some hairy client code. we should get a good opinion though on where in general the default extensions are set. I'm not sure !

crates/goose-server/src/routes/extension.rs
DOsinga: ooh, good catch. probably no longer compiles on windows. it does beg the question what this windows specific code is doing here and why it is installing a specific version of node, but still.

crates/goose/src/agents/todo_extension.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    fn get_tools() -> Vec<Tool> {
+        fn create_schema(json_value: Value) -> Arc<Map<String, Value>> {
+            Arc::new(json_value.as_object().unwrap().clone())
+        }
+
+        vec![
+            Tool {
+                name: "todo_read".into(),
+                description: Some("Read the current TODO list content".into()),
```
angelahning: description needs to be expanded. Same for the write tool below.

crates/goose-cli/src/commands/configure.rs
```diff
@@ -1048,23 +1014,16 @@ pub fn configure_extensions_dialog() -> Result<(), Box<dyn Error>> {
...
-                        } else {
-                            Ok(())
-                        }
-                    })
-                    .interact()?;
-                Some(desc)
-            } else {
-                None
-            };
+            let description = cliclack::input("Enter a description for this extension:")
```
michaelneale, line 1017: so descriptions are required now - but not often used, how come mandatory now effectively?

crates/goose/src/agents/agent.rs
michaelneale: yeah this is the best bit!

crates/goose/src/agents/extension.rs
```diff
@@ -32,6 +33,37 @@ impl ProcessExit {
     }
 }
 
+pub static PLATFORM_EXTENSIONS: Lazy<HashMap<&'static str, PlatformExtensionDef>> =
+    Lazy::new(|| {
+        let mut map = HashMap::new();
+
+        map.insert(
+            todo_extension::EXTENSION_NAME,
```
michaelneale, line 42: much nicer looking

crates/goose/src/agents/extension.rs
michaelneale: more so than just binary, as the developer MCP etc are also in the binary, but run out of process, these run in process?

crates/goose/src/agents/todo_extension.rs
michaelneale: so this replaces the prompt for existing todos? (or it is still in system prompt?). Existing todos seem to think they are a file which always looks odd, so probably good to play around with this

crates/goose/src/agents/todo_extension.rs
DOsinga: this is a sketch - the idea is to look at the architecture

crates/goose-cli/src/commands/configure.rs
DOsinga: well, we were using them all over the place, just for when we didn't have them we would try to put something together ourselves. this is just an effort to make it all look a bit more like each other.

crates/goose/src/agents/todo_extension.rs
DOsinga: yes, will be. currently todo read & write are tools without an extension, which means there is no extension prompt which means we include it in the system prompt which means that if you switch off todo, it still knows about it. so that's one of the things we're trying to fix here

crates/goose/src/agents/extension.rs
DOsinga: changed it into something better

crates/goose/src/agents/todo_tools.rs
DOsinga: they have now

crates/goose-server/src/routes/extension.rs
DOsinga: fixed

crates/goose-cli/src/commands/configure.rs
michaelneale: that is ok - just wanted to make sure deliberate and won't cause problems loading older configurations

michaelneale:
LGTM - when ready/builds any thing to manually test and watch out for @DOsinga ?

crates/goose/src/agents/todo_extension.rs
```diff
@@ -0,0 +1,263 @@
...
+        }
+    }
+
+    fn get_tools() -> Vec<Tool> {
+        fn create_schema(json_value: Value) -> Arc<Map<String, Value>> {
+            Arc::new(json_value.as_object().unwrap().clone())
+        }
+
+        vec![
+            Tool {
```
alexhancock: Can do `Tool::new` for a bit nicer syntax if we want

crates/goose/src/agents/todo_extension.rs
```diff
@@ -0,0 +1,263 @@
...
+
+pub struct TodoClient {
+    info: InitializeResult,
+    context: PlatformExtensionContext,
+}
+
+impl TodoClient {
+    pub fn new(context: PlatformExtensionContext) -> Result<Self> {
+        let info = InitializeResult {
+            protocol_version: ProtocolVersion::V_2025_03_26,
```
alexhancock, line 29: if there is a constant for the june version we can start using it. i was going to take a pass and make that the default in rmcp. we more closely resemble the june version than march.

DOsinga: sounds good. of course this is internal only so for now I doubt we'll check it, bit still

crates/goose/src/agents/todo_extension.rs
DOsinga: oh nice

crates/goose/src/agents/todo_extension.rs
```diff
@@ -0,0 +1,263 @@
...
+                annotations: None,
+                output_schema: None,
+            },
+        ]
+    }
+}
+
+#[async_trait]
+impl McpClientTrait for TodoClient {
+    async fn list_resources(
```
jamadeo, line 217: if these are going to act like mcp servers, it would be really nice to be able to define them using rmcp's server sdk, e.g.

```
impl Todo {
    #[tool(name = "todo_add")]
    pub async fn todo_add(...)
}
```

they could still be built with the platform extension context and have all the same access, but they'd be easier to write and easily port-able to out-of-process MCP servers

jamadeo (commented):
Nice. I think passing the context to a per-session client is great, and opens these up to be really quite powerful.

I'm fine with description being non-optional in general, but I don't think users should have to create a description when adding an MCP server. (e.g. I go find the github mcp, add it, I should not have to also type "github" in a description field.)

michaelneale:
image


image


but my config is: 

```yaml
GOOSE_MODEL: claude-sonnet-4-5-20250929
ANTHROPIC_HOST: https://api.anthropic.com
GOOSE_PROVIDER: anthropic
OTEL_EXPORTER_OTLP_ENDPOINT: https://goose-otlp-proxy.sqprod.co
GOOSE_RECIPE_GITHUB_REPO: squareup/goose-recipes
extensions:
  developer:
    enabled: true
    type: builtin
    name: developer
    display_name: Developer Tools
    description: null
    timeout: 300
    bundled: true
    available_tools: []
```


so something is broken with the extension loading (it won't load or even recognise a built in) specifically with this branch

michaelneale (changes requested):
isn't picking up built in extensions or configured extensions (see comment)

DOsinga:
> isn't picking up built in extensions or configured extensions (see comment)

I couldn't reproduce this, but it did asked me if I wanted to fix the install of the todo extension. that should be fixed now. it might be the same thing.  this is what I am seeing:

image

michaelneale:
@DOsinga still the same - not showing any extensions or finding them, and have pulled latest

image


no idea why - but any other branch is fine, it is just this one

michaelneale:
ok testing with production build `cargo build -r` is where I am seeing it not work for some reason...

michaelneale (approved):
ok think have hammered on this enough - cli does work now - so I am happy (weird how it loaded different - I guess longer term that code path will be replaced with goosed so all good). 

I also note that todo now shows up an extension you can turn off, but it is not in the config.yaml - which means you can't "add it" later on - is this an issue? (maybe it will always just show up in any list which is fine and ideal). Again - cli only.

(openapi schema isn't happy though - but when build is ok I think this makes sense)

DOsinga:
> ok think have hammered on this enough - cli does work now - so I am happy (weird how it loaded different - I guess longer term that code path will be replaced with goosed so all good).
> 
> I also note that todo now shows up an extension you can turn off, but it is not in the config.yaml - which means you can't "add it" later on - is this an issue? (maybe it will always just show up in any list which is fine and ideal). Again - cli only.
> 
> (openapi schema isn't happy though - but when build is ok I think this makes sense)

yeah, good analysis. todo always shows up, as it is a platform tool, so all you can do is switch it off. if you delete it from the config, we add it again. I think this is how all bundled tools should work? not sure. also not sure on how it works now and when they are re-added (i.e. like with developer tool, the desktop will adde them if deleted?). might need to do a follow up to clean this up

PR #4871, opened 2025-09-29
fix mcp integration test flakiness

jamadeo:
Concurrent `cargo run` commands don't work so well. This changes it to build the binary and then use that.


jamadeo:
Thanks @michaelneale for fixing the breaking test. The passing tests are still somewhat flaky for me when I run them locally, although I've yet to see them fail on CI in this way.

I tracked it down to the fact that it uses `cargo run` to run the MCP replay process. It seems that when cargo locks itself, the process is killed somewhere and the connection closed.

alexhancock:
when they flake, how do they flake?

i've experienced difficulty with recording (other output included etc), but not once we have a clean set of recordings made

michaelneale:
I think no longer needed - we had a bad merge to main which had a failing check (shoudlnt
' have been merged) which caused this issue I believe.

jamadeo:
> when they flake, how do they flake?

@alexhancock this affected replay mode, where the concurrent `cargo run` commands might mean one or more of the server processes exits without doing anything. The error you'd see is something like "connection closed before init response" or something like that.

alexhancock:
> > when they flake, how do they flake?
> 
> @alexhancock this affected replay mode, where the concurrent `cargo run` commands might mean one or more of the server processes exits without doing anything. The error you'd see is something like "connection closed before init response" or something like that.

I see

PR #4872, opened 2025-09-29
Don't set agent props twice

DOsinga:
Had another look at the code we ended up merging and I think we missed a trick; that lock returns the agent to the rest of the code so the way it was, it would always reset the provider and scheduler.


crates/goose/src/execution/manager.rs
```diff
@@ -91,43 +91,40 @@ impl AgentManager {
...
 
-        match &mode {
-            SessionExecutionMode::Interactive | SessionExecutionMode::Background => {
-                debug!("Setting scheduler on agent for session {}", session_id);
-                agent.set_scheduler(Arc::clone(&self.scheduler)).await;
+        if is_new {
+            match &mode {
+                SessionExecutionMode::Interactive | SessionExecutionMode::Background => {
+                    debug!("Setting scheduler on agent for session {}", session_id);
+                    agent.set_scheduler(Arc::clone(&self.scheduler)).await;
```
alexhancock: I know this is existing code but why do we `agent.set_scheduler` for an interactive session?

alexhancock (approved):
It looks good to me. Perhaps add a test that would have failed before the fix.

crates/goose/src/execution/manager.rs
DOsinga: it's not a great reason, but goose can create schedules for itself and in order to that, the agent needs to have access to the scheduling service

alexhancock: ah, ok

DOsinga:
> It looks good to me. Perhaps add a test that would have failed before the fix.

I dont think you can really. this is better because it doesn't lock up the entire agent manager while creating an agent, but the previous situation would still work

PR #4873, opened 2025-09-29
Dead code cleanup

DOsinga:
Dead code that is really not actually used it seems


michaelneale:
yep LGTM - it is and always has been enforced in the desktop client only (I think at one point we wanted to try it on the agent/cli side but that was abandoned)

PR #4832, opened 2025-09-25
lowercase g in goose

angiejones:
changing occurrences of Goose to goose - first pass

PR #4874, opened 2025-09-29
Revert "lowercase g in goose"

michaelneale:
Reverts block/goose#4832

seeing if this resolves the build (maybe some case insensitivity inconsistency)

we may need to do this, and then do a different change just with the casing of the doc site


alexhancock:
what build issues were you noticing @michaelneale? `main` has been ok today on my side

michaelneale:
@alexhancock oh this was an older attempted - took me a few hours but I was able to patch up main. The lowercase g removed some data which was needed by the tests and was failing, so added it back in, fixed openapi stuff etc - but if that didn't work, rolling back was an option (honestly the rename PR was broken so am not sure how it was merged, it shouldn't have been in that form) - but my aim was to get it clean before your work day so no one noticed!

alexhancock:
> my aim was to get it clean before your work day so no one noticed!

You did marvelously!

PR #4878, opened 2025-09-29
feat(cli): add `path` & `limit` to `session list` command

exitcode0:
adds `-p/--path` and `-l/--limit` to `goose session list` command

## Pull Request Description

---

**Email**:

### DIFF
--- a/crates/goose-cli/src/cli.rs
+++ b/crates/goose-cli/src/cli.rs
@@ -128,7 +128,7 @@ enum SessionCommand {
             long = "limit",
             help = "Limit the number of results",
         )]
-        limit: Option<u32>,
+        limit: Option<usize>,
     },
     #[command(about = "Remove sessions. Runs interactively if no ID or regex is provided.")]
     Remove {
@@ -198,7 +198,7 @@ enum SchedulerCommand {
         id: String,
         /// Maximum number of sessions to return
         #[arg(long, help = "Maximum number of sessions to return")]
-        limit: Option<u32>,
+        limit: Option<usize>,
     },
     /// Run a scheduled job immediately
     #[command(about = "Run a scheduled job immediately")]
--- a/crates/goose-cli/src/cli.rs
+++ b/crates/goose-cli/src/cli.rs
@@ -98,9 +98,6 @@ fn parse_key_val(s: &str) -> Result<(String, String), String> {
 enum SessionCommand {
     #[command(about = "List all available sessions")]
     List {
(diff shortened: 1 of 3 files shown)


crates/goose-cli/src/commands/session.rs
```diff
@@ -114,14 +114,29 @@ pub async fn handle_session_remove(id: Option<String>, regex_string: Option<Stri
...
+
     if ascending {
         sessions.sort_by(|a, b| a.updated_at.cmp(&b.updated_at));
     } else {
         sessions.sort_by(|a, b| b.updated_at.cmp(&a.updated_at));
     }
 
+    if let Some(n) = limit {
+        sessions.truncate(n as usize);
+    }
```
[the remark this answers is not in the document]
exitcode0: 🤞 i've addressed this correctly in this follow-up commit [832c0fe](https://github.com/block/goose/pull/4878/commits/832c0fe4b9a7e58c801befff05dd5aa63f16bb95)

crates/goose-cli/src/cli.rs
```diff
@@ -184,7 +194,7 @@ enum SchedulerCommand {
         id: String,
         /// Maximum number of sessions to return
```
DOsinga: the command and the arg description are the same. remove

crates/goose-cli/src/cli.rs
```diff
@@ -115,6 +115,16 @@ enum SessionCommand {
             long_help = "Sort sessions by date in ascending order (oldest first). Default is descending order (newest first)."
         )]
         ascending: bool,
+
+        #[arg(
+            short = 'p',
+            long = "path",
+            help = "Filter sessions by working directory"
+        )]
+        path: Option<PathBuf>,
```
DOsinga: can we name this working_dir? we use path elsewhere to refer to where the session lives. that's no longer a thing since we moved to a db, but still it might confuse people

crates/goose-cli/src/commands/session.rs
```diff
@@ -114,14 +114,35 @@ pub async fn handle_session_remove(id: Option<String>, regex_string: Option<Stri
     remove_sessions(matched_sessions).await
 }
 
-pub async fn handle_session_list(verbose: bool, format: String, ascending: bool) -> Result<()> {
+pub async fn handle_session_list(
+    verbose: bool,
```
DOsinga: while you are at it, can you drop this flag? doesn't seem to do anything sensible

PR #4880, opened 2025-09-29
make agent manager singleton

yingjiehe-xyz:
Agent manager should be created every time and can be shared in different classes


crates/goose/src/execution/manager.rs
```diff
@@ -44,6 +54,17 @@ impl AgentManager {
         Ok(manager)
     }
 
+    // Public method to get or initialize the singleton instance
+    pub async fn instance(max_sessions: Option<usize>) -> Result<Arc<Self>> {
```
DOsinga: a singleton shouldn't take a parameter. that or it should fail if you try to get one with a different max_sessions

crates/goose/src/execution/manager.rs
```diff
@@ -12,17 +12,27 @@ use etcetera::{choose_app_strategy, AppStrategy};
...
     default_provider: Arc<RwLock<Option<Arc<dyn crate::providers::base::Provider>>>>,
 }
 
 impl AgentManager {
-    pub async fn new(max_sessions: Option<usize>) -> Result<Self> {
+    // This new method is ONLY used for TEST.
+    pub async fn new_for_test(max_sessions: Option<usize>) -> Result<Self> {
+        // Same implementation as the private new() method
+        // This bypasses the singleton pattern for tests
+        Self::new(max_sessions).await
```
DOsinga: I don't think this is the way; we end up testing a non singleton behavior while in prod we have a singleton.

I would suggest adding something like:

```
#[cfg(test)]
   pub async fn reset_for_test() {
       unsafe {
           AGENT_MANAGER.take();
       }
   }
```

that way you can call that in the tests when you need a new singleton.

crates/goose/src/execution/manager.rs
yingjiehe-xyz: Done

crates/goose/src/execution/manager.rs
yingjiehe-xyz: `AGENT_MANAGER.take();` should not work here as we borrow the immutable as mutable, updated code

crates/goose/src/execution/manager.rs
```diff
@@ -12,25 +12,41 @@ use etcetera::{choose_app_strategy, AppStrategy};
...
 pub struct AgentManager {
     sessions: Arc<RwLock<LruCache<String, Arc<Agent>>>>,
     scheduler: Arc<dyn SchedulerTrait>,
     default_provider: Arc<RwLock<Option<Arc<dyn crate::providers::base::Provider>>>>,
 }
 
 impl AgentManager {
-    pub async fn new(max_sessions: Option<usize>) -> Result<Self> {
+    /// Reset the global singleton - ONLY for testing
+    pub fn reset_for_test() {
```
DOsinga, line 30: you do want to annotate this with test only though

crates/goose/tests/execution_tests.rs
```diff
@@ -25,8 +25,10 @@ mod execution_tests {
     }
 
     #[tokio::test]
+    #[serial]
     async fn test_session_isolation() {
-        let manager = AgentManager::new(None).await.unwrap();
+        AgentManager::reset_for_test();
```
DOsinga, line 30: why do we need to reset these? I'd imagine you only want to reset for testing for situations where you actually want to test that a new singleton does somethign different

crates/goose/src/execution/manager.rs
```diff
@@ -44,6 +60,17 @@ impl AgentManager {
         Ok(manager)
     }
 
+    /// Get or initialize the singleton instance
```
DOsinga: here and elsewhere, delete the comments they don't add

crates/goose/tests/execution_tests.rs
```diff
@@ -263,59 +266,85 @@ mod execution_tests {
...
         let result = manager.remove_session(&session).await;
         assert!(result.is_err());
         assert!(result.unwrap_err().to_string().contains("not found"));
+
+        AgentManager::reset_for_test();
+    }
+
+    #[tokio::test]
+    #[serial]
+    async fn test_singleton_persistence_across_calls() {
```
DOsinga: I dont think this test is adding anything

crates/goose/src/execution/manager.rs
yingjiehe-xyz: I don't think it work here, we can use `#[cfg(test)]` if the tests are agent class are in the same package, otherwise, tests cannot find the `reset_for_test`

see error:
```
no function or associated item named `reset_for_test` found for struct `AgentManager` in the current scope
```

crates/goose/tests/execution_tests.rs
yingjiehe-xyz: Done

crates/goose/tests/execution_tests.rs
yingjiehe-xyz: I think it is better to reset every time to keep the tests isolated from each other

DOsinga: yeah,  no. you should isolate these tests from the rest of the system, but if you write tests for a singleton and then reset the singleton all the time, what are you testing?

yingjiehe-xyz: Do you mean that we want to cover the singleton access crossing multiple test cases? Unit tests should only cover each simple functionalities of the methods, we have a test case to test the multiple access as the same time, and others for the main logic, to make all these tests isolated, we need such reset.

PR #4884, opened 2025-09-29
fix(providers): update Claude Sonnet 4 model identifier

jmdaly:
## Pull Request Description

Remove date suffix from claude-sonnet-4-20250514 model name in GitHub Copilot provider configuration. Updates both the main models list and streaming models list to use the "claude-sonnet-4" identifier, which is the model name in GitHub's API.

Prior to this, whenever Sonnet 4 was used with the `github_copilot` provider in goose, issuing a request to goose would result in being returned to the prompt each time. Then I'd have to send another message to get it to keep going. After that message, it would return to the prompt, etc. This fix seems to make Sonnet 4 work correctly, in the way the other models do.

---

**Email**: jmdaly@gmail.com


angelahning (approved):
interesting. Thanks!

PR #4885, opened 2025-09-29
alexhancock/mcp-crate-cleanup

alexhancock:
Re-introducing https://github.com/block/goose/pull/4800 with a fix commit for some argument formatting issues that came up with providers

https://github.com/block/goose/commit/ca160bf7cfb8e252d2cd9f3b666caf8c54eb3e75


alexhancock:
The line counts (add/remove) are now the same, but I promise it's better code now 😆 

More tests, less custom stuff, more using a standard lib

amed-xyz (approved):

PR #4888, opened 2025-09-29
feat(nightly): build nightlies from main shas

alexhancock:
We ran into a tag permission issue, and there are reasons we don't want to change those repository rules.

```
Creating tag: 1.8.0-nightly.20250924
remote: error: GH013: Repository rule violations found for refs/tags/1.8.0-nightly.20250924.        
remote: Review all repository rules at https://github.com/block/goose/rules?ref=refs%2Ftags%2F1.8.0-nightly.20250924        
remote: 
remote: - Cannot create ref due to creations being restricted.        
remote: 
To https://github.com/block/goose
 ! [remote rejected]       1.8.0-nightly.20250924 -> 1.8.0-nightly.20250924 (push declined due to repository rule violations)
```

This PR attempts to build nightlies directly from main shas instead

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,4 +1,5 @@
-# This workflow is for nightly releases, automatically triggered at midnight US Eastern
+# A release from main. Automatically triggered at midnight US Eastern.
+
 name: Nightly Build
 
 on:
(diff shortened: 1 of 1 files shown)


alexhancock:
This may need further revs to get a couple things working, but it's in a better state than current. Merging

PR #4890, opened 2025-09-29
Lifei/investigation render recipe

lifeizhou-ap:
## Pull Request Description

---

**Email**:


DOsinga:
is this stale?

lifeizhou-ap:
this is only for investigation. will create a separate pr for this piece of work.

PR #4892, opened 2025-09-30
feat: add commit command for AI-generated commits

capttrousers:
## Pull Request Description

Add new 'goose commit' command that analyzes git changes and generates conventional commit messages using an interactive AI session.

The command supports both staged and unstaged changes, with options to review git history for context. It integrates seamlessly with goose's existing session infrastructure and supports all standard CLI options.

Implementation details:
- Add CommitConfig struct to encapsulate command parameters
- Add handle_commit() function in commands/commit.rs
- Add Command::Commit variant to CLI enum with comprehensive options
- Add indoc dependency (2.0.5) for cleaner multi-line strings
- Use --staged-only flag to commit only staged changes
- Support standard goose options: extensions, provider, model, debug, etc.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2701,6 +2701,7 @@ dependencies = [
  "goose-mcp",
  "http 1.2.0",
  "indicatif",
+ "indoc",
  "is-terminal",
  "jsonschema",
  "nix 0.30.1",
--- a/crates/goose-cli/Cargo.toml
+++ b/crates/goose-cli/Cargo.toml
@@ -58,6 +58,7 @@ tokio-util = { version = "0.7.15", features = ["compat"] }
 is-terminal = "0.4.16"
 anstream = "0.6.18"
 url = "2.5.7"
+indoc = "2.0.5"
 
 [target.'cfg(target_os = "windows")'.dependencies]
 winapi = { version = "0.3", features = ["wincred"] }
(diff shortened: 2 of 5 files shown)


DOsinga:
should this not be a goose recipe?

capttrousers:
@DOsinga I suppose it could be a recipe instead, although I don't have any experience using those. I could see this being a slash command available in the GUI too.

I was inspired by something I swear I saw in the Claude Code docs a while back where it had a top level `claude commit` command. But now I can't find it 🤷

blackgirlbytes:
i think this is cool 🔥

blackgirlbytes:
if you did want it to be a recipe though..we have a recipe cookbook that you can contribute to https://block.github.io/goose/recipes/

submissions are eligible for prizes!!

DOsinga:
> @DOsinga I suppose it could be a recipe instead, although I don't have any experience using those. I could see this being a slash command available in the GUI too.
> 
> I was inspired by something I swear I saw in the Claude Code docs a while back where it had a top level `claude commit` command. But now I can't find it 🤷

so with a recipe you can ask the user for parameters that then get included in the prompt etc, so that's kinda nice

we're looking into making recipes runnable through slash command, so that's a great suggestion

capttrousers:
I like the idea of using the recipe system because that's a good extension point. 

But I also like the simplicity of just running `goose commit` in a dirty repo, to have it kick off an interactive session after it's committed everything, and gotten context on the current work.

What do yall think about having `goose commit` just be an alias to the "git commit" recipe? are there any recipes that ship with the goose installable?

DOsinga:
we currently don't ship recipes with goose. you can set a github repository as a source though, maybe we should start doing that for the OSS version to? /cc @angiejones

DOsinga:
I'm going to close this. thanks for the contribution though. can you create this as a recipe and submit that? we'll follow up to see how we best integrate submitted recipes in goose

PR #4895, opened 2025-09-30
Lifei/create save recipe to file

lifeizhou-ap:
## Pull Request Description

- Created endpoints so that frontend can call the server side to import recipe and save recipe to files

  1.  `/recipes/save_to_file` 
      - to be used in the import, edit, create recipe flows
      - include validate the duplicate files and duplicate title in the server side so that we can remove the logic of validate recipe schema and these validation on the frontend
      - Now it only should accept the recipe (not the recipe wrapper any more. ie. with name, is_global, is_archived....)
     
  2.  `/recipe/parse`
     - to be used to parse the recipe when import recipe from file.  For importing recipe from deeplink, we have already used decode_recipe endpoint

- Removed the is_global attribute from list_recipes endpoint payload
- Created a general ErrorResponse to return error message from server side

**Note**
I only changed the ImportFileDialog  a bit to just test whether these end points works. I did not make much changes/removal on Frontend to avoid merge conflicts before Zane's PR is merged


crates/goose/src/recipe/recipe_library.rs
```diff
@@ -0,0 +1,102 @@
...
+use std::path::PathBuf;
+
+pub fn get_recipe_library_dir(is_global: bool) -> PathBuf {
+    if is_global {
+        choose_app_strategy(APP_STRATEGY.clone())
+            .expect("goose requires a home dir")
+            .config_dir()
+            .join("recipes")
+    } else {
+        std::env::current_dir().unwrap().join(".goose/recipes")
```
DOsinga, line 17: unfortunately this is not going to work. we'd need the directory of the session here, not the current directory.

crates/goose-server/src/routes/errors.rs
```diff
@@ -0,0 +1,24 @@
...
+
+impl IntoResponse for ErrorResponse {
+    fn into_response(self) -> Response {
+        let body = Json(serde_json::json!({
+            "message": self.message,
+        }));
+
+        (self.status, body).into_response()
+    }
+}
```
DOsinga, line 24: this is cute and we certainly need something like this, but I would leave it out for now; the real refactor here is to have something that converts from our errors to a ErrorResponse globally and then apply it everywhere. would love to have it, but I don't think it is in scope /cc @jamadeo

ui/desktop/src/recipe/recipe_management.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export async function parseRecipeFromFile(fileContent: string): Promise<Recipe> {
+  let response = await parseRecipe({
+    body: {
+      content: fileContent,
+    },
+    throwOnError: true,
+  });
+  return response.data.recipe;
+}
```
DOsinga: I know we do this elsewhere, but I am not really a fan of these thin layers that hide where things happen - can we not just call things directly?

ui/desktop/src/components/recipes/ImportRecipeForm.tsx
```diff
@@ -171,25 +143,24 @@ export default function ImportRecipeForm({ isOpen, onClose, onSuccess }: ImportR
...
         recipe.title = value.recipeTitle.trim();
 
-        const titleValidationError = await validateTitleUniqueness(
-          value.recipeTitle.trim(),
-          value.global
-        );
-        if (titleValidationError) {
-          throw new Error(titleValidationError);
-        }
+        // const titleValidationError = await validateTitleUniqueness(value.recipeTitle.trim());
```
DOsinga: what? why would names have to be unique :(

crates/goose-server/src/routes/recipe.rs
```diff
@@ -286,6 +352,8 @@ pub fn routes(state: Arc<AppState>) -> Router {
         .route("/recipes/scan", post(scan_recipe))
         .route("/recipes/list", get(list_recipes))
         .route("/recipes/delete", post(delete_recipe))
+        .route("/recipes/save_to_file", post(save_recipe_to_file))
```
DOsinga: let's call this just save and save_recipe. no need to tell people we save this to file

crates/goose/src/recipe/recipe_library.rs
```diff
@@ -0,0 +1,102 @@
...
+        "untitled-recipe".to_string()
+    } else {
+        base_name
+    };
+    format!("{}.yaml", filename)
+}
+
+pub fn save_recipe_to_file(
+    recipe: Recipe,
+    is_global: Option<bool>,
```
DOsinga, line 70: I think we should drop this now. all writes should go to the global version - we can read from local for a while though

crates/goose/src/recipe/recipe_library.rs
```diff
@@ -0,0 +1,102 @@
...
+                    continue;
+                };
+                recipes_with_path.push((path, recipe));
+            }
+        }
+    }
+    Ok(recipes_with_path)
+}
+
+pub fn list_all_recipes_from_library() -> Result<Vec<(PathBuf, Recipe)>> {
```
DOsinga, line 43: I would prefer it if we would integrate this with how the CLI works; there we have a function that reads recipes from various places; can we not add this to that so we get a clear overview

crates/goose-server/src/routes/recipe.rs
```diff
@@ -286,6 +352,8 @@ pub fn routes(state: Arc<AppState>) -> Router {
...
+        .route("/recipes/parse", post(parse_recipe))
```
DOsinga, line 356: should we replace /decode with this too? and really make parse so that it also takes base64 etc

crates/goose-server/src/routes/recipe.rs
amed-xyz: Was thinking of `save_to_disk` but yeah I agree the extra info isn't needed

crates/goose-server/src/routes/recipe_utils.rs
```diff
@@ -31,59 +28,31 @@ fn short_id_from_path(path: &str) -> String {
...
-    recipe_manifests_with_path.extend(load_recipes_from_path(&local_recipe_path, false)?);
-    recipe_manifests_with_path.extend(load_recipes_from_path(&global_recipe_path, true)?);
+    for (file_path, recipe) in recipes_with_path {
+        let Ok(last_modified) = fs::metadata(file_path.clone())
+            .map(|m| chrono::DateTime::<chrono::Utc>::from(m.modified().unwrap()).to_rfc3339())
+        else {
+            continue;
+        };
+        let recipe_metadata =
+            RecipeManifestMetadata::from_yaml_file(&file_path).unwrap_or_else(|_| {
```
amed-xyz, line 41: The current import UI also supports JSON so we may want to consider that too

crates/goose/src/recipe/recipe_library.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+}
+
+pub fn list_recipes_from_library(is_global: bool) -> Result<Vec<(PathBuf, Recipe)>> {
+    let path = get_recipe_library_dir(is_global);
+    let mut recipes_with_path = Vec::new();
+    if path.exists() {
+        for entry in fs::read_dir(path)? {
+            let path = entry?.path();
+            if path.extension() == Some("yaml".as_ref()) {
```
amed-xyz: ditto

crates/goose/src/recipe/recipe_library.rs
amed-xyz: This should also be reflected in the UI by removing the radio buttons

ui/desktop/src/components/recipes/ImportRecipeForm.tsx
amed-xyz: Name and title are different attributes. Name is what's currently displayed in the UI, title is only used for the file name. We added uniqueness check because otherwise it'd overwrite existing recipes altogether, which is also a confusing UX

ui/desktop/src/components/recipes/ImportRecipeForm.tsx
```diff
@@ -171,25 +143,24 @@ export default function ImportRecipeForm({ isOpen, onClose, onSuccess }: ImportR
...
-        const validationResult = validateRecipe(recipe);
-        if (!validationResult.success) {
-          const errorMessages = getValidationErrorMessages(validationResult.errors);
-          throw new Error(`Recipe validation failed: ${errorMessages.join(', ')}`);
-        }
+        // const validationResult = validateRecipe(recipe);
+        // if (!validationResult.success) {
+        //   const errorMessages = getValidationErrorMessages(validationResult.errors);
+        //   throw new Error(`Recipe validation failed: ${errorMessages.join(', ')}`);
+        // }
```
amed-xyz: Why are we commenting out this?

ui/desktop/src/components/recipes/ImportRecipeForm.tsx
DOsinga: uhm, if name and title are different attributes and title is use for the file *name*, I think we are in trouble

crates/goose-server/src/routes/recipe_utils.rs
DOsinga: yaml is a superset of json so that should be alright

crates/goose-server/src/routes/errors.rs
jamadeo: axum is very permissive about what you can return: https://docs.rs/axum/latest/axum/response/index.html

You can already return a (StatusCode, String) and it will convert

DOsinga: cool. we should do this for everything some time. maybe goose can

ui/desktop/src/components/recipes/ImportRecipeForm.tsx
lifeizhou-ap: As noted in the PR description, the ImportRecipeFrom.tsx change is only for testing.  The UI code change will start after Zane's PR is merged

crates/goose/src/recipe/recipe_library.rs
lifeizhou-ap: How about the current `non global` recipes that persists on user's machine? (ie: they were saved on the current working directory before) Shall we still support them?

Are we also going to have recipes in the directory of the session in the future?

crates/goose/src/recipe/recipe_library.rs
lifeizhou-ap: will address in the next PR

crates/goose/src/recipe/recipe_library.rs
lifeizhou-ap: We can have a separate PR for this

ui/desktop/src/components/recipes/ImportRecipeForm.tsx
lifeizhou-ap: i feel name and title are very confusing in the frontend, and they are used interchangeably. That's why I took this chance to make it simpler. 

Currently `name` is defined in the wrapper of the recipe in the recipe management.  Since we don't want the wrap version of the recipe, we can use recipe title instead to display on the UI.  


Also we can the title to work out the file name for new recipes or imported recipe.  

For the existing recipe, we use the recipe id to work out the existing recipe file path at the server side.  It would not use the name or recipe title.

ui/desktop/src/recipe/recipe_management.ts
lifeizhou-ap: These functions are used in multiple times, that's why I extract them here

DOsinga: but what function, this is just calling parseRecipe and naming it parseRecipeFromFile? why the overhead

lifeizhou-ap: At the moment the in ImportFileModal calls parseRecipe 2 times.  The UI change I included is just for testing. I have revert UI change in this PR and I am going to merge this backend change PR soon.

When we get the latest changes from Zane's PR, we can check whether the Modal has to call parseRecipe multiple time

Issue #4893, opened 2025-09-30
Cant name a session in cli


jsibbison-square:
**Describe the bug**

Since #4648 we can't name a cli session anymore with `goose session -n name`. It tries to resume a session and fails. 

**To Reproduce**
`goose session -n brand_new_name`

produces

`Error: No session found with name 'brand_new_name'`

**Expected behavior**
We should be able to name sessions for 2 reasons:
- Avoid session name conflicts that prevents parallel goose usage. Previously the naming scheme was second based datestamp so that would fail if you created 2 sessions in the same second. Not sure if the new db approach makes this redundant
- Enables resuming named sessions.

**Please provide following information:**
 - **OS & Arch:** osx
 - **Interface:** CLI
 - **Version:** HEAD (later than 1.9 release)

DOsinga:
yeah, I think we should drop the concept of named sessions. what we called names really wasn't, it was ids:
>Previously the naming scheme was second based datestamp so that would fail if you created 2 sessions in the same second. Not sure if the new db approach makes this redundant

it wouldn't fail. it would just silently override. the current approach is that the keys are date_<idx> with idx being unique. since elsewhere in the system the description field effectively works as the name, that seems like a much better approach and in line with how the desktop works.

note that you can resume a session based on the auto generated description. the CLI doesn't allow you to override the description. the desktop does but unless you have 3 or more user messages, the system will override what you do again

named sessions to me suggests that you can have multiple sessions with the same name, which we never supported

jsibbison-square:
> it wouldn't fail. it would just silently override.

In our use cases we would see it fail with a file lock error iirc. So I guess it would try to override but the file was locked so it would fail ending the session.

I can still produce this failure intermittently now (database is locked error) by running this as a single command to start 2 sessions at the same second:
```bash
./target/debug/goose run -t "say hello only" >run1.log 2>&1 &
./target/debug/goose run -t "say hello only" >run2.log 2>&1 &
wait
```

output
```bash
cat run1.log

thread 'main' panicked at goose/main/crates/goose-cli/src/session/builder.rs:280:10:
called `Result::unwrap()` on an `Err` value: error returned from database: (code: 5) database is locked

Caused by:
    (code: 5) database is locked
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

cat run2.log
starting session | provider: databricks model: goose-gpt-4-1
    working directory: goose/main
Hello!
```

[...]

PR #4896, opened 2025-09-30
Allow better concurrent access

DOsinga:
helps with #4893


DOsinga:
sqlite is pretty speedy for that sort of thing though. definitely faster than reading the file, parsing the json and writing the file back which is what we were previously doing. but yeah, you wouldn't want to power a high volume website with this. then again we can always have a postgres alternative relatively easy now.

PR #4897, opened 2025-09-30
remove anyOf from create_task tool

tlongwell-block:


r0x0d (approved):
can confirm that the solution worked! 

Image

Kvadratni (approved):
Should be fine

PR #4900, opened 2025-09-30
fix optional recipe schema zod validation

zanesq:
## Pull Request Description
Fixes optional recipe validation error when saving a recipe mid chat. Verified locally saving mid chat and from recipe manager.

Fixes
```
Recipe validation failed: author.metadata: Required
```


katzdave (approved):

ui/desktop/src/recipe/validation.ts
```diff
@@ -187,40 +187,30 @@ function openApiSchemaToZod(schema: Record<string, unknown>): z.ZodTypeAny {
           shape[propName] = openApiSchemaToZod(propSchema as Record<string, unknown>);
         }
 
-        // Make optional properties optional
```
DOsinga, line 190: what is this anyway? why do we have our own validation here?

Issue #4881, opened 2025-09-29
CLI developer tool interruption - WARN rmcp::service: Error sending response, id: 72.h


zanesq:
**Describe the bug**
from zacherydrayer

I interrupted goose to give it some new instructions, and it interrupted my prompt with an error message mid-paste - and while i'm _assuming_ this went to stdout and won't make it to stdin, I still made a new goose session to be safe.

```
─── shell | developer ──────────────────────────
command: [REDACTED: 81 chars]

^CThe existing call to developer__shell was interrupted. How would you like to proceed?
Context: ○○○○○○○○○○ 0% (0/200000 tokens)
( O)> [redacted by zanesq user was typing answer and the following message was displayed mid type] 
2025-09-29T14:11:07.072400Z  WARN rmcp::service: Error sending response, id: 72.h
    at /Users/runner/work/goose/goose/.hermit/rust/registry/src/index.crates.io-1949cf8c6b5b557f/rmcp-0.6.2/src/service.rs:825
```

**Expected behavior**
Should allow answering without this warning

**Please provide following information:**
 - **OS & Arch:** Mac arm
 - **Interface:** CLI
 - **Version:** v1.8.0

**Additional context**

PR #4902, opened 2025-09-30
CLI: dont show logs to user

angelahning:
We currently show all the warning messages on the CLI. They could be useful for debugging or identifying potential problems for developers, but they are probably useless and confusing to the CLI users. 

This PR filters the console logging to just ERRORS (the file logs still track everything fyi)

related: https://github.com/block/goose/issues/4881


DOsinga:
should we even do that? as the linked issue shows, we chuck the error right into the "UI" - is there a case where we log an error and not bubble it up to the user?

michaelneale:
I like this - the warning messages aren't helpful inline in the CLI and mostly look bad

michaelneale (approved):

PR #4909, opened 2025-10-01
feat: Disable llm session description generation to reduce latency

jsibbison-square:
## Pull Request Description

**Before**
Every session during the first few messages an llm is called with messages to generate an accurate session description.

**After**
By setting `DISABLE_GENERATED_SESSION_NAMING=true` env variable the session description will always be 'no session name' and llm will not be called

**Why**
This reduces latency on the first few turns of the agents by avoid an extra llm call (even with a fast model its still time). When we are doing automation use cases of the repetitive tasks there is no value to this description.

**Usage**
When running the command `DISABLE_GENERATED_SESSION_NAMING=true  ./target/debug/goose run  -t "say hello twice"` this is what the database looks like when true (top line) and default (false) second line in the session database.
image

### DIFF
--- a/crates/goose/src/session/session_manager.rs
+++ b/crates/goose/src/session/session_manager.rs
@@ -238,7 +238,15 @@ impl SessionManager {
             .count();
 
         if user_message_count <= MSG_COUNT_FOR_SESSION_NAME_GENERATION {
-            let description = provider.generate_session_name(&conversation).await?;
+            let description = if std::env::var("DISABLE_GENERATED_SESSION_NAMING")
+                .unwrap_or_default()
+                .to_lowercase()
+                == "true"
+            {
+                "no session name".to_string()
+            } else {
+                provider.generate_session_name(&conversation).await?
+            };
             Self::update_session(id)
                 .description(description)
                 .apply()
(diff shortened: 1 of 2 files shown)


jamadeo:
+1 to addressing this, but we should make it fast all the time. What about making session naming a fire-and-forget operation where we update the description separately once the provider returns something? cc @DOsinga

jamadeo:
another option that might work well for your use case is the ability to set your own session name, so you can put something useful there while also avoiding the slow generated part

angelahning:
+1 to making session naming fire-and-forget. I am too lazy to set my own session names.

DOsinga:
> +1 to addressing this, but we should make it fast all the time. What about making session naming a fire-and-forget operation where we update the description separately once the provider returns something? cc @DOsinga

it is that. where are we seeing this slow down @jsibbison-square ?

we run this like:


            tokio::spawn(async move {
                if let Err(e) =
                    SessionManager::maybe_update_description(&session_id, provider).await
                {
                    warn!("Failed to generate session description: {}", e);
                }
            });

jsibbison-square:
Oh you are saying this is happening out of band. In that case this is only a minimal cost thing (its probably a couple of cents per 1000 runs) and I'm fine to leave this out if we aren't keen on this solution.

Was more noticable when gemini 1-5 was removed and it defaulted to the main active model which is more expensive.

DOsinga:
> Oh you are saying this is happening out of band. In that case this is only a minimal cost thing (its probably a couple of cents per 1000 runs) and I'm fine to leave this out if we aren't keen on this solution.
> 
> Was more noticable when gemini 1-5 was removed and it defaulted to the main active model which is more expensive.

are you concerned about cost here or about the latency? it is supposed to use the cheap model and yeah when 1-5 went away it went back to the expensive one.

if we wanted to support this, I think the way would be something like Jack said; make it possible to set the description but in a way that the agent then doesn't override it. which would be a good to have feature either way, you can change the description in the desktop, but if you haven't hit 3 user messages it won't stick

jsibbison-square:
Latency was the primary concern but you showed it is out of band so that's all good. Cost is a lower priority, just some use cases just don't need it. Happy for this to be closed and we can look at jacks way at a later stage.

DOsinga:
alright, thanks

PR #4911, opened 2025-10-01
Add type field to empty schemas for anthropic

jamadeo:
According to their own model (Claude) `{"type":"object"}` is equivalent to `{}`. So, either Claude's wrong, or Anthropic is wrong.

¯\_(ツ)_/¯


michaelneale (approved):
this worked for me. Tested with anthropic + databricks (passthrough anthropic)

PR #4913, opened 2025-10-01
fix: session timestamps

jamadeo:
- use datetimes not strings
- parse session file names as local times (they were generated with `Local::now().format("%Y%m%d_%H%M%S").to_string()`) not UTC


crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -317,8 +317,8 @@ mod tests {
             id: "test_session".to_string(),
             working_dir: PathBuf::from(working_dir),
             description: "Test session".to_string(),
-            created_at: "2024-01-01T00:00:00Z".to_string(),
-            updated_at: "2024-01-01T00:00:00Z".to_string(),
+            created_at: Default::default(),
+            updated_at: Default::default(),
```
DOsinga, line 321: this could cause flakiness in tests. probably they would catch something then though

ui/desktop/openapi.json
```diff
@@ -3590,7 +3590,8 @@
             "nullable": true
           },
           "created_at": {
-            "type": "string"
+            "type": "string",
+            "format": "date-time"
```
DOsinga, line 3594: does this not require client side changes

crates/goose/src/context_mgmt/auto_compact.rs
jamadeo: would it? DateTime::default() returns unix epoch 0, not now()

ui/desktop/openapi.json
jamadeo: I think this is just an openapi annotation. If it changed the generated code we'd see that in this diff, but it seems that it does not: https://github.com/block/goose/blob/12f7ef5900be42a48445e757bc2a444a193f6b86/ui/desktop/src/api/types.gen.ts#L712-L713

DOsinga: fair enough, you'd think it would turn into js Dates though those have their own issues I remember

jamadeo: turns out it's a plugin: https://heyapi.dev/openapi-ts/plugins/transformers#dates

PR #4914, opened 2025-10-01
live testing

michaelneale:
trying a separate workflow for live testing


DOsinga:
where do we want to go with this? attach it to the nightly? we do have a bunch of integration like tests in our repostory too, we could use those maybe? like in goose/examples

.github/workflows/pr-smoke-test.yml
```diff
@@ -0,0 +1,126 @@
...
+      
+      - name: Check for code changes
+        uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # pin@v3
+        id: filter
+        with:
+          filters: |
+            code:
+              - '!documentation/**'
+
+  build-binary:
```
jamadeo, line 25: at some point we should refactor these so we only have one "build" step that is called from different workflows, rather than repeating. our desktop bundle step should, for example, take advantage of all of the compiling done by the CLI binary builders

michaelneale:
@DOsinga I think we need to find it a way to be on every PR, it has happened again.

michaelneale:
@DOsinga @jamadeo FYI when I updated this yesterday, it correctly picked up the openai breakage (due to schema) - now updating it to main to check if it is healthy, which is a good sign!

.github/workflows/pr-smoke-test.yml
```diff
@@ -0,0 +1,126 @@
...
+      - name: Make Binary Executable
+        run: chmod +x target/release/goose
+
+      - name: Run Smoke Test
+        env:
+          GOOSE_PROVIDER: ${{ matrix.provider }}
+          GOOSE_MODEL: ${{ matrix.provider == 'anthropic' && 'claude-sonnet-4-5-20250929' || 'gpt-5' }}
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          HOME: /tmp/goose-home
```
DOsinga, line 97: this is great, can we add more model provider combinations? databricks and google I think we break more in practice

michaelneale:
closing this as I prefer other approach with script

PR #4919, opened 2025-10-01
chore: put test in the name

jamadeo:
This job runs lint and test, so it should say so in the title


jamadeo:
I wonder if changing the name is what's causing the check to be stuck on pending. This might only affect this PR but I worry about it breaking something on merge too. Maybe the branch protection rules somehow keep a reference by name?

Screenshot 2025-10-02 at 9 27 05 AM
Screenshot 2025-10-02 at 9 25 16 AM

PR #4922, opened 2025-10-01
feat: lazy infinite scroller for session history view

jamadeo:
Session history at the API level is quite fast now thanks to the SQLite switch. But the client still renders thousands of session history tiles somewhat slowly. This speeds things up by adding a lazy infinite scroll. If we want to speed even more, we could actually paginate at the API level, but we'd have to implement some other solution for search, or fetch them all when you cmd+F.


jamadeo:
Yeah, something virtualized is also a good option. I imagine though with session history you either pick something very close to the top, or you search, and scrolling a lot is probably uncommon.

zanesq:
True! The only issue is it can take up a lot of memory with long session history lists since its in the dom without virtualization

Issue #4513, opened 2025-09-04
goose gets stuck even on non-blocking requests


blackgirlbytes:
**Describe the bug**

goose seems to get stuck on simple tasks such as reading/accessing a file. Typically, this happens when goose runs a long-running task like npm run dev..then other operations get blocked, but it seems to be happening in other cases as well. 

Community member: @/fabian_andgo had the following session where goose got stuck, but there was no long running task and the context window was fairly low. When asked why it was taking so long goose said, "yeah sorry about that"

```
Let me check the taskAnalyzer.ts file to see if there might be an issue with the implementation:


─── text_editor | developer ──────────────────────────
path: ~/D/a/0/t/s/services/taskAnalyzer.ts
command: view

Now let me check if there are any environment variables set up properly for the API key:


─── text_editor | developer ──────────────────────────
path: ~/Documents/ai_test/03_cli_task_splitter/task-splitter/.env
command: view

Let me check what's in the .gooseignore file:


─── shell | developer ──────────────────────────
command: cat .gooseignore
Let me check what's in the .env file:


Context: ●●○○○○○○○○ 19% (50833/262144 tokens)
( O)>
```

**To Reproduce**

**Expected behavior**
goose should continue to run on simple tasks like reading a file.

taniacryptid:
Replying to this issue to say I just ran into this issue as well when asking goose to look at a local file within its directory, and started happening after the latest release! Also linking back to Discord thread with other community members reporting it in case: https://discord.com/channels/1287729918100246654/1413076002690699264/1413175010511683786

ewann:
> Replying to this issue to say I just ran into this issue as well when asking goose to look at a local file within its directory, and started happening after the latest release! Also linking back to Discord thread with other community members reporting it in case: https://discord.com/channels/1287729918100246654/1413076002690699264/1413175010511683786

I wonder if it's worth capturing provider & model info too, particularly is at risk of getting a fair amount of "me too". For example: I appear to experience this a lot with qwen3* models, but qwen2.5-coder not so much / not at all.

alexhancock:
I have to say I haven't seen it at all recently with 1.7.0, but appears some people are.

I just merged this change which should give us some additional logging when this happens: https://github.com/block/goose/pull/3888

We will include this in the next release, and then when it happens we can find the associated log and see if an LLM or tool interaction hung

jmdaly:
[...]

PR #4924, opened 2025-10-01
Make token counter saver

DOsinga:
fixes things up a bit. most crucially dies less

### DIFF
--- a/crates/goose/src/token_counter.rs
+++ b/crates/goose/src/token_counter.rs
@@ -8,25 +8,20 @@ use tokio::sync::OnceCell;
 
 use crate::conversation::message::Message;
 
-// Global tokenizer instance to avoid repeated initialization
 static TOKENIZER: OnceCell<Arc<CoreBPE>> = OnceCell::const_new();
 
-// Cache size limits to prevent unbounded growth
 const MAX_TOKEN_CACHE_SIZE: usize = 10_000;
 
-/// Async token counter with caching capabilities
 pub struct AsyncTokenCounter {
     tokenizer: Arc<CoreBPE>,
-    token_cache: Arc<DashMap<u64, usize>>, // content hash -> token count
+    token_cache: Arc<DashMap<u64, usize>>,
 }
 
-/// Legacy synchronous token counter for backward compatibility
 pub struct TokenCounter {
     tokenizer: Arc<CoreBPE>,
 }
 
 impl AsyncTokenCounter {
-    /// Creates a new async token counter with caching
     pub async fn new() -> Result<Self, String> {
         let tokenizer = get_tokenizer().await?;
         Ok(Self {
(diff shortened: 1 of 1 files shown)


crates/goose/src/token_counter.rs
```diff
@@ -86,34 +74,35 @@ impl AsyncTokenCounter {
...
+                            if let Some(enum_values) = value.get("enum").and_then(|v| v.as_array())
+                            {
                                 func_token_count =
-                                    func_token_count.saturating_add_signed(enum_init);
+                                    func_token_count.saturating_add_signed(ENUM_INIT);
                                 for item in enum_values {
                                     if let Some(item_str) = item.as_str() {
-                                        func_token_count += enum_item;
+                                        func_token_count =
+                                            func_token_count.saturating_add_signed(ENUM_INIT);
```
jamadeo: is this not supposed to be ENUM_ITEM?

crates/goose/src/token_counter.rs
```diff
@@ -239,27 +215,32 @@ impl TokenCounter {
...
+                            if let Some(enum_values) = value.get("enum").and_then(|v| v.as_array())
+                            {
                                 func_token_count =
-                                    func_token_count.saturating_add_signed(enum_init); // Add tokens if property has enum list
+                                    func_token_count.saturating_add_signed(ENUM_INIT);
                                 for item in enum_values {
                                     if let Some(item_str) = item.as_str() {
-                                        func_token_count += enum_item;
+                                        func_token_count =
+                                            func_token_count.saturating_add_signed(ENUM_INIT);
```
jamadeo: also here.

How is this different from the one in AsyncTokenCounter?

jamadeo (approved):

crates/goose/src/token_counter.rs
```diff
@@ -341,40 +316,33 @@ impl TokenCounter {
...
-/// Fixed encoding for all tokenization - using o200k_base for GPT-4o and o1 models
 async fn get_tokenizer() -> Result<Arc<CoreBPE>, String> {
     let tokenizer = TOKENIZER
         .get_or_init(|| async {
-            match tiktoken_rs::o200k_base() {
-                Ok(bpe) => Arc::new(bpe),
-                Err(e) => panic!("Failed to initialize o200k_base tokenizer: {}", e),
-            }
+            tiktoken_rs::o200k_base().map(Arc::new).unwrap_or_else(|e| {
+                eprintln!("Failed to initialize o200k_base tokenizer: {}", e);
```
jamadeo: this looks like a log message the user wouldn't know what to do with

katzdave:
Where are you seeing this actually get used, local models? The counting should be done generally in provider metadata with this as a fallback.

DOsinga:
> Where are you seeing this actually get used, local models? The counting should be done generally in provider metadata with this as a fallback.

we see crashes here and there. most recently: https://github.com/block/goose/issues/4513

crates/goose/src/token_counter.rs
DOsinga: yeah, I know, but I don't know what else we should do. maybe still crash?

crates/goose/src/token_counter.rs
DOsinga: this code is clearly duplicated

crates/goose/src/token_counter.rs
DOsinga: oops yes

PR #4930, opened 2025-10-01
Fix/mcp documentation cleanup

aegntic:
## Pull Request Description

---

**Email**:

### DIFF
--- a/MCP_CLEANUP_NOTES.md
+++ b/MCP_CLEANUP_NOTES.md
@@ -1,0 +1,75 @@
+# MCP Server Documentation Cleanup - Issue #4121
+
+## Completed Changes
+
+### 1. Brave Search MCP Server (FIXED ✅)
+- **Issue**: Server has been archived and is no longer maintained
+- **Action**: Added deprecation warning and recommended alternatives
+- **Files Modified**: `documentation/docs/mcp/brave-mcp.md`
+- **Impact**: Users will be warned about security issues and guided to better alternatives
+
+### 2. Puppeteer MCP Server Link (FIXED ✅)
+- **Issue**: Broken documentation link returning 404 error
+- **Action**: Fixed URL from `/tree/HEAD/src/puppeteer` to `/tree/main/src/puppeteer`
+- **Files Modified**: `documentation/docs/mcp/puppeteer-mcp.md`
+- **Impact**: Documentation now links to valid repository location
+
+### 3. Figma MCP Server (VERIFIED ✅)
+- **Issue**: Mentioned in #4110 as potentially problematic
+- **Status**: Verified to be correct - uses official Figma Dev Mode server
+- **Action**: No changes needed - documentation is accurate
+
+## MCP Servers Audited (56 total)
+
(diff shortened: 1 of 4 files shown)


download_cli.sh
```diff
@@ -36,6 +36,20 @@ if ! command -v tar >/dev/null 2>&1 && ! command -v unzip >/dev/null 2>&1; then
   exit 1
 fi
 
+# Check for required extraction tools based on detected OS
+if [ "$OS" = "windows" ]; then
```
iandouglas, line 40: where are you setting `$OS` above this point to detect if the user is on windows?

download_cli.sh
```diff
@@ -87,6 +122,16 @@ case "$ARCH" in
     ;;
 esac
 
+# Debug output (safely handle undefined variables)
+echo "WINDIR: ${WINDIR:-<not set>}"
```
iandouglas, line 126: do we have a 'debug' flag that we could encapsulate these within so we only print it if we ask for debug info?

download_cli.sh
```diff
@@ -58,12 +72,33 @@ else
 fi
 
 # --- 3) Detect OS/Architecture ---
-OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+# Better OS detection for Windows environments
```
iandouglas, line 75: I've seen a handful of other scripts in the repo that are trying to do this OS detection, and feel like we could probably standardize on a 'helper method' format where we call that other code to determine the OS so we don't repeat this so often.

MCP_CLEANUP_NOTES.md
```diff
@@ -0,0 +1,75 @@
+# MCP Server Documentation Cleanup - Issue #4121
```
iandouglas, line 1: does this file need to be in the goose repo?

documentation/docs/mcp/brave-mcp.md
```diff
@@ -8,9 +8,17 @@ import TabItem from '@theme/TabItem';
 import YouTubeShortEmbed from '@site/src/components/YouTubeShortEmbed';
 import GooseDesktopInstaller from '@site/src/components/GooseDesktopInstaller';
 
+:::warning Deprecated
+The Brave Search MCP Server has been **archived** and is no longer maintained. The server may have security vulnerabilities and compatibility issues. We recommend using alternative search MCP servers like [Tavily Search](./tavily-mcp.md) or [Browserbase](./browserbase-mcp.md) for better reliability and security.
```
iandouglas, line 12: i don't think I'd recommend Browserbase here, that's more of a browser automation tool, where Tavily is meant to be a search tool.

dianed-square: I think we just need update this tutorial to use https://github.com/brave/brave-search-mcp-server

iandouglas: yeah, good point @dianed-square

iandouglas:
let's close this PR and revamp it to point to the new Brave MCP server

Issue #4540, opened 2025-09-05
Fails to register cursor-agent provider


jmmaloney4:
**Describe the bug**

Fails to register cursor-agent provider:

```
nix shell github:numtide/nix-ai-tools#goose-cli github:numtide/nix-ai-tools#cursor-agent

goose configure

This will update your existing config file
  if you prefer, you can edit it directly at /Users/jack/.config/goose/config.yaml

┌   goose-configure
│
◇  What would you like to configure?
│  Configure Providers
│
◇  Which model provider should we use?
│  Cursor Agent
│
◇  Model fetch complete
│
◇  Enter a model from that provider:
│  gpt-5
│
◇  Request failed: Command failed with exit code: Some(1)
│
└  Failed to configure provider: init chat completion request with tool did not succeed.
```

Note: Please check the common issues on https://block.github.io/goose/docs/troubleshooting before filing a report

**To Reproduce**

**Expected behavior**

**Screenshots**

**Please provide following information:**
 - **OS & Arch:** [e.g. Ubuntu 22.04 x86]
 - **Interface:** [UI/CLI]
 - **Version:** [e.g. v1.0.2]
 - **Extensions enabled:** [e.g. Computer Controller, Figma]
 - **Provider & Model:** [e.g. Google - gemini-1.5-pro]

**Additional context**

cbruyndoncx:
please EDIT fill out the template , not all linux versions are working/supported equally. I am working on provider bugs right now in v1.7.0

jmmaloney4:
@cbruyndoncx Edited. I'm actually on macOS. Let me know if any other information would be helpful.

justyns:
I'm running into a similar issue with similar versions on macos as well.

- `gpt-5` gives the `Failed to configure provider: init chat completion request with tool did not succeed.` error in OP
- `claude-4-sonnet` doesn't give an error, but hangs on `Checking your configuration...` and never seems to go past that

I also see the "Model fetch complete" message before entering a model, but there's no selection/auto-complete for the model names.

castilloadrian:
Any update on how to get this working?

bharmann:
@castilloadrian @justyns @jmmaloney4 

Had the same issue - I did several things and not sure which worked, so here's the list:
Verify permissions are:  `your username:staff` in `~/.config/goose` 

``` code 
cd ~/.config
ls -al
```

and also for all the files inside that goose folder, mainly:  `config.yaml`

``` code 
cd goose
ls -al
```

[...]

PR #4942, opened 2025-10-02
fix: Issue #4540: `goose configure` -> Cursor Agent succeeds

ElectricToy:
## Pull Request Description

As described in [issue #4540](https://github.com/block/goose/issues/4540), on Mac (at least) the Cursor Agent provider could not be configured via `goose configure` because the process failed with error: 

```
◇  Request failed: Command failed with exit code: Some(1)
│
└  Failed to configure provider: init chat completion request with tool did not succeed.
```

This was not (at least in my case) a permissions error, but a result of two difficulties with cursor-agent's model parameter. 

First, for some reason the documented "-m" argument to `cursor-agent` does not work in current versions; the full "--model" form must be used. 

Second, "gpt-5" was not recognized as a valid by `cursor-agent` (at least on my system, though the model is certainly available and listed under certain circumstances). Goose does not yet seem to implement real-time model enumeration for `cursor-agent`, so the quick fix is to use "auto" as the default. In effect, replacing the command arguments "-m gpt-5" with "--model auto" enables configuration to successfully conclude.

```
┌   goose-configure 
│
│  Configure Providers 
│
│  Cursor Agent 
│
◇  Model fetch complete
│
◇  Enter a model from that provider:
│  auto
│
◓  Checking your configuration...                                                                                                                                
└  Configuration saved successfully
```

### DIFF
--- a/crates/goose/src/providers/cursor_agent.rs
+++ b/crates/goose/src/providers/cursor_agent.rs
@@ -15,8 +15,8 @@ use crate::impl_provider_default;
 use crate::model::ModelConfig;
 use rmcp::model::Tool;
 
-pub const CURSOR_AGENT_DEFAULT_MODEL: &str = "gpt-5";
-pub const CURSOR_AGENT_KNOWN_MODELS: &[&str] = &["gpt-5", "opus-4.1", "sonnet-4"];
+pub const CURSOR_AGENT_DEFAULT_MODEL: &str = "auto";
+pub const CURSOR_AGENT_KNOWN_MODELS: &[&str] = &["auto", "gpt-5", "opus-4.1", "sonnet-4"];
 
 pub const CURSOR_AGENT_DOC_URL: &str = "https://docs.cursor.com/en/cli/overview";
 
@@ -267,7 +267,7 @@ impl CursorAgentProvider {
 
         // Only pass model parameter if it's in the known models list
         if CURSOR_AGENT_KNOWN_MODELS.contains(&self.model.model_name.as_str()) {
-            cmd.arg("-m").arg(&self.model.model_name);
+            cmd.arg("--model").arg(&self.model.model_name);
         }
 
         cmd.arg("-p")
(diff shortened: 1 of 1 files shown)


michaelneale:
thanks @ElectricToy do you mind doing the DCO thing so commits are ok?

ElectricToy:
@michaelneale no problem, done.

crates/goose/src/providers/cursor_agent.rs
```diff
@@ -15,8 +15,8 @@ use crate::impl_provider_default;
 use crate::model::ModelConfig;
 use rmcp::model::Tool;
 
-pub const CURSOR_AGENT_DEFAULT_MODEL: &str = "gpt-5";
-pub const CURSOR_AGENT_KNOWN_MODELS: &[&str] = &["gpt-5", "opus-4.1", "sonnet-4"];
+pub const CURSOR_AGENT_DEFAULT_MODEL: &str = "auto";
+pub const CURSOR_AGENT_KNOWN_MODELS: &[&str] = &["auto", "gpt-5", "opus-4.1", "sonnet-4", "claude"];
```
loganblevins: curious why we're adding `claude` here. does it work with goose already and was simply missed?

ElectricToy: `claude-agent` (at least here) lists it as an available model, perhaps as a proxy for a self-selecting sonnet version? I can't verify that it works with goose yet, though, because I haven't been able to get `goose` to work with any model yet (hence the attempted bug fix). Because `goose` is not yet querying `cursor-agent` models and won't present this as an option for users, its presence here will do no harm, but I'll amend it out until I can verify that `goose` can talk to `cursor-agent` this way.

ElectricToy: Okay @loganblevins, that's amended. Good catch.

loganblevins: LGTM!  I don't have write access to stamp this however. Feel free to open for review and our maintainers can take a peek I'm sure 😄

michaelneale:
thanks @ElectricToy !

PR #4963, opened 2025-10-02
Add reply to the routes

DOsinga:
More stuff into documented routes


DOsinga:
> nice (I guess need to update to main and re-generate? but seems good, surprised reply wasn't in there!).

i think we're still missing bits of the extension management too

PR #4966, opened 2025-10-02
Fix_converstaion handles mismatching visibility on tool call requests/response

katzdave:


crates/goose/src/conversation/mod.rs
```diff
@@ -532,8 +577,8 @@ mod tests {
 
         assert_eq!(fixed.len(), 5);
         assert_eq!(issues.len(), 2);
-        assert!(issues[0].contains("Removed orphaned tool request"));
-        assert!(issues[1].contains("Merged consecutive assistant messages"));
+        assert!(issues[0].contains("Merged consecutive assistant messages"));
+        assert!(issues[1].contains("Removed orphaned tool request"));
```
DOsinga, line 581: why did this switch?

DOsinga (commented):
can you add a test to this that shows the situation of what we want to cover here?

katzdave:
I don't really like this fix overall. Switching that order is part of why it works, but I can validate this works with a broken session file. 

I think we should go with https://github.com/block/goose/pull/4968, but still testing it more

DOsinga:
ok, since this is obsolete now, closing

PR #4967, opened 2025-10-03
Generate Swift, Python and Bash clients

DOsinga:
generate all the clients


iandouglas:
a goose api??? oooooh

(grabs his popcorn)

PR #4968, opened 2025-10-03
Remove orphaned tool calls before compaction

katzdave:


crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -135,17 +137,28 @@ pub async fn check_compaction_needed(
...
+    let mut messages_to_process = messages.to_vec();
+    
+    // Check if the last assistant message contains a tool request
+    // If so, remove it to prevent orphaned tool responses after compaction
+    if let Some(last_assistant_pos) = messages_to_process.iter().rposition(|m| matches!(m.role, rmcp::model::Role::Assistant)) {
+        if messages_to_process[last_assistant_pos].is_tool_call() {
+            info!("Removing last assistant message with pending tool request before compaction");
+            messages_to_process.remove(last_assistant_pos);
+        }
+    }
```
DOsinga, line 143: can't we just call the conversation fixer here to be safe?

katzdave: Will try that; would be nice to not have to add new logic.

DOsinga: it could possibly help with some other corner cases like if there's an empty message in there during streaming that we recently saw. maybe also change the type here to conversation from Vec<Message> - fixes some of that too

katzdave: Yeah doesn't quite have this new logic so keeping both, but threw it in too.

katzdave:
OK have this pretty extensively tested new. The new logic consistently triggers and haven't seen the orphaned tool call error after maxing out the context window on 10 sessions.

crates/goose/src/context_mgmt/auto_compact.rs
jamadeo: now that you have the conversation fixer I don't think you need this check

katzdave: It doesn't handle the visible/non-visible messages now but added some logic for that.

jamadeo (approved):

PR #4970, opened 2025-10-03
Add cli flag to auto open recipe deeplink

keverly:
## What Changed
Adds a new `--open` flag to the `goose recipe deeplink` command. This flag automatically opens the generated deeplink in the Goose Desktop app.

## Why?
Simplifies the workflow of developing and testing recipes. 

## Demo

https://github.com/user-attachments/assets/8084efb8-8e1f-4bf2-8b3b-0bf90f029194

### DIFF
--- a/crates/goose-cli/Cargo.toml
+++ b/crates/goose-cli/Cargo.toml
@@ -58,6 +58,7 @@ tokio-util = { version = "0.7.15", features = ["compat"] }
 is-terminal = "0.4.16"
 anstream = "0.6.18"
 url = "2.5.7"
+open = "5.3.2"
 
 [target.'cfg(target_os = "windows")'.dependencies]
 winapi = { version = "0.3", features = ["wincred"] }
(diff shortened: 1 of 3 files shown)


alexhancock:
I like it! But how about `goose recipe open` as a new subcommand instead of an arg to the deep link generation. Given the deep link never even shows up in the way it is currently, it feels more like something that should just be an explicit command, and then the deep link subcommand should be for generating deep links (to share etc)

keverly:
> I like it! But how about `goose recipe open` as a new subcommand instead of an arg to the deep link generation. Given the deep link never even shows up in the way it is currently, it feels more like something that should just be an explicit command, and then the deep link subcommand should be for generating deep links (to share etc)

Yeah, that could definitely work. As far as I know, you can only open a recipe in the UI via deeplink. So the `goose recipe open` implementation would still generate a deeplink behind the scenes and open it. It just wouldn't print the link out to the console. Does that sound like the right approach?

PR #4974, opened 2025-10-03
chore: Used common function to list recipes in local machine

lifeizhou-ap:
## Summary
Used common function to list recipes in local machine

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [X] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Manual testing and existing unit tests

### DIFF
--- a/crates/goose-server/src/routes/recipe.rs
+++ b/crates/goose-server/src/routes/recipe.rs
@@ -4,7 +4,6 @@ use std::sync::Arc;
 
 use axum::routing::get;
 use axum::{extract::State, http::StatusCode, routing::post, Json, Router};
-use goose::conversation::{message::Message, Conversation};
 use goose::recipe::local_recipes;
 use goose::recipe::Recipe;
 use goose::recipe_deeplink;


crates/goose/src/recipe/local_recipes.rs
```diff
@@ -0,0 +1,134 @@
...
+    } else {
+        std::env::current_dir().unwrap().join(".goose/recipes")
+    }
+}
+
+fn local_recipe_dirs() -> Vec<PathBuf> {
+    let mut local_dirs = vec![PathBuf::from(".")];
+
+    if let Ok(recipe_path_env) = env::var(GOOSE_RECIPE_PATH_ENV_VAR) {
+        let path_separator = if cfg!(windows) { ';' } else { ':' };
```
DOsinga, line 30: can we get this from config so you can set that up outside of env variables (desktop doesn't use them)

crates/goose/src/recipe/local_recipes.rs
```diff
@@ -0,0 +1,134 @@
...
+        local_dirs.extend(recipe_path_env.split(path_separator).map(PathBuf::from));
+    }
+    local_dirs.push(get_recipe_library_dir(true));
+    local_dirs.push(get_recipe_library_dir(false));
+
+    local_dirs
+}
+
+pub fn retrieve_local_recipe_file(recipe_name: &str) -> Result<RecipeFile> {
```
DOsinga: we should probably unify on naming here and elsewhere. let's call everything load that actually loads & returns? so here, load_local_recipe?

crates/goose/src/recipe/search_local_recipes.rs
```diff
@@ -0,0 +1,134 @@
...
+        .join(":");
+    Err(anyhow!(
+        "ℹ️  Failed to retrieve {}.yaml or {}.json in {}",
+        recipe_name,
+        recipe_name,
+        search_dirs_str
+    ))
+}
+
+pub fn discover_local_recipes() -> Result<Vec<(PathBuf, Recipe)>> {
```
DOsinga: list_local_recipes?

crates/goose/src/recipe/local_recipes.rs
```diff
@@ -0,0 +1,134 @@
...
+        || recipe_name.contains('\\')
+        || recipe_name.starts_with('~')
+        || recipe_name.starts_with('.')
+}
+
+fn is_file_name(recipe_name: &str) -> bool {
+    Path::new(recipe_name).extension().is_some()
+}
+
+fn read_recipe_in_dir(dir: &Path, recipe_name: &str) -> Result<RecipeFile> {
```
DOsinga: load_recipes_from_dir?

crates/goose/src/recipe/recipe_library.rs
```diff
@@ -1,52 +1,9 @@
-use crate::config::APP_STRATEGY;
```
DOsinga, line 1: I'd consider just merging this with search_local_recipes and renaming that last one to local_recipes and just do all the local_recipe work there

crates/goose/src/recipe/local_recipes.rs
lifeizhou-ap: This returns RecipeFile which include file contents, file path, and parent file path.  maybe load_local_recipe_file?

PR #4976, opened 2025-10-03
Fix nix flake double copy

FredeHoey:
## Summary
Setting `src = ./.;` in the nix flake, will cause it to first be copied to the nix store and then copied from the nix store to the source directory of the project. Setting `src = self;`, removes this behaviour and only copies the current directory to the nix store as the base

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [x] Build / Release
- [ ] Other (specify below)

### Testing
Manually tested by running nix build

### Related Issues
None


AnthonyRonning:
Good catch. I tested it locally and it worked.

DOsinga:
can you do the DCO thing?

FredeHoey:
@DOsinga done :)

PR #4977, opened 2025-10-03
Allow empty API Key when registering custom provider

### LINKED ISSUE #4797: Make optional to use API Key when registering a custom provider
While registering a custom provider (to my local ramalama server), I noticed that it is required to add an API Key to the custom provider, whoever, in ramalama (at least), the usage of API Key is optional. I think that goose should allow for empty API Key in case the custom provider does not need/accept it.

Make the API Key field during the custom provider registration optional

An alternative was putting an random string while registering & using the custom provider.

**Additional context**
None.

r0x0d:
## Summary

Some providers like ollama/ramalama does not require an API Key to serve models locally, for that, we can make this option accept empty values.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Related Issues
Closes #4797

### DIFF
--- a/crates/goose-cli/src/commands/configure.rs
+++ b/crates/goose-cli/src/commands/configure.rs
@@ -1940,7 +1940,10 @@ fn add_provider() -> Result<(), Box<dyn Error>> {
         })
         .interact()?;
 
-    let api_key: String = cliclack::password("API key:").mask('▪').interact()?;
+    let api_key: String = cliclack::password("API key:")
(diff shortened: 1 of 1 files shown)


r0x0d:
np! happy to help

PR #4987, opened 2025-10-03
Remove orphaned tool calls before compacting

katzdave:
Remove orphaned tool calls before compacting + run fix conversation


katzdave:
ehh also trying to merge some duplicated logic in perform_compaction + check_and_compact.

katzdave:
replaced with: https://github.com/block/goose/pull/5059

PR #4989, opened 2025-10-03
fix: show PowerShell PATH instructions for Windows users

prempyla:
##  Problem

Goose CLI install script shows **Unix-style PATH instructions** on Windows (e.g., `.bashrc`, `.zshrc`).
This confuses Windows users, who primarily use **PowerShell**.
##  Before

```text
Warning: goose installed, but /home/rob-admin/.local/bin is not in your PATH.
Add it to your PATH by editing ~/.bashrc, ~/.zshrc, or similar...
```
##  After

```text
Warning: goose installed, but C:\Users\<user>\.local\bin is not in your PATH.
To add goose to your PATH in PowerShell:

$profilePath = $PROFILE
if (!(Test-Path $profilePath)) { New-Item -Path $profilePath -ItemType File -Force }
Add-Content -Path $profilePath -Value '$env:PATH = "$env:USERPROFILE\.local\bin;$env:PATH"'
. $PROFILE

Alternatively: goose configure
```
##  Fix

* Detect Windows and show **PowerShell instructions**.
* Keep Bash/Zsh for Unix systems.
* Added alternatives: `goose configure` or rerun script.
##  Benefit

Clearer setup for Windows users → **less confusion, smoother onboarding**.

### DIFF
--- a/download_cli.sh
+++ b/download_cli.sh
@@ -243,6 +243,7 @@ fi
 if [[ ":$PATH:" != *":$GOOSE_BIN_DIR:"* ]]; then
   echo ""
   echo "Warning: goose installed, but $GOOSE_BIN_DIR is not in your PATH."
+  
   if [ "$OS" = "windows" ]; then
     echo "To add goose to your PATH in PowerShell:"
     echo ""
@@ -257,14 +258,37 @@ if [[ ":$PATH:" != *":$GOOSE_BIN_DIR:"* ]]; then
     echo "    goose configure"
     echo "or rerun this install script after updating your PATH."
   else
-    echo "Add it to your PATH by editing ~/.bashrc, ~/.zshrc, or similar:"
-    echo "    export PATH=\"$GOOSE_BIN_DIR:\$PATH\""
-    echo "Then reload your shell (e.g. 'source ~/.bashrc', 'source ~/.zshrc') to apply changes."
+    SHELL_NAME=$(basename "$SHELL")
+    
     echo ""
-    echo "Alternatively, you can run:"
-    echo "    goose configure"
-    echo "or rerun this install script after updating your PATH."
+    echo "The \$GOOSE_BIN_DIR is not in your PATH."
+    echo "What would you like to do?"
+    echo "1) Add it for me"
+    echo "2) I'll add it myself, show instructions"
+    
+    read -p "Enter choice [1/2]: " choice
+    
+    case "$choice" in
+      1)
+        RC_FILE="$HOME/.${SHELL_NAME}rc"
+        echo "Adding \$GOOSE_BIN_DIR to $RC_FILE..."
+        echo "export PATH=\"$GOOSE_BIN_DIR:\$PATH\"" >> "$RC_FILE"
+        echo "Done! Reload your shell or run 'source $RC_FILE' to apply changes."
+        ;;
+      2)
+        echo ""
+        echo "Add it to your PATH by editing ~/.${SHELL_NAME}rc or similar:"
+        echo "    export PATH=\"$GOOSE_BIN_DIR:\$PATH\""
+        echo "Then reload your shell (e.g. 'source ~/.${SHELL_NAME}rc') to apply changes."
+        ;;
+      *)
+        echo "Invalid choice. Please add \$GOOSE_BIN_DIR to your PATH manually."
+        ;;
+    esac
   fi
+  
   echo ""
 fi
+
+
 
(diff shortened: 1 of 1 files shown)


alexhancock (approved):

download_cli.sh
```diff
@@ -237,12 +237,35 @@ else
...
+    echo "    export PATH=\"$GOOSE_BIN_DIR:\$PATH\""
+    echo "Then reload your shell (e.g. 'source ~/.bashrc', 'source ~/.zshrc') to apply changes."
+    echo ""
+    echo "Alternatively, you can run:"
+    echo "    goose configure"
+    echo "or rerun this install script after updating your PATH."
+  fi
   echo ""
 fi
+# ...existing code...
```
iandouglas: can we remove this comment? there's no other 'existing code' at the point.

download_cli.sh
```diff
@@ -237,12 +237,35 @@ else
...
+    echo 'if (!(Test-Path $profilePath)) { New-Item -Path $profilePath -ItemType File -Force }'
+    echo 'Add-Content -Path $profilePath -Value ''$env:PATH = "$env:USERPROFILE\.local\bin;$env:PATH"'''
+    echo "# Reload profile or restart PowerShell"
+    echo '. $PROFILE'
+    echo ""
+    echo "Alternatively, you can run:"
+    echo "    goose configure"
+    echo "or rerun this install script after updating your PATH."
+  else
+    echo "Add it to your PATH by editing ~/.bashrc, ~/.zshrc, or similar:"
```
iandouglas: if they're on windows, we probably don't want to repeat bash/zsh/etc instructions here, that was the point of the `Add-Content` call for powershell, correct?

download_cli.sh
alexhancock: yeah this is a good catch. remove the comment line 241 too

prempyla:
Hi @iandouglas,
I updated the Windows PATH instructions to only show PowerShell guidance, without repeating bash/zsh instructions.

download_cli.sh
```diff
@@ -237,12 +237,34 @@ else
...
+    echo ""
+    echo "# Add to your PowerShell profile"
+    echo '$profilePath = $PROFILE'
+    echo 'if (!(Test-Path $profilePath)) { New-Item -Path $profilePath -ItemType File -Force }'
+    echo 'Add-Content -Path $profilePath -Value ''$env:PATH = "$env:USERPROFILE\.local\bin;$env:PATH"'''
+    echo "# Reload profile or restart PowerShell"
+    echo '. $PROFILE'
+    echo ""
+    echo "Alternatively, you can run:"
+    echo "    goose configure"
```
iandouglas, line 258: One last comment here -- you're presenting this 'alternative' on line 256 and 264 almost like it's something they can do INSTEAD of modifying their path? How does running 'goose configure' help set up their PATH on their behalf if this download script couldn't do it?

I kinda feel like we could get more intelligent here, where we just prompt the user "hey do you want me to add the goose installation folder to your PATH environment" and just go run these commands or a an `echo "export PATH=/goose/bin/etc:$PATH" >> ~/.bashrc` because we could determine from `$SHELL` which shell they're using on a Mac/Linux system.

maybe something like:

The $GOOSE_BIN_DIR is not in your path, what would you like to do:
1. Add it for me
2. I'll add it myself, give me instructions

and then if they're on Linux/Mac we could prompt them regardless of their choice:

We think you're using $SHELL, should we add it to ~/.$(basename $SHELL)rc for you?

prempyla:
Hi @iandouglas,
I updated the PATH instructions to be interactive for Mac/Linux, automatically detecting the user’s shell and offering to add $GOOSE_BIN_DIR to their PATH, while keeping PowerShell instructions for Windows. The old “goose configure” alternative is no longer needed in the prompt

download_cli.sh
```diff
@@ -237,12 +237,58 @@ else
...
+    echo "or rerun this install script after updating your PATH."
+  else
+    SHELL_NAME=$(basename "$SHELL")
```
alexhancock, line 261: The scope expanded a bit on this PR and feels like one we'd want to now test a bit on different envs/shells to make sure there are no hitches.

Either that or we go back to only adding the improvement for windows, which I think was great as it was after iterations with @iandouglas !

prempyla:
Thanks for the feedback!  @iandouglas  @alexhancock  I see two possible paths here:

I can revert this PR back to just the Windows PATH fix (which was already reviewed positively) so it’s small and mergeable quickly.

Or, I can continue with the broader interactive improvement, and test it on macOS (zsh & bash) and Linux via Docker. For Windows, I’d need help testing or I could try adding a CI job to cover it.

Which direction would you prefer for this repo?

PR #5009, opened 2025-10-04
feat(providers): add Mistral AI provider

glinf:
- implement Mistral provider with OpenAI-compatible requests and streaming

- register the provider so it can be constructed via the registry

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:   

**Email**:

### DIFF
--- a/crates/goose/src/providers/factory.rs
+++ b/crates/goose/src/providers/factory.rs
@@ -15,6 +15,7 @@ use super::{
     groq::GroqProvider,
     lead_worker::LeadWorkerProvider,
     litellm::LiteLLMProvider,
+    mistral::MistralProvider,
     ollama::OllamaProvider,
     openai::OpenAiProvider,
     openrouter::OpenRouterProvider,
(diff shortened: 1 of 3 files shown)


DOsinga:
Is this implementation not almost completely identical to openai.rs? can this not be a custom provider instead?

glinf:
Appreciate the close read!

Isn't what you’re seeing the same pattern Goose already use for other OpenAI-compatible backends (eg: Groq, xAI, Azure OpenAI) ?

Each of those files largely shares the request/response glue code but still ships as a dedicated provider so goose can surface vendor defaults, auth, doc links, and model metadata without asking users to hand-author a custom-provider stanza.

IMHO Mistral deserve that first-class slot for the same reasons: we publish our own defaults (mistral-medium-latest with a mistral-small-* fast path), expose the public v1/models index so the registry stays in sync with our releases, and handle our endpoint/telemetry quirks (stream toggles, error mapping, optional custom headers) directly in the provider.

More importantly, Mistral isn’t just “another OpenAI clone”. It's a frontier lab shipping original weights and contributing them to the [open-models ecosystem](https://huggingface.co/mistralai/models). Giving those models a first-party provider advertises that availability, keeps our docs discoverable,and makes it trivial for goose users to use Mistral models.

A generic custom-provider entry would lose all of that context and force every user to recreate it by hand.

DOsinga:
I agree Mistral deserves a first rank listing, I am just trying to figure out how we make sure we don't keep copying the openai code every time and then when we need to modify the original have to modify it in multiple places etc.

can we ship this in a custom provider format, but include it by default for example?

glinf:
> can we ship this in a custom provider format, but include it by default for example?

Got it! I see what you mean (I think). I've revised and tested. Let me know what you think?

DOsinga:
so this is what I am thinking:
https://github.com/block/goose/pull/5084

you should be able to just add add the json to the binary and it will show up in the list like any native provider. how does that sound?

DOsinga:
the declarative provider change is now merged, you should be able to use that

glinf:
sounds good i'll try to have this weekend, sorry for the delay

glinf:
Nice work, works well.

PR #5011, opened 2025-10-04
Initial work on goose-server openapi clients

tlongwell-block:
Adds a Python client library for Goose at `clients/python/`. The client wraps the OpenAPI-generated code to provide a more ergonomic Python API.

## Why

The OpenAPI generator produces functional but verbose code. Python developers expect:
- Simple interfaces for common operations
- Native streaming support for SSE
- Context managers for resource lifecycle
- Pythonic method names and patterns

### Core Client (~1,000 LOC)
- **Simple API**: `client.chat("Hello")` instead of managing request objects
- **SSE Streaming**: Native support via httpx-sse, yields text chunks
- **Session Management**: Context managers handle cleanup automatically
- **Tool Confirmations**: Support for non-autonomous mode with three patterns (auto/manual/conditional)

### Generated Foundation (~30,000 LOC)
- OpenAPI generator creates all models and API methods
- We fix import paths automatically (generator uses incorrect absolute imports)
- Generated code remains accessible for advanced use cases

### Design Decisions

**Two-tier architecture**: Simple client covers ~80% of use cases, generated client available for complex needs.

**SSE in core, not extension**: Streaming is fundamental to Goose interactions.

**Direct HTTP for `/confirm`**: Tool confirmations bypass generated client for simplicity.

**Optional extensions**: Retry logic and session persistence are opt-in to keep core lightweight.

## Build Automation

- `just generate-openapi-clients`: Regenerates Python client from OpenAPI spec
- Pre-commit hook: Auto-regenerates on `openapi.json` changes
- Import paths fixed automatically via sed

## Testing

10/13 integration tests passing. Core functionality (chat, streaming, sessions, confirmations) verified.

## Usage

```python
from goose_client import GooseClient

client = GooseClient(api_key="...")
response = client.chat("What is 2+2?")

# With tool confirmations
for chunk in client.stream_chat_with_confirmations("Create a file", auto_confirm="allow_once"):
    print(chunk, end="")
```


alexhancock:
I'm not sure about supporting something like this directly as an export of the goose codebase. We're moving to https://agentclientprotocol.com/overview/introduction soon for this interface too so maybe that's the time to formalize a way for people to gen/maintain clients as it'll be more forward looking.

Cool to see it in such idiomatic python though

PR #5015, opened 2025-10-05
Upgrade electron for macOS Tahoe compatibility

zanesq:
## Summary
macOS Tahoe introduced some rendering issues with electron, upgrading to fix it. Tested app in dev and bundle locally.
 
https://www.theregister.com/2025/10/02/macos_26_electron_slowdown/


DOsinga:
does this remove the border around the icon? /cc @spencrmartin ?

alexhancock:
@DOsinga @spencrmartin can follow up if any issues with icon border, but this seems good to me. Merging

PR #5016, opened 2025-10-05
Make it startable from playwright and also isolate

DOsinga:
To make goose automatic testable, we need to open a debug port for playwright. to make the runs reproducible we also need to specify a folder where logs, data etc end up. to do that we need to centralize the app strategy thing. this does that.

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -151,6 +151,11 @@ async function ensureTempDirExists(): Promise<string> {
 
 if (started) app.quit();
 
+if (process.env.ENABLE_PLAYWRIGHT) {
+  console.log('[Main] Enabling Playwright remote debugging on port 9222');
+  app.commandLine.appendSwitch('remote-debugging-port', '9222');
+}
+
 // In development mode, force registration as the default protocol client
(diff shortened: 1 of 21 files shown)


DOsinga:
> consolidation of the app_strategy stuff

yeah that was long overdue

ui/desktop/src/main.ts
```diff
@@ -151,6 +151,11 @@ async function ensureTempDirExists(): Promise<string> {
 
 if (started) app.quit();
 
+if (process.env.ENABLE_PLAYWRIGHT) {
+  console.log('[Main] Enabling Playwright remote debugging on port 9222');
+  app.commandLine.appendSwitch('remote-debugging-port', '9222');
```
jamadeo, line 156: if it's a command-line flag, why use an environment variable? does `electron-forge start -- --remote-debugging-port=9222` not do the same?

jamadeo (approved):

ui/desktop/src/main.ts
DOsinga: mostly because the easier plumbing through the just -> npm run -> electron pipeline

PR #5017, opened 2025-10-06
Feat: Added custom headers and toggle keyring CLI options

BlairAllan:
## Summary

Add support for custom HTTP headers and a CLI option to toggle keyring storage. Updated docs to explain how to add these headers and provided an example.

### Type of Change

- [x] Feature

### Testing

- Added unit tests for custom headers in custom_providers.rs to verify that the headers are saved
and loaded correctly
- Manually tested the CLI flow for adding custom headers
- Manually tested the keyring toggle functionality via goose configure > Settings > Secret
Storage


taniacryptid:
Let us know when this is ready for review, or if you have questions!

crates/goose/src/config/custom_providers.rs
```diff
@@ -213,3 +214,142 @@ pub fn register_custom_providers(
     }
     Ok(())
 }
+
+#[cfg(test)]
```
DOsinga: delete these tests, I don't think they do much here

crates/goose-cli/src/commands/configure.rs
```diff
@@ -1348,6 +1356,58 @@ pub fn configure_tool_output_dialog() -> Result<(), Box<dyn Error>> {
     Ok(())
 }
 
+pub fn configure_keyring_dialog() -> Result<(), Box<dyn Error>> {
+    let config = Config::global();
+
+    // Check current setting
+    let currently_disabled = std::env::var("GOOSE_DISABLE_KEYRING").is_ok()
+        || config.get_param::<String>("GOOSE_DISABLE_KEYRING").is_ok();
```
DOsinga: get_param already checks the environment variable so you don't need to do that here. also remove the comment.

you might want to check if the environment variable is set independently and warn about it since if it is set, changing the settings is not going to make a difference

crates/goose-server/src/routes/config_management.rs
```diff
@@ -627,6 +628,7 @@ pub async fn create_custom_provider(
         request.api_key,
         request.models,
         request.supports_streaming,
+        request.headers,
```
DOsinga, line 710: this seems fine, but unless we hook it up on the other side, I don't think this bring us much?

DOsinga (approved):

documentation/docs/getting-started/providers.md
```diff
@@ -279,6 +279,72 @@ goose supports using custom OpenAI-compatible endpoints, which is particularly u
 For enterprise deployments, you can pre-configure these values using environment variables or configuration files to ensure consistent governance across your organization.
 :::
 
+### Custom OpenAI-Compatible Providers
+
+Goose supports adding custom OpenAI-compatible providers with additional authentication headers. This is useful for:
```
iandouglas: "goose" should be lowercased in the documentation

documentation/docs/getting-started/providers.md
```diff
@@ -279,6 +279,72 @@ goose supports using custom OpenAI-compatible endpoints, which is particularly u
...
+- Corporate API gateways that require extra authentication headers
+- Multi-tenant systems needing tenant identifiers
+- Services requiring additional API keys or tokens
+
+#### Configuring Custom Providers
+
+<Tabs groupId="interface">
+  <TabItem value="ui" label="Goose Desktop" default>
```
iandouglas: not sure if we're going to lowercase these tab headings, but thinking we should? @dianed-square @emma-squared any thoughts here?

documentation/docs/getting-started/providers.md
```diff
@@ -279,6 +279,72 @@ goose supports using custom OpenAI-compatible endpoints, which is particularly u
...
+       - Enter header value (will be masked for security)
+       - Add more headers as needed
+  </TabItem>
+</Tabs>
+
+#### Manual Configuration
+
+You can also manually configure custom providers by creating a JSON file in:
+- macOS/Linux: `~/.config/goose/custom_providers/your_provider.json`
+- Windows: `%APPDATA%\goose\custom_providers\your_provider.json`
```
iandouglas: I think these files have to start with the word "custom_" don't they? @DOsinga

iandouglas (changes requested):

documentation/docs/getting-started/providers.md
DOsinga: I don't think they have to. We for some reason do it that way, but I think this should work fine

documentation/docs/getting-started/providers.md
iandouglas: I did go through and lowercase these in other docs/tutorials pages, so we should lowercase 'goose' here too

blackgirlbytes:
here's how to fix the [DCO check](https://github.com/block/goose/pull/5017/checks?check_run_id=53651855931)

crates/goose-server/src/routes/config_management.rs
BlairAllan: Yes it would need to be added to the UI for it to work - it was just added because I set it to None to avoid the `create_custom_provider()` missing arg error but thought this would be more useful in the future.

DOsinga:
sorry @BlairAllan for letting this linger - there was a merge conflict I tried to fix, but we changed the interface and I think I made it worse. can you have a look?

DOsinga:
there's a linter error now @BlairAllan

BlairAllan:
> there's a linter error now @BlairAllan 

This should all be fixed now

documentation/docs/getting-started/providers.md
```diff
@@ -285,6 +285,72 @@ goose supports using custom OpenAI-compatible endpoints, which is particularly u
...
+After adding the configuration, set the API key:
+```bash
+export CUSTOM_CORP_API_API_KEY="your-api-key"
+```
+
+Then use the provider:
+```bash
+goose session start --provider custom_corp_api
+```
+
```
dianed-square:
```suggestion

```

dianed-square: Thanks @BlairAllan ! But removing this doc update for now because we don't want the docs to be ahead of the latest release. We'll add it back later.

PR #5019, opened 2025-10-06
fix: use dynamic port allocation for OAuth server

frozendevil:
## Summary
Replace hardcoded port 8020 with OS-assigned dynamic port to prevent conflicts with other applications. The OAuth server now binds to port 0 by default (when no specific port is provided in redirect URL), allowing the OS to assign an available port automatically.

Changes:
- Modified OAuth flow to bind to dynamic port and use actual assigned port
- Updated default redirect URL to not include hardcoded port
- Added helper methods to handle dynamic redirect URLs
- Maintains backward compatibility for explicit port specifications

Resolves port conflicts during OAuth authentication flow.

### Type of Change
- [ x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Unit tests pass. 
### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:   

**Email**:


michaelneale (approved):
nice one - @frozendevil if you can update to main and push to trigger a build, we can get this in. really nice catch! we probably have other places we make this mistake too

frozendevil:
cool, thanks! The fork is up-to-date now, but I don't have permission to kickoff the workflows

michaelneale:
I don't hate all the comments: 

```
// If no port is specified (or port is explicitly 0), let the OS assign one
```

is nice as I always forget you can do that!
(comments are for people that forhget, like me, not those that forever remember like @DOsinga !)

nice one!

frozendevil:
Yeah, my reasoning for leaving the comments was that the port-0 behavior is something that lots of folks haven't encountered (or have forgotten). I've cleaned up the other comments though. 

Re: `get_authorization_url`—the resolved URL is passed to both `get_authorization_url` and `exchange_code_for_token`, so the current spot seemed like the lowest common parent for both of those. Maybe I'm misunderstanding something though?

PR #5020, opened 2025-10-06
fix: backwards compatible parsing recipe file

lifeizhou-ap:
## Summary
Issue: The recipes are not loaded when the description on extension is null. This is related to the change in this PR https://github.com/block/goose/pull/4868

Changes in this PR:
- If the existing recipe file has extension with null or empty, populate it with extension name
- use yaml parser to parse both json and yaml file
- removed out of date comments

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing

Unit test and manual testing


crates/goose/src/recipe/mod.rs
```diff
@@ -21,57 +21,6 @@ fn default_version() -> String {
...
-///     context: None,
-///     activities: None,
-///     author: None,
-///     settings: None,
-///     parameters: None,
-///     response: None,
-///     sub_recipes: None,
-///     retry: None,
-/// };
-///
```
DOsinga, line 74: thanks for deleting this!

crates/goose/src/recipe/mod.rs
```diff
@@ -309,24 +244,34 @@ impl Recipe {
...
-                if let Some(nested_recipe) = yaml_value.get("recipe") {
-                    serde_yaml::from_value(nested_recipe.clone())?
-                } else {
-                    serde_yaml::from_str(content)?
+        // Parse using YAML parser (since JSON is a subset of YAML, this handles both)
+        let mut value: serde_yaml::Value = serde_yaml::from_str(content)
+            .map_err(|e| anyhow::anyhow!("Failed to parse recipe content as YAML/JSON: {}", e))?;
+
+        // Handle nested legacy recipe format
+        if let Some(nested_recipe) = value.get("recipe") {
```
DOsinga, line 252: ugh

lifeizhou-ap: this is legacy format that we have to maintain :(

PR #5021, opened 2025-10-06
fix: check server is ready so that we can decode the recipe deeplink

lifeizhou-ap:
## Summary
**Issue:** When deeplink is triggered in chrome, a new window is created but the recipe is not loaded. 
**Why:**  There is a race condition that the server is not up when triggering decode recipe api call

Screenshot 2025-10-06 at 2 57 48 pm

**Fix:** 
-  Check server is ready when creating a new window (in create_chat). 
-  This will the centralised place to check server status.  Removed the existing check in `get-goosed-host-port` handling as it is redundant now

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Manual testing

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -593,6 +593,13 @@ const createChat = async (
   );
   goosedClients.set(mainWindow.id, goosedClient);
 
+  console.log('[Main] Waiting for backend server to be ready...');
+  const serverReady = await checkServerStatus(goosedClient);
+  if (!serverReady) {
+    throw new Error('Backend server failed to start in time');
+  }
+  console.log('[Main] Backend server is ready');
+
   // Let windowStateKeeper manage the window
   mainWindowState.manage(mainWindow);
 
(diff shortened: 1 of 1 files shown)


ui/desktop/src/main.ts
```diff
@@ -593,6 +593,13 @@ const createChat = async (
   );
   goosedClients.set(mainWindow.id, goosedClient);
 
+  console.log('[Main] Waiting for backend server to be ready...');
+  const serverReady = await checkServerStatus(goosedClient);
+  if (!serverReady) {
+    throw new Error('Backend server failed to start in time');
+  }
+  console.log('[Main] Backend server is ready');
+
```
DOsinga, line 601: can you remove the log statements? either this works and we dont care or it fails and it says so

DOsinga (approved):

PR #5022, opened 2025-10-06
Applied server side call to parse and save recipe

lifeizhou-ap:
## Summary
Client side to call `recipes/save` and `recipes/parse` to import and save recipe

**Client side**
- call `recipes/save` and `recipes/parse` to import and save recipe
- Passed recipeId when start a recipe session so that the recipe file can be viewed or edit in that session
- Removed the validation for importing and saving recipe on the client side.  Validation errors are returned from server side and are shown in the toast on client side
- Removed "title" (Not the recipe title), "name" and "save location" when saving recipe
- Removed the "Save recipe" in Recipe session at the bottom of the menu
- Removed recipeStorage

**Server side**
- Applied recipe validation in `recipes/save` and `recipes/parse` endpoints
- Refactored recipe validation so that both CLI and Desktop can use the same validation. 
- Removed is_global in save_recipe request
- Removed name from list recipe response
- Changed the way to handle backwards compatibility of required `description` on `extension`.  Reason:  current implementation does work well to get the error location of invalid recipe, because we load the recipe as a Value first,  update the extension description to "" and parse the Value to recipe.  so the validation error loses the lines where the error occurs. 

### Type of Change

- [X] Feature
- [ ] Bug fix
- [X] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
unit test and manual testing

### DIFF
--- a/crates/goose-cli/src/commands/recipe.rs
+++ b/crates/goose-cli/src/commands/recipe.rs
@@ -9,20 +9,15 @@ use goose::recipe_deeplink;
 pub fn handle_validate(recipe_name: &str) -> Result<()> {
     // Load and validate the recipe file
     let recipe_file = load_recipe_file(recipe_name)?;
-    match validate_recipe_template_from_file(&recipe_file) {
-        Ok(_) => {
-            println!("{} recipe file is valid", style("✓").green().bold());
-            Ok(())
-        }
-        Err(err) => {
-            println!(
-                "{} recipe file is invalid: {}",
-                style("✗").red().bold(),
-                err
-            );
-            Ok(())
-        }
-    }
+    validate_recipe_template_from_file(&recipe_file).map_err(|err| {
+        anyhow::anyhow!(
+            "{} recipe file is invalid: {}",
+            style("✗").red().bold(),
+            err
+        )
+    })?;
+    println!("{} recipe file is valid", style("✓").green().bold());
+    Ok(())
 }
 
 pub fn handle_deeplink(recipe_name: &str) -> Result<String> {
@@ -187,20 +182,6 @@ prompt: "Test prompt content {{ name }}"
 instructions: "Test instructions"
 "#;
 
-    const RECIPE_WITH_INVALID_JSON_SCHEMA: &str = r#"
-title: "Test Recipe with Invalid JSON Schema"
-description: "A test recipe with invalid JSON schema"
-prompt: "Test prompt content"
-instructions: "Test instructions"
-response:
-  json_schema:
-    type: invalid_type
-    properties:
-      result:
-        type: unknown_type
-    required: "should_be_array_not_string"
-"#;
-
     #[test]
     fn test_handle_deeplink_valid_recipe() {
         let temp_dir = TempDir::new().expect("Failed to create temp directory");
@@ -263,23 +244,6 @@ response:
             create_test_recipe_file(&temp_dir, "test_recipe.yaml", INVALID_RECIPE_CONTENT);
         let result = handle_validate(&recipe_path);
         assert!(result.is_err());
-    }
-
-    #[test]
-    fn test_handle_validation_recipe_with_invalid_json_schema() {
-        let temp_dir = TempDir::new().expect("Failed to create temp directory");
-        let recipe_path = create_test_recipe_file(
-            &temp_dir,
-            "test_recipe.yaml",
-            RECIPE_WITH_INVALID_JSON_SCHEMA,
-        );
-
-        let result = handle_validate(&recipe_path);
-        assert!(result.is_err());
-        assert!(result
-            .unwrap_err()
-            .to_string()
-            .contains("JSON schema validation failed"));
     }
 
     #[test]


crates/goose/src/recipe/validate_recipe.rs
```diff
@@ -0,0 +1,135 @@
...
+        .to_string();
+
+    validate_recipe_template(&recipe_file.content, Some(recipe_dir))
+}
+
+pub fn validate_recipe_template_from_content(recipe_content: &str) -> Result<Recipe> {
+    validate_recipe_template(recipe_content, None)
+}
+
+fn validate_recipe_template(recipe_content: &str, recipe_dir: Option<String>) -> Result<Recipe> {
```
DOsinga: this is is a lot of intermediate functions for validation, consider inlining some

crates/goose-cli/src/commands/recipe.rs
```diff
@@ -17,16 +17,10 @@ use goose::recipe_deeplink;
...
-            Ok(())
-        }
-        Err(err) => {
-            println!("{} {}", style("✗").red().bold(), err);
-            Err(err)
-        }
-    }
+    let recipe_file = load_recipe_file(recipe_name)?;
+    validate_recipe_template_from_file(&recipe_file)?;
+    println!("{} recipe file is valid", style("✓").green().bold());
```
DOsinga: are we missing the pretty errors here?

crates/goose-server/src/routes/recipe.rs
```diff
@@ -333,6 +354,85 @@ async fn save_recipe(
...
 }
 
+fn json_rejection_to_error_response(rejection: JsonRejection) -> ErrorResponse {
+    ErrorResponse {
+        message: format_json_rejection_message(&rejection),
+        status: StatusCode::BAD_REQUEST,
+    }
+}
+
+fn validate_recipe(recipe: &Recipe) -> Result<(), ErrorResponse> {
```
zanesq, line 364: Does this report specific recipe schema validation errors like the front end was doing before this change? Ideally it has more info than just recipe schema validation failed.

crates/goose-server/Cargo.toml
```diff
@@ -39,6 +39,7 @@ utoipa = { version = "4.1", features = ["axum_extras", "chrono"] }
 reqwest = { version = "0.12.9", features = ["json", "rustls-tls", "blocking", "multipart"], default-features = false }
 tokio-util = "0.7.15"
 uuid = { version = "1.11", features = ["v4"] }
+serde_path_to_error = "0.1.20"
```
DOsinga, line 42: oh that's cool

ui/desktop/src/hooks/useRecipeManager.ts
```diff
@@ -286,8 +286,12 @@ export const useRecipeManager = (chat: ChatType, recipe?: Recipe | null) => {
     });
   };
 
+  const recipeId: string | null =
+    (window.appConfig.get('recipeId') as string | null | undefined) ?? null;
+
   return {
     recipe: finalRecipe,
+    recipeId,
```
DOsinga, line 294: hmm, we want to get rid of useRecipeManager

crates/goose-server/src/routes/recipe.rs
lifeizhou-ap: Thanks @zanesq ! At moment it will show the user which attribute is missing such as title or with line number. For save recipe for files, it is clear for the user.  However,  for `import recipe via deeplink` or new recipe, it might not make sense using the line number. We can enhance later.

lifeizhou-ap:
> Looking good overall! Let's make sure all validations are ported over from the frontend. For instance, it's currently allowing recipes with empty instructions

Thanks @amed-xyz ! I did not notice the instruction is optional now :) will add this into validation!

lifeizhou-ap:
> Users won't see the local recipes after upgrading. Should we add migration logic to “import” the locally saved recipes into the global store or still read from local but not allow saving there anymore?

We still load the local recipe from global and local (current directory).  For existing local recipes, since we know its path, when we save it, it still saves to the original path. But for any new recipes, it will save to the global. 

However, it might be helpful to have some migration to move the local to global in the future.  maybe better to let user know before doing it?

zanesq:
> > Users won't see the local recipes after upgrading. Should we add migration logic to “import” the locally saved recipes into the global store or still read from local but not allow saving there anymore?
> 
> We still load the local recipe from global and local (current directory). For existing local recipes, since we know its path, when we save it, it still saves to the original path. But for any new recipes, it will save to the global.
> 
> However, it might be helpful to have some migration to move the local to global in the future. maybe better to let user know before doing it?

Ah ok then I think we are good with this as is, thanks for clarifying!

crates/goose-cli/src/commands/recipe.rs
lifeizhou-ap: I removed because the error was print here and printed again at the top level main.  Now i made a change to return the error with style

PR #5023, opened 2025-10-06
Update system.md with softer subagent language

tlongwell-block:


crates/goose/src/prompts/system.md
```diff
@@ -51,7 +51,7 @@ No extensions are defined. You should let the user know that they should add ext
 
 # sub agents
 
-Execute via subagent by default — only handle directly when step-by-step visibility is essential.
+Execute via subagent when appropriate — handle directly when step-by-step visibility is essential.
```
DOsinga: shouldn't we just replace this with a sentence that says what appropriate is? 

Execute self contained tasks where step-by-step visibility is not important through subagents

DOsinga (approved):

crates/goose/src/prompts/system.md
tlongwell-block: Updated

DOsinga (approved):

PR #5027, opened 2025-10-06
Platform Extension MOIM (Minus One Info Message)

tlongwell-block:
Implements ephemeral context injection at the agent boundary to automatically include timestamp and TODO content in LLM context without modifying conversation history.

## Problem

Previously, accessing TODO content required explicit `todo__read` tool calls, cluttering conversation history and consuming tokens. Timestamps were injected via template variables in system prompts, creating inconsistency across different prompt variants and preventing effective prompt caching.

## Solution

MOIM (Minus One Info Message) injects contextual information directly into the recent context during agent message preparation. This injection contains:

- Current timestamp (always)
- TODO content from session metadata (when available)
- Any other extension-provided context via the `get_moim()` trait method

The injection happens in `agent.rs` at the message preparation stage, ensuring the context is included in the LLM call but never persisted to storage or returned in conversation history.

## Architecture Benefits

- Zero changes to core data structures (Message, Conversation)
- No persistence layer modifications  
- Clean extension boundaries via the `get_moim()` trait method
- No impact on existing sessions or conversation replay
- Extensions can opt-in to MOIM by implementing the trait method

## Implementation Details

- `ExtensionManager::collect_moim()` aggregates MOIM content from all active extensions
- `moim::inject_moim()` injects collected MOIM into context
- TODO extension implements `get_moim()` to provide task content
- Timestamp injection moved from template variables to MOIM system

## Testing

Comprehensive test coverage in `crates/goose/tests/moim_tests.rs` with proper test isolation using `serial_test` for tests that modify environment variables.

## Migration

- Removed `todo__read` tool (breaking change for recipes using it)
- Updated `todo__write` description to reflect automatic availability
- System prompts now receive timestamp via MOIM instead of template variables
- Removed `{{current_date_time}}` template variable from system prompts

## User Impact

- TODO content is now automatically available in context after writing
- No need to explicitly read TODO content, reducing token usage
- Cleaner conversation history without repetitive read operations
- Consistent timestamp availability across all prompt variants


crates/goose/tests/moim_tests.rs
```diff
@@ -0,0 +1,42 @@
+use goose::agents::extension::ExtensionConfig;
```
DOsinga: delete this file. the first test can only break if we change our mind on how we represent the time. the second test tests the same thing as the first

crates/goose-cli/src/commands/acp.rs
```diff
@@ -774,7 +774,7 @@ print(\"hello, world\")
             format_tool_name("platform__manage_extensions"),
             "Platform: Manage Extensions"
         );
-        assert_eq!(format_tool_name("todo__read"), "Todo: Read");
```
DOsinga, line 874: this seems unrelated

crates/goose-cli/src/scenario_tests/scenarios.rs
```diff
@@ -8,10 +8,13 @@ mod tests {
     use crate::scenario_tests::scenario_runner::run_scenario;
     use anyhow::Result;
     use goose::conversation::message::Message;
+    use serial_test::serial;
 
     #[tokio::test]
+    #[serial]
     async fn test_what_is_your_name() -> Result<()> {
-        run_scenario(
+        std::env::set_var("GOOSE_MOIM_ENABLED", "false");
```
DOsinga: as we discussed before, this is not how you want to set environment variables for a test. but also, why do we need to disable this? the point of these is to show that everything still works if you have a new feature, and then disabling the new feature while testing seems not great

crates/goose-cli/src/session/output.rs
```diff
@@ -495,13 +495,10 @@ fn render_dynamic_task_request(call: &CallToolRequestParam, debug: bool) {
 fn render_todo_request(call: &CallToolRequestParam, _debug: bool) {
     print_tool_header(call);
 
-    // For todo tools, always show the full content without redaction
+    // For todo_write, always show the full content without redaction
```
DOsinga: just remove the comment

crates/goose/tests/mcp_integration_test.rs
```diff
@@ -132,11 +133,14 @@ enum TestMode {
     vec![]
 )]
 #[tokio::test]
+#[serial]
 async fn test_replayed_session(
     command: Vec<&str>,
     tool_calls: Vec<CallToolRequestParam>,
     required_envs: Vec<&str>,
 ) {
+    // Disable MOIM for replayed sessions to ensure consistency with recorded data
```
DOsinga: let's not

DOsinga (changes requested):
This looks very promising. I'll have another look later, but meanwhile:

* I think you shouldn't bother with finding an insertion point. you are doing this just before sending it to the provider which means that if you make a mistake we'll get a catastrophic error. I think you want to prefix this to the user message just after:

```if let Some(session_config) = &session {```

block. that way it doesn't go to the storage, but you also don't change the structure of the conversation. you probably don't need a moim.rs then either

* I don't think we want a setting for this unless we also have that in the UI and you don't want that. we should not switch this off in the scenario tests as you do right now, we should rerun those and make sure everything is still working

* can you remove the LLM comments that don't add anything and cut any tests that don't really test anything

File: crates/goose-cli/src/commands/acp.rs (hunk 2 above)
Comment (tlongwell-block) [reply crates/goose-cli/src/commands/acp.rs:874]: We should remove the entire test_format_tool_name_with_extension if this isn't useful

**Comment by tlongwell-block** [comment]
> * I think you shouldn't bother with finding an insertion point. you are doing this just before sending it to the provider which means that if you make a mistake we'll get a catastrophic error. I think you want to prefix this to the user message just after:
> 
> `if let Some(session_config) = &session {`
> 
> block. that way it doesn't go to the storage, but you also don't change the structure of the conversation. you probably don't need a moim.rs then either

Unfortunately, counting on the user message would make this useless for recipes. Long-context independent work is where Goose needs this feature the most

File: crates/goose/tests/mcp_integration_test.rs (hunk 5 above)
Comment (tlongwell-block) [reply crates/goose/tests/mcp_integration_test.rs]: Okay, so I have gotten rid of the config for MOIM entirely as you suggested. It cannot be disabled now

However, for recorded tests, we can't use it. This is because the timestamp it passes in always changes.

I've updated the test provider to simply ignore the moim message for scenario tests

**Comment by michaelneale** [comment]
if this is adjusting the system message each time - doesn't that mean that there is no prompt caching? (as right at the start the content diverges?) - or do I misunderstand?

**Comment by DOsinga** [comment]
> if this is adjusting the system message each time - doesn't that mean that there is no prompt caching? (as right at the start the content diverges?) - or do I misunderstand?

instead of doing that, it is trying to add a synthetic message to the conversation with changeable info towards the bottom of the conversation.

**Comment by tlongwell-block** [comment]
@DOsinga can you take a look again? Think I hit your changes

Though, through benchmarking, I found I had to make the message positioning slightly more complicated again in order to get the best results. But just a little

**Comment by blackgirlbytes** [comment]
hmm, wondering why my team got tagged in this. i dont see any docs changes

**Comment by tlongwell-block** [comment]
> hmm, wondering why my team got tagged in this. i dont see any docs changes

My bad! I didn't merge `main` in correctly earlier and a change in the docs auto tagged your team in

File: crates/goose/src/agents/agent.rs [hunk 6]
```diff
@@ -1102,10 +1102,17 @@ impl Agent {
                     }
                 }
 
+                // Inject MOIM ephemeral context just before provider call
```
DOsinga: please remove these types of comments before offering things for review

it says just before provider call, but that's not relevant in this context at all. the function is called inject_moim, so no need to repeat this

crates/goose/src/agents/extension_manager.rs
```diff
@@ -1107,6 +1107,32 @@ impl ExtensionManager {
             .get(&name.into())
             .map(|ext| ext.get_client())
     }
+
+    /// Collect and aggregate MOIM content from all platform extensions.
```
DOsinga: remove. this is not true. we're also injecting the timestamp

crates/goose/src/agents/extension_manager.rs
```diff
@@ -1107,6 +1107,32 @@ impl ExtensionManager {
...
+    pub async fn collect_moim(&self) -> Option<String> {
+        use chrono::Local;
+
+        let timestamp = Local::now().format("%Y-%m-%d %H:%M:%S").to_string();
+        let mut content = format!("<info-msg>\nDatetime: {}\n", timestamp);
```
DOsinga, line 1169: what's info-msg? I think this would be better if it would return a vec<str>

crates/goose/src/agents/mcp_client.rs
```diff
@@ -72,6 +72,10 @@ pub trait McpClientTrait: Send + Sync {
     async fn subscribe(&self) -> mpsc::Receiver<ServerNotification>;
 
     fn get_info(&self) -> Option<&InitializeResult>;
+
+    async fn get_moim(&self) -> Option<String> {
+        None // Default: most extensions won't provide MOIM
```
DOsinga: comment

crates/goose/src/agents/moim.rs
```diff
@@ -0,0 +1,241 @@
+use crate::agents::extension_manager::ExtensionManager;
+use crate::agents::SessionConfig;
+use crate::conversation::message::{Message, MessageContent};
+use uuid::Uuid;
+
+/// Inject MOIM (Minus One Info Message) into conversation.
+///
+/// MOIM provides ephemeral context that's included in LLM calls
+/// but never persisted to conversation history.
+pub async fn inject_moim(
```
DOsinga, line 11: this should take a conversation and return a conversation

crates/goose/src/agents/moim.rs
```diff
@@ -0,0 +1,241 @@
...
+    messages: &[Message],
+    extension_manager: &ExtensionManager,
+    _session: &Option<SessionConfig>,
```
DOsinga: if not used, remove

crates/goose/src/agents/todo_extension.rs
```diff
@@ -154,61 +137,37 @@ impl TodoClient {
...
-                read_only_hint: Some(false),
-                destructive_hint: Some(true),
-                idempotent_hint: Some(false),
-                open_world_hint: Some(false),
-            })
-        ]
+                "#}
+            .to_string(),
+            object!({
+                "type": "object",
```
DOsinga: I know this is pre-existing, but don't hand roll json schema

crates/goose/src/agents/moim.rs
```diff
@@ -0,0 +1,241 @@
...
+        // Verify the tool pair is still together (positions 2 and 3)
+        assert!(result[2]
+            .content
+            .iter()
+            .any(|c| matches!(c, MessageContent::ToolRequest(_))));
+        assert!(result[3]
+            .content
+            .iter()
+            .any(|c| matches!(c, MessageContent::ToolResponse(_))));
+    }
```
DOsinga, line 136: these tests are very implementation dependent. we want to guard for creating invalid results from inserting the message

crates/goose/src/agents/moim.rs
```diff
@@ -0,0 +1,241 @@
...
+            .any(|c| matches!(c, MessageContent::ToolResponse(_)));
+
+        if prev_has_tool_calls && curr_has_tool_responses {
+            tracing::debug!("MOIM: Adjusting position to avoid breaking tool pair");
+            return last_pos.saturating_sub(1);
+        }
+    }
+
+    last_pos
+}
```
DOsinga, line 41: this code scares me a lot. anything we do here will break goose in a way that is going to be very hard to debug

I also don't think this is correct a single tool class message can contain more than one tool request which then would result in

\<toolcall\>
\<tool response\>
\<tool response\>

and so we'd think it would be ok to do this

I think we need to do a few things here:

* Consider whether we really want to mess with toolcalling. if you look at how compaction broke conversations multiple times, going with something safe like just appending the moim to the user message that was inserted

* Replace the LLM generated tests by careful constructed tests that actually test scenarios that would break

* Probably run the conversation fixer over the result of the moim insertion and if it fails, don't do it. and then report that failure back into the logs and have a look at that

crates/goose/src/agents/moim.rs
tlongwell-block: Done

crates/goose/src/agents/extension_manager.rs
tlongwell-block: Just xml-style delimiter to better allow the agent/llm to discern that the moim is not part of the actual user's message

crates/goose-cli/src/scenario_tests/scenario_runner.rs
```diff
@@ -141,6 +141,32 @@ where
...
+                None => {
+                    goose::config::Config::global()
+                        .delete("GOOSE_MOIM_ENABLED")
+                        .ok();
+                }
+            }
+        }
+    }
+    let _restore_guard = MoimGuard(original_moim);
+
```
DOsinga, line 145: I think we've been over this before. this just leads to flaky tests. you can't make global changes in tests. also, moim needs to be part of the scenario tests

crates/goose-cli/src/scenario_tests/scenarios.rs
```diff
@@ -7,8 +7,10 @@ mod tests {
     use crate::scenario_tests::mock_client::WEATHER_TYPE;
     use crate::scenario_tests::scenario_runner::run_scenario;
     use anyhow::Result;
+    use serial_test::serial;
 
     #[tokio::test]
+    #[serial]
```
DOsinga: this doesn't help

crates/goose/src/agents/agent.rs
```diff
@@ -926,10 +926,15 @@ impl Agent {
                     }
                 }
 
+                let conversation_with_moim = super::moim::inject_moim(
+                    conversation.clone(),
```
DOsinga, line 950: this replaces conversation, so I don't think we need to clone here

crates/goose/src/agents/extension_manager.rs
```diff
@@ -1167,6 +1167,30 @@ impl ExtensionManager {
             .get(&name.into())
             .map(|ext| ext.get_client())
     }
+
+    pub async fn collect_moim(&self) -> Option<String> {
+        use chrono::Local;
+
+        let timestamp = Local::now().format("%Y-%m-%d %H:%M:%S").to_string();
+        let mut content = format!("<info-msg>\nDatetime: {}\n", timestamp);
```
DOsinga, line 1169: a list with join would be more memory effiicient

crates/goose/src/agents/extension_manager.rs
```diff
@@ -1167,6 +1167,30 @@ impl ExtensionManager {
             .get(&name.into())
             .map(|ext| ext.get_client())
     }
+
+    pub async fn collect_moim(&self) -> Option<String> {
+        use chrono::Local;
```
DOsinga: any particular reason to do a local import?

crates/goose/src/agents/moim.rs
```diff
@@ -0,0 +1,64 @@
+use crate::agents::extension_manager::ExtensionManager;
+use crate::conversation::message::Message;
+use crate::conversation::Conversation;
+use rmcp::model::Role;
+
+/// The Minus One Info message provides ephemeral context that's included in LLM calls
+/// but never persisted to conversation history.
```
DOsinga: that comment seems to be not related to the function, you know what I think fo taht

crates/goose/src/agents/moim.rs
DOsinga: I think this point still stands

crates/goose/src/agents/todo_extension.rs
```diff
@@ -286,4 +245,22 @@ impl McpClientTrait for TodoClient {
...
+                    .map(|state| state.content)
+                    .filter(|c| !c.trim().is_empty())
+            }
+            Err(e) => {
+                tracing::debug!("Could not read session for MOIM: {}", e);
+                None
+            }
+        };
+
+        content.map(|c| format!("Current tasks and notes:\n{}\n", c))
```
DOsinga: this seems a tad convoluted, no? something like;

```
    let session_id = self.context.session_id.as_ref()?;
    let metadata = SessionManager::get_session(session_id, false).await.ok()?;
    let state = extension_data::TodoState::from_extension_data(&metadata.extension_data)?;
    
    Some(format!("Current tasks and notes:\n{}\n", state.content))
```

crates/goose/src/agents/moim.rs
```diff
@@ -0,0 +1,64 @@
...
+    if let Some(moim) = extension_manager.collect_moim().await {
+        let mut messages = conversation.messages().clone();
+        let idx = messages
+            .iter()
+            .rposition(|m| m.role == Role::Assistant)
+            .unwrap_or(messages.len());
+        messages.insert(idx, Message::user().with_text(moim));
+        return Conversation::new_unvalidated(messages);
+    }
+    conversation
```
DOsinga, line 40: this still scares me a lot. you're changing the messages at a point where we just crash the agent

this feels like it might break a toolcall (assistant) and toolresult (user) pair. 

I would feel much more comfortable if this didn't return an unvalidated conversation. and if we had tests that actually had tool calling in them and all varieties that typically make things break

DOsinga (approved):

michaelneale:
I think worth pushing on this as I expect it could help performance a lot and also (I expect) gain more with actual prompt caching.

tlongwell-block:
> I think worth pushing on this as I expect it could help performance a lot and also (I expect) gain more with actual prompt caching.

Going to work hard on ensuring safety of this today before merging in and following up on @DOsinga's review. Will try to get it in today

crates/goose/src/agents/moim.rs
tlongwell-block: Okay, calling fix_conversation now and returning the output from there instead of the directly modified unvalidated conversation

tlongwell-block: This ends up creating the behavior we talked about in Slack where the moim gets attached to an existing message. Without me having to add that logic here. So that works out well. The tests reflect this.

crates/goose/src/agents/moim.rs
tlongwell-block: Made the new tests more robust. Now we count on fix_conversation to return a valid conversation for us, too

crates/goose/src/agents/moim.rs
```diff
@@ -0,0 +1,147 @@
...
+        let mut messages = conversation.messages().clone();
+        let idx = messages
+            .iter()
+            .rposition(|m| m.role == Role::Assistant)
+            .unwrap_or(messages.len());
+        messages.insert(idx, Message::user().with_text(moim));
+
+        let (fixed, _issues) = fix_conversation(Conversation::new_unvalidated(messages));
+        return fixed;
+    }
```
DOsinga, line 39: I would suggest here to return the origianl conversation if there are issues. 

and then in the tests, do one that has a bunch of cases, but the cases are all just conversations and verify that the conversation returned are different from what we put in.

tlongwell-block: That's what I originally did, but `fix_conversation` considers that behavior we're counting on now (merging messages of the same type) to be an `issue`

I can check for that specific issue, though, and return `conversation` for any other issues that arise. let me try that out

PR #5028, opened 2025-10-06
fix: missing cmake and protobuf for windows build, deduplicate sh/pws…

vlascik:
…h build commands

## Summary
Fix: install missing cmake and protobuf to docker image for windows release build started from cmd/powershell, to align with other platforms.

Build on Windows is broken otherwise.

Original problem was that the release-windows docker build command `apt-get install -y mingw-w64 protobuf-compiler cmake`... was missing `cmake` and `protobuf-compiler` when started from `powershell`, and not `sh`. Apparently someone added it to `sh`, but forgot to put it to the `powershell` part, so if the build was started from powershell, it didn't work.

I extracted the command to a variable, this way it's in one place and people won't have to remember to make the changes twice.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [x] Build / Release
- [ ] Other (specify below)

### Related Issues
Another stab at PR at https://github.com/block/goose/pull/4816

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -138,14 +138,18 @@ copy-binary-windows:
     } else { \
         Write-Host 'Windows goose-scheduler-executor binary not found.' -ForegroundColor Yellow; \
     }"
-    @if [ -f ./temporal-service/temporal-service.exe ]; then \
-        echo "Copying Windows temporal-service binary..."; \
-        cp -p ./temporal-service/temporal-service.exe ./ui/desktop/src/bin/; \
-    else \
-        echo "Windows temporal-service binary not found. Building it..."; \
-        cd temporal-service && GOOS=windows GOARCH=amd64 go build -o temporal-service.exe main.go && cp temporal-service.exe ../ui/desktop/src/bin/; \
-    fi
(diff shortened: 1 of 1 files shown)


vlascik:
@alexhancock pinging for the new PR for https://github.com/block/goose/pull/5028

DOsinga:
this seems stuck; I think it is because we had a bad commit on main - can you sync to latest?

vlascik:
@DOsinga ok, I had to force push, but It should be rebased to main now.

There was also one more problem with `copy-binary-windows` recipe, half of it was in powershell, half of it was in `sh` for some reason, so obviously it didn't work on Windows, so I've rewritten it to powershell too. I've added it here, I don't think it would warrant a separate PR, but if you'd prefer it that way, I can remove those changes.

DOsinga:
let's get it in!

DOsinga:
w00t

PR #5034, opened 2025-10-06
Add support for headers in extensions deeplinks

amed-xyz:
## Summary
Extensions deeplinks support specifying env vars but not http header vars 

### Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
open a link such as

```
goose://extension?type=streamable_http&url=https%3A%2F%2Fapi.githubcopilot.com%2Fmcp%2F&id=github-mcp&name=GitHub&description=GitHub%20repository%20management%20and%20operations&header=Authorization%3Dyour%20token
```

### DIFF
--- a/documentation/src/pages/extensions/detail.tsx
+++ b/documentation/src/pages/extensions/detail.tsx
@@ -167,6 +167,34 @@ const getDocumentationPath = (serverId: string): string => {
                     </div>
                   )}
 
+                  {server.headers && server.headers.length > 0 && (
+                    <div className="space-y-4">
+                      <h2 className="text-lg font-medium text-textStandard m-0">
+                        Request Headers
+                      </h2>
+                      <div>
+                        {server.headers.map((header) => (
+                          <div
+                            key={header.name}
+                            className="border-b border-borderSubtle py-4 first:pt-0 last:border-0"
+                          >
+                            <div className="text-sm text-textStandard font-medium">
+                              {header.name}
+                            </div>
+                            <div className="text-textSubtle text-sm mt-1">
+                              {header.description}
+                            </div>
+                            {header.required && (
+                              <Badge variant="secondary" className="mt-2">
+                                Required
+                              </Badge>
+                            )}
+                          </div>
+                        ))}
+                      </div>
+                    </div>
+                  )}
+
                   <div className="card-footer">
                     {githubStars !== null && (
                       <a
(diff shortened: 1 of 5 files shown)


documentation/src/pages/extensions/detail.tsx
```diff
@@ -167,6 +167,34 @@ const getDocumentationPath = (serverId: string): string => {
                     </div>
                   )}
 
+                  {server.headers && server.headers.length > 0 && (
```
dianed-square, line 170: Hi @amed-xyz . I added similar support in a different PR but yours is cleaner so I'll rebase yours into my branch! I also hid the Environment Variables section in the details card for streamable http extension types. I can still plan to do that in my branch if you think that's a valid assumption.

dianed-square: To clarify, streamable http extensions would only show a Request Headers section

dianed-square (approved):

documentation/src/types/server.ts
```diff
@@ -14,4 +14,9 @@ export interface MCPServer {
     description: string;
     required: boolean;
   }[];
+  headers?: {
+    name: string;
+    description: string;
+    required: boolean;
+  }[];
```
DOsinga, line 21: what is this?

documentation/static/servers.json
```diff
@@ -288,10 +288,11 @@
...
     "is_builtin": false,
     "endorsed": true,
-    "environmentVariables": [
+    "environmentVariables": [],
+    "headers": [
       {
-        "name": "GITHUB_PERSONAL_ACCESS_TOKEN",
-        "description": "Required environment variable",
+        "name": "Authorization",
+        "description": "GitHub Personal Access Token",
```
DOsinga: should we add something here where it says you need to start that with Bearer

documentation/src/pages/extensions/detail.tsx
amed-xyz: Hi @dianed-square thanks for the review! I'm not sure if there's valid cases for a streamable http extension to have env vars, maybe @blackgirlbytes would more about it!

documentation/src/types/server.ts
amed-xyz: it's the `MCPServer` type used across the extensions section

Issue #4894, opened 2025-09-30
goose:// deep links are ignored on Linux


IA386:
Describe the bug

  Deep linking with goose:// URLs does not function correctly on Linux. When a deep link is opened using xdg-open, the Goose application 
  launches, but it fails to process the URL. The console logs indicate that the application is not aware of the "pending deep link," 
  suggesting that the command-line arguments used by xdg-open are not being handled properly at startup.

  To Reproduce

   1. Ensure the Goose application is not running.
   2. Open a terminal on a Linux (GNOME) desktop.
   3. Execute the following command: xdg-open 
      "goose://extension?cmd=uvx&arg=mcp-server-fetch&id=fetch&name=Fetch&description=Web%20content%20fetching%20and%20processing%20capabilities"
   4. The Goose application starts, but the action specified in the URL (installing the 'Fetch' extension) is not performed.
   5. The console output shows the following, confirming the link was ignored:

        Failed to register global hotkey
        Opening URL:  file:///usr/lib/goose/resources/app.asar/.vite/renderer/main_window/index.html#/?
        React ready event received
        No pending deep link to process
        [main] React ready - window is prepared for deep links

  Expected behavior

  The Goose application should launch and correctly parse the goose:// URL, triggering the corresponding action. In the example provided, the 
  "Fetch" extension should be added to the application.

  Screenshots

  N/A

  Please provide following information:
    - OS & Arch: Debian GNU/Linux 12 (GNOME 43.9) x86_64
    - Interface: UI
    - Version: v1.9.0
    - Extensions enabled: N/A
    - Provider & Model: N/A

  Additional context

  The issue appears to stem from a platform check in ui/desktop/src/main.ts. The code responsible for parsing protocol URLs from command-line 
  arguments is currently restricted to Windows (if (process.platform === 'win32')). This logic should be extended to include Linux (if 
  (process.platform !== 'darwin')) to correctly handle deep links passed via xdg-open.

PR #5041, opened 2025-10-06
Fix linux deeplinks not working

zanesq:
## Summary
Basically any non mac should get the lock, it was restricted to win32 only.

Fixes https://github.com/block/goose/issues/4894


michaelneale (approved):
nice catch

Issue #5039, opened 2025-10-06
Auto-focus prompt input on home screen for immediate typing


tfsq:
**Please explain the motivation behind the feature request.**
Currently, when navigating to the home screen, users must click on the prompt input box before they can start typing. This creates unnecessary friction in the user experience. The problem is that users naturally expect to start typing immediately when they see a prompt interface, similar to how search bars work in browsers, Spotlight on macOS, or command palettes in various applications. This extra click interrupts the flow of thought and slows down the interaction, especially for users who prefer keyboard-driven workflows.

This feature would unlock:

- Faster prompt entry for quick questions or commands
- Improved accessibility for keyboard-first users
- Reduced cognitive load by eliminating an unnecessary interaction step
- Better alignment with standard UX patterns in search and command interfaces

**Describe the solution you'd like**
When the home screen loads or when a user navigates to the home screen, the prompt input box should automatically receive focus. This would allow users to immediately start typing without having to click on the input field first. The cursor should be blinking in the prompt box, ready to accept text input as soon as the home screen is displayed.

**Describe alternatives you've considered**

- Adding a keyboard shortcut (like / or Cmd+K) to quickly focus the prompt input box
- Implementing a global hotkey that brings up a quick prompt window from anywhere in the application
- Keeping the current behavior but adding a visual cue or animation to draw attention to the prompt box

**Additional context**
This is a common pattern in modern applications where the primary action is text input. Examples include Google's homepage, VS Code's command palette, Raycast, Alfred, and many other productivity tools. The feature would make the Goose Desktop application feel more responsive and aligned with user expectations from similar interfaces.

- I have verified this does not duplicate an existing feature request

DOsinga:
I like it

zanesq:
Sounds good, just a note that currently the input is disabled until the chat is fully initialized with extensions activated because technically the chat shouldn't start until the extensions are fully enabled. So you could add the focus to that call when its enabled. This is temporary eventually we will move everything to the backend and the user won't have to wait until chat is ready.

DOsinga:
we could enable to input, but dissable the submit though, couldn't we? I want to start typing before everything is ready!

zanesq:
yeah good point!

PR #5043, opened 2025-10-06
Allow auto focus and typing while chat is initializing

zanesq:
## Summary
Remove disabled input while agent is not ready or extensions are loading
I think we got a little heavy handed with this, user feedback made it clear that they should still be able to type right away 
Submit is still disabled until ready.

fixes https://github.com/block/goose/issues/5039

### DIFF
--- a/ui/desktop/src/components/ChatInput.tsx
+++ b/ui/desktop/src/components/ChatInput.tsx
@@ -1183,13 +1183,7 @@ export default function ChatInput({
     isExtensionsLoading;
 
   const isUserInputDisabled =
-    isAnyImageLoading ||
-    isAnyDroppedFileLoading ||
-    isRecording ||
-    isTranscribing ||
-    isCompacting ||
-    !agentIsReady ||
-    isExtensionsLoading;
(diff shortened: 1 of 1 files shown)


amed-xyz (approved):

ui/desktop/src/components/ChatInput.tsx
```diff
@@ -1183,13 +1183,7 @@ export default function ChatInput({
...
 
   const isUserInputDisabled =
-    isAnyImageLoading ||
-    isAnyDroppedFileLoading ||
-    isRecording ||
-    isTranscribing ||
-    isCompacting ||
-    !agentIsReady ||
-    isExtensionsLoading;
+    isAnyImageLoading || isAnyDroppedFileLoading || isRecording || isTranscribing || isCompacting;
```
DOsinga, line 1186: does this still stop the user from submitting?

zanesq: Yes, above this code is `isSubmitButtonDisabled` which includes the `!agentIsReady || isExtensionsLoading` logic I removed from here so its still disabled based on these conditions.

PR #5053, opened 2025-10-07
Art vandelay: Import & Export

DOsinga:
Make it possible to import and export sessions directly from the desktop

### DIFF
--- a/crates/goose-server/src/routes/session.rs
+++ b/crates/goose-server/src/routes/session.rs
@@ -34,6 +34,12 @@ pub struct UpdateSessionUserRecipeValuesRequest {
     user_recipe_values: HashMap<String, String>,
 }
 
+#[derive(Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct ImportSessionRequest {
+    json: String,
+}
+
 const MAX_DESCRIPTION_LENGTH: usize = 200;
 
 #[utoipa::path(
@@ -228,7 +234,7 @@ async fn export_session(Path(session_id): Path<String>) -> Result<Json<String>, 
 #[utoipa::path(
     post,
     path = "/sessions/import",
-    request_body = String,
+    request_body = ImportSessionRequest,
     responses(
         (status = 200, description = "Session imported successfully", body = Session),
         (status = 401, description = "Unauthorized - Invalid or missing API key"),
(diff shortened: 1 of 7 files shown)


crates/goose-server/src/routes/session.rs
```diff
@@ -199,11 +206,63 @@ async fn delete_session(Path(session_id): Path<String>) -> Result<StatusCode, St
...
+    get,
+    path = "/sessions/{session_id}/export",
+    params(
+        ("session_id" = String, Path, description = "Unique identifier for the session")
+    ),
+    responses(
+        (status = 200, description = "Session exported successfully", body = String),
+        (status = 401, description = "Unauthorized - Invalid or missing API key"),
+        (status = 404, description = "Session not found"),
+        (status = 500, description = "Internal server error")
```
jamadeo, line 219: these just add to the openapi, you don't *need* to enumerate them, but you can if you want to. (on our other endpoints, I don't think they're always telling the truth)

crates/goose/src/session/session_manager.rs
```diff
@@ -895,6 +903,41 @@ impl SessionStorage {
...
+                .extension_data(import.extension_data)
+                .total_tokens(import.total_tokens)
+                .input_tokens(import.input_tokens)
+                .output_tokens(import.output_tokens)
+                .accumulated_total_tokens(import.accumulated_total_tokens)
+                .accumulated_input_tokens(import.accumulated_input_tokens)
+                .accumulated_output_tokens(import.accumulated_output_tokens)
+                .schedule_id(import.schedule_id)
+                .recipe(import.recipe)
+                .user_recipe_values(import.user_recipe_values),
```
jamadeo, line 930: consider constructing the struct itself without the builder so any added new fields are caught at compile time

jamadeo (approved):

crates/goose/src/session/session_manager.rs
DOsinga: I had that at first, but it becomes awkward since we still want to call create_session at this point so then we need to update only the other fields?

ui/desktop/src/components/sessions/SessionListView.tsx
```diff
@@ -427,14 +438,66 @@ const SessionListView: React.FC<SessionListViewProps> = React.memo(
...
 
+    const handleExportSession = useCallback(async (session: Session, e: React.MouseEvent) => {
+      e.stopPropagation();
+
+      const response = await exportSession({
+        path: { session_id: session.id },
+        throwOnError: true,
+      });
+
+      const json = response.data;
```
alexhancock, line 449: the below works but feels a bit like coaxing the electron renderer code into downloading the file. would there be a natural spot to do this rust-side as part of `exportSession`?

DOsinga: I think the cleaner API is for the server to return the contents of the exported file and let the download be handled by the browser; imagine the server runs on a different host - no suddenly that file appears in a different file system. or if we run this code on a browser and you have set up your browser to put the downloads somewhere else. the server wouldn't know about that

PR #5055, opened 2025-10-07
handle non-existent sessions gracefully instead of panicking

r0x0d:
## Summary

This bug got introduced in when merging  dc292883b7 (#4648).  The problem was that the code was trying to load a session id from the database that does not exist.

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:   

**Email**:


DOsinga:
What exactly are you trying to fix here? if the session doesn't exist in the database and you try to resume it, I think we should exit. We can make the error reporting better if that is the case, but creating a new session doesn't seem like the right fix

r0x0d:
> What exactly are you trying to fix here? if the session doesn't exist in the database and you try to resume it, I think we should exit. We can make the error reporting better if that is the case, but creating a new session doesn't seem like the right fix

Hi, @DOsinga. Sorry for the lack of explanation. 

I posted the error I got here: https://github.com/block/goose/pull/4648#discussion_r2411270199. Since #4648 got merged, I couldn't run goose properly. My patch fixed it, but not sure if that's the correct approach. 

Even if I clean up the session's folder, goose can't really start up. I can test in a new environment (without goose being present previously) and post here tomorrow.

DOsinga:
thanks for the context. so are you seeing this every time you start goose? that's pretty bad if so. my concern is a little bit that if we have a sesion id and we can't find that session id, how did we get that session id? if the user provided it, we should error out. it really doesn't exist. but if we somehow think we created one and now have one that doesn't exist, we should try to fix that code path. can you print the session id that it fails with? and see if you can get a stack trace out where it came from?

r0x0d:
> thanks for the context. so are you seeing this every time you start goose? that's pretty bad if so. my concern is a little bit that if we have a sesion id and we can't find that session id, how did we get that session id? if the user provided it, we should error out. it really doesn't exist. but if we somehow think we created one and now have one that doesn't exist, we should try to fix that code path. can you print the session id that it fails with? and see if you can get a stack trace out where it came from?


Yeah, totally. That's a very good idea. Let me put this in a draft and try to debug more.

DOsinga:
I think this can be closed now since we solved it in a different way!

r0x0d:
> I think this can be closed now since we solved it in a different way!

oh, cool! let me close then.

Issue #2331, opened 2025-04-24
Platform tools


gzuuus:
**Please explain the motivation behind the feature request.** 
I've noticed that from the last updates goose now includes some built-in tools that cannot be disabled `platform__search_available_extensions` and `platform__enable_extension` This is something confusing because I cant see them beign enabled, and how to disable them. 

**Describe the solution you'd like**
Enable/disable platform tools like any other tool 

**Additional context** 
My LLM was getting confused with these tools and I haven't found the way to disable them.

- [x] I have verified this does not duplicate an existing feature request

wendytang:
Hi @gzuuus, why would you need to disable them? These tools are built in to assist with extensions, so might you have triggered them by asking about extensions?

gzuuus:
Hey! Thank you for your response. There are several reasons why I would like the ability to enable or disable the platform tools. First, it allows for better control over what I am executing. Additionally, regarding the context window, if these tools are always enabled and cannot be disabled, they are injected every time I send a message to the models. While this may be negligible for larger models, it can be distracting for smaller ones.

Moreover, there is some overlap in names and descriptions with other MCP servers I am using and developing, which is annoying since it confuse the models about which tools to utilize. Finally, not all users require this assistance, so it would be beneficial to have the option to disable these tools

yokonao:
I would like to support this proposal!

The motivation behind being able to toggle Platform tools ON/OFF is rooted in several key benefits:

- **Increased Agent Controllability:** Providing the ability to disable these tools would offer users greater control over the Agent's actions and prevent it from unnecessarily searching for or managing extensions in contexts where such operations are not desired or relevant.
- **Context Compression:** By removing unused tools from the Agent's operational scope, we can potentially reduce the cognitive load and context window size for the LLM, leading to more efficient and focused reasoning.
- **Enhanced Security:** In certain environments or for specific tasks, limiting the Agent's ability to discover and manipulate extensions can reduce potential security risks, especially if the extensions involve sensitive operations like shell or filesystem access.

One of Goose's significant attractions, in my opinion, is its dynamic enable/disable capability for built-in tools. For example, even the fundamental developer extension can be explicitly disabled. I highly value the ability to restrict the use of tools like shell or filesystem operations when they are not required for a particular task. Extending similar customization to the Platform tools would align with this philosophy and further enhance Goose's flexibility.

xek:
[...]

PR #5057, opened 2025-10-07
Rewrite extension management tools

angelahning:
## Summary

This PR refactors extension management tools from being hardcoded in the Agent into a proper platform extension (ExtensionManagerClient), allowing it to be enabled/disabled like any other extension.

### Type of Change

- [x] Feature
- [ ] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing

Manually tested on UI and CLI

### Related Issues
Addesses https://github.com/block/goose/issues/2331

Before:  

After: 
Agent should not have access to the extension management tools when the extension_manager extension is disabled 
Screenshot 2025-10-07 at 2 49 08 PM

Re-enabled the extension, agent can now autonomously discover and enable/disable tools.
Screenshot 2025-10-07 at 2 49 49 PM
Screenshot 2025-10-07 at 2 49 59 PM

### DIFF
--- a/crates/goose/src/agents/extension_manager_extension.rs
+++ b/crates/goose/src/agents/extension_manager_extension.rs
@@ -73,6 +73,7 @@ pub const READ_RESOURCE_TOOL_NAME: &str = "read_resource";
 pub const LIST_RESOURCES_TOOL_NAME: &str = "list_resources";
 pub const SEARCH_AVAILABLE_EXTENSIONS_TOOL_NAME: &str = "search_available_extensions";
 pub const MANAGE_EXTENSIONS_TOOL_NAME: &str = "manage_extensions";
+pub const MANAGE_EXTENSIONS_TOOL_NAME_COMPLETE: &str = "extensionmanager__manage_extensions";
 
 pub struct ExtensionManagerClient {
     info: InitializeResult,
--- a/crates/goose/src/permission/permission_inspector.rs
+++ b/crates/goose/src/permission/permission_inspector.rs
@@ -1,4 +1,4 @@
-use crate::agents::extension_manager_extension::MANAGE_EXTENSIONS_TOOL_NAME;
+use crate::agents::extension_manager_extension::MANAGE_EXTENSIONS_TOOL_NAME_COMPLETE;
 use crate::config::permission::PermissionLevel;
 use crate::config::PermissionManager;
 use crate::conversation::message::{Message, ToolRequest};
@@ -164,7 +164,7 @@ impl ToolInspector for PermissionInspector {
                         InspectionAction::Allow
                     }
                     // 4. Special case for extension management
-                    else if tool_name == MANAGE_EXTENSIONS_TOOL_NAME {
+                    else if tool_name == MANAGE_EXTENSIONS_TOOL_NAME_COMPLETE {
                         InspectionAction::RequireApproval(Some(
                             "Extension management requires approval for security".to_string(),
                         ))
@@ -189,7 +189,7 @@ impl ToolInspector for PermissionInspector {
                     }
                     InspectionAction::Deny => "User permission denies this tool".to_string(),
                     InspectionAction::RequireApproval(_) => {
-                        if tool_name == MANAGE_EXTENSIONS_TOOL_NAME {
+                        if tool_name == MANAGE_EXTENSIONS_TOOL_NAME_COMPLETE {
                             "Extension management requires user approval".to_string()
                         } else {
                             "Tool requires user approval".to_string()
--- a/.github/workflows/bundle-desktop-intel.yml
+++ b/.github/workflows/bundle-desktop-intel.yml
@@ -192,10 +192,10 @@ jobs:
           source_job_url="https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
           unsigned_url="s3://block-goose-artifacts-bucket-production/unsigned/goose-${GITHUB_SHA}-${{ github.run_id }}-intel.zip"
 
-          zip -q -u -r out/goose-darwin-x64/goose_intel_mac.zip entitlements.plist
+          zip -q -u -r out/Goose-darwin-x64/Goose_intel_mac.zip entitlements.plist
 
           # upload unsigned goose to transfer bucket so it can be passed to lambda
-          aws s3 cp --quiet out/goose-darwin-x64/goose_intel_mac.zip "${unsigned_url}"
+          aws s3 cp --quiet out/Goose-darwin-x64/Goose_intel_mac.zip "${unsigned_url}"
 
           # begin signing
           echo "🚀 launching signing process"
@@ -247,7 +247,7 @@ jobs:
 
           # download the signed app from S3
           echo "⬇️ downloading signed app"
-          aws s3 cp --quiet "${signed_url}" out/goose-darwin-x64/goose_intel_mac.zip
+          aws s3 cp --quiet "${signed_url}" out/Goose-darwin-x64/Goose_intel_mac.zip
         working-directory: ui/desktop
 
       - name: Final cleanup before artifact upload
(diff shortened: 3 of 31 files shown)


DOsinga:
closes #5068

clippy-baselines/too_many_lines.txt
```diff
@@ -34,6 +34,7 @@ crates/goose-server/src/routes/reply.rs::reply_handler
 crates/goose/src/agents/agent.rs::create_recipe
 crates/goose/src/agents/agent.rs::reply_internal
 crates/goose/src/agents/extension_manager.rs::add_extension
+crates/goose/src/agents/extension_manager_extension.rs::manage_extensions_impl
```
DOsinga: hmm, this seems a bit suspicious; this is a new file, no? these exceptions are only supposed to be for pre-existing things

crates/goose-cli/src/commands/configure.rs
```diff
@@ -1447,8 +1447,7 @@ pub async fn configure_tool_permissions_dialog() -> Result<(), Box<dyn Error>> {
         .await
         .into_iter()
         .filter(|tool| {
-            tool.name != PLATFORM_LIST_RESOURCES_TOOL_NAME
-                && tool.name != PLATFORM_READ_RESOURCE_TOOL_NAME
+            tool.name != LIST_RESOURCES_TOOL_NAME && tool.name != READ_RESOURCE_TOOL_NAME
```
DOsinga: I think we no longer want this. was always odd

crates/goose/src/agents/extension.rs
```diff
@@ -235,6 +255,7 @@ pub enum ExtensionConfig {
         /// The name used to identify this extension
         name: String,
         description: String,
+        display_name: Option<String>, // needed for the UI
```
DOsinga: why do we need this for the UI?

crates/goose/src/agents/extension_manager_extension.rs
```diff
@@ -0,0 +1,535 @@
...
+};
+use rmcp::object;
+use serde_json::Value;
+use std::sync::Arc;
+use tokio::sync::mpsc;
+use tokio_util::sync::CancellationToken;
+use tracing::error;
+
+pub static EXTENSION_NAME: &str = "extensionmanager";
+pub static DISPLAY_NAME: &str = "Extension Manager";
```
DOsinga: Why do we need both?

crates/goose/src/agents/extension_manager_extension.rs
```diff
@@ -0,0 +1,535 @@
...
+    MissingParameter { param_name: String },
+
+    #[error("Invalid action: {action}. Must be 'enable' or 'disable'")]
+    InvalidAction { action: String },
+
+    #[error("Extension operation failed: {message}")]
+    OperationFailed { message: String },
+}
+
+// Tool name constants
```
DOsinga: that's not very informative

crates/goose/src/agents/extension_manager_extension.rs
```diff
@@ -0,0 +1,535 @@
...
+                }),
+            ).annotate(ToolAnnotations {
+                title: Some("Discover extensions".to_string()),
+                read_only_hint: Some(true),
+                destructive_hint: Some(false),
+                idempotent_hint: Some(false),
+                open_world_hint: Some(false),
+            }),
+            Tool::new(
+                MANAGE_EXTENSIONS_TOOL_NAME.to_string(),
```
DOsinga, line 374: I would suggest to split this into two tools, one to enable and one to disable.

crates/goose/src/agents/extension_manager_extension.rs
```diff
@@ -0,0 +1,535 @@
...
+
+    async fn get_tools(&self) -> Vec<Tool> {
+        let mut tools = vec![
+            Tool::new(
+                SEARCH_AVAILABLE_EXTENSIONS_TOOL_NAME.to_string(),
+                "Searches for additional extensions available to help complete tasks.
+        Use this tool when you're unable to find a specific feature or functionality you need to complete your task, or when standard approaches aren't working.
+        These extensions might provide the exact tools needed to solve your problem.
+        If you find a relevant one, consider using your tools to enable it.".to_string(),
+                object!({
```
DOsinga: this feels really brittle - we have had issues with wrong json schema here. @jamadeo & @alexhancock is there a better way?

crates/goose/src/agents/extension_manager_extension.rs
```diff
@@ -0,0 +1,535 @@
...
+        &self,
+        name: &str,
+        arguments: Option<JsonObject>,
+        _cancellation_token: CancellationToken,
+    ) -> Result<CallToolResult, Error> {
+        let result = match name {
+            SEARCH_AVAILABLE_EXTENSIONS_TOOL_NAME => self
+                .handle_search_available_extensions()
+                .await
+                .map_err(|e| ExtensionManagerToolError::OperationFailed {
```
DOsinga: does returns a different type of error. any reason for that? if not, I think you could just collapse this thing and have result be future that you then await and map error

crates/goose/src/permission/permission_inspector.rs
```diff
@@ -164,7 +164,7 @@ impl ToolInspector for PermissionInspector {
                         InspectionAction::Allow
                     }
                     // 4. Special case for extension management
-                    else if tool_name == PLATFORM_MANAGE_EXTENSIONS_TOOL_NAME {
+                    else if tool_name == MANAGE_EXTENSIONS_TOOL_NAME {
```
DOsinga: ugh

ui/desktop/src/components/settings/extensions/subcomponents/ExtensionList.tsx
```diff
@@ -81,10 +81,12 @@ export default function ExtensionList({
...
 // Helper functions
 export function getFriendlyTitle(extension: FixedExtensionEntry): string {
-  const name = (extension.type === 'builtin' && extension.display_name) || extension.name;
+  const name =
+    ((extension.type === 'builtin' || extension.type === 'platform') && extension.display_name) ||
+    extension.name;
   return name
     .split(/[-_]/) // Split on hyphens and underscores
-    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+    .map((word: string) => word.charAt(0).toUpperCase() + word.slice(1))
```
DOsinga: I don't think you need the type annotation here unless name somehow now has a different type? also I think this goes to show we really don't need display_name at all. I would have deleted it also from builtin if it wasn't for backwards compatibility

crates/goose/src/agents/router_tools.rs
```diff
@@ -1,6 +1,6 @@
-use super::platform_tools::{
-    PLATFORM_LIST_RESOURCES_TOOL_NAME, PLATFORM_MANAGE_EXTENSIONS_TOOL_NAME,
-    PLATFORM_READ_RESOURCE_TOOL_NAME, PLATFORM_SEARCH_AVAILABLE_EXTENSIONS_TOOL_NAME,
+use crate::agents::extension_manager_extension::{
```
DOsinga, line 1: shouldn't this tool also be part of the extension_manager_extension? if you disable the extension_manager_extension_manager_manager, this will just be confusing to the agent

DOsinga:
closes #2331

crates/goose/src/agents/router_tools.rs
angelahning: I will add it.

angelahning: On second thought, the router tools should be separate from the extension_manager_extension. These tools are used for picking tools from an extension for a user query. And they should continue to work even if user disables the dynamic loading of extensions. So we should make that a separate extension.

crates/goose/src/agents/extension_manager_extension.rs
angelahning: somewhere in the code keeps modify the name "Extension Manager" to "extensionmanager", and then failing with a Unknown platform extension error. I spent sometime figuring out why that is, and decided it is faster to just add a display name like the built-ins.

DOsinga: yeah, I don't think that's a good reason. having names and display names is in general a recipe for confusion and mismatches, if we can't track them down ourselves that's even worse.

angelahning: i will track it down

angelahning:
- updated system prompt
- added unit tests
- the router stuff should be its own tool, not migrated in this PR
- fixed existing bug in UI where it was mutating the extenstion name

PR #5063, opened 2025-10-08
Remove duplicate prepare_reply_context call

jamadeo:
We were calling `prepare_reply_context` twice (as long as auto-compaction didn't happen). The cost of this was highly dependent on your extensions, since this function calls list_tools for all of them.

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -963,40 +963,29 @@ impl Agent {
         session: Option<SessionConfig>,
         cancel_token: Option<CancellationToken>,
     ) -> Result<BoxStream<'_, Result<AgentEvent>>> {
-        // Handle auto-compaction before processing
-        let (conversation, compaction_msg, _summarization_usage) = match self
+        let compaction_result = self
             .handle_auto_compaction(unfixed_conversation.messages(), &session)
-            .await?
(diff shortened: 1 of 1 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -963,40 +963,29 @@ impl Agent {
...
 
-                // Continue with normal reply processing using compacted messages
                 let mut reply_stream = self.reply_internal(conversation, session, cancel_token).await?;
                 while let Some(event) = reply_stream.next().await {
                     yield event?;
                 }
-            }));
+            }))
+        } else {
+            self.reply_internal(unfixed_conversation, session, cancel_token)
```
DOsinga, line 986: why do we flush this?

jamadeo: what do you mean? why is there an .await?

DOsinga: ah, I misclicked my comment. both arms now end on reply_internal, but one arm flushes the stream, the other returns

jamadeo: the first branch creates a new stream, with two compaction-related items followed by the contents of the reply_internal stream, and the second just returns the reply_internal stream

PR #5064, opened 2025-10-08
Fix bedrock tool input schema

alexyao2015:
## Summary
The todo extension added in https://github.com/block/goose/pull/4868
causes errors with bedrock because bedrock expects an input schema.

Without this fix, bedrock tool calls fail with

Error: Server error: Failed to call Bedrock:
ValidationException(ValidationException { message: Some("The value at
toolConfig.tools.3.toolSpec.inputSchema.json.type must be one of the
following: object."), meta: ErrorMetadata { code:
Some("ValidationException"), message: Some("The value at
toolConfig.tools.3.toolSpec.inputSchema.json.type must be one of the
following: object."), extras: Some({"aws_request_id":
"9813e4c5-607f-47fc-8a6a-4387ef28acb7"}) } })

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Tested by seeing that an error no longer occurs after this patch.

### DIFF
--- a/crates/goose/src/providers/formats/bedrock.rs
+++ b/crates/goose/src/providers/formats/bedrock.rs
@@ -188,6 +188,14 @@ pub fn to_bedrock_tool_config(tools: &[Tool]) -> Result<bedrock::ToolConfigurati
 }
 
 pub fn to_bedrock_tool(tool: &Tool) -> Result<bedrock::Tool> {
+    let mut input_schema = tool.input_schema.as_ref().clone();
+
+    // If the schema doesn't have a "type" field, add it
+    // This is required by Bedrock
+    if !input_schema.contains_key("type") {
+        input_schema.insert("type".to_string(), Value::String("object".to_string()));
+    }
+
     Ok(bedrock::Tool::ToolSpec(
         bedrock::ToolSpecification::builder()
             .name(tool.name.to_string())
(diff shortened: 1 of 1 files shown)


taniacryptid:
Tagged @DOsinga & @alexhancock to review, thank you @alexyao2015 !

crates/goose/src/providers/formats/bedrock.rs
```diff
@@ -188,6 +188,14 @@ pub fn to_bedrock_tool_config(tools: &[Tool]) -> Result<bedrock::ToolConfigurati
 }
 
 pub fn to_bedrock_tool(tool: &Tool) -> Result<bedrock::Tool> {
+    let mut input_schema = tool.input_schema.as_ref().clone();
+
+    // If the schema doesn't have a "type" field, add it
+    // This is required by Bedrock
+    if !input_schema.contains_key("type") {
+        input_schema.insert("type".to_string(), Value::String("object".to_string()));
+    }
```
DOsinga, line 197: I'm sure this is needed, but it doesn't seem to me like this is in the right place. if the todo tool somehow managed to declare itself without a type in the input schema, we should add a check to extension_manager to make sure all tools have that and make sure that whatever way we setup tools, does the right thing here

alexyao2015: I see the exact same implementation here for the anthropic provider. https://github.com/block/goose/blob/4ffc1429f640c4e07c517e16e7ad652d734d29b7/crates/goose/src/providers/formats/anthropic.rs#L174

DOsinga:
yeah, that's not great. let's get this in but see if we can fix this on a higher level

PR #5065, opened 2025-10-08
Fix mcp large response race condition

alexyao2015:
## Summary
When multiple subagents are running in parallel and they call an mcp
tool that returns a large response, these responses may all be returned
at the exact same second which can result in subagents receiving paths
containing the incorrect information.

We fix this by adding microseconds to the filename so that this is very
unlikely to occur anymore. Maybe a better solution is to check if the
file exists, but I think this is sufficient for now.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Tested that this race condition no longer occurs when the main agent calls subagents to run in parallel.

### DIFF
--- a/crates/goose/src/agents/large_response_handler.rs
+++ b/crates/goose/src/agents/large_response_handler.rs
@@ -64,7 +64,7 @@ fn write_large_text_to_file(content: &str) -> Result<String, std::io::Error> {
     std::fs::create_dir_all(&temp_dir)?;
 
     // Generate a unique filename with timestamp
-    let timestamp = Utc::now().format("%Y%m%d_%H%M%S");
+    let timestamp = Utc::now().format("%Y%m%d_%H%M%S%.6f");
     let filename = format!("mcp_response_{}.txt", timestamp);
     let file_path = temp_dir.join(&filename);
 


alexyao2015:
This is not just for debugging. If an mcp server provides a large response to the agent, it's stored in a file which can be erroneously overwritten with incorrect content in the current implementation.

DOsinga:
well, that is for debugging yes. to be able to see what the large response was. I think this is something we used to have trouble with and can probably go /cc @michaelneale

alexyao2015:
If by debugging, you mean the alternative is using an in memory temporary file, I suppose so? This file is currently read by the agent using shell commands if the developer tools extension is enabled, so it's definitely not used only for debugging as you might think it is.

With the race condition existing as it is today, sub agents get confused because they are fed the wrong data when they receive a response from the mcp server.

DOsinga:
ok, fine. let's give it a real unique name though instead of still a timestmap

michaelneale:
yes, this makes sense, and wow, this is a wild edge case. Yeah some MCPs will spew a lot, so this directs to a file, if they happen to have same name... ouch. It would rarely happen as responses now have to be pretty large (if we make it too small, it slows down common MCP calls). 

Belatedly approve this @alexyao2015 wow, that is cool fix.

PR #5072, opened 2025-10-08
Improve Rust analysis output

tlongwell-block:
Adds reference tracking to the analyze tool for Rust code, showing type associations and usage patterns.

### Changes

- Added reference query support for Rust to capture type usage (method receivers, struct literals, field types, variable annotations, parameters)
- Enhanced formatter to display references section showing methods grouped by type, type instantiations, and type usage in fields/vars/params
- Added helper functions to walk AST and associate methods with their impl types
- Added test coverage for new functionality

### Example Output

```
R: methods[summarize_context(Agent) truncate_context(Agent)]; types[Self]; fields[String Vec]; params[Message]
```

The reference section (R:) now shows which methods belong to which types, making it easier to understand code structure at a glance.

### DIFF
--- a/crates/goose-mcp/src/developer/analyze/parser.rs
+++ b/crates/goose-mcp/src/developer/analyze/parser.rs
@@ -380,11 +380,22 @@ impl ElementExtractor {
                             ast_recursion_limit,
                         );
                         if let Some(method_name) = method_name {
-                            (
-                                ReferenceType::MethodDefinition,
-                                method_name,
-                                Some(text.to_string()),
-                            )
+                            // For Rust, we need to find the impl type since self_parameter text is just "self"
+                            let type_name = if language == "rust" {
+                                Self::find_impl_type_for_self(&node, source)
+                            } else {
+                                Some(text.to_string())
+                            };
+
+                            if let Some(type_name) = type_name {
+                                (
+                                    ReferenceType::MethodDefinition,
+                                    method_name,
+                                    Some(type_name),
+                                )
+                            } else {
+                                continue;
+                            }
                         } else {
                             continue;
                         }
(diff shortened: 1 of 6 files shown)


crates/goose-mcp/src/developer/analyze/parser.rs
```diff
@@ -428,6 +439,25 @@ impl ElementExtractor {
             .and_then(|handler| handler(receiver_node, source, ast_recursion_limit))
     }
 
+    fn find_impl_type_for_self(node: &tree_sitter::Node, source: &str) -> Option<String> {
```
wpfleger96: Can we move this function to `languages/rust.rs` to keep the parser module language-agnostic?

I know this is a special case for Rust bc of its `self` type, so maybe we could add `find_receiver_type_handler: Option<FindReceiverTypeHandler> ` to the `LanguageInfo` struct. Then we can move this logic to `rust.rs` as-is, register the handler to Rust's `LanguageInfo`, and update the calls to use the handler pattern (like we do for `extract_function_name_handler`/`find_method_for_receiver_handler`)

crates/goose-mcp/src/developer/analyze/parser.rs
```diff
@@ -380,11 +380,22 @@ impl ElementExtractor {
                             ast_recursion_limit,
                         );
                         if let Some(method_name) = method_name {
-                            (
-                                ReferenceType::MethodDefinition,
-                                method_name,
-                                Some(text.to_string()),
-                            )
+                            // For Rust, we need to find the impl type since self_parameter text is just "self"
```
wpfleger96: After the change [here](url) we can update this to call the language-specific handler

crates/goose-mcp/src/developer/analyze/parser.rs
tlongwell-block: Done

PR #5075, opened 2025-10-08
Fix: Prevent cross-contamination of cache data across analysis modes

### LINKED ISSUE #5074: Cross-contamination of cache data across analysis modes in Developer extension
**Describe the bug**

When using the `analyze` tool in the Developer extension, the analysis results cache keys on `(path, modified_time)` but does not include the analysis mode, causing an order-dependent bug: when directory analysis runs before focused analysis in the same session, the cached Structure-mode data is incorrectly returned for subsequent Semantic-mode requests, resulting in incomplete results bc of missing call chains and references.

**To Reproduce**
1. Start a fresh Goose session with the Developer extension
2. Prompt the model to analyze the current directory in Structure mode (default mode for directories)
    - This caches all files with Structure mode data (basic counts only, no calls/references)
4. Then prompt the model to run focused analysis on any arbitrary symbol you know exists in those files in the codebase

**Expected behavior**
Focused analysis of the symbol should show incoming/outgoing call chains and detailed references.

Instead, the focused analysis will return missing or incomplete call chains and references for the given symbol.

 - **Interface:** CLI
 - **Version:** unreleased, built off latest main

wpfleger96:
## Summary
This PR fixes a cache contamination bug where the `analyze` tool's (from Developer extension) LRU cache would incorrectly return cached results from one analysis mode (e.g., Structure) when a different mode (e.g., Semantic) was requested. The root cause was that the cache key only included the file path and modification time, but not the analysis mode, allowing different modes to overwrite each other's cached data. 

The fix adds the `AnalysisMode` enum as a third component to the `CacheKey` struct, ensuring that each combination of (path, modified time, mode) creates a separate cache entry. This prevents Structure mode's lightweight directory data from contaminating Semantic mode's detailed AST analysis results, which was problematic when both modes were used sequentially in the same session and would cause the tool to return incomplete results.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
I added a unit test for validating cache separation, and I also tested this manually on a local codebase 

### Related Issues
Fixes #5074

### DIFF
--- a/crates/goose-mcp/src/developer/analyze/cache.rs
+++ b/crates/goose-mcp/src/developer/analyze/cache.rs
@@ -5,7 +5,7 @@ use std::sync::{Arc, Mutex};
 use std::time::SystemTime;
 
 use super::lock_or_recover;
-use crate::developer::analyze::types::AnalysisResult;
+use crate::developer::analyze::types::{AnalysisMode, AnalysisResult};
 
 #[derive(Clone)]
 pub struct AnalysisCache {
@@ -18,6 +18,7 @@ pub struct AnalysisCache {
 struct CacheKey {
     path: PathBuf,
     modified: SystemTime,
+    mode: AnalysisMode,
 }
 
 impl AnalysisCache {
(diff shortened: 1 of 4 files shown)


wpfleger96:
Now when I run the same test that I did in #5074, we see the complete and accurate results:
```
Yes! This time the focused analysis found more complete results. The key improvements are:

**Better Coverage:**
- **11 incoming chains** (vs 10 previously) - now includes the cross-package reference
- **3 files analyzed** (F1, F2, F3) instead of just 2

**Cross-Package Reference Found:**
- `F1:23` - This is the reference in `/Users/wpfleger/Development/oncall-buddy/cmd/main.go` at line 23, which corresponds to the `Run *oncallbuddy.RunCmd` field we found with ripgrep

**Complete File Mapping:**
- F1: `/Users/wpfleger/Development/oncall-buddy/cmd/main.go` (the main command structure)
- F2: `/Users/wpfleger/Development/oncall-buddy/pkg/oncallbuddy/main.go` (where RunCmd is defined)
- F3: `/Users/wpfleger/Development/oncall-buddy/pkg/oncallbuddy/main_test.go` (the test file)

**Definition Location:**
- Shows `F2:87 - RunCmd` which correctly identifies the struct definition at line 87 in the main.go file

So this second run captured the complete picture of the `RunCmd` symbol, including:
1. The struct type definition (line 87)
2. All its methods and their call chains
3. The cross-package reference from cmd/main.go
4. All test references
```

tlongwell-block:
Confirmed the bug and the fix locally

PR #5076, opened 2025-10-08
add openmetadata recipe

PubChimps:
## Summary
Adding an openmetadata example to the recipe catalog!

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [x] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
This has been tested, please let me know if I can help setting up an openmetadata env to test or share a recording of my test

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:   

**Email**: 
nick.acosta@getcollate.io

### DIFF
--- a/documentation/src/pages/recipes/data/recipes/use-openmetadata.yaml
+++ b/documentation/src/pages/recipes/data/recipes/use-openmetadata.yaml
@@ -12,7 +12,6 @@ extensions:
   - http://localhost:8585/mcp
   - --auth-server-url=http://localhost:8585/mcp
   - --client-id=openmetadata
-  - --verbose
   - --clean
   - --header
   - Authorization:${AUTH_HEADER}
@@ -21,10 +20,8 @@ extensions:
   - AUTH_HEADER
   timeout: 300
   description: ''
-  bundled: null
+  bundled: false
 settings:
-  goose_provider: openai
-  goose_model: gpt-4o
   temperature: 0.0
 activities:
 - Generate SQL Given FQN


PubChimps:
Would love some clarification on extensions in recipes... if I already have an openmetadata extension set up in goose this recipe works, I thought the recipe would create an extension if there wasn't one, but it does not, this is expected, right?

documentation/src/pages/recipes/data/recipes/use-openmetadata.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - --clean
+  - --header
+  - Authorization:${AUTH_HEADER}
+  envs: {}
+  env_keys:
+  - AUTH_HEADER
+  timeout: 300
+  description: ''
+  bundled: null
+settings:
```
iandouglas, line 24: I think you could keep the temperature here, but I'd probably remove the openai/gpt-4o requirement as this recipe should run with any tool-enabled LLM when goose installs and uses your MCP server.

documentation/src/pages/recipes/data/recipes/use-openmetadata.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - --verbose
+  - --clean
+  - --header
+  - Authorization:${AUTH_HEADER}
+  envs: {}
+  env_keys:
+  - AUTH_HEADER
+  timeout: 300
+  description: ''
+  bundled: null
```
iandouglas: this should be a boolean of `false` instead of `null`

documentation/src/pages/recipes/data/recipes/use-openmetadata.yaml
```diff
@@ -0,0 +1,50 @@
...
+  args:
+  - -y
+  - mcp-remote
+  - http://localhost:8585/mcp
+  - --auth-server-url=http://localhost:8585/mcp
+  - --client-id=openmetadata
+  - --verbose
+  - --clean
+  - --header
+  - Authorization:${AUTH_HEADER}
```
iandouglas, line 17: quick question, does this need to be in quotes for string safety or to require a space after the colon? ie `Authorization: ${AUTH_HEADER}`

documentation/src/pages/recipes/data/recipes/use-openmetadata.yaml
```diff
@@ -0,0 +1,50 @@
...
+- type: stdio
+  name: openmetadata
+  cmd: npx
+  args:
+  - -y
+  - mcp-remote
+  - http://localhost:8585/mcp
+  - --auth-server-url=http://localhost:8585/mcp
+  - --client-id=openmetadata
+  - --verbose
```
iandouglas: is the verbose flag required?

documentation/src/pages/recipes/data/recipes/use-openmetadata.yaml
PubChimps: no, this is working on my end

iandouglas:
> if I already have an openmetadata extension set up in goose this recipe works, I thought the recipe would create an extension if there wasn't one, but it does not, this is expected, right?

I believe if you have an extension that matches the name/title then goose will use the one you already have; defining the whole structure of the extension within the recipe is more for first-time extension installations.

PR #5079, opened 2025-10-08
Clean room implementation of the chat process

DOsinga:
WIP

### DIFF
--- a/ui/desktop/src/App.tsx
+++ b/ui/desktop/src/App.tsx
@@ -99,10 +99,6 @@ const PairRouteWrapper = ({
       chat={chat}
       setChat={setChat}
       setView={setView}
-      agentState={agentState}
-      loadCurrentChat={loadCurrentChat}
-      setFatalError={setFatalError}
-      setAgentWaitingMessage={setAgentWaitingMessage}
       setIsGoosehintsModalOpen={setIsGoosehintsModalOpen}
       resumeSessionId={resumeSessionId}
       initialMessage={initialMessage}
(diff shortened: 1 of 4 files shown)


ui/desktop/src/hooks/useAgent.ts
```diff
@@ -78,9 +76,7 @@ export function useAgent(): UseAgentReturn {
           sessionId: agentSession.id,
           title: agentSession.recipe?.title || agentSession.description,
           messageHistoryIndex: 0,
-          messages: messages?.map((message: ApiMessage) =>
```
zanesq, line 81: @DOsinga I think we might need to bring this back for now. The front end currently relies on an id and timestamp.

ui/desktop/src/components/context_management/index.ts
```diff
@@ -1,149 +1,29 @@
...
-      `Context management failed: ${error || 'Unknown error'}\n\nStart a new session.`
-    );
+  // Check for errors in the result
+  if (result.error) {
+    throw new Error(`Context management failed: ${result.error}`);
   }
-}
-
-// Function to convert API Message to frontend Message
-export function convertApiMessageToFrontendMessage(apiMessage: ApiMessage): FrontendMessage {
```
zanesq, line 55: removing all these api message frontend conversion functions seems a bit scary

zanesq: nvm verified they aren't needed anymore 👍

ui/desktop/src/hooks/useAgent.ts
zanesq: All good, was seeing react warnings with messages "Encountered two children with the same key, `null`. Keys should be unique so that components maintain their identity across updates." but this exists on main also. gtg

PR #5082, opened 2025-10-08
Unify subrecipe and subagent execution through shared recipe pipeline

tlongwell-block:
Replaces subprocess-based subrecipe execution with in-process transformation, unifying the execution path between subrecipes and subagents while bringing full Recipe feature support to subagents.

### What Changed
- Subrecipes now execute in-process instead of spawning `goose run --recipe`
- Removed TaskType enum - both subrecipes and inline recipes are just recipes
- Unified execution through single `run_complete_subagent_task()` path
- Added ModelOverrideProvider for recipe-specific model configuration

### Backend Changes for Subrecipes
Subrecipes previously spawned a subprocess (`goose run --recipe`) which naturally supported all Recipe features. Now they:
- Load and build recipes directly at task creation time
- Execute through the same in-process pipeline as subagents
- Eliminate 170+ lines of subprocess management code
- Run faster without process overhead

### New Capabilities for Subagents
Subagents (inline recipes) now support the full Recipe specification:
- Recipe context (initial conversation messages)
- Recipe activities (execution logging)
- Recipe retry configuration
- Recipe response schemas with FinalOutputTool validation
- **Recipe settings for provider/model overrides**

### Implementation Details
Both subrecipes and subagents now flow through the same execution path. At task creation, subrecipes are loaded and transformed into Recipe objects, making them indistinguishable from inline recipes during execution. The ModelOverrideProvider enables recipe-specific settings without creating new provider instances.

This moves toward the unified execution model outlined in #4389, where all agent work flows through consistent pipelines regardless of entry point.


DOsinga:
@yingjiehe-xyz can you do a first pass as you touched this last

yingjiehe-xyz (commented):
subagent parts looks good, not fully familiar with sub-recipe

tlongwell-block:
@lifeizhou-ap can you check out the subrecipe changes here?

crates/goose/src/agents/subagent_execution_tool/tasks.rs
```diff
@@ -123,150 +117,3 @@ async fn handle_inline_recipe_task(
...
-}
-
-fn process_output(stdout_output: String) -> Result<Value, String> {
-    let last_line = stdout_output
-        .lines()
-        .filter(|line| !line.trim().is_empty())
-        .next_back()
-        .unwrap_or("");
-
-    if let Some(json_string) = extract_json_from_line(last_line) {
```
lifeizhou-ap, line 267: when we ran sub-recipe in the command line, we check here whether the result is in a json response. This was specified in recipe (response/json_schema payload). If this is a json response, we only return it as tool response to the main agent.  

Now when we switch to the `agent` sub-recipe run, does it maintain the same behaviour?

crates/goose/src/agents/subagent_execution_tool/task_execution_tracker.rs
```diff
@@ -28,25 +27,8 @@ const THROTTLE_INTERVAL_MS: u64 = 250;
 const COMPLETION_NOTIFICATION_DELAY_MS: u64 = 500;
 
 fn format_task_metadata(task_info: &TaskInfo) -> String {
-    if let Some(params) = task_info.task.get_command_parameters() {
```
lifeizhou-ap, line 31: I found displaying parameter value in the console when running subrecipe task is helpful to the users. 

I am sorry that when I suggest pass  the resolved recipe here, I did not think of that we would lose this piece of  information. 

How about adding an additional field called task_metadata in TaskPayload? and add the param values in the task_metadata. Something like

```
task_metadata
{ 
   recipe_parameter_values: ....
} 
```

crates/goose/src/agents/subagent_execution_tool/tasks.rs
lifeizhou-ap: It seems it is handled in run_complete_subagent_task.  Would you please help confirm? Thank you!

crates/goose/src/agents/subagent_execution_tool/tasks.rs
```diff
@@ -41,70 +37,68 @@ pub async fn process_task(
...
     }
 
-    let instruction = recipe
-        .instructions
-        .or(recipe.prompt)
-        .ok_or_else(|| "No instructions or prompt in recipe".to_string())?;
+    // Apply recipe provider settings if specified
+    if let Some(ref settings) = recipe.settings {
+        // Validate: provider requires model
+        if settings.goose_provider.is_some() && settings.goose_model.is_none() {
```
lifeizhou-ap: in recipe, if settings.goose_model is optional, it will use the goose model in the configuration by default.  same as settings.goose_provider. 

We may use the parent recipe provider and model as default instead of the configuration file

crates/goose/src/agents/subagent_execution_tool/tasks.rs
```diff
@@ -41,70 +37,68 @@ pub async fn process_task(
...
+
+        // Apply settings based on what's specified
+        if let (Some(provider), Some(model)) = (&settings.goose_provider, &settings.goose_model) {
+            // Full replacement: new provider and model
+            let model_config =
+                ModelConfig::new_or_fail(model).with_temperature(settings.temperature);
+            task_config.provider = providers::create(provider, model_config)
+                .await
+                .map_err(|e| format!("Failed to create provider '{}': {}", provider, e))?;
+        } else if settings.goose_model.is_some() || settings.temperature.is_some() {
```
lifeizhou-ap: To make this code a bit simpler, In "else if" block: Shall we just create a provider and set to the task config same as the "if" block? before that we can set the default value if the setting does not specify some of the attributes?

lifeizhou-ap:
Hi @tlongwell-block,

Thanks for removing cli code for running subrecipe!   I've added a couple of comments regarding the sub-recipe behaviour.  Also  please test with the subrecipe features in CLI and Desktop to check the behaviours and  the output.  Thank you!

crates/goose/src/agents/subagent_execution_tool/tasks.rs
tlongwell-block: Yes, I we have it with final_output to enforce returning json if requested just like before https://github.com/block/goose/pull/5082/files#diff-3eacfb692b710bd09c1879dfe3f3b936e0331b39038ea22a38fb073f016029bdR33

crates/goose/src/agents/subagent_execution_tool/tasks.rs
tlongwell-block: This if/else branch does a few things

* by default, subagents and subrecipes inherit the parent's provider and model
* we allow override just the model
* we allow overriding the model AND the provider

We do not allow overriding just the provider, as that doesn't make a lot of sense

We not not require specifying either. Inheritance is the default

crates/goose/src/agents/subagent_execution_tool/tasks.rs
tlongwell-block: see https://github.com/block/goose/pull/5082#discussion_r2440818579 as for why this logic is required

crates/goose/src/agents/subagent_execution_tool/task_execution_tracker.rs
tlongwell-block: Added parameter_values

lifeizhou-ap (approved):

crates/goose/src/session/session_manager.rs
```diff
@@ -628,6 +635,9 @@ impl SessionStorage {
     }
 
     async fn create_session(&self, working_dir: PathBuf, description: String) -> Result<Session> {
+        // Mutex to prevent concurrent session creation race conditions
+        let _guard = self.create_session_mutex.lock().await;
```
DOsinga: No. delete this. you keep coming back to this, but this is not the way

crates/goose/src/providers/model_override.rs
```diff
@@ -0,0 +1,210 @@
...
+    ) -> Result<(Message, ProviderUsage), ProviderError> {
+        // Use our override config
+        self.inner
+            .complete_with_model(&self.model_config, system, messages, tools)
+            .await
+    }
+
+    // Streaming: Disable for model override cases
+    fn supports_streaming(&self) -> bool {
+        false // Force non-streaming path for overrides
```
DOsinga: why?

crates/goose/src/providers/model_override.rs
```diff
@@ -0,0 +1,210 @@
...
+    LeadWorkerProviderTrait, MessageStream, Provider, ProviderMetadata, ProviderUsage,
+};
+use super::errors::ProviderError;
+use super::retry::RetryConfig;
+use crate::conversation::message::Message;
+use crate::model::ModelConfig;
+use rmcp::model::Tool;
+
+/// A provider wrapper that overrides the model configuration
+pub struct ModelOverrideProvider {
```
DOsinga: I am not sure I follow here, is all you want to do is create a new provider with a specific model config? can't you just create a new provider instead?

crates/goose/src/agents/subagent_execution_tool/task_execution_tracker.rs
```diff
@@ -28,25 +27,23 @@ const THROTTLE_INTERVAL_MS: u64 = 250;
...
+            // Filter out internal parameters like recipe_dir
+            let mut param_strs: Vec<String> = params
+                .iter()
+                .filter(|(k, _)| k.as_str() != "recipe_dir")
+                .map(|(k, v)| format!("{}={}", k, v))
+                .collect();
+            if !param_strs.is_empty() {
+                param_strs.sort(); // Sort for consistent display
+                return param_strs.join(", ");
+            }
```
DOsinga, line 43: I would move the join  & sort into the previous statement

crates/goose/src/agents/subagent_execution_tool/task_execution_tracker.rs
```diff
@@ -151,7 +148,7 @@ impl TaskExecutionTracker {
     async fn format_line(&self, task_info: Option<&TaskInfo>, line: &str) -> String {
         if let Some(task_info) = task_info {
             let task_name = get_task_name(task_info);
-            let task_type = task_info.task.task_type.clone();
+            let task_type = "recipe";
```
DOsinga: this should probably be a constant RECIPE_TASK_TYPE

crates/goose/src/agents/subagent_execution_tool/task_execution_tracker.rs
```diff
@@ -232,7 +229,7 @@ impl TaskExecutionTracker {
                         }
                     }),
                     current_output: task_info.current_output.clone(),
-                    task_type: task_info.task.task_type.to_string(),
+                    task_type: "recipe".to_string(),
```
DOsinga: use RECIPE_TASK_TYPE here

crates/goose/src/agents/subagent_execution_tool/task_types.rs
```diff
@@ -16,59 +17,21 @@ pub enum ExecutionMode {
...
-            TaskType::InlineRecipe => write!(f, "inline_recipe"),
-            TaskType::SubRecipe => write!(f, "sub_recipe"),
-        }
-    }
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TaskPayload {
+    pub recipe: Recipe,
+    #[serde(default)]
+    pub return_last_only: bool,
+    #[serde(default)]
```
DOsinga: don't use default on a bool

crates/goose/src/agents/subagent_execution_tool/tasks.rs
```diff
@@ -41,70 +37,68 @@ pub async fn process_task(
 
 async fn get_task_result(
```
DOsinga: just replace this with handle_recipe_task

crates/goose/src/agents/subagent_execution_tool/tasks.rs
```diff
@@ -41,70 +37,68 @@ pub async fn process_task(
...
-        .unwrap_or(false);
-
-    // If extensions are explicitly provided in the recipe (even if empty),
-    // override the task_config extensions. Empty array means no extensions.
-    if let Some(exts) = recipe.extensions {
+    // Apply recipe extensions if specified
+    // - None: inherit parent extensions (default)
+    // - Some([]): explicitly no extensions
+    // - Some([...]): use only specified extensions
+    if let Some(ref exts) = recipe.extensions {
```
DOsinga, line 43: this has proven problematic; we recently undid this for subtasks when subtasks were launched and then said they did not have extensions. if we want to support subtasks without tools they should have a special system prompt so we don't tell them use tools and then not give them tools /cc @jamadeo

crates/goose/src/agents/subagent_execution_tool/tasks.rs
```diff
@@ -41,70 +37,68 @@ pub async fn process_task(
...
-        .ok_or_else(|| "No instructions or prompt in recipe".to_string())?;
+    // Apply recipe provider settings if specified
+    if let Some(ref settings) = recipe.settings {
+        // Validate: provider requires model
+        if settings.goose_provider.is_some() && settings.goose_model.is_none() {
+            return Err("Recipe specifies provider but no model".to_string());
+        }
+
+        // Apply settings based on what's specified
+        if let (Some(provider), Some(model)) = (&settings.goose_provider, &settings.goose_model) {
```
DOsinga: if we have a custom model or a custom provider, create a new provider based on either of that - in theory we could switch providers here but not models even?

crates/goose/src/agents/subagent_execution_tool/tasks_manager.rs
```diff
@@ -57,19 +55,26 @@ impl TasksManager {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use serde_json::json;
+    use crate::agents::subagent_execution_tool::task_types::TaskPayload;
+    use crate::recipe::Recipe;
 
     fn create_test_task(id: &str, sub_recipe_name: &str) -> Task {
+        let recipe = Recipe::builder()
```
DOsinga, line 62: now would be a good time to consider whether you want to keep most of the tests here

crates/goose/src/agents/subagent_execution_tool/utils/mod.rs
```diff
@@ -3,10 +3,8 @@ use std::collections::HashMap;
 use crate::agents::subagent_execution_tool::task_types::{TaskInfo, TaskStatus};
 
 pub fn get_task_name(task_info: &TaskInfo) -> &str {
-    task_info
-        .task
-        .get_sub_recipe_name()
-        .unwrap_or(&task_info.task.id)
+    // Return the recipe title (always present as it's a required field)
```
DOsinga: remove this comment!

crates/goose/src/agents/subagent_execution_tool/utils/tests.rs
```diff
@@ -20,76 +20,51 @@ mod test_get_task_name {
...
-        let task_info = create_task_info_with_defaults(malformed_task, TaskStatus::Pending);
-
-        assert_eq!(get_task_name(&task_info), "task_4");
-    }
 }
 
 mod count_by_status {
     use super::*;
 
     fn create_test_task(id: &str, status: TaskStatus) -> TaskInfo {
```
DOsinga, line 51: we have two of these. use the other one

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -2,30 +2,37 @@ use crate::{
...
 };
 use anyhow::{anyhow, Result};
 use futures::StreamExt;
 use rmcp::model::{ErrorCode, ErrorData};
 use std::future::Future;
 use std::pin::Pin;
-use tracing::debug;
+use tracing::{debug, info};
+
+// Type alias for the complex return type
```
DOsinga: I can see that

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -127,17 +137,40 @@ fn get_agent_messages(
...
+        let has_response_schema = recipe.response.is_some();
+        if let Some(response) = recipe.response {
+            agent.add_final_output_tool(response).await;
+        }
+
+        // Build initial conversation with context if provided
+        let mut initial_messages = Vec::new();
+
+        // Add context as initial messages if provided
+        if let Some(context_items) = recipe.context {
```
DOsinga: I don't think we use context. let's delete it

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -127,17 +137,40 @@ fn get_agent_messages(
...
         }
 
-        let mut conversation =
-            Conversation::new_unvalidated(
-                vec![Message::user().with_text(text_instruction.clone())],
-            );
+        // Add FinalOutputTool if response schema is specified
+        let has_response_schema = recipe.response.is_some();
+        if let Some(response) = recipe.response {
+            agent.add_final_output_tool(response).await;
```
DOsinga: hmm, we're now doing this in three different places, can't we just add this agent.reply or there about?

DOsinga (changes requested):

crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs
```diff
@@ -54,27 +56,39 @@ fn extract_task_parameters(params: &Value) -> Vec<Value> {
...
-                task_type: TaskType::SubRecipe,
-                payload,
-            }
-        })
-        .collect();
+) -> Result<Vec<Task>> {
+    let mut tasks = Vec::new();
+
+    for task_command_param in command_params {
+        // Load the recipe file for each task (can't clone RecipeFile)
```
jamadeo: you can just make the RecipeFile `derive(Clone)`

but even better, it looks like build_recipe_from_template and the functions it calls should just be able to borrow the content anyway, they don't need to take ownership

loading the same file N times is definitely not necessary

crates/goose/src/agents/subagent_execution_tool/task_execution_tracker.rs
jamadeo: or change the field in EventTaskInfo to an Enum type

crates/goose/src/agents/subagent_execution_tool/tasks.rs
jamadeo: right, we had models creating tasks using an empty extensions list where they definitely should not have been restricted from using extensions. example "write 'hello world' to hello.txt" spawned with extensions=[].

I'm not sure *why* that was the case, but I don't think the model intended to explicitly disallow extension use.

crates/goose/src/providers/model_override.rs
jamadeo: it's also accomplishing the same thing as the `complete_with_model` method on the Provider trait

we probably don't want two ways of doing the same thing

crates/goose/src/session/session_manager.rs
tlongwell-block: I added this in when testing haiku and hitting a subagent session creation race condition that caused failures. This was a simple fix. I'll take it out of this PR

crates/goose/src/providers/model_override.rs
tlongwell-block: Overrides are only used for subagents and subrecipes, so a human won't be watching text stream in. No reason to support streaming afaik

crates/goose/src/agents/subagent_execution_tool/tasks.rs
tlongwell-block: @DOsinga that was a bug with the implementation in https://github.com/block/goose/pull/4828 that made an unspecified extension list default to no extensions, which was not the correct behavior

This setup here is how subagents work now and the llms do well with it

crates/goose/src/providers/model_override.rs
tlongwell-block: Yeah, turns out this is nontrivial. I did initially go with complete_with_model, but that work touched a lot more of the codebase than I wanted to for this PR when I put it together for the first draft

The thinking from my notes:

`complete_with_model` doesn't help because the agent always calls `provider.complete()` throughout its execution (never `complete_with_model`), and refactoring to use it would require threading an optional ModelConfig through the entire agent architecture, touching 5-7 files and breaking APIs. Creating a new provider isn't viable either since we only have `Arc<dyn Provider>` with no way to get the provider type name needed for `providers::create()`. ModelOverrideProvider cleanly wraps any provider to override just the model config while keeping everything else transparent to the agent - it's basically the Decorator pattern doing exactly what it's meant to do in about 100 lines with zero API changes.

crates/goose/src/agents/subagent_execution_tool/tasks.rs
tlongwell-block: We want subrecipes and subagents to be able to override
* provider and model
* just the model

Being able to override the provider but use the same model is... messy, probably? Most providers use different naming schemes for the same model (think custom Databricks names vs prefixed openrouter names vs non-prefixed first party llm names). I think that would be much more trouble than it is worth to support

crates/goose/src/agents/subagent_handler.rs
tlongwell-block: I just made sure it worked with the new execution pathway. I can definitely remove it as a feature, but the goal of this PR was just a feature complete cutover away from the cli shell out

crates/goose/src/agents/subagent_execution_tool/task_execution_tracker.rs
tlongwell-block: I think we could do this if I add itertools as a dep for the `goose` crate. Otherwise, I think this is as clean as we can make this unless you had something specific in mind?

crates/goose/src/agents/subagent_execution_tool/tasks_manager.rs
tlongwell-block: I think so for this PR since it is already chunky. Though, yes, I would like separate test files for almost everything.

crates/goose/src/agents/subagent_handler.rs
tlongwell-block: Great catch. Added apply_recipe_components

crates/goose/src/agents/subagent_execution_tool/utils/tests.rs
tlongwell-block: I think we'd need something like a `test_helpers` module to consolidate? This PR is probably not the right place for that rework

crates/goose/src/providers/model_override.rs
DOsinga: I don't think it is the way though. adding the name to the provider seems like a much better approach and something that is missing elsewhere

crates/goose/src/agents/subagent_handler.rs
DOsinga: we need to clean up the cruft as we go, that's the only way

crates/goose/src/providers/model_override.rs
tlongwell-block: Yeah, agree. I will take a crack at it again and try to keep it relatively minimal so we don't blow up this PR

tlongwell-block: Okay, this gets messy really quickly. Needed to modify all 30~ providers to make this work as suggested. If you want to add a `provider_name` to every provider, we would be able to instantiate a new provider when the model is overridden. But we do lose anything around rate limiting and connection pooling. It also does pull in auth again from scratch, which may not be what we want when we want to *only* override the model and not the full provider

I also looked at setting a model override on Agent itself in lieu of changing every provider, but then that makes Agent messier and it starts to depend on the internals of the providers more than it probably should. It also requires changes to reply_internal

I think `ModelOverrideProvider` ends up being cleaner and much lower maintenance going forward since it works with any current and future provider. It's a straightforward solution

DOsinga: I am nog convinced. all providers should be created using the factory::create method. we know the name there. that is the name we should set. that name should be independent of the class that actually does the work.

tlongwell-block: This is the most minimal setup I've been able to put together so far https://github.com/block/goose/pull/5338

We can merge it into this PR from there if it looks good

tlongwell-block: Able to do this thanks to https://github.com/block/goose/pull/5399

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
```diff
@@ -116,7 +113,7 @@ pub fn create_dynamic_task_tool() -> Tool {
 
     Tool::new(
         DYNAMIC_TASK_TOOL_NAME_PREFIX.to_string(),
```
DOsinga, line 115: above, can you remove the .expects()? it would crash the server if that broke

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
```diff
@@ -81,9 +81,6 @@ pub struct TaskParameter {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub retry: Option<JsonObject>,
 
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub context: Option<Vec<String>>,
-
     #[serde(skip_serializing_if = "Option::is_none")]
```
DOsinga, line 84: so yeah, this entire struct, why can't it be a recipe? that way the tool would have no choice but provide us with a valid recipe, rather than Vec<JsonObject>s that might or might not fit what we expect?

that would solve: https://github.com/block/goose/issues/5241

crates/goose/src/agents/subagent_execution_tool/task_execution_tracker.rs
```diff
@@ -28,25 +29,23 @@ const THROTTLE_INTERVAL_MS: u64 = 250;
 const COMPLETION_NOTIFICATION_DELAY_MS: u64 = 500;
 
 fn format_task_metadata(task_info: &TaskInfo) -> String {
-    if let Some(params) = task_info.task.get_command_parameters() {
-        if params.is_empty() {
-            return String::new();
+    // If we have parameter values, format them nicely
+    if let Some(ref params) = task_info.task.payload.parameter_values {
+        if !params.is_empty() {
+            // Filter out internal parameters like recipe_dir
```
DOsinga: remove

DOsinga (approved):
good improvement. can you have a look at including the recipe as a schema element? that would really land this

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
tlongwell-block: Yes, we should do this. I've opened https://github.com/block/goose/issues/5620 to keep this PR reasonable

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
tlongwell-block: This ends up opening a bit of a can of worms. This should get looked at when we move the tools to a platform extension

PR #5083, opened 2025-10-08
Clean room cleanup and resume chat in same window

zanesq:
## Summary

- fixed a couple backwards compatibility changes that were breaking loading a session with tool calls and message conversion for compaction - fixed the failing tests also (goose helped me)
- added flag conditional resume chat session in current window from hub only so we can test that path easier, had some related changes in useAgent but seems to work ok the session loads at least
- turned the flag to false for merging to main
- verified chat and summarization is working like it was with flag off

### DIFF
--- a/ui/desktop/src/components/context_management/ContextManager.tsx
+++ b/ui/desktop/src/components/context_management/ContextManager.tsx
@@ -53,21 +53,14 @@ export const ContextManagerProvider: React.FC<{ children: React.ReactNode }> = (
           sessionId: sessionId,
         });
 
-        // Convert server messages to include id and created fields if missing
-        const convertedMessages = summaryResponse.messages.map((msg) => ({
-          ...msg,
-          id: msg.id || `msg-${Date.now()}-${Math.random().toString(36).substring(2, 10)}`,
-          created: msg.created || Math.floor(Date.now() / 1000),
-        }));
-
-        setMessages(convertedMessages);
+        setMessages(summaryResponse.messages);
 
         // Only automatically submit the continuation message for auto-compaction (context limit reached)
         // Manual compaction should just compact without continuing the conversation
         if (!isManual) {
           // Automatically submit the continuation message to continue the conversation
           // This should be the third message (index 2) which contains the "I ran into a context length exceeded error..." text
-          const continuationMessage = convertedMessages[2];
+          const continuationMessage = summaryResponse.messages[2];
           if (continuationMessage) {
             setTimeout(() => {
               append(continuationMessage);
--- a/ui/desktop/src/components/context_management/__tests__/ContextManager.test.tsx
+++ b/ui/desktop/src/components/context_management/__tests__/ContextManager.test.tsx
@@ -7,7 +7,6 @@ import { ContextManageResponse, Message } from '../../../api';
 // Mock the context management functions
 vi.mock('../index', () => ({
   manageContextFromBackend: vi.fn(),
-  convertApiMessageToFrontendMessage: vi.fn(),
 }));
 
 const mockManageContextFromBackend = vi.mocked(contextManagement.manageContextFromBackend);
@@ -102,6 +101,7 @@ describe('ContextManager', () => {
   describe('handleAutoCompaction', () => {
     it('should successfully perform auto compaction with server-provided messages', async () => {
       // Mock the backend response with 3 messages: marker, summary, continuation
+      // Note: Server messages may not have id/created, which will be added by the code
       mockManageContextFromBackend.mockResolvedValue({
         messages: [
           {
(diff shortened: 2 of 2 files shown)


ui/desktop/src/updates.ts
```diff
@@ -3,4 +3,4 @@ export const COST_TRACKING_ENABLED = true;
 export const ANNOUNCEMENTS_ENABLED = false;
 export const CONFIGURATION_ENABLED = true;
 
-export const USE_NEW_CHAT = true;
+export const USE_NEW_CHAT = false;
```
DOsinga: should we make this an environment variable instead?

ui/desktop/src/components/BaseChat2.tsx
```diff
@@ -74,6 +74,13 @@ function BaseChatContent({
 
   const [messages, setMessages] = useState(chat?.messages || []);
 
+  // Update messages when chat changes (e.g., when resuming a session)
+  useEffect(() => {
+    if (chat?.messages) {
+      setMessages(chat.messages);
+    }
+  }, [chat?.messages, chat?.sessionId]);
```
DOsinga, line 82: this seems good for now, but needs further refactoring I think. I think we should just get rid of useAgent and load the session here from the server - we can add a caching layer to make it look smooth

ui/desktop/src/components/ToolCallWithResponse.tsx
```diff
@@ -42,8 +42,15 @@ export default function ToolCallWithResponse({
...
-  const toolCall = toolRequest.toolCall as { name: string; arguments: Record<string, unknown> };
-  if (!toolCall) {
+  // Handle both the wrapped ToolResult format and the unwrapped format
+  // The server serializes ToolResult<T> as { status: "success", value: T } or { status: "error", error: string }
+  const toolCallData = toolRequest.toolCall as Record<string, unknown>;
+  const toolCall =
+    toolCallData?.status === 'success'
+      ? (toolCallData.value as { name: string; arguments: Record<string, unknown> })
+      : (toolCallData as { name: string; arguments: Record<string, unknown> });
+
```
DOsinga, line 52: I can see that we need this, but does this work with the old setup? looks vaguely scary

ui/desktop/src/components/context_management/ContextManager.tsx
```diff
@@ -53,14 +53,21 @@ export const ContextManagerProvider: React.FC<{ children: React.ReactNode }> = (
           sessionId: sessionId,
         });
 
-        setMessages(summaryResponse.messages);
+        // Convert server messages to include id and created fields if missing
+        const convertedMessages = summaryResponse.messages.map((msg) => ({
+          ...msg,
+          id: msg.id || `msg-${Date.now()}-${Math.random().toString(36).substring(2, 10)}`,
+          created: msg.created || Math.floor(Date.now() / 1000),
+        }));
```
DOsinga: yeah, this should not be needed here

ui/desktop/src/hooks/useAgent.ts
```diff
@@ -62,24 +62,35 @@ export function useAgent(): UseAgentReturn {
...
-        const messages = agentSession.conversation || [];
-        return {
-          sessionId: agentSession.id,
-          title: agentSession.recipe?.title || agentSession.description,
-          messageHistoryIndex: 0,
-          messages,
-          recipe: agentSession.recipe,
-          recipeParameters: agentSession.user_recipe_values || null,
-        };
+      // If agent is already initialized, we can resume without full reinitialization
```
DOsinga: I think we should get rid of useAgent in the new flow

ui/desktop/src/components/ToolCallWithResponse.tsx
zanesq: yes, agreed verified it fixes the issue with the old setup

ui/desktop/src/hooks/useAgent.ts
zanesq: ok will come back to this

ui/desktop/src/updates.ts
zanesq: good call will do that and merge after

ui/desktop/src/hooks/useAgent.ts
DOsinga: it might be safer to revert this? this changes the behavior for the current setup which seems brittle

ui/desktop/src/components/context_management/ContextManager.tsx
zanesq: good call updated the test instead

ui/desktop/src/updates.ts
zanesq: changed it to use alpha flag since we already have that, `npm run start-alpha-gui`

Issue #4397, opened 2025-08-28
Adding custom provider adds 2 duplicates; Can't add LM-Studio (Custom Provider) on Windows 11


lorenz-graf:
**Describe the bug**
Can't get LM-Studio to work since adding a new custom provider wrongly adds two providers (one seems to be a copy of openai). Probably because of that the custom provider doesn't work. My attempt was adding LM-Studio.

There is a video and conversation about the issue reproduction in this thread https://discord.com/channels/1287729918100246654/1409870301784838328

**To Reproduce**
1. Be on windows
2. Add Custom Provider
3. Use OpenAI compatible provider
4. Two providers are created

**Expected behavior**
It should only create one provider (e.g. LM-Studio)

**Screenshots**
Check the discord thread for screenshots and screen recordings.

**Please provide following information:**
 - **OS & Arch:** windows 11
 - **Interface:** ui
 - **Version:** 1.6.0
 - **Extensions enabled:** none
 - **Provider & Model:** qwen3 coder 30B (but didn't get to interact since the provider didn't work)

**Additional context**
please check the discord thread for any more details.

bstin:
The same happens on Linux trying to add a custom provider for `llama.cpp` providers. 

In addition to adding **duplicate** providers, it seems to be completely **ignoring** the port setting of URL. 

See pics. 

1. Define New provider (notice port). 

<img width="600" height="602" alt="Image" src="https://github.com/user-attachments/assets/51afa7ad-a33b-4b27-b441-3f90d0967a41" />

2. After adding notice now 2 providers have been added?!?
<img width="1092" height="903" alt="Image" src="https://github.com/user-attachments/assets/6dbbea41-aeae-4973-9b13-b5ab620bbe4b" />

3. Trying to use either of the providers doesn't work. Notice port is missing from URL if I choose the "right" one. 

<img width="904" height="732" alt="Image" src="https://github.com/user-attachments/assets/224b1707-1f8a-4f9c-aa2f-bfbaaff26c45" />

cbruyndoncx:
see https://github.com/block/goose/pull/4514 for the fix for duplicates (any custom provider)

port setting not explicitly addressed yet, would need to retest to make sure

taniacryptid:
This bug is also happening for a [community member trying to set up with ollama](https://discord.com/channels/1287729918100246654/1413532084286525470/1413554456414654544)! @alexhancock

DOsinga:
this should be fixed after #5084

[...]

Issue #4029, opened 2025-08-12
OpenAI key becomes the string "true" in the edit dialog


sings-to-bees-on-wednesdays:
**Describe the bug**

If you configure the OpenAI provider, but then go to edit it, the API key is reset to the string "true".

(It shows up as ••••)

**To Reproduce**
1. Start Goose for the first time
2. Configure OpenAI provider
3. Go to Settings -> Configure Providers -> OpenAI

**Expected behavior**
Should preserve entered values.

**Screenshots**

<img width="1033" height="1137" alt="Image" src="https://github.com/user-attachments/assets/96b2ea15-adf7-4b69-9bd6-10ffd272fa3e" />

**Please provide following information:**
 - **OS & Arch:** Windows 11
 - **Interface:** UI
 - **Version:** 1.3.0
 - **Extensions enabled:** No
 - **Provider & Model:** OpenAI

**Additional context**

cbruyndoncx:
Fixed as part of https://github.com/block/goose/pull/4557 PR for UI custom providers

DOsinga:
also #5084

zanesq:
Is this fixed now can we close?

cbruyndoncx:
@zanesq I think so, i have not seen this happening any more while testing the custom providers. I believe it is the same 'edit' code independent of provider type. 
PS I can not close this myself.

PR #5084, opened 2025-10-08
Declarative providers

DOsinga:
## Providers loaded from the file system

### DIFF
--- a/crates/goose/src/providers/utils.rs
+++ b/crates/goose/src/providers/utils.rs
@@ -937,12 +937,12 @@ mod tests {
                     "The model 'gpt-5' does not exist (code: model_not_found, type: invalid_request_error) (status 404)".to_string(),
                 )),
             ),
-            // Non-JSON body error (tests parse failure path)
+            // Non-JSON body error (tests 413 PAYLOAD_TOO_LARGE -> ContextLengthExceeded)
             (
                 413,
                 Some(Value::String("Payload Too Large".to_string())),
-                Err(ProviderError::RequestFailed(
-                    "Request failed with status: 413 Payload Too Large".to_string(),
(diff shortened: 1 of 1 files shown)


DOsinga:
fixes #4397

DOsinga:
fixes #4221

DOsinga:
fixes #4029

DOsinga:
@michaelneale you were saying that custom providers were confusing and so they were. or really it was all kinda broken. this refactors how things work and allows one to actually edit a custom provider(!). it also bring sin the concept of internal custom providers which should allow us to reduce code duplication. I did it with groq and it worked well. it even now allows streaming with groq where before we didn't have that.

there's still work to be done. the frontend has a checkbox for no api key required, but that doesn't seem to be hooked up to anything. there's probably a lot more features (toolshim?) that you'd want to be able to set from the json file but we can bring them in one by one

michaelneale:
very nice @DOsinga - didn't get a chance ot try it but I love this approach, this seems the right approach.

DOsinga:
fixes #5148

crates/goose/src/providers/utils.rs
```diff
@@ -937,12 +937,12 @@ mod tests {
...
                 )),
             ),
-            // Non-JSON body error (tests parse failure path)
+            // Non-JSON body error (tests 413 PAYLOAD_TOO_LARGE -> ContextLengthExceeded)
             (
                 413,
                 Some(Value::String("Payload Too Large".to_string())),
-                Err(ProviderError::RequestFailed(
-                    "Request failed with status: 413 Payload Too Large".to_string(),
+                Err(ProviderError::ContextLengthExceeded(
```
DOsinga, line 944: hmmm, how is this related ...

Issue #5044, opened 2025-10-06
Using `--name` Without `--resume` No Longer Works As Expected


wpfleger96:
**Describe the bug**

The `--name` flag for `goose session` and `goose run` commands is broken. When used without `--resume`, it incorrectly searches for an existing session instead of creating a new session with the specified name which used to work and has regressed

This seems to have been introduced in #4648, and I touched on this in #4937 but didn't address the core bug 😅 

**To Reproduce**
1. Run: `goose session --name test-session`
2. Observe error: "No session found with name 'test-session'"

**Expected behavior**
1. A new session should be created with description field set to "test-session"
3. The session ID should be auto-generated in the format `YYYYMMDD_X`
4. The user can later resume this session using `goose session -r --name test-session`

**Please provide following information:**
 - **OS & Arch:** Darwin (Sequoia 15.6.1) arm64
 - **Interface:** CLI
 - **Version:** unreleased, built off latest main
 - **Extensions enabled:** `computercontroller`, `developer`
 - **Provider & Model:** Databricks + Claude Sonnet 4

DOsinga:
hmm, that's annoying. one of the issues is that if you set the name of a session, the system will still override it, so we need a flag indicating that the user has set it

wpfleger96:
@DOsinga yeah, it's a little confusing what the relationship between a session's ID, name, and description is supposed to be (and which are supposed to be user-configurable)

in the session DB schema we have:
- `id`
    - autogenerated as `YYYYMMDD_X` (`X` is auto incrementing per-day)
    - the help text for `--session-id` implies that you can specify custom IDs for new sessions, but this doesn't actually seem like it's user-configurable in the code (always auto generated)
- `description`
    - Initially defaults to `CLI Session`, gets updated with LLM-generated value after 3 turns
    - Even if `--name` did work here, the custom name would get overwritten after 3 turns (as you already mentioned)

Historically you could create custom named sessions with `--name`, but IIRC that would actually set the session ID to whatever you passed to `--name` since there is no such "name" field for a session. For example in `oncall-buddy`, we'd use custom session IDs like `YYYYMMDD_HHMMSS_OncallBuddy` to denote `oncall-buddy` vs other Goose sessions. But since the SQLite DB migration session IDs are auto generated which breaks that previous behavior.

I think if we can clarify whether we want a session to have a name, a description, or both, and clarify if either/both should be user-configurable, that would help clear this up

DOsinga:
[...]

PR #5085, opened 2025-10-08
Standardize Session Name Attribute

wpfleger96:
## Summary
This PR standardizes session attribute naming by migrating from the `description` field to `name` throughout the codebase, with full backwards compatibility support. The change introduces a `user_set_name` boolean flag to differentiate between user-provided names and auto-generated names, preventing AI from overwriting custom session names.

## Key Changes
Backend: Updated `SessionManager` and `Session` struct to use `name` field with `user_set_name` flag. Database schema migration to v3 renames the `sessions.description` column to `sessions.name` and adds `user_set_name` column. Serde alias ensures backwards compatibility for JSON deserialization. The `maybe_update_name` method now respects user-set names.

Frontend: Added `sessionCompat.ts` utility module providing `getSessionName()` helper that gracefully handles both old (description) and new (name) formats. Updated components (`SessionHistoryView`, `SessionItem`, `SessionListView`, `SessionsView`) to use the compatibility layer. OpenAPI schema and TypeScript types updated to reflect new field names.

Database Migration: Includes `goose-db-helper.sh` script for safe schema migrations with backup capabilities. Migration tests ensure data integrity during the transition. Database changes are non-destructive with automatic backups before migration.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
I added 2 unit tests for this work, and I manually tested with my local Goose 

### Related Issues
Related to #5044 

Before:  

After:

### DIFF
--- a/crates/goose-cli/src/session/builder.rs
+++ b/crates/goose-cli/src/session/builder.rs
@@ -313,7 +313,7 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> CliSession {
     } else {
         let session = SessionManager::create_session(
             std::env::current_dir().unwrap(),
-            "CLI session".to_string(),
+            "CLI Session".to_string(),
         )
         .await
         .unwrap();


DOsinga:
let me know when you need a review

wpfleger96:
@DOsinga this should be ready for review now!

let me know what you think about the `goose-db-helper.sh` script, it was pretty helpful for me while testing but it ended up getting pretty massive... maybe that should be refactored into a proper Python package

wpfleger96:
[Test failure](https://github.com/block/goose/actions/runs/18546001927/job/52866139240?pr=5085) is unrelated and addressed in #5207

crates/goose-cli/src/cli.rs
```diff
@@ -75,7 +75,7 @@ async fn get_session_id(identifier: Identifier) -> Result<String> {
 
         sessions
             .into_iter()
-            .find(|s| s.id == name || s.description.contains(&name))
+            .find(|s| s.name == name)
```
DOsinga: are we ok with this change in functionality

crates/goose-server/src/routes/session.rs
```diff
@@ -135,7 +135,7 @@ async fn update_session_description(
     }
 
     SessionManager::update_session(&session_id)
-        .description(request.description)
+        .name(request.description)
```
DOsinga: should we not rename this too?

crates/goose-server/src/routes/schedule.rs
```diff
@@ -69,7 +69,7 @@ fn default_limit() -> u32 {
 #[serde(rename_all = "camelCase")]
 pub struct SessionDisplayInfo {
     id: String,          // Derived from session_name (filename)
-    name: String,        // From metadata.description
+    name: String,        // From metadata.name
     created_at: String,  // Derived from session_name, in ISO 8601 format
     working_dir: String, // from metadata.working_dir (as String)
```
DOsinga: can you remove these comments, the third one for one isn't true

crates/goose/src/session/session_manager.rs
```diff
@@ -27,7 +27,10 @@ pub struct Session {
     pub id: String,
     #[schema(value_type = String)]
     pub working_dir: PathBuf,
-    pub description: String,
+    #[serde(alias = "description")]
+    pub name: String,
+    #[serde(default)]
+    pub user_set_name: bool,
```
DOsinga, line 34: can we get away with not having these #[serde tags and handle it in the migration?

crates/goose/src/session/session_manager.rs
```diff
@@ -166,10 +176,10 @@ impl SessionManager {
             .map(Arc::clone)
     }
 
-    pub async fn create_session(working_dir: PathBuf, description: String) -> Result<Session> {
+    pub async fn create_session(working_dir: PathBuf, name: Option<String>) -> Result<Session> {
```
DOsinga: why do we want to make the session name optional? I could see the case for dropping this completely and just either wait for the system or the user to set it

crates/goose/src/session/session_manager.rs
```diff
@@ -88,8 +93,13 @@ impl SessionUpdateBuilder {
...
     }
 
-    pub fn description(mut self, description: impl Into<String>) -> Self {
-        self.description = Some(description.into());
+    pub fn name(mut self, name: impl Into<String>) -> Self {
+        self.name = Some(name.into());
+        self
+    }
+
+    pub fn user_set_name(mut self, user_set: bool) -> Self {
```
DOsinga: this probably shouldn't be a public thing, instead add this to the name(..) method and then add a separate method system_name or some such - you don't want callers to have to bother with a separate boolean I think

ui/desktop/src/components/BaseChat2.tsx
```diff
@@ -107,7 +108,7 @@ function BaseChatContent({
           const conversation = session.conversation || [];
           const loadedChat: ChatType = {
             sessionId: session.id,
-            title: session.description || 'Untitled Chat',
+            title: getSessionName(session) || 'Untitled Chat',
```
DOsinga: should we rename title here too?

ui/desktop/src/utils/sessionCompat.ts
```diff
@@ -0,0 +1,23 @@
+import type { Session } from '../api/types.gen';
```
DOsinga: why do we need this? the Session type is updated, no? and this comes from the database?

DOsinga (approved):

wpfleger96:
@DOsinga I think I addressed all your feedback, would you mind giving this another glance over before I merge?

ui/desktop/src/components/sessions/SessionHistoryView.tsx
```diff
@@ -270,7 +270,7 @@ const SessionHistoryView: React.FC<SessionHistoryViewProps> = ({
         <div className="flex-1 flex flex-col min-h-0 px-8">
           <SessionHeader
             onBack={onBack}
-            title={session.description || 'Session Details'}
+            title={session.name || 'Session Details'}
```
DOsinga: probably happened before, but since name is now required, the only reason we need the || is when the user explicitly sets the name to "" - maybe we should just block that instead?

crates/goose/src/session/session_manager.rs
```diff
@@ -88,8 +94,17 @@ impl SessionUpdateBuilder {
...
+        self.user_set_name = Some(true);
+        self
+    }
+
+    /// System-generated name (allows LLM to update it later)
+    pub fn system_name(mut self, name: impl Into<String>) -> Self {
+        self.name = Some(name.into());
+        self.user_set_name = Some(false);
         self
     }
```
DOsinga, line 113: remove the comments or change the name of the functions

crates/goose/src/session/session_manager.rs
```diff
@@ -166,10 +181,11 @@ impl SessionManager {
             .map(Arc::clone)
     }
 
-    pub async fn create_session(working_dir: PathBuf, description: String) -> Result<Session> {
+    /// `None` allows LLM to automatically generate the name, `Some` preserves the user's choice.
```
DOsinga: I don't think you want that behavior. We now set the system names to CLI session, can we just revert to the previous behavior? that or just drop the name from the create_session and make it explicit at the calling side

crates/goose/src/session/session_manager.rs
```diff
@@ -1143,4 +1205,61 @@ mod tests {
...
+        let storage = Arc::new(SessionStorage::create(&db_path).await.unwrap());
+
+        let imported = storage.import_session(OLD_FORMAT_JSON).await.unwrap();
+
+        assert_eq!(imported.name, "Old format session");
+        assert!(imported.user_set_name);
+        assert_eq!(imported.working_dir, PathBuf::from("/tmp/test"));
+    }
+
+    #[tokio::test]
```
DOsinga: this test doesn't make sense to me

wpfleger96:
@DOsinga ok this should be ready for another glance over, per your feedback I reverted back to requiring a name when starting a session

DOsinga:
yeah merge when ready!

michaelneale:
this isn't backwards compatible if you roll back a version of goose (which may happen): 

image




```
17:46:43.054 › from renderer: [UNHANDLED REJECTION] Error: Failed to create session: error returned from database: (code: 1) table sessions has no column named description
```


or in cli: 


thread 'main' panicked at /Users/micn/Development/goose/crates/goose-cli/src/session/builder.rs:319:10:
called `Result::unwrap()` on an `Err` value: error returned from database: (code: 1) table sessions has no column named description

So this is only one way compatible. I think need to roll it back unless there is a strategy for this.

DOsinga:
it's a good point. we should roll this back and then do it by adding the column name and keep description around for a while

wpfleger96:
👍 great point, let's roll back and  I'll try again next week and be careful to make this more backwards compatible

DOsinga:
thank you @wpfleger96 !

PR #5088, opened 2025-10-09
basecamp2 - Resume chat session from server with a cache

zanesq:
## Summary
Added support for resuming chat session from server with a cache. Removed reliance on useAgent in Pair2.

Note: this gets us most of the way there but still some rough edges with race conditions from hub (if a session is loaded while the empty chat is still initializing) and improvements to the initial state before loading the chat session.

### DIFF
--- a/ui/desktop/src/components/BaseChat2.tsx
+++ b/ui/desktop/src/components/BaseChat2.tsx
@@ -10,16 +10,17 @@ import { MainPanelLayout } from './Layout/MainPanelLayout';
 import ChatInput from './ChatInput';
 import { ScrollArea, ScrollAreaHandle } from './ui/scroll-area';
 import { useFileDrop } from '../hooks/useFileDrop';
-import { Message } from '../api';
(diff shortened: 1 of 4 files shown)


ui/desktop/src/components/BaseChat2.tsx
```diff
@@ -72,17 +76,62 @@ function BaseChatContent({
...
+  // Load session on mount if resumeSessionId is provided
+  useEffect(() => {
+    const needsLoad = resumeSessionId && !hasLoadedSessionRef.current;
+
+    if (needsLoad) {
+      hasLoadedSessionRef.current = true;
+      setSessionLoadError(null);
+
+      // Set chat to empty session to indicate loading state
+      // todo: set to null instead and handle that in other places
```
zanesq, line 94: I originally set chat to `null` but it started having cascading effects all over the codebase setting the types to `ChatType | null` so opted for a less disruptive change for now and figured we can come back to it later

PR #5090, opened 2025-10-09
Sort providers in alphabetical vs random

katzdave:
I know we'd also discussed potentially having different tiers of provider support within our testing; is this something we want to potentially expose here?


katzdave:
> do we need to sort them in the backend and the frontend? I would just go frontend.
> 
> I've been working on this: #5084 that does introduce the notion of a prefered provider

cli on goose configure drives me absolutely insane with random order.

PR #5091, opened 2025-10-09
Provider error proxy for simulating various types of errors

katzdave:
Still in flight as I encounter issues + test more providers but already starting to be useful in uncovering some issues of errors we handle poorly.

Providers to test:
- [x] Openai
- [x] Anthropic
- [x] Google
- [x] Databricks
- [ ] Openrouter
- [ ] Tetrate


scripts/provider-error-proxy/pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[project]
```
DOsinga, line 1: for a one file python thing, you can just add the dependencies to to main file

DOsinga (approved):

michaelneale:
great idea - I think anthropic is really needed as it seems to have the most issues (and is one of the more popular 1st party ones).

DOsinga:
have a look at what copilot said and let's merge this?

PR #5092, opened 2025-10-09
Add dev and alpha environment indicator

zanesq:
## Summary
Added subtle environment badge in bottom left sidebar if dev or alpha to help distinguish between the production app and local dev environment.

Verified locally does not show in the bundled app.

Screenshot 2025-10-09 at 9 12 21 AM
Screenshot 2025-10-09 at 9 04 54 AM
Screenshot 2025-10-09 at 9 05 24 AM

### DIFF
--- a/ui/desktop/src/components/GooseSidebar/EnvironmentBadge.tsx
+++ b/ui/desktop/src/components/GooseSidebar/EnvironmentBadge.tsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import { Tooltip, TooltipContent, TooltipTrigger } from '../ui/Tooltip';
+
+const EnvironmentBadge: React.FC = () => {
+  const isAlpha = process.env.ALPHA;
+  const isDevelopment = import.meta.env.DEV;
+
+  // Don't show badge in production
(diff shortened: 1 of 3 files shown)


ui/desktop/src/components/GooseSidebar/EnvironmentBadge.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { Tooltip, TooltipContent, TooltipTrigger } from '../ui/Tooltip';
+
+const EnvironmentBadge: React.FC = () => {
+  const isAlpha = process.env.ALPHA;
+  const isDevelopment = import.meta.env.DEV;
```
DOsinga, line 6: do we set these? there's also a test to see if this has node things set

zanesq: yup `import.meta.env.DEV` is provided by Vite
`process.env.ALPHA` is set with `"start-alpha-gui": "ALPHA=true npm run start-gui"` in package.json

PR #5094, opened 2025-10-09
Fallback session name

katzdave:
Goose completely crashes if the provider throws an error on this request.

I wonder if we should have a mechanism to try again like:

user_message_count % 10 update the session name, but might be confusing to people if these start changing. Could also match this unnamed session string with that pattern and only retry those occasionally.


katzdave:
Doesn't seem to be happening anymore; definitely broken on 1.9.x but fixed somewhere in the middle.

PR #5102, opened 2025-10-09
feat(mcp): add support for mcp sampling

alexhancock:
## feature

Adds [MCP sampling](https://modelcontextprotocol.io/specification/2025-06-18/client/sampling) support to goose, so that goose routes requests between extensions and providers when needed. The flow is similar to the below diagram from the MCP docs:
Screenshot 2025-10-09 at 5 15 02 PM
NOTE: I did not add response review as it seemed like the extension would handle what came back and do with it what it can, and it would be an additional UX speedbump.

## goose specific UX

Added a human-in-the-loop approval flow with a modal that allows the user to confirm or deny the sampling message before it is sent to the llm
Screenshot 2025-10-09 at 1 45 08 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2597,6 +2597,7 @@ dependencies = [
  "axum 0.8.1",
  "base64 0.21.7",
  "blake3",
+ "bytes",
  "chrono",
  "criterion",
  "ctor",
--- a/crates/goose-server/src/commands/agent.rs
+++ b/crates/goose-server/src/commands/agent.rs
@@ -34,6 +34,7 @@ pub async fn run() -> Result<()> {
         .allow_headers(Any);
 
     let app = crate::routes::configure(app_state)
+        .await
         .layer(middleware::from_fn_with_state(
             secret_key.clone(),
             check_token,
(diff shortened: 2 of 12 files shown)


alexhancock:
@DOsinga Thanks. I had the thought as well that elicitation will work a similar way, and agreed it's also similar shaped to tool call approval. I can spike on a way to unify the approach (maybe something in or added to ACP can also help standardize further?)

As to whether we even need human in the loop, I think so.

Other MCP clients implement it and it's recommended strongly as a SHOULD in the MCP spec here: https://modelcontextprotocol.io/specification/2025-06-18/client/sampling#user-interaction-model

I think the reason is so that an MCP server doesn't just silently use your model connection for its own purposes. Multiple ways to solve that of course as you say, but this seems to be the standard approach to giving the user visibility and authority. The UX implemented here is the same as currently implemented by other major clients (vscode, etc).

alexhancock:
Switching to draft while I work out a more generalized approach to human in the loop approval that can work for more things than just sampling

alexhancock:
Working in a couple different branches now. There will be an updated PR for this soon.

PR #5103, opened 2025-10-09
Added global per session cacheing and sse session streaming endpoint and support to basechat2

zanesq:
## Summary
Added chat session specific cacheing in global context and streaming support for chat in preparation for multi chat in same window.

Still some rough edges with session resume from hub when the default chat hasn't initialized fully but works nice otherwise.

Also fixed an issue with progressive message list where messages re-rendering as the response is generated happening on main also.

https://github.com/user-attachments/assets/0a818f46-6e85-4b1e-952f-05b574f265be

### DIFF
--- a/crates/goose-server/src/routes/reply.rs
+++ b/crates/goose-server/src/routes/reply.rs
@@ -85,6 +85,8 @@ pub struct ChatRequest {
     session_id: String,
     recipe_name: Option<String>,
     recipe_version: Option<String>,
+    #[serde(default)]
+    background_mode: bool,
 }
 
 pub struct SseResponse {
@@ -144,7 +146,12 @@ enum MessageEvent {
     Ping,
 }
 
-async fn stream_event(event: MessageEvent, tx: &mpsc::Sender<String>) -> bool {
+async fn stream_event(
+    event: MessageEvent,
+    tx: &mpsc::Sender<String>,
+    cancel_token: &CancellationToken,
+    background_mode: bool,
+) -> bool {
     let json = serde_json::to_string(&event).unwrap_or_else(|e| {
         format!(
             r#"{{"type":"Error","error":"Failed to serialize event: {}"}}"#,
@@ -152,10 +159,14 @@ async fn stream_event(event: MessageEvent, tx: &mpsc::Sender<String>) -> bool {
         )
     });
     if tx.send(format!("data: {}\n\n", json)).await.is_err() {
-        tracing::info!("client hung up - continuing agent task");
-        // Don't cancel the agent task when client disconnects
-        // Agent should continue running and the session will remain in_use=true
-        // so that when the client reconnects via the session stream, they can see progress
+        if background_mode {
+            tracing::info!(
+                "client disconnected - continuing agent task in background (ALPHA mode)"
+            );
+        } else {
+            tracing::info!("client disconnected - cancelling agent task (backwards compatibility)");
+            cancel_token.cancel();
+        }
         return false;
     }
     true
@@ -210,12 +221,19 @@ pub async fn reply(
     let cancel_token = CancellationToken::new();
 
     let messages = Conversation::new_unvalidated(request.messages);
+    let background_mode = request.background_mode;
+
+    if background_mode {
+        let mut tokens = state.session_cancellation_tokens.lock().await;
+        tokens.insert(session_id.clone(), cancel_token.clone());
+    }
 
     let task_cancel = cancel_token.clone();
     let task_tx = tx.clone();
+    let task_state = state.clone();
 
     drop(tokio::spawn(async move {
-        let agent = match state.get_agent(session_id.clone()).await {
+        let agent = match task_state.get_agent(session_id.clone()).await {
             Ok(agent) => agent,
             Err(e) => {
                 tracing::error!("Failed to get session agent: {}", e);
@@ -224,6 +242,8 @@ pub async fn reply(
                         error: format!("Failed to get session agent: {}", e),
                     },
                     &task_tx,
+                    &task_cancel,
+                    background_mode,
                 )
                 .await;
                 return;
@@ -239,6 +259,8 @@ pub async fn reply(
                         error: format!("Failed to read session: {}", e),
                     },
                     &task_tx,
+                    &task_cancel,
+                    background_mode,
                 )
                 .await;
                 return;
@@ -270,6 +292,8 @@ pub async fn reply(
                         error: e.to_string(),
                     },
                     &task_tx,
+                    &task_cancel,
+                    background_mode,
                 )
                 .await;
                 return;
@@ -288,7 +312,7 @@ pub async fn reply(
                 }
                 _ = heartbeat_interval.tick() => {
                     if client_connected {
-                        client_connected = stream_event(MessageEvent::Ping, &tx).await;
+                        client_connected = stream_event(MessageEvent::Ping, &tx, &task_cancel, background_mode).await;
                     }
                 }
                 response = timeout(Duration::from_millis(500), stream.next()) => {
@@ -302,7 +326,7 @@ pub async fn reply(
 
                             // Only send message to client if it's user_visible and client is still connected
                             if message.is_user_visible() && client_connected {
-                                client_connected = stream_event(MessageEvent::Message { message }, &tx).await;
+                                client_connected = stream_event(MessageEvent::Message { message }, &tx, &task_cancel, background_mode).await;
                             }
                         }
                         Ok(Some(Ok(AgentEvent::HistoryReplaced(new_messages)))) => {
@@ -313,7 +337,7 @@ pub async fn reply(
                         }
                         Ok(Some(Ok(AgentEvent::ModelChange { model, mode }))) => {
                             if client_connected {
-                                client_connected = stream_event(MessageEvent::ModelChange { model, mode }, &tx).await;
+                                client_connected = stream_event(MessageEvent::ModelChange { model, mode }, &tx, &task_cancel, background_mode).await;
                             }
                         }
                         Ok(Some(Ok(AgentEvent::McpNotification((request_id, n))))) => {
@@ -321,7 +345,7 @@ pub async fn reply(
                                 client_connected = stream_event(MessageEvent::Notification{
                                     request_id: request_id.clone(),
                                     message: n,
-                                }, &tx).await;
+                                }, &tx, &task_cancel, background_mode).await;
                             }
                         }
 
@@ -333,6 +357,8 @@ pub async fn reply(
                                         error: e.to_string(),
                                     },
                                     &tx,
+                                    &task_cancel,
+                                    background_mode,
                                 ).await;
                             }
                             break;
@@ -406,8 +432,15 @@ pub async fn reply(
                 reason: "stop".to_string(),
             },
             &task_tx,
+            &task_cancel,
+            background_mode,
         )
         .await;
+
+        if background_mode {
+            let mut tokens = task_state.session_cancellation_tokens.lock().await;
+            tokens.remove(&session_id);
+        }
     }));
     Ok(SseResponse::new(stream))
 }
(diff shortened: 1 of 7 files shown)


crates/goose/src/context_mgmt/auto_compact.rs
```diff
@@ -331,6 +331,7 @@ mod tests {
             conversation: Some(conversation),
             message_count,
             user_recipe_values: None,
+            in_use: false,
```
michaelneale, line 335: feels weird this is repeated here but I assume is following pattern of other fields. (auto compact would always happen on an in_use session I would think)

michaelneale:
very cool - some Q's: what would happen if a separate process starts and streams, will it pick up the same content (and the in_use flag?)

is there a quick way to have a session id and know if it is in_use or not without streaming from it?

crates/goose-server/src/routes/reply.rs
```diff
@@ -215,10 +215,15 @@ pub async fn reply(
     let task_tx = tx.clone();
 
     drop(tokio::spawn(async move {
+        if let Err(e) = SessionManager::mark_in_use(&session_id, true).await {
+            tracing::error!("Failed to mark session as in use: {}", e);
```
DOsinga: this should not be in the routes but in agent reply. also we should check whether the session is in use and raise an error if so. and we need to think about how we make sure we don't crash and now the session is locked forever. maybe this should be a time stamp that we update every tool loop instead? then we can ignore it if is like 10m old

crates/goose-server/src/routes/session.rs
```diff
@@ -256,10 +267,188 @@ async fn import_session(
...
+impl IntoResponse for SessionSseResponse {
+    fn into_response(self) -> axum::response::Response {
+        let stream = self;
+        let body = axum::body::Body::from_stream(stream);
+
+        http::Response::builder()
+            .header("Content-Type", "text/event-stream")
+            .header("Cache-Control", "no-cache")
+            .header("Connection", "keep-alive")
+            .header("X-Accel-Buffering", "no") // Disable nginx buffering
```
DOsinga: nginx?

crates/goose-server/src/routes/session.rs
```diff
@@ -256,10 +267,188 @@ async fn import_session(
...
+}
+
+async fn stream_session_event(event: SessionEvent, tx: &mpsc::Sender<String>) -> bool {
+    let json = serde_json::to_string(&event).unwrap_or_else(|e| {
+        format!(
+            r#"{{"type":"Error","error":"Failed to serialize event: {}"}}"#,
+            e
+        )
+    });
+    tx.send(format!("data: {}\n\n", json)).await.is_ok()
```
DOsinga, line 321: uh, let's not hand craft json

crates/goose-server/src/routes/session.rs
```diff
@@ -256,10 +267,188 @@ async fn import_session(
...
+        (status = 200, description = "Session stream initiated", content_type = "text/event-stream"),
+        (status = 404, description = "Session not found"),
+        (status = 500, description = "Internal server error")
+    ),
+    security(
+        ("api_key" = [])
+    ),
+    tag = "Session Management"
+)]
+async fn stream_session(
```
DOsinga, line 340: I think this is doing too much work to live in routes. and we should make the streaming compatible with agent.reply so that in BaseChat(2) we can check if the session is active and if so stream that or if it is not, ask the user for input and when they have done, stream the agent reply

crates/goose/src/context_mgmt/auto_compact.rs
zanesq: true.. added a comment its needed to fulfill the contract but its not used by compaction

zanesq:
> very cool - some Q's: what would happen if a separate process starts and streams, will it pick up the same content (and the in_use flag?)

Yes! Check out the attached demo video if someone hooks into the stream it will work.

> 
> is there a quick way to have a session id and know if it is in_use or not without streaming from it?

Yes! There's already a quick way to check if a session is in use without streaming using the existing `GET /sessions/{session_id}` endpoint and it now returns in_use

zanesq:
@DOsinga thanks for the feedback, addressed the comments. See video, I added an indicator that the session is in progress for now on refresh or in a separate chat window. We can adjust this display later, added something basic for now.

I will follow up after this with proper stop chat that stops the stream as well currently its only front end.

crates/goose/src/context_mgmt/auto_compact.rs
DOsinga: I would delete the comment again - this is to setup a testing session and so it is not in_use. compacting itself should set it in use (if one client hits compact, another client should not be able to send messages to it)

crates/goose/src/session/session_manager.rs
```diff
@@ -1116,4 +1261,284 @@ mod tests {
...
+
+        let session = storage
+            .create_session(PathBuf::from("/tmp/test"), "Test".to_string())
+            .await
+            .unwrap();
+
+        let initial_updated_at = session.updated_at;
+
+        // Wait a bit to ensure timestamp difference (SQLite datetime has second precision)
+        tokio::time::sleep(Duration::from_secs(2)).await;
```
DOsinga, line 1370: no

crates/goose-server/src/state.rs
```diff
@@ -6,13 +6,17 @@ use std::path::PathBuf;
...
+
 #[derive(Clone)]
 pub struct AppState {
     pub(crate) agent_manager: Arc<AgentManager>,
     pub recipe_file_hash_map: Arc<Mutex<HashMap<String, PathBuf>>>,
     pub session_counter: Arc<AtomicUsize>,
     /// Tracks sessions that have already emitted recipe telemetry to prevent double counting.
     recipe_session_tracker: Arc<Mutex<HashSet<String>>>,
+    /// Tracks cancellation tokens for active sessions
+    pub(crate) session_cancellation_tokens: Arc<Mutex<HashMap<String, CancellationToken>>>,
```
DOsinga, line 19: this seems a bit confusing; do we want to keep track of the active sessions in the app state or in the session manager?

crates/goose-server/src/routes/reply.rs
```diff
@@ -455,13 +492,57 @@ pub async fn confirm_permission(
...
+    responses(
+        (status = 200, description = "Agent task cancelled successfully"),
+        (status = 404, description = "Session not found"),
+        (status = 500, description = "Internal server error")
+    )
+)]
+pub async fn cancel_agent(
+    State(state): State<Arc<AppState>>,
+    Json(request): Json<CancelRequest>,
+) -> Result<Json<Value>, StatusCode> {
```
DOsinga, line 513: dont return untyped json

crates/goose-server/src/routes/reply.rs
```diff
@@ -455,13 +492,57 @@ pub async fn confirm_permission(
...
+}
+
 pub fn routes(state: Arc<AppState>) -> Router {
     Router::new()
         .route(
             "/reply",
             post(reply).layer(DefaultBodyLimit::max(50 * 1024 * 1024)),
         )
         .route("/confirm", post(confirm_permission))
+        .route("/cancel", post(cancel_agent))
```
DOsinga, line 545: I don't see this being registered in openapi?

crates/goose/src/session/session_manager.rs
```diff
@@ -241,6 +247,102 @@ impl SessionManager {
             Ok(())
         }
     }
+
+    pub async fn mark_in_use(id: &str, in_use: bool) -> Result<()> {
+        Self::instance().await?.mark_in_use(id, in_use).await
```
DOsinga, line 252: I don't think we want to do this using the session manager, but this should go through the builder pattern

crates/goose/src/session/session_manager.rs
```diff
@@ -241,6 +247,102 @@ impl SessionManager {
...
+    }
+
+    pub async fn heartbeat(id: &str) -> Result<()> {
+        Self::instance().await?.heartbeat(id).await
```
DOsinga, line 256: and then you don't need the heartbeat

crates/goose/src/session/session_manager.rs
```diff
@@ -241,6 +247,102 @@ impl SessionManager {
...
+
+    pub async fn stream_updates(
+        session_id: String,
+    ) -> Result<Pin<Box<dyn Stream<Item = Result<Session>> + Send>>> {
+        // Verify session exists first
+        let initial_session = Self::get_session(&session_id, true).await?;
+
+        let stream = async_stream::stream! {
+            // Check if session is stale using STALE_LOCK_MINUTES threshold
+            let is_actually_in_use = Self::is_session_in_use(&session_id, STALE_LOCK_MINUTES).await.unwrap_or(false);
```
DOsinga, line 274: I think you want to bubble up, no?

crates/goose/src/session/session_manager.rs
```diff
@@ -241,6 +247,102 @@ impl SessionManager {
...
+
+            // Session is in use, continue streaming until it's no longer in use
+            let mut interval = tokio::time::interval(Duration::from_secs(1));
+            let mut last_message_count: Option<usize> = Some(session_to_yield.message_count);
+            let mut last_updated_at: Option<String> = Some(session_to_yield.updated_at.to_rfc3339());
+
+            loop {
+                interval.tick().await;
+
+                match Self::get_session(&session_id, true).await {
```
DOsinga, line 305: you're hammering the database here only to see if it got updated

ui/desktop/src/types/message.ts
```diff
@@ -97,12 +100,18 @@ export function getTextContent(message: Message): string {
 }
 
 export function getToolRequests(message: Message): (ToolRequest & { type: 'toolRequest' })[] {
+  if (!message.content) {
+    return [];
+  }
```
DOsinga, line 105: why do we need these changes?

ui/desktop/src/hooks/useChatEngine.ts
```diff
@@ -282,6 +282,22 @@ export const useChatEngine = ({
     setLastInteractionTime(Date.now());
     stopPowerSaveBlocker();
 
+    // In ALPHA mode, when user explicitly stops, we need to cancel the agent task
+    // This is different from a page refresh where we want the agent to continue
+    if (process.env.ALPHA === 'true' && chat.sessionId) {
+      fetch(getApiUrl('/cancel'), {
```
DOsinga, line 288: don't call the api like this. does this even work without injecting the secret?

ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -43,15 +43,30 @@ export function useChatStream({
...
+  // Keep refs in sync with props
+  messagesRef.current = messages;
+  sessionIdRef.current = sessionId;
+
   const handleSubmit = useCallback(
     async (userMessage: string) => {
+      // Abort any existing stream
+      if (abortControllerRef.current) {
+        abortControllerRef.current.abort();
+        abortControllerRef.current = null;
```
DOsinga, line 59: don't we want to throw an error here instead? or really you shouldn't be able to get here

ui/desktop/src/contexts/SessionStreamContext.tsx
```diff
@@ -0,0 +1,362 @@
...
+
+const TextDecoder = globalThis.TextDecoder;
+
+/**
+ * Event types from the session SSE stream
+ */
+type SessionEvent =
+  | { type: 'Session'; session: Session }
+  | { type: 'Error'; error: string }
+  | { type: 'Ping' };
```
DOsinga, line 21: this type mimics what we declare on the server. we should use the server one. also this has an extra Ping one.

ui/desktop/src/components/ChatInput.tsx
```diff
@@ -1423,19 +1431,21 @@ export default function ChatInput({
...
+                        ? 'Compacting conversation...'
+                        : isAnyImageLoading
+                          ? 'Waiting for images to save...'
+                          : isAnyDroppedFileLoading
+                            ? 'Processing dropped files...'
+                            : isRecording
+                              ? 'Recording...'
+                              : isTranscribing
+                                ? 'Transcribing...'
+                                : (chatContext?.agentWaitingMessage ?? 'Send')}
```
DOsinga, line 1448: we need to refactor this

zanesq:
ok makes sense, closing

PR #5124, opened 2025-10-10
Initial draft of context platorm extension (do not review)

### LINKED ISSUE #5104: Context files (.goosehints, AGENTS.md) should be loaded without requiring developer extension
## Problem

Currently, `.goosehints` and `AGENTS.md` context files are only loaded when the developer extension is explicitly enabled with `--with-builtin developer`. This creates an unexpected dependency where basic context loading requires a specific extension to be active.

## Current Behavior

- Context files like `.goosehints` and `AGENTS.md` are not loaded in non-interactive runs unless `--with-builtin developer` is specified
- The hint loading functionality is implemented within the developer extension's `get_info()` method
- Users expect these context files to be available regardless of which extensions are enabled

## Expected Behavior

Context files should be loaded as part of core Goose functionality, independent of any specific extension being enabled. These files provide important project-specific context that should be available in all modes of operation.

## Technical Details

The `load_hint_files` function in `crates/goose-mcp/src/developer/goose_hints/load_hints.rs` is currently called only from within the developer extension. This functionality should be moved to core Goose initialization to ensure context files are always available when present in the project directory.

## Impact

This affects users who run Goose in non-interactive modes or with minimal extension configurations, as they lose access to important project context that should be universally available.

tlongwell-block:
Context files (`.goosehints`, `AGENTS.md`) now load automatically without requiring the developer extension to be explicitly enabled. This functionality has been moved to a dedicated platform extension that's enabled by default.

### Why
Previously, users had to run goose with `--with-builtin developer` to access their context files, even when they didn't need developer tools. This created an unnecessary dependency between basic context loading and the developer extension.

### How
- Created a new `context` platform extension that handles hint file loading
- Moved the hint loading logic from `goose-mcp` to `goose` core
- Set the context extension to be enabled by default
- Removed hint loading responsibility from the developer extension

### Testing
- All 20 existing hint loading tests pass
- Verified context files load without developer extension
- Confirmed backward compatibility when developer extension is enabled

Fixes #5104


DOsinga:
closing because stale

Issue #5110, opened 2025-10-10
Installation of recipe or extension from deeplink repeats on new window/chat


zanesq:
**Describe the bug**
Run an installation of recipe or extension from deeplink and the install prompt shows like normal but after refresh or a new chat it shows again.

**To Reproduce**
1. Install recipe or extension from deeplink
2. Start a new chat/window
3. Install screen shows again even though we already installed it.

**Expected behavior**
After install process (success or failure) It shouldn't show the installation screen again in a new window. Deeplink should be cleared after use.

**Please provide following information:**
 - **OS & Arch:** Mac Arm
 - **Interface:** UI
 - **Version:** > v1.9.1 
 - **Extensions enabled:** Developer
 - **Provider & Model:** Claude

**Additional context**
This is a pre-existing issue but more prominent now in 1.10.0 because it requires users to start a new chat to enable a newly installed extension.

PR #5128, opened 2025-10-10
Clear deeplinks after use

zanesq:
## Summary
Fixes https://github.com/block/goose/issues/5110

Also noticed the new recipe warning wasn't showing, there was a race condition happening when a recipe was loaded, finalRecipe would be set before the recipe prop arrived. Yes, the `useRecipeManager` code is a mess, we plan on refactoring a lot of this in the upcoming basecamp2 rework.

NOTE: this fixes the installation dialog from showing again in a new chat/window which solves the immediate issue with launching a new chat after installing an extension but there is another semi related [issue](https://github.com/block/goose/issues/5130) if you refresh the same window after installation it will show the modal again.

### DIFF
--- a/ui/desktop/src/hooks/useRecipeManager.ts
+++ b/ui/desktop/src/hooks/useRecipeManager.ts
@@ -83,6 +83,8 @@ export const useRecipeManager = (chat: ChatType, recipe?: Recipe | null) => {
       }
 
       if (finalRecipe) {
+        hasCheckedRecipeRef.current = true;
+
         // If the recipe comes from session metadata (not from navigation state),
         // it means it was already accepted in a previous session, so auto-accept it
         const hasMessages = chat.messages.length > 0;
@@ -92,11 +94,8 @@ export const useRecipeManager = (chat: ChatType, recipe?: Recipe | null) => {
           // Recipe loaded from session metadata should be automatically accepted
           setRecipeAccepted(true);
           setIsRecipeWarningModalOpen(false);
-          hasCheckedRecipeRef.current = true;
           return;
         }
-
-        hasCheckedRecipeRef.current = true;
 
         try {
           const hasAccepted = await window.electron.hasAcceptedRecipeBefore(finalRecipe);


ui/desktop/src/hooks/useRecipeManager.ts
```diff
@@ -75,17 +76,29 @@ export const useRecipeManager = (chat: ChatType, recipe?: Recipe | null) => {
 
   useEffect(() => {
     const checkRecipeAcceptance = async () => {
+      // Create a unique key for this recipe to prevent duplicate checks
+      const recipeKey = finalRecipe ? `${finalRecipe.title}-${finalRecipe.instructions}` : null;
```
alexhancock: Would this need to include all aspects of the recipe? Like extensions included as well?

zanesq: good point at first I didn't think it was necessary but now I think it should behave the same as the other checks, will update

zanesq: ended up removing these checks altogether, wasn't necessary since we can just check if we've checked the recipe or not in the same chat session

zanesq:
this should be gtg, verified and tested locally with a few recipes (old and new)

ui/desktop/src/hooks/useRecipeManager.ts
```diff
@@ -75,17 +77,27 @@ export const useRecipeManager = (chat: ChatType, recipe?: Recipe | null) => {
...
         // it means it was already accepted in a previous session, so auto-accept it
-        const isFromSessionMetadata = !recipe && finalRecipe;
+        const hasMessages = chat.messages.length > 0;
+        const isFromSessionMetadata = !recipe && finalRecipe && hasMessages;
 
         if (isFromSessionMetadata) {
           // Recipe loaded from session metadata should be automatically accepted
           setRecipeAccepted(true);
+          setIsRecipeWarningModalOpen(false);
+          hasCheckedRecipeRef.current = true;
```
alexhancock: should we just move this to happen one time immediately after the early return check?

zanesq: Thanks makes sense, needed to put it after the check for the recipe so we handle the no recipe case.. the real issue here is its re-rendering multiple times so there is a race condition before the recipe is ready. This is a hack around that, will come back to that multiple rendering issue when we refactor this soon.

alexhancock: ok got it

Issue #5140, opened 2025-10-12
Sub-agents should not run in non autonomous mode


cruftex:
**Describe the bug**

Goose is somehow starting a subagent. Where is the output of it? What did it do?

**To Reproduce**

See https://github.com/block/goose/issues/5139

**Expected behavior**

I get a response what was done

In the session jsonl file there is one line with 75440 chars. It looks like there is some info about what happened.

**Screenshots**

<img width="525" height="375" alt="Image" src="https://github.com/user-attachments/assets/496a3f70-cf29-430e-9de1-7bccb0fed167" />

**Please provide following information:**
 - **OS & Arch:** Ubuntu 24.04
 - **Interface:** cli
 - **Version:**  1.9.3
 - **Extensions enabled:** Developer Extension
 - **Provider & Model:** local

DOsinga:
thanks for the report. I agree, sub agents should not run when you are not in autonomous mode

cruftex:
@DOsinga 

Thanks for checking on this!

The mode was in APPROVE mode on the CLI. Here is the config:

````
GOOSE_MODE: approve
GOOSE_MODEL: qwen/qwen3-coder-30b
OPENAI_HOST: http://localhost:1234
extensions:
  computercontroller:
    available_tools: []
    bundled: true
    description: null
    display_name: Computer Controller
    enabled: true
    name: computercontroller
    timeout: 300
    type: builtin
  developer:
    available_tools: []
    bundled: true
    description: null
    display_name: Developer Tools
    enabled: true
    name: developer
    timeout: 300
    type: builtin
OPENAI_API_KEY: ignore
GOOSE_PROVIDER: openai
````

All together I see the following problems with this, which should be addressed:

- There was no plan displayed what the subagent should do
- I only could approve whether the subagent runs or not without any information
- The outcome of the subagent was not displayed
- I don't understand why one (as in 1) agent is started, instead of staying in the main flow

tlongwell-block:
> thanks for the report. I agree, sub agents should not run when you are not in autonomous mode

This is done in https://github.com/block/goose/pull/5149

The conversion of the subagent tools to platform extensions should be done, but this will prevent unwanted behavior until that major revision is complete

PR #5149, opened 2025-10-13
Minimally disable subagents when not in autonomous model

tlongwell-block:
Fixes https://github.com/block/goose/issues/5140

Minimal change required. This will change once the create dynamic task and subagent execute tools are converted to platform extensions.

### DIFF
--- a/crates/goose-cli/src/commands/configure.rs
+++ b/crates/goose-cli/src/commands/configure.rs
@@ -423,7 +423,7 @@ pub async fn configure_provider_dialog() -> Result<bool, Box<dyn Error>> {
     let mut available_providers = providers().await;
 
     // Sort providers alphabetically by display name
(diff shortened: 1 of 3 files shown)


crates/goose/src/agents/reply_parts.rs
```diff
@@ -34,6 +34,11 @@ async fn toolshim_postprocess(
 impl Agent {
     /// Prepares tools and system prompt for a provider request
     pub async fn prepare_tools_and_prompt(&self) -> anyhow::Result<(Vec<Tool>, Vec<Tool>, String)> {
+        // Determine the current mode from config
+        let config = crate::config::Config::global();
+        let mode = config.get_param("GOOSE_MODE").unwrap_or("auto".to_string());
+        let is_autonomous = mode == "auto";
```
DOsinga: this doesn't need the comment or the intermediate variable and move the declaration to where it is used

PR #5159, opened 2025-10-14
remove extra whitespace from message

jamadeo:
Some providers actually raise 400 for this

### DIFF
--- a/crates/goose/src/agents/context.rs
+++ b/crates/goose/src/agents/context.rs
@@ -110,11 +110,13 @@ impl Agent {
         final_token_counts.push(summary_tokens);
 
         // Add an assistant message to continue the conversation (agent_visible=true, user_visible=false)
-        let assistant_message = Message::assistant().with_text("
-            The previous message contains a summary that was prepared because a context limit was reached.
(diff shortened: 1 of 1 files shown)


crates/goose/src/agents/context.rs
```diff
@@ -110,11 +110,13 @@ impl Agent {
...
-            The previous message contains a summary that was prepared because a context limit was reached.
-            Do not mention that you read a summary or that conversation summarization occurred
-            Just continue the conversation naturally based on the summarized context
-        ").with_metadata(MessageMetadata::agent_only());
+        let assistant_message = Message::assistant()
+            .with_text(
+                "The previous message contains a summary that was prepared because a context limit was reached.
+Do not mention that you read a summary or that conversation summarization occurred
+Just continue the conversation naturally based on the summarized context"
+            )
```
DOsinga, line 118: how did this work before?

jamadeo: I don't know, this code has apparently been here for a while. Which provider did you see the 400 with?

jamadeo: maybe we had a .trim() upstream somewhere that got removed?

PR #5160, opened 2025-10-14
chore: include vendored node executable

alexhancock:
* Makes it so `node` is also bundled/vendored like `npx`
* Same setup content as the `npx` previously had, but deduped between that script and the new `node`
* Tested `npx` and `node` based extensions

### DIFF
--- a/ui/desktop/src/bin/node-setup-common.sh
+++ b/ui/desktop/src/bin/node-setup-common.sh
@@ -1,0 +1,101 @@
+#!/bin/bash
+
+# Common setup script for node and npx
+# This script sets up hermit and node.js environment
+
+# Enable strict mode to exit on errors and unset variables
+set -euo pipefail
+
+# Set log file
+LOG_FILE="/tmp/mcp.log"
+
+# Clear the log file at the start
+> "$LOG_FILE"
+
+# Function for logging
+log() {
+    local MESSAGE="$1"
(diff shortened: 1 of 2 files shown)


ui/desktop/src/bin/node-setup-common.sh
```diff
@@ -0,0 +1,101 @@
...
+
+log "Starting node setup (common)."
+
+# Ensure ~/.config/goose/mcp-hermit/bin exists
+log "Creating directory ~/.config/goose/mcp-hermit/bin if it does not exist."
+mkdir -p ~/.config/goose/mcp-hermit/bin
+
+# Change to the ~/.config/goose/mcp-hermit directory
+log "Changing to directory ~/.config/goose/mcp-hermit."
+cd ~/.config/goose/mcp-hermit
```
jamadeo, line 32: fwiw I think a lot of this is also in the uvx script (the hermit part), could eventually do a bit more deduplicating

alexhancock:
> Does it not need to be added here too?
> 
> https://github.com/block/goose/blob/5581ca33bfddbeccc2b1acc54cac94be68c95315/ui/desktop/src/components/settings/extensions/utils.ts#L193

Investigated and found that this adds the whole shims dir to the path: https://github.com/block/goose/blob/73e9e1af2e716d5730dac2ef1781f12ee86a2d64/ui/desktop/src/goosed.ts#L162-L163

We need to do a larger cleanup where we remove the concept of `replaceWithShims`, and just have the shells which run extensions include the hermit directory in the PATH, but this will be a followup

PR #5165, opened 2025-10-14
Propagate session ID in LLM and MCP requests

codefromthecrypt:
## Summary

Adds `goose-session-id` header to HTTP requests and `_meta.goose-session-id` to MCP operations when a session exists. Enables external observability tools like Envoy AI Gateway to correlate requests by session without requiring OpenTelemetry in Goose.

Implementation:
- Task-local storage for session context propagation
- HTTP header injection in api_client (case-insensitive via reqwest)
- MCP metadata injection preserving existing _meta fields (case-insensitive cleanup)
- Integration and unit tests for both HTTP and MCP propagation

Example use case: Configure Envoy AI Gateway with `OTEL_AIGW_SPAN_REQUEST_HEADER_ATTRIBUTES=goose-session-id:session.id` to send traces to Arize Phoenix grouped by session.

### Type of Change
- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing

added tests including an integrated one

### Related Issues

Relates to https://github.com/envoyproxy/ai-gateway/issues/1221 which asked for this in other agent frameworks

### DIFF
--- a/crates/goose/tests/session_id_propagation_test.rs
+++ b/crates/goose/tests/session_id_propagation_test.rs
@@ -81,7 +81,7 @@ async fn setup_mock_server() -> (MockServer, HeaderCapture, Box<dyn Provider>) {
     (mock_server, capture, provider)
 }
 
-async fn make_request(provider: &Box<dyn Provider>, session_id: Option<&str>) {
+async fn make_request(provider: &dyn Provider, session_id: Option<&str>) {
     let message = Message::user().with_text("test message");
     let request_fn = async {
         provider
@@ -104,7 +104,7 @@ async fn make_request(provider: &Box<dyn Provider>, session_id: Option<&str>) {
 async fn test_session_id_propagation_to_llm() {
     let (_, capture, provider) = setup_mock_server().await;
 
-    make_request(&provider, Some("integration-test-session-123")).await;
+    make_request(provider.as_ref(), Some("integration-test-session-123")).await;
 
     assert_eq!(
         capture.get_captured(),
@@ -116,7 +116,7 @@ async fn test_session_id_propagation_to_llm() {
 async fn test_no_session_id_when_absent() {
     let (_, capture, provider) = setup_mock_server().await;
 
-    make_request(&provider, None).await;
+    make_request(provider.as_ref(), None).await;
 
     assert_eq!(capture.get_captured(), vec![None]);
 }
@@ -126,9 +126,9 @@ async fn test_session_id_matches_across_calls() {
     let (_, capture, provider) = setup_mock_server().await;
 
     let test_session_id = "consistent-session-456";
-    make_request(&provider, Some(test_session_id)).await;
-    make_request(&provider, Some(test_session_id)).await;
-    make_request(&provider, Some(test_session_id)).await;
+    make_request(provider.as_ref(), Some(test_session_id)).await;
+    make_request(provider.as_ref(), Some(test_session_id)).await;
+    make_request(provider.as_ref(), Some(test_session_id)).await;
 
     assert_eq!(
         capture.get_captured(),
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
very flexible on this one, too, header naming.. whether it is configurable.. all that!

crates/goose/tests/session_id_propagation_test.rs
```diff
@@ -0,0 +1,360 @@
...
+                    "completion_tokens": 5,
+                    "total_tokens": 15
+                }
+            }))
+        })
+        .mount(&mock_server)
+        .await;
+
+    // Configure OpenAI provider to use mock server
+    std::env::set_var("OPENAI_HOST", mock_server.uri());
```
jamadeo: can we do this without environment vars? I think it should be possible. Maybe OpenAIProvider::from_custom_config.

that way you can probably avoid the `#[serial]` on the test

crates/goose/src/agents/mcp_client.rs
```diff
@@ -389,3 +389,184 @@ impl McpClientTrait for McpClient {
...
+            .map(|meta| meta.0.clone())
+            .unwrap_or_default();
+
+        // Remove any existing session ID keys with different casings (case-insensitive)
+        // Note: JsonObject (serde_json::Map) is case-sensitive, so we use retain to filter
+        meta_map.retain(|k, _| !k.eq_ignore_ascii_case("goose-session-id"));
+
+        // Insert with canonical casing
+        meta_map.insert(
+            "goose-session-id".to_string(),
```
jamadeo: I think it's worth defining this as a const somewhere `const GOOSE_SESSION_ID = ...`

crates/goose/src/providers/api_client.rs
```diff
@@ -369,6 +369,11 @@ impl<'a> ApiRequestBuilder<'a> {
         let mut request = request_builder(url, &self.client.client);
         request = request.headers(self.headers.clone());
 
+        // Inject session ID header if available
```
jamadeo: try to avoid redundant comments like this (llms love to create them though :)) even if they're already all over the goose codebase

jamadeo (approved):
seems very useful! if you could have a quick look at any redundant comments and pare them down that would be great. also if possible avoid using environment variables in tests, I think this should be possible in your case

codefromthecrypt:
thanks for the advice! I was AWOL but coming online again now. will touch this up tomorrow UTC+8

crates/goose/src/providers/api_client.rs
codefromthecrypt: no problemo, and haha busted

crates/goose/tests/session_id_propagation_test.rs
codefromthecrypt: thanks for the tip. from_custom_config is close, but it still errs out if the api key env isn't there. I'm spiking new() that's hidden meanwhile and we can iterate if that's not a good idea

codefromthecrypt:
ok rebased and also tried to shred the lines in a good way. if I made things worse lemme know

crates/goose/src/providers/mod.rs
```diff
@@ -1,5 +1,5 @@
 pub mod anthropic;
-mod api_client;
+pub mod api_client;
```
codefromthecrypt, line 2: needed to expose this for tests in order to write them without ENV dependencies

codefromthecrypt:
cool I think the live provider red X is due to it not being on a protected branch. probably safe to review!

jamadeo (approved):
Looks great, and yeah, those provider tests aren't configured correctly right now. (they need access to secrets for api keys, but those aren't available on forks)

michaelneale:
over to you @jamadeo

PR #5166, opened 2025-10-14
chore: clean up random unused files

alexhancock:
Removing random unused files which seem like they were added unintentionally


recipe-inline-python.yaml
```diff
@@ -1,87 +0,0 @@
-version: "1.0.0"
```
alexhancock, line 1: @DOsinga this one I wasn't sure about. Do we want to keep it in?

DOsinga: can go

Issue #5167, opened 2025-10-14
OS: unbound variable installing Goose in GitHub Action runner after release of v.1.10.0


Dan-Wuensch:
**Describe the bug**

Note: Please check the common issues on https://block.github.io/goose/docs/troubleshooting before filing a report

**To Reproduce**
* Install goose via `curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh | CONFIGURE=false INSTALL_PATH=/home/runner/.local/bin bash` in a GitHub Action

**Expected behavior**
Latest version installs and runs in the container

**Screenshots**

**Please provide following information:**
 - **OS & Arch:** Ubuntu 24.04.3 LTS, GitHub runner version: '2.328.0'
 - **Interface:** CLI, CI pipeline approach
 - **Version:** v.1.10.0 
 - **Extensions enabled:** Developer
 - **Provider & Model:** AWS Bedrock (choice of provider is irrelevant to this issue, however)

**Additional context**
As soon as v.1.10.0 became available, our CI pipeline started failing on every run with the following error installing goose 🪿 :

"bash: line 40: OS: unbound variable"

Our workaround is by modifying the install script locally to download last known working version 1.9.1 via [goose-x86_64-unknown-linux-gnu.tar.bz2](https://github.com/block/goose/releases/download/v1.9.1/goose-x86_64-unknown-linux-gnu.tar.bz2)

Longer term, having standard practice be pinning a version instead of auto-updates will work better for CI/CD sers, since Goose installs fresh on every run

PR #4928, opened 2025-10-01
fix: improve Windows OS detection in CLI installation script

aegntic:
- Enhanced Windows detection for Git Bash, MSYS2, Cygwin, and WSL environments
- Added multiple detection methods: WINDIR, OSTYPE, /proc/version, mount points, Windows executables
- Fixed issue where Git Bash on Windows incorrectly detected as Linux
- Added debug output to help troubleshoot OS detection issues
- Improved extraction tool validation based on detected OS
- Fixes #4848 where CLI installation failed due to downloading Linux tar.bz2 instead of Windows zip

Closes #4848

## Pull Request Description

---

**Email**:

PR #5169, opened 2025-10-14
fix: adjust download_cli.sh to tolerate no OS variable

alexhancock:
Fixing a bug that came with the changes in https://github.com/block/goose/pull/4928 which blocks `goose update` and usage of the `download_cli.sh` script

Resolves https://github.com/block/goose/issues/5167


Dan-Wuensch:
Thank you!

PR #5175, opened 2025-10-15
feat: goose native gguf model serving: an embedded model serving infrastructure

michaelneale:
* Aim:  Just Work with off the shelf gguf models*, no ollama or lmstudio etc needed.

This uses `llama-server` (which is automatically downloaded if needed, and then platform optimised) and can run any .gguf model from `~/.models` dir. It tries to detect if the model can do tool calling, and if not, will do a fallback emulation (assuming not in chat only mode). 

image

Screenshot 2025-10-15 at 3 05 25 pm

For example: 

```
GOOSE_PROVIDER=embedded GOOSE_MODEL=qwen2.5-7b-instruct-q3_k_m.gguf ./target/release/goose session
```

or if you want it to force using (very fast) tool call emulation: 

```
EMBEDDED_FORCE_TOOL_EMULATION=true GOOSE_PROVIDER=embedded GOOSE_MODEL=qwen2.5-7b-instruct-q3_k_m.gguf ./target/release/goose session
```

try this out with: 

qwen2.5-7b-instruct-q3_k_m.gguf
gemma-3n-E4B-it-Q6_K.gguf
gpt-oss-20b-Q3_K_M.gguf

for different sizes. 

This will suggest some models, and any models < 10G in size it will automatically use tool emulation if not in chat mode (> 10G it will test if it claims it can do tool calling and use, eg gpt-oss:20b will use native tool calling as large enough).

It also, out of the box, will offer to download a qwen 2.5 and gpt oss 20b model (small and larger) which are tested to work (for some value of work!) in the background. 

image


iandouglas:
would the user be able to point to their own llama.cpp with this (or search the system path for an existing llama.cpp)? ie, if I'm on AMD hardware, I might want to use a compiled llama.cpp with ROCm support, versus a generic llama.cpp build that might use my CPU instead of GPU. If I have llama.cpp hosted on my network, i'd love to do something similar to OLLAMA_HOST but maybe this would be LLAMACPP_HOST maybe?

michaelneale:
@iandouglas this PR is not for that - you can already do that, with the custom providers, you can point to anything now. In this case you provider the llama-server and off you go, doesn't seem related to this (the whole point is to make it turn key for places where it may not work)

(I think this does, or did, look for a llama-server on PATH, and use if it it can, in which case it would pick up that installed/compiled version), llama-server runs llama.cpp (this branch doesn't run llama.cpp in process -I have another one that does that but not sure about it). so maybe that answers it? (but if you are aware enough to have a better llama installed, then you would know how to use it and probably expose it via your own tools or lmstudio?)

iandouglas:
@michaelneale question about functionality ... would it be helpful, or would it be over-stepping, to examine the user's RAM and downloading a model small enough to fit? gpt-oss-20b won't fit on many machines if they only have 16GB of RAM and running apps/Chrome/etc without hitting swap. Maybe defer to smaller models by default?

michaelneale:
@iandouglas yeah was thinking that - going to make some suggested ones - a small one and a larger one and can suggest accordingly, but people can be very unaware of their platform (and in that case - not sure how we fail - just say you don't have enough power to use this provider? also not sure how to detect GPU vs CPU etc, that I don't think is solved that I am aware of). Could detect apple silicon capability, but anything else I actually am not sure how to.

iandouglas:
>  not sure how to detect GPU vs CPU

Yeah, we definitely don't want to start tracking a database of CPUs and GPUs ourselves. :) I do see Rusy library called `hardware-query` that will give you hardware info, including GPU detection and capabilities like CUDA, ROCm, DirectML, etc., but sounds like a lot of extra logic on our side to evaluate.

michaelneale:
just got some final touches to make it work in the GUI nicely

crates/goose/src/providers/embedded.rs
```diff
@@ -0,0 +1,1454 @@
...
+                    match Command::new("sh").arg("-c").arg(command).output() {
+                        Ok(output) => {
+                            let stdout = String::from_utf8_lossy(&output.stdout);
+                            let stderr = String::from_utf8_lossy(&output.stderr);
+
+                            // Truncate if output is too large
+                            let truncate_if_large = |s: &str| -> String {
+                                if s.len() > MAX_TOOL_OUTPUT_SIZE {
+                                    // Use char_indices to find a safe truncation point
+                                    let truncate_at = s
```
DOsinga: we have safe_truncate

crates/goose/src/providers/embedded.rs
```diff
@@ -0,0 +1,1752 @@
...
+            std::io::copy(&mut file, &mut outfile)?;
+
+            // Preserve permissions on Unix
+            #[cfg(unix)]
+            {
+                use std::os::unix::fs::PermissionsExt;
+                if let Some(mode) = file.unix_mode() {
+                    std::fs::set_permissions(&outpath, std::fs::Permissions::from_mode(mode))?;
+                }
+            }
```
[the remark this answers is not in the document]
michaelneale: lol

crates/goose-server/src/routes/utils.rs
```diff
@@ -94,6 +94,11 @@ pub fn inspect_keys(
 pub fn check_provider_configured(metadata: &ProviderMetadata, provider_type: ProviderType) -> bool {
     let config = Config::global();
 
+    // Special case: Embedded provider is always available (local GGUF models)
```
michaelneale, line 97: can't think of a better way to do this (pretty sure need to have this??)

crates/goose/Cargo.toml
```diff
@@ -100,19 +100,20 @@ dashmap = "6.1"
...
 
 oauth2 = "5.0.0"
+sysinfo = "0.37.2"
+zip = "0.6"
 schemars = { version = "1.0.4", default-features = false, features = ["derive"] }
 insta = "1.43.2"
 paste = "1.0.0"
 shellexpand = "3.1.1"
 indexmap = "2.12.0"
+libc = "0.2.177"
```
michaelneale, line 114: so we can manage llama-server

crates/goose/Cargo.toml
```diff
@@ -100,19 +100,20 @@ dashmap = "6.1"
 ahash = "0.8"
 tokio-util = "0.7.15"
 unicode-normalization = "0.1"
-zip = "0.6"
 sys-info = "0.9"
 
 oauth2 = "5.0.0"
+sysinfo = "0.37.2"
+zip = "0.6"
```
michaelneale, line 107: needed to unpack and handle gguf files and match it up with OS capabilities

zanesq:
checking if this is still in progress?

alexhancock:
@zanesq I want to pick this one up but it will likely come after I finish sandbox/code mode MCP

tlongwell-block:
@michaelneale did you look at this (I think) pure rust inference library? https://github.com/EricLBuehler/mistral.rs

I had some good luck playing with it in goose

DOsinga:
let's discuss later today where we want to take this if anywhere? I do like this, but this has obviously gone stale quite a bit

michaelneale:
bringing this back for reference only

PR #5177, opened 2025-10-15
feat: enhance goose to search sessions for easy recall

michaelneale:
They say an elephant never forgets - but goose does!

A platform tool to search session history (smartly) now we have sqlite which is fast and right there. 
Will do a quasi semantic fuzzy search (basically creative words and LIKE statement!) to find relevant sessions, dig things out, and you can also have it poke at session to see more detail if it needs to (or it will automatically). 
Screenshot 2025-10-15 at 6 43 41 pm

Screenshot 2025-10-15 at 6 44 07 pm

### DIFF
--- a/crates/goose/src/session/session_manager.rs
+++ b/crates/goose/src/session/session_manager.rs
@@ -71,29 +71,6 @@ pub struct SessionInsights {
 
 pub type SessionId = String;
 
-#[derive(Debug, Clone, Serialize)]
-pub struct ChatRecallResult {
-    pub session_id: String,
-    pub session_description: String,
-    pub session_working_dir: String,
-    pub last_activity: DateTime<Utc>,
-    pub total_messages_in_session: usize,
-    pub messages: Vec<ChatRecallMessage>,
-}
-
-#[derive(Debug, Clone, Serialize)]
-pub struct ChatRecallMessage {
-    pub role: String,
-    pub content: String,
-    pub timestamp: DateTime<Utc>,
-}
-
-#[derive(Debug, Serialize)]
-pub struct ChatRecallResults {
-    pub results: Vec<ChatRecallResult>,
-    pub total_matches: usize,
-}
-
 impl SessionUpdateBuilder {
     fn new(session_id: String) -> Self {
         Self {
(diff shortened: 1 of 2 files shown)


crates/goose/src/session/session_manager.rs
```diff
@@ -241,6 +266,19 @@ impl SessionManager {
...
+        exclude_session_id: Option<String>,
+    ) -> Result<ChatRecallResults> {
+        Self::instance()
+            .await?
+            .search_chat_history(query, limit, after_date, before_date, exclude_session_id)
+            .await
+    }
 }
 
 pub struct SessionStorage {
```
DOsinga, line 261: this is a lot of lines, maybe break it up a bit. in fact it could probably live in its own file

crates/goose/src/agents/chat_recall_extension.rs
```diff
@@ -0,0 +1,347 @@
...
+        vec![Tool::new(
+            "chat_recall".to_string(),
+            indoc! {r#"
+                Search past chat or load session summaries. Use when it is clear user expects some memory or context.
+
+                search mode (query): Use multiple keywords/synonyms. Returns messages grouped by session, ordered by recency. Supports date filters.
+                load mode (session_id): Returns first/last 3 messages of a session.
+            "#}
+            .to_string(),
+            object!({
```
DOsinga: I know we do this in todo, but we shouldn't! handrolling of json schema is dangerous and will weirdly crash odd providers. Can you use the #[tool( macro instead /cc @alexhancock

crates/goose/src/session/session_manager.rs
michaelneale: yep, will, I felt bad about adding that exception for clippy.

crates/goose/src/agents/chat_recall_extension.rs
michaelneale: oh! I didn't know we could do this in non mcp-launched extensions (ie in process) - neat if so!

michaelneale: hrm looked at this, it isn't really obvious how to use that macro and rmcp lib outside of running an mcp server, really designed to run out of process...

michaelneale: did switch approaches so it is less hoppefully problematic

DOsinga:
Not sure if you noticed, but this was referred since it showed errors for a bunch of people

PR #5186, opened 2025-10-15
Compaction overhaul

DOsinga:


crates/goose/src/agents/subagent_handler.rs
```diff
@@ -141,22 +141,23 @@ fn get_agent_messages(
...
             match message_result {
-                Ok(AgentEvent::Message(msg)) => session_messages.push(msg),
-                Ok(AgentEvent::McpNotification(_))
-                | Ok(AgentEvent::ModelChange { .. })
-                | Ok(AgentEvent::HistoryReplaced(_)) => {}
+                Ok(AgentEvent::Message(msg)) => conversation.push(msg),
+                Ok(AgentEvent::McpNotification(_)) | Ok(AgentEvent::ModelChange { .. }) => {}
+                Ok(AgentEvent::HistoryReplaced(updated_conversation)) => {
+                    conversation = updated_conversation;
+                }
```
jamadeo, line 153: does this mean sub-agents never compacted?

crates/goose/src/context_mgmt/mod.rs
```diff
@@ -1,6 +1,338 @@
...
+///   - `Vec<usize>`: Indices of removed messages (empty if no compaction)
+///   - `Option<ProviderUsage>`: Provider usage from summarization (if compaction occurred)
+pub async fn check_and_compact_messages(
+    agent: &Agent,
+    messages_with_user_message: &[Message],
+    force_compact: bool,
+    preserve_last_user_message: bool,
+    threshold_override: Option<f64>,
+    session_metadata: Option<&crate::session::Session>,
+) -> std::result::Result<(bool, Conversation, Vec<usize>, Option<ProviderUsage>), anyhow::Error> {
```
jamadeo, line 76: should this return AutoCompactionResult?

crates/goose/src/context_mgmt/mod.rs
```diff
@@ -1,6 +1,338 @@
...
+    // Set role to user as it will be used in following conversation as user content
+    response.role = Role::User;
+
+    // Ensure we have token counts, estimating if necessary
+    provider_usage
+        .ensure_tokens(&system_prompt, &summarization_request, &response, &[])
+        .await
+        .map_err(|e| anyhow::anyhow!("Failed to ensure usage tokens: {}", e))?;
+
+    std::prelude::rust_2015::Ok(Some((response, provider_usage)))
```
jamadeo, line 337: strange qualifier here

crates/goose-server/src/routes/context.rs
```diff
@@ -48,28 +46,20 @@ async fn manage_context(
...
+        &agent,
+        &request.messages,
+        true,
+        false,
+        None,
+        None,
+    )
+    .await
+    .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+    // TODO(Douwe): store into db
```
jamadeo, line 59: this seems important! I guess right now the new conversation will be on the client, and then on the next message we'll do the replacement in the agent reply handler?

DOsinga: yes, that's how it works. we'll probably print a warning in the reply that the format is unexpected and then go on.

Issue #3960, opened 2025-08-08
Tool-Calling Loop Interrupted Mid-Task


Innomen:
Describe the bug
Goose unexpectedly cancels a multi-step tool call (glob → text_editor) even though the local LLM backend successfully completed its task. This appears to be an internal orchestration issue, possibly a timeout, rather than a crash or LLM failure.

To Reproduce

Start Goose and a local LLM backend (Ollama or llama.cpp).

Use the provided prompt to initiate a task that requires analyzing and potentially editing a file (e.g., mod_api.gd).

Goose begins the task, but the tool-calling loop is interrupted midway through.

An "Execution error" message is displayed in the UI, and the Goose logs show a task cancellation (srv cancel_tasks).

Expected behavior
Goose should allow the local LLM enough time to complete its response and should not cancel the task if the LLM backend returns a successful response code. The tool-calling loop should continue as orchestrated.

Screenshots
(Attach screenshots of the Goose UI showing the "Execution error" and your terminal logs showing the LLM backend's successful response followed by Goose's cancel_tasks log.)

Please provide following information:
OS & Arch: Garuda Linux DGE (Dr460nized Gaming Edition) on x86_64

Interface: UI

Version: v1.2.0

Extensions enabled: Computer Controller (implied by glob and text_editor tool calls)

Provider & Model: Local LLM via Ollama or llama.cpp - Qwen3-Coder-30B-A3B-Instruct-UD-Q5_K_XL.gguf

Additional context
The LLM backend logs show that the evaluation was completed successfully, but Goose's logs indicate the task was canceled internally. This points to a potential disconnect or timeout issue within Goose, as it is canceling the task before it can receive and process the LLM's response.

Full Prompt Used:

Please analyze the current state of mod_api.gd and tell me:
1. What dependency-related code exists (if any)
2. Are there any syntax errors or incomplete functions?
3. Is the basic mod loading still functional?
4. What needs to be fixed or completed?

If there are issues, please fix them and ensure the core mod system still works.

towry:
Happens in goose v1.3.1

```
( O)> /Users/towry/Downloads/H5-登录@1x.jpg convert this into html page
I'll help you convert the login page image into an HTML page. I'll first analyze the design using the design_extractor tool, then generate the HTML code using the frontend_coder tool.

Let me start by extracting the design from the image:

─── create_task_design_extractor | subrecipe ──────────────────────────
task_parameters:
    -
        extraction_detail: comprehensive
[...]

PR #5189, opened 2025-10-15
fix: #3960 better approach to input schema for dynamic task params

alexhancock:
Use rust structs to derive the input schema for the `dynamic_task__create_task` tool

Part of (or maybe all the) work for #3960


alexhancock:
Ah yes, I forgot to apply the douwe-anti-vibecoding filter to the commit :) Will do now

PR #5191, opened 2025-10-15
Readonly developer extension

tlongwell-block:


crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -569,6 +581,30 @@ impl DeveloperServer {
         }
     }
 
+    /// Helper function to check if an operation is allowed in read-only mode
+    /// Returns an error if in read-only mode and the operation requires write access
+    fn require_write_access(&self, operation: &str) -> Result<(), ErrorData> {
+        if self.readonly {
+            // Special message for text_editor commands
+            if operation.starts_with("Command '") {
```
katzdave, line 589: Do we need this check? Since you only stuck this inside the non-readonly functions can't we just return an error if readonly isn't set.

crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -167,9 +167,17 @@ fn load_prompt_files() -> HashMap<String, Prompt> {
     prompts
 }
 
+/// Check if developer extension is in read-only mode
+fn is_readonly_mode() -> bool {
+    std::env::var("DEVELOPER_READONLY")
```
katzdave, line 172: Seems like a good enough mechanism. Would work for recipes.

katzdave (commented):
Overall I like it, and I think the env variable is a good enough way to just get this out.

Not sure what the best longterm solution is, maybe even a separate readonly_developer tool. 

Although I feel like readonly would be a common option we'd want to limit other mcp servers with.

crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -3863,4 +3911,359 @@ Additional instructions here.
...
+    async fn test_readonly_mode_initialization() {
+        // Test without environment variable
+        let server = create_test_server();
+        assert!(
+            !server.readonly,
+            "Should not be in read-only mode by default"
+        );
+
+        // Test with environment variable set to "true"
+        std::env::set_var("DEVELOPER_READONLY", "true");
```
DOsinga, line 3927: No! don't set environment variables in tests like this

crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -3863,4 +3911,359 @@ Additional instructions here.
             cleanup_test_service(running_service, peer);
         });
     }
+
+    // Read-only mode tests
+    #[tokio::test]
+    #[serial]
+    async fn test_readonly_mode_initialization() {
```
DOsinga, line 3918: delete this test, it is doing nothing

DOsinga:
closing > 2 weeks draft PRs

PR #5192, opened 2025-10-15
feat: Stable system prompt and tool ordering for more cache hits

jsibbison-square:
## Summary
When running multiple goose instances for the same tasks (eg. recipes) we currently do not get any llm cache hits on the initial request. From my testing this happens for multiple reasons:
- We have a second-based timestamp in the system prompt
- Tools are added in a random order
- Extension instructions come in a random order

This PR makes the timestamp truncated at the hour to allow for decent time accuracy but greatly improved cache hit potential and sorts the extensions (for instructions) and tools by name so they always appear in the same order.

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [x] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
I've used an api key with openai and downloaded their usage csv to verify that caching on the same initial request run in subsequent sessions with the developer mcp results in almost 100% cached input token usage.

### DIFF
--- a/crates/goose/src/prompts/system.md
+++ b/crates/goose/src/prompts/system.md
@@ -1,6 +1,8 @@
 You are a general-purpose AI agent called goose, created by Block, the parent company of Square, CashApp, and Tidal.
 goose is being developed as an open-source software project.
 
+The current date is {{current_date_time}}.
+
 goose uses LLM providers with tool calling capability. You can be used with different language models (gpt-4o,
 claude-sonnet-4, o1, llama-3.2, deepseek-r1, etc).
 These models have varying knowledge cut-off dates depending on when they were trained, but typically it's between 5-10


cloud-on-prem (approved):

crates/goose/src/agents/prompt_manager.rs
```diff
@@ -317,4 +320,31 @@ mod tests {
         assert!(result.contains("Extension help"));
         assert!(result.contains("hidden instructions"));
     }
+
+    #[test]
+    fn test_build_system_prompt_orders_extensions_by_name() {
```
DOsinga: delete this test

tlongwell-block:
Yeah, https://github.com/block/goose/pull/5027 removes the timestamp entirely from the system prompt

jsibbison-square:
@tlongwell-block So should I just delete the timestamp line in this PR so we get caching sooner or do you want me to not make that change and have your PR do that? I matched your change to delete the timestamp for system_prompt.md so shouldn't conflict when it merges

Once you respond I'll be able to merge.

crates/goose/src/agents/prompt_manager.rs
jsibbison-square: I thought it was useful to avoid regressing but have deleted it.

tlongwell-block:
> @tlongwell-block So should I just delete the timestamp line in this PR so we get caching sooner or do you want me to not make that change and have your PR do that? I matched your change to delete the timestamp for system_prompt.md so shouldn't conflict when it merges
> 
> Once you respond I'll be able to merge.

Until https://github.com/block/goose/pull/5027 is merged, this would mean goose won't know the time at all

@DOsinga, what do you think about merging https://github.com/block/goose/pull/5027 right after this one? If we're still not sure about MOIM as a concept, I'd say leave the timestamp in here @jsibbison-square. I will say that the MOIM build is my daily driver and works very well, including on the benchmarks.

jsibbison-square:
Ok I'll put the timestamp back at just the hourly granuality that way we get greater caching and you don't need to rush the other PR.

crates/goose/src/prompts/system.md
```diff
@@ -1,6 +1,8 @@
 You are a general-purpose AI agent called goose, created by Block, the parent company of Square, CashApp, and Tidal.
 goose is being developed as an open-source software project.
 
+The current date is {{current_date_time}}.
```
DOsinga, line 4: we could consider adding this to the end so we get most of the system prompt cached

jsibbison-square: Makes sense, will leave it unchanged here so tylers pr doesn't get a conflict when he removes the timestamp (we changed the timestamp to be at an hour granuality so we still get improved caching benefits).

PR #5194, opened 2025-10-16
Skip hidden & real format

DOsinga:
## Summary

compaction itself ran out of tokens because of how we created the thing to compact. we did not filter out the agent invisible messages, which basically means that after a few compaction turns you run out of space since we kept sending the already compacting messages over.

the other issue is that we would send raw json of the messages, including the id, visibility and everything which also uses 60% more tokens than this PR

### DIFF
--- a/crates/goose/src/context_mgmt/mod.rs
+++ b/crates/goose/src/context_mgmt/mod.rs
@@ -353,7 +353,9 @@ fn format_message_for_compacting(msg: &Message) -> String {
                 if let Ok(contents) = &res.tool_result {
                     let text_items: Vec<String> = contents
                         .iter()
-                        .filter_map(|content| content.as_text().map(|text_str| text_str.text.clone()))
+                        .filter_map(|content| {
+                            content.as_text().map(|text_str| text_str.text.clone())
+                        })
                         .collect();
 
                     if !text_items.is_empty() {


crates/goose/src/context_mgmt/mod.rs
```diff
@@ -293,46 +291,110 @@ async fn check_compaction_needed(
...
 
-    std::prelude::rust_2015::Ok(Some((response, provider_usage)))
+    Ok(Some((response, provider_usage)))
+}
+
+fn format_message_for_compacting(msg: &Message) -> String {
+    let content_parts: Vec<String> = msg
+        .content
+        .iter()
+        .map(|content| match content {
```
jamadeo, line 337: this is obviously a big improvement and we should do it.

how do we make sure we aren't removing something important here and then seeing decreased quality of compaction?

michaelneale:
trying this out

crates/goose/src/context_mgmt/mod.rs
michaelneale: all that noise around its and syntax adds a lot of overhead, so hard to imagine a world in which this isn't better all round (even ignoring it actually working beyond one compaction!)

DOsinga: I think we should add a bunch of compaction to the scenario tests, large-ish conversation, compaction and then needline in the haystack? /cc @katzdave

katzdave: Agreed. And would like some LLMish judge tests, to at least have high level sanity checks that we don't forget key information.

PR #5198, opened 2025-10-16
compaction: basic test and some small clean-up

jamadeo:
This started as code review feedback but then I wrote a test.


jamadeo:
took on new life in another PR

PR #5206, opened 2025-10-16
fix: revert built app name to uppercase Goose

zanesq:
## Problem

PR #4832 lowercased 'Goose' everywhere, but this is causing issues with:
- Case-sensitive filesystems breaking CI workflows
- Shortcuts and integrations expecting uppercase name  
- Platform conventions (macOS .app bundles typically use Title Case)

## Solution

This PR reverts **only the built application names** to use uppercase 'Goose' while keeping the lowercase branding in documentation and user-facing text.

## Changes

- **ui/desktop/package.json**: `productName` back to 'Goose'
- **All workflow files**: paths and artifact names use `Goose-*` format
- **macOS**: `Goose.app` instead of `goose.app`
- **Windows**: `Goose.exe` instead of `goose.exe`
- **Artifact names**: `Goose-darwin-arm64`, `Goose-darwin-x64`, `Goose-win32-x64`, `Goose-linux-x64`

## Testing

- ✅ Verified JSON syntax in package.json
- ✅ Ran cargo fmt check
- ✅ Verified no trailing whitespace or syntax errors

This ensures compatibility with existing tooling and platform expectations while maintaining the lowercase branding initiative for user-facing content.

### DIFF
--- a/ui/desktop/forge.config.ts
+++ b/ui/desktop/forge.config.ts
@@ -65,8 +65,8 @@ module.exports = {
     {
       name: '@electron-forge/maker-deb',
       config: {
-        name: 'goose',
-        bin: 'goose',
+        name: 'Goose',
+        bin: 'Goose',
         maintainer: 'Block, Inc.',
         homepage: 'https://block.github.io/goose/',
         categories: ['Development'],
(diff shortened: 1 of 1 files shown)


Abhijay007:
We might need to release a new version to test this change, maybe 1.10.3, to make it able to available to update in Goose

PR #5207, opened 2025-10-16
delete flaky pricing integration tests

wpfleger96:
## Summary
Fixes race condition causing `test_pricing_cache_performance` test to fail intermittently in CI with:
```
"No such file or directory (os error 2)"
```

Currently all 4 pricing integration tests share the same `GOOSE_CACHE_DIR` environment variable. Since environment variables are process-wide, when tests run in parallel (`cargo test --jobs 2` in CI), one test can overwrite another's env var value. This creates a race window where Test A sets the env var, Test B overwrites it, then Test A reads Test B's (potentially deleted) path and fails. This is resulting in [test failures](https://github.com/block/goose/actions/runs/18546001927/job/52866139240?pr=5085) on my unrelated PR

#3546 previously added `TempDir` isolation, but this only provided per-test directories and didn't prevent the tests from running in parallel and overwriting each other's environment variables.

~~To address this I added `#[serial]` attributes to all 4 pricing integration tests. This forces sequential execution of these tests only, while other tests continue running in parallel.~~

UPDATE: per feedback I'm just going to delete the flaky tests

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
 I had trouble reproducing this locally, so I created a synthetic test (`race_condition_demo.rs`) locally that deliberately triggers the race condition:

```rust
  #[tokio::test]
  async fn race_test_1() {
      let temp_dir = TempDir::new().unwrap();
      std::env::set_var("GOOSE_CACHE_DIR", temp_dir.path());
      sleep(Duration::from_millis(50)).await;  // Widen race window

      let read_path = std::env::var("GOOSE_CACHE_DIR").unwrap();
      // Fails if Test 2 overwrote the env var
      assert!(Path::new(&read_path).exists());
  }

  #[tokio::test]
  async fn race_test_2() {
      sleep(Duration::from_millis(10)).await;  // Let Test 1 start first
      let temp_dir = TempDir::new().unwrap();
      std::env::set_var("GOOSE_CACHE_DIR", temp_dir.path());  // Overwrites Test 1!
      drop(temp_dir);  // Delete directory Test 1 might try to use
  }
```

Results:
- Without `#[serial]`: Test 1 panics with `RACE CONDITION! Path doesn't exist: /tmp/.tmppTBwjl` (Test 2's deleted path)
- With `#[serial]`: Both tests pass—Test 1 reads its own path correctly

 ~~Additional validation: Stress tested the fixed pricing tests with 10 consecutive runs using `--jobs 4`. All runs passed.~~


crates/goose/tests/pricing_integration_test.rs
```diff
@@ -1,8 +1,10 @@
 use goose::providers::pricing::{get_model_pricing, initialize_pricing_cache, refresh_pricing};
+use serial_test::serial;
 use std::time::Instant;
 use tempfile::TempDir;
 
 #[tokio::test]
```
DOsinga: yeah, this is not the solution. we should just not set environment variables in tests. delete the tests I would say or maybe keep the one that does any real work

PR #5209, opened 2025-10-16
Revert "feat: enhance goose to search sessions for easy recall (#5177)"

alexhancock:
This reverts commit d142f4267d20a639982d3e51acad4652a9a26da6 temporarily

We saw it adding multiple copies of the chat_recall to config under two keys (`chat_recall` and `chatrecall`)

And it also fails to launch with error: `Error activating extension: Failed to add extension configuration, error: ConfigError("Unknown platform extension: chatrecall")`


angelahning (approved):

alexhancock:
Sorry @michaelneale!

Ran into some issues with this one on main and trying unblock a release from main

I have re-added this code in https://github.com/block/goose/pull/5211

If you can verify `just run-ui` works there to enable the extension, and it doesn't double add it to config would be great

PR #5212, opened 2025-10-16
Break compaction back into check_ and do_ compaction

katzdave:
This exposes clear points (marked with TODOs) as to where we can notify the client before + after compaction.

Also move the public api to use conversation instead of messages
__________

Following this up with more clear messaging on

Adding GooseThinkingMessage. Non agent visible message, that can set the state of the goose is thinking spinner. Will use this to say 'Compacting conversation...'

Refactor ConversationCompacted into SystemNotification; General messages that display in-line in the client, but aren't agent_visible. No special logic on trying to process these, they just render inline with the conversation wherever they fall.

### DIFF
--- a/crates/goose/src/context_mgmt/mod.rs
+++ b/crates/goose/src/context_mgmt/mod.rs
@@ -12,18 +12,6 @@ use tracing::{debug, info};
 
 pub const DEFAULT_COMPACTION_THRESHOLD: f64 = 0.8;
 
-/// Result of auto-compaction check
-#[derive(Debug)]
-pub struct AutoCompactResult {
-    /// Whether compaction was performed
-    pub compacted: bool,
-    /// The messages after potential compaction
-    pub messages: Conversation,
-    /// Provider usage from summarization (if compaction occurred)
-    /// This contains the actual token counts after compaction
-    pub summarization_usage: Option<crate::providers::base::ProviderUsage>,
-}
-
 #[derive(Serialize)]
 struct SummarizeContext {
     messages: String,


crates/goose/src/context_mgmt/mod.rs
```diff
@@ -198,34 +159,33 @@ Just continue the conversation naturally based on the summarized context"
...
-/// * `messages` - The current message history
+/// * `conversation` - The current conversation history
 /// * `threshold_override` - Optional threshold override (defaults to GOOSE_AUTO_COMPACT_THRESHOLD config)
 /// * `session_metadata` - Optional session metadata containing actual token counts
 ///
 /// # Returns
-/// * `CompactionCheckResult` containing detailed information about compaction needs
-async fn check_compaction_needed(
+/// * `CompactionCheckResult` containing whether compaction is needed and usage details
+pub async fn check_if_compaction_needed(
```
DOsinga, line 140: can't we just return a boolean here? also delete the intro comments

crates/goose/src/context_mgmt/mod.rs
```diff
@@ -46,101 +46,62 @@ struct SummarizeContext {
...
+            (&messages[..messages.len() - 1], Some(last_message.clone()))
+        } else if preserve_last_user_message {
+            // Last message is not a user message, but we want to preserve the most recent user message
+            // Find the most recent user message and copy it (don't remove from history)
+            let most_recent_user_message = messages
+                .iter()
+                .rev()
+                .find(|msg| matches!(msg.role, rmcp::model::Role::User))
+                .cloned();
+            (messages.as_slice(), most_recent_user_message)
```
DOsinga, line 59: this whole block strikes me as weird. there's two cases here, either the last message is a user message (i.e. the user said something and we decided to auto compact or the user said something, our autocompacting didn't trigger but we ran out tokens never the less, i.e. user has compact limit at 99% and posts something long) or it is not (i.e. we ran out of context while processing a tool call reply)

we set preserve_last_user_message to true if this is triggered from the tool calling loop, but in all other cases, the last message should be a user message. so I don't think we need this flag at all.

crates/goose/src/context_mgmt/mod.rs
katzdave: Yeah good call, I don't think we're actually doing anything interesting with CompactionCheckResult. Mostly just want to display the GOOSE_AUTO_COMPACT_THRESHOLD that we exceeded.

crates/goose/src/context_mgmt/mod.rs
katzdave: The first case where the last message is a user one, is from auto-compact. 

The second exists for the difference of manual summarize vs tool calling case you described.

In manual compaction we don't actually want to re-trigger the agent loop (and the last message is probably an agent message from the last response), but if we hit a tool call ContextLimitExceeded, I want to find and re-process the most recent user message and continue the agent loop automatically.

DOsinga: yes I understand. I'm just saying there's no difference. 
> but if we hit a tool call ContextLimitExceeded, I want to find and re-process the most recent user message and continue the agent loop automatically.

that is one way, but the other is that we ran out of context with a user mesage. so in either case you are looking for the last user message.

The other thing we're missing here I think is that tool responses are also user mesages, so you'll typically return that for the tool loop case.

PR #5214, opened 2025-10-16
fix the auto updater

jamadeo:
Pull download links from a manifest in the release, giving us flexibility to (re)name them


jamadeo:
closing this for now as we've had some other work here

PR #5215, opened 2025-10-16
FIX: prefer linux in WSL and add INSTALL_OS override for CLI

cbruyndoncx:
## Summary
FIX: prefer linux in WSL and add INSTALL_OS override for download_cli.sh

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [x] Build / Release
- [ ] Other (specify below)

### Testing
Tested with and without INSTALL_OS env var set, without now on WSL CLI defaults to linux.
With env var set, respects the choice windows or linux.

### Related Issues
Relates to #4928

### DIFF
--- a/download_cli.sh
+++ b/download_cli.sh
@@ -72,26 +72,52 @@ else
 fi
 
 # --- 3) Detect OS/Architecture ---
-# Better OS detection for Windows environments
-if [[ "${WINDIR:-}" ]] || [[ "${windir:-}" ]] || [[ "$OSTYPE" == "msys" ]] || [[ "$OSTYPE" == "cygwin" ]]; then
+# Allow explicit override for automation or when auto-detection is wrong:
+#   INSTALL_OS=linux|windows|darwin
+if [ -n "${INSTALL_OS:-}" ]; then
+  case "${INSTALL_OS}" in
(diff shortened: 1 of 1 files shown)


download_cli.sh
```diff
@@ -72,26 +72,52 @@ else
...
 
 # --- 3) Detect OS/Architecture ---
-# Better OS detection for Windows environments
-if [[ "${WINDIR:-}" ]] || [[ "${windir:-}" ]] || [[ "$OSTYPE" == "msys" ]] || [[ "$OSTYPE" == "cygwin" ]]; then
-    OS="windows"
-elif [[ -f "/proc/version" ]] && grep -q "Microsoft\|WSL" /proc/version 2>/dev/null; then
-    # WSL detection
+# Allow explicit override for automation or when auto-detection is wrong:
+#   INSTALL_OS=linux|windows|darwin
+if [ -n "${INSTALL_OS:-}" ]; then
```
alexhancock, line 77: nit: I would name it `OVERRIDE_OS` to signify what it is used for as `INSTALL_OS` and `OS` are kind of similar.

jamadeo:
related but non-conflicting I think: #5221

PR #5218, opened 2025-10-16
Replace compaction notifications with system notifications

katzdave:
Replace compaction notifications with more general system notifications. Deletes a bunch of compaction specific state in the client + the REST route for manual compaction.

The client sends a manual compact trigger message in the stream that the agent detects + processes. This makes the auto-compact + manual compact code nearly identical on the backend, with just some differences in the messaging + whether or not we resume the conversation.

The only downside with this approach is the compaction message shows up in the chat + message stream but I think we could avoid that in a follow up with some kind of way to inject additional metadata.

This also adds two types of SystemNotifications
- Inline just triggers special rendering ('compaction complete') etc.
- ThinkingMessage is also very lightweight; this must be at the end of the stream, and if it's detected there, the client will replace the thinking message with that. This does not actually change the thinking state of the client.

https://github.com/user-attachments/assets/f8ebf149-cb74-4f77-b5a2-0f40f81fecb6

### DIFF
--- a/crates/goose/src/conversation/message.rs
+++ b/crates/goose/src/conversation/message.rs
@@ -713,9 +713,7 @@ impl Message {
 
 #[cfg(test)]
 mod tests {
-    use crate::conversation::message::{
-        Message, MessageContent, MessageMetadata,
-    };
+    use crate::conversation::message::{Message, MessageContent, MessageMetadata};
     use crate::conversation::*;
     use rmcp::model::{
         AnnotateAble, CallToolRequestParam, PromptMessage, PromptMessageContent, PromptMessageRole,


katzdave:
Hmm actually what I've found is using the REST for manual compaction seems to create a lot of friction; needed a way on the client to control the thinking state.

Having it send a stream message instead, which puts it in the agent loop so it can share a closer code path with auto-compact + out of context compact.

alexhancock:
UX question: does it work to type "Please compact this conversation" the same way it does if the message is auto-inserted by clicking the button?

If it could, I actually don't even think of that as an issue for the UX compared to now, and opens up a nice way to tell it to compact without taking hands off keyboard.

crates/goose/src/agents/agent.rs
```diff
@@ -745,77 +747,99 @@ impl Agent {
...
-        };
-
-        let check_result = crate::context_mgmt::check_if_compaction_needed(
-            self,
-            &unfixed_conversation,
-            None,
-            session_metadata.as_ref(),
-        )
-        .await;
+        let is_manual_compact = unfixed_conversation.messages().last().is_some_and(|msg| {
```
DOsinga, line 750: this seems rather convoluted; I think it is because you want the manual compaction to use the same stream as we do for the agent reply? what was wrong with the old way?

crates/goose/src/agents/agent.rs
```diff
@@ -1138,23 +1162,29 @@ impl Agent {
                             }
                         }
                         Err(ProviderError::ContextLengthExceeded(_error_msg)) => {
-                            info!("Context length exceeded, attempting compaction");
+                            yield AgentEvent::Message(
+                                Message::assistant().with_system_notification(
+                                    SystemNotificationType::InlineMessage,
+                                    "Context limit reached. Attempting to compact and continue conversation...",
```
DOsinga: remove attempt, or even do we need two messages here?

crates/goose/src/context_mgmt/mod.rs
```diff
@@ -93,14 +93,6 @@ pub async fn compact_messages(
...
-    // Add the compaction marker (user_visible=true, agent_visible=false)
-    let compaction_marker = Message::assistant()
-        .with_conversation_compacted("Conversation compacted and summarized")
-        .with_metadata(MessageMetadata::user_only());
-    let compaction_marker_tokens: usize = 0; // Not counted since agent_visible=false
-    final_messages.push(compaction_marker);
-    final_token_counts.push(compaction_marker_tokens);
-
     // Add the summary message (agent_visible=true, user_visible=false)
     let summary_msg = summary_message.with_metadata(MessageMetadata::agent_only());
```
DOsinga, line 97: do we still need this block below? I thought we had already deleted that

crates/goose/src/conversation/message.rs
```diff
@@ -1182,4 +1203,30 @@ mod tests {
         assert!(metadata.user_visible);
         assert!(metadata.agent_visible);
     }
+
+    #[test]
```
DOsinga: delete

ui/desktop/src/utils/thinkingMessage.ts
```diff
@@ -0,0 +1,20 @@
+import { Message } from '../api';
+import { ChatState } from '../types/chatState';
+export function getThinkingMessage(messages: Message[], chatState: ChatState): string | undefined {
```
DOsinga: I think this should live in messages.ts. it should probably not take the chatState at all though, thinking is a property of the messages. in fact since when you call this you already check on message count and chat state, it could just be a property on a single message

ui/desktop/src/components/sessions/SessionHistoryView.tsx
```diff
@@ -101,29 +100,27 @@ const SessionMessages: React.FC<{
               </Button>
             </div>
           ) : filteredMessages?.length > 0 ? (
-            <ContextManagerProvider>
```
DOsinga, line 104: hurray! I think this causes rerenders

ui/desktop/src/hooks/useMessageStream.ts
```diff
@@ -307,6 +307,16 @@ export function useMessageStream({
                       mutateChatState(ChatState.WaitingForUserInput);
                     }
 
+                    const hasThinkingMessage = newMessage.content.some(
+                      (content) =>
+                        content.type === 'systemNotification' &&
```
DOsinga: I think this pattern appears in a bunch of places; as suggested below, let's just create a method in messages.ts

crates/goose/src/agents/agent.rs
katzdave: The old way triggered from REST vs a stream. It just made it a lot more complicated to keep the client in the right 'thinking' state. It also sent the client's view of the messages and used that which would cause a bit of rendering jitter when we did history replaced.

Could maybe split these more, but I think it's generally fine, main differences here are messaging + whether or not we continue the stream.

crates/goose/src/agents/agent.rs
katzdave: I think the system notification is still nice to render, because you'll see that. Removing attempt; it *should* confidently succeed!

crates/goose/src/context_mgmt/mod.rs
katzdave: The assistant_message? The summary is a user message historically; I think we could make it an agent one and then we don't need this. Just don't want the conversation to end in a user message unless we're trying to continue.

crates/goose/src/conversation/message.rs
katzdave: done

ui/desktop/src/utils/thinkingMessage.ts
katzdave: Done

ui/desktop/src/hooks/useMessageStream.ts
katzdave: done

katzdave:
> UX question: does it work to type "Please compact this conversation" the same way it does if the message is auto-inserted by clicking the button?
> 
> If it could, I actually don't even think of that as an issue for the UX compared to now, and opens up a nice way to tell it to compact without taking hands off keyboard.

Yeah it does. I considered /compact to mirror the cli but that didn't look as nice.

I think a future feature that could be cool is running a fast model on user messages to determine specific intents such as compaction or recipe generation.

PR #5221, opened 2025-10-16
fix: use Windows-compatible default path for CLI installation

### LINKED ISSUE #4851: Default path for CLI install on Windows with Bash does not exist, but script does not fail
**Describe the bug**

Default path is used from bash, in my case `/home/rob/.local/bin/goose.exe`. This does not translate to anything in Windows, and the installation docs show that you need to load the .local path, so when I do so goose is still not available. 

```
Warning: Goose installed, but /home/rob/.local/bin is not in your PATH.
Add it to your PATH by editing ~/.bashrc, ~/.zshrc, or similar:
    export PATH="/home/rob/.local/bin:$PATH"
Then reload your shell (e.g. 'source ~/.bashrc', 'source ~/.zshrc') to apply changes.
```

**To Reproduce**
1. Run the install script with bash `curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh | bash`
2. Follow the docs to update the path (shows to use `.local`)
3. Try to run goose and fail

```
goose: The term 'goose' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```

**Expected behavior**
Use a path that exists and things might start to work 😄.

aegntic:
Fixes #4851 - Address issue where default path does not exist on Windows.

Problem:
- Script used $HOME/.local/bin as default on all platforms
- On Windows with bash, $HOME resolves to Unix-style paths that do not exist
- Users get PATH instructions for non-existent directories

Solution:
- Added OS-aware default path selection:
  * Native Windows: $USERPROFILE/goose (Windows user profile)
  * WSL: $HOME/.local/bin (Linux-style path)
  * Linux/macOS: $HOME/.local/bin (existing behavior)
- Preserves custom GOOSE_BIN_DIR environment variable when set

Test Results:
- Script syntax validation: PASSED
- Windows path detection logic: PASSED
- Directory creation logic: PASSED
- Custom path preservation: PASSED

Users on Windows environments will now get a working default installation path and appropriate PATH setup instructions.

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:   

**Email**:

### DIFF
--- a/download_cli.sh
+++ b/download_cli.sh
@@ -54,7 +54,23 @@ fi
 # --- 2) Variables ---
 REPO="block/goose"
 OUT_FILE="goose"
-GOOSE_BIN_DIR="${GOOSE_BIN_DIR:-"$HOME/.local/bin"}"
+
+# Set default bin directory based on detected OS environment
+if [[ "${WINDIR:-}" ]] || [[ "${windir:-}" ]] || [[ "$OSTYPE" == "msys" ]] || [[ "$OSTYPE" == "cygwin" ]]; then
+    # Native Windows environments - use Windows user profile path
+    DEFAULT_BIN_DIR="$USERPROFILE/goose"
+elif [[ -f "/proc/version" ]] && grep -q "Microsoft\|WSL" /proc/version 2>/dev/null; then
+    # WSL - use Linux-style path but make sure it exists
+    DEFAULT_BIN_DIR="$HOME/.local/bin"
+elif [[ "$PWD" =~ ^/mnt/[a-zA-Z]/ ]]; then
+    # WSL mount point detection
+    DEFAULT_BIN_DIR="$HOME/.local/bin"
+else
+    # Default for Linux/macOS
+    DEFAULT_BIN_DIR="$HOME/.local/bin"
+fi
+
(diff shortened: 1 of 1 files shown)


download_cli.sh
```diff
@@ -54,7 +54,23 @@ fi
...
+elif [[ -f "/proc/version" ]] && grep -q "Microsoft\|WSL" /proc/version 2>/dev/null; then
+    # WSL - use Linux-style path but make sure it exists
+    DEFAULT_BIN_DIR="$HOME/.local/bin"
+elif [[ "$PWD" =~ ^/mnt/[a-zA-Z]/ ]]; then
+    # WSL mount point detection
+    DEFAULT_BIN_DIR="$HOME/.local/bin"
+else
+    # Default for Linux/macOS
+    DEFAULT_BIN_DIR="$HOME/.local/bin"
+fi
```
jamadeo, line 71: is there a reason to preserve three else clauses that all do the same thing? suggest simplifying it to a single else

PR #5223, opened 2025-10-17
migrating back with new chatrecall non underscore name

michaelneale:
fixes the reverted recall: https://github.com/block/goose/pull/5177 tool

It was broken as it was called "chat_recall" but the GUI code strips (for some reason) underscores when it saves config.yaml (cli doesn't) which will break any extension with _ in its name (this just removes it from the name) - same otherwise.

### DIFF
--- a/crates/goose/src/agents/chatrecall_extension.rs
+++ b/crates/goose/src/agents/chatrecall_extension.rs
@@ -114,7 +114,7 @@ impl ChatRecallClient {
 
                     let mut output = format!(
                         "Session: {} (ID: {})\nWorking Dir: {}\nTotal Messages: {}\n\n",
-                        loaded_session.description,
+                        loaded_session.name,
                         sid,
                         loaded_session.working_dir.display(),
                         total


alexhancock (approved):

michaelneale:
FYI this is meant to be off by default, but I think the default setting for all buildins is ignored currently (ie this and todo will always be on until you turn it off)

is that ok?

DOsinga:
huh, we should fix both those things in the frontend. it has no business trying to change the name of the extensions. I think a similar thing might have happened with @angelahning 's change.

DOsinga:
I looked into this briefly and this is such a mess. both the client and the server have a normalize name function. but they are not the same. one tolerates underscores, the other does not. and obviously the CLI takes a different path.

@alexhancock @zanesq your handles came up in the code about this, do you know why we do this at all? we also seem to expand the path to the exectuable in the client. this all seems a generous invitation to bugs

michaelneale:
I wonder if expanding to executable was for security (ie so it can only run the binaries that came with it?) - struggling to think why that would be (it may not have it on PATH otherwise). 

@DOsinga should I hold off merging this for stabilisation?

alexhancock:
@DOsinga @michaelneale I also found many things in this area that were confusing and messy last week. Multiple attempts at different places to replace paths to executable versions with "shims" etc. It seemed to me perhaps the best thing to do is simply not to touch any commands, but make sure the dir with the hermit executables (`node`, `npx`, `uvx`, etc) is in the `$PATH`

I can look at this in a separate change

DOsinga:
> @DOsinga @michaelneale I also found many things in this area that were confusing and messy last week. Multiple attempts at different places to replace paths to executable versions with "shims" etc. It seemed to me perhaps the best thing to do is simply not to touch any commands, but make sure the dir with the hermit executables (`node`, `npx`, `uvx`, etc) is in the `$PATH`
> 
> I can look at this in a separate change

It doesn't make any sense what we are doing right now, we mess around with names and executable etc but only in some paths. then in other paths we either do some of it on the server or not at all. we should treat the extension definitions as readonly. rip it all out! and then when it comes to actually executing the thing, maybe at that point we need to resolve the path to the executable, but that should be in one place in the extension manager I would thinkg

alexhancock:
@DOsinga makes sense - I can give it a go

michaelneale:
yeah - I think setting $PATH correctly is the go. Previously we would inherit the PATH from users (interactive) so didn't want to touch that, but now we dont' so can take more control I think. Ideally we just want hermit for pre-installing the things we know are problematic (python and node tool chains mainly)

michaelneale:
@alexhancock want me to hold on this until you are happy with things, tag me when ready and I can pick up this PR again? and check it (could even try slapping in an underscore again!) - or this is orthogonal?

alexhancock:
@michaelneale 

@DOsinga and I had a change in progress today, but if you've tested this and it doesn't have the double add bug, I say go ahead and then it'll just be another improvement when we merge the thing that improves extension management.

DOsinga:
do we know where the double add comes from? I would like to overhaul how we set the default list of extensions in a sensible way, but I don't quite understand how it is supposed to work also with the block version of this

michaelneale:
@DOsinga if by double add you mean chatrecall and chat_recall - in that case it was as cli allowed it, and then desktop (incorrectly) renamed it and added it with new name (won't do that in this case anymore).

PR #5226, opened 2025-10-17
chore: first cut at seeing if we can remove this dependency

michaelneale:
For desktop, default scheduler is fine, and this avoids a much more complex build tool chain and dependencies. 
If this works, we can consider temporal as an add on via some other means, vs built in

### DIFF
--- a/.github/workflows/pr-smoke-test.yml
+++ b/.github/workflows/pr-smoke-test.yml
@@ -5,9 +5,9 @@ on:
   workflow_dispatch:
     inputs:
       branch:
-        description: 'Branch to test'
+        description: "Branch to test"
         required: true
-        default: 'main'
+        default: "main"
         type: string
 
 name: Live Provider Tests
@@ -105,6 +105,7 @@ jobs:
           DATABRICKS_HOST: ${{ secrets.DATABRICKS_HOST }}
           DATABRICKS_TOKEN: ${{ secrets.DATABRICKS_TOKEN }}
           OPENROUTER_API_KEY: ${{ secrets.OPENROUTER_API_KEY }}
+          TETRATE_API_KEY: ${{ secrets.TETRATE_API_KEY }}
           HOME: /tmp/goose-home
           GOOSE_DISABLE_KEYRING: 1
           SKIP_BUILD: 1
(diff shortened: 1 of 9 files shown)


alexhancock (approved):
Only thing I see is some docs that should be removed.

```
$ git grep temporal -- documentation
documentation/docs/guides/environment-variables.md:| `GOOSE_SCHEDULER_TYPE` | Controls which scheduler Goose uses for [scheduled recipes](/docs/guides/recipes/session-recipes.md#schedule-recipe) | "legacy" or "temporal" | "legacy" (Goose's built-in cron scheduler) |
documentation/docs/guides/environment-variables.md:| `GOOSE_TEMPORAL_BIN` | Optional custom path to your Temporal binary | /path/to/temporal-service | None |
documentation/docs/guides/environment-variables.md:export GOOSE_SCHEDULER_TYPE=temporal
documentation/docs/guides/environment-variables.md:export GOOSE_TEMPORAL_BIN=/path/to/temporal-service
documentation/docs/guides/recipes/session-recipes.md:When scheduling Goose recipes with the CLI, you can use Goose's built-in cron scheduler (default), or the [Temporal scheduler](https://docs.temporal.io/evaluate/development-production-features/schedules) (requires the Temporal CLI). Switch from the default legacy scheduler by setting the `GOOSE_SCHEDULER_TYPE` [environment variable](/docs/guides/environment-variables.md#session-management):
documentation/docs/guides/recipes/session-recipes.md:  export GOOSE_SCHEDULER_TYPE=temporal
documentation/src/pages/recipes/data/recipes/subrecipes/chart-generator.yaml:     D. Time Series (if temporal data exists)
```

alexhancock:
Some stuff in `download_cli.ps1` too which doesn't seem to be removed yet

After that and the docs, LGTM!

DOsinga:
My theory was always that we needed this because Windows doesn't support cron, but the LLMs that be tell me tokio-cron-scheduler is supported on windows /cc @Kvadratni

Kvadratni:
Nah, the only reason I added that is to have an independent Daemon that runs even if goose is closed. Temporal had a lot of promise feature wise but it is overkill in the end

michaelneale:
I think if there was another way to extend desktop so it could hook in to things running on machine (if temporal installed) that could be useful for truly long running things. Of course now, we are expecting that people have goose desktop running semi long term anyway so it isn't a terrible solution for simple scheduling needs.

DOsinga:
I agree

michaelneale:
@alexhancock I removed appropriate ones from docs (some were the word "temporal" just meaning time) - also from the download scripts.

blackgirlbytes (approved):

blackgirlbytes:
looks good -- doc wise

alexhancock:
This LGTM - going for it

PR #5229, opened 2025-10-17
fix: correct schema for openai tools

michaelneale:
Currently openai won't work due to: 

Request failed: Invalid schema for function 'dynamic_task__create_task': In context=('properties', 'extensions', 'type', '0'), array schema items is not an object. (code: invalid_function_parameters, type: invalid_request_error) (status 400).

as that tool generates an invalid schema for openai, which started with: https://github.com/block/goose/pull/5189

### DIFF
--- a/crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
+++ b/crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
@@ -47,6 +47,8 @@ impl From<ExecutionModeParam> for ExecutionMode {
     }
 }
 
+type JsonObject = serde_json::Map<String, Value>;
+
 /// Parameters for a single task
 #[derive(Debug, Serialize, Deserialize, JsonSchema)]
 pub struct TaskParameter {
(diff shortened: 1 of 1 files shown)


crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
```diff
@@ -63,18 +63,23 @@ pub struct TaskParameter {
     pub description: Option<String>,
 
     #[serde(skip_serializing_if = "Option::is_none")]
+    #[schemars(with = "Option<Vec<serde_json::Map<String, Value>>>")]
```
jamadeo: I think the right fix is to change the `Value`s here to Map<String, Value>, not just in the schemars annotation but in the actual type

jamadeo: `items: true` is apparently a valid way in json schema to say anything goes, but openai doesn't like that. But also, it's not what we want (or at least not what the prior hand-rolled schema had). we want objects only, not other values

jamadeo: unrelated to this issue, but how does this tool work if we are just putting plain objects in the schema? are we telling the model somewhere else what those parameters look like?

DOsinga: ah, this is how this works! we should just insert the json schema derived from an actual Recipe definition in here @tlongwell-block @yingjiehe-xyz

michaelneale: yeah I think we have hit this with openai before - things don't always adhere to what is technically allowed.

DOsinga: we can always fix things up in the provider formatter as we do for google, but here we just need to generate the correct schema, not use Value

Issue #4030, opened 2025-08-12
Downloaded extensions don't work in Goose Desktop for Windows


sings-to-bees-on-wednesdays:
**Describe the bug**

You can download extensions for Goose for Windows, but Goose isn't able to actually load them.

**To Reproduce**
1. Click "Browse Extensions"
2. Find an extension you like
3. Click "Install"
4. Click "OK"
5. See error

**Expected behavior**
Extensions should load

**Screenshots**

<img width="734" height="416" alt="Image" src="https://github.com/user-attachments/assets/336f4283-2ff9-4414-8933-97e946aeb27a" />

**Please provide following information:**
 - **OS & Arch:** Windows 11
 - **Interface:** UI
 - **Version:** 1.3.0
 - **Extensions enabled:** Any
 - **Provider & Model:** N/A

**Additional context**
The error (when you click "Copy Error") is:

```
Error activating extension: Node.js installer script not found
```

rromanw:
Bump, any ideas what to do ?

milllan:
still not working for me :(

DOsinga:
can you copy the error when this pops up and share here?

understood-the-assignment:
@DOsinga This was already done, you can find it at the bottom of the issue description above. That's the entire contents of the copied error.

DOsinga:
> Error activating extension: Node.js installer script not found

I see. well, that's no good.

vlascik:
The whole idea with node.js installer is not great and doesn't really work. 

I have a fix in the making at least for cases where node.js is installed, but not in the default Program files folder (it tries to read the node path from registry instead), but the whole Windows situation is bad - the builds are constantly broken, extensions do not work, node MCPs don't work, etc.

Good Windows support is going to need a lot more automated testing infrastructure.

Anyway, in the meantime, I will try to post a few fixes in next few days.

DOsinga:
unfortunately you are right, the key issue is that we do not have anybody on the core team that uses windows on a day to day basis

vlascik:
@DOsinga Yeah, figured out that much :) Given that there's still about as many developers on Windows as there are on Mac + Linux combined, not to mention the end-user ratios, that's probably not a great situation...

Btw. there is a way to run Windows in docker containers, so maybe at least some tests could be set up? At least verifying that the app builds, starts, and can perform some basic functions.

[...]

PR #5233, opened 2025-10-17
Fix: get node path from Windows registry, clearer error messages

vlascik:
## Summary
On Windows when adding `npx` extensions, Goose tries to find `node.exe` at default paths in `Program Files`, and then fails for everyone that installed `node` elsewhere. This PR tries to also find the path from the registry, which should mitigate the problem for people that already have node installed, but in a different folder.

Also the error messages are made a bit clearer about what the problem is, so that users can try to rectify it on their own.

The node.js install script doesn't seem to quite work, but that's left for another PR.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Build on Windows

### Related Issues
Relates to https://github.com/block/goose/issues/4030

### DIFF
--- a/crates/goose-server/src/routes/extension.rs
+++ b/crates/goose-server/src/routes/extension.rs
@@ -37,8 +37,8 @@ async fn add_extension(
     if let ExtensionConfig::Stdio { cmd, .. } = &request.config {
         if cmd.ends_with("npx.cmd") || cmd.ends_with("npx") {
             // Check if Node.js is installed in standard locations
-            let mut node_exists = std::path::Path::new(r"C:\\Program Files\\nodejs\\node.exe").exists()
-                || std::path::Path::new(r"C:\\Program Files (x86)\\nodejs\\node.exe").exists();
+            let mut node_exists = std::path::Path::new(r"C:\Program Files\nodejs\node.exe").exists()
+                || std::path::Path::new(r"C:\Program Files (x86)\nodejs\node.exe").exists();
 
             // Also check Windows registry: HKEY_LOCAL_MACHINE\\SOFTWARE\\Node.js InstallPath
             if !node_exists {


crates/goose-server/src/routes/extension.rs
```diff
@@ -32,12 +32,44 @@ async fn add_extension(
 
     // If this is a Stdio extension that uses npx, check for Node.js installation
     #[cfg(target_os = "windows")]
+    use winreg::enums::{HKEY_LOCAL_MACHINE, KEY_READ};
```
DOsinga, line 35: this looks great. does this mean we can delete the code in the API handler that tries to do the same thing? but only in the API so that would not work for the CLI I think?

vlascik: what code do you mean? All I can see are some shell scripts trying to install node.js again in a default directory, which is probably not going to work for people that already have node installed elsewhere anyway (the chances are their install will just be updated).

vlascik: I see what you mean now, there is a new PR that moved  this code to crates/goose-server/src/routes/agent.rs now, is this what you meant?

DOsinga: yeah sorry I moved this - do you want to do that or should I just copy your code over if that is faster

vlascik: @DOsinga just copy it, it will be faster.

vlascik:
Files moved around, so I did a new rebase of this PR at https://github.com/block/goose/pull/5731 to make review/merge easier.

PR #5234, opened 2025-10-17
fix Windows build ring crate missing advapi32 link

vlascik:
## Summary
Windows build fails due to missing link to advapi32.dll needed to build `ring` crate.
```
error: linking with `link.exe` failed: exit code: 1120
  |
  = note: "C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.44.35207\\bin\\HostX64\\x64\\link.exe" "/NOLOGO" 
...
goose\\target\\debug\\build\\ring-33b93150c5211568\\build_script_build-33b93150c5211568.exe" "/OPT:REF,NOICF" "/DEBUG" "/PDBALTPATH:%_PDB%" "/NATVIS:<sysroot>\\lib\\rustlib\\etc\\intrinsic.natvis" "/NATVIS:<sysroot>\\lib\\rustlib\\etc\\liballoc.natvis" "/NATVIS:<sysroot>\\lib\\rustlib\\etc\\libcore.natvis" "/NATVIS:<sysroot>\\lib\\rustlib\\etc\\libstd.natvis"
...
  = note: libjobserver-de3391caf3272945.rlib(jobserver-de3391caf3272945.jobserver.37a54c1ed64fde8f-cgu.3.rcgu.o) : error LNK2019: unresolved external symbol SystemFunction036 referenced in function _ZN9jobserver3imp9getrandom17h9e1b1b056175cb58E␍

          P:\goose\target\debug\build\ring-33b93150c5211568\build_script_build-33b93150c5211568.exe : fatal error LNK1120: 1 unresolved externals␍
```          

GPT suggested doing this fix:

`.cargo/config.toml`
```
[target.x86_64-pc-windows-msvc]
rustflags = ["-C", "link-args=advapi32.lib"]

[target.x86_64-pc-windows-gnu]
rustflags = ["-C", "link-args=-ladvapi32"]
```
and this works, but adding the link this way looks fairly unstandard, maybe someone with better Rust skill could suggest a better/more canonical way to do it? I don't have much time to delve into this.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [x] Build / Release
- [ ] Other (specify below)

### Testing
Windows build

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,5 @@
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-C", "link-args=advapi32.lib"]
+
+[target.x86_64-pc-windows-gnu]
+rustflags = ["-C", "link-args=-ladvapi32"]


alexhancock:
We modify these files during ci https://github.com/block/goose/blob/cd2ea778f0b1132dce1e7bace6466dbb09c845c4/.github/workflows/build-cli.yml#L166

I don't know much of the history here, but perhaps this fix needs to be there if it is affecting ci builds?

Can you say more when/where you saw build failures related to this @vlascik ?

vlascik:
> We modify these files during ci
> 
> https://github.com/block/goose/blob/cd2ea778f0b1132dce1e7bace6466dbb09c845c4/.github/workflows/build-cli.yml#L166
> 
> I don't know much of the history here, but perhaps this fix needs to be there if it is affecting ci builds?
> 
> Can you say more when/where you saw build failures related to this @vlascik ?

As seen above, this also breaks build of `ring` crate on Windows (without docker), apparently the build needs a link to `advapi32.dll`, which is a Windows encryption library, and it's missing. 

Not sure if this is a common problem with `ring` or something about `goose`, and there probably is a better way to fix it, but without this fix I was unable to finish `cargo build` (and therefore run/debug the builds) on Windows.

I'm afraid this requires more rust/Win32api knowledge than I have and I don't know much more about this.

What I would suggest though, is to setup at least some real Windows testing, at least to make sure the builds can run on the platform and apps start or something - because as it is, the Windows version breaks constantly.

michaelneale:
thanks @vlascik - some windows packaging/impl work underway so hopefully will be addressed shortly.

PR #5235, opened 2025-10-17
fix: re-sync package-lock.json

alexhancock:
package-lock.json got out of sync in version control - this re-syncs it


alexhancock:
Super minor - merging

PR #5236, opened 2025-10-17
improve provider request logging a bit

jamadeo:
- handle streaming responses (write chunks to the file)
- capture errors from the provider

### DIFF
--- a/crates/goose/src/providers/snowflake.rs
+++ b/crates/goose/src/providers/snowflake.rs
@@ -309,7 +309,6 @@ impl Provider for SnowflakeProvider {
     ) -> Result<(Message, ProviderUsage), ProviderError> {
         let payload = create_request(model_config, system, messages, tools)?;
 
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/utils.rs
```diff
@@ -452,40 +455,95 @@ pub fn unescape_json_values(value: &Value) -> Value {
...
-        "output": response,
-        "usage": usage,
-    });
-
-    if let Err(e) = std::fs::write(
-        log_path(0),
-        serde_json::to_string_pretty(&data).unwrap_or_default(),
-    ) {
-        tracing::warn!("Failed to write log file: {}", e);
+    fn finish(&mut self) -> Result<()> {
```
DOsinga, line 525: we could put a lock on this to avoid a not very likely race condition

DOsinga (approved):

PR #5237, opened 2025-10-17
Next camp

DOsinga:
This makes one window sort of workable for new chats, resume chats and recipes. Also brings back recipe acceptance and security scanning to BaseChat2

Things to do

- [x] check on why starting from hub flickers
- [x] allow to start a chat from pair if no chat is open yet
- [x] hub starts also seem to render the conversation twice
- [ ] hook up recipe variables entry
- [ ] centralize recipe launching 
- [ ] find other places where we still open a new window
- [ ] integrate with @lifeizhou-ap 's work for creating recipes with system prompts the right way


zanesq:
Will look into the following in another branch off this one.

- check on why starting from hub flickers
- hub starts also seem to render the conversation twice

While it should be default to launch in same window we should continue to support open in a new window for sessions and recipes and add another button for that option (users might not want to always overwrite their current chat window) -- can follow up with this later

DOsinga:
supporting multiple windows seems fine, but we then do need to make sure it talks to the same goosed still

zanesq:
tried a new chat on this branch and getting

```
12:56:38.554 › goosed stderr for port 58183 and dir /Users/zane/Development/goose-main:   2025-10-17T19:56:38.554157Z ERROR goosed::routes::reply: Failed to read session for : Session not found
    at crates/goose-server/src/routes/reply.rs:240
```

image

DOsinga:
oh is this when you start a chat from pair? I think we might have to still implement that too!

zanesq:
ah yes from pair, I thought that was working when we were sharing earlier but must have been mistaken, its working from hub can use that for now 👍

DOsinga:
only if a conversation is already loaded. can look into this next week

zanesq:
@DOsinga fixed above issue when new chat in pair and reload

PR #5245, opened 2025-10-17
nextcamp - fix flickering

zanesq:
## Summary
- fixed flickering when rendering messages (used a ref instead of a cache)
- fixed initial load flickering popular topics before starting chat
- brought back goose is working on it chatstate display
- enhanced the logging and added a flag

### DIFF
--- a/ui/desktop/src/hooks/useChatStream.ts
+++ b/ui/desktop/src/hooks/useChatStream.ts
@@ -5,6 +5,7 @@ import { getApiUrl } from '../config';
 import { createUserMessage } from '../types/message';
 
 const TextDecoder = globalThis.TextDecoder;
+const resultsCache = new Map<string, { messages: Message[]; session: Session }>();
 
 // Debug logging - set to false in production
 const DEBUG_CHAT_STREAM = true;
@@ -218,6 +219,12 @@ export function useChatStream({
   const [chatState, setChatState] = useState<ChatState>(ChatState.Idle);
   const abortControllerRef = useRef<AbortController | null>(null);
 
+  useEffect(() => {
+    if (session) {
+      resultsCache.set(sessionId, { session, messages });
+    }
+  }, [sessionId, session, messages]);
+
   const renderCountRef = useRef(0);
   renderCountRef.current += 1;
   console.log(`useChatStream render #${renderCountRef.current}, ${session?.id}`);
@@ -366,10 +373,16 @@ export function useChatStream({
     setChatState(ChatState.Idle);
   }, []);
 
+  const cached = resultsCache.get(sessionId);
+  const maybe_cached_messages = session ? messages : cached?.messages || [];
+  const maybe_cached_session = session ?? cached?.session;
+
+  console.log('>> returning', sessionId, Date.now(), maybe_cached_messages, chatState);
+
   return {
     sessionLoadError,
-    messages,
-    session,
+    messages: maybe_cached_messages,
+    session: maybe_cached_session,
     chatState,
     handleSubmit,
     stopStreaming,


DOsinga:
this seems to remove the pre-caching though

ui/desktop/src/components/BaseChat2.tsx
```diff
@@ -257,7 +272,7 @@ function BaseChatContent({
             <div className="absolute bottom-1 left-4 z-20 pointer-events-none">
               <LoadingGoose
                 message={
-                  messages.length === 0
+                  messages.length === 0 && chatState === ChatState.Thinking
```
DOsinga: we should catch this all with chatState and then just have a mapping from chatState to message here I think

ui/desktop/src/components/BaseChat2.tsx
```diff
@@ -87,11 +87,26 @@ function BaseChatContent({
...
+      initialMessage &&
+      session &&
+      !hasSubmittedInitialMessage.current &&
+      initialMessagesLength.current === 0
+    ) {
+      hasSubmittedInitialMessage.current = true;
       handleSubmit(initialMessage);
     }
-  }, [initialMessage, session, messages, handleSubmit]);
+  }, [initialMessage, session, handleSubmit]);
```
DOsinga: we just removed so much crufty hard to follow conditional code and now it is already coming back :(

this is trying to auto-submit initialMessage exactly once, but I don't think it works when switching sessions - hasSubmittedInitialMessage and initialMessagesLength never reset when sessionId changes. Even if we fixed that bug, the code is hard to follow with three separate refs, two effects with overlapping dependencies, and implicit coordination between them

I think we can make this work better if we move this into useChatStream; it can just decide after it has loaded the session from the server whether it should submit immediately and it is already scoped to the sessionId

ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -69,9 +98,12 @@ function pushMessage(currentMessages: Message[], incomingMsg: Message): Message[
 async function streamFromResponse(
   response: Response,
   initialMessages: Message[],
-  setMessages: (messages: Message[], log: string) => void,
+  updateMessages: (messages: Message[]) => void,
```
DOsinga, line 103: any particular reason to rename this?

ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -197,59 +274,101 @@ export function useChatStream({
           },
           throwOnError: true,
         });
-        const session = response.data;
-        console.log('resume agent returned', session?.id, session?.conversation?.length);
-        setSession(session);
-        setMessagesAndLog(session?.conversation || [], 'load-session');
+        if (cancelled) return;
+
+        const sessionData = response.data;
```
DOsinga: why rename this? it is the session here

ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -145,47 +209,60 @@ export function useChatStream({
...
+        from: sessionIdRef.current.slice(0, 8),
+        to: sessionId.slice(0, 8),
+      });
+      sessionIdRef.current = sessionId;
+      setMessagesAndLog([], 'session-reset');
+      setSession(undefined);
+      setSessionLoadError(undefined);
+      setChatState(ChatState.Idle);
+    }
+  }, [sessionId, setMessagesAndLog]);
```
DOsinga: this useEffect has the same dependencies as the one below. I don't think there's a reason to have both. it also introduces a sessionIdRef that only checks if the sessionId has changed, but the sessionId is in the dependencies, so the useEffect already only runs when sessionId changes, no need for that ref.

I think all we want to do is add the resets to the block below

ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -197,59 +274,101 @@ export function useChatStream({
...
+          onFinish
+        );
 
-      await streamFromResponse(response, currentMessages, setMessagesAndLog, onFinish);
+        log.stream('stream-complete');
+      } catch (error) {
+        if (error instanceof Error && error.name !== 'AbortError') {
+          log.error('submit failed', error);
+          onFinish('Submit error: ' + error.message);
+        } else if (error instanceof Error && error.name === 'AbortError') {
```
DOsinga: this can just be an else; but is also feels like duplication of what we're doing in stream from. so if we want to keep the exception catch here, we should remove the other one

ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -197,59 +274,101 @@ export function useChatStream({
...
+        log.error('session load failed', error);
         setSessionLoadError(error instanceof Error ? error.message : String(error));
+
+        log.state(ChatState.Idle, { reason: 'session load error' });
         setChatState(ChatState.Idle);
       }
     })();
+
+    return () => {
+      cancelled = true;
```
DOsinga, line 301: good idea

ui/desktop/src/hooks/useChatStream.ts
zanesq: just trying to avoid confusion and make it clear that its doing more than setMessages (setMessagesAndLog), really we're updating messages happy to change it back if you think that is confusing also

zanesq:
Unless I'm missing something I don't think the pre-caching was needed as we still have to wait for the server response and it was creating a race condition where the component had to coordinate between its own loaded state and useChatStream's state. Now we have a single source of truth in useChatStream and no flickering issues.

Was the pre-cache intended to avoid rendering the existing chat messages when switching back to the chat?

ui/desktop/src/components/BaseChat2.tsx
zanesq: good call 👍

ui/desktop/src/components/BaseChat2.tsx
```diff
@@ -184,6 +180,15 @@ function BaseChatContent({
...
   const initialPrompt = messages.length == 0 && recipe?.prompt ? recipe.prompt : '';
+
+  // Map chatState to LoadingGoose message
+  const getLoadingMessage = (): string | undefined => {
+    if (isCompacting) return 'goose is compacting the conversation...';
+    if (messages.length === 0 && chatState === ChatState.Thinking) {
+      return 'loading conversation...';
+    }
+    return undefined;
+  };
```
DOsinga, line 191: What I meant to say, but we can maybe do this in some other PR that we should move to a world where the useChatStream determines the chatState and BaseChat does not contain any logic. So isCompacting should just be a chatState, as should loading be. I checked and we are already doing this in the LoadingGoose component, so let's move the messaging we generate to one point

ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -6,7 +6,36 @@ import { createUserMessage } from '../types/message';
 
 const TextDecoder = globalThis.TextDecoder;
 
-const resultsCache = new Map<string, { messages: Message[]; session: Session }>();
+// Debug logging - set to false in production
+const DEBUG_CHAT_STREAM = true;
```
DOsinga, line 11: we should delete this once we are sure it works really

ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -197,59 +266,110 @@ export function useChatStream({
...
+          onFinish('Submit error: ' + (error instanceof Error ? error.message : String(error)));
+        }
+      }
     },
-    [sessionId, onFinish, setMessagesAndLog]
+    [sessionId, setMessagesAndLog, onFinish]
   );
 
+  useEffect(() => {
+    if (initialMessage && session && messages.length === 0 && chatState === ChatState.Idle) {
```
DOsinga, line 363: This is an improvement. I think we can do even better (maybe in a follow up) and make ChatState into a true statemachine that allows us to see exactly what's going on.

that way all we have to do here is depend on chatState and if it is autoSubmit and there is an initialMessage

DOsinga:
> Unless I'm missing something I don't think the pre-caching was needed as we still have to wait for the server response and it was creating a race condition where the component had to coordinate between its own loaded state and useChatStream's state. Now we have a single source of truth in useChatStream and no flickering issues.
> 
> Was the pre-cache intended to avoid rendering the existing chat messages when switching back to the chat?

Be careful deleting things that you don't know what they are for; Chesterton had his fence.

so the goal of the precaching is that we have a version of the chat available immediately upon load when we hit a chat we've already worked with. in the current world switching back to a chat that you've worked with will reset the provider, model & extensions (in line with how it works now). this can be slow, so getting a copy immediately would be preferable.

I'm not sure I am convinced this was the reason for the flickering though; the flickering only seemed to happen when we created a new conversation from the hub, in which cache there was no pre-cache by definition - the flickering happened during the streaming when we already have a session and the precache should be overwritten. there could be a bug in all of this of course, but I think this is the way to do it.

PR #5250, opened 2025-10-18
Revert "Standardize Session Name Attribute"

michaelneale:
Reverts block/goose#5085

that change won't allow people to roll back a version of goose: 

image

17:46:43.054 › from renderer: [UNHANDLED REJECTION] Error: Failed to create session: error returned from database: (code: 1) table sessions has no column named description

```
thread 'main' panicked at /Users/micn/Development/goose/crates/goose-cli/src/session/builder.rs:319:10:
called `Result::unwrap()` on an `Err` value: error returned from database: (code: 1) table sessions has no column named description
```


wpfleger96:
@michaelneale I’m AFK and can’t take a closer look rn but I’m pretty sure that failing test is flaky

DOsinga:
we should also not die on this type of errors; move the session db to a backup and create a new one?

michaelneale:
ugh... how can I now roll back a version to get my sessions back (don't really care about them just need to test stuff!)

wpfleger96:
@michaelneale this is a little hacky, but you can use my (now-reverted) DB helper script: https://github.com/wpfleger96/goose/blob/2389e98f16643c84b8093d15b59effe14734f68e/scripts/goose-db-helper.sh

you should be able to `goose-db-helper.sh generate-migrations` then `goose-db-helper.sh migrate-to 3` where v3 is the current most recent schema version (v4 was the one that got reverted)

PR #5251, opened 2025-10-18
goose remote access

michaelneale:
allows user to open a tunnel to a goosed instance which can be used by remove mobile client(s)
This is for clients reaching in to goose (desktop) - via a tunnel (not for goose desktop to access remote goosed server - that will be a later change). 

image

image

this uses https://github.com/michaelneale/lapstone-tunnel or alternatively tailscale

### DIFF
--- a/ui/desktop/src/components/settings/tunnel/TunnelSection.tsx
+++ b/ui/desktop/src/components/settings/tunnel/TunnelSection.tsx
@@ -25,7 +25,7 @@ const STATUS_MESSAGES = {
   disabled: 'Tunnel is disabled',
 } as const;
 
-const IOS_APP_STORE_URL = 'https://apps.apple.com/app/goose-ai-assistant/id6738286074';
+const IOS_APP_STORE_URL = 'https://apps.apple.com/us/app/goose-ai/id6752889295';
 
 export default function TunnelSection() {
   const [tunnelInfo, setTunnelInfo] = useState<TunnelInfo>({


ui/desktop/src/components/settings/app/AppSettingsSection.tsx
```diff
@@ -37,6 +39,13 @@ export default function AppSettingsSection({ scrollToSection }: AppSettingsSecti
     setIsMacOS(window.electron.platform === 'darwin');
   }, []);
 
+  // Check if platform supports Tailscale tunnel (macOS or Linux)
```
DOsinga: yes, well, that sort of comments

ui/desktop/TUNNEL_IMPLEMENTATION.md
```diff
@@ -0,0 +1,144 @@
+# Tailscale Tunnel Implementation - Complete
```
DOsinga: I don't think we want to add this

ui/desktop/src/components/settings/app/AppSettingsSection.tsx
```diff
@@ -394,6 +403,35 @@ export default function AppSettingsSection({ scrollToSection }: AppSettingsSecti
         </CardContent>
       </Card>
 
+      {/* Experimental Features - only show on macOS and Linux, and not hidden by env var */}
```
DOsinga: this seems overly complicated; we should probably just use the ALPHA flag we already have for now to test with the team and then when it works somewhat, add a section here to configure it

ui/desktop/src/components/settings/tunnel/TunnelSection.tsx
```diff
@@ -0,0 +1,384 @@
...
+  }, []);
+
+  const checkTunnelStatus = async () => {
+    try {
+      const status = await window.electron.getTunnelStatus();
+      setTunnelStatus(status as TunnelStatus);
+    } catch (err) {
+      console.error('Error checking tunnel status:', err);
+    }
+  };
```
DOsinga: I think we can clean this up a bit; have just one useEffect, merge window.electron.getTunner* and properly type it. also figure out what to do if it fails instead of just dropping the exception

ui/desktop/src/components/settings/tunnel/TunnelSection.tsx
```diff
@@ -0,0 +1,384 @@
...
+    setError(null);
+    setTunnelStatus({ state: 'starting', info: null });
+
+    try {
+      const tunnelInfo = await window.electron.startTunnel();
+      setTunnelStatus({ state: 'running', info: tunnelInfo as TunnelInfo });
+      setShowQRModal(true);
+    } catch (err) {
+      console.error('Error starting tunnel:', err);
+      setError(err instanceof Error ? err.message : 'Failed to start tunnel');
```
DOsinga: we have a new fangled errorMessage() method now for this

ui/desktop/src/components/settings/tunnel/TunnelSection.tsx
```diff
@@ -0,0 +1,384 @@
...
+  ipv6: string;
+  hostname: string;
+  secret: string;
+  port: number;
+}
+
+interface TunnelStatus {
+  state: 'idle' | 'starting' | 'running' | 'error';
+  info: TunnelInfo | null;
+}
```
DOsinga: make this:

```
type TunnelStatus =
  | { state: 'idle' }
  | { state: 'starting' }
  | { state: 'running'; info: TunnelInfo }
  | { state: 'error'; error: string };
```
  to take avantage of the type system

ui/desktop/src/components/settings/tunnel/TunnelSection.tsx
```diff
@@ -0,0 +1,384 @@
...
+    const tunnelUrl = tunnelStatus.info.ipv4
+      ? `http://${tunnelStatus.info.ipv4}:${tunnelStatus.info.port}`
+      : tunnelStatus.info.url;
+
+    const configJson = JSON.stringify({
+      url: tunnelUrl,
+      secret: tunnelStatus.info.secret,
+    });
+    const urlEncodedConfig = encodeURIComponent(configJson);
+    return `goosechat://configure?data=${urlEncodedConfig}`;
```
DOsinga, line 119: so in the mobile app we register this schema and then you can configure it just from the camera app?

ui/desktop/src/components/settings/tunnel/TunnelSection.tsx
```diff
@@ -0,0 +1,384 @@
...
+    }
+  };
+
+  // Generate QR code data
+  const getQRCodeData = () => {
+    if (!tunnelStatus.info) return '';
+
+    // Use the actual tunnel URL (works for both Tailscale and Lapstone)
+    // For Tailscale: uses ipv4 if available, otherwise falls back to url
+    // For Lapstone: always uses the public Cloudflare URL
```
DOsinga: this comment is harder to read than the actual code, so I would drop it. Since we're getting this from the other side anyway, why not always have a url in tunnelStatus and we can worry about it there?

ui/desktop/src/utils/tunnel.ts
```diff
@@ -0,0 +1,347 @@
...
+  return settings.tunnelMode || 'lapstone';
+}
+
+export function setTunnelMode(mode: TunnelMode): void {
+  const settings = loadSettings();
+  settings.tunnelMode = mode;
+  saveSettings(settings);
+}
+
+export interface TunnelInfo {
```
DOsinga: this looks like a duplicate

ui/desktop/src/utils/tunnel.ts
```diff
@@ -0,0 +1,347 @@
...
+  hostname: string;
+  secret: string;
+  port: number;
+  pids: {
+    goosed: number;
+    tailscale_serve: number;
+  };
+}
+
+export type TunnelState = 'idle' | 'starting' | 'running' | 'error';
```
DOsinga: also known as TunnelStatus elsewhere

ui/desktop/src/utils/settings.ts
```diff
@@ -9,13 +9,18 @@ export interface EnvToggles {
...
 export interface Settings {
   envToggles: EnvToggles;
   showMenuBarIcon: boolean;
   showDockIcon: boolean;
   schedulingEngine: SchedulingEngine;
   enableWakelock: boolean;
+  tunnelSecret?: string;
+  tunnelAgentId?: string;
+  tunnelMode?: TunnelMode;
+  tunnelAutoStart?: boolean;
```
DOsinga: we should not be using this - settings should go through the server, saving the settings here will potentially miss changes we made since.

ui/desktop/src/components/settings/tunnel/TunnelSection.tsx
michaelneale: heh - this one was actually, ironically by me (and left for me!) yes is confusing.

ui/desktop/src/components/settings/tunnel/TunnelSection.tsx
michaelneale: yes, that is correct.

ui/desktop/src/utils/settings.ts
michaelneale: yes, that makes sense. I was thinking goose-server didn't need to care about this, but could shift logic there.

ui/desktop/src/components/settings/app/AppSettingsSection.tsx
michaelneale: yep, can use that, and not have it in a folded experimental section?

michaelneale:
@DOsinga 

```
Right now we duplicate a bunch of code starting goosed, but we already have a goosed and we are already connected to it, so there's no need for that.
```

Yes - it would need to have a stable secret for the exchange however (different for local electron, but we could change that). 

```
In the ideal world I think goosed would do all this itself. It would see if there's a tunnel config and if so start it and handle all the traffic directly. Maybe we should have a quick video chat about this? It seems related to other ways we talk to goosed like ACP and maybe even the web
```

yes was trying to re-use as much as possible with existing one, as most of the work (originally) was launching goosed (and the tunnel client code is natively in ts/web socket - wonder what that looks like in rust, can take a look but not optimistic it wouldn't be a lot more complex), but benefit is it lives in the one goosed (like other things will like acp) - but the idea also was that this is just http to goosed, it wasn't really meant to be another protocol (the tunnel isn't super complex, and in the case of tailscale it is just ngrok like port forwarding). Restarting/network awareness is good to have (and hard to get right, lots of manual testing)

michaelneale:
@DOsinga ported it to rust side now! (that was fiddly - mostly to test it with various network connections)

crates/goose-server/src/tunnel/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+            anyhow::bail!("Tunnel is already running or starting");
+        }
+        *state = TunnelState::Starting;
+        drop(state);
+
+        let config = self.config.read().await.clone();
+
+        let tunnel_secret = config.secret.clone().unwrap_or_else(generate_secret);
+        let server_secret = std::env::var("GOOSE_SERVER__SECRET_KEY")
+            .expect("GOOSE_SERVER__SECRET_KEY must be set for tunnel to work");
```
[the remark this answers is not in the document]
michaelneale: nope, there is no case when it wouldn't be set, the expect is the problem, this is more like an assertion to help debugging in weird future cases and is used elsewhere.

crates/goose-server/src/tunnel/lapstone.rs
```diff
@@ -0,0 +1,548 @@
...
+            loop {
+                tokio::time::sleep(Duration::from_secs(60)).await;
+                let elapsed = last_activity_clone.read().await.elapsed();
+                if elapsed > Duration::from_secs(IDLE_TIMEOUT_SECS) {
+                    warn!(
+                        "No activity for {} minutes, forcing reconnect",
+                        IDLE_TIMEOUT_SECS / 60
+                    );
+                    break;
+                }
```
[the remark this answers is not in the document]
michaelneale: this is not true, and clearly by design so it keeps things alive if all else fails, and it reconnects

Justfile
```diff
@@ -187,13 +187,18 @@ lint-ui:
 
 # make GUI with latest binary
 make-ui:
+    @just release-binary
+    cd ui/desktop && APPLE_IDENTITY="" APPLE_ID="" APPLE_ID_PASSWORD="" APPLE_APP_SPECIFIC_PASSWORD="" APPLE_TEAM_ID="" npm run bundle:default
+
+# make GUI with signing and notarization (requires Apple Developer credentials)
+make-ui-signed:
     @just release-binary
     cd ui/desktop && npm run bundle:default
```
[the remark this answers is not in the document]
michaelneale: nope, this is by design. wrong again, and again.

crates/goose-server/src/tunnel/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+            anyhow::bail!("Tunnel is already running or starting");
+        }
+        *state = TunnelState::Starting;
+        drop(state);
+
+        let config = self.config.read().await.clone();
+
+        let tunnel_secret = config.secret.clone().unwrap_or_else(generate_secret);
+        let server_secret =
+            std::env::var("GOOSE_SERVER__SECRET_KEY").unwrap_or_else(|_| "test".to_string());
```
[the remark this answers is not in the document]
michaelneale: nope, same as core.

crates/goose-server/src/tunnel/lapstone.rs
```diff
@@ -0,0 +1,627 @@
...
+            response,
+            status,
+            headers_map,
+            request_id,
+            message.path,
+            ws_tx,
+        )
+        .await?;
+    } else {
+        let body = response.text().await.unwrap_or_default();
```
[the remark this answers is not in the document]
michaelneale: no that is incorrect, again

crates/goose-server/src/tunnel/mod.rs
```diff
@@ -0,0 +1,263 @@
...
+                                agent_id,
+                                lapstone_handle.clone(),
+                                new_restart_tx,
+                            )
+                            .await
+                            {
+                                Ok(_) => {
+                                    *state.write().await = TunnelState::Running;
+                                    tracing::info!("Tunnel restarted successfully");
+                                    restart_rx = new_restart_rx;
```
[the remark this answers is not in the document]
michaelneale: this is harmfully wrong feedback

crates/goose-server/src/commands/agent.rs
```diff
@@ -59,6 +59,13 @@ pub async fn run() -> Result<()> {
     let listener = tokio::net::TcpListener::bind(settings.socket_addr()).await?;
     info!("listening on {}", listener.local_addr()?);
 
+    let tunnel_manager = app_state.tunnel_manager.clone();
+    tokio::spawn(async move {
+        // Give time for GUI to render before doing autostart
+        tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
```
DOsinga: why do we need time for the GUI to render here? and why 3 seconds

crates/goose-server/src/tunnel/config.rs
```diff
@@ -0,0 +1,49 @@
+use super::TunnelConfig;
```
DOsinga: I don't think this module adds all that much. we create a layer between the config object we already use but can't we just call everything directly? now when you change the auto start we also write the secret etc.

just read and write it when we need it, is what I would think

crates/goose-server/src/tunnel/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+        }
+    }
+
+    pub async fn start(&self) -> anyhow::Result<TunnelInfo> {
+        let mut state = self.state.write().await;
+        if *state != TunnelState::Idle {
+            anyhow::bail!("Tunnel is already running or starting");
+        }
+        *state = TunnelState::Starting;
+        drop(state);
```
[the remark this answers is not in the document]
DOsinga: yeah holding the lock longer seems decent

crates/goose-server/src/tunnel/mod.rs
```diff
@@ -0,0 +1,263 @@
...
+                        let auto_start = config.read().await.auto_start;
+                        if auto_start {
+                            tracing::warn!("Tunnel connection lost, initiating restart...");
+
+                            lapstone::stop(lapstone_handle.clone()).await;
+                            *state.write().await = TunnelState::Idle;
+
+                            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+
+                            *state.write().await = TunnelState::Starting;
```
DOsinga: the restart logic here is rather similar to what we are doing with the starting logic. I would suggest factoring it out in one method that you call twice

ui/desktop/src/utils/tunnel.ts
```diff
@@ -0,0 +1,75 @@
+import log from './logger';
```
DOsinga: I don't think you need this file; all of this is generated through the API, no? so you can call that instead

ui/desktop/src/main.ts
```diff
@@ -1310,6 +1310,37 @@ ipcMain.handle('get-wakelock-state', () => {
   }
 });
 
+ipcMain.handle('start-tunnel', async (_event, baseUrl: string) => {
```
DOsinga: this can just go through the API

ui/desktop/src/preload.ts
```diff
@@ -119,6 +120,9 @@ type ElectronAPI = {
   hasAcceptedRecipeBefore: (recipe: Recipe) => Promise<boolean>;
   recordRecipeHash: (recipe: Recipe) => Promise<boolean>;
   openDirectoryInExplorer: (directoryPath: string) => Promise<boolean>;
+  startTunnel: (baseUrl: string) => Promise<TunnelInfo>;
```
DOsinga: idem

ui/desktop/src/components/settings/tunnel/TunnelSection.tsx
```diff
@@ -0,0 +1,256 @@
...
+            <div className="p-3 bg-red-100 dark:bg-red-900/20 border border-red-300 dark:border-red-800 rounded text-sm text-red-800 dark:text-red-200">
+              {error}
+            </div>
+          )}
+
+          <div className="flex items-center justify-between">
+            <div>
+              <h3 className="text-text-default text-xs">Tunnel Status</h3>
+              <p className="text-xs text-text-muted max-w-md mt-[2px]">
+                {tunnelInfo.state === 'idle' && 'Tunnel is not running'}
```
DOsinga: use a look up table for this

ui/desktop/src/components/settings/tunnel/TunnelSection.tsx
```diff
@@ -0,0 +1,256 @@
...
+                  </Button>
+                  <Button onClick={handleStopTunnel} variant="destructive" size="sm">
+                    Stop Tunnel
+                  </Button>
+                </>
+              )}
+              {tunnelInfo.state === 'error' && (
+                <Button onClick={handleStartTunnel} variant="default" size="sm">
+                  Retry
+                </Button>
```
DOsinga, line 190: this all reads a bit confusing. error & idle seem to have the same effect; I would probbaly keep one button and then do the state machine in the click (but the button doesn't change variant then maybe)

crates/goose-server/src/tunnel/ed25519.rs
```diff
@@ -0,0 +1,176 @@
...
+                .try_into()
+                .map_err(|_| anyhow::anyhow!("Invalid signature length"))?,
+        );
+
+        self.public_key.verify(message.as_bytes(), &signature)?;
+        Ok(())
+    }
+}
+
+#[cfg(test)]
```
DOsinga: yeah these tests I don't think are adding much. delete

crates/goose-server/src/tunnel/ed25519.rs
```diff
@@ -0,0 +1,176 @@
...
+        if age < -MAX_AGE_SECS {
+            anyhow::bail!("Signature timestamp too far in future");
+        }
+
+        let body_hash = if let Some(body_content) = body {
+            let hash = Sha256::digest(body_content.as_bytes());
+            hex::encode(hash)
+        } else {
+            String::new()
+        };
```
DOsinga: this produces a different hash for empty string vs none, maybe just default body ot String:new() and then follow the happy path?

crates/goose-server/src/routes/tunnel.rs
```diff
@@ -0,0 +1,82 @@
...
+                }),
+            )
+                .into_response()
+        }
+    }
+}
+
+/// Stop the tunnel
+///
+/// Stops the currently running tunnel and optionally clears the auto-start setting.
```
DOsinga: ah yes can you remove these comments. stop the tunnel doesn't help and I think the other one is a lie

crates/goose-server/src/commands/agent.rs
michaelneale: yeah may not now - in the past for some reason, it wouldn't let it always start correctly subsequently - never could work out why. Takes a long time for that problem to show up (ie run for 30 mins, restart) so hard to know.

crates/goose-server/src/tunnel/ed25519.rs
michaelneale: did help me find a timing bug - but I think I may not keep this code anyway (the whole file)

ui/desktop/src/components/settings/tunnel/TunnelSection.tsx
michaelneale: yeah I don't even want a retry actually - do or do not, there is no retry (as yoda says)

crates/goose-server/src/tunnel/mod.rs
michaelneale: yes good idea - the restart is a fairly late addition (where I made it actually do similar to start - in fact, I want it to do EXACTLY the same thing, so yes).

ui/desktop/src/utils/tunnel.ts
michaelneale: oh yeah - the hey api thing... forgot about that.

crates/goose-server/src/tunnel/config.rs
michaelneale: yeah I think this is a remnant from porting it from ts.

crates/goose-server/src/tunnel/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+        }
+    }
+
+    async fn start_tunnel_internal(&self) -> anyhow::Result<(TunnelInfo, mpsc::Receiver<()>)> {
+        let config = self.config.read().await.clone();
+        let server_port = get_server_port()?;
+
+        let tunnel_secret = config.secret.clone().unwrap_or_else(generate_secret);
+        let server_secret =
+            std::env::var("GOOSE_SERVER__SECRET_KEY").unwrap_or_else(|_| "test".to_string());
```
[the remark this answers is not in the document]
michaelneale: have mentioned before, this is from elsewhere, and not an issue.

michaelneale:
ok did another cleanout @DOsinga

crates/goose-server/src/tunnel/lapstone.rs
```diff
@@ -0,0 +1,611 @@
...
+    #[serde(rename = "requestId")]
+    request_id: String,
+    status: u16,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    headers: Option<HashMap<String, String>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    body: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    error: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
DOsinga, line 79: a lot of these fields are optional; can we make the ones (especially the booleans) non optional? a boolean is true or false

crates/goose-server/src/routes/tunnel.rs
```diff
@@ -0,0 +1,78 @@
...
+        (status = 200, description = "Tunnel stopped successfully"),
+        (status = 500, description = "Internal server error", body = ErrorResponse)
+    )
+)]
+pub async fn stop_tunnel(State(state): State<Arc<AppState>>) -> Response {
+    state.tunnel_manager.stop(true).await;
+    StatusCode::OK.into_response()
+}
+
+/// Get tunnel info
```
DOsinga, line 59: can you remove the comments that just reflect the names? maybe also drop /api from the paths bring it more in line w the rest

crates/goose-server/src/tunnel/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub url: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub hostname: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub secret: Option<String>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct TunnelConfig {
```
DOsinga: I don't think I would keep this around as a separate class; just read & write to the config directly

crates/goose-server/src/tunnel/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+pub enum TunnelState {
+    #[default]
+    Idle,
+    Starting,
+    Running,
+    Error,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct TunnelInfo {
```
DOsinga, line 30: these fields feel like they don't need to be optional

crates/goose-server/src/tunnel/lapstone.rs
michaelneale: you know LLMs, they are never able to commit, are they.

michaelneale: actually - no they need to be, this is http tunnelling, may not be errors, may not be body (pretty common). boolean can probably default though, but this is network code.

crates/goose-server/src/routes/tunnel.rs
michaelneale: these are deliberately left in (FYI) as they are used in api docs (but are probably not needed) _ just so you know it isn't slop (and I do miss helpful comments)

michaelneale: cleaned

crates/goose-server/src/tunnel/lapstone.rs
```diff
@@ -0,0 +1,607 @@
...
+}
+
+async fn handle_chunked_response(
+    body: String,
+    status: u16,
+    headers_map: HashMap<String, String>,
+    request_id: String,
+    message_path: String,
+    ws_tx: WebSocketSender,
+) -> Result<()> {
```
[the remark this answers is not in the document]
michaelneale: what ... on earth kind of comment is this?

crates/goose-server/src/tunnel/lapstone.rs
```diff
@@ -0,0 +1,607 @@
...
+            response,
+            status,
+            headers_map,
+            request_id,
+            message.path,
+            ws_tx,
+        )
+        .await?;
+    } else {
+        let body = response.text().await.unwrap_or_default();
```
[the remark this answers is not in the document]
michaelneale: *sigh* we have been over this before

Issue #5242, opened 2025-10-17
Recent Update Has Broken Gemini support


jatinsu:
**Describe the bug**
when updating to a recent goose version using a google gemini model returns the following message:

`
Ran into this error: Request failed: Request failed with status: 400 Bad Request. Message: Invalid JSON payload received. Unknown name "$ref" at 'tools.function_declarations[3].parameters.properties[0].value.any_of[0]': Cannot find field..
`

I have not checked which version broke this, but it works in version v1.0.16

**To Reproduce**
run `goose update` and then `goose configure` and select Google Gemini model

**Expected behavior**
goose to return an output

**Screenshots**
v1.0.16

<img width="1903" height="381" alt="Image" src="https://github.com/user-attachments/assets/eec42518-9ba7-41ac-b2ce-57e170d06cb0" />

v1.11.0

<img width="1911" height="468" alt="Image" src="https://github.com/user-attachments/assets/c1e17e76-7946-4267-8993-4558087a46bd" />

**Please provide following information:**
 - **OS & Arch:** Fedora 42
 - **Interface:** CLI
 - **Version:** 1.11.0
 - **Extensions enabled:** None
 - **Provider & Model:** Google - gemini-2.0-flash

**Additional context**

DOsinga:
this is odd. I tried with gemini-2.5-flash and that worked, but I do get this error on 2.0 - can you try the newer model?

jatinsu:
Yep! Tried it with gemini-2.5-flash as well

<img width="992" height="512" alt="Image" src="https://github.com/user-attachments/assets/0cea627e-1217-44c7-96fa-24a0ba670e79" />

cyberone:
Reproducible for me in the GUI as well. However, cli version works with the 2.5-flash model.

atlasoft:
same 
```
goose --version
 1.11.0
```

```
starting session | provider: google lead model: gemini-2.5-pro worker model: gemini-2.5-pro
    session id: 20251018_24
    working directory: 

goose is running! Enter your instructions, or try asking what goose can do.

Context: ○○○○○○○○○○ 0% (0/1000000 tokens)
( O)> tell me a joke

Ran into this error: Request failed: Request failed with status: 400 Bad Request. Message: Invalid JSON payload received. Unknown name "$ref" at 'tools.function_declarations[3].parameters.properties[0].value.any_of[0]': Cannot find field..

Please retry if you think this is a transient or recoverable error.
```

atlasoft:
[...]

PR #5257, opened 2025-10-18
Skip subagents for gemini

DOsinga:
Our subagents tool declaration does not work on gemini and it keeps braking, so disabling it now for that provider. also introduces a central switch so if other things break we can easily skip sub-agents. also cleans up how we create prompts and avoid dying on them.

fixes: #5242

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -1746,13 +1746,8 @@ mod tests {
         );
 
         let prompt_manager = agent.prompt_manager.lock().await;
-        let system_prompt = prompt_manager.build_system_prompt(
-            vec![],
-            None,
-            Value::Null,
-            &"gpt-4o".to_string(),
-            false,
-        );
+        let system_prompt =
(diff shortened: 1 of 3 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -1754,7 +1747,7 @@ mod tests {
 
         let prompt_manager = agent.prompt_manager.lock().await;
         let system_prompt =
-            prompt_manager.build_system_prompt(vec![], None, Value::Null, None, false);
+            prompt_manager.build_system_prompt(vec![], None, Value::Null, "gpt-4o", false);
```
amed-xyz, line 1750: why is this hardcoded now?

amed-xyz (approved):

crates/goose/src/agents/agent.rs
DOsinga: this is a test, so we have to provide it something. the agent here isn't set up with a model. the test doesn't seem to do much so we could also drop it

God-damnit-all:
This didn't manage to fix the issue unfortunately.

Issue #5137, opened 2025-10-12
Goose configure writes empty config / Ubuntu 24.04 / KeePassXC Secret Service


cruftex:
**Describe the bug**

Trying to do initial setup with custom provider. Goose just writes an empty config.

**To Reproduce**

````
~/tmp/goose-test$ cat ~/.config/goose/config.yaml
cat: /home/jeans/.config/goose/config.yaml: No such file or directory
~/tmp/goose-test$ goose configure

This will update your existing config file
  if you prefer, you can edit it directly at /home/jeans/.config/goose/config.yaml

┌   goose-configure 
│
◇  What would you like to configure?
│  Custom Providers 
│
◇  What would you like to do?
│  Add A Custom Provider 
│
◇  What type of API is this?
│  OpenAI Compatible 
│
◇  What should we call this provider?
│  Local LM Studio
│
◇  Provider API URL:
│  http://localhost:1234
│
◇  API key:
│  ▪▪▪▪
│
◇  Available models (seperate with commas):
│  qwen/qwen3-coder-30b
│
◇  Does this provider support streaming responses?
│  Yes 
│
~/tmp/goose-test$ echo $?
0
~/tmp/goose-test$ cat /home/jeans/.config/goose/config.yaml
{}
````

I am using Ubuntu 24.04 and KDE Neon and keepassxc with Secret Service support.
I noticed that the unlock dialog of keepassxc popped up when I frist try to install.
However, I don't want to store anything there, the key is not needed for local tests anyway.

The documentation says something about a keyring and the environment variable OPENAI_KEY should be set. This did not change anything.

When I open the vault and run the setup again, Goose writes a new secret into it and stops with:

````
└  Custom provider added: Local LM Studio
````

Still `config.yaml` is empty:

````
goose info -v
Goose Version:
  Version:          1.9.3

Goose Locations:
  Config file:      /home/jeans/.config/goose/config.yaml
  Sessions dir:     /home/jeans/.local/share/goose/sessions
  Logs dir:         /home/jeans/.local/state/goose/logs

Goose Configuration:
  No configuration values set
  Run 'goose configure' to configure goose
````

**Expected behavior**

An error message with a reasonable hint if configuration is not possible

If the problem is the secret storage, a working documented way to it switch off secret storage

Also, since it says `/home/jeans/.config/goose/config.yaml` the expectation is that ONLY this file will be changes. Secrets should only be written when instructed or with confirmation.

**Screenshots**

**Please provide following information:**
[...]

Issue #2472, opened 2025-05-08
WSL Ubuntu 22.04-- Cannot Add Provider- Constantly Exiting At Final Step


SigmaEnterprise:
No matter what I try on WSL on windows ,it keeps exiting when I try to load model at final step it keeps exiting and never saves config settings

WSL Ubuntu 22.04

Any help would be appreciated

It always just exits at that final step

![Image](https://github.com/user-attachments/assets/84cc310e-7b0e-4396-9300-1b5cf2da71d4)

megafetis:
+

bjornbyte:
I believe I'm hitting the same bug (Windows 11 WSL2) trying to configure goose with copilot:
```
❯ goose configure

Welcome to goose! Let's get you set up with a provider.
  you can rerun this command later to update your configuration

┌   goose-configure
│
◇  Which model provider should we use?
│  Github Copilot
│
◇  Enter a model from that provider:
│  gpt-4o
│
◇  Execution error: failed to get api info after 3 attempts
│
└  Failed to configure provider: init chat completion request with tool did not succeed.

  Warning: We did not save your config, inspect your credentials
   and run 'goose configure' again to ensure goose can connect
```

I tried to check the logs but the path reported by `goose info` for logs (and other things), `~/.local/state/goose/logs` does not exist.

bjornbyte:
although I just tried configuring with google gemini and succeeded, so maybe a different but similar problem. Let me know if I should open a separate ticket.

GitMurf:
You need to disable the secure storage stuff by setting this env variable: `export GOOSE_DISABLE_KEYRING=true`

I got copilot to work after I did this.

I did try gemini as well and got that to work first so if for whatever reason copilot doesn't work right away, try getting google gemini to work first and then try copilot again.

bjornbyte:
that worked, thanks!

SigmaEnterprise:
I'm using it with no problems on MX Linux Libretto 23.6

DOsinga:
yeah, export GOOSE_DISABLE_KEYRING=1 seems to do the trick. however we should default to that if writing to the keyring fails. it seems hard to keep this sort of thing working on all linux varieties

PR #5260, opened 2025-10-18
Dont exit silently when storing api key fails

DOsinga:
Mostly for linux; key storage doesn't work and we silently exit.

fixes: #5137
fixes: #2472

### DIFF
--- a/crates/goose-cli/src/commands/configure.rs
+++ b/crates/goose-cli/src/commands/configure.rs
@@ -11,7 +11,10 @@ use goose::config::extensions::{
     name_to_key, remove_extension, set_extension, set_extension_enabled,
 };
 use goose::config::permission::PermissionLevel;
-use goose::config::{Config, ConfigError, ExperimentManager, ExtensionEntry, PermissionManager};
+use goose::config::signup_tetrate::TetrateAuth;
+use goose::config::{
+    configure_tetrate, Config, ConfigError, ExperimentManager, ExtensionEntry, PermissionManager,
+};
 use goose::conversation::message::Message;
 use goose::model::ModelConfig;
 use goose::providers::{create, providers};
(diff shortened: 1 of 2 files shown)


crates/goose-cli/src/commands/configure.rs
```diff
@@ -465,7 +481,9 @@ pub async fn configure_provider_dialog() -> Result<bool, Box<dyn Error>> {
                     .interact()?
                 {
                     if key.secret {
-                        config.set_secret(&key.name, Value::String(env_value))?;
```
jamadeo, line 465: this already returns an error, so I looked at why we'd want to turn that into a Ok(false), and it's because we throw it away here:

https://github.com/block/goose/blob/1b532cc8168f7456c35a56a43851f71628c3831d/crates/goose-cli/src/cli.rs#L781

I think a better fix would be to just not throw it away there, no?

crates/goose-cli/src/commands/configure.rs
```diff
@@ -411,7 +411,23 @@ fn select_model_from_list(
...
+    key_name: &str,
+    value: String,
+) -> Result<bool, Box<dyn Error>> {
+    match config.set_secret(key_name, Value::String(value)) {
+        Ok(_) => Ok(true),
+        Err(e) => {
+            cliclack::outro(style(format!(
+                "Failed to store {} securely: {}. Please ensure your system's secure storage is accessible. Alternatively you can run with GOOSE_DISABLE_KEYRING=true or set the key in your environment variables",
+                key_name, e
+            )).on_red().white())?;
```
jamadeo, line 427: still feels like this should be an Err and the function should return Result<(), Box<dyn Error>> but this is still an improvement, and fixing that might mean more tinkering with the callers than you had hoped

cruftex:
Nice! Thanks!

PR #5263, opened 2025-10-20
live testing script

michaelneale:
Trying to run the live testing script as part of CI

### DIFF
--- a/scripts/test_providers.sh
+++ b/scripts/test_providers.sh
@@ -19,8 +19,21 @@ PROVIDERS=(
   "openai:gpt-4o:gpt-4o-mini:gpt-3.5-turbo"
   "anthropic:claude-sonnet-4-5-20250929:claude-opus-4-1-20250805"
   "google:gemini-2.5-pro:gemini-2.5-pro:gemini-2.5-flash"
-  #"databricks:databricks-claude-sonnet-4:gemini-2-5-flash:gpt-4o"
 )
+
+# In CI, only run Databricks tests if DATABRICKS_HOST and DATABRICKS_TOKEN are set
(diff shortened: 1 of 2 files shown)


scripts/test_providers.sh
```diff
@@ -3,18 +3,23 @@ if [ -f .env ]; then
...
 
 PROVIDERS=(
-  "openrouter:anthropic/claude-sonnet-4.5:google/gemini-flash-2.5:qwen/qwen3-coder"
+  "openrouter:anthropic/claude-sonnet-4.5:qwen/qwen3-coder"
   "openai:gpt-4o:gpt-4o-mini:gpt-3.5-turbo"
-  "anthropic:claude-sonnet-4-0:claude-3-7-sonnet-latest"
+  "anthropic:claude-sonnet-4-5-20250929:claude-opus-4-1-20250805"
   "google:gemini-2.5-pro:gemini-2.5-pro:gemini-2.5-flash"
-  "databricks:databricks-claude-sonnet-4:gemini-2-5-flash:gpt-4o"
+  #"databricks:databricks-claude-sonnet-4:gemini-2-5-flash:gpt-4o"
```
DOsinga: ideally we'd put this back for people at home following along, but rather get this in sooner than later.

michaelneale: hrm - wonder if can change script so that it keeps state, ie once it passes, won't let it fail (but is ok for now)

michaelneale: @DOsinga enhanced it so if you run it locally it will run databricks. but for now in CI will not run it until it sees the host configured (so we can use same script in both places)

PR #5243, opened 2025-10-17
Revert "Rewrite extension management tools"

wpfleger96:
Reverts block/goose#5057

We are seeing failures on `main` with the new `Extension Manager` extension:
```
🐚 ~/Development/goose main ./target/debug/goose
◆  Would you like me to help debug the 'Extension Manager' extension failure?
│  ○ Yes  / ● No
```

Currently I'm only able to build and run Goose on `main` if I disable the 'Extension Manager' extension via config:
```
  extensionmanager:
    available_tools: []
    bundled: true
    description: Enable extension management tools for discovering, enabling, and disabling extensions
    enabled: false
    name: Extension Manager
    type: platform
```

PR #5273, opened 2025-10-20
Revert "Revert "Rewrite extension management tools""

angelahning:
Reverts block/goose#5243

Fixed issues with the extension manager extension - it was getting added with the key "extensionmanager" but the look up was done using the name "Extension Manager".


DOsinga:
> We should look at whether the normalization/key rewriting is really necessary at all. I know i've seen goose rename extension keys in my config and I can't explain why it's necessary. (Outside of this change's scope of course)

yes. @alexhancock and me had a crack and removing some cruft: https://github.com/block/goose/pull/5281 makes things go through the API while also removing name and cmd normalization. needs more testing, but I think it is the way

PR #5276, opened 2025-10-20
fix: include Windows GNU CLI artifact in releases

zanesq:
The goose-x86_64-pc-windows-gnu.zip artifact was excluded from releases starting in v1.10.3 due to a case-sensitive pattern change in the release workflow.

This fix adds 'goose-*.zip' pattern alongside 'Goose*.zip' to ensure both CLI and desktop Windows builds are included in releases.

fixes https://github.com/block/goose/discussions/5275

Fixes the issue where Windows GNU CLI builds were missing from:
- v1.10.3
- v1.11.0
- v1.11.1

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -105,6 +105,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           artifacts: |
             goose-*.tar.bz2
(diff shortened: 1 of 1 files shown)


zanesq:
Ignore that mention pasted wrong link :)

PR #5280, opened 2025-10-20
fix: Show recipe content when recipe parsing fails

jsibbison-square:
## Summary
Currently when a recipe fails to be rendered we get an error without seeing the content of the recipe. This makes it very hard to debug particularly when the cause of the failure is the injected content from parameters.

Current example errors:
`Error: Recipe parsing failed: Unsupported format. Expected JSON or YAML.`
and
`Error: Recipe parsing failed: Failed to parse recipe: could not find expected ':' at line 16 column 1, while scanning a simple key at line 13 column 1`

This PR adds the recipe content after parameters are substituted in, into the recipe parsing error to make it easier to debug.

New errors (fi parameter has alignment that breaks yaml):
```
Error: Recipe parsing failed: Failed to parse recipe: could not find expected ':' at line 16 column 1, while scanning a simple key at line 13 column 1

Raw recipe content:
version: 1.0.0
title: "404Portfolio"
description: "Create personalized, creative 404 pages using public profile data"
parameters:
  - key: fi
    input_type: file
    requirement: required
    description: "fi"
prompt: |
  Say Only "HI" to the user.

   Hi
There

 What's
Up

extensions:
  - type: builtin
    name: developer
  - type: builtin
    name: computercontroller
``` 

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Manually tested before and after with a valid and invalid recipe.

### DIFF
--- a/crates/goose/src/recipe/build_recipe/mod.rs
+++ b/crates/goose/src/recipe/build_recipe/mod.rs
@@ -15,8 +15,13 @@ pub enum RecipeError {
     MissingParams { parameters: Vec<String> },
     #[error("Template rendering failed: {source}")]
     TemplateRendering { source: anyhow::Error },
-    #[error("Recipe parsing failed: {source}")]
-    RecipeParsing { source: anyhow::Error },
+    #[error("Recipe parsing failed: {source}\n\nRaw recipe content:\n{content}")]
+    RecipeParsing {
+        source: anyhow::Error,
+        content: String,
+    },
+    #[error("Unknown sub-recipe path: {path}")]
+    UnknownSubRecipePath { path: String },
 }
 
 fn render_recipe_template<F>(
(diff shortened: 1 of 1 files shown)


lifeizhou-ap:
will this include all the recipe contents? The error is used in both cli and Desktop now. If the recipe is long, it will have a very long error message. Is the error message with line number helpful for debugging in CLI? 

The assumption is: the user has the recipe file and knows the recipe content already, so it does not need to print the whole contents in the error message.  But maybe there are some use case which requires the contents in the error message?

jsibbison-square:
> will this include all the recipe contents? The error is used in both cli and Desktop now. If the recipe is long, it will have a very long error message. Is the error message with line number helpful for debugging in CLI?
> 
> The assumption is: the user has the recipe file and knows the recipe content already, so it does not need to print the whole contents in the error message. But maybe there are some use case which requires the contents in the error message?

It includes the whole content with parameters injected. Just the line number is not useful because we need to see how the content is rendered into the recipe to understand why the format breaks the recipe.

DOsinga:
do we still want to get this in? @lifeizhou-ap if so can you have a look?

jsibbison-square:
Deferring unless others have demand for it

PR #5281, opened 2025-10-20
Agent extension api

DOsinga:
Remove most of the add to and remove from the code base and use the API. try to do the best for windows and showing things, but eh

### DIFF
--- a/crates/goose-server/src/routes/agent.rs
+++ b/crates/goose-server/src/routes/agent.rs
@@ -510,7 +510,7 @@ async fn agent_add_extension(
                     eprintln!("Node.js installation completed");
                 } else {
                     return Err(ErrorResponse::internal(format!(
-                        "Node.js installer script not found at: {}",
+                        "Node.js not detected and no installer script not found at: {}",
                         install_script.display()
                     )));
                 }


crates/goose-server/src/routes/agent.rs
```diff
@@ -451,6 +464,88 @@ async fn update_router_tool_selector(
...
+
+            if !node_exists {
+                let cmd_path = std::path::Path::new(&cmd);
+                let script_dir = cmd_path.parent().ok_or(StatusCode::INTERNAL_SERVER_ERROR)?;
+                let install_script = script_dir.join("install-node.cmd");
+
+                if install_script.exists() {
+                    eprintln!("Installing Node.js...");
+                    let output = std::process::Command::new(&install_script)
+                        .arg("https://nodejs.org/dist/v23.10.0/node-v23.10.0-x64.msi")
```
jamadeo, line 497: is there a better url we could use here, "stable" or something like that?

crates/goose-server/src/routes/agent.rs
```diff
@@ -451,6 +464,88 @@ async fn update_router_tool_selector(
...
+                    if !output.status.success() {
+                        return Err(ErrorResponse::internal(format!(
+                            "Failed to install Node.js: {}",
+                            String::from_utf8_lossy(&output.stderr)
+                        )));
+                    }
+                    eprintln!("Node.js installation completed");
+                } else {
+                    return Err(ErrorResponse::internal(format!(
+                        "Node.js installer script not found at: {}",
```
jamadeo: "node.js not detected on system, and no installer script found"

jamadeo (approved):

crates/goose-server/src/routes/agent.rs
DOsinga: pinning does feel better, actually. though I doubt we'll ever update this

PR #5287, opened 2025-10-21
fix: handle non exist subrecipe path

lifeizhou-ap:
## Summary
Issue: 
In Desktop UI is frozen after user enters the parameters of the recipe.  

Why: 
The subrecipe path in the main recipe does not exist

Root cause: 
when we load the recipe and the subrecipe path does not exist, the code skipped the error

Fix:  
When the recipe is ready to load (required parameters are available), check the subrecipe path and display a toast showing the error

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Unit test and manual testing

### DIFF
--- a/crates/goose-cli/src/recipes/extract_from_cli.rs
+++ b/crates/goose-cli/src/recipes/extract_from_cli.rs
@@ -240,7 +240,7 @@ settings:
   temperature: 0.7
 sub_recipes:
 - path: existing_sub_recipe.yaml
-  name: existing_sub_recipe        
+  name: existing_sub_recipe
 response:
   json_schema:
     type: object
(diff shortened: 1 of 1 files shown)


DOsinga:
is that related to https://github.com/block/goose/issues/5288 ? that last one might warrant a quicker fix

zanesq:
I don't think so actually this is just validation that the sub recipe is missing so not session related

DOsinga:
how did this freeze? that doesn't seem like a state we should be able to get in to at all

zanesq:
Freezing was not the best term it wasn't allowing the param form to submit due too a backend error from the missing sub recipe so it felt like the app was not doing anything when clicked, no error just wouldn't submit. The issue is the param entry form doesn't handle the backend error and show it to the user so nothing happened. We should follow up with better error handling for param entry form but Lifei's change circumvents it for now because the recipe won't validate and run now so the user would never reach the param form.

Similar issue now though currently its just a backend error that the recipe didn't load because its missing a sub recipe. We should bubble that up to the user so its clear why it won't load the recipe after clicking the run button from recipe library. Should also show this error message when loaded from a deeplink.

zanesq (approved):
Nevermind doesn't seem to be happening now gtg!

PR #5289, opened 2025-10-21
improvement: propagate McpErrors directly into ToolCallResult

alexhancock:
**Issue:**

Before this change, we wrapped all errors which occured during tool calling in a:

```
ErrorData::new(ErrorCode::INTERNAL_ERROR, e.to_string(), None)
```

There are some issues with this. Only the message (albeit with original code embedded) was included, the error itself has an incorrect code, and no additional information about the error in the third param.

**Change:**

This change propgates the wrapped `ErrorData` from `ServiceError::McpError`s during tool calling into the `ToolCallResult` so that the model has better information about what happened. These will contain the proper error codes, original message, and additional data where the mcp sdk provides it.

**Testing**

I tested with the following prompt:

> Call the text_editor str_replace tool to replace the header of the README.md, but leave out the old_str param, which should cause and let me test an arguments error. Tell me exactly what the error says

**Before (actual error wrapped and obscured):**
Screenshot 2025-10-21 at 10 28 48 AM

**After (actual error included directly):**
Screenshot 2025-10-21 at 10 24 24 AM

Sonnet 4.5 did a valiant effort trying to include some info about the wrapped error, but other models are likely to be confused.

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -940,7 +940,9 @@ impl ExtensionManager {
                 .map(|call| call.content)
                 .map_err(|e| match e {
                     ServiceError::McpError(error_data) => error_data,
-                    _ => ErrorData::new(ErrorCode::INTERNAL_ERROR, e.to_string(), e.maybe_to_value()),
+                    _ => {
+                        ErrorData::new(ErrorCode::INTERNAL_ERROR, e.to_string(), e.maybe_to_value())
+                    }
                 })
         };
 


crates/goose/src/agents/extension_manager.rs
```diff
@@ -937,7 +937,10 @@ impl ExtensionManager {
                 .call_tool(&tool_name, arguments, cancellation_token)
                 .await
                 .map(|call| call.content)
-                .map_err(|e| ErrorData::new(ErrorCode::INTERNAL_ERROR, e.to_string(), None))
+                .map_err(|e| match e {
+                    ServiceError::McpError(error_data) => error_data,
+                    _ => ErrorData::new(ErrorCode::INTERNAL_ERROR, e.to_string(), None),
```
alexhancock: still thinking a bit. may want to consider further expansion of these things too?

`e.maybe_to_value()` is another option for the third param instead of `None`

DOsinga: we can provide even more context in the optional data field. might be useful for debugging and also to send to the agent back

alexhancock: @DOsinga agreed. does https://github.com/block/goose/pull/5289/commits/8d7158d69cbda2013c8ebb637a5b8bbd93f00f06 look right to you for it?

DOsinga:
nice

PR #5290, opened 2025-10-21
Remove reliance on localstorage for pendingScheduleDeepLink when scheduling a recipe

zanesq:
## Summary
Follow up for https://github.com/block/goose/pull/5217 to remove reliance on localStorage for pendingScheduleDeepLink

### DIFF
--- a/ui/desktop/src/components/schedule/SchedulesView.tsx
+++ b/ui/desktop/src/components/schedule/SchedulesView.tsx
@@ -1,4 +1,5 @@
 import React, { useState, useEffect, useCallback, useMemo } from 'react';
+import { useLocation } from 'react-router-dom';
 import {
   listSchedules,
   createSchedule,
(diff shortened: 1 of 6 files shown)


ui/desktop/src/main.ts
```diff
@@ -991,8 +991,6 @@ ipcMain.on('react-ready', () => {
     log.info('No pending deep link to process');
   }
 
-  // We don't need to handle pending deep links here anymore
-  // since we're handling them in the window creation flow
```
DOsinga, line 995: good riddance

ui/desktop/src/components/schedule/SchedulesView.tsx
```diff
@@ -219,6 +222,7 @@ const SchedulesView: React.FC<SchedulesViewProps> = ({ onClose: _onClose }) => {
   const [isEditModalOpen, setIsEditModalOpen] = useState(false);
   const [editingSchedule, setEditingSchedule] = useState<ScheduledJob | null>(null);
   const [isRefreshing, setIsRefreshing] = useState(false);
+  const [pendingDeepLink, setPendingDeepLink] = useState<string | null>(null);
 
   // Individual loading states for each action to prevent double-clicks
```
DOsinga, line 227: this is pre-existing code below, but it feels like we should just have a state per schedule, no?

DOsinga (approved):

ui/desktop/src/components/schedule/SchedulesView.tsx
zanesq: yes good call, will come back to when we work on scheduler soon cc @amed-xyz

PR #5296, opened 2025-10-21
nextcamp - fix new chat from pair and reload

zanesq:
## Summary
Fixed new chat from pair to get sessionID and also reload from a new chat in pair by adding sessionID to the url after a chat starts


zanesq:
https://github.com/block/goose/pull/5237

Issue #5298, opened 2025-10-21
Using an unauthenticated cursor-agent returns opaque error


alexeyre:
**Describe the bug**

When using an unauthenticated cursor-agent the error returned is `Some(1)`, which doesn't indicate to the user the steps required to rectify the error. Although likely not a common scenario if the user is already using cursor-agent, goose should indicate when the configuration is incorrect.

**To Reproduce**
1. Run `cursor-agent logout` if logged in
2. Launch Goose with the cursor-agent provider
3. Attempt to start a chat

**Expected behavior**
Goose error should point the user towards the cursor authentication flow.

**Please provide following information:**
 - **OS & Arch:** MacOS 26.1 Beta (25B5062e)
 - **Interface:** UI
 - **Version:** v1.11.0
 - **Extensions enabled:** todo, developer, extension-manager
 - **Provider & Model:** cursor-agent auto

DOsinga:
I suppose we can close this now!

PR #5300, opened 2025-10-21
improvement: add useful error message when attempting to use unauthenticated cursor-agent

alexeyre:
## Summary
Adds logic to check the authentication status of cursor-agent via `cursor-agent status` in `get_authentication_status` and adds a check in `execute_command`. Previous behaviour was to output an error of `Some(1)` when trying to use cursor-agent while unauthenticated. 

### Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing

Manual testing with an unauthenticated cursor-agent, ran tests and linter.

### Related Issues
Relates to #5298

### DIFF
--- a/crates/goose/src/providers/cursor_agent.rs
+++ b/crates/goose/src/providers/cursor_agent.rs
@@ -259,12 +259,6 @@ impl CursorAgentProvider {
         messages: &[Message],
         _tools: &[Tool],
     ) -> Result<Vec<String>, ProviderError> {
-        if !self.get_authentication_status().await {
-            return Err(ProviderError::Authentication(
-                "You are not logged in to cursor-agent. Please run 'cursor-agent login' to authenticate first."
-                    .to_string()));
-        }
-
         let prompt = self.messages_to_cursor_agent_format(system, messages);
 
         if std::env::var("GOOSE_CURSOR_AGENT_DEBUG").is_ok() {
@@ -336,6 +330,11 @@ impl CursorAgentProvider {
         })?;
 
         if !exit_status.success() {
+            if !self.get_authentication_status().await {
+                return Err(ProviderError::Authentication(
+                    "You are not logged in to cursor-agent. Please run 'cursor-agent login' to authenticate first."
+                        .to_string()));
+            }
             return Err(ProviderError::RequestFailed(format!(
                 "Command failed with exit code: {:?}",
                 exit_status.code()


crates/goose/src/providers/cursor_agent.rs
```diff
@@ -98,6 +98,66 @@ impl CursorAgentProvider {
         None
     }
 
+    /// Check if the user is authenticated with cursor-agent CLI
```
DOsinga: this seems like a lot of code for what it does. can we do something like:

```rust
async fn get_authentication_status(&self) -> bool {
    Command::new(&self.command)
        .arg("status")
        .output()
        .await
        .ok()
        .map(|output| String::from_utf8_lossy(&output.stdout).contains("✓ Logged in as"))
        .unwrap_or(false)
}
```

crates/goose/src/providers/cursor_agent.rs
```diff
@@ -245,6 +305,25 @@ impl CursorAgentProvider {
...
+            }
+            Err(e) => {
+                tracing::warn!("Failed to check authentication status: {}", e);
+                // If we can't check auth status, we should fail rather than proceed
+                return Err(ProviderError::Authentication(
+                    format!("Unable to verify cursor-agent authentication status: {}. Please ensure cursor-agent CLI is installed and accessible.", e)
+                ));
+            }
+        }
+
```
DOsinga: is this the place to check this? if so, let's just call the other thing (which should just return a boolean) and return an error if it says false?

ahldr: agreed, initial implementation was overly complicated, I'll implement and test your version and commit

ahldr:
@DOsinga implemented your refactored version, the logic is correct but error message appears to be coming via the chat rather than the error dialogue I first encountered so I'm unsure whether this is appropriate to commit as-is. Not super familiar with the codebase yet so very possible I'm not throwing the error in the correct way, feedback very much appreciated!

alexeyre:
I don't think this is in a state to be merged, the performance of `cursor-agent status` is pretty slow so I don't think it should be called every single time `execute_command` is used. It should just be cached and re-visited if we encounter an error, but unsure if clutter like that belongs in the whole `CursorAgentProvider` struct. 
```
hyperfine "cursor-agent status"
Benchmark 1: cursor-agent status
  Time (mean ± σ):     790.8 ms ± 151.7 ms    [User: 318.9 ms, System: 46.3 ms]
  Range (min … max):   690.0 ms … 1198.2 ms    10 runs
```

crates/goose/src/providers/cursor_agent.rs
```diff
@@ -47,6 +47,17 @@ impl CursorAgentProvider {
...
 
+    /// Get authentication status from cursor-agent
+    async fn get_authentication_status(&self) -> bool {
+        Command::new(&self.command)
+            .arg("status")
+            .output()
+            .await
+            .ok()
+            .map(|output| String::from_utf8_lossy(&output.stdout).contains("✓ Logged in as"))
+            .unwrap_or(false)
```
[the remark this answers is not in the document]
ahldr: `cursor-agent status` returns an exit code of 0 and uses stdout rather than stderr regardless of if the user is logged in or not.

ahldr:
@DOsinga if this is still wanted I'm unable to run the required approving workflows to allow merging

DOsinga:
thank you @alexeyre !

PR #5303, opened 2025-10-21
chip away at the stringly typed

jamadeo:
Uses a struct for system prompt templating, and a new type for goose mode.


jamadeo:
I'll re-open once #5313 is in

PR #5305, opened 2025-10-21
add a system prompt snapshot test

jamadeo:
we should do more of them

### DIFF
--- a/crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__basic.snap
+++ b/crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__basic.snap
@@ -18,9 +18,12 @@ Extensions allow other applications to provide context to goose. Extensions conn
 tools.
 You are capable of dynamically plugging into new extensions and learning how to use them. You solve higher level
 problems using the tools in these extensions, and can interact with multiple at once.
-Use the search_available_extensions tool to find additional extensions to enable to help with your task. To enable
-extensions, use the enable_extension tool and provide the extension_name. You should only enable extensions found from
-the search_available_extensions tool.
+
+If the Extension Manager extension is enabled, you can use the search_available_extensions tool to discover additional
+extensions that can help with your task. To enable or disable extensions, use the manage_extensions tool with the
+extension_name. You should only enable extensions found from the search_available_extensions tool.
+If Extension Manager is not available, you can only work with currently enabled extensions and cannot dynamically load
+new ones.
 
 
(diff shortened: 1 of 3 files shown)


crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__basic.snap
```diff
@@ -0,0 +1,65 @@
...
+
+
+No extensions are defined. You should let the user know that they should add extensions.
+
+
+
+
+# Suggestion
+
+""
```
jamadeo, line 36: this is obviously kinda broken, but leaving this PR so we can see the fix when we fix it

crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__basic.snap
```diff
@@ -0,0 +1,65 @@
...
+source: crates/goose/src/agents/prompt_manager.rs
+expression: system_prompt
+---
+You are a general-purpose AI agent called goose, created by Block, the parent company of Square, CashApp, and Tidal.
+goose is being developed as an open-source software project.
+
+The current date is 1970-01-01 00:00:00.
+
+goose uses LLM providers with tool calling capability. You can be used with different language models (gpt-4o,
+claude-sonnet-4, o1, llama-3.2, deepseek-r1, etc).
```
DOsinga, line 11: that can probably go

DOsinga (approved):
nice work

crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__basic.snap
```diff
@@ -0,0 +1,65 @@
...
+
+Extensions allow other applications to provide context to goose. Extensions connect goose to different data sources and
+tools.
+You are capable of dynamically plugging into new extensions and learning how to use them. You solve higher level
+problems using the tools in these extensions, and can interact with multiple at once.
+
+If the Extension Manager extension is enabled, you can use the search_available_extensions tool to discover additional
+extensions that can help with your task. To enable or disable extensions, use the manage_extensions tool with the
+extension_name. You should only enable extensions found from the search_available_extensions tool.
+If Extension Manager is not available, you can only work with currently enabled extensions and cannot dynamically load
```
jamadeo, line 25: this also seems like we could do better. We shouldn't say "if tool x is available" because we know when we build this prompt what tools are available

PR #5306, opened 2025-10-21
Unify loading goose messages and usechatstream determines chat state

zanesq:
## Summary
Follow up for next camp to unify loading goose messages and usechatstream determines chat state

### DIFF
--- a/ui/desktop/src/types/message.ts
+++ b/ui/desktop/src/types/message.ts
@@ -3,6 +3,9 @@ import { Message, ToolConfirmationRequest, ToolRequest, ToolResponse } from '../
 export type ToolRequestMessageContent = ToolRequest & { type: 'toolRequest' };
 export type ToolResponseMessageContent = ToolResponse & { type: 'toolResponse' };
 
+// Compaction response message - must match backend constant
+const COMPACTION_THINKING_TEXT = 'goose is compacting the conversation...';
+
 export function createUserMessage(text: string): Message {
   return {
     id: generateMessageId(),
@@ -92,7 +95,7 @@ export function getCompactingMessage(message: Message | undefined): string | und
 
   for (const content of message.content) {
     if (content.type === 'systemNotification' && content.notificationType === 'thinkingMessage') {
-      if (content.msg.toLowerCase().includes('compact')) {
+      if (content.msg === COMPACTION_THINKING_TEXT) {
         return content.msg;
       }
     }


zanesq:
Put into draft while working with upstream changes

ui/desktop/src/types/message.ts
```diff
@@ -84,3 +84,19 @@ export function getThinkingMessage(message: Message | undefined): string | undef
...
 }
+
+export function getCompactingMessage(message: Message | undefined): string | undefined {
+  if (!message || message.role !== 'assistant') {
+    return undefined;
+  }
+
+  for (const content of message.content) {
+    if (content.type === 'systemNotification' && content.notificationType === 'thinkingMessage') {
+      if (content.msg.toLowerCase().includes('compact')) {
```
zanesq: Wasn't sure if there was a better way here, could change it to look for the exact message `goose is compacting the conversation...` matching the backend?

DOsinga: why is this of type thinking and not just of type compacting or some such?

zanesq: was wondering that also looks like we only have 2 `SystemNotificationType` from the backend `'thinkingMessage' | 'inlineMessage'` didn't think it was worth refactoring the backend at this point but can follow up after

ui/desktop/src/components/LoadingGoose.tsx
```diff
@@ -6,16 +6,25 @@ import { ChatState } from '../types/chatState';
...
+const LoadingGoose = ({
+  message,
+  chatState = ChatState.Idle,
+  isLoadingConversation = false,
+}: LoadingGooseProps) => {
   // Determine the appropriate message based on state
   const getLoadingMessage = () => {
     if (message) return message; // Custom message takes priority
 
+    if (isLoadingConversation && chatState === ChatState.Thinking) {
```
DOsinga: can't we make Loading a chat state?

ui/desktop/src/components/LoadingGoose.tsx
```diff
@@ -6,16 +6,25 @@ import { ChatState } from '../types/chatState';
...
+      return 'loading conversation...';
+    }
     if (chatState === ChatState.Thinking) return 'goose is thinking…';
     if (chatState === ChatState.Streaming) return 'goose is working on it…';
     if (chatState === ChatState.WaitingForUserInput) return 'goose is waiting…';
+    if (chatState === ChatState.Compacting) return 'goose is compacting the conversation...';
```
DOsinga: this should be a look up table

ui/desktop/src/components/LoadingGoose.tsx
```diff
@@ -33,6 +42,8 @@ const LoadingGoose = ({ message, chatState = ChatState.Idle }: LoadingGooseProps
           <FlyingBird className="flex-shrink-0" cycleInterval={150} />
         ) : chatState === ChatState.WaitingForUserInput ? (
           <AnimatedIcons className="flex-shrink-0" cycleInterval={600} variant="waiting" />
+        ) : chatState === ChatState.Compacting ? (
```
DOsinga: as should this

ui/desktop/src/types/message.ts
zanesq: ended up adding an exact match with the backend message I noticed we were doing that in other places

PR #5311, opened 2025-10-22
fix: skip temperature for goose-gpt-5 model

lifeizhou-ap:
## Summary
**Issue:** When specifying temperature with `goose-gpt-5` model (eg: run a recipe), the session exits with "bad request error"
**Why:** gpt-5 only allows default (1) temperature 
**Fix:** Ignore the temperature for `goose-gpt-5` (same thing has been handled in the openai)

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### DIFF
--- a/crates/goose/src/providers/formats/databricks.rs
+++ b/crates/goose/src/providers/formats/databricks.rs
@@ -625,7 +625,7 @@ pub fn create_request(
             }
         }
 
-        // o1 models use max_completion_tokens instead of max_tokens
+        // open ai reasoning models use max_completion_tokens instead of max_tokens
         if let Some(tokens) = model_config.max_tokens {
             let key = if is_openai_reasoning_model {
                 "max_completion_tokens"


jsibbison-square:
We should make the same change for the openai provider too: https://github.com/mnicholson110/goose/blob/88947546c8ef64f70bb2ea60c7b2774881f2db84/crates/goose/src/providers/formats/openai.rs#L625

lifeizhou-ap:
> We should make the same change for the openai provider too: https://github.com/mnicholson110/goose/blob/88947546c8ef64f70bb2ea60c7b2774881f2db84/crates/goose/src/providers/formats/openai.rs#L625

Hi @jsibbison-square,
In the main branch, as I mentioned in this PR description, the openai provider has already been handled the gpt-5 for skipping temperature. https://github.com/block/goose/blob/2b968d87740ebe10dc84a9a4a90f2eaac86929d5/crates/goose/src/providers/formats/openai.rs#L629

crates/goose/src/providers/formats/databricks.rs
```diff
@@ -625,7 +627,7 @@ pub fn create_request(
 
         // o1 models use max_completion_tokens instead of max_tokens
```
alexhancock: nit: update this comment too

alexhancock (approved):
nice!

crates/goose/src/providers/formats/databricks.rs
```diff
@@ -625,7 +625,7 @@ pub fn create_request(
             }
         }
 
-        // o1 models use max_completion_tokens instead of max_tokens
+        // open ai reasoning models use max_completion_tokens instead of max_tokens
```
DOsinga, line 628: do we need this at all? in fact shouldn't we just check whether we have max_tokens in there and if not check if there is max_completion_tokens in there?

lifeizhou-ap: By reading the open ai doc [here](https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_completion_tokens), max_token is deprecated.  

We may have a separate PR to use `max_completion_tokens` and test it.

PR #5315, opened 2025-10-22
fix: synchronize local message state after conversation compaction

katzdave:
## Summary

Fixes a critical bug where message visibility metadata was not properly synchronized between client and server after compaction, causing repeated compaction failures and context overflow issues.

## The Problem

When the server performs conversation compaction:
1. Server compacts 400 messages → sends `UpdateConversation` event with 10 messages (old ones marked `agentVisible=false`)
2. Client updates React state ✓
3. **Bug**: Local `currentMessages` variable still has 400 messages with `agentVisible=true` ✗
4. New messages get appended to the stale 400-message array
5. React state gets overwritten with incorrect metadata
6. Next request sends 400+ messages marked `agentVisible=true`
7. Server tries to compact again but context is too large → compaction fails

This created a persistent mismatch between client and server state where the client thought all 400 messages were visible to the agent, while the server had correctly marked most as `agentVisible=false`.

## The Fix

Synchronize the local `currentMessages` variable when `UpdateConversation` event is received in both streaming handlers:

- **`useChatStream.ts`**: Added `currentMessages = event.conversation` to sync local variable with compacted state
- **`useMessageStream.ts`**: Added `currentMessages = parsedEvent.conversation` to sync local variable with compacted state

This ensures that subsequent messages are appended to the **compacted** conversation with correct visibility metadata, not the stale 400-message array.

## Testing

- [x] Type checks pass
- [x] ESLint passes
- [x] Prettier formatting applied

## Impact

- **Before**: Repeated compaction failures, context overflow errors
- **After**: Proper state synchronization, compaction works reliably

### DIFF
--- a/ui/desktop/src/hooks/useChatStream.ts
+++ b/ui/desktop/src/hooks/useChatStream.ts
@@ -176,7 +176,6 @@ async function streamFromResponse(
             }
             case 'UpdateConversation': {
               log.messages('conversation-update', event.conversation.length);
-              // Synchronize local variable with compacted conversation
               currentMessages = event.conversation;
               // This calls the wrapped setMessagesAndLog with 'streaming' context
               updateMessages(event.conversation);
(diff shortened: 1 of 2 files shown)


katzdave:
OK my agent was a little eager, needs some more testing.

PR #5316, opened 2025-10-22
Set up Datadog metrics for prompt injection detection

dorien-koelemeijer:
## Summary
Adding some initial metrics to be consumed in Datadog. 

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [x] Other (specify below)
Observability improvement. 

### Related Issues
Related PR: https://github.com/block/goose/pull/4237/

### DIFF
--- a/crates/goose/src/security/mod.rs
+++ b/crates/goose/src/security/mod.rs
@@ -80,43 +80,38 @@ impl SecurityManager {
                 // Get threshold from config - only flag things above threshold
                 let config_threshold = scanner.get_threshold_from_config();
 
-                if analysis_result.is_malicious && analysis_result.confidence > config_threshold {
-                    // Generate a globally unique finding ID for each security finding
+                if analysis_result.is_malicious {
+                    let above_threshold = analysis_result.confidence > config_threshold;
                     let finding_id = format!("SEC-{}", Uuid::new_v4().simple());
 
                     tracing::warn!(
                         counter.goose.prompt_injection_finding = 1,
                         gauge.goose.prompt_injection_confidence_score = analysis_result.confidence,
-                        above_threshold = true,
+                        above_threshold = above_threshold,
                         tool_name = %tool_call.name,
                         tool_request_id = %tool_request.id,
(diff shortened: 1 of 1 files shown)


crates/goose/src/security/mod.rs
```diff
@@ -106,6 +115,9 @@ impl SecurityManager {
                     });
                 } else if analysis_result.is_malicious {
                     tracing::warn!(
+                        counter.goose.prompt_injection_finding = 1,
```
DOsinga: you're logging essentially the same thing on 76, 96 and 117 - I think you probably just want to log the same record just with different values; suggest to unify that.

dorien-koelemeijer: Having a look. I might just leave it for now since I want to play around with the metrics and have the feeling I'll be doing more updates anyway. Line 76 shouldn't send anything to Datadog for now, so I could potentially remove that.

dorien-koelemeijer: I think I may leave it as is for now, I'm planning to play around with metrics and logs a bit and will need to determine what will stay and needs to go longer term anyway, so can do one big uplift at that stage

DOsinga: that arguments rather cuts both ways though; why not simplify it for now while you figure out what you want?

dorien-koelemeijer: True, could do, not sure if it'll get simplified by a lot though, since I think it'll actually only be line 76 and 96 that will be simplified

dorien-koelemeijer: Fair enough, have cleaned up. Would you mind unblocking the 3 workflows checks below that require approval (I can't unblock these myself)? Thanks!

dorien-koelemeijer:
Created https://github.com/block/goose/pull/5385 instead (not from forked repo)

PR #5323, opened 2025-10-22
Diagnostics

DOsinga:
Makes it possible to share a diagnostics zip file with the team that contains the important details to help resolve an issue.

image

image

### DIFF
--- a/crates/goose-server/src/routes/status.rs
+++ b/crates/goose-server/src/routes/status.rs
@@ -30,7 +30,7 @@ async fn diagnostics(Path(session_id): Path<String>) -> impl IntoResponse {
                 ),
                 (
                     http::header::CONTENT_DISPOSITION,
-                    HeaderValue::from_str(&filename).map_err(|e| StatusCode::BAD_REQUEST)?,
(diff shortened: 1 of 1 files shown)


crates/goose-server/src/routes/status.rs
```diff
@@ -0,0 +1,46 @@
...
+        Ok(zip_data) => {
+            let filename = format!("attachment; filename=\"diagnostics_{}.zip\"", session_id);
+            let headers = [
+                (
+                    axum::http::header::CONTENT_TYPE,
+                    HeaderValue::from_static("application/zip"),
+                ),
+                (
+                    axum::http::header::CONTENT_DISPOSITION,
+                    HeaderValue::from_str(&filename).unwrap(),
```
jamadeo: you have an .unwrap here, I think it's highly unlikely to fail to make a header value out of your file name, but might as well map to a 500

jamadeo (approved):

crates/goose-server/src/routes/status.rs
DOsinga: yeah, we should really get around automatically doing that at some point

PR #5325, opened 2025-10-22
feat/fix: set optional config params, and don't overwrite unset secrets

jamadeo:
This changes the provider settings modals to:
- show optional config parameters in a default-hidden section
- don't resubmit values the user doesn't change
- show a masked version of secret values instead of nothing
- be simpler

OpenAI looks like

Screenshot 2025-10-22 at 11 22 59 PM

Screenshot 2025-10-22 at 11 23 28 PM

### DIFF
--- a/ui/desktop/src/components/settings/providers/modal/ProviderConfiguationModal.tsx
+++ b/ui/desktop/src/components/settings/providers/modal/ProviderConfiguationModal.tsx
@@ -175,7 +175,10 @@ export default function ProviderConfigurationModal({
             onDelete={handleDelete}
             showDeleteConfirmation={showDeleteConfirmation}
             onConfirmDelete={handleConfirmDelete}
-            onCancelDelete={() => {}}
+            onCancelDelete={() => {
+              setIsActiveProvider(false);
+              setShowDeleteConfirmation(false);
+            }}
             canDelete={isConfigured && !isActiveProvider}
             providerName={provider.metadata.display_name}
             isActiveProvider={isActiveProvider}
--- a/ui/desktop/src/components/settings/providers/modal/subcomponents/forms/DefaultProviderSetupForm.tsx
+++ b/ui/desktop/src/components/settings/providers/modal/subcomponents/forms/DefaultProviderSetupForm.tsx
@@ -51,24 +51,27 @@ export default function DefaultProviderSetupForm({
 
   const loadConfigValues = useCallback(async () => {
     setIsLoading(true);
-    const values: { [k: string]: ConfigInput } = {};
+    try {
+      const values: { [k: string]: ConfigInput } = {};
 
-    for (const parameter of parameters) {
-      const configKey = `${parameter.name}`;
-      const configValue = (await read(configKey, parameter.secret || false)) as ConfigValue;
+      for (const parameter of parameters) {
+        const configKey = `${parameter.name}`;
+        const configValue = (await read(configKey, parameter.secret || false)) as ConfigValue;
 
-      if (configValue) {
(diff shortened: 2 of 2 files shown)


ui/desktop/src/components/settings/providers/modal/ProviderConfiguationModal.tsx
```diff
@@ -175,7 +175,10 @@ export default function ProviderConfigurationModal({
             onDelete={handleDelete}
             showDeleteConfirmation={showDeleteConfirmation}
             onConfirmDelete={handleConfirmDelete}
-            onCancelDelete={() => {}}
+            onCancelDelete={() => {
```
zanesq, line 178: Do we want to change the provider's "active" state? Will this mark providers as inactive when the user simply cancels the confirmation dialog?

ui/desktop/src/components/settings/providers/modal/ProviderConfiguationModal.tsx
```diff
@@ -7,198 +7,123 @@ import {
...
   const [validationErrors, setValidationErrors] = useState<Record<string, string>>({});
   const { upsert, remove } = useConfig();
   const { getCurrentModelAndProvider } = useModelAndProvider();
-  const { isOpen, currentProvider, modalProps, closeModal } = useProviderModal();
-  const [configValues, setConfigValues] = useState<Record<string, string>>({});
+  const [configValues, setConfigValues] = useState<Record<string, ConfigInput>>({});
   const [showDeleteConfirmation, setShowDeleteConfirmation] = useState(false);
-  const [isActiveProvider, setIsActiveProvider] = useState(false); // New state for tracking active provider
-  const [requiredParameters, setRequiredParameters] = useState<ConfigKey[]>([]); // New state for tracking active provider
+  const [isActiveProvider, setIsActiveProvider] = useState(false);
```
DOsinga, line 36: oh, is this no longer a new state?

ui/desktop/src/components/settings/providers/modal/ProviderConfiguationModal.tsx
DOsinga: activeProvider is a shadow variable that is supposed to be true only for the provider that is the current one. I think we can safely get rid of the entire thing and just at the top read in what the current provider is and then everywhere where we check activeProvider, just check if that provider is the current provider. we're really playing fast and loose with setting and unsetting this as is

ui/desktop/src/components/ui/input.tsx
```diff
@@ -8,7 +8,7 @@ const Input = React.forwardRef<HTMLInputElement, React.ComponentProps<'input'>>(
       <input
         type={type}
         className={cn(
-          'flex h-9 w-full rounded-md border focus:border-border-strong hover:border-border-strong bg-background-default px-3 py-1 text-base transition-colors file:border-0 file:bg-transparent file:text-sm file:font-medium file:text-foreground placeholder:text-textPlaceholder focus-visible:outline-none disabled:cursor-not-allowed disabled:opacity-50 md:text-sm',
+          'flex h-9 w-full rounded-md border focus:border-border-strong hover:border-border-strong bg-background-default px-3 py-1 text-base transition-colors file:border-0 file:bg-transparent file:text-sm file:font-medium file:text-foreground placeholder:text-textPlaceholder placeholder:font-light focus-visible:outline-none disabled:cursor-not-allowed disabled:opacity-50 md:text-sm',
```
DOsinga, line 11: hmm?

DOsinga (approved):

PR #5330, opened 2025-10-22
fix: remove dependency on gsap library

maniksurtani:
## Summary
This change removes a dependency on the `gsap` library, used by ui/desktop. The library is used to animate text on the greeting screen on startup. 

The library is problematic - it is not open source. Instead, it uses this rather badly-written bespoke license:
https://gsap.com/community/standard-license/ This is not an open source license because it has prohibitions on certain competing uses and also it's not clear it even allows commercial distribution.

`gsap`'s highly limited use in this codebase can easily be achieved using the native JS Web Animations API. The Web Animations API is optimised at the browser level so this should perform better too, though that isn't the primary goal of this change.

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Manual testing 

No change in UI from before and after the change.

### DIFF
--- a/ui/desktop/src/hooks/use-text-animator.tsx
+++ b/ui/desktop/src/hooks/use-text-animator.tsx
@@ -110,6 +110,7 @@ export class TextAnimator {
   splitter!: TextSplitter;
   originalChars!: string[];
   activeAnimations: globalThis.Animation[] = [];
+  activeTimeouts: ReturnType<typeof setTimeout>[] = [];
 
   constructor(textElement: HTMLElement) {
     if (!textElement || !(textElement instanceof HTMLElement)) {
@@ -124,7 +125,7 @@ export class TextAnimator {
     this.splitter = new TextSplitter(this.textElement, {
       splitTypeTypes: ['words', 'chars'],
     });
-    this.originalChars = this.splitter.getChars().map((char) => char.innerHTML);
+    this.originalChars = this.splitter.getChars().map((char) => char.textContent || '');
   }
 
   animate() {
@@ -133,7 +134,7 @@ export class TextAnimator {
     const chars = this.splitter.getChars();
 
     chars.forEach((char, position) => {
-      const initialHTML = char.innerHTML;
+      const initialText = char.textContent || '';
 
       char.style.opacity = '1';
       char.style.display = 'inline-block';
@@ -142,19 +143,16 @@ export class TextAnimator {
       const animation = char.animate(
         [
           {
-            // 0%
             opacity: 1,
             color: '#666',
             fontFamily: 'Cash Sans Mono',
             fontWeight: '300',
           },
           {
-            // 50%
             opacity: 0.5,
             color: '#999',
           },
           {
-            // 100%
             opacity: 1,
             color: 'inherit',
             fontFamily: 'inherit',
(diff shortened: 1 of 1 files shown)


ui/desktop/src/hooks/use-text-animator.tsx
```diff
@@ -134,63 +134,76 @@ export class TextAnimator {
...
-              color: '#666', // Add color change
-            });
+      char.style.opacity = '1';
+      char.style.display = 'inline-block';
+      char.style.position = 'relative';
+
+      const animation = char.animate(
+        [
+          {
+            // 0%
```
DOsinga: remove these comments - I think they are confusing. they indicated transparency but are about opacity.

ui/desktop/src/hooks/use-text-animator.tsx
```diff
@@ -134,63 +134,76 @@ export class TextAnimator {
...
-          repeatDelay: 0.05, // Increased delay
-          delay: position * 0.03, // Reduced delay between chars
-          innerHTML: () => lettersAndSymbols[Math.floor(Math.random() * lettersAndSymbols.length)],
-          opacity: 1,
+        ],
+        {
+          duration: 300, // Total duration for all iterations
+          easing: 'ease-in-out',
+          delay: position * 30, // Stagger the start of each animation
+          iterations: 1, // We'll handle repeats manually to change innerHTML
```
DOsinga: this comment should also go

ui/desktop/src/hooks/use-text-animator.tsx
```diff
@@ -134,63 +134,76 @@ export class TextAnimator {
...
       );
+
+      this.activeAnimations.push(animation);
+
+      let iteration = 0;
+      const maxIterations = 2;
+
+      const animateCharacterChange = () => {
+        if (iteration < maxIterations) {
+          char.innerHTML = lettersAndSymbols[Math.floor(Math.random() * lettersAndSymbols.length)];
```
zanesq: can we try `textContent` here instead of `innerHTML`?

ui/desktop/src/hooks/use-text-animator.tsx
```diff
@@ -134,63 +134,76 @@ export class TextAnimator {
...
+          setTimeout(animateCharacterChange, 100);
+          iteration++;
+        } else {
+          char.innerHTML = initialHTML;
+        }
+      };
+
+      setTimeout(animateCharacterChange, position * 30);
```
zanesq: Can we track and clear these timeouts for cleanup purposes?

Something like 
```
activeTimeouts: number[] = [];
const id = window.setTimeout(...);
this.activeTimeouts.push(id);
```

In reset()
```
this.activeTimeouts.forEach(clearTimeout);
this.activeTimeouts = [];
```

michaelneale:
nice one, yeah good catch.

maniksurtani:
> Can we add respecting the users reduced-motion preferences?

@zanesq let's do that in a separate PR since it changes the behaviour. This PR won't change any behaviour but is just a refactor.

PR #5337, opened 2025-10-23
Create recipe dir on save

DOsinga:
From Discord, if there's no recipe dir, it fails

### DIFF
--- a/scripts/test_providers.sh
+++ b/scripts/test_providers.sh
@@ -15,7 +15,7 @@ fi
 SCRIPT_DIR=$(pwd)
 
 PROVIDERS=(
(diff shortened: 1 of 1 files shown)


scripts/test_providers.sh
```diff
@@ -15,7 +15,7 @@ fi
 SCRIPT_DIR=$(pwd)
 
 PROVIDERS=(
-  "openrouter:anthropic/claude-sonnet-4.5:qwen/qwen3-coder"
+  "openrouter:meta-llama/llama-3-70b-instruct:anthropic/claude-sonnet-4.5:qwen/qwen3-coder"
```
jamadeo: want to keep this?

DOsinga: yeah good catch

Issue #3538, opened 2025-07-20
Bug: Pre and post summarization shows same token count, token count not immediately updated


cbruyndoncx:
**Describe the bug**
Pre and post summarization shows same token count, only after next llm call this is updated

$ goose session -r
resuming session | provider: openrouter model: anthropic/claude-sonnet-4
    logging to /home/cb/.local/share/goose/sessions/20250720_141858.jsonl
    working directory: /home/cb/projects/github/goose

Goose is running! Enter your instructions, or try asking what goose can do.

**Context: ●●●●●●●○○○ 68% (136479/200000 tokens)**
**( O)> /summarize**
◇  Are you sure you want to summarize this conversation? This will condense the message history.
│  Yes
│
Summarizing conversation...
Conversation has been summarized.
Key information has been preserved while reducing context length.
**Context: ●●●●●●●○○○ 68% (136479/200000 tokens)**
Carine B (BE - CET TZ) — 14:45
**after the next call, it shows correctly 
Context: ●●●○○○○○○○ 26% (52824/200000 tokens)**

**Expected behavior**
After summarization; correctly reflecting the reduced token count - or an estimate or a message saying it will be updated after next call - anything that gives confidence that it worked

**Please provide following information:**
 - **OS & Arch:** Ubuntu 22.04 x86
 - **Interface:** CLI
 - **Version:** v1.1.3
 - **Extensions enabled:** developer
 - **Provider & Model:** openrouter anthropic/claude-sonnet-4

blackgirlbytes:
I am having this issue as well

DOsinga:
is fixed in main!

PR #5342, opened 2025-10-23
Stream token usage on every agent message

katzdave:
Token usage gets saved to the session every turn, so just pull that. 

- Some clientside renames to make more clear what each token means
- Make sure we always sum input + output tokens -> total tokens even if the provider doesn't provide

### DIFF
--- a/ui/desktop/src/hooks/useChatStream.ts
+++ b/ui/desktop/src/hooks/useChatStream.ts
@@ -61,7 +61,7 @@ interface UseChatStreamReturn {
   setRecipeUserParams: (values: Record<string, string>) => Promise<void>;
   stopStreaming: () => void;
   sessionLoadError?: string;
-  tokenState?: TokenState;
+  tokenState: TokenState;
 }
 
 function pushMessage(currentMessages: Message[], incomingMsg: Message): Message[] {
@@ -90,7 +90,7 @@ async function streamFromResponse(
   stream: AsyncIterable<MessageEvent>,
   initialMessages: Message[],
   updateMessages: (messages: Message[]) => void,
-  updateTokenState: (tokenState?: TokenState) => void,
+  updateTokenState: (tokenState: TokenState) => void,
   updateChatState: (state: ChatState) => void,
   onFinish: (error?: string) => void
 ): Promise<void> {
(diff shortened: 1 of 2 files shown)


crates/goose/src/session/session_manager.rs
```diff
@@ -44,6 +44,18 @@ pub struct Session {
...
+pub struct TokenState {
+    /// Current turn token counts
+    pub input_tokens: Option<i32>,
+    pub output_tokens: Option<i32>,
+    pub total_tokens: Option<i32>,
+    /// Accumulated token counts across all turns
+    pub accumulated_input_tokens: Option<i32>,
+    pub accumulated_output_tokens: Option<i32>,
+    pub accumulated_total_tokens: Option<i32>,
+}
```
DOsinga: this seems like a good idea to have, but I would expect then to also have a way to read this on a session and a way to set this on the builder? would probably clean up the rest of the code

zanesq:
@katzdave can you resolve the conflicts when you get a chance?

katzdave:
> @katzdave can you resolve the conflicts when you get a chance?

Done

ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -60,6 +61,7 @@ interface UseChatStreamReturn {
   setRecipeUserParams: (values: Record<string, string>) => Promise<void>;
   stopStreaming: () => void;
   sessionLoadError?: string;
+  tokenState?: TokenState;
```
zanesq: Should this be optional? When would tokenState not be available?

zanesq (approved):
LGTM tested locally and token count is much better.

One question: what happens when multiple agents/requests update the same session concurrently would it lose the updates?

zanesq:
I think this fixes https://github.com/block/goose/issues/3538 also in my testing

ui/desktop/src/hooks/useChatStream.ts
katzdave: Fixed; had thought it was more widely used.

katzdave:
> LGTM tested locally and token count is much better.
> 
> One question: what happens when multiple agents/requests update the same session concurrently would it lose the updates?

I think it should be fine, since all of the updates are just pulling the latest state from the session on every message.

> I think this fixes #3538 also in my testing

Nice! I think desktop will need one more small fix for this but, very close.

PR #5344, opened 2025-10-23
mcp resource tools should always be enabled

angelahning:
## Summary
The read_resource and list_resources are part of the extension manager, which can be toggled-off. We shouldn't do that as this is how goose make the MCP resources available to the models, and the only way for users to reach the MCP resources. This PR separates them into a different mcp client called "core", which we can add more core tools in later, such as the llm tool router tool.  

### Type of Change

- [x] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Added tests. See`core_extension_tests`

**Email**:

### DIFF
--- a/crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__one_extension.snap
+++ b/crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__one_extension.snap
@@ -39,20 +39,12 @@ and platform__list_resources on this extension.
 how to use this extension
 
 
-# LLM Tool Selection Instructions
-    Important: the user has opted to dynamically enable tools, so although an extension could be enabled, \
-    please invoke the llm search tool to actually retrieve the most relevant tools to use according to the user's messages.
-    For example, if the user has 3 extensions enabled, but they are asking for a tool to read a pdf file, \
-    you would invoke the llm_search tool to find the most relevant read pdf tool.
-    By dynamically enabling tools, you (goose) as the agent save context window space and allow the user to dynamically retrieve the most relevant tools.
-    Be sure to format a query packed with relevant keywords to search for the most relevant tools.
-    In addition to the extension names available to you, you also have platform extension tools available to you.
-    The platform extension contains the following tools:
-    - search_available_extensions
-    - manage_extensions
-    - read_resource
-    - list_resources
-    
+Important: the user has opted to dynamically enable tools, so although an extension could be enabled,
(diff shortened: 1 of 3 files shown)


crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__one_extension.snap
```diff
@@ -39,20 +39,12 @@ and platform__list_resources on this extension.
...
-    Be sure to format a query packed with relevant keywords to search for the most relevant tools.
-    In addition to the extension names available to you, you also have platform extension tools available to you.
-    The platform extension contains the following tools:
-    - search_available_extensions
-    - manage_extensions
-    - read_resource
-    - list_resources
-    
+Important: the user has opted to dynamically enable tools, so although an extension could be enabled,
+please invoke the llm search tool to actually retrieve the most relevant tools to use according to the user's messages.
```
alexhancock, line 43: If we include this, we would want to be consistent with `llm_search` vs the `llm search` that shows up here

crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__one_extension.snap
```diff
@@ -39,20 +39,12 @@ and platform__list_resources on this extension.
...
-    By dynamically enabling tools, you (goose) as the agent save context window space and allow the user to dynamically retrieve the most relevant tools.
-    Be sure to format a query packed with relevant keywords to search for the most relevant tools.
-    In addition to the extension names available to you, you also have platform extension tools available to you.
-    The platform extension contains the following tools:
-    - search_available_extensions
-    - manage_extensions
-    - read_resource
-    - list_resources
-    
+Important: the user has opted to dynamically enable tools, so although an extension could be enabled,
```
alexhancock, line 42: I'm not sure about these instructions. Shouldn't we just advertise all the tools and then include a more minimal message something like: 

"if you don't find anything relevant in the standard enabled tools you can search for others that may be relevant via the `llm_search` tool and then dynamically enable an extension that has that tool"

zanesq:
Checking if still in progress?

michaelneale:
I think can close this for now - re-open if want to look at it again. I think with new emerging code-mode will make a bunch of things work better.

Issue #5045, opened 2025-10-06
UI Scheduler Issues


zanesq:
**Describe the bug**

Scheduler functionality isn't running at the exact time and its not able to load the session from a run. Also when trying to setup a schedule using cron and run it manually I got a unexpected response format error.

**To Reproduce**
1. Create a schedule from a recipe
2. Run schedule
3. See errors

**Expected behavior**
It should launch at the specified time and it should allow loading the session information.

**Screenshots**
<img width="941" height="799" alt="Image" src="https://github.com/user-attachments/assets/fa46b3ba-f0fa-42b5-b488-692a3538eba0" />
<img width="1027" height="582" alt="Image" src="https://github.com/user-attachments/assets/a290ade7-b590-42ec-8594-10c112303927" />
<img width="831" height="870" alt="Image" src="https://github.com/user-attachments/assets/361fccef-148d-4020-9975-f627c2529c7d" />

**Please provide following information:**
 - **OS & Arch:** [Mac Arm]
 - **Interface:** [UI]
 - **Version:** [v1.9.1]
 - **Extensions enabled:** [Developer]
 - **Provider & Model:** [Claude 4]

**Additional context**

junweiz-block:
<img width="447" height="569" alt="Image" src="https://github.com/user-attachments/assets/b644916a-0a62-4ee0-95a1-eb674bc6b1cc" />
<img width="1022" height="876" alt="Image" src="https://github.com/user-attachments/assets/375f7c33-1495-4fce-829b-53042c87263b" />

I have a scheduler set up and it supposes to run at `9:48 am` daily, My local time zone is `Australia/Melbourne`, when the time arrives, the scheduler doesn't run.

When I manually trigger a run, the new run shows a created at timestamp that's in the future.

<img width="831" height="885" alt="Image" src="https://github.com/user-attachments/assets/ca867e51-998b-421e-ba53-d37746950c25" />

When I click that run, Goose is unable to load run detail

<img width="1024" height="670" alt="Image" src="https://github.com/user-attachments/assets/7b11fda6-1aae-492d-9b2b-5d2f37770693" />

Please provide following information:

OS & Arch: [Mac Arm]
Interface: [UI]
Version: [v1.9.1]
Extensions enabled: [Developer, Context 7, Fetch, Memory, Playwright, Slack, Websearch, Linear]
Provider & Model: [Claude 4 Sonnet]

Abhijay007:
[...]

PR #5346, opened 2025-10-23
Improve schedule execution visibility for recipes that failed

amed-xyz:
### Problem

When a recipe has no `prompt` field, the scheduler:

1. Creates a session in the database
2. Skips agent execution (logs warning)
3. Updates session with `schedule_id`
4. Returns "success" with empty session
5. Sessions with 0 messages don't appear in UI
6. **Users have no visibility that their schedule failed**

### Solution

Create visible failed sessions with error messages that show up in the UI, so users can see what went wrong.

### Note
This is primarily for backwards compatibility. The root cause issue is that we allow schedules to be created off recipes with empty prompts, but that'll be addressed in a follow up PR.

### Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Related Issues
Relates to https://github.com/block/goose/issues/5045

Before:  
Screenshot 2025-10-23 at 4 29 03 PM

After:   
Screenshot 2025-10-23 at 4 29 35 PM

### DIFF
--- a/crates/goose/src/scheduler.rs
+++ b/crates/goose/src/scheduler.rs
@@ -655,7 +655,8 @@ impl Scheduler {
                 schedule_sessions.push((session.id.clone(), session));
             }
         }
-        schedule_sessions.sort_by(|a, b| b.0.cmp(&a.0));
+        // Sort by created_at timestamp, newest first
+        schedule_sessions.sort_by(|a, b| b.1.created_at.cmp(&a.1.created_at));
 
         let result_sessions: Vec<(String, Session)> =
             schedule_sessions.into_iter().take(limit).collect();


iandouglas:
Question on an adjacent workflow: if a user turns a session into a recipe, or asks goose to make a recipe, does a prompt get added? It might seem odd for a user to ask goose to make them a recipe, then try to schedule it, and then see a message that it didn't run because there was no prompt. In the GitHub issue thread I think we had talked about having goose inject a prompt like "go follow the user instructions" if no prompt was present.

https://github.com/block/goose/issues/5045#issuecomment-3379243840

crates/goose/src/scheduler.rs
```diff
@@ -655,7 +655,8 @@ impl Scheduler {
                 schedule_sessions.push((session.id.clone(), session));
             }
         }
-        schedule_sessions.sort_by(|a, b| b.0.cmp(&a.0));
+        // Sort by created_at timestamp, newest first
+        schedule_sessions.sort_by(|a, b| b.1.created_at.cmp(&a.1.created_at));
```
amed-xyz, line 659: Newer sessions were at the bottom of the list, again causing a confusing UX

amed-xyz:
> Question on an adjacent workflow: if a user turns a session into a recipe, or asks goose to make a recipe, does a prompt get added? It might seem odd for a user to ask goose to make them a recipe, then try to schedule it, and then see a message that it didn't run because there was no prompt. In the GitHub issue thread I think we had talked about having goose inject a prompt like "go follow the user instructions" if no prompt was present.
> 
> [#5045 (comment)](https://github.com/block/goose/issues/5045#issuecomment-3379243840)

@iandouglas good point. Next thing I was planning to do was to explicitly validate recipe prompt presence for schedules and ask the user for it, but adding _"go follow the user instructions"_ sounds even simpler as long as the instructions provide enough guidance

amed-xyz:
@iandouglas any thoughts on https://github.com/block/goose/pull/5359 instead?

iandouglas:
> @iandouglas any thoughts on #5359 instead?

left a comment over there, yeah that sounds perfect.

amed-xyz:
Closing in favour of https://github.com/block/goose/pull/5359

PR #5352, opened 2025-10-24
Add bottom menu extension selection

zanesq:
## Summary
Added bottom menu extension selector gated by ALPHA flag for upcoming work

image

### DIFF
--- a/ui/desktop/src/components/bottom_menu/BottomMenuExtensionSelection.tsx
+++ b/ui/desktop/src/components/bottom_menu/BottomMenuExtensionSelection.tsx
@@ -6,7 +6,7 @@ import { Switch } from '../ui/switch';
 import { FixedExtensionEntry, useConfig } from '../ConfigContext';
 import { toggleExtension } from '../settings/extensions/extension-manager';
 import { toastService } from '../../toasts';
-import { getFriendlyTitle, getSubtitle } from '../settings/extensions/subcomponents/ExtensionList';
+import { getFriendlyTitle } from '../settings/extensions/subcomponents/ExtensionList';
 
 interface BottomMenuExtensionSelectionProps {
   sessionId: string;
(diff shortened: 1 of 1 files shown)


ui/desktop/src/components/bottom_menu/BottomMenuExtensionSelection.tsx
```diff
@@ -0,0 +1,159 @@
...
+      if (!aBundled && bBundled) return 1;
+
+      // Finally sort alphabetically within each group
+      return a.name.localeCompare(b.name);
+    });
+  }, [extensionsList]);
+
+  const fetchExtensions = useCallback(async () => {
+    await getExtensions(true);
+  }, [getExtensions]);
```
DOsinga: why wrap this in a callback? we can just call getExtensions where it is needed, no?

ui/desktop/src/components/bottom_menu/BottomMenuExtensionSelection.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+export const BottomMenuExtensionSelection = ({ sessionId }: BottomMenuExtensionSelectionProps) => {
+  const [searchQuery, setSearchQuery] = useState('');
+  const [isOpen, setIsOpen] = useState(false);
+  const { extensionsList, addExtension, getExtensions } = useConfig();
+
+  const extensions = useMemo(() => {
+    if (extensionsList.length === 0) {
+      return [];
+    }
```
DOsinga: I don't think checking this gets you anything

ui/desktop/src/components/bottom_menu/BottomMenuExtensionSelection.tsx
```diff
@@ -0,0 +1,159 @@
...
+  }, [extensions, searchQuery]);
+
+  const sortedExtensions = useMemo(() => {
+    return [...filteredExtensions].sort((a, b) => {
+      if (a.enabled === b.enabled) {
+        return a.name.localeCompare(b.name);
+      }
+      return a.enabled ? -1 : 1;
+    });
+  }, [filteredExtensions]);
```
DOsinga, line 83: we seem to have two sorts going on here. let's combine them into one. also the first sort looks at bundled which I don't think really gives us anything.  if we wanted to do this, prioritize anything that is of type {Builtin, Platform, Frontend} I would say

ui/desktop/src/components/bottom_menu/BottomMenuExtensionSelection.tsx
```diff
@@ -0,0 +1,159 @@
...
+        });
+
+        await fetchExtensions();
+      } catch (error) {
+        toastService.error({
+          title: 'Extension Error',
+          msg: `Failed to ${extensionConfig.enabled ? 'disable' : 'enable'} ${extensionConfig.name}`,
+          traceback: error instanceof Error ? error.message : String(error),
+        });
+        await fetchExtensions();
```
DOsinga: hmm, do we want to do this in the error clause? if it failed on 67 it will fail again. why do we need to fetch extensions again anyway?

ui/desktop/src/components/bottom_menu/ExtensionItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+  name: string;
+  title: string;
+  description: string;
+}
+
+interface ExtensionItemProps {
+  extension: Extension;
+  isEnabled: boolean;
+  onToggle: (enabled: boolean) => void;
+}
```
DOsinga: Let's flatten this. I also would drop the description. this is a compact view and the descriptions hardly ever add anything anyway

DOsinga (approved):

ui/desktop/src/components/bottom_menu/ExtensionItem.tsx
zanesq: my bad this wasn't even used after I refactored things, removed

PR #5354, opened 2025-10-24
Auto-compact Threshold UI improvements

katzdave:
- On the backend, don't treat disabled as 80%; disabled is off (but you can still compact with out of context error)

Client UI
- Properly loads state in the context window so multiple windows stay in sync (it's a global shared config)
- No longer has any UI around 'disabling it'. You can set it from 1-100%. Setting it to 100% still effectively disables auto-compact, but better communicates that your client will still compact itself if you run out of context. "Auto-compact at 100%"
- Fix big threshold dots not showing up at some thresholds.

Screenshot 2025-10-24 at 1 18 44 PM

### DIFF
--- a/crates/goose/src/context_mgmt/mod.rs
+++ b/crates/goose/src/context_mgmt/mod.rs
@@ -167,7 +167,7 @@ pub async fn check_if_compaction_needed(
     let usage_ratio = current_tokens as f64 / context_limit as f64;
 
     let needs_compaction = if threshold <= 0.0 || threshold >= 1.0 {
-        false
+        false // Auto-compact is disabled.
     } else {
         usage_ratio > threshold
     };


crates/goose/src/context_mgmt/mod.rs
```diff
@@ -167,7 +167,7 @@ pub async fn check_if_compaction_needed(
     let usage_ratio = current_tokens as f64 / context_limit as f64;
 
     let needs_compaction = if threshold <= 0.0 || threshold >= 1.0 {
-        usage_ratio > DEFAULT_COMPACTION_THRESHOLD
+        false
```
zanesq: looks a bit confusing to me, maybe a comment why we're disabling here?

zanesq (approved):
LGTM minor comment

PR #5355, opened 2025-10-24
No recursive subagents

tlongwell-block:


crates/goose/src/agents/agent.rs
```diff
@@ -664,6 +666,10 @@ impl Agent {
         Ok(())
     }
 
+    pub fn mark_as_subagent(&mut self) {
+        self.is_subagent = true;
```
DOsinga, line 670: this doesn't seem like a property that should be mutable

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -109,7 +109,7 @@ fn get_agent_messages(
         .map_err(|e| anyhow!("Failed to create a session for sub agent: {}", e))?;
 
         let agent = agent_manager
-            .get_or_create_agent(session.id.clone())
+            .get_or_create_agent(session.id.clone(), true) // true = is_subagent
```
DOsinga, line 112: don't do this

alexhancock:
@tlongwell-block Are you able to give thoughts or take a rev on this per @DOsinga's comment?

> I don't know about this approach. it feels heavy handed given what we are trying to do here. the real solution would be to make creating sub agents a real extension and then just not enable that extension for subagents

tlongwell-block:
> @tlongwell-block Are you able to give thoughts or take a rev on this per @DOsinga's comment?
> 
> > I don't know about this approach. it feels heavy handed given what we are trying to do here. the real solution would be to make creating sub agents a real extension and then just not enable that extension for subagents

I totally agree. This is hamfisted and not the right way.

I mocked up moving subagents/subrecipes tools to a platform extension and it it just a big change. Want to get https://github.com/block/goose/pull/5082 in before proposing something that significant.

This PR was just an option as an emergency fix if we needed one for a new release.

Since we seem to have figured out the endless subagent spawning root cause (no developer tools enabled, which caused goose to spawn endless subagents to try to do anything) we could just address that with a system prompt change or a tool description change.

tlongwell-block:
Closing this for now. Don't think this is a good idea

PR #5356, opened 2025-10-24
Configurable search paths

jamadeo:
This lets you set paths in your config to be appended to a PATH variable when Goose runs things. For now just extensions but perhaps we use this for other processes too, like cli providers?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2660,6 +2660,7 @@ dependencies = [
  "serde_yaml",
  "serial_test",
  "sha2",
+ "shellexpand",
  "sqlx",
  "sys-info",
  "temp-env",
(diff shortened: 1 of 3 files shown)


crates/goose/src/agents/extension_manager.rs
```diff
@@ -182,6 +183,12 @@ async fn child_process_client(
     command.process_group(0);
     #[cfg(windows)]
     command.creation_flags(CREATE_NO_WINDOW_FLAG);
+
+    command.env(
```
michaelneale, line 187: this overrides the main $PATH right? so user would need to know to set everything they need as not appending? (can the config include $PATH as an auto expanding var?)

jamadeo: command.env overrides, but the search_path_var() function will do the appending

michaelneale: ah ok, nice.

PR #5360, opened 2025-10-24
fix: --session-id shouldn't work without --resume, but --name should

### LINKED ISSUE #5044: Using `--name` Without `--resume` No Longer Works As Expected

wpfleger96:
## Summary
This PR fixes the regression where `--name` flag no longer worked without `--resume` flag, restoring the ability to create new sessions with custom names.

- Added CLI boundary validation to ensure `--session-id` can only be used with `--resume`, catching invalid combinations before session creation

### Implementation Details
- Validation for `--session-id` without `--resume` happens at CLI boundary in both Session and Run commands

### Behavior
- `--name` without `--resume`: Creates NEW session with custom name
- `--name` with `--resume`: Resumes existing session with that name (errors if not found)
- `--session-id` without `--resume`: Shows clear error message
- `--session-id` with `--resume`: Resumes existing session with that ID (errors if not found)

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Manually tested locally

### Related Issues
 Fixes #5044


crates/goose-cli/src/commands/bench.rs
```diff
@@ -33,10 +33,10 @@ impl BenchBaseSession for CliSession {
 }
 pub async fn agent_generator(
     requirements: ExtensionRequirements,
-    session_id: String,
+    _session_id: String,
 ) -> BenchAgent {
     let base_session = build_session(SessionBuilderConfig {
-        session_id: Some(session_id),
+        identifier: None,
```
DOsinga: what's happening here?

crates/goose-cli/src/cli.rs
```diff
@@ -64,29 +63,9 @@ struct Identifier {
         long_help = "Legacy parameter for backward compatibility. Extracts session ID from the file path (e.g., '/path/to/20250325_200615.
 jsonl' -> '20250325_200615')."
     )]
-    path: Option<PathBuf>,
+    pub path: Option<PathBuf>,
 }
 
-async fn get_session_id(identifier: Identifier) -> Result<String> {
```
DOsinga, line 70: Why do we need to move this? the whole Identifier is part of the CLI  command line stuff and should ideally not be used outside of it. Getting to a real session-id as quickly as possible seems like the right pattern

PR #5366, opened 2025-10-24
Removed drafts and agentIsReady in ChatInput

zanesq:
## Summary

- Removed drafts, the original purpose was a workaround pre new ui / routes when your draft prompt would get erased after navigating to the extensions page. We can bring this back in a better form by integrating with the session and support CLI also.
- Removed chat context / agentIsReady in ChatInput since it was redundant we already have a isLoading state that should cover this.


ui/desktop/src/App.tsx
```diff
@@ -620,13 +619,11 @@ export function AppInner() {
...
-        </HashRouter>
-        <AnnouncementModal />
-      </ModelAndProviderProvider>
-    </DraftProvider>
+    <ModelAndProviderProvider>
+      <HashRouter>
+        <AppInner />
+      </HashRouter>
+      <AnnouncementModal />
+    </ModelAndProviderProvider>
```
DOsinga, line 645: woohoo! two more to go! well, also this:

```
      <Suspense fallback={SuspenseLoader()}>
        <ConfigProvider>
          <ErrorBoundary>
            <App />
          </ErrorBoundary>
        </ConfigProvider>
      </Suspense>
```

PR #5367, opened 2025-10-24
feat(mcp): support sampling in a scoped way

alexhancock:
Implements basic support for MCP Sampling

* MCP servers can now request sampling and it will be handled by goose to route a sampling message to the provider for completion
* Works for desktop and CLI
* Does not yet implement human in the loop approval of these, which is a `SHOULD` for clients in the MCP spec, but added significant overhead to the initial implementation so the idea is to get a scoped and working implementation out and then iterate
* Sampling is supposed to be text, image, or audio. This does not handle audio, because interestingly goose does not handle audio message content. Can be a followup when we support it generally.

Demos of it working in CLI and Desktop with the [everything](https://github.com/modelcontextprotocol/servers/tree/main/src/everything) server the prompt

> Use the sampleLLM tool in the everything extension to have the mcp server request sampling from the model. Have the mcp server ask the model for a quote from the great gatsby

Screenshot 2025-10-24 at 5 28 49 PM

Screenshot 2025-10-24 at 5 31 13 PM

### DIFF
--- a/crates/goose/src/agents/mcp_client.rs
+++ b/crates/goose/src/agents/mcp_client.rs
@@ -1,21 +1,24 @@
-use rmcp::model::JsonObject;
+use crate::agents::types::SharedProvider;
+use rmcp::model::{Content, ErrorCode, JsonObject};
 /// MCP client implementation for Goose
 use rmcp::{
     model::{
         CallToolRequest, CallToolRequestParam, CallToolResult, CancelledNotification,
         CancelledNotificationMethod, CancelledNotificationParam, ClientCapabilities, ClientInfo,
-        ClientRequest, GetPromptRequest, GetPromptRequestParam, GetPromptResult, Implementation,
-        InitializeResult, ListPromptsRequest, ListPromptsResult, ListResourcesRequest,
-        ListResourcesResult, ListToolsRequest, ListToolsResult, LoggingMessageNotification,
+        ClientRequest, CreateMessageRequestParam, CreateMessageResult, GetPromptRequest,
+        GetPromptRequestParam, GetPromptResult, Implementation, InitializeResult,
+        ListPromptsRequest, ListPromptsResult, ListResourcesRequest, ListResourcesResult,
+        ListToolsRequest, ListToolsResult, LoggingMessageNotification,
         LoggingMessageNotificationMethod, PaginatedRequestParam, ProgressNotification,
         ProgressNotificationMethod, ProtocolVersion, ReadResourceRequest, ReadResourceRequestParam,
-        ReadResourceResult, RequestId, ServerNotification, ServerResult,
+        ReadResourceResult, RequestId, Role, SamplingMessage, ServerNotification, ServerResult,
     },
     service::{
-        ClientInitializeError, PeerRequestOptions, RequestHandle, RunningService, ServiceRole,
+        ClientInitializeError, PeerRequestOptions, RequestContext, RequestHandle, RunningService,
+        ServiceRole,
     },
     transport::IntoTransport,
-    ClientHandler, Peer, RoleClient, ServiceError, ServiceExt,
+    ClientHandler, ErrorData, Peer, RoleClient, ServiceError, ServiceExt,
 };
 use serde_json::Value;
 use std::{sync::Arc, time::Duration};
@@ -76,12 +79,17 @@ pub trait McpClientTrait: Send + Sync {
 
 pub struct GooseClient {
     notification_handlers: Arc<Mutex<Vec<Sender<ServerNotification>>>>,
+    provider: SharedProvider,
 }
 
 impl GooseClient {
-    pub fn new(handlers: Arc<Mutex<Vec<Sender<ServerNotification>>>>) -> Self {
+    pub fn new(
+        handlers: Arc<Mutex<Vec<Sender<ServerNotification>>>>,
+        provider: SharedProvider,
+    ) -> Self {
         GooseClient {
             notification_handlers: handlers,
+            provider,
         }
     }
 }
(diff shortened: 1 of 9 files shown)


crates/goose/src/agents/mcp_client.rs
```diff
@@ -1,22 +1,26 @@
...
     service::{
-        ClientInitializeError, PeerRequestOptions, RequestHandle, RunningService, ServiceRole,
+        ClientInitializeError, PeerRequestOptions, RequestContext, RequestHandle, RunningService,
+        ServiceRole,
     },
     transport::IntoTransport,
-    ClientHandler, Peer, RoleClient, ServiceError, ServiceExt,
+    ClientHandler, ErrorData as McpError, ErrorData, Peer, RoleClient, ServiceError, ServiceExt,
 };
+use schemars::_private::NoSerialize;
```
jamadeo: suspicious-looking import!

crates/goose/src/agents/extension_manager_extension.rs
```diff
@@ -246,7 +246,7 @@ impl ExtensionManagerClient {
         };
 
         let result = extension_manager
-            .add_extension(config)
+            .add_extension(config, None)
```
jamadeo: probably fine for now, but I suppose we will want to allow this to work too

jamadeo (approved):
nice!

crates/goose/src/agents/mcp_client.rs
```diff
@@ -127,10 +136,85 @@ impl ClientHandler for GooseClient {
...
+            .unwrap_or("You are a general-purpose AI agent called goose");
+
+        let (response, usage) = provider
+            .complete(system_prompt, &messages, &[])
+            .await
+            .map_err(|e| {
+                ErrorData::new(
+                    ErrorCode::INTERNAL_ERROR,
+                    "Unexpected error while completing the prompt",
+                    e.maybe_to_value(),
```
DOsinga: why maybe_to_value()? I think ti would return None here, we should probably return the relevant string though

alexhancock: Adjusted

crates/goose/src/agents/extension_manager_extension.rs
alexhancock: In theory I fixed this in the most recent commit, which changes the approach of how the extension manager has a reference to a provider. Didn't text explicitly for sampling through extension_manager managed extensions, but I think it may work now. I'll add a note to follow up on it.

crates/goose/src/agents/mcp_client.rs
alexhancock: It was needed for `e.maybe_to_value()` but no longer using that, so it's gone!

alexhancock:
@jamadeo @DOsinga Updated it to always uses the agent's current provider via a shared reference. Verified by the `llm_request` logs that it uses different models for sampling after switching providers.

crates/goose/src/agents/types.rs
```diff
@@ -9,6 +10,13 @@ use utoipa::ToSchema;
 /// Type alias for the tool result channel receiver
 pub type ToolResultReceiver = Arc<Mutex<mpsc::Receiver<(String, ToolResult<Vec<Content>>)>>>;
 
+/// This is used when we want to share the agent's current provider
+/// There are a lot of components to this definition so breaking it down:
+/// `Arc` enables shared ownership across threads or async contexts
+/// `Mutex` ensures mutable access is synchronized
+/// `Option` represents that a provider may or may not be set
+pub type SharedProvider = Arc<Mutex<Option<Arc<dyn Provider>>>>;
```
DOsinga, line 14: you know what I am going to say about ths comment

crates/goose/src/agents/mcp_client.rs
```diff
@@ -127,10 +135,87 @@ impl ClientHandler for GooseClient {
...
+            .await
+            .as_ref()
+            .ok_or(ErrorData::new(
+                ErrorCode::INTERNAL_ERROR,
+                "Could not use provider",
+                None,
+            ))?
+            .clone();
+
+        // go from MCP sampling messages to a version we can send to the provider
```
DOsinga: if you want to say that, use better variable names

crates/goose/src/agents/mcp_client.rs
```diff
@@ -127,10 +135,87 @@ impl ClientHandler for GooseClient {
...
+            .map_err(|e| {
+                ErrorData::new(
+                    ErrorCode::INTERNAL_ERROR,
+                    "Unexpected error while completing the prompt",
+                    Some(Value::from(e.to_string())),
+                )
+            })?;
+
+        // convert back to MCP messages
+        let response_content = if let Some(content) = response.content.first() {
```
DOsinga: only first?

crates/goose/src/agents/types.rs
alexhancock: But I wrote this one in an attempt to explain why it's such a crazy looking type! 😆 

I suppose it is self-evident if you're familiar with rust, but in many cases I don't know that developers working in goose and who may need to work with values of this type would understand all these pieces and why they are important in this context.

DOsinga: maybe - to me the comment is confusing because it explains what Arc does, but not why we have it twice. Explaining Mutex and Option seems rather too much.

I would take:
/// Double Arc allows cheap provider swaps while sharing concurrent access

if that is indeed the reasoning here

alexhancock: yes, that is the reason and a better comment - will update

alexhancock:
Draft of a couple different approaches to integration testing for sampling https://github.com/block/goose/pull/5456

Wanted to make a separate PR to get feedback on which one seems most valuable

@jamadeo @DOsinga

Issue #5048, opened 2025-10-07
any MCP server extensions I add to the goose config fail to start when running goose UI


shaunie2fly:
any MCP server extensions I add to the goose config fail to start when running goose UI .   Goose cli loads them without an issue.

Error activating extension: Failed to add extension configuration, error: ProcessExit(ProcessExit { stderr: "2025-10-07 06:26:46 - Starting npx setup script.\n2025-10-07 06:26:46 - Creating directory ~/.config/goose/mcp-hermit/bin if it does not exist.\n2025-10-07 06:26:46 - Changing to directory ~/.config/goose/mcp-hermit.\n2025-10-07 06:26:46 - Hermit binary already exists. Skipping download.\n2025-10-07 06:26:46 - setting hermit cache to be local for MCP servers\n2025-10-07 06:26:46 - Updated PATH to include ~/.config/goose/mcp-hermit/bin.\n2025-10-07 06:26:46 - Checking for hermit in PATH.\n2025-10-07 06:26:46 - Initializing hermit.\nfatal:hermit: open /home/shaunbotsis/.config/goose/mcp-hermit/bin/hermit: text file busy\n2025-10-07 06:26:46 - An error occurred. Exiting with status 1.\n", source: ConnectionClosed("initialize response") })

(base) shaun@homepc-big:~$ uname -a
Linux homepc-big 6.16.3-76061603-generic #202508231538~1758561135~22.04~171c8de SMP PREEMPT_DYNAMIC Mon S x86_64 x86_64 x86_64 GNU/Linux

**To Reproduce**
1. add a extension 
2. start goose
3. See error

**Expected behavior**
The extension should install 

**Please provide following information:**
 - **OS & Arch:** [e.g. Ubuntu 22.04 x86]
 - **Interface:** [UI/CLI]
 - **Version:** [e.g. v1.0.2]
 - **Extensions enabled:** [e.g. Computer Controller, Figma]
 - **Provider & Model:** [e.g. Google - gemini-1.5-pro]

**Additional context**

DOsinga:
the error message suggests that hermit is in use and that's why this fails. how many extensions do you have activated? also can you do:

    ps aux | grep hermit

and see if there's anything stale?

IA386:
I think it is related to: #2351

michaelneale:
thought: I think we shoudln't use hermit/shims at all on linux - anyone disagree? Linux is more likely to have them ready at hand.

effndc:
[...]

Issue #3030, opened 2025-06-23
Unable to add extensions to Goose


alicecoulthard:
My Goose extensions are not sticking - when selections are remade I am getting this error

Error activating extension: Failed to add extension configuration, error: Initialization(Stdio { name: "googlecalendar", cmd: "/Applications/Goose.app/Contents/Resources/bin/uvx", args: ["mcp_gcal@latest"], envs: Envs { map: {} }, env_keys: [], timeout: Some(300), description: None, bundled: None }, McpServerError { method: "initialize", server: "", source: ServerBoxError(StdioProcessError("Unable to download extension, ensure WARP is on or try re-authenticating.")) })

Extensions requested
Google Calendar
Image Generator
Pdfreader
Websearch

DOsinga:
this should be fixed on main by now. feel free to reopen if not

reefer42:
adding extensions not working on debian or kali. it just relaunches the client with nothing new happening.

wdawn:
ubuntu, etensions wont install. 

Error activating extension: Failed to add extension configuration, error: ProcessExit(ProcessExit { stderr: "2025-09-16 21:44:45 - Starting uvx setup script.\n2025-09-16 21:44:45 - Creating directory ~/.config/goose/mcp-hermit/bin if it does not exist.\n2025-09-16 21:44:45 - Changing to directory ~/.config/goose/mcp-hermit.\n2025-09-16 21:44:45 - Hermit binary already exists. Skipping download.\n2025-09-16 21:44:45 - setting hermit cache to be local for MCP servers\n2025-09-16 21:44:45 - Updated PATH to include ~/.config/goose/mcp-hermit/bin.\n2025-09-16 21:44:45 - Checking for hermit in PATH.\n2025-09-16 21:44:45 - Initializing hermit.\nfatal:hermit: open /home/user/.config/goose/mcp-hermit/bin/hermit: text file busy\n2025-09-16 21:44:45 - An error occurred. Exiting with status 1.\n", source: ConnectionClosed("initialize response") })

Issue #2351, opened 2025-04-25
Bug: Unable to add extension on linux


Jax-Tsai-zero:
```bash
Error: Error activating extension: Failed to add extension configuration, error: Initialization(Stdio { name: "pdfreader", cmd: "/usr/lib/goose-desktop/resources/bin/uvx", args: ["mcp-read-pdf"], envs: Envs { map: {} }, timeout: Some(300), description: None, bundled: None }, McpServerError { method: "initialize", server: "", source: ServerBoxError(StdioProcessError("fatal:hermit: open /home/Jax-Tsai/.config/goose/mcp-hermit/bin/hermit: text file busy\n")) })
```

platform: arch linux

dbirks:
Hey thanks for opening this issue. I was having troubles with this too, and just recently figured out a fix for it. Just pushed it to the AUR in this commit: https://aur.archlinux.org/cgit/aur.git/commit/?h=goose-desktop&id=bb616687159479ddaae952f80a5364ae720a7f4d

ahau-square:
@Jax-Tsai-zero are you still seeing this issue? 

Thanks for the fix @dbirks !

phlax:
im hitting this on a debian host - is there no way to avoid hermit - id never heard of it but afaict its not somethign i want at all

phlax:
ah - i see - just replace the hermit scripts - wfm - thanks @dbirks

gbtami:
dbirk suggestion worked for me as well on Ubuntu
I suggest to devs to add uv and npm packages as required dependencies of .deb/.rpm packages.

Hmm, uv is so new it has no .deb in debian/ubuntu at all.

ryzen3100:
@dbirks 
Thanks it worked for Ubuntu distrobox container (Arch host).
```
sudo rm "/usr/lib/goose/resources/bin/npx"
sudo ln -s /usr/bin/npx "/usr/lib/goose/resources/bin/npx"
```

samrocksc:
On ubuntu, running into the same issue, the symlink works though.  Seems like it's an electron issue.

harkamals:
> [@dbirks](https://github.com/dbirks) Thanks it worked for Ubuntu distrobox container (Arch host).
> 
> ```
> sudo rm "/usr/lib/goose/resources/bin/npx"
> sudo ln -s /usr/bin/npx "/usr/lib/goose/resources/bin/npx"
> ```

thank you!

airtonix:
[...]

PR #5372, opened 2025-10-25
fix hermit `text file busy` issues on linux

The-Best-Codes:
## Summary
For several months, Linux users have frequently been unable to add STDIO MCP extensions (e.g., extensions using the `npx` command, etc.). Obviously this is a **key functionality** and affects a **lot** of MCP servers!

The goose setup script fails on Linux with a `text file busy` error. I suspect this is because Hermit tries to self-update its running binary, which Linux locks, unlike macOS. The fix uses a temporary `bin/hermit` copy on Linux to prevent lock conflicts, redirects activation output to a log file for clean stderr, and skips these changes on macOS to preserve existing behavior.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
I've manually tested this on Linux. **I need a macOS user to test these changes.** Steps to test on macOS:

1. Checkout the PR locally or open it in a codespace. To check out the PR locally, ensure you have the GitHub CLI installed, then run `gh pr checkout 5372` in the `goose` directory. To open the PR in a codespace, view the PR in GitHub and click `Code` in the top right of the PR, then `Codespaces > Create codespace on best/fix-hermit-lock`.
2. Disable & enable a STDIO MCP server. You can do this in goose desktop by running `just run-ui` from the `goose` directory to start the desktop UI, then navigating to `Extensions` in the sidebar. Once there, find any MCP server that uses an STDIO command such as `npx`, disable it, and enable it again. If you don't see any MCP servers that use `npx`, add a new one, `npx -y @angiejones/mcp-selenium` for example. If you see an error, please report it to me! If not, everything should be fine. Play around with goose a bit to make sure.
3. If you are using a codespace, you will need to test using the goose CLI.

### Related Issues
Relates to #2351, #5048, #3030

### DIFF
--- a/ui/desktop/src/bin/node-setup-common.sh
+++ b/ui/desktop/src/bin/node-setup-common.sh
@@ -57,19 +57,33 @@ log "Updated PATH to include ~/.config/goose/mcp-hermit/bin."
 log "Checking for hermit in PATH."
 which hermit >> "$LOG_FILE"
 
-# Fix hermit self-update lock issues on Linux
-if [[ "$(uname -s)" == "Linux" ]]; then
-    log "Creating temp dir with bin subdirectory for hermit copy to avoid self-update locks and bin directory issues."
-    HERMIT_TMP_DIR="/tmp/hermit_tmp_$$/bin"
-    mkdir -p "$HERMIT_TMP_DIR"
-    cp ~/.config/goose/mcp-hermit/bin/hermit "$HERMIT_TMP_DIR/hermit"
-    chmod +x "$HERMIT_TMP_DIR/hermit"
-    export PATH="$HERMIT_TMP_DIR:$PATH"
-fi
+# Check if hermit environment is already initialized (only run init on first setup)
+if [ ! -f bin/activate-hermit ]; then
+    log "Hermit environment not yet initialized. Setting up hermit."
 
-# Initialize hermit
-log "Initializing hermit."
-hermit init >> "$LOG_FILE"
+    # Fix hermit self-update lock issues on Linux by using temp binary for init only
+    if [[ "$(uname -s)" == "Linux" ]]; then
+        log "Creating temp dir with bin subdirectory for hermit copy to avoid self-update locks."
+        HERMIT_TMP_DIR="/tmp/hermit_tmp_$$/bin"
+        mkdir -p "$HERMIT_TMP_DIR"
+        cp ~/.config/goose/mcp-hermit/bin/hermit "$HERMIT_TMP_DIR/hermit"
+        chmod +x "$HERMIT_TMP_DIR/hermit"
+        export PATH="$HERMIT_TMP_DIR:$PATH"
+        HERMIT_CLEANUP_DIR="/tmp/hermit_tmp_$$"
+    fi
+
+    # Initialize hermit
+    log "Initializing hermit."
+    hermit init >> "$LOG_FILE"
+
+    # Clean up temp dir if it was created
+    if [[ -n "${HERMIT_CLEANUP_DIR:-}" ]]; then
+        log "Cleaning up temporary hermit binary directory."
+        rm -rf "$HERMIT_CLEANUP_DIR"
+    fi
+else
+    log "Hermit environment already initialized. Skipping init."
+fi
 
 # Activate the environment with output redirected to log
 if [[ "$(uname -s)" == "Linux" ]]; then
@@ -80,11 +94,6 @@ fi
 # Install Node.js using hermit
 log "Installing Node.js with hermit."
 hermit install node >> "$LOG_FILE"
-
-# Clean up temp dir
-if [[ "$(uname -s)" == "Linux" ]]; then
-    rm -rf "/tmp/hermit_tmp_$$"
-fi
 
 # Verify installations
 log "Verifying installation locations:"


The-Best-Codes (commented):
Ready for review!

The-Best-Codes:
:warning: This PR is ready for review, but **do not merge it**. It needs to be tested on macOS.

IA386:
This issue also affects **jbang** and **uvx** scripts.
The implementation should only run if **Hermit** is not installed — inside this block:

```bash
# Check if Hermit binary exists and download if not
if [ ! -f ~/.config/goose/mcp-hermit/bin/hermit ]; then
```

The **“text file busy”** lock issue occurs only on the first start, when Hermit is not yet installed.

The-Best-Codes:
@IA386 Yep I'm aware that is affects `uvx` STDIO servers as well… haven't tested with `jbang` but it should resolve issues with it as well.

> The “text file busy” lock issue occurs only on the first start, when Hermit is not yet installed.

This is what I thought at first but not what I experienced in my testing, possibly because cleanup doesn't happen properly on failure, and it gets stuck in a “locked” state after that? Not sure.

IA386:
At least to my understanding of the issue, there’s no cleanup process involved — the binary just dies. Your solution won’t work if someone tries to run it for the first time with a uvx-based MCP or jbang (haven’t seen any MCPs available, but I haven’t really looked either).

So in its current form, this PR doesn’t fully address the issue across different environments unless those scripts are updated as well.

Regarding the implementation — I don’t like that it will execute every time. On a successful install, subsequent runs will still copy an already replaced bin/hermit script.

If I can make a suggestion — please see my comment on #2351. I’m not completely happy with that approach either, but it only triggers once, while the Hermit environment is not yet bootstrapped.

IMHO, the best solution would be to keep the binary with its original name inside the mcp-hermit folder and leave it as is.

ui/desktop/src/bin/node-setup-common.sh
```diff
@@ -57,14 +57,35 @@ log "Updated PATH to include ~/.config/goose/mcp-hermit/bin."
 log "Checking for hermit in PATH."
 which hermit >> "$LOG_FILE"
 
+# Fix hermit self-update lock issues on Linux
```
alexhancock: It seems this introduced a second copy of hermit in a temp dir only for init/activate.

Could we just replace the copy and all associated logic in `~/.config/goose/mcp-hermit/` with this approach altogether? Everything would become a lot simpler in this script it seems to me if we did that.

The-Best-Codes: So you're saying like remove the persistent directory (`~/.config/goose/mcp-hermit/`) altogether and just use the temp directory approach for the binary? I could see that. I'll try it out and see how it works.

alexhancock: Yes. I'd go all persistent if we can fix issues associated with it, or if not then consider an approach where we always use a temporary one. Having a second copy temporarily to do init + activate feels strange to me.

The-Best-Codes: @alexhancock I tested both all-persistent and all-temporary approaches, but they all fail on Linux at hermit init
due to file locking.
All-persistent fails because hermit can't execute init when the binary is locked by the kernel; all-temp method fails because hermit's persistent state gets deleted on cleanup which causes errors.

If you had a different approach in mind LMK.

alexhancock: Thanks for testing.

I guess I don't understand how only having a temp directory for init + activate and then removing it and having another directory in perpetuity works

The-Best-Codes: @alexhancock Basically this is the flow on Linux:

1. The temp binary copy is only for safe execution during initialization
2. The persistent directory and binary remain untouched
3. Each run on Linux creates a new temp copy, then deletes it after setup
4. The persistent directory is where the actual state/tools are cached

So if I understand you correctly, you are concerned about step 3? And suggesting maybe creating a temp copy just once and then reusing that temp copy? I will test that out as well :+1:

rinormaloku:
Looking forward to a fix on this issue, it is making this unusable in linux.

ui/desktop/src/bin/node-setup-common.sh
The-Best-Codes: @alexhancock I updated the logic so that it only uses the temp dir once (when activating hermit) and never after that. It appears to be working on my system! Can you test on macOS, please?

The-Best-Codes (commented):
Changes are ready for review.

alexhancock:
Nice. Thanks for the revisions on this @The-Best-Codes

rinormaloku:
Note: After upgrading I had to delete the mcp hermit directory and then it worked:  `rm -rf ~/.config/goose/mcp-hermit`

The-Best-Codes:
@rinormaloku Yep :100: 
I was waiting for the next official release of goose so I could clarify that in the Discord. I should have clarified it here. Thanks!

rinormaloku:
Thanks for contributing this fix @The-Best-Codes

PR #5391, opened 2025-10-27
Filter preserved user messages to be text only.

katzdave:
Avoids orphaning a tool call.

In the case of out of context error we keep the last user request.

Been testing with the proxy + this query

`run ls, get the 3rd biggest file. Count the words. Mod the number of words with the number of files. Get another file and do the same. Run this loop 7 times.`

### DIFF
--- a/crates/goose/src/context_mgmt/mod.rs
+++ b/crates/goose/src/context_mgmt/mod.rs
@@ -78,8 +78,7 @@ pub async fn compact_messages(
     };
 
     // Check if the most recent message is a user message with text content only
-    let (messages_to_compact, preserved_user_text) = if let Some(last_message) = messages.last()
-    {
+    let (messages_to_compact, preserved_user_text) = if let Some(last_message) = messages.last() {
         if matches!(last_message.role, rmcp::model::Role::User) && has_text_only(last_message) {
             // Remove the last user message before compaction and preserve its text
             (&messages[..messages.len() - 1], extract_text(last_message))


crates/goose/src/context_mgmt/mod.rs
```diff
@@ -42,19 +42,26 @@ pub async fn compact_messages(
...
         } else if preserve_last_user_message {
-            // Last message is not a user message, but we want to preserve the most recent user message
-            // Find the most recent user message and copy it (don't remove from history)
+            // Last message is not a user message with text, but we want to preserve the most recent user message with text
+            // Find the most recent user message with text content and copy it (don't remove from history)
             let most_recent_user_message = messages
                 .iter()
                 .rev()
-                .find(|msg| matches!(msg.role, rmcp::model::Role::User))
+                .find(|msg| matches!(msg.role, rmcp::model::Role::User) && has_text_content(msg))
```
jamadeo: we should check that the message does not contain a tool result, not that it contains text (I'm not sure how likely in practice, but you could have both)

katzdave: Makes sense. Maybe check for both, or something very specific? contains exactly 1 text and no tool result?

jamadeo: the most important thing is that we don't keep the tool result around without a tool request, because that will trigger 400s from providers. So, yeah, anything that includes that check is probably sufficient for a quick fix

katzdave: Done

DOsinga: I would say, don't preserve messages at all. extract the text you want to keep and construct a new message. that way we avoid all the weirdness

crates/goose/src/context_mgmt/mod.rs
```diff
@@ -42,19 +42,34 @@ pub async fn compact_messages(
 
     let messages = conversation.messages();
 
-    // Check if the most recent message is a user message
+    // Helper to check if a message has text content ONLY (no tool requests/responses)
```
DOsinga: remove the comment

crates/goose/src/context_mgmt/mod.rs
```diff
@@ -290,3 +305,258 @@ fn format_message_for_compacting(msg: &Message) -> String {
         format!("[{}]: {}", role_str, content_parts.join("\n"))
     }
 }
+
+#[cfg(test)]
```
DOsinga: These tests read like LLM generated tests and verify your implementation, not the actual breakage. Can we replace this with what Jack had? What we want to test here is a series of scenarios where we call compaction and then verify that after compaction the assistant visible messages form a conversation that is valid according to the conversation fixer. And those scenarios should reflect the actual breakagaes that we've seen over the last few months and be extendable any time we see a new breakage.

crates/goose/src/context_mgmt/mod.rs
katzdave: done

crates/goose/src/context_mgmt/mod.rs
katzdave: done

crates/goose/src/context_mgmt/mod.rs
katzdave: done

crates/goose/src/context_mgmt/mod.rs
```diff
@@ -42,21 +42,55 @@ pub async fn compact_messages(
...
-            let most_recent_user_message = messages
+            // Last message is not a user message with text only, but we want to preserve the most recent user message with text only
+            // Find the most recent user message with text content only and extract its text
+            let preserved_text = messages
                 .iter()
                 .rev()
-                .find(|msg| matches!(msg.role, rmcp::model::Role::User))
-                .cloned();
-            (messages.as_slice(), most_recent_user_message)
+                .find(|msg| matches!(msg.role, rmcp::model::Role::User) && has_text_only(msg))
```
jamadeo, line 91: is this definitely what we want though? that user message might be pretty far back in a tool calling loop...

for a patch, should we maybe just include the last user message *iff* it is a text, and do nothing if it isn't? that feels more conservative, and we can explore other ways to improve the quality in main

DOsinga: I think that might have been better, but I'd rather get something out sooner than later. we can then make a plan of what we really want

katzdave: merging as is since this mirrors the intended previous behavior, but lets keep the discussion open as to what exactly we want.

The other thing we might need to worry about if we continue the loop is with smaller models it could be possible to get stuck in an infinite loop.

PR #5392, opened 2025-10-27
fix compaction in tool loops

jamadeo:


crates/goose/src/context_mgmt/mod.rs
```diff
@@ -17,55 +16,85 @@ struct SummarizeContext {
...
+        .with_text(CONVSERATION_CONTINUATION_TEXT)
+        .with_metadata(MessageMetadata::agent_only())
+}
+
+fn tool_loop_continuation_message(last_user_message: Option<&Message>) -> Message {
+    let prompt = if let Some(user_message) = last_user_message {
+        format!(
+            "{}\nThe last user message was:\n\n{}",
+            TOOL_LOOP_CONTINUATION_TEXT,
+            user_message.as_concat_text()
```
katzdave, line 54: seems like a good way to include the last user message 👍

crates/goose/src/context_mgmt/mod.rs
```diff
@@ -17,55 +16,85 @@ struct SummarizeContext {
...
-        } else if preserve_last_user_message {
-            // Last message is not a user message, but we want to preserve the most recent user message
-            // Find the most recent user message and copy it (don't remove from history)
+    // split the conversation into messages to be compacted, and continuation messages
+    let (messages_to_compact, continuation_messages) = match messages.last() {
+        Some(message) if is_user_submitted_text(message) => (
+            &messages[..messages.len() - 1],
+            vec![conversation_continuation_message(), message.clone()],
+        ),
+        Some(message) if is_tool_response_message(message) => {
```
katzdave, line 84: This check on is_tool_response_message feels a little iffy to me.

For manual compact, I guess it won't happen under normal circumstances, but is that guaranteed for the case where the agent loop is running and runs out of context?

Would consider passing a boolean is_manual_compact passed into compact_messages.

crates/goose/src/context_mgmt/mod.rs
```diff
@@ -101,25 +130,15 @@ pub async fn compact_messages(
...
-    final_messages.push(assistant_message);
     final_token_counts.push(assistant_message_tokens);
 
-    // Add back the preserved user message if it exists
-    if let Some(user_message) = preserved_user_message {
-        final_messages.push(user_message);
-    }
+    let mut new_messages = vec![summary_msg];
+    new_messages.extend(continuation_messages);
+    let (new_messages, _issues) = merge_consecutive_messages(new_messages);
```
katzdave, line 140: Nice use of merge messages. What should we do with issues? Ideally there are none, but would be nice to know.

katzdave:
FYI, noticed one more issue:

[Joke request.json](https://github.com/user-attachments/files/23215449/Joke.request.json)

Since I changed manual compaction to actually send a message, we're now picking up that message (as the most recent user message) and moving it after the compaction.

Maybe another argument for just passing in a manual_compaction flag and skipping the whole last user search entirely; I think otherwise we always just want to find the most recent user message and might not need to special case the last one.

DOsinga:
should we close this one now? or do we need to copy the test?

katzdave:
> should we close this one now? or do we need to copy the test?

Yeah closing. Moved the test to https://github.com/block/goose/pull/5618

Still porting a little more functionality there

Issue #4468, opened 2025-09-01
Markdown syntax highlighting not working in Terminal.app on macOS


blackgirlbytes:
**Describe the bug**

goose CLI does not display markdown syntax highlighting when running in Terminal.app on macOS, despite the terminal supporting full color and formatting capabilities. All markdown content is rendered as plain text without colors, bold formatting, or syntax highlighting for code blocks.

## **Expected Behavior**
When goose responds with markdown content, it should display:
- ✅ **Colored headers** (e.g., `# Header` in bright color)
- ✅ **Bold text** (e.g., `**bold**` rendered as bold)
- ✅ **Syntax-highlighted code blocks** (e.g., Rust code with proper colors)
- ✅ **Formatted lists, quotes, and links**

## **Actual Behavior**
All markdown content is rendered as plain text:
- ❌ Headers appear as regular text with `#` symbols
- ❌ Bold markers `**text**` are shown literally
- ❌ Code blocks have no syntax highlighting
- ❌ No colors or formatting applied

However, I did find that the syntax highlighting shows in other apps such as Cursor's terminal or ghostty

### ghostty
<img width="560" height="587" alt="Image" src="https://github.com/user-attachments/assets/b408a053-9369-46c1-b976-9d1e076955b6" />
### Cursor's terminal
<img width="1039" height="433" alt="Image" src="https://github.com/user-attachments/assets/89f4faf4-00b1-4a50-ab04-9e5619de2a03" />
### MacOS terminal app
<img width="1281" height="755" alt="Image" src="https://github.com/user-attachments/assets/a81c92d4-dec9-4e26-8ee7-82bbe46f6f4d" />

## **Environment**
- **OS**: macOS (Terminal.app)
- **goose version**: 1.6.0
- **Terminal**: Terminal.app (default macOS terminal)
- **Shell**: bash/zsh
- **TERM**: `xterm-256color`

I asked goose, and it suggested that this could be the root cause
## **Root Cause**
The issue is in `crates/goose-cli/src/session/output.rs` in the `print_markdown()` function:

```rust
fn print_markdown(content: &str, theme: Theme) {
    if std::io::stdout().is_terminal() {  // ← Too strict!
        // bat syntax highlighting
    } else {
        print!("{}", content);  // ← Falls back to plain text
    }
}
```

The `std::io::stdout().is_terminal()` check is **too restrictive** and fails in many legitimate terminal environments, including:
- IDE integrated terminals
- Terminal multiplexers (tmux, screen)
[...]

PR #5393, opened 2025-10-27
fix: enable markdown syntax highlighting in macOS Terminal.app

ConstantTime:
## Summary
`print_markdown` used to bail out whenever `Stdout::is_terminal()` returned false. On macOS Terminal.app that check is flaky, so the CLI fell back to literal markdown (no colored headers/bold/syntax highlighting). This change centralizes the decision in `should_use_colored_output()`, layering multiple signals so we still feed the response through `bat` whenever the terminal can actually understand ANSI.

Key changes:
- treat `NO_COLOR` as the top-level kill-switch (`env_no_color()` now returns true when the var is set)
- add `should_use_colored_output()` that:
  * returns false immediately if `NO_COLOR`/`TERM=dumb` is present
  * returns true if `stdout().is_terminal()` succeeds
  * otherwise inspects `TERM`, `TERM_PROGRAM`, and `COLORTERM` so macOS Terminal.app (and other color-capable terms such as `xterm-256color`) still get color even when `is_terminal()` lies
- reuse that helper in both `print_markdown` and `render_text_no_newlines` so all CLI paths respect the same policy
- switch `bat::PrettyPrinter::colored_output(!env_no_color())` to line up with the inverted, corrected `env_no_color()` semantics

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing

- `cargo fmt`
- `cargo build`
- Manual:
  * Terminal.app interactive session (main vs branch) with markdown-heavy prompts
  * Pipe to `cat` and `cat -v` on both builds to validate the ANSI output behaviour
  * `NO_COLOR=1` run for plain-text regression coverage

### Related Issues
Relates to #4468  

Before:  

Screenshot 2025-10-27 at 9 52 06 PM

After:   

Screenshot 2025-10-27 at 9 44 22 PM

---

**Email**: rak1729e@gmail.com

### DIFF
--- a/crates/goose-cli/src/session/output.rs
+++ b/crates/goose-cli/src/session/output.rs
@@ -202,7 +202,7 @@ pub fn render_text(text: &str, color: Option<Color>, dim: bool) {
 }
 
 pub fn render_text_no_newlines(text: &str, color: Option<Color>, dim: bool) {
-    if !std::io::stdout().is_terminal() {
+    if !should_use_colored_output() {
         println!("{}", text);
         return;
     }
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/session/output.rs
```diff
@@ -535,16 +535,52 @@ fn print_tool_header(call: &CallToolRequestParam) {
 
 // Respect NO_COLOR, as https://crates.io/crates/console already does
 pub fn env_no_color() -> bool {
-    // if NO_COLOR is defined at all disable colors
-    std::env::var_os("NO_COLOR").is_none()
+    // if NO_COLOR is defined at all, disable colors
+    std::env::var_os("NO_COLOR").is_some()
```
ConstantTime, line 539: colors enabled now in default case.

taniacryptid:
Thank you for the contribution, @ConstantTime ! As we wait for the dev team to review, can you also fix the DCO check? [You can follow instructions here to do so](https://github.com/block/goose/pull/5393/checks?check_run_id=53897652482)

taniacryptid:
@DOsinga @alexhancock @zanesq @jamadeo markdown PR for review!

DOsinga:
thanks for your contribution. however I am still not sure what we're trying to fix here.

your example screenshot shows that if we go through a pipe, we will now be using colors. this is not what we want, we only want to use colors when we are displaying directly to the screen. when the output is redirected you want non-ansi code because you are probably going to programmatically process it.

taniacryptid:
> thanks for your contribution. however I am still not sure what we're trying to fix here.
> 
> your example screenshot shows that if we go through a pipe, we will now be using colors. this is not what we want, we only want to use colors when we are displaying directly to the screen. when the output is redirected you want non-ansi code because you are probably going to programmatically process it.

The main fix we want is as described in the original bug, 

> goose CLI does not display markdown syntax highlighting when running in Terminal.app on macOS, despite the terminal supporting full color and formatting capabilities

taniacryptid:
Since I didn't get any followup items addressed on this issue, I'll be taking the hacktoberfest label off of this contribution in the interim.

I will also close this issue in the interim, and hopefully you @ConstantTime or another contributor are able to get your contribution in for something like this, knowing the direction the devs want as mentioned by @DOsinga .

ConstantTime:
> Since I didn't get any followup items addressed on this issue, I'll be taking the hacktoberfest label off of this contribution in the interim.
> 
> I will also close this issue in the interim, and hopefully you @ConstantTime or another contributor are able to get your contribution in for something like this, knowing the direction the devs want as mentioned by @DOsinga .

oh sorry @taniandjerry. It's been a super busy week for me at office and I missed this thread somehow. Closing this PR makes sense as per your reasoning. Going through comments from @.DOsinga to understand the gap better.

PR #5398, opened 2025-10-27
Use session IDs as task IDs for subagents instead of UUIDs

tlongwell-block:
Replace UUID task IDs with session IDs to reduce LLM transcription errors.

## The Problem
LLMs can mistype UUIDs when referencing tasks:
- `a1b2c3d4-e5f6-7890-abcd-ef1234567890` → easy to corrupt
- 36 characters of random hex → higher error probability

## The Solution  
Session IDs are more LLM-friendly:
- `20251027_1234` → less likely to corrup
- 13 characters with a pattern → lower error probability

## Result
Fewer failed task executions due to ID typos, improving Goose's reliability when orchestrating subagents.

Additionally, this task ID format is human-readable and meaningful


tlongwell-block:
Resolves https://github.com/block/goose/issues/4955

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -94,22 +95,30 @@ pub async fn run_complete_subagent_task(
...
         let parent_session_id = task_config.parent_session_id;
         let working_dir = task_config.parent_working_dir;
-        let session = SessionManager::create_session(
-            working_dir.clone(),
-            format!("Subagent task for: {}", parent_session_id),
-        )
-        .await
-        .map_err(|e| anyhow!("Failed to create a session for sub agent: {}", e))?;
+
+        let session_id = match session_id {
```
DOsinga: so session-ids are now non optional, so I don't think here and elsewhere you need this. in general it does feel weird to create a session here though

DOsinga (commented):
we should get this in

crates/goose/src/agents/subagent_handler.rs
tlongwell-block: Make id mandatory and removed fallthrough creation

crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs
```diff
@@ -51,30 +52,39 @@ fn extract_task_parameters(params: &Value) -> Vec<Value> {
...
-            });
-            Task {
-                id: uuid::Uuid::new_v4().to_string(),
-                task_type: TaskType::SubRecipe,
-                payload,
+) -> Result<Vec<Task>> {
+    let mut tasks = Vec::new();
+
+    for task_command_param in command_params {
+        // Create a session for this task - use its ID as the task ID
```
DOsinga: delete the comment

crates/goose/src/agents/recipe_tools/sub_recipe_tools.rs
```diff
@@ -51,30 +52,39 @@ fn extract_task_parameters(params: &Value) -> Vec<Value> {
...
+        let session = SessionManager::create_session(
+            std::env::current_dir().unwrap_or_else(|_| std::path::PathBuf::from(".")),
```
DOsinga: this should not use the current working dir, but the working dir for the subagent or whatever is associated with it. previously we were using: task_config.parent_working_dir;

DOsinga (approved):

Issue #5358, opened 2025-10-24
Provider/Model Config Should Persist in Session


wpfleger96:
**Please explain the motivation behind the feature request.**
Similar to #4295, we should persist provider/model so that resumed sessions retain the same config as the "original" session

**Describe the solution you'd like**
When I run `goose session --resume --session-id <ID>` I'd like the resumed session to retain the same provider/model config that session `<ID>` originally did, whether the "original" session was started via CLI/UI/recipe/etc

- [x] I have verified this does not duplicate an existing feature request

PR #5399, opened 2025-10-27
Enable runtime access to provider name

wpfleger96:
This PR adds runtime access to provider names by introducing a simple name field and instance method to the Provider trait. This enables session management code to identify which provider is currently in use, laying the groundwork for provider/model configuration persistence across session resumption.

Session management code interacts with providers as trait objects (`&dyn Provider`). To persist provider/model configuration when saving a session, we need to identify which provider is currently in use. Previously, provider metadata was only accessible via static methods, which can't be called on trait objects. This blocked the ability to save and restore the correct provider during session resumption.

The implementation adds a simple `name` field to each provider that gets initialized from their static metadata. For standard providers (like OpenAI, Anthropic), the name comes directly from their metadata. For custom providers (like "custom_deepseek"), the name is their custom identifier—not their underlying engine's name. This ensures custom providers are correctly distinguished from their base implementations.

Implementation details:
- Added `get_name(&self) -> &str` instance method to Provider trait
- Updated all 21 provider implementations to store and return a `name: String` field
- Standard providers initialize with `name: Self::metadata().name`
- Custom providers receive their name via `from_custom_config(model, config, provider_name)`
- Declarative provider registration passes the custom provider's name to ensure proper identification
- Updated all 7 mock providers in tests to implement the new trait method

Relates to #5358

### DIFF
--- a/crates/goose/src/config/declarative_providers.rs
+++ b/crates/goose/src/config/declarative_providers.rs
@@ -293,45 +293,24 @@ pub fn register_declarative_provider(
 
     match config.engine {
         ProviderEngine::OpenAI => {
-            let provider_name = config.name.clone();
             registry.register_with_name::<OpenAiProvider, _>(
                 &config,
                 provider_type,
-                move |model| {
-                    OpenAiProvider::from_custom_config(
-                        model,
-                        config_clone.clone(),
-                        provider_name.clone(),
-                    )
-                },
+                move |model| OpenAiProvider::from_custom_config(model, config_clone.clone()),
             );
         }
         ProviderEngine::Ollama => {
-            let provider_name = config.name.clone();
             registry.register_with_name::<OllamaProvider, _>(
                 &config,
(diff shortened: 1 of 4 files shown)


crates/goose/src/config/declarative_providers.rs
```diff
@@ -293,24 +293,45 @@ pub fn register_declarative_provider(
...
+            let provider_name = config.name.clone();
             registry.register_with_name::<OpenAiProvider, _>(
                 &config,
                 provider_type,
-                move |model| OpenAiProvider::from_custom_config(model, config_clone.clone()),
+                move |model| {
+                    OpenAiProvider::from_custom_config(
+                        model,
+                        config_clone.clone(),
+                        provider_name.clone(),
```
DOsinga: any reason to pass this in separately? name sits in the config, and the method is called from_custom_config, so I'd just grab it there

PR #5401, opened 2025-10-27
fix: restore --resume functionality for most recent session

wpfleger96:
This PR fixes the `--resume` flag to correctly resume the most recent session when used without additional parameters, restoring functionality that I accidentally regressed in PR #5360

After PR #5360, using `goose run --resume` or `goose session --resume` without specifying a session name or ID would create a new session instead of resuming the last used session

Implementation details:
- Modified `get_or_create_session_id()` function in `crates/goose-cli/src/cli.rs` to check the `resume` flag when no identifier is provided
- When `resume` is true and no identifier given, fetches sessions via `SessionManager::list_sessions()` and resumes the first (most recent by `updated_at`)
- When `resume` is false and no identifier given, creates a new session as before
- Preserves all existing behavior for `--name` and `--session-id` parameters from PR #5360

### DIFF
--- a/crates/goose-cli/src/cli.rs
+++ b/crates/goose-cli/src/cli.rs
@@ -78,15 +78,13 @@ async fn get_or_create_session_id(
 
     let Some(id) = identifier else {
         if resume {
-            // Resume most recent session
             let sessions = SessionManager::list_sessions().await?;
             let session_id = sessions
                 .first()
                 .map(|s| s.id.clone())
-                .ok_or_else(|| anyhow::anyhow!("No sessions found to resume"))?;
+                .ok_or_else(|| anyhow::anyhow!("No session found to resume"))?;
             return Ok(Some(session_id));
         } else {
-            // Create new session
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/cli.rs
```diff
@@ -77,10 +77,21 @@ async fn get_or_create_session_id(
...
-            SessionManager::create_session(std::env::current_dir()?, "CLI Session".to_string())
-                .await?;
-        return Ok(Some(session.id));
+        if resume {
+            // Resume most recent session
+            let sessions = SessionManager::list_sessions().await?;
+            let session_id = sessions
+                .first()
+                .map(|s| s.id.clone())
+                .ok_or_else(|| anyhow::anyhow!("No sessions found to resume"))?;
```
DOsinga: should this be session, singular?

DOsinga (approved):
as always, remove the comments if you want to make me happy

Issue #2493, opened 2025-05-09
history.txt should be stored in ~/.local/share  (XDG_DATA_HOME or XDG_STATE_HOME)


bendavis78:
`history.txt` is currently stored in `~/.config`, however it is considered to be log or stateful data as opposed to configuration data. As such, it should be stored in `XDG_DATA_HOME` (`~/.local/share`). This is important because one might have a portable "dotfiles" repo where various files under `~/.config` (`XDG_CONFIG_HOME`) are kept under version control. 

Essentially,
```rust
        let history_file = choose_app_strategy(crate::APP_STRATEGY.clone())
            .expect("goose requires a home dir")
            .in_config_dir("history.txt");
```

should be changed to:
```rust
        let history_file = choose_app_strategy(crate::APP_STRATEGY.clone())
            .expect("goose requires a home dir")
            .in_data_dir("history.txt");
```

One slight challenge is this would need a migration step to move or copy the file on upgrade.

See [here](https://wiki.archlinux.org/title/XDG_Base_Directory) for more info.

DOsinga:
I think I agree. we could just try to read it from the old location but store it under the new one, no? do you want to make it so @bendavis78 ?

r0x0d:
I wonder, is the `history.txt` still necessary? It feels that the same information is already present in the session.db, but in a much better format.

DOsinga:
yeah, that is true - if the mood strikes you feel free to send a pr to just source it from there.

PR #5410, opened 2025-10-28
move history txt to state dir

zanesq:
## Summary
move history txt to state dir

closes https://github.com/block/goose/issues/2493

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -406,25 +406,20 @@ impl CliSession {
             }
         }
 
-        let history_loaded = if history_file.exists() {
-            if let Err(err) = editor.load_history(&history_file) {
+        let history_files = [&history_file, &old_history_file];
+        let load_from = history_files.iter().find(|f| f.exists());
+
+        if let Some(file) = load_from {
+            if let Err(err) = editor.load_history(file) {
                 eprintln!("Warning: Failed to load command history: {}", err);
             }
-            true
-        } else if old_history_file.exists() {
-            if let Err(err) = editor.load_history(&old_history_file) {
-                eprintln!("Warning: Failed to load command history: {}", err);
-            }
-            true
-        } else {
-            false
-        };
+        }
 
         let save_history =
             |editor: &mut rustyline::Editor<GooseCompleter, rustyline::history::DefaultHistory>| {
                 if let Err(err) = editor.save_history(&history_file) {
                     eprintln!("Warning: Failed to save command history: {}", err);
-                } else if history_loaded && old_history_file.exists() {
+                } else if old_history_file.exists() {
                     if let Err(err) = std::fs::remove_file(&old_history_file) {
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/session/mod.rs
```diff
@@ -397,7 +397,33 @@ impl CliSession {
         let completer = GooseCompleter::new(self.completion_cache.clone());
         editor.set_helper(Some(completer));
 
-        let history_file = Paths::config_dir().join("history.txt");
+        let history_file = Paths::state_dir().join("history.txt");
+
+        // Migrate from old location if needed
+        let old_history_file = Paths::config_dir().join("history.txt");
+        if old_history_file.exists() && !history_file.exists() {
+            if let Some(parent) = history_file.parent() {
```
DOsinga: this is duped code

crates/goose-cli/src/session/mod.rs
```diff
@@ -397,7 +397,33 @@ impl CliSession {
...
+                    // Successfully copied, now try to remove old file
+                    if let Err(remove_err) = std::fs::remove_file(&old_history_file) {
+                        eprintln!(
+                            "Warning: Migrated history file but couldn't remove old file: {}",
+                            remove_err
+                        );
+                    }
+                }
+            }
+        }
```
DOsinga, line 402: the above seems more complicated than it should be. we already check if the history file exists. if we also check whether the old history exists we can then just load from old if new doesn't exist and delete old when new does exist

DOsinga (approved):

crates/goose-cli/src/session/mod.rs
```diff
@@ -398,49 +398,36 @@ impl CliSession {
...
+            true
+        } else {
+            false
+        };
 
         let save_history =
             |editor: &mut rustyline::Editor<GooseCompleter, rustyline::history::DefaultHistory>| {
                 if let Err(err) = editor.save_history(&history_file) {
                     eprintln!("Warning: Failed to save command history: {}", err);
+                } else if history_loaded && old_history_file.exists() {
```
DOsinga: no need to check on history_loaded here I think

crates/goose-cli/src/session/mod.rs
```diff
@@ -398,49 +398,36 @@ impl CliSession {
...
-        }
+            true
+        } else if old_history_file.exists() {
+            if let Err(err) = editor.load_history(&old_history_file) {
+                eprintln!("Warning: Failed to load command history: {}", err);
+            }
+            true
+        } else {
+            false
+        };
```
DOsinga: I'd probably go with something like:

let load_from = [&history_file, &old_history_file]
    .iter()
    .find(|f| f.exists());

if let Some(file) = load_from {
    if let Err(err) = editor.load_history(file) {
        eprintln!("Warning: Failed to load command history: {}", err);
    }
}

DOsinga (approved):

PR #5416, opened 2025-10-28
Summarization iteratively removes tool call responses

katzdave:
This will help solidify compaction, so even in the event of picking up context from a subagent, or shifting the model it will have a much stronger chance of succeeding.

Will wait on https://github.com/block/goose/pull/5392, and build a test with the scaffolding that provides.


DOsinga:
closing since in draft > 2 weeks

PR #5419, opened 2025-10-28
Persist provider name and model config in the session

### LINKED ISSUE #5358: Provider/Model Config Should Persist in Session

wpfleger96:
## Summary
Building off #5399, this PR enables provider name/model config persistence by storing the configs in the `Session` struct. Now if a user resumes a previous session with `--resume`, the previous session's provider and model config will be automatically reloaded instead of defaulting to the provider/model from global config.

### Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Manually tested

### Related Issues
Fixes #5358

### DIFF
--- a/crates/goose/examples/agent.rs
+++ b/crates/goose/examples/agent.rs
@@ -10,12 +10,10 @@ use goose::session::SessionManager;
 use std::path::PathBuf;
 
 #[tokio::main]
-async fn main() {
+async fn main() -> anyhow::Result<()> {
     let _ = dotenv();
 
-    let provider = create_with_named_model("databricks", DATABRICKS_DEFAULT_MODEL)
-        .await
-        .expect("Couldn't create provider");
+    let provider = create_with_named_model("databricks", DATABRICKS_DEFAULT_MODEL).await?;
 
     let agent = Agent::new();
 
@@ -35,7 +33,7 @@ async fn main() {
         DEFAULT_EXTENSION_TIMEOUT,
     )
     .with_args(vec!["mcp", "developer"]);
-    agent.add_extension(config).await.unwrap();
+    agent.add_extension(config).await?;
 
     println!("Extensions:");
     for extension in agent.list_extensions().await {
@@ -52,13 +50,12 @@ async fn main() {
     let user_message = Message::user()
         .with_text("can you summarize the readme.md in this dir using just a haiku?");
 
-    let mut stream = agent
-        .reply(user_message, session_config, None)
-        .await
-        .unwrap();
+    let mut stream = agent.reply(user_message, session_config, None).await?;
 
     while let Some(Ok(AgentEvent::Message(message))) = stream.next().await {
-        println!("{}", serde_json::to_string_pretty(&message).unwrap());
+        println!("{}", serde_json::to_string_pretty(&message)?);
         println!("\n");
     }
+
+    Ok(())
 }


crates/goose/src/agents/agent.rs
```diff
@@ -1198,10 +1199,19 @@ impl Agent {
     pub async fn update_provider(&self, provider: Arc<dyn Provider>) -> Result<()> {
         let mut current_provider = self.provider.lock().await;
         *current_provider = Some(provider.clone());
+        drop(current_provider);
```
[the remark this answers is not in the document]
wpfleger96: resolved in https://github.com/block/goose/pull/5419/commits/a15bd93365ef4cda2874b1257aff7adce0eb5ac7

crates/goose/src/scheduler.rs
```diff
@@ -1186,15 +1180,27 @@ async fn run_scheduled_job_internal(
...
+            error: format!("Failed to set provider on agent: {}", e),
+        });
+    }
+
+    if let Err(e) = agent.persist_provider_config(&session.id).await {
+        return Err(JobExecutionError {
+            job_id: job.id.clone(),
+            error: format!("Failed to save provider config to session: {}", e),
+        });
+    }
```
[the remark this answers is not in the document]
wpfleger96: this is intentional, we should "fail open" and let the session continue if provider persistence fails

crates/goose/src/agents/agent.rs
```diff
@@ -1199,9 +1200,17 @@ impl Agent {
         let mut current_provider = self.provider.lock().await;
         *current_provider = Some(provider.clone());
 
-        self.update_router_tool_selector(Some(provider), None)
-            .await?;
-        Ok(())
+        self.update_router_tool_selector(Some(provider), None).await
+    }
+
+    pub async fn persist_provider_config(&self, session_id: &str) -> Result<()> {
```
DOsinga: it feels to me that this shouldn't be a something we need to call; if you want to change the provider/model on the agent, you should call that and then it will always save it to the session that it has. or is the problem that we don't have a session id?

wpfleger96: @DOsinga yep the problem is we need a session ID to persist provider/model config to

there are other cases where `update_provider()` is called without `persist_provider_config()` when we're not in a session e.g. when a user is running `goose configure`

DOsinga: hmm, that shouldn't be the case though.

if you run `goose configure` you don't have an agent running, so there is no session to change, so that's alright. we have a high level confusion on what it means to change the provider, does it change the global setting or does it also change the running agent. but it feels to me that unless we actually do the call on the agent, we shouldn't persist this. does that make sense?

wpfleger96: I dug into the code a bit more and I agree there's some high level confusion on what it means to change the provider but I'm thinking the root cause is really the the unclear Agent-Session relationship

it seems like there's an implicit one-to-one relationship with Agent and Session, i.e. every Agent is tied to a specific session. the only exceptions to this relationship are the same places where the code calls `update_provider()` without `persist_provider_config()`:
1. `goose configure` requires an Agent (without a session) for the tool permission configuration option because this flow lazy loads the provider from the Agent
    - we need this because to support MCP sampling/elicitation requires Provider to be accessible (via Agent)
    - except this feels like an anti-pattern because `goose configure` only starts the MCP server processes so it can call `list_tools()` to populate the available tool permission settings
    - sampling/elicitation would never be called here since we're in config, not an actual session
    - to change this we'd need to implement a way to query extension metadata without starting the full MCP server
2. Web interface: the experimental `goose web` command uses a single shared Agent that serves all web sessions, no associated Session object
3. Unit tests 
4. Debug mode: when the user chooses to have goose help debug extension failures, the debug session creates an Agent but not a persisted Session

I think this unclear relationship leads to the current problem: Agent doesn't have a clear identity so it doesn't know if it's in the "production agent" context (1:1 with sessions, should persist state) or "utility/testing" agent context (no session, ephemeral contexts used by tests or config) and because of that a `session_id` isn't easily available to an Agent. that means either we need `update_provider()` to conditionally persist based on presence of `session_id` (this is the approach I tried that we previously discussed) or you'd have to split out persistence into a separate method which is the current approach I chose with `persist_provider_config()`.

moving forward I'm thinking we could do either of:
1. Make Agent session-aware (i.e. add `session_id` as optional field to `Agent` struct). then `update_provider()` knows that when `session_id` is present (in a "production" session) it should also persist each time it runs, and when session_id isn't present (testing/debug sessions) it will skip persistence
2. Try to refactor single `Agent` type into separate `SessionAgent` and `UtilityAgent` (`TestAgent`? something like that) to make the distinction more explicit 

@DOsinga what do you think?

crates/goose-cli/src/commands/acp.rs
```diff
@@ -230,18 +230,22 @@ impl GooseAcpAgent {
         };
         let provider = create(&provider_name, model_config).await?;
 
-        // Create a shared agent instance
+        let session = SessionManager::create_session(
+            std::env::current_dir().unwrap_or_else(|_| std::path::PathBuf::from(".")),
+            "ACP Session".to_string(),
+            SessionType::Hidden,
+        )
+        .await?;
```
DOsinga, line 238: You'll need to sync to main - we now have one session per acp session, so we need to line up with that and make sure we call set_agent for that where we create the session (not the agent)

crates/goose-cli/src/commands/web.rs
```diff
@@ -158,12 +158,17 @@ pub async fn handle_web(
...
+        std::env::current_dir().unwrap_or_else(|_| std::path::PathBuf::from(".")),
+        "Web Agent Initialization".to_string(),
+        SessionType::Hidden,
+    )
+    .await?;
+
     let agent = Agent::new();
     let provider = goose::providers::create(&provider_name, model_config).await?;
-    agent.update_provider(provider).await?;
+    agent.update_provider(provider, &init_session.id).await?;
```
DOsinga, line 170: I think we should move this to where the session is created, not create a specific session for it and also update the agent there

then if you resume a session by going to /<session_id> we can in theory load the provider, model back from the session, which is ultimately the goal of this whole exercise

crates/goose/examples/agent.rs
```diff
@@ -18,7 +18,16 @@ async fn main() {
...
     let agent = Agent::new();
-    let _ = agent.update_provider(provider).await;
+
+    let session = SessionManager::create_session(
+        PathBuf::default(),
+        "max-turn-test".to_string(),
+        SessionType::Hidden,
+    )
+    .await
+    .expect("session manager creation failed");
```
DOsinga: well that's not true. let's just ? this?

DOsinga (approved):

crates/goose-cli/src/commands/acp.rs
wpfleger96: discussed and will address in a separate PR migrating ACP and Web to `AgentManager`

crates/goose-cli/src/commands/web.rs
wpfleger96: discussed and will address in a separate PR migrating ACP and Web to `AgentManager`

Issue #3526, opened 2025-07-19
Installation fails due to chrome-sandbox file conflict with existing application


ig-perez:
**Describe the bug**

When installing the RPM, the transaction aborts because Goose’s package provides its own chrome-sandbox binary, which conflicts with a file already present on the system from another application.

```shell
Transaction failed: Rpm transaction failed.
Warning: skipped OpenPGP checks for 1 package from repository: @commandline
  - file /usr/lib/.build-id/e3/48af15b8a6fcd2c4941da68ed6a01e8c5be38d from install of Goose-1.1.3-1.x86_64.rpm conflicts with file from existing package
```

**To Reproduce**
1. Download Goose-1.1.3-1.x86_64.rpm from the releases page.
2. Run: `sudo rpm -i Goose-1.1.3-1.x86_64.rpm`
3. Observe the file-conflict error on chrome-sandbox.

**Expected behavior**
The RPM should install Goose without overwriting or conflicting with binaries from other applications. Ideally, `chrome-sandbox` should be placed in a Goosed-owned directory (e.g. /opt/goose/) or be namespaced to avoid collisions.

**Screenshots**

**Please provide following information:**
 - **OS & Arch:** Fedora Linux 38 x86_64
 - **Interface:** UI
 - **Version:** v1.1.3
 - **Extensions enabled:** None
 - **Provider & Model:** None

**Additional context**
Installing via rpm2cpio and extracting into a local folder works around the issue, but proper packaging is needed to avoid system-wide conflicts.

PR #5421, opened 2025-10-28
rpm install in /opt/Goose to avoid conflicts with chrome-sandbox

zanesq:
rpm install in /opt/Goose to avoid conflicts with chrome-sandbox from other apps

fixes https://github.com/block/goose/issues/3526

don't have linux to test this but it should work


michaelneale (approved):
hard to test but can keep an eye on it.

PR #5422, opened 2025-10-28
add clippy warning for string_slice

### LINKED ISSUE #3476: clippy shuld check clippy::string_slice
Does this feature solve a particular problem you have been experiencing? What opportunities or use cases would be unlocked with this feature?

* Splitting a string by an array is not safe.
   * #3264
   * #3263

* add `clippy::string_slice` for lint
   * https://rust-lang.github.io/rust-clippy/master/index.html#string_slice

**Additional context**

zanesq:
## Summary
add clippy warning for string_slice
closes https://github.com/block/goose/issues/3476


katzdave (approved):

zanesq:
was hoping to just add warnings for now but its killing the ci linting, looking into it

zanesq:
Goose says

The `-D warnings` flag in the CI script is **promoting all warnings to errors**, including your new `string_slice` warning.

When you set `string_slice = "warn"` in `Cargo.toml`, it enables the lint as a warning. But then the CI script runs with `-D warnings`, which says "deny (treat as error) ALL warnings" - including your new `string_slice` warnings.

So you have a few options:

**Option 1:** Fix the actual string slicing issues (the safest approach)
- This is what we started doing earlier with the `playback.rs` fix

**Option 2:** Add the lint at the "deny" level instead of "warn"
- Change `string_slice = "warn"` to `string_slice = "deny"` in `Cargo.toml`
- This makes it explicit that it's an error, not just a warning being promoted

**Option 3:** Suppress existing violations with `#[allow(clippy::string_slice)]`
- This is what we did with "Option B" earlier
- Allows existing code to pass while catching new violations

**Option 4:** Modify the CI script to allow this specific warning
- Add `#[allow(clippy::string_slice)]` at the crate level temporarily

Given that your PR intent is to **enable the lint going forward**, I'd recommend **Option 1** (fix the issues) or **Option 3** (suppress existing ones).

amed-xyz (approved):

DOsinga:
yeah, I guess we need to fix all those occurrences

DOsinga:
I'll take this over

jamadeo:
@DOsinga you could use the baseline thing that @lifeizhou-ap set up if we want to merge this for future code without fixing all of the existing

DOsinga:
yeah we could, but I'll just have goose do it

PR #5427, opened 2025-10-28
Improve system theme detection and fallback

zanesq:
Be more explicit when checking for system theme setting and add current theme detection fallback

Bonus: added multi window sync

https://github.com/user-attachments/assets/d891bb0c-4cf9-49a4-bb60-d5053f8155df

fixes https://github.com/block/goose/issues/4245

### DIFF
--- a/ui/desktop/src/components/GooseSidebar/ThemeSelector.tsx
+++ b/ui/desktop/src/components/GooseSidebar/ThemeSelector.tsx
@@ -36,6 +36,14 @@ const setThemeModeStorage = (mode: 'light' | 'dark' | 'system') => {
     localStorage.setItem('use_system_theme', 'false');
     localStorage.setItem('theme', mode);
   }
+
+  const themeData = {
+    mode,
+    useSystemTheme: mode === 'system',
+    theme: mode === 'system' ? '' : mode,
+  };
+
+  window.electron?.broadcastThemeChange(themeData);
 };
 
 const ThemeSelector: React.FC<ThemeSelectorProps> = ({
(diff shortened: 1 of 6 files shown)


ui/desktop/src/components/GooseSidebar/ThemeSelector.tsx
```diff
@@ -14,12 +14,19 @@ const ThemeSelector: React.FC<ThemeSelectorProps> = ({
   horizontal = false,
 }) => {
   const [themeMode, setThemeMode] = useState<'light' | 'dark' | 'system'>(() => {
-    const savedUseSystemTheme = localStorage.getItem('use_system_theme') === 'true';
-    if (savedUseSystemTheme) {
+    const savedUseSystemTheme = localStorage.getItem('use_system_theme');
```
DOsinga: this is better, but it retains the awkwardness of the original; the setter and the getter needs to be kept in sync with what we store in local storage. I think it would be better to just forego the useState pattern and use localStorage directly, implement getThemeMode and setThemeMode directly

ui/desktop/src/components/GooseSidebar/ThemeSelector.tsx
```diff
@@ -14,12 +14,19 @@ const ThemeSelector: React.FC<ThemeSelectorProps> = ({
...
     }
-    const savedTheme = localStorage.getItem('theme');
-    return savedTheme === 'dark' ? 'dark' : 'light';
+    if (savedTheme) {
+      return savedTheme === 'dark' ? 'dark' : 'light';
+    }
+
+    // No explicit preference: infer from current DOM state (set by index.html)
+    // This preserves the system preference that was applied on initial load
+    const isDarkInDOM = document.documentElement.classList.contains('dark');
```
DOsinga: like this feels awkward, if you had getThemeMode you could just use this as the fallback if it wasn't stored

amed-xyz (approved):

God-damnit-all:
Seems to have broken the Windows CLI build (again).

God-damnit-all:
I seem to have accidentally triggered a workflow with my last comment...

Issue #4524, opened 2025-09-04
Allow the chat window to be forced to the top


simonsickle:
**Please explain the motivation behind the feature request.**
Sometimes I want to click into another view like my IDE or chrome window and I want goose to say at the top, just not focused anymore. Other apps, like the android emulator, seem to have functionality to force the app to the front and keep it there even when you click away.

**Describe the solution you'd like**
I would like to enable a setting in the top bar of the system (ie, under window) for a specific goose window to force this session of goose to always be on top. A hotkey shortcut would also be nice!

**Describe alternatives you've considered**
I currently work around this by splitting goose and my IDE but on smaller screens this does not work well. I'd rather just drag it out of the way if I need to see under it.

- [x] I have verified this does not duplicate an existing feature request

PR #5429, opened 2025-10-28
Add menu option for setting window always on top

zanesq:
## Summary
Add menu option for setting window always on top with hotkey command
Verified works locally

closes https://github.com/block/goose/issues/4524

image

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -484,7 +484,6 @@ let appConfig = {
 const windowMap = new Map<number, BrowserWindow>();
 const goosedClients = new Map<number, Client>();
 const windowPowerSaveBlockers = new Map<number, number>();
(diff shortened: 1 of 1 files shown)


ui/desktop/src/main.ts
```diff
@@ -483,9 +483,8 @@ let appConfig = {
 
 const windowMap = new Map<number, BrowserWindow>();
 const goosedClients = new Map<number, Client>();
-
-// Track power save blockers per window
-const windowPowerSaveBlockers = new Map<number, number>(); // windowId -> blockerId
+const windowPowerSaveBlockers = new Map<number, number>();
+const windowAlwaysOnTop = new Map<number, boolean>();
```
jamadeo: why do we need to keep track in a map?

PR #5444, opened 2025-10-29
fix: accept string or number for analyze tool depth parameters

tlongwell-block:
The `analyze` tool in the developer MCP extension expects numeric parameters (`max_depth` and `follow_depth`) as integers, but AI agents sometimes pass these as strings (e.g., `"2"` instead of `2`), causing deserialization failures. This simple fix adds a flexible deserializer that accepts both formats, making the tool more forgiving and agent-friendly without changing its behavior—similar to how web forms accept both numeric and string inputs for number fields.


tlongwell-block:
Closed in favor of https://github.com/block/goose/pull/5478

PR #5445, opened 2025-10-29
fix: don't use user specific paths in mcp integration tests

alexhancock:
We shouldn't use something like `~/goose/crates/goose/tests/tmp/goose.txt` in an MCP integration test, as the final path will be different on each developer who may want to record's laptop, causing churn. I also ran into an issue because I don't store goose in my home dir.

This makes it use an absolute path to a temp dir on the machine.


crates/goose/tests/mcp_integration_test.rs
```diff
@@ -108,22 +108,22 @@ enum TestMode {
...
         // Test shell command to verify file was modified
         CallToolRequestParam { name: "shell".into(), arguments: Some(object!({
-            "command": "cat ~/goose/crates/goose/tests/tmp/goose.txt"
+            "command": "cat /tmp/goose_test/goose.txt"
         })) },
         // Test text_editor tool to restore original content
         CallToolRequestParam { name: "text_editor".into(), arguments: Some(object!({
             "command": "str_replace",
-            "path": "~/goose/crates/goose/tests/tmp/goose.txt",
+            "path": "/tmp/goose_test/goose.txt",
```
jamadeo, line 126: we could just delete these tests. The purpose of the snapshot test was to verify the protocol messaging, not so much the servers themselves. The developer mcp server should probably have its own integration test instead

alexhancock: I see the logic, yeah. I think I'll merge this as an improvement on current state, then we should consider quality integration tests for the developer and other built in servers.

PR #5450, opened 2025-10-29
Goose recover

DOsinga:
## Summary
Makes it easy to pipe errors into goose. Implements it for extension install failures

### DIFF
--- a/ui/desktop/src/sessions.ts
+++ b/ui/desktop/src/sessions.ts
@@ -1,5 +1,5 @@
 import { Session, startAgent } from './api';
-import { setViewType } from './hooks/useNavigation';
+import type { setViewType } from './hooks/useNavigation';
 
 export function resumeSession(session: Session, setView: setViewType) {
   if (process.env.ALPHA) {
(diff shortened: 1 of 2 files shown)


ui/desktop/src/sessions.ts
```diff
@@ -1,19 +1,17 @@
-import { Session } from './api';
+import { Session, startAgent } from './api';
+import { setViewType } from './hooks/useNavigation';
```
zanesq: should be a type-only import `import type { setViewType } from './hooks/useNavigation';` to follow our conventions elsewhere and without the type-only modifier can cause TypeScript errors

zanesq (approved):
LGTM tested locally, minor nit with type import.

ui/desktop/src/sessions.ts
DOsinga: huh. can you say more about that? what type of typescript errors?

DOsinga: ah, I think I understand. I think in this case it can't really cause errors since we're going to need useNavigation further in the call stack, but still

Issue #4709, opened 2025-09-23
🍳 Advanced Recipe Submission #7


taniandjerry:
# 🍳 Advanced Recipe Submission 

Thanks for contributing to the Goose Cookbook!  
This issue represents an **Advanced Recipe Slot (8–10 points)**.  

Advanced recipes use **3 or more extensions** and/or **multiple sub-recipes** to orchestrate complex workflows.  

---

## ✅ Acceptance Criteria  
Your recipe will be accepted if it:  
- Uses **3+ extensions** in a single workflow.  
- Optionally (or preferably) includes **multiple sub-recipes**.  
- Demonstrates orchestration, parameter passing, conditional logic, or parallel execution.  
- Includes required YAML fields: `title`, `description`, `instructions`, and at least one `activity`.  
- Validates against the Recipe Reference Guide.  
- Has been tested locally in Goose.  
- Follows the Cookbook Contributing Guide.  

---

## 📋 Submission Checklist  
Before opening your PR, make sure you've:  
- [ ] Chosen a unique and descriptive filename (e.g., `code-review-pipeline.yaml`).  
- [ ] Matched the recipe `title` to the filename.  
- [ ] Confirmed your YAML validates.  
- [ ] Listed all extensions and sub-recipes used.  
- [ ] Included clear instructions and multiple activities.  
- [ ] Verified the recipe works locally in Goose.  

---

## ✨ Inspiration Examples  
*(These are just ideas — you can submit **any real workflow** you've used with Goose.)*  
- Code review pipeline with sub-recipes (security + quality checks).  
- Smart analyzer that conditionally chooses sub-recipes.  
- Research assistant chaining multiple sources into a report.  
- Project bootstrapper (generate starter templates).  
- Data visualization helper (convert CSV/JSON to charts).  

---

## 📚 Resources  
- [Recipe Reference Guide](https://block.github.io/goose/docs/guides/recipes/recipe-reference)  
- [Cookbook Contributing Guide](https://github.com/block/goose/blob/main/CONTRIBUTING_RECIPES.md#-contributing-recipes-to-goose-cookbook)  
- [Goose Recipes Tutorial (YouTube)](https://youtu.be/8rTliYrQ6Iw)  
- [Existing Recipes](https://github.com/block/goose/tree/main/documentation/src/pages/recipes/data/recipes)  

---

## 🚀 Next Steps  
1. Claim this slot by commenting `.take` below.  
2. Fork the repository and create a new branch for your recipe.
[...]

PR #5451, opened 2025-10-29
technical debt tracker recipe

Better-Boy:
Fixes https://github.com/block/goose/issues/4709


taniacryptid:
Tagging teammates to help review since it otherwise looks good! @blackgirlbytes @EbonyLouis

PR #5456, opened 2025-10-29
chore(tests/mcp): testing for MCP sampling

alexhancock:
Two approaches to tests that will ensure MCP sampling continues to work (following the merge of https://github.com/block/goose/pull/5367)

- 1. `mcp_integration_tests.rs` - Make sure all codepaths related to sampling work correctly assuming the llm responds predictably. No assertions on content here, or integration with a real provider, just making sure the messages are handled properly.

- 2. A new `scripts/test_mcp.sh` script, similar to `scripts/test_providers.sh` but which does two things:
    - Calls a tool that causes sampling to occur across a range of models
    - Uses a separate fast model as a judge to determine if the interaction went through and actually performed sampling.


alexhancock:
@jamadeo @DOsinga testing is now set up here for sampling.

Douwe and I discussed how we should work to merge `test_providers.sh` `test_subrecipes.sh` and `test_mcp.sh` into a unified approach of tests matrixed across the the providers/models with some ability to choose what runs where, but makes more sense in a followup.

DOsinga (approved):

Issue #5459, opened 2025-10-29
[Docs] Update Apify stdio


Better-Boy:
**Describe the bug**

Currently, goose community extensions doc for apify has the following command

```
goose session --with-extension "npx -y apify-mcp-server@latest"
```
ref - https://block.github.io/goose/extensions/

<img width="1807" height="690" alt="Image" src="https://github.com/user-attachments/assets/e478c42a-be35-4bf2-b81a-2d29805843db" />

npm `apify-mcp-server@latest` package is no longer available. Instead apify uses `apify/actors-mcp-server` package.

Refer [Apify Stdio MCP npm package](https://docs.apify.com/platform/integrations/mcp#local-stdio)

Update the package to the right one.

Better-Boy:
.take

Better-Boy:
@taniandjerry when I was creating youtube short for MCP for apify, found this bug. Please tag hacktoberfest and small.

taniacryptid:
Oh since you found this bug, I'm happy to make it a quick Hacktoberfest issue. For reference for the time remaining, not opening or labeling any new issues as we await final passes on some open PRs. Thank you so much for contributing this!

Navadeep0007:
.take

PR #5460, opened 2025-10-29
apify doc fix

Better-Boy:
Fixes https://github.com/block/goose/issues/5459


taniacryptid:
Let me tag @dianed-square @blackgirlbytes specifically to triple check, but this looks good to me!

dianed-square:
Thanks for catching!

PR #5461, opened 2025-10-29
only run fork on branch PRs

michaelneale:
We don't need to run these on forks, also they won't work as require access to secrets, so they will only be run from branches.

### DIFF
--- a/.github/workflows/pr-smoke-test.yml
+++ b/.github/workflows/pr-smoke-test.yml
@@ -13,8 +13,16 @@ on:
 name: Live Provider Tests
 
 jobs:
+  check-fork:
+    runs-on: ubuntu-latest
+    # Skip entire workflow for PRs from forks (they don't have access to secrets)
+    if: github.event_name == 'workflow_dispatch' || github.event.pull_request.head.repo.full_name == github.repository
+    steps:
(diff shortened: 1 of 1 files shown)


taniacryptid:
Maybe rebase for failed CI test? [Line 459](https://github.com/block/goose/actions/runs/18923354932/job/54024888367?pr=5461#step:6:459) is it because the expected 128000 context limit for gpt-4o is supposed to be something different?

jamadeo:
If we want to, we can still do this in a way such that they'll work. `pull_request_target` can run on a fork and have secrets access, but of course we'll need to be careful with that since it will run untrusted code. It would still have to be run only on approval

DOsinga:
> Maybe rebase for failed CI test? [Line 459](https://github.com/block/goose/actions/runs/18923354932/job/54024888367?pr=5461#step:6:459) is it because the expected 128000 context limit for gpt-4o is supposed to be something different?

that test is somehow flaky

taniacryptid:
> If we want to, we can still do this in a way such that they'll work. `pull_request_target` can run on a fork and have secrets access, but of course we'll need to be careful with that since it will run untrusted code. It would still have to be run only on approval

o:

PR #5463, opened 2025-10-29
Stringly typed config

jamadeo:
- remove some .get(<string>) calls to config
- make the router thing a boolean (it was a string)

### DIFF
--- a/crates/goose/src/config/goose_mode.rs
+++ b/crates/goose/src/config/goose_mode.rs
@@ -1,3 +1,5 @@
+use std::str::FromStr;
+
 use serde::{Deserialize, Serialize};
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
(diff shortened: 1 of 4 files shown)


crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
```diff
@@ -93,7 +94,7 @@ pub struct TaskParameter {
 
 pub fn should_enabled_subagents(model_name: &str) -> bool {
     let config = crate::config::Config::global();
-    let is_autonomous = config.get_param("GOOSE_MODE").unwrap_or("auto".to_string()) == "auto";
+    let is_autonomous = config.get_goose_mode().unwrap_or(GooseMode::Auto) == GooseMode::Auto;
```
DOsinga, line 97: here and elsewhere; having a typed get_goose_mode is of course so much better, but should we add a default value to the pattern? we never read goose mode without defaulting to auto if it is not set

crates/goose/src/config/goose_mode.rs
```diff
@@ -0,0 +1,24 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum GooseMode {
+    Auto,
+    Approve,
+    SmartApprove,
```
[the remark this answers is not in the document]
DOsinga: shut up robot

crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
jamadeo: yeah, we should do that. every config needs some kind of default value (it might be None of course)

PR #5465, opened 2025-10-29
use app.isPackaged instead of checking for node env development

zanesq:
## Summary
use app.isPackaged instead of checking for node env development


DOsinga:
I'm going to take this over if you don't mind while you are absent @zanesq

jamadeo (approved):

PR #5468, opened 2025-10-30
keep the order of keys in config.yaml

jamadeo:
serde_yaml::Mapping preserves key order, but we were throwing that away and using HashMaps. No more!

### DIFF
--- a/crates/goose/src/config/base.rs
+++ b/crates/goose/src/config/base.rs
@@ -610,7 +610,7 @@ impl Config {
     /// - There is an error serializing the value
     pub fn set_param<V: Serialize>(&self, key: &str, value: V) -> Result<(), ConfigError> {
         let _guard = self.guard.lock().unwrap();
-        let mut values = self.load_values()?;
+        let mut values = self.load()?;
         values.insert(serde_yaml::to_value(key)?, serde_yaml::to_value(value)?);
         self.save_values(values)
     }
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/commands/info.rs
```diff
@@ -40,7 +40,7 @@ pub fn handle_info(verbose: bool) -> Result<()> {
     // Print verbose info if requested
     if verbose {
         println!("\n{}", style("goose Configuration:").cyan().bold());
-        match config.load_values() {
```
DOsinga, line 43: pre existing, but I think we can now just let the error bubble up

crates/goose/src/agents/extension.rs
```diff
@@ -215,6 +215,8 @@ pub enum ExtensionConfig {
     Sse {
         /// The name used to identify this extension
         name: String,
+        #[serde(default)]
+        #[schema(required)]
```
DOsinga, line 219: aha! that's the way

crates/goose/src/providers/bedrock.rs
```diff
@@ -61,7 +61,7 @@ impl BedrockProvider {
             }
         };
 
-        set_aws_env_vars(config.load_values());
+        set_aws_env_vars(config.all_values());
         set_aws_env_vars(config.load_secrets());
```
DOsinga: hmmm

jamadeo: is that a reaction to the fact that we export any secret that starts with AWS_ to the environment? because, yeah...

PR #5478, opened 2025-10-30
fix: Add schema-aware numeric coercion for MCP tool arguments

tlongwell-block:
Some LLMs quote numeric arguments when calling MCP tools (e.g., sending `"42"` instead of `42`), which can cause tool execution failures. This change adds schema-aware coercion that automatically converts string representations of numbers to their proper numeric types, but only when the tool's schema explicitly indicates a parameter should be numeric. The implementation is conservative - if no schema is available, no coercion is performed to avoid accidentally modifying strings that should remain as strings.

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -49,7 +49,7 @@ use rmcp::model::{
     CallToolRequestParam, Content, ErrorCode, ErrorData, GetPromptResult, Prompt,
     ServerNotification, Tool,
 };
-use serde_json::Value;
+use serde_json::{json, Value};
 use tokio::sync::{mpsc, Mutex};
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, info, instrument, warn};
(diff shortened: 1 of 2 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -383,15 +383,66 @@ impl Agent {
...
     }
 
+    /// Apply schema-aware numeric coercion to tool arguments
+    /// Converts string representations of numbers to actual numbers based on the tool's schema
+    async fn coerce_numeric_arguments(&self, tool_call: &mut CallToolRequestParam) {
+        let Some(ref mut args) = tool_call.arguments else {
+            return;
+        };
+
+        let tools = self.list_tools(None).await;
```
DOsinga: we're looking only for one tool, list_tools will actually go and call all mcp servers to ask for their tools - that could be rather expensive! now that I say that, I worry that we are doing this willy-nilly anyway

still though, I think we should move this code to the place where we actually get the toolrequests, in categorize_tools - which I see already has a list of tools (unused!)

crates/goose/src/agents/agent.rs
```diff
@@ -383,15 +383,66 @@ impl Agent {
...
+                Some(Value::String(t)) => t == "number" || t == "integer",
+                Some(Value::Array(types)) => types
+                    .iter()
+                    .any(|t| t.as_str().is_some_and(|s| s == "number" || s == "integer")),
+                _ => false,
+            };
+
+            if should_be_number {
+                if let Ok(n) = s.parse::<f64>() {
+                    // Preserve integer types when possible
```
DOsinga: json doesn't have a distinction between integers and floats, so this is all a bit of a no-op

we should consider to do the same for booleans maybe?

crates/goose/src/agents/agent.rs
```diff
@@ -383,15 +383,66 @@ impl Agent {
...
     #[instrument(skip(self, tool_call, request_id), fields(input, output))]
     pub async fn dispatch_tool_call(
         &self,
-        tool_call: CallToolRequestParam,
+        mut tool_call: CallToolRequestParam,
         request_id: String,
         cancellation_token: Option<CancellationToken>,
         session: Option<SessionConfig>,
     ) -> (String, Result<ToolCallResult, ErrorData>) {
+        self.coerce_numeric_arguments(&mut tool_call).await;
```
DOsinga: I don't think you want to do this in a mutable way - just make it functional

DOsinga (approved):
this is definitely an improvement, but I think it should live somewhere else

PR #5480, opened 2025-10-30
Manual compaction counting fix + cli cleanup

katzdave:
Token counts updated in the stream immediately after compaction.

Remove lots of unused code from compaction; only return the provider usage.

Modify update_session_metrics to support using summarization usage to estimate new context window (with system prompt token estimate).

Remove lots of custom code in CLI around token counting + manual summarization trigger

Rename /summarize -> /compact.


crates/goose/src/agents/reply_parts.rs
```diff
@@ -273,11 +276,26 @@ impl Agent {
         let accumulated_output =
             accumulate(session.accumulated_output_tokens, usage.usage.output_tokens);
 
+        let (current_total, current_input, current_output) = if is_compaction_usage {
+            // After compaction: summary output becomes new input context
```
jamadeo, line 278: after manual compaction, should this not really just go to zero? have we sent the manually compacted conversation to the model yet or only to the client? although I could see how that would be confusing to a user

it would be good to avoid a hard coded system prompt estimate, just because that could really be all over the place

I suppose we could do something like use the token counts from the first message you send, but that would be complicated to do

crates/goose/src/context_mgmt/mod.rs
```diff
@@ -31,13 +31,12 @@ struct SummarizeContext {
 /// # Returns
 /// * A tuple containing:
 ///   - `Conversation`: The compacted messages
-///   - `Vec<usize>`: Token counts for each message
-///   - `Option<ProviderUsage>`: Provider usage from summarization
```
jamadeo, line 35: nice, I saw we didn't use this. There's some code on the frontend you could clean up with this too

jamadeo (approved):

crates/goose/src/context_mgmt/mod.rs
katzdave: ack will take another pass on the frontend too.

crates/goose/src/agents/reply_parts.rs
katzdave: Yeah I think this one is more of a question of what we want to show the user. Let's try dropping it to the summary size, there will be a bit of whiplash for the user when the system prompt re-enters on the next message, but I don't love that constant either.

PR #5481, opened 2025-10-30
Fix: Compaction client <-> server sync 

katzdave:
Just noticed we regressed on this;

Leading to the message following compaction restoring the old messages (that were no longer visible) to the session.

Added some scary comments since it's pretty hard to see the impact of that one line.


zanesq:
is there a test to catch this?

Issue #5288, opened 2025-10-21
Blocking BUG - v1.10+ Subrecipes blocked by missing session, even if explicit pre-existing session_id passed in


cbruyndoncx:
Blocking BUG - v1.10+ Subrecipes blocked by missing session, even if explicit pre-existing session_id passed in

Unable to run in v1.10 or v1.11.1, 

in cli 
* --no-session does not work
* using --session-id 20251021_2 (which i noted from interactive session), starts main recipe, but fails on sub-recipes with same error

Running: /home/cb/.local/bin/goose run --recipe main_recipe.yaml --quiet -r --session-id 20251021_2 --params update_frequency= --params limit=100 --params days=2
Loading recipe: Run full youtubeiq pipeline (preprocess -> fetch -> process)
Description: Top-level recipe to fetch video from channels or playlists, and per-video processing using sub-recipes.
Parameters used to load this recipe:
   update_frequency:
   limit: 100
   days: 2

─── create_task_preprocess_channels | subrecipe ──────────────────────────
task_parameters:
    -
        days: 2
        limit: 100
        update_frequency: weekly

─── execute_task | subagent ──────────────────────────
execution_mode: sequential
task_ids: d672b21b-1638-4684-a92c-1d49e15e27f4

─── create_task_fetch_videos | subrecipe ──────────────────────────
task_parameters:
    -
        channelIds: ["UCdoadna9HFHsxXWhafhNvKw","UCyaN6mg5u8Cjy2ZI4ikWaug","UCpoWCikpTk6Ysw4GSI___hw","UCQo_1uyUB4wf-5u7OsVIGHg","UCENWgdIISKK2M7xGeG1ZhW...
        days: 2

─── execute_task | subagent ──────────────────────────
execution_mode: sequential
task_ids: a3c55349-fc79-4094-89cc-362760bf27ed

─── create_task_driver | subrecipe ──────────────────────────
task_parameters:
    -
        videoIds: ["kREtJ91lIw4","GePFC-P99pk","7whH1RpeD0E","hwPBiBLbjzA","DTBTl6vwkb0","0W7WKhA7szU","c9tPUc6HJfQ","u8V45xsnkGA","2KqxVl5g7lw","BYirio4...

─── execute_task | subagent ──────────────────────────
execution_mode: sequential
task_ids: 7341bff7-3ad6-4bb1-9066-6e22230a2938

─── final_output | recipe ──────────────────────────
[...]

PR #5486, opened 2025-10-30
fix: disable subagents unless env var explicitly provided

alexhancock:
Running into a number of issues with subagents, such as subagents...

1. not returning, with no visibility into what they are doing in the UX
2. flooding session history with tens or hundreds of small sessions which seem to spiral and get very far away from the original top level work
3. hanging indefinitely on the cli

Seems appropriate to disable them across the board for now until these kinds of issues can be addressed

### DIFF
--- a/crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
+++ b/crates/goose/src/agents/recipe_tools/dynamic_task_tools.rs
@@ -93,6 +93,9 @@ pub struct TaskParameter {
 }
 
 pub fn should_enabled_subagents(model_name: &str) -> bool {
+    if std::env::var("GOOSE_ENABLE_SUBAGENTS").is_err() {
+        return false;
+    }
     let config = crate::config::Config::global();
     let is_autonomous = config.get_goose_mode().unwrap_or(GooseMode::Auto) == GooseMode::Auto;
     if !is_autonomous {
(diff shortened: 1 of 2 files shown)


DOsinga:
hmm, does this impact sub-recipes? in that case we might want to rethink this, because that is a valid use case

cbruyndoncx:
I am unable to use my recipe / sub-recipe setup since 1.10.x as the whole session management is not working. It wont create the necessary sessions, and it will not run without session. If sub-recipes would run without sessions my issue would be solved . I had them all setup with no-session. In general i think if you have a lot of sub-recipe/sub-agent runs, like in my case, the occasional hickup is not important if the overall success rate is still good. I have been adding additional logging to persist the errors i got so i can follow up, not needing the session details. Possibly others do similarly, when relying on subagents and sub recipes.

alexhancock:
> I am unable to use my recipe / sub-recipe setup since 1.10.x as the whole session management is not working. It wont create the necessary sessions, and it will not run without session. If sub-recipes would run without sessions my issue would be solved . I had them all setup with no-session. In general i think if you have a lot of sub-recipe/sub-agent runs, like in my case, the occasional hickup is not important if the overall success rate is still good. I have been adding additional logging to persist the errors i got so i can follow up, not needing the session details. Possibly others do similarly, when relying on subagents and sub recipes.

Thanks for the feedback on sub-recipes @cbruyndoncx 

Is there a separate issue for session <-> sub-recipe issues that describe it in more detail?

cbruyndoncx:
issue #5288

alexhancock:
Unclear that we want to do this this way - closing out for now

PR #5494, opened 2025-10-31
fix: customised recipe to yaml string to avoid minininjia parsing error

lifeizhou-ap:
## Summary
This fix is for a recipe with backtick and {{if}} block

## Issue
serde_yaml serialization uses internal heuristics to decide between `|` and quoted text with escaped`\n` and `\"`, and the quoted form breaks MiniJinja parsing.  It seems that if the texts contains backtick, it tends to use the quoted texts without "|"

Example before:

`prompt: "Hello \\\"World\\\"\\n{% if user == \\\"admin\\\" %}Welcome{% endif %}"`

After fix:
```
prompt: |
   Hello "World"
  {% if user == "admin" %}Welcome{% endif %}
```

## Changes
Apply extra format after the yaml serialization to convert the quoted multiline texts to the texts with "|"

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Unit testing and manual testing

### DIFF
--- a/crates/goose/src/recipe/mod.rs
+++ b/crates/goose/src/recipe/mod.rs
@@ -237,8 +237,10 @@ impl Recipe {
     }
 
     pub fn to_yaml(&self) -> Result<String> {
-        let recipe_yaml = serde_yaml::to_string(self).map_err(|err| anyhow::anyhow!("Failed to serialize recipe: {}", err))?;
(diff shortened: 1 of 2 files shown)


lifeizhou-ap:
.bundle

lifeizhou-ap:
.bundle

Issue #5383, opened 2025-10-26
Scheduling existing recipes - missing deeplink and creates duplicate recipe


angiejones:
## Bug Description

When a user presses the **Schedule** button on an existing recipe in the desktop app (v1.11.0), two problems occur:

### Issue 1: Recipe Deeplink Not Auto-Populated
The scheduler dialog is not auto-populated with the recipe's deeplink, requiring the user to manually fill in the recipe path/URL.

**Expected behavior:** The recipe deeplink should be automatically populated in the scheduler when scheduling an existing recipe.

**Actual behavior:** The recipe field is empty and must be filled in manually.

### Issue 2: Duplicate Recipe Created
Once the recipe is scheduled, a duplicate version of the existing recipe is created in the Recipe view.

**Expected behavior:** Scheduling an existing recipe should not create a duplicate entry in the Recipe view.

**Actual behavior:** A duplicate recipe entry appears in the Recipe view after scheduling.

## Environment
- **Goose Version:** 1.11.0
- **Platform:** Desktop app

## Steps to Reproduce
1. Open the desktop app
2. Navigate to an existing recipe in the Recipe view
3. Click the "Schedule" button on the recipe
4. Observe that the recipe deeplink is not pre-filled in the scheduler dialog
5. Fill in the schedule details and create the scheduled job
6. Return to the Recipe view
7. Observe that a duplicate of the recipe now appears

## Impact
- Users must manually enter recipe information that should be automatically populated
- Recipe view becomes cluttered with duplicate entries
- Poor user experience when scheduling recipes

zanesq:
@amed-xyz might have fixed this on main

DOsinga:
reverting this feature would also be an option

amed-xyz:
Issue 1 was fixed by https://github.com/block/goose/pull/5290
Issue 2 is going to be fixed by https://github.com/block/goose/pull/5513 once merged

PR #5513, opened 2025-10-31
Fix duplicate recipe upon schedule creation

amed-xyz:
## Summary

Current behavior: Once a recipe is scheduled, a duplicate version of the existing recipe is created in the Recipe view. This is due temp files are created in `GOOSE_WORKING_DIR` during schedule creation but never cleaned up.

Expected behavior: Scheduling an existing recipe should not create a duplicate entry in the Recipe view.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing

Success Path:

- Create a schedule from a recipe
- Submit successfully
- Check `GOOSE_WORKING_DIR` - temp file should be deleted

Cancel Path:

- Create a schedule from a recipe
- Click Cancel before submitting
- Check `GOOSE_WORKING_DIR` - temp file should be deleted

### Related Issues
Relates to https://github.com/block/goose/issues/5383

### DIFF
--- a/ui/desktop/src/components/schedule/SchedulesView.tsx
+++ b/ui/desktop/src/components/schedule/SchedulesView.tsx
@@ -361,10 +361,6 @@ const SchedulesView: React.FC<SchedulesViewProps> = ({ onClose: _onClose }) => {
     setIsSubmitting(true);
     setSubmitApiError(null);
 
-    if (tempFilePath) {
-      setPendingTempFile(tempFilePath);
-    }
-
     try {
       await createSchedule(payload);
       await fetchSchedules();


ui/desktop/src/components/schedule/CreateScheduleModal.tsx
```diff
@@ -23,6 +23,7 @@ export interface NewSchedulePayload {
   recipe_source: string;
   cron: string;
   execution_mode?: string;
+  temp_file_path?: string;
```
angiejones: Should this field be part of the `NewSchedulePayload` interface? Feels like an implementation detail that is now leaking into the API contract. The backend's `/schedule/create` endpoint doesn't expect or use this field from what I can see.

**Suggested approach (from goose):**

1. Remove `temp_file_path` from the `NewSchedulePayload` interface
2. Pass it separately to the submit handler, or
3. Filter it out before sending to the API

**Example fix:**

```typescript
// In SchedulesView.tsx handleCreateScheduleSubmit:
const handleCreateScheduleSubmit = async (payload: NewSchedulePayload) => {
  setIsSubmitting(true);
  setSubmitApiError(null);

  // Extract temp_file_path before sending to API
  const { temp_file_path, ...apiPayload } = payload;

  if (temp_file_path) {
    setPendingTempFile(temp_file_path);
  }

  try {
    await createSchedule(apiPayload); // Send clean payload without temp_file_path
    // ... rest of the logic
```

ui/desktop/src/components/schedule/SchedulesView.tsx
```diff
@@ -343,15 +352,37 @@ const SchedulesView: React.FC<SchedulesViewProps> = ({ onClose: _onClose }) => {
   const handleCreateScheduleSubmit = async (payload: NewSchedulePayload) => {
     setIsSubmitting(true);
     setSubmitApiError(null);
+
+    if (payload.temp_file_path) {
+      setPendingTempFile(payload.temp_file_path);
+    }
```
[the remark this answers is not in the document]
amed-xyz: realistically there's no way the user can close the modal during the API call, but I added some logic for cleaner error logs

ui/desktop/src/components/schedule/CreateScheduleModal.tsx
amed-xyz: Good point! Extracted `temp_file_path` out of the payload.

DOsinga (commented):
wouldn't be easier not to create the temporary file in the goose working dir (whatever that is?) despite the final handlers etc, this still seems a bit brittle.

ideally we would just not use a file location, but just send the recipe to the end point instead of a file path

DOsinga:
do we still want to get this in? do you have thoughts on my last remark?

amed-xyz:
> wouldn't be easier not to create the temporary file in the goose working dir (whatever that is?) despite the final handlers etc, this still seems a bit brittle.

Yes, maybe? These changes still fix the underlying bug which has been around for a long while, they can be merged as-is or pivoted to your approach, however I don't have bandwidth for the latter.

zanesq:
I'll take it from here thanks @amed-xyz!

zanesq:
Tested and I'm not seeing this behavior anymore so closing for now.

PR #5516, opened 2025-10-31
Standardize CLI argument flags and update documentation

wpfleger96:
This PR establishes consistent flag naming patterns across all Goose CLI commands, eliminating confusion around `--id` vs `--session-id`, conflicting short flags, and mismatched documentation. Previously, session commands mixed `--id` and `--session-id`, the `--name` flag had a misleading id alias, and schedule commands lacked short flags entirely.

Before this change, users encountered inconsistent flag names across commands, making it unclear when to use `--id` versus `--session-id`. The Identifier struct's `--name` flag incorrectly aliased to id, and the `-p` short flag conflicted between path and working_dir usage. Schedule commands used generic `--id`, and documentation showed flags that didn't match the actual implementation.

- Updated Identifier struct to use `alias = "id"` on `--session-id` (instead of on `--name`), removed `-p` short flag from legacy `--path` parameter, and kept `-n` for name
- Changed `SessionCommand::Remove` to use `--session-id` and separate `--name` flag, updating `handle_session_remove` to accept three distinct parameters
- Renamed all schedule command flags to `--schedule-id` keeping `--id` as a backwards compatible alias, updating handlers in `schedule.rs`
- Fixed `SessionCommand::List` working_dir to use `-w` with `-p` as backward-compatible alias
- Updated help strings to clarify "recurring scheduled job" for schedule-id and fixed session `--resume` help to reference `--session-id` instead of `--path`
- Marked ServicesStatus and ServicesStop commands as `[Deprecated]`
- Added "Flag Naming Conventions" section to documentation and updated all examples to reflect standardized patterns

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)


crates/goose-cli/src/cli.rs
```diff
@@ -177,7 +177,7 @@ enum SessionCommand {
         ascending: bool,
 
         #[arg(
-            short = 'p',
+            short = 'w',
```
katzdave, line 190: I agree w makes more sense. Maybe we go w and p to keep this the same?

crates/goose-cli/src/cli.rs
```diff
@@ -58,7 +58,7 @@ pub struct Identifier {
     pub session_id: Option<String>,
 
     #[arg(
-        short,
+        short = 'p',
```
katzdave: Probably not important to add this since we hopefully won't keep this around for too much longer.

crates/goose-cli/src/cli.rs
```diff
@@ -40,16 +40,16 @@ struct Cli {
...
         value_name = "NAME",
         help = "Name for the chat session (e.g., 'project-x')",
-        long_help = "Specify a name for your chat session. When used with --resume, will resume this specific session if it exists.",
-        alias = "id"
+        long_help = "Specify a name for your chat session. When used with --resume, will resume this specific session if it exists."
     )]
     pub name: Option<String>,
 
     #[arg(
+        short = 'i',
```
katzdave: 'id' would be slightly longer but just feels a lot more clear.

crates/goose-cli/src/cli.rs
```diff
@@ -235,8 +240,8 @@ enum SessionCommand {
 enum SchedulerCommand {
     #[command(about = "Add a new scheduled job")]
     Add {
-        #[arg(long, help = "Unique ID for the job")]
-        id: String,
+        #[arg(short = 'i', long = "schedule-id", help = "Unique ID for the job")]
```
katzdave: I don't like that we overload i to mean session and schedule-id. 

I'd be biased towards just not having a short for this (but I also don't use the scheduler on CLI at all).

katzdave (commented):
Nice change! Let's get some more opinions from the team on this one since it's largely a matter of taste.

crates/goose-cli/src/cli.rs
cbruyndoncx: if the explanation is unique ID for the job, then it should be job-id , as a schedule implies a series of job imho, or is it really a session-id ?

wpfleger96: looking through the code it seems like this param refers to the unique ID of a recurring scheduled job, not a single execution/job instance. so the help string is misleading 😓

crates/goose-cli/src/cli.rs
wpfleger96: great idea, I kept `w` but readded `p` as an alias for backwards compatibility

crates/goose-cli/src/cli.rs
```diff
@@ -235,8 +240,11 @@ enum SessionCommand {
 enum SchedulerCommand {
     #[command(about = "Add a new scheduled job")]
     Add {
-        #[arg(long, help = "Unique ID for the job")]
-        id: String,
+        #[arg(
+            long = "schedule-id",
+            help = "Unique ID for the recurring scheduled job"
+        )]
+        schedule_id: String,
```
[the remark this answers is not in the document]
wpfleger96: this is intentional, see https://github.com/block/goose/pull/5516#discussion_r2496176948

crates/goose-cli/src/cli.rs
wpfleger96: great point, removed this

crates/goose-cli/src/cli.rs
wpfleger96: removed the `i` short and replaced with `id` alias

crates/goose-cli/src/cli.rs
wpfleger96: addressed in latest commit: https://github.com/block/goose/pull/5516/commits/4fbb701a9eb54abf3d91d7f330c91d9df2eb4c32

documentation/docs/guides/goose-cli-commands.md
```diff
@@ -93,11 +112,12 @@ Start or resume interactive chat sessions.
...
 ```bash
 # Start a basic session
-goose session --name my-project
+goose session -n my-project
 
 # Resume a previous session
-goose session --resume --name my-project
-goose session --resume --id 2025250620_013617
+goose session --resume -n my-project
+goose session --resume -i 2025250620_013617
```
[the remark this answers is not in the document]
wpfleger96: fixed

crates/goose-cli/src/cli.rs
```diff
@@ -177,7 +176,8 @@ enum SessionCommand {
         ascending: bool,
 
         #[arg(
-            short = 'p',
+            short = 'w',
+            short_alias = 'p',
```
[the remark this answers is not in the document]
wpfleger96: Copilot hallucinated this, at least `short_alias` is working for me locally

DOsinga (approved):

dianed-square:
Thanks @wpfleger96! Would you also be able to update the feature docs that reference the changed flags? Specifically:
- guides/recipes/session-recipes.md): --id → --schedule-id
- troubleshooting/diagnostics-and-reporting.md: --id → --session-id

It would be better to show the new canonical flag names even though the old flags still work via aliases.

wpfleger96:
@dianed-square great catch thank you! just updated those feature docs and a couple more places I found while testing

dianed-square (approved):

Issue #5506, opened 2025-10-31
Gemini 2.5 Flash: 400 Bad Request – Invalid JSON "type" field in function_declarations with Goose 1.12.1


tirath5u:
**Describe the bug**

After a fresh install of Goose 1.12.1 and using the Gemini 2.5 Flash model, I encounter an error when running requests:

Request failed: Request failed with status: 400 Bad Request. Message: Invalid JSON payload received. Unknown name "type" at 'tools.function_declarations[20].parameters.properties[0].value': Proto field is not repeating, cannot start list. Invalid JSON payload received. Unknown name "type" at 'tools.function_declarations[22].parameters.properties[1].value': Proto field is not repeating, cannot start list.

The error indicates an invalid JSON payload with unknown or unexpected "type" fields in the function declarations, causing Gemini API to reject the request.

**To Reproduce**

Fresh install Goose version 1.12.1

Configure Goose to use Google Gemini 2.5 Flash model as the provider

Run a request or session that triggers Gemini tool/function calling

Observe the 400 Bad Request error with invalid JSON payload related to "type" fields

Expected behavior

Gemini 2.5 Flash model should accept the valid function declaration schema generated by Goose without rejecting the request. The API calls should succeed without JSON schema errors.
This was working fine prior to the install - I was using 1.93 befor this. 

Screenshots

Attached

Please provide following information:

OS & Arch: Windows 10 Pro x64

Interface: CLI (or UI if applicable)

Version: 1.12.1

Extensions enabled: Attached screenshot

![Image](https://github.com/user-attachments/assets/4590a50e-eaa4-437a-815d-95762765f193)

![Image](https://github.com/user-attachments/assets/a19c29df-8eaf-4608-974e-55ddc0f413c2)

Provider & Model: Google - gemini-2.5-flash

Additional context

This appears to be an incompatibility or schema mismatch between Goose's function declaration JSON and the latest Gemini API expectations. The error resembles known issues where Proto fields expect single objects but receive arrays/lists or unsupported fields like "type". Checking Goose's docs troubleshooting and existing GitHub issues shows this is a current problem with the Gemini 2.5 Flash integration. Updating Goose or the schema format may be necessary to resolve this.

atlasoft:
+100

DOsinga:
I tested this and I think #5530 indeed fixes this (again)

PR #5530, opened 2025-11-02
Fixes Gemini API parse issue by converting nullable type arrays to single types in tool schemas

psilvis:
## Summary
Fixes issue where Gemini API rejected tool schemas containing type arrays like ["string", "null"]. Gemini's API expects type fields to be single string values.

The root cause was in the process_map function, which was cloning values for most keys instead of calling process_value. The process_value function already had logic to convert type arrays to single values, but it was only being called for specific keys (properties, items, anyOf, allOf), not for the type key itself.

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
I manually reproduced the issue using Gemini 2.5 Flash as the provider, then tested again with this change to verify the fix. This PR also includes a unit test.

### Related Issues
Relates to [#5506](https://github.com/block/goose/issues/5506)

Before:  
Screenshot 2025-11-02 at 10 05 48 AM

After:   
Screenshot 2025-11-02 at 10 06 04 AM

### DIFF
--- a/crates/goose/src/providers/formats/google.rs
+++ b/crates/goose/src/providers/formats/google.rs
@@ -238,7 +238,7 @@ pub fn process_map(map: &Map<String, Value>, parent_key: Option<&str>) -> Value 
                         value.clone()
                     }
                 }
-                _ => value.clone(),
+                _ => process_value(value, Some(key.as_str())),
             };
 
             Some((key.clone(), processed_value))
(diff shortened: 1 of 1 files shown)


angiejones:
thank you so much for this fix! can you sign the [DCO](https://github.com/block/goose/pull/5530/checks?check_run_id=54304673741):

```
git commit --amend --no-edit --signoff
git push -f origin fix-gemini-api-parsing-issue
```

angiejones (approved):

tirath5u:
Thanks. When will this fix be available in the application?  @psilvis

Issue #5528, opened 2025-11-02
Frequent "Server error: None" after multiple prompts in both CLI and Desktop


blackgirlbytes:
Based on [Discord thread ](https://discord.com/channels/1287729918100246654/1434269750195912905)
## Description
Both the CLI and Desktop applications frequently encounter a "Server error: None" after a couple of prompts, requiring users to retry or restart.

## Error Message
```
Ran into this error: Server error: None.
Please retry if you think this is a transient or recoverable error.
```

## Environment
- **Affected Components**: CLI and Desktop applications
- **Models Tested**:
  - Qwen-3-coder (local)
  - Tetrate Agent Router Service models:
    - OpenAI
    - Claude
    - Google

## Steps to Reproduce
1. Start either the CLI or Desktop application
2. Send a few prompts (typically occurs after "a couple of prompts")
3. Error occurs intermittently

## Expected Behavior
The application should continue to process prompts without encountering server errors.

## Actual Behavior
- Error occurs frequently after multiple prompts
- Sometimes the application recovers without intervention
- Sometimes requires restart to continue working

## Additional Context
- The error appears to be intermittent and not tied to a specific model provider
- Occurs across different model types (local and cloud-based)
- The error message lacks specific details ("None"), making debugging difficult

JEFH507:
I have been running into the same issue since last night. I was using Tetrate GPT5 and now Openrouter GPT5, I tried other models (same router providers) got same error always. Sometimes it start the process others just crash. 

At the beginning it will recover, but now all conversations fall into the same very quickly.
I am using pop os, desktop.

michaelneale:
I am trying openrouter w GPT-5, unable to see it there, will look at tetrate.

michaelneale:
@JEFH507 I can't see it with any prividers other than tetrate, is it possible that you are still actually using tetrate in the settings?

michaelneale:
I am seeing 500 responses from tetrate agent router service, so may be an outage, openrouter seems fine, anthropic and openai 1st party, databricks is fine, I also tested with qwen3-coder/ollama and was ok.

JEFH507:
I was in openrouter gpt5 yes.
Update now:
I tested open router and tetrate in gpt 5, none worked.

Then tested tetrate with antropic claude sonet, did not work.

Now tested openrouter with cloude sonnet, **that one work.**

michaelneale:
@JEFH507 you sure you are using right model name for openrouter: 

```
~/Downloads/goose session
[...]

PR #5532, opened 2025-11-02
fix: improve server error messages to include HTTP status code

angiejones:
## Summary

Previously, when a server error (5xx) occurred without a response body, the error message would display 'Server error: None', which was not helpful for debugging.

This change improves error messages by:
- Including the HTTP status code (e.g., 'HTTP 500')
- Providing a clear message when no response body is received
- Maintaining detailed error info when a response body is present

Example error messages:
- Before: 'Server error: None'
- After: 'HTTP 500: No response body received from server'

This affects all OpenAI-compatible providers (OpenAI, Claude, Google, Ollama, Azure, Databricks, GitHub Copilot, etc.) and Google-compatible endpoints.

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
ran all the tests in the utils module, and updated them to match new response

### Related Issues
related to #5528. this gives a better message but doesn't solve the problem of frequently running into 500 errors.

### DIFF
--- a/crates/goose/src/providers/utils.rs
+++ b/crates/goose/src/providers/utils.rs
@@ -70,7 +70,7 @@ fn check_context_length_exceeded(text: &str) -> bool {
 
 fn format_server_error_message(status_code: u16, payload: Option<&Value>) -> String {
     if let Some(p) = payload {
-        format!("HTTP {}: {:?}", status_code, p)
+        format!("HTTP {}: {}", status_code, p.to_string())
     } else {
         format!(
             "HTTP {}: No response body received from server",
@@ -90,7 +90,7 @@ pub fn map_http_error_to_provider_error(
                 "Authentication failed. Please ensure your API keys are valid and have the required permissions. \
         Status: {}{}",
                 status,
-                payload.as_ref().map(|p| format!(". Response: {:?}", p)).unwrap_or_default()
+                payload.as_ref().map(|p| format!(". Response: {}", p.to_string())).unwrap_or_default()
             );
             ProviderError::Authentication(message)
         }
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/utils.rs
```diff
@@ -1112,13 +1123,13 @@ mod tests {
             (
                 StatusCode::INTERNAL_SERVER_ERROR,
                 None,
-                ProviderError::ServerError("None".to_string()),
+                ProviderError::ServerError("HTTP 500: No response body received from server".to_string()),
```
DOsinga: use the format_server_error_message here

crates/goose/src/providers/utils.rs
```diff
@@ -1112,13 +1123,13 @@ mod tests {
...
             ),
             // is_server_error() with payload
             (
                 StatusCode::BAD_GATEWAY,
                 Some(json!({"error": "upstream error"})),
-                ProviderError::ServerError("Some(Object {\"error\": String(\"upstream error\")})".to_string()),
+                ProviderError::ServerError("HTTP 502: {\"error\":\"upstream error\"}".to_string()),
```
DOsinga: and also here

crates/goose/src/providers/utils.rs
```diff
@@ -68,6 +68,17 @@ fn check_context_length_exceeded(text: &str) -> bool {
         .any(|phrase| text_lower.contains(phrase))
 }
 
+fn format_server_error_message(status_code: u16, payload: Option<&Value>) -> String {
```
DOsinga: shouldn't status_code by of ype StatusCode here?

DOsinga (approved):

crates/goose/src/providers/utils.rs
angiejones: eh, then I need to add `.as_u16()` inside the function. but guess that's negligible. will fix...

PR #5545, opened 2025-11-03
feat(ui): add custom accent color theme support

jom-sq:
## Summary

Add custom accent color theming system that allows users to select a custom color that themes the entire UI, working seamlessly with both light and dark modes.

### Type of Change

- [x] Feature
- [x] Refactor / Code quality

**Features:**
- Custom color picker with preset colors and ability to add custom colors
- Modal dialog for adding and saving custom colors to personal palette (up to 12 colors)
- Native CSS `color-mix()` for dynamic color manipulation
- Automatic theme generation: backgrounds get visible tint, text colors stay readable
- Support for both light and dark modes with proper contrast
- Persistent theme preferences via localStorage
- Initial theme loading to prevent flash of unstyled content

**Components:**
- `CustomColorPicker`: UI for selecting and adding custom colors with modal dialog
- `ThemeSelector`: Enhanced theme selection with custom color toggle switch
- `colorUtils`: Lightweight utilities using native CSS features

**Implementation Highlights:**
- Uses native CSS `color-mix()` instead of custom color conversion functions
- Graduated tinting: backgrounds (8%), mid-tones (4-6%), text (2-3%)
- Accent color automatically lightened in dark mode for visibility
- Simple, maintainable codebase with minimal custom logic

### Testing:
- Component tests for CustomColorPicker and ThemeSelector
- Proper mocking and accessibility checks
- Manual testing across light/dark modes with various colors

Before

https://github.com/user-attachments/assets/ccf6917e-901d-4bd3-8a15-a22e111cdcc0

After

https://github.com/user-attachments/assets/49244a64-61e0-4d97-9b8f-8f361f1a33af

---

**Note:** This implementation was simplified based on feedback to use native CSS features instead of reimplementing color math in JavaScript


DOsinga:
thanks for taking this initiative. I do have some initial feedback here - this seems like a lot of code for what should be a relatively small feature - setting the color accent on the index.html. do we need our own color picker here and implement conversions between different color spaces? I would probably also drop the tests. they test the implementation not so much the results. I think maintaining them will be more expensive than the benefit of them catching future bugs.

spencrmartin:
Hey @jom-sq 

Pulled it down but want to get an alignment around the functionality I should be seeing. Within light mode the accent selector changes colours for primary elements but nothing really else (toggles, main chat ctas, radio buttons)

Screenshot 2025-11-04 at 12 39 19 PM

While within dark mode it seems to only change the mainapp base colour - CTAs etc are all the retained darkmode element. 

Screenshot 2025-11-04 at 12 40 36 PM

jom-sq:
> Hey @jom-sq
> 
> Pulled it down but want to get an alignment around the functionality I should be seeing. Within light mode the accent selector changes colours for primary elements but nothing really else (toggles, main chat ctas, radio buttons)
> 
> Screenshot 2025-11-04 at 12 39 19 PM
> While within dark mode it seems to only change the mainapp base colour - CTAs etc are all the retained darkmode element.
> 
> Screenshot 2025-11-04 at 12 40 36 PM

Hi @spencrmartin ! Yeah exactly, I didn't want to branch out too deep and change colouring everywhere but I could expand to make it affect more components, just worried about usability etc and would have to test each thoroughly. Not many elements are using the --color-accent CSS variable

jom-sq:
> thanks for taking this initiative. I do have some initial feedback here - this seems like a lot of code for what should be a relatively small feature - setting the color accent on the index.html. do we need our own color picker here and implement conversions between different color spaces? I would probably also drop the tests. they test the implementation not so much the results. I think maintaining them will be more expensive than the benefit of them catching future bugs.

@DOsinga yeah you're right, this is pretty over-engineered. The intent was to generate a theme palette from a colour, but it's overkill for just the accent feature and can be iterated on later...

I'll simplify it to remove some of the conversions, use the native colour picker and fix it up a bit (remove tests etc)

spencrmartin:
Regarding the overall UX - would suggest we bury hex selections and custom swatch selections within a flow. 

Example here:
https://github.com/user-attachments/assets/e616bc6a-339b-4972-8e6a-d6a6ec09556d

Getting the user to "add" a custom colour so they can retain some of their custom selections

spencrmartin:
I couldnt find the branch but I have some work here you may want to leverage 

https://github.com/block/goose/pull/5566

jom-sq:
> I couldnt find the branch but I have some work here you may want to leverage
> 
> #5566

@spencrmartin thank you for that! I added an option to delete custom colours on top of this since it would just stay there :) we can add an "edit" too but depends how useful that is.. I think being able to delete and recreate a new one is still decent UX (compared to having them stuck there)

If you can re-review 🙏

PR #5548, opened 2025-11-03
Sessions required

DOsinga:
## Summary

Makes session ids required, hides sessions of non user type and sends only the user message to the end point

also gets rid of tests that are pointless or don't actually succeed


crates/goose/src/agents/agent.rs
```diff
@@ -731,183 +727,132 @@ impl Agent {
         }
     }
 
-    #[instrument(skip(self, unfixed_conversation, session), fields(user_message))]
+    #[instrument(skip(self, user_message, session_config), fields(user_message))]
     pub async fn reply(
         &self,
-        unfixed_conversation: Conversation,
-        session: Option<SessionConfig>,
+        user_message: Message,
```
katzdave, line 733: Fantastic!

crates/goose/src/agents/types.rs
```diff
@@ -87,11 +86,9 @@ pub struct SessionConfig {
     /// Unique identifier for the session
     pub id: String,
     /// Working directory for the session
-    pub working_dir: PathBuf,
+    //pub working_dir: PathBuf,
```
katzdave: is this supposed to be deleted?

crates/goose-server/src/routes/reply.rs
```diff
@@ -257,17 +257,30 @@ pub async fn reply(
...
         let session_config = SessionConfig {
             id: session_id.clone(),
-            working_dir: session.working_dir.clone(),
             schedule_id: session.schedule_id.clone(),
-            execution_mode: None,
             max_turns: None,
             retry_config: None,
         };
 
+        let user_message = match messages.last() {
```
katzdave, line 265: Seems like we're still passing in all messages from the client + building/modifying all_messages below. Is that intentional?

crates/goose/src/agents/types.rs
DOsinga: yes. let me delete it for real. the session config has a working directory and a session id which maps to a session which also has a working dir. they were the same in practice, but we should only have one source of truth

crates/goose-server/src/routes/reply.rs
DOsinga: we're not really

* we're still sending the entire conversation over, but only pick up the last message
* we keep track of all_messages here separately but only for telemetry. this is somewhat misguided, I think it is measuring the wrong thing and it should happen in agent.reply, not here, but this PR is big enough as is

katzdave (approved):

PR #5551, opened 2025-11-03
fix: remove qwen3-coder from provider/mcp smoke tests

alexhancock:
Noticing the `qwen3-coder` often produces failing test results in test_providers.sh an test_mcp.sh because it doesn't reliably call the tools we expect

Most recent example where it simply didn't call the `sampleLLM` tool when it is directly told to do so by the test script https://github.com/block/goose/actions/runs/19044249135/job/54389311803?pr=5550


amed-xyz (approved):

michaelneale:
@alexhancock are you sure it isn't a regression, as it was I thought reliably working. The aim of these is to catch drift in Goose so fi it adds something that breaks popular models then we have a regression. 

(I think this is https://openrouter.ai/qwen/qwen3-coder model?) - if it doesn't work for this, we have regressed something in goose.

PR #5552, opened 2025-11-03
cli: adjust default theme to address #1905

### LINKED ISSUE #1905: CLI: Solarized themes display with very low contrast
**Describe the bug**
Here's what it looks like with Solarized Light:
Image
The light theme is the only one that looks OK with Solarized Light.

Here's what it looks like with Solarized Dark:

Image
The dark and ANSI themes look OK with Solarized Dark.

(The first one is with Goose's dark theme)

**Expected behavior**
I'd expect to be able to read the text, regardless of whether using light or dark versions of terminal themes.

Ideally, goose would either switch between dark and light depending on system themes, or just stick to ANSI (and/or allow it to be configurable, see #1521)

 - **OS & Arch:** MacOS on iTerm2
 - **Interface:** CLI
 - **Version:** v1.0.6

**Additional context**

alexhancock:
Using `Theme::Ansi` by default leads to better rendering in a range of common terminal themes

Fixes https://github.com/block/goose/issues/1905

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -495,6 +495,10 @@ impl CliSession {
 
                     let current = output::get_theme();
                     let new_theme = match current {
+                        output::Theme::Ansi => {
+                            println!("Switching to Light theme");
+                            output::Theme::Light
+                        }
                         output::Theme::Light => {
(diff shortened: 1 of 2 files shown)


michaelneale (approved):
oh nice, I didn't know about this!

PR #5554, opened 2025-11-03
Better search paths and handling of CLI providers

### LINKED ISSUE #4133: Can't use Claude Code on Windows
**Describe the bug**

1. Select Claude Code as model provider
2. Exec chat "Develop a tamagotchi game that lives on my computer and follows a pixelated styling"
3. Error: `Request failed: Failed to spawn Claude CLI command 'C:\Users\Sammy\AppData\Roaming\npm\claude': %1 is not a valid Win32 application. (os error 193). Make sure the Claude Code CLI is installed and in your PATH, or set CLAUDE_CODE_COMMAND in your config to the correct path.`

**Expected behavior**
Command should not fail

**Screenshots**

Image

claude is a recognized command on my system:

Image

 - **OS & Arch:** Windows 11 Pro 24H2
 - **Interface:** [UI/CLI]
 - **Version:** 1.4.0
 - **Extensions enabled:** Computer Controller, Developer, Playwright, Vscode, Memory
 - **Provider & Model:** Claude Code

**Additional context**

jamadeo:
The big changes here are better binary resolution using the search paths from your config plus some common npm paths, and better handling of provider setups when we can't actually create the provider.

Before, we'd let you configure a provider in goose desktop even if we couldn't create it in goosed. Now, you'll see something like this:

Screenshot 2025-11-04 at 3 23 21 PM

in the provider config section, and if you do add it to your config anyway, you'll see:

Screenshot 2025-11-04 at 3 23 53 PM

if you try to use it.

fixes #4133


rafalohaki:
btw when i use goose, its breaking path of my nodejs installed with scoop.sh 
its likely reinstalling nodejs after i run goose

crates/goose-server/src/routes/config_management.rs
```diff
@@ -734,6 +746,41 @@ pub async fn update_custom_provider(
...
+        .map_err(|err| (StatusCode::BAD_REQUEST, err.to_string()))?;
+    Ok(())
+}
+
+#[utoipa::path(
+    post,
+    path = "/config/set_provider",
+    request_body = SetProviderRequest,
+)]
+pub async fn set_provider(
```
DOsinga: should we call this one set_config_provider and the other set_agent_provider for easier reading?

crates/goose/src/config/search_path.rs
```diff
@@ -1,25 +1,121 @@
...
+            .unwrap_or_default();
+
+        #[cfg(unix)]
+        {
+            paths.push("/usr/local/bin".into());
+            paths.push("~/.local/bin".into());
+        }
+
+        if cfg!(target_os = "macos") {
+            paths.push("/opt/homebrew/bin".into());
```
DOsinga, line 29: macports uses: /opt/local/bin

crates/goose/src/config/search_path.rs
```diff
@@ -1,25 +1,121 @@
...
+impl SearchPaths {
+    pub fn builder() -> Self {
+        let mut paths = Config::global()
+            .get_goose_search_paths()
+            .unwrap_or_default();
+
+        #[cfg(unix)]
+        {
+            paths.push("/usr/local/bin".into());
+            paths.push("~/.local/bin".into());
```
DOsinga: also `/usr/bin` or is that understood

crates/goose/src/config/search_path.rs
```diff
@@ -1,25 +1,121 @@
...
+        }
+
+        if cfg!(target_os = "macos") {
+            paths.push("/opt/homebrew/bin".into());
+        }
+
```
DOsinga, line 32: my llm reluctantly suggests:

#[cfg(windows)]
{
    paths.push("%APPDATA%\\npm".into());
    paths.push("%USERPROFILE%\\.local\\bin".into());

DOsinga (approved):
nice work!

crates/goose-server/src/routes/config_management.rs
jamadeo: yeah, good idea

crates/goose/src/config/search_path.rs
jamadeo: on mac/desktop PATH is `/usr/bin:/bin:/usr/sbin:/sbin` (started via launchctl) so I think it's safe to assume that will always be there. I'm not entirely sure of the situation on linux but it's probably also in the system path?

crates/goose/src/config/search_path.rs
jamadeo: appdata/npm is covered in the with_npm bit. For home/.local/bin, yeah that can just be pulled out of the if cfg(unix) arm since shellexpand will do the right thing on windows

PR #5453, opened 2025-10-29
chore: adds .yarn to gitignore

anthonydmays:
Signed-off-by: Anthony D. Mays <anthony@morganlatimer.com>

## Summary
Adds `.yarn` folder to .gitignore in `documentation/` project to prevent adding 1K+ files to git after installing deps.

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [x] Build / Release
- [ ] Other (specify below)

### Testing
Manual testing

PR #5555, opened 2025-11-03
chore: remove yarn usage from documentation

alexhancock:
Remove `yarn` usage from documentation. We don't use yarn elsewhere in the project, and we had both `yarn.lock` and `package-lock.json` in this dir. We can just use `npm` equivalents for these 

cc @block/goose-devrel


blackgirlbytes (approved):
okay that's fine

alexhancock:
will it be annoying for your workflow at all @blackgirlbytes & co?

was just doing this to simplify after reviewing https://github.com/block/goose/pull/5453

PR #5557, opened 2025-11-03
info: print location of sessions.db via goose info

alexhancock:
Now that we use `sessions.db` for storage, updating the output of `goose info` to show it


crates/goose-cli/src/commands/info.rs
```diff
@@ -19,7 +20,7 @@ pub fn handle_info(verbose: bool) -> Result<()> {
     // Define the labels and their corresponding path values once.
     let paths = [
         ("Config dir:", config_dir),
-        ("Sessions dir:", sessions_dir.display().to_string()),
+        ("Sessions DB:", sessions_db.display().to_string()),
```
DOsinga: could add (sqlite)

DOsinga (approved):

Issue #4365, opened 2025-08-27
goose Generates Huge Log Files in Hidden .local Folder on Mac


corydouthat-sq:
**Describe the bug**
I have found that goose frequently hangs indefinitely with simple commands. This doesn't always cause a problem, but sometimes if I forget to stop it, it will generate log files in the 100s of GBs until it fills up the drive. See screenshot. 

**To Reproduce**
1. Use goose UI on Mac for normal research, code analysis, or questions. 
2. goose hangs on a complex request.
3. Forget to stop it, it never times out.
4. Eventually the Macbook starts running very slowly once it is nearly out of drive space

**Expected behavior**
- Log files would be capped at a certain size
- Log files and caches would be in a non-hidden location and could be cleared from within the app
- goose would have a reasonable default timeout

**Screenshots**
<img width="908" height="442" alt="Image" src="https://github.com/user-attachments/assets/f112c2ac-174c-47bc-8b45-6320955ee392" />

**Please provide following information:**
 - macOS Sequoia 15.6.1, Apple Silicon M1
 - UI
 - Version: 1.5.0-block.202508202012...
 - Extensions: Developer

**Additional context**
In this most recent instance, the logs were called ...-goosed.log and ...-mcp-developer.log, but I had to delete them.

towry:
I can confirm this issue. https://github.com/towry/dots/blob/next/scripts/freeup-my-tiny-macbook.sh#L29

naveendhaterwal:
.take

zanesq:
related https://github.com/block/goose/pull/4985

DOsinga:
thanks, @zanesq - yeah I took a stab at this. If you want to fork the PR and continue to work on that (take @jamadeo 's ) suggestions in for example, that could work!

taniacryptid:
Let us know if you have any questions, @naveenkumar29052006 !

taniacryptid:
Hi @naveenkumar29052006 - wanted to check if you are still working on this!

naveendhaterwal:
> Hi [@naveenkumar29052006](https://github.com/naveenkumar29052006) - wanted to check if you are still working on this!

yes @taniandjerry i am almost fininshed with my work

Shreyanshsingh23:
.take

taniacryptid:
As it seems @Shreyanshsingh23 that @naveenkumar29052006 is still working on this, you may explore other open issues!

taniacryptid:
[...]

PR #5561, opened 2025-11-04
feat: log rotation

### LINKED ISSUE #4965: goosed runaway logging with no rotation under ~/.local/state/goose/logs/server on macOS — logs up to 630 GB per file
Hey team,
Ran into this issue with Goose today and used ChatGPT to resolve. Goose suggested this was a result of a bug and suggested I file this bug report + drafted it up for me. Let me know if you need anything else!

Summary
- On macOS, the Goose server (goosed) is writing enormous daily log files under ~/.local/state/goose/logs/server with no apparent rotation or sane size limits. This filled my disk completely.
- This appears to be runaway logging and/or an excessively verbose default log level combined with missing/ineffective rotation.

Environment
- OS: macOS 15.6.1
- Goose Desktop: 1.9.0-block.202509242155-cd4ca (1.9.0-block.202509242155-cd4ca)
- How goosed starts: Unknown (likely background via Desktop; I did not consciously start it)
- XDG_STATE_HOME: Unknown

Impact
- Severity: Critical
- Effect: Disk fills to 100%, risking data loss and system instability. Required manual truncation of logs to recover space.

Observed behavior
- Logs written to: ~/.local/state/goose/logs/server/
- Example file sizes:
  - 20250930_111045-goosed.log → 630 GB
  - 20250930_111618-goosed.log → 26 GB
  - 20251001_goosed.log → 121 GB
- Behavior persisted until I manually truncated the files and killed goosed.

Expected behavior
- Default logging should:
  - Use a sane default level (info or warn).
  - Rotate by size and/or time automatically.
  - Enforce a hard cap to prevent any single file from exceeding a safe size (e.g., 50–200 MB).
- Disk should never be at risk from default logging behavior.

Steps to reproduce (best effort)
1) Install/run Goose Desktop 1.9.0-block.202509242155-cd4ca on macOS 15.6.1.
2) Allow the app/server to run in the background.
3) Over time (hours to a day), logs accumulate under ~/.local/state/goose/logs/server and can grow to hundreds of GB without rotation.

Temporary mitigation taken
- Truncated the large .log files (truncate -s 0 …) to recover space.
- Did not disable the server because I’m unsure if I need it; would prefer a product-side fix.

Notes
- Path is XDG-style; I didn’t intentionally set XDG_STATE_HOME (unknown if set by the app environment). On macOS, it may be more conventional to log under ~/Library/Logs/Goose or use macOS’s unified logging, but the main issue is the lack of rotation and extreme verbosity.
- I’m happy to provide additional diagnostics if helpful.

Requested fix
- Set a safe default log level (warn or info).
- Add size- and time-based rotation by default.
- Enforce a hard maximum per file and a total retention cap.
- Consider platform-appropriate default log locations and rotation mechanisms on macOS.
- Provide a user-facing toggle for server logging level and retention in Desktop settings, or ensure server inherits a safe default.

Optional diagnostics you can request from me
- Check if XDG_STATE_HOME is set:
  - echo "$XDG_STATE_HOME"
- List log sizes:
  - ls -lahS ~/.local/state/goose/logs/server | head -n 50
- Find open file handles (if space doesn’t return after deletion):
  - sudo lsof -nP | grep goosed

Thank you! This issue is blocking due to the potential for complete disk exhaustion.

alexhancock:
Simple log rotation to prevent runaway log file storage. Will fix https://github.com/block/goose/issues/4965 and https://github.com/block/goose/issues/4365

Approach:

* Rotates log content to new files daily
* Deletes log files after two weeks
* Sets up a new `llm` subdirectory for `llm-request` logs so they can use the same cleanup strategy

Notes:

* There wasn't a super natural place to call the cleanup function for llm requests, so I coupled calls to it when we do cleanup for `cli` and `server` dirs
* This will require a one time manual `rm` of logs at the top level of the log dir, as I moved llm-requests to a new subdir, but after that should work for everything automatically

### DIFF
--- a/crates/goose-server/src/logging.rs
+++ b/crates/goose-server/src/logging.rs
@@ -1,5 +1,4 @@
 use anyhow::{Context, Result};
-use std::path::PathBuf;
 use tracing_appender::rolling::Rotation;
 use tracing_subscriber::{
     filter::LevelFilter, fmt, layer::SubscriberExt, util::SubscriberInitExt, EnvFilter, Layer,
@@ -8,20 +7,15 @@ use tracing_subscriber::{
 
 use goose::tracing::{langfuse_layer, otlp_layer};
 
-/// Returns the directory where log files should be stored.
-/// Creates the directory structure if it doesn't exist.
-fn get_log_directory() -> Result<PathBuf> {
-    goose::logging::get_log_directory("server", true)
-}
-
 /// Sets up the logging infrastructure for the application.
 /// This includes:
 /// - File-based logging with JSON formatting (DEBUG level)
 /// - Console output for development (INFO level)
 /// - Optional Langfuse integration (DEBUG level)
 pub fn setup_logging(name: Option<&str>) -> Result<()> {
-    // Set up file appender for goose module logs
-    let log_dir = get_log_directory()?;
+    let _ = goose::logging::cleanup_old_logs("server");
+    let _ = goose::logging::cleanup_old_logs("llm");
+    let log_dir = goose::logging::get_log_directory("server", true)?;
     let timestamp = chrono::Local::now().format("%Y%m%d_%H%M%S").to_string();
 
     // Create log file name by prefixing with timestamp
@@ -31,9 +25,9 @@ pub fn setup_logging(name: Option<&str>) -> Result<()> {
         format!("{}.log", timestamp)
     };
 
-    // Create non-rolling file appender for detailed logs
+    // Create daily rolling file appender for detailed logs
     let file_appender =
-        tracing_appender::rolling::RollingFileAppender::new(Rotation::NEVER, log_dir, log_filename);
+        tracing_appender::rolling::RollingFileAppender::new(Rotation::DAILY, log_dir, log_filename);
 
     // Create JSON file logging layer
     let file_layer = fmt::layer()
--- a/crates/goose-server/src/logging.rs
+++ b/crates/goose-server/src/logging.rs
@@ -13,21 +13,18 @@ use goose::tracing::{langfuse_layer, otlp_layer};
 /// - Console output for development (INFO level)
 /// - Optional Langfuse integration (DEBUG level)
 pub fn setup_logging(name: Option<&str>) -> Result<()> {
-    let _ = goose::logging::cleanup_old_logs("server");
-    let _ = goose::logging::cleanup_old_logs("llm");
-    let log_dir = goose::logging::get_log_directory("server", true)?;
+    let log_dir = goose::logging::prepare_log_directory("server", true)?;
     let timestamp = chrono::Local::now().format("%Y%m%d_%H%M%S").to_string();
-
-    // Create log file name by prefixing with timestamp
     let log_filename = if name.is_some() {
         format!("{}-{}.log", timestamp, name.unwrap())
     } else {
         format!("{}.log", timestamp)
     };
-
-    // Create daily rolling file appender for detailed logs
-    let file_appender =
-        tracing_appender::rolling::RollingFileAppender::new(Rotation::DAILY, log_dir, log_filename);
+    let file_appender = tracing_appender::rolling::RollingFileAppender::new(
+        Rotation::NEVER, // we do manual rotation via file naming and cleanup_old_logs
+        log_dir,
+        log_filename,
+    );
 
     // Create JSON file logging layer
     let file_layer = fmt::layer()
--- a/crates/goose/src/logging.rs
+++ b/crates/goose/src/logging.rs
@@ -9,22 +9,21 @@ use std::time::{Duration, SystemTime};
 ///
 /// # Arguments
 ///
-/// * `component` - The component name (e.g., "cli", "server", "debug")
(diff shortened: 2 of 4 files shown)


sheikhlimon:
I think it solves this #4365 one too... i was working on this then saw new requirements.. so I put it on hold..

crates/goose-server/src/logging.rs
```diff
@@ -8,20 +7,15 @@ use tracing_subscriber::{
...
 /// This includes:
 /// - File-based logging with JSON formatting (DEBUG level)
 /// - Console output for development (INFO level)
 /// - Optional Langfuse integration (DEBUG level)
 pub fn setup_logging(name: Option<&str>) -> Result<()> {
-    // Set up file appender for goose module logs
-    let log_dir = get_log_directory()?;
+    let _ = goose::logging::cleanup_old_logs("server");
+    let _ = goose::logging::cleanup_old_logs("llm");
+    let log_dir = goose::logging::get_log_directory("server", true)?;
```
DOsinga: we actually have two setup_logging methods, one for the server and one for the cli. the get_log_directory is shared though. I think the right move would be to either just have one and use the name here (don't put it in a server or cli folder, just have a server.log, cli.log and an mcp.log) but that's a rather bigger refactor. or we can just rename this function to prepare_log_directory that also does the clean up? definitely need to set up the log rotation also for the cli though

crates/goose-server/src/logging.rs
```diff
@@ -31,9 +25,9 @@ pub fn setup_logging(name: Option<&str>) -> Result<()> {
         format!("{}.log", timestamp)
     };
 
-    // Create non-rolling file appender for detailed logs
+    // Create daily rolling file appender for detailed logs
     let file_appender =
-        tracing_appender::rolling::RollingFileAppender::new(Rotation::NEVER, log_dir, log_filename);
+        tracing_appender::rolling::RollingFileAppender::new(Rotation::DAILY, log_dir, log_filename);
```
DOsinga: I don't think this works - we were previously doing our own timestamping, so this is just going to produce log files of the form:

20251104_143052.log.2025-11-04

so we can use the rotation, but then just get rid of the old timestamp code or just leave it as is

crates/goose/src/logging.rs
```diff
@@ -29,6 +30,42 @@ pub fn get_log_directory(component: &str, use_date_subdir: bool) -> Result<PathB
...
+pub fn cleanup_old_logs(component: &str) -> Result<()> {
+    let base_log_dir = Paths::in_state_dir("logs");
+    let component_dir = base_log_dir.join(component);
+
+    if !component_dir.exists() {
+        return Ok(());
+    }
+
+    let two_weeks = SystemTime::now() - Duration::from_secs(14 * 24 * 60 * 60);
+    let entries = fs::read_dir(&component_dir).context("Failed to read log directory")?;
```
DOsinga: I'd drop the context

crates/goose/src/logging.rs
```diff
@@ -29,6 +30,42 @@ pub fn get_log_directory(component: &str, use_date_subdir: bool) -> Result<PathB
...
+
+    for entry in entries.flatten() {
+        let path = entry.path();
+
+        if let Ok(metadata) = entry.metadata() {
+            if let Ok(modified) = metadata.modified() {
+                if modified < two_weeks {
+                    if path.is_dir() {
```
DOsinga: I would only remove dirs here - that's what we expect and if there are files, maybe somebody put them there for a reason? i.e. copied a log file out of the sub folder for save keeping? also not sure about logging inside of a logger

DOsinga (approved):

alexhancock:
@DOsinga Addressed all feedback. Reaping now works the same way for all three of  `cli, server, llm` using a unified `prepare_log_directory` which also does cleanup. File structure now looks like:

```
13:38:55 ~/.local/state/goose $ tree logs
logs
├── cli
│   └── 2025-11-04
│       ├── 20251104_132253.log
│       └── 20251104_132300.log
├── llm
│   └── 2025-11-04
│       ├── llm_request.0.jsonl
│       ├── llm_request.1.jsonl
│       ├── llm_request.2.jsonl
│       └── llm_request.3.jsonl
└── server
    └── 2025-11-04
        ├── 20251104_133833-goosed.log
        └── 20251104_133844-mcp-developer.log
```

The content of the llm ones is a little different in the `.jsonl` format so I left the files themselves alone for that, but made the directory structure nicer.

crates/goose/src/providers/utils.rs
```diff
@@ -529,7 +527,7 @@ impl RequestLog {
     fn finish(&mut self) -> Result<()> {
         if let Some(mut writer) = self.writer.take() {
             writer.flush()?;
-            let logs_dir = Paths::in_state_dir("logs");
+            let logs_dir = crate::logging::prepare_log_directory("llm", true)?;
```
DOsinga, line 530: we should undo this; the llm logs are special and huge; keep only 10 of them and we send them out with the diagnostics

alexhancock: will do

PR #5570, opened 2025-11-04
Support platform tools through CLI

DOsinga:
Since platform tools are built-in but have their own type you couldn't explicitly enable them through the CLI. This changes that

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -299,22 +299,22 @@ impl CliSession {
     /// * `builtin_name` - Name of the builtin extension(s), comma separated
     pub async fn add_builtin(&mut self, builtin_name: String) -> Result<()> {
         for name in builtin_name.split(',') {
-            let extension_name = name.trim().to_string();
+            let extension_name = name.trim();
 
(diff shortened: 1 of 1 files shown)


alexhancock:
(rebased this because I want to see if the provider tests are more reliable after recent merges)

crates/goose-cli/src/session/mod.rs
```diff
@@ -299,15 +299,24 @@ impl CliSession {
...
+                    name: extension_name.to_string(),
+                    bundled: None,
+                    description: name.to_string(),
+                    available_tools: Vec::new(),
+                }
+            } else {
+                ExtensionConfig::Builtin {
+                    name: extension_name.to_string(),
+                    display_name: None,
+                    timeout: None,
```
alexhancock, line 315: what is the effect of the switch to have no timeout vs the prior

```
Some(goose::config::DEFAULT_EXTENSION_TIMEOUT)
```
?

alexhancock (approved):

crates/goose-cli/src/session/mod.rs
alexhancock: Interested to know, but doesn't seem like a big deal to me as we wouldn't expect the builtins to time out conceptually

Issue #5491, opened 2025-10-30
Scheduler does not run


aprilatmc:
Below is the schedule information. It never ran as scheduled.

<img width="746" height="393" alt="Image" src="https://github.com/user-attachments/assets/9b985ff7-9fed-4199-8cc6-18bb509914f3" />

cbruyndoncx:
I reported similar on windows, scheduler is being worked on. 
What OS are you using ? Seems the template is completely missing, did you not get a standard set of fields to fill out for the bug ?

DOsinga:
does it not run or does it not run at the expected time? I think I've seen timezone issues where if I came back 24 hours later, I could see it had run, just not at the hour I expected

cbruyndoncx:
apparently not at the expected time, but still no results, no session data nothing logged as far as i can tell

<img width="1472" height="1224" alt="Image" src="https://github.com/user-attachments/assets/02b9ceee-462b-42ff-993e-e4f653fe9698" />

DOsinga:
so #5571 is now merged and should fix this. the main issue seemed to be that we tried to run schedules in UTC

DOsinga:
closing this since fixed on main

cbruyndoncx:
@DOsinga when i entered to edit a schedule, the hour selector, upon edit starts with 1 , I tried entering 9 AM, but it was on PM, so i got Honk Error happened, and nothing else. I think it was because 19 hrs PM is non existent.
Entering minutes, it sets to 00 adding my 09, as you can see in the crontab . So using hte keyboard messes things up on windows. 

<img width="1838" height="794" alt="Image" src="https://github.com/user-attachments/assets/ee831baa-4851-4706-b2a9-393fb572ecae" />

cbruyndoncx:
As soon as you edit and get an illegal time result, it gives 
<img width="1040" height="814" alt="Image" src="https://github.com/user-attachments/assets/1f97698c-292c-4c20-bb37-77a46dcf99cd" />

cbruyndoncx:
now the session runs on schedule, but i dont see results, when i run interactively, i get , how do i debug this further in the GUI ?

<img width="950" height="366" alt="Image" src="https://github.com/user-attachments/assets/acab28f7-2c09-4c6a-95e4-e369b97b64a9" />

cbruyndoncx:
It did run on schedule, and interactively with error, but no history: 

[...]

PR #5571, opened 2025-11-04
Scheduler cleanup

DOsinga:
Cleaning up the scheduler and verify that it actually works


DOsinga:
fixes #5491 (hopefully)

crates/goose/src/scheduler.rs
```diff
@@ -186,427 +149,253 @@ impl Scheduler {
...
+        let cron_parts: Vec<&str> = job.cron.split_whitespace().collect();
+        let cron = match cron_parts.len() {
+            5 => {
+                tracing::warn!(
+                    "Job '{}' has legacy 5-field cron '{}', converting to 6-field",
+                    job.id,
+                    job.cron
+                );
+                format!("0 {}", job.cron)
+            }
```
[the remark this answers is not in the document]
jamadeo: yeah, I don't think this needs a warning at all

crates/goose/src/scheduler.rs
```diff
@@ -186,427 +146,252 @@ impl Scheduler {
...
-            )));
-        }
-        if !original_recipe_path.is_file() {
-            return Err(SchedulerError::RecipeLoadError(format!(
-                "Original recipe source is not a file: {}",
-                original_job_spec.source
-            )));
-        }
+    fn create_cron_task(&self, job: ScheduledJob) -> Result<Job, SchedulerError> {
+        let job_for_task = job.clone();
```
jamadeo, line 166: I don't think this needs to own `job`

`create_cron_task(&self, job: &ScheduledJob)`

crates/goose/src/scheduler.rs
```diff
@@ -186,427 +146,252 @@ impl Scheduler {
...
+            .and_then(|ext| ext.to_str())
+            .unwrap_or("yaml");
+
+        let destination_filename = format!("{}.{}", original_job_spec.id, original_extension);
+        let destination_recipe_path = scheduled_recipes_dir.join(destination_filename);
+
+        fs::copy(original_recipe_path, &destination_recipe_path)?;
+
+        let mut stored_job = original_job_spec;
+        stored_job.source = destination_recipe_path.to_string_lossy().into_owned();
```
jamadeo, line 308: seems like ScheduledJob.source should be a PathBuf not a String

crates/goose/src/scheduler.rs
```diff
@@ -186,427 +146,252 @@ impl Scheduler {
...
         }
 
-        let list: Vec<ScheduledJob> = serde_json::from_str(&data).map_err(|e| {
-            SchedulerError::PersistError(format!("Failed to deserialize schedules.json: {}", e))
-        })?;
+        let list: Vec<ScheduledJob> = match serde_json::from_str(&data) {
+            Ok(jobs) => jobs,
+            Err(e) => {
+                tracing::error!(
+                    "Failed to parse schedules.json: {}. Starting with empty schedule list.",
```
jamadeo, line 351: I guess this should say the actual path

crates/goose/src/scheduler.rs
```diff
@@ -619,23 +404,26 @@ impl Scheduler {
...
-            Ok(())
-        } else {
-            Err(SchedulerError::JobNotFound(id.to_string()))
+        let path = Path::new(&recipe_path);
+        if path.exists() {
+            fs::remove_file(path)?;
         }
+
+        persist_jobs(&self.storage_path, &self.jobs).await?;
+        Ok(())
```
jamadeo, line 426: so we have to remove it from three or four places but each might fail, do we want to not abort on the first fail?

crates/goose/src/scheduler.rs
```diff
@@ -155,22 +115,22 @@ pub struct ScheduledJob {
     pub process_start_time: Option<DateTime<Utc>>,
 }
 
-async fn persist_jobs_from_arc(
+async fn persist_jobs(
```
jamadeo, line 118: I was going to say why isn't this a method on the scheduler, but I see it's because we are calling it in the job tasks where we only give it a pointer to the internal map

I probably would have one this where the job tasks write to a channel and the scheduler just has one task that saves whenever updates come through, that feels a bit safer and cleaner, but this is already a big refactor. Maybe for another time.

crates/goose/src/scheduler.rs
```diff
@@ -1045,156 +642,118 @@ struct JobExecutionError {
...
     jobs_arc: Option<Arc<Mutex<JobsMap>>>,
     job_id: Option<String>,
-) -> std::result::Result<String, JobExecutionError> {
-    tracing::info!("Executing job: {} (Source: {})", job.id, job.source);
+    cancel_token: CancellationToken,
+) -> Result<String, JobExecutionError> {
+    if job.source.is_empty() {
+        // Shortcut for testing
+        return Ok(job.id.to_string());
+    }
```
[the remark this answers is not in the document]
jamadeo: is this actually here for testing? it does seem weird to just do nothing if the job.source is empty

crates/goose/src/scheduler.rs
```diff
@@ -1021,180 +619,73 @@ impl Scheduler {
...
-                })
-            }
-            _ => Err(JobExecutionError {
-                job_id: job.id.clone(),
-                error: format!(
-                    "Unsupported recipe file extension '{}' for: {}",
-                    extension, job.source
-                ),
-            }),
+            "json" | "jsonl" => serde_json::from_str(&recipe_content)?,
```
jamadeo, line 655: can recipes be .jsonl files? also do we need this check, can we not always call serde_yaml?

crates/goose/src/scheduler.rs
```diff
@@ -1260,10 +737,9 @@ async fn run_scheduled_job_internal(
         .apply()
         .await
     {
-        tracing::error!("[Job {}] Failed to update session metadata: {}", job.id, e);
+        tracing::error!("Failed to update session: {}", e);
     }
 
-    tracing::info!("Finished job: {}", job.id);
```
jamadeo, line 1266: I know we're all about reducing the logging but it does seem like logging the start/stop of each job execution would not be a bad idea.

Or is it that we log it outside of this function?

jamadeo (approved):
nice

the nits mostly refer to old code, feel free to ignore them

crates/goose/src/scheduler.rs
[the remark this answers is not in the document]
DOsinga: I'm leaving it in as this was a cause of errors. it should all be rational now

crates/goose/src/scheduler.rs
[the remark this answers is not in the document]
DOsinga: hmm, yeah

so I wanted to have a test that tests the actual scheduler. but I don't want to do anything. I experimented with various test only code, but that got my warnings of not used code, so if you have a better suggestion

DOsinga:
thanks @jamadeo - what are you saying is all true. but I want to get this in, so I am mostly going to leave it for another time. if that :)

PR #5574, opened 2025-11-05
fix: gemini flash -> pro for mcp smoke tests

alexhancock:
Should be more reliable at the cost of a small amount of build time


michaelneale (approved):

katzdave:
Still seems to be failing with pro; should we just drop/modify that specific test condition?

alexhancock:
@katzdave The failure was actually unrelated to the model here. I think this should be more reliable. If it proves not to be once in main + everyone's PRs we can revisit.

PR #5575, opened 2025-11-05
fix: move goosehints/AGENTS.md handling to goose, and out of developer extension

alexhancock:
Move `.goosehints` and `AGENTS.md` etc handling out of the developer extension and into goose itself. This shouldn't depend on an extension being active to work.

This works at present, but I want to take a few revs on the  the new code before merge.


crates/goose/src/hints/load_hints.rs
```diff
@@ -350,6 +341,57 @@ mod tests {
         assert!(hints.contains("Subdir .goosehints content"));
     }
 
+    #[test]
+    #[serial]
```
DOsinga: serial doesn't protect you; it just means that this test will run serial with all the other serial tests. we're changing the current directory here, so that will make other tests flaky. do we need to change the current directory?

crates/goose/src/hints/load_hints.rs
```diff
@@ -350,6 +341,57 @@ mod tests {
...
+
+        // Should contain the .goosehints content
+        assert!(hints.contains("Project Information"));
+        assert!(hints.contains("Additional instructions here"));
+
+        // Should contain the referenced files' content
+        assert!(hints.contains("# Project README"));
+        assert!(hints.contains("This is the project documentation"));
+        assert!(hints.contains("# Development Guide"));
+        assert!(hints.contains("Follow these steps"));
```
DOsinga: no need to check 2 lines for each file (or even inject two lines)

crates/goose/src/hints/load_hints.rs
```diff
@@ -350,6 +341,57 @@ mod tests {
...
+            temp_dir.path(),
+            &[GOOSE_HINTS_FILENAME.to_string()],
+            &gitignore,
+        );
+
+        // Should contain the .goosehints content
+        assert!(hints.contains("Project Information"));
+        assert!(hints.contains("Additional instructions here"));
+
+        // Should contain the referenced files' content
```
DOsinga: should comments that cover the asserts

crates/goose/src/hints/load_hints.rs
```diff
@@ -150,14 +147,8 @@ mod tests {
     fn test_global_goosehints() {
```
DOsinga: delete this test please

crates/goose/src/hints/load_hints.rs
```diff
@@ -60,16 +61,10 @@ pub fn load_hint_files(
 
     for hints_filename in hints_filenames {
         // Global hints
-        // choose_app_strategy().config_dir()
+        // Paths::config_dir()
         // - macOS/Linux: ~/.config/goose/
         // - Windows:     ~\AppData\Roaming\Block\goose\config\
```
DOsinga: delete

crates/goose/src/hints/load_hints.rs
```diff
@@ -60,16 +61,10 @@ pub fn load_hint_files(
...
-        // keep previous behavior of expanding ~/.config in case this fails
-        let global_hints_path = choose_app_strategy(crate::APP_STRATEGY.clone())
-            .map(|strategy| strategy.in_config_dir(hints_filename))
-            .unwrap_or_else(|_| {
-                let path_str = format!("~/.config/goose/{}", hints_filename);
-                PathBuf::from(shellexpand::tilde(&path_str).to_string())
-            });
+        let global_hints_path = Paths::in_config_dir(hints_filename);
 
         if let Some(parent) = global_hints_path.parent() {
```
DOsinga: why are we creating this directory? we're just reading from it?

crates/goose/src/agents/agent.rs
```diff
@@ -830,6 +831,34 @@ impl Agent {
         session: Session,
         cancel_token: Option<CancellationToken>,
     ) -> Result<BoxStream<'_, Result<AgentEvent>>> {
+        // Load hint files and add to system prompt
+        let config = Config::global();
```
DOsinga: I don't think we should do this in reply_internal - that one is big enough as is. instead add this functionality to the prompt-manager itself and then add it to the call to the prompt manager:

        let prompt_manager = self.prompt_manager.lock().await;
        let system_prompt = prompt_manager
            .builder(model_name)
            .with_extensions(extensions_info.into_iter())
            .with_frontend_instructions(self.frontend_instructions.lock().await.clone())
            .with_extension_and_tool_counts(extension_count, tool_count)
            .build();

DOsinga (approved):

PR #5581, opened 2025-11-05
Add uv for uvx in Justfile

jamadeo:
The Justfile recipes `get-tag-version` and `prepare-release` need uvx, and we don't manage uv with hermit. Any of those other things could arguably change, but for now, just add it where we need it.


DOsinga (approved):
v1 -> v6 eh?

PR #5583, opened 2025-11-05
fix: make ACP cancellation work

alexhancock:
Fixes cancellation by properly storing the cancel token on the session. Prior to this, hitting the cancel button in clients like zed etc didn't do anything during tool calling loops. Now it works.


alexhancock:
Replacing this with https://github.com/block/goose/pull/5588 which also makes `ToolCallLocations` work with line numbers and file names for the developer extension tools

PR #5588, opened 2025-11-05
acp: ToolCallLocations and working cancellation

alexhancock:
Add more information to `ToolCallLocations` such as line numbers etc. This makes it possible to follow work goose is doing in an ACP client.

Also, make cancellation work by attaching a cancel token to the session. Right now hitting the stop button in ACP clients does nothing.

Replaces https://github.com/block/goose/pull/5583

Demo

https://github.com/user-attachments/assets/62c2504e-310b-49d2-9c56-e246a1beac8e

### DIFF
--- a/crates/goose-cli/src/commands/acp.rs
+++ b/crates/goose-cli/src/commands/acp.rs
@@ -7,10 +7,11 @@ use goose::agents::{Agent, SessionConfig};
 use goose::config::{get_all_extensions, Config};
 use goose::conversation::message::{Message, MessageContent};
 use goose::conversation::Conversation;
+use goose::mcp_utils::ToolResult;
 use goose::providers::create;
 use goose::session::session_manager::SessionType;
 use goose::session::SessionManager;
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/commands/acp.rs
```diff
@@ -33,6 +35,116 @@ struct GooseAcpAgent {
...
+            caps.get(2)?.as_str().parse::<usize>().ok()?
+        };
+        return Some((start, end));
+    }
+    None
+}
+
+/// Extract the first line number from code snippet (e.g., "123: some code")
+fn extract_first_line_number(text: &str) -> Option<usize> {
+    // Look for pattern like "123: " at the start of a line within a code block
```
michaelneale, line 140: can probably tell goose to drop the commentary, as it is just saying what is in the code (annoys me so much it does this, but I always have to tell it tp rip it out0

michaelneale (approved):
can probably rip out the un-necessary LLM comments they like to put in, but lets go!

Issue #5576, opened 2025-11-05
Excessive Database Writes from Agent Stream Messages in goose Web


blackgirlbytes:
This is based on a [Discord thread](https://discord.com/channels/1287729918100246654/1287729920319033345/1434849335933145132) - reported by a community member

 goose web is storing every message yielded by the agent's reply stream as a separate entry in the session database, resulting in excessive database writes and fragmented conversation history.

## Environment
- **Component**: goose Web Server (`goose-server`)
- **File**: `crates/goose-server/src/routes/reply.rs`
- **Endpoint**: `/reply` (streaming chat endpoint)

## Problem Description
The `/reply` endpoint processes agent responses by consuming a stream of messages from `agent.reply()`. Currently, **every message yielded by this stream is immediately added to the session database**, causing:

1. **Excessive I/O operations**: Each streamed message triggers a background database write to local storage
2. **Fragmented conversation history**: Sessions contain multiple intermediate messages instead of clean conversation turns
3. **Performance degradation**: Multiple file writes for each message in a single response

## Current Behavior
When an agent processes a request that involves tool calls, the stream yields multiple messages:
- Initial assistant response
- Tool call messages  
- Tool response messages
- Follow-up assistant responses
- Additional tool interactions (if any)

**All of these messages are currently being stored as separate database entries.**

## The contributor suggested that
Ideally, only the **final consolidated agent reply** should be stored in the session database, representing the complete response to the user's request. Intermediate tool calls and responses are part of the agent's internal reasoning process and shouldn't clutter the conversation history.

## Code Location
**File**: `crates/goose-server/src/routes/reply.rs`  
**Function**: `handler()` (the `/reply` endpoint handler)  
**Problematic section** (lines ~165-190):

## Impact
- **Performance**: Multiple unnecessary file writes to local storage per user request
- **Storage**: Inflated session files with redundant intermediate messages  
- **Resource usage**: Increased disk I/O operations on the local machine

## Reproduction Steps
1. Start Goose Web server locally
2. Send a request to `/reply` that triggers tool usage (e.g., a request requiring file operations)
3. Observe the session file - it will contain multiple message entries for a single user request
4. Check server logs - multiple "storing session history" operations will be logged

DOsinga:
FWIW I don't think it would actually reduce database operations if we fix this since we would still want to write the message fragments as they come to the DB, but yeah, it looks like we end up writing a new message for every fragment that comes in

michaelneale:
[...]

PR #5598, opened 2025-11-06
fix: write messages from session in a batch for efficiency

michaelneale:
fixes: https://github.com/block/goose/issues/5576

this will write it at once, vs message by message. I have tested this cli and desktop, and even with mobile following along reading the DB, and seemed fine (fast even?)

@DOsinga I don't know what downside of doing this is vs status quo, I don't think it is a massive saving, "felt fast" but surely there aren't that many messages coming in.

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -1177,9 +1177,7 @@ impl Agent {
                     }
                 }
 
-                for msg in &messages_to_add {
-                    SessionManager::add_message(&session_config.id, msg).await?;
-                }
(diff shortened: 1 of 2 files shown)


DOsinga:
I actually don't think this is the fix for the reported issue; that was talking about how web hammers the database. if you do a web session and then open that web session in goose, you can clearly see what happens. web writes streaming messages in one by one and then agent writes the resulting message. so the fix is really simple:

https://github.com/block/goose/pull/5601

alexhancock:
#5601 has been merged. going to close this out now, but feel free to reopen if there is still something missing @michaelneale

michaelneale:
yep, not needed, and not in right area.

PR #5601, opened 2025-11-06
Avoid web double write

DOsinga:
Goose web was writing incoming messages to the session, rather than relying on the agent to do it


alexhancock (approved):

alexhancock:
`gemini-2.5-flash` failed the mcp test in this one. will get less flaky now that we have merged https://github.com/block/goose/pull/5574

merging this one

PR #5606, opened 2025-11-06
Token state not showing on load, or after message is finished.

### LINKED ISSUE #5604: (release 1.13.0) token counts are zero after first message
After the first message in a new chat, the token counts are 0.

I've verified that they do come from goosed as zero:

```
"token_state":{"inputTokens":0,"outputTokens":0,"totalTokens":0,"accumulatedInputTokens":0,"accumulatedOutputTokens":0,"accumulatedTotalTokens":0}
```

katzdave:
Two fixes:

On client side use || an issue with truthyness of 0.

On server add token_state to Finish. This ensures the last turn is always included; update_session_metrics doesn't run until after the last Message.

fixes #5604

### DIFF
--- a/crates/goose-server/src/routes/reply.rs
+++ b/crates/goose-server/src/routes/reply.rs
@@ -161,7 +161,11 @@ async fn get_token_state(session_id: &str) -> TokenState {
             accumulated_total_tokens: session.accumulated_total_tokens.unwrap_or(0),
         },
         Err(e) => {
-            tracing::warn!("Failed to fetch session token state for {}: {}", session_id, e);
+            tracing::warn!(
+                "Failed to fetch session token state for {}: {}",
+                session_id,
+                e
+            );
             TokenState {
                 input_tokens: 0,
                 output_tokens: 0,
--- a/crates/goose-server/src/routes/reply.rs
+++ b/crates/goose-server/src/routes/reply.rs
@@ -133,6 +133,7 @@ pub enum MessageEvent {
     },
     Finish {
         reason: String,
+        token_state: TokenState,
     },
     ModelChange {
         model: String,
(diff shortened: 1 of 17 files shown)


crates/goose-server/src/routes/reply.rs
```diff
@@ -149,6 +150,34 @@ pub enum MessageEvent {
...
+                e
+            );
+            TokenState {
+                input_tokens: 0,
+                output_tokens: 0,
+                total_tokens: 0,
+                accumulated_input_tokens: 0,
+                accumulated_output_tokens: 0,
+                accumulated_total_tokens: 0,
+            }
```
jamadeo: TokenState could derive Default (if it doesn't already) and this could be TokenState::default()

then this could alaso be `get_session(session_id).await.map( .. ).unwrap_or_default()`

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -443,12 +443,12 @@ function BaseChatContent({
...
             setView={setView}
-            totalTokens={tokenState?.totalTokens ?? sessionTokenCount}
+            totalTokens={tokenState?.totalTokens || sessionTokenCount}
             accumulatedInputTokens={
-              tokenState?.accumulatedInputTokens ?? sessionInputTokens ?? localInputTokens
+              tokenState?.accumulatedInputTokens || sessionInputTokens || localInputTokens
             }
             accumulatedOutputTokens={
-              tokenState?.accumulatedOutputTokens ?? sessionOutputTokens ?? localOutputTokens
+              tokenState?.accumulatedOutputTokens || sessionOutputTokens || localOutputTokens
```
jamadeo, line 451: What are these three sources of token counts and do we need them all? tokenState, sessionOutputTokens, localOutputTokens

jamadeo (approved):
I'm not quite following the different client-side sources of token counts, do we have a clear idea on why they all exist?

ui/desktop/src/components/BaseChat.tsx
DOsinga: fwiw this is all gone in basecamp2

katzdave: Im not convinced sessionOutputTokens is working right; I think that's the reason we were seeing this issue. The stream seems like a reasonable place to do this (although I suspect that might be used for session load because the stream code path is not hit there). 

Also really don't know why we have token estimates in the client.

Will take a deeper pass and try to clean shop here.

crates/goose-server/src/routes/reply.rs
katzdave: Done

ui/desktop/src/components/BaseChat.tsx
jamadeo: > fwiw this is all gone in basecamp2

👍

jamadeo: If it's all going away soon, I wouldn't bother tracking it all down if this gets it working

DOsinga:
does this work in basecamp2?

jamadeo:
> does this work in basecamp2?

If I set ALPHA=true, I still see the issue with and without this PR

PR #5617, opened 2025-11-06
Run the developer mcp in a login shell

jamadeo:
I'm not sure we want to do this, but if we do, this would probably be the way

### DIFF
--- a/crates/goose-mcp/src/mcp_server_runner.rs
+++ b/crates/goose-mcp/src/mcp_server_runner.rs
@@ -3,6 +3,9 @@ use crate::{
 };
 use anyhow::{anyhow, Result};
 use rmcp::{transport::stdio, ServiceExt};
+use std::{env, os::unix::process::CommandExt, process::Command};
+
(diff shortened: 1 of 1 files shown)


michaelneale:
I think this may be worthwhile, previously, it would inherit it from the forced loading on macos, but now people may find it is missing things they expect are on their $PATH, so this would be closer to what you run if you run from cli, so principle of least surprise: I think yes? 

What is downside to this?

wpfleger96:
I just opened #5680 before I realized this draft PR was already open for this, FWIW I am definitely in favor of running a login shell bc I've found the developer extension in goose UI/desktop to be pretty limited now - it can no longer run any non system PATH command which includes `rg`, `node`/`npm`, `pip`/`uv`, `jq`, `yq`, etc. Not having `rg` in particular is tough since goose makes heavy use of it during local development and when exploring local codebases

This also breaks Presidio MCP for Block users in UI/desktop, because that MCP implementation uses `subprocess.run()` to launch `sq presidio` which similarly fails since `sq` is also installed via Homebrew

jamadeo:
done in #5774

PR #5622, opened 2025-11-07
feat: remove hints from recipe prompts

alexhancock:
Right now hints are included in the prompt for recipes. I think it makes sense for recipes to be as consistent as possible in any dir, so this change removes hints.

Anything we should consider here?


lifeizhou-ap:
> Right now hints are included in the prompt for recipes. I think it makes sense for recipes to be as consistent as possible in any dir, so this change removes hints.
> 
> Anything we should consider here?

I agree. The recipe is built from the context in the current session, not from project knowledge in the current dir

lifeizhou-ap (approved):

DOsinga (approved):

PR #5623, opened 2025-11-07
Add ML-based prompt injection detection 

dorien-koelemeijer:
## Summary
This PR adds BERT model prompt injection evaluation alongside the existing pattern-based approach.

### Context
https://docs.google.com/document/d/1GNvriNWLAaJUMpWE1heBapxFshQEED5ZYfn1ixJvGCE/edit?tab=t.0#heading=h.rj4p6llxqvrq 

### Key Changes
* Updated PromptInjectionScanner to support both pattern-based + ML-based scanning, using the highest confidence score from either method to determine threats. ML-based scanning also looks at recent user messages in addition to tool call content. 
* Added ClassificationClient - Generic HTTP client that is compatible with HuggingFace's text classification API, supporting both internal users (Gondola hosted BERT models) and external/oss users (direct endpoints, such as using the hugging face API). The ML inference team is creating an additional wrapper endpoint that allows us to use their BatchInfer API with the same inputs and outputs as the Hugging Face text classification API. 
* Updated UI settings to allow users to configure what models to use if they decide to enable ML-based prompt injection detection. Have included `BERT` in the name of the variables so we can create a follow-up PR to also allow LLM-as-a-judge type prompt injection detection by re-using people's providers, and hopefully prevent confusing names of variables. 

### Planned follow-up PRs / Related PRs
* Add a bunch of variables to goose-releases (will open PR shortly)
* Re-use provider to do ML-based prompt injection detection rather than users requiring setup of the BERT models (mostly for oss/external users)
* Provide reference implementation for users who want to run their BERT models locally (I've already prepared this but didn't want to make this PR super huge + I am not sure where best to store this info in the repo)
* Waiting for this PR to get merged: https://github.com/squareup/gondola/pull/1085

### Type of Change
- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Screenshots of changes

**Internal:**
Screenshot 2025-11-26 at 11 51 36 am

Screenshot 2025-11-26 at 11 54 04 am

**External:**
Screenshot 2025-11-26 at 11 57 58 am

Screenshot 2025-11-26 at 11 58 46 am

### DIFF
--- a/documentation/docs/guides/config-files.md
+++ b/documentation/docs/guides/config-files.md
@@ -50,10 +50,6 @@ The following settings can be configured at the root level of your config.yaml f
 <!-- | `SECURITY_PROMPT_CLASSIFIER_ENABLED` | Enable ML-based prompt injection detection for advanced threat identification | true/false | false | No | -->
 <!-- | `SECURITY_PROMPT_CLASSIFIER_MODEL` | Specify the BERT ML model to use for prompt injection detection | String | "ProtectAI DeBERTa" | No | -->
 
-:::info Automatic Multi-Model Configuration
-The experimental [AutoPilot](/docs/guides/multi-model/autopilot) feature provides intelligent, context-aware model switching. Configure models for different roles using the `x-advanced-models` setting.
-:::
-
 ## Experimental Features
 
 These settings enable experimental features that are in active development. These may change or be removed in future releases.


crates/goose/src/security/prompt_ml_detector.rs
```diff
@@ -0,0 +1,125 @@
...
+        let exp_malicious = logits[1].exp();
+        let sum = exp_safe + exp_malicious;
+        let confidence = (exp_malicious / sum) as f32;
+
+        tracing::info!(
+            prob_safe = %(exp_safe / sum),
+            prob_malicious = %(exp_malicious / sum),
+            confidence = %confidence,
+            "ML detection raw results"
+        );
```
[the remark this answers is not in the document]
dorien-koelemeijer: Will have a more detailed look into metrics/logs for changes on this PR

crates/goose/src/security/gondola.rs
```diff
@@ -0,0 +1,272 @@
...
+pub struct DoubleListValue {
+    pub double_values: Vec<f64>,
+}
+
+impl GondolaProvider {
+    pub fn new() -> Result<Self> {
+        let config = Config::global();
+
+        let endpoint = config
+            .get_param::<String>("GONDOLA_ENDPOINT")
```
dorien-koelemeijer: Todo: add this to goose-releases. Also, how do I deal with this for non-internal goose?

michaelneale:
I don't think this should be added as a provider, as we don't want it grouped with others, this needs to be its own part of the security module I think

crates/goose/src/security/gondola.rs
michaelneale: what is gondola? is it something people would potentialy host themselves?

michaelneale: is it possible (if bert models small enough) that these could run locally (for OSS) or via ollama etc?

dorien-koelemeijer: https://github.com/squareup/gondola

It's a Cash managed platform for hosting and training models. I talked with Douwe about running these models locally (which is what we did in the first iteration of this), but they're too big to be downloaded and run locally.

dorien-koelemeijer:
> I don't think this should be added as a provider, as we don't want it grouped with others, this needs to be its own part of the security module I think

Okay, that sounds good, will move it there :)

crates/goose/src/security/gondola.rs
michaelneale: that looks like a closed source thing so can't be referred to directly in code, need an abstraction that can work with it - is the api standard?

DOsinga: yeah, this does seem like a concern; we can't add something to the OSS code base that is not accessible to everybody

crates/goose/src/security/patterns.rs
```diff
@@ -436,7 +435,7 @@ mod tests {
     #[test]
     fn test_rm_rf_detection() {
```
DOsinga: delete these tests

ui/desktop/src/main.ts
```diff
@@ -545,6 +568,7 @@ const createChat = async (
           recipeId: recipeId,
           recipeDeeplink: recipeDeeplink,
           scheduledJobId: scheduledJobId,
+          ML_MODEL_MAPPING: process.env.ML_MODEL_MAPPING,
```
michaelneale: shoudl this be called something appropriately scoped - like it is SECURITY_ something like others?

dorien-koelemeijer: yeah, good point, will update, thanks!

michaelneale:
at first I wasn't sure about all the complexity in the GUI for this, but it looks pretty nice. What happens if one picks enable prompt injection, is the list pre-populated and will it "Just Work" by downloading a model? (for default users)

crates/goose/src/security/classification_client.rs
```diff
@@ -0,0 +1,326 @@
...
+        assert!(client.is_ok());
+        let client = client.unwrap();
+        assert_eq!(client.endpoint_url, "http://localhost:8000/classify");
+        assert!(client.auth_token.is_none());
+    }
+
+    #[test]
+    fn test_from_model_name_without_mapping() {
+        // Should fail when SECURITY_ML_MODEL_MAPPING is not set
+        std::env::remove_var("SECURITY_ML_MODEL_MAPPING");
```
DOsinga: Don't use environment variables in tests

documentation/docs/guides/security/classification-api-spec.md
```diff
@@ -0,0 +1,113 @@
+# Classification API Specification
+
+This document defines the API that Goose uses for ML-based prompt injection detection.
+
+## Overview
+
+Goose requires a classification endpoint that can analyze text and return a score indicating the likelihood of prompt injection. This API follows the **HuggingFace Inference API format** for text classification, making it compatible with HuggingFace Inference Endpoints to allow for easy usage in OSS goose.
```
DOsinga: OSS?

crates/goose/src/security/prompt_ml_detector.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl MlDetector {
+    pub fn new(client: ClassificationClient) -> Self {
+        Self { client }
+    }
+
+    pub fn new_from_config() -> Result<Self> {
+        let config = Config::global();
+
+        // Filter out empty strings - treat them as None
```
DOsinga: why?

crates/goose/src/security/prompt_ml_detector.rs
```diff
@@ -0,0 +1,68 @@
...
+        let model_name = config
+            .get_param::<String>("SECURITY_PROMPT_BERT_MODEL")
```
DOsinga: is there a reason we put BERT into these variables? that feels like an implementation detail, no?

documentation/docs/guides/security/classification-api-spec.md
```diff
@@ -0,0 +1,113 @@
...
+# [[{"label": "INJECTION", "score": 0.98}, {"label": "SAFE", "score": 0.02}]]
+```
+
+## Implementation Guidelines - TBD
+
+### Text Preprocessing
+
+Implementations SHOULD:
+- Normalize whitespace (collapse multiple spaces/newlines)
+- Trim leading/trailing whitespace
```
DOsinga: should it? leading whitespaces can have meaning in prompts for sure

crates/goose/src/security/classification_client.rs
```diff
@@ -0,0 +1,326 @@
...
+        let classification_response: ClassificationResponse = response
+            .json()
+            .await
+            .context("Failed to parse classification response")?;
+
+        let batch_result = classification_response
+            .first()
+            .context("Classification API returned empty response")?;
+
+        let is_probabilities = batch_result
```
DOsinga, line 167: how does this work?

crates/goose/src/security/classification_client.rs
```diff
@@ -0,0 +1,326 @@
...
+                label: label.label.clone(),
+                score: exp_score / sum_exp,
+            })
+            .collect();
+
+        Ok(normalized)
+    }
+}
+
+#[cfg(test)]
```
DOsinga: do these test test anything

crates/goose/src/security/scanner.rs
```diff
@@ -11,167 +15,216 @@ pub struct ScanResult {
...
+    }
 
-        let explanation = if matches.len() > 3 {
-            format!(
-                "Detected {} security threats:\n{}\n... and {} more",
-                matches.len(),
-                explanations.join("\n"),
-                matches.len() - 3
-            )
+    fn select_highest_confidence_result(
```
DOsinga, line 172: this should probably just be inlined

DOsinga (commented):
this strikes me as all more complicated than it maybe has to be? couldn't we just have one endpoint in the config where we post to and then just call that? scan_with_ml calls ml_detector.scan which calls classify which does the http call, it just all seems a bit much for something we mostly want to see if it works.

dorien-koelemeijer:
> this strikes me as all more complicated than it maybe has to be? couldn't we just have one endpoint in the config where we post to and then just call that? scan_with_ml calls ml_detector.scan which calls classify which does the http call, it just all seems a bit much for something we mostly want to see if it works.

Could you give an example of what endpoint this would be, if we would have just one endpoint? It'd be easier to give an answer if you're a little more specific about what you're thinking.

documentation/docs/guides/security/classification-api-spec.md
dorien-koelemeijer: Non-internal users

crates/goose/src/security/prompt_ml_detector.rs
dorien-koelemeijer: There will be other types of ML scanning potentially (i.e. llm-as-a-judge approach), so didn't want to call it generic ML scanning. Happy with suggestions though on other naming, I thought BERT was still generic enough/not so much of an implementation detail, but maybe I'm wrong there

crates/goose/src/security/prompt_ml_detector.rs
dorien-koelemeijer: Because of the different configs (i.e. endpoint vs model name). Could probably come up with a different solution here though.

crates/goose/src/security/classification_client.rs
dorien-koelemeijer: It checks whether results that are returned are in a range between 0 and 1, to determine whether we should convert them or not (gondola-hosted models don't do this conversion in their classify endpoint, whereas HF for instance does).

documentation/docs/guides/security/classification-api-spec.md
dorien-koelemeijer: Generally improves model consistency (I think I will actually remove this from the classification API spec doc for now, this chunking and normalization probably something we need to solve in goose)

ui/desktop/src/components/settings/security/SecurityToggle.tsx
```diff
@@ -1,26 +1,65 @@
...
+      const mappingEnv = window.appConfig?.get('SECURITY_ML_MODEL_MAPPING') as string | undefined;
+      if (mappingEnv) {
+        const mapping = JSON.parse(mappingEnv);
+        return Object.keys(mapping).map((modelName) => ({
+          value: modelName,
+          label: modelName,
+        }));
+      }
+    } catch (error) {
+      console.warn('Failed to parse SECURITY_ML_MODEL_MAPPING:', error);
```
DOsinga: should we drop this on the ground?

documentation/docs/guides/security/classification-api-spec.md
DOsinga: yeah, I understand, but we should write the documentation for the OSS people. I would just drop anything after ", making"

crates/goose/src/security/scanner.rs
```diff
@@ -11,167 +16,254 @@ pub struct ScanResult {
...
+        } else {
+            "Security threat detected".to_string()
+        }
+    }
+
+    fn extract_user_messages(&self, messages: &[Message], limit: usize) -> Vec<String> {
+        messages
+            .iter()
+            .rev()
+            .filter(|m| matches!(m.role, rmcp::model::Role::User))
```
DOsinga: I thought I had mentioned this, but can't find it back; you probably want to use effective_role here, since tool responses are also user messages for some reason

DOsinga (approved):

crates/goose/src/security/scanner.rs
dorien-koelemeijer: Thanks, didn't realise this, have addressed

documentation/docs/guides/config-files.md
```diff
@@ -45,8 +45,14 @@ The following settings can be configured at the root level of your config.yaml f
 | `GOOSE_AUTO_COMPACT_THRESHOLD` | Set the percentage threshold at which goose [automatically summarizes your session](/docs/guides/sessions/smart-context-management#automatic-compaction). | Float between 0.0 and 1.0 (disabled at 0.0)| 0.8 | No |
 | `otel_exporter_otlp_endpoint` | OTLP endpoint URL for [observability](/docs/guides/environment-variables#opentelemetry-protocol-otlp) | URL (e.g., `http://localhost:4318`) | None | No |
 | `otel_exporter_otlp_timeout` | Export timeout in milliseconds for [observability](/docs/guides/environment-variables#opentelemetry-protocol-otlp) | Integer (ms) | 10000 | No |
-| `SECURITY_PROMPT_ENABLED` | Enable [prompt injection detection](/docs/guides/security/prompt-injection-detection) to identify potentially harmful commands | true/false | false | No |
-| `SECURITY_PROMPT_THRESHOLD` | Sensitivity threshold for [prompt injection detection](/docs/guides/security/prompt-injection-detection) (higher = stricter) | Float between 0.01 and 1.0 | 0.7 | No |
+| `SECURITY_PROMPT_ENABLED` | Enable prompt injection detection to identify potentially harmful commands | true/false | false | No |
```
dianed-square: @dorien-koelemeijer Can you please revert the changes to this existing topic for now? Otherwise the docs will go out before this feature is available in a release.

We'll add them back in after the feature is released (except the note below because autopilot has been removed)

documentation/docs/guides/security/classification-api-spec.md
```diff
@@ -0,0 +1,89 @@
+# Classification API Specification
```
dianed-square:
```suggestion
---
title: Classification API Specification
unlisted: true
---
```

documentation/docs/guides/config-files.md
dorien-koelemeijer: Thanks for the comments! Do you mean it's best to revert all changes to this file for now?

dianed-square (approved):
Thanks!

PR #5624, opened 2025-11-07
fix: propagate session ID to LLM request headers in rename task

codefromthecrypt:
## Summary

Wrap session name generation task in session_context so that session ID injection works via api_client.rs. 

Without this, rename task requests lacked the goose-session-id header despite having the session ID as a parameter. 

This means you wouldn't get all requests grouped by the same ID, as the rename task is orphaned. By backfilling tests on this, we also pave the way to test more difficult otel trace propagation.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

I know what I want, but still struggle with rust, so I used an agent to investigate the several ways of how to do what I want. I also used it to adapt some similar mermaid diagrams to this, so it is easy to understand the mechanics.

### Testing

Unit test matrix covers success, error, and panic cases. Integration test verifies
goose-session-id header appears in rename task's LLM requests without timing dependencies.

Sets foundation for future trace context propagation tests.

#### Before: ApiClient cannot see the session ID on rename

```mermaid
sequenceDiagram
    participant Agent
    participant ReplyInternal
    participant NamingTask
    participant ApiClient
    participant LLM

    Agent->>ReplyInternal: reply(session_id=abc-123)

    rect rgb(255, 200, 200)
    Note over ReplyInternal,LLM: Before: rename task missing session header
    ReplyInternal->>NamingTask: tokio::spawn(without context)
    NamingTask->>ApiClient: generate_session_name()
    Note over ApiClient: current_session_id() → None
    ApiClient->>LLM: POST /chat (no goose-session-id)
    LLM-->>NamingTask: "Generated Name"
    end

    Note over NamingTask: Testing requires sleep-based timing
```

#### After: ApiClient can see the session ID on rename

```mermaid
sequenceDiagram
    participant Agent
    participant ReplyInternal
    participant NamingTask
    participant ApiClient
    participant LLM

    Agent->>ReplyInternal: reply(session_id=abc-123)

    rect rgb(191, 223, 255)
    Note over ReplyInternal,LLM: After: session ID injected in headers
    ReplyInternal->>NamingTask: tokio::spawn(with_session_id(abc-123))
    NamingTask->>ApiClient: generate_session_name()
    Note over ApiClient: current_session_id() → abc-123
    ApiClient->>LLM: POST /chat (goose-session-id: abc-123)
    LLM-->>NamingTask: "Generated Name"
    end

    Note over ReplyInternal: Process messages...
    ReplyInternal->>NamingTask: await handle
    NamingTask-->>ReplyInternal: complete
    Note over Agent: Task completion verified (deterministic)
```

### Related

Follows up https://github.com/block/goose/pull/5165

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -873,14 +873,17 @@ impl Agent {
         let reply_span = tracing::Span::current();
         self.reset_retry_attempts().await;
 
+        let working_dir = session.working_dir.clone();
         let provider = self.provider().await?;
         let session_id = session_config.id.clone();
-        let working_dir = session.working_dir.clone();
-        tokio::spawn(async move {
-            if let Err(e) = SessionManager::maybe_update_name(&session_id, provider).await {
-                warn!("Failed to generate session description: {}", e);
-            }
-        });
+        let naming_handle = tokio::spawn(crate::session_context::with_session_id(
+            Some(session_id.clone()),
+            async move {
+                if let Err(e) = SessionManager::maybe_update_name(&session_id, provider).await {
+                    warn!("Failed to generate session description: {}", e);
+                }
+            },
+        ));
 
         Ok(Box::pin(async_stream::try_stream! {
             let _ = reply_span.enter();
@@ -1220,6 +1223,8 @@ impl Agent {
 
                 tokio::task::yield_now().await;
             }
+
+            let _ = naming_handle.await;
         }))
     }
 
@@ -1528,7 +1533,13 @@ impl Agent {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::model::ModelConfig;
+    use crate::providers::base::ProviderUsage;
     use crate::recipe::Response;
+    use crate::session::session_manager::SessionType;
+    use crate::session::SessionManager;
+    use async_trait::async_trait;
+    use test_case::test_case;
 
     #[tokio::test]
     async fn test_add_final_output_tool() -> Result<()> {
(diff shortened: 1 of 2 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -1210,6 +1222,11 @@ impl Agent {
 
                 tokio::task::yield_now().await;
             }
+
+            // Wait for naming task to complete if it was spawned
+            if let Some(handle) = naming_handle {
+                let _ = handle.await;
```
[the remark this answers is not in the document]
codefromthecrypt: note to human: do we actually desire to panic on rename fail? maybe I should make a comment saying why this is intentionally not panicing? (because the rename is a low value task)

codefromthecrypt:
marking draft as I didn't realize we have copilot now, will undraft when done satisfying it.

crates/goose/src/agents/agent.rs
```diff
@@ -863,14 +863,25 @@ impl Agent {
         let reply_span = tracing::Span::current();
         self.reset_retry_attempts().await;
 
-        let provider = self.provider().await?;
+        // Only spawn naming task if user hasn't provided a name
         let session_id = session_config.id.clone();
         let working_dir = session.working_dir.clone();
```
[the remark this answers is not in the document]
codefromthecrypt: invalid

crates/goose/src/agents/agent.rs
```diff
@@ -863,14 +863,26 @@ impl Agent {
...
```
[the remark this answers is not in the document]
codefromthecrypt: invalid you just didn't look far enough

crates/goose/src/agents/agent.rs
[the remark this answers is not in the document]
codefromthecrypt: satisfied the bot by saying why

crates/goose/src/agents/agent.rs
```diff
@@ -1210,6 +1215,14 @@ impl Agent {
 
                 tokio::task::yield_now().await;
             }
+
+            // Wait for naming task to complete if it was spawned
+            // We don't propagate panics since session naming is a low-value background task
```
codefromthecrypt: I added this to satisfy the bot even though it isn't typical in this project to spam the user on things that don't matter. open to advice from a human on it

DOsinga: I disagree with the bot. these warning just go to the log, we won't be looking into it.

DOsinga (commented):
Thanks for doing this. I am not sure I understand the fix though. 

you check:
    let naming_handle = if !session.user_set_name {
    
but in the `maybe_update_name` we have the same check already:

```
if session.user_set_name {
            return Ok(());
        }
```

so in which conditions does this not work?

codefromthecrypt:
@DOsinga good point about the check being internal there. I think probably in the process of trying to make a black box test I may have goofed about this improving the opt-out of the llm call, where it just opts out of spawning a task which could outlive the calling function.

I had several things in progress since several weeks ago on a local branch, and got things mixed up.

So TL;DR: this doesn't obviate the expensive LLM call (already done) all it does is ensure the task doesn't escape the closure of the calling function, which you could argue is a helpful or not. If you want it to escape, it makes things a bit harder to do with replay tests. The fact that there are none yet, that I'm aware of, might reduce the impact on this point.

Over to you.

codefromthecrypt:
@DOsinga so, I chopped this off a pending otel propagation PR, and lost some signal in doing so. I redid this PR for the essentials which are not about whether rename happens or not.

1. make rename testable without sleeps (bind the task to the outer closure)
2. fix the missing session ID propagation when renaming the session (#5165 missed this)
3. backfill missing tests so that when we go back for otel, there's an example of how to reach this part.

crates/goose/src/agents/agent.rs
```diff
@@ -873,14 +873,16 @@ impl Agent {
...
         let provider = self.provider().await?;
         let session_id = session_config.id.clone();
-        let working_dir = session.working_dir.clone();
-        tokio::spawn(async move {
-            if let Err(e) = SessionManager::maybe_update_name(&session_id, provider).await {
-                warn!("Failed to generate session description: {}", e);
-            }
-        });
+        let naming_handle = tokio::spawn(
+            crate::session_context::with_session_id(Some(session_id.clone()), async move {
```
codefromthecrypt: without this, even though the session id is propagated manually, it is not available to api_client which looks for it in context, this is the missing piece and also a place for folks to look at when they get around to otel. Any async seams need a hard look or things like session or in the future trace propagation will break.

crates/goose/src/agents/agent.rs
```diff
@@ -1587,4 +1597,142 @@ mod tests {
...
+
+        fn get_model_config(&self) -> ModelConfig {
+            self.model_config.clone()
+        }
+    }
+
+    // Verifies session ID is visible in generate_session_name when user hasn't provided a name.
+    // When user provides name, maybe_update_name early-returns and session ID isn't captured.
+    #[test_case(NamingBehavior::Success, None, "Generated Name", true)]
+    #[test_case(NamingBehavior::Error, None, "initial", true)]
```
codefromthecrypt, line 1677: error and panic cases here are just to make sure our ignoring the same doesn't accidentally crash the critical code. we didn't have a problem so far, but it is easy to create one, so these cases help IMHO

codefromthecrypt:
opened https://github.com/block/goose/issues/5649 as especially as even trivial workflows need approval, it is annoying to find one missed then repeat the rebase drift loop again. I think we can make it easier to not fail

crates/goose/src/agents/agent.rs
```diff
@@ -873,14 +873,17 @@ impl Agent {
...
+        let working_dir = session.working_dir.clone();
         let provider = self.provider().await?;
         let session_id = session_config.id.clone();
-        let working_dir = session.working_dir.clone();
-        tokio::spawn(async move {
-            if let Err(e) = SessionManager::maybe_update_name(&session_id, provider).await {
-                warn!("Failed to generate session description: {}", e);
-            }
-        });
+        let naming_handle = tokio::spawn(crate::session_context::with_session_id(
```
codefromthecrypt, line 879: @alexhancock so this code and the test below makes sure the session rename task is also grouped by the session.id, which ack is not something people always care about, but if you think systemically there's probably no value in intentionally orphaning it. I noticed your ACP pr, if it top-levels the session ID stuff, will end up needing a similar line in spawn. https://github.com/block/goose/pull/5657

codefromthecrypt:
gonna save our time on this. Anyone who wants to harvest tests or approach here at the point it becomes relevant, do so. Meanwhile I think review and rebase duty is better spent elsewhere.

codefromthecrypt:
https://github.com/block/goose/pull/7052 will obviate this at least for tests, as testing was the tricky part as the request never finishes contained to any function for sure. Meanwhile, what we did was special case a separate HTTP handler in tests to work around this.

codefromthecrypt:
also the root issue for propagation of session id was already solved by doing it explicitly in https://github.com/block/goose/pull/6584

PR #5634, opened 2025-11-07
Scheduler recipe execution unification

tlongwell-block:
Use `run_complete_subagent_task`, which internally uses `AgentManager`. The scheduler no longer creates `Agent` instances directly.


tlongwell-block:
Closing until https://github.com/block/goose/pull/5571 is merged in

Issue #5527, opened 2025-11-02
Multiple file picker popups open when double-clicking the “Select File” button


shailedrathakur300:
When I click the “Select File” button once, it correctly opens a file picker popup.
However, if I double-click or click the button multiple times quickly, multiple popups open simultaneously instead of restricting to one.

This causes multiple overlapping file dialogs, making it confusing and hard to close them all.

**To Reproduce**

Go to the file upload section (e.g., /home/shailendra/ or similar directory view).
Click the “Select File” button once — file picker opens (expected).
Click the same button twice or rapidly multiple times.
Multiple file picker popups appear instead of just one.

**Expected behavior**

Only one file picker popup should appear at a time, even if the button is clicked multiple times rapidly

OS : Ubuntu 24.04 x86_64
Interface: UI (Desktop)
Extensions enabled: None
Provider & Model: Anthropic Claude Sonnet-4

<img width="815" height="93" alt="Image" src="https://github.com/user-attachments/assets/164ae996-9331-42a6-858b-c4f938a1e6a0" />
<img width="1544" height="886" alt="Image" src="https://github.com/user-attachments/assets/a955da91-a4f0-47f7-9e3d-cafa13bc3523" />

Lymah123:
Hi @angiejones, can I work on this?

DOsinga:
go for it

DOsinga:
should be fixed now!

PR #5640, opened 2025-11-08
Bug: prevent multiple file picker dialogs on double-click

Lymah123:
## Summary

Add fileDialogOpen guard flag to select-file-or-directory IPC handler to prevent
race condition where rapid double-clicks would spawn multiple overlapping
file picker dialogs. Uses try-finally to ensure flag is always reset.

Also add .cargo/ to .gitignore to exclude local cargo config.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Related Issues
Relates to #5527 
Discussion: LINK (if any)

Before:  
image

After:   
Screenshot (30)
Screenshot (31)

### Submitting a Recipe?

**Email**:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,4 +72,3 @@ result
 
 # Goose self-test artifacts
 gooseselftest/
-.cargo/
--- a/documentation/src/pages/community/data/community-content.json
+++ b/documentation/src/pages/community/data/community-content.json
@@ -298,19 +298,6 @@
         "hacktoberfest"
       ],
       "hacktoberfest": true
-    },
-    {
-      "title": "Managing goose Configurations Across Multiple Projects",
-      "author": {
-        "name": "Halimah",
-        "handle": "lymah123"
-      },
-      "type": "blog",
-      "url": "https://dev.to/lymah/managing-goose-configurations-across-multiple-projects-58i3",
-      "thumbnail": "https://i.imgur.com/VvM8sMR.png",
-      "submittedDate": "2025-10-30",
-      "tags": ["hacktoberfest"],
-      "hacktoberfest": true
     }
   ]
 }
--- a/ui/desktop/src/components/ChatInput.tsx
+++ b/ui/desktop/src/components/ChatInput.tsx
@@ -122,6 +122,7 @@ export default function ChatInput({
   const [displayValue, setDisplayValue] = useState(initialValue); // For immediate visual feedback
   const [isFocused, setIsFocused] = useState(false);
   const [pastedImages, setPastedImages] = useState<PastedImage[]>([]);
+  const [isFileDialogOpen, setIsFileDialogOpen] = useState(false);
 
   // Derived state - chatState != Idle means we're in some form of loading state
   const isLoading = chatState !== ChatState.Idle;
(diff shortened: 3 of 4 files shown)


Lymah123:
Hi @DOsinga! I will be waiting for your review.

Lymah123:
Hi!
Just to confirm: would you prefer we handle this by disabling the select button in `ChatInput.tsx` while the `dialog` is open, instead of managing it in the Electron main process? Want to make sure I'm taking the right approach, thanks! @DOsinga

DOsinga:
> Hi! Just to confirm: would you prefer we handle this by disabling the select button in `ChatInput.tsx` while the `dialog` is open, instead of managing it in the Electron main process? Want to make sure I'm taking the right approach, thanks! @DOsinga

yes, I think that would work better - thanks

Lymah123:
> > Hi! Just to confirm: would you prefer we handle this by disabling the select button in `ChatInput.tsx` while the `dialog` is open, instead of managing it in the Electron main process? Want to make sure I'm taking the right approach, thanks! @DOsinga
> 
> yes, I think that would work better - thanks

Alright. I will work on the requested changes. Thanks

documentation/src/pages/community/data/community-content.json
```diff
@@ -298,6 +298,19 @@
         "hacktoberfest"
       ],
       "hacktoberfest": true
+    },
```
alexhancock: These changes don't seem relevant to the intended scope of the PR. Are they left in unintentionally?

Lymah123: I will remove those files once I am done making the requested changes. They were from my past PRs.

documentation/src/pages/community/data/community-content.json
Abhijay007, line 1: This file shouldn't be deleted; just your changes need to be reverted

Lymah123: Fixed. Thanks

ui/desktop/src/components/ChatInput.tsx
```diff
@@ -122,6 +122,7 @@ export default function ChatInput({
   const [displayValue, setDisplayValue] = useState(initialValue); // For immediate visual feedback
   const [isFocused, setIsFocused] = useState(false);
   const [pastedImages, setPastedImages] = useState<PastedImage[]>([]);
+  const [isFileDialogOpen, setIsFileDialogOpen] = useState(false); // NEW: Track file dialog state
```
blackgirlbytes: just removing extra comments that we may not need
```suggestion
  const [isFileDialogOpen, setIsFileDialogOpen] = useState(false); 
```

ui/desktop/src/components/ChatInput.tsx
```diff
@@ -1454,19 +1463,23 @@ export default function ChatInput({
         <DirSwitcher className="mr-0" />
         <div className="w-px h-4 bg-border-default mx-2" />
 
+        {/* Attach button now disables when dialog is open */}
```
blackgirlbytes: i think we may not need this comment either
```suggestion
```

.gitignore
```diff
@@ -72,3 +72,4 @@ result
 
 # Goose self-test artifacts
 gooseselftest/
+.cargo/
```
Abhijay007: You can remove this change

ui/desktop/src/components/ChatInput.tsx
```diff
@@ -1453,20 +1462,23 @@ export default function ChatInput({
...
               size="sm"
-              className="flex items-center justify-center text-text-default/70 hover:text-text-default text-xs cursor-pointer transition-colors"
+              className="flex items-center justify-center text-text-default/70 hover:text-text-default text-xs cursor-pointer transition-colors disabled:opacity-50 disabled:cursor-not-allowed"
             >
               <Attach className="w-4 h-4" />
             </Button>
           </TooltipTrigger>
-          <TooltipContent>Attach file or directory</TooltipContent>
+          <TooltipContent>
+            {isFileDialogOpen ? 'Dialog open...' : 'Attach file or directory'}
```
Abhijay007, line 1480: This here isn’t working, when a button is disabled, radix UI tooltips don’t show on hover by default, so you might need to add an alternate fix for showing dialog is open or maybe just disable grey button is fine or you can show a disabled point there.  

Also, when I’m in different tabs, I’m not able to open multiple dialogs, but that might be because we’re only fixing it for the chat input, so maybe we can ignore that for now

Lymah123: Ok, I will look into it.

DOsinga:
I merged the other one since that didn't have a conflict, but I believe the approach is the same, so thanks!

PR #5643, opened 2025-11-08
Show errors on failure

DOsinga:
## Summary

not sure how this was supposed to work. If we failed to start goosed, we'd just throw an Error which the user never sees. also waiting 120 seconds on the error, I don't think users even see that. it also improves `goose info` so it doesn't crash if the config can't be created and prints for the files it has info on if they exist and if we could create them. also removes the unused and confusing scheduler settings

this improves this by showing the stderr on failure, early exit on error: detection, shortening the wait time and setting logging to console to non-ansi. also changes the context to actually say what went wrong


crates/goose-server/src/logging.rs
```diff
@@ -34,35 +31,27 @@ pub fn setup_logging(name: Option<&str>) -> Result<()> {
         .with_ansi(false)
         .with_file(true);
 
-    // Create console logging layer for development - INFO and above only
+    let base_env_filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| {
+        EnvFilter::new("")
+            .add_directive("mcp_client=info".parse().unwrap())
```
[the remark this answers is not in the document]
alexhancock: is the thought just to make it consistent with goose-server?

ui/desktop/src/goosed.ts
```diff
@@ -183,7 +174,14 @@ export const startGoosed = async (
   });
 
   goosedProcess.stderr?.on('data', (data: Buffer) => {
-    log.error(`goosed stderr for port ${port} and dir ${dir}: ${data.toString()}`);
+    const lines = data
+      .toString()
+      .split('\n')
+      .filter((l) => l.trim());
+    lines.forEach((line) => {
+      log.error(`goosed stderr for port ${port} and dir ${dir}: ${line}`);
```
alexhancock, line 182: this would repeat `goosed stderr for port ${port} and dir ${dir}:` for every line in each data chunk

maybe something like:

```
const lines = data
  .toString()
  .split('\n')
  .filter((l) => l.trim());
stderrLines.push(...lines);
log.error(`goosed stderr for port ${port} and dir ${dir}: ${lines.join("\n");}`);

```

?

ui/desktop/src/main.ts
```diff
@@ -1167,22 +1145,6 @@ ipcMain.handle('get-goosed-host-port', async (event) => {
...
 });
 
-ipcMain.handle('set-scheduling-engine', async (_event, engine: string) => {
-  try {
-    const settings = loadSettings();
-    settings.schedulingEngine = engine as SchedulingEngine;
-    saveSettings(settings);
-
-    // Update the environment variable immediately
-    updateSchedulingEngineEnvironment(settings.schedulingEngine);
```
alexhancock, line 1177: we should remove the definition of this too

ui/desktop/src/components/settings/scheduler/SchedulerSection.tsx
```diff
@@ -1,113 +0,0 @@
-import { useState, useEffect } from 'react';
-import { SchedulingEngine, Settings } from '../../../utils/settings';
```
alexhancock, line 2: remove `SchedulingEngine` from `utils/settings` too right?

ui/desktop/src/main.ts
```diff
@@ -594,23 +574,21 @@ const createChat = async (
   );
   goosedClients.set(mainWindow.id, goosedClient);
 
-  console.log('[Main] Waiting for backend server to be ready...');
-  const serverReady = await checkServerStatus(goosedClient);
+  const serverReady = await checkServerStatus(goosedClient, errorLog);
   if (!serverReady) {
-    throw new Error('Backend server failed to start in time');
+    dialog.showMessageBoxSync({
```
alexhancock, line 579: maybe https://www.electronjs.org/docs/latest/api/dialog#dialogshowerrorboxtitle-content as it's meant for errors?

alexhancock (approved):

ui/desktop/src/main.ts
DOsinga: I tried that, it doesn't respect the newlines and so made it harder to read

ui/desktop/src/goosed.ts
DOsinga: well, yes, that is the existing behavior

ui/desktop/src/main.ts
DOsinga: good point. makes me wonder why our linter didn't catch the unused stuff

ui/desktop/src/components/settings/scheduler/SchedulerSection.tsx
DOsinga: done

ui/desktop/src/main.ts
alexhancock: got it

ui/desktop/src/goosed.ts
alexhancock: The existing is:

```
  goosedProcess.stderr?.on('data', (data: Buffer) => {
    log.error(`goosed stderr for port ${port} and dir ${dir}: ${data.toString()}`);
    ...
```

so `"goosed stderr for port ${port} and dir ${dir}:"` shows up in the logs N times where N is the number of `data` chunks. my understanding of the change is it would now show up L times where L is the number of lines in each `data` chunk, which could be tens/hundreds of lines per chunk right?

I could also be missing something though

crates/goose-server/src/logging.rs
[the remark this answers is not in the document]
DOsinga: yeah. I think we could go further even. mcp is rather chatty and any time I want to look into the logs, there's a huge log statement saying something unparseable (though it has gotten better) I think. but in general we have way too many debug's in there. we should do something about that, but not showing them is a start

ui/desktop/src/goosed.ts
DOsinga: technically, true, but in practice L & N are the same

we can still remove it, maybe logging that it came from a certain directory was useful at some point? I don't know that anybody looks at these though

Issue #5293, opened 2025-10-21
`goose recipe list` can return duplicated entries


benoitf:
**Describe the bug**

duplicated entries using `goose recipe list` depending on env var or directories

**To Reproduce**
1. provide as env var the default recipe storage

`GOOSE_RECIPE_PATH="$HOME/.config/goose/recipes" goose recipe list --format json`

```
GOOSE_RECIPE_PATH="$HOME/.config/goose/recipes" goose recipe list --format json | jq .                                                                                               
[
  {
    "name": "my-task",
    "source": "Local",
    "path": "$HOME/.config/goose/recipes/my-task.yaml",
    "title": "my-task,
    "description": "my-task"
  },
  {
    "name": "my-task",
    "source": "Local",
    "path": $HOME/.config/goose/recipes/my-task.yaml",
    "title": "my-task,
    "description": "my-task"
  },
]
```

or  if you go to the `$HOME/.config/goose` folder and try the env variable or not, you may have up to 3 duplicates. As it'll list from current folder, default folder and env variable folder.

IMHO It should filter out duplicated locations

**Screenshots**

**Please provide following information:**
 - **OS & Arch:**  macOS latest
 - **Interface:** [CLI]
 - **Version:** 1.11.1
 - **Extensions enabled:** none
 - **Provider & Model:** none

**Additional context**
was working in v1.9.3, fails in v1.10 and v1.11

DOsinga:
yup that is true. doesn't strike me as terrible, if you have the same place to search for multiple times, but still

benoitf:
It's a regression since v1.9.3

zanesq:
@Abhijay007 checking if still working on this?

Abhijay007:
> [@Abhijay007](https://github.com/Abhijay007) checking if still working on this?

Yeah @zanesq , I am ooo for a few days, will get back and fix this soon

DOsinga:
how's it going @Abhijay007 ? can I help?

Abhijay007:
> how's it going [@Abhijay007](https://github.com/Abhijay007) ? can I help?

Hi @DOsinga, I was out for some work and just came back a few days ago. I was working on another PR and resolved it yesterday, apologies for the delay on this one 😅 . I’ll resolve it soon :)

PR #5645, opened 2025-11-09
fix: goose recipe list can return duplicated entries

Abhijay007:
Closes: #5293 

## Pull Request Description

This PR addresses an issue where running goose recipe list resulted in duplicate recipe entries when the `GOOSE_RECIPE_PATH ` environment variable was set to the same directory as the default global recipe directory (`~/.config/goose/recipes`).

The root cause was that the duplication occurred because the `local_recipe_dirs()` function aggregates recipe directories from multiple sources: the current working directory, directories specified via `GOOSE_RECIPE_PATH`, the global recipe directory, and local project-specific recipe directories. When `GOOSE_RECIPE_PATH` pointed to the default global recipe directory, the same path was included more than once. Since these paths were not normalized before being processed, the application scanned the same directory multiple times, resulting in duplicate entries when listing recipes.

### Changes Made

To resolve this issue, deduplication logic was added to the local_recipe_dirs() function. The solution works as follows:

- Canonicalize paths: Convert all collected paths into their absolute canonical forms (resolving symlinks and normalization like . and ..).
- Sort paths: Ensure paths are sorted so duplicates are adjacent.
- Remove duplicates: Use dedup() to eliminate repeated entries.

This ensures that each recipe directory is processed exactly once.

NOTE: this could be solved with HashSet too, but I saw that sort and dedup is used in `lead_worked.rs` so I used a similar approach

### Type of Change
- [x] Bug fix

### AI Assistance
- [x] This PR was created or reviewed with AI assistance - used Goose for approach validation 

### Testing

Tested with the same command shared in the issue 

```
GOOSE_RECIPE_PATH="$HOME/.config/goose/recipes" goose recipe list --format json | jq   
```

### DIFF
--- a/crates/goose/src/recipe/local_recipes.rs
+++ b/crates/goose/src/recipe/local_recipes.rs
@@ -29,7 +29,13 @@ fn local_recipe_dirs() -> Vec<PathBuf> {
     local_dirs.push(get_recipe_library_dir(true));
     local_dirs.push(get_recipe_library_dir(false));
 
-    local_dirs
+    let mut dirs: Vec<PathBuf> = local_dirs
+        .into_iter()
+        .map(|dir| dir.canonicalize().unwrap_or(dir))
+        .collect();
+    dirs.sort();
+    dirs.dedup();
+    dirs
 }
 
 pub fn load_local_recipe_file(recipe_name: &str) -> Result<RecipeFile> {


Abhijay007:
> I suppose this can still create dupes through github, but hey, this is a great improvement

okay I will look into this, thanks for the feedback :)

DOsinga:
nah, don't worry about it. that's just nitpicking

PR #5646, opened 2025-11-09
custom instructions for copilot reviews

angiejones:
### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [X] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -22,17 +22,17 @@
 - Resource leaks (files, connections, memory)
 - Off-by-one errors or boundary conditions
 - Incorrect error propagation (using `unwrap()` inappropriately)
+- Optional types that don't need to be optional
+- Booleans that should default to false but are set as optional
+- Error context that doesn't add useful information (e.g., `.context("Failed to do X")` when error already says it failed)
+- Overly defensive code that adds unnecessary checks
+- Unnecessary comments that just restate what the code already shows (remove them)
 
 ### Architecture & Patterns
 - Code that violates existing patterns in the codebase
 - Missing error handling (should use `anyhow::Result`)
 - Async/await misuse or blocking operations in async contexts
 - Improper trait implementations
-
-### Testing Gaps
-- New features without tests
-- Changed logic without corresponding test updates
-- Missing error case coverage
 
 ## Skip These (Low Value)
 
@@ -42,6 +42,8 @@
 - Obvious code that doesn't need explanation
 - Suggestions to add comments for self-documenting code
 - Refactoring suggestions unless there's a clear bug or maintainability issue
+- Listing multiple potential issues in one comment (choose the single most critical issue)
+- Suggestions to add more logging (the codebase needs less logging, not more)
 
 ## Response Format
 
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,70 @@
+# GitHub Copilot Code Review Instructions
+
+## Review Philosophy
+- Only comment when you have HIGH CONFIDENCE (>80%) that an issue exists
+- Be concise: one sentence per comment when possible
+- Focus on actionable feedback, not observations
+- Skip comments on style that clippy/rustfmt will catch
+
+## Priority Areas (Review These)
+
+### Security & Safety
+- Unsafe code blocks without justification
+- Command injection risks (shell commands, user input)
+- Path traversal vulnerabilities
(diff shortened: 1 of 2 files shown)


michaelneale (approved):
looks good @angiejones where did you get those guidelines from? 

I think over time if we see it flag things incorrectly which are due to specific quirks of how goose works vs other style of apps, we can easily add the tips to here. 

I likeit.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,70 @@
...
+- Off-by-one errors or boundary conditions
+- Incorrect error propagation (using `unwrap()` inappropriately)
+
+### Architecture & Patterns
+- Code that violates existing patterns in the codebase
+- Missing error handling (should use `anyhow::Result`)
+- Async/await misuse or blocking operations in async contexts
+- Improper trait implementations
+
+### Testing Gaps
```
DOsinga: can we drop these? our current code base has way more tests with a negative value than tests that are missing; I would take bug fixes that don't have a matching test that make future regressions impossible

DOsinga (approved):
thanks for doing this, we do need something here. I wonder though if we should just all go through the reviews we got from copilot and collect what was useful and what was not.

DOsinga:
fwiw, the three comments I got on this:

https://github.com/block/goose/pull/5643

were all spot on

tlongwell-block:
This is a great idea. Something to keep in mind, copilot [does read](https://github.blog/changelog/2025-08-28-copilot-coding-agent-now-supports-agents-md-custom-instructions/) AGENTS.md in the root and in subdirectories. Though hard to tell if that is just the "coding" agent, or copilot in general, including the review bot.

(Maybe goose should check subdirs too when it is working. That would be a nice feature that would allow per-crate instructions, @DOsinga?)

So, we could probably reduce duplication a bit between the instructions file here and agents.

DOsinga:
I actually have some notes on what goes often wrong in our code:

Do not make things optional that don't need to be. The compiler is annoying but is your friend
Remove useless LLM comments - at best they take up space, at worst they will be outdated soon and will confuse
Beware for Goose suggesting to check 3 different things. It is probably only one
Booleans should default to false and not be optional
Beware of the context manager that doesn't need context
Clean up your logging
Avoid overly defensive code

tlongwell-block (approved):

tlongwell-block:
> I actually have some notes on what goes often wrong in our code:
> 
> Do not make things optional that don't need to be. The compiler is annoying but is your friend Remove useless LLM comments - at best they take up space, at worst they will be outdated soon and will confuse Beware for Goose suggesting to check 3 different things. It is probably only one Booleans should default to false and not be optional Beware of the context manager that doesn't need context Clean up your logging Avoid overly defensive code

I would probably put this in AGENTS.md so all AI working on the repo will follow these guidelines?

Looks like copilot review does respect agents, too https://docs.github.com/en/copilot/concepts/agents/code-review#providing-instructions-for-copilot-code-reviews

angiejones:
> > I actually have some notes on what goes often wrong in our code:
> > Do not make things optional that don't need to be. The compiler is annoying but is your friend Remove useless LLM comments - at best they take up space, at worst they will be outdated soon and will confuse Beware for Goose suggesting to check 3 different things. It is probably only one Booleans should default to false and not be optional Beware of the context manager that doesn't need context Clean up your logging Avoid overly defensive code
> 
> I would probably put this in AGENTS.md so all AI working on the repo will follow these guidelines?
> 
> Looks like copilot review does respect agents, too https://docs.github.com/en/copilot/concepts/agents/code-review#providing-instructions-for-copilot-code-reviews

you want to move everything to AGENTS.md and delete the copilot custom instructions file, or just add these new entries to AGENTS.md?

DOsinga:
it's a good point that we include AGENTS.md anyway, but I would still think repeating the important bits as explicit code review instructions (pay attention to optional things that shouldn't) wouldn't hurt

angiejones:
> it's a good point that we include AGENTS.md anyway, but I would still think repeating the important bits as explicit code review instructions (pay attention to optional things that shouldn't) wouldn't hurt

added coding guidelines to AGENTS.md and review guidelines to the copilot custom instructions
https://github.com/block/goose/pull/5646/commits/2ccc3a795195364f0f7fc42ebcf2939de7f77fd0

PR #5648, opened 2025-11-10
Add tool calling emulation for OpenAI compatible providers 

kirill670:
## Summary

Added tool calling emulation for OpenAI compatible providers.

### Type of Change

- [ x ] Feature
- [ x ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

- [ x ] This PR was created or reviewed with AI assistance

### Testing
Unit tests

### Related issues 
None

Before:  
None yet
After:   
None yet
### Submitting a Recipe?

**Email**:

### DIFF
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -55,6 +55,7 @@ pub struct OpenAiProvider {
     model: ModelConfig,
     custom_headers: Option<HashMap<String, String>>,
     supports_streaming: bool,
+    supports_tool_calling: bool,
     name: String,
 }
 
@@ -78,6 +79,11 @@ impl OpenAiProvider {
             .ok()
             .map(parse_custom_headers);
         let timeout_secs: u64 = config.get_param("OPENAI_TIMEOUT").unwrap_or(600);
+        let supports_tool_calling: bool = config
+            .get_param("OPENAI_SUPPORTS_TOOL_CALLING")
+            .unwrap_or_else(|_| "true".to_string())
+            .parse()
+            .unwrap_or(true);
 
         let auth = AuthMethod::BearerToken(api_key);
         let mut api_client =
@@ -109,6 +115,7 @@ impl OpenAiProvider {
             model,
             custom_headers,
             supports_streaming: true,
+            supports_tool_calling,
             name: Self::metadata().name,
         })
     }
@@ -123,6 +130,7 @@ impl OpenAiProvider {
             model,
             custom_headers: None,
             supports_streaming: true,
+            supports_tool_calling: true,
             name: Self::metadata().name,
         }
     }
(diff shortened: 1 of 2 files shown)


michaelneale:
thanks @kirill670 do you mind doing the DCO signing thing? https://github.com/block/goose/pull/5648/checks?check_run_id=54928685902

(also a quick cargo fmt and ./script/clippy-lint.sh)

crates/goose/src/providers/openai.rs
```diff
@@ -238,31 +248,67 @@ impl Provider for OpenAiProvider {
...
+                    Usage::default()
+                });
+
+            let model = get_model(&json_response);
+            log.write(&json_response, Some(&usage))?;
+
+            let interpreter = OllamaInterpreter::new()?;
+            message = augment_message_with_tool_calls(&interpreter, message, tools).await?;
+
+            Ok((message, ProviderUsage::new(model, usage)))
```
DOsinga, line 310: the if and else branch are identical. why not leave it all as is, but then add at the end, this last bit?

```
if !self.supports_tool_calling && !tools.is_empty() {
        use super::toolshim::{augment_message_with_tool_calls, OllamaInterpreter};
        let interpreter = OllamaInterpreter::new()?;
        message = augment_message_with_tool_calls(&interpreter, message, tools).await?;
    }
```

and then instead of not streaming below when you need tool shimming, you can do the same below? maybe also lift out the use statement?

crates/goose/src/providers/openai.rs
```diff
@@ -78,6 +79,11 @@ impl OpenAiProvider {
             .ok()
             .map(parse_custom_headers);
         let timeout_secs: u64 = config.get_param("OPENAI_TIMEOUT").unwrap_or(600);
+        let supports_tool_calling: bool = config
+            .get_param("OPENAI_SUPPORTS_TOOL_CALLING")
+            .unwrap_or_else(|_| "true".to_string())
+            .parse()
+            .unwrap_or(true);
```
DOsinga, line 86: I think we can do without this; if you want to have an openai provider that does tool emulation, just use the declarative provider

michaelneale:
I wonder if we get it working on openai, then we make sure ollama really is just using openai compatibility (not sure how close it is, but worth thinking about so) - but ideally want it on both, but they should be getting closer together.

I am also curious how well people are finding this emulation working

DOsinga:
it is already pretty close indeed; our ollama provider doesn't have an API key and why should it, but some people run their own ollama version on a remote server with an API key and I've told them to just use the OpenAi compatible custom provider and that seems to work

DOsinga:
also seems to be running into some checks - can you do the DCO thing also?

michaelneale:
closing as a bit old, feel free to re-open.

Issue #5626, opened 2025-11-07
In an ACP session via Zed, goose cannot see provided @ context


pjv:
When using goose via ACP through Zed, there is an opportunity to provide context to the model using `@` which looks like this:

<img width="456" height="307" alt="Image" src="https://github.com/user-attachments/assets/d553697a-121c-46d8-9a55-28c3415f793b" /><br />

But when you do, goose apparently cannot see the provided context:

<img width="635" height="129" alt="Image" src="https://github.com/user-attachments/assets/e8509a23-e194-4795-bb6a-28e340939ffc" /><br />

opencode's ACP implementation is able to see the provided context:

<img width="637" height="364" alt="Image" src="https://github.com/user-attachments/assets/299e99c1-4b96-4603-a793-7f39c268c9fa" />

---

**Please provide the following information**
- **OS & Arch:** macOS 15.7.1 x86
- **Interface:** ACP
- **Version:** 1.13.0
- **Extensions enabled:** developer, tutorial
- **Provider & Model:** ollama - qwen3:4b-instruct-2507-q8_0

alexhancock:
Hi @pjv I just merged a change which should let you export diagnostics from the session via this https://block.github.io/goose/docs/troubleshooting/diagnostics-and-reporting/#generating-diagnostics

We will need to wait until it's released if you want to use a released version of goose, but if you're up to run goose from source on main you can clone the repo and add this to your zed config (Option + Command/Windows Button + ,)

```
{
  "theme": "One Dark",
  "agent_servers": {
    "goose": {
      "command": "cargo",
      "args": [
        "+nightly",
        "-Z",
        "unstable-options",
        "-C",
        "/WHEREVER YOU PUT/goose",
        "run",
        "--bin",
        "goose",
        "acp"
      ],
      "env": {}
    }
  }
}
```

I'd like see the message logs from a time this happens and a diagnostics package will give us that

pjv:
hi @alexhancock sorry, yesterday my starlink self-borked doing a firmware update so i was functionally wounded. Back to semi-functionality today. Goose has a release of version 1.14.0 in the mean time which I was hoping includes your new patch. I've updated to that version but when i create an ACP session and recapitulate my actions from the OP, i do not see that ACP session in `goose session list` so I guess not.

unfortunately cloning the goose repo over my pokey 5G temporary starlink replacement would take hours so i think i will have to wait until a release that includes your patch.

alexhancock:
[...]

PR #5657, opened 2025-11-10
feat: SessionManager integration for acp sessions

alexhancock:
SessionManager integration for ACP based sessions

1. Shares the session id between goose sessions and ACP sessions
2. Names sessions appropriately
3. Writes all messages

This will allow us to see ACP sessions in session history, collect diagnostics for them by session id, etc.

Example showing persistence of a session from an ACP client:

Screenshot 2025-11-10 at 3 11 25 PM

(As a followup, we can now implement `load_session` for ACP from goose's storage as well, but a bit more complexity to that so leaving it in a sketch commit for now https://github.com/block/goose/commit/f46f804d4a06e3d9e5a4f480617d150046e2758e)

Using the diagnostics from a session we should be able to better diagnose https://github.com/block/goose/issues/5626

### DIFF
--- a/crates/goose-cli/src/commands/acp.rs
+++ b/crates/goose-cli/src/commands/acp.rs
@@ -605,8 +605,12 @@ impl acp::Agent for GooseAcpAgent {
     ) -> Result<acp::NewSessionResponse, acp::Error> {
         info!("ACP: Received new session request {:?}", args);
 
-        // Generate a unique session ID
-        let session_id = uuid::Uuid::new_v4().to_string();
+        let goose_session = SessionManager::create_session(
+            std::env::current_dir().unwrap_or_default(),
+            "ACP Session".to_string(), // just an initial name - may be replaced by maybe_update_name
+            SessionType::User,
+        )
+        .await?;
 
         let session = GooseAcpSession {
             messages: Conversation::new_unvalidated(Vec::new()),
@@ -615,15 +619,14 @@ impl acp::Agent for GooseAcpAgent {
             cancel_token: None,
         };
 
-        // Store the session
         let mut sessions = self.sessions.lock().await;
-        sessions.insert(session_id.clone(), session);
+        sessions.insert(goose_session.id.clone(), session);
 
-        info!("Created new session with ID: {}", session_id);
+        info!("Created new ACP/goose session {}", goose_session.id);
 
         Ok(acp::NewSessionResponse {
-            session_id: acp::SessionId(session_id.into()),
-            modes: None, // TODO: Implement session modes if needed
+            session_id: acp::SessionId(goose_session.id.into()),
+            modes: None,
             meta: None,
         })
     }
@@ -649,15 +652,9 @@ impl acp::Agent for GooseAcpAgent {
     }
 
     async fn prompt(&self, args: acp::PromptRequest) -> Result<acp::PromptResponse, acp::Error> {
-        info!("ACP: Received prompt request {:?}", args);
-
-        // Get the session
         let session_id = args.session_id.0.to_string();
-
-        // Create and store cancellation token for this prompt
         let cancel_token = CancellationToken::new();
 
-        // Store the cancel token in the session BEFORE starting the prompt
         {
             let mut sessions = self.sessions.lock().await;
             let session = sessions
@@ -668,21 +665,13 @@ impl acp::Agent for GooseAcpAgent {
 
         let user_message = self.convert_acp_prompt_to_message(args.prompt);
 
-        let session = SessionManager::create_session(
-            std::env::current_dir().unwrap_or_default(),
-            "ACP Session".to_string(),
-            SessionType::Hidden,
-        )
-        .await?;
-
         let session_config = SessionConfig {
-            id: session.id.clone(),
+            id: session_id.clone(),
             schedule_id: None,
             max_turns: None,
             retry_config: None,
         };
 
-        // Get agent's reply through the Goose agent
         let mut stream = self
             .agent
             .reply(user_message, session_config, Some(cancel_token.clone()))
@@ -694,12 +683,9 @@ impl acp::Agent for GooseAcpAgent {
 
         use futures::StreamExt;
 
-        // Track if we were cancelled
         let mut was_cancelled = false;
 
-        // Process the agent's response stream
         while let Some(event) = stream.next().await {
-            // Check if we've been cancelled
             if cancel_token.is_cancelled() {
                 was_cancelled = true;
                 break;
@@ -707,24 +693,19 @@ impl acp::Agent for GooseAcpAgent {
 
             match event {
                 Ok(goose::agents::AgentEvent::Message(message)) => {
-                    // Re-acquire the lock to add message to conversation
                     let mut sessions = self.sessions.lock().await;
                     let session = sessions
                         .get_mut(&session_id)
                         .ok_or_else(acp::Error::invalid_params)?;
 
-                    // Add to conversation
                     session.messages.push(message.clone());
 
-                    // Process message content, including tool calls
                     for content_item in &message.content {
                         self.handle_message_content(content_item, &args.session_id, session)
                             .await?;
                     }
                 }
-                Ok(_) => {
-                    // Ignore other events for now
-                }
+                Ok(_) => {}
                 Err(e) => {
                     error!("Error in agent response stream: {}", e);
                     return Err(acp::Error::internal_error());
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/commands/acp.rs
```diff
@@ -605,8 +605,12 @@ impl acp::Agent for GooseAcpAgent {
     ) -> Result<acp::NewSessionResponse, acp::Error> {
         info!("ACP: Received new session request {:?}", args);
 
-        // Generate a unique session ID
-        let session_id = uuid::Uuid::new_v4().to_string();
+        let goose_session = SessionManager::create_session(
+            std::env::current_dir().unwrap_or_default(),
+            "ACP Session".to_string(), // just an initial name - may be replaced by maybe_update_name
```
DOsinga, line 610: hmmm?

crates/goose-cli/src/commands/acp.rs
```diff
@@ -667,22 +666,28 @@ impl acp::Agent for GooseAcpAgent {
...
+            acp::Error::internal_error()
+        })?;
+        let session_id_for_naming = session_id.clone();
+        tokio::spawn(async move {
+            if let Err(e) =
+                SessionManager::maybe_update_name(&session_id_for_naming, provider).await
+            {
+                warn!("Failed to generate session name: {}", e);
+            }
+        });
```
DOsinga: this feels like code copied from agent.reply which we are calling below

crates/goose-cli/src/commands/acp.rs
```diff
@@ -694,45 +699,40 @@ impl acp::Agent for GooseAcpAgent {
...
                 was_cancelled = true;
                 break;
             }
 
             match event {
                 Ok(goose::agents::AgentEvent::Message(message)) => {
-                    // Re-acquire the lock to add message to conversation
+                    if let Err(e) = SessionManager::add_message(&session_id, &message).await {
+                        error!("Failed to write message to session: {}", e);
+                    }
```
DOsinga: I don't think you want to do this. agent already keeps the conversation up to date.

DOsinga (approved):
seems like some of this code is copied from agent.reply but we're also calling agent.reply, so we should be good without it.

there's also `async fn load_session(` in the code that is currently implemented that we could implement now.

we should also decide whether we want to give these sessions user visibility - before the sessions were created as hidden, now they are not

crates/goose-cli/src/commands/acp.rs
michaelneale: would goose automatically update this after a few turns (via the LLM session naming)? If so, can probably leave it out

michaelneale:
seems like a good step towards it - does this mean that on acp client side they duplicate session data or is this just makes acp aware that there is a session?

crates/goose-cli/src/commands/acp.rs
codefromthecrypt: I'm not sure it is really helpful to intentionally give no useful initial name, right? there's a race condition and the first inference request will have a worse first name right? If this is an issue of it being too verbose to set the initial name, maybe we can find a better way to rejig it so it isn't so in your face?

crates/goose-cli/src/commands/acp.rs
```diff
@@ -667,22 +666,28 @@ impl acp::Agent for GooseAcpAgent {
...
             retry_config: None,
         };
 
-        // Get agent's reply through the Goose agent
+        let provider = self.agent.provider().await.map_err(|e| {
+            error!("Failed to get provider: {}", e);
+            acp::Error::internal_error()
+        })?;
+        let session_id_for_naming = session_id.clone();
+        tokio::spawn(async move {
```
codefromthecrypt: made a comment here about this spawn which I think if we had a test we would find it loses the session ID context which means outbound HTTP requests won't have the session.id header you just pinned https://github.com/block/goose/pull/5624/files#r2513696604

crates/goose-cli/src/commands/acp.rs
DOsinga: goose should name this automatically yeah. but also depending on what we decide on visibility here, people won't be seeing this name I think (unless I miss something in the ACP protocol)

codefromthecrypt:
@alexhancock feel free to punt me somewhere (acp discussion) and also take my opinions as a grain of salt..

TL;DR; maybe we need to know who should own the session name/title and meanwhile there seems to be an opinion that how session names are done in goose is already good enough(?).

Below for longer thoughts as I have looked at this perhaps too much.

<details>

---

### ACP session ID vs name/title and who should own it vs who does now

So the ACP session ID is important and I assume we want to pair these together, so that there is coherence and correlation between the editor and the agent. E.g. [session setup](https://agentclientprotocol.com/protocol/session-setup#session-id)

What is less clear to me, is who the ultimate owner of the session name (or the [title in ACP](https://agentclientprotocol.com/rfds/session-list))

> title - Optional human-readable title (could be auto-generated from first prompt)

I feel it would be very likely this could become a protocol concern as the owner of the editor is likely to allow the user to change the title, right? In the same vane, it could be that they want the editor to control the prompt (instead of using goose's prompt string). These are guesses. I added a comment to a [thread on ACP](https://github.com/orgs/agentclientprotocol/discussions/60#discussioncomment-14936950) about it


### Goose session naming strategy opinions in case anyone cares

Goose has session name/title approach which is largely like others that preceded our code for it, but there are some differences and maybe the rationale could be kicked around if anything should result in a change or not.

Since we are talking about ACP and ACP is from Zed, let's look at theirs right..

Zed's ACP thread is named *independent* of this goose naming, so it s a fair thing to use as an approach comparison.

What I mean is when Zed is *not* using ACP, it summarizes with a prompt like this:
```
Generate a concise 3-7 word title for this conversation, omitting punctuation.
Go straight to the title, without any preamble and prefix like `Here's a concise suggestion:...` or `Title:`.
If the conversation is about a specific subject, include it in the title.
Be descriptive. DO NOT speak in the first person.
```

Some prompts are a bit different than ours, use more or less messages, and/or have token limits.

Then, there's the model that is used for generation. The last open source version of sourcegraph/cody and current open source projects like Zed and OpenCode have fast or small choices used for the summarizer. This is not to say goose should change, others just use whatever the default model is, but having found junk in my session names and also folks being concerned about where/when and how to do our session name task.. maybe more research could help as there are others of course with the same problem.

There are of course other tools who intentionally choose to not generate names with LLMs instead choosing first line, or just letting people click "edit" on the title and done.

If there is interest in optimizing this, I would be glad to help but if so would want an eagerly interested maintainer as a champion to help finish it since there is often a lot of drift on any change I end up having to rebase.. I would want to make sure the time spent is mutually valuable.

</details>

alexhancock:
@codefromthecrypt Thanks for the thoughts.

> opinion that how session names are done in goose is already good enough

Yes, this is the stance I am taking with this change

@benbrandt does that present any issues from an ACP perspective?

alexhancock:
@DOsinga Removed the things `agent.reply` already does - thanks for catching

crates/goose-cli/src/commands/acp.rs
alexhancock: Updated

crates/goose-cli/src/commands/acp.rs
alexhancock: Updated

alexhancock:
Going to merge. Can always update/iterate if we want different naming/visibility. But I like the consistency with other sessions here.

benbrandt:
Sorry catching up here @alexhancock what is the specific question?
In my mind the agent would "own" the title generation, but maybe I am misunderstanding the question?

DOsinga:
> @alexhancock feel free to punt me somewhere (acp discussion) and also take my opinions as a grain of salt..
> 
> TL;DR; maybe we need to know who should own the session name/title and meanwhile there seems to be an opinion that how session names are done in goose is already good enough(?).
> 
> Below for longer thoughts as I have looked at this perhaps too much.

Just for my understanding, does this matter? we name the session in our db, because, well we do. but do we ever show this in the context of ACP? if we can get to the name of the session according to ACP/zed, we could fix the name to that, or possibly ACP: <their name>. the only place the user will see this, is in our session logs. I think

benbrandt:
Yep you all can do as you like. I got asked for a comment, and have no context on what any of this is doing, so just trying to help.

codefromthecrypt:
I think the reason why this is a difficult discussion is because ACP is a work in progress and includes limited functionality. Zed is not the only editor, as well, but it is what we are looking at.

Think about sessions where you go get a historical one to resume. The names being different on the frontend vs the backend would probably be an accident. If ACP is to replace a native integration, it would feel natural to have only one session title/name, hence the "ownership" concern. I feel like folks don't want to think about this now, and that's perfectly valid for the same reasons as things being early days. We can revisit it later!

PR #5659, opened 2025-11-10
fix: disallow runaway subagent chains

alexhancock:
We need to disable subagents being able to create other subagents. It might be cleaner to not even advertise the tool, but looks like it would be a bigger refactor to do so

@tlongwell-block @DOsinga Let me know if this looks like a good fix to you for the immediate term

Demo:

Screenshot 2025-11-10 at 4 27 06 PM


DOsinga (approved):
we can go with this for now since we need something, but it should ideally be done in `should_enabled_subagents` I think

PR #5668, opened 2025-11-11
feat: provider detection

michaelneale:
This is more a PoC at this stage, but wasn't hard to do - when desktop starts, it can sniff out environment to see if there is an anthropic, openai or openrouter api key already available, and offer with one click to just use it: 

Screenshot 2025-11-11 at 4 00 37 pm

Screenshot 2025-11-11 at 4 01 42 pm


michaelneale:
@DOsinga yeah this was very much a sketch to see if it coudl work and what it feels like (and can work!). 

I like the idea of auto detect env, then if they have an api to paste, and falling back to something like some built in signup to an LLM so they don't have to leave goose no matter what path they take?

michaelneale:
closing this as stale.

Issue #5651, opened 2025-11-10
Context progress bar not resetting after /clear command


catbilyeu:
**Describe the bug**

💡 Before filing, please check common issues:  
https://block.github.io/goose/docs/troubleshooting  

📦 To help us debug faster, attach your **diagnostics zip** if possible.  
👉 How to capture it: https://block.github.io/goose/docs/troubleshooting/diagnostics-and-reporting/

---

**To Reproduce**
1. Start a Goose CLI session
  2. Send several messages to build up context usage (e.g., ask multiple questions)
  3. Observe the context progress bar showing token usage (e.g., Context: ●●●●●●●○○○ 70% (189053/272000 tokens))
  4. Type /clear to clear the chat context
  5. Observe that the context progress bar still shows the old percentage instead of resetting to 0%

---

**Expected behavior**
  After running /clear, the context progress bar should reset to show 0% usage with all empty circles:
  Context: ○○○○○○○○○○ 0% (0/272000 tokens)

---

**Screenshots**
<img width="987" height="204" alt="Image" src="https://github.com/user-attachments/assets/18b31d90-3dff-407b-9e6f-7f573d7eb643" />

---

**Please provide the following information**
- **OS & Arch:** macOS 26
- **Interface:** CLI
- **Version:** 1.13.0
- **Extensions enabled:** chromedevtools, developer, computercontroller, todo
- **Provider & Model:** OpenAI – gpt-5

---

**Additional context**

PR #5679, opened 2025-11-11
Fix: /clear doesn't update session metrics

katzdave:
Fixes. https://github.com/block/goose/issues/5651


katzdave:
oops saw there's already a fix out here for this

PR #5687, opened 2025-11-11
fix: make image processing work in github copilot provider

### LINKED ISSUE #5497: Image processing fails when using GitHub CoPilot as LLM provider
Trying to reference an image file whenever using GitHub CoPilot as a LLM provider, fails with an error  "Request failed: missing required Copilot-Vision-Request header for vision requests (code: ) (status 400)"

**To Reproduce**
1. Go to codename goose, a desktop app running on Windows
2. Select GoPilot->gpt-5-mini Model
3. Go to new chat and upload an image file
4. Ask goose to tell what is on the file
5. See error above

**Expected behavior**
goose should be able to read the file and understand what is in it, whenever an LLM supporting vision is selected. 

 - **OS & Arch:** Windows Server 2022 Standard
 - **Interface:** UI
 - **Version:** 38.2.1
 - **Extensions enabled:** official Atlassian, Azure DevOps, Brave MCPs, Developer Tools
 - **Provider & Model:** CoPilot / gpt-mini-5

**Additional context**
If changing the provider to Openrouter and selecting the same model gpt-5-mini, the image is correctly processed.

alexhancock:
Fixes https://github.com/block/goose/issues/5497 by sending the `Copilot-Vision-Request` header with a value of true when there is a processed image in the message

Working example after the change:

Screenshot 2025-11-11 at 3 50 58 PM

### DIFF
--- a/crates/goose/src/providers/githubcopilot.rs
+++ b/crates/goose/src/providers/githubcopilot.rs
@@ -132,6 +132,31 @@ pub struct GithubCopilotProvider {
 }
 
 impl GithubCopilotProvider {
+    fn payload_contains_image(payload: &Value) -> bool {
+        if let Some(messages) = payload.get("messages").and_then(|m| m.as_array()) {
+            for msg in messages {
+                if let Some(content) = msg.get("content") {
+                    if content.is_array() {
+                        for item in content.as_array().unwrap() {
+                            if let Some(t) = item.get("type").and_then(|v| v.as_str()) {
+                                if t == "image_url" || t == "image" {
+                                    return true;
+                                }
+                            }
+                        }
+                    } else if content.is_object() {
+                        if let Some(t) = content.get("type").and_then(|v| v.as_str()) {
+                            if t == "image_url" || t == "image" {
+                                return true;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        false
+    }
+
     pub async fn from_env(model: ModelConfig) -> Result<Self> {
         let client = Client::builder()
             .timeout(Duration::from_secs(600))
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/githubcopilot.rs
```diff
@@ -132,6 +132,31 @@ pub struct GithubCopilotProvider {
 }
 
 impl GithubCopilotProvider {
+    fn payload_contains_image(payload: &Value) -> bool {
+        if let Some(messages) = payload.get("messages").and_then(|m| m.as_array()) {
+            for msg in messages {
+                if let Some(content) = msg.get("content") {
```
DOsinga: you could do this more functional, like messages.iter.any and then cast the is_object to a single list but up to you

crates/goose/src/providers/githubcopilot.rs
```diff
@@ -164,10 +189,14 @@ impl GithubCopilotProvider {
         let (endpoint, token) = self.get_api_info().await?;
         let url = url::Url::parse(&format!("{}/chat/completions", endpoint))
             .map_err(|e| ProviderError::RequestFailed(format!("Invalid base URL: {e}")))?;
+
+        // Build headers and include Copilot-Vision-Request when payload contains processed images
```
DOsinga: I'd drop this comment, but do what it says :)

adding payload to get_github_headers seems a tad awkward. if you just break up that builder below you just go:
```
let mut request = self.client.post(..).headers()
if has_images:
  request = request.header(that_header)
response = request.header(..).send.await?
```
or some such

then you can leave the other bits alone

DOsinga (approved):

PR #5688, opened 2025-11-11
fix: do not take into account gitignore in developer mcp

### LINKED ISSUE #2911: .gooseignore (.gitignore) blocks commands that happen to match files
**Describe the bug**

I have `build/` in my `.gitignore` file. I am now trying to get goose to compile some go code to fix the errors, but the command `go build` is matching and being blocked from being run.

**To Reproduce**

1. Add a directory `build/` to a `.gitignore` or `.gooseignore`
2. Ask goose to run a command with build in it like `go build`

**Expected behavior**

Should still run the command since the command doesn't match the build directory. 

**Screenshots**

```
( O)> Fix the compilation errors for this command "go build -o /dev/null ./cmd/tools/bindingsreplayer"
I'll fix it. I'll run the command to see the errors first.
─── shell | developer ──────────────────────────
command: go build -o /dev/null ./cmd/tools/bindingsreplayer


I see that the `go build` command is restricted in this environment. I can use other Go tools to check for compilation errors. I'll start by running `go vet` on the package to identify any issues.
```

 - **OS & Arch:** -- macos 15
 - **Interface:** CLI
 - **Version:** ` 1.0.27-canary+1326db2`
 - **Extensions enabled:** `Developer`
 - **Provider & Model:** Gemini Pro 06-05

**Additional context**

alexhancock:
Do not couple gitignore to gooseignore when considering commands operating on file paths in the developer MCP

Fixes https://github.com/block/goose/issues/2911

Example of a command working which would not work right now after this fix:

Screenshot 2025-11-11 at 4 09 40 PM


michaelneale (approved):
I think this makes sense, as does seem wrong. Main concern was goose rat-holing into node_modules and .venv and other derived directories wasting context, but I think it is smarter now, so not really needed.

PR #5694, opened 2025-11-12
Canonical models for Providers

katzdave:
Build core canonical model implementation

Two scripts
- build_canonical_model -> pulls all models from openrouter and loads them into our json file
- canonical_model_checker -> produces the full mappings from our providers as to which (provider, models) are being mapped into which canonical models.

Providers can use map_to_canonical_model to lookup this info (if the model is mapped). 

Once we're settled on this will further integrate it in providers and pull + add more information to these.

Currently integrated in model filtering.

Pricing is next; The [hit rate ](https://github.com/block/goose/compare/dkatz/canonical-model-pricing?expand=1#diff-fa985c805e6526d22d512e21a4d478b6125ee62ba404229f1ccfdbe522e345f5) for the current system is pretty low outside of openrouter.

### DIFF
--- a/crates/goose-server/src/routes/config_management.rs
+++ b/crates/goose-server/src/routes/config_management.rs
@@ -51,7 +51,6 @@ pub struct ConfigKeyQuery {
     pub is_secret: bool,
 }
 
-
 #[derive(Serialize, ToSchema)]
 pub struct ConfigResponse {
     pub config: HashMap<String, Value>,
--- a/crates/goose/src/providers/canonical/name_builder.rs
+++ b/crates/goose/src/providers/canonical/name_builder.rs
@@ -1,9 +1,7 @@
 use once_cell::sync::Lazy;
 use regex::Regex;
 
-static NORMALIZE_VERSION_RE: Lazy<Regex> = Lazy::new(|| {
-    Regex::new(r"-(\d)-(\d)(-|$)").unwrap()
-});
+static NORMALIZE_VERSION_RE: Lazy<Regex> = Lazy::new(|| Regex::new(r"-(\d)-(\d)(-|$)").unwrap());
 
 static STRIP_PATTERNS: Lazy<Vec<Regex>> = Lazy::new(|| {
     vec![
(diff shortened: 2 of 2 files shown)


crates/goose/src/providers/canonical/data/canonical_mapping_report.json
```diff
@@ -0,0 +1,5271 @@
+{
```
DOsinga, line 1: I guess the other one is the dupe

crates/goose/canonical_mapping_report.json
```diff
@@ -0,0 +1,4413 @@
+{
```
DOsinga, line 1: these models list a lot of stuff that seems to be block specific

michaelneale:
nice - @katzdave how would you like this to be tested/reviewed, as a lot of new stuff

katzdave:
> nice - @katzdave how would you like this to be tested/reviewed, as a lot of new stuff

Mostly just if this is the general abstraction we want? Its all pretty standalone right now map_to_canonical_model will let you pull the data from within the provider. Otherwise canonical models holds the current state (you can update it via re-running build_canonical_models).

The mapping report isn't strictly necessary, but lets you see the diff between how existing models got mapped so you can audit changes to the canonical models json more carefully.

Thinking we get some consensus on what we want to store in the repo and then can do the integration across more providers incrementally

crates/goose/src/providers/anthropic.rs
```diff
@@ -233,7 +233,7 @@ impl Provider for AnthropicProvider {
         }
 
         let json = response.payload.unwrap_or_default();
-        let arr = match json.get("models").and_then(|v| v.as_array()) {
+        let arr = match json.get("data").and_then(|v| v.as_array()) {
```
DOsinga, line 245: huh? was this broken?

crates/goose/src/providers/base.rs
```diff
@@ -415,6 +416,66 @@ pub trait Provider: Send + Sync {
...
+                map_to_canonical_model(provider_name, model, registry)
+                    .and_then(|canonical_id| registry.get(&canonical_id))
+                    .map(|m| m.input_modalities.contains(&"text".to_string()))
+                    .unwrap_or(false)
+            })
+            .cloned()
+            .collect();
+
+        if recommended_models.is_empty() {
+            Ok(Some(all_models))
```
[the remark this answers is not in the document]
katzdave: can filter all models and that's just a bad experience to return nothing.

crates/goose-server/src/routes/config_management.rs
```diff
@@ -51,6 +51,12 @@ pub struct ConfigKeyQuery {
     pub is_secret: bool,
 }
 
+#[derive(Deserialize, ToSchema, utoipa::IntoParams)]
+pub struct ModelFilterQuery {
+    #[serde(default)]
+    pub show_all: bool,
```
DOsinga: let's not default booleans if it can be avoided.  also do we need this? are we not sure about ourselves that we can just always filter (the user can still override by typing the model name)

crates/goose/examples/build_canonical_models.rs
```diff
@@ -0,0 +1,254 @@
...
+///
+use anyhow::{Context, Result};
+use goose::providers::canonical::{
+    canonical_name, CanonicalModel, CanonicalModelRegistry, Pricing,
+};
+use serde_json::Value;
+use std::collections::HashMap;
+
+const OPENROUTER_API_URL: &str = "https://openrouter.ai/api/v1/models";
+const ALLOWED_PROVIDERS: &[&str] = &[
```
DOsinga, line 15: could we base this on our own providers?

crates/goose/src/providers/base.rs
```diff
@@ -415,6 +416,66 @@ pub trait Provider: Send + Sync {
...
+        let all_models = match self.fetch_supported_models().await? {
+            Some(models) => models,
+            None => return Ok(None),
+        };
+
+        let registry = CanonicalModelRegistry::bundled().map_err(|e| {
+            ProviderError::ExecutionError(format!("Failed to load canonical registry: {}", e))
+        })?;
+
+        use super::canonical::map_to_canonical_model;
```
DOsinga: why the inline import?

crates/goose/src/providers/base.rs
```diff
@@ -415,6 +416,66 @@ pub trait Provider: Send + Sync {
         Ok(None)
     }
 
+    /// Fetch models filtered by canonical registry and usability
+    async fn fetch_recommended_models(&self) -> Result<Option<Vec<String>>, ProviderError> {
+        let force_show_all = std::env::var("FORCE_SHOW_ALL_MODELS")
```
DOsinga: why do we have this environment variable?

crates/goose/examples/canonical_model_checker.rs
```diff
@@ -0,0 +1,369 @@
+/// Canonical Model Checker
```
DOsinga, line 1: should this be in examples?

in general how are we going to integrate this with the overall process. how long does it take to run all of this? can we add it to the CI and then fail when it is not up to date? or add it to the release process /cc @jamaedo

crates/goose/src/providers/canonical/name_builder.rs
```diff
@@ -0,0 +1,869 @@
...
+    vec![
+        Regex::new(r"-latest$").expect("Failed to compile -latest regex"),
+        Regex::new(r"-preview(-\d+)*$").expect("Failed to compile -preview regex"),
+        Regex::new(r"-exp(-\d+)*$").expect("Failed to compile -exp regex"),
+        Regex::new(r":exacto$").expect("Failed to compile :exacto regex"),
+        Regex::new(r"-\d{8}$").expect("Failed to compile date regex"),
+        Regex::new(r"-\d{4}-\d{2}-\d{2}$").expect("Failed to compile full-date regex"),
+        Regex::new(r"-v\d+(\.\d+)*$").expect("Failed to compile version regex"),
+        Regex::new(r"-\d{3,}$").expect("Failed to compile patch regex"),
+        Regex::new(r"-bedrock$").expect("Failed to compile -bedrock regex"),
```
DOsinga: why the expects here? does it make things clearer?

crates/goose/src/providers/canonical/name_builder.rs
```diff
@@ -0,0 +1,869 @@
...
+
+    // OpenRouter models are already in canonical format
+    if provider == "openrouter" {
+        model_base
+    } else {
+        format!("{}/{}", provider, model_base)
+    }
+}
+
+/// Try to map a provider/model pair to a canonical model
```
DOsinga, line 55: this has a bunch of trivial comments in there that can be deleted, but also repeats the same pattern where you generate a bunch of name patterns and then see if they exists but they are all of the form:

let candidate = canonical_name(provider, &model_stripped);
        if registry.get(&candidate).is_some() {
        
can we simplify?

crates/goose/src/providers/canonical/name_builder.rs
```diff
@@ -0,0 +1,869 @@
...
+        || model_lower.starts_with("codestral")
+        || model_lower.starts_with("ministral")
+        || model_lower.starts_with("pixtral")
+        || model_lower.starts_with("devstral")
+        || model_lower.starts_with("voxtral")
+    {
+        return Some("mistralai");
+    }
+
+    if model_lower.starts_with("deepseek") || model_lower.contains("deepseek") {
```
DOsinga: this is also very repetitive. but also, why test if it starts with something and then check if it contains something? contains >= startswith no?

crates/goose/src/providers/canonical/name_builder.rs
```diff
@@ -0,0 +1,869 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_canonical_names() {
+        // canonical_name tests
+        assert_eq!(
```
DOsinga: again, very repetitive - use parameterized test. also this is an insane long list of tests. strip them down to the ones that we actually need.

crates/goose/examples/canonical_model_checker.rs
katzdave: /cc @jamadeo (working tag)

I was thinking add to the release process. Run build_canonical_models, followed by canonical_model_checker.

So then at least we have some record in the release logs of the model diff.

pricing.rs uses a local cache (refreshes weekly), so that's also worth considering if we're worried about freshness, but I think releases should fresh enough.

crates/goose/examples/build_canonical_models.rs
katzdave: It's based on all the ones I have keys for. I took off databricks because it's a bunch of block specific model names, but can potentially add that back (and a lot of the logic in the name converter specifically addresses some of the cases I saw there).

crates/goose-server/src/routes/config_management.rs
katzdave: Yeah that's fair. Got rid of all of the boolean + the cli flag.

So the auto-complete will always filter based on this, but you can ignore it and just type whatever.

crates/goose/src/providers/base.rs
katzdave: gone

crates/goose/src/providers/canonical/name_builder.rs
katzdave: Done. Now a single parameterized test on just the outermost fn.

PR #5699, opened 2025-11-12
Improve AWS credential loading and configuration handling in BedrockProvider 

fbalicchia:
This PR refactors the BedrockProvider::from_env implementation to make AWS configuration loading more robust and user-friendly, especially for environments using AWS SSO or multiple profiles.

⸻

Key Changes
• Replaced aws_config::load_from_env() with aws_config::defaults(BehaviorVersion::latest()) to support:
• AWS SSO logins (aws sso login --profile )
• Named profiles (AWS_PROFILE)
• Explicit region configuration (AWS_REGION)
• Added explicit handling for:
• AWS_PROFILE → sets the active AWS profile
• AWS_REGION → sets the AWS region dynamically

### DIFF
--- a/crates/goose/src/providers/bedrock.rs
+++ b/crates/goose/src/providers/bedrock.rs
@@ -22,12 +22,13 @@ use super::formats::bedrock::{
 pub const BEDROCK_DOC_LINK: &str =
     "https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html";
 
-pub const BEDROCK_DEFAULT_MODEL: &str = "anthropic.claude-sonnet-4-20250514-v1:0";
+pub const BEDROCK_DEFAULT_MODEL: &str = "us.anthropic.claude-sonnet-4-5-20250929-v1:0";
 pub const BEDROCK_KNOWN_MODELS: &[&str] = &[
-    "anthropic.claude-sonnet-4-20250514-v1:0",
-    "anthropic.claude-3-7-sonnet-20250219-v1:0",
-    "anthropic.claude-opus-4-20250514-v1:0",
-    "anthropic.claude-opus-4-1-20250805-v1:0",
+    "us.anthropic.claude-sonnet-4-5-20250929-v1:0",
+    "us.anthropic.claude-sonnet-4-20250514-v1:0",
+    "us.anthropic.claude-3-7-sonnet-20250219-v1:0",
+    "us.anthropic.claude-opus-4-20250514-v1:0",
+    "us.anthropic.claude-opus-4-1-20250805-v1:0",
 ];
 
 pub const BEDROCK_DEFAULT_MAX_RETRIES: usize = 6;
@@ -67,7 +68,6 @@ impl BedrockProvider {
         // Use load_defaults() which supports AWS SSO, profiles, and environment variables
         let mut loader = aws_config::defaults(aws_config::BehaviorVersion::latest());
 
-        // Check for AWS_PROFILE configuration
         if let Ok(profile_name) = config.get_param::<String>("AWS_PROFILE") {
             if !profile_name.is_empty() {
                 loader = loader.profile_name(&profile_name);
@@ -204,7 +204,7 @@ impl Provider for BedrockProvider {
             BEDROCK_DOC_LINK,
             vec![
                 ConfigKey::new("AWS_PROFILE", true, false, Some("default")),
-                ConfigKey::new("AWS_REGION", true, false, Some("us-east-1")),
+                ConfigKey::new("AWS_REGION", true, false, None),
             ],
         )
     }


fbalicchia:
@DOsinga have you any update ?

crates/goose/src/providers/bedrock.rs
```diff
@@ -175,11 +198,14 @@ impl Provider for BedrockProvider {
...
             "Amazon Bedrock",
-            "Run models through Amazon Bedrock. You may have to set 'AWS_' environment variables to configure authentication.",
+            "Run models through Amazon Bedrock. Supports AWS SSO profiles - run 'aws sso login --profile <profile-name>' before using. Configure with AWS_PROFILE and AWS_REGION, or use environment variables/credentials.",
             BEDROCK_DEFAULT_MODEL,
             BEDROCK_KNOWN_MODELS.to_vec(),
             BEDROCK_DOC_LINK,
-            vec![ConfigKey::new("AWS_PROFILE", true, false, Some("default"))],
+            vec![
+                ConfigKey::new("AWS_PROFILE", true, false, Some("default")),
+                ConfigKey::new("AWS_REGION", true, false, Some("us-east-1")),
```
michaelneale: @fbalicchia did you mean to add us east 1 hard coded in?

crates/goose/src/providers/bedrock.rs
```diff
@@ -22,12 +22,12 @@ use super::formats::bedrock::{
...
     "https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html";
 
-pub const BEDROCK_DEFAULT_MODEL: &str = "us.anthropic.claude-sonnet-4-20250514-v1:0";
+pub const BEDROCK_DEFAULT_MODEL: &str = "anthropic.claude-sonnet-4-20250514-v1:0";
 pub const BEDROCK_KNOWN_MODELS: &[&str] = &[
-    "us.anthropic.claude-sonnet-4-20250514-v1:0",
-    "us.anthropic.claude-3-7-sonnet-20250219-v1:0",
-    "us.anthropic.claude-opus-4-20250514-v1:0",
-    "us.anthropic.claude-opus-4-1-20250805-v1:0",
+    "anthropic.claude-sonnet-4-20250514-v1:0",
```
michaelneale: odd - it wouldn't work without the prefix here for me, but may be different setup with sso vs not?

crates/goose/src/providers/bedrock.rs
```diff
@@ -62,16 +64,39 @@ impl BedrockProvider {
         set_aws_env_vars(config.all_values());
         set_aws_env_vars(config.all_secrets());
 
-        let sdk_config = aws_config::load_from_env().await;
+        // Use load_defaults() which supports AWS SSO, profiles, and environment variables
+        let mut loader = aws_config::defaults(aws_config::BehaviorVersion::latest());
+
+        // Check for AWS_PROFILE configuration
```
michaelneale: nit: would it be ok to trim some of the one line comments, I know goose/others like to add them all the time, but they can add noise/mistakes in future as code drifts (ie only really needed ones - like the one on like 67 which talks about SSO is great).

michaelneale (changes requested):
thanks @fbalicchia - devilishly hard to test all the edge cases for bedrock implementations isn't it. 

I had some nit/comments in review - but main thing is: have had issues before with it not really refreshing creds and then goose basically crashes/stops working until you close and open, is that likely to be an issue with this with the `client` at the BedrockProvider level? or will it load things when it needs? 

If you think this is ok, then would just love small comments addressed (and also - do we need us-east-1 in there?) and lets get this in, and can see how it goes?

crates/goose/src/providers/bedrock.rs
```diff
@@ -37,6 +37,8 @@ pub const BEDROCK_DEFAULT_MAX_RETRY_INTERVAL_MS: u64 = 120_000;
 
 #[derive(Debug, serde::Serialize)]
 pub struct BedrockProvider {
+    #[serde(skip)]
+    client: Client,
```
michaelneale, line 42: I think ok, but in the past heard that it would fail to refresh things when creds expire (but I see it is calling load_from_env in same scope so probably fine?)

crates/goose/src/providers/bedrock.rs
fbalicchia: Thanks for the review.
Good catch. Honestly, I haven’t tested the credential-refreshing use case, but I think the behavior is currently the same. I agree with you that it should reload credentials when needed or provide some kind of refresh mechanism when the token expires.

Regarding the region, I added a default one because it’s usually where AWS releases new things first. That said, it should definitely be configurable by the user.  I'll put default empty

crates/goose/src/providers/bedrock.rs
fbalicchia: No—this doesn’t depend on SSO here. It depends on which profile we want to use by default, whether the global profile or the inference profile. I’ll revert it

fbalicchia:
@michaelneale Done thanks

michaelneale:
thank you @fbalicchia

PR #5700, opened 2025-11-12
Switch to Swatinem/rust-cache

jamadeo:
This action does all sorts of optimizations that are useful to us. Our cache is way too big and most of it isn't useful (it caches a bunch of incremental build artifacts from our own code, when we really just want to cache compiled dependencies)


jamadeo:
merged as part of #5702

PR #5701, opened 2025-11-12
Fix: Always show autocompact threshold ui

katzdave:
This does duplicate the default threshold in the UI, but I think that's fine since it's so easy to change there (if it actually can render).

### DIFF
--- a/ui/desktop/src/components/alerts/AlertBox.tsx
+++ b/ui/desktop/src/components/alerts/AlertBox.tsx
@@ -138,7 +138,7 @@ export const AlertBox = ({ alert, className }: AlertBoxProps) => {
                       handleSaveThreshold();
                     } else if (e.key === 'Escape') {
                       setIsEditingThreshold(false);
-                      const resetValue = currentThreshold ? Math.round(currentThreshold * 100) : 80;
+                      const resetValue = Math.round(currentThreshold * 100);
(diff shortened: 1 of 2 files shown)


ui/desktop/src/components/alerts/AlertBox.tsx
```diff
@@ -49,7 +49,8 @@ export const AlertBox = ({ alert, className }: AlertBoxProps) => {
     loadThreshold();
   }, [read]);
 
-  const currentThreshold = loadedThreshold !== null ? loadedThreshold : alert.autoCompactThreshold;
+  const currentThreshold =
+    loadedThreshold !== null ? loadedThreshold : (alert.autoCompactThreshold ?? 0.8);
```
DOsinga: can we just make loadedThreshold not optional and initialize to 0.8? also delete autoCompactThreshold, it doesn't seem like we're using it (maybe we used to and that's how this broke?)

DOsinga (approved):

ui/desktop/src/components/alerts/AlertBox.tsx
katzdave: Done. I am still keeping threshold value + loadedThreshold. This lets the value sync across multiple goose windows if it changes (pretty important since the auto-compact will trigger based on the config value).

PR #5702, opened 2025-11-12
faster, cheaper (pick two): improve CI workflow and switch to free github runner

jamadeo:
This:
- splits jobs in CI
- uses better caching
- uses the free github runner

with the end goal of opening up our workflows to run on fork PRs without approval

### DIFF
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -78,29 +78,19 @@ jobs:
         if: matrix.use-cross
         run: source ./bin/activate-hermit && cargo install cross --git https://github.com/cross-rs/cross
 
-      # Cache Cargo registry and git dependencies for Windows builds
-      - name: Cache Cargo registry (Windows)
-        if: matrix.use-docker
-        uses: actions/cache@2f8e54208210a422b2efd51efaa6bd6d7ca8920f
+      # Cache Cargo artifacts for cross-platform builds
+      - name: Cache Cargo artifacts (Linux/macOS)
+        if: matrix.use-cross
+        uses: Swatinem/rust-cache@98c8021b550208e191a6a3145459bfc9fb29c4c0 # v2
         with:
-          path: |
-            ~/.cargo/registry/index
-            ~/.cargo/registry/cache
-            ~/.cargo/git/db
-          key: ${{ runner.os }}-cargo-registry-${{ hashFiles('Cargo.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-cargo-registry-
+          key: ${{ matrix.architecture }}-${{ matrix.target-suffix }}
 
-      # Cache compiled dependencies (target/release/deps) for Windows builds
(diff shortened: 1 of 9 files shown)


.github/workflows/build-cli.yml
```diff
@@ -78,29 +78,19 @@ jobs:
         if: matrix.use-cross
         run: source ./bin/activate-hermit && cargo install cross --git https://github.com/cross-rs/cross
 
-      # Cache Cargo registry and git dependencies for Windows builds
-      - name: Cache Cargo registry (Windows)
-        if: matrix.use-docker
-        uses: actions/cache@2f8e54208210a422b2efd51efaa6bd6d7ca8920f
+      # Cache Cargo artifacts for cross-platform builds
```
DOsinga: might as well drop the comment (or change it into the name below)

.github/workflows/ci.yml
```diff
@@ -41,63 +41,79 @@ jobs:
...
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # pin@v4
+
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+
+      - uses: Swatinem/rust-cache@v2
+
       - name: Lint
         run: |
           source ./bin/activate-hermit
+          hermit uninstall rustup
```
DOsinga, line 94: why is that needed?

DOsinga (approved):

.github/workflows/ci.yml
jamadeo: maybe this could do with a comment

The swatinem/rust-cache github action is the go-to for caching rust projects because it does a good job of filtering out all of the incremental artifacts created from compiling your own project code, and keeping the dependencies. But there's one problem with our setup: it distinguishes between first and third party code by looking at the path on disk, and if the artifact source path prefix-matches your project directory, it marks it as (non-cached) project code. When we use hermit, the CARGO_HOME is within the project directory, whereas in a typical setup it would be at `~/.cargo`.

Here, since this step needs both rust (for the openapi code gen) plus npm and just, it made sense to just remove the rust bit from hermit while keeping the other parts.

This problem goes away if we decide to stop using hermit for the rust toolchain.

.github/workflows/ci.yml
```diff
@@ -41,63 +41,80 @@ jobs:
...
+        uses: Swatinem/rust-cache@v2
 
       - name: Build and Test
         run: |
           gnome-keyring-daemon --components=secrets --daemonize --unlock <<< 'foobar'
-          source ../bin/activate-hermit
           export CARGO_INCREMENTAL=0
-          cargo test --jobs 2
+          cargo test -- --skip scenario_tests::scenarios::tests
+          cargo test --jobs 1 scenario_tests::scenarios::tests
```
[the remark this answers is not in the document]
jamadeo: just wrong

Issue #3919, opened 2025-08-07
Enable slash command support in Goose Desktop when using Claude Code and other CLI providers


stnkvcs:
Please explain the motivation behind the feature request.
When using Claude Code as a provider in Goose Desktop, slash commands (such as /t for theme toggling, /help, /plan, etc.) are not recognized or processed correctly. Instead of executing the command, Goose interprets them as regular text input. This is problematic because:

These slash commands are documented as standard features in the Goose CLI and should provide consistent functionality across all interfaces
Users coming from the CLI expect these commands to work in the Desktop client
Important workflow commands like /plan (for reviewing AI intentions), /summarize (for context management), and /recipe (for saving session recipes) are inaccessible
Theme switching via /t would improve readability and user comfort during long coding sessions

This issue specifically occurs when using Claude Code as the provider - the commands are typed but Goose responds as if they're regular conversation text rather than commands.
Describe the solution you'd like
Enable full slash command support in Goose Desktop when using Claude Code (and other CLI providers) as the backend. Specifically:

Recognize and process all documented slash commands (/t, /help, /plan, /summarize, /recipe, /prompt, /mode, etc.)
Provide visual feedback when a slash command is recognized (e.g., syntax highlighting or autocomplete)
Ensure feature parity between Goose CLI and Goose Desktop interfaces
Add slash command documentation/hints in the Desktop UI (perhaps in a help menu or tooltip)

Describe alternatives you've considered

Using Goose CLI instead - This works but loses the benefits of the Desktop GUI (better visual layout, easier session management, integrated file handling)
GUI buttons/menus for commands - Adding UI buttons for common commands like theme switching, but this doesn't solve the problem for power users who prefer keyboard shortcuts
Keyboard shortcuts - Implementing keyboard shortcuts (e.g., Cmd+T for theme toggle) instead of slash commands, though this breaks consistency with the CLI interface

Additional context

This issue occurs specifically with Claude Code as the provider configured via GOOSE_PROVIDER=claude-code
The documentation clearly shows these commands should be available during active sessions
Example of current behavior: Typing /t results in Goose responding "I see you typed '/t' - I'm not sure what specific command or action you're looking for there" instead of toggling the theme
This affects all slash commands, not just theme toggling
Having these commands work would significantly improve the developer experience, especially for users who frequently switch between CLI and Desktop interfaces

- [x] I have verified this does not duplicate an existing feature request

DOsinga:
[...]

PR #5718, opened 2025-11-13
Slash commands

DOsinga:
## Summary

implements slash commands.

https://github.com/user-attachments/assets/b012d08d-891f-4224-be2b-7b806b83eebb


Abhijay007:
It's good, but just have a thought around this, as we're soon going to have the same slash commands same as CLI in the desktop app, based on the changes/featues requested in this issue: https://github.com/block/goose/issues/3919
 and ongoing work in this branch: https://github.com/block/goose/compare/spence+alexhancock/text-input-rebased.

This implementation is good, but maybe it's not the best way to call recipes and extensions as reference or to load them. I shared this earlier in the Discord discussion, it might be better if we use `@` with fuzzy search for calling recipes and extensions. It would feel more convenient for people coming to Goose from platforms like Zed or Cursor, since they already use`@` to trigger features or reference anything in the chat, like past conversations, configs, or files etc.

maybe something like this : very basic and example implementation I created way back : 

addContent

not sure but I think this might be better to adopt and use :)

crates/goose/src/slash_commands.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+pub fn resolve_slash_command(command: &str) -> Option<Recipe> {
+    let recipe_path = get_recipe_for_command(command)?;
+
+    if !recipe_path.exists() {
+        return None;
+    }
+    let recipe_content = std::fs::read_to_string(&recipe_path).ok()?;
+    let recipe: Recipe = serde_yaml::from_str(&recipe_content).ok()?;
```
lifeizhou-ap: - could use Recipe.from_content function with consistent logic. 
- this function is used in agent reply function to get the recipe from slash command and then extract the prompts and instructions.  If the recipe has parameters, currently the parameters won't be resolved.  Shall we ask the agent to prompt users to enter the parameters?

DOsinga:
> It's good, but just have a thought around this, as we're soon going to have the same slash commands same as CLI in the desktop app, based on the changes/featues requested in this issue: #3919 and ongoing work in this branch: https://github.com/block/goose/compare/spence+alexhancock/text-input-rebased.
> 
> This implementation is good, but maybe it's not the best way to call recipes and extensions as reference or to load them. I shared this earlier in the Discord discussion, it might be better if we use `@` with fuzzy search for calling recipes and extensions. It would feel more convenient for people coming to Goose from platforms like Zed or Cursor, since they already use`@` to trigger features or reference anything in the chat, like past conversations, configs, or files etc.
> 
> maybe something like this : very basic and example implementation I created way back :
>
> not sure but I think this might be better to adopt and use :)

maybe. custom slash commands seem pretty standard - you want to execute a command and it just expands. fuzzy search is I think a bit of a different thing where you want to add something to a conversation like a file - adding a recipe that way would insert the recipe contents in the conversation I would think?

crates/goose/src/slash_commands.rs
DOsinga: yeah good points.  parameters I think we should get from suggestions/autocomplete /cc @spencrmartin

Abhijay007:
> maybe. custom slash commands seem pretty standard - you want to execute a command and it just expands. fuzzy search is I think a bit of a different thing where you want to add something to a conversation like a file - adding a recipe that way would insert the recipe contents in the conversation I would think? 

Yeah but I think if the recipe contents get inserted, it will get executed, not like just adding it as prompt. It’s like calling the recipe. But yeah maybe slash commands can be the way to go for now, and later we might edit it in fuzzy search to execute directly after selecting it after seeing the response if needed.

I was thinking of it like: if I want to call the `/joke` recipe, I just do `@`, go to recipes and search for `joke`, and once selected it becomes `/joke` in the chat, and on hitting enter it just gets executed."

crates/goose-server/src/routes/config_management.rs
```diff
@@ -390,6 +407,30 @@ pub async fn get_provider_models(
     }
 }
 
+#[utoipa::path(
+    get,
+    path = "/config/slash_commands",
```
alexhancock, line 412: I feel like this might be more natural to have its own set of routes vs being under `config`. I think of `config_managment.rs` routes as just about changing settings, but with this change slash commands are becoming a feature of their own. Similar to how recipes have their own route file.

DOsinga: maybe? this does just write something into the settings though. it is more similar to extensions and providers; sure they are their own features, but in terms of routes here we expose how to change it in the config file.

recipes are stored in their own folder(s).

jonandersen:
@DOsinga looking to add support for this into https://github.com/block/ai-rules/
is there a repo level folder we can use e.g `.agents/commands` ?

PR #5706, opened 2025-11-12
Next camp refactor live

zanesq:
## Summary
Follow up to release next camp refactoring efforts removing ALPHA flag condition and related cleanup.

This enables launching sessions and recipes in the same window along with various refactors to improve how we load chats on the front end.

- [x] Fix sending message from hub not starting chat and flickering
- [x] Fix quick launcher not working
- [x] Fix create and run recipe during chat
- [x] Added proper error handling ui for failed session loading (hide chat for now)
- [x] Fixed recipe and extension deeplinks not working
- [x] Removed debug logging
- [x] Add option to launch recipes and sessions in new windows
- [x] Fix manual approval mode not asking for approval mid chat [https://github.com/block/goose/pull/5726]
- [x] Bring over missing functionality from useChatEngine (power saver and notifications, etc) and remove from project [https://github.com/block/goose/pull/5734]
- [x] Implement message editing functionality on the server [https://github.com/block/goose/pull/5813]

closes https://github.com/block/goose/issues/4980

PR #5726, opened 2025-11-13
Next camp tool notifications

zanesq:
## Summary
Add back tool notifications support for manual approval mid chat.
Verified locally

follow up for https://github.com/block/goose/pull/5706

### DIFF
--- a/crates/goose/src/agents/tool_execution.rs
+++ b/crates/goose/src/agents/tool_execution.rs
@@ -71,12 +71,14 @@ impl Agent {
                             }
                         });
 
-                    let confirmation = Message::assistant().with_tool_confirmation_request(
-                        request.id.clone(),
-                        tool_call.name.to_string().clone(),
-                        tool_call.arguments.clone().unwrap_or_default(),
-                        security_message,
-                    );
+                    let confirmation = Message::assistant()
+                        .with_tool_confirmation_request(
+                            request.id.clone(),
+                            tool_call.name.to_string().clone(),
+                            tool_call.arguments.clone().unwrap_or_default(),
+                            security_message,
+                        )
+                        .user_only();
                     yield confirmation;
 
                     let mut rx = self.confirmation_rx.lock().await;
--- a/ui/desktop/src/components/GooseMessage.tsx
+++ b/ui/desktop/src/components/GooseMessage.tsx
@@ -9,11 +9,11 @@ import {
   getToolRequests,
   getToolResponses,
   getToolConfirmationContent,
+  NotificationEvent,
 } from '../types/message';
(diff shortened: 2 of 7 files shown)


zanesq:
Closing while I look into a backend solution that doesn't modify messages on the front end

crates/goose/src/agents/tool_execution.rs
```diff
@@ -71,7 +71,7 @@ impl Agent {
                             }
                         });
 
-                    let confirmation = Message::user().with_tool_confirmation_request(
+                    let confirmation = Message::assistant().with_tool_confirmation_request(
```
zanesq: I looked up what llms expect for this role and there is no clear standard. Goose says using Message::assistant() for tool confirmations is fine and arguably more semantically correct. The assistant is asking for permission, so it makes sense for it to come from the assistant role. Since these messages are filtered out before sending to the LLM, the role only matters for frontend rendering.

DOsinga: can't we/shouldn't we just set this to agent invisible?

ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -17,6 +17,8 @@ import { errorMessage } from '../utils/conversionUtils';
 
 const resultsCache = new Map<string, { messages: Message[]; session: Session }>();
 
+export type NotificationEvent = Extract<MessageEvent, { type: 'Notification' }>;
```
DOsinga: I would flip the dependency around here and move this to message.ts since it is a related type

ui/desktop/src/components/GooseMessage.tsx
```diff
@@ -112,9 +109,20 @@ export default function GooseMessage({
...
+            await confirmPermission({
+              body: {
+                session_id: sessionId,
+                id: toolConfirmationContent.id,
+                action: 'deny',
+              },
+              throwOnError: true,
+            });
+          } catch (error) {
+            console.error('Failed to send tool cancellation to backend:', error);
```
DOsinga, line 123: this will probably not fail, but we should be louder about it than logging since the agent I think will hang now

ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -32,6 +34,7 @@ interface UseChatStreamReturn {
   stopStreaming: () => void;
   sessionLoadError?: string;
   tokenState: TokenState;
+  notifications: NotificationEvent[];
```
DOsinga: I think the only consumer of this converts this to a map sorted by request_id - I would suggest to do that on construction

DOsinga (approved):

PR #5731, opened 2025-11-13
fix: get win node path from registry

vlascik:
## Summary
Current rebase of https://github.com/block/goose/pull/5233

On Windows when adding npx extensions, Goose tries to find node.exe at default paths in Program Files, and then fails for everyone that installed node elsewhere. This PR tries to also find the path from the registry, which should mitigate the problem for people that already have node installed, but in a different folder.

Also the error messages are made a bit clearer about what the problem is, so that users can try to rectify it on their own.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Build on Windows

### Related Issues
Relates to https://github.com/block/goose/issues/4030

### DIFF
--- a/crates/goose-server/src/routes/agent.rs
+++ b/crates/goose-server/src/routes/agent.rs
@@ -482,7 +482,8 @@ async fn agent_add_extension(
     State(state): State<Arc<AppState>>,
     Json(request): Json<AddExtensionRequest>,
 ) -> Result<StatusCode, ErrorResponse> {
-    if cfg!(target_os = "windows") {
+    #[cfg(windows)]
+    {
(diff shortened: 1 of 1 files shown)


jamadeo (approved):

vlascik:
@jamadeo linux build had a problem, it should be fixed now, but I think the action rerun requires a new approval? Could you please restart it?

Thank you.

PR #5733, opened 2025-11-14
Pr 5175 nowarmup verboseprompt

ATrueLight4:
## Summary

--no-warmup loads larger models faster, --verbose-prompt, see the actual prompts

### Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing

Manual
### Related Issues
Relates to #5175 
Discussion: LINK (if any)

Before:  

After:   

### Submitting a Recipe?

**Email**:

### DIFF
--- a/crates/goose/src/providers/embedded.rs
+++ b/crates/goose/src/providers/embedded.rs
@@ -648,6 +648,8 @@ impl ServerProcess {
             .arg("--json-schema")
             .arg("{}")
             .arg("--verbose")
+            .arg("--verbose-prompt")
+            .arg("--no-warmup")
             .stdout(std::process::Stdio::null())
             .stderr(std::process::Stdio::null());
 


ATrueLight4:
so I am in a quandary... I think my PR for adding --verbose-prompt and --no-warmup should be closed and deleted...

 I noticed that the llama-embedded branch was updated and stdio and stderr are now going to null, so no need for --verbose or --verbose-prompt... And if I add the --no-warmup to a new PR, for not spawning a window for llama-server when running the GUI in Windows the PR I just created is no longer needed...
 
Next I just tested not spawning a window, which never happened in CLI, and the change now no longer spawns a window for the llama-server in GUI under Windows and with the --no-warmup both CLI and GUI starts the larger gpt-oss in time to respond to first prompt given...

Problem... GUI version of goose still doesn't kill the llama-server process at least under Windows. The CLI did, does, and still does, stop the llama-server when the goose CLI stops

But with no window means user would need to open taskmanager or something to kill the llama-server process

I can create a PR with this change to not spawn a window for the llama-server under Windows, but making sure the GUI version of goose kills the llama-server process should probably occur first...

DOsinga:
can you sync to main to clean up the PR?

also could do with a bit more explanation of what the plan is here

zanesq:
@ATrueLight4 still in progress?

PR #5741, opened 2025-11-14
fix: Implement a CredentialStore for auth

### LINKED ISSUE #5259: MCP OAuth refresh_token not used correctly
**Describe the bug**

Goose in its current state does not (at least reliably?) accept the refresh_token when supplied by the auth server. It does not submit "offline_access" as a scope either, which may be required.

**To Reproduce**
1. Add an MCP server that follows the spec for auth (OAuth2.1 with DCR)
2. Wait a few minutes/hours for the access token to expire.
3. Goose fires off the auth flow again.

**Expected behavior**
The refresh_token should be persisted and used to update the access token. The old access token should be thrown away.

**Screenshots**
NA

 - **OS & Arch:** Apple M3/Darwin
 - **Interface:** CLI
 - **Version:** 1.11.0
 - **Extensions enabled:** A test MCP server requiring authentication and following the current MCP auth spec.
 - **Provider & Model:** Local (Qwen3 running on a 8xMi325X) (probably not relevant)

**Additional context**

For what it's worth, these two branches fix the issue. However, it's most definitely AI slop as I am not a Rust programmer.

https://github.com/ianschenck/rust-sdk/tree/ian/refresh-token-callback (rust-sdk)
https://github.com/ianschenck/goose/tree/ian/get-refresh-token (goose)

alexhancock:
Fixes https://github.com/block/goose/issues/5259

* Implements a new CredentialStore for goose to use to persist oauth token information on both initial auth and during token exchanges
* Ensures new refresh tokens will always be used

More context in https://github.com/modelcontextprotocol/rust-sdk/pull/542

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ uninlined_format_args = "allow"
 string_slice = "warn"
 
 [workspace.dependencies]
-rmcp = { version = "0.8.5", features = ["schemars", "auth"] }
+rmcp = { version = "0.9.0", features = ["schemars", "auth"] }
 
 # Patch for Windows cross-compilation issue with crunchy
 [patch.crates-io]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -281,9 +281,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.87"
+version = "0.1.89"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d556ec1359574147ec0c4fc5eb525f3f23263a592b1a9c07e0a75b427de55c97"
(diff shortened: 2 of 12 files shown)


Cargo.toml
```diff
@@ -15,7 +15,8 @@ uninlined_format_args = "allow"
 string_slice = "warn"
 
 [workspace.dependencies]
-rmcp = { version = "0.8.5", features = ["schemars", "auth"] }
+# rmcp = { version = "0.8.5", features = ["schemars", "auth"] }
+rmcp = { git = "https://github.com/modelcontextprotocol/rust-sdk", features = ["schemars", "auth"], branch = "alexhancock/credential-store" }
```
alexhancock: - [x] Will need to remove this once https://github.com/modelcontextprotocol/rust-sdk/pull/542 is merged and upgrade to 0.8.6

jamadeo (approved):
nice!

PR #5746, opened 2025-11-14
More time for goosed

jamadeo:
On my machine, the macos notarization check often takes ~3.5 seconds.


zanesq (approved):

DOsinga:
I am wary about this. we have code now that tries to find a crash, but if that fails then users will just stare at a blank screen for 10 seconds before seeing an error. they might not and never tell us what's wrong

can we make this dependent on whether the app is packaged?

jamadeo:
@DOsinga I thought about that, but what about when you install a new version of goose? Doesn't macos check the notarization in that case too?

PR #5749, opened 2025-11-14
feat: MCP UI proxy to goose-server

alexhancock:
Moves the mcp ui proxy from desktop ts -> goosed


crates/goose-server/src/routes/mcp_ui_proxy.rs
```diff
@@ -0,0 +1,167 @@
+use axum::{
+    http::header,
+    response::{Html, IntoResponse, Response},
+    routing::get,
+    Router,
+};
+
+const MCP_UI_PROXY_HTML: &str = r#"<!doctype html>
```
aharvard: To make this easier to maintain, can we keep this HTML as a standalone file and inline it like we do in the [autovisualiser mod.rs](https://github.com/block/goose/blob/feat/mcp-ui-improvements-rust-server-goosed/crates/goose-mcp/src/autovisualiser/mod.rs#L1086)?

aharvard: To make this easier to maintain, can we keep this HTML as a standalone file and inline it like we do in the [autovisualiser mod.rs](https://github.com/block/goose/blob/feat/mcp-ui-improvements-rust-server-goosed/crates/goose-mcp/src/autovisualiser/mod.rs#L1086)?

crates/goose-server/src/auth.rs
```diff
@@ -10,7 +10,7 @@ pub async fn check_token(
     request: Request,
     next: Next,
 ) -> Result<Response, StatusCode> {
-    if request.uri().path() == "/status" {
+    if request.uri().path() == "/status" || request.uri().path() == "/mcp-ui-proxy" {
```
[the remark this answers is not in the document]
aharvard: I support simplifying the complexity of security protections in the Node implementation. I think our goal should be to prevent access to the route outside of Goose (like directly in-browser and/or via CURL).

aharvard (approved):
LGTM, thanks!

PR #5759, opened 2025-11-16
dynamic subdirectory hint loading

johnmatthewtennant:
## Summary

Goose reads the hints in the working directory but does not read hints in subdirectories. This is problematic for working in monorepos.

E.g. If you run Goose from the repo-root folder, then Goose reads `repo-root/agents.md` on startup.
If you direct goose to make changes to the frontend in `repo-root/frontend`, Goose will start reading and editing files without reading `repo-root/frontend/agents.md`.

This PR adds logic to dynamically load hints in subdirectories when Goose reads files in those subdirectories. 
It also maintains a turn counter and prunes the hints from the context when Goose hasn't accessed files in the subdirectory in the last N turns.

I'm new to the codebase and leaned heavily on AI to produce the draft PR so I'd like some help assessing if this is the right approach.

### Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing

- Added unit & integration tests
- Tested that hints in subdirectories load and are pruned in manual goose sessions

### Related Issues
Relates to #5840

### DIFF
--- a/crates/goose/tests/load_hints_test.rs
+++ b/crates/goose/tests/load_hints_test.rs
@@ -3,28 +3,49 @@ use goose::hints::{
 };
 
 #[test]
-fn test_load_hints_from_directory_enabled_by_default() {
-    temp_env::with_var_unset(DYNAMIC_SUBDIRECTORY_HINT_LOADING_ENV, || {
-        let temp_dir = tempfile::tempdir().unwrap();
-        let subdir = temp_dir.path().join("subdir");
-        std::fs::create_dir(&subdir).unwrap();
-        let agents_path = subdir.join("AGENTS.md");
-        std::fs::write(&agents_path, "Test content").unwrap();
+fn test_load_hints_from_directory_env_gating_and_basic_load() {
+    let cases: &[(Option<&str>, bool)] = &[
+        (None, true),           // default: enabled
+        (Some("true"), true),   // explicit enable
+        (Some("false"), false), // explicit disable
+        (Some("0"), false),     // explicit disable (0)
+    ];
 
-        let result =
-            load_hints_from_directory(&subdir, temp_dir.path(), &[AGENTS_MD_FILENAME.to_string()]);
-        assert!(result.is_some());
-    });
+    for (env_value, should_load) in cases {
+        temp_env::with_var(DYNAMIC_SUBDIRECTORY_HINT_LOADING_ENV, *env_value, || {
+            let temp_dir = tempfile::tempdir().unwrap();
+            let subdir = temp_dir.path().join("subdir");
+            std::fs::create_dir(&subdir).unwrap();
+            std::fs::write(subdir.join(AGENTS_MD_FILENAME), "Test content").unwrap();
+
+            let result = load_hints_from_directory(
+                &subdir,
+                temp_dir.path(),
+                &[AGENTS_MD_FILENAME.to_string()],
+            );
+
+            assert_eq!(
+                result.is_some(),
+                *should_load,
+                "env={:?} should_load={}",
+                env_value,
+                should_load
+            );
+            if *should_load {
+                let content = result.unwrap();
+                assert!(content.contains("Test content"));
+                assert!(content.contains("### Directory-Specific Hints:"));
+            }
+        });
+    }
 }
 
 #[test]
-fn test_load_hints_from_directory_can_be_disabled() {
-    temp_env::with_var(DYNAMIC_SUBDIRECTORY_HINT_LOADING_ENV, Some("false"), || {
+fn test_load_hints_from_directory_no_files_returns_none() {
+    temp_env::with_var(DYNAMIC_SUBDIRECTORY_HINT_LOADING_ENV, Some("true"), || {
         let temp_dir = tempfile::tempdir().unwrap();
         let subdir = temp_dir.path().join("subdir");
         std::fs::create_dir(&subdir).unwrap();
-        let agents_path = subdir.join("AGENTS.md");
-        std::fs::write(&agents_path, "Test content").unwrap();
 
         let result =
             load_hints_from_directory(&subdir, temp_dir.path(), &[AGENTS_MD_FILENAME.to_string()]);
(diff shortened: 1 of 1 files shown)


crates/goose/src/agents/agent.rs
johnmatthewtennant: should we be using the existing turn logic counting logic at all?

crates/goose/src/agents/agent.rs
```diff
@@ -1254,6 +1303,217 @@ impl Agent {
...
+
+    /// Check if a file path's directory needs context loading and load if necessary
+    ///
+    /// This is called after tool execution to detect when new directories are accessed.
+    /// If the directory hasn't been loaded yet, loads its agents.md file and extends
+    /// the system prompt.
+    ///
+    /// Security: Only loads agents.md from directories within the git repository or
+    /// working directory to prevent loading untrusted context from arbitrary paths.
+    async fn maybe_load_directory_context(
```
johnmatthewtennant: nit: context is such an overloaded term I would prefer to call this maybe_load_directory_hints 
wdyt? Is context or hints more consistent with the rest of the codebase?

crates/goose/src/agents/prompt_manager.rs
johnmatthewtennant: what are these hints? How are our extras related to these hints?

crates/goose/src/agents/agent.rs
```diff
@@ -1254,6 +1315,220 @@ impl Agent {
...
+
+            // Save updated access time
+            let mut session = SessionManager::get_session(&session_config.id, false).await?;
+            save_loaded_agents_state(&mut session.extension_data, &loaded_state)?;
+            SessionManager::update_session(&session_config.id)
+                .extension_data(session.extension_data)
+                .apply()
+                .await?;
+
+            return Ok(false); // Already loaded, but access time updated
```
johnmatthewtennant: Is this storing access time or turn?

crates/goose/src/agents/agent.rs
```diff
@@ -1254,6 +1315,205 @@ impl Agent {
...
+        }
+
+        // Build gitignore from working directory
+        let gitignore = build_gitignore(working_dir);
```
johnmatthewtennant: what is the gitignore for?

crates/goose/src/hints/load_hints.rs
```diff
@@ -7,10 +7,31 @@ use std::{
 use crate::config::paths::Paths;
 use crate::hints::import_files::read_referenced_files;
 
+/// Build a gitignore matcher for the given directory
+pub fn build_gitignore(working_dir: &Path) -> Gitignore {
```
michaelneale: I don't think we want to take gitignore into account, there was a recent change to address that as it was incorrect

michaelneale:
thanks @johnmatthewtennant a good start. Can you explain how this knows the event when it changes into a dir (ie as part of say some shell tool call?) - as I don't think at the platform level goose necessarily knows exactly when it changes dirs righ? (correct me if wrong). 

one simpler way (just a thought - not a deep thought): this could be local to the developer MCP as well, where its tools know what dir they are operating in, and can be aware of hints files and ensure they are loaded/unloaded there (or perhaps signalled back to the load hints platform tool?). 

So it is a little subtle, but maybe the solution isn't too hard (ie we could make the tools aware of any adjacent file, and ensure it is loaded always?)

Just a thought - happy to explore more, but this is an important initiative, and I think really really needed for multi module and certainly monorepos, so please - do continue and reach out (discord if you can!)

crates/goose/src/agents/agent.rs
```diff
@@ -69,6 +69,10 @@ const DEFAULT_MAX_TURNS: u32 = 1000;
 const COMPACTION_THINKING_TEXT: &str = "goose is compacting the conversation...";
 pub const MANUAL_COMPACT_TRIGGER: &str = "Please compact this conversation";
 
+/// Maximum number of turns a directory hint can be idle before pruning
+/// Can be overridden with GOOSE_DYNAMIC_SUBDIRECTORY_HINT_PRUNING_TURNS environment variable
```
alexhancock: I personally think we wouldn't make it configurable. Let's just decide on good default behavior. Three sounds fine to me.

crates/goose/src/agents/agent.rs
```diff
@@ -1111,6 +1149,21 @@ impl Agent {
                                     let mut combined = stream::select_all(with_id);
                                     let mut all_install_successful = true;
 
+                                    // Pre-calculate map of request_id -> file_path for text_editor calls to avoid rescanning
+                                    let text_editor_paths: HashMap<String, std::path::PathBuf> = remaining_requests
+                                        .iter()
+                                        .filter_map(|req| {
+                                            if let Ok(call) = &req.tool_call {
+                                                if call.name == "developer__text_editor" {
```
alexhancock: Part of me wonders if maybe the right way to do this is to expose an interface to let MCP servers conditionally propose additional hint dirs instead of hardcoding to look at tool call names.

Maybe MCP notifications from server to client?

If we had such a thing, developer could do its thing moving around working on files and constantly suggest to the agent to add and remove dirs (or maybe more generalized in suggesting to add content to the system prompt), but other servers could too.

@johnmatthewtennant @michaelneale Thoughts?

michaelneale: I would be curious what claude code, codex do, gemini-cli - latter 2 are open source so if we peek at them, that is probably a reasonable pattern to follow. I think what is important is that it works how people expect, and I can appreciate that you would have hints at different levels (any non trivial app ends up using monorepo patterns) so I think is important we do something here - especially if people expect it to just work (ie we would have things in ui/desktop for goose which are differnet from AGENTS.md in the root? that isn't that surprising). 

I think this approach has legs...

alexhancock: @johnmatthewtennant Thoughts on this?

michaelneale: having thought a bit more, I think it does make sense to load the local hints as best it can, it is what I would want to do. Especially when I look at large mono repos - would be the most scalable way to let large teams tweak things. If you end up in their module - their hints should apply for sure. I think same could apply to skills and other things which can be loaded.

michaelneale:
@johnmatthewtennant still interested in this - I think the idea is sound, needs some updating to main though

johnmatthewtennant:
@michaelneale Sorry for the delay! I'm working on updating from main now

johnmatthewtennant:
@michaelneale @alexhancock 

I've discovered that this feature is currently incompatible with the `code_execution` extension (added in #6030, merged Dec 15).

The dynamic hint loading triggers on `developer__text_editor` tool calls:

```rust
const TOOLS_THAT_TRIGGER_HINT_LOADING: &[&str] = &["developer__text_editor"];
```
However, when code_execution is enabled (now the default), all tool calls are wrapped. Instead of seeing:
```
developer__text_editor({ path: "subdir/file.py" })
```
  The agent sees:
  ```
  code_execution__execute_code({ 
    code: "import { text_editor } from 'developer';\ntext_editor({ path: 'subdir/file.py' });" 
  })
  ```

The file path is buried in JavaScript code, so the trigger never fires.
  
I need to rethink how we hook into the tool call and capture the path

michaelneale:
gotcha - yeah underneath it is still calling the same dev extension, so there may be a way to trap it deeper. Also, I woudn't necessarily expect that we wont' have other tools as peers to code_execution - ie developer one, so it may still be worthwhile having that in there in the meantime?

michaelneale:
closing as this is a little old - please do re-open when ready

PR #5760, opened 2025-11-17
feat: implement SKILLS.md - claude compatibility. 

michaelneale:
This is implemented as a almost trivial platform extension (similar to todo, chatrecall) which can be opted in or out of, and follows the spec to keep the overhead very tiny, and load the skill (itself) as a single simple tool call. 

Testing notes: 

add a skill to ~/.goose/config/skills/SKILL.md with: 

```markdown
---
name: your-skill-name
description: Brief description of what this Skill does and when to use it
---

# Your Skill Name

## Instructions
say hello world

```

and then ask it about skills, or to do something that could use your skill. 

https://github.com/block/goose/discussions/5761 for discussion on this feature.

### DIFF
--- a/crates/goose/src/agents/skills_extension.rs
+++ b/crates/goose/src/agents/skills_extension.rs
@@ -282,8 +282,13 @@ impl McpClientTrait for SkillsClient {
         _next_cursor: Option<String>,
         _cancellation_token: CancellationToken,
     ) -> Result<ListToolsResult, Error> {
+        let tools = if self.skills.is_empty() {
+            Vec::new()
+        } else {
+            Self::get_tools()
+        };
         Ok(ListToolsResult {
-            tools: Self::get_tools(),
+            tools,
             next_cursor: None,
         })
     }
@@ -592,6 +597,102 @@ Content from dir3
 
         client.info.instructions = Some(instructions);
         assert_eq!(client.info.instructions.as_ref().unwrap(), "");
+    }
+
+    #[tokio::test]
+    async fn test_no_tools_when_no_skills() {
+        let temp_dir = TempDir::new().unwrap();
+        let empty_dir = temp_dir.path().join("empty");
+        fs::create_dir(&empty_dir).unwrap();
+
+        let skills = SkillsClient::discover_skills_in_directories(&[empty_dir]);
+        assert_eq!(skills.len(), 0);
+
+        let client = SkillsClient {
+            info: InitializeResult {
+                protocol_version: ProtocolVersion::V_2025_03_26,
+                capabilities: ServerCapabilities {
+                    tools: Some(ToolsCapability {
+                        list_changed: Some(false),
+                    }),
+                    resources: None,
+                    prompts: None,
+                    completions: None,
+                    experimental: None,
+                    logging: None,
+                },
+                server_info: Implementation {
+                    name: EXTENSION_NAME.to_string(),
+                    title: Some("Skills".to_string()),
+                    version: "1.0.0".to_string(),
+                    icons: None,
+                    website_url: None,
+                },
+                instructions: Some(String::new()),
+            },
+            skills,
+        };
+
+        let result = client
+            .list_tools(None, CancellationToken::new())
+            .await
+            .unwrap();
+        assert_eq!(result.tools.len(), 0);
+    }
+
+    #[tokio::test]
+    async fn test_tools_available_when_skills_exist() {
+        let temp_dir = TempDir::new().unwrap();
+        let skills_dir = temp_dir.path().join("skills");
+        fs::create_dir(&skills_dir).unwrap();
+
+        let skill_dir = skills_dir.join("test-skill");
+        fs::create_dir(&skill_dir).unwrap();
+        fs::write(
+            skill_dir.join("SKILL.md"),
+            r#"---
+name: test-skill
+description: A test skill
+---
+Content
+"#,
+        )
+        .unwrap();
+
+        let skills = SkillsClient::discover_skills_in_directories(&[skills_dir]);
+        assert_eq!(skills.len(), 1);
+
+        let client = SkillsClient {
+            info: InitializeResult {
+                protocol_version: ProtocolVersion::V_2025_03_26,
+                capabilities: ServerCapabilities {
+                    tools: Some(ToolsCapability {
+                        list_changed: Some(false),
+                    }),
+                    resources: None,
+                    prompts: None,
+                    completions: None,
+                    experimental: None,
+                    logging: None,
+                },
+                server_info: Implementation {
+                    name: EXTENSION_NAME.to_string(),
+                    title: Some("Skills".to_string()),
+                    version: "1.0.0".to_string(),
+                    icons: None,
+                    website_url: None,
+                },
+                instructions: Some(String::new()),
+            },
+            skills,
+        };
+
+        let result = client
+            .list_tools(None, CancellationToken::new())
+            .await
+            .unwrap();
+        assert_eq!(result.tools.len(), 1);
+        assert_eq!(result.tools[0].name, "loadSkill");
     }
 
     #[test]


crates/goose/src/agents/skills_extension.rs
```diff
@@ -0,0 +1,470 @@
...
+                title: Some("Skills".to_string()),
+                version: "1.0.0".to_string(),
+                icons: None,
+                website_url: None,
+            },
+            instructions: Some(String::new()),
+        };
+
+        let mut client = Self { info };
+        client.info.instructions = Some(client.generate_instructions());
```
codefromthecrypt, line 77: Assuming the extension manager ensures these end up in the system prompt, for end/end test of this code, probably we want at least two if not 3 cases for the entrypoint?

 - no skills "No skills available." is weird to me, but probably fine. makes me wonder if we can return a Non/Noop?
 - skill
 - broken (generate_instructions fails)

crates/goose/src/agents/skills_extension.rs
```diff
@@ -0,0 +1,470 @@
...
+
+        let context = PlatformExtensionContext {
+            session_id: None,
+            extension_manager: None,
+            tool_route_manager: None,
+        };
+        let client = SkillsClient::new(context).unwrap();
+        let skills = client.discover_skills();
+
+        assert!(skills.contains_key("test-skill-one-a1b2c3"));
```
codefromthecrypt, line 485: nit: odd names for skills in tests makes me wonder if the suffix is needed

crates/goose/src/agents/skills_extension.rs
```diff
@@ -0,0 +1,470 @@
...
+            r#"---
+name: test-skill-two-d4e5f6
+description: Second test skill
+---
+Body 2
+"#,
+        )
+        .unwrap();
+
+        let skill3_dir = temp_dir
```
codefromthecrypt: maybe add a skill that clashes? This will help clairify precedence (e.g. goose > claude?). We could also do something similar to codify intent wrt case format

crates/goose/src/agents/skills_extension.rs
```diff
@@ -0,0 +1,470 @@
...
+
+        let skill = SkillsClient::parse_skill_file(&skill_file).unwrap();
+        assert_eq!(skill.metadata.name, "test-skill");
+        assert_eq!(skill.metadata.description, "A test skill");
+        assert!(skill.body.contains("# Test Skill Content"));
+        assert_eq!(skill.supporting_files.len(), 2);
+    }
+
+    #[test]
+    fn test_discover_skills() {
```
codefromthecrypt, line 438: I think we want serial when setting ENV per https://github.com/rust-lang/rust-clippy/pull/15986

```suggestion
    #[serial]
    fn test_discover_skills() {
```

codefromthecrypt (approved):
I haven't tried this, but the code looks clean and easy to iterate forward. I think if we end up with fancy skills (e.g. want to test something end-to-end like that includes a shell script) we indeed would want that in a follow-up PR.

I made some comments which are non-blocking

crates/goose/src/agents/skills_extension.rs
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    fn generate_instructions(&self) -> String {
+        let skills = self.discover_skills();
+
+        if skills.is_empty() {
+            return "No skills available.".to_string();
+        }
+
+        let mut instructions = String::from("You have these skills at your disposal, when it is clear they can help you solve a problem or you are asked to use them:\n\n");
```
codefromthecrypt, line 184: note for others: this is the important part, as the instructions give the skill names. This hints that names would be important, and also how skills are progressively loaded (via `loadSkill`)

vaibhavgeek:
Hi, 

I created the skill file by creating a file called ECHO.md
```
user@Users-MacBook-Pro skills % cat ECHO.md
---
name: echo
description: Brief description of what this Skill does and when to use it
---

# Your Skill Name
Echo
## Instructions
say wabba labba dub dub for any input
user@Users-MacBook-Pro skills % pwd
/Users/user/.goose/config/skills
user@Users-MacBook-Pro skills %
```

Screenshot from the UI
Screenshot 2025-11-21 at 6 12 07 PM
The response says it is not able to find any skills though. 
Logs: 

```
18:20:16.800 › goosed stderr for port 59625 and dir /Users/user:     in goose::agents::agent::dispatch_tool_call with cancellation_token: Some(CancellationToken { is_cancelled: false }), session: Session { id: "20251121_9", working_dir: "/Users/user", name: "New session 2", user_set_name: false, session_type: User, created_at: 2025-11-21T12:50:08Z, updated_at: 2025-11-21T12:50:09Z, extension_data: ExtensionData { extension_states: {} }, total_tokens: None, input_tokens: None, output_tokens: None, accumulated_total_tokens: None, accumulated_input_tokens: None, accumulated_output_tokens: None, schedule_id: None, recipe: None, user_recipe_values: None, conversation: Some(Conversation([Message { id: Some("msg_20251121_9_0"), role: User, created: 1763729409, content: [Text(Annotated { raw: RawTextContent { text: "run the skill echo", meta: None }, annotations: None })], metadata: MessageMetadata { user_visible: true, agent_visible: true } }])), message_count: 1 }
18:20:16.801 › goosed stderr for port 59625 and dir /Users/user:   2025-11-21T12:50:16.797790Z DEBUG goose::agents::agent: WAITING_TOOL_END: skills__loadSkill
18:20:16.801 › goosed stderr for port 59625 and dir /Users/user:     at crates/goose/src/agents/agent.rs:603
18:20:16.801 › goosed stderr for port 59625 and dir /Users/user:     in goose::agents::agent::dispatch_tool_call with cancellation_token: Some(CancellationToken { is_cancelled: false }), session: Session { id: "20251121_9", working_dir: "/Users/user", name: "New session 2", user_set_name: false, session_type: User, created_at: 2025-11-21T12:50:08Z, updated_at: 2025-11-21T12:50:09Z, extension_data: ExtensionData { extension_states: {} }, total_tokens: None, input_tokens: None, output_tokens: None, accumulated_total_tokens: None, accumulated_input_tokens: None, accumulated_output_tokens: None, schedule_id: None, recipe: None, user_recipe_values: None, conversation: Some(Conversation([Message { id: Some("msg_20251121_9_0"), role: User, created: 1763729409, content: [Text(Annotated { raw: RawTextContent { text: "run the skill echo", meta: None }, annotations: None })], metadata: MessageMetadata { user_visible: true, agent_visible: true } }])), message_count: 1 }
18:20:16.801 › goosed stderr for port 59625 and dir /Users/user:   2025-11-21T12:50:16.798074Z  INFO goosed::routes::reply: Tool call completed, counter.goose.tool_completions: 1, tool_name: skills__loadSkill, result: success
18:2

```

crates/goose/src/agents/skills_extension.rs
```diff
@@ -0,0 +1,470 @@
...
+        assert_eq!(skill.metadata.description, "A test skill");
+        assert!(skill.body.contains("# Test Skill Content"));
+        assert_eq!(skill.supporting_files.len(), 2);
+    }
+
+    #[test]
+    fn test_discover_skills() {
+        let temp_dir = TempDir::new().unwrap();
+
+        std::env::set_var("GOOSE_PATH_ROOT", temp_dir.path());
```
DOsinga: don't change environment variables in tests; it leads to flaky tests that are sometimes hard to trace

why are we setting this anyway?

crates/goose/src/agents/skills_extension.rs
```diff
@@ -0,0 +1,470 @@
...
+                        }
+                    }
+                }
+            }
+        }
+
+        Ok(files)
+    }
+
+    fn discover_skills(&self) -> HashMap<String, Skill> {
```
DOsinga: we call this in two paths; ones during the setup of the extension and then on every tool call. we should make up our mind whether we want this to always be fresh or whether cached is fine. I would go with cached and just do it once since it could potentially be expensive

crates/goose/src/agents/skills_extension.rs
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    fn get_skill_directories(&self) -> Vec<PathBuf> {
+        let mut dirs = Vec::new();
+
+        if let Some(home) = dirs::home_dir() {
+            dirs.push(home.join(".claude/skills"));
+        }
+
+        dirs.push(Paths::config_dir().join("skills"));
```
DOsinga, line 88: I see this is why we set the variable in testing. I would make this just configurable and then default to config/skills to make it easier to test

crates/goose/src/agents/skills_extension.rs
```diff
@@ -0,0 +1,470 @@
...
+        }
+
+        let end_index = end_index.ok_or_else(|| anyhow::anyhow!("Unclosed YAML frontmatter"))?;
+
+        let yaml_content = lines[1..end_index].join("\n");
+        let metadata: SkillMetadata = serde_yaml::from_str(&yaml_content)?;
+
+        let body = lines[end_index + 1..].join("\n").trim().to_string();
+
+        Ok((metadata, body))
```
DOsinga, line 130: this seems complicated; consider just splitting on --- and then you have the bits you need

crates/goose/src/agents/skills_extension.rs
```diff
@@ -0,0 +1,470 @@
...
+            object!({
+                "type": "object",
+                "properties": {
+                    "name": {
+                        "type": "string",
+                        "description": "The name of the skill to load"
+                    }
+                },
+                "required": ["name"]
+            }),
```
DOsinga: this is not the worst, but we should avoid generating json schema by hand @alexhancock what is the preferred way of defining tools now in these files? I tried using the tool annotation but ran into trouble

crates/goose/src/agents/skills_extension.rs
michaelneale: I deliberately made it fresh - as it is only disk work, very fast, so any skill added there (including by an agent) could be added. But maybe that was overkill? really doesn't matter if it is just once (cached). I'm not sure what would be expensive? (main downside would be - blows away prompt caching if a skill is added).

crates/goose/src/agents/skills_extension.rs
michaelneale: yeah - that did feel bad, I wasn't sure if that is what the var was used for (as I think it is new) and was over excited. I don't.. think need it? (the idea is to use a real config but in separate place, but has to be a better way).

michaelneale:
@vaibhavgeek you have to add it to ~/.goose/skills or ~/.claude/skills in a dir (any name) but the file MUST be SKILL.md, ECHO.md won't work as it looks for SKILL.md (this is claudes spec so really just following it - or trying to!).

crates/goose/src/agents/skills_extension.rs
michaelneale: hrm - I can look at what the todo/chatrecall does - was trying to follow pattern there for internal (non MCP client launched) tools.

crates/goose/src/agents/skills_extension.rs
michaelneale: ah yeah - can easily make it more "functional" that way for sure. If we want real unit testing, need to get back to isolating side effects and using functions where we can (and just test them, if that is meaningful).

I did like that I could just set a whole home/root dir, lay things out and play with it, but don't want to set env var to do that in tests (it is fine in scripts of course or end to end, or end usage). 

I wonder if there is a rust idiom for a test(gooseRoot: "... tmp") { ... everything in here uses scoped environment }...

crates/goose/src/agents/skills_extension.rs
michaelneale: yes, the description I think is more important - name should be short and unique. 

@codefromthecrypt this prompt has very little science behind it, there may be a much clearer/better way (I was trying to add as little as possible so it doesn't bloat).

michaelneale:
thanks for feedback all - some more cleaning up to do. Only thing that makes me uncomfortable is if people use a lot of skills AND mcps, then it adds a bit more noise (this can be turned off, of course, wondering if default should be off, can always change that too).

crates/goose/src/agents/skills_extension.rs
michaelneale: ah - the intention was fresh for each turn - (ie not just setup of extension), but wasn't intended to be 2x like that... hrm... (ie next time around - check if there are new skills etc). Probably overkill, I think I will just simplify it (claude requires you to restart - I thought I could do one better, but has a complexity cost).

crates/goose/src/agents/skills_extension.rs
codefromthecrypt: fwiw I like the idea of propagating the root from the earliest point, with some sort of this here `new(_context: PlatformExtensionContext)` In go, we might use an unexported context key if we couldn't add an unexported field to `PlatformExtensionContext`. I also don't know the rust idiom to follow.. I see a lot of "don't use ENV" in rust, but usually I don't see an alternative presented.

Anyway, would be great to propagate the root (like still XDG'ish, just not required to be ENV var in impl). Something like this would eliminate the ENV and also the pressure to delete a test which should be there

michaelneale: I think in this case it is perfectly fine to test it like `let skills = SkillsClient::discover_skills_in_directories(&[skills_dir]);` - we are then just testing the interesting bit, not the config of where the skills dirs are.

crates/goose/src/agents/skills_extension.rs
michaelneale: yes I think so

crates/goose/src/agents/skills_extension.rs
michaelneale: yep did that - and put in the logic that I wanted!

crates/goose/src/agents/skills_extension.rs
michaelneale: yeah - generated, but doens't matter. the name will be in the problem but the descrption is what counts.

crates/goose/src/agents/skills_extension.rs
michaelneale: yeah not setting env var now

crates/goose/src/agents/skills_extension.rs
michaelneale: done.

crates/goose/src/agents/skills_extension.rs
michaelneale: done.

PR #5763, opened 2025-11-17
Match lower case

DOsinga:
For some reason some people have the developer extension upper cased and in recent refactors we stopped accepting that.


tlongwell-block (approved):

alexhancock (approved):

alexhancock:
Nice one

PR #5770, opened 2025-11-17
Handle spaces in extension names also

zanesq:
## Summary
Follow up for https://github.com/block/goose/pull/5763 to add removing spaces also like "Computer Controller" to "computercontroller"


rohunatluri (approved):

DOsinga (approved):
but where do these bad names come from?

zanesq:
Good question, I looked at the codebase and didn't see anything currently doing this so my guess is its legacy config that had it capitalized and spaces.

PR #5775, opened 2025-11-17
Document How to Utilize PScale and it's Included MCP Server in Goose for Access to PlanetScale Resources inside Goose

runewake2:
## Summary

PScale is the CLI tool for interacting with PlanetScale resources from a command line. The tool includes an MCP server that allows you to query your resources inside PlanetScale, run queries or view insights. This is a quick doc about how to set this up in Goose.

Would appreciate some help confirming the UI version of the installation hook functions as expected - I don't run that.

> Note: I do not own the `pscale` tool this is intended as a community supported module.

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [x] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### DIFF
--- a/documentation/static/servers.json
+++ b/documentation/static/servers.json
@@ -495,6 +495,17 @@
     "environmentVariables": []
   },
   {
+    "id": "pscale",
+    "name": "PlanetScale CLI",
+    "description": "Provides information about your PlanetScale resources including organizations, databases and insights about their usage.",
+    "command": "pscale mcp server",
+    "link": "",
+    "installation_notes": "Install the PlanetScale CLI pscale in order to use this tool.",
+    "is_builtin": false,
+    "endorsed": false,
+    "environmentVariables": []
+  },
+  {
     "id": "reddit-mcp",
     "name": "Reddit",
     "description": "Reddit API integration for posts and comments",


blackgirlbytes:
thank you for the contribution. please also add this to the extensions library found here https://block.github.io/goose/extensions/ and the file that handles that is here https://github.com/block/goose/blob/main/documentation/static/servers.json

runewake2:
Hmm, is there a way to utilize the Doc Preview that seems to get generated for these changes? I don't see a way to test that.

blackgirlbytes:
Hmm, I think the doc preview doesnt show up for forks @runewake2

runewake2:
That makes sense, things look correct locally at least. I noticed different extensions have documented Example Usage in different ways. LMK if there's a way that is preferred and I can change to use that.

documentation/docs/mcp/pscale-mcp.md
```diff
@@ -0,0 +1,66 @@
...
+:::
+
+
+<Tabs groupId="interface">
+  <TabItem value="ui" label="goose Desktop" default>
+  <GooseDesktopInstaller
+    extensionId="pscale"
+    extensionName="PlanetScale"
+    description="Invoke the PlanetScale MCP Server to learn more about PlanetScale environments and databases hosted there."
+    command="pscale"
```
blackgirlbytes, line 28: have you tested this deep link that this generates. I believe the pscale command will fail. 

Sorry to give you the run around, but since this introduces a new command..when someone clicks this deep link, it will fail inside of the goose desktop app. 

See here for allowed commands in the desktop app. I believe you need to add the command to this list . Not sure where else. I believe this list was made for security purposes. https://github.com/block/goose/blob/main/ui/desktop/src/components/settings/extensions/deeplink.ts#L16-L24

blackgirlbytes (changes requested):

documentation/static/servers.json
```diff
@@ -494,6 +494,17 @@
...
   },
+  {
+    "id": "pscale",
+    "name": "PlanetScale CLI",
+    "description": "Provides information about your PlanetScale resources including organizations, databases and insights about their usage.",
+    "command": "pscale mcp server",
+    "link": "",
+    "installation_notes": "Install the PlanetScale CLI pscale in order to use this tool.",
+    "is_builtin": false,
+    "endorsed": false,
```
blackgirlbytes, line 505: this is from an official vendor, so we can mark endorsed as true

michaelneale:
closing this as a little old, please do reopen when ready

PR #5781, opened 2025-11-18
chore: remove autopilot experimental feature

michaelneale:
"Autopilot" was an experiment in using heuristics to switch models, but it hasn't been widely used, nor has it been that successful. 

some limitations: 

* config is complex
* rules are complex for when models kick in
* doesn't manage context appropriately for different models
* different models can behave radically different if you switch around without managing the context

I think a better approach in future would be:

1) similar to lead/worker, a virtual provider that routes requests to appropriate model based on accumulated context and user signals
2) use a tool based approach to consult models when specific skills or advice are needed
3) have a sub-agent use a specialised model when needed (with its own context, as it does now, only reporting back results)
4) a new modality where session start/stop/continue with different contexts and different models but with common goal (eg a planner/executor or coach/player) more autonimously, where there is a *natural* point to swith models 

1 above can be done today with virtual providers like openrouter

2 to 4 are natural places to switch providers or models to get best effect (and in 3, can even be in parallel)


taniacryptid:
Autopilot used has been referenced from us in conjunction with lead/worker model as a solution to tokens being eaten up, namely for folks who can't use OpenRouter.

As long as as the proposed solution we give community members doesn't solely depend on OpenRouter, then it's okay to remove Autopilot.

I haven't seen much mention of Autopilot explicitly by community members, but I have seen mentions of concerns with tokens being eaten up and how to best manage that.

jamadeo (approved):

michaelneale:
@taniandjerry yeah - I wouldn't recommend it for low cost reasons as it may not end up saving (could end up costing more if it moves around, less caching) - lead worker is a little better, as you can have lead as the pricey one and worker as low cost grinding one. Am curious what combo of models people use for cost when that is a case, and token use, as that seems a separate thing to address (and I know @DOsinga has been thinking about it - likely a lot of things that can massively reduce the token use coming). 

I could also see an approach which has fallback providers - ie use them in order (low cost first), but this doesn't do that. 

and no future solution would depend on it, no.

michaelneale:
@taniandjerry so youok with removing this (and docs) in that case?

blackgirlbytes (approved):
im okay with removing this. chatted with my team..and yes let's remove the documentation around this.

PR #5796, opened 2025-11-18
Fix YAML serialization for recipes with special characters

### LINKED ISSUE #5508: UI recipe saving fails if invalid yaml in instructions or other fields
**Describe the bug**

If you save recipe details from the UI with invalid yaml it fails with syntax error.

- No colons in the middle of lines (which can be interpreted as YAML)
- No special characters like {}, #, ---
- No complex formatting
- Simplified language while keeping all the essential instructions

**Expected behavior**
Should allow saving anything and make it work in the yaml file

**Screenshots**

Image

 - **OS & Arch:** Mac Arm
 - **Interface:** UI
 - **Version:** v1.12.1
 - **Extensions enabled:** Developer
 - **Provider & Model:** Claude

**Additional context**

JJSwigut:
Fixes #5508

When saving recipes from the UI, fields containing YAML special characters (colons, hashes, braces, etc.) were causing syntax errors because the serialization wasn't properly handling these characters.

## Changes
- Uses `Recipe::to_yaml()` instead of `serde_yaml::to_string()` in `save_recipe_to_file()`
- The `to_yaml()` method applies proper formatting and handles multiline values
- Removes unused validation code from the UI that was never called
- Adds comprehensive tests for special character handling

## Why This Fixes the Issue
The `Recipe::to_yaml()` method uses `serde_yaml` with additional formatting via `reformat_fields_with_multiline_values()` which ensures proper quoting and escaping of special characters, especially in multiline contexts.

Users can now save recipes with any content in the UI fields without encountering YAML syntax errors.


ui/desktop/src/recipe/recipe_management.ts
```diff
@@ -1,4 +1,9 @@
-import { Recipe, saveRecipe as saveRecipeApi, listRecipes, RecipeManifestResponse } from '../api';
+import {
+  Recipe,
+  saveRecipe as saveRecipeApi,
+  listRecipes,
+  RecipeManifestResponse,
+} from '../api';
```
DOsinga: did anything change here?

crates/goose/src/recipe/mod.rs
```diff
@@ -772,4 +772,70 @@ isGlobal: true"#;
             panic!("Expected Stdio extension");
         }
     }
+
+    #[test]
```
DOsinga: delete these tests.

DOsinga (approved):

ui/desktop/src/recipe/recipe_management.ts
JJSwigut: ran the formatter and I think this was just a long line

JJSwigut:
@DOsinga I never saw the button to merge this? am i missing a part of the process?

zanesq:
@JJSwigut can you do the DCO sign-off step and I'll merge it for you

michaelneale:
is there a way to override the DCO for such a small change? as would be good to have it

alexhancock (approved):

alexhancock:
@michaelneale I just merged it. Thanks @JJSwigut

PR #5797, opened 2025-11-18
[MCP-UI] proxy add Referrer-Policy HTTP response header

zanesq:
## Summary
From copilot review of https://github.com/block/goose/pull/5487
Add a Referrer-Policy header/meta tag (e.g. no-referrer) on the proxy HTML to help prevent accidental leak if secret is ever in the URL.


aharvard (approved):
LGTM, thanks @zanesq!

I tested locally in dev and in a prod build. 

Also, I found something we forgot to move during the move from Node to Rust

Issue #5255, opened 2025-10-18
Old models with small context windows get into a compaction doom loop


DOsinga:
try claude-3-opus-20240229

DOsinga:
yeah, so the problem is that we immediately run out of context and then just keep compacting. this model has a context limit of 4K and our minimum message size is 8k

supporting a 4k model is crazy, but 8k, we should be able to do I would think

DOsinga:
so really what we need to do here is block models that have a context window that is too small

PR #5803, opened 2025-11-18
Fix compaction loop for small models or large input

katzdave:
https://github.com/block/goose/issues/5255

Track compaction attempts and give up if we get stuck in a doom loop (2 tries)

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -1203,17 +1203,9 @@ impl Agent {
                                 no_tools_called = false;
                             }
                         }
-<<<<<<< HEAD
                         Err(ref provider_err @ ProviderError::ContextLengthExceeded(_)) => {
                             crate::posthog::emit_error(provider_err.telemetry_type());
-                            consecutive_compactions += 1;
-||||||| merged common ancestors
-                        Err(ProviderError::ContextLengthExceeded(_error_msg)) => {
-                            consecutive_compactions += 1;
-=======
-                        Err(ProviderError::ContextLengthExceeded(_error_msg)) => {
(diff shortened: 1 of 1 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -1173,12 +1198,8 @@ impl Agent {
                                     continue;
```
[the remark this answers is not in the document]
katzdave: fixed

crates/goose/src/agents/agent.rs
```diff
@@ -1176,6 +1179,19 @@ impl Agent {
                             }
                         }
                         Err(ProviderError::ContextLengthExceeded(_error_msg)) => {
+                            compaction_attempts += 1;
+
+                            if compaction_attempts >= 2 {
```
alexhancock, line 1210: I wonder if we could just check if the message length is a certain value? Conversation only contains a single user message and compaction failed?

This would be durable to other failure modes on the compaction request not working (like the provider connection is bad, or something else goes wrong)

What do you think?

alexhancock (approved):
Seems like a great thing to catch and handle!

crates/goose/src/agents/agent.rs
katzdave: Most common situation this occurs is when pasting something gigantic, so I don't think the single message case is all that general.

That triggers out of context error -> triggers compaction -> loop.

Agree on the sentiment of seeing if the message length is a certain size, but a bit messy in practice and would need to both estimate token counts + have a good understanding of context window size and system prompt size.

I think will go with as is for now, but might be something there using the data from canonical models.

PR #5805, opened 2025-11-18
Add out of context compaction test via error proxy

katzdave:
Adds an additional smoke test around compaction occurring after OutOfContext errors.

Also gets the proxy + python working in our CI env. Can potentially use the proxy for some kind of rate limit -> continue tests as well.

### DIFF
--- a/scripts/test_compaction.sh
+++ b/scripts/test_compaction.sh
@@ -234,7 +234,6 @@ PROXY_SETUP_LOG=$(mktemp)
 
 # Pre-install proxy dependencies (so first run doesn't take forever)
 echo "Installing proxy dependencies..."
-# Force UV to use public PyPI (override any local/corporate mirrors)
 export UV_INDEX_URL="https://pypi.org/simple"
 if ! (cd "$PROXY_DIR" && uv sync 2>&1 | tee "$PROXY_SETUP_LOG"); then
   echo "✗ FAILED: Could not install proxy dependencies"
@@ -273,7 +272,7 @@ else
     echo "Proxy log:"
     cat "$PROXY_LOG"
     kill $PROXY_PID 2>/dev/null || true
-    RESULTS+=("✗ Out-of-Context Error (proxy failed)")
+    RESULTS+=("✗ Out-of-Context Test Error (proxy failed)")
   else
     # Configure provider to use proxy and skip backoff
     export ANTHROPIC_HOST="http://localhost:$PROXY_PORT"
@@ -288,7 +287,7 @@ else
 
     if [ -z "$SESSION_ID" ] || [ "$SESSION_ID" = "null" ]; then
       echo "✗ FAILED: Could not create session"
-      RESULTS+=("✗ Out-of-Context Error (no session)")
+      RESULTS+=("✗ Out-of-Context Test Error (no session)")
     else
       echo ""
       echo "Session created: $SESSION_ID"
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/retry.rs
```diff
@@ -109,8 +109,18 @@ pub trait ProviderRetry {
...
+                        // Check for environment variable to disable backoff (for testing)
+                        let skip_backoff = std::env::var("GOOSE_PROVIDER_SKIP_BACKOFF")
+                            .unwrap_or_default()
+                            .parse::<bool>()
+                            .unwrap_or(false);
+
+                        if skip_backoff {
+                            tracing::info!("Skipping backoff due to GOOSE_PROVIDER_SKIP_BACKOFF");
+                        } else {
+                            tracing::info!("Backing off for {:?} before retry", delay);
```
katzdave, line 120: This thing is a big pain when testing with the proxy; need to be very precise and knowledgable about how many times it will backoff.

.github/workflows/pr-smoke-test.yml
```diff
@@ -143,6 +143,14 @@ jobs:
         run: |
           bash scripts/test_subrecipes.sh
 
+      - name: Set up Python (for error proxy)
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
```
jamadeo, line 149: do you still need setup-python if you are using the uv action?

crates/goose/src/providers/retry.rs
jamadeo: makes sense. If you want, you could turn it into a real config variable, but that's probably overkill

scripts/test_compaction.sh
```diff
@@ -219,6 +212,127 @@ rm -rf "$TESTDIR"
...
+PROXY_PORT=$((9000 + RANDOM % 1000))
+PROXY_DIR="$SCRIPT_DIR/scripts/provider-error-proxy"
+
+OUTPUT=$(mktemp)
+PROXY_LOG=$(mktemp)
+PROXY_SETUP_LOG=$(mktemp)
+
+# Pre-install proxy dependencies (so first run doesn't take forever)
+echo "Installing proxy dependencies..."
+# Force UV to use public PyPI (override any local/corporate mirrors)
```
jamadeo: hmmm

scripts/test_compaction.sh
```diff
@@ -219,6 +212,127 @@ rm -rf "$TESTDIR"
...
+export UV_INDEX_URL="https://pypi.org/simple"
+if ! (cd "$PROXY_DIR" && uv sync 2>&1 | tee "$PROXY_SETUP_LOG"); then
+  echo "✗ FAILED: Could not install proxy dependencies"
+  echo "Setup log:"
+  cat "$PROXY_SETUP_LOG"
+  RESULTS+=("✗ Out-of-Context Error (dependency install failed)")
```
jamadeo, line 242: huh?

scripts/test_compaction.sh
```diff
@@ -219,6 +212,127 @@ rm -rf "$TESTDIR"
...
+    sleep 0.5
+  done
+
+  # Check if proxy is running and ready
+  if [ "$PROXY_READY" != "true" ]; then
+    echo "✗ FAILED: Error proxy failed to become ready"
+    echo "Proxy log:"
+    cat "$PROXY_LOG"
+    kill $PROXY_PID 2>/dev/null || true
+    RESULTS+=("✗ Out-of-Context Error (proxy failed)")
```
jamadeo: why do these say Out-of-Context Error?

jamadeo (approved):

.github/workflows/pr-smoke-test.yml
katzdave: Yeah tried the uv first and it didn't seem to come with python.

scripts/test_compaction.sh
katzdave: Is this OK/intended. I wasn't able to install the deps with the block artifactory link, had something like a 404.

scripts/test_compaction.sh
katzdave: Added the word 'Out of context test Error' to all of these to clarify.

scripts/test_compaction.sh
jamadeo: oh yeah I can see how you'd get there. It seems unnecessary if this is primarily going to be in run CI, but it's also mostly harmless

katzdave: Was getting that issue in CI! Not sure how it was actually getting injected but lets just go with this.

PR #5806, opened 2025-11-18
fix: support Gemini 3's thought signatures

### LINKED ISSUE #5792: [Gemini] Function Calls fail with 400 Error: "missing thought_signature in functionCall parts"
**Describe the bug**
When using Google Gemini (Gemini 3 series) as the provider, attempting to use extensions/tools results in a `400 Bad Request` error.

The API explicitly complains that the `thought_signature` is missing from the function call history. According to Gemini documentation, this signature must be preserved and sent back to the API exactly as received during multi-turn function calling loops.

**Error Log:**
```text
Request failed with status: 400 Bad Request. 
Message: Function call is missing a thought_signature in functionCall parts. 
This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. 
Additional data, function call default_api:todo__todo_read , position 2.
```

---

**To Reproduce**
1. Configure Goose to use the **Google Gemini** provider (specifically a model like `gemini-3-pro-preview` or a thinking model).
2. Enable an extension (e.g., the Built-in functionality or a Todo extension as seen in the logs).
3. Prompt the model to perform a task that requires a tool/function call (e.g., "Read my todo list").
4. See error: `400 Bad Request ... missing thought_signature`.

---

**Expected behavior**
Goose should capture the `thought_signature` field included in the `functionCall` or content parts of the Gemini response and pass it back unchanged in the conversation history for subsequent requests. This ensures the tool output validation passes on the API side.

---

- **OS & Arch:** [e.g. Mac M1 / Ubuntu 22.04 / Windows 11]
- **Interface:** [UI / CLI]
- **Version:** [e.g. v1.0.2]
- **Extensions enabled:** [e.g. Memory, Computer Controller]
- **Provider & Model:** [e.g. Google – gemini-3-pro-preview]

---

**Additional context**
This seems to be a strict validation requirement introduced in newer Gemini models (Gemini 3 Pro).

**References from Gemini API Docs:**
> "As a general rule, if you receive a thought signature in a model response, you should pass it back exactly as received when sending the conversation history in the next turn. When using Gemini 3 Pro, you must pass back thought signatures during function calling, otherwise you will get a validation error (4xx status code)."

> "The first functionCall part in each step of the current turn must include its thought_signature. If you omit a thought_signature... the request will fail with a 400 error."

Documentation Link: https://ai.google.dev/gemini-api/docs/thought-signatures

**Proposed Solution for Devs:**
Ensure the internal handling of `functionCall` objects or Content `parts` includes the `thought_signature` property (encrypted string) and includes it when constructing the message history payload for the next turn.

SalvatoreT:
## Summary
Gemini 3 requires [Thought Signatures](https://ai.google.dev/gemini-api/docs/thought-signatures) to use it. This PR implements it.

### Type of Change
- [x] Feature
- [x] Bug fix

### AI Assistance

### Testing
Unit tests + using the command line with `gemini-3-pro-preview`

### Related Issues
fixes #5792

### DIFF
--- a/crates/goose/src/providers/formats/google.rs
+++ b/crates/goose/src/providers/formats/google.rs
@@ -53,9 +53,14 @@ pub fn format_messages(messages: &[Message]) -> Vec<Value> {
                                 }
                             }
 
-                            parts.push(json!({
-                                "functionCall": function_call_part
-                            }));
+                            let mut part = Map::new();
+                            part.insert("functionCall".to_string(), json!(function_call_part));
+
+                            if let Some(signature) = &request.thought_signature {
+                                part.insert("thoughtSignature".to_string(), json!(signature));
+                            }
+
+                            parts.push(json!(part));
                         }
                         Err(e) => {
                             parts.push(json!({"text":format!("Error: {}", e)}));
@@ -120,6 +125,12 @@ pub fn format_messages(messages: &[Message]) -> Vec<Value> {
                                 parts.push(json!({"text":format!("Error: {}", e)}));
                             }
                         }
+                    }
+                    MessageContent::Thinking(thinking) => {
+                        let mut part = Map::new();
+                        part.insert("text".to_string(), json!(thinking.thinking));
+                        part.insert("thoughtSignature".to_string(), json!(thinking.signature));
+                        parts.push(json!(part));
                     }
 
                     _ => {}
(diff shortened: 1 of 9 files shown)


The-Best-Codes (commented):
Looks good, I didn't test it locally but will soon

crates/goose/src/providers/formats/google.rs
```diff
@@ -53,9 +53,15 @@ pub fn format_messages(messages: &[Message]) -> Vec<Value> {
                                 }
                             }
 
-                            parts.push(json!({
-                                "functionCall": function_call_part
-                            }));
+                            let mut part = Map::new();
+                            part.insert("functionCall".to_string(), json!(function_call_part));
+
+                            // Preserve thought signature if present
```
DOsinga: remove the redundant comment

crates/goose/src/providers/formats/google.rs
```diff
@@ -269,8 +281,19 @@ pub fn response_to_message(response: Value) -> Result<Message> {
         .unwrap_or(&binding);
 
     for part in parts {
+        // Extract thought signature if present
```
DOsinga: same

crates/goose/src/providers/formats/google.rs
```diff
@@ -890,4 +914,191 @@ mod tests {
         assert_eq!(regular_field["type"], "number");
         assert_eq!(regular_field["description"], "A regular number field");
     }
+
+    #[test]
```
DOsinga: https://block-eng-blog.stage.sqprod.co/blog/the-high-cost-of-free-testing

instead let's just add the new model to the scripts/test_providers.sh

SalvatoreT: Ah, here's the public post: https://engineering.block.xyz/blog/the-high-cost-of-free-testing

Very timely post (for me)!

michaelneale:
just running a live test now (as can't from forks)...

PR #5809, opened 2025-11-19
Fix keyboard shortcut conflict for Focus Goose Window

### LINKED ISSUE #5782: Cmd+Option+Shift+G assigned to both Focus Window and Quick Launcher
**Describe the bug**

Two different actions are assigned to the same keyboard shortcut: `Cmd+Option+Shift+G` (macOS) or `Ctrl+Alt+Shift+G` (Windows/Linux). The File menu shows "Focus goose Window" with this shortcut, but the global shortcut registration opens the Quick Launcher popup instead. This creates a conflict where the keyboard shortcut behavior doesn't match what's documented in the menu.

---

**To Reproduce**
1. Open goose Desktop.
2. Look at the File menu and note "Focus goose Window" shows `Cmd+Option+Shift+G` (macOS) or `Ctrl+Alt+Shift+G` (Windows/Linux).
3. Use the keyboard shortcut.
4. Observe the Quick Launcher popup opening instead of focusing the goose window.

---

**Expected behavior**

Each action should have a unique keyboard shortcut. Either:
- The "Focus goose Window" menu item should display a different shortcut (or no shortcut), OR
- The Quick Launcher should use a different global shortcut

---

**Screenshots**

Not applicable.

---

- **OS & Arch:** macOS Ventura 13.5 ARM64
- **Interface:** UI (goose Desktop)
- **Version:** v1.14.0
- **Extensions enabled:** None
- **Provider & Model:** OpenAI – gpt-4

---

**Additional context**

In `main.ts`, the global shortcut is registered to `createLauncher()` while the File menu item shows the same shortcut but calls `focusWindow()`. The global shortcut takes precedence, making the menu's displayed shortcut invalid.

JJSwigut:
- Changed Focus Goose Window shortcut from Cmd+Option+Shift+G to Cmd+Option+G
- Registered new global shortcut for Focus Window (Cmd+Option+G)
- Quick Launcher remains on Cmd+Option+Shift+G
- Resolves conflict where both actions shared the same shortcut

Fixes #5782

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -1771,6 +1771,14 @@ async function appMain() {
     console.error('Error registering launcher hotkey:', e);
   }
 
+  try {
+    globalShortcut.register('CommandOrControl+Alt+G', () => {
+      focusWindow();
+    });
+  } catch (e) {
+    console.error('Error registering focus window hotkey:', e);
+  }
+
   session.defaultSession.webRequest.onBeforeSendHeaders((details, callback) => {
     details.requestHeaders['Origin'] = 'http://localhost:5173';
     callback({ cancel: false, requestHeaders: details.requestHeaders });
@@ -1948,7 +1956,7 @@ async function appMain() {
     fileMenu.submenu.append(
       new MenuItem({
         label: 'Focus Goose Window',
-        accelerator: 'CmdOrCtrl+Alt+Shift+G',
+        accelerator: 'CmdOrCtrl+Alt+G',
         click() {
           focusWindow();
         },


ui/desktop/src/main.ts
```diff
@@ -1771,6 +1771,14 @@ async function appMain() {
...
   }
 
+  try {
+    globalShortcut.register('CommandOrControl+Alt+G', () => {
+      focusWindow();
+    });
+  } catch (e) {
+    console.error('Error registering focus window hotkey:', e);
+  }
+
```
Abhijay007, line 1781: Just a question, do we need to add this here, as we already have an accelerator for the focus menu?

PR #5813, opened 2025-11-19
Next Camp Live - Added editing messages functionality

zanesq:
## Summary
Follow up for [next camp live](https://github.com/block/goose/pull/5706) to add editing messages on the server. Added another option to fork the conversation into a new session also.

Screenshot 2025-11-19 at 9 49 49 AM

### DIFF
--- a/ui/desktop/src/components/BaseChat.tsx
+++ b/ui/desktop/src/components/BaseChat.tsx
@@ -85,20 +85,13 @@ function BaseChatContent({
 
   const onStreamFinish = useCallback(() => {}, []);
 
-  const [sessionLoaded, setSessionLoaded] = useState(false);
-  const [hasSubmittedInitialMessage, setHasSubmittedInitialMessage] = useState(false);
-  const [hasTriggeredAgentForFork, setHasTriggeredAgentForFork] = useState(false);
   const [isCreateRecipeModalOpen, setIsCreateRecipeModalOpen] = useState(false);
+  const hasAutoSubmittedRef = useRef(false);
 
+  // Reset auto-submit flag when session changes
   useEffect(() => {
-    setSessionLoaded(false);
-    setHasSubmittedInitialMessage(false);
-    setHasTriggeredAgentForFork(false);
+    hasAutoSubmittedRef.current = false;
   }, [sessionId]);
-
-  const handleSessionLoaded = useCallback(() => {
-    setSessionLoaded(true);
-  }, []);
 
   const {
     session,
@@ -114,39 +107,25 @@ function BaseChatContent({
   } = useChatStream({
     sessionId,
     onStreamFinish,
-    onSessionLoaded: handleSessionLoaded,
   });
 
   useEffect(() => {
-    const shouldStartAgent = searchParams.get('shouldStartAgent') === 'true';
-    if (sessionLoaded && initialMessage && !hasSubmittedInitialMessage) {
-      setHasSubmittedInitialMessage(true);
-      handleSubmit(initialMessage);
+    if (!session || hasAutoSubmittedRef.current) {
       return;
     }
 
-    if (
-      sessionLoaded &&
-      shouldStartAgent &&
-      messages.length > 0 &&
-      !hasTriggeredAgentForFork &&
-      !initialMessage &&
-      chatState === ChatState.Idle
-    ) {
-      setHasTriggeredAgentForFork(true);
+    const shouldStartAgent = searchParams.get('shouldStartAgent') === 'true';
+
+    if (initialMessage) {
+      // Submit the initial message (e.g., from fork)
+      hasAutoSubmittedRef.current = true;
+      handleSubmit(initialMessage);
+    } else if (shouldStartAgent) {
+      // Trigger agent to continue with existing conversation
+      hasAutoSubmittedRef.current = true;
       handleSubmit('');
     }
-  }, [
-    sessionLoaded,
-    initialMessage,
-    hasSubmittedInitialMessage,
-    searchParams,
-    messages.length,
-    hasTriggeredAgentForFork,
-    chatState,
-    handleSubmit,
-    messages,
-  ]);
+  }, [session, initialMessage, searchParams, handleSubmit]);
 
   const handleFormSubmit = (e: React.FormEvent) => {
     const customEvent = e as unknown as CustomEvent;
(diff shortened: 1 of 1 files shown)


crates/goose-server/src/routes/reply.rs
```diff
@@ -85,6 +85,8 @@ pub struct ChatRequest {
     session_id: String,
     recipe_name: Option<String>,
     recipe_version: Option<String>,
+    #[serde(default)]
+    skip_add_user_message: bool,
```
DOsinga: rule number one, no optional

crates/goose/src/agents/agent.rs
```diff
@@ -776,6 +776,18 @@ impl Agent {
...
+            .await
+    }
+
+    #[instrument(skip(self, user_message, session_config), fields(user_message))]
+    pub async fn reply_with_options(
+        &self,
+        user_message: Message,
+        session_config: SessionConfig,
+        cancel_token: Option<CancellationToken>,
+        skip_add_message: bool,
```
DOsinga: why split this function? if we need an extra parameter, just add the parameter

crates/goose/src/conversation/message.rs
```diff
@@ -468,6 +468,8 @@ impl MessageMetadata {
 #[serde(rename_all = "camelCase")]
 pub struct Message {
     pub id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub row_id: Option<i64>,
```
DOsinga: why do we need a row_id here? can't we just use the id we already have?

crates/goose-server/src/routes/reply.rs
```diff
@@ -300,10 +302,11 @@ pub async fn reply(
         };
 
         let mut stream = match agent
-            .reply(
+            .reply_with_options(
```
DOsinga: this strikes me as cumbersome

what I would do/expect here is when we fork the conversation we fork/truncate without the new message. in the case of edit you can then just call the normal submit which will add the edited message to the conversation just as before.

in the case of fork, you can just do the same thing as we do in hub and immediately submit the new message

crates/goose-server/src/routes/session.rs
```diff
@@ -307,6 +338,82 @@ async fn import_session(
...
+        (status = 401, description = "Unauthorized - Invalid or missing API key"),
+        (status = 404, description = "Session or message not found"),
+        (status = 500, description = "Internal server error")
+    ),
+    security(
+        ("api_key" = [])
+    ),
+    tag = "Session Management"
+)]
+async fn edit_message(
```
DOsinga, line 354: I think here you'd just want two simple primitive; copy which takes a session id and returns a new one with an id and truncate which takes a session id and a message id and then deletes all messages including and after that message id

crates/goose/src/session/session_manager.rs
```diff
@@ -290,6 +290,28 @@ impl SessionManager {
         Self::instance().await?.import_session(json).await
     }
 
+    pub async fn fork_session_at_message(
```
DOsinga: which then can be relfected here

crates/goose/src/session/session_manager.rs
```diff
@@ -1137,6 +1160,118 @@ impl SessionStorage {
         self.get_session(&session.id, true).await
     }
 
+    async fn fork_session_at_message(
+        &self,
+        session_id: &str,
+        message_row_id: i64,
+        new_message_content: String,
+    ) -> Result<Session> {
+        use crate::conversation::message::MessageContent;
```
DOsinga: why the local import

crates/goose/src/session/session_manager.rs
```diff
@@ -1137,6 +1160,118 @@ impl SessionStorage {
...
+            };
+
+            let content = serde_json::from_str(&content_json)?;
+            let metadata = metadata_json
+                .and_then(|json| serde_json::from_str(&json).ok())
+                .unwrap_or_default();
+
+            let mut message = Message::new(role, created_timestamp, content);
+            message.metadata = metadata;
+            messages.push(message);
```
DOsinga: this is all duplicated code, no?

we should either add a flag to get_conversation to return messages only up to the desired message or and that's slower but simpler, just get the whole conversation and then add a truncate method to conversation class (haha)

I'd go with the last one

crates/goose/src/session/session_manager.rs
```diff
@@ -1137,6 +1160,118 @@ impl SessionStorage {
...
+        session_id: &str,
+        message_row_id: i64,
+        new_message_content: String,
+    ) -> Result<Session> {
+        use crate::conversation::message::MessageContent;
+
+        let original_session = self.get_session(session_id, false).await?;
+        let rows = sqlx::query_as::<_, (String, String, i64, Option<String>)>(
+            "SELECT role, content_json, created_timestamp, metadata_json FROM messages WHERE session_id = ? AND id < ? ORDER BY id",
+        )
```
DOsinga: we normally sort by timestamp - sorting on id probably works. but it is more consistent to use the timestamp. and from that perspective, use the timestamp also to decide where to truncate the conversation. that is what the user sees

crates/goose/src/session/session_manager.rs
```diff
@@ -1137,6 +1160,118 @@ impl SessionStorage {
...
+
+        let rows = sqlx::query_as::<_, (String, String, i64, Option<String>)>(
+            "SELECT role, content_json, created_timestamp, metadata_json FROM messages WHERE session_id = ? AND id < ? ORDER BY id",
+        )
+            .bind(session_id)
+            .bind(message_row_id)
+            .fetch_all(&self.pool)
+            .await?;
+
+        let mut messages = Vec::new();
```
DOsinga: even more code duplication

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -120,6 +125,29 @@ function BaseChatContent({
...
+    }
+  }, [
+    sessionLoaded,
+    searchParams,
+    messages.length,
+    hasTriggeredAgentForFork,
+    chatState,
+    handleSubmit,
+    messages,
+  ]);
```
DOsinga: these checks make me feel very uneasy. this looks brittle and very specific and reminds me of the sins of old we had in the recipe manager. I don't think you need this at all if we don't add the message

PR #5817, opened 2025-11-19
Revert "fix: Add schema-aware numeric coercion for MCP tool arguments…

jamadeo:
This is a good change to make but as implemented it calls .list_tools() on every token chunk response from a provider. .list_tools() is potentially very expensive, as it queries all extensions for their tools. This is especially noticeable if you have remote extensions, as they tend to have a higher latency.

reverts #5478

This reverts commit https://github.com/block/goose/commit/78fd31a47ae1a618463de751459156b71448f6fc.

### DIFF
--- a/crates/goose/src/agents/reply_parts.rs
+++ b/crates/goose/src/agents/reply_parts.rs
@@ -3,7 +3,6 @@ use std::sync::Arc;
 
 use async_stream::try_stream;
 use futures::stream::StreamExt;
-use serde_json::{json, Value};
 use tracing::debug;
 
 use super::super::agents::Agent;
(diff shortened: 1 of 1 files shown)


DOsinga (approved):

jamadeo:
@tlongwell-block yes that would do it. The important thing to avoid is calls to list_tools() that query all MCP servers.

jamadeo:
@tlongwell-block did you end up making the above change? if so we can close this PR, if not let's merge this if it's still an active issue because it could be quite bad for anyone connected to remote MCPs

tlongwell-block:
> @tlongwell-block did you end up making the above change? if so we can close this PR, if not let's merge this if it's still an active issue because it could be quite bad for anyone connected to remote MCPs

I can open that PR today. I'll link it here when its ready

alexhancock:
@jamadeo @tlongwell-block Is this one still relevant, or replaced with another PR?

tlongwell-block:
@alexhancock @jamadeo https://github.com/block/goose/pull/6138 should fix this issue

Issue #5777, opened 2025-11-17
Please isolate your environment


klonuo:
```bat
Microsoft Windows [Version 10.0.26100.7171]
(c) Microsoft Corporation. All rights reserved.

C:\Users\klo>where uvx
C:\Users\klo\AppData\Local\Goose\bin\uvx.exe
C:\Users\klo\.local\bin\uvx.exe

C:\Users\klo>where npx
C:\Program Files\nodejs\npx
C:\Program Files\nodejs\npx.cmd
C:\Users\klo\AppData\Local\Goose\bin\npx.cmd
C:\Users\klo\AppData\Roaming\npm\npx
C:\Users\klo\AppData\Roaming\npm\npx.cmd

C:\Users\klo>uvx --from git+https://github.com/oraios/serena serena start-mcp-server --help
error: Could not find the `uv` binary at: C:\Users\klo\AppData\Local\Goose\bin\uv.exe

C:\Users\klo>
```

You are overshadowing global `uvx` by injecting into PATH

Please try to use better method

michaelneale:
thanks @klonuo yeah - reading about `setx` (not a windows user here) that isnt' the right approach at all - https://github.com/block/goose/pull/5821 will remove that (not sure if you wanted to take a look at that). 

Another option (and need to think this over, maybe with @jamadeo) is we don't at all bundle uvx et al with windows - use whatever is already there (it is needed for MCP extensions etc) - a more extreme change.

jamadeo:
Yikes that is bad. Thanks @michaelneale for removing that.

PR #5821, opened 2025-11-20
fix: remove setx calls to not permanently edit the windows shell PATH

michaelneale:
fixes: #5777 

`setx` seems like it is not a great thing to call, it should not mess with users env, even if it limits what it can do.

@jamadeo not sure who is best equipped with windows to think on this, but it seems like we shoulen't be messing with PATH that exists outside of goose.

### DIFF
--- a/ui/desktop/src/utils/winShims.ts
+++ b/ui/desktop/src/utils/winShims.ts
@@ -1,7 +1,6 @@
 import fs from 'node:fs';
 import path from 'node:path';
 import os from 'node:os';
-import { spawn } from 'child_process';
 import log from './logger';
 
 /**
@@ -21,8 +20,8 @@ export async function ensureWinShims(): Promise<void> {
   try {
     await fs.promises.mkdir(tgtDir, { recursive: true });
 
-    // Only copy the command-line tools, NOT goosed.exe (which should always be used locally)
-    const shims = ['uvx.exe', 'npx.cmd', 'install-node.cmd'];
+    // Copy command-line tools, NOT goosed.exe (which should always be used locally)
+    const shims = ['uvx.exe', 'uv.exe', 'npx.cmd', 'install-node.cmd'];
 
     await Promise.all(
       shims.map(async (shim) => {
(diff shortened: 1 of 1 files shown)


ui/desktop/src/utils/winShims.ts
```diff
@@ -39,12 +38,12 @@ export async function ensureWinShims(): Promise<void> {
       })
     );
 
-    // Prepend to PATH **for this process & all children**.
-    // Make sure our bin directory is at the VERY BEGINNING of PATH
```
klonuo, line 43: This comment explains it - it injects at the beginning of PATH and as such is the first valid command, as PATH search path is executed from left to right. If it would just append at the end it wouldn't make such problem, but not changing PATH environment variable w/o user consent would be even greater.

ui/desktop/src/utils/winShims.ts
```diff
@@ -54,71 +53,10 @@ export async function ensureWinShims(): Promise<void> {
         // Remove it from its current position and add to beginning
         pathParts.splice(binDirIndex, 1);
         process.env.PATH = `${tgtDir}${path.delimiter}${pathParts.join(path.delimiter)}`;
-        log.info(`Moved ${tgtDir} to beginning of PATH for current process`);
+        log.info(`Moved ${tgtDir} to beginning of PATH for Goose processes only`);
       }
     }
-
-    // Optional: Persist PATH for user's external PowerShell/CMD sessions
```
jamadeo, line 61: The "Optional" here says it all: this was probably inserted by llm and barely glanced at. We should definitely not be adding these to the user's PATH.

jamadeo (approved):

ui/desktop/src/utils/winShims.ts
michaelneale: yep, thanks for bringing it up @klonuo (and goose made it clear it was bad, and even when I switched models a few times, as I didn't believe that it would do that!). Its been decades since I was a windows dev and the registry gives me chills (not the good kind).

PR #5822, opened 2025-11-20
chore: some old code hanging around, and mention configure cli

michaelneale:
There was this old provider code which is no longer needed, and thought I should make it clear for copilot that you currently need to use the cli to do the oauth login.


codefromthecrypt (approved):
good tips. thanks for the cleanup

alexhancock:
Nice

PR #5825, opened 2025-11-20
Disallow subagents with no extensions

### LINKED ISSUE #5812: Claude sonnet 4 is very eager to create sub-agents, and goose does not handle it well
I ran the "create a tamagotchi game" test, and:

- the main agent is instructed to build a game
- the main agent immediately creates a sub-agent to handle the whole thing
- the sub agent tries to also create a sub-agent
- the recursive sub-agent call fails (as it should)
- the sub-agent then replies with some code directly (no developer tools are called) and reports success to the main agent
- the main agent summarizes this reply

so in the end, goose reported that everything worked. And no code was written.

goose version: 1.15.0 (pre release e9b9dcc3634bcc1a12504e33a29baaea631f6ddc)

[diagnostics_20251119_7.zip](https://github.com/user-attachments/files/23633476/diagnostics_20251119_7.zip)

jamadeo:
If we present no tools to the model except a "execute subagent task" tool, the model will try to use that tool to do everything.

fixes #5812

### DIFF
--- a/scripts/test_compaction.sh
+++ b/scripts/test_compaction.sh
@@ -151,7 +151,7 @@ echo ""
 # TEST 2: Auto Compaction
 # ==================================================
 echo "---------------------------------------------------"
-echo "TEST 2: Auto Compaction via threshold (0.01)"
+echo "TEST 2: Auto Compaction via threshold (0.005)"
 echo "---------------------------------------------------"
 
 TESTDIR=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -693,19 +693,41 @@ impl Agent {
         Ok(())
     }
 
+    pub async fn subagents_enabled(&self) -> bool {
+        let config = crate::config::Config::global();
```
DOsinga, line 644: now that this is in the agent context (yeah!) we should be able to add the session_type to this too and not enable subagents if the session type is subagent. that way instead of failing the creation of a subagent when a subagent is trying to create one, it will not even know about it

DOsinga (approved):

Issue #5824, opened 2025-11-20
Feature Request: Add Session Cleanup API to Prevent MCP Process Accumulation


wpfleger96:
## Problem

When using Goose programmatically via the API in long-running applications (e.g., Slack bots), MCP processes accumulate indefinitely and are never cleaned up, causing memory leaks that eventually lead to OOM crashes.

## Evidence

In our Kubernetes-deployed Slack bot, we observe:
- Linear memory growth corresponding to each new Slack thread interaction
- Multiple instances of the same MCP servers accumulating (6+ copies of each configured MCP server)
- Memory never decreases - processes persist until pod restart
- Each new conversation thread spawns a complete new set of MCP processes

## Current Behavior

1. Each session spawns new MCP processes via ExtensionManager
2. Sessions created via API persist indefinitely
3. No mechanism exists to clean up sessions or their associated MCP processes
4. The \`ChildWithCleanup\` Drop trait is never triggered because Extension objects are never removed from the ExtensionManager

## Proposed Solution

Add a session cleanup API endpoint that:
- Removes the session from memory
- Cleans up associated extensions and MCP processes
- Properly triggers Drop implementations to terminate child processes

Suggested endpoint: \`DELETE /agent/session/{session_id}\`

## Use Case

This primarily affects:
- Long-running bot deployments (Slack, Discord, etc.)
- Any programmatic API usage where the client persists across multiple sessions
- Kubernetes/containerized deployments where memory limits are enforced

The issue doesn't occur in CLI usage because the process terminates after each task, triggering automatic cleanup.

PR #5826, opened 2025-11-20
Add /agent/stop endpoint, make max active agents configurable

tlongwell-block:
Resolves https://github.com/block/goose/issues/5824

### DIFF
--- a/crates/goose/src/execution/manager.rs
+++ b/crates/goose/src/execution/manager.rs
@@ -39,7 +39,7 @@ impl AgentManager {
         let scheduler = Scheduler::new(schedule_file_path).await?;
 
         let capacity = NonZeroUsize::new(max_sessions.unwrap_or(DEFAULT_MAX_SESSION))
-            .unwrap_or_else(|| NonZeroUsize::new(10).unwrap());
+            .unwrap_or_else(|| NonZeroUsize::new(100).unwrap());
 
         let manager = Self {
             sessions: Arc::new(RwLock::new(LruCache::new(capacity))),


crates/goose/src/execution/manager.rs
```diff
@@ -52,7 +53,10 @@ impl AgentManager {
     pub async fn instance() -> Result<Arc<Self>> {
         AGENT_MANAGER
             .get_or_try_init(|| async {
-                let manager = Self::new(Some(DEFAULT_MAX_SESSION)).await?;
+                let max_sessions = Config::global()
+                    .get_goose_max_active_agents()
```
DOsinga, line 57: this is a singleton so making it depend on a configuration variable only somewhat works

more to the point and I think we discussed this previously, I don't think the agent manager should be a cache and definitely not one with max=10 by default - we still have a rumbling subagent problem and that would mean if you start 10 subagents, the main agent dies.

so adding an agent/stop path is the right way, but the clients should either always handle this or the agents should be properly resumed (now it loads the default provider, no extensions etc)

tlongwell-block: >so adding an agent/stop path is the right way, but the clients should either always handle this or the agents should be properly resumed (now it loads the default provider, no extensions etc)

I think https://github.com/block/goose/pull/5419 resolves some of this

tlongwell-block: restored default to 100

DOsinga: > I think https://github.com/block/goose/pull/5419 resolves some of this

it's definitely on the way there. the next step would be to store the active extensions also and then we can move code from resume_agent to here and make sure that everybody uses the agent manager and then we can kill agents and also have agents restore themselves after a system restart

PR #5831, opened 2025-11-20
goose tui

tlongwell-block:
This PR introduces a terminal user interface for goose, along with a first-party Rust client library for goose-server.

### TUI Features

A lightweight, keyboard-driven interface for goose that runs entirely in your terminal:

- **Embedded server**: No separate processes—goose-server runs in-process
- **Rich rendering**: Markdown support with syntax-highlighted code blocks
- **Tool inspection**: View available tools and their schemas. Create slash commands to run tools
- **Vim-style navigation**: Normal/Editing/Visual modes with familiar keybindings
- **Session management**: Resume previous conversations or start fresh. Support conversation forking.
- **Live configuration**: Switch providers, models, and extensions without restarting
- **Smart context**: Automatic project analysis on new sessions
- **Copy mode**: Select text to copy with keyboard+mouse or copy entire message/tool response to clipboard
- **Themes**: Beautiful built-in themes and support for custom themes defined in your goose config
- **Multiple modes**: Full TUI, lightweight CLI mode, server only, or headless for automation

The TUI provides a fast, distraction-free way to interact with goose for developers who prefer staying in the terminal.

### goose-client

A reusable Rust client library for goose-server that provides:

- Type-safe API bindings for all goose-server endpoints
- SSE streaming for real-time message events
- Async/await support with tokio
- Minimal dependencies (reqwest, serde)

Having a first-party Rust client offers several benefits:

1. **Type safety**: Compile-time guarantees that client/server contracts match
2. **Reference implementation**: Demonstrates best practices for building goose-server clients
3. **Ecosystem growth**: Makes it easier for Rust developers to build custom goose interfaces

The client is designed to be lightweight and focused—it handles HTTP communication and serialization, while leaving UI concerns to consumers like goose-tui.

goose_logo_tui
tui_screen2
retro

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4863,6 +4863,7 @@ dependencies = [
  "tracing-appender",
  "tracing-subscriber",
  "unicode-width 0.2.2",
+ "url",
  "urlencoding",
  "uuid",
 ]
--- a/crates/goose-tui/Cargo.toml
+++ b/crates/goose-tui/Cargo.toml
@@ -56,6 +56,7 @@ bzip2 = "0.5"
 open = "5.3.2"
 reqwest = { version = "0.13", default-features = false, features = ["blocking", "rustls"] }
 urlencoding = "2.1"
+url.workspace = true
 
 [features]
 disable-update = []
--- a/crates/goose-tui/src/commands/recipe.rs
+++ b/crates/goose-tui/src/commands/recipe.rs
@@ -19,8 +19,44 @@ pub fn handle_validate(recipe_name: &str) -> Result<()> {
     Ok(())
 }
 
-pub fn handle_deeplink(recipe_name: &str, params: &[String]) -> Result<String> {
-    let params_map = parse_params(params)?;
+pub fn handle_explain(recipe_name: &str, params: &[(String, String)]) -> Result<()> {
+    let rf = read_recipe_file(recipe_name)?;
+    let recipe = validate_recipe_template_from_file(&rf)?;
+    println!("Title: {}", recipe.title);
+    println!("Description: {}", recipe.description);
+    if let Some(prompt) = &recipe.prompt {
+        println!("Prompt: {}", prompt);
+    }
+    if !params.is_empty() {
+        println!("\nParameters provided:");
+        for (k, v) in params {
+            println!("  {} = {}", k, v);
+        }
+    }
+    Ok(())
+}
+
+pub fn handle_render(recipe_name: &str, params: &[(String, String)]) -> Result<()> {
+    let rf = read_recipe_file(recipe_name)?;
+    let recipe = validate_recipe_template_from_file(&rf)?;
+    match serde_yaml::to_string(&recipe) {
+        Ok(yaml) => {
+            if !params.is_empty() {
+                println!("# Parameters:");
+                for (k, v) in params {
+                    println!("#   {} = {}", k, v);
+                }
+                println!();
+            }
+            println!("{}", yaml);
+            Ok(())
+        }
+        Err(e) => anyhow::bail!("Failed to serialize recipe: {}", e),
+    }
+}
+
+pub fn handle_deeplink(recipe_name: &str, params: &[(String, String)]) -> Result<()> {
+    let params_map: HashMap<String, String> = params.iter().cloned().collect();
     match generate_deeplink(recipe_name, params_map) {
         Ok((deeplink_url, recipe)) => {
             println!(
@@ -29,7 +65,7 @@ pub fn handle_deeplink(recipe_name: &str, params: &[String]) -> Result<String> {
                 recipe.title
             );
             println!("{}", deeplink_url);
-            Ok(deeplink_url)
+            Ok(())
         }
         Err(err) => {
             println!(
@@ -42,7 +78,7 @@ pub fn handle_deeplink(recipe_name: &str, params: &[String]) -> Result<String> {
     }
 }
 
-pub fn handle_open(recipe_name: &str, params: &[String]) -> Result<()> {
+pub fn handle_open(recipe_name: &str, params: &[(String, String)]) -> Result<()> {
     handle_open_with(
         recipe_name,
         params,
@@ -53,7 +89,7 @@ pub fn handle_open(recipe_name: &str, params: &[String]) -> Result<()> {
 
 fn handle_open_with<F, W>(
     recipe_name: &str,
-    params: &[String],
+    params: &[(String, String)],
     opener: F,
     out: &mut W,
 ) -> Result<()>
@@ -61,7 +97,7 @@ where
     F: FnOnce(&str) -> std::io::Result<()>,
     W: std::io::Write,
 {
-    let params_map = parse_params(params)?;
+    let params_map: HashMap<String, String> = params.iter().cloned().collect();
     match generate_deeplink(recipe_name, params_map) {
         Ok((deeplink_url, recipe)) => match opener(&deeplink_url) {
             Ok(_) => {
@@ -103,31 +139,14 @@ pub fn handle_list(_format: &str, _verbose: bool) -> Result<()> {
     anyhow::bail!("recipe list not yet supported in goose-tui - TODO: integrate recipe search")
 }
 
-fn parse_params(params: &[String]) -> Result<HashMap<String, String>> {
-    let mut params_map = HashMap::new();
-    for param in params {
-        let parts: Vec<&str> = param.splitn(2, '=').collect();
-        if parts.len() != 2 {
-            return Err(anyhow::anyhow!(
-                "Invalid parameter format: '{}'. Expected format: key=value",
-                param
-            ));
-        }
-        params_map.insert(parts[0].to_string(), parts[1].to_string());
-    }
-    Ok(params_map)
-}
-
 /// Load a recipe file by path.
-/// TODO: Integrate goose-cli's recipes::search_recipe::load_recipe_file for name-based lookup
 fn load_recipe_file(recipe_name: &str) -> Result<String> {
-    // For now, treat recipe_name as a direct file path
     let path = std::path::Path::new(recipe_name);
     if path.exists() {
         Ok(recipe_name.to_string())
     } else {
         anyhow::bail!(
-            "Recipe file not found: {}. In goose-tui, provide a direct file path.",
+            "Recipe file not found: {}. Provide a direct file path.",
             recipe_name
         )
     }
@@ -202,25 +221,6 @@ instructions: "Test instructions"
 
         let result = handle_deeplink(&recipe_path, &[]);
         assert!(result.is_ok());
-        let url = result.unwrap();
-        assert!(url.starts_with("goose://recipe?config="));
-        let encoded_part = url.strip_prefix("goose://recipe?config=").unwrap();
-        assert!(!encoded_part.is_empty());
-    }
-
-    #[test]
-    fn test_handle_deeplink_with_parameters() {
-        let temp_dir = TempDir::new().expect("Failed to create temp directory");
-        let recipe_path =
-            create_test_recipe_file(&temp_dir, "test_recipe.yaml", VALID_RECIPE_CONTENT);
-
-        let params = vec!["name=John".to_string(), "age=30".to_string()];
-        let result = handle_deeplink(&recipe_path, &params);
-        assert!(result.is_ok());
-        let url = result.unwrap();
-        assert!(url.starts_with("goose://recipe?config="));
-        assert!(url.contains("&name=John"));
-        assert!(url.contains("&age=30"));
     }
 
     #[test]
@@ -234,7 +234,7 @@ instructions: "Test instructions"
 
     fn run_handle_open(
         recipe_path: &str,
-        params: &[String],
+        params: &[(String, String)],
         opener_result: std::io::Result<()>,
     ) -> (Result<()>, String, String) {
         let captured_url = std::cell::RefCell::new(String::new());
@@ -271,13 +271,13 @@ instructions: "Test instructions"
         let recipe_path =
             create_test_recipe_file(&temp_dir, "test_recipe.yaml", VALID_RECIPE_CONTENT);
 
-        let (base_url, _) = generate_deeplink(&recipe_path, HashMap::new()).unwrap();
-
-        let params = vec!["name=Alice".to_string(), "role=developer".to_string()];
+        let params = vec![
+            ("name".to_string(), "Alice".to_string()),
+            ("role".to_string(), "developer".to_string()),
+        ];
         let (result, captured_url, _) = run_handle_open(&recipe_path, &params, Ok(()));
 
         assert!(result.is_ok());
-        assert!(captured_url.starts_with(&base_url));
         assert!(captured_url.contains("&name=Alice"));
         assert!(captured_url.contains("&role=developer"));
     }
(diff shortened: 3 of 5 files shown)


michaelneale:
I love this. We need this @DOsinga !

DOsinga:
love this too. my one question is, do we want to do this in rust? I have two reasons why we should consider typescript. one, I think the TUI support in typescript is better and it is just quicker to develop. the other is that from a platform perspective it would just be neat to have two client implementations that both use the goosed protocol and show to the community how that works and invite them to build more.

tlongwell-block:
> love this too. my one question is, do we want to do this in rust? I have two reasons why we should consider typescript. one, I think the TUI support in typescript is better and it is just quicker to develop. the other is that from a platform perspective it would just be neat to have two client implementations that both use the goosed protocol and show to the community how that works and invite them to build more.

This does use the goosed protocol (check out crates/goose-tui/src/client.rs)

And... I did look at typescript, too. But it is slightly messier. By using rust, we get to reuse a lot from the rest of goose, and embed goose-server directly without having to ship it as a separate file

ratatui is definitely not as easy to make aesthetic and responsive, though

michaelneale:
yeah building on goosed makes sense, but I didn think about a non rust one, it does mean more distirbution hassle but then developers only are using this, so yeah https://github.com/sst/opentui or even using the codex (opensource) code as a client could be doable (or adapting/lifting) - thoughts @tlongwell-block ? rattatui I always found challenging. I guess downside is people need node installed?

tlongwell-block:
> yeah building on goosed makes sense, but I didn think about a non rust one, it does mean more distirbution hassle but then developers only are using this, so yeah https://github.com/sst/opentui or even using the codex (opensource) code as a client could be doable (or adapting/lifting) - thoughts @tlongwell-block ? rattatui I always found challenging. I guess downside is people need node installed?

I'm still playing with it right now, but I'm surprised at how nice it is. Will have a more informed opinion by the end of the day, I think.

Also, I think there's a really good opportunity here wrt code reuse. I'm going to refactor this a little today to make this work more easily accessible to other projects.

And isn't there something kind of romantic about rust?

tlongwell-block:
I broke the client bits into crates/goose-client

I think having a first party rust client that uses the actual structs/enums/etc from goose/goose-server adds a lot of value to the project. What do you think, @DOsinga?

michaelneale:
I am liking this more now - so the goose-client exists mostly to expose things, I wonder if we can simplify that (ie not have another module), also we could drop the main cli so that the goose-tui becomes the one goose binary (and perhaps we can unify the bianry so goose/goosed are same thing - just default is tui?)

Anyway, can dive into those details shortly - but so far looking good, and yeah, if we are to do a TUI, I think it makes sense to do it in rust @DOsinga for now, mostly as I looked around, didn't see any suitable ready to go TUIs in ts/js (certainly not that arent' really just libraries and would take a fair bit of work - but that may happen one day, but for now, its really just apis for TUIs no matter what, so ts, or rust, have to pick one?)

crates/goose-client/src/lib.rs
```diff
@@ -0,0 +1,365 @@
+mod types;
```
michaelneale, line 1: so this really exists to provide glue between goose (core) and server?

crates/goose-server/src/commands/agent.rs
```diff
@@ -27,6 +29,38 @@ async fn shutdown_signal() {
     let _ = tokio::signal::ctrl_c().await;
 }
 
+pub async fn build_app() -> Result<(Router, TcpListener)> {
```
michaelneale, line 33: curious what this does (or why it is build_app - the name?)

crates/goose-server/src/main.rs
```diff
@@ -28,35 +20,29 @@ enum Commands {
...
+        Commands::Mcp { name } => {
+            logging::setup_logging(Some(&format!("mcp-{name}")))?;
+            let server = McpCommand::from_str(name)
+                .map_err(|e| anyhow::anyhow!("Invalid MCP server: {}", e))?;
             match server {
                 McpCommand::AutoVisualiser => serve(AutoVisualiserRouter::new()).await?,
                 McpCommand::ComputerController => serve(ComputerControllerServer::new()).await?,
                 McpCommand::Memory => serve(MemoryServer::new()).await?,
                 McpCommand::Tutorial => serve(TutorialServer::new()).await?,
-                McpCommand::Developer => {
```
michaelneale: this functionality is pushed down now in to dev extension?

crates/goose-tui/AGENTS.md
```diff
@@ -0,0 +1,91 @@
...
+cargo run -p goose-tui
+cargo run -p goose-tui -- --session <ID>
+cargo run -p goose-tui -- --recipe <FILE>
+cargo run -p goose-tui -- --recipe <FILE> --headless
+```
+
+## Structure
+
+```
+src/
```
michaelneale: do we want much detail for agents? (or they can work it out when needed?). (small)

michaelneale:
using this as a daily: 

image

not sure what causes it to keep that persistent message for "clippy" there? anytime it is working, the spinner is next to it (must be some most recent status message?)

(sorry for screen caps)

image


I guess that is showing a command, but not sure what is meant to be in body of the dev tool (is empty in that case)

tlongwell-block:
> I am liking this more now - so the goose-client exists mostly to expose things, I wonder if we can simplify that (ie not have another module), also we could drop the main cli so that the goose-tui becomes the one goose binary (and perhaps we can unify the bianry so goose/goosed are same thing - just default is tui?)

We can absolutely move goose-client back into the tui crate. (That's how it was originally.)

Just a matter of taste. I thought it was independent enough to show a good example of a goose-server client, but it isn't particularly useful as its own crate.

tlongwell-block:
> using this as a daily:
> 
> image
> not sure what causes it to keep that persistent message for "clippy" there? anytime it is working, the spinner is next to it (must be some most recent status message?)

That is the current item on the todo list goose is working on. Hit Ctrl+T to see the full list. I am going to improve its todo list usage a bit to make this status line more useful

> image
> I guess that is showing a command, but not sure what is meant to be in body of the dev tool (is empty in that case)

The yellow line ( > developer__shell (args hidden)) is the tool call, which you can highlight and click (or hit Enter) on to view the full tool call.

The box below that is the tool response. In this case, it appears the tool didn't return any results

As a "goose power user", I wanted to be able to see both in full when I chose to

crates/goose-client/src/lib.rs
tlongwell-block: Yes, this is just a generic http client for a goose-server instance

In an ideal world, all the data structs would be available in something like a goose-types crate that we could use in the client *and* in the server to have rust-enforced guarantees.

Right now I'm pulling them from goose-server directly, which is fine for a first-party use like the TUI, but not great if we wanted to publish a goose-client crate.

crates/goose-server/src/commands/agent.rs
tlongwell-block: This allows running goose-server as an embedded part of the TUI

crates/goose-tui/AGENTS.md
tlongwell-block: I think they could definitely work out what they need to look at by name/directory structure themselves in most cases, but I have found that having this tree saves a bunch of work for the agent when resuming dev on the TUI

crates/goose-server/src/main.rs
tlongwell-block: Yes. We end up not needing this here anymore.

But, this change should not be in this PR, so I have reverted it. Great catch 🙏

crates/goose-client/src/lib.rs
tlongwell-block: I didn't want to include a refactor of the server in this PR, though, so goose-types will have to wait until later. But, as far as the client and tui are concerned, it would be a trivial cutover

michaelneale:
@tlongwell-block ok that makes sense, as long as deliberate. I idd think it odd it kept that there, but that just means it is still in todo I guess? (which is odd itself, but not on the part of the TUI).

crates/goose-tui/AGENTS.md
michaelneale: hrm interesting... fair enough then.

crates/goose-server/src/commands/agent.rs
michaelneale: why is called build_ then?

crates/goose-client/src/lib.rs
michaelneale: yep, it looked small.

michaelneale:
I would love to see this in the goose-cli crate, vs goose-tui - if it was up to me? (I wouldn't want to keep old CLI as well as this, but maybe it is a cut over? but don't want to maintain 2)

michaelneale:
cc @jamadeo @DOsinga would love your take, would really like this to be default goose command vs goose-tui I think (and we still have to support recipes, run and all the other cli constructs somehow) and then eventually retire the old goose-cli code? (and could also perhaps combine some of this with goose-server?) but could be in a phased approach (the way it is now, is a new binary, which means people wont' really discover or use it)

DOsinga:
> cc @jamadeo @DOsinga would love your take, would really like this to be default goose command vs goose-tui I think (and we still have to support recipes, run and all the other cli constructs somehow) and then eventually retire the old goose-cli code? (and could also perhaps combine some of this with goose-server?) but could be in a phased approach (the way it is now, is a new binary, which means people wont' really discover or use it)

yeah, I'm all for thinking ahead and the CLI is a bit old in tooth. but then again, I am not sure that doing product development by PR is the right way in this case. Personally I don't think Rust would be my first choice for a user interface, whether GUI or TUI. I know a lot of work went into this already, but would it be an idea to take a step back and post a discussion about where we want to go with this first?

tlongwell-block:
> yeah, I'm all for thinking ahead and the CLI is a bit old in tooth. but then again, I am not sure that doing product development by PR is the right way in this case. Personally I don't think Rust would be my first choice for a user interface, whether GUI or TUI. I know a lot of work went into this already, but would it be an idea to take a step back and post a discussion about where we want to go with this first?

I thought the exact same thing wrt rust for frontend work. I did full working mockups in `ink` and `ratatui`. Besides some initial teething issues, I ended up greatly preferring `ratatui` and the ability to pull in from the goose codebase directly where needed.

I don't know how folks feel about a TUI for goose in general, though. It may not be the right direction.

I ended up working on this after spending a lot of time with gemini-cli TUI in order to test gemini 3 and found there were strong points that I wanted in my goose experience that I couldn't get with the goose cli.

michaelneale:
@DOsinga @tlongwell-block I'm not sure practically how a TUI which is not in rust would work with goose - would mean need to install goose via npm (for example) instead of a monolithic binary as of now (pros/cons?) - and doesn't there always need to be a "cli" like experience in goose (and the cli is getting a little dated) anyway? Disagree that this  a bad way to explore "product by PR" in this specific case (well assuming it isn't backbraking manual! sorry if it was @tlongwell-block!) - I tried to find other TUIs I liked and other than actually driving it, it is hard to try them out. There really isn't a way to mock/design things like this other than trying it out IMO. 

could move this to a discussion to continue? 

* do we want TUI at all - is it a passing fashion?
* do it in rust, or bring along another dependency if people want cli
* update cli in place
* > whether GUI or TUI: I wouldn't at all group those together for chosing rust or not (I am thinking just replacement for goose-cli but perhaps others have bigger things in mind?)
* move to the @goose terminal enhancement and deprecate goose-cli crate (!)

many good options (I will open this as a discussion, seems mixed engagement there but it is picking up).

tlongwell-block:
Chatting in https://github.com/block/goose/discussions/5958

zanesq:
Tested locally and works well! UI looks really good also.

I'm not too familiar with TUIs so a couple things that stick out to me:
1. mouse support for highlighting and copying text would be nice
2. it pops the message details everywhere I click with the mouse which feels a bit strange/annoying to me, maybe the click trigger could be localized to the message itself or an icon at the top for more details?

tlongwell-block:
> Tested locally and works well! UI looks really good also.
> 
> I'm not too familiar with TUIs so a couple things that stick out to me:
> 
> 1. mouse support for highlighting and copying text would be nice
> 2. it pops the message details everywhere I click with the mouse which feels a bit strange/annoying to me, maybe the click trigger could be localized to the message itself or an icon at the top for more details?

Thanks for testing this out!

Mouse support is there, but requires hitting Ctrl+S to shift into copy mode first. Aside from that, pressing 'c' when viewing the details popup on a message lets you copy the whole message. This setup is at par with other TUIs afaik, but I am still working on making this better. It should be easier.

Yes, I can see that clicking behavior not making sense. It's doing the same thing as `enter` and opening the highlighted message (which is normally the last one unless you scroll back). Let me fix this. No one would think clicking would open the highlighted message instead of whatever they're clicking on.

tlongwell-block:
Something else to think about here: since this is a pure rust implementation, we are able to bundle the actual goose-server for the TUI to use.

Which means we don't need to build a separate goose-server binary to ship with the desktop UI if we adopt the TUI.

The TUI binary can do `goose-tui server` already to run goose-server without a UI, which the desktop app can use.

So... anyone who would install the desktop UI gets a TUI for free, essentially. With very little added weight to the distribution.

codefromthecrypt:
@tlongwell-block probably doesn't apply to here but in IDEs like jetbrains claude has intense flickering. Not sure we've anything in common with it, but could be worth a review to see if there's anything in pretty tui here that could result in flickering https://github.com/anthropics/claude-code/issues/769

taniacryptid:
Posting @tlongwell-block's update in context for hands-on review 🙏 

> Hi folks  - would really appreciate some eyes (and hands-on time). This adds a `goose-tui` binary that’s a nice terminal experience (and it also has a plain `--cli` mode and a `server` only mode). A big upside is that it lets us move everything onto goose-server.
> 
> Today, `goose-cli` has its own backend setup. This PR is a path to having every `goose` UI share the same backend machinery and behavior.
> 
> We can merge it as `goose-tui` first, ask people to try it for a few weeks, iterate on feedback, and then over the next couple months start deprecating `goose-cli` and the standalone `goose-server/goosed` binary once we’re confident it’s solid and we’re not breaking anything anyone cares about.
> 
> Or maybe we don't want a TUI for goose and we don't want to mess with the `goose cli` that's already working fine, I'd just like to close out or move forward on this PR.

@block/goose-core-maintainers @block/goose-core-maintainers

alexhancock:
@taniandjerry @tlongwell-block We definitely need a new terminal UX!

I think we should have a chat about the future of the cli. As part of https://github.com/block/goose/issues/6642 I prototyped an InkJS based TUI that was incredibly powerful, and implemented in a very small number of files due to expressiveness of TS and the ACP interface.

https://github.com/block/goose/compare/main...alexhancock/goosed-acp-and-new-cli#diff-0c954f5db4ecda500f284985d4cad2bc18bd3d3b52b2efff237dad38ea7a1e18

I built a prototype that had a single agent mode as well as an orchestrator mode that could spin up multiple agents each working on worktrees. I came away feeling like this is definitely the way forward for new clients, including our own.

Let's discuss next week?

cc @DOsinga

albina-popova:
@DOsinga is there any chance this PR would be merged soon? Would much appreciate that.

tlongwell-block:
Will contribute to https://github.com/block/goose/tree/main/ui/acp instead

deenthekiid-blip:
Do everything

deenthekiid-blip:
Go on

PR #5833, opened 2025-11-21
Groq configure fix 

neiii:
## Summary
Fixed `goose configure` command for Groq (d) provider. Previously, if the `groq_api_key` was unset, the cli would not ask for it and error out at the end of the setup. That said, I still think there are some issues with how goose handles Groq's conversations, with all of them being stuck on "Context limit reached. Compacting to continue conversation..."

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance
Made with goose :P 

### Testing
Manual testing + the following:
```
cargo test -p goose --lib providers::factory::tests
cargo test -p goose --lib providers::formats::openai::tests
cargo test -p goose --lib providers::factory::tests::test_groq_provider_config_keys (per commit)
```

Previously:
broken

Now
fixed

### DIFF
--- a/crates/goose/src/providers/provider_registry.rs
+++ b/crates/goose/src/providers/provider_registry.rs
@@ -98,7 +98,6 @@ impl ProviderRegistry {
 
         let mut config_keys = base_metadata.config_keys.clone();
 
-        // Force the API key to be platform-specific over `OPENAI_API_KEY`
         if let Some(api_key_index) = config_keys
             .iter()
             .position(|key| key.required && key.secret)


crates/goose/src/providers/factory.rs
```diff
@@ -372,4 +372,44 @@ mod tests {
         _guard.set("GOOSE_CONTEXT_LIMIT", "64000");
         let _result = create_lead_worker_from_env("openai", &default_model, "gpt-4o");
     }
+
+    #[tokio::test]
+    async fn test_groq_provider_config_keys() {
+        // Test that Groq provider shows correct config keys
+        let providers_list = providers().await;
+
+        // Find the Groq provider
```
michaelneale: probably can drop the one line comments, I know goose et al like to throw them in! but they can confuse later edits (especially if they drift away from what it sreally happening).

crates/goose/src/providers/provider_registry.rs
```diff
@@ -96,14 +96,21 @@ impl ProviderRegistry {
             })
             .collect();
 
+        // Generate appropriate config keys based on the provider engine and configuration
+        let mut config_keys = base_metadata.config_keys.clone();
+        // Update the first config key (API key) to use the provider-specific environment variable
+        if !config_keys.is_empty() {
+            config_keys[0] = super::base::ConfigKey::new(&config.api_key_env, true, true, None);
```
michaelneale: I don't quite follow what this is doing - are there other side effects to this (oobviously beyond making groq work! which is nice)

neiii:
> thanks @neiii - just some nits (other comments) to fix, but also - can you explain more why it needs to set specifically the first key (ie overwrite whatever was there in the config), that seems odd as it would effect config for other providers too?

you're right, this is a rather silly implementation, not sure how I overlooked this. I'm currently debugging another implementation, by the looks of it, Groq is not the only affected provider; any openAI compatible provider is affected, including deepseek, and mistral. I'll update in a bit

DOsinga:
thanks @neiii - let's get groq work here completely. I agree with @michaelneale that blindly copying the key into the first parameter might have side effects though.  a better but more involved solution would be to move the api_key out of the config_keys in ProviderMetaData and just have it there as an explicit field.

neiii:
hey @DOsinga, I pushed a different implementation just as you were writing the message. Now, it gets the position of the OPENAI_API_KEY and then swaps it with the provider's API key. Looks good or would you like more change?

In terms of getting Groq fixed completely, quite frankly I'm not sure what's causing the issue, I've went back-and-forth with it for 2 hours yesterday, I'll tinker some more tonight. The requests go through (and it shows on Groq's dashboard), it's just that when Goose is processing them for some reason it sees it as a context overflow and tries to infinitely compress it, causing the requests to spam. I'll tinker some more tonight

DOsinga:
ok, thanks for the info, very useful. replacing OPEN_AI_KEY is a good start and helps for this for sure, but custom providers based on Anthropic (not very common to be sure) will run into the same issue. we can leave your solution in place though

as for groq, the reason this happens can be either that we have the token count on the models they return wrong so we think we're immediately out of context tokens, or there is something in the reply like "too many" and it is a 500 http error. If you can share a diagnostics report that could be helpful

neiii:
@DOsinga I checked again with Groq now on a paid plan, the (bigger context) models seem to work, I suppose it was a tier issue (?). Is there anything else blocking for this change to be implemented?

crates/goose/src/providers/provider_registry.rs
```diff
@@ -96,14 +96,25 @@ impl ProviderRegistry {
             })
             .collect();
 
+        let mut config_keys = base_metadata.config_keys.clone();
+
+        // Force the API key to be platform-specific over `OPENAI_API_KEY`
+        if let Some(api_key_index) = config_keys
```
DOsinga, line 101: can  you remove the comment? code speaks for itself

crates/goose/src/providers/factory.rs
```diff
@@ -372,4 +372,138 @@ mod tests {
         _guard.set("GOOSE_CONTEXT_LIMIT", "64000");
         let _result = create_lead_worker_from_env("openai", &default_model, "gpt-4o");
     }
+
+    #[tokio::test]
+    async fn test_openai_provider_config_keys() {
```
DOsinga: these tests all test the same thing for different providers. you can replace them with a simple lookup table and call providers_list just one and then go through those providers and for each hit check that the key is indeed the right one.

crates/goose/src/providers/provider_registry.rs
michaelneale: I actually disagree - this is an important quirk in this case and warrants a comment

michaelneale (approved):
looks good, mostly agree with @DOsinga on final nits, very cool!

neiii:
@DOsinga compressed the tests and re-added the comment. Anything else?

crates/goose/src/providers/provider_registry.rs
DOsinga: sorry to disagree again here. the comment says it is overwriting OPENAI_API_KEY but that is just not what it is doing.

if we want to keep a comment here, it should say something like don't inherit the api_key key from the base provider, but use the one from the custom provider

michaelneale: oh in that case, your point doubly stands! if it is wrong then that is really bad (and I guess... as a rule, why comments get stale and can mislead) fair enough! 

Yes, remove the comment! I stand corrected. I assumed it meant what it said which is not great!

PR #5837, opened 2025-11-21
OpenAI responses streaming

katzdave:
This PR implements support for OpenAI's Responses API with streaming

  - Implemented full support for OpenAI's Responses API format with streaming capabilities
  - Added new openai_responses.rs format handler for the responses endpoint
  - Updated OpenAI provider to handle both chat completions and responses formats
  - Improved token counting and usage tracking for the new format

### DIFF
--- a/crates/goose/src/providers/formats/openai_responses.rs
+++ b/crates/goose/src/providers/formats/openai_responses.rs
@@ -312,6 +312,7 @@ fn add_function_call_outputs(input_items: &mut Vec<Value>, messages: &[Message])
                 match &response.tool_result {
                     Ok(contents) => {
                         let text_content: Vec<String> = contents
+                            .content
                             .iter()
                             .filter_map(|c| {
                                 if let RawContent::Text(t) = c.deref() {


katzdave:
Still figuring out how to get usage data for the streaming, but otherwise this is working pretty well. Interestingly usage comes through when there's a tool call.

katzdave:
@michaelneale This seems to be working pretty well for me. Want to take it for test drive?

michaelneale (approved):
nice one @katzdave tested it with gpt-5.2 and codex 5.1 and worked nicely! awesome

cbruyndoncx:
@michaelneale Hi guys , I am doing a project where i need deep research, what is the outlook for this landing in the next release ?

crates/goose/src/providers/formats/openai_responses.rs
```diff
@@ -0,0 +1,697 @@
...
+                response_str.strip_prefix("data: ").unwrap()
+            } else if response_str.starts_with("event: ") {
+                // Skip event type lines
+                continue;
+            } else {
+                // Try to parse as-is in case there's no prefix
+                &response_str
+            };
+
+            // Skip [DONE] marker
```
DOsinga: I can see that

crates/goose/src/providers/formats/openai_responses.rs
```diff
@@ -0,0 +1,697 @@
...
+                ResponsesStreamEvent::OutputItemDone { item, .. } => {
+                    output_items.push(item);
+                }
+
+                ResponsesStreamEvent::OutputTextDone { .. } => {
+                    // Text is already complete from deltas, this is just a summary event
+                }
+
+                ResponsesStreamEvent::ResponseCompleted { response, .. } => {
+                    // Always set final usage (use default if not provided)
```
DOsinga: same

crates/goose/src/providers/openai.rs
```diff
@@ -184,13 +191,25 @@ impl OpenAiProvider {
         })
     }
 
+    fn uses_responses_api(model_name: &str) -> bool {
+        model_name.starts_with("gpt-5-codex") || model_name.starts_with("gpt-5.1-codex")
```
DOsinga, line 195: after your model normalization this might work, but right now, does this cover like databricks? also, people might change the base_path to v1/responses for models that support both

katzdave: I don't actually see the codex models inside of databricks. I gave goose-gpt-5.2 a try and it failed to configure very cryptically with:

Request failed: Bad request (400): {"external_model_provider":"databricks-model-serving","external_model_error":{"error_code":"BAD_REQUEST","message":"BAD_REQUEST: Invalid role in the chat message. The role must be one of: \"system\", \"user\", \"assistant\", \"tool\", \"function\"."}}

Saw there was another thread about that.

Good callout on the base path. We probably need to parse that + also be aware about which models support both.

I did a quick test of the openrouter endpoint I'm using for canonical models and seems like you can infer which models support completions/responses based on the supported_parameters part of the payload.

___________

So I think we follow up with adding something to canonical model around supported_formats.

For the openai provider, if the model supports both, trust the user on the base path. If the model only supports responses, continue to modify the base path suffix for them.

PR #5838, opened 2025-11-21
fix(#5626): handle multi-text content better for openai format models

alexhancock:
We weren't properly handling when messages have multiple text chunks in `providers/formats/openai.rs`

This surfaced in https://github.com/block/goose/issues/5626 when encountering a message with content like this:

```
"content": [
  {
    "type": "text",
    "text": "--- Resource: file:///Users/pjv/development/projects/test%20python%20project/test2.py ---\nfrom typing import NamedTuple\n\n\nclass Person(NamedTuple):\n  name: str\n  age: int | None = None\n\n\nperson = Person(\"Alice\", 25)\n\n---\n"
  },
  {
    "type": "text",
    "text": " what does this code do?"
  }
],
```

Previously when a message contained multiple text content chunks, we would overwrite the earlier ones and replace them with the last one. So this message `what does this code do?` would lead to an answer like `what code?`

**Example**

**Before:**

Screenshot 2025-11-21 at 3 26 02 PM

**After:**

Screenshot 2025-11-21 at 3 29 51 PM

### DIFF
--- a/crates/goose/src/providers/formats/openai.rs
+++ b/crates/goose/src/providers/formats/openai.rs
@@ -63,25 +63,23 @@ pub fn format_messages(messages: &[Message], image_format: &ImageFormat) -> Vec<
         });
 
         let mut output = Vec::new();
+        let mut text_parts = Vec::new();
 
         for content in &message.content {
             match content {
                 MessageContent::Text(text) => {
                     if !text.text.is_empty() {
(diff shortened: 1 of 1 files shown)


alexhancock:
Closing in favor of https://github.com/block/goose/pull/5839 which had some similar code to fix images in acp

PR #5841, opened 2025-11-21
fix sessions coming back with empty messages

zanesq:
## Summary
Noticed that loading sessions from history was coming back with empty messages and not loading the session.

In this [PR](https://github.com/block/goose/pull/5419) `resume_agent` `get_session` parameter `include_messages` was changed to `false` but we do want the messages for the front end to load the session in chat now after [next camp](https://github.com/block/goose/pull/5706)


DOsinga:
sounds like a genuine meta merge conflict

zanesq:
@wpfleger96 np! It worked before the next camp merge. Mostly tagging you to follow up and verify it still works with your original intent in the PR after this change.

PR #5843, opened 2025-11-22
experiment: just in time loading of claude and codex sessions

michaelneale:


crates/goose/src/session/claude_code.rs
```diff
@@ -0,0 +1,273 @@
...
+    cwd: Option<String>,
+    timestamp: Option<String>,
+    message: Option<ClaudeMessageWrapper>,
+}
+
+#[derive(Debug, Deserialize)]
+struct ClaudeMessageWrapper {
+    role: Option<String>,
+    content: serde_json::Value,
+}
```
DOsinga, line 26: are all these fields optional? feels like they are probably pretty required, no?

crates/goose/src/session/claude_code.rs
```diff
@@ -0,0 +1,273 @@
...
+
+pub fn list_claude_code_sessions() -> Result<Vec<(String, PathBuf, DateTime<Utc>)>> {
+    let home = dirs::home_dir().ok_or_else(|| anyhow::anyhow!("No home dir"))?;
```
DOsinga, line 29: we should add home_dir to Paths:: and use that. can in a follow up

crates/goose/src/session/claude_code.rs
```diff
@@ -0,0 +1,273 @@
...
+    let updated_at = latest_timestamp.unwrap_or_else(Utc::now);
+
+    Ok((session_id, PathBuf::from(working_dir), updated_at))
+}
+
+pub fn load_claude_code_session(session_id: &str) -> Result<Conversation> {
+    let home = dirs::home_dir().ok_or_else(|| anyhow::anyhow!("No home dir"))?;
+    let projects_dir = home.join(".claude").join("projects");
+
+    for entry in std::fs::read_dir(projects_dir)? {
```
DOsinga, line 118: this code is the same as on 41; factor our either the iteration or do the dumb thing and fish out the session out of the list

crates/goose/src/session/codex.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    Ok((session_id, PathBuf::from(working_dir), updated_at))
+}
+
+pub fn load_codex_session(session_id: &str) -> Result<Conversation> {
+    let home = dirs::home_dir().ok_or_else(|| anyhow::anyhow!("No home dir"))?;
+    let sessions_dir = home.join(".codex").join("sessions");
+
+    // Walk through all directories to find the session file
+    for year_entry in std::fs::read_dir(&sessions_dir)? {
```
DOsinga, line 154: same as block on 47

ui/desktop/src/components/settings/sessions/SessionSharingSection.tsx
```diff
@@ -150,6 +152,14 @@ export default function SessionSharingSection() {
     }
   };
 
+  const externalSessionsEnabled =
+    config['GOOSE_ENABLE_EXTERNAL_SESSIONS'] === true ||
+    config['GOOSE_ENABLE_EXTERNAL_SESSIONS'] === 'true';
```
DOsinga, line 157: which one is it?

ui/desktop/src/components/settings/sessions/SessionSharingSection.tsx
```diff
@@ -247,6 +257,26 @@ export default function SessionSharingSection() {
           </div>
         </CardContent>
       </Card>
+
+      <Card className="pb-2">
+        <CardHeader className="pb-0">
+          <CardTitle>External Sessions</CardTitle>
+          <CardDescription>
+            Allow goose to access sessions from Claude Code and Codex editors. When enabled, goose
+            can read conversation history from these tools.
```
DOsinga, line 266: what happens if you load those in the new world where we automatically resume the conversation? presumably you can't resume these?

michaelneale: yeah - you can, it imports them (as best it can) into a brand new goose session (so you continue them on in goose).

PR #5847, opened 2025-11-22
[draft] [feat] add term command

baxen:
Term command is a new way to use goose without the builtin REPL
- instead the session is tied right into your terminal

### Type of Change
- [x] Feature

### DIFF
--- a/documentation/docs/guides/terminal-integration.md
+++ b/documentation/docs/guides/terminal-integration.md
@@ -1,3 +1,6 @@
+---
+unlisted: true
+---
 # Terminal Integration
 
 The `goose term` commands let you talk to goose directly from your shell prompt. Instead of switching to a separate REPL session, you stay in your terminal and call goose when you need it.


documentation/docs/guides/terminal-integration.md
```diff
@@ -0,0 +1,75 @@
+# Terminal Integration
+
+The `goose term` commands let you talk to goose directly from your shell prompt. Instead of switching to a separate REPL session, you stay in your terminal and call goose when you need it.
+
+```bash
+gt "what does this error mean?"
+```
+
+Goose responds, you read the answer, and you're back at your prompt. The conversation lives alongside your work, not in a separate window you have to manage.
```
angiejones, line 12:
```suggestion
goose responds, you read the answer, and you're back at your prompt. The conversation lives alongside your work, not in a separate window you have to manage.
```

documentation/docs/guides/terminal-integration.md
```diff
@@ -0,0 +1,75 @@
+# Terminal Integration
+
+The `goose term` commands let you talk to goose directly from your shell prompt. Instead of switching to a separate REPL session, you stay in your terminal and call goose when you need it.
```
blackgirlbytes, line 6: This is such a cool feature, but since it's not released yet, can you make the document unlisted? We can remove it when it gets released.

The way you do that..is in the frontmatter write "unlisted: true"

blackgirlbytes, line 6: This is such a cool feature, but since it's not released yet, can you make the document unlisted? We can remove it when it gets released.

The way you do that..is in the frontmatter write "unlisted: true"

blackgirlbytes: And noting to myself that we should have this mentioned in the CLI commands section

blackgirlbytes (approved):

michaelneale:
nice  - I did note when I ran something which didn't involve a tool call, it didn't always return results... (ie ask "how are you") - if it uses a tool call seems all ok., but intermittent... digging in. 

Also ... 

```zsh
command_not_found_handler() {
  local cmd="$*"
  gt "$cmd"
}
```

I added that to it and works well (even with the eval, automatically, if not too invasive!). 

Very cool and innovative! (I also wonder if we could add a nice little oh my zsh like decoration to know goose is on)

I do think, somehow, dont' want each invocation to show up as a "term..." shell in the session history if possible?

crates/goose/src/session/session_manager.rs
```diff
@@ -19,7 +19,7 @@ use tokio::sync::OnceCell;
 use tracing::{info, warn};
 use utoipa::ToSchema;
 
-const CURRENT_SCHEMA_VERSION: i32 = 6;
+const CURRENT_SCHEMA_VERSION: i32 = 7;
```
michaelneale, line 22: I guess if we are changing the schema - we can make the term sessions not show up the usual way? (wonder if there is a less invasive way).

crates/goose/src/session/session_manager.rs
```diff
@@ -1286,6 +1389,55 @@ impl SessionStorage {
         .execute()
         .await
     }
+
+    async fn add_shell_command(
```
michaelneale, line 1393: I'm confused as to why we need to track these separately? as a special case?

crates/goose/src/session/session_manager.rs
baxen: like having these term: specific sessions not show up in the history for things like the goose desktop? could do!

crates/goose/src/session/session_manager.rs
baxen: this is used by the preexec hook so that goose actually gets a record of the commands that you ran by hand in your terminal between interactions with goose. could make this optional but i find it very useful to not have to tell it what i was doing

crates/goose-cli/src/commands/term.rs
```diff
@@ -0,0 +1,243 @@
...
+/// Ensure a terminal session exists, creating it if necessary
+async fn ensure_terminal_session(
+    session_name: String,
+    working_dir: std::path::PathBuf,
+) -> Result<()> {
+    if SessionManager::get_session(&session_name, false)
+        .await
+        .is_err()
+    {
+        let session = SessionManager::create_session_with_id(
```
DOsinga, line 19: why do we need to create a session with a specific id, why not just use the name of the session here instead? or alternatively, we could create the session the first time and then write that session id to the config file; that way you wouldn't have to set the environment variable either

crates/goose-cli/src/commands/term.rs
```diff
@@ -0,0 +1,243 @@
...
+            .user_provided_name(session_name)
+            .apply()
+            .await?;
+    }
+    Ok(())
+}
+
+/// Handle `goose term init <shell>` - print shell initialization script
+pub fn handle_term_init(shell: &str) -> Result<()> {
+    let terminal_id = Uuid::new_v4().to_string();
```
DOsinga, line 37: actually you could just create the session here and then use the session id of that session as the terminal_id

crates/goose-cli/src/commands/term.rs
```diff
@@ -0,0 +1,243 @@
...
+        Err(_) => return Ok(()), // Silent exit if no terminal ID
+    };
+
+    let session_name = format!("{}{}", TERMINAL_SESSION_PREFIX, terminal_id);
+
+    // Get tokens from session or 0 if none started yet in this terminal
+    let session = SessionManager::get_session(&session_name, false).await.ok();
+    let total_tokens = session.as_ref().and_then(|s| s.total_tokens).unwrap_or(0) as usize;
+
+    let model_name = Config::global()
```
DOsinga, line 204: we store the model and provider now in the session - should probably get it from there

crates/goose-cli/src/commands/term.rs
```diff
@@ -0,0 +1,243 @@
...
+                .await?;
+            session_name.clone()
+        }
+        Err(_) => {
+            ensure_terminal_session(session_name.clone(), working_dir).await?;
+            session_name.clone()
+        }
+    };
+
+    let commands = SessionManager::get_shell_commands_since_last_message(&session_id).await?;
```
DOsinga, line 164: I'm not entirely sure I follow what we are fetching here. doesn't each session add at least one message to the conversation so isn't this then automatically empty?

in general though, adding shell commands to the session manager seems rather specific for this application. couldn't we just list the messages from session and exclude the <shell_history></shell_history> from that to get the same result? or if that doesn't work, maybe add a message content type?

crates/goose/src/session/session_manager.rs
michaelneale: yeah - I have a PR that does that, Session::Hidden is fine.

michaelneale:
my enhancements so far: https://github.com/block/goose/pull/5852

michaelneale:
ok @baxen @DOsinga https://github.com/block/goose/pull/5852 - is my take, this doesn't change session manager, and makes sessions hidden type etc - but seems to work the same if you wan to try it? not sure if the approach for storing the terminal info is better/worse than putting it in sessions, but it is more isolated change.

michaelneale:
all of this has been pulled over here: https://github.com/block/goose/pull/5887

PR #5850, opened 2025-11-23
Cors and token

DOsinga:
Fixes https://github.com/block/goose/security/advisories/GHSA-3cq5-5r4q-jg5w#event-502791

### DIFF
--- a/crates/goose-cli/src/commands/web.rs
+++ b/crates/goose-cli/src/commands/web.rs
@@ -3,7 +3,7 @@ use axum::response::Redirect;
 use axum::{
     extract::{
         ws::{Message, WebSocket, WebSocketUpgrade},
-        Request, State,
+        Query, Request, State,
     },
     http::StatusCode,
     middleware::{self, Next},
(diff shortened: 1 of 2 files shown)


crates/goose-cli/src/commands/web.rs
```diff
@@ -87,17 +88,14 @@ async fn auth_middleware(
     req: Request,
     next: Next,
 ) -> Result<Response, StatusCode> {
-    // Skip auth for health check
     if req.uri().path() == "/api/health" {
         return Ok(next.run(req).await);
     }
 
-    // If no auth token is configured, skip authentication entirely
```
michaelneale, line 95: this is actually a helpful and IMO necessary comment, I don't like just blindly deleting oneline coments.

michaelneale (approved):
seems reasonable (although gets rid of useful comments.

crates/goose-cli/src/commands/web.rs
DOsinga: interestingly enough, goose deleted this comment. which it probably did because I have general settings that it should delete useless comments.

I would still argue that the comment says the same thing as the code below though.

michaelneale: it never ever wants to with mine - I guess it tends to follow "house style" over what system prompt says! @DOsinga sometimes I wonder if we need to put some logic in the editor tool for it to to return an error if it detects single line comment (error can be "this is an inane comment, please either remove and try again or consider if it is really needed) or something?

PR #5852, opened 2025-11-23
feat: enhancing new terminal feature

michaelneale:
Building on: https://github.com/block/goose/pull/5847 

Some small tweaks, naming session, aliases, and more. Doens't use session DB to catch up on missed terminal entries keeping things cleanerl 

Also - you can use `gt type anything you want` - no quotes necessary. 
also just type in `any command you want` and it will send it to goose

This also doesn't use changes to schema or session manager - an alternative approach.

### DIFF
--- a/crates/goose-cli/src/commands/term.rs
+++ b/crates/goose-cli/src/commands/term.rs
@@ -16,13 +16,11 @@ async fn get_or_create_terminal_session(working_dir: PathBuf) -> Result<String> 
         working_dir.to_string_lossy()
     );
 
-    // Find existing session by name
     let sessions = SessionManager::list_sessions().await?;
     if let Some(session) = sessions.iter().find(|s| s.name == session_name) {
         return Ok(session.id.clone());
     }
 
-    // Create new session
     let session =
         SessionManager::create_session(working_dir, session_name.clone(), Default::default())
             .await?;


michaelneale:
some small enhancements: https://github.com/block/goose/pull/5852 will take a look at other feedback

documentation/docs/guides/terminal-integration.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+With this enabled, any command that doesn't exist will automatically be sent to goose:
+
+```bash
+$ analyze_logs
+🪿 Command 'analyze_logs' not found. Asking goose...
+```
+
+Goose will interpret what you meant and either suggest the correct command or help you accomplish the task.
```
blackgirlbytes:
```suggestion
goose will interpret what you meant and either suggest the correct command or help you accomplish the task.
```

crates/goose-cli/src/commands/term.rs
```diff
@@ -0,0 +1,333 @@
...
+    /// Map from working directory path to session ID
+    terminal_sessions: HashMap<String, String>,
+}
+
+impl TerminalConfig {
+    fn config_path() -> Result<PathBuf> {
+        let home = dirs::home_dir().ok_or_else(|| anyhow!("Could not determine home directory"))?;
+        let config_dir = home.join(".config").join("goose");
+        fs::create_dir_all(&config_dir)?;
+        Ok(config_dir.join("term-sessions.json"))
```
DOsinga: use Paths:: to get to the config dir - that way it can be overwritten

michaelneale: yes - I wonder if it should be `state` dir too?

michaelneale:
@DOsinga 

```json
{
  "terminal_sessions": {
    "/Users/micn/Development/goose": "20251124_2",
    "/Users/micn/Documents/code/goose": "20251124_3"
  }
}
``` yeah it just glues it to session id.

crates/goose-cli/src/commands/term.rs
michaelneale: @DOsinga it also needs a short term storage as it captures commands between gt invocations so it can add them in (doesn't need to be in session though - as they will end up in session for historical use, but while goose isnt' running, need to accumulate somewhere)- that make sense?

michaelneale:
@DOsinga @baxen how is this looking? 

I note that `gt` is often used by graphite (a growing popular tool) so I augmented it with "@goose" or "@g" which felt more natural. This is functionally equivalent otherwise, but doesn't change what is in the session.

michaelneale:
work has shifted over here: https://github.com/block/goose/pull/5887/

PR #5853, opened 2025-11-24
urgent: github broke runners for macos

michaelneale:
https://github.com/actions/runner-images/issues/13341
https://github.com/actions/runner/issues/4134

looks like github broke their macos runners


michaelneale:
merging this as it is sunday/monday and is needed

Issue #5752, opened 2025-11-15
When goose makes a web request, hint to the server that markdown is preferred


simonsickle:
**Please explain the motivation behind the feature request.**
I’d like goose to hint to web servers that it prefers markdown, but will also take other content types. This is helpful when requesting / reading docs and is much more token efficient than the agent reading HTML/JS/CSS

**Describe the solution you'd like**
Simply adding Accepts: text/markdown to outgoing requests would be sufficient. 

**Additional context**

https://x.com/bcherny/status/1988860326306087102?s=46

- [x] I have verified this does not duplicate an existing feature request

michaelneale:
thanks @simonsickle: https://github.com/block/goose/pull/5854 here it is!

PR #5854, opened 2025-11-24
chore: suggest using text/markdown when fetching content

michaelneale:
fixes https://github.com/block/goose/issues/5752 - which means it may get clearer content back for agent consonsumption.

Please explain the motivation behind the feature request.
I’d like goose to hint to web servers that it prefers markdown, but will also take other content types. This is helpful when requesting / reading docs and is much more token efficient than the agent reading HTML/JS/CSS

Describe the solution you'd like
Simply adding Accepts: text/markdown to outgoing requests would be sufficient.

Additional context

https://x.com/bcherny/status/1988860326306087102?s=46

(brings it in line with direction of other agents).

### DIFF
--- a/crates/goose-mcp/src/developer/rmcp_developer.rs
+++ b/crates/goose-mcp/src/developer/rmcp_developer.rs
@@ -325,6 +325,8 @@ impl ServerHandler for DeveloperServer {
             **Important**: Each shell command runs in its own process. Things like directory changes or
             sourcing files do not persist between tool calls. So you may need to repeat them each time by
             stringing together commands.
+
+            If fetching web content, consider adding Accept: text/markdown header
         "#};
 
         let windows_specific = indoc! {r#"
--- a/crates/goose-mcp/src/computercontroller/mod.rs
+++ b/crates/goose-mcp/src/computercontroller/mod.rs
@@ -492,13 +492,19 @@ impl ComputerControllerServer {
         let save_as = params.save_as;
 
         // Fetch the content
-        let response = self.http_client.get(url).send().await.map_err(|e| {
-            ErrorData::new(
-                ErrorCode::INTERNAL_ERROR,
-                format!("Failed to fetch URL: {}", e),
-                None,
-            )
-        })?;
+        let response = self
+            .http_client
+            .get(url)
+            .header("Accept", "text/markdown, */*")
+            .send()
+            .await
+            .map_err(|e| {
+                ErrorData::new(
+                    ErrorCode::INTERNAL_ERROR,
+                    format!("Failed to fetch URL: {}", e),
+                    None,
+                )
+            })?;
 
         let status = response.status();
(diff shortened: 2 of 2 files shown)


crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -325,6 +325,8 @@ impl ServerHandler for DeveloperServer {
             **Important**: Each shell command runs in its own process. Things like directory changes or
             sourcing files do not persist between tool calls. So you may need to repeat them each time by
             stringing together commands.
+
+            If fetching web content, consider adding Accept: text/markdown header
```
simonsickle, line 329: Should we add the reason why to add the header? Right now it would take some extra context to know why it’s needed or when it’s appropriate, so maybe inlining that context here is good?

michaelneale: yeah - I suspect so. I am always wary of "adding more" to a prompt that effectively ends up in a system prompt, so am trying to be brief - any ideas of the densest possible way to say that? (I am also not sure if the shell() tool is the right place, but there is no developer tool that is "fetch" - there are a ton of extensions that are, and the computer controller one...). Maybe don't even need it here and over time models will know (via pre-train) to do this when appropriate? (ie don't touch this one, but leave the one in computer controller) @simonsickle ?

simonsickle: I thought about this a bit more. When I have written rules, I typically like to follow the following pattern: (a) short, (b) conditional, (c) with explicit carve-outs

so, my attempt here following that pattern is


“For HTTP GETs that retrieve human-readable documentation or prose (e.g., READMEs, guides, specs, blog posts), include ‘Accept: text/markdown, */*’ to request markdown when available (token-efficient, easy to parse). Do not add this header for API/JSON requests, binary/media downloads, or HTML scraping.”

this tells the LLMs when and why we should use the header but also why we should not use the header (if you ask it to scrape HTML or call a JSON endpoint specifically). 

So we have any cli benchmarks we could test this with?

michaelneale: there is tbench - but still working on getting that a) regular and reliable (it costs about $200+ each run and takes some time) and b) let it run on branches too (may be overkillf or that). That seems a bit much wording for that prompt, would rather it be part of a bigger developer MCP refactoring (ie take things out, add other things). Perhaps it is time to bite the bullet and have a built in fetch tool? (I expect that is now common?)

alexhancock (approved):

crates/goose-mcp/src/developer/rmcp_developer.rs
alexhancock: Interesting discussion here, but in the interest of testing this out I am going to merge with it as-is. We can always iterate on prompts.

PR #5855, opened 2025-11-24
Fix multi tool calling

DOsinga:
## Summary
Fix how we handle multiple tools, we need to interleave requests and responses, not bunch them up


DOsinga:
In the old code we collected all the tool requests in one message and all the tool responses in another one and then appended those to the conversation, which doesn't work on at least databricks

PR #5862, opened 2025-11-24
[draft] [feat] new shell driven extension

baxen:
## Summary

This is an experiment that could replace developer. It didn't work a year ago, but LLMs have come a long way and want to test some now very old assumptions

### Type of Change
- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance


michaelneale:
nice @baxen worth while trying this with something like tbench - what is best way to stretch its wings?

crates/goose-server/src/main.rs
```diff
@@ -56,6 +57,15 @@ async fn main() -> anyhow::Result<()> {
                     )
                     .await?
                 }
+                McpCommand::Shell => {
+                    let bash_env = Paths::config_dir().join(".bash_env");
```
michaelneale, line 62: we are also trying to make it work with users preferred shell - zsh for example

michaelneale:
closing this but keepign branch so we can try to bring it back when code-mode is more mainstream

PR #5868, opened 2025-11-25
Update cleanup marker logic for Fedora users.

sudocarlos:
Update cleanup marker logic for Fedora users.

## Summary

Removes the double quotes from `CLEANUP_MARKER` for Fedora systems experiencing an error `touch: cannot access '~/.config/goose/.mcp-hermit-cleanup-v1': No such file or directory` when trying to add extensions.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing

Manual testing. I made this change in my system to resolve the issue when I encountered it. I verified the same in a fresh Fedora Workstation 43 VM, too. 

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:   

### Submitting a Recipe?

**Email**:

### DIFF
--- a/ui/desktop/src/bin/node-setup-common.sh
+++ b/ui/desktop/src/bin/node-setup-common.sh
@@ -24,7 +24,7 @@ trap 'log "An error occurred. Exiting with status $?."' ERR
 log "Starting node setup (common)."
 
 # One-time cleanup for existing Linux users to fix locking issues
-CLEANUP_MARKER="${HOME}/.config/goose/.mcp-hermit-cleanup-v1"
+CLEANUP_MARKER=${HOME}/.config/goose/.mcp-hermit-cleanup-v1
 if [[ "$(uname -s)" == "Linux" ]] && [ ! -f "$CLEANUP_MARKER" ]; then
     log "Performing one-time cleanup of old mcp-hermit directory to fix locking issues."
     if [ -d ${HOME}/.config/goose/mcp-hermit ]; then


michaelneale:
weird - what is adding double quotes? If that is an issue - could probably remove it for all, this can't just be fedora? (could try single quotes too).

michaelneale:
@sudocarlos if you can do the DCO thing as well, would be appreciated, and can get it in (still very strange! but I have seen weirder things and clearly this worked for you when packaged up).

ui/desktop/src/bin/node-setup-common.sh
```diff
@@ -24,7 +24,11 @@ trap 'log "An error occurred. Exiting with status $?."' ERR
 log "Starting node setup (common)."
 
 # One-time cleanup for existing Linux users to fix locking issues
-CLEANUP_MARKER="~/.config/goose/.mcp-hermit-cleanup-v1"
+if grep -q "Fedora" /etc/os-release; then
```
The-Best-Codes: I don't think `/etc/os-release` exists on macOS, so that will cause an error there. I think it might be better to find what specifically is causing the issue for you on Fedora (is it the shell? the OS? or something else?) and detect that in general to apply the fix.

The-Best-Codes (changes requested):

ui/desktop/src/bin/node-setup-common.sh
michaelneale: I think this path is only for linux though? (and only ever was?)

sudocarlos: I made another change to check if it is a Linux system first. I don't think it's a good idea to omit the quotes for all Linux system, since it is generally best-practice to use the quotes. In the case of Fedora, however, the shell encounters issues with shell expansion.

The-Best-Codes: @sudocarlos Yes, my bad lol. I forgot there's a Linux check right before that script runs. If there wasn't, the script would error on macOS (`grep: /etc/os-release: No such file or directory`) but I think it should work how it is.

My other question is, do we know that this is just an issue with Fedora? Or is it because a specific shell you're using? Or maybe it's because of a broader issue like the base OS (Red Hat)?
We want to find the root cause, not just treat a symptom.

ui/desktop/src/bin/node-setup-common.sh
```diff
@@ -24,45 +24,45 @@ trap 'log "An error occurred. Exiting with status $?."' ERR
 log "Starting node setup (common)."
 
 # One-time cleanup for existing Linux users to fix locking issues
-CLEANUP_MARKER="~/.config/goose/.mcp-hermit-cleanup-v1"
+CLEANUP_MARKER="${HOME}/.config/goose/.mcp-hermit-cleanup-v1"
```
The-Best-Codes: Remove the quotes, it should work fine with them but better without them for consistency.

The-Best-Codes (changes requested):

michaelneale (approved):
I think should be great. thanks all

The-Best-Codes:
Merged @sudocarlos, thank you!

PR #5870, opened 2025-11-25
feat: external session loading (claude and codex)

michaelneale:
This lets you turn on (off by default) loading of claude and codex sessions. It doesn't import them until you click "resume" in goose (then it will run them as a goose session).

### DIFF
--- a/crates/goose-cli/src/session/builder.rs
+++ b/crates/goose-cli/src/session/builder.rs
@@ -355,8 +355,8 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> CliSession {
         session.id
     } else if session_config.resume {
         if let Some(session_id) = session_config.session_id {
-            match SessionManager::get_session(&session_id, false).await {
-                Ok(_) => session_id,
+            match SessionManager::ensure_local_session(&session_id).await {
+                Ok(id) => id,
                 Err(_) => {
                     output::render_error(&format!(
                         "Cannot resume session {} - no such session exists",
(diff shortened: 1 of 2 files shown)


zanesq (approved):
LGTM!

michaelneale:
@codefromthecrypt thanks - yeah still mulling this over: 

I could see a future where something that works exactly like this just drives those other agents via ACP, I could also imagine this porting to using ACP to load/list session (into goose)
And I could imagine a future feature where you can push-button import external sessions into goose up front (via ACP). 

so a lot hinges on ACP (long term ideally am not reading the disk format) longer term, but for now this (and I can add gemini cli) is more a read/just in time import, you think that is still worth while as things are today (and revisit with ACP in future as it matures?)

codefromthecrypt:
Agreed the ACP is basically a direction for the future not today, as they don't have full support for loading (things stuck in PRs etc, and some specs around this are RFD not fully merged yet)

TL;DR; I meant to say there are some kinks we might work out now with the loading/listing as done in this PR. In the future we should try to delete this if ACP overlaps which looks like it might.

michaelneale:
@codefromthecrypt should I merge this or close it in favour of ACP given we agree it is a good approach (I know slightly different implementation and aims - but wonder if keep it simple with just acp?) - if you think this is ok otherwise happy to merge it in too.

codefromthecrypt:
I will use this as a comparison either way, so if you want to merge and i refactor to ACP i think that works too right?

codefromthecrypt:
Gun to head answer: give me the weekend and if ACP session load stuff works out for both claude and Gemini, then we can obviate this one with the other PR. I will be back in your timezone before monday.

michaelneale:
Closing this after discussion with @codefromthecrypt, we are going to look at more direct ACP integration which will allow both this but also run sessions in claude as well.

PR #5871, opened 2025-11-25
Revert "fix: do not load active extensions when no extensions in the recipe"

DOsinga:
Reverts block/goose#5282

after all the discussion, I think it is better to revert this. this breaks the scheduler and a lot of recipes in the wild. let me know what internal systems we are using that would need to respect this.

/cc @jamadeo @tlongwell-block


jsibbison-square:
We can work around it by clearing the extensions from the config.yaml, feel free to revert.

PR #5873, opened 2025-11-25
OpenRouter & Xai streaming

DOsinga:
## Summary

Remove some duplication on how we do streaming and also add it to xai


crates/goose/src/providers/openai.rs
```diff
@@ -238,7 +230,14 @@ impl Provider for OpenAiProvider {
...
         tools: &[Tool],
     ) -> Result<(Message, ProviderUsage), ProviderError> {
-        let payload = create_request(model_config, system, messages, tools, &ImageFormat::OpenAi)?;
+        let payload = create_request(
+            model_config,
+            system,
+            messages,
+            tools,
+            &ImageFormat::OpenAi,
+            false,
```
michaelneale: I wish the rules didn't make this so vertical... ugh...

michaelneale:
code looks good - main worry and maybe need some manual testing, making sure we are accounting for tokens correctly (I know we had past issues with that with anthropic, but probably long ago)

kmk142789 (approved):

zanesq:
@DOsinga still in progress?

PR #5874, opened 2025-11-25
Add tooltips for icons in sidebar and recipes

zanesq:
## Summary
Added tooltips to icons in sidebar and recipes because users have difficulties understanding what these icons do. 

Yes there are too many icons in recipes we will come back to that.

image
image
<br>
image
<br>
image
<br>
image
<br>
image
<br>
image
<br>
image
<br>
image


zanesq:
closing since there doesn't seem to be an appetite for it

PR #5875, opened 2025-11-25
Fix recipe scheduler in recipes view

zanesq:
## Summary
Fixes issue where create/edit schedule from recipes is persisting the original value instead of allowing changes. Also fixed the initial display of the cron picker to show double zeros for minutes.

### DIFF
--- a/ui/desktop/src/components/schedule/CronPicker.tsx
+++ b/ui/desktop/src/components/schedule/CronPicker.tsx
@@ -92,7 +92,7 @@ export const CronPicker: React.FC<CronPickerProps> = ({ schedule, onChange, isVa
   useEffect(() => {
     const parsed = parseCron(schedule?.cron || '');
     setPeriod(parsed.period);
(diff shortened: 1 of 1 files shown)


ui/desktop/src/components/schedule/CronPicker.tsx
```diff
@@ -92,8 +92,8 @@ export const CronPicker: React.FC<CronPickerProps> = ({ schedule, onChange, isVa
   useEffect(() => {
     const parsed = parseCron(schedule?.cron || '');
     setPeriod(parsed.period);
-    setSecond(parsed.second === '*' ? '0' : parsed.second);
-    setMinute(parsed.minute === '*' ? '0' : parsed.minute);
+    setSecond(parsed.second === '*' ? '0' : parsed.second.padStart(2, '0'));
```
DOsinga: should we not double the 00 also?

zanesq:
Actually the persisting issue cropped up after I added the tooltips in https://github.com/block/goose/pull/5874, the tooltip component must be causing it to re-render.

PR #5877, opened 2025-11-25
fix: adjust strange spacing in agent.rs

alexhancock:
Noticed this strange spacing issue in `agent.rs` when working on something else, and corrected it to make the code easier to read

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -927,352 +927,352 @@ impl Agent {
         });
 
         Ok(Box::pin(async_stream::try_stream! {
-                    let _ = reply_span.enter();
-                    let mut turns_taken = 0u32;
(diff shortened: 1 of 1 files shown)


alexhancock:
This is the only actual change, and still passes the formatter 

Screenshot 2025-11-25 at 1 30 10 PM

DOsinga:
I noticed that too, but neither my editor nor cargo fmt would budge

I think the real problem is that reply_internal is too nested and then our line length and format start to conflict. but the streaming setup makes it hard to refactor

alexhancock:
> I noticed that too, but neither my editor nor cargo fmt would budge
> 
> I think the real problem is that reply_internal is too nested and then our line length and format start to conflict. but the streaming setup makes it hard to refactor

this whitespace change makes it I think how it should be and still passes the formatter

so we may not get automatic corrections if it gets messed up again, but seems better than current state?

Issue #5865, opened 2025-11-24
Feature Request: Add session fork functionality


wpfleger96:
**Motivation**

After finishing work in a session, I often want to branch into parallel tasks that share the same context. For example, after implementing a feature, I want to simultaneously write tests (in one session) and docs (in another) - both need the implementation context but should run independently.

Currently this requires re-explaining context in new sessions or working sequentially.

**Proposed Solution**

Add a `--fork` flag:
```bash
goose session --resume --fork
goose session --resume --name <session-name> --fork
goose session --resume --session-id <session-id> --fork
```

Creates a new session with all messages copied from the source, allowing independent continuation.

**Alternatives Considered**

- Manual DB duplication (error-prone, not user-friendly)
- Export/import session JSON (more complex than needed)

**Additional Context**

- Claude Code has this via `--fork-session` - proven UX
- Desktop UI: fork button in session history would be a nice future addition

- [x] I have verified this does not duplicate an existing feature request

DOsinga:
we have the functionality in the desktop (/cc @zanesq) - there's now a primitive to copy a conversation (which it looks like what you want) and a truncate (in the desktop you can edit a message further back in time and then fork from there, or wipe the rest of the conversation and continue with the new message).

wpfleger96:
@DOsinga / @zanesq I took a stab at implementing this in #5882 which builds on top of that copy session primitive - wondering what you think!

PR #5882, opened 2025-11-25
Add session forking capability

wpfleger96:
This PR adds session forking (duplication) across CLI, backend API, and desktop UI with a unified, simplified approach

Based on feedback this PR was simplified from the original implementation

I often do something like this in Claude Code by running `claude --resume <id> --fork-session`:
```
                                                                        /-> (still original): User2 ("write tests for x")
Original: User1 ("implement x for me") -> LLM1 ("ok implemented x") ---|
                                                                        \-> forked: User3 ("write docs for x")
```
After the change in this PR, now I can do in goose by `goose --resume --session-id <id> --fork`. 

### Unified Fork API
- Renamed `edit_message` endpoint → `fork` and added boolean `truncate` and `copy` parameters
- Single endpoint handles all use cases: full copy, edit-in-place, and branch-at-point
- Removed redundant `fork_session()` wrapper (thin wrapper around `copy_session()`) from original PR implementation
- Removed over-engineered UI code: `useForkSession` hook, ChatInput fork button from original PR implementation

### Backend Changes
- `POST /sessions/{id}/fork` with `ForkRequest { truncate: bool, copy: bool, timestamp?: i64 }`
- Validation: `truncate=true` requires `timestamp`, returns structured `ErrorResponse`
- Uses existing `copy_session()` and `truncate_conversation()` primitives
- Fixed `copy_session()` to preserve `provider_name` and `model_config`

### CLI Implementation
- `--fork` flag (requires `--resume`) duplicates sessions
- Calls `copy_session()` directly (no HTTP endpoint needed)
- Simplified session resolution to eliminate duplication

### Desktop UI
- Added "Duplicate" button in session list using `forkSession({ truncate: false, copy: true })`
- Updated existing edit message fork to use new unified API
- Removed fork button from ChatInput and useForkSession hook

## Use Case Matrix

| truncate | copy | Result | Use Case |
|----------|------|--------|----------|
| `false` | `true` | Full session copy | Duplicate for parallel work (my use case for this PR) |
| `true` | `false` | Edit in place | Truncate current session (existing `Edit in Place` UI option) |
| `true` | `true` | Branch at point | Copy + truncate at timestamp (exiting `Fork Session` UI option) |

Addresses #5865


crates/goose-server/src/routes/session.rs
```diff
@@ -390,6 +390,30 @@ async fn edit_message(
...
+        ("api_key" = [])
+    ),
+    tag = "Session Management"
+)]
+async fn fork_session(Path(session_id): Path<String>) -> Result<Json<Session>, StatusCode> {
+    let forked_session = SessionManager::fork_session(&session_id)
+        .await
+        .map_err(|e| {
+            tracing::error!("Failed to fork session: {}", e);
+            StatusCode::INTERNAL_SERVER_ERROR
```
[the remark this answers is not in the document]
wpfleger96: addressed in `ff0c6120012a59c91df27299b200220ecdf9947c`

ui/desktop/openapi.json
```diff
@@ -2124,6 +2124,47 @@
...
+        "operationId": "fork_session",
+        "parameters": [
+          {
+            "name": "session_id",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
+          }
```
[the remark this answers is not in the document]
wpfleger96: addressed in `ff0c6120012a59c91df27299b200220ecdf9947c`

crates/goose/src/session/session_manager.rs
```diff
@@ -1255,6 +1267,11 @@ impl SessionStorage {
         self.get_session(&new_session.id, true).await
     }
 
+    async fn fork_session(&self, session_id: &str) -> Result<Session> {
```
DOsinga: so fork session is the same as copy session, except for that it also copies the name? not sure we need that

crates/goose/src/session/session_manager.rs
```diff
@@ -1492,4 +1509,63 @@ mod tests {
         assert!(imported.user_set_name);
         assert_eq!(imported.working_dir, PathBuf::from("/tmp/test"));
     }
+
+    #[tokio::test]
+    async fn test_fork_session() {
```
DOsinga: this doesn't test anything other that indeed it copies the name. I'd drop it

crates/goose-cli/src/cli.rs
```diff
@@ -75,14 +75,22 @@ jsonl' -> '20250325_200615')."
 async fn get_or_create_session_id(
     identifier: Option<Identifier>,
     resume: bool,
+    fork: bool,
     no_session: bool,
```
DOsinga, line 317: I'm not sure it makes sense to have three booleans here

crates/goose-cli/src/cli.rs
```diff
@@ -101,9 +109,22 @@ async fn get_or_create_session_id(
...
+        } else {
+            Ok(Some(session_id))
+        }
     } else if let Some(name) = id.name {
-        if resume {
+        if fork {
+            let sessions = SessionManager::list_sessions().await?;
+            let source_session = sessions
+                .into_iter()
+                .find(|s| s.name == name || s.id == name)
```
DOsinga: this is duplicated code I think. shouldn't this just reuse all the resume code and then if you fork in the end make a copy of that session

crates/goose-cli/src/session/builder.rs
```diff
@@ -30,6 +30,8 @@ pub struct SessionBuilderConfig {
     pub session_id: Option<String>,
     /// Whether to resume an existing session
     pub resume: bool,
+    /// Whether to fork an existing session (creates new session with copied history)
```
DOsinga: history?

ui/desktop/src/hooks/useForkSession.ts
```diff
@@ -0,0 +1,36 @@
+import { useState } from 'react';
```
DOsinga: I don't think this warrants a hook

ui/desktop/src/hooks/useForkSession.ts
```diff
@@ -0,0 +1,36 @@
...
+  const [isForking, setIsForking] = useState(false);
+
+  const forkAndOpenWindow = async (sessionId: string) => {
+    if (isForking) return;
+    setIsForking(true);
+    try {
+      const response = await forkSession({ path: { session_id: sessionId }, throwOnError: true });
+      const forkedSession = response.data;
+      if (!forkedSession) {
+        throw new Error('Fork response missing session data');
```
DOsinga: why check this if when it fails you get an exception

ui/desktop/src/hooks/useForkSession.ts
```diff
@@ -0,0 +1,36 @@
...
+      }
+      toast.success(`Forked session: ${forkedSession.name}`);
+      window.electron.createChatWindow(
```
DOsinga: no need to create a new window in the new world.

zanesq:
Just reiterating what @DOsinga said the desktop now has fork session support in chat

image

wpfleger96:
@DOsinga @zanesq just simplified the PR based your feedback, what do y'all think now? I named this "fork" since that's what Claude Code calls the equivalent functionality I'm trying to implement, but I think I made it confusing since there's already a "Fork Session" in goose UI. what I want here is really "duplicate" instead of "fork" 

| truncate | copy | Result | Use Case |
|----------|------|--------|----------|
| `false` | `true` | Full session copy | Duplicate for parallel work (my use case for this PR) |
| `true` | `false` | Edit in place | Truncate current session (existing `Edit in Place` UI option) |
| `true` | `true` | Branch at point | Copy + truncate at timestamp (exiting `Fork Session` UI option) |

zanesq:
LGTM Front end changes look good also tested locally in the desktop app and everything is working. I'll defer to @DOsinga for the cli changes

crates/goose-server/src/routes/session.rs
```diff
@@ -336,63 +325,86 @@ async fn import_session(
...
+                }
+            })?;
+
+        copied.id
+    } else {
+        session_id.clone()
+    };
+
+    if request.truncate {
+        SessionManager::truncate_conversation(&target_session_id, request.timestamp.unwrap())
```
DOsinga: copilot is a little pedantic here, but yeah, we might as well just go with or_default(0) or some such

PR #5885, opened 2025-11-26
fix: use a lock to ensure only need to run tunnel just in case multiple go…

michaelneale:
there can be multiple goosed but should only be one tunnel running.

small one - would rather not need to do this, but as things are now, we let many goosed fly (pun intended) so as this is in process, don't really want to double up (not sure what would happen - probably nothing functionally good!)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2871,6 +2871,7 @@ dependencies = [
  "chrono",
  "clap",
  "config",
+ "fs2",
  "futures",
  "goose",
  "goose-mcp",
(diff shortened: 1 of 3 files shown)


michaelneale:
@alexhancock no known port - the only port used is randomly assigned for goosed (at the moment) - I guess if that was persisted it could check if it is still active and use that as a lock? (I don't think the tunnel itself uses a port - it just internally passes from websocket to the temporary port goosed is using).

alexhancock:
> @alexhancock no known port - the only port used is randomly assigned for goosed (at the moment) - I guess if that was persisted it could check if it is still active and use that as a lock? (I don't think the tunnel itself uses a port - it just internally passes from websocket to the temporary port goosed is using).

I figured as much - file based seems fine!

PR #5887, opened 2025-11-26
feat: @goose in terminal (native terminal support)

DOsinga:
## Summary

Douwe's take on this saga :)

 
image

### DIFF
--- a/documentation/docs/guides/terminal-integration.md
+++ b/documentation/docs/guides/terminal-integration.md
@@ -104,7 +104,7 @@ You won't notice any delay. The logging happens asynchronously after your comman
 4. Optionally installs a command-not-found handler (with `--default`)
 
 The hook runs `goose term log <command> &` in the background, which appends to the goose session.
-When you run `@goose`, goose reads commands from goose session any commands that happened since it 
+When you run `@goose`, goose reads from the goose session any commands that happened since it 
 was last called and incorporates them in the next call.
 
 ## Session Management


crates/goose-cli/src/commands/term.rs
```diff
@@ -0,0 +1,267 @@
...
+    let session_id = std::env::var("GOOSE_SESSION_ID").map_err(|_| {
+        anyhow!("GOOSE_SESSION_ID not set. Run 'eval \"$(goose term init <shell>)\"' first.")
+    })?;
+
+    let message = Message::new(
+        Role::User,
+        chrono::Utc::now().timestamp_millis(),
+        vec![MessageContent::text(command)],
+    )
+    .with_metadata(MessageMetadata::user_only());
```
DOsinga, line 190: we could leave this out since we truncate again

crates/goose-cli/src/commands/term.rs
```diff
@@ -0,0 +1,306 @@
...
+            Shell::Zsh => &ZSH_CONFIG,
+            Shell::Fish => &FISH_CONFIG,
+            Shell::Powershell => &POWERSHELL_CONFIG,
+        }
+    }
+}
+
+static BASH_CONFIG: ShellConfig = ShellConfig {
+    script_template: r#"export GOOSE_SESSION_ID="{session_id}"
+alias gt='{goose_bin} term run'
```
baxen: nit: i'd remove this now, but i like having both `@goose` and `@g`

crates/goose-cli/src/commands/term.rs
baxen: do messages have an LLM only state? because these would be weird/spammy if you saw them in a GUI i think

documentation/docs/guides/terminal-integration.md
```diff
@@ -0,0 +1,122 @@
+---
+unlisted: true
+---
+# Terminal Integration
```
baxen, line 4: a todo for future, would be nice to show how to set this up with 
https://starship.rs/
and
https://github.com/ianthehenry/zsh-autoquoter

crates/goose-cli/src/commands/term.rs
michaelneale: if they are like what I saw, isn't end of the world.

crates/goose/src/session/session_manager.rs
```diff
@@ -1121,25 +1128,40 @@ impl SessionStorage {
...
-        )
-            .fetch_all(&self.pool)
+            SELECT s.id, s.working_dir, s.name, s.description, s.user_set_name, s.session_type, s.created_at, s.updated_at, s.extension_data,
+                   s.total_tokens, s.input_tokens, s.output_tokens,
+                   s.accumulated_total_tokens, s.accumulated_input_tokens, s.accumulated_output_tokens,
+                   s.schedule_id, s.recipe_json, s.user_recipe_values_json,
+                   s.provider_name, s.model_config_json,
+                   COUNT(m.id) as message_count
+            FROM sessions s
+            INNER JOIN messages m ON s.id = m.session_id
```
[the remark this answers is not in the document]
michaelneale: @DOsinga this sounds so convincing but I am not so sure (also if it is true, I don't think this is a bad thing... I was wondering how do we get a fresh session)?

michaelneale (approved):

michaelneale:
ok, needs blessing from @blackgirlbytes (as this is a new PR with doc change) and I think we can get this in

blackgirlbytes (approved):

PR #5897, opened 2025-11-26
feat: ActionRequired

alexhancock:
Changing the way tool confirmation requests work to be powered by a more generalized "action required" infrastructure which will scale to other use-cases.

No user facing changes for desktop or CLI. Just reorganizing code.

The next thing that will use this is MCP elicitation support, which I have started in df0bd2f19081b9c1f4374cf5bb1eda36c6987b9b and I have seen enough to know this "action required" system is a good foundation.


michaelneale:
nice @alexhancock, in this case was the copilot feedback on the money?

crates/goose/src/providers/formats/anthropic.rs
```diff
@@ -91,6 +91,9 @@ pub fn format_messages(messages: &[Message]) -> Vec<Value> {
                 MessageContent::ToolConfirmationRequest(_tool_confirmation_request) => {
                     // Skip tool confirmation requests
                 }
+                MessageContent::ActionRequired(_action_required) => {
+                    // Skip action required messages - they're for UI only
```
michaelneale, line 95: this one ignores but other one like bedrock returns a value - expected to be different?

alexhancock: will look later tonight

alexhancock: The bedrock one is more about what the provider gets as content in those cases I believe. It doesn't have any meaningful content either way so I will keep with the pattern in each file matching what `ToolConfirmationRequest` did in each format.

PR #5905, opened 2025-11-27
5527 multiple file popups

benjamindrussell:
## Summary

Added state to track file picker status, disabled the button when file picker open.
Also, not mentioned in the original issue but the same behavior occurs for the directory chooser as well so I went ahead and fixed it.
Added state to track directory chooser status, disabled the button when directory chooser open.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing

Manual Testing

### Related Issues
Relates to #5527 
Discussion: LINK (if any)

Before:  
image

After:
image

image

### Submitting a Recipe?

**Email**:

### DIFF
--- a/ui/desktop/src/components/bottom_menu/DirSwitcher.tsx
+++ b/ui/desktop/src/components/bottom_menu/DirSwitcher.tsx
@@ -8,12 +8,24 @@ interface DirSwitcherProps {
 
 export const DirSwitcher: React.FC<DirSwitcherProps> = ({ className = '' }) => {
   const [isTooltipOpen, setIsTooltipOpen] = useState(false);
+  const [isDirectoryChooserOpen, setIsDirectoryChooserOpen] = useState(false);
 
   const handleDirectoryChange = async () => {
-    window.electron.directoryChooser(true);
+    if (isDirectoryChooserOpen) return;
+    setIsDirectoryChooserOpen(true);
+    try {
+      await window.electron.directoryChooser(true);
+    } finally {
+      setIsDirectoryChooserOpen(false);
+    }
   };
 
   const handleDirectoryClick = async (event: React.MouseEvent) => {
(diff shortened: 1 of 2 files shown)


The-Best-Codes:
FYI, there is a PR (#5640) for the issue already. The author of that PR was the first to begin work on issue #5527, and they are assigned to it.

ui/desktop/src/components/bottom_menu/DirSwitcher.tsx
```diff
@@ -8,12 +8,24 @@ interface DirSwitcherProps {
...
+      setIsDirectoryChooserOpen(false);
+    }
   };
 
   const handleDirectoryClick = async (event: React.MouseEvent) => {
+    if (isDirectoryChooserOpen) {
+      event.preventDefault();
+      event.stopPropagation();
+      return;
+    }
```
The-Best-Codes, line 28: Just out of curiosity, can you explain why just using `return;` won't work here?

benjamindrussell: Yeah, returning only exits the handler, it doesn’t cancel the browser’s default behavior or stop the event from bubbling. We could remove preventDefault and add a type="button" to the button, since the default type is submit. TBH it would probably still work fine with just return.

benjamindrussell:
> FYI, there is a PR (#5640) for the issue already. The author of that PR was the first to begin work on issue #5527, and they are assigned to it.

Ah I see, I'm happy to change the scope of this PR to just fix the behavior of the directory chooser if necessary.

PR #2252, opened 2025-04-17
goose/providers: add Venice.ai private open-source LLM

faces-of-eth:
## Description

Implements Venice.ai as a new model provider with support for various models including:

- llama-3.3-70b (65K context)
- qwen-2.5-coder-32b for code optimization
- qwen-2.5-vl for vision capabilities
- Additional specialized models

The implementation includes:

- Live model list fetching for new model releases
- Full provider API integration
- UI configuration settings
- Improved error handling for model fetching
- Support for Venice.ai's authentication system

## Changes

- Added Venice.ai provider implementation
- Updated provider factory to include Venice.ai
- Added UI components for Venice.ai configuration
- Improved model fetching and error handling
- Added Venice.ai models to the hardcoded list

## Testing

Tested with multiple Venice.ai models including general purpose models and specialized ones for code and vision tasks.

Closes #2239

Screenshot 2025-04-17 at 3 41 13 PM
Screenshot 2025-04-17 at 3 40 09 PM

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210278349818735

PR #5932, opened 2025-12-01
fix: anthropic provider model fetching

tlongwell-block:
looks like this was introduced in https://github.com/block/goose/pull/2252 and may never have worked

reference https://console.anthropic.com/docs/en/api/models/list


katzdave (approved):

alexhancock:
good catch - thanks!

PR #5477, opened 2025-10-30
Integration/pr 4950 5147

spencrmartin:
## Summary

Screenshot 2025-10-30 at 10 25 33 AM

### Type of Change

- [x] Feature
- [ ] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [x] Other (specify below)

### Related Issues
https://github.com/block/goose/pull/4950
https://github.com/block/goose/pull/5147

PR #5147, opened 2025-10-12
Auto detect providers

DOsinga:
Auto detect

PR #5955, opened 2025-12-02
Onboarding detect provider from api key

zanesq:
## Summary
Resurrecting https://github.com/block/goose/pull/5477 from @spencrmartin and integrating the api key detection logic from @DOsinga in https://github.com/block/goose/pull/5147

verified working locally

image

### DIFF
--- a/ui/desktop/src/components/ProviderGuard.tsx
+++ b/ui/desktop/src/components/ProviderGuard.tsx
@@ -213,8 +213,8 @@ export default function ProviderGuard({ didSelectProvider, children }: ProviderG
                   <h1 className="text-2xl sm:text-4xl font-light text-left">Welcome to Goose</h1>
                 </div>
                 <p className="text-text-muted text-base sm:text-lg mt-4 sm:mt-6">
-                  Since it's your first time here, let's get you setup with a provider so we can
-                  make incredible work together.
+                  Since it’s your first time here, let’s get you set up with an AI provider so goose
+                  can work its magic.
                 </p>
               </div>
 


ui/desktop/src/components/ProgressiveMessageList.tsx
```diff
@@ -187,7 +187,7 @@ export default function ProgressiveMessageList({
         if (hasInlineSystemNotification(message)) {
           return (
             <div
-              key={message.id && `${message.id}-${message.content.length}`}
+              key={message.id || `message-${index}-${message.role || 'unknown'}`}
```
zanesq: unrelated fix from @spencrmartin "Resolve React key uniqueness warning in ProgressiveMessageList" looks legit to me

michaelneale:
@zanesq lovely - do you think we can as part of this enhance this so it could even detect from (interactive?) env that it has a named key (when I tried that, when it does work, is just delightful as zero config) - if you like I can look at ways of layering it on with this (as it would require some hand testing to make sure it works packaged vs justfile style) - ie use the -l style shell, sniff out 3 of the common key styles and suggest it automatically? (I can add it to this one if you like - or if you would rather a follow on LMK)

zanesq:
@michaelneale sure please go ahead, this branch is fine.. it sounds like a great enhancement and you would be better suited to test it etc. Can always follow up after this merges if you think it will take a while.

michaelneale:
@zanesq a good start but 2 missing things we need, and I think this should be a priority PR: 

1 - it isn't configuring goose properly - extensions completely empty, even built in ones, so something about how it bootstraps is a very different path: 

image

2 - when it did set it up, say with anthropic, it just picked what i thought was a random model (ie haiku) - not sure of logic of that but should have a recommended one for each maybe? (not sure how that works). 

Otherwise - I really like this, could be a great improvement IMO, just those 2

crates/goose/src/providers/auto_detect.rs
```diff
@@ -0,0 +1,163 @@
...
+    if trimmed_key.starts_with("AIza") {
+        return Some("google");
+    }
+
+    // Groq keys start with gsk_
+    if trimmed_key.starts_with("gsk_") {
+        return Some("groq");
+    }
+
+    // xAI keys start with xai-
```
michaelneale: probably can drop these claude comments!

crates/goose-server/src/routes/config_management.rs
```diff
@@ -131,6 +134,23 @@ pub struct SlashCommandsResponse {
...
+}
+
+#[derive(Serialize, ToSchema)]
+pub struct DetectProviderResponse {
+    pub provider_name: String,
+    pub models: Vec<String>,
+}
+
+#[derive(Serialize, ToSchema)]
+pub struct DetectProviderError {
```
DOsinga: I dont think we are using this

crates/goose/src/providers/auto_detect.rs
```diff
@@ -0,0 +1,163 @@
+use crate::model::ModelConfig;
+
+/// Detect the provider based on API key format
+fn detect_provider_from_key_format(api_key: &str) -> Option<&'static str> {
```
DOsinga: I think we discussed this; let's not do this at all. If the user enters an invalid API key we will just go on and configure the provider with it. also it is rather brittle.

crates/goose/src/providers/auto_detect.rs
```diff
@@ -0,0 +1,163 @@
...
+    if trimmed_key.starts_with("xai-") {
+        return Some("xai");
+    }
+
+    None
+}
+
+/// Test a specific provider with the API key by calling its API
+async fn test_provider(provider_name: &str, api_key: &str) -> Option<(String, Vec<String>)> {
+    let env_key = match provider_name {
```
DOsinga: this is now duplicated code

DOsinga (approved):
Can we revert this to my original implementation? https://github.com/block/goose/pull/5147

ui/desktop/src/components/ProgressiveMessageList.tsx
DOsinga: shouldn't that just be message.id?

ui/desktop/src/components/ApiKeyTester.tsx
```diff
@@ -0,0 +1,232 @@
...
+              onKeyDown={(e) => {
+                if (e.key === 'Enter' && !isLoading && hasInput) {
+                  testApiKey();
+                }
+              }}
+            />
+            <Button
+              onClick={testApiKey}
+              disabled={isLoading || !hasInput}
+              variant={hasInput && !isLoading ? 'default' : 'secondary'}
```
DOsinga: disabled and variant follow the same logic differently formulated, let's make that easier to read

ui/desktop/src/components/ApiKeyTester.tsx
```diff
@@ -0,0 +1,232 @@
...
+  model?: string;
+  totalModels?: number;
+  error?: string;
+  suggestions?: string[];
+}
+
+export default function ApiKeyTester({ onSuccess, onStartTesting }: ApiKeyTesterProps) {
+  const [apiKey, setApiKey] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [testResults, setTestResults] = useState<TestResult[]>([]);
```
DOsinga: this shouldn't be a list, we only ever have one result

ui/desktop/src/components/ApiKeyTester.tsx
```diff
@@ -0,0 +1,232 @@
...
+          success: false,
+          error: 'Could not detect provider from API key',
+          suggestions: [
+            'Make sure you are using a valid API key from OpenAI, Anthropic, Google, Groq, or xAI',
+            'Check that the key is complete and not truncated',
+            'Verify your API key is active and has sufficient credits',
+            'For local Ollama setup, use the "Other Providers" section below',
+          ],
+        },
+      ]);
```
DOsinga: If you are going to ignore the result here, no need to set all this into the result; just keep the result from the API around and have an error boolean

ui/desktop/src/components/ApiKeyTester.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+      toastService.error({
+        title: 'Detection Failed',
+        msg: 'Could not detect provider. Please check your API key and try again.',
+        traceback: error instanceof Error ? error.stack || '' : '',
```
DOsinga: this doesn't strike me as particularly useful - showing a toast while also displaying the error in the UI. pick on

ui/desktop/src/components/ProviderGuard.tsx
```diff
@@ -280,103 +209,64 @@ export default function ProviderGuard({ didSelectProvider, children }: ProviderG
                 </div>
                 <p className="text-text-muted text-base sm:text-lg mt-4 sm:mt-6">
                   Since it's your first time here, let's get you setup with a provider so we can
-                  make incredible work together. Scroll down to see options.
+                  make incredible work together.
```
DOsinga: can we fix this text too?

ui/desktop/src/components/ApiKeyTester.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+        await upsert('GOOSE_PROVIDER', provider_name, false);
+        await upsert('GOOSE_MODEL', models[0], false);
+
+        toastService.success({
+          title: 'Success!',
+          msg: `Configured ${provider_name} with ${models.length} models available`,
+        });
+
+        onSuccess(provider_name, models[0]);
```
DOsinga: the caller should do the config update. also deriving the API key name here from the provider seems fragile; would be better to return that from the backend since that actually tests something

ui/desktop/src/components/ProgressiveMessageList.tsx
zanesq: id can be undefined based on the type.. I think its leftover from when we used to modify messages front end. I'll leave it for now until we can come back to the type and check that its now enforced everywhere. Updated to use message timestamp also.

ui/desktop/src/components/ProviderGuard.tsx
zanesq: what do you think the text should be?

ui/desktop/src/components/ProgressiveMessageList.tsx
DOsinga: It needs to be optional since messages only get their id once they get written to the database. if we want to keep this in, why the double || and the appending of the message role?

zanesq: Got it, I simplified it to ```key={message.id ?? `msg-${index}-${message.created}`}``` in the latest commit 👍

ui/desktop/src/components/ProviderGuard.tsx
DOsinga: drop the scroll down, this should all fix the current window. setup is two words. I don't know the user knows what a provider is. so maybe:

Since it’s your first time here, let’s get you set up with an AI provider so goose can work its magic.

ui/desktop/src/components/ProgressiveMessageList.tsx
DOsinga: 👍

zanesq:
@michaelneale good catch I was able to replicate the missing extensions issue intermittently using a custom `GOOSE_PATH_ROOT` both with the api key detection and with a custom provider, and the extensions show on refresh sometimes and after re-launching. I think its a pre-existing issue since we didn't touch any related config code in this branch and we aren't doing anything special other than what we did before by loading the provider and model and redirecting to home. I will look into that in another PR.

Regarding the preferred model choice, agreed it would be nice addition. I will look into that in a separate PR. Let me know if you have some ideas of how we could do that. Since it would need to be a hard coded list that might not match user preferences I think the best option would be to show a model chooser dialog when onboarding after selecting a provider?

DOsinga:
> Regarding the preferred model choice, agreed it would be nice addition. I will look into that in a separate PR. Let me know if you have some ideas of how we could do that. Since it would need to be a hard coded list that might not match user preferences I think the best option would be to show a model chooser dialog when onboarding after selecting a provider?

the detection code does return a list of models already, so it shouldn't be too hard to show that in a dropdown after detection maybe?

zanesq:
Yeah was thinking not difficult just where it would fit in the flow, should probably do it for configure providers also not just the auto key detection so basically after any provider is configured for the first time.. will see what I can come up with real quick

zanesq:
@DOsinga I think it works well! https://github.com/block/goose/pull/6005

PR #5967, opened 2025-12-03
fix: `final assistant content cannot end with trailing whitespace` error from Anthropic

tlongwell-block:


crates/goose/src/conversation/mod.rs
```diff
@@ -257,6 +258,32 @@ fn merge_text_content_items(messages: Vec<Message>) -> (Vec<Message>, Vec<String
     )
 }
 
+fn trim_assistant_text_whitespace(messages: Vec<Message>) -> (Vec<Message>, Vec<String>) {
```
alexhancock, line 261: Interesting. I might just add a `fn trim` which trims both leading and trailing whitespace of all messages. Would be simpler, and in theory doesn't cause any issues. Only makes messages smaller.

tlongwell-block: Okay, looked into this a bit and tried out some edge cases. We could get into trouble with messages that contain things like ascii art, tables, charts, or other space-dependent content in user message. So, I'm just going to leave this as trimming only the trailing whitespace on assistant messages

tlongwell-block: I could limit the trimming to only the very latest assistant message since that appears to be the only one Anthropic cares about per that message. Any thoughts there @alexhancock ?

DOsinga: this stops things from crashing so that's good and agree, leading whitespace you can't trim in general. do we know where the issue comes from though? it would be better if we could fix it at the source.

tlongwell-block: If the error is to be believed at face value, it's an assistant message causing the issue.

So that rules out tool responses, the moim, and special/control input from the user/client. So my thought is that the LLM itself is generating a trailing whitespace for some reason? They do get weird sometimes once context gets long enough.

tool call seems likely and fits with what I was seeing

DOsinga (approved):

Issue #4115, opened 2025-08-15
Remote agent control


gabeosx:
**Please explain the motivation behind the feature request.**
I have deployed Goose CLI on a remote Linux environment. I would like to be able to control the agent using the GUI on my desktop, rather than through a terminal on SSH

**Describe the solution you'd like**
The ability to deploy Goose in 1 or more locations and control it centrally from the desktop application.

- [x] I have verified this does not duplicate an existing feature request

DOsinga:
does this do the trick #5061

zanesq:
@michaelneale do we have support for this? I see that PR was closed.

michaelneale:
https://github.com/michaelneale/goose-ios 

along with: https://github.com/block/goose/pull/5251 

will do it with a tunnelling system. 
Don't be put off by ios - could be any client to "goosed" (desktop doesn't yet do remote access though but could?)

woudl this be along the lines of what you are looking for @gabeosx ?

michaelneale:
@zanesq yeah moved to a new PR - would like to get that one merged in next so it is in mainstream

michaelneale:
I'll see about letting the electron client connect to a remote as part of this

petermchale:
I work on highly regulated genomics data that sits in a protected environment. Being able to run a desktop on a local computer that is connected via ssh to some sort of 'goose server" running in the protected environment would be helpful. For example, I could loop at data visualizations in the desktop app, which would be impossible via the goose CLI in a terminal.

michaelneale:
@petermchale excellent use case for that, yeah I think doable, taking a look. @zanesq no support yet, as wasn't happy with old approach, taking another run.

michaelneale:
@petermchale @gabeosx and @zanesq 

please take a look: https://github.com/block/goose/pull/5978 

<img width="1886" height="784" alt="Image" src="https://github.com/user-attachments/assets/d91f3b38-e73d-4936-8d8f-8957d1d1619a" />


you can then run a goose server with: 
```
./target/release/goosed agent
```

and use the defaults in the form (127.0.0.1, port 3000, secret is 'test')


Would this meet the need as I understand it?

PR #5978, opened 2025-12-04
feat: external goosed server

michaelneale:
fixes: https://github.com/block/goose/issues/4115 which has been requested a few times. Let the goose app connect to any goose server.

example motivation: 

```
I work on highly regulated genomics data that sits in a protected environment. Being able to run a desktop on a local computer that is connected via ssh to some sort of 'goose server" running in the protected environment would be helpful. For example, I could loop at data visualizations in the desktop app, which would be impossible via the goose CLI in a terminal.
```

image

image

If you accidentally start it and it can't connect: 

Image

you can then run a goose server with: 
```
./target/release/goosed agent
```

and use the defaults in the form (127.0.0.1, port 3000, secret is 'test')

### DIFF
--- a/ui/desktop/src/components/settings/app/ExternalBackendSection.tsx
+++ b/ui/desktop/src/components/settings/app/ExternalBackendSection.tsx
@@ -96,9 +96,10 @@ export default function ExternalBackendSection() {
   };
 
   return (
-    <Card className="rounded-lg">
-      <CardHeader className="pb-0">
-        <CardTitle className="mb-1">Goose Server</CardTitle>
+    <section id="external-backend" className="space-y-4 pr-4 mt-1">
+      <Card className="pb-2">
+        <CardHeader className="pb-0">
+          <CardTitle>Goose Server</CardTitle>
         <CardDescription>
           By default goose launches a server for you, use this to connect to an external goose
           server
@@ -174,5 +175,6 @@ export default function ExternalBackendSection() {
         )}
       </CardContent>
     </Card>
+    </section>
   );
 }
--- a/ui/desktop/src/components/MCPUIResourceRenderer.tsx
+++ b/ui/desktop/src/components/MCPUIResourceRenderer.tsx
@@ -100,10 +100,10 @@ export default function MCPUIResourceRenderer({
 
     const fetchProxyUrl = async () => {
       try {
-        const baseUrl = await window.electron.getGoosedHostPort();
+        const gooseApiHost = await window.electron.getGoosedHostPort();
         const secretKey = await window.electron.getSecretKey();
-        if (baseUrl && secretKey) {
-          setProxyUrl(`${baseUrl}/mcp-ui-proxy?secret=${encodeURIComponent(secretKey)}`);
+        if (gooseApiHost && secretKey) {
+          setProxyUrl(`${gooseApiHost}/mcp-ui-proxy?secret=${encodeURIComponent(secretKey)}`);
         } else {
           console.error('Failed to get goosed host/port or secret key');
         }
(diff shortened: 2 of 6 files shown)


ui/desktop/src/main.ts
```diff
@@ -1763,38 +1813,47 @@ async function appMain() {
     }
   });
 
+  const buildConnectSrc = (): string => {
+    const sources = [
```
michaelneale, line 1815: this is for CSP when the goose-server is remote - or an attempt to cc @aharvard

ui/desktop/src/main.ts
```diff
@@ -567,24 +578,52 @@ const createChat = async (
...
+          ...settings,
+          externalGoosed: {
+            enabled: false,
+            url: settings.externalGoosed?.url || '',
+            secret: settings.externalGoosed?.secret || '',
+          },
+        };
+        saveSettings(updatedSettings);
+        mainWindow.destroy();
+        return createChat(app, initialMessage, dir);
```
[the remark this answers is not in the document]
michaelneale: I don't think this comment makes any sense, it is a fall back

ui/desktop/src/main.ts
```diff
@@ -534,14 +545,14 @@ const createChat = async (
...
-      // Enable features needed for Web Speech API
       webSecurity: true,
       nodeIntegration: false,
       contextIsolation: true,
       additionalArguments: [
         JSON.stringify({
           ...appConfig,
-          GOOSE_PORT: port,
+          GOOSE_API_HOST: baseUrl,
+          GOOSE_PORT: 0,
```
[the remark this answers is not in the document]
DOsinga: this seems like a good point. We should make sure we're not using GOOSE_PORT anywhere except for when starting goosed. also, we're calling it GOOSE_API_HOST here instead of baseUrl. I like that better. how would you feel about renaming it like that everywhere?

ui/desktop/src/utils/settings.ts
```diff
@@ -7,11 +7,18 @@ export interface EnvToggles {
...
+  url: string;
+  secret: string;
+}
+
 export interface Settings {
   envToggles: EnvToggles;
   showMenuBarIcon: boolean;
   showDockIcon: boolean;
   enableWakelock: boolean;
+  externalGoosed?: ExternalGoosedConfig;
```
DOsinga, line 21: this file really needs to die since it keeps its own settings outside of settings that keep centrally in the server. of course this is the exception since you need these settings before you can connect to the server. thoughts?

ui/desktop/src/components/settings/app/ExternalBackendSection.tsx
```diff
@@ -0,0 +1,180 @@
...
+  useEffect(() => {
+    loadSettings();
+  }, []);
+
+  const loadSettings = async () => {
+    const settings = (await window.electron.getSettings()) as Settings | null;
+    if (settings?.externalGoosed) {
+      setEnabled(settings.externalGoosed.enabled);
+      setUrl(settings.externalGoosed.url || '');
+      setSecret(settings.externalGoosed.secret || '');
```
DOsinga: you have one useState per setting element here. might be better to just use the entire object and then set it up in a way that if you forget to set one, it errors out

you could then also have one method to change a field in here and save the blob back to the settings

ui/desktop/src/utils/settings.ts
michaelneale: yes - although I don't think as part of this change. we need to probably rethink how everything is launched and flip it so that something else bootstraps (although I think that would be an extra world of hurt to launch a rust binary which then launches the electron stuff - vs other way around, it likely adds a lot of platform specific code which electron is doing right now - if we add a new bootstrap crate perhaps then electron could bootstrap the bootstrap etc, so entry point the same but order is different). Also some settings here are very specific to front end/electron only so that is probably ok.

ui/desktop/src/main.ts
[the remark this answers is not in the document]
michaelneale: yeah I think that makes sense

PR #6000, opened 2025-12-05
Fix PATH on Windows for extensions

jamadeo:


crates/goose-server/src/routes/agent.rs
```diff
@@ -482,78 +482,6 @@ async fn agent_add_extension(
     State(state): State<Arc<AppState>>,
     Json(request): Json<AddExtensionRequest>,
 ) -> Result<StatusCode, ErrorResponse> {
-    #[cfg(windows)]
```
DOsinga, line 507: hurray

PR #6005, opened 2025-12-05
Show model selector after configuring a provider

zanesq:
## Summary
Show model selector after selecting a provider in welcome or the provider grid. Follow up for https://github.com/block/goose/pull/5955

Updated the copy and icon in the selector to make more sense and moved the doc link to the bottom because it was too close / interfering with the provider dropdown click.

Defaults provider selector to the chosen provider before launch.

Verified locally from all provider setup paths.

Screenshot 2025-12-05 at 2 31 25 PM
Screenshot 2025-12-05 at 2 30 38 PM

### DIFF
--- a/ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
+++ b/ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
@@ -20,6 +20,44 @@ import Model, { getProviderMetadata, fetchModelsForProviders } from '../modelInt
 import { getPredefinedModelsFromEnv, shouldShowPredefinedModels } from '../predefinedModelsUtils';
 import { ProviderType } from '../../../../api';
 
+const PREFERRED_MODEL_PATTERNS = [
+  /claude-sonnet-4/i,
+  /claude-4/i,
+  /gpt-4o(?!-mini)/i,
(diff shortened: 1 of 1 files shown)


ui/desktop/src/components/ProviderGuard.tsx
```diff
@@ -24,6 +26,10 @@ export default function ProviderGuard({ didSelectProvider, children }: ProviderG
   const [showFirstTimeSetup, setShowFirstTimeSetup] = useState(false);
   const [showOllamaSetup, setShowOllamaSetup] = useState(false);
   const [userInActiveSetup, setUserInActiveSetup] = useState(false);
+  const [showSwitchModelModal, setShowSwitchModelModal] = useState(false);
```
DOsinga, line 29: these don't strike me as independent booleans, might be good at some point to change them into an annotated enum like thing

zanesq:
@DOsinga we already have logic to set it to the first one without the model selector. Do you want me to set it behind the scenes to the first one or just the model selector? Regarding the preferred ones, I think it might be difficult to manage that list (hard coded, can get out of date etc) but let me know if you have ideas on that.

zanesq:
@DOsinga this work for now? https://github.com/block/goose/pull/6018

PR #6021, opened 2025-12-08
Hide advanced recipe options under expandable content

zanesq:
## Summary
Visually simplify the recipe form by hiding the advanced options under expandable content

Expands automatically if there are advanced values when editing a recipe.

Also fixed a minor bug where the recipe icon in bottom bar was showing in an empty chat with no messages and improved the text a bit to not have ellipsis everywhere.

Added a link to the recipe docs also.

image

image

### DIFF
--- a/ui/desktop/src/components/recipes/CreateEditRecipeModal.tsx
+++ b/ui/desktop/src/components/recipes/CreateEditRecipeModal.tsx
@@ -1,9 +1,9 @@
 import React, { useState, useEffect, useCallback } from 'react';
 import { useForm } from '@tanstack/react-form';
 import { Recipe, generateDeepLink, Parameter } from '../../recipe';
+import { Check, ExternalLink, Play, Save, X } from 'lucide-react';
 import { Geese } from '../icons/Geese';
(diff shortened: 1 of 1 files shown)


ui/desktop/src/components/ChatInput.tsx
```diff
@@ -1507,7 +1507,7 @@ export default function ChatInput({
               <BottomMenuExtensionSelection sessionId={sessionId} />
             </>
           )}
-          {sessionId && (
+          {sessionId && messages.length > 0 && (
```
michaelneale, line 1510: oh - what would this do before?

zanesq: It would show an empty recipe if you clicked it with no recipes.

I ended up pushing this [change](https://github.com/block/goose/pull/6022) separately for the upcoming release but it wasn't a live bug in a current release, was on main from a previous refactor to move the recipe icon in the bottom bar.

Whitewolf111 (approved):
Hi

Whitewolf111:
-yj4dno

PR #6024, opened 2025-12-09
fix: metrics on posthog

michaelneale:
This records anon metrics on session create (vs on start) to https://posthog.com/

for this to work, needs to have opt in/out for the setting, and wording changes around what data goes where
(this is similar to what other OSS projects do)

**Opt-out:** `GOOSE_TELEMETRY_OFF=1`

@zanesq if we are to use this - will need to opt out of course (separate to OTEL in this case).


zanesq (approved):
verified working!

PR #6030, opened 2025-12-09
feat(mcp): platform extension for "code mode" MCP tool calling

alexhancock:
Implements the idea of "code mode"  or "sandbox mode" for MCP

**Refs**
https://blog.cloudflare.com/code-mode/
https://www.anthropic.com/engineering/code-execution-with-mcp
https://github.com/block/goose/discussions/5899

**Architecture**
* New `code_execution` platform extension
* When enabled, this extension makes all other tools invisible to the model in the traditional sense
* Model now has two tools `read_module` to read a tool's implementation code to know how to call it, and `execute_code` to send code to run to call tool(s)
* Generates a programmatic API to all enabled MCP server tools
* Has two tools
  * `read_module` with the ability to read the source code implementing one tool call
  * `execute_code` with instructions to the model on how it should write code
* Publishes the tree of modules available in format `servers/:server_name/:tool_name.js` to the model via `get_moim`
* Dispatches tool calls in a separate async thread, as the main thread running boa NativeFunctions are `!Send`

**Diagram for tool call dispatching in present state**
tool-calling

### DIFF
--- a/crates/goose/src/agents/code_execution_extension.rs
+++ b/crates/goose/src/agents/code_execution_extension.rs
@@ -39,8 +39,10 @@ struct ExecuteCodeParams {
 
 #[derive(Debug, Serialize, Deserialize, JsonSchema)]
 struct ReadModuleParams {
-    /// Module path: "server" for all tools, "server/tool" for one tool.
-    path: String,
+    /// Module path format:
+    /// - For entire server: "server_name"
+    /// - For specific tool: "server_name/tool_name"
+    module_path: String,
 }
 
 #[derive(Debug, Serialize, Deserialize, JsonSchema)]
@@ -59,6 +61,14 @@ enum SearchTerms {
     Multiple(Vec<String>),
 }
 
+#[derive(Debug, Default, Deserialize)]
+struct InputSchema {
+    #[serde(default)]
+    properties: BTreeMap<String, Value>,
+    #[serde(default)]
+    required: Vec<String>,
+}
+
 struct ToolInfo {
     server_name: String,
     tool_name: String,
@@ -72,19 +82,20 @@ impl ToolInfo {
         let (server_name, tool_name) = tool.name.as_ref().split_once("__")?;
         let param_names = get_parameter_names(tool);
 
+        let schema: InputSchema =
+            serde_json::from_value(Value::Object(tool.input_schema.as_ref().clone()))
+                .unwrap_or_default();
+
         let params = param_names
             .iter()
             .map(|name| {
-                let schema = &tool.input_schema;
-                let prop = schema
-                    .get("properties")
-                    .and_then(|p| p.get(name))
-                    .and_then(|v| v.as_object());
-                let required = schema
-                    .get("required")
-                    .and_then(|r| r.as_array())
-                    .is_some_and(|arr| arr.iter().any(|v| v.as_str() == Some(name)));
-                let ty = prop.and_then(|p| p.get("type")?.as_str()).unwrap_or("any");
+                let ty = schema
+                    .properties
+                    .get(name)
+                    .and_then(|p| p.get("type"))
+                    .and_then(|t| t.as_str())
+                    .unwrap_or("any");
+                let required = schema.required.contains(name);
                 (name.clone(), ty.to_string(), required)
             })
             .collect();
@@ -120,15 +131,15 @@ thread_local! {
 }
 
 fn create_server_module(server_tools: &[&ToolInfo], ctx: &mut Context) -> Module {
-    let export_names: Vec<JsString> = server_tools
+    let (export_names, tool_data): (Vec<JsString>, Vec<(String, String)>) = server_tools
         .iter()
-        .map(|t| js_string!(t.tool_name.as_str()))
-        .collect();
-
-    let tool_data: Vec<(String, String)> = server_tools
-        .iter()
-        .map(|t| (t.tool_name.clone(), t.full_name.clone()))
-        .collect();
+        .map(|t| {
+            (
+                js_string!(t.tool_name.as_str()),
+                (t.tool_name.clone(), t.full_name.clone()),
+            )
+        })
+        .unzip();
 
     Module::synthetic(
         &export_names,
@@ -149,7 +160,7 @@ fn create_server_module(server_tools: &[&ToolInfo], ctx: &mut Context) -> Module
     )
 }
 
-fn create_tool_function(full_name: String) -> NativeFunction {
+fn create_tool_function(full_tool_name: String) -> NativeFunction {
     NativeFunction::from_copy_closure_with_captures(
         |_this, args, full_name: &String, ctx| {
             let args_json = args
@@ -178,7 +189,7 @@ fn create_tool_function(full_name: String) -> NativeFunction {
                 .map(|result| JsValue::from(js_string!(result.as_str())))
                 .map_err(|e| JsNativeError::error().with_message(e).into())
         },
-        full_name,
+        full_tool_name,
     )
 }
 
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -26,7 +26,6 @@ use tokio_stream::wrappers::ReceiverStream;
 use tokio_util::sync::CancellationToken;
 use tracing::{error, warn};
 
-use super::code_execution_extension::EXTENSION_NAME as CODE_EXECUTION_NAME;
 use super::extension::{
     ExtensionConfig, ExtensionError, ExtensionInfo, ExtensionResult, PlatformExtensionContext,
     ToolInfo, PLATFORM_EXTENSIONS,
@@ -599,25 +598,12 @@ impl ExtensionManager {
             .insert(name, Extension::new(config, client, info, temp_dir));
     }
 
-    /// Get extensions info for building the system prompt.
-    /// When code_execution is enabled, it routes tools from other extensions through its
-    /// execute_code interface, so those extensions should not appear as separate sections
-    /// in the system prompt (their instructions would be irrelevant/confusing).
+    /// Get extensions info for building the system prompt
     pub async fn get_extensions_info(&self) -> Vec<ExtensionInfo> {
-        let extensions = self.extensions.lock().await;
-        let code_exec_enabled = extensions.contains_key(CODE_EXECUTION_NAME);
-
-        extensions
+        self.extensions
+            .lock()
+            .await
             .iter()
-            .filter(|(name, _)| {
-                if code_exec_enabled {
-                    // When code_execution is enabled, only show code_execution itself
-                    // Other extensions' tools are accessed via code_execution's execute_code
-                    name.as_str() == CODE_EXECUTION_NAME
-                } else {
-                    true
-                }
-            })
             .map(|(name, ext)| {
                 ExtensionInfo::new(
                     name,
--- a/crates/goose/src/agents/prompt_manager.rs
+++ b/crates/goose/src/agents/prompt_manager.rs
@@ -55,6 +55,7 @@ pub struct SystemPromptBuilder<'a, M> {
     router_enabled: bool,
     subagents_enabled: bool,
     hints: Option<String>,
+    code_execution_mode: bool,
 }
 
 impl<'a> SystemPromptBuilder<'a, PromptManager> {
@@ -86,6 +87,11 @@ impl<'a> SystemPromptBuilder<'a, PromptManager> {
 
     pub fn with_router_enabled(mut self, enabled: bool) -> Self {
         self.router_enabled = enabled;
+        self
+    }
+
+    pub fn with_code_execution_mode(mut self, enabled: bool) -> Self {
+        self.code_execution_mode = enabled;
         self
     }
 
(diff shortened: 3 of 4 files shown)


alexhancock:
This implementation saves on intermediate tool results not flowing to the model, but doesn't yet address progressive discovery of the interfaces themselves (via a tree of files, resources, etc). I will look into this tomorrow and push an update.

michaelneale:
nice - also for compatibility, in this mode I think `enabled: true` in the config just means that it is available to the code mode environment, not the LLM (functionally the same, but implemented differently). can even start trying what it is like with all possible extensions "turned on" to see how well it works and how efficient it can be!

domdomegg:
> This implementation saves on intermediate tool results not flowing to the model, but doesn't yet address progressive discovery of the interfaces themselves (via a tree of files, resources, etc). I will look into this tomorrow and push an update.

I think for this we've found one of the following works well:
- real or virtual filesystem, with servers as folders and tools as files
- telling the model in system prompt what servers are available (and maybe tool names), and then having a tool that is like `get_tools(server_name)` and `get_tool(server_name, tool_name)`

crates/goose/src/agents/extension_manager.rs
```diff
@@ -598,12 +599,25 @@ impl ExtensionManager {
             .insert(name, Extension::new(config, client, info, temp_dir));
     }
 
-    /// Get extensions info
+    /// Get extensions info for building the system prompt.
```
DOsinga: we also call this function for some reason when generating a recipe. either way, do we need to modify this at all? to me it looks like if we have code execution on, we don't even insert the extensions in the prompt (or we should).

crates/goose/src/agents/extension_manager.rs
```diff
@@ -1249,16 +1285,26 @@ impl ExtensionManager {
...
-        for (name, extension) in extensions.iter() {
-            if let ExtensionConfig::Platform { .. } = &extension.config {
-                let client = extension.get_client();
-                let client_guard = client.lock().await;
-                if let Some(moim_content) = client_guard.get_moim().await {
-                    tracing::debug!("MOIM content from {}: {} chars", name, moim_content.len());
-                    content.push('\n');
-                    content.push_str(&moim_content);
-                }
+        let platform_clients: Vec<(String, McpClientBox)> = {
```
DOsinga, line 1274: presmably this is to avoid hanging on to the lock? nice.

does MOIM work with code execution?

crates/goose/src/agents/extension_manager.rs
```diff
@@ -598,12 +599,25 @@ impl ExtensionManager {
...
+    /// Get extensions info for building the system prompt.
+    /// When code_execution is enabled, it routes tools from other extensions through its
+    /// execute_code interface, so those extensions should not appear as separate sections
+    /// in the system prompt (their instructions would be irrelevant/confusing).
     pub async fn get_extensions_info(&self) -> Vec<ExtensionInfo> {
-        self.extensions
-            .lock()
-            .await
+        let extensions = self.extensions.lock().await;
+        let code_exec_enabled = extensions.contains_key(CODE_EXECUTION_NAME);
```
DOsinga: we try to determine this in three different ways, we should probably only use the utility function below.

crates/goose/src/agents/prompt_manager.rs
```diff
@@ -138,6 +139,11 @@ impl<'a> SystemPromptBuilder<'a, PromptManager> {
             })
             .collect();
 
+        // Detect code_execution mode: when enabled, only code_execution extension is passed
```
DOsinga: why do we need to detect that here? arent we injecting this in the builder?

crates/goose/src/agents/extension_manager.rs
alexhancock: > presmably this is to avoid hanging on to the lock? nice.

yes, exactly

alexhancock: can you clarify the question about moim? i use it in a certain way for this change, but want to make sure I answer the right question.

DOsinga: let me have a closer look at the rest of the code. I mostly wondered if we should block the Moim for the rest of the extensions or not.

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -0,0 +1,905 @@
...
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema)]
+struct ExecuteCodeParams {
+    /// JavaScript code with ES6 imports for MCP tools.
+    code: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema)]
+struct ReadModuleParams {
+    /// Module path: "server" for all tools, "server/tool" for one tool.
```
DOsinga: I can't believe I am saying this, but assuming this comment becomes a doc string in the tool, can we expand this a bit better? I had to read it twice before I understood it.

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -0,0 +1,905 @@
...
+    path: String,
```
DOsinga: and maybe just call this module_path then

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -0,0 +1,905 @@
...
+                    .and_then(|p| p.get(name))
+                    .and_then(|v| v.as_object());
+                let required = schema
+                    .get("required")
+                    .and_then(|r| r.as_array())
+                    .is_some_and(|arr| arr.iter().any(|v| v.as_str() == Some(name)));
+                let ty = prop.and_then(|p| p.get("type")?.as_str()).unwrap_or("any");
+                (name.clone(), ty.to_string(), required)
+            })
+            .collect();
```
DOsinga, line 101: is this the best way to parse the json? do we not have a rust object that matches this that we can use to read into?

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -0,0 +1,905 @@
...
+fn create_server_module(server_tools: &[&ToolInfo], ctx: &mut Context) -> Module {
+    let export_names: Vec<JsString> = server_tools
+        .iter()
+        .map(|t| js_string!(t.tool_name.as_str()))
+        .collect();
+
+    let tool_data: Vec<(String, String)> = server_tools
+        .iter()
+        .map(|t| (t.tool_name.clone(), t.full_name.clone()))
+        .collect();
```
DOsinga: nit: you could avoid looping over this twice and unzip in one go

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -0,0 +1,905 @@
...
+            },
+            tool_data,
+        ),
+        None,
+        None,
+        ctx,
+    )
+}
+
+fn create_tool_function(full_name: String) -> NativeFunction {
```
DOsinga: maybe call this full_tool_name to indicate that this is server__tool ?

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -0,0 +1,905 @@
...
+                        result = Some(format!("{before}\n{last}\n__result__ = {name};"));
+                    }
+                    break;
+                }
+            }
+            result.unwrap_or_else(|| {
+                format!("{before}\n__result__ = {};", last.trim_end_matches(';'))
+            })
+        }
+    };
```
DOsinga, line 252: this code looks fragile. it makes a good effort, but there's also a lot of corner cases to be covered here. have we considered to just inject a record_result(...) function and tell the LLM to call that to, eh, record a result?

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -0,0 +1,905 @@
...
+                BATCH MULTIPLE TOOL CALLS INTO ONE execute_code CALL.
+
+                This extension exists to reduce round-trips. When a task requires multiple tool calls:
+                - WRONG: Multiple execute_code calls, each with one tool
+                - RIGHT: One execute_code call with a script that calls all needed tools
+
+                Workflow:
+                    1. Use read_module("server") to discover tools and signatures
+                    2. Write ONE script that imports and calls ALL tools needed for the task
+                    3. Chain results: use output from one tool as input to the next
```
DOsinga, line 311: this is obviously a thing where we can quickly iterate. currently seems very focussed on batching just tool calls, not so much on what code can give you otherwise. one example that springs to mind is the handling of tasks with many entries where by default the LLM only does the first 7

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -0,0 +1,905 @@
...
+                        const status = shell({ command: "git status" });
+                        { files, readme, status }
+                        ```
+
+                        SYNTAX:
+                        - Import: import { tool1, tool2 } from "serverName";
+                        - Call: toolName({ param1: value, param2: value })
+                        - All calls are synchronous, return strings
+                        - Last expression is the result
+                        - No comments in code
```
DOsinga, line 612: ha!

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -0,0 +1,905 @@
...
+            indoc::indoc! {r#"
+                ALWAYS batch multiple tool operations into ONE execute_code call.
+                - WRONG: Separate execute_code calls for read file, then write file
+                - RIGHT: One execute_code with a script that reads AND writes
+
+                Modules: {}
+
+                Use read_module("name") to see tool signatures before calling unfamiliar tools.
+            "#},
+            server_list.join(", ")
```
DOsinga, line 750: I wonder if we should rely on read_module and search if the number of tools is not that many. LLMs can very easily consume large interfaces in one go without getting confused. We could even supply it with the interface that we inject directly so it can call that instead of:
* search for the function
* import the function
* call the function

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -0,0 +1,905 @@
...
+            .to_string();
+
+        let tools = self.get_tool_infos().await;
+        let (call_tx, call_rx) = mpsc::unbounded_channel();
+        let tool_handler = tokio::spawn(Self::run_tool_handler(
+            call_rx,
+            self.context.extension_manager.clone(),
+        ));
+
+        let js_result = tokio::task::spawn_blocking(move || run_js_module(&code, &tools, call_tx))
```
DOsinga, line 354: if the agent writes an infinite loop, are we toast here? for external MCP servers I think we timeout, but this is all in process so it hangs our agent I think

DOsinga (approved):
this is awesome. I have many thoughts, left some as comments, but we should try and ship this quickly

at this point it is up to the user to use code mode by enabling that extension and then the other extensions are only accessible through this. have we tried allowing both paths?

crates/goose/src/agents/code_execution_extension.rs
alexhancock: i put this one in there for you!

alexhancock: it was funny that it at first generated the kinds of comments we often see in PR review for vibecode, so I added this

crates/goose/src/agents/code_execution_extension.rs
alexhancock: yes, but I also think as the writes more and more complex code where it potentially could wait for async things etc in one run, having a timeout is a little different than with a single tool call.

instinct on what a good timeout would be?

crates/goose/src/agents/code_execution_extension.rs
alexhancock: yeah agree, can iterate on all prompting. i think these ones are performing well for now, but expect that we will change them as we learn more.

crates/goose/src/agents/code_execution_extension.rs
alexhancock: i did this at the beginning but @michaelneale's instinct was to move more towards a search/read/execute approach to pull as much as we could out of the context window initially

given we've done more manual testing with this approach I am going to stick with this for now. but we can iterate on this if there is a clear change we find to make that is beneficial.

crates/goose/src/agents/code_execution_extension.rs
michaelneale: this really wouldn't be that different to say using shell to do writes with sed and awk or even bash type of thing is it? did that have a timeout?

crates/goose/src/agents/code_execution_extension.rs
michaelneale: yeah - could have it add in a small number, but even some popular MCPs (like github) overload it right from the start so unless we special case some built in ones... seems consistent to let it discover? (but maybe that is worth it if we note it being an issue?)

michaelneale: main one would be say shell/editor ones so it intrinsically knows, which would save (sometimes) one search/lookup up front (but that is all)

The main thing is we don't load up an unbounded set of tools (extension names are usually reasonable and modest, but tools in MCP world do not seem to be reasonable)

crates/goose/src/agents/code_execution_extension.rs
jamadeo: We can give it a default and then let the agent try again with its own value for timeout if it needs to.

PR #6035, opened 2025-12-10
fix(google): Fix 400 Bad Request error with Gemini 3 thought signatures

rabi:
## Summary
Gemini 3 models require thought_signature in tool calls and responses, and thinking content must be echoed back in the conversation loop.
    Changes:
    - Replace thought_signature field with generic metadata field on ToolRequest/ToolResponse
    - Add ProviderMetadata type alias for provider-specific data
    - Google provider stores thoughtSignature in metadata
    - Helper functions for Google-specific metadata extraction
    - Preserve thinking content in conversation when tool calls are processed
    - Handle signature inheritance for multi-part responses
    - Text-only responses treated as final answer (not thinking)
  
### Type of Change
- [ ] Feature
- [x] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Unit tests and manual testing

### Related Issues
Relates to #5792
Relates to #6091

### DIFF
--- a/crates/goose/src/conversation/message.rs
+++ b/crates/goose/src/conversation/message.rs
@@ -53,6 +53,10 @@ where
     Ok(content)
 }
 
+/// Provider-specific metadata for tool requests/responses.
+/// Allows providers to store custom data without polluting the core model.
+pub type ProviderMetadata = serde_json::Map<String, serde_json::Value>;
+
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 #[derive(ToSchema)]
@@ -62,7 +66,8 @@ pub struct ToolRequest {
     #[schema(value_type = Object)]
     pub tool_call: ToolResult<CallToolRequestParam>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub thought_signature: Option<String>,
+    #[schema(value_type = Object)]
+    pub metadata: Option<ProviderMetadata>,
 }
 
 impl ToolRequest {
@@ -89,6 +94,9 @@ pub struct ToolResponse {
     #[serde(with = "tool_result_serde")]
     #[schema(value_type = Object)]
     pub tool_result: ToolResult<CallToolResult>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(value_type = Object)]
+    pub metadata: Option<ProviderMetadata>,
 }
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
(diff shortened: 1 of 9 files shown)


DOsinga:
thanks @rabi for diving on this. this is a good start, but I am not sure it does everything yet. the spec says that gemini can reply with thinking messages and that we then have to keep the conversation going.

other things to note: shouldn't we just always have an optional thought signature on toolcall/response messages? the testing seems rather convoluted, I'd rather do one big round-trip thing where we construct a conversation in our native format and, convert it to google, have some asserts and then convert it back and then verify that it is the same. that would be then a parttern we can easily copy into other format checkers (since they need to be cross provider compatible)

let me know if you want to continue working on this, I do really appreciate the effort and gemini has been annoying in how they change their api without telling us

DOsinga:
here's the related bug report: https://github.com/block/goose/issues/6091

rabi:
> let me know if you want to continue working on this, I do really appreciate the effort and gemini has been annoying in how they change their api without telling us

Thanks @DOsinga for reviewing. I think I've addressed your comments in the last push. Let me know.

cgwalters:
FWIW I did a skim through this PR and it looks sane, though I am not an expert in some of the core tool calling APIs. Having a generic extensible metadata looks like the right thing to do for sure.

It also does fix the problem for me.

crates/goose/src/providers/formats/google.rs
```diff
@@ -9,10 +9,28 @@ use rmcp::model::{
...
-use crate::conversation::message::{Message, MessageContent};
+use crate::conversation::message::{Message, MessageContent, ProviderMetadata};
 use serde_json::{json, Map, Value};
 use std::ops::Deref;
 
+/// Google-specific metadata key for thought signatures
+pub const THOUGHT_SIGNATURE_KEY: &str = "thoughtSignature";
+
+/// Create metadata with a thought signature
+pub fn metadata_with_signature(signature: &str) -> ProviderMetadata {
```
DOsinga, line 18: same here

DOsinga (approved):
I like, looks clean - if you could remove the more obvious comments, I'd say it is good to go

alexhancock (approved):
Yeah I concur with @DOsinga let's remove the comments which aren't adding much to what the code says, then get this in

rabi:
> I like, looks clean - if you could remove the more obvious comments, I'd say it is good to go

Done!

crates/goose/src/conversation/message.rs
```diff
@@ -62,7 +66,8 @@ pub struct ToolRequest {
     #[schema(value_type = Object)]
     pub tool_call: ToolResult<CallToolRequestParam>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub thought_signature: Option<String>,
+    #[schema(value_type = Object)]
+    pub metadata: Option<ProviderMetadata>,
```
DOsinga, line 70: does this break compatibility with sessions where we've already written out the thought_signature?

rabi: No, it does not. I tested resuming session with builds before and after the PR and it worked fine. I also tested with older gemini models like `gemini-2.0-flash` and did not see any issues. Also, I notice that the thought_signature is a recent addition https://github.com/block/goose/commit/cfdf01567d6fed3e465408210d46d6c83dc6a58c.

DOsinga:
excellent, thank you so much!

Issue #6034, opened 2025-12-10
Doesnt work on ubuntu 25.10 - libssl.so.1.1 error


anupanup2001:
**Goose v1.16.0 fails to start on Ubuntu 25.10**

**To Reproduce**
1. Download the deb file for [v1.16.0](https://github.com/block/goose/releases/download/v1.16.0/goose_1.16.0_amd64.deb)
2. Install it by `sudo dpkg -i goose_1.16.0_amd64.deb`
3. Open the newly installed goose application
4. See error as per the screenshot `The backend server failed to start.`

**Expected behavior**
Goose should start

**Screenshots**
<img width="726" height="456" alt="Image" src="https://github.com/user-attachments/assets/faaad3e4-7af8-4ec5-9162-323e0ca77226" />

**Please provide the following information**
- **OS & Arch:** Ubuntu 25.10 amd64
- **Interface:** UI 
- **Version:** 1.16.0
- **Provider & Model:** [e.g. Google – gemini-2.5-flash]

---

**Additional context**
The previous version 1.15.0 worked fine. This ssl version which goose depends on seems to be a much older version 1.1 and most ubuntu versions after 22.04+ use openssl 3.0 or newer and would have libssl.so.3

Antsiscool:
I am seeing the same error with Ubuntu 24.04.3

elijahsgh:
Same in Fedora
Kind of weird to see it linked against libssl 1.1

sheikhlimon:
i guess its broken on arch as well

Edit: The issue is that goose is using OpenSSL version 0.10.73 which requires libssl.so.1.1, but the latest Docker images are installing libssl3 (OpenSSL 3.x).

Here's what's happening:
1. Dockerfile:41 installs libssl3 (OpenSSL 3.x)
2. Cargo.lock:4593 shows OpenSSL 0.10.73 dependency which expects libssl.so.1.1 (OpenSSL 1.1.x)

Solution: Update the Dockerfile to install the legacy OpenSSL 1.1.x libraries alongside libssl3

or 

```cargo
tokio-tungstenite = { version = "0.28.0", features = ["native-tls"] }
```
to
```cargo
tokio-tungstenite = { version = "0.28.0", features = ["rustls-tls-native-roots"] }
```

DOsinga:
yeah, thanks @sheikhlimon looks like that is it. we'll post a patch

zanesq:
[...]

PR #6051, opened 2025-12-10
fix: linux goosed crashing libssl error

zanesq:
## Summary

fixes https://github.com/block/goose/issues/6034

### Root Cause

The bug was introduced in **PR #5251 "goose remote access"** which added the `tokio-tungstenite` dependency with the `native-tls` feature to `crates/goose-server/Cargo.toml`:

```toml
tokio-tungstenite = { version = "0.28.0", features = ["native-tls"] }
```

On Linux, `native-tls` uses OpenSSL via the system's `libssl`. The build environment (likely Ubuntu 20.04 or similar) links against **OpenSSL 1.1** (`libssl.so.1.1`), but modern Linux distributions ship with **OpenSSL 3.x** (`libssl.so.3`):

- Ubuntu 22.04+ → OpenSSL 3.x
- Ubuntu 25.10 → OpenSSL 3.x  
- Fedora → OpenSSL 3.x
- Arch Linux → OpenSSL 3.x

This causes a runtime error: `libssl.so.1.1: cannot open shared object file`.

### Fix

Change `tokio-tungstenite` from `native-tls` to `rustls-tls-native-roots`:

```diff
-tokio-tungstenite = { version = "0.28.0", features = ["native-tls"] }
+tokio-tungstenite = { version = "0.28.0", features = ["rustls-tls-native-roots"] }
```

**Why this works:**
- `rustls` is a pure Rust TLS implementation - no system OpenSSL dependency
- `rustls-tls-native-roots` uses the system's certificate store for trust anchors
- This is consistent with `reqwest` in the same file which already uses `rustls-tls`

### Files Changed

- `crates/goose-server/Cargo.toml` - one line change

The fix removes `native-tls`, `tokio-native-tls`, `openssl`, `openssl-probe`, and `openssl-sys` from the dependency tree, replacing them with `rustls` and `tokio-rustls`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2913,6 +2913,7 @@ dependencies = [
  "rand 0.9.2",
  "reqwest 0.12.12",
  "rmcp 0.9.1",
+ "rustls 0.23.31",
  "schemars",
  "serde",
  "serde_json",
@@ -5846,6 +5847,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c0ebcbd2f03de0fc1122ad9bb24b127a5a6cd51d72604a3f3c50ac459762b6cc"
 dependencies = [
  "aws-lc-rs",
+ "log",
  "once_cell",
  "ring",
  "rustls-pki-types",
--- a/crates/goose-server/Cargo.toml
+++ b/crates/goose-server/Cargo.toml
@@ -45,6 +45,7 @@ rand = "0.9.2"
 hex = "0.4.3"
 socket2 = "0.6.1"
 fs2 = "0.4.3"
+rustls = { version = "0.23", features = ["ring"] }
 
 [target.'cfg(windows)'.dependencies]
 winreg = { version = "0.55.0" }
--- a/crates/goose-server/src/tunnel/lapstone.rs
+++ b/crates/goose-server/src/tunnel/lapstone.rs
@@ -477,6 +477,8 @@ async fn run_single_connection(
     server_secret: String,
     restart_tx: mpsc::Sender<()>,
 ) {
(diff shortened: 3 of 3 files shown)


zanesq:
@michaelneale can you follow up and check if the goose remote access feature still works after this change?

zanesq:
looking into failing tunnel test

sheikhlimon:
yeah it was pulling in `rustls` with `aws-lc-rs` default provider

zanesq:
Fix for tests failing:

Using rustls-tls-native-roots instead of native-tls uses rustls (a pure Rust TLS implementation) which doesn't depend on system OpenSSL. However, rustls 0.23.x requires an explicit crypto provider to be installed at runtime so I added the ring feature and the install_default() call.

DOsinga:
>Using rustls-tls-native-roots instead of native-tls uses rustls (a pure Rust TLS implementation) which doesn't depend on system OpenSSL. However, rustls 0.23.x requires an explicit crypto provider to be installed at runtime so I added the ring feature and the install_default() call.

can you say more about this?

zanesq:
Using rustls-tls-native-roots instead of native-tls switches to rustls, a pure Rust TLS implementation that doesn't depend on system OpenSSL. The native-roots part means it still uses the system's trusted certificate store for validating server certificates.

However, rustls 0.23.x no longer bundles a default cryptographic backend—it requires an explicit crypto provider to be installed at runtime. The library supports two options: ring (mature, widely used) or aws-lc-rs (AWS's libcrypto). Since both were already in our dependency tree via other crates, I chose ring and added the install_default() call to register it as the process-wide crypto provider before any TLS connections are made.

PR #6057, opened 2025-12-10
Add support for changing working dir and extensions in same window/session

zanesq:
## Summary
Changing working dir restarts the agent and sets it to the session
Extensions serve as new chat defaults and can be applied before starting a new chat or during chat.
Extensions are loaded server side now with the agent startup

Pre-requisite for [multichat](https://github.com/block/goose/pull/5937)

https://github.com/user-attachments/assets/1b987228-d600-41d7-aaab-a03d693b1599

### DIFF
--- a/ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
+++ b/ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
@@ -1,6 +1,6 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useRef } from 'react';
 import { FileText, Clock, Home, Puzzle, History } from 'lucide-react';
-import { useNavigate } from 'react-router-dom';
+import { useNavigate, useSearchParams } from 'react-router-dom';
 import {
   SidebarContent,
   SidebarFooter,
@@ -96,7 +96,16 @@ const menuItems: NavigationEntry[] = [
 
 const AppSidebar: React.FC<SidebarProps> = ({ currentPath }) => {
   const navigate = useNavigate();
+  const [searchParams] = useSearchParams();
   const chatContext = useChatContext();
+  const lastSessionIdRef = useRef<string | null>(null);
+  const currentSessionId = currentPath === '/pair' ? searchParams.get('resumeSessionId') : null;
+
+  useEffect(() => {
+    if (currentSessionId) {
+      lastSessionIdRef.current = currentSessionId;
+    }
+  }, [currentSessionId]);
 
   useEffect(() => {
     const timer = setTimeout(() => {
@@ -130,6 +139,17 @@ const AppSidebar: React.FC<SidebarProps> = ({ currentPath }) => {
     return currentPath === path;
   };
 
+  const handleNavigation = (path: string) => {
+    // For /pair, preserve the current session if one exists
+    // Priority: current URL param > last known session > context
+    const sessionId = currentSessionId || lastSessionIdRef.current || chatContext?.chat?.sessionId;
+    if (path === '/pair' && sessionId && sessionId.length > 0) {
+      navigate(`/pair?resumeSessionId=${sessionId}`);
+    } else {
+      navigate(path);
+    }
+  };
+
   const renderMenuItem = (entry: NavigationEntry, index: number) => {
     if (entry.type === 'separator') {
       return <SidebarSeparator key={index} />;
@@ -144,7 +164,7 @@ const AppSidebar: React.FC<SidebarProps> = ({ currentPath }) => {
             <SidebarMenuItem>
               <SidebarMenuButton
                 data-testid={`sidebar-${entry.label.toLowerCase()}-button`}
-                onClick={() => navigate(entry.path)}
+                onClick={() => handleNavigation(entry.path)}
                 isActive={isActivePath(entry.path)}
                 tooltip={entry.tooltip}
                 className="w-full justify-start px-3 rounded-lg h-fit hover:bg-background-medium/50 transition-all duration-200 data-[active=true]:bg-background-medium"


zanesq:
Put in draft while I work on a few more places using working dir that need to be addressed.

crates/goose-server/tests/test_working_dir_update.rs
```diff
@@ -0,0 +1,53 @@
+#[cfg(test)]
```
DOsinga: this test doesn't make any sense. it's not testing whether you can change working dirs, it is testing whether you can change a field in the session database. if we really feel like that is special, add it to the testing there

ui/desktop/src/store/newChatState.ts
```diff
@@ -0,0 +1,37 @@
+// Store for new chat configuration
+// Acts as a cache that can be updated from UI or synced from session
+// Resets on page refresh - defaults to window.appConfig.get('GOOSE_WORKING_DIR')
```
DOsinga: this doesn't seem like the way. this is yet another shadow settings thing. the way we handle GOOSE_WORKING_DIR is a complete mess, but we should fix that, not stack stuff on top

ui/desktop/src/main.ts
```diff
@@ -1188,9 +1188,25 @@ ipcMain.handle('open-external', async (_event, url: string) => {
...
+  try {
+    if (!dir) {
+      return false;
+    }
+    addRecentDir(dir);
+    return true;
+  } catch (error) {
+    console.error('Error adding recent dir:', error);
+    return false;
+  }
```
DOsinga, line 1202: why the plumbing, this can just call addRecentDir no? that can't fail it is vibe coded with exception handlers to the brim and we ignore the return value

ui/desktop/src/main.ts
```diff
@@ -1188,9 +1188,25 @@ ipcMain.handle('open-external', async (_event, url: string) => {
   }
 });
 
-// Handle directory chooser
-ipcMain.handle('directory-chooser', (_event) => {
-  return openDirectoryDialog();
+ipcMain.handle('directory-chooser', async () => {
+  const result = await dialog.showOpenDialog({
+    properties: ['openFile', 'openDirectory', 'createDirectory'],
+    defaultPath: os.homedir(),
```
DOsinga, line 1195: open file? also, default to homedir?

ui/desktop/src/components/settings/app/ExternalBackendSection.tsx
```diff
@@ -100,81 +100,81 @@ export default function ExternalBackendSection() {
       <Card className="pb-2">
```
DOsinga, line 100: did anything change here, or just the formatter?

ui/desktop/src/components/ChatInput.tsx
```diff
@@ -143,6 +144,27 @@ export default function ChatInput({
...
+    const fetchSessionWorkingDir = async () => {
+      try {
+        const response = await getSession({ path: { session_id: sessionId } });
+        if (response.data?.working_dir) {
+          setSessionWorkingDir(response.data.working_dir);
+        }
+      } catch (error) {
+        console.error('[ChatInput] Failed to fetch session working dir:', error);
+      }
+    };
```
DOsinga, line 173: if we fetch the working dir here, shouldn't we just pass that on to the DirSwitcher instead of having it pick its own? probably be nicer if the dirswitcher had a callback on what should happen when the dir switched

crates/goose/src/agents/agent.rs
```diff
@@ -581,7 +581,11 @@ impl Agent {
         Ok(())
     }
 
-    pub async fn add_extension(&self, extension: ExtensionConfig) -> ExtensionResult<()> {
+    pub async fn add_extension(
+        &self,
+        extension: ExtensionConfig,
+        working_dir: Option<std::path::PathBuf>,
+    ) -> ExtensionResult<()> {
```
DOsinga: this doesn't feel great to me. an agent has a working_dir. if you add an extension and that extension needs to know what the working dir is, it should just get it from the agent

crates/goose-server/src/routes/agent.rs
```diff
@@ -592,6 +587,158 @@ async fn stop_agent(
...
+            status: StatusCode::INTERNAL_SERVER_ERROR,
+        })
+}
+
+async fn restore_agent_extensions(
+    agent: Arc<Agent>,
+    working_dir: &std::path::Path,
+) -> Result<(), ErrorResponse> {
+    let working_dir_buf = working_dir.to_path_buf();
+    let enabled_configs = goose::config::get_enabled_extensions();
```
DOsinga: this is getting the currently enabled extensions. so if you change the working dir, the agent restarts and we might get a different set of extensions

crates/goose-server/src/routes/agent.rs
```diff
@@ -592,6 +587,158 @@ async fn stop_agent(
     Ok(StatusCode::OK)
 }
 
+async fn restore_agent_provider(
```
DOsinga: it's good to factor these out, but they don't belong in routes I think

ui/desktop/src/components/bottom_menu/DirSwitcher.tsx
```diff
@@ -1,23 +1,84 @@
...
+
+    if (sessionId) {
+      try {
+        await updateSessionWorkingDir({
+          path: { session_id: sessionId },
+          body: { workingDir: newDir },
+        });
+
+        try {
+          await restartAgent({ body: { session_id: sessionId } });
```
DOsinga: this shouldn't be a separate call. we should have an agentUpdateWorkingDir that might also restart the agent - the client shouldn't know about this implementation detail

ui/desktop/src/components/settings/app/ExternalBackendSection.tsx
zanesq: checked locally against the previous version, just the formatting for some reason the gh diff viewer is making it appear that more is happening

ui/desktop/src/store/newChatState.ts
zanesq: Good call, removed this and changed to use session as source of truth and pass working dir rather than cacheing it in https://github.com/block/goose/pull/6057/commits/960285bfae03a0628d6f53f43148e43e0bbeb064

crates/goose/src/agents/agent.rs
zanesq: Agreed, added working_dir to agent synced with session in this commit https://github.com/block/goose/pull/6057/commits/6b4d5709a8c65342f4eb89b02ef3b4121db6ad31

crates/goose-server/src/routes/agent.rs
zanesq: Will address in follow up extensions PR shortly

zanesq:
@DOsinga mind taking one more pass before I merge?

crates/goose/src/agents/agent.rs
```diff
@@ -173,9 +174,19 @@ impl Agent {
             scheduler_service: Mutex::new(None),
             retry_manager: RetryManager::new(),
             tool_inspection_manager: Self::create_default_tool_inspection_manager(),
+            working_dir: Mutex::new(None),
```
DOsinga: agent shouldn't have a working_dir; we already store that in the session, can't have two sources of truth

crates/goose-server/src/routes/agent_utils.rs
```diff
@@ -0,0 +1,85 @@
+use crate::routes::errors::ErrorResponse;
```
DOsinga: let's call this something else. and it also does not belong in routes. I'm sure we're doing restore_agent_provider-y stuff in core goose so it should live there as just some added functionality to agent.rs::Agent

crates/goose-server/src/routes/agent_utils.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+pub async fn restore_agent_extensions(
+    agent: Arc<Agent>,
+    working_dir: &std::path::Path,
+) -> Result<(), ErrorResponse> {
+    // Set the agent's working directory before adding extensions
+    agent.set_working_dir(working_dir.to_path_buf()).await;
+
+    let enabled_configs = goose::config::get_enabled_extensions();
```
DOsinga: this is still going to override the extensions we have set

crates/goose/src/agents/agent.rs
zanesq: makes sense, moved working_dir to session so only one source of truth

crates/goose-cli/src/commands/acp.rs
```diff
@@ -241,6 +241,7 @@ fn format_tool_name(tool_name: &str) -> String {
 }
 
 async fn add_builtins(agent: &Agent, builtins: Vec<String>) {
+    let working_dir = std::env::current_dir().ok();
```
DOsinga: any time we see this, things don't smell good - what's the current directory for acp?

crates/goose-server/src/routes/session.rs
```diff
@@ -181,6 +190,51 @@ async fn update_session_name(
...
+    SessionManager::update_session(&session_id)
+        .working_dir(path)
+        .apply()
+        .await
+        .map_err(|e| {
+            tracing::error!("Failed to update session working directory: {}", e);
+            StatusCode::INTERNAL_SERVER_ERROR
+        })?;
+
+    Ok(StatusCode::OK)
```
DOsinga: why do we need this? this just seems to change the working_dir without telling the agent about it while we also have something where do change that with the same parameters. I didn't see any uses of this.

crates/goose-server/src/routes/reply.rs
```diff
@@ -274,6 +274,8 @@ pub async fn reply(
             }
         };
 
+        std::env::set_var("GOOSE_WORKING_DIR", &session.working_dir);
```
DOsinga: why do we need this? goosed is handling multiple agents, we can't be setting environment variables

crates/goose-server/src/routes/agent.rs
```diff
@@ -34,7 +36,48 @@ use std::path::PathBuf;
...
+        .persist_extension_state(session_id)
+        .await
+        .map_err(|e| {
+            error!("Failed to persist extension state: {}", e);
+            ErrorResponse {
+                message: format!("Failed to persist extension state: {}", e),
+                status: StatusCode::INTERNAL_SERVER_ERROR,
+            }
+        })
+}
```
DOsinga: these functions above just seem to be forwarding calls to the agent.

crates/goose/src/agents/agent.rs
```diff
@@ -565,7 +573,97 @@ impl Agent {
         Ok(())
     }
 
-    pub async fn add_extension(&self, extension: ExtensionConfig) -> ExtensionResult<()> {
+    /// Save current extension state to session by session_id
+    /// Simpler version of save_extension_state that just takes a session_id
```
DOsinga: I don't understand this comment

crates/goose/src/agents/agent.rs
```diff
@@ -565,7 +573,97 @@ impl Agent {
...
+            .extension_data(extension_data)
+            .apply()
+            .await?;
+
+        Ok(())
+    }
+
+    /// Load extensions from session into the agent
+    /// Skips extensions that are already loaded
+    pub async fn load_extensions_from_session(
```
DOsinga, line 598: this function seems to be called only once, which doesn't seem right; shouldn't we be calling this everywhere where we start an agent? really I think this should be done through the agentmanager and we should just always use agent_for_session which would then get us the right agent.

crates/goose/src/agents/agent.rs
```diff
@@ -565,7 +573,97 @@ impl Agent {
...
+            })
+            .collect::<Vec<_>>();
+
+        futures::future::join_all(extension_futures).await
+    }
+
+    pub async fn add_extension(
+        &self,
+        extension: ExtensionConfig,
+        working_dir: Option<std::path::PathBuf>,
```
DOsinga: I don't think we should pass the working_dir in here and if we do, it shouldn't be optional; an agent should have one working_dir, this just makes it possible to have multiple source of the truth

crates/goose/src/agents/agent.rs
```diff
@@ -565,7 +573,97 @@ impl Agent {
         Ok(())
     }
 
-    pub async fn add_extension(&self, extension: ExtensionConfig) -> ExtensionResult<()> {
+    /// Save current extension state to session by session_id
+    /// Simpler version of save_extension_state that just takes a session_id
+    pub async fn persist_extension_state(&self, session_id: &str) -> Result<()> {
```
DOsinga, line 577: why do we need this? shouldn't we just always when you call add_extension on agent write it to the session?

crates/goose/src/agents/agent.rs
```diff
@@ -565,7 +573,97 @@ impl Agent {
...
+    /// Load extensions from session into the agent
+    /// Skips extensions that are already loaded
+    pub async fn load_extensions_from_session(
+        self: &Arc<Self>,
+        session: &Session,
+    ) -> Vec<ExtensionLoadResult> {
+        let working_dir = session.working_dir.clone();
+
+        // Try to load session-specific extensions first, fall back to global config
+        let session_extensions =
```
DOsinga, line 602: we should not need to do this; the extensions should always be in the session, no?

crates/goose-server/src/routes/agent.rs
zanesq: thanks was left over from refactoring, inlined instead

crates/goose-cli/src/commands/acp.rs
zanesq: good catch changed to cwd

crates/goose/src/agents/agent.rs
zanesq: We call it in 3 places in agent.rs start_agent, resume_agent and restart_agent_internal but good point we need a single "get me a fully configured agent for this session" method but I think it should be in a follow up larger refactor because there is a lot of considerations that need to be made (sub agents use parent's extensions not session extensions).

crates/goose/src/agents/agent.rs
zanesq: Agreed, removed and changed it to look it up from the session instead for now.

crates/goose/src/agents/agent.rs
zanesq: good point, changed to do it in add_extension 👍

crates/goose/src/agents/agent.rs
zanesq: the idea was backwards compatibility for when we didn't have extensions in sessions but I went ahead and removed that for now and logged it since its a rare case

zanesq:
@DOsinga thanks for the review! I updated it to not thread working_dir everywhere and I think we're at a good middle ground for now to unblock multi chat. Definitely agree that we should come back to the overall vision in a follow up.

crates/goose-server/src/routes/agent.rs
```diff
@@ -571,6 +571,17 @@ async fn agent_add_extension(
         ErrorResponse::internal(format!("Failed to add extension: {}", e))
     })?;
 
+    // Persist here rather than in add_extension to ensure we only save state
```
DOsinga, line 574: No. This should not happen in routes. this means that anything that hits the same path (ACP?) will not have this persisted.

codefromthecrypt:
Agreed on design debt that is accumulating though not specific to this PR. Makes me think we could use a clean room spec sheet so whatever v2 design that handles ACP well (e.g. clean separation of agent from session) doesn't mess up multichat. Almost sounds pow-wow like several in a room for a week worthy to sort all this out quickly as slowly iterating small design fixes not being able to address all .. or compromise to not impact all.. this approach is very time consuming.

crates/goose-server/src/routes/agent.rs
zanesq: Yes I remember you flagging that but it turned into a rabbit hole at the last minute. Here is the follow up: https://github.com/block/goose/pull/6685

PR #6058, opened 2025-12-10
revert /reply to old behavior when full conversation provided

tlongwell-block:
Allow passing full conversation to /reply if needed. This supports the previous behavior of the /reply endpoint before the session required overhaul as part of the agent manager work. This is required for some uses of goose-server where session contents may have more than one source of truth, like the goose slackbot.

### DIFF
--- a/crates/goose-server/src/routes/reply.rs
+++ b/crates/goose-server/src/routes/reply.rs
@@ -258,7 +258,7 @@ pub async fn reply(
             }
         };
 
-        let session = match SessionManager::get_session(&session_id, false).await {
+        let session = match SessionManager::get_session(&session_id, true).await {
             Ok(metadata) => metadata,
             Err(e) => {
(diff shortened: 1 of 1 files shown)


crates/goose-server/src/routes/reply.rs
```diff
@@ -316,7 +317,7 @@ pub async fn reply(
             }
         };
 
-        let mut all_messages = messages.clone();
+        let mut all_messages = Conversation::new_unvalidated(vec![user_message.clone()]);
```
DOsinga: this doesn't strike me as quite right; we keep a shadow administration of the messages here I think just for tracking purposes. we should probably move that to agent, but until we do, we should in the block above retrieve the conversation so far from the db if it is not given and then call the agent only with the user message

DOsinga (approved):
I think we do need to keep the full conversation in the loop

PR #6065, opened 2025-12-11
Feat: Add SAP AI Core provider

augustine-d-nguyen:
## Summary

This PR adds support for SAP AI Core as a new LLM provider in Goose. SAP AI Core is SAP's AI infrastructure platform hosted on SAP BTP (Business Technology Platform), which provides enterprise-grade hosting for various language models including GPT-4o, Claude 3.5 Sonnet, and others.

This integration enables Goose users within SAP ecosystems to leverage their existing SAP AI Core infrastructure, benefiting from:
- Enterprise security and compliance features
- Integration with SAP's authentication and authorization mechanisms
- Support for multiple LLM models hosted in SAP BTP AI Core
- Cost tracking and governance through SAP's billing systems

### Implementation Details
- Added SAP AI Core provider implementation following Goose's provider interface pattern
- Implemented authentication using SAP AI Core credentials (OAuth2/service keys)
- Added configuration support for SAP AI Core endpoints and deployment URLs
- Included proper error handling and retry logic for API calls
- Added support for streaming responses
- Documented configuration requirements in provider documentation

### Type of Change
- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [x] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance
- [x] This PR was created with AI assistance (Goose + SAP AI Core)

### Testing

**Unit Tests:**
- Added unit tests for SAP AI Core provider initialization
- Added tests for authentication flow
- Added tests for request/response handling
- Added tests for error handling and edge cases

**Integration Tests:**
- Tested against SAP AI Core staging/test environment
- Verified compatibility with multiple models (GPT-4o, Claude 3.5 Sonnet)
- Tested streaming and non-streaming responses
- Verified proper handling of rate limits and API errors

**Manual Testing:**
- Successfully authenticated with SAP AI Core using service keys
- Verified end-to-end conversation flow with various models
- Tested configuration via environment variables and config files
- Confirmed proper error messages for misconfiguration scenarios

Before:
N/A
After:   
Provider Configuration:
image
image

Chat with Claude hosted in SAP AI Core (with cost info)
image

### DIFF
--- a/crates/goose/src/providers/formats/sap_ai_core.rs
+++ b/crates/goose/src/providers/formats/sap_ai_core.rs
@@ -837,6 +837,7 @@ pub fn get_streaming_response_usage(
 mod tests {
     use super::*;
     use crate::conversation::message::MessageContent;
+    use rmcp::model::CallToolResult;
 
     #[test]
     fn test_create_request() {
@@ -945,9 +946,12 @@ mod tests {
             ),
             Message::user().with_tool_response(
                 "call_123",
-                Ok(vec![Content::text(
-                    "The weather in New York is sunny, 75°F",
-                )]),
+                Ok(CallToolResult {
+                    content: vec![Content::text("The weather in New York is sunny, 75°F")],
+                    structured_content: None,
+                    is_error: Some(false),
+                    meta: None,
+                }),
             ),
             Message::assistant().with_text("The weather in New York is sunny and 75°F today!"),
         ];


crates/goose/src/providers/token_manager.rs
```diff
@@ -0,0 +1,226 @@
+use std::sync::{Arc, RwLock};
+use std::time::{Duration, SystemTime};
+
+#[derive(Clone, Debug)]
```
michaelneale: should this be part of the sap code vs its own top level module/file?

crates/goose/src/providers/sap_ai_core.rs
```diff
@@ -0,0 +1,955 @@
+use anyhow::Result;
```
michaelneale, line 1: this has same name as the file in the formats dir? is this really the provider?

michaelneale:
thanks @augustine-d-nguyen - had some comments. This seems like a fair bit of code - is the SAP provide not underneath an openai compatible LLM endpoint in some fashion (that would simplify things a lot) - I wasn't able to find out.

crates/goose/src/providers/token_manager.rs
augustine-d-nguyen: I designed this struct as a utility to manage the JWT token lifecycle. It caches the token to prevent the provider from requesting a new token for every message sent. If no other providers will use it, I can move it into the SAP code module.

crates/goose/src/providers/sap_ai_core.rs
augustine-d-nguyen: Yes, I'm following the pattern used by other providers where the provider and formatter share the same name. The file in the `formats` directory is the message formatter, while the one in the `providers` directory is the provider itself. Additionally, since there are many message payload types, I split them into a separate `*_payload.rs` file for better organization.

augustine-d-nguyen:
> thanks @augustine-d-nguyen - had some comments. This seems like a fair bit of code - is the SAP provide not underneath an openai compatible LLM endpoint in some fashion (that would simplify things a lot) - I wasn't able to find out.

Hi @michaelneale, thank you for your review. SAP AI Core is using its own format for its orchestration API. It's not compatible to OpenAI API spec

crates/goose/src/providers/sap_ai_core.rs
michaelneale: oh right - and "sap_ai_core" = then "AI Core" is the product (?) name of the SAP endpoint? (I interpreted _core as meaning it is some utility/library, but sounds like this is just name, so yeah, pattern makes sense)

crates/goose/src/providers/token_manager.rs
michaelneale: I would be curious if others could use it to leave it there, it not could be folded in to keep things clean

crates/goose/src/providers/sap_ai_core.rs
augustine-d-nguyen: Yeah, the AI Core is the service name, the full name is SAP AI Core, it's a service on BTP platform with additional functions around the LLM models, such as orchestration API:  rounding, masking....

crates/goose/src/providers/token_manager.rs
augustine-d-nguyen: Currently, only the SAP provider needs it since it's the only one using JWT-based authentication with token caching. I'll move the token_manager into the SAP provider module to keep things clean, as you suggested. If other providers need similar functionality in the future, we can always extract it then.

clouatre (commented):
Thanks for the contribution. I see you've already agreed to move `token_manager` into the SAP module per @michaelneale's feedback.

### Additional Issues

**1. Missing integration tests**
- Only unit tests for `TokenManager` in `sap_ai_core.rs`
- No tests in `crates/goose/tests/`
- No update to `goose-self-test.yaml`
- Project guideline: "When adding features, update goose-self-test.yaml, rebuild, then run \`goose run --recipe goose-self-test.yaml\` to validate"

**2. Excessive logging**
- Multiple `tracing::info!()` calls for success paths (12 total in `formats/sap_ai_core.rs`)
- Project guideline: "Clean up existing logs, don't add more unless for errors or security events"
- Examples: "Successfully loaded image from path", "Successfully serialized tool call arguments", "This chunk is not an error"
- For comparison: `formats/openai.rs` and `anthropic.rs` have 0 info-level logs

### Request Changes

1. Add integration tests under `crates/goose/tests/`
2. Update `goose-self-test.yaml` with SAP provider test case
3. Remove success-path info logs (keep only error-level)

alexhancock:
I'm sorry but this seems like a lot of code using new data types for almost everything and a new API format that only this will use. No other provider will use this, we don't have a way to test it ourselves, and we aren't sure how much interest there will be in it. Thank you for working on it, but for the reasons above I'll close it out for now.

We'd be more than happy to have it as a declaratively defined JSON based provider.

Issue #3196, opened 2025-07-01
Add a hotkey to show full tool output


rolznz:
**Please explain the motivation behind the feature request.**
This is how my tool output looks like, but I can't see the invoice - all I can see is "...".
```
I'll pay the Lightning invoice to get the weather data for Wellington.
─── pay_invoice | alby ──────────────────────────
invoice: ...
```

**Describe the solution you'd like**
Claude code allows you to press a hotkey (ctrl+r) to toggle displaying the full tool output

**Describe alternatives you've considered**
-

**Additional context**
-

- [x] I have verified this does not duplicate an existing feature request

aegntic:
.take

taniacryptid:
Hi @aegntic - wanted to check on your progress with this issue and see if you needed any help :)

aegntic:
I've got a couple other things I have to take care of irl and I will return to the flock when I'm done 👍🏻✅

taniacryptid:
thank you for the update! <3

taniacryptid:
Hi @aegntic - wanted to tag you and see how it's going for this issue! As we are in the last week of Hacktoberfest, wanted to make sure you have the support you need and also check if you have trouble with this issue. Thank you!

taniacryptid:
Hi @aegntic, I am writing to let you know I'm removing the hacktoberfest labels off of this issue. You are free to contribute to this if you have a PR ready! 📕 and if it is before Hacktoberfest ends, I will be able to add those labels as well.

Since Hacktoberfest ends in two days, and the team needs that remaining day to finalize PRs and rankings, all open issues without an open PR are being closed or having Hacktoberfest labels removed. This way, in-progress submissions can bee finalized along with final leaderboard rankings. Thank you so much for participating!

Abhijay007:
Hi @aegntic are you still looking into this one ?

clouatre:
I'm experiencing this same issue with `gh auth login`, which makes GitHub authentication impossible through goose.

**Reproduction:**
1. Ask goose to run: `gh auth login`
2. The device code is truncated, while the URL is visible
3. The command hangs waiting for authentication, but without the device code it's impossible to complete
4. The only way out is CTRL+C

**Expected behavior:**
The full device code and URL should be visible (like in kiro-cli, which handles this correctly, and completes the full authentication).

**Comparison:**

**In kiro-cli (works correctly):**
```
I will run the following command: gh auth login (using tool: shell)
Purpose: Authenticate with GitHub CLI

[...]

PR #6067, opened 2025-12-11
feat: add hotkey to toggle full tool output display

Abhijay007:
Closes: #3196

## PR Description

This PR aims to implement a feature request to allow users to view complete tool output without truncation. It resolves issues where important information (such as Lightning invoice codes and GitHub device codes) was being cut off with `"..."` ellipsis.

## Changes made : 

- Add `/r` command to toggle full tool output mode
- Modify `print_value()` to respect toggle when truncating
- Add `ToggleFullToolOutput` handler in session loop
- Update help text with `/r` command in `input.rs`
- Add unit tests for toggle functionality

## Type of Change

- [x] Feature
- [x] Bug fix

## AI Assistance
- [x] This PR was created or reviewed with AI assistance — used Goose for approach validation.

## Testing

Added tests for the fix and also tested via CLI by running different commands.

## Screenshots / Demos (for UX changes)

**Before:**  
Screenshot 2025-12-11 195316

**After:**  
Screenshot 2025-12-11 195358

### DIFF
--- a/crates/goose-cli/src/session/output.rs
+++ b/crates/goose-cli/src/session/output.rs
@@ -91,9 +91,9 @@ pub fn get_theme() -> Theme {
 
 pub fn toggle_full_tool_output() -> bool {
     SHOW_FULL_TOOL_OUTPUT.with(|s| {
-        let new_value = !*s.borrow();
-        *s.borrow_mut() = new_value;
-        new_value
+        let mut val = s.borrow_mut();
+        *val = !*val;
+        *val
     })
 }
 


crates/goose-cli/src/session/output.rs
```diff
@@ -88,6 +89,18 @@ pub fn get_theme() -> Theme {
     CURRENT_THEME.with(|t| *t.borrow())
 }
 
+pub fn toggle_full_tool_output() -> bool {
+    SHOW_FULL_TOOL_OUTPUT.with(|s| {
+        let new_value = !*s.borrow();
+        *s.borrow_mut() = new_value;
+        new_value
+    })
```
DOsinga, line 97: you can use borrow_mut for one go:

```
pub fn toggle_full_tool_output() -> bool {
    SHOW_FULL_TOOL_OUTPUT.with(|s| {
        let mut val = s.borrow_mut();
        *val = !*val;
        *val
    })
}
```

Abhijay007:
> why `/r`?

The original issue mentioned Claude code uses ctrl+r so as we are using slash commands, I thought /r would be better

crates/goose-cli/src/session/output.rs
Abhijay007: ca97b9e

PR #6068, opened 2025-12-11
Add /goose issue solver github workflow

tlongwell-block:
Adds a workflow that lets maintainers trigger goose on any issue by commenting `/goose` as the first part of a comment.                                                                                                  
                                                                                                                                                                                          
goose will analyze the issue, implement a fix, and open a draft PR if successful.                                                                                                         
                                                                                                                                                                                          
- Only OWNER/MEMBER can trigger                                                                                                                                                           
- Creates draft PRs for human review                                                                                                                                                      
- Recipe is customizable via the GOOSE_RECIPE env variable so this action can be used in other repos

### DIFF
--- a/.github/workflows/goose-issue-solver.yml
+++ b/.github/workflows/goose-issue-solver.yml
@@ -59,7 +59,7 @@ env:
       ## Phase 4: Implement
       - [ ] Implement minimal fix per /tmp/requirements.md
       - [ ] Before adding anything, check: is it in the requirements? If not, don't.
-      - [ ] Max 10 files, no .github/, no lock files, no secrets
+      - [ ] No .github/, no lock files, no secrets
 
       ## Phase 5: Verify
       - [ ] source bin/activate-hermit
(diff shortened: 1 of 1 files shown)


.github/workflows/goose-issue-solver.yml
```diff
@@ -0,0 +1,197 @@
...
+
+      ## Phase 3: Plan
+      - [ ] Decide on implementation approach
+      - [ ] For nontrivial issues, use subagents to evaluate architecture or implementation choices
+      - [ ] Update TODO with specific changes to make
+
+      ## Phase 4: Implement
+      - [ ] Implement minimal fix per /tmp/requirements.md
+      - [ ] Before adding anything, check: is it in the requirements? If not, don't.
+      - [ ] Max 10 files, no .github/, no lock files, no secrets
```
jamadeo: max 10 files seems arbitrary?

.github/workflows/goose-issue-solver.yml
```diff
@@ -0,0 +1,197 @@
...
+          gh api /repos/${{ github.repository }}/issues/$ISSUE_NUMBER > /tmp/issue.json
+
+          echo "number=$ISSUE_NUMBER" >> $GITHUB_OUTPUT
+
+          echo "title<<TITLE_EOF" >> $GITHUB_OUTPUT
+          jq -r '.title' /tmp/issue.json >> $GITHUB_OUTPUT
+          echo "TITLE_EOF" >> $GITHUB_OUTPUT
+
+          echo "body<<BODY_EOF" >> $GITHUB_OUTPUT
+          jq -r '.body // ""' /tmp/issue.json >> $GITHUB_OUTPUT
```
jamadeo: this could break if the issue has BODY_EOF in its body. Unlikely but not impossible

.github/workflows/goose-issue-solver.yml
```diff
@@ -0,0 +1,197 @@
...
+          apt-get update
+          apt-get install -y gh
+
+      - name: Get issue details
+        id: issue
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ISSUE_NUMBER: ${{ github.event.issue.number || github.event.inputs.issue_number }}
+        run: |
+          gh api /repos/${{ github.repository }}/issues/$ISSUE_NUMBER > /tmp/issue.json
```
jamadeo: security-wise I don't think you want to read the issue body here, as it may be changed between the maintainer commenting /goose and this action being run

to prevent that, use the body from the event payload: https://docs.github.com/en/webhooks/webhook-events-and-payloads#issue_comment

.github/workflows/goose-issue-solver.yml
tlongwell-block: Removed that limitation. I had imagined this slash command would only be useful for tiny fixes, but I think we could use it for bigger work with opus 4.5 and gpt 5.2 probably

.github/workflows/goose-issue-solver.yml
tlongwell-block: Great catch. This has been fixed when the action is kicked off from a comment

.github/workflows/goose-issue-solver.yml
tlongwell-block: Refactored to remove this entirely

PR #6072, opened 2025-12-11
Add HTTP proxy support for Goose GUI (#5836)

### LINKED ISSUE #5836: Goose GUI does not support explicit HTTP proxy servers
**Describe the bug**

---

**To Reproduce**
1. Connect to network that does not use proxies
2. Configure providers and models, it works
3. Connect to network that requires explicit HTTP proxy servers (aka must be configured in ENV within shells or network settings for system apps)
4. Attempt to configure providers and models, it will fail.
5. Repate using `goose` cli, it will succeed.
---

**Expected behavior**
GUI app would load proxy configurations from system network configuration or as defined within user shell ENV.
If app cannot load system proxy configurations, it should allow user to specify HTTPS Proxy and no proxy rules.

---

**Screenshots**
Image showing success when using network without proxy:
Image

Image of attempt to configure when on network with mandatory proxy:
Image

---

- **OS & Arch:** macOS Darwin 25.1.0
- **Interface:** GUI
- **Version:** 1.15.0
- **Extensions enabled:** N/A
- **Provider & Model:** All other than local ollama

---

I believe this also causes new installs of Goose to fail the "first run" tasks when using GUI, using CLI within shell with ENV set correctly functions as expected.

shanto12:
## Summary
This PR adds HTTP proxy configuration support to the Goose desktop GUI application to address issue #5836.

Added a `configureProxy()` function that:
- Reads HTTP_PROXY, HTTPS_PROXY, and NO_PROXY environment variables
- Configures Electron's session proxy using `session.defaultSession.setProxy()`
- Falls back to system proxy detection if environment variables are not set
- Includes proper error handling and logging

The function is called during app initialization in `app.whenReady()` to ensure proxy settings are applied before any network requests are made.

## Type of Change
- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

## Testing
The implementation follows standard Electron proxy configuration patterns. Testing should include:
- Setting HTTP_PROXY/HTTPS_PROXY environment variables and verifying the app uses the proxy
- Testing with NO_PROXY to verify bypass rules work correctly
- Testing on corporate networks with explicit proxy requirements
- Verifying system proxy detection works when environment variables are not set

## Related Issues
Closes #5836

## Additional Notes
This is a foundational implementation that enables proxy support. Future enhancements could include:
- UI settings for manual proxy configuration
- Proxy authentication support
- PAC (Proxy Auto-Configuration) file support

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -132,6 +132,40 @@ async function ensureTempDirExists(): Promise<string> {
   return gooseTempDir;
 }
 
+// Configure HTTP proxy settings for the application
+async function configureProxy() {
+  try {
+    // Check for proxy environment variables
+    const httpsProxy = process.env.HTTPS_PROXY || process.env.https_proxy;
+    const httpProxy = process.env.HTTP_PROXY || process.env.http_proxy;
+    const noProxy = process.env.NO_PROXY || process.env.no_proxy || '';
+    
+    const proxyUrl = httpsProxy || httpProxy;
+    
+    if (proxyUrl) {
+      console.log('[Main] Configuring proxy:', proxyUrl);
+      
+      // Set proxy configuration for the default session
+      await session.defaultSession.setProxy({
+        proxyRules: proxyUrl,
+        proxyBypassRules: noProxy
+      });
+      
+      console.log('[Main] Proxy configured successfully');
+    } else {
+      // Try to resolve system proxy settings
+      const systemProxy = await session.defaultSession.resolveProxy('https://example.com');
+      if (systemProxy && systemProxy !== 'DIRECT') {
+        console.log('[Main] Using system proxy:', systemProxy);
+      } else {
+        console.log('[Main] No proxy configured');
+      }
+    }
+  } catch (error) {
+    console.error('[Main] Failed to configure proxy:', error);
+  }
+}
+
 if (started) app.quit();
 
 if (process.env.ENABLE_PLAYWRIGHT) {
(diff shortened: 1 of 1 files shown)


michaelneale:
thanks @shanto12  - do you mind signing the DCO stuff for your commits?

shanto12:
Thanks @michaelneale! I'll sign off the commits with DCO. Will update shortly with:

```
git commit --amend --signoff
git push --force-with-lease
```

Regarding the Codex feedback about proxy configuration placement - the `configureProxy()` call in `app.whenReady()` is intentional. This ensures proxy settings are applied before any network requests, including during the initial app launch. The function gracefully falls back to system proxy detection if environment variables aren't set, so it won't break normal startup.

Thanks for reviewing!

michaelneale:
thanks @shanto12  - can you do the signing/push so we can merge this in?

ui/desktop/src/main.ts
```diff
@@ -132,6 +132,40 @@ async function ensureTempDirExists(): Promise<string> {
   return gooseTempDir;
 }
 
+// Configure HTTP proxy settings for the application
+async function configureProxy() {
+  try {
+    // Check for proxy environment variables
```
michaelneale: can probably edit out the obvious comments added

shanto12:
@michaelneale I appreciate your patience! I'd like to sign off the commits with DCO as requested. However, signing commits requires command-line git access with `git commit --amend --signoff` and `git push --force-with-lease`.

Since I'm currently working through the browser interface, I'm unable to perform these git operations directly. 

Would you be able to either:
1. Add the DCO sign-off on my behalf (if that's permissible in your workflow), or  
2. Accept the PR with the understanding that I acknowledge the Developer Certificate of Origin for these changes?

I confirm that I have authored this contribution and agree to the terms of the DCO.

Alternatively, if there's a way to sign commits through the GitHub web interface that I'm not aware of, please let me know!

michaelneale:
@shanto12 hrm - I don't think I can override that, but I could make a fresh PR with this change, but I would have to manually attribute to you in comments and the PR (but you may not show up in commit history). Would that be ok?

michaelneale:
I am goign to close this and make it a separate branch PR

PR #6074, opened 2025-12-11
Tool reply meta

DOsinga:
## Summary

Make goose keep the toolresults around


aharvard:
I just tested. 

Should we be seeing structured content in the tool result?

image
image

Also, should goose be aware of meta? 
image
image

DOsinga:
No, for now we're not changing what we are actually sending to the LLM; just what gets send to the client. we should look into that separately since it does feel like that currently we are missing some stuff.

aharvard:
> No, for now we're not changing what we are actually sending to the LLM; just what gets send to the client. we should look into that separately since it does feel like that currently we are missing some stuff.

Oh, I see! OK, cool.

Is there an easy way to inspect the client data in goose UI?

DOsinga:
depends on what you want to see. creating a diagnostics report is probably the  best thing you can do. it gives you an export of the current conversation and some jsonl files of recent stuff we sent to the providers

aharvard:
> depends on what you want to see. creating a diagnostics report is probably the best thing you can do. it gives you an export of the current conversation and some jsonl files of recent stuff we sent to the providers

Thanks! Had never clicked on the bug button before (embarrassed to admit) ... for folks following along:
image

Issue #6078, opened 2025-12-11
Contextual Parameter Passing to Sub Recipes is broken


dgole:
**Describe the bug**

I'm trying to get context based parameter passing working.  I'm trying to replicate the [travel planner shown in the docs](https://dev-guides.sqprod.co/docs/tools/goose/guides/subrecipes#context-based-parameter-passing).  I've implemented the main recipe and two subrecipes locally at ~/.config/goose/recipes/.  I can run the recipe but it doesn't seem like the parameter passing actually works.  The main recipe does correctly set the param for the subtask, but then when the subtask actually runs it doesn't look like the parm is set for it.   Then the main recipe goes rogue and ignores the subrecipes and does the travel planning task itself.  

Here is an example session: goose://sessions/e4e75a80-aee0-4790-9001-361fbd173d82

This doc on FizzBuzz with subagents shows the same behavior:
https://docs.google.com/document/d/16fcYvIvF-3UPQ5A3CCKiqkGyJuOmFdArnjPPjho6Ulk/edit?tab=t.0

Slack thread with more detail:
https://sq-block.slack.com/archives/C06PBQ52MCK/p1765415277029179

---

**To Reproduce**
Steps to reproduce the behavior: Follow [your own guide on implementing context based parameter passing](https://block.github.io/goose/docs/guides/recipes/subrecipes/#context-based-parameter-passing) and observe that it doesn't work right.  

---

**Expected behavior**
Subrecipes should receive the context based params that the main agent passes and be able to act on them.  Currently that never works and the main agent creates dynamic subtasks to try to achieve the same behavior rather than actually using subrecipes as intended.  

---

**Screenshots**

<img width="996" height="1244" alt="Image" src="https://github.com/user-attachments/assets/8dafe74b-f06e-4459-b02c-3818390fe50b" />

---

**Please provide the following information**
- **OS & Arch:** Apple M1 Max Tahoe 26.1
- **Interface:** Both Goose Desktop and CLI
- **Version:** Version 1.15.0-block.202511212016-e9b9d (1.15.0-block.202511212016-e9b9d)
- **Extensions enabled:** Doesn't Matter
- **Provider & Model:** Anthropic goose-claude-4-5-sonnet

---

**Additional context**
This is a bummer, I really want to use this and its frustrating that the team seemed to fully build and document this and everything but then apparently never actually tested it.

DOsinga:
seems fixed on main

PR #6084, opened 2025-12-11
Prompt fix to encourage waiting for params from previous subrecipe

katzdave:
Fixes https://github.com/block/goose/issues/6078.

I don't see a mechanism where we would pipe the output from one recipe directly to a parameter of another, so this encourages the agent to think critically about if there might be a dependency here. If there is, it should run the subrecipes one at a time, and fill the parameters on the subsequent subrecipe afterwards.


katzdave:
The first subrecipe itself still doesn't work for the weather; it complains it doesn't have access to a useful tool so maybe we should update the docs on that. But the second recipe works now and has the weather context from the first.

katzdave:
Closing as I see a larger subrecipe change on the way, and I feel like this is something the agent should be able to correct for after seeing a task for on its own.

PR #6095, opened 2025-12-12
feat: add mcp app renderer

aharvard:
## Summary

Implements the **front-end foundation for MCP Apps** in Goose Desktop, following the [MCP Apps specification](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx). This PR sets up the rendering infrastructure, communication bridge, and mock handlers—preparing for a subsequent PR that will connect to actual MCP server resources.

Related: https://github.com/block/goose/discussions/6069

### New Components (`ui/desktop/src/components/McpApps/`)

- **`McpAppRenderer.tsx`** — Renders MCP App UI resources in sandboxed iframes with security boundaries
- **`useSandboxBridge.ts`** — React hook managing the postMessage-based JSON-RPC 2.0 communication between host (Goose) and guest (iframe), including:
  - Sandbox initialization lifecycle (`sandbox-ready` → `initialize` → `initialized`)
  - Host context propagation (theme, viewport, locale, timezone, platform)
  - Tool lifecycle notifications (`tool-input`, `tool-input-partial`, `tool-result`, `tool-cancelled`)
  - Resource teardown on unmount
- **`types.ts`** — TypeScript types for JSON-RPC messages, MCP App resources (`ui://` scheme), host context, and all guest message types
- **`utils.ts`** — Message creation helpers and **stubbed handlers** for MCP protocol methods
- **`mockAppData.ts`** — Demo MCP App HTML with interactive buttons to test all protocol methods

### Spec Alignment

- Uses `ui://` URI scheme for UI resources
- Implements `text/html;profile=mcp-app` MIME type
- Supports CSP metadata (`connectDomains`, `resourceDomains`, `prefersBorder`)
- Full initialization handshake per spec
- Host context with theme, viewport, locale, platform, device capabilities
- CSP enforcement on the outer sandbox proxy

## Current State

The `McpAppRenderer` integration in `ToolCallWithResponse.tsx` is **commented out** because Goose needs to dynamically fetch UI resources via `resources/read` when a tool's `_meta.ui.resourceUri` is present. The mock data demonstrates the expected flow.

## Handlers Status

| Method | Status | Action Required |
|--------|--------|-----------------|
| `ui/open-link` | ✅ Implemented | — |
| `ui/message` | ✅ Implemented | — |
| `ui/notifications/size-changed` | ✅ Implemented | — |
| `tools/call` | 🔜 Stubbed | Forward to MCP server; enforce `"app"` visibility |
| `resources/list` | 🔜 Stubbed | Return available resources from MCP server |
| `resources/read` | 🔜 Stubbed | Fetch resource content from MCP server |
| `resources/templates/list` | 🔜 Stubbed | Return resource templates from MCP server |
| `prompts/list` | 🔜 Stubbed | Return available prompts from MCP server |
| `notifications/message` | 🔜 Stubbed | Forward log message to MCP server |
| `ping` | 🔜 Stubbed | Forward to MCP server and return response |

## Future: @mcp-ui/client SDK

The `useSandboxBridge` hook is a temporary implementation. Once the `@mcp-ui/client` package provides its `AppRenderer` component, we can refactor to pass props directly to that component—which will internalize the sandbox communication logic. The handler functions in `utils.ts` are designed to be reusable and can be plugged directly into the SDK component's callback props.

## Next Steps (Subsequent PRs)

1. **Resource fetching** — Wire up `resources/read` to fetch UI resources when `_meta.ui.resourceUri` is present on tool results
2. **Handler integration** — Connect stubbed handlers to forward requests to the appropriate MCP server
3. **Tool visibility** — Implement filtering for `["model"]` vs `["app"]` visibility
4. **Host theming** — Provide `hostContext.styles.variables` with standardized CSS custom properties (colors, typography, border radius) so MCP Apps can match Goose's visual style
5. **Custom fonts** — Provide `hostContext.styles.css.fonts` with `@font-face` rules for Goose's fonts so MCP Apps can use them


crates/goose-server/src/auth.rs
```diff
@@ -10,7 +10,10 @@ pub async fn check_token(
     request: Request,
     next: Next,
 ) -> Result<Response, StatusCode> {
-    if request.uri().path() == "/status" || request.uri().path() == "/mcp-ui-proxy" {
+    if request.uri().path() == "/status"
+        || request.uri().path() == "/mcp-ui-proxy"
+        || request.uri().path() == "/mcp-app-proxy"
+    {
```
DOsinga, line 16: is three the time we introduce an allow list?

crates/goose-server/src/routes/mcp_app_proxy.rs
```diff
@@ -0,0 +1,53 @@
+use axum::{
```
DOsinga, line 1: this is very similar to mcp_ui_proxy.rs

I would suggest to just change the route to

/proxy/mcp_app and /proxy/mcp_ui (or /mcp-proxy/ui etc)

and then just serve the relvant HMTL based on the parameter

ui/desktop/src/components/McpApps/MCP_APPS_GLOSSARY.md
```diff
@@ -0,0 +1,410 @@
+# MCP Apps Protocol Glossary
```
DOsinga: do we need this to get merged or is it only a working document useful while working on this feature? if the first, I worry how we keep it up to date with the emerging standard

ui/desktop/src/components/ToolCallWithResponse.tsx
```diff
@@ -96,6 +98,30 @@ export default function ToolCallWithResponse({
             return null;
           }
         })}
+
+      {/* MCP Apps - This data will be coming from a resources/read result. */}
+      {mockResourceReadResult.contents.map((content) => {
+        // Transform Goose's internal tool result format to MCP Apps spec format
+        // Goose internal: { status: "success", value: [...content] } or { status: "error", error: "..." }
+        // MCP Apps spec: { content: [...], structuredContent?: {...}, _meta?: {...}, isError?: boolean }
```
DOsinga: this should be "upgraded" on main, we now have that format I think

ui/desktop/src/components/McpApps/useSandboxBridge.ts
```diff
@@ -0,0 +1,325 @@
...
+  const [iframeHeight, setIframeHeight] = useState(200);
+  const [isSandboxReady, setIsSandboxReady] = useState(false);
+  const [isGuestInitialized, setIsGuestInitialized] = useState(false);
+
+  // Fetch proxy URL on mount
+  useEffect(() => {
+    fetchMcpAppProxyUrl().then(setProxyUrl);
+  }, []);
+
+  // Reset state when resource changes
```
DOsinga: drop the comment

crates/goose-server/src/routes/templates/mcp_app_proxy.html
```diff
@@ -0,0 +1,197 @@
...
+            "object-src 'none'",
+            "base-uri 'self'"
+          ];
+
+          return directives.join('; ');
+        }
+
+        /**
+         * Create the Guest UI iframe with the provided HTML and CSP.
+         */
```
DOsinga: this comment doesn't add much

ui/desktop/src/components/McpApps/useSandboxBridge.ts
```diff
@@ -0,0 +1,325 @@
...
+          return;
+
+        case 'ui/initialize': {
+          if (!('id' in data) || data.id === undefined) return;
+          console.log('🐛 McpAppRenderer: Guest UI requesting initialization');
+          const request = data as JsonRpcRequest;
+
+          // Build host context with all fields from spec
+          const iframe = iframeRef.current;
+          const hostContext: HostContext = {
```
DOsinga: you don't need to type this

ui/desktop/src/components/McpApps/utils.ts
```diff
@@ -0,0 +1,152 @@
...
+    params: hostContext as Record<string, unknown>,
+  };
+}
+
+/**
+ * Create an initialize response with host capabilities and context.
+ */
+export function createInitializeResponse(
+  requestId: string | number,
+  hostContext?: HostContext
```
DOsinga: this doesn't need to be optional

ui/desktop/src/components/McpApps/utils.ts
```diff
@@ -0,0 +1,152 @@
...
+): JsonRpcResponse {
+  return {
+    jsonrpc: '2.0',
+    id: requestId,
+    result: {
+      protocolVersion: '2025-06-18', // THIS IS THE MCP APP PROTOCOL VERSION GOOSE SUPPORTS
```
DOsinga: probably make this a constant then rather than comment

ui/desktop/src/components/McpApps/useSandboxBridge.ts
```diff
@@ -0,0 +1,325 @@
...
+      iframe.contentWindow.postMessage(message, '*');
+    }
+  }, []);
+
+  const flushPendingMessages = useCallback(() => {
+    pendingMessagesRef.current.forEach((msg) => sendToSandbox(msg));
+    pendingMessagesRef.current = [];
+  }, [sendToSandbox]);
+
+  const handleMessage = useCallback(
```
DOsinga: I think you are fighting the type system here. if you would just declare the types as something like

```
interface SandboxReadyNotification {
  jsonrpc: '2.0';
  method: 'ui/notifications/sandbox-ready';
}

interface InitializeRequest {
  jsonrpc: '2.0';
  id: string | number;
  method: 'ui/initialize';
}

...

type IncomingGuestMessage =
  | SandboxReadyNotification
  | InitializeRequest
  | InitializedNotification
  ...
```

you could then just switch on method and get this all for free:

```
switch (msg.method) {
    case 'ui/notifications/size-changed':
      setIframeHeight(Math.max(200, msg.params.height));
      break;
```

ui/desktop/src/components/McpApps/utils.ts
```diff
@@ -0,0 +1,152 @@
...
+    if (baseUrl && secretKey) {
+      return `${baseUrl}/mcp-app-proxy?secret=${encodeURIComponent(secretKey)}`;
+    }
+    console.error('Failed to get goosed host/port or secret key');
+    return null;
+  } catch (error) {
+    console.error('Error fetching MCP App Proxy URL:', error);
+    return null;
+  }
+}
```
DOsinga: we should find a way to do this through the API instead of hand composing this, adding the secret, checking the secret by hand. for one thing we could extend the auto checker of the secret to also look in, say __secret

ui/desktop/src/components/McpApps/utils.ts
```diff
@@ -0,0 +1,152 @@
...
+/**
+ * Get the current theme from localStorage.
+ */
+export function getCurrentTheme(): 'light' | 'dark' {
+  const useSystemTheme = localStorage.getItem('use_system_theme') === 'true';
+  if (useSystemTheme) {
+    return window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
+  }
+  return localStorage.getItem('theme') === 'dark' ? 'dark' : 'light';
+}
```
DOsinga: this looks like it is copied from the brittle code that handles that on the other side. we should generalize

ui/desktop/src/components/McpApps/types.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export type JsonRpcMessage = JsonRpcNotification | JsonRpcRequest | JsonRpcResponse;
+
+// MCP App resource type
+export interface McpAppResource {
+  uri: `ui://${string}`;
+  description?: string;
+  mimeType: `text/html;profile=mcp-app`;
+  text?: string;
```
DOsinga: should this be optional?

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
```diff
@@ -0,0 +1,83 @@
...
+ * Temporary Goose implementation while waiting for official SDK components.
+ *
+ * @see SEP-1865 https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx
+ */
+
+import { useSandboxBridge } from './useSandboxBridge';
+import { McpAppResource, ToolInput, ToolResult } from './types';
+import { cn } from '../../utils';
+
+export type { McpAppResource, ToolInput, ToolResult } from './types';
```
DOsinga: you are importing this twice? does that compile?

crates/goose-server/src/auth.rs
aharvard: `/mcp-ui-proxy` will eventually go away, and I don't have strong opinions here. Maybe if we start opening up more routes?

ui/desktop/src/components/McpApps/MCP_APPS_GLOSSARY.md
aharvard: Using as a tool to help me work on the feature and communicate what's going on... I was confused by all the method names. I want to remove before merging.

- [x] remove MCP_APPS_GLOSSARY.md prior to merge

crates/goose-server/src/routes/mcp_app_proxy.rs
aharvard: Whoops... Yesterday, I dug pretty deep into the iframe approach and realized I messed some things up. I think I got it right now. I just pushed [this new commit](https://github.com/block/goose/pull/6095/changes/f4d36fd6b1b1ad7b3e47f88dfaf5d90f1b181f21). Sorry for the churn.

Before: The outer sandbox had a static, permissive CSP that allowed any HTTPS domain, relying on JavaScript to build and inject a restrictive CSP into the inner iframe.

After: The outer sandbox CSP is now dynamically built in Rust based on the app's declared domains, making it the single source of truth — the inner iframe needs no CSP since it can't exceed the outer's restrictions.

Based on these changes, `mcp_app_proxy` is now unique from `mcp_ui_proxy` — and implemented closer to spec.

Do we consolidate code into a single route that handles both MCP-UI and MCP Apps (with CSP construction) HTML, or keep them as separate files? Thinking ahead to the MCP-UI deprecation, can we keep them separate?

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
aharvard: No, that second one is an export. I used to rely on a barrel file approach, where I used these types in other places. Besides being pretty sloppy, we don't need this anymore. Removed.

ui/desktop/src/components/McpApps/utils.ts
aharvard: Good call. I added a constant to the file. 

And for clarity, this is actually the MCP protocol version; my comment had an error about it being specifically the MCP App protocol version. 

Since the protocol version is part of the MCP initialization lifecycle, should we pull from a single source for all Goose? Does that sound interesting / possible?

ui/desktop/src/components/ToolCallWithResponse.tsx
aharvard: cool, I'll take a look.

- [x] Leverage updated format for tool results merged to main recently

ui/desktop/src/components/McpApps/types.ts
aharvard: Yeah, it can be `text` or `blob`

```
text?: string;                // HTML content as string
blob?: string;                // OR base64-encoded HTML
```

Refer to https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx#ui-resource-format

Technically, it probably should be a discriminated union, but that feels a little bit heavy.

Keeping `text` as optional and also adding optional `blob`.

ui/desktop/src/components/McpApps/useSandboxBridge.ts
aharvard: I tried removing it and noticed that TypeScript infers the `displayMode` of `inline` as a `string` rather than the literal union types defined in the type. 

We would need to either write `as const` next to the `displayMode` property or `createInitializeResponse(msg.id, hostContext as HostContext)`.

To me, being explicit here feels cleaner than casting. Also, I think the DX is a little bit better, too. You get immediate feedback on how the const is either missing properties or using the wrong values.

ui/desktop/src/components/McpApps/utils.ts
aharvard: good catch, will fix!

ui/desktop/src/components/McpApps/useSandboxBridge.ts
aharvard: Much better! Thanks.

ui/desktop/src/components/McpApps/utils.ts
aharvard: Yeah, maybe some sort of React hook?


- [x] Make theme fetching better.

ui/desktop/src/components/McpApps/utils.ts
aharvard: Do you think we should address that as a fast follow-up or incorporate it into this PR?

ui/desktop/src/components/ToolCallWithResponse.tsx
aharvard: sweet! 

image

aharvard: modified slightly, now passing in`value`
image

ui/desktop/src/components/McpApps/utils.ts
aharvard: opened a stand-alone PR for it: https://github.com/block/goose/pull/6137

aharvard: merged https://github.com/block/goose/pull/6137 and folded in

crates/goose-server/src/routes/templates/mcp_app_proxy.html
```diff
@@ -0,0 +1,135 @@
...
+            createGuestIframe(html);
+            return;
+          }
+
+          // Forward all other messages to Guest UI (if it exists)
+          // This includes lifecycle messages like responses to ui/initialize
+          if (guestIframe && guestIframe.contentWindow) {
+            guestIframe
+              .contentWindow
+              .postMessage(data, '*');
```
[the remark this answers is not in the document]
aharvard: won't address

ui/desktop/src/components/McpApps/useSandboxBridge.ts
```diff
@@ -0,0 +1,352 @@
...
+  useEffect(() => {
+    setIsSandboxReady(false);
+    setIsGuestInitialized(false);
+    pendingMessagesRef.current = [];
+  }, [resourceUri]);
+
+  const sendToSandbox = useCallback((message: JsonRpcMessage) => {
+    const iframe = iframeRef.current;
+    if (iframe?.contentWindow) {
+      iframe.contentWindow.postMessage(message, '*');
```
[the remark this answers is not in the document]
aharvard: won't address

crates/goose-server/src/routes/templates/mcp_app_proxy.html
```diff
@@ -0,0 +1,135 @@
...
+        window.addEventListener('message', handleMessage);
+
+        // Notify Host that Sandbox is ready
+        window
+          .parent
+          .postMessage({
+            jsonrpc: '2.0',
+            method: 'ui/notifications/sandbox-ready',
+            params: {}
+          }, '*');
```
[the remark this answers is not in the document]
aharvard: won't address

crates/goose-server/src/routes/templates/mcp_app_proxy.html
```diff
@@ -0,0 +1,135 @@
...
+          var data = event.data;
+          if (!data || typeof data !== 'object') {
+            return;
+          }
+
+          // Forward all messages from Guest UI to Host
+          // The Sandbox does NOT create its own requests - just relays
+          window
+            .parent
+            .postMessage(data, '*');
```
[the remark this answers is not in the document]
aharvard: won't address

ui/desktop/src/components/McpApps/useSandboxBridge.ts
```diff
@@ -0,0 +1,352 @@
...
+    }
+  }, []);
+
```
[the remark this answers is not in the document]
aharvard: won't address

ui/desktop/src/components/McpApps/useSandboxBridge.ts
```diff
@@ -0,0 +1,352 @@
...
+  // Send resource teardown request when component unmounts
+  useEffect(() => {
+    const currentSendToSandbox = sendToSandbox;
+    const checkInitialized = () => isGuestInitialized;
+
+    return () => {
+      if (checkInitialized()) {
+        const { message } = createResourceTeardownRequest('Component unmounting');
+        currentSendToSandbox(message);
```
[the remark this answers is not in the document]
aharvard: addressed as part of other feedback

ui/desktop/src/components/McpApps/types.ts
```diff
@@ -0,0 +1,278 @@
...
+export interface ToolInputPartial {
+  arguments: Record<string, unknown>;
+}
+
+/** Tool result passed to the MCP App (matches MCP CallToolResult) */
+export interface ToolResult {
+  _meta?: Record<string, unknown>;
+  content: unknown[];
+  isError?: boolean;
+  structuredContent?: Record<string, unknown>;
```
DOsinga: no. these types need to come from the server, no saying you copied it

ui/desktop/src/components/McpApps/utils.ts
```diff
@@ -0,0 +1,414 @@
...
+ * This is a notification, so no response is sent.
+ * Returns a handler function that updates iframe height.
+ */
+export function handleSizeChanged(setIframeHeight: (height: number) => void) {
+  return (msg: SizeChangedNotification): null => {
+    const newHeight = Math.max(DEFAULT_IFRAME_HEIGHT, msg.params.height);
+    setIframeHeight(newHeight);
+    return null;
+  };
+}
```
DOsinga, line 37: I still don't think this file is adding much. these are simple types, creating constructors for them like this just loses type information

PR #6096, opened 2025-12-12
Keep both instructions and prompts when using subrecipes

katzdave:
Fixes https://github.com/block/goose/issues/6078

The parameter passing was working fine, but we were only keeping instructions and not prompt which just told it it was a weather specialist and didn't have the actual parameter injection about the location it should look up. 

recipe had:

```
instructions: |
  You are a weather data specialist. Gather current weather information
  including temperature, conditions, and seasonal context.
prompt: |
  Get the current weather for {{ location }}.
```

### DIFF
--- a/crates/goose/src/scheduler.rs
+++ b/crates/goose/src/scheduler.rs
@@ -749,11 +749,12 @@ async fn execute_job(
         job_def.current_session_id = Some(session.id.clone());
     }
 
-    let prompt_text = recipe
-        .prompt
-        .as_ref()
-        .or(recipe.instructions.as_ref())
-        .unwrap();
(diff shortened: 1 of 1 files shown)


jamadeo (approved):

jamadeo:
hmm @tlongwell-block has a related change #6121, maybe that is the way to go?

tlongwell-block:
> hmm @tlongwell-block has a related change #6121, maybe that is the way to go?

I think #6121 is a bit more comprehensive

alexhancock:
@tlongwell-block @katzdave I am looking to merge or close open PRs. Is this one still what we want to do or has #6121 addressed the same thing?

katzdave:
Oops no worries, agree lets go with #6121

PR #6098, opened 2025-12-13
fix: use system not developer role in db

baxen:
Databricks natively provided openai models use "system" not "developer". The gateway can also pass through to first party API, where new models are supposed to use "developer". However openai automatically converts system to developer when necessary, so this works with both

### DIFF
--- a/crates/goose/src/providers/formats/databricks.rs
+++ b/crates/goose/src/providers/formats/databricks.rs
@@ -533,12 +533,7 @@ pub fn create_request(
     };
 
     let system_message = DatabricksMessage {
-        role: if is_openai_reasoning_model {
-            "developer"
-        } else {
-            "system"
-        }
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/formats/databricks.rs
```diff
@@ -1024,7 +1019,7 @@ mod tests {
             "model": "o1",
             "messages": [
                 {
-                    "role": "developer",
+                    "role": "system",
```
DOsinga: this makes this a good candidate to just delete the test

all we really want to test here is that reasoning_effort comes back as medium. anything else we test just suffers from if other things change (like here) it breaks

baxen: yeah good call, consolidated!

Issue #6047, opened 2025-12-10
Desktop: Custom providers with empty API Key considered "not configured"


dianed-square:
**Describe the bug**

CLI and Desktop handle "no auth required" custom providers differently. E.g. when creating an Ollama-compatible custom provider via CLI, users can press `Enter` to skip the API key prompt, which stores an empty string (""). This provider is then accessible in CLI, but Desktop's `is_configured` logic checks `!s.is_empty()`, causing these providers to appear "not configured" in the Desktop UI even though they're fully functional.

---

**To Reproduce**
1. Run `goose configure`
2. Select `Custom Providers` > `Add A Custom Provider`
3. Choose `Ollama Compatible` as the API type
4. Enter provider details, and press `Enter` to skip the API key prompt (leaving it empty)
5. Complete the setup
6. Open goose Desktop > Settings > Models > Configure providers
7. Observe that the custom provider does not have a green checkmark and appears "not configured", which means it won't show up in the providers list for selection

---

**Expected behavior**
Custom providers with empty API keys (valid for local/no-auth endpoints) should show as "configured" in Desktop if they were successfully created via CLI.

---

**Screenshots**
N/A

---

**Please provide the following information**
- **OS & Arch:** macOS
- **Interface:** CLI to create, UI to observe
- **Version:** v1.16.0
- **Extensions enabled:** N/A
- **Provider & Model:** Custom Ollama-compatible provider

---

**Additional context**
The issue appears to be in Desktop's `is_configured` logic, which checks `!s.is_empty()` for the API key. An empty string (`""`) is stored in the keychain when the user skips the prompt, which fails this check.

Possible solutions:
- CLI could store a placeholder value (e.g., `"notrequired"` like Desktop does) instead of empty string
- Desktop could treat empty string as valid for Ollama-compatible providers
- Align on a convention for "no auth required" across both interfaces

PR #6105, opened 2025-12-13
fix : Custom providers with empty API keys show as configured in desktop

Abhijay007:
Closes: #6047

## PR Description

This PR aims to resolve the issue with custom providers that were created via the CLI with an empty API key (valid for local or no-auth endpoints like Ollama) were incorrectly shown as **not configured** in the Desktop UI.

This was caused by `check_provider_configured()` validating that the API key was non-empty instead of checking whether it existed in the config.

## Solution

Updated `check_provider_configured()` to treat a provider as configured if the API key exists in the config, even if it is empty.

**File modified**
- `crates/goose-server/src/routes/utils.rs`

## Type of Change
- [x] Bug fix

## Testing

- Tested in Desktop UI by creating a custom provider via CLI with an empty API key

## Screenshots / Demos

Before: 
image

 After: 
afterEmptyString

### DIFF
--- a/crates/goose-server/src/routes/utils.rs
+++ b/crates/goose-server/src/routes/utils.rs
@@ -94,13 +94,11 @@ pub fn inspect_keys(
 pub fn check_provider_configured(metadata: &ProviderMetadata, provider_type: ProviderType) -> bool {
     let config = Config::global();
 
-    // TODO(Douwe): if the provider doesn't need an API key, it should be considered configured always
     if provider_type == ProviderType::Custom || provider_type == ProviderType::Declarative {
         if let Ok(loaded_provider) = load_provider(metadata.name.as_str()) {
             return config
                 .get_secret::<String>(&loaded_provider.config.api_key_env)
-                .map(|s| !s.is_empty())
-                .unwrap_or(false);
+                .is_ok();
         }
     }
     // Special case: Zero-config providers (no config keys)


crates/goose-server/src/routes/utils.rs
```diff
@@ -94,13 +94,11 @@ pub fn inspect_keys(
...
     let config = Config::global();
 
-    // TODO(Douwe): if the provider doesn't need an API key, it should be considered configured always
     if provider_type == ProviderType::Custom || provider_type == ProviderType::Declarative {
         if let Ok(loaded_provider) = load_provider(metadata.name.as_str()) {
             return config
                 .get_secret::<String>(&loaded_provider.config.api_key_env)
-                .map(|s| !s.is_empty())
-                .unwrap_or(false);
+                .is_ok();
```
DOsinga, line 101: doesn't this lead to the situation where we think custom providers are configured that are not because they miss the API key?

angiejones (approved):

DOsinga (approved):

crates/goose-server/src/routes/utils.rs
Abhijay007: For the desktop app, we have a checkbox to control this behavior, marking it as local mode.
For the CLI, it works the same as before, if the user explicitly adds it, we treat it as added; otherwise, it is passed through as-is. I don’t think there is any explicit API-level validation or check for this in the CLI. 

image

alexhancock (approved):

Issue #6046, opened 2025-12-10
Desktop edit custom provider: Provider Type always defaults to OpenAI


dianed-square:
**Describe the bug**

When editing a custom provider in goose Desktop, the "Provider Type" dropdown always defaults to "OpenAI Compatible" regardless of the provider's actual engine type (Anthropic Compatible or Ollama Compatible). If the user saves any changes without manually re-selecting the correct provider type, the engine is overwritten to "openai", corrupting the provider configuration.

---

**To Reproduce**
1. Create a custom provider using CLI or Desktop with `Anthropic Compatible` or `Ollama Compatible` as the provider type
2. Go to Settings > Models > Configure providers
3. Click on the custom provider to edit it
4. Observe that "Provider Type" dropdown shows "OpenAI Compatible" instead of the actual type
5. Make any change (e.g., update display name) and click "Update Provider"
6. Check the provider's JSON config file in `~/.config/goose/custom_providers/` - the `engine` field is now `"openai"` instead of the original value

---

**Expected behavior**
The "Provider Type" dropdown should display the provider's actual engine type when editing, and saving changes should preserve the original engine unless the user explicitly changes it.

---

**Screenshots**
N/A

---

**Please provide the following information**
- **OS & Arch:** macOS
- **Interface:** UI (goose Desktop)
- **Version:** v1.15.0 and v1.16.0 (confirmed in both)
- **Extensions enabled:** N/A
- **Provider & Model:** Custom provider (Anthropic Compatible or Ollama Compatible)

---

**Additional context**
This is a data corruption bug. Users may unknowingly break their custom provider configurations by making unrelated edits. The issue appears to be in the Desktop UI's edit form not properly loading the existing engine value into the dropdown.

PR #6106, opened 2025-12-13
fix : preserve provider engine type when editing custom providers

Abhijay007:
Closes: #6046

## PR Description

When editing a custom provider in Goose Desktop, the **Provider Type** dropdown always defaulted to **OpenAI Compatible**, regardless of the provider’s actual engine (Anthropic or Ollama). Saving the form without reselecting the correct type overwrote the engine to `openai`, resulting in an incorrect configuration.

This issue was caused by a mismatch between frontend engine values (using `_compatible` suffixes) and the backend’s expected engine values.

## Changes Made

file modified: `ui/desktop/src/components/settings/providers/modal/subcomponents/forms/CustomProviderForm.tsx`
- Updated engine values to match backend (`openai`, `anthropic`, `ollama`)
- Fixed default engine initialization using `initialData.engine`
- Removed unnecessary mapping logic in `useEffect`
- Updated dropdown options and label logic accordingly

file modified: `ui/desktop/src/components/settings/providers/ProviderGrid.tsx`
- Removed redundant `.toLowerCase()` call when passing the engine value

## Type of Change
- [x] Bug fix

## Testing

- Tested in Desktop UI by editing and saving custom OpenAI , Anthropic , and Ollama-based providers

## Screenshots / Demos

**Before:** 

beforeDropDown

**After:**

afterDropDown

### DIFF
--- a/ui/desktop/src/components/settings/providers/modal/subcomponents/forms/CustomProviderForm.tsx
+++ b/ui/desktop/src/components/settings/providers/modal/subcomponents/forms/CustomProviderForm.tsx
@@ -19,7 +19,7 @@ export default function CustomProviderForm({
   initialData,
   isEditable,
 }: CustomProviderFormProps) {
-  const [engine, setEngine] = useState('openai');
+  const [engine, setEngine] = useState('openai_compatible');
   const [displayName, setDisplayName] = useState('');
   const [apiUrl, setApiUrl] = useState('');
   const [apiKey, setApiKey] = useState('');
@@ -30,7 +30,12 @@ export default function CustomProviderForm({
 
   useEffect(() => {
     if (initialData) {
-      setEngine(initialData.engine);
+      const engineMap: Record<string, string> = {
+        openai: 'openai_compatible',
+        anthropic: 'anthropic_compatible',
+        ollama: 'ollama_compatible',
+      };
+      setEngine(engineMap[initialData.engine] || 'openai_compatible');
       setDisplayName(initialData.display_name);
       setApiUrl(initialData.api_url);
       setModels(initialData.models.join(', '));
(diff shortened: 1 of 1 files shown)


DOsinga (commented):
the description of the PR says that this is due to a mismatch of the backend and the frontend and the _compatible suffix, but what the PR does is remove the toLowercase (which seems reasonable) - is this what you wanted to do?

Abhijay007:
> the description of the PR says that this is due to a mismatch of the backend and the frontend and the _compatible suffix, but what the PR does is remove the toLowercase (which seems reasonable) - is this what you wanted to do?

Oh yes, that was the issue I initially suspected. However, I received a review from Copilot mentioning that this change would break the backend, as the backend would reject it with an “Invalid provider type” error.

I looked into it further and found that the problem was due to a lowercase mismatch, the values were being sent in lowercase, which caused the backend validation to fail. Because of this, I reverted those changes and pushed a fix.

For reference, you can check the Copilot review comments.

DOsinga:
hmm, not sure I follow still; does the backend use different values from what it sends and what it receives? if that is the case, can't we just fix that in the backend?

Abhijay007:
> hmm, not sure I follow still; does the backend use different values from what it sends and what it receives? if that is the case, can't we just fix that in the backend?

The issue is that the backend expects engine values with the `_compatible suffix` (for example, `openai_compatible`, `anthropic_compatible`, `ollama_compatible`). The Rust code in `crates/goose/src/config/declarative_providers.rs`  explicitly matches against these exact strings.

If we change these values to `openai`, `anthropic`, or `ollama`, the backend rejects them with an “Invalid provider type” error. So removing the suffix would break the backend.

I revisited the issue and found that the reason the provider was defaulting back incorrectly was not due to the suffix, but because the value was being converted to lowercase, so instead of making changes in the backend (which may have affected other things) I changed the lowercase value conversion. This caused a mismatch, making the backend think it was receiving an invalid engine value. As a result, the UI fell back to showing the engine type from which the custom provider was originally created.

In short, we don’t need to update or remove any suffix values. The fix was simply to remove the lowercase conversion that was causing the mismatch and backend validation error in provider grid and provider form.

DOsinga:
but can't we change the backend so it accepts this too?

Abhijay007:
> but can't we change the backend so it accepts this too?

I don't think that is needed, as that might affect the backend api response, which might be getting used at other places too, and if we update that for this change, we might need to look into that as well, and the current change will not affect it

PR #6112, opened 2025-12-14
feat: subagent personalities

tlongwell-block:
Add built-in investigator and planner subagent personalities

Subagents can now use two specialist modes without defining custom subrecipes:
• investigator - Read-only research into codebases and systems
• planner - Strategic decomposition before complex tasks

Additional personalities can be added later by placing recipes in crates/goose/src/prompts/subrecipes/

Usage: `subrecipe: "investigator"` or `subrecipe: "planner"` with instructions passed via the `instructions` parameter.

**Implementation:**
- Extends `prompt_template.rs` to serve bundled subrecipes from existing `CORE_PROMPTS_DIR`
- Adds `investigator.yaml` and `planner.yaml` to `src/prompts/subrecipes/`
- Extracts `params_to_values()` and `combine_instructions()` helpers to eliminate duplication in subagent_tool.rs
- Adds `max_turns` field to Recipe struct/builder for turn limits in subrecipes

**Why this approach:**
Reuses existing prompt infrastructure rather than creating parallel systems. Bundled subrecipes are discovered via `iter_bundled_subrecipes()` and built with the same `build_recipe_from_template()` path as user-defined subrecipes.

Builds on the work in https://github.com/block/goose/pull/5893

### DIFF
--- a/ui/desktop/src/api/types.gen.ts
+++ b/ui/desktop/src/api/types.gen.ts
@@ -631,6 +631,7 @@ export type Recipe = {
     description: string;
     extensions?: Array<ExtensionConfig> | null;
     instructions?: string | null;
+    max_turns?: number | null;
     parameters?: Array<RecipeParameter> | null;
     prompt?: string | null;
     response?: Response | null;


tlongwell-block:
Probably should reconsider this in the context of recipes/skills later

DOsinga:
yes. maybe the two of us can have a little chat about this and figure out the diverse dimensions of skills, recipes, claude plugins, apps and agents.md include hierarchies and come up with a plan?

PR #6119, opened 2025-12-15
Summarize old tool calls

DOsinga:
## Summary

Implements tool call/response compacting on the fly. does in a task so should not increase latency. also makes sure that all messages actually have a message_id that is written to the db and retrieved

### DIFF
--- a/crates/goose/src/context_mgmt/mod.rs
+++ b/crates/goose/src/context_mgmt/mod.rs
@@ -712,10 +712,11 @@ mod tests {
                 Message::assistant()
                     .with_tool_request(
                         call_id,
-                        Ok(CallToolRequestParam {
+                        Ok(CallToolRequestParams {
                             task: None,
                             name: tool_name.to_string().into(),
                             arguments: None,
+                            meta: None,
                         }),
                     )
                     .with_id(call_id),
--- a/crates/goose/src/session/session_manager.rs
+++ b/crates/goose/src/session/session_manager.rs
@@ -587,6 +587,7 @@ impl SessionStorage {
             r#"
             CREATE TABLE messages (
                 id INTEGER PRIMARY KEY AUTOINCREMENT,
+                message_id TEXT,
                 session_id TEXT NOT NULL REFERENCES sessions(id),
                 role TEXT NOT NULL,
                 content_json TEXT NOT NULL,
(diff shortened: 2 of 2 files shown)


crates/goose/src/context_mgmt/mod.rs
```diff
@@ -414,6 +418,114 @@ fn format_message_for_compacting(msg: &Message) -> String {
...
+    let mut tool_call_count = 0;
+    let mut first_tool_call_id = None;
+
+    for msg in messages.iter() {
+        if !msg.is_agent_visible() {
+            continue;
+        }
+
+        for content in &msg.content {
+            if let MessageContent::ToolRequest(req) = content {
```
katzdave, line 439: Should we filter on the response size? Is it worth bothering trying to summarize if its very small.

crates/goose/src/agents/agent.rs
```diff
@@ -1327,6 +1342,37 @@ impl Agent {
...
+                        for msg in matching {
+                            msg.metadata = msg.metadata.with_agent_invisible();
+                            if let Some(id) = &msg.id {
+                                SessionManager::update_message_metadata(&session_config.id, id, |metadata| {
+                                    metadata.with_agent_invisible()
+                                }).await?;
+                            }
+                        }
+                        conversation = Conversation::new_unvalidated(updated_messages);
+                        messages_to_add.push(summary_msg);
```
katzdave, line 1560: Is this summary getting added to the end? (rather than where the original tool call was?)

crates/goose/src/context_mgmt/mod.rs
DOsinga: I tried to measure; an empty tool call/response is 300 tokens, a summary of one sentence (an ls was executed and three files, one of them readme, was found) is about 100, so even that does save something

crates/goose/src/agents/agent.rs
DOsinga: well, sort of. the summary message gets the same timestamp as the toolcall, so it should sort the right way

katzdave: ah got it, missed the timestamp sorting.

zanesq:
@DOsinga still want to get this in?

DOsinga:
yes I do. been going back and forth on how it exactly should work, let's see what we can do today

PR #6121, opened 2025-12-15
fix: use instructions for system prompt and prompt for user message in subagents

tlongwell-block:
Fixes incorrect handling of `recipe.instructions` and `recipe.prompt` in the subagent system. Previously, both fields were merged together and used for BOTH the system prompt AND the first user message, which was redundant and semantically incorrect.

## The Bug

In `subagent_tool.rs`, the `build_subrecipe()` function merged `prompt` into `instructions`:
```rust
// Before: prompt was appended to instructions
combined.push_str(prompt);
recipe.instructions = Some(combined);
```

Then in `subagent_handler.rs`, this merged blob was used for both:
- The system prompt template (`task_instructions`)
- The user message (`Message::user().with_text(...)`)

## The Fix

- **`instructions`** → system prompt (agent personality/role/guidelines)
- **`prompt`** → user message (the actual task to execute)

This matches how the CLI handles recipes in `extract_from_cli.rs`.

## Changes

| File | Change |
|------|--------|
| `subagent_tool.rs` | Remove merge logic; only append parent's extra instructions |
| `subagent_handler.rs` | Use `instructions` for system prompt, `prompt` for user message |
| `subagent_system.md` | Conditional rendering for empty instructions |

## Backward Compatibility

- Recipes with only `instructions` still work (used for both, as before)
- Recipes with only `prompt` still work
- Ad-hoc subagents (instructions parameter only) still work
- Augmented mode (subrecipe + extra instructions) still works

### DIFF
--- a/crates/goose/src/agents/subagent_handler.rs
+++ b/crates/goose/src/agents/subagent_handler.rs
@@ -121,8 +121,7 @@ fn get_agent_messages(
         let user_task = recipe
             .prompt
             .clone()
-            .or(recipe.instructions.clone())
-            .ok_or_else(|| anyhow!("Recipe has no instructions or prompt"))?;
+            .unwrap_or_else(|| "Begin.".to_string());
 
         let agent_manager = AgentManager::instance()
             .await


tlongwell-block:
/goose give me a minimal review on this change

crates/goose/src/agents/subagent_handler.rs
```diff
@@ -117,10 +117,11 @@ fn get_agent_messages(
...
 ) -> AgentMessagesFuture {
     Box::pin(async move {
-        let text_instruction = recipe
-            .instructions
+        let system_instructions = recipe.instructions.clone().unwrap_or_default();
+        let user_task = recipe
+            .prompt
             .clone()
-            .or(recipe.prompt.clone())
+            .or(recipe.instructions.clone())
```
DOsinga: this should probably not also default to instructions, since a recipe might not have a prompt, in which case we should probably just say, ehm, "do it" ?

DOsinga (approved):
I think goose missed this?

crates/goose/src/agents/subagent_handler.rs
tlongwell-block: Great catch. This was just the previous behavior and is not great. I've changed it to just say "Begin." as the first user message when there is no prompt, but there are instructions.

PR #6122, opened 2025-12-15
Add more posthog analytics (privacy aware)

zanesq:
## Summary

Expands analytics coverage across both the backend (Rust) and frontend (TypeScript) to provide better insights into feature usage and error patterns.

All frontend analytics are sent through a new endpoint and the existing posthog.rs api integration. I first attempted the posthog js library but it was having authentication issues and we reuse a lot of code anyway so might as well leave it in the backend for now.

### Key Design Decisions
- **Privacy-first**: Extension names only tracked for builtin extensions; recipe names never tracked
- **Success/failure tracking**: All operations track both success and failure with error details
- **Source attribution**: Events include `source: "backend"` or `source: "frontend"` for debugging
- **Consistent error handling**: `getErrorType()` helper standardizes error classification

#### New Events
- **`slash_command_used`** - Tracks slash command usage with:
  - `command_type`: "builtin" | "recipe" | "unknown"
  - `command_name`: Only for builtin commands (privacy-safe)
  - `success`: true for valid commands, false for unknown
- **`schedule_job_started`** / **`schedule_job_completed`** - Tracks automated scheduled job execution with duration

#### Enhanced `session_started` Event
Now includes additional properties:
- `is_dev` - Whether running in debug/dev mode
- `db_schema_version` - Current database schema version
- Router/lead model settings: `setting_mode`, `setting_max_turns`, `setting_router_enabled`, `setting_lead_model`, `setting_lead_provider`, `setting_lead_turns`, `setting_lead_failure_threshold`, `setting_lead_fallback_turns`

#### Enhanced `error` Event
- Added `error_category` classification
- Added `source: "backend"` marker
- New `ErrorContext` struct with optional `component` and `action` fields
- Added provider/model context to error events

#### New Infrastructure
- **`emit_event()`** - Generic event emitter for custom events with arbitrary properties
- **`emit_slash_command_used()`** - Dedicated slash command tracking
- **`emit_error_with_context()`** - Error tracking with additional context
- New `/telemetry/event` API endpoint for frontend-to-backend event forwarding

---

#### New Analytics Module (`utils/analytics.ts`)
Comprehensive type-safe analytics including:

**Schedule Events:**
- `schedule_created` (with source_type: file/deeplink)
- `schedule_updated`, `schedule_deleted`, `schedule_paused`, `schedule_unpaused`
- `schedule_run_now`, `schedule_killed`, `schedule_inspected`, `schedule_viewed`

**Recipe Events:**
- `recipe_created`, `recipe_imported`, `recipe_edited`, `recipe_deleted`
- `recipe_started` (with in_new_window flag)
- `recipe_deeplink_copied`
- `recipe_scheduled`, `recipe_slash_command_set` (with action: add/edit/remove)

**Extension Events:**
- `extension_added`, `extension_enabled`, `extension_disabled`, `extension_deleted`
- Includes `is_builtin` flag; `extension_name` only tracked for builtin extensions

**Chat Input Bar Events:**
- `input_file_attached` (file/directory)
- `input_voice_dictation` (start/stop/transcribed/error with duration)
- `input_mode_changed` (from/to mode)
- `input_diagnostics_opened`, `input_create_recipe_opened`, `input_edit_recipe_opened`

**Core Events:**
- `page_view` with referrer tracking
- `model_changed`, `settings_tab_viewed`, `setting_toggled`
- `telemetry_preference_set`
- Enhanced error tracking: `error_occurred`, `app_crashed`, `app_reloaded`
- Onboarding funnel: `onboarding_started`, `onboarding_provider_selected`, `onboarding_completed`, `onboarding_abandoned`, `onboarding_setup_failed`

#### New Hook (`hooks/useAnalytics.ts`)
- `usePageViewTracking()` - Automatic page view tracking on route changes

#### Components Updated
- `SchedulesView.tsx` - Schedule CRUD operations
- `ScheduleDetailView.tsx` - Schedule inspection/kill
- `RecipesView.tsx` - Recipe management
- `ChatInput.tsx` - File attachments, voice dictation, mode changes
- `BottomMenuModeSelection.tsx` - Mode switching
- `extension-manager.ts` - Extension lifecycle
- `SettingsView.tsx`, `TelemetrySettings.tsx`, `SecurityToggle.tsx`, etc. - Settings toggles
- `ErrorBoundary.tsx` - Crash tracking
- `ProviderGuard.tsx` - Provider setup tracking
- `SwitchModelModal.tsx` - Model changes
- `App.tsx` / `renderer.tsx` - Global error handlers and page view tracking

### DIFF
--- a/crates/goose/src/posthog.rs
+++ b/crates/goose/src/posthog.rs
@@ -237,18 +237,7 @@ pub fn emit_error_with_context(error_type: &str, context: ErrorContext) {
     });
 }
 
-/// Type of slash command for analytics tracking
-#[derive(Debug, Clone)]
-pub enum SlashCommandType {
-    /// Builtin command (e.g., /compact) - name is safe to track
-    Builtin(String),
-    /// Recipe command - name is NOT tracked for privacy
-    Recipe,
-    /// Unknown/unrecognized command
-    Unknown,
-}
-
-pub fn emit_slash_command_used(command_type: SlashCommandType) {
+pub fn emit_custom_slash_command_used() {
     if !is_telemetry_enabled() {
         return;
     }
@@ -256,7 +245,7 @@ pub fn emit_slash_command_used(command_type: SlashCommandType) {
     let installation = load_or_create_installation();
 
     tokio::spawn(async move {
-        let _ = send_slash_command_event(&installation, command_type).await;
+        let _ = send_custom_slash_command_event(&installation).await;
     });
 }
 
@@ -300,25 +289,10 @@ async fn send_error_event(
     client.capture(event).await.map_err(|e| format!("{:?}", e))
 }
 
-async fn send_slash_command_event(
-    installation: &InstallationData,
-    command_type: SlashCommandType,
-) -> Result<(), String> {
+async fn send_custom_slash_command_event(installation: &InstallationData) -> Result<(), String> {
     let client = posthog_rs::client(POSTHOG_API_KEY).await;
-    let mut event = posthog_rs::Event::new("slash_command_used", &installation.installation_id);
-
-    // Set success and command_type based on the enum
-    let (success, type_str, command_name) = match &command_type {
-        SlashCommandType::Builtin(name) => (true, "builtin", Some(name.clone())),
-        SlashCommandType::Recipe => (true, "recipe", None),
-        SlashCommandType::Unknown => (false, "unknown", None),
-    };
-
-    event.insert_prop("success", success).ok();
-    event.insert_prop("command_type", type_str).ok();
-    if let Some(name) = command_name {
-        event.insert_prop("command_name", name).ok();
-    }
+    let mut event =
+        posthog_rs::Event::new("custom_slash_command_used", &installation.installation_id);
 
     event.insert_prop("source", "backend").ok();
     event.insert_prop("version", env!("CARGO_PKG_VERSION")).ok();
@@ -328,14 +302,6 @@ async fn send_slash_command_event(
 
     if let Some(platform_version) = get_platform_version() {
         event.insert_prop("platform_version", platform_version).ok();
-    }
-
-    let config = Config::global();
-    if let Ok(provider) = config.get_param::<String>("GOOSE_PROVIDER") {
-        event.insert_prop("provider", provider).ok();
-    }
-    if let Ok(model) = config.get_param::<String>("GOOSE_MODEL") {
-        event.insert_prop("model", model).ok();
     }
 
     client.capture(event).await.map_err(|e| format!("{:?}", e))
--- a/ui/desktop/src/components/schedule/ScheduleDetailView.tsx
+++ b/ui/desktop/src/components/schedule/ScheduleDetailView.tsx
@@ -21,15 +21,7 @@ import { Loader2, Pause, Play, Edit, Square, Eye } from 'lucide-react';
 import cronstrue from 'cronstrue';
 import { formatToLocalDateWithTimezone } from '../../utils/date';
 import { getSession, Session } from '../../api';
-import {
-  trackScheduleRunNow,
-  trackScheduleUpdated,
-  trackSchedulePaused,
-  trackScheduleUnpaused,
-  trackScheduleKilled,
-  trackScheduleInspected,
-  getErrorType,
-} from '../../utils/analytics';
+import { trackScheduleRunNow, getErrorType } from '../../utils/analytics';
 
 interface ScheduleSessionMeta {
   id: string;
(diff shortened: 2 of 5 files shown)


crates/goose-server/src/routes/telemetry.rs
```diff
@@ -0,0 +1,45 @@
...
+#[derive(Debug, Deserialize, ToSchema)]
+pub struct TelemetryEventRequest {
+    pub event_name: String,
+    #[serde(default)]
+    pub properties: HashMap<String, serde_json::Value>,
+}
+
+#[utoipa::path(
+    post,
+    path = "/telemetry/event",
```
michaelneale, line 19: I think this is ok - as it means we could in theory swap out what is at the backend from posthog

crates/goose/src/posthog.rs
```diff
@@ -218,15 +233,94 @@ pub fn emit_error(error_type: &str) {
...
-        let _ = send_error_event(&installation, &error_type).await;
+        let _ = send_error_event(&installation, &error_type, context).await;
+    });
+}
+
+/// Type of slash command for analytics tracking
+#[derive(Debug, Clone)]
+pub enum SlashCommandType {
+    /// Builtin command (e.g., /compact) - name is safe to track
+    Builtin(String),
```
michaelneale: how come we want to track slash commands? will this be the same from GUI vs CLI? (I didn't realise they were that heavily used - but I guess idea is to track it? not sure why but seems fairly fine grained but I guess there is no info in it)

ui/desktop/src/components/schedule/ScheduleDetailView.tsx
```diff
@@ -21,6 +21,15 @@ import { Loader2, Pause, Play, Edit, Square, Eye } from 'lucide-react';
 import cronstrue from 'cronstrue';
 import { formatToLocalDateWithTimezone } from '../../utils/date';
 import { getSession, Session } from '../../api';
+import {
```
michaelneale: with recipes and schedule tracking - do we want to track it this fine grained vs just seeing if people use either feature in some meaningful way? (maybe need to do this but seems like a lot!)

ui/desktop/src/renderer.tsx
```diff
@@ -4,9 +4,13 @@ import { ConfigProvider } from './components/ConfigContext';
 import { ErrorBoundary } from './components/ErrorBoundary';
 import SuspenseLoader from './suspense-loader';
 import { client } from './api/client.gen';
+import { setTelemetryEnabled } from './utils/analytics';
+import { readConfig } from './api';
 
 const App = lazy(() => import('./App'));
 
+const TELEMETRY_CONFIG_KEY = 'GOOSE_TELEMETRY_ENABLED';
```
michaelneale, line 12: I would have thought the backend would take care of this?

michaelneale:
I think looks ok - is a lot of extra stuff, just had some Qs if we realy want schedules/recipes that fine grained (I guess idea is to know if used or not) but seems ok

crates/goose/src/posthog.rs
zanesq: yeah just to see how much its used, then again maybe we don't need the fine grain just general usage on non built ins would be enough, will follow up with that

ui/desktop/src/renderer.tsx
zanesq: yeah it does also but the frontend needs to check the config also so it can avoid sending all the events for no reason if it is disabled

ui/desktop/src/components/schedule/ScheduleDetailView.tsx
zanesq: hard to say the point is to see what fine grained features are really needed cluttering up our ui... will take another look and see if can be pruned more

crates/goose/src/posthog.rs
zanesq: removed built in tracking logic

ui/desktop/src/components/schedule/ScheduleDetailView.tsx
zanesq: good call removed some low value fine-grained tracking for these scheduling features but I think the recipe ones are valid to get an idea if they are even used

trackScheduleUpdated - editing schedules
trackSchedulePaused / trackScheduleUnpaused - pause toggle
trackScheduleKilled - killing running jobs
trackScheduleInspected - inspecting running jobs
trackScheduleViewed - viewing schedule details

zanesq:
@michaelneale thanks valid concerns, removed a few low signal ones. Will leave this open for a bit longer in case anyone has other feedback.

PR #6124, opened 2025-12-15
feat: add goose PR reviewer workflow

tlongwell-block:
Adds `/goose` support for PRs. Reviews the diff, explores the codebase for context, and posts a structured comment with findings. Skips build validation since CI handles that.

### DIFF
--- a/.github/workflows/goose-pr-reviewer.yml
+++ b/.github/workflows/goose-pr-reviewer.yml
@@ -2,9 +2,7 @@
 #
 # Automated PR review using goose AI agent.
 #
-# Triggers:
-#   - Comment "/goose [optional instructions]" on a PR (OWNER/MEMBER only)
-#   - Manual workflow_dispatch with PR number
+# Trigger: Comment "/goose [optional instructions]" on a PR (OWNER/MEMBER only)
 #
 # Examples:
 #   /goose
@@ -18,19 +16,15 @@
 #   - GOOSE_PROVIDER: LLM provider (default: anthropic)
 #   - GOOSE_MODEL: Model name (default: claude-sonnet-4-5)
 #
-# Security: PR content could prompt-inject the agent; only trigger on PRs you trust.
+# Security:
+#   - PR content could prompt-inject the agent; only trigger on PRs you trust.
+#   - Do not add workflow_dispatch: API calls fetch mutable data, enabling TOCTOU attacks.
 
 name: goose PR Reviewer
 
 on:
   issue_comment:
     types: [created]
-  workflow_dispatch:
-    inputs:
-      pr_number:
-        description: 'PR number to review'
-        required: true
-        type: string
 
 env:
   GOOSE_RECIPE: |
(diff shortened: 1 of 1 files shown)


.github/workflows/goose-pr-reviewer.yml
```diff
@@ -0,0 +1,245 @@
...
+      - name: Install tools
+        run: |
+          apt-get update
+          apt-get install -y jq gettext curl
+          curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg
+          echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+          apt-get update
+          apt-get install -y gh
+
+      - name: Acknowledge trigger
```
jamadeo: consider moving this up to the first step so the user gets quicker feedback

.github/workflows/goose-pr-reviewer.yml
```diff
@@ -0,0 +1,245 @@
...
+          fetch-depth: 1
+
+      - name: Install tools
+        run: |
+          apt-get update
+          apt-get install -y jq gettext curl
+          curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg
+          echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+          apt-get update
+          apt-get install -y gh
```
jamadeo, line 168: I don't think you need to install gh, it should always exist on github runners

.github/workflows/goose-pr-reviewer.yml
```diff
@@ -0,0 +1,245 @@
...
+          gh api --method POST /repos/${{ github.repository }}/issues/comments/${{ github.event.comment.id }}/reactions -f content=eyes
+
+      - name: Get PR details
+        id: pr
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          PR_NUM="${{ github.event.issue.number || github.event.inputs.pr_number }}"
+          
+          gh api /repos/${{ github.repository }}/pulls/$PR_NUM > /tmp/pr.json
```
jamadeo: similar to the issue solver, it should probably get this from the event payload

jamadeo (approved):

.github/workflows/goose-pr-reviewer.yml
tlongwell-block: We're running in the `goose` official Docker container, which doesn't have it baked in

.github/workflows/goose-pr-reviewer.yml
tlongwell-block: It is here to enable it to use the `gh` cli to do so. But this is a simple curl command. Let me change that

.github/workflows/goose-pr-reviewer.yml
tlongwell-block: I removed the workflow dispatch trigger so we could directly pull this from the event info

.github/workflows/goose-pr-reviewer.yml
tlongwell-block: Done

PR #6125, opened 2025-12-15
Don't show subagent tool when we're a subagent

jamadeo:
follow up from https://github.com/block/goose/pull/5825#discussion_r2547053970

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -46,7 +46,7 @@ use crate::recipe::{Author, Recipe, Response, Settings, SubRecipe};
 use crate::scheduler_trait::SchedulerTrait;
 use crate::security::security_inspector::SecurityInspector;
 use crate::session::extension_data::{EnabledExtensionsState, ExtensionState};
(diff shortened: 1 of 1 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -654,6 +652,17 @@ impl Agent {
         {
             return false;
         }
+        if let Some(ref session_id) = self.extension_manager.get_context().await.session_id {
```
jamadeo, line 655: @DOsinga I'm reading the session type here by way of the extension manager + platform extension context, which feels a bit roundabout but is also currently the most direct way.

Do we want to tie an agent to a session more directly? we of course map session id to agent, but the agent doesn't necessarily know its session

alexhancock: in the long run agents should be per-session right? with multiple sessions being possible per app instance and those two concepts coupled.

jamadeo: yeah we map session ID to agent right now, I'm just thinking should we keep a pointer in the other direction too

PR #6127, opened 2025-12-15
Regenerate canonical models when release branch is created.

katzdave:
How this will work:

Runs build_canonical_models -> creates canonical_models.json
Runs canonical_model_checker -> creates canonical_mapping_report.json

canonical_model_checker needs secrets to access our common providers. 

canonical_mapping_report.json isn't that useful, but what canonical_model_checker prints to stdout about which model mappings were changed/added is, so we might want to expose that a little better. 

Can discuss more, but I think on release is a reasonable cadence to update the canonical model mapping + gives us a chance to review if anything catastrophic happened.

### DIFF
--- a/crates/goose/src/providers/canonical/data/canonical_mapping_report.json
+++ b/crates/goose/src/providers/canonical/data/canonical_mapping_report.json
@@ -1,5 +1,5 @@
 {
-  "timestamp": "2025-12-17T19:41:56.317785+00:00",
+  "timestamp": "2025-12-18T16:22:56.520+00:00",
   "unmapped_models": [
     {
       "provider": "openai",
@@ -2367,10 +2367,6 @@
     },
     {
       "provider": "openrouter",
-      "model": "google/gemini-3-flash-preview"
-    },
-    {
-      "provider": "openrouter",
       "model": "google/gemma-3-27b-it:free"
     },
     {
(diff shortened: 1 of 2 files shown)


Justfile
```diff
@@ -301,7 +301,9 @@ prepare-release version:
     # used to update Cargo.lock after we've bumped versions in Cargo.toml
     @cargo update --workspace
     @just set-openapi-version {{ version }}
-    @git add Cargo.toml Cargo.lock ui/desktop/package.json ui/desktop/package-lock.json ui/desktop/openapi.json
+    @cargo run --example build_canonical_models
+    @cargo run --example canonical_model_checker
+    @git add Cargo.toml Cargo.lock ui/desktop/package.json ui/desktop/package-lock.json ui/desktop/openapi.json crates/goose/src/providers/canonical/data/canonical_models.json crates/goose/src/providers/canonical/data/canonical_mapping_report.json
```
jamadeo: probably time to break this up like

```
@git add \
    Cargo.toml \
    Cargo.lock \
    ... etc
```

katzdave: Done

katzdave:
> Makes sense to me to put this here. How likely do you think it is to fail? If it fails, the release PR never gets created, which is not great but if it's pretty reliable that's good enough.
> 
> I don't think these should be "examples" though. You can create regular binary targets with like this:
> 
> https://github.com/block/goose/blob/ae8ecbe0906fd972246be62e09b00ee1679c9b0b/crates/goose-server/Cargo.toml#L57
> 
> and then run with `cargo run --bin build_canonical_models`

Done. Made a new binary and combined these.

Only reasons it would fail are:

1) Network issues openrouter/breaking schema change
2) Missing secret for the checker

Should both generally be fine and I'd want to know if it breaks.

PR #6130, opened 2025-12-15
Integrate pricing with canonical model

katzdave:
Main changes are removing pricing.rs which maintains model pricing via a cached config file which is refreshed from open router. Use canonical models instead (same data source, but much higher hit rate on name matching). 

Stop fetching all of the models/prices into the client, and trying to maintain that state. There was some UI buried in settings about when the last pricing data refresh occurred. Only do targeted fetches with (provider, model).

Result is no meaningful difference in behavior but many more models/providers supported with pricing (databricks!).

A slight loss on the freshness, but will continue to follow up on keeping canonical models as up to date as possible.

### DIFF
--- a/ui/desktop/src/utils/pricing.ts
+++ b/ui/desktop/src/utils/pricing.ts
@@ -7,19 +7,6 @@ export async function fetchModelPricing(
   provider: string,
   model: string
 ): Promise<PricingData | null> {
-  // For local/free providers, return zero cost immediately
-  const freeProviders = ['ollama', 'local', 'localhost'];
-  if (freeProviders.includes(provider.toLowerCase())) {
-    return {
-      provider,
-      model,
-      input_token_cost: 0,
-      output_token_cost: 0,
-      currency: '$',
-      context_length: null,
-    };
-  }
-
   try {
     const response = await getPricing({
       body: { provider, model },


ui/desktop/src/hooks/useCostTracking.ts
```diff
@@ -32,46 +32,53 @@ export const useCostTracking = ({
...
   // Handle model changes and accumulate costs
   useEffect(() => {
-    if (
-      prevModelRef.current !== undefined &&
-      prevProviderRef.current !== undefined &&
-      (prevModelRef.current !== currentModel || prevProviderRef.current !== currentProvider)
-    ) {
-      // Model/provider has changed, save the costs for the previous model
-      const prevKey = `${prevProviderRef.current}/${prevModelRef.current}`;
+    const handleModelChange = async () => {
```
katzdave, line 35: This is client side only and doesn't get persisted... 

Filed: https://github.com/block/goose/issues/6141

Will fix in a followup.

ui/desktop/src/utils/pricing.ts
```diff
@@ -0,0 +1,60 @@
+import { getApiUrl } from '../config';
+
+export interface ModelCostInfo {
```
alexhancock: should we make this a type in the openapi api and get it from there instead? that way we could use the generated SDK for fetching it too.

katzdave: Done

alexhancock: can't we remove this type now and use `PricingData`? https://github.com/block/goose/pull/6130/files#diff-270e779202976f7f9b473ced6ebce8debeba7c7d7e5085991d08d5a363f1f40fR530

that's what I was after

after that, LGTM for a merge!

katzdave: Ah misunderstood slightly. Done

PR #6133, opened 2025-12-16
chore: code mode tool reporting

michaelneale:
This is an attempt to make it show the underlying tool calls in the UI when "code mode" is used: 

image

cc @alexhancock 

don't like how clumsy this feels and special case-ey, but I think we need something functionally like this?


tlongwell-block:
/goose review this, focusing on the ui/ux specifically. Also, have three subagents research and make suggestions for more elegant implementations if possible. It may not be possible. Make a single suggestion for an alternative implementation if you find one that is as complete while being more straightforward and minimal.

michaelneale:
yeah I don't like that it implies that it needs a bespoke UI implementation (it should not - if that review is to believed @tlongwell-block !)

tlongwell-block:
> yeah I don't like that it implies that it needs a bespoke UI implementation (it should not - if that review is to believed @tlongwell-block !)

No, that review was terrible. Please ignore. Sorry for the spam. Working on the goose reviewer and this was too much for it.

PR #6137, opened 2025-12-16
feat: Centralize theme management with ThemeContext

aharvard:
## Summary

This PR centralizes theme management into a single React context, eliminating scattered localStorage reads and brittle event-based synchronization.

> [!NOTE]
> This PR also supports the MCP Apps work by providing a clean, reactive way to get the current theme. See [related discussion](https://github.com/block/goose/pull/6095#discussion_r2621452625).

### Problem

Theme state was previously managed in multiple places with no single source of truth:
- `ThemeSelector.tsx` - read/write localStorage, managed local state
- `App.tsx` - listened for Electron IPC, wrote to localStorage, dispatched synthetic storage events
- `MCPUIResourceRenderer.tsx` - read directly from localStorage (and ignored system theme preference!)
- Various components watching `StorageEvent` which doesn't fire for same-window changes

### Solution

Created `ThemeContext` that provides:
- `userThemePreference`: What the user chose (`'light' | 'dark' | 'system'`)
- `setUserThemePreference`: Function to change preference
- `resolvedTheme`: The actual theme to apply (`'light' | 'dark'`)

### Changes

- **New**: `src/contexts/ThemeContext.tsx` - Single source of truth for theme state
- **Updated**: `ThemeSelector.tsx` - Now uses `useTheme()` hook (simplified from ~150 to ~75 lines)
- **Updated**: `App.tsx` - Wraps with `ThemeProvider`, removed duplicated theme handling
- **Updated**: `MCPUIResourceRenderer.tsx` - Uses `resolvedTheme` (fixes bug where system theme was ignored)

### Benefits

- ✅ Single source of truth via React context
- ✅ Automatic re-renders when theme changes
- ✅ No more localStorage polling or synthetic storage events
- ✅ Type-safe theme values
- ✅ Fixes bug in MCPUIResourceRenderer that ignored system theme preference
- ✅ ~80 fewer lines of code overall

### DIFF
--- a/ui/desktop/src/contexts/ThemeContext.tsx
+++ b/ui/desktop/src/contexts/ThemeContext.tsx
@@ -96,24 +96,10 @@ export function ThemeProvider({ children }: ThemeProviderProps) {
     if (!window.electron) return;
 
     const handleThemeChanged = (_event: unknown, ...args: unknown[]) => {
-      const themeData = args[0];
-
-      // Validate IPC payload structure
-      if (
-        typeof themeData !== 'object' ||
-        themeData === null ||
-        typeof (themeData as Record<string, unknown>).useSystemTheme !== 'boolean' ||
-        typeof (themeData as Record<string, unknown>).theme !== 'string'
-      ) {
-        console.warn('Invalid theme-changed IPC payload:', themeData);
-        return;
-      }
-
-      const { useSystemTheme, theme } = themeData as { useSystemTheme: boolean; theme: string };
-
-      const newPreference: ThemePreference = useSystemTheme
+      const themeData = args[0] as { useSystemTheme: boolean; theme: string };
+      const newPreference: ThemePreference = themeData.useSystemTheme
         ? 'system'
-        : theme === 'dark'
+        : themeData.theme === 'dark'
           ? 'dark'
           : 'light';
 


ui/desktop/src/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+function applyThemeToDocument(theme: ResolvedTheme): void {
+  if (theme === 'dark') {
+    document.documentElement.classList.add('dark');
+    document.documentElement.classList.remove('light');
+  } else {
+    document.documentElement.classList.remove('dark');
+    document.documentElement.classList.add('light');
+  }
+}
```
DOsinga, line 52: we could simplify this to 
const to_remove = theme == 'dark'? 'light': 'dark'

ui/desktop/src/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,161 @@
...
+    };
+
+    window.electron.on('theme-changed', handleThemeChanged);
+    return () => {
+      window.electron.off('theme-changed', handleThemeChanged);
+    };
+  }, []);
+
+  // Apply theme on initial mount
+  useEffect(() => {
```
DOsinga, line 118: remove the comment - this is not true, we apply the theme when it changes. also everywhere we call setResolvedThem(...) and then applyThemToDocument(...) remove the second call and just rely on this effect

ui/desktop/src/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,161 @@
...
+  }, [userThemePreference]);
+
+  // Listen for theme changes from other windows (via Electron IPC)
+  useEffect(() => {
+    if (!window.electron) return;
+
+    const handleThemeChanged = (_event: unknown, ...args: unknown[]) => {
+      const themeData = args[0];
+
+      // Validate IPC payload structure
```
DOsinga: that doesn't seem necesary - we should just type the theme data and use it on both sides:

themeData: { mode: string; useSystemTheme: boolean; theme: string }

DOsinga (approved):
Nice! thanks for factoring this out

zanesq (approved):
Tested works well thanks!

PR #6143, opened 2025-12-16
fix: MCP UI not rendering due to CallToolResult structure change

aharvard:
## Summary

MCP UI was not rendering on main because the tool result structure changed from an array to an object.

## Problem

The `toolResult.value` changed from:
- **Old format**: `Content[]` (array directly)
- **New format**: `{ content: Content[], structuredContent?, isError?, _meta? }` (CallToolResult object)

The existing code expected `toolResult.value` to be an array, so `getToolResultValue` returned `null` and MCP UI resources were never rendered.

This regression was introduced by #6074 which changed the tool result structure to preserve full CallToolResult metadata.

## Solution

- Added typed interfaces for `ToolResultData` matching `rmcp::model::CallToolResult` (camelCase serialization)
- Created `getToolResultContent` helper that:
  - Extracts `content` array from the new structure
  - Filters by audience annotation
- Simplified `toolResults` assignment using the new helper

## Testing

Tested with diagnostic session logs comparing 1.16 (working) vs main (broken) to verify the fix correctly extracts content from the new structure.

### DIFF
--- a/ui/desktop/src/components/ToolCallWithResponse.tsx
+++ b/ui/desktop/src/components/ToolCallWithResponse.tsx
@@ -15,7 +15,7 @@ import { ChevronRight, FlaskConical } from 'lucide-react';
 import { TooltipWrapper } from './settings/providers/subcomponents/buttons/TooltipWrapper';
 import MCPUIResourceRenderer from './MCPUIResourceRenderer';
 import { isUIResource } from '@mcp-ui/client';
-import { Content, EmbeddedResource } from '../api';
+import { CallToolResponse, Content, EmbeddedResource } from '../api';
 
 interface ToolCallWithResponseProps {
   isCancelledMessage: boolean;
@@ -30,7 +30,7 @@ function getToolResultContent(toolResult: Record<string, unknown>): Content[] {
   if (toolResult.status !== 'success') {
     return [];
   }
-  const value = toolResult.value as { content: Content[] };
+  const value = toolResult.value as CallToolResponse;
   return value.content.filter((item) => {
     const annotations = (item as { annotations?: { audience?: string[] } }).annotations;
     return !annotations?.audience || annotations.audience.includes('user');


aharvard:
This regression was introduced by #6074 which changed the tool result structure to preserve full CallToolResult metadata.

ui/desktop/src/components/ToolCallWithResponse.tsx
```diff
@@ -26,11 +26,43 @@ interface ToolCallWithResponseProps {
...
-    return toolResult.value as Content[];
+/**
+ * Represents the inner value of a successful tool result.
+ * Maps to rmcp::model::CallToolResult which uses camelCase serialization.
+ */
+interface CallToolResultValue {
+  content: Content[];
+  structuredContent?: unknown;
+  isError?: boolean;
+  _meta?: unknown;
```
DOsinga: you should use the version we get from the api, not declare your own. I'm confused though since the API version doesn't seem to use camelCase?

ui/desktop/src/components/ToolCallWithResponse.tsx
```diff
@@ -26,11 +26,43 @@ interface ToolCallWithResponseProps {
...
+ */
+interface ToolResultSuccess {
+  status: 'success';
+  value: CallToolResultValue;
+}
+
+interface ToolResultError {
+  status: 'error';
+  error: string;
+}
```
DOsinga: I wouldn't bother with these since your doing a bunch of casting anyway

ui/desktop/src/components/ToolCallWithResponse.tsx
```diff
@@ -26,11 +26,43 @@ interface ToolCallWithResponseProps {
...
+
+type ToolResultData = ToolResultSuccess | ToolResultError;
+
+function getToolResultContent(toolResult: Record<string, unknown>): Content[] {
```
DOsinga, line 29: I think if you drop all the types, you can just go with

```
function getToolResultContent(toolResult: Record<string, unknown>): Content[] {
  if (toolResult.status !== 'success') {
    return [];
  }
  const value = toolResult.value as CallToolResponse;
  return value.content.filter((item) => {
    const annotations = (item as { annotations?: { audience?: string[] } }).annotations;
    return !annotations?.audience || annotations.audience.includes('user');
  });
}
```

ui/desktop/src/components/ToolCallWithResponse.tsx
```diff
@@ -253,12 +285,9 @@ function ToolCallView({
     }
   }, [toolResponse, startTime]);
 
-  const toolResults: Content[] =
-    loadingStatus === 'success' && Array.isArray(toolResponse?.toolResult.value)
-      ? toolResponse!.toolResult.value.filter((item) => {
-          const audience = item.annotations?.audience as string[] | undefined;
-          return !audience || audience.includes('user');
```
DOsinga, line 260: good clean-up here

aharvard:
feedback addressed, thanks!

ui/desktop/src/components/ToolCallWithResponse.tsx
```diff
@@ -26,11 +26,15 @@ interface ToolCallWithResponseProps {
...
+  if (toolResult.status !== 'success') {
+    return [];
   }
-  return null;
+  const value = toolResult.value as { content: Content[] };
+  return value.content.filter((item) => {
+    const annotations = (item as { annotations?: { audience?: string[] } }).annotations;
+    return !annotations?.audience || annotations.audience.includes('user');
+  });
 }
```
DOsinga, line 38: this is fine, but casting the value to CallToolResponse from the api would be even better and avoid this sort of problem in the future

aharvard: happy to address this, no prob!

Issue #6147, opened 2025-12-17
Bug: read_module tool fails with "Missing required parameter: path"


rabi:
**Bug Description**

When using the read_module tool to discover tool signatures, the tool fails with the error: Missing required parameter: path

This occurs even when the LLM correctly provides the parameter according to the tool's JSON schema.

**Root Cause**

There was a mismatch between the parameter name defined in the schema and the parameter name expected by the handler:

- Schema (ReadModuleParams struct): defines the parameter as module_path
- Handler (handle_read_module function): looks for a parameter named path

When the LLM reads the tool schema, it correctly uses module_path, but the handler rejects it because it's looking for path.

**To Reproduce**
 
- Start a Goose session with the code_execution extension enabled (this is typically enabled by default)
- Ask Goose to use an unfamiliar tool from another MCP extension
- Observe the error:  Error: Missing required parameter: path

**Expected behavior**
The read_module tool should return the tool signature and description for jira/search_issues, allowing Goose to understand the required parameters before calling the tool.

**Actual Behavior**
The tool fails with "Missing required parameter: path" even though the module_path parameter was correctly provided according to the tool's JSON schema.

PR #6148, opened 2025-12-17
fix: use correct parameter name in read_module handler

rabi:
## Summary

Handler is looking for 'path' but the ReadModuleParams schema defines 'module_path'. This caused the tool to fail with 'Missing required parameter: path' when the LLM correctly used 'module_path'.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Unit and manual testing with main branch build

### Related Issues
Relates to:  https://github.com/block/goose/issues/6147

### DIFF
--- a/crates/goose/src/agents/code_execution_extension.rs
+++ b/crates/goose/src/agents/code_execution_extension.rs
@@ -365,9 +365,9 @@ impl CodeExecutionClient {
     ) -> Result<Vec<Content>, String> {
         let path = arguments
             .as_ref()
-            .and_then(|a| a.get("path"))
+            .and_then(|a| a.get("module_path"))
             .and_then(|v| v.as_str())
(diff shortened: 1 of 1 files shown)


rabi:
> Nice. thanks. we could consider turning it into a string constant so we are always sure we use the same thing in the three places where it is used

We could, but I've left it as inline string literal for consistency with existing codebase patterns. Please see similar usage in `ExecuteCodeParams.code`, `SearchModulesParams.terms,` and throughout `providers/utils.rs`, `providers/formats/google.rs`. If we want to adopt constants, that could be a separate effort across the codebase probably(?).

DOsinga:
fair enough. we should though

PR #6149, opened 2025-12-17
fix: we don't need to warn about tool count when in code mode

michaelneale:
when code mode is enabled, we don't enumerate all the tools, so we don't need to warn about size, or add this extra noise to the system prompt as it isn't relevant.


codefromthecrypt (approved):

DOsinga:
I would be ok with removing the entire toolcounting actually. I think it made more sense when models where very bad at multiple tools; right now it really depends on the model so we should either do this dynamically which seems hard or just delete the checks and not confuse the user.

alexhancock:
good catch, thank you!

PR #6151, opened 2025-12-17
add bedrock prompt caching

fbalicchia:
## Summary
Implemented prompt caching for Anthropic Claude models on AWS Bedrock to reduce costs 
Introduced an intelligent cache point placement strategy that complies with AWS Bedrock’s four cache point limitation.
Added the BEDROCK_ENABLE_CACHING configuration parameter

### Testing
  - Verify caching is enabled for Claude models by default
  - Test that cache points are correctly placed in system prompts
  - Test that cache points are correctly placed in user/assistant messages
  - Verify cache point limit (4 max) is respected

### DIFF
--- a/crates/goose/src/security/scanner.rs
+++ b/crates/goose/src/security/scanner.rs
@@ -46,7 +46,7 @@ impl PromptInjectionScanner {
     fn create_classifier_from_config() -> Result<ClassificationClient> {
         let config = Config::global();
 
-        let model_name = config
+        let mut model_name = config
             .get_param::<String>("SECURITY_PROMPT_CLASSIFIER_MODEL")
             .ok()
             .filter(|s| !s.trim().is_empty());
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/bedrock.rs
```diff
@@ -128,6 +129,17 @@ impl BedrockProvider {
         }
     }
 
+    fn should_enable_caching(&self, model_name: &str) -> bool {
+        let config = crate::config::Config::global();
+        if let Ok(enabled) = config.get_param::<bool>("BEDROCK_ENABLE_CACHING") {
```
DOsinga: shouldn't we do this by default? i.e. change it to DISSABLE?

crates/goose/src/providers/formats/bedrock.rs
```diff
@@ -18,15 +18,33 @@ use super::super::base::Usage;
...
+) -> Result<bedrock::Message> {
+    let mut content_blocks: Vec<bedrock::ContentBlock> = message
+        .content
+        .iter()
+        .map(to_bedrock_message_content)
+        .collect::<Result<_>>()?;
+
+    // Add cache point after the last content block if caching is enabled
+    // This allows AWS Bedrock to cache the entire message content
+    if enable_caching && !content_blocks.is_empty() {
```
DOsinga, line 34: I don't think you need the comment here

fbalicchia:
Hi @DOsinga , I’ve pushed some updates to this PR. When you have a moment, could you please take another look? Thanks!

fbalicchia:
After making a bit of a mess with the branch I close this pr in favor of [6463](https://github.com/block/goose/pull/6463)

Issue #6140, opened 2025-12-16
Desktop: Opening a session to view it modifies session history order


dianed-square:
**Describe the bug**

Opening a session from the History tab to view it (read only, no messages sent) modifies the session's "last used" timestamp, causing it to jump to the top of the history list.

---

**To Reproduce**
1. Go to the History tab
2. Note the order of sessions
3. Click on an older session to open it
4. Go back to the History tab
5. See that the session you just viewed has moved to the top of the list

---

**Expected behavior**
Opening a session to view it should not modify the session or change its position in the history order.

---

**Screenshots**
n/a

---

**Please provide the following information**
- **OS & Arch:** macOS
- **Interface:** UI (Goose Desktop)
- **Version:** v1.16.0
- **Extensions enabled:** N/A - not extension-related
- **Provider & Model:** N/A - not provider-related

PR #6156, opened 2025-12-17
fix : opening a session to view it modifies session history order in desktop

Abhijay007:
Closes: #6140

## PR Description

This PR fixes an issue where opening a session from the History tab in read-only mode updated the session’s `updated_at` timestamp, causing it to move to the top of the history list and break chronological ordering.
The root cause was that `resume_agent()` was calling `update_provider()`, which persisted provider data to the database and updated the `updated_at` timestamp even when the session was only being viewed.

### Changes Made

To fix this, a new `load_provider()` method was introduced to restore the provider state in memory only, without writing to the database or updating timestamps. `resume_agent()` now uses this method when loading sessions from history.

- Added `load_provider()` to restore provider state without persistence  
- Updated `resume_agent()` to use `load_provider()` instead of `update_provider()`

## Type of Change
- [x] Bug fix

## AI Assistance
- [x] This PR was created or reviewed with AI assistance - used Goose to test different approaches for the solution

## Testing

Tested in the Desktop UI by opening sessions from the History tab and confirming the session order remains unchanged.

## Screenshots / Demos

**Before:**
beforeHistory

**After:** 

afterHistory

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -1326,6 +1326,14 @@ impl Agent {
             .context("Failed to persist provider config to session")
     }
 
+    pub async fn load_provider(&self, provider: Arc<dyn Provider>) -> Result<()> {
+        let mut current_provider = self.provider.lock().await;
+        *current_provider = Some(provider.clone());
+
+        self.update_router_tool_selector(Some(provider.clone()), None)
+            .await
+    }
+
     pub async fn update_router_tool_selector(
         &self,
         provider: Option<Arc<dyn Provider>>,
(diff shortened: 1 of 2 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -1375,6 +1326,14 @@ impl Agent {
             .context("Failed to persist provider config to session")
     }
 
+    pub async fn load_provider(&self, provider: Arc<dyn Provider>) -> Result<()> {
+        let mut current_provider = self.provider.lock().await;
+        *current_provider = Some(provider.clone());
+
+        self.update_router_tool_selector(Some(provider.clone()), None)
```
michaelneale, line 1333: odd that this is needed?

PR #6157, opened 2025-12-17
fix: support global .gooseignore and negation patterns

### LINKED ISSUE #6155: Global .gooseignore negation pattern not working
## Description

When trying to access a `.env.example` file in a project, Goose blocks access with the error:

```
Access to '/path/to/project/.env.example' is restricted by .gooseignore
```

## Environment

- **Local .gooseignore**: Does not exist in the project directory
- **Global .gooseignore** at `~/.config/goose/.gooseignore` contains:
  ```
  !.env.example
  ```

## Expected Behavior

The `!` prefix should negate the pattern and explicitly **allow** access to `.env.example` files, following the standard gitignore-style negation syntax.

## Actual Behavior

Goose still blocks access to `.env.example` files despite the negation pattern in the global `.gooseignore`.

## Possible Causes

1. The negation pattern (`!`) isn't being parsed correctly
2. There's a default internal rule blocking `.env*` files that the negation isn't overriding
3. The global `.gooseignore` file isn't being read/applied properly

## Steps to Reproduce

1. Create a global `.gooseignore` at `~/.config/goose/.gooseignore` with content `!.env.example`
2. Ensure no local `.gooseignore` exists in the project
3. Try to view or edit a `.env.example` file using Goose
4. Observe the access restriction error

leocavalcante:
## Summary

This PR fixes issue #6155 where global `.gooseignore` negation patterns were not working to allow access to files like `.env.example`.

## Problem

The previous implementation had two issues:
1. The global `.gooseignore` file at `~/.config/goose/.gooseignore` was never being loaded
2. Default patterns were only applied when no local `.gooseignore` existed, making it impossible to negate them

## Solution

Changed `build_ignore_patterns()` to:
1. **Always apply default patterns first** (`**/.env`, `**/.env.*`, `**/secrets.*`) - lowest priority
2. **Load global `.gooseignore`** from `~/.config/goose/.gooseignore` - medium priority
3. **Load local `.gooseignore`** - highest priority

This follows the standard gitignore pattern matching semantics where later patterns can override earlier ones, and negation patterns (prefixed with `!`) can allow access to files that would otherwise be ignored.

## Example

With this fix, users can now add `!.env.example` to their global or local `.gooseignore` to allow access to `.env.example` files while still blocking `.env` and `.env.local`.

## Testing

Added 3 new tests:
- `test_global_gooseignore_negation_pattern` - verifies global negation patterns work
- `test_local_gooseignore_overrides_global` - verifies local patterns override global
- `test_local_gooseignore_negation_overrides_defaults` - verifies local negation patterns override defaults

All existing ignore-related tests continue to pass.

Fixes #6155


crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -1284,17 +1285,50 @@ impl DeveloperServer {
...
+        // Check GOOSE_PATH_ROOT first (used for testing), then fall back to etcetera
+        let global_ignore_path = if let Ok(test_root) = std::env::var("GOOSE_PATH_ROOT") {
+            // For testing: GOOSE_PATH_ROOT/config/.gooseignore
+            Some(
+                std::path::PathBuf::from(test_root)
+                    .join("config")
+                    .join(".gooseignore"),
+            )
+        } else {
+            etcetera::choose_app_strategy(etcetera::AppStrategyArgs {
```
katzdave: What's going on here? Don't want anything block specific. Can we mirror this:
https://github.com/block/goose/blob/bf95995e1d9f75385f3bebe16e97e6459873af57/crates/goose-mcp/src/memory/mod.rs#L185

crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -1284,17 +1285,50 @@ impl DeveloperServer {
...
+
+        let has_local_ignore = local_ignore_path.is_file();
+        let has_global_ignore = global_ignore_path
+            .as_ref()
+            .map(|p| p.is_file())
+            .unwrap_or(false);
+
+        // 1. Always add default patterns first (lowest priority)
+        //    This allows them to be negated by global or local .gooseignore files
+        let _ = builder.add_line(None, "**/.env");
```
katzdave: Old code only used these when we didn't explicitly have an ignore file; I think that makes sense.

crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -3237,6 +3271,110 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    #[serial]
+    async fn test_global_gooseignore_negation_pattern() {
```
katzdave: These tests are a bit longwinded.

Can we have a single parameterized test with inputs of:

global .gooseignore contents.
local .gooseignore contents.
expected output

crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -1284,17 +1285,40 @@ impl DeveloperServer {
...
+        let _ = builder.add_line(None, "**/.env");
+        let _ = builder.add_line(None, "**/.env.*");
+        let _ = builder.add_line(None, "**/secrets.*");
+
+        // 2. Add global .gooseignore patterns (medium priority)
+        //    These can override defaults but are overridden by local patterns
+        if let Some(ref global_path) = global_ignore_path {
+            if has_global_ignore {
+                let _ = builder.add(global_path);
+            }
```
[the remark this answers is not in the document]
katzdave: +1

leocavalcante:
@katzdave I've addressed all the PR review comments:

- Using `choose_app_strategy(APP_STRATEGY)` pattern from memory/mod.rs
- Simplified the redundant nested check for has_global_ignore  
- Converted the three verbose tests to a single parameterized test
- Using HOME env var for test isolation instead of GOOSE_PATH_ROOT

Ready for another review when you get a chance!

crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -1284,17 +1285,29 @@ impl DeveloperServer {
     fn build_ignore_patterns(cwd: &PathBuf) -> Gitignore {
         let mut builder = GitignoreBuilder::new(cwd);
         let local_ignore_path = cwd.join(".gooseignore");
-        let mut has_ignore_file = false;
 
-        if local_ignore_path.is_file() {
-            let _ = builder.add(local_ignore_path);
-            has_ignore_file = true;
+        // Get the global config directory for .gooseignore
+        // Uses the same pattern as memory/mod.rs for consistency
```
katzdave: rm comments. And below

crates/goose-mcp/src/developer/rmcp_developer.rs
katzdave: Let's only do these defaults if we don't have a local/global ignore file.

crates/goose-mcp/src/developer/rmcp_developer.rs
```diff
@@ -3237,6 +3250,85 @@ mod tests {
         );
     }
 
+    /// Test case for gooseignore patterns
```
katzdave: Sorry, this still ended up being pretty complicated; I was hoping we could have something that just tested some of the diff cases without actually interacting with the filesystem. 

If that's not possible, let's just get rid of the test completely because the logic in build_ignore_patterns isn't too heavy.

PR #6160, opened 2025-12-17
refactor: subagents as platform extension (enables subagents in code mode)

tlongwell-block:
Moves the subagent tool from being a special case in `Agent::dispatch_tool_call` to a proper platform extension, following the same pattern as todo, skills, chatrecall, and code_execution extensions.

### Motivation

1. **Consistency**: The subagent tool had special-case handling in the agent's dispatch logic
2. **Code mode support**: Platform extensions are accessible from `code_execution` - this change enables subagent delegation from JavaScript code

### Changes

- New `SubagentClient` implementing `McpClientTrait` 
- New `call_tool_deferred` trait method allowing extensions to return deferred futures (with sensible default for sync tools)
- `PlatformExtensionContext` extended with `working_dir` and `sub_recipes`
- `TaskConfig` simplified (session/working_dir now obtained from context)
- `sub_recipes` changed from `Mutex` to `Arc<RwLock>` for shared read access
- Change from `"path"` to `"module_path"` in `handle_read_module` fixes a pre-existing bug where the code looked for `"path"` but the schema defined `module_path`.

### DIFF
--- a/scripts/test_subrecipes.sh
+++ b/scripts/test_subrecipes.sh
@@ -77,13 +77,13 @@ check_recipe_output() {
   local tmpfile=$1
   local mode=$2
   
-  # Check for unified subagent tool invocation (new format: "─── subagent |")
-  if grep -q "─── subagent" "$tmpfile"; then
-    echo "✓ SUCCESS: Subagent tool invoked"
-    RESULTS+=("✓ Subagent tool invocation ($mode)")
+  # Check for delegate tool invocation (format: "─── delegate | subagent")
+  if grep -q "─── delegate" "$tmpfile"; then
+    echo "✓ SUCCESS: Delegate tool invoked"
+    RESULTS+=("✓ Delegate tool invocation ($mode)")
   else
-    echo "✗ FAILED: No evidence of subagent tool invocation"
-    RESULTS+=("✗ Subagent tool invocation ($mode)")
+    echo "✗ FAILED: No evidence of delegate tool invocation"
+    RESULTS+=("✗ Delegate tool invocation ($mode)")
   fi
   
   # Check that both subrecipes were called (shown as "subrecipe: <name>" in output)


tlongwell-block:
/goose trying to make `subagent` fit in as a platform extension rather than a special case. Also need it to work with code_execution. Is this an ideal move from the custom agent-registered subagent tool to a more standard platform extension? Any concerns? Is it at least as cleanly implemented/architected as the other platform extensions?

crates/goose/src/agents/subagent_client.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    async fn build_tool(&self) -> Tool {
+        let sub_recipes = self.get_sub_recipes().await;
+        let sub_recipes_vec: Vec<_> = sub_recipes.values().cloned().collect();
+        create_subagent_tool(&sub_recipes_vec)
+    }
+}
+
+#[async_trait]
+impl McpClientTrait for SubagentClient {
```
jamadeo, line 91: just a thought (not feedback on this PR)

We could probably benefit from an extension construct that isn't necessarily an MCP extension. We're making good use of the platform extension construct, but having them "pretend" to be MCP servers results in us needing some odd code like this.

crates/goose/src/agents/mcp_client.rs
```diff
@@ -64,6 +65,22 @@ pub trait McpClientTrait: Send + Sync {
         cancel_token: CancellationToken,
     ) -> Result<CallToolResult, Error>;
 
+    async fn call_tool_deferred(
+        &self,
+        name: &str,
+        arguments: Option<JsonObject>,
+        cancel_token: CancellationToken,
+    ) -> Result<ToolCallResult, Error> {
```
jamadeo: oof. call_tool returns a CallToolResult and call_tool_deferred returns a ToolCallResult

crates/goose/src/agents/extension_manager.rs
```diff
@@ -1080,26 +1085,24 @@ impl ExtensionManager {
...
-                    ServiceError::McpError(error_data) => error_data,
-                    _ => {
-                        ErrorData::new(ErrorCode::INTERNAL_ERROR, e.to_string(), e.maybe_to_value())
-                    }
-                })
-        };
+        let mut result = client
+            .lock()
+            .await
+            .call_tool_deferred(&tool_name, arguments, cancellation_token)
```
jamadeo: instead of a new function, how about changing this so that McpClientTrait::call_tool always returns a notification stream, with the default impl calling .subscribe()?

I'm also seeing that all of the platform extensions just return a dummy channel for subscribe, but ToolCallResult has an Option for the notifcations stream, so it would make more sense to have the trait fn itself return an Option

crates/goose/src/agents/agent.rs
```diff
@@ -430,20 +435,8 @@ impl Agent {
         tool_call: CallToolRequestParam,
         request_id: String,
         cancellation_token: Option<CancellationToken>,
-        session: &Session,
+        _session: &Session,
```
jamadeo: if possible, just remove the parameter

crates/goose/src/agents/subagent_client.rs
```diff
@@ -0,0 +1,207 @@
...
+    info: InitializeResult,
+}
+
+impl SubagentClient {
+    pub fn new(context: PlatformExtensionContext) -> Result<Self> {
+        Ok(Self {
+            context,
+            info: InitializeResult {
+                protocol_version: ProtocolVersion::V_2025_03_26,
+                capabilities: ServerCapabilities {
```
jamadeo: ServerCapabilities has a builder 

```
ServerCapabilities::builder().enable_tools().build()
```

crates/goose/src/agents/agent.rs
```diff
@@ -267,6 +264,10 @@ impl Agent {
         let initial_messages = conversation.messages().clone();
         let config = Config::global();
 
+        if let Err(e) = self.enable_subagent_extension().await {
```
jamadeo: this seems like an odd place to enable the extension

it looks like this will enable it even if the user has toggled it off -- also I don't think this will disable if something else changes, like tool approve mode or model

crates/goose/src/agents/extension_manager.rs
tlongwell-block: I think this is a good idea, but I don't necessarily want to do it in this PR.

crates/goose/src/agents/mcp_client.rs
tlongwell-block: Changed to DeferredToolCall

tlongwell-block:
I think I need to make goose reviewer even more critical

crates/goose-cli/src/session/builder.rs
```diff
@@ -496,6 +497,8 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> CliSession {
         }
     }
 
+    agent_ptr.ensure_subagent_for_recipes().await;
```
jamadeo: this makes me think that sub recipes should be part of the subagent platform extension itself, not the agent

at the least, I would say do this within the agent such that it isn't the caller's responsibility to always call this. But then you still have this situation where the Agent has sub_recipes, but they're non functional unless it's been configured with this extension

tlongwell-block: I agree. But I want to do that in a separate PR, potentially one for this overhaul https://github.com/block/goose/discussions/6202

tlongwell-block:
> This mostly looks good to me, and definitely makes sense to have subagents as a platform extension!
> 
> Trying to understand the need for the new `call_tool_deferred` and the switch from `ToolCallResult -> DeferredToolCall`. Can you provide some background on it?

I needed that so that we could actually run parallel subagents when goose makes multiple subagent tool calls in one turn

| "parallelism is hard" - commit a6a2886

I ended up wrapping the async call in the default blocking path to make sure subagent could be used through that more standard way, too

tlongwell-block:
https://github.com/block/goose/pull/6964 is a better version of this

PR #6164, opened 2025-12-18
fix: when changing modes mid chat or in settings need to notify

michaelneale:
When changing chat modes - you need to currently restart sessions. This will tell you via a simple toast to do that if you change when in active chat or in settings.

### DIFF
--- a/ui/desktop/src/components/ChatInput.tsx
+++ b/ui/desktop/src/components/ChatInput.tsx
@@ -1547,7 +1547,7 @@ export default function ChatInput({
             </div>
           </Tooltip>
           <div className="w-px h-4 bg-border-default mx-2" />
-          <BottomMenuModeSelection />
+          <BottomMenuModeSelection hasActiveSession={messages.length > 0} />
           {sessionId && process.env.ALPHA && (
             <>
               <div className="w-px h-4 bg-border-default mx-2" />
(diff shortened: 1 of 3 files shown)


zanesq:
Thanks! I looked into it and we don't need to change mid chat it applies on the next message reply currently so we only need to do this on the settings page.

DOsinga:
FWIW I have seen on model change that it doesn't show this in the UI until after reload

michaelneale:
@DOsinga oh so we should add the warning for model change as well you mean?

(this is partly only temporary until multi chat lands)

zanesq:
model change seems to work fine mid chat for me, example switching to gemini (model response fails for other reasons but you can see it working)

https://github.com/user-attachments/assets/3fd4fac6-5730-4e60-83c7-f97cf9e14b56

michaelneale:
so may be ok as is @zanesq ?

zanesq:
I think model switching mid chat is ok so no need for a notice there.
Mode change mid chat also works mid chat currently so no need for it there also. Can you remove it from mid chat and only keep it in settings?

michaelneale:
hrm - when I tried it, it didn't seem to actually change it mid chat - you sure @zanesq ? (I just put it in chat only - and it still called tools)

zanesq:
I haven't tried it just looking at the code I see it applies the mode on the next `reply`.. will try it first thing tomorrow to verify

zanesq:
@michaelneale strange it works for me on `main` see below. Also for some reason if I test on this branch it never gets past the first prompt "create a test file in my desktop" so there might be a bug in this branch?

https://github.com/user-attachments/assets/d34d90b0-5c83-4f2e-8910-b569e4e1854c

michaelneale:
hrm yeah - not sure what has gone wrong but will take another run at it.

PR #6166, opened 2025-12-18
chore: manually update version

alexhancock:
We need to do this this time because the PR for 1.17.0 couldn't merge cleanly

https://github.com/block/goose/pull/6131

We can update the release workflow to only actually merge a version bump like this in the future to avoid conflicts, but we should do this manually for this one


katzdave:
So what happened? Did you trigger the release pipeline manually; It usually gets triggered from merging that branch now right?

PR #6177, opened 2025-12-18
fix(code-mode): improve tool signatures for LLM discovery

### LINKED ISSUE #6147: Bug: read_module tool fails with "Missing required parameter: path"

codefromthecrypt:
## Summary

Allows the following prompt to work occasionally with code mode in dumber LLM like gpt-5-nano
> Search for web_scrape and text_editor tools. Use them to save https://httpbingo.org/get to /tmp/result.txt.

- Parse enum types from JSON Schema (oneOf+const, enum arrays)
- Dereference $ref using unbinder for schemars-generated schemas
- Extract return types from output_schema (default: string)
- Remove JS-like syntax from descriptions that confused LLMs

Fixes web_scrape showing save_as as "any" instead of "text" | "json" | "binary"

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing

Unified test_mcp_tool_signature with real MCP server data
- Covers github, filesystem, memory schema patterns

Added a test and used the mentioned prompt successfully
```bash
$  cargo run -p goose-cli --release -- session --with-builtin code_execution,developer,computercontroller                                                                                                             
   Compiling goose v1.17.0 (/Users/codefromthecrypt/oss/goose/crates/goose)
   Compiling goose-bench v1.17.0 (/Users/codefromthecrypt/oss/goose/crates/goose-bench)
   Compiling goose-cli v1.17.0 (/Users/codefromthecrypt/oss/goose/crates/goose-cli)
    Finished `release` profile [optimized] target(s) in 47.95s
     Running `target/release/goose session --with-builtin code_execution,developer,computercontroller`
starting session | provider: tetrate model: gpt-5-nano
    session id: 20251219_3
    working directory: /Users/codefromthecrypt/oss/goose

Context: ○○○○○○○○○○ 0% (0/272000 tokens)
( O)> Search for web_scrape and text_editor tools. Use them to save https://httpbingo.org/get to /tmp/result.txt.

terms: ["web_scrape", "text_editor"]


module_path: computercontroller


module_path: developer


module_path: extensionmanager


module_path: github


─── execute_code | code_execution ──────────────────────────
code: import { web_scrape } from "computercontroller";
import { text_editor } from "developer";                                                                                                                                                                                  
                                                                                                                                                                                                                          
const content = web_scrape({ save_as: "text", url: "https://httpbingo.org/get" });                                                                                                                                        
text_editor({ path: "/tmp/result.tx...                                                                                                                                                                                    

## Result

- Action completed: Saved the contents of https://httpbingo.org/get to /tmp/result.txt in a single batched operation.
- Internal note: The content was also cached locally at the tool’s temp location:
  - /Users/codefromthecrypt/.cache/goose/computer_controller/web_20251219_104714.txt

## Verification commands (choose one)

- View the full file:
  - cat /tmp/result.txt
- Preview the first 200 lines:
  - head -n 200 /tmp/result.txt
- If you expect JSON and want a quick check:
  - head -n 50 /tmp/result.txt

## What’s in the file

- The page at httpbingo.org/get typically returns a simple JSON response with details about the request (args, headers, URL). If you’d like, I can parse and display specific fields (e.g., URL, headers) or convert it to another format (JSON, pretty-printed, etc.).                                                                                                                                                                             

## Next steps

- Save to a different path or filename?
- Fetch another URL and save similarly?
- Parse and extract specific fields from the saved content? If yes, tell me which fields you want.

⏱️  Elapsed time: 54.07s
Context: ○○○○○○○○○○ 2% (6168/272000 tokens)
```

### Related Issues

Fixes #6147

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -174,11 +174,14 @@ fn require_str_parameter<'a>(v: &'a serde_json::Value, name: &str) -> Result<&'a
 }
 
 pub fn get_parameter_names(tool: &Tool) -> Vec<String> {
-    tool.input_schema
+    let mut names: Vec<String> = tool
+        .input_schema
         .get("properties")
         .and_then(|props| props.as_object())
         .map(|props| props.keys().cloned().collect())
-        .unwrap_or_default()
+        .unwrap_or_default();
+    names.sort();
+    names
 }
 
 impl Default for ExtensionManager {


crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -82,19 +82,31 @@ impl ToolInfo {
...
             .iter()
             .map(|name| {
                 let ty = schema
                     .properties
                     .get(name)
-                    .and_then(|p| p.get("type"))
-                    .and_then(|t| t.as_str())
-                    .unwrap_or("any");
+                    .and_then(|p| {
+                        // Check for oneOf with const (schemars generates this for enums)
```
codefromthecrypt: if this comment is too much, happy to scrub it. This was the key issue in WebScrape not working, and it was getting gnarly to handle these by hand. Also, there are other cases besides oneOf we probably need to handle soon. Just didn't want to go too much in one PR

codefromthecrypt:
race condition.. I'll fix this back to module_path everywhere

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -39,10 +39,10 @@ struct ExecuteCodeParams {
 
 #[derive(Debug, Serialize, Deserialize, JsonSchema)]
 struct ReadModuleParams {
-    /// Module path format:
+    /// Path format:
     /// - For entire server: "server_name"
     /// - For specific tool: "server_name/tool_name"
-    module_path: String,
+    path: String,
```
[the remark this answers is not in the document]
michaelneale: is this intentional @codefromthecrypt  - I think @tlongwell-block mentioned something like this too

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -656,9 +669,9 @@ impl McpClientTrait for CodeExecutionClient {
                         Search for tools by name or description across all available modules.
 
                         USAGE:
-                        - Single term: search_modules({ terms: "file" })
-                        - Multiple terms: search_modules({ terms: ["git", "shell"] })
-                        - Regex patterns: search_modules({ terms: "sh.*", regex: true })
+                        - Single term: terms="file"
```
michaelneale: it is ok to not say "search_modules" now? still works?

crates/goose/src/agents/code_execution_extension.rs
michaelneale: is ok - but could be expanded to say that it can happen (I didn't know!) - anything else we are missing?

crates/goose/src/agents/code_execution_extension.rs
codefromthecrypt: no you are right. it is just doing it without it looking like javascript. what was happening was the llm was executing this directly. that said, my change went too far here search_modules should be in!

codefromthecrypt:
@michaelneale on
> is ok - but could be expanded to say that it can happen (I didn't know!) - anything else we are missing?

if we are keen on using the schema dep, I can add more cases to the handling besides oneOf so no bombs like other things coming up as 'any' making the llm guess what the params might be.

michaelneale:
@codefromthecrypt sure - also that Live test failure - is that legit that it is tripping up qwen? (or can't be, as this is only code mode now?)

crates/goose/src/agents/code_execution_extension.rs
codefromthecrypt: ok I expanded with all the real data reasonable to use.. smallest signatures that trigger things from all mcp servers we have blogs about (only chose one per pattern)

crates/goose/src/agents/code_execution_extension.rs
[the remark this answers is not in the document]
codefromthecrypt: scrubbed this out

crates/goose/src/agents/code_execution_extension.rs
codefromthecrypt: should be done now

codefromthecrypt:
will take out of draft when done

codefromthecrypt:
@michaelneale I think we're good. I looked at the live tests and think that qwen fail earlier was just a flake. I've pushed several times to re-trigger and it hasn't failed again.

codefromthecrypt:
@michaelneale gonna mergeroo this one. The live provider tests are flakey this hasn't anything to do with it. In the last week ~1 in 5 times across any branch flaked.

PR #6183, opened 2025-12-19
chore: cover code mode with end to end provider tests

michaelneale:
Attempting to cover the new code mode with same tests for providers. Should work the same as stock developer extension


codefromthecrypt (approved):
ah so the main change here is we are still doing "list the files...", just that we are adding code execution to do it which forces the other extensions inside. Same test

alexhancock:
Nice work @michaelneale !

PR #6187, opened 2025-12-19
Fix GCP Vertex AI global endpoint support for Gemini 3 models

clouatre:
## Summary
Fixes the GCP Vertex AI provider to support the global endpoint required for Gemini 3 models. The previous implementation incorrectly constructed the URL as `https://global-aiplatform.googleapis.com` instead of `https://aiplatform.googleapis.com`.

### Type of Change
- [x] Bug fix

### AI Assistance
- [x] This PR was created or reviewed with AI assistance

### Testing
**Unit Tests:**
- All 11 tests passing (4 new tests added, 7 existing tests still passing)
- New tests: `test_gcp_location_display`, `test_gcp_location_is_global`, `test_build_host_url_global`, `test_build_host_url_regional`
- Code formatted and linted: `cargo fmt --check` ✓, `./scripts/clippy-lint.sh` ✓

**Manual Live Testing (100% Working):**
Built release binary and tested with actual GCP Vertex AI API:
- ✅ `gemini-3-flash-preview` with `GCP_LOCATION=global` - **SUCCESS**
  - Question: "What is the capital of France?"
  - Response: "Paris" (correct)
  - Confirms global endpoint URL works: `https://aiplatform.googleapis.com`
- ✅ `gemini-3-pro-preview` with `GCP_LOCATION=global` - **SUCCESS**
  - Question: "What is 5 times 7?"
  - Response: "35" (correct)
  - Confirms global endpoint URL works for different Gemini 3 model
- ✅ `gemini-2.0-flash-001` with `GCP_LOCATION=us-central1` - **SUCCESS**
  - Question: "What is 2+2?"
  - Response: "4" (correct)
  - Baseline verification that regional endpoints still work

**Verification:**
- No 404 errors (which would indicate wrong URL construction)
- Responses stored in session database confirm successful API calls
- Both Gemini 3 models now work with global endpoint (previously failed)

### Related Issues
Fixes #6186

## Changes
- Added `GcpLocation::Global` variant with `is_global()` helper method
- Added `build_host_url()` method to handle global vs regional URL construction:
  - Global: `https://aiplatform.googleapis.com`
  - Regional: `https://{location}-aiplatform.googleapis.com`
- Removed `known_location()` method and fallback retry logic (simplification)
- Removed `TryFrom<&str>` impl for `GcpLocation` (no longer needed)

## Impact
- Net -34 lines (92 removed, 58 added)
[...]

Issue #6186, opened 2025-12-19
GCP Vertex AI provider doesn't support global endpoint for Gemini 3 models


clouatre:
**Describe the bug**

Gemini 3 models (`gemini-3-pro-preview`, `gemini-3-flash-preview`) require the global endpoint on Vertex AI. Setting `GCP_LOCATION=global` does not work because the provider constructs an invalid URL.

In `crates/goose/src/providers/gcpvertexai.rs`:
```rust
let host = format!("https://{}-aiplatform.googleapis.com", location);
```

This produces `https://global-aiplatform.googleapis.com` instead of `https://aiplatform.googleapis.com`.

---

**To Reproduce**
1. Set config:
   ```yaml
   GOOSE_PROVIDER: gcp_vertex_ai
   GOOSE_MODEL: gemini-3-pro-preview
   GCP_LOCATION: global
   ```
2. Run `goose session`
3. Send any message
4. Request fails

---

**Expected behavior**

When `GCP_LOCATION=global`, use endpoint `https://aiplatform.googleapis.com` with `locations/global` in the path.

---

**Please provide the following information**
- **OS & Arch:** macOS 26.1 arm64
- **Interface:** CLI
- **Version:** 1.17.0
- **Extensions enabled:** Developer Tools
- **Provider & Model:** GCP Vertex AI – gemini-3-pro-preview

---

**Additional context**

Regional endpoint returns 404:
```bash
curl -X POST "https://us-central1-aiplatform.googleapis.com/.../gemini-3-pro-preview:generateContent"
# 404 NOT_FOUND
```

Global endpoint works:
```bash
curl -X POST "https://aiplatform.googleapis.com/v1/projects/{project}/locations/global/publishers/google/models/gemini-3-pro-preview:generateContent"
# Success
```

PR #6197, opened 2025-12-19
Claude 3.7 is out. we had some harcoded stuff

DOsinga:
## Summary
removed refrences to 3.7 and simplified some code


crates/goose-cli/src/session/output.rs
```diff
@@ -809,10 +809,9 @@ fn normalize_model_name(model: &str) -> String {
         result = re_date.replace(&result, "").to_string();
     }
 
-    // Convert version numbers like -3-7- to -3.7- (e.g., claude-3-7-sonnet -> claude-3.7-sonnet)
-    let re_version = Regex::new(r"-(\d+)-(\d+)-").unwrap();
-    if re_version.is_match(&result) {
-        result = re_version.replace(&result, "-$1.$2-").to_string();
+    let re_version_separator_pattern = Regex::new(r"-(\d+)-(\d+)-").unwrap();
```
katzdave: map_to_canonical_model should be able to do this for you (although maybe not if the model ends up getting removed from openrouter)

crates/goose/src/providers/formats/anthropic.rs
```diff
@@ -394,9 +394,9 @@ pub fn create_request(
         return Err(anyhow!("No valid messages to send to Anthropic API"));
     }
 
-    // https://docs.anthropic.com/en/docs/about-claude/models/all-models#model-comparison-table
-    // Claude 3.7 supports max output tokens up to 8192
-    let max_tokens = model_config.max_tokens.unwrap_or(8192);
+    // https://platform.claude.com/docs/en/about-claude/models/overview
+    // 64k output tokens works for most claude models:
+    let max_tokens = model_config.max_tokens.unwrap_or(64 * 1024);
```
katzdave: We should use max_completion_tokens from canonical model (and more generally across all providers). I can do a followup pass with that. 

Seems like the data there correctly has 64k.

katzdave (approved):

clouatre:
Hi @DOsinga - I noticed our PRs both touch `gcpvertexai.rs`. PR #6187 addresses issue #6186 (global endpoint support for Gemini 3 models) and was submitted yesterday.

Your refactoring to simplify the model enums looks great! When you rebase to resolve conflicts, the key pieces from #6187 that need to stay are:
- `GcpLocation::Global` variant with `is_global()` helper
- `build_host_url()` method for global vs regional URL construction

These handle the case where Gemini 3 models need `https://aiplatform.googleapis.com` rather than `https://global-aiplatform.googleapis.com`.

Happy to help coordinate if needed!

PR #6205, opened 2025-12-19
feat: improved UX for tool calls via execute_code

alexhancock:
**Change**

Improves the UX when in code mode. Makes a new `tool_graph` argument the model will provide to outline the flow of execution that will occur with the code it wrote. The client then uses this to render information about what the code it wrote will do (how many tools were called, what tools were called, etc) still with an option to see the code.

**Demos**

Desktop:
Screenshot 2025-12-19 at 3 39 52 PM

CLI:
Screenshot 2025-12-19 at 3 56 45 PM

### DIFF
--- a/crates/goose-cli/src/session/output.rs
+++ b/crates/goose-cli/src/session/output.rs
@@ -271,6 +271,7 @@ fn render_tool_request(req: &ToolRequest, theme: Theme, debug: bool) {
         Ok(call) => match call.name.to_string().as_str() {
             "developer__text_editor" => render_text_editor_request(call, debug),
             "developer__shell" => render_shell_request(call, debug),
+            "code_execution__execute_code" => render_execute_code_request(call, debug),
             "subagent" => render_subagent_request(call, debug),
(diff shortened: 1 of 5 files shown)


codefromthecrypt (approved):
Nice I personally like this as when auditing manually found sometimes not sure what it intends vs what it wrote

michaelneale:
I like it, one suggestion, in GUI, could it say "N tools used" (don't need to add the detail around execute code as that looks a bit extra detailed)? If there is just 1 or 2, could it show it right next to it (not sure if that would be better but just another thought) but just to keep it simple, but yeah, this is great. 
This is the sort of thing I was hoping for! nice one!

codefromthecrypt:
Also this seems neat to add to traces at some point.. like I wonder if there would be a "code mode eval"

alexhancock:
@michaelneale Good suggestions - updated and merging now!

hongquan:
Please expose this detail via ACP, too.
I'm using Goose inside [Toad](https://github.com/batrachianai/toad).

PR #6210, opened 2025-12-20
fix deeplink recipe launch cold start

zanesq:
## Summary

Fix deep link handling on cold start (app not running) for extension, session, and recipe deep links.

Also noticed deeplinks stopped launching from cold start after nextcamp.

Goose created this but I wasn't able to test cold start launch it would only launch my goose app in applications and copying to Applications wasn't getting around it so we'll need to see after the next release. Verified deeplinks still open when app is running so there should be no regression here only hopeful improvement :) 

fixes https://github.com/block/goose/issues/3972

### Problem

When clicking a `goose://` deep link while Goose is not running:
1. **Extension deep links** failed silently - the app launched but the extension install modal never appeared
2. **Recipe deep links** failed silently - the app launched but the recipe didn't run
3. **Duplicate windows** could be created - both the `open-url` handler and `appMain()` would create windows

### Root Cause

Two issues in the `open-url` event handler:

1. **Missing `app.whenReady()` call**: On macOS, `open-url` can fire before the app is ready. The handler was calling `createChat()` (which creates a `BrowserWindow`) without waiting for the app to be ready, causing the window creation to fail silently.

2. **Race condition for IPC messages**: For extension/session deep links, IPC messages (`add-extension`, `open-shared-session`) were sent immediately after window creation, before React had mounted and registered its listeners. The messages were lost.

### Solution

1. Added `await app.whenReady()` in the `open-url` handler before creating any windows (critical fix for cold start)
2. Added `openUrlHandledLaunch` flag to track when `open-url` has already created a window for a deep link
3. Modified `open-url` handler to defer IPC messages for extension/session deep links until React is ready (stored in `pendingDeepLink`)
4. Modified `appMain()` to skip window creation if `openUrlHandledLaunch` is true
5. Modified `react-ready` IPC handler to process `pendingDeepLink` by sending the appropriate IPC message to the now-ready window

### Result

- App properly launches on cold start deep links (was silently failing before)
- Single window opens (no duplicates)
- Extension install modal appears correctly when clicking extension deep links
- Recipe deep links now work correctly on cold start
- Session sharing deep links now work on cold start

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -227,7 +227,6 @@ if (process.platform !== 'darwin') {
 
 let firstOpenWindow: BrowserWindow;
 let pendingDeepLink: string | null = null;
-// Track if open-url already handled window creation during cold start
 let openUrlHandledLaunch = false;
 
 async function handleProtocolUrl(url: string) {
@@ -310,7 +309,6 @@ app.on('open-url', async (_event, url) => {
 
     log.info('[Main] Received open-url event:', url);
 
-    // Wait for app to be ready before creating window (critical for cold start)
     await app.whenReady();
 
     const recentDirs = loadRecentDirs();
@@ -1914,8 +1912,6 @@ async function appMain() {
 
   const { dirPath } = parseArgs();
 
-  // Only create a new window if open-url didn't already handle the launch
-  // This prevents duplicate windows when launching via deep link on cold start
   if (!openUrlHandledLaunch) {
     await createNewWindow(app, dirPath);
   } else {


ui/desktop/src/main.ts
```diff
@@ -305,19 +307,24 @@ let windowDeeplinkURL: string | null = null;
...
     if (parsedUrl.hostname === 'bot' || parsedUrl.hostname === 'recipe') {
       console.log('[Main] Detected bot/recipe URL, creating new chat window');
+      openUrlHandledLaunch = true;
       const deeplinkData = parseRecipeDeeplink(url);
       if (deeplinkData) {
         windowDeeplinkURL = url;
       }
       const scheduledJobId = parsedUrl.searchParams.get('scheduledJob');
 
+      // Wait for app to be ready before creating window (critical for cold start)
```
alexhancock: i'd nuke the comments

DOsinga: goose run --text "look at the current branch and compare with main. remove any comments that were introduced that were introduce that just say what the code underneath is doing, i.e. commenting on what or how and not on why."

zanesq: sure will do, actually I did review these and they seem relevant for the cold start edge case for future devs

PR #6217, opened 2025-12-21
feat: default todo message

tlongwell-block:
Improve todo usage for autonomous work:
- Remove "2+ steps" threshold from todo instructions
- Show default reminder when todo is empty

Goose tends to forget requirements during longer autonomous tasks. This nudges it to write things down immediately rather than deciding whether a task is "complex enough" to warrant tracking.

### DIFF
--- a/crates/goose/src/agents/todo_extension.rs
+++ b/crates/goose/src/agents/todo_extension.rs
@@ -50,30 +50,24 @@ impl TodoClient {
                 icons: None,
                 website_url: None,
             },
-            instructions: Some(indoc! {r#"
-                Task Management
-
-                Your TODO content is automatically available in your context.
+            instructions: Some(
+                indoc! {r#"
+                Your todo content is automatically available in your context.
 
                 Workflow:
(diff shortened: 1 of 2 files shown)


crates/goose/src/agents/todo_extension.rs
```diff
@@ -50,31 +50,24 @@ impl TodoClient {
...
-
                 Template:
-                - [ ] Implement feature X
-                  - [ ] Update API
-                  - [ ] Write tests
-                  - [ ] Run tests
-                  - [ ] Run lint
-                - [ ] Blocked: waiting on credentials
-            "#}.to_string()),
+                - [x] Requirement 1
```
michaelneale, line 63: is this missing some template wording @tlongwell-block ?

tlongwell-block: I think it's right in the diff. Just hard to grok in the unified view. I trimmed it down a lot because I think it was a bit repetitious

Issue #4018, opened 2025-08-12
Goose requesting keychain password repeatedly


kentcdodds:
**Describe the bug**

I just installed the latest version of Goose and every time I open it up, I get the following error:

<img width="546" height="284" alt="Image" src="https://github.com/user-attachments/assets/8260efc9-e162-4464-b327-dde888191bbd" />

**To Reproduce**
1. Install Goose (I tried downloading and using homebrew)
2. Open Goose
3. See the request
4. Provide the password
5. See the request appear again

**Expected behavior**

I expect to only have to fill that in once.

**Please provide following information:**
 - **OS & Arch:** macOS 15.6 arm64
 - **Interface:** UI
 - **Version:** v1.3.0 (latest)
 - **Extensions enabled:** N/A
 - **Provider & Model:** N/A

michaelneale:
could this be related to https://github.com/block/goose/pull/4007 @jamadeo ?

jamadeo:
I don't think so @michaelneale , that PR introduces a new use of the keyring where we weren't using it before.

@kentcdodds @michaelneale I think you're seeing this because we changed the keys used for the code signing. Are you clicking "Always allow" after entering your password? If so, it should ask only once.

kentcdodds:
Verified using "always allow" makes it go away after entering it twice. I thought I'd done that already but it's working now. Thanks!

PR #6219, opened 2025-12-21
fix: prevent keychain requests during cargo test

codefromthecrypt:
## Summary

When `OPENAI_API_KEY` is set via environment variable, Goose still prompted for keychain access because secondary config lookups (e.g. `OPENAI_CUSTOM_HEADERS`) fell through to keyring.

This adds `get_secrets(primary, maybe_secret)` that checks if the primary key exists in environment variables. If it does, all lookups use env-only mode, completely bypassing keyring. Secondary keys also fall back to `get_param()` automatically.

### Type of Change
- [x] Bug fix

### AI Assistance

### Testing

Verified no keychain access occurs by temporarily adding a panic in the keyring path - all tests still passed.

### Related Issues
Relates to #4018

### DIFF
--- a/crates/goose/src/config/base.rs
+++ b/crates/goose/src/config/base.rs
@@ -19,9 +19,6 @@ const KEYRING_SERVICE: &str = "goose";
 const KEYRING_USERNAME: &str = "secrets";
 pub const CONFIG_YAML_NAME: &str = "config.yaml";
 
-#[cfg(test)]
-const TEST_KEYRING_SERVICE: &str = "goose-test";
-
 #[derive(Error, Debug)]
 pub enum ConfigError {
     #[error("Configuration value not found: {0}")]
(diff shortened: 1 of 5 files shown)


michaelneale:
well worthwhile when running local unit tests for productivity.

crates/goose/src/config/base.rs
```diff
@@ -19,9 +19,6 @@ const KEYRING_SERVICE: &str = "goose";
 const KEYRING_USERNAME: &str = "secrets";
 pub const CONFIG_YAML_NAME: &str = "config.yaml";
 
-#[cfg(test)]
-const TEST_KEYRING_SERVICE: &str = "goose-test";
```
codefromthecrypt, line 23: FYI, this may look scary, but AFAICT tests never actually used the keyring.

The tests calling `Config::new(..., TEST_KEYRING_SERVICE)` only exercised set_param/get_param - which read/write YAML config, not secrets. The keyring service parameter was passed but never touched.

So, removing this is mainly to prevent us from accidentally depending on keyring access in tests that don't actually need it.

michaelneale: heh. wow.

crates/goose/src/config/base.rs
```diff
@@ -729,6 +726,31 @@ impl Config {
             .and_then(|v| Ok(serde_json::from_value(v.clone())?))
     }
 
+    /// Get secrets. If primary is in env, use env for all keys. Otherwise use secret storage.
```
michaelneale, line 729: so - prefer env? (what if you only want it in secrets? is that not needed here?)

crates/goose/src/providers/openai.rs
```diff
@@ -67,7 +67,8 @@ impl OpenAiProvider {
         let model = model.with_fast(OPEN_AI_DEFAULT_FAST_MODEL.to_string());
 
         let config = crate::config::Config::global();
-        let api_key: String = config.get_secret("OPENAI_API_KEY")?;
+        let secrets = config.get_secrets("OPENAI_API_KEY", &["OPENAI_CUSTOM_HEADERS"])?;
```
michaelneale, line 70: ah I think I get it - we still have get_secret which works as before (required in secrets?) vs get_secrets which will use env if it finds the key there (and then get all other values as well).

michaelneale: actually I think this is simiar to original get_secret but this is for convenience.. yeah I like it.

michaelneale:
yeah I think I like this, makes sense. Also will make it easier for goose to operate on itself as won't be unignorable dialogs popping up when you set it on a task.

crates/goose/src/config/base.rs
codefromthecrypt: yeah if we want to add an ignore-env mode we'd want to propagate it down here in a future change.

codefromthecrypt:
missed a couple spots.. hopefully these are the last https://github.com/block/goose/pull/6231

PR #6228, opened 2025-12-22
feat: option to stream json - jsonl really

michaelneale:
--output stream-json

so it can stream output to stdout as it goes, simmilar to other cli tools.

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -1146,7 +1146,6 @@ impl CliSession {
                                         },
                                     };
 
-                                    // Handle stream-json mode for notifications
                                     if is_stream_json_mode {
                                         emit_stream_event(&StreamEvent::Notification {
                                             extension_id: extension_id.clone(),


crates/goose-cli/src/session/mod.rs
```diff
@@ -814,6 +850,14 @@ impl CliSession {
     ) -> Result<()> {
         // Cache the output format check to avoid repeated string comparisons in the hot loop
         let is_json_mode = self.output_format == "json";
```
DOsinga, line 851: that comment above is dumb, let's remove. I also think that actually switching on the output format will lead to more readable code; now I have to think whether is_stream_json_mode and is_json_mode are mutually exclusive or that one is a specialization of the other

crates/goose-cli/src/session/mod.rs
```diff
@@ -1101,9 +1147,15 @@ impl CliSession {
                                         },
                                     };
 
+                                    // Handle stream-json mode for notifications
+                                    if is_stream_json_mode {
```
DOsinga, line 1149: shouldn't we just always do this and have the client ignore it if not present/

michaelneale: I wonder if clients that were using --json would break though - as they expect one blob at the end right vs a whole lot of jsonl like blobs?

DOsinga: oh maybe I am misreading the code

crates/goose-cli/src/session/mod.rs
michaelneale: I think streaming is more a specialisation.

crates/goose-cli/src/cli.rs
```diff
@@ -753,13 +753,13 @@ enum Command {
         )]
         additional_sub_recipes: Vec<String>,
 
-        /// Output format (text, json)
+        /// Output format (text, json, stream-json)
```
michaelneale, line 756: @DOsinga I guess at this level they are distinct - and this is consistent with what others like codex do

crates/goose-cli/src/session/mod.rs
DOsinga: I think it would be better to keep it as strings or use an enum but up to you

PR #6231, opened 2025-12-22
test: fix recipe and audio tests to avoid side effects

codefromthecrypt:
## Summary
Fix tests that cause side effects during `cargo test`:
- **recipe.rs**: Tests were launching Goose Desktop via `open::that()`. Added injectable opener pattern with `handle_open_with()` so tests can mock the opener and verify URLs without launching external apps. Added failure case tests.
- **audio.rs**: Test was triggering keychain access. Added `EnvVarGuard` for tokio-safe env var manipulation to set `OPENAI_API_KEY` before `AppState::new()`.

### Type of Change
- [x] Tests

### AI Assistance

### Testing
- `cargo test` passes with no keychain popups or Goose Desktop launches
- Added `test_handle_open_opener_fails` and `test_handle_open_invalid_recipe` to verify error handling paths

### Related Issues
Completes #6219


codefromthecrypt:
ran `cargo test` manually 3 times in a row and no keychain requests

crates/goose-cli/src/commands/recipe.rs
```diff
@@ -250,47 +266,82 @@ instructions: "Test instructions"
...
+        let temp_dir = TempDir::new().unwrap();
         let recipe_path =
             create_test_recipe_file(&temp_dir, "test_recipe.yaml", VALID_RECIPE_CONTENT);
 
+        let (base_url, _) = generate_deeplink(&recipe_path, HashMap::new()).unwrap();
+
         let params = vec!["name=Alice".to_string(), "role=developer".to_string()];
-        let result = handle_open(&recipe_path, &params);
-        // The result may be Ok or Err depending on whether the system can open the URL
-        // In a test environment, it will likely fail to open, but that's fine
```
DOsinga, line 284: this sounds like an LLM wanting to write a test but not knowing how to

crates/goose-server/src/routes/audio.rs
```diff
@@ -393,8 +393,43 @@ mod tests {
...
+                    None => std::env::remove_var(key),
+                }
+            }
+        }
+    }
+
     #[tokio::test(flavor = "multi_thread")]
     async fn test_transcribe_endpoint_requires_auth() {
+        let _guard = EnvVarGuard::new(&["OPENAI_API_KEY"]);
+        _guard.set("OPENAI_API_KEY", "fake-openai-no-keyring");
```
DOsinga: I don't think this actually helps, does it? if another test runs at the same time, it will pick up this fake OPEN_API_KEY

michaelneale: isn't EnvVarGuard for test scoped environments and would clear between tests? https://crates.io/crates/env-guard

DOsinga: oh, you are right I guess. still though, if it locks the process

codefromthecrypt: if ok to use a small crate from another, I can add this one that is more explicit about its intent and change the two call sites (this is a copy of the same code in another test) https://github.com/LucasPickering/env-lock

michaelneale: that one looks good

crates/goose/src/providers/factory.rs
```diff
@@ -362,24 +308,16 @@ mod tests {
...
             ModelConfig::new_or_fail("gpt-3.5-turbo").with_context_limit(Some(16_000));
 
         let _result = create_lead_worker_from_env("openai", &default_model, "gpt-4o");
-
-        _guard.set("GOOSE_WORKER_CONTEXT_LIMIT", "32000");
-        let _result = create_lead_worker_from_env("openai", &default_model, "gpt-4o");
-
-        _guard.set("GOOSE_CONTEXT_LIMIT", "64000");
-        let _result = create_lead_worker_from_env("openai", &default_model, "gpt-4o");
     }
```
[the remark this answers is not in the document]
codefromthecrypt: actually this test doesn't do anything, except show it doesn't crash with ENV, I will rewrite it

crates/goose-server/src/routes/audio.rs
codefromthecrypt: fixed a bug once I actually made the tests assert something. PTAL!

crates/goose/src/providers/factory.rs
```diff
@@ -222,14 +222,10 @@ fn create_worker_model_config(default_model: &ModelConfig) -> Result<ModelConfig
...
 
-    if let Ok(limit_str) = global_config.get_param::<String>("GOOSE_WORKER_CONTEXT_LIMIT") {
-        if let Ok(limit) = limit_str.parse::<usize>() {
-            worker_config = worker_config.with_context_limit(Some(limit));
-        }
-    } else if let Ok(limit_str) = global_config.get_param::<String>("GOOSE_CONTEXT_LIMIT") {
-        if let Ok(limit) = limit_str.parse::<usize>() {
-            worker_config = worker_config.with_context_limit(Some(limit));
-        }
+    if let Ok(limit) = global_config.get_param::<usize>("GOOSE_WORKER_CONTEXT_LIMIT") {
```
codefromthecrypt, line 225: unrelated bug but fixed while here. We didn't actually read anything in our previous tests so they were not really helpful before

crates/goose/src/providers/factory.rs
```diff
@@ -238,148 +234,76 @@ fn create_worker_model_config(default_model: &ModelConfig) -> Result<ModelConfig
...
+            ("OPENAI_API_KEY", Some("fake-openai-no-keyring")),
         ]);
 
-        _guard.set("OPENAI_API_KEY", "fake-openai-no-keyring");
-        _guard.set("GOOSE_LEAD_MODEL", "grok-3");
-
-        let result = create("openai", ModelConfig::new_or_fail("gpt-4o-mini")).await;
-
-        match result {
-            Ok(_) => {}
```
codefromthecrypt, line 326: rewrote all the assertions here as they did nothing

PR #6237, opened 2025-12-22
Bump release version to 1.18.0

katzdave:
Bumping release version manually since release PR isn't mergable.https://github.com/block/goose/pull/6194

Also re-ran canonical model generation.


jamadeo:
re-ran the actions on https://github.com/block/goose/pull/6194

Issue #6196, opened 2025-12-19
issues with LaTeX rendering in agent response


zanesq:
**Describe the bug**

Two issues have come up from the [KaTeX](https://sq-block.slack.com/archives/C06PBQ52MCK/p1766011595780379?thread_ts=1766011301.677329&cid=C06PBQ52MCK) implementation

Unfortunately I don't have diagnostics or sessions to share because they are internal.

1. treating _ as a subscript notation

<img width="1110" height="230" alt="Image" src="https://github.com/user-attachments/assets/221989a3-fe2e-4d40-86ff-e82e6ee0a187" />

From users: 
you'll want to escape any _ as \_ in the markup

NOTE: potential [fix](https://github.com/block/goose/pull/6162) opened but subsequently closed due to not having time to wrap it up, could be reopened and tested/verified cc @alex-ameen 

2. run on statements

<img width="596" height="55" alt="Image" src="https://github.com/user-attachments/assets/63a0c4cf-7dfe-4340-b61a-65f58a5551b6" />

**Please provide the following information**
- **OS & Arch:** [Mac ARM]
- **Interface:** [UI]
- **Version:** [v1.16.1]
- **Extensions enabled:** [e.g. Computer Controller, Figma]
- **Provider & Model:** [e.g. Google – gemini-1.5-pro]

---

**Additional context**

alex-ameen:
Yeah I had a hunch right off the bat that it was a LaTeX markup because that's the only one I know that treats `x_2` as "x sub-script 2".

It was showing up for me when Goose tried to display shell scripts it was editing which contain multiple `$FOO_BAR a b c 1 2 3 $BAZ` type blocks which are wrapped in `$..$` activating the LaTeX engine.

The use of `_` in variables there gets confused as subscripts.

You should tell Goose NOT to attempt to run LaTeX engine unless it is absolutely certain the user is editing LaTeX documents, or is working on their PhD thesis ( in which case Goose should assume all text should be treated as LaTeX ;) )

PR #6242, opened 2025-12-22
fix: prevent KaTeX from treating underscores as subscripts in plain text

Abhijay007:
Closes: https://github.com/block/goose/issues/6196

## PR Summary

This PR fixes an issue with KaTeX math rendering where underscores (`_`) were incorrectly treated as subscripts. As a result, text such as `$FOO_BAR` or `variable_name` was rendered in math mode, making code snippets and shell examples unreadable.

The root cause was that  `remark-math` plugin treats single-dollar expressions (`$text$`) as inline math by default. When users write shell variables like `$FOO_BAR`, it unintentionally enters math mode, and KaTeX interprets `_` as a subscript operator.

### Configuration Fix
- Set `singleDollarTextMath: false` in the `remarkMath` plugin.
- Single `$` is now treated as plain text, preventing accidental math rendering.
- Users must use `$$...$$` for display math.

### Files Modified
- `ui/desktop/src/components/MarkdownContent.tsx` — added configuration and preprocessing logic.

## Trade-off

**Note:** Users must now use `$$...$$` for math instead of `$...$`.

**Why this is acceptable:**
- Goose is primarily a coding assistant, where code and shell commands are more common than math.
- This matches the behavior of platforms
- This approach aligns with recommendations from the `remark-math` maintainers and is adopted by other platforms (e.g., GitLab) facing similar issues.

## Type of Change
- [x] Bug fix

## AI Assistance
- [x] This PR was created or reviewed with AI assistance.

## Testing

Tested in the desktop UI with:
- Plain text containing `$...$`
- Math expressions using `$$...$$`
- Shell commands and code snippets to ensure underscores render correctly.

## Screenshots / Demos (UX)

**Before:**

beforeKatex

**After:**

afterKatex

### DIFF
--- a/ui/desktop/src/components/MarkdownContent.test.tsx
+++ b/ui/desktop/src/components/MarkdownContent.test.tsx
@@ -390,4 +390,57 @@ Another very long URL: https://www.example.com/very/long/path/with/many/segments
       expect(markdownContainer).toHaveClass('prose-a:overflow-wrap-anywhere');
     });
   });
+
+  describe('KaTeX Math Rendering - singleDollarTextMath: false', () => {
+    it('treats single dollar signs as plain text', async () => {
+      const content = 'The formula $x_i$ represents the i-th element.';
+
+      const { container } = render(<MarkdownContent content={content} />);
+
+      await waitFor(() => {
+        const katexElements = container.querySelectorAll('.katex');
+        expect(katexElements.length).toBe(0);
+        expect(container).toHaveTextContent('$x_i$');
+      });
+    });
+
+    it('renders double dollar signs as display math', async () => {
+      const content = `Calculate
+
+$$
+x^2 + y^2
+$$
+
+for the result.`;
+
+      const { container } = render(<MarkdownContent content={content} />);
+
+      await waitFor(() => {
+        const katexDisplay = container.querySelector('.katex-display');
+        expect(katexDisplay).toBeInTheDocument();
+      });
+    });
+
+    it('handles shell commands without triggering math mode', async () => {
+      const content = 'Run echo "$FOO_BAR" to see the value.';
+
+      const { container } = render(<MarkdownContent content={content} />);
+
+      await waitFor(() => {
+        const katexElements = container.querySelectorAll('.katex');
+        expect(katexElements.length).toBe(0);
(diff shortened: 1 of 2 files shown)


DOsinga:
why does `$FOO_BAR` enter math mode? shouldn't there be a closing tag $

Abhijay007:
> why does `$FOO_BAR` enter math mode? shouldn't there be a closing tag $

The thing is with the default `remark-math` configuration, `singleDollarTextMath` is enabled (`true`) by default, which means a single `$` is enough to trigger inline math mode. When the parser encounters a `$`, it starts looking for a closing `$`. 

However, within math mode, the underscore character (`_`) is interpreted as a LaTeX subscript operator. As a result, a string like `$FOO_BAR` is parsed as the start of inline math, where `FOO_BAR` is interpreted as `FOO` with `BAR` as a subscript. Even if a closing `$` is missing, `KaTeX`/`remark-math` still attempts to parse it as incomplete math, which leads to rendering issues.

DOsinga:
yes, but my concern is that if you run this test:

  describe('KaTeX Math Rendering - singleDollarTextMath: false', () => {
    it('treats single dollar signs as plain text', async () => {
      const content = 'The formula $x_i represents the i-th element.';

      const { container } = render(<MarkdownContent content={content} />);

      await waitFor(() => {
        const katexElements = container.querySelectorAll('.katex');
        expect(katexElements.length).toBe(0);
        expect(container).toHaveTextContent('$x_i');
      });
    });
    });


on main, it succeeds, so it is not that if it fails to find a closing $ it executes.

Abhijay007:
> yes, but my concern is that if you run this test:
> 
> describe('KaTeX Math Rendering - singleDollarTextMath: false', () => { it('treats single dollar signs as plain text', async () => { const content = 'The formula $x_i represents the i-th element.';
> 
> ```
>   const { container } = render(<MarkdownContent content={content} />);
> 
>   await waitFor(() => {
>     const katexElements = container.querySelectorAll('.katex');
>     expect(katexElements.length).toBe(0);
>     expect(container).toHaveTextContent('$x_i');
>   });
> });
> });
> ```
> 
> on main, it succeeds, so it is not that if it fails to find a closing $ it executes.

Um I tried this test on main, it's failing for me for the same reason that it counts it as a subscript

PR #6245, opened 2025-12-22
Clean up

DOsinga:
## Summary

Simplify things a bit. Frittered lives and all that


DOsinga:
FYI - I'm just going to merge this, we can revisit when you are back @aharvard

PR #6246, opened 2025-12-22
chore: Update gemini versions in test_providers.sh

michaelneale:
cover gemini 3 flash (preview)


michaelneale:
yeah - eventually we drop the -preview.

PR #6248, opened 2025-12-23
Add streaming support to bedrock provider

cronus42:
## Summary
Add streaming support to bedrock provider

### Type of Change
- [X] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [X] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested on debian linux

### DIFF
--- a/crates/goose/src/session/session_manager.rs
+++ b/crates/goose/src/session/session_manager.rs
@@ -1074,50 +1074,57 @@ impl SessionStorage {
         Ok(Conversation::new_unvalidated(messages))
     }
 
+    /// Returns true if a message is a "simple" assistant text message that is safe to
+    /// merge with adjacent assistant messages for consolidation purposes. This is
+    /// intentionally provider-agnostic and only inspects internal MessageContent.
+    fn is_simple_assistant_text(message: &Message) -> bool {
+        if message.role != rmcp::model::Role::Assistant {
+            return false;
+        }
+
+        if message.content.len() != 1 {
+            return false;
+        }
+
+        matches!(message.content.first(), Some(MessageContent::Text(_)))
+            && !message.content.iter().any(|c| {
+                matches!(
+                    c,
+                    MessageContent::ToolRequest(_) | MessageContent::ToolResponse(_)
+                )
+            })
+    }
+
     async fn add_message(&self, session_id: &str, message: &Message) -> Result<()> {
         let pool = self.pool().await?;
         let mut tx = pool.begin().await?;
 
-        // Check if we should merge with the last message to prevent fragmentation
-        // This handles streaming responses that create multiple message chunks
-        let should_merge = if message.role == rmcp::model::Role::Assistant {
-            // Only consider merging if this is a simple text-only message
-            let is_simple_text = message.content.len() == 1
-                && matches!(message.content.first(), Some(MessageContent::Text(_)));
+        // Check if we should merge with the last message to prevent fragmentation.
+        // This handles streaming responses that create multiple assistant text chunks.
+        let should_merge = if Self::is_simple_assistant_text(message) {
+            // Get the last message for this session.
+            let last_msg = sqlx::query_as::<_, (String, String)>(
+                "SELECT role, content_json FROM messages WHERE session_id = ? ORDER BY created_timestamp DESC LIMIT 1",
+            )
+            .bind(session_id)
+            .fetch_optional(&mut *tx)
+            .await?;
 
-            if is_simple_text {
-                // Get the last message for this session
-                let last_msg = sqlx::query_as::<_, (String, String)>(
-                    "SELECT role, content_json FROM messages WHERE session_id = ? ORDER BY created_timestamp DESC LIMIT 1"
-                )
-                .bind(session_id)
-                .fetch_optional(&mut *tx)
-                .await?;
-
-                if let Some((last_role, last_content_json)) = last_msg {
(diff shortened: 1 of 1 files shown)


dianed-square:
Hi @cronus42 . Looks like documentation/TODO-BEDROCK.md is a goose artifact

cronus42:
removed the doc, the tests pass when a valid credential is present in the environment with bedrock access. This was assisted by claude and I am not a rust dev, but I am a dev whose trying to be responsible with it.

blackgirlbytes:
looks like your DCO check is failing! Here's how to fix it: https://github.com/block/goose/pull/6248/checks?check_run_id=58942416145

michaelneale:
nice - I think if you update to main, check things are happy again may be in happier state, I thin kthis is worthwhile having. 

Also - don't forget to do the DCO check and sign commits - thanks for this!

michaelneale:
hi @cronus42 do you mind signing the DCO (and if you can update to main - some conflicts to resolve but hopefully just mechanical). would be good to have this.

zanesq:
@cronus42 checking if still interested in wrapping this up?

cronus42:
I think I have all the tests passing and such now

crates/goose/src/session/session_manager.rs
```diff
@@ -482,7 +489,13 @@ impl SessionStorage {
             .busy_timeout(std::time::Duration::from_secs(5))
             .journal_mode(sqlx::sqlite::SqliteJournalMode::Wal);
 
-        SqlitePoolOptions::new().connect_lazy_with(options)
+        // Use a single connection for the session database to avoid SQLite write-lock
```
michaelneale: not sure what all the changes are in session manager - shouldn't these changes be isolated to the provider for bedrock?

ui/desktop/src/components/bottom_menu/CostTracker.tsx
```diff
@@ -70,6 +70,7 @@ export function CostTracker({ inputTokens = 0, outputTokens = 0, sessionCosts }:
   }
 
   const calculateCost = (): number => {
+    // Note: Pricing data from backend is per million tokens, so we divide by 1_000_000
```
michaelneale: not sure of connection here to bedrock?

crates/goose/src/session/session_manager.rs
cronus42: this was just me trying to get the CI tests to pass, should probably move it to another PR

ui/desktop/src/components/bottom_menu/CostTracker.tsx
cronus42: I needed to know how much money I was spending and the cost tracker wasn't working for bedrock provider agents, can also break this off

michaelneale: ah no worries - yeah probably need to sort that out!

crates/goose/src/session/session_manager.rs
michaelneale: there are a lot of other changes to session_manager.rs as well - are they needed?

cronus42: the other changes in session manager are mechanism to consolidate streamed messages into the history so you don't end up with a thought per line when you load the chat history

michaelneale: ah ok - I think we need to have some more testing of that. Other providers would be improved by this wouldn't they?

cronus42: I think that other providers have their own aggregators for openai and anthropic formats? I moved the functionality into a helper in SessionStorage. added tests for several cases

michaelneale: in that case I'm not sure why this needs this much change to session_storeage.rs which is a pretty core bit of code, jsut for bedrock? that doesn't seem the right place for it.

cronus42: The helper could be used by other providers, I just don't really know what other providers, similar to bedrock in implementation, might need it. It looks like the other providers have aggregation at other layers than bedrock does.

DOsinga:
@cronus42 - sorry about the slow replies, do we still want to get this in? can you merge in main?

crates/goose/src/session/session_manager.rs
```diff
@@ -1067,25 +1074,109 @@ impl SessionStorage {
...
+                )
+            })
+    }
+
     async fn add_message(&self, session_id: &str, message: &Message) -> Result<()> {
         let pool = self.pool().await?;
         let mut tx = pool.begin().await?;
 
-        let metadata_json = serde_json::to_string(&message.metadata)?;
+        // Check if we should merge with the last message to prevent fragmentation.
```
DOsinga, line 1102: I don't think this is the way. the other streaming providers just return messages/chunks with the same id and that is what we handle. adding this logic to the database layer kills the performance

michaelneale:
closing this as has moved to a branch

crates/goose/src/session/session_manager.rs
DOsinga: can you explain this more?

the other providers all have the pattern where the streaming messages that need to be concatenated have the same id. we use the Conversation class which if you append a message that has an id we already have just appends it to that message, nice and transparent.

why does that approach not work here?

Issue #6251, opened 2025-12-23
update goose daemon endpoints


baxen:
We're approaching having multiple clients for the goose daemon - especially with goose-ios app finalizing.

The current daemon grew organically to support the electron app, but has left behind some confusing terminology and conventions as we introduced the session database, simplified our config, etc.

This issue will track a refactor across several PRs (might continue breaking this down):

- [ ] #6252 Simplify and consolidate initialization procedure in electron
- [ ] #6260 Move all "agent" apis to "session" apis
- [ ] Simplify config endpoints
- [ ] Standardize the rest (naming and REST conventions)

The biggest change is removing the "agent" apis in favor of session. They're currently mixing concerns, and for the client thinking of sessions is better matched with the concerns. The agent state then becomes a backend implementation detail. 

For **now** we will still keep extension state management client side for consistency and the ability to render session loading progress. But we plan to make the agent take on ownership of extension management in future work.

baxen:
Here's a summary of the now-in-pr updates to agent/sessions 

| Before | After |
|--------|-------|
| `POST /agent/start` | `POST /sessions` |
| `POST /agent/resume` | `POST /sessions/{session_id}/open` |
| `POST /agent/stop` | **DEPRECATED** |
| `POST /agent/update_from_session` | **DEPRECATED** (folded into `/open`) |
| `POST /agent/update_provider` | **DEPRECATED** (provider set via `/open` or create) |
| `POST /agent/add_extension` | `POST /sessions/{session_id}/extensions` |
| `POST /agent/remove_extension` | `DELETE /sessions/{session_id}/extensions/{extension_name}` |
| `GET /agent/tools?session_id=...` | `GET /sessions/{session_id}/tools` |
| `POST /agent/call_tool` | `POST /sessions/{session_id}/tools/{tool_name}/call` |
| `POST /agent/read_resource` | `POST /sessions/{session_id}/resources/read` |
| `POST /action-required/tool-confirmation` | `POST /sessions/{session_id}/confirmations` |
| `PUT /sessions/{id}/user_recipe_values` | `PUT /sessions/{session_id}/recipe` |
| `POST /sessions/{id}/edit_message` (Fork) | `POST /sessions/{session_id}/messages/{timestamp}/fork` |
| `POST /sessions/{id}/edit_message` (InPlace) | `POST /sessions/{session_id}/messages/{timestamp}/edit` |
| `GET /sessions` | `GET /sessions` *(unchanged)* |
[...]

PR #6252, opened 2025-12-23
feat(daemon): refactor session init endpoints

baxen:
add new API endpoints for session lifecycle management:
- POST /sessions/create: Create and initialize a new session with provider, model, recipe, and extensions in a single call
- POST /sessions/{id}/open: Open an existing session and setup agent
- POST /sessions/{id}/extensions: Add extensions to an active session

Consolidates both new window and load from history to use the same extension management and show the toast in both cases

Part 1 of #6251

https://github.com/user-attachments/assets/4c4948d2-e48c-4f5c-8c2a-fd0069913b33


baxen:
to simplify the diff on this one only adds the new endpoints and uses them in a followup pr for #6251 will cleanup unused methods

crates/goose-server/src/routes/session.rs
```diff
@@ -74,8 +87,309 @@ pub struct EditMessageResponse {
...
+}
+
+#[derive(Deserialize, ToSchema)]
+pub struct AddSessionExtensionRequest {
+    pub config: ExtensionConfig,
+}
+
 const MAX_NAME_LENGTH: usize = 200;
 
+async fn setup_agent_for_session(
```
DOsinga, line 120: I think if we move this to routes/agents.rs and then hook it up to agent/start and agent/resume you get the same effect, but less duplication

baxen: yes agree! it's on the next pr #6260

wareagle461 (approved):

michaelneale:
yep this looks ok to me - I like it so far. Doesn't break anything yet! over to you @DOsinga

alexhancock:
this and https://github.com/block/goose/pull/6260 LGTM 

@DOsinga what do you think?

michaelneale:
I say go for it

katzdave:
@baxen still want to get this in or can we close?

michaelneale:
can close this for now - I think drifted a bit, reopen if still relevant?

PR #6264, opened 2025-12-24
feat: add a max tokens env var

baxen:


crates/goose/src/model.rs
```diff
@@ -184,6 +185,27 @@ impl ModelConfig {
         }
     }
 
+    fn parse_max_tokens() -> Result<Option<i32>, ConfigError> {
+        if let Ok(val) = std::env::var("GOOSE_MAX_TOKENS") {
```
DOsinga: rather than using an environment variable, should we make this go through the config module, which checks environment variables anyway and has some light templating for types going on?

baxen: yeah sounds good. i changed just this one field for now rather than the full model config. can come back to doing the other ones in this file that also don't use config

PR #6274, opened 2025-12-25
chore: refactor interactive session to reduce line count

baxen:
## Summary

Refactors the `interactive` function in `crates/goose-cli/src/session/mod.rs` from 275 lines to 35 lines by extracting input handling, command processing, and UI rendering into separate helpers.

### New Types
- **`InputAction` enum**: Controls main loop flow (Continue, Break, Proceed)
- **`HistoryManager` struct**: Encapsulates command history loading/saving logic

### Extracted Helper Methods
- `create_editor()`: Sets up the rustyline editor with completion
- `handle_input()`: Dispatches input commands to appropriate handlers
- `handle_message_input()`: Processes user messages in normal/plan modes
- `handle_toggle_theme()` / `handle_select_theme()`: Theme switching
- `handle_goose_mode()`: Mode switching (auto, approve, chat, smart_approve)
- `handle_plan_mode()`: Plan mode entry and initial message processing
- `handle_clear()`: Session clearing
- `handle_recipe()`: Recipe generation
- `handle_compact()`: Conversation compaction

## Result

The `interactive` function is now 35 lines, well under the 200 line target.

## Testing

- All 112 tests pass
- Clippy lint passes
- Code formatted with cargo fmt

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -106,6 +106,53 @@ pub enum RunMode {
     Plan,
 }
 
+struct HistoryManager {
+    history_file: PathBuf,
+    old_history_file: PathBuf,
+}
+
+impl HistoryManager {
+    fn new() -> Self {
+        Self {
+            history_file: Paths::state_dir().join("history.txt"),
+            old_history_file: Paths::config_dir().join("history.txt"),
+        }
+    }
+
+    fn load(
+        &self,
+        editor: &mut rustyline::Editor<GooseCompleter, rustyline::history::DefaultHistory>,
+    ) {
+        if let Some(parent) = self.history_file.parent() {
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/session/mod.rs
```diff
@@ -106,6 +106,53 @@ pub enum RunMode {
     Plan,
 }
 
+struct HistoryManager {
+    history_file: PathBuf,
+    old_history_file: PathBuf,
+}
+
+impl HistoryManager {
```
DOsinga, line 114: FWIW we should just get rid of this and base it on the session manager I think and then extent it to the desktop (keeps its own history) too maybe?

baxen: agree yeah

PR #4344, opened 2025-08-26
chore: Bump xcap to 0.7.0

shawn111:
xcap 0.7.0
- improve wayland support
- support android build (no function, just enable build)

PR #6284, opened 2025-12-26
feat(code-mode): use server names for MCP extensions

### LINKED ISSUE #6188: code mode: better LLM UX on runtime added extensions

Right now, in code mode an inline MCP server gets a pretty bad name. I don't think it is easy for the LLM to guess it. 

```bash
$ cargo run -p goose-cli --release -- run --text "Search for list_directory and read_text_file tools. Use them to list /tmp and read the first .txt file." --with-builtin code_execution --with-extension "npx -y @modelcontextprotocol/server-filesystem /tmp"
    Finished `release` profile [optimized] target(s) in 0.25s
     Running `target/release/goose run --text 'Search for list_directory and read_text_file tools. Use them to list /tmp and read the first .txt file.' --with-builtin code_execution --with-extension 'npx -y @modelcontextprotocol/server-filesystem /tmp'`
starting session | provider: tetrate model: gpt-5-nano
    session id: 20251219_11
    working directory: /Users/codefromthecrypt/oss/goose

terms: list_directory, read_text_file


module_path: tmp_9zspqbtd/list_directory


module_path: tmp_9zspqbtd/read_text_file
--snip--
```

I wonder if when there is no conflict, we can use the server name or parse some part of it. Probably other projects have a similar concern we can study.

codefromthecrypt:
## Summary

Makes code mode usable in ACP and improves extension naming for better LLM accuracy.

**Code mode in ACP:** Adds `--with-builtin` arg so ACP clients can enable code_execution and developer extensions.

**Better extension names:** When no name is configured (e.g., CLI `--with-streamable-http-extension`), extensions now use server-declared names instead of anonymous identifiers. User-configured names from goose config or ACP sessions are still preferred.

Before (CLI arg, no configured name):
```bash
$ goose run --with-streamable-http-extension 'https://mcp.kiwi.com'   -t 'Use kiwi to find the fastest itinerary from BKI to SYD tomorrow'
--snip--
─── search-flight | mcpkiwicom_yfvx6ext ──────────────────────────
```

After:
```bash
─── search-flight | kiwi-mcp-server ──────────────────────────
```

LLMs can now match extension names to server info in their prompts. Collision suffix added when duplicate names exist.

### Type of Change
- [x] Feature
- [x] Tests

### AI Assistance

### Testing

**New integration test:** `test_acp_with_builtin_and_mcp` - verifies code_execution and MCP servers work together in ACP mode.

**Manual:**
```bash
goose run --with-streamable-http-extension 'https://mcp.kiwi.com'   -t 'Use kiwi to find the fastest itinerary from BKI to SYD tomorrow'
```

### Related Issues
Closes #6188


crates/goose/src/agents/extension_manager.rs
```diff
@@ -145,6 +146,36 @@ fn normalize(input: String) -> String {
...
+    }
+
+    let suffix: String = rand::thread_rng()
+        .sample_iter(Alphanumeric)
+        .take(6)
+        .map(char::from)
+        .collect();
+
+    format!("{base}_{suffix}")
+}
```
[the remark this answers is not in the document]
codefromthecrypt: done

crates/goose/src/agents/extension_manager.rs
```diff
@@ -1781,4 +1814,40 @@ mod tests {
...
+            ServerInfo {
+                server_info: Implementation {
+                    name: name.into(),
+                    ..Default::default()
+                },
+                ..Default::default()
+            }
+        }
+
+        #[test_case("kiwi", Some("kiwi-mcp-server"), None, "^kiwi$" ; "ACP session prefers explicit name")]
```
codefromthecrypt: FYI: all of these are real server names scraped from the public MCP impls

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -892,7 +892,9 @@ impl McpClientTrait for CodeExecutionClient {
 
         let mut servers: BTreeSet<&str> = BTreeSet::new();
         for tool in &tools {
-            servers.insert(&tool.server_name);
+            if tool.server_name != "extensionmanager" {
```
codefromthecrypt: this is the part for code mode UX because unless you hide extensionmanager it will try to iterate its tools

codefromthecrypt: removed this as it was a feature not a bug

crates/goose/src/providers/oauth.rs
```diff
@@ -99,7 +99,7 @@ async fn get_workspace_endpoints(host: &str) -> Result<OidcEndpoints> {
     if !resp.status().is_success() {
         return Err(anyhow::anyhow!(
             "Failed to get OIDC configuration from {}",
-            oidc_url.to_string()
+            oidc_url
```
michaelneale, line 102: unrelated change?

codefromthecrypt: yeah will revert this good catch

crates/goose/tests/acp_integration_test.rs
```diff
@@ -184,17 +250,21 @@ fn extract_text(updates: &[SessionNotification]) -> String {
...
+    }
+    cmd.stdin(Stdio::piped())
         .stdout(Stdio::piped())
         .stderr(Stdio::piped())
         .env("GOOSE_PROVIDER", "openai")
         .env("GOOSE_MODEL", "gpt-5-nano")
         .env("GOOSE_MODE", "approve")
         .env("OPENAI_HOST", mock_server.uri())
         .env("OPENAI_API_KEY", "test-key")
+        .env("GOOSE_PATH_ROOT", data_root)
```
codefromthecrypt, line 267: added this to isolate the DBs between tests, but allow us to intentionally share them. This will be very important in subsequent change for session load, resume and list capabilities

michaelneale: makes sense.

crates/goose/Cargo.toml
```diff
@@ -121,7 +121,8 @@ unbinder = "0.1.7"
 winapi = { version = "0.3", features = ["wincred"] }
 
 [dev-dependencies]
-sacp = "9.0.0"
+sacp = "10.0.0-alpha.1"
+agent-client-protocol-schema = "0.10.5"
```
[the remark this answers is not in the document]
codefromthecrypt: no this intentionally moves off very old versions in preparation of an upgrade

crates/goose-cli/Cargo.toml
```diff
@@ -19,7 +19,8 @@ goose = { path = "../goose" }
 goose-bench = { path = "../goose-bench" }
 goose-mcp = { path = "../goose-mcp" }
 rmcp = { workspace = true }
-sacp = "9.0.0"
+sacp = "10.0.0-alpha.1"
+agent-client-protocol-schema = "0.10.5"
```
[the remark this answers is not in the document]
codefromthecrypt: no this intentionally moves off very old versions in preparation of an upgrade

michaelneale:
I'm not sure why - but the GUI isn't working with this - never get a response, no matter what I do (I just flicked to main and tried again) - if I turn off code mode it is fine.

codefromthecrypt:
@michaelneale good check and I promise to use the gui also from now on or at least until all the things load extensions the same way.

What happened is that on main we load the same extension multiple times. My change was overzealous wrt collisions and it added a suffix if an extension already exists. This is wrong even if loading the same extension multiple times is inefficient.

I will fix the code to only do **collision avoidance on anonymous extensions**. There is a real chance they will collide as all the paths are like /mcp and sometimes even the server name is fake 'StatelessApp'.

codefromthecrypt:
verified the UI (which always names servers) and anonymous CLI ones work

Screenshot 2025-12-29 at 3 01 01 PM


```bash
     Running `target/debug/goose run --with-streamable-http-extension 'https://mcp.kiwi.com' -t 'Use kiwi to find the fastest itinerary from BKI to SYD tomorrow'`
starting session | provider: tetrate model: gpt-5
    session id: 20251229_9
    working directory: /Users/codefromthecrypt/oss/goose

module_path: kiwi-mcp-server
```

michaelneale:
nice one @codefromthecrypt - I wonder what legit test would catch that with gui (perhaps a goosed e2e? I know we have some various states of GUI e2e, but not sure if they always run).

crates/goose-cli/Cargo.toml
```diff
@@ -19,7 +19,8 @@ goose = { path = "../goose" }
 goose-bench = { path = "../goose-bench" }
 goose-mcp = { path = "../goose-mcp" }
 rmcp = { workspace = true }
-sacp = "9.0.0"
+sacp = "10.0.0-alpha.1"
```
michaelneale: is it ok to have an alpha dep?

crates/goose/tests/test_data/openai_tool_call_response.txt
```diff
@@ -1,9 +1,10 @@
...
 
-data: {"id":"chatcmpl-test1","object":"chat.completion.chunk","created":1766309803,"model":"gpt-5-nano","choices":[{"index":0,"delta":{},"finish_reason":"tool_calls"}]}
+data: {"id":"chatcmpl-CqqCVVtD16yj37EZocLFkGNMhHZFS","object":"chat.completion.chunk","created":1766709751,"model":"gpt-5-nano-2025-08-07","service_tier":"default","system_fingerprint":null,"choices":[{"index":0,"delta":{},"finish_reason":"tool_calls"}],"usage":null,"obfuscation":"k965c2jCwUF"}
 
-data: {"id":"chatcmpl-test1","object":"chat.completion.chunk","created":1766309803,"model":"gpt-5-nano","choices":[],"usage":{"prompt_tokens":100,"completion_tokens":50,"total_tokens":150}}
+data: {"id":"chatcmpl-CqqCVVtD16yj37EZocLFkGNMhHZFS","object":"chat.completion.chunk","created":1766709751,"model":"gpt-5-nano-2025-08-07","service_tier":"default","system_fingerprint":null,"choices":[],"usage":{"prompt_tokens":2320,"completion_tokens":149,"total_tokens":2469,"prompt_tokens_details":{"cached_tokens":0,"audio_tokens":0},"completion_tokens_details":{"reasoning_tokens":128,"audio_tokens":0,"accepted_prediction_tokens":0,"rejected_prediction_tokens":0}},"obfuscation":"7Wxwg9X1OBwbjfE"}
```
michaelneale, line 7: a lot more data in these responses - presumably more complete now!

michaelneale:
looks good to me now - main Q around alpha dep (and that the test data changes are deliberate)

crates/goose-cli/Cargo.toml
codefromthecrypt: right now I think so, since ACP isn't mainline, but I will bump it to alpha3. I have active discussions on the ACP zulip with the maintainer to make sure it has the features we need before it goes final

crates/goose/tests/test_data/openai_tool_call_response.txt
codefromthecrypt: yeah the test data is different because with this change, I can use exactly the same data from the CLI for ACP tests since the tool names are predictable. This will help later when we do anything obs related to have real data.

codefromthecrypt:
thanks @michaelneale I'll be on the hook for sacp updates, with no holiday except NY planned next two weeks plus. I'm happy they are iterating and listening so the alpha thing is a good thing!

codefromthecrypt:
@michaelneale wrt e2e I think I can make a playwright to catch things like this with similar faked responses.. It is on my brain, but will work on it after the goosed ACP cli providers stuff

crates/goose/src/providers/oauth.rs
codefromthecrypt: incidental anyhow bump from 1.0.99 → 1.0.100 makes it possible for clippy to figure out the to_string is redundant.

codefromthecrypt:
xcap needed an update to be compatible with core-foundation 0.10.0 → 0.10.1

I'm running a manual flow to capture windows, my screen etc to make sure everything is keen

codefromthecrypt:
Manually tested code affected by the xcap by using the following prompts:

* list all open windows on my screen
* capture a screenshot of the window titled Calculator
* capture a screenshot of my main display

Screenshot 2025-12-30 at 9 21 37 AM

---
Note: Goose doesn't work with capture on macos until authorized, you need to authorize the binary you built from source.

So you need to..
1. Edit Justfile so that it has consistent signatures between Goose.app and goosed
2. `just make-ui` then unzip the result somewhere
3. `open "x-apple.systempreferences:com.apple.preference.security?Privacy_ScreenCapture"`
4. Hit the + sign and drag the Goose binary into that 
5. Then launch the same binary

```diff
diff --git a/Justfile b/Justfile
index e7a6ea9d93..ca75de10aa 100644
--- a/Justfile
+++ b/Justfile
@@ -199,7 +199,18 @@ generate-openapi:
 lint-ui:
     cd ui/desktop && npm run lint:check
 
-# make GUI with latest binary
+# Signing added for screenshots on MacOS
+# open "x-apple.systempreferences:com.apple.preference.security?Privacy_ScreenCapture"
+# and authorize ui/desktop/out/Goose-darwin-arm64/Goose.app before running Goose
+[macos]
+make-ui:
+    @just release-binary
+    cd ui/desktop && npm run bundle:default
+    codesign --force --deep --sign - --entitlements ui/desktop/entitlements.plist ui/desktop/out/Goose-darwin-arm64/Goose.app/Contents/Resources/bin/goosed
+    codesign --force --deep --sign - --entitlements ui/desktop/entitlements.plist ui/desktop/out/Goose-darwin-arm64/Goose.app
+
+[linux]
+[windows]
 make-ui:
     @just release-binary
     cd ui/desktop && npm run bundle:default
```


For example, you should have Goose here, but the important thing is that it is from the path you are actually running
Screenshot 2025-12-30 at 9 36 45 AM


If you don't do this, list windows will partially work only able to read the background image and system menubar. This has nothing to do with this PR, just I wanted to write this down for future self.

codefromthecrypt:
noticed xcap update was attempted in #4344 and the main thing was the CI packages. I used the same literal packages from xcap themselves, so 🤞 it will pass

crates/goose-mcp/Cargo.toml
```diff
@@ -27,7 +27,7 @@ schemars = "1.0"
 lazy_static = "1.5"
 shellexpand = "3.1.0"
 indoc = "2.0.5"
-xcap = "0.0.14"
+xcap = "=0.4.0"
```
codefromthecrypt, line 30: this is the revlock version that allows everything else to pass. Movement past this in #6301

codefromthecrypt:
re-ran all steps manually, choosing xcap 0.4.0 which is the latest version that needs no new platform deps on linux.

codefromthecrypt:
thanks for all the support @michaelneale!

codefromthecrypt:
fyi opened this https://github.com/nashaofu/xcap/pull/250 and once it is merged and when we can upgrade will be a lot easier to detect subtle permissions problem on screen capture that need the user to overcome.

PR #6286, opened 2025-12-26
MCP Apps Plumbing

DOsinga:
## Summary

Hook the client up to the actual sever. Works now with at least some MCP servers


ui/desktop/src/components/ToolCallWithResponse.tsx
```diff
@@ -47,7 +58,49 @@ function isEmbeddedResource(content: Content): content is EmbeddedResource {
...
+  const extensionName = toolCall.name.split('__')[0];
+
+  return (
+    <div className="mt-3">
+      <McpAppRenderer
+        resourceUri={resourceUri}
+        toolInput={{ arguments: toolCall.arguments }}
+        extensionName={extensionName}
+        sessionId={sessionId}
+        append={append}
```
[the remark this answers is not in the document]
DOsinga: huh?

ui/desktop/src/components/sessions/SessionViewComponents.tsx
aharvard, line 1: Does this file change need to be part of this PR?

DOsinga: not really; we're dragging in session-ids into sessionviews and so this would have to maybe updated but then it is not used

aharvard (approved):
Very good! Confirmed it's working with/ my demo MCP App!

Image

michaelneale:
nice - cli code isn't entirely happy: error[E0063]: missing field `tool_meta` in initializer of `goose::conversation::message::ToolRequest`
Error:     --> crates/goose-cli/src/session/export.rs:1244:28
     |
1244 |         let tool_request = ToolRequest {
     |                            ^^^^^^^^^^^ missing `tool_meta`
 
 
 is best way to test it what @aharvard did?

crates/goose/src/agents/extension_manager.rs
```diff
@@ -706,14 +707,13 @@ impl ExtensionManager {
                                 input_schema: tool.input_schema,
                                 annotations: tool.annotations,
                                 output_schema: tool.output_schema,
-                                icons: None,
-                                title: None,
-                                meta: None,
+                                icons: tool.icons,
```
michaelneale, line 716: huh - so will show the tool it is adding?

ui/desktop/src/api/types.gen.ts
```diff
@@ -690,7 +690,12 @@ export type ReadResourceRequest = {
 };
 
 export type ReadResourceResponse = {
-    html: string;
+    _meta?: {
```
michaelneale, line 689: yeah this seems a reasonable extension to what I remember from mcp ui, nice, a generalisation for apps

crates/goose/src/agents/extension_manager.rs
DOsinga: no, this is just for completion sake. we're not really using icons as far as I can tell, but dropping these values on the floor can only be bad :)

aharvard:
> is best way to test it what @aharvard did?

@michaelneale if you want to give it a spin...

**endpoint**: https://mcp-app-bench.onrender.com/mcp 

**prompt**: "run mcp app bench"

crates/goose/src/agents/extension_manager.rs
```diff
@@ -784,14 +783,17 @@ impl ExtensionManager {
...
         // If extension name is provided, we can just look it up
         if extension_name.is_some() {
-            let result = self
-                .read_resource_from_extension(
-                    uri,
-                    extension_name.unwrap(),
-                    cancellation_token.clone(),
-                    true,
-                )
+            let read_result = self
```
jamadeo, line 786: not your code but `if thing.is_some() { ... thing.unwrap() ... }` should be an `if let Some(inner) = thing {... `

ui/desktop/src/components/ToolCallWithResponse.tsx
```diff
@@ -47,7 +71,57 @@ function isEmbeddedResource(content: Content): content is EmbeddedResource {
...
+  }
+
+  if (!resourceUri) return null;
+  if (requestWithMeta.toolCall.status !== 'success') return null;
+
+  const extensionName = requestWithMeta.toolCall.value.name.split('__')[0];
+
+  let toolResult: CallToolResponse | undefined;
+  if (toolResponse) {
+    const resultWithMeta = toolResponse.toolResult as ToolResultWithMeta;
```
jamadeo, line 97: can we type this properly?

DOsinga: If you see a way? I've been going back and forth here, but toolResponse comes through the API from the rcmp classes which have a hard to time getting converted to proper typescript, so we end up with something semi-typed that we cast

PR #6288, opened 2025-12-27
Feat azure OpenAI API migration to response API + minor UI improvement

bioinfornatics:
# Azure OpenAI Responses API support & auth UX improvements

This PR updates Azure OpenAI support in both the CLI and desktop app to match the Azure portal’s Responses API (including gpt‑5.x models) and makes configuration of authentication and endpoints much easier.

## Highlights

- Works out of the box with Responses API URLs copied directly from the Azure portal.
- Clear choice between API key and Entra ID authentication for Azure OpenAI, in both CLI and desktop UI.
- Safer endpoint field behaviour in the desktop app (no more surprise re-population after clearing).
- Fixes 404 "Resource not found" errors caused by mismatched `api-version` values.

---

## Backend / CLI

- **Auth mode switch for Azure OpenAI**

  - New `AZURE_OPENAI_AUTH_TYPE` config key with two modes:
    - `api_key` → sends `api-key: <AZURE_OPENAI_API_KEY>`
    - `entra_id` → sends `Authorization: Bearer <token from Azure CLI / default credentials>`
  - This allows configuring Azure OpenAI either with a static key or with Entra ID, matching the Azure portal’s options.

- **Endpoint and API version handling aligned with the portal**

  - `AZURE_OPENAI_ENDPOINT` now accepts:
    - just the host (`https://<resource>.cognitiveservices.azure.com`)
    - host + path (`…/openai/responses`)
    - a full URL copied from the portal (`…/openai/responses?api-version=2025-04-01-preview`)
  - The value is normalized into:
    - a base endpoint (`scheme://host[:port]`)
    - an optional `api-version` extracted from the query string.
  - `AZURE_OPENAI_API_VERSION` is chosen with this priority:
    1. explicit `AZURE_OPENAI_API_VERSION` in config (if set)
    2. `api-version` parsed from `AZURE_OPENAI_ENDPOINT`
    3. default `AZURE_DEFAULT_API_VERSION`
  - This removes the mismatch where curl with `api-version=2025-04-01-preview` worked but goose used `2025-04-01` and got 404.

- **Model list / defaults**

  - Extends the known Azure models to include `gpt-5.1` and related variants.
  - Updates the default API version to the Responses API so new configs work with the latest models by default.

---

## Desktop UI

- **Authentication type dropdown for Azure OpenAI**

  - In the provider configuration modal (when provider is `azure_openai`):
    - Adds an **Authentication Type** dropdown with:
      - **Key Authentication** → `AZURE_OPENAI_AUTH_TYPE=api_key`
      - **Entra ID Authentication** → `AZURE_OPENAI_AUTH_TYPE=entra_id`
    - When **Entra ID** is selected, the API key field is hidden; when **Key** is selected, the field is shown.
  - The chosen auth type is persisted via the config API and used by the backend.

- **“Paste the portal URL” endpoint UX**

  - Watches edits to `AZURE_OPENAI_ENDPOINT` (with a short debounce ~800 ms) and applies:
    - **a) Endpoint only (no path, no api-version)** → leave as-is.
    - **b) Endpoint + path (e.g. `…/openai/responses`)** → rewrite to just the base endpoint; leave `AZURE_OPENAI_API_VERSION` unchanged.
    - **c) Endpoint + path + `api-version`** → rewrite to base endpoint and set `AZURE_OPENAI_API_VERSION` from the query param.
  - Only applies for the `azure_openai` provider.
  - Edge cases:
    - If the user clears the field, it stays empty (no auto-repopulation).
    - If the URL is incomplete/invalid while typing, no normalization is performed.

- **Fix for config field re-population**

  - Fixes a bug where clearing a field (like the Endpoint) would cause the original server value to reappear.
  - The form now distinguishes:
    - “value explicitly set (including empty string)” vs
    - “no value set yet”,
    and only falls back to the server value in the latter case.


documentation/docs/getting-started/providers.md
```diff
@@ -24,7 +24,7 @@ goose is compatible with a wide range of LLM providers, allowing you to choose a
...
-| [Azure OpenAI](https://learn.microsoft.com/en-us/azure/ai-services/openai/) | Access Azure-hosted OpenAI models, including GPT-4 and GPT-3.5. Supports both API key and Azure credential chain authentication.                                                                                          | `AZURE_OPENAI_ENDPOINT`, `AZURE_OPENAI_DEPLOYMENT_NAME`, `AZURE_OPENAI_API_KEY` (optional)                                                                                           |
+| [Azure OpenAI](https://learn.microsoft.com/en-us/azure/ai-services/openai/) | Access Azure-hosted OpenAI models, including GPT-4 and GPT-3.5. Supports both API key and Azure Entra ID / Azure credential chain authentication.                                                                       | `AZURE_OPENAI_ENDPOINT`, `AZURE_OPENAI_API_VERSION` (optional), `AZURE_OPENAI_AUTH_TYPE` (optional, `"api_key"` or `"entra_id"`), `AZURE_OPENAI_API_KEY` (optional)                         |
```
dianed-square: Hi @bioinfornatics - thanks for updating! But can you please revert these documentation changes for now? Otherwise the docs will go out before this feature is available in a release.

We'll add them back in after the feature is released.

jmservier: Thanks for the clarification.
 I revert the documentation changes for now and keep the implementation as-is.

bioinfornatics: Thanks @dianed-square  for the clarification.
I revert the documentation changes for now and keep the implementation as-is.

crates/goose/src/providers/base.rs
```diff
@@ -175,6 +175,19 @@ impl ProviderMetadata {
 }
 
 /// Configuration key metadata for provider setup
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct AuthModeChoice {
+    /// Valeur stockée dans la configuration (ex: "api_key", "entra_id")
```
DOsinga: eh let's keep comments in english if needed at all

ui/desktop/src/components/settings/providers/modal/subcomponents/forms/DefaultProviderSetupForm.tsx
```diff
@@ -12,6 +13,17 @@ export interface ConfigInput {
   value?: string;
 }
 
+type AuthModeChoice = {
+  value: string;
+  label: string;
+  description?: string | null;
+  requires_api_key: boolean;
+};
```
DOsinga: this is a dupe, you should get this from the API

ui/desktop/src/components/settings/providers/modal/subcomponents/forms/DefaultProviderSetupForm.tsx
```diff
@@ -79,6 +112,99 @@ export default function DefaultProviderSetupForm({
...
+            }
+          }
+
+          return changed ? next : prev;
+        });
+      } catch {
+        // Ignore parse errors; user may still be typing an incomplete URL or has pasted invalid text.
+        // We avoid being intrusive in these cases.
+      }
+    }, 800); // ~0.8s debounce to allow human typing
```
DOsinga: using a timer for this sort of thing is not the right way to do this. also having three ways of getting this done doesn't make sense to me. pick one.

DOsinga:
@katzdave can you have a look at this since you looked at the one we did for openai itself? we should make sure to reuse as much code as possible. haven't looked too closely in the auth mode stuff, but left some preliminary comments

crates/goose/src/providers/base.rs
bioinfornatics: Thanks, sorry I let those french comments. 
The AuthModeChoice struct now has all its doc comments in English.

ui/desktop/src/components/settings/providers/modal/subcomponents/forms/DefaultProviderSetupForm.tsx
bioinfornatics: The duplicate frontend type for auth modes has been removed; now relies on the API-generated AuthModeChoice.

ui/desktop/src/components/settings/providers/modal/subcomponents/forms/DefaultProviderSetupForm.tsx
bioinfornatics: Thanks, that makes sense. I’ve removed the Azure‑specific timer‑based normalization effect from the generic provider setup form, so this component no longer contains provider‑specific behavior that runs implicitly in the background. This also simplifies the logic: there is now a single, explicit way to handle configuration, without extra paths that update values after a delay.

For now, Azure is the only provider exposing this “full endpoint URL with api-version in the query string” shape, so I didn’t see a clean way to generalize that behavior in the shared form. If we want to reintroduce this UX shortcut later, I’d prefer to do it in an Azure‑specific helper or component, or drive it from backend metadata, so the base form can stay generic and easier to reason about.

bioinfornatics:
I will be happy to get your review @kathawthorne ;-)
thanks for your insight

To me the most important thing is to support response api instead of the deprecated chat completion

kathawthorne:
were you referring to @katzdave?

bioinfornatics:
> were you referring to @katzdave?

Yes sorry 😅

crates/goose/src/providers/azure.rs
```diff
@@ -92,19 +152,19 @@ impl AzureProvider {
...
+    /// POST /openai/responses?api-version=...
+    ///
+    /// The deployment name is passed via the `model` field in the request body.
     async fn post(&self, payload: &Value) -> Result<Value, ProviderError> {
-        // Build the path for Azure OpenAI
-        let path = format!(
-            "openai/deployments/{}/chat/completions?api-version={}",
-            self.deployment_name, self.api_version
-        );
+        let path = format!("openai/responses?api-version={}", self.api_version);
```
katzdave, line 171: Do we want to move totally remove completions API, is responses fully backwards compatible with old versions, or should we switch on models here?

crates/goose-cli/src/commands/configure.rs
```diff
@@ -630,6 +636,116 @@ pub async fn configure_provider_dialog() -> anyhow::Result<bool> {
...
+                .unwrap_or(false);
+            if has_env_key || has_stored_key {
+                "api_key".to_string()
+            } else {
+                "entra_id".to_string()
+            }
+        }
+    };
+
+    let auth_type = cliclack::select("How would you like to authenticate with Azure OpenAI?")
```
katzdave, line 766: This full cli auth flow seems rather heavy compared to other providers. How well does the desktop case work this; just want to ensure that both paths are viable. Can potentially try to offload more logic into the provider.

crates/goose/src/providers/base.rs
```diff
@@ -175,6 +175,19 @@ impl ProviderMetadata {
 }
 
 /// Configuration key metadata for provider setup
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct AuthModeChoice {
```
katzdave: Is there any way to get this configuration to work without introducing this new abstraction. Could we just have a few different keys and tooltips/error messages on what needs to be configured together, because this makes a pretty big change to the provider flow in general, and is only used for this provider.

katzdave:
@bioinfornatics can you take another pass at testing. I fixed the conflicts + simplified things by removing the auth mode abstraction.

crates/goose/src/providers/azure.rs
bioinfornatics: from OpenAI
```
Incremental migration

The Responses API is a superset of the Chat Completions API. The Chat Completions API will also continue to be supported. As such, you can incrementally adopt the Responses API if desired. You can migrate user flows who would benefit from improved reasoning models to the Responses API while keeping other flows on the Chat Completions API until you're ready for a full migration.

As a best practice, we encourage all users to migrate to the Responses API to take advantage of the latest features and improvements from OpenAI.
```

source: https://platform.openai.com/docs/guides/migrate-to-responses

bioinfornatics:
> @bioinfornatics can you take another pass at testing. I fixed the conflicts + simplified things by removing the auth mode abstraction.

yes I will do, thanks @katzdave 

One thing come to my mind, how it is done with google gravity as same as azure you have also auth user flow

Generally speaking, using API keys in an agentic system is considered a poor security practice. API keys can be suitable for individual use cases, such as experimenting with new features or running quick tests. However, in a professional environment, it is essential to properly track user intent and identify which agent is performing each action in order to ensure accountability and secure access control.

katzdave:
> > @bioinfornatics can you take another pass at testing. I fixed the conflicts + simplified things by removing the auth mode abstraction.
> 
> yes I will do, thanks @katzdave
> 
> One thing come to my mind, how it is done with google gravity as same as azure you have also auth user flow
> 
> Generally speaking, using API keys in an agentic system is considered a poor security practice. API keys can be suitable for individual use cases, such as experimenting with new features or running quick tests. However, in a professional environment, it is essential to properly track user intent and identify which agent is performing each action in order to ensure accountability and secure access control.

I think if we can get this version tested on both CLI/desktop let's ship it.

I agree with you that there should be abstraction around multiple auth types, but maybe an easier flow would be some kind of array type like

auth_options: []

and then you can have some mutually exclusive form fields rather than enforcing specific rules around auth mode. Let's think about that for a follow up!

DOsinga:
I'm going to close this. would love to have gotten this in, so feel free to ping or reopen, but it's been inactive for too long

PR #6290, opened 2025-12-28
fix(code-execution): support bracket notation and multiple import styles

rabi:
## Summary
Use bracket notation for tool calls to support tool names that are not
valid JS identifiers (e.g. names with hyphens like 'search-flight'):
```
    - server["tool"]({ ... }) syntax for all tools
    - import * as server from "server" for imports
```
    
Also export a namespace object with the server's name. Models sometimes
use named imports (import { github }) when the module only exported
individual tool functions, causing 'could not find export' errors.
Adding the server name as a named export makes both styles work:
```
    - import * as github from "github"; github["tool"]({})
    - import { github } from "github"; github["tool"]({})
```

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Unit tested and tested with mcp using tool names with hyphen.

### DIFF
--- a/crates/goose/src/agents/code_execution_extension.rs
+++ b/crates/goose/src/agents/code_execution_extension.rs
@@ -257,8 +257,8 @@ impl ToolInfo {
             .join(", ");
         let desc = self.description.lines().next().unwrap_or("");
         format!(
-            "{}({{ {params} }}): {} - {desc}",
-            self.tool_name, self.return_type
+            "{}[\"{}\"]({{{params}}}): {} - {desc}",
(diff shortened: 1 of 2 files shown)


alexhancock (approved):

alexhancock:
agree, makes sense. merging. thanks for the contribution!

Issue #6295, opened 2025-12-29
download_cli.sh script fails when piped into bash


irbekrm:
**Describe the bug**

The `download_cli.sh` script fails at the step that asks users whether to add $GOOSE_BIN_DIR to PATH when the script is piped into bash as suggested by the [Goose install instructions](https://block.github.io/goose/docs/quickstart/).

This seems to be because the [`read` command](https://github.com/block/goose/blob/31eaa472099c920a4a1d4cb89e0eb8737535bc5b/download_cli.sh#L331) assumes that the stdin is terminal, but if the script is piped into bash, the stdin is the script itself, so the command does not wait for user's input.

---

**To Reproduce**
Run the install script against a new $GOOSE_BIN_DIR to ensure that the path where the user has to enter choice for whether add it to PATH is hit:
1. `mkdir -p /tmp/goose`
2. `curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh | GOOSE_BIN_DIR bash`
3. Follow the interactive steps
4. Observe that the script does not wait for you to enter choice for whether to add $GOOSE_BIN_DIR to PATH before the script fails at that step:
```
Warning: goose installed, but /tmp/goose is not in your PATH.

The $GOOSE_BIN_DIR is not in your PATH.
What would you like to do?
1) Add it for me
2) I'll add it myself, show instructions
Invalid choice. Please add $GOOSE_BIN_DIR to your PATH manually.
```

---

**Expected behavior**
`curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh | bash` succeeds when ran first time.

---

**Please provide the following information**
- **OS & Arch:** Ubuntu 24.04 x86
- **Interface:** download script
- **Version:** From latest main 31eaa472099c920a4a1d4cb89e0eb8737535bc5b
- **Extensions enabled:** N/A
- **Provider & Model:** N/A

irbekrm:
Should be fixed in https://github.com/block/goose/pull/6296

PR #6296, opened 2025-12-29
fix(install script): ensure script works when piped into bash

irbekrm:
## Summary

Fixes the `download_cli.sh` installation script to work correctly when piped into bash (e.g., curl ... | bash), such as suggested by the [Goose CLI install instructions](https://block.github.io/goose/docs/quickstart/). Previously, the read command would fail because stdin was the script itself rather than the terminal, preventing the script from waiting for user input when asking whether to add the goose bin directory to PATH.

- Adds conditional logic to detect if stdin is a terminal and reads from /dev/tty when it's not

The approach is similar to [rustup script](https://github.com/rust-lang/rustup/blob/86e0ac1453520bb8e4ba3b28b7ae4c1eba36b93c/rustup-init.sh#L177)

### Type of Change

- [ ] Feature
- [ x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

- [ x] This PR was created or reviewed with AI assistance

### Testing
I have manually tested that with this change the script succeeds when piped into bash (which was not working previously) as well as still works when ran directly on both Mac and Linux.
 
### Related Issues
Relates to #6295

Before:  
<details>
Screenshot from 2025-12-29 10-37-08

</details>

After:   
<details>
Screenshot from 2025-12-29 10-39-41

</details>

### DIFF
--- a/download_cli.sh
+++ b/download_cli.sh
@@ -328,7 +328,17 @@ if [[ ":$PATH:" != *":$GOOSE_BIN_DIR:"* ]]; then
       echo "1) Add it for me"
       echo "2) I'll add it myself, show instructions"
 
-      read -p "Enter choice [1/2]: " choice
+      # Check whether stdin is a terminal. If it is not (for example, if
+      # this script has been piped into bash), we need to explicitly read user's
+      # choice from /dev/tty.
+      if [ -t 0 ]; then # terminal
+        read -p "Enter choice [1/2]: " choice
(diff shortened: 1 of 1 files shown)


michaelneale (approved):
thanks!

PR #6309, opened 2025-12-31
fix(openai): make tool_call arguments optional and fix silent stream termination

rabi:
## Summary

Some LLM providers (e.g., vLLM with QwQ models) send tool_call chunks where the first chunk contains only the function name without arguments, and subsequent chunks contain the arguments. This caused deserialization to fail with 'missing field arguments'.

The deserialization error was silently treated as end-of-stream because 'while let Some(Ok(...))' doesn't match 'Some(Err(...))' - it just exits the loop as if the stream had completed normally.

Changes:
- Added #[serde(default)] to DeltaToolCallFunction.arguments in openai.rs to handle missing arguments by defaulting to empty string
- Fixed stream consumption pattern: changed 'while let Some(Ok(...))' to 'while let Some(result)' with explicit error propagation via '?'
- Added test for stream error propagation

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Unit and manual testing with vLLM.

### DIFF
--- a/crates/goose/src/agents/reply_parts.rs
+++ b/crates/goose/src/agents/reply_parts.rs
@@ -235,7 +235,9 @@ impl Agent {
         };
 
         Ok(Box::pin(try_stream! {
-            while let Some(Ok((mut message, usage))) = stream.next().await {
+            while let Some(result) = stream.next().await {
+                let (mut message, usage) = result?;
+
                 // Store the model information in the global store
                 if let Some(usage) = usage.as_ref() {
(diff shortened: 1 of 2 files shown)


jamadeo (approved):
Nice catch!

PR #6315, opened 2026-01-01
refactor: include hidden dotfiles folders in file picker search

### LINKED ISSUE #6211: @ file picker excludes hidden folders (dotfiles like .github) from search results
**Describe the bug**

The `@` file picker excludes hidden folders (those prefixed with `.`) from search results, making it impossible to reference important configuration files like GitHub workflows, VS Code settings, and other dotfiles.

When attempting to use `@` to reference a file in `.github/workflows/`, the file picker does not show these files even though they exist in the project and are visible in the file tree.

---

**To Reproduce**
1. Open a project that contains a `.github` folder with workflow files (e.g., `.github/workflows/deploy.yml`)
2. In the chat interface, type `@` to open the file picker
3. Search for the workflow file (e.g., type "deploy" or "workflow")
4. Observe that files within `.github` folder do not appear in the search results
5. Note that the file is visible in the folder tree on the right side panel in VS Code

---

**Expected behavior**

The `@` file picker should include files from hidden folders (dotfiles) in search results, especially since these folders often contain critical development configuration:
- `.github/workflows/` - CI/CD configurations
- `.vscode/` - Editor settings
- `.config/` - Application configs
- Other dotfiles tracked in git

Alternatively, there should be a toggle/setting to show/hide dotfiles in the file picker.

---

**Screenshots**

Image

---

- **OS & Arch:** macOS Tahoe 26.1 (25B78)
- **Interface:** UI
- **Version:** 1.17.0
- **Extensions enabled:** Auto Visualiser, Chatrecall, Code Execution, Context7, Developer, Extension Manager, Linear, Memory, Notion, Skills

- **Provider & Model:** anthropic/claude-sonnet-4.5 — OpenRouter

---

**Additional context**

It probably makes sense not to show all hidden files (although you could respect the OS' settings for that)

**Possible solutions:**
1. Include all git-tracked files regardless of hidden status (I'd probably suggest this one as this is more of a developer specific issue)
2. Add a setting to toggle "Show hidden files in @ picker"
3. Respect `.gitignore` but include tracked dotfiles (maybe you want to access build artifacts for some reason, so maybe don't opt for this)
4. Add a special indicator in the picker for hidden files but still show them

Abhijay007:
closes #6211 

### PR description: 

Refactor the file picker to include hidden configuration folders in search results. Previously, all files and folders starting with `.` were excluded, making it impossible to reference important configuration files like GitHub workflows, VS Code settings, and other dotfiles.

### Changes made

- **Added allowlist for hidden directories**: Created `allowedHiddenDirs` array containing important configuration directories:
  - `.github` - GitHub workflows and actions
  - `.vscode` - VS Code settings
 etc .....
 
- **Updated filtering logic**: Modified the file picker to check if hidden items are in the allowlist before skipping them, instead of blanket-excluding all dotfiles
- **Removed from skip list**: Removed `.vscode` and `.idea` from `skipDirs` since they're now explicitly allowed
- **Documentation update**: Updated file management guide to document the new hidden folder support

### Type of Change
- [x] Refactor / Code quality

### Testing

Tested in the desktop UI

Before:  

image

After:   

image

### DIFF
--- a/ui/desktop/src/components/MentionPopover.tsx
+++ b/ui/desktop/src/components/MentionPopover.tsx
@@ -170,7 +170,6 @@ const MentionPopover = forwardRef<
             '.Trash',
           ];
 
-          // directories should be included in search
           const allowedHiddenDirs = [
             '.github',
             '.vscode',


ui/desktop/src/components/MentionPopover.tsx
```diff
@@ -172,6 +170,18 @@ const MentionPopover = forwardRef<
             '.Trash',
           ];
 
+          // directories should be included in search
```
angiejones:
```suggestion

```

ui/desktop/src/components/MentionPopover.tsx
```diff
@@ -172,6 +170,18 @@ const MentionPopover = forwardRef<
...
+          const allowedHiddenDirs = [
```
angiejones, line 173: hmm, is there a different approach that will allow all hidden directories, not just a hardcoded list?

ui/desktop/src/components/MentionPopover.tsx
```diff
@@ -192,8 +202,13 @@ const MentionPopover = forwardRef<
...
-            // Skip hidden items and common ignore patterns
-            if (item.startsWith('.') || skipDirsAtDepth.includes(item)) {
+            // Skip items in the skip list
+            if (skipDirsAtDepth.includes(item)) {
+              continue;
+            }
+
+            // Skip hidden items except for allowed hidden directories
+            if (item.startsWith('.') && !allowedHiddenDirs.includes(item)) {
               continue;
```
angiejones, line 211: why skip?

ui/desktop/src/components/MentionPopover.tsx
Abhijay007: It's like we skipped some of the files, just like we did before. I am not sure if we should allow all of them

ui/desktop/src/components/MentionPopover.tsx
Abhijay007: just like before, we skipped some of the dotfiles and others for security purposes, like .git  and system, etc if we want, we can allow all

alexhancock: i think just include them all, as this is for when the user is selecting right?

Abhijay007: yea, when using file picker

blackgirlbytes (approved):

PR #6320, opened 2026-01-02
feat: add caching support for Anthropic models via Databricks

baxen:
Add cache_control fields to requests for Claude models to enable prompt caching:
- System message gets cache_control
- Last two user messages get cache_control (for incremental caching)
- Last tool definition gets cache_control (so all tools are cached as prefix)

This reduces costs when using Claude models via Databricks by caching frequently repeated content like system prompts and tool definitions.

### DIFF
--- a/crates/goose/src/providers/formats/databricks.rs
+++ b/crates/goose/src/providers/formats/databricks.rs
@@ -445,6 +445,92 @@ struct StreamingChunk {
     model: String,
 }
 
+/// Check if the model name indicates a Claude/Anthropic model that supports cache control.
+fn is_claude_model(model_name: &str) -> bool {
+    model_name.contains("claude")
+}
+
+/// Add Anthropic-style cache_control fields to the request payload for Claude models.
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/formats/databricks.rs
```diff
@@ -407,6 +445,92 @@ struct StreamingChunk {
...
+            if message.get("role") == Some(&json!("user")) {
+                if let Some(content) = message.get_mut("content") {
+                    if let Some(content_str) = content.as_str() {
+                        *content = json!([{
+                            "type": "text",
+                            "text": content_str,
+                            "cache_control": { "type": "ephemeral" }
+                        }]);
+                    } else if let Some(content_array) = content.as_array_mut() {
+                        // Content is already an array, add cache_control to the last element
```
michaelneale, line 480: probably can drop little comments like this (thanks claude opus!) but others I found helpful (assuming they are true!)

PR #6321, opened 2026-01-02
Have goose issue solver solve issues like a goose contributor

tlongwell-block:
Analyzed PRs from top goose contributors to push goose's code generation in a similar direction


baxen (approved):
These look great, nice and dense

PR #6322, opened 2026-01-02
feat: reasoning_content in API for reasoning models

### LINKED ISSUE #6192: [Feature] reasoning_content in API for reasoning models like DeepSeek
To better support reasoning models like DeepSeek, adding a [reasoning_content](https://api-docs.deepseek.com/guides/reasoning_model) parameter to the API would be great so that users can see the steps in the reasoning process.

[tool-calls](https://api-docs.deepseek.com/guides/thinking_mode#tool-calls)

from openai import OpenAI
client = OpenAI(api_key="", base_url="https://api.deepseek.com")

messages = [{"role": "user", "content": "9.11 and 9.8, which is greater?"}]
response = client.chat.completions.create(
    model="deepseek-reasoner",
    messages=messages
)

**+ reasoning_content = response.choices[0].message.reasoning_content**
content = response.choices[0].message.content

Abhijay007:
closes #6192 

## PR description 

This PR adds support for the `reasoning_content` field used by reasoning models like DeepSeek's reasoner, enabling goose to capture and display the step-by-step reasoning process that these models expose separately from their main response content.

### Type of Change
- [x] Feature

### AI Assistance
- [x] This PR was created or reviewed with AI assistance, used goose for approach and code 

### Testing

Tested on the desktop UI using the `DeepSeek-Reasoner` with sample “thinking” prompts.

deepLearning


taniacryptid:
Thank you for working on this and the other PRs! Tagging @block/goose-maintainers @block/goose-core-maintainers for each to have a look ❤️

crates/goose/src/context_mgmt/mod.rs
```diff
@@ -398,6 +398,7 @@ fn format_message_for_compacting(msg: &Message) -> String {
             MessageContent::SystemNotification(notification) => {
                 format!("system_notification: {}", notification.msg)
             }
+            MessageContent::Reasoning(reasoning) => format!("reasoning: {}", reasoning.text),
```
jamadeo: I wonder if we need to include these at all for compacting. I'm not sure it would improve the result. The model's reasoning->output content generation is already effectively "compacting"

crates/goose/src/conversation/message.rs
```diff
@@ -187,6 +189,7 @@ pub enum MessageContent {
     Thinking(ThinkingContent),
     RedactedThinking(RedactedThinkingContent),
     SystemNotification(SystemNotificationContent),
+    Reasoning(ReasoningContent),
```
jamadeo, line 197: we do have a Thinking message type, is that not really the same thing? I know OpenAI calls it reasoning, but we probably don't need to distinguish within the internal message type

zanesq:
@Abhijay007 checking if still in progress?

Abhijay007:
> @Abhijay007 checking if still in progress?

Will look into this soon, will share an update

michange:
Which reasoning/thinking models should this PR support?
What about ACP (= Agent Client Protocol) standard for client↔agent communication. Goose is adopting it. May it eventually cover event streaming ?

crates/goose/src/conversation/message.rs
alexhancock: Yes, I agree could be good to use something generic that normalizes over the same concept across providers.

ui/desktop/src/components/sessions/SessionViewComponents.tsx
```diff
@@ -128,6 +159,18 @@ export const SessionMessages: React.FC<SessionMessagesProps> = ({
                       </div>
 
                       <div className="flex flex-col w-full">
+                        {/* Reasoning content */}
+                        {reasoningContent && (
+                          <details className="bg-bgSubtle border border-borderSubtle rounded p-2 mb-2">
```
alexhancock: I personally would make two UX tweaks. Perhaps try with no border so it shows up more like goose's plain text responses, and then make the font smaller.

crates/goose/src/context_mgmt/mod.rs
alexhancock: Yeah I agree, I would remove it for the compaction case

DOsinga:
can we get this landed? @Abhijay007 - if you address @alexhancock 's last comments and match main, let's merge?

Abhijay007:
> can we get this landed? @Abhijay007 - if you address @alexhancock 's last comments and match main, let's merge?

Will look into this today will resolve it asap

PR #6324, opened 2026-01-04
fix: Update tetrate signup step

michaelneale:
Make it clearer for the signup options and makes it clear there is more to scroll down to: 

image

with free starter access and recommended badge as before


angiejones:
why are there 2 "Recommended" ribbons? it's a bit confusing

michaelneale:
@angiejones yeah we used to have that - basically one is BYO token option, other is signup and have it set up for you - ie "which path do you want" which I think makes sense? could have a different wording on each ribbon or something? (ie really - do you have a api key: go here, you don't: go there?) - something like that?

michaelneale:
@angiejones 

image

this should make it clearer

angiejones (approved):

baxen (approved):

codefromthecrypt:
thanks so much @michaelneale and @angiejones for thinking about coherency here!

PR #6325, opened 2026-01-04
feat: add auto submit for recipes that have been accepted

### LINKED ISSUE #6087: Add auto submit for recipes that have been accepted
Now that we have an accept recipe prompt in place I think we can turn back on auto submit recipe after running it if there is an initial prompt filled in the recipe.

Abhijay007:
closes #6087 

## PR description
This PR adds auto-submit functionality for recipes that have been accepted, allowing recipe prompts to be automatically submitted when a session is opened with a recipe but no existing messages.

### Type of Change
- [x] Feature

### AI Assistance
- [x] This PR was created or reviewed with AI assistance, used Copilot for the PR description, and Goose for the change 

### Testing
Tested in the desktop UI on Windows with recipes that have a prompt

Before:  
autoSubmitBefore

After:   

autoSubmitafter

### DIFF
--- a/ui/desktop/src/hooks/useAutoSubmit.ts
+++ b/ui/desktop/src/hooks/useAutoSubmit.ts
@@ -50,6 +50,11 @@ export function useAutoSubmit({
     );
   }, [sessionId]);
 
+  const hasUnfilledParameters = useCallback((session: Session) => {
+    const recipe = session.recipe;
+    return recipe?.parameters && recipe.parameters.length > 0 && !session.user_recipe_values;
+  }, []);
+
   // Auto-submit logic
   useEffect(() => {
     const currentSessionId = searchParams.get('resumeSessionId');
@@ -67,11 +72,7 @@ export function useAutoSubmit({
     // Scenario 1: New session with initial message from Hub
     // Hub always creates new sessions, so message_count will be 0
     if (initialMessage && session.message_count === 0 && messages.length === 0) {
-      const recipe = session.recipe;
-      const hasUnfilledParameters =
-        recipe?.parameters && recipe.parameters.length > 0 && !session.user_recipe_values;
-
-      if (!hasUnfilledParameters) {
+      if (!hasUnfilledParameters(session)) {
         hasAutoSubmittedRef.current = true;
         handleSubmit(initialMessage);
         clearInitialMessage();
@@ -92,11 +93,7 @@ export function useAutoSubmit({
 
     // Scenario 3: Resume with shouldStartAgent (continue existing conversation)
     if (shouldStartAgent) {
-      const recipe = session.recipe;
-      const hasUnfilledParameters =
-        recipe?.parameters && recipe.parameters.length > 0 && !session.user_recipe_values;
-
-      if (!hasUnfilledParameters) {
+      if (!hasUnfilledParameters(session)) {
         hasAutoSubmittedRef.current = true;
         handleSubmit({ msg: '', images: [] });
       }
@@ -111,6 +108,7 @@ export function useAutoSubmit({
     messages.length,
     chatState,
     clearInitialMessage,
+    hasUnfilledParameters,
   ]);
 
   return {


taniacryptid:
Thank you for working on this and the other PRs! Tagging @block/goose-maintainers @block/goose-core-maintainers for each to have a look ❤️

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -123,6 +123,14 @@ function BaseChatContent({
       .reverse();
   }, [messages]);
 
+  const recipe = session?.recipe;
+
+  const autoSubmit = useMemo(() => {
+    if (!session || initialMessage) return false;
+    const hasMessages = (session.conversation?.length ?? 0) > 0;
+    return !!(recipe?.prompt && !hasMessages && hasNotAcceptedRecipe === false);
```
The-Best-Codes: Would it be better to use `Boolean` here instead of `!!`?

Abhijay007: well both work the same way. I went with `!!` because it's the common pattern in typeScript and matches what we do elsewhere in the file

The-Best-Codes (approved):

ui/desktop/src/components/ChatInput.tsx
```diff
@@ -1001,6 +1011,18 @@ export default function ChatInput({
...
+  useEffect(() => {
+    performSubmitRef.current = performSubmit;
+  }, [performSubmit]);
+
+  useEffect(() => {
+    if (!!autoSubmit && !didAutoSubmit) {
+      setDidAutoSubmit(true);
+      performSubmitRef.current(displayValue);
+    }
+  }, [autoSubmit, didAutoSubmit, displayValue]);
```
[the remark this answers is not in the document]
Abhijay007: [9da4f72](https://github.com/block/goose/pull/6325/commits/9da4f720528a4fe2684e6944f5203306eaddad94)

zanesq:
@Abhijay007 appreciate you taking this on.. after looking at this more the current approach feels overcomplex and brittle. I think we might want to rewrite this rather than tacking on more logic. Do you have any ideas to make this better keeping in mind that auto submit from hub is also brittle? It's ok if we decide to punt on this for now and we can come back to it later when we do an upcoming rewrite.

Abhijay007:
> @Abhijay007 appreciate you taking this on.. after looking at this more the current approach feels overcomplex and brittle. I think we might want to rewrite this rather than tacking on more logic. Do you have any ideas to make this better keeping in mind that auto submit from hub is also brittle? It's ok if we decide to punt on this for now and we can come back to it later when we do an upcoming rewrite.

I will look into that and share how we can improve this

zanesq:
hi @Abhijay007 checking if you still want to work on this or should we close it?

Abhijay007:
Hi @zanesq I will push changes soon am just got occupied I bit I will try to resolve few of my PRs tomorrow

DOsinga:
also this @Abhijay007 - if it is too much everything, just let us know and we can find a hand-off appreciate all the work

Abhijay007:
> I don't think this is the right approach. we now have 4 scenario's for useAutoSubmit? and it has two extra parameters derived from the recipe state but then does an extra recipe substitution? we need to rethink this instead of making this more complicated

Hi @DOsinga I tried to simplify the auto-submit implementation based on your feedback. Instead of adding a 4th scenario with extra parameters, I made `RecipesView` pass the recipe prompt as initialMessage (just like Hub does), which automatically works with existing Scenario 1. 

also added a parameter check to Scenario 1 to prevent auto-submit before params are filled. This removes all the complexity that you mentioned, while still addressing all three issues: only auto-submits if recipe has a prompt, waits for parameters to be filled, and no flashing accept modal.

ui/desktop/src/components/recipes/RecipesView.tsx
```diff
@@ -156,6 +156,7 @@ export default function RecipesView() {
       setView('pair', {
         disableAnimation: true,
         resumeSessionId: session.id,
+        initialMessage: recipe.prompt ? { msg: recipe.prompt, images: [] } : undefined,
```
DOsinga, line 159: ??

ui/desktop/src/hooks/useAutoSubmit.ts
```diff
@@ -87,8 +92,15 @@ export function useAutoSubmit({
...
-      handleSubmit({ msg: '', images: [] });
+      const recipe = session.recipe;
+      const hasUnfilledParameters =
+        recipe?.parameters && recipe.parameters.length > 0 && !session.user_recipe_values;
+
+      if (!hasUnfilledParameters) {
+        hasAutoSubmittedRef.current = true;
+        handleSubmit({ msg: '', images: [] });
+      }
+      return;
```
DOsinga, line 100: this code is duplicated and I think it belongs in the caller

Abhijay007: yea, I will extract the parameter checking logic into a hasUnfilledParameters helper function to remove duplication between Scenario 1 and 3

ui/desktop/src/components/recipes/RecipesView.tsx
Abhijay007: Passing the recipe prompt as `initialMessage` so it auto-submits using the existing Scenario 1 logic (same pattern as Hub) for ref see `ui/desktop/src/components/Hub.tsx:63-67` where Hub does the same thing

DOsinga: yes, I understand. ?:undefined === ?? is what I meant

ui/desktop/src/hooks/useAutoSubmit.ts
Abhijay007: Wait, did you mean removing scenario 3?

DOsinga (approved):

Issue #5907, opened 2025-11-28
Goose CLI logs: total_tokens is always 0


agardnerIT:
**Describe the bug**

Goose CLI session logs always have the `total_tokens` field set to zero `0`.

---

**To Reproduce**
1. Start a new session: `goose session`
2. Ask goose and the LLM to create a Python file to do something
3. Exit the session: `exit`

The logs show the session ending but `total_tokens: 0`

```
$ goose session
starting session | provider: github_copilot model: gpt-4.1
    session id: 20251128_5
    working directory: /workspaces/ai-sre-workflow-demo

goose is running! Enter your instructions, or try asking what goose can do.

Context: ○○○○○○○○○○ 0% (0/1000000 tokens)
( O)> create a python file in /tmp/foo.py which prints the numbers 1, 2 and 3

─── text_editor | developer ──────────────────────────
path: /tmp/foo.py
command: write
file_text: print(1)
print(2)
print(3)


# Info Received

- **Datetime**: 2025-11-28 04:01:16

---

# Action Taken

- Created a simple Python script (`/tmp/foo.py`) with example print statements for testing.

---

# Summary Table

| Component        | Details                                 |
|------------------|-----------------------------------------|
| Info Message     | Datetime: 2025-11-28 04:01:16           |
| Script Creation  | /tmp/foo.py with `print(1)\nprint(2)\nprint(3)` |

If you have further instructions or a specific task you want to work on, please let me know!

⏱️  Elapsed time: 5.04s
Context: ○○○○○○○○○○ 0% (0/1000000 tokens)
( O)> exit
Closing session. Session ID: 20251128_5
```

Looking at the logs:

```
{"timestamp":"2025-11-28T04:01:20.810082Z","level":"INFO","fields":{"message":"Session completed","counter.goose.session_completions":1,"session_type":"new","exit_type":"normal","duration_ms":23707,"total_tokens":0,"message_count":4},"target":"goose_cli::cli"}
{"timestamp":"2025-11-28T04:01:20.810236Z","level":"INFO","fields":{"message":"Session duration","counter.goose.session_duration_ms":23707,"session_type":"new"},"target":"goose_cli::cli"}
```

## Goose Version

```
1.15.0
```

morpheus9393:
Same behaviour here.

marcindanisz:
same here for openrouter's models

DOsinga:
@lifeizhou-ap did you assign yourself to this? by default I'd think @katzdave could have a look

[...]

PR #6331, opened 2026-01-05
make sure the models are using streaming properly

lifeizhou-ap:
## Summary
- Root cause: usage is not set in the response data with the streaming models.
- Fix:  Introduced the stream function to be consistent to the other providers. Deleted unused file.

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
- Manual testing
- Smoke Test

### Related Issues
Relates to #ISSUE_ID: https://github.com/block/goose/issues/5907 
Discussion: LINK (if any)

### DIFF
--- a/crates/goose/src/providers/githubcopilot.rs
+++ b/crates/goose/src/providers/githubcopilot.rs
@@ -480,7 +480,6 @@ impl Provider for GithubCopilotProvider {
         stream_openai_compat(response, log)
     }
 
-    /// Fetch supported models from GitHub Copliot; returns Err on failure, Ok(None) if not present
     async fn fetch_supported_models(&self) -> Result<Option<Vec<String>>, ProviderError> {
         let (endpoint, token) = self.get_api_info().await?;
         let url = format!("{}/models", endpoint);


crates/goose/src/providers/githubcopilot.rs
```diff
@@ -489,6 +496,49 @@ impl Provider for GithubCopilotProvider {
...
+                let (endpoint, token) = self.get_api_info().await?;
+                let url = url::Url::parse(&format!("{}/chat/completions", endpoint)).map_err(|e| ProviderError::RequestFailed(format!("Invalid base URL: {e}")))?;
+
+                let headers = self.get_github_headers();
+
+                let mut request = self
+                    .client
+                    .post(url)
+                    .headers(headers)
+                    .header("Authorization", format!("Bearer {}", token));
```
DOsinga: is this right? we have authorization types for the client, we shouldn't be setting this by hand I would think

PR #6337, opened 2026-01-05
Skip the smoke tests for dependabot PRs

jamadeo:
As documented [here](https://docs.github.com/en/code-security/dependabot/troubleshooting-dependabot/troubleshooting-dependabot-on-github-actions#troubleshooting-failures-when-dependabot-triggers-existing-workflows), dependabot PRs are like fork PRs in that they don't have access to secrets. This means the smoke tests can't run on them.


jamadeo:
Running on PRs with secrets access makes the API tokens easily steal-able. If we can find a way around that we could make it happen

PR #6341, opened 2026-01-05
Improve PR preview site artifact clean-up

jamadeo:
The previous logic only cleaned up pr-previews which left behind "empty" directories (containing only a hidden .jekyll-ignore file), which seemed to be the case for some but not all of the since-removed preview directories.

This improves it so that we always remove everything under pr-previews/ that isn't a currently-deployed preview directory.

### DIFF
--- a/scripts/clean-gh-pages.sh
+++ b/scripts/clean-gh-pages.sh
@@ -1,4 +1,3 @@
-
 #!/bin/bash
 
 set -e  # Exit on error


jamadeo:
I think we can actually get rid of this script/step entirely if we set force_orphan on the gh-pages deployment as done in #6340

jamadeo:
closing in favor of https://github.com/block/goose/pull/6340

PR #6267, opened 2025-12-24
fix(code-execution): auto-wrap multi-line function calls, serialize results as JSON

rabi:
## Summary

The code execution extension's result wrapping logic only looked at the last line of code, not the complete multi-line statement. When code ended with continuation characters like `});`, the wrapping produced invalid JavaScript syntax like `__result__ = }))`, causing parse errors and infinite retry loops.

 - Extract wrap_code_for_result() to handle multiline expressions ending with });
 - Detect continuation lines (closing brackets, method chaining)
 - Find expression start for proper wrapping instead of wrapping interior lines
 - Serialize results using to_json() instead of display() to show actual array/object contents
 - Update LLM instructions for clearer code formatting guidance

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

- [x] This PR was created or reviewed with AI assistance

### Testing
Unit tests and manual testing.

### Related Issues

Closes: #6266

PR #6343, opened 2026-01-05
fix: clean up result recording for code mode

alexhancock:
Improves the quality of how we have the rust part of the code execution extension collect results from JS execution

Per this comment initially https://github.com/block/goose/pull/6030#discussion_r2620518841 I should have gotten to it on initial implementation

This should avoid the need for improvements and tweaks to the `__result__` handling https://github.com/block/goose/pull/6267

Demo:

Screenshot 2026-01-05 at 4 50 43 PM

### DIFF
--- a/crates/goose/src/agents/code_execution_extension.rs
+++ b/crates/goose/src/agents/code_execution_extension.rs
@@ -5,7 +5,6 @@ use anyhow::Result;
 use async_trait::async_trait;
 use boa_engine::builtins::promise::PromiseState;
 use boa_engine::module::{MapModuleLoader, Module, SyntheticModuleInitializer};
-use boa_engine::property::Attribute;
 use boa_engine::{js_string, Context, JsNativeError, JsString, JsValue, NativeFunction, Source};
 use indoc::indoc;
 use regex::Regex;
@@ -266,6 +265,8 @@ impl ToolInfo {
 thread_local! {
     static CALL_TX: std::cell::RefCell<Option<mpsc::UnboundedSender<ToolCallRequest>>> =
         const { std::cell::RefCell::new(None) };
+    static RESULT_CELL: std::cell::RefCell<Option<String>> =
+        const { std::cell::RefCell::new(None) };
 }
 
 fn create_server_module(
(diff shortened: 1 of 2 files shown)


crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -766,15 +740,14 @@ impl McpClientTrait for CodeExecutionClient {
...
                         const readme = text_editor({ path: "./README.md", command: "view" });
                         const status = shell({ command: "git status" });
-                        { files, readme, status }
+                        record_result({ files, readme, status });
                         ```
 
                         SYNTAX:
                         - Import: import { tool1, tool2 } from "serverName";
                         - Call: toolName({ param1: value, param2: value })
+                        - Result: record_result(value) - call this to return a value from the script
```
rabi, line 749: Yeah , looks simpler and better. However, I've seen at times smaller models ignore these type of instructions.

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -360,19 +361,22 @@ fn run_js_module(
...
 
-    ctx.register_global_property(
-        js_string!("__result__"),
-        JsValue::undefined(),
-        Attribute::WRITABLE,
-    )
-    .map_err(|e| format!("Failed to register __result__: {e}"))?;
+    let record_result = NativeFunction::from_copy_closure(|_this, args, _ctx| {
+        let value = args.first().cloned().unwrap_or(JsValue::undefined());
+        RESULT_CELL.with(|cell| *cell.borrow_mut() = Some(value.display().to_string()));
```
rabi, line 374: value.display() doesn't properly serialize arrays/objects as JSON. Maybe I'll propose a followup later.

crates/goose/src/agents/code_execution_extension.rs
DOsinga: we do something similar for the main agent for the final result (in the context of recipes). there we check whether the call was actually made and if not, just remind the model to do it if it doesn't. we could add that here too - have the model generate the code, run the code, see that it doesn't call record, stick the js back into the model saying, you said this, but you didn't call record.

DOsinga: obviously for a follow up

DOsinga (approved):

crates/goose/src/agents/code_execution_extension.rs
rabi: proposed https://github.com/block/goose/pull/6495

PR #6349, opened 2026-01-06
feat: http proxy support from shanto12

michaelneale:
this is ported from https://github.com/block/goose/pull/6072 thanks to @shanto12


ui/desktop/src/main.ts
```diff
@@ -132,6 +132,40 @@ async function ensureTempDirExists(): Promise<string> {
...
+      // Try to resolve system proxy settings
+      const systemProxy = await session.defaultSession.resolveProxy('https://example.com');
+      if (systemProxy && systemProxy !== 'DIRECT') {
+        console.log('[Main] Using system proxy:', systemProxy);
+      } else {
+        console.log('[Main] No proxy configured');
+      }
+    }
+  } catch (error) {
+    console.error('[Main] Failed to configure proxy:', error);
```
DOsinga: I think we might just want to crash here - the user configured a proxy and we just print, eh, didn't work. we should be louder about that

ui/desktop/src/main.ts
```diff
@@ -1797,6 +1831,9 @@ const focusWindow = () => {
 };
 
 async function appMain() {
+  // Configure HTTP proxy settings before initializing the app
```
DOsinga: delete the comment

ui/desktop/src/main.ts
```diff
@@ -132,6 +132,40 @@ async function ensureTempDirExists(): Promise<string> {
...
+      });
+      
+      console.log('[Main] Proxy configured successfully');
+    } else {
+      // Try to resolve system proxy settings
+      const systemProxy = await session.defaultSession.resolveProxy('https://example.com');
+      if (systemProxy && systemProxy !== 'DIRECT') {
+        console.log('[Main] Using system proxy:', systemProxy);
+      } else {
+        console.log('[Main] No proxy configured');
```
DOsinga: this is pretty much the default, so no need to log this.

ui/desktop/src/main.ts
```diff
@@ -132,6 +132,40 @@ async function ensureTempDirExists(): Promise<string> {
...
+      console.log('[Main] Configuring proxy:', proxyUrl);
+      
+      // Set proxy configuration for the default session
+      await session.defaultSession.setProxy({
+        proxyRules: proxyUrl,
+        proxyBypassRules: noProxy
+      });
+      
+      console.log('[Main] Proxy configured successfully');
+    } else {
```
DOsinga: this else clause is prety much a no-op, no? it checks whether there is a system proxy and if so logs that it does. I would drop the entire bit.

God-damnit-all:
> * Implements system proxy detection as fallback when environment variables are not set

It seems the fallback also applies if it can't connect to the proxy for some reason, which isn't very desirable behavior.

PR #6350, opened 2026-01-06
fix: we load extensions when agent starts so don't do it up front

michaelneale:
Previously it would load the extensions on GUI start - and then when you start interacting, it would load them any way, this removes that un-necessary initial load.

### DIFF
--- a/ui/desktop/src/App.tsx
+++ b/ui/desktop/src/App.tsx
@@ -51,10 +51,10 @@ function PageViewTracker() {
 }
 
 // Route Components
-const HubRouteWrapper = ({ isExtensionsLoading }: { isExtensionsLoading: boolean }) => {
+const HubRouteWrapper = () => {
   const setView = useNavigation();
 
-  return <Hub setView={setView} isExtensionsLoading={isExtensionsLoading} />;
+  return <Hub setView={setView} />;
 };
 
 const PairRouteWrapper = ({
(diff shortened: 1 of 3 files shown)


ui/desktop/src/App.tsx
```diff
@@ -398,27 +396,8 @@ export function AppInner() {
...
-            // the onboarding flow will trigger
-          } else {
-            throw e;
-          }
-        }
-      })();
-    }
-  }, [loadCurrentChat, setAgentWaitingMessage, navigate, loadingHub, setChat]);
+  // Don't pre-load session/extensions on Hub - wait until user actually starts chatting
+  // This avoids the slow extension loading on app startup
```
DOsinga: this is commenting code that doesn't exist?

ui/desktop/src/App.tsx
```diff
@@ -363,15 +363,14 @@ const ExtensionsRoute = () => {
 
 export function AppInner() {
   const [fatalError, setFatalError] = useState<string | null>(null);
-  const [agentWaitingMessage, setAgentWaitingMessage] = useState<string | null>(null);
+  const [agentWaitingMessage] = useState<string | null>(null);
```
DOsinga: what's the idea here? I think you want to rip this and the one below out and any dependent code

michaelneale:
@DOsinga yeah this was a late night hack, I wasn't even sure I was reading things right, can tidy this up for sure.

ui/desktop/src/App.tsx
michaelneale: yeah I don't think it is needed now... probably

ui/desktop/src/App.tsx
michaelneale: hehe - those comments are literally from my prompt... good job goose!

michaelneale:
thanks @DOsinga @tlongwell-block there was even more I could delete!

PR #6352, opened 2026-01-06
fix(code_execution): handle model quirks with tool calls

rabi:
## Summary
- Auto-prefix unprefixed tool names (read_module -> code_execution__read_module)
- Parse stringified JSON arrays in search_modules terms parameter
- Add note that all tool calls are synchronous (no async/await)
- Improve search_modules documentation to clarify array parameter format

Fixes issues where some models (e.g. Nemotron) strip tool prefixes or stringify array parameters.

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested with vLLM and  Nemotron Hybrid MoE Model (nemotron-3-nano-30b-a3b)

### DIFF
--- a/crates/goose/src/agents/code_execution_extension.rs
+++ b/crates/goose/src/agents/code_execution_extension.rs
@@ -466,6 +466,8 @@ impl CodeExecutionClient {
                 - WRONG: Multiple execute_code calls, each with one tool
                 - RIGHT: One execute_code call with a script that calls all needed tools
 
+                IMPORTANT: All tool calls are SYNCHRONOUS. Do NOT use async/await.
+
                 Workflow:
                     1. Use the read_module tool to discover tools and signatures
                     2. Write ONE script that imports and calls ALL tools needed for the task
@@ -572,12 +574,16 @@ impl CodeExecutionClient {
             .and_then(|a| a.get("terms"))
             .ok_or("Missing required parameter: terms")?;
 
-        let terms_vec = if let Some(s) = terms.as_str() {
-            vec![s.to_string()]
-        } else if let Some(arr) = terms.as_array() {
+        let terms_vec = if let Some(arr) = terms.as_array() {
             arr.iter()
                 .filter_map(|v| v.as_str().map(String::from))
                 .collect()
+        } else if let Some(s) = terms.as_str() {
+            if s.starts_with('[') && s.ends_with(']') {
+                serde_json::from_str::<Vec<String>>(s).unwrap_or_else(|_| vec![s.to_string()])
+            } else {
+                vec![s.to_string()]
+            }
         } else {
             return Err("Parameter 'terms' must be a string or array of strings".to_string());
         };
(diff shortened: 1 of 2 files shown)


crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -466,6 +466,8 @@ impl CodeExecutionClient {
                 - WRONG: Multiple execute_code calls, each with one tool
                 - RIGHT: One execute_code call with a script that calls all needed tools
 
+                IMPORTANT: All tool calls are SYNCHRONOUS. Do NOT use async/await.
```
michaelneale, line 469: hrm - I guess this is true - are we sure? 
Also, I wonder if there is some other way to enforce this than prompting, but I think this is is ok @alexhancock

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -830,9 +836,11 @@ impl McpClientTrait for CodeExecutionClient {
...
 
                         USAGE:
-                        - Single term: search_modules with terms="file"
-                        - Multiple terms: search_modules with terms=["git", "shell"]
-                        - Regex patterns: search_modules with terms="sh.*", regex=true
+                        - Single term: terms="github" (just a plain string)
+                        - Multiple terms: terms=["git", "shell"] (a JSON array, NOT a string)
+                        - Regex patterns: terms="sh.*", regex=true
+
+                        IMPORTANT: Do NOT stringify arrays. Use terms=["a","b"] not terms="[\"a\",\"b\"]"
```
michaelneale, line 843: have found saying NOT can use up attention and hurt performance, wonder if we need that so explicitly (if there is a way to feedback errors then that can help)

crates/goose/src/agents/extension_manager.rs
```diff
@@ -1022,17 +1022,30 @@ impl ExtensionManager {
         tool_call: CallToolRequestParam,
         cancellation_token: CancellationToken,
     ) -> Result<ToolCallResult> {
+        // Some models strip the tool prefix, so auto-add it for known code_execution tools
```
michaelneale, line 1025: is this special casing for code execution as it needs to prefix it, but in other cases - they work?

michaelneale:
thanks @rabi - cc @alexhancock shoudl we do some benchmarks/checking with frontier models to make sure this is no worse (I am worryed about the use of "not" in the prompts as it consumes extra attention heads).

crates/goose/src/agents/extension_manager.rs
rabi: I think `code_execution` tools (execute_code, read_module, search_modules) are the only tools the model calls directly when `code_execution` extension is enabled.  Model is expected to call the tools with the extension prefix, but I suspect when all visible tools share the same prefix and the instruction is to use <tool name> without prefix, some models strip them.

crates/goose/src/agents/code_execution_extension.rs
alexhancock: I'm confused by this part of the code @rabi . Isn't the builtin function we provide async, and therefor don't any calls the model writes to functions which call a tool need to be `await`ed?

rabi: It seems tools are exposed to the model as synchronous JavaScript functions via the Boa JS engine. `create_tool_function` wraps each tool using `blocking_recv()` which blocks until the async Rust handler responds. So, model-generated code I think should use `const x = tool({...})` not `await tool({...})`.

alexhancock: Makes sense @rabi. I had misremembered. Merged

PR #6353, opened 2026-01-06
fix: lazy extension load

michaelneale:
this is an alternative to: https://github.com/block/goose/pull/6350

this attempts to only load extensions when they are needed (as in launch)


DOsinga:
Isn't this just going to make the response slower by however long it takes to load the extension?

PR #6356, opened 2026-01-06
Bump rust toolchain to 1.92 (current stable)

jamadeo:


alexhancock (approved):

crates/goose/src/providers/formats/databricks.rs
```diff
@@ -370,43 +370,6 @@ pub fn response_to_message(response: &Value) -> anyhow::Result<Message> {
     ))
 }
 
-#[derive(Serialize, Deserialize, Debug)]
```
alexhancock, line 373: were these just unused?

jamadeo: Yeah seems so, and I'm not sure why it wasn't flagged by the build before, but I'm guessing some improvements were made between 1.88 and 1.92 to finding dead code

PR #6363, opened 2026-01-06
feat(goose-acp): enable parallel sessions with isolated agent state

codefromthecrypt:
## Summary

This PR makes it possible to run ACP agents and sessions in parallel, with critical state isolated to each agent instance.

Before, we were using global singletons for most state, which made it impossible to change configuration without ENV or subprocess isolation. Moreover, unlike platform, built-in extensions launched subprocesses which implied you need a binary entrypoint to run them.

This PR extracts the most critical state into structs and ensures consistent threading to call sites. It also reconfigures built-in extensions to run in-process, via duplex streams.

Legacy entrypoints retain global singletons despite ACP entrypoints being isolated. This was done to allow progress, particularly to implement goose-server with ACP. This is also due to the pending PR to replace the CLI. We don't want to do too much work for code that is going to be replaced.

The most important impact is the new goose-acp crate which contains isolated code for critical paths even though the CLI entrypoint that uses it will act the same.

Later in the tests section you will see how the combination of these changes allow more ACP integration tests than before to complete in less than a second.

### Type of Change
- [x] Feature
- [x] Refactor / Code quality
- [x] Tests

### AI Assistance

### Testing

- Now that built-in are in-process like platform extensions, we don't need to build binaries to run ACP tests.
- Since we parameterized the `GooseAcpAgent` to accept variables via `AgentConfig`, tests don't rely on ENV variables.

Above and other changes allow us to safely run tests in parallel, removing env-lock, serial and reset patterns.
This means despite having more ACP integration tests, our performance is incredibly faster.

| Crate     | MAIN   | THIS PR | Speedup  |
|-----------|--------|---------|----------|
| goose-acp | 42.22s | 0.77s   | **55x**  |

### Works cited

What this PR does is a step towards a more modular codebase which avoids global singletons.
Some projects I studied for reference were:
- [hyperswitch](https://github.com/juspay/hyperswitch) - AppState takes storage config
- [lemmy](https://github.com/LemmyNet/lemmy) - LemmyContext created once, passed to handlers
- [typst](https://github.com/typst/typst) - Engine requires all dependencies at construction

### Related Issues
- #6305


crates/goose/tests/mcp_replays/github-mcp-serverstdio
```diff
@@ -7,6 +7,6 @@ STDERR: time=2025-12-11T17:58:47.640-05:00 level=INFO msg="server session connec
...
 STDIN: {"jsonrpc":"2.0","method":"notifications/initialized"}
 STDERR: time=2025-12-11T17:58:47.642-05:00 level=INFO msg="session initialized"
-STDIN: {"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"_meta":{"progressToken":0},"name":"get_file_contents","arguments":{"owner":"block","path":"README.md","repo":"goose","sha":"ab62b863c1666232a67048b6c4e10007a2a5b83c"}}}
+STDIN: {"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"_meta":{"goose-session-id":"test-session-id","progressToken":0},"name":"get_file_contents","arguments":{"owner":"block","path":"README.md","repo":"goose","sha":"ab62b863c1666232a67048b6c4e10007a2a5b83c"}}}
```
codefromthecrypt, line 10: now that session is propagated via parameter in the case we always know it, the expectation file bumped for good reason.

crates/goose/tests/mcp_integration_test.rs
```diff
@@ -60,6 +75,44 @@ impl Provider for MockProvider {
     }
 }
 
+fn build_and_get_binary_path() -> PathBuf {
```
codefromthecrypt, line 78: now that builtin extensions no longer require being launched as subprocesses (`mcp xxx` args), ACP doesn't need a build to run tests. This leaves only one path where we launch processes in tests, the old one for MCP replay, only used when recording. This is put back to exactly the same code from before we needed to extract a common cargo build utility.

crates/goose/src/scheduler_trait.rs
```diff
@@ -39,3 +40,104 @@ pub trait SchedulerTrait: Send + Sync {
         sched_id: &str,
     ) -> Result<Option<(String, DateTime<Utc>)>, SchedulerError>;
 }
+
+const UNAVAILABLE_MESSAGE: &str = "Scheduler not available. This tool only works in server mode.";
```
codefromthecrypt, line 44: before we had a single call site we noticed scheduler in use and disabled it for CLI sessions. This extracts so it isn't optional. Unless there's a good reason to have CLI sessions fail on scheduler, I recommend in this PR or the next just letting them work and removing this workaround and/or only using it in tests

crates/goose/src/session/session_manager.rs
```diff
@@ -61,7 +60,8 @@ impl std::str::FromStr for SessionType {
     }
 }
 
-static SESSION_STORAGE: OnceCell<Arc<SessionStorage>> = OnceCell::const_new();
+static SESSION_STORAGE: LazyLock<Arc<SessionStorage>> =
+    LazyLock::new(|| Arc::new(SessionStorage::new(Paths::data_dir())));
```
codefromthecrypt, line 64: in order to propagate and also still have a singleton I looked around at about 10 rust projects and this lazy lock pattern seemed the best, as old sites still get a global singleton until/unless updated, and new can propagate through in all cases the actual connect stuff is deferred to first.

crates/goose/src/permission/permission_confirmation.rs
```diff
@@ -7,6 +7,7 @@ pub enum Permission {
     AllowOnce,
     Cancel,
     DenyOnce,
+    AlwaysDeny,
```
codefromthecrypt, line 10: we were missing this even though our permissions system allows persistence of deny. This is needed for ACP and the motivation was testing permissions which had a bug and they now are testable without clobbering or ENV

crates/goose-acp/tests/server_test.rs
```diff
@@ -0,0 +1,400 @@
...
+
+                test_fn(cx.clone(), session.session_id, updates).await;
+                Ok(())
+            }
+        })
+        .await
+        .unwrap();
+}
+
+#[test_case(Some(PermissionOptionKind::AllowAlways), ToolCallStatus::Completed, "user:\n  always_allow:\n  - lookup__get_code\n  ask_before: []\n  never_allow: []\n"; "allow_always")]
```
codefromthecrypt, line 342: this is the new test and notice no ACP test requires serial anymore, and none use ENV or clobber eachother's database or permissions files

crates/goose/src/config/permission.rs
```diff
@@ -3,8 +3,14 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::fs;
 use std::path::{Path, PathBuf};
+use std::sync::{Arc, LazyLock, RwLock};
 use utoipa::ToSchema;
 
+const PERMISSION_FILE: &str = "permission.yaml";
+
+static PERMISSION_MANAGER: LazyLock<Arc<PermissionManager>> =
```
codefromthecrypt, line 11: permission manager needs to be a singleton where used, so I used exact same pattern as session storage. before depending on call site reads might come from memory which could have problems when not shared. now it is always shared and pushed down consistently

codefromthecrypt:
can't change the head branch without a new PR so re-opened here so it is a branch off this repo instead of my fork  https://github.com/block/goose/pull/6392

PR #6372, opened 2026-01-07
fix: update MCP Apps _meta.ui.resourceUri to use nested format (SEP-1865)

### LINKED ISSUE #6364: metadata format mismatch for MCP Apps stops UI from rendering
In implementing the newest release of goose (1.19.0) that includes MCP UI Apps Support - I ran into an issue where the draft spec is looking for `_meta.ui.resourceUri` but goose is expecting `_meta['ui/resourceUri']` and this makes it so the UI does not render

I used claude to figure out this out, but did find that reference in the draft spec SEP-1865 so I believe it to be accurate and fixing this in my implementation worked to allow rendering of the UI.

---

**To Reproduce**
I am not able to provide an environment to reporduce as I'm working on a local version of the apollo mcp server that i cannot push to share.  However, I am happy to show you in Discord if needed what I did (handle is AmandaM).  The Apollo MCP server is built in Rust and the edits I made to it are based on the draft spec.

tl;dr this blocked rendering of the UI for me

---

**Expected behavior**
With MCP-UI apps implemented in my MCP server, I expect UI to render

---

**Screenshots**

---

- **OS & Arch:** [macos sequoia]
- **Interface:** [UI]
- **Version:** [1.19.0]
- **Extensions enabled:** [e.g. Developer, Todo, my localhost http server]
- **Provider & Model:** [e.g. Anthropic - Claude SOnnet 4.5]

---

**Additional context**

aharvard:
## Summary

Fixes #6364

The draft spec SEP-1865 expects `_meta.ui.resourceUri` (nested object structure) but goose was expecting `_meta['ui/resourceUri']` (flat key with slash), which prevented MCP Apps UI from rendering.

## Changes

Updated `ui/desktop/src/components/ToolCallWithResponse.tsx` to use the correct nested format:

- Changed `UiMeta` type from flat key to nested structure
- Updated `maybeRenderMCPApp` function to access `_meta?.ui?.resourceUri` instead of `_meta?.['ui/resourceUri']`

## Testing

- TypeScript compiles without errors
- ESLint passes

### DIFF
--- a/ui/desktop/src/components/ToolCallWithResponse.tsx
+++ b/ui/desktop/src/components/ToolCallWithResponse.tsx
@@ -24,19 +24,21 @@ interface ToolGraphNode {
   depends_on: number[];
 }
 
+type UiMeta = {
+  ui?: {
+    resourceUri?: string;
+  };
+};
+
 type ToolResultWithMeta = {
   status?: string;
   value?: CallToolResponse & {
-    _meta?: {
-      'ui/resourceUri'?: string;
-    };
+    _meta?: UiMeta;
   };
 };
 
 type ToolRequestWithMeta = ToolRequestMessageContent & {
-  _meta?: {
-    'ui/resourceUri'?: string;
-  };
+  _meta?: UiMeta;
   toolCall: {
     status: 'success';
     value: {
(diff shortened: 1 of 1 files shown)


DOsinga:
how come this worked with the mcp servers we tested before?

aharvard:
> how come this worked with the mcp servers we tested before?

Because both the servers and goose were based on an earlier iteration of the draft spec.

Refer to deprecation note here https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx#resource-discovery
image

DOsinga:
should we support both for now?

aharvard:
> should we support both for now?

no, I don't think we need to.

PR #6374, opened 2026-01-07
fix: prevent infinite loop of tool-input notifications in MCP Apps

### LINKED ISSUE #6373: MCP Apps infinite loop during render
goose appears to send an infinite loop of `ui/notifications/tool-input` after tool result - appeared to cause re-rendering issues although this was inconsistent. Adding debugging and restarting the MCP server allowed my UI to render as expected after discovering this issue.

example log (not the full output as it is long)

```
🔍 Debug Log:
8:43:33 AM: 🚀 Initializing MCP App
8:43:39 AM: 📥 Received tool-input
{
  "calendarId": "cal-OCfFpty2kvE6LLB",
  "limit": 5
}
8:43:39 AM: ✅ Received tool-result
{
  "content": [
    {
      "type": "text",
      "text": "{\"result\":{\"data\":{\"eventAnalytics\":{\"totalEvents\":5,\"totalAttendees\":226,\"averageAttendeesPerEvent\":45.2,\"checkInRate\":0.1504424778761062,\"events\":[{\"id\":\"evt-lai9A77GhlAUyqR\",\"name\":\"Apollo and AWS in SF\",\"date\":\"2025-03-06T01:00:00.000Z\",\"attendeeCount\":26,\"checkedInCount\":0,\"location\":\"San Francisco, California\"},{\"id\":\"evt-sEkpMqp0F04rRV5\",\"name\":\"MCP Server Builder Series\",\"date\":\"202
8:43:39 AM: 🎯 Processing tool result
8:43:39 AM: 🎨 Rendering dashboard
{}
8:43:39 AM: 📥 Received tool-input
{
  "calendarId": "cal-OCfFpty2kvE6LLB",
  "limit": 5
}
8:43:39 AM: 📥 Received tool-input
{
  "calendarId": "cal-OCfFpty2kvE6LLB",
  "limit": 5
}
8:43:39 AM: 📥 Received tool-input
{
  "calendarId": "cal-OCfFpty2kvE6LLB",
  "limit": 5
}
```
---

**To Reproduce**
1. Add debug logs to your html for an MCP app iframe
2. Execute any tool 
3. observe console results in the iframe

---

**Expected behavior**
UI updates as expected with tool results and no re-rendering, no additional tool-input notifications are sent from goose

---

**Screenshots**
Example debug code added:
Image

Example debug output in iframe (working version)

Image

---

- **OS & Arch:** [macos seqouia]
- **Interface:** [UI]
- **Version:** [v1.19.0]
- **Extensions enabled:** [Developer, ToDo, local http mcp server]
- **Provider & Model:** [sonnet 4.5]

---

aharvard:
## Summary

Fixes #6373

The MCP Apps sandbox bridge was sending an infinite loop of `ui/notifications/tool-input` notifications after receiving a tool result, causing re-rendering issues.

For testing, I used my demo mcp app: https://mcp-app-bench.onrender.com/mcp

### Before
Note the debug log panel on the right. Tool input is received many times when it should not be.
image

### After
Note the debug panel on the right. tool input and tool results are received once.
image

## Root Cause

The issue was caused by two problems:

1. **Unstable object references**: When `toolInput`, `toolInputPartial`, or `toolResult` props are passed as inline objects, a new object reference is created on every render. This triggered useEffect hooks to run repeatedly, sending duplicate notifications.

2. **Timing issue**: The useEffect hooks for sending tool data run before the guest iframe sends `ui/notifications/initialized`, so the initial tool data was never sent.

### Memoize props in `McpAppRenderer.tsx`
Added `useMemo` with `JSON.stringify` in the dependency array for `toolInput`, `toolInputPartial`, and `toolResult` to ensure stable object references when content has not changed.

### Send pending data on initialization in `useSandboxBridge.ts`
When the guest sends `ui/notifications/initialized`, we now check for any pending `toolInput` or `toolResult` and send them immediately.

## Testing

- TypeScript compiles without errors
- ESLint passes

### DIFF
--- a/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
+++ b/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
@@ -6,7 +6,7 @@
  * @see SEP-1865 https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx
  */
 
-import { useState, useCallback, useEffect, useMemo } from 'react';
+import { useState, useCallback, useEffect } from 'react';
 import { useSandboxBridge } from './useSandboxBridge';
 import {
   ToolInput,
@@ -160,30 +160,13 @@ export default function McpAppRenderer({
     setIframeHeight(newHeight);
   }, []);
 
-  // Memoize tool props to prevent unnecessary re-renders and duplicate notifications
-  // These props are often passed as inline objects which create new references on each render
-  const memoizedToolInput = useMemo(
-    () => toolInput,
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [JSON.stringify(toolInput)]
-  );
-  const memoizedToolInputPartial = useMemo(
-    () => toolInputPartial,
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [JSON.stringify(toolInputPartial)]
-  );
-  const memoizedToolResult = useMemo(
-    () => toolResult,
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [JSON.stringify(toolResult)]
-  );
   const { iframeRef, proxyUrl } = useSandboxBridge({
     resourceHtml: resourceHtml || '',
     resourceCsp,
     resourceUri,
-    toolInput: memoizedToolInput,
-    toolInputPartial: memoizedToolInputPartial,
-    toolResult: memoizedToolResult,
+    toolInput,
+    toolInputPartial,
+    toolResult,
     toolCancelled,
     onMcpRequest: handleMcpRequest,
     onSizeChanged: handleSizeChanged,
--- a/ui/desktop/src/components/ToolCallWithResponse.tsx
+++ b/ui/desktop/src/components/ToolCallWithResponse.tsx
@@ -1,6 +1,6 @@
 import { ToolIconWithStatus, ToolCallStatus } from './ToolCallStatusIndicator';
 import { getToolCallIcon } from '../utils/toolIconMapping';
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useRef, useState, useMemo } from 'react';
 import { Button } from './ui/button';
 import { ToolCallArguments, ToolCallArgumentValue } from './ToolCallArguments';
 import MarkdownContent from './MarkdownContent';
@@ -71,12 +71,19 @@ function isEmbeddedResource(content: Content): content is EmbeddedResource {
   return 'resource' in content && typeof (content as Record<string, unknown>).resource === 'object';
 }
 
-function maybeRenderMCPApp(
-  toolRequest: ToolRequestMessageContent,
-  toolResponse: ToolResponseMessageContent | undefined,
-  sessionId: string,
-  append?: (value: string) => void
-): React.ReactNode {
+interface McpAppWrapperProps {
+  toolRequest: ToolRequestMessageContent;
+  toolResponse?: ToolResponseMessageContent;
+  sessionId: string;
+  append?: (value: string) => void;
+}
+
+function McpAppWrapper({
+  toolRequest,
+  toolResponse,
+  sessionId,
+  append,
+}: McpAppWrapperProps): React.ReactNode {
   const requestWithMeta = toolRequest as ToolRequestWithMeta;
   let resourceUri = requestWithMeta._meta?.['ui/resourceUri'];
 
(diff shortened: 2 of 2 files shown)


ui/desktop/src/components/McpApps/useSandboxBridge.ts
```diff
@@ -54,13 +54,19 @@ export function useSandboxBridge(options: SandboxBridgeOptions): SandboxBridgeRe
   const isGuestInitializedRef = useRef(false);
   const [proxyUrl, setProxyUrl] = useState<string | null>(null);
 
+  // Track last sent values to prevent duplicate notifications
+  const lastSentToolInputRef = useRef<string | null>(null);
+  const lastSentToolResultRef = useRef<string | null>(null);
```
DOsinga: useRef seems overkill here; if we wanted to do it this way, we can just serialize on compare

ui/desktop/src/components/McpApps/useSandboxBridge.ts
```diff
@@ -194,8 +204,13 @@ export function useSandboxBridge(options: SandboxBridgeOptions): SandboxBridgeRe
     });
   }, [toolInputPartial, sendToSandbox]);
 
+  // Send tool result notification when it changes
+  // Use JSON stringification to compare values and prevent duplicate notifications
   useEffect(() => {
     if (!isGuestInitializedRef.current || !toolResult) return;
+    const serialized = JSON.stringify(toolResult);
+    if (serialized === lastSentToolResultRef.current) return;
+    lastSentToolResultRef.current = serialized;
```
DOsinga: we have three times the same code, but we're only fixing it twice; presumably toolInputPartial needs this too?

ui/desktop/src/components/McpApps/useSandboxBridge.ts
```diff
@@ -176,8 +182,12 @@ export function useSandboxBridge(options: SandboxBridgeOptions): SandboxBridgeRe
   }, [handleJsonRpcMessage]);
 
   // Send tool input notification when it changes
+  // Use JSON stringification to compare values and prevent duplicate notifications
```
DOsinga: I don't think this is the way. Can't we memoize the inputs to this component instead so the values don't change on re-rerender? that seems cleaner and easier to maintain

ui/desktop/src/components/McpApps/useSandboxBridge.ts
aharvard: ah yes, good catch

ui/desktop/src/components/McpApps/useSandboxBridge.ts
aharvard: agreed. will pivot to that approach.

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
```diff
@@ -160,13 +160,30 @@ export default function McpAppRenderer({
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [JSON.stringify(toolInput)]
+  );
+  const memoizedToolInputPartial = useMemo(
+    () => toolInputPartial,
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [JSON.stringify(toolInputPartial)]
+  );
+  const memoizedToolResult = useMemo(
+    () => toolResult,
```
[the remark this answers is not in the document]
DOsinga: hmm. good point from copilot @aharvard - doing it this way would not really be better than just having the JSON.stringify in the dependencies of where we use this though so I think we have to go further up the stack

    toolInput={{ arguments: requestWithMeta.toolCall.value.arguments || {} }}

seems like a likely culprit. when you think about it, the conversation *should* be stable. it does feel iffy though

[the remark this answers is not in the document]
aharvard: @DOsinga Updated per your feedback - moved the fix higher up the stack. Converted `maybeRenderMCPApp` to a proper component (`McpAppWrapper`) and added `useMemo` there to memoize `toolInput` and `toolResult` before passing them down. This keeps `useSandboxBridge` clean and addresses the root cause at the source. 

Ready for another look when you have a chance!

aharvard:
Gonna merge. 

As a reminder, we'll be able to delete a bunch of related code when this ships in the client SDK https://github.com/MCP-UI-Org/mcp-ui/pull/147 (might be available for us to use as soon as this week... gonna keep my eye on it)

Issue #6366, opened 2026-01-06
MCP Apps: `ui/notifications/size-changed` doesn't change the width


khandrew1:
**Describe the bug**
When `ui/notifications/size-changed` gets sent, Goose is deliberately ignoring the width. Is there a reason for this?
 
Reference: https://github.com/block/goose/blob/HEAD/ui/desktop/src/components/McpApps/McpAppRenderer.tsx#L158-L161

---

**To Reproduce**
Sent a size changed notification with new height and width on an MCP App (with auto resize off) and only the height changed.

---

**Expected behavior**
The width also gets updated

---
**Please provide the following information**
- **OS & Arch:** MacOS Tahoe 26.1
- **Interface:** UI
- **Version:** v1.19.0
---

DOsinga:
it's a good question. when a widget/app renders inside of a conversation, should it not just take the width of the parent? we could make it less wide, but wider doesn't work I think. I imagined that the width was more for a pop-out or standalone mode. /cc @aharvard

khandrew1:
Seems there's been an update to the spec regarding viewports. My thought process is that height/width should be controlled by the widget because the Host can already declare both a `maxHeight` and `maxWidth` in the host context. 

Ref: https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx#host-context-in-mcpuiinitializeresult (See `containerDimensions`).

aharvard:
@khandrew1, great reference. I was just about to provide more details about that.`hostContext.containerDimensions` is a new, breaking change and goose is still sending `hostContext.viewport`. (I've got a forthcoming PR in the works to pivot goose to `containerDimensions`)

Additionally, goose sends `hostContext.displayMode` of `inline` and `hostContext.availableDisplayModes` of `['inline']`. The intent here is that:
1. the host can have dynamic sizing logic based on mode (when we provide more modes, we can ignore certain sizes sent by the guest UI)
2. the guest UI can check what mode they are in and alter the design, if needed

To @DOsinga's point, I too would be worried about an app setting the width larger than the chat window. This makes me wonder if my comment will address the edge case: https://github.com/block/goose/pull/6376#discussion_r2669888535

aharvard:
FYI, I just opened a thread on the MCP Contributors Discord for the mcp-apps-wg to get some more guidance around this topic https://discord.com/channels/1358869848138059966/1458566824231833731

owtaylor:
[...]

PR #6376, opened 2026-01-07
fix(ui): respect width parameter in MCP app size-changed notifications

Abhijay007:
closes: #6366

## PR Description

This PR fixes a bug where the MCP app iframe was ignoring the width parameter in size-changed notifications. The fix enables MCP apps to dynamically control both their width and height.

- Adds state management for iframe width
- Updates the size-changed handler to capture and apply the width parameter
- Applies dynamic width to iframe styling with fallback to 100%

### Type of Change
- [x] Bug fix

### AI Assistance

### DIFF
--- a/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
+++ b/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
@@ -203,9 +203,7 @@ export default function McpAppRenderer({
   const handleSizeChanged = useCallback((height: number, width?: number) => {
     const newHeight = Math.max(DEFAULT_IFRAME_HEIGHT, height);
     setIframeHeight(newHeight);
-    if (width !== undefined) {
-      setIframeWidth(width);
-    }
+    setIframeWidth(width ?? null);
   }, []);
 
   const { iframeRef, proxyUrl } = useSandboxBridge({
@@ -268,6 +266,7 @@ export default function McpAppRenderer({
           src={proxyUrl}
           style={{
             width: iframeWidth ? `${iframeWidth}px` : '100%',
+            maxWidth: '100%',
             height: `${iframeHeight}px`,
             border: 'none',
             overflow: 'hidden',


ui/desktop/src/components/McpApps/McpAppRenderer.tsx
```diff
@@ -155,9 +156,12 @@ export default function McpAppRenderer({
     [append, sessionId, extensionName]
   );
 
-  const handleSizeChanged = useCallback((height: number, _width?: number) => {
+  const handleSizeChanged = useCallback((height: number, width?: number) => {
     const newHeight = Math.max(DEFAULT_IFRAME_HEIGHT, height);
     setIframeHeight(newHeight);
+    if (width !== undefined) {
+      setIframeWidth(width);
+    }
```
DOsinga: I don't think you want to have the if condition here; if the width changes from defined to undefined, the width should snap back to 100%

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
```diff
@@ -197,7 +201,7 @@ export default function McpAppRenderer({
           ref={iframeRef}
           src={proxyUrl}
           style={{
-            width: '100%',
+            width: iframeWidth ? `${iframeWidth}px` : '100%',
```
DOsinga, line 268: I don't know if we can do this is in css, but can we clip the iframe here to the size of the client? we don't want these widgets to be able to bork the conversation

DOsinga (approved):

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
aharvard: We _should_ be able to guard against apps being too wide for the chat container by also declaring `maxWidth: '100%'`

I have not tested this to ensure that it works.

@Abhijay007, can you give that a shot?

Abhijay007: Yea sure will look into this

Abhijay007:
Hi @zanesq can you please test it once with different test MCP apps

PR #6385, opened 2026-01-07
fix ResultsFormat error when loading old sessions

zanesq:
Follow up for https://github.com/block/goose/pull/6094

### Problem
Users reported an error when loading/chatting after upgrading the desktop app: stream error: failed to read session: data did not match any variant of untagged enum ResultFormat at line 1 column 136

### Root Cause
The issue was a schema mismatch between old and new session data formats:

- **Old format** (`ToolCall`): `arguments: serde_json::Value` (could be any JSON type)
- **New format** (`CallToolRequestParam`): `arguments: Option<JsonObject>` (must be an object or null)

When deserializing old sessions where `arguments` was a string, array, or number, the new deserialization logic failed because it couldn't match any variant of the `ResultFormat` enum.

### Solution
Modified `crates/goose/src/conversation/tool_result_serde.rs` to handle legacy `ToolCall` format:

1. Added a `LegacyToolCall` struct to accept `arguments: serde_json::Value`
2. Updated the `deserialize` function to include a `LegacySuccess` variant for backward compatibility
3. Added conversion logic in `LegacyToolCall::into_call_tool_request_param()`:
   - JSON object → used directly as `Some(map)`
   - `null` → converted to `None`
   - Other types (string, number, array, boolean) → wrapped in `{"value": <original>}`

### Files Changed

**`crates/goose/src/conversation/tool_result_serde.rs`**
- Added `LegacyToolCall` struct with `arguments: serde_json::Value`
- Added `into_call_tool_request_param()` conversion method
- Modified `deserialize` function to handle both new and legacy formats

**`crates/goose/src/conversation/message.rs`**
- Added 5 new unit tests covering legacy argument types:
  - `test_legacy_tool_request_with_value_arguments` (string)
  - `test_legacy_tool_request_with_array_arguments`
  - `test_legacy_tool_request_with_number_arguments`
  - `test_legacy_tool_request_with_null_arguments`
  - `test_legacy_tool_request_with_object_arguments`

### Verification
- ✅ Code compiles successfully
- ✅ All 27 message tests pass
- ✅ Clippy checks pass
- ✅ Code formatted with `cargo fmt`

### DIFF
--- a/crates/goose/src/conversation/message.rs
+++ b/crates/goose/src/conversation/message.rs
@@ -1440,196 +1440,85 @@ mod tests {
     }
 
     #[test]
-    fn test_legacy_tool_request_with_value_arguments() {
-        // Old ToolCall format had arguments as serde_json::Value, not Option<JsonObject>
-        // This tests backward compatibility with sessions that had non-object arguments
-        let legacy_json = r#"{
-            "role": "assistant",
-            "created": 1640995200,
-            "content": [{
-                "type": "toolRequest",
-                "id": "tool123",
-                "toolCall": {
-                    "status": "success",
-                    "value": {
-                        "name": "test_tool",
-                        "arguments": "string_argument"
-                    }
-                }
-            }],
-            "metadata": { "agentVisible": true, "userVisible": true }
-        }"#;
-
-        let message: Message = serde_json::from_str(legacy_json).unwrap();
-        assert_eq!(message.content.len(), 1);
-
-        if let MessageContent::ToolRequest(request) = &message.content[0] {
-            assert_eq!(request.id, "tool123");
-            if let Ok(tool_call) = &request.tool_call {
-                assert_eq!(tool_call.name, "test_tool");
-                // Non-object arguments should be wrapped in {"value": ...}
-                assert!(tool_call.arguments.is_some());
-                let args = tool_call.arguments.as_ref().unwrap();
-                assert_eq!(
-                    args.get("value"),
-                    Some(&serde_json::json!("string_argument"))
-                );
-            } else {
-                panic!("Expected successful tool call");
-            }
-        } else {
-            panic!("Expected ToolRequest content");
+    fn test_tool_request_with_value_arguments_backward_compatibility() {
+        struct TestCase {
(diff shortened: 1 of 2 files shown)


crates/goose/src/conversation/message.rs
```diff
@@ -1438,4 +1438,198 @@ mod tests {
             panic!("Expected ToolResponse content");
         }
     }
+
+    #[test]
```
DOsinga, line 1442: these tests are rather repetitive - if you wanted to do this, turn them into parameterized tests that only have the difference between the cases rather than repeating yourself.

but we should maybe think one deeper on how we avoid breaking this in the future - it happened before when we change the message format. adding tests like this is just very specifically going to test for this, doesn't seem that likely that we'll catch future issues.

wouldn't it be better to just check in a bunch of historic conversations and verify that we can load them?

crates/goose/src/conversation/tool_result_serde.rs
```diff
@@ -25,22 +25,59 @@ where
     }
 }
 
-pub fn deserialize<'de, T, D>(deserializer: D) -> Result<ToolResult<T>, D::Error>
+/// Legacy ToolCall format from older sessions (pre-rmcp migration).
+/// The old format had `arguments: Value` instead of `arguments: Option<JsonObject>`.
```
DOsinga: I think it would be better to remove the comments here and also not call it just legacy; surely we'll have more changes to formats coming up, representing in the name what's going on here would be better. LegacyToolCallWithValue? longer names beats comments here for sure.

crates/goose/src/conversation/tool_result_serde.rs
```diff
@@ -25,22 +25,59 @@ where
...
-    T: Deserialize<'de>,
     D: Deserializer<'de>,
 {
     #[derive(Deserialize)]
     #[serde(untagged)]
-    enum ResultFormat<T> {
-        Success { status: String, value: T },
-        Error { status: String, error: String },
+    enum ResultFormat {
+        NewSuccess {
```
DOsinga: same here; New is not going stay that way

PR #6392, opened 2026-01-08
feat(goose-acp): enable parallel sessions with isolated agent state

codefromthecrypt:
## Summary

This PR makes it possible to run ACP agents and sessions in parallel, with critical state isolated to each agent instance.

Before, we were using global singletons for most state, which made it impossible to change configuration without ENV or subprocess isolation. Moreover, unlike platform, built-in extensions launched subprocesses which implied you need a binary entrypoint to run them.

This PR extracts the most critical state into structs and ensures consistent threading to call sites. It also reconfigures built-in extensions to run in-process, via duplex streams.

Legacy entrypoints retain global singletons despite ACP entrypoints being isolated. This was done to allow progress, particularly to implement goose-server with ACP. This is also due to the pending PR to replace the CLI. We don't want to do too much work for code that is going to be replaced.

The most important impact is the new goose-acp crate which contains isolated code for critical paths even though the CLI entrypoint that uses it will act the same.

Later in the tests section you will see how the combination of these changes allow more ACP integration tests than before to complete in less than a second.

### Type of Change
- [x] Feature
- [x] Refactor / Code quality
- [x] Tests

### AI Assistance

### Testing

- Now that built-in are in-process like platform extensions, we don't need to build binaries to run ACP tests.
- Since we parameterized the `GooseAcpAgent` to accept variables via `AgentConfig`, tests don't rely on ENV variables.

Above and other changes allow us to safely run tests in parallel, removing env-lock, serial and reset patterns.
This means despite having more ACP integration tests, our performance is incredibly faster.

| Crate     | MAIN   | THIS PR | Speedup  |
|-----------|--------|---------|----------|
| goose-acp | 42.22s | 0.77s   | **55x**  |

### Works cited

What this PR does is a step towards a more modular codebase which avoids global singletons.
Some projects I studied for reference were:
- [hyperswitch](https://github.com/juspay/hyperswitch) - AppState takes storage config
- [lemmy](https://github.com/LemmyNet/lemmy) - LemmyContext created once, passed to handlers
- [typst](https://github.com/typst/typst) - Engine requires all dependencies at construction

### Related Issues
- #6305

### DIFF
--- a/crates/goose/src/agents/mcp_client.rs
+++ b/crates/goose/src/agents/mcp_client.rs
@@ -37,6 +37,23 @@ pub type BoxError = Box<dyn std::error::Error + Sync + Send>;
 
 pub type Error = rmcp::ServiceError;
 
+#[derive(Clone, Debug)]
+pub struct McpMeta {
+    pub session_id: String,
+}
+
+impl McpMeta {
+    pub fn new(session_id: impl Into<String>) -> Self {
+        Self {
+            session_id: session_id.into(),
+        }
+    }
+
+    fn inject_into_extensions(&self, extensions: Extensions) -> Extensions {
+        inject_session_id_into_extensions(extensions, &self.session_id)
+    }
+}
+
 #[async_trait::async_trait]
 pub trait McpClientTrait: Send + Sync {
     async fn list_tools(
@@ -49,7 +66,7 @@ pub trait McpClientTrait: Send + Sync {
         &self,
         name: &str,
         arguments: Option<JsonObject>,
-        session_id: &str,
+        meta: McpMeta,
         cancel_token: CancellationToken,
     ) -> Result<CallToolResult, Error>;
 
@@ -452,7 +469,7 @@ impl McpClientTrait for McpClient {
         &self,
         name: &str,
         arguments: Option<JsonObject>,
-        session_id: &str,
+        meta: McpMeta,
         cancel_token: CancellationToken,
     ) -> Result<CallToolResult, Error> {
         let res = self
@@ -463,7 +480,7 @@ impl McpClientTrait for McpClient {
                         arguments,
                     },
                     method: Default::default(),
-                    extensions: inject_session_id_into_extensions(Default::default(), session_id),
+                    extensions: meta.inject_into_extensions(Default::default()),
                 }),
                 cancel_token,
             )
--- a/crates/goose/src/config/permission.rs
+++ b/crates/goose/src/config/permission.rs
@@ -43,10 +43,12 @@ impl PermissionManager {
     pub fn new(config_dir: PathBuf) -> Self {
         let permission_path = config_dir.join(PERMISSION_FILE);
         let permission_map = if permission_path.exists() {
-            let file_contents = fs::read_to_string(&permission_path).unwrap();
+            let file_contents =
+                fs::read_to_string(&permission_path).expect("Failed to read permission.yaml");
             serde_yaml::from_str(&file_contents).unwrap_or_else(|_| HashMap::new())
         } else {
-            fs::create_dir_all(&config_dir).unwrap();
+            // Consolidate directory creation for re-use in global singleton or ACP.
+            fs::create_dir_all(&config_dir).expect("Failed to create config directory");
             HashMap::new()
         };
         PermissionManager {
@@ -149,8 +151,9 @@ impl PermissionManager {
         }
 
         // Serialize the updated permission map and write it back to the config file
-        let yaml_content = serde_yaml::to_string(&*map).unwrap();
-        fs::write(&self.config_path, yaml_content).unwrap();
+        let yaml_content =
+            serde_yaml::to_string(&*map).expect("Failed to serialize permission config");
+        fs::write(&self.config_path, yaml_content).expect("Failed to write to permission.yaml");
     }
 
     /// Removes all entries where the principal name starts with the given extension name.
@@ -168,8 +171,9 @@ impl PermissionManager {
                 .retain(|p| !p.starts_with(extension_name));
         }
 
-        let yaml_content = serde_yaml::to_string(&*map).unwrap();
-        fs::write(&self.config_path, yaml_content).unwrap();
+        let yaml_content =
+            serde_yaml::to_string(&*map).expect("Failed to serialize permission config");
+        fs::write(&self.config_path, yaml_content).expect("Failed to write to permission.yaml");
     }
 }
 
--- a/crates/goose/src/session/session_manager.rs
+++ b/crates/goose/src/session/session_manager.rs
@@ -473,7 +473,7 @@ impl sqlx::FromRow<'_, sqlx::sqlite::SqliteRow> for Session {
 impl SessionStorage {
     fn create_pool(path: &Path) -> Pool<Sqlite> {
         if let Some(parent) = path.parent() {
-            fs::create_dir_all(parent).unwrap();
+            fs::create_dir_all(parent).expect("Failed to create session database directory");
         }
 
         let options = SqliteConnectOptions::new()
--- a/crates/goose-cli/src/scenario_tests/mock_client.rs
+++ b/crates/goose-cli/src/scenario_tests/mock_client.rs
@@ -1,7 +1,7 @@
 //! MockClient is a mock implementation of the McpClientTrait for testing purposes.
 //! add a tool you want to have around and then add the client to the extension router
 
-use goose::agents::mcp_client::{Error, McpClientTrait};
+use goose::agents::mcp_client::{Error, McpClientTrait, McpMeta};
 use rmcp::{
     model::{
         CallToolResult, Content, ErrorData, GetPromptResult, ListPromptsResult,
@@ -94,7 +94,7 @@ impl McpClientTrait for MockClient {
         &self,
         name: &str,
         arguments: Option<serde_json::Map<String, Value>>,
-        _session_id: &str,
+        _meta: McpMeta,
         _cancel_token: CancellationToken,
     ) -> Result<CallToolResult, Error> {
         if let Some(handler) = self.handlers.get(name) {
--- a/crates/goose/src/agents/chatrecall_extension.rs
+++ b/crates/goose/src/agents/chatrecall_extension.rs
@@ -1,5 +1,5 @@
 use crate::agents::extension::PlatformExtensionContext;
-use crate::agents::mcp_client::{Error, McpClientTrait};
+use crate::agents::mcp_client::{Error, McpClientTrait, McpMeta};
 use anyhow::Result;
 use async_trait::async_trait;
 use indoc::indoc;
@@ -294,9 +294,10 @@ impl McpClientTrait for ChatRecallClient {
         &self,
         name: &str,
         arguments: Option<JsonObject>,
-        session_id: &str,
+        meta: McpMeta,
         _cancellation_token: CancellationToken,
     ) -> Result<CallToolResult, Error> {
+        let session_id = &meta.session_id;
         let content = match name {
             "chatrecall" => self.handle_chatrecall(session_id, arguments).await,
             _ => Err(format!("Unknown tool: {}", name)),
--- a/crates/goose/src/agents/code_execution_extension.rs
+++ b/crates/goose/src/agents/code_execution_extension.rs
@@ -1,6 +1,6 @@
 use crate::agents::extension::PlatformExtensionContext;
 use crate::agents::extension_manager::get_parameter_names;
-use crate::agents::mcp_client::{Error, McpClientTrait};
+use crate::agents::mcp_client::{Error, McpClientTrait, McpMeta};
 use anyhow::Result;
 use async_trait::async_trait;
 use boa_engine::builtins::promise::PromiseState;
@@ -824,11 +824,11 @@ impl McpClientTrait for CodeExecutionClient {
         &self,
         name: &str,
         arguments: Option<JsonObject>,
-        session_id: &str,
+        meta: McpMeta,
         _cancellation_token: CancellationToken,
     ) -> Result<CallToolResult, Error> {
         let content = match name {
-            "execute_code" => self.handle_execute_code(session_id, arguments).await,
+            "execute_code" => self.handle_execute_code(&meta.session_id, arguments).await,
             "read_module" => self.handle_read_module(arguments).await,
             "search_modules" => self.handle_search_modules(arguments).await,
             _ => Err(format!("Unknown tool: {name}")),
@@ -902,7 +902,7 @@ mod tests {
             .call_tool(
                 "execute_code",
                 Some(args),
-                "test-session-id",
+                McpMeta::new("test-session-id"),
                 CancellationToken::new(),
             )
             .await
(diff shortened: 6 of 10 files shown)


crates/goose/tests/mcp_replays/github-mcp-serverstdio
```diff
@@ -7,6 +7,6 @@ STDERR: time=2025-12-11T17:58:47.640-05:00 level=INFO msg="server session connec
...
 STDIN: {"jsonrpc":"2.0","method":"notifications/initialized"}
 STDERR: time=2025-12-11T17:58:47.642-05:00 level=INFO msg="session initialized"
-STDIN: {"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"_meta":{"progressToken":0},"name":"get_file_contents","arguments":{"owner":"block","path":"README.md","repo":"goose","sha":"ab62b863c1666232a67048b6c4e10007a2a5b83c"}}}
+STDIN: {"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"_meta":{"goose-session-id":"test-session-id","progressToken":0},"name":"get_file_contents","arguments":{"owner":"block","path":"README.md","repo":"goose","sha":"ab62b863c1666232a67048b6c4e10007a2a5b83c"}}}
```
codefromthecrypt, line 10: now that session is propagated via parameter in the case we always know it, the expectation file bumped for good reason.

crates/goose/tests/mcp_integration_test.rs
```diff
@@ -60,6 +75,44 @@ impl Provider for MockProvider {
     }
 }
 
+fn build_and_get_binary_path() -> PathBuf {
```
codefromthecrypt, line 78: now that builtin extensions no longer require being launched as subprocesses (`mcp xxx` args), ACP doesn't need a build to run tests. This leaves only one path where we launch processes in tests, the old one for MCP replay, only used when recording. This is put back to exactly the same code from before we needed to extract a common cargo build utility.

crates/goose/src/scheduler_trait.rs
```diff
@@ -39,3 +40,104 @@ pub trait SchedulerTrait: Send + Sync {
         sched_id: &str,
     ) -> Result<Option<(String, DateTime<Utc>)>, SchedulerError>;
 }
+
+const UNAVAILABLE_MESSAGE: &str = "Scheduler not available. This tool only works in server mode.";
```
codefromthecrypt: before we had a single call site we noticed scheduler in use and disabled it for CLI sessions. This extracts so it isn't optional. Unless there's a good reason to have CLI sessions fail on scheduler, I recommend in this PR or the next just letting them work and removing this workaround and/or only using it in tests

crates/goose/src/session/session_manager.rs
```diff
@@ -61,7 +60,8 @@ impl std::str::FromStr for SessionType {
     }
 }
 
-static SESSION_STORAGE: OnceCell<Arc<SessionStorage>> = OnceCell::const_new();
+static SESSION_STORAGE: LazyLock<Arc<SessionStorage>> =
+    LazyLock::new(|| Arc::new(SessionStorage::new(Paths::data_dir())));
```
codefromthecrypt, line 64: in order to propagate and also still have a singleton I looked around at about 10 rust projects and this lazy lock pattern seemed the best, as old sites still get a global singleton until/unless updated, and new can propagate through in all cases the actual connect stuff is deferred to first.

crates/goose/src/permission/permission_confirmation.rs
```diff
@@ -7,6 +7,7 @@ pub enum Permission {
     AllowOnce,
     Cancel,
     DenyOnce,
+    AlwaysDeny,
```
codefromthecrypt, line 10: we were missing this even though our permissions system allows persistence of deny. This is needed for ACP and the motivation was testing permissions which had a bug and they now are testable without clobbering or ENV

crates/goose-acp/tests/server_test.rs
```diff
@@ -0,0 +1,400 @@
...
+
+                test_fn(cx.clone(), session.session_id, updates).await;
+                Ok(())
+            }
+        })
+        .await
+        .unwrap();
+}
+
+#[test_case(Some(PermissionOptionKind::AllowAlways), ToolCallStatus::Completed, "user:\n  always_allow:\n  - lookup__get_code\n  ask_before: []\n  never_allow: []\n"; "allow_always")]
```
codefromthecrypt, line 342: this is the new test and notice no ACP test requires serial anymore, and none use ENV or clobber eachother's database or permissions files

codefromthecrypt (commented):
notes on the important things

crates/goose/src/config/permission.rs
```diff
@@ -3,8 +3,14 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::fs;
 use std::path::{Path, PathBuf};
+use std::sync::{Arc, LazyLock, RwLock};
 use utoipa::ToSchema;
 
+const PERMISSION_FILE: &str = "permission.yaml";
+
+static PERMISSION_MANAGER: LazyLock<Arc<PermissionManager>> =
```
codefromthecrypt, line 11: permission manager needs to be a singleton where used, so I used exact same pattern as session storage. before depending on call site reads might come from memory which could have problems when not shared. now it is always shared and pushed down consistently

crates/goose-bench/Cargo.toml
```diff
@@ -11,21 +11,21 @@ description.workspace = true
 workspace = true
 
 [dependencies]
-anyhow = "1.0"
+anyhow = { workspace = true }
```
michaelneale, line 14: nice - this does what I think it does which makes good sense.

michaelneale:
so far working really well!

michaelneale:
I also like that it is half as many processes for usual workloads. I want to stamp this but need to work out how to get enough consensus but LGTM - I will try next running it bundled

codefromthecrypt:
ps confirmed in zed the missing "reject always" works and updates my permissions file properly

Screenshot 2026-01-08 at 12 10 45 PM

```bash
$ cat ~/.config/goose/permission.yaml
user:
  always_allow: []
  ask_before: []
  never_allow:
  - kiwi__search-flight
```

codefromthecrypt:
ps verified there's no bug with us about the double check vs double x on deny. I'll raise a PR with zed to fix that, right now the only icon they have for double is approve

codefromthecrypt:
opened https://github.com/zed-industries/zed/pull/46329 on zed for the unrelated UI glitch I noticed testing this

codefromthecrypt:
@angiejones @alexhancock FYI I tested in webstorm (jetbrains) and it works. One note is that current version of Jetbrains doesn't respond to permissions requests like Zed, etc. So, don't change the GOOSE_MODE default from auto until that's not the case.

Screenshot 2026-01-08 at 2 21 57 PM

codefromthecrypt:
also checked 3rd party MCP also works in jetbrains same permissions caveat which isn't new to this PR and probably already solved, just not released in latest jetbrains.

Screenshot 2026-01-08 at 2 11 40 PM
Screenshot 2026-01-08 at 2 24 53 PM

codefromthecrypt:
rebased no code changes since last review otherwise

codefromthecrypt:
this PR takes quite a lot of effort to rebase every couple days. let me know if there is anything I can do to get this in so my "goose time" isn't dominated by rebasing.

crates/goose/src/agents/mcp_client.rs
```diff
@@ -61,27 +49,50 @@ pub trait McpClientTrait: Send + Sync {
         &self,
         name: &str,
         arguments: Option<JsonObject>,
+        session_id: &str,
```
baxen: I get why we need session id now in so many downstream tools (no more consistent way to lookup the process's current session). I like how it becomes _meta in the MCP spec. Can be a followup, but i'd cnsider having a "meta" arg in call_tool instead of session_id and always add session id into it

crates/goose/src/config/permission.rs
```diff
@@ -23,62 +29,44 @@ pub struct PermissionConfig {
...
-        // Load the existing configuration file or create an empty map if the file doesn't exist
-        let permission_map = if config_path.exists() {
-            // Load the configuration file
-            let file_contents =
-                fs::read_to_string(&config_path).expect("Failed to read permission.yaml");
+impl PermissionManager {
+    pub fn new(config_dir: PathBuf) -> Self {
+        let permission_path = config_dir.join(PERMISSION_FILE);
+        let permission_map = if permission_path.exists() {
+            let file_contents = fs::read_to_string(&permission_path).unwrap();
```
baxen: nit: unwrap -> expect (old code had it)

crates/goose/src/config/permission.rs
```diff
@@ -23,62 +29,44 @@ pub struct PermissionConfig {
...
             serde_yaml::from_str(&file_contents).unwrap_or_else(|_| HashMap::new())
         } else {
-            HashMap::new() // No config file, create an empty map
+            fs::create_dir_all(&config_dir).unwrap();
```
baxen: nit: missing why we need to create dir here? not a serious concern though

crates/goose/src/session/session_manager.rs
```diff
@@ -479,52 +471,61 @@ impl sqlx::FromRow<'_, sqlx::sqlite::SqliteRow> for Session {
...
-            }
-
-            storage
-        };
-
-        Ok(storage)
-    }
+    fn create_pool(path: &Path) -> Pool<Sqlite> {
+        if let Some(parent) = path.parent() {
+            fs::create_dir_all(parent).unwrap();
```
baxen: nit: could use an expect here for a better error message (we had a lot of access issues with some peoples xdg dirs)

crates/goose-mcp/src/lib.rs
```diff
@@ -19,3 +21,52 @@ pub use computercontroller::ComputerControllerServer;
...
 pub use tutorial::TutorialServer;
+
+pub type SpawnServerFn = fn(tokio::io::DuplexStream, tokio::io::DuplexStream);
+
+pub struct BuiltinDef {
+    pub name: &'static str,
+    pub spawn_server: SpawnServerFn,
+}
+
+fn spawn_and_serve<S>(
```
baxen, line 32: this is great, really nice to clean up the processes we were leaving here

crates/goose-acp/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+description.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+goose = { path = "../goose" }
+goose-mcp = { path = "../goose-mcp" }
+rmcp = { workspace = true }
+sacp = "10.1.0"
```
baxen, line 17: discussed in discord, will bring this to the official rust sdk once they finish their refactor

baxen (approved):
LGTM, a lot of great changes in here for the architecture to enable this

A lot of my comments i think can be followups, along with eventually getting the rest of the global state out like you mentioned (should make the desktop app multi chat less confusing too!)

crates/goose/src/agents/mcp_client.rs
codefromthecrypt: good point. when we get around to otel we'll need to implicitly add trace IDs, but maybe we wouldn't if we were passing meta (since we could do it manually just like we are for session ID)

crates/goose/src/config/permission.rs
codefromthecrypt: we have a choice to ensure the directory is there in the beginning or later when we persist. Not all code paths to this point ensure the config dir is there.. I wanted to put a TODO why but I have noticed not everyone likes to have TODOs.. if it were a TODO, I would say TODO: ensure all XDG paths are ensured at all entrypoints

crates/goose/src/session/session_manager.rs
codefromthecrypt: sgtm

codefromthecrypt:
thanks for the look @baxen I added a commit which I think got the bulk of things

DOsinga:
oops this totally slipped my mind. I'll have a look too since this is kinda big?

crates/goose/src/scheduler_trait.rs
DOsinga: I'm not sure I completely follow; the reason the CLI didn't have the scheduler is that a CLI session just only runs that session so it is never going to run scheduled jobs. we could still allow the CLI to set-up scheduled jobs, but you'd need the desktop to actually trigger them if that makes sense

DOsinga:
I went through this once. it's clearly an improvement and we should possibly just merge as is, but after this I think the architecture would be left in a slightly confusing state.

This is solid progress. The dependency injection via AgentConfig is the right move, and making the core components instantiable (vs global singletons) unblocks library mode. Let's merge this forward.

  What Works Well

  1. AgentConfig pattern - bundling SessionManager, PermissionManager, scheduler into config for DI
  2. Agent.with_config() - enables multiple independent instances
  3. Threading session_id through methods - makes Agent more session-agnostic
  4. Test improvements - no more #[serial] or unsafe reset hacks

  What's Confusing

  1. AgentManager + Agent both have SessionManager

  pub struct AgentManager {
      sessions: LruCache<String, Arc<Agent>>,  // session_id -> Agent
      session_manager: Arc<SessionManager>,     // Also has session manager
  }

  pub struct Agent {
      config: AgentConfig {
          session_manager: Arc<SessionManager>,  // Agent also has it
      }
  }

  AgentManager finds the Agent for a session_id, but Agent also needs SessionManager. This double indirection is confusing - why does AgentManager need to know about sessions at all if it's just caching agents?

  2. What goes in AgentConfig vs on Agent?

  pub struct AgentConfig {
      session_manager, permission_manager, scheduler, goose_mode  // In config
  }

  pub struct Agent {
      provider: SharedProvider,              // NOT in config
      extension_manager: Arc<ExtensionManager>,  // NOT in config
      config: AgentConfig,
  }

  Why are provider and extension_manager outside AgentConfig? What's the principle for what goes where?

  3. Two competing patterns

  - goose-server: Uses AgentManager, creates 1 Agent per session
  - goose-acp: Bypasses AgentManager, uses 1 Agent for all sessions

  Both work, but it's unclear when to use which pattern or what Agent's relationship to sessions is meant to be.

  4. Provider in multiple places

  - Agent.provider (mutable via update_provider)
  - Session.provider (string stored in DB)
  - GooseAcpConfig.provider (for initialization)

  What's the source of truth?


In general though this does feel like a good improvement. longer term though I think we might want to deprecate the whole notion of agents as a thing.  we have sessions and an extension manager. then when you want to do something you can start an agent loop with a session and your preferences (goose mode, number of turns?). it then asks the (global?) extension manager to get it the extensions it needs. for stateless mcp, it can just reuse what is already there and that should be the general case. maybe sometimes it needs to start an extra one if one is busy. but other than that, the agent loop can just be a function without any agent state. thoughts?

codefromthecrypt:
@DOsinga I agree things aren't yet coherent and there are still some mixed responsibilities.. there are a lot of call sites to old constructors and in some cases agent == agent+session

I was told that we are killing the old CLI which is a large part of some copy/paste stuff, and after that it would be goose-server+acp only, let's do another major arch pass when that's in!

PR #6393, opened 2026-01-08
feat(gcp-vertex): add model list with org policy filtering

rabi:
## Summary
- Add static KNOWN_MODELS list for Claude and Gemini models
- Filter models by GCP org policy (constraints/vertexai.allowedModels)
- Preserve unlisted model entry via allows_unlisted_models flag
- Replace ClaudeVersion/GeminiVersion enums with simple string storage

Note: Vertex AI does not provide a public REST API to list available foundation/publisher models, hence the static KNOWN_MODELS approach.

### Type of Change
- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Unit tests and tested with Vertex AI.

### DIFF
--- a/crates/goose-cli/src/commands/configure.rs
+++ b/crates/goose-cli/src/commands/configure.rs
@@ -436,12 +436,12 @@ fn select_model_from_list(
     provider_meta: &goose::providers::base::ProviderMetadata,
 ) -> anyhow::Result<String> {
     const MAX_MODELS: usize = 10;
+    const UNLISTED_MODEL_KEY: &str = "__unlisted__";
+
     // Smart model selection:
     // If we have more than MAX_MODELS models, show the recommended models with additional search option.
     // Otherwise, show all models without search.
-
     if models.len() > MAX_MODELS {
-        // Get recommended models from provider metadata
         let recommended_models: Vec<String> = provider_meta
             .known_models
             .iter()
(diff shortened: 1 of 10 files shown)


rabi:
@jamadeo Hey PTAL. This removes some cruft and makes is easier to select/provide models when configuring gcp_vertex_ai provider.

crates/goose-cli/src/commands/configure.rs
```diff
@@ -401,32 +401,68 @@ fn select_model_from_list(
                 ),
             );
 
+            if provider_meta.allows_unlisted_models {
+                model_items.push((
+                    UNLISTED_MODEL_KEY.to_string(),
+                    "Enter a model not listed...".to_string(),
+                    "For models blocked by org policy or newly released",
```
jamadeo: this text seems somewhat provider-specific, I'd just leave the extra text here empty as it's pretty self explanatory

crates/goose-cli/src/commands/configure.rs
```diff
@@ -401,32 +401,68 @@ fn select_model_from_list(
...
-                    .iter()
-                    .map(|m| (m, m.as_str(), ""))
-                    .collect::<Vec<_>>(),
-            )
-            .interact()?
-            .to_string())
+        let mut model_items: Vec<(String, String, &str)> =
+            models.iter().map(|m| (m.clone(), m.clone(), "")).collect();
+
+        if provider_meta.allows_unlisted_models {
```
jamadeo, line 493: Maybe we should just always allow this for any provider? Since we now have the ability, we might as well let the user opt to do this

crates/goose-cli/src/commands/configure.rs
rabi: Thanks, hint is not relevant, removed it.

crates/goose-cli/src/commands/configure.rs
rabi: I think most providers expose a model‑listing API (Vertex AI is the exception), so offering an ‘unlisted model’ option for those providers is less useful and can lead to invalid selections and subsequent error when configuring the provider.

PR #6396, opened 2026-01-08
feat: add generic HTTP LLM provider support

shiningstone23:
Add a flexible provider that can connect to any HTTP-based LLM API using configuration-driven request/response mapping.

Features:
- JSON configuration schema for defining custom LLM API integrations
- Template variable substitution for requests (${VAR}, ${VAR:number}, etc.)
- JSONPath-based response parsing ($.content, $.error, etc.)
- Multiple authentication methods (Bearer, Header, Basic, Query)
- SSE/Chunked streaming support
- Tool injection for LLMs without native tool calling support
- Skip SSL verification option for self-signed certificates
- Custom environment variable prefix support

Usage:
1. Create a JSON config file in ~/.config/goose/custom_providers/
2. Define endpoint, auth, request template, and response parsing
3. Run `goose configure` to set up the provider

Example config: crates/goose/examples/generic_http_provider.json

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:

### DIFF
--- a/crates/goose/examples/generic_http_provider.json
+++ b/crates/goose/examples/generic_http_provider.json
@@ -1,0 +1,56 @@
+{
+  "name": "custom_llm",
+  "display_name": "Custom LLM Provider",
+  "description": "Custom LLM provider using generic HTTP provider",
+  "engine": "generic_http",
+  "env_prefix": "CUSTOM_LLM",
+  "endpoint": {
+    "url_template": "http://localhost:8765/v1/chat/messages",
+    "method": "POST",
+    "timeout_seconds": 120,
+    "skip_ssl_verify": false
+  },
+  "auth": {
+    "type": "header",
+    "header_name": "Authorization",
+    "value_template": "Bearer ${TOKEN}"
+  },
+  "headers": {
+    "Content-Type": "application/json"
+  },
+  "request": {
+    "template": {
+      "model": "${MODEL}",
+      "messages": [{"role": "user", "content": "${PROMPT}"}],
+      "stream": "${IS_STREAM:bool}"
+    }
+  },
+  "response": {
+    "content_path": "$.content",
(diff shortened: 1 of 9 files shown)


DOsinga:
thanks for the contribution. what is the advantage over this over the current approach of custom providers? I see this has more fields, but can't we just add those to custom providers, rather than ending up with what is effectively two different json formats to specify a generic provider? haven't looked closely at the code so I might be missing something here of course.

shiningstone23:
Thanks for the review!

The main use case for this generic HTTP provider is supporting internal/private LLM endpoints 
that have different URL structures or authentication requirements from standard providers.

For example, in my company we have an internal LLM gateway with:
- Custom endpoint URL pattern
- Specific authentication headers (x-openapi-token with Bearer format)
- Custom request/response field mappings

The current custom provider approach works great for OpenAI-compatible APIs, but doesn't 
support cases where:
1. The URL structure is completely different
2. Authentication uses non-standard headers
3. Request/response field names don't match OpenAI format
4. Tool calling needs to be injected via system prompt (for LLMs without native support)

This generic HTTP provider fills that gap by allowing users to define arbitrary 
request templates, response parsing paths, and authentication methods through JSON 
configuration.

Let me know if you'd like me to adjust the approach!

DOsinga:
yeah, my strong preference would be to build out the declarative providers with more features so we can support more providers. just linked to a related PR.

DOsinga:
closing this for now

Issue #6395, opened 2026-01-08
Extension failure should display an error


alecthomas:
**Describe the bug**

I ran the Goose CLI and got this:

```
~ $ goose
◇  Would you like me to help debug the 'fetch' extension failure?
│  Yes 
│
🔧 Starting debugging session...
Analyzing the extension failure...
I'll help you diagnose the "fetch" extension startup issue. Let me investigate the common causes of this "No such file or directory" error.
```

Ideally of course, there would be no failure, but in the case that there is, I'd expect to see output of the failure.

---

**To Reproduce**

I'm not sure how to reproduce it.

---

**Expected behavior**

I expect to see the error text.

---

**Please provide the following information**
- **OS & Arch:** Darwin arm64
- **Interface:** CLI
- **Version:** 1.19.1
- **Extensions enabled:** Developer, Extension Manager, Skills, Todo
- **Provider & Model:** Sonnet 4.5

DOsinga:
you tend to get these errors if there's anything up with the VPN; internal MCPs don't run if it is not up, some externals don't run if it is. sigh. but you are right, we should do a better job reporting here.

DOsinga:
closing because fixed

PR #6398, opened 2026-01-08
CLI show extension errors

DOsinga:
## Summary
We offered help to fix but didn't show what was broken when an extension didn't load in the CLI

Before:
image

After:
image

Fixes:
#6395


alecthomas (approved):
Yay, thanks Douwe!

PR #6400, opened 2026-01-08
fix(google): treat signed text as regular content in streaming

rabi:
## Summary
Change streaming to use SignedTextAsRegularText instead of SkipSignedText. Signatures are tracked via last_signature and attached to function calls through effective_signature = signature.or(last_signature).
 
Gemini 2.5 models include thoughtSignature on the first streaming chunk regardless of whether function calls follow. The old SkipSignedText behavior discarded text parts with signatures, causing content loss.
 
 Signature placement by model:
 - Gemini 2.5: first part (with or without function calls)
 - Gemini 3 with function calls: first function call part
 - Gemini 3 without function calls: last part if thought generated

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Unit test and tested with 2.5 models locally to verify the fix for their unique behaviour.

### DIFF
--- a/crates/goose/src/providers/formats/google.rs
+++ b/crates/goose/src/providers/formats/google.rs
@@ -309,7 +309,6 @@ pub fn process_map(map: &Map<String, Value>, parent_key: Option<&str>) -> Value 
 
 #[derive(Clone, Copy)]
 enum SignedTextHandling {
-    SkipSignedText,
     SignedTextAsThinking,
     SignedTextAsRegularText,
 }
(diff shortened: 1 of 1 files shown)


rabi:
@jamadeo Hey! PTAL. This would fix a unique behavior of Gemini 2.5 models.

PR #6401, opened 2026-01-08
feat(providers): add custom headers support to Anthropic provider

trjordan:
## Summary

- Adds custom headers support to `AnthropicProvider::from_custom_config`
- Enables Anthropic-compatible proxies (like Portkey) that require custom headers for authentication
- Mirrors the existing implementation in `OpenAiProvider::from_custom_config`

## Test plan

- [x] `cargo check` passes
- [x] `cargo test -p goose --lib anthropic` - all 8 tests pass
- [x] `cargo clippy` - no new warnings
- [ ] Manual testing with Portkey proxy

### DIFF
--- a/crates/goose/src/providers/anthropic.rs
+++ b/crates/goose/src/providers/anthropic.rs
@@ -91,8 +91,19 @@ impl AnthropicProvider {
             key: api_key,
         };
 
-        let api_client = ApiClient::new(config.base_url, auth)?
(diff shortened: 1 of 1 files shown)


DOsinga:
thanks for doing this. I noticed this is also missing from Ollama so added that here: https://github.com/block/goose/pull/6774 - I'm going to continue there, so closing this one now.

PR #6402, opened 2026-01-08
fix: custom provider URI changes now work without restart

### LINKED ISSUE #6303: Editing model settings to change URI from https to just http is NOT picked up until app restart.
**Describe the bug**

When working with local (ollama) LLM servers, when changing the uri from https to just http (ollama doesn't need https if on the local area network), the chat model session still tries to use https when contacting the local network service.  I had to restart the app to get it to be recognized and just for good measure, changed off to a different model and then changed back.

---

**To Reproduce**
1. Create a local ollama model configuration (I suspect it will happen with anything)
2. set the url to be https based like :  https://myserver.local:11434/v1/chat/completions
3. Attempt to use in a chat session, a 404 error will be issued
4. Go back to the model configuration and change the uri from "https" to "http"
5. Try again.  chat will still try to use "http" to contact local network server.

---

**Expected behavior**
Expect that after updating the uri in model configurator, that the chat will pick up the change.

---

- **OS & Arch:** [Macbook Pro M2]
- **Interface:** [UI]
- **Version:** [e.g. v1.18.0]

Abhijay007:
closes #6303

### Type of Change
- [x] Bug fix

### AI Assistance

Before:  

After:

### DIFF
--- a/crates/goose-server/src/routes/config_management.rs
+++ b/crates/goose-server/src/routes/config_management.rs
@@ -780,6 +780,12 @@ pub async fn update_custom_provider(
         tracing::warn!("Failed to refresh custom providers after update: {}", e);
     }
 
+    if let Ok(manager) = goose::execution::manager::AgentManager::instance().await {
+        if let Err(e) = manager.refresh_provider_for_sessions(&id).await {
+            tracing::warn!("Failed to refresh provider for existing sessions: {}", e);
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


DOsinga:
closing this since it is draft and > 2 weeks. I get the intention, but I think we should fix this differently - the providers should just be shared between sessions (they should be stateless) and then we can just update it there. that seems more robust in general.

PR #6419, opened 2026-01-09
fix: inconsistent API url requirement between desktop and CLI versions

### LINKED ISSUE #6354: Inconsistent API Url requirement between desktop and CLI versions
**Describe the bug**

When using local openai compatible providers, the desktop version expect something like http://192.168.1.5:8000/ without "v1" but the CLI version (using goose configure) uses http://192.168.1.5:8000/v1

---
**Steps to reproduce**

Install latest desktop and CLI goose on MacOS, use goose configure to add custom provider, require API url like http://192.168.1.5:8000/v1

However, when adding custom provider through desktop version, it requires http://192.168.1.5:8000 without "v1"

---

**Expected behavior**
Consistent for both versions, i.e., requires standard openai api compatible url like 
http://192.168.1.5:8000/v1

Abhijay007:
Closes #6354 

## PR descripton 

This PR fixes an inconsistency where the desktop version and CLI version of goose required different API URL formats for custom OpenAI-compatible providers. Previously, the desktop and CLI showed /v1/messages as the placeholder, but the actual path construction logic didn't properly handle the standard OpenAI format of /v1 as a base path.

### Type of Change
- [x] Bug fix

### AI Assistance

- [x] This PR was created or reviewed with AI assistance, Copilot for PR description

### DIFF
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -154,9 +154,7 @@ impl OpenAiProvider {
             format!("{}://{}", url.scheme(), url.host_str().unwrap_or(""))
         };
         let base_path = url.path().trim_start_matches('/').to_string();
-        let base_path = if base_path.is_empty() {
-            "v1/chat/completions".to_string()
-        } else if base_path == "v1" || base_path == "v1/" {
+        let base_path = if base_path.is_empty() || base_path == "v1" || base_path == "v1/" {
             "v1/chat/completions".to_string()
         } else {
             base_path


crates/goose/src/providers/openai.rs
```diff
@@ -156,8 +156,12 @@ impl OpenAiProvider {
...
         let base_path = if base_path.is_empty() {
             "v1/chat/completions".to_string()
-        } else {
+        } else if base_path.ends_with("/chat/completions") {
             base_path
+        } else if base_path.ends_with('/') {
+            format!("{}chat/completions", base_path)
+        } else {
+            format!("{}/chat/completions", base_path)
         };
```
DOsinga, line 161: this strikes me as a tad aggressive. I could see autocompleting v1/ or v1, but other than that, if the user explicitly gives us /my-endpoint, should we override it?

Abhijay007: Oh yea it is, let revise it to be more conservative in its behavior, maybe to only auto-completes for an empty path, v1, or v1/ (the standard OpenAI format) and other explicitly provided path is preserved as it is respecting the user choice

DOsinga: yeah, that sounds like the right way to go about it!

PR #6420, opened 2026-01-09
User configurable templates

DOsinga:
## Summary

Simplify how we do templating and make it possible for users to edit prompts

### Example

Change the system prompt to make replies in Dutch by default:

image

### DIFF
--- a/crates/goose/src/prompt_template.rs
+++ b/crates/goose/src/prompt_template.rs
@@ -104,7 +104,6 @@ pub fn render_template<T: Serialize>(name: &str, context: &T) -> Result<String, 
 }
 
 pub fn get_template(name: &str) -> Option<Template> {
-    // Find description in registry
     let (_, description) = TEMPLATE_REGISTRY.iter().find(|(n, _)| *n == name)?;
 
     let default_content = CORE_PROMPTS_DIR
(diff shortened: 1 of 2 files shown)


crates/goose/src/prompt_template.rs
```diff
@@ -1,240 +1,237 @@
...
+        "desktop_prompt.md",
+        "Additional context provided when using the desktop application",
+    ),
+    (
+        "desktop_recipe_instruction.md",
+        "Prompt template for recipe instructions in desktop mode",
+    ),
+    (
+        "plan.md",
+        "Prompt used when goose creates step-by-step plans. CLI only",
```
alexhancock, line 40: not specifically related to this changeset but why is this CLI only?

alexhancock (approved):
looks like a good improvement and I like the configurability!

I do wonder if we should now include edited template content in diagnostic reports. It feels like we cannot guarantee good behavior as much once people start editing content of these.

crates/goose/src/prompt_template.rs
DOsinga: I don't know. the CLI has a plan mode, the desktop does not

DOsinga:
> looks like a good improvement and I like the configurability!
> 
> I do wonder if we should now include edited template content in diagnostic reports. It feels like we cannot guarantee good behavior as much once people start editing content of these.

they should be part of the llm logs, but yeah, should be easy enough to do

PR #6422, opened 2026-01-09
Swap most references of Block -> AAIF

jamadeo:
I had Opus do a swap, paying attention to anything that would break something


katzdave:
@alexhancock, @jamadeo still working on this?

jamadeo:
> @alexhancock, @jamadeo still working on this?

@katzdave yeah, but it's awaiting some blockers for the actual repo move, and looks like we've got our fair share of conflicts, so I'll close and we can reopen either with this or a new one.

Issue #6406, opened 2026-01-08
Goose session tabs (single window)


seanwarner-tfh:
Here's a draft feature request for multiple session tabs in Goose Desktop:

---

**Please explain the motivation behind the feature request.**

This feature would solve the significant workflow disruption of managing multiple concurrent projects or conversation contexts within Goose Desktop. Currently, users must either:
- Lose context by switching topics within a single session
- Run multiple Goose Desktop instances, which consumes excessive system resources and creates window management overhead

This feature would unlock several key use cases:
- **Multi-project development**: Managing different codebases, documentation projects, or research tasks simultaneously without context switching penalties
- **Parallel AI team coordination**: Running separate conversations with different AI agents or teams (like the MAIteam system) for specialized work streams
- **Context preservation**: Maintaining long-running conversations about specific topics while being able to quickly switch to urgent queries or different project phases
- **Enhanced productivity**: Reducing the cognitive overhead of context switching and the time lost in re-establishing conversation context

**Describe the solution you'd like**

Implement a tabbed interface within Goose Desktop similar to modern browsers or IDEs:
- **Tab bar** at the top of the window showing session names/titles
- **New tab button** (+) to create fresh sessions
- **Tab management**: Right-click context menu for renaming, closing, duplicating tabs
- **Session persistence**: Tabs maintain their conversation history and context when switching between them
- **Visual indicators**: Show active/busy states when AI is processing in background tabs
- **Keyboard shortcuts**: Standard tab navigation (Ctrl+Tab, Ctrl+Shift+Tab, Ctrl+W to close, Ctrl+T for new tab)

Optional enhancements:
- Tab grouping for related conversations
- Session templates for common workflows
- Export/import tab sessions for sharing or backup

**Describe alternatives you've considered**

1. **Multiple Goose Desktop instances**: Resource-intensive and creates window management complexity
2. **Context switching within single session**: Loses conversation history and requires manual context re-establishment
3. **External session management**: Using external tools to track conversation contexts, but this adds friction and breaks the integrated workflow
4. **Conversation recipes**: While helpful for structured workflows, they don't solve the fundamental multi-context problem

**Additional context**

This feature would be particularly valuable for:
- Professional developers managing multiple repositories or projects
- Users working with multi-agent AI systems (like MAIteam) where different agents handle specialized tasks
- Long-form research projects where different topics need separate conversation threads
- Teams that need to maintain separate contexts for different stakeholders or project phases

[...]

Issue #4437, opened 2025-08-29
UI for managing multiple agent tasks in one place


yissachar:
**Please explain the motivation behind the feature request.**
I run a lot of agents concurrently on separate tasks. Currently in the UI I need a new window for each agent I'm running. If using the CLI then I need a new terminal window/tab. Either way, it's hard to keep track of everything in progress, and to know whether the agent is still in progress or waiting for more input.

**Describe the solution you'd like**
A UI where I can see all my current sessions and easily switch between them in one window. Indicator on each session if it's in progress or finished (waiting for further input). Description of the session (can be auto-generated based on the initial prompt), and show the working dir. This is probably a UI only feature, not for CLI.

- [x] I have verified this does not duplicate an existing feature request

zanesq:
Similar to https://github.com/smtg-ai/claude-squad

DOsinga:
we're actively working on this!

PR #6428, opened 2026-01-09
Multi chat

zanesq:
## Summary
Multi chat in one window working with existing sidebar navigation.

- Shows up to 10 recent chats on the left side with link to session history.
- Active sessions are simply hidden from view using css to keep running in the background.
- Scrollable sidebar to view bottom menu items.
- Session names are defaulted to "New Chat X" until they are automatically renamed from the backend using basic polling.
- Progress indicators for in progress, done and error states (cleared upon session selection in sidebar).
- Handles delete and renames from session history
- Handles edits/forks
- Shows chef icon for recipes (we can add to these icons later for other types of chats/forks etc)

closes https://github.com/block/goose/issues/4437
closes https://github.com/block/goose/issues/6406

https://github.com/user-attachments/assets/f7669af0-8683-4e3a-8ef0-658992fb92f6

### DIFF
--- a/ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
+++ b/ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
@@ -171,9 +171,10 @@ const SessionList = React.memo<{
     const nextIds = nextProps.sessions.map((s) => s.id).join(',');
     if (prevIds !== nextIds) return false;
 
-    // Check if any session name changed
+    // Check if any session name or message_count changed
     for (let i = 0; i < prevProps.sessions.length; i++) {
       if (prevProps.sessions[i].name !== nextProps.sessions[i].name) return false;
+      if (prevProps.sessions[i].message_count !== nextProps.sessions[i].message_count) return false;
     }
 
     // Check if any session's status has changed
@@ -345,7 +346,11 @@ const AppSidebar: React.FC<SidebarProps> = ({ currentPath }) => {
       const { sessionId, newName } = (event as CustomEvent<{ sessionId: string; newName: string }>)
         .detail;
       setRecentSessions((prev) =>
-        prev.map((s) => (s.id === sessionId ? { ...s, name: newName } : s))
+        prev.map((s) =>
+          s.id === sessionId
+            ? { ...s, name: newName, message_count: Math.max(s.message_count, 1) }
+            : s
+        )
       );
     };
 
--- a/ui/desktop/src/hooks/useChatStream.ts
+++ b/ui/desktop/src/hooks/useChatStream.ts
@@ -274,10 +274,20 @@ export function useChatStream({
   // Refs for values needed in callbacks without causing re-renders
   const abortControllerRef = useRef<AbortController | null>(null);
   const lastInteractionTimeRef = useRef<number>(Date.now());
+  const namePollingRef = useRef<ReturnType<typeof setTimeout> | null>(null);
 
   // Ref to access latest state in callbacks (avoids stale closures)
   const stateRef = useRef(state);
   stateRef.current = state;
+
+  useEffect(() => {
+    return () => {
+      if (namePollingRef.current) {
+        clearTimeout(namePollingRef.current);
+        namePollingRef.current = null;
+      }
+    };
+  }, [sessionId]);
 
   useEffect(() => {
     if (state.session) {
@@ -287,6 +297,11 @@ export function useChatStream({
 
   const onFinish = useCallback(
     async (error?: string): Promise<void> => {
+      if (namePollingRef.current) {
+        clearTimeout(namePollingRef.current);
+        namePollingRef.current = null;
+      }
+
       dispatch({ type: 'STREAM_FINISH', payload: error });
 
       const timeSinceLastInteraction = Date.now() - lastInteractionTimeRef.current;
@@ -452,6 +467,53 @@ export function useChatStream({
       // Emit session-created event for first message in a new session
       if (!hasExistingMessages && hasNewMessage) {
         window.dispatchEvent(new CustomEvent(AppEvents.SESSION_CREATED));
+
+        // Start polling for session name update during streaming
+        // The backend generates the name in parallel with the response
+        const pollForName = async (attempts = 0) => {
+          if (attempts >= 20) return; // Max 20 attempts (10 seconds)
+
+          try {
+            const response = await getSession({
+              path: { session_id: sessionId },
+              throwOnError: true,
+            });
+            const currentState = stateRef.current;
+            const currentName = currentState.session?.name;
+            const newName = response.data?.name;
+
+            // Check if name has changed from the initial name
+            if (newName && newName !== currentName) {
+              dispatch({
+                type: 'SET_SESSION',
+                payload: currentState.session
+                  ? { ...currentState.session, name: newName }
+                  : undefined,
+              });
+              window.dispatchEvent(
+                new CustomEvent(AppEvents.SESSION_RENAMED, {
+                  detail: { sessionId, newName },
+                })
+              );
+              return; // Stop polling once name is updated
+            }
+          } catch {
+            // Silently continue polling
+          }
+
+          // Continue polling if still streaming
+          const latestState = stateRef.current;
+          if (
+            latestState.chatState === ChatState.Streaming ||
+            latestState.chatState === ChatState.Thinking ||
+            latestState.chatState === ChatState.Compacting
+          ) {
+            namePollingRef.current = setTimeout(() => pollForName(attempts + 1), 500);
+          }
+        };
+
+        // Start polling after a short delay to give backend time to start name generation
+        namePollingRef.current = setTimeout(() => pollForName(0), 1000);
       }
 
       const newMessage = hasNewMessage


crates/goose/src/session/session_manager.rs
```diff
@@ -1139,7 +1139,7 @@ impl SessionStorage {
                    s.provider_name, s.model_config_json,
                    COUNT(m.id) as message_count
             FROM sessions s
-            INNER JOIN messages m ON s.id = m.session_id
+            LEFT JOIN messages m ON s.id = m.session_id
```
zanesq: Left sidebar needs to show sessions without messages for "New Chat"

ui/desktop/src/components/ChatInput.tsx
```diff
@@ -569,28 +565,38 @@ export default function ChatInput({
...
+  const minTextareaHeight = 38;
+
   const debouncedAutosize = useMemo(
     () =>
       debounce((element: HTMLTextAreaElement) => {
-        element.style.height = '0px'; // Reset height
+        // Store current scroll position to prevent jump
+        const scrollTop = element.scrollTop;
+
+        // Temporarily set to auto to measure natural height, but use minHeight to prevent collapse
```
zanesq, line 587: Fixed a pre-existing bug where the chat input placeholder was disappearing/jumping when starting a new chat

ui/desktop/src/components/ChatInput.tsx
```diff
@@ -1361,21 +1361,44 @@ export default function ChatInput({
...
+                          ? 'Recording...'
+                          : isTranscribing
+                            ? 'Transcribing...'
+                            : chatState === ChatState.RestartingAgent
+                              ? 'Restarting session...'
+                              : !hasSubmittableContent
+                                ? 'Type a message to send'
+                                : 'Send'}
+                  </p>
+                </TooltipContent>
```
DOsinga, line 1407: let's factor this out into a separate function that can then just do if ... return. this is getting out of hand! unless you want to do a datadriven approach here

crates/goose/src/session/session_manager.rs
DOsinga: hmm, yeah, but we can't do this like this. we create sessions without messages too often and it will just overwhelm all the overviews that we have.

alternatives: make this a flag to this method? probably end up with a convenience method for the current behavior.

or do this at the client. the client knows about the started sessions that don't have messages yet.

ui/desktop/src/App.tsx
```diff
@@ -58,53 +59,28 @@ const HubRouteWrapper = () => {
...
 
 const PairRouteWrapper = ({
-  chat,
-  setChat,
+  activeSessions,
 }: {
-  chat: ChatType;
-  setChat: (chat: ChatType) => void;
+  activeSessions: Array<{ sessionId: string; initialMessage?: string }>;
+  setActiveSessions: (sessions: Array<{ sessionId: string; initialMessage?: string }>) => void;
```
DOsinga, line 71: this suggests the client approach - we can just add to the active sessions when they don't have messages yet

ui/desktop/src/components/SessionIndicators.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Visual indicators for session status
+ * - Error: red alert icon (highest priority)
+ * - Streaming: animated spinner with GPU acceleration
+ * - Has activity: green dot (shown after streaming completes until viewed)
+ */
+export const SessionIndicators = React.memo<SessionIndicatorsProps>(
+  ({ isStreaming, hasUnread, hasError, className = '' }) => {
+    if (!isStreaming && !hasUnread && !hasError) {
```
DOsinga: this can be more readable if it is just an if ... else construct

if hasError: return <alert>
if isStreaming return <spinner>
if hasUnread: return <round thing>
return null

ui/desktop/src/sessions.ts
```diff
@@ -1,13 +1,42 @@
...
+  if (!name) return false;
+
+  // Check for backend default name pattern "New session X" or frontend default "New Chat"
+  const backendDefaultPattern = /^New session \d+$/i;
+  return (
+    name === DEFAULT_CHAT_TITLE ||
+    name.startsWith(DEFAULT_CHAT_TITLE) ||
+    backendDefaultPattern.test(name)
+  );
+}
```
DOsinga, line 13: this is rather fragile; I don't think we should do it like this. alternatives: the client already knows that this is a newly started conversation so could handle it like that. Or just change what we call these on the server. New Session (x) was just for debugging

ui/desktop/src/utils/navigationUtils.ts
```diff
@@ -41,9 +41,16 @@ export const createNavigationHandler = (navigate: NavigateFunction) => {
       case 'chat':
         navigate('/', { state: options });
         break;
-      case 'pair':
-        navigate('/pair', { state: options });
+      case 'pair': {
+        // Build URL with search params if resumeSessionId is provided
+        const searchParams = new URLSearchParams();
+        if (options?.resumeSessionId) {
+          searchParams.set('resumeSessionId', options.resumeSessionId);
```
DOsinga, line 51: this seems unrelated to the current change and makes me feel uneasy; we play fast and loose with routing here already. maybe we can let this one slide if we pinkie promise to clean this up once and for all?

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -129,48 +157,57 @@ function BaseChatContent({
       .reverse();
   }, [messages]);
 
+  // Auto-submit initial message for new sessions or forked sessions
```
DOsinga: I don't think we can do it this way. this has the different scenarios, the multi facetted if statements that try to mimic a specific situation the app might be useRef global variables that we used to have in the recipe manager with all the hard to read code and hard to main aspects of it. we need to rethink what we're doing here.

ui/desktop/src/components/ui/optimized-spinner.tsx
```diff
@@ -0,0 +1,69 @@
...
+  sm: 'w-4 h-4',
+  md: 'w-6 h-6',
+  lg: 'w-8 h-8',
+  xl: 'w-10 h-10',
+};
+
+/**
+ * GPU-accelerated spinner component for smooth performance
+ * Uses CSS transforms and containment for optimal rendering
+ */
```
DOsinga: this is not true. the implementation below is misleading. The translateZ(0), will-change, backface-visibility, perspective stuff is outdated advice from ~2012. Modern browsers composite CSS animations efficiently without hints. Worse, will-change can actually hurt performance if overused (it reserves GPU memory).

I don't think this needs to be separate component at all.

ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -172,6 +174,28 @@ export function useChatStream({
...
 
+  // Keep refs in sync with state for use in callbacks
+  // Note: We also update these synchronously when setting state to avoid stale closures
+  useEffect(() => {
+    sessionRef.current = session;
+  }, [session]);
+
+  useEffect(() => {
+    chatStateRef.current = chatState;
+  }, [chatState]);
```
DOsinga: yeah this is getting out of hand. we now have 4 refs for stale state? we have to refactor this. maybe useReducer

crates/goose/src/session/session_manager.rs
zanesq: ok good call, changed back and now tracked on the client instead

ui/desktop/src/hooks/useChatStream.ts
zanesq: good call changed to use a reducer for all these states

ui/desktop/src/utils/navigationUtils.ts
zanesq: Its actually needed so the sidebar can look up the current session and for the refresh case. Open to suggestions butnot really seeing another option here its pretty standard to use a query param for navigation related ids and refresh. Added comments to make it a bit clearer why we do it.

ui/desktop/src/components/BaseChat.tsx
zanesq: I pulled this into a hook to be more manageable for now

zanesq:
@DOsinga thanks for the feedback! Agreed on a lot of those points and refactored, pls take a second look. Ultimately we need to add a global state library to reduce our components and get rid of the window events and conditionals and refs everywhere. I will follow up with that next to make all this a lot more manageable but I think this is good enough for now with the current architecture to get the feature live.

crates/goose-server/src/routes/agent.rs
```diff
@@ -205,8 +204,7 @@ async fn start_agent(
         }
     }
 
-    let counter = state.session_counter.fetch_add(1, Ordering::SeqCst) + 1;
-    let name = format!("New session {}", counter);
+    let name = "New Chat".to_string();
```
zanesq, line 211: Since the naming and numbering was mainly for debugging it made sense to unify it with what we want to display in the ui sidebar rather than trying to detect and paper over it in the display.

ui/desktop/src/components/UserMessage.tsx
```diff
@@ -255,13 +253,6 @@ export default function UserMessage({ message, onMessageUpdate }: UserMessagePro
             </div>
           </div>
         )}
-
-        {/* Edited indicator */}
```
zanesq, line 259: Removed this edited indicator since it was only front end and would be lost on refresh anyway.. I think if we want to keep it we should integrate it with the backend but ultimately it doesn't provide much value to the user imo.

ui/desktop/src/App.tsx
```diff
@@ -134,45 +120,33 @@ const PairRouteWrapper = ({
...
+  // Add resumed session to active sessions if not already there
   useEffect(() => {
-    if (sessionId && capturedInitialMessage && sessionIdFromState) {
-      const timer = setTimeout(() => {
-        setCapturedInitialMessage(undefined);
-      }, 100);
-      return () => clearTimeout(timer);
+    if (resumeSessionId && !activeSessions.some((s) => s.sessionId === resumeSessionId)) {
+      window.dispatchEvent(
+        new CustomEvent('add-active-session', {
```
alexhancock: I might group all these string event names into a module that exports symbols that can be reused across the codebase, so we don't ever typo one.

Something like

```
export default {
  ADD_ACTIVE_SESSION: 'add-active-session'
  ...
}
```

or something even better with an enum perhaps?

ui/desktop/src/App.tsx
```diff
@@ -366,11 +340,82 @@ export function AppInner() {
...
+
+      setActiveSessions((prev) => {
+        const existingIndex = prev.findIndex((s) => s.sessionId === sessionId);
+
+        let updated: Array<{ sessionId: string; initialMessage?: string }>;
+
+        if (existingIndex !== -1) {
+          // Session exists - move to end and update initialMessage if provided
+          const existing = prev[existingIndex];
+          const updatedSession = initialMessage ? { ...existing, initialMessage } : existing;
```
alexhancock: why does this modify the `initialMessage`?

ui/desktop/src/App.tsx
zanesq: good call will update 👍

zanesq: done but touching a lot of files now so would like to merge this sooner than later

ui/desktop/src/App.tsx
zanesq: good catch it was from my testing and wasn't needed anymore, removed

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -55,6 +53,7 @@ interface BaseChatProps {
   showPopularTopics?: boolean;
   suppressEmptyState: boolean;
   sessionId: string;
+  isActiveSession?: boolean;
```
DOsinga: this is a pattern the AIs like very much - optional booleans that then get a default assigned. let's watch out for it! in general, no optional booleans unless undefined has a particular meaning (in which case we'd probably want to ask ourselves, do we need an enum instead?)

ui/desktop/src/App.tsx
```diff
@@ -366,11 +341,77 @@ export function AppInner() {
...
+          updated = [...prev.slice(0, existingIndex), ...prev.slice(existingIndex + 1), existing];
+        } else {
+          const newSession = { sessionId, initialMessage };
+          updated = [...prev, newSession];
+        }
+
+        // LRU eviction: keep only the last MAX_ACTIVE_SESSIONS
+        if (updated.length > MAX_ACTIVE_SESSIONS) {
+          updated = updated.slice(updated.length - MAX_ACTIVE_SESSIONS);
+        }
```
DOsinga, line 379: we can move this in the if - if we don't append there's no reason to slice. also, if we move the existing session, we drop the initialMessage on the floor. is that intended?

ui/desktop/src/App.tsx
```diff
@@ -366,11 +341,77 @@ export function AppInner() {
...
+    return () => {
+      window.removeEventListener(
+        AppEvents.ADD_ACTIVE_SESSION,
+        handleAddActiveSession as unknown as (event: Event) => void
+      );
+      window.removeEventListener(
+        AppEvents.CLEAR_INITIAL_MESSAGE,
+        handleClearInitialMessage as unknown as (event: Event) => void
+      );
+    };
```
DOsinga, line 402: these type cast strike me as rather ugly; wouldn't you say that doing the casting inside the event handlers would be better? or use a type map.

ui/desktop/src/App.tsx
```diff
@@ -529,22 +580,31 @@ export function AppInner() {
...
+        isProcessingRef.current = true;
+        console.log('[App] Processing initial message from launcher:', initialMessage);
+        navigate('/pair', {
+          state: {
+            initialMessage,
+          },
+        });
+        setTimeout(() => {
+          isProcessingRef.current = false;
+        }, 1000);
```
DOsinga, line 595: wow. timeout and a useRef? that's a tad smelly. now that we have multi-chat, can we not just send the message to the chat and then we get to pair it will already be there?

ui/desktop/src/components/sessions/SessionListView.tsx
```diff
@@ -416,6 +419,12 @@ const SessionListView: React.FC<SessionListViewProps> = React.memo(
       setSessions((prevSessions) =>
         prevSessions.map((s) => (s.id === sessionId ? { ...s, name: newDescription } : s))
       );
+      // Notify sidebar of the rename
```
DOsinga: are you? we're just telling anybody, no? in fact we should update the title of the current conversation too - there's an issue for that

ui/desktop/src/components/SessionIndicators.tsx
```diff
@@ -0,0 +1,47 @@
+import { AlertCircle, Loader2 } from 'lucide-react';
+import React from 'react';
+
+interface SessionIndicatorsProps {
+  isStreaming?: boolean;
+  hasUnread?: boolean;
+  hasError?: boolean;
+  className?: string;
```
DOsinga: none of these should be optional - we use this thing once - either they are set or not needed (like for classname I think)

ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
```diff
@@ -102,19 +93,169 @@ const menuItems: NavigationEntry[] = [
...
+  if (!shouldShowNewChatTitle(session)) {
+    return session.name;
+  }
+  // If session has a recipe, show the recipe title
+  if (session.recipe?.title) {
+    return session.recipe.title;
+  }
+  // Otherwise show default "New Chat"
+  return DEFAULT_CHAT_TITLE;
+};
```
DOsinga, line 102: this can all do without the comments

ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
```diff
@@ -102,19 +93,169 @@ const menuItems: NavigationEntry[] = [
...
+const SessionList = React.memo<{
+  sessions: Session[];
+  activeSessionId: string | undefined;
+  getSessionStatus: (
+    sessionId: string
+  ) => { streamState: string; hasUnreadActivity: boolean } | undefined;
+  onSessionClick: (session: Session) => void;
+}>(
+  ({ sessions, activeSessionId, getSessionStatus, onSessionClick }) => {
+    // Sort sessions so empty new chats always appear at the top
```
DOsinga: is that so? shouldShowNewChatTitle doesn't just check for no messages. also shouldn't these be sorted on the last modified date of the last message?

ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
```diff
@@ -131,6 +272,122 @@ const AppSidebar: React.FC<SidebarProps> = ({ currentPath }) => {
...
+
+          const shouldContinue = pollCount < maxPolls && (sessionWithDefaultName || pollCount < 5);
+
+          if (shouldContinue) {
+            const timeoutId = setTimeout(pollForUpdates, pollIntervalMs);
+            pollingTimeouts.push(timeoutId);
+          } else {
+            isPolling = false;
+          }
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
DOsinga: why this dissable?

ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
```diff
@@ -131,6 +272,122 @@ const AppSidebar: React.FC<SidebarProps> = ({ currentPath }) => {
...
+                if (seen.has(s.id)) return false;
+                seen.add(s.id);
+                return true;
+              })
+              .slice(0, 10);
+          });
+
+          const sessionWithDefaultName = apiSessions.find((s) => shouldShowNewChatTitle(s));
+
+          const shouldContinue = pollCount < maxPolls && (sessionWithDefaultName || pollCount < 5);
```
DOsinga, line 321: this is confusing to me, why keep polling if the name has changed?

ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
```diff
@@ -131,6 +272,122 @@ const AppSidebar: React.FC<SidebarProps> = ({ currentPath }) => {
...
+        return;
+      }
+
+      isPolling = true;
+      const pollIntervalMs = 300;
+      const maxPollDurationMs = 10000;
+      const maxPolls = maxPollDurationMs / pollIntervalMs;
+      let pollCount = 0;
+
+      const pollForUpdates = async () => {
```
DOsinga, line 295: so this whole polling buseiness is always suspicious. I get why we do this, since the backend changes the name out of the critical path. we should be able to solve this with a message in the stream though. can be done in a follow up. :pinky_swear:

ui/desktop/src/hooks/useAutoSubmit.ts
```diff
@@ -0,0 +1,103 @@
...
+    sessionId,
+    messages.length,
+    chatState,
+    clearInitialMessage,
+  ]);
+
+  return {
+    hasAutoSubmitted: hasAutoSubmittedRef.current,
+  };
+}
```
DOsinga, line 103: I don't know man, I feel like we might be making this whole pending message business way too complicated and react is just getting in our way here.

what if we just had something super simple like a global(!) variable that is a map from sessionId to pending message and then when you reach BaseChat, we check that, send that message out and clear the entry in the map?

ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -540,35 +643,39 @@ export function useChatStream({
...
     setChatState,
     handleSubmit,
     submitElicitationResponse,
     stopStreaming,
     setRecipeUserParams,
-    tokenState,
+    tokenState: state.tokenState,
     notifications: notificationsMap,
     onMessageUpdate,
   };
```
DOsinga, line 743: I'm still not wild about this. I think we can turn this in a true multi chat state component that just listens to multiple conversations, but I guess that is out of scope for this.

ui/desktop/src/vite-env.d.ts
```diff
@@ -45,11 +45,24 @@ declare module '*.md?raw' {
...
+      sessionId: string;
+      initialMessage?: string;
+    }>;
+    'clear-initial-message': CustomEvent<{
+      sessionId: string;
+    }>;
+    responseStyleChanged: CustomEvent;
+    'session-created': CustomEvent<{ session?: import('./api').Session }>;
+    'session-deleted': CustomEvent<{ sessionId: string }>;
+    'session-renamed': CustomEvent<{ sessionId: string; newName: string }>;
```
DOsinga, line 64: huh. if we have a window event map, why the casting on top?

ui/desktop/src/components/BaseChat.tsx
```diff
@@ -115,6 +105,40 @@ function BaseChatContent({
...
+    initialMessage,
+    handleSubmit,
+  });
+
+  useEffect(() => {
+    let streamState: 'idle' | 'loading' | 'streaming' | 'error' = 'idle';
+    if (chatState === ChatState.LoadingConversation) {
+      streamState = 'loading';
+    } else if (
+      chatState === ChatState.Streaming ||
```
alexhancock, line 127: should we just define a new attribute on `ChatState`?

ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
zanesq: Comment was confusing so I removed it. It just relies on the backend which is sorted by last modified already or puts new chat on top.

ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
zanesq: the name gets updated again as the session progresses but yeah agreed its confusing, I'll come back to this in another pr with a refactor to move to streaming and remove the polling like you mentioned in the other comment

ui/desktop/src/components/sessions/SessionListView.tsx
zanesq: ok removed comment, it does update the title also, I merged in that issue/fix from main already and verified its working

ui/desktop/src/hooks/useAutoSubmit.ts
zanesq: Good point but we might be able to handle this better with a global state library so I will follow up with that also

ui/desktop/src/hooks/useChatStream.ts
zanesq: noted, will consider in the upcoming refactor to global state

ui/desktop/src/App.tsx
zanesq: yeah this is hacky, I tried using the existing system and it would load multiple chat sessions so its to get around the quick launcher sending multiple events from the other electron window.. will come back to this also

ui/desktop/src/components/BaseChat.tsx
zanesq: ChatState doesn't quite fit because ChatState is an enum for UI state transitions, not a data container but good point about simplifying the data flow. I'll look at this with the global state update next. Also good suggestion about better multi chat rust integration, I'll think about what that looks like and get a plan together.

zanesq:
Sounds good will merge asap after some final testing. The goose logo was an upstream change that added it to the sidebar so I moved it to the top right above chat in my latest commit.

Not seeing the only one session issue will try to repro that real quick.

PR #6431, opened 2026-01-11
feat: add Microsoft Foundry provider

andrey-moor:
## Summary

Add support for Microsoft Foundry (Azure AI Foundry) as a new provider, enabling access to multiple AI models through Azure infrastructure.

### Features
- Supports Claude models via Anthropic API format
- Supports GPT, Mistral, Llama, and DeepSeek models via OpenAI API format
- Automatic API routing based on model name prefix
- Full streaming support for both API formats
- Standard error handling and retry logic

### Configuration
| Variable | Required | Secret | Description |
|----------|----------|--------|-------------|
| `MICROSOFT_FOUNDRY_RESOURCE` | Yes | No | Azure resource name |
| `MICROSOFT_FOUNDRY_API_KEY` | Yes | Yes | API key from Azure portal |

### Supported Models
- **Claude**: claude-sonnet-4-5 (default), claude-haiku-4-5 (fast), claude-opus-4-5, claude-opus-4-1
- **OpenAI**: gpt-4o, gpt-4o-mini
- **Others**: mistral-large, llama-3.1-70b, deepseek-v3

## Type of Change
- [x] New feature (non-breaking change that adds functionality)

## AI Assistance
- [x] This PR was created with AI assistance

## Testing
- Unit tests for model detection and metadata validation
- Manual testing with real Microsoft Foundry credentials
- Verified Claude model responses via API
- Build and clippy checks pass

## References
- [Microsoft Foundry Claude Documentation](https://learn.microsoft.com/en-us/azure/ai-foundry/foundry-models/how-to/use-foundry-models-claude)
- [Azure AI Model Inference](https://learn.microsoft.com/en-us/azure/ai-foundry/model-inference/concepts/endpoints)

### DIFF
--- a/crates/goose/src/config/declarative_providers.rs
+++ b/crates/goose/src/config/declarative_providers.rs
@@ -97,6 +97,9 @@ pub struct DeclarativeProviderConfig {
     /// Explicit default model (instead of first in models list).
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub default_model: Option<String>,
+    /// Optional authentication configuration (used by routed providers).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub auth: Option<AuthConfig>,
 }
 
 impl DeclarativeProviderConfig {
@@ -224,6 +227,7 @@ pub fn create_custom_provider(
         routes: None,
         config_keys: None,
         default_model: None,
+        auth: None,
     };
 
     let custom_providers_dir = custom_providers_dir();
@@ -279,6 +283,7 @@ pub fn update_custom_provider(
             routes: existing_config.routes,
             config_keys: existing_config.config_keys,
             default_model: existing_config.default_model,
+            auth: existing_config.auth,
         };
 
         let file_path = custom_providers_dir().join(format!("{}.json", id));
@@ -536,6 +541,7 @@ fn build_route_config(
         routes: None, // Single-engine config for the route
         config_keys: base.config_keys.clone(),
         default_model: base.default_model.clone(),
+        auth: route.auth.clone(),
     })
 }
 
--- a/crates/goose/src/providers/anthropic.rs
+++ b/crates/goose/src/providers/anthropic.rs
@@ -112,9 +112,24 @@ impl AnthropicProvider {
             base_path
         };
 
-        let auth = AuthMethod::ApiKey {
-            header_name: "x-api-key".to_string(),
-            key: api_key,
+        // Use auth config if provided, otherwise default to x-api-key header
+        let auth = if let Some(auth_config) = &config.auth {
+            if auth_config.auth_type == "header" {
+                AuthMethod::ApiKey {
+                    header_name: auth_config.name.clone(),
+                    key: api_key,
+                }
+            } else {
+                AuthMethod::ApiKey {
+                    header_name: "x-api-key".to_string(),
+                    key: api_key,
+                }
+            }
+        } else {
+            AuthMethod::ApiKey {
+                header_name: "x-api-key".to_string(),
+                key: api_key,
+            }
         };
 
         let api_client =
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -153,15 +153,32 @@ impl OpenAiProvider {
         } else {
             format!("{}://{}", url.scheme(), url.host_str().unwrap_or(""))
         };
-        let base_path = url.path().trim_start_matches('/').to_string();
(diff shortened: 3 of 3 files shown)


bioinfornatics:
@andrey-moor, to my knowledge, there are two ways for end users to consume an LLM in Azure AI Foundry:

- API key authentication
- Microsoft Entra ID authentication

With your PR, do you support both authentication methods?

DOsinga:
I don't know how this all works, but we do have an azure provider currently. Does this add/replace that? Why do we need both?

andrey-moor:
@DOsinga @bioinfornatics  The existing azure provider is for Azure OpenAI Service (dedicated OpenAI deployments). This new provider is for Microsoft AI Foundry's (ex Azure AI Foundry) Model-as-a-Service offerings, which include Claude, Mistral, Llama, and DeepSeek in addition to GPT models. They use different endpoints and APIs.

Currently this PR supports API key authentication only. Entra ID (DefaultAzureCredential) support could be added in a follow-up PR by reusing the existing AzureAuth infrastructure from the Azure OpenAI provider. I wanted to keep this initial PR focused, but happy to add Entra ID support if we feel this is needed before we merge.

andrey-moor:
@DOsinga @alexhancock I saw the declarative formats, but the current system has some limitations for this case:
  1. Single engine - Foundry (and I believe we'll find similar patterns with other cloud providers) needs Anthropic format for Claude, OpenAI for others
  2. Static base_url - Foundry needs https://{resource}.services.ai.azure.com/... where resource comes from env var
  3. Single endpoint - Claude uses /anthropic/v1/messages, others use /models/chat/completions

To support this declaratively, at minimum we'd need to add URL templating and model-prefix-based routing. Happy to contribute that if that's what we need, but it's a bigger architectural change.

alexhancock:
@andrey-moor Thanks for looking into it!

My view is the more providers we can enable declaratively, the better. If we ran into limitations that prevented this one being implemented that way, fixing the underlying limitations will unblock us and likely enable more providers to be added declaratively in future as well.

Would you be willing to look at URL templating? Let's think about model routing and if there is a clean way to map different models <-> different paths.

DOsinga:
expanding how we do declarative providers would definitely be preferred. 

https://github.com/block/goose/pull/6396 seems related and has ideas of what we could add here

andrey-moor:
Agreed, declarative feels like the right way to scale out the provider support. Let me think about it a little bit and I propose a potential solution to it

crates/goose/src/config/declarative_providers.rs
```diff
@@ -19,18 +20,66 @@ pub fn custom_providers_dir() -> std::path::PathBuf {
...
+    /// Header name for the API key (e.g., "x-api-key", "api-key", "Authorization")
     pub name: String,
+}
+
+/// Configuration for a single route that maps models to an engine and endpoint.
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct RouteConfig {
+    /// Pattern to match model names against
+    #[serde(rename = "match")]
+    pub match_pattern: ModelMatch,
```
DOsinga, line 53: the comment for this field appears three times in the code. also not sure it warrants its own subtype.

DOsinga (commented):
sorry for taking so long. I like the general ideas here, but I feel the routing is a bridge too far. could we not for now just have two different declarative providers here, one for anthropic and one for openai?

bioinfornatics:
Hi maintainers , jumping in as a user.

Just wanted to share that this functionality would be very valuable on our side. We’re currently blocked to use Azure Foundry + Entra ID, and this PR seems to address exactly that gap.
Finally Azure Foundry is like `Open Router`


I realize reviews take time, but I wanted to signal real-world demand for this feature. Happy to provide feedback or validation from a user perspective if that helps move things forward.

Thanks for your work on Goose!

DOsinga:
thanks @bioinfornatics - that's very valuable

so I like this too and I think builiding out the declarative providers is really important. it is just that the routing component of this seems like a big change that feels a tad unnatural

my instinct would be to remove that for now and just say, if you want to access a provider based on conditions, just create two declarative providers. that would make it possible to use this functionality. we could then follow up if needed with something like a meta provider where you can virtually merge one or more providers and you just pick the model and the system than figures out what provider you really need. that could be useful in general.

(actually maybe that should be the default, just select a model and the system figures out which provider you mean)

does this make sense?

DOsinga:
ok, how do we want to progress here? can we do two providers and expand the custom provider setup?

katzdave:
Closing for now as its been a couple weeks, but feel free to re-open.

PR #6439, opened 2026-01-12
Fix: Small update in how ML-based prompt injection determines final result

dorien-koelemeijer:
## Summary
Small update in how ML-based prompt injection detection, evaluating conversation context, impacts what findings will be flagged and which ones won't. To reduce false positives, if the prompt injection evaluation of conversation context results in a SAFE result, only flag tool call findings that have a CRITICAL result (just to be on the safe side). 

This means:
- Safe context + non-critical pattern → suppressed
- Safe context + critical pattern → flagged
- Unsafe context + any pattern → flagged 
(if above threshold that user set of course)

### Type of Change
- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [x] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Local testing using `just run-ui` and various test cases


crates/goose/src/security/scanner.rs
```diff
@@ -169,12 +189,37 @@ impl PromptInjectionScanner {
...
+                max_pattern_risk = ?tool_result.pattern_matches.first().map(|m| &m.threat.risk_level),
+                "Suppression conditions met: safe context + non-critical patterns"
+            );
+
+            DetailedScanResult {
+                confidence: 0.0,
+                pattern_matches: Vec::new(),
+                ml_confidence: context_result.ml_confidence,
+            }
+        } else if tool_result.confidence >= context_result.confidence {
```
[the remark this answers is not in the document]
dorien-koelemeijer: This is the entire point of the PR? Maybe i'm misunderstanding copilot, but the idea is to reduce false positives for the time being and generally increase sensitivity again later on. Same goes for comment below

crates/goose/src/security/scanner.rs
```diff
@@ -59,6 +59,23 @@ impl PromptInjectionScanner {
...
+        if model_name.is_none() {
+            if let Ok(mapping_json) = std::env::var("SECURITY_ML_MODEL_MAPPING") {
+                if let Ok(mapping) = serde_json::from_str::<
+                    crate::security::classification_client::ModelMappingConfig,
+                >(&mapping_json)
+                {
+                    if let Some(first_model) = mapping.models.keys().next() {
+                        tracing::info!(
+                            default_model = %first_model,
+                            "SECURITY_ML_MODEL_MAPPING available but no model selected - using first available model as default"
```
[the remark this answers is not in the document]
dorien-koelemeijer: Not really a critical comment

crates/goose/src/security/scanner.rs
```diff
@@ -314,4 +359,132 @@ mod tests {
         assert!(result.is_malicious);
         assert!(result.explanation.contains("Security threat"));
     }
+
+    #[test]
```
DOsinga: this is 130 lines of code to essentially test 2 booleans and how they interact. Delete

crates/goose/src/security/scanner.rs
```diff
@@ -169,12 +189,37 @@ impl PromptInjectionScanner {
...
+                .all(|m| m.threat.risk_level != crate::security::patterns::RiskLevel::Critical);
+
+        if context_is_safe && tool_has_only_non_critical {
+            tracing::info!("Suppressing non-critical pattern match due to safe context evaluation");
+            tracing::debug!(
+                context_ml_confidence = ?context_result.ml_confidence,
+                pattern_count = tool_result.pattern_matches.len(),
+                max_pattern_risk = ?tool_result.pattern_matches.first().map(|m| &m.threat.risk_level),
+                "Suppression conditions met: safe context + non-critical patterns"
+            );
```
DOsinga: do you want to info & debug log here?

dorien-koelemeijer:
> given that we increased the cut-off for pattern matching, does this make a change

It will slightly, if conversation context seems fully safe, I don't think it's necessary to alert on a tool call. However, it's a critical tool call finding, we'd probably want to safe rather than sorry (it's not only for pattern matching, but also going forward when we expand on the pattern matching)

crates/goose/src/security/scanner.rs
dorien-koelemeijer: Good point, this was helpful for local testing but can probably get rid of these for now, ty!

Issue #6405, opened 2026-01-08
CLI headless recipe runs fail with 'Scheduler not available' error on Linux AARCH64 (v1.18.0+)


blackgirlbytes:
## Bug Description

When running recipes in headless mode via CLI on Linux AARCH64, tool calls consistently fail with the error:

```
-32603: Scheduler not available. This tool only works in server mode.
```

Additionally, some extensions (like `mcp-kubernetes`) fail with:
```
-32603: Transport closed
```

Note: `filesystem` extension works fine, and interactive mode works correctly.

## Environment
- **OS:** DGX Spark Linux AARCH64
- **Mode:** CLI headless (`goose --recipe xxx`)
- **LLM Provider:** Ollama
- **Workload:** ~20 recipes with 9-10 subrecipes running concurrently
- **Affected versions:** v1.18.0, v1.19.1
- **Working version:** v1.15.0

## Steps to Reproduce
1. Install goose v1.18.0 or later on Linux AARCH64
2. Run a recipe in headless mode: `goose --recipe <recipe_file>`
3. Observe the scheduler error when tool calls are made

## Expected Behavior
Recipe should execute successfully as it does in:
- Interactive mode (same system)
- v1.15.0 headless mode

## Actual Behavior
- Tool calls fail with `-32603: Scheduler not available` error
- Some extensions fail with `-32603: Transport closed`

## Relevant Code
The error originates from [schedule_tool.rs#L28](https://github.com/block/goose/blob/d97b6a52e8e2923c8a71e8459995876af31b8a61/crates/goose/src/agents/schedule_tool.rs#L28):

```rust
let scheduler = match self.scheduler_service.lock().await.as_ref() {
    Some(s) => s.clone(),
    None => {
        return Err(ErrorData::new(
            ErrorCode::INTERNAL_ERROR,
            "Scheduler not available. This tool only works in server mode.".to_string(),
            None,
        ))
    }
};
```

## Analysis
It appears that in CLI headless mode, the `scheduler_service` is not being initialized, causing the schedule tool to fail. This regression was introduced somewhere between v1.15.0 and v1.18.0.

The issue may be related to how the agent is initialized differently in CLI mode vs server mode, where the scheduler service is expected but not available.

The concurrent workload (9-10 subrecipes running simultaneously) and use of Ollama as the LLM provider may also be relevant factors.

## Additional Context
- Same recipe works perfectly on v1.15.0
- Interactive mode always works (even on affected versions)
- Issue is consistent and reproducible
[...]

PR #6442, opened 2026-01-12
Fix: exclude platform_schedule_tool in CLI

lifeizhou-ap:
## Summary
Fixed "Scheduler not available" error in  issue https://github.com/block/goose/issues/6405

Changes:
1. Exclude schedule tool in CLI mode - Only add platform__manage_schedule tool when scheduler is available. In CLI mode, there's no persistent scheduler service running, so the tool would fail if the LLM tried to use it.

2. Unified schedule file - CLI and Desktop now both use schedule.json (was using different files). Using schedule.json because Desktop users already have data in this file for scheduled job.  For `schedules.json` created by CLI, the data in this file are not used to execute the recipe with schedule, it only used in `run now` which triggers the job manually. Losing this data has very minimal user impact.

3.  Refactoring: create_scheduler() helper to reduce duplication

4. Improved tool description - Added "Goose" to description and all actions to prevent confusion with external schedulers

### Type of Change

- [ ] Feature
- [X] Bug fix
- [X] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing

Unit test and Manually Testing

### DIFF
--- a/crates/goose/tests/agent.rs
+++ b/crates/goose/tests/agent.rs
@@ -130,7 +130,7 @@ mod tests {
                 .description
                 .clone()
                 .unwrap_or_default()
-                .contains("Manage Goose's internal scheduled recipe execution"));
+                .contains("Manage goose's internal scheduled recipe execution"));
         }
 
         #[tokio::test]
@@ -163,7 +163,7 @@ mod tests {
                 .description
                 .clone()
                 .unwrap_or_default()
-                .contains("Manage Goose's internal scheduled recipe execution"));
+                .contains("Manage goose's internal scheduled recipe execution"));
 
             // Verify the tool has the expected actions in its schema
             if let Some(properties) = tool.input_schema.get("properties") {
@@ -464,7 +464,7 @@ mod tests {
                 config: ExtensionConfig::Platform {
                     name: "todo".to_string(),
                     description:
-                        "Enable a todo list for Goose so it can keep track of what it is doing"
+                        "Enable a todo list for goose so it can keep track of what it is doing"
                             .to_string(),
                     bundled: Some(true),
                     available_tools: vec![],


crates/goose/src/agents/platform_tools.rs
```diff
@@ -7,18 +7,18 @@ pub fn manage_schedule_tool() -> Tool {
...
-            - "sessions": List execution history for a job
+            - "list": List all Goose scheduled jobs
+            - "create": Create a new Goose scheduled job from a recipe file
+            - "run_now": Execute a Goose scheduled job immediately
+            - "pause": Pause a Goose scheduled job
+            - "unpause": Resume a paused Goose scheduled job
+            - "delete": Remove a Goose scheduled job
+            - "kill": Terminate a currently running Goose scheduled job
+            - "inspect": Get details about a running Goose scheduled job
+            - "sessions": List execution history for a Goose scheduled job
```
DOsinga: goose should not be capitalized. why is this change happening?

crates/goose/src/agents/reply_parts.rs
```diff
@@ -481,9 +481,7 @@ mod tests {
         let (tools, _toolshim_tools, _system_prompt) =
             agent.prepare_tools_and_prompt(&working_dir).await?;
 
-        // Ensure both platform and frontend tools are present
```
DOsinga, line 484: or just delete this test

scripts/test_compaction.sh
```diff
@@ -164,8 +164,10 @@ export GOOSE_AUTO_COMPACT_THRESHOLD=0.005
 
 OUTPUT=$(mktemp)
 
-echo "Step 1: Creating session with first message..."
-(cd "$TESTDIR" && "$GOOSE_BIN" run --text "hello" 2>&1) | tee "$OUTPUT"
+LONG_RESPONSE_PROMPT="Count from 1 to 200, one number per line."
+
+echo "Step 1: Creating session with first message (generating tokens for threshold)..."
+(cd "$TESTDIR" && "$GOOSE_BIN" run --text "$LONG_RESPONSE_PROMPT" 2>&1) | tee "$OUTPUT"
```
DOsinga, line 170: what happened here? from a different branch maybe?

DOsinga (approved):

crates/goose/src/agents/platform_tools.rs
lifeizhou-ap: I am sorry that I forgot to mention this change in the PR summary.  In the original issue, user has some prompts related to k8s that could be related to schedule and agent triggers the platform_schedule_tool.  I guess the agent got confused with k8s scheduler with goose scheduler.   That's why I added `goose` in the tool description, so that the agent won't use this tool when having the context with non goose schedule.  

I will change to the `goose` instead  of `Goose`

scripts/test_compaction.sh
lifeizhou-ap: Because we removed the platform_schedule_tool from CLI, the context usage is getting smaller and the original test prompt cannot reach to the context size to trigger the compact.  So I changed the prompt to generate a bit longer response so that it reached to the threshold of 0.5% to trigger the compact

crates/goose/src/agents/reply_parts.rs
lifeizhou-ap: I checked the code, it seems it still worthwhile to use this test to test the ordering of the tools.  I changed the test name to reflect the purpose of this test.

PR #6445, opened 2026-01-12
feat: Native PKCE Auth for Tetrate via Custom URL Scheme

raj-subhankar:
## Summary

This PR moves Tetrate authentication from a server-driven flow (Rust PKCE + localhost callback) to a client-driven flow (Electron PKCE + goose:// deep-link callback). 
It also adds a /tetrate/verify endpoint so the Electron client can send the authorization code and PKCE verifier for server-side token exchange.

                                                           
### Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

https://github.com/user-attachments/assets/c864cd49-886c-4e00-b36c-40a4e22837f3

After:   

https://github.com/user-attachments/assets/1ed3caf2-f417-4721-b3c4-6192e5c8618e

### DIFF
--- a/ui/desktop/src/components/ProviderGuard.tsx
+++ b/ui/desktop/src/components/ProviderGuard.tsx
@@ -40,6 +40,7 @@ export default function ProviderGuard({ didSelectProvider, children }: ProviderG
   const [isTetrateSetupInProgress, setIsTetrateSetupInProgress] = useState(false);
   const onboardingTracked = useRef(false);
   const tetrateSetupRunId = useRef(0);
+  const tetrateSuccessTimerId = useRef<ReturnType<typeof setTimeout> | null>(null);
   const scrollContainerRef = useRef<HTMLDivElement>(null);
   const [showScrollIndicator, setShowScrollIndicator] = useState(true);
 
@@ -69,6 +70,7 @@ export default function ProviderGuard({ didSelectProvider, children }: ProviderG
     title: string;
     message: string;
     showProgress?: boolean;
+    showSuccess?: boolean;
     showRetry: boolean;
     closeLabel?: string;
     autoClose?: number;
@@ -107,9 +109,20 @@ export default function ProviderGuard({ didSelectProvider, children }: ProviderG
       setIsTetrateSetupInProgress(false);
 
       if (result.success) {
-        setTetrateSetupState(null);
-        setSwitchModelProvider('tetrate');
-        setShowSwitchModelModal(true);
+        setTetrateSetupState({
+          show: true,
+          title: 'Setup Complete',
+          message: 'Your account has been connected successfully.',
+          showSuccess: true,
+          showProgress: false,
+          showRetry: false,
+        });
+        tetrateSuccessTimerId.current = setTimeout(() => {
+          tetrateSuccessTimerId.current = null;
+          setTetrateSetupState(null);
+          setSwitchModelProvider('tetrate');
+          setShowSwitchModelModal(true);
+        }, 1500);
       } else {
         trackOnboardingSetupFailed('tetrate', result.message);
         setTetrateSetupState({
@@ -165,6 +178,16 @@ export default function ProviderGuard({ didSelectProvider, children }: ProviderG
   };
 
   const handleCancelTetrateSetup = async () => {
+    if (tetrateSuccessTimerId.current) {
+      // User dismissed the success modal early — still advance to model selection
+      clearTimeout(tetrateSuccessTimerId.current);
+      tetrateSuccessTimerId.current = null;
+      setTetrateSetupState(null);
+      setSwitchModelProvider('tetrate');
+      setShowSwitchModelModal(true);
+      return;
+    }
+
     if (!isTetrateSetupInProgress) {
       setTetrateSetupState(null);
       return;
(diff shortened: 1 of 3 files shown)


ui/desktop/package.json
```diff
@@ -101,6 +105,7 @@
     "@electron-forge/plugin-fuses": "^7.10.2",
     "@electron-forge/plugin-vite": "^7.10.2",
     "@electron/fuses": "^1.8.0",
+    "@electron/node-gyp": "git+https://github.com/electron/node-gyp.git#06b29aafb7708acef8b3669835c8a7857ebc92d2",
```
johnlanda: We could use a published version here rather than a git commit. 

Latest is `10.2.0-electron.2` https://www.npmjs.com/package/@electron/node-gyp

ui/desktop/src/native/auth_session/binding.gyp
```diff
@@ -0,0 +1,19 @@
+{
+  "targets": [
+    {
+      "target_name": "auth_session",
+      "sources": ["src/auth_session.mm"],
+      "defines": ["NAPI_DISABLE_CPP_EXCEPTIONS"],
+      "xcode_settings": {
+        "CLANG_ENABLE_OBJC_ARC": "YES",
+        "MACOSX_DEPLOYMENT_TARGET": "12.0",
```
johnlanda: Let's make sure this is aligned with the Mac OSX versions supported by goose.

crates/goose-server/src/routes/setup.rs
```diff
@@ -67,6 +74,43 @@ async fn start_openrouter_setup() -> Result<Json<SetupResponse>, StatusCode> {
...
+    ),
+)]
+async fn verify_tetrate_setup(
+    Json(request): Json<TetrateVerifyRequest>,
+) -> Result<Json<SetupResponse>, StatusCode> {
+    let api_key =
+        match TetrateAuth::exchange_code_with_verifier(request.code, request.code_verifier).await {
+            Ok(api_key) => api_key,
+            Err(e) => {
+                return Ok(Json(SetupResponse {
```
johnlanda, line 80: Should we return a different status code here rather than a 200 on failure?

ui/desktop/src/native/auth_session/src/auth_session.mm
```diff
@@ -0,0 +1,247 @@
...
+  napi_threadsafe_function tsfn;
+} AuthRequest;
+
+typedef struct AuthResult {
+  bool success;
+  std::string message;
+} AuthResult;
+
+static NSMutableSet<ASWebAuthenticationSession *> *gActiveSessions = nil;
+static GooseAuthPresentationContextProvider *gPresentationContextProvider = nil;
```
johnlanda: Will these globals cause issues if multiple auth flows are triggered at the same time? Is it possible to trigger multiple at the same time?

michaelneale:
oh interesting, looks like some merge conflicts (most are generated code though). I didn't know about this on macos - is there an example of this? seems a fair bit of work, but macos is pretty common so may be ok (OTOH spinning up a little server isn't the end of the world if smooth enough)

crates/goose-server/src/routes/setup.rs
raj-subhankar: Returning 200 with success: false was an existing behavior and matches the pattern used by the Openrouter setup flow.

ui/desktop/src/native/auth_session/src/auth_session.mm
raj-subhankar: ASWebAuthenticationSession only supports a single login session at a time, so I think this should be fine.

ui/desktop/src/native/auth_session/binding.gyp
raj-subhankar: Goose uses Electron 39, which has a minimum macOS 12 requirement, so this should be fine.

raj-subhankar:
hey @michaelneale 👋

Yeah, spinning up a server for auth is totally fine. The main goal here was to keep the user in the app, instead of opening a browser tab and leaving them there. To make that work, we had to change the server based approach as well. 
Claude Desktop comes to mind as a good example of this.

Also I’ve fixed the conflicts and updated the PR description.

michaelneale:
yeah this did work when I tried it - is nice. May be overkill as I think previous was was similarly smooth (but with native browser I guess)

raj-subhankar:
@michaelneale I agree this is a significant change, and for most cases the previous login flow did feel smooth.

That said, I do run into issues on when using Goose alongside local dev projects. Since Goose spins up the login server on localhost:3000, if you already have something running on that port the login just fails silently, with no indication of why.

While we could fix this by not hardcoding the port, the new flow addresses it as well and has the added benefit of feeling more native.

michaelneale:
@raj-subhankar is there a way to do it with a dynamic port (or this macos native way) but use the native browser vs a window?

raj-subhankar:
@michaelneale Yes, we could do this with a custom URL scheme using the external browser.

However, I’m advocating for ASWebAuthenticationSession because it’s the best of both worlds.

To clarify: on macOS this uses the user’s default browser (Chrome, Safari, etc.), not an embedded webview, and it shares the browser’s cookies, so you still get the same SSO experience.

The “window” you see is the OS delegating the auth request to the default browser. The key advantage is lifecycle management: the system lets us automatically dismiss the window as soon as authentication completes.

With a external browser flow, the user is left with a “Success” tab they have to manually close before returning to the app.

Given that this is Apple’s recommended approach for native app authentication, and the standard used by apps like Slack and VS Code, is there a specific UX concern you have with the modal window?

michaelneale:
I think we can close this, with a port fix and possibly https://github.com/block/goose/pull/7166 for responding to low balance I think we could be in a good place.

PR #6446, opened 2026-01-12
Spell check setting

### LINKED ISSUE #6423: Spell Check
There is no option to remove spell check. This is incredibly annoying having spell check constantly underline every word I type. Add an option to remove the spell check that is always on. The spell check makes the app unusable. Thanks.

DOsinga:
## Summary

fixes #6423

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -1421,7 +1421,6 @@ ipcMain.handle('get-wakelock-state', () => {
   }
 });
 
-// Handle spellcheck setting
 ipcMain.handle('set-spellcheck', async (_event, enable: boolean) => {
   try {
     const settings = loadSettings();


ui/desktop/src/main.ts
```diff
@@ -1421,6 +1421,29 @@ ipcMain.handle('get-wakelock-state', () => {
   }
 });
 
+// Handle spellcheck setting
```
alexhancock: i imagine it will give you great satisfaction to remove this comment before merging :P

alexhancock (approved):

ui/desktop/src/main.ts
DOsinga: haha thanks. to be honest I am wavering. it feels like a fight I am losing and dumb comments aren't as bad as dumb tests maybe?

DOsinga: `goose run --text "look at the current branch and compare with main. remove any comments that were introduced that were introduce that just say what the code underneath is doing, i.e. commenting on what or how and not on why."`

PR #6448, opened 2026-01-12
Safe npm package upgrades and removed unused ai package

zanesq:
## Summary
Upgraded safe packages and noticed we aren't using the `ai` package so removed it.

Looks like hey api changed their imports again also.

Tested and verified locally


zanesq:
Putting in draft while I look into the typescript errors from upgrade

PR #6458, opened 2026-01-13
Standalone mcp apps

DOsinga:
## Summary

makes it possible to launch standalone apps that are declared as MCP servers

image

### DIFF
--- a/crates/goose-server/src/routes/agent.rs
+++ b/crates/goose-server/src/routes/agent.rs
@@ -998,7 +998,10 @@ async fn list_apps(
 
         for extension_name in active_extensions {
             if let Err(e) = cache.delete_extension_apps(&extension_name) {
-                warn!("Failed to clean cache for extension {}: {}", extension_name, e);
+                warn!(
+                    "Failed to clean cache for extension {}: {}",
+                    extension_name, e
+                );
             }
         }
 
--- a/ui/desktop/src/hooks/useChatStream.ts
+++ b/ui/desktop/src/hooks/useChatStream.ts
@@ -276,9 +276,8 @@ export function useChatStream({
         });
         setChatState(ChatState.Idle);
 
-        // Populate apps cache in the background (don't await)
         listApps({
-          throwOnError: false,
+          throwOnError: true,
           query: { session_id: sessionId },
         }).catch((err) => {
           console.warn('Failed to populate apps cache:', err);


michaelneale:
@DOsinga got a demo app to poke around with you like to show?

ui/desktop/src/api/types.gen.ts
```diff
@@ -364,6 +364,17 @@ export type GetToolsQuery = {
...
+export type GooseApp = {
+    description?: string | null;
+    height?: number | null;
+    html: string;
+    mcpServer?: string | null;
+    name: string;
+    resizable?: boolean | null;
+    resourceUri: string;
+    width?: number | null;
+};
```
aharvard: For awareness, MCP Apps are conforming to MCP Resources schema via `UIResource`, see `UIResource` [documented here](https://github.com/modelcontextprotocol/ext-apps/blob/db1b48552b0b74f12aadd54d61fe2c36b34792c5/specification/draft/apps.mdx#ui-resource-format).

Is `GooseApp` a superset of `UIResource`? If so, can we align our type definition to conform? 

I think properties like `height`, `mcpServer` (can we infer this?), `width`, `resizable` probably belong under a `_meta.ui.goose` key... or defined in a smart way that work well with `UIResource`.

... and properties like `description`, `name`, `html` (as `text` or `blob`) are already standard fields in the MCP resource schema.

aharvard:
@michaelneale, once you connect to an MCP server with apps (I used https://mcp-app-bench.onrender.com/mcp), their app resources appear. See screenshot — I have one MCP server that I've set up as a bench testing tool, and all the relevant UI resources show up as standalone apps. It’s not exactly my mental model of what _an_ MCP App is, but it’s still pretty cool!

This raises the question: should we ship a very simple goose app so users can try it out? Maybe a clock app?

image

DOsinga:
I have a standard stock quote app that I use. it currently has the access token baked in, so need to clean that up, but it allows you to get, eh, stock quotes.

we could quickly publish a little clock thing too - it wouldn't really do all that much in terms of toolcalling of course (timezone?) but how we store that is still left as an exercise for the reader here

our we do this as is and quickly follow up with a goose platform MPC that handles this elegantly. including a clock in that would work easy

aharvard:
> we could quickly publish a little clock thing too

just whipped up some HTML, [clock.html](https://github.com/user-attachments/files/24621995/clock.html)

https://github.com/user-attachments/assets/b069bc7a-3239-4b1e-80a6-e75aa1d4f8bf

aharvard:
> a standard stock quote app 

would be pretty sweet as well

ui/desktop/src/api/types.gen.ts
DOsinga: this is a good point. the GooseApp struct is inherited from a previous version of GooseApps that predates the McpApp story.  it does need the mcpServer so it can make toolcalls back (an embedded mcp app is inside a tool result, so it knows what mcp server created it).  I would like a future where an MCP app can have more than one mcp server actually

DOsinga:
I refactored this a bit, taking the mcp resource thing in as @aharvard suggested - my preference would be to merge this soon-ish and quickly follow up with step 4 of the plan. thoughts?

we should also think about how we do state management for these. I'm thinking something like, we associate a session with each app and it always renders the last toolcall in that session and then there is a config box which all it does is send a message to the agent asking for a toolcall.

that way if you had like @aharvard 's clock example, the toocall could set the type of clock and if you hit configure you could just tell goose, show me a swiss clock type of thing

aharvard (approved):

aharvard:
> my preference would be to merge this soon-ish and quickly follow up with step 4 of the plan

@DOsinga, agreed — and I would like to consider us refining how the apps page presents stuff to users... we can iterate on that layout easily enough over time

> we should also think about how we do state management for these. I'm thinking something like, we associate a session with each app and it always renders the last toolcall in that session and then there is a config box which all it does is send a message to the agent asking for a toolcall.

I think i follow, last tool call makes sense... also FYI context mutation is something I'd like to land for basic MCP Apps soonish... might want to leverage that as well somehow  https://github.com/block/goose/issues/6472

> that way if you had like @aharvard 's clock example, the toocall could set the type of clock and if you hit configure you could just tell goose, show me a swiss clock type of thing

yeah, worth exploring statefullness w/ local storage as well (how I did it in my clock HTML)

PR #6465, opened 2026-01-13
feat(ui): support prefersBorder option for MCP Apps

aharvard:
## Summary

Adds support for the `_meta.ui.prefersBorder` option in MCP App resources. When an MCP App returns `prefersBorder: false` in its resource metadata, the UI wrapper border and rounded corners are removed, allowing the app to render edge-to-edge.

## Changes

- Read `prefersBorder` from `_meta.ui` in the resource response (defaults to `true`)
- Conditionally apply border/rounded styling based on this preference
- Refactored resource state into a single `ResourceData` object for cleaner code

## Usage

MCP Apps can now specify border preference in their resource response:

```json
{
  "uri": "ui://widget/example.html",
  "mimeType": "text/html;profile=mcp-app",
  "_meta": {
    "ui": {
      "prefersBorder": false
    }
  },
  "text": "<!doctype html>..."
}
```

## Before / After

| Before | After |
|--------|-------|
| image | image |

*Before: MCP App with default border and rounded corners*
*After: MCP App with `prefersBorder: false` - no wrapper border*

### DIFF
--- a/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
+++ b/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
@@ -193,16 +193,6 @@ export default function McpAppRenderer({
     );
   }
 
-  if (!resource.html) {
-    return (
-      <div className={cn('p-4 bg-bgApp border border-borderSubtle rounded-lg')}>
-        <div className="flex items-center justify-center" style={{ minHeight: '200px' }}>
-          Loading MCP app...
-        </div>
-      </div>
-    );
-  }
-
   return (
     <div
       className={cn(
@@ -210,7 +200,7 @@ export default function McpAppRenderer({
         resource.prefersBorder ? 'border border-borderSubtle rounded-lg' : 'my-6'
       )}
     >
-      {proxyUrl ? (
+      {resource.html && proxyUrl ? (
         <iframe
           ref={iframeRef}
           src={proxyUrl}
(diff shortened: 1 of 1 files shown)


ui/desktop/src/components/McpApps/McpAppRenderer.tsx
```diff
@@ -174,15 +187,20 @@ export default function McpAppRenderer({
...
     return (
-      <div className="mt-3 p-4 border border-red-500 rounded-lg bg-red-50 dark:bg-red-900/20">
+      <div className="p-4 border border-red-500 rounded-lg bg-red-50 dark:bg-red-900/20">
         <div className="text-red-700 dark:text-red-300">Failed to load MCP app: {error}</div>
       </div>
     );
   }
 
-  if (!resourceHtml) {
+  if (!resource.html) {
```
DOsinga: consider unifying the effective two arms of the if now that the payload is growing, i.e. 

const toRender = resource.html ? <iframe> : "Loading MCP app..."

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
```diff
@@ -174,25 +187,20 @@ export default function McpAppRenderer({
...
   return (
-    <div className={cn('mt-3 bg-bgApp', 'border border-borderSubtle rounded-lg overflow-hidden')}>
-      {proxyUrl ? (
+    <div
+      className={cn(
+        'bg-bgApp overflow-hidden',
+        resource.prefersBorder ? 'border border-borderSubtle rounded-lg' : 'my-6'
+      )}
+    >
+      {resource.html && proxyUrl ? (
```
aharvard, line 203: @DOsinga PTAL

I unified `resource.html` and `proxyUrl` as we don't need a discrete state for `proxyUrl` loading 

Also, leaving this inline instead of extracting as a const. I noticed a hard-to-reproduce issue where the iframe ref _might_ have gotten lost or something... the rendered mcp app would be sitting there w/ no messages being sent down. Maybe a closure issue? Not sure. But I think what I got works.

PR #6475, opened 2026-01-13
Ask for permission in the CLI

DOsinga:
## Summary

On the CLI we don't ask to track. this addresses that.

### DIFF
--- a/crates/goose-cli/src/commands/configure.rs
+++ b/crates/goose-cli/src/commands/configure.rs
@@ -40,8 +40,6 @@ pub async fn handle_configure() -> anyhow::Result<()> {
     }
 }
 
-/// Prompt user for telemetry consent.
-/// Returns true if user opted in, false if opted out.
 pub fn configure_telemetry_consent_dialog() -> anyhow::Result<bool> {
     let config = Config::global();
 


crates/goose-cli/src/commands/configure.rs
```diff
@@ -39,16 +40,78 @@ pub async fn handle_configure() -> anyhow::Result<()> {
...
-async fn handle_first_time_setup(config: &Config) -> anyhow::Result<()> {
+pub fn configure_telemetry_consent_dialog() -> anyhow::Result<bool> {
+    let config = Config::global();
+
+    println!();
+    println!("{}", style("Help improve goose").bold());
+    println!();
+    println!(
+        "{}",
+        style("Would you like to help improve goose by sharing anonymous usage data?").dim()
```
jamadeo, line 51: How annoying would it be to have the text in just one place for CLI and UI?

DOsinga: I don't think it is worth it, tbh. We could make it into a larger effort and share more text of course, but bridging between the CLI and the react code, that's quite a bit apart.

PR #6478, opened 2026-01-13
fix: require auth when running goose on non loopback address

michaelneale:
Small change - loopback (localhost) etc is fine - but if exposed on the network this will ask for an auth-token to be set.

### DIFF
--- a/crates/goose-cli/src/cli.rs
+++ b/crates/goose-cli/src/cli.rs
@@ -832,10 +832,10 @@ enum Command {
         #[arg(long, help = "Authentication token to secure the web interface")]
         auth_token: Option<String>,
 
-        /// Allow running without authentication on non-loopback addresses (unsafe)
+        /// Allow running without authentication when exposed on the network (unsafe)
         #[arg(
             long,
-            help = "Skip auth requirement for non-loopback addresses (unsafe)"
+            help = "Skip auth requirement when exposed on the network (unsafe)"
         )]
         no_auth: bool,
     },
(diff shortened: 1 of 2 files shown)


michaelneale:
cc @DOsinga

crates/goose-cli/src/cli.rs
```diff
@@ -831,6 +831,13 @@ enum Command {
         /// Authentication token for both Basic Auth (password) and Bearer token
         #[arg(long, help = "Authentication token to secure the web interface")]
         auth_token: Option<String>,
+
+        /// Allow running without authentication on non-loopback addresses (unsafe)
+        #[arg(
+            long,
+            help = "Skip auth requirement for non-loopback addresses (unsafe)"
```
DOsinga: non-loopback addresses might be unclear to most users. maybe just say, expose server to local network without auth?

crates/goose-cli/src/commands/web.rs
```diff
@@ -125,12 +125,36 @@ async fn auth_middleware(
...
+        "localhost" | "127.0.0.1" | "::1" => true,
+        _ => {
+            if let Ok(addr) = host.parse::<std::net::IpAddr>() {
+                addr.is_loopback()
+            } else {
+                false
+            }
+        }
+    }
+}
```
DOsinga, line 133: fn is_loopback_address(host: &str) -> bool {
    host == "localhost" || host.parse::<std::net::IpAddr>().map_or(false, |a| a.is_loopback())
}

seems better. technically 12.1.2.3 is also loopback

michaelneale: woah what - it is? haha

michaelneale: must be a programmatic way to determin this

crates/goose-cli/src/cli.rs
michaelneale: yeah true

PR #6492, opened 2026-01-14
Dont show MCP UI/Apps until tool is approved

### LINKED ISSUE #6481: MCP Apps: UI shown before user approves tool call (Manual mode)
If a tool call that has an associated MCP App resource needs approval, then the behavior is strange: the UI is shown, and the tool call arguments are passed to it via a `ui/notifications/tool-input`, but the tool isn't actually called until the user approves it.

This is problematical on a couple of levels

 - It's very confusing to the user - after the model called the tool, something has already happened, but they are being asked to approve the tool call.
 - The UI resource is likely in a half-filled state - while this is a normal transient state, it's again going to be confusing to the user what is going on.
 - The resource can make external network connections as allowed in its configuration, and act on or exfiltrate the tool call input parameters

---

**To Reproduce**
1. Run https://github.com/modelcontextprotocol/ext-apps/tree/main/examples/wiki-explorer-server (`npm run dev`) and add it as a goose extension
2. Try to use it via the model
3. The UI is displayed, in a partially initialized state, and the permission request is after that
4. When the user approves the tool call, the UI finishes initialization and the model receives the tool call results
---

**Expected behavior**
The user's permission is requested *before* the UI resource is displayed.

---

**Screenshots**

Before approval:

Image

After approval:

Image

(OTOH if the UI only needs the tool inputs and not the response, like say, https://github.com/modelcontextprotocol/ext-apps/tree/main/examples/sheet-music-server then the user will be even more puzzled by the approval prompt.)

---

- **OS & Arch:** Any
- **Interface:** UI
- **Version:** v1.19.1
- **Extensions enabled:** None
- - **Provider & Model:** Any

---

DOsinga:
## Summary

fixes #6481

### DIFF
--- a/ui/desktop/src/components/ToolCallWithResponse.tsx
+++ b/ui/desktop/src/components/ToolCallWithResponse.tsx
@@ -55,6 +55,7 @@ interface ToolCallWithResponseProps {
   toolResponse?: ToolResponseMessageContent;
   notifications?: NotificationEvent[];
   isStreamingMessage?: boolean;
+  isPendingApproval: boolean;
   append?: (value: string) => void;
 }
 
@@ -106,10 +107,8 @@ function McpAppWrapper({
       ? requestWithMeta.toolCall.value.arguments
       : undefined;
 
-  // Memoize toolInput to prevent unnecessary re-renders
   const toolInput = useMemo(() => ({ arguments: toolArguments || {} }), [toolArguments]);
 
-  // Memoize toolResult to prevent unnecessary re-renders
   const toolResult = useMemo(() => {
     if (!toolResponse) return undefined;
     const resultWithMeta = toolResponse.toolResult as ToolResultWithMeta;
@@ -149,6 +148,7 @@ export default function ToolCallWithResponse({
   toolResponse,
   notifications,
   isStreamingMessage,
+  isPendingApproval,
   append,
 }: ToolCallWithResponseProps) {
   // Handle both the wrapped ToolResult format and the unwrapped format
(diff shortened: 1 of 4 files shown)


ui/desktop/src/components/ToolCallWithResponse.tsx
```diff
@@ -169,6 +169,8 @@ export default function ToolCallWithResponse({
     requestWithMeta._meta?.ui?.resourceUri || resultWithMeta?.value?._meta?.ui?.resourceUri
   );
 
+  const shouldShowMcpContent = !isPendingApproval;
```
aharvard, line 172: nit: maybe use `!isPendingApproval` directly for conditions since `content` is a meaningful term in the MCP spec? 

```ts
!isPendingApproval && !hasMcpAppResourceURI ...t( )
```

DOsinga: what do you mean with content is a meaningful term? I kinda like introducing variables like this to indicate intent. if you !isPendingApproval you have to think a moment, why do we only show this when we have no pending approval, while at the time of the variable introduction it just says, ah we should show content when there is no pending approvals. but if you think it is confusing, we can rename the var?

aharvard: MCP `Content` https://modelcontextprotocol.io/specification/2025-11-25/schema#content (heads up, page might no auto scroll to the right section) 

Yeah, intent indication sounds good. 

Maybe `shouldShowMcpApp` is clearer than `shouldShowMcpContent` and follows the pattern of `hasMcpAppResourceURI`?

PR #6493, opened 2026-01-14
fixed 0 token in openrouter steaming

lifeizhou-ap:
## Summary
Fix Issue  https://github.com/block/goose/issues/5907

**Root Cause**
When sending api request to openrouter via streaming mode, the response chunks with usage comes in the chunk without finish_reason. Example

```
- choices: [(content=Some(""), finish_reason=None, has_usage=true)], usage: Some({prompt_tokens: 5897, ...}
```

**Fix**
Skip the check for `finish_reason` to get the usage

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Manual testing and also tested open_ai, databricks and github copilot provider using the same open_ai logic

### DIFF
--- a/crates/goose/src/providers/formats/openai.rs
+++ b/crates/goose/src/providers/formats/openai.rs
@@ -609,8 +609,6 @@ where
                     msg = msg.with_id(id);
                 }
 
-                // Always yield usage when present - OpenRouter sends usage in a separate
-                // chunk that may have empty content and no finish_reason
                 yield (
                     Some(msg),
                     usage,


DOsinga:
you should ask the author of that line

lifeizhou-ap:
To be safe, I also tried Ollama, Xai and Tetrate providers which uses openai format, and it works.

I traced the origin of the logic, it was introduced in this PR [https://github.com/block/goose/pull/2677/](https://github.com/block/goose/pull/2677/changes). 

It was a while ago, maybe this is not valid any more.

jamadeo:
> To be safe, I also tried Ollama, Xai and Tetrate providers which uses openai format, and it works.
> 
> I traced the origin of the logic, it was introduced in this PR [https://github.com/block/goose/pull/2677/](https://github.com/block/goose/pull/2677/changes).
> 
> It was a while ago, maybe this is not valid any more.

oh yeah that was me. If only I could remember why I did it that way. Thanks for the fix :)

lifeizhou-ap:
Hi @jamadeo,

I found why the check was there :)  When I tested this PR, I only checked the used tokens for the context window. not the accumulated tokens to calculate the cost. 

fix: https://github.com/block/goose/pull/6587

PR #6497, opened 2026-01-14
fix: improve code execution mode tool handling

rabi:
## Summary
- Surface extension server instructions in system prompt by removing conditional code_execution_mode block from system.md
- Use prefixed tool names (code_execution__*) consistently in MOIM, server instructions, and tool descriptions to prevent model confusion
- Add explicit "NEVER use JSON.parse()" warnings since tool results are already parsed JavaScript objects
- Set RepetitionInspector limit to 5 to prevent infinite loops when model repeatedly calls same tools

### Type of Change
- [ ] Feature
- [x] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested locally with Gemini and hybrid models like Nemotron.

### DIFF
--- a/crates/goose/src/prompts/system.md
+++ b/crates/goose/src/prompts/system.md
@@ -5,8 +5,6 @@ goose uses LLM providers with tool calling capability. You can be used with diff
 claude-sonnet-4, o1, llama-3.2, deepseek-r1, etc).
 These models have varying knowledge cut-off dates depending on when they were trained, but typically it's between 5-10
 months prior to the current date.
-{% if not code_execution_mode %}
-
 # Extensions
 
 Extensions allow other applications to provide context to goose. Extensions connect goose to different data sources and
@@ -40,7 +38,6 @@ and extensionmanager__list_resources on this extension.
 
 {% else %}
 No extensions are defined. You should let the user know that they should add extensions.
-{% endif %}
 {% endif %}
 
 {% if extension_tool_limits is defined and not code_execution_mode %}
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -236,7 +236,8 @@ impl Agent {
         )));
 
         // Add repetition inspector (lower priority - basic repetition checking)
-        tool_inspection_manager.add_inspector(Box::new(RepetitionInspector::new(None)));
+        // Limit to 5 repetitions to allow model self-correction while preventing infinite loops
+        tool_inspection_manager.add_inspector(Box::new(RepetitionInspector::new(Some(5))));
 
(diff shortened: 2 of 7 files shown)


rabi:
Hey @alexhancock, PTAL. This makes code_execution work better with smaller hybrid models like Nemotron.

crates/goose/src/prompts/system.md
```diff
@@ -5,6 +5,23 @@ goose uses LLM providers with tool calling capability. You can be used with diff
 claude-sonnet-4, o1, llama-3.2, deepseek-r1, etc).
 These models have varying knowledge cut-off dates depending on when they were trained, but typically it's between 5-10
 months prior to the current date.
+{% if code_execution_mode %}
```
alexhancock: I worry that this prompt info is duplicative with:

Server instructions: https://github.com/block/goose/pull/6497/changes#diff-40aeaabae1c639df10ef13ae9edf99b906c8bcbf761b0270597d74284b5fb80aR436

Tool description: https://github.com/block/goose/pull/6497/changes#diff-40aeaabae1c639df10ef13ae9edf99b906c8bcbf761b0270597d74284b5fb80aR735

It looks like right now maybe those never make it into the system prompt because of the `{% if not code_execution_mode %}` block?

But I am confused because I thought in my original change I did have a way to surface this info to the model.

What do you think about refactoring this such that when code mode is active the server instructions + tool descriptions end up included in this, and we use that as the way of surfacing this info to the model vs hardcoding a new copy here.

rabi: Thanks Alex. Yeah, I think the tool descriptions are sent (as part of the tool spec), but the server instructions don't surface to the system prompt when code_execution_mode is true. 

The naming inconsistency i.e tools are prefixed (code_execution__execute_code) in tool list but server instructions and MOIM using unprefixed names (execute_code) confuses the model and it ends up many times making calls without prefix and going in a loop. Also, the model sees the tool signatures and wants to call them immediately, forgetting the code_execution wrapper is required.

I can change to surface the server instructions and test. However, I've seen https://github.com/block/goose/pull/6765 and it surely looks promising and may work better with smaller models.  if we decide to go with that, I can test that with Nemotron and abandon this PR.

rabi: I've updated the PR and with my local testing it seems to work well with Nemotron-3-nano.

rabi:
I've proposed a separate improvement [PR](https://github.com/block/goose/pull/7035) after testing with pctx and smaller models.

PR #6498, opened 2026-01-14
Create default gooseignore file when missing

mattjoyce:
## Summary
  - automatically create `.gooseignore` the first time we need the built-in safety patterns so the defaults are visible/
  editable instead of hidden in code
  - share those defaults through a single constant and cover them with an explicit test
  ## Testing
  - `cargo fmt`
  - `cargo test -p goose-mcp test_default_patterns_when_no_ignore_files`

### DIFF
--- a/crates/goose-mcp/src/developer/rmcp_developer.rs
+++ b/crates/goose-mcp/src/developer/rmcp_developer.rs
@@ -121,6 +121,27 @@ pub struct PromptArgumentTemplate {
 // Embeds the prompts directory to the build
 static PROMPTS_DIR: Dir = include_dir!("$CARGO_MANIFEST_DIR/src/developer/prompts");
 
+const DEFAULT_GOOSEIGNORE_CONTENT: &str = concat!(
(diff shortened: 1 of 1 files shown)


blackgirlbytes:
love that the default is there to keep average users from messing up their secrets files, but if theyre more advanced..they have the option to edit the ignore file.

mattjoyce:
Thinking about it, we can help user more.   I have added some guidance into the default .gooseignore, including some other examples of files that users might unintentionally expose.

zanesq:
@mattjoyce just realized that we should probably add a default `.gooseignore` to this project also or add it to `.gitignore` if we don't think we need one do you want to follow up with that change?

mattjoyce:
> @mattjoyce just realized that we should probably add a default `.gooseignore` to this project also or add it to `.gitignore` if we don't think we need one do you want to follow up with that change?

Done.  I think .gitignore is most robust.
https://github.com/block/goose/pull/6826

PR #6412, opened 2026-01-09
Avoid using cliclack.confirm in non-interactive session

lifeizhou-ap:
## Summary
Fixes Issue [#6236](https://github.com/block/goose/issues/6236)

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

- [X] This PR was created or reviewed with AI assistance

### Testing
Manual testing

PR #6502, opened 2026-01-14
allow --with-builtin extension on cli resume

Evanfeenstra:
## Summary
Included builtin extension names in the check_missing_extensions_or_exit check, so that you can add `--with-builtin` to cli call while also including `--resume`

The problem:

`goose run --with-builtin developer --output-format stream-json --name session1 -t "read the readme"`
`goose run --with-builtin developer --output-format stream-json --name session1 -t "summarize our convo" --resume`
results in: "Extension(s) developer from previous session are no longer available. Restore for this session?"

### Type of Change

- [ ] Feature
- [ x ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

- [ x ] This PR was created or reviewed with AI assistance

### Testing
Yes, I verified this works by running the flow locally from ./target/debug/goose, and the interative cli prompt did not showup

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:

### DIFF
--- a/crates/goose-cli/src/session/builder.rs
+++ b/crates/goose-cli/src/session/builder.rs
@@ -204,14 +204,13 @@ async fn offer_extension_debugging_help(
 
 fn check_missing_extensions_or_exit(
     saved_extensions: &[ExtensionConfig],
-    cli_extension_names: &HashSet<String>,
+    cli_builtin_names: &HashSet<String>,
     interactive: bool,
 ) {
     let missing: Vec<_> = saved_extensions
         .iter()
         .filter(|ext| {
-            !cli_extension_names.contains(&ext.name())
-                && get_extension_by_name(&ext.name()).is_none()
+            !cli_builtin_names.contains(&ext.name()) && get_extension_by_name(&ext.name()).is_none()
         })
         .cloned()
         .collect();
@@ -454,8 +453,8 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> CliSession {
         eprintln!("{}", style(format!("Warning: {}", warning)).yellow());
     }
 
-    // Collect CLI-provided extension names to avoid false "missing" warnings on resume
-    let cli_extension_names: HashSet<String> = session_config.builtins.iter().cloned().collect();
+    // Collect CLI-provided builtin names to avoid false "missing" warnings on resume
+    let cli_builtin_names: HashSet<String> = session_config.builtins.iter().cloned().collect();
 
     // If we get extensions_override, only run those extensions and none other
     let extensions_to_run: Vec<_> = if let Some(extensions) = session_config.extensions_override {
@@ -468,7 +467,7 @@ pub async fn build_session(session_config: SessionBuilderConfig) -> CliSession {
                 {
                     check_missing_extensions_or_exit(
                         &saved_state.extensions,
-                        &cli_extension_names,
+                        &cli_builtin_names,
                         session_config.interactive,
                     );
                     saved_state.extensions


Evanfeenstra:
maybe this is already solved with https://github.com/block/goose/pull/6412 .. but still, this should not be an error at all IMO

blackgirlbytes:
> maybe this is already solved with #6412 .. but still, this should not be an erro

I stand corrected from our thread in Discord. only half the problem is resolved with 6412. After looking at your PR, this would solve the problem fully

lifeizhou-ap:
Hi @Evanfeenstra,

Thank you for raising the issue and your contribution. 

I can still reproduce this issue in `interactive` mode.  I found the issue does not only happen to the `built-in` extensions, it also happens to ` --with-extension` cli arguments.   You can expand the condition that you added to include the extensions passed with `--with-extensions` and `--with-streamable-http-extension`

**However, we may have another option to fix this problem:** 
I am now curious about why we have `check_missing_extensions_or_exit` when resuming the session.  

- It only checks whether the original saved extensions exist in the goose config file
- the error message `Extension(s) {} from previous session are no longer available. Restore for this session?` seems confusing.  I copied the error message from the original code while fixing in this [PR](https://github.com/block/goose/pull/6412/changes.  I should have thought about this in-depth.   

The same prompt would also appear when resuming a recipe if the extension is specified in recipe but not in goose config file. 

Since we have extension per session now shall we remove it? cc @wpfleger96 @DOsinga

wpfleger96:
when resuming sessions, my opinion has been that I should always get the same set of extensions that I had in the original session (instead of resuming the session with your default set of extensions from `~/.config/goose/config.yaml`) - I originally started working toward this in #4331

the reason why you see `Extension(s) {} from previous session are no longer available. Restore for this session?` is because of the change I made in #5974 - goose is detecting that the previous session you're resuming was originally started with an ephemeral extension (`--with-builtin`/`--with-extension`/etc) and it's asking if you want to reload that ephemeral extension into the resumed session. this aligns with my same opinion: when resuming a session you should get the same extensions that were in the original session, NOT your default set of extensions in `~/.config/goose/config.yaml` 

I think I'm a little confused on the issue though. if I start a session with `--with-builtin developer`, then I later resume that session, based on the current code I would expect goose to interactively prompt and ask me if I want to restore the `developer` extension into my resumed session. are we saying that instead, goose should automatically restore the `developer` extension into the resumed session instead of prompting the user to confirm?

lifeizhou-ap:
> when resuming sessions, my opinion has been that I should always get the same set of extensions that I had in the original session (instead of resuming the session with your default set of extensions from `~/.config/goose/config.yaml`) - I originally started working toward this in #4331
> 
> the reason why you see `Extension(s) {} from previous session are no longer available. Restore for this session?` is because of the change I made in #5974 - goose is detecting that the previous session you're resuming was originally started with an ephemeral extension (`--with-builtin`/`--with-extension`/etc) and it's asking if you want to reload that ephemeral extension into the resumed session. this aligns with my same opinion: when resuming a session you should get the same extensions that were in the original session, NOT your default set of extensions in `~/.config/goose/config.yaml`
> 
> I think I'm a little confused on the issue though. if I start a session with `--with-builtin developer`, then I later resume that session, based on the current code I would expect goose to interactively prompt and ask me if I want to restore the `developer` extension into my resumed session. are we saying that instead, goose should automatically restore the `developer` extension into the resumed session instead of prompting the user to confirm?

Thank you @wpfleger96 for the reply. 

The function is checking whether the original saved extensions exist in the **goose config file**, not the extension in the resume session exists in the original session.  Please correct me if I am wrong. 

Current steps:
-  user runs `goose run --with-extension "npx -y @modelcontextprotocol/server-memory"  --output-format stream-json --name session1 -t "read the readme"`, the `server-memory` is persisted in the session.
- when user runs resume, the function checks whether `server-memory` exists in the extensions in the goose config. Since users does not have this extension in the config (which is common, same as recipe sessions), it will prompt user to confirm.

lifeizhou-ap:
Hi @Evanfeenstra 

@wpfleger96 and I had a sync chat and we both thought the `check_missing_extensions_or_exit` is redundant now.  I created a fix PR https://github.com/block/goose/pull/6759 to remove it and we won't have the user prompt any more. 

I have merged https://github.com/block/goose/pull/6759 to `main` branch, would you please check the `main` branch out to confirm the issue does not occur to you? If it works for you, would you please close this PR?   Thank you!

Evanfeenstra:
Thank you! makes is easier to automate stuff on servers or in CI pipelines. Worked on `main` for me

PR #6503, opened 2026-01-14
feat: advertise MCP Apps UI extension capability

aharvard:
## Summary

Implements early support for [SEP-1724](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1724) by advertising the `io.modelcontextprotocol/ui` extension capability during MCP initialization.

This allows goose to signal to MCP servers that it supports [MCP Apps](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx#client-host-capabilities).

## Changes

- Added `capabilities_extensions.rs` module with a `WithExtensions` HTTP client wrapper
- The wrapper implements `StreamableHttpClient` and intercepts `initialize` requests to inject the `extensions` field into capabilities
- HTTP-based extensions now use `WithExtensions(http_client)` in `extension_manager.rs`

## Why an HTTP client wrapper?

The `rmcp` crate's `ClientCapabilities` struct doesn't yet support the `extensions` field from SEP-1724. Rather than forking rmcp, this approach:
1. Wraps `reqwest::Client` and implements `StreamableHttpClient`
2. Intercepts `initialize` messages before they're sent
3. Converts to `serde_json::Value`, injects `extensions`, and sends as raw JSON body

This ensures the `extensions` field survives serialization (unlike a generic transport wrapper approach which would lose the field during rmcp's internal deserialization).

## Limitations

- **HTTP transports only**: This injection works for HTTP-based MCP extensions
- **Stdio transports**: Cannot inject extensions without upstream rmcp changes (noted in code comments)

## Initialize message now includes:

```json
{
  "method": "initialize",
  "params": {
    "capabilities": {
      "sampling": {},
      "elicitation": {},
      "extensions": {
        "io.modelcontextprotocol/ui": {
          "mimeTypes": ["text/html;profile=mcp-app"]
        }
      }
    }
  }
}
```

## Testing

- Unit tests verify the JSON injection works correctly
- Tests verify non-initialize messages are passed through unchanged
- Confirmed working via server logs showing the `extensions` field in initialize requests


aharvard:
This PR is not ready for review yet

aharvard:
Ok, it's ready now

crates/goose/src/agents/capabilities_extensions.rs
```diff
@@ -0,0 +1,250 @@
...
+        // MCP Apps UI extension
+        // https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx#client-host-capabilities
+        "io.modelcontextprotocol/ui": {
+            "mimeTypes": ["text/html;profile=mcp-app"]
+        }
+    })
+}
+
+#[cfg(test)]
+mod tests {
```
DOsinga, line 212: this is not testing anything

aharvard:
Good point, @DOsinga. 

I think [SEP 1724](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1724) is getting closer to being merged (I spotted a draft blog post announcing it, but no formal confirmation when). 

@alexhancock, if we lift this up to RMCP _before_ [SEP 1724](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1724) merges, would that be a problem?

Honestly, might be best to wait for the SEP to merge, ship in RMCP, then upgrade goose (instead of monkey patching).

DOsinga:
it probably is the best way, yeah. can we add experimental features in a non harmfull way to rmcp @alexhancock - this might come up more if we want to have goose be built on rmcp and be head of the standard sometimes.

aharvard:
I don't think we would need to do this now (since the SEPs for extensions and mcp apps is on near horizon)... in the future for other non-standard capablities we can leverage `ClientCapabilities["experimental"]` https://modelcontextprotocol.io/specification/2025-11-25/schema#clientcapabilities.

Does RMCP provide the `experimental` property?

jamadeo:
yep, rmcp does support https://docs.rs/rmcp/latest/rmcp/model/type.ExperimentalCapabilities.html

alexhancock:
Yes there are experimental capabilities and @bolinfest also helped add support for custom requests recently.  My hope is this gives us good primitives to do things that go beyond the spec with rmcp itself not having a lot of custom stuff.

aharvard:
Thanks for the discussion! I'll close this PR in favor of waiting for RMCP to provide goose with `capabilities.extensions` once SEP-1724 merges

Issue #6144, opened 2025-12-16
MCP Client connection is never properly closed


Ben-Lichtman:
**Describe the bug**

This is caused by an issue with the rmcp crate: https://github.com/modelcontextprotocol/rust-sdk/issues/572

ie. the connection to an MCP server is never getting formally closed, which can cause the session to stay open on the server side. If this is an authenticated connection this can bar the user from creating a new client until the old session expires.

On goose, this means that once I close the application, I get authentications errors when connecting to the same extension server again until the old session expires on the server side (can take a long time).

A workaround I have found for this is to call `.cancel()` on the client before dropping it, which appears to close the connection explicitly, though ideally this can be done automatically on Drop (async drop issues aside), and to actually close the transport, rather than cancelling it.

katzdave:
Maybe @alexhancock has some insight.

alexhancock:
I can look into this first week of Jan if now one else gets to it first in the rust-sdk

alexhancock:
https://github.com/modelcontextprotocol/rust-sdk/pull/588 should resolve this in goose. Will roll out a new version of the rust-sdk soon and then update to it in goose.

alexhancock:
This should be resolved

PR #6516, opened 2026-01-15
chore: upgrade rmcp

alexhancock:
A relatively straightforward upgrade to the new RMCP. This will close out https://github.com/block/goose/issues/6144 and perhaps more


jamadeo (approved):
Looks good, I think there are a few more places to add the new struct field.

PR #6521, opened 2026-01-15
feat(cli): allow custom timeout for streamable HTTP extensions

Jazzcort:
## Summary
Updates argument parsing to support a `url timeout=<seconds>` format, allowing users to override the default extension timeout.

Also want to add this feature to `--with-extension` (stdio extension), but not sure about what would be the best design of the format since the current format already included envs (Format: "ENV1=val1 ENV2=val2 command args..."). We can either add it at the end (after args...) like what I did for '--with-streamable-http-extension' or having a reserved env like `EXT_TIMEOUT`. The first approach might interfere the command if there are some edge cases, the later approach makes the behavior of passing customized timeout not consistent between `--with-extension` and `--with-streamable-http-extension`. 

### Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing

```shell
goose session --with-streamable-http-extension "http://localhost:<port> timeout=<seconds>"
```
Use the above command to add a streamable http extension to the session with a customized timeout other than the default.
Maybe try eprintln!() the config to check if the timeout is successfully parsed.

### DIFF
--- a/crates/goose-cli/src/cli.rs
+++ b/crates/goose-cli/src/cli.rs
@@ -102,6 +102,38 @@ pub struct SessionOptions {
     pub max_turns: Option<u32>,
 }
 
+#[derive(Debug, Clone)]
+pub struct StreamableHttpOptions {
+    pub url: String,
+    pub timeout: u64,
+}
+
+fn parse_streamable_http_extension(input: &str) -> Result<StreamableHttpOptions, String> {
+    let mut input_iter = input.split_whitespace();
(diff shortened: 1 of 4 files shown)


DOsinga:
I've resolve the merge conflicts in https://github.com/block/goose/pull/6772 and will merge that one. thank you!

Jazzcort:
Thank you so much for reviewing!

Issue #6063, opened 2025-12-11
ElevenLabs API Key Not Persisting in Voice Dictation Settings


WebWorX-io:
**Describe the bug**
When configuring ElevenLabs for Voice Dictation in Goose Desktop, the API key appears to save successfully (UI shows "configured" in green text) but does not actually persist. Upon returning to Chat mode, the microphone icon displays an error: "ElevenLabs API Key is not configured. Set it up in Settings > Chat > Voice Dictation."

📦 To help us debug faster, attach your **diagnostics zip** if possible.  

[diagnostics_20251211_3.zip](https://github.com/user-attachments/files/24095880/diagnostics_20251211_3.zip)

---

**To Reproduce**
Steps to Reproduce
1. Open Goose Desktop
2. Navigate to Settings > Chat > Voice Dictation
3. Select ElevenLabs as the voice dictation provider
4. Enter a valid ElevenLabs API key in the provided field
5. Click outside the API key field
6. Observe "configured" appears in green text (suggesting successful save)
7. Return to Chat mode
8. Click the microphone icon

---

**Expected behavior**
The ElevenLabs API key should persist after being entered, and voice dictation should function when clicking the microphone icon in Chat mode.

**Actual Behavior**
- The UI incorrectly indicates the key is "configured"
- The key is not saved to any persistent storage
- Returning to Chat mode shows error: "ElevenLabs API Key is not configured. Set it up in Settings > Chat > Voice Dictation."
- The issue persists across app restarts

---

**Screenshots**
No screenshots.

---

**System & app information**
- **OS & Arch:** [macOS 14.7.6, Intel]
- **Interface:** [GUI]
- **Version:** [Version 1.16.1 (1.16.1)]
- **Extensions enabled:** [Auto Visualiser, Chatrecall, Computer Controller, Developer, Extension Manager, Memory, Skills, Todo, Tutorial]
- **Provider & Model:** [anthropic/claude-opus-4.5]
- **ElevenLabs Account Type:** [Free]

---

**Additional context**
Investigation Notes
Checked the following storage locations - none contain the ElevenLabs API key after configuration:

~/.config/goose/config.yaml - No ElevenLabs entries
macOS Keychain - No ElevenLabs-related entries found
~/Library/Application Support/Goose/ local storage - No ElevenLabs data
Manually adding ELEVENLABS_API_KEY to config.yaml did not resolve the issue.

Additional Context
[...]

PR #6557, opened 2026-01-17
Fix: ElevenLabs API Key Not Persisting

Abhijay007:
Closes: #6063 

### PR description

This PR fixes a bug where ElevenLabs API keys were not persisting in the desktop UI.

### Changes:

- Fixed key existence detection logic in both the settings hook and key input component
- Improved UI styling for provider dropdown (overflow, width, text wrapping)

### Type of Change
- [x] Bug fix

### AI Assistance

### Testing

Tested in Desktop UI with the whole functionality 

Before:  
elevanLabsFix

After:   

elevanLabsFixAfter

### DIFF
--- a/ui/desktop/package-lock.json
+++ b/ui/desktop/package-lock.json
@@ -217,6 +217,7 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.5",
@@ -586,6 +587,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -629,6 +631,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=18"
       }
@@ -1086,6 +1089,7 @@
       "integrity": "sha512-zx0EIq78WlY/lBb1uXlziZmDZI4ubcCXIMJ4uGjXzZW0nS19TjSPeXPAjzzTmKQlJUZm0SbmZhPKP7tuQ1SsEw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "chalk": "^4.1.1",
         "fs-extra": "^9.0.1",
@@ -2655,6 +2659,7 @@
       "integrity": "sha512-yl43JD/86CIj3Mz5mvvLJqAOfIup7ncxfJ0Btnl0/v5TouVUyeEdcpknfgc+yMevS/48oH9WAkkw93m7otLb/A==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@inquirer/checkbox": "^3.0.1",
         "@inquirer/confirm": "^4.0.1",
@@ -6221,8 +6226,7 @@
       "resolved": "https://registry.npmjs.org/@types/aria-query/-/aria-query-5.0.4.tgz",
       "integrity": "sha512-rfT93uj5s0PRL7EzccGMs3brplhcrghnDoV26NqKhCAS1hVo+WdNsPvE/yb6ilfr5hi2MEk6d5EWJTKdxg8jVw==",
       "dev": true,
-      "license": "MIT",
-      "peer": true
+      "license": "MIT"
     },
     "node_modules/@types/babel__core": {
       "version": "7.20.5",
@@ -6510,6 +6514,7 @@
       "integrity": "sha512-QoiaXANRkSXK6p0Duvt56W208du4P9Uye9hWLWgGMDTEoKPhuenzNcC4vGUmrNkiOKTlIrBoyNQYNpSwfEZXSg==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "undici-types": "~7.16.0"
       }
@@ -6551,6 +6556,7 @@
       "resolved": "https://registry.npmjs.org/@types/react/-/react-19.2.8.tgz",
       "integrity": "sha512-3MbSL37jEchWZz2p2mjntRZtPt837ij10ApxKfgmXCTuHWagYg7iA5bqPw6C8BMPfwidlvfPI/fxOc42HLhcyg==",
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "csstype": "^3.2.2"
       }
@@ -6561,6 +6567,7 @@
       "integrity": "sha512-jp2L/eY6fn+KgVVQAOqYItbF0VY/YApe5Mz2F0aykSO8gx31bYCZyvSeYxCHKvzHG5eZjc+zyaS5BrBWya2+kQ==",
       "devOptional": true,
       "license": "MIT",
+      "peer": true,
       "peerDependencies": {
         "@types/react": "^19.2.0"
       }
@@ -6701,6 +6708,7 @@
       "integrity": "sha512-npiaib8XzbjtzS2N4HlqPvlpxpmZ14FjSJrteZpPxGUaYPlvhzlzUZ4mZyABo0EFrOWnvyd0Xxroq//hKhtAWg==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@typescript-eslint/scope-manager": "8.53.0",
         "@typescript-eslint/types": "8.53.0",
@@ -7086,6 +7094,7 @@
       "integrity": "sha512-hRDjg6dlDz7JlZAvjbiCdAJ3SDG+NH8tjZe21vjxfvT2ssYAn72SRXMge3dKKABm3bIJ3C+3wdunIdur8PHEAw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@vitest/utils": "4.0.17",
         "fflate": "^0.8.2",
@@ -7334,6 +7343,7 @@
       "integrity": "sha512-NZyJarBfL7nWwIq+FDL6Zp/yHEhePMNnnJ0y3qfieCrmNvYct8uvtiV41UvlSe6apAfk0fY1FbWx+NwfmpvtTg==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "bin": {
         "acorn": "bin/acorn"
       },
@@ -7959,6 +7969,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "baseline-browser-mapping": "^2.9.0",
         "caniuse-lite": "^1.0.30001759",
@@ -9257,8 +9268,7 @@
       "resolved": "https://registry.npmjs.org/dom-accessibility-api/-/dom-accessibility-api-0.5.16.tgz",
       "integrity": "sha512-X7BJ2yElsnOJ30pZF4uIIDfBEVgF4XEBxL9Bxhy6dnrm5hkzqmsWHGTiHqRiITNhMyFLyAiWndIJP7Z1NTteDg==",
       "dev": true,
-      "license": "MIT",
-      "peer": true
+      "license": "MIT"
     },
     "node_modules/dom-helpers": {
       "version": "5.2.1",
@@ -9316,6 +9326,7 @@
       "dev": true,
       "hasInstallScript": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@electron/get": "^2.0.0",
         "@types/node": "^22.7.7",
@@ -10330,6 +10341,7 @@
       "integrity": "sha512-LEyamqS7W5HB3ujJyvi0HQK/dtVINZvd5mAAp9eT5S/ujByGjiZLCzPcHVzuXbpJDJF/cxwHlfceVUDZ2lnSTw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@eslint-community/eslint-utils": "^4.8.0",
         "@eslint-community/regexpp": "^4.12.1",
@@ -10789,6 +10801,7 @@
       "resolved": "https://registry.npmjs.org/express/-/express-5.2.1.tgz",
       "integrity": "sha512-hIS4idWWai69NezIdRt2xFVofaF4j+6INOpJlVOLDO8zXGpUVEVzIYk12UUi2JzjEzWL3IOAxcTubgz9Po0yXw==",
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "accepts": "^2.0.0",
         "body-parser": "^2.2.1",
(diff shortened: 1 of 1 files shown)


ui/desktop/src/components/settings/dictation/ElevenLabsKeyInput.tsx
```diff
@@ -45,23 +55,32 @@ export const ElevenLabsKeyInput = () => {
...
+        return;
+      }
+
+      const trimmedKey = elevenLabsApiKey.trim();
+
+      if (trimmedKey) {
+        if (trimmedKey.length < 32) {
+          setHasElevenLabsKey(false);
+          return;
+        }
```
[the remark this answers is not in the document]
Abhijay007: [4fece4e](https://github.com/block/goose/pull/6557/commits/4fece4eabd7965c5baff8b4615a920f5453e4ea3)

ui/desktop/src/components/settings/dictation/ElevenLabsKeyInput.tsx
```diff
@@ -78,10 +97,15 @@ export const ElevenLabsKeyInput = () => {
...
       </div>
       <Input
-        type="password"
+        type="text"
         value={elevenLabsApiKey}
         onChange={(e) => handleElevenLabsKeyChange(e.target.value)}
         onBlur={saveElevenLabsKey}
+        onFocus={(e) => {
+          if (e.target.value === '••••••••••••••••') {
+            setElevenLabsApiKey('');
```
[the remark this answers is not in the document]
Abhijay007: [4fece4e](https://github.com/block/goose/pull/6557/commits/4fece4eabd7965c5baff8b4615a920f5453e4ea3)

ui/desktop/src/components/settings/dictation/ElevenLabsKeyInput.tsx
```diff
@@ -82,12 +109,19 @@ export const ElevenLabsKeyInput = () => {
...
+            elevenLabsApiKeyRef.current = '';
+          }
+        }}
```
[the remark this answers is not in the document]
Abhijay007: [649fa8c](https://github.com/block/goose/pull/6557/commits/649fa8c47de17812ea209668da92109c8159a115)

ui/desktop/src/components/settings/dictation/ElevenLabsKeyInput.tsx
```diff
@@ -7,16 +7,29 @@ export const ElevenLabsKeyInput = () => {
...
       setIsLoadingKey(true);
       try {
         const keyExists = await read(ELEVENLABS_API_KEY, true);
-        if (keyExists === true) {
+        const hasKey =
+          keyExists !== null &&
+          keyExists !== '' &&
+          (typeof keyExists === 'object' && 'masked_value' in keyExists
+            ? keyExists.masked_value
+            : keyExists);
```
lifeizhou-ap: if `keyExists` is a non empty string, the `hasKey` will also return true which is not expected behaviour

The `keyExist` variable name is a bit confusing now, can rename it to `response`.

While fixing another issue in this PR https://github.com/block/goose/pull/6662, I have created a function `isSecretKeyConfigured`  to check hasKey and are used in both ElevenLabsKeyInput.ts here and useDictationSettings.ts.   We can use this function.

ui/desktop/src/components/settings/dictation/ElevenLabsKeyInput.tsx
```diff
@@ -45,26 +57,47 @@ export const ElevenLabsKeyInput = () => {
...
-        console.log('Saving ElevenLabs API key to secure storage...');
-        await upsert(ELEVENLABS_API_KEY, elevenLabsApiKey, true);
+      if (elevenLabsApiKey === '••••••••••••••••') {
+        return;
+      }
+
+      const trimmedKey = elevenLabsApiKey.trim();
+
+      if (trimmedKey) {
+        if (trimmedKey.length < 32) {
```
lifeizhou-ap: Do you know where the ‘minimum 32 characters’ requirement comes from? Wondering if it's something we need to apply.

ui/desktop/src/hooks/useDictationSettings.ts
```diff
@@ -34,7 +34,7 @@ export const useDictationSettings = () => {
       // Load ElevenLabs API key from storage (non-secret for frontend access)
       try {
         const keyExists = await read(ELEVENLABS_API_KEY, true);
-        if (keyExists === true) {
+        if (keyExists !== null) {
```
lifeizhou-ap: the logic is not consistent with the ElevenLabsKeyInput.ts

ui/desktop/src/components/settings/dictation/ElevenLabsKeyInput.tsx
```diff
@@ -7,16 +7,29 @@ export const ElevenLabsKeyInput = () => {
...
+        const hasKey =
+          keyExists !== null &&
+          keyExists !== '' &&
+          (typeof keyExists === 'object' && 'masked_value' in keyExists
+            ? keyExists.masked_value
+            : keyExists);
+
+        if (hasKey) {
           setHasElevenLabsKey(true);
+          setElevenLabsApiKey('••••••••••••••••');
```
lifeizhou-ap: I am thinking whether we should set '••••••••••••••••'. This makes code a bit complicated and introduces a few code duplications to check the input with this string and the setWasMasked state.   Also I feel user might be confused as their input is also masked and looks similar

If I understand correctly, this is to solve the existing problem: when the api key has configured, user focuses on the api key input, and blur, then it will unset the api key at the backend. 

Maybe we can have 

- When loading the Setting page, it only show
    - whether the api key is configured -> (configured) label 
    - add/update button (if api key has configured, show `update` otherwise `add`) 
    - no text input box
- When user clicks the add/update button, the text input appears and user enters the api key, then submit.

Will this be cleared to the user?

ui/desktop/src/components/settings/dictation/ElevenLabsKeyInput.tsx
```diff
@@ -45,26 +57,47 @@ export const ElevenLabsKeyInput = () => {
...
+        setElevenLabsApiKey('••••••••••••••••');
+        setValidationError('');
+        setWasMasked(false);
+      } else if (wasMasked) {
+        setElevenLabsApiKey('••••••••••••••••');
+        setWasMasked(false);
       } else {
-        console.log('Removing ElevenLabs API key from secure storage...');
-        await upsert(ELEVENLABS_API_KEY, null, true);
+        await remove(ELEVENLABS_API_KEY, true);
```
lifeizhou-ap: Nice one!

ui/desktop/src/components/settings/dictation/ElevenLabsKeyInput.tsx
Abhijay007: well I added this as it's taking a single letter too, so I thought instead of allowing it maybe we should set the minimum character limit

ui/desktop/src/components/settings/dictation/ProviderSelector.tsx
```diff
@@ -1,7 +1,6 @@
 import { useState, useEffect } from 'react';
 import { ChevronDown } from 'lucide-react';
 import { DictationProvider, DictationSettings } from '../../../hooks/useDictationSettings';
-import { DICTATION_PROVIDER_ELEVENLABS } from '../../../hooks/dictationConstants';
 import { useConfig } from '../../ConfigContext';
```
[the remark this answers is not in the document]
Abhijay007: 7fec68f

Abhijay007:
Updated UX 

updateProvider

cc : @lifeizhou-ap

lifeizhou-ap:
Thank you @Abhijay007!

One thing I noticed the changes to optimise api call in the [PR](https://github.com/block/goose/pull/6662) have been removed in this PR.  Just wondering whether it is intentional or not.

Abhijay007:
> Thank you @Abhijay007!
> 
> One thing I noticed the changes to optimise api call in the [PR](https://github.com/block/goose/pull/6662) have been removed in this PR. Just wondering whether it is intentional or not.

oh yea, thanks @lifeizhou-ap for letting me know I made some mistakes while resolving merge conflicts and missed those changes 😅 , I updated em now

ui/desktop/src/components/settings/dictation/ElevenLabsKeyInput.tsx
```diff
@@ -8,71 +9,75 @@ export const ElevenLabsKeyInput = () => {
...
+      setElevenLabsApiKey('');
+      setValidationError('');
+      setIsEditing(false);
+      await loadKey();
     } catch (error) {
-      console.error('Error saving ElevenLabs API key:', error);
+      console.error(error);
+      setValidationError('Failed to save API key');
     }
   };
```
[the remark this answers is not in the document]
Abhijay007: It got changed, we changed UX

taniacryptid:
It looks like comments were addressed, tagging @lifeizhou-ap and @zanesq again!

ui/desktop/src/components/settings/dictation/ElevenLabsKeyInput.tsx
```diff
@@ -8,71 +9,75 @@ export const ElevenLabsKeyInput = () => {
...
+      if (!trimmedKey) {
+        setValidationError('API key is required');
+        return;
       }
+
+      await upsert(ELEVENLABS_API_KEY, trimmedKey, true);
+      setElevenLabsApiKey('');
+      setValidationError('');
+      setIsEditing(false);
+      await loadKey();
```
lifeizhou-ap, line 55: Hi @Abhijay007 just curious why we hav to loadKey again in the upsert  and remove? It added an extra call to the server side and it seems unnecessary?  Shall we just update the cache and hasElevenLabsKey instead of the additional call?

ui/desktop/src/hooks/useDictationSettings.ts
```diff
@@ -8,7 +8,7 @@ import {
   isSecretKeyConfigured,
 } from './dictationConstants';
 
-export type DictationProvider = 'openai' | typeof DICTATION_PROVIDER_ELEVENLABS | null;
+export type DictationProvider = 'openai' | 'elevenlabs' | null;
```
lifeizhou-ap: shall we use the constant here

ui/desktop/package-lock.json
```diff
@@ -217,7 +217,6 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
```
DOsinga, line 220: I see this flip back and forth. best not to include unless making changes

ui/desktop/src/components/settings/dictation/ElevenLabsKeyInput.tsx
```diff
@@ -1,5 +1,6 @@
-import { useState, useEffect, useRef } from 'react';
+import { useState, useEffect, useCallback } from 'react';
```
DOsinga, line 1: frankly, this whole module feels rather complicated. aren't we just setting an API key? there's validation code, but all it seems to do is check that the API key is set. or that the server failed but that's not very useful.

ui/desktop/src/hooks/useDictationSettings.ts
DOsinga: they should both be constants, no?

lifeizhou-ap:
> they should both be constants, no?

Hi @DOsinga Yes.  This is my fault because I did not extract `openai` as constant in another PR.  I was intending to extract in the same PR, but `openai` is used in different contexts.  It was not just used in dictation, it is also used in provider, and other etc.  So I did not do it and thought about doing later. 

Now I realised that I has overthought about the risk of extracting the `openai` constant.  It is ok to only replace dictation related `openai`.

ui/desktop/package-lock.json
lifeizhou-ap: yes, I had this too.  I guess we may use different `npm` versions.

I am using npm 11.6.0 @Abhijay007 would you please help check your npm and node version? I guess you might be using the older version. 

I'll also create a small pr to enforce newer version of npm

lifeizhou-ap:
Hi @Abhijay007, 

I also checked out your branch and it works well!

One feedback about the "Remove" button in UI.

This is your current UI
Screenshot 2026-01-27 at 10 34 08 am. With the "Remove" below the input text box, user may think it only removes the text in the input box instead of removing the api key. 

I feel it will be clearer to user if we put next to "Update API Key".  WDTY? (does not need to be in this PR. we can address this later if this makes sense)
Screenshot 2026-01-27 at 10 34 17 am

ui/desktop/package-lock.json
lifeizhou-ap: PR created https://github.com/block/goose/pull/6727, wait for @Abhijay007 to provide his npm and node version to help us validate the cause of the problem.

zanesq: The node version should come from hermit, its locked in at v24.10.0 currently

lifeizhou-ap: yeah, it depends on whether hermit auto activation is turned on or not on the machine. Maybe our contributer has not set this up.  https://cashapp.github.io/hermit/usage/shell/?h=#shell-hooks

ui/desktop/src/components/settings/dictation/ElevenLabsKeyInput.tsx
Abhijay007: it was mentioned here by copoilt : https://github.com/block/goose/pull/6557#discussion_r2722501050 so I thought of implementing it

Abhijay007:
> Hi @Abhijay007,
> 
> I also checked out your branch and it works well!
> 
> One feedback about the "Remove" button in UI.
> 
> This is your current UI Screenshot 2026-01-27 at 10 34 08 am. With the "Remove" below the input text box, user may think it only removes the text in the input box instead of removing the api key.
> 
> I feel it will be clearer to user if we put next to "Update API Key". WDTY? (does not need to be in this PR. we can address this later if this makes sense) Screenshot 2026-01-27 at 10 34 17 am

I think it's nice if we handle it in another PR

ui/desktop/src/components/settings/dictation/ElevenLabsKeyInput.tsx
Abhijay007: > frankly, this whole module feels rather complicated. aren't we just setting an API key? there's validation code, but all it seems to do is check that the API key is set. or that the server failed but that's not very useful.

@lifeizhou-ap , @DOsinga should I consider this update in this PR or make another one for this refactor ?

lifeizhou-ap: Hi @Abhijay007,

Let's make another PR for this refactor.

lifeizhou-ap:
> > they should both be constants, no?
> 
> Hi @DOsinga Yes. This is my fault because I did not extract `openai` as constant in another PR. I was intending to extract in the same PR, but `openai` is used in different contexts. It was not just used in dictation, it is also used in provider, and other etc. So I did not do it and thought about doing later.
> 
> Now I realised that I has overthought about the risk of extracting the `openai` constant. It is ok to only replace dictation related `openai`.

Thank you @Abhijay007 to extract it!

ui/desktop/package-lock.json
lifeizhou-ap: Hi @Abhijay007,

Would you please provide us your npm and node version to help us validate the cause of the problem. (ie: without "peer": true in your package-lock.json file when running npm install) Thank you!

Abhijay007: node version : v24.0. 0, npm version : 11.3.0

PR #6575, opened 2026-01-19
feat: add native structured output support for providers

rabi:
## Summary
Implements native structured output at the provider level, saving LLM roundtrips with tool calls.

- Add supports_structured_output() trait method to Provider
- Implement for OpenAI (gpt-4o, gpt-4.x, o1/o3/o4 models)
- Implement for Anthropic (Claude 4.x models)
- Implement for Google (Gemini 3.x models)
- Pass recipe json_schema to provider's response_format when supported
- Add CustomProviderParams struct to simplify provider config APIs
- Add supports_structured_output field to DeclarativeProviderConfig

### Type of Change
- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested locally with google provider.

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -131,6 +131,7 @@ pub struct Agent {
     pub(super) retry_manager: RetryManager,
     pub(super) tool_inspection_manager: ToolInspectionManager,
     container: Mutex<Option<Container>>,
+    pub(super) response_schema: Mutex<Option<serde_json::Value>>,
 }
 
 #[derive(Clone, Debug)]
@@ -216,6 +217,7 @@ impl Agent {
             retry_manager: RetryManager::new(),
             tool_inspection_manager: Self::create_tool_inspection_manager(permission_manager),
             container: Mutex::new(None),
+            response_schema: Mutex::new(None),
         }
     }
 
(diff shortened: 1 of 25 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -458,14 +460,22 @@ impl Agent {
...
             self.add_sub_recipes(sub_recipes).await;
         }
 
-        if include_final_output {
-            if let Some(response) = response {
-                self.add_final_output_tool(response).await;
+        if let Some(ref response) = response {
+            if let Some(schema) = &response.json_schema {
+                if provider_supports_structured_output {
+                    *self.response_schema.lock().await = Some(schema.clone());
```
DOsinga, line 472: I'm not sure about this. when an agent runs a recipe, it will do a bunch of toolcalling and say things in between and then only when it is done, it will call the final toolcall and if it doesn't, we will remind it to

with this change, we force the agent to immediately return the content in the response_schema, effectively blocking toolcalling (if I read the code correctly)

rabi: Thanks @DOsinga for having a look.

Regarding your concern (if I have understood correctly),  I think response_format/output_format from provider only constrains the format of the final output. We can trust the models to call tools when needed, and only provide the output in the expected format when done.

From Anthropic docs...

"When combined, Claude can call tools with guaranteed-valid parameters AND return structured JSON responses. This is useful for agentic workflows where you need both reliable tool calls and structured final outputs."

https://platform.claude.com/docs/en/build-with-claude/structured-outputs#using-both-features-together

I have seen similar statement openai and gemini docs.

If this makes sense let me know, I'll rebase the change.

DOsinga:
I'm not entirely convinced. normally what models seem to do is something like:

let me have a look in the directory
<toolcall>
oh, I see, there's a file missing

I need to do something about that
<toolcall>
there it is, all good now

those text bits seem to be important, at one point we were missing them which made goose repeat itself, because it never had in the conversations the anchor points that it had done things.

the way I understand it, is that these structured outputs would break that

rabi:
> those text bits seem to be important, at one point we were missing them which made goose repeat itself, because it never had in the conversations the anchor points that it had done things.

Did you notice that behavior with any provider?  I've not seen those anchor text bits removed when using response_schema, as it only applies to the output (final turn). It just changes the final output according to the schema from what I've seen. Having said that, I'll do some more testing with different providers.

rabi:
Thanks @DOsinga. After testing with more providers, I confirmed that some Anthropic models do drop the interleaved text responses when structured output is enabled, even though I didn't observe actual repetition in my tests or the final outcome. Since this could be a risk as you pointed out, I'll close this for now and revisit in the future.

PR #6584, opened 2026-01-20
fix(goose): propagate session_id across providers and MCP

codefromthecrypt:
## Summary
Propagate explicit `session_id` through provider and MCP requests, allowing coherent session IDs everywhere we can.

We use a fallback for two cases:
- connection-scoped for unsolicited MCP sampling
- new UUIDs for request paths that have no user session.

This notably fixes session-id propagation such as from servers who do MCP sampling in [health checks](https://github.com/cyrup-ai/sweetmcp/blob/17cbc64c054a6f06cebb0d83f409d041b2a4a68b/packages/sweetmcp-axum/src/ui/health/mod.rs#L206).
This also fixes session ID propagation in session rename requests originally started in #5624.

Importantly, by backfilling ACP integration hooks, we now can fail any change that breaks session ID propagation.
This is important because before it was subtle, like task locals which might not even be the correct context. For example,
MCP sampling callbacks do not share a task local with a tool call that caused it!

### Type of Change

- [ ] Feature
- [x] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
- ACP fixtures cover streamable HTTP + MCP metadata coherence, including session list and session rename requests.
- MCP client unit tests cover request injection (list/read/call/prompt) and sampling callback session resolution.

### Related Issues
Resumes from #5624

### DIFF
--- a/crates/goose/src/agents/reply_parts.rs
+++ b/crates/goose/src/agents/reply_parts.rs
@@ -8,6 +8,7 @@ use tracing::debug;
 
 use super::super::agents::Agent;
 use crate::agents::code_execution_extension::EXTENSION_NAME as CODE_EXECUTION_EXTENSION;
+use crate::agents::subagent_tool::SUBAGENT_TOOL_NAME;
 use crate::conversation::message::{Message, MessageContent, ToolRequest};
 use crate::conversation::Conversation;
 use crate::providers::base::{stream_from_single_message, MessageStream, Provider, ProviderUsage};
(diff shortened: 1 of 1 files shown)


crates/goose-acp/tests/common.rs
```diff
@@ -14,57 +17,128 @@ use wiremock::{Mock, MockServer, ResponseTemplate};
...
             }
-        })
-        .mount(&mock_server)
-        .await;
+            None => Ok(()),
+        }
+    }
 
-    mock_server
+    /// Calling this ensures incidental requests that might error asynchronously, such as
```
codefromthecrypt, line 61: this is a very important part of ACP tests as it tracks any session ID mismatch including no session ID. if any propagation breaks in the future, and it is reproducible by ACP, this will be the tripwire

michaelneale:
I think this is ok, I think @zanesq and @DOsinga should give their blessing as fairly cross cutting but should flush out some weirdness. 

One small conflict to resolve, and that live test failure is a one off I think... worth retrying.

codefromthecrypt:
the live one must have been a flake. merging this one in as without it sessions are lost through acp and also we have some edge cases even in normal paths (like session rename) which have incoherent session id propagation.

PR #6585, opened 2026-01-20
feat(google): add configurable thinking level for Gemini 3 models

rabi:
## Summary

- Defaults to 'low' instead of Google's default 'high' for better latency
- Supports 'low' and 'high' levels for Gemini 3 models

### Type of Change
- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Unit testing and local testing with gemini3 models.

### DIFF
--- a/crates/goose/src/providers/formats/google.rs
+++ b/crates/goose/src/providers/formats/google.rs
@@ -533,6 +533,21 @@ struct GenerationConfig {
     temperature: Option<f64>,
     #[serde(skip_serializing_if = "Option::is_none")]
     max_output_tokens: Option<i32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    thinking_config: Option<ThinkingConfig>,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "lowercase")]
+enum ThinkingLevel {
+    Low,
+    High,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct ThinkingConfig {
+    thinking_level: ThinkingLevel,
 }
 
 #[derive(Serialize)]
@@ -544,6 +559,45 @@ struct GoogleRequest<'a> {
     tools: Option<ToolsWrapper>,
     #[serde(skip_serializing_if = "Option::is_none")]
     generation_config: Option<GenerationConfig>,
+}
+
+fn get_thinking_config(model_config: &ModelConfig) -> Option<ThinkingConfig> {
+    if !model_config
+        .model_name
+        .to_lowercase()
+        .starts_with("gemini-3")
+    {
+        return None;
+    }
+
+    let thinking_level_str = model_config
+        .request_params
+        .as_ref()
+        .and_then(|params| params.get("thinking_level"))
+        .and_then(|v| v.as_str())
+        .map(|s| s.to_lowercase())
+        .or_else(|| {
+            crate::config::Config::global()
+                .get_param::<String>("gemini3_thinking_level")
+                .ok()
+                .map(|s| s.to_lowercase())
+        })
+        .unwrap_or_else(|| "low".to_string());
+
+    let thinking_level = match thinking_level_str.as_str() {
+        "high" => ThinkingLevel::High,
+        "low" => ThinkingLevel::Low,
+        invalid => {
+            tracing::warn!(
+                "Invalid thinking level '{}' for model '{}'. Valid levels: low, high. Using 'low'.",
+                invalid,
+                model_config.model_name,
+            );
+            ThinkingLevel::Low
+        }
+    };
+
+    Some(ThinkingConfig { thinking_level })
 }
 
 pub fn create_request(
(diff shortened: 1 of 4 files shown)


crates/goose/src/providers/formats/google.rs
```diff
@@ -662,15 +702,20 @@ pub fn create_request(
...
+    let thinking_config = get_thinking_config(&model_config.model_name);
+
+    let generation_config = if model_config.temperature.is_some()
+        || model_config.max_tokens.is_some()
+        || thinking_config.is_some()
+    {
+        Some(GenerationConfig {
+            temperature: model_config.temperature.map(|t| t as f64),
+            max_output_tokens: model_config.max_tokens,
+            thinking_config,
```
katzdave, line 626: Are older models e.g. 2.5 flash happy to accept this new field?

documentation/docs/guides/environment-variables.md
```diff
@@ -139,6 +139,26 @@ export DATABRICKS_BACKOFF_MULTIPLIER=2               # each retry waits 2x longe
...
+Available levels:
+- **low** - Reduced reasoning, good for most tasks (default in goose)
+- **high** - Full reasoning depth (Google's default)
+
+**Examples**
+
+```bash
+# Use high thinking for complex reasoning tasks
+export GEMINI_THINKING_LEVEL=high
+```
```
dianed-square:
```suggestion
<!--
### Model-Specific Configuration

#### Google Gemini 3

| Variable | Purpose | Values | Default |
|----------|---------|---------|---------|
| `GEMINI_THINKING_LEVEL` | Controls thinking depth for Gemini 3 models | "low", "high" | "low" |

This setting applies to **Gemini 3 models** (`gemini-3-pro-preview`, `gemini-3-pro-image-preview`) on both Google AI and GCP Vertex AI providers. Google's default is "high", but goose uses "low" for better latency.

Available levels:
- **low** - Reduced reasoning, good for most tasks (default in goose)
- **high** - Full reasoning depth (Google's default)

**Examples**

```bash
# Use high thinking for complex reasoning tasks
export GEMINI_THINKING_LEVEL=high
```
-->
```

Comment (dianed-square) [reply documentation/docs/guides/environment-variables.md]: @rabi Thanks for contributing! Can we comment out this section (temporarily) to keep the public docs in sync with the latest released version?

File: crates/goose/src/providers/formats/google.rs (hunk 1 above)
Comment (rabi) [reply crates/goose/src/providers/formats/google.rs:626]: Yeah, API is backward compatible and we're sending `None` for non gemini3 models. I've not seen any issues in my testing with 2.5/2.0-flash. Older models use `ThinkingBudget` to control thinking. If we want we can implement that as well.

**Comment by rabi** [comment]
> I think is ok with me. Would rather not add more env things but maybe in this case and just for this provider. Must be some other way to do config which is relevant to the provider? (I think you can add a variable to the provider and let people set it there - may be better @rabi ?)

I think users would want these to be session specific, so prpbably won't make sense to have it in provider config. The same pattern of env vars are already there for others. CLAUDE_THINKING_ENABLED/CLAUDE_THINKING_BUDGET for Anthropic.

**Comment by rabi** [comment]
> I think you can add a variable to the provider and let people set it there

@michaelneale  Hey! Are you ok with the explanation or still want it in provider config? Even if we set in provider config, it can be overridden by env vars[1] and also it does not apply to all models (only gemini 3 models) for google provider.

Else, we can merge this to reduce the latency as we now default to 'low' thinking for our use-cases.

[1] https://github.com/block/goose/blob/main/crates/goose/src/config/base.rs#L637

File: ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx [hunk 3]
```diff
@@ -101,6 +106,9 @@ export const SwitchModelModal = ({
   const [loadingModels, setLoadingModels] = useState<boolean>(false);
   const [userClearedModel, setUserClearedModel] = useState(false);
   const [providerErrors, setProviderErrors] = useState<Record<string, string>>({});
+  const [thinkingLevel, setThinkingLevel] = useState<string>('low');
+
+  const isGemini3Model = model.toLowerCase().startsWith('gemini-3');
```
michaelneale: this seems duplicated and also not used in another place?

michaelneale:
thanks - is there some way to normalise this, this seems very very hard coded to gemini-3 for something they will probably change their mind over - can it be mapped from some normalised thing in the provider?

ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
rabi: This is for custom model input and L156 checks for predefined models. I'll update to use it there as well, but it would be complex check.

rabi:
> thanks - is there some way to normalise this, this seems very very hard coded to gemini-3 for something they will probably change their mind over - can it be mapped from some normalised thing in the provider?

It's a gemini-3 thing and we've these kind of checks elsewhere too. They may change things in the future but we can't predict what api changes they would make. That's the reason I had earlier proposed to just use an env var. There is no way we can give the option of thinking leval to the user before a model is selected during configuration.

Issue #1785, opened 2025-03-20
Enable Link-Time Optimization (LTO) and codegen-units = 1


zamazan4ik:
Hi!

I noticed that in the `Cargo.toml` file Link-Time Optimization (LTO) for the project is not enabled. I suggest switching it on since it will reduce the binary size (always a good thing to have) and will likely improve the application's performance. If you want to read more about LTO and its possible modes, I recommend starting from [this](https://doc.rust-lang.org/rustc/codegen-options/index.html#lto) Rustc documentation.

I think you can enable LTO only for the Release builds so as not to sacrifice the developers' experience while working on the project, since LTO consumes an additional amount of time to finish the compilation routine. In this case, we can create a dedicated `[profile.optimized-dev]` profile where LTO will be disabled (so developers experience will not be affected). If we enable it on the Cargo profile level for the Release profile, users, who install the application with `cargo install`, will get the LTO-optimized version of the app "automatically". E.g., check `cargo-outdated` Release [profile](https://github.com/kbknapp/cargo-outdated/blob/master/Cargo.toml#L48). You also could be interested in other optimization options like `codegen-units = 1` - it also brings improvements over the current defaults.

Basically, it can be enabled with the following lines:

```
[profile.release]
codegen-units = 1
lto = true
```

Probably, you will also will be interested in enabling other optimizations - I just tested these two for the project.

I have made quick tests (AMD Ryzen 5900x, Fedora 41, Rust 1.85.1, the latest version of the project at the moment, `cargo build --release` command) - here are the results. Here are results for some binaries (non-exhaustive list):

| Binary\Build mode | goose | goosed | mcp-server |
| --- | --- | --- | --- |
| Release | 77 Mib | 71 Mib | 6.1 Mib |
| Release + `codegen-units = 1` + ThinLTO | 61 Mib | 55 Mib | 4.7 Mib |
| Release + `codegen-units = 1` + FatLTO | 54 Mib | 48 Mib | 4.1 Mib |

Clean build times:

* Release: 1m 35s
* Release + `codegen-units = 1` + ThinLTO: 2m 51s
* Release + `codegen-units = 1` + FatLTO: 5m 03s

Thank you.

PR #6586, opened 2026-01-20
fix(1785): enable lto and codegen-units in release build

lsytj0413:
## Summary

binary size:

- release：goose 126M，goosed 113M
- release + lto + codegen-units：goose 90M，goosed 81M

build time:

- release: 6m26.735s
- relase + lto + codegen-units: 12m29.807s

Fix: https://github.com/block/goose/issues/1785

### Type of Change

- [x] Feature

### AI Assistance

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,3 +36,7 @@ reqwest = { version = "0.12.28", default-features = false }
 tower = "0.5.2"
 tower-http = "0.6.8"
 url = "2.5.8"
+
+[profile.release]
+codegen-units = 1
+lto = true


lsytj0413:
@michaelneale  could you please take a look at this pr？

lsytj0413:
@lifeizhou-ap  could you please take a look at this pr？

Cargo.toml
```diff
@@ -36,3 +36,7 @@ reqwest = { version = "0.12.28", default-features = false }
 tower = "0.5.2"
 tower-http = "0.6.8"
 url = "2.5.8"
+
+[profile.release]
+codegen-units = 1
+lto = true
```
jh-block, line 42: Thanks for the contribution, it will be good to have a more optimised release build. I think it would be better to use `lto = "thin"` here, what do you think? ThinLTO achieves optimisation results very close to "fat" LTO but with substantially less build time penalty

jh-block: Actually, I see in the linked issue that all three were compared and fat LTO is doing better at reducing binary size. I'll merge this as-is for now and if build times become an issue we can revisit. Thanks again!

PR #6587, opened 2026-01-20
Lifei/fixed accumulated token count

lifeizhou-ap:
## Summary
Fix wrong accumulated token count for databricks claude models.

**Why**
For Databricks Claude models, the usage is in all the chunks with same `prompt_token`. But only the last one with finish_reason has `completion_token`.  

With the change in this PR https://github.com/block/goose/pull/6493, for each api, 
**Actual:** we adds `prompt_token` in all the chunks
**Expected:** we should only add once, which can use the chunk with finish_reason.

**What**
- Only use the usage with `completion_token`. If `completion_token` is null, this means we don't have the usage yet. 
- Added usage in the tool_call_chunk too (I think this was an existing issue)
- Added tests with api response test data

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Unit tests and manual tests

### DIFF
--- a/crates/goose/src/providers/formats/openai.rs
+++ b/crates/goose/src/providers/formats/openai.rs
@@ -518,7 +518,7 @@ where
                                 let tool_chunk: StreamingChunk = serde_json::from_str(line)
                                     .map_err(|e| anyhow!("Failed to parse streaming chunk: {}: {:?}", e, &line))?;
 
-                                if let Some(chunk_usage) = extract_usage_witqh_output_tokens(&tool_chunk) {
+                                if let Some(chunk_usage) = extract_usage_with_output_tokens(&tool_chunk) {
                                     usage = Some(chunk_usage);
                                 }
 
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/formats/openai.rs
```diff
@@ -1406,6 +1417,69 @@ mod tests {
...
+            result.usage_count
+        );
+
+        let usage = result.usage.as_ref().expect("Expected usage to be present");
+        assert_eq!(usage.usage.input_tokens, Some(expected_input));
+        assert_eq!(usage.usage.output_tokens, Some(expected_output));
+        assert_eq!(usage.usage.total_tokens, Some(expected_total));
+    }
+
     #[tokio::test]
```
DOsinga, line 1483: I haven't looked at the details since this just feels like a large dump, but not sure we need three tests here to verify that we are now doing the right thing?

jamadeo (commented):
Is it the same across all providers @lifeizhou-ap ? You mention databricks/claude but since this code is in formats/openai.rs I'd be worried if we're missing some specific logic per-provider for token tracking

crates/goose/src/providers/formats/openai.rs
lifeizhou-ap: The bug is caused by different structures of the streaming chunks in api response.   So I added the tests with the test data with variations for different providers.  This will provide harness when we changing the logic in the future.

lifeizhou-ap:
> Is it the same across all providers @lifeizhou-ap ? You mention databricks/claude but since this code is in formats/openai.rs I'd be worried if we're missing some specific logic per-provider for token tracking

Yes, I was worrying about this too including the last PR. I manually tested with different providers again.  Also I added the chunk payload in new tests to make sure future change won't break this.

I am not sure whether I understand your question correctly, happy to have a chat!

PR #6589, opened 2026-01-20
fix: better name and desc for code mode

alexhancock:
Better name and description for for the code mode extension

Draft for now. Need to think of a way for it to update everyone's config files actually as the desc gets saved there. I need to do another change later to make it do a migration.

### DIFF
--- a/crates/goose/src/agents/code_execution_extension.rs
+++ b/crates/goose/src/agents/code_execution_extension.rs
@@ -434,7 +434,7 @@ impl CodeExecutionClient {
             },
             server_info: Implementation {
                 name: EXTENSION_NAME.to_string(),
-                title: Some("Code Execution".to_string()),
+                title: Some("Code Mode".to_string()),
                 version: "1.0.0".to_string(),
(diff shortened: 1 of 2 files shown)


DOsinga:
Should we even mention javascript? might be confusing to some users (do I need to know JS?)

alexhancock:
@DOsinga Good point. I need to work out some sort of migrator before I can do this so it edits the entry in config, but will remove ref to JavaScript before moving from draft -> ready.

michaelneale:
another thought - rename the whole thing (even the id/key) so it is like a brand new (default on) extension.

So we change: 

```
pub static EXTENSION_NAME: &str = "code_execution";
```

```
pub static EXTENSION_NAME: &str = "code_mode";
```

etc - and it is on by default. I guess challenge then is there is a dead config for code_execution so would need to handle that 'migration' by explicly ignoring or removing it? this lets these thigns default to on.

alexhancock:
Closing this for a new approach that also adds a migrator

PR #6603, opened 2026-01-21
feat: a thinking tool (sometimes called an oracle)

michaelneale:
This is a simple built in tool which uses an alternative model via a provider to formulate a plan or think. A common pattern used by successful agents when you have multiple models available or models specialising in reasoning (and keeps it out of the main agent loop)


codefromthecrypt (commented):
I like this, the following is just things that come to mind looking at a bunch of agents.

1. this seems like a "plan" tool, and so the planning/thinking model makes more sense in my mind anyway
2. MCP has sampling, and it has the ability to request a specific model right now... so I wonder if it is always required to use a different provider?
3. since in many cases your current provider has access to a thinking model or already is one, I wonder if we can default this better to prefer the same provider. this also makes tests easier.
4. If we need to use multiple providers at once, we could have a goose convention like provider--model or some other unlikely delimited to switch to another one. That way the sampling or otherwise request doesn't need to think multi-provider

None of these things have to happen, though I do think the naming of oracle probably while neat could change and be easier to guess what it is. Besides the name itself all the other things above are impleentation details we can do now or later without disrupting the user.

michaelneale:
yeah good point on potentially re-using the single provider (most common config). Another thought I had was to break this out as a discrete MCP (outside of goose) which can work anywhere which needs thinking. With all my knowledge encoded in it, like the one ring. I think I have said too much already...

Issue #6556, opened 2026-01-17
Feature Request: Full Support for MCP Roots Protocol in Goose


bioinfornatics:
## Summary

Please add **full support for the MCP “Roots” capability** in the Goose client, as defined in the official Model Context Protocol documentation:

👉 https://modelcontextprotocol.io/docs/learn/client-concepts#roots

At the moment, Goose can launch `@modelcontextprotocol/server-filesystem`, but it **does not advertise or manage MCP roots**. As a result, when the filesystem server is started without static directory arguments, it receives **no allowed directories**, and all file operations are denied.

This prevents the use of **dynamic roots**, which are a core concept of MCP.

---

## Problem Statement

According to the MCP specification:

- **Roots are owned and controlled by the client**, not the server.
- Servers (like `server-filesystem`) rely on the client to send roots dynamically.
- If no roots are provided, filesystem access must be denied for security reasons.

In Goose today:

- The filesystem server is launched successfully.
- The server exposes the Roots capability.
- **Goose never sends any roots**, resulting in:
  - `list_allowed_directories` returning an empty array
  - `write_file` / `read_file` failing with  
    `Access denied - path outside allowed directories`

This makes the documented “dynamic roots” mode unusable.

---

### 1. Roots Discovery

Goose should determine one or more workspace/project roots, for example:

- The currently opened folder
- The active workspace
- User-configured project directories

### 2. Roots Advertisement (MCP)

Goose should notify MCP servers of root changes using the **Roots protocol**, e.g.:

```json
{
  "jsonrpc": "2.0",
  "method": "roots/list_changed",
  "params": {
    "added": [
      {
        "uri": "file:///home/user/projects/my-project",
        "name": "my-project"
      }
    ],
    "removed": []
  }
}
```

This must happen:
  - On initial connection
  - When the active project/workspace changes<

### 3. Dynamic Updates
When the user switches projects, Goose should update roots accordingly.
Filesystem servers should immediately reflect the new allowed directories.

## Acceptance Criteria

Requirement | Status
-- | --
Goose detects workspace/project roots | ❌
Goose sends roots/list_changed to MCP servers | ❌
Filesystem server receives non-empty roots | ❌
File read/write works without static CLI args | ❌
Roots update on workspace switch | ❌
Automated test covering roots negotiation | ⭕

## Why This Matters
Implementing MCP Roots support in Goose would:
- Enable **secure, dynamic filesystem access**
- Remove the need for static CLI arguments when starting servers<
[...]

PR #6612, opened 2026-01-21
fix: actually set the working dir for extensions from session

### LINKED ISSUE #6610: Bug: Working directory not passed to ExtensionManager when extensions are loaded
## Summary

When a user changes the working directory via the folder picker in Goose Desktop, the session's `working_dir` is updated but extensions are loaded using the **goose-server process's current directory** instead of the session's working directory. This causes shell commands and other extension operations to run in the wrong directory.

Image

Image

## User Report

Discord thread: https://discord.com/channels/1287729918100246654/1287729920319033345/1463483040511557795

The user reported that:
- The folder picker in the UI showed `.../ios_development/dialysispal_v02/`
- But Goose responded: "I see - it looks like your current working directory shows `dialysispal_v02` but we're actually operating in `/Users/alex/Downloads`"

### Diagnostics

From the user's diagnostics export (`diagnostics_20260121_4.zip`):
- **Session `working_dir`**: `/Users/alex/Projects/ios_development/dialysispal_v02`
- **`<info-msg>` injected into conversation**: `Working directory: /Users/alex/Downloads`

[diagnostics_20260121_4.zip](https://github.com/user-attachments/files/24766173/diagnostics_20260121_4.zip)

The mismatch confirms the bug.

## Root Cause Analysis

The bug is in `crates/goose/src/agents/extension_manager.rs`:

```rust
async fn resolve_working_dir(&self) -> PathBuf {
    // Fall back to current_dir - working_dir is passed through the call chain from session
    std::env::current_dir().unwrap_or_default()
}
```

The comment says "working_dir is passed through the call chain from session" but **it is not actually passed through**. The method just returns `std::env::current_dir()` which is the goose-server process's working directory, not the session's.

### Call chain that needs fixing:

```
load_extensions_from_session(session) 
  → agent.add_extension(config)  // has access to session.working_dir but doesn't pass it
    → extension_manager.add_extension(config)  // needs working_dir parameter
      → resolve_working_dir()  // currently returns wrong directory
```

## Impact

- Shell commands run in the wrong directory
- The `<info-msg>` shows the wrong working directory to the model
- Memory extension looks for `.goose/memory` in the wrong location
- Any extension that uses `GOOSE_WORKING_DIR` env var gets the wrong path

## Suggested Fix

Either:

1. **Add `working_dir` field to `ExtensionManager`** and update it when session changes
2. **Pass `working_dir` explicitly through the call chain** from `load_extensions_from_session` → `add_extension` → `extension_manager.add_extension`

The key files to modify:
- `crates/goose/src/agents/extension_manager.rs` - `add_extension()` and `resolve_working_dir()`
- `crates/goose/src/agents/agent.rs` - `load_extensions_from_session()` and `add_extension()`

## Environment

- App Version: 1.20.1
- OS: macOS
- Model: GLM-4.7 (via custom provider)

zanesq:
## Summary
The root cause was two-fold: first, the ExtensionManager::add_extension method wasn't accepting a working directory parameter, so the session's working_dir was never propagated from Agent::load_extensions_from_session. Second, for builtin extensions like developer that run in-process (not as child processes), the GOOSE_WORKING_DIR environment variable wasn't being set in the current process's environment—it was only being set for child process extensions. The fix adds a working_dir parameter through the extension loading chain and ensures GOOSE_WORKING_DIR is set in the current process when loading builtin extensions, so that when you change the working directory in the UI and extensions are reloaded, the developer extension's shell commands now correctly use the new directory.

image

closes https://github.com/block/goose/issues/6610

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -634,6 +634,7 @@ impl Agent {
 
     /// Load extensions from session into the agent
     /// Skips extensions that are already loaded
+    /// Uses the session's working_dir for extension initialization
     pub async fn load_extensions_from_session(
         self: &Arc<Self>,
         session: &Session,
@@ -651,11 +652,15 @@ impl Agent {
             }
         };
 
+        // Capture the session's working_dir to pass to extensions
+        let working_dir = session.working_dir.clone();
+
         let extension_futures = enabled_configs
             .into_iter()
             .map(|config| {
                 let config_clone = config.clone();
                 let agent_ref = self.clone();
+                let working_dir_clone = working_dir.clone();
 
                 async move {
                     let name = config_clone.name().to_string();
@@ -674,7 +679,10 @@ impl Agent {
                         };
                     }
 
-                    match agent_ref.add_extension(config_clone).await {
+                    match agent_ref
+                        .add_extension_with_working_dir(config_clone, Some(working_dir_clone))
+                        .await
+                    {
                         Ok(_) => ExtensionLoadResult {
                             name,
                             success: true,
@@ -698,6 +706,14 @@ impl Agent {
     }
 
     pub async fn add_extension(&self, extension: ExtensionConfig) -> ExtensionResult<()> {
+        self.add_extension_with_working_dir(extension, None).await
+    }
+
+    pub async fn add_extension_with_working_dir(
+        &self,
+        extension: ExtensionConfig,
+        working_dir: Option<std::path::PathBuf>,
+    ) -> ExtensionResult<()> {
         match &extension {
             ExtensionConfig::Frontend {
                 tools,
@@ -726,7 +742,7 @@ impl Agent {
             }
             _ => {
                 self.extension_manager
-                    .add_extension(extension.clone())
+                    .add_extension_with_working_dir(extension.clone(), working_dir)
                     .await?;
             }
         }
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -466,13 +466,6 @@ impl ExtensionManager {
         &self.context
     }
 
-    /// Resolve the working directory for an extension.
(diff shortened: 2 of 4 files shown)


DOsinga:
I think we should use the roots feature of MCP for the developer tool and then implement this for all extensions

zanesq:
Agreed we should follow up with that refactor

zanesq:
Rust test failure was pre-existing on the `goose-acp` crate

```
thread 'test_permission_persistence::allow_always' (4774) has overflowed its stack
error: test failed, to rerun pass `-p goose-acp --test server_test`

Caused by:
  process didn't exit successfully: `/home/runner/work/goose/goose/target/debug/deps/server_test-d0cb26eab9d9a4eb --skip 'scenario_tests::scenarios::tests'` (signal: 6, SIGABRT: process abort signal)
Error: Process completed with exit code 101.
```

jamadeo:
> I think we should use the roots feature of MCP for the developer tool and then implement this for all extensions

Sounds so much like the right way to do it I'm not sure why we haven't already

DOsinga:
> > I think we should use the roots feature of MCP for the developer tool and then implement this for all extensions
> 
> Sounds so much like the right way to do it I'm not sure why we haven't already

turns out roots should restrict the MCP to the working_dir. is that what we imagine a working_dir does? maybe? /cc @alexhancock

alexhancock:
Yes, I think it could make sense to group https://github.com/block/goose/issues/6556 with fixing this problem

@zanesq Want to give a try at adding roots support? We would have the MCP client respond to `roots/list` with the goose working dir. Then have the built in servers: developer/etc request it.

I had thought I'd do this this week, so either way, let me know!

zanesq:
Sure I can take a crack at it this week, can we merge this one for now to go with multi chat? Verified it does fix the issue in its current state.

DOsinga (approved):
good fix, but I think it also shows the brittleness of the current plumbing

PR #6616, opened 2026-01-21
Speed up Databricks provider init by removing fetch of supported models

jh-block:
## Summary

The list of supported models was being fetched from Databricks every time the provider was initialised, taking around 1.6s on my machine. This shows up to the user as a 1.6s delay before the first model response can start to be received.

We only needed this list to determine if our preferred "fast model", used for compacting and for generating the titles of sessions (currently Gemini 2.5 Flash) was supported. If it isn't supported, we fall back to the main model. But we already have a fallback to the main model implemented if calling the fast model fails (e.g. because it's not found), so there's no need to check upfront.

This PR removes the fetch of supported models and assumes the fast model is available, relying on the already implemented fallback to the main model if it turns out not to be. On my machine, it reduces the time to first token after sending the first prompt from 1.6s to around 200ms.

I also updated the Playwright tests (which were failing due to drift between the actual UI structure and what the tests expected) and added some performance tests which were used to find the issue being fixed in the PR.

It may be easier to review by commit.

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [x] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

- [x] This PR was created or reviewed with AI assistance (Claude Code)

### Testing

I manually tested to verify the performance improvement.

### DIFF
--- a/crates/goose/src/providers/databricks.rs
+++ b/crates/goose/src/providers/databricks.rs
@@ -134,7 +134,6 @@ impl DatabricksProvider {
         let api_client =
             ApiClient::with_timeout(host, auth_method, Duration::from_secs(DEFAULT_TIMEOUT_SECS))?;
 
-        // Create the provider without the fast model first
         let mut provider = Self {
             api_client,
             auth,
@@ -143,31 +142,7 @@ impl DatabricksProvider {
             retry_config,
             name: Self::metadata().name,
         };
-
-        // Check if the default fast model exists in the workspace
-        // Generate UUID for this initialization request since no user session exists yet
-        let session_id = uuid::Uuid::new_v4().to_string();
-        let model_with_fast =
-            if let Ok(Some(models)) = provider.fetch_supported_models(&session_id).await {
-                if models.contains(&DATABRICKS_DEFAULT_FAST_MODEL.to_string()) {
-                    tracing::debug!(
-                        "Found {} in Databricks workspace, setting as fast model",
-                        DATABRICKS_DEFAULT_FAST_MODEL
-                    );
-                    model.with_fast(DATABRICKS_DEFAULT_FAST_MODEL.to_string())
-                } else {
-                    tracing::debug!(
-                        "{} not found in Databricks workspace, not setting fast model",
-                        DATABRICKS_DEFAULT_FAST_MODEL
-                    );
-                    model
-                }
-            } else {
-                tracing::debug!("Could not fetch Databricks models, not setting fast model");
-                model
-            };
-
-        provider.model = model_with_fast;
+        provider.model = model.with_fast(DATABRICKS_DEFAULT_FAST_MODEL.to_string());
         Ok(provider)
     }
 


zanesq:
Thanks! I tested the desktop app locally and its working well and does seem faster to load a chat.

I also tested the e2e changes and there are still failing tests for me but overall its an improvement so I don't think it should block this PR and we can follow up with fixing the remaining tests.

A major change is it launches the electron app from scratch with each test whereas before it would use the same running electron app so that has tradeoffs like taking longer to run the entire suite. Ultimately I think we should pair these tests down to only what we need for a release smoke test from our checklist so we can automate them on releases next. The rest of the functionality can be covered by our vitest unit/functional tests.

jh-block:
> A major change is it launches the electron app from scratch with each test whereas before it would use the same running electron app so that has tradeoffs like taking longer to run the entire suite. Ultimately I think we should pair these tests down to only what we need for a release smoke test from our checklist so we can automate them on releases next. The rest of the functionality can be covered by our vitest unit/functional tests.

Yeah, one advantage of the isolation is that we could potentially allow playwright to run the tests in parallel without the tests interfering with each other. It is necessary for the performance test to have a fresh instance to give more repeatable results (so that the test is not sometimes testing a cold launch and sometimes a warm instance) but we could allow the other tests to reuse instances if desired.

crates/goose/src/providers/databricks.rs
```diff
@@ -144,30 +144,11 @@ impl DatabricksProvider {
...
-            } else {
-                tracing::debug!("Could not fetch Databricks models, not setting fast model");
-                model
-            };
-
-        provider.model = model_with_fast;
+        tracing::debug!(
+            "Setting {} as fast model (will fall back to main model if unavailable)",
+            DATABRICKS_DEFAULT_FAST_MODEL
+        );
```
DOsinga: maybe delete that log statement. same as w the rest, we'll see when it happens

ui/desktop/tests/e2e/app.spec.ts
```diff
@@ -28,23 +25,23 @@ const test = base.extend<TestFixtures>({
   provider: [providers[0], { option: true }], // Default to first provider (Databricks)
 });
 
-// Store mainWindow reference
-let mainWindow;
+// Store mainWindow reference (will be set from goosePage in beforeEach)
```
DOsinga: in beforeEach?

ui/desktop/tests/e2e/fixtures.ts
```diff
@@ -0,0 +1,170 @@
...
+  goosePage: async ({}, use, testInfo) => {
+    console.log(`Launching fresh Electron app for test: ${testInfo.title}`);
+
+    let appProcess: ChildProcess | null = null;
+    let browser: Browser | null = null;
+
+    try {
+      // Assign a unique debug port for this test to enable parallel execution
+      // Base port 9222, offset by worker index * 100 + parallel slot
+      const debugPort = 9222 + (testInfo.parallelIndex * 10);
```
DOsinga, line 40: can we not just find a free port here? we're in node land, right?

jh-block: we're not the ones binding the port, that's happening in electron after we spawn it (but we need to choose it here so that we know the port to connect to later).

so, even if we wrote the code here to find a free port, there'd be a race condition as someone could bind it before electron starts up and devtools binds it. I'm not sure it would be overall better.

PR #6619, opened 2026-01-21
Native images

DOsinga:
## Summary

Upload images directly. Also does a better job on pasting images in:

image


zanesq (approved):
Nice good improvement! Code LGTM and verified its working as intended uploading and processing images in chat locally.

PR #6620, opened 2026-01-21
fix: macOS keychain infinite prompt loop   

### LINKED ISSUE #6595: [Bug] macOS Keychain infinite prompt loop - 'Allow' vs 'Always Allow' causes corrupted state
## Describe the bug

On macOS, users experience an **infinite keychain password prompt loop** when using Goose Desktop with GitHub Copilot provider. The prompts continue indefinitely until the user clicks "Always Allow" — clicking just "Allow" causes prompts to recur on every keychain operation.

**Critical finding**: If user clicks "Allow" first (instead of "Always Allow"), they enter a corrupted state that requires **two** subsequent "Always Allow" clicks to recover.

## Root Cause Analysis

We traced through the code and confirmed via extensive testing:

### Why the loop happens

1. **CLI and Desktop use different binaries** with different code signatures:
   - CLI: `~/.local/bin/goose` (cdhash-based ACL, unsigned)
   - Desktop: `/Applications/Goose.app/Contents/Resources/bin/goosed` (Team ID-based ACL, signed by Block Inc.)

2. When CLI authenticates first, the keychain entry's ACL only includes the CLI's cdhash. Desktop's `goosed` binary has a different signature and isn't authorized.

3. **Each keychain operation is a separate prompt** when user clicks "Allow":
   - `get_secret()` = 1 READ operation
   - `set_secret()` = 1 READ + 1 WRITE operation (reads existing secrets, then writes updated)

4. **Retry loops compound the problem**:
   ```
   with_retry (3 retries) — providers/retry.rs
       └─ post() 
           └─ get_api_info() (3 internal retries)
               └─ refresh_api_info()
                   └─ config.get_secret() ← KEYCHAIN READ
   ```

### Test Results

| First Click | Subsequent Clicks | Result |
|-------------|-------------------|--------|
| **Always Allow** | — | ✅ Works immediately |
| **Allow** | Allow (20+ times) | ❌ Infinite loop |
| **Allow** | Always Allow | ❌ Fails first time |
| **Allow** | Always Allow (2nd) | ✅ Works |

The "two Always Allow clicks needed" behavior occurs because:
1. First "Always Allow" authorizes the **read** operation
2. Second "Always Allow" authorizes the **write** operation (since `set_secret()` does read+write)

### ACL Evidence

After clicking "Always Allow", both binaries are properly authorized:
```
applications (2):
    0: /Applications/Goose.app/Contents/Resources/bin/goosed (OK)
        requirement: identifier goosed and anchor apple generic and certificate leaf[subject.OU] = EYF346PHUG
    1: /Users/jacobhales/.local/bin/goose (OK)
        requirement: cdhash H"3443534d499f2ae1d2cc22fe2571c1f66a69085a"
```

Note: Desktop uses **Team ID-based requirement** (stable across updates), while CLI uses **cdhash** (breaks on updates).

### Fix 1: Cache OAuth token in memory (eliminates retry prompts)

```rust
// In GithubCopilotProvider struct, add:
oauth_token: tokio::sync::OnceCell<String>,

// In refresh_api_info(), cache after first read:
let token = self.oauth_token.get_or_try_init(|| async {
    config.get_secret::<String>("GITHUB_COPILOT_TOKEN")
}).await?;
```

### Fix 2: Pre-keychain UX guidance

Before triggering the first keychain access, show an in-app dialog:
> "macOS will ask for keychain access. Click **'Always Allow'** to avoid repeated prompts."

### Fix 3: Document the CLI→Desktop workflow

If users authenticate via CLI first, they need to click "Always Allow" when Desktop prompts (since it's a different binary).

## To Reproduce

1. Clear keychain: `security delete-generic-password -s goose`
2. Clear cache: `rm -rf ~/.config/goose/githubcopilot/`
3. Run `goose configure` → select GitHub Copilot → complete OAuth
4. Launch Goose Desktop
5. When keychain prompt appears, click **"Allow"** (not "Always Allow")
6. Observe: prompts continue indefinitely (20+ tested)

**To escape**: Click "Always Allow" twice (once for read, once for write)

## Expected Behavior

- Single keychain prompt on first access
- Clear guidance to click "Always Allow"
- No prompts after authorization (until app update changes binary signature)

## Environment

- **OS & Arch**: macOS 15.2 (Sequoia) arm64
- **Interface**: Desktop + CLI
- **Version**: v1.20.1 (Homebrew)
- **Provider & Model**: GitHub Copilot (claude-sonnet-4.5)

## Additional Context

- Related closed issues: #1734, #4018, #1238 (all closed with "click Always Allow" but issue persists for users)
- Non-technical users don't understand "Allow" vs "Always Allow" difference
- The Desktop binary uses Team ID-based ACL (survives updates), but CLI uses cdhash (breaks on updates)

## References

- Keychain access: [`config/base.rs:542-560`](https://github.com/block/goose/blob/main/crates/goose/src/config/base.rs#L542) (`all_secrets()`)
- Set secret (read+write): [`config/base.rs:765-790`](https://github.com/block/goose/blob/main/crates/goose/src/config/base.rs#L765) (`set_secret()`)
- Retry loop: [`providers/githubcopilot.rs:200-221`](https://github.com/block/goose/blob/main/crates/goose/src/providers/githubcopilot.rs#L200) (`get_api_info()`)

sheikhlimon:
## Summary
**Problem**: Each call to `get_secret()` hits keychain separately, causing multiple prompts when loading provider config with multiple secret parameters. 
**Solution**: Add in-memory cache to `all_secrets()` so keychain is accessed once per config load.
 - Before: N keychain prompts for N secret parameters when loading provider config
 - After: 1 keychain prompt per config load, cached for subsequent reads in same session

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Related Issues
Fixes #6595

### DIFF
--- a/crates/goose/src/config/base.rs
+++ b/crates/goose/src/config/base.rs
@@ -547,7 +547,6 @@ impl Config {
         let mut cache = self.secrets_cache.lock().unwrap();
 
         let values = if let Some(ref cached_secrets) = *cache {
-            tracing::debug!("secrets cache hit");
             cached_secrets.clone()
         } else {
             tracing::debug!("secrets cache miss, fetching from storage");
@@ -578,7 +577,6 @@ impl Config {
             };
 
             *cache = Some(loaded.clone());
-            tracing::debug!("secrets cached");
             loaded
         };
 
@@ -888,7 +886,6 @@ impl Config {
     fn invalidate_secrets_cache(&self) {
         let mut cache = self.secrets_cache.lock().unwrap();
         *cache = None;
-        tracing::debug!("secrets cache invalidated");
     }
 
     /// Check if an error string indicates a keyring availability issue that should trigger fallback


michaelneale (approved):
I think this would be very nice @codefromthecrypt looks ok to you (once it is a clean build)

sheikhlimon:
> For this code.. I think there's a potential lazy loading race (e.g., multiple threads could check None outside the lock and all try to load, causing redundant hits).

Thanks for reviewing this!  I’ve implemented your suggested fix by holding the lock across the entire cache check and load operation.
This prevents multiple threads from observing `None` and loading simultaneously. The first thread to acquire the lock performs the load, and subsequent callers wait and then use the cached result.

codefromthecrypt:
thanks again for the help!

PR #6625, opened 2026-01-22
Swap canonical model from openrouter to models.dev

katzdave:
## Summary
Makes the lookups a lot more straightforward since models.dev has data keyed by (provider, model). Gets rid of lots of hacky text parsing. Still leaving in some level of text parsing for 'meta-providers' like databricks and bedrock. Sort by release date (works great!).

### Testing

Tested Model filtering, sorting cost with:
- [ x ] OpenAI
- [ x ] Databricks
- [ x ] Anthropic
- [ x ] Google
- [ x ] xai
- [ x ] OpenRouter
- [] Tetrate (need key)

### Coming in Followup

- Use canonical model for context limits. Want to aggressively delete a bunch of code here + verify with some oneoff scripts that we don't lose context data for major providers.

### DIFF
--- a/crates/goose/src/providers/xai.rs
+++ b/crates/goose/src/providers/xai.rs
@@ -178,46 +178,32 @@ impl Provider for XaiProvider {
     }
 
     async fn fetch_supported_models(&self) -> Result<Option<Vec<String>>, ProviderError> {
-        let response = match self.api_client.response_get(None, "models").await {
-            Ok(response) => response,
-            Err(e) => {
-                tracing::warn!(
-                    "Failed to fetch models from XAI API: {}, falling back to known models",
-                    e
-                );
-                let models: Vec<String> = XAI_KNOWN_MODELS.iter().map(|s| s.to_string()).collect();
-                return Ok(Some(models));
-            }
-        };
+        let response = self.api_client.response_get(None, "models").await?;
+        let json = handle_response_openai_compat(response).await?;
 
-        let json = match handle_response_openai_compat(response).await {
-            Ok(json) => json,
-            Err(e) => {
-                tracing::warn!(
-                    "Failed to parse XAI models response: {}, falling back to known models",
-                    e
-                );
-                let models: Vec<String> = XAI_KNOWN_MODELS.iter().map(|s| s.to_string()).collect();
-                return Ok(Some(models));
-            }
-        };
-
-        if let Some(data) = json.get("data").and_then(|v| v.as_array()) {
-            let models: Vec<String> = data
-                .iter()
-                .filter_map(|m| {
-                    m.get("id")
-                        .and_then(|id| id.as_str())
-                        .map(|s| s.to_string())
-                })
-                .collect();
-            if !models.is_empty() {
-                return Ok(Some(models));
-            }
+        if let Some(err_obj) = json.get("error") {
+            let msg = err_obj
+                .get("message")
+                .and_then(|v| v.as_str())
+                .unwrap_or("unknown error");
+            return Err(ProviderError::Authentication(msg.to_string()));
         }
 
-        // Fallback to known models if API doesn't return what we expect
-        let models: Vec<String> = XAI_KNOWN_MODELS.iter().map(|s| s.to_string()).collect();
-        Ok(Some(models))
+        let data = json.get("data").and_then(|v| v.as_array());
+        match data {
+            Some(arr) => {
+                let mut models: Vec<String> = arr
+                    .iter()
+                    .filter_map(|m| {
+                        m.get("id")
+                            .and_then(|id| id.as_str())
+                            .map(|s| s.to_string())
+                    })
+                    .collect();
+                models.sort();
+                Ok(Some(models))
+            }
+            None => Ok(None),
+        }
     }
 }


crates/goose-server/src/routes/config_management.rs
```diff
@@ -505,17 +505,17 @@ pub async fn get_pricing(
...
-    ) {
+    if let (Some(input_cost), Some(output_cost)) =
+        (canonical_model.cost.input, canonical_model.cost.output)
+    {
         pricing_data.push(PricingData {
             provider: query.provider.clone(),
             model: query.model.clone(),
-            input_token_cost: input_cost,
-            output_token_cost: output_cost,
+            // Canonical model costs are per million tokens, convert to per-token
```
DOsinga, line 514: needing this comment suggests that we need to refactor this more either at the beginning or at the end - let's have one unit we pass around

crates/goose/src/providers/canonical/model.rs
```diff
@@ -1,53 +1,102 @@
...
+/// Modalities supported by a model (mirrors models.dev structure)
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct Modalities {
+    /// Input modalities (e.g., ["text", "image", "pdf"])
+    #[serde(default)]
+    pub input: Vec<String>,
+
+    /// Output modalities (e.g., ["text"])
+    #[serde(default)]
+    pub output: Vec<String>,
```
DOsinga: can we make these enums instead of string, ignoring the ones we don't know about?

crates/goose/src/providers/canonical/model.rs
```diff
@@ -1,53 +1,102 @@
...
+}
+
+/// Pricing/cost information for a model
```
DOsinga: remove

crates/goose/src/providers/canonical/model.rs
```diff
@@ -1,53 +1,102 @@
...
+/// Token limits for a model
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct Limit {
+    /// Maximum context window size in tokens
+    pub context: usize,
 
-    /// Cost per image
+    /// Maximum output/completion tokens
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub image: Option<f64>,
```
DOsinga, line 20: do we not have this anymore? we're probably not using it anyway, but still

crates/goose/src/providers/canonical/model.rs
```diff
@@ -1,53 +1,102 @@
...
+    pub output: Option<usize>,
 }
 
-/// Canonical representation of a model
+/// Canonical representation of a model (mirrors models.dev API structure)
```
DOsinga: remove addition to comment

crates/goose/src/providers/canonical/model.rs
```diff
@@ -1,53 +1,102 @@
...
     #[serde(default)]
-    pub output_modalities: Vec<String>,
+    pub modalities: Modalities,
 
-    /// Whether the model supports tool calling
+    /// Whether the model has open weights
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub open_weights: Option<bool>,
+
+    /// Pricing information
```
DOsinga, line 113: remove

crates/goose/src/providers/utils.rs
```diff
@@ -90,6 +90,7 @@ fn check_context_length_exceeded(text: &str) -> bool {
         "max_tokens",
         "decrease input length",
         "context limit",
+        "maximum prompt length",
```
DOsinga, line 93: did we run into this? where? the string "too long" above certainly has no place here

crates/goose/src/providers/xai.rs
```diff
@@ -176,4 +176,48 @@ impl Provider for XaiProvider {
...
+                        .and_then(|id| id.as_str())
+                        .map(|s| s.to_string())
+                })
+                .collect();
+            if !models.is_empty() {
+                return Ok(Some(models));
+            }
+        }
+
+        // Fallback to known models if API doesn't return what we expect
```
DOsinga: I'm not sure we want to do that. I just merged something that has can't reach the provider - if this happens either the provider is down or the user is using some alternate host. we should just bubble up the error or return [] - whatever it is we do elsewhere

ui/desktop/src/api/types.gen.ts
```diff
@@ -572,15 +572,15 @@ export type ModelInfo = {
...
+     * Cost per token for input in USD (optional)
      */
     input_token_cost?: number | null;
     /**
      * The name of the model
      */
     name: string;
     /**
-     * Cost per token for output (optional)
+     * Cost per token for output in USD (optional)
```
DOsinga, line 583: can we remove (optional) in the comment?

crates/goose/src/providers/utils.rs
katzdave: hit this with x.ai testing.

katzdave: A couple of these definitely feel over generic, but I'm a little hesitant to change them. iirc too long was the databricks issue but don't remember the full error message.

ui/desktop/src/api/types.gen.ts
katzdave: done

crates/goose-server/src/routes/config_management.rs
katzdave: My followup pr deletes the pricing API and replaces it with a full canonical model fetch (or at least just the fields the client needs, but gives us a place to layer on more). will keep as cost / mtokens until we render.

crates/goose/src/providers/canonical/model.rs
katzdave: Yeah don't have it anymore. Definitely weren't using it, I had just pulled all the fields in from openrouter, but still a slight loss.

crates/goose/src/providers/canonical/model.rs
katzdave: Done

crates/goose/src/providers/canonical/model.rs
katzdave: Done

PR #6637, opened 2026-01-22
Remove dependency on goose-mcp from goose crate

jamadeo:
goose-mcp contains all of our "built-in" mcp servers. At present we do compile these into the main goose binary (goose-server and goose-cli), but better to keep the goose crate itself independent, as other binaries are built on it that don't need to drag in all of the various developer/analysis tool dependencies.

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -565,9 +565,11 @@ impl ExtensionManager {
             }
             ExtensionConfig::Builtin { name, timeout, .. } => {
                 let timeout_duration = Duration::from_secs(timeout.unwrap_or(300));
-                let extension_fn = get_builtin_extension(name.as_str()).ok_or_else(|| {
-                    ExtensionError::ConfigError(format!("Unknown builtin extension: {}", name))
-                })?;
+                let normalized_name = name_to_key(name);
+                let extension_fn =
+                    get_builtin_extension(normalized_name.as_str()).ok_or_else(|| {
+                        ExtensionError::ConfigError(format!("Unknown builtin extension: {}", name))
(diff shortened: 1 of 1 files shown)


crates/goose/src/execution/manager.rs
```diff
@@ -44,6 +44,22 @@ impl AgentManager {
...
+                let schedule_file_path = Paths::data_dir().join("schedule.json");
+                let session_manager = Arc::new(SessionManager::instance());
+                let manager =
+                    Self::new(session_manager, schedule_file_path, Some(max_sessions)).await?;
+                Ok(Arc::new(manager))
+            })
+            .await
+            .cloned()
+    }
+
```
jh-block: This new `initialize()` method seems to be identical to the `instance()` method below. Could probably just use `instance()` wherever we're calling this?

crates/goose/src/builtin_extension.rs
```diff
@@ -0,0 +1,24 @@
...
+}
+
+/// Register multiple builtin extensions from a HashMap
+pub fn register_builtin_extensions(extensions: HashMap<&'static str, SpawnServerFn>) {
+    let mut registry = BUILTIN_REGISTRY.write().unwrap();
+    registry.extend(extensions);
+}
+
+/// Get a copy of all registered builtin extensions
+pub fn get_builtin_extensions() -> HashMap<&'static str, SpawnServerFn> {
```
jh-block: How often is this called? Could potentially wrap the `HashMap` in an `Arc` (inside the `RwLock`) to make cloning it ~free, or provide a `get_builtin_extension(name: &'static str) -> SpawnServerFn` so that we don't have to return the whole registry.

PR #6646, opened 2026-01-22
Strip the audience for compacting

DOsinga:
## Summary

We weren't stripping audience = user tags.

my conversation to be compacted came in at 229,887 before, now it is 128,516 - to be honest seems like almost too steep a change


DOsinga:
I think the format code is actually wrong; the LLMs are supposed to ignore this already so it should all be good to send them this; just dont want to compact it.

PR #6657, opened 2026-01-23
fix(goose): only send agent-session-id when a session exists

codefromthecrypt:
## Summary
Stop adding the `agent-session-id` header unless there is a current session. This keeps headers 1:1 with session manager records and avoids confusing or conflicting session IDs.

`None` is for configuration or pre-session paths (model discovery, provider detection, auth probes, canonical model checks).

This also backfills propagation of non-basic OpenAI apis so they can also see `agent-session-id` (e.g., Anthropic, Bedrock, Vertex etc). This is important because proxies like Envoy AI Gateway and Ollama started supporting non-OpenAI endpoints recently. With this, they can do log correlation on goose session ID.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [x] Refactor / Code quality
- [x] Tests
- [ ] Performance improvement
- [ ] Documentation
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Added some edge case tests.

### Related Issues
Tuneup of #6626

### DIFF
--- a/crates/goose/src/providers/base.rs
+++ b/crates/goose/src/providers/base.rs
@@ -368,9 +368,12 @@ pub trait Provider: Send + Sync {
 
     // Internal implementation of complete, used by complete_fast and complete
     // Providers should override this to implement their actual completion logic
+    //
+    /// # Parameters
+    /// - `session_id`: Use `None` only for configuration or pre-session tasks.
     async fn complete_with_model(
         &self,
-        session_id: &str,
+        session_id: Option<&str>,
         model_config: &ModelConfig,
         system: &str,
         messages: &[Message],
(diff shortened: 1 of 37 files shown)


crates/goose/src/providers/base.rs
```diff
@@ -368,9 +368,12 @@ pub trait Provider: Send + Sync {
 
     // Internal implementation of complete, used by complete_fast and complete
     // Providers should override this to implement their actual completion logic
+    //
+    /// # Parameters
+    /// - `session_id`: Use `None` only for configuration or pre-session tasks.
     async fn complete_with_model(
```
codefromthecrypt, line 374: this is the key place though some others had to switch to Option to support pre-session tasks

PR #6658, opened 2026-01-23
feat: smart start directories and files

michaelneale:
Detect what user has been working on their desktop (outside of the app), and suggest where they go to start as default experience:

image


lifeizhou-ap:
Hey @michaelneale 👋 Just checking in on this draft. Are you still actively working on it? If not, would you mind closing it? We can always reopen later if you pick it back up. Thanks!

PR #6662, opened 2026-01-23
Optimise load config in UI

lifeizhou-ap:
## Summary

1. Optimise the load time when app starts or new session starts
   - Reduce the call to check ELEVENLAB_API_KEY exists or not. This call is expensive as takes around 3s because it checks keychain at the backend
       - Consolidate dictation settings retrieval to eliminate duplicate hook calls which makes the duplicated api call
       - Only check ELEVENLAB_API_KEY when the provider is eleven_lab, otherwise no need to check
       - Add module-level caching for ElevenLabs API key checks to avoid redundant keychain reads
      
   -  Fix getProviders callback memoization in ConfigContext using a ref to prevent unnecessary recreations

2. Enhanced user experience to show the puzzle of current session extension only after the extensions are loaded.

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [X] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing

Manual testing by adding sleep in the `/config/read` endpoint

**Before**
https://github.com/user-attachments/assets/2818649f-fce3-48de-806f-861de3612d74

**After**
https://github.com/user-attachments/assets/56546fdc-c722-4d5a-a7b5-5adde2f66b19

### DIFF
--- a/crates/goose-server/src/routes/config_management.rs
+++ b/crates/goose-server/src/routes/config_management.rs
@@ -222,11 +222,6 @@ fn mask_secret(secret: Value) -> String {
 pub async fn read_config(
     Json(query): Json<ConfigKeyQuery>,
 ) -> Result<Json<ConfigValueResponse>, StatusCode> {
-    // TODO: Remove this - simulating slow keychain access for debugging
-    if query.is_secret {
-        tokio::time::sleep(std::time::Duration::from_secs(3)).await;
-    }
-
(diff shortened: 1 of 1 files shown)


zanesq:
Thanks!

PR #6682, opened 2026-01-23
fix: Manual compaction does not update context window.

katzdave:
## Summary
Simple fix, but added some pretty extensive tests with assertions about conversation state + token counts after all 3 compaction trigger types.

### DIFF
--- a/crates/goose/tests/compaction.rs
+++ b/crates/goose/tests/compaction.rs
@@ -118,8 +118,10 @@ impl Provider for MockCompactionProvider {
 
         // Simulate context limit: if input > 20k tokens and we haven't compacted yet, fail
         const CONTEXT_LIMIT: i32 = 20000;
-        if !is_compaction && input_tokens > CONTEXT_LIMIT
-            && !self.has_compacted.load(Ordering::SeqCst) {
+        if !is_compaction
+            && input_tokens > CONTEXT_LIMIT
+            && !self.has_compacted.load(Ordering::SeqCst)
+        {
             return Err(ProviderError::ContextLengthExceeded(format!(
                 "Context limit exceeded: {} > {}",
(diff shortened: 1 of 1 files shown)


crates/goose/src/agents/execute_commands.rs
```diff
@@ -98,6 +98,15 @@ impl Agent {
             .replace_conversation(session_id, &compacted_conversation)
             .await?;
 
+        let session_config = super::types::SessionConfig {
+            id: session_id.to_string(),
+            schedule_id: session.schedule_id,
+            max_turns: None,
+            retry_config: None,
+        };
```
DOsinga: uhm, that doesn't seem great. change update_session_metrics to just take a session_id or get your hands on the actual SessionConfig (I think we can do the first thing)

crates/goose/tests/agent.rs
```diff
@@ -567,4 +567,768 @@ mod tests {
             );
         }
     }
+
+    #[cfg(test)]
+    mod compaction_tests {
```
DOsinga: this is a lot of lines; should this live in its own test module maybe? also the setup of three tests seems similar; can we unify?

katzdave: Moved it. Lots of shared code on the setup of the provider and session + validations on post-compaction session state. Maybe could squeeze a little more with the messages, but I think the individual tests tell a clear story of the context before/after compaction and I am seeding the one that triggers out of context with extra messages.

crates/goose/src/agents/execute_commands.rs
katzdave: Done although, this schedule_id leaked out? We had it in all places so just continuing to pass it along.

PR #6685, opened 2026-01-24
Refactor: move persisting extension to session outside of route

zanesq:
## Summary
Follow up for https://github.com/block/goose/pull/6057#discussion_r2721522189 to move persisting the extension to the session outside of the route so its available to other code paths.

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -734,13 +734,18 @@ impl Agent {
         extension: ExtensionConfig,
         session_id: &str,
     ) -> ExtensionResult<()> {
-        let working_dir = self
+        let session = self
             .config
             .session_manager
             .get_session(session_id, false)
             .await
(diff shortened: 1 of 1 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -704,14 +704,30 @@ impl Agent {
...
     }
 
-    pub async fn add_extension(&self, extension: ExtensionConfig) -> ExtensionResult<()> {
-        self.add_extension_with_working_dir(extension, None).await
+    pub async fn add_extension(
+        &self,
+        extension: ExtensionConfig,
+        session_id: Option<&str>,
+    ) -> ExtensionResult<()> {
+        self.add_extension_internal(extension, None, session_id)
```
DOsinga: session_id should not be optional here. add_extension_with_working_dir and add_extension_internal are the same right now

this should all just be one thing - add extension with a session id. that session has your working dir.

PR #6711, opened 2026-01-26
Keyboard short cuts

DOsinga:
## Summary

Collects all our keyboard short cuts and makes them configurable:

image

### DIFF
--- a/ui/desktop/src/components/settings/keyboard/ShortcutRecorder.tsx
+++ b/ui/desktop/src/components/settings/keyboard/ShortcutRecorder.tsx
@@ -59,10 +59,11 @@ export function ShortcutRecorder({
       parts.push('Shift');
     }
 
-    let key = e.key;
+    let key = e.code && e.code.startsWith('Key') ? e.code.replace('Key', '') : e.key;
 
     const keyMap: Record<string, string> = {
       ' ': 'Space',
+      Space: 'Space',
       ArrowUp: 'Up',
       ArrowDown: 'Down',
       ArrowLeft: 'Left',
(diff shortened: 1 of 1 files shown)


zanesq:
When I run this branch locally I get the following error, I think it's related to the copilot review above? I put up a [PR](https://github.com/block/goose/pull/6716) with a fix from goose if you want to merge it into this branch.
```
08:45:24.213 › from renderer: [ERROR] ReferenceError: __dirname is not defined

    at Lazy (<anonymous>)
    at ErrorBoundary (http://localhost:5173/src/components/ErrorBoundary.tsx:89:5)
    at ConfigProvider (http://localhost:5173/src/components/ConfigContext.tsx:24:34)
    at Suspense (<anonymous>)

```

After this fix and running I tested the keyboard settings and it is putting a strange character when I tried to change focus goose window to simply `option + G` and it doesn't seem to work it still works with the original `alt + cmd + G`
image

Couple of other issues:

1. Noticed a bug where the enable/disable toggle is resetting it back to the original default command, maybe related to it not saving it above?

https://github.com/user-attachments/assets/7648b509-bd3f-4e2a-9e47-05078bdd6eb6

2. We should probably update the app menu to show the saved shortcut also (which will require restart bc electron pulls this in at app init)

image

DOsinga:
@zanesq 

1) it resets it back to the original if you toggle; yes that is by design.  my thinking is toggle it off is deleting the short-cut, toggling it on recreates it. it's also an easy way to reset.

2) I think we do this update. it does say in the UI for these short cuts that you need to restart the app

the other bug with the weird characters if you hit cmd-option-N should now also be fixed

PR #6716, opened 2026-01-26
fix __dirname is not defined

zanesq:
Follow for https://github.com/block/goose/pull/6711 to fix

```
08:45:24.213 › from renderer: [ERROR] ReferenceError: __dirname is not defined

    at Lazy (<anonymous>)
    at ErrorBoundary (http://localhost:5173/src/components/ErrorBoundary.tsx:89:5)
    at ConfigProvider (http://localhost:5173/src/components/ConfigContext.tsx:24:34)
    at Suspense (<anonymous>)

```


DOsinga:
thanks for diving on to this. I guess this was the original reason why getSettings returned an unknown then :) 

I am not sure this is the way. shouldn't we make it so that everything is importable, except for main.ts? or maybe we should have an electron folder? I just think we need to keep the electron stuff nice and separated.

Issue #6703, opened 2026-01-25
Shell tool output duplicated to LLM: audience filtering not applied to tool results


GordianDziwis:
**Describe the bug**

Shell tool output is duplicated in the function result returned to the agent. The user sees normal single output in their terminal, but the agent receives every line twice.

This wastes context window tokens and causes confusion - in my session, the agent incorrectly reported that config files had duplicated content when they did not.

---

**To Reproduce**
1. Start a goose CLI session
2. Have the agent run any shell command, e.g. `uname -a`
3. The agent receives the output twice in the tool result

---

**Expected behavior**

Agent receives the output once.

**Actual behavior**

Agent receives the output twice. Looking at the session JSON, the `toolResult.value.content` array contains two identical text entries:

```json
"content": [
  {
    "type": "text",
    "text": "extensions:\n  developer:\n  ...",
    "annotations": {
      "audience": ["assistant"]
    }
  },
  {
    "type": "text",
    "text": "extensions:\n  developer:\n  ...",   // <-- IDENTICAL TEXT
    "annotations": {
      "audience": ["user"],
      "priority": 0.0
    }
  }
]
```

The code in `crates/goose-mcp/src/developer/mod.rs` intentionally returns two Content items with different audiences, but for shell output they contain the **same text**:

```rust
Ok(vec![
    Content::text(output_str.clone()).with_audience(vec![Role::Assistant]),
    Content::text(output_str)
        .with_audience(vec![Role::User])
        .with_priority(0.0),
])
```

The bug is that both audience-tagged content items are being sent to the LLM context, when only the `assistant` audience should be.

---

**Screenshots**

N/A - the duplication is not visible to the user, only in what the agent receives.

---

**Please provide the following information**
- **OS & Arch:** Arch Linux x86_64 (kernel 6.18.5-arch1-1)
- **Interface:** CLI
- **Version:** 1.21.1
- **Extensions enabled:** developer, extensionmanager, todo
- **Provider & Model:** Anthropic – claude-opus-4-5-20251101

---

**Additional context**

The issue is in how tool results are filtered before being sent to the LLM. The `audience` annotation should filter content so only `assistant`-targeted content goes to the model, but currently both are included.

See `crates/goose-mcp/src/developer/mod.rs` for the shell tool implementation that returns dual-audience content.

knotseaborg:
I am facing the same issue.

rabi:
[...]

PR #6728, opened 2026-01-27
refactor: centralize audience filtering before providers receive messages

rabi:
## Summary
Move filtering to stream_response_from_provider() so providers receive pre-filtered messages.

### Type of Change
- [ ] Feature
- [] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Tested with anthropic provider.

```
Before the fix:

"content": "Linux ramishra-thinkpadp1gen7.rmtin.csb 6.18.5-200.fc43.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Jan 11 17:09:32 UTC 2026 x86_64 GNU/Linux\n\nLinux ramishra-thinkpadp1gen7.rmtin.csb 6.18.5-200.fc43.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Jan 11 17:09:32 UTC 2026 x86_64 GNU/Linux\n"
The uname -a output appeared twice (separated by a blank line).

After the fix:
"content": "Linux ramishra-thinkpadp1gen7.rmtin.csb 6.18.5-200.fc43.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Jan 11 17:09:32 UTC 2026 x86_64 GNU/Linux\n"
The output appears only once.

```

Closes: #6703

### DIFF
--- a/crates/goose/src/agents/reply_parts.rs
+++ b/crates/goose/src/agents/reply_parts.rs
@@ -182,11 +182,17 @@ impl Agent {
     ) -> Result<MessageStream, ProviderError> {
         let config = provider.get_model_config();
 
+        let filtered_messages: Vec<Message> = messages
+            .iter()
+            .filter(|m| m.is_agent_visible())
+            .map(|m| m.agent_visible_content())
+            .collect();
+
         // Convert tool messages to text if toolshim is enabled
         let messages_for_provider = if config.toolshim {
-            convert_tool_messages_to_text(messages)
+            convert_tool_messages_to_text(&filtered_messages)
         } else {
-            Conversation::new_unvalidated(messages.to_vec())
+            Conversation::new_unvalidated(filtered_messages)
         };
 
         // Clone owned data to move into the async stream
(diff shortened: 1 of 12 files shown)


DOsinga:
so https://github.com/block/goose/pull/6646/files#diff-0ceb972208a9cbe6f665656b11f613ea51588bedeeace7bfeae766302d8ea332 introduces filter_for_audience, can we use that here?

rabi:
> so https://github.com/block/goose/pull/6646/files#diff-0ceb972208a9cbe6f665656b11f613ea51588bedeeace7bfeae766302d8ea332 introduces filter_for_audience, can we use that here?

That's a good idea. Moving filtering earlier in the pipeline to the agent level using agent_visible_content() would be cleaner and easier to maintain, but it involves cloning entire Message structures upfront rather than filtering during extraction. Happy to refactor if the tradeoff is acceptable.

DOsinga:
yeah, but you could just call that on the filter_map(|c| { ? that should be a content object that has the filter, no?

rabi:
> yeah, but you could just call that on the filter_map(|c| { ? that should be a content object that has the filter, no?

Isn't `filter_for_audience` a method on `MessageContent` (goose type) and not on `Content` (rmcp type in tool results)? At present, inline filtering operates on `Content` items directly. I think what you're suggesting would involve adding `filter_for_audience` method to Content in rmcp crate or extension trait pattern to add methods to external types.

DOsinga (approved):

PR #6737, opened 2026-01-27
fix: read_resource_tool deadlock causing test_compaction to hang

lifeizhou-ap:
## Summary
**Why:**
test_compaction smoke test hung infinitely, cancelled after **6 hours** by GitHub Actions. This happened frequently now.  https://github.com/block/goose/actions/workflows/pr-smoke-test.yml?query=is%3Acancelled

**Root Cause:**
The compact smoke test does not have developer extension enabled setup. Without developer extension, the LLM sometimes tried to use extensionmanager__read_resource to read a file (eg: read hello.txt). This caused a deadlock:

> dispatch_tool_call("extensionmanager__read_resource")
>   └── Acquires LOCK on extensionmanager
>         └── read_resource_tool() iterates ALL extensions
>               └── Tries extensionmanager → needs same LOCK → 💀 DEADLOCK

The bug was intermittent due to HashMap's random iteration order - sometimes apps was checked first (works), sometimes extensionmanager was first (deadlock).

**How to reproduce the problem**
> goose run --text "use the read_resource tool to read the resource with uri ui://apps/clock"

This does not guarantee the command always hangs (due to the order of extensions HashMap), but much more frequent than the compact smoke test. 

**Fix:**
Only iterate extensions that actually support resources (matching existing list_resources() pattern)

**Note**
In current `test_compaction.sh`
> (cd "$TESTDIR" && "$GOOSE_BIN" run --text "list files and read hello.txt" 2>&1) | tee "$OUTPUT"

We could change it later with something simpler without developer tool (eg: describe Melbourne AU in 100 words) as the test is mainly for compaction feature.  However, I still leave the prompt as it is for now so that we can observe whether the problem is fixed.

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Manual testing with the command that reproduces the problem


crates/goose/src/prompts/system.md
```diff
@@ -31,8 +31,8 @@ in your tool specification.
 ## {{extension.name}}
 
 {% if extension.has_resources %}
-{{extension.name}} supports resources, you can use platform__read_resource,
-and platform__list_resources on this extension.
+{{extension.name}} supports resources, you can use extensionmanager__read_resource,
+and extensionmanager__list_resources on this extension.
```
lifeizhou-ap, line 35: At the moment, even with `platform__read_resource` in the system prompt , LLM is smart to call `extensionmanager__read_resource` tool.  However, it is better to update with proper extension name.

DOsinga (approved):
thanks! nice work

Issue #6577, opened 2026-01-19
AWS Bedrock - add support for API keys


andreswebs:
## **Please explain the motivation behind the feature request:**

Amazon Bedrock [announced](https://aws.amazon.com/about-aws/whats-new/2025/07/amazon-bedrock-api-keys-for-streamlined-development/) in July 2025 support for API Keys as auth method as an alternative to IAM credentials.

This will simplify onboarding and configuration when using the provider.

## **Describe the solution you'd like:**

By default the AWS SDKs look for an environment variable named `AWS_BEARER_TOKEN_BEDROCK`, which can also be passed as a bearer token in a direct HTTP request (`Authorization: Bearer $AWS_BEARER_TOKEN_BEDROCK`).

We'd like to have support for this same env var in the Goose AWS Bedrock provider, in addition to the already supported AWS credentials chain.

- [x] I have verified this does not duplicate an existing feature request

andreswebs:
This has been completed

hkolasani:
I have tried to configure Bedrock as a provider and used the AWS_BEARER_TOKEN_BEDROCK and not able to get that working. 
I keep getting this error. The API token is valid and it's working fine outside of Goose

_reply_stream{session.id=20260327_1}:stream_response_from_provider{session.id=20260327_1}: goose::providers::retry: crates/goose/src/providers/retry.rs: Request failed, retrying (1/6): ServerError("Failed to call Bedrock: Unhandled(Unhandled { source: ErrorMetadata { code: Some(\"UnrecognizedClientException\"), message: Some(\"The security token included in the request is invalid.\"), extras: Some({\"aws_request_id\": \"XXXXXXXX\"}) }, meta: ErrorMetadata { code: Some(\"UnrecognizedClientException\"), message: Some(\"The security token included in the request is invalid.\"), extras: Some({\"aws_request_id\": \"XXXXXXX"}) } })")_

hkolasani:
> This has been completed

@andreswebs  is this working? 

I have tried to configure Bedrock as a provider and used the AWS_BEARER_TOKEN_BEDROCK and not able to get that working.
I keep getting this error. The API token is valid and it's working fine outside of Goose

[...]

PR #6739, opened 2026-01-27
feat(goose): add support for AWS_BEARER_TOKEN_BEDROCK environment variable

andreswebs:
## Summary

Add support for bearer token authentication in the AWS Bedrock provider via the `AWS_BEARER_TOKEN_BEDROCK` environment variable.

This change allows users to authenticate with Amazon Bedrock using a bearer token as an alternative to standard AWS credentials (IAM/SSO). When `AWS_BEARER_TOKEN_BEDROCK` is set, the provider uses bearer token authentication; otherwise, it falls back to the existing credential-based authentication flow.

Changes:

- Add `AWS_BEARER_TOKEN_BEDROCK` as a new optional secret config key
- Refactor client creation into create_client_with_credentials helper method
- Make `AWS_PROFILE` optional (since bearer token auth doesn't require it)
- Make `AWS_REGION` required at runtime when `AWS_BEARER_TOKEN_BEDROCK` is used
- Update provider metadata description to document the new authentication option

### Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing

- Added unit tests for metadata configuration:
  - Config keys count validation
  - Bearer token config key presence
  - Bearer token marked as secret and optional
  - `AWS_PROFILE` marked as optional
  - Provider description mentions bearer token auth
  - Added integration test `test_bedrock_provider_bearer_token` that validates authentication flow with bearer token (requires - `AWS_BEARER_TOKEN_BEDROCK` and `AWS_REGION` env vars)

### Related Issues

Relates to <https://github.com/block/goose/issues/6577>

### DIFF
--- a/crates/goose/src/providers/bedrock.rs
+++ b/crates/goose/src/providers/bedrock.rs
@@ -64,8 +64,34 @@ impl BedrockProvider {
             }
         };
 
+        let filtered_secrets = config.all_secrets().map(|map| {
+            map.into_iter()
+                .filter(|(key, _)| key != "AWS_BEARER_TOKEN_BEDROCK")
+                .collect()
+        });
+
         set_aws_env_vars(config.all_values());
-        set_aws_env_vars(config.all_secrets());
+        set_aws_env_vars(filtered_secrets);
+
+        // Check for bearer token first to determine if region is required
+        let bearer_token = match config.get_secret::<String>("AWS_BEARER_TOKEN_BEDROCK") {
+            Ok(token) => {
+                let token = token.trim().to_string();
+                if token.is_empty() {
+                    None
+                } else {
+                    Some(token)
+                }
+            }
+            Err(_) => None,
+        };
+
+        // Get AWS_REGION from config if explicitly set (optional - SDK can resolve from other sources)
+        let region = match config.get_param::<String>("AWS_REGION") {
+            Ok(r) if !r.is_empty() => Some(r),
+            Ok(_) => None,
+            Err(_) => None,
+        };
 
         // Use load_defaults() which supports AWS SSO, profiles, and environment variables
         let mut loader = aws_config::defaults(aws_config::BehaviorVersion::latest());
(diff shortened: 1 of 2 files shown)


crates/goose/src/providers/bedrock.rs
```diff
@@ -67,6 +66,32 @@ impl BedrockProvider {
...
+            Ok(token) => {
+                let token = token.trim().to_string();
+                if token.is_empty() {
+                    None
+                } else {
+                    Some(token)
+                }
+            }
+            Err(ConfigError::NotFound(_)) => None,
+            Err(e) => {
```
michaelneale: won't this fail for legitimate cases where you aren't using a bearer token?

michaelneale:
thanks @andreswebs - just wanted to confirm as a little hard to look at just diffs, but this will be fine for non bearer token cases as before - and the new stuff only kick in when there is a bearer token variable configured? is that right?

crates/goose/src/providers/bedrock.rs
andreswebs: Yes - my original intention was to fail on all ConfigErrors other than the NotFound - so things like keychain access. But that's not good, indeed. If the user denies keychain access, for example, this wouldn't fallback to the other env vars (AWS_PROFILE).

So I'll update it to set the token to None for all errors, and let it fallback to the normal credential chain in those cases.

michaelneale (approved):
Looks good - the one copilot comment on potentially filtering out `AWS_BEARER_TOKEN_BEDROCK` is a nice to have I think @andreswebs if you wanted to include that and test it, but otherwise LGTM

andreswebs:
Awesome, thanks @michaelneale 
I'll include that and re-test in a bit

michaelneale:
thanks for seeing this through @andreswebs and appreciate all the permutation testing and patience!

crates/goose/tests/providers.rs
```diff
@@ -395,6 +395,15 @@ async fn test_provider(
...
         load_env();
 
+        // Check required_vars BEFORE applying env_modifications to avoid
+        // leaving the environment mutated when skipping
+        let missing_vars = required_vars.iter().any(|var| std::env::var(var).is_err());
+        if missing_vars {
+            println!("Skipping {} tests - credentials not configured", name);
+            TEST_REPORT.record_skip(name);
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
andreswebs: I'll leave this alone for now, as it impacts other tests.

Issue #6732, opened 2026-01-27
Got "Unsaved Changes" dialog asked when there's no changes to extension configuration (Desktop v1.12.1)


KiKaraage:
**Describe the bug**

Goose shown "Unsaved Changes" dialog when user didn't changed any settings for an extension. Unnecessary friction when user just want to check their extension config

Related components: ExtensionModal > `hasFormChanges`
https://github.com/block/goose/blob/e9024b7c190a66fcc14d5531c07ac1c66ae223f9/ui/desktop/src/components/settings/extensions/modal/ExtensionModal.tsx#L47-L87

---

**To Reproduce**
1. Open Goose. Click "Extension" menu in sidebar.
2. Click settings/gear icon for non-built-in extensions, let the modal open
3. Do nothing
4. Close it right away, using the X button on top or Cancel button at the bottom
5. See the "Unsaved Changes" dialog shown

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/c61d3d09-0f7c-4772-817b-554b076c4c8a" />

---

**Expected behavior**
Goose should detect config changes properly and not ask/show the "Unsaved Changes" dialog when user try to close the popup after doing nothing/if there's no extension config edited yet.
---

**Screen Recording**
https://github.com/user-attachments/assets/64e659ee-165b-48d0-b5dc-190b7dd10339

---

**Please provide the following information**
- **OS & Arch:** Linux - Bluefin based on Fedora 43, 64-bit
- **Interface:** UI
- **Version:** v.1.12.1 Desktop
- **Extensions added:** Linux MCP Server and Tavily. Default extensions enabled.

PR #6743, opened 2026-01-27
fix unsaved changes showing when no changes in extensions

zanesq:
## Summary

**Issue:** The "Unsaved Changes" dialog was appearing when closing the extension configuration modal even when no changes were made. This happened for extensions that had existing environment variables.

**Root Cause:** In `ExtensionModal.tsx`, the `hasFormChanges()` function had a check called `envVarsHaveText` that would return `true` if any environment variable had text in it, regardless of whether it was modified or just loaded from the initial data:

```javascript
// This was the problematic check
const envVarsHaveText = formData.envVars.some(
  (envVar) =>
    (envVar.key.trim() !== '' || envVar.value.trim() !== '') &&
    envVar.value !== '••••••••'
);
```

This check was meant to catch cases where users typed something but didn't explicitly mark it as edited. However, it caused false positives for existing extensions with environment variables because those env vars already have text in them.

**Fix:** Removed the `envVarsHaveText` check. The other existing checks already cover all necessary cases:
- `envVarsChanged` - detects when existing env vars are edited (`isEdited === true`)
- `envVarsAdded` - detects when new env vars are added (length comparison)
- `envVarsRemoved` - detects when env vars are removed (length comparison)
- `hasPendingInput` - detects when user is typing in the "new env var" input fields

**Files Changed:**
1. `ui/desktop/src/components/settings/extensions/modal/ExtensionModal.tsx` - Removed the redundant `envVarsHaveText` check
2. `ui/desktop/src/components/settings/extensions/modal/ExtensionModal.test.tsx` - Added a test to verify the fix

closes https://github.com/block/goose/issues/6732

### DIFF
--- a/ui/desktop/src/components/settings/extensions/modal/ExtensionModal.test.tsx
+++ b/ui/desktop/src/components/settings/extensions/modal/ExtensionModal.test.tsx
@@ -44,6 +44,123 @@ describe('ExtensionModal', () => {
     expect(screen.queryByText('Unsaved Changes')).not.toBeInTheDocument();
   });
 
+  it('shows unsaved changes dialog when name is modified', async () => {
+    const user = userEvent.setup();
+    const mockOnSubmit = vi.fn();
+    const mockOnClose = vi.fn();
+
+    const initialData: ExtensionFormData = {
+      name: 'Original Name',
+      description: 'An existing extension',
+      type: 'stdio',
+      cmd: 'npx some-mcp-server',
+      endpoint: '',
+      enabled: true,
+      timeout: 300,
+      envVars: [],
+      headers: [],
+    };
+
+    render(
+      <ExtensionModal
+        title="Edit Extension"
+        initialData={initialData}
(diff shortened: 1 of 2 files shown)


Abhijay007:
Hi @zanesq, I tested this, but I wanted to check, Is this change only targeting envVars/request header changes, or does it also cover overall extension changes?

When I change the extension name, description, etc., I don’t see any dialog showing up like before cancelling. It wasn’t showing previously either, but I just wanted to confirm whether we wanted to include those cases as well.

zanesq:
@Abhijay007 good catch I will add name and description and also I noticed port is not triggering it.

Abhijay007:
> @Abhijay007 good catch I will add name and description and also I noticed port is not triggering it.

I tried editing port it was triggering it for me 🤔 I tried changing it's value in jetbrain extension

zanesq:
@Abhijay007 my bad meant `timeout` field, updated

PR #6745, opened 2026-01-27
refactor providers a bit to add lower level request logging

jamadeo:
Refactors the way providers build and execute requests so we can use the request logger at a lower level, and miss nothing.


DOsinga:
still working on this?

jamadeo:
I'm going to make a new version of this.

PR #6749, opened 2026-01-27
Use errorMessage

DOsinga:
## Summary

We have a handy errorMessage uitlity that we weren't using all over


documentation/src/pages/recipes/index.tsx
```diff
@@ -8,6 +8,7 @@ import { Button } from "@site/src/components/ui/button";
 import { SidebarFilter, type SidebarFilterGroup } from "@site/src/components/ui/sidebar-filter";
 import { Menu, X } from "lucide-react";
 import Link from '@docusaurus/Link';
+import { errorMessage } from '../../../ui/desktop/src/utils/conversionUtils';
```
zanesq: Are we sure we want to import from desktop what if we want to split out the documentation from this repo one day?

angiejones: good point

PR #6773, opened 2026-01-28
Fix/filter audience 6703 local

DOsinga:
## Summary

Follow up on: https://github.com/block/goose/pull/6728#issuecomment-3805912481

### DIFF
--- a/crates/goose/src/providers/formats/databricks.rs
+++ b/crates/goose/src/providers/formats/databricks.rs
@@ -45,11 +45,7 @@ fn format_tool_response(
 
     match &response.tool_result {
         Ok(call_result) => {
-            let abridged: Vec<_> = call_result
-                .content
-                .iter()
-                .map(|c| c.raw.clone())
-                .collect();
+            let abridged: Vec<_> = call_result.content.iter().map(|c| c.raw.clone()).collect();
 
             let mut tool_content = Vec::new();
             let mut image_messages = Vec::new();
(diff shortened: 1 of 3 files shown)


crates/goose/src/conversation/message.rs
```diff
@@ -285,10 +285,8 @@ impl MessageContent {
                     .cloned()
                     .collect();
 
-                if filtered_content.is_empty() {
-                    return None;
-                }
-
+                // Preserve ToolResponse even when content is empty - some providers
+                // (like Google) need to handle empty tool responses specially
```
jamadeo, line 289: without the deleted code, this comment seems a little out of place, you could just drop it

crates/goose/src/providers/formats/anthropic.rs
```diff
@@ -34,7 +34,13 @@ const DATA_FIELD: &str = "data";
 pub fn format_messages(messages: &[Message]) -> Vec<Value> {
     let mut anthropic_messages = Vec::new();
 
-    for message in messages.iter().filter(|m| m.is_agent_visible()) {
+    let filtered_messages: Vec<Message> = messages
+        .iter()
+        .filter(|m| m.is_agent_visible())
+        .map(|m| m.agent_visible_content())
```
jamadeo, line 40: maybe agent_visible_content should return an `Option<Message>` (None if is_agent_visible would return false) and you could .filter_map these

DOsinga:
> It seems like we should be doing this before the provider ever gets the messages. Like the .complete() methods on Provider should take a Conversation type that we produce with Conversation::for_agent or something like that

I agree. this is an adaption of a PR that came in though, so don't want to change too much

katzdave (approved):

rabi:
> It seems like we should be doing this before the provider ever gets the messages. Like the .complete() methods on Provider should take a Conversation type that we produce with Conversation::for_agent or something like that

Thanks @DOsinga @jamadeo for this. I've updated https://github.com/block/goose/pull/6728 to move the filtering to `stream_response_from_provider()` in `reply_parts.rs`, so providers receive pre-filtered messages. Please check if that makes sense.

PR #6776, opened 2026-01-28
fix: address compilation issue in main

alexhancock:
`cargo build` is broken at the moment. this addresses it by using `StreamableHttpOptions` where it should be used, and also using the new `timeout` attr available


The-Best-Codes:
Just tested and code compiles fine for me (with `cargo check`, `cargo clippy`, and `cargo build` on Linux)

Abhijay007 (approved):
build sucess LGTM

PR #6777, opened 2026-01-28
Fix: Recipe Extensions Not Loading in Desktop

zanesq:
### Problem
Recipes with an empty `extensions: []` array would cause `resolve_extensions_for_new_session` 
to return no extensions, instead of falling back to the user's default enabled extensions.

### Solution
Strip empty extension arrays in the frontend before saving/running recipes.

### Files Changed

**ui/desktop/src/recipe/index.ts**
- Added `stripEmptyExtensions()` utility function that removes the `extensions` property 
  if it's an empty array

**ui/desktop/src/components/recipes/RecipesView.tsx**
- `handleStartRecipeChat()`: Strip empty extensions in memory before running a recipe
- Removed unused `recipeId` parameter

**ui/desktop/src/components/recipes/CreateEditRecipeModal.tsx**
- `getCurrentRecipe()`: Use `stripEmptyExtensions()` when building recipe to save

**ui/desktop/src/components/recipes/CreateRecipeFromSessionModal.tsx**
- `handleCreateRecipe()`: Simply don't set `extensions` at all (instead of setting `[]`)

**ui/desktop/src/components/schedule/ScheduleModal.tsx**
- `recipeToYaml()`: Strip empty extensions before converting recipe to YAML for scheduling

### How it works
- If a recipe has `extensions: []`, it gets stripped to `extensions: undefined`
- Backend's `resolve_extensions_for_new_session(None, override_extensions)` then falls 
  through to use the UI's enabled extensions (desktop) or `get_enabled_extensions()` (scheduler)

### DIFF
--- a/ui/desktop/src/components/recipes/CreateEditRecipeModal.tsx
+++ b/ui/desktop/src/components/recipes/CreateEditRecipeModal.tsx
@@ -1,6 +1,6 @@
 import React, { useState, useEffect, useCallback } from 'react';
 import { useForm } from '@tanstack/react-form';
-import { Recipe, generateDeepLink, Parameter, stripEmptyExtensions } from '../../recipe';
+import { Recipe, generateDeepLink, Parameter } from '../../recipe';
 import { Check, ExternalLink, Play, Save, X } from 'lucide-react';
 import { Geese } from '../icons/Geese';
 import Copy from '../icons/Copy';
@@ -81,12 +81,8 @@ export default function CreateEditRecipeModal({
   const [copied, setCopied] = useState(false);
   const [isSaving, setIsSaving] = useState(false);
 
-  // Initialize selected extensions for the recipe
-  const [recipeExtensions] = useState<ExtensionConfig[]>(() => {
-    if (recipe?.extensions) {
-      return recipe.extensions;
-    }
-    return [];
+  const [recipeExtensions] = useState<ExtensionConfig[] | undefined>(() => {
+    return recipe?.extensions ?? undefined;
   });
 
   // Reset form when recipe changes
@@ -132,14 +128,11 @@ export default function CreateEditRecipeModal({
       }
     }
 
-    // Strip envs to avoid leaking secrets
-    const extensionsWithoutEnvs = recipeExtensions.map((extension) =>
+    const extensions = recipeExtensions?.map((extension) =>
       'envs' in extension ? { ...extension, envs: undefined } : extension
-    ) as ExtensionConfig[];
+    ) as ExtensionConfig[] | undefined;
 
-    // Use stripEmptyExtensions to avoid saving empty extensions array
-    // Empty extensions array would prevent default extensions from loading
-    return stripEmptyExtensions({
+    return {
       ...recipe,
       title,
       description,
@@ -148,8 +141,8 @@ export default function CreateEditRecipeModal({
       prompt: prompt || undefined,
       parameters: formattedParameters,
       response: responseConfig,
-      extensions: extensionsWithoutEnvs,
-    }) as Recipe;
+      extensions,
+    };
   }, [
     recipe,
     title,
--- a/ui/desktop/src/components/recipes/CreateRecipeFromSessionModal.tsx
+++ b/ui/desktop/src/components/recipes/CreateRecipeFromSessionModal.tsx
@@ -156,8 +156,6 @@ export default function CreateRecipeFromSessionModal({
 
     setIsCreating(true);
     try {
-      // Create the recipe object from form data
-      // Don't set extensions - let the recipe use default enabled extensions
       const recipe: Recipe = {
         title: formData.title,
         description: formData.description,
--- a/ui/desktop/src/components/schedule/ScheduleModal.tsx
+++ b/ui/desktop/src/components/schedule/ScheduleModal.tsx
@@ -4,7 +4,7 @@ import { Button } from '../ui/button';
 import { Input } from '../ui/input';
 import { ScheduledJob } from '../../schedule';
 import { CronPicker } from './CronPicker';
-import { Recipe, decodeRecipe, stripEmptyExtensions } from '../../recipe';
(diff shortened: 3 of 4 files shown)


DOsinga:
I still don't understand the issue. this seems to be like the correct behavior. if the recipe specifies extensions, we should use those. if the recipe specifies an empty list, we should run with no extensions. if a recipe does not specify extensions, then we should use the one that the system has. /cc @lifeizhou-ap

angiejones:
>  if a recipe does not specify extensions, then we should use the one that the system has. /cc @lifeizhou-ap

but it's not using the ones the system has. it loads as 0 extensions enabled and then is unable to run

zanesq:
I discussed it with @DOsinga gonna take a different approach and handle it in the desktop code. The main issue is desktop is writing an empty extensions array rather than no extensions.

ui/desktop/src/recipe/index.ts
```diff
@@ -76,3 +76,16 @@ export async function generateDeepLink(recipe: Recipe): Promise<string> {
   const encoded = await encodeRecipe(recipe);
   return `goose://recipe?config=${encoded}`;
 }
+
+/**
+ * Strips empty extensions array from a recipe.
+ * This ensures recipes without explicit extensions will use the default enabled extensions
+ * from the user's config in the desktop rather than loading with no extensions.
+ */
+export function stripEmptyExtensions<T extends { extensions?: unknown[] | null }>(recipe: T): T {
```
DOsinga: why this complicated type? shouldn't this just be Recipe in, Recipe out? that would save us some casting above too

lifeizhou-ap:
> I discussed it with @DOsinga gonna take a different approach and handle it in the desktop code. The main issue is desktop is writing an empty extensions array rather than no extensions.

Yes! I noticed this late yesterday when testing the sub recipe feature.  When user creates recipe via UI it includes "extensions: []" in the file.   Previously it does not matter in Desktop because Desktop ignores the extensions in the recipe.  Since we unified the recipe extension behaviour in Desktop, CLI and Scheduler recently, Desktop user has different experiences now.  I was about fixing this today.  Thanks @zanesq for fixing this

ui/desktop/src/components/schedule/ScheduleModal.tsx
```diff
@@ -80,25 +80,28 @@ async function parseDeepLink(deepLink: string): Promise<Recipe | null> {
 }
 
 function recipeToYaml(recipe: Recipe): string {
+  // Strip empty extensions to ensure scheduled recipes use default enabled extensions
```
DOsinga: hmm, we discussed this, that's not how the scheduler works

ui/desktop/src/components/schedule/ScheduleModal.tsx
```diff
@@ -80,25 +80,28 @@ async function parseDeepLink(deepLink: string): Promise<Recipe | null> {
...
 
-  if (recipe.activities && recipe.activities.length > 0) {
-    cleanRecipe.activities = recipe.activities;
+  if (processedRecipe.activities && processedRecipe.activities.length > 0) {
+    cleanRecipe.activities = processedRecipe.activities;
   }
 
-  if (recipe.extensions && recipe.extensions.length > 0) {
-    cleanRecipe.extensions = recipe.extensions.map((ext) => {
+  if (processedRecipe.extensions && processedRecipe.extensions.length > 0) {
```
DOsinga: and we're stripping it anyway here, so this feels like a no op

ui/desktop/src/components/recipes/CreateRecipeFromSessionModal.tsx
```diff
@@ -157,6 +157,7 @@ export default function CreateRecipeFromSessionModal({
     setIsCreating(true);
     try {
       // Create the recipe object from form data
+      // Don't set extensions - let the recipe use default enabled extensions
```
DOsinga: typical LLM comment

ui/desktop/src/components/recipes/CreateEditRecipeModal.tsx
```diff
@@ -132,7 +132,14 @@ export default function CreateEditRecipeModal({
...
     }
 
-    return {
+    // Strip envs to avoid leaking secrets
+    const extensionsWithoutEnvs = recipeExtensions.map((extension) =>
+      'envs' in extension ? { ...extension, envs: undefined } : extension
+    ) as ExtensionConfig[];
+
+    // Use stripEmptyExtensions to avoid saving empty extensions array
+    // Empty extensions array would prevent default extensions from loading
```
DOsinga: do we? we do want to make the distinction between empty lists and no recipes. if you now load a recipe that pre-existed and had an empty list of extensions, and you save it, it now has no extensions.

you want to make sure that if we create a recipe the default extensions are undefined and then respect whatever it said already

ui/desktop/src/components/schedule/ScheduleModal.tsx
lifeizhou-ap: Yes, scheduler always respect the extension config in recipe, not like Desktop. Maybe we should not apply this stripEmptyExtensions here.

ui/desktop/src/components/recipes/CreateEditRecipeModal.tsx
```diff
@@ -81,12 +81,8 @@ export default function CreateEditRecipeModal({
   const [copied, setCopied] = useState(false);
   const [isSaving, setIsSaving] = useState(false);
 
-  // Initialize selected extensions for the recipe
-  const [recipeExtensions] = useState<ExtensionConfig[]>(() => {
-    if (recipe?.extensions) {
-      return recipe.extensions;
-    }
-    return [];
+  const [recipeExtensions] = useState<ExtensionConfig[] | undefined>(() => {
```
lifeizhou-ap, line 84: That's great! I was about suggesting this too! You are quick :)

Issue #6456, opened 2026-01-12
Goose crashes on Mac Os


u35tpus:
**Describe the bug**

💡 Before filing, please check common issues:  
https://block.github.io/goose/docs/troubleshooting  

📦 To help us debug faster, attach your **diagnostics zip** if possible.  
👉 How to capture it: https://block.github.io/goose/docs/troubleshooting/diagnostics-and-reporting/

Goose crashes several times a day on my Macbook.
I have attached crash report.

---

**To Reproduce**
I can not tell how to reproduce it exactly. Usually it happens in long sessions after several compactions of context.
---

**Expected behavior**
I expect that crashes will not occur at all or not that often at least.

---

**Screenshots**

---

**Please provide the following information**
everything is in the attached report, but i can repeat here for convenience.

Identifier:          com.electron.goose
Version:             1.18.0 (1.18.0)
Code Type:           ARM-64 (Native)
Role:                Foreground
Parent Process:      launchd [1]
Coalition:           com.electron.goose [25931]
User ID:             501

Date/Time:           2026-01-12 23:59:20.5766 +0100
Launch Time:         2026-01-12 21:24:28.2781 +0100
Hardware Model:      Mac15,7
OS Version:          macOS 26.2 (25C56)

---

**Additional context**

[goose-report.13Jan.00.00.txt](https://github.com/user-attachments/files/24575530/goose-report.13Jan.00.00.txt)

zanesq:
I've seen this also.. hoping the [electron package updates](https://github.com/block/goose/pull/6450) in the next release fix it need to run main locally to see if it still happens

blackgirlbytes:
Context for anyone who is also experiencing this :

I think this an Electron bug.. It is a known Electron framework issue affecting macOS 26 (Tahoe).

I found several open issues with the exact same crash signature:
- [electron/electron#49281](https://github.com/electron/electron/issues/49281) - macOS 26 ARM64 crashes
- [electron/electron#49185](https://github.com/electron/electron/issues/49185) - TestFlight SIGTRAP

Your crash (`ares_dns_rr_get_ttl`, `EXC_BREAKPOINT/SIGTRAP`) matches these exactly. It is a memory corruption bug in the c-ares DNS library that Electron uses, and macOS 26 seems to have exposed it.


As Zane said above, hopefully electron will fix this in their next release

zanesq:
[...]

PR #6779, opened 2026-01-28
upgrade electron to 39.3.0

zanesq:
## Summary
upgrade electron to minor safe version 39.3.0

attempt to fix https://github.com/block/goose/issues/6456

if this doesn't work we can try to upgrade to 40.0.0 next (less safe major upgrade)

both releases have crash fixes so hopefully one of them does the trick


zanesq:
Just that its a major upgrade to `0.0` so a bit riskier, would like to try this first to be on the safe side. Both have stability updates.

PR #6793, opened 2026-01-29
fix: migrate tables if they are missing things

michaelneale:
This is a sensitive one - but I got into a state where somehow my db was missing a migration and I had to have goose manually add it. 

This is an attempt to automate that so the schema can be repaired. Not sure if this is a terrible idea but @zanesq @jamadeo @DOsinga if you think it is - worth consideration (not easy how to test this TBH)


DOsinga:
So I think this is fixed in : https://github.com/block/goose/pull/6809 - closing this, let me know if you don't agree

PR #6803, opened 2026-01-29
acp: load configured extensions and refactor tests

codefromthecrypt:
## Summary

This allows ACP to read and use the global extensions defined in goose config file. As these are global, they are loaded only once during initialization, not per session, the same way builtins are.

This also opportunistically refactors tests for the upcoming #6305 (providers), so that the diff there is less huge. Notably the tests now check the session_id more strictly and allow exact re-use between client (acp provider) and server (`goose acp`) modes.

### Type of Change
- [x] Bug fix
- [x] Refactor / Code quality

### AI Assistance

### Testing
New integration test that writes a config file for an MCP server and proves it was loaded as an extension by using it.


crates/goose-acp/tests/common_tests/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+pub async fn run_configured_extension<S: Session>() {
+    let temp_dir = tempfile::tempdir().unwrap();
+    let expected_session_id = ExpectedSessionId::default();
+    let prompt = "Use the get_code tool and output only its result.";
+    let mcp = McpFixture::new(expected_session_id.clone()).await;
+
+    let config_yaml = format!(
+        "extensions:\n  lookup:\n    enabled: true\n    type: streamable_http\n    name: lookup\n    description: Lookup server\n    uri: \"{}\"\n",
```
codefromthecrypt, line 193: this is the new test

zhixiongdu027:
I've also encountered the issue where the global configuration doesn't take effect in ACP mode. Please help move this forward as soon as possible. Thanks!

michaelneale:
merging as has hit me and others.

PR #6809, opened 2026-01-29
Session manager fixes

DOsinga:
## Summary

This fixes two issues:

* We weren't running migration inside of a transaction, which means that if you had a large number of messages and had goose running twice, we could have both upgrading, both failing and we're broken
* A true merge conflict about inserting messages for fork.


michaelneale:
nice catch - and I did see comments in the past about outside of tx so this explains why, never thought about the failure mode of multiple things hitting sqlite - amazing how well it holds up

PR #6811, opened 2026-01-29
Filter Apps page to only show standalone Goose Apps

aharvard:
## Summary

Filter the Apps page to only display apps from the "apps" extension (vibe coded apps built by Goose).

## Changes

- Client-side only change in `AppsView.tsx`
- Filters apps using `mcpServers?.includes('apps')` - the same pattern already used for the "Custom app" label
- No backend changes required

## Why

Previously, the Apps page displayed all MCP UI resources from all extensions, including chat-dependent MCP apps that require a chat context to function. This caused confusion since those apps cannot be launched standalone.

Now only apps created by Goose (vibe coded apps) are shown on the Apps page.

### DIFF
--- a/crates/goose-server/src/routes/agent.rs
+++ b/crates/goose-server/src/routes/agent.rs
@@ -978,7 +978,7 @@ async fn list_apps(
     let cache = McpAppCache::new().ok();
 
     let Some(session_id) = params.session_id else {
-        let apps: Vec<GooseApp> = cache
+        let apps = cache
             .as_ref()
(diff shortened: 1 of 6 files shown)


DOsinga:
as discussed, I think we should do the filtering on the client

aharvard:
@DOsinga, we're client-side only now. LMK if this is cool.

aharvard:
@DOsinga, friendly ping for when you have a moment

PR #6817, opened 2026-01-29
Remove apps from the client

DOsinga:
## Summary

Apps are currently broken


alexhancock:
If I `cargo run --bin goose` it adds apps. So I think we probably need to comment out the rust block which defines/adds the platform extension right?

https://github.com/block/goose/blob/01e607a12d0745cd871347a9054e3cf325489ab2/crates/goose/src/agents/extension.rs#L58-L67

DOsinga:
closing since we don't want this in main

PR #6824, opened 2026-01-30
Remove autogenerated .gooseignore files that don't belong in repo

DOsinga:
## Summary

What is says on the tin


michaelneale (approved):
goose added these automatically and goose taketh away haha

PR #6825, opened 2026-01-30
Fix case-insensitive matching for builtin extension names

DOsinga:
## Summary

normalize the name when doing a lookup.


michaelneale:
built locally and this works. Only follow on is why do we even use name for anything really programmatic other than display... but yes this fixes it.

michaelneale (approved):

PR #6832, opened 2026-01-30
refactor(providers): extract ProviderDef trait and OpenAiCompatibleProvider

codefromthecrypt:
## Summary
Separate provider construction from runtime behavior to reduce boilerplate and prepare for ACP-backed providers.

**`ProviderDef`**: new trait with `metadata()` and `from_env()`, split from `Provider`. Simplifies registration from a closure-based call to a single type parameter:

```diff
-        registry.register::<XaiProvider, _>(|m| Box::pin(XaiProvider::from_env(m)), false);
+        registry.register::<XaiProvider>(false);
```

The name follows Rust conventions (`FromStr`, `Default`) rather than OOP "factory" terminology. The implementing structs are zero-sized types used purely as type-level descriptors, with an associated type that ensures compile-time verification that the constructed provider implements `Provider`.

**`OpenAiCompatibleProvider`**: reusable struct that implements `Provider` for any OpenAI-compatible API. Providers like xAI and Azure go from ~200 lines (full `Provider` impl with duplicated HTTP/streaming logic) to ~70 lines (just metadata + client setup). Applied initially to "xai" and "azure" to prove the concept; others can follow.

This is vital for ACP-backed providers, as there could be over a dozen and differ only on aspects such as permission and mode mappings.

### Type of Change
- [x] Refactor / Code quality

### AI Assistance

### Testing

Ran the normal tests.

### Related Issues

Last major refactor needed to reduce the huge diff in ACP provider PR #6605

### DIFF
--- a/crates/goose/src/providers/azure.rs
+++ b/crates/goose/src/providers/azure.rs
@@ -82,16 +82,18 @@ impl ProviderDef for AzureProvider {
             })?;
 
             let auth_provider = AzureAuthProvider { auth };
-            let api_client = ApiClient::new(endpoint, AuthMethod::Custom(Box::new(auth_provider)))?
+            let host = format!(
+                "{}/openai/deployments/{}",
+                endpoint.trim_end_matches('/'),
+                deployment_name
+            );
+            let api_client = ApiClient::new(host, AuthMethod::Custom(Box::new(auth_provider)))?
                 .with_query(vec![("api-version".to_string(), api_version)]);
-
-            let path = format!("openai/deployments/{}", deployment_name);
 
             Ok(OpenAiCompatibleProvider::new(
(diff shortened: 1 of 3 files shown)


crates/goose-mcp/.gooseignore
```diff
@@ -0,0 +1,18 @@
+# This file is created automatically if no .gooseignore exists.
```
michaelneale: can remove this - new version of goose doesn't add these all over - was a bad bug!

codefromthecrypt: no my bad will kill that

michaelneale (approved):
I think this is ok - main concern was azure shift to streaming and missing methods for loading models @codefromthecrypt

Issue #6694, opened 2026-01-24
Stream output of Claude Code CLI provider


joshrnoll:
**Please explain the motivation behind the feature request.**
When using goose CLI with the Claude Code CLI provider, goose's output is not streamed. User gets the loading animation until goose is complete with the request. This makes it difficult to catch when goose is going in the wrong direction and the user wants to Ctrl+C and try a different prompt.

**Describe the solution you'd like**
Support streaming output via the Claude Code CLI provider. The anthropic API should support this.

**Describe alternatives you've considered**
The best alternative is to use claude code directly. But we all know that goose is cooler 🪿

**Additional context**
n/a

- [x] I have verified this does not duplicate an existing feature request

PR #6833, opened 2026-01-30
feat: add streaming support for Claude Code CLI provider

rabi:
## Summary
Implement stream() method using Claude CLI's --output-format stream-json and --include-partial-messages flags for incremental text output. Parse NDJSON events (content_block_delta, message_start, message_delta, result) to yield text chunks and usage information.

### Type of Change
- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Unit tests for parsing steaming events and tested locally.

Closes: #6694

### DIFF
--- a/crates/goose/src/providers/claude_code.rs
+++ b/crates/goose/src/providers/claude_code.rs
@@ -1,16 +1,21 @@
 use anyhow::Result;
+use async_stream::try_stream;
 use async_trait::async_trait;
 use futures::future::BoxFuture;
-use rmcp::model::Role;
+use rmcp::model::{Role, Tool};
 use serde_json::{json, Value};
 use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
+use std::sync::Arc;
 use tempfile::NamedTempFile;
 use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
 use tokio::process::Command;
 
-use super::base::{ConfigKey, Provider, ProviderDef, ProviderMetadata, ProviderUsage, Usage};
+use super::base::{
+    stream_from_single_message, ConfigKey, MessageStream, Provider, ProviderDef, ProviderMetadata,
+    ProviderUsage, Usage,
+};
 use super::errors::ProviderError;
 use super::utils::{filter_extensions_from_system_prompt, RequestLog};
 use crate::config::base::ClaudeCodeCommand;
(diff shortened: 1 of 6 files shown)


rabi:
@jamadeo Hey! it's probably the last one in the series of streaming support. PTAL when possible.

rabi:
Hmm. I've to now rewrite the streaming implementation fresh as we moved to persistent child process that stays alive across turns vs spawning a fresh claude CLI process per request in https://github.com/block/goose/pull/7029 that merged quickly.

crates/goose/src/providers/claude_code.rs
```diff
@@ -65,11 +108,10 @@ impl ClaudeCodeProvider {
             command: resolved_command,
             model,
             name: CLAUDE_CODE_PROVIDER_NAME.to_string(),
-            cli_process: tokio::sync::OnceCell::new(),
+            cli_process: Arc::new(tokio::sync::Mutex::new(None)),
```
codefromthecrypt: since we aren't respawning can you revert the process control logic back? Options are a smell unless proven otherwise, and this change as it sits shouldn't need it. let's pare down the logic to the streaming part?

rabi:
> what you can do instead is run the providers.rs test manually for the claude,

I don't see claude_code provider testing in @crates/goose/tests/providers.rs atm, unless you're asking me to add it. ProviderTester::run_test_suite() won't work as-is for claude_code as it does not support tools or /models api, so it's just about running test_basic_response(), which I have already tested locally.

codefromthecrypt:
right the test refactor was caged to another PR. I extracted this which should hopefully land before your next change https://github.com/block/goose/pull/7108

rabi:
> right the test refactor was caged to another PR. I extracted this which should hopefully land before your next change #7108

I'll rebase my PR once you're done with all your changes as I don't want to keep doing it again and again, let me know.

codefromthecrypt:
If you prefer I can just do this. The changes I have been doing was to level set and begin testing claude and codex as both were very broken and will be replaced by acp soon. Lemme know

codefromthecrypt:
@rabi cool you're on it! so I'm not doing anything else related to the provider, so your merge is safe this time ;)

I have something personal come up so I won't be here to support the merge, so hopefully someone else can help.

rabi:
> the one for gemini might have some overlapping code, probably not worth reusing?

Thanks @DOsinga, the new functions added in these two commits are similar in pattern but differ in return types and error handling due to different JSON schemas. Having said that, I agree that there are some existing repeated functions across all CLI providers in existing code (ex. generate_simple_session_description(), is_session_description_request()) that are copy pasted across providers. I'll create a follow-up PR to extract the truly shared ones into a common module.

rabi:
> > the one for gemini might have some overlapping code, probably not worth reusing?

Now that https://github.com/block/goose/pull/7118 has meged, I've refactored what could be moved to a common  module as suggested, but it has bloated this PR.

Issue #6835, opened 2026-01-30
HTTP Proxy Environment Variables Not Honored by Rust Backend


tltsaia:
# HTTP Proxy Environment Variables Not Honored by Rust Backend

## Summary

Documentation claims goose supports `HTTP_PROXY`/`HTTPS_PROXY` environment variables, but the Rust backend doesn't implement this feature, causing connection failures in corporate proxy environments.

## Environment

- goose v1.22.0, Windows, Desktop app
- Network: Corporate proxy at `10.1.250.30:8080`
- Provider: Azure OpenAI at `company.com`

## Problem

```powershell
SET HTTPS_PROXY=http://10.1.250.30:8080
SET AZURE_OPENAI_ENDPOINT=https://company.com
```

**Error:**
```
Request failed: error sending request for url 
(https://company.com/...) 
(failed to connect to company.com)
```

**Root Cause:** `reqwest::Client` doesn't read environment variables without `.use_system_proxy()`

- ✅ Electron UI configures proxy correctly (`ui/desktop/src/main.ts:80`)
- ❌ Rust backend ignores proxy settings (`crates/goose/src/providers/api_client.rs:210`)

## Solution

Add `.use_system_proxy()` to all `Client::builder()` calls:

**`crates/goose/src/providers/api_client.rs`:**
```rust
// Line 210
let mut client_builder = Client::builder()
    .timeout(timeout)
    .use_system_proxy();  // Add this

// Line 232  
let mut client_builder = Client::builder()
    .timeout(self.timeout)
    .default_headers(self.default_headers.clone())
    .use_system_proxy();  // Add this
```

**Also fix:**
- `crates/goose/src/providers/githubcopilot.rs:158`
- `crates/goose/src/providers/gcpvertexai.rs:161`
- `crates/goose/src/providers/toolshim.rs:73`

## Impact

**Severity:** High - Blocks enterprise users behind corporate proxies

**Affected:** Any deployment requiring HTTP proxy for LLM provider connections

**Workaround:** Use OS-level proxy tools (Proxifier, Netch) - requires admin rights

---

**Labels:** `bug` `networking` `enterprise` documentation

DOsinga:
hello & thanks for the report and for using goose. I would like to remark that AI created reports like this don't have all that much value; `use_system_proxy()` does not exist for example.

tltsaia:
[...]

PR #6840, opened 2026-01-30
Proper proxy support on windows & mac

DOsinga:
## Summary

Fixes: https://github.com/block/goose/issues/6835

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6951,9 +6951,9 @@ dependencies = [
 
 [[package]]
 name = "pctx_code_mode"
-version = "0.2.2"
+version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5788a3f1aaa7bc0fdcbae6d9ff24a080a6cc0d202d8affa8f1c29097e40a504a"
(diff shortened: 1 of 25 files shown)


DOsinga:
goose tells me we need these in the cargo files to make it work. it does work for me on mac this way and before did not.

jh-block:
Copilot seems to be wrong; system proxies are enabled by default so building with the feature should be enough: https://docs.rs/reqwest/latest/reqwest/#proxies

jh-block:
> Interesting to see the other default features that we're not enabling (`default-features = false`): https://docs.rs/crate/reqwest/0.12.28/features#default / https://docs.rs/reqwest/latest/reqwest/#optional-features
> 
> `http2` and `charset` seem like features we should probably also add?

Because of feature unification across the workspace, I think they do end up enabled, because the `crates/goose/Cargo.toml` has:

```
reqwest = { workspace = true, features = ["rustls-tls-native-roots", "json", "cookies", "gzip", "brotli", "deflate", "zstd", "charset", "http2", "stream", "blocking", "multipart"], default-features = false }
```

jamadeo:
aha, got it, thanks @jh-block

PR #6849, opened 2026-01-30
More providers for testing

jamadeo:
Also run them in parallel

### DIFF
--- a/scripts/test_providers.sh
+++ b/scripts/test_providers.sh
@@ -284,29 +284,50 @@ total_jobs=${#JOBS[@]}
 echo "Starting $total_jobs tests..."
 echo ""
 
-# Run tests in parallel
-running_jobs=0
-for job in "${JOBS[@]}"; do
-  IFS='|' read -r provider model idx <<< "$job"
+# Run first test sequentially if any jobs exist
+if [ $total_jobs -gt 0 ]; then
+  echo "Running first test sequentially..."
+  first_job="${JOBS[0]}"
+  IFS='|' read -r provider model idx <<< "$first_job"
 
   result_file="$RESULTS_DIR/result_$idx"
   output_file="$RESULTS_DIR/output_$idx"
   meta_file="$RESULTS_DIR/meta_$idx"
   echo "$provider|$model" > "$meta_file"
 
-  # Run test in background
-  run_test "$provider" "$model" "$result_file" "$output_file" &
-  ((running_jobs++))
+  # Run first test and wait for it to complete
+  run_test "$provider" "$model" "$result_file" "$output_file"
+  echo "First test completed."
+  echo ""
+fi
 
-  # Wait if we've hit the parallel limit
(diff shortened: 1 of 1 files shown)


jamadeo:
As of now, this is what's added but not run:

```
⚠️  Skipping Databricks tests (DATABRICKS_HOST and DATABRICKS_TOKEN required)
⚠️  Skipping Azure OpenAI tests (AZURE_OPENAI_ENDPOINT and AZURE_OPENAI_DEPLOYMENT_NAME required)
⚠️  Skipping AWS Bedrock tests (AWS_REGION and AWS_PROFILE or AWS credentials required)
⚠️  Skipping GCP Vertex AI tests (GCP_PROJECT_ID required)
⚠️  Skipping Snowflake tests (SNOWFLAKE_HOST and SNOWFLAKE_TOKEN required)
⚠️  Skipping Venice tests (VENICE_API_KEY required)
⚠️  Skipping LiteLLM tests (LITELLM_API_KEY required)
⚠️  Skipping Ollama tests (OLLAMA_HOST required or ollama must be installed)
⚠️  Skipping SageMaker TGI tests (SAGEMAKER_ENDPOINT_NAME and AWS_REGION required)
⚠️  Skipping GitHub Copilot tests (OAuth setup required - run 'goose configure' first)
⚠️  Skipping ChatGPT Codex tests (OAuth setup required - run 'goose configure' first)
⚠️  Skipping Claude Code CLI tests ('claude' CLI tool required)
⚠️  Skipping Codex CLI tests ('codex' CLI tool required)
⚠️  Skipping Gemini CLI tests ('gemini' CLI tool required)
⚠️  Skipping Cursor Agent tests ('cursor-agent' CLI tool required)
```

scripts/test_providers.sh
```diff
@@ -111,52 +224,147 @@ should_skip_provider() {
...
+    ((job_index++))
+  done
+done
+
+total_jobs=${#JOBS[@]}
+echo "Starting $total_jobs tests..."
+echo ""
+
+# Run first test sequentially if any jobs exist
+if [ $total_jobs -gt 0 ]; then
```
jamadeo, line 288: Running one alone before the rest concurrently is kind of silly, but solves two small pain points. If you immediately start N concurrent sessions, then:

- locally, you'll get the keychain password prompt N times, even if you click "always allow"
- in CI, they all try to create the sqlite database at the same time, and that leads to things breaking. Maybe we should fix that but it also seems highly unlikely to happen in the real world

DOsinga: it's also useful since most likely if you break something it is broken for all providers. this way it tells you that immediately maybe

michaelneale: yeah the faster inner loop - I think is reasonable

jamadeo:
> Code mode helping a lot?

I think the main difference there is that the other one also runs a compaction test which takes 4-5 min right now. Maybe that should be split out of sequence

PR #6857, opened 2026-01-31
docs: White label guide

maniksurtani:
### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [X] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### DIFF
--- a/documentation/docs/guides/custom-distributions.md
+++ b/documentation/docs/guides/custom-distributions.md
@@ -6,7 +6,7 @@ sidebar_label: Custom Distributions
 
 # Custom Distributions
 
-goose is designed to be forked and customized. You can create your own "distro" of goose — preconfigured with specific providers, bundled extensions, custom branding, and tailored workflows for your organization or audience.
+goose is designed to be forked and customized. You can create your own "distro" of goose preconfigured with specific providers, bundled extensions, custom branding, and tailored workflows for your organization or audience.
 
 ## What you can customize
 


katzdave:
@block/goose-devrel do you think worth having here in the main repo? Or do we have similar info in the goose docs?

WHITE_LABEL.md
```diff
@@ -0,0 +1,804 @@
+# White-Labeling Goose
+
+This guide explains how to create custom distributions of goose tailored to your organization's needs—whether that's preconfigured models, custom tools, branded interfaces, or entirely new user experiences.
+
+## Overview
+
+Goose's architecture is designed for extensibility. Organizations can create "remixed" versions that:
```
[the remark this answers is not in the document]
angiejones: goose (lowercase) please

WHITE_LABEL.md
```diff
@@ -0,0 +1,804 @@
+# White-Labeling Goose
```
angiejones: white labeling doesn't seem like the right term for open source projects. how about using "Custom Distributions"?

angiejones: do you think this would be best as a top level file in the repo (as you have it now) or a guide on the doc site (with the readme pointing to this link)?

angiejones:
> @block/goose-devrel do you think worth having here in the main repo? Or do we have similar info in the goose docs?

ha, i had the same thought. i think a doc is better, and the README can point to it

maniksurtani:
How about this:

1. We move this content to the docs site
2. We point to it from README.md
3. We still keep this file (WHITE_LABEL.md -- or better, call it CUSTOM_DISTROS.md) which also links to the docs site, just so we have that added top-level visibility. 

WDYT @angiejones @michaelneale?

michaelneale:
ok, made some docs, and tidied up, moved it to distros md (I think .md makes sense - as it is a dev time activity - and want people to discover it even as they browse - added a link from readme too) doc is just to help highlight it. 

cc @angiejones for your consideration

WHITE_LABEL.md
michaelneale: decided to keep top level, but have a doc refer to it (maximum discovery for developers - who it is for. Documentation is just for discovery when people don't browse via the repo, but the real word is in the repo md)

Issue #6858, opened 2026-01-31
Install script fails because download link doesn't exist


IAmMarcellus:
**Describe the bug**

Running `curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh | bash` to do a fresh install in linux fails with this output:
```
cli.sh | bash
WINDIR: <not set>
OSTYPE: linux-gnu
uname -s: Linux
uname -m: x86_64
PWD: /home/marcellus/Projects/AlgoCryptoTradingBot-hybrid
Detected OS: linux with ARCH x86_64
Downloading stable release: goose-x86_64-unknown-linux-gnu.tar.bz2...
Error: Failed to download https://github.com/block/goose/releases/download/stable/goose-x86_64-unknown-linux-gnu.tar.bz2
```

Its happening because `https://github.com/block/goose/releases/download/stable/goose-x86_64-unknown-linux-gnu.tar.bz2` doesn't exist (the [most recent CI job](https://github.com/block/goose/actions/runs/21529244002) on  release failed, maybe that's why?).

The same url `/1.21.1/` in place of `/stable/` does exist so it works if you set `GOOSE_VERSION` before running the install command/script.

---

**To Reproduce**
1. Make sure GOOSE_VERSION is unset in your shell
2. Run `curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh | bash`
3. See error

---

**Expected behavior**
Goose installs

---

**Please provide the following information**
- **OS & Arch:** Ubuntu 22.04 x86
- **Interface:** CLI
- **Version:** v1.22.1
- **Extensions enabled:** N/A
- **Provider & Model:** N/A

---

**Additional context**

IAmMarcellus:
Created a PR for it: https://github.com/block/goose/pull/6859

Cyber-Syntax:
Same error with cli and `goose update` command in version 1.18.0:

```
goose update
WINDIR: <not set>
OSTYPE: linux-gnu
uname -s: Linux
uname -m: x86_64
PWD: /home/developer/Downloads
Detected OS: linux with ARCH x86_64
Downloading stable release: goose-x86_64-unknown-linux-gnu.tar.bz2...
Error: Failed to download https://github.com/block/goose/releases/download/stable/goose-x86_64-unknown-linux-gnu.tar.bz2
```
```
curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh | bash
WINDIR: <not set>
OSTYPE: linux-gnu
[...]

PR #6859, opened 2026-01-31
Fallback to latest version during failed install

IAmMarcellus:
## Summary

Fix for https://github.com/block/goose/issues/6858

Installs fail because sometime `goose-x86_64-unknown-linux-gnu.tar.bz2` doesn't exist at `github.com/block/goose/releases/download/stable` but it exists at `/goose/releases/download/<version_number>`.

Now, if the download fails and a GOOSE_VERSION isn't specified, it will get the latest version from github and retry with that version explicitly. This avoids a failure when the `stable` tag is missing its download files.

### Type of Change

- [ ] Feature
- [ X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Script is working locally for me and there isn't a download file on the stable tag right now.

### Related Issues
Relates to #6858

### DIFF
--- a/download_cli.sh
+++ b/download_cli.sh
@@ -202,7 +202,7 @@ if ! curl -sLf "$DOWNLOAD_URL" --output "$FILE"; then
     LATEST_TAG=$(curl -s https://api.github.com/repos/block/goose/releases/latest | \
       grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
     if [ -z "$LATEST_TAG" ]; then
-      echo "Error: Failed to download $DOWNLOAD_URL"
+      echo "Error: Failed to download $DOWNLOAD_URL and latest tag unavailable"
       exit 1
     fi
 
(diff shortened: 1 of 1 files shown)


download_cli.sh
```diff
@@ -197,8 +197,27 @@ DOWNLOAD_URL="https://github.com/$REPO/releases/download/$RELEASE_TAG/$FILE"
...
+    DOWNLOAD_URL="https://github.com/$REPO/releases/download/$LATEST_TAG/$FILE"
+    if curl -sLf "$DOWNLOAD_URL" --output "$FILE"; then
+      # Fallback succeeded
+      :
+    else
+      echo "Error: Failed to download $DOWNLOAD_URL"
+      exit 1
+    fi
+  else
+    echo "Error: Failed to download $DOWNLOAD_URL"
```
kravtsun: ❓  Are these two error messages the same because the problem is the same?

If not, is that possible to make them different, so that a user could understand the problem more precise?

IAmMarcellus: Sure! Fixed

DOsinga:
ported this over so we can merge without the DCO - thanks @IAmMarcellus for taking care of this

PR #6861, opened 2026-01-31
fix: Anthropic prompt caching behavior

GordianDziwis:
## Summary

- Move MOIM insertion to end of conversation to preserve cache prefix. Previously inserted after last assistant message, breaking cache continuity
- Refactor cache_control logic to cache only last valid content block Simplifies code from nested loops to functional chain 
- Excludes MOIM messages (<info-msg>) from caching Only caches valid block types: text, image, tool_use, tool_result

A clean solution would be to add MOIM messages as an explicit message type in goose model.

### Type of Change
- [x] Bug fix

### AI Assistance

### Testing

Updated existing tests.

### DIFF
--- a/crates/goose/src/agents/moim.rs
+++ b/crates/goose/src/agents/moim.rs
@@ -4,6 +4,8 @@ use crate::conversation::{fix_conversation, Conversation};
 use rmcp::model::Role;
 use std::path::Path;
 
+pub const MOIM_PREFIX: &str = "<info-msg>";
+
 // Test-only utility. Do not use in production code. No `test` directive due to call outside crate.
 thread_local! {
     pub static SKIP: std::cell::Cell<bool> = const { std::cell::Cell::new(false) };
@@ -24,11 +26,11 @@ pub async fn inject_moim(
         .await
     {
         let mut messages = conversation.messages().clone();
-        let idx = messages
+        let index = messages
             .iter()
             .rposition(|m| m.role == Role::Assistant)
-            .unwrap_or(0);
-        messages.insert(idx, Message::user().with_text(moim));
+            .map_or(0, |index| index + 1);
+        messages.insert(index, Message::user().with_text(moim));
 
         let (fixed, issues) = fix_conversation(Conversation::new_unvalidated(messages));
 
@@ -54,7 +56,7 @@ mod tests {
     use std::path::PathBuf;
 
     #[tokio::test]
-    async fn test_moim_injection_before_assistant() {
+    async fn test_moim_injection_after_assistant() {
         let temp_dir = tempfile::tempdir().unwrap();
         let em = ExtensionManager::new_without_provider(temp_dir.path().to_path_buf());
         let working_dir = PathBuf::from("/test/dir");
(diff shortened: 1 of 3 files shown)


GordianDziwis:
Do I need to create an issue first to get this reviewed?

GordianDziwis:
Found another case where the cache gets invalidated:

Left: last request
Right: previous request

image

It looks like the tool call disappeared.

GordianDziwis:
https://github.com/block/goose/issues/6941

rabi:
> Previously inserted after last assistant message, breaking cache continuity

I think it was before the last assistant message, no? Probably there is some reason for that though not documented. The move to end-placement sounds reasonable to me though. Could you consider adding a cacheable: bool field to MessageMetadata instead of string-matching <info-msg> in the Anthropic formatter - this would make the solution provider-agnostic and explicit in the data model?

GordianDziwis:
@rabi should I also encode the MOIN message as a type?

rabi:
> @rabi should I also encode the MOIN message as a type?

I was thinking about using something like..

```
pub struct MessageMetadata {
    pub user_visible: bool,
    pub agent_visible: bool,
    #[serde(default = "default_true")]
    pub cacheable: bool,
}

```
Then in moim.rs, mark MOIM as non-cacheable

```
messages.push(
    Message::user()
        .with_text(moim)
        .not_cacheable()  // sets metadata.cacheable = false
);
```

Then providers check message.metadata.cacheable instead of parsing content.

GordianDziwis:
Thanks for suggestions @rabi!

  I ended up going a different direction, mainly because Anthropic's caching works at the content block level rather than the message level. A single message can have multiple content blocks, and we need to mark the last cacheable block across the entire conversation - which we can only determine after the final payload is assembled.

  This is also why I moved MOIM to the end of the conversation: it keeps all the cacheable content in a contiguous prefix, and then we apply cache markers as a final transformation step.

  I separated the caching logic into its own add_cache_markers(payload) function that runs after conversion. This keeps the format functions as pure converters and makes it easy to swap in different caching strategies if needed. It's also much easier to debug cache control behavior when it's isolated in one place rather than interleaved with the conversion logic.

  I did add MOIM_PREFIX as a constant in moim.rs to address the string matching issue. 

  Let me know if you have any concerns with this approach!

crates/goose/src/agents/moim.rs
```diff
@@ -24,11 +25,8 @@ pub async fn inject_moim(
         .await
     {
         let mut messages = conversation.messages().clone();
-        let idx = messages
-            .iter()
-            .rposition(|m| m.role == Role::Assistant)
-            .unwrap_or(0);
-        messages.insert(idx, Message::user().with_text(moim));
+        // Append MOIM at the end to preserve cache prefix
+        messages.push(Message::user().with_text(moim));
```
tlongwell-block: This ends up distracting the LLM too much, hence why we push it back.

The MOIM should be recent in context, but never the latest message. The behavior gets weird.

GordianDziwis: I did put it after the last assistant message and before the last user message. So the history won't be rewritten and the moim message is not the last message.

michaelneale:
@GordianDziwis sorry it is late - but interested in updating this?

michaelneale:
closing due to activity, please do we open when time

PR #6863, opened 2026-01-31
feat: pass env to shell command

lsytj0413:
### Type of Change

- [x] Bug fix

### AI Assistance

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:


alexhancock:
@jamadeo Remind me, did we used to do this and we removed it? Trying to remember if there is some reason why we don't want to

jamadeo (approved):
@alexhancock I do think this is (was) a bug - I'm not sure why we wouldn't pass this. It's related, but separate from the back-and-forth we did with setting PATH using an interactive shell. Thanks for the fix @lsytj0413

PR #6872, opened 2026-01-31
Ef/moonshot provider

Evanfeenstra:
## Summary
Added `moonshot` provider. Since kimi reasoning output is in a different format that other models, it cant be added as a custom provider

### Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
- unit tests
- testing manually (tool use, streaming responses)

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:

### DIFF
--- a/crates/goose/src/providers/formats/moonshot.rs
+++ b/crates/goose/src/providers/formats/moonshot.rs
@@ -4,7 +4,7 @@ use crate::providers::formats::openai;
 use anyhow::anyhow;
 use async_stream::try_stream;
 use futures::Stream;
-use rmcp::model::{object, CallToolRequestParam, ErrorCode, ErrorData, Role};
+use rmcp::model::{object, CallToolRequestParams, ErrorCode, ErrorData, Role};
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
 use std::borrow::Cow;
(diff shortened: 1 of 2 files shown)


DOsinga:
before we merge this, can we think about whether we can do this with a custom provider? I see we need changes to the formatter, but can we add custom rendering so we can accomplish that? the provider code itself looks like a fairily standard openai provider?

making it easier to add custom/declarative providers makes the whole system stronger!

DOsinga:
any more thoughts on making this a custom provider? what would be needed?

Evanfeenstra:
@DOsinga sry been super busy lately!

what do you think? Should we make some small changes to openai formatting code to support this? Or create totally separate provider code and just try to use some parts of the openai format?

DOsinga:
I think using openai is the way - there's already some support there for different thinking modes. I could imagine this all ends with having a canonical provider class where you can specify how it handles thinking bits, similar to how you can now say, this one uses streaming. thoughts?

stebbins:
@DOsinga were you thinking something more like this approach?

[moonshot support PR](https://github.com/block/goose/pull/6962) / [kimi code support PR](https://github.com/block/goose/pull/7222)

note; they both implement the same changes to openai provider. I _think_ this might improve deepseek `reasoning_content` handling as well, but I can't test as I don't have access.

DOsinga:
hi! yes - just approved that one with the note that it would be good if we could make sure deepseek still works :)

DOsinga:
ok, so closing this and we can merge the others

PR #6263, opened 2025-12-24
feat: add OpenAI Codex CLI provider

RamXX:
## Summary

Add a new CLI provider that integrates with OpenAI's Codex CLI tool, allowing users to use their ChatGPT Plus/Pro subscription through goose's interface. This follows the same pattern as the existing Claude Code provider.

**Features:**
- Subprocess-based provider using `codex exec` command for non-interactive mode
- Configurable reasoning effort levels (low/medium/high) via `CODEX_REASONING_EFFORT`
- Optional skills support via `CODEX_ENABLE_SKILLS`
- GOOSE_MODE permission mapping (auto->--yolo, smart-approve->--full-auto, etc.)
- JSONL response parsing for structured output
- Automatic filtering of goose extensions from system prompts
- Optional `--skip-git-repo-check` via `CODEX_SKIP_GIT_CHECK`

**Supported models (Codex CLI v0.77.0):**
- gpt-5.2-codex, gpt-5.2: 400K (auto-compacting)
- gpt-5.1-codex-max, gpt-5.1-codex-mini: 256K

**Configuration:**
- `GOOSE_PROVIDER=codex` to enable
- `CODEX_COMMAND` for custom CLI path (default: codex)
- `CODEX_REASONING_EFFORT`: low, medium, high (default: high)
- `CODEX_ENABLE_SKILLS`: true/false (default: true)
- `CODEX_SKIP_GIT_CHECK`: true/false (default: false)

### Type of Change
- [x] Feature
- [x] Documentation
- [x] Tests

### AI Assistance
- [x] This PR was created or reviewed with AI assistance

This PR was developed with Claude Code assistance. All code was reviewed, tested manually with real Codex CLI integration, and validated against existing patterns in the codebase (particularly claude_code.rs). Unit tests achieve comprehensive coverage of the provider functionality.

## Test plan

- [x] Unit tests (20 tests, all passing): `cargo test --package goose --lib providers::codex::tests`
- [x] Manual integration testing with actual Codex CLI
- [x] Verified JSONL parsing matches real Codex output format
- [x] Tested `CODEX_SKIP_GIT_CHECK=true` for non-git directories

**Manual test:**
```bash
echo "What is 2+2? Reply in one word only." | GOOSE_PROVIDER=codex goose run -q -i -
# Output: 4
```

## Requirements

- Codex CLI installed (`npm i -g @openai/codex` or `brew install --cask codex`)
- Active ChatGPT Plus/Pro subscription or OpenAI API credits
[...]

PR #6874, opened 2026-02-01
fix(codex): propagate extended PATH to codex subprocess

RamXX:
## Summary

When the Codex CLI provider spawns `codex exec`, the subprocess inherits the parent process's PATH. When goose is launched from the desktop app (Electron), this PATH is significantly more limited than a terminal session -- it lacks paths added by nvm, volta, pnpm, fnm, and other Node version managers. This causes the codex subprocess to fail finding Node.js and other dependencies.

This fix adds PATH propagation using `SearchPaths::builder().with_npm().path()`, matching the existing pattern already used by `gemini_cli.rs`. The subprocess now gets access to common binary locations (`~/.local/bin`, `/opt/homebrew/bin`, `~/.npm-global/bin`, etc.) plus the system PATH.

### Type of Change
- [x] Bug fix

### AI Assistance

### Testing
- All 30 existing codex provider unit tests pass
- `cargo fmt`, `cargo clippy -D warnings` clean
- Manual testing: verified `codex exec` works with the extended PATH
- The fix is a 4-line addition following an established pattern (gemini_cli.rs)

### Related Issues
Relates to #6263
Reported by @Aerasyn: https://github.com/block/goose/pull/6263#issuecomment-3812629554
Also noted by @michaelneale during original review: "I had some issues with my codex install and the desktop - might have to have a follow on, but ran out of time for it. Something to do with path maybe."

### DIFF
--- a/crates/goose/src/providers/codex.rs
+++ b/crates/goose/src/providers/codex.rs
@@ -176,6 +176,13 @@ impl CodexProvider {
         let mut cmd = Command::new(&self.command);
         configure_command_no_window(&mut cmd);
 
+        // Propagate extended PATH so the codex subprocess can find Node.js
+        // and other dependencies (especially when launched from the desktop app
+        // where the inherited PATH is limited).
+        if let Ok(path) = SearchPaths::builder().with_npm().path() {
+            cmd.env("PATH", path);
+        }
+
         // Use 'exec' subcommand for non-interactive mode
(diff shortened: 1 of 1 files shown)


RamXX:
Thanks @michaelneale . Goose is moving so fast I can't keep up! thanks for the tip. I'll check it out.

PR #6877, opened 2026-02-01
Careless whisper

DOsinga:
## Summary

Trans-atlantic plane wifi FTW. 

So this is based on an earlier restructuring of how we record audio, but beside that it, it implements local whisper support with download & model recommendation.

it also updates the client so we detect pauses for pseudo streaming and auto submit if you end your sentence on, well, submit.

bonus: also adds groq as a provider (faster than openai)

### DIFF
--- a/crates/goose/src/dictation/whisper.rs
+++ b/crates/goose/src/dictation/whisper.rs
@@ -3,7 +3,8 @@
 //! This module provides local audio transcription using OpenAI's Whisper model
 //! via the Candle ML framework. It supports loading GGUF quantized models for
 //! efficient CPU inference.
-//! Heavily "inspired" by the reference implementation
+//! Heavily "inspired" by the Candle Whisper example:
+//! https://github.com/huggingface/candle/tree/main/candle-examples/whisper
 
 use crate::config::paths::Paths;
 


crates/goose/Cargo.toml
```diff
@@ -7,6 +7,12 @@ license.workspace = true
 repository.workspace = true
 description.workspace = true
 
+[features]
```
michaelneale, line 10: will this mean it only works on macos?

DOsinga: I think it means we compile this with this for all platforms, but it will only trigger on macos. we should test of course!

michaelneale (approved):
seems ok - was able to download, it worked with my accent (but not sure if i was doing it wrong as it didn't submit for me - but that was ok). 

Probably some valid nits from copilot but nice (assuming I tried the right thing). I wonder if we could make it more obvious we reocmmend the local one. The word "Offline" at first made me think it was busted! 

(could also have the microphone icon back on main chat which takes you to the config as the config seems so complex now, or ... maybe rich I should say!)

jh-block (approved):
Looks good! I think we can do a little better with the platform-specific candle features. If we move the `candle-core` and `candle-nn` deps to be target-specific dependencies, we can automatically set the features on the appropriate targets without needing our own cuda/metal features:

untested:
```
[dependencies]
# ...
candle-core = { version = "0.8" }
candle-nn = { version = "0.8" }

[target.'cfg(target_os = "macos")'.dependencies]
candle-core = { version = "0.8", features = ["metal"] }
candle-nn = { version = "0.8", features = ["metal"] }

[target.'cfg(target_os = "linux")'.dependencies]
candle-core = { version = "0.8", features = ["cuda"] }
candle-nn = { version = "0.8", features = ["cuda"] }
```

(and remove the features on our crate as no longer needed)

DOsinga:
> seems ok - was able to download, it worked with my accent (but not sure if i was doing it wrong as it didn't submit for me - but that was ok).

it only submits if you end your sentence on submit. unfortunately though whisper doesn't process single words. so if you enter a sentence, then pause for the transcribing to kick in and then just say submit, it won't do anything. I tried playing with padding but it doesn't really help. mostly people will just append submit directly I think, just when you test not so much.

> (could also have the microphone icon back on main chat which takes you to the config as the config seems so complex now, or ... maybe rich I should say!)

oh, I like that - always show the microphone button and link to the setting you mean?

> Looks good! I think we can do a little better with the platform-specific candle features. If we move the candle-core and candle-nn deps to be target-specific dependencies, we can automatically set the features on the appropriate targets without needing our own cuda/metal features:

yeah great idea. I'll do it in a follow up though, gotta keep the PR count up! or really, I would like to ship this in todays release.

DOsinga:
actually @DOsinga that feels true for metal, but cuda you could have on any platform, no?

taniacryptid:
Random commentary: Living for this reference, actually stoked for local whisper support!!

PR #6882, opened 2026-02-01
fix: add XML tool call parsing fallback for Qwen3-coder via Ollama

kuccello:
When using Qwen3-coder model through Ollama with many tools (6+), the model outputs XML-style tool calls in the content field instead of using the native JSON tool_calls format. This causes Goose to fail to execute any tools.

This commit adds XML parsing as a fallback mechanism:

- Add parse_xml_tool_calls() helper function using regex to parse <function=name><parameter=key>value</parameter></function> format
- Modify response_to_message() to check for XML tool calls when no JSON tool_calls are found in the response
- Modify response_to_streaming_message() to accumulate text and parse XML tool calls when stream completes
- Add comprehensive unit tests for XML parsing

The fix is backward-compatible and only activates when JSON tool_calls are absent, ensuring existing providers continue to work normally.

Tested with Qwen3-coder:latest via Ollama with 11 developer tools.

### Type of Change

- [X] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [X] Tests
- [ ] Security fix
- [ ] Build / Release
- [X] Other (specify below)
    - qwen3-coder behavior (other qwen models too)

### AI Assistance

### Testing

- Unit tests passing: `cargo test -p goose --lib -- formats::openai::tests` - green
- Clippy check: `cargo fmt -p goose && cargo clippy -p goose --lib -- -D warnings`

### Related Issues
Relates to #6883
Discussion: LINK (if any)

Before:  

After:

### DIFF
--- a/crates/goose/src/providers/formats/mod.rs
+++ b/crates/goose/src/providers/formats/mod.rs
@@ -3,6 +3,7 @@ pub mod bedrock;
 pub mod databricks;
 pub mod gcpvertexai;
 pub mod google;
+pub mod ollama;
 pub mod openai;
 pub mod openai_responses;
 pub mod openrouter;
--- a/crates/goose/src/providers/formats/ollama.rs
+++ b/crates/goose/src/providers/formats/ollama.rs
@@ -1,0 +1,422 @@
+//! Ollama-specific response handling with XML tool call fallback.
+//!
+//! Some models running through Ollama (notably Qwen3-coder) output XML-style tool calls
+//! when given many tools (6+), instead of using the native JSON tool_calls format.
+//! This module wraps the standard OpenAI response parsing with XML fallback logic,
+//! isolating this behavior to the Ollama provider only.
+//!
+//! Known affected models:
+//! - qwen3-coder
+//! - qwen3-coder-32b
+
+use crate::conversation::message::{Message, MessageContent};
+use crate::providers::base::ProviderUsage;
+use crate::providers::utils::is_valid_function_name;
+use async_stream::try_stream;
+use chrono;
+use futures::Stream;
+use regex::Regex;
+use rmcp::model::{object, CallToolRequestParams, ErrorCode, ErrorData, Role};
+use serde_json::Value;
+use std::borrow::Cow;
+use uuid::Uuid;
+
+pub use super::openai::{
+    create_request, format_messages, format_tools, get_usage, validate_tool_schemas,
+};
+
+/// Parse XML-style tool calls from content (Ollama/Qwen3-coder fallback format).
+///
+/// Format: `<function=name><parameter=key>value</parameter>...</function>`
+///
+/// Returns a tuple of (prefix_text, tool_calls) where prefix_text is any text before the first function tag.
+pub fn parse_xml_tool_calls(content: &str) -> (Option<String>, Vec<MessageContent>) {
+    let mut tool_calls = Vec::new();
+
+    let function_re = Regex::new(r"<function=([^>]+)>([\s\S]*?)</function>").unwrap();
+    let param_re = Regex::new(r"<parameter=([^>]+)>([\s\S]*?)</parameter>").unwrap();
+
+    let prefix = content
+        .find("<function=")
+        .and_then(|idx| content.get(..idx))
+        .map(|s| s.trim())
+        .filter(|s| !s.is_empty())
+        .map(|s| s.to_string());
+
+    for func_cap in function_re.captures_iter(content) {
+        let function_name = func_cap[1].trim().to_string();
+        let function_body = &func_cap[2];
+        let mut arguments = serde_json::Map::new();
+        for param_cap in param_re.captures_iter(function_body) {
+            let param_name = param_cap[1].trim().to_string();
+            let param_value = param_cap[2].trim().to_string();
+            arguments.insert(param_name, serde_json::Value::String(param_value));
+        }
+
+        let id = Uuid::new_v4().to_string();
+
+        if is_valid_function_name(&function_name) {
+            tool_calls.push(MessageContent::tool_request(
+                id,
+                Ok(CallToolRequestParams {
+                    meta: None,
+                    task: None,
+                    name: function_name.into(),
+                    arguments: Some(object(serde_json::Value::Object(arguments))),
+                }),
+            ));
+        } else {
+            let error = ErrorData {
+                code: ErrorCode::INVALID_REQUEST,
+                message: Cow::from(format!(
+                    "The provided function name '{}' had invalid characters, it must match this regex [a-zA-Z0-9_-]+",
+                    function_name
+                )),
+                data: None,
+            };
+            tool_calls.push(MessageContent::tool_request(id, Err(error)));
+        }
+    }
+
+    (prefix, tool_calls)
+}
+
+/// Convert OpenAI's API response to internal Message format, with XML tool call fallback.
+///
+/// This wraps the standard OpenAI response parsing and adds XML fallback for models
+/// like Qwen3-coder that output XML tool calls when given many tools.
+pub fn response_to_message(response: &Value) -> anyhow::Result<Message> {
+    let message = super::openai::response_to_message(response)?;
+
+    let has_tool_requests = message
+        .content
+        .iter()
+        .any(|c| matches!(c, MessageContent::ToolRequest(_)));
+
+    if has_tool_requests {
+        return Ok(message);
+    }
+
+    let original = response
+        .get("choices")
+        .and_then(|c| c.get(0))
+        .and_then(|m| m.get("message"));
+
+    if let Some(original) = original {
+        if let Some(text) = original.get("content").and_then(|c| c.as_str()) {
+            if text.contains("<function=") {
+                let (prefix, xml_tool_calls) = parse_xml_tool_calls(text);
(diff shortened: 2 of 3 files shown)


crates/goose/src/providers/formats/openai.rs
```diff
@@ -17,6 +18,66 @@ use serde::{Deserialize, Serialize};
...
+/// Format: <function=name><parameter=key>value</parameter>...</function>
+/// Returns a tuple of (prefix_text, tool_calls) where prefix_text is any text before the first function tag
+fn parse_xml_tool_calls(content: &str) -> (Option<String>, Vec<MessageContent>) {
+    let mut tool_calls = Vec::new();
+
+    // Match <function=name>...</function> blocks
+    let function_re = Regex::new(r"<function=([^>]+)>([\s\S]*?)</function>").unwrap();
+    let param_re = Regex::new(r"<parameter=([^>]+)>([\s\S]*?)</parameter>").unwrap();
+
+    // Find prefix text (anything before the first <function= tag)
```
michaelneale: could drop this comment and the match one above - they get a little too keen in duplicating what code does which can cause future bugs (the doc for the function is fine though)

crates/goose/src/providers/formats/openai.rs
```diff
@@ -17,6 +18,66 @@ use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
 use std::borrow::Cow;
 use std::ops::Deref;
+use uuid::Uuid;
+
+/// Parse XML-style tool calls from content (Qwen3-coder fallback format when many tools are provided)
```
michaelneale: this is likely for other models too - this is in the very heavily used openai.rs - so we want it to be clear if it is general or specific (and also that it won't do something incorrect!)

crates/goose/src/providers/formats/openai.rs
```diff
@@ -620,6 +711,42 @@ where
                 )
             } else if chunk.choices[0].delta.content.is_some() {
                 let text = chunk.choices[0].delta.content.as_ref().unwrap();
+
+                // Accumulate text for potential XML tool call detection
```
michaelneale: ditto for these comments - once claude catches on to comments it won't stop haha

crates/goose/src/providers/formats/openai.rs
```diff
@@ -1661,4 +1788,190 @@ data: [DONE]
 
         Ok(())
     }
+
+    // Tests for XML tool call parsing (Qwen3-coder fallback format)
+
+    #[test]
+    fn test_parse_xml_tool_calls_single() {
```
michaelneale: nice to have these tests

michaelneale:
nice - have pushed this here: https://github.com/block/goose/pull/6886 to get to run live tests, but I think this looks good. If that passes can merge this. thanks @kuccello

michaelneale (approved):
nice!

crates/goose/src/providers/formats/openai.rs
kuccello: will modify

crates/goose/src/providers/formats/openai.rs
kuccello: Added a commit that addresses comments on comments ;-)

kuccello:
I added a commit to address the comments:

- Added xml_detected flag - Tracks when XML tool call markers (<function=) are detected in the accumulated text
    - Modified streaming logic - Once XML is detected:
    - Text chunks are buffered instead of yielded
- On final chunk, only the parsed tool calls are emitted (or the accumulated text if parsing fails)

This change is to prevent noisy duplicate response text showing up in the conversation flow.

michaelneale:
thanks @kuccello - looks nice - BTW I am working on a extensive model/extension combo to test with open models - and this will solve a recurring issue I see in test suites.

codefromthecrypt:
@kuccello so I'm concerned about doing more stuff like this in the openai provider as it puts main openai at risk..

you can see conditional body handling scattered per provider in some cases, such as this in the tetrate one

```rust
        // Handle Google-compatible model responses differently
        if is_google_model(payload) {
            return handle_response_google_compat(response).await;
        }

        // For OpenAI-compatible models, parse the response body to JSON
        let response_body = handle_response_openai_compat(response)
            .await
            .map_err(|e| ProviderError::RequestFailed(format!("Failed to parse response: {e}")))?;
```

Agree normal responses are a lot less tricky than tool responses.

https://github.com/block/goose/pull/6832 introduces a different type `OpenAiCompatibleProvider` to help isolate code that is about clones and not in any way about openai the platform (`openai.rs`). In its first revision, this does not support body customizations. It only migrated xai and azure to it, not yet ollama.

For now, it would be best to move this code to the ollama provider because it has templating that isn't guaranteed to act the same in other providers. Since it is left alone and not using `OpenAiCompatibleProvider`, https://github.com/block/goose/pull/6832 shouldn't interfere too much with your work.

If you run into a problem in refactoring and would prefer a hand, I can try to reproduce your work in a way that doesn't affect the main OpenAiProvider.

michaelneale:
@codefromthecrypt @kuccello we could also move to a branch so it is covered by e2e tests with live providers while at it? really good to get this in

kuccello:
@codefromthecrypt I will make an adjustment based on your comment

kuccello:
Here's a summary of the changes for the Qwen3-coder XML tool call fix:

## Problem

When using Qwen3-coder through Ollama with many tools (6+), the model outputs XML-style tool calls instead of the expected JSON format:

```xml
<function=developer__text_editor><parameter=command>view</parameter><parameter=path>/some/path</parameter></function>
```

Instead of:

```json
{"tool_calls": [{"function": {"name": "developer__text_editor", "arguments": {"command": "view", "path": "/some/path"}}}]}
```

## Architectural Solution

I implemented an **Ollama-specific post-processing wrapper** that isolates the XML parsing logic to the Ollama provider only, without modifying the core OpenAI format handling.

### Files Changed

1. **`crates/goose/src/providers/formats/ollama.rs`** (new file)
   - Contains `parse_xml_tool_calls()` - regex-based XML parser
   - Contains `response_to_message()` - wraps standard OpenAI parsing with XML fallback
   - Contains `response_to_streaming_message_ollama()` - streaming wrapper that buffers text when XML markers are detected

2. **`crates/goose/src/providers/formats/mod.rs`**
   - Added `pub mod ollama;` to expose the new module

3. **`crates/goose/src/providers/openai_compatible.rs`**
   - Added `stream_ollama()` function that uses the Ollama-specific streaming handler

4. **`crates/goose/src/providers/ollama.rs`**
   - Updated imports to use `formats::ollama` instead of `formats::openai`
   - Calls `stream_ollama()` for streaming responses

### Key Design Decisions

1. **Isolation**: The XML parsing is completely isolated to Ollama. The core `formats/openai.rs` remains unchanged, so other OpenAI-compatible providers (Azure, XAI, etc.) are unaffected.

2. **Post-Processing Wrapper Pattern**: Rather than exposing internal streaming types from `openai.rs`, the Ollama streaming handler wraps the standard OpenAI stream and post-processes `Message` objects. This avoids tight coupling.

3. **Streaming Buffering**: When XML markers (`<function=`) are detected during streaming, text chunks are buffered instead of being emitted immediately. At stream end, the buffered content is parsed for XML tool calls. This prevents duplicate output (raw XML fragments followed by parsed tool calls).

4. **Re-exports**: `formats/ollama.rs` re-exports the standard OpenAI utilities (`create_request`, `format_messages`, `format_tools`, `get_usage`, `validate_tool_schemas`) so the Ollama provider can use them without importing from two places.

### Data Flow

```
Ollama API Response
        │
        ▼
┌─────────────────────────────────┐
│  openai::response_to_message()  │  ← Standard JSON parsing
└─────────────────────────────────┘
        │
        ▼
   Has tool_calls?
        │
   ┌────┴────┐
   │ Yes     │ No
   ▼         ▼
 Return   Contains "<function="?
           │
      ┌────┴────┐
      │ Yes     │ No
      ▼         ▼
  Parse XML   Return original
  tool calls  message
      │
      ▼
  Return message
  with tool requests
```

### Why This Approach?

- **Minimal risk**: Core OpenAI handling untouched
- **No public API changes**: Internal types in `openai.rs` remain private
- **Follows existing patterns**: Similar to how `handle_response_google_compat` is used for Google-specific handling in other providers
- **Extensible**: If other Ollama models exhibit similar behavior, they'll automatically benefit

crates/goose/src/providers/openai_compatible.rs
```diff
@@ -287,3 +288,29 @@ pub fn stream_openai_compat(
         }
     }))
 }
+
+/// Ollama-specific streaming handler with XML tool call fallback.
+/// Uses the Ollama format module which buffers text when XML tool calls are detected,
+/// preventing duplicate content from being emitted to the UI.
+pub fn stream_ollama(
```
codefromthecrypt: this is an abstraction break, as we are putting provider-specific logic into the openai_compatible.rs

can you please contain changes related to ollama to files named ollama?

kuccello:
All comments addressed

codefromthecrypt (approved):

codefromthecrypt:
@kuccello thanks for your patience. very helpful!

timohuovinen:
🚀 !!! Thank you all

Issue #6562, opened 2026-01-18
Recipe in the Goose Desktop UI, the recipe details show the globally-selected provider/model instead of the provider/model specified inside the recipe's YAML config file.


n93013:
**Description the bug**

When viewing a Recipe in the Goose Desktop UI, the recipe details show the globally-selected provider/model instead of the provider/model specified inside the recipe's YAML. In other words, even though the recipe YAML explicitly sets a different provider/model, the Recipe details page displays (and therefore appears to indicate) the global LLM selection, which is misleading and makes it hard to confirm which model the recipe will run with.

---

**To Reproduce**

- In Goose Desktop Settings → Provider / Models, set a global provider & model (for example: "OpenAI — gpt-4o").
- Create a recipe YAML that explicitly specifies a different provider/model. Example 

```
version: 1.0.0
title: Example recipe that targets a different model
description: Demo
goose_provider: anthropic
goose_model: claude-sonnet-4
```

- Open the Recipes view in the Goose UI and either Import the recipe or open the recipe file from the UI.
- Inspect the recipe details / metadata /footer pane (where it displays Provider & Model).
See the error: the UI shows the global provider/model ("OpenAI — gpt-4o" in the example) instead of the recipe's specified provider/model ("anthropic / claude-sonnet-4").

---

**Expected behavior**
- The recipe page should show the provider & model declared in the recipe YAML (if present).
- If the UI must show the global model somewhere, it should be clearly separated from the recipe's model so users cannot mistake the global model for the recipe's configured model.

---

**Screenshots**
![Image](https://github.com/user-attachments/assets/2baa8785-6705-4666-a084-75773b69d3cf)

---

OS & Arch: macOS (Tahoe - 26.1, Mac Mini - M.2)
Interface: UI (Desktop)
Version: Goose Desktop version v1.20.0
Provider & Model: OpenAI gpt-5-mini

PR #6884, opened 2026-02-01
feat: load provider/model specified inside the recipe config

Abhijay007:
closes: https://github.com/block/goose/issues/6562

## PR description 

The Recipe details view is likely pulling the provider/model from the global settings state instead of reading it from the recipe's YAML configuration.

### Type of Change

- [x] Feature

### AI Assistance

### Testing

Tested in the desktop UI with recipes and different cases 

Before:  

After:

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -1566,30 +1566,18 @@ impl Agent {
     pub async fn restore_provider_from_session(&self, session: &Session) -> Result<()> {
         let config = Config::global();
 
-        let recipe_provider = session
-            .recipe
-            .as_ref()
-            .and_then(|r| r.settings.as_ref())
-            .and_then(|s| s.goose_provider.clone());
-
-        let recipe_model = session
-            .recipe
-            .as_ref()
-            .and_then(|r| r.settings.as_ref())
-            .and_then(|s| s.goose_model.clone());
-
         let provider_name = session
             .provider_name
             .clone()
-            .or(recipe_provider)
             .or_else(|| config.get_goose_provider().ok())
             .ok_or_else(|| anyhow!("Could not configure agent: missing provider"))?;
 
         let model_config = match session.model_config.clone() {
             Some(saved_config) => saved_config,
             None => {
-                let model_name = recipe_model
-                    .or_else(|| config.get_goose_model().ok())
+                let model_name = config
+                    .get_goose_model()
+                    .ok()
                     .ok_or_else(|| anyhow!("Could not configure agent: missing model"))?;
                 crate::model::ModelConfig::new(&model_name)
                     .map_err(|e| anyhow!("Could not configure agent: invalid model {}", e))?
--- a/crates/goose-server/src/routes/agent.rs
+++ b/crates/goose-server/src/routes/agent.rs
@@ -254,18 +254,27 @@ async fn start_agent(
     }
 
     if let Some(recipe) = original_recipe {
-        manager
-            .update(&session.id)
-            .recipe(Some(recipe))
-            .apply()
-            .await
-            .map_err(|err| {
-                error!("Failed to update session with recipe: {}", err);
-                ErrorResponse {
-                    message: format!("Failed to update session with recipe: {}", err),
-                    status: StatusCode::INTERNAL_SERVER_ERROR,
+        let mut update = manager.update(&session.id).recipe(Some(recipe.clone()));
+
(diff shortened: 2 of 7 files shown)


zanesq:
@Abhijay007 these copilot comments look legit can you take a look? Also maybe this should be rebased/pointed at `Abhijay007:feat/editModelGUI` since they should probably merge to main at the same time?

zanesq:
I tested this and I don't think its actually using the selected model once the recipe is loaded, it shows it at the bottom as selected but its still querying the global model in the chat.

zanesq:
@Abhijay007 I realized that provider/model is already specified in the session there's just something missing, goose says:

Now I see the issue clearly. Looking at `restore_provider_from_session`:

```rust
pub async fn restore_provider_from_session(&self, session: &Session) -> Result<()> {

    let provider_name = session
        .provider_name
        .clone()
        .or_else(|| config.get_goose_provider().ok())  // Falls back to global config
        .ok_or_else(|| anyhow!("Could not configure agent: missing provider"))?;

    let model_config = match session.model_config.clone() {
        Some(saved_config) => saved_config,
        None => {
            // Falls back to global config
            let model_name = config
                .get_goose_model()
                .map_err(|_| anyhow!("Could not configure agent: missing model"))?;
            crate::model::ModelConfig::new(&model_name)
                .map_err(|e| anyhow!("Could not configure agent: invalid model {}", e))?
        }
    };
    // ...
}
```

The problem is that **when a session is created with a recipe**, the recipe's `settings.goose_provider` and `settings.goose_model` are **never saved to the session's `provider_name` and `model_config` fields**.

Looking at `start_agent`:
1. Session is created
2. Extensions are resolved from recipe and saved to session
3. Recipe is saved to session
4. But **provider/model from recipe settings are NOT saved to session**

Then when `resume_agent` calls `restore_provider_from_session`:
- `session.provider_name` is `None` (never set)
- `session.model_config` is `None` (never set)
- So it falls back to global config, ignoring the recipe's settings

## The Fix

In `start_agent`, after saving the recipe to the session, we need to also save the provider/model from the recipe settings:

```rust
// In start_agent, after saving the recipe:
if let Some(ref recipe) = original_recipe {
    if let Some(ref settings) = recipe.settings {
        // Save recipe's provider/model to session so restore_provider_from_session uses them
        let mut update = manager.update(&session.id);
        
        if let Some(ref provider) = settings.goose_provider {
            update = update.provider_name(provider);
        }
        
        if let Some(ref model) = settings.goose_model {
            if let Ok(model_config) = ModelConfig::new(model) {
                update = update.model_config(model_config);
            }
        }
        
        update.apply().await.map_err(|err| {
            error!("Failed to update session with recipe provider/model: {}", err);
            ErrorResponse {
                message: format!("Failed to update session with recipe provider/model: {}", err),
                status: StatusCode::INTERNAL_SERVER_ERROR,
            }
        })?;
    }
}
```

This way, when `restore_provider_from_session` is called, it will find the recipe's provider/model in the session and use them instead of falling back to global config.

zanesq:
@Abhijay007 I found the issue and pushed a fix, test it out and lmk what you think when you get a chance.

**agent.rs** - restore_provider_from_session now checks recipe.settings as a fallback between session-level overrides and global config, so the agent actually uses the recipe's model.

**ModelsBottomBar.tsx** - Tracks userChangedModel flag. When false, displays recipe model/provider. When true (after user submits a new model), displays the user's choice. Cancel doesn't flip the flag.

**SwitchModelModal.tsx** - New initialModel prop so the modal pre-selects the recipe's model when opened during a recipe session.

crates/goose/src/agents/agent.rs
```diff
@@ -1633,18 +1633,31 @@ impl Agent {
...
+            .recipe
+            .as_ref()
+            .and_then(|r| r.settings.as_ref())
+            .and_then(|s| s.goose_provider.clone());
+
+        let recipe_model = session
+            .recipe
+            .as_ref()
+            .and_then(|r| r.settings.as_ref())
+            .and_then(|s| s.goose_model.clone());
```
DOsinga: we're looking up the settings twice here - could do it once

zanesq:
Just pushed a better fix for the model/provider loading and bottom bar display syncing

Abhijay007:
> Just pushed a better fix for the model/provider loading and bottom bar display syncing

checking it now

Abhijay007:
Tested LGTM

PR #6895, opened 2026-02-02
feat: WebSocket transport for goose-acp

jh-block:
Adds WebSocket transport alongside the existing streamable HTTP transport for ACP, providing an alternative simpler implementation for bidirectional communication.

## Protocol Details

**Connection Model**: A WebSocket connection maps one-to-one with an ACP session.

**Message Format**: JSON-RPC 2.0 (same as the streamable HTTP transport) over WebSocket text frames. All message types (requests, responses, notifications) share the same connection with native bidirectional support.

**Lifecycle**:
1. WebSocket upgrade to `ws://host:port/acp`. Session ID is returned in an `Acp-Session-Id` header in the same way as for streamable HTTP.
2. Bidirectional messaging: same protocol as used for streamable HTTP, but all over the single connection
3. Graceful or abnormal WS connection close with cleanup

## Implementation Choices

**Shared Components**: Extracted `ReceiverToAsyncRead` and `SenderToAsyncWrite` adapters into `adapters.rs`, enabling both HTTP and WebSocket transports to reuse the same `serve()` protocol handler with consistent framing.

**Architecture**:
- `TransportSession` manages per-connection state for both http and ws transports: bidirectional mpsc channels and agent task handle
- A `tokio::select!` loop concurrently routes messages: client→agent and agent→client
- Uses axum's built-in WebSocket handling with automatic ping/pong support

**Server Integration**: Both transports are served by the same HTTP server via merged Axum routers, allowing clients to choose their preferred transport at will.

### DIFF
--- a/crates/goose-acp/src/adapters.rs
+++ b/crates/goose-acp/src/adapters.rs
@@ -1,0 +1,123 @@
+//! Shared adapter classes for converting mpsc channels to AsyncRead/AsyncWrite streams
+//! Used by both HTTP and WebSocket transports
+
+use std::{
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tokio::sync::mpsc;
+use tracing::error;
+
+/// Converts an mpsc::Receiver<String> to AsyncRead
+/// Each message is terminated with a newline for JSON-RPC framing
+pub(crate) struct ReceiverToAsyncRead {
+    rx: mpsc::Receiver<String>,
+    buffer: Vec<u8>,
+    pos: usize,
+}
+
+impl ReceiverToAsyncRead {
+    pub(crate) fn new(rx: mpsc::Receiver<String>) -> Self {
+        Self {
+            rx,
+            buffer: Vec::new(),
+            pos: 0,
+        }
+    }
+}
+
+impl tokio::io::AsyncRead for ReceiverToAsyncRead {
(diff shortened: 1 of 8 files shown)


crates/goose-acp/src/adapters.rs
```diff
@@ -0,0 +1,123 @@
...
+
+impl tokio::io::AsyncWrite for SenderToAsyncWrite {
+    fn poll_write(
+        mut self: Pin<&mut Self>,
+        _cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<std::io::Result<usize>> {
+        self.buffer.extend_from_slice(buf);
+
+        while let Some(pos) = self.buffer.iter().position(|&b| b == b'\n') {
```
alexhancock, line 88: is there any way this could catch `\n` in content?

jh-block: (this isn't new code, just extracted out of the http transport impl)

If the protocol client is well-behaved, the JSON-RPC messages should not contain any actual newline characters, and they're not valid within JSON strings (a newline within a string would be encoded as two characters: backslash followed by n).

If the protocol client somehow manages to fail at that, then we'd end up trying to parse an incomplete JSON documents and error out, which I think is fine

PR #6900, opened 2026-02-02
Make ctrl-c  work like in claude code

DOsinga:
## Summary

This requires double ctrl-c to abort the CLI and displays that, bringing it in line with how claude code works


Cargo.toml
```diff
@@ -32,7 +32,7 @@ wiremock = "0.6"
 serial_test = "3.2.0"
 test-case = "3.3.1"
 base64 = "0.22.1"
-reqwest = { version = "0.12.28", default-features = false }
+reqwest = { version = "0.12.28", default-features = false, features = ["multipart"] }
```
jh-block, line 35: 🤨 seems unrelated to the rest?

DOsinga: hmm yeah, will revert

PR #6901, opened 2026-02-02
fix: make apps work in built copies of goose

alexhancock:
The old code in the `launch-app` IPC handler tried to construct the standalone app URL by extracting the origin from the current window's URL:

```typescript
const currentUrl = launchingWindow.webContents.getURL();
const baseUrl = new URL(currentUrl).origin;
const standaloneUrl = `${baseUrl}/#/standalone-app?...`;
```

- In **development** (`just run-ui`), the main window URL is `http://localhost:5173/...`, so `origin` returns `http://localhost:5173`
- In **production** (built app), the main window loads from a `file://` URL like `file:///C:/path/to/app/index.html#/...`. For `file://` URLs, `new URL(...).origin` returns the string `"null"`

This caused the standalone app window to try loading a URL like `null/#/standalone-app?...`. The fix constructs the URL the same way the main window URL is constructed elsewhere in the codebase:

```typescript
const url = MAIN_WINDOW_VITE_DEV_SERVER_URL
  ? new URL(MAIN_WINDOW_VITE_DEV_SERVER_URL)
  : pathToFileURL(path.join(__dirname, `../renderer/${MAIN_WINDOW_VITE_NAME}/index.html`));
```

- In **dev mode**: Uses the Vite dev server URL (`http://localhost:5173`)
- In **production**: Uses `pathToFileURL()` to properly construct a `file://` URL pointing to the bundled HTML file

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -394,6 +394,12 @@ async function handleFileOpen(filePath: string) {
 declare var MAIN_WINDOW_VITE_DEV_SERVER_URL: string;
 declare var MAIN_WINDOW_VITE_NAME: string;
 
+function getAppUrl(): URL {
+  return MAIN_WINDOW_VITE_DEV_SERVER_URL
+    ? new URL(MAIN_WINDOW_VITE_DEV_SERVER_URL)
+    : pathToFileURL(path.join(__dirname, `../renderer/${MAIN_WINDOW_VITE_NAME}/index.html`));
+}
+
 // Parse command line arguments
 const parseArgs = () => {
   let dirPath = null;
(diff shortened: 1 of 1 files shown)


ui/desktop/src/main.ts
```diff
@@ -2269,14 +2266,21 @@ async function appMain() {
...
-        `${baseUrl}/#/standalone-app?` +
-        `resourceUri=${encodeURIComponent(gooseApp.uri)}` +
-        `&extensionName=${encodeURIComponent(extensionName)}` +
-        `&appName=${encodeURIComponent(gooseApp.name)}` +
-        `&workingDir=${encodeURIComponent(workingDir)}`;
-
-      await appWindow.loadURL(standaloneUrl);
+
+      // Construct URL the same way as main window - using dev server in dev mode,
+      // or file:// URL in production. This fixes app loading on Windows builds.
```
jamadeo: I'd delete the comment about "fixing", since this is how it should have always been really

DOsinga: drop the comments?  I don't know how this is related to Windows builds

ui/desktop/src/main.ts
```diff
@@ -2269,14 +2266,21 @@ async function appMain() {
...
+      const url = MAIN_WINDOW_VITE_DEV_SERVER_URL
```
DOsinga: this is literalyy the same as for the main window indeed, can we not just reuse that?

PR #6907, opened 2026-02-03
fix deep links not working in markdown

zanesq:
## Summary
Clicking a deeplink in markdown was opening a blank window and not working

ReactMarkdown's default `urlTransform` only allows `http`, `https`, `mailto`, `irc`, `ircs`, and `xmpp` protocols.
All other protocols (like `spotify:`, `vscode:`, `slack:`) were being stripped to empty strings.

Also added security checks now that we are overwriting the default rendering for the links. Known safe protocols are rendered and allowed. Unknown protocols pop a warning confirmation first. Risky ones are not rendered and blocked.

Verified working now with deeplinks in markdown and regular links as before.

Screenshot 2026-02-02 at 3 10 13 PM

### DIFF
--- a/ui/desktop/src/utils/urlSecurity.ts
+++ b/ui/desktop/src/utils/urlSecurity.ts
@@ -17,24 +17,20 @@ export const BLOCKED_PROTOCOLS = [
 
 // Protocols that are safe to open without confirmation
 export const SAFE_PROTOCOLS = [
-  // Web
   'http:',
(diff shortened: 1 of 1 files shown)


zanesq:
I just asked goose for the most common deeplinks, obviously cant get them all but its a good start I think. Can always add more later if needed and if not recognized it will prompt the user to open it.

PR #6912, opened 2026-02-03
fix(acp): fixtures now raise content mismatch errors

codefromthecrypt:
## Summary

Fix ACP test fixture to show clear error messages on content mismatch. Before, it returned empty string which defeated the logic involved.

### Type of Change
- [x] Bug fix
- [x] Tests

### AI Assistance

### Testing
Introduced a typo and fixed the before and after issue

### Related Issues
Relates to #6765

### Diff

I introduced a typo (space) in the request match expression to trigger a failure.

Before
```
  json atoms at path "(root)" are not equal:                                                                                                                                           
      lhs:                                                                                                                                                                             
          ""                                                                                                                                                                           
      rhs:                                                                                                                                                                             
          {                                                                                                                                                                            
            "messages": [                                                                                                                                                              
              ...                                                                                                                                                                      
```
                                                                                                                                                                             
After:
```
  Expected body to contain:                                                                                                                                                            
    lookup[\"get_ code\"]({}): string - Get the code                                                                                                                                   
                                                                                                                                                                                       
  Actual body:                                                                                                                                                                         
    {"messages":[{"content":"You are a general-purpose AI agent called goose...     
```

### DIFF
--- a/crates/goose-acp/tests/fixtures/mod.rs
+++ b/crates/goose-acp/tests/fixtures/mod.rs
@@ -1,7 +1,6 @@
 #![recursion_limit = "256"]
 #![allow(unused_attributes)]
 
-use assert_json_diff::{assert_json_matches_no_panic, CompareMode, Config};
 use async_trait::async_trait;
 use fs_err as fs;
 use goose::builtin_extension::register_builtin_extensions;
@@ -153,8 +152,9 @@ impl OpenAiFixture {
                 let queue = queue.clone();
                 let expected_session_id = expected_session_id.clone();
                 move |req: &wiremock::Request| {
-                    let body = String::from_utf8_lossy(&req.body);
+                    let body = std::str::from_utf8(&req.body).unwrap_or("");
 
+                    // Validate session ID header
                     let actual = req
                         .headers
                         .get(SESSION_ID_HEADER)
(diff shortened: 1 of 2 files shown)


crates/goose-acp/tests/fixtures/mod.rs
```diff
@@ -174,28 +174,30 @@ impl OpenAiFixture {
...
-                    let diff =
-                        assert_json_matches_no_panic(&exp, &act, Config::new(CompareMode::Strict))
-                            .unwrap_err();
+                    drop(q);
+
+                    // If there was no body, the request was unexpected. Otherwise, it is a mismatch.
+                    let message = if expected_body.is_empty() {
+                        format!("Unexpected request:\n  {}", body)
+                    } else {
+                        format!(
```
katzdave, line 192: Did we lose something here with actual diff of the bodies together? Seems like we're only checking against empty.

codefromthecrypt: yeah the logic was wrong before as it was a substring search.  So, what was happening is we were doing a diff of a substring vs an entire json body.

so it was doing a json diff of `format!(r#"</info-msg>\n{prompt}""#)` vs the actual json body.

This was a relic of an old version which did complete request matching, and is a bug in the current design. that's why json diff isn't useful here.

Make sense?

katzdave: Got it, thanks for clarifying!

katzdave (approved):

crates/goose-acp/tests/fixtures/mod.rs
codefromthecrypt: ps #6969

Issue #6909, opened 2026-02-03
Bug: Working directory shared across sessions via process-wide env var


blackgirlbytes:
## Description

When multiple chat sessions are active, changing the working directory in one session (or the Home screen) affects all other sessions. This is because the developer extension reads its working directory from a **process-wide environment variable** (`GOOSE_WORKING_DIR`), creating two sources of truth.

## Steps to Reproduce

1. Open Goose Desktop and start a chat session in `/project-a`
2. Navigate to the Home screen
3. Change the directory to `/project-b` (or start a new session there)
4. Return to the original chat session in `/project-a`
5. Run a shell command - it will execute in `/project-b` instead of `/project-a`

## Expected Behavior

Each chat session should maintain its own working directory, independent of other sessions.

## Actual Behavior

Changing the working directory anywhere (Home screen or another session) overwrites the `GOOSE_WORKING_DIR` environment variable for the entire `goosed` process, affecting all active sessions.

## Root Cause

The developer extension reads its working directory from a global env var:

```rust
// In crates/goose-mcp/src/developer/rmcp_developer.rs
let working_dir = std::env::var("GOOSE_WORKING_DIR")
    .ok()
    .map(std::path::PathBuf::from);
```

And when sessions start, they set this process-wide:

```rust
// In crates/goose/src/agents/extension_manager.rs
std::env::set_var("GOOSE_WORKING_DIR", &effective_working_dir);
```

Since `std::env::set_var` affects the entire process, all sessions end up sharing the same working directory.

## Suggested Fix

The developer extension should receive its working directory through its initialization context or store it per-instance, rather than reading from a global environment variable.

## Environment

- Version: v1.22.2
- OS: Ubuntu (reported), likely affects all platforms

## Related

- This is **different from** #6612, which fixed extensions not receiving the session's working directory at initialization
- #6789 fixed preserving working directory when creating new chats, but doesn't address the shared state issue

## Reporter

Originally reported by a community member on the forum.

zanesq:
Got a fix in the works!

PR #6920, opened 2026-02-03
fix: per session working dir isolation

zanesq:
## Problem
Working directory was not isolated per chat session. Changing the working directory in one
session would affect other sessions because:

1. **Frontend**: A module-level global `currentWorkingDir` in `workingDir.ts` was shared
   across all sessions
2. **Backend**: `std::env::set_var("GOOSE_WORKING_DIR", ...)` set a process-global
   environment variable that all sessions read from

## Solution
Pass the working directory per-request through MCP metadata instead of using global state.

### Frontend Changes
- **`ui/desktop/src/utils/workingDir.ts`**: Removed `currentWorkingDir` global and
  `setCurrentWorkingDir()` function
- **`ui/desktop/src/components/bottom_menu/DirSwitcher.tsx`**: Removed call to
  `setCurrentWorkingDir()`

### Backend Changes

**Core plumbing:**
- **`session_context.rs`**: Added `WORKING_DIR_HEADER` constant for MCP metadata key
- **`mcp_client.rs`**: 
  - Added `working_dir` parameter to `McpClientTrait::call_tool()`
  - Created `inject_session_context_into_extensions()` and `inject_session_context_into_request()`
    to inject both session_id and working_dir into MCP request metadata
  - Updated `send_request_with_context()` to handle working directory injection

**Extension manager:**
- **`extension_manager.rs`**: 
  - Added `working_dir: Option<&Path>` parameter to `dispatch_tool_call()`
  - Removed `std::env::set_var("GOOSE_WORKING_DIR", ...)` for builtin extensions
  - Passes working_dir through to `McpClient::call_tool()`

**Developer extension:**
- **`rmcp_developer.rs`**:
  - Added `extract_working_dir_from_meta()` helper to read working_dir from MCP context
  - Modified `shell()` to extract working_dir from request metadata
  - Modified `execute_shell_command()` to accept `working_dir: Option<PathBuf>` parameter
    instead of reading from environment variable

**Call sites updated:**
- `agent.rs` - passes `Some(session.working_dir.as_path())`
- `goose-server/routes/agent.rs` - passes `None` (direct API calls)
- `code_execution_extension.rs` - passes `None`
- All `McpClientTrait` implementations updated with `_working_dir` parameter
- All test mocks updated

## Result
Each session's working directory is now passed through the request chain and isolated
from other sessions. No more global state pollution.

fixes https://github.com/block/goose/issues/6909


crates/goose/src/agents/agent.rs
```diff
@@ -581,6 +581,7 @@ impl Agent {
                 .dispatch_tool_call(
                     &session.id,
                     tool_call.clone(),
+                    Some(session.working_dir.as_path()),
```
jamadeo, line 584: maybe dispatch_tool_call should just take a reference to the session itself? now that we're passing two fields from it

zanesq: I looked into this and its a pretty large refactor because not all call sites have access to a full session and it would add additional overhead to create a session and/or mock a fake one etc. Gonna punt on this for now and we can come back to it later..

jamadeo: that makes sense - thanks for looking into it

DOsinga:
this looks a lot nicer than what we have now, thanks for doing this.

but a propos now, shouldn't this be matched by the deletion of a bunch of corresponding lines?

pub async fn add_extension( still takes a working_dir for example

zanesq:
The `working_dir` parameter in `add_extension` is used for setting the working directory of child processes `effective_working_dir ` so its still needed there but agreed it could be better organized by computing effective_working_dir lazily only where needed instead. I'll do a follow up PR 👍

DOsinga:
thanks @zanesq - you are right that's still needed I guess

zanesq:
np here is the follow up just making it a bit clearer for now why we do this https://github.com/block/goose/pull/6958

PR #6933, opened 2026-02-03
feat: add local inference provider with llama.cpp backend and HuggingFace model management

DOsinga:
## Summary

Adds a local inference provider that enables running language models directly on-device using llama.cpp, with full integration across CLI, server, and desktop UI.

### Core Changes

- **Local inference provider** (`crates/goose/src/providers/local_inference/`): New provider with modular architecture split into submodules:
  - `inference_native_tools`: Inference path for models with native tool-calling support
  - `inference_emulated_tools`: Inference path for models without native tool support, using text-based `$ command` and ````execute` block detection
  - `inference_engine`: Shared inference primitives — context management, sampling, and token generation
  - `tool_parsing`: Parsing tool calls from model output (JSON and XML formats)
  - `hf_models`: HuggingFace Hub model search and GGUF file discovery
  - `local_model_registry`: Local model download and lifecycle management

- **Server routes** (`crates/goose-server/src/routes/local_inference.rs`): REST endpoints for model download, listing, deletion, and status

- **CLI integration** (`crates/goose-cli/`): New `local-models` subcommand for managing downloaded models, plus provider selection support

- **Desktop UI** (`ui/desktop/src/components/`):
  - `LocalModelSetup.tsx`: First-run setup flow for downloading models
  - `LocalInferenceSettings.tsx`: Settings panel for model management
  - `HuggingFaceModelSearch.tsx`: Search and download models from HuggingFace
  - `ModelSettingsPanel.tsx`: Per-model configuration

### Additional Changes

- Tiny model system prompt for resource-constrained local models
- Session name generation support for local inference conversations
- Download progress display improvements

### Known Limitations (deferred to future iterations)

- **Memory management is best-effort**: `estimate_max_context_for_memory` uses 50% of available memory for KV cache, which is a rough heuristic. `available_inference_memory_bytes` picks the max of any single accelerator device, which may not be correct for multi-GPU setups or unified memory architectures. On unified-memory Macs, free memory fluctuates with system load, so a model that fits during estimation may OOM during inference. There is currently no fallback or recovery for OOM — it will crash the process.

- **Model unloading strategy is naive**: When loading a new model, all other models are unloaded. This is fine for single-model usage but the code structure (`HashMap` of model slots) suggests multi-model was considered. A smarter eviction strategy (e.g. LRU) could be added if multi-model support is needed.

- **Featured models list is hardcoded**: The `FEATURED_MODELS` constant in `local_model_registry.rs` is a static list of HuggingFace model specs. There is no mechanism to update it without a code change — it will go stale as newer/better models are released. A future iteration could fetch this list from a remote config or allow user customization.

### DIFF
--- a/crates/goose-acp/tests/common_tests/mod.rs
+++ b/crates/goose-acp/tests/common_tests/mod.rs
@@ -171,7 +171,6 @@ pub async fn run_model_list<C: Connection>() {
             "gpt-4o-2024-05-13",
             "gpt-4o-2024-08-06",
             "gpt-4o-2024-11-20",
-
             "gpt-4",
             "gpt-4-0613",
             "gpt-4-turbo",


crates/goose-server/src/routes/dictation.rs
```diff
@@ -262,6 +262,8 @@ pub async fn download_model(Path(model_id): Path<String>) -> Result<StatusCode,
             model.id.to_string(),
             model.url.to_string(),
             model.local_path(),
+            Some(whisper::LOCAL_WHISPER_MODEL_CONFIG_KEY.to_string()),
+            Some(model.id.to_string()),
```
jamadeo: It seems a bit out of place to put the config change in the download manager--think this is the only place it's used? The caller could just set the config value

crates/goose/src/agents/extension_manager.rs
```diff
@@ -236,7 +236,18 @@ async fn child_process_client(
         command.env("PATH", path);
     }
 
-    if let Some(dir) = working_dir {
+    // Set GOOSE_CONTEXT_SIZE env var for the child process from provider's model config
+    if let Some(provider_arc) = provider.lock().await.as_ref() {
+        let context_limit = provider_arc.get_model_config().context_limit();
+        command.env("GOOSE_CONTEXT_SIZE", context_limit.to_string());
```
jamadeo: What is this used for?

crates/goose/src/agents/extension_manager.rs
```diff
@@ -1509,6 +1521,14 @@ impl ExtensionManager {
         session_id: &str,
         working_dir: &std::path::Path,
     ) -> Option<String> {
+        if let Ok(provider_guard) = self.provider.try_lock() {
+            if let Some(provider) = provider_guard.as_ref() {
+                if provider.get_model_config().context_limit() < 9 * 1024 * 1024 {
```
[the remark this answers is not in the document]
jamadeo: Is the intention here to omit this content for models with small context windows?

crates/goose/src/agents/prompt_manager.rs
```diff
@@ -42,6 +42,7 @@ struct SystemPromptContext {
     max_extensions: usize,
     max_tools: usize,
     code_execution_mode: bool,
+    small_model: bool,
```
jamadeo: I might be missing it but I don't think this is being used?

crates/goose/src/providers/local_inference/inference_emulated_tools.rs
```diff
@@ -0,0 +1,681 @@
...
+use super::inference_engine::{
+    create_and_prefill_context, generation_loop, validate_and_compute_context, GenerationContext,
+    TokenAction,
+};
+use super::{finalize_usage, StreamSender, CODE_EXECUTION_TOOL, SHELL_TOOL};
+
+/// Bytes to hold back from streaming in code mode: length of `` ```execute\n ``
+/// plus the preceding `\n`, so the parser doesn't emit text that turns out to be
+/// the start of an execute fence.
+const HOLD_BACK_CODE_MODE: usize = 12;
```
jamadeo: could do something like `const HOLD_BACK_CODE_MODE: usize = " ```execute\n".len();`, likewise for below

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
aharvard, line 1: Were these changes to the MCP App renderer intentional? This PR removed @alexhancock work from merged in here https://github.com/block/goose/pull/7039

aharvard: fixed via https://github.com/block/goose/pull/7366

PR #6934, opened 2026-02-04
Simplified custom model flow with canonical models

katzdave:
## Summary
Use canonical models to provide a new flow for initializing custom models that are part of our models database.

- Remove ALLOWED_PROVIDERS for canonical models; let's use them for everything (necessary to pickup all possible custom providers)
- Introduce provider metadata catalog which lists providers + compatibility.
- Serve the catalog to the client
- Add a simplified flow to pre-create most fields for these entries
- Attach non-null catalog_id. Will follow up with some kind of auto-update models flow (since those are hard-coded)

https://github.com/user-attachments/assets/b234ad59-aba4-40f5-8117-fe7c03a6a2ce


lifeizhou-ap (approved):
Great! Now it is easy to add new providers!

Issue #6894, opened 2026-02-02
ACP integration with Zed cannot modify files when following the documentation


maikeriva:
**Describe the bug**

Goose ACP integration with the Zed IDE is unable to interact with files. The other ACP providers in Zed seem to work fine. The integration might be broken or the documentation might be lacking to illustrate some necessary steps to provide goose the correct information to leverage the functionality

---

**To Reproduce**
1. Install Goose CLI and configure it with a provider (for example mistral-medium-latest)
2. Install [Zed](https://zed.dev/download)
3. Follow instructions at https://block.github.io/goose/docs/guides/acp-clients/#example-zed-editor-setup
4. In Zed's agent panel open a new thread with goose's external agent
5. Ask it to edit any file, goose seems unable to do so

---

**Expected behavior**
Goose's ACP integration with Zed should work like other ACP integrations

---

**Please provide the following information**
- **OS & Arch:** Fedora 43 (amd64)
- **Interface:** CLI
- **Version:** v1.22.1
- **Extensions enabled:** Default-provided
- **Provider & Model:** mistral-medium-latest (any should exhibit the same issue)

rabi:
I think https://github.com/block/goose/commit/b0c6373cf88b62786acf6850019aaff21508e6b4 has not been released yet. You can build goose from main branch and use it or try enabling developer extension.

```
{
  "agent_servers": {
    "goose": {
      "type": "custom",
      "command": "goose",
      "args": ["acp", "--with-builtin", "developer"],
      "env": {},
    },
  },
}

```

maikeriva:
Thanks! After altering the config it seems to sort of work, but it still lacks access (or perhaps understanding?) of Zed's overall capabilities. For example, it doesn't show me a preview of the changes in the IDE like other external agents do, but rather only prompt to allow/deny edits. Is this a limitation on goose or Zed side?

rabi:
I think plumbing for buffered preview before applying changes does not exist yet in goose/goose-acp. Could be an RFE.

maikeriva:
Thanks, now I have a better understanding of it. I think it would be great to have, and one of the last pieces of the puzzle towards a very well integrated system-wide LLM assistant in Linux.

rabi:
I've proposed https://github.com/block/goose/pull/6940. You can give it a shot and provide feedback. It works well for me.

maikeriva:
[...]

PR #6940, opened 2026-02-04
feat(acp): delegate developer file I/O through ACP fs methods

rabi:
## Summary
When an ACP client (e.g. Zed) advertises fs.writeTextFile and fs.readTextFile capabilities, the developer extension routes file reads/writes through the client instead of hitting disk directly. This lets the editor track modifications and return unsaved buffer contents.

### Type of Change
- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested locally with zed.

### Related Issues
Relates to #6894

After:   
image

### DIFF
--- a/crates/goose-acp/src/server.rs
+++ b/crates/goose-acp/src/server.rs
@@ -2,6 +2,7 @@ use crate::custom_requests::*;
 use anyhow::Result;
 use fs_err as fs;
 use goose::agents::extension::{Envs, PLATFORM_EXTENSIONS};
+use goose::agents::platform_extensions::DeveloperFileIo;
 use goose::agents::{Agent, AgentConfig, ExtensionConfig, GoosePlatform, SessionConfig};
 use goose::builtin_extension::register_builtin_extensions;
 use goose::config::base::CONFIG_YAML_NAME;
@@ -22,21 +23,22 @@ use goose_acp_macros::custom_methods;
 use rmcp::model::{CallToolResult, RawContent, ResourceContents, Role};
 use sacp::schema::{
     AgentCapabilities, AuthMethod, AuthenticateRequest, AuthenticateResponse, BlobResourceContents,
-    CancelNotification, Content, ContentBlock, ContentChunk, EmbeddedResource,
+    CancelNotification, ClientCapabilities, Content, ContentBlock, ContentChunk, EmbeddedResource,
     EmbeddedResourceResource, ImageContent, InitializeRequest, InitializeResponse,
     ListSessionsResponse, LoadSessionRequest, LoadSessionResponse, McpCapabilities, McpServer,
     ModelId, ModelInfo, NewSessionRequest, NewSessionResponse, PermissionOption,
-    PermissionOptionKind, PromptCapabilities, PromptRequest, PromptResponse,
-    RequestPermissionOutcome, RequestPermissionRequest, ResourceLink, SessionCapabilities,
-    SessionId, SessionInfo, SessionListCapabilities, SessionModelState, SessionNotification,
-    SessionUpdate, SetSessionModelRequest, SetSessionModelResponse, StopReason, TextContent,
-    TextResourceContents, ToolCall, ToolCallContent, ToolCallId, ToolCallLocation, ToolCallStatus,
-    ToolCallUpdate, ToolCallUpdateFields, ToolKind,
+    PermissionOptionKind, PromptCapabilities, PromptRequest, PromptResponse, ReadTextFileRequest,
+    ReadTextFileResponse, RequestPermissionOutcome, RequestPermissionRequest, ResourceLink,
+    SessionCapabilities, SessionId, SessionInfo, SessionListCapabilities, SessionModelState,
+    SessionNotification, SessionUpdate, SetSessionModelRequest, SetSessionModelResponse,
+    StopReason, TextContent, TextResourceContents, ToolCall, ToolCallContent, ToolCallId,
+    ToolCallLocation, ToolCallStatus, ToolCallUpdate, ToolCallUpdateFields, ToolKind,
+    WriteTextFileRequest, WriteTextFileResponse,
 };
 use sacp::{AgentToClient, ByteStreams, Handled, JrConnectionCx, JrMessageHandler, MessageCx};
 use std::collections::HashMap;
 use std::sync::Arc;
-use tokio::sync::Mutex;
+use tokio::sync::{Mutex, RwLock};
 use tokio_util::compat::{TokioAsyncReadCompatExt as _, TokioAsyncWriteCompatExt as _};
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, info, warn};
@@ -60,6 +62,7 @@ pub struct GooseAcpAgent {
     goose_mode: goose::config::GooseMode,
     disable_session_naming: bool,
     builtins: Vec<String>,
+    client_capabilities: RwLock<ClientCapabilities>,
 }
 
 fn mcp_server_to_extension_config(mcp_server: McpServer) -> Result<ExtensionConfig, String> {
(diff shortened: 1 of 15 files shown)


michaelneale:
hrm - yeah this is not a small change, although pretty cool. Main thing is this changes the developer tool to have a new mode, that is main thing that concerns, but otherwise would make editor experience nicer, wonder if there is a way to package this to not be as cross cutting, as you will rarely want "write mode" etc?

rabi:
> hrm - yeah this is not a small change, although pretty cool. Main thing is this changes the developer tool to have a new mode, that is main thing that concerns, but otherwise would make editor experience nicer, wonder if there is a way to package this to not be as cross cutting, as you will rarely want "write mode" etc?

It's based on client capability to write. If you think there could be scenarios where clients won't need it irrespective of them advertising their capabilities, we can probably make it configurable, but we would still need the new developer mode. 

```
        let client_can_write = self
            .client_capabilities
            .get()
            .is_some_and(|c| c.fs_write_text_file);
        let write_mode = if client_can_write && !has_code_execution {
            WriteMode::Deferred
        } else {
            WriteMode::Direct
        };
```

michaelneale:
@rabi yeah - I think that we really want this only available when actually used from visual editors don't you think? otherwise it is adding more work for the LLM to do? if could polish that off I think is a great idea

rabi:
> @rabi yeah - I think that we really want this only available when actually used from visual editors don't you think? otherwise it is adding more work for the LLM to do? if could polish that off I think is a great idea

OK, Thanks @michaelneale, I'll expose the feature behind  `--acp-editor` flag for clients to enable it . So it will be behind both the flag and client capability to write files. The same flag can be used for other features in the future too.

codefromthecrypt:
Hey @rabi, I filed #7451 to capture what I think the minimal approach looks like here.

The core idea is the same (route file I/O through ACP), but the implementation can be a lot simpler if we lean on what the spec already defines:

- **Reads too, not just writes.** `fs/read_text_file` means the editor can return unsaved buffer contents. This PR only handles writes.
- **No custom MIME type or `_meta.old_text`.** The spec's `fs/write_text_file` just takes `path` + `content`. The client decides how to present diffs. We don't need `application/x-goose-file-diff` or `FileDiff` structs.
- **No `--acp-editor` flag.** Capability negotiation during `initialize` is the feature gate. If the client advertises `readTextFile`/`writeTextFile`, delegate. If not, fall back to disk. No new CLI flags needed.
- **No double permission.** The existing `session/request_permission` flow handles tool approval. A second managed-write permission prompt isn't in the spec.
- **No in-process duplex pipe.** The developer extension's disk I/O functions (`File::open`, `fs::write`, `read_to_string`) just need to become pluggable so they can go through ACP instead of disk. No need to rewire how the extension connects.

The `view_range` to `line`+`limit` translation and the checklist in #7451 should give a clearer starting point. Happy to discuss any of it.

rabi:
@codefromthecrypt  Thanks for checking this and creating the issue as well.

> Reads too, not just writes. fs/read_text_file means the editor can return unsaved buffer contents. This PR only handles writes.

We can implement these incrementally right? I kept this limited to write ad read as a followup. As you can see I've added `read_text_file` in as TODO.

> No custom MIME type or _meta.old_text. The spec's fs/write_text_file just takes path + content. The client decides how to present diffs. We don't need application/x-goose-file-diff or FileDiff structs.

Yeah, that's a good point. I'll change.

> No --acp-editor flag. Capability negotiation during initialize is the feature gate. If the client advertises readTextFile/writeTextFile, delegate. If not, fall back to disk. No new CLI flags needed.

It was not in my original PR. Added after @michaelneale asked for it.

> No double permission. The existing session/request_permission flow handles tool approval. A second managed-write permission prompt isn't in the spec.

I observed cases where approval didn’t trigger as expected in my testing, so I added a temporary guard. I’ll re-verify and remove if unnecessary.

>No in-process duplex pipe. The developer extension's disk I/O functions (File::open, fs::write, read_to_string) just need to become pluggable so they can go through ACP instead of disk. No need to rewire how the extension connects.

Right, that would be larger refactor of developer extension with broader touch points I wanted to avoid, but I can include it if we want.

codefromthecrypt:
thanks for accepting the feedback so far ;) ack that there are a couple design points going on.

Here's some unsolicited feedback with TL;DR; maybe chop off a targeted smaller PR from this, with eventually landing this as the goal. The first, is probably just tuning the developer extension (builtin) so that the impl of reads are possible to adapt in with an ACP client., This is like a wiring concern and shouldn't be huge. with that in, other things are simpler to progress maybe.

---

Something that can help is that when we look at this change, probably the easiest one is the read file not the write. One of the reasons it takes me so long to get changes together is that I make a PR (such as this one) then hack it to bits until all the pieces needed are in.

This is why #6605 has taken so long, for example. For example, the infra carve-out for MCP came in this, and in some ways it is similar to the design issue you are hinting at https://github.com/block/goose/pull/6972

Basically, you have a chicken-egg, right? you are looking to have a way to control which impl is used for editing tools that Goose tells the LLM it has. So, how do you get ACP to be the "impl" of the developer plugin, especially considering the chicken-egg is you don't know if ACP will be supported by the client until *after* it connects.

So, one way to think about it is the "builtin" are already added to ACP via #6284 I mean that's how I was testing code mode, right? So, developer is a builtin.. so one way to chop this up is to choose the easiest thing, which may be read, not write, and do only that.. surgically, in a PR. It isn't required to do both read and write in the same one.. this would prove you've solved the bootstrapping concern in the right abstraction.

The other thing that might help, is while we have hints everywhere somethings are not very explicit. I force the agent to re-read this every few minutes to cut down on cruft.. 

[CODING_STYLE.md](https://github.com/user-attachments/files/25522124/CODING_STYLE.md)

rabi:
> The first, is probably just tuning the developer extension (builtin) so that the impl of reads are possible to adapt in with an ACP client

I think I've addressed all the feedback in the last update, including developer extension changes. So it should be good now.

michaelneale:
@rabi just today (I think?) there is a new developer tool, does this work with that? otherwise LGTM I think

rabi:
> @rabi just today (I think?) there is a new developer tool, does this work with that? otherwise LGTM I think

@michaelneale If you mean https://github.com/block/goose/pull/7466, does not look like it's merged yet.

crates/goose-acp/src/server.rs
```diff
@@ -600,48 +754,167 @@ fn outcome_to_confirmation(outcome: &RequestPermissionOutcome) -> PermissionConf
...
+        _ => None,
+    }
+}
+
+struct ToolCallContentWithDiffs {
+    content: Vec<ToolCallContent>,
+    diffs: Vec<ManagedFileWrite>,
+}
+
+struct ToolPermissionRequestArgs {
```
codefromthecrypt: this is a bit strange.. I think if an agent created this to avoid clippy warnings it is probably best to undo it. as fr as I can tell, we create this structure then destructure it immediately.

crates/goose-acp/src/server.rs
```diff
@@ -496,23 +598,75 @@ impl GooseAcpAgent {
             )),
         ))?;
 
+        let supports_write = self.supports_fs_write();
+        let mut apply_failed = false;
+
+        if !diffs.is_empty() {
```
codefromthecrypt: this is mixing abstraction concepts, the server code shouldn't be handling diffs, even if some agents aren't designed perfectly. I raised https://github.com/block/goose/issues/7451 to suggest designing first before implementing. In there are some hints and source bases to explore.

The issue with doing things this way is that it is tech debt that takes a lot more attention to undo vs just do cleanly in the first place.

crates/goose-acp/src/server.rs
```diff
@@ -600,48 +754,167 @@ fn outcome_to_confirmation(outcome: &RequestPermissionOutcome) -> PermissionConf
     }
 }
 
-fn build_tool_call_content(tool_result: &ToolResult<CallToolResult>) -> Vec<ToolCallContent> {
+struct ManagedFileWrite {
+    path: std::path::PathBuf,
+    new_text: String,
+}
+
+fn decode_uri_component(value: &str) -> Option<String> {
```
codefromthecrypt: server should not have the responsibility for being a URI encoder/decoder. look for existing crates or a utility, or in any case mention why custom decoding is an ACP concern required to implement this.

crates/goose-acp/src/server.rs
```diff
@@ -671,9 +955,100 @@ impl GooseAcpAgent {
...
+            .is_some_and(|c| c.fs.write_text_file)
+    }
+
+    pub fn supports_fs_read(&self) -> bool {
+        self.client_capabilities
+            .get()
+            .is_some_and(|c| c.fs.read_text_file)
+    }
+
+    fn build_io_strategy(
```
codefromthecrypt: this is a broken abstraction. The server's job is only to push the file capability to our dev extension which consumes it optionally as a trait. When that's missing, it uses the default implementation. What you can see happening here, is server inheriting a mix of concerns which ends up with having tight coupling and confusing code, where the core server handler is now also a diff machine.

rabi: Thanks @codefromthecrypt , I think most of the complexity came from bridging ACP through the old goose-mcp developer path, which mixed transport and file-diff concerns across layers.

Now that developer is a platform extension after https://github.com/block/goose/pull/7466, ACP can now inject capability-gated read/write delegates directly as you mentioned, keeping server transport-only and letting clients render diffs natively. 

I still think developer_file_io delegate on PlatformExtenstionContext would be extension specific leakage. 

I'll rebase the PR make it aligned with the new developer platform extension. If there are still concerns from your side around the design, feel free to take over.

codefromthecrypt:
Ran my acp-test suite against this branch (commit a3dda06). Write delegation now passes, nice! But read delegation still fails.

The `read_file` closure is wired all the way through: `build_developer_file_io()` creates it, `DeveloperFileIo` carries it, `EditTools` stores it. But it's only ever called inside `file_edit_with_cwd()` to read a file before applying an edit. There's no standalone `read` tool in the developer extension — the tool list is `write`, `edit`, `shell`, `tree`.

So when asked to read a file, goose falls back to `shell` (`cat`) or `computercontroller: cache` (`view`), bypassing ACP delegation entirely.

Is `read` intentionally left out? The plumbing is all there — it just needs a `read` tool in `mod.rs` + a `file_read_with_cwd` method on `EditTools` to expose it?

cc @michaelneale as I'm really very out of time and must stop for a day or two

codefromthecrypt:
Nice work on the delegation plumbing — the build_developer_file_io closures in server.rs:341-399 correctly wire up ReadTextFileRequest/WriteTextFileRequest through ACP when the client advertises capabilities.

However, during testing with Zed I noticed a gap: there is no standalone **read** tool exposed by the developer extension. The read_file closure is wired into EditTools, but it is only called internally from file_edit_with_cwd (edit.rs:78) as part of the edit flow (read current content → find/replace → write back). The 4 tools in mod.rs:102-152 are write, edit, shell, tree — no read.

So when the LLM needs to read a file, it falls back to shell (cat file), which bypasses ACP delegation entirely. This means fs/read_text_file is never actually invoked as a standalone operation — only as a side-effect of edit.

Is this intentional? Without a read tool that delegates through ACP, the client cannot return unsaved buffer contents for standalone file reads.

codefromthecrypt:
TL;DR; add a read tool to developer that uses the existing read_file closure, then make sure there's a test for it that proves dispatch works. That's the cleanest. otherwise we need to merge something incomplete then make another couple PRs to follow-up, something odd because reads are more frequent editor use than writes.

rabi:
> TL;DR; add a read tool to developer that uses the existing read_file closure, then make sure there's a test for it that proves dispatch works.

I think I had initially kept `read_text_file` capability  wiring as a future change for that same reason. Anyway, I think addition of read tool to the dev extension should be a separate PR. But I'll add it here as a separate commit if we want.

jamadeo:
Sorry I am late to get up to speed here (hadn't seen this PR til now). I definitely think goose should support these edit tools.

I implemented this in https://github.com/block/goose/pull/7388 as a separate set of tools. I think that's the way to go, but I'm open to hearing otherwise!

rabi:
@jamadeo, Hey! Thanks for jumping in. From what I've seen, existing ACP implementations like gemini-cli[1] use the same pattern i.e delegate existing file tool I/O through ACP client capabilities rather than exposing a separate set of ACP-specific tools.

A. Capability negotiation at initialize time
B. Per-capability fallback for readTextFile/writeTextFile independently
C. Session-scoped wiring during newSession/loadSession
D. Same ACP protocol methods (readTextFile, writeTextFile)

Gemini-cli uses an interface/service replacement pattern; goose uses closure injection based on earlier feedback.


@codefromthecrypt,  `developer_file_io` being `Option` threaded through the stack adds unnecessary wiring is a fair point. I can refactor `DeveloperFileIo` to always be present (defaulting to `default_local()`) rather than `Option`, which would eliminate the `None` sites and simplify the wiring. 

On the "special-casing" concern,  in the current code, `EditTools` calls `(self.read_file)(path)` and `(self.write_file)(path, content)` uniformly. There are no ACP-specific if-statements at the tool execution level; the branching happens once at construction time in `build_developer_file_io`. That said, I'm open to hearing which specific paths you see as problematic.

On test infrastructure -- could you clarify the record/replay concern? I want to make sure the tests align with the existing ACP test patterns.

I'm confused by the shift between the Feb 27 review ("looks very close... I don't mind if someone else approves merges") and now the suggestion to abandon. I want to understand what specifically changed between those reviews beyond the optionality point, so I can address it. This is a feature asked by many users and something we wanted to use internally. I'm always for constructive feedback as the intent is to make goose better and learn in the process. We won't get everything right the first time. Having said that, if we think this implementation is completely off, I'm fine to go with a better implementation you propose.

[1] https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/zed-integration/zedIntegration.ts

codefromthecrypt:
Ps I have a complete impl for this as well, just cant push it until off flight. It corrects all the things mentioned with tests for all 4 options (pass through and delegated read vs write) as well a properly annotated read tool. I have not seen the other PRs since. If the PR I made isnt helpful I will drop it after raising it. However, I think it is worth raising just so that future folks and coding agents alike have a reference.

codefromthecrypt:
https://github.com/block/goose/pull/7388 is the most elegant way out. Thanks for the work on this folks!

I will raise a follow up PR to that to complete test backfill etc.

codefromthecrypt:
https://github.com/block/goose/pull/7668 is the code I had locally as promised.. I will figure out if it is easier to rewrite it or add a commit to switch to the style of https://github.com/block/goose/pull/7388

PR #6947, opened 2026-02-04
feat(mcp-apps): add Permission Policy support for sandbox iframes

aharvard:
## Summary

MCP Apps run in sandboxed iframes for security. Some apps need access to browser capabilities like the camera, microphone, or geolocation—for example, a voice recording app or a location-based search.

This PR implements support for `UIResourceMeta.permissions`, allowing MCP App servers to declare which browser capabilities their UI needs. The host (Goose) reads these permissions and sets the appropriate `allow` attribute on the sandbox iframe, enabling the [Permissions Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Permissions_Policy) for those features.

Per the spec, hosts MAY honor these permissions, and apps SHOULD NOT assume they're granted—they should use JS feature detection as a fallback.

## Supported Permissions

| Permission | Iframe `allow` attribute |
|------------|---------------------------|
| `camera` | `camera` |
| `microphone` | `microphone` |
| `geolocation` | `geolocation` |
| `clipboardWrite` | `clipboard-write` |

## Changes

**Backend (Rust)**
- Added `PermissionsMetadata` struct with optional boolean fields for each permission
- Updated `UiMetadata` to include the new `permissions` field

**Frontend (TypeScript)**
- Pass permissions from resource metadata through the sandbox bridge
- Updated `mcp_app_proxy.html` to build the iframe `allow` attribute from requested permissions

## Example

An MCP App server can declare permissions in its resource metadata:

```json
{
  "_meta": {
    "ui": {
      "permissions": {
        "camera": true,
        "microphone": true
      }
    }
  }
}
```

Goose will then render the iframe with:
```html
<iframe allow="camera; microphone" ...>
```

## References

- [MCP Apps Spec - UIResourceMeta.permissions](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx)
- [MDN - Permissions Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Permissions_Policy)


aharvard:
There are a lot of formatting changes across the generated files. Not sure what's going on there.

crates/goose/src/goose_apps/resource.rs
```diff
@@ -14,13 +14,36 @@ pub struct CspMetadata {
...
 /// UI-specific metadata for MCP resources
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema, PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
 pub struct UiMetadata {
     /// Content Security Policy configuration
     #[serde(skip_serializing_if = "Option::is_none")]
     pub csp: Option<CspMetadata>,
+    /// Sandbox permissions requested by the UI
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub permissions: Option<PermissionsMetadata>,
```
DOsinga: can we make this not Option<> and just provide a default implementation? same as for the booleans above. if we just make them non option booelan and initialize them as false, the generated code becomes cleaner and the client easier to read

Issue #2371, opened 2025-04-27
Feature Request: Clearer UI to associate a tool call and its approval in single step with multiple tool calls


m-roberts:
**Please explain the motivation behind the feature request.**

When an extension makes several tool call requests in a single step, they all appear sequentially, followed by the tool call approval for each (also sequentially). Each tool call approval says "Goose would like to call the above tool. Allow?" but when there are several, it can be hard to quickly inspect what step you are approving. If there are enough it can even require scrolling up to find which tool is going to be invoked.

This disrupts the flow of the conversation and can lead to erroneous actions being taken by the user by accident.

**Describe the solution you'd like**

Even if an extension makes multiple tool call requests in a single "step", it would be better if the tool approval step was more clearly associated with the actual tool invocation. The exact approach to do this is certainly something that doesn't necessarily have one best way, as I can see the benefit of seeing what the follow-up tool calls are, so simply blocking them from showing up until earlier tool calls are handled might be undesirable.

One way to do this "elegantly" could be to integrate the manual tool approval UI into the tool call box itself, such that output actually appears below where the user accepted the tool call.

**Describe alternatives you've considered**

I haven't really thought too much about this, though I think it's fairly clear that the current UI, when running without auto-accept mode, is problematic.

**Additional context**

Here are some screenshots of a conversation I had which was making multiple edits to Airtable - 1 update, and multiple creates. If I wasn't diligent, I could have easily missed the fact that I was actually initially approving an update tool call, not a create.

Note that this is from session history, hence why "Tool confirmation is not available".

<img width="1800" alt="Image" src="https://github.com/user-attachments/assets/46539e0f-a183-4c09-8194-6b917480669e" />

<img width="1800" alt="Image" src="https://github.com/user-attachments/assets/ee48d09e-604e-4828-ba01-e5cce47c42c0" />

- [x] I have verified this does not duplicate an existing feature request

Abhijay007:
@taniandjerry , @DOsinga, I think this one is resolved in one of the latest releases

PR #6949, opened 2026-02-04
Ask separately for confirmation

DOsinga:
## Summary

Fixes: https://github.com/block/goose/issues/2371

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6951,9 +6951,9 @@ dependencies = [
 
 [[package]]
 name = "pctx_code_mode"
-version = "0.2.2"
+version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5788a3f1aaa7bc0fdcbae6d9ff24a080a6cc0d202d8affa8f1c29097e40a504a"
+checksum = "1006bf745381b3726592032cbee129f27873296810be199e04dfe759820e96f8"
 dependencies = [
  "futures",
  "pctx_code_execution_runtime",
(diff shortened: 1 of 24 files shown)


zanesq:
Ran it locally it works

Couple of things goose picked up that seem worth looking into

---------------

Missing deny_once handling - The Permission enum includes deny_once but the UI only offers "Deny" which maps to always_deny. Is this intentional? The old code had DENY = 'deny' which is different from always_deny.

isStreaming prop change - Changed from isStreaming?: boolean with default false to required isStreaming: boolean. Need to verify all call sites pass this prop.

findConfirmationForToolAcrossMessages defined inside render - This function is recreated on every render. Could be memoized or moved outside the component.

cancel permission - The enum includes cancel but it's not used in the UI. Is this for programmatic cancellation only?

No error handling UI - handleAction in ToolApprovalButtons calls the API but doesn't show errors to users if the call fails.

The PR removes the "Change" button that let users modify permissions after approval. Was this intentional? Users can no longer change their decision after the fact from this UI.

DOsinga:
> Missing deny_once handling - The Permission enum includes deny_once but the UI only offers "Deny" which maps to always_deny. Is this intentional? The old code had DENY = 'deny' which is different from always_deny.

that's not true. we have allow, allow once & deny (which is deny once)
 
> isStreaming prop change - Changed from isStreaming?: boolean with default false to required isStreaming: boolean. Need to verify all call sites pass this prop.

the compiler would have picked up on that.

> findConfirmationForToolAcrossMessages defined inside render - This function is recreated on every render. Could be memoized or moved outside the component.

why does it matter that it gets recreated every time

> cancel permission - The enum includes cancel but it's not used in the UI. Is this for programmatic cancellation only?

correct, but there's no change here. the enum always had cancel in there, before we were using custom strings with 
all the risks involved, now we use the enum in typescript too.

> No error handling UI - handleAction in ToolApprovalButtons calls the API but doesn't show errors to users if the call fails.

same as before. 

> The PR removes the "Change" button that let users modify permissions after approval. Was this intentional? Users can no longer change their decision after the fact from this UI.

the Change button never allowed the user to modify their permission. it was a short cut to settings. just like this confused you, I think it confuses all users, so yeah removed

Issue #6954, opened 2026-02-04
"Edit In Place" feature only appends instead of edits


bzqzheng:
Expected Behavior
When editing a message in place, the existing message and its subsequent answer should be modified/repliaced.

Actual behavior
Instead of modifying the existing conversation, Goose appends the newly edited query to the bottom of the chat. The original query and its output remain in the conversaion, creating duplicate/conflicting context.
If previous query contains an extension operation (i.e. search on Slack), this action will be duplicated and executed again.

version - `1.1.22.2`

bzqzheng:
Found the root cause - code [ref](https://github.com/block/goose/blob/6ae72a6cf0408d0ab90c54935d9f1fc0c82a88b3/ui/desktop/src/hooks/useChatStream.ts#L753)

```
// When editType is 'edit' (Edit in Place)
  const response = await forkSession({
    path: { session_id: sessionId },
    body: {
      timestamp: message.created,
      truncate: true,
      copy: editType === 'fork',  // false when editType is 'edit'
    },
    throwOnError: true,
  });

  const targetSessionId = response.data?.sessionId;

  if (editType === 'fork') {
    // Forking logic here
  } else {
    // Edit in place logic - THIS IS THE ISSUE
    const { getSession } = await import('../api');
    const sessionResponse = await getSession({
      path: { session_id: targetSessionId },
      throwOnError: true,
    });

    if (sessionResponse.data?.conversation) {
      dispatch({ type: 'SET_MESSAGES', payload: sessionResponse.data.conversation });
    }
    // This is the problem - it's submitting the edited content as a new message
    await handleSubmit({ msg: newContent, images: [] });
  }
```

PR #6955, opened 2026-02-04
Fix 'Edit In Place' feature appending instead of editing messages

bzqzheng:
## Summary
- Fixes issue #6954 where "Edit In Place" feature appends the edited content as a new message instead of replacing the existing message
- Modifies the `onMessageUpdate` function to properly update the content of the edited message instead of creating a new one
- Preserves existing image content if present in the message being edited

## Test plan
1. Open an existing conversation with multiple messages
2. Hover over a user message and click the "Edit" button
3. Modify the message content
4. Click "Edit in Place"
5. Verify that the edited content replaces the original message and doesn't appear as a new message
6. Verify that messages after the edited message are removed as expected

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/ui/desktop/src/hooks/useChatStream.ts
+++ b/ui/desktop/src/hooks/useChatStream.ts
@@ -705,6 +705,8 @@ export function useChatStream({
     async (messageId: string, newContent: string, editType: 'fork' | 'edit' = 'fork') => {
       const currentState = stateRef.current;
 
+      dispatch({ type: 'SET_CHAT_STATE', payload: ChatState.Thinking });
+
       try {
         const { forkSession } = await import('../api');
         const message = currentState.messages.find((m) => m.id === messageId);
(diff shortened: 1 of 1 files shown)


ui/desktop/src/hooks/useChatStream.ts
```diff
@@ -748,9 +748,63 @@ export function useChatStream({
           });
 
           if (sessionResponse.data?.conversation) {
+            const updatedMessages = [...sessionResponse.data.conversation];
+            
+            // Find the last user message (which should be the one we're editing)
+            if (updatedMessages.length > 0) {
+              const lastMessage = updatedMessages[updatedMessages.length - 1];
+              if (lastMessage.role === 'user') {
+                // Update the content of the last message with new content
```
katzdave: Can we delete most of these comments.

katzdave:
Can you try reloading a session from history; does it follow the expectation there where the old message is removed/replaced?

bzqzheng:
> Can you try reloading a session from history; does it follow the expectation there where the old message is removed/replaced?

not sure if i understand this question.

I'm experiencing the same issue from reloading the past sessions;

PR #6962, opened 2026-02-04
feat: support for Moonshot LLMs (Kimi 2.5 etc)

### LINKED ISSUE #6902: [Bug] Kimi k2.5 API Error: "thinking is enabled but reasoning_content is missing" in tool calls
I am encountering a blocking 400 error when using the Kimi k2.5 model (Moonshot AI) with Goose on macOS.

The error occurs specifically after the model attempts to use a tool. It appears that when "thinking/reasoning" is enabled on the model, the API requires the reasoning_content field to be preserved in the message history for assistant messages. If Goose strips this field or fails to send it back in the subsequent request, the API rejects the request.

---

**To Reproduce**
1. Configure Goose to use Kimi k2.5 (via Moonshot AI / OpenAI compatible provider).
2. Start a conversation that requires the agent to use a tool (e.g., creating a file or running a terminal command).
3. Allow the model to generate a response that includes both "thinking" (reasoning) and a tool call.
4. Observe the error when the tool output is submitted back to the model.

---

**Expected behavior**
Goose should preserve the reasoning_content (or equivalent thinking field) from the assistant's previous message when submitting tool outputs back to the API, preventing the validation error.

---

**Screenshots**
[Screenshot](https://share.cleanshot.com/0XB2pt4d)

---

- **OS & Arch:** [Macos 15.7.3]
- **Interface:** [UI]
- **Version:** [v1.22.2]
- **Provider & Model:** [Moonshot – Kimi K2.5]

stebbins:
## Summary
- add support for Moonshot LLMs 
- add moonshot provider declaration 
- support `reasoning_content` when these models use thinking mode + tool call
- source for context etc https://platform.moonshot.ai/docs/introduction#kimi-k2-model

### Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing

- CI tests
- manual build and testing with Moonshot providers
- NOTE: I do not have OpenAI access, so cannot directly manually "regression" test OpenAI 

### Related Issues
Closes #6902  

Before:  

image

After:   

image

### DIFF
--- a/crates/goose/src/providers/formats/openai.rs
+++ b/crates/goose/src/providers/formats/openai.rs
@@ -82,7 +82,7 @@ pub fn format_messages(messages: &[Message], image_format: &ImageFormat) -> Vec<
         let mut output = Vec::new();
         let mut content_array = Vec::new();
         let mut text_array = Vec::new();
-        let mut reasoning_text: Option<String> = None;
+        let mut reasoning_text = String::new();
 
         for content in &message.content {
             match content {
@@ -116,7 +116,7 @@ pub fn format_messages(messages: &[Message], image_format: &ImageFormat) -> Vec<
                     continue;
                 }
                 MessageContent::Reasoning(r) => {
-                    reasoning_text = Some(r.text.clone());
+                    reasoning_text.push_str(&r.text);
                 }
                 MessageContent::ToolRequest(request) => match &request.tool_call {
                     Ok(tool_call) => {
@@ -278,15 +278,11 @@ pub fn format_messages(messages: &[Message], image_format: &ImageFormat) -> Vec<
             converted["content"] = json!(null);
         }
 
-        // DeepSeek requires reasoning_content field when tool_calls are present
-        // Set it to the captured reasoning text, or empty string if not present
-        if converted.get("tool_calls").is_some() {
-            let reasoning = reasoning_text.unwrap_or_default();
-            converted["reasoning_content"] = json!(reasoning);
-        } else if let Some(reasoning) = reasoning_text {
-            if !reasoning.is_empty() {
-                converted["reasoning_content"] = json!(reasoning);
-            }
+        // Include reasoning_content only when non-empty.
+        // Kimi rejects empty reasoning_content (""), so we must omit it entirely
+        // when there's no reasoning to send.
+        if !reasoning_text.is_empty() {
+            converted["reasoning_content"] = json!(reasoning_text);
         }
 
         if converted.get("content").is_some() || converted.get("tool_calls").is_some() {
@@ -542,6 +538,7 @@ where
         use futures::StreamExt;
 
         let mut accumulated_reasoning: Vec<Value> = Vec::new();
+        let mut accumulated_reasoning_content = String::new();
 
         'outer: while let Some(response) = stream.next().await {
             if response.as_ref().is_ok_and(|s| s == "data: [DONE]") {
@@ -561,6 +558,9 @@ where
             if !chunk.choices.is_empty() {
                 if let Some(details) = &chunk.choices[0].delta.reasoning_details {
                     accumulated_reasoning.extend(details.iter().cloned());
+                }
+                if let Some(rc) = &chunk.choices[0].delta.reasoning_content {
+                    accumulated_reasoning_content.push_str(rc);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


Tyler-Hardin:
Have you tested a longer session? I rebased your PR onto main and it seems to fail with the same error after multiple turns.

```
$ ./target/release/goose
starting session | provider: custom_kimi model: kimi-k2.5
    session id: 20260210_7
    working directory: /x/qscratch/tyler/goose

Context: ○○○○○○○○○○ 0% (0/131072 tokens)
( O)> Write me an install.sh script that runs cargo build --release and, if it succeeds, copies ./target/release/goose to /export/mana/bin, but first creates a hardlink of the existing copy to a file in the same dir called goose.bk.$(date %Y%m%d.%H%M%S). not sure if my directions will actually achieve what i want -- i want to backup the original without modifying the timestamps. i kinda prefer this hardlink method because it seems more atomic. if we mv the original file first, for a split second there will be no goose installed. but if hardlink causes the backup to accidentally be overwritten, then use mv instead.

you attempted this previously. look at what's currently there in install.sh and tell me if you think it's good to use

path: /x/qscratch/tyler/goose/install.sh
command: view

Looking at the current `install.sh`, here's my assessment:

## What's Good

The script is **well-designed and safe to use**. It actually implements a **better** approach than your original description:

| Your Request | What the Script Does | Assessment |
|-------------|---------------------|------------|
| Hardlink existing, then copy new | Copy new to temp → hardlink existing → atomic mv | ✅ **Better** - The `mv` is truly atomic; your approach would have a window where the file is being copied |
| `goose.bk.$(date %Y%m%d.%H%M%S)` | `goose.bk.$(date +%Y%m%d.%H%M%S)` | ✅ Correct (the `+` is required for `date` command) |
| Fallback to mv if hardlink fails | Catches `ln` failure with `2>/dev/null` and falls back | ✅ Implemented |
| Atomic operation | Uses temp file + atomic `mv` | ✅ Yes - at no point is there no `goose` binary |

## Potential Issue

The backup could be overwritten if you run multiple installs within the same second (same timestamp). To fix this:
path: /x/qscratch/tyler/goose/install.sh
command: str_replace
old_str:     BACKUP="$DEST_DIR/goose.bk.$(date +%Y%m%d.%H%M%S)"
new_str:     BACKUP="$DEST_DIR/goose.bk.$(date +%Y%m%d.%H%M%S).$$"

Ran into this error: Request failed: Bad request (400): thinking is enabled but reasoning_content is missing in assistant tool call message at index 5.

Please retry if you think this is a transient or recoverable error.

⏱️  Elapsed time: 34.95s
Context: ●○○○○○○○○○ 5% (7171/131072 tokens)
```

Tyler-Hardin:
My guess is that moonshot might not be echoing back the full reasoning chain. They may only send you back the most recent one. Would be hypocritical of them, but would also explain the error. Idk tho, didn't look deeply.

stebbins:
@Tyler-Hardin thank you for testing, i will take a look

Tyler-Hardin:
Thanks for all your work on this. It was already working well for me before your last commit. (Not to suggest the last commit was unneeded, just that the last edge case you fixed wasn't something I had encountered yet.)

Tyler-Hardin:
By the way, fun fact, there's a bug introduced by 2bc368948b32e8094c47e79be014ea85c30f6759 that causes text_editor view to return nothing. I thought it might be a bug in your patch but it's actually in main, predating your changes.

stebbins:
@Tyler-Hardin cheers and thanks for the heads up. I've been running into more "edge cases" where the reasoning context is lost so trying a few different approaches

stebbins:
@Tyler-Hardin I think we're good!

crates/goose/src/providers/formats/openai.rs
```diff
@@ -278,15 +278,11 @@ pub fn format_messages(messages: &[Message], image_format: &ImageFormat) -> Vec<
...
-            let reasoning = reasoning_text.unwrap_or_default();
-            converted["reasoning_content"] = json!(reasoning);
-        } else if let Some(reasoning) = reasoning_text {
-            if !reasoning.is_empty() {
-                converted["reasoning_content"] = json!(reasoning);
-            }
+        // Include reasoning_content only when non-empty.
+        // Kimi rejects empty reasoning_content (""), so we must omit it entirely
+        // when there's no reasoning to send.
+        if !reasoning_text.is_empty() {
```
DOsinga, line 284: you are removing the existing comment here, which makes me slightly worried - is this now compatible with both Kimi & Deepseek?

DOsinga:
ok, I merged the other one, I think we can close this now. let me know if not. and thanks for the contribution @stebbins

PR #6964, opened 2026-02-04
feat: replace subagent and skills with unified summon extension

tlongwell-block:
Introduces the `summon` extension to unify goose's delegation and knowledge-loading capabilities.

## New Tools
| Tool | Purpose | Example |
|------|---------|---------|
| `load` | Discover sources or inject knowledge | `load()` lists all, `load(source: "rust-patterns")` loads content |
| `delegate` | Run task in isolated subagent | `delegate(instructions: "...")` or `delegate(source: "review", async: true)` |

## Features
- **Source discovery**: Recipes, skills, agents from local/global paths with priority resolution
- **Async execution**: Background tasks with turn tracking and MOIM status reporting
- **Unprefixed tools**: Cleaner tool names without `extension__` prefix

## Removed
- `subagent_tool.rs`, `skills_extension.rs`, `builtin_skills/` - replaced by summon

Closes #6202

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -199,10 +199,12 @@ pub fn get_parameter_names(tool: &Tool) -> Vec<String> {
     names
 }
 
+const TOOL_EXTENSION_META_KEY: &str = "goose_extension";
+
 pub fn get_tool_owner(tool: &Tool) -> Option<String> {
     tool.meta
         .as_ref()
-        .and_then(|m| m.0.get("goose_extension"))
+        .and_then(|m| m.0.get(TOOL_EXTENSION_META_KEY))
         .and_then(|v| v.as_str())
         .map(|s| s.to_string())
 }
@@ -913,7 +915,7 @@ impl ExtensionManager {
                                 .map(|m| m.0.clone())
                                 .unwrap_or_default();
                             meta_map.insert(
-                                "goose_extension".to_string(),
+                                TOOL_EXTENSION_META_KEY.to_string(),
                                 serde_json::Value::String(name.clone()),
                             );
 


DOsinga:
I haven't looked at the details here, but this feels like the right way to do this.

michaelneale:
@tlongwell-block so I think it is good to revisit skills, my main Q is - for the case of just skills, will it still have thge lightweight approach of name/description pairs in system prompt (always) an then a simple tool to load it progressively? as that is key for skills to be useful, dont' want to confuse it with other things it could do (even claude is not doing a great job at skills now as it gets distracted) - that is my main feedback/query but otherwise I think this makes a lot of sense and probably makes it more efficient I would hope.

tlongwell-block:
> @tlongwell-block so I think it is good to revisit skills, my main Q is - for the case of just skills, will it still have thge lightweight approach of name/description pairs in system prompt (always) an then a simple tool to load it progressively? as that is key for skills to be useful, dont' want to confuse it with other things it could do (even claude is not doing a great job at skills now as it gets distracted) - that is my main feedback/query but otherwise I think this makes a lot of sense and probably makes it more efficient I would hope.

Right now, the agent has to call load() to see the list of skills and a brief description. I was considering putting the list in the system prompt or the moim, but internally folks are just putting a 70+ skill repo in path, which... that's not ideal

Hence this current setup. I'm not sure exactly the best approach

tlongwell-block:
With my setup, I've found the only consistent way to get it to use skills is to reference them in AGENTS.md or my global .goosehints

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -111,10 +112,16 @@ impl CodeExecutionClient {
         let mut cfgs = vec![];
         for tool in tools {
             let full_name = tool.name.to_string();
-            let (server_name, tool_name) = full_name.split_once("__")?;
+            let (namespace, name) = if let Some((server, tool_name)) = full_name.split_once("__") {
+                (server.to_string(), tool_name.to_string())
+            } else if let Some(owner) = get_tool_owner(&tool) {
+                (owner, full_name)
+            } else {
+                continue;
```
jamadeo, line 120: this seems a bit to implicit -- tools become available in code mode iff they have a `__` or we tag them with `goose_extension`? It would be good to have a more explicit way of doing this

crates/goose/src/agents/extension.rs
```diff
@@ -162,12 +168,15 @@ impl PlatformExtensionContext {
...
 
+/// Definition for a platform extension that runs in-process with direct agent access.
 #[derive(Debug, Clone)]
 pub struct PlatformExtensionDef {
     pub name: &'static str,
     pub display_name: &'static str,
     pub description: &'static str,
     pub default_enabled: bool,
+    /// If true, tools are exposed without extension prefix for intuitive first-class use.
+    pub unprefixed_tools: bool,
```
jamadeo, line 193: to the other point about defining these explicitly, what about something like

```
pub first_class_tools: bool
```

where first-class tools don't get prefixed names and also exist alongside the `execute` tool (don't become code modules)

crates/goose/src/agents/extension_manager.rs
```diff
@@ -199,6 +199,31 @@ pub fn get_parameter_names(tool: &Tool) -> Vec<String> {
     names
 }
 
+pub fn get_tool_owner(tool: &Tool) -> Option<String> {
+    tool.meta
+        .as_ref()
+        .and_then(|m| m.0.get("goose_extension"))
+        .and_then(|v| v.as_str())
+        .map(|s| s.to_string())
+}
```
jamadeo, line 210: I'd suggest something like

```
trait ExtensionTool {
    fn get_extension_name(&self) -> Option<&str>;
    fn set_extension_name(&mut self, name: String);
}

impl ExtensionTool for Tool {
    fn get_extension_name(&self) -> Option<&str> {
        self.meta
            .as_ref()
            .and_then(|m| m.get("goose_extension"))
            .and_then(|v| v.as_str())
    }

    fn set_extension_name(&mut self, name: String) {
        let mut meta = self.meta.take().unwrap_or_default();
        meta.insert("goose_extension".to_string(), Value::String(name));
        self.meta = Some(meta)
    }
}
```

so you don't need to use the key everywhere or define it as a constant

crates/goose/src/agents/extension_manager.rs
```diff
@@ -199,6 +199,31 @@ pub fn get_parameter_names(tool: &Tool) -> Vec<String> {
...
+            .get(name_to_key(name).as_str())
+            .is_some_and(|def| def.unprefixed_tools)
+    } else {
+        false
+    }
+}
+
+/// Result of resolving a tool call to its owning extension
+struct ResolvedTool {
+    client_name: String,
```
jamadeo: extension_name ?

crates/goose/src/agents/reply_parts.rs
```diff
@@ -151,12 +150,16 @@ impl Agent {
             .is_extension_enabled(CODE_EXECUTION_EXTENSION)
             .await;
         if code_execution_active {
-            let code_exec_prefix = format!("{CODE_EXECUTION_EXTENSION}__");
-            let skills_prefix = format!("{SKILLS_EXTENSION}__");
+            let allowed_extensions = [CODE_EXECUTION_EXTENSION, SUMMON_EXTENSION];
```
jamadeo: these should probably come from the plaform definitions, where we decide whether they exist in code mode or as first-class tools

crates/goose/src/providers/testprovider.rs
```diff
@@ -72,9 +72,29 @@ impl TestProvider {
     }
 
     fn hash_input(messages: &[Message]) -> String {
+        use crate::conversation::message::MessageContent;
+
+        // Strip internal metadata (e.g. tool_meta/_meta) from content before hashing.
+        // This metadata is used for internal routing (like goose_extension ownership)
+        // and isn't part of the semantic input the LLM sees, so it shouldn't affect
+        // replay matching.
```
jamadeo, line 80: is this always going to be true? also, what breaks if it's kept in?

crates/goose/src/session/session_manager.rs
```diff
@@ -381,6 +381,7 @@ pub struct SessionStorage {
     pool: Pool<Sqlite>,
     initialized: tokio::sync::OnceCell<()>,
     session_dir: PathBuf,
+    write_lock: tokio::sync::Mutex<()>,
```
jamadeo: can you say something about why we need this lock? cc @DOsinga

crates/goose/src/recipe/mod.rs
```diff
@@ -226,6 +226,24 @@ pub struct RecipeBuilder {
 }
 
 impl Recipe {
+    fn ensure_summon_for_subrecipes(&mut self) {
```
jamadeo, line 229: you already know my thoughts on this :) but, I can see that fixing it is going to be its own refactor

crates/goose/src/session/session_manager.rs
tlongwell-block: Having a few dozen subagents across a few different sessions at once causes db lock errors to crop up and kill the agents

There were three ways to fix this, I figured:

* retries with backoff, which I thought was ugly
* switch to a DB that allows row level multi-writes
* lock

I thought locking would be sensible here. And it's only on writes

crates/goose/src/agents/extension.rs
tlongwell-block: I think `unprefixed_tools` is clearer and relatively self-documenting

And we don't necessarily want to exclude unprefixed_tools from code mode. load/delegate work well in code mode and should be available there

crates/goose/src/agents/code_execution_extension.rs
tlongwell-block: Sorry, this is not super clear on what is happening.

Before, when we didn't have unprefixed/first class tools, we could just split the tool name to get the server/extension name from it

Now we still try to do that to preserve the old behavior as much as possible, but fall back to the new `get_tool_owner` when we encounter an unprefixed tool

crates/goose/src/agents/extension_manager.rs
tlongwell-block: Right now it's used in only two places. I think it's probably reasonable to keep as is and move to the trait like you have here in a follow up if we end up expanding this machinery more.

crates/goose/src/providers/testprovider.rs
tlongwell-block: It should be. No reason to send explicitly internal metadata to providers.

It breaks replay tests without getting us anything in return since I expect we'll expand this in the future

crates/goose/src/agents/extension.rs
tlongwell-block: Actually, I will use `first_class_tools` to fix the issue in your other comment about where we defined allowed extensions

So I'll leave it as is here

crates/goose/src/recipe/mod.rs
tlongwell-block: Yes, 100%... this is not ideal. But it's how the backend is setup right now and I was trying to avoid touching it as much as possible in this PR

crates/goose/src/agents/extension_manager.rs
tlongwell-block: Done

crates/goose/src/agents/reply_parts.rs
tlongwell-block: Done

crates/goose/src/agents/code_execution_extension.rs
tlongwell-block: Now filtering happens in reply_parts.rs via is_first_class_extension(), which checks unprefixed_tools on the platform extension def.

crates/goose/src/agents/extension_manager.rs
jamadeo: Sure, but it's more about encapsulating the implementation, not code re-use. In any case, if you're going to read and write directly to meta, at least set the `"goose_extension"` as a const somewhere.

crates/goose/src/agents/code_execution_extension.rs
jamadeo: Ok, do we ever have a case then where get_tool_owner() returns None? Are we then adding that metadata to all tools?

It would be nice to get rid of the splitting on `__` entirely, and you're part way there by attaching the extension name directly to the tool

crates/goose/src/session/session_manager.rs
tlongwell-block: Reverted lock. Increased busy_timeout

crates/goose/src/agents/extension_manager.rs
tlongwell-block: Done

clouatre:
The regression introduced here (moving `delegate` into the `summon` platform extension without updating the auto-injection logic) is addressed in #7360.

PR #6966, opened 2026-02-04
custom provider form minor improvements

zanesq:
## Summary
Minor custom provider experience improvements

- fix checkboxes not visible
- we needed to use the native checkbox we don't actually use the radix checkbox anywhere in the codebase so its not styled properly (can come back to that later)
- noticed when editing the previous model from a different provider was still hanging around
- moved api key to an authentication section to help visually distinguish it
- changed it to default unchecked with hide and show field since most local providers don't need an api key, added some helpful text about that
- improved model default selection in the switch modal while I was in there

image
image

### DIFF
--- a/ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
+++ b/ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
@@ -83,7 +83,11 @@ export const SwitchModelModal = ({
   const [provider, setProvider] = useState<string | null>(
     initialProvider || currentProvider || null
   );
-  const [model, setModel] = useState<string>(currentModel || '');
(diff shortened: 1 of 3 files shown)


ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
```diff
@@ -83,7 +83,11 @@ export const SwitchModelModal = ({
   const [provider, setProvider] = useState<string | null>(
     initialProvider || currentProvider || null
   );
-  const [model, setModel] = useState<string>(currentModel || '');
+  // Only use currentModel if we're not switching to a different provider
```
zanesq, line 86: thought this logic was worthy of a comment but can remove if people disagree

PR #6970, opened 2026-02-04
Fix 'Edit In Place' and 'Fork Session' features

### LINKED ISSUE #6954: "Edit In Place" feature only appends instead of edits

bzqzheng:
## Summary
Fixes #6954 - "Edit In Place" and "Fork Session" features were broken. Both now work correctly.

**✅ Verified with local UI testing**

### 1. Edit In Place - Duplicated Instead of Replacing
**Issue:** Edited messages were appended as duplicates instead of replacing the original.

**Cause:** Code tried to modify the wrong message (the one before the edit target) due to incorrect assumption about truncation behavior.

**Fix:** Create new message with edited content + pass `conversation_so_far` explicitly to backend.

### 2. Fork Session - Race Condition & Empty History
**Issue:** Forked sessions showed no history, appeared laggy, or threw `TypeError: Cannot read property 'trim'`.

**Causes:**
- Auto-submit fired before messages loaded (race condition)
- `initialMessage` was a string, not `UserInput` object
- Backend didn't receive explicit conversation state

**Fix:** 
- Wait for `messages.length > 0` before auto-submitting
- Convert string to `{msg, images}` format  
- Pass `conversation_so_far` to ensure correct state

## Changes
- `src/hooks/useChatStream.ts` - Edit In Place logic + explicit conversation passing
- `src/hooks/useAutoSubmit.ts` - Fork timing + type safety

## Verification
Both features tested and working:
- ✅ Edit In Place: Message replaces correctly, no duplication
- ✅ Fork Session: History loads → edited message auto-submits → new response

🤖 *Generated with [Claude Code](https://claude.ai/claude-code)*

### DIFF
--- a/ui/desktop/src/App.tsx
+++ b/ui/desktop/src/App.tsx
@@ -593,7 +593,7 @@ export function AppInner() {
         console.log('[App] Processing initial message from launcher:', initialMessage);
         navigate('/pair', {
           state: {
-            initialMessage,
+            initialMessage: { msg: initialMessage, images: [] },
           },
         });
         setTimeout(() => {
--- a/ui/desktop/src/components/BaseChat.tsx
+++ b/ui/desktop/src/components/BaseChat.tsx
@@ -279,7 +279,7 @@ export default function BaseChat({
       navigate(`/pair?${params.toString()}`, {
         state: {
           disableAnimation: true,
-          initialMessage: editedMessage,
+          initialMessage: editedMessage ? { msg: editedMessage, images: [] } : undefined,
         },
       });
     };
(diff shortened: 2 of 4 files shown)


zanesq:
Testing locally, may go to monday

PR #6974, opened 2026-02-05
Fast reply

DOsinga:
## Summary

What if we offer the same prompt to a slow and a fast model and pick whichever one wins?

### DIFF
--- a/ui/desktop/src/components/ChatInput.tsx
+++ b/ui/desktop/src/components/ChatInput.tsx
@@ -276,10 +276,9 @@ export default function ChatInput({
         ? filteredText.replace(/\bsubmit[.,!?;'"\s]*$/i, '').trim()
         : filteredText;
 
-      const newValue =
-        displayValue.trim() && cleanedText
-          ? `${displayValue.trim()} ${cleanedText}`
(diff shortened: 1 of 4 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -1168,11 +1168,11 @@ impl Agent {
                     &working_dir,
                 ).await;
 
-                let mut stream = Self::stream_response_from_provider(
+                let mut stream = Self::reply_fast_and_slow(
```
katzdave, line 1171: Why are we calling here?

Seems like a pure win for fast model use cases like compaction though (especially given the fast model is flaky on databricks).

DOsinga: this is an idea I was discussing online and could probably do with a discussion

the idea is that we call both the fast and the slow model at the same time, but the fast model comes with a very simple prompt asking the model for a one line answer, a one line shell command or just saying it can't do it. if that can do it and returns before the slow model, we cancel the slow model and have a much faster response time. if it returns it cannot do it, we keep waiting on the slow model

DOsinga:
let's keep thinking

Issue #5716, opened 2025-11-13
Packaging goose to Fedora


r0x0d:
**Please explain the motivation behind the feature request.**
Fedora does not have a terminal AI assistant at the moment package in the main repositories, and I would like to package it for you guys.

For now, this would be mainly for tackling the CLI portion of Goose, and perhaps, later, we could figure out how to get Goose Desktop UI in it as well. 

**Describe the solution you'd like**
Having a rpm to package this to Fedora in upstream

**Describe alternatives you've considered**
None

**Additional context**
I have already a build going through COPR here https://github.com/rhel-lightspeed/goose. Would love some reviews and opinions on it. 

Package review for fedora being tracked at: https://bugzilla.redhat.com/show_bug.cgi?id=2428704

- [x] I have verified this does not duplicate an existing feature request

DOsinga:
Sounds great! How can we best help?

r0x0d:
> Sounds great! How can we best help?

Hey, @DOsinga! 

So, I think for now, the packaging worked out very well (at least in COPR). My next steps are getting to Fedora repos. Getting your ack in here is a huge help, as I believe this is one of the criteria for pushing to Fedora (IIRC). 

One think I would like to ask for support here is to understand if the generated artifacts `goose` and `goosed` is really the only thing that needs to existing in the user system. I know that when we do a `cargo build --release`, other binaries are generated as well. Do I need to include those too?

r0x0d:
Okay, so, I have been looking at building goose without vendoring the dependencies, and right now, it seems that I have around [75 errors](https://download.copr.fedorainfracloud.org/results/r0x0d/goose/fedora-rawhide-x86_64/09821399-goose/builder-live.log.gz), but, some of them I can fix by updating dependencies or patching them downstream. 

I will open a PR to get the following deps updated in goose: 

```
 Problem 60: nothing provides requested (crate(tree-sitter-go/default) >= 0.21.0 with crate(tree-sitter-go/default) < 0.22.0~)
 Problem 61: nothing provides requested (crate(tree-sitter-java/default) >= 0.21.0 with crate(tree-sitter-java/default) < 0.22.0~)
 Problem 62: nothing provides requested (crate(tree-sitter-javascript/default) >= 0.21.0 with crate(tree-sitter-javascript/default) < 0.22.0~)
[...]

PR #6980, opened 2026-02-05
Implement manpage generation for goose-cli

r0x0d:
## Summary

This patch brings the capabilities of generating manpages using CLI descriptions.

This process is not automated as it would probably not be included in every release or target, so a `just` target was created to easy the work whenever it is needed.

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [x] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Related Issues
Relates to #5716  
Discussion: LINK (if any)

Before:  

After:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1854,6 +1854,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c3e64b0cc0439b12df2fa678eae89a1c56a529fd067a9115f7827f1fffd22b32"
 
 [[package]]
+name = "clap_mangen"
+version = "0.2.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "439ea63a92086df93893164221ad4f24142086d535b3a0957b9b9bea2dc86301"
+dependencies = [
+ "clap",
+ "roff",
+]
+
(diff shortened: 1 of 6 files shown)


r0x0d:
This will help a lot for packaging in Fedora, but I'm pretty sure that it will also help every other package distribution!

katzdave:
What do the actual manpages look like? Should they be in git?

r0x0d:
> What do the actual manpages look like? Should they be in git?

Ah, good question! No, definitely not. 

Manpages are something we want to generate on each release to give people some insights on the CLI args/commands and some usage, when convenient. For instance, take a look at this example (online manpage: https://linux.die.net/man/1/3ddesk). 

If you spawn any linux container (e.g; `docker run -it fedora:latest bash`) and run `man man`, you will se the docs for the `man` command. 

This will be helpful for downstream packagers to provide some easy documentation for goose, and it's sub-commands. I know that the user could just run `goose --help`, but man pages could contain more information in the future, like `usage`, `config files schema` and etc.

PR #6984, opened 2026-02-05
Smoke tests: split compaction test and use debug build

jamadeo:
- use a debug build for the smoke tests
- split the compaction tests to their own job so they can run concurrently


alexhancock (approved):

alexhancock:
Nice! Could be some time-savings

PR #6988, opened 2026-02-05
feat: add User-Agent header to MCP HTTP requests

aharvard:
## Summary

Adds a `goose/{version}` User-Agent header to all HTTP requests sent to MCP servers via streamable HTTP transport.

## Why

MCP server operators need visibility into which clients are connecting to their servers. A User-Agent header enables:

- **Analytics** - Understanding client distribution and usage patterns
- **Debugging** - Correlating server-side issues with specific client versions
- **Compatibility** - Detecting outdated clients that may need different handling
- **Security** - Identifying and blocking problematic client versions if needed

This is standard practice for HTTP clients and aligns with how other MCP clients identify themselves.

## Changes

- Added `goose_user_agent()` helper function that returns `goose/{CARGO_PKG_VERSION}`
- Added User-Agent header to initial HTTP client in `create_streamable_http_client()`
- Added User-Agent header to OAuth retry flow HTTP client

## Format

The header follows the standard format:
```
User-Agent: goose/1.23.0
```

## Testing

- All 17 extension_manager tests pass
- `cargo clippy` passes
- `cargo fmt` applied

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -372,9 +372,8 @@ fn substitute_env_vars(value: &str, env_map: &HashMap<String, String>) -> String
     result
 }
 
-fn goose_user_agent() -> String {
-    format!("goose/{}", env!("CARGO_PKG_VERSION"))
-}
+const GOOSE_USER_AGENT: reqwest::header::HeaderValue =
+    reqwest::header::HeaderValue::from_static(concat!("goose/", env!("CARGO_PKG_VERSION")));
 
 async fn create_streamable_http_client(
     uri: &str,
@@ -386,12 +385,7 @@ async fn create_streamable_http_client(
 ) -> ExtensionResult<Box<dyn McpClientTrait>> {
     let mut default_headers = HeaderMap::new();
 
-    default_headers.insert(
-        reqwest::header::USER_AGENT,
-        goose_user_agent()
-            .parse()
-            .expect("valid user agent header value"),
-    );
+    default_headers.insert(reqwest::header::USER_AGENT, GOOSE_USER_AGENT);
 
     for (key, value) in headers {
         let substituted_value = substitute_env_vars(value, all_envs);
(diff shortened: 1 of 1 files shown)


crates/goose/src/agents/extension_manager.rs
```diff
@@ -381,6 +385,14 @@ async fn create_streamable_http_client(
     provider: SharedProvider,
 ) -> ExtensionResult<Box<dyn McpClientTrait>> {
     let mut default_headers = HeaderMap::new();
+
+    default_headers.insert(
+        reqwest::header::USER_AGENT,
+        goose_user_agent()
+            .parse()
+            .expect("valid user agent header value"),
```
katzdave: Don't think we can to panic on this. Maybe have a regex check on the pkg version, and otherwise just send goose/?

crates/goose/src/agents/extension_manager.rs
```diff
@@ -414,7 +426,20 @@ async fn create_streamable_http_client(
         let am = oauth_flow(&uri.to_string(), &name.to_string())
             .await
             .map_err(|_| ExtensionError::SetupError("auth error".to_string()))?;
-        let auth_client = AuthClient::new(reqwest::Client::default(), am);
+        let mut auth_headers = HeaderMap::new();
+        auth_headers.insert(
+            reqwest::header::USER_AGENT,
+            goose_user_agent()
+                .parse()
+                .expect("valid user agent header value"),
```
katzdave: and here.

crates/goose/src/agents/extension_manager.rs
```diff
@@ -381,6 +385,14 @@ async fn create_streamable_http_client(
     provider: SharedProvider,
 ) -> ExtensionResult<Box<dyn McpClientTrait>> {
     let mut default_headers = HeaderMap::new();
+
+    default_headers.insert(
+        reqwest::header::USER_AGENT,
+        goose_user_agent()
+            .parse()
+            .expect("valid user agent header value"),
+    );
```
DOsinga: this feels odd, goose_user_agent could be done at compile time I think and even though parse will never fail, it would kill the agent if it did.

can we something like:

const GOOSE_USER_AGENT = HeaderValue::from_static(concat!("goose/", env!("CARGO_PKG_VERSION")));

and then here insert that?

crates/goose/src/agents/extension_manager.rs
DOsinga: yes

aharvard:
Hey @DOsinga @katzdave — good catch! Updated to use a compile-time constant with `HeaderValue::from_static(concat!(...))`. No more runtime parsing or potential panics. 👍

PR #7032, opened 2026-02-06
fix(ci): switch from cargo-audit to cargo-deny for advisory scanning

codefromthecrypt:
## Summary

Switch CI from `cargo-audit` to `cargo-deny` for advisory scanning. `cargo-deny` uses the actual dependency graph instead of scanning the lockfile, which eliminates false positives like the rsa advisory from unused sqlx-mysql.

### Type of Change
- [x] Security fix
- [x] Build / Release

### AI Assistance

### Testing

```bash
cargo deny check advisories
```

### Related Issues

Follows up on #7031

### DIFF
--- a/.github/workflows/cargo-deny.yml
+++ b/.github/workflows/cargo-deny.yml
@@ -1,9 +1,9 @@
-name: "Cargo Audit"
+name: "Cargo Deny"
 on:
   push:
     paths:
       # Run if workflow changes
-      - '.github/workflows/cargo-audit.yml'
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
@michaelneale figure you might care about this.. main difference here is it is more precise about deps actually used. it makes it more clear that things that are unmaintained are not audited, though I think we could revisit that. It might be a fool's errand to try to make goose solely depend on maintained things as it applies transitively

PR #6972, opened 2026-02-05
feat: MCP support for agentic CLI providers

codefromthecrypt:
## Summary

Agentic CLI providers (`claude-code`, `codex`) can now pass MCP extensions to the underlying agent!

This works by adding `Vec<ExtensionConfig>` to `ProviderDef::from_env`, and threading it as necessary. This allows agentic providers to parse MCP extensions.
- `claude_code.rs` converts extensions to `--mcp-config` JSON
- `codex.rs` converts them to `-c mcp_servers.*` TOML overrides.

There was another dependency to avoid empty key names in the MCP config;
- Extension names are derived at construction time (from URI host or command basename), so `key()` always has a name to normalize.
- Secrets referenced via `env_keys` are resolved at runtime by the extension manager, never persisted to the session database.

### Type of Change
- [x] Feature
- [x] Tests

### AI Assistance
- [x] This PR was created or reviewed with AI assistance

### Testing

```shell
$ cargo test --test providers -- test_claude_code_provider test_codex_provider --nocapture
   Compiling goose v1.23.0 (/Users/codefromthecrypt/oss/goose/crates/goose)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 12.36s
     Running tests/providers.rs (target/debug/deps/providers-71b9d9da4d87344c)

running 2 tests
=== codex::model_listing ===
[crates/goose/tests/providers.rs:315:9] &models = [
    "gpt-5.2-codex",
    "gpt-5.2",
    "gpt-5.1-codex-max",
    "gpt-5.1-codex-mini",
]
===================
=== claude-code::model_listing ===
[crates/goose/tests/providers.rs:315:9] &models = [
    "default",
    "sonnet",
    "haiku",
]
===================
=== codex::basic_response === Hello!
=== claude-code::basic_response === Hello! 👋 How can I help you today?
=== claude-code::tool_usage === test-uuid-12345-67890
=== codex::tool_usage === test-uuid-12345-67890
=== claude-code::image_content === the image is a simple white background with black text that reads:

**"hello goose! this is a test image."**
=== codex::image_content === the image is a white background with centered black text that reads: “hello goose! this is a test image.”
=== claude-code::model_switch (default -> sonnet) === Hello! 👋 How can I help you today?
test test_claude_code_provider ... ok
=== codex::context_length_exceeded_error ===
[crates/goose/tests/providers.rs:234:9] &result = Err(
    ContextLengthExceeded(
[...]

PR #7033, opened 2026-02-06
feat: add image support and improve error resilience for Codex

codefromthecrypt:
## Summary

Fix image support for both Codex providers. Previously, users could paste or attach images in goose desktop, but both the Codex CLI and ChatGPT Codex providers silently dropped them — the image appeared in the chat UI but was never sent to the model.

**Codex CLI provider** (`codex`): Images are now decoded from base64, written to temp files, and passed to the CLI via the `-i` flag.

**ChatGPT Codex provider** (`chatgpt_codex`): Images are now serialized as `input_image` content items with a `data:` URI in the Responses API payload.

### Additional improvements (Codex CLI)

- **Model list**: Added `fetch_supported_models` so the UI model dropdown populates correctly
- **Error resilience**: Context window exhaustion and rate limit errors are now classified correctly, enabling automatic context compaction and backoff instead of hard failure. Partial responses are preserved on non-zero exit.
- **Hang prevention**: A stderr drain prevents pipe buffer deadlock when the CLI writes large error output.

### Type of Change
- [x] Feature
- [x] Bug fix

### AI Assistance

### Testing

Image path

run `just run-ui` then attach an test_image.jpg, ask "what's in this image?":

### Related Issues

Extracted from #6972

Before — Codex CLI (goose v1.23.1, image visible in chat but silently dropped):
codex-cli-before

Before — ChatGPT Codex (goose v1.23.1, image visible in chat but silently dropped):
chatgpt-codex-before

After — Codex CLI (image passed via `codex exec -i <FILE>`):
codex-cli-after

After — ChatGPT Codex (image sent as `input_image` in API payload):
chatgpt-codex-after

### DIFF
--- a/crates/goose/src/providers/codex.rs
+++ b/crates/goose/src/providers/codex.rs
@@ -516,7 +516,18 @@ fn prepare_input(
                     let decoded = BASE64.decode(&img.data).map_err(|e| {
                         ProviderError::RequestFailed(format!("Failed to decode image: {}", e))
                     })?;
-                    let ext = img.mime_type.split('/').next_back().unwrap_or("png");
+                    // Codex only supports png and jpeg:
+                    // https://github.com/openai/codex/blob/aea7610c/codex-rs/utils/image/src/lib.rs#L162-L167
+                    let ext = match img.mime_type.as_str() {
+                        "image/png" => "png",
+                        "image/jpeg" => "jpg",
+                        _ => {
+                            return Err(ProviderError::RequestFailed(format!(
+                                "Unsupported image MIME type for Codex: {}",
+                                img.mime_type
+                            )));
+                        }
+                    };
                     let mut tmp = tempfile::Builder::new()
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/codex.rs
```diff
@@ -161,7 +137,10 @@ impl CodexProvider {
         messages: &[Message],
         _tools: &[Tool],
     ) -> Result<Vec<String>, ProviderError> {
-        let prompt = self.messages_to_prompt(system, messages);
+        // Single pass: text → prompt (stdin), images → temp files (-i flags)
+        let image_dir = Paths::state_dir().join("codex/images");
+        std::fs::create_dir_all(&image_dir).ok();
```
jh-block, line 142: nit: ignoring the possible error here means that we'll fail later, inside prepare_input, when we try to create the temp file, which could be more confusing than if we failed here.

PR #7035, opened 2026-02-06
fix: fix snake_case function names in code_execution instructions

rabi:
## Summary
The execute tool description used snake_case function names (e.g., `Developer.text_editor`, `Github.create_issue`) but the TypeScript runtime expects camelCase (`Developer.textEditor`, `Github.createIssue`) since pctx_codegen converts all MCP tool names to camelCase. This causes TS2551 type errors at runtime.

Also:
- Added explicit "Function names are always camelCase" rule with examples.
- Added HANDLING RETURN VALUES section: many MCP tools return wrapper objects not raw arrays, causing `TypeError: x.filter is not a function` when LLMs assume array returns.
- Corrected available APIs: `fetch()` (host-restricted) and `console` methods are available, not forbidden. Removed Deno-specific references.

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested locally with nemotron nano and the results are much better.

### DIFF
--- a/crates/goose/src/agents/code_execution_extension.rs
+++ b/crates/goose/src/agents/code_execution_extension.rs
@@ -342,7 +342,7 @@ impl McpClientTrait for CodeExecutionClient {
                         Get detailed type information for specific functions.
 
                         Provide a list of function identifiers in the format "Namespace.functionName"
-                        (e.g., "Developer.shell", "Github.create_issue").
+                        (e.g., "Developer.shell", "Github.createIssue").
 
                         Returns full TypeScript interface definitions with parameter types,
                         return types, and descriptions for the requested functions.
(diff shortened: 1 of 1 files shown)


codefromthecrypt (approved):

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -384,13 +384,16 @@ impl McpClientTrait for CodeExecutionClient {
                         KEY RULES:
                         - Code MUST define an async function named `run()`
                         - All function calls are async - use `await`
-                        - Access functions as Namespace.functionName() (e.g., Developer.shell, Github.create_issue)
+                        - Function names are always camelCase (e.g., Developer.textEditor, Github.listIssues, Github.createIssue)
                         - Return value from `run()` is the result, all `console.log()` output will be returned as well.
-                        - Only functions from `list_functions()` are available - no `fetch()`, fs, or other Node/Deno APIs
+                        - Only functions from `list_functions()` are available, plus `fetch()` (restricted to allowed hosts) and `console` methods
```
alexhancock: should we add `console` to `list_functions` instead?

i also imagine keeping 

```
no `fetch()`, fs, or other Node/Deno APIs
```

as-is is ok right?

rabi: I think console is part of the sandbox runtime layer, not the tool/MCP layer. Adding it to list_functions would require modifying the upstream pctx_code_mode, right?

pctx mentions fetch as host restricted https://github.com/portofcontext/pctx/blob/main/crates/pctx_code_execution_runtime/src/lib.rs#L34, but yeah we can leave it as is.

alexhancock (approved):

PR #7039, opened 2026-02-06
feat(apps): add support for MCP apps to sample

alexhancock:
**Summary**

* Sparked by conversations with @DOsinga and @liady
* Adds support from sampling from inside MCP Apps
* Uses the existing postmessage bridge, but sends new `sampling/createMessage` reqs
* Makes a new route in `goosed` for handling
* Allows for text or images
* Adds an example app with a simple chat interface as a demonstration

**Demo**

Screenshot 2026-02-06 at 2 06 20 PM

**Request Flow**

```
┌─────────────────────────────────────────────────────────────────────┐
│                         Desktop (Electron)                          │
│  ┌──────────────┐      ┌──────────────────┐      ┌──────────────┐  │
│  │   MCP App    │      │  McpAppRenderer  │      │  goose-server│  │
│  │   (iframe)   │      │   (React host)   │      │   (goosed)   │  │
│  └──────┬───────┘      └────────┬─────────┘      └──────┬───────┘  │
│         │                       │                       │          │
│         │ postMessage           │                       │          │
│         │ sampling/createMessage│                       │          │
│         │──────────────────────>│                       │          │
│         │                       │                       │          │
│         │                       │  POST /sessions/      │          │
│         │                       │  {id}/sampling/message│          │
│         │                       │──────────────────────>│          │
│         │                       │                       │          │
│         │                       │                       │ provider │
│         │                       │                       │.complete │
│         │                       │                       │──────┐   │
│         │                       │                       │      │   │
│         │                       │                       │<─────┘   │
│         │                       │                       │          │
│         │                       │    CreateMessageResp  │          │
│         │                       │<──────────────────────│          │
│         │                       │                       │          │
│         │  response (model,     │                       │          │
│         │  role, content)       │                       │          │
│         │<──────────────────────│                       │          │
│         │                       │                       │          │
└─────────┴───────────────────────┴───────────────────────┴──────────┘
```

### DIFF
--- a/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
+++ b/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
@@ -40,6 +40,8 @@ import {
   McpAppToolInputPartial,
   McpAppToolResult,
   DimensionLayout,
+  SamplingCreateMessageParams,
+  SamplingCreateMessageResponse,
 } from './types';
 
 const DEFAULT_IFRAME_HEIGHT = 200;
@@ -185,6 +187,7 @@ function appReducer(state: AppState, action: AppAction): AppState {
 
   switch (action.type) {
     case 'FETCH_RESOURCE':
+      if (state.status === 'ready') return state;
       return { status: 'loading_resource', html, meta };
 
     case 'RESOURCE_LOADED':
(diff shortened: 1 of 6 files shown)


alexhancock:
cc @liady

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
```diff
@@ -232,14 +239,41 @@ export default function McpAppRenderer({
...
+              })),
+              systemPrompt,
+              maxTokens,
+            }),
+          });
+          if (!response.ok) {
+            throw new Error(`Sampling request failed: ${response.statusText}`);
+          }
+          return (await response.json()) as McpMethodResponse['sampling/createMessage'];
+        }
```
aharvard: this is fantastic! 

FYI, @alexhancock  I'm working to integrate `<AppRenderer/>` from `@mcp-ui/client` over [here](https://github.com/block/goose/pull/7013/changes#diff-29ea4a5f599b916170bfa9de29bed5ee5048f973a360339e7175b9773cb87d68R434-R450). It has dedicated props for callbacks that map to a subset of  JSON RPC messages supported by the MCP Apps spec. Don't think `sampling/createMessage` is supported on that side of things yet. Maybe I'm wrong? 

@liady, how can we support `sampling/createMessage` while using `<AppRenderer />`?  

++ @idosal, @ochafik, @infoxicator for input as well

aharvard: I discussed with @idosal that we might be able to have a generic `onRequest` prop added ` <AppRender />`, as a catch-all that can allow goose to listen for unsupported/non-standard JSON RPC methods. Not to say that `sampling/createMessage` is nonstandard, just not supported by the client SDK...

infoxicator:
Awesome stuff @alexhancock! looks like you did manage to spin something cool on the plane ride back!

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
alexhancock: Yeah, I can wait for https://github.com/block/goose/pull/7013 to go in and then adjust however needed. Feels like it would only be minor changes to use a better set of types etc.

Will stay tuned to this thread

DOsinga:
interesting - I like this of course, but i think we can do this simpler by just adding a tool to the apps platform extension, call_llm (or sample but I always found that confusing). we already have the plumbing in mcp apps for an app to call the mcp server that created it, so that should then all work. all we need to do is then add this to the prompt we use to generate or update the apps. let me know if you think this makes sense or we can chat about this?

alexhancock:
> interesting - I like this of course, but i think we can do this simpler by just adding a tool to the apps platform extension, call_llm (or sample but I always found that confusing). we already have the plumbing in mcp apps for an app to call the mcp server that created it, so that should then all work. all we need to do is then add this to the prompt we use to generate or update the apps. let me know if you think this makes sense or we can chat about this?

It does makes sense in the context of goose and the presence of that server, but I was trying to think of a way to do it that could work when the app is run in hosts other than goose as well. Would still need plumbing added in other hosts of course, but an app sending a `sampling/createMessage` across the postmessage bridge was my attempt at this. I am still familiarizing myself with apps. Does this make sense?

DOsinga:
yeah, but it will only work in other providers if this becomes part of the standard. we should push for that! but until that happens, what do we do? /cc @aharvard ?

DOsinga:
@idosal do you think adding sampling to the protocol like this would make sense?

aharvard:
> yeah, but it will only work in other providers if this becomes part of the standard. we should push for that! but until that happens, what do we do? /cc @aharvard ?

As to what we do, we can now use the new `onFallbackRequest` prop, that **_just_** landed (ty! @idosal) in `@mcp-ui/clients` `v1.6.1`. It's designed for hosts to experiment with implementing JSON RPC message-related content that might be promoted as a [standard message](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx#standard-mcp-messages) for MCP Apps. 

@alexhancock I upgraded to `v1.6.1` and stubbed the handler here: https://github.com/block/goose/pull/7208

Handling `sampling/createMessage` also feels like it could be easily promoted to become a standard for MCP Apps soon. For us that would show up as a prop on the `<AppRenderer />`, something like  `onCreateSamplingMessage`, so long as the broader MCP standard keeps it in place. Has there been talk of deprecation?

alexhancock:
> > yeah, but it will only work in other providers if this becomes part of the standard. we should push for that! but until that happens, what do we do? /cc @aharvard ?
> 
> As to what we do, we can now use the new `onFallbackRequest` prop, that **_just_** landed (ty! @idosal) in `@mcp-ui/clients` `v1.6.1`. It's designed for hosts to experiment with implementing JSON RPC message-related content that might be promoted as a [standard message](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx#standard-mcp-messages) for MCP Apps.
> 
> @alexhancock I upgraded to `v1.6.1` and stubbed the handler here: #7208
> 
> Handling `sampling/createMessage` also feels like it could be easily promoted to become a standard for MCP Apps soon. For us that would show up as a prop on the `<AppRenderer />`, something like `onCreateSamplingMessage`, so long as the broader MCP standard keeps it in place. Has there been talk of deprecation?

Yes this all sounds right to me

Putting together a proposal here https://gist.github.com/alexhancock/2bb8b62500f33ccb29a0edb4aa7c5342

@liady @idosal @DOsinga @aharvard @ochafik What's the best forum to actually submit something like this?

aharvard (approved):
This is gonna be a fundamental unlock. Great work!

PR #6946, opened 2026-02-04
Remove build-dependencies section from Cargo.toml

r0x0d:
## Summary

Looks like the build.rs script that was using tokio/reqwest was deleted in commit 75454b96d, and this is a leftover.

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [x] Other (specify below)

Cleanup technical debt

### AI Assistance

- [ ] This PR was created or reviewed with AI assistance

### Testing

Ran `cargo test` locally

### Related Issues
Relates to https://github.com/block/goose/pull/3115, #5716
Discussion: LINK (if any)

### Screenshots/Demos (for UX changes)
Before:  

After:

PR #7043, opened 2026-02-06
chore: update build deps

alexhancock:
An updated version of https://github.com/block/goose/pull/6946 that removes the `json` and `rustls-tls-native-roots` entries without removing the whole `build-dependencies` section

Reason: a newer change from @DOsinga added `system-proxy`. Douwe, unless you confirm that also isn't needed this seems like the best resolution for https://github.com/block/goose/pull/6946 for now

### DIFF
--- a/crates/goose/Cargo.toml
+++ b/crates/goose/Cargo.toml
@@ -17,8 +17,6 @@ workspace = true
 [build-dependencies]
 tokio = { workspace = true }
 reqwest = { workspace = true, features = [
-    "json",
-    "rustls-tls-native-roots",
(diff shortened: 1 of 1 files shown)


r0x0d:
Please, feel free to close mine if this is better suited.

alexhancock:
Looking at it closer, it looks like we can just remove it. Can you rebase yours @r0x0d ?

r0x0d:
> Looking at it closer, it looks like we can just remove it. Can you rebase yours @r0x0d ?

sure! let me rebase.

PR #7050, opened 2026-02-06
feat(ui): add session content search via API

### LINKED ISSUE #7045: Add content search for past sessions
## Problem
Session search currently only matches session titles. Cannot search within session content/messages.

## Expected Behavior
Search should index and match against full session content, not just titles.

## Use Case
Finding past conversations by topic, tool used, or specific text discussed, when the session title doesn't capture it.
As is, current search functionality is pretty useless.

tfsq:
## Summary
Add server-side content search for sessions, replacing the client-side title-only filtering.

## Changes
- **Backend**: Added `/sessions/search` endpoint that calls existing `search_chat_history` function
- **Frontend**: Updated `SessionListView` to call search API with fallback to client-side filtering

## How it works
1. User types in session search box
2. Debounced search term triggers API call to `/sessions/search?query=...`
3. Backend searches message content using existing `ChatHistorySearch`
4. Frontend filters session list to matching session IDs
5. Falls back to client-side title search if API fails

Closes #7045


crates/goose-server/src/routes/session.rs
```diff
@@ -510,3 +511,136 @@ pub fn routes(state: Arc<AppState>) -> Router {
...
+    ),
+    security(
+        ("api_key" = [])
+    ),
+    tag = "Session Management"
+)]
+async fn search_sessions(
+    State(state): State<Arc<AppState>>,
+    axum::extract::Query(params): axum::extract::Query<SearchSessionsQuery>,
+) -> Result<Json<SearchSessionsResponse>, StatusCode> {
```
DOsinga: I wonder if it would make sense to return an actual Conversation object here with the truncated messages. it is all local so it should be fast enough and that way we don't create an extra type which is handy when we want to do other things with the returned session

ui/desktop/src/components/sessions/SessionListView.tsx
```diff
@@ -352,32 +353,53 @@ const SessionListView: React.FC<SessionListViewProps> = React.memo(
...
+          const matchedSessionIds = new Set(resp.data.results.map((r: { sessionId: string }) => r.sessionId));
+          const filtered = sessions.filter((session) => matchedSessionIds.has(session.id));
+          
+          startTransition(() => {
+            setFilteredSessions(filtered);
+            setSearchResults(filtered.length > 0 ? { count: filtered.length, currentIndex: 1 } : null);
+          });
+        } catch (error) {
+          console.error('Search failed, falling back to client-side:', error);
+          // Fallback to client-side search on error
```
DOsinga: I would not. the backend shouldn't error.

ui/desktop/src/components/sessions/SessionListView.tsx
```diff
@@ -352,32 +353,53 @@ const SessionListView: React.FC<SessionListViewProps> = React.memo(
...
+          const resp = await searchSessions({
+            query: { query: debouncedSearchTerm },
+            throwOnError: true,
+          });
+          
+          const matchedSessionIds = new Set(resp.data.results.map((r: { sessionId: string }) => r.sessionId));
+          const filtered = sessions.filter((session) => matchedSessionIds.has(session.id));
+          
+          startTransition(() => {
+            setFilteredSessions(filtered);
```
DOsinga, line 368: ah so for now we're not actually returning the matches? maybe just return sessions without messages then?

PR #7052, opened 2026-02-06
Add global config switch to disable automatic session naming

baxen:
Adds a `GOOSE_DISABLE_SESSION_NAMING` configuration flag that lets users turn off the LLM-powered automatic session naming feature. This is useful in headless/CI environments or for users who find the background naming calls unnecessary, reducing token usage and latency.


codefromthecrypt:
https://github.com/block/goose/pull/7062 follow-up so it is used in acp and scenario tests

PR #7054, opened 2026-02-06
feat: Add summarize tool for deterministic reads

baxen:
Introduces a built-in `summarize` tool that lets the agent load files/directories and get an LLM-generated summary in a single call—a lightweight alternative to spawning a subagent when you already know what to analyze.

### DIFF
--- a/crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__all_platform_extensions.snap
+++ b/crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__all_platform_extensions.snap
@@ -94,6 +94,9 @@ and file sizes. When you need to search, prefer rg which correctly respects giti
 content. Then use cat or sed to gather the context you need, always reading before editing.
 Use write and edit to efficiently make changes. Test and verify as appropriate.
 
+## summarize
+
+
 ## summon
 
 ### Instructions


crates/goose/src/agents/agent.rs
```diff
@@ -567,6 +570,33 @@ impl Agent {
                 session.working_dir.clone(),
                 cancellation_token,
             )
+        } else if tool_call.name == SUMMARIZE_TOOL_NAME {
```
DOsinga: I don't think we want to inject this straight into the agent. you should wire this up through the platform extension mechanism

crates/goose/src/agents/code_summary_tool.rs
```diff
@@ -0,0 +1,506 @@
...
+pub struct SummarizeSettings {
+    pub provider: Option<String>,
+    pub model: Option<String>,
+    pub temperature: Option<f32>,
+    #[serde(flatten)]
+    pub extra: Option<HashMap<String, Value>>,
+}
+
+pub fn create_summarize_tool() -> Tool {
+    let schema = json!({
```
DOsinga: handcrafting json schema is rather fragile and this is just mirrorring the declaration that you have above, no?

crates/goose/src/agents/code_summary_tool.rs
```diff
@@ -0,0 +1,506 @@
...
+    lines: usize,
+}
+
+fn build_gitignore(working_dir: &Path) -> Gitignore {
+    let mut builder = GitignoreBuilder::new(working_dir);
+
+    // Add .gitignore if it exists
+    let gitignore_path = working_dir.join(".gitignore");
+    if gitignore_path.is_file() {
+        let _ = builder.add(&gitignore_path);
```
DOsinga: in our prompt manager we're not actually adding the .gitignore file, is that one faulty?

DOsinga (approved):
Seems like a useful tool to have and save context. we should bring it inline with the other build in tools though that now all are platform extensions

PR #7057, opened 2026-02-06
Always sync bundled extensions

zanesq:
## Summary
After https://github.com/block/goose/pull/6568 we stopped getting new bundled extensions after updating. This changes it so new bundled extensions are always applied.

PR #6568 (f3c2f8e8032) added a conditional check that only synced bundled extensions when no bundled extensions existed (i.e., fresh installs only):

```
// BROKEN CODE (before fix)
const hasBundledExtensions = extensions.some(
  (ext) => ext.type === 'builtin' && 'bundled' in ext && ext.bundled
);

if (!hasBundledExtensions) {
  await syncBundledExtensions(extensions, addExtensionForSync);
}
```
This broke the ability for existing users to receive new bundled extensions added to `bundled-extensions.json` in subsequent releases.

The fix removes this gate so `syncBundledExtensions` always runs, and the function itself correctly skips extensions that already exist with `bundled: true` while adding any new ones.

### DIFF
--- a/ui/desktop/src/components/ConfigContext.tsx
+++ b/ui/desktop/src/components/ConfigContext.tsx
@@ -207,28 +207,24 @@ export const ConfigProvider: React.FC<ConfigProviderProps> = ({ children }) => {
         const extensionsResponse = await apiGetExtensions();
         let extensions = extensionsResponse.data?.extensions || [];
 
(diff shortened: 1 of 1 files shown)


michaelneale:
I think this is ok - slightly worried about a distant memory of a bug that was caused here recently

Issue #7022, opened 2026-02-06
Groq provider: Only 5 models shown via dropdown despite having 20 models available


KiKaraage:
**Please provide the following information**
Fedora 43 x64. Desktop v1.23.0 via Flatpak.

---
**To Reproduce**
1. Install a fresh instance of Goose Desktop
2. Open Goose and "Welcome to Goose" screen with "Quick Setup with API Key" field will be shown
3. Insert a Groq API Key (`gsk_***`). Goose will show "Detected groq" with **(20 models available)** subcaption
<img width="704" height="759" alt="Image" src="https://github.com/user-attachments/assets/3d9633cb-89a8-489c-b5e2-c8ac6eb7e5e2" />
4. "Choose Model" modal will show up, but **only 5 models are shown**: gpt-oss-120b, llama-3.1-8b-instant, llama-3.3-70b-versatile, llama-guard-4-12b, and gpt-oss-20b

<img width="704" height="759" alt="Image" src="https://github.com/user-attachments/assets/e004c4c5-1599-408c-84e4-3d3720a3f3f9" />

---

**Expected behavior**
All 20 models should be shown, including Qwen3 and Kimi K2 Instruct

---

**Additional context: The Culprit**

https://github.com/block/goose/blob/44ac5dc1cee5685abf394c9c0ab1c34e496a63f2/crates/goose/src/providers/declarative/groq.json#L1-L31

This matches the available option on the dropdown, so editing groq.json should be sufficient to fix this bug. (I will do this and submit the PR)

***Reference**: https://github.com/block/goose/commit/bbcfe7f57175297a294d8e719694996f5a62afb8 - Editing declarative mistral.json file to make Devstral 2 & Devstral Small 2 models available on model dropdown list*

katzdave:
Is groq (d) the intended provider or is this an xai key? If it is the case of groq (d), then yes seems like the declarative provider would be the best place.

I'm cleaning shop with this new concept of canonical models based on models.dev data and haven't looked into the declarative providers much; Ideally we'd be able to pick up even more models from there somehow.

KiKaraage:
> Is groq (d) the intended provider or is this an xai key? If it is the case of groq (d), then yes seems like the declarative provider would be the best place.
> 
> I'm cleaning shop with this new concept of canonical models based on models.dev data and haven't looked into the declarative providers much; Ideally we'd be able to pick up even more models from there somehow.

[...]

PR #7065, opened 2026-02-07
fix: use dynamic model listing for all providers

rabi:
## Summary
Remove the early return in get_provider_models that returned static model names for declarative/custom providers. Now all providers go through fetch_recommended_models() first, with static known_models from metadata as fallback when fetching fails or returns empty. Respect allows_unlisted_models for free-text model input. Aligns both cli and ui behavior.

### Type of Change
- [ ] Feature
- [x] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Unit tests and tested locally.

### Related Issues
Relates to #7022

### DIFF
--- a/crates/goose-server/src/routes/config_management.rs
+++ b/crates/goose-server/src/routes/config_management.rs
@@ -372,19 +372,6 @@ pub async fn providers() -> Result<Json<Vec<ProviderDetails>>, ErrorResponse> {
 pub async fn get_provider_models(
     Path(name): Path<String>,
 ) -> Result<Json<Vec<String>>, ErrorResponse> {
-    let loaded_provider = goose::config::declarative_providers::load_provider(name.as_str()).ok();
-    // TODO(Douwe): support a get models url for custom providers
-    if let Some(loaded_provider) = loaded_provider {
-        return Ok(Json(
-            loaded_provider
-                .config
-                .models
-                .into_iter()
-                .map(|m| m.name)
-                .collect::<Vec<_>>(),
-        ));
-    }
-
     let all = get_providers().await.into_iter().collect::<Vec<_>>();
     let Some((metadata, provider_type)) = all.into_iter().find(|(m, _)| m.name == name) else {
         return Err(ErrorResponse::bad_request(format!(
@@ -403,10 +390,11 @@ pub async fn get_provider_models(
     let provider = goose::providers::create(&name, model_config).await?;
 
     let models_result = provider.fetch_recommended_models().await;
-
-    match models_result {
-        Ok(Some(models)) => Ok(Json(models)),
-        Ok(None) => Ok(Json(Vec::new())),
+    if let Err(ref e) = models_result {
+        tracing::warn!("Model fetch failed for provider '{}': {e}", name);
+    }
+    match metadata.resolve_models_with_fallback(models_result) {
+        Ok(models) => Ok(Json(models)),
         Err(provider_error) => Err(provider_error.into()),
     }
 }
--- a/crates/goose-cli/src/commands/configure.rs
+++ b/crates/goose-cli/src/commands/configure.rs
@@ -691,20 +691,21 @@ pub async fn configure_provider_dialog() -> anyhow::Result<bool> {
     };
     spin.stop(style("Model fetch complete").green());
 
-    // Select a model: on fetch error show styled error and abort; if Some(models), show list; if None, free-text input
-    let model: String = match models_res {
-        Err(e) => {
-            // Provider hook error
-            cliclack::outro(style(e.to_string()).on_red().white())?;
-            return Ok(false);
-        }
-        Ok(Some(models)) => select_model_from_list(&models, provider_meta)?,
-        Ok(None) => {
-            let default_model =
-                std::env::var("GOOSE_MODEL").unwrap_or(provider_meta.default_model.clone());
+    if let Err(ref e) = models_res {
+        let _ = cliclack::log::warning(format!("Could not fetch models: {e}"));
+    }
+    let model: String = match provider_meta.resolve_models_with_fallback(models_res) {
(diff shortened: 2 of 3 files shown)


crates/goose-server/src/routes/config_management.rs
```diff
@@ -372,19 +372,6 @@ pub async fn providers() -> Result<Json<Vec<ProviderDetails>>, ErrorResponse> {
 pub async fn get_provider_models(
     Path(name): Path<String>,
 ) -> Result<Json<Vec<String>>, ErrorResponse> {
-    let loaded_provider = goose::config::declarative_providers::load_provider(name.as_str()).ok();
-    // TODO(Douwe): support a get models url for custom providers
```
DOsinga, line 376: W00t!

DOsinga (commented):
so this seems like an improvement, but it looks like we're returning the static list when the dynamic list throws an error. should we not propogate that error and only return the static list if the remote return an empty list, i.e. is not implemented? we now use the fetch models in the desktop as a way to establish whether we actually have a working provider. before we had something similar as what I think is going on here and if you had your ollama misconfigured, it would show you 5 random models rather than telling you the provider didn't work.

rabi:
> but it looks like we're returning the static list when the dynamic list throws an error. should we not propogate that error and only return the static list if the remote return an empty list, i.e. is not implemented? we now use the fetch models in the desktop as a way to establish whether we actually have a working provider. before we had something similar as what I think is going on here and if you had your ollama misconfigured, it would show you 5 random models rather than telling you the provider didn't work.

Thanks @DOsinga for looking at it. I had tried to keep it as UX friendly as possible.

1. I think we're not using random models for fallback but a list of known models in the metadata like ‎crates/goose/src/providers/declarative/groq.json which we can trust

2. Also, I wanted to keep the existing behavior of listing known models

3. There may be some declarative providers that may not have have api to list models and fetch_supported_models would return an error in that case, so the fallback to known models seems necessary

4. I agree the desktop shouldn't rely on model listing as a provider health check — that was already broken for declarative providers (old code returned static models without any API call). I'll file a follow-up to add proper provider verification to the desktop, similar to the CLI's test_provider_configuration() which makes a real API call. That would cleanly separate "list models for the user" from "verify the provider", if that works?

codefromthecrypt:
I agree with @DOsinga that we should stop papering over model list failures. I implemented the alternative here https://github.com/block/goose/pull/7074

rabi:
> I agree with @DOsinga that we should stop papering over model list failures. I implemented the alternative here #7074

Declarative providers may not have a model listing API, we make the API call, but the actual inference endpoint might not implement /models. Built-in providers like Vertex AI handle this by overriding fetch_supported_models with a known model list, but declarative providers can't do that as they're defined by JSON and use the underlying provider's fetch_supported_models which makes a real API call, Having said that, if we want to go that route we can, though the curated model list we maintain would be redundant for the purpose

codefromthecrypt:
@rabi My read looking at the logic is that we are suggesting models out of a list of potentially hundreds due to fewer working with tools or tested with goose. this leaves:

* providers who implement list model (honestly should be all except few exception) and which are suggested
* providers who are hacking around lack of a model list and the suggested == only known lists. (esp with ACP this should approach zero after we delete CLI providers)

That's the idea in simplifying this problem. The more 3 level if statements we have, the harder it is to maintain and also the more unlike this codebase becomes vs model abstractions. I think as long as we make docs more coherent, we can de-tempt folks from being "too smart" with errors and still have a recommended list.

My 2p

codefromthecrypt:
obviated by https://github.com/block/goose/pull/7074

rabi:
> obviated by https://github.com/block/goose/pull/7074

As expected #7074 https://github.com/block/goose/blob/main/crates/goose/src/providers/claude_code.rs#L497,  broke the current workflow with cli configure, the only way you can workaround it is by forcing the model with env vars.. 

```
RUST_LOG=debug ./target/debug/goose configure


┌   goose-configure 
│
│  Configure Providers 
│
│  Claude Code CLI 
│
●  CLAUDE_CODE_COMMAND is already configured
│  
│  No 
│
◇  Model fetch complete
│
◆  Select a model:
│  ● sonnet 
│  ○ opus 

◇  Select a model:
│  opus 
│
◐  Checking your configuration...                                                                                                                                                                                   └  Configuration saved successfully to /tmp/.config/goose/config.yaml

RUST_LOG=debug ./target/debug/goose session
starting session | provider: claude-code model: opus
    session id: 20260209_4
    working directory: /tmp/work/LLM/goose

Context: ○○○○○○○○○○ 0% (0/128000 tokens)
( O)> can you write a poem of 20 lines with start and end markers?
API Error: 404 {"error":{"code":404,"message":"Publisher Model `projects/itpc-gcp-core-pe-eng-
claude/locations/us-east5/publishers/anthropic/models/claude-opus-4-1@20250805` was not found or your project does not have access to it. Please ensure you are using a valid model version. For more information, see: https://cloud.google.com/vertex-ai/generative-ai/docs/learn/model-versions","status":"NOT_FOUND"}}

⏱️  Elapsed time: 6.43s
Context: ○○○○○○○○○○ 0% (0/128000 tokens)
( O)> Press Ctrl+C again to exit, or type new instructions to continue
Closing session. Session ID: 20260209_4

GOOSE_PROVIDER=claude-code GOOSE_MODEL=claude-opus-4.6 ./target/debug/goose session
starting session | provider: claude-code model: claude-opus-4.6
    session id: 20260209_6
    working directory: /tmp/work/LLM/goose

( O)> can you write a poem of 20 lines with start and end markers?
I'll write a 20-line poem for you with start and end markers.

---START---
In circuits deep where silicon dreams flow,

```

PR #7029, opened 2026-02-06
feat(claude-code): use stream-json protocol for persistent sessions

codefromthecrypt:
## Summary

Switch the Claude Code provider from one-shot `-p` invocations to the stream-json protocol (`--input-format stream-json --output-format stream-json`). This is the same protocol used by the official Claude Agent SDKs (Go, TypeScript, Elixir) to communicate with the CLI as a persistent subprocess.

The old approach spawned a new `claude -p <messages_json>` process per turn, passing the entire conversation as a CLI argument. This hit OS argument size limits (E2BIG) on long conversations and lost all session context between turns. The new approach keeps a single CLI process alive, sends messages as NDJSON on stdin, and reads responses line-by-line from stdout -- giving goose persistent multi-turn sessions, image support, and proper error handling for free.

### Type of Change
- [x] Refactor / Code quality
- [x] Bug fix

### AI Assistance
- [x] This PR was created or reviewed with AI assistance

### Testing

```bash
GOOSE_PROVIDER=claude-code GOOSE_MODEL=sonnet target/release/goose run --text "say just hello"
# starting session | provider: claude-code model: sonnet
#     session id: 20260206_26
#     working directory: /Users/codefromthecrypt/oss/goose-goosed
# Hello! 👋
```

### Related Issues
Relates to https://github.com/block/goose/pull/6972

PR #7070, opened 2026-02-07
fix(claude-code): respawn persistent CLI process on broken pipe

rabi:
## Summary

Replace OnceCell with Mutex<Option<CliProcess>> so a dead CLI process can be detected and respawned. The persistent stream-json process (introduced in #7029) dies when the user uses Ctrl+C, leaving the OnceCell permanently poisoned with a stale process whose stdin pipe is closed.

### Type of Change
- [ ] Feature
- [x] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested locally and Ctrl+C works as expected.

### DIFF
--- a/crates/goose/src/providers/claude_code.rs
+++ b/crates/goose/src/providers/claude_code.rs
@@ -29,13 +29,68 @@ struct CliProcess {
     child: tokio::process::Child,
     stdin: tokio::process::ChildStdin,
     reader: BufReader<tokio::process::ChildStdout>,
-    #[allow(dead_code)]
     stderr_handle: tokio::task::JoinHandle<String>,
     messages_sent: usize,
 }
 
+impl CliProcess {
+    async fn send_and_read(
+        &mut self,
+        content_blocks: &[Value],
+    ) -> Result<Vec<String>, ProviderError> {
+        let ndjson_line = build_stream_json_input(content_blocks);
+        self.stdin
+            .write_all(ndjson_line.as_bytes())
+            .await
+            .map_err(|e| {
+                ProviderError::RequestFailed(format!("Failed to write to stdin: {}", e))
+            })?;
+        self.stdin.write_all(b"\n").await.map_err(|e| {
+            ProviderError::RequestFailed(format!("Failed to write newline to stdin: {}", e))
+        })?;
+
+        let mut lines = Vec::new();
+        let mut line = String::new();
+
+        loop {
+            line.clear();
+            match self.reader.read_line(&mut line).await {
+                Ok(0) => {
+                    return Err(ProviderError::RequestFailed(
+                        "Claude CLI process terminated unexpectedly".to_string(),
+                    ));
+                }
+                Ok(_) => {
+                    let trimmed = line.trim();
+                    if trimmed.is_empty() {
+                        continue;
+                    }
+                    lines.push(trimmed.to_string());
+
+                    if let Ok(parsed) = serde_json::from_str::<Value>(trimmed) {
+                        match parsed.get("type").and_then(|t| t.as_str()) {
+                            Some("result") => break,
+                            Some("error") => break,
+                            _ => {}
+                        }
+                    }
+                }
+                Err(e) => {
+                    return Err(ProviderError::RequestFailed(format!(
+                        "Failed to read output: {}",
+                        e
+                    )));
+                }
+            }
+        }
+
+        Ok(lines)
+    }
+}
+
 impl Drop for CliProcess {
     fn drop(&mut self) {
+        self.stderr_handle.abort();
         let _ = self.child.start_kill();
     }
 }
@@ -52,7 +107,7 @@ pub struct ClaudeCodeProvider {
     #[serde(skip)]
     name: String,
     #[serde(skip)]
-    cli_process: tokio::sync::OnceCell<tokio::sync::Mutex<CliProcess>>,
+    cli_process: tokio::sync::Mutex<Option<CliProcess>>,
 }
 
 impl ClaudeCodeProvider {
@@ -65,7 +120,7 @@ impl ClaudeCodeProvider {
             command: resolved_command,
             model,
             name: CLAUDE_CODE_PROVIDER_NAME.to_string(),
-            cli_process: tokio::sync::OnceCell::new(),
+            cli_process: tokio::sync::Mutex::new(None),
         })
     }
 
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
curious is this the right solution? "dies when the user uses Ctrl+C," wouldn't we want to not propagate the ctrl-c instead? respawning would lose state.

codefromthecrypt:
basically I'm not sure the flow, but in other tools I've worked on there's either intent that ctrl+c means exit the process (which should propagate) or it means stop the current command (should not).

maybe look at how other claude-code runners work as even if we want to restart the subprocess instead of handling the signal.. if we do it that way, it is best to cite why.

codefromthecrypt:
spent some time for a more thorough answer. 

---

The child dies on Ctrl-C because it inherits the parent's process group and receives SIGINT directly. Respawning works around this but has side effects:

- **Silent context loss** — the respawned CLI has no memory of the previous conversation. Message replay doesn't restore Claude Code's internal state (tool results, file edits, etc.).
- **Fragile error detection** — `is_recoverable` matches `"Broken pipe"` as a substring, which is OS/locale-dependent. Other death modes (SIGKILL, OOM) won't match.
- **Single retry, no backoff** — if the respawn also fails immediately, the error propagates with no further recovery.

The root cause fix is to spawn the child in its own process group so it doesn't receive SIGINT. For reference, [forge-core][forge-group] does this with `group_spawn()`. That one-line change would eliminate the need for respawn logic entirely.

<details><summary>See also: how other projects manage the Claude Code subprocess</summary>

| Project | Signal handling | Persistent NDJSON | Auto-restart |
|---------|-----------------|-------------------|--------------|
| [claude-agent-sdk-go][sdk-go] (Go) | [None][sdk-go-nosig] — no OS signal trapping. Close does [stdin close + kill after 5s][sdk-go-close]. Interrupt is a [protocol message][sdk-go-interrupt], not a signal. | [Yes][sdk-go-ndjson] — one process at Connect(), `--output-format stream-json --input-format stream-json`, [bufio.Scanner line reader][sdk-go-read] | [None][sdk-go-alive] — `IsAlive()` is stubbed (`return true`, TODO comment). EOF closes channel; consumer must re-`Connect()`. |
| [forge-core][forge] (Rust) | [kill_on_drop][forge-kill] + [group_spawn][forge-group] — process group killed when handle drops. No explicit signal trapping. | [Yes][forge-ndjson] — `--output-format=stream-json --input-format=stream-json` via [ProtocolPeer][forge-peer] on stdin/stdout. | [None][forge-spawn] — new process per execution. Supports `--fork-session`/`--resume` but no auto-respawn. |
| [mcpc][mcpc] (TypeScript) | [kill()][mcpc-kill] on `forceCleanup()`. No SIGINT/SIGTERM trapping on the parent process. | [Yes][mcpc-ndjson] — `spawn()` with `ndJsonStream(input, output)` over persistent stdin/stdout pipes. | [None][mcpc-connect] — if the process dies, session fails. No retry or respawn logic. |

</details>

---

[sdk-go]: https://github.com/anthropics/claude-agent-sdk-go
[sdk-go-nosig]: https://github.com/anthropics/claude-agent-sdk-go/blob/main/subprocess.go#L1-L12
[sdk-go-close]: https://github.com/anthropics/claude-agent-sdk-go/blob/main/transport.go#L327-L363
[sdk-go-interrupt]: https://github.com/anthropics/claude-agent-sdk-go/blob/main/client.go#L735-L745
[sdk-go-ndjson]: https://github.com/anthropics/claude-agent-sdk-go/blob/main/transport.go#L110-L113
[sdk-go-read]: https://github.com/anthropics/claude-agent-sdk-go/blob/main/transport.go#L279-L320
[sdk-go-alive]: https://github.com/anthropics/claude-agent-sdk-go/blob/main/subprocess.go#L137-L148
[forge]: https://github.com/namastexlabs/forge-core
[forge-kill]: https://github.com/namastexlabs/forge-core/blob/main/crates/executors/src/executors/claude.rs#L213
[forge-group]: https://github.com/namastexlabs/forge-core/blob/main/crates/executors/src/executors/claude.rs#L220
[forge-ndjson]: https://github.com/namastexlabs/forge-core/blob/main/crates/executors/src/executors/claude.rs#L106-L111
[forge-peer]: https://github.com/namastexlabs/forge-core/blob/main/crates/executors/src/executors/claude.rs#L239
[forge-spawn]: https://github.com/namastexlabs/forge-core/blob/main/crates/executors/src/executors/claude.rs#L202-L220
[mcpc]: https://github.com/mcpc-tech/mcpc
[mcpc-kill]: https://github.com/mcpc-tech/mcpc/blob/main/packages/acp-ai-provider/src/language-model.ts#L594-L611
[mcpc-ndjson]: https://github.com/mcpc-tech/mcpc/blob/main/packages/acp-ai-provider/src/language-model.ts#L310-L332
[mcpc-connect]: https://github.com/mcpc-tech/mcpc/blob/main/packages/acp-ai-provider/src/language-model.ts#L301-L333

rabi:
I looked into the group_spawn() approach in forge-core, the command_group crate uses calls pre_exec + setpgid in an **unsafe** block under the hood and would also introduce orphan-process risks if the parent exits without cleaning up the separate process group. From what I can tell, forge-core spawns a new process per execution, so they never face this persistent-process problem in the first place.
 
Silent context loss: Context loss would happen regardless if the process dies for any reason, Respawn at least lets the user continue rather than hitting a permanent error. The message replay sends the full conversation history.

Fragile error detection: Good point, but the try_wait() proactive check handles the common case, and the "terminated unexpectedly" string (which we control) covers the rest. But I can see if this can be improved, probably by preserving ErrorKind through the error chain.

Single retry, no backoff: I'm not sure backoff makes sense here. If the respawn fails, it's because the binary is gone or permissions changed,retrying won't help unlike a transient network error.

codefromthecrypt:
@rabi mainly I'm concerned about respawn in general especially if the rationale to do this is about "dies when the user uses Ctrl+C" having dealt with things like double ctrl+c problems in the past there are enough concerns and hard to test areas as they are.

I'll leave this to another reviewer if you insist on incidental respawn. I think if a process dies it is a problem and the cure worse than the disease applies. If we are killing a process when we shouldn't this seems just not the right fix.

Good luck either way out!

rabi:
@codefromthecrypt We had a new child process for every conversation turn before, and one has to use double Ctrl+C to exit, but you can stop a response using Ctrl+C (more amplified when using streaming). After the change to use a persistent process, the user gets below error after Ctrl+C and there is no way to continue without starting a new session.

```
   Context: ○○○○○○○○○○ 0% (100/200000 tokens)
    ( O)> write a poem
    Ran into this error: Request failed: Failed to write to stdin: Broken pipe (os error 32).
    
    Please retry if you think this is a transient or recoverable error.
    
    ⏱️  Elapsed time: 0.13s
    Context: ○○○○○○○○○○ 0% (257/200000 tokens)
    ( O)> Write a poem again
    Ran into this error: Request failed: Failed to write to stdin: Broken pipe (os error 32).
    
    Please retry if you think this is a transient or recoverable error.
```

I understand the concern about respawn hiding real problems, that's probably a valid worry. But Ctrl+C is just the most reproducible trigger; the persistent process can also die from OOM, crashes, etc. The respawn is meant as general resilience for a long-lived process, not specifically a Ctrl+C workaround. Open to alternatives, if there's a cleaner approach, but to me this seemed like the least invasive option since it's invisible to the user.

alexhancock (approved):
It makes sense to me as an issue, and the code looks reasonable to me.

May want @codefromthecrypt eyes on it as well

codefromthecrypt (changes requested):
the ctrl-C to cancel a stream should not stop any subprocess, not MCP, CLI or ACP processes.

The issue I have with this is that we are papering over a design mistake where we are papering over the cancel flow problem. I've attempted to explain this, having personally dealt with this in other projects which deal with the double ctrl-c problem.

I don't think the "extra reslience" argument holds in this context, and when we move to ACP and delete this provider, we won't control that process directly anyway.

I will raise an alternative.

codefromthecrypt:
First PR to expect will do this:

1. add a practical test with the double ctrl C so we know the impact on stream cancel and our fix
2. the fix to the cancel propagation problem

That way, we can keep separate any "enhanced resilience" concepts like potentially auto-restarting in a loop, as doing anything like that, if we wanted to do that would be for any subprocess we launch (MCP, ACP, and CLI while they exist) or we'd need to explain why only do this in claude.

codefromthecrypt:
also, noticed with claude only complete_with_model is defined, not streaming, so this is important part of the problem statement.

https://github.com/block/goose/blob/main/crates/goose/src/providers/claude_code.rs#L507

alexhancock:
Ah I missed that you'd already reviewed @codefromthecrypt; I just looked at the diff and weighed in, missed the previous conversation.

codefromthecrypt:
this peels off the first part. we were inconsistent between MCP and CLI providers, so it is an easy fix.

Note I wasn't able to reproduce the ctrl-c killing things, possibly due to platform or linux in use. we can retry after with exact platform info to see what's going on. https://github.com/block/goose/pull/7083

rabi:
> Note I wasn't able to reproduce the ctrl-c killing things

Did you try to do ctrl+c during the wait (before the response)?

```
( O)> can you write a poem of 20 lines with start and end markers?
◒  Decoding human intent...                                                                                                                                                                                         ◐  Decoding human intent...                                                                                                                                                                                         Interrupted before the model replied and removed the last message.

⏱️  Elapsed time: 6.03s
( O)> can you write a poem of 20 lines with start and end markers?
Ran into this error: Request failed: Failed to write to stdin: Broken pipe (os error 32).

Please retry if you think this is a transient or recoverable error.


```
> also, noticed with claude only complete_with_model is defined, not streaming, so this is important part of the problem statement.

Yes it is, I think I mentioned earlir in the conversation that the issue was noticed in the context of the streaming implementation I was doing in https://github.com/block/goose/pull/6833, but it applies to when we're not using streaming as demonstrated above.

codefromthecrypt:
thanks for the clarification @rabi please check based on latest main (build etc) without streaming change. if it kills the subprocess, please report back the Claude Code etc version and platform. if it doesn't kill the subprocess, let's close this PR out so we can focus!

cheers

rabi:
> thanks for the clarification @rabi please check based on latest main (build etc) without streaming change. if it kills the subprocess, please report back the Claude Code etc version and platform. if it doesn't kill the subprocess, let's close this PR out so we can focus!
> 
> cheers

Yeah, thanks! https://github.com/block/goose/pull/7083 would fix that issue. But I'll keep my concern about child dying without the parent knowing about it (one has to start  a new session) for another day :smile: I've updated https://github.com/block/goose/pull/6833 to implement streaming and also handle the draining of leftover NDJSON events from the cancelled response. Have a look when possible.

PR #7076, opened 2026-02-08
fix: ensure assistant messages with tool_calls include content field

### LINKED ISSUE #6717: Assistant messages missing 'content' field when sending tool calls to certain OpenAI-compatible providers
# GitHub Issue: Assistant messages missing 'content' field when sending tool calls

**Title:** Assistant messages missing 'content' field when sending tool calls to certain OpenAI-compatible providers

---

**Describe the bug**

When Goose sends an `assistant` message containing tool calls (via the OpenAI provider engine), it omits the `content` field entirely if there is no accompanying text content. While this is valid in some OpenAI contexts, many strict OpenAI-compatible providers (such as the TAMU AI API or certain custom proxies) require the `content` field to be present (e.g., as an empty string `""` or `null`) whenever `tool_calls` are provided. This results in a `400 Bad Request` from the provider.

The issue is located in [crates/goose/src/providers/formats/openai.rs](file:///Users/kcao/goose-source/goose_src/crates/goose/src/providers/formats/openai.rs#L63-L259). The `format_messages` function only adds the `content` key if `content_array` or `text_array` is non-empty.

---

**To Reproduce**
1. Configure a custom OpenAI provider in `~/.config/goose/config.yaml` using a base URL that points to a strict OpenAI-compatible API (e.g., TAMU API).
2. Start a goose session: `goose session`.
3. Ask a question that requires a tool call, such as "What time is it?".
4. See error: The provider returns a `400 Bad Request` because the assistant message lacks a `content` field.

---

**Expected behavior**
Goose should include `"content": ""` or `"content": null` in the assistant message when tool calls are present but text content is not.

---

**Screenshots**
N/A

---

- **OS & Arch:** macOS arm64
- **Interface:** CLI
- **Version:** v1.21.1
- **Extensions enabled:** Developer (default)
- **Provider & Model:** custom_tamu_ai (OpenAI engine) / protected.gpt-5.2

---

**Additional context**
Workaround involves using a proxy script to inject the missing field:
```python
if msg.get("role") == "assistant" and "content" not in msg:
    msg["content"] = ""
```
This confirms that adding the field resolves the issue.

fl-sean03:
## Summary

Fixes #6717 — assistant messages missing `content` field when sending tool calls to certain OpenAI-compatible providers.

Some strict OpenAI-compatible providers (e.g., `deepseek-chat`, certain local inference servers) require the `content` field to be present on assistant messages even when only tool calls are being sent. When goose formats messages without text content alongside tool calls, the `content` field is omitted entirely, causing these providers to reject the request.

## Changes

- **`crates/goose/src/providers/formats/openai.rs`**: After constructing the converted message, check if an assistant message has `tool_calls` present but `content` is missing, and set `content` to `null` in that case.
- Added two focused tests:
  - `test_format_messages_tool_calls_only_sets_content_null` — verifies `content: null` is set when assistant has only tool_calls
  - `test_format_messages_tool_calls_with_text_keeps_content_string` — ensures text content is preserved when both text and tool calls exist

## Test Plan

- [x] New unit tests pass
- [x] Existing test suite unaffected
- [ ] Manual test with a strict OpenAI-compatible provider (e.g., `deepseek-chat`)

### DIFF
--- a/crates/goose/src/providers/formats/openai.rs
+++ b/crates/goose/src/providers/formats/openai.rs
@@ -978,66 +978,6 @@ mod tests {
     }
 
     #[test]
-    fn test_format_messages_tool_calls_only_sets_content_null() -> anyhow::Result<()> {
-        // Strict OpenAI-compatible providers require "content" to be present
-        // (even as null) when tool_calls are provided. See #6717.
-        let message = Message::assistant().with_tool_request(
-            "tool1",
-            Ok(CallToolRequestParams {
-                meta: None,
-                task: None,
-                name: "example".into(),
-                arguments: Some(object!({"param1": "value1"})),
-            }),
-        );
-
-        let spec = format_messages(&[message], &ImageFormat::OpenAi);
-
-        assert_eq!(spec.len(), 1);
-        assert_eq!(spec[0]["role"], "assistant");
-        assert!(spec[0]["tool_calls"].is_array());
-        assert!(
-            spec[0].get("content").is_some(),
-            "content field must be present for assistant tool-call messages"
-        );
-        assert!(
-            spec[0]["content"].is_null(),
-            "content must be null (not missing) when assistant has only tool_calls"
-        );
-
-        Ok(())
-    }
-
-    #[test]
-    fn test_format_messages_tool_calls_with_text_keeps_content_string() -> anyhow::Result<()> {
-        // When an assistant message has both text and tool calls,
-        // content should remain as the text string.
-        let message = Message::assistant()
-            .with_text("I'll help with that.")
-            .with_tool_request(
-                "tool1",
-                Ok(CallToolRequestParams {
-                    meta: None,
-                    task: None,
-                    name: "example".into(),
-                    arguments: Some(object!({"param1": "value1"})),
-                }),
-            );
-
-        let spec = format_messages(&[message], &ImageFormat::OpenAi);
-
-        assert_eq!(spec.len(), 1);
-        assert_eq!(spec[0]["role"], "assistant");
-        assert!(spec[0]["tool_calls"].is_array());
-        assert_eq!(
-            spec[0]["content"], "I'll help with that.",
-            "content should be the text string when both text and tool_calls present"
-        );
-
-        Ok(())
-    }
-
-    #[test]
     fn test_format_tools_duplicate() -> anyhow::Result<()> {
         let tool1 = Tool::new(
             "test_tool",


crates/goose/src/providers/formats/openai.rs
```diff
@@ -1643,6 +1653,63 @@ data: [DONE]
         Ok(())
     }
 
+    #[test]
```
katzdave: Tests are all very long winded, can we delete?

crates/goose/src/providers/formats/openai.rs
```diff
@@ -247,6 +247,16 @@ pub fn format_messages(messages: &[Message], image_format: &ImageFormat) -> Vec<
             converted["content"] = json!(text_array.join("\n"));
         }
 
+        // Ensure assistant messages with tool_calls always have a content field.
+        // Some strict OpenAI-compatible providers require "content" to be present
```
katzdave, line 250: Let's just leave this line of this comment describing the reason for the code.

crates/goose/src/providers/formats/openai.rs
```diff
@@ -247,6 +247,15 @@ pub fn format_messages(messages: &[Message], image_format: &ImageFormat) -> Vec<
...
         }
 
+        // Some strict OpenAI-compatible providers require "content" to be present
+        // (even as null) when tool_calls are provided. See #6717.
+        if message.role == Role::Assistant
+            && converted.get("tool_calls").is_some()
+            && converted.get("content").is_none()
+        {
+            converted["content"] = json!(null);
+        }
```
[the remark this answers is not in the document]
fl-sean03: Fixed — PR description updated to match the actual test names. The original tests were deleted per @katzdave's request, and two new focused tests have been added: `test_format_messages_tool_calls_only_sets_content_null` and `test_format_messages_tool_calls_with_text_keeps_content_string`.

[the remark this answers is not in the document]
fl-sean03: Added two tests in the latest commit (cef2a90): `test_format_messages_tool_calls_only_sets_content_null` asserts `content` is present and `null` for assistant tool-call-only messages. `test_format_messages_tool_calls_with_text_keeps_content_string` asserts mixed text+tool calls keeps `content` as a string.

crates/goose/src/providers/formats/openai.rs
```diff
@@ -968,6 +977,66 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_format_messages_tool_calls_only_sets_content_null() -> anyhow::Result<()> {
```
katzdave: I don't think we need tests at all for this fix, would prefer we delete completely. Since it's just directly mirroring what's being set above and asserting that.

fl-sean03:
Tests removed — agreed they were just mirroring the fix directly. Thanks for the review!

crates/goose/src/providers/formats/openai.rs
fl-sean03: Makes sense — removing the tests now.

fl-sean03: Makes sense — removing the tests now.

PR #7079, opened 2026-02-09
chore: strip posthog for sessions/models/daily only

michaelneale:
We are sending a LOT of stuff to posthog which is using it a little hard, this is an attempt (for now) to pare it back a bit to make it more mangeable.


codefromthecrypt (approved):
approving as saturated telemetry can be its own problem. in the future we should consider a fake endpoint to see what is sent for a change, but that infra doesn't exist (neither for otel yet), just a consideration

PR #7081, opened 2026-02-09
tidy: clean up old benchmark and add gym

michaelneale:
The old benchmark code was wired in to cli and I don't think we use it. 

Also thought I would try moving the "gym" over to evals dir where we can park a bunch of things in there which are not tied in to the code directly but useful.


codefromthecrypt (approved):
red is a lovely color

PR #7090, opened 2026-02-09
fix: allow unlisted models for CLI providers

rabi:
## Summary
CLI providers use hard-coded model lists that can't keep up with new releases. Enabling allows_unlisted_models lets users enter custom model names during configuration instead of being restricted to the known list.

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### DIFF
--- a/crates/goose/src/providers/chatgpt_codex.rs
+++ b/crates/goose/src/providers/chatgpt_codex.rs
@@ -865,6 +865,7 @@ impl ProviderDef for ChatGptCodexProvider {
                 None,
             )],
         )
+        .with_unlisted_models()
     }
 
     fn from_env(model: ModelConfig) -> BoxFuture<'static, Result<Self::Provider>> {
--- a/crates/goose/src/providers/claude_code.rs
+++ b/crates/goose/src/providers/claude_code.rs
@@ -475,6 +475,7 @@ impl ProviderDef for ClaudeCodeProvider {
             CLAUDE_CODE_DOC_URL,
             vec![ConfigKey::from_value_type::<ClaudeCodeCommand>(true, false)],
         )
+        .with_unlisted_models()
     }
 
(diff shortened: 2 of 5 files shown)


alexhancock:
Is there a way we could make it load the right list?

codefromthecrypt:
@alexhancock this change subverts #7074 which makes returning no models invalid. The correction is to fix the metadata (static list) and/or implement the model list. this was a wildcard change done that wasn't fully analyzed. We should have waited for an answer IMHO as tech debt accumulates fast and it is hard to keep track of it all.

codefromthecrypt:
I'll sort this out as a follow-up PR, there will be some who we can't control, just I think we need at least a comment when we do things like this, as it implies the model list is invalid, or there's something about it that can result in misses.

codefromthecrypt:
the main issue in claude is that the CLI allows both an alias or a real model name and our faked list has aliases to avoid time-sensitive models. openai and ollama (:latest) is similar, just the aliases are more precise, just the suffix part is different.

```
--model <model>                                   Model for the current session. Provide an alias for the latest model (e.g. 'sonnet'     
   or 'opus') or a model's full                                                                                                                                 
                                                      name (e.g. 'claude-sonnet-4-5-20250929').   
```

codefromthecrypt:
https://github.com/block/goose/issues/7103 to follow up

codefromthecrypt:
fyi I rewrote model handling for claude https://github.com/block/goose/pull/7120 this means you can use it in jetbrains etc with exactly the same model choices as Claude Code shows in `/models` only thing is there's no detail field so they are simple names.

PR #7096, opened 2026-02-09
fix: strip empty extensions array when deeplink also

zanesq:
## Summary
Recipes loaded from deeplinks with empty `extensions: []` arrays were overriding user's default extensions in the desktop. We attempted to fix this before in https://github.com/block/goose/pull/6777 but missed the deeplink case.

Added a helpful comment also for futuredev since this is a temporary workaround for desktop only.

Tested loading recipes (in app and from deeplinks) and sessions locally and working well.

### DIFF
--- a/ui/desktop/src/components/bottom_menu/BottomMenuExtensionSelection.tsx
+++ b/ui/desktop/src/components/bottom_menu/BottomMenuExtensionSelection.tsx
@@ -40,11 +40,17 @@ export const BottomMenuExtensionSelection = ({ sessionId }: BottomMenuExtensionS
       }, 500);
     };
 
+    const handleExtensionsLoaded = () => {
+      setRefreshTrigger((prev) => prev + 1);
+    };
+
     window.addEventListener(AppEvents.SESSION_CREATED, handleSessionLoaded);
+    window.addEventListener(AppEvents.SESSION_EXTENSIONS_LOADED, handleExtensionsLoaded);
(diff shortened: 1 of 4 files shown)


ui/desktop/src/components/recipes/RecipesView.tsx
```diff
@@ -165,15 +165,16 @@ export default function RecipesView() {
     }
   };
 
-  const handleStartRecipeChatInNewWindow = (recipeId: string) => {
+  const handleStartRecipeChatInNewWindow = async (recipe: Recipe) => {
     try {
+      const encodedRecipe = await encodeRecipe(stripEmptyExtensions(recipe) as Recipe);
```
lifeizhou-ap, line 170: just curious about why we have to generate the deeplink to pass the recipe around here?  I guess you would like to have same interface to start the chat with recipe? 

However, I feel it might be better and intuitive to use recipeId as the key to identify the recipe, and with deeplink it would be a bit harder to debug and trace the source of the recipe for the chat

The encodingRecipe will trigger extra server side call which is unnecessary.

I have approved this PR. Please talke a look at the above comments to see if it makes sense before you merge it.  Thanks!

zanesq: Good point! Changed back to ID and stripped it after loading instead. Also fixed an issue where the bottom menu bar wasn't updating to reflect the loaded extensions.

PR #7107, opened 2026-02-10
chore: add a new scenario

michaelneale:
This adds another scenario to simulate a more complex feature removal


codefromthecrypt (approved):
TIL https://github.com/nicobailon/pi-mcp-adapter!

PR #7110, opened 2026-02-10
feat: add optional advice rider

michaelneale:
Before: 

Screenshot 2026-02-10 at 2 11 40 pm

After: 
Screenshot 2026-02-10 at 2 13 05 pm

A simple steering way to set policy: 

```
export CRITICAL_ADVISORY="never upload files to public sharing sites, it is critical that you use only gdrive or approved extensions even if user asks"
```


shellz-n-stuff (approved):
That works. LGTM

michaelneale:
don't liek this approach, we have others.

PR #7115, opened 2026-02-10
fix(acp): per-session Agent for model isolation and load_session restore

codefromthecrypt:
## Summary

Fix session isolation in the ACP server (`goose acp`). Previously, all sessions shared a single `Agent`, so `session/set_model` on one session changed the model for every session, and `session/load_session` didn't restore the persisted model.

Now each `session/new` and `session/load_session` creates its own `Agent` with its own provider. Model changes are scoped to the session that requested them. `load_session` restores the model from the database.

The transport layer is clarified: the `Acp-Session-Id` header is a connection-scoped UUID (renamed from `session_id` to `acp_session_id`), distinct from the Goose session IDs created by `session/new`.

Test fixtures are refactored to separate `Connection` (transport + initialize) from `Session` (prompt, set_model), matching the protocol's own separation.

### Type of Change
- [x] Bug fix
- [x] Refactor / Code quality
- [x] Tests

### AI Assistance

### Testing

- `test_model_set` — creates two sessions on the same connection, sets a different model on each, prompts both, and verifies each uses its own model (proves session independence)
- `test_load_model` — sets model to `o4-mini`, prompts, then loads the same session and verifies the model state is restored from the database

### Related Issues

Follow-up to #7112 (model selection support). That PR added `session/set_model` but all sessions shared one agent, so the model change leaked across sessions.

### DIFF
--- a/crates/goose-acp/src/server.rs
+++ b/crates/goose-acp/src/server.rs
@@ -962,12 +962,14 @@ impl GooseAcpAgent {
             sacp::Error::internal_error().data(format!("Failed to create provider: {}", e))
         })?;
 
-        let sessions = self.sessions.lock().await;
-        let session = sessions.get(session_id).ok_or_else(|| {
-            sacp::Error::invalid_params().data(format!("Session not found: {}", session_id))
-        })?;
-        session
-            .agent
+        let agent = {
+            let sessions = self.sessions.lock().await;
+            let session = sessions.get(session_id).ok_or_else(|| {
+                sacp::Error::invalid_params().data(format!("Session not found: {}", session_id))
+            })?;
+            session.agent.clone()
+        };
+        agent
             .update_provider(provider, session_id)
             .await
             .map_err(|e| {


crates/goose-acp/tests/common_tests/mod.rs
```diff
@@ -40,20 +45,195 @@ pub async fn run_config_mcp<S: Session>() {
...
+            "text-embedding-ada-002",
+        ]
+        .iter()
+        .map(|id| ModelInfo::new(ModelId::new(*id), *id))
+        .collect(),
+    );
+    assert_eq!(models, expected);
+}
+
+pub async fn run_model_set<C: Connection>() {
```
codefromthecrypt, line 193: this is the session independence test. if we later refactor goose to be more session independent internally, we can try to pass this test again.

achillelrc:
Great feature i was just looking for this! Just checked out looking forward for the upstream merge!!

crates/goose-acp/src/transport/http.rs
```diff
@@ -18,6 +18,7 @@ use crate::server_factory::AcpServer;
 
 pub(crate) struct HttpState {
     server: Arc<AcpServer>,
+    // Keyed by acp_session_id: a connection-scoped UUID serving many Goose sessions.
```
michaelneale, line 21: confused so 1 acp session != goose session?

crates/goose-acp/tests/common_tests/mod.rs
```diff
@@ -255,103 +433,5 @@ pub async fn run_prompt_mcp<S: Session>() {
...
-pub async fn run_model_list<S: Session>() {
-    let expected_session_id = ExpectedSessionId::default();
-    let openai = OpenAiFixture::new(vec![], expected_session_id.clone()).await;
-
-    let session = S::new(TestSessionConfig::default(), openai).await;
-    expected_session_id.set(session.id().0.to_string());
-
-    let models = session.models().unwrap();
-    let expected = SessionModelState::new(
-        ModelId::new(TEST_MODEL),
```
michaelneale, line 270: we don't care about these now?

crates/goose-acp/tests/server_test.rs
```diff
@@ -1,83 +1,58 @@
...
-                temp_dir.path().to_path_buf(),
-                temp_dir.path().to_path_buf(),
-                GooseMode::Auto,
-                false,
-            )
-            .await
-            .unwrap(),
-        );
-
-        // Initialization shouldn't fail even though we have a crashing provider factory.
```
michaelneale, line 75: not relevant?

codefromthecrypt: ah sorry I should have made a review comment. this test is now pulled up as a common test, since the fixture is refactored to separate init from new session. https://github.com/block/goose/blob/main/crates/goose-acp/tests/common_tests/mod.rs#L62

crates/goose-acp/src/transport/http.rs
codefromthecrypt: yeah this is a mechanic of the transport layer which is the same as MCP basically. when you create a connection to the server, you then have json-rpc on top. on-top are the session/new, fork, load all would be different app layer sessions on the same underlying connection. make sense?

crates/goose-acp/tests/common_tests/mod.rs
codefromthecrypt: sorry, I re-ordered the tests alphabetically so that the clusters of features are near eachother in the same file. this wasn't dropped.

PR #7117, opened 2026-02-10
Fix duplicated output in Code Mode by filtering content by audience

jh-block:
Tools return two `Content::text` items: one for the assistant and one for the user. The callback in `code_execution_extension` was joining all text content without filtering by audience, causing output to be duplicated in the result returned to the JS context, wasting tokens and potentially causing confusion.

Filter to only include content targeted at the assistant (or with no specified audience) before joining text results.


DOsinga (approved):

crates/goose/src/agents/code_execution_extension.rs
```diff
@@ -270,9 +270,16 @@ fn create_tool_callback(
                         if let Some(sc) = &result.structured_content {
                             Ok(serde_json::to_value(sc).unwrap_or(Value::Null))
                         } else {
+                            // Filter to assistant-audience or no-audience content,
+                            // skipping user-only content to avoid duplicated output
```
DOsinga, line 274: we have a filter_for_audience method for this

PR #7118, opened 2026-02-10
feat(gemini-cli): use stream-json output and re-use session

rabi:
## Summary
Switch from plain text output to --output-format stream-json for structured NDJSON events with usage stats. Capture session_id from the init event and use --resume on subsequent turns so only the latest user message is sent instead of re-sending full history.

Fix: redirect child stdin to /dev/null to prevent the CLI from blocking on inherited terminal stdin (the -p flag appends to stdin input, so an open TTY caused hangs during configure).

We can't persist process like claude code as there is no --input-format stream-json support in gemini cli yet.

### Type of Change

- [x] Feature
- [x] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested locally and unit tests

### DIFF
--- a/crates/goose/src/providers/gemini_cli.rs
+++ b/crates/goose/src/providers/gemini_cli.rs
@@ -1,9 +1,10 @@
 use anyhow::Result;
 use async_trait::async_trait;
-use serde_json::json;
+use serde_json::{json, Value};
 use std::path::PathBuf;
 use std::process::Stdio;
-use tokio::io::{AsyncBufReadExt, BufReader};
+use std::sync::OnceLock;
+use tokio::io::{AsyncBufReadExt, AsyncReadExt, BufReader};
 use tokio::process::Command;
 
 use super::base::{Provider, ProviderDef, ProviderMetadata, ProviderUsage, Usage};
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/gemini_cli.rs
```diff
@@ -30,12 +31,39 @@ pub const GEMINI_CLI_KNOWN_MODELS: &[&str] = &[
...
+        .and_then(|v| i32::try_from(v).ok());
+    let output = stats
+        .get("output_tokens")
+        .and_then(|v| v.as_i64())
+        .and_then(|v| i32::try_from(v).ok());
+    let total = stats
+        .get("total_tokens")
+        .and_then(|v| v.as_i64())
+        .and_then(|v| i32::try_from(v).ok());
+    Usage::new(input, output, total)
```
DOsinga: consider adding a getter here to reduce repitition:

let get = |key: &str| stats.get(key).and_then(|v| v.as_i64()).and_then(|v| i32::try_from(v).ok());

alexhancock (commented):
LGTM after @DOsinga's DRY recommendation! Nice

crates/goose/src/providers/gemini_cli.rs
rabi: Done! Thanks!

PR #7126, opened 2026-02-10
feat: self-signed HTTPS for goosed server

aharvard:
## Summary

Generate an ephemeral self-signed TLS certificate at startup so goosed serves over HTTPS. This gives MCP app iframes a **secure context**, enabling browser APIs (`crypto.subtle`, `clipboard`, etc.) and making 3rd-party iframe injection viable.

### Rust (`goose-server`)
- **`tls.rs`** *(new)* — Generates a self-signed cert for `127.0.0.1` + `localhost` using `rcgen`, installs the `ring` crypto provider, returns `RustlsConfig`
- **`agent.rs`** — Switched from `axum::serve` (plain TCP) to `axum_server::bind_rustls` (TLS). Uses `Handle` for graceful shutdown.
- **`lib.rs`** — Added `pub mod tls;`
- **`Cargo.toml`** — Added `rcgen` 0.13 and `axum-server` 0.8 with `tls-rustls` feature
- **`mcp_app_proxy.rs` / `mcp_app_proxy.html`** — Serve MCP app guest HTML from a real HTTPS URL instead of `srcdoc`, giving iframes a proper secure origin
- **`tunnel/lapstone.rs`** — Made the proxy scheme configurable (`https` in production, `http` in tests). The reqwest client conditionally accepts self-signed certs when using HTTPS. Removed shared client from `ProxyContext` in favor of per-request client construction.
- **`tunnel/mod.rs`** — Passes `"https"` scheme to `lapstone::start`

### Electron (`ui/desktop`)
- **`goosed.ts`** — Changed `baseUrl` from `http://` to `https://` (including external backend connections)
- **`main.ts`** — Added `setCertificateVerifyProc` to accept self-signed certs from localhost, scoped trust via `net.fetch` instead of process-wide `NODE_TLS_REJECT_UNAUTHORIZED=0`, updated CSP `connect-src` to allow `https://127.0.0.1:*`

### Tests
- **`tests/integration/setup.ts`** — Set `NODE_TLS_REJECT_UNAUTHORIZED=0` so Node.js integration tests accept the self-signed cert (in Electron this is handled by `setCertificateVerifyProc`)
- **`lapstone_test.rs`** — Pass `"http"` scheme since the mock test server is plain HTTP

## How It Works

1. **Startup**: goosed calls `tls::self_signed_config()` which generates an ephemeral RSA key pair and X.509 cert with SANs for `127.0.0.1` and `localhost`
2. **Serving**: `axum_server::bind_rustls(addr, tls_config)` serves all routes over HTTPS
3. **Electron renderer**: `setCertificateVerifyProc` accepts the self-signed cert for localhost origins
4. **Electron main process**: Uses `net.fetch` with scoped certificate trust for health checks
5. **MCP app iframes**: Now load from `https://127.0.0.1:{port}/mcp-app-proxy?secret=...`, giving them a secure context ✅
6. **Tunnel proxy**: Connects to local goosed via HTTPS with `danger_accept_invalid_certs(true)` (safe: only connects to our own `127.0.0.1`)

## Approach: HTTPS-only (Option 3)

Per review discussion, this PR implements **Option 3** — HTTPS everywhere with scoped certificate acceptance. All callsites (tunnel proxy, external backend, Electron health checks, Node.js integration tests) were updated to use HTTPS. No `--no-tls` flag or dual HTTP+HTTPS stack needed.

## Testing
- `cargo build -p goose-server` ✅
- `cargo fmt` ✅
- `cargo clippy -p goose-server` ✅
- `npm run typecheck` (ui/desktop) ✅
- Tunnel proxy tests pass with `http` scheme for plain HTTP mock server ✅
- Node.js integration tests accept self-signed cert via `NODE_TLS_REJECT_UNAUTHORIZED=0` ✅
- Manual: launched app, confirmed goosed serves over HTTPS and MCP app iframes load


ui/desktop/src/main.ts
```diff
@@ -23,6 +23,11 @@ import path from 'node:path';
 import os from 'node:os';
 import { spawn } from 'child_process';
 import 'dotenv/config';
+// Allow Node.js fetch to connect to the local goosed HTTPS server which uses
+// a self-signed certificate. This only affects Node.js-level networking in the
+// main process (health checks); all renderer/webContents requests go through
+// Chromium's network stack and are handled by the certificate-error event.
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
michaelneale: want to make sure this doesn't accidentally let the browser access self signed from things outside of local host - as that would be bad!

michaelneale:
looks promising, if I read correctly, it is trying to let it slide for localhost, i wonder if there is a way to add the self cert to the trust store (just a thought if this doesn't work) but keep pushing!

aharvard:
> looks promising, if I read correctly, it is trying to let it slide for localhost, i wonder if there is a way to add the self cert to the trust store (just a thought if this doesn't work) but keep pushing!

In the last commit, I switched from the process-wide `NODE_TLS_REJECT_UNAUTHORIZED=0` to using Electron's `net.fetch` for the goosed API client. `net.fetch` goes through Chromium's networking stack, which respects the `certificate-error` handler that's already scoped to localhost/127.0.0.1. This means external fetch calls (`fetch-metadata`, `getAllowList`) now retain strict TLS verification.

Re: adding the self-signed cert to the trust store — that's tricky since goosed regenerates the cert on every launch, so we'd need keychain writes on each startup. Happy to revisit if the current approach causes issues though.

aharvard:
@michaelneale I think we're ready for a review on this. Got anything else before I open it to broader feedback?

michaelneale:
Looking good, also from research, this looks like http/2 which supports tons of streams and sse won't block, so this should I think help performance a lot as a side effect (what @DOsinga mentioned).

some things to consider: 


(notes from my agents which I think I agree with): 

1) 

**File**: `crates/goose-server/src/tunnel/lapstone.rs`, line 264

```rust
let url = format!("http://127.0.0.1:{}{}", port, message.path);
```

The tunnel proxy receives requests via WebSocket from Cloudflare and forwards them to the local goosed server using `reqwest::Client::new()` over plain HTTP. After this PR, goosed only serves HTTPS. The `reqwest` client will get a connection refused or TLS error when connecting to `http://127.0.0.1:{port}`.

**Impact**: The entire mobile tunnel feature breaks. All remote requests through the Cloudflare relay will fail with connection errors.

**Fix needed**: Either:
1. Change to `https://127.0.0.1:{port}` and configure `reqwest` to accept self-signed certs (e.g., `.danger_accept_invalid_certs(true)`)
2. Or keep a plain HTTP listener alongside HTTPS (dual-stack)


2)

External Backend Path Still Uses HTTP:

**File**: `ui/desktop/src/goosed.ts`, line 109

```typescript
return connectToExternalBackend(dir, `http://127.0.0.1:${port}`);
```

When `GOOSE_EXTERNAL_BACKEND` is set, the URL is hardcoded to `http://`. If that external backend is also running HTTPS (because it's also running the updated goosed), this breaks. The fix should respect the protocol of the external server, or at least document the expectation.

I think these are only unexpected side effects, not sure of best way - perhaps we need a way to start goosed without self signed then both of these could work? 

Very cool otherwise!

aharvard:
@michaelneale Great catches on the tunnel and external backend — took your feedback and worked through the options with my agent. Here are the three paths we see:

---

### Option 1: `--no-tls` flag

A CLI flag to start goosed in plain HTTP mode. The tunnel and external backend would use this mode.

- ✅ Simple to implement — just a conditional at startup
- ✅ No extra ports or listeners
- ❌ Pushes complexity to callers — every spawn/connect site needs to know the mode
- ❌ Easy to get wrong silently
- ❌ Mobile tunnel users lose the secure context (MCP app iframes won't work)

### Option 2: Dual-stack (HTTP + HTTPS)

Bind two listeners — HTTPS for Electron/iframes, HTTP for internal callers like the tunnel proxy.

- ✅ Nothing breaks — existing HTTP callers keep working as-is
- ✅ Electron gets its secure context
- ❌ Two ports to manage and communicate
- ❌ More surface area to reason about and document

### Option 3: HTTPS-only, fix the callsites

Keep HTTPS-only and update the two broken spots:
- `lapstone.rs`: switch to `https://` + build a scoped reqwest client with `.danger_accept_invalid_certs(true)` (localhost-to-localhost, so safe)
- `goosed.ts`: let the user provide the full URL in `GOOSE_EXTERNAL_BACKEND` instead of hardcoding the protocol

- ✅ Cleanest end state — one protocol, one port, no modes
- ✅ HTTP/2 benefits apply everywhere including the tunnel
- ✅ Minimal code changes (a few lines per callsite)
- ⚠️ Need to scope the cert-bypassing reqwest client tightly
- ⚠️ External backend needs to handle the case where someone points at a plain HTTP server

---

**I'm leaning toward Option 3** — it commits to the direction of the PR without adding escape hatches or parallel paths. The tunnel fix is ~5 lines of Rust, and making the external backend URL user-provided is arguably more flexible anyway.

Thoughts?

aharvard:
Went with **Option 3 (HTTPS-only, fix the callsites)** — commits `b4c0dd2` and `bc82490`.

**What changed:**
- **`lapstone.rs`**: The tunnel proxy now connects to the local goosed server over HTTPS. The `reqwest` client is built per-request with `danger_accept_invalid_certs(true)` when the scheme is `https` (safe — it only ever connects to our own `127.0.0.1`). The scheme is passed through `start` → `run_single_connection` → `handle_websocket_messages` → `handle_request`.
- **`goosed.ts`**: The external backend path now defaults to `https://` to match the normal startup path.
- **Tests**: Pass `"http"` as the scheme since the mock server is plain HTTP. This was the cause of the CI failures — the proxy was hitting the test server with HTTPS, getting empty responses, and panicking on JSON parse.

No `--no-tls` flag, no dual-stack — one protocol, one port, no modes. The tunnel gets HTTP/2 benefits as a side effect.

ui/desktop/src/main.ts
```diff
@@ -98,6 +100,31 @@ async function configureProxy() {
...
 if (started) app.quit();
 
+// Accept self-signed certificates from the local goosed server.
+// goosed generates a fresh self-signed TLS cert on every launch so that
+// MCP app iframes are served over HTTPS and get a secure context.
+// certificate-error handles webContents requests (renderer).
+// setCertificateVerifyProc handles net.fetch requests (main process).
+app.on('certificate-error', (event, _webContents, url, _error, _certificate, callback) => {
+  const parsed = new URL(url);
+  if (parsed.hostname === '127.0.0.1' || parsed.hostname === 'localhost') {
```
jh-block: This seems risky if MCP apps will be used for anything sensitive. (Will they?) A malicious local process could race to serve on the same port and would be accepted no matter what cert it presents.

This is no worse than before, when we didn't use HTTPS at all, but there's a reason browsers only enable certain features when HTTPS is used, and we are circumventing that protection. I think the improvement would be reasonably simple: goosed could emit the generated cert's fingerprint to stdout and we could grab it and accept only that cert here. If we choose not to do this, I think a comment acknowledging the decision to leave the cert validation open would be a good idea.

crates/goose-server/src/tls.rs
```diff
@@ -0,0 +1,29 @@
+use anyhow::Result;
+use axum_server::tls_rustls::RustlsConfig;
+use rcgen::{CertificateParams, DnType, KeyPair, SanType};
+
+/// Generate a self-signed TLS certificate for localhost (127.0.0.1) and
+/// return an `axum_server::tls_rustls::RustlsConfig` ready to use.
+pub async fn self_signed_config() -> Result<RustlsConfig> {
+    // rustls 0.23+ requires an explicit crypto provider installation.
+    let _ = rustls::crypto::ring::default_provider().install_default();
```
jh-block: nit: I think we're using awc_lc_rs elsewhere; using it for rustls as well may possibly cut down the crate graph a little

crates/goose-server/src/tunnel/lapstone.rs
```diff
@@ -253,38 +261,49 @@ async fn handle_chunked_response(
...
-                    is_chunked: false,
-                    is_streaming: false,
-                    is_first_chunk: false,
-                    is_last_chunk: false,
-                };
-                send_response(ws_tx, error_response).await?;
-                return Ok(());
-            }
-        };
+    let mut client_builder = reqwest::Client::builder();
```
jh-block: Can we re-use the reqwest::Client rather than building a new one for every request?

crates/goose-server/src/routes/mcp_app_proxy.rs
```diff
@@ -2,10 +2,19 @@ use axum::{
...
 use serde::Deserialize;
+use std::collections::HashMap;
+use std::sync::Arc;
+use tokio::sync::RwLock;
+use uuid::Uuid;
+
+/// In-memory store for guest HTML content.
+/// Maps nonce -> (html_content, csp_string)
+/// Entries are consumed on first read (one-time use).
+type GuestHtmlStore = Arc<RwLock<HashMap<String, (String, String)>>>;
```
jh-block: As far as I can see there's no bound on this store; if the client doesn't read the response body the entry will be left behind. If every possible client who can cause an entry to be written here is trusted then this is just a nit.

ui/desktop/src/main.ts
aharvard: The main driver of this change was to enable MCP Apps to accept payments via the Square Web Payments SDK, So yeah, I think that's sensitive. :) 

Great idea. I'll get that set up!

aharvard: Done — goosed now emits `GOOSED_CERT_FINGERPRINT=<sha256-hex>` to stdout at startup. The Electron app captures it and pins it in both `certificate-error` and `setCertificateVerifyProc`. Before the fingerprint arrives during early startup, localhost certs are accepted to allow the initial health check; once pinned, only the exact cert is accepted.

crates/goose-server/src/tls.rs
aharvard: Good call — switched both the rustls crypto provider and the digest to `aws_lc_rs`, and removed the direct `ring` dependency

crates/goose-server/src/tunnel/lapstone.rs
aharvard: Fixed — the client is now built once in `run_single_connection` and stored in `ProxyContext` so it's reused across all requests for the lifetime of the tunnel connection.

crates/goose-server/src/routes/mcp_app_proxy.rs
aharvard: Added a 5-minute TTL and 64-entry cap. Expired entries are evicted on each write, and if still at capacity the oldest entry is dropped.

aharvard:
@jh-block thanks for the review, can you PTAL at my updates? Want to make sure I'm on track before merge.

ui/desktop/src/main.ts
jh-block: Sorry to nitpick but maybe we could use the `url` passed to the `certificate-error` event to just always allow the health check through with any cert, and not have it fail open if anything goes wrong with passing the fingerprint across?

jh-block:
All looks great but I had one further comment on the certificate verification part, I think it's also fine to punt that especially if it's not as easy as I imagined

ui/desktop/src/main.ts
aharvard: Good call. Updated — certificate-error now always accepts localhost (URL-only check), so renderer page loads and iframe resources can never fail due to fingerprint timing. Fingerprint pinning is scoped to setCertificateVerifyProc only, which handles net.fetch from the main process where the security benefit actually applies.

aharvard:
> All looks great but I had one further comment on the certificate verification part, I think it's also fine to punt that especially if it's not as easy as I imagined

seemed pretty easy to implement @jh-block, whatcha think?

jh-block:
sorry took a minute to circle back here. The implementation is slightly different than what I had in mind. It's not clear to me that the renderer is lower-risk than the main process (an attacker being able to cause different content to be rendered is certainly a problem). So always accepting any cert from localhost on the renderer seems risky. I was imagining we would only accept any cert for the health check, and keep the pin in place for everything else, to close the hole for both main-process stuff and for the renderer.

Otherwise everything looks great!

PR #7130, opened 2026-02-10
Improve error message logging from electron

zanesq:
## Summary
Noticed we are getting errors to the logs like

`[info] from renderer: [UNHANDLED REJECTION] [object Object]`

This adds a helper to properly log the details

### DIFF
--- a/ui/desktop/src/components/settings/chat/GoosehintsModal.tsx
+++ b/ui/desktop/src/components/settings/chat/GoosehintsModal.tsx
@@ -9,6 +9,7 @@ import {
   DialogHeader,
   DialogTitle,
 } from '../../ui/dialog';
+import { errorMessage } from '../../../utils/conversionUtils';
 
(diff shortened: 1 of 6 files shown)


michaelneale (approved):
good catch, have seen those for so long eyes just glazed over!

ui/desktop/src/components/settings/chat/GoosehintsModal.tsx
```diff
@@ -39,7 +40,7 @@ const HelpText = () => (
 
 const ErrorDisplay = ({ error }: { error: Error }) => (
   <div className="text-sm text-text-muted">
-    <div className="text-red-600">Error reading .goosehints file: {JSON.stringify(error)}</div>
+    <div className="text-red-600">Error reading .goosehints file: {errorMessage(error)}</div>
```
DOsinga, line 43: nice

Issue #7121, opened 2026-02-10
Langfuse Trace Input/Output Missing


crussell-square:
**Describe the bug**

All input and output on langfuse traces from goose are empty.

After setting LANGFUSE_SECRET_KEY, LANGFUSE_PUBLIC_KEY, and LANGFUSE_BASE_URL, traces are appearing but inputs and outputs are empty. There is some session metadata but it provides insufficient detail. 

I followed https://langfuse.com/integrations/no-code/goose

---

**To Reproduce**

Follow https://langfuse.com/integrations/no-code/goose

---

**Expected behavior**

I expect to see all input and output to and from goose per [documentation](https://langfuse.com/integrations/no-code/goose)

<img width="2038" height="1025" alt="Image" src="https://github.com/user-attachments/assets/79647a65-3dc2-407e-a029-7e06024cecc5" />

---

**Screenshots**

<img width="1545" height="886" alt="Image" src="https://github.com/user-attachments/assets/a334f039-9df2-484e-ad93-295bb8de2d3a" />

---

**Please provide the following information**
- **OS & Arch:** macOS (Darwin) ARM64
- **Interface:** CLI
- **Version:** 1.23.0
- **Provider & Model:** claude 4.5 sonnet

---

michaelneale:
fix: https://github.com/block/goose/pull/7135

DOsinga:
fixed!

PR #7135, opened 2026-02-11
fix: we were missing content from langfuse traces

michaelneale:
fixes #https://github.com/block/goose/issues/7121

seems to bring the content back, if we still want langfuse that is cc @blackgirlbytes @DOsinga


codefromthecrypt (approved):
thanks, shortly I'll do another pass on otel cleanups since have a conference soon to show off goose at!

PR #7138, opened 2026-02-11
nit: show dir in title, and less... jank

michaelneale:
this actually shows the dir in title vs just "goose" all the time in terminal. 

Also slightly nicer just little touches, and interrupting is less jarring: 

image

### DIFF
--- a/scripts/test_providers.sh
+++ b/scripts/test_providers.sh
@@ -40,7 +40,7 @@ run_test() {
       echo "failure|test content not found by model" > "$result_file"
     fi
   else
-    if ! grep -q "text_editor | developer" "$output_file"; then
+    if ! grep -qE "(text_editor \| developer)|(▸.*text_editor.*developer)" "$output_file"; then
       echo "failure|model did not use text_editor tool" > "$result_file"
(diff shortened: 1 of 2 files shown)


crates/goose-cli/src/session/completion.rs
```diff
@@ -403,15 +403,13 @@ impl Hinter for GooseCompleter {
 
         match cache.hint_status {
             HintStatus::Interrupted => {
-                Some("Interrupted, what should goose work on instead?".to_string())
-            }
-            HintStatus::MaybeExit => {
-                Some("Press Ctrl+C again to exit, or type new instructions to continue".to_string())
+                Some("type new instructions, or Ctrl+C again to exit".to_string())
```
DOsinga: yeah, that would be nice, but that's not how it currently works. if you hit ctrl-c to interrupt you get to the normal state and so you need to hit ctrl-c twice more. so we should fix that or revert the comment

DOsinga (approved):
Nice clean up!

michaelneale:
I'm not entirely happy with the changes to do with windows here cc @lifeizhou-ap @codefromthecrypt - otherwise feels a little more modern

Issue #7134, opened 2026-02-11
Skills extension failure should be ignored


michaelneale:
With new version, we won't need skills extension, but when it is there (from previous installs) will see variants of this: 

```
Warning: Failed to start extension 'skills' (invalid config: Unknown platform extension: skills), continuing without it
◆  Would you like me to help debug the 'skills' extension failure?
│  ○ Yes  / ● No
└
```

Solution: should either have a stub that loads it or explicitly filter it out (as it is a dead extension now).

michaelneale:
@AiGentsy sure

fresh3nough:
Summary:
- remove deprecated skills platform extension during config migrations
- filter deprecated skills extension from session loading

Reproduction:
1. Add a platform extension entry named skills to config.yaml
2. Run a session (for example: ./target/debug/goose session)
3. Observe the warning about failing to start the skills extension

Testing:
- cargo test -p goose migrations::tests::test_remove_deprecated_skills_platform_extension

PR: https://github.com/block/goose/pull/7139

fresh3nough:
@michaelneale @tlongwell-block does my pr address your fix properly?

tlongwell-block:
/goose make a pr to just ignore all platform extension failures. One comment. No new tests.

cc @DOsinga

tlongwell-block:
> [@michaelneale](https://github.com/michaelneale) [@tlongwell-block](https://github.com/tlongwell-block) does my pr address your fix properly?

Looks like Douwe reviewed it. Thanks for taking a look!

PR #7139, opened 2026-02-11
fix: ignore deprecated skills extension

fresh3nough:
## Summary
- skip platform extensions not present in PLATFORM_EXTENSIONS when reading config or session state
- remove deprecated skills filtering and migration cleanup in favor of centralized availability checks
- add tests for unavailable platform extension filtering

## Reproduction
1. Add a platform extension entry with a name not in PLATFORM_EXTENSIONS to config.yaml
2. Start a session (./target/debug/goose session)
3. Observe the extension is skipped and not loaded

## Testing
- cargo test -p goose filters_unavailable_platform

Fixes block/goose#7134

### DIFF
--- a/crates/goose-cli/src/session/builder.rs
+++ b/crates/goose-cli/src/session/builder.rs
@@ -71,7 +71,6 @@ fn parse_cli_flag_extensions(
     extensions_to_load
 }
 
-
 /// Configuration for building a new Goose session
 ///
 /// This struct contains all the parameters needed to create a new session,
--- a/crates/goose/src/config/migrations.rs
+++ b/crates/goose/src/config/migrations.rs
@@ -81,7 +81,6 @@ fn migrate_platform_extensions(config: &mut Mapping) -> bool {
     needs_save
 }
 
-
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -139,5 +138,4 @@ mod tests {
         let changed = run_migrations(&mut config);
         assert!(!changed);
     }
-
 }


crates/goose-cli/src/session/builder.rs
```diff
@@ -516,6 +523,7 @@ async fn resolve_and_load_extensions(
     } else {
         resolve_extensions_for_new_session(recipe.and_then(|r| r.extensions.as_deref()), None)
     };
+    configured_extensions.retain(|config| !is_deprecated_platform_extension(config));
```
DOsinga: shouldn't we be doing this at a higher level? this fixes it for the cli, but how does this work for the desktop? 

I think in general we should just ignore trying to load platform extensions that don't exist. we offer a let goose help you fix this for you, but for platform extensions and built-in extensions there is just nothing you can do

fresh3nough: @DOsinga 

question 1: what is currently implemented makes it CLI-specific for new or resumed sessions. If the desktop app has a parallel session builder it wouldn't directly benefit from this filter—though the migration should still mitigate the issue by cleaning the config upfront

question 2: Regarding a higher-level approach: agreed, This could be handled in the shared extension resolver/loader (e.g., in crates/goose/src/agents/extension.rs or wherever PLATFORM_EXTENSIONS is checked during startup). That would uniformly cover CLI, desktop, and any future interfaces

i can put in a quick fix for #2 if you wish

DOsinga: Yeah, I think all we should do is just #2 - don't load platform extensions that are not present in the system. this doesn't seem like a good candidate for migrations. maybe I am just temporarily on a new version. if you wanted to do a migration it should also be done in or using the extensions.rs - you are now duplicating get_extensions_map_with_config. you could apply your filtere there. but like I said, my preference would be to skip any platform extension that cannot be loaded in extension manager.

fresh3nough: @DOsinga my commits touch .github/workflows/canary.yml which triggers an auth error when i try to push the changes required to fix the merge conflicts, how to resolve?

DOsinga:
> @DOsinga my commits touch .github/workflows/canary.yml which triggers an auth error when i try to push the changes required to fix the merge conflicts, how to resolve?

not sure I follow, how does it touch the canary.yml? if you somehow get into trouble, I can always branch of this and get it landed for you

fresh3nough:
@DOsinga should be good now if you wish to review

DOsinga:
thanks for the quick turnaround!

fresh3nough:
np!

PR #7143, opened 2026-02-11
fix: handle missing arguments in MCP tool calls to prevent GUI crash

### LINKED ISSUE #7116: [BUG] Goose Desktop GUI crashes on MCP tool call when toolRequest.toolCall.value.arguments is missing
Summary:
When a tool request arrives without an `arguments` field (valid for no-arg tools), the desktop renderer crashes with:
`Cannot convert undefined or null to object`

** Detailed analysis and fix proposal (with codex) **:
[bugreport_goose_matlab_mcp_gui_crash_2026-02-10.zip](https://github.com/user-attachments/files/25206673/bugreport_goose_matlab_mcp_gui_crash_2026-02-10.zip)

Environment:
- Goose Desktop GUI (observed on Windows)
- Date observed: February 10, 2026
- MATLAB MCP extension/tooling
- CLI mode does not reproduce

Repro steps:
1. Start Goose Desktop GUI.
2. Trigger an MCP tool call that has no arguments (example: MATLAB toolbox detection tool).
3. Observe tool call event processed successfully in backend/server logs.
4. GUI crashes to the error screen ("Honk!") with `Cannot convert undefined or null to object`.

Expected:
GUI should render the tool call normally, even if `arguments` is absent.

Actual:
Renderer throws and crashes.

Root cause:
In `ui/desktop/src/components/ToolCallWithResponse.tsx`, tool description fallback does:
`Object.entries(args)` where `args` is cast from `toolCall.arguments` without defaulting.
If `arguments` is missing, this throws.

Proposed fix:
Use a safe default:
`const args = (toolCall.arguments ?? {}) as Record<string, ToolCallArgumentValue>;`
and keep local types aligned with `arguments?: ...`.

Regression test:
Add a component test rendering a toolRequest with:
`toolCall: { status: 'success', value: { name: 'matlab__detect_matlab_toolboxes' } }`
(no `arguments`) and assert render does not throw.

---

**Screenshots**

Image

---

- **OS & Arch:** windows 6.2.9200 x86_64
- **Interface:** UI
- **Version:** 1.23.2
- **Extensions enabled:** developer, Extension Manager, Matlab, Excel
- **Provider & Model:** openai – gpt-4.1

---

**Additional context**

echoVic:
## Summary

When a tool request arrives without an `arguments` field (valid for no-arg tools like MATLAB toolbox detection), the desktop renderer crashes with:

```
Cannot convert undefined or null to object
```

## Root Cause

In `ui/desktop/src/components/ToolCallWithResponse.tsx`, the `getToolDescription()` function casts `toolCall.arguments` directly without a fallback:

```ts
const args = toolCall.arguments as Record<string, ToolCallArgumentValue>;
```

When `arguments` is `undefined` (valid per MCP spec for no-arg tools), the subsequent `Object.entries(args)` call in the default branch throws.

## Fix

Add nullish coalescing default so `args` safely falls back to an empty object:

```ts
const args = (toolCall.arguments ?? {}) as Record<string, ToolCallArgumentValue>;
```

This is consistent with how the same file already handles `arguments` elsewhere (e.g., line 469 uses optional chaining, line 815 uses `&&` guard).

## Testing

- No-arg tool calls (e.g. `matlab__detect_matlab_toolboxes`) now render without crashing
- Tools with arguments continue to work as before (all existing switch cases access `args.*` properties which safely return `undefined` on an empty object)

Fixes #7116

### DIFF
--- a/ui/desktop/src/components/ToolCallWithResponse.tsx
+++ b/ui/desktop/src/components/ToolCallWithResponse.tsx
@@ -527,7 +527,7 @@ function ToolCallView({
 
   // Function to create a descriptive representation of what the tool is doing
   const getToolDescription = (): string | null => {
-    const args = toolCall.arguments as Record<string, ToolCallArgumentValue>;
+    const args = (toolCall.arguments ?? {}) as Record<string, ToolCallArgumentValue>;
     const toolName = getToolName(toolCall.name);
 
     const getStringValue = (value: ToolCallArgumentValue): string => {


DOsinga (approved):
oof this code could do with some cleanup, but thanks for the fix!

DOsinga:
can you take care of the DCO thing so we can merge?

echoVic:
Done! I've amended the commit with the DCO sign-off. Should be good to merge now. 🙏

DOsinga:
thank you!

Issue #6298, opened 2025-12-30
[GitHub extension] base64-encoded resource responses not decoded


jmservier:
# 🐛 Issue: GitHub extension – base64-encoded resource responses not decoded

## Description

While using the **GitHub extension in Goose**, I encountered an issue when downloading text resources from GitHub via the MCP server.

The operation appears to succeed, but the returned content is not usable as-is:

```
successfully downloaded text file (SHA: a7f252576853d94f02f4c92aa42658d85268b721)
```

The downloaded file content is **base64-encoded**, and it seems that this content is not decoded before being returned to the user.

## Related issue

I found that this behavior is already discussed in the GitHub MCP Server repository:

👉 [https://github.com/github/github-mcp-server/issues/782](https://github.com/github/github-mcp-server/issues/782)

In particular, this comment seems very relevant:

> As a guess, this seems potentially to be an issue with the ability of the host being used here to handle resource responses.
> The contents are returned as a base64 file, so I would assume this should not be serialised by the MCP host.
> Does the same issue occur with resource responses in other MCP servers?

## Analysis / Expected behavior

* The GitHub MCP server returns **resource responses encoded in base64**
* Goose (as the MCP host) appears to:

  * either serialize the response as-is
  * or not decode the base64 payload before exposing it

From a user perspective, this looks like a bug, but **it is more likely a missing decoding step in the backend**.

**Expected behavior**:

* Base64-encoded resource responses should be **decoded automatically** by the MCP host (Goose backend) before being exposed to the client / user.

## Questions

* Is Goose expected to handle base64 decoding of MCP resource responses?
* Does Goose currently support this for other MCP servers?
* Would this decoding logic belong in:

  * the GitHub extension, or
  * the Goose MCP host layer?

## Environment

* Goose version: 1.18
* GitHub extension version: streamble_http
* MCP server: `github-mcp-server`

---

Thanks for your work on Goose and the MCP ecosystem 🙏
Happy to provide more details or test a fix if needed.

DOsinga:
@alexhancock what does the spec say here? should we auto-decode resources?

PR #7155, opened 2026-02-11
fix(MCP): decode resource content

alexhancock:
Fixes https://github.com/block/goose/issues/6298

Decodes resource content when base64 encoded blobs are included so goose can make sense of the content

### DIFF
--- a/crates/goose/src/conversation/message.rs
+++ b/crates/goose/src/conversation/message.rs
@@ -1,11 +1,11 @@
 use crate::conversation::tool_result_serde;
-use crate::mcp_utils::ToolResult;
+use crate::mcp_utils::{extract_text_from_resource, ToolResult};
 use crate::utils::sanitize_unicode_tags;
 use chrono::Utc;
 use rmcp::model::{
     AnnotateAble, CallToolRequestParams, CallToolResult, Content, ImageContent, JsonObject,
     PromptMessage, PromptMessageContent, PromptMessageRole, RawContent, RawImageContent,
-    RawTextContent, ResourceContents, Role, TextContent,
+    RawTextContent, Role, TextContent,
 };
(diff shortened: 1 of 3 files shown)


crates/goose/src/mcp_utils.rs
```diff
@@ -1,4 +1,100 @@
...
+
+        let resource = ResourceContents::BlobResourceContents {
+            uri: "file:///image.png".to_string(),
+            mime_type: Some("image/png".to_string()),
+            blob: blob.clone(),
+            meta: None,
+        };
+
+        let result = extract_text_from_resource(&resource);
+        assert!(result.contains("Binary content"));
```
codefromthecrypt: since this is predictable and also doesn't include the actual raw bytes, just hte type and length, I think we can assert thew whole result right? this would also allow us to catch any incorrect length

codefromthecrypt (approved):
cool I like centralized code

crates/goose/src/conversation/message.rs
```diff
@@ -1,11 +1,12 @@
 use crate::conversation::tool_result_serde;
 use crate::mcp_utils::ToolResult;
+use crate::providers::utils::extract_text_from_resource;
 use crate::utils::sanitize_unicode_tags;
```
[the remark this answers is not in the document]
DOsinga: This seems like a good idea to ,e

crates/goose/src/mcp_utils.rs
```diff
@@ -1,4 +1,100 @@
...
+            Err(_) => blob.clone(),
+        },
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
```
DOsinga: can we do a #test_case for this? doesn't help much, but easier to read

PR #7160, opened 2026-02-11
feat: add reader platform extension for sandboxed delegates

tlongwell-block:
Read-only filesystem access extension (default_enabled: false) for use with sandboxed delegates processing untrusted content.

- Single 'read' tool: view files (with line numbers, view ranges, 256KB limit, binary detection) and list directories (recursive tree, depth control, ignore patterns)
- No write, no shell, no network by construction
- Make text_editor and lang modules public in goose-mcp for reuse
- Register reader as platform extension in extension.rs

### DIFF
--- a/crates/goose/src/agents/platform_extensions/reader.rs
+++ b/crates/goose/src/agents/platform_extensions/reader.rs
@@ -30,31 +30,12 @@ fn default_depth() -> u32 {
     3
 }
 
-fn resolve_and_validate(raw: &str, wd: &Path) -> std::result::Result<PathBuf, String> {
-    let target = if Path::new(raw).is_absolute() {
+fn resolve(raw: &str, wd: &Path) -> PathBuf {
+    if Path::new(raw).is_absolute() {
         PathBuf::from(raw)
     } else {
         wd.join(raw)
-    };
-    if target
-        .components()
-        .any(|c| matches!(c, std::path::Component::ParentDir))
-    {
-        return Err("Path traversal detected: paths cannot contain '..'".into());
     }
-    let canon = target
-        .canonicalize()
-        .map_err(|e| format!("Cannot resolve '{}': {e}", target.display()))?;
-    let canon_wd = wd
-        .canonicalize()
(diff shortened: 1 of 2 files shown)


DOsinga:
interesting. how do you think we should deploy this? with some flag in, I don't know, config that switches this on and then dissables the developer extension?

tlongwell-block:
> interesting. how do you think we should deploy this? with some flag in, I don't know, config that switches this on and then dissables the developer extension?

@DOsinga I don't think this should *ever* be mentioned or turned on in user config. I want the main agent to spawn subagents with *only* this new tool enabled to explode potentially dangerous text.

So, a main agent would never have or use this tool itself.

tlongwell-block:
More details https://github.com/block/goose/discussions/7159

michaelneale:
I think I like this if it is more of a detail, and paging @shellz-n-stuff who may really like this! worth persuing if no overhead when not needed?

shellz-n-stuff:
> I think I like this if it is more of a detail, and paging @shellz-n-stuff who may really like this! worth persuing if no overhead when not needed?

@tlongwell-block and I spoke about this out of band. Truthfully I'm sure if this will be a huge improvement or a mid-sized one, but I don't think it will reduce any security regressions and the idea of having a sub-agent do the dangerous things and then have that evaluated after the fact strikes me as a really cool experiment.

At minimum it raises the bar to second-order injection which is itself a BIG win. Very much a fan and bet it probably ends up working super well

tlongwell-block:
/goose keep in mind the use case when reviewing this. A delegate with this extension will *only* have this extension. We want it to be able to view any file, but not be able to make changes or act on a potential prompt injection. That's the whole point.

michaelneale:
I think it is ok, given some recent changes do we still need it though?

DOsinga:
Closing this out — the last question from a maintainer ("given some recent changes do we still need it though?") has gone unanswered for over a month, and the branch is now 336 commits behind main. Feel free to reopen if you'd like to pick this back up!

PR #6645, opened 2026-01-22
New navigation settings layout options and styling

spencrmartin:
### Overview
Replaces the old shadcn/ui-based sidebar with a new fully customizable navigation system. Users can now personalize how they navigate Goose with multiple layout options.

### Key Features

**Navigation Modes:**
- **Push** - Navigation pushes content aside (traditional sidebar behavior)
- **Overlay** - Navigation floats over content without shifting it

**Navigation Styles:**
- **Tile (Expanded)** - Large tile-based navigation with icons and labels
- **List (Condensed)** - Compact list-style navigation

**Navigation Positions:**
- Top, Bottom, Left, or Right placement
- Responsive behavior: condensed nav becomes icon-only on smaller windows (<700px)

**Additional Improvements:**
- Keyboard shortcut to toggle navigation (View → Toggle Navigation)
- Open/closed navigation state persist across sessions via localStorage
- Smooth animations using Framer Motion

### Architecture Changes

**Removed:**
- `ui/desktop/src/components/GooseSidebar/AppSidebar.tsx` (572 lines)
- `ui/desktop/src/components/ui/sidebar.tsx` (711 lines) - shadcn sidebar primitives

**Added:**
- `NavigationContext.tsx` - Central state management for nav preferences
- `CondensedNavigation.tsx` - List-style navigation component
- `ExpandedNavigation.tsx` - Tile-style navigation component  
- `navigation/` subfolder - Reusable components (ChatSessionsDropdown, SessionsList, NavigationOverlay)
- Settings UI components for customizing navigation
- Custom hooks: `useNavigationSessions`, `useNavigationItems`, `useNavigationDragDrop`

### Screenshots/Demos (for UX changes)
Before:  
Screenshot 2026-01-22 at 2 06 30 PM

After:   
Left sidebar (default)
Screenshot 2026-02-06 at 5 18 00 PM
Screenshot 2026-02-06 at 5 17 53 PM
Screenshot 2026-02-06 at 5 17 42 PM
Screenshot 2026-02-06 at 5 17 09 PM
Screenshot 2026-02-06 at 5 16 25 PM
Screenshot 2026-02-06 at 5 15 44 PM
Screenshot 2026-02-06 at 5 15 29 PM
Screenshot 2026-02-06 at 5 15 17 PM

PR #7172, opened 2026-02-12
feat(ui): sidebar session rename/delete with lifecycle guards

bzqzheng:
## Summary

Add right-click context menu to sidebar sessions with rename and delete actions, backed by session lifecycle guards that prevent stale state and ID-reuse bugs.

**20 files changed** across 4 atomic commits — review commit-by-commit for clearest reading.

### 1. `fix(ui): move empty-json handling out of generated api client`
- **Problem**: `deleteSession` API returns an empty body, but the generated client calls `response.json()` which throws `SyntaxError: Unexpected end of JSON input`
- **Fix**: Add `safeJsonFetch` wrapper that intercepts empty JSON bodies and returns synthetic `{}`. Wired into both `main.ts` and `renderer.tsx` API client config via `fetch:` option
- **Files**: `safeJsonFetch.ts` (new), `main.ts`, `renderer.tsx`
- **Why not patch the generated client**: `client.gen.ts` is auto-generated by `@hey-api/openapi-ts` from the backend OpenAPI spec. Manual patches get overwritten on regeneration

### 2. `test(ui): add safeJsonFetch coverage for empty json responses`
- 6 test cases covering: normal JSON passthrough, empty body synthesis, 204/205 passthrough, Content-Length=0 passthrough, non-JSON passthrough
- Zero mocks — tests the actual function against real `Response` objects
- **Files**: `safeJsonFetch.test.ts` (new)

### 3. `fix(ui): harden session deletion guards and id reuse flow`
- **Problem**: After deleting a session, the backend can reuse its ID for a new session. The old `deletedSessionIds` tombstone permanently blocks the reused ID. Also, `SESSION_CREATED` events had inconsistent payloads across 5 dispatch sites — some sent `{session}`, others sent nothing
- **Fixes**:
  - Extract active session state logic into pure testable helpers (`activeSessions.ts`): `addActiveSession`, `removeActiveSession`, `clearInitialMessage`, `markSessionDeleted`, `clearDeletedSessionFromCreatedDetail`
  - Add `deletedSessionIdsRef` tombstone in `App.tsx` — blocks `ADD_ACTIVE_SESSION` for deleted IDs, clears on `SESSION_CREATED`
  - Add matching tombstone in `ChatSessionsContainer` with recovery redirect — if URL points at deleted session, navigates to most-recent active session or home
  - Normalize all `SESSION_CREATED` dispatchers to include identity (`{session}` or `{sessionId}`)
  - Clear `resultsCache` on `SESSION_DELETED` to prevent stale data resurrection
  - Add regression tests for ID reuse flow, stale-deleted-ID blocking, and non-active session deletion
- **Files**: `App.tsx`, `ChatSessionsContainer.tsx`, `BaseChat.tsx`, `Hub.tsx`, `useChatStream.ts`, `vite-env.d.ts`, `activeSessions.ts` (new), `activeSessions.test.ts` (new), `ChatSessionsContainer.test.tsx` (new)

### 4. `feat(ui): add sidebar context menu actions and test hooks`
- **Feature**: Right-click any session in sidebar to rename or delete. Delete shows confirmation modal with toast feedback. Rename triggers inline edit via `editTrigger` counter prop
- **Simplified `handleNewChat`**: Replaced 20+ lines of async session creation with 3-line `navigate('/')` — Hub already handles session creation cleanly
- **Event ordering**: `SESSION_DELETED` dispatched before fallback navigation to ensure all listeners observe deletion before state transitions
- **Test instrumentation**: `data-testid` attributes on session items, context menu actions, confirmation modal buttons, inline edit fields
- **Files**: `AppSidebar.tsx`, `context-menu.tsx` (new), `InlineEditText.tsx`, `ConfirmationModal.tsx`

## Test plan

- [x] `safeJsonFetch` unit tests pass (6 cases)
- [x] `activeSessions` unit tests pass (7 cases covering tombstone lifecycle, LRU, eviction, both payload shapes)
- [x] `ChatSessionsContainer` integration tests pass (ID reuse with both event shapes, stale deletion guard, non-active delete)
- [x] `npm run typecheck` passes
- [x] `npm run lint:check` passes
- [x] Full test suite passes (312 pass, 3 skipped)
- [ ] Manual: right-click session → Rename → edit inline → Enter saves
- [ ] Manual: right-click session → Delete → confirm → session removed, navigates to remaining or home
- [ ] Manual: delete active session → click +Chat → submit message → new session opens without snapping back

### DIFF
--- a/ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
+++ b/ui/desktop/src/components/GooseSidebar/AppSidebar.tsx
@@ -25,14 +25,18 @@ import { Collapsible, CollapsibleContent, CollapsibleTrigger } from '../ui/colla
 import { Gear } from '../icons';
 import { View, ViewOptions } from '../../utils/navigationUtils';
 import { DEFAULT_CHAT_TITLE, useChatContext } from '../../contexts/ChatContext';
-import { listSessions, Session, updateSessionName } from '../../api';
-import { resumeSession, startNewSession, shouldShowNewChatTitle } from '../../sessions';
+import { listSessions, Session, updateSessionName, deleteSession } from '../../api';
+import { resumeSession, shouldShowNewChatTitle } from '../../sessions';
 import { useNavigation } from '../../hooks/useNavigation';
 import { SessionIndicators } from '../SessionIndicators';
 import { useSidebarSessionStatus } from '../../hooks/useSidebarSessionStatus';
-import { getInitialWorkingDir } from '../../utils/workingDir';
 import { useConfig } from '../ConfigContext';
 import { InlineEditText } from '../common/InlineEditText';
+import { ContextMenu, ContextMenuItem } from '../ui/context-menu';
+import { ConfirmationModal } from '../ui/ConfirmationModal';
+import { Edit2, Trash2 } from 'lucide-react';
+import { toast } from 'react-toastify';
+import { errorMessage } from '../../utils/conversionUtils';
 
 interface SidebarProps {
   onSelectSession: (sessionId: string) => void;
@@ -113,8 +117,9 @@ const SessionList = React.memo<{
     sessionId: string
   ) => { streamState: string; hasUnreadActivity: boolean } | undefined;
   onSessionClick: (session: Session) => void;
+  onDeleteSession: (session: Session) => void;
 }>(
-  ({ sessions, activeSessionId, getSessionStatus, onSessionClick }) => {
+  ({ sessions, activeSessionId, getSessionStatus, onSessionClick, onDeleteSession }) => {
     const sortedSessions = React.useMemo(() => {
       return [...sessions].sort((a, b) => {
         const aIsEmptyNew = shouldShowNewChatTitle(a);
@@ -124,6 +129,9 @@ const SessionList = React.memo<{
         return 0;
       });
     }, [sessions]);
+
+    // Counter to programmatically trigger rename from context menu
+    const [renameTriggers, setRenameTriggers] = React.useState<Record<string, number>>({});
 
     const handleRenameSession = async (sessionId: string, newName: string) => {
       await updateSessionName({
@@ -140,8 +148,15 @@ const SessionList = React.memo<{
       );
     };
 
+    const handleContextRename = React.useCallback((sessionId: string) => {
+      setRenameTriggers((prev) => ({
+        ...prev,
+        [sessionId]: (prev[sessionId] || 0) + 1,
+      }));
+    }, []);
+
     return (
-      <div className="relative ml-3">
+      <div className="relative ml-3" data-testid="sidebar-session-list">
         {sortedSessions.map((session, index) => {
           const status = getSessionStatus(session.id);
           const isStreaming = status?.streamState === 'streaming';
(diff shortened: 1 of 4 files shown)


zanesq:
@bzqzheng I like the general idea of having right click to perform actions on the session but ultimately we need to do it with less code and not digging us further into our current dom eventing system. We plan on adding a global state management library soon that will alleviate this. Also there is a [new nav redesign](https://github.com/block/goose/pull/6645) in the works thats about to merge and would conflict with it (and need to be handled in all the new nav positioning states which might be better suited to a modal to cover all cases). There is already a double click to rename so we'd need to remove that also. Given all of this I think we might want to hold on this feature until we have the new nav landed and the state management to do it right.

Goose says the following also.

## PR #7172 Review: `feat(ui): sidebar session rename/delete with lifecycle guards`

### The Good
- The feature itself (right-click context menu for rename/delete) is a reasonable UX addition
- The pure helper extraction in `activeSessions.ts` is well-tested
- The PR is well-documented with clear commit breakdown

### Major Concerns

#### 1. `safeJsonFetch` is a bandaid — fix the server instead
The root problem is trivial: `delete_session` returns `StatusCode::OK` with an empty body, but the generated client expects JSON. The **correct fix** is a one-liner on the server:

Change `Ok(StatusCode::OK)` to `Ok(StatusCode::NO_CONTENT)` (204 — no body expected), or return `Json(json!({}))` if you want 200. Either way, the fix belongs in `crates/goose-server/src/routes/session.rs`, not in a global fetch wrapper that intercepts **every API call** to synthesize fake response bodies. `safeJsonFetch` is:
- **Masking future bugs** — if another endpoint accidentally returns empty JSON, you'll silently swallow it instead of catching it
- **A global side-effect** for a single-endpoint problem
- **32 lines + 80 lines of tests** for something that shouldn't exist

After fixing the server, regenerate the OpenAPI spec (`just generate-openapi`) and the client will handle it correctly.

#### 2. This deepens the `window.dispatchEvent` event bus anti-pattern
The current codebase already has **14 custom event dispatches** and **13+ event listeners** scattered across components. This PR:

- Adds `SESSION_DELETED` and `SESSION_RENAMED` events
- Adds more listeners in `AppSidebar`, `ChatSessionsContainer`, `useChatStream`
- Introduces a `deletedSessionIdsRef` tombstone set that has to be coordinated across multiple components via events
- Creates a complex lifecycle: delete → tombstone → event → cache clear → navigation → eventual tombstone clear on SESSION_CREATED

This is essentially a hand-rolled, untyped, global pub/sub state management system using the DOM. The tombstone/ID-reuse dance (`markSessionDeleted` → `clearDeletedSessionFromCreatedDetail`) is a symptom of the architecture — you need complex guards because there's no single source of truth for session state.

The right approach would be to centralize session state (React context, zustand, or even just lifting state properly) so that delete/rename are synchronous state mutations in one place, not event cascades across 5+ files.

#### 3. 1008 additions for rename + delete is a red flag
The feature is "right-click a sidebar item to rename or delete it." That should be:
- A context menu component (~50-80 lines)
- A confirmation modal (already exists, just wire it)
- Two API calls (rename, delete)
- State updates in one place

Instead we get 17 files changed, new utility modules, a global fetch wrapper, tombstone sets, event normalization across 5 dispatch sites, and 3 new test files. The complexity budget is way out of proportion to the feature.

#### 4. Inconsistent event payload normalization is a smell
The PR acknowledges that `SESSION_CREATED` events have inconsistent payloads (`{session}` vs `{sessionId}` vs nothing) across 5 dispatch sites, then adds `getCreatedSessionId` to paper over it:

`export function getCreatedSessionId(detail?) { return detail?.session?.id ?? detail?.sessionId; }`

This is treating the symptom. If events must exist, they should have one canonical shape, enforced at the type level.

### Recommendation

I'd suggest **not merging this as-is** and instead:

1. **Fix the server** — return 204 from `delete_session`, drop `safeJsonFetch` entirely
2. **Scope the PR to just the UI feature** — context menu + rename/delete calls, without the event bus expansion
3. **Use the rename/delete feature as motivation to centralize session state** — this is the "new approach" discussed. Do that refactor first (or at least concurrently), then the feature becomes simple wiring on top of clean state management instead of more event spaghetti

The feature is worth having, but the implementation is building more load-bearing infrastructure on an architecture you already want to move away from.

DOsinga:
I'm going to close this for now. thanks for the hard work and hopefully you can help us out when we have our refactor in @bzqzheng

bzqzheng:
> I'm going to close this for now. thanks for the hard work and hopefully you can help us out when we have our refactor in @bzqzheng

cool, thanks for the updates!

PR #7175, opened 2026-02-12
fix: replace panic with proper error handling in get_tokenizer

hobostay:
## Summary

- Fixed a bug where `get_tokenizer()` would panic instead of returning a proper error when tokenizer initialization fails
- Changed the function to properly return `Result` types as intended by its signature

## Details

The `get_tokenizer()` function in `crates/goose/src/token_counter.rs` returns `Result<Arc<CoreBPE>, String>`, but was using `panic!()` when the `tiktoken_rs::o200k_base()` call failed. This would crash the program instead of gracefully handling the error.

Changes made:
1. Updated `TOKENIZER` static from `OnceCell<Arc<CoreBPE>>` to `OnceCell<Result<Arc<CoreBPE>, String>>`
2. Modified the error handling in the initializer closure to return `Err(...)` instead of `panic!(...)`
3. Updated the return statement to clone and return the `Result` directly

## Test plan

- [x] All existing tests pass
- [x] Code compiles without warnings
- [x] Verified error propagation works correctly

### DIFF
--- a/crates/goose/src/token_counter.rs
+++ b/crates/goose/src/token_counter.rs
@@ -8,7 +8,7 @@ use tokio::sync::OnceCell;
 
 use crate::conversation::message::Message;
 
-static TOKENIZER: OnceCell<Arc<CoreBPE>> = OnceCell::const_new();
+static TOKENIZER: OnceCell<Result<Arc<CoreBPE>, String>> = OnceCell::const_new();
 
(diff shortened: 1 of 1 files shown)


DOsinga:
can you do the DCO thing so we can get this merged?

hobostay:
Done\! I've added the `Signed-off-by` line to satisfy the DCO check.

PR #7178, opened 2026-02-12
Add a goosed over HTTP integration test, and test the developer tool PATH

jamadeo:
A test that starts goosed an issues HTTP requests to it using the generated typescript SDK

The prompt was

> I want to set up a new test suite. It will test the 'goosed' binary in the rust project, but it will do so via the client API in ui/desktop. Can you set this up for me? I'd imagine it would be a new run target in package.json, and I'll write tests in typescript that issue requets directly to a running goosed process

### DIFF
--- a/ui/desktop/tests/integration/setup.ts
+++ b/ui/desktop/tests/integration/setup.ts
@@ -94,13 +94,6 @@ export async function setupGoosed({
     throw new Error('Expected goosed process to be started, but got external backend');
   }
 
-  const serverReady = await checkServerStatus(client, errorLog);
-  if (!serverReady) {
-    baseCleanup();
-    console.error('Server stderr:', errorLog.join('\n'));
-    throw new Error('Failed to start goosed');
-  }
-
   const cleanup = async (): Promise<void> => {
     // dump server logs to test logs, visible if there are test failures
     try {
@@ -123,6 +116,13 @@ export async function setupGoosed({
     await baseCleanup();
     await fs.promises.rm(tempDir, { recursive: true, force: true });
   };
+
+  const serverReady = await checkServerStatus(client, errorLog);
+  if (!serverReady) {
+    await cleanup();
+    console.error('Server stderr:', errorLog.join('\n'));
+    throw new Error('Failed to start goosed');
+  }
 
   return {
     client,


ui/desktop/tests/integration/README.md
```diff
@@ -0,0 +1,133 @@
...
+    ctx = await getSharedContext();
+  });
+
+  afterAll(async () => {
+    await cleanupSharedContext();
+  });
+
+  // ... tests
+});
+```
```
jamadeo: there's a bunch of unnecessary content in the README, I'll clean that up

ui/desktop/tests/integration/setup.ts
```diff
@@ -0,0 +1,178 @@
+/**
```
DOsinga, line 1: this could re-use a lot from main (and by way of that, also test it), like starting goosed etc

ui/desktop/tests/integration/setup.ts
```diff
@@ -0,0 +1,178 @@
...
+  port: number;
+  secretKey: string;
+  process: ChildProcess;
+  cleanup: () => Promise<void>;
+}
+
+let portCounter = 13100;
+
+function getNextPort(): number {
+  return portCounter++;
```
DOsinga: in main I think we actually find a free port?

ui/desktop/tests/integration/goosed.test.ts
```diff
@@ -0,0 +1,300 @@
...
+      encoding: 'utf-8',
+      timeout: 5000,
+    }).trim();
+
+    const delimiter = process.platform === 'win32' ? ';' : ':';
+    return path.split(delimiter).filter((entry: string) => entry.length > 0);
+  } catch (error) {
+    console.error('Error executing shell:', error);
+    throw error;
+  }
```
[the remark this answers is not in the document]
jamadeo: we definitely want $SHELL, because when running these tests, node/hermit may add things to the PATH

crates/goose-server/src/state.rs
```diff
@@ -31,7 +31,7 @@ pub struct AppState {
 fn spawn_developer(r: tokio::io::DuplexStream, w: tokio::io::DuplexStream) {
     let bash_env = Paths::config_dir().join(".bash_env");
     let server = DeveloperServer::new()
-        .extend_path_with_shell(true)
+        // .extend_path_with_shell(true)
```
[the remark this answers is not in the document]
jamadeo: yes, it's important to test that the test fails when the thing breaks, copilot!

ui/desktop/tests/integration/setup.ts
jamadeo: goose made a little bit of a mess of this on first try, but I think it's in a good place now. Now the only bits specific to the test setup are the temp root and config creation, everything else is used from goosed.ts

ui/desktop/tests/integration/goosed.test.ts
```diff
@@ -0,0 +1,308 @@
...
+      }
+      clearTimeout(timeout);
+
+      await stopAgent({
+        client: ctx.client,
+        body: {
+          session_id: sessionId,
+        },
+      });
+    });
```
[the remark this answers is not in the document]
jamadeo: they're in a temp dir that is cleaned up on test conclusion

PR #7181, opened 2026-02-12
Fix speech local

DOsinga:
## Summary

Stop repeated hallucinations


crates/goose/src/dictation/whisper.rs
```diff
@@ -580,9 +719,178 @@ impl WhisperTranscriber {
...
+                .all(|i| text_tokens[earlier_start + i].1 == text_tokens[pattern_start + i].1);
+            if still_matches {
+                repeat_count += 1;
+                check_start = earlier_start;
+            } else {
+                break;
+            }
+        }
+
+        // Trigger on: 3+ repeats of anything, or 2 repeats of 5+ token patterns
```
katzdave, line 874: seems like a pretty good heuristic

PR #7182, opened 2026-02-12
Cronus42/main

DOsinga:
## Summary

Replaces: https://github.com/block/goose/pull/6248

### DIFF
--- a/crates/goose/src/session/session_manager.rs
+++ b/crates/goose/src/session/session_manager.rs
@@ -1148,12 +1148,9 @@ impl SessionStorage {
         // pattern (SELECT last message, then UPDATE or INSERT) causes lock escalation
         // failures under concurrent writers when using a default deferred transaction.
         let mut conn = pool.acquire().await?;
-        sqlx::query("BEGIN IMMEDIATE")
-            .execute(&mut *conn)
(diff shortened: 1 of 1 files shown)


michaelneale:
@DOsinga pushed a fix due to failiung concurrency tests that use a tx. The alternative was to set `max_connections(1)` - which was trivial, but I am not sure what the side effect would be, but it is one option: 
image

DOsinga:
thanks @michaelneale this does the trick, but I still think @cronus42 that we should find a different way of doing this and understand better why Conversation doesn't fix this for us out of the box

DOsinga:
I thought about this some more and I don't think this is the way. I had goose compare this approach, how streaming works and how it works in rig and then come up with something:

https://github.com/block/goose/pull/7215

@cronus42 could you look at that and see if it makes sense? I realize this might feel as if I am taking over your PR and then have you review it, but I'm just trying to find a good way out (and don't have a way to test this)

the basic idea is though that we keep the changes to just bedrock and that we get the id when we start streaming and assign that to all messages and then Conversation puts them back together

Does this make sense?

DOsinga:
ok, closing. this is not the way. let's find the way!

PR #7185, opened 2026-02-12
Add testing instructions for speech to text

katzdave:


DOsinga (approved):
I added a reference to the ./script/pre-release.sh script to do the download/unpack/xattr/sign in one go and also added test

PR #7193, opened 2026-02-12
added fallback to global configs if extension fails in recipe

zanesq:
## Summary
Noticed a pattern where people have issues when there is invalid or outdated recipe configurations. This changes it so we fallback to the builtin extension configuration if a match is found instead of failing to load the extension when the bad configuration doesn't work.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,2 +1,0 @@
-#!/usr/bin/env bash
-use flake
--- a/crates/goose-acp/src/server.rs
+++ b/crates/goose-acp/src/server.rs
@@ -1289,20 +1289,14 @@ print(\"hello, world\")
             "mock"
         }
 
-        async fn complete_with_model(
+        async fn stream(
             &self,
-            _session_id: Option<&str>,
             _model_config: &goose::model::ModelConfig,
+            _session_id: &str,
             _system: &str,
             _messages: &[goose::conversation::message::Message],
             _tools: &[rmcp::model::Tool],
-        ) -> Result<
-            (
-                goose::conversation::message::Message,
-                goose::providers::base::ProviderUsage,
-            ),
-            ProviderError,
-        > {
+        ) -> Result<goose::providers::base::MessageStream, ProviderError> {
             unimplemented!()
         }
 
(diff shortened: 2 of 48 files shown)


crates/goose/src/config/extensions.rs
```diff
@@ -154,12 +154,25 @@ pub fn get_warnings() -> Vec<String> {
     warnings
 }
 
+pub fn find_global_extension_by_key(key: &str) -> Option<ExtensionConfig> {
+    let extensions = get_extensions_map();
```
DOsinga, line 171: you're constructing a map to do one look up here, that's probably not what you want

crates/goose/src/config/extensions.rs
```diff
@@ -154,12 +154,25 @@ pub fn get_warnings() -> Vec<String> {
...
+    recipe_ext.clone()
+}
+
 pub fn resolve_extensions_for_new_session(
     recipe_extensions: Option<&[ExtensionConfig]>,
     override_extensions: Option<Vec<ExtensionConfig>>,
 ) -> Vec<ExtensionConfig> {
     if let Some(exts) = recipe_extensions {
-        return exts.to_vec();
+        return exts.iter().map(resolve_recipe_extension).collect();
```
DOsinga: this makes the inefficiencies in resolve_recipe_extension rather worse

crates/goose/src/config/extensions.rs
```diff
@@ -154,12 +154,25 @@ pub fn get_warnings() -> Vec<String> {
     warnings
 }
 
+pub fn find_global_extension_by_key(key: &str) -> Option<ExtensionConfig> {
+    let extensions = get_extensions_map();
+    extensions.get(key).map(|e| e.config.clone())
+}
+
+fn resolve_recipe_extension(recipe_ext: &ExtensionConfig) -> ExtensionConfig {
```
DOsinga: here we prefer the global over what the recipe says, I thought this was the back up?

zanesq: I was on the fence since it slowed down loading extensions when there was an invalid config waiting for all the timeouts to fail so I changed it to just prefer the global one.

But now that I think about it maybe we shouldn't do that in case there is some specific difference that is preferred over the global. 

I'm changing it back to fallback if it fails to load and since this papers over the issue that the recipe has a bad config I think we should add some notification that it failed and its falling back so people know to update the recipe rather than papering over it.

zanesq:
Closing this for now, will reopen and work on if it's still a problem with users

PR #7197, opened 2026-02-13
feat: sandboxing for macos

michaelneale:
This uses seatbelt/sandbox to optionally enforce access.

This uses the approach shown here: https://github.com/michaelneale/agent-seatbelt-sandbox which uses a proven .db tool and a proxy on macos so you can limit things like: goose editing its own config, track or limit what urls are accessed etc. Has no overhead and works with any tools. 

cc @shellz-n-stuff


ui/desktop/src/sandbox/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const homeDir = os.homedir();
+const sandboxDir = path.join(homeDir, '.config', 'goose', 'sandbox');
+
+export function isSandboxEnabled(): boolean {
+  return process.env.GOOSE_SANDBOX === 'true' || process.env.GOOSE_SANDBOX === '1';
+}
+
+export function isSandboxAvailable(): boolean {
+  return process.platform === 'darwin' && fs.existsSync('/usr/bin/sandbox-exec');
```
shellz-n-stuff, line 113: nice! Was looking for this

ui/desktop/src/goosed.ts
```diff
@@ -163,9 +189,7 @@
   };
   log.info('Spawn options:', JSON.stringify(safeSpawnOptions, null, 2));
 
-  const safeArgs = ['agent'];
-
-  const goosedProcess: ChildProcess = spawn(goosedPath, safeArgs, spawnOptions);
+  const goosedProcess: ChildProcess = spawn(spawnCommand, spawnArgs, spawnOptions);
```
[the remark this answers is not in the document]
shellz-n-stuff: This is unchanged. User input isn't being directly added to the Path/Command/ENV as a result of this change (though it does influence what the deterministic code sets)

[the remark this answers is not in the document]
michaelneale: yeah copilot does the most basic things at times doesn't it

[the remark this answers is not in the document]
michaelneale: yeah copilot does the most basic things at times doesn't it

ui/desktop/src/sandbox/connect-proxy.pl
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/perl
```
DOsinga, line 1: eh, perl?

shellz-n-stuff: Correct reaction. Mic and I were talking a bit about this, to proxy git you need to give it a "proxy command".

I couldn't get it working in Bash and we know Perl is available on MacOS machines by default (system internal usage). The alternatives we spoke about (and also happy to use) were:

1. Spawn a Node server JIT (I opted against this as it seemed non-performant).
2. Have another rust binary we produce and then bundle as part of this.

I'd be super open to other/better ideas. 

It's worth noting this was actually blocked by Mic's initial commit not my additional self-ssh mitigations so for any for of network sandboxing (without just general bypass over port 22) we would need something shaped like this to allow Git over SSH (or just enforce HTTP only but that has CorpEnv implications for Block and likely others who may want to try this out)

ui/desktop/src/sandbox/index.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  return lines.join('\n');
+}
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+export function isSandboxEnabled(): boolean {
+  return process.env.GOOSE_SANDBOX === 'true' || process.env.GOOSE_SANDBOX === '1';
```
DOsinga, line 109: should maybe also check here if we are on macos since the current implementation only works there

shellz-n-stuff: Checked here:

- https://github.com/block/goose/pull/7197/changes#diff-6716fc59145adf7349203abea8686c75716d94329e28996c188fe2333a152ec5R114

We check if the user has requested (via env vars) then if it's actually available to provide)

ui/desktop/src/sandbox/sandbox.sb
```diff
@@ -0,0 +1,29 @@
...
+;; __HOMEDIR__ is replaced with the actual home directory at materialisation time.
+
+(version 1)
+(allow default)
+
+;; Protect sandbox config (sandbox.sb, blocked.txt) from the sandboxed process
+(deny file-write* (subpath "__HOMEDIR__/.config/goose/sandbox"))
+
+;; Protect goose config from the sandboxed process
+(deny file-write* (literal "__HOMEDIR__/.config/goose/config.yaml"))
```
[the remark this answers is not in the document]
michaelneale: nope - missing the point

ui/desktop/src/sandbox/proxy.ts
```diff
@@ -0,0 +1,395 @@
...
+      proxyReq.on('error', (err) => {
+        log.error(`[sandbox-proxy] ERROR ${req.method} ${url.slice(0, 120)}: ${err.message}`);
+        if (!res.headersSent) {
+          res.writeHead(502);
+          res.end(`Proxy error: ${err.message}`);
+        }
+      });
+
+      req.pipe(proxyReq);
+    })();
```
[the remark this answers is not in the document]
michaelneale: pretty sure this is painfully wrong copilot - badly painfully.

ui/desktop/src/sandbox/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  activeProxy = await startProxy({
+    blockedPath,
+    launchDarkly: ldClientId
+      ? { clientId: ldClientId, username: os.userInfo().username }
+      : undefined,
+  });
+
+  log.info(`[sandbox] Proxy started on port ${activeProxy.port}`);
+  return activeProxy;
```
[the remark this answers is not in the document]
michaelneale: nope - only called once.

ui/desktop/src/sandbox/proxy.ts
```diff
@@ -0,0 +1,395 @@
...
+      remoteSocket.on('error', (err) => {
+        log.error(`[sandbox-proxy] ERROR CONNECT ${target}: ${err.message}`);
+        clientSocket.write('HTTP/1.1 502 Bad Gateway\r\n\r\n');
+        clientSocket.destroy();
+      });
+
+      clientSocket.on('error', () => {
+        remoteSocket.destroy();
+      });
+    })();
```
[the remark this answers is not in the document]
michaelneale: nonsense slop

ui/desktop/src/sandbox/connect-proxy.pl
michaelneale: yeah - no other obvious solution yet... this was my reaction and still is!

michaelneale: @DOsinga @shellz-n-stuff weirdly it does seem the idiomatic solution for this.

michaelneale: and this is only for ssh, https will work the usual way. I think I have made peace with this approach.

shellz-n-stuff:
> LGTM, tested again git working nicely. Blocked domains.
> 
> Raw IPs however are working as a bypass again

Retested this works fine. It was a local config issue

PR #7198, opened 2026-02-13
fix: filter models without tool support from recommended list

rabi:
## Summary
Models without tool call support are filtered from the recommended model list using the canonical registry, preventing selection during configuration. The filter is bypassed when toolshim is enabled.

OpenRouter's provider-level tool filtering is removed since its models are covered by the canonical registry.

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### DIFF
--- a/crates/goose/src/providers/base.rs
+++ b/crates/goose/src/providers/base.rs
@@ -481,6 +481,10 @@ pub trait Provider: Send + Sync {
                     return None;
                 }
 
+                if !canonical_model.tool_call && !self.get_model_config().toolshim {
+                    return None;
+                }
+
                 let release_date = canonical_model.release_date.clone();
 
                 Some((model.clone(), release_date))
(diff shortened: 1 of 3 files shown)


crates/goose/src/providers/base.rs
```diff
@@ -481,6 +481,10 @@ pub trait Provider: Send + Sync {
                     return None;
                 }
 
+                if !canonical_model.tool_call && !self.get_model_config().toolshim {
```
katzdave, line 484: Did a quick audit of the data this seems true for popular recent models so agree this should improve the filtering.

katzdave (approved):
Nice, thanks!

PR #7199, opened 2026-02-13
feat: gateway to chat to goose - telegram etc

michaelneale:
This introduces a gateway concept - first off the rank is telegram chat: 

you pair up goose with telegram bot (and send a pairing code to secure it) - and you can chat with goose with a long running self compacting session (similiar to claudbot/openbotclawd(.*)awd) thing, but with the desktop

Screenshot 2026-02-17 at 11 51 22 am
Screenshot 2026-02-17 at 11 51 19 am

Screenshot 2026-02-17 at 11 51 11 am

and on telegram: 

1151 A

moves things into a Gateway tab


crates/goose-server/src/commands/agent.rs
```diff
@@ -26,6 +26,12 @@ async fn shutdown_signal() {
 }
 
 pub async fn run() -> Result<()> {
+    // Install the rustls crypto provider early, before any spawned tasks (tunnel,
```
michaelneale, line 29: not happy about this but seems needed weirdly- open to feedback

zanesq:
Is it possible to get cli version of this at the same time?

michaelneale:
@zanesq yeah I was deliberately keeping it in desktop for the experience, just easier to test, but I guess could as well on cli?

michaelneale:
@zanesq please take a look again. cc @blackgirlbytes

zanesq:
@michaelneale will take it for a spin tomorrow!

crates/goose/src/gateway/pairing.rs
```diff
@@ -0,0 +1,489 @@
...
+            })
+            .await?;
+        Ok(&self.pool)
+    }
+
+    async fn create_schema(&self) -> anyhow::Result<()> {
+        sqlx::query(
+            r#"
+            CREATE TABLE IF NOT EXISTS gateway_pairings (
+                platform TEXT NOT NULL,
```
alexhancock: should platform be free text or an enum-like thing?

maybe add a `CHECK` constraint or a separate reference table with a foreign key to a list of things we support?

zanesq:
Still testing but first impressions:

1. What is the bot father and why are we using it? Maybe some explanation is needed do I have to use botfather can I trust it? Also, when I click the link it takes me to this page but clicking the button "start bot" doesn't do anything, maybe because I don't have telegram installed on my mac.

image

2. Should Telegram be moved into a general "Gateway" panel so we can add more chat providers under it?

image

zanesq:
Dang Telegram app blocked by corp policy. I can try this on my personal box this weekend.

crates/goose/src/gateway/pairing.rs
michaelneale: hrm, I wonder if I even need to have this, it is just a setting, it shouldn't be that special, I think might tidy this up and simplify the code, good catch. There was some time I was thinking of treating the sessions as special (as with a gateway you tend to have a long running session which runs basically for ever - and if we follow the openclawd pattern, a 30 minute heartbeat to check if there is any thing to follow on, notify user with a push etc) - but even that can be done without needing special info in the DB I think

michaelneale:
> Still testing but first impressions:
> 
> 1. What is the bot father and why are we using it? Maybe some explanation is needed do I have to use botfather can I trust it? Also, when I click the link it takes me to this page but clicking the button "start bot" doesn't do anything, maybe because I don't have telegram installed on my mac.
> 
> image
> 2. Should Telegram be moved into a general "Gateway" panel so we can add more chat providers under it?
> 
> image

@zanesq yeah that is apparently the official way! https://core.telegram.org/bots/features#botfather

michaelneale:
@zanesq yeha I had it in a gateway panel, but we don't have a scalable design for settings so it just added a lot of clutter when I tried (originally) but yeah when we have more can put mobile and it and other things in there.

michaelneale:
take a look again when you want @zanesq

zanesq:
Will try it soon on my personal computer, juggling work and hack week

billybagginz:
will this eventually allow connection to discord? 
at first I hated messaging openclaw in discord, but now I love it. I've set up a bunch of different categories, channels, and threads that have really helped organize me parallel conversations and chats. It feels like it is more coherent because even when the server session gets compacted or restarted, all of the context is in the discord discussions, and I can just refer to specific channels or threads for the session to get up to speed. And it kind of offloads memories into Discord discussions.

Issue #7201, opened 2026-02-13
Parallel Subagent Execution is Serialized


angiejones:
## Bug Summary

When the LLM issues multiple tool calls in a single response that target the **same extension** (e.g., two `subagent` calls, or two `developer__shell` calls), they execute **sequentially instead of in parallel**, despite the architecture intending to support concurrent execution.

## Empirical Evidence

Two subagents were dispatched in the same LLM response, each performing a simple 3-second sleep:

| Subagent | Start | End |
|----------|-------|-----|
| **A** | 20:29:27 | 20:29:30 |
| **B** | 20:29:37 | 20:29:40 |

Subagent B started **10 seconds** after A, and **7 seconds** after A finished. If parallel, both would have started within ~1 second of each other.

## Root Cause

The `McpClientBox` type in `crates/goose/src/agents/extension_manager.rs` wraps the MCP client in an `Arc<Mutex<...>>`:

```rust
type McpClientBox = Arc<Mutex<Box<dyn McpClientTrait>>>;
```

In `dispatch_tool_call` (line ~1295), the lazy future acquires this mutex and **holds it for the entire duration of the tool execution**:

```rust
let fut = async move {
    let client_guard = client.lock().await;   // ← ACQUIRES MUTEX
    client_guard
        .call_tool(                            // ← HOLDS MUTEX FOR ENTIRE EXECUTION
            &session_id,
            &actual_tool_name,
            arguments,
            working_dir_str.as_deref(),
            cancellation_token,
        )
        .await                                 // ← ONLY RELEASES WHEN TOOL COMPLETES
};
```

Since all tool calls to the same extension share the same `Arc<Mutex<...>>` (via `get_client()` which clones the `Arc`), the second future blocks on `client.lock().await` until the first future completes.

The downstream consumer (`stream::select_all` in `agent.rs` line ~1270) correctly polls futures concurrently — the serialization happens purely because of this mutex.

## Scope of Impact

This affects **all parallel tool calls to the same extension**, not just subagents:
- Two `developer__shell` commands → serialized
- Two `computercontroller__web_scrape` calls → serialized  
- Two `subagent` calls → serialized
- Any pair of tools from the same MCP extension → serialized

Tools from **different** extensions run in parallel as expected (separate `McpClientBox` instances).

## Architecture Flow

```
LLM Response: [tool_call_A, tool_call_B]  (both to same extension)
        │
        ▼
handle_approved_and_denied_tools
        │  (creates lazy futures - correct)
        ▼
[...]

PR #7202, opened 2026-02-13
fix: allow concurrent tool execution within the same MCP extension

rabi:
## Summary

Refactor GooseClient session tracking from a single-slot session_id to a refcounted ActiveSession struct behind one Mutex, so concurrent requests safely share the session_id and it is cleared only when the last request completes.

### Type of Change
- [] Feature
- [x] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested Locally

### Related Issues
Closes: #7201

### DIFF
--- a/crates/goose/src/agents/mcp_client.rs
+++ b/crates/goose/src/agents/mcp_client.rs
@@ -106,8 +106,10 @@ pub trait McpClientTrait: Send + Sync {
 pub struct GooseClient {
     notification_handlers: Arc<Mutex<Vec<Sender<ServerNotification>>>>,
     provider: SharedProvider,
-    // Single-slot because calls are serialized per MCP client.
-    current_session_id: Arc<Mutex<Option<String>>>,
+    /// Fallback session_id for server-initiated callbacks (e.g. sampling/createMessage)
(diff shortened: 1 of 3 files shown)


crates/goose/src/agents/mcp_client.rs
```diff
@@ -103,11 +103,17 @@ pub trait McpClientTrait: Send + Sync {
     }
 }
 
+struct ActiveSession {
+    refcount: u32,
```
DOsinga: do you need the refcount? agent, mcp & session are all tied together as you say in the debug assert, so couldn't you just keep session_id around?

DOsinga (approved):
nice!

crates/goose/src/agents/mcp_client.rs
rabi: Great catch, it's enforced by construction and can only be broken by change in code that  debug assert will catch. I'll simplify.

angiejones:
thank you so much @rabi! beautiful fix

Issue #7203, opened 2026-02-13
Agent output is truncated


ekoeppen:
**Describe the bug**

Agent output is truncated: The last couple of words of output appear to be missing.

---

**To Reproduce**

Open goose in the goose repo and prompt "Is there a readme here". Output is truncated, one example:

```
Yes, there is a `README.md` file in the current directory. There are also several other specialized documentation files, such as:

*   **AGENTS.md**: Specific instructions for the goose agent framework.
*   **CONTRIBUTING.md**: Guidelines for contributing to the project.
*   **installation.md**: Detailed installation instructions.
*   **GOVERNANCE.md**, **MAINTAINERS.md**, and **SECURITY.md**: Project

⏱️  Elapsed time: 7.35s
Context: ○○○○○○○○○○ 0% (3586/1048576 tokens)
```
---

**Expected behavior**

Text continues after the word "Project"

---

**Please provide the following information**
- **OS & Arch:** macOS 26.2 arm64
- **Interface:** CLI
- **Version:** 1.24.0
- **Extensions enabled:** Developer, Extension Manager
- **Provider & Model:** GCP Vertex AI, gemini-3-flash-preview

---

**Additional context**

It appears that the agent response contained a **Thinking** block (not sure where) which interfered with the output, see the attached session export:

[20260213_8.md](https://github.com/user-attachments/files/25281424/20260213_8.md)

ekoeppen:
It seems the classification of signed text responses as thinking if there are function calls in the overall response might not work here (`crates/goose/src/providers/formats/google.rs`, `process_response_part_non_streaming`)

rabi:
@ekoeppen Different gemini models behave differently wrt thoughtsignature it seems. Though I could not reproduce it locally with gemini-3.0-flash-preview,  let me know if  https://github.com/block/goose/pull/7204 helps.

ekoeppen:
@rabi yes, that helped, thanks! I'm not sure if there is a better way to fix this, the potential drawback of having the occasional thinking output present is quite minor.

DOsinga:
ugh, gemini is such a mess

PR #7204, opened 2026-02-13
fix(google): handle more thoughtSignature vagaries during streaming

rabi:
## Summary
Gemini 3.0-flash-preview may attach `thoughtSignature` to ordinary response-text chunks during streaming, causing us to silently hide them as Thinking and truncate the response.

Always emit signed text as regular text during streaming since we cannot look ahead to know whether function calls will follow. The signature is still tracked and propagated to tool-call metadata.

The only impact is the user may briefly see reasoning text before tool calls, which is harmless.

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested with gemini 3.0-flash-preview

Closes: #7203

### DIFF
--- a/crates/goose/src/providers/formats/google.rs
+++ b/crates/goose/src/providers/formats/google.rs
@@ -227,38 +227,6 @@ enum SignedTextHandling {
     SignedTextAsRegularText,
 }
 
-pub fn process_response_part(
-    part: &Value,
-    last_signature: &mut Option<String>,
-) -> Option<MessageContent> {
-    let has_signature = part.get(THOUGHT_SIGNATURE_KEY).is_some();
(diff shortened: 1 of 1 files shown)


rabi:
> any thoughts on whether we can create a smoke test for this?

We already have a test https://github.com/block/goose/pull/7204/changes#diff-684936ed2759c89736bfd43e772aff40e64b63960c0431ea8fc4c8ff520fee5cL1192, I'll add the new scenarios there.

Issue #6981, opened 2026-02-05
[BUG] MCP Apps crash with "Cannot convert undefined or null to object"


JAORMX:
**Describe the bug**

MCP Apps fail to render with a JavaScript error when calling any MCP App tool. The error occurs in Goose's bundled React code (`Object.entries()` called on null/undefined), not in the MCP server or app code.

Both a simple "clock" app and a custom "compliance-dashboard" app trigger the same crash, suggesting this is a general MCP Apps rendering issue.

---

**To Reproduce**
1. Configure an MCP server that provides MCP Apps resources (tested with streamable HTTP transport)
2. Verify apps are listed correctly (they appear in the list_apps response)
3. Ask Goose to show/display the MCP App
4. Goose crashes with "Honk! An error occurred"

---

**Expected behavior**

MCP Apps should render in an iframe when the associated tool is called. The same MCP Apps work correctly in `basic-host` from the ext-apps SDK.

---

**Screenshots**

Error displayed:
```
Honk!
An error occurred.
Cannot convert undefined or null to object
```

Console stack trace:
```
TypeError: Cannot convert undefined or null to object
    at Object.entries (<anonymous>)
    at k (App-D22GTiTp.js:1737:30407)
    at _ (App-D22GTiTp.js:1737:30607)
    at poe (App-D22GTiTp.js:1737:30984)
    at Qs (index-C1WyBZUE.js:9:48084)
    at fu (index-C1WyBZUE.js:9:70876)
    at up (index-C1WyBZUE.js:9:81208)
    at Hp (index-C1WyBZUE.js:9:116953)
    at Qv (index-C1WyBZUE.js:9:115999)
    at Du (index-C1WyBZUE.js:9:115831)
```

Network `list_apps` response shows apps are correctly registered:
```json
{
  "apps": [
    {
      "uri": "ui://apps/clock",
      "name": "clock",
      "mimeType": "text/html;profile=mcp-app",
      "mcpServers": ["apps"],
      "width": 300,
      "height": 300,
      "resizable": false
    },
    {
      "uri": "ui://minder/compliance-dashboard",
      "name": "Compliance Dashboard",
      "mimeType": "text/html;profile=mcp-app",
      "mcpServers": ["minder"],
      "width": 800,
      "height": 600,
      "resizable": true
    }
  ]
}
```

---

**Please provide the following information**
- **OS & Arch:** Fedora 43, Linux 6.18.7-200.fc43.x86_64
- **Interface:** UI (Flatpak)
[...]

PR #7213, opened 2026-02-13
Prevent MCP Apps crash when tool arguments are null/undefined

Sahil8383:
## Summary
Fixes MCP Apps crashing with "Cannot convert undefined or null to object" when rendering tool calls. The crash occurred when the server sent `null`/`undefined` for tool arguments and our code called `Object.entries()` or passed non-objects into the MCP UI SDK. This PR adds null-safety in tool-call and MCP app rendering and updates the lockfile so `@mcp-ui/client` 6.x (which exports `AppRenderer`) is installed.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Related Issues
Relates to #6981

### DIFF
--- a/ui/desktop/package-lock.json
+++ b/ui/desktop/package-lock.json
@@ -219,7 +219,6 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -589,7 +588,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       },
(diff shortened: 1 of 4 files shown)


Sahil8383:
@DOsinga can you review this PR?

DOsinga:
Thanks for the quick turnaround, but I don't think this is the solution. The bug mentions that no apps render, not even the built-in clock app. That uses the StandAloneAppView, which calls McpAppRenderer, so the changes in the ToolCall* classes can't be involved in the crash.

StandAloneAppView though doesn't set any of the properties that you are modifying.

Sahil8383:
Let me give it another thought @DOsinga ,
since reproducing the issue is a bit difficult, So my first thought was the `Object.entries()` because if null or undefined is passed there then we can get the `TypeError: Cannot convert undefined or null to object` error.

DOsinga:
I understand where you are coming from, but generally the goose setup is pretty type safe - the backend is in rust, the calls are automatically converted to json and there are matching ts definitions created too. to assume that anything where the type definition says it cannot be null to be null, we'll need a theory.

Sahil8383:
Yeah @DOsinga makes sense, let me give this issue a good thought.

DOsinga:
I'm going to close this. when we have a better plan, happy to look at that PR

PR #7217, opened 2026-02-13
Added cmd to validate bundled extensions json

zanesq:
## Summary
Problem: The goose-releases repo has a bundled-extensions.json that defines extensions
shipped with the internal desktop app. Two extensions (Asana and Neighborhood) were added
with "type": "http", but the app only recognizes "builtin", "stdio", and "streamable_http".
When the sync loop hit Asana, it crashed and silently dropped every extension after it.
There was no way to catch this before shipping. The release built fine and the error only
surfaced at runtime.

Solution: Added a hidden `validate-extensions` command to both the goose CLI and goosed
server binaries. It reads a bundled-extensions.json file, attempts to deserialize each
entry into the same ExtensionConfig enum the runtime uses, and reports all errors with
the extension name and index. It specifically detects common mistakes like unknown type
values (listing the valid ones) and "url" vs "uri" field confusion on streamable_http
entries. It exits with code 1 on any error, making it suitable as a CI gate step in the
release workflow right after the Rust build, before packaging. The list of valid types is
defined as a constant on ExtensionConfig itself so it lives next to the enum definition
and has a test ensuring it stays in sync.

### DIFF
--- a/crates/goose-server/src/main.rs
+++ b/crates/goose-server/src/main.rs
@@ -69,19 +69,12 @@ async fn main() -> anyhow::Result<()> {
             }
         }
         Commands::ValidateExtensions { path } => {
-            let result = validate_extensions::validate_bundled_extensions(&path)?;
-            if result.errors.is_empty() {
-                println!("✓ All {} extensions validated successfully.", result.total);
-            } else {
-                eprintln!(
-                    "✗ Found {} error(s) in {} extension(s):\n",
-                    result.errors.len(),
-                    result.errors.len()
-                );
-                for (i, error) in result.errors.iter().enumerate() {
-                    eprintln!("  [{}] {}", i, error);
+            match validate_extensions::validate_bundled_extensions(&path) {
+                Ok(msg) => println!("{msg}"),
+                Err(e) => {
+                    eprintln!("{e}");
+                    std::process::exit(1);
                 }
-                std::process::exit(1);
             }
         }
     }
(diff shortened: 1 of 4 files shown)


crates/goose/src/agents/extension.rs
```diff
@@ -374,6 +374,16 @@ impl ExtensionConfig {
         }
     }
 
+    pub const VALID_TYPES: &[&str] = &[
```
DOsinga: I don't think we should need this. in the validate we could just restructure the loop body to:

Parse BundledExtensionEntry (metadata) — same as now
Do the type-specific hint checks (url/uri, missing cmd) on the raw JSON
Try serde_json::from_value::<ExtensionConfig> — this catches unknown types and any other structural issues

crates/goose-server/src/main.rs
```diff
@@ -59,6 +68,22 @@ async fn main() -> anyhow::Result<()> {
...
+                    "✗ Found {} error(s) in {} extension(s):\n",
+                    result.errors.len(),
+                    result.errors.len()
+                );
+                for (i, error) in result.errors.iter().enumerate() {
+                    eprintln!("  [{}] {}", i, error);
+                }
+                std::process::exit(1);
+            }
+        }
```
DOsinga, line 79: this is duplicated code. just have the validate_bundled_extension(...) return all this text and print it.

PR #7219, opened 2026-02-13
fix: fork session

angiejones:
## Summary
The `Fork Session` feature was not working if the message was not edited. 

In `UserMessage.tsx`, the handleSave function had this flow:

```
const handleSave = (editType: 'fork' | 'edit' = 'fork') => {
  setIsEditing(false);                              // ← always closes the textbox
  if (editContent.trim() === textContent.trim()) {  // ← if content unchanged...
    return;                                          // ← ...bail out early!
  }
  onMessageUpdate(message.id, editContent, editType); // ← fork never reached
};
```

When you click "Fork Session", the most common case is that you haven't changed the message text — you just want to create a new session from that point. But the early return on the content-unchanged check prevented onMessageUpdate from ever being called. The edit textbox closed (setIsEditing(false)) but no fork happened.

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
tested locally

### DIFF
--- a/ui/desktop/src/components/ProgressiveMessageList.tsx
+++ b/ui/desktop/src/components/ProgressiveMessageList.tsx
@@ -36,7 +36,7 @@ interface ProgressiveMessageListProps {
   // Custom render function for messages
   renderMessage?: (message: Message, index: number) => React.ReactNode | null;
   isStreamingMessage?: boolean; // Whether messages are currently being streamed
-  onMessageUpdate?: (messageId: string, newContent: string) => void;
+  onMessageUpdate?: (messageId: string, newContent: string, editType?: 'fork' | 'edit') => void;
   onRenderingComplete?: () => void; // Callback when all messages are rendered
   submitElicitationResponse?: (
     elicitationId: string,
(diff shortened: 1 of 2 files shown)


zanesq (approved):
Thanks! Tested locally LGTM

ui/desktop/src/components/ProgressiveMessageList.tsx
```diff
@@ -36,7 +36,7 @@ interface ProgressiveMessageListProps {
   // Custom render function for messages
   renderMessage?: (message: Message, index: number) => React.ReactNode | null;
   isStreamingMessage?: boolean; // Whether messages are currently being streamed
-  onMessageUpdate?: (messageId: string, newContent: string) => void;
+  onMessageUpdate?: (messageId: string, newContent: string, editType?: 'fork' | 'edit') => void;
```
DOsinga, line 39: do we need this change? I don't see the callers being updated to provide a value?

angiejones: UserMessage has this which has 3 args:

```
      if (onMessageUpdate && message.id) {
        onMessageUpdate(message.id, editContent, editType);
      }
```

but ProgressiveMessageList defied it as only accepting 2 args, so was trying to fix the type

Issue #6816, opened 2026-01-29
Extensions fail to load on Windows


Joshua-Hell:
Being on Windows, browsing an extension and adding it to Goose Desktop is fine. After adding it you can see that the parts of the command are getting escaped with a backslash. (`npx -y \@upstash/context7-mcp`)

After adding the extension and creating a new chat you can see that it fails with the following message: `process quit before initialization: stderr = `

On Linux this works fine as expected because the backslash is no problem in a Linux shell. Windows on the other hand fails to start the MCP. You can view the same behaviour when running the escaped command directly.

Error running manually with an escaped `@` sign:
```shell
PS C:\Users\joshu> npx -y \@upstash/context7-mcp
npm error code ENOENT
npm error syscall open
npm error path C:\@upstash\context7-mcp\package.json
npm error errno -4058
npm error enoent Could not read package.json: Error: ENOENT: no such file or directory, open 'C:\@upstash\context7-mcp\package.json'
npm error enoent This is related to npm not being able to find a file.
npm error enoent
npm error A complete log of this run can be found in: C:\Users\joshu\scoop\persist\nodejs-lts\cache\_logs\2026-01-29T16_29_20_751Z-debug-0.log
```

Starting without:
```
PS C:\Users\joshu> npx -y @upstash/context7-mcp
WARNING: Using default CLIENT_IP_ENCRYPTION_KEY.
Context7 Documentation MCP Server v2.1.1 running on stdio
```

**diagnostics zip**

[diagnostics_20260129_2.zip](https://github.com/user-attachments/files/24942827/diagnostics_20260129_2.zip)

**Expected behavior**

No error message and a starting MCP.

---

**Screenshots**

<img width="599" height="685" alt="Image" src="https://github.com/user-attachments/assets/3aad822a-5e25-4b61-9ab0-95254bcdb381" />

<img width="269" height="332" alt="Image" src="https://github.com/user-attachments/assets/040bdba6-155d-4183-8950-710e43739c38" />

---

- **OS & Arch:** Windows 11 Home 25H2
- **Interface:** Goose Desktop
- **Version:** 1.21.2
- **Extensions enabled:** *Context7*, Developer, Extension Manager, Skills, Todo
- **Provider & Model:** Ollama - qwen2.5:7b (happens with non-ollama models too)

---

**Additional context**

I installed Goose Desktop using the Scoop package manager.
I'm by no means a Windows expert, but I believe it's because of the escaping of special characters.

DOsinga:
[...]

PR #7220, opened 2026-02-14
install node in goose dir

lifeizhou-ap:
## Summary
Fixed issue https://github.com/block/goose/issues/6816 that cannot load extensions starting with `npx` cmd in windows

Root cause: `npx.cmd` tries to install node which may requires system admin permission and then fails

Fix: install node in goose folder which does not require system admin permission and the node executable is self contained.

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [X] Build / Release
- [ ] Other (specify below)

### Testing
Testing the canary build artifact in Windows and user also tested it

### DIFF
--- a/ui/desktop/src/platform/windows/bin/npx.cmd
+++ b/ui/desktop/src/platform/windows/bin/npx.cmd
@@ -11,42 +11,7 @@ if exist "%GOOSE_NODE_DIR%\node-v%NODE_VERSION%.installed" (
     exit /b !errorlevel!
 )
 
-REM === Search PATH for a real npx (one that has node.exe alongside it) ===
-for /f "tokens=*" %%i in ('where npx.cmd 2^>nul') do (
-    if exist "%%~dpi\node.exe" (
-        "%%i" %*
-        exit /b !errorlevel!
-    )
-)
-
-REM === Check common locations not always on PATH ===
-for %%p in (
-    "C:\Program Files\nodejs"
-    "C:\Program Files (x86)\nodejs"
-) do (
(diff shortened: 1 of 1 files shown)


michaelneale:
would this mean we don't use npx - or still do but it falls back to using this node? (the whole idea of npx shim is that it uses hermit to install node in its own space - if it isn't, then I think the paths are getting mixed up - it shouldnt' be installing anything to the system?)

michaelneale:
would this mean we don't use npx - or still do but it falls back to using this node? (the whole idea of npx shim is that it uses hermit to install node in its own space - if it isn't, then I think the paths are getting mixed up - it shouldnt' be installing anything to the system?)

lifeizhou-ap:
> would this mean we don't use npx - or still do but it falls back to using this node? (the whole idea of npx shim is that it uses hermit to install node in its own space - if it isn't, then I think the paths are getting mixed up - it shouldnt' be installing anything to the system?)

Hermit does not support windows. To install npx on Windows, we need to install node. Before this PR, goose tries to install node in the system level.  With this PR. node is installed in goose specific dir, not the system level. In this case it is similar to use hermit to install on linux/macos

IIUR, your question is more about: if user has already installed node on their system level, shall we still install node on goose dir or just use the system level node?  

With this PR, we just use the system level node. We can make more change to ignore the system level node which seems cleaner

michaelneale:
@lifeizhou-ap oh of course! yes windows, LGTM then!

Issue #7223, opened 2026-02-14
Streaming responses clash with markdown text output rendering


ekoeppen:
**Describe the bug**

In case the model responds with markdown text in streaming mode, the response is partly rendered incorrectly. The root cause is that the responses are split across logical markdown components, and at least `bat` is not able to render this as it is likely stateless. E.g if the first block is `**[README.md](./`, followed by `README.md)**`, the result is not bold.

---

**To Reproduce**

1. Enter a prompt which will likely result in a larger markdown response, e.g. in the goose directory "summarize the markdown files here in a table"
2. Observe whether text attributes such as bold are applied correctly

---

**Expected behavior**

All text attributes are applied correctly 

---

**Screenshots**

<img width="547" height="275" alt="Image" src="https://github.com/user-attachments/assets/a1f1c6a9-b14d-4e22-ae3f-be1e6caa7877" />

---

**Please provide the following information**
- **OS & Arch:** macOS 26.2 arm64
- **Interface:** CLI
- **Version:** 1.24.0
- **Extensions enabled:** developer, code_execution
- **Provider & Model:** GCP Vertex AI, gemini-3-flash-preview

---

**Additional context**

Logs with show the splitting across text responses:

[llm_request.0.json](https://github.com/user-attachments/files/25312277/llm_request.0.json) (this is actually a `jsonl` file.)

ekoeppen:
A very crude way to fix this is to disable streaming responses, which I personally don't like that much anyway, I prefer the responses in one go, and find text trickling in rather distracting, but that is maybe another topic.

The correct technical solution is probably best done in the renderer (`output.rs`), which needs to a) consider whether the content is structured such as markdown, and b) if it is structured, ensure it can produce correct output by e.g. buffering as needed. There would still need to be an indication to the renderer that the response is complete.

DOsinga:
interesting. so the desktop just rerenders, but that might indeed be beyond what terminal can do. detecting whether we are in a block of markdown seems tricky though - I could have **hello in my text and it would stop all rendering (which you might be fine with).

ekoeppen:
I've been through a similar topic when implementing an HTML renderer when modem speeds were the norm, and it is pretty difficult (or rather impossible) without maintaining state somewhere if the input comes in irregular shapes. Especially tables were a killer.

[...]

PR #7231, opened 2026-02-15
fix: Accumulate text messages for batched rendering

ekoeppen:
In cases where the text format returned by the LLM is structured such as Markdown, streaming responses cannot reliably be rendered if the text messages arrive in separate blocks. Collect all text messages in a turn and render them at once after the turn completes.

This is a minimal change for the issue at hand only, possible followup changes would be to move more module state into the renderer struct, e.g. theme or quiet/verbose output.

## Summary

Introduce a struct holding the accumulated text messages, collect messages from streaming responses, and render at the end of the message stream.

Note that this changes the visual appearance of responses to appear at once instead of incrementally.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing

Manual testing and `cargo test`

### Related Issues
Relates to #7223 

After: 

image

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -164,6 +164,7 @@ pub struct CliSession {
     edit_mode: Option<EditMode>,
     retry_config: Option<RetryConfig>,
     output_format: String,
+    renderer: output::Renderer,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -259,6 +260,7 @@ impl CliSession {
             edit_mode,
             retry_config,
             output_format,
+            renderer: output::Renderer::new(),
         }
     }
 
@@ -769,7 +771,7 @@ impl CliSession {
 
         self.messages.clear();
         tracing::info!("Chat context cleared by user.");
-        output::render_message(
+        self.renderer.render_message(
             &Message::assistant().with_text("Chat context cleared.\n"),
             self.debug,
         );
@@ -847,7 +849,7 @@ impl CliSession {
                 &[],
             )
             .await?;
-        output::render_message(&plan_response, self.debug);
+        self.renderer.render_message(&plan_response, self.debug);
         output::hide_thinking();
         let planner_response_type = classify_planner_response(
             &self.session_id,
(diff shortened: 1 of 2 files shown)


DOsinga:
hi, I'm sorry, based on your comments on the other I gave the buffering suggestion you mentioned a go: https://github.com/block/goose/pull/7233 - basically your approach but with entitity detection. wanna have a look if that is what you were thinking?

ekoeppen:
> hi, I'm sorry, based on your comments on the other I gave the buffering suggestion you mentioned a go: #7233 - basically your approach but with entitity detection. wanna have a look if that is what you were thinking?

No worries, thanks for taking time with this! Yes, your changes do the trick as well. I've rebased an experimental branch which uses `termimad` instead of `bat`, and also that looked fine. It does have the advantage of keeping the incremental output mode.

I did think about some simpler approaches in context of your solution, e.g. just render if we get a newline char at the end (that's mostly a 60% solution), or render [until we get two newlines in a row](https://github.com/block/goose/commit/f4f464af62e63bf8151a2edf310c7602e4aafc7e) (makes it less incremental), in case you're put off by the extra complexity.

This topic has however great yak shaving potential, and I'm happy to leave this up to you and the other maintainers how to approach this, even if you decide it's not worth fixing this right now :)

DOsinga:
> This topic has however great yak shaving potential, and I'm happy to leave this up to you and the other maintainers how to approach this, even if you decide it's not worth fixing this right now :)

that is very true. I added table rendering to my branch. right now the output that we have is somewhat syntax highlighted but still more or less standard markdown, which I think is kinda nice. if we go all fancy we would lose that (though if you switch it off for non terminal use cases maybe it doesn't matter?)

if you have a pointer to the termimad version, I'd give that a spin. goose suggested that to me too as an option

ekoeppen:
> if you have a pointer to the termimad version, I'd give that a spin. goose suggested that to me too as an option

Sure, this is the branch: https://github.com/ekoeppen/goose/tree/termimad-rendering. The actual change is surprisingly small: https://github.com/block/goose/commit/4836faab19c911cc28aa118c4b334efe60b7abc2

I also tested loading custom themes, that has a bit more impact due to passing the theme around which carries now a theme parameter: https://github.com/ekoeppen/goose/tree/termimad-custom-themes

Termimad has the drawback that it doesn't do syntax highlighting at all.

ekoeppen:
Closing, done in #7233

PR #7233, opened 2026-02-15
Streaming markdown

DOsinga:
## Summary

Supports streaming markdown, somewhat


DOsinga:
yeah, if you open a code block with ``` and it never closes, it will stop streaming and flush it at the end. but right now we often just break syntax highlighting for code etc.

but yeah, we should think about the strategy when it comes to clients. I'm kinda thinking we keep the CLI, maybe slimming it down and then have N other clients that live outside our main repo? goose mobile, a TUI, etc

Issue #7245, opened 2026-02-16
OpenAI API Host falls back to default when saving other fields


bavadim:
## Problem
In OpenAI provider settings, updating another field and saving can make the effective `API Host` fall back to the default host, even if the `API Host` field itself was not changed.

The UI may still show the previously entered `API Host`, which hides the regression.

## Impact
Requests are sent to the default OpenAI endpoint instead of the user-configured host (proxy/self-hosted gateway), causing unexpected routing and failures.

## Steps to Reproduce
1. Open **OpenAI** provider settings.
2. Set a custom value in **API Host** and save.
3. Reopen OpenAI settings.
4. Change any field **except** `API Host` (for example API key), then click **Save**.
5. Validate effective config by making a request / running provider validation and checking which host is used.

## Expected
If `API Host` was not edited, its previously saved custom value remains the effective runtime host.

## Actual
After saving changes to another field, effective host can revert to OpenAI default while UI still shows the old custom `API Host` text.

## Scope
Provider settings save flow in desktop UI (confirmed with OpenAI `API Host`).

PR #7246, opened 2026-02-16
fix(ui): prevent OpenAI API Host fallback when saving other fields

bavadim:
## Summary
Fixes a UI bug in OpenAI provider settings where saving changes to a non-host field can cause the effective `API Host` to fall back to default (`https://api.openai.com`) even when `API Host` was not edited.

The UI can still display the previously entered host text, so the regression is easy to miss.

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
- Unit test (new):
  - `ui/desktop/src/components/settings/providers/modal/subcomponents/handlers/DefaultSubmitHandler.test.ts`
  - Covers OpenAI regression: update `OPENAI_API_KEY` without editing `OPENAI_HOST`, and assert host does not revert to default.
- Verification result:
  - Test fails on `main` (host falls back to default).
  - Test passes with this fix branch.

### Related Issues
Relates to #7245
Discussion: https://github.com/block/goose/issues/7245

Before:
Changing another OpenAI field and saving could make effective `API Host` revert to default while UI still showed old host text.

After:
Saving changes to non-host fields preserves previously configured `API Host`.

### DIFF
--- a/crates/goose/examples/tetrate_auth.rs
+++ b/crates/goose/examples/tetrate_auth.rs
@@ -10,13 +10,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Create new PKCE auth flow
     let mut auth_flow = TetrateAuth::new()?;
 
-    // Get the auth URL that would be opened
-    let auth_url = auth_flow.get_auth_url();
-    println!("Auth URL: {}", auth_url);
-    println!("\nStarting authentication flow...");
+    println!("Starting authentication flow...");
     println!("This will:");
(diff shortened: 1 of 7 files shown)


DOsinga:
thanks! somebody else also fixed this. since they have the DCO check, going to merge that one, but appreciate your efforts!

PR #7247, opened 2026-02-16
Everything is streaming

DOsinga:
## Summary

make stream() the required method to implement. provides a respond() default implementation based on that. requires a model config (/cc @katzdave ).


lifeizhou-ap:
The code is much simpler now with the change! 

We might have to keep the non streaming version for `anthropic`, `openai` and `ollama` as they support `from_custom_config` which has `supports_streaming` configuration.   

Another option could be: removing supports_streaming from custom config since most of the models are now supporting streaming.  However users may lose flexibility to use some model that does not support streaming.

DOsinga:
> Another option could be: removing supports_streaming from custom config since most of the models are now supporting streaming. However users may lose flexibility to use some model that does not support streaming.

ah yes, good point. fixed

PR #7248, opened 2026-02-16
fix(ui): preserve server config values on partial provider config save

### LINKED ISSUE #7245: OpenAI API Host falls back to default when saving other fields

clayarnoldg2m:
## Summary
- Fixes #7245 — OpenAI API Host (and other non-secret config fields) silently revert to default when saving a partial provider configuration
- The submit filter in `ProviderConfiguationModal.tsx` now includes fields with existing non-masked server values (`serverValue` as plain string), not just fields where the user typed a new value
- Masked/secret values are correctly excluded from the fallback (they remain on the server unchanged)

## Test plan
- [ ] Configure a provider with both API Host and API Key set
- [ ] Edit only the API Key (leave Host untouched) and save
- [ ] Verify the Host value persists (previously it silently reverted to default)
- [ ] Verify secret fields (masked values) still work correctly when unchanged
- [ ] TypeScript type check passes: `npx tsc --noEmit`

### DIFF
--- a/ui/desktop/src/components/settings/providers/modal/ProviderConfiguationModal.tsx
+++ b/ui/desktop/src/components/settings/providers/modal/ProviderConfiguationModal.tsx
@@ -105,8 +105,15 @@ export default function ProviderConfigurationModal({
 
     const toSubmit = Object.fromEntries(
       Object.entries(configValues)
-        .filter(([_k, entry]) => !!entry.value)
-        .map(([k, entry]) => [k, entry.value || ''])
+        .filter(
+          ([_k, entry]) =>
+            !!entry.value ||
+            (entry.serverValue != null && typeof entry.serverValue === 'string')
+        )
+        .map(([k, entry]) => [
+          k,
+          entry.value ?? (typeof entry.serverValue === 'string' ? entry.serverValue : ''),
+        ])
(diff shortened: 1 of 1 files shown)


DOsinga:
thanks, same as https://github.com/block/goose/pull/7246

PR #7250, opened 2026-02-16
fix: use explicit max_tokens in apps extension to prevent truncation

### LINKED ISSUE #7239: Bug: Apps extension inner LLM call has no max_tokens, causing truncation and 'missing field html' error
## Summary

The Apps extension's inner LLM call to generate HTML app content does not set `max_tokens`, so the API defaults to 8,192 output tokens. This is insufficient for generating any non-trivial HTML app. The response gets truncated (`finish_reason: "length"`), the `html` field is never written, and the user sees a cryptic error:

```
Error: Failed to parse tool response: missing field `html`
```

## Reproduction

1. Use the Databricks provider with `databricks-claude-opus-4-6` (or any provider where the default max output tokens is ~8K)
2. Ask goose to create a non-trivial app (e.g., "Create a figure skating game app")
3. The `apps__create_app` tool returns `Error: Failed to parse tool response: missing field html` every time

## Root Cause

In `crates/goose/src/agents/apps_extension.rs`, `generate_new_app_content()` and `generate_updated_app_content()` call:

```rust
let (response, _usage) = provider
    .complete(session_id, &system_prompt, &messages, &tools)
    .await
```

This uses `provider.complete()` which inherits the provider's `ModelConfig`. When `max_tokens` is `None` in the config, no `max_tokens` is sent to the API, and the Databricks/Claude API defaults to 8,192 output tokens.

The inner LLM is asked to call a `create_app_content` tool where the `html` field (a complete HTML/CSS/JS app) is serialized as part of the JSON arguments. The LLM generates fields in order — `name`, `description`, `width`, `height`, `resizable` — and hits the 8,192 token ceiling before ever reaching `html`.

**Evidence from diagnostic logs** — all 3 inner LLM calls show the same pattern:

| Inner LLM Call | finish_reason | output_tokens | Has html? |
|----------------|--------------|---------------|-----------|
| Attempt 1      | `length`     | 8,192         | No        |
| Attempt 2      | `length`     | 8,192         | No        |
| Attempt 3      | `length`     | 8,192         | No        |

The returned arguments each time:
```json
{"name":"figure-skating-championship","description":"Interactive figure skating game...","height":700,"resizable":false,"width":900}
```

No `html` field — the LLM ran out of tokens before starting it.

## Suggested Fix

1. **Set a higher `max_tokens`** for the inner LLM call — use `complete_with_model()` with a `ModelConfig` that overrides `max_tokens` to something like 16,384+ (or even higher for complex apps)
2. **Detect truncation** — check `finish_reason` / usage in the response and return a meaningful error instead of the cryptic serde error
3. Both ideally

## Environment

- Provider: Databricks
- Model: `databricks-claude-opus-4-6`
- `max_tokens` in config: `null` (not set)
- API default: 8,192 output tokens

clayarnoldg2m:
## Summary
- Replaced `provider.complete()` with `provider.complete_with_model()` in the apps platform extension, passing a `ModelConfig` with explicit `max_tokens = 16384`
- Applied to both `generate_new_app_content()` and `generate_updated_app_content()` in `crates/goose/src/agents/platform_extensions/apps.rs`
- Without explicit `max_tokens`, the provider's default can be too low, causing generated app HTML to be truncated

## Test plan
- [x] `cargo check` — clean
- [x] `cargo clippy -- -D warnings` — clean
- [x] `cargo test -p goose` — all passing
- [x] Verified by tester1

Fixes #7239

### DIFF
--- a/crates/goose/src/agents/platform_extensions/apps.rs
+++ b/crates/goose/src/agents/platform_extensions/apps.rs
@@ -290,8 +290,11 @@ impl AppsManagerClient {
         let messages = vec![Message::user().with_text(&user_prompt)];
         let tools = vec![Self::create_app_content_tool()];
 
+        let mut model_config = provider.get_model_config();
+        model_config.max_tokens = Some(16384);
+
         let (response, _usage) = provider
-            .complete(session_id, &system_prompt, &messages, &tools)
+            .complete_with_model(Some(session_id), &model_config, &system_prompt, &messages, &tools)
             .await
             .map_err(|e| format!("LLM call failed: {}", e))?;
 
@@ -321,8 +324,11 @@ impl AppsManagerClient {
         let messages = vec![Message::user().with_text(&user_prompt)];
         let tools = vec![Self::update_app_content_tool()];
 
+        let mut model_config = provider.get_model_config();
+        model_config.max_tokens = Some(16384);
+
         let (response, _usage) = provider
-            .complete(session_id, &system_prompt, &messages, &tools)
+            .complete_with_model(Some(session_id), &model_config, &system_prompt, &messages, &tools)
             .await
             .map_err(|e| format!("LLM call failed: {}", e))?;
 


DOsinga:
closing, I don't think this is the way

PR #7252, opened 2026-02-16
fix: handle reasoning_content for Kimi/thinking models

### LINKED ISSUE #6902: [Bug] Kimi k2.5 API Error: "thinking is enabled but reasoning_content is missing" in tool calls

clayarnoldg2m:
## Summary
- Preserves `reasoning_content` when splitting parallel tool calls in `agent.rs` — providers like Kimi require it on all assistant messages with tool_calls when thinking mode is enabled
- Omits `reasoning_content` field entirely when empty in `format_messages` instead of sending `""` — Kimi rejects empty reasoning_content
- Properly accumulates `reasoning_content` chunks during streaming and emits as `MessageContent::reasoning()`

## Context
Fixes #6902. Kimi (Moonshot) and other thinking models that use `reasoning_content` (similar to DeepSeek) were failing because:
1. When parallel tool calls are split into separate messages, the reasoning content was lost
2. An empty `reasoning_content: ""` was sent even when no reasoning occurred, which Kimi rejects
3. Streaming `reasoning_content` chunks were not accumulated, so reasoning was lost in streamed responses

Reference: PR #6962 identified these issues. This PR implements the three core fixes cleanly.

## Test plan
- [x] `cargo test -p goose --lib` — 666 tests pass
- [x] `cargo clippy -p goose --all-targets -- -D warnings` — clean
- [ ] Manual verification with DeepSeek reasoning model (reasoning still works)
- [ ] Manual verification with Kimi thinking model (tool calls + reasoning work)

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -1346,8 +1346,9 @@ impl Agent {
                                     }
                                 }
 
-                                // Preserve thinking content from the original response
+                                // Preserve thinking/reasoning content from the original response
                                 // Gemini (and other thinking models) require thinking to be echoed back
+                                // Kimi/DeepSeek require reasoning_content on assistant tool call messages
                                 let thinking_content: Vec<MessageContent> = response.content.iter()
                                     .filter(|c| matches!(c, MessageContent::Thinking(_)))
                                     .cloned()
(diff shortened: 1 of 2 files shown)


michaelneale:
Full CI passed on branch PR: https://github.com/block/goose/pull/7262

PR #7255, opened 2026-02-16
refactor: remove allows_unlisted_models flag, always allow custom model entry

### LINKED ISSUE #7103: Remove allows_unlisted_models — always allow custom model entry

We keep merging one-liner PRs that flip `allows_unlisted_models` to `true` per provider:
- #6761 — OpenRouter
- #6393 — GCP Vertex AI
- #7090 — claude-code, codex, gemini-cli, cursor-agent, chatgpt-codex

The underlying problem is that our `known_models` lists don't match what the APIs actually return. I ran `fetch_supported_models()` against the real APIs:

- **OpenAI**: Our known_models has 11 entries but the API returns 115 models. Many valid chat models (`gpt-5`, `gpt-5-nano`, `gpt-5-pro`, `gpt-5.2-codex`) aren't in our list, and each also has a dated variant (`gpt-5-nano-2025-08-07`). Without unlisted enabled, users can't select any of them.
- **Ollama**: API returns names with tags (`qwen3:latest`, `qwen3:4b`) but known_models lists bare names (`qwen3`). Zero of our 4 known models match the fetched list due to the `:tag` mismatch.
- **Claude Code**: `fetch_supported_models()` returns only `["sonnet", "opus"]` — hardcoded, no API. You can't enter `claude-sonnet-4-5-20250929` without unlisted enabled.

The lists serve two purposes: (a) recommending known-good text+tools models, and (b) blocking users from entering arbitrary names. But (b) doesn't hold up — `GOOSE_MODEL` env var bypasses the UI entirely, and the canonical registry already resolves aliases to real names.

Remove the `allows_unlisted_models` field from `ProviderMetadata` and the `with_unlisted_models()` builder method entirely. The configure UI would always show "Enter a model not listed..." alongside the curated list. If a user enters a model name the provider doesn't support, they get an API error immediately — no need to gate this in the UI.

Default `allows_unlisted_models` to `true` instead of removing it. But if no provider has a reason to set it `false`, the flag is dead code. The simpler option is to delete it.

**Additional context**

#7065 attempted a broader fix ("use dynamic model listing for all providers") but was closed. #5520 and #4834 also tried to address the same root issue from different angles.

clayarnoldg2m:
## Summary
- Removes the `allows_unlisted_models` field from `ProviderMetadata` and its `with_unlisted_models()` builder method
- All providers now always allow users to enter custom model names not in the fetched list
- Simplifies CLI `configure.rs` to unconditionally show "Enter a model not listed..." option
- Updates `RecipeModelSelector.tsx` and `SwitchModelModal.tsx` to always show custom model entry
- Regenerates OpenAPI schema and TypeScript types (field fully removed from API)
- 16 files changed, 56 deletions, 16 insertions

## Test plan
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo test` — passes (2 pre-existing integration failures: claude_code, xai)
- [x] `npx tsc --noEmit` — clean
- [x] `just generate-openapi` — regenerated, no `allows_unlisted_models` references remain

Fixes #7103

### DIFF
--- a/crates/goose-cli/src/commands/configure.rs
+++ b/crates/goose-cli/src/commands/configure.rs
@@ -465,13 +465,11 @@ fn select_model_from_list(
                 ),
             );
 
-            if provider_meta.allows_unlisted_models {
-                model_items.push((
-                    UNLISTED_MODEL_KEY.to_string(),
-                    "Enter a model not listed...".to_string(),
-                    "",
-                ));
-            }
+            model_items.push((
+                UNLISTED_MODEL_KEY.to_string(),
+                "Enter a model not listed...".to_string(),
+                "",
+            ));
 
             let selection = cliclack::select("Select a model:")
                 .items(&model_items)
@@ -491,13 +489,11 @@ fn select_model_from_list(
         let mut model_items: Vec<(String, String, &str)> =
             models.iter().map(|m| (m.clone(), m.clone(), "")).collect();
 
-        if provider_meta.allows_unlisted_models {
-            model_items.push((
-                UNLISTED_MODEL_KEY.to_string(),
-                "Enter a model not listed...".to_string(),
-                "",
-            ));
-        }
+        model_items.push((
+            UNLISTED_MODEL_KEY.to_string(),
+            "Enter a model not listed...".to_string(),
+            "",
+        ));
 
         let selection = cliclack::select("Select a model:")
             .items(&model_items)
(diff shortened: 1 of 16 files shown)


michaelneale:
I am ok with this - I can see that we have been adding the flag one by one, but why even have it? this reduces complexity, I like it, but @codefromthecrypt - you ok with it? stamp if you think so

PR #7256, opened 2026-02-16
fix(ui): prevent MCP app iframe from staying stuck at compressed width

### LINKED ISSUE #6818: Resize goose desktop shrink the mcp app and it never grows back.
**Describe the bug**

Due to the change from #6376, now goose desktop bind the iframe width with the inner app width through `ui/notifications/size-changed`. However, `ui/notifications/size-changed` will only get triggered when there is not enough space in the app which cause the problem that once the app gets compressed it never grows back - see the screen recording. 

P.S.  I don't really know what would be the use case for the host to do something meaningful by using this size-change info from the app. Please enlighten me if anyone thinks of one. 😁 

---

**To Reproduce**
1. Resize the goose desktop to compress the mcp app
2. Resize the goose desktop back to larger window width
3. Mcp app remains its compressed width 

---

**Expected behavior**
I think the app should utilize the full width of its container.

---

**Screenshots**

https://github.com/user-attachments/assets/e6ba3844-aedf-45a1-9448-91623996dd3e

---

- **OS & Arch:** macOS M3 pro
- **Interface:** UI
- **Version:** debug mode with 5bffc1c8ef5d806d3b623132f8cc14cf4fc9bf7a
- **Provider & Model:** Google – gemini-3-pro-preview

---

clayarnoldg2m:
## Summary
- Fixes #6818 — MCP app iframe gets stuck at compressed width after window resize
- The `size-changed` notification from MCP apps was setting a fixed pixel `width` on the container, which never grew back when the window expanded (the app can't detect size changes beyond its own container bounds)
- Changed to use the app-reported width as `maxWidth` instead of `width`, so the container stays fluid (`width: 100%`) and naturally expands with the window while still respecting the app's preferred maximum

## Root cause
PR #6376 introduced width tracking from `ui/notifications/size-changed`. When an app reported its width (e.g., `400px` when compressed), the container was pinned to that exact width. Since the app is inside the container, it couldn't detect the window growing beyond `400px` to send a larger width notification — a feedback loop that kept it stuck.

## Test plan
- [ ] Open an MCP app in Goose desktop
- [ ] Shrink the window to compress the app
- [ ] Expand the window back — app should now fill available width (previously stayed compressed)
- [ ] Verify apps that report a preferred width still get `maxWidth` capping
- [ ] `npx tsc --noEmit` passes
- [ ] `cargo clippy` clean

### DIFF
--- a/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
+++ b/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
@@ -547,8 +547,8 @@ export default function McpAppRenderer({
   const containerStyle = isExpandedView
     ? { width: '100%', height: '100%' }
     : {
-        width: iframeWidth !== null ? `${iframeWidth}px` : '100%',
-        maxWidth: '100%',
+        width: '100%',
+        maxWidth: iframeWidth !== null ? `${iframeWidth}px` : '100%',
         height: `${iframeHeight || DEFAULT_IFRAME_HEIGHT}px`,
       };
 


DOsinga:
I'm not seeing this issue on main with my toy MCP app, it just always the width of the screen. thoughts @aharvard ?

aharvard:
@DOsinga it depends on how the app is sending width data in the size change message. Regardless, what we have for goose right now is incorrect. Will share notes!

DOsinga:
ok, so looks like we can close this then, thanks for the check @aharvard

aharvard:
Here's the PR that reverts and sends container dimensions: https://github.com/block/goose/pull/7300

PR #7257, opened 2026-02-16
fix: preserve symlinks when saving config

### LINKED ISSUE #6847: Feature request: do not move the config.yaml file to bak - this breaks dotfiles (symlinks)
**Please explain the motivation behind the feature request:**

Currently, configuration changes move the original configuration file to a new location such as `~/.config/goose/config.yaml.bak`, `~/.config/goose/config.yaml.bak.1`, `~/.config/goose/config.yaml.bak.2`, etc.

This breaks any set up that uses symlinks pointing to a central "dotfiles" location.

Instead of moving the original file, create a copy of it, then modify it in place.

**Describe alternatives you've considered:**

None. The current system completely breaks my workflow, making it impossible to save Goose configs in my dotfiles.

**Additional context:**

None.

clayarnoldg2m:
## Summary
- Resolves symlinks via `canonicalize()` in `save_values` before the atomic write+rename, so symlinked config files (e.g. dotfiles) are preserved instead of being replaced with regular files
- Removes the automatic backup mechanism (`create_backup_if_needed`, `rotate_backups`, `try_restore_from_backup`, `get_backup_paths`) which created `.bak` files on every write — redundant since explicit backup is available via the `config/backup` API endpoint
- Simplifies corrupted config recovery to just reset to defaults instead of attempting `.bak` file recovery

## Context
Closes #6847. Users who symlink `~/.config/goose/config.yaml` to a central dotfiles directory had their symlinks destroyed on every config save. The root cause was `std::fs::rename(temp_path, config_path)` which replaces the symlink inode with a new regular file.

Per [maintainer guidance](https://github.com/block/goose/issues/6847#issuecomment-3825704778), the automatic backup mechanism is a legacy holdover that can be removed entirely.

## Test plan
- [x] `cargo test -p goose --lib` — 667 tests pass (net +1 from new symlink test)
- [x] `cargo clippy -p goose --all-targets -- -D warnings` — clean
- [x] New test `test_symlinked_config_preserved` verifies symlink survives write and data is correct
- [ ] Manual verification: symlink `~/.config/goose/config.yaml` to another location, modify config, verify symlink intact

### DIFF
--- a/crates/goose/src/config/base.rs
+++ b/crates/goose/src/config/base.rs
@@ -275,21 +275,12 @@ impl Config {
         let mut values = if self.config_path.exists() {
             self.load_values_with_recovery()?
         } else {
-            // Config file doesn't exist, try to recover from backup first
-            tracing::info!("Config file doesn't exist, attempting recovery from backup");
+            tracing::info!("Config file doesn't exist, creating default configuration");
 
-            if let Ok(backup_values) = self.try_restore_from_backup() {
-                tracing::info!("Successfully restored config from backup");
-                backup_values
-            } else {
-                // No backup available, create a default config
-                tracing::info!("No backup found, creating default configuration");
+            // Try to load from init-config.yaml if it exists, otherwise use empty config
+            let default_config = self.load_init_config_if_exists().unwrap_or_default();
 
-                // Try to load from init-config.yaml if it exists, otherwise use empty config
-                let default_config = self.load_init_config_if_exists().unwrap_or_default();
-
-                self.create_and_save_default_config(default_config)?
-            }
+            self.create_and_save_default_config(default_config)?
         };
 
         // Run migrations on the loaded config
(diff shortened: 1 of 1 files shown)


crates/goose/src/config/base.rs
```diff
@@ -1242,44 +1094,24 @@ mod tests {
...
-        for (i, path) in backup_paths.iter().enumerate() {
-            println!(
-                "After second write - Backup {} exists: {}",
-                i,
-                path.exists()
-            );
-        }
-
         // Corrupt the main config file
         std::fs::write(config_file.path(), "invalid: yaml: content: [unclosed")?;
```
DOsinga: remove this test

DOsinga:
the diff seems to show changes from other PRs, can you clean that up?

DOsinga:
closing since no reaction

PR #7258, opened 2026-02-16
fix: truncate long tool output for user display

### LINKED ISSUE #7025: Long tool output is not always truncated when shown to the user
My agent called `ls -R` using the developer tools (despite the fact that it shouldn't).
As a result the output hit the maximum character limit.

But the issue is that the entire output was displayed to me, showing thousands of lines.
It seems like the expected behavior

---

**To Reproduce**
1. Make a tool call with many lines of output

---

**Expected behavior**
A small number of lines are shown

---

**Screenshots**

---

- **OS & Arch:** [e.g. Ubuntu 22.04 x86]
- **Interface:** [UI / CLI]
- **Version:** [e.g. v1.0.2]
- **Provider & Model:** [e.g. Google – gemini-1.5-pro]

[diagnostics_20260206_10.zip](https://github.com/user-attachments/files/25119109/diagnostics_20260206_10.zip)

---

**Additional context**
I have been trying to use an alternative system prompt which showed better behavior for me. It still includes the notes about extensions (in particular no "ls -R") but gpt-oss is dumb

clayarnoldg2m:
## Summary
- Tool output exceeding 10,000 characters is now truncated for user display with an indicator showing `[output truncated: showing N of M characters]`
- Applied at both display layers: server SSE stream (UI) and CLI terminal output
- Full content is preserved for agent processing — only the display copy is truncated
- CLI users can still toggle full output display with the existing setting

### Changes
- `crates/goose/src/utils.rs`: Added `truncate_tool_text_for_display()` and `truncate_message_for_display()` utilities with 5 new tests
- `crates/goose-server/src/routes/reply.rs`: Truncate message before SSE serialization to UI
- `crates/goose-cli/src/session/output.rs`: Truncate in `render_tool_response()` (respects existing `show_full_tool_output` toggle)

## Test plan
- [x] `cargo test -p goose --lib -- utils::tests` — 15/15 passing (5 new)
- [x] `cargo clippy -p goose -- -D warnings` — clean
- [x] New tests verify: short text passes through, long text truncated with indicator, message-level truncation for tool responses

Fixes #7025

### DIFF
--- a/crates/goose-cli/src/session/output.rs
+++ b/crates/goose-cli/src/session/output.rs
@@ -8,7 +8,7 @@ use goose::conversation::message::{
 use goose::providers::canonical::maybe_get_canonical_model;
 #[cfg(target_os = "windows")]
 use goose::subprocess::SubprocessExt;
-use goose::utils::safe_truncate;
+use goose::utils::{safe_truncate, truncate_tool_text_for_display};
 use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
 use rmcp::model::{CallToolRequestParams, JsonObject, PromptArgument};
 use serde_json::Value;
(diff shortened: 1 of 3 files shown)


DOsinga:
I appreciate the effort, but this is claude code's work without you having had a look at it. the diff is not clean and contains commits from other PRs from you. I also don't think this is the way. for the desktop we already hide the output of tools by default. for the desktop, I think you want to be able to see this - we're not a TUI, so you can't recover this. furthermore this breaks scripting

appreciate the enthusiasm, but please have a look at things before creating PR and look at what the agent has done

PR #7284, opened 2026-02-17
fix: prevent SageMaker TGI from being marked configured when only Bedrock keys are set

### LINKED ISSUE #7268: Bedrock Provider sets value on AWS Sage Manager
**To Reproduce**
1. Go to Set new Provider
2. Click on Amazon Bedrock -> Configure
3. Set variables like AWS_PROFILE , AWS_REGION , AWS_BEARER_TOKEN_BEDROCK
4. Click Enable Provider

Result : Amazon SageMaker TGI results enable

---

**Expected behavior**
Bedrock provider to become available

---

**Screenshots**

Image

---

- **OS & Arch:** Windows 11
- **Interface:** UI
- **Version:** 1.24.0
- **Extensions enabled:** Not relevant (I suppose)
- **Provider & Model:** None yet

---

**Additional context**

ibaaaaal:
## Summary

Fixes #7268 — Configuring Amazon Bedrock causes Amazon SageMaker TGI to appear as configured.

This PR also fixes a related issue: Amazon Bedrock itself was not appearing in the provider selection dropdown because all its config keys were optional.

### Issue 1: SageMaker TGI incorrectly marked as configured
`SAGEMAKER_ENDPOINT_NAME` was declared as `required=false` in the provider metadata, while `AWS_REGION` and `AWS_PROFILE` were `required=true` with defaults. When a user configures Bedrock, the shared `AWS_REGION`/`AWS_PROFILE` keys get stored globally. The `check_provider_configured()` logic then sees SageMaker TGI's required keys are satisfied and marks it as configured, even though no SageMaker endpoint was actually set up.

### Issue 2: Bedrock not appearing in provider dropdown  
Amazon Bedrock had all config keys marked as `required=false`. The `check_provider_configured()` function returns `false` for providers with zero required keys, causing Bedrock to never appear as "configured" in the provider selection dropdown.

### SageMaker TGI fix (`sagemaker_tgi.rs`):
- Mark `SAGEMAKER_ENDPOINT_NAME` as `required=true` — it is already required at runtime
- Mark `AWS_REGION` and `AWS_PROFILE` as `required=false` — consistent with Bedrock

### Bedrock fix (`bedrock.rs`):
- Mark `AWS_REGION` as `required=true` with default `"us-east-1"` — this ensures Bedrock is properly marked as configured when the region is set
- Update test assertion to reflect correct expected behavior

## Validation

- `cargo fmt --check` — pass
- `cargo clippy -D warnings` — pass  
- `cargo test --release -p goose` — 687 tests pass (4 unrelated failures in prompt_manager are pre-existing)


lifeizhou-ap (approved):
Thanks for fixing it!

PR #7296, opened 2026-02-18
fix(cli): handle Reasoning content and fix streaming thinking display

rabi:
## Summary

Add MessageContent::Reasoning to both CLI renderers. Fix streaming to print a single "Thinking:" header then stream dimmed text, instead of repeating the header per chunk.

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested locally.

### DIFF
--- a/crates/goose-cli/src/session/output.rs
+++ b/crates/goose-cli/src/session/output.rs
@@ -16,7 +16,7 @@ use std::cell::RefCell;
 use std::collections::HashMap;
 use std::io::{Error, IsTerminal, Write};
 use std::path::Path;
-use std::sync::Arc;
+use std::sync::{Arc, LazyLock};
 use std::time::Duration;
 
 use super::streaming_buffer::MarkdownBuffer;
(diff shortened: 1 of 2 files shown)


crates/goose-cli/src/session/output.rs
```diff
@@ -410,6 +412,38 @@ pub fn goose_mode_message(text: &str) {
...
+}
+
+fn render_thinking_streaming(
+    text: &str,
+    buffer: &mut MarkdownBuffer,
+    header_shown: &mut bool,
+    theme: Theme,
+) {
+    if should_show_thinking() {
+        flush_markdown_buffer(buffer, theme);
```
DOsinga, line 437: presumably this doesn't happen half way rendering of markdown? not that we could do much about it

Issue #7288, opened 2026-02-18
Enhancement: Include recipe parameter details within generated `load` details


jcwilk:
**Please explain the motivation behind the feature request.**
The `load` mechanism of the `Summon` loads up information about (among other things) recipes so that it has enough context to know what to pass to `delegate` so that it can spin off the recipe successfully. However, one thing that is *not* included in this `load` information is the list of parameters, whether they're required, what their type is, and what the description of the parameter is. The only way it can get information about the parameters is by intuiting them from the names and positions of the interpolated variables.

This has a consequence of reducing how nimbly and accurately recipes can be used with agents possibly needing to do some guess and check before they get it right - not ideal when we already have the full details of the parameters available but out of reach.

**Describe the solution you'd like**
Simply include the full parameter info in some form in the `load` context.

**Describe alternatives you've considered**
I guess the main alternative would be not doing anything and letting it continue to guess and check, which does work generally, but aside from it being less effort I can't see a benefit there. Slightly less context used I suppose since parameter specifications are probably a little token heavy? Seems worth it IMO.

**Additional context**
This came out of investigations into the following issue: https://github.com/block/goose/issues/7273 which I will be closing momentarily in lieu of this one since running recipes within misc conversations is no longer a missing feature given the existence of Summon.

- [x] I have verified this does not duplicate an existing feature request

michaelneale:
could this sort of thing help? https://github.com/block/goose/pull/7297 thanks @jcwilk

jcwilk:
@michaelneale wow you're quick, I was piddling away at a less elegant version so I'll take my learnings about the issue to your nicer PR instead 🙌

PR #7297, opened 2026-02-18
feat: include recipe parameter details in Summon load context (#7288)

michaelneale:
candidate fix for https://github.com/block/goose/issues/7288

### DIFF
--- a/crates/goose/src/agents/platform_extensions/summon.rs
+++ b/crates/goose/src/agents/platform_extensions/summon.rs
@@ -15,7 +15,7 @@ use crate::config::Config;
 use crate::providers;
 use crate::recipe::build_recipe::build_recipe_from_template;
 use crate::recipe::local_recipes::load_local_recipe_file;
-use crate::recipe::{Recipe, Settings, RECIPE_FILE_EXTENSIONS};
+use crate::recipe::{Recipe, RecipeParameter, Settings, RECIPE_FILE_EXTENSIONS};
 use crate::session::extension_data::EnabledExtensionsState;
 use crate::session::SessionType;
 use anyhow::Result;
@@ -45,6 +45,7 @@ pub struct Source {
     pub description: String,
     pub path: PathBuf,
     pub content: String,
+    pub parameters: Option<Vec<RecipeParameter>>,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
@@ -71,9 +72,37 @@ impl std::fmt::Display for SourceKind {
 impl Source {
     /// Format the source content for loading into context
     pub fn to_load_text(&self) -> String {
+        let params_section = if let Some(params) = &self.parameters {
+            if !params.is_empty() {
+                let param_lines: Vec<String> = params
+                    .iter()
+                    .map(|p| {
+                        let mut line = format!(
+                            "- **{}** (type: {}, {}): {}",
+                            p.key, p.input_type, p.requirement, p.description
+                        );
+                        if let Some(default) = &p.default {
+                            line.push_str(&format!(" [default: {}]", default));
+                        }
+                        if let Some(options) = &p.options {
+                            if !options.is_empty() {
+                                line.push_str(&format!(" [options: {}]", options.join(", ")));
+                            }
+                        }
+                        line
+                    })
+                    .collect();
+                format!("\n\n### Parameters\n\n{}", param_lines.join("\n"))
+            } else {
+                String::new()
+            }
+        } else {
+            String::new()
+        };
+
         format!(
-            "## {} ({})\n\n{}\n\n### Content\n\n{}",
-            self.name, self.kind, self.description, self.content
+            "## {} ({})\n\n{}{}\n\n### Content\n\n{}",
+            self.name, self.kind, self.description, params_section, self.content
         )
     }
 }
@@ -174,6 +203,7 @@ fn parse_skill_content(content: &str, path: PathBuf) -> Option<Source> {
         description: metadata.description,
         path,
         content: body,
+        parameters: None,
     })
 }
 
@@ -195,6 +225,7 @@ fn parse_agent_content(content: &str, path: PathBuf) -> Option<Source> {
         description,
         path,
         content: body,
+        parameters: None,
     })
 }
 
@@ -604,7 +635,8 @@ impl SummonClient {
             }
             seen.insert(sr.name.clone());
 
-            let description = self.build_subrecipe_description(sr).await;
+            let (description, parameters) =
+                self.build_subrecipe_description_and_params(sr).await;
 
             sources.push(Source {
                 name: sr.name.clone(),
@@ -612,13 +644,23 @@ impl SummonClient {
                 description,
                 path: PathBuf::from(&sr.path),
                 content: String::new(),
+                parameters,
             });
         }
     }
 
-    async fn build_subrecipe_description(&self, sr: &crate::recipe::SubRecipe) -> String {
+    async fn build_subrecipe_description_and_params(
+        &self,
+        sr: &crate::recipe::SubRecipe,
+    ) -> (String, Option<Vec<RecipeParameter>>) {
         if let Some(desc) = &sr.description {
-            return desc.clone();
+            // Even with an explicit description, try to load params from the recipe file
+            if let Ok(recipe_file) = load_local_recipe_file(&sr.path) {
+                if let Ok(recipe) = Recipe::from_content(&recipe_file.content) {
+                    return (desc.clone(), recipe.parameters.clone());
+                }
+            }
+            return (desc.clone(), None);
         }
 
         if let Ok(recipe_file) = load_local_recipe_file(&sr.path) {
@@ -633,11 +675,11 @@ impl SummonClient {
                     }
                 }
 
-                return desc;
+                return (desc, recipe.parameters.clone());
             }
         }
 
-        format!("Subrecipe from {}", sr.path)
+        (format!("Subrecipe from {}", sr.path), None)
     }
 
     fn scan_recipes_dir(
@@ -682,6 +724,7 @@ impl SummonClient {
                         description: recipe.description.clone(),
                         path: path.clone(),
                         content: recipe.instructions.clone().unwrap_or_default(),
+                        parameters: recipe.parameters.clone(),
                     });
                 }
                 Err(e) => {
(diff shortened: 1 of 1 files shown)


crates/goose/src/agents/platform_extensions/summon.rs
```diff
@@ -604,21 +635,32 @@ impl SummonClient {
...
             });
         }
     }
 
-    async fn build_subrecipe_description(&self, sr: &crate::recipe::SubRecipe) -> String {
+    async fn build_subrecipe_description_and_params(
+        &self,
+        sr: &crate::recipe::SubRecipe,
+    ) -> (String, Option<Vec<RecipeParameter>>) {
         if let Some(desc) = &sr.description {
```
jcwilk, line 656: If I'm understanding the code correctly, I think the prior-existing short circuit was here mostly (maybe it's for testing purposes so you can pass in a value without needing the files to be all reconfigured?) to avoid needing to do a file lookup for performance reasons - so if we need to do the file lookup for the parameters anyways then I think the short circuit no longer has a purpose?

If so, maybe it would make more sense to put this condition inside the pre-existing `Ok(recipe_file)...Ok(recipe)` block since the two paths mostly share the same code, maybe maintaining the same "prioritize the in-memory recipe data and fallback to the file-stored values" with something like:
```
let description = sr
                    .description
                    .as_ref()
                    .cloned()
                    .unwrap_or_else(|| recipe.description.clone());
```

? Apologies if I'm overlooking something.

crates/goose/src/agents/platform_extensions/summon.rs
jcwilk, line 674: is it right to keep injecting param data into the description if we're now passing the parameters back separately and formatting them explicitly down in `handle_load_discovery`?

crates/goose/src/agents/platform_extensions/summon.rs
```diff
@@ -71,9 +72,37 @@ impl std::fmt::Display for SourceKind {
...
+            }
+        } else {
+            String::new()
+        };
+
         format!(
-            "## {} ({})\n\n{}\n\n### Content\n\n{}",
-            self.name, self.kind, self.description, self.content
+            "## {} ({})\n\n{}{}\n\n### Content\n\n{}",
+            self.name, self.kind, self.description, params_section, self.content
```
jcwilk, line 105: Does it still make sense to return the full content of a recipe in `to_load_text`? My understanding is that `delegate` uses a different flow, so when the recipe is actually "executed" it won't depend on this data flow. The content of a recipe may be substantial so I was thinking it might make sense to keep the content of the recipe "private" (much like an MCP server for example, you hopefully have a description and a list of parameters but you don't need the actual source code to be able to invoke it) especially since now that we're describing parameters explicitly, it's no longer necessary to surface the parameters via the content (and where they're interpolated etc)

crates/goose/src/agents/platform_extensions/summon.rs
```diff
@@ -1000,6 +1043,14 @@ impl SummonClient {
                         source.name,
                         truncate(&source.description, 60)
                     ));
+                    if let Some(params) = &source.parameters {
```
jcwilk, line 1046: what do you think about consolidating the "discovery" parameter presentation (here) with the "load" presentation (above, currently inlined in `to_load_text`)? Perhaps a shared function could take a flag for whether it should truncate param descriptions or not? Aside from truncating the description, which might hypothetically be long, it seems like we'd mostly want the same information available between the two WRT options, defaults, etc

jcwilk: here's one I've been playing with, which doesn't currently truncate at all:
```rust
fn format_parameters_section(parameters: Option<&[RecipeParameter]>) -> String {
    match parameters {
        None | Some([]) => "  Parameters: None".to_string(),
        Some(params) => {
            let mut out = String::from("  Parameters:\n");
            for p in params {
                let req = match &p.requirement {
                    RecipeParameterRequirement::Required => "required",
                    RecipeParameterRequirement::Optional | RecipeParameterRequirement::UserPrompt => {
                        "optional"
                    }
                };
                let type_str = match &p.input_type {
                    RecipeParameterInputType::Select => {
                        if let Some(ref opts) = p.options {
                            format!("select: {}", opts.join(", "))
                        } else {
                            "select".to_string()
                        }
                    }
                    other => format!("{}", other),
                };
                let default_suffix = p
                    .default
                    .as_ref()
                    .map(|d| format!(", default: \"{}\"", d))
                    .unwrap_or_default();
                let bullet = format!(
                    "{} ({}, {}{}): {}",
                    p.key, req, type_str, default_suffix, p.description
                );
                out.push_str(&format!("    • {}\n", bullet));
            }
            out.pop();
            out
        }
    }
}
```

It ends up as something like:
```
• goose-self-test - A comprehensive meta-testing recipe
  Parameters:
    • test_phases (optional, string): Which test phases to run
• hello-world - A sample recipe
  Parameters: None
```

The nesting is kind of awkward but was going for something that's tight and neat while very explicit/clear about what's being specified where.

Very hazy at best understanding of the precedence of where the specifics around these output formats come from though, so apologies if it seems like I'm imposing guesswork on you here - just trying to share some tentative findings in case they're helpful.

jcwilk:
Apologies for the deluge of comments, still getting my head around Rust/Goose best practices - and thank you for jumping on my issue so quickly :raised_hands: :raised_hands:

michaelneale:
no worries @jcwilk - the issue isn't totally clear to me but based on what you observed, are you able to try this branch to see if it makes that bit better? (then I will decide if should keep it or not). 

Thanks in advance!

crates/goose/src/agents/platform_extensions/summon.rs
michaelneale: that nesting is no where near as bad as I have seen elsewhere, it is fine!

michaelneale:
might tab @tlongwell-block who had the vision around "summon" consolidation... for any thoughts.

jcwilk:
@michaelneale sure thing! I'll try it when I get home - at the rust/goose meetup at block hq at the moment 😅 looking forward to seeing what y'all have in the pipes

jcwilk:
Okay, I made a somewhat contrived recipe that executes a compiled binary (ie, goose can't simply read the source code to figure out how the binary works) which does nothing except verify the passed parameters match allowed values, throwing an unhelpful error if they do not. Here's the recipe:

```
version: 1.0.0
title: API Client Generator
description: Generate a client library for your API specification.
instructions: |-
  Run the api-client-generator script. It expects exactly six positional arguments in order.
  Pass each parameter value as the corresponding argument (first parameter = first argument, second = second, and so on).
  The script is at ./api-client-generator

  1. language: {{ language }}
  2. style: {{ style }}
  3. async_mode: {{ async_mode }}
  4. output_path: {{ output_path }}
  5. spec_format: {{ spec_format }}
prompt: Execute the script by feeding the ordered parameters in as CLI arguments.
activities: []
parameters:
- key: language
  input_type: select
  requirement: required
  description: Target language
  options:
  - typescript
  - python
  - go
- key: style
  input_type: select
  requirement: required
  description: Code style
  options:
  - class-based
  - functional
  - minimal
- key: async_mode
  input_type: boolean
  requirement: optional
  description: Use async/await patterns
  default: 'true'
- key: output_path
  input_type: string
  requirement: required
  description: Where to write the generated client
- key: spec_format
  input_type: select
  requirement: required
  description: API spec format
  options:
  - openapi3
  - openapi2
```

Using gpt-4o I start a new chat with just developer+summon and tell it "Run the api generator recipe with reasonable values using summon's load/delegate tooling" and it `load`s (the discovery-based load doesn't include select value options) and then runs the subagent via delegate, the subagent attempts to run the script a few times with the bogus values it got passed from the parent agent and then gives up. If I paste in the full recipe yaml then it's able to call `delegate` successfully with valid parameters.

I realize we're getting into a bit of an edge case here, but it just seemed to me that if goose has a great spec for storing and validating recipe parameters then we might as well try to get the agent to be able to fully leverage that. But maybe it's a balancing act between keeping the `load`s lean too?

jcwilk:
In case you want the binary, just compile with `rustc api-client-generator.rs -o api-client-generator`

```api-client-generator.rs
//! API client generator - expects 5 positional args: language style async_mode output_path spec_format

use std::env;
use std::process::exit;

fn fail() -> ! {
    eprintln!("failed due to parameters!");
    exit(1);
}

fn main() {
    let args: Vec<String> = env::args().skip(1).collect();
    if args.len() != 5 {
        fail();
    }

    let (lang, style, async_mode, output_path, spec_format) = (
        &args[0],
        &args[1],
        &args[2],
        &args[3],
        &args[4],
    );

    let valid_lang = ["typescript", "python", "go"].contains(&lang.as_str());
    let valid_style = ["class-based", "functional", "minimal"].contains(&style.as_str());
    let invalid_combo = lang == "go" && style == "class-based";
    let valid_async = ["true", "false"].contains(&async_mode.as_str());
    let valid_format = ["openapi3", "openapi2"].contains(&spec_format.as_str());

    if !valid_lang || !valid_style || invalid_combo || !valid_async || output_path.is_empty()
        || !valid_format
    {
        fail();
    }

    println!(
        "OK: {} {} {} {} {}",
        lang, style, async_mode, output_path, spec_format
    );
}
```

tlongwell-block:
/goose I thought we already had the parameters for subrecipes in the MOIM. Is that right? Are we missing that for recipes?

jcwilk:
@tlongwell-block - I think I'm a bit overbudget on taking up y'all's time on this issue, so I'll try to be concise.

Goose's message was correct that both in `main` and this branch, no recipe information is included in the MOIM, the agent must invoke `Summon.load` - before `Summon.load` was added there was no way for the agent to get any recipe information at all.

Goose's message glossed over an important detail though, there's two different ways that `Summon.load` can be invoked:
* `load()` - Retrieve all recipes (and other things) for discovery. ALL agents which work with recipes in any way MUST call this unless they already have a nudge towards a specifically named recipe. This PR adds *partial* exposure of parameters to these results. (an improvement! however, things like defaults and option values are not exposed and descriptions are tightly truncated)
* `load(recipe)` - Retrieve full details for one recipe in particular. This PR adds full exposure of the parameters to these results. However, agents don't necessarily call this if they think they already have all the information from `load()` as illustrated by my above evidence. Also, it not only includes the full description and parameters, but it also includes the full *content* of the recipe which seems like wasted tokens/headspace now that we're exposing the full parameters specification.

FWIW I don't see harm in the PR as-is, it's an improvement vs `main`, just figured it's worth going the whole 9 yards while we have it in our heads but I'll defer to y'all's judgement about what that is exactly. Thanks for your time.

lifeizhou-ap:
Hey @michaelneale 👋 Just checking in on this draft. Are you still actively working on it? If not, would you mind closing it? We can always reopen later if you pick it back up. Thanks!

jcwilk:
@michaelneale Closing the loop from my side -- I invested a good amount of effort testing this branch and documenting remaining gaps, and it looks like the PR was shelved without any follow-up. Totally understand priorities shift, but a brief "we're moving on" would have gone a long way. Flagging in case it's useful for contributor engagement going forward. Thanks for keeping Goose open, and I wish y'all well :wave:

PR #7300, opened 2026-02-18
fix(ui): revert app-driven iframe width and send containerDimensions per ext-apps spec

### LINKED ISSUE #6818: Resize goose desktop shrink the mcp app and it never grows back.

aharvard:
## Summary

Fixes #6818 — Reverts #6376 and implements `containerDimensions` in the host context per the [ext-apps spec](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx#container-dimensions).

### Commit 1: Revert app-driven iframe width from #6376

PR #6376 let apps drive the container width via `size-changed` notifications. Per the spec and [owtaylor's analysis](https://github.com/block/goose/issues/6818#issuecomment-3873557786), the host owns width in inline mode — apps should not resize it. This commit removes:

- `iframeWidth` state and `setIframeWidth`
- Width handling from `handleSizeChanged` (height-only now)
- Conditional width styling — container is always `width: 100%`

### Commit 2: Send `containerDimensions` in hostContext

Adds a `getContainerDimensions()` function that maps display modes to the spec's dimension modes. Each axis is independently **fixed**, **flexible**, or **unbounded**:

| Mode | Spec Field | Meaning |
|---|---|---|
| Fixed | `width` / `height` | Host controls size, view fills it |
| Flexible | `maxWidth` / `maxHeight` | View controls size, up to the max |
| Unbounded | Field omitted | View controls size, no limit |

Display mode → layout mapping:

| Display Mode | Width | Height |
|---|---|---|
| `inline` | fixed | unbounded |
| `fullscreen` | fixed | fixed |
| `standalone` | fixed | fixed |
| `pip` | fixed | fixed |

Container dimensions are measured from the actual container element via `ResizeObserver` and update reactively on resize.

Types (`DimensionMode`, `DimensionLayout`) are exported from `McpApps/types.ts`.

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] `npx eslint` clean
- [x] `npx vitest run` — 340/340 tests pass
- [x] Open an MCP app inline — should fill chat width, grow vertically
- [x] Resize the Goose window — app width should follow
- [x] Open an MCP app in standalone window — should fill the window
- [x] Resize standalone window — app should fill new size

### DIFF
--- a/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
+++ b/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
@@ -51,8 +51,7 @@ const DISPLAY_MODE_LAYOUTS: Record<GooseDisplayMode, DimensionLayout> = {
   fullscreen: { width: 'fixed', height: 'fixed' },
   standalone: { width: 'fixed', height: 'fixed' },
   pip: { width: 'fixed', height: 'fixed' },
-  // Example: a sidebar where the app controls its height up to the visible area.
-  // sidecar: { width: 'fixed', height: 'flexible' },
+  // sidecar: { width: 'fixed', height: 'flexible' }, // example on how to use flexible layout
 };
 
 function getContainerDimensions(
@@ -63,8 +62,11 @@ function getContainerDimensions(
   const layout = DISPLAY_MODE_LAYOUTS[displayMode] ?? DISPLAY_MODE_LAYOUTS.inline;
 
   // Only require a measurement for axes that are fixed or flexible (unbounded axes are omitted).
-  if (layout.width !== 'unbounded' && measuredWidth <= 0) return undefined;
-  if (layout.height !== 'unbounded' && measuredHeight <= 0) return undefined;
+  if (
+    (layout.width !== 'unbounded' && measuredWidth <= 0) ||
+    (layout.height !== 'unbounded' && measuredHeight <= 0)
+  )
+    return undefined;
 
   const widthDimension = (() => {
     switch (layout.width) {
@@ -431,14 +433,11 @@ export default function McpAppRenderer({
     []
   );
 
-  const handleSizeChanged = useCallback(
-    ({ height }: McpUiSizeChangedNotification['params']) => {
-      if (height !== undefined && height > 0) {
-        setIframeHeight(height);
-      }
-    },
-    []
-  );
+  const handleSizeChanged = useCallback(({ height }: McpUiSizeChangedNotification['params']) => {
+    if (height !== undefined && height > 0) {
+      setIframeHeight(height);
+    }
+  }, []);
 
   // Track the container's pixel dimensions so we can report them to apps via containerDimensions.
   useEffect(() => {
@@ -447,10 +446,9 @@ export default function McpAppRenderer({
 
     const observer = new ResizeObserver((entries) => {
       for (const entry of entries) {
-        const width = Math.round(entry.contentBoxSize[0]?.inlineSize ?? entry.contentRect.width);
-        const height = Math.round(entry.contentBoxSize[0]?.blockSize ?? entry.contentRect.height);
-        setContainerWidth((prev) => (prev !== width ? width : prev));
-        setContainerHeight((prev) => (prev !== height ? height : prev));
+        const { width, height } = entry.contentRect;
+        setContainerWidth((prev) => (prev !== Math.round(width) ? Math.round(width) : prev));
+        setContainerHeight((prev) => (prev !== Math.round(height) ? Math.round(height) : prev));
       }
     });
 


ui/desktop/src/components/McpApps/McpAppRenderer.tsx
```diff
@@ -453,7 +509,7 @@ export default function McpAppRenderer({
       displayMode: displayMode as McpUiDisplayMode,
       availableDisplayModes:
         displayMode === 'standalone' ? [displayMode as McpUiDisplayMode] : AVAILABLE_DISPLAY_MODES,
-      // todo: containerDimensions: {} (depends on displayMode)
+      containerDimensions: getContainerDimensions(displayMode, containerWidth, containerHeight),
```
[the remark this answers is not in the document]
aharvard: goose is a superset, won't change

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
```diff
@@ -39,12 +39,56 @@ import {
...
+
+function getContainerDimensions(
+  displayMode: GooseDisplayMode,
+  measuredWidth: number,
+  measuredHeight: number
+): McpUiHostContext['containerDimensions'] {
+  if (measuredWidth <= 0 || measuredHeight <= 0) return undefined;
+
+  const layout = DISPLAY_MODE_LAYOUTS[displayMode] ?? DISPLAY_MODE_LAYOUTS.inline;
+
```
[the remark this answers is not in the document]
aharvard: addressed

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
```diff
@@ -39,12 +39,58 @@ import {
...
 
 const AVAILABLE_DISPLAY_MODES: McpUiDisplayMode[] = ['inline'];
 
+const DISPLAY_MODE_LAYOUTS: Record<GooseDisplayMode, DimensionLayout> = {
+  inline: { width: 'fixed', height: 'unbounded' },
+  fullscreen: { width: 'fixed', height: 'fixed' },
+  standalone: { width: 'fixed', height: 'fixed' },
+  pip: { width: 'fixed', height: 'fixed' },
+  // Example: a sidebar where the app controls its height up to the visible area.
+  // sidecar: { width: 'fixed', height: 'flexible' },
```
DOsinga: not sure what this comment is for, maybe dorp?

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
```diff
@@ -39,12 +39,58 @@ import {
...
+function getContainerDimensions(
+  displayMode: GooseDisplayMode,
+  measuredWidth: number,
+  measuredHeight: number
+): McpUiHostContext['containerDimensions'] {
+  const layout = DISPLAY_MODE_LAYOUTS[displayMode] ?? DISPLAY_MODE_LAYOUTS.inline;
+
+  // Only require a measurement for axes that are fixed or flexible (unbounded axes are omitted).
+  if (layout.width !== 'unbounded' && measuredWidth <= 0) return undefined;
+  if (layout.height !== 'unbounded' && measuredHeight <= 0) return undefined;
```
DOsinga: we could collapse these into one if

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
```diff
@@ -383,23 +431,33 @@ export default function McpAppRenderer({
...
+  useEffect(() => {
+    const el = containerRef.current;
+    if (!el) return;
+
+    const observer = new ResizeObserver((entries) => {
+      for (const entry of entries) {
+        const width = Math.round(entry.contentBoxSize[0]?.inlineSize ?? entry.contentRect.width);
+        const height = Math.round(entry.contentBoxSize[0]?.blockSize ?? entry.contentRect.height);
+        setContainerWidth((prev) => (prev !== width ? width : prev));
+        setContainerHeight((prev) => (prev !== height ? height : prev));
```
DOsinga: eh, isn't that just width/height

ui/desktop/src/components/McpApps/types.ts
```diff
@@ -14,6 +14,19 @@ export type SandboxPermissions = string;
...
+ *   flexible  – sends maxWidth/maxHeight (view controls, up to max; host resizes via size-changed)
+ *   unbounded – field omitted (view controls with no limit; host resizes via size-changed)
+ */
+export type DimensionMode = 'fixed' | 'flexible' | 'unbounded';
+
+export interface DimensionLayout {
+  width: DimensionMode;
+  height: DimensionMode;
+}
+
```
DOsinga, line 29: shouldn't we get these from the server spec?

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
aharvard: The spec's rules for when to use width/height vs maxWidth/maxHeight vs omitting fields are a bit hard to follow, so I abstracted it into a layout map using more intuitive terms: fixed, flexible, unbounded. We don't have a flexible case yet, but I wanted to capture it now so we don't have to re-derive it later.

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
aharvard: 👍 done

ui/desktop/src/components/McpApps/McpAppRenderer.tsx
aharvard: yep, simplified to just contentRect.width/height — we don't need logical properties for vertical writing modes.

ui/desktop/src/components/McpApps/types.ts
aharvard: The spec exports the `containerDimensions` shape on `McpUiHostContex`t but not a `DimensionMode`/`DimensionLayout` type — those are our internal mapping for deciding which fields to populate per display mode. If the spec ever exports them we'll switch.

PR #7322, opened 2026-02-18
fix(goose-acp): heap allocations

alexhancock:
`goose-acp` frequently errors out now when clients connect to it via HTTP

```
2026-02-18T14:44:06.016245Z  INFO goose_acp::transport::http: Session created acp_session_id=565b645e-dd6d-4cf9-94d4-0ceff7fd0016

thread 'tokio-runtime-worker' (7050194) has overflowed its stack
zsh: abort      cargo run -p goose-acp
```

this addresses it by `Box`ing things

### DIFF
--- a/crates/goose-acp/src/server.rs
+++ b/crates/goose-acp/src/server.rs
@@ -1193,112 +1193,125 @@ impl JrMessageHandler for GooseAcpHandler {
         "goose-acp"
     }
 
-    async fn handle_message(
+    fn handle_message(
         &mut self,
         message: MessageCx,
         cx: JrConnectionCx<AgentToClient>,
-    ) -> Result<Handled<MessageCx>, sacp::Error> {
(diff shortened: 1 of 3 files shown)


crates/goose-acp/src/server.rs
```diff
@@ -1005,103 +1005,116 @@ impl JrMessageHandler for GooseAcpHandler {
...
-                        _ => Err(sacp::Error::method_not_found()),
                     }
-                }
-            })
-            .await
-            .map(|()| Handled::Yes)
+                })
+                .await
+                .map(|()| Handled::Yes)
+        })
```
jamadeo, line 1291: The copilot explanation is convincing. That is quite a lot of .awaits, each one growing the size of the state machine. 85kb is quite a lot less than the [default 2mb](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.thread_stack_size) though.

jamadeo (approved):

PR #7333, opened 2026-02-18
Improve link confirmation modal

zanesq:
## Summary
Fixed the confirm-open-link modal not scrolling when URLs are very long, 
making the action buttons unreachable.

Replaced the native Electron `showMessageBox` dialog with the existing React 
`ConfirmationModal` component for the external link confirmation flow. The 
`ConfirmationModal` now supports an optional scrollable detail section with 
max viewport height constraints, so long URLs wrap and scroll while the 
Cancel/Open buttons always remain visible and accessible.

Also added error handling for when no application is found to open the 
link protocol, and improved keyboard focus visibility on the modal buttons 
using the theme accent color instead of the barely-visible default ring.

Screenshot 2026-02-18 at 2 57 53 PM
Screenshot 2026-02-18 at 2 57 42 PM


lifeizhou-ap (approved):
Nice!

PR #7304, opened 2026-02-18
feat: add Moonshot and Kimi Code declarative providers

DOsinga:
## Summary

Add declarative provider configurations for Moonshot AI (Kimi) models, combining the work from PRs #6962 and #7222.

### New Provider Files
- **moonshot.json**: General Moonshot/Kimi models
  - kimi-latest (131K context)
  - kimi-thinking-preview (131K context)
  - kimi-k2-0711 (131K context)
  - kimi-k2 (262K context)
  - moonshot-v1-8k (8K context)
  - moonshot-v1-32k (32K context)

- **kimi.json**: Coding-optimized Kimi models
  - kimi-for-coding (262K context)
  - kimi-code (262K context)

Both providers use the Moonshot API (`api.moonshot.cn`) with `MOONSHOT_API_KEY`.

## Related PRs
- Closes #6962 (feat: support for Moonshot LLMs)
- Closes #7222 (feat: add Kimi Code provider)
- #7294 (already merged - reasoning field alias support)

## Notes
The reasoning_content handling changes from #6962 and #7222 are already in main (likely from #7294 or earlier work), so this PR only adds the missing declarative provider configurations.

## Testing
- `cargo build -p goose` ✅

PR #7339, opened 2026-02-19
feat: add Cerebras provider support

Kvadratni:
## Summary
- Add Cerebras as a provider using OpenAI-compatible API
- Supports production models (llama3.1-8b, gpt-oss-120b) and preview models (qwen-3-235b-a22b-instruct-2507, zai-glm-4.7)
- Configuration: CEREBRAS_API_KEY (required), CEREBRAS_HOST (optional)

## Testing
- Verified with curl and goose CLI
- All models work correctly

### DIFF
--- a/crates/goose/src/providers/declarative/cerebras.json
+++ b/crates/goose/src/providers/declarative/cerebras.json
@@ -1,0 +1,27 @@
+{
+  "name": "cerebras",
+  "engine": "openai",
+  "display_name": "Cerebras",
+  "description": "Fast inference on Cerebras wafer-scale engines",
+  "api_key_env": "CEREBRAS_API_KEY",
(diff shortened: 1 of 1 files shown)


crates/goose/src/providers/cerebras.rs
```diff
@@ -0,0 +1,67 @@
+use super::api_client::{ApiClient, AuthMethod};
```
alexhancock: Only Q I have would be whether this could be implemented the declarative way, with a json only approach

Like https://github.com/block/goose/pull/7304

Kvadratni: oh cool. will do

katzdave (approved):

PR #7344, opened 2026-02-19
fix(ci): deflake smoke tests for Google models

michaelneale:
## Problem

The Live Provider Tests Smoke Tests have been flaking heavily on Google models (`gemini-2.5-pro` and `gemini-3-flash-preview`), causing ~75% of all Smoke Test failures across PRs. These are all flakes — re-triggered runs pass.

The root cause is the uppercase transformation prompt:

> *Use the text_editor view command to read ./input.txt, then output this file's contents in UPPERCASE*

Gemini models interpret "output in UPPERCASE" as a style instruction for their response rather than a transformation of specific file content. They read the file correctly but then hallucinate uppercase text instead:
- `HELLO. I AM A LARGE LANGUAGE MODEL. I AM GOOSE...`
- `HELLO WORLD. THIS IS A TEST. THE QUICK BROWN FOX...`
- `INPUT.TXT-ABC123` (uppercased the filename, not the content)

## Fix

Replace the uppercase transformation test with a two-file read-back test using random tokens per run (`smoke-alpha-$RANDOM`, `smoke-bravo-$RANDOM`).

This still verifies:
1. **Tool use** — `text_editor` must be called (same grep check as before)
2. **Actual file reading** — random tokens can't be guessed or hallucinated, so their presence in the output proves the model read the files

No transformation needed, just echo back. The prompt asks the model to reply with ONLY the file contents — much less ambiguous than asking for a case transformation.

### DIFF
--- a/scripts/test_providers.sh
+++ b/scripts/test_providers.sh
@@ -23,9 +23,9 @@ run_test() {
     cp "$TEST_FILE" "$testdir/test-content.txt"
     prompt="read ./test-content.txt and output its contents exactly"
   else
-    # Write two files with unique tokens, ask model to read both and combine them.
-    # This tests tool use (must call text_editor twice) and proves the model actually
-    # read the file contents (must output both tokens which only exist in those files).
+    # Write two files with unique random tokens. Validation checks that text_editor
+    # was used and that both tokens appear in the output, proving the model actually
+    # read the files (random tokens can't be guessed or hallucinated).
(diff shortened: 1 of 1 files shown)


michaelneale:
@DOsinga the perfect one for an agent to solve - just told it to look through history for patterns, make a branch and try some things (and made sure it stayed true to intent of test!)

PR #7349, opened 2026-02-19
Use prompt for no tools

DOsinga:
## Summary

Only replace system prompts when we have tools

/cc @katzdave can we get from the models.dev the info whether the model supports toolcalling? then we could do this on a higher level

### DIFF
--- a/crates/goose/src/providers/local_inference.rs
+++ b/crates/goose/src/providers/local_inference.rs
@@ -622,7 +622,10 @@ impl Provider for LocalInferenceProvider {
         // Models that support native OpenAI-compatible tool-call JSON use the
         // native path (template-based tool calling with JSON output). All other
         // models use the emulator which parses `$ command` and ```execute blocks.
-        let use_emulator = !model_settings.native_tool_calling;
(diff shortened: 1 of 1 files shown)


katzdave:
> ## Summary
> Only replace system prompts when we have tools
> 
> /cc @katzdave can we get from the models.dev the info whether the model supports toolcalling? then we could do this on a higher level

The data for local models is pretty sharded between many different providers, and mostly has larger ones that are worth serving on the cloud.

```
{
    "id": "ollama-cloud/qwen3-coder:480b",
    "name": "qwen3-coder:480b",
    "family": "qwen",
    "attachment": false,
    "reasoning": false,
    "tool_call": true,
    "release_date": "2025-07-22",
    "last_updated": "2026-01-19",
    "modalities": {
      "input": [
        "text"
      ],
      "output": [
        "text"
      ]
    },
    "open_weights": true,
    "cost": {},
    "limit": {
      "context": 262144,
      "output": 65536
    }
  },
```
The [openrouter json ](https://openrouter.ai/api/v1/models) had a bit more coverage + keyed on things like qwen, so I'd like to mix that in.

Issue #6642, opened 2026-01-22
project: goosed to ACP-over-HTTP


alexhancock:
**Goal:** Replace goosed's custom SSE-based streaming API with a standards-based ACP (Agent Communication Protocol) over HTTP interface, enabling better client portability, protocol standardization, and eventually a unified interface for all goose clients.

### Current Architecture

- **goose-cli** (Rust): Direct in-process agent communication
- **goosed** (goose-server): Custom REST + SSE streaming API consumed by the Electron desktop app

### Proposed Architecture

A new **ACP-over-HTTP** server that:
- Implements the standardized ACP (Agent Communication Protocol) specification
- Uses JSON-RPC 2.0 over HTTP with SSE streaming as the transport
- Enables any ACP-compatible client to interact with goose

## Prototype

See the prototype branch: https://github.com/block/goose/tree/alexhancock/goosed-acp-and-new-cli
Example of single agent (video):  [ref](https://discord.com/channels/1287729918100246654/1408153538537721966/1461444855543762986)
Example of multi-agent orchestration (screenshots): [ref](https://discord.com/channels/1287729918100246654/1408153538537721966/1463661278202695754)

---

## Phase 1: Stabilize ACP Server

**Goal:** Production-ready ACP server that can fully replace goosed

### Tasks

- [X] Session persistence and resumption (`session/load`)
- [x] Enough goosed features to power alternative alpha CLI (extensions, modalities, streaming, etc)

### Success Criteria
- Core goosed features available via ACP-over-HTTP
- Integration tests passing

---

## Phase 2: TypeScript TUI Alpha

**Goal:** Feature-complete TUI which will be an evolution of the current CLI

### Tasks

- [ ] Core features of existing CLI supported
  - [X] Messages & tool calling
  - [ ] Extensions (add, remove, edit config)
  - [ ] Session (list, resume, export)
  - `-t` and `-i` modes for zero-shotting
- [ ] Make MCP features work (sampling, elicitation, etc)
- [X] UX polish (syntax highlighting, markdown rendering, diffs, themes?)
- [X] Build process that makes a single binary

### Success Criteria
- TUI usable as daily driver
- Docs

---

## Phase 3: Desktop Update

**Goal:** Updated desktop app which uses goose-acp

### Tasks

- [ ] Integrate an ACP client (using `goose acp` over stdio or `goose serve` with http/ws into new desktop app

### Success Criteria
- Desktop app multi-chat works with goose's ACP support as the server

---

[...]

PR #7362, opened 2026-02-19
feat: TUI client of goose-acp

alexhancock:
A new TUI that serves as the first goose-acp client to use the streamable http transport, this PR the basis for phase II of this plan https://github.com/block/goose/issues/6642

## Demo

Screenshot 2026-02-19 at 12 48 30 PM

https://github.com/user-attachments/assets/88302b45-a4c4-45f7-8ed3-8f0d17e56016

## Detail

* Imports the new `GooseClient` and types from `ui/acp` which we will use as the client SDK for interacting with `goose-acp` over the HTTP transport. (But doesn't use it yet as everything the TUI does is covered by the standard TS ACP SDK. We should consider routing the standard things through GooseClient too, and I or @jamadeo can do that in a followup.
* Implements a fullscreen TUI experience
* Supports rendering messages + tool calls and interacting with a goose agent using ACP over HTTP

### DIFF
--- a/ui/acp/generate-schema.ts
+++ b/ui/acp/generate-schema.ts
@@ -28,7 +28,6 @@ main().catch((err) => {
 async function main() {
   const schemaSrc = await fs.readFile(SCHEMA_PATH, "utf8");
   const jsonSchema = JSON.parse(
-    // Convert JSON Schema $defs refs to OpenAPI component refs
     schemaSrc.replaceAll("#/$defs/", "#/components/schemas/"),
   );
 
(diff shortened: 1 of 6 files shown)


ui/text/README.md
```diff
@@ -0,0 +1,16 @@
+# goose ACP TUI
+
+Early stage and part of goose's broader move to ACP
+
+https://github.com/block/goose/issues/6642
+https://github.com/block/goose/discussions/7309
+
+## Running
+
+1. Run the server `cargo run -p goose-acp --bin goose-acp-server`
```
jamadeo, line 10: What's your thinking on how this should work in the future? should the tui app manage a goose acp process?

alexhancock: Yeah I think we'll do two things

1. Provide a version with it bundled into a single executable
2. Let people `npx @block/goose --server $URL`

PR #7373, opened 2026-02-20
Option to add changeable defaults in goose-releases

dorien-koelemeijer:
## Summary
Add option to add changeable defaults in goose-releases. Currently, there's an option to add defaults in `init-config` and `bundles.py` - the former will only add defaults when there is no config file, and the latter will override everything (even the user's config). This PR adds defaults that will be added if the user hasn't set certain variables in their local config file, but will never override the user's config. 

A PR will be made in goose-releases to add the defaults.yaml file and add some initial variables. 

### Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [x] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing

Local/manual testing using `just run-ui`

### DIFF
--- a/crates/goose/src/config/base.rs
+++ b/crates/goose/src/config/base.rs
@@ -1046,7 +1046,6 @@ fn find_workspace_or_exe_root() -> Option<PathBuf> {
         path = parent.to_path_buf();
     }
 
-    // No workspace found, return directory containing the executable
     Some(exe_dir)
 }
 
@@ -1791,7 +1790,6 @@ mod tests {
             defaults_file.path(),
         )
         .unwrap();
-        // Return defaults_file to keep it alive (NamedTempFile deletes on drop)
         (config, defaults_file)
     }
 


michaelneale:
yep I think this is ok - did you test it with the app packaged via `just make-ui` - as behavior of file finding can vary, but I think this makes sense. May want to document it too for people packaging things up. Do you think this could also possibly retire the old .py script which hard codes in variables?

dorien-koelemeijer:
> yep I think this is ok - did you test it with the app packaged via `just make-ui` - as behavior of file finding can vary, but I think this makes sense. May want to document it too for people packaging things up. Do you think this could also possibly retire the old .py script which hard codes in variables?

I will test again to make sure. I don't think we can retire the old bundles.py script, because it has its own purpose (hardcoding env vars that the user should not be able to change). I will confirm this though to make 100% sure that this is the case.

Issue #7364, opened 2026-02-19
Summon extension causes agent to get confused when using async


alexyao2015:
**Describe the bug**
Since https://github.com/block/goose/pull/6964

Using gpt-5.2-codex and telling the agent to use a subagent to call a tool in an mcp server, it gets confused when it delegates the task then immediately tries to load it. The agent then calls a sleep command of a random number of seconds. The extension should instead notify the agent when the task is done instead of relying on the agent to guess when the task is done.

See the logs of the agent attempting to load the result of a task and it just sleeping
```
  ▸ load
    source: 20260219_5
    cancel: false

Annotated {
    raw: Text(
        RawTextContent {
            text: "Error: Task '20260219_5' is still running. Use load(source: \"20260219_5\", cancel: true) to stop.",
            meta: None,
        },
    ),
    annotations: None,
}

  ▸ shell developer
    command: sleep 296

-32603: request timeout after PT30S
  ▸ shell developer
    command: sleep 1

Annotated {
    raw: Text(
        RawTextContent {
            text: "",
            meta: None,
        },
    ),
    annotations: Some(
        Annotations {
            audience: Some(
                [
                    User,
                ],
            ),
            priority: Some(
                0.0,
            ),
            last_modified: None,
        },
    ),
}

  ▸ load
    source: 20260219_5
    cancel: false
```

alexhancock:
I suspect https://github.com/block/goose/pull/7377 resolved this

PR #7377, opened 2026-02-20
fix(summon): stop MOIM from telling models to sleep while waiting for tasks

rabi:
## Summary

The MOIM hint for running background tasks injected "sleep N to wait" into the conversation, causing some models to call `developer__shell: sleep N` which times out or wastes the entire turn. Replace with guidance to use load() which waits for the task result.

Also stop load() from killing async delegate tasks after 300s — return the task to the background and let the model decide whether to wait again or cancel.

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested locally.

### Related Issues
Relates to #7364

### DIFF
--- a/crates/goose/src/agents/platform_extensions/summon.rs
+++ b/crates/goose/src/agents/platform_extensions/summon.rs
@@ -330,10 +330,12 @@ impl SummonClient {
             "Load knowledge into your current context or discover available sources.\n\n\
              Call with no arguments to list all available sources (subrecipes, recipes, skills, agents).\n\
              Call with a source name to load its content into your context.\n\
-             For background tasks: load(source: \"task_id\", cancel: true) stops and returns output.\n\n\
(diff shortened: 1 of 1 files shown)


crates/goose/src/agents/platform_extensions/summon.rs
```diff
@@ -896,45 +898,39 @@ impl SummonClient {
                 }
             }
 
-            let description = task.description.clone();
-            let mut handle = task.handle;
+            let mut task = task;
```
jh-block: nit: the `task` binding could just be declared `mut` where it's created above.

PR #7380, opened 2026-02-20
feat(desktop): add minimal i18n foundation for desktop UI strings

bavadim:
## Summary
This PR adds a minimal, low-maintenance i18n foundation for Goose Desktop without introducing new maintained locales in upstream.

What is included:
- A lightweight translation runtime in `ui/desktop/src/i18n/index.ts`
- English dictionary `ui/desktop/src/i18n/en.json` as the canonical default
- UI string wiring to `t('key', 'fallback')` across key desktop surfaces
- Locale-aware date/time formatting via `GOOSE_LOCALE` (`currentLocaleTag()`)
- i18n unit tests in `ui/desktop/src/i18n/i18n.test.ts`
- Documentation update in `CUSTOM_DISTROS.md` for localization in custom distros

What is intentionally NOT included:
- No upstream commitment to maintain multiple languages now
- No language selector UI
- No additional locale catalogs in this PR
- No CI guardrails/check scripts

Why this should be accepted now:
- Keeps upstream default behavior (English-first, safe fallbacks)
- Reduces hardcoded UI text and creates a stable path for downstream/fork localization
- Minimal operational burden on core maintainers
- Aligns with current project velocity: infrastructure first, language packs later as optional follow-ups

### Type of Change
- [x] Feature
- [ ] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [x] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
- `cd ui/desktop && npm run test:run -- src/i18n/i18n.test.ts`
- `cd ui/desktop && npm run typecheck`
- Manual smoke review of updated UI surfaces using translation function fallbacks

### Related Issues
Relates to #2376  
Discussion: https://github.com/block/goose/issues/2376  
Discussion: https://github.com/block/goose/pull/4230

Before:  
UI strings were hardcoded in many desktop components.

After:  
UI strings are routed through i18n keys with English fallback, and locale tag is respected for date/time formatting.

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -451,14 +451,6 @@ const getBundledConfig = (): BundledConfig => {
 const { defaultProvider, defaultModel, predefinedModels, baseUrlShare, version } =
   getBundledConfig();
 
-const resolveGoosePathRoot = (): string | undefined => {
-  const pathRoot = process.env.GOOSE_PATH_ROOT?.trim();
-  if (pathRoot) {
-    return expandTilde(pathRoot);
-  }
-  return undefined;
-};
-
 const GENERATED_SECRET = crypto.randomBytes(32).toString('hex');
 
 const getServerSecret = (settings: Settings): string => {
@@ -476,7 +468,6 @@ let appConfig = {
   GOOSE_DEFAULT_MODEL: defaultModel,
   GOOSE_PREDEFINED_MODELS: predefinedModels,
   GOOSE_API_HOST: 'http://127.0.0.1',
-  GOOSE_PATH_ROOT: resolveGoosePathRoot(),
   GOOSE_WORKING_DIR: '',
   // If GOOSE_ALLOWLIST_WARNING env var is not set, defaults to false (strict blocking mode)
   GOOSE_ALLOWLIST_WARNING: process.env.GOOSE_ALLOWLIST_WARNING === 'true',
@@ -507,9 +498,7 @@ const createChat = async (
   const goosedResult = await startGoosed({
     serverSecret,
     dir: dir || os.homedir(),
-    env: {
-      GOOSE_PATH_ROOT: appConfig.GOOSE_PATH_ROOT as string | undefined,
-    },
+    env: { GOOSE_PATH_ROOT: process.env.GOOSE_PATH_ROOT },
     externalGoosed: settings.externalGoosed,
     isPackaged: app.isPackaged,
     resourcesPath: app.isPackaged ? process.resourcesPath : undefined,


ui/desktop/src/i18n/index.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+export function currentLocale(): string {
+  return resolveLocale(getConfiguredLocaleRaw(), Object.keys(dictionaries));
+}
+
+export function currentLocaleTag(): string {
+  const raw = getConfiguredLocaleRaw();
+  if (raw) {
+    return normalizeLocaleTag(raw);
```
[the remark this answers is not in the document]
bavadim: Good catch, fixed in `fix(desktop): validate locale tag before timestamp formatting`.

Changes:
- Strip POSIX locale suffixes (like `.UTF-8` and `@modifier`) before using locale tags
- Validate/canonicalize locale tags via `Intl.getCanonicalLocales`
- Fallback to `en-US` when locale is invalid
- Added tests for `en_US.UTF-8` normalization and invalid tag fallback

This prevents `RangeError` in `toLocaleTimeString`/`toLocaleDateString` paths.

ui/desktop/src/components/GroupedExtensionLoadingToast.tsx
```diff
@@ -81,7 +93,10 @@ export function GroupedExtensionLoadingToast({
                   <div className="font-medium text-base">{getSummaryText()}</div>
                   {errorCount > 0 && (
                     <div className="text-sm opacity-90">
-                      {errorCount} extension{errorCount !== 1 ? 's' : ''} failed to load
+                      {t('extensions.failed_count', '{count} extension{suffix} failed to load', {
```
jh-block, line 96: It's not clear to me how this framework will support pluralisation in most languages, where a simple suffix is not enough. I might have missed something though. To give an example, in French this string would have one of these forms:

Singular: 1 extension n'a pas pu être chargée
Plural: 10 extensions n'ont pas pu être chargées

How would a hypothetical `i18n/fr.json` implement this translation?

ui/desktop/src/components/conversation/SearchBar.tsx
```diff
@@ -37,7 +38,7 @@ export const SearchBar: React.FC<SearchBarProps> = ({
   searchResults,
   inputRef: externalInputRef,
   initialSearchTerm = '',
-  placeholder = 'Search conversation...',
+  placeholder = t('common.search_conversation', 'Search conversation...'),
```
jh-block, line 41: This gets called at module load time; are we sure locale config is always already available then? It might be safer to move the t() call into the component body

ui/desktop/src/components/extensions/ExtensionsView.tsx
```diff
@@ -95,19 +96,23 @@ export default function ExtensionsView({
...
-              <h1 className="text-4xl font-light">Extensions</h1>
+              <h1 className="text-4xl font-light">{t('extensions.title', 'Extensions')}</h1>
             </div>
-            <p className="text-sm text-text-secondary mb-2">
-              These extensions use the Model Context Protocol (MCP). They can expand Goose's
-              capabilities using three main components: Prompts, Resources, and Tools.{' '}
-              {getSearchShortcutText()} to search.
+            <p className="text-sm text-text-muted mb-2">
+              {t(
+                'extensions.description',
```
jh-block, line 106: This key doesn't exist in `i18n/en.json` -- could we add a lint that all keys used in t() calls in code actually exist in the base translation file? If we have that, maybe we don't need to duplicate the default text in both the t() call and the translation file.

ui/desktop/src/components/GroupedExtensionLoadingToast.tsx
jh-block: Due to this issue I decided to take a stab at a different approach using ICU MessageFormat for the translation strings (which is flexible enough to support the wide variety of different plural/gender/etc complexities that different languages have, and also has the advantage of being a standard with a lot of tooling support)

That PR is in #8105 -- I'll close this one. Would welcome any review/discussion/help over there

PR #7381, opened 2026-02-20
Client settings

DOsinga:
## Summary

Move away from using localStorage

### DIFF
--- a/ui/desktop/src/contexts/ThemeContext.tsx
+++ b/ui/desktop/src/contexts/ThemeContext.tsx
@@ -7,7 +7,6 @@ interface ThemeContextValue {
   userThemePreference: ThemePreference;
   setUserThemePreference: (pref: ThemePreference) => void;
   resolvedTheme: ResolvedTheme;
-  isLoading: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


DOsinga:
Thanks for testing, @zanesq 

>Thanks for tackling this! Tested locally and looks good. Only piece of feedback from goose worth noting is:
Prototype Pollution Protection is Good, But (settings as any)[key] = value is Risky
The validSettingKeys allowlist in main.ts is a good security measure. However, the value parameter is completely unvalidated — a renderer compromise could write arbitrary data structures into any allowed key. For example, writing a massive blob to seenAnnouncementIds or a non-object to keyboardShortcuts.
Recommendation: Consider adding basic type validation per key (e.g., typeof value === 'boolean' for boolean keys, Array.isArray(value) for array keys). Not critical for v1, but worth a TODO.

yeah, I don't know - we're not protecting against an compromised renderer here. even checking the keys seems not all that useful since typescript should protect us. but that is something somebody might override easier, I think.

if you can think of a way to this with introspection rather than explicit, I'd be up for it

PR #7388, opened 2026-02-20
ACP editor tools

jamadeo:
This adds tools to use ACP's [file system](https://agentclientprotocol.com/protocol/file-system) and [terminals](https://agentclientprotocol.com/protocol/terminals) methods to call back to the ACP client to perform file edits and shell commands. It's enabled when the client advertises this capability.

This makes for a nicer experience when using goose over ACP in an editor like Zed.

Before:
https://github.com/user-attachments/assets/c3ee6a05-a4ba-411b-a881-02ae909b6274

After:
https://github.com/user-attachments/assets/247730ee-c83e-454f-b71c-b4b07630305a

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -937,7 +937,6 @@ impl ExtensionManager {
             .collect();
 
         let cancel_token = CancellationToken::default();
-        let num_clients = clients.len();
         let client_futures = clients.into_iter().map(|(name, config, client)| {
             let cancel_token = cancel_token.clone();
             let ext_name = name.clone();
@@ -954,7 +953,7 @@ impl ExtensionManager {
                     }
                 };
 
-                let expose_unprefixed = num_clients == 1 || is_unprefixed_extension(&config);
+                let expose_unprefixed = is_unprefixed_extension(&config);
 
                 loop {
                     for tool in client_tools.tools {


crates/goose-acp/src/tools.rs
```diff
@@ -0,0 +1,404 @@
...
+use rmcp::{
+    model::{CallToolResult, Content, InitializeResult, JsonObject, ListToolsResult, Tool},
+    ErrorData, ServiceError,
+};
+use sacp::{AgentToClient, JrConnectionCx};
+use schemars::{schema_for, JsonSchema};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use tokio_util::sync::CancellationToken;
+
+struct Capabilities {
```
rabi, line 22: Should we not use ClientCapabilities rather than having a separate struct?

crates/goose-acp/src/tools.rs
```diff
@@ -0,0 +1,400 @@
...
+        Ok(CallToolResult::success(vec![Content::text(content)]))
+    }
+
+    async fn write(
+        &self,
+        ctx: &goose::agents::ToolCallContext,
+        params: WriteParams,
+    ) -> Result<CallToolResult, ServiceError> {
+        let path = self.working_dir.join(params.path);
+        let content = read_file(&path, &self.cx, self.session_id.clone()).await?;
```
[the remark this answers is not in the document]
rabi: This seems like a valid concern, no?

crates/goose-acp/src/server.rs
```diff
@@ -343,9 +346,18 @@ impl GooseAcpAgent {
...
+        let exclude_developer = {
+            let capabilities = self.client_capabilities.lock().await;
+            capabilities.terminal
+                || capabilities.fs.read_text_file
+                || capabilities.fs.write_text_file
+        };
         let config_path = self.config_dir.join(CONFIG_YAML_NAME);
         if let Ok(config_file) = Config::new(&config_path, "goose") {
-            let extensions = get_enabled_extensions_with_config(&config_file);
+            let mut extensions = get_enabled_extensions_with_config(&config_file);
```
rabi, line 357: The `exclude_developer` check seems to filter only config-loaded extensions.  If passed via `--with-builtin developer`, it would create duplicate/conflicting tools? Should we remove the --with-builtin flag  completely now that user is expected to enable extensions via config and then we filter based on that?

crates/goose-acp/src/server.rs
```diff
@@ -60,6 +61,7 @@ pub struct GooseAcpAgent {
     goose_mode: goose::config::GooseMode,
     disable_session_naming: bool,
     builtins: Vec<String>,
+    client_capabilities: Mutex<ClientCapabilities>,
```
rabi, line 64: Maybe we use tokio::sync::OnceCell? As it's only set during initialize and then can be read freely?

jamadeo:
> If we all feel this is the way to go, then it's ok. My major reason for not approaching it this way as to not maintaining two different implementations of developer tools. If the plan is to move CLI/GUI to ACP clients, then this duplication is temporary and fine. Otherwise, a `DeveloperBackend` trait with capability-gated backend wiring would atleast keep the tool definitions in sync.

I agree with the reasoning, but one problem which we have now is that the ACP editor methods are a bit more constrained than how the new developer tools work in goose today. For example, if we want all files reads to go through ACP when you are connected to an editor, we would not want goose to read using the shell (`cat`/`head`/etc). But goose's standard developer tools now lack a readFile method entirely, meaning all reads happen with the shell.

This does mean that the tools seen by the model are different when you are using goose with an ACP-enabled editor vs as a standalone agent. And maybe that's okay? Arguably, as a user, I'd expect goose to be more autonomous when run on its own, and more interactive through an editor. But I'm open to arguments in the other direction.

codefromthecrypt:
@jamadeo https://github.com/block/goose/pull/7668 I raised last night swaps out the tool impl but re-uses exactly the same tool defs, so I *think* this balances the debate, correct me if I'm wrong. There's a whole lot more ACP work to complete, so hoping we can close this topic out. There are a lot more complicated things left and time budget is a thing or we'll never get through the spec.

rabi:
> But goose's standard developer tools now lack a readFile method entirely, meaning all reads happen with the shell.

@jamadeo Thanks. Yeah that's why I had implemented the read tool in #6940, so that tools remain the same but we decide how they behave based on the delegate registered.

jamadeo:
most of this is in #7668

PR #7393, opened 2026-02-20
Defer images to avoid interleaving with tool call messages

katzdave:
##  Summary
Fix Anthropic tool_use ids were found without tool_result blocks error when tool responses contain images

- When a tool response includes an image, the OpenAI-compatible format layers (databricks.rs,
openai.rs) must split it into a role: "tool" message (text only, since OpenAI tool messages don't
support images) plus a separate role: "user" message carrying the image. With multiple tool calls
returning images, these user image messages interleave with tool messages, breaking Anthropic's
requirement that all tool_result blocks appear consecutively after the assistant message.
- Add a post-processing pass in both format modules that defers user image messages until after
the tool group completes


jamadeo:
I think this is fixed now though I'm not entirely sure where - tested this with tool calls returning multiple images and multiple tool calls returning images / databricks provider.

PR #7395, opened 2026-02-20
Package ACP-based TUI

jamadeo:
There are several ways we could go about distributing [the ACP-based TUI](https://github.com/block/goose/tree/main/ui/text).

This PR uses NPM to distribute the app, and creates platform-specific NPM packages that contain the native binaries. A user can then e.g. `npx @block/goose`, which will install both the TUI app and the appropriate native goose server.


npm/goose-acp-server-darwin-arm64/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@block/goose-acp-server-darwin-arm64",
+  "version": "0.1.0",
```
jamadeo, line 3: these should maybe match the Cargo.toml version?

ui/text/scripts/build-native-packages.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
```
jamadeo, line 1: we probably don't need this script. We should run the packaging step in the release GHA workflow and pull the appropriate binaries from there

alexhancock (commented):
LGTM 

Test failures look related to API keys which we'll need to resolve out of band of this PR

jamadeo:
closing in favor of #7666

PR #7399, opened 2026-02-20
dev: add cmake to hermitized env

alexhancock:
With the local inference changes in, we need `cmake`. This PR adds it via the existing hermit activation.


michaelneale (approved):
not sure about those failures - we lost some api keys I guess?

alexhancock:
Yeah they're unrelated so I'll merge this

PR #7403, opened 2026-02-20
(feat) Session mananger backend mongodb

dstieglitz:
## Summary

This PR introduces a configurable storage back-end for sessions. The sqlite backed is unchanged and enabled by default. Included is a mongodb backend that stores sessions in an identical document as would be exported by the export of a sqlite session (some additional fields as well).

Much of the code is written by AI, author is not a rust developer and is requesting a review.

### Type of Change
- [X ] Feature
- [ ] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [X ] Documentation
- [X ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance
- [X ] This PR was created or reviewed with AI assistance
Developer is NOT a rust developer and used Claude to generate code and tests. Architecture was human-guided, and code was human-reviewed. Tests were written for automated tested and manual testing was performed on author's cluster (by a human).

### Testing
This PR was tested using the project's test harness (feature tests added), as well as by deploying to a test environment we're running with Kubernetes

### DIFF
--- a/crates/goose/src/session/mongodb_storage.rs
+++ b/crates/goose/src/session/mongodb_storage.rs
@@ -358,7 +358,11 @@ impl SessionStorageBackend for MongoDbSessionStorage {
         name: String,
         session_type: SessionType,
     ) -> Result<Session> {
-        let id = self.generate_session_id().await?;
+        // Allow overriding session ID via environment variable (for linking with external systems)
+        let id = match std::env::var("GOOSE_SESSION_ID") {
+            Ok(env_id) if !env_id.is_empty() => env_id,
+            _ => self.generate_session_id().await?,
+        };
         let now = mongodb::bson::DateTime::now();
 
         let doc = doc! {


DOsinga:
can I ask what the use case is? I have no mongodb set up so I can't test this. it adds 2500 lines - who will maintain that?

dstieglitz:
We're running goose in a cluster with parallel agents and it would be
useful for them to share sessions among each other for knowledge sharing
(and for retrieval via API). Additionally, this enables centralized
reporting for token usage through prometheus/grafana with an appropriate
exporter.

We tested it using a plain Docker-based MongoDB image that you can run in a
single command (and destroy just as easily).

alexhancock:
@dstieglitz I love the idea and see the motivation for sure!

Where I have concerns as a maintainer of goose is the volume of custom code required just for mongo. What would we do if someone else came and wanted to add an exporter/manager for N other DBs?

Thinking about it more generally what if we made it so you could configure a shared SQLite db for all your geese (instead of just the in process file based one)? Something accessible on a mounted filesystem or a server running `sqld`...

If we did this, would it address the need? It looks like there are then a variety of off-the-shelf tools people could then leverage to export data from a shared SQLite db to other data storage solutions (ex https://litestream.io) as you need with no maintenance required in goose.

Let me know your thoughts

dstieglitz:
@alexhancock makes sense. A shared SQLIte would definitely enable this kind of feature and would be a good balance between maintainability and functionality.

If you keep the plugin interface and add some minimal code to allow plugins or extensions, I can move the MongoDB code to an external project and maintain it, and others can extend as well and create community-maintained extensions.

DOsinga:
going to close this for now based on the discussion - making the sqlite session db configurable is the way to go. thanks for thinking about this @dstieglitz

PR #7406, opened 2026-02-21
add flag to hide select voice providers

zanesq:
## Summary
In https://github.com/block/goose/pull/5095 we removed a flag we use internally to hide eleven labs voice dictation. Added back a new flag `DICTATION_ALL_PROVIDERS_ENABLED` to encompass multiple voice providers we want to hide.


dianed-square (approved):

zanesq:
@DOsinga good call was thinking that over the weekend, will follow up with that change

PR #7407, opened 2026-02-21
change voice dictation dropdown to radix dropdown

zanesq:
## Summary
Noticed I wasn't able to click out of the voice dictation dropdown. Changing it to use our radix dropdown component so we get these for free:

- Click outside to close
- Escape key to close
- Keyboard navigation
- ARIA roles and accessibility

Slight difference with selected style but looks ok to me.
Screenshot 2026-02-20 at 5 08 03 PM


Abhijay007 (approved):
Hi @zanesq looks good to me, just one thing that Esc key in Windows does not close it but the whole setting section and bringing it back to the chat interface

zanesq:
Good catch @Abhijay007! I went ahead and fixed it in the related PR https://github.com/block/goose/pull/7406. I left the esc key to go back to home from settings for now however since it was pre-existing, will come back to that.

PR #7414, opened 2026-02-22
fix: prevent tool-pair summarization from being silently undone

### LINKED ISSUE #7413: Tool-pair summarization creates hidden messages that are silently deleted on next client round-trip
## Summary

When tool-pair summarization fires, it creates server-side hidden messages (`userVisible: false, agentVisible: true`) and mutates prior message metadata without emitting a `HistoryReplaced` event. The UI never learns about these changes. On the next user turn, the UI sends `conversation_so_far` containing its stale copy of the conversation, and the server calls `replace_conversation` with it — permanently deleting the hidden summaries and potentially reverting metadata changes the server just made.

This is in contrast to full compaction (`compact_messages`), which correctly emits `HistoryReplaced` and keeps the UI in sync.

## Affected Code

- `crates/goose/src/agents/agent.rs` — tool-pair summarization block (~L1580–1608): calls `update_message_metadata` and `add_message` but never emits `AgentEvent::HistoryReplaced`
- `crates/goose-server/src/routes/reply.rs` (~L288–305): unconditionally calls `replace_conversation` with client-supplied `conversation_so_far`
- `crates/goose/src/agents/agent.rs` (~L951, ~L1034, ~L1493): the three places where `HistoryReplaced` IS emitted — tool-pair summarization is the missing fourth

## Reproduction

**Setup:** Set `GOOSE_TOOL_CALL_CUTOFF: 3` in `~/.config/goose/config.yaml` to make summarization trigger quickly.

1. Start a fresh session
2. Send a message that forces chained tool calls:
   > "List the files in /tmp, read the 3 most recently modified ones, and tell me what you find"
3. After the response, query the session DB:
   ```sql
   SELECT message_id, role, metadata_json
   FROM messages
   WHERE session_id = (SELECT id FROM sessions ORDER BY updated_at DESC LIMIT 1)
   ORDER BY created_timestamp;
   ```
   Observe: a new `userVisible: false, agentVisible: true` message (the hidden summary) and one or two messages now marked `agentVisible: false`.

4. Send any follow-up message ("thanks")
5. Query the DB again

**Expected:** hidden summary persists; `agentVisible: false` metadata preserved.

**Actual:** hidden summary is deleted; server had to re-run summarization from scratch. Other messages also dropped.

## Evidence

Captured from a live test session:

**After turn 1:**
```
msg_6df69722  assistant  agentVisible:false   ← summarized, hidden from AI
msg_7e3e673e  user       agentVisible:false   ← summarized, hidden from AI
msg_803f5a4f  user       userVisible:false, agentVisible:true  ← hidden summary
```

**After sending "thanks":**
```
msg_011qCnLc  assistant  agentVisible:false   ← different message ID, msg_6df69722 gone
msg_7e3e673e  user       agentVisible:false   ← preserved
msg_cc2f0950  user       userVisible:false, agentVisible:true  ← brand new summary
```

`msg_803f5a4f` is completely gone. The server recreated the work it just did.

## Impact

- Tool-pair summarization silently undoes itself every turn
- Agent context is not compressed as intended — tool pairs the system tried to retire may re-enter the context window
- In long sessions where stale tool results (e.g. failed extension checks) were being compressed away, those results can persist in context longer than expected, contributing to stale reasoning loops
- Severity: medium. No data loss visible to the user; the system partially self-heals by re-running summarization. But the intended token economy and context hygiene of tool-pair summarization is effectively broken.

## Root Cause

`maybe_summarize_tool_pair` in `agent.rs` calls `update_message_metadata` and `session_manager.add_message` but does not emit `AgentEvent::HistoryReplaced`. The UI's local state therefore never reflects the server's metadata mutations. On the next `/reply` request, `conversation_so_far` carries the stale pre-mutation state, and `reply.rs` calls `replace_conversation` unconditionally, overwriting the server's changes.

## Proposed Fix

**Immediate (low risk):** After applying tool-pair summarization changes in `agent.rs`, emit `AgentEvent::HistoryReplaced(conversation.clone())`. This is identical to how full compaction notifies the UI. The UI will refresh its local state and subsequent `conversation_so_far` payloads will be accurate.

**Follow-up (tracked separately):** The `conversation_so_far` ownership model is inherently fragile — the client should not be the source of truth for server-managed state. Consider a server-authoritative reply contract where the client can only append new messages, not replace existing ones.

## References

- Full compaction emits correctly: `agent.rs` ~L1034
- Recovery compaction emits correctly: `agent.rs` ~L1493
- Slash command compaction emits correctly: `agent.rs` ~L951
- Tool-pair summarization: `agent.rs` ~L1580–1608 (missing emission)
- `AgentEvent::HistoryReplaced` handling in UI: `useChatStream.ts` ~L294

bzqzheng:
## What
Fix a conversation state desync where tool-pair summarization creates server-side hidden messages but the UI never learns about these changes. On the next user turn, the UI sends its stale conversation state, and the server's `replace_conversation` (DELETE ALL + INSERT) permanently deletes the hidden summaries — making the summarization silently undo itself every time.

## Changes
- **Server:** Emit `AgentEvent::HistoryReplaced` at the end of agent loop iterations where conversation was mutated (tool-pair summarization added hidden messages)
- **UI:** Remove `conversation_so_far` from normal `handleSubmit` turns (server already has authoritative state in DB)
- **Tests:** Add two integration tests in `crates/goose/tests/tool_pair_summarization.rs`:
  - `test_history_replaced_emitted_after_tool_pair_summarization` — proves the fix works: HistoryReplaced is emitted with hidden summary and original tool pair marked agent-invisible
  - `test_stale_conversation_overwrites_hidden_summary` — proves why the fix matters: demonstrates that without HistoryReplaced, stale UI state would silently wipe server-side summaries via replace_conversation

## Why both changes
- Server-side emit keeps UI state accurate after tool-pair summarization
- UI stops sending stale conversation_so_far so server's work can't be overwritten
- Guard `!did_recovery_compact_this_iteration` prevents duplicate HistoryReplaced when recovery compaction already emitted one

## Verification
- Both new integration tests pass: `cargo test --package goose --test tool_pair_summarization`
- Tests use `#[serial]` to avoid env var races with `GOOSE_TOOL_CALL_CUTOFF`
- No mock LLM calls needed — tests construct Agent with MockSummarizationProvider

## References
- Fixes #7413

### DIFF
--- a/crates/goose/tests/tool_pair_summarization.rs
+++ b/crates/goose/tests/tool_pair_summarization.rs
@@ -1,0 +1,379 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use futures::StreamExt;
+use goose::agents::{Agent, AgentEvent, SessionConfig};
+use goose::conversation::message::{Message, MessageContent};
+use goose::conversation::Conversation;
+use goose::model::ModelConfig;
+use goose::providers::base::{
+    stream_from_single_message, MessageStream, Provider, ProviderDef, ProviderMetadata,
+    ProviderUsage, Usage,
+};
+use goose::providers::errors::ProviderError;
+use goose::session::session_manager::SessionType;
+use goose::session::Session;
+use rmcp::model::{AnnotateAble, CallToolRequestParams, RawContent, Tool};
+use serial_test::serial;
+use std::sync::Arc;
+use tempfile::TempDir;
+
+// ---------------------------------------------------------------------------
+// Mock provider that recognises summarization calls via the system prompt
+// ---------------------------------------------------------------------------
+
+struct MockSummarizationProvider;
+
+impl MockSummarizationProvider {
+    fn new() -> Self {
+        Self
+    }
+}
+
+#[async_trait]
+impl Provider for MockSummarizationProvider {
+    async fn stream(
+        &self,
+        _model_config: &ModelConfig,
+        _session_id: &str,
+        system_prompt: &str,
+        _messages: &[Message],
+        _tools: &[Tool],
+    ) -> Result<MessageStream, ProviderError> {
+        // complete_fast → complete → stream; the summarization path passes the
+        // indoc system prompt containing "summarize a tool call".
+        let is_summarization = system_prompt
+            .to_lowercase()
+            .contains("summarize a tool call");
+
+        let message = if is_summarization {
+            Message::assistant().with_text("A call to shell was made to list files")
+        } else {
+            // Regular reply — no tool requests so the agent loop exits.
+            Message::assistant().with_text("Done.")
+        };
+
+        let usage = ProviderUsage::new(
+            "mock-model".to_string(),
+            Usage::new(Some(100), Some(50), Some(150)),
+        );
+
+        Ok(stream_from_single_message(message, usage))
+    }
+
+    fn get_model_config(&self) -> ModelConfig {
+        ModelConfig::new("mock-model").unwrap()
+    }
+
+    fn get_name(&self) -> &str {
+        "mock-summarization"
+    }
+}
+
+impl ProviderDef for MockSummarizationProvider {
+    type Provider = Self;
+
+    fn metadata() -> ProviderMetadata {
+        ProviderMetadata {
+            name: "mock".to_string(),
+            display_name: "Mock Summarization Provider".to_string(),
+            description: "Mock provider for tool-pair summarization testing".to_string(),
+            default_model: "mock-model".to_string(),
+            known_models: vec![],
+            model_doc_link: "".to_string(),
+            config_keys: vec![],
+            allows_unlisted_models: false,
+        }
+    }
+
+    fn from_env(
+        _model: ModelConfig,
+        _extensions: Vec<goose::config::ExtensionConfig>,
+    ) -> futures::future::BoxFuture<'static, anyhow::Result<Self>> {
+        Box::pin(async { Ok(Self::new()) })
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+/// Build a tool-request / tool-response pair linked by `call_id`.
+/// Both messages carry `.with_id()` — required by the `msg.id.is_some()`
+/// guard at agent.rs:1586.
+fn create_tool_pair(
+    call_id: &str,
+    response_id: &str,
+    tool_name: &str,
+    response_text: &str,
+) -> Vec<Message> {
+    vec![
+        Message::assistant()
+            .with_tool_request(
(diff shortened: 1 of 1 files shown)


DOsinga:
> Fixes a conversation state desync where tool-pair summarization creates server-side hidden messages (userVisible: false, agentVisible: true) and mutates prior message metadata, but the UI never learns about these changes. On the next user turn, the UI sends its stale conversation_so_far, and the server blindly calls replace_conversation — permanently deleting the hidden summaries and reverting metadata mutations. 

I don't think that is true. that is how it used to work a long time agao, but now reply(...) just takes the user message and fetches the conversation from the session manager - whatever the UI has doesn't enter into it.

If we have this bug though, we should fix it! Can we come up with a test that reproduces this? Since this compaction stuff doesn't actually require us to call the LLM that sounds doable, does it not?

bzqzheng:
Thanks for the review! The `None` path does fetch from the session manager — but currently on `main`, the UI [sends `conversation_so_far` on every turn](https://github.com/block/goose/blob/35866e4ad7/ui/desktop/src/hooks/useChatStream.ts#L593) with existing messages, and the server [calls `replace_conversation()` when that field is present](https://github.com/block/goose/blob/35866e4ad7/crates/goose-server/src/routes/reply.rs#L288-L305), which does a full DELETE + INSERT before processing the new message. So the UI inadvertently overwrites the server's state each turn.

I reproduced this deterministically — full steps in [#7413](https://github.com/block/goose/issues/7413). With `GOOSE_TOOL_CALL_CUTOFF: 3`: after turn 1, the server creates a hidden summary (`userVisible:false, agentVisible:true`). On the next turn, that summary is gone — the UI's `conversation_so_far` never included it, and `replace_conversation` wiped it.

**Context on why I looked into this:** I kept seeing an outdated "Jira extension failed to load" error resurface across turns during personal usage, even after the extension was working. Digging into the code, I found tool-pair summarization was compressing that old failure into a hidden summary that persisted as stale agent context. The desync bug amplifies it (summaries get wiped and recreated each turn from truncated history), but the stale-context problem itself is a separate concern worth its own issue.

**On tests:** Fully agree — very testable without LLM calls. Happy to add a test that mocks `complete_fast()`, creates tool pairs past the cutoff, runs summarization, then asserts hidden summaries survive a round-trip.

DOsinga:
thanks again for looking into this. I think this is just an unfortunate regression and we should just not call the server with conversation_so_far - I have a small PR to make it so

DOsinga:
here's my take: https://github.com/block/goose/pull/7438

bzqzheng:
> here's my take: #7438

sorry for the delayed review. your PR looks good to me. closing this PR.

Issue #7415, opened 2026-02-22
Tool-pair summarization is too aggressive for large context models


angiejones:
## Problem

Tool-pair summarization (`maybe_summarize_tool_pair`) fires after 10 tool calls (`GOOSE_TOOL_CALL_CUTOFF` default) regardless of context window size or token usage. On a 1M context model at 7% usage, it still summarizes one tool pair every turn — destroying detailed tool outputs the user has plenty of room for.

Three issues:

### 1. Fixed cutoff ignores context window size

The default of 10 was reasonable for 128K models but is far too aggressive for 200K-1M models. A session doing code editing hits 10 tool calls in minutes.

### 2. One-per-turn cadence

Once the cutoff is exceeded, it summarizes the oldest tool pair on **every subsequent turn** indefinitely. A 400-message session triggers hundreds of summarization calls.

### 3. Summaries inject as `Role::User` with no framing

The summary is set to `Role::User` with `MessageMetadata::agent_only()`. The agent interprets it as something the user said. When a summary describes a past error (e.g., "A shell command was executed to check if gh was installed, but it was not found"), the agent responds to it as a current problem — even if it was resolved 200 messages ago. Users have to repeatedly tell goose to ignore its own summaries.

## Observed impact

In multiple 200-400 message sessions on Claude Opus 4.6 (1M context):
- Compaction summaries about resolved errors kept resurfacing as "ghost" messages
- Agent repeatedly responded to stale summaries instead of the user's actual message  
- Users had to say "that's the compaction message, ignore it" multiple times per session

## Suggested fixes

1. **Scale cutoff with context window** — e.g., `context_limit / 10_000` clamped to 10-100
2. **Batch summarization** — when cutoff is exceeded, summarize N oldest pairs at once, then don't fire again until the cutoff is exceeded again (instead of one-per-turn forever)
3. **Fix summary framing** — either don't use `Role::User`, or wrap summaries with clear framing so the agent knows it's a historical record, not a current user message

## Workaround

```yaml
# In config or environment
GOOSE_TOOL_CALL_CUTOFF: 200
```

## Related

- #7413 — tool-pair summarization sync bug (summaries silently deleted on next client round-trip)

## References

- `crates/goose/src/context_mgmt/mod.rs` — `tool_id_to_summarize()`, `summarize_tool_call()`
- `crates/goose/src/agents/agent.rs` — `tool_call_cut_off`, `maybe_summarize_tool_pair` usage
- Introduced in #6119 (2026-01-28)

[...]

Issue #6930, opened 2026-02-03
not a bug, however I simply ran out of tokens and was not alerted that my tokens were running low although I set guardrails to alert me


normancomics:
not a bug I simply ran out of tokens and was not alerted that my tokens were running low although I set guardrails to alert me

katzdave:
Closing since nothing actionable. You can view token usage for a session by hovering over the (what starts as green) dot below the chat.

PR #7418, opened 2026-02-22
feat: expose context window utilization to agent via MOIM

tlongwell-block:
## Summary

Add token/context usage info to the per-turn `<info-msg>` (MOIM — Minus One Info Message) injection so the agent can see how much of its context window is consumed and make informed decisions (e.g., summarize early, hand off, or keep going).

The agent now sees a line like:
```
Context: ~85k/128k tokens used (66%)
```

## Implementation

~22 lines in `collect_moim()` in `extension_manager.rs`, which already has access to both the provider (`context_limit`) and `session_manager` (`total_tokens`). No new files, no new dependencies, no new abstractions.

### Fail-open guarantees (4 layers)
1. `if let Ok(session)` — session lookup fails → skip silently
2. `if let (Some(total), Some(config))` — missing token data or model config → skip silently
3. `total > 0` — negative or zero tokens (first turn, corrupt state) → skip silently
4. `limit > 0` — zero context limit → skip silently (also prevents division by zero)

## Why this matters

The agent currently has **zero visibility** into its own context usage. It can't tell if it's at 20% or 95% of its context window. This leads to:
- Agents running into compaction unexpectedly
- No ability to proactively summarize or hand off when running low
- Context management decisions happening around the agent instead of with it

Related issues: #1703, #7415, #7413, #6930, #5255, #3485

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -1608,8 +1608,15 @@ impl ExtensionManager {
             working_dir.display()
         );
 
-        if let Ok(session) = self.context.session_manager.get_session(session_id, false).await {
-            if let (Some(total), Some(config)) = (session.total_tokens, session.model_config.as_ref()) {
+        if let Ok(session) = self
+            .context
+            .session_manager
+            .get_session(session_id, false)
+            .await
+        {
+            if let (Some(total), Some(config)) =
+                (session.total_tokens, session.model_config.as_ref())
+            {
                 let limit = config.context_limit();
                 if total > 0 && limit > 0 {
                     let pct = (total as f64 / limit as f64 * 100.0).round() as u32;


crates/goose/src/agents/extension_manager.rs
```diff
@@ -1608,6 +1608,28 @@ impl ExtensionManager {
...
+            if let (Some(total), Some(config)) =
+                (session.total_tokens, session.model_config.as_ref())
+            {
+                let limit = config.context_limit();
+                if total > 0 && limit > 0 {
+                    let pct = (total as f64 / limit as f64 * 100.0).round() as u32;
+                    content.push_str(&format!(
+                        "Context: ~{}k/{}k tokens used ({}%)\n",
+                        total / 1000,
+                        limit / 1000,
```
DOsinga, line 1626: 1000 or 1024? this rounds down, probably fine, but actually rounding would be more accurate

tlongwell-block: Should be 1000, I think, since they're tokens

Issue #6941, opened 2026-02-04
Anthropic prompt caching behavior


GordianDziwis:
**Describe the bug**

The Anthropic prompt caching implementation had two issues that affected cache efficiency:

1. **Cache prefix continuity broken**: MOIM (Message of Instruction Mode) messages were being inserted after the last assistant message, which broke the cache prefix continuity required by Anthropic's caching system.

2. **Cache control logic complexity**: The cache_control logic used nested loops and didn't properly exclude MOIM messages (`<info-msg>`) from caching, potentially caching invalid content types.

---

**To Reproduce**

1. Use goose with Anthropic provider (Claude models)
2. Have a conversation with multiple turns
3. Observe MOIM insertion point in conversation
4. Check that cache_control is applied to MOIM messages when it shouldn't be
5. Notice reduced cache hit rates due to broken cache prefix

---

**Expected behavior**

- MOIM messages should be inserted at the end of the conversation to preserve Anthropic's cache prefix continuity
- Only valid content block types (text, image, tool_use, tool_result) should have cache_control applied
- MOIM messages (`<info-msg>`) should be excluded from caching

---

**Please provide the following information**
- **OS & Arch:** [All platforms]
- **Interface:** [UI / CLI]
- **Version:** [Affected versions prior to commit 64c4bc1]
- **Extensions enabled:** [All]
- **Provider & Model:** [Anthropic – Claude models with prompt caching enabled]

---

**Additional context**

This fix improves:
- Cache hit rates for Anthropic conversations
- Cost efficiency when using prompt caching
- Code maintainability with simplified cache control logic

The issue affected all users of goose with Anthropic provider when prompt caching was enabled.

tlongwell-block:
Hi, @GordianDziwis 👋 

The MOIM should always be **very** recent in the context (that's why it is named the Minus One Info Message, being close to top of context-1).

It should only prevent a tiny part of context from being cached. Are you seeing different behavior?

alexhancock:
Closing due to inactivity/no reply

Issue #5997, opened 2025-12-05
Snowflake provider + Developer Tools: "assistant role in the final position when tools are used" + -32602: tool not found


Taooo-habitus:
```bash
─── text_editor | developer__text_editordeveloper__text_editordeveloper__text_editordeveloper ──────────────────────────

◇  Goose would like to call the above tool, do you allow?
│  Always Allow 
│
-32602: tool not foundRan into this error: Request failed: Request failed with status: 400 Bad Request. Message: invalid request parameters: Requests cannot contain `assistant` role in the final position when `tools` are used..

Please retry if you think this is a transient or recoverable error.
```

Sometimes developer tool calls fail when using Snowflake Cortex :(
And retry works sometimes, but not always.

blackgirlbytes:
too you have ` GOOSE_ENABLE_ROUTER` set to true by any chance?

4t8dd:
@Taooo-habitus have you ever resolve this? I got the same issue now.
@blackgirlbytes what does this env do in the backend?

DOsinga:
can you provide a diagnostics report @4t8dd or @Taooo-habitus ? it would be useful to see which model is doing this.

4t8dd:
@DOsinga diagnosis report does not help you I think. The error message here is enough. 
This is the constraint from models in snowflake. The last message role should not be asssitant when there is a tool calling.
And the official says this is from the protocol of openai api. But I never find this.

jamadeo:
The linked PR looks like it fixes at least one way this could happen. To properly diagnose this though we would need a diagnostics bundle to see the message sequence that caused it. It's true that the error says what's wrong, but it doesn't tell us how we created that sequence. Feel free to re-open if you see this again and can get a diagnostics report.

PR #7424, opened 2026-02-22
fix: handle truncated tool calls that break conversation alternation

tlongwell-block:
## Summary

When a model response hits `max_tokens` mid-tool-call, the streaming handler creates a `tool_request` with a parse error (truncated JSON). The agent loop counts it as a tool request but silently skips it — no tool request message, no tool response added to the conversation. The text preamble from the truncated response becomes a dangling trailing assistant message. The loop continues, and on the next iteration the conversation is sent to the API ending with an assistant message, which crashes on providers that reject prefill (Claude 4.6+, Snowflake, others).

This was latent — older models silently tolerated trailing assistant messages via prefill support. Claude Opus 4.6 (Feb 5) removed prefill as a documented breaking change, exposing it. Multiple users have reported the resulting `"does not support assistant message prefill"` errors since early February.

## The fix

When a tool call can't be parsed, append a plain user text message telling the model the call was likely truncated and to try breaking the task into smaller steps. This maintains conversation alternation without using tool protocol types that serialize inconsistently across providers.

## Type of Change
- [x] Bug fix

## Testing

All existing tests pass. The bug requires the full agent streaming loop to reproduce (provider stream → truncated tool call → silent skip → trailing assistant), so a unit test is not practical without significant test infrastructure investment.

## See also
- #7421 — patches a downstream symptom of the same bug in MOIM injection
- #5997 — same error class reported with Snowflake provider
- #6941 — related Anthropic prompt caching issue

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -1443,6 +1443,16 @@ impl Agent {
                                                                 .lock().await.clone();
                                         yield AgentEvent::Message(final_response.clone());
                                         messages_to_add.push(final_response);
+                                    } else {
+                                        let error_msg = format!(
+                                            "[system: Tool call could not be parsed: {}. The response may have been truncated. Try breaking the task into smaller steps.]",
+                                            request.tool_call.as_ref().unwrap_err(),
+                                        );
+                                        let error_response = Message::user()
+                                            .with_generated_id()
+                                            .with_text(&error_msg);
+                                        yield AgentEvent::Message(error_response.clone());
+                                        messages_to_add.push(error_response);
                                     }
                                 }
 


tlongwell-block:
Testing with my local goose

```
  ━━╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌ 9% 18k/200k
🪿 Now I want you to help me test my fix. YOU are built with this change included in YOUR code. Do you understand that? I want you to aknowledge that, then I want to to attempt to directly write a giant file to ~/scratch. Ideally, you should see that error from the PR and it should tell you to try doing it in multiple tool calls. Don't do that, just let me know what error you get and if the PR fix is working as intended.



Yes, I understand — I'm running on a build of goose that includes PR #7424. So if my response gets truncated mid-tool-call due to hitting `max_tokens`, I should see that new error message telling me the tool call couldn't be parsed and to break the task into smaller steps.

Let me try to write a massive file in a single tool call. I'll generate something absurdly large to try to trigger the truncation.-32602: Could not interpret tool use parameters for id toolu_018gedtVovzfTUzZo4KThmnm: EOF while parsing an object at line 1 column 75[system: Tool call could not be parsed: -32602: Could not interpret tool use parameters for id toolu_018gedtVovzfTUzZo4KThmnm: EOF while parsing an object at line 1 column 75. The response may have been truncated. Try breaking the task into smaller steps.]There it is! The PR fix is working exactly as intended. Here's what happened:

1. **I attempted** to write a massive file in a single tool call
2. **My response hit `max_tokens`** mid-tool-call, truncating the JSON payload
3. **Before the fix**: This would have been silently skipped, leaving a dangling assistant message, and the next API call would crash with `"does not support assistant message prefill"`
4. **With the fix**: I got the new error message:

> `[system: Tool call could not be parsed: -32602: Could not interpret tool use parameters for id toolu_018gedtVovzfTUzZo4KThmnm: EOF while parsing an object at line 1 column 75. The response may have been truncated. Try breaking the task into smaller steps.]`

Key observations:
- **Conversation alternation is maintained** — the error was injected as a user message, so the conversation stays valid (assistant → user)
- **I didn't crash** — the agent loop continued normally and I'm able to respond to you right now
- **The error is actionable** — it tells me the parse failed, why (truncated JSON / EOF), and what to do about it (break into smaller steps)
- **The error code `-32602`** (invalid params) confirms it was indeed truncated JSON that couldn't be deserialized
```

DOsinga:
this is good. we could recover from this though if we would do the pre-fill thing, yeah? or do we actually have that, it is just that using anthropic through databricks kills this?

tlongwell-block:
> this is good. we could recover from this though if we would do the pre-fill thing, yeah? or do we actually have that, it is just that using anthropic through databricks kills this?

The agent should recover automatically after getting the error message and continue working with smaller tools calls without user intervention

DOsinga:
> The agent should recover automatically after getting the error message and continue working with smaller tools calls without user intervention

no, I understand. it is just that if I remember correctly, anthropic has this thing where you can start a request with the the assistant pre-filled and so if you run out of output tokens, you can repost what you had so far as prefill and you can create arbitrarily large replies. I wonder if that is already supported, just not through databricks

PR #7426, opened 2026-02-22
feat: simplify the text editor to be more like pi

michaelneale:
the text editor really fails hard on small models by trying to edit the whole file, this makes it more like the `pi` simple editor approach, which is able to run eval/gym tasks like "remove the feature" whereas the existing editor will not. 

More an exemplar than a hard implementation for now...

### DIFF
--- a/crates/goose-mcp/src/developer/rmcp_developer.rs
+++ b/crates/goose-mcp/src/developer/rmcp_developer.rs
@@ -1835,10 +1835,9 @@ mod tests {
             .as_text()
             .unwrap();
 
-        assert!(
-            assistant_content.text.contains("The file")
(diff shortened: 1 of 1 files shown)


crates/goose-mcp/src/developer/text_editor.rs
```diff
@@ -848,25 +848,34 @@ pub async fn text_editor_replace(
...
         .split(old_str)
         .next()
         .expect("should split on already matched content")
         .matches('\n')
-        .count();
+        .count()
+        + 1; // 1-indexed
+
+    let summary = format!(
+        "Successfully edited {} (replaced {} lines with {} lines at line {})",
```
DOsinga: you could drop the ()

PR #7429, opened 2026-02-23
fix: use BEGIN IMMEDIATE to prevent SQLite deadlocks

rabi:
## Summary
Fixes `Internal error: "Error in agent response stream: error returned from database: (code: 5) database is locked"` seen at times.

SessionManager is Arc-shared across concurrent tokio tasks with a multi-connection pool. DEFERRED transactions deadlock when two connections both hold shared locks and try to upgrade to write. BEGIN IMMEDIATE acquires the write lock upfront, letting contenders wait via busy_timeout instead of deadlocking.

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested locally.

### DIFF
--- a/crates/goose/src/session/session_manager.rs
+++ b/crates/goose/src/session/session_manager.rs
@@ -665,7 +665,7 @@ impl SessionStorage {
     }
 
     async fn import_legacy_session(pool: &Pool<Sqlite>, session: &Session) -> Result<()> {
-        let mut tx = pool.begin().await?;
+        let mut tx = pool.begin_with("BEGIN IMMEDIATE").await?;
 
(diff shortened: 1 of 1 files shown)


rabi:
> is it worth extracting this into a helper to make it clearer to users that this is the pattern to use?

I thought it would be over engineering to create wrapper for an one-liner that's self explanatory. But I can do it if we want.

DOsinga:
nah, I agree, this is fine. merging

PR #7438, opened 2026-02-23
The client is not the source of truth

DOsinga:
## Summary

We did it again. The client is overriding the data from the server, causing subtle bugs.


DOsinga:
fixes this: https://github.com/block/goose/pull/6970

bzqzheng:
hey sorry for the delayed review. this looks good to me, thanks for addressing the issue!

PR #7452, opened 2026-02-23
fixed build error with conflicting C runtime library (STILL NOT WORKING)

lifeizhou-ap:
**Bundle windows failed during complication**

 ### Original root cause:

  - Windows native linking broke after adding llama-cpp-2 (llama-cpp-sys-2).
  - Two overlapping issues appeared:
      1. CRT mismatch between /MT and /MD.
      2. v8 and llama-cpp-sys-2 linked together with duplicate C++ symbols.

  What /MT vs /MD means:

  - /MT = link the static MSVC C/C++ runtime into the binary (libcmt).
  - /MD = use the dynamic MSVC runtime DLLs (msvcrt/msvcprt).
  - Mixing them in one link graph often causes LNK2038, LNK4098, and symbol conflicts.

 ### What we changed:

  - Disabled tokenizers default features (esaxx_fast off, onig only).
  - Aligned Windows CRT settings for Rust/CMake/llama toward dynamic runtime.
  - Upgraded llama-cpp-2 / llama-cpp-sys-2 from 0.1.133 to 0.1.136.

  **Current status:**

  - Tokenizers/CRT mismatch symptoms improved.
  - Windows still fails on v8 + llama-cpp-sys-2 collision:
      - duplicate std::exception_ptr symbols
      - LNK4098 + LNK1169


katzdave (approved):

lifeizhou-ap:
Hi @jh-block,

I tried to fix this bundle windows, but it still does not work.  https://github.com/block/goose/actions/runs/22337431964/job/64633151030

I have raised an issue in https://github.com/utilityai/llama-cpp-rs/issues/940

In the meantime I am wondering whether we shall disable local inference provider in Windows at the moment?

jurrejelle:
I ran into this issue when trying to build locally on windows yesterday, I thought it was a bug on my end so this fix is very much appreciated! <3

lifeizhou-ap:
Closed as it was fixed in https://github.com/block/goose/pull/7511

PR #7454, opened 2026-02-23
Add lmstudio declarative provider and fix up a few rough edges

zanesq:
## Summary

Added LM Studio as a built-in provider for Goose so its easier to setup. Also fixed several issues 
that made the local model experience poor.

PROBLEMS:
- No built-in LM Studio support — users had to manually create a custom provider or use the openai compatible
- A recently merged PR (#7304) introduced malformed JSON configs for Kimi/Moonshot 
  that silently broke ALL declarative provider loading (Groq, DeepSeek, Cerebras, Mistral, Inception, OVHcloud all disappeared from the UI)
- When local models returned errors in streaming responses (like context length 
  exceeded), users saw cryptic "Failed to parse streaming chunk: missing field 
  choices" instead of the actual error message from the server
- Local models that use inline </think> tags for chain-of-thought reasoning had 
  raw tags and thinking content leaked into both chat messages and session names
- The "Show reasoning" UI section had minimal styling compared to "Show thinking"
- New chat sessions that failed to load got stuck — clicking "Start New Chat" 
  kept reusing the broken session instead of creating a fresh one

FIXES:
- Added LM Studio declarative provider config (OpenAI-compatible, localhost:1234, 
  no auth, dynamic model fetching)
- Fixed kimi.json and moonshot.json to use the correct schema, restoring all 
  declarative providers
- Made declarative provider loading resilient — one bad JSON file no longer breaks 
  all providers
- Improved OpenAI streaming parser to extract actual server error messages from 
  non-standard error responses embedded in streams
- Added </think> tag detection in the streaming parser that buffers thinking 
  content and emits it as proper Reasoning content (shown in a collapsible UI 
  section) while only showing the actual response as text
- Fixed session naming to strip inline thinking content so sessions get meaningful 
  names instead of "1. **Analyze the Request:**..."
- Styled the "Show reasoning" collapsible section with rounded background and 
  border matching the existing "Show thinking" section
- Fixed "Start New Chat" to skip errored sessions instead of reusing them, by 
  tracking session error state via existing status update events

Visually differentiate thinking message in chat and fix it so inline thinking is pulled out
before
image
after
Screenshot 2026-02-23 at 3 21 04 PM

Thinking message was getting in the session autogenerated title like this before the fix
Screenshot 2026-02-23 at 2 23 28 PM

### DIFF
--- a/crates/goose/src/providers/base.rs
+++ b/crates/goose/src/providers/base.rs
@@ -27,17 +27,18 @@ fn strip_xml_tags(text: &str) -> String {
     static BLOCK_RE: LazyLock<Regex> = LazyLock::new(|| {
         Regex::new(r"(?s)<([a-zA-Z][a-zA-Z0-9_]*)[^>]*>.*?</[a-zA-Z][a-zA-Z0-9_]*>").unwrap()
     });
+    // Orphan closing tags with everything before them (e.g. thinking content
+    // from local models that output `reasoning...</think>answer` with no opening tag)
+    static ORPHAN_CLOSE_RE: LazyLock<Regex> =
+        LazyLock::new(|| Regex::new(r"(?s)^.*</[a-zA-Z][a-zA-Z0-9_]*>").unwrap());
     static TAG_RE: LazyLock<Regex> =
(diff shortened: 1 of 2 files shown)


zanesq:
Closing for more investigation / testing, will put up a separate PR for just the declarative provider changes

PR #7457, opened 2026-02-24
fix(summon): restore skill supporting files and directory path in load output

angiejones:
## Summary

When the unified summon extension replaced `skills_extension` in v1.25.0 (#6964), skill loading stopped communicating the skill directory path and supporting files (scripts, templates, etc.) to the agent. This caused the agent to load the skill content successfully but then have no idea where bundled scripts live, leading it to search the entire filesystem.

Restores the old behavior by collecting supporting files during skill discovery and including the skill directory + file listing in the load output.

### Type of Change
- [x] Bug fix
- [ ] Feature
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
- All 9 existing summon unit tests pass
- New `test_skill_supporting_files_discovered` test verifies skills with bundled files get their directory path and supporting files populated correctly

### Related Issues
Regression introduced in #6964

### DIFF
--- a/crates/goose/src/agents/platform_extensions/summon.rs
+++ b/crates/goose/src/agents/platform_extensions/summon.rs
@@ -45,6 +45,7 @@ pub struct Source {
     pub description: String,
     pub path: PathBuf,
     pub content: String,
+    pub supporting_files: Vec<PathBuf>,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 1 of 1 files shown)


angiejones:
> good catch @angiejones.
> 
> I wonder what other agents do - I guess it is convenient to list them in one go like this (could also just put it to root of dir and let whatever FS tool find it - pi does that, but this way is probably snappier, as long as we don't blow up the context with content of files but it should be reading them progressively on demand still).

yeah, followed the spec to do the 3 tier progressive disclosure. load output would be something like

```
## Supporting Files

Skill directory: /path/to/.goose/skills/my-skill

The following supporting files are available:
- myscript.sh
- templates/report.txt

Use the file tools to read these files or run scripts as directed.
```

Issue #7432, opened 2026-02-23
Bug: Corrupted permission.yaml silently resets all permissions — denied tools become allowed


marlonbarreto-git:
## Bug Description

When `permission.yaml` contains malformed YAML (from a partial write, manual editing error, or disk corruption), `PermissionManager::new` silently replaces the entire permission map with an empty `HashMap`. No warning is logged. Tools that were explicitly set to `never_allow` become implicitly allowed.

## Affected Code

**File:** `crates/goose/src/config/permission.rs`, line 48

```rust
pub fn new(config_dir: PathBuf) -> Self {
    let permission_path = config_dir.join(PERMISSION_FILE);
    let permission_map = if permission_path.exists() {
        let file_contents =
            fs::read_to_string(&permission_path).expect("Failed to read permission.yaml");
        serde_yaml::from_str(&file_contents).unwrap_or_else(|_| HashMap::new())
        //                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        //  Parse error silently discarded — returns empty map, no log
    } else {
        fs::create_dir_all(&config_dir).expect("Failed to create config directory");
        HashMap::new()
    };
    // ...
}
```

## Security Impact

1. User configures `permission.yaml` with `never_allow` entries for sensitive tools (e.g., file system access, shell execution)
2. The YAML file gets corrupted (partial write during crash, manual edit with syntax error, disk error)
3. On next process start, `serde_yaml::from_str` returns `Err(...)`
4. `unwrap_or_else(|_| HashMap::new())` silently swallows the error
5. `permission_map` is now `{}` — all `never_allow` entries are gone
6. Previously blocked tools are now permitted with **zero user notification**

This is a **fail-open** behavior where security configuration corruption leads to elevated permissions.

## Steps to Reproduce

1. Configure `permission.yaml` with restricted permissions:
   ```yaml
   shell_execute:
     permission: never_allow
   file_write:
     permission: never_allow
   ```
2. Corrupt the file (e.g., truncate it, add invalid YAML characters)
3. Restart goose
4. Observe: no error/warning in logs, all tools are now permitted

## Expected Behavior

At minimum, a `tracing::warn!` or `tracing::error!` should be emitted when the parse fails, alerting the user that their permission configuration is being reset. Ideally, the process should refuse to start with corrupted security configuration (fail-closed).

## Proposed Solution

**Option A (recommended — fail-closed):**
```rust
[...]

PR #7458, opened 2026-02-24
fix: panic on corrupted permission.yaml instead of silently allowing all (#7432)

michaelneale:
Tentative fix for #7432.

When permission.yaml is corrupted/malformed, the YAML parse error was silently swallowed and an empty HashMap returned — wiping all never_allow entries. Denied tools became silently re-allowed.

Fix: panic with a clear error message pointing to the corrupted file. Consistent with existing pattern (file read already uses .expect()). Includes test.

### DIFF
--- a/crates/goose/src/config/permission.rs
+++ b/crates/goose/src/config/permission.rs
@@ -4,6 +4,7 @@ use std::collections::HashMap;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, LazyLock, RwLock};
+use tracing;
 use utoipa::ToSchema;
 
 const PERMISSION_FILE: &str = "permission.yaml";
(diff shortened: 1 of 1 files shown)


michaelneale:
hey @marlonbarreto-git is this the kind of thing that would be better - not sure if panic, but will at least stop it. coudl also revert to most strict but then that could be annoying too if it wasn't that before? reset things? offer to help> ideas?

marlonbarreto-git:
Hey @michaelneale — thanks for picking this up! The fix is a clear improvement over the silent reset. Some thoughts on the approach:

**Panicking is the right call here** — this is a security-relevant config file. A corrupted `permission.yaml` silently becoming an empty HashMap means `never_allow` rules disappear, which is the worst-case failure mode for a permissions system. Better to fail loud than to silently grant access.

The existing code already uses `.expect()` for the `read_to_string` call (line 47), so panicking on corrupt content is consistent with how the file-read failure is handled. If you can't read it → panic. If you can read it but it's garbage → should also panic.

One small thought: if you wanted to be slightly more graceful in the future, the alternative would be to fall back to a **deny-all** default (`PermissionLevel::NeverAllow` for everything) instead of panicking — that way a corrupted file still fails safe. But that's a bigger change and the panic is fine for now since it matches the existing pattern.

The test looks clean — `#[should_panic]` is the right way to assert this. 👍

michaelneale:
@marlonbarreto-git yeah I think the silent failure of it muddling along could be frustrating, so better to blow up like this. I don't know how a file like that gets corrupted but not great!

Issue #7425, opened 2026-02-22
400: The conversation must end with a user message


angiejones:
MOIM injection kills sessions when the conversation ends with an assistant message (e.g. text-only response during retry). The conversation gets sent to the API ending with an assistant message, and models that don't support prefill reject it with a 400.

`inject_moim` always inserts before the last assistant message:

```
[0] User:      "clean up comments in check.sh"
[1] Assistant:  "Let me first look at the current state of check.sh."
[2] Assistant:  [tool_request: text_editor view check.sh]
...
[9] User:       [tool_response: <file contents>]
[10] Assistant: "Now I'll clean this up..."   ← text-only, no tool call
```

code inserts MOIM at index 10 (before the last assistant):
```rust
let mut messages = conversation.messages().clone();
let idx = messages
    .iter()
    .rposition(|m| m.role == Role::Assistant)  // finds message [10]
    .unwrap_or(0);
messages.insert(idx, Message::user().with_text(moim));  // inserts BEFORE [10]
```

This creates:
```
[9]  User:      [tool_response]
[10] User:      <MOIM>              ← inserted here
[11] Assistant:  "Now I'll clean this up..."  ← trailing
```

Results in a 400 error from Anthropic: 
> "This model does not support assistant message prefill. The conversation must end with a user message."

tlongwell-block:
Should be fixed in https://github.com/block/goose/pull/7424

DOsinga:
closing because fixed

PR #7459, opened 2026-02-24
fix: MOIM injection causing trailing assistant message (#7425)

michaelneale:
Tentative fix for #7425.

When a conversation ends with a text-only assistant message (no tool call), MOIM was inserted before it, leaving the assistant message trailing. Anthropic rejects this with 400 'conversation must end with user message'.

Fix: insert MOIM after the last assistant message when it's the final message. Includes regression test.

Note: supersedes the approach on `fix/moim-trailing-assistant` with a cleaner match-based implementation and a more thorough regression test that also verifies the trailing assistant message is preserved (not dropped).

### DIFF
--- a/crates/goose/src/agents/moim.rs
+++ b/crates/goose/src/agents/moim.rs
@@ -24,10 +24,18 @@ pub async fn inject_moim(
         .await
     {
         let mut messages = conversation.messages().clone();
-        let idx = messages
-            .iter()
-            .rposition(|m| m.role == Role::Assistant)
-            .unwrap_or(0);
+        let last_assistant_idx = messages.iter().rposition(|m| m.role == Role::Assistant);
+
+        // If the conversation ends with an assistant message (text-only, no tool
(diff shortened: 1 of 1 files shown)


michaelneale:
hey @tlongwell-block does this one make sense?

michaelneale:
I think this may make sense @angiejones

tlongwell-block:
Fixed in https://github.com/block/goose/pull/7424

tlongwell-block:
Fat-fingered that close, sorry! Yeah, @michaelneale we looked at this in https://github.com/block/goose/pull/7421

It is not caused by the MOIM. Just truncated tool call requests

Also, it is important that the MOIM *not* be the latest message in context or it distracts goose too much

michaelneale:
@tlongwell-block well given that, should we close this one anyway?

tlongwell-block:
Yes, I think so. It just felt rude closing your PR 🙃

PR #7468, opened 2026-02-24
fix: too much tool summarization

michaelneale:
every 10 calls is too much, this fixes: https://github.com/block/goose/issues/7415

### DIFF
--- a/crates/goose/src/context_mgmt/mod.rs
+++ b/crates/goose/src/context_mgmt/mod.rs
@@ -417,13 +417,13 @@ fn format_message_for_compacting(msg: &Message) -> String {
 /// Compute a smart tool call cutoff based on the effective context budget.
 /// When a compaction threshold is set, the effective budget is context_limit * threshold,
 /// since compaction will fire at that point anyway.
-/// Returns effective_limit / 10_000, clamped to [10, 100].
+/// Returns effective_limit / 10_000, clamped to [10, 500].
 pub fn compute_tool_call_cutoff(context_limit: usize, compaction_threshold: Option<f64>) -> usize {
     let effective_limit = match compaction_threshold {
         Some(t) if t > 0.0 && t < 1.0 => (context_limit as f64 * t) as usize,
         _ => context_limit,
     };
-    (effective_limit / 10_000).clamp(10, 100)
+    (effective_limit / 10_000).clamp(10, 500)
 }
 
 /// Find tool call IDs to summarize. Returns the oldest unsummarized tool call IDs
@@ -525,8 +525,6 @@ pub async fn summarize_tool_call(
     Ok(response.with_generated_id())
 }
 
-const TOOL_PAIR_BATCH_SIZE: usize = 3;
-
 pub fn maybe_summarize_tool_pairs(
     provider: Arc<dyn Provider>,
     session_id: String,
@@ -535,8 +533,10 @@ pub fn maybe_summarize_tool_pairs(
     protect_last_n: usize,
 ) -> JoinHandle<Vec<(Message, String)>> {
     tokio::spawn(async move {
-        let tool_ids =
-            tool_ids_to_summarize(&conversation, cutoff, TOOL_PAIR_BATCH_SIZE, protect_last_n);
+        // Batch size is 20% of cutoff (min 3) so summarization fires in discrete
+        // bursts with room to breathe before the next batch triggers.
+        let batch_size = (cutoff / 5).max(3);
+        let tool_ids = tool_ids_to_summarize(&conversation, cutoff, batch_size, protect_last_n);
         let mut results = Vec::new();
         for tool_id in tool_ids {
             match summarize_tool_call(provider.as_ref(), &session_id, &conversation, &tool_id).await
@@ -858,8 +858,8 @@ mod tests {
         // Clamp at minimum
         assert_eq!(compute_tool_call_cutoff(50_000, None), 10);
         assert_eq!(compute_tool_call_cutoff(10_000, None), 10);
-        // Clamp at maximum
-        assert_eq!(compute_tool_call_cutoff(2_000_000, None), 100);
+        // Clamp at maximum (500)
+        assert_eq!(compute_tool_call_cutoff(10_000_000, None), 500);
         // With compaction threshold — uses effective budget
         assert_eq!(compute_tool_call_cutoff(200_000, Some(0.3)), 10); // 60K effective
         assert_eq!(compute_tool_call_cutoff(1_000_000, Some(0.5)), 50); // 500K effective


crates/goose/src/context_mgmt/mod.rs
```diff
@@ -502,25 +512,27 @@ pub async fn summarize_tool_call(
...
     session_id: String,
     conversation: Conversation,
     cutoff: usize,
-) -> JoinHandle<Option<(Message, String)>> {
+) -> JoinHandle<Vec<(Message, String)>> {
     tokio::spawn(async move {
-        if let Some(tool_id) = tool_id_to_summarize(&conversation, cutoff) {
+        let tool_ids = tool_ids_to_summarize(&conversation, cutoff, TOOL_PAIR_BATCH_SIZE);
+        let mut results = Vec::new();
+        for tool_id in tool_ids {
```
jh-block, line 541: Could we do these in parallel (futures::join_all / tokio JoinSet)? They're independent afaict, so it will be a big speedup with remote providers and local inference should serialise them itself

crates/goose/src/context_mgmt/mod.rs
```diff
@@ -485,10 +496,9 @@ pub async fn summarize_tool_call(
                 structured output. So the tool might ask to look up something on github and the
                 reply might be a json document. So you could reply with something like:
 
-                "A call to github was made to get the project status"
-
-                if that is what it was.
+                "[Historical tool call summary] A call to github was made to get the project status"
 
+                if that is what it was. Always start your reply with "[Historical tool call summary]".
```
jh-block, line 514: Could it maybe make sense to apply the prefix ourselves rather than relying on the LLM to follow the instruction?

crates/goose/src/agents/agent.rs
```diff
@@ -1147,7 +1158,7 @@ impl Agent {
                     break;
                 }
 
-                let tool_pair_summarization_task = crate::context_mgmt::maybe_summarize_tool_pair(
+                let tool_pair_summarization_task = crate::context_mgmt::maybe_summarize_tool_pairs(
                     self.provider().await?,
                     session_config.id.clone(),
                     conversation.clone(),
```
jh-block, line 1182: This is pre-existing, so no action needed on this PR, but I noticed we often clone the whole conversation to pass into tasks etc (also in agent.rs, acp code, etc). It might be interesting to look at making it cheaply cloneable since for large conversations (potentially including images etc) it's not a trivial copy

crates/goose/src/context_mgmt/mod.rs
DOsinga: In practice we shouldn't have too many in parallel though - every time we add a toolcall/reply pair, we can summarize an old one

crates/goose/src/agents/agent.rs
DOsinga: yeah, good point. mostly we don't need to copy the messages at all

crates/goose/src/context_mgmt/mod.rs
DOsinga: I don't know, in my experience if the LLM encountered these messages, it was just very confused. the messages we produce now look like:

```
The codebase was analyzed to understand the project structure and Cargo.toml configuration was retrieved. The analysis showed a Rust workspace project (version 1.24.0) with multiple crates, containing build scripts, documentation, and test utilities. The Cargo.toml revealed workspace dependencies and configuration for the "goose" AI agent project.
```

that seems pretty alright? if the sorting is off, the agent will say, why are you giving me a summary of the work we did, is what I found. maybe we should ask it make it active so it knows it did it? or maybe it should be a message by the agent? Now I understand: <...>

jh-block: Now I have a complete understanding

crates/goose/src/agents/agent.rs
```diff
@@ -351,15 +351,26 @@ impl Agent {
...
             .await?;
 
+        let tool_call_cut_off = match Config::global().get_param::<usize>("GOOSE_TOOL_CALL_CUTOFF")
+        {
+            Ok(v) => v,
+            Err(_) => {
+                let context_limit = self
+                    .provider()
+                    .await
+                    .map(|p| p.get_model_config().context_limit())
```
DOsinga, line 361: I think we should probably apply the auto-compaction limit here too. if I am a user with a 200k limit but I don't want to spend money on tokens at that level, I would probably set the autocompact to pretty low; taking that as signal that we want to do this more aggresiveily then makes sense

crates/goose/src/context_mgmt/mod.rs
```diff
@@ -414,13 +414,23 @@ fn format_message_for_compacting(msg: &Message) -> String {
...
+/// Compute a smart tool call cutoff based on the model's context window size.
+/// Larger context windows can hold more tool calls before summarization is needed.
+/// Returns context_limit / 10_000, clamped to [10, 100].
+pub fn compute_tool_call_cutoff(context_limit: usize) -> usize {
+    (context_limit / 10_000).clamp(10, 100)
+}
+
+/// Find tool call IDs to summarize. Returns the oldest unsummarized tool call IDs
+/// that exceed the cutoff, up to `batch_size` at once.
+pub fn tool_ids_to_summarize(
```
DOsinga, line 432: ideally we would also insert here where the current turn started and never summarize beyond that so that if you have a very long back & forth toolcalling sequence it doesn't start to eat its own tail /cc @baxen

crates/goose/src/context_mgmt/mod.rs
michaelneale: even for local, or mesh, concurrent can be ok as the bottleneck may be rpc to rpc leaving other layers idle so parallel never a terrible idea.. but shoudl we do that now?

crates/goose/src/context_mgmt/mod.rs
michaelneale: the way I noted it was happening was self reflection from those confusing messages in the past

crates/goose/src/agents/agent.rs
michaelneale: right so f(context_limit . threshold_chosen) type of thing @DOsinga ?

michaelneale: or rather, if they have it compat at 30k instead of 200k full window, then use the 30k as the input to the calculation of when to compact tools?

michaelneale: yep, PTAL

michaelneale:
@DOsinga address 2 things, protecting latest one, and using the compaction threshold as scaling factor  for tool call compaction - you think this is worthwhile to get in? or have something else cooking?

michaelneale:
I think will need to take another run at this

DOsinga:
I can adopt this and make it work the way we said this week?

PR #7472, opened 2026-02-24
Fix out of order messages

DOsinga:
## Summary

we weren't sorting the messages by creation, but rather by database timestamp. also make sure the messages are explicitly sorted after tool call/response summarization

### DIFF
--- a/crates/goose/src/conversation/mod.rs
+++ b/crates/goose/src/conversation/mod.rs
@@ -103,10 +103,6 @@ impl Conversation {
         self.0.clear();
     }
 
-    pub fn sort_by_created(&mut self) {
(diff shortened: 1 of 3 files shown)


crates/goose/src/conversation/mod.rs
```diff
@@ -103,6 +103,10 @@ impl Conversation {
         self.0.clear();
     }
 
+    pub fn sort_by_created(&mut self) {
+        self.0.sort_by_key(|m| m.created);
+    }
+
```
jamadeo: what about making Conversation always keep them sorted? this method could be private and called on every mutable method

PR #7474, opened 2026-02-24
Apps token limit

DOsinga:
## Summary

makes the canonical model config drive the output tokens that we set for the providers

### DIFF
--- a/crates/goose/src/model.rs
+++ b/crates/goose/src/model.rs
@@ -280,7 +280,7 @@ impl ModelConfig {
             return tokens;
         }
 
-        16_384
+        4_096
     }
 
     pub fn new_or_fail(model_name: &str) -> ModelConfig {


crates/goose/src/providers/formats/databricks.rs
```diff
@@ -613,19 +613,19 @@ pub fn create_request(
...
+            .ok()
+            .and_then(|s| s.parse().ok())
+            .unwrap_or(DEFAULT_THINKING_BUDGET);
+
+        // With thinking enabled, max_tokens must include both output and thinking budget
+        let max_tokens = model_config.max_output_tokens() + budget_tokens;
+        payload
+            .as_object_mut()
+            .unwrap()
+            .insert("max_tokens".to_string(), json!(max_tokens));
```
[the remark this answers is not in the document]
DOsinga: i32 is 2 billion

crates/goose/src/model.rs
```diff
@@ -280,7 +280,6 @@ impl ModelConfig {
             return tokens;
         }
 
-        // Priority 2: Global default
         4_096
```
katzdave, line 283: Was separately thinking about making this bigger. Too small seems strictly worse as it can lead to truncation.

DOsinga: I had it higher for a bit, but presumably we almost always have a real value here so then erring on a lower side is probably better

PR #7476, opened 2026-02-24
fix(apps): forward _meta in tool results to MCP Apps

### LINKED ISSUE #7435: Goose 1.25.0 does not pass `toolResult._meta` to MCP Apps
**Describe the bug**

Per [this section of the MCP Apps spec](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx#2-tool-result-via-uinotificationstool-result-notification), the host is supposed to forward the entire tool execution result to the MCP App view, including the `_meta` property.

Goose 1.24.0 did this correctly, but Goose 1.25.0 no longer includes `_meta`. The regression was introduced in https://github.com/block/goose/pull/7013, specifically here:

https://github.com/block/goose/blob/717d17a35f8c366877af59515a4ba01aae0b0f53/ui/desktop/src/components/McpApps/McpAppRenderer.tsx#L637-L640

---

**To Reproduce**

```ts
// server
registerAppTool(
  server,
  "show-mcp-app",
  {
    title: "Show MCP App",
    description: "Show MCP App",
    _meta: { ui: { resourceUri: MCP_APP_URI } },
  },
  async () => {
    return {
      content: [{ type: "text", text: "App" }],
      _meta: { message: "hello world" },
    };
  },
);

// client
app.ontoolresult = (result) => {
  console.log(result);
}
```

---

**Expected behavior**
When the MCP App renders in the host, it should print an object containing `"hello world"` to the console.

---

**Screenshots**
n/a

---

- **OS & Arch:** macOS 15.7.4. arm64
- **Interface:** UI
- **Version:** 1.25.0
- **Extensions enabled:** MCP Apps
- **Provider & Model:** n/a

---

**Additional context**

aharvard:
## Summary

Fixes #7435

The `_meta` property from tool results was being silently dropped when forwarding to MCP Apps via the `@mcp-ui/client` SDK's `AppRenderer` component. This is a regression introduced in #7013.

## Root Cause

When #7013 refactored `McpAppRenderer` to use the SDK's `AppRenderer`, the custom `McpAppToolResult` type was created without `_meta`, and the conversion logic in the `appToolResult` memo only mapped `content` and `structuredContent`.

The upstream data (from rmcp's `CallToolResult`) **does** include `_meta` on the wire — it was just being discarded at the TypeScript boundary.

## Changes

- **`types.ts`**: Add `_meta?: { [key: string]: unknown }` to `McpAppToolResult`
- **`McpAppRenderer.tsx`**: Forward `_meta` in the `appToolResult` memo

## Testing

- TypeScript compiles clean
- ESLint passes
- All 340 UI tests pass
- Manually verified with an MCP App that reads `toolResult._meta`

### DIFF
--- a/ui/desktop/src/components/McpApps/types.ts
+++ b/ui/desktop/src/components/McpApps/types.ts
@@ -40,6 +40,7 @@ export type McpAppToolCancelled = McpUiToolCancelledNotification['params'];
 export type McpAppToolResult = {
   content: Content[];
   structuredContent?: unknown;
+  _meta?: { [key: string]: unknown };
 };
 
 export type SamplingMessage = {
--- a/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
+++ b/ui/desktop/src/components/McpApps/McpAppRenderer.tsx
@@ -637,6 +637,7 @@ export default function McpAppRenderer({
     return {
       content: toolResult.content as unknown as CallToolResult['content'],
       structuredContent: toolResult.structuredContent as { [key: string]: unknown } | undefined,
+      _meta: toolResult._meta,
     };
   }, [toolResult]);
 


ui/desktop/src/components/McpApps/types.ts
```diff
@@ -40,6 +40,7 @@ export type McpAppToolCancelled = McpUiToolCancelledNotification['params'];
 export type McpAppToolResult = {
```
DOsinga: do we need this type? isn't this just CallToolResult?

aharvard: No, we don't, but upon further investigation, we are sending camel case over the wire and should make updates in the backend. Will follow up with another PR for that.

aharvard: here's the follow-up PR https://github.com/block/goose/pull/7487

PR #7484, opened 2026-02-24
fix(acp): don't fail session creation when model listing is unavailable

### LINKED ISSUE #7427: Cannot use Azure OpenAI deployment endpoints (no `/models` endpoint)
# Cannot use Azure OpenAI deployment endpoints (no `/models` endpoint)

## Environment
- **Goose Desktop**: v1.25.0
- **OS**: Windows 11
- **Provider**: Azure OpenAI (deployment-specific endpoint)

## Problem

I'm trying to use Goose with my company's Azure OpenAI endpoint, but Goose Desktop shows an error during setup: 

> "Could not contact provider - Failed to fetch models: Server error (500 Internal Server Error)" or "Resource not found (404)"

However, the API works perfectly when I test it directly with curl/PowerShell - I can send messages and get responses without any issues.

## What's happening

Our company provides Azure OpenAI access through a custom gateway with deployment-specific endpoints:
```
https://internal-api.company.com/openai/deployments/gpt-4o/chat/completions
```

When I configure this as a custom provider in Goose, the validation fails because:
1. Goose's model validation expects to find `/models` at a predictable location
2. With Azure OpenAI deployment URLs (which include the full path to `chat/completions`), Goose can't properly construct the models endpoint path
3. The validation fails with a "Resource not found" or "Server error", even though the chat endpoint itself works perfectly

This is a standard Azure OpenAI deployment URL format where the model is already specified in the URL itself, so there's no separate models listing endpoint exposed.

## Why I need this

Many organizations use Azure OpenAI with deployment-specific endpoints rather than the full Azure OpenAI API. This is especially common when:
- Companies run Azure OpenAI behind a custom gateway/proxy for security and access control
- IT provides pre-configured deployment URLs for different teams (e.g., `deployments/gpt-4o`, `deployments/gpt-4o-mini`)
- Access is restricted to specific deployments only, without exposing the management/listing APIs
- The endpoint follows Azure OpenAI's deployment pattern: `/openai/deployments/{deployment-name}/chat/completions`

In Azure OpenAI's design, when you have a deployment URL, you already know which model you're using - it's in the URL itself. The `/models` listing endpoint is typically used at the account level, not the deployment level. So our IT-provided deployment endpoint doesn't need (or expose) a models listing, but it's perfectly valid and functional for chatting.

I'd love to use Goose with our company's Azure OpenAI setup, but currently can't get past the setup screen because of this validation requirement.

## Expected behavior

It would be great if Goose could either:
1. Use the model list I specify in the custom provider JSON file, without requiring API validation
2. Show a warning instead of an error when `/models` returns 404, and let me proceed anyway
3. Have a `skip_model_validation: true` option in the provider config

I already know which model I'm using (it's in the endpoint URL), so I don't need Goose to fetch the list dynamically.

## Current workaround

I found I can work around this by:
1. Closing the error dialog when it appears
2. Setting environment variables (`GOOSE_PROVIDER` and `GOOSE_MODEL`) before launching
3. Starting to type in the chat - it actually works!

But this is confusing because the error message makes it seem like the provider is broken, when it actually works fine for chatting.

## What would help

If there was a simple config option like:

```json
{
  "name": "company_azure_openai",
  "engine": "openai",
  "base_url": "https://internal-api.company.com/openai/deployments/gpt-4o/chat/completions",
  "models": [
    {"name": "gpt-4o", "context_limit": 128000}
  ],
  "headers": {
    "api-key": "xxx",
    "api-version": "2024-12-01-preview"
  },
  "skip_model_validation": true
}
```

Then Goose would trust the models I listed instead of trying to fetch them from the API.

## Example

Here's how my Azure OpenAI deployment endpoint works (tested with PowerShell):

```powershell
# This works perfectly - I can chat with Azure OpenAI through our gateway
Invoke-WebRequest -Uri 'https://internal-api.company.com/openai/deployments/gpt-4o/chat/completions' `
  -Method POST `
  -Headers @{'api-key'='xxx'; 'api-version'='2024-12-01-preview'} `
  -Body '{"messages":[{"role":"user","content":"Hello"}]}' `
  -ContentType 'application/json'
# Returns 200 OK with chat response ✅

# But when Goose tries to validate this provider, it can't find/construct
# the correct models endpoint path, causing validation to fail
# Result: Can't use it in Goose Desktop UI ❌
```

**My custom provider config**:
```json
{
  "name": "company_azure_openai",
  "engine": "openai",
  "base_url": "https://internal-api.company.com/openai/deployments/gpt-4o/chat/completions",
  "models": [{"name": "gpt-4o", "context_limit": 128000}],
  "headers": {
    "api-key": "xxx",
    "api-version": "2024-12-01-preview"
  },
  "requires_auth": false
}
```

## Additional notes

I'm not a developer or Rust expert, just a user trying to connect Goose to our company's Azure OpenAI deployment. This seems like a common enterprise setup where IT provides deployment-specific URLs rather than full Azure OpenAI account access. I'd be happy to test any changes or provide more information if it helps!

Thanks for considering this feature request - Goose is awesome and I'd love to use it with Azure OpenAI! 🙏

idosavion:
## Summary

Fixes #7427

`build_model_state` treats model listing as a hard gate on session creation — if the GET to the models endpoint fails, `on_new_session` returns an error and the session never starts. This breaks proxy-based and deployment-specific providers that only implement the chat completions endpoint (no `/models` route).

Now `build_model_state` catches the error, logs a warning, and returns an empty model list. The session starts normally; model selection in the UI just won't be available. This matches providers like Azure and xAI that already return empty model lists through the default `fetch_supported_models` impl.

### Type of Change
- [x] Bug fix
- [x] Tests

### AI Assistance

### Testing

- `test_build_model_state::fetch_error_falls_back_to_current_model_only` — previously asserted `Err(_)`, now asserts the function returns the current model with an empty available-models list
- `cargo fmt`, `cargo clippy -D warnings` clean

### Related Issues

Follow-up to #7112 (model selection support). That PR added `build_model_state` but made it fatal, so any provider without a models endpoint can't start sessions.

### DIFF
--- a/crates/goose-acp/src/server.rs
+++ b/crates/goose-acp/src/server.rs
@@ -276,20 +276,21 @@ async fn add_extensions(agent: &Agent, extensions: Vec<ExtensionConfig>) {
     }
 }
 
-async fn build_model_state(
-    provider: &dyn Provider,
-    current_model: &str,
-) -> Result<SessionModelState, sacp::Error> {
-    let models = provider.fetch_recommended_models().await.map_err(|e| {
-        sacp::Error::internal_error().data(format!("Failed to fetch models: {}", e))
-    })?;
-    Ok(SessionModelState::new(
+async fn build_model_state(provider: &dyn Provider, current_model: &str) -> SessionModelState {
+    let models = match provider.fetch_recommended_models().await {
+        Ok(models) => models,
+        Err(e) => {
+            warn!(error = %e, "failed to fetch models, model selection will be unavailable");
+            vec![]
+        }
+    };
+    SessionModelState::new(
         ModelId::new(current_model),
         models
             .iter()
             .map(|name| ModelInfo::new(ModelId::new(&**name), &**name))
             .collect(),
-    ))
+    )
 }
 
 impl GooseAcpAgent {
@@ -728,7 +729,7 @@ impl GooseAcpAgent {
         );
 
         let model_state =
-            build_model_state(&*provider, &provider.get_model_config().model_name).await?;
+            build_model_state(&*provider, &provider.get_model_config().model_name).await;
 
         Ok(NewSessionResponse::new(SessionId::new(goose_session.id)).models(model_state))
     }
@@ -853,7 +854,7 @@ impl GooseAcpAgent {
         );
 
         let model_state =
-            build_model_state(&*provider, &provider.get_model_config().model_name).await?;
+            build_model_state(&*provider, &provider.get_model_config().model_name).await;
 
         Ok(LoadSessionResponse::new().models(model_state))
     }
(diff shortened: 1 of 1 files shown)


idosavion:
For sanity, built the custom binary and verified the log is emitted as expected and session is starting:
```
{
  "level": "WARN",
  "message": "failed to fetch models, model selection will be unavailable",
  "error": "Request failed: Resource not found (404): ",
  "target": "goose_acp::server"
}
```

idosavion:
@katzdave - thanks for the quick review!
Anything else I need to do in order to get this merged and deployed as part of the next path/minor version?

katzdave:
> @katzdave - thanks for the quick review! Anything else I need to do in order to get this merged and deployed as part of the next path/minor version?

Nope sorry let's merge!

PR #7487, opened 2026-02-24
fix(api): use camelCase in CallToolResponse and add type discriminators to ContentBlock

aharvard:
## Summary

Fixes the `CallToolResponse` API to use **camelCase** field names matching the MCP spec, and adds proper `type` discriminator fields to content blocks so generated TypeScript types align with the SDK's `CallToolResult`.

### Problem

1. `CallToolResponse` serialized fields as snake_case (`is_error`, `structured_content`) but the MCP spec and SDK expect camelCase (`isError`, `structuredContent`)
2. The `Content` type from rmcp includes a `type` discriminator on the wire (via `#[serde(tag = "type")]`) but utoipa can't reflect this for external types, so the generated OpenAPI schema and TypeScript types were missing it
3. `McpAppToolResult` was a redundant intermediate type that duplicated `CallToolResult`

### Changes

**Rust Backend** (`agent.rs`):
- Added `#[serde(rename_all = "camelCase")]` to `CallToolResponse`
- Changed `content` to `Vec<Value>` with schema-only `ContentBlock` types that include `type` discriminator literals (`"text"`, `"image"`, `"resource"`, `"audio"`, `"resource_link"`)
- Manual `ToSchema` impls to work around utoipa's inability to reflect rmcp's `#[serde(tag = "type")]`

**Frontend**:
- Removed redundant `McpAppToolResult` type
- `McpAppRenderer` now accepts `CallToolResult` directly — no conversion useMemo
- `ToolCallWithResponse` uses `ContentBlock` (with proper `type` field) instead of old `Content`

### Why schema-only types?

rmcp's schemars schema is correct (it reflects the `type` discriminator), but goose uses utoipa for OpenAPI generation. utoipa can't derive the right schema for rmcp's externally-defined tagged enum. The schema-only types exist solely to produce a correct OpenAPI spec — actual serialization goes through `serde_json::Value`.

Follows up on #7476

### DIFF
--- a/ui/desktop/src/components/ToolCallWithResponse.tsx
+++ b/ui/desktop/src/components/ToolCallWithResponse.tsx
@@ -124,7 +124,6 @@ function McpAppWrapper({
   const toolInput = { arguments: toolArguments || {} };
 
   const resultWithMeta = toolResponse?.toolResult as ToolResultWithMeta | undefined;
-  // Wire format matches CallToolResult but generated types differ structurally
   const toolResult =
     resultWithMeta?.status === 'success' && resultWithMeta.value
       ? (resultWithMeta.value as unknown as CallToolResult)


DOsinga:
> Why schema-only types?

I am not sure about this. It looks complicated and since it constructs the types independently of the actual Rust code, only give us the illusion of type safety

aharvard:
> > Why schema-only types?
> 
> I am not sure about this. It looks complicated and since it constructs the types independently of the actual Rust code, only give us the illusion of type safety

The goal is to get a discriminated union for TS, the current way RMCP approaches this lacks the `type` field to discriminate from. Do you think there something we can adjust in RMCP that makes both Rust and our generated types happier? cc @alexhancock

jamadeo:
@aharvard I pulled your branch and tried adding support to the existing schemars-to-utoipa macros we have in openapi.rs, and I think the result looks like what we want: https://github.com/block/goose/pull/7792 (that's a PR targeting this branch)

Feel free to merge it into this if it accomplishes what you're doing. The types are defined slightly differently, but you still get a discriminated union at the end

aharvard:
thanks @jamadeo, just merged in your PR (will work on cleaning up this PR next)

jamadeo (approved):

PR #7492, opened 2026-02-24
fix: test fixes for the new develop extension

alexhancock:
Iterating on test fixes


alexhancock:
these are now fixed in the `develop2` branch

PR #7511, opened 2026-02-25
Fix Windows MSVC linking issues

jh-block:
Fixes several linking conflicts when building on Windows MSVC targets.

## Changes

1. **Disable default features on `tokenizers`** — The `esaxx-rs` transitive dependency (via tokenizers' default features) has optional native code that unconditionally links against the static CRT, causing conflicts. Retains needed functionality via the `onig` feature.

2. **Switch `llama-cpp-2` to fork** — Uses a git dependency on jh-block/llama-cpp-rs (`goose-patches` branch, [changes here](https://github.com/utilityai/llama-cpp-rs/compare/main...jh-block:llama-cpp-rs:goose-patches)) which contains a fix to detect whether the static-crt target feature is enabled rather than assuming dynamic.

3. **Patch `v8` crate via `v8-goose`** — The upstream `v8` crate detects CRT linkage using `CARGO_FEATURE_CRT_STATIC`, which is testing a non-existent crate feature instead of correctly checking the target feature. The patched version correctly uses `CARGO_CFG_TARGET_FEATURE` instead. Implemented as a thin `vendor/v8` shim that re-exports `v8-goose` (published on crates.io), because a git dependency would pull in all the v8 submodules which are enormous.

4. **Allow duplicate symbols on Windows MSVC** — Both `v8` and `llama-cpp-rs` statically link identical `std::exception_ptr` symbols from the C++ runtime. Adds `/FORCE:MULTIPLE` linker flag for MSVC targets to accept these duplicate definitions.

## Testing

I've tested debug and release builds on my x86_64 Windows machine; they build successfully and function correctly.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,5 @@
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-C", "link-args=/FORCE:MULTIPLE"]
+
+[target.aarch64-pc-windows-msvc]
+rustflags = ["-C", "link-args=/FORCE:MULTIPLE"]


lifeizhou-ap:
Nice one! 

I did not find we set CARGO_CFG_TARGET_FEATURE, so we use dynamic for both V8 and llama cpp v2?  Just to make sure my understanding is correct. 

Also maybe we can add a step in CI to compile with windows runner so that next time we can have earlier feedback, could be a follow PR.

jh-block:
> I did not find we set CARGO_CFG_TARGET_FEATURE, so we use dynamic for both V8 and llama cpp v2? Just to make sure my understanding is correct.

We leave it at the default, which is dynamic CRT on Windows. This is unchanged from before this PR; this PR is just rationalising how all our dependencies handle CRT selection so that they all match, since you can't mix static and dynamic CRT.

The way to change that to static, if we ever wanted to change it, would be to add "-Ctarget-feature=+crt-static" to RUSTFLAGS (in env var or in `.cargo/config.toml`). `CARGO_CFG_TARGET_FEATURE` is an env var that Cargo itself sets when running build scripts.

> Also maybe we can add a step in CI to compile with windows runner so that next time we can have earlier feedback, could be a follow PR.

Yes, I think this is a good idea - I will take a look at this.

kskarthik:
Interesting!

BTW `v8-goose` is required for goose-cli ? https://github.com/block/goose/discussions/7272

jh-block:
> Interesting!
> 
> BTW `v8-goose` is required for goose-cli ? #7272

The code mode extension (allowing the agent to write code that calls tools, rather than tying tool calls together with inference) uses v8, and it's currently non-optional. I can take a look at making that a cargo feature if you are OK with building without code mode.

kskarthik:
Due to the nature of OBS which disallows downloading binaries during build time, we need to build v8 from source, which might take a long time. Having it optional is appreciated.

But, let me see how long it takes to build v8 now.

jh-block:
@kskarthik I opened #7567 to make it optional

PR #7529, opened 2026-02-26
chore: openai reasoning model cleanup

lifeizhou-ap:
## Summary
- removed deprecate max_token request param in openai compatible api
- use canonical model reasoning data to check whether model is reasoning model

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [X] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Unit test and manual testing

### DIFF
--- a/crates/goose/src/providers/formats/openai_responses.rs
+++ b/crates/goose/src/providers/formats/openai_responses.rs
@@ -827,7 +827,7 @@ mod tests {
             toolshim_model: None,
             fast_model_config: None,
             request_params: None,
-            reasoning: None
+            reasoning: None,
         };
 
         let messages = vec![


crates/goose/src/model.rs
```diff
@@ -275,6 +278,18 @@ impl ModelConfig {
         self.context_limit.unwrap_or(DEFAULT_CONTEXT_LIMIT)
     }
 
+    pub fn is_openai_reasoning_model(&self) -> bool {
+        const PROVIDER_PREFIXES: &[&str] = &["goose-", "databricks-"];
```
michaelneale: provider or model name? as that seems odd?

lifeizhou-ap:
> much nicer but provider/model name prefix is a little confusing for me.

yes @michaelneale! I was a bit confused initially too.  Looking into the code,  I found the databricks provider currently is much for block databricks which we know the convention of our model names.  If we want to have a general purpose databricks provider, we have to have a common schema for returning the servingpoints.

crates/goose/src/model.rs
lifeizhou-ap: should be DATABRICKS_MODEL_NAME_PREFIXES

PR #7531, opened 2026-02-26
refactor: split text_editor into read/edit/write tools with fuzzy matching and unified diffs

jh-block:
## Summary

Ports key ideas from pi's text editor implementation to improve goose's developer tools.

### Tool split
- Replaces the monolithic `text_editor` tool (with `command` subparam and 7 optional fields) with three focused tools: `read_file`, `edit_file`, `write_file`
- Simpler schemas mean fewer LLM errors — each tool only has the parameters it needs
- Removes `insert` (line-number-based insertion is error-prone) and `undo_edit` (rarely used effectively)
- Removes in-memory `file_history` state

### Fuzzy text matching
- Tries exact match first, falls back to fuzzy normalization
- Strips trailing whitespace per line
- Normalizes smart quotes (`\u201C\u201D\u2018\u2019` → `"'`), Unicode dashes (em-dash, en-dash → `-`), and special spaces (NBSP etc → ` `)
- Fixes common LLM-introduced character differences that previously caused match failures

### Encoding preservation
- BOM stripping/preservation for UTF-8 files
- Detects and preserves original line endings (CRLF/LF) instead of forcing platform default

### Unified diff output
- User-facing edit output now shows a proper unified diff with line numbers and context (using `similar` crate) instead of a raw code snippet

### Downstream updates
- CLI output rendering, export, and ACP server updated to recognize new tool names alongside legacy `developer__text_editor`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4492,6 +4492,7 @@ dependencies = [
  "serial_test",
  "shell-words",
  "shellexpand",
+ "similar",
  "tempfile",
  "tokio",
  "tokio-stream",
--- a/crates/goose-acp/src/server.rs
+++ b/crates/goose-acp/src/server.rs
@@ -109,7 +109,15 @@ fn extract_tool_locations(
     let mut locations = Vec::new();
 
     if let Ok(tool_call) = &tool_request.tool_call {
-        if tool_call.name != "developer__text_editor" {
+        let tool_name = tool_call.name.to_string();
+        let is_file_tool = matches!(
+            tool_name.as_str(),
+            "developer__text_editor"
+                | "developer__read_file"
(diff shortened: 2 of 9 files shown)


michaelneale:
yeah interesting - I think would want to have a mode which does not ask for diffs for small models - which is tricky, I know, but don't see another way to scale down at this time. 

There is actually one other way @jh-block: fast apply models. Not many are open yet, but there may be some soon, this means the editing can be offloaded to a different model and it can make a smaller LLM work better.

lifeizhou-ap:
Hey @jh-block 👋 Closing this draft as part of some housekeeping on stale PRs. Feel free to reopen anytime if you pick this back up!

PR #7534, opened 2026-02-26
Rewrite recipe validation tests

tpbrown:
Rewrite validation.test.ts from 3 trivial tests (32 lines) to 31 comprehensive tests (518 lines).

Covers all resolveRefs code paths: $ref resolution (simple, chained, broken, non-object), allOf merging (with $ref, spread ordering, empty arrays, null entries), oneOf/anyOf variant resolution, nested object property recursion, array items resolution, passthrough for plain schemas, non-object inputs, title/description defaults, and fallback when Recipe schema is missing or spec has no components.

Uses vi.doMock to inject controlled OpenAPI specs per test group. Real-spec integration tests verify actual openapi.json resolution.


tpbrown:
Whoops. My scheduler & recipe picked up your repo instead of the one of the target repos.

Sorry about that!

PR #7538, opened 2026-02-26
Nanogpt custom canonical loader

katzdave:
The data in models.dev is pretty out of date, but nanogpt serves their own model info with similar data.

Adding a way for us to replace model info for specific providers from other sources.

### DIFF
--- a/crates/goose-acp/tests/common_tests/mod.rs
+++ b/crates/goose-acp/tests/common_tests/mod.rs
@@ -13,9 +13,7 @@ use goose::config::GooseMode;
 use goose::providers::provider_registry::ProviderConstructor;
 use goose_acp::server::GooseAcpAgent;
 use goose_test_support::{ExpectedSessionId, McpFixture, FAKE_CODE, TEST_MODEL};
-use sacp::schema::{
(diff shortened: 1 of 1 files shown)


crates/goose-acp/tests/common_tests/mod.rs
```diff
@@ -122,65 +120,72 @@ pub async fn run_model_list<C: Connection>() {
...
-        .map(|id| ModelInfo::new(ModelId::new(*id), *id))
-        .collect(),
-    );
-    assert_eq!(models, expected);
+        .map(|m| m.model_id.0.to_string())
+        .collect();
+    actual_ids.sort();
+
+    let mut expected_ids: Vec<String> = expected_model_ids.iter().map(|s| s.to_string()).collect();
+    expected_ids.sort();
```
katzdave, line 186: The models shift around in order if their dates get updated on refreshing models.dev.

This just treats it as a set now. (although I suspect this is also going to cause problems when new models come)

lifeizhou-ap:
Close this PR as model.dev has included nanogpt models recently.  

updated canonical models in this PR https://github.com/block/goose/pull/7920

PR #7541, opened 2026-02-26
Improve custom provider creation experience

katzdave:
Rather than showing your existing provider, the custom provider panel now shows this (i'm using random characters as an API key) immediately after custom provider creation. 

Screenshot 2026-02-26 at 1 38 42 PM

Some of the providers seem to fetch the models no problem with an invalid key, but others do not; I think this warning is a decent signal that something might be wrong without the noisy full error. And then if you make a conversation, you get the error about invalid API key.

### DIFF
--- a/ui/desktop/src/components/ConfigContext.tsx
+++ b/ui/desktop/src/components/ConfigContext.tsx
@@ -179,7 +179,7 @@ export const ConfigProvider: React.FC<ConfigProviderProps> = ({ children }) => {
         return providersData;
       } catch (error) {
         console.error('Failed to fetch providers:', error);
-        return [];
+        return providersListRef.current;
       }
     }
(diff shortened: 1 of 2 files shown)


ui/desktop/src/components/settings/providers/ProviderSettingsPage.tsx
```diff
@@ -50,11 +50,10 @@ export default function ProviderSettings({
   }, []); // Intentionally not including loadProviders in deps to prevent reloading
 
   // This function will be passed to ProviderGrid for manual refreshes after config changes
-  const refreshProviders = useCallback(() => {
+  const refreshProviders = useCallback(async () => {
```
katzdave, line 53: This is the biggest issue, not blocking here.

PR #7466, opened 2026-02-24
feat: simplify developer extension

baxen:
## Summary
- Simplifies the developer extension by removing ~11,400 lines of code across analyze, editor models, text editor, and rmcp developer modules
- Replaces complex code analysis (`analyze/`), editor model integrations (`editor_models/`), and the large rmcp developer implementation with streamlined `edit.rs`, `tree.rs`, and `shell.rs` modules
- Adds a new platform extension for developer functionality in `goose/src/agents/platform_extensions/developer.rs`
- Updates configuration, security scanning, and session handling to support the simplified architecture

Issue #5155, opened 2025-10-14
[bug] goose gets into endless loop with developer extension


dord44:
**Describe the bug**

we observed that goose gets into endless loop with the developer extension turned on when we work on our repo:

```
Context: ○○○○○○○○○○ 0% (0/400000 tokens)
( O)> inspect the project

─── analyze | developer ──────────────────────────
focus: 
follow_depth: 2
force: false
max_depth: 2
path: /Users/dor/radioquebrada


─── analyze | developer ──────────────────────────
focus: 
follow_depth: 2
force: false
max_depth: 3
path: /Users/dor/radioquebrada
```

this goes on, with `max_depth` increasing forever. it doesn't happen when we work on a different directory. turning off the developer extension fixes the endless loop. different LLMs have been tried and the problem was persistent. different prompts all lead to the same result.

**To Reproduce**
- clone `https://github.com/dord44/dord44.github.io` with latest main (commit `41914bb7d77766957af7cc093b90519fb02a8fdd`)
- start goose with developer extension in that directory
- ask something like "explore the project" 

**Expected behavior**
no endless loop

**Screenshots**

<img width="694" height="283" alt="Image" src="https://github.com/user-attachments/assets/5a4f2844-5656-42ed-bfef-fb8e9e5bed8e" />

**Please provide following information:**
 - **OS & Arch:** macOS
 - **Interface:** both UI and CLI
 - **Version:** 1.9.3
 - **Extensions enabled:** Developer Extension
 - **Provider & Model:** GPT-5 et al.

DOsinga:
hmm, analyze seems to be causing this? /cc @tlongwell-block @wpfleger96

wpfleger96:
I can take a look at this

DOsinga:
should we also move analyze into its own extension? smaller models have a hard time with the size of the developer extension

alexhancock:
can we disable the analyze tool until we know what's going on with it? I'd like to see us bias towards reverting/disabling things and then re-enabling when known to be stable.

tlongwell-block:
What OS are you on @dord44 ?

tlongwell-block:
Also, I wouldn't expect any tool to cause an infinite loop in the agent as long as it is returning and not doing a prompt injection

tlongwell-block:
Perhaps this is related https://discord.com/channels/1287729918100246654/1426951272451539054

Don't see this issue on my end on mac

```
[...]

PR #7542, opened 2026-02-26
feat: add analyze platform extension with tree-sitter AST parsing

tlongwell-block:
Rebuild the analyze tool as a standalone platform extension after https://github.com/block/goose/pull/7466 merged in, as suggested by DOsinga in Oct 2025 (issue #5155). Single unprefixed 'analyze' tool with three auto-selected modes:

- Structure: directory overview with LOC/function/class counts, language breakdown
- Semantic: per-file functions with signatures and return types, classes with fields and inheritance, imports, call frequency (•N)
- Focused: symbol call graph with BFS incoming/outgoing chains, cross-file resolution, test/production caller separation

10 languages: Rust, Python, JavaScript, TypeScript, TSX, Go, Java, Kotlin, Swift, Ruby. Tree-sitter queries extract functions, classes, imports, and call sites. Rayon parallelism for directory analysis.

Key design decisions:
- 1 tool, 3 auto-selected modes (vs narsil-mcp's 90 tools)
- Compact LLM-optimized output format
- Function signatures with params and return types inline
- Class inheritance shown for all languages
- File-scoped call graph nodes with language isolation
- Deduplicated edges (HashSet) prevent BFS chain explosion
- Tree-formatted chain output with shared-prefix grouping
- Test callers separated from production callers
- Multiline format for large files (>10 symbols)
- Subagent delegation pattern for token budget management

Architecture: 5 files, ~2k LOC, no file over 650 LOC
- mod.rs: client, McpClientTrait impl, mode selection, tests
- parser.rs: tree-sitter parsing, signatures, inheritance extraction
- languages.rs: language registry (pure data, one entry per language)
- format.rs: output formatting, tree dedup, test separation
- graph.rs: call graph, BFS traversal, scoped resolution

Also includes analyze_cli binary for ad-hoc testing.

### DIFF
--- a/crates/goose/src/agents/platform_extensions/analyze/mod.rs
+++ b/crates/goose/src/agents/platform_extensions/analyze/mod.rs
@@ -301,6 +301,7 @@ mod tests {
         PlatformExtensionContext {
             extension_manager: None,
             session_manager: Arc::new(SessionManager::new(std::env::temp_dir())),
+            session: None,
         }
     }
 


crates/goose/src/bin/analyze_cli.rs
```diff
@@ -0,0 +1,88 @@
+//! CLI wrapper for the analyze platform extension.
```
michaelneale, line 1: how come there is an analyze cli separately?

Cargo.toml
```diff
@@ -69,6 +69,17 @@ opentelemetry-appender-tracing = "0.31"
 opentelemetry-stdout = { version = "0.31", features = ["trace", "metrics", "logs"] }
 tracing-opentelemetry = "0.32"
 
+rayon = "1.10"
```
michaelneale, line 72: any idea of heft added by these? I guess we had them before?

crates/goose/src/agents/platform_extensions/mod.rs
```diff
@@ -24,6 +25,19 @@ pub static PLATFORM_EXTENSIONS: Lazy<HashMap<&'static str, PlatformExtensionDef>
...
         let mut map = HashMap::new();
 
+        map.insert(
+            analyze::EXTENSION_NAME,
+            PlatformExtensionDef {
+                name: analyze::EXTENSION_NAME,
+                display_name: "Analyze",
+                description:
+                    "Analyze code structure with tree-sitter: directory overviews, file details, symbol call graphs",
+                default_enabled: true,
```
michaelneale, line 37: could we make this not on by default - let platform put it on?

Cargo.toml
tlongwell-block: Shouldn't be noticeable since this was a dep just a few days ago already

crates/goose/src/agents/platform_extensions/mod.rs
tlongwell-block: It had been on by default, and makes goose *much* better in big repos that devs care about. I think on by default is definitely the way to go here. It's one tool and surprisingly light

crates/goose/src/bin/analyze_cli.rs
tlongwell-block: It lets the agents test it as they develop it. And humans can use it, too. Not sure its worth keeping in, but it makes dev on this tool much, much easier

PR #7560, opened 2026-02-27
feat: make the text bar persistent and add a queue for messages

alexhancock:
Couple things in the ACP-based TUI...

* Make the text entry bar persistent
* Add support for queing up messages

Nice small-ish improvements that set the stage next for slash commands and more to be added

Demo:

Screenshot 2026-02-27 at 10 29 29 AM


alexhancock:
> we should talk about message queuing though. I don't think we should do it. I think if you enter a message while the agent is running, it should just be inserted as an extra message to steer the current process

A cool idea - let's pair on it!

PR #7568, opened 2026-02-27
fix(desktop): harden settings.json read/write path

### LINKED ISSUE #7556: Desktop: settings.json parse is unguarded; corrupted file can crash app at startup
### Severity
Medium (robustness / user lockout)

### Where
`ui/desktop/src/main.ts`

### What was observed
`getSettings()` reads `settings.json` and does a direct `JSON.parse` with no try/catch fallback.

`updateSettings()` writes the file directly, without temp-file + rename behavior, increasing the chance of partial writes on crash/power loss.

### Why this matters
A single truncated/corrupted JSON file can brick the UI until the user manually finds and deletes it.

WhiskerSage:
# PR Title

fix(desktop): guard settings.json parse and write settings atomically

# PR Body

## Summary

Fixes a desktop startup robustness issue where a corrupted `settings.json` can crash app startup.

## Problem

`getSettings()` in `ui/desktop/src/main.ts` directly parsed `settings.json` with no parse guard.  
A truncated/corrupted file could throw during startup and block the app from launching.

`updateSettings()` also wrote directly to `settings.json`, which increases risk of partial writes during interruption.

## Changes

- Add safe fallback in `getSettings()`:
  - return defaults when file is missing
  - wrap parse/read in `try/catch`
  - log warning and fallback to `defaultSettings` on parse failure
- Harden `updateSettings()` write path:
  - write to `settings.json.tmp` first
  - replace via `rename`
  - handle Windows replace edge cases (`EEXIST` / `EPERM` / `ENOTEMPTY`) by removing destination then renaming
  - clean up temp file in `finally`

### Repro on unpatched code

1. Write invalid JSON to `%APPDATA%/Goose/settings.json`
2. Start desktop app
3. Startup flow fails in main process path

### Verify on patched code

1. Keep invalid `%APPDATA%/Goose/settings.json`
2. Start desktop app
3. App launches successfully; settings fallback is applied

Also ran:

- `npm run typecheck` in `ui/desktop` (pass)

## Risk

Low:

- Changes are isolated to settings read/write logic in desktop main process
- Existing behavior is preserved for valid settings files
- Failure mode is improved from startup crash to safe defaults

Fixes #7556

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -135,7 +135,6 @@ function updateSettings(modifier: (settings: Settings) => void): void {
     fsSync.renameSync(SETTINGS_FILE, backupSettingsFile);
     try {
       fsSync.renameSync(tempSettingsFile, SETTINGS_FILE);
-      fsSync.unlinkSync(backupSettingsFile);
     } catch (replaceError) {
       if (fsSync.existsSync(backupSettingsFile)) {
         if (fsSync.existsSync(SETTINGS_FILE)) {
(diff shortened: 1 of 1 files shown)


DOsinga:
this is fixed in a simpler way that I think we prefer for now. thanks for helping though!

PR #7571, opened 2026-02-27
fix: pass OAuth scopes to DCR and extract granted_scopes from token response

peschee:
## Summary

Two fixes for OAuth scope handling when connecting to MCP servers:

- **Pass scopes to DCR registration**: When an MCP server returns `WWW-Authenticate: Bearer scope="..."`, the scopes are parsed but never included in the Dynamic Client Registration (DCR) request. This patches rmcp via `[patch.crates-io]` to include a `scope` field in the DCR request per [RFC 7591](https://datatracker.ietf.org/doc/html/rfc7591). Upstream PR: modelcontextprotocol/rust-sdk#705
- **Extract granted_scopes from token response**: Previously `granted_scopes: vec![]` was always saved. Now the actual scopes from the token response are extracted and stored in credentials.

## Test plan

- [x] `cargo check -p goose` compiles without errors
- [ ] Connect to an MCP server that returns scopes in `WWW-Authenticate` and verify the DCR registration request includes the `scope` field
- [ ] Verify stored credentials contain the granted scopes after OAuth flow completes
- [ ] Once upstream rmcp PR is merged and released, update to remove the `[patch.crates-io]` entry

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,4 +83,3 @@ tree-sitter-typescript = "0.23"
 
 [patch.crates-io]
 v8 = { path = "vendor/v8" }
-rmcp = { git = "https://github.com/peschee/rust-sdk.git", branch = "fix-dcr-scopes" }


crates/goose/src/agents/platform_extensions/developer/shell.rs
```diff
@@ -429,6 +429,8 @@ mod tests {
         let path1 = save_full_output("first").unwrap();
         let path2 = save_full_output("second").unwrap();
         assert_eq!(path1, path2);
-        assert_eq!(std::fs::read_to_string(&path2).unwrap(), "second");
+        // Note: we intentionally don't assert file content here because
+        // parallel tests (render_output_truncates_*) share the same static
+        // temp file and can overwrite the content between our write and read.
```
jamadeo, line 521: yeah, this appears to be a flaky test. Looks like removing this assert loses very little actual test usefulness, so this seems fine to me

peschee:
@jamadeo upstream has been merged

jamadeo:
thanks @peschee , once #7619 merges, we can remove the dependency override

PR #7576, opened 2026-02-27
Restore old system prompt behavior around subagents

jamadeo:
with @alexhancock

### DIFF
--- a/crates/goose/src/agents/platform_extensions/summon.rs
+++ b/crates/goose/src/agents/platform_extensions/summon.rs
@@ -499,10 +499,12 @@ impl SummonClient {
             let mut instructions = "".to_string();
             let sources = discover_filesystem_sources(&session.working_dir);
 
-            let skills: Vec<&Source> = sources
+            let mut skills: Vec<&Source> = sources
                 .iter()
                 .filter(|s| s.kind == SourceKind::Skill || s.kind == SourceKind::BuiltinSkill)
                 .collect();
+
+            skills.sort_by(|a, b| (&a.name, &a.path).cmp(&(&b.name, &b.path)));
 
             if !skills.is_empty() {
                 instructions.push_str("\n\nYou have these skills at your disposal, when it is clear they can help you solve a problem or you are asked to use them:");


crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__all_platform_extensions.snap
```diff
@@ -87,7 +87,10 @@ Use write and edit to efficiently make changes. Test and verify as appropriate.
 ## summon
 
 ### Instructions
-Load knowledge and delegate tasks to subagents using the summon extension.
+
+
+You have these skills at your disposal, when it is clear they can help you solve a problem or you are asked to use them:
+• goose-doc-guide - Reference goose documentation to create, configure, or explain goose-specific features like recipes, extensions, sessions, and providers. You MUST fetch relevant goose docs before answering. You MUST NOT rely on training data or assumptions for any goose-specific fields, values, names, syntax, or commands.
```
DOsinga, line 93: not for now, but maybe where this is generated we can throw in some skills so they actually show up here. hmm, actually, actually, should we even show this if there are no skills?

jamadeo: this `goose-doc-guide` is a skill that currently always appears, because it's built into the binary

crates/goose/src/agents/platform_extensions/summon.rs
```diff
@@ -278,6 +495,28 @@ impl Drop for SummonClient {
...
+                .iter()
+                .filter(|s| s.kind == SourceKind::Skill || s.kind == SourceKind::BuiltinSkill)
+                .collect();
+
+            skills.sort_by(|a, b| (&a.name, &a.path).cmp(&(&b.name, &b.path)));
+
+            if !skills.is_empty() {
+                instructions.push_str("\n\nYou have these skills at your disposal, when it is clear they can help you solve a problem or you are asked to use them:");
+                for skill in &skills {
+                    instructions.push_str(&format!("\n• {} - {}", skill.name, skill.description));
```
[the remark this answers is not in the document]
jamadeo: This is a good point, but I think in general we trust the working directory in goose. WDYT @DOsinga @alexhancock ?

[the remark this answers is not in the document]
DOsinga: I think this is how skills are supposed to work - you install an evil skill and your agent becomes evil. YOLO all the way

tlongwell-block, line 512: Should we put a limit on the max number of skills in the description? The skill monorepos can have dozens or hundreds

Issue #6957, opened 2026-02-04
GitHub Copilot authentication stuck on "Signing in..." in Windows Desktop (v1.23.0)


Rouba-M-S:
## Description
GitHub Copilot authentication in Goose Desktop on Windows gets stuck at the "Signing in..." screen. The browser window does not automatically open, and no authentication code is displayed or copied to the clipboard.

## Environment
- **OS**: Windows 11
- **Goose Version**: v1.23.0
- **Installation Method**: Direct download (Goose-win32-x64.zip)
- **Goose Type**: Desktop Application
- **GitHub Copilot**: Active subscription confirmed

## Steps to Reproduce
1. Launch Goose Desktop on Windows
2. Select "Other Providers" from the welcome screen
3. Click "Configure providers"
4. Select "GitHub Copilot" from the provider list
5. Click "Configure" or "Sign in with GitHub Copilot"
6. Application displays: "Sign in with your GitHub Copilot account to use this provider" with "Signing in..." and "A browser window will open for you to complete the login."

## Expected Behavior
- Browser should automatically open to `https://github.com/login/device`
- An 8-character authentication code should be displayed in the Goose window
- The code should be automatically copied to clipboard
- User can complete authentication in the browser

## Actual Behavior
- Browser window **does not open** automatically
- **No authentication code** is displayed in the Goose application
- **No code is copied** to clipboard
- Application remains stuck on "Signing in..." screen indefinitely
- No error messages are shown
- No entries are created in `~/.config/goose/` directory

## Attempted Workarounds
- Manually opening `https://github.com/login/device` in browser (no code to enter)
- Checking clipboard with Ctrl+V (nothing pasted)
- Restarting Goose Desktop application
- Running Goose as Administrator
- Checking Windows Firewall settings
- Checking antivirus exclusions

## Additional Context
- The same GitHub Copilot subscription works correctly in VS Code
- Similar issues have been reported for WSL (#6607) but this is native Windows Desktop
- Issue #3068 mentions GitHub Copilot asking for token instead of device flow on Windows
- The Goose Desktop process is running (visible in Task Manager)
- No error logs are generated in `%USERPROFILE%\.config\goose\`

## Logs
No log files are created. The `~/.config/goose/` directory remains empty after attempting authentication.

## Workaround
Using Google Gemini as an alternative provider works correctly with API key authentication.

## Impact
Users on Windows cannot authenticate with GitHub Copilot through the Desktop application, forcing them to:
[...]

PR #7577, opened 2026-02-27
fix: open browser and copy code for github copilot device oauth flow

sheikhlimon:
## Summary
Relates to #6957                                               
 
This PR updates GitHub Copilot's OAuth device code flow to show the code in the UI instead of silently copying it to the clipboard, addressing the UX concern raised in the PR review.

### Changes:
- Added DeviceCodeModal component with code display and auto-polling for completion
- Added discriminated union response types to the OAuth endpoint to keep it compatible with other providers
- Added /config/providers/{name}/oauth/completion endpoint for non-blocking completion checks
- Updated backend to return device code info immediately and poll for completion
- Removed arboard dependency (no longer silently copying to clipboard)

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
- Added 3 serialization tests for DeviceCodeData struct


crates/goose/src/providers/githubcopilot.rs
```diff
@@ -273,9 +274,20 @@ impl GithubCopilotProvider {
     async fn login(&self) -> Result<String> {
         let device_code_info = self.get_device_code().await?;
 
+        let code = device_code_info.user_code.as_str();
+
+        if let Ok(mut clipboard) = arboard::Clipboard::new() {
+            if clipboard.set_text(code).is_ok() {
```
jamadeo: We should avoid writing to the clipboard like this without an explicit action -- the user might have something there they don't want overwritten. Also, I'm not sure the user will know that his code is in their clipboard when the browser opens.

It's more work, but I would suggest looking into modifying the existing /config/providers/{provider}/oauth flow to allow rendering a code in the UI element.

jamadeo (changes requested):
This is a clever way around the issue, but I'm not sure I'd want goose writing directly to my clipboard.

What would it take to properly show the code in the desktop UI?

crates/goose/src/providers/githubcopilot.rs
sheikhlimon: Thanks for pointing that out - the clipboard concern makes sense

sheikhlimon:
> This is a clever way around the issue, but I'm not sure I'd want goose writing directly to my clipboard.
> 
> What would it take to properly show the code in the desktop UI?

From what I understand, this might involve two changes:
1. Updating the `/config/providers/{provider}/oauth` flow to return the device code so the desktop UI can render it.
2. Adding a UI element to display the code with an explicit copy option.

That way the clipboard write could be removed, since the user would be able to copy it themselves.

Let me know if that sounds roughly like the direction you had in mind.

jamadeo:
@sheikhlimon yes that sounds exactly right! For the oauth handler, however we change it will still need to be compatible with the other providers using it, but this is definitely the way to go.

sheikhlimon:
@jamadeo I tried implementing the approach we discussed and got close, but I’m running into some limitations and can’t get the final flow to work reliably.

I don’t think I’ll be able to complete this PR properly, so it’s probably best to close it rather than leave it half-working. Happy for someone else to pick it up if needed.

Also, #8019 covers what I initially aimed for and does a better job in terms of UX.

michaelneale:
thanks @sheikhlimon - will close this for now, but nice work.

PR #7581, opened 2026-02-27
fix(desktop): harden settings.json read/write path

fresh3nough:
Closes https://github.com/block/goose/issues/7556

## Problem

`getSettings()` in `ui/desktop/src/main.ts` reads `settings.json` with a bare `JSON.parse` and no try/catch. A single corrupted, truncated, or empty settings file crashes the app at startup and locks the user out until they manually find and delete the file.

`updateSettings()` writes the file directly with `writeFileSync`, meaning a crash or power loss during the write can leave a partially-written (corrupt) file on disk.

## Changes

- Extracted `getSettings` and `updateSettings` into a new testable module `ui/desktop/src/utils/settingsManager.ts`, parameterized by file path.
- Wrapped `JSON.parse` in a try/catch in `getSettings`. On parse failure, logs a warning and returns `defaultSettings` instead of crashing.
- Changed `updateSettings` to write atomically: writes to a `.tmp` file first, then uses `renameSync` to move it to the target path. This prevents partial writes from corrupting `settings.json`.
- Updated `main.ts` to delegate to the new module.

## Tests

Added `ui/desktop/src/utils/__tests__/settingsManager.test.ts` with 7 tests covering:
- Returns defaults when no settings file exists
- Reads valid settings from file
- Does not throw on corrupted JSON (returns defaults)
- Does not throw on empty file (returns defaults)
- Does not throw on truncated JSON (returns defaults)
- Persists a settings change
- Uses atomic write (temp file + rename) instead of direct writeFileSync

## Reproduction steps

1. `cd ui/desktop && npm install`
2. Run tests: `npx vitest run src/utils/__tests__/settingsManager.test.ts`
3. All 7 tests pass
4. Full suite: `npx vitest run` -- 305 tests pass, 0 regressions
5. `npm run lint:check` passes clean

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -30,7 +30,11 @@ import { ensureWinShims } from './utils/winShims';
 import { addRecentDir, loadRecentDirs } from './utils/recentDirs';
 import { formatAppName, errorMessage } from './utils/conversionUtils';
 import type { Settings } from './utils/settings';
-import { defaultKeyboardShortcuts, getKeyboardShortcuts } from './utils/settings';
+import { getKeyboardShortcuts } from './utils/settings';
+import {
+  getSettings as getSettingsFromFile,
+  updateSettings as updateSettingsFromFile,
+} from './utils/settingsManager';
 import * as crypto from 'crypto';
 import * as yaml from 'yaml';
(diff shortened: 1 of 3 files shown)


DOsinga:
yeah, let's keep it simple that way. thanks for contributing though!

PR #7583, opened 2026-02-27
feat(summon): make skill supporting files individually loadable via load()

wpfleger96:
## Summary

- Supporting files in a skill directory are now individually loadable via `load(source: "skill-name/relative/path")` — e.g. `load(source: "oncall-buddy-help/references/technical-ops.md")`
- Loading a skill (`load(source: "skill-name")`) now lists available supporting files with their `load()` names rather than inlining their content
- All file types (markdown, scripts, configs) are loadable this way, not just `.md` files
- Path traversal protection: canonicalization check prevents symlinks from reading files outside the skill directory
- When a namespaced path isn't found, the error message lists available files for that skill

## Problem

Skill environments without filesystem access (e.g. Block-internal Goose Slackbot deployments) had no way to reach supporting reference files. The previous approach — listing paths with "Use the file tools to read these files" — only worked for environments with the developer extension.

## Behavior

```
load(source: "oncall-buddy-help")
```
Now outputs:
```
## Supporting Files

Skill directory: ~/.claude/skills/oncall-buddy-help

The following supporting files are available:
- references/technical-ops.md → load(source: "oncall-buddy-help/references/technical-ops.md")
- references/diagnostics.md → load(source: "oncall-buddy-help/references/diagnostics.md")

Use load(source: "<skill-name>/<path>") to load individual files into context, or use file tools to read/run them directly.
```

The LLM can then selectively load only the reference it needs, preserving progressive disclosure while remaining fully functional in filesystem-restricted environments.

### DIFF
--- a/crates/goose/src/agents/platform_extensions/summon.rs
+++ b/crates/goose/src/agents/platform_extensions/summon.rs
@@ -169,6 +169,14 @@ fn parse_frontmatter<T: for<'de> Deserialize<'de>>(content: &str) -> Option<(T, 
 fn parse_skill_content(content: &str, path: PathBuf) -> Option<Source> {
     let (metadata, body): (SkillMetadata, String) = parse_frontmatter(content)?;
 
+    if metadata.name.contains('/') {
+        warn!(
+            "Skill name '{}' contains '/' which is not allowed, skipping",
+            metadata.name
+        );
+        return None;
+    }
+
     Some(Source {
         name: metadata.name,
         kind: SourceKind::Skill,
@@ -703,7 +711,8 @@ impl SummonClient {
             return Ok(Some(source));
         }
 
-        if let Some((skill_name, relative_path)) = name.split_once('/') {
+        if let Some((skill_name, raw_relative_path)) = name.split_once('/') {
+            let relative_path = raw_relative_path.replace('\\', "/");
             if let Some(skill) = sources.iter().find(|s| {
                 s.name == skill_name
                     && matches!(s.kind, SourceKind::Skill | SourceKind::BuiltinSkill)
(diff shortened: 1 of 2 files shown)


tlongwell-block:
I think https://github.com/block/goose/pull/7457 might help here?

wpfleger96:
@tlongwell-block so #7457 enables the Slackbot to see that the skill's supplemental/reference material exists, but it's still unreachable/unusable

the skill is structured like:
```
oncall-buddy-help
├── references
│   ├── diagnostics.md
│   ├── hints-template.md
│   └── technical-ops.md
└── SKILL.md
```

and the agent can `load(source: oncall-buddy-help)`, which loads **only** the content from `SKILL.md`. to be able to load the skill's reference content for deeper debugging, it needs to be able to do:
```
load(source: references/technical-ops)
```
and the model does try this, but it fails with:
```
Error: Source 'references/technical-ops' not found. Use load() to see available sources.
```

because currently `load()` **only** loads the root level `SKILL.md` content and for any supplemental content, examples, scripts, etc. the model would need a shell filesystem access to read the file contents directly which the Slackbot can't do.

if we're going to abstract the `SKILL.md` file away by calling `load(source: <skill_name>)` then I'm thinking it makes sense to do the same for skill reference files/supplemental content instead of having a dual path of `load()` vs raw file read

michaelneale:
I think this is ok @wpfleger96  - downside is that this is less "progressive loading" which is one of the tenets of skills. So downside is if there are skills which are mountains and mountains of .md files, it will blow out context. 

tagging @tlongwell-block who would be aware of this limitation and may have thoughts. I think it is ok but I don't know how bloated skills are (the ones I have seen in squareup/agents I think may end up a bit bloated, but I guess you only load one or two).

wpfleger96:
@michaelneale @tlongwell-block 🙃 sorry I totally botched this PR, my intention was to preserve progressive disclosure here not break it!

I just pushed a new change to address this - now instead of inlining supporting files on load, each supporting file is now individually addressable via load():
```
load(source: "oncall-buddy-help/references/technical-ops.md")
```

Loading the base skill still just loads `SKILL.md` and lists available supporting files with their `load()` names so the LLM knows how to reach them when needed:
```
## Supporting Files

The following supporting files are available:
- references/technical-ops.md → load(source: "oncall-buddy-help/references/technical-ops.md")
- references/diagnostics.md → load(source: "oncall-buddy-help/references/diagnostics.md")

Use load(source: "<skill-name>/<path>") to load individual files into context...
```

My intention here was to make skill supplemental files reachable for geese without filesystem access, but not break skill progressive disclosure! What do y'all think

PR #7591, opened 2026-02-28
fix: feed back errors for tool calls with invalid names/params

### LINKED ISSUE #7589: goose sometimes fail to do tool calls to edit files
Running goose version 1.26.1 with gpt-5.3-codex. Initially tried to write a short poem to a text file and it was successful. But any implementation of a small rust command line tool fails to write files in my project directory. Example output from my goose session below:

```
  ▸ execute 3 tool calls
    1. todo/todoWrite Initialize todo list with explicit and implicit project requirements
    2. developer/shell Scaffold/update Rust CLI project files, write Cargo.toml and src/main.rs, then build (uses 1)
    3. todo/todoWrite Mark implemented tasks as completed after successful build (uses 2)

Implemented attempt hit a tooling mismatch: the `Todo.todoWrite` 
API here doesn’t accept a `todos` field (different schema than expected), 
so that batch call failed before file writes.

I can proceed immediately by rerunning in **one batched execute call** (as requested) to:
1. scaffold/update the Rust project,
2. write `Cargo.toml` and `src/main.rs`,
3. run `cargo build`.
```

And I continuously need to answer "yes, please go ahead" after exits like this:

```
Thanks — I see the exact remaining issues (HTML string quotes and `'` escape in `esc()`).

I’ll do one more **single batched execute** to:
- patch `src/main.rs` with proper escaped string literals,
- rebuild with `cargo build`,
- and report success.
```

It seems like goose erroneously ends up in an exit state when it was supposed to continue on its own.

- **OS & Arch:** [MacOS ARM]
- **Interface:** [CLI]
- **Version:** [1.26.1]
- **Extensions enabled:** [Developer tools]
- **Provider & Model:** [OpenAI – gpt-5.3-codex]

---

fresh3nough:
## Summary

Fixes #7589

When a model generates tool calls with invalid function names (e.g. `todo/todoWrite` with a `/` character) or unparseable JSON arguments, the tool call parsing in the provider layer creates `Err` tool requests. Previously these errored tool calls were **silently dropped** from the conversation -- no error response was ever sent back to the model. This caused two problems:

1. The model never received feedback about why its tool calls failed, so it could not correct itself
2. The agent loop would continue with an unchanged conversation (or the model would generate a text-only response describing what it *would* do, causing the agent to exit prematurely and require the user to say "yes, please go ahead")

## Root Cause

In the agent loop (`agent.rs`), the final tool request processing loop at line ~1363 only processes tool requests where `tool_call.is_ok()`. Errored tool calls (`tool_call.is_err()`) are skipped entirely -- no tool request message, no tool response message, no error feedback. The model is left blind to the failure.

## Fix

After processing valid tool calls, collect error messages from any errored tool requests and send them back to the model as a user text message. This gives the model explicit feedback like:

```
The following tool calls could not be processed:
- The provided function name 'todo/todoWrite' had invalid characters, it must match this regex [a-zA-Z0-9_-]+
Please check the available tool names and parameters and try again.
```

The model can then correct its tool names (e.g. use `todo__todoWrite` instead of `todo/todoWrite`) and continue autonomously without requiring user intervention.

## Testing

- Added two new unit tests in `reply_parts.rs` that verify errored tool calls are correctly categorized and that error feedback messages are properly constructed
- All 575 existing tests pass
- cargo fmt and cargo clippy clean

## Steps to Reproduce (before fix)

1. Start a goose session with OpenAI gpt-5.3-codex (or any model that occasionally hallucinates tool names with `/` separators instead of `__`)
2. Ask goose to build a small Rust CLI project
3. Observe the model generates tool calls like `todo/todoWrite`, `developer/shell`
4. These tool calls are silently dropped; the model outputs text like "I can proceed immediately by rerunning..." but stops
5. User must manually say "yes, please go ahead" to continue

After the fix, the model receives error feedback and can correct its tool names automatically.


DOsinga:
Thanks for the PR — this is a real and annoying bug worth fixing. A few concerns before this can land:

**Implementation bug: double-handling of errored tool calls**

The new block that yields specific error feedback for errored `remaining_requests` runs *before* the existing handler in the subsequent `for request in frontend_requests.iter().chain(remaining_requests.iter())` loop (around line 1474 in `agent.rs`). That existing handler also encounters the same errored requests and: (1) yields a second conflicting generic "could not be parsed" message, and (2) sets `exit_chat = true`, terminating the session. So rather than letting the model self-correct (which is the intent), both messages fire and the chat exits anyway. The fix needs to either skip errored requests in the existing loop once they've been handled, or replace that handler's behaviour entirely.

**Tests are testing the wrong thing**

The new tests in `reply_parts.rs` verify that `categorize_tool_requests` correctly routes errored calls into `remaining_requests` — which is true, but that's a pre-existing invariant, not the behaviour this PR adds. The thing that actually matters is: does the agent deliver the error feedback to the model and continue the conversation rather than exiting? That should be tested end-to-end using the existing integration test infrastructure in `crates/goose/tests/` (see e.g. `agents_test.rs`) with a mocked provider that returns a malformed tool call — checking that the next LLM turn receives a `ToolResponse` message containing the error text.

fresh3nough:
@reviewer Thanks for the detailed review -- spot on.

Plan:
- **Double-handling**: filtered errored requests out of `remaining_requests` before the main for-loop (via `valid_requests` partition), so the generic `else` handler with `exit_chat = true` never sees them.
- **Tests**: removed the reply_parts.rs unit tests (they only verified a pre-existing categorize_tool_requests invariant). Replaced with a proper e2e integration test in `tests/agent.rs` -- mocked provider returns an errored tool call (ErrorData), then verifies the error feedback message is delivered and the chat continues to a second provider call.

Pushed update.

Issue #7478, opened 2026-02-24
BUG: Schedules Added via CLI not showing up in Goose UI


amandacrawford-block:
When adding new goose schedules via CLI, the new schedules are not being displayed in the app UI. In order for me to view these schedules, I have to quit Goose and reopen the UI for them to display.

💡 Before filing, please check common issues:  
https://block.github.io/goose/docs/troubleshooting  

📦 To help us debug faster, attach your **diagnostics zip** if possible.  
👉 How to capture it: https://block.github.io/goose/docs/troubleshooting/diagnostics-and-reporting/

---

**To Reproduce**
1. Add new schedule using Goose CLI: `goose schedule add --schedule-id test_example --cron "0 10,14,16 * * 1-5" --recipe-source link-to-recipe.yaml
2. Check if schedule was added via CLI: `goose schedule list`
3. Open Goose UI -> Scheduler -> Check if new test_example schedule is displayed

---

**Expected behavior**
The test_example should be displayed in the list of schedules

---

**Screenshots**

---

**Please provide the following information**
- **OS & Arch:** Mac OS M4 Max Tahoe 26.3
- **Interface:** Both UI and CLI
- **Version:** Version 1.25.0-block.202602182139-cd798 (1.25.0-block.202602182139-cd798)
- **Extensions enabled:** A ton, assume the max
- **Provider & Model:** Goose-Databricks

---

**Additional context**

PR #7594, opened 2026-03-01
fix(scheduler): schedules added via CLI showing up in UI

Abhijay007:
closes https://github.com/block/goose/issues/7478

## Summary

- [x] Bug fix

### AI Assistance

### Testing
Tested by reproducing the bug

### DIFF
--- a/crates/goose/src/scheduler.rs
+++ b/crates/goose/src/scheduler.rs
@@ -462,7 +462,77 @@ impl Scheduler {
         }
     }
 
+    async fn sync_from_storage(&self) {
+        if !self.storage_path.exists() {
+            return;
+        }
+        let data = match fs::read_to_string(&self.storage_path) {
+            Ok(d) => d,
+            Err(_) => return,
+        };
+        if data.trim().is_empty() {
+            return;
+        }
+        let disk_jobs: Vec<ScheduledJob> = match serde_json::from_str(&data) {
+            Ok(jobs) => jobs,
+            Err(_) => return,
+        };
+
+        let disk_ids: std::collections::HashSet<String> =
+            disk_jobs.iter().map(|j| j.id.clone()).collect();
+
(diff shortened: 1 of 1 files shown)


jamadeo:
Thanks for the fix @Abhijay007 , but I wonder if maybe this is better done with a directory/file watcher in the goose server. That way the user doesn't have to refresh their list view for the scheduler itself to pick up a new item.

Abhijay007:
> Thanks for the fix @Abhijay007 , but I wonder if maybe this is better done with a directory/file watcher in the goose server. That way the user doesn't have to refresh their list view for the scheduler itself to pick up a new item.

It shows us after 15 sec without refresh or restarting, but yea let me look into possibility of doing it by file watcher in goose server.

jamadeo:
> > Thanks for the fix @Abhijay007 , but I wonder if maybe this is better done with a directory/file watcher in the goose server. That way the user doesn't have to refresh their list view for the scheduler itself to pick up a new item.
> 
> It shows us after 15 sec without refresh or restarting, but yea let me look into possibility of doing it bua file watcher in goose server.

Oh I see, so list_scheduled_jobs is called on an interval? that is probably fine then

Abhijay007:
> > > Thanks for the fix @Abhijay007 , but I wonder if maybe this is better done with a directory/file watcher in the goose server. That way the user doesn't have to refresh their list view for the scheduler itself to pick up a new item.
> > 
> > 
> > It shows us after 15 sec without refresh or restarting, but yea let me look into possibility of doing it bua file watcher in goose server.
> 
> Oh I see, so list_scheduled_jobs is called on an interval? that is probably fine then

Yes. The UI already polls GET /schedule/list every 15 seconds, so the server’s list_scheduled_jobs() runs on that interval. The change only makes that call sync from disk before returning, so new CLI-added schedules show up on the next poll without any refresh.

PR #7600, opened 2026-03-02
feat: claude sub first pass

michaelneale:
Seeing if this can work with a claude subscription


DOsinga:
do we not think that using the acp protocol for this sort of thing is strictly better /cc @codefromthecrypt ?

michaelneale:
yeah I am not sure I want to do this is against ToS - acp is right way

PR #7602, opened 2026-03-02
fix: scope empty session reuse to current window to prevent session mixing

### LINKED ISSUE #7601: Desktop: Sessions can get mixed up/overwritten during upgrade with multiple windows open
## Description

A user reported losing an entire chat session when upgrading from 1.25.0 to 1.26.1 with 4-5 chat windows open. Two sessions got "mixed up" - one was completely overwritten/lost, and the other received the name that should have belonged to the lost session.

## Steps to Reproduce

1. Open Goose Desktop with multiple chat windows/tabs (4-5)
2. Have at least two sessions that are still showing as "New Chat" (unnamed, no messages yet or just started)
3. Upgrade Goose (e.g., 1.25.0 → 1.26.1)
4. After restart, observe that sessions may be mixed up - one session's content may be lost and another may have the wrong name

## Expected Behavior

All sessions should be preserved with their correct content and names after an upgrade, regardless of how many windows are open.

## Actual Behavior

- One session was completely lost/overwritten
- Another session received the name that belonged to the lost session

## Root Cause Analysis

Investigation identified several contributing factors in the 1.25.0 → 1.26.1 changes:

### 1. In-memory session cache doesn't survive restarts
The `resultsCache` in `useChatStream.ts` caches `{messages, session}` by sessionId but is not persisted:
```typescript
```
When the app restarts during upgrade, this cache is cleared and sessions are reloaded from SQLite, but window-to-session associations may be lost.

### 2. "New Chat" session reuse logic
In `useNavigationSessions.ts`, the `handleNewChat` function reuses existing empty sessions:
```typescript
const emptyNewSession = sessionsRef.current.find((s) => shouldShowNewChatTitle(s));
if (emptyNewSession) {
  resumeSession(emptyNewSession, setView);
}
```
With multiple unnamed sessions across windows, this could route to the wrong session after a restart.

### 3. Async session naming race condition
Session names are generated asynchronously via `maybe_update_name()` which spawns a background task. If the app restarts during this process, names could be written to incorrect sessions.

### 4. Client/server sync changes (PR #7438)
The "client is not the source of truth" change removed `conversation_so_far` from client requests, changing how session state synchronizes. This was intended to fix bugs but may have edge cases during restarts.

## Related Issues

- #7402 - Desktop: Switching model in one tab incorrectly updates model display in all tabs (demonstrates cross-tab state isolation issues)

## Suggested Fixes

1. **Persist window-to-session mapping** - Store the `activeSessions` array to survive restarts, or use Electron's session restoration APIs

2. **Add session integrity verification** - When loading a session into a window, verify the content matches expectations (e.g., check message count, first message content)

3. **Scope "New Chat" reuse per-window** - The empty session reuse logic should be window-scoped, not global

4. **Add upgrade safeguards** - Consider gracefully closing/saving all sessions before an upgrade, or blocking upgrades when multiple windows have unsaved state

5. **Add session content hashing** - Store a hash of session content that can be verified after restart to detect corruption/mixing

6. **Automatic backup before upgrade** - Automatically export all open sessions to disk (e.g., as JSON/markdown files) before killing the app for an upgrade. This provides a recovery path even if something goes wrong during the upgrade process.

## Impact

**High** - Users can lose substantial work with no way to recover it. This is especially problematic for users who keep multiple long-running sessions open.

## Environment

- OS: macOS
- Versions affected: Upgrade from 1.25.0 to 1.26.1 (likely affects other upgrade paths too)
- Interface: Desktop (Electron)

## User Quote

> "The work on this app is amazing but please show care not to lose data with these super fast paced changes and upgrades...this has cost me a lot of time and could have been much worse had it been another chat."

fresh3nough:
## Problem

Fixes #7601

When multiple goose desktop windows are open during an upgrade or restart, all windows reload and call `handleNewChat`. The previous implementation used a global `sessions.find(s => shouldShowNewChatTitle(s))` which picked the **first** empty session from the shared session list. Every window would grab the same empty session, causing:

- Session data mixing/overwriting between windows
- Loss of conversation history
- Confusing UX where multiple windows show the same session

## Root Cause

In `AppSidebar.tsx`, `handleNewChat` (line 432) searched the entire `recentSessions` list for any empty session to reuse. Since all windows load the same session list from the API, they all race for the same first empty session.

## Fix

Changed `handleNewChat` to only reuse the current window's own active session (identified by `activeSessionId` from the URL search params) if that session is empty. If the window has no active session or the active session already has messages, a new session is always created.

This prevents cross-window collisions because each window has its own distinct `activeSessionId`.

## Changes

- **`ui/desktop/src/components/GooseSidebar/AppSidebar.tsx`**: Replaced global empty-session lookup with window-scoped active session check
- **`ui/desktop/src/__tests__/sessions.test.ts`**: Added 9 tests covering `shouldShowNewChatTitle` and the scoped reuse logic, including a test that demonstrates the old bug

## Testing

All 307 existing tests pass plus 9 new tests. ESLint clean.

### DIFF
--- a/ui/desktop/src/__tests__/sessions.test.ts
+++ b/ui/desktop/src/__tests__/sessions.test.ts
@@ -1,0 +1,108 @@
+import { describe, it, expect } from 'vitest';
+import { shouldShowNewChatTitle } from '../sessions';
+import type { Session } from '../api';
+
+// Helper to build a minimal Session object for testing.
+function makeSession(overrides: Partial<Session> = {}): Session {
+  return {
+    id: 'sess-1',
+    name: 'untitled',
+    message_count: 0,
+    created_at: new Date().toISOString(),
+    updated_at: new Date().toISOString(),
+    working_dir: '/tmp',
+    extension_data: { active: [], installed: [] },
+    ...overrides,
+  };
+}
+
+describe('shouldShowNewChatTitle', () => {
+  it('returns true for an empty session without a user-set name', () => {
+    const session = makeSession({ message_count: 0, user_set_name: false });
+    expect(shouldShowNewChatTitle(session)).toBe(true);
+  });
+
+  it('returns false when the session has messages', () => {
+    const session = makeSession({ message_count: 3, user_set_name: false });
+    expect(shouldShowNewChatTitle(session)).toBe(false);
+  });
+
+  it('returns false when the user has set a custom name', () => {
+    const session = makeSession({ message_count: 0, user_set_name: true });
+    expect(shouldShowNewChatTitle(session)).toBe(false);
+  });
+
+  it('returns false when the session has a recipe', () => {
+    const session = makeSession({
+      message_count: 0,
+      user_set_name: false,
+      recipe: { title: 'Recipe', steps: [] } as unknown as Session['recipe'],
+    });
+    expect(shouldShowNewChatTitle(session)).toBe(false);
+  });
+});
+
+describe('session reuse scoping (fix for #7601)', () => {
+  // Simulates the core logic extracted from handleNewChat in useNavigationSessions.ts.
+  // Before the fix: `sessions.find(s => shouldShowNewChatTitle(s))` picked the
+  // first global empty session regardless of which window called it.
+  // After the fix: only the current window's activeSessionId is considered.
+  function findReusableSession(
+    sessions: Session[],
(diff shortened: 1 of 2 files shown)


jh-block:
thanks!

fresh3nough:
np

PR #7604, opened 2026-03-02
feat: return structured {stdout, stderr} from shell tool with output schema

jh-block:
## Problem

When using code mode with local models, the model naturally writes `result.stdout` and `result.stderr` to access shell output. However, the shell tool was:

1. Merging stdout and stderr into a single string
2. Not declaring an output schema, so code mode generated `Promise<any>` as the return type
3. Describing itself as "return both stdout/stderr" which implied a structured object

This caused `result.stdout` and `result.stderr` to both be `undefined`.

## Changes

- **Structured output schema**: added `ShellOutput { stdout, stderr, exit_code, timed_out }` with `.with_output_schema::<ShellOutput>()` so `pctx_code_mode` generates typed TypeScript interfaces instead of `Promise<any>`
- **Structured content on all paths**: every return path (success, non-zero exit, timeout, parse error, spawn failure) attaches a `ShellOutput`-shaped `structured_content` so code-mode callers always get `{stdout, stderr, exit_code, timed_out}`
- **Interleaved display text**: output is collected as tagged lines preserving chronological arrival order across stdout/stderr for the human-facing text content
- **Raw empty strings in structured fields**: empty streams are `""` in structured content (the `"(no output)"` placeholder is only used in display text)
- **Truncation on structured fields**: stdout and stderr in `structured_content` are individually truncated via the same 2000-line/50KB limits, with overflow saved to per-stream temp files
- **Single-copy memory**: output is stored once as tagged lines; stdout, stderr, and interleaved strings are derived on demand without duplicate buffering
- **Updated description**: clarified that the tool returns an object with separate fields

## Testing

All 24 developer extension tests pass, including a new test for per-label overflow file separation.

### DIFF
--- a/crates/goose/src/scheduler.rs
+++ b/crates/goose/src/scheduler.rs
@@ -462,7 +462,77 @@ impl Scheduler {
         }
     }
 
+    async fn sync_from_storage(&self) {
+        if !self.storage_path.exists() {
+            return;
+        }
+        let data = match fs::read_to_string(&self.storage_path) {
+            Ok(d) => d,
+            Err(_) => return,
+        };
+        if data.trim().is_empty() {
+            return;
+        }
+        let disk_jobs: Vec<ScheduledJob> = match serde_json::from_str(&data) {
+            Ok(jobs) => jobs,
+            Err(_) => return,
+        };
+
+        let disk_ids: std::collections::HashSet<String> =
+            disk_jobs.iter().map(|j| j.id.clone()).collect();
+
+        let (jobs_to_add, jobs_to_remove): (Vec<ScheduledJob>, Vec<(String, JobId)>) = {
+            let jobs_guard = self.jobs.lock().await;
+            let to_add = disk_jobs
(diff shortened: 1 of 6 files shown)


crates/goose/src/agents/platform_extensions/developer/shell.rs
```diff
@@ -228,31 +281,58 @@ fn build_shell_command(command_line: &str) -> tokio::process::Command {
...
+        }
+        *started = true;
+        target.push_str(text);
+    }
+    (stdout, stderr, interleaved)
+}
+
+/// Collect lines from stdout and stderr in arrival order, tagging each with its source.
+/// Returns a vec of (is_stderr, line_text) preserving interleaved ordering.
+async fn collect_tagged_lines(
```
DOsinga, line 312: why not just return the three buffers instead of packaging it up like that and then unpacking it again?

DOsinga (approved):

PR #7606, opened 2026-03-02
fix: GooseMode changes applied to existing cached sessions

### LINKED ISSUE #7603: Bug: GooseMode changes not applied to existing cached sessions
## Summary

When a user changes the Goose mode (auto, approve, smart_approve, chat) via the GUI settings, existing cached agent sessions do not pick up the new mode. This can lead to unexpected behavior and significant cost overruns when users believe they've switched to a more restrictive mode but the session continues operating in the original mode.

## Root Cause

The `AgentManager` caches agents in an LRU cache (up to 100 sessions). When an agent is created, the `GooseMode` is read once from config and stored in `AgentConfig.goose_mode`. When the same session is accessed again, the cached agent is returned with its original mode - the config is never re-read.

**Relevant code in `crates/goose/src/execution/manager.rs`:**
```rust
pub async fn get_or_create_agent(&self, session_id: String) -> Result<Arc<Agent>> {
    // Returns cached agent with OLD mode
    if let Some(existing) = sessions.get(&session_id) {
        return Ok(Arc::clone(existing));
    }
    
    // Only reads mode when creating NEW agent
    let mode = Config::global().get_goose_mode().unwrap_or(GooseMode::Auto);
    let config = AgentConfig::new(..., mode, ...);
    // ...
}
```

**`AgentConfig` in `crates/goose/src/agents/agent.rs`:**
```rust
pub struct AgentConfig {
    pub goose_mode: GooseMode,  // Set once at creation, never updated
    // ...
}
```

## Steps to Reproduce

1. Start a new chat session in the GUI
2. Note the current mode (e.g., `auto`)
3. Change the mode to `smart_approve` or `approve` in settings
4. Continue using the **same** chat session
5. Observe that tools still execute without approval prompts (original `auto` behavior)

## Impact

- **Cost overruns**: Users who switch to restrictive modes expecting approval prompts may leave Goose running, resulting in unexpected API costs
- **Security/Safety**: Users may believe they have approval controls enabled when they don't
- **User trust**: Unexpected behavior erodes confidence in the mode system

## Real-world impact

A user reported switching to "smart" mode, leaving Goose running, and returning to find their project spend had gone from $4.50 to $70+, then to $184+ after reopening the same chat window (context load-in with the wrong mode).

## Suggested Fix

Option A: Re-read `goose_mode` from config at the start of each reply/request rather than caching it in the agent.

Option B: Invalidate/update cached agents when mode changes in config (would require config change notification mechanism).

Option C: When mode changes in GUI, restart/recreate agents for active sessions.

## Environment

- Goose version: v1.26.0+
- Affects: GUI (Desktop app)
- Config location: `~/.config/goose/config.yaml`

## Workaround

Users should:
1. Restart the Goose GUI app after changing mode settings
2. Start a new chat session after mode changes (don't continue existing sessions)

Abhijay007:
closes #7603 

### Type of Change
- [x] Bug fix

### AI Assistance

### Testing
testing rn 

Before:  

After:

### DIFF
--- a/crates/goose/src/execution/manager.rs
+++ b/crates/goose/src/execution/manager.rs
@@ -84,10 +84,12 @@ impl AgentManager {
     }
 
     pub async fn get_or_create_agent(&self, session_id: String) -> Result<Arc<Agent>> {
-        let mode = self.default_mode;
         {
             let mut sessions = self.sessions.write().await;
             if let Some(existing) = sessions.get(&session_id) {
+                let mode = Config::global()
+                    .get_goose_mode()
+                    .unwrap_or(self.default_mode);
                 existing.update_goose_mode(mode).await;
                 return Ok(Arc::clone(existing));
             }
@@ -97,7 +99,7 @@ impl AgentManager {
             Arc::clone(&self.session_manager),
             permission_manager,
             Some(Arc::clone(&self.scheduler)),
-            mode,
+            self.default_mode,
             Config::global()
                 .get_goose_disable_session_naming()
                 .unwrap_or(false),
@@ -132,6 +134,9 @@ impl AgentManager {
 
         let mut sessions = self.sessions.write().await;
         if let Some(existing) = sessions.get(&session_id) {
+            let mode = Config::global()
+                .get_goose_mode()
+                .unwrap_or(self.default_mode);
             existing.update_goose_mode(mode).await;
             Ok(Arc::clone(existing))
         } else {


crates/goose/src/agents/agent.rs
```diff
@@ -356,7 +356,9 @@ impl Agent {
             tools,
             toolshim_tools,
             system_prompt,
-            goose_mode: self.config.goose_mode,
+            goose_mode: Config::global()
+                .get_goose_mode()
+                .unwrap_or(self.config.goose_mode),
```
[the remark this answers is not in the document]
Abhijay007: Hmm it seems valid let me check that

crates/goose/src/execution/manager.rs
```diff
@@ -78,6 +78,8 @@ impl AgentManager {
         {
             let mut sessions = self.sessions.write().await;
             if let Some(existing) = sessions.get(&session_id) {
+                let mode = Config::global().get_goose_mode().unwrap_or(GooseMode::Auto);
+                existing.update_goose_mode(mode).await;
```
jamadeo, line 93: I suppose this is what we're doing now for loading an existing session, but we should probably also store this on the session. Then, to really be careful, only "downgrade" modes, but warn/ask before upgrading

jamadeo (approved):

crates/goose/src/execution/manager.rs
```diff
@@ -78,6 +78,8 @@ impl AgentManager {
         {
             let mut sessions = self.sessions.write().await;
             if let Some(existing) = sessions.get(&session_id) {
+                let mode = Config::global().get_goose_mode().unwrap_or(GooseMode::Auto);
```
codefromthecrypt: reading globals always makes me cringe.. is there another place we can get this value looking up the stack?

Abhijay007: maybe we can call Config::global() one level up the stack in `state.rs`, at the HTTP request boundary, and pass in as a parameter ? this is what you mean right 👀

crates/goose/src/execution/manager.rs
Abhijay007: The mode is already stored per-agent in `current_goose_mode: Mutex<GooseMode>` and updated on every request via `update_goose_mode()` , and regarding the the upgrade/downgrade guard, I thinnk we can held off on implementing that silently rn, since blocking a user's explicit settings change (e.g. Approve → Auto) without a UI prompt would be surprising ig.

maybe we can hanlde this as a follow-up PR once we have a proper confirmation flow in the UI flow ?

crates/goose/src/gateway/handler.rs
```diff
@@ -269,9 +269,12 @@ impl GatewayHandler {
             let _ = self.agent_manager.remove_session(session_id).await;
         }
 
+        let mode = crate::config::Config::global()
+            .get_goose_mode()
+            .unwrap_or(crate::config::GooseMode::Auto);
         let agent = self
             .agent_manager
```
codefromthecrypt, line 273: try to keep config based initialization of defaults in one place, as close. e.g. agentmanager is like this;

```
        AGENT_MANAGER
            .get_or_try_init(|| async {
                let max_sessions = Config::global()
                    .get_goose_max_active_agents()
                    .unwrap_or(DEFAULT_MAX_SESSION);
```

codefromthecrypt: so what I mean is to see if you can avoid a signature change for `get_or_create_agent` by having agent manager store a field for the default/first mode similar to how `AgentConfig` does

codefromthecrypt:
@Abhijay007 I think this is solving the issue per its suggestion, but the suggestion isn't really valid as global config can be not consistent with the session in play. This is especially the case in ACP where the mode is literally set and mutable in the session. We can't conflate the written config file with an arbitrary session, and even if we went to memory, it would be out-of-sync with the session (db) state. There are patterns where we bind config to the session, and I think that's the way forward. I'll draft a PR.

Sorry I didn't notice as I didn't look deeply at the problem and the solution, rather only the solution (here) and I think this was going the wrong path in hindsight. I made the summary comment about it here https://github.com/block/goose/issues/7603#issuecomment-4020554018

Abhijay007:
> @Abhijay007 I think this is solving the issue per its suggestion, but the suggestion isn't really valid as global config can be not consistent with the session in play. This is especially the case in ACP where the mode is literally set and mutable in the session. We can't conflate the written config file with an arbitrary session, and even if we went to memory, it would be out-of-sync with the session (db) state. There are patterns where we bind config to the session, and I think that's the way forward. I'll draft a PR.
> 
> Sorry I didn't notice as I didn't look deeply at the problem and the solution, rather only the solution (here) and I think this was going the wrong path in hindsight. I made the summary comment about it here [#7603 (comment)](https://github.com/block/goose/issues/7603#issuecomment-4020554018)

Oh no problem, and thanks for the detailed review, will wait for you PR on the same and will continue it from there :)

codefromthecrypt:
@Abhijay007 thanks.. and yeah doing the whole thing is a massive amount of work, vs the workaround. I'll take you up on the offer to review/iterate on the PR as often the velocity in goose is bounded by reviewal more than anything else.

DOsinga:
I think for now the workaround could be to just remove goose mode from the sessionconfig/agent and just always read it from the Config::global. that would not be perfect but would at least work predictable

codefromthecrypt:
https://github.com/block/goose/pull/7801 takes over and fixes ACP too

DOsinga:
closing this in favor of #7801

PR #7623, opened 2026-03-03
fix: disable all platform extensions by default except for developer

alexhancock:
Anecdotally I find many people are using goose only with `developer` enabled and it works well

Putting this up for consideration to make that the only platform extension enabled by default

### DIFF
--- a/crates/goose/src/agents/platform_extensions/mod.rs
+++ b/crates/goose/src/agents/platform_extensions/mod.rs
@@ -34,7 +34,7 @@ pub static PLATFORM_EXTENSIONS: Lazy<HashMap<&'static str, PlatformExtensionDef>
                 display_name: "Analyze",
                 description:
                     "Analyze code structure with tree-sitter: directory overviews, file details, symbol call graphs",
-                default_enabled: true,
+                default_enabled: false,
                 unprefixed_tools: true,
                 client_factory: |ctx| Box::new(analyze::AnalyzeClient::new(ctx).unwrap()),
             },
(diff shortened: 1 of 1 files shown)


tlongwell-block:
We probably shouldn’t disable summon by default since skills depend on it

tlongwell-block:
And I would leave analyze on too because goose flails in monorepos without it or a local rg

we probably don’t need todo with opus and codex anymore. But todo is a huge boost for weaker models

alexhancock:
> We probably shouldn’t disable summon by default since skills depend on it

on it. i think we should make skills part of the core of goose.

tlongwell-block:
It feels like core at this point! :) It’s a platform extension and the skill tool is not prefixed. It’s as built in feeling for the agent as possible

angiejones:
Todo is pretty helpful and "checklists" are common in other harnesses by default, so it's become expected behavior

alexhancock:
@tlongwell-block @angiejones 

I agree about TODOs! For that and for Skill loading I think should it should just be part of `goose` and not even be a platform extension. I will make a PR tomorrow to this effect for consideration. And then I'll likely update this PR to have: 

* developer
* computer controller
* extension manager

as the only default-enabled extensions (as TODO and Skill support will be part of native goose)

everything else I personally think should be at people's discretion to opt-in

tlongwell-block:
We spent a good bit of time moving tools *out* of being directly integrated with the agent in order to move them to platform extensions to make tooling easier to reason about. I don't know if integrating them more tightly improves the AX, but it will make it a bit harder to reason about/manage them on our side imo.

@DOsinga, I know you had thinking about integrated tooling moving into platform extensions?

alexhancock:
I have a new idea about this I will circulate soon instead of this approach

PR #7626, opened 2026-03-03
Improve the formatting of tool calls, show thinking, treat Reasoning and Thinking as the same thing (sorry Kant)

jh-block:
- Trim whitespace from tool parameter values and output text before rendering in code blocks in the expandable tool details
- Simplify ToolCallArguments with a unified `formatValue` helper, render expanded values as `<pre>` instead of MarkdownContent
- Simplify reasoning/thinking content display from collapsible `<details>` to inline italic text
- Remove `MessageContent::Reasoning` variant — `MessageContent::reasoning()` now produces `Thinking` content, so the UI only ever sees `thinking` type
- Old sessions with serialized `reasoning` content are migrated to `thinking` during deserialization
- Remove `ReasoningContent` struct, `as_reasoning()`, and all `Reasoning` match arms across provider formats, agent, context management, CLI output, and OpenAPI schema
- UI renamed `getReasoningContent` → `getThinkingContent`, only handles `thinking` content type


crates/goose/src/providers/formats/anthropic.rs
```diff
@@ -965,6 +963,21 @@ mod tests {
...
+            Message::assistant().with_text("Hi there"),
+        ];
+
+        let spec = format_messages(&messages);
+
+        assert_eq!(spec.len(), 1);
+        assert_eq!(spec[0]["role"], "assistant");
+        assert_eq!(spec[0]["content"][0]["type"], "text");
+        assert_eq!(spec[0]["content"][0]["text"], "Hi there");
+    }
```
DOsinga, line 956: I would drop this test. or just reduce to the check that it is one message. but probably drop

ui/desktop/src/components/ToolCallArguments.tsx
```diff
@@ -14,6 +13,13 @@ interface ToolCallArgumentsProps {
   args: Record<string, ToolCallArgumentValue>;
 }
 
+function formatValue(value: ToolCallArgumentValue): string {
+  if (typeof value === 'string') return value;
+  if (Array.isArray(value)) return JSON.stringify(value, null, 2);
```
DOsinga: The Array.isArray branch is redundant. arrays are objects, and both cases do the same thing

PR #7628, opened 2026-03-03
Add unified thinking effort control across all providers

jh-block:
## Summary

Replaces the fragmented per-provider thinking controls with a single unified **Thinking Effort** setting (`off`/`low`/`medium`/`high`/`max`) that works across all providers.

### Backend (crates/goose)
- **model.rs**: New `ThinkingEffort` enum with `FromStr`/`Display`/`Serialize`/`Deserialize`. Read from `request_params["thinking_effort"]` or `GOOSE_THINKING_EFFORT` env var via `ModelConfig::thinking_effort()`.
- **model.rs**: `normalize_effort_suffix()` automatically migrates legacy model names like `o3-mini-high` — strips the suffix and sets `thinking_effort` in `request_params` at construction time. The suffix is always stripped from the model name to avoid sending invalid identifiers to providers, while explicit `thinking_effort` in request_params takes priority over the suffix-derived value.
- **model.rs**: `with_request_params()` now merges incoming params into existing ones instead of replacing them, preserving suffix-derived `thinking_effort` when the server applies caller-provided request params.
- **OpenAI format**: Maps effort → `reasoning_effort` (`low`/`medium`/`high`), defaults to `medium`.
- **Anthropic format**: Maps effort → `ThinkingType` (adaptive for 4-6 models, enabled for others) and Anthropic-native effort levels. Maps to budget tokens for enabled mode.
- **Databricks format**: Maps to `reasoning_effort` for OpenAI models and `budget_tokens` for Claude models.
- **Google format**: Maps to `thinking_level` (`low`/`high`) for Gemini 3 models.
- **config/base.rs**: Replaced 4 legacy config values with single `GOOSE_THINKING_EFFORT`.
- **CLI configure**: Simplified from model-specific multi-step thinking config to a single effort dropdown. Thinking support detection uses fallible `ModelConfig::new()` instead of `new_or_fail()` to avoid panicking on invalid env/config values.

### Backward Compatibility
- Model names with effort suffixes (e.g., `o3-mini-high`) are automatically migrated: the suffix is always stripped from the model name, and converted to a `thinking_effort` request param unless one is already explicitly set.

### Removed Legacy Config
- `CLAUDE_THINKING_TYPE`, `CLAUDE_THINKING_EFFORT`, `CLAUDE_THINKING_BUDGET`, `CLAUDE_THINKING_ENABLED`
- `GEMINI3_THINKING_LEVEL`

### UI (ui/desktop)
- Replaces three model-specific controls (Gemini thinking level, Claude thinking type/effort/budget) with a single **Thinking Effort** dropdown
- Shown for any model that supports thinking (Claude, OpenAI reasoning models, Gemini 3)
- Persisted via `GOOSE_THINKING_EFFORT` config key

## Testing
- 8 unit tests for `ThinkingEffort` parsing, env var, request_params override, suffix migration
- All existing provider tests updated and passing (878 total in goose crate)
- All 340 UI tests pass

### DIFF
--- a/crates/goose/src/providers/formats/anthropic.rs
+++ b/crates/goose/src/providers/formats/anthropic.rs
@@ -52,6 +52,16 @@ pub fn thinking_type(model_config: &ModelConfig) -> ThinkingType {
 
     let is_adaptive_model = supports_adaptive_thinking(&model_config.model_name);
 
+    // Unified thinking effort takes priority
+    if let Some(effort) = model_config.thinking_effort() {
+        use crate::model::ThinkingEffort;
+        return match effort {
+            ThinkingEffort::Off => ThinkingType::Disabled,
+            _ if is_adaptive_model => ThinkingType::Adaptive,
+            _ => ThinkingType::Enabled,
+        };
+    }
+
     if let Some(s) =
         model_config.get_config_param::<String>("thinking_type", "CLAUDE_THINKING_TYPE")
     {
@@ -469,6 +479,17 @@ pub fn get_usage(data: &Value) -> Result<Usage> {
 }
 
 pub fn thinking_effort(model_config: &ModelConfig) -> ThinkingEffort {
+    // Unified thinking effort takes priority
+    if let Some(effort) = model_config.thinking_effort() {
+        return match effort {
+            crate::model::ThinkingEffort::Off => ThinkingEffort::Low,
+            crate::model::ThinkingEffort::Low => ThinkingEffort::Low,
+            crate::model::ThinkingEffort::Medium => ThinkingEffort::Medium,
+            crate::model::ThinkingEffort::High => ThinkingEffort::High,
+            crate::model::ThinkingEffort::Max => ThinkingEffort::Max,
+        };
+    }
+
     match model_config.get_config_param::<String>("effort", "CLAUDE_THINKING_EFFORT") {
         Some(s) => s.parse().unwrap_or_else(|e| {
             tracing::warn!("{e}, defaulting to 'high'");
--- a/crates/goose/src/providers/formats/anthropic.rs
+++ b/crates/goose/src/providers/formats/anthropic.rs
@@ -51,43 +51,14 @@ pub fn thinking_type(model_config: &ModelConfig) -> ThinkingType {
     }
 
     let is_adaptive_model = supports_adaptive_thinking(&model_config.model_name);
+    let effort = model_config
+        .thinking_effort()
+        .unwrap_or(crate::model::ThinkingEffort::Off);
 
-    // Unified thinking effort takes priority
-    if let Some(effort) = model_config.thinking_effort() {
-        use crate::model::ThinkingEffort;
-        return match effort {
-            ThinkingEffort::Off => ThinkingType::Disabled,
-            _ if is_adaptive_model => ThinkingType::Adaptive,
-            _ => ThinkingType::Enabled,
-        };
-    }
-
-    if let Some(s) =
-        model_config.get_config_param::<String>("thinking_type", "CLAUDE_THINKING_TYPE")
-    {
-        let tt = s.parse::<ThinkingType>().unwrap_or_else(|e| {
-            tracing::warn!("{e}");
-            ThinkingType::Disabled
-        });
-        if tt == ThinkingType::Adaptive && !is_adaptive_model {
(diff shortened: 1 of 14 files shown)


documentation/docs/guides/cli-providers.md
```diff
@@ -310,7 +310,7 @@ The following models are recognized and passed to the Claude CLI via the `--mode
 | `GOOSE_PROVIDER` | Set to `codex` to use this provider | None |
 | `GOOSE_MODEL` | Model to use (only known models are passed to CLI) | `gpt-5.2-codex` |
 | `CODEX_COMMAND` | Path to the Codex CLI command | `codex` |
-| `CODEX_REASONING_EFFORT` | Reasoning effort level: `low`, `medium`, `high`, or `xhigh` (`none` is only supported on non-codex models like `gpt-5.2`) | `high` |
```
blackgirlbytes: can you take the docs out or put them in a separate PR since the docs get deployed immediately? And the feature wont be deployed immediately?

I'm creating a solution so folks wont have to do this in the future, but that's the workaround for now. Thank you!

crates/goose/src/providers/formats/anthropic.rs
```diff
@@ -457,12 +438,16 @@ pub fn get_usage(data: &Value) -> Result<Usage> {
...
+    let effort = model_config
+        .thinking_effort()
+        .unwrap_or(crate::model::ThinkingEffort::High);
+    match effort {
+        crate::model::ThinkingEffort::Off | crate::model::ThinkingEffort::Low => {
+            ThinkingEffort::Low
+        }
+        crate::model::ThinkingEffort::Medium => ThinkingEffort::Medium,
+        crate::model::ThinkingEffort::High => ThinkingEffort::High,
+        crate::model::ThinkingEffort::Max => ThinkingEffort::Max,
```
DOsinga: this looks distinctly odd. I would suggest to drop the local ThinkingEffort enum, we now have the Display/from string on it,

crates/goose/src/providers/formats/anthropic.rs
```diff
@@ -1095,20 +1080,15 @@ mod tests {
         let payload = create_request(&config, "system", &messages, &[])?;
 
         assert_eq!(payload["thinking"]["type"], "enabled");
-        assert_eq!(payload["thinking"]["budget_tokens"], 10000);
-        assert_eq!(payload["max_tokens"], 4096 + 10000);
+        assert_eq!(payload["thinking"]["budget_tokens"], 16000);
+        assert_eq!(payload["max_tokens"], 4096 + 16000);
```
DOsinga: this test is not doing much given the size. consider using cfg_with_effort and then asserting that budget_tokens > 0 and max_tokens = 4096 + budget_tokens

ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
```diff
@@ -21,27 +21,54 @@ import { getPredefinedModelsFromEnv, shouldShowPredefinedModels } from '../prede
...
-  { value: 'low', label: 'Low - Minimal thinking, fastest responses' },
   { value: 'medium', label: 'Medium - Moderate thinking' },
-  { value: 'high', label: 'High - Deep reasoning (default)' },
+  { value: 'high', label: 'High - Deep reasoning' },
   { value: 'max', label: 'Max - No constraints on thinking depth' },
 ];
 
 function isClaudeModel(name: string | null | undefined): boolean {
-  return !!name && name.toLowerCase().startsWith('claude-');
+  return !!name && name.toLowerCase().includes('claude');
```
DOsinga: !!name?.

DOsinga (approved):
the client side thinking support handling could really do with a clean up

DOsinga (approved):

ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
jh-block: double-bang is a fairly common JS thing, the first ! converts it to a boolean (so checks if it's falsey) and then the second one negates that (so checks if it's truthy)

checking if the type is string might be more clear though

blackgirlbytes (approved):

jh-block:
Superseded by #9242

PR #7630, opened 2026-03-03
Merge platform/builtin extensions

jamadeo:
The distinction between builtin and platform need not be in the config file -- goose knows what's what. Right now, it only leads to bugs.


jamadeo:
~~This doesn't quite work as we want it to (yet) - the platform extension migration will change builtins to platform on load~~

This should work. There are at least two places where we'd change from builtin to platform, and with this change, we don't need to do that.

alexhancock (approved):

PR #7631, opened 2026-03-03
feat(extensions): add Unix socket transport for StreamableHttp extensions

wpfleger96:
Adds optional Unix domain socket transport for `StreamableHttp` MCP extensions, enabling Goose to route HTTP traffic through a local Unix socket rather than direct TCP.

Some deployment environments (e.g., Kubernetes pods with a service mesh sidecar) require outbound HTTP to pass through a proxy that is only accessible via a Unix domain socket — DNS-based hostnames do not resolve over the normal network stack in these contexts. Without this, `StreamableHttp` extensions targeting service-mesh-only endpoints are unreachable.

- Adds optional `socket` field to `ExtensionConfig::StreamableHttp` (config file and goosed API; no CLI flag — this is an infrastructure concern, not a user-facing setting)
- Implements `StreamableHttpClient` trait via `UnixSocketHttpClient` using `hyper` + `tokio::net::UnixStream`; plugged in through `StreamableHttpClientTransport::with_client()` — the reqwest path is unchanged
- Supports both filesystem sockets (`/var/run/proxy.sock`) and Linux abstract sockets (`@egress.sock` → null-byte prefix)
- Non-unix platforms return a clear config error when `socket` is set; all new deps are `cfg(unix)`-gated


alexhancock:
@wpfleger96 this looks cool!

What would you think about porting it to https://github.com/modelcontextprotocol/rust-sdk as an optional feature, to allow any client that needs to do streamable http over UDS to do it with the sdk?

This would get it working for goose, but also any other MCP host that uses rmcp (codex etc)

wpfleger96:
@alexhancock great idea, that didn't even cross my mind. I'm down to give it a shot! could I ping you there for a review when I'm ready?

alexhancock:
> @alexhancock great idea, that didn't even cross my mind. I'm down to give it a shot! could I ping you there for a review when I'm ready?

I am heading on vacation for a week tomorrow, so ping @jamadeo and @daleseo

crates/goose/src/agents/extension_manager.rs
```diff
@@ -483,6 +513,97 @@ async fn create_streamable_http_client(
...
+        default_headers.insert(header_name, val);
+    }
+
+    let retry_headers = {
+        let mut h = default_headers.clone();
+        h.remove(&http::header::AUTHORIZATION);
+        h
+    };
+
+    let unix_client = UnixSocketHttpClient::new(uri, socket_path, default_headers);
```
jamadeo, line 562: would something like

```
let unix_client = reqwest::Client::builder()
    .unix_socket(socket_path)
    .build()?;
let transport = StreamableHttpClientTransport::with_client(
    unix_client,
    StreamableHttpClientTransportConfig {
        uri: uri.into(),
        ..Default::default()
    },
);
```

have worked here?

michaelneale:
I like the idea of this, if we can bake it a bit more as discussed above. Unrelated but I always wanted the chromium front end to talk socket transport but at the time I couldn't make it work. An under appreciated tool.

wpfleger96:
Superseded by #7860, which uses the upstream rmcp Unix domain socket client (rmcp 1.3.0) instead of the local implementation.

Issue #7410, opened 2026-02-21
bug(windows): MSVC link failure in goosed when v8 and llama-cpp are linked together (LNK2038/LNK2005/LNK1169)


bavadim:
Windows MSVC desktop builds fail at link stage when `goosed` links both `v8/deno_core` and `llama-cpp-sys-2` in the same binary. Observed failures include CRT/runtime-library mismatch and duplicate C++ exception symbol definitions.

This appears to be a regression after local inference (`llama.cpp`) integration (commit `ddd35f6d4`, 2026-02-19).

**To Reproduce**
  
  1. Use Windows MSVC environment (`windows-latest` / VS2022).
  2. Ensure build includes both dependency groups:
     - `deno_core` / `v8`
     - `llama-cpp-2` / `llama-cpp-sys-2`
  3. Run:
     - `rustup target add x86_64-pc-windows-msvc`
     - `cargo clean`
     - `cargo build --release -p goose-server --target x86_64-pc-windows-msvc --verbose`
  4. Observe linker failure near final `goosed.exe` link step.

CI reproduction:
  1. Trigger desktop Windows workflow.
  2. Open Windows build job (`Build goosed (Windows)`).
  3. See linker signatures below.

**Expected behavior**
  
Windows MSVC build should complete successfully without runtime-library mismatch or duplicate symbol linker errors.

  **Screenshots**

  N/A (CI run/job links and log excerpt below).

  **System information**
  
  - **OS & Arch:** Windows Server 2025 x64 (`windows-latest`)
  - **Interface:** CI / desktop workflow
  - **Version:** 1.24.0 line
  - **Extensions enabled:** N/A
  - **Provider & Model:** N/A

  **Additional context**

  Relevant log excerpt:

  error: linking with `link.exe` failed: exit code: 1169

  libv8-...rlib(exception.obj) : error LNK2005:
  std::exception_ptr::exception_ptr(...) already defined in
  libllama_cpp_sys_2-...rlib(llama-model-loader.obj)

  libv8-...rlib(exception.obj) : error LNK2005:
  std::exception_ptr::~exception_ptr(void) already defined in
  libllama_cpp_sys_2-...rlib(llama-model-loader.obj)

  libv8-...rlib(exception.obj) : error LNK2005:
  std::current_exception(void) already defined in
  libllama_cpp_sys_2-...rlib(llama-model-loader.obj)

  libv8-...rlib(exception.obj) : error LNK2005:
[...]

PR #7633, opened 2026-03-04
fix: prevent abort in local inference 

KubeCat:
Fix a bug "Rust cannot catch foreign exceptions" abort that occurs when using Local Inference with models that have long Jinja2 chat templates (e.g. Qwen3 GGUF models).

### The problem

1.  `apply_chat_template()` calls `llama_chat_apply_template()` which internally does a C++ map `.at()` lookup on the template string. 
2. When the string is a full Jinja2 template (not a short name like "chatml"), this throws `std::out_of_range`. 
3. The exception is normally caught in C++, but when goose's other native dependencies are linked into the binary, the C++ exception-handling ABI breaks and the exception propagates across the Rust FFI boundary, causing a process abort.

### The solution

Switch from `apply_chat_template()` to  `apply_chat_template_with_tools_oaicompat()` in `inference_emulated_tools.rs`

### Why?

This function has its own C++ try-catch wrapper. When called with `tools=None`, it produces the same prompt. This aligns the emulated tools path with the native tools path, which already uses the oaicompat variant.
 

### Testing

A few integration tests were improved, the integration tests for the local inference had been abandoned and marked with `#[ignore]`, this is problematic because I could not understand very clearly how to actually test my changes. The changes in testing are as follows:

1. Fix broken model ID in local_inference_integration.rs, add TEST_MODEL env var override, removed perf assertion since it would never pass because the model was never truly in a cold start state.
2. Separate cold vs warm perf benchmark with TEST_MODEL env var support in `local_inference_perf.rs`

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality 
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [x] Build / Release
- [ ] Other (specify below)

### AI Assistance
- 56 unit tests pass (parsing, engine, hf_models)
- Integration tests pass with Llama 1B and Qwen3 32B, on both CPU and CUDA
- Qwen3 32B was the crashing model now works

 ### Testing
 - Tested end-to-end with `bartowski/Qwen_Qwen3-32B-GGUF:Q4_K_M` via `test_provider_configuration` —
 previously aborted, now passes.
 - Verified the crash is specific to the goose binary (standalone llama-cpp-2 tests pass because the
 C++ ABI isn't disturbed by additional native deps).
 - Verified that short template names ("chatml") never triggered the crash — only full Jinja2
 template strings.
 - Existing unit tests for emulated tools parsing, tool parsing, and inference engine all continue to
  pass (they test downstream of the template step).

 ### Related Issues

- Related to #7410 same root cause (C++ exception-handling ABI clash between llama-cpp and other native deps). That issue manifests as a Windows MSVC link failure; this one manifests as a Linux runtime abort.
- Follow-up to #7511 which fixed the Windows MSVC link-time side of the v8/llama-cpp C++ ABI conflict (`/FORCE:MULTIPLE` for duplicate `std::exception_ptr` symbols). This PR addresses the Linux runtime side where the exception *links* but doesn't *propagate* correctly across the Rust FFI boundary.
- Discovered an issue where `CUDA` will crash when context_size is null in registry which falls through to n_ctx_train, despite the fact that estimate_max_context_for_memory should cap it.

image

After: Model loads and generates normally.
image

With `CUDA`: 
Screenshot_20260304_023801

### DIFF
--- a/crates/goose-cli/Cargo.toml
+++ b/crates/goose-cli/Cargo.toml
@@ -72,6 +72,7 @@ winapi = { version = "0.3", features = ["wincred"] }
 [features]
 default = ["code-mode"]
 code-mode = ["goose/code-mode", "goose-acp/code-mode"]
+cuda = ["goose/cuda"]
 # disables the update command
 disable-update = []
 
--- a/crates/goose/src/providers/local_inference/inference_emulated_tools.rs
+++ b/crates/goose/src/providers/local_inference/inference_emulated_tools.rs
@@ -359,10 +359,19 @@ pub(super) fn generate_with_emulated_tools(
     ctx: &mut GenerationContext<'_>,
     code_mode_enabled: bool,
 ) -> Result<(), ProviderError> {
+    // Use oaicompat variant — its C++ wrapper catches exceptions that would
+    // otherwise abort the process when other native libs disturb the C++ ABI.
     let prompt = ctx
         .loaded
         .model
-        .apply_chat_template(&ctx.loaded.template, ctx.chat_messages, true)
+        .apply_chat_template_with_tools_oaicompat(
+            &ctx.loaded.template,
+            ctx.chat_messages,
+            None, // no tools for emulated path
+            None, // no json_schema
+            true, // add_generation_prompt
+        )
+        .map(|r| r.prompt)
         .map_err(|e| {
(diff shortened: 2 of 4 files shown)


jh-block (approved):
Thank you for this!

PR #7643, opened 2026-03-04
Integrate Sentry error tracking across Electron and Rust

jh-block:
## Summary

Adds Sentry error tracking across the full stack — Electron desktop app and Rust backend — with the DSN injected at build time so it never appears in source code.

### Sentry integration (Electron + Rust)
- **Main process**: `@sentry/electron/main` initialized at app startup
- **Preload**: `@sentry/electron/preload` imported for IPC bridge between main and renderer
- **Renderer**: `@sentry/electron/renderer` initialized for frontend error capture
- **Rust**: `sentry` crate with `tower`, `tower-http`, and `tracing` features; `NewSentryLayer` and `SentryHttpLayer` added to axum for distributed tracing

### Telemetry consent gating
- Sentry events are gated on user telemetry consent via `beforeSend`/`beforeSendTransaction` hooks (Electron) and `is_telemetry_enabled()` check (Rust)
- Users must opt in to telemetry before any Sentry data is sent

### Build-time DSN injection
The Sentry DSN is **not** hard-coded in source. Instead, it's injected at build time via the `SENTRY_DSN` environment variable:
- **Rust**: `option_env!("SENTRY_DSN")` reads the env var at compile time; if unset, Sentry is fully disabled
- **Electron**: `process.env.SENTRY_DSN` injected via Vite `define` block; if empty, Sentry SDK initializes in disabled mode
- **CI/CD**: `SENTRY_DSN` passed as a GitHub Actions secret through release and canary workflows to all bundle-desktop reusable workflows
- **Cross-compilation**: `SENTRY_DSN` added to `Cross.toml` passthrough list for Linux cross-compiled builds

This means:
| Build scenario | Sentry enabled? |
|---|---|
| Official release/canary build (CI has `SENTRY_DSN` secret) | Yes |
| Clone + local build (no env var set) | No |
| Clone + own DSN (`SENTRY_DSN=... cargo build`) | Yes, with their DSN |

### Environment tagging
- Electron passes `GOOSE_ENVIRONMENT` env var when spawning goosed (`production` for packaged app, `development` otherwise)
- Rust reads `GOOSE_ENVIRONMENT`, defaulting to `development` if unset

### Setup required
After merging, add a repository secret named `SENTRY_DSN` with the DSN value.

## Test plan
- [ ] Verify `cargo check -p goose-server` compiles without `SENTRY_DSN` set (no DSN baked in)
- [ ] Verify `SENTRY_DSN=test cargo check -p goose-server` compiles with DSN baked in
- [ ] Verify Electron app starts without errors when `SENTRY_DSN` is not set
- [ ] After adding the repo secret, verify a canary build produces binaries with Sentry enabled

### DIFF
--- a/crates/goose/src/posthog.rs
+++ b/crates/goose/src/posthog.rs
@@ -59,7 +59,9 @@ pub fn is_telemetry_enabled() -> bool {
 /// changed so that Sentry callbacks and other hot paths see the new
 /// state immediately.
 pub fn set_telemetry_enabled(enabled: bool) {
-    TELEMETRY_ENABLED.store(enabled, Ordering::Relaxed);
+    // GOOSE_TELEMETRY_OFF is a hard kill switch; never allow re-enabling.
+    let effective = enabled && !TELEMETRY_DISABLED_BY_ENV.load(Ordering::Relaxed);
+    TELEMETRY_ENABLED.store(effective, Ordering::Relaxed);
 }
 
 // ============================================================================


lifeizhou-ap:
Hi @jh-block,

The PR looks good!

Just want to confirm Sentry is only for error tracking here, not product analytics (that stays in PostHog)?

A few comments:

- `posthog.rs` now has shared consent logic used by both PostHog and Sentry. Maybe rename to telemetry.rs? 

-  Shall we also apply sentry in the CLI?  maybe a followup PR?

- emit_error and emit_error_with_context are already disabled in . Shall we move them to use Sentry (sentry::capture_message()) instead? some of them are in the goose core, maybe we have to add sentry in goose.  Or we just remove them.

PR #7646, opened 2026-03-04
refactor: big bang goose-mcp -> platform extensions

alexhancock:
A big bang move of all `goose-mcp` extensions to the platform extension model. The PR makes a couple other changes as well:

1. The only default enabled things are now `computer_controller` and `developer`. Should add `extension_manager` to this as well
2. Removed the concept of `bundled` as a config attr for extensions. This will likely necessitate changes to distros which use the JSON file to extend the default included platform extensions


alexhancock:
will do another rev of this atop the new changes

Issue #7636, opened 2026-03-04
Switch Models dialog blocks provider selection when model listing API fails (e.g., Anthropic via Azure AI Foundry)


blackgirlbytes:
## Bug Description

When the dynamic model listing API call fails for a configured provider, the Switch Models dialog shows a dead-end error banner with no way to proceed. The user cannot select a model, enter a custom model name, or switch to that provider at all. Combined with the restriction that the currently active provider cannot be deleted, this can leave users completely stuck — unable to switch back to a previously working provider.

## Steps to Reproduce

1. Configure Anthropic (e.g., via Azure AI Foundry where `ANTHROPIC_HOST` points to an Azure endpoint)
2. Use Anthropic for chat — works fine (`v1/messages` is supported)
3. Add and switch to another provider (e.g., ChatGPT Codex, Azure OpenAI, or Gemini)
4. Open the Switch Models dialog and select Anthropic from the provider dropdown
5. **Result:** Error banner "Could not contact provider" with no model selector or custom model input

The user is now stuck:
- Cannot switch to Anthropic (model listing fails → dead-end error)
- Cannot delete the current provider (it is "currently in use")
- Only workaround is manually editing `~/.config/goose/config.yaml`

<img width="1333" height="1303" alt="Image" src="https://github.com/user-attachments/assets/eb274277-9409-4ebe-ad22-bf9d996b2958" />

## Context: How we got here

PR #6744 ("Show provider failures") intentionally removed a `known_models` fallback that previously existed. The old code in `modelInterface.ts` was:

```typescript
let models = await getProviderModelsFunc(providerName);
if ((!models || models.length === 0) && p.metadata.known_models?.length) {
    models = p.metadata.known_models.map((m) => m.name);
}
```

The old `getProviderModels` wrapper in `ConfigContext.tsx` silently swallowed API errors and returned `[]`, which then triggered the `known_models` fallback. For providers like Ollama, this was confusing — users saw hardcoded default models that had nothing to do with what was actually available.

PR #6744 correctly fixed the silent error swallowing by using `throwOnError: true` and surfacing the real error. But it also removed the `known_models` fallback entirely and replaced it with a dead-end error banner that blocks the user completely. The fix went from one extreme (silently show wrong models) to the other (block the user entirely).

### Dead-end error UI with no escape hatch (`SwitchModelModal.tsx`)

[...]

PR #7649, opened 2026-03-04
fix: restore known_models fallback for all providers on model listing failure

Abhijay007:
closes  https://github.com/block/goose/issues/7636

### Type of Change

- [ ] Feature
- [x] Bug fix

### AI Assistance

Before:  

After:

### DIFF
--- a/ui/desktop/src/components/settings/models/modelInterface.ts
+++ b/ui/desktop/src/components/settings/models/modelInterface.ts
@@ -72,7 +72,8 @@ export async function fetchModelsForProviders(
       const models = response.data || [];
       return { provider: p, models, error: null, warning: null };
     } catch (e: unknown) {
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
note: the issue that raised for this is a 404, which is a misconfiguration. we should identify what the actual issue is, if it is an openai endpoint that does not include the models endpoint (emulation) that's config not something to paper over. 

we already allow users to hand-enter models

cc @DOsinga as I think this is basically re-introducing the bug where we subverted our health check, and if you look closely at the issue this is raised against, it is a 404

DOsinga:
yes @codefromthecrypt I agree. we use the models as a health check. there's a different PR that implements this only for custom providers and on the server. so I am going to close this

Issue #7570, opened 2026-02-27
docs: document summon extension requirement for delegate and load tools


clouatre:
**Please explain the motivation behind the feature request.**

PR #6964 (v1.24.0) moved `delegate` and `load` from unconditional builtins into the `summon` platform extension. Recipes with an explicit `extensions:` block that omit `summon` silently lose access to both tools. There is no error, no warning, and no log message.

This requirement is not documented anywhere. The subagents guide, recipe reference, subrecipes page, and tutorials all describe `delegate` without mentioning that it depends on the `summon` extension.

Related:
- #6964 - refactor that introduced the breaking change
- #7355 - regression report (closed, workaround: add `summon` to extensions)
- #7360 - auto-injection fix (closed, maintainer prefers explicit declaration)
- #7457 - separate regression from #6964 (merged)

**Describe the solution you'd like**

Document the `summon` extension requirement in the pages that reference `delegate` or `load`:

1. **[Subagents guide](https://block.github.io/goose/docs/guides/subagents)** - add a prerequisites note that `delegate` is provided by the `summon` extension and must be declared in recipes with an explicit `extensions:` block.
2. **[Recipe reference](https://block.github.io/goose/docs/guides/recipes/recipe-reference)** - note under the `extensions` field that `summon` is required for `delegate` and `load`.
3. **[Subrecipes guide](https://block.github.io/goose/docs/guides/recipes/subrecipes)** - same note.
4. **[Subagents tutorial](https://block.github.io/goose/docs/tutorials/subagents)** - same note.

**Describe alternatives you've considered**

- A runtime warning when a recipe session starts without `summon` but uses `delegate`. This would help users diagnose the issue without reading docs. Proposed in #7355, not pursued.
- Auto-injecting `summon` for all recipes that reference `delegate` in instructions. Proposed in #7360, rejected by maintainer.

**Additional context**

The silent failure mode makes this particularly difficult to debug. In our case it took several hours, initially misattributed to an unrelated regression (#7353).

- [x] I have verified this does not duplicate an existing feature request

DOsinga:
We're working on this — PR #9231 adds the missing documentation for the summon extension requirement across the subagents, recipe reference, and subrecipes pages. Snoozing while the PR is in review.

PR #7663, opened 2026-03-04
fix(openai): handle server error chunks in streaming responses

### LINKED ISSUE #7645: Stream decode error when using subagents/summon extension (regression in v1.25.0+)
**Describe the bug**

Since v1.25.0, using the summon extension / subagents causes a stream decode error that breaks workflows. The issue does not occur in v1.23 or v1.24 (with Summon disabled). Multiple users are affected.

The error message is:

```
Ran into this error: Request failed:
Stream decode error: Failed to parse streaming chunk: missing field choices at line 1 column 99:
"""error": ("code":500,"message":"Invalid diff: now finding less tool calls!","type":"server_error"))
```

Rolling back to v1.24 resolves the issue. At least one user confirmed that subagent/summon invocations almost always fail in the latest two versions.

---

**To Reproduce**
1. Use Goose v1.25.0 or later
2. Enable the Summon extension
3. Run a workflow that triggers subagent delegation
4. Observe the stream decode error

---

**Expected behavior**

Subagent delegation and streaming should work without errors, as they did in v1.23/v1.24.

---

**Environment details from reporters**
- **Provider & Model:** Local llama.cpp (localhost), though other users report similar failures with different setups
- **Versions affected:** v1.25.0+
- **Versions working:** v1.23, v1.24 (with Summon disabled)
- **Extensions enabled:** Summon

---

**Additional context**

From community discussion:
- One user confirmed the problem disappears entirely on v1.24 rollback and is likely related to the subagents feature introduced/changed in v1.25.0+.
- Another user confirmed that summoning subagents almost always fails in the latest two versions.
- A Goose developer noted that running multiple agents concurrently against a local llama.cpp server could cause problems and may require additional configuration around slots and KV cache setup.
- There may be overlap with other known subagent/summon issues (see #7364, #7570).

Related issues: #7364, #7570

fresh3nough:
## Problem

When an OpenAI-compatible server (e.g. llama.cpp) returns an error during streaming, it sends a JSON chunk with an `error` field instead of the expected `choices` field. The `StreamingChunk` deserialization fails with `missing field choices`, producing a confusing error:

```
Stream decode error: Failed to parse streaming chunk: missing field choices at line 1 column 99:
```

This is particularly triggered by subagents/summon, which create concurrent streaming sessions that can overwhelm local LLM servers (regression since the summon extension was introduced).

## Fix

- Add `#[serde(default)]` to `StreamingChunk.choices` so error-only chunks can be deserialized without crashing
- Add optional `error` field to `StreamingChunk` to capture server error responses  
- Add `check_streaming_error()` helper that propagates server errors with clear messages including the original error code, type, and message
- Check for errors in both the main streaming loop and the inner tool-call accumulation loop
- 3 new tests covering: error-only chunks, errors during tool call streaming, and rate limit errors

This follows the same pattern already used by the Google provider (`formats/google.rs` lines 469-479) which already handles streaming errors correctly.

## Testing

- All 31 existing tests in `formats::openai::tests` pass (no regressions)
- 3 new tests verify error chunk handling:
  - `test_streaming_error_chunk_returns_server_error` - reproduces the exact error from #7645
  - `test_streaming_error_chunk_during_tool_calls` - error mid-tool-call accumulation
  - `test_streaming_error_chunk_with_no_choices_no_crash` - rate limit errors
- `cargo clippy` and `cargo fmt` clean

Fixes #7645
Related: #7364, #7570

### DIFF
--- a/crates/goose/src/providers/formats/openai.rs
+++ b/crates/goose/src/providers/formats/openai.rs
@@ -65,12 +65,21 @@ struct StreamingChoice {
 }
 
 #[derive(Serialize, Deserialize, Debug)]
+struct StreamingError {
+    message: Option<String>,
+    r#type: Option<String>,
+    code: Option<Value>,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
 struct StreamingChunk {
+    #[serde(default)]
     choices: Vec<StreamingChoice>,
     created: Option<i64>,
     id: Option<String>,
     usage: Option<Value>,
     model: Option<String>,
+    error: Option<StreamingError>,
 }
 
 pub fn format_messages(messages: &[Message], image_format: &ImageFormat) -> Vec<Value> {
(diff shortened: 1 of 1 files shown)


DOsinga:
Thanks for the fix @fresh3nough! Unfortunately this has been superseded by #8031 which landed the same fix (with a slightly different approach — using `ProviderError::ServerError` so the error is retryable, and handling both the OpenAI and vLLM error formats). Closing this one out, but the contribution is appreciated — it clearly highlighted a real bug that needed fixing!

PR #7666, opened 2026-03-04
feat: TUI UX improvements

alexhancock:
Various improvements to the core UX of using the new TUI

* Consolidates core into a single top level file called `tui.tsx` and implements other features in a single `.ts` file each (tool calls, markdown support)
* Rich tool calls (shows args and results - tab to open completely)
* Markdown rendering in text content
* Scrolling with arrows

**Demo**

Screenshot 2026-03-04 at 5 38 23 PM


alexhancock:
this merged as part of #7683

PR #7669, opened 2026-03-05
fix: write to real file if config.yaml is symlink

lsytj0413:
## Summary

If config.yaml is a symlink, rename to the real file

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Fix: https://github.com/block/goose/issues/6847

Before:  

After:

### DIFF
--- a/crates/goose/src/config/base.rs
+++ b/crates/goose/src/config/base.rs
@@ -459,20 +459,57 @@ impl Config {
         load_init_config_from_workspace()
     }
 
+    fn config_write_target_path(&self) -> Result<PathBuf, ConfigError> {
+        let mut path = self.config_path.clone();
+
+        // Follow symlinks so we update the target file without replacing the link itself.
+        const MAX_SYMLINK_HOPS: usize = 1;
+        let mut hops = 0usize;
+        loop {
+            match std::fs::symlink_metadata(&path) {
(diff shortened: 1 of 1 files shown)


crates/goose/src/config/base.rs
```diff
@@ -459,20 +459,57 @@ impl Config {
         load_init_config_from_workspace()
     }
 
+    fn config_write_target_path(&self) -> Result<PathBuf, ConfigError> {
```
jamadeo, line 462: The goal makes sense to me, but why not use `fs::canonicalize`?

lsytj0413: Since the config_path may not exist, we should return it as the target_path in this scenario.

jamadeo: That makes sense. It just seems like symlink resolution is something we should let the OS do, not on our own in a loop. But I'm not sure there's a better way if we also want to preserve the creating of parent directories.

PR #7677, opened 2026-03-05
fix: increase Ollama retry config + add transient-only mode

### LINKED ISSUE #7635: Stops waiting for ollama before specified timeout reached
**Describe the bug**

When trying to use Ollama with any decent model I run into error:
"Ran into this error: Server error: Server error (500 Internal Server Error): health resp: Get "http://127.0.0.1:60954/health": dial tcp 127.0.0.1:60954: connectex: No connection could be made because the target machine actively refused it..

Please retry if you think this is a transient or recoverable error." 

It seems to wait Ollama for 5-10 seconds to respond and gives up.

This does not happen if i use very small model such as 2b parameter variant of Qwen3.5.
With small model it patiently waits for model to respond even over minute.

[diagnostics_20260304_3.zip](https://github.com/user-attachments/files/25735459/diagnostics_20260304_3.zip)

---

**To Reproduce**
1. Use local Ollama
2. Select model that takes while to respond. In my case with RTX3060 12GB and 64GB system ram Qwen3.5 35b model that still works fine in Ollama but takes 10-15 seconds to start responding.
3. Ask it anything
4. See the error in chat
---

**Expected behavior**
Get response

---

- **OS & Arch:** Windows 11
- **Interface:** UI
- **Version:** 1.26.1
- **Extensions enabled:** Apps, Code mode, Computer Controller, Developer, Extension Manger, Memory, Summon, Todo, Top Of Mind. But note with these work fine on smaller model
- **Provider & Model:** Local Ollama qwen3.5:35b

---

**Additional context**
Qwen3.5:35b from my local Ollama works fine on Ollama terminal and for other tools that use Ollama as backend.

fresh3nough:
## Problem
When using Ollama with large models (e.g. Qwen3.5 35b), goose gives up after ~7 seconds with a 500 'connection refused' error. Ollama returns HTTP 500 while loading a model into memory, which can take 10-120s for large models on consumer hardware.

The default retry config (3 retries with 1s/2s/4s backoff = ~7s total) is insufficient for this scenario.

## Fix
Override `retry_config()` in `OllamaProvider` with values tuned for local model loading:
- **10 retries** (up from 3)
- **2s initial interval** (up from 1s)
- **1.5x backoff multiplier** (down from 2.0, more gradual ramp)
- **15s max interval** (down from 30s)

This provides ~100s of total retry wait time (even with worst-case jitter, >60s), which handles models that take up to ~2 minutes to load.

## Testing
- Added unit tests verifying the retry config values and that total wait time exceeds 60s
- All existing Ollama tests continue to pass
- `cargo clippy` clean

Closes #7635


crates/goose/src/providers/ollama.rs
```diff
@@ -191,6 +199,15 @@ impl Provider for OllamaProvider {
         self.model.clone()
     }
 
+    fn retry_config(&self) -> RetryConfig {
+        RetryConfig::new(
```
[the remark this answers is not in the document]
fresh3nough: Gate the extended backoff only to transient loading errors (connection refused during model startup, 5xx). Keep immediate fail for client errors (4xx). Update retry predicate in ollama.rs + retry.rs.

crates/goose/src/providers/ollama.rs
```diff
@@ -349,4 +367,73 @@ mod tests {
...
+            let capped = std::cmp::min(base_delay, OLLAMA_MAX_RETRY_INTERVAL_MS);
+            base_total_ms += capped;
+        }
+        let min_total_secs = (base_total_ms as f64 * 0.8) / 1000.0;
+
+        assert!(
+            min_total_secs >= 60.0,
+            "Minimum total retry wait time ({min_total_secs:.1}s) should be >= 60s"
+        );
+    }
```
DOsinga: these tests are not doing much I think, better drop them

fresh3nough: dropped

fresh3nough:
in addition im updating pr title to extending transient mode

DOsinga:
so I am thinking we shouldn't fix this using the retry mechanism - it would just retry any ollama call even the real faulty one > 1m. is there a different way of doing this? something that is more targeted to the issue

fresh3nough:
gotcha, couple options:

1)

Long First-Byte Timeout + Normal Chunk Timeout (builds on your earlier Ollama work) --> Use tokio::time::timeout only until the first SSE line/chunk arrives (like the “defer stall timeout” pattern we discussed before). --> After first chunk → revert to 30s per-chunk. --> This handles slow model loading without touching retries at all.

or

2)

Smart Error-Pattern Retry Predicate
Only extend retries when the error matches known “model still loading” patterns:
Common Ollama signatures during cold start (from real issues + community reports):
HTTP 503 Service Unavailable
HTTP 500 with body containing: "llama runner", "loading model", "server not yet available", "timed out waiting for llama runner to start", or "model is loading"

In code: add a new method in OllamaProvider:Rustfn 

is_model_loading_error(&self, err: &reqwest::Error) -> bool {
    if err.status() == Some(StatusCode::SERVICE_UNAVAILABLE) { return true; }
    let body = err.to_string().to_lowercase();
    body.contains("llama runner") || body.contains("loading model") || ...
}

Then update the should_retry predicate to use this only for Ollama (and only up to ~120s total). Real errors fail instantly. This is precise, zero user impact, and easy to maintain

2 is more what I was thinking tbh

PR #7683, opened 2026-03-05
feat: combine TUI UX from alexhancock/tui-goodness with publishing config from jackamadeo/package-tui

alexhancock:
Combines

https://github.com/block/goose/pull/7395
https://github.com/block/goose/pull/7666

making it possible for us to publish all three of the npm packages that will power the new TUI with an embedded `goose-acp`

I added one more architectural improvement I thought of which was to bring the stream creation into `ui/acp`


ui/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "private": true,
```
michaelneale, line 2: what does private do here? will this change how the desktop is built at all?

alexhancock: it just means this itself won't be registered as a publishable module by `npm`

we only want to publish the constituent packages

PR #7689, opened 2026-03-06
fix(shell): prevent hang when command backgrounds a child process

rabi:
## Summary

When a command spawns a background process that inherits stdout/stderr pipes, the pipe reader blocks forever after the shell exits. Fix by streaming output through an mpsc channel with a 1-second drain timeout after child.wait() returns, so captured output is returned without waiting for the grandchild to exit.

### Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Unit tests and tested locally.

Before:  

```
🪿 Run this shell command and tell me the output: echo 'BEFORE' && sleep 300 & echo 'AFTER'. If you see both BEFORE and AFTER in the output, the test passed.



  ▸ shell
    command: echo 'BEFORE' && sleep 300 & echo 'AFTER'
    timeout_secs: 10

```
stuck forever...

After:   
```
🪿 Run this shell command and tell me the output: echo 'BEFORE' && sleep 300 & echo 'AFTER'. If you see both BEFORE and AFTER in the output, the test passed.



  ▸ shell
    command: echo 'BEFORE' && sleep 300 & echo 'AFTER'
    timeout_secs: 10

AFTER
BEFOREBoth **BEFORE** and **AFTER** appear in the output — **the test passed!** ✅

Here's what happened:
- `echo 'BEFORE' && sleep 300` was sent to the **background** (via `&`), so `BEFORE` printed but `sleep 300` runs in the background without blocking.
- `echo 'AFTER'` ran immediately in the **foreground**, printing `AFTER`.

```

### DIFF
--- a/crates/goose/src/agents/platform_extensions/developer/shell.rs
+++ b/crates/goose/src/agents/platform_extensions/developer/shell.rs
@@ -25,7 +25,7 @@ pub struct ShellParams {
     pub timeout_secs: Option<u64>,
 }
 
-#[derive(Debug, Serialize, JsonSchema)]
+#[derive(Debug, Serialize, Deserialize, JsonSchema)]
 pub struct ShellOutput {
     pub stdout: String,
     pub stderr: String,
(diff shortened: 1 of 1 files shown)


lifeizhou-ap (commented):
Nice fix — the channel + drain timeout approach is clean and solves the problem well. One minor thought:

**Drain timeout duration:** `OUTPUT_DRAIN_TIMEOUT_SECS` is currently 5 seconds. In practice, useful output arrives in the channel almost instantly (before the process even exits), so the 5s is really just a safety margin. Would you consider reducing it to 1-2 seconds? It would make commands with accidental background processes feel snappier. Not a blocker — 5s is fine too, just something to consider.

lifeizhou-ap (approved):

rabi:
> Would you consider reducing it to 1-2 seconds? 

Yeah. that makes sense thanks. I've mentioned 1s in the commit message whereas somehow it was 5s in the code. We can change it to milliseconds as well.

PR #7696, opened 2026-03-06
delete goose web

DOsinga:
## Summary

What it says on the label


blackgirlbytes:
kk, did we check if people really dont use this? My team did use it heavily (although no longer here)

atline:
Can I understand the reason for this?

PR #7707, opened 2026-03-06
fix(ui): sync sidebar sessions after chat history deletion

zuosus:
## Summary
This PR fixes a UI sync issue where deleting a chat from `Chat -> Show All -> Chat History` did not immediately remove that topic from the sidebar list until app restart.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### DIFF
--- a/ui/desktop/src/hooks/useNavigationSessions.ts
+++ b/ui/desktop/src/hooks/useNavigationSessions.ts
@@ -140,8 +140,7 @@ export function useNavigationSessions(options: UseNavigationSessionsOptions = {}
 
   useEffect(() => {
     const handleSessionDeleted = (event: Event) => {
-      const { sessionId } = (event as CustomEvent<{ sessionId?: string }>).detail || {};
-      if (!sessionId) return;
+      const { sessionId } = (event as CustomEvent<{ sessionId: string }>).detail;
 
       setRecentSessions((prev) => prev.filter((session) => session.id !== sessionId));
       sessionsRef.current = sessionsRef.current.filter((session) => session.id !== sessionId);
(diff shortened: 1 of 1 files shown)


DOsinga:
Thanks for the PR — this is a nice bug fix that wires up listeners for events that were clearly meant to be consumed.

A couple of things to address before we can merge:

1. **Remove the redundant `fetchSessions()` call in `handleSessionDeleted`.** You already optimistically filter the session out of both `recentSessions` and `sessionsRef.current`, which gives instant UI feedback. The extra `fetchSessions()` call is unnecessary overhead (a full API round-trip) — the local state update is sufficient here, and it matches how `handleSessionRenamed` works (local update only, no re-fetch). If you want belt-and-suspenders consistency, the existing polling in `handleSessionCreated` already covers that path.

2. **Use the correct types from `vite-env.d.ts` instead of optional fields.** The event detail types in `vite-env.d.ts` declare `sessionId` (and `newName`) as required, but your casts use optional fields (`sessionId?: string`, `newName?: string`). Please match the declared types — e.g. `(event as CustomEvent<{ sessionId: string }>)` — so the type safety is consistent with the rest of the codebase.

3. **DCO sign-off is failing.** Please amend your commit(s) with `git commit --amend --signoff` and force-push to fix the DCO check.

zuosus:
Thanks @DOsinga , issues fixed.

Issue #7715, opened 2026-03-07
bad connection with to ollama container


f1-outsourcing:
**Describe the bug**

Typing at the console "hi" hangs the ollama container on max cpu

goose container
```
wget 'https://github.com/block/goose/releases/download/v1.27.0/goose-x86_64-unknown-linux-gnu.tar.bz2'

docker run -v /dev/log:/dev/log -it --network host debian:bookworm-slim /bin/bash
docker cp goose goose:/usr/local/bin/
apt-get update
apt-get install libgomp1 ca-certificates
```

ollama container 

```
docker run -v /dev/log:/dev/log -v /home/data:/root/.ollama --network host --cpus=6 ollama/ollama:latest

docker exec -it ollama ollama pull llama3 (does not work with goose? (tools/function calling) use llama3.1)
docker exec -it ollama ollama list
```

tests like these are fine, and produce immediate results

`docker exec -it ollama ollama run llama3.1 "hi"`

```
curl -X POST http://localhost:11434/api/generate \
  -H "Content-Type: application/json" \
  -d '{
    "model": "llama3.1",
    "prompt": "hi",
    "stream": true
  }'
```

However typing "hi" in goose takes forever, load increases to max cpu on the ollama container, however does not seem to end.

PR #7723, opened 2026-03-08
fix: prevent Ollama provider from hanging on tool-calling requests

fresh3nough:
## Summary

Strip unsupported OpenAI-specific fields (`stream_options`, `max_completion_tokens`) from the request payload sent to Ollama's `/v1/chat/completions` endpoint. Convert `max_completion_tokens` to Ollama's native `options.num_predict`. Add a 30-second per-chunk streaming timeout so stalled Ollama connections fail fast instead of hanging for the full 600s timeout.

## Problem

When goose sends a chat completion request to Ollama (especially with tool definitions), Ollama hangs indefinitely at max CPU. Direct `curl` and `ollama run` work fine. The root cause is that goose's Ollama provider reuses the generic OpenAI `create_request` which emits fields Ollama does not support:

- `stream_options: {"include_usage": true}` -- not recognized by Ollama
- `max_completion_tokens` -- should be `num_predict` inside Ollama's `options` object

## Changes

- **`apply_ollama_options`**: Now strips `stream_options` and converts `max_completion_tokens` to `options.num_predict` before sending to Ollama
- **`stream_ollama`**: Added a 30-second per-chunk timeout so stalled streams produce a clear error instead of hanging silently
- **Tests**: Added 3 new tests:
  - Bug reproduction: verifies `create_request` produces the problematic fields
  - Fix verification: verifies `apply_ollama_options` removes them correctly
  - Streaming timeout: verifies stalled streams produce an error within the timeout window

Fixes https://github.com/block/goose/issues/7715


fresh3nough:
Not at all I'll take care of it today

fresh3nough:
should be good now @DOsinga @michaelneale

kasjens:
Heads up — #8493 builds on top of this PR's work. The `stream_options` removal and per-chunk timeout you introduced here were solid fixes at the time. Since Ollama now supports `stream_options` (ollama/ollama#4448), we've gated it behind `OLLAMA_STREAM_USAGE` (default: true) instead of unconditionally stripping it, and made the chunk timeout configurable. Thanks for the original fix — the timeout infrastructure made these improvements possible.

PR #7726, opened 2026-03-08
fix: persist manage_extensions changes to config.yaml

jeffa-block:
## Summary

The manage_extensions tool disables/enables extensions at runtime but doesn't 
persist the change to config.yaml. Next session, the extension reverts to its 
previous state. This adds a call to the existing `set_extension_enabled()` 
function after each successful enable/disable so the change is saved to disk.

### Type of Change
- [x] Bug fix

### AI Assistance

### Testing
- `cargo check -p goose` passes
- The fix uses existing `set_extension_enabled()` and `name_to_key()` functions 
  that are already tested elsewhere in the codebase

### Related Issues
New bug — no existing issue.

### DIFF
--- a/crates/goose/src/agents/platform_extensions/ext_manager.rs
+++ b/crates/goose/src/agents/platform_extensions/ext_manager.rs
@@ -1,6 +1,7 @@
 use crate::agents::extension::PlatformExtensionContext;
 use crate::agents::mcp_client::{Error, McpClientTrait};
-use crate::config::get_extension_by_name;
+use crate::config::extensions::name_to_key;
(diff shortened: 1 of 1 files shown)


lifeizhou-ap:
Hi @jeffa-block,

Thanks for putting up this PR.  

In goose,  "manage_extensions tool disables/enables extensions at runtime" is the expected behaviour.  

We have documented [here](https://block.github.io/goose/docs/getting-started/using-extensions/):
> Extension Manager: Discover, enable, and disable extensions dynamically during sessions

I'm going to close this PR, but we appreciate the contribution!

PR #7730, opened 2026-03-08
use constant-time comparison for auth token validation

### LINKED ISSUE #7710: Security: Timing Attack Vulnerability in Authentication Middleware
# Security: Timing Attack Vulnerability in Authentication Middleware

## Summary
The authentication middleware in `crates/goose-server/src/auth.rs` uses non-constant-time string comparison (`==`) when validating the `X-Secret-Key` header against the configured secret. This makes the authentication vulnerable to timing attacks, allowing attackers to potentially deduce the secret key byte-by-byte through precise timing measurements.

## Vulnerability Details

**Location:** `crates/goose-server/src/auth.rs:22-23`

```rust
match secret_key {
    Some(key) if key == state => Ok(next.run(request).await),
    _ => Err(StatusCode::UNAUTHORIZED),
}
```

The `key == state` comparison uses Rust's default `PartialEq` implementation for `&str`, which performs **variable-time comparison** — it stops at the first differing byte. This timing difference can be measured by attackers to recover the secret incrementally.

## Attack Scenario

1. Attacker sends requests with controlled `X-Secret-Key` values
2. By measuring response times, attacker detects when the first byte matches (slightly longer processing time)
3. Attacker iterates through all possible bytes for position 2, repeating the process
4. Eventually recovers the entire secret key

While network jitter may make this harder, local or low-latency environments (e.g., same data center, VPN, or side-channel attacks) significantly increase feasibility.

## Recommended Fix

Use **constant-time comparison** to prevent timing side-channels:

### Option 1: Use `subtle` crate (recommended)
```rust
// Add to Cargo.toml:
// subtle = "2.6"

use subtle::ConstantTimeEq;

match secret_key {
    Some(key) if key.as_bytes().ct_eq(state.as_bytes()).into() => {
        Ok(next.run(request).await)
    }
    _ => Err(StatusCode::UNAUTHORIZED),
}
```

### Option 2: Manual constant-time comparison
```rust
fn constant_time_compare(a: &str, b: &str) -> bool {
    if a.len() != b.len() {
        return false;
    }
    a.as_bytes()
        .iter()
        .zip(b.as_bytes())
        .fold(0u8, |acc, (x, y)| acc | (x ^ y))
        == 0
}

match secret_key {
    Some(key) if constant_time_compare(key, &state) => {
        Ok(next.run(request).await)
    }
    _ => Err(StatusCode::UNAUTHORIZED),
}
```

## Impact

- **Severity:** Medium to High (depending on deployment context)
- **Affected:** All deployments using authentication middleware with `X-Secret-Key`
- **Exploitability:** Requires precise timing measurements; easier in low-latency or local environments

## References

- [CWE-208: Observable Timing Discrepancy](https://cwe.mitre.org/data/definitions/208.html)
- [OWASP: Timing Attack](https://owasp.org/www-community/attacks/Timing_attack)
- [`subtle` crate documentation](https://docs.rs/subtle/)

---

**Reporter:** Zesty Cody (OpenClaw Code Analyst)  
**Detection Date:** 2026-03-07

extrasmall0:
The auth middleware in `auth.rs` uses `==` for comparing the `X-Secret-Key` header against the stored secret. Rust's default string comparison short-circuits on the first differing byte, which leaks timing information that could let an attacker recover the key incrementally.

This swaps the comparison for `subtle::ConstantTimeEq::ct_eq`, which always compares every byte regardless of where they differ. `subtle` is already in the dependency tree (pulled in transitively), so this just adds it as a direct dep for `goose-server`.

The explicit `key.len() == state.len()` guard before `ct_eq` is technically redundant (ct_eq on unequal-length slices returns 0), but makes the intent clearer and avoids doing the full comparison when lengths obviously differ.

Closes #7710

### DIFF
--- a/crates/goose-server/src/auth.rs
+++ b/crates/goose-server/src/auth.rs
@@ -24,9 +24,7 @@ pub async fn check_token(
         .and_then(|value| value.to_str().ok());
 
     match secret_key {
-        Some(key)
-            if key.len() == state.len() && bool::from(key.as_bytes().ct_eq(state.as_bytes())) =>
-        {
+        Some(key) if bool::from(key.as_bytes().ct_eq(state.as_bytes())) => {
             Ok(next.run(request).await)
         }
         _ => Err(StatusCode::UNAUTHORIZED),


jh-block:
Doesn't the explicit length guard let an attacker know whether their key is the right length or not based on the timing? I don't think it makes the intent clearer and doing the full computation every time is rather the point of constant time comparison. Happy to merge if that is removed.

extrasmall0:
Good point — the length check does leak that info through timing. Removed it in 99e1d11, ct_eq handles different lengths fine on its own.

extrasmall0:
Thanks for the quick review and merge @jh-block!

PR #7738, opened 2026-03-09
fix: provider test infrastructure

codefromthecrypt:
## Summary

Fix Ollama, scheduler, and LiteLLM test failures and make it easy to add new providers like ACP.

- Introduce `ProviderTestConfig` with `with_llm_provider` and `with_agentic_provider` for isolated and easier setup.
- Isolate each provider sub-test with its own Agent and session, removing extension re-registration and session ID complexity
- Make ollama pass: Ollama VL model unreliable for tool calls; split into qwen3/qwen3-vl
- Cleanups: Scheduler env leakage into keyring; LiteLLM hardcoded host overwrite

### Type of Change
- [x] Bug fix
- [x] Refactor / Code quality
- [x] Tests

### AI Assistance

### Testing

```
$ OLLAMA_HOST=http://localhost:11434 cargo test --test providers -- --nocapture

--snip--

test result: ok. 16 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 201.63s

============== Providers ==============
⏭️ Anthropic
⏭️ Azure
⏭️ Databricks
⏭️ Google
⏭️ LiteLLM
✅ Ollama
⏭️ OpenRouter
⏭️ SageMakerTgi
⏭️ Snowflake
⏭️ Xai
⏭️ aws_bedrock
✅ claude-code
✅ codex
⏭️ openai
=======================================
```

### DIFF
--- a/crates/goose/src/scheduler.rs
+++ b/crates/goose/src/scheduler.rs
@@ -1013,6 +1013,12 @@ mod tests {
 
     #[tokio::test]
     async fn test_job_runs_on_schedule() {
+        let _guard = env_lock::lock_env([
+            ("GOOSE_PROVIDER", Some("openai")),
+            ("GOOSE_MODEL", Some("gpt-4o")),
+            ("OPENAI_API_KEY", Some("fake-openai-no-keyring")),
+            ("OPENAI_CUSTOM_HEADERS", Some("")),
+        ]);
         let temp_dir = tempdir().unwrap();
         let storage_path = temp_dir.path().join("schedule.json");
         let recipe_path = create_test_recipe(temp_dir.path(), "scheduled_job");
@@ -1039,6 +1045,12 @@ mod tests {
 
     #[tokio::test]
     async fn test_paused_job_does_not_run() {
+        let _guard = env_lock::lock_env([
+            ("GOOSE_PROVIDER", Some("openai")),
+            ("GOOSE_MODEL", Some("gpt-4o")),
+            ("OPENAI_API_KEY", Some("fake-openai-no-keyring")),
+            ("OPENAI_CUSTOM_HEADERS", Some("")),
+        ]);
         let temp_dir = tempdir().unwrap();
         let storage_path = temp_dir.path().join("schedule.json");
         let recipe_path = create_test_recipe(temp_dir.path(), "paused_job");
@@ -1066,6 +1078,12 @@ mod tests {
 
     #[tokio::test]
     async fn test_job_with_no_prompt_does_not_panic() {
+        let _guard = env_lock::lock_env([
+            ("GOOSE_PROVIDER", Some("openai")),
+            ("GOOSE_MODEL", Some("gpt-4o")),
+            ("OPENAI_API_KEY", Some("fake-openai-no-keyring")),
+            ("OPENAI_CUSTOM_HEADERS", Some("")),
+        ]);
         let temp_dir = tempdir().unwrap();
         let recipe_path = temp_dir.path().join("no_prompt.yaml");
         fs::write(
--- a/crates/goose/tests/providers.rs
+++ b/crates/goose/tests/providers.rs
@@ -1,10 +1,7 @@
 use anyhow::Result;
 use dotenvy::dotenv;
 use futures::StreamExt;
-use goose::agents::extension_manager::ExtensionManagerCapabilities;
-use goose::agents::{
-    Agent, AgentConfig, AgentEvent, ExtensionManager, GoosePlatform, PromptManager, SessionConfig,
-};
+use goose::agents::{Agent, AgentConfig, AgentEvent, GoosePlatform, PromptManager, SessionConfig};
 use goose::config::{ExtensionConfig, GooseMode, PermissionManager};
 use goose::conversation::message::{ActionRequiredData, Message, MessageContent};
 use goose::permission::permission_confirmation::PrincipalType;
@@ -25,7 +22,7 @@ use goose::providers::sagemaker_tgi::SAGEMAKER_TGI_DEFAULT_MODEL;
 use goose::providers::snowflake::SNOWFLAKE_DEFAULT_MODEL;
 use goose::providers::xai::XAI_DEFAULT_MODEL;
 use goose::session::{SessionManager, SessionType};
-use goose_test_support::{ExpectedSessionId, McpFixture, FAKE_CODE, TEST_SESSION_ID};
+use goose_test_support::{ExpectedSessionId, McpFixture, FAKE_CODE};
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
 use tokio_util::sync::CancellationToken;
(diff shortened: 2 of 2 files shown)


crates/goose/src/scheduler.rs
```diff
@@ -1013,6 +1013,12 @@ mod tests {
 
     #[tokio::test]
     async fn test_job_runs_on_schedule() {
+        let _guard = env_lock::lock_env([
```
codefromthecrypt, line 1016: this is the pattern we use to make sure that the local goose config isn't read (e.g. my provider was set to tetrate, and this test ran trying to access its creds)

crates/goose/tests/providers.rs
```diff
@@ -93,42 +90,216 @@ lazy_static::lazy_static! {
     static ref ENV_LOCK: Mutex<()> = Mutex::new(());
 }
 
-struct ProviderTester {
-    provider: Arc<dyn Provider>,
+fn mode_str(mode: GooseMode) -> &'static str {
```
codefromthecrypt: this cruft is removable when https://github.com/block/goose/pull/7706 is merged, and I'll follow up on that race condition

crates/goose/tests/providers.rs
```diff
@@ -93,42 +90,216 @@ lazy_static::lazy_static! {
...
-impl ProviderTester {
-    fn new(
-        provider: Arc<dyn Provider>,
-        name: String,
-        extension_manager: Arc<ExtensionManager>,
-        is_cli_provider: bool,
-        model_switch_name: Option<String>,
-        mcp_extension: ExtensionConfig,
+impl ProviderTestConfig {
+    fn with_llm_provider(
```
codefromthecrypt, line 124: llm providers are all remote apis and they don't internalize tool calls like agents do. This is how providers typically work in goose

crates/goose/tests/providers.rs
```diff
@@ -93,42 +90,216 @@ lazy_static::lazy_static! {
...
+        self.expect_context_length_exceeded = v;
+        self
+    }
+
+    fn context_length_exceeded(mut self, token_count: usize) -> Self {
+        self.context_length_exceeded = token_count;
+        self
+    }
+
+    fn with_agentic_provider(name: &'static str, model_name: &'static str, binary: &str) -> Self {
```
codefromthecrypt, line 181: agentic providers currently are all (both CLI and ACP) binary based, even if they can also read ENV. so the guard for them is at least about the binary.

- *right now* we don't have the means to propagate session IDs to their backend LLM or MCP apis, though it is possible one might support this via the corresponding specs.
- *right now* they aren't semantically compatible with smart approve, though this might change in the future.
- we didn't run content length tests on these in the past because they accept huge contexts and that results in extreme delays

crates/goose/tests/providers.rs
```diff
@@ -93,42 +90,216 @@ lazy_static::lazy_static! {
...
+        }
+    }
+
+    async fn run(self) -> Result<()> {
+        test_provider(self).await
+    }
+}
+
+impl ProviderFixture {
+    async fn setup(config: &ProviderTestConfig, mode: GooseMode) -> Result<Self> {
```
codefromthecrypt, line 198: Before, it was really complicated in the tests themselves because they switched from default auto to a specific mode. This isolates each test to a provider instance so it doesn't need to switch modes. If later we want to add a runtime mode switch test, we could, but that's a different feature than testing a specific mode behavior.

crates/goose/tests/providers.rs
```diff
@@ -93,42 +90,216 @@ lazy_static::lazy_static! {
...
+            display_name: Some("Developer".to_string()),
+            timeout: None,
+            bundled: None,
+            available_tools: vec![],
+        };
+
+        let provider = create_with_named_model(
+            &config.name.to_lowercase(),
+            config.model_name,
+            vec![mcp_extension.clone(), developer_extension.clone()],
```
codefromthecrypt, line 227: now all tests enable an MCP and developer extension, as it doesn't harm tests that don't use them and also removes test-site complexity of loading up the extension manager.

crates/goose/tests/providers.rs
```diff
@@ -163,26 +334,23 @@ impl ProviderTester {
...
             .extension_manager
-            .dispatch_tool_call(session_id, params, None, CancellationToken::new())
+            .dispatch_tool_call(&self.session_id, params, None, CancellationToken::new())
             .await
-            .expect("dispatch failed")
+            .unwrap()
             .result
             .await
-            .expect("tool call failed");
+            .unwrap();
```
codefromthecrypt, line 337: I removed the expect messages here as typically in our tests we don't add them, leaving this inconsistent with other tests. However, I can put them back if folks want.

crates/goose/tests/providers.rs
```diff
@@ -227,39 +388,27 @@ impl ProviderTester {
...
+            .tool_roundtrip("Use the get_code tool and output only its result.", None)
             .await?;
         let text = response.as_concat_text();
-        assert!(
-            text.contains(FAKE_CODE),
-            "Expected lookup code '{}' in final response, got: {}",
-            FAKE_CODE,
-            text
-        );
+        assert!(text.contains(FAKE_CODE), "{text}");
```
codefromthecrypt, line 387: when the assert fail can be due to a flake, I do add that into the message. otherwise the line that failed is fine.

crates/goose/tests/providers.rs
```diff
@@ -227,39 +388,27 @@ impl ProviderTester {
...
 
-    async fn test_context_length_exceeded_error(&self, session_id: &str) -> Result<()> {
-        let large_message_content = if self.name.to_lowercase() == "google" {
-            "hello ".repeat(1_300_000)
-        } else {
-            "hello ".repeat(300_000)
-        };
-
+    async fn test_context_length_exceeded_error(&self) -> Result<()> {
+        // "hello " ≈ 2 tokens across common tokenizers
```
codefromthecrypt, line 393: usually we don't add comments, but it was unclear to me why we are counting hellos. the test is really a function of tokens, so I think this is the better way forward.

crates/goose/tests/providers.rs
```diff
@@ -270,70 +419,62 @@ impl ProviderTester {
         dbg!(&result);
         println!("===================");
 
-        let name_lower = self.name.to_lowercase();
-        if name_lower == "ollama" || name_lower == "openrouter" {
```
codefromthecrypt, line 274: you can see this stuff has been high maintenance, hence moving to a feature based test system vs remmebering to update names in various places.

crates/goose/tests/providers.rs
```diff
@@ -509,30 +552,29 @@ impl ProviderTester {
     }
 
     async fn test_permission_allow(&self) -> Result<()> {
+        let test_file = tempfile::NamedTempFile::new()?;
         self.run_permission_test(
-            GooseMode::Approve,
             Permission::AllowOnce,
             true,
-            "Use the get_code tool and output only its result.",
+            &format!("Write the word 'hello' to {}", test_file.path().display()),
```
codefromthecrypt, line 550: changed this because we add read-only attribute to the MCP get_code which means in some agents it is not called as they permit read-only without permission. All agents I've used can write files, and so regardless of LLM (developer extension) or an agent (claude etc) we know that this write to a new path will require a permission guard.

crates/goose/tests/providers.rs
```diff
@@ -693,238 +694,161 @@ async fn test_provider(
...
-    test_provider(
-        "Ollama",
-        "qwen3-vl",
-        Some("qwen3"),
-        &["OLLAMA_HOST"],
-        None,
-        false,
-    )
-    .await
+    ProviderTestConfig::with_llm_provider("Ollama", "qwen3", &["OLLAMA_HOST"])
```
codefromthecrypt, line 752: split image model from the default one because qwen3-vl is really bad at tool execution and fails routinely.

crates/goose/tests/providers.rs
```diff
@@ -693,238 +694,161 @@ async fn test_provider(
...
+        .image_model("qwen3-vl")
+        // Above qwen3's 40960 context_length but small enough for Ollama's 600s timeout
```
codefromthecrypt, line 754: this was really not intuitive to hunt down, but yeah we were not checking anything except that the contentlengthexceeded failure. the prior config actually failed just with a different error due to overloading the ollama process!

crates/goose/tests/providers.rs
```diff
@@ -693,238 +694,161 @@ async fn test_provider(
...
-    test_provider(
-        "claude-code",
-        CLAUDE_CODE_DEFAULT_MODEL,
-        Some("sonnet"),
-        &[],
-        None,
-        true,
-    )
-    .await
+    ProviderTestConfig::with_agentic_provider("claude-code", CLAUDE_CODE_DEFAULT_MODEL, "claude")
```
codefromthecrypt, line 831: much less cruft this way, which is needed as there are many untested agentic providers

crates/goose/tests/providers.rs
```diff
@@ -559,124 +600,84 @@ fn load_env() {
...
-    } else {
-        Some(ExpectedSessionId::default())
-    };
-    let mcp = McpFixture::new(expected_session_id.clone()).await;
-    if let Some(ref id) = expected_session_id {
-        id.set(TEST_SESSION_ID);
+    if run_test(GooseMode::Auto).await.is_err() {
+        println!("Skipping {} tests - failed to create provider", name);
+        TEST_REPORT.record_skip(name);
+        return Ok(());
```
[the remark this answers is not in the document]
codefromthecrypt: 👎 it is a bigger break if Agent::update_provider/add_extension are failing and this file is not trying to test that.

crates/goose/tests/providers.rs
```diff
@@ -553,124 +591,84 @@ fn load_env() {
...
-
-        // Check required_vars BEFORE applying env_modifications to avoid
-        // leaving the environment mutated when skipping
-        let missing_vars = required_vars.iter().any(|var| std::env::var(var).is_err());
-        if missing_vars {
+        if config
+            .required_vars
+            .iter()
+            .any(|var| std::env::var(var).is_err())
+        {
```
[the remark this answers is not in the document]
codefromthecrypt: 👎 env-lock serializes so there is no race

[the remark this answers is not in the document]
codefromthecrypt: 👎 I already told you about this

PR #7745, opened 2026-03-09
fix: server route improvements — conflict helper, metric name, provider check

bioinfornatics:
## Summary

Three targeted fixes to server route handlers:

1. **Conflict helper** — fixes edge case in session conflict resolution
2. **Metric name** — corrects telemetry metric naming for consistency
3. **Provider check** — adds missing provider availability validation

### Changes
- 3 files modified
- 9 insertions, 6 deletions (net +3 lines)

### Verification
- ✅ cargo check
- ✅ cargo clippy (0 warnings)
- ✅ cargo test (5 server tests pass)
- ✅ cargo fmt

### DIFF
--- a/crates/goose-server/src/routes/errors.rs
+++ b/crates/goose-server/src/routes/errors.rs
@@ -31,6 +31,14 @@ impl ErrorResponse {
         }
     }
 
+    #[allow(dead_code)]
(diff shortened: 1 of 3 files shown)


DOsinga:

PR #7746, opened 2026-03-09
refactor: simplify ComputerController — remove peekaboo dependency (-373 lines)

bioinfornatics:
## Summary

Removes the `peekaboo` dependency from `ComputerController`, replacing its complex screen capture logic with a simpler direct implementation.

### Changes
- 4 files changed
- 44 insertions, 417 deletions (**net -373 lines** 🎉)
- Removes external dependency

### What changed
- Replaced peekaboo-based screen capture with direct implementation
- Simplified the ComputerController interface
- Removed unused helper code

### Verification
- ✅ cargo check
- ✅ cargo clippy (0 warnings)
- ✅ cargo fmt


jh-block:
The simpler direct implementation seems to just be AppleScript? AppleScript and direct computer control seem like separate tools; in particular the UI testing use case is not solved well by AppleScript (as with AS you target elements rather than coordinates). I think the AppleScript tool would be a great contribution but also think the peekaboo integration has value.

jamadeo:
The peekaboo implementation was added recently in #7342. I am all for simplifying in general but perhaps we should have a discussion first? With @michaelneale at least given he just worked on this. Feel free to open an issue or start a chat on the discord and we can discuss.

PR #7748, opened 2026-03-09
feat: editor model providers — MorphLLM, OpenAI-compatible, and Relace

bioinfornatics:
## Summary

Adds three new editor/coding-focused model providers:

1. **MorphLLM** — Morph Labs' code editing model
2. **OpenAI-compatible** — generic provider for OpenAI-API-compatible endpoints
3. **Relace** — Relace's fast code completion model

### Changes
- 7 files changed, 699 insertions
- All new files under `crates/goose/src/providers/`
- Provider registrations in `factory.rs`

### Verification
- ✅ cargo check
- ✅ cargo clippy (0 warnings)
- ✅ cargo fmt


DOsinga:
this is an interesting idea, but again needs broader discussion. I don't think the developer tool should have its own provider management

DOsinga:

PR #7751, opened 2026-03-09
feat: Developer Analyze Tool — AST-based code analysis engine

bioinfornatics:
## Summary

Adds a Developer Analyze Tool — an AST-based code analysis engine for the Developer MCP extension.

### Features
- AST parsing and symbol extraction for multiple languages
- Code structure analysis (functions, classes, imports, exports)
- Dependency graph construction
- Integration with the developer extension

### Verification
- ✅ cargo check
- ✅ cargo clippy (0 warnings)
- ✅ cargo fmt


DOsinga:

PR #7753, opened 2026-03-09
feat: Extension Registry — discovery, install, and lifecycle management

bioinfornatics:
## Summary

Adds an Extension Registry system for discovering, installing, and managing MCP extensions.

### Features
- Registry protocol for extension discovery
- Extension metadata and versioning
- Install/uninstall lifecycle management
- Server-side registry routes

### Changes
- 24 files changed, ~4,258 insertions

### Verification
- ✅ cargo check
- ✅ cargo clippy (0 warnings)
- ✅ cargo fmt


DOsinga:

PR #7754, opened 2026-03-09
feat: prompt templates for specialist agents

bioinfornatics:
## Summary

Adds prompt templates for specialist agent behaviors and capabilities.

### Features
- Structured prompt templates for different agent specializations
- Template interpolation and composition
- Integration with agent configuration

### Verification
- ✅ cargo check
- ✅ cargo clippy (0 warnings)
- ✅ cargo fmt


DOsinga:
I think these should probably just be skills. but also not part of core goose until we are really sure.  you could set up something like https://github.com/block/goosetown/blob/main/README.md or contrbute there

PR #7755, opened 2026-03-09
ci: update workflow configurations

bioinfornatics:
## Summary

Updates CI/CD workflow configurations for improved build and test pipelines.

### Verification
- Non-Rust changes (GitHub Actions YAML files)
- No build impact


.github/workflows/release-branches.yml
```diff
@@ -27,20 +27,4 @@ jobs:
             [📱 Download macOS Desktop App (arm64, unsigned)](https://nightly.link/${{ github.repository }}/actions/runs/${{ github.run_id }}/Goose-darwin-arm64.zip)
 
             **Instructions:**
-            
-            The easiest way is to just run the following script:
-            
-            `./scripts/pre-release.sh`
```
DOsinga, line 33: we need to keep this

DOsinga (commented):
I don't know what we're trying to achieve here, but I opened a random file and that deletes something we want to keep. so I think I'll close this for now.

PR #7758, opened 2026-03-09
fix(moim): avoid injecting context into active tool-call loops

### LINKED ISSUE #6293: Gemini causing empty replies after tool use
## Summary

When sending `functionResponse` back to Gemini after a tool call, the code uses the internal tracking ID (`response.id`) instead of the original function name. This violates the Gemini API contract and causes Gemini to return empty responses.

## Evidence

From a user's diagnostic logs (gemini-2.5-flash-lite):
- **75 out of 76 tool responses resulted in empty assistant replies**
- 100% of responses after `functionResponse` were empty
- 100% of responses after `text` messages were non-empty

Example from logs:
```
functionCall.name = "developer__text_editor"   ✓ Correct
functionResponse.name = "d9X7Azly"              ✗ Wrong (random ID)
```

## Root Cause

In `crates/goose/src/providers/formats/google.rs` (lines 130-131 and 148-149):

```rust
function_response.insert("name".to_string(), json!(response.id));
```

The `response.id` is a random 8-character alphanumeric string generated internally for tracking, not the function name.

## Expected Behavior

Per [Google's Gemini API docs](https://ai.google.dev/gemini-api/docs/function-calling), `functionResponse.name` must match the `functionCall.name` that was originally sent.

## Comparison with Other Providers

| Provider | Field | Value Used | Correct? |
|----------|-------|------------|----------|
| Anthropic | `tool_use_id` | `response.id` | ✅ Yes (ID-based) |
| OpenAI | `tool_call_id` | `response.id` | ✅ Yes (ID-based) |
| Google/Gemini | `name` | `response.id` | ❌ **Bug** (should be function name) |

## Proposed Fix

The `functionResponse.name` should use the original function name. This requires either:
1. Storing the function name in `ToolResponse` when created (from the corresponding `ToolRequest`)
2. Looking up the function name from the matching `ToolRequest` when formatting the response

## User Impact

Users experience:
- "No response after tool use"
- Tools execute but Goose goes silent
- Files modified but no acknowledgment from the model

## Notes

- This code has been present since V1.0 (`1c9a7c0b050`)

rabi:
## Summary
MOIM insertion now only happens before the last assistant message that does not contain a tool request; if no such anchor exists, MOIM is skipped for that turn. This preserves tool-call sequencing and prevents MOIM from being inserted into in-flight tool loops that breaks for gemini models.

### Type of Change
- [] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested locally with the reproducer in #6293

### Related Issues

Closes: #6293

### DIFF
--- a/crates/goose/src/agents/moim.rs
+++ b/crates/goose/src/agents/moim.rs
@@ -1,5 +1,5 @@
 use crate::agents::extension_manager::ExtensionManager;
-use crate::conversation::message::Message;
+use crate::conversation::message::{Message, MessageContent};
 use crate::conversation::{fix_conversation, Conversation};
 use rmcp::model::Role;
 use std::path::Path;
@@ -7,6 +7,13 @@ use std::path::Path;
 // Test-only utility. Do not use in production code. No `test` directive due to call outside crate.
 thread_local! {
     pub static SKIP: std::cell::Cell<bool> = const { std::cell::Cell::new(false) };
+}
+
+fn has_tool_request(message: &Message) -> bool {
+    message
+        .content
+        .iter()
+        .any(|c| matches!(c, MessageContent::ToolRequest(_)))
 }
 
 pub async fn inject_moim(
@@ -24,10 +31,16 @@ pub async fn inject_moim(
         .await
     {
         let mut messages = conversation.messages().clone();
-        let idx = messages
+        let idx = if let Some(pos) = messages
             .iter()
-            .rposition(|m| m.role == Role::Assistant)
-            .unwrap_or(0);
+            .rposition(|m| m.role == Role::Assistant && !has_tool_request(m))
+        {
+            pos
+        } else if !messages.iter().any(|m| m.role == Role::Assistant) {
+            0
+        } else {
+            return conversation;
+        };
         messages.insert(idx, Message::user().with_text(moim));
 
         let (fixed, issues) = fix_conversation(Conversation::new_unvalidated(messages));
@@ -54,7 +67,7 @@ mod tests {
     use std::path::PathBuf;
 
     #[tokio::test]
-    async fn test_moim_injection_before_assistant() {
+    async fn test_moim_injected_before_assistant() {
         let temp_dir = tempfile::tempdir().unwrap();
         let em = ExtensionManager::new_without_provider(temp_dir.path().to_path_buf());
         let working_dir = PathBuf::from("/test/dir");
@@ -77,13 +90,12 @@ mod tests {
             .filter_map(|c| c.as_text())
             .collect::<Vec<_>>()
             .join("");
-        assert!(merged_content.contains("Hello"));
         assert!(merged_content.contains("<info-msg>"));
         assert!(merged_content.contains("Working directory: /test/dir"));
     }
 
     #[tokio::test]
-    async fn test_moim_injection_no_assistant() {
+    async fn test_moim_injected_without_assistant_message() {
         let temp_dir = tempfile::tempdir().unwrap();
         let em = ExtensionManager::new_without_provider(temp_dir.path().to_path_buf());
         let working_dir = PathBuf::from("/test/dir");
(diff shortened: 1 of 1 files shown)


tlongwell-block (changes requested):
MOIM is already skipped if it introduces an error.

Please fix fix_conversation to catch an error if there is one.

Also, MOIM doesn't break tool call pairs. That should be checked for in fix_conversation anyway.

DOsinga:
hmm, I think we should make sure that fix_conversation can fix any obvious errors, but MOIM should not rely on that - ideally fix_conversation sits there as an in case of emergency (I have had some ideas to turn it off in dev so we see the errors). if MOIM breaks gemini and then we rely on fix_conversation, we're effectively doing the same thing as this PR.

tlongwell-block:
> hmm, I think we should make sure that fix_conversation can fix any obvious errors, but MOIM should not rely on that - ideally fix_conversation sits there as an in case of emergency (I have had some ideas to turn it off in dev so we see the errors). if MOIM breaks gemini and then we rely on fix_conversation, we're effectively doing the same thing as this PR.

100%. MOIM reports when errors occurs and we should fix them at the source. But afaik fix_conversation is *not* reporting errors here, though it apparently should be!

rabi:
Thanks @DOsinga @tlongwell-block for having a look! Here is what I've noticed..

MOIM injection between tool pairs passes through fix_conversation silently and it reaches gemin provider where it responds with an empty response. The reason is probably because merge_consecutive_messages uses effective_role, which treats user(tool_response) as "tool" and user(text) as "user". When MOIM (user text) is injected between a user(tool_response) and an assistant(tool_call), the merge step sees different effective roles and neither merges nor flags them. The rest of the pipeline also doesn't detect consecutive Role::User messages.

I also tried changing merge_consecutive_messages to use message.role instead of effective_role, which does merge the messages, but the resulting Content block mixes functionResponse and text parts, and Gemini echoes the MOIM text verbatim in its response.

owtaylor:
> Also, MOIM doesn't break tool call pairs. That should be checked for in fix_conversation anyway.

To be clear, the problem here is not breaking up a tool call pair, but rather breaking up the "current turn" in Google's terminology - (https://ai.google.dev/gemini-api/docs/thought-signatures) - the series of assistant messages and tool responses leading up to the final answer. Maybe there's some variation that would make Gemini OK with that, but it's not clear what that is or how we would find it. Given current information it should be avoided.

With:

```
<user message>
<tool call>
<tool response>
.... this goes on for thousands of tokens, 10s of minutes....
<tool call>
<tool response>
```

The only safe place to inject the MOIM for Gemini seems to be right at the beginning. But that might not be effective at the goals of MOIM (See: https://github.com/block/goose/pull/5027#issuecomment-3373363235.) Does this need to be some sort of provider flag?

jamadeo:
So, what is our conclusion here? @tlongwell-block @DOsinga I think either we decide the gemini-compatible ordering is sufficient for all uses of MOIM, or we have to make it's behavior provider-specific.

rabi:
> I think either we decide the gemini-compatible ordering is sufficient for all uses of MOIM, or we have to make it's behavior provider-specific.

I agree, let me know and I'll update the PR, if required.

michaelneale:
hey might close this, @rabi if it is still needed, please open again and lets get it to top of queue

owtaylor:
Nothing was fixed here, doesn't make sense to close it. The question in https://github.com/aaif-goose/goose/pull/7758#issuecomment-4041836430 still stands.

jamadeo:
A suggestion from @DOsinga is to prepend the MOIM to the last user message itself rather than create a new one. That seems reasonable to me. @tlongwell-block ?

PR #7764, opened 2026-03-09
feat(acp): add ACP methods to support programmatic goose clients

wpfleger96:
The ACP protocol was missing several capabilities needed by external clients that talk to a running `goose-acp-server` over the network.

Previously, clients like the goosed-slackbot had to use the HTTP API (`goose-server`) because ACP lacked a health check, a way to inject session instructions, and the ability to switch LLM providers at runtime. This closes those gaps.

- Adds `_goose/health` returning `{"status": "ok"}` — for clients that poll at startup before accepting traffic
- Adds `_goose/session/set_instructions` — injects a string into the agent's system prompt under the `"recipe"` key, equivalent to the `POST /agent/update_from_session` HTTP pattern but taking the final instructions string directly
- Enhances `session/set_model` with an optional `provider` field — when present, switches the full provider+model via the global registry (same pathOST /agent/update_provider`); empty string is treated as absent to avoid confusing registry errors
- Verifies `_goose/session/get` already returns token metrics (`input_tokens`, `output_tokens`, `accumulated_*`, `model_config`) via `Session` struct serialization; adds a test asserting their presence

### DIFF
--- a/crates/goose-acp/src/custom_requests.rs
+++ b/crates/goose-acp/src/custom_requests.rs
@@ -119,6 +119,24 @@ pub struct GetExtensionsResponse {
     pub warnings: Vec<String>,
 }
 
+/// Apply system prompt instructions to an active session.
+/// Equivalent to POST /agent/update_from_session in the HTTP API.
+#[derive(Debug, Deserialize, JsonSchema)]
+pub struct SetSessionInstructionsRequest {
+    pub session_id: String,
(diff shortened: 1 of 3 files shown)


lifeizhou-ap:
Hey @wpfleger96 👋 Closing this draft as part of some housekeeping on stale PRs. Feel free to reopen anytime if you pick this back up!

PR #7765, opened 2026-03-09
fix: remap max_completion_tokens to max_tokens for OpenAI-compatible providers

### LINKED ISSUE #7762: Request fails with 422: max_completion_tokens not permitted when using Mistral provider
**Describe the bug**

When using Goose with a Mistral provider, sending any message results in a `422 Unprocessable Entity` error. The Mistral API rejects the `max_completion_tokens` parameter, which Goose appears to send by default but is not supported by Mistral's API.

Error message:
```
Request failed: Request failed with status 422 Unprocessable Entity: {"object":"error","message":{"detail":[{"type":"extra_forbidden","loc":["body","max_completion_tokens"],"msg":"Extra inputs are not permitted","input":16384}]},"type":"invalid_request_error","param":null,"code":null}
```

---

**To Reproduce**
1. Configure Goose to use a Mistral provider/model
2. Start a Goose session (`goose`)
3. Send any message (e.g. `hello`)
4. See the 422 error

---

**Expected behavior**

Goose should either omit unsupported parameters like `max_completion_tokens` when calling the Mistral API, or map them to Mistral-compatible equivalents (e.g. `max_tokens`).

---

**Screenshots**

N/A

---

- **OS & Arch:** macOS (Apple Silicon)
- **Interface:** CLI & App
- **Version:** 1.27.2
- **Extensions enabled:** default
- **Provider & Model:** Mistral

---

**Additional context**

Mistral's API does not accept `max_completion_tokens` as a valid parameter. It appears Goose is sending this OpenAI-style parameter without filtering it for Mistral compatibility. The `input` value of `16384` in the error suggests this is a hardcoded default.

fresh3nough:
## Summary

When using OpenAI-compatible providers like Mistral via the declarative provider system, the OpenAI-specific parameter `max_completion_tokens` (used for O-series models) is not recognized and causes 422 errors.

## Root Cause

`OpenAiProvider` is used by both native OpenAI and declarative OpenAI-compatible providers (Mistral, Groq, etc.). The `create_request()` function sets `max_completion_tokens` for O-series models, but this parameter is not recognized by non-OpenAI APIs like Mistral, which expect `max_tokens` instead.

## Fix

Added `sanitize_request_for_compat()` method to `OpenAiProvider` that remaps `max_completion_tokens` to `max_tokens` for any non-native OpenAI provider (`name != "openai"`). Applied in the `stream` code path (the active chat completions path). If `max_tokens` already exists, the existing value is preserved.

## Validation

- 4 unit tests added and passing
- `cargo fmt` clean
- `cargo clippy` clean

Closes #7762

### DIFF
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -251,8 +251,22 @@ impl OpenAiProvider {
         Self::is_responses_model(model_name)
     }
 
+    /// Providers known to reject `max_completion_tokens` and require
+    /// the legacy `max_tokens` field instead.
+    const PROVIDERS_NEEDING_MAX_TOKENS_REMAP: &[&str] = &[
+        "cerebras",
+        "custom_deepseek",
+        "groq",
+        "inception",
+        "kimi",
+        "lmstudio",
+        "mistral",
+        "moonshot",
+        "ovhcloud",
+    ];
+
     fn sanitize_request_for_compat(&self, mut payload: serde_json::Value) -> serde_json::Value {
-        if self.name == OPEN_AI_PROVIDER_NAME {
+        if !Self::PROVIDERS_NEEDING_MAX_TOKENS_REMAP.contains(&self.name.as_str()) {
             return payload;
         }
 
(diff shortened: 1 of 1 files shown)


jamadeo:
(forgot to cite my sources)

https://github.com/badlogic/pi-mono/blob/4535415300f1599cd83807e1d1dd7f2615aabf70/packages/ai/src/types.ts#L239

https://github.com/badlogic/pi-mono/blob/4535415300f1599cd83807e1d1dd7f2615aabf70/packages/ai/src/providers/openai-completions.ts#L757

I do not know if pi is *the* thing to reference here, but that seems pretty thorough

fresh3nough:
Good call — narrowed it to Mistral/Groq for safety in this PR. Long-term I think we should move these mappings into the declarative config

PR #7772, opened 2026-03-09
feat: load hints in nested subdirs

### LINKED ISSUE #5840: Dynamically load hints in subdirectories
Goose reads the hints in the working directory but does not read hints in subdirectories. This is problematic for working in monorepos.

E.g. If you run Goose from the repo-root folder, then Goose reads `repo-root/agents.md` on startup.
If you direct goose to make changes to the frontend in `repo-root/frontend`, Goose will start reading and editing files without reading `repo-root/frontend/agents.md`.

Claude code and cursor do read hints when they access files in subdirectories

I have a draft [PR](https://github.com/block/goose/pull/5759) that adds logic to dynamically load hints in subdirectories when Goose reads files in those subdirectories.
It also maintains a turn counter and prunes the hints from the context when Goose hasn't accessed files in the subdirectory in the last N turns.

I'm new to the codebase and leaned heavily on AI to produce the draft PR so I'd like some help assessing if this is the right approach.

alexhancock:
fixes #5840 - a new version of https://github.com/block/goose/pull/5759 that can also work when code mode is on

verified

* where `.goosehints` says to "say bonjour at the end of responses"
* where `nested/.goosehints` says to "say adieu at the end of responses"
* by running:

```
16:13:51 ~/Development/test-ground $ ~/Development/goose/target/release/goose run --text "read nested/doc.md and tell me what it says"

    __( O)>  ● new session · databricks goose-claude-4-6-opus
   \____)    20260309_27 · /Users/alexhancock/Development/test-ground
     L L     goose is ready

  ▸ shell
    command: cat nested/doc.md

HELLOThe file `nested/doc.md` simply contains the text:

**HELLO**

That's it — just a single word greeting. Nothing else in the file.

Bonjour! Adieu! 👋
```

### DIFF
--- a/crates/goose/src/hints/load_hints.rs
+++ b/crates/goose/src/hints/load_hints.rs
@@ -10,6 +10,156 @@ use crate::hints::import_files::read_referenced_files;
 pub const GOOSE_HINTS_FILENAME: &str = ".goosehints";
 pub const AGENTS_MD_FILENAME: &str = "AGENTS.md";
 
+pub fn get_context_filenames() -> Vec<String> {
+    use crate::config::Config;
+
+    Config::global()
+        .get_param::<Vec<String>>("CONTEXT_FILE_NAMES")
+        .unwrap_or_else(|_| {
+            vec![
+                GOOSE_HINTS_FILENAME.to_string(),
+                AGENTS_MD_FILENAME.to_string(),
+            ]
+        })
+}
+
+#[derive(Default)]
+pub struct SubdirectoryHintTracker {
+    loaded_dirs: HashSet<PathBuf>,
+    pending_dirs: Vec<PathBuf>,
+    hints_filenames: Vec<String>,
+}
+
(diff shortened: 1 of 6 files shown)


crates/goose/src/hints/load_hints.rs
```diff
@@ -10,6 +10,163 @@ use crate::hints::import_files::read_referenced_files;
...
+/// Tracks which subdirectories have been accessed during a session and loads
+/// hint files from those directories on demand.
+#[derive(Default)]
+pub struct SubdirectoryHintTracker {
+    loaded_dirs: HashSet<PathBuf>,
+    pending_dirs: Vec<PathBuf>,
+    hints_filenames: Vec<String>,
+}
+
+impl SubdirectoryHintTracker {
```
DOsinga, line 33: can we add this to the prompt manager? right now it ends up being part of the agent struct and it feels like the agent should really not need to worry about this.

PR #7775, opened 2026-03-09
chore: utoipa upgrade to 5

alexhancock:
related to https://github.com/block/goose/issues/7008, but doesn't completely resolve it

we will still need to have `pctx` make a change as well


jh-block:
see also #7665

alexhancock:
closing in favor of the more comprehensive https://github.com/block/goose/pull/7665!

Issue #4307, opened 2025-08-23
Reference environment variable in http extension header value


giacomochiarella:
**Describe the bug**
I'm not sure if this is a bug. I would expect so just because otherwise it does not make any sense in my head to have environment variables in http extensions.
It looks impossible to reference an environment variable in an header value in http extension
I want to be able to add an environment variable (which could be a secret), let's name it MY_VAR, and then add a new header with MY_VAR referenced in the value of the header like $MY_VAR or ${MY_VAR} or {{MY_VAR}} or %MY_VAR%.
I've tried all of the above and none worked.

**To Reproduce**
1. Add a new Goose http extension
2. Add a new environment variable
3. Add a new header
4. Reference the environment variable
5. enable it
6. add a log into your mcp serverthe log will print the variable name instead of the variable value

**Expected behavior**
I expect the environment variable name is resolved in the header value when referenced

**Please provide following information:**
 - **OS & Arch:** Mac OS Sequoia
 - **Version:** 1.6.0
 - **Provider & Model:** Google - gemini-2.5-flash

giacomochiarella:
I've just noticed environment variable resolution does not work also in the endpoint field. I tried this url
http://127.0.0.1:11000/mcp?par1=$MY_VAR
and what I get in the mcp server is the string $MY_VAR, not the actual value stored in MY_VAR env var

DOsinga:
@amed-xyz - does this seem reasonable?

amed-xyz:
I think this can be split into a bug and a feature:

1. Bug: while one can specify `ENV` variables in the UI for `HTTP` extensions, they're actually [never used](https://github.com/block/goose/blob/6b6b71f3b4540ef7daf6d522c337bb9cd81de73b/crates/goose/src/agents/extension_manager.rs#L348-L354). We actually don't have a known use case for that yet, so we should remove it from the UI since it's misleading.
2. Feature: we don't support variable interpolation on either of `HTTP`, `ENV` or `URL`. I suspect mainly because we haven't seen a use case yet, but could make sense supporting it.

giacomochiarella:
@amed-xyz if I want to add a token in the url (like an api key) currently I need to set it up in the url as plain string, instead, env var would hide the value (as far as I saw env var values are hidden once they are set)

jh-block:
This is now [implemented](https://github.com/block/goose/pull/7782)

PR #7782, opened 2026-03-10
Support secret interpolation in streamable HTTP extension URLs

jh-block:
## Summary

Allow users to store secret values (API keys, tokens) in the system keyring and interpolate them into streamable HTTP extension URLs and headers using `${VAR}` or `$VAR` syntax. Addresses the feature request in #4307.

### Changes
- **Backend**: Apply `substitute_env_vars()` to the URI (previously only applied to headers) in both the config resolution path and runtime launch path
- **UI**: Show the Environment Variables section for streamable HTTP extensions (previously only shown for stdio extensions), enabling users to add and manage keyring-stored secrets
- **Fix recursive expansion**: `substitute_env_vars()` now scans the original input for `$VAR` patterns instead of the post-substitution output, preventing double-expansion when a secret value contains `$OTHER_VAR` syntax
- **Detect secret rotation**: Store a resolved config snapshot (with secrets substituted) in memory at client-creation time, so that `add_extension` detects when a secret value has changed even if `env_keys` and URI are unchanged
- **Tests**: Added test cases for URI variable substitution and recursive expansion prevention

Secrets are stored securely in the system keyring (macOS Keychain, Linux Secret Service, Windows Credential Manager) and resolved at runtime — they never appear in plaintext in config.yaml or on disk.

### Type of Change
- [x] Feature
- [x] Bug fix

### AI Assistance

### Testing
All 34 extension-related tests pass, including new tests for URI substitution (`http_env_key_uri_substitution`) and recursive expansion prevention (`test_substitute_env_vars_no_recursive_expansion`).

### Related Issues
Relates to #4307

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -60,6 +60,11 @@ static RE_ENV_SIMPLE: Lazy<regex::Regex> =
 
 struct Extension {
     pub config: ExtensionConfig,
+    /// Resolved config snapshot (with secrets from keyring substituted)
+    /// captured at client-creation time. Used to detect secret rotation
+    /// without re-reading the keyring on every comparison. Only held in
+    /// memory — never serialized to disk.
+    resolved_config: ExtensionConfig,
 
     client: McpClientBox,
     server_info: Option<ServerInfo>,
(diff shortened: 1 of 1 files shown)


DOsinga:
not sure we should support two ways for substitutions - or do we do that already elsewhere?

jh-block:
do you mean ${VAR} vs $VAR? It's just consistent with what the shell does I guess. We're already doing it for env var substitution into headers (which was pre-existing, this just does it in the URL as well)

DOsinga:
eh, if we use both patterns already for the other thing I guess that ship has sailed

PR #7784, opened 2026-03-10
Fix model selector showing wrong model in tabs

### LINKED ISSUE #7402: Desktop: Switching model in one tab incorrectly updates model display in all tabs
**Describe the bug**

When switching models in one session tab, the model selector UI incorrectly updates to show the new model in all other open session tabs, even though each session continues to use its original model.

---

**To Reproduce**

1. Open Desktop and ask goose to give you the AGENT_SESSION_ID from 2 chat session tabs
2. Check the database to confirm the model that's being used in the two sessions (in this example, `goose-claude-4-5-opus`):
   ```bash
   sqlite3 ~/.local/share/goose/sessions/sessions.db "SELECT id, provider_name, model_config_json FROM sessions WHERE id IN ('20260219_23', '20260220_10');"
   ```
3. In tab 1, change the model (in this example, `databricks-claude-haiku-4-5`)
4. Switch back to tab 2
5. Observe that the model selector in both tab 1 and 2 show `databricks-claude-haiku-4-5`
6. Check the database again to confirm the model that's being used in the two sessions:
   - Tab 1 is `databricks-claude-haiku-4-5`
   - Tab 2 is `goose-claude-4-5-opus`

The database check confirms that tab 2's session is still using its original model, but the UI incorrectly displays the changed model from tab 1.

---

**Expected behavior**

Each session tab should display its own model in the model selector, independent of model changes in other tabs.

---

**Screenshots**

N/A

---

**System Information**

- OS: macOS
- goose Version: [current version]
- Interface: Desktop

---

**Additional context**

**Root cause:** The `ModelAndProviderContext` React context state (`currentModel` and `currentProvider`) is global across all tabs. When `changeModel()` is called, it updates this global state (lines 70-71 in `ui/desktop/src/components/ModelAndProviderContext.tsx`), causing all tabs to display the same model name.

**Code reference:** `ui/desktop/src/components/ModelAndProviderContext.tsx` lines 43-85, specifically lines 70-71 where `setCurrentProvider()` and `setCurrentModel()` update global state.

**Impact:** This is a UI display bug only - sessions actually maintain their own models correctly in the database and use the correct model for API calls. The bug creates user confusion about which model each session is using.

jh-block:
## Summary
Fixed the model selector display being incorrectly synchronized across all tabs when changing a model in one tab. The bug occurred because the global `ModelAndProviderContext` state was being updated whenever any session's model changed, causing all tabs to display the same model name regardless of their actual session models.

### Type of Change
- [x] Bug fix

### Key Changes
- Made model display per-session: each tab now shows its own session's model
- Passed `sessionModel`/`sessionProvider` as props through the component tree instead of reading from global context
- Added local override state in `ModelsBottomBar` for immediate UI feedback before session re-fetches
- Only update the config default when `changeModel()` is called without a `sessionId` (existing chats don't affect defaults for new chats)

### Testing
- All existing tests pass (4 pre-existing failures in App.test.tsx unrelated to this change)
- TypeScript type checking passes
- Component tree data flow verified

### Related Issues
Fixes #7402

### AI Assistance

### DIFF
--- a/ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
+++ b/ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState, useCallback } from 'react';
+import { useEffect, useState, useCallback, useRef } from 'react';
 import { Bot, ExternalLink } from 'lucide-react';
 
 import {
@@ -277,18 +277,18 @@ export const SwitchModelModal = ({
     }
   }, [usePredefinedModels, currentModel]);
 
-  // For manual mode: sync provider/model state when session data arrives
-  // after the modal has already mounted (only if user hasn't changed them).
-  // Skip when initialProvider forces a different provider — model is
-  // intentionally empty so the user picks one for that provider.
+  // For manual mode: one-time sync of provider/model when session data
+  // arrives after the modal has already mounted. Uses a ref so it only
+  // fires once and doesn't interfere with user-driven changes (e.g.
+  // switching provider clears model intentionally).
+  const manualSyncDone = useRef(false);
   useEffect(() => {
-    if (usePredefinedModels) return;
+    if (usePredefinedModels || manualSyncDone.current) return;
     if (initialProvider && initialProvider !== currentProvider) return;
-    if (currentProvider && !provider) {
-      setProvider(currentProvider);
-    }
-    if (currentModel && !model) {
-      setModel(currentModel);
+    if (currentModel && currentProvider) {
+      if (!provider) setProvider(currentProvider);
+      if (!model) setModel(currentModel);
+      manualSyncDone.current = true;
     }
   }, [currentModel, currentProvider, usePredefinedModels, provider, model, initialProvider]);
 


ui/desktop/src/components/bottom_menu/CostTracker.tsx
```diff
@@ -15,10 +14,11 @@ interface CostTrackerProps {
       totalCost: number;
     };
   };
+  model?: string | null;
```
DOsinga: this shouldn't be optional (or nullable really) since how are we costtracking without a model/provider?

ui/desktop/src/components/settings/models/bottom_bar/ModelsBottomBar.tsx
```diff
@@ -24,15 +24,26 @@ interface ModelsBottomBarProps {
   dropdownRef: React.RefObject<HTMLDivElement>;
   setView: (view: View) => void;
   alerts: Alert[];
+  sessionModel?: string | null;
+  sessionProvider?: string | null;
+  onModelChanged?: (override: { model: string; provider: string }) => void;
```
DOsinga: same - that evenhandler is always set, so no need to be optional

ui/desktop/src/hooks/useCostTracking.ts
```diff
@@ -26,7 +25,8 @@ export const useCostTracking = ({
     };
   }>({});
 
-  const { currentModel, currentProvider } = useModelAndProvider();
+  const currentModel = session?.model_config?.model_name ?? null;
+  const currentProvider = session?.provider_name ?? null;
```
DOsinga: yeah so if we somehow don't have a model here, the whole thing falls apart. maybe just return undefined in that case? it shouldn't happen but that feels like a better data model

PR #7801, opened 2026-03-11
feat(acp): add session/set_mode handler

codefromthecrypt:
## Summary

The ACP server advertises available modes in `session/new` and `session/load` responses but has no `session/set_mode` handler. Zed's mode selector populates from these responses but mode changes are silently ignored. There are no tests for mode validation or error paths, and the server and provider test fixtures diverge: the server fixture uses raw ACP session IDs while the provider fixture uses goose session IDs, preventing shared test coverage.

This adds mode validation and aligns the test fixtures. `session/set_mode` validates the mode string and session existence, and rejects mode changes since per-session persistence is not yet implemented (#7603). The `ExpectedSessionId` trait lets server and provider tests share the same test functions. Both paths exercise identical error cases.

### Type of Change
- [x] Bug fix
- [x] Tests

### AI Assistance

#### Zed (ACP)

Kill goose and clear session/log state:
```bash
$ pkill -f goose
$ rm -rf ~/.local/state/goose/logs ~/.local/share/goose/sessions
$ rm -rf ~/Library/Logs/Zed/ ~/Library/Application\ Support/Zed/threads/
```

Build the release binary:
```bash
$ just release-binary
```

Add to `~/.config/zed/settings.json`:
```json
"agent_servers": {
  "goose": {
    "type": "custom",
    "command": "/path/to/goose-2/target/release/goose",
    "args": ["acp", "--with-builtin", "developer"],
    "env": {
      "RUST_LOG": "debug"
    }
  }
}
```

**Mode selection (expected rejection):**

Open Zed's agent panel. The mode selector (bottom bar, next to model selector) shows "auto" as the default.

1. Click the mode selector -- all four modes appear (auto, approve, smart_approve, chat) with descriptions
2. Switch to "approve"
3. Verify the rejection in goose logs:
```bash
$ grep -h "Mode change not supported" ~/.local/state/goose/logs/cli/*/*.log
```
```
{"timestamp":"...","level":"WARN","fields":{"message":"Sending error response","id":"Number(3)","error":"Error { code: -32602: Invalid params, message: \"Invalid params\", data: Some(String(\"Mode change not supported: session is auto, requested approve\")) }"},"target":"sacp::jsonrpc::outgoing_actor",...}
```

### Related Issues

Prerequisite for #7603

### Screenshots

Zed supports mode selection, so I used that.

#### Before

Goose has no mode selection, just model:

Screenshot 2026-03-13 at 1 18 51 PM

#### After

The selector briefly shows "approve" then reverts to "auto" -- Zed [optimistically updates][zed-optimistic] the UI, then [silently reverts][zed-revert] when the server returns an error. The error exists until modes are session scoped in #7603

mode selection but disabled

[zed-optimistic]: https://github.com/zed-industries/zed/blob/main/crates/agent_servers/src/acp.rs#L1029
[zed-revert]: https://github.com/zed-industries/zed/blob/main/crates/agent_ui/src/mode_selector.rs#L66-L67

### DIFF
--- a/crates/goose/src/acp/provider.rs
+++ b/crates/goose/src/acp/provider.rs
@@ -13,6 +13,7 @@ use sacp::{ClientToAgent, JrConnectionCx};
 use std::collections::{HashMap, HashSet};
 use std::path::PathBuf;
 use std::process::Stdio;
+use std::str::FromStr;
 use std::sync::{Arc, Mutex};
 use tokio::process::{Child, Command};
 use tokio::sync::{mpsc, oneshot, Mutex as TokioMutex};
@@ -43,10 +44,10 @@ enum ClientRequest {
     NewSession {
         response_tx: oneshot::Sender<Result<NewSessionResponse>>,
     },
-    SetModel {
-        session_id: SessionId,
-        model_id: String,
-        response_tx: oneshot::Sender<Result<()>>,
+    Untyped {
+        method: String,
+        params: serde_json::Value,
+        response_tx: oneshot::Sender<Result<serde_json::Value>>,
     },
     Prompt {
         session_id: SessionId,
@@ -77,7 +78,7 @@ enum AcpUpdate {
 pub struct AcpProvider {
     name: String,
     model: ModelConfig,
-    goose_mode: GooseMode,
+    goose_mode: Arc<Mutex<GooseMode>>,
     tx: mpsc::Sender<ClientRequest>,
     permission_mapping: PermissionMapping,
     rejected_tool_calls: Arc<TokioMutex<HashSet<String>>>,
@@ -107,8 +108,10 @@ impl AcpProvider {
         let (init_tx, init_rx) = oneshot::channel();
         let permission_mapping = config.permission_mapping.clone();
         let rejected_tool_calls = Arc::new(TokioMutex::new(HashSet::new()));
+        let goose_mode = Arc::new(Mutex::new(goose_mode));
 
-        tokio::spawn(run_client_loop(config, rx, init_tx));
+        let client_loop = AcpClientLoop::new(config, goose_mode.clone());
+        tokio::spawn(client_loop.spawn(rx, init_tx));
 
         let init_response = init_rx
             .await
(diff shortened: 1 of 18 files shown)


codefromthecrypt:
this PR is too big, so I will split it into two.

1. Persist goose mode to memory and DB on change (mode visibility fix for everything)
2. ACP support of session mode in general (this code has no idea about the DB and leaves only one patch for sync'ing)

codefromthecrypt:
fyi I rescoped this PR to the ACP prep part of mode support. I have a local branch ready to add session-scope mode (and persist) that will fix https://github.com/block/goose/issues/7603

This is a pre-req as doing both in the same change was too blurry, especially as a lot of the session persistence logic was plumbing to server and the Goose UI (not affecting ACP in really any way).

crates/goose/src/providers/base.rs
```diff
@@ -710,6 +711,10 @@ pub trait Provider: Send + Sync {
     ) -> bool {
         false
     }
+
+    async fn update_mode(&self, _session_id: &str, _mode: GooseMode) -> Result<(), ProviderError> {
```
codefromthecrypt, line 715: so this is a callback which agentic providers can implement to sync their mode for a given session. It is basically the same way permissions integration in ACP works.

crates/goose/src/acp/provider.rs
```diff
@@ -422,31 +455,161 @@ impl Drop for AcpProvider {
...
+            }
+        };
+
+        match self.run_with_child(child, &mut rx, init_tx).await {
+            Ok(()) => tracing::debug!("ACP protocol loop exited cleanly"),
+            Err(e) => tracing::error!(error = %e, "ACP protocol loop error"),
         }
     }
+
+    async fn run_with_child(
```
codefromthecrypt, line 493: this is mechanically moving around because things were getting hairy in the older layout, especially as we add things like mode and later other things.

codefromthecrypt:
will rebase after https://github.com/block/goose/pull/7843 which will drift it

codefromthecrypt:
side note: I notice the whole permissions integration is hairy, but there's a lot of code moving around with high frequency. I'd like to let that settle before refactoring, I think with some message types we can avoid some of the duplication mode/permission logic that is handled both in ACP internally and also somewhat similarly in the agent loops.

codefromthecrypt:
https://github.com/block/goose/pull/7854 is the mode mutablility/persistence thing split off from this. it keeps things less coupled.. when both of these are in, connecting acp to updatable mode will be less to review

Issue #7781, opened 2026-03-10
Weird problem with double quotes in shell on Windows platform.


0xfk0:
**Cannot use double quotes in any windows shell-command.**

See an example:
```
running findstr /n "class MainForm" *.cs

Tool Details
command
findstr /n "class MainForm" *.cs

Output
FINDSTR: Cannot open MainForm"

Command exited with code 1
```

Model tries to use distinct commands to do basic tasks, tries to use powershell. But if the command contains double quotes, command interpreted incorrectly and gives an error.

Looks like word-splitting on spaces exist despite of double quotes. Details unknown.

Due to this bug shell on Windows is mostly unusable. Model can spend up to ten steps to do primitive action like reading a file. Frequently model finishes with writing python-script and running it, or something like this.

[diagnostics_20260310_1.zip](https://github.com/user-attachments/files/25856871/diagnostics_20260310_1.zip)

---

**To Reproduce**
1. Run Goose on windows
2. Ask a model to run the command `findstr /n "word1 word2" *.txt`
3. You will see error message like: `FINDSTR: Cannot open word2"`
4. Pay attention, shell command is split on spaces despite of existing double quotes.

---

**Expected behavior**

Such commands as shown in example must not give an error.

---

**Please provide the following information**
- **OS & Arch:** windows 6.2.9200 x86_64
- **Interface:** UI
- **Version:** 1.27.2
- **Extensions enabled:** developer
- **Provider & Model:** custom_deepinfra – zai-org/GLM-4.7

0xfk0:
I checked out this (https://github.com/fresh3nough/goose/pull/3) branch, and build it:
```
  419  2026-03-10 19:10:41  git checkout fresh3nough/fix/windows-shell-double-quotes-7781
  420  2026-03-10 19:11:05  cargo build --release
  421  2026-03-10 20:15:37  cd ui/desktop
  422  2026-03-10 20:15:39  npm install
  423  2026-03-10 20:41:02  cd ../..
  424  2026-03-10 20:41:41  cargo build --release -p goose-server
  426  2026-03-10 21:06:18  cd -
  429  2026-03-10 21:07:14  xcopy /b ..\..\target\release\goosed.exe src\bin\
  430  2026-03-10 21:07:31  npm run make -- --targets=@electron-forge/maker-zip
```

Now I give following command to the model:
```
Please run shell-command: `findstr /n "word1 word2" *.txt` and provide me results. If shell returns an error please stop immediatelly.
```

And the result is:
[...]

PR #7810, opened 2026-03-11
fix: avoid word splitting by space for windows shell commands (#7781)

0xfk0:
The problem: when running shell-commands on Windows, Goose splits command to words by spaces and when passes every word as a separate argument to command interpreter (cmd.exe).

For example, when model tries to call a command: `find /I "word1 word2" *.txt` this command passed to command interpreter as (each line is a separate argument):

  - `cmd.exe`
  - `/C`
  - `find`
  - `/I`
  - `"word1`
  - `word2"`
  - `*.txt`

Command interpreter (cmd.exe) just passes discrete arguments as is to `find` program. So instead of one argument `find` program accepts a string which is split into two argument. This causes following errors.

Generally, this bug prevents using Goose on Windows (majority of commmands issued by the model include double quotes). Due to this bug model cannot use PowerShell, cannot use quoted file names, etc...

The root cause of the issue is with how command line is constructed:

```rust
let mut command = tokio::process::Command::new("cmd");
command.arg("/C").arg(command_line);
```

Even though `command_line` is passed as a single argument, if it contains spaces and quotes, the way Windows processes command line arguments can still cause issues with quote parsing.

Solution:

here's a Windows-specific method in tokio:process::Command called `raw_arg`:

> __raw_arg__ Available on Windows only.
> Append literal text to the command line without any quoting or escaping.
> This is useful for passing arguments to `cmd.exe /c`, which doesn't follow
> `CommandLineToArgvW` escaping rules.

Now command line arguments for example shown above looks as:

  - `cmd.exe`
  - `/C`
  - `find /I "word1 word2" *.txt`

Issue fixed.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:

### DIFF
--- a/crates/goose/src/agents/platform_extensions/developer/shell.rs
+++ b/crates/goose/src/agents/platform_extensions/developer/shell.rs
@@ -266,7 +266,7 @@ fn build_shell_command(command_line: &str) -> tokio::process::Command {
     #[cfg(windows)]
     let mut command = {
         let mut command = tokio::process::Command::new("cmd");
-        command.arg("/C").arg(command_line);
+        command.arg("/C").raw_arg(command_line);
         command
     };
 


DOsinga:
thanks!

PR #7812, opened 2026-03-11
fix: retry on authentication failure with credential refresh

wpfleger96:
## Summary

When goosed gets a 403 from a provider (e.g., Databricks token expired after auto-rotation), it now invalidates the in-memory secrets cache, re-reads `secrets.yaml` from disk, and retries once. Previously, authentication errors were terminal — the session broke with "Invalid access token" and no recovery.

## Context

Databricks tokens auto-rotate every 30 days in AWS Secrets Manager. The Block internal Slackbot fetches the token at pod startup and writes `secrets.yaml`. Goosed caches it in memory forever via `Config::all_secrets()`. When the token rotates while pods are running, goosed uses a stale token and returns 403 to all users.

This fix works with a sidecar container that periodically refreshes `secrets.yaml` from AWS SM. The sidecar keeps the file fresh; this change makes goosed pick up the new credentials on auth failure.

## Changes

- Add `Provider::refresh_credentials()` to the `Provider` trait (`base.rs`) with a default `Err` return — providers opt in by overriding
- `DatabricksProvider` overrides `refresh_credentials()` to invalidate the secrets cache (`databricks.rs`)
- Make `DatabricksAuthProvider::get_auth_header()` re-read the token from `Config::global()` on each call instead of using the value baked in at construction — matches how the OAuth path already works, and means cache invalidation actually takes effect without rebuilding the provider
- Add auth retry to the provider retry layer (`retry.rs`): on `ProviderError::Authentication`, call `refresh_credentials()` and retry once if it succeeds
- Remove the previous agent-loop implementation of auth retry from `agent.rs`
- Make `invalidate_secrets_cache()` public on `Config` (`base.rs`)

### DIFF
--- a/crates/goose/src/providers/base.rs
+++ b/crates/goose/src/providers/base.rs
@@ -707,11 +707,6 @@ pub trait Provider: Send + Sync {
         ))
     }
 
-    /// Refresh cached credentials after an authentication failure.
-    ///
-    /// Providers that support credential refresh (e.g. re-reading a rotated
-    /// token from disk) should override this. The provider retry loop calls
-    /// this on `ProviderError::Authentication` before retrying once.
     async fn refresh_credentials(&self) -> Result<(), ProviderError> {
         Err(ProviderError::NotImplemented(
             "credential refresh not supported by this provider".to_string(),
--- a/crates/goose/src/providers/databricks.rs
+++ b/crates/goose/src/providers/databricks.rs
@@ -227,11 +227,8 @@ impl DatabricksProvider {
     }
 
     pub fn from_params(host: String, api_key: String, model: ModelConfig) -> Result<Self> {
+        let token_cache = Arc::new(Mutex::new(Some(api_key.clone())));
         let auth = DatabricksAuth::token(api_key);
-        let token_cache = Arc::new(Mutex::new(match &auth {
-            DatabricksAuth::Token(t) => Some(t.clone()),
-            _ => None,
-        }));
         let auth_method = AuthMethod::Custom(Box::new(DatabricksAuthProvider {
             auth: auth.clone(),
             token_cache: token_cache.clone(),


jh-block:
Hi @wpfleger96, thanks for the contribution! Can you address the Codex code review comment and then this looks good to merge.

crates/goose/src/providers/base.rs
```diff
@@ -700,6 +699,17 @@ pub trait Provider: Send + Sync {
         ))
     }
 
+    /// Refresh cached credentials after an authentication failure.
+    ///
+    /// Providers that support credential refresh (e.g. re-reading a rotated
+    /// token from disk) should override this. The provider retry loop calls
+    /// this on `ProviderError::Authentication` before retrying once.
```
DOsinga: remove the comment

crates/goose/src/providers/databricks.rs
```diff
@@ -140,8 +160,15 @@ impl DatabricksProvider {
             DatabricksAuth::oauth(host.clone())
         };
 
-        let auth_method =
-            AuthMethod::Custom(Box::new(DatabricksAuthProvider { auth: auth.clone() }));
+        let token_cache = Arc::new(Mutex::new(match &auth {
+            DatabricksAuth::Token(t) => Some(t.clone()),
+            _ => None,
+        }));
```
DOsinga, line 166: let token_cache = Arc::new(Mutex::new(Some(api_key.clone())));

crates/goose/src/providers/databricks.rs
```diff
@@ -112,6 +130,8 @@ pub struct DatabricksProvider {
     fast_retry_config: RetryConfig,
     #[serde(skip)]
     name: String,
+    #[serde(skip)]
+    token_cache: Arc<Mutex<Option<String>>>,
```
DOsinga, line 134: I don't think we serialize this, but should we skip?

wpfleger96: I think we need to bc `DatabricksProvider` derives `serde::Serialize` but `Arc<Mutex<Option<String>>>` doesn't implement it, so this fails to compile without the skip. similar to the other fields `api_client`/`retry_config`/etc

Issue #7463, opened 2026-02-24
bug: Cursor position offset in CLI input prompt on Linux (same class as #3364)


nabeelshirazi:
On Linux, the CLI input prompt has a cursor positioning offset that causes the first character of typed input to appear separated from the rest by a space. For example, typing "what" displays as "w hat".
This is the same class of issue as #3364 (Windows cursor shift), which was fixed in PR #4464 by stripping ANSI escape sequences on Windows. The Linux prompt still uses styled ANSI sequences and the 🪿 emoji, both of which can cause rustyline to miscalculate display width.

To Reproduce

1. Run goose (interactive mode) on Linux
2. Type any message at the prompt
3. Observe the first character is visually separated from the rest of the input

Expected behavior
Text should appear contiguous as typed, with no spurious space after the first character.  Currently it looks like:
🪿w hat" instead of "what"

Environment

- OS & Arch: Linux 6.14.0-37-generic (x86_64)
- Interface: CLI (VS Code integrated terminal / Cursor IDE)
- Version: 1.25.0
- Provider & Model: lemonade / Qwen3-Coder-30B-A3B-Instruct-GGUF

Additional context

- The Windows variant of this bug (#3364) was fixed in PR #4464 by conditionally stripping ANSI escape sequences from the rustyline prompt via cfg!(target_os = "windows").
- The root cause is in rustyline's width calculation for ANSI escape sequences and/or the 🪿 (U+1FABF) emoji. Upstream tracking: kkawakam/rustyline#562, pending fix: kkawakam/rustyline#890.
- A possible fix would be to extend the ANSI-stripping workaround from PR #4464 to Linux, or to wrap the styled portions of the prompt in rustyline's special markers (\x01/\x02) so it can correctly exclude them from width calculation.

PR #7819, opened 2026-03-11
Fix terminal UI

alexyao2015:
## Summary
This fixes a few small QoL issues with the current console. 

1) This fixes the strange cursor offset in #7463. #7813 attempted to fix this, but it still has this issue after this change. This change fixes it.
2) Ctrl+C out of `goose configure` causes the console cursor to disappear. This is also fixed by this change.

### Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Tested the two issues were fixed

### DIFF
--- a/crates/goose-cli/src/main.rs
+++ b/crates/goose-cli/src/main.rs
@@ -7,6 +7,18 @@ async fn main() -> Result<()> {
         eprintln!("Warning: Failed to initialize logging: {}", e);
     }
 
+    // Install a non-terminating SIGINT handler so that interactive prompts
+    // (e.g. cliclack) can clean up the terminal cursor when Ctrl+C is pressed.
+    // Without this, the `console` crate raises SIGINT on Ctrl+C and the default
+    // handler kills the process before cursor restoration can run.
+    // The loop ensures there is always an active listener so that SIGINT is never
+    // silently discarded after the first Ctrl+C.
+    tokio::spawn(async {
+        loop {
+            tokio::signal::ctrl_c().await.ok();
+        }
+    });
+
     let result = cli().await;
 
     if goose::otel::otlp::is_otlp_initialized() {
--- a/crates/goose-cli/src/session/input.rs
+++ b/crates/goose-cli/src/session/input.rs
@@ -400,7 +400,13 @@ fn parse_plan_command(input: String) -> Option<InputResult> {
 }
 
 fn get_input_prompt_string() -> String {
-    "🪿 ".to_string()
+    // Use CHA (Cursor Horizontal Absolute) escape to fix cursor positioning.
+    // The goose emoji (U+1FABF) is 2 columns per unicode-width (Unicode 16.0),
+    // but some terminals (e.g. xterm.js/VS Code) render it as 1 column using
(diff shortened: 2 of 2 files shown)


DOsinga:
the concern flagged by codex seems worth looking into - you are addressing a real issue, but maybe the fix is too broad? let me know and we'll get this in

alexyao2015:
Made changes. Testing on the terminal, this also resolves the issue

alexyao2015:
I went ahead and reverted to the original change. `configure_telemetry_consent_dialog` also caused the same issue where the cursor is lost if the user ctrl+c'd out of it. I don't think this should have any ill effect on other listeners as suggested by codex in the earlier review.

DOsinga:
Thanks for tackling both of these terminal quirks — the underlying issues (cursor disappearing on Ctrl+C, prompt offset in VS Code) are real and worth fixing.

That said, the Codex P1 on the SIGINT handler is a genuine concern that needs to be resolved before this can land. The current approach spawns an infinite `ctrl_c()` loop at process startup that consumes **all** SIGINT signals for the lifetime of the process. This means any long-running or stuck subcommand — network calls during `goose configure`, a hanging provider check, etc. — becomes uninterruptible via Ctrl+C, which is a regression from the default behaviour.

Our suggestion would be to **drop the SIGINT loop entirely** and instead look for a narrower fix — e.g. hooking into `cliclack`'s own cleanup path, or catching the signal only for the duration of the interactive prompt and then re-raising/propagating it. The cursor-restore issue is really a `cliclack` responsibility; if their prompt doesn't restore the cursor on signal receipt that's worth raising upstream with them.

The P2 on the `\x1b[3G` CHA escape is lower priority but also needs attention — baking an unconditional ANSI escape into the prompt string will render literal escape garbage on terminals without VT support (Windows console paths, etc.). It should be gated on a terminal capability check.

Marking `needs_human` for a closer look once the SIGINT approach is reconsidered.

alexyao2015:
I looked at this again. It seems like there would need to be a major refactor of the app to properly handle signals where a global handler is installed and propagates back down to goose. 

Its clear that this is the way cliclack is built and it's goose's responsibility to handle signals properly.
https://github.com/fadeevab/cliclack/blob/a296aa6c9827a353a914858f520d1273a4693c9b/src/lib.rs#L40-L41


It does appear that this fix does break exiting out of oauth handlers for example. The cliclack library itself enables exiting using `esc`. Maybe the easiest fix is to just add a note saying to press `esc` to exit.

alexyao2015:
This change is the what is recommended in the library examples to handle this issue. 

https://github.com/fadeevab/cliclack/blob/a296aa6c9827a353a914858f520d1273a4693c9b/examples/basic.rs#L13

DOsinga:
I think we should remove the changes to the ctrl-c handling for now

DOsinga:
closing since stale. happy to merge the thing without the ctrl-c changes

PR #7829, opened 2026-03-12
chore: remove redundant husky prepare script

sheikhlimon:
## Summary
The pre-commit hook is manually configured at the git root (`.husky/pre-commit`) and correctly runs lint-staged for ui/desktop. The prepare script in the subdirectory can't work (no .git folder there) and causes warnings.

### Testing
Tested manually


DOsinga:
thanks!

Issue #5959, opened 2025-12-03
`goose session list` does not play nicely in pipes


greggyb:
Running `goose session list | head -n1` results in a main thread panic d/t the closed pipe:

```shell
$ RUST_BACKTRACE=full goose session list | head -n1
Available sessions:

thread 'main' panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: Broken pipe (os error 32)
stack backtrace:
   0:     0x5d7a8ee5b6b2 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h46a716bba2450163
   1:     0x5d7a8ee89173 - core::fmt::write::h275e5980d7008551
   2:     0x5d7a8ee56453 - std::io::Write::write_fmt::h561a66a0340b6995
   3:     0x5d7a8ee5b502 - std::sys::backtrace::BacktraceLock::print::hafb9d5969adc39a0
   4:     0x5d7a8ee5dc52 - std::panicking::default_hook::{{closure}}::hae2e97a5c4b2b777
   5:     0x5d7a8ee5da55 - std::panicking::default_hook::h3db1b505cfc4eb79
   6:     0x5d7a8ee5e5f2 - std::panicking::rust_panic_with_hook::h409da73ddef13937
   7:     0x5d7a8ee5e39a - std::panicking::begin_panic_handler::{{closure}}::h159b61b27f96a9c2
   8:     0x5d7a8ee5bba9 - std::sys::backtrace::__rust_end_short_backtrace::h5b56844d75e766fc
   9:     0x5d7a8ee5e02d - __rustc[4794b31dd7191200]::rust_begin_unwind
  10:     0x5d7a8b66a750 - core::panicking::panic_fmt::hc8737e8cca20a7c8
  11:     0x5d7a8ee54f9c - std::io::stdio::_print::h915f3273edec6464
  12:     0x5d7a8b7afb55 - goose_cli::cli::cli::{{closure}}::h3a5a4fa94d5731c0
  13:     0x5d7a8b7fde61 - goose::main::{{closure}}::hb9c653b64795145e
  14:     0x5d7a8b869b67 - tokio::runtime::park::CachedParkThread::block_on::h5124da2fae34f125
  15:     0x5d7a8bced561 - tokio::runtime::context::runtime::enter_runtime::h260ba4e8f3dff308
[...]

PR #7858, opened 2026-03-13
fix: stop writing without error when hitting broken pipe for goose session list

lifeizhou-ap:
## Summary
fix Issue https://github.com/block/goose/issues/5959

### Change
created a small helper that treats stdout BrokenPipe as normal early exit, preventing panics when output is piped for `goose session list`

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

### Testing
Manual testing

### DIFF
--- a/crates/goose-cli/src/commands/session.rs
+++ b/crates/goose-cli/src/commands/session.rs
@@ -7,7 +7,7 @@ use goose::session::{generate_diagnostics, Session, SessionManager};
 use goose::utils::safe_truncate;
 use regex::Regex;
 use std::fs;
-use std::io::Write;
+use std::io::{self, Write};
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


lifeizhou-ap:
> this works but goose tells me we can also fix this in one sweet go with:
> 
> ```rust
> // In main(), before anything else:
> unsafe { libc::signal(libc::SIGPIPE, libc::SIG_DFL); }
> ```

Yeah, I saw the same suggestions.  However, it is in main() and will be applied to all the commands. I feel it is a bit risky especially for long-running interactive sessions.

Issue #7793, opened 2026-03-10
Goose Desktop crash: EXC_BREAKPOINT (SIGTRAP) in Electron Framework after ~22 hours uptime


cvigrassblock:
### Description
Goose Desktop (Electron app) crashed with `EXC_BREAKPOINT (SIGTRAP)` after running for approximately 22 hours. The crash occurred on the main thread (`CrBrowserMain`) inside the Electron Framework, appearing to involve V8 garbage collection / stream read handling.

### Environment
- **Goose Version:** 1.27.2-block.202603061938-3a15b
- **Electron Framework:** 40.7.0
- **OS:** macOS 26.3 (25D125)
- **Hardware:** Mac16,8 (MacBook Pro, Apple M4 Pro, 24 GB RAM)
- **Architecture:** ARM-64 (Native)

### Crash Details
- **Date/Time:** 2026-03-10 11:40:24 PDT
- **Launch Time:** 2026-03-09 13:50:57 PDT (~21h 49m uptime)
- **Exception Type:** `EXC_BREAKPOINT (SIGTRAP)`
- **Exception Codes:** `0x0000000000000001, 0x0000000111179d3c`
- **Termination Reason:** Namespace SIGNAL, Code 5, Trace/BPT trap: 5
- **Triggered by Thread:** 0 (CrBrowserMain, main thread)

### Crash Stack Trace (Thread 0 — Crashed)
```
Thread 0 Crashed:: CrBrowserMain Dispatch queue: com.apple.main-thread
0   Electron Framework    ares_dns_rr_get_ttl + 3719688
1   Electron Framework    ares_dns_rr_get_ttl + 3719688
2   Electron Framework    ares_dns_rr_get_ttl + 3719712
3   Electron Framework    ares_dns_rr_get_ttl + 3719740
4   Electron Framework    node::PrincipalRealm::messaging_deserialize_create_object() const + 347072
5   Electron Framework    fontations_ffi$cxxbridge1$BridgeMappingIndex$operator$sizeof + 4587960
6   Electron Framework    fontations_ffi$cxxbridge1$BridgeMappingIndex$operator$sizeof + 4587844
7   Electron Framework    v8::CppHeap::CollectGarbageInYoungGenerationForTesting(cppgc::EmbedderStackState) + 93636
8   Electron Framework    v8::SharedMemoryStatistics::SharedMemoryStatistics() + 66344
...
49  Electron Framework    v8::Function::Call(...) + 6672
[...]

PR #7862, opened 2026-03-13
hail mary to stop from crashing

DOsinga:
See the crash: https://github.com/block/goose/issues/7793

we crash after prolonged periods of time in the GC. this switches of maglev for macos & arm. it also refreshes all the windows if we wake up after > 8 hours.

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -10,6 +10,7 @@ import {
   MenuItem,
   net,
   Notification,
+  powerMonitor,
   powerSaveBlocker,
   screen,
   session,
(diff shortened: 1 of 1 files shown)


ui/desktop/src/main.ts
```diff
@@ -116,6 +117,14 @@ async function configureProxy() {
 
 if (started) app.quit();
 
+// Mitigate V8 TurboFan/Maglev crash after macOS system wake (EXC_BREAKPOINT / SIGTRAP).
+// V8's tiered compilation can queue deferred optimization tasks on the CFRunLoop.
+// After a long sleep these fire with stale CompilationDependencies, hitting a CHECK().
+// Disabling Maglev removes the intermediate tier that most aggressively defers work.
+if (process.platform === 'darwin' && process.arch === 'arm64') {
+  app.commandLine.appendSwitch('js-flags', '--no-maglev');
```
jamadeo, line 125: what are the other implications of this?

PR #7874, opened 2026-03-14
chore(ui): remove debug console.log statements from App.tsx

jeffa-block:
## Why

App.tsx contains 7 `console.log` statements used for development-time debugging of IPC events and lifecycle signals. These appear in DevTools in production builds and provide no value to end users.

## What

Remove all `console.log` calls from `ui/desktop/src/App.tsx`:
- reactReady signal notification
- Keyboard shortcut setup trace
- View change request logging
- New-chat event trace
- Initial message IPC event logging (received / processing / duplicate)

All `console.error` calls for actual error handling are **preserved**.

Also removes a now-empty `else if` branch that only contained a removed log statement.

## How to review

Single file, 15 lines deleted. Verify each removal was a `console.log` (not `console.error`), and that no logic was affected.

## Testing

- `npm run build` in `ui/desktop` — no compile errors
- Manual: open Goose desktop, verify IPC events (new chat, view switching, initial message) still work. The only difference is silence in DevTools console.

### DIFF
--- a/ui/desktop/src/App.tsx
+++ b/ui/desktop/src/App.tsx
@@ -412,7 +412,6 @@ export function AppInner() {
   const { addExtension } = useConfig();
 
   useEffect(() => {
-    console.log('Sending reactReady signal to Electron');
     try {
       window.electron.reactReady();
     } catch (error) {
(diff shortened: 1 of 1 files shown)


DOsinga:
Squashed into #8142

PR #7878, opened 2026-03-14
chore(ui): remove stale comment in CardHeader.tsx

jeffa-block:
## Why

A comment reading `// Remove the console.log completely` was left behind after the console.log it referenced was already removed in a previous commit.

## What

Remove the stale comment from `CardHeader.tsx` line 24.

## How to review

1 line deleted. The comment was the only change — no logic affected.

## Testing

No runtime impact — comment-only change.


DOsinga:
Squashed into #8142

PR #7880, opened 2026-03-14
chore(ui): remove debug console.log from SessionSharingSection

jeffa-block:
## Why

`SessionSharingSection.tsx` logs the `GOOSE_BASE_URL_SHARE` environment variable value to the browser DevTools console on every render. This leaks configuration to anyone with DevTools open.

## What

Remove 1 `console.log` statement from `ui/desktop/src/components/settings/sessions/SessionSharingSection.tsx`.

## How to review

Single file, 1 line deleted. The `envBaseUrlShare` variable is still read and used — only the debug log is removed.

## Testing

No runtime impact. Session sharing settings continue to work as before.


DOsinga:
Squashed into #8142

PR #7893, opened 2026-03-14
Some safety features for reviewing untrusted code

jamadeo:
The goal is to be able to run a code review against untrusted code. Doing so of course opens up the agent to prompt injection, so the features here are designed to limit the capabilities of the agent such that that isn't a problem.

Two features controlled by environment variables help do this:
- `GOOSE_SHELL_ALLOWED_COMMANDS`: when set, the developer shell tool is limited to running only these commands. We also don't `bash -c`, so there's no shell expansion.
- `GOOSE_SHELL_ENV_SCRUB`: anything set here will be removed from the shell tool's ENV, so you can prevent the agent from seeing e.g. API keys that are exported

This also has:
- an MCP server designed to make it possible for the agent to add comments and complete a code review without writing files
- a recipe that references this server and contains a code review prompt
- a GHA workflow that runs goose in an environment restricted using the new features and using the new recipe

There are probably still holes here but this is a start.

### DIFF
--- a/.github/recipes/code-review.yaml
+++ b/.github/recipes/code-review.yaml
@@ -29,8 +29,8 @@ prompt: |
   The base branch is checked out in the working directory.
   Use the tools you have to review the diff and examine the code changes and context.
   Use the code review tools to add feedback on specific parts of the diff.
-  When finished, call the `finish_review` tool. There's no need to add a summary, but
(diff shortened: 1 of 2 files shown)


jamadeo:
I think I'm going to try a different direction

PR #7902, opened 2026-03-15
fix: use platform-appropriate shell examples in code_execution tool

jeffa-block:
## Summary

The `execute` tool description includes example shell commands (`ls -la`, `cat ./README.md`) that are unavailable on Windows. When the model follows these examples on Windows, it generates failing commands.

## Changes

Extract the description into `execute_tool_description()` and use `cfg!(windows)` to substitute Windows equivalents (`dir /b`, `type README.md`).

### Before (Windows)
```typescript
const files = await Developer.shell({ command: "ls -la" });      // ❌ fails
const readme = await Developer.shell({ command: "cat ./README.md" }); // ❌ fails
```

### After (Windows)
```typescript
const files = await Developer.shell({ command: "dir /b" });           // ✅
const readme = await Developer.shell({ command: "type README.md" });  // ✅
```

Unix examples remain unchanged.

Companion to #7898 which fixes the same issue in the developer extension instructions.

### DIFF
--- a/crates/goose/src/agents/platform_extensions/code_execution.rs
+++ b/crates/goose/src/agents/platform_extensions/code_execution.rs
@@ -51,6 +51,61 @@ pub struct ExecuteWithToolGraph {
     tool_graph: Vec<ToolGraphNode>,
 }
 
+/// Build the execute tool description with platform-appropriate shell examples.
+fn execute_tool_description() -> String {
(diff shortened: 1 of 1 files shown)


DOsinga:
Closing — the inline tool description this PR modified has been extracted to the `pctx_code_mode` crate (`descriptions::tools` constants). The platform-appropriate shell examples would need to be addressed there instead.

PR #7906, opened 2026-03-15
fix: support OPENAI_BASE_URL as alias for OPENAI_HOST

### LINKED ISSUE #7904: openai provider 401 with xAI API despite valid key
## Problem

When using `GOOSE_PROVIDER=openai` with xAI's API (`OPENAI_BASE_URL=https://api.x.ai/v1`), goose returns 401 Unauthorized despite the API key being valid.

## Environment

- goose version: 1.27.2
- macOS Darwin 25.2.0 (arm64)
- xAI subscription active, key valid

## Steps to reproduce

```bash
export OPENAI_API_KEY="$XAI_API_KEY"    # valid xai-... key
export OPENAI_BASE_URL="https://api.x.ai/v1"
GOOSE_PROVIDER=openai GOOSE_MODEL=grok-4-fast-reasoning goose run -t "Say hello"
```

## Expected behavior

Goose connects to xAI and receives a response.

## Actual behavior

```
Authentication error: Authentication failed. Status: 401 Unauthorized.
Response: Incorrect API key provided: xai-Pv3C****BlhI.
```

## Proof the key works

The exact same key succeeds with curl:

```bash
curl -s https://api.x.ai/v1/chat/completions \
  -H "Authorization: Bearer $XAI_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"model":"grok-4-fast-reasoning","messages":[{"role":"user","content":"hello"}],"max_tokens":5}'
# 200 OK — works fine
```

Also works with tools in the request body (like goose would send):
```bash
curl -s https://api.x.ai/v1/chat/completions \
  -H "Authorization: Bearer $XAI_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"model":"grok-4-fast-reasoning","messages":[{"role":"user","content":"hello"}],"max_tokens":5,"tools":[{"type":"function","function":{"name":"test","description":"test","parameters":{"type":"object","properties":{}}}}]}'
# 200 OK — works with tools too
```

And `GET /v1/models` works:
```bash
curl -s https://api.x.ai/v1/models -H "Authorization: Bearer $XAI_API_KEY"
# 200 OK — lists all grok models
```

## Notes

- The key IS being passed correctly (shown in the error message, matches the xAI key)
- No keychain entry exists for goose/OPENAI_API_KEY
- `GOOSE_PROVIDER=openrouter` with the same model works fine as a workaround
- Goose config.yaml does not have any stored openai credentials
- Tested with both `grok-4-1-fast` and `grok-4-fast-reasoning` model names — same result
- The Rust openai SDK may be modifying the auth header or request in a way xAI rejects

## Related

- #7903 (Azure AI auth incompatibility — same pattern: valid key, 401 from goose)

vincenzopalazzo:
## Summary

- Adds `OPENAI_BASE_URL` as a fallback alias for `OPENAI_HOST` when configuring the OpenAI provider's base URL

## Why

The [official OpenAI Python SDK](https://github.com/openai/openai-python) uses `OPENAI_BASE_URL` as its standard environment variable for custom API endpoints. This convention is followed across the broader OpenAI-compatible ecosystem — LiteLLM, LangChain, and virtually every tool that wraps the OpenAI API all honor `OPENAI_BASE_URL`.

When users point Goose at third-party OpenAI-compatible APIs (xAI, Azure AI, etc.) using the widely-expected `OPENAI_BASE_URL`, Goose silently ignores it and falls back to `https://api.openai.com`. The user's valid API key is then sent to OpenAI's servers, resulting in a confusing 401 Unauthorized error that looks like a key problem but is actually a host mismatch.

`OPENAI_HOST` remains the primary parameter — `OPENAI_BASE_URL` is only checked as a fallback, so existing configurations are completely unaffected.

## Test plan

- [x] `cargo check -p goose` passes
- [x] Existing OpenAI provider tests pass
- [ ] Verify `OPENAI_HOST` still takes priority when both are set
- [ ] Verify `OPENAI_BASE_URL` is used when `OPENAI_HOST` is not set

Closes #7904

### DIFF
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -224,36 +224,36 @@ impl OpenAiProvider {
     /// Parse an `OPENAI_BASE_URL` value into `(host, base_path)`.
     ///
     /// The OpenAI SDK convention includes the version path in the URL
-    /// (e.g. `https://api.x.ai/v1`).  We split this into a scheme+authority
-    /// host and prepend the URL path to the default chat-completions
-    /// endpoint so that `build_url` produces the correct final URL
-    /// without duplicating path segments like `/v1/v1/…`.
+    /// (e.g. `https://api.x.ai/v1`), so `{base_url}/chat/completions` is
+    /// the chat endpoint.  We keep the full URL (including its path) as
+    /// the host for `ApiClient` and use `chat/completions` as the
+    /// base_path.  `ApiClient::build_url` joins host and path, giving
+    /// e.g. `https://api.x.ai/v1/chat/completions` without duplicating
+    /// segments.
+    ///
+    /// When the URL has no path (e.g. `https://api.openai.com`), we fall
+    /// back to the standard `v1/chat/completions` default.
     fn parse_base_url(base_url: &str, config: &crate::config::Config) -> Result<(String, String)> {
         let parsed = url::Url::parse(base_url)
             .map_err(|e| anyhow::anyhow!("Invalid OPENAI_BASE_URL '{}': {}", base_url, e))?;
 
-        let host = if let Some(port) = parsed.port() {
-            format!(
-                "{}://{}:{}",
-                parsed.scheme(),
-                parsed.host_str().unwrap_or(""),
-                port
-            )
-        } else {
-            format!("{}://{}", parsed.scheme(), parsed.host_str().unwrap_or(""))
-        };
+        // Use the full URL (scheme + authority + path) as host so that
+        // proxy path prefixes like /openai/v1 are preserved.
+        let host = base_url.trim_end_matches('/').to_string();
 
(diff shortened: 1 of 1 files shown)


DOsinga:
Thanks for the contribution! The idea of supporting `OPENAI_BASE_URL` as an alias is a good one.

However, this PR has several unaddressed Codex review comments — including at least one P1 regression: when neither `OPENAI_HOST` nor `OPENAI_BASE_URL` is set, the new code hardcodes `OPEN_AI_DEFAULT_BASE_PATH` and silently ignores any `OPENAI_BASE_PATH` the user may have configured. The original code always read that env var. There are also two other P1s around path-stripping edge cases.

Please address (or explain) the Codex comments before re-opening or submitting a new PR. See our policy: https://github.com/block/goose/blob/main/CONTRIBUTING.md#ai-code-reviews

Happy to see a revised version!

PR #7922, opened 2026-03-16
feat: show live context window usage during multi-turn sessions

dale-lakes:
## Summary

Adds a persistent status bar at the bottom of the terminal that shows context window usage in real-time during multi-turn tool-calling sessions.

```
  ━━━━━━╌╌╌╌╌╌╌╌╌╌╌╌╌╌ 30% 78k/262k tokens
```

## Problem

During long agentic sessions with many tool calls (e.g. codebase analysis with 30+ tool calls), there's no way to see how full the context window is getting. The existing `/context` command only works between turns at the REPL prompt, not during streaming multi-turn execution. Users only discover context overflow when the session fails.

## Implementation

- **New `AgentEvent::ContextUsage` variant** — emitted after each provider response with current token count and context limit
- **Persistent `ProgressBar`** in the existing `McpSpinners` `MultiProgress` — stays pinned at the bottom while output scrolls above
- **Color-coded** — green (<50%), yellow (50-85%), red (>85%) to quickly signal when compaction is approaching
- **Non-intrusive** — cleared when the session returns to the prompt; only appears during streaming

## Test plan

- [ ] Run a multi-turn tool-calling session and verify the bar appears and updates after each turn
- [ ] Verify color transitions from green → yellow → red as context fills
- [ ] Verify the bar clears when returning to the REPL prompt
- [ ] Verify non-streaming / single-turn responses don't show a stale bar
- [ ] Verify subagent and gateway paths handle the new event variant without errors

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -1070,6 +1070,9 @@ impl CliSession {
                         Some(Ok(AgentEvent::HistoryReplaced(updated_conversation))) => {
                             self.messages = updated_conversation;
                         }
+                        Some(Ok(AgentEvent::ContextUsage { used_tokens, total_tokens })) => {
+                            progress_bars.update_context(used_tokens, total_tokens);
+                        }
                         Some(Ok(AgentEvent::ModelChange { model, mode })) => {
                             if is_stream_json_mode {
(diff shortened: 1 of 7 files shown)


lifeizhou-ap:
Hey @dale-lakes 👋 Closing this draft as part of some housekeeping on stale PRs. Feel free to reopen anytime if you pick this back up!

PR #7931, opened 2026-03-16
fix(claude-code): remove incorrect agent_visible filter on user message

### LINKED ISSUE #7930: Claude Code provider drops user messages with agent_visible: false metadata
## Bug Description

The Claude Code CLI provider silently drops user messages when they have `agent_visible: false` in their metadata, resulting in empty requests being sent to Claude. This causes the CLI to appear unresponsive while the desktop app works fine.

## Root Cause

In `crates/goose/src/providers/claude_code.rs`, the `last_user_content_blocks` function at line 284 filters by `is_agent_visible()`:

```rust
fn last_user_content_blocks(&self, messages: &[Message]) -> Vec<Value> {
    let msgs = match messages.iter().rev().find(|m| m.role == Role::User) {
        Some(msg) => std::slice::from_ref(msg),
        None => messages,
    };
    let mut blocks: Vec<Value> = Vec::new();
    for message in msgs.iter().filter(|m| m.is_agent_visible()) {  // <-- BUG HERE
        // ...
    }
}
```

Since `msgs` contains only the single last user message, if that message has `agent_visible: false`, the filter removes it entirely, resulting in empty content blocks being sent to Claude.

## When Introduced

Commit `4abf91e72b` on February 10, 2026 (PR #7108 "fix: isolate claude-code sessions via stream-json session_id")

The refactor changed from `messages_to_content_blocks` (which iterated all messages) to `last_user_content_blocks` (which finds only the last user message). The `is_agent_visible()` filter was incorrectly preserved from the original function.

## Symptoms

- CLI appears to hang or not respond to user input
- Session shows `total_tokens: 0` 
- Desktop app works fine (uses `Message::user()` with default metadata where `agent_visible: true`)
- No obvious error messages in logs

## Proposed Fix

Remove the `.filter(|m| m.is_agent_visible())` from line 284. The visibility filtering made sense when iterating over conversation history, but not when we've already specifically selected the user's message to send.

```rust
for message in msgs.iter() {  // Remove the filter
```

## Affected Versions

v1.27.2 and likely all versions since February 10, 2026

blackgirlbytes:
## Summary

Fixes #7930 - Claude Code provider silently drops user messages in CLI

Reported in Discord by community member https://discord.com/channels/1287729918100246654/1287729920319033345/1483114312720584794

## Problem

The `last_user_content_blocks` function finds the last user message to send to Claude Code CLI, but then incorrectly filters it by `is_agent_visible()`. Since we explicitly selected this message to send, the visibility filter should not apply.

When a message has `agent_visible: false` in its metadata, the filter removes it entirely, resulting in **empty content being sent to Claude**. This causes the CLI to appear to hang with no response.

## Root Cause

This was introduced in commit 4abf91e72b (PR #7108, Feb 10, 2026) when refactoring from `messages_to_content_blocks` to `last_user_content_blocks`. The original function iterated over all messages (where filtering made sense), but the new function only processes the single last user message - the filter was incorrectly preserved.

## Fix

Remove the `.filter(|m| m.is_agent_visible())` since we explicitly want to send the user's input regardless of visibility metadata.

## Testing

- Existing tests pass
- The fix is a one-line change with clear semantics

### DIFF
--- a/crates/goose/src/providers/claude_code.rs
+++ b/crates/goose/src/providers/claude_code.rs
@@ -275,17 +275,15 @@ pub struct ClaudeCodeProvider {
 }
 
 impl ClaudeCodeProvider {
+    /// Build content blocks from the last user message only. The CLI maintains
+    /// conversation context internally per session_id.
     fn last_user_content_blocks(&self, messages: &[Message]) -> Vec<Value> {
-        let (msgs, filter_visibility): (&[Message], bool) =
-            match messages.iter().rev().find(|m| m.role == Role::User) {
-                Some(msg) => (std::slice::from_ref(msg), false),
-                None => (messages, true),
-            };
+        let msgs = match messages.iter().rev().find(|m| m.role == Role::User) {
+            Some(msg) => std::slice::from_ref(msg),
+            None => messages,
+        };
         let mut blocks: Vec<Value> = Vec::new();
-        for message in msgs
-            .iter()
-            .filter(|m| !filter_visibility || m.is_agent_visible())
-        {
+        for message in msgs {
             let prefix = match message.role {
                 Role::User => "Human: ",
                 Role::Assistant => "Assistant: ",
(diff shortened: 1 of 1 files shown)


blackgirlbytes:
/goose review this PR . Tell me if there's any issues you found in this change.

codefromthecrypt:
ps I added a commit to feed the bot, but I think we should soon plan getting off double-maintenance of both acp and non-acp variants of the claude provider. a lot of this hairy stuff should be in one place (in the acp adapter we use as a lib)

PR #7932, opened 2026-03-16
Add a code review step which uses a short-lived provider token

jamadeo:
## Summary
Adds a new code review step which proxies its LLM requests using a short-lived Github OIDC token. This lets us make provider calls from an untrusted context for e.g. code reviews (included in this PR) or live provider integration tests.

This PR also contains the source code for a cloudflare worker that can be used to do this proxying.

### DIFF
--- a/.github/workflows/code-review.yml
+++ b/.github/workflows/code-review.yml
@@ -102,6 +102,7 @@ jobs:
           GOOSE_PROVIDER: anthropic-oidc-proxy
           GOOSE_MODEL: ${{ vars.GOOSE_CODE_REVIEW_MODEL || 'claude-opus-4-6' }}
           OIDC_PROXY_URL: ${{ github.event.inputs.oidc_proxy_url || vars.OIDC_PROXY_URL }}
+          REVIEW_INSTRUCTIONS: ${{ github.event.inputs.review_instructions || 'Review the changes for correctness' }}
         run: |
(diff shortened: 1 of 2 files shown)


.github/scripts/pr-review-mcp.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env -S uv run --script
```
michaelneale, line 1: nice and brief! was wondering if a skill and gh cli could do this but this is more crisp

jamadeo: That is possible, but if goose is to invoke the gh cli directly, we need to give it a GITHUB_TOKEN, which would then be vulnerable to stealing via promp injection. So instead we give goose the least privilege/secrets possible (only the short-lived token for the LLM proxy) and have it stage the whole review to an artifact, which is later uploaded by a step with the right privilege.

We could just have goose write to a json file directly, but that felt like asking more of the LLM vs giving it tools.

jamadeo:
> I like it. Does this mean we can also retire the codex review bot which is starting to get really noisy and harmful?

Eventually that would be great, but right now I think the codex review is probably a lot better at general code reviews. My thinking was that we use this new method to add specific things to look for, and if we can get to a point where the prompts work well enough to do general reviews, that's a great too.

Issue #7934, opened 2026-03-16
Configurable OPENAI_FAST_MODEL


elijahsgh:
**Please explain the motivation behind the feature request.**
Goose will attempt to use gpt-4o-mini as a fast model in the openai provider. This is not configurable. When used with a non-OpenAI server, it consistently attempts (and fails, then defaults) back to the other model.

**Describe the solution you'd like**
I would like to make this configurable.

**Describe alternatives you've considered**
I don't believe there is an alternative.

**Additional context**
I created a small change to the openai provider [here](https://github.com/block/goose/pull/7935)

- [x] I have verified this does not duplicate an existing feature request

DOsinga:
thanks. we should probably do this through the custom providers and add it as a field to DeclarativeProviderConfig

elijahsgh:
I did my best with the help of goose but I can't get past the `just generate-openapi`
Fails for some reason with a MODULE_NOT_FOUND

PR #7935, opened 2026-03-16
Adding configurable OPENAI_FAST_MODEL param

elijahsgh:
## Summary

The OpenAI provider currently has a default fast model (gpt-4o-mini) that it attempts to use. When using the OpenAI provider with another provider, such as Lemonade Server, goose will consistently attempt to call this model even though it doesn't exist.

We should make this configurable.

### Type of Change

- [ ] Feature
- [ ] Bug fix
- [x] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### AI Assistance

The original 

### Testing

Manual build and retry.

### Related Issues
Relates to #7934
Discussion: NA

Before:  
NA
After:   
NA

### DIFF
--- a/crates/goose-server/src/routes/agent.rs
+++ b/crates/goose-server/src/routes/agent.rs
@@ -385,7 +385,7 @@ async fn resume_agent(
             }
         })?;
 
-    let extension_results = if payload.load_model_and_extensions {
+    let (extension_results, session) = if payload.load_model_and_extensions {
         let agent = state
             .get_agent_for_route(payload.session_id.clone())
             .await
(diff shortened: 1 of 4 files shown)


DOsinga:
thanks. we should probably do this through the custom providers and add it as a field to DeclarativeProviderConfig

DOsinga:
can you do the DCO check @elijahsgh

elijahsgh:
I'm going to close this because I goofed up the signing but also because I think this is misdirected.
There's definitely a problem here where using the OpenAI Provider uses a fast model, but there also seems to be an odd problem where using the custom provider sends double calls and is not configurable.

We definitely should be able to change this fast model (or disable it) across both types of providers, but I don't think this PR does that well.

PR #7951, opened 2026-03-17
fix: removed double dash in pnpm command

lifeizhou-ap:
## Summary
Removed the extra double dash that passed to pnpm.  

### Root cause
The double dash worked with npm. But with pnpm, when the command with double dash forwarded to electron command and the double dashes are still there.  This caused the extra arguments to be dropped.

### Changes
- Removed the double dash
- Added the bundle desktop intel in the canary so that we know the failure earlier rather than release stage
- Added bundle desktop intel in manual bundle for testing

### Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Refactor / Code quality
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Security fix
- [ ] Build / Release
- [ ] Other (specify below)

### Testing
Testing by CI


.github/workflows/bundle-desktop-manual.yml
```diff
@@ -17,3 +17,12 @@ jobs:
     with:
       signing: false
       ref: ${{ inputs.branch }}
+
+  bundle-desktop-intel-unsigned:
```
michaelneale, line 21: did you mean to add this one in?

lifeizhou-ap: The main issue is the `--` in the `pnpm run make -- ...`.

I accidentally found the macos intel did not bundled due to this issue while working on issue.   So in this PR I included the macos intel bundle as part of the canary so that we can find any issue earlier. 

> also I found why the rest works fine in canary with pnpm `--`
because the github agent are x64, so even with `--` without the actual arguments, they worked fine.

PR #7956, opened 2026-03-17
fix: unregister goosed child process's  listener

michaelneale:
this may stop crashes after very long sessions (to do with v8)

shows up after many hours, days as this: 

image


DOsinga:
I don't think so. We don't print on stdout I think other than the fingerprint - the logging goes to stderr and I disconnected that last week (should probably do this the same way)

and even so,  if I look at goosed's logs overnight there's almost nothing there. the LLMs keep going on about memory pressure (97Mb!). it can't hurt of course

michaelneale:
yeah 97 is nothing really ... but thought was worth a try

PR #7966, opened 2026-03-17
Revert message flush & test

DOsinga:
## Summary
https://github.com/block/goose/pull/7850 tried to keep messages in sync, but it now seems that was not needed

### Testing
Includes a test

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -1631,6 +1631,9 @@ impl Agent {
                                 Ok(should_retry) => {
                                     if should_retry {
                                         info!("Retry logic triggered, restarting agent loop");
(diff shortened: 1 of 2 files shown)


DOsinga:
@jh-block @codefromthecrypt @rabi -- the previous fix broke how we concat messages and in the process of fixing that, I decided to add a test and then that test showed that the fix wasn't needed, so I wonder how we reproduce the original problem.

FWIW, if I start goose ask it to list the files in the current folder and examine one by one whether they contain the word goose and then hit ctrl-c after the first toolcall, I can dump the session and see that it indeed got captured in the session.

codefromthecrypt:
oh wow, we ran around in circles, huh. thanks for the explainer test!

PR #7973, opened 2026-03-18
fix: add periodic health check to detect backend unavailability

kyledef:
## Summary

Adds a periodic health check that pings the goosed `/status` endpoint every 30 seconds when the session is idle, proactively detecting backend unavailability **between** user messages.

## Problem

The goose server sends heartbeat pings only **during active reply streams** (500ms interval). Between user messages — which could be minutes or hours — there is **zero keepalive**. This means:

- The TCP connection can silently die (macOS sleep/wake, network changes)
- The goosed process could crash without the frontend knowing
- The agent could be evicted from the LRU cache

The user only discovers the problem when they try to send their next message and it fails. Multiple users have reported sessions that "just hang" after periods of inactivity, requiring a full restart.

## Approach

Rather than attempting transport-level retry on `/reply` (which is unsafe — see #7972), this PR detects problems **early** so the user is informed before their next submit fails:

1. **Detect** — ping `/status` every 30s while idle
2. **Debounce** — require 3 consecutive failures to avoid false positives from transient blips
3. **Alert** — show a toast when the backend is confirmed unreachable
4. **Recover** — when the backend comes back, call `resumeAgent` to restore the agent (handles LRU eviction, server restarts, extension reloading)

The health check is **disabled during active streaming** since the server already sends 500ms heartbeat pings in that path.

## Changes

- **New `useHealthCheck` hook** (`ui/desktop/src/hooks/useHealthCheck.ts`, 96 lines):
  - Pings `/status` every 30 seconds when enabled
  - 3 consecutive failures → toast warning ("Backend unreachable")
  - Fires `onRecovered` callback when connection is restored
  - Properly cleans up intervals on unmount and when disabled

- **Integrated into `useChatStream`**:
  - Active when: session is loaded AND `chatState === Idle`
  - Disabled when: streaming (server heartbeats cover this)
  - On recovery: calls `resumeAgent` to restore the session silently

## What this does NOT do

- Does not retry `/reply` requests (not safe — see #7972 for rationale)
- Does not auto-resend user messages
- Does not add any backend changes — reuses the existing `/status` endpoint

## Relationship to #7972

| PR | Covers | Mechanism |
|----|--------|-----------|
| #7972 | Failures **during** submit | `resumeAgent` fallback + honest error messaging |
| **#7973 (this)** | Failures **between** messages | Periodic health check + proactive recovery |

Together they cover both failure windows without unsafe transport retry.

## Testing

- Verified balanced braces and correct TypeScript syntax
- Hook uses `useCallback` for stable references to avoid unnecessary re-renders
- Interval properly cleaned up on unmount and `enabled` toggle
- Rust backend builds cleanly (no server-side changes)

## Files Changed

- `ui/desktop/src/hooks/useHealthCheck.ts` — **NEW** health check hook
- `ui/desktop/src/hooks/useChatStream.ts` — integration with health check
- `ui/desktop/package.json` — patch version bump to 1.27.2

### DIFF
--- a/ui/desktop/package.json
+++ b/ui/desktop/package.json
@@ -1,7 +1,7 @@
 {
   "name": "goose-app",
   "productName": "Goose",
-  "version": "1.27.0",
+  "version": "1.27.2",
   "description": "Goose App",
   "engines": {
     "node": "^24.10.0",
--- a/ui/desktop/src/hooks/useChatStream.ts
+++ b/ui/desktop/src/hooks/useChatStream.ts
@@ -27,6 +27,7 @@ import {
 import { errorMessage } from '../utils/conversionUtils';
 import { showExtensionLoadResults } from '../utils/extensionErrorUtils';
 import { maybeHandlePlatformEvent } from '../utils/platform_events';
+import { useHealthCheck } from './useHealthCheck';
 
 const resultsCache = new Map<string, { messages: Message[]; session: Session }>();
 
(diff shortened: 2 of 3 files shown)


DOsinga:
Hey @kyledef, thanks for taking the time to put this together — periodic health checking to detect backend unavailability is a genuinely useful idea, and the overall structure of the hook is reasonable.

We're closing this for a couple of reasons:

1. **Multiple PRs opened in quick succession** — you opened several PRs on the same day without waiting for earlier ones to land. We ask contributors to submit one at a time and wait for feedback before opening more. See: https://github.com/block/goose/blob/main/CONTRIBUTING.md#pull-requests

2. **Unaddressed Codex review comments** — there are two P2 issues flagged in the automated review that haven't been addressed:
   - State (`consecutiveFailuresRef` / `wasUnreachableRef`) is not reset when the hook is disabled, so stale state from a previous session can trigger false warnings on the next enable.
   - `setInterval` can fire overlapping async `status()` calls if the server is slow to respond, leading to out-of-order resolution.

   Please address these before resubmitting. See: https://github.com/block/goose/blob/main/CONTRIBUTING.md#ai-code-reviews

If you'd like to pick this back up, we'd suggest focusing on one PR at a time and fixing the two issues above first. Happy to review again once those are sorted. Thanks!

PR #7978, opened 2026-03-18
feat: idempotent reply submission and resumable SSE streams

kyledef:
## Problem

Goose's `/reply` endpoint is a one-shot turn submission with no reconnection support. If the SSE connection drops mid-stream (sleep/wake, Wi-Fi handoff, network blip), the client has two bad options:

1. **Don't retry** — the response is lost and the session appears hung
2. **Retry the POST** — a new agent turn starts, duplicating tool calls (shell commands, file edits, MCP calls)

This is the root cause of the widely-reported "goose hangs after inactivity" issue. PRs #7972 and #7973 provide safe immediate mitigations (agent recovery + health checks), but they cannot solve the fundamental problem: there's no way to reconnect to an in-flight reply.

## Solution

Add protocol-level support for idempotent turn submission and resumable SSE streams.

### How it works

1. **Client generates a `reply_id`** (UUID) for each submit and includes it in the `ChatRequest` body
2. **Server tracks active replies** by `(session_id, reply_id)` with event buffering and broadcast
3. **On retry**, the server recognizes the duplicate `reply_id` and returns a reconnect stream that:
   - Replays all buffered events from the original reply
   - Live-tails new events via broadcast channel
   - Does NOT start a new agent turn
4. **SSE `id:` fields** are emitted on each event, enabling `Last-Event-ID` based reconnection
5. **Cleanup** happens 60s after the reply finishes, giving time for late reconnections

### What changed

**Backend (`crates/goose-server/`)**
- `ChatRequest`: new optional `reply_id` field (fully backward compatible)
- `ActiveReply` struct: event buffer, broadcast channel, monotonic event IDs, cancellation token
- `reply()` function: check for existing active reply before starting new turn
- `stream_event_with_active_reply()`: emit SSE `id:` fields, buffer events, broadcast to subscribers
- `build_reconnect_stream()`: replay buffered events + live-tail new ones
- `AppState`: `active_replies` HashMap for tracking in-flight replies

**Frontend (`ui/desktop/`)**
- `generateReplyId()`: uses `crypto.randomUUID()` with fallback
- All 3 `reply()` call sites pass `reply_id` in the body
- `sseMaxRetryAttempts: 3` — now safe because retries reconnect instead of re-executing
- `ChatRequest` type and OpenAPI schema updated

### Tests

| Layer | Tests | Status |
|-------|-------|--------|
| Backend unit | 7 new: ChatRequest deserialization (with/without/null reply_id), ActiveReply buffering, broadcast, finished flag, monotonic IDs | ✅ All passing |
| Integration | 1 new: start reply → abort mid-stream → reconnect with same reply_id → verify replay | ✅ Syntactically valid |
| Existing | `test_reply_endpoint` updated with `reply_id: None` | ✅ Passing |

## Backward Compatibility

Fully backward compatible. When `reply_id` is `None` (or omitted), behavior is identical to today:
- No active reply tracking
- No event buffering
- No SSE `id:` fields
- `sseMaxRetryAttempts` has no effect (same as before)

## Relationship to other PRs

| PR | Purpose | Status |
|----|---------|--------|
| #7972 | Safe agent recovery on submit failure + accurate disconnect messaging | Ready for review |
| #7973 | Periodic health check between messages | Ready for review |
| **This PR** | Protocol-level idempotent reply + resumable streaming | Draft |

#7972 and #7973 are safe to merge independently. This PR builds on the same problem space but is a larger protocol change that warrants more review.

## Future work

- [ ] `Last-Event-ID` header handling on reconnect (client already sends it; server needs to use it to skip already-received events)
- [ ] Memory pressure monitoring for event buffers on long-running replies
- [ ] Configurable buffer retention (currently 60s after finish)
- [ ] Consider ACP-style transport model for desktop chat long-term

### DIFF
--- a/crates/goose-server/src/routes/reply.rs
+++ b/crates/goose-server/src/routes/reply.rs
@@ -699,4 +699,112 @@ mod tests {
             assert_eq!(response.status(), StatusCode::OK);
         }
     }
+
+    mod active_reply_tests {
+        use super::*;
+
+        #[tokio::test]
+        async fn test_chat_request_deserializes_without_reply_id() {
+            let json = r#"{
+                "user_message": {"role": "user", "created": 0, "content": [{"type": "text", "text": "hello"}], "metadata": {"userVisible": true, "agentVisible": true}},
+                "session_id": "test-session"
+            }"#;
+            let request: ChatRequest = serde_json::from_str(json).unwrap();
+            assert!(request.reply_id.is_none());
+            assert_eq!(request.session_id, "test-session");
+        }
+
+        #[tokio::test]
+        async fn test_chat_request_deserializes_with_reply_id() {
+            let json = r#"{
+                "user_message": {"role": "user", "created": 0, "content": [{"type": "text", "text": "hello"}], "metadata": {"userVisible": true, "agentVisible": true}},
+                "session_id": "test-session",
+                "reply_id": "abc-123"
+            }"#;
+            let request: ChatRequest = serde_json::from_str(json).unwrap();
+            assert_eq!(request.reply_id, Some("abc-123".to_string()));
+        }
+
+        #[tokio::test]
+        async fn test_chat_request_deserializes_with_null_reply_id() {
+            let json = r#"{
+                "user_message": {"role": "user", "created": 0, "content": [{"type": "text", "text": "hello"}], "metadata": {"userVisible": true, "agentVisible": true}},
+                "session_id": "test-session",
+                "reply_id": null
+            }"#;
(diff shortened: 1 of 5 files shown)


jh-block:
Hi @kyledef, thanks for this contribution. I realise it's still draft but wanted to point to #7834 which I hoped to land soon, are there any significant differences between the approaches / should we combine them somehow?

We already plan to switch to ACP so #7834 is just to improve things in the meantime while we still use our custom protocol

kyledef:
Closing. Even with the updated done in the other session I still have problems with keeping and resuming state. However, I'm not in a position to solve it given the complexity.

PR #7992, opened 2026-03-18
fix: SSE reconnect followup - reset refs, prevent duplication, sort active requests

jh-block:
## Summary
Follow-up fixes to #7834 (SSE reconnect):

- **Reset request-tracking refs on session switch** — clear all active/pending/buffer refs in the session-load effect cleanup so switching sessions doesn't leak state from the previous session's in-flight request
- **Prevent overflow duplication** — clear active request state in `reloadConversation` so the SSE reconnect doesn't reattach and replay deltas on top of the freshly-loaded conversation after a buffer overflow
- **Sort ActiveRequests by ID** — use lexicographic sort on uuidv7 IDs to always pick the newest reply, not an arbitrary HashMap key order

### DIFF
--- a/ui/desktop/src/hooks/useChatStream.ts
+++ b/ui/desktop/src/hooks/useChatStream.ts
@@ -373,9 +373,11 @@ export function useChatStream({
   const reloadingConversationRef = useRef(false);
   const namePollingRef = useRef<ReturnType<typeof setTimeout> | null>(null);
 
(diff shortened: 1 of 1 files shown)


DOsinga:
Closing in favour of a fresh PR that addresses the remaining Codex P1 issues (React commit timing race and A→B→A generation counter). Thanks for the work on this — the session-switch cleanup and overflow guard approach are the right direction.

PR #7996, opened 2026-03-18
fix: strip message wrapper in CLI session title generation

### LINKED ISSUE #7990: CLI providers generate session titles from prompt wrapper instead of user content
## Bug Description

When using CLI-based providers (`gemini-cli`, `claude_code`, `codex`, `cursor_agent`), all new chat sessions are titled **`---BEGIN USER MESSAGES---`** (or a variation like `---BEGIN USER MESSAGE`) instead of a meaningful generated title.

## Root Cause

The bug is in the interaction between `generate_session_name()` in `base.rs` and `generate_simple_session_description()` in `cli_common.rs`.

**Step 1** — `generate_session_name()` (`base.rs` line ~662) wraps the user messages in markers before calling the LLM:

```rust
let user_text = format!(
    "---BEGIN USER MESSAGES---\n{}\n---END USER MESSAGES---\n\nGenerate a short title for the above messages.",
    context.join("\n")
);
let message = Message::user().with_text(&user_text);
```

**Step 2** — CLI providers short-circuit the LLM call via `is_session_description_request()` (which matches `"four words or less"` in the system prompt) and dispatch to `generate_simple_session_description()` instead.

**Step 3** — `generate_simple_session_description()` (`cli_common.rs` line ~43) naively takes the **first 4 words** of the first user message:

```rust
text.split_whitespace()
    .take(4)
    .collect::<Vec<_>>()
    .join(" ")
```

Since the first user message is the wrapped prompt (`"---BEGIN USER MESSAGES---\n<actual content>..."`), the first 4 words are `"---BEGIN USER MESSAGES--- <first_real_word>"` — producing the broken title.

## Affected Providers

All CLI-based providers that use `cli_common::generate_simple_session_description`:
- `gemini_cli.rs` (line ~211)
- `claude_code.rs` (line ~697)
- `codex.rs` (lines ~683, ~1157, ~1177)
- `cursor_agent.rs` (line ~337)

## Suggested Fix

In `cli_common.rs`, update `generate_simple_session_description` to extract the actual user content from within the markers before taking the first 4 words:

```rust
.map(|text| {
    // Extract actual content from between the session-naming markers if present
    let actual_text = if let Some(start_idx) = text.find("---BEGIN USER MESSAGES---\n") {
        let after = &text[start_idx + "---BEGIN USER MESSAGES---\n".len()..];
        if let Some(end_idx) = after.find("\n---END USER MESSAGES---") {
            &after[..end_idx]
        } else {
            after
        }
    } else {
        text.as_str()
    };

    actual_text
        .split_whitespace()
        .take(4)
        .collect::<Vec<_>>()
        .join(" ")
})
```

This fix is isolated to a single function and corrects the behavior for all affected providers simultaneously.

## Steps to Reproduce

1. Configure Goose to use any CLI-based provider (`gemini-cli`, `claude_code`, `codex`, or `cursor_agent`)
2. Start a new chat session
3. Observe that the session title is set to `---BEGIN USER MESSAGES---` instead of a descriptive title

extrasmall0:
CLI providers (gemini-cli, claude_code, codex, cursor_agent) all generate session titles like `---BEGIN USER MESSAGES---` because `generate_simple_session_description()` grabs the first 4 words from the raw message — which includes the wrapper that `generate_session_name()` adds around the actual user content.

This strips the `---BEGIN USER MESSAGES---` / `---END USER MESSAGES---` markers before extracting words, so titles actually reflect what the user said.

Single-file change, pretty straightforward.

Closes #7990

### DIFF
--- a/crates/goose/src/providers/cli_common.rs
+++ b/crates/goose/src/providers/cli_common.rs
@@ -3,6 +3,7 @@ use serde_json::Value;
 use super::base::{ProviderUsage, Usage};
 use super::errors::ProviderError;
 use crate::conversation::message::{Message, MessageContent};
+use crate::utils::safe_truncate;
 use rmcp::model::Role;
 
 pub(crate) fn extract_usage_tokens(usage_info: &Value) -> Usage {
@@ -75,7 +76,7 @@ pub(crate) fn generate_simple_session_description(
             if desc.is_empty() {
                 "Simple task".to_string()
             } else {
-                desc
+                safe_truncate(&desc, 100)
             }
         })
         .unwrap_or_else(|| "Simple task".to_string());


crates/goose/src/providers/cli_common.rs
```diff
@@ -50,10 +50,30 @@ pub(crate) fn generate_simple_session_description(
             })
         })
         .map(|text| {
-            text.split_whitespace()
+            // Strip the wrapper added by generate_session_name so we get
+            // the actual user content instead of "---BEGIN USER MESSAGES---".
+            let stripped = text
+                .strip_prefix("---BEGIN USER MESSAGES---")
```
DOsinga: this seems rather brittle - if we change the marker on the other side of this, it will break. at a minimum use constants, but it would be even better if there was just a flag on providers that tells them whether to use this stripped down algorithm or to do the full call and check that at the callers side.

extrasmall0: yeah good point, the string matching is fragile. let me refactor — I'll extract the markers into constants at minimum, and also look into whether we can avoid the wrapping altogether by the time it hits generate_simple_session_description.

extrasmall0:
refactored — markers are now constants (SESSION_NAME_PREFIX / SESSION_NAME_SUFFIX). also looked into skipping the wrap entirely but that touches the provider trait boundary so keeping it scoped for now. lmk if this looks better

PR #7999, opened 2026-03-18
Orchestration

DOsinga:
## Summary
As part of project Palmate, introduce a orchestration extension. default off & hidden

### DIFF
--- a/crates/goose-server/src/routes/session_events.rs
+++ b/crates/goose-server/src/routes/session_events.rs
@@ -1,5 +1,6 @@
 use crate::routes::errors::ErrorResponse;
 use crate::routes::reply::{get_token_state, track_tool_telemetry, MessageEvent};
+use crate::session_event_bus::RequestGuard;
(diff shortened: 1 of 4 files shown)


tlongwell-block:
Pushed a fix commit (f7a4f1021) addressing concurrency and cleanup issues in the orchestrator and session event paths:

- Self-send guard prevents deadlock when orchestrator messages its own session
- Parent cancel token wired into `send_message` via `tokio::select!`; child stream dropped before busy-token release to prevent overlapping replies
- Atomic `try_register_cancel_token` / `try_register_request` replace check-then-act patterns
- `CancelTokenGuard` and `RequestGuard` Drop impls ensure cleanup on panic or early return
- Relative `working_dir` resolved against parent session directory instead of server CWD
- `remove_session` now cancels active work before dropping the cancel token
- Removed non-atomic `register_cancel_token` to prevent reintroduction of the race

PR #8004, opened 2026-03-19
feat: upgrade ink for performance, and use stdio not http

michaelneale:
Don't use http, use studio as better, more secure (and means can have many running)

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -3,8 +3,5 @@
   "workspaces": [
     "acp",
     "text"
-  ],
-  "overrides": {
-    "react": "19.2.4"
-  }
+  ]
 }
--- a/ui/text/package.json
+++ b/ui/text/package.json
@@ -39,6 +39,9 @@
     "@block/goose-acp-server-linux-x64": "0.1.0",
     "@block/goose-acp-server-win32-x64": "0.1.0"
   },
+  "overrides": {
+    "react": "19.2.4"
+  },
   "devDependencies": {
     "@types/marked-terminal": "^6.1.1",
     "@types/node": "^25.2.3",


alexhancock:
> (and means can have many running)

worth thinking through what the limitations might be for the http version in this area, but I agree using stdio for tui makes sense

ui/package.json
```diff
@@ -3,5 +3,8 @@
   "workspaces": [
     "acp",
     "text"
-  ]
+  ],
+  "overrides": {
+    "react": "19.2.4"
```
alexhancock: why do we need this? Ink not depending on a version we like?

could this be in `ui/text/package.json` instead?

jamadeo:
architecturally stdio makes sense. I think this began life as a PoC for the http transport

crates/goose-acp/src/bin/server.rs
```diff
@@ -40,18 +36,40 @@ async fn main() -> Result<()> {
...
+        tracing_subscriber::registry()
+            .with(filter)
+            .with(
+                tracing_subscriber::fmt::layer()
+                    .with_target(true)
+                    .with_writer(std::io::stderr),
+            )
+            .init();
+
+        goose_acp::server::run(builtins).await
```
jamadeo: we now have two entrypoints for acp over stdio: `goose acp` and `goose-acp-server --stdio`

michaelneale: hrm... makes me thinks this is a duplicate maybe?

michaelneale:
@jamadeo or should we stick with http as GUI will use it, and keeps thing ready for having one daemon running ? (I can patch the http one so if launches free port and secret so it works like it is one biinary for now?)

crates/goose-acp/src/bin/server.rs
michaelneale: yeah switched to using goose acp

alexhancock:
> @jamadeo or should we stick with http as GUI will use it, and keeps thing ready for having one daemon running ? (I can patch the http one so if launches free port and secret so it works like it is one biinary for now?)

No big deal either way what transport the TUI uses for now. If there isn't a performance issue with using HTTP I think it's good to use it as a reference implementation, but not required.

I do think it would be good if `goose-acp` can do either, and then `npx @aaif/goose acp` (and `goose acp` if we want to keep that in the small rust based scripting cli we'll keep) becomes a literal passthrough to the server running in stdio mode.

So this PR looks good to go to me

PR #8010, opened 2026-03-19
feat: add delete button to apps UI

### LINKED ISSUE #7965: Ability to delete apps from the apps section - Goose Desktop (Windows)
Hi,

I've found myself searching the documentation, and the internet to no avail. Deleting session folders and app folder contents just to continuiously still find myself to have the same ca. 30 apps that iv'e made in the beginning after downloading and running Goose Desktop for the first time.

Naturally I wanted to get rid of all these 'tests'. But instead of being able to delete specific apps (no menu option available), nor being able to just delete specific apps from the apps folder, I ended up having to trash all "block / goose" folders - which perfectly deleted all apps - but of course deleted .. *all* apps, and in turn also *all settings*. 

Meaning I had to completely set up everything from zero again.

Way too late I've discovered I can prompt AI to have an app deleted. By giving it the app name and the removal request.

It should just be as simple as with deleting a chat session - I'd love to request a remove button on each of the generated apps, and allow to delete speciffic apps. And most importantly without the need to prompt an AI to delete something I've created.  (That feels a bit overkill for a simple task like this :))

I can't be the only one who would find this useful, so I hope this feature finds its way into one of the next updates.

Besides that - great product - love it so far ! 🥇

jh-block:
## Summary
Users can now delete custom apps directly from the Apps UI with a confirmation modal, instead of requiring chat commands. Implements a DELETE endpoint and corresponding UI button with the Trash icon.

### Testing
- Verify trash icon appears on custom app cards
- Click trash icon → confirmation modal shows
- Confirm deletion → app is removed and list refreshes
- Cancel → modal closes without changes

### Related Issues
Resolves #7965


DOsinga:
I'm not sure I follow this PR; the only apps that are deletable are apps generated by the apps platform extension and they are recognizeable by having apps in their mcp server, so the client could just detect that and only then show a delete button and then we just forward the /apps/delete request to the apps platform extensions (need to clean up the cache too of course) - right now it looks like the deleteable flag on apps platform extension is false?

jh-block:
> I'm not sure I follow this PR; the only apps that are deletable are apps generated by the apps platform extension and they are recognizeable by having apps in their mcp server, so the client could just detect that and only then show a delete button and then we just forward the /apps/delete request to the apps platform extensions (need to clean up the cache too of course) - right now it looks like the deleteable flag on apps platform extension is false?

`deletable` should be true for anything the user created (e.g. ask goose to build a calculator for you and it should appear with a delete button -- currently these are impossible to delete unless you prompt the agent to do it which is not really discoverable) and false for anything that was bundled in an extension (e.g. the two sample apps that come with the Apps extension)

DOsinga:
> > I'm not sure I follow this PR; the only apps that are deletable are apps generated by the apps platform extension and they are recognizeable by having apps in their mcp server, so the client could just detect that and only then show a delete button and then we just forward the /apps/delete request to the apps platform extensions (need to clean up the cache too of course) - right now it looks like the deleteable flag on apps platform extension is false?
> 
> `deletable` should be true for anything the user created (e.g. ask goose to build a calculator for you and it should appear with a delete button -- currently these are impossible to delete unless you prompt the agent to do it which is not really discoverable) and false for anything that was bundled in an extension (e.g. the two sample apps that come with the Apps extension)

yeah, I understand that, but to me, it seems the way should implement this, is to have the route call the apps extension delete app path (that's the only way it can be deleted) and then that can return a boolean or an error and if it correctly deletes the app, clean up the cache. no inside information needed beyond that. and in the front end only show the delete button if apps is in the mcp servers of the app

PR #8024, opened 2026-03-19
fix: prevent panics when truncating non-ASCII strings in agent log helpers

cgarrett-square:
## Summary

Two log helper functions in `agent.rs` used byte-index slicing (`&s[..N]`) to truncate strings, which panics at runtime if the cutoff byte falls in the middle of a multi-byte UTF-8 character (e.g. emoji, accented characters, CJK text).

This caused `tokio-runtime-worker` thread panics that killed the agent mid-session whenever a user message or tool response contained non-ASCII content near the truncation boundary.

**Affected sites:**
- `derive_call_reason` — truncated user message preview at byte 100
- `format_messages_for_log` — truncated tool response body at byte 2000

## Fix

Replace byte-index slicing with `char_indices().nth(N)` to find a safe UTF-8 character boundary before slicing. This is the standard Rust idiom for character-count-based truncation.

## Test plan

- [x] Added unit tests for both sites covering: ASCII truncation, short string (no truncation), and multi-byte input (emoji / CJK) that would have panicked before the fix
- [ ] Manually send a message containing emoji or non-ASCII characters and confirm the agent no longer panics mid-session

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -82,8 +82,9 @@ fn derive_call_reason(turns_taken: u32, conversation: &Conversation) -> String {
                 })
             })
             .unwrap_or_default();
-        let truncated = if user_text.len() > 100 {
-            format!("{}...", &user_text[..100])
+        let truncated = if user_text.chars().count() > 100 {
+            let end = user_text.char_indices().nth(100).map(|(i, _)| i).unwrap_or(user_text.len());
+            format!("{}...", &user_text[..end])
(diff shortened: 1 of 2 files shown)


DOsinga (commented):
my agent went ahead and was a bit too eager; still though the PR description talks about an issue introduced in this PR

lifeizhou-ap:
Hey @cgarrett-square 👋 Closing this draft as part of some housekeeping on stale PRs. Feel free to reopen anytime if you pick this back up!

PR #8026, opened 2026-03-20
feat(providers): Refactor LM Studio to native provider and fix model discovery/filtering

nelsonr462:
### Issues

1. When attempting to customize/setup LM Studio, due to the declarative config, no changes in `config.yaml` files were respected when attempting to use a custom port or other modifications to the current LM Studio provider.
2. When connected, not all models were being displayed, due to recommended filtering hiding models that may already be loaded into memory.

As a result of the above, attempting to use LM Studio for things like Lead/Worker configurations would also break and fall back to settings/models that were unrelated to what was already configured in LM Studio or even within the Goose configs/settings themselves.

Videos of the above issues are attached below:

- Video 1 shows that when attempting to go in and modify LM Studio, a non-skippable modal appears requesting an API key, when none is required. Even when populated, the modal does not close.
[Screencast_20260319_204341.webm](https://github.com/user-attachments/assets/d5408eff-305b-47f1-88cc-a905cfe45449)

- Video 2 shows that attempting to change settings related to your loaded LM Studio config (like lead/workers) yields only a partial list of what models are actually available to LM Studio. There is a brief snapshot of the logs from the server clearly showing a much larger payload being sent to Goose.
[Screencast_20260319_211808.webm](https://github.com/user-attachments/assets/f753be4a-57e1-49bc-8fed-5a98bd163170)

It would then follow that other local running model providers would run into the same issue, given that "Recommended filters" are being run for every provider by default.

### Changelog

- Replace declarative LM Studio configuration with a native Rust provider.
- Enable full configurability for LM Studio host, port, and timeout in the Goose UI.
- Implement model discovery for LM Studio via the /v1/models endpoint.
- Override RECOMMENDED model filtering for LM Studio and Ollama to ensure all local models appear in the UI without being mistakenly hidden by naming heuristics.
- Update relevant docs for environment variables related to LM Studio

## Summary
With a native provider, now LM Studio behaves like other provider configs, with a configurable endpoint/port, timeout, and an optional key.
image

Clicking submit now proceeds to a proper model selection screen with full list of locally-loadable models that are ready in LM Studio.
image

With these settings now being respected, Goose can now properly utilize multiple models loaded into memory for Lead/Worker models.

### Testing
- All unit/integration tests ran locally, passed
- Manual testing and validation with LM Studio and Goose running on Bazzite via Strix Halo-based machine.

### Related Issues
None.

Before:  (above)

After:   (above)

### DIFF
--- a/documentation/docs/getting-started/providers.md
+++ b/documentation/docs/getting-started/providers.md
@@ -35,7 +35,7 @@ goose is compatible with a wide range of LLM providers, allowing you to choose a
 | [GitHub Copilot](https://docs.github.com/en/copilot/using-github-copilot/ai-models) | Access to AI models from OpenAI, Anthropic, Google, and other providers through GitHub's Copilot infrastructure. **GitHub account with Copilot access required.** | No manual key. Uses [device flow authentication](#github-copilot-authentication) for both CLI and Desktop. |
 | [Groq](https://groq.com/)                                                   | High-performance inference hardware and tools for LLMs.                                                                                                                                                                   | `GROQ_API_KEY`                                                                                                                                                                      |
 | [LiteLLM](https://docs.litellm.ai/docs/) | LiteLLM proxy supporting multiple models with automatic prompt caching and unified API access. | `LITELLM_HOST`, `LITELLM_BASE_PATH` (optional), `LITELLM_API_KEY` (optional), `LITELLM_CUSTOM_HEADERS` (optional), `LITELLM_TIMEOUT` (optional) |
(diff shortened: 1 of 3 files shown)


DOsinga:
apart from https://github.com/block/goose/blob/main/CONTRIBUTING.md#getting-started -- this is not the way. if custom providers to fit your need, change that & add more options.

nelsonr462:
noted - also just saw https://github.com/block/goose/discussions/1344#discussioncomment-14541974

will try to fit this better into custom config, thanks for direction @DOsinga 

sorry, was just a bit overeager after finding this project :sweat_smile:

PR #8027, opened 2026-03-20
fix: handle mid-stream error events in OpenAI SSE streaming

### LINKED ISSUE #8021: OpenAI provider crashes on mid-stream error events from local inference servers
## Problem

When an OpenAI-compatible inference server (e.g., Exo, vLLM, SGLang) sends a mid-stream error event during SSE streaming, Goose crashes with:

```
Stream decode error: error decoding response body
```

This happens because `response_to_streaming_message` in `providers/formats/openai.rs` deserializes every `data:` line as a `StreamingChunk` (`serde_json::from_str`). When the server sends an error object like:

```
data: {"error": {"message": "Internal server error", "type": "InternalServerError", "code": 500}}
```

...deserialization fails because there's no `choices` field, and the error propagates as `ProviderError::RequestFailed("Stream decode error: ...")`, killing the session.

## Expected behavior

Goose should check for an `error` key in parsed SSE data before attempting to deserialize as a `StreamingChunk`. This is the same pattern used by the official OpenAI Python client ([`openai-python/_streaming.py`](https://github.com/openai/openai-python/blob/main/src/openai/_streaming.py)), which checks `data.get("error")` and raises `APIError`.

## Context

Sending `{"error": {...}}` as an SSE `data:` event is the de facto standard for mid-stream errors across OpenAI-compatible servers:

- **OpenAI**: Their Python client explicitly handles this format
- **vLLM**: Sends `{"object":"error","message":"...","code":N}` as a data event
- **SGLang**: Sends `{"error":{"message":"...","type":"...","code":N}}` as a data event
- **Exo**: Sends `{"error":{"message":"...","type":"InternalServerError","code":500}}`

## Suggested fix

In `response_to_streaming_message` (around line 651 of `providers/formats/openai.rs`), before deserializing as `StreamingChunk`, parse the JSON and check for an `error` key. If found, return a `ProviderError` with the error message instead of a deserialization error.

dale-lakes:
## Summary

- Check for `error` key in SSE data events before deserializing as `StreamingChunk`
- Return a clear error message instead of crashing with "Stream decode error"

## Problem

When an OpenAI-compatible server sends a mid-stream error like `{"error": {"message": "Internal server error", ...}}`, `response_to_streaming_message` tries to deserialize it as a `StreamingChunk` and fails because there's no `choices` field. This propagates as `ProviderError::RequestFailed("Stream decode error: error decoding response body")`, killing the session.

This error format is used by vLLM, SGLang, Exo, and OpenAI itself. The official OpenAI Python client handles it by checking `data.get("error")` before deserialization.

## Fix

Added `parse_streaming_chunk()` that checks for an `error` key in the JSON before attempting `StreamingChunk` deserialization. Applied to both deserialization sites in `response_to_streaming_message`.

Closes #8021

### DIFF
--- a/crates/goose/src/providers/formats/openai.rs
+++ b/crates/goose/src/providers/formats/openai.rs
@@ -528,6 +528,20 @@ fn strip_data_prefix(line: &str) -> Option<&str> {
     line.strip_prefix("data: ").map(|s| s.trim())
 }
 
+fn parse_streaming_chunk(line: &str) -> anyhow::Result<StreamingChunk> {
+    if let Ok(value) = serde_json::from_str::<Value>(line) {
+        if let Some(error) = value.get("error") {
+            let message = error
+                .get("message")
+                .and_then(|m| m.as_str())
+                .unwrap_or("Unknown server error");
+            return Err(anyhow!("Server error during streaming: {}", message));
+        }
+    }
+    serde_json::from_str(line)
+        .map_err(|e| anyhow!("Failed to parse streaming chunk: {}: {:?}", e, line))
+}
+
 pub fn response_to_streaming_message<S>(
     mut stream: S,
 ) -> impl Stream<Item = anyhow::Result<(Option<Message>, Option<ProviderUsage>)>> + 'static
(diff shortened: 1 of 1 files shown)


DOsinga:
thanks @spitfire55 - in order to merge faster, we just push to branches these days. since yours is on a branch, I'll continue on #8301 and will close this

PR #8043, opened 2026-03-20
feat: bypass canonical model filtering for LM Studio

nelsonr462:
Directly addresses:
This enables LM Studio users to discover all available models without canonical registry filtering, as their model names are dynamic and don't match the canonical list.

This also allows for LM Studio users to define custom ports/host if needed via an env variable.

## Summary
- Add `skip_model_filtering()` trait method to Provider (defaults to false)
- Implement in OpenAiProvider to enable unfiltered models for no-auth providers
  - Set via `skip_model_filter: config.name == "lmstudio"` in `from_declarative()`
  - Only affects built-in LM Studio config, won't impact other no-auth/declarative configs
- Update LM Studio declarative config:
  - Dynamic `${LMSTUDIO_HOST}` env var instead of hardcoded URL
  - `dynamic_models: true` flag
- Show UI warning for LM Studio users about tool-calling compatibility
- LM Studio can now discover all available models without canonical registry filtering

### Testing
All unit/integration tests ran locally, passed
Manual testing and validation with LM Studio and Goose 

### Related Issues
Relates to: https://github.com/block/goose/pull/8026

Before:  

Model selection modal only displaying filtered results
image

After: 

Now showing unfiltered results from LMStudio, with warning
image

image

### DIFF
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -212,7 +212,7 @@ impl OpenAiProvider {
             custom_headers: config.headers,
             supports_streaming: config.supports_streaming.unwrap_or(true),
             name: config.name.clone(),
-            skip_model_filter: !config.requires_auth,
+            skip_model_filter: config.name == "lmstudio",
         })
     }
 


DOsinga:
Thanks for digging into this! The root cause is real — LM Studio's locally-served model names don't match the canonical registry so they all get filtered out.

We've opened a minimal fix in #8052 that adds a `skip_canonical_filtering` boolean directly to `DeclarativeProviderConfig` and sets it to `true` in `lmstudio.json`. This avoids the name-string hardcode and generalises cleanly to any future declarative provider with the same characteristic. Closing this one in favour of that.

PR #8052, opened 2026-03-21
fix: add skip_canonical_filtering to declarative providers

DOsinga:
## Summary

Declarative providers serving arbitrary local models (e.g. LM Studio) have model names that won't match the canonical registry, so the existing filtering in `fetch_recommended_models` returns an empty list.

## Details

Adds a `skip_canonical_filtering` boolean field to `DeclarativeProviderConfig` (defaults `false`, so no behaviour change for existing providers). When `true`, `fetch_recommended_models` returns all models from `fetch_supported_models` directly without consulting the canonical registry.

Sets `skip_canonical_filtering: true` in `lmstudio.json` so locally-served models are listed as-is.

This is a clean, minimal fix for the same problem raised in #8043 and #8001 — without name-string hardcoding or a new provider-type enum.

Closes #8043
cc #8001

### DIFF
--- a/ui/desktop/openapi.json
+++ b/ui/desktop/openapi.json
@@ -4633,6 +4633,9 @@
           "requires_auth": {
             "type": "boolean"
           },
+          "skip_canonical_filtering": {
+            "type": "boolean"
+          },
           "supports_streaming": {
             "type": "boolean",
             "nullable": true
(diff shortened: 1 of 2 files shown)


DOsinga:
@nelsonr462 and @octogonz - I think this does what we want to get done here? let me know if not.

PR #8054, opened 2026-03-21
fix(ci): produce .tar.gz archives for Zed ACP registry compatibility

### LINKED ISSUE #8047: Can't connect to Goose ACP in Zed
When attempting to use the Goose Agent ACP in Zed I get this error: 

unsupported archive type in URL: https://github.com/block/goose/releases/download/v1.28.0/goose-x86_64-unknown-linux-gnu.tar.bz2

However i already have v1.28.0 CLI installed and working with my local LLMs

I tried both methods to connect. 

- Registry:
I get the following error message:

Image

- Custom:
```
{
  "agent_servers": {
    "goose": {
      "type": "custom",
      "command": "goose",
      "args": ["acp"],
      "env": {},
    },
```
I get the following error message:

Image

I'm using Ubuntu 24.04 x86. I use Goose v1.28.0 with Ollama and Zed v0.228.0, and I have both the CLI and desktop versions of Goose. Thanks in advance for your support!

vincenzopalazzo:
## Summary

- Adds `.tar.gz` archive output alongside the existing `.tar.bz2` for Linux/macOS CLI builds
- Updates all release workflows (release, nightly, canary) to include `goose-*.tar.gz` in published artifacts

## Problem

Zed's agent server store ([`agent_server_store.rs`](https://github.com/zed-industries/zed/blob/main/crates/project/src/agent_server_store.rs)) only supports `.tar.gz`, `.tgz`, and `.zip` when downloading agent binaries:

```rust
let asset_kind = if archive_url.ends_with(".zip") {
    AssetKind::Zip
} else if archive_url.ends_with(".tar.gz") || archive_url.ends_with(".tgz") {
    AssetKind::TarGz
} else {
    anyhow::bail!("unsupported archive type in URL: {}", archive_url);
};
```

Goose currently only publishes `.tar.bz2` for Linux/macOS, causing Zed users to get `"unsupported archive type in URL"` when installing Goose via the ACP registry.

## Solution

Produce **both** `.tar.bz2` (backward compat) and `.tar.gz` (Zed compat) in the build pipeline. This is a minimal change — one extra `tar` command in `build-cli.yml` and glob additions in the release workflows.

Once merged and released, the ACP registry entry for Goose can be updated to point to the `.tar.gz` URLs.

## Test plan

- [ ] Verify CI builds produce both `.tar.bz2` and `.tar.gz` artifacts
- [ ] Verify release assets include both archive formats
- [ ] After ACP registry update, verify Zed can install Goose via registry method

Closes #8047

### DIFF
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -144,7 +144,6 @@ jobs:
           # Copy the goose binary
           cp "target/${TARGET}/release/goose" "target/${TARGET}/release/goose-package/"
 
-          # Create the tar archives (bz2 for backward compat, gz for Zed/ACP registry compat)
           cd "target/${TARGET}/release"
           tar -cjf "goose-${TARGET}.tar.bz2" -C goose-package .
           tar -czf "goose-${TARGET}.tar.gz" -C goose-package .
@@ -163,7 +162,7 @@ jobs:
 
           cd "target/${TARGET}/release"
           7z a -tzip "goose-${TARGET}.zip" goose-package/
-          echo "ARTIFACT=target/${TARGET}/release/goose-${TARGET}.zip" >> $GITHUB_ENV
+          echo "ARTIFACT_ZIP=target/${TARGET}/release/goose-${TARGET}.zip" >> $GITHUB_ENV
 
       - name: Upload CLI artifact
         uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f  # v6.0.0
@@ -172,4 +171,5 @@ jobs:
           path: |
             ${{ env.ARTIFACT_BZ2 }}
             ${{ env.ARTIFACT_GZ }}
-            ${{ env.ARTIFACT }}
+            ${{ env.ARTIFACT_ZIP }}
+
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -134,6 +134,7 @@ jobs:
         with:
           subject-path: |
             goose-*.tar.bz2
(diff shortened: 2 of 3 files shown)


aleclarson:
Note that tar.bz2 *is* supported by the ACP registry:
https://github.com/agentclientprotocol/registry/blob/e77c08070ae89d4e76a0169021bf9974ca2ca51c/agent.schema.json#L96

So I would consider this a bug in Zed's implementation, FWIW.

vincenzopalazzo:
Hey @aleclarson, good catch! You are right, the ACP registry schema does support `.tar.bz2` — but looking at the Zed codebase, there is no handler for it.

The archive URL matching in [`agent_server_store.rs`](https://github.com/zed-industries/zed/blob/main/crates/project/src/agent_server_store.rs) only handles `.zip`, `.tar.gz`, and `.tgz` — anything else bails with `"unsupported archive type"`.

The `AssetKind` enum in [`github.rs`](https://github.com/zed-industries/zed/blob/main/crates/http_client/src/github.rs) has no `TarBz2` variant, and the extraction logic in [`github_download.rs`](https://github.com/zed-industries/zed/blob/main/crates/http_client/src/github_download.rs) has no bzip2 decoder at all.

So yeah, producing both formats here is the pragmatic fix — it unblocks Zed users now. Might be worth opening an issue on `zed-industries/zed` to close this gap on their side too.

vincenzopalazzo:
In the meantime, I also fixed the zed bug https://github.com/zed-industries/zed/pull/52188, so now with the zed fix, we should have the full workflow completed

PR #8070, opened 2026-03-23
feat: add GOOSE_SHOW_FULL_OUTPUT config to disable tool output truncation

DOsinga:
Closes #7919

Wires up the existing `SHOW_FULL_TOOL_OUTPUT` thread-local (previously hardcoded to `false`) to read from config/env at initialization.

When `GOOSE_SHOW_FULL_OUTPUT=true` is set (via environment variable or config file), tool call arguments are displayed in full instead of being truncated to terminal width. Useful for debugging tool calls with large arguments.

Co-authored-by: Dale Lakes <6843636+spitfire55@users.noreply.github.com>


DOsinga:
Closing — approved the original PR #7919 directly instead.

PR #8073, opened 2026-03-23
Update tar version to avoid CVE-2026-33056

r0x0d:
## Summary

We received a CVE report in downstream builds for tar <= 0.4.44 and the fix is present in >= 0.4.45, and to avoid the version being stuck on an old version, this patch updates the Cargo.toml to have tar >= 0.4.45

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:

### DIFF
--- a/crates/goose-cli/Cargo.toml
+++ b/crates/goose-cli/Cargo.toml
@@ -48,7 +48,7 @@ shlex = "1.3.0"
 async-trait = { workspace = true }
 base64 = { workspace = true }
 regex = { workspace = true }
-tar = "0.4"
+tar = "0.4.45"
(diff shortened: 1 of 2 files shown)


crates/goose-cli/Cargo.toml
```diff
@@ -48,7 +48,7 @@ shlex = "1.3.0"
 async-trait = { workspace = true }
 base64 = { workspace = true }
 regex = { workspace = true }
-tar = "0.4"
+tar = ">= 0.4.45"
```
r0x0d:
```suggestion
tar = "^0.4.45"
```

Issue #3468, opened 2025-07-17
Feat: add keyring support for DataBricks


exitcode0:
**Please explain the motivation behind the feature request.**
I've noticed that the Databricks integration does not currently use the keyring for token storage, instead it writes tokens out to disk

**Describe the solution you'd like**
we should investigate keyring support for this integration to better protect the tokens

**Describe alternatives you've considered**
n/a

**Additional context**
There might be considerations for other Github Issues / feature requests, e.g. #3137 

- [x] I have verified this does not duplicate an existing feature request

michaelneale:
@exitcode0 is there a way to make keyring less annoying on macos? ? (not really relevant exactly to this, but wow... it is frustrating to use for me)

DOsinga:
Thanks for raising this @exitcode0. We've looked into this and can confirm:

- **PAT path** (`DATABRICKS_TOKEN`) already uses the system keyring via `config.get_secret()`, so that's covered.
- **OAuth path** is the real gap — the `TokenCache` in `oauth.rs` writes access/refresh tokens as plaintext JSON to `~/.config/goose/databricks/oauth/`, bypassing the keyring entirely.

This is a valid security improvement. The fix would involve refactoring `TokenCache` to use the existing `Config::set_secret`/`get_secret` infrastructure (which already handles keyring + file fallback + `GOOSE_DISABLE_KEYRING`), with token migration from existing files.

We're snoozing this for a few weeks to give the community a chance to weigh in on priority and any design preferences (e.g. the macOS keyring UX concern raised by @michaelneale). If there's interest, we'd welcome a contribution here — the scope is mostly contained to `crates/goose/src/providers/oauth.rs`.

PR #8074, opened 2026-03-23
auth: databricks OAUTH using keyring for storage

alexhancock:
Resolves https://github.com/block/goose/issues/3468

We can discuss tradeoffs of doing this. Seems more secure, but like @michaelneale I also do not love the user experience of having goose use keychain on mac for everything (password entry requests without explanation)

### DIFF
--- a/crates/goose/src/providers/oauth.rs
+++ b/crates/goose/src/providers/oauth.rs
@@ -1,4 +1,3 @@
-use crate::config::paths::Paths;
 use anyhow::Result;
 use axum::{extract::Query, response::Html, routing::get, Router};
 use base64::Engine;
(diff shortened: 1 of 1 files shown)


alexhancock:
Closing this out as it's not the right place for this code if we want to do it. `oauth.rs` is a general purpose utility for multiple providers. Not databricks specific. I can take another rev if we decide we want this.

DOsinga (approved):
Fixed two issues from the Codex review before approving:

1. ** after initial browser login is now best-effort** — a keyring denial (e.g. user clicks "Deny" on macOS Keychain prompt) after a successful OAuth dance no longer fails the whole auth flow. Matches the existing treatment on the refresh path.

2. ** always runs the filesystem cleanup** — keyring errors are now warned-and-continued so the legacy on-disk plaintext token files are always removed, even when the keyring is unavailable.

Also removed doc-comments that restated the field names / code.

DOsinga:
sorry goose picked it up any way

PR #8086, opened 2026-03-24
chore: clean up stray recipe.yaml

alexhancock:
Removing a file that seems to be checked in by mistake from somewhere


DOsinga:
I wonder if we should just have a folder person-recipes or some such that is in .gitignore

PR #8097, opened 2026-03-25
feat: allow GOOSE_CLI_SHOW_THINKING to be set in config.yaml

dale-lakes:
Previously `GOOSE_CLI_SHOW_THINKING` could only be set as an environment variable. Now it can also be set in config.yaml as a boolean, matching how other goose settings work.

### DIFF
--- a/crates/goose-cli/src/session/output.rs
+++ b/crates/goose-cli/src/session/output.rs
@@ -17,7 +17,6 @@ use std::cell::RefCell;
 use std::collections::HashMap;
 use std::io::{Error, IsTerminal, Write};
 use std::path::Path;
-use std::sync::LazyLock;
 use std::time::Duration;
 
 use super::streaming_buffer::MarkdownBuffer;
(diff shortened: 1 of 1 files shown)


crates/goose-cli/src/session/output.rs
```diff
@@ -447,7 +447,11 @@ pub fn goose_mode_message(text: &str) {
 }
 
 static SHOW_THINKING: LazyLock<bool> = LazyLock::new(|| {
-    std::env::var("GOOSE_CLI_SHOW_THINKING").is_ok() && std::io::stdout().is_terminal()
+    let from_env = std::env::var("GOOSE_CLI_SHOW_THINKING").is_ok();
+    let from_config = Config::global()
+        .get_param::<bool>("GOOSE_CLI_SHOW_THINKING")
+        .unwrap_or(false);
+    (from_env || from_config) && std::io::stdout().is_terminal()
```
DOsinga: Config already reads from the env variables, so this is double. I'll fix.

DOsinga (approved):
Config::get_param already checks the env var (uppercased key) before falling back to config.yaml, so the separate env-var check and the LazyLock static were redundant. Simplified to a direct call in should_show_thinking().

PR #8111, opened 2026-03-25
fix: gate code-mode tool filtering behind cfg(feature) for --no-defau…

r0x0d:
## Summary
The `if code_execution_active` block references `code_execution` and `pctx_code_mode` symbols that only exist when the `code-mode` feature is enabled. Without the compile-time gate, `cargo check --no-default-features` fails with unresolved module errors even though the branch is dead at runtime.

### Testing
`cargo check --no-default-features`

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:


DOsinga (approved):
LGTM! Correct fix for the  compile error — the  gate matches the existing pattern used just above for .

PR #8128, opened 2026-03-26
fix: remove unsupported cwd parameter from changesets action

alexhancock:
## Problem

The npm publish workflow was failing with the error:
```
Error: There is no .changeset directory in this project
```

This was caused by the `changesets/action@v1` not supporting the `cwd` parameter that was being used in the workflow.

## Solution

Removed the unsupported `cwd: ui` parameter and instead prefixed the publish and version commands with `cd ui &&` to run them from the correct directory where the `.changeset` folder exists.

### Changes
- `publish: pnpm run release` → `publish: cd ui && pnpm run release`
- `version: pnpm run version` → `version: cd ui && pnpm run version`
- Removed: `cwd: ui`

## Verification

✅ Workflow run [#23572031847](https://github.com/block/goose/actions/runs/23572031847) completed successfully
- All 6 build jobs passed
- Release to npm job passed (46s)
- No ".changeset directory" error

## References

- Failed run: https://github.com/block/goose/actions/runs/23564809405/job/68619943297
- Fixed run: https://github.com/block/goose/actions/runs/23572031847

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
 
-      - uses: actions-rust-lang/setup-rust-toolchain@v1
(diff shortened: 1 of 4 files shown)


DOsinga:
My agent had a different take. Let me know what you think

PR #8129, opened 2026-03-26
feat: Gemini OAuth provider

michaelneale:
This is, similar to chatgpt, using gemini oauth PKCE to login to get access to api. 

This should replace gemini cli and gemini acp providers 

---

### What

Adds a new `gemini_oauth` provider that lets users sign in with their Google account to use Gemini models — no API key needed.

### How it works

1. **OAuth flow** — Opens browser for Google sign-in, runs a local callback server, exchanges the auth code for tokens using PKCE. Uses the same public installed-app OAuth credentials as the Gemini CLI.

2. **Code Assist API** — Talks to `cloudcode-pa.googleapis.com` (the same backend the Gemini CLI uses for "Login with Google"). The standard `generativelanguage.googleapis.com` API does **not** accept OAuth tokens for content generation — it is API-key only.

3. **User onboarding** — First-time users go through `loadCodeAssist` / `onboardUser` to get a project ID, cached alongside tokens at `~/.config/goose/gemini_oauth/tokens.json`.

4. **Request/response adaptation** — Wraps outgoing requests in the Code Assist envelope format and unwraps incoming SSE response chunks so the existing Google format parser handles the actual content unchanged.

5. **Token management** — Automatic caching, refresh, and re-auth if refresh fails.

### Files changed

- `crates/goose/src/providers/gemini_oauth.rs` — new provider (~800 lines)
- `crates/goose/src/providers/mod.rs` — module registration
- `crates/goose/src/providers/init.rs` — provider registration (visible to users)
- `crates/goose/src/providers/canonical/name_builder.rs` — maps `gemini_oauth` → `google` for model name resolution

### Context

The Gemini CLI's "Login with Google" OAuth flow does NOT use the public Gemini API. It uses Google's internal Code Assist API, which has a different request envelope format (wraps the standard Gemini request body) and requires user onboarding. This provider replicates that flow in Rust.


bin/corepack
```diff
@@ -1 +1 @@
-.node-24.10.0.pkg
```
lifeizhou-ap, line 1: can we upgrade node in a separate pr?  we need to change the version in other places such as windows

PR #8133, opened 2026-03-26
feat: optionally enforce path confinement in developer extension

### LINKED ISSUE #7587: Draft: Developer platform extension write/edit follow symlinks (boundary ambiguity)
### Category
Security boundary ambiguity / unintended file access via symlink resolution

### Where
- `crates/goose/src/agents/platform_extensions/developer/edit.rs`
  - `file_edit_with_cwd`: uses `fs::read_to_string(&path)`
  - `file_write_with_cwd`: uses `fs::write(path, &params.content)`

### What was observed
The platform developer extension `write` and `edit` operations resolve user-supplied paths and then perform normal filesystem read/write operations that follow symlinks.

There is no canonical-target enforcement before IO and no symlink rejection/no-follow behavior in this path.

### Why this matters
If callers treat `path` as a boundary indicator (for example, “inside workspace”), symlink targets can violate that expectation.

A symlink located in an expected-safe location can point to a different file, causing reads/writes to affect the target rather than the link path itself.

### Reproduction
1. Create a workspace file symlink, e.g. `workspace/notes.txt -> /tmp/real_target.txt`.
2. Invoke platform developer `write` or `edit` against `workspace/notes.txt`.
3. Observe that the operation reads/writes `/tmp/real_target.txt` (the symlink target).

### Scope clarification
This report is specifically about the current platform extension path (`crates/goose/src/agents/platform_extensions/developer/*`). It does not claim a `.gooseignore` bypass.

jh-block:
## Summary

Replaces the unchecked `resolve_path` with `validate_and_resolve_path` that enforces confinement to the working directory (allowance is made for a future feature that allows paths to be added to the allowed set). This prevents absolute path escape (`/etc/passwd`), `..` traversal (`../../etc/passwd`), and symlink-based escapes from reaching files outside the working directory.

Confinement is off by default to match the existing behaviour where working directory is just a base path. It can be enabled by setting the GOOSE_CONFINEMENT config key to true.

### Testing

New unit tests covering: absolute path rejection, `..` traversal rejection, symlink escape rejection, symlink within working dir allowed, new file write within working dir, `..` escape on write, and no-confinement mode.

### Related Issues

Fixes #7587

### DIFF
--- a/crates/goose-acp/src/fs.rs
+++ b/crates/goose-acp/src/fs.rs
@@ -4,7 +4,7 @@ use async_trait::async_trait;
 use fs_err as fs;
 use goose::agents::mcp_client::{Error as McpError, McpClientTrait};
 use goose::agents::platform_extensions::developer::edit::{
-    resolve_path, string_replace, FileEditParams, FileReadParams, FileWriteParams,
+    string_replace, validate_and_resolve_path, FileEditParams, FileReadParams, FileWriteParams,
 };
 use goose::agents::platform_extensions::developer::shell::{ShellParams, OUTPUT_LIMIT_BYTES};
 use goose::agents::platform_extensions::developer::DeveloperClient;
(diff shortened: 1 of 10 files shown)


DOsinga:
I'm not sure this helps really; an agent can still write a delete_all.sh script and execute that. isn't this just fake security? I'm pretty sure agents would consider that if they run into an error.

jh-block:
> I'm not sure this helps really; an agent can still write a delete_all.sh script and execute that. isn't this just fake security? I'm pretty sure agents would consider that if they run into an error.

I think if you combine it with something which allows only reading/writing/listing files (so, just being able to disable the shell tool in the developer extension), you have something potentially interesting. I agree the shell tool is an escape to basically anything like this

PR #8135, opened 2026-03-26
fix: improve tool call expansion behavior

alexhancock:
## Problem

Two UI issues with tool call expansion in the desktop app:

1. **Last tool call stays expanded**: When a tool completes but is the last one in a streaming message, it remains expanded if it had logs or progress during execution.

2. **Tools with no content show chevron**: Tools like `todo__todo_write` that have no arguments and return simple success responses show an expand/collapse chevron but have nothing to display when clicked.

### Fix 1: Auto-collapse after completion
Added `isForceCollapse` prop to `ToolCallExpandable` that automatically collapses tools after they finish loading, but only if the user hasn't manually interacted with the expansion state. This is set to `true` when:
- Tool is not currently rendering progress (`!isRenderingProgress`)
- Tool has finished loading (`loadingStatus !== 'loading'`)

### Fix 2: Hide chevron when no content
Added `hasContent` check that:
- Inspects the children prop to determine if there's displayable content
- Hides the expand/collapse chevron when no content exists
- Disables the button to prevent clicking

## Testing

- [x] TypeScript compilation passes
- [x] ESLint passes
- [x] Prettier formatting applied

## Behavior Changes

**Before:**
- Last tool in message stayed expanded after completion
- Empty tools showed clickable chevron with no content

**After:**
- Tools auto-collapse after completion (unless manually expanded)
- Empty tools show no chevron and are non-interactive
- User manual expansion is preserved

Fixes the issues reported in the conversation.

### DIFF
--- a/ui/desktop/src/components/ToolCallWithResponse.tsx
+++ b/ui/desktop/src/components/ToolCallWithResponse.tsx
@@ -291,6 +291,7 @@ interface ToolCallExpandableProps {
   isStartExpanded?: boolean;
   isForceExpand?: boolean;
   isForceCollapse?: boolean;
+  hasContent?: boolean;
   children: React.ReactNode;
   className?: string;
 }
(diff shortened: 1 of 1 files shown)


alexhancock:
This one was a draft unintentionally put up by goose

https://github.com/block/goose/pull/8136 will be the solution

Issue #7969, opened 2026-03-17
Load source tool cannot read existing file (Windows-specific).


0xfk0:
**Load source cannot read file**

I see an error:
```
Error: Source 'c:\Users\localuser\any2txt\Program.cs' not found. Use load() to see available sources.
```

The model retries with `type` command and the same file. And I see file contents.
Looks like "Load source" tool doesn't work correctly with windows-specific paths?

Please see the logs:
[diagnostics_20260317_13.zip](https://github.com/user-attachments/files/26065068/diagnostics_20260317_13.zip)

---

**To Reproduce**
1. Ask a model to read source file, give absolute path to the file.
2. Tool call gives an error.
---

**Expected behavior**
Error should not happen.

---

**Screenshots**

---

**Please provide the following information**
- **OS & Arch:** windows 6.2.9200 x86_64
- **Interface:** UI
- **Version:** 1.27.2
- **Extensions enabled:** summon, developer
- **Provider & Model:** custom_deepinfra – zai-org/GLM-5

DOsinga:
hmm, it looks like `summon` picked this up with `load` - not sure why the model would decide to do that. 

the problem is that the load() reports that the file cannot be found, which is not true; the is not a recipe or a skill - opus 4.6 is smart enough to figure this out. but I think if we just had a better error this model would work too. 

/cc @tlongwell-block

majiayu000:
Hi, I've been looking into this and traced the root cause.

**Proposed approach:** Find the load() tool implementation in the summon extension (likely in Rust under crates/ or extensions/). Update the error message from generic 'Source X not found' to something like 'Source X is not a registered recipe or skill. To read file contents, use the read_file tool instead.' This gives the LLM enough context to retry with the correct tool. May also want to add Windows path normalization (backslash→forward slash) if the source lookup does path matching.

I'll open a draft PR shortly. Happy to adjust based on your preference.

DOsinga:
Closing this — the `load` tool is working as designed. It's for loading recipes, skills, and agents, not arbitrary files. When the LLM passes a file path to `load()`, that's the model picking the wrong tool. The tool description and error message already provide enough context for the model to self-correct and use the developer extension's file-reading tools instead. This has also become less common with newer models.

PR #8143, opened 2026-03-26
fix(summon): reject absolute file paths in load source tool

balcsida:
## Summary

When an LLM passes an absolute file path (e.g. `c:\Users\localuser\file.cs` on Windows or `/home/user/file.rs` on Unix) to the `load()` tool, `resolve_source()` fails with a confusing "Source not found" error. This is because `split_once('/')` either fails on Windows backslash paths or incorrectly splits on the drive letter separator.

This adds an early check in `resolve_source()` to detect absolute paths and return a clear error message directing the LLM to use file tools (e.g. `read_file`) instead.

### Testing

- Added `test_load_source_rejects_windows_absolute_path` — covers both `c:\...` and `C:/...` paths
- Added `test_load_source_rejects_unix_absolute_path` — covers `/home/...` paths
- All 27 existing summon tests continue to pass

### Related Issues
Relates to #7969

N/A

### DIFF
--- a/crates/goose/src/agents/platform_extensions/summon.rs
+++ b/crates/goose/src/agents/platform_extensions/summon.rs
@@ -756,6 +756,22 @@ impl SummonClient {
         name: &str,
         working_dir: &Path,
     ) -> Result<Option<Source>, String> {
+        // Detect absolute file paths — these aren't valid source names.
+        // On Unix: starts with '/'
+        // On Windows: starts with drive letter like 'C:\' or 'C:/'
+        let looks_like_absolute_path = Path::new(name).is_absolute()
+            || (name.len() >= 3
+                && name.as_bytes()[0].is_ascii_alphabetic()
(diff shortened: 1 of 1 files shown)


DOsinga:
Hey @balcsida, thanks for looking into this! 👋

Since the original issue was filed, the `handle_load_source` error handling has been improved — when a source isn't found, it already returns a clear message like:

> Source 'c:\Users\localuser\file.cs' not found. Use load() to see available sources.

It also does fuzzy matching to suggest similar source names, and for skill sub-paths it lists available files. So the LLM already gets enough information to course-correct and use the right tool (e.g. `read_file`).

Adding a special case for absolute paths would be papering over what's fundamentally an LLM tool-selection issue — the existing generic "not found" path handles it well. We'd rather keep `resolve_source` simple and let the clear error message do the work.

Thanks again for the contribution!

PR #8149, opened 2026-03-27
feat(security): add egress logging inspector

dorien-koelemeijer:
## Summary

Adds a new `EgressInspector` that runs in observation-only mode alongside the existing `SecurityInspector`. It detects outbound network destinations in shell commands and web tool calls, logging them via `tracing::info!` without blocking anything. This gives visibility into what egress traffic goose is generating during a session.

The inspector recognises the following egress patterns:
* HTTP/HTTPS/FTP URLs
* Git SSH remotes (git@host:org/repo)
* S3 buckets (s3://)
* GCS buckets (gs://)
* SCP and rsync targets
* SSH connections
* Docker push/login targets
* npm and cargo package publishes

Tool coverage goes beyond shell commands — it also handles web fetch and browser navigation tools, matching both exact tool names and MCP-namespaced variants (e.g. mcp__developer__shell). The suffix-matching approach (e.g. ends_with("__shell")) means MCP servers that namespace their tools will be covered automatically without needing to enumerate every possible name. Multiple argument key names are tried (command, cmd, script, input for shell; url, uri, endpoint for web tools) to handle the variety of conventions used across different MCP servers.

Destinations are deduplicated within a single inspection pass so the same URL appearing multiple times in one session doesn't produce duplicate log entries. npm/cargo publish detection uses anchored regex rather than substring matching to avoid false positives from commands like echo "npm publish".

Each detected destination is logged with egress_kind, domain, and destination fields. 
Note: the destination field may contain sensitive data (credentials embedded in URLs, internal hostnames). Sanitisation before any external log export (e.g. OTLP) should be handled at the transport layer and is out of scope for this PR (another PR will be opened shortly to address this). 

### Testing
Unit tests cover all supported egress pattern types — URLs, git SSH remotes, S3, GCS, SCP, rsync, SSH, Docker, and npm/cargo publish — including false positive cases for the publish detection.

End-to-end testing via `just run-ui` was attempted but the server's file logger does not currently write to disk (pre-existing issue, not introduced by this PR). Tracing output is available on stderr when running goosed agent directly, which confirmed the inspector is wired into the pipeline correctly. Full end-to-end validation of live egress detection is deferred until the logging infrastructure issue is resolved.

### DIFF
--- a/crates/goose/src/security/egress_inspector.rs
+++ b/crates/goose/src/security/egress_inspector.rs
@@ -170,9 +170,12 @@ fn extract_destinations(command: &str) -> Vec<EgressDestination> {
 }
 
 fn extract_domain_from_url(url: &str) -> Option<String> {
-    let after_scheme = url.find("://").map(|i| &url[i + 3..]).unwrap_or(url);
+    let after_scheme = url
+        .find("://")
+        .and_then(|i| url.get(i + 3..))
+        .unwrap_or(url);
     let authority = after_scheme.split('/').next()?;
-    let host_port = authority.split('@').last()?;
+    let host_port = authority.split('@').next_back()?;
     let host = if host_port.contains('[') {
         host_port
             .split(']')


michaelneale:
I think a reasonable idea - and any serious enterprise would love to use this IMO - so yeah, lets do this, doesn't seem to hurt things?

crates/goose/src/security/egress_inspector.rs
```diff
@@ -0,0 +1,372 @@
...
+        if !domain.is_empty() {
+            destinations.push(EgressDestination {
+                kind: "url".to_string(),
+                destination: url,
+                domain,
+            });
+        }
+    }
+
+    static GIT_SSH_RE: OnceLock<Regex> = OnceLock::new();
```
michaelneale, line 49: would it lose much if things were a bit less bespoke - really anything potentially matching a host, could have shipped out for later analysis?

dorien-koelemeijer: Yeah, I like the idea, I'm just scared it will be too broad that way. I'll play around and see if iI can find a good way to do this that isn't too broad so the data doesn't become unusable.

dorien-koelemeijer: Made an attempt, it's a bit lame, but it's a bit more broad at least. I think maybe we proceed and see how it works and make changes as necessary?

PR #8155, opened 2026-03-27
refactor: goose-acp-server -> goose binary for TUI

alexhancock:
Cleanup now that `ui/acp` package uses `goose acp` instead of the `goose-acp-server`. Everything now uses the `goose` binary wrapped in a native npm package.

We can get back to having server support when we make `goose serve` do what `goose-acp-server` does now per https://github.com/block/goose/discussions/7697

https://github.com/block/goose/actions/runs/23658726270 is a test publishing build. I will move PR out of draft if it doesn't introduce any new issues


alexhancock:
> Is there a rust binary target to remove too?

I want to keep the goose-acp-server implementation there so that it's intact and can be moved to `goose serve` soon as part of https://github.com/block/goose/discussions/7697

Unless I misunderstood you and there is still cleanup that can be done regardless. Happy to follow up.

PR #8158, opened 2026-03-27
fix: use `overflow: clip` to not disrupt sticky ToC

The-Best-Codes:
Fixes a regression in #6082 that caused the Table of Contents for pages to not be sticky.

## Before

Pay attention to the ToC sidebar on the right.

https://github.com/user-attachments/assets/d412092e-2bc9-4fd5-897b-245a00f73b74

## After

https://github.com/user-attachments/assets/5a3f4071-7d5a-4411-ad47-8a40a390892d

### DIFF
--- a/documentation/src/css/custom.css
+++ b/documentation/src/css/custom.css
@@ -244,7 +244,7 @@ html {
 .hero--logo {
   display: 'flex';
   justify-content: 'center';
-  margin-bottom: '1rem' 
+  margin-bottom: '1rem'
 }
 
 .pill-button {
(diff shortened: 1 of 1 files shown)


documentation/src/css/custom.css
```diff
@@ -244,7 +244,7 @@ html {
 .hero--logo {
   display: 'flex';
   justify-content: 'center';
-  margin-bottom: '1rem'
```
The-Best-Codes, line 247: Zed did this automatically when I saved the file

PR #8165, opened 2026-03-28
feat: Add missing ACP custom methods for prompts, provider_info, plan_prompt, and session/clear

### LINKED ISSUE #7225: CLI-via-goosed: add missing goosed API endpoints and remove remaining self.agent calls

fresh3nough:
## Summary

Adds ACP custom method handlers to close the gaps identified in #7225. This is a focused PR for just the new custom methods (without `goose serve`), addressing the feedback from the closed #7720.

### New ACP custom methods (server.rs, custom_requests.rs)

- `_goose/session/clear` - clear conversation history and reset token counts
- `_goose/agent/plan_prompt` - get the plan prompt string for a session
- `_goose/agent/provider_info` - get provider name, model, context limit, and token usage
- `_goose/config/prompt_info` - get details for a single prompt by name
- `_goose/config/prompts` - list extension prompts for a session

### Feedback from #7720 addressed

- **No `goose serve` subcommand** - scoped to custom methods only
- **Correct test types** - uses existing `AcpServerConnection`, `SessionData`, shared `send_custom`
- **serde consistency** - all new request types include `#[serde(rename_all = "camelCase")]`
- **No race condition** - session/clear cancels via CancellationToken and clears immediately, no arbitrary timeouts

### Tests

7 integration tests in `crates/goose-acp/tests/custom_requests_test.rs` covering all new methods including error cases (invalid sessions, nonexistent prompts).

Test results: 11 passed (4 existing + 7 new), 0 failed

## Verification

- `cargo test -p goose-acp --test custom_requests_test` -- 11/11 passed
- `cargo clippy -p goose-acp --all-targets` -- clean
- `cargo fmt` -- clean

Closes #7225
Supersedes #7720

### DIFF
--- a/crates/goose-acp/src/custom_requests.rs
+++ b/crates/goose-acp/src/custom_requests.rs
@@ -127,6 +127,68 @@ pub struct GetExtensionsResponse {
     pub warnings: Vec<String>,
 }
 
+/// Clear conversation history and reset token counts for a session.
+#[derive(Debug, Deserialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct ClearSessionRequest {
+    pub session_id: String,
+}
+
+/// Get the plan prompt string for a session.
+#[derive(Debug, Deserialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct GetPlanPromptRequest {
+    pub session_id: String,
+}
+
+#[derive(Debug, Serialize, JsonSchema)]
+pub struct GetPlanPromptResponse {
+    pub prompt: String,
+}
+
+/// Get provider name, model, context limit, and token usage for a session.
+#[derive(Debug, Deserialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct GetProviderInfoRequest {
+    pub session_id: String,
+}
+
+#[derive(Debug, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct GetProviderInfoResponse {
+    pub provider_name: String,
+    pub model_name: String,
+    pub context_limit: usize,
+    pub total_tokens: Option<i32>,
+    pub input_tokens: Option<i32>,
+    pub output_tokens: Option<i32>,
+}
+
+/// Get details for a single prompt by name.
+#[derive(Debug, Deserialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct GetPromptInfoRequest {
+    pub session_id: String,
+    pub name: String,
+}
+
+#[derive(Debug, Serialize, JsonSchema)]
+pub struct GetPromptInfoResponse {
+    pub prompt: serde_json::Value,
+}
+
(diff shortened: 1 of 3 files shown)


DOsinga:
Minor question on the API surface: `ListPromptsResponse` and `GetPromptInfoResponse` both use `serde_json::Value` for their payload fields. Since the underlying `Prompt` type from the MCP SDK already implements `Serialize`, is there a reason we can't keep the type information around and use a typed field (e.g. `prompts: HashMap<String, Vec<Prompt>>` / `prompt: Prompt`) instead of erasing to `Value`? That would give callers compile-time guarantees about the shape of the response and avoid an extra serialise-then-deserialise round-trip if anyone ever wants to consume these structs from Rust.

fresh3nough:
nope you are totally correct that is very feasible switch, i really chose the route I did for flexibility reasons with dynamic schemas, i will implement the solution you mention its clearly the type safe fix

DOsinga:
Thanks for working on this @fresh3nough — sorry we've been slow getting to it.

Unfortunately the `goose-acp` crate was merged into `crates/goose/src/acp/` in #8726, so this PR now targets files that no longer exist on main and can't be merged without a full rewrite.

Additionally, `_goose/agent/plan_prompt` is now obsolete — plan mode no longer works that way.

The other endpoints (`_goose/session/clear`, `_goose/agent/provider_info`, `_goose/config/prompts`, `_goose/config/prompt_info`) and the concurrent prompt tracking refactor are still needed per #7225. If you'd like to take another pass targeting `crates/goose/src/acp/server.rs` we'd welcome that — just drop the plan_prompt endpoint.

Thanks again for the contribution!

PR #8175, opened 2026-03-28
fix(openai): improve streaming tool argument parse errors for large payloads

### LINKED ISSUE #8167: bug: Tool call arguments parsing fails with EOF error for large payloads
## Problem

When invoking certain tools (e.g., `gh command to create a new issue`, or writing to `DESIGN.md`), the tool call arguments parsing fails with:

```
Could not interpret tool use parameters for id call_36d45d842b654a26900b87e8: EOF while parsing a string at line 1 column 15596
```

This indicates the JSON string containing tool arguments is being truncated or malformed.

## Context

- Error occurs when tool call arguments are very long (>15KB)
- The JSON parser hits EOF unexpectedly, suggesting the input is truncated
- Happens with specific tool invocations like GitHub CLI commands or file writes

## Possible Causes

1. **Stream truncation**: The response stream containing tool arguments is being truncated before fully read
2. **Buffer overflow**: The buffer for storing tool arguments has a size limit that's being hit
3. **Encoding issue**: The arguments string may contain special characters that aren't being handled correctly

## Steps to Reproduce

1. Use a model that generates long tool call arguments
2. Invoke a tool that requires many parameters (e.g., `gh issue create` with long body, or write large file)
3. Observe the EOF parsing error

## Expected Behavior

Tool call arguments should be parsed correctly regardless of size, or there should be a clear error message indicating size limits.

treebird7:
## Summary

When OpenAI streaming splits tool call argument JSON across many chunks, arguments exceeding ~15 KB can arrive as truncated JSON (no closing brace), causing `serde_json::from_str` to fail with a cryptic "EOF while parsing a string" error. The agent cannot proceed and the tool call is silently lost (fixes #8167).

**Root cause:** The streaming path in `formats/openai.rs` used bare `serde_json::from_str` instead of `safely_parse_json` (which is already used on the non-streaming path at line 455). When the accumulated arguments string is truncated mid-JSON, parsing fails with no useful diagnostic.

**Changes (`crates/goose/src/providers/formats/openai.rs`):**
- Switches streaming parse to `safely_parse_json` (consistent with non-streaming path)
- Detects truncation by checking whether accumulated args end with `}` or `]`
- Emits user-actionable error: _"Tool arguments appear truncated after streaming (N bytes received, no closing brace). Try breaking the operation into smaller pieces."_
- Adds 2 tests: truncated args assert `truncated after streaming` in error; valid 16 KB JSON asserts successful parse

## Test plan

- [ ] `cargo test --package goose providers::formats::openai` — all 49 + 2 new tests pass
- [ ] Reproduce with a tool call that generates > 15 KB arguments (e.g. `gh issue create` with large body): confirm error message is now actionable rather than cryptic EOF
- [ ] Confirm non-streaming path unaffected

🤖 Generated with [Goose](https://github.com/block/goose) — contributed by goose, using goose!

### DIFF
--- a/crates/goose/src/providers/formats/openai.rs
+++ b/crates/goose/src/providers/formats/openai.rs
@@ -750,7 +750,21 @@ where
                         let parsed = if arguments.is_empty() {
                             Ok(json!({}))
                         } else {
-                            serde_json::from_str::<Value>(arguments)
+                            safely_parse_json(arguments).map_err(|e| {
+                                let trimmed = arguments.trim_end();
+                                if !trimmed.ends_with('}') && !trimmed.ends_with(']') {
+                                    format!(
+                                        "Tool arguments appear truncated after streaming \
+                                         ({} bytes received, no closing brace). \
+                                         Try breaking the operation into smaller pieces. \
+                                         Original error: {}",
+                                        arguments.len(),
+                                        e
+                                    )
+                                } else {
+                                    format!("Could not parse tool arguments ({} bytes): {}", arguments.len(), e)
+                                }
+                            })
                         };
 
                         let metadata = if let Some(sig) = &last_signature {
@@ -772,12 +786,12 @@ where
                                     metadata.as_ref(),
                                 )
                             },
-                            Err(e) => {
+                            Err(msg) => {
                                 let error = ErrorData {
                                     code: ErrorCode::INVALID_PARAMS,
                                     message: Cow::from(format!(
                                         "Could not interpret tool use parameters for id {}: {}",
-                                        id, e
+                                        id, msg
                                     )),
                                     data: None,
                                 };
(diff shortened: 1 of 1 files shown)


DOsinga:
Thanks for the contribution! Closing this for a couple of reasons:

1. **Multiple PRs without prior merged contributions** — as a first-time contributor, please wait for earlier PRs to land before opening new ones. See [Pull Requests](https://github.com/block/goose/blob/main/CONTRIBUTING.md#pull-requests).

2. **AI-generated code without sufficient author review** — the truncation heuristic (`ends_with`) is unreliable, the error message is inconsistent (Copilot flagged this and it wasn't addressed), the error gets double-wrapped, and the tests don't exercise the actual streaming accumulation path. See [Quick Responsible AI Tips](https://github.com/block/goose/blob/main/CONTRIBUTING.md#quick-responsible-ai-tips).

The core idea — using `safely_parse_json` on the streaming path to match the non-streaming path — is correct and we'll get that fix in separately.

PR #8179, opened 2026-03-28
Initial native TUI

lu-zero:
It is an initial attempt to provide a tui using iocraft.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4684,7 +4684,6 @@ dependencies = [
  "tokio-util",
  "tracing",
  "tracing-subscriber",
- "uuid",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


DOsinga:
Thanks for the effort here! Unfortunately we need to close this for a few reasons:

- As a first-time contributor, a 3k-line PR adding an entirely new crate is too large for us to review without an established track record. Please see our [Getting Started](https://github.com/block/goose/blob/main/CONTRIBUTING.md#getting-started) guide.
- A feature of this scope should start with a discussion or issue before a PR — see [Discussions, Issues and PRs](https://github.com/block/goose/blob/main/CONTRIBUTING.md#discussions-issues-and-prs).
- There are also 18 unaddressed Codex review comments — see [AI Code Reviews](https://github.com/block/goose/blob/main/CONTRIBUTING.md#ai-code-reviews).

It's also worth noting that goose already has a TUI in `crates/goose-cli/src/ui/text` — that's the path we're investing in. If you're interested in improving the terminal experience, contributing there would be the way to go.

Please don't take this personally — we'd love to see smaller contributions from you first to build up trust, and then we can work up from there!

lu-zero:
it is understandable, I forgot to mark it as draft to begin with, I'll try to see if the typescript ui can get back some changes, hopefully as tiny prs :) I'll move the native tui on a separate project to continue experimenting on the native side so we can compare the two approach

PR #8181, opened 2026-03-29
fix: propagate token usage from ACP provider responses

### LINKED ISSUE #8132: fix(acp): ACP client provider discards token usage from PromptResponse
## Problem

When using an ACP backend (claude-acp, codex-acp), `Provider::complete()` always returns `ProviderUsage` with all-zero token counts. Token tracking, cost reporting, and context-window monitoring are broken for any ACP-based provider.

## Root cause — three-point gap in `crates/goose/src/acp/provider.rs`

1. **Enum**: `AcpUpdate::Complete(StopReason)` only carries `StopReason`, no room for usage ([provider.rs L118](https://github.com/block/goose/blob/main/crates/goose/src/acp/provider.rs#L118))
2. **Sender**: When `PromptResponse` arrives, only `r.stop_reason` is forwarded; `r.usage` is silently dropped ([provider.rs L1112](https://github.com/block/goose/blob/main/crates/goose/src/acp/provider.rs#L1112))
3. **Receiver**: The stream handler breaks out of the loop on `Complete` without yielding any `ProviderUsage` ([provider.rs L711-712](https://github.com/block/goose/blob/main/crates/goose/src/acp/provider.rs#L711))

Since the stream never yields a non-`None` usage, `collect_stream()` in `providers/base.rs` falls back to `ProviderUsage::new("unknown", Usage::default())`.

## ACP protocol support exists

- `agent-client-protocol-schema` v0.10.8 defines `PromptResponse.usage: Option<Usage>` behind the `unstable_session_usage` feature
- Goose enables this via `features = ["unstable"]` in `crates/goose/Cargo.toml`
- The `Usage` struct has: `total_tokens`, `input_tokens`, `output_tokens`, plus optional `thought_tokens`, `cached_read_tokens`, `cached_write_tokens`

## ACP agent adapter status

| Agent adapter | Populates `PromptResponse.usage`? | Sends `UsageUpdate` notifications? |
|---|---|---|
| [claude-agent-acp](https://github.com/agentclientprotocol/claude-agent-acp) | **Yes** — accumulates from every Claude API result | Yes (context window + cost) |
| [codex-acp](https://github.com/zed-industries/codex-acp) (Zed) | **No** — returns bare `PromptResponse::new(stop_reason)` | Yes (context window only) |
| goose-acp server (`crates/goose-acp/src/server.rs`) | **No** — returns bare `PromptResponse::new(stop_reason)` | — |

Note: `UsageUpdate` (a mid-stream `SessionNotification`) only carries context-window-level data (`used`, `size`, optional `cost`), **not** per-turn input/output token breakdown. It cannot substitute for `PromptResponse.usage`.

### Part A: ACP client provider (`crates/goose/src/acp/provider.rs`) — consume usage

Three targeted changes:

**1. Extend `AcpUpdate::Complete` to carry usage:**

```rust
use agent_client_protocol_schema::Usage as AcpUsage;

enum AcpUpdate {
    // ...existing variants...
    Complete(StopReason, Option<AcpUsage>),
    // ...
}
```

**2. Extract `r.usage` when sending the complete signal (~L1112):**

```rust
Ok(r) => {
    let _ = response_tx.try_send(AcpUpdate::Complete(r.stop_reason, r.usage));
}
```

**3. Map ACP `Usage` to `ProviderUsage` and yield on the final stream item (~L711):**

```rust
AcpUpdate::Complete(_reason, acp_usage) => {
    if let Some(u) = acp_usage {
        let provider_usage = ProviderUsage::new(
            model_config.model_name.clone(),
            Usage {
                input_tokens: Some(u.input_tokens as i32),
                output_tokens: Some(u.output_tokens as i32),
                total_tokens: Some(u.total_tokens as i32),
            },
        );
        yield (None, Some(provider_usage));
    }
    break;
}
```

### Part B: ACP server (`crates/goose-acp/src/server.rs`) — emit usage

The Goose ACP server (when Goose acts as an ACP agent) returns bare `PromptResponse::new(stop_reason)` without populating `.usage()`. Since Goose already tracks `ProviderUsage` internally per agent turn, it should accumulate and return it:

```rust
let response = PromptResponse::new(stop_reason)
    .usage(acp_schema::Usage::new(
        total_tokens,
        input_tokens,
        output_tokens,
    ));
Ok(response)
```

## Notes

- The `Usage` type from `agent-client-protocol-schema` uses `u64` while Goose's internal `Usage` uses `Option<i32>`. The mapping truncates, which is fine for practical token counts but worth noting.
- `AcpUsage` also has optional `thought_tokens`, `cached_read_tokens`, `cached_write_tokens` which Goose's `Usage` does not model. These could be surfaced later.
- For agents that don't populate `PromptResponse.usage` (like codex-acp today), the fix degrades gracefully — `r.usage` is `None`, no `ProviderUsage` is yielded, and the existing `"unknown"` fallback applies. As those adapters add usage reporting, it starts working automatically.

xr843:
## Summary

Fixes #8132 — ACP client provider discards token usage data from `PromptResponse`, causing all-zero token counts for ACP-based providers (claude-acp, codex-acp). This breaks cost reporting and context window monitoring.

## Changes

Three targeted changes in `crates/goose/src/acp/provider.rs`:

1. **Extend `AcpUpdate::Complete`** to carry `Option<AcpUsage>` alongside `StopReason`
2. **Forward `r.usage`** when sending the complete signal from the prompt response handler
3. **Map ACP `Usage` to `ProviderUsage`** and yield it as the final stream item before breaking

## Graceful degradation

For ACP agents that don't yet populate `PromptResponse.usage` (e.g. codex-acp, goose-acp server), `r.usage` is `None`, no `ProviderUsage` is yielded, and the existing `"unknown"` fallback in `collect_stream()` applies unchanged. As those adapters add usage reporting, it starts working automatically.

## Notes

- `AcpUsage` fields are `u64`, Goose's `Usage` uses `Option<i32>` — the `as i32` cast is safe for practical token counts
- `AcpUsage` also has optional `thought_tokens`, `cached_read_tokens`, `cached_write_tokens` which Goose's `Usage` doesn't model yet — these could be surfaced in a follow-up

Credit to the issue reporter for the thorough root cause analysis and proposed fix.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/goose/src/acp/provider.rs
+++ b/crates/goose/src/acp/provider.rs
@@ -1,3 +1,4 @@
+use agent_client_protocol_schema::Usage as AcpUsage;
 use agent_client_protocol_schema::AGENT_METHOD_NAMES;
 use anyhow::{Context, Result};
 use async_stream::try_stream;
@@ -30,7 +31,7 @@ use crate::conversation::message::{Message, MessageContent};
 use crate::model::ModelConfig;
 use crate::permission::permission_confirmation::PrincipalType;
 use crate::permission::{Permission, PermissionConfirmation};
-use crate::providers::base::{MessageStream, PermissionRouting, Provider};
+use crate::providers::base::{MessageStream, PermissionRouting, Provider, ProviderUsage, Usage};
 use crate::providers::errors::ProviderError;
 use crate::subprocess::configure_subprocess;
 
@@ -115,7 +116,7 @@ enum AcpUpdate {
         request: Box<RequestPermissionRequest>,
         response_tx: oneshot::Sender<RequestPermissionResponse>,
     },
-    Complete(StopReason),
+    Complete(StopReason, Option<AcpUsage>),
     Error(String),
 }
 
@@ -635,6 +636,7 @@ impl Provider for AcpProvider {
             .map_err(|_| ProviderError::RequestFailed("goose_mode lock poisoned".into()))?;
 
         let reject_all_tools = goose_mode == GooseMode::Chat;
+        let model_name = model_config.model_name.clone();
 
         Ok(Box::pin(try_stream! {
             // ACP agents execute tools internally. Goose never dispatches tool calls;
@@ -708,7 +710,19 @@ impl Provider for AcpProvider {
                         let response = map_permission_response(&permission_mapping, &request, decision);
                         let _ = response_tx.send(response);
                     }
-                    AcpUpdate::Complete(_reason) => {
+                    AcpUpdate::Complete(_reason, acp_usage) => {
+                        if let Some(u) = acp_usage {
+                            let provider_usage = ProviderUsage::new(
+                                model_name.clone(),
+                                Usage {
+                                    input_tokens: Some(u.input_tokens as i32),
+                                    output_tokens: Some(u.output_tokens as i32),
+                                    total_tokens: Some(u.total_tokens as i32),
+                                    ..Default::default()
+                                },
+                            );
+                            yield (None, Some(provider_usage));
+                        }
                         break;
                     }
                     AcpUpdate::Error(e) => {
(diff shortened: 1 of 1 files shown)


xr843:
Closing this per @DOsinga's feedback — focusing on #8182 first as suggested. Will revisit this after #8182 lands. Thanks for the guidance!

PR #8191, opened 2026-03-29
fix: allow LM Studio API key to be saved/update from provider settings

### LINKED ISSUE #8125: Provider config modal: "Update Provider" button does nothing after entering API key
LM Studio works fine without an API key, but adding an API key in the provider configuration does not work.

**To Reproduce**
1. Open **LM Studio** provider settings.
2. Enter an API key.
3. Click **Update Provider**.

**Actual behavior**
Nothing happens when **Update Provider** is clicked. The API key is not saved, and there is no visible error.

**Expected behavior**
The API key should save successfully, or the UI should show a validation/error message.

**Additional context**
The LM Studio API itself is working fine. The issue appears to be specific to saving the API key in the provider configuration UI.

Abhijay007:
closes https://github.com/block/goose/issues/8125

Before:  

After:

### DIFF
--- a/crates/goose/src/config/declarative_providers.rs
+++ b/crates/goose/src/config/declarative_providers.rs
@@ -426,11 +426,31 @@ pub fn register_declarative_providers(
     let dir = custom_providers_dir();
     let custom_providers = load_custom_providers(&dir)?;
     let fixed_providers = load_fixed_providers()?;
+
+    let fixed_names: std::collections::HashSet<&str> =
+        fixed_providers.iter().map(|c| c.name.as_str()).collect();
+
+    let (shadows, true_custom): (Vec<_>, Vec<_>) = custom_providers
+        .into_iter()
+        .partition(|c| fixed_names.contains(c.name.as_str()));
+
+    let shadow_map: HashMap<String, DeclarativeProviderConfig> =
+        shadows.into_iter().map(|c| (c.name.clone(), c)).collect();
+
     for config in fixed_providers {
+        let config = if let Some(shadow) = shadow_map.get(&config.name) {
+            DeclarativeProviderConfig {
+                requires_auth: shadow.requires_auth,
+                api_key_env: shadow.api_key_env.clone(),
+                ..config
+            }
+        } else {
+            config
+        };
         register_declarative_provider(registry, config, ProviderType::Declarative);
     }
 
-    for config in custom_providers {
+    for config in true_custom {
         register_declarative_provider(registry, config, ProviderType::Custom);
     }
 
--- a/crates/goose/src/config/declarative_providers.rs
+++ b/crates/goose/src/config/declarative_providers.rs
@@ -254,14 +254,14 @@ pub fn update_custom_provider(params: UpdateCustomProviderParams) -> Result<()> 
         String::new()
     };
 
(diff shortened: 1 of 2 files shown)


DOsinga:
Thanks for looking into this — the bug is real and the UI validation fix makes sense.

However, the shadow file approach is more complex than it needs to be. The secret is already being stored in the keyring by `update_custom_provider` (the `config.set_secret` call happens before the `if editable` check). The only reason the fixed provider does not pick it up is that `load_provider` returns the fixed config as-is, with `requires_auth: false` and `api_key_env: ""`.

Since `api_key_env` is deterministic (`generate_api_key_name(id)` → `"{ID}_API_KEY"`), you can detect at load time whether the user has configured auth by checking if a secret exists in the keyring:

```rust
// in load_provider, after loading a fixed provider config
let api_key_name = generate_api_key_name(id);
let global_config = Config::global();
if global_config.get_secret::<String>(&api_key_name).is_ok() {
    config.requires_auth = true;
    config.api_key_env = api_key_name;
}
```

Same thing in `register_declarative_providers` for the fixed providers loop.

That removes the need for shadow files, `find_fixed_provider`, the partition/merge logic in `register_declarative_providers`, and the new branches in `update_custom_provider` — roughly 60 lines of code replaced by a few lines. The keyring already *is* the source of truth for whether the user configured auth.

Abhijay007:
> Thanks for looking into this — the bug is real and the UI validation fix makes sense.
> 
> However, the shadow file approach is more complex than it needs to be. The secret is already being stored in the keyring by `update_custom_provider` (the `config.set_secret` call happens before the `if editable` check). The only reason the fixed provider does not pick it up is that `load_provider` returns the fixed config as-is, with `requires_auth: false` and `api_key_env: ""`.
> 
> Since `api_key_env` is deterministic (`generate_api_key_name(id)` → `"{ID}_API_KEY"`), you can detect at load time whether the user has configured auth by checking if a secret exists in the keyring:
> 
> ```rust
> // in load_provider, after loading a fixed provider config
> let api_key_name = generate_api_key_name(id);
> let global_config = Config::global();
> if global_config.get_secret::<String>(&api_key_name).is_ok() {
>     config.requires_auth = true;
>     config.api_key_env = api_key_name;
> }
> ```
> 
> Same thing in `register_declarative_providers` for the fixed providers loop.
> 
> That removes the need for shadow files, `find_fixed_provider`, the partition/merge logic in `register_declarative_providers`, and the new branches in `update_custom_provider` — roughly 60 lines of code replaced by a few lines. The keyring already _is_ the source of truth for whether the user configured auth.

Sure, thanks for the review @DOsinga will look into this

lifeizhou-ap:
Hi @Abhijay007 I noticed that this PR has been inactive for a while.  Would you like to close it for now and reopen it when you have time to work on it in the future?

Abhijay007:
> Hi @Abhijay007 I noticed that this PR has been inactive for a while. Would you like to close it for now and reopen it when you have time to work on it in the future?

Sure, will resolve it soon

DOsinga:
how about it @Abhijay007 -- we can also close it for now and then reopen when active again?

DOsinga:
closing this for now

PR #8207, opened 2026-03-30
feat: show $0.00 cost for local Ollama provider sessions

### LINKED ISSUE #8192: feat: show $0.00 (or hide) cost line for local Ollama provider sessions
## Summary

When using the `ollama` provider, goose-cli displays the cost line using cloud pricing tables, producing figures like `Cost: $0.0006 USD (4107 tokens: in 4096, out 11)` even though actual Ollama inference cost is $0.00.

## Steps to reproduce

1. Configure goose with `GOOSE_PROVIDER: ollama` and `GOOSE_MODEL: mistral-nemo` (or any Ollama model)
2. Set `GOOSE_CLI_SHOW_COST: true`
3. Send any message
4. Observe: cost line shows non-zero value matching approximate OpenRouter pricing for that model

## Expected behavior

For Ollama provider turns: show `Cost: $0.00 USD` or suppress the cost line entirely. The token count portion (`4107 tokens: in 4096, out 11`) is still useful and ideally retained.

## Context

In multi-provider sessions (Ollama default with occasional cloud model switches), `GOOSE_CLI_SHOW_COST` must remain `true` so cloud costs are tracked when the user switches models mid-session. A per-turn provider check would resolve this cleanly.

## Suggested approach

Check the provider on each response turn. If `ollama`, display `$0.00` or omit the cost. If cloud provider, display actual cost as today.

fresh3nough:
## Summary

When using the `ollama` provider, goose displays cloud pricing (e.g. `Cost: $0.0006 USD`) even though Ollama inference is free. This PR adds a per-provider zero-cost check so that Ollama sessions correctly show `$0.0000` while preserving token counts.

## Changes

- **CLI** (`crates/goose-cli/src/session/output.rs`): Added `is_zero_cost_provider()` helper; `estimate_cost_usd()` returns `Some(0.0)` for Ollama, so the cost line reads `$0.0000 USD` with token counts intact.
- **Server** (`crates/goose-server/src/routes/config_management.rs`): Added matching `is_zero_cost_provider()` helper; `get_canonical_model_info` zeroes out all cost fields for Ollama so the desktop UI also displays zero cost.
- **Tests**: Two new unit tests validating Ollama returns zero cost and OpenAI returns nonzero cost.

## How to test

1. Configure goose with `GOOSE_PROVIDER=ollama`, `GOOSE_MODEL=mistral-nemo`, `GOOSE_CLI_SHOW_COST=true`
2. Send a message
3. Verify cost line shows `$0.0000 USD` with token counts

Closes #8192

### DIFF
--- a/crates/goose-cli/src/session/output.rs
+++ b/crates/goose-cli/src/session/output.rs
@@ -1414,12 +1414,21 @@ pub fn display_context_usage(total_tokens: usize, context_limit: usize) {
     );
 }
 
+/// Returns true for providers that run models locally at zero cost.
+fn is_zero_cost_provider(provider: &str) -> bool {
+    provider == "ollama"
+}
+
 fn estimate_cost_usd(
     provider: &str,
     model: &str,
     input_tokens: usize,
     output_tokens: usize,
 ) -> Option<f64> {
+    if is_zero_cost_provider(provider) {
+        return Some(0.0);
+    }
+
     let canonical_model = maybe_get_canonical_model(provider, model)?;
 
     let input_cost_per_token = canonical_model.cost.input? / 1_000_000.0;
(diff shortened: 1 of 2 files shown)


DOsinga:
Closing this — the fix is at the wrong layer. Hardcoding zero-cost checks at the CLI and server printing level means every consumer of cost data has to remember to special-case ollama. The right fix is at the provider/canonical model level so that when ollama is the provider, cost is zero at the source. That way all call sites get correct data without duplication.

PR #8209, opened 2026-03-30
feat: goose serve

alexhancock:
Part of https://github.com/block/goose/discussions/7697 and a step towards making the new terminal UX and desktop apps be able to use the http/websocket transport for ACP again

(Draft right now because I want to make sure a client can still connect and use it)


alexhancock:
The TUI now works with this!

Verified with

```
npm start -- --server http://127.0.0.1:3284
```

PR #8212, opened 2026-03-30
merge 'goosed' binary into 'goose'

jamadeo:
`goosed` becomes `goose server`


alexhancock:
Do we want to do this or do we want to just do https://github.com/block/goose/pull/8209, move clients to it, and then `rm -rf goose-server`?

I wonder about whether we'd create more new usage of `goose-server` by doing this that would be harder to unwind.

jamadeo:
Yes I don't think this is how we want to do it.

PR #8216, opened 2026-03-31
only run windows compile on main branch

lifeizhou-ap:
### Why
currently we triggers rust compilation on windows when new changes are pushed into pr.  This step takes long time and if the step failed it does not prevent the change from merging into `main`

### What
Only run in the `main` branch and it already makes easy to trace the changes if this step fails

### Testing
build pipeline

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,8 +79,7 @@ jobs:
   rust-build-windows:
     name: Build Rust Project on Windows
     runs-on: windows-latest
-    needs: changes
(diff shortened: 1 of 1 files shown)


michaelneale (approved):
oh good idea, does it mean we may miss out the signals until later though?

lifeizhou-ap:
> oh good idea, does it mean we may miss out the signals until later though?

I added it recently the window compile in ci recently, but i found it takes long and most time it does not fail. Putting in the main will be easier for us to identify which pr caused the issue

PR #8227, opened 2026-03-31
rust acp client for extension methods

jamadeo:
Make the extension methods usable directly with sacp.

The right way to do this would be to have a separate crate with the custom methods that doesn't depend on the rest of goose


jamadeo:
@alexhancock that part is all [sacp](https://github.com/agentclientprotocol/symposium-acp)'s api. I think what you'd use if you want something like that is [spawn_connection](https://docs.rs/sacp/latest/sacp/struct.ConnectionTo.html#method.spawn_connection)

PR #8228, opened 2026-03-31
mesh: download fresh binary on every start, kill mesh when goose exits

michaelneale:
## Summary

Two critical changes to mesh-llm lifecycle management:

### 1. Always download the latest binary before starting
`start-mesh` now fetches the latest mesh-llm release from GitHub on every start (macOS Apple Silicon only). If the download fails, it falls back to whatever binary is already installed. This ensures users always run the latest mesh-llm version.

The separate `download-mesh` IPC handler has been removed since `start-mesh` subsumes it entirely.

### 2. Mesh process dies with Goose
Previously, mesh-llm was spawned with `detached: true` and `child.unref()`, meaning it outlived the Goose process. Now:
- Spawned as a regular child process of Electron (no detach, no unref)
- OS sends SIGHUP when Electron exits
- Explicit SIGTERM in `will-quit` handler as belt-and-suspenders
- If mesh is already running on port 9337, `start-mesh` skips both download and spawn

### UI simplifications
- On macOS, `not-installed` is treated as `stopped` since start handles the download — no separate download button needed
- On non-macOS, `not-installed` shows install guide only (download/start is macOS-only)
- Start button shows "Downloading latest mesh-llm" status first on macOS, then "Starting"
- Removed dead code: `downloadMesh` function, `download-mesh` preload bridge, `Download` icon import

### Files changed
- `ui/desktop/src/main.ts` — mesh lifecycle handlers
- `ui/desktop/src/preload.ts` — removed download-mesh bridge
- `ui/desktop/src/components/settings/mesh/MeshSettings.tsx` — simplified UI flow

### DIFF
--- a/ui/desktop/src/mesh.ts
+++ b/ui/desktop/src/mesh.ts
@@ -5,7 +5,7 @@
  * probe the API port to detect an externally-running mesh.
  */
 
-import { execFile, execFileSync, execSync, spawn } from 'child_process';
+import { execFile, execFileSync, spawn } from 'child_process';
 import path from 'node:path';
 import os from 'node:os';
 import fsSync from 'node:fs';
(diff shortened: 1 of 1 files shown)


alexhancock (approved):
Code LGTM

Big picture does this only work from the desktop app right now?

michaelneale:
@alexhancock yes - by design, as needs to be persistent, daemon like (youy can run it other ways too natively).

Issue #6025, opened 2025-12-09
Stack Overflow on 'goose configure'


d432456463:
Running the goose configure command on a fresh build from source on Windows consistently results in a
  thread 'main' has overflowed its stack crash. This issue occurs before any output is printed to the
  console and appears to be a fundamental bug in the command's startup or provider initialization
  sequence, as it persists even when all providers are disabled.

  Environment

   * goose-cli Versions Tested: v1.15.0 (latest master), v1.14.0
   * Operating System: Windows

  Steps to Reproduce

   1. Get Source Code: Download the source code for goose-cli (e.g., v1.14.0 from the releases page).
   2. Setup Windows Build Environment:
       * Install Rust via rustup.
       * Install Visual Studio Build Tools with the "Desktop development with C++" workload.
   3. Apply Build Fixes: Two workarounds are required to get the project to compile on Windows:
       * Fix `crunchy` dependency: The git dependency for crunchy fails to fetch. It must be downloaded
         manually and patched in Cargo.toml to use a local path:

   1         # In Cargo.toml
   2         [patch.crates-io]
   3         # crunchy = { git = "...", branch = "..." } # This fails
   4         crunchy = { path = "../path/to/local/crunchy/source" } # This works
       * Fix Linker Error: The build fails with LNK2019: unresolved external symbol SystemFunction036.
         This is resolved by creating a .cargo/config.toml file with the following content:
   1         [target.x86_64-pc-windows-msvc]
   2         rustflags = ["-C", "link-arg=advapi32.lib"]
   4. Build the CLI: Run cargo build -p goose-cli. The build should complete successfully.
   5. Run the command: Execute the built binary:
   1     .\target\debug\goose.exe configure

  Expected Behavior

  The goose configure command should start and display a list of available providers for configuration.

  Actual Behavior

  The command immediately crashes with the following message, and no other output is produced:
   1 thread 'main' has overflowed its stack

  Debugging Steps Taken / Additional Context

  This appears to be a deep, blocking bug in the application's startup logic.

   1. No Stack Trace: Setting the RUST_BACKTRACE=1 environment variable does not produce a stack trace,
      making direct debugging very difficult.
   2. Provider Isolation Test: The most critical finding is that this crash occurs even when all
      providers are commented out from the init_registry function in
      crates/goose/src/providers/factory.rs. This strongly indicates the bug is not in any specific
      provider's code but in the CLI's core logic that runs before or during provider registration for
      the configure command.
[...]

PR #8234, opened 2026-04-01
build: raise default stack reserve to 8 MB

kevklam:
## Summary
See #6025 - on Windows, the default stack size is only 1mb, compared to MacOS/Linux where 8mb is more typical. Rust async futures are fairly demanding stack wise, and debug builds fail at runtime with stack overflow. 

### Testing
Manual testing - the repro steps are simple (Windows only):
1. cargo build -p goose-cli
2. run 'goose acp' and send an initialize payload via stdio
3. application terminates with stack overflow

Issue repros without the stack increase, and does not repro with the stack increase.

### Related Issues
Relates to #6025

n/a


DOsinga:
can you do the DCO thing so we can get this in?

kevklam:
Done - thanks! Not familiar with DCO, but seems relatively straightforward.

PR #8237, opened 2026-04-01
Port provider tests to typescript

jamadeo:
test_providers.sh has evolved beyond the simplicity of a shell script.


DOsinga (approved):
Clean port from bash to TypeScript. Pushed a fix commit to:

- Resolve `.env` from the repo root via `__dirname` instead of `process.cwd()`, matching the old shell script behavior when running from `ui/desktop`
- Strip surrounding quotes from dotenv values so `KEY="value"` works correctly
- Give flaky tests a 120s timeout so the try/catch handler runs before vitest kills the test externally

PR #8244, opened 2026-04-01
refactor: skills as its own platform ext

alexhancock:
Move skill support to its own hidden extension which will be default on, so that skill support is not dependent on the user choosing to enable `summon`

(Draft for now - need to verify and test this)


michaelneale:
does this add a bit more tool use overhead vs one summon (?) tool? is summon on by default? (should it be? I think it is, which is fine for now) - but assume that skills can be even lower level?

PR #8260, opened 2026-04-02
fix(ollama): wire worker model and fix thinking leak in summaries

santiagoaraoz2001-sketch:
## Summary

Three fixes for Ollama-based agentic sessions that cause context pollution, wasted compute, and model looping:

- **Wire `GOOSE_WORKER_MODEL` as the fast model for Ollama provider** — Unlike cloud providers (Anthropic -> claude-haiku, OpenAI -> gpt-4o-mini, etc.), the Ollama provider never called `model.with_fast()`. This meant `complete_fast()` always fell back to the expensive main model for tool-pair summarization and compaction, wasting compute on a 120B+ model for simple summaries.

- **Strip thinking/reasoning content from tool-pair summaries** — When the summarization model is a thinking model (Nemotron 3 Super, Qwen3, etc.), its reasoning tokens leaked into summary messages stored as `Role::User` content. This injected ghost "user thinking" messages (e.g., "We need to summarize: a tool call to shell command...") into the conversation, confusing the main model and polluting context.

- **Improve system, compaction, and subagent prompts** — Added efficiency guidelines (do not re-derive plans), error recovery patterns (do not retry identical failed commands), working memory instructions (do not re-run pwd/git status), and anti-looping directives. These changes are especially impactful for local/open-weight models that are more susceptible to looping.

## Motivation

Observed during an 8-hour Goose session using Nemotron 3 Super (120B) via Ollama:
- 505 messages in the session, ~70% were duplicates or ghost thinking
- The model repeated "We are in the..." re-orientation **108 times**
- 70 user-role thinking messages injected by the summarization system
- The worker model (devstral-small-2) was configured but **never called** (0 requests in logs)
- The model retried the same failed `gh pr create` command 8 times without diagnosing the root cause

## Test plan

- [x] `cargo test -p goose --lib` — 1098 passed, 0 failed
- [x] Snapshot tests updated for new system prompt content
- [x] `cargo check -p goose` — clean compilation
- [ ] Manual test: configure GOOSE_WORKER_MODEL with Ollama and verify complete_fast() uses the worker model
- [ ] Manual test: run a thinking model session and verify tool-pair summaries contain no thinking blocks

## Files changed

| File | Change |
|------|--------|
| `crates/goose/src/providers/ollama.rs` | Read GOOSE_WORKER_MODEL from config and wire as fast model |
| `crates/goose/src/context_mgmt/mod.rs` | Strip Thinking/RedactedThinking content from summaries |
| `crates/goose/src/prompts/system.md` | Add efficiency, error recovery, working memory, and tool usage guidelines |
| `crates/goose/src/prompts/compaction.md` | Exclude thinking blocks, preserve working state |
| `crates/goose/src/prompts/subagent_system.md` | Add anti-looping and error recovery rules |
| `crates/goose/src/agents/snapshots/*.snap` | Updated snapshot tests for new system prompt |

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__all_platform_extensions.snap
+++ b/crates/goose/src/agents/snapshots/goose__agents__prompt_manager__tests__all_platform_extensions.snap
@@ -128,3 +128,30 @@ Template:
 # Response Guidelines
 
 Use Markdown formatting for all responses.
+
+## Efficiency
+
+- Act, don't plan repeatedly. State your intent once, then execute.
+- Never re-derive your full plan from scratch each turn. Reference your earlier reasoning instead.
+- Use the minimum number of tool calls needed. Batch related operations when possible.
+- When a tool call succeeds, move to the next step immediately without restating the plan.
+
+## Error Recovery
+
+- When a tool call fails, read the error message carefully before retrying.
+- Never retry the exact same command that just failed without changing something. Diagnose the root cause first.
+- After 2 consecutive failures on the same task, stop and explain the blocker to the user rather than continuing to retry.
+- For git operations: verify branch state and remote state before creating PRs or pushing.
+
+## Working Memory
+
+- Keep a mental model of your current directory, git branch, and files you have created or modified.
+- Do not run `pwd`, `git status`, or `ls` repeatedly if nothing has changed since your last check.
+- When resuming after tool output, continue from where you left off. Do not restart your reasoning from the beginning.
+
+## Tool Usage
+
+- Prefer reading files before editing them.
+- Prefer a single comprehensive shell command over many small sequential ones.
+- When creating multiple files, plan the dependency order to avoid broken imports.
+- Always verify that imports reference files that exist or that you will create in the current plan.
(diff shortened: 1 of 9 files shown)


DOsinga:
Thanks for the detailed write-up and for digging into the Ollama experience — the problems you identified (thinking content leaking into summaries, worker model not being wired, model looping) are real issues worth solving.

However, this PR bundles several distinct changes — worker model wiring for Ollama, thinking content stripping in summarization, and significant system/compaction/subagent prompt additions — that each affect different parts of the codebase and all users, not just Ollama users. Changes of this scope need prior discussion before a PR, per our [contributing guidelines](https://github.com/block/goose/blob/main/CONTRIBUTING.md#discussions-issues-and-prs).

A couple of specific notes:

- **Worker model for Ollama**: since Ollama models are user-managed, a [custom provider configuration](https://block.github.io/goose/docs/getting-started/providers#custom-provider) is likely the right approach here rather than special-casing `GOOSE_WORKER_MODEL` in the Ollama provider code.
- **Prompt changes**: the 27 lines of new behavioral instructions added to the system prompt (efficiency, error recovery, working memory, tool usage) affect every goose session across all providers. These are opinionated and need careful discussion and testing before landing.
- **Thinking content stripping**: the `retain` filter keeps only `Text` variants, which would also strip `Image`, `ToolRequest`, etc. — the intent is right but the implementation is too broad.

I'd encourage you to open an issue describing the problems you observed (the session data you shared is great evidence) and discuss the approach before putting together a new PR. Breaking this into smaller, focused changes would also make review much easier.

PR #8267, opened 2026-04-02
revert deployment target change

jamadeo:
It broke code mode


jamadeo:
actually, this may have been a red herring

PR #8276, opened 2026-04-03
feat: implement dynamic tool discovery via MCP notifications/tools/li…

makanilani:
# Pull Request

## Title
feat: implement dynamic tool discovery via MCP notifications/tools/list_changed

## URL
https://github.com/block/goose/compare/main...makanilani:goose:dynamic-tool-discovery

---

### Summary

When MCP servers send `notifications/tools/list_changed` (e.g., Google Colab servers that register tools dynamically), Goose now invalidates its tool cache so the next agent turn sees the updated tool set.

### Problem

Currently, Goose calls `tools/list` once during initialization and never checks for updates. The `rmcp` crate's `ClientHandler` trait provides `on_tool_list_changed()` with a default no-op implementation, and `GooseClient` does not override it — so the notification is silently dropped.

### Changes

- Add `ToolCacheInvalidator` trait in `mcp_client.rs`
- Implement `ToolCacheInvalidator` for `ExtensionManager`
- Add `tool_cache_invalidator` field to `GooseClient`
- Implement `on_tool_list_changed` handler in `GooseClient`'s `ClientHandler` impl (overrides rmcp's default no-op)
- Thread invalidator reference through all MCP client construction paths (StreamableHttp, Stdio, Builtin, InlinePython)

### Architecture

```
MCP Server → notifications/tools/list_changed
  → GooseClient::on_tool_list_changed() (new)
    → ExtensionManager::invalidate_tools_cache_and_bump_version() (existing)
      → Agent reads fresh tools on next turn via get_prefixed_tools()
```

### Files Changed

- `Cargo.lock`
- `crates/goose/src/agents/mcp_client.rs`
- `crates/goose/src/agents/extension_manager.rs`

### Testing
Observed  Google colab mcp server works after the change.

---
Co-authored-by: Qwen-Coder <qwen-coder@alibabacloud.com>

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -15,7 +15,7 @@ use std::collections::HashMap;
 use std::path::PathBuf;
 use std::process::Stdio;
 use std::sync::atomic::{AtomicU64, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, Weak};
 use std::time::Duration;
 use tempfile::{tempdir, TempDir};
 use tokio::io::AsyncReadExt;
@@ -34,7 +34,9 @@ use super::tool_execution::{ToolCallContext, ToolCallResult};
 use super::types::SharedProvider;
 use crate::agents::extension::{Envs, ProcessExit};
 use crate::agents::extension_malware_check;
-use crate::agents::mcp_client::{GooseMcpClientCapabilities, McpClient, McpClientTrait};
+use crate::agents::mcp_client::{
+    GooseMcpClientCapabilities, McpClient, McpClientTrait, ToolCacheInvalidator,
+};
 use crate::builtin_extension::get_builtin_extension;
 use crate::config::extensions::name_to_key;
 use crate::config::search_path::SearchPaths;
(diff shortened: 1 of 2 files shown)


DOsinga:
Thanks for the PR — handling `notifications/tools/list_changed` is something we do need.

A few things to address before we can move forward:

1. **Reference cycle (Codex P1):** The `Arc::clone(self) as Arc<dyn ToolCacheInvalidator>` creates a strong reference cycle (`ExtensionManager` → `extensions` → `McpClient` → `GooseClient` → `tool_cache_invalidator` → `ExtensionManager`). Please use `Arc::downgrade` with a `Weak`-based wrapper instead.

2. **Tests:** The two tests added are testing implementation details (that a field is stored, that a mock works) rather than the actual feature. `test_on_tool_list_changed_calls_invalidator` never actually calls `on_tool_list_changed` — it just calls `invalidate_tools()` directly on the stored field. Please either write tests that exercise the real notification path or remove the trivial tests.

3. **DCO:** The commit is missing the required `Signed-off-by` line — CI is failing on the DCO check. See the check output for instructions on how to fix this.

4. **Cargo.lock:** There's a spurious `goose-sdk` version bump in the lockfile that's unrelated to this change — please revert that.

/cc @alexhancock

Issue #8274, opened 2026-04-03
bug: MOIM injection produces hundreds of warnings per session (log spam)


eugenio:
## Description

The `tom` (Top Of Mind) extension's MOIM (Message of Immediate Mind) injection triggers warning messages on nearly every LLM call. In a single session, **142 occurrences** of these warnings were logged:

```
MOIM injection caused unexpected issues: ["Removed trailing assistant message"]
MOIM injection caused unexpected issues: ["Merged consecutive assistant messages", "Merged consecutive assistant messages"]
MOIM injection caused unexpected issues: ["Merged text content", "Merged consecutive user messages"]
```

## Environment

- Goose CLI on Windows 11 (latest version)
- Provider: OpenAI-compatible proxy (non-Anthropic models)
- Extensions enabled: tom (Top Of Mind) is a platform/bundled extension

## Steps to Reproduce

1. Run `goose session` with the `tom` extension enabled (default)
2. Have a conversation of 10+ turns
3. Check CLI logs — observe MOIM warnings on nearly every turn

## Expected Behavior

Either:
- MOIM injection should not produce invalid message sequences that require correction, OR
- The auto-correction should be silent (debug-level, not warn-level), since it is a known and handled side effect

## Actual Behavior

MOIM injection inserts a user message into the conversation on every LLM call, creating invalid sequences (consecutive same-role messages, trailing assistant messages). The correction code fixes these but emits a WARN-level log for each fix. In long sessions this produces massive log pollution (142+ warnings).

## Suggested Fix

Downgrade the `"MOIM injection caused unexpected issues"` message from WARN to DEBUG level, since the issues are automatically resolved and do not indicate a real problem. Alternatively, batch the corrections and emit a single summary warning per session instead of per-turn.

## Related Issues

- #6293 (Open) — Gemini empty replies caused by MOIM injection (functional breakage, not just log spam)
- #7425 (Closed, fixed) — 400 error from trailing assistant message after MOIM injection

This issue specifically tracks the **log spam** aspect, not the functional breakage for Gemini models.

Issue #8273, opened 2026-04-03
bug: Summon extension emits 'missing field name' frontmatter warnings on every startup


eugenio:
## Description

Every Goose session startup emits exactly 2 warnings from the summon extension:

```
WARN goose::agents::platform_extensions::summon — Failed to parse frontmatter: missing field `name`
```

The warnings repeat on every subagent invocation (e.g., via `delegate` or `load`), producing 18+ occurrences in a single long session.

## Environment

- Goose CLI on Windows 11 (latest version)
- Extensions enabled: nadirclaw-usage, analyze, todo, summon, developer, apps, tom, extensionmanager

## Steps to Reproduce

1. Run `goose session` with the `summon` extension enabled (default)
2. Check CLI logs at `%APPDATA%/Block/goose/data/logs/cli/`
3. Observe 2x `Failed to parse frontmatter: missing field 'name'` warnings at startup
4. Use any subagent delegation — warnings repeat

## Expected Behavior

No warnings on startup. Extension files that are not valid summon agents should be silently skipped, or the specific files causing the warning should be identified in the message.

## Actual Behavior

Two files are being scanned by the summon extension's frontmatter parser that lack the required `name` field. The warning does not identify which files are failing, making it impossible to fix from the user side.

## Impact

Cosmetic — no functional failure, but pollutes logs and makes it harder to spot real errors. In long sessions with multiple delegations, the repeated warnings add significant noise.

txhno:
I've opened a fix here: https://github.com/block/goose/pull/8338

The change keeps malformed agent-frontmatter warnings, but adds the file path and stops warning for plain markdown files in agent directories that just happen to have YAML frontmatter without a `name` field.

I couldn't run `cargo test` in this environment because the Rust toolchain isn't installed, so the PR notes that validation limitation explicitly.

Issue #8272, opened 2026-04-03
bug: Tool call JSON parse failures with large payloads and streaming fragments


eugenio:
## Description

Tool calls fail mid-session with JSON parse errors. Two distinct failure modes observed:

### Small payload truncation (shell tool)
```
Tool call could not be parsed: -32602: Could not interpret tool use parameters for id functions.shell:13: EOF while parsing a string at line 1 column 17
```

### Large payload truncation (write tool)
```
Tool call could not be parsed: -32602: Could not interpret tool use parameters for id functions.write:19: expected ',' or '}' at line 1 column 27409
Tool call could not be parsed: -32602: Could not interpret tool use parameters for id functions.write:22: expected ',' or '}' at line 1 column 58872
```

## Environment

- Goose CLI on Windows 11
- Provider: OpenAI-compatible proxy (NadirClaw) routing to Bedrock/Model Runner backends
- `GOOSE_TOOLSHIM=true`
- Models: DeepSeek V3.2 (eco), Kimi K2.5 (premium)

## Steps to Reproduce

1. Configure Goose with a non-Anthropic model via OpenAI-compatible proxy
2. Enable `GOOSE_TOOLSHIM=true`
3. Ask Goose to write a large file (>15KB) or execute a shell command
4. Observe JSON parse failure in logs

## Expected Behavior

Tool call arguments should be fully parsed regardless of payload size. If JSON is malformed, Goose should attempt repair (e.g., closing unclosed strings, fixing truncation) rather than failing silently.

## Actual Behavior

The `safely_parse_json` utility only handles control character escaping. It does not handle:
- Truncated JSON (unclosed strings, missing closing braces)
- Large payloads that exceed some internal buffer

## Related Issues

- #8167 (closed, no fix merged) — same EOF error for large payloads
- #2892 (closed, no fix merged) — unescaped control characters in JSON args
- #6688 (closed) — feature request for intelligent JSON repair layer for local models

Since #8167 and #2892 were closed without a definitive fix, reopening as a new issue to track.

DOsinga:
Thanks for the detailed report! This was fixed by PR #8208 (merged March 30, before this issue was filed) which switched the streaming path from raw `serde_json::from_str` to `safely_parse_json`, which includes `repair_truncated_json` for handling exactly the truncation/EOF errors you're seeing.

Additionally, PR #8310 (merged today) further improved the toolshim parsing path with better marker handling and more robust JSON extraction.

[...]

PR #8280, opened 2026-04-03
fix: repair truncated JSON, reduce summon/MOIM log noise

eugenio:
## Summary

Three targeted fixes for log noise and tool call reliability when using non-Anthropic models via OpenAI-compatible proxies:

- **Truncated JSON repair** (`safely_parse_json`): Adds a third fallback that closes unclosed strings, braces, and brackets in truncated tool call arguments. Fixes the `-32602: EOF while parsing a string at line 1 column 27409` errors that occur when models stream large `write` tool payloads. Ref: #8272
- **Summon frontmatter warning → debug**: Downgrades `"Failed to parse frontmatter: missing field name"` from WARN to DEBUG and includes the source file path. These warnings fire on every startup (2x) and every subagent invocation for non-extension markdown files that lack a `name` field. Ref: #8273
- **MOIM expected issues list**: Adds `"Merged text content"`, `"Removed trailing assistant message"`, and `"Trimmed trailing whitespace from assistant message"` to the expected-issues filter in `inject_moim()`. These are normal side-effects of MOIM injection but were triggering WARN-level logs on every LLM call (142+ per session). Ref: #8274

## Test plan

- [x] `test_safely_parse_json` updated — truncated JSON now repaired instead of erroring
- [x] `test_repair_truncated_json` added — covers unclosed strings, nested objects, arrays, large payloads
- [x] All 57 related unit tests pass (`providers::utils`, `conversation`, `agents::moim`)
- [ ] Manual: verify WARN-level summon/MOIM messages no longer appear in CLI logs

### DIFF
--- a/crates/goose/src/providers/utils.rs
+++ b/crates/goose/src/providers/utils.rs
@@ -533,8 +533,13 @@ fn repair_truncated_json(s: &str) -> Option<String> {
         return None;
     }
 
-    // Close an unclosed string
+    // Close an unclosed string. If the last character was a backslash
+    // (escape_next is still true), the appended quote would be escaped
+    // and produce invalid JSON. Drop the dangling backslash first.
     if in_string {
+        if escape_next {
+            repaired.pop(); // remove trailing backslash
+        }
         repaired.push('"');
     }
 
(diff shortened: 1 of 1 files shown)


eugenio:
Superseded by #8310 which includes the truncated JSON repair and all other fixes from this PR.

Issue #8275, opened 2026-04-03
bug: Toolshim OllamaInterpreter returns noop for all streaming tool call fragments


eugenio:
## Description

When `GOOSE_TOOLSHIM=true`, the toolshim sends individual streaming chunks to the local Ollama model for interpretation. Small models (e.g., `qwen3:1.7b`) return `noop` for every fragment, failing to recognize any tool calls.

### Example

A single `todo__todo_write` call streamed in 5 chunks:

| Chunk | Sent to Ollama | Result |
|---|---|---|
| 1 | `"name": "todo__todo_write"` | noop |
| 2 | `"arguments": {"content": "-` | noop |
| 3 | `[ ] Say hello in one word` | noop |
| 4 | `[x] Respond with one word greeting` | noop |
| 5 | `} }` | noop |

The interpreter receives each fragment in isolation and cannot determine it is part of a tool call.

## Environment

- Goose CLI on Windows 11
- Provider: OpenAI-compatible proxy (NadirClaw)
- `GOOSE_TOOLSHIM=true`
- Ollama model for interpreter: `qwen3:1.7b`
- Main model streams tool calls across multiple SSE chunks

## Expected Behavior

The toolshim should buffer streaming chunks and send a complete (or sufficiently large) tool call payload to the interpreter model, rather than sending each chunk individually.

## Actual Behavior

Each streaming chunk is sent separately to the OllamaInterpreter. Small models cannot recognize partial tool call fragments and return `noop` for all of them. The toolshim is effectively non-functional for streaming tool calls with small local models.

## Suggested Fix

Add a buffering/accumulation layer in the toolshim that collects streaming chunks until a complete tool call boundary is detected (e.g., matching opening/closing braces for JSON), then sends the assembled payload to the interpreter.

## Related Issues

- #8270 (Open) — Developer extension hardcodes Ollama for message augmentation
- #8269 (Open) — Toolshim documentation gap
- #3748 (Closed) — Qwen model tool call support with Ollama

sunilkumarvalmiki:
## Repro + fix validation (local branch)

I reproduced the fragment/noop behavior and validated a local fix.

### Before
Using the current toolshim interpreter prompt/format against Ollama:
- Fragmented streamed tool-call pieces each yielded `{"tool_calls":[{"name":"noop"...}]}`
- A full assembled payload yielded a real tool call

This confirms per-fragment interpretation causes false noops for streamed tool-call JSON.

### Fix applied locally
[...]

PR #8281, opened 2026-04-03
feat: add LlamaCppInterpreter for toolshim, fix streaming chunk handling

eugenio:
## Summary

- **LlamaCppInterpreter**: New `ToolInterpreter` implementation that uses the built-in llama.cpp backend instead of requiring an external Ollama server. Configured via `GOOSE_TOOLSHIM_MODEL` env var (default: `bartowski/Llama-3.2-1B-Instruct-GGUF:Q4_K_M`). Falls back to `OllamaInterpreter` when the `local-inference` feature is disabled or the model is unavailable.
- **Streaming fix**: `toolshim_postprocess` now runs only on the final streaming message (when `usage` is present) instead of on every intermediate chunk. This prevents the interpreter model from seeing incomplete text fragments and returning `noop` for all of them. Ref: #8275

### Motivation

The toolshim currently hardcodes an `OllamaInterpreter` that requires a running Ollama server on `localhost:11434`. This is a friction point for users who:
1. Don't want to install/manage a separate Ollama service
2. Already have the `local-inference` feature compiled in
3. Use Windows where Ollama runs as a service that consumes resources even when idle

The `LlamaCppInterpreter` reuses the existing `InferenceRuntime` and model cache, loading a small (~800MB) model on first use.

### Visibility changes

- `inference_engine` module: `pub(super)` → `pub(crate)` for `LoadedModel`, `TokenAction`, `validate_and_compute_context`, `create_and_prefill_context`, `generation_loop`, `build_context_params`, `estimate_max_context_for_memory`, `effective_context_size`
- `InferenceRuntime::get_or_create_model_slot`: `fn` → `pub fn`
- `LocalInferenceProvider::load_model_sync_public`: new `pub(crate)` wrapper

## Test plan

- [x] Compiles with `--features local-inference` (full default features)
- [x] Compiles without `local-inference` (graceful fallback to Ollama)
- [x] All existing unit tests pass (1014/1022 — 8 pre-existing env-specific failures)
- [ ] Manual: verify `goose session` with `GOOSE_TOOLSHIM=true` uses llama.cpp (check logs for `LlamaCpp tool interpreter response`)
- [ ] Manual: verify streaming tool calls are interpreted correctly with small models

### DIFF
--- a/crates/goose/src/agents/reply_parts.rs
+++ b/crates/goose/src/agents/reply_parts.rs
@@ -125,16 +125,21 @@ async fn toolshim_postprocess(
     response: Message,
     toolshim_tools: &[Tool],
 ) -> Result<Message, ProviderError> {
-    // Try llama.cpp interpreter first (no external server needed)
+    // Try llama.cpp interpreter first (no external server needed).
+    // Fall back to Ollama if creation OR augmentation fails.
     #[cfg(feature = "local-inference")]
     {
         match LlamaCppInterpreter::new() {
             Ok(interpreter) => {
-                return augment_message_with_tool_calls(&interpreter, response, toolshim_tools)
-                    .await
-                    .map_err(|e| {
-                        ProviderError::ExecutionError(format!("Failed to augment message: {}", e))
-                    });
+                match augment_message_with_tool_calls(&interpreter, response.clone(), toolshim_tools).await {
+                    Ok(msg) => return Ok(msg),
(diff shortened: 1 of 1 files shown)


eugenio:
Superseded by #8310 which includes all changes from this PR (LlamaCppInterpreter, streaming chunk accumulation, direct parsers) plus additional marker sanitization fixes.

PR #8286, opened 2026-04-03
Custom goose distros without re-builds

jamadeo:
Allows customizing a goose distribution without re-building. Can be done either within the electron package (resources dir) or externally in user data (e.g. `~/Library/Application\ Support/Goose/distro`)


jamadeo:
so one thing that this kind of misses is how it interacts with goose CLI. Once a config is written, the CLI has it, but bundling distro information in the electron app should ideally only apply to customizations to that UI

PR #8301, opened 2026-04-04
feat: add ssh-stateless bundled extension

fresh3nough:
## Summary
- add the built-in ssh-stateless extension and its Python implementation
- register the extension in bundled extension metadata for the desktop app and settings UI
- add Python and TypeScript tests for the new bundled extension

a la warp mode

### DIFF
--- a/extensions/ssh-stateless/pyproject.toml
+++ b/extensions/ssh-stateless/pyproject.toml
@@ -1,0 +1,13 @@
+[project]
+name = "ssh-stateless"
+version = "0.1.0"
+description = "Stateless SSH MCP extension for goose"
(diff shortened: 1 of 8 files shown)


fresh3nough:
working on this if it is of interest @DOsinga if not i will keep for myself

michaelneale:
hi @fresh3nough - not sure this should be bundled like this - could be a separate mcp, but as a bundled bit of python doesn't seem to fit. Is there a better place for this?

fresh3nough:
@michaelneale gotcha, yes i agree -> Extract the SSH logic into a standalone MCP server package is the way to go, i will get started on this today

PR #8304, opened 2026-04-04
fix: handle missing notification settings app on Linux

### LINKED ISSUE #8292: Desktop crash on Linux: Notifications "Open Settings" tries to spawn gnome-control-center (ENOENT)
Hi team,

I found another Desktop issue on Linux.

In **Settings → App → Notifications**, clicking **Open Settings** crashes Goose Desktop and shows this error:

```text
spawn gnome-control-center ENOENT
```

From the UI, Goose says that notifications are managed by the OS, so the button appears to be intended to open the system notification settings. However, on my setup it seems to assume a GNOME environment and tries to launch `gnome-control-center` directly.

That does not work on my system and causes Goose Desktop to fail instead of handling the situation gracefully.

## Expected behavior
- Goose Desktop should not crash when clicking **Open Settings**.
- On non-GNOME Linux desktops, it should either:
  - use a desktop-appropriate settings integration,
  - use xdg/portal-based behavior if available,
  - or show a non-fatal message that system notification settings cannot be opened automatically.

## Actual behavior
- Clicking **Open Settings** leads to an error screen.
- The error shown is:

```text
spawn gnome-control-center ENOENT
```

## Environment
- OS: Arch Linux
- Session: Wayland / Hyprland
- Goose Desktop version: 1.29.0

Thanks.

mvanhorn:
## Summary

Fixes a crash when clicking "Open Settings" in Settings > App > Notifications on Linux desktops that don't have GNOME, KDE, or XFCE installed.

## Changes

In `ui/desktop/src/main.ts`, the `open-notifications-settings` IPC handler used `spawn()` inside `try/catch` blocks for each desktop environment. Node.js `spawn()` does not throw synchronously on ENOENT; it emits the error asynchronously via the ChildProcess 'error' event. The `try/catch` blocks never fired, so the app crashed with `spawn gnome-control-center ENOENT`.

Fix: added a `canSpawn(cmd)` helper that uses `execFileSync('which', [cmd])` to check if the binary exists before calling `spawn()`. Each DE check now uses `if (canSpawn(...))` instead of `try { spawn(...) } catch`. When no settings app is found, returns `false` instead of crashing.

## Testing

Verified the logic by reading the Node.js `child_process` docs confirming `spawn()` ENOENT behavior. The fix is a straightforward control flow change with no new dependencies.

Fixes #8292

This contribution was developed with AI assistance (Codex).

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -1469,11 +1469,6 @@ ipcMain.handle('open-notifications-settings', async () => {
         return true;
       }
 
-      // Fallback: Try to open general settings
-      if (canSpawn('gnome-control-center')) {
-        spawn('gnome-control-center');
-        return true;
-      }
       console.warn('Could not find a suitable settings application for Linux');
       return false;
     } else {


DOsinga (approved):
Good fix — `spawn()` indeed doesn't throw synchronously on ENOENT, so the existing try/catch was dead code. The `canSpawn` approach using `which` is clean and correct.

I pushed a small cleanup commit removing the redundant fallback `canSpawn('gnome-control-center')` check — if it wasn't found the first time, checking again without args won't help.

One thing to fix: your original commit is missing the DCO sign-off. Please amend it with `git commit --amend -s` and force-push so the DCO check passes.

mvanhorn:
Thanks for the review and cleanup, @DOsinga. Good call removing the redundant fallback check.

PR #8310, opened 2026-04-04
fix: prevent tool-use marker leakage in toolshim output

eugenio:
## Summary

- **Fix streaming marker leakage**: Accumulate full toolshim response before post-processing so markers spanning multiple stream chunks (`<|tool_calls_section_begin|>`, `<|tool_call_begin|>`, etc.) are detected and stripped before reaching the UI.
- **Fix parse-failure marker leakage**: Add catch-all `sanitize_residual_markers()` on every exit path of `augment_message_with_tool_calls`, so malformed markers are stripped even when parsing yields zero tool calls.
- **Add direct tokenized + inline JSON parsers**: Bypass the interpreter for faster, more reliable tool call extraction from model output.
- **Add local llama.cpp toolshim backend**: New `GOOSE_TOOLSHIM_BACKEND=local` option alongside the existing Ollama backend.
- **Add frontend defense-in-depth**: `stripToolCallMarkers()` in `message.ts` as a last-resort UI-level sanitizer.
- **Add truncated JSON repair**: `safely_parse_json()` now recovers from truncated/malformed JSON tool arguments.
- **Increase main thread stack size**: 8MB stack to prevent overflows on Windows.

## Test plan

- [x] 15 toolshim unit tests pass (4 new regression tests for marker leakage)
- [x] Smoke test: plain reply flow — clean output, no markers
- [x] Smoke test: tool-use flow — tool call rendered cleanly, no raw markers in output
- [x] CI build and test suite


crates/goose/src/providers/toolshim.rs
```diff
@@ -47,6 +49,481 @@ use uuid::Uuid;
...
+        out.push('\\');
+    }
+
+    out
+}
+
+fn parse_json_value_tolerant(input: &str) -> Option<Value> {
+    serde_json::from_str::<Value>(input).ok().or_else(|| {
+        let escaped = escape_invalid_backslashes_in_json_strings(input);
+        serde_json::from_str::<Value>(&escaped).ok()
```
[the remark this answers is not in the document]
eugenio: This is a known trade-off. The first `serde_json::from_str` path succeeds because `{"path":"C:\temp\foo"}` is technically valid JSON — `\t` and `\f` are legal escape sequences. Running the backslash repair unconditionally would risk corrupting valid JSON that intentionally uses `\t`/`\n`/`\r`.

In practice, models that know these markers (Mistral, etc.) emit properly escaped `\` for Windows paths — verified by the existing `parses_tokenized_tool_call_with_windows_path_arguments` test. This edge case would require the model to emit raw unescaped backslashes, which is a pre-existing limitation of any JSON-based tool call format, not a regression from this PR.

michaelneale:
thanks @eugenio  - pretty chunky changes - did you find this worked ok for you? I am curious about the 8MB stack bit, that is probably the most sensitive change - LMK, I like this.

crates/goose-cli/src/main.rs
```diff
@@ -17,3 +16,21 @@ async fn main() -> Result<()> {
 
     result
 }
+
+fn main() -> Result<()> {
+    let handle = std::thread::Builder::new()
+        .name("goose-cli-main".to_string())
```
michaelneale, line 22: not sure about name and the limit here - wonder if there is another way. FYI we are moving to a TUI in ui/text, and goose serve - so may want to consider that path as well?

ui/desktop/src/types/message.ts
```diff
@@ -94,9 +94,23 @@ export function getTextAndImageContent(message: Message): {
     }
   }
 
+  // Strip tool call markers only from assistant messages (not user-authored text)
```
michaelneale: I think this should be done lower level, not just in desktop?

michaelneale: I think could drop changes to message.ts if it is working.

michaelneale:
@eugenio happy to take this on and clean it up if you like

eugenio:
Thank you @michaelneale. The changes  seem to be working well on my side. I haven't had any more json leaking since then, but I didn't try every model. So it might need more thorough testing.

eugenio:
> @eugenio happy to take this on and clean it up if you like

Thank you, that would be awesome! Just let me know if you need anything from my side!

iaincalderfh:
FYI I tested this patch using Goose v1.30.0 CLI and Qwen3-Coder-Next from an OpenAI-compatible provider.  Before the patch:

`-32602: Could not interpret tool use parameters for id call_79c77242cf4c4c7ebcdbd62c: EOF while parsing an object at line 1 column 1A tool call could not be parsed — the response may have been truncated. Try breaking the task into smaller steps or resending your message.`. 

After the patch, it works well and I have not seen the tool error again.

michaelneale (approved):

PR #8312, opened 2026-04-04
fix: don't show onboarding when server is unreachable

### LINKED ISSUE #8311: fix: improve disconnect recovery UX — don't show onboarding to existing users
## Summary

When a session fails with "Lost connection to server", the only recovery option ("Go home") can dump existing users into the onboarding Welcome screen. The error screen itself offers no retry, no context, and no way to recover.

### Bug 1: "Go home" sends configured users to onboarding

**Repro:** Have two Goose windows open. Kill/restart goosed (or lose network). One window shows "Failed to Load Session". Click "Go home".

**What happens:** `setView('chat')` navigates to `/`, which is wrapped in `OnboardingGuard`. The guard calls `readConfig('GOOSE_PROVIDER')` — a POST to goosed. If goosed is still recovering, the catch block sets `hasProvider = false` and renders the full "Welcome to goose — Connect an AI model provider" screen.

**File:** `ui/desktop/src/components/onboarding/OnboardingGuard.tsx:42-44`
```tsx
} catch (error) {
  console.error('Error checking provider:', error);
  setHasProvider(false);  // ← Wrong: assumes no provider when server is just unreachable
}
```

**Impact:** Existing users think they need to reconfigure. Some do, breaking their setup.

### Bug 2: No retry or context recovery on the error screen

**File:** `ui/desktop/src/components/BaseChat.tsx:342-370`

The error screen shows a red banner and a single "Go home" button. No retry. No session ID. No way to recover the last message. The SSE layer (`useSessionEvents.ts`) already has reconnect logic with exponential backoff — but the UI error screen was never updated to leverage it.

### PR1: OnboardingGuard fail-open (1 file, +9/-2 lines)

Persist a `localStorage` flag (`goose_has_provider`) on successful provider detection. On config read failure, check the flag before falling through to onboarding. Existing users see the error screen instead of Welcome.

```diff
+        const configured = provider.trim() !== '';
+        setHasProvider(configured);
+        if (configured) {
+          localStorage.setItem('goose_has_provider', 'true');
+        }
       } catch (error) {
         console.error('Error checking provider:', error);
-        setHasProvider(false);
+        const previouslyConfigured = localStorage.getItem('goose_has_provider') === 'true';
+        setHasProvider(previouslyConfigured);
```

### PR2: Error screen retry + new session buttons (2 files, +33/-12 lines)

Replace the single "Go home" button with:
1. **"Retry connection"** — clears error state, re-triggers session load via a `retryCount` state in the `useEffect` dependency array
2. **"New session"** — navigates to hub (same as old "Go home" but clearer label)
3. **Session ID** — displayed and selectable for context recovery / bug reports

## Slack Evidence

| Date | Channel | Quote |
|---|---|---|
| 03/04 | #goose-help | "I'm running into Failed to Load Session / Stream error: Lost connection to server" |
| 03/04 | #ai-help-community | "I am facing Failed to load session issue... I restarted goose and restarted my machine" |
| 03/04 | #goose-help | "The Goose error reads: Failed to Load Session - Stream Error - Lost connection to Server" |
| 24/03 | #goose-slackbot | "exposing some kind of 'clear/reset session' functionality would help this which has been a common request" |
| 16/03 | #nexus | "every now and then im getting this error, and have to abandon the session and start a new one. Is there a way to recover from this?" |

## Prior Art

| Reference | Status | Notes |
|---|---|---|
| [#1520](https://github.com/block/goose/issues/1520) — Retry button | Closed | Never shipped for disconnect errors |
| [#7972](https://github.com/block/goose/pull/7972) — Safe agent recovery | Closed | @kyledef attempted this. Closed for process reasons, not technical. |
| [#7834](https://github.com/block/goose/pull/7834) — SSE separation | Merged | SSE reconnects automatically now, but UI error screen wasn't updated |
| [#8298](https://github.com/block/goose/issues/8298) — Preserve context on retry | Open | Different scope — automated retry context, not disconnect screen |

## Edge Cases

1. **First-time user with no localStorage flag** — `getItem` returns `null`, `=== 'true'` is false → falls through to onboarding as expected. No regression.
2. **User clears browser storage** — same as first-time, falls through to onboarding. Acceptable — they'll set up once and the flag is re-persisted.
3. **Retry when server is permanently down** — retry clears error, shows loading state, then error reappears after `resumeAgent` fails again. User can retry again or click "New session". No infinite loop.
4. **Retry when server recovers** — retry clears error, `resumeAgent` succeeds, session loads normally. Best case.

jeffa-block:
When `readConfig` fails (goosed down), `OnboardingGuard` falls through to `setHasProvider(false)`, showing the Welcome screen to existing users. This is confusing — they think they need to reconfigure.

**Fix:** Persist a `localStorage` flag (`goose_has_provider`) on successful provider detection. On config read failure, check the flag before falling through to onboarding. Existing users see the error screen instead of Welcome.

Fixes #8311 (PR 1 of 2)

## Changes
- **OnboardingGuard.tsx**: persist flag on success; check it on failure before showing onboarding (+9/-2)
- **OnboardingGuard.test.tsx**: 6 tests covering success path, fail-open for existing users, first-time user fallback, and edge cases (+149)

## Testing
1. Configure a provider normally → localStorage flag is set
2. Kill goosed → navigate to `/` → see error screen, NOT Welcome
3. First-time user (no flag) → still sees Welcome as expected
4. `pnpm vitest run src/components/onboarding/OnboardingGuard.test.tsx` → 6/6 pass

### DIFF
--- a/ui/desktop/src/components/onboarding/OnboardingGuard.test.tsx
+++ b/ui/desktop/src/components/onboarding/OnboardingGuard.test.tsx
@@ -1,0 +1,149 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, waitFor } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+import OnboardingGuard from './OnboardingGuard';
+
+// --- localStorage shim (jsdom in this repo has a broken localStorage) ---
+
+const localStorageStore: Record<string, string> = {};
+
+const localStorageMock = {
+  getItem: vi.fn((key: string) => localStorageStore[key] ?? null),
+  setItem: vi.fn((key: string, value: string) => {
+    localStorageStore[key] = value;
+  }),
+  removeItem: vi.fn((key: string) => {
+    delete localStorageStore[key];
+  }),
+  clear: vi.fn(() => {
+    for (const key of Object.keys(localStorageStore)) {
+      delete localStorageStore[key];
+    }
+  }),
+  get length() {
+    return Object.keys(localStorageStore).length;
+  },
+  key: vi.fn((index: number) => Object.keys(localStorageStore)[index] ?? null),
+};
+
+Object.defineProperty(globalThis, 'localStorage', {
+  value: localStorageMock,
+  writable: true,
+});
+
+// --- Mocks ---
+
+const mockRead = vi.fn();
+
+vi.mock('../ConfigContext', () => ({
+  useConfig: () => ({
+    read: mockRead,
+    upsert: vi.fn(),
+    getProviders: vi.fn().mockResolvedValue([]),
+  }),
+}));
+
+vi.mock('../ModelAndProviderContext', () => ({
+  useModelAndProvider: () => ({
+    refreshCurrentModelAndProvider: vi.fn(),
+  }),
+}));
+
+vi.mock('../../utils/analytics', () => ({
+  trackOnboardingStarted: vi.fn(),
+  trackOnboardingCompleted: vi.fn(),
+  trackOnboardingProviderSelected: vi.fn(),
+  trackTelemetryPreference: vi.fn(),
+  setTelemetryEnabled: vi.fn(),
+}));
+
+// Mock child components that have their own complex dependencies
+vi.mock('./ProviderSelector', () => ({
+  default: () => <div data-testid="provider-selector">ProviderSelector</div>,
+}));
+
+vi.mock('./OnboardingSuccess', () => ({
+  default: () => <div data-testid="onboarding-success">OnboardingSuccess</div>,
+}));
+
+const renderGuard = () =>
+  render(
+    <MemoryRouter>
+      <OnboardingGuard>
+        <div data-testid="app-content">App Content</div>
+      </OnboardingGuard>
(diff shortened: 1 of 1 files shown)


lifeizhou-ap:
Hi @jeffa-block Thank you so much for the issue analysis and fix!

With the local storage approach, we have to sync the local storage with other features.  For example, use can click "reset provider and model" in the settings, and in this case we need to sync the local storage to set the has provider to false. 

The main issue is that the error scenario is not handled.  I created another PR to handle the error https://github.com/aaif-goose/goose/pull/8394 WDYT?

jeffa-block:
yeah the localStorage sync concern is fair — I had a second PR planned for retry + error screen but yours handles it earlier in the flow which is cleaner. will close this in favour of #8394 👍

DOsinga:
For context on the closure: the underlying bug has already been fixed on main by @lifeizhou-ap in #8394, which takes a different approach — retry logic with backoff + a dedicated error screen with a Retry button, rather than a localStorage flag. This avoids the stale-flag issue that Codex flagged in review. Thanks for the well-written diagnosis in #8311 though, Jeff — it was spot on. 🙏

PR #8313, opened 2026-04-04
feat: add retry and new session buttons to disconnect error screen

### LINKED ISSUE #8311: fix: improve disconnect recovery UX — don't show onboarding to existing users

jeffa-block:
Replace the single "Go home" button on the disconnect error screen with:
- **Retry connection** — clears error state, re-triggers session load via `retryCount` in the effect deps
- **New session** — clearer label for the existing navigation
- **Session ID** — copyable for bug reports and session recovery

Fixes #8311 (PR 2 of 2)

## Changes
- **useChatStream.ts**: add `retrySessionLoad()` callback using `retryCount` state in session load effect deps (+11/-2)
- **BaseChat.tsx**: replace error screen with retry + new session buttons + session ID display (+22/-10)

## Testing
Manual — BaseChat has no existing test coverage and is tightly coupled to `useChatStream` internals. If reviewers want unit tests, happy to extract a `<SessionErrorScreen>` component as a follow-up.

1. Trigger "Lost connection to server" error (kill goosed mid-session)
2. Click "Retry connection" → loading state → session resumes (if server is back) or error reappears
3. Click "New session" → navigates to hub
4. Session ID is visible and selectable


michaelneale:
thanks @jeffa-block got a minute to update to main (looks like some collision, but I assume still a bug)

DOsinga (commented):
I merged main (which had an i18n conflict) and fixed the cache bug Codex flagged — `retrySessionLoad` now calls `resultsCache.delete(sessionId)` before bumping `retryCount`, so the retry actually hits the backend instead of returning stale cached data. Also added the new i18n keys (`retryConnection`, `newSession`) to replace the removed `goHome` key.

One question: when does the user actually see this disconnect screen? We had a regression recently where the error showed up because goosed had died entirely. In that case, "Retry connection" won't help — the backend process is gone and there's nothing to reconnect to. Is this PR specifically targeting transient network blips (where goosed is still alive), or should the retry also attempt to restart goosed? If it's the latter, we should probably surface that distinction in the UI (e.g. "Server is not running" vs "Lost connection") so users know whether retry is likely to work.

DOsinga:
Closing — the main scenario where users see this disconnect screen is when goosed has died entirely, and in that case a "Retry connection" button won't help since there's nothing to reconnect to. The question about whether this should also attempt to restart goosed (or at least distinguish "server not running" from "transient blip") was raised in review but never addressed. Happy to revisit if someone wants to pick this up with that distinction built in.

DOsinga:
going to close this due to staleness. feel free to reopen and we can discuss my question

Issue #8317, opened 2026-04-05
Bug: dynamic_models field in custom provider config is ignored


yart:
### Description

The `dynamic_models` field in custom provider JSON configuration is defined in the schema but **not used** in the actual model loading logic. This causes Goose to always fetch models dynamically from the provider's `/v1/models` API endpoint, ignoring the static model list defined in the configuration file.

### Expected Behavior

When `dynamic_models: false` is set in a custom provider config:
- Goose should use **only** the models listed in the `models` array
- The `/v1/models` API endpoint should **not** be called

When `dynamic_models: true` or `dynamic_models: null`:
- Goose should fetch models dynamically from the API (current behavior)

### Actual Behavior

Regardless of the `dynamic_models` value, Goose always:
1. Attempts to fetch models from `/v1/models` API
2. Only falls back to the static list if the API returns 404
3. If the API succeeds, the static model list is completely ignored

### Steps to Reproduce

1. Create a custom provider config at `~/.config/goose/custom_providers/custom_omniroute.json`:
```json
{
  "name": "custom_omniroute",
  "engine": "openai",
  "display_name": "OmniRoute",
  "api_key_env": "CUSTOM_OMNIROUTE_API_KEY",
  "base_url": "http://localhost:20128",
  "dynamic_models": false,
  "models": [
    {"name": "kr/claude-sonnet-4.5", "context_limit": 128000},
    {"name": "codex/gpt-5.4", "context_limit": 128000}
  ]
}
```

2. Start a provider that returns 500+ models from `/v1/models`
3. Launch Goose and check available models
4. **Result**: All 500+ models from the API are shown, not just the 2 configured models

### Root Cause

In `crates/goose/src/providers/openai.rs`, the `from_custom_config` method sets `custom_models`:

```rust
let custom_models = if !config.models.is_empty() {
    Some(config.models.iter().map(|m| m.name.clone()).collect())
} else {
    None
};
```

But `fetch_supported_models` always tries the API first:

```rust
async fn fetch_supported_models(&self) -> Result<Vec<String>, ProviderError> {
    if let Some(custom_models) = &self.custom_models {
        match self.fetch_models_from_api().await {
            Ok(models) => return Ok(models),  // ← API result used, ignoring custom_models
            Err(e) if e.is_endpoint_not_found() => {
                return Ok(custom_models.clone());  // ← Only used on 404
            }
            Err(e) => return Err(e),
        }
    }
    self.fetch_models_from_api().await
}
```

[...]

PR #8319, opened 2026-04-05
[claude-generated] fix: respect dynamic_models=false in custom provider config

sjhddh:
## Summary

- Fixes `dynamic_models` field in custom provider config being completely ignored (#8317)
- When `dynamic_models: false` is set, Goose now uses the static `models` list from the config instead of always fetching from the provider's `/v1/models` API endpoint
- Applies the fix to both OpenAI and Anthropic provider engines

## Problem

The `dynamic_models` field was defined in `DeclarativeProviderConfig` but never checked in either provider's `from_custom_config()` method. Two issues:

1. **`from_custom_config`**: `custom_models` was set whenever the `models` array was non-empty, regardless of the `dynamic_models` flag
2. **`fetch_supported_models`**: Even when `custom_models` was set, the method always tried the API first and only used the static list as a 404 fallback — so providers returning hundreds of models from their API would ignore the user's curated list entirely

## Fix

**In `from_custom_config` (both `openai.rs` and `anthropic.rs`):**
```rust
// Before: always set custom_models if models list is non-empty
let custom_models = if !config.models.is_empty() { ... }

// After: only set custom_models when dynamic_models is explicitly false
let custom_models = if config.dynamic_models == Some(false) && !config.models.is_empty() { ... }
```

**In `fetch_supported_models` (both providers):**
```rust
// Before: try API first, only use custom_models on 404
if let Some(custom_models) = &self.custom_models {
    match self.fetch_models_from_api().await {
        Ok(models) => return Ok(models),  // API wins, static list ignored
        ...
    }
}

// After: custom_models means the user wants static list only
if let Some(custom_models) = &self.custom_models {
    return Ok(custom_models.clone());
}
```

## Behavior matrix

| `dynamic_models` | `models` array | Result |
|---|---|---|
| `false` | non-empty | Use static list only (fixed) |
| `true` | any | Fetch from API (unchanged) |
| `null`/absent | any | Fetch from API (unchanged) |

## Test plan

- [ ] Verify with a custom provider config that has `dynamic_models: false` and a static model list — only the listed models should appear
- [ ] Verify default behavior (no `dynamic_models` field) still fetches from API
- [ ] Verify `dynamic_models: true` still fetches from API
- [ ] Run existing test suite: `cargo test -p goose`

### DIFF
--- a/crates/goose/src/providers/anthropic.rs
+++ b/crates/goose/src/providers/anthropic.rs
@@ -121,7 +121,9 @@ impl AnthropicProvider {
             ));
         }
 
-        let custom_models = if !config.models.is_empty() {
+        // When dynamic_models is explicitly false and a static model list is provided,
+        // use only the static list and skip the models API call.
+        let custom_models = if config.dynamic_models == Some(false) && !config.models.is_empty() {
             Some(config.models.iter().map(|m| m.name.clone()).collect())
         } else {
             None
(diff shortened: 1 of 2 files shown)


DOsinga:
Thanks for the PR — the underlying issue (#8317) is real, but the Codex review flagged a regression that needs to be addressed before this can move forward: when `create_custom_provider` sets `dynamic_models: None`, the new logic drops the static model fallback, breaking providers created via CLI/UI whose backends don't implement `/v1/models`.

Please see our guidelines on addressing automated review feedback: https://github.com/block/goose/blob/main/CONTRIBUTING.md#ai-code-reviews

PR #8328, opened 2026-04-05
fix(cli): display platform-correct secrets path in keyring config dialog

### LINKED ISSUE #8326: fix(cli): keyring config dialog shows hardcoded Unix path instead of platform-specific secrets path
## Description

When running `goose configure` > `goose settings` > `Secret Storage`, the warning and confirmation messages display a hardcoded `~/.config/goose/secrets.yaml` path regardless of the platform. This path is only correct on Linux.

On Windows, the actual secrets file path is `C:\Users\<user>\AppData\Roaming\Block\goose\config\secrets.yaml`, and on macOS it's `~/Library/Application Support/Block/goose/secrets.yaml`.

The rest of the same dialog correctly uses the platform-specific path (e.g. "Configuration saved successfully to C:\Users\iamsu\AppData\Roaming\Block\goose\config\config.yaml"), making the hardcoded `~/.config/goose/` inconsistent and misleading.

## Steps to Reproduce

1. On Windows (or macOS), run `goose configure`
2. Select **goose settings** > **Secret Storage**
3. Observe the warning: `Note: Disabling the keyring stores secrets in a plain text file (~/.config/goose/secrets.yaml)`
4. Select "File-based Storage"
5. Observe the confirmation: `Secret storage set to file (~/.config/goose/secrets.yaml). Keep this file secure!`

## Expected Behavior

The messages should show the platform-correct path, e.g.:
- Windows: `C:\Users\iamsu\AppData\Roaming\Block\goose\config\secrets.yaml`
- macOS: `/Users/iamsu/Library/Application Support/Block/goose/secrets.yaml`
- Linux: `/home/iamsu/.config/goose/secrets.yaml`

## Actual Behavior

Both messages show `~/.config/goose/secrets.yaml` on all platforms.

## Environment

- **OS & Arch:** Windows 11 x86_64
- **Interface:** CLI
- **Version:** 1.29.1

## Root Cause

In `crates/goose-cli/src/commands/configure.rs`, lines 1481 and 1508 use hardcoded string literals instead of computing the path via `Paths::config_dir().join("secrets.yaml")` (which is already imported and used elsewhere in the same file).

SushaanthSrinivasan:
## Summary

- The keyring configuration dialog in `goose configure` hardcoded `~/.config/goose/secrets.yaml` in warning and confirmation messages
- This path is only correct on Linux — on Windows the actual path is `C:\Users\<user>\AppData\Roaming\Block\goose\config\secrets.yaml`, on macOS it's `~/Library/Application Support/Block/goose/`
- Replaced hardcoded string with `Paths::config_dir().join("secrets.yaml")`, which is already imported and used elsewhere in the same file

Fixes #8326

## Before / After

**Before** (hardcoded Unix path on Windows):
old

**After** (correct platform path):
new

## Test plan

- [x] Built and tested on Windows 11 — path now shows correct platform path
- [x] Compilation verified on Linux (WSL) via `cargo check`

### DIFF
--- a/crates/goose-cli/src/commands/configure.rs
+++ b/crates/goose-cli/src/commands/configure.rs
@@ -1478,7 +1478,11 @@ pub fn configure_keyring_dialog() -> anyhow::Result<()> {
     };
 
     let _ = cliclack::log::info(format!("Current secret storage: {}", current_status));
-    let _ = cliclack::log::warning("Note: Disabling the keyring stores secrets in a plain text file (~/.config/goose/secrets.yaml)");
+    let secrets_path = Paths::config_dir().join("secrets.yaml");
+    let _ = cliclack::log::warning(format!(
+        "Note: Disabling the keyring stores secrets in a plain text file ({})",
+        secrets_path.display()
+    ));
 
     let storage_option = cliclack::select("How would you like to store secrets?")
         .item(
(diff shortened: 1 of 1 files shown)


michaelneale:
thanks @SushaanthSrinivasan

PR #8335, opened 2026-04-06
feat: add Ollama Cloud as a declarative provider

michaelneale:
Adds Ollama Cloud as a declarative provider — a minimal 13-line JSON config pointing at ollama.com's OpenAI-compatible API with dynamic model discovery.

Simplified version of #8189 — just the declarative JSON and docs entry, no other changes needed.

cc @vincenzopalazzo @DOsinga


vincenzopalazzo:
LGTM

crates/goose/src/providers/declarative/ollama_cloud.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "ollama_cloud",
+  "engine": "openai",
+  "display_name": "Ollama Cloud",
+  "description": "Access hosted models on ollama.com via OpenAI-compatible API",
+  "api_key_env": "OLLAMA_CLOUD_API_KEY",
+  "base_url": "https://ollama.com/v1/chat/completions",
+  "models": [],
```
vincenzopalazzo, line 8: are the module fetch in automagically now in goose?

michaelneale: models? I think so (worth a test) - FYI ths was just so you could try it regarding your PR (if you want to do it there, that is cool)

vincenzopalazzo: I can try to test today and I can also push it to the finish line if you want :)

vincenzopalazzo: Ok I imported your work on https://github.com/aaif-goose/goose/pull/8189#issuecomment-4199474745 and make the test where the model are fetch automagically (when dynamic is set to `true` in the JSON)! 

Hope this helps

michaelneale:
thanks for this, I think a reasonable idea (although is it a bit weird if it is sharing ollamas remote resources?) - wold it also make sense to have ollama itself as a fairly vanilla backend to share if people already run it? wdyt? (just for single models, not splits etc)

vincenzopalazzo:
@michaelneale Just dropped a comment in https://github.com/aaif-goose/goose/pull/8189#issuecomment-4204602044

jamadeo:
let's merge #8189, it just awaits a documentation code owner now

PR #8336, opened 2026-04-06
fix: include unmapped models in recommended list

### LINKED ISSUE #8321: Goose silently drops unmapped models from /v1/models, breaking OpenRouter, OmniRoute, and cutting off new models
This is related to #8317, highlighting a broader architectural flaw in how Goose handles model lists from providers.

Currently, Goose fetches the list of available models from a provider's API (e.g., OpenRouter or an OpenAI-compatible proxy like OmniRoute) but then **silently drops any model that is not hardcoded in its internal `CanonicalModelRegistry`**.

**The Code at Fault (`crates/goose/src/providers/base.rs`):**
```rust
async fn fetch_recommended_models(&self) -> Result<Vec<String>, ProviderError> {
    let all_models = self.fetch_supported_models().await?;
    // ...
    let mut models_with_dates: Vec<(String, Option<String>)> = all_models
        .iter()
        .filter_map(|model| {
            // SILENT DROP: If map_to_canonical_model fails, the model is hidden from the UI
            let canonical_id = map_to_canonical_model(provider_name, model, registry)?;
            // ...
```

**Why this is broken:**
1. **New models are instantly unusable:** Models like `claude-opus-4-6-thinking` or `gemini-3.1-pro` cannot be selected in the UI because Goose's internal registry is outdated. Users are forced to type them manually.
2. **OpenRouter is crippled:** OpenRouter returns hundreds of models (e.g., `openrouter/qwen/qwen3.6-plus:free`), but Goose hides almost all of them because it doesn't have a mapping for every OpenRouter ID.
3. **Custom Proxies (OmniRoute, LiteLLM) break:** If a proxy uses prefixes or custom aliases, Goose drops them entirely.

There is a hidden flag `skip_canonical_filtering` for declarative providers, but it is entirely undocumented and fails to resolve the issue for built-in providers (like OpenRouter) or complex setups.

**Expected behavior:**
Goose should trust the provider. If an endpoint returns a model, it should be available in the dropdown UI. Filtering by a hardcoded registry should be **opt-in** (or at least provide a "Show All" toggle in the UI), rather than forcefully dropping models that the user's API key actually has access to.

Bahtya:
## Summary

Fixes #8321

Previously, models not found in `CanonicalModelRegistry` were silently dropped from the recommended models list. This made newly released models (e.g. `claude-opus-4-6-thinking`, `gemini-3.1-pro`, OpenRouter models) invisible in the UI dropdown, forcing users to type model names manually.

## Fix

In `fetch_recommended_models()`, when `map_to_canonical_model()` returns `None` (model not in registry), include the model anyway with `None` as its release date. Unmapped models are sorted after dated models in the dropdown.

**Before:** Unmapped models silently dropped → not selectable in UI
**After:** All models from provider's `/v1/models` endpoint are available

## Changes

- `crates/goose/src/providers/base.rs`: Changed `filter_map` to include unmapped models instead of silently dropping them (10 lines added, 1 removed)

## Testing

The fix preserves existing behavior for mapped models (sorted by release date) while adding unmapped models at the end of the list.

### DIFF
--- a/crates/goose/src/providers/base.rs
+++ b/crates/goose/src/providers/base.rs
@@ -1205,7 +1205,9 @@ mod tests {
         assert!(is_likely_non_chat_model("qwen3-asr-flash"));
         assert!(is_likely_non_chat_model("imagen-4.0-generate-001"));
         assert!(is_likely_non_chat_model("stable-diffusion-xl"));
-        assert!(is_likely_non_chat_model("gemini-2.0-flash-exp-image-generation"));
+        assert!(is_likely_non_chat_model(
+            "gemini-2.0-flash-exp-image-generation"
+        ));
 
         // Should NOT be filtered — these are chat models
         assert!(!is_likely_non_chat_model("gpt-4o"));
--- a/crates/goose/src/providers/base.rs
+++ b/crates/goose/src/providers/base.rs
@@ -603,10 +603,19 @@ pub trait Provider: Send + Sync {
         let provider_name = self.get_name();
 
         // Get all text-capable models with their release dates
+        // Models that can't be mapped to a canonical ID are still included
+        // (without a release date) so users can select newly released models.
         let mut models_with_dates: Vec<(String, Option<String>)> = all_models
             .iter()
             .filter_map(|model| {
-                let canonical_id = map_to_canonical_model(provider_name, model, registry)?;
+                let canonical_id = match map_to_canonical_model(provider_name, model, registry) {
+                    Some(id) => id,
+                    None => {
+                        // Unmapped model — include it anyway so new/unknown models
+                        // are available in the UI dropdown.
+                        return Some((model.clone(), None));
+                    }
+                };
 
                 let (provider, model_name) = canonical_id.split_once('/')?;
                 let canonical_model = registry.get(provider, model_name)?;
--- a/crates/goose/src/providers/base.rs
+++ b/crates/goose/src/providers/base.rs
@@ -500,6 +500,27 @@ pub enum PermissionRouting {
     Noop,
 }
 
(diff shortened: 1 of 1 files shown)


michaelneale:
this seems reasonable to me. Any reason why not @DOsinga as I am sure we have gone back and forth on this. @Bahtya mind signing the SCO so can get the build clean? thanks!

Bahtya:
Thanks @michaelneale! I've amended the commit with the DCO sign-off and force-pushed. Let me know if there's anything else needed.

Bahtya:
Thanks @michaelneale! I've amended the commit with the DCO sign-off and force-pushed. Let me know if there's anything else needed.

jh-block:
As codex says, won't this cause models that don't support chat (vision/audio models with no text modality) to appear in the drop-down?

Bahtya:
@jh-block Good catch! You're right that unmapped models bypassed the canonical registry's text-modality and tool_call filters. I've addressed this in 212cd3c by adding a name-based heuristic (`is_likely_non_chat_model`) that filters out known non-chat model patterns from the unmapped path:

- **Embedding models**: names containing "embed" (text-embedding-*, cohere-embed-*)
- **TTS models**: names containing "tts"
- **Image generation**: "dall", "imagen", "stable-diffusion", "image-generation"
- **Speech recognition**: "whisper", "-asr", "asr-"

This is a best-effort heuristic since we have no metadata for unmapped models. For mapped models, the existing canonical registry filters (text modality + tool_call support) continue to apply as before. The tradeoff is that a few edge-case non-chat models with unusual names could still slip through, but this covers all the common patterns visible in our canonical registry data.

PR #8338, opened 2026-04-06
fix: reduce summon agent frontmatter warning noise

### LINKED ISSUE #8273: bug: Summon extension emits 'missing field name' frontmatter warnings on every startup

txhno:
Fixes #8273

## Summary
- stop warning when agent directories contain markdown files with YAML frontmatter but no name field
- keep warning for genuinely invalid agent frontmatter, and include the file path in that warning
- add focused tests covering both cases

## Validation
- static diff review only in this environment
- could not run cargo test or cargo fmt here because the Rust toolchain is not installed

### DIFF
--- a/crates/goose/src/agents/platform_extensions/summon.rs
+++ b/crates/goose/src/agents/platform_extensions/summon.rs
@@ -150,22 +150,52 @@ struct AgentMetadata {
 }
 
 fn parse_frontmatter<T: for<'de> Deserialize<'de>>(content: &str) -> Option<(T, String)> {
+    match try_parse_frontmatter(content) {
+        Ok(parsed) => parsed,
+        Err(e) => {
+            warn!("Failed to parse frontmatter: {}", e);
+            None
+        }
+    }
+}
+
+fn try_parse_frontmatter<T: for<'de> Deserialize<'de>>(
+    content: &str,
+) -> std::result::Result<Option<(T, String)>, serde_yaml::Error> {
     let parts: Vec<&str> = content.split("---").collect();
     if parts.len() < 3 {
-        return None;
+        return Ok(None);
     }
 
     let yaml_content = parts[1].trim();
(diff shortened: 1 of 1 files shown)


DOsinga:
Thank you for the contribution and for linking it to a real issue (#8273) — the underlying problem is worth fixing. Unfortunately, we can't accept a PR where the code was never compiled or tested by the author. Please see our [contributing guidelines on responsible AI usage](https://github.com/block/goose/blob/main/CONTRIBUTING.md#quick-responsible-ai-tips) for more context. If you're able to set up the Rust toolchain and verify your changes locally, we'd welcome a fresh attempt!

PR #8344, opened 2026-04-06
feat(acp): introduce threads

baxen:


crates/goose-acp/src/server.rs
```diff
@@ -936,57 +958,48 @@ impl GooseAcpAgent {
...
+                args.cwd.clone(),
+                requested_provider.as_deref(),
+                None,
+            )
+            .await?;
 
-        let session_id = SessionId::new(goose_session.id.clone());
+        let internal_session_id = goose_session.id.clone();
+        // Use the internal session ID for agent creation (it keys into the sessions DB).
+        let agent_session_id = SessionId::new(internal_session_id.clone());
```
alexhancock: i might name this something like `wrapped_internal_session_id`

alexhancock (approved):
The implementation looks solid and disambiguates two concepts which are conflated in the current code, so that's good. One idea I had since we discussed the other day would be to not introduce a new noun (Thread), but rather just call them "ACPSession"

So `thread_id` -> `acp_session_id` etc

goose can probably make the change in a clean commit and I think the code would end up easier to understand.

PR #8353, opened 2026-04-07
fix(cli): show actual provider error in configuration failure message

### LINKED ISSUE #8352: Provider config failure shows generic error instead of actual error message
**Describe the bug**

When `goose configure` fails during the provider configuration check, the error dialog shows a hardcoded generic message regardless of the actual failure reason:

```
Failed to configure provider: init chat completion request with tool did not succeed.
```

The real error is briefly shown in the spinner above, but immediately overwritten by this generic message. Whether it's a 404 (model not found), an auth error, or a model that doesn't support tool calling — the user always sees the same unhelpful message.

---

**To Reproduce**
1. Run `goose configure`
2. Select a provider (e.g., OpenAI)
3. Choose "Enter a model not listed..." and type a model name that doesn't exist
4. Observe the error output — the real error appears in the spinner but the final message is generic

---

**Expected behavior**

The final error message should include the actual error from the provider, so users can diagnose the issue (wrong model name, bad API key, unsupported feature, etc.)

---

**Screenshots**

Image

(sorry for the terminal theme readability lol)

---

- **OS & Arch:** Windows 11 x86_64
- **Interface:** CLI
- **Version:** v1.29.1
- **Extensions enabled:** developer
- **Provider & Model:** OpenAI – (various models that fail configuration)

---

**Additional context**

I hit this while trying different setups — a local model via sglang that didn't support tool calling, and a model name that didn't exist on Groq. Both showed the exact same generic error, making it impossible to tell what actually went wrong without checking logs.

SushaanthSrinivasan:
## Summary
- Replace hardcoded generic error message in `goose configure` with the actual provider error
- Users can now diagnose configuration failures (wrong model name, auth issues, unsupported features) without digging into logs

Fixes #8352

## Screenshot

Top: before (generic message), Bottom: after (actual error shown)

Before and after comparison

(sorry for the terminal theme readability lol)

## Test plan
- [x] Run `goose configure` with a non-existent model name — shows the actual 404 error instead of generic message
- [x] Run `goose configure` with a valid model — succeeds as before (no change to happy path)
- [x] `cargo check -p goose-cli` passes

### DIFF
--- a/crates/goose-cli/src/commands/configure.rs
+++ b/crates/goose-cli/src/commands/configure.rs
@@ -1,12 +1,12 @@
 use crate::recipes::github_recipe::GOOSE_RECIPE_GITHUB_REPO_CONFIG_KEY;
 use cliclack::spinner;
 use console::style;
-use goose::agents::Agent;
-use goose::agents::extension::{PLATFORM_EXTENSIONS, ToolInfo};
+use goose::agents::extension::{ToolInfo, PLATFORM_EXTENSIONS};
 use goose::agents::extension_manager::get_parameter_names;
-use goose::agents::{ExtensionConfig, extension::Envs};
+use goose::agents::Agent;
+use goose::agents::{extension::Envs, ExtensionConfig};
 use goose::config::declarative_providers::{
-    CreateCustomProviderParams, create_custom_provider, remove_custom_provider,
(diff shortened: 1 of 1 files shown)


SushaanthSrinivasan:
thanks for the merge! will make sure to include the DCO signoff from next time

PR #8378, opened 2026-04-07
mergeable configs + cleanup

jamadeo:
This PR introduces mergeable config files, so goose will pull configuration from multiple sources: /etc/goose/config.yaml, and ~/.config/goose/config.yaml. You can have an extension defined in the global config and enable/disable it in your user config.

Also, remove the backup/restore logic.


jamadeo:
@codex review

DOsinga (approved):

PR #8400, opened 2026-04-08
fix(desktop): accept self-signed certs from configured external goosed host

Gandalf-Le-Dev:
## Summary

  - The `certificate-error` and `setCertificateVerifyProc` handlers in `main.ts` only accepted self-signed certificates from `localhost`, rejecting all other hosts
  - This makes remote connection to an external goosed server impossible over HTTPS, since goosed generates a self-signed cert
  - Combined with the CSP blocking plain HTTP to non-localhost hosts, this creates a catch-22 where no protocol works for remote backends
  - Add `isTrustedHost()` that also trusts the user-configured `externalGoosed` host, so remote backends work with self-signed certs

  ## Test plan

  - [ ] Configure an external goosed server on a remote host with `GOOSE_SERVER__SECRET_KEY`
  - [ ] In Goose Desktop, set remote connection URL to `https://<remote-host>:<port>` with the secret key
  - [ ] Verify connection succeeds (previously failed with "Could not connect to external backend")
  - [ ] Verify localhost connections still work as before
  - [ ] Verify untrusted hosts are still rejected

fixes #8376

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4604,6 +4604,7 @@ dependencies = [
  "hex",
  "http 1.4.0",
  "openssl",
+ "pem",
  "rand 0.8.5",
  "rcgen",
  "reqwest 0.13.2",
--- a/crates/goose-server/Cargo.toml
+++ b/crates/goose-server/Cargo.toml
@@ -76,6 +76,7 @@ rcgen = "0.13"
 axum-server = { version = "0.8.0" }
 aws-lc-rs = { version = "1.16.0", optional = true }
 openssl = { version = "0.10", optional = true }
+pem = "3.0.6"
 
 [target.'cfg(windows)'.dependencies]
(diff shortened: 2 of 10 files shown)


jh-block:
I made some changes to tighten this up a bit.

With the local backend we pin the fingerprint to what goosed outputs when it starts up.

With remote backends we can't automatically know the fingerprint of the cert that was generated before connecting, but we now allow the user to specify the fingerprint if known (and I added support on the goosed side for providing a cert + key, so the fingerprint can be known), and if not specified, we do TOFU (pin the fingerprint upon first connect) as a basic protection against MITM during the session.

lifeizhou-ap (approved):

Gandalf-Le-Dev:
nice, thanks for the improvement, do you know when it will release ?

PR #8411, opened 2026-04-08
fix: allow manual model entry when provider model listing fails

### LINKED ISSUE #7636: Switch Models dialog blocks provider selection when model listing API fails (e.g., Anthropic via Azure AI Foundry)

Yagi-Michael:
Closes https://github.com/block/goose/issues/7636

## Summary
- When a provider's model listing API fails (e.g., `/v1/models` returns 404 through a proxy like Azure AI Foundry), the Switch Models dialog showed a dead-end red error banner with no way to proceed
- Replace the blocking error with a yellow warning banner + text input so users can still type a model name manually
- **Pure UI escape hatch — no backend changes, no health check bypass.** The `/v1/models` health check still runs and the error is still surfaced to the user as a warning. This does not restore the `known_models` fallback or touch `modelInterface.ts` at all.

### Note on prior PR #7649
That PR was closed because it restored the `known_models` fallback for all providers in `modelInterface.ts`, which subverted the health check introduced in PR #6744. This PR takes a different approach — it only changes the UI layer (`SwitchModelModal.tsx`) to add a manual input alongside the existing warning, so the health check behavior is fully preserved.

## Test plan
- [x] Set `ANTHROPIC_HOST` to a bad endpoint (e.g., `http://localhost:9999`) in `~/.config/goose/config.yaml`
- [x] Open Switch Models → select Anthropic
- [x] **Before**: red dead-end error, no model input, user is stuck
- [x] **After**: yellow warning banner with text input to manually enter a model name
- [x] Verify normal providers (with working model listing) still show the dropdown as usual
- [x] Verify local provider still shows the "download models" UI when no models are downloaded

### DIFF
--- a/ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
+++ b/ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
@@ -115,14 +115,6 @@ const i18n = defineMessages({
     id: 'switchModelModal.checkProviderConfig',
     defaultMessage: 'Check your provider configuration in Settings → Providers',
   },
-  couldNotFetchModels: {
-    id: 'switchModelModal.couldNotFetchModels',
-    defaultMessage: 'Could not fetch models from provider',
-  },
-  manualEntryHint: {
-    id: 'switchModelModal.manualEntryHint',
-    defaultMessage: 'You can still enter a model name manually, or check your provider configuration in Settings → Providers.',
-  },
   loadingModels: {
     id: 'switchModelModal.loadingModels',
     defaultMessage: 'Loading models…',
@@ -855,13 +847,13 @@ export const SwitchModelModal = ({
                         <div className="flex items-start">
                           <div className="flex-1">
                             <h3 className="text-sm font-medium text-yellow-800 dark:text-yellow-200">
-                              {intl.formatMessage(i18n.couldNotFetchModels)}
+                              {intl.formatMessage(i18n.couldNotContactProvider)}
                             </h3>
                             <div className="mt-1 text-sm text-yellow-700 dark:text-yellow-300">
                               {providerErrors[provider]}
                             </div>
                             <div className="mt-2 text-xs text-yellow-600 dark:text-yellow-400">
-                              {intl.formatMessage(i18n.manualEntryHint)}
+                              {intl.formatMessage(i18n.checkProviderConfig)}
                             </div>
                           </div>
                         </div>
--- a/ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
+++ b/ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx
@@ -115,6 +115,14 @@ const i18n = defineMessages({
     id: 'switchModelModal.checkProviderConfig',
     defaultMessage: 'Check your provider configuration in Settings → Providers',
   },
+  couldNotFetchModels: {
+    id: 'switchModelModal.couldNotFetchModels',
+    defaultMessage: 'Could not fetch models from provider',
+  },
+  manualEntryHint: {
+    id: 'switchModelModal.manualEntryHint',
+    defaultMessage: 'You can still enter a model name manually, or check your provider configuration in Settings → Providers.',
+  },
   loadingModels: {
     id: 'switchModelModal.loadingModels',
     defaultMessage: 'Loading models…',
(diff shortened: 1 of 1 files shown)


Yagi-Michael:
Before: image
After: 
image

Yagi-Michael:
Ah sorry, just saw using `i18n`. Will fix and reuse the strings already made so try not to add new loc

jamadeo (approved):
I don't know that this would often lead to a valid state, but getting un stuck is probably worth allowing this

Yagi-Michael:
> I don't know that this would often lead to a valid state, but getting un stuck is probably worth allowing this

Yeah, I don't think it does, is mostly just to unblock the UI and was an easy change without causing too much chaos from rogue open source contributor like myself, I know it must be a dread going through all the PRs so tried to keep it lean and at least "unblock" the user. Thank you for merging :goose:

PR #8462, opened 2026-04-10
feat(providers): add llama-swap declarative provider

caniko:
## Summary
Adds a new fixed declarative provider entry for [llama-swap](https://github.com/mostlygeek/llama-swap) at `crates/goose/src/providers/declarative/llama_swap.json`. No Rust code changes are required: `register_declarative_providers` picks it up automatically via `include_dir!` and dispatches it through the existing `OpenAiProvider::from_custom_config` path.

Config choices:
* `engine: "openai"` because llama-swap exposes `/v1/chat/completions` and `/v1/models` with full OpenAI compatibility.
* `LLAMA_SWAP_HOST` env var with a default of `http://localhost:8080` (matches llama-swap's documented default) so users can point at a non-default host/port through the normal settings UI.
* `dynamic_models: true` so the user's YAML-configured model list is fetched from `/v1/models` at runtime rather than hardcoded.
* `requires_auth: false` and `skip_canonical_filtering: true` to match the `lmstudio` pattern, since llama-swap runs unauthenticated by default and model names are user-defined.

### Testing
* Added `test_llama_swap_json_deserializes` in `crates/goose/src/config/declarative_providers.rs` alongside the existing `test_tanzu_json_deserializes`, asserting all shipped defaults (name, engine, base_url template, env var default, `dynamic_models`, `skip_canonical_filtering`, etc.).
* `cargo test -p goose --lib config::declarative_providers` passes locally (9/9).
* Manual end-to-end smoke test against a local llama-swap on `:8080` is planned before merge: verify the provider shows up in the picker with `LLAMA_SWAP_HOST` pre-filled, `/v1/models` populates dynamically, and both streaming and non-streaming completions work. A second pass with a non-default `LLAMA_SWAP_HOST` will confirm the env var override path.

### Related Issues
N/A

N/A (no UI code changes; the new provider appears in the existing provider list using its declarative metadata).

### DIFF
--- a/crates/goose/src/config/declarative_providers.rs
+++ b/crates/goose/src/config/declarative_providers.rs
@@ -531,6 +531,34 @@ mod tests {
     }
 
     #[test]
+    fn test_llama_swap_json_deserializes() {
+        let json = include_str!("../providers/declarative/llama_swap.json");
+        let config: DeclarativeProviderConfig =
+            serde_json::from_str(json).expect("llama_swap.json should parse");
+        assert_eq!(config.name, "llama_swap");
+        assert_eq!(config.display_name, "Llama Swap");
+        assert!(matches!(config.engine, ProviderEngine::OpenAI));
+        assert_eq!(config.api_key_env, "");
+        assert!(!config.requires_auth);
+        assert!(config.skip_canonical_filtering);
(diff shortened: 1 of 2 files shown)


jamadeo (approved):

jamadeo:
Thank you!

Issue #7150, opened 2026-02-11
Kimi Code Support


KevinBerthier:
**Please explain the motivation behind the feature request.**
I added Kimi code (that have an Anthropic compatible API) but it's not working

**Describe the solution you'd like**
Support Kimi Code subscription

**Describe alternatives you've considered**
Tried Anthropic compatible as described here https://www.kimi.com/code/docs/en/more/third-party-agents.html#claude-code
and OpenAI compatible as described here https://www.kimi.com/code/docs/en/more/third-party-agents.html#configure-the-kimi-code-model-1

**Additional context**

- [x] I have verified this does not duplicate an existing feature request

stebbins:
dupe of https://github.com/block/goose/issues/6902 ? or is kimi code subscription different?

KevinBerthier:
@stebbins it looks different 
Kimi code docs https://www.kimi.com/code/docs/en/more/third-party-agents.html (base URL is api.kimi.com)
Kimi API docs https://platform.moonshot.ai/docs/guide/agent-support.en-US (base URL is api.moonshot.ai)

DOsinga:
can you share a diagnostics report? I am guessing it is probably a path thing (goose requires the full path not /v1 or so)

KevinBerthier:
[diagnostics_20260212_1.zip](https://github.com/user-attachments/files/25283953/diagnostics_20260212_1.zip)

stebbins:
@DOsinga I don't have Kimi Code subscription so I can't verify myself, but presumably, there would _also_ be the issue of how `reasoning_content` is handled. That's what I ran into when setting up access via Moonshot API, at least.

I have a working PR for the direct Moonshot API integration: https://github.com/block/goose/pull/6962

I've also attempted directly resolving this issue: https://github.com/block/goose/pull/7222

@KevinBerthier  Mind taking a look and testing?

spazbg:
Not sure how to build locally and test the fixes, but can't wait to see native Kimi K2.5 support in the official Goose app :) Thank you for the hard work, guys

jh-block:
Implemented in #8466, thanks @soilSpoon!

PR #8466, opened 2026-04-10
feat: add Kimi Code provider with OAuth device flow authentication

soilSpoon:
## Summary

Add a Kimi Code provider with OAuth 2.0 Device Authorization Grant (RFC 8628) authentication.

The existing `kimi.json` declarative provider was broken — it pointed to `api.moonshot.cn` (Moonshot's direct API) with a `MOONSHOT_API_KEY`, rather than the Kimi Code platform (`api.kimi.com/coding`) which uses OAuth. This PR replaces it with a hand-coded provider that works correctly.

**How authentication works:**
1. First use: `goose configure` → select "Kimi Code" → browser opens + user code copied to clipboard → user approves on `auth.kimi.com` → token saved to `~/.config/goose/kimicode/token.json` (mode `0o600`)
2. Subsequent requests: access token reused from memory/disk cache, auto-refreshed 5 min before expiry via `refresh_token`
3. Token fully expired: device flow restarts automatically

Follows the same patterns as `GithubCopilotProvider` (device code flow) and `AnthropicProvider` (Anthropic-compatible SSE streaming).

Key implementation details:
- OAuth host: `https://auth.kimi.com`, Client ID: `17e5f671-d194-4dfb-9706-5516cb48c098` (public client ID, same as [kimi-cli](https://github.com/MoonshotAI/kimi-cli) and [oh-my-pi](https://github.com/can1357/oh-my-pi))
- RFC 8628 `slow_down` error handled — polling interval increases by 5 s per server request
- Supports `kimi-k2.5` and `kimi-k2-thinking` models
- Reuses the provider's `reqwest::Client` for API calls (no per-request client construction)

### Testing

Manual testing plan:
- [ ] `goose configure` → select "Kimi Code" → complete OAuth device flow in browser
- [ ] Verify `~/.config/goose/kimicode/token.json` exists with `0o600` permissions
- [ ] Run a session with `kimi-k2.5` and confirm streaming responses work
- [ ] Verify token refresh: manually set `expires_at` to a past time in the token file and confirm auto-refresh triggers

### Related Issues

Relates to #7150
Discussion: (none)


jh-block:
Re-assigned to myself since the associated issue is assigned to me

crates/goose/src/providers/kimicode.rs
```diff
@@ -0,0 +1,640 @@
...
+                .post(format!("{}/api/oauth/token", KIMI_AUTH_HOST))
+                .headers(self.kimi_headers())
+                .form(&PollReq {
+                    client_id: KIMI_CODE_CLIENT_ID,
+                    device_code,
+                    grant_type: "urn:ietf:params:oauth:grant-type:device_code",
+                })
+                .send()
+                .await
+                .context("failed to poll for token")?
```
jh-block: need a raise_for_status() here to handle possible HTTP errors

crates/goose/src/providers/kimicode.rs
```diff
@@ -0,0 +1,640 @@
...
+            "setup_steps should be populated"
+        );
+    }
+
+    #[test]
+    fn fetch_supported_models_returns_known_models() {
+        let known: Vec<String> = KIMI_CODE_KNOWN_MODELS
+            .iter()
+            .map(|s| s.to_string())
+            .collect();
```
jh-block: this test isn't actually testing anything; it's just testing that a constant contains the values that it contains

jh-block (commented):
just a couple of minor issues. there's also quite a bit of duplication between `get_access_token` and `configure_oauth`. maybe `configure_oauth` can call `get_access_token`, or a shared helper can be extracted?

eventually we might want to extract a common oauth device flow helper that multiple providers can use, but no need to do that as part of this PR.

crates/goose/src/providers/kimicode.rs
soilSpoon: Addressed in commit a5113f5d1, but not as a literal `error_for_status()` — the OAuth device flow (RFC 8628) returns `authorization_pending` and `slow_down` as 4xx responses with a JSON body, so raising on any non-2xx would break polling entirely (that was the original Codex P1 on the first commit).

The new code in `poll_for_token` reads the body first, attempts to parse it as `PollResp`, and only surfaces the HTTP status when parsing fails **and** the status is non-success. That preserves RFC 8628 semantics while still giving full status + body context on real transport errors.

Covered by:
- `poll_for_token_handles_authorization_pending_then_success` (400 + pending body)
- `poll_for_token_surfaces_http_error_on_unparseable_body` (502 + unparseable body)

crates/goose/src/providers/kimicode.rs
soilSpoon: Agreed — removed in commit a5113f5d1 and replaced with wiremock-backed tests that exercise real refresh/poll behavior:

- `use_or_refresh_returns_fresh_token_without_calling_endpoint`
- `use_or_refresh_falls_back_to_existing_token_when_refresh_fails`
- `use_or_refresh_returns_new_token_on_successful_refresh`
- `use_or_refresh_preserves_refresh_token_when_server_omits_it`
- `poll_for_token_handles_authorization_pending_then_success`
- `poll_for_token_accepts_response_without_refresh_token`
- `poll_for_token_surfaces_http_error_on_unparseable_body`
- `kimi_headers_skips_invalid_device_id_without_panic`
- `validates_device_id_rejects_invalid_bytes`

crates/goose/src/providers/kimicode.rs
```diff
@@ -0,0 +1,828 @@
...
+                    }
+                    return Err(
+                        anyhow::Error::new(e).context("failed to parse token poll response")
+                    );
+                }
+            };
+
+            if let (Some(access_token), Some(refresh_token)) =
+                (resp.access_token, resp.refresh_token)
+            {
```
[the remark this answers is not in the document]
soilSpoon: Fixed in commit 3b04aa803. `poll_for_token` now succeeds on just `access_token` and defaults `refresh_token` to empty when the server omits it, per RFC 6749. Covered by new test `poll_for_token_accepts_response_without_refresh_token`.

crates/goose/src/providers/kimicode.rs
```diff
@@ -0,0 +1,828 @@
...
+        struct RefreshReq<'a> {
+            client_id: &'a str,
+            grant_type: &'static str,
+            refresh_token: &'a str,
+        }
+        #[derive(Deserialize)]
+        struct RefreshResp {
+            access_token: String,
+            refresh_token: String,
+            expires_in: Option<i64>,
```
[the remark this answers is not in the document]
soilSpoon: Fixed in commit 3b04aa803. `RefreshResp.refresh_token` is now `Option<String>`, and `do_refresh_token` falls back to reusing the prior refresh_token when the server omits one, per RFC 6749 §6. Covered by new test `use_or_refresh_preserves_refresh_token_when_server_omits_it`.

soilSpoon:
@jh-block Thanks for the review. Quick status update — all inline feedback is addressed:

- **Duplication between `get_access_token` and `configure_oauth`** — extracted `ensure_token` (the cache → refresh → device-flow ladder) and `use_or_refresh` (per-token freshness/refresh/fallback) in commit a5113f5d1. Both entry points now go through `ensure_token`, so the duplicated branching is gone.
- **`raise_for_status()` on the poll request** — replied inline. Short version: a plain `error_for_status()` would break RFC 8628, so I read the body first and only surface the HTTP status when parsing fails AND status is non-2xx. Covered by `poll_for_token_surfaces_http_error_on_unparseable_body` and `poll_for_token_handles_authorization_pending_then_success`.
- **Trivial constants-only test** — removed and replaced with 9 wiremock-backed tests.

Also addressed during this round:

- Device-code panic on corrupted `device_id` file → validated at load and regenerated.
- Missing `kimi_code_configured` marker → persisted on successful `configure_oauth` so the server's `check_provider_configured` reports the provider as ready.
- Refresh-failure fallback → reuses the still-unexpired token instead of forcing device flow.
- Hardcoded attempt cap → replaced with a deadline derived from `expires_in`.
- `poll_for_token` no longer requires `refresh_token` in the response, and `do_refresh_token` reuses the prior refresh_token when the server omits one (both per RFC 6749; 3b04aa803).
- Named constants for fallback timeouts, warn on corrupted `token.json`, `eprintln!` instead of `println!`, and outcome-focused `tracing` on refresh/fallback/device-flow events.

**Follow-ups intentionally out of scope (happy to split into separate PRs):**

- Shared RFC 8628 device-flow helper for `kimicode` + `githubcopilot` — matches your suggestion. Only these two currently use device flow (`chatgpt_codex` and `gemini_oauth` use authorization-code variants).
- Typed request struct for the `stream` field (currently `payload.as_object_mut().unwrap().insert(...)`) — touches shared `formats::anthropic::create_request`.
- Optional process-level single-flight protection against concurrent device-flow logins (file lock on `token.json`).

All tests pass, clippy clean. Ready for another look.

jh-block (approved):

Issue #8479, opened 2026-04-11
Ollama provider: context window shows 0% / token counts are always null


kasjens:
## Describe the bug

When using Goose with the Ollama provider, the context window indicator permanently shows `0%` (e.g., `0% 0/128k`) and token counts are never tracked, even though Ollama's API returns valid token usage data in its responses.

## To Reproduce

1. Configure Goose with Ollama provider in `~/.config/goose/config.yaml` (set `GOOSE_PROVIDER: ollama`, `GOOSE_MODEL: deepseek-v3.2:cloud`, `OLLAMA_HOST: localhost:11434`)
2. Start a new session: `goose session`
3. Send any message
4. Observe the context window bar — it stays at 0%

## Evidence: Ollama does return token counts

A direct API call to the same model/endpoint returns valid usage data:

```bash
curl -s http://localhost:11434/api/chat \
  -d '{"model":"deepseek-v3.2:cloud","messages":[{"role":"user","content":"hi"}],"stream":false}'
```

Response includes:

```json
{
  "prompt_eval_count": 5,
  "eval_count": 128,
  "done": true
}
```

However, inspecting the Goose sessions database shows all token fields are `null`:

```sql
SELECT total_tokens, input_tokens, output_tokens FROM sessions ORDER BY updated_at DESC LIMIT 3;
-- Returns: (None, None, None) for all Ollama sessions

SELECT tokens FROM messages WHERE session_id='...' LIMIT 5;
-- Returns: (None,) for all messages
```

## Root cause hypothesis

Goose's Ollama provider does not map Ollama's `prompt_eval_count` and `eval_count` response fields to its internal token tracking (`input_tokens`, `output_tokens`, `total_tokens`). Ollama uses different field names than OpenAI-compatible APIs (`prompt_eval_count`/`eval_count` vs `prompt_tokens`/`completion_tokens`).

## Expected behavior

The context window should reflect actual token usage based on the `prompt_eval_count` and `eval_count` values returned by Ollama's API.

## Environment

- **Goose version:** 1.30.0
- **OS:** Ubuntu Linux (6.17.0-20-generic)
- **Ollama:** latest (port 11434)
- **Models tested:** `deepseek-v3.2:cloud`, `qwen3.5:9b` (local)
- **Interface:** CLI

## Related issues

- #6588 — Similar 0% token tracking bug for Google/OpenRouter providers (closed/fixed, but fix did not cover Ollama)
- #1703 — Token usage tracking feature request

[...]

Issue #8476, opened 2026-04-11
There appears to be a regression in Goose when using an Ollama backend.


AiRC-ai:
**Describe the bug**

There appears to be a regression in Goose when using an **Ollama backend**.

This issue does **not** occur in **v1.27.2**, but it **does** occur in **v1.30.0**.

In v1.30.0, requests through the Ollama backend frequently fail with:

`Request failed: Stream decode error: Ollama stream stalled: no data received for 30s. This may indicate the model is overwhelmed by the request payload. Try a smaller model or reduce the number of tools.`

The same Ollama workflow worked in v1.27.2, so this appears to be a behavioral change or regression introduced after that release.

---

**To Reproduce**
1. Install and run **Goose v1.30.0**
2. Configure Goose to use an **Ollama backend**
3. Start a chat using a model served by Ollama
4. Send a prompt and wait for the streamed response
5. Observe the request fail with the 30-second stalled stream error

For comparison:
1. Run the same general workflow in **Goose v1.27.2**
2. Use the same Ollama backend/setup
3. Observe that the issue does not occur there

---

**Expected behavior**

Requests using the Ollama backend should behave at least as reliably as they did in v1.27.2. Goose should not prematurely fail with a stream stall error when the same backend/setup previously worked in the older version.

---

**Screenshots**

Attached screenshot shows the exact error message in the UI.

---

**Please provide the following information**
- **OS & Arch:** macOS host environment; also using Kali WSL in related workflow
- **Interface:** UI
- **Version:** v1.30.0 affected; v1.27.2 works
- **Extensions enabled:** Standard enabled set in current Goose UI session
- **Provider & Model:** Ollama backend; reproduced across models

---

**Additional context**

Important detail: this looks like a **version regression**.

- **Working:** v1.27.2
- **Failing:** v1.30.0

This suggests the issue may be related to a change in:
- streaming response handling
- timeout handling
- Ollama provider integration
- tool/context payload handling
- initial token wait behavior

Observed error:

`Request failed: Stream decode error: Ollama stream stalled: no data received for 30s. This may indicate the model is overwhelmed by the request payload. Try a smaller model or reduce the number of tools.`

The message suggests payload/tool load, but since the same backend flow worked in v1.27.2, this seems more likely to be a regression in Goose’s handling rather than just model size alone.

Potential areas to review:
- any Ollama integration changes between v1.27.2 and v1.30.0
- timeout defaults for streamed responses
- handling of slow first-token latency
[...]

PR #8493, opened 2026-04-12
fix: enable token usage tracking and configurable stream timeout for Ollama provider

kasjens:
## Summary

Three related fixes for the Ollama provider:

1. **Token usage tracking**: The provider was unconditionally stripping `stream_options: {"include_usage": true}` from requests (#7723), preventing Ollama from returning token counts in streaming responses. This is now gated behind `OLLAMA_STREAM_USAGE` (default: `true`) so modern Ollama builds get usage tracking while older builds can opt out with `OLLAMA_STREAM_USAGE=false`. Invalid values are handled safely — a warning is logged and `stream_options` is disabled.

2. **Fallback usage parsing**: Added fallback for Ollama-native token fields (`prompt_eval_count`, `eval_count`) in `get_usage()`, so token tracking works even when Ollama doesn't translate to standard OpenAI field names (`prompt_tokens`, `completion_tokens`). OpenAI fields take precedence when both are present. Null OpenAI fields (e.g. `"completion_tokens": null`) correctly fall through to the Ollama-native fields instead of silently dropping usage.

3. **Configurable stream timeout**: The hardcoded 30s per-chunk timeout was too aggressive for slower models (CPU inference, large parameter counts, complex reasoning). The timeout is now configurable via a resolution chain: `OLLAMA_STREAM_TIMEOUT` > `GOOSE_STREAM_TIMEOUT` > `OLLAMA_TIMEOUT` > default (120s). Zero values are treated as invalid and skipped to prevent immediate stall errors on every chunk.

### Testing

- All 29 related tests pass (zero failures)
- **6 timeout resolution tests**: verify the fallback chain, default behavior, and zero-value rejection
- **3 usage parsing tests**: verify Ollama-native field fallback, OpenAI field precedence, and null OpenAI field fallthrough
- **2 stream_options gating tests**: verify `OLLAMA_STREAM_USAGE` default-on and opt-out behavior

### Related Issues
Relates to #8479
Relates to #8476
Relates to #7723

### DIFF
--- a/crates/goose/src/providers/ollama.rs
+++ b/crates/goose/src/providers/ollama.rs
@@ -353,11 +353,27 @@ impl Provider for OllamaProvider {
     }
 }
 
-/// Per-chunk timeout for Ollama streaming responses.
-/// If no new raw SSE data arrives within this duration, the connection is considered dead.
-/// Set to 120s to accommodate slower models (CPU inference, large parameter counts,
-/// or complex reasoning) that may pause for extended periods between tokens.
-const OLLAMA_CHUNK_TIMEOUT_SECS: u64 = 120;
+/// Default per-chunk timeout for Ollama streaming responses (seconds).
+/// Configurable via OLLAMA_STREAM_TIMEOUT, GOOSE_STREAM_TIMEOUT, or falls back
+/// to OLLAMA_TIMEOUT. Set high to accommodate slower models (CPU inference,
+/// large parameter counts, complex reasoning).
+const OLLAMA_DEFAULT_CHUNK_TIMEOUT_SECS: u64 = 120;
+
+/// Resolve the per-chunk stream timeout from config.
+/// Priority: OLLAMA_STREAM_TIMEOUT > GOOSE_STREAM_TIMEOUT > OLLAMA_TIMEOUT > default (120s).
+fn resolve_ollama_chunk_timeout() -> u64 {
+    let config = crate::config::Config::global();
+
+    if let Ok(val) = config.get_param::<u64>("OLLAMA_STREAM_TIMEOUT") {
+        return val;
+    }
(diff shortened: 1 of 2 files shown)


crates/goose/src/providers/ollama.rs
```diff
@@ -68,10 +68,36 @@ fn resolve_ollama_num_ctx(model_config: &ModelConfig) -> Option<usize> {
     input_limit.or(model_config.context_limit)
 }
 
+fn resolve_ollama_stream_usage() -> bool {
```
jamadeo, line 71: Can we do without a config flag? Maybe https://docs.ollama.com/api-reference/get-version and flag this on a minimum version?

jamadeo (commented):
Nice one, thank you @kasjens . Is it possible we can avoid the config parameter by looking at the ollama version?

michaelneale (approved):
I think this is worth having in, thanks!

crates/goose/src/providers/ollama.rs
kasjens: Thanks @jamadeo! I considered that but went with the config flag because:

1. **Ollama isn't always direct** — users behind proxies or compatible API servers (LiteLLM, LocalAI) may not expose `/api/version`.
2. **Version ≠ capability** — custom builds/forks may support `stream_options` without a recognizable version string.
3. **It defaults to enabled** — so modern installs work out of the box. Only users on older builds need to set `OLLAMA_STREAM_USAGE=false`.

Happy to add version detection as a best-effort first pass with the config flag as fallback if you'd prefer that approach!

jamadeo: Makes sense, and seems like most of the time you'd never add it at all

PR #8494, opened 2026-04-12
fix: use api. subdomain for enterprise GitHub Copilot token URL

### LINKED ISSUE #7716: Github Enterprise Cloud support for Github provider
For enterprises Github provides own instances with custom "my-enterprise.ghe.com" domains. Those should be fully compatible to the standard github copilot. 

For the GitHub Copilot provider the configure dialog should ask for a base url. The default should be github.com, but one should be able to provide an own base like "my-enterprise.ghe.com". The base url should also be saved, and used for the llm calls later.

As an alternative the url could be provided or overridden with an environment variable like GH_HOST. I would prefer the stable configuration in the goose config.

stefri:
The enterprise copilot token URL was incorrectly using a path prefix (/api/copilot_internal/v2/token) which returns a 404. The correct pattern, consistent with how api.github.com works for github.com, is to use the api. subdomain:

  https://api.<host>/copilot_internal/v2/token

Also restores correct use of the computed copilot_token_url variable so that the GITHUB_COPILOT_TOKEN_URL override still works.

Fixes #7716

## Summary
Removed "api" from the url-path and added it as subdomain to the Github Enterprise Host.

### Testing
Tested the implementation against our real our-company.ghe.com, instance within the provided dev-container. I can provide further information which company that is in a private conversation and I'm happy to help with any further testing.

### Related Issues
Relates to #7716

### DIFF
--- a/crates/goose/src/providers/githubcopilot.rs
+++ b/crates/goose/src/providers/githubcopilot.rs
@@ -82,7 +82,7 @@ impl GithubCopilotUrls {
             let base = format!("https://{}", host);
             let copilot_token_url = copilot_token_url
                 .map(|u| u.trim_end_matches('/').to_string())
-                .unwrap_or_else(|| format!("{}/api/copilot_internal/v2/token", base));
+                .unwrap_or_else(|| format!("https://api.{}/copilot_internal/v2/token", host));
             Self {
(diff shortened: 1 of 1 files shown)


lifeizhou-ap (approved):
Thanks a lot for the fix!

PR #8497, opened 2026-04-12
fix: inherit subagent provider and model overrides

AiRC-ai:
## Summary
This change ensures that subagents correctly inherit provider and model configurations when explicit overrides are not provided.

It introduces a consistent normalization and resolution flow that:
- Treats `"inherit"` (including whitespace variations) as **no override**
- Applies a clear precedence order for provider and model resolution:
  1. Delegate parameters
  2. Recipe settings
  3. Environment variables (`GOOSE_SUBAGENT_PROVIDER`, `GOOSE_SUBAGENT_MODEL`)
  4. Parent session defaults (provider only)

## Key Changes
- Added helper functions:
  - `normalize_delegate_override` to sanitize override values
  - `resolve_delegate_provider_name` for provider resolution
  - `resolve_delegate_model_override` for model resolution
- Updated agent parsing to exclude `"inherit"` from metadata-derived labels
- Refactored provider/model resolution logic in `SummonClient` to use centralized helpers
- Ensured `"inherit"` does not propagate into runtime configuration

## Behavior Improvements
- Prevents `"inherit"` from being treated as a literal provider/model
- Ensures predictable fallback behavior
- Aligns subagent configuration with parent session defaults when appropriate

## Testing
Added tests to validate:
- `"inherit"` is ignored in agent frontmatter
- `"inherit"` does not produce recipe settings
- Model override resolution correctly ignores `"inherit"`
- Provider resolution falls back to session when all overrides are `"inherit"`

## Notes
This change improves consistency and reduces edge cases where invalid or unintended configuration values could be applied during subagent execution.

PR #8499, opened 2026-04-12
[codex] fix declarative provider auth updates and Ollama Cloud model selection

AiRC-ai:
## Summary
This stacks the declarative-provider auth update flow with two follow-up fixes needed to make the Ollama Cloud configuration path work end to end.

Changes included:
- allow fixed declarative providers to save auth updates from the provider settings modal
- persist auth overrides for fixed declarative providers instead of treating them like fully editable custom providers
- await provider refresh before opening model selection after an update, so the model picker uses fresh provider state
- skip canonical filtering for `ollama_cloud` so the full live cloud model list is shown instead of an over-filtered subset

## Root cause
There were two separate issues in the provider flow:
- fixed declarative providers could accept an API key in the UI, but the update path did not fully persist and reload those auth overrides in a way the rest of Goose treated as configured
- after a successful update, the model picker could open against stale cached provider state, and Ollama Cloud's live model list was additionally being filtered too aggressively by canonical-model filtering

## Impact
This makes the Ollama Cloud provider flow usable from the desktop settings UI:
- `Update Provider` persists the API key for the fixed declarative provider
- the follow-up model picker opens with refreshed provider state
- the provider shows the live Ollama Cloud models instead of a nearly empty list

## Validation
- `source bin/activate-hermit && cd ui/desktop && pnpm run typecheck`
- rebuilt and packaged the macOS desktop app locally with `just package-ui`
- manually verified the desktop flow:
  - add/update Ollama Cloud API key
  - open model picker after update
  - see populated cloud models and select one successfully

## Notes
The first two commits are the `#8191`-equivalent declarative-provider fixes on top of `main`; the last commit is the follow-up for the refreshed model picker and Ollama Cloud model list behavior.

### DIFF
--- a/crates/goose/src/config/declarative_providers.rs
+++ b/crates/goose/src/config/declarative_providers.rs
@@ -388,6 +388,7 @@ pub fn load_provider(id: &str) -> Result<LoadedProvider> {
 
     Err(anyhow::anyhow!("Provider not found: {}", id))
 }
+
 pub fn load_custom_providers(dir: &Path) -> Result<Vec<DeclarativeProviderConfig>> {
     if !dir.exists() {
         return Ok(Vec::new());
--- a/crates/goose/src/config/declarative_providers.rs
+++ b/crates/goose/src/config/declarative_providers.rs
@@ -302,22 +302,11 @@ pub fn update_custom_provider(params: UpdateCustomProviderParams) -> Result<()> 
             fast_model: existing_config.fast_model.clone(),
         })
     } else {
-        let custom_file = custom_providers_dir().join(format!("{}.json", params.id));
-        if params.requires_auth {
-            Some(DeclarativeProviderConfig {
-                requires_auth: true,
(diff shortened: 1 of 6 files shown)


lifeizhou-ap:
Thanks for the contribution, @AiRC-ai! We noticed you have 3 open PRs (#8497, #8499, #8500) opened around the same time — per our [contribution guidelines](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#pull-requests) we'd suggest focusing on one at a time so we can give it proper review. There are also a couple of unaddressed [Codex review comments](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#ai-code-reviews) on this PR (one flags a real regression where custom providers whose name collides with a fixed provider would have their settings silently dropped). We're going to close this one for now — please feel free to reopen your top-priority PR with the Codex feedback addressed, and we'll take a look!

PR #8502, opened 2026-04-13
fix: use sqlx chrono decode for thread timestamps instead of manual parsing

matt2e:
## Summary

- Thread timestamps (`created_at`, `updated_at`, `archived_at`) were parsed from raw strings using `str::parse()`, which expects RFC 3339 format. SQLite's `CURRENT_TIMESTAMP` produces `YYYY-MM-DD HH:MM:SS` (no `T` separator, no `Z` suffix), so parsing always failed and silently fell back to `Utc::now()` — making every session show as "now" in the desktop app.
- Fix: change `ThreadRow` to decode timestamps as `DateTime<Utc>` directly via sqlx's built-in chrono support, which handles SQLite's timestamp format natively. This removes all manual parsing and fallback logic.

## Known issue

Existing rows written by `save_session` used sqlx's default `DateTime<Utc>` encoding (RFC 3339 with fractional seconds), while `CURRENT_TIMESTAMP` / `datetime('now')` produce SQL format. Both formats are read correctly by sqlx chrono decode, but **SQLite string-based `ORDER BY` on the timestamp columns will sort incorrectly** when the column contains a mix of formats (RFC 3339 timestamps sort after SQL-format ones due to the `T` character). A migration to normalize existing timestamps to a single format would fix this.

## Test plan

- [x] Verify `cargo check` passes
- [x] Confirm thread list in desktop app shows correct relative timestamps instead of "now"
- [x] Verify newly created sessions get correct timestamps
- [x] Check that mixed-format sorting is acceptable for now (old sessions may appear out of order)


lifeizhou-ap:
Hi @matt2e,

Nice bug find! The root cause diagnosis is spot on — CURRENT_TIMESTAMP produces SQL format while .parse() expects RFC 3339.

However, I think the fix can be simpler. The bug is only in thread_manager.rs — session_manager.rs already works correctly because it uses sqlx's built-in chrono Decode, which handles multiple timestamp formats automatically.

We can fix this by changing the timestamp fields in ThreadRow from String to DateTime<Utc>:

https://github.com/aaif-goose/goose/blob/5fa2a8b82159c29335013128c0656a9c7d4aa508/crates/goose/src/session/thread_manager.rs#L57

```
type ThreadRow = (
    String,
    String,
    bool,
    Option<String>,
    DateTime<Utc>,          // was String
    DateTime<Utc>,          // was String
    Option<DateTime<Utc>>,  // was Option<String>
    String,
    Option<String>,
    i64,
);
```

This lets sqlx decode the timestamps directly (same mechanism that Session's FromRow impl already uses), so we can drop parse_timestamp() and the manual parsing in thread_from_row(). No changes to session_manager.rs needed.

matt2e:
Good thinking @lifeizhou-ap! Moved to that solution, and added a known issue note in the PR about how sorting will still be broken due to the multiple formats being written. We can sort that out later

PR #8504, opened 2026-04-13
fix(recipe): read prompted file parameters

vonbai:
## Summary

- Read file contents for prompted recipe parameters declared with `input_type: file`.
- Preserve the existing prompt-building flow for other parameter input types.
- Add regression coverage so a file parameter is inserted as contents instead of a literal path string.

## Validation

- `cargo fmt --check`
- `cargo test -p goose --no-default-features --features rustls-tls recipe::build_recipe::tests::file_parameter_tests::test_build_recipe_user_prompt_file_parameter_reads_file_content`

Note: the focused test run emitted the existing `providers` test dead-code warning for `clear_env`; validation still passed.


lifeizhou-ap (approved):
Thanks @vonbai for your contribution! I pushed a follow-up to consolidate the file-read into one place.

PR #8513, opened 2026-04-13
feat: onboarding UX for the TUI

alexhancock:
## Summary
* Adds an onboarding flow to the new TUI
* Adds an AGENTS.md to help the agent code effectively in the TUI
* Changes it so `npm start` always runs the rust side of goose from local source (compiling if needed)
* Refactors some parts of the TUI code out into separate modules to make the code more scalable to future screens

### Testing
Manual

### Related Issues
N/A

https://github.com/user-attachments/assets/58711b4c-22a3-49d3-bfd8-990fcc955e6c


ui/text/src/constants.tsx
```diff
@@ -0,0 +1,82 @@
...
+  allow_always: "Always allow",
+  reject_once: "Reject once",
+  reject_always: "Always reject",
+};
+
+export const PERMISSION_KEYS: Record<string, string> = {
+  allow_once: "y",
+  allow_always: "a",
+  reject_once: "n",
+  reject_always: "r",
```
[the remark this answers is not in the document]
alexhancock: I am going to remove/redo the code around this anyway in a coming change, so ignoring

ui/text/scripts/dev-binary.mjs
```diff
@@ -0,0 +1,103 @@
...
+    console.log("Cargo.lock changed, needs rebuild");
+    return true;
+  }
+
+  // Check if goose-acp crate sources are newer than the binary
+  const acpDir = join(projectRoot, "crates", "goose-acp");
+  if (existsSync(acpDir)) {
+    const result = spawnSync(
+      "find",
+      [acpDir, "-type", "f", "(", "-name", "*.rs", "-o", "-name", "Cargo.toml", ")", "-newer", binaryPath],
```
jh-block, line 51: I think this might miss changes in the goose-cli/ crate itself?
Also other crates like goose.
Can we not just call cargo unconditionally and let it do nothing if nothing changed?

ui/text/src/onboarding.tsx
```diff
@@ -0,0 +1,674 @@
...
+          <Box marginTop={1} flexDirection="column">
+            <Box>
+              <Text color={CRANBERRY} bold>
+                {"❯ "}
+              </Text>
+              {masked ? (
+                <PasswordInput
+                  key={`password-${currentKey.name}-${inputKey}`}
+                  placeholder={currentKey.name}
+                  onChange={handleChange}
```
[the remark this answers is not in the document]
jh-block: the codex comment seems legit -- the field can get cleared when toggling masked/unmasked

ui/text/src/onboarding.tsx
```diff
@@ -0,0 +1,674 @@
...
+          )}
+      </Box>
+    </Box>
+  );
+});
+
+interface SuccessScreenProps {
+  provider: ProviderDetailEntry | null;
+  height: number;
+  onComplete: () => void;
```
jh-block: this is not called anywhere within this component

ui/text/src/tui.tsx
```diff
@@ -434,85 +220,120 @@ function buildContentLines({
...
-  lines.push(
-    <Box key="u-prompt" width={width} height={1}>
-      <Text color={CRANBERRY} bold>{"❯ "}</Text>
-      {collapsedUserPrompt(turn.userText, width - 4)}
-    </Box>,
-  );
+  const safeWidth = Math.max(width, 20);
+
+  // Render user prompt  
+  const turnId = turn.userText.slice(0, 20).replace(/[^a-zA-Z0-9]/g, '') || 'turn';
```
jh-block: This ends up being used as a React key, and it seems pretty easy for there to be dupes with this method of generating it. Why not use the index?

ui/text/AGENTS.md
```diff
@@ -0,0 +1,446 @@
+# AGENTS.md - Working with Ink CLI Applications
```
jh-block, line 1: This is quite large, will end up in the context of anyone working on the tui, and mostly talks about Ink in general -- I suspect modern models might perform just as well if we only mention the things that are specific to this project rather than giving an Ink tutorial

ui/text/scripts/dev-binary.mjs
alexhancock: I intentionally made it only recompile if it notices changes in `goose-acp` itself. It's not perfect but I think there will be a lot of changes that are irrelevant and force recompiles if we look at all the crates. I figure we can always do it manually when we know it's needed, but save time in other cases.

ui/text/AGENTS.md
alexhancock: Yeah, a good callout! I originally did development that way but found coding agents (goose) were constantly causing rendering issues in the app with code they wrote. I think it's because they are trained on so much react code that runs in browsers, that they offer completions for various things that completely break rendering in Ink apps.

This file is an effort to get them to be able to effectively write UX code atop this library choice. It lead to improved generations in this branch.

michaelneale (approved):
looks ok - one Q was why ModelAndProviderContext.tsx changed for tui change?

jh-block (approved):

alexhancock:
> looks ok - one Q was why ModelAndProviderContext.tsx changed for tui change?

formatting

PR #8544, opened 2026-04-15
fix: links in chat could not be opened

matt2e:
Screenshot 2026-04-15 at 3 51 12 pm

## Summary
- Streamdown showed a dialog asking if the user wanted to open an external link, but that link just tried to navigate the browser to the link, which does nothing in a tauri app
- We have to intercept the links being clicked in messages and show our own dialog, which opens the link in the user's default browser by calling back into the tauri backend

### DIFF
--- a/ui/goose2/src/features/chat/lib/artifactPathPolicyCore.ts
+++ b/ui/goose2/src/features/chat/lib/artifactPathPolicyCore.ts
@@ -2,6 +2,7 @@ import {
   collectCommandArgPathCandidates,
   extractToolNamePathCandidates,
 } from "@/features/chat/lib/artifactPathCommandExtraction";
+import { isExternalHref } from "@/shared/lib/isExternalHref";
 
 export type ArtifactCandidateSource =
   | "arg_key"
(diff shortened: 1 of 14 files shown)


morgmart (approved):

ui/goose2/src/features/chat/lib/artifactPathPolicyCore.ts
```diff
@@ -145,16 +146,7 @@ function hasKnownScheme(value: string): boolean {
...
-  if (!href) return false;
-  const lower = href.trim().toLowerCase();
-  return (
-    lower.startsWith("http://") ||
-    lower.startsWith("https://") ||
-    lower.startsWith("mailto:") ||
-    lower.startsWith("tel:")
-  );
-}
+export { isExternalHref };
```
lifeizhou-ap, line 149: Instead of the import-then-re-export shim, they should just update these two spots to import from @/shared/lib/isExternalHref directly and remove the re-export from artifactPathPolicyCore

ui/goose2/src/features/chat/ui/__tests__/ContextPanel.test.tsx
```diff
@@ -34,6 +34,7 @@ vi.mock("@tauri-apps/api/event", () => ({
 
 vi.mock("@tauri-apps/plugin-opener", () => ({
```
lifeizhou-ap, line 35: we can move this mock to `src/test/setup.ts` and can remove the duplicated mock setup in individual tests

lifeizhou-ap:
Looks good!  

I left some comments.  Before merging this PR, would you please address these including the i18n one from codex.  

Also one thought: would it be possible to keep external markdown links as `a href=...` and intercept the click to open the modal, instead of rendering them as `button`?

something like
 ```
<a
      href={url}
      onClick={(e) => {
        e.preventDefault();
        openModal(url);
      }}
    >
      ...
    </a>
```

That would keep the external-link flow the same, while preserving normal link semantics and accessibility. Since the modal and openUrl() handling are already in place, it seems like a pretty small change from the current implementation.

lifeizhou-ap (approved):

PR #8546, opened 2026-04-15
Update patterns for pattern-based command injection

dorien-koelemeijer:
## Summary
Update patterns for pattern-based command injection detection due to still having a high false positive rate for this type of detection. The command injection classifier false positive rates are very low, but if users aren't connected to WARP, there is an automatic fallback to pattern-based command injection detection. 

### Testing
Manual testing running `just run-ui`

### DIFF
--- a/crates/goose/src/security/patterns.rs
+++ b/crates/goose/src/security/patterns.rs
@@ -47,16 +47,12 @@ impl RiskLevel {
 /// Comprehensive list of dangerous command patterns
 pub const THREAT_PATTERNS: &[ThreatPattern] = &[
     // Critical filesystem destruction patterns
-    ThreatPattern {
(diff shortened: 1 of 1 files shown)


jh-block:
Thanks @dorien-koelemeijer. Can you address the latest feedback from the Codex bot; I think to some extent it just reflects the impossibility of having a watertight protection using regex, but those two do seem fixable

dorien-koelemeijer:
> Thanks @dorien-koelemeijer. Can you address the latest feedback from the Codex bot; I think to some extent it just reflects the impossibility of having a watertight protection using regex, but those two do seem fixable

Hi, sorry for late response - got sidetracked by other things. Will finish this off now

shellz-n-stuff (approved):
Nice one!

PR #8555, opened 2026-04-15
refactor: @aaif/goose-acp -> @aaif/goose-sdk

alexhancock:
Just a rename


jamadeo (approved):

jamadeo:
note to one of us that this will require adding the trusted publisher config on the npmjs.com side of things too

PR #8562, opened 2026-04-15
fix: one more rename

alexhancock:
Was somehow missed


jamadeo (approved):

alexhancock:
this is an obvious fix, so not waiting for CI

PR #8567, opened 2026-04-15
Masking in diagnostics package

michaelneale:
Adds a preventative masking pass to the generated diagnostics zip. All text content (config.yaml, server logs, LLM request logs, session JSON) is now scanned before being written to the zip — any long, high-entropy tokens that look like they shouldn't be shared are replaced with `[REDACTED]`.

Detection uses Shannon entropy combined with character-composition heuristics: long random-looking strings made of alphanumeric chars and hyphens get masked, while normal diagnostic values (URLs, model names, paths, descriptions, timestamps) pass through unchanged.

This only affects the diagnostics zip export — no changes to how config is stored or read at runtime.

### DIFF
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -130,19 +130,10 @@ jobs:
                 echo "- $name@$version"
               fi
             done
-            echo ""
-            if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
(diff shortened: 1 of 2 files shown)


DOsinga:
do we know where this came from? we shouldn't be logging keys etc in the log files in the first place of course

michaelneale:
@DOsinga yeah I have no idea - but I think sometimes people paste in keys into configs (I have seen other agents recommend this TBH)

PR #8580, opened 2026-04-16
failed the script when bundle:default fails and cleanup "alpha" 

lifeizhou-ap:
## Summary
-  Fail release "Build App" when electron-forge make failed.  Fixed in the `bundle: default` task by removing the unnecessary echo message at the end

-  Removed all "alpha" related task and code.  It is not used any more

### Testing
CI build


michaelneale (approved):
nice cleanup

PR #8583, opened 2026-04-16
fix: work around rmcp OAuth discovery bug for remote MCP servers

jh-block:
## Summary

Fixes the OAuth flow never opening a browser for remote MCP servers that require authentication (e.g. Attio). The root cause is an rmcp bug where resource metadata discovery fails fatally when the MCP server returns HTTP 200 with non-JSON content at its base URL, preventing fallback to `.well-known/oauth-authorization-server` discovery.

This PR adds a fallback in `oauth_flow` that catches the discovery failure and retries by fetching `.well-known/oauth-authorization-server` directly, using rmcp's public `AuthorizationManager`/`AuthorizationSession` APIs. It also logs the actual error when `oauth_flow` fails, instead of silently swallowing it.

### Testing

Manual testing with the Attio MCP server (`https://mcp.attio.com/mcp`).

### Related Issues
Relates to #8453
Upstream PR: https://github.com/modelcontextprotocol/rust-sdk/pull/810

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -492,7 +492,15 @@ async fn create_streamable_http_client(
                     .await?,
                 ))
             }
-            Err(_) => Ok(Box::new(client_res?)),
+            Err(e) => {
+                tracing::warn!(
+                    extension = %name,
+                    uri = %uri,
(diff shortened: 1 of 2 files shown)


jh-block:
Superseded by #8618

PR #8587, opened 2026-04-16
Split code signing from build

jamadeo:
Code signing can fail for all sorts of reasons. This lets us retry without rebuilding.


DOsinga (approved):
Looks good — one thing to double-check: Codex flagged that the unsigned intermediate artifacts (which upload the entire output directory including the .zip) could collide with the signed .zip when the release job downloads everything with `merge-multiple: true`. That seems like a real concern worth verifying before merge.

PR #8589, opened 2026-04-16
chore: set goose binaries as executable in package.json

jamadeo:
- sets goose binaries as "bin" targets in package.json
- moves binaries to be dependencies of goose-sdk
- exports resolveGooseBinary from goose-sdk
- removes post installs


ui/sdk/package.json
```diff
@@ -35,6 +35,13 @@
   "peerDependencies": {
     "@agentclientprotocol/sdk": "*"
   },
+  "optionalDependencies": {
```
alexhancock, line 48: do we need to add logic to the SDK package to find these and use the binary? feels like moving that from `ui/text` to `ui/sdk` is productive?

alexhancock (approved):

ui/sdk/package.json
jamadeo: yup, had the same thought a few minutes ago. I'll push some more commits here

alexhancock (approved):

Issue #8592, opened 2026-04-16
Capability negotiation for MCP Apps in Goose2


aharvard:
### Parent issue

- Parent issue: [Implement MCP Apps in Goose2](https://github.com/aaif-goose/goose/issues/8591)

### Summary

Implement the MCP Apps capability handshake between Goose2, Goose ACP, and the downstream MCP server.

This work makes Goose2 explicitly opt into MCP Apps support and makes Goose ACP advertise the MCP Apps extension downstream only for clients that actually support it.

### Sequence diagram

```mermaid
sequenceDiagram
    participant UI as Goose2 UI
    participant Tauri as Tauri ACP Manager
    participant ACP as Goose ACP Server
    participant MCP as MCP Server

    UI->>Tauri: initialize
    Tauri->>ACP: initialize
    ACP->>ACP: Parse Goose2 capability metadata
    ACP->>MCP: initialize
    MCP-->>ACP: initialize result
    ACP-->>Tauri: initialize result
    Tauri-->>UI: ACP transport ready
```

### This issue covers

- Goose2 ACP `initialize`
- `clientCapabilities.meta["com.block.goose2"]`
- Goose ACP parsing of that metadata
- downstream MCP `initialize`
- conditional advertisement of the MCP Apps extension

### Key rules

- Goose2 declares the MCP Apps capability
- Goose ACP translates it
- only then does Goose advertise the MCP Apps extension in its MCP initialize request to the downstream MCP server
- text/TUI clients remain on the non-MCP-App path

### Open question

- `clientCapabilities.meta` is ACP-defined, but `com.block.goose2` is our custom namespaced envelope inside it
- we should decide whether `com.block.goose2` is the right long-term namespace and payload shape, or whether this should be generalized if multiple desktop clients may eventually declare the same MCP Apps capability

### Acceptance criteria

- Goose2 sends `clientCapabilities.meta["com.block.goose2"]`
- Goose ACP detects the MCP Apps capability only from client-declared metadata
- downstream MCP `initialize` includes the MCP Apps extension only for that session/client path
- text/TUI clients remain on the non-MCP-App path

aharvard:
taking another route

PR #8595, opened 2026-04-16
add MCP Apps ACP surface and capability negotiation

aharvard:
This is PR 1 of 3 in a stacked PR sequence for MCP Apps in Goose2.

The full feature delivery depends on all three PRs merging. [PR 2](https://github.com/aaif-goose/goose/pull/8596) stacks on this one, and [PR 3](https://github.com/aaif-goose/goose/pull/8597) stacks on PR 2.

The overall goal of this work is to let Goose2 behave like a real MCP Apps host. This PR handles the backend negotiation layer that everything else depends on. It teaches Goose ACP how to understand Goose2's MCP Apps capability metadata, adds the ACP request surface the desktop app needs, and makes Goose advertise the MCP Apps extension to downstream MCP servers only when the connected client actually supports it.

In other words, this is the groundwork PR. It does not try to render anything in Goose2 yet, but it puts the ACP and MCP pieces in place so the later Goose2 PRs have something solid to build on.

If you're reviewing this slice and want to see how it's supposed to work, start with the sequence diagram and notes in [#8592](https://github.com/aaif-goose/goose/issues/8592).

Related to #8592.


aharvard:
taking another route

Issue #8594, opened 2026-04-16
MCP App View bridge and bootstrap in Goose2


aharvard:
### Parent issue

- Parent issue: [Implement MCP Apps in Goose2](https://github.com/aaif-goose/goose/issues/8591)

### Summary

Implement the renderer and sandbox bridge that let Goose2 host a real MCP App inline once the app-backed render path has already been selected.

This work makes `McpAppView` responsible for rendering, initialization, host context, proxying app requests back through ACP, and handling host interaction requests from the app view.

### Out of scope for this issue

- extra display modes beyond inline. See [Display Modes](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx#display-modes)
- model context updating as a supported end-to-end feature. See [Requests (View → Host)](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx#requests-view--host). This section includes `ui/update-model-context`.

These features are explicitly out of scope for this issue, even if partial plumbing already exists.

### Sequence diagram

```mermaid
sequenceDiagram
    participant UI as Goose2 UI
    participant Host as McpAppView Renderer
    participant Proxy as Sandbox Proxy
    participant View as MCP App View
    participant Tauri as Tauri ACP Manager
    participant ACP as Goose ACP Server
    participant MCP as MCP Server

    UI->>Host: Mounted McpAppView
    Host->>Proxy: Load sandbox proxy iframe
    Proxy-->>Host: iframe onLoad
    Host->>Proxy: ui/notifications/sandbox-resource-ready
    Proxy->>View: Write app HTML into inner iframe
    View->>Proxy: ui/initialize
    Proxy->>Host: Forward ui/initialize
    Host-->>Proxy: ui/initialize result with hostContext
    Proxy-->>View: Forward ui/initialize result
    View->>Proxy: ui/notifications/initialized
    Proxy->>Host: Forward initialized notification
    Host->>Proxy: ui/notifications/tool-input
    Host->>Proxy: ui/notifications/tool-result
    alt App requests host-proxied MCP methods
        View->>Proxy: MCP request
        Proxy->>Host: Forward MCP request
        Host->>Tauri: ACP invoke
        Tauri->>ACP: goose/... request
        ACP->>MCP: MCP request
        MCP-->>ACP: MCP result
        ACP-->>Tauri: goose/... result
        Tauri-->>Host: ACP result
        Host-->>Proxy: MCP result
[...]

PR #8596, opened 2026-04-16
add Goose2 MCP App bridge bootstrap

aharvard:
This is PR 2 of 3 in a stacked PR sequence for MCP Apps in Goose2.

The full feature delivery depends on all three PRs merging. This PR stacks on [PR 1](https://github.com/aaif-goose/goose/pull/8595), and [PR 3](https://github.com/aaif-goose/goose/pull/8597) stacks on top of this one.

With the ACP surface in place, this PR is the big host/bootstrap pass for Goose2 itself. It adds the sandbox proxy, the Tauri command plumbing, and the initial inline app host so Goose2 can actually load an MCP App resource, bootstrap the app view, send host context and tool updates, and proxy app requests back through ACP.

This is the PR where Goose2 starts acting like an app host rather than just a chat UI. The follow-up PR tightens the render decision and makes it `tools/list`-driven, but this one is what gets the bridge and runtime in place in the first place.

If you're reviewing this slice and want to see how it's supposed to work, start with the sequence diagram and notes in [#8594](https://github.com/aaif-goose/goose/issues/8594).

Related to #8594.
Base PR: [#8595](https://github.com/aaif-goose/goose/pull/8595).


aharvard:
taking another route

Issue #8593, opened 2026-04-16
Inline MCP App rendering in Goose2


aharvard:
### Parent issue

- Parent issue: [Implement MCP Apps in Goose2](https://github.com/aaif-goose/goose/issues/8591)

### Summary

Implement the session-bound catalog hydration and render-decision path that lets Goose2 choose between a normal tool card and an inline MCP App.

This work makes `tools/list` metadata the authoritative linkage source for app-backed tools and moves the app-render decision out of raw-output heuristics.

### Sequence diagram

```mermaid
sequenceDiagram
    participant User
    participant UI as Goose2 UI
    participant Tauri as Tauri ACP Manager
    participant ACP as Goose ACP Server
    participant MCP as MCP Server
    participant Host as McpAppView Renderer

    UI->>Tauri: acp_prepare_session
    Tauri->>ACP: session/new
    ACP-->>Tauri: session/new result
    Tauri-->>UI: acp:session_bound
    UI->>Tauri: goose/tools
    Tauri->>ACP: goose/tools
    ACP->>MCP: tools/list
    MCP-->>ACP: tools/list result
    ACP-->>Tauri: goose/tools result
    UI->>UI: Cache tools/list metadata
    User->>UI: Send prompt
    UI->>Tauri: acp_send_message
    Tauri->>ACP: prompt
    ACP->>MCP: tools/call
    MCP-->>ACP: tools/call result
    ACP-->>Tauri: notifications/session/update
    Tauri-->>UI: acp:tool_result with rawOutput
    UI->>UI: Match tool result to cached tools/list metadata
    alt Matched tool metadata includes _meta.ui.resourceUri
        UI->>Host: Render McpAppView
        Host->>Tauri: acpReadResource
        Tauri->>ACP: goose/resource/read
        ACP->>MCP: resources/read
        MCP-->>ACP: resources/read result
        ACP-->>Tauri: goose/resource/read result
        Tauri-->>Host: HTML resource
        Host-->>UI: Mount sandbox iframe inline
    else No MCP app linkage
        UI->>UI: Render standard tool response
    end
```

### This issue covers

- ACP session binding
- `tools/list` hydration after `acp:session_bound`
- per-session tool catalog caching
- preservation of canonical tool identity via `catalogName`
- app-render decision based on cached `tools/list` metadata
- `resources/read` for app-backed tools

### Key rules

- `tools/list` metadata is the authoritative linkage source
[...]

PR #8597, opened 2026-04-16
align Goose2 MCP App rendering with tools/list metadata

aharvard:
This is PR 3 of 3 in a stacked PR sequence for MCP Apps in Goose2.

The full feature delivery depends on all three PRs merging. This PR stacks on [PR 2](https://github.com/aaif-goose/goose/pull/8596), which itself stacks on [PR 1](https://github.com/aaif-goose/goose/pull/8595).

The previous PR gets the inline app host working, but it still makes the render decision in a way that's more heuristic than we'd like. This PR moves that decision onto cached `tools/list` metadata, warms the catalog after session binding, and makes `McpAppView` more renderer-only by passing it a resolved catalog entry instead of asking it to discover app linkage on its own.

This is the cleanup and architectural alignment pass. It brings the code closer to the flow we documented: `tools/list` is the source of truth for app linkage, and Goose2 only takes the inline app path when the matched catalog entry says it should.

If you're reviewing this slice and want to see how it's supposed to work, start with the sequence diagram and notes in [#8593](https://github.com/aaif-goose/goose/issues/8593).

Related to #8593.
Base PR: [#8596](https://github.com/aaif-goose/goose/pull/8596).

### DIFF
--- a/ui/goose2/src/features/chat/hooks/useAcpStream.ts
+++ b/ui/goose2/src/features/chat/hooks/useAcpStream.ts
@@ -29,6 +29,7 @@ import type {
   AcpUsageUpdatePayload,
   AcpReplayCompletePayload,
 } from "./acpStreamTypes";
+import { warmSessionToolsCatalog } from "../ui/mcpAppCatalog";
 
 function getAssistantProviderId(sessionId: string): string | undefined {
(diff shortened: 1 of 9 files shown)


aharvard:
taking another route

PR #8602, opened 2026-04-16
fix(goose2): require a per-process token for the localhost ACP websocket bridge

### LINKED ISSUE #8601: goose2 localhost ACP websocket bridge lacks an explicit client-auth gate
**Describe the bug**

goose2 starts a localhost ACP WebSocket bridge for `goose serve`, but the reviewed transport path does not appear to enforce an explicit client-auth mechanism, and the ACP router still allows any origin.

On current `main`, goose2 constructs the bridge URL as `ws://127.0.0.1:<port>/acp` and launches `goose serve --host 127.0.0.1 --port <random>`. In the ACP transport, `/acp` is exposed directly and the router still uses `CorsLayer::new().allow_origin(Any)`.

I realize localhost-only transport may be an intentional part of the threat model, so I’m framing this as a trust-boundary / hardening issue rather than a remote-RCE claim. The concern is that goose2 depends on a browser-reachable localhost ACP control channel without a visible transport-level auth gate.

---

**To Reproduce**
1. Review the current goose2 ACP bridge path on `main`:
   - `ui/goose2/src-tauri/src/services/acp/goose_serve.rs`
   - `crates/goose-acp/src/transport.rs`
   - `crates/goose-acp/src/transport/http.rs`
   - `crates/goose-acp/src/transport/websocket.rs`
2. Observe that goose2 still constructs `ws://127.0.0.1:<port>/acp` and starts `goose serve` on loopback.
3. Observe that the ACP router still exposes `/acp` and sets `allow_origin(Any)`.
4. Observe that the ACP transport path does not show a visible `X-Secret-Key`, `Authorization`, or explicit `Origin` validation requirement for the WebSocket bridge path.

---

**Expected behavior**
If goose2 relies on a localhost ACP control bridge, I would expect an explicit app-controlled trust boundary for websocket clients (for example a per-process bridge secret/nonce, origin validation, or another transport-level gate), or documented rationale that localhost-only access is the accepted model.

---

**Screenshots**
N/A

---

- **OS & Arch:** macOS arm64
- **Interface:** UI (goose2) / ACP transport
- **Version:** main branch at `93e6f8d521f049efebe08c64f77eb90df0a71ff1` (verified on 2026-04-17 KST)
- **Extensions enabled:** N/A
- **Provider & Model:** N/A

---

**Additional context**
Fresh code-path evidence from the current branch:
- `ui/goose2/src-tauri/src/services/acp/goose_serve.rs:34`
  - bridge URL is `ws://127.0.0.1:{port}/acp`
- `ui/goose2/src-tauri/src/services/acp/goose_serve.rs:75`
  - goose2 starts `goose serve --host 127.0.0.1 --port ...`
- `crates/goose-acp/src/transport.rs:103`
  - router uses `.allow_origin(Any)`
- `crates/goose-acp/src/transport.rs:126`
  - `/acp` route is exposed directly
- `crates/goose-acp/src/transport/http.rs:237`
  - follow-up HTTP requests require `Acp-Session-Id`, but the ACP transport path does not visibly require a secret/auth header

I also checked for an existing upstream duplicate/fix on 2026-04-17 KST and did not find a convincing existing issue for this exact trust-boundary concern.

shaun0927:
## Summary

Hardens the goose2 localhost ACP websocket bridge with a per-process bridge token.

### Problem

On current `main`, goose2 starts `goose serve` on loopback and connects to `ws://127.0.0.1:<port>/acp`. The ACP websocket transport path did not show an explicit client-auth gate, while the ACP router still allowed any origin.

That leaves goose2 depending on a browser-reachable localhost ACP control channel without an app-controlled transport secret.

### Fix

- generate a random bridge token when goose2 spawns `goose serve`
- pass that token to the child process via `GOOSE_ACP_WS_TOKEN`
- append the token to the internal websocket bridge URL returned by `GooseServeProcess::ws_url()`
- reject ACP websocket upgrades when the configured token is missing or does not match
- add focused unit tests for token parsing / matching and the authenticated websocket URL helper

### Scope

This intentionally keeps the change narrow:
- it hardens the goose2 websocket bridge path
- it does **not** redesign the entire ACP auth model
- it does **not** make shared ACP auth mandatory for non-goose2 clients by default

### Testing

- `cargo fmt --all`
- `cargo test -p goose-acp --lib websocket::tests --no-default-features --features rustls-tls -- --nocapture` *(still compiling locally at the time of PR creation; leaving the exact command here for easy reproduction)*

### Related Issues
Fixes #8601

### DIFF
--- a/crates/goose-acp/src/transport.rs
+++ b/crates/goose-acp/src/transport.rs
@@ -4,18 +4,18 @@ pub mod websocket;
 use std::sync::Arc;
 
 use axum::{
+    Router,
     body::Body,
     extract::{
-        ws::{rejection::WebSocketUpgradeRejection, WebSocketUpgrade},
         State,
+        ws::{WebSocketUpgrade, rejection::WebSocketUpgradeRejection},
     },
-    http::{header, Method, Request},
+    http::{Method, Request, header},
     response::Response,
     routing::{delete, get, post},
-    Router,
 };
 use serde_json::Value;
-use tokio::sync::{mpsc, Mutex};
+use tokio::sync::{Mutex, mpsc};
 use tower_http::cors::{Any, CorsLayer};
 
 use crate::server_factory::AcpServer;
@@ -86,7 +86,7 @@ async fn handle_get(
     request: Request<Body>,
 ) -> Response {
     match ws_upgrade {
-        Ok(ws) => websocket::handle_get(state.1, ws).await,
+        Ok(ws) => websocket::handle_get(state.1, ws, request).await,
         Err(_) => http::handle_get(state.0, request).await,
     }
 }
--- a/crates/goose-acp/src/transport/http.rs
+++ b/crates/goose-acp/src/transport/http.rs
@@ -8,7 +8,7 @@ use axum::{
 use http_body_util::BodyExt;
 use serde_json::Value;
 use std::{collections::HashMap, convert::Infallible, sync::Arc, time::Duration};
-use tokio::sync::{mpsc, Mutex, RwLock};
+use tokio::sync::{Mutex, RwLock, mpsc};
 use tokio_util::compat::{TokioAsyncReadCompatExt, TokioAsyncWriteCompatExt};
 use tracing::{error, info};
 
--- a/crates/goose-acp/src/transport/websocket.rs
+++ b/crates/goose-acp/src/transport/websocket.rs
@@ -1,18 +1,21 @@
 use anyhow::Result;
 use axum::{
+    body::Body,
     extract::ws::{Message, WebSocket, WebSocketUpgrade},
-    http::StatusCode,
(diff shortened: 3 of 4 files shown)


DOsinga:
Thanks for the contribution, but closing this one. The issue and PR both read like automated security scanner output with a generated fix — the formulaic structure, the exhaustive file-path evidence, and the 3-minute turnaround from issue to PR all point that way.

The finding itself is marginal: any local process that can connect to localhost can also read the env vars of the goose serve process, so the token doesn't meaningfully raise the security bar. And putting secrets in query strings is itself an anti-pattern.

Please see https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#quick-responsible-ai-tips — we expect authors to review AI-generated work before submitting.

PR #8609, opened 2026-04-16
feat(goose2): voice dictation via direct-ACP pattern

tulsi-builder:
## Summary

Wires voice dictation into goose2 following the post-#8549/#8582 architecture: frontend calls ACP custom methods directly over the goose-serve WebSocket, no Tauri middleware. Supersedes #8565.

### What lands

**Eight ACP custom methods on the goose server**, consumed directly via the regenerated `@aaif/goose-sdk` `GooseClient`:
- `_goose/dictation/transcribe`, `/config`
- `_goose/dictation/models/{list, download, download/progress, cancel, delete}`
- `_goose/dictation/model/select`

**Frontend UI** for local Whisper model management (`ui/goose2/src/features/settings/ui/LocalWhisperModels.tsx`): list + download-with-progress + cancel + delete + select, with 750ms progress polling that auto-stops when nothing is downloading. Replaces the pre-existing "Local model download is not yet available" placeholder.

**Chat input improvements:**
- Mic toggle with auto-submit keyword
- One-click send while mic is still recording (was two-click before)
- `useAudioDevices` now reacts to OS-level mic permission changes via `navigator.permissions.query`
- Defensive filter on empty `deviceId` to avoid Radix Select crashes when `enumerateDevices()` runs before `getUserMedia`

**On the Rust side:**
- `on_dictation_transcribe` now reads per-provider selected model from config (via `dictation_selected_model` helper) instead of always using hardcoded defaults
- `on_dictation_model_select` validates `is_downloaded()` for Local before persisting
- `local-inference` feature forwarded from `goose-cli` to `goose-acp` so `is_configured(Local)` resolves correctly

### Why not Tauri commands
- #8549 moved session/chat ACP off the Rust middleware and onto a direct WebSocket
- #8582 deleted the Tauri ACP manager entirely
- Dictation follows the same pattern: ACP custom methods on the server, consumed directly by frontend

Only OS-bound operations would justify Tauri wrappers; `navigator.mediaDevices.getUserMedia` handles the mic permission prompt in the renderer, so no new Tauri commands were needed.

### Testing

- [x] `cargo check -p goose-acp` (default features) — passes
- [x] `cargo test -p goose-acp -p goose-sdk` — passes
- [x] `pnpm tsc --noEmit` — clean
- [x] `pnpm vitest run` — 51 files / 412 tests passing (9 new tests for SDK-routed dictation functions)
- [x] Manual end-to-end on macOS: local Whisper `tiny` model download → select → transcribe-in-chat flow works
- [ ] OpenAI / Groq / ElevenLabs paths compile and route correctly but were not manually tested (no API keys available — teammates with keys should validate)

### Related Issues

- Builds on #8549 (merged)
- Builds on #8582 (merged)
- Supersedes #8565

### Screenshots/Demos

https://github.com/user-attachments/assets/6b3c5012-2e97-4ac7-b057-0a32c19a6115


crates/goose-cli/Cargo.toml
```diff
@@ -71,7 +71,7 @@ winapi = { workspace = true }
 [features]
 default = ["code-mode", "local-inference", "aws-providers", "telemetry", "otel", "rustls-tls"]
 code-mode = ["goose/code-mode", "goose-acp/code-mode"]
-local-inference = ["goose/local-inference"]
+local-inference = ["goose/local-inference", "goose-acp/local-inference"]
```
alexhancock, line 74: I am a bit lost what the modifications to the `local-inference` feature defs are up to here! Do you know of a purpose? I'd guess you can remove this and the similar line in `crates/goose-acp/Cargo.toml` if not

alexhancock (commented):
It looks pretty good to me! Made one minor comment and @jamadeo may want to have a look as well to confirm the custom type defs fit into the architecture we've planned nicely.

Eventually we'll put all this knowledge into the `AGENTS.md` and it will be automatic but good to review more closely early on

crates/goose-cli/Cargo.toml
tulsi-builder: It's load-bearing. `crates/goose-acp/src/server.rs` has `#[cfg(feature = "local-inference")]` blocks for the Local dictation transcribe path and all six local-model handlers; without the forward, `goose-cli`'s default build activates `goose/local-inference` but not `goose-acp/local-inference`, those blocks are compiled out, and `is_configured(Local)` always returns false — the original bug. Removing either line silently breaks Local Whisper in the default binary.

jamadeo: I think this will make more sense if (when?) we merge the goose-acp crate into the goose crate. Given where we're going with this, I don't think it makes sense that they remain separated

ui/goose2/src/features/chat/hooks/useVoiceInputPreferences.ts
```diff
@@ -0,0 +1,177 @@
...
+  DEFAULT_AUTO_SUBMIT_PHRASES_RAW,
+  VOICE_AUTO_SUBMIT_PHRASES_STORAGE_KEY,
+  VOICE_DICTATION_PREFERRED_MIC_STORAGE_KEY,
+  VOICE_DICTATION_PROVIDER_STORAGE_KEY,
+  normalizeDictationProvider,
+  parseAutoSubmitPhrases,
+} from "../lib/voiceInput";
+import type { DictationProvider } from "@/shared/types/dictation";
+
+const VOICE_INPUT_PREFERENCES_EVENT = "goose:voice-input-preferences";
```
jamadeo, line 14: We should be using values from the goose config, not local storage, for these settings

jamadeo (commented):
Looks overall good, but for the most part we shouldn't use localStorage for settings and instead prefer the goose config. There may be some exceptions to this if the setting is truly relevant only to this app, but I think all the ones here would be better off in the user's config file

ui/goose2/src/features/chat/hooks/useVoiceInputPreferences.ts
tulsi-builder: Done in `a4c78c58f3`. `useVoiceInputPreferences` now reads/writes via `_goose/config/{read,upsert,remove}` over ACP. Three keys: `VOICE_DICTATION_PROVIDER`, `VOICE_DICTATION_PREFERRED_MIC`, `VOICE_AUTO_SUBMIT_PHRASES`.

waiting for checks

ui/goose2/src/features/settings/ui/LocalWhisperModels.tsx
```diff
@@ -0,0 +1,325 @@
...
+    };
+
+    const interval = window.setInterval(() => {
+      void tick();
+    }, POLL_INTERVAL_MS);
+    return () => {
+      cancelled = true;
+      window.clearInterval(interval);
+    };
+  }, [downloadingIds, refresh]);
```
jh-block, line 125: Because this `useEffect` depends on `downloadingIds` and also updates `downloadingIds` unconditionally, it will re-subscribe on every tick, potentially affecting the responsiveness of the app. Make sure `downloadingIds` is only updated if it's different from the previous Set.

ui/goose2/src/features/chat/hooks/useVoiceInputPreferences.ts
```diff
@@ -0,0 +1,211 @@
...
+      setRawAutoSubmitPhrasesState(
+        phrasesResult.value ?? DEFAULT_AUTO_SUBMIT_PHRASES_RAW,
+      );
+    }
+
+    if (!providerResult.ok) {
+      if (micResult.ok) {
+        setPreferredMicrophoneIdState(micResult.value);
+      }
+      return;
```
jh-block, line 84: If this return is hit, since `isHydrated` wasn't set, everything is stuck in the "waiting for hydration" state. Given that it is depending on a config read over the network, a retry might be a good idea here

jh-block (approved):
This looks great, I left a couple of comments but they could also be dealt with as follow-ups if you prefer.

PR #8614, opened 2026-04-17
chore: adding opus 4.7

michaelneale:
Opus 4.7 is out - this teaches goose about it - context size etc - just routine update. 

This description was written by hand, by mic, old school.


jh-block:
There are a bunch of changes to other models here as well, is that intentional?

baxen (approved):
LGTM

PR #8615, opened 2026-04-17
chore: package the goose binary in the goose2 tauri app

jamadeo:
Rather than look in known locations for goose, bundle it in the app.


lifeizhou-ap (approved):

lifeizhou-ap:
@jamadeo  Looks good! 

`just dev` does not work on my local machine in this branch and I pushed an extra commit to make it working.  also cleaned up some scripts which is not required as goose2 is in the goose repo now

PR #8619, opened 2026-04-17
refactor(providers): extract shared OAuth device-flow helper

soilSpoon:
## Summary

- Extract shared RFC 8628 device-code flow into `providers::oauth_device_flow` module
- Migrate `kimicode` and `githubcopilot` to use the shared helper
- Fix RFC 8628 compliance bugs in `githubcopilot`'s polling implementation

## Motivation

Both `kimicode` and `githubcopilot` implemented the OAuth device-code flow independently. The two copies diverged enough to hide bugs in the `githubcopilot` implementation:

- Called `error_for_status()` before parsing the JSON body, so RFC 8628 error payloads like `authorization_pending` (returned with 4xx per §3.5) could surface as plain HTTP errors
- Ignored `slow_down` responses (§3.5: client MUST increase interval by 5s)
- Used a hardcoded 36 × 5s poll window instead of the server's `expires_in`

The new shared helper (`oauth_device_flow.rs`) implements the full RFC 8628 spec once, and both providers configure it via `DeviceFlowConfig`.

## Changes

- **New**: `crates/goose/src/providers/oauth_device_flow.rs` — `run_device_flow`, `poll_for_tokens`, `refresh_device_flow_token`, with 9 wiremock tests
- **kimicode.rs**: `device_flow_login`, `poll_for_token`, `do_refresh_token` replaced with thin wrappers over the helper
- **githubcopilot.rs**: `login`, `get_device_code`, `poll_for_access_token` replaced; RFC 8628 bugs fixed as a side effect
- **Net**: +250 in helper, −370 across both providers (**~315 lines reduced**)

## Test plan

- [ ] 9 new wiremock tests in `oauth_device_flow`: pending → success, slow_down backoff, timeout, HTTP error body, access_denied, missing refresh_token, refresh happy-path
- [ ] 12 existing kimicode tests pass (3 duplicated poll tests removed — covered by helper)
- [ ] Existing githubcopilot tests pass
- [ ] `cargo clippy -p goose --all-targets -- -D warnings` clean

Follow-up to #8466 and #8588.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/crates/goose/src/providers/kimicode.rs
+++ b/crates/goose/src/providers/kimicode.rs
@@ -270,35 +270,31 @@ impl KimiCodeProvider {
         }
     }
 
-    fn device_flow_config(&self) -> (String, String, DeviceFlowConfig<'_>) {
+    async fn device_flow_login(&self) -> Result<KimiToken> {
         let device_auth_url = format!("{}/api/oauth/device_authorization", self.auth_host);
         let token_url = format!("{}/api/oauth/token", self.auth_host);
-        // Borrow via caller-owned strings — see how callers destructure this.
-        (
-            device_auth_url,
-            token_url,
-            DeviceFlowConfig {
-                device_auth_url: "",
-                token_url: "",
-                client_id: KIMI_CODE_CLIENT_ID,
-                scopes: None,
-                extra_headers: self.kimi_headers(),
-                encoding: RequestEncoding::Form,
-            },
-        )
-    }
-
-    async fn device_flow_login(&self) -> Result<KimiToken> {
-        let (device_auth_url, token_url, mut cfg) = self.device_flow_config();
-        cfg.device_auth_url = &device_auth_url;
-        cfg.token_url = &token_url;
+        let cfg = DeviceFlowConfig {
+            device_auth_url: &device_auth_url,
+            token_url: &token_url,
+            client_id: KIMI_CODE_CLIENT_ID,
+            scopes: None,
+            extra_headers: self.kimi_headers(),
+            encoding: RequestEncoding::Form,
+        };
         let tokens = run_device_flow(&self.client, &cfg).await?;
         Ok(tokens_to_kimi(tokens, ""))
     }
 
     async fn do_refresh_token(&self, refresh_token: &str) -> Result<KimiToken> {
-        let (_device_auth_url, token_url, mut cfg) = self.device_flow_config();
-        cfg.token_url = &token_url;
+        let token_url = format!("{}/api/oauth/token", self.auth_host);
+        let cfg = DeviceFlowConfig {
+            device_auth_url: "",
+            token_url: &token_url,
+            client_id: KIMI_CODE_CLIENT_ID,
+            scopes: None,
+            extra_headers: self.kimi_headers(),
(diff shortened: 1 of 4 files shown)


crates/goose/src/providers/oauth_device_flow.rs
```diff
@@ -0,0 +1,553 @@
...
+    let builder = match cfg.encoding {
+        RequestEncoding::Form => builder.form(body),
+        RequestEncoding::Json => builder.json(body),
+    };
+    builder.send().await
+}
+
+fn announce_user_action(device: &DeviceCodeResponse) {
+    if let Ok(mut clipboard) = arboard::Clipboard::new() {
+        let _ = clipboard.set_text(&device.user_code);
```
jh-block: I think we might want to log failure here; silently not copying the code to the clipboard on failure might be a bit confusing

crates/goose/src/providers/kimicode.rs
```diff
@@ -261,201 +271,34 @@ impl KimiCodeProvider {
...
-            .context("failed to refresh token")?
-            .error_for_status()
-            .context("token refresh failed")?
-            .json()
-            .await
-            .context("failed to parse token refresh response")?;
-
+        let token_url = format!("{}/api/oauth/token", self.auth_host);
+        let cfg = DeviceFlowConfig {
+            device_auth_url: "",
```
jh-block: Make this field an Option rather than using empty string to indicate absence

crates/goose/src/providers/kimicode.rs
```diff
@@ -261,201 +271,34 @@ impl KimiCodeProvider {
...
+        let cfg = DeviceFlowConfig {
+            device_auth_url: &device_auth_url,
+            token_url: &token_url,
+            client_id: KIMI_CODE_CLIENT_ID,
+            scopes: None,
+            extra_headers: self.kimi_headers(),
+            encoding: RequestEncoding::Form,
+        };
+        let tokens = run_device_flow(&self.client, &cfg).await?;
+        Ok(tokens_to_kimi(tokens, ""))
```
jh-block: Similarly here, None should be used to indicate no prior refresh, rather than empty string

crates/goose/src/providers/oauth_device_flow.rs
soilSpoon: Done — now logs with `tracing::warn!` on clipboard failure.

crates/goose/src/providers/kimicode.rs
soilSpoon: Done — `DeviceFlowConfig.device_auth_url` is now `Option<&str>`. Callers pass `Some(...)` for device-flow login and `None` for refresh-only paths.

crates/goose/src/providers/kimicode.rs
soilSpoon: Done — `tokens_to_kimi` now takes `Option<&str>` for `prior_refresh`. Fresh login passes `None`, refresh passes `Some(existing_refresh)`.

jh-block (approved):

jh-block:
thank you @soilSpoon!

PR #8623, opened 2026-04-17
[goose2] MCP Apps: translate ACP host capabilities into MCP initialization

aharvard:
**Category:** improvement
**User Impact:** Goose 2 now tells downstream MCP servers which host is rendering MCP Apps UI, and advertises MCP Apps support from explicit host declarations instead of a coarse desktop-platform guess.

**Problem:** Goose runtime is the MCP client of record, but MCP Apps rendering is a host capability owned by the ACP client. Before this change, Goose inferred downstream `capabilities.extensions` from `GoosePlatform`, which blurred ACP client vs MCP client responsibilities and made host identity and host capability sourcing too coarse.

**Solution:** Goose 2 now declares MCP Apps host capability and host identity during ACP `initialize`, `goose serve` stores that state per ACP connection, and Goose runtime translates it into a normal downstream MCP `initialize`. Goose runtime remains the MCP client of record, but it no longer invents host UI capability on its own; Goose 2 is now the source of truth for host rendering capability and user-facing identity, with the older platform behavior kept as a compatibility fallback.

The shortest accurate mental model for this stack is:

- Goose 2 is the ACP client and MCP Apps host.
- `goose serve` is the ACP server front door.
- Goose runtime lives behind that front door in the same backend stack.
- Goose runtime is the MCP client that connects outward to MCP servers.

```mermaid
sequenceDiagram
    participant UI as "Goose 2 UI (ACP client / MCP Apps host)"
    participant Serve as "goose serve (ACP server entry point)"
    participant Acp as "ACP session handler (ACP server)"
    participant Runtime as "Goose runtime (in-process MCP client)"
    participant Mcp as "External MCP server"

    UI->>Serve: open ACP WebSocket
    Serve->>Acp: create connection-scoped ACP handler
    UI->>Acp: initialize(clientInfo + host capabilities)
    Acp->>Acp: store per-connection host declarations
    UI->>Acp: newSession / loadSession / prompt
    Acp->>Runtime: create or load agent with host info
    Runtime->>Mcp: initialize(clientInfo + capabilities.extensions)
    Mcp-->>Runtime: tools/resources/app responses
    Runtime-->>Acp: assistant text + tool results
    Acp-->>UI: sessionUpdate notifications
```

The downstream MCP `initialize` payload now changes like this:

```diff
 {
   "jsonrpc": "2.0",
   "id": 0,
   "method": "initialize",
   "params": {
     "protocolVersion": "2025-03-26",
     "capabilities": {
-      "extensions": {},
+      "extensions": {
+        "io.modelcontextprotocol/ui": {
+          "mimeTypes": [
+            "text/html;profile=mcp-app"
+          ]
+        }
+      },
       "roots": {},
       "sampling": {},
       "elicitation": {}
     },
     "clientInfo": {
-      "name": "goose-cli",
+      "name": "goose2",
-      "version": "1.31.0"
+      "version": "0.1.0"
     }
   }
 }
```

That role split is the main architectural intent of this PR:

- Goose 2 is the source of truth for host rendering capability and user-facing identity.
- `goose serve` is the ACP boundary where those declarations are captured per connection.
- Goose runtime is still the MCP client of record, but now translates host declarations into ordinary downstream MCP initialization instead of inferring them from platform alone.

This is part 1 of 3, tracked in https://github.com/aaif-goose/goose/issues/8591

### DIFF
--- a/crates/goose-cli/src/cli.rs
+++ b/crates/goose-cli/src/cli.rs
@@ -1,6 +1,7 @@
 use anyhow::Result;
 use clap::{Args, CommandFactory, Parser, Subcommand};
 use clap_complete::{generate, Shell as ClapShell};
+use goose::agents::GoosePlatform;
 use goose::builtin_extension::register_builtin_extensions;
 use goose::config::{Config, GooseMode};
 #[cfg(feature = "telemetry")]
@@ -1075,11 +1076,20 @@ async fn handle_serve_command(host: String, port: u16, builtins: Vec<String>) ->
     } else {
         builtins
     };
+    let goose_platform = match std::env::var("GOOSE_PLATFORM").ok().as_deref() {
+        Some("desktop") | Some("goose-desktop") => GoosePlatform::GooseDesktop,
+        Some("cli") | Some("goose-cli") | None => GoosePlatform::GooseCli,
+        Some(other) => {
+            warn!("Unknown GOOSE_PLATFORM value '{other}', defaulting ACP server to goose-cli");
+            GoosePlatform::GooseCli
+        }
+    };
 
     let server = Arc::new(AcpServer::new(AcpServerFactoryConfig {
         builtins,
         data_dir: Paths::data_dir(),
         config_dir: Paths::config_dir(),
+        goose_platform,
     }));
     let router = goose_acp::transport::create_router(server);
 
--- a/crates/goose-cli/src/cli.rs
+++ b/crates/goose-cli/src/cli.rs
@@ -1079,20 +1079,12 @@ async fn handle_serve_command(host: String, port: u16, builtins: Vec<String>) ->
     } else {
         builtins
     };
-    let goose_platform = match std::env::var("GOOSE_PLATFORM").ok().as_deref() {
-        Some("desktop") | Some("goose-desktop") => GoosePlatform::GooseDesktop,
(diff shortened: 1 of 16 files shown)


aharvard:
@DOsinga / @alexhancock — I think y'all might be the closest to this work, can y'all plz TAL? Thanks!

crates/goose-cli/src/cli.rs
```diff
@@ -1078,11 +1079,20 @@ async fn handle_serve_command(host: String, port: u16, builtins: Vec<String>) ->
     } else {
         builtins
     };
+    let goose_platform = match std::env::var("GOOSE_PLATFORM").ok().as_deref() {
```
jamadeo: it doesn't seem right that serve should look at GOOSE_PLATFORM at all, or maybe I'm missing something there? IMO it should look only at the capabilities in the initialization handshake

jamadeo (approved):
Nice. My only point of confusion is around why we would need the GOOSE_PLATFORM env variable in an ACP context. If we can do without, I think we should

crates/goose-cli/src/cli.rs
aharvard: Good call. I removed the GOOSE_PLATFORM check from goose serve so the ACP server no longer infers desktop vs CLI from env there. For this path we now rely on the initialize handshake to declare host extensions/capabilities.

PR #8624, opened 2026-04-17
goose2 session working dir

lifeizhou-ap:
## Summary

This PR cleans up ui/goose2 ACP session working directory handling.

### Changes

- Renamed path-related variables and state so their purpose is clearer.

- Introduced `sessionCwdSelection.ts` to centralize ACP session working directory resolution.

- Added a Tauri-backed path resolver so path joining and ~ expansion are centralized instead of built ad hoc in ts.
 
- Cleaned up the ACP boundary so acpPrepareSession now requires a real workingDir.

### Discussion
I intentionally kept the Tauri backend thin in this PR. The backend only handles low-level path resolution, while frontend still decides which path should be used for the ACP session working directory. This gives us one place for OS-specific path handling like joining segments and expanding ~, without moving app concept (eg project) into the Tauri backend.

I also thought about pushing more of this into goose serve or even Goose core. But it becomes a bigger design question,  we need to decide whether Goose core should know about project-related concepts , and if so, how the domain like project, workspace, artifact root, and session working directory should be represented (eg domain design)

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:

### DIFF
--- a/ui/goose2/scripts/check-file-sizes.mjs
+++ b/ui/goose2/scripts/check-file-sizes.mjs
@@ -11,7 +11,7 @@ const EXCEPTIONS = {
       "Drag-and-drop handlers for session-to-project moves and project reorder, plus activeProjectId highlight.",
   },
   "src/features/chat/ui/ChatView.tsx": {
-    limit: 535,
+    limit: 560,
     justification:
       "ACP prewarm guards, project-aware working dir selection, working context sync, and chat bootstrapping still live together here.",
   },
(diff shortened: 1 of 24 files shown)


baxen (approved):
This looks good directionally. A few followups around the default path and artifact dir conventions but i plan to just send a followup PR instead - i think we can merge this one!

jamadeo:
Yes I think this is the right way to do this. An alternative would be https://v2.tauri.app/plugin/file-system/. As for goose core, I think we'll want to settle on how projects should be implemented before we'd move it into there

PR #8626, opened 2026-04-17
fix: eliminate ~5s delay when opening Switch Models panel with local provider

jh-block:
**Category:** fix
**User Impact:** The "Switch models" panel now opens instantly instead of showing "Loading models…" for ~5 seconds when the local inference provider is selected.

**Problem:** Opening the Switch Models panel triggered a long blocking load. Two things contributed: (1) the `list_local_models` endpoint called `ensure_featured_models_in_registry()` which made up to 6 sequential HTTP requests to HuggingFace to resolve metadata for featured models — models the UI immediately filtered out since only downloaded models are selectable, and (2) the modal eagerly fetched model lists from every configured provider in parallel via `Promise.all`, blocking the UI until the slowest provider responded, even though only one provider's models are displayed at a time.

**Solution:** Removed the unnecessary HuggingFace sync from the model listing endpoint and moved it to a dedicated `POST /local-inference/sync-featured` endpoint called only by the Settings → Models management page that actually needs it. Parallelized the HF API calls with `join_all` for when the sync is needed. Changed the modal to lazy-load models per-provider on selection instead of eagerly fetching all, with results cached so switching back is instant.

<details>

**crates/goose-server/src/routes/local_inference.rs**
Removed `ensure_featured_models_in_registry()` from `list_local_models` so it returns instantly from the registry. Added a new `POST /local-inference/sync-featured` endpoint. Restructured `ensure_featured_models_in_registry` into 3 phases (check registry → resolve concurrently via `join_all` → update registry) to parallelize HuggingFace API calls.

**crates/goose-server/src/openapi.rs**
Registered the new `sync_featured_models` endpoint in the OpenAPI spec.

**ui/desktop/openapi.json**
Generated OpenAPI spec with the new endpoint.

**ui/desktop/src/api/sdk.gen.ts**
Generated SDK client with `syncFeaturedModels` function.

**ui/desktop/src/api/types.gen.ts**
Generated types for the new endpoint.

**ui/desktop/src/api/index.ts**
Generated export for `syncFeaturedModels`.

**ui/desktop/src/components/settings/localInference/LocalInferenceSettings.tsx**
Calls `syncFeaturedModels()` before `listLocalModels()` so the Settings → Models management page still populates featured models for browsing/downloading.

**ui/desktop/src/components/settings/models/subcomponents/SwitchModelModal.tsx**
Split the initial load into two phases: provider list fetch (fast) and per-provider model fetch (on demand when selected). Results are cached in a ref so switching back to a previously-loaded provider is instant.

</details>

## Reproduction Steps

1. Configure the local inference provider (and optionally one or more cloud providers)
2. Open a chat session and click the model selector to open the "Switch models" panel
3. Select the local inference provider from the dropdown
4. Observe that the model list loads instantly instead of showing "Loading models…" for several seconds
5. Switch to another provider — its models load on first selection, then are cached for subsequent switches


crates/goose-server/src/routes/local_inference.rs
```diff
@@ -55,9 +56,17 @@ pub struct LocalModelResponse {
 }
 
 async fn ensure_featured_models_in_registry() -> Result<(), ErrorResponse> {
-    let mut entries_to_add = Vec::new();
     let mut mmproj_downloads_needed: Vec<(String, String, PathBuf)> = Vec::new();
 
+    // Phase 1: Check registry to find which featured models need HF resolution.
```
DOsinga: can we drop the phasing comments? I find that they are extra likely to rot

DOsinga (approved):

crates/goose-server/src/routes/local_inference.rs
jh-block: oops

Issue #8591, opened 2026-04-16
Implement MCP Apps in Goose2


aharvard:
> [!IMPORTANT]
>  NOTE: This parent issue and the related child issues/PRs are being refactored to leverage recent ACP client work.

Implement MCP Apps in Goose2 and expand ACP so Goose2 can behave like an MCP Apps host.

This work is landing as three stacked PRs, but this parent issue is the one-stop place to track the full feature.

When all three pieces land together, Goose2 should be able to negotiate MCP Apps capability, decide when a tool result should render as an app, and then bootstrap the inline app view inside the chat UI.

This parent issue tracks the three delivery areas in the order they are landing:

1. [Capability negotiation](https://github.com/aaif-goose/goose/issues/8592): teach Goose2 to declare MCP Apps capability during ACP initialize, teach Goose ACP to understand that metadata, and only advertise the MCP Apps extension in its MCP initialize request to the downstream MCP server when the connected client actually supports it.
2. [MCP App View bridge and bootstrap](https://github.com/aaif-goose/goose/issues/8594): load the sandbox proxy, bootstrap the MCP App View, deliver host context and tool updates, and handle host-proxied MCP methods plus host interaction requests like opening a link.
3. [Inline app rendering](https://github.com/aaif-goose/goose/issues/8593): hydrate `tools/list` after session binding, cache the catalog per ACP session, and use that catalog metadata to decide between normal tool rendering and the inline MCP App path.

Demo of a functional prototype:

VIDEO HERE

aharvard:
Initial implementation complete!

PR #8632, opened 2026-04-17
[goose2] MCP Apps: hydrate and replay app payloads in Goose2

aharvard:
**Category:** improvement  
**User Impact:** Goose2 can now show a session-scoped MCP App payload in chat and replay that same payload when an older chat is reopened.  
**Stacked On:** #8623  
**Problem:** PR #8623 establishes the MCP Apps capability handshake and host identity, but Goose2 still needed a durable way to surface an app payload after an MCP tool call completes. Without that, Goose2 had no backend-owned, replayable representation of the MCP App resource associated with a completed tool interaction.  
**Solution:** This PR introduces backend-owned MCP App association and resource hydration in Goose runtime. When an app-capable tool completes, Goose runtime reads the associated `ui://` resource, stores the resulting snapshot on `tool_result.meta.goose.mcpApp`, and forwards that through ACP `tool_call_update._meta.goose` for both live updates and replay. Goose2 now renders that payload through `McpAppView` as structured JSON; the next step is replacing that payload view with the real HTML renderer and bridge.

### Sequence

```mermaid
sequenceDiagram
    participant UI as "Goose 2 (ACP client / MCP Apps host)"
    participant ACP as "goose serve (ACP server)"
    participant Runtime as "Goose runtime (MCP client)"
    participant MCP as "MCP server"

    UI->>ACP: prompt
    ACP->>Runtime: agent.reply(...)
    Runtime->>MCP: callTool(...)
    MCP-->>Runtime: CallToolResult
    Runtime->>MCP: readResource(sessionId, ui://...)
    MCP-->>Runtime: ReadResourceResult
    Runtime->>Runtime: attach tool_result.meta.goose.mcpApp
    Runtime-->>ACP: completed ToolResponse
    ACP-->>UI: tool_call_update + _meta.goose.mcpApp
    UI->>UI: build McpApp payload
    UI->>UI: render McpAppView (JSON for now)

    Note over UI,ACP: later, on session reload
    UI->>ACP: loadSession(...)
    ACP->>ACP: replay stored ToolResponse from thread history
    ACP-->>UI: replayed tool_call_update + _meta.goose.mcpApp
    UI->>UI: reconstruct McpAppView from replay
```

### What This PR Introduces

This is intentionally the "payload-first" step.

This PR is the first backend implementation of MCP App payload hydration and replay in Goose2. Its job is to create a stable, session-scoped payload at tool-completion time, persist that payload with the tool result, and let Goose2 reconstruct the same `MCP App` view later when a session is reopened.

The current `McpAppView` is deliberately a structured payload view rather than the final HTML renderer. That keeps this PR focused on the data model, persistence, and replay path. The next PR can then build the real app surface on top of that payload with the iframe, proxy, and host bridge behavior.

### Other ACP Clients

Replayed MCP App payloads should remain part of session history even when an older Goose2-created session is later opened by another ACP client. Clients that do not render MCP Apps can ignore `_meta.goose.mcpApp`, while the session-scoped capability advertisement from #8623 still matters for new prompts: it prevents Goose runtime from doing fresh MCP App resource reads for a client that does not currently advertise MCP Apps support.

### Reproduction Steps

1. Add and activate this MCP server as a Goose extension: `https://mcp-app-bench.onrender.com/mcp`
2. Start a new session and prompt `run the app bench`
3. Note the JSON payload in view
 
image

This is part 2 of 3, tracked in #8591


morgmart (approved):

crates/goose/src/agents/extension_manager.rs
```diff
@@ -1432,17 +1534,6 @@ impl ExtensionManager {
...
-                    extension_name: owner,
-                    actual_tool_name: actual.to_string(),
-                    client,
-                });
-            }
-        }
-
         let tools = self.get_all_tools_cached(session_id).await.map_err(|e| {
             ErrorData::new(
                 ErrorCode::INTERNAL_ERROR,
```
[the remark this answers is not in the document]
aharvard: I don’t think this reproduces with the current implementation. resolve_tool now intentionally checks the listed tool set first so prefixed calls still respect available_tools filtering and preserve MCP app metadata. Also, fetch_all_tools() already degrades individual list_tools failures to warnings/empty results, so unrelated extension timeouts don’t usually cause get_all_tools_cached() to fail before the prefixed fallback runs. The remaining coupling is only on a true top-level internal cache/list failure, which is much narrower than the scenario described.

crates/goose/src/agents/extension_manager.rs
```diff
@@ -1530,15 +1670,32 @@ impl ExtensionManager {
...
                     ServiceError::McpError(error_data) => error_data,
                     _ => {
                         ErrorData::new(ErrorCode::INTERNAL_ERROR, e.to_string(), e.maybe_to_value())
                     }
-                })
+                })?;
+
+            remove_untrusted_mcp_app_meta(&mut result);
+
+            if should_hydrate_mcp_app && result.is_error != Some(true) {
```
jamadeo, line 1685: You could also consider fetching the resource even if the client doesn't support MCP apps, that way the snapshot is there in case you reconnect with a client that does support them

jamadeo (approved):
This looks great. I do want to dive more into the client replay support but that shouldn't block this PR.

Only question I had is whether you just want to fetch all the time, regardless of client support

crates/goose/src/agents/extension_manager.rs
aharvard: Great point! I think for now we should just fetch only in supporting clients. We're thinking about this a little bit differently with some applied work at Block for MCP servers. We're providing adaptive responses based off client capabilities, so we're not even sending MCP Apps down when the clients don't support them. This allows the server-side to control the user experience to be best for the client's capabilities.

PR #8633, opened 2026-04-17
feat(providers): strip chain-of-thought markers from custom provider output

### LINKED ISSUE #8590: Chain-of-Thought Leakage Breaking Chat Readability & Poor User Experience
Currently, when using custom LLM providers (via LLMProxy or OpenAI-compatible proxies), Goose displays the model's internal "Chain of Thought" or planning steps (e.g., "Structuring the comparison", "I need to provide a general comparison") as part of the normal chat output. This creates a "Thought Leakage" issue where the UI gets cluttered with internal reasoning, making the actual answer hard to read. It also sometimes causes repetitive loops in responses. This feature would significantly improve usability for users relying on local or proxy-based LLM setups where provider-side filtering is not available.

---

I would like a configurable setting (toggle or regex-based filter) called **"Sanitize Internal Reasoning"**. This feature should automatically detect and hide or collapse common chain-of-thought patterns (such as lines starting with "Considering", "Structuring", "I will", etc.) before rendering them in the chat UI. If possible, these filtered sections can be shown as a collapsed dropdown (e.g., "Show hidden reasoning") instead of being completely removed.

---

* **Prompt Injection (.goosehints):** Attempted to instruct the model to avoid showing reasoning, but this is unreliable and often ignored.
* **Concise Mode:** Enabled concise response style in settings, but it does not prevent reasoning leakage when the model outputs it as plain text.
* **Manual Ignoring:** Currently ignoring these sections manually, which is inefficient and disrupts readability.

---

**Additional context**
This issue primarily affects users using Virtual Keys and local proxy setups (like configure manually provider), where raw model output is passed directly to the UI without any post-processing or filtering.

Example issue:

* The agent repeatedly prints lines like "Structuring the comparison" or "I need to provide a general comparison" before the actual response.

---

* [x] I have verified this does not duplicate an existing feature request

mvanhorn:
## Summary

Custom OpenAI-compatible providers (DeepSeek R1, Qwen-style proxies, LLMProxy, etc.) emit reasoning inline in the content field as `<think>...</think>` or `<thinking>...</thinking>` blocks rather than using the structured `reasoning_content` JSON field. Goose surfaced those blocks directly in chat, cluttering the UI and sometimes triggering repetition.

This change adds a streaming-safe `ThinkFilter` / `split_think_blocks` helper in `crates/goose/src/providers/base.rs` and wires it into both the non-streaming (`response_to_message`) and streaming (`response_to_streaming_message`) paths in `crates/goose/src/providers/formats/openai.rs`. Inline thinking is routed into `MessageContent::thinking`; cleaned content flows through `MessageContent::text`. Structured `reasoning_content` / `reasoning` fields still take precedence, so providers that already send reasoning as a first-class field do not get duplicated output.

Streaming semantics: when a chunk ends on a partial tag prefix (`<thi`, `<think`, `</thinkin`, etc.), the filter holds that suffix until it can decide whether it belongs to a think block or is regular content. A chunk ending mid-tag will never leak characters to the UI that later turn out to be inside a think block.

### Testing

- `cargo fmt --check`: clean
- `cargo clippy -p goose --all-targets -- -D warnings`: clean
- `cargo test -p goose --lib providers::base::tests`: 23 passed (includes new think filter tests)
- `cargo test -p goose --lib providers::formats::openai::tests`: all passed (includes new regression tests)

New unit tests cover:
- Full-content splitter for `<think>`, `<thinking>`, attributes (`<think class="x">`), case-insensitive tags, multiple blocks, plain content, unterminated opens
- Streaming across partial tag boundaries (chunks like `["<thi", "nk>x</thi", "nk>y"]`)
- Non-think tags (`<table>`) are treated as content
- Structured `reasoning_content` path still wins when both are present

### Related Issues
Closes #8590

Prior art: zed's similar fix landed in zed-industries/zed#53488 ("Strip think tags from text stream") for an analogous UI-leakage issue.

This contribution was developed with AI assistance (Codex).

### DIFF
--- a/crates/goose/src/providers/base.rs
+++ b/crates/goose/src/providers/base.rs
@@ -24,6 +24,214 @@ use std::pin::Pin;
 use std::sync::LazyLock;
 use std::sync::Mutex;
 
+#[derive(Debug, Default, PartialEq, Eq)]
+pub struct FilterOut {
+    pub content: String,
+    pub thinking: String,
+}
+
+pub struct ThinkFilter {
+    buffer: String,
+    inside_think: bool,
+    think_depth: usize,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum ThinkTag {
+    Open,
+    Close,
+}
+
+enum BufferEvent {
+    Tag {
+        pos: usize,
+        end: usize,
+        kind: ThinkTag,
+    },
+    Partial(usize),
+}
+
+impl ThinkFilter {
+    pub fn new() -> Self {
+        Self {
+            buffer: String::new(),
+            inside_think: false,
+            think_depth: 0,
+        }
+    }
+
+    pub fn push(&mut self, chunk: &str) -> FilterOut {
+        self.buffer.push_str(chunk);
+        self.process_buffer()
+    }
+
+    pub fn finish(mut self) -> FilterOut {
+        let mut out = self.process_buffer();
+        if !self.buffer.is_empty() {
+            if self.inside_think {
+                out.thinking.push_str(&self.buffer);
+            } else {
+                out.content.push_str(&self.buffer);
+            }
+            self.buffer.clear();
+        }
+        out
+    }
+
+    fn process_buffer(&mut self) -> FilterOut {
+        let mut out = FilterOut::default();
+
+        loop {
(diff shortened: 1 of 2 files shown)


mvanhorn:
Apologies, this was created on the wrong fork. Re-submitting on block/goose.

PR #8648, opened 2026-04-19
fix: overlay config writes onto symlinked paths

caniko:
## Summary
Fix Goose config writes for declarative config managers such as Home Manager. When `config.yaml` or a custom provider JSON lives at a symlinked logical path, Goose should write an overlay file at the logical location instead of following the symlink into an immutable target.

## Changes
- add a shared atomic helper that writes to the logical config path and replaces a symlink with a regular file overlay
- use that helper for `config.yaml` writes in `crates/goose/src/config/base.rs`
- use the same helper for writable custom-provider JSON files in `crates/goose/src/config/declarative_providers.rs`
- add regression tests for symlinked `config.yaml`, a short symlink chain, and symlinked custom-provider JSON files

## Validation
- `nix develop -c cargo fmt --all`
- `nix develop -c cargo test -p goose --no-default-features --features rustls-tls write_replaces_symlink_with_overlay_file -- --nocapture`
- `nix develop -c cargo test -p goose --no-default-features --features rustls-tls write_replaces_head_of_symlink_chain -- --nocapture`
- `nix develop -c cargo test -p goose --no-default-features --features rustls-tls write_custom_provider_replaces_symlink_with_overlay_file -- --nocapture`

### DIFF
--- a/crates/goose/src/config/base.rs
+++ b/crates/goose/src/config/base.rs
@@ -34,6 +34,32 @@ fn write_secrets_file(path: &Path, content: &str) -> std::io::Result<()> {
     }
 }
 
+pub(crate) fn atomic_write_overlay(path: &Path, content: &[u8]) -> Result<(), ConfigError> {
+    if let Some(parent) = path.parent() {
+        std::fs::create_dir_all(parent).map_err(|e| ConfigError::DirectoryError(e.to_string()))?;
+    }
+
+    let temp_path = path.with_extension("tmp");
+
+    {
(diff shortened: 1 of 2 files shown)


caniko:
Closing this out because the writable-vs-immutable mismatch is owned by the Home Manager integration layer on the canix side. We are switching the fix to HM-managed seeded regular files for `config.yaml`, `permission.yaml`, and `custom_providers/*.json` instead of changing Goose's generic config-write behavior upstream.

PR #8650, opened 2026-04-20
used default fallback provider and model if default are set

lifeizhou-ap:
## Summary
Onboarding flow:  Fallback to default provider and model if this is specified in the environment variable if there is no provider configured.


michaelneale (approved):
there are 2 lines removed with session storage and local - I assume that was deliberate? but this makes sense otherwise.

lifeizhou-ap:
> there are 2 lines removed with session storage and local - I assume that was deliberate? but this makes sense otherwise.

yes, it is deliberate as a cleanup

Issue #8459, opened 2026-04-10
Backend server fails to start on fresh install (Windows desktop app)


khanhdoan1710:
**Describe the bug**

Goose desktop app shows "The backend server failed to start" right after installation

---

**To Reproduce**
1. Download
2. Unzip
3. Run Goose.exe

---

**Expected behavior**
Goose to run successfully

---

**Screenshots**

<img width="1916" height="1024" alt="Image" src="https://github.com/user-attachments/assets/0c325b58-af24-4e97-9fba-458dea94c7cd" />

---

**Please provide the following information**
- **OS & Arch:** Windows 11
- **Interface:** UI
- **Version:** Newest
- **Extensions enabled:** None
- **Provider & Model:** None

---

**Additional context**
- Fresh install

cidtrips:
Same issued with new install of 1.31.0

Server log files are created, but are completely empty.

lifeizhou-ap:
Hi @khanhdoan1710 @cidtrips ,

I cannot reproduce the issue on my Windows (Windows 10).   I am still trying to find the root cause.  Just wondering whether it works if you relaunch goose? 

I also created a branch [here](https://github.com/aaif-goose/goose/pull/8656), added more logs to figure out the root cause, and created a bundled desktop app.  Would you please download the  https://github.com/aaif-goose/goose/actions/runs/24656065228/artifacts/6528007994 and share the json file in C:\Users\<username>\AppData\Roaming\Goose\logs\startup\ when you have the issue.   Thank you!

The bundled app is not signed so you may have to click "Run anyway".

cidtrips:
@lifeizhou-ap 

Yes, happens regardless of number of restart attempts.

New message:

<img width="559" height="507" alt="Image" src="https://github.com/user-attachments/assets/36d94f5e-da7f-4017-ae0b-481032deb022" />

Here is your JSON:

{
  "attemptId": "goosed-startup-2026-04-21T01-24-03.149Z-23416.json",
  "startedAt": "2026-04-21T01:24:03.149Z",
  "goosedPath": "C:\\Users\\<redacted>\\Downloads\\windows-unsigned\\resources\\bin\\goosed.exe",
  "workingDir": "C:\\Users\\<redacted>",
  "baseUrl": "https://127.0.0.1:55169",
  "pid": 17908,
  "certFingerprintSeen": false,
  "healthCheckSucceeded": false,
  "childExitCode": 3221225477,
  "childExitSignal": null,
  "stderrTail": [
[...]

PR #8656, opened 2026-04-20
fix: lazy init local inference runtime in router instead of app start

lifeizhou-ap:
## Summary
Fix Issue: https://github.com/aaif-goose/goose/issues/8459

### Changes
- Init the local inference runtime in local inference router instead of the goosed started
- Added minimal startup logs for debug in the future and added diagnose doc (I was thinking about including this in the `doctor` but the log file is generated on the client side and server side does not know unless we put hard coded directory, we can think about more of this when start up error occurs more often)

### Testing
Tested local inference feature


ui/desktop/src/goosed.ts
```diff
@@ -187,10 +292,12 @@ export const startGoosed = async (options: StartGoosedOptions): Promise<GoosedRe
...
     logger = defaultLogger,
+    diagnosticsDir,
   } = options;
 
   const errorLog: string[] = [];
   const workingDir = dir || os.homedir();
+  const startupTrace = createStartupDiagnostics(diagnosticsDir, workingDir, null, null);
 
   if (externalGoosed?.enabled && externalGoosed.url) {
     const url = externalGoosed.url.replace(/\/$/, '');
```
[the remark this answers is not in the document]
lifeizhou-ap: fixed

crates/goose-server/src/state.rs
```diff
@@ -48,11 +48,18 @@ impl AppState {
...
-            inference_runtime: InferenceRuntime::get_or_init(),
+            inference_runtime: Arc::new(OnceLock::new()),
             session_buses: Arc::new(Mutex::new(HashMap::new())),
         }))
     }
 
+    #[cfg(feature = "local-inference")]
+    pub fn get_inference_runtime(&self) -> Arc<InferenceRuntime> {
+        self.inference_runtime
+            .get_or_init(InferenceRuntime::get_or_init)
```
jh-block: Maybe we can make this fallible (make InferenceRuntime::get_or_init return a Result rather than panicking, and return Result in this function) so that we can gracefully handle problems with starting the inference runtime

ui/desktop/src/startupDiagnostics.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  for (const startupLog of startupLogs.slice(STARTUP_LOGS_TO_KEEP)) {
+    fs.unlinkSync(startupLog.filePath);
+  }
+};
+
+export const createStartupDiagnostics = (
+  diagnosticsDir: string | undefined,
+  workingDir: string,
+  goosedPath: string | null,
```
jh-block: The only call site of this function passes null for `goosedPath` and `baseUrl` and then mutates `startupTrace.diagnostics.goosedPath` / `.baseUrl` later. Either drop the parameters, or use them instead of the manual mutation of the StartupTrace

ui/desktop/src/startupDiagnostics.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+export interface StartupTrace {
+  diagnosticsPath: string;
+  diagnostics: StartupDiagnostics;
+  record: (name: string, details?: Record<string, unknown>) => void;
+  flush: () => void;
+}
+
+const STARTUP_TAIL_LIMIT = 40;
```
jh-block: I think this needs to be a little higher (80 or even 100), the diag in 8459 was truncated right before the actual crash.

ui/desktop/src/startupDiagnostics.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (name === 'healthcheck_success') {
+      diagnostics.healthCheckSucceeded = true;
+    }
+    diagnostics.events.push({
+      name,
+      at: new Date().toISOString(),
+      elapsedMs: Date.now() - monotonicStart,
+      ...(details ? { details } : {}),
+    });
+    flush();
```
jh-block, line 108: flushing after every line might slow down startup a little, but it's probably not too bad

jh-block (approved):
This is a good mitigation, allowing users for whom local inference fails to start to still use the rest of the app, but I think we should go further and find out why local inference isn't working for them and fix it -- one clue in #8459 is the MSVC version issue. I am happy to take a closer look at that!

crates/goose-server/src/state.rs
lifeizhou-ap: That's a good idea!

jh-block (approved):

DOsinga (approved):

PR #8658, opened 2026-04-20
fix(developer): run shell tool under bash/sh regardless of login shell

### LINKED ISSUE #7348: CLI is incompatible to fish
when using fish as shell, CLI fails to modify files. (It can initially write files.)

after some inquiries, goose with gpt-codex was able to diagnose the issue and get around it (see below); however, I was near to give up at the point, having wasted two hours with failed experiments

the CLI should detect the active shell and either internally use another shell or at least warn it is incompatible to fish

---

```
prompt: try if you can write a file in the current directory

(works)

prompt: the tool calls somehow don't work; I can see goose_write_test.txt (so you CAN write a new file), but other_file.py has not been changed (modify time has not been updated since I started this session); what's going wrong?

answer: Thanks for the patience — you were right: the earlier writes were failing because I was using a `cat <<EOF > file` redirection under **fish**, and fish doesn’t allow that syntax in the way I used it. That’s why the file’s mtime never changed.

I’ve now rewritten the file using `/bin/sh`, which **does support here-doc redirection**, so the write actually succeeded. The modification time was updated.
```

jh-block:
## Problem

The developer extension's `shell` tool previously fell back to the user's `$SHELL` when `/bin/bash` wasn't present. Users with a non-POSIX login shell (fish, csh, tcsh, nu, ...) would then see failures whenever the model emitted common idioms like heredoc redirection (`cat <<EOF > file`), which those shells don't support.

Fixes #7348.

## Fix

- Always run commands under `bash` (when on PATH) or `sh`; `GOOSE_SHELL` still overrides.
- Include the selected shell's name in the `shell` tool description so the model tailors syntax accordingly (and `GOOSE_SHELL` users get the dialect they asked for).
- Tidy-up: use `which` to probe for bash instead of hard-coded paths, drop a redundant Flatpak short-circuit (the `flatpak-spawn --host` wrapping is unchanged), store shells as basenames so PATH resolution does the work, and factor Windows/Unix basename handling behind a shared helper.

### DIFF
--- a/crates/goose/src/agents/platform_extensions/developer/mod.rs
+++ b/crates/goose/src/agents/platform_extensions/developer/mod.rs
@@ -15,7 +15,7 @@ use rmcp::model::{
 };
 use schemars::{schema_for, JsonSchema};
 use serde_json::Value;
-use shell::{ShellOutput, ShellParams, ShellTool};
+use shell::{shell_display_name, ShellOutput, ShellParams, ShellTool};
 use std::sync::Arc;
 use tokio_util::sync::CancellationToken;
 use tree::{TreeParams, TreeTool};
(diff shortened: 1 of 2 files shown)


jh-block:
superseded by #8659

PR #8667, opened 2026-04-20
upgrade goose sdk and tui to be compatible with the latest agentclientprotocol/sdk package

lifeizhou-ap:
## Summary
- Upgrade version on agentclientprotocol/sdk in goose-sdk and tui
- Make the code compatible with new version
- Upgrade the version of goose-sdk (ui/sdk) and goose (ui/text)

### Question and Discussions
1. @alexhancock can you please help testing TUI?  I started the Tui (even with the old version) locally, I cannot move the cursor or use the keys to show the existing provider and models on the Tui.  Maybe I don't know how to use it properly

2.   I noticed ui/sdk, ui/text and ui/goose-binary have the same version.  I guess this is easier to manage the versions for these packages in the same repo.  Shall i upgrade the goose-binary to have the same version as sdk and text

3.   Shall we use `changeset` to do the changelog and versioning, I guess changeset probably have better flow to add changelog, or we can manually do it.  Either way works. WDYT?

4. If we want to have same versions for these 3 packages, I can add another check in the publish npm workflow. 

### Testing
Have trouble testing at the moment


alexhancock (approved):
I tested with the TUI and it works great via `cd ui/text && npm start`

lifeizhou-ap:
> I tested with the TUI and it works great via `cd ui/text && npm start`

Thank you!

lifeizhou-ap:
Hi @alexhancock,

Would you mind taking another look at items 2, 3, and 4 under ‘Questions and Discussions’?  I would like to make sure the versions are correct (same version for all three npm packages or different versions are ok) Thank you! 

cc: @jamadeo

PR #8696, opened 2026-04-20
fix: handle commas within values for CUSTOM_HEADERS

### LINKED ISSUE #8495: Not respecting OPENAI_CUSTOM_HEADERS that have commas in values
**Describe the bug**

OPENAI_CUSTOM_HEADERS parsing does not allow for commas in the mapped values (example of this is x-litellm-tags - as documented here: https://docs.litellm.ai/docs/proxy/request_headers. - which supports a comma delimited list, but the parsing method here doesn't support it)

So in short, you can not have more than one value mapped to any header (because commas are assumed to not be in the value).

---

**To Reproduce**
Set or pass any custom header that has a comma in the value - see that only the first value is respected because the logic splits on the command and then sends the other value as a header. For example:

export OPENAI_CUSTOM_HEADERS=x-litellm-tags=tag1,tag2 

Review of this function identifies the problem - it just blindly splits on commas first, without seeing if thy are within quotes for example. https://github.com/aaif-goose/goose/blob/3f5277538d5f6df1a88faca5e77cecc0ee208ae1/crates/goose/src/providers/openai.rs#L596

---

**Expected behavior**

I expect that commas within quotes will be respectd. The following should parse cleanly:

x-litellm-tags="tag1,tag2",x-litellm-spend-logs-metadata="{\"a\":\"b\",\"c\":\"d\"}"
x-litellm-tags='tag1,tag2',x-litellm-spend-logs-metadata="{'a':'b','c':'d'}"

should parse these:

x-litellm-tags=tag1,tag2
x-litellm-spend-logs-metadata={'a':'b','c':'d'}

---

- **OS & Arch:** Docker Linux 
- **Interface:** CLI
- **Version:** 1.27.0
- **Extensions enabled:** N/A
- **Provider & Model:** openai (custom provider, in this case using openai to point to litellm)

----

AI Suggested this change:

Method to change: 
https://github.com/aaif-goose/goose/blob/3f5277538d5f6df1a88faca5e77cecc0ee208ae1/crates/goose/src/providers/openai.rs#L596

AI Suggested Change:
```
fn parse_custom_headers(s: String) -> std::collections::HashMap<String, String> {

    fn split_on_commas_outside_quotes(input: &str) -> Vec<&str> {
        let mut parts = Vec::new();
        let mut start = 0usize;

        let mut in_quotes = false;
        let mut prev_was_backslash = false;

        for (i, ch) in input.char_indices() {
            match ch {
                '"' if !prev_was_backslash => in_quotes = !in_quotes,
                ',' if !in_quotes => {
                    parts.push(input[start..i].trim());
                    start = i + 1;
                }
                _ => {}
            }

            // Only treat backslash as escaping the *next* character.
            prev_was_backslash = ch == '\\' && !prev_was_backslash;
        }

        parts.push(input[start..].trim());
        parts.into_iter().filter(|p| !p.is_empty()).collect()
    }

    fn unquote_and_unescape(v: &str) -> String {
        let v = v.trim();
        let inner = if v.len() >= 2 && v.starts_with('"') && v.ends_with('"') {
            &v[1..v.len() - 1]
        } else {
            v
        };

        // Minimal unescape: \" and \\ are common in env var strings.
        let mut out = String::with_capacity(inner.len());
        let mut chars = inner.chars();
        while let Some(c) = chars.next() {
            if c == '\\' {
                if let Some(next) = chars.next() {
                    match next {
                        '"' => out.push('"'),
                        '\\' => out.push('\\'),
                        // keep unknown escapes as-is (e.g., \n) rather than guessing
                        other => {
                            out.push('\\');
                            out.push(other);
                        }
                    }
                } else {
                    out.push('\\');
                }
            } else {
                out.push(c);
            }
        }
        out
    }

    split_on_commas_outside_quotes(&s)
        .into_iter()
        .filter_map(|header| {
            let (k, v) = header.split_once('=')?;
            let key = k.trim();
            if key.is_empty() {
                return None;
            }
            let value = unquote_and_unescape(v);
            Some((key.to_string(), value))
        })
        .collect()
}
```

with tests like this:

```
#[test]
fn parse_custom_headers_respects_commas_in_quotes() {
    let headers = r#"A="1,2",B=3"#.to_string();
    let parsed = parse_custom_headers(headers);
    assert_eq!(parsed.get("A").unwrap(), "1,2");
    assert_eq!(parsed.get("B").unwrap(), "3");
}

#[test]
fn parse_custom_headers_trims_and_unquotes() {
    let headers = r#" A = "x, y" , B = z "#.to_string();
    let parsed = parse_custom_headers(headers);
    assert_eq!(parsed.get("A").unwrap(), "x, y");
    assert_eq!(parsed.get("B").unwrap(), "z");
}

#[test]
fn parse_custom_headers_supports_escaped_quotes() {
    let headers = r#"A="hello \"world\"",B=ok"#.to_string();
    let parsed = parse_custom_headers(headers);
    assert_eq!(parsed.get("A").unwrap(), r#"hello "world""#);
    assert_eq!(parsed.get("B").unwrap(), "ok");
}
```

sunilkumarvalmiki:
Resolves #8495. Improves \parse_custom_headers\ to correctly handle commas inside quoted strings in \OPENAI_CUSTOM_HEADERS\ variables.

### DIFF
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -683,7 +683,7 @@ fn split_on_commas_outside_quotes(s: &str) -> Vec<String> {
 }
 
 fn parse_custom_headers(s: String) -> HashMap<String, String> {
-    fn unquote_and_unescape(val: &str) -> String {
+    fn normalize_header_value(val: &str) -> String {
         let trimmed = val.trim();
         let Some(inner) = trimmed.strip_prefix('"').and_then(|s| s.strip_suffix('"')) else {
             return trimmed.to_string();
@@ -710,7 +710,12 @@ fn parse_custom_headers(s: String) -> HashMap<String, String> {
                 out.push(c);
             }
         }
-        out
+
+        if inner.contains(',') {
+            out
+        } else {
+            format!("\"{}\"", out)
+        }
     }
 
     split_on_commas_outside_quotes(&s)
@@ -721,7 +726,7 @@ fn parse_custom_headers(s: String) -> HashMap<String, String> {
             if key.is_empty() {
                 return None;
             }
-            let value = unquote_and_unescape(v);
+            let value = normalize_header_value(v);
             Some((key.to_string(), value))
         })
         .collect()
@@ -824,7 +829,15 @@ mod tests {
     fn test_parse_custom_headers_supports_escaped_quotes() {
         let headers = r#"A="hello \"world\"",B=ok"#.to_string();
         let parsed = parse_custom_headers(headers);
-        assert_eq!(parsed.get("A").unwrap(), r#"hello "world""#);
+        assert_eq!(parsed.get("A").unwrap(), r#""hello "world"""#);
+        assert_eq!(parsed.get("B").unwrap(), "ok");
+    }
+
+    #[test]
+    fn test_parse_custom_headers_preserves_literal_quotes_without_commas() {
+        let headers = r#"If-None-Match="abc",B=ok"#.to_string();
+        let parsed = parse_custom_headers(headers);
+        assert_eq!(parsed.get("If-None-Match").unwrap(), r#""abc""#);
         assert_eq!(parsed.get("B").unwrap(), "ok");
     }
 
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -664,7 +664,7 @@ fn split_on_commas_outside_quotes(s: &str) -> Vec<String> {
         if escape_next {
             current.push(c);
             escape_next = false;
-        } else if c == '\\' {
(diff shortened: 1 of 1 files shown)


lifeizhou-ap:
Thanks for the fix, @sunilkumarvalmiki — and nice work linking it to #8495! We noticed you have 12 open PRs, nearly all opened on the same day within an hour of each other. Per our [contribution guidelines](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#pull-requests), we'd recommend focusing on one at a time. We're closing these for now — feel free to reopen your top choice, and once it lands you can open the next one. If you reopen this one, there's also an unresolved Codex P2 about `in_quotes` toggling on literal `"` in unquoted values that would be worth addressing.

PR #8700, opened 2026-04-20
fix(providers): handle missing delta field in streaming chunks

### LINKED ISSUE #8641: Ran into this error: Request failed: Failed to parse streaming chunk: missing field delta
**Describe the bug**

Goose responses fail with:

```
Ran into this error: Request failed: Failed to parse streaming chunk: missing field delta: "{"choices":[{"content_filter_offsets":{"check_offset":54781,"start_offset":54781,"end_offset":55778},"content_filter_results":{"hate":{"filtered":false,"severity":"safe"},"self_harm":{"filtered":false,"severity":"safe"},"sexual":{"filtered":false,"severity":"safe"},"violence":{"filtered":false,"severity":"safe"}},"finish_reason":null,"index":0}],"created":0,"id":"","model":"","object":""}".
```

---

- **OS & Arch:** macOS 15.7.4
- **Interface:** UI
- **Version:** 1.31
- **Extensions enabled:** 

Image

- **Provider & Model:** Azure OpenAI/GPT 5.4

sunilkumarvalmiki:
**Category:** fix
**User Impact:** Users will no longer encounter streaming crashes with providers (like Azure OpenAI) that omit the delta object in filtering or final chunks.
**Problem:** Some providers conform to the streaming format but omit the delta field entirely when yielding final content-filter results rather than yielding delta: {}. This caused a direct serde validation failure during stream parsing.
**Solution:** Added #[serde(default)] to the delta field and derived Default for the schema so omitted fields safely deserialize to a null delta message, allowing the stream to complete successfully.

<details>

**openai.rs**
Derived Default on Delta and relaxed the validation on StreamingChoice::delta so it uses the default (null content) fallback if the field is missing from the provider's chunk.

</details>

Fixes #8641

### DIFF
--- a/crates/goose/src/providers/formats/openai.rs
+++ b/crates/goose/src/providers/formats/openai.rs
@@ -64,7 +64,7 @@ struct ContentPart {
     thought_signature: Option<String>,
 }
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, Default)]
 struct Delta {
     #[serde(default)]
     content: Option<DeltaContent>,
@@ -77,6 +77,7 @@ struct Delta {
 
 #[derive(Serialize, Deserialize, Debug)]
 struct StreamingChoice {
+    #[serde(default)]
     delta: Delta,
     index: Option<i32>,
     finish_reason: Option<String>,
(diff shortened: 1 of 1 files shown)


lifeizhou-ap:
Hi @sunilkumarvalmiki The delta change looks good and should fix the reported Azure case.  
 
 I pushed a commit so that we can finalise this pr soon.  I kept the delta defaulting change and reverted the usage change and added a regression test for the missing-delta case.

lifeizhou-ap (approved):

PR #8710, opened 2026-04-21
consolidate provider ACP methods onto inventory

baxen:
**Category:** improvement
**User Impact:** Users now get provider setup and selection data from a single ACP inventory endpoint instead of separate provider list and details methods.
**Problem:** Goose exposed separate ACP methods for provider list, provider details, and provider inventory even though clients in this repo were converging on the same underlying provider data. That duplication created drift in the contracts, including inconsistent configuration metadata and redundant maintenance across server handlers, schema generation, and UI consumers.
**Solution:** This change makes `_goose/providers/inventory` the single read surface for provider metadata plus live inventory state, migrates the in-repo clients to that endpoint, and removes the legacy list/details ACP methods and compat-only inventory fields. The result is a smaller ACP surface with one contract to maintain, while preserving the internal provider-option helper used for session config.

<details>

**crates/goose-acp/acp-meta.json**
Removed the legacy provider list and details method registrations so ACP metadata only advertises inventory-based provider reads.

**crates/goose-acp/acp-schema.json**
Regenerated the ACP schema to reflect the richer inventory entry shape, required metadata fields, and removal of the deprecated provider list/details methods.

**crates/goose-acp/src/server.rs**
Expanded inventory DTO mapping to include provider catalog metadata, removed the old ACP list/details handlers, and kept provider-option building as an internal helper for session config.

**crates/goose-acp/tests/custom_requests_test.rs**
Replaced the old list-provider coverage with inventory metadata assertions and restored the missing `#[test]` registration so the new path actually runs.

**crates/goose-sdk/src/custom_requests.rs**
Promoted inventory entry metadata into the SDK request/response contract, made catalog fields required, and removed the old list/details request and response types.

**crates/goose/src/providers/inventory/mod.rs**
Moved provider catalog metadata onto inventory entries and descriptors so the inventory service can serve both setup metadata and live model inventory from one source of truth.

**ui/goose2/src/shared/api/acpApi.ts**
Switched provider discovery to inventory-backed entries while preserving the synthetic default Goose provider for picker flows.

**ui/goose2/tests/e2e/fixtures/tauri-mock.ts**
Dropped the unused legacy provider-list mock branch now that UI callers only use inventory.

**ui/sdk/src/generated/client.gen.ts**
Regenerated the typed SDK client so provider inventory is the only provider-read custom method and its richer response shape is reflected in generated calls.

**ui/sdk/src/generated/index.ts**
Regenerated exported SDK method/type metadata after removing the legacy provider list/details ACP methods.

**ui/sdk/src/generated/types.gen.ts**
Regenerated TypeScript types so provider inventory entries carry required catalog/setup metadata and the old list/details types disappear.

**ui/sdk/src/generated/zod.gen.ts**
Regenerated Zod schemas so inventory metadata fields are required and the removed list/details contracts no longer parse.

**ui/text/src/configure.tsx**
Migrated the text configure flow to inventory-backed provider metadata, model loading, and configured-state checks.

**ui/text/src/onboarding.tsx**
Migrated onboarding provider selection and setup to inventory-backed metadata so it no longer depends on the removed details method.

</details>

1. Run `cargo test` and confirm the ACP schema, SDK generation, and provider inventory changes compile and the unit/integration suite passes apart from any pre-existing live provider integration failures in your environment.
2. Open the text onboarding or configure flow and verify provider selection, setup prompts, and model selection still work while sourcing metadata from `_goose/providers/inventory`.
3. Inspect the generated ACP schema and SDK outputs to confirm `_goose/providers/list` and `_goose/providers/details` no longer appear and inventory entries expose `description`, `defaultModel`, `providerType`, `configKeys`, and `setupSteps`.

**Validation:** `cargo fmt` passed. `cargo test` completed with two failures in `crates/goose/tests/providers.rs`: `test_codex_provider` failed because the current Codex ChatGPT account does not support `gpt-5.2-codex`, and `test_anthropic_provider` failed on an existing live tool-output assertion mismatch.


crates/goose-sdk/src/custom_requests.rs
```diff
@@ -181,24 +181,6 @@ pub struct RemoveSecretRequest {
     pub key: String,
 }
 
-/// List providers available through goose, including the config-default sentinel.
-#[derive(Debug, Default, Clone, Serialize, Deserialize, JsonSchema, JsonRpcRequest)]
-#[request(method = "_goose/providers/list", response = ListProvidersResponse)]
```
alexhancock, line 186: nit: I would personally keep this naming for the converged thing. "Inventory" feels like it's something that shrinks and grows and is always changes but providers is more something we'll just keep adding to.

ui/sdk/src/generated/client.gen.ts
```diff
@@ -121,20 +115,6 @@ export class GooseExtClient {
...
   }
 
-  async GooseProvidersList(
-    params: ListProvidersRequest,
-  ): Promise<ListProvidersResponse> {
-    const raw = await this.conn.extMethod("_goose/providers/list", params);
-    return zListProvidersResponse.parse(raw) as ListProvidersResponse;
-  }
-
-  async GooseProvidersDetails(
```
alexhancock, line 147: the TUI calls this one a couple times

https://github.com/aaif-goose/goose/blob/8eda6fdabc108b82700860e5099693687f8c0734/ui/text/src/configure.tsx#L381

https://github.com/aaif-goose/goose/blob/8eda6fdabc108b82700860e5099693687f8c0734/ui/text/src/onboarding.tsx#L538

alexhancock (approved):
LGTM after the two commented things are addressed!

ui/sdk/src/generated/client.gen.ts
baxen: yes update the use of this in this PR!

PR #8719, opened 2026-04-21
fix: avoid sqlite session creation panic on legacy dbs

rtzll:
## Summary
- fix a startup crash when Goose creates a new session against an older SQLite `sessions` table shape
- stop using `INSERT ... RETURNING *` during session creation
- insert the session row first, then load it with an explicit `SELECT`, which avoids the `sqlx-sqlite` row-shape
panic seen after upgrading to `1.31.1`

### Testing
- manually reproduced the crash with an existing user `sessions.db` copied into an isolated `GOOSE_PATH_ROOT`
- confirmed the failing path was `SessionStorage::create_session` during `INSERT ... RETURNING *`
- validated the behavior by rebuilding the copied `sessions` table and confirming `goose` started normally
- built and ran the patched CLI locally against the existing session DB and confirmed Goose reached the normal interactive prompt without the SQLite panic

### DIFF
--- a/crates/goose/src/session/session_manager.rs
+++ b/crates/goose/src/session/session_manager.rs
@@ -1084,33 +1084,48 @@ impl SessionStorage {
         let mut tx = pool.begin_with("BEGIN IMMEDIATE").await?;
 
         let today = chrono::Utc::now().format("%Y%m%d").to_string();
-        let session = sqlx::query_as(
+        let next_id = (sqlx::query_scalar::<_, Option<i64>>(
             r#"
-                INSERT INTO sessions (id, name, user_set_name, session_type, working_dir, extension_data, goose_mode)
-                VALUES (
-                    ? || '_' || CAST(COALESCE((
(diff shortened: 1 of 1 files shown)


jh-block:
Can you provide some more detail about the crash & the "legacy DB" that it was applying against? Either here or in an associated issue if there is one. It sounds like we may have a faulty migration if it was to do with the DB schema not matching the DML being executed.

rtzll:
After upgrading to `1.31.1`I ran `goose` in the terminal which produced the following output:
```bash
thread 'sqlx-sqlite-worker-0' (18837320) panicked at /Users/brew/Library/Caches/Homebrew/cargo_cache/registry/src/index.crates.io-1949cf8c6b5b557f/sqlx-sqlite-0.8.6/src/row.rs:43:46:
index out of bounds: the len is 21 but the index is 21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: no rows returned by a query that expected to return at least one row
```

Now `schema_version = 11` and when I run `goose` it does not crash. It was only the very first time after the upgrade.

PR #8720, opened 2026-04-21
feat: goose2 message bubble + action tray

tellaho:
**Category:** improvement
**User Impact:** Goose 2 chat messages now show a bubble-scoped actions tray with a small downward reveal, and copy confirmation stays visible until it clears.

**Problem:** Message actions were tied to the full message row, so the hover target felt broader than the visible bubble. Copy confirmation was also easy to miss because the tray disappeared as soon as hover ended.

**Solution:** This update mostly refreshes the message presentation: the actions tray is visually attached to each bubble, reveals just beneath the message, and uses a subtle motion treatment that keeps the layout steady. It also keeps copy confirmation visible for the full timeout so the feedback reads clearly.

<details>

**ui/goose2/src/features/chat/ui/MessageBubble.tsx**
Reworked the message actions presentation from a row-level hover treatment into a bubble-scoped tray with reserved space below each message. The copy action now uses the shared clipboard hook so the tray stays visible while the copied checkmark is active.

**ui/goose2/src/features/chat/ui/__tests__/MessageBubble.test.tsx**
Expanded the focused UI coverage to assert the muted user bubble shell, the reserved actions tray structure, and the copy confirmation lifecycle. Added clipboard mocking and timer control so the new behavior is exercised deterministically.

</details>

## Reproduction Steps

1. Open Goose 2 and load any chat with at least one assistant response.
2. Hover directly over a message bubble and confirm the actions tray appears for that bubble, animating down beneath it.
3. Move the pointer away and confirm the tray fades out while the message layout stays steady.
4. Click the copy action on a message and confirm the icon switches to a checkmark.
5. Move the pointer away and confirm the tray remains visible while the copied state is active, then disappears once the confirmation timeout ends.

## Demo
image

| hover | button | active | assistant |
| ----------- | ----------- | ----------- | ----------- | 
| image | image |  image | image |

### DIFF
--- a/ui/goose2/scripts/check-file-sizes.mjs
+++ b/ui/goose2/scripts/check-file-sizes.mjs
@@ -65,6 +65,11 @@ const EXCEPTIONS = {
     justification:
       "Voice dictation send/stop guards, attachment handling, and mention/picker coordination still share one chat composer component.",
   },
+  "src/features/chat/ui/MessageBubble.tsx": {
+    limit: 520,
+    justification:
+      "Bubble rendering still owns assistant identity, grouped tool output, attachments, and the inline actions tray pending a later extraction pass.",
+  },
   "src/features/chat/ui/__tests__/ChatInput.test.tsx": {
     limit: 520,
     justification:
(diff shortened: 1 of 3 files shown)


aharvard (approved):
nice

PR #8722, opened 2026-04-21
fix: reuse goose2 vite server on port conflict

tellaho:
**Category:** fix
**User Impact:** Engineers can rerun `just goose2 dev` in the same Goose2 worktree without getting blocked by a stale Vite server on the worktree's assigned port.
**Problem:** `just goose2 dev` hashes the worktree path into a stable Vite port and starts Vite with `--strictPort`. If a previous Goose2 Vite server is still bound to that same port, rerunning the recipe fails immediately with a port-in-use error even though the existing server is already the right one for that worktree.
**Solution:** Route Goose2 dev startup through a small helper script that reuses an existing Vite process when it belongs to the current worktree and otherwise surfaces a clearer next step. The Just recipes also treat the hashed port as a default so engineers can override it with `VITE_PORT` when they need to.

<details>

**ui/goose2/justfile**
Treats the hashed Vite port as a default instead of a hardcoded value. Both dev entry points now delegate startup to the helper script, and the `kill` recipe respects the same `VITE_PORT` override so the workflow stays consistent.

**ui/goose2/scripts/start-dev-vite.sh**
Adds a small startup wrapper for Goose2's Vite dev server. It reuses an existing Vite listener when that process belongs to the current worktree and otherwise exits with a targeted message that points engineers to `just goose2 kill` or a custom `VITE_PORT` override.

</details>

1. In a Goose2 worktree, run `just goose2 dev`.
2. Leave the Vite server running, then rerun `just goose2 dev` from that same worktree.
3. On `main`, the second run fails with `Port <hashed-port> is already in use`.
4. On this branch, the helper script reuses the existing Vite server for that worktree and the dev startup continues.
5. Optionally rerun with `VITE_PORT=<free-port> just goose2 dev` to confirm the new override path also works.


DOsinga (approved):
Merged main and addressed the codex P1 comment: changed the process-detection pattern from `*"vite --port ${VITE_PORT}"*` to `*"--port ${VITE_PORT} --strictPort"*`. This matches exactly what the script passes to vite on the last line, regardless of whether the process shows up as `node /path/to/vite.js` or `vite`. Combined with the cwd check, it's robust and specific.

Also resolved the merge conflict with main's new `kill branch=""` recipe by keeping the worktree-lookup logic and applying the VITE_PORT override in the else branch.

PR #8726, opened 2026-04-21
merge goose-acp crate into goose

jamadeo:
ACP is the future of goose "core", so there isn't really any reason to separate it.


angiejones (approved):
approving for `documentation`

alexhancock (approved):

DOsinga (approved):

PR #8731, opened 2026-04-21
Dedupe and organize skills/sources

jamadeo:


alexhancock (approved):
Looks good for organizing skills in project vs global

Will there be a followup UI one to allow for creation in either a project or global?

baxen (approved):
i like it

jamadeo:
> Looks good for organizing skills in project vs global
> 
> Will there be a followup UI one to allow for creation in either a project or global?

@alexhancock yeah, there should be more UI work around project-discovered skills. I'm not even sure how you access them in the current app.

This PR has grown quite big with skills editing and I'm still not even sure it goes where we want it to, but I'll merge and we can keep iterating.

Issue #8699, opened 2026-04-20
Create AGENTS.md for goose 2.0 architecture guidelines


lifeizhou-ap:
Create an `AGENTS.md` file that coding agents can follow to ensure the new goose 2.0 architecture is adhered to when contributing code.

alexhancock:
complete

PR #8732, opened 2026-04-21
docs: AGENTS.md section on goose2 desktop backend architecture

alexhancock:
Addresses #8699

In the goose2 AGENTS.md this replaces the "Backend Architecture" section with a fuller "Architecture" section that (1) states the single communication path explicitly, (2) walks through the skills-as-sources PR as a canonical worked example, (3) calls out the typed SDK methods as the preferred shape with `extMethod` as an escape hatch, (4) enumerates the narrow cases where `invoke()` is still correct, and (5) lists the specific anti-patterns to avoid. The goal is that an agent picking up a new feature has enough guidance to structure it correctly given the still mixed state of client <-> server interactions in goose2.

### DIFF
--- a/ui/goose2/AGENTS.md
+++ b/ui/goose2/AGENTS.md
@@ -131,16 +131,61 @@ ThemeProvider manages three axes:
 - `tauri-plugin-window-state` persists window size and position.
 - Traffic light offset: `pl-20` (80px) to accommodate macOS window controls.
 
-## Backend Architecture
+## Architecture
 
-All AI communication goes through **ACP (Agent Client Protocol)**:
-- The Tauri app starts a long-lived `goose serve` process and exposes its WebSocket URL.
-- The frontend connects directly to `goose serve` over WebSocket and handles ACP notifications in TypeScript.
+**All frontend ↔ backend communication in goose2 flows through a single path:**
 
-For non AI communication, such as configuration:
-- Use **Tauri commands** (`invoke()` from `@tauri-apps/api/core`) for request/response operations (sessions, personas, skills, projects, etc.).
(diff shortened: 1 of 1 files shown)


ui/goose2/AGENTS.md
```diff
@@ -131,16 +131,61 @@ ThemeProvider manages three axes:
...
+
+For a minimal frontend `api/` wrapper using the typed shape, see `ui/goose2/src/features/providers/api/inventory.ts` — ~30 lines, typed SDK calls, thin adapter. For a fully worked end-to-end feature including OS-keychain handling and progress streaming, see the voice dictation feature ([#8609](https://github.com/block/goose/pull/8609)) and `ui/goose2/src/shared/api/dictation.ts`.
+
+### When `invoke()` is still appropriate
+
+Tauri commands (`invoke()` from `@tauri-apps/api/core`) are reserved for things that genuinely belong to the desktop shell, not to `goose` core. In practice that means:
+
+- `get_goose_serve_url` — bootstrapping the ACP connection.
+- Secret storage owned by the OS keychain (e.g. `save_provider_field`, `delete_provider_config` — note dictation still uses these for writing API keys into the OS keychain, because that's a shell concern).
```
jamadeo, line 177: > Secret storage owned by the OS keychain

do we want to be doing this via ACP+ too?

jamadeo (approved):

ui/goose2/AGENTS.md
alexhancock: we probably can/should yes. i will remove this line.

alexhancock: will do in a followup when I am doing something else to avoid CI cycle wait

PR #8736, opened 2026-04-21
alexhancock/tui-improvements

alexhancock:
## Summary
A grab bag of three TUI improvements

1. Remove some permissioning code which looks to be around from early/experimental commits to this client that weren't being used
2. Fix scrolling which broke somehow, and make resize handling more robust
3. Redo the "expandable tool calls" UX. You can now arrow key between them and hit space to open one.

### Testing
Local usage

### Related Issues
N/A

https://github.com/user-attachments/assets/878e0eaf-525e-4b5b-9c47-c69f09d1f3ee


michaelneale (approved):
Can't wait for "goose" default cli to delegate to this!

alexhancock:
@michaelneale A few more things IMO and then I think we can consider! 'Agree will be great

PR #8747, opened 2026-04-22
feat: add /model slash command to CLI for session model switching

### LINKED ISSUE #8863: Feature Request (Goose CLI): Mid-Session Model Switching (`/model` slash command)

Currently, there is no way to switch the LLM model during an active Goose CLI session. If a user wants to change from `claude-opus-4.6` to `claude-sonnet-4` (or any other model), they must exit the session, set a new `GOOSE_MODEL` environment variable, and restart or resume. This breaks flow and loses conversational momentum.

Real-world use cases where this is painful:
- **Cost optimization mid-task**: You start a session with a powerful model (e.g., `claude-opus-4.6`) for complex planning, then want to switch to a cheaper/faster model (e.g., `claude-sonnet-4`) for straightforward code generation or simple follow-ups — without leaving the session.
- **Model comparison**: When evaluating responses, developers want to ask the same question to different models within one session to compare quality, speed, or token usage.
- **Hitting rate limits**: When a provider rate-limits one model, you want to quickly switch to another without losing your session context.
- **Plan vs Execute**: Goose already supports `GOOSE_PLANNER_PROVIDER` / `GOOSE_PLANNER_MODEL` for `/plan` mode, proving the architecture can handle multi-model sessions. Extending this to ad-hoc switching is a natural evolution.
- **Provider-hosted multi-model access**: Providers like `github_copilot` expose multiple models (Claude, GPT, Gemini) under one authentication. Users should be able to leverage this flexibility without session restarts.

A `/model` slash command in the CLI session with the following behavior:

| Command                    | Behavior                                                                                                                 |
|----------------------------|--------------------------------------------------------------------------------------------------------------------------|
| `/model`                   | Display the currently active provider and model                                                                          |
| `/model list` or `/models` | List all available models from the configured provider (query the provider API if possible, otherwise show known models) |
| `/model <model_name>`      | Switch the active model for the remainder of the session (e.g., `/model claude-sonnet-4`)                                |
| `/model reset`             | Revert to the model defined in `config.yaml` / environment                                                               |
The model switch should:
- Take effect immediately on the next turn
- Preserve full conversation history and context
- Display a confirmation message (e.g., `✓ Switched to claude-sonnet-4 via github_copilot`)
- Be session-scoped only — not modify `config.yaml` or environment variables
- Validate the model name against the provider before switching, and show an error if invalid

1. **Exit and re-export env var**: Current workaround  `export GOOSE_MODEL=claude-sonnet-4 && goose session --resume`. Works but disrupts flow, requires terminal knowledge, and the user must remember exact model names with no discovery mechanism.

2. **Separate profiles/aliases**: Creating shell aliases like `alias goose-sonnet='GOOSE_MODEL=claude-sonnet-4 goose session'`. Functional but doesn't allow switching within a session, and requires manual setup per model.

3. **Config file editing mid-session**: Editing `config.yaml` and hoping it hot-reloads. It doesn't — changes only apply on next session start.

4. **Goose Desktop app**: The Desktop UI already has a model selector dropdown that works mid-session. The CLI should have feature parity.

**Additional context**

- Goose Desktop already supports mid-session model switching via a dropdown, so the backend architecture likely supports this. The CLI just needs a slash command interface to it.
- The `goose local-models list` command exists for local GGUF models, but there's no equivalent for remote provider models. `/model list` would fill this gap for both local and remote.
- This would pair well with the existing `/plan` mode — users could `/model gpt-4.1` for planning, then `/model claude-opus-4.6` for execution, all in one session.
- Related: token usage display per-model after switching would help users track cost across model switches within a session.

---

baxen:
## Summary

Add a `/model` command to the CLI that lets users view or switch the active model within a session without restarting.

### Usage

- `/model` — displays the current model and provider
- `/model <name>` — switches to the given model, keeping the same provider

### Details

- Carries over temperature and toolshim settings when switching models
- Guard rails for ACP providers and providers that manage their own context (e.g. Responses API)
- No-op when already using the requested model
- Tab-completion stub for future model name completion
- Reuses `build_switched_model_config` helper to construct the new `ModelConfig` with canonical limits for the target model

### Bug fix

Fixed ordering of `starts_with` checks in tab completion dispatch — `/model` is now checked before `/mode` so that typing `/model` doesn't incorrectly fall into mode-flag completion (since `"/model".starts_with("/mode")` is true).

### Files changed

- `crates/goose-cli/src/session/completion.rs` — `/model` in command list, completion stub, fixed dispatch ordering
- `crates/goose-cli/src/session/input.rs` — parse `/model` command, help text, tests
- `crates/goose-cli/src/session/mod.rs` — `handle_model`, `build_switched_model_config`, tests
- `crates/goose-cli/Cargo.toml` — added `env-lock` dev dependency for test

Fixes #8863


baxen:
@builderbot review

kyledef (approved):
Just missing a version bump

mohsanabbas (approved):

DOsinga (commented):
Nice work on this — the feature itself looks solid and the guard rails are well thought out.

One architectural request: we have a central slash command dispatcher in `crates/goose/src/agents/execute_commands.rs` where all shared commands live (`/compact`, `/clear`, `/prompts`, `/prompt`, `/skills`, `/doctor`). Could you move the model-switching business logic there?

Specifically:
1. Add `"model"` to the `COMMANDS` array and the `match` in `execute_command()`
2. Add a `handle_model_command(&self, params, session_id)` method on `Agent` that does the current model display (no args) and the provider-switch flow (with args) — including the ACP / manages_own_context guards
3. Move `build_switched_model_config` into the goose crate (e.g. alongside the handler or in `model.rs`)
4. Keep the CLI-specific bits (tab completion, `InputResult::Model` parsing) in `goose-cli`, but have them delegate to `agent.execute_command("/model ...", session_id)` or call the new method directly

This way both CLI and Desktop get the `/model` command through the shared path, and the Desktop won't need a separate REST-only flow for the same operation.

(ref: #8863)

alexhancock (approved):

Issue #8727, opened 2026-04-21
Goose CLI Will Not Launch


aronfoster:
**Describe the bug**

When running goose session from the terminal, the process immediately suspends with tty output errors. The suspension happens consistently on every invocation. It occurs using `zsh` and `bash`.

```
[afoster@xxx]~/path/to/project% goose session

    __( O)>  ● new session · model_name model
   \____)    20260421_3 · /home/afoster/path/to/project
     L L     goose is ready
  ╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌ 0% 0/262k
zsh: suspended (tty output)  ~/opt/goose/bin/goose session

```

---

**To Reproduce**
1. Launch goose CLI from zsh or bash terminal
2. Run goose session                                                                                                                          
3. Observe immediate suspension: zsh: suspended (tty output) (or equivalent for bash)

---

**Expected behavior**
goose session should start without suspension and be able to read user input normally.

---

**Screenshots**
N/A

---

**Please provide the following information**
- **OS & Arch:** RHEL 9 x86_64
- **Interface:** CLI
- **Version:** Latest main (953fe961b) — issue began with commit b7aea3dc9 
- **Extensions enabled:** chatrecall, code_execution, orchestrator, summarize
- **Provider & Model:** [vLLM](https://github.com/vllm-project/vllm), all models

---

**Additional context**
I identified the breaking commit as b7aea3dc9 ("warm shell PATH lookup during session init #8631"). The code change was from LazyLock to spawn_blocking(resolve_login_shell_path), which spawns bash -l -i to resolve the user's PATH.

Proposed fix (verified working):

```
  // File: crates/goose/src/agents/platform_extensions/developer/shell.rs
  // Around line 179-187, change from:
 
      } else {
          std::process::Command::new(&shell)
              .args(["-l", "-i", "-c", "echo $PATH"])
              .stdin(Stdio::null())
              .stdout(Stdio::piped())
              .stderr(Stdio::null())
              .spawn()
              .ok()?
      };

  // To:
      } else {
          // Spawn detached from terminal to prevent affecting parent's process group.
          #[cfg(unix)]
          let mut cmd = std::process::Command::new(&shell);
          #[cfg(unix)]
          {
              use std::os::unix::process::CommandExt;
              cmd.args(["-l", "-i", "-c", "echo $PATH"])
                  .stdin(Stdio::null())
                  .stdout(Stdio::piped())
[...]

PR #8777, opened 2026-04-23
fix(shell): spawn login shell in separate process group on Unix

aronfoster:
## Summary
Prevents the shell command from affecting the parent process group when resolving the login shell path. This fixes issues where the shell spawn could interfere with the parent terminal session.

Uses process_group(0) on Unix to create a new process group, ensuring the child shell is detached from the parent's terminal control.

 ### Testing                                                                                      
  - `cargo test` passes                                                                           
  - `cargo clippy` and `cargo fmt` run cleanly                                                    
  - Verified fix works on RHEL 9 (the original failure environment)  

### Related Issues
Relates to #8727  
Discussion: https://github.com/aaif-goose/goose/issues/8727

### DIFF
--- a/crates/goose/src/agents/platform_extensions/developer/shell.rs
+++ b/crates/goose/src/agents/platform_extensions/developer/shell.rs
@@ -177,13 +177,19 @@ fn resolve_login_shell_path() -> Option<String> {
             .spawn()
             .ok()?
     } else {
-        std::process::Command::new(&shell)
-            .args(["-l", "-i", "-c", "echo $PATH"])
+        let mut cmd = std::process::Command::new(&shell);
+        cmd.args(["-l", "-i", "-c", "echo $PATH"])
             .stdin(Stdio::null())
             .stdout(Stdio::piped())
-            .stderr(Stdio::null())
-            .spawn()
-            .ok()?
+            .stderr(Stdio::null());
+
+        #[cfg(unix)]
+        {
+            use std::os::unix::process::CommandExt;
+            cmd.process_group(0);
+        }
+
+        cmd.spawn().ok()?
     };
 
     let mut stdout = child.stdout.take()?;


aronfoster:
Note that I haven't tested this on any other OS, including Windows.

crates/goose/src/agents/platform_extensions/developer/shell.rs
```diff
@@ -177,13 +177,25 @@ fn resolve_login_shell_path() -> Option<String> {
...
+        let cmd = std::process::Command::new(&shell)
             .args(["-l", "-i", "-c", "echo $PATH"])
             .stdin(Stdio::null())
             .stdout(Stdio::piped())
-            .stderr(Stdio::null())
-            .spawn()
-            .ok()?
+            .stderr(Stdio::null());
+
+        cmd.spawn().ok()?
```
umago, line 192: I think we can simplify the approach here, may I suggest:

```rust
let mut cmd = std::process::Command::new(&shell);                                                                                                                            
cmd.args(["-l", "-i", "-c", "echo $PATH"])                                                                                                                                   
    .stdin(Stdio::null())                                                                                                                                                    
    .stdout(Stdio::piped())                                                                                                                                                  
    .stderr(Stdio::null());                                                                                                                                                  
                                                                                                                                                                             
#[cfg(unix)]                                                                                                                                                                 
{                                                                                                                                                                            
    use std::os::unix::process::CommandExt;                                                                                                                                  
    cmd.process_group(0);                                                                                                                                                    
}                                                                                                                                                                            
                                                                                                                                                                             
cmd.spawn().ok()?  
```

umago (changes requested):

umago (approved):
LGTM, tested locally on my machine (Fedora 43) and it no longer launch in the background:

```bash
🐚 [lucas@fedora goose]$ ./target/debug/goose

    __( O)>  ● new session · claude-acp claude-sonnet-4-6
   \____)    20260423_1 · /home/lucas/Projects/goose
     L L     goose is ready
  ╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌ 0% 0/1.0M
🪿
```

crates/goose/src/agents/platform_extensions/developer/shell.rs
aronfoster: Updated

jamadeo:
this is a good fix, thank you @aronfoster . We merged #8804 which addresses the same issue using process sessions instead of groups.

Issue #8756, opened 2026-04-22
[BUG] Tool use error on macOS (python vs. python3)


tomrannosaurus:
On some tool calls, some models will invoke 'python'. But, on macOS, 'python' does not exist. python 2.7 was removed from macOS on a previous release. 

It would be possible to link python3 install to python on a system level, but that is not recommended as the intended behavior is to have any 'python' calls go unanswered (so that any old scripts calling 'python' and expecting to find 2.7 don't get redirected to 3.xx)

In my opinion, tool calls should always use python3 for this reason. I do not think this was a model issue (although it could be a model/harness interaction issue), because I use gpt5.4 in other harnesses and this has not happened. 

---

**To Reproduce**
A very basic repository (a simple GitHub pages markdown), prompt requested a review of the single page for technical accuracy. 

---

**Screenshots**

<img width="2050" height="476" alt="Image" src="https://github.com/user-attachments/assets/8bd8f5dd-8061-4021-870e-94080c9783c9" />

---

**Please provide the following information**
- **OS & Arch:** macos 24.6.0 aarch64
- **Interface:** UI
- **Version:** 1.31.1
- **Extensions enabled:** chatrecall, summarize, analyze, tom, skills, Extension Manager, developer, todo, code_execution, summon, apps, memory, PDF Reader
- **Provider & Model:** chatgpt_codex – gpt-5.4

---

**Additional context**
The automated bug report generator gave my system OS as above, but this is from my system directly:
```
x@x ~ % system_profiler SPHardwareDataType
Hardware:

    Hardware Overview:

      Model Name: MacBook Pro
      Model Identifier: Mac14,6
      Model Number: Z174000EGLL/A
      Chip: Apple M2 Max
      Total Number of Cores: 12 (8 performance and 4 efficiency)
      Memory: 32 GB
      System Firmware Version: mBoot-18000.101.7
      OS Loader Version: 11881.140.96.701.1

x:~ x$ /usr/bin/sw_vers
ProductName:		macOS
ProductVersion:		15.7.5
BuildVersion:		24G624

```

DOsinga:
Thanks for reporting this! This is ultimately a model behavior issue — whether the model chooses to invoke `python` vs `python3` depends on how well it follows instructions, and that varies across providers and model versions.

[...]

PR #8784, opened 2026-04-23
fix: use python3 in developer extension instructions for macOS/Linux compatibility

treebird7:
## Summary

Modern macOS (12+) removed the `python` binary — only `python3` is available.
Many Linux distributions follow the same convention. When the model uses the
developer extension to run Python scripts, it may invoke `python` directly,
which fails on these systems.

This adds a clear instruction to the non-Windows developer extension system
prompt to always use `python3` instead of `python`.

### Testing

- No existing tests cover the content of `developer_instructions()` — tests
  in mod.rs are functional tool-call tests (shell, write, edit)
- The change is a string addition inside the existing `cfg!(windows)` else
  branch, so Windows behaviour is unaffected

### Related Issues

Relates to #8756

### DIFF
--- a/crates/goose/src/agents/platform_extensions/developer/mod.rs
+++ b/crates/goose/src/agents/platform_extensions/developer/mod.rs
@@ -59,7 +59,7 @@ fn developer_instructions() -> &'static str {
             Use write and edit to efficiently make changes. Test and verify as appropriate.
 
             When running Python scripts or commands, always use `python3` instead of `python`.
-            The `python` command is not available on modern macOS and many Linux distributions.
+
         "}
     }
 }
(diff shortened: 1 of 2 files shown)


crates/goose/src/agents/platform_extensions/developer/mod.rs
```diff
@@ -57,6 +57,9 @@ fn developer_instructions() -> &'static str {
             and file sizes. When you need to search, prefer rg which correctly respects gitignored
             content. Then use cat or sed to gather the context you need, always reading before editing.
             Use write and edit to efficiently make changes. Test and verify as appropriate.
+
+            When running Python scripts or commands, always use `python3` instead of `python`.
+            The `python` command is not available on modern macOS and many Linux distributions.
```
jh-block: I think the second sentence can be removed; it doesn't shape model behaviour and wastes tokens.

jh-block (changes requested):

crates/goose/src/agents/platform_extensions/developer/mod.rs
treebird7: Smart thinking, deleted that sentence.

jh-block (approved):

jh-block:
looks good but there's a test failure, looks like whitespace in a snapshot

PR #8787, opened 2026-04-23
refactor: make ACP server smaller

alexhancock:
## Summary
* Code size reduction for the ACP server
* I wasn't sure if anyone was using all the `perf` logs so I ripped them out, but feel free to let me know if that is wrong @baxen @codefromthecrypt 

### Testing
Tests and using goose locally

### Related Issues
N/A

N/A

### DIFF
--- a/crates/goose/src/acp/server.rs
+++ b/crates/goose/src/acp/server.rs
@@ -78,6 +78,34 @@ pub type AcpProviderFactory = Arc<
         + Sync,
 >;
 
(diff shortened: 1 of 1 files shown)


codefromthecrypt (approved):
don't recall about the perf stuff but apologies if that was me.

the map_err stuff was for reversability of the errors from acp client to agent, but if it is keen without it or internal_err works well enough sounds fine to me.

the acp server indeed has become massive and needs to be both pared down and probably broken up. thanks for the progress.

PR #8793, opened 2026-04-24
fix(goose): exclude preprompt from session title generation

matt2e:
## Issue
When creating a session in a project, the additional preprompt text we provide to the agent to generate a session name meant often the session names were about the preprompt, not the actual user's message:
Screenshot 2026-04-24 at 1 49 27 pm

## Summary
- Filters out assistant-audience (preprompt) content blocks from user messages when generating session titles, so titles reflect what the user actually said
- Extracts preprompt context separately and includes it as background context in the title generation prompt, explicitly instructing the model not to base the title on it
- Updates the simple session description fallback to strip the new background context section

## Test plan
- [x] Verify session titles no longer incorporate preprompt/system instructions
- [x] Verify preprompt context is still available to the model for understanding but not title generation
- [x] Verify fallback simple session description still works correctly

### DIFF
--- a/crates/goose/src/providers/base.rs
+++ b/crates/goose/src/providers/base.rs
@@ -731,14 +731,42 @@ pub trait Provider: Send + Sync {
         ))
     }
 
-    /// Returns the first 3 user messages as strings for session naming
+    /// Returns the first 3 user messages as strings for session naming,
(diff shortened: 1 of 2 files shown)


jamadeo:
Why is the pre-prompt in a user message and not part of the system prompt?

morgmart (approved):

PR #8795, opened 2026-04-24
fix(providers): honor dynamic_models: false in declarative provider configs

### LINKED ISSUE #8317: Bug: dynamic_models field in custom provider config is ignored

guptamihir418:
# fix(providers): honor `dynamic_models: false` in declarative provider configs

Fixes #8317.

## What was broken

`DeclarativeProviderConfig.dynamic_models` (`crates/goose/src/config/declarative_providers.rs`) is defined as `Option<bool>` but was ignored by `OpenAiProvider` and `AnthropicProvider`. Both providers always tried `/v1/models` first and only fell back to the static `models` list on a 404. Users who set `"dynamic_models": false` still saw the provider's full dynamic list.

## What changed

Added a 3-state `dynamic_models: Option<bool>` field to `OpenAiProvider` and `AnthropicProvider`, threaded it through `from_custom_config`, and short-circuited `fetch_supported_models`:

| `dynamic_models` | `models` | Behavior |
|---|---|---|
| `Some(false)` | non-empty | Return static list, no API call *(the fix)* |
| `Some(false)` | empty | Construction error |
| `Some(true)` or `None` | any | Try API, fall back to static on 404 *(unchanged)* |

Config validation now runs at the top of `from_custom_config` — before secret lookup, URL parsing, or API-client construction — so misconfigurations fail fast and don't first surface as downstream errors.

## Why not the reporter's suggested patch?

The issue's suggested fix removes the static-list fallback entirely when `dynamic_models` is anything other than `Some(false)`. That would regress 10 of the 15 built-in declarative providers (cerebras, deepseek, groq, inception, lmstudio, minimax, mistral, moonshot, novita, ovhcloud, tensorix) — they ship with an unset `dynamic_models` and a non-empty `models` list and depend on today's 404 fallback. This PR preserves that behavior for the unset/`Some(true)` cases.

## Tests

- `fetch_supported_models_static_only_skips_api` — wiremock returning 500 for any request; if the fix is broken and the API is hit, the test fails loudly.
- `fetch_supported_models_dynamic_true_hits_api` — mock returns models at `/v1/models`; asserts the API list overrides the static list.
- `fetch_supported_models_none_falls_back_on_404` — regression test: proves the legacy fallback still works for `dynamic_models: None`, which is what 10 of the 15 built-in declarative providers rely on.
- `from_custom_config_rejects_static_only_without_models` — asserts the contradictory-config error with message check.
- `from_custom_config_preserves_dynamic_flag` (OpenAI) — asserts the field is threaded through on construction.

New test module added to `anthropic.rs` (previously had no tests); mirrored cases there.

## Out of scope

- `OllamaProvider` has no `custom_models` field and discovers models via `/api/tags` against a live daemon — `dynamic_models: false` semantics don't map to it.
- `supports_inventory_refresh` wiring (`declarative_providers.rs:468,487,506`, gating a separate 24h staleness/refresh UX) is intentionally untouched.
- `skip_canonical_filtering` is a separate axis (which IDs to keep from the returned list vs whether to fetch at all) and isn't conflated.

## Verification

- `cargo check -p goose --lib` passes for the library changes.
- The new tests follow the wiremock pattern established in `providers/kimicode.rs` and `providers/chatgpt_codex.rs`.
- I was unable to run `cargo test` or an end-to-end repro locally in my development environment due to an unrelated filesystem issue. The full test suite will run via CI on this PR; reviewers, please confirm the tests pass there before merging.


jh-block:
Thanks for the contribution. Overall this looks good, but checks are failing: there are formatting issues (please run `cargo fmt`) and a build error to resolve.

DOsinga (approved):
LGTM — the fix is correct and well-scoped. I pushed a commit to fix the build (missing struct fields in test helpers) and trimmed the tests down to the ones that actually test the new behavior rather than re-asserting pre-existing paths.

DOsinga:
@guptamihir418 Nice fix! One thing needed before merge: your commit is missing the DCO sign-off. Please amend it with:

```
git commit --amend -s
git push --force-with-lease
```

This adds the `Signed-off-by:` trailer required by the project's DCO check.

PR #8804, opened 2026-04-24
fix(shell): prevent login-shell PATH probe from suspending goose on startup

maxamillion:
## Summary
resolve_login_shell_path() spawns `bash -l -i -c "echo $PATH"` to pick up the user's full PATH. The -i flag triggers bash's interactive job-control setup, which calls TIOCSPGRP to steal the terminal foreground from goose. While in the background, any terminal read in goose's startup path causes the kernel to send SIGTTIN to the entire process group, suspending goose.

Fix: use process-wrap's ProcessSession wrapper to spawn the child in a new session (via setsid). This detaches the child from the controlling terminal so TIOCSPGRP cannot succeed. PATH resolution is unaffected; -i is preserved so that ~/.bashrc is still sourced for users whose ~/.bash_profile does not chain to it.

### Testing
manual testing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4391,6 +4391,7 @@ dependencies = [
  "pem",
  "pkcs1",
  "pkcs8",
+ "process-wrap",
  "pulldown-cmark",
  "rand 0.8.5",
  "rayon",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4525,6 +4525,7 @@ dependencies = [
  "indoc",
  "lopdf",
  "once_cell",
+ "process-wrap",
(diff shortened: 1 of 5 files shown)


aronfoster:
See https://github.com/aaif-goose/goose/pull/8777 for another PR attempting to address this issue.

jamadeo (approved):
This seems to be the most robust fix to this issue. I'm not entirely sure how this broke, but it definitely broke, and process session isolation looks like the best way to ensure it doesn't break again.

michaelneale:
nice - yeah this seems better. There was a change which made it block on it, was what I noted (just in last few days)

Issue #6915, opened 2026-02-03
Goose consumes an enormous number of GitHub Copilot premium requests for simple things


vladiliescu:
Invoking goose with a simple request (in my case `thoroughly analyze this project from a security standpoint. create a SECTODO.md with your findings.`) will use about 16 (**16!**) premium requests before it comes back to the user.

Repro:
- Clone smolagents (https://github.com/huggingface/smolagents.git)
- Start goose, let it loose with the command above (`thoroughly analyze this project from a security standpoint. create a SECTODO.md with your findings.`)
- Watch the premium requests get drained (I was at 4.2% before goose, now I'm at 5.3%. I'm a Pro+ subscriber, so I have 1500 premium requests per month => 1500*(5.3-4.2)/100=**16,5** premium requests for a simple task)

By comparison, for the exact same request, Copilot CLI uses `gpt-5.2-codex           1.3m in, 22.2k out, 880.3k cached (Est. 1 Premium request)`. I'd expect goose to do the same.

I've attached the diagnostics log.

[diagnostics_20260203_3.zip](https://github.com/user-attachments/files/25048893/diagnostics_20260203_3.zip)

OS/App Details
- **OS & Arch:** MacOS Sequoia 15.7.3 (24G419)
- **Interface:** CLI
- **Version:** 1.14.2
- **Extensions enabled:** - Extension Manager, chatrecall, developer, todo
- **Provider & Model:** github_copilot, model: gpt-5.2

As a side note, this is not an isolated incident, I had noticed something similar last year (severe drainage of premium requests), and I subsequently switched to another agentic cli hoping this would get fixed in the meantime. It hasn't so here I am raising a bug about it :).

alterphp:
I get the same behaviour with Goose Desktop.

OS/App Details

    OS & Arch: Ubuntu 24.04.3
    Interface: Desktop
    Version: 1.23.0
    Extensions enabled: Extension Manager, Apps, Developer, Skills, Todo
    Provider & Model: github_copilot, model: sonnet 4.5

Even asking to Goose to compress the context (with `/compress` command) seems to consume premium requests. I'm pretty sure Github Copilot is not a convenient provider for tools like Goose, but consuming so many premium requests seems to be a misuse of Github Copilot APIs ?

vladiliescu:
[...]

PR #8809, opened 2026-04-24
fix(providers): set X-Initiator header on GitHub Copilot requests

itscooper:
## Summary

GitHub Copilot's API uses the `X-Initiator` header to distinguish between requests initiated directly by the user and internal agent calls (e.g. context compaction, session title generation). Setting it correctly ensures usage is attributed and metered appropriately on GitHub's side. In my personal tests this prevents over-billing (as per issue #6915 ).

## How it works

A `tokio::task_local!` flag (`IS_AGENT_CALL`) is set to `true` inside `complete()`, which is the entry point for all agent-internal calls. `stream()` then reads this flag -- along with checking whether the last message is a tool response -- to determine the correct header value:

- **`X-Initiator: user`** -- `stream()` calls where `IS_AGENT_CALL` is unset and the last message is not a tool response
- **`X-Initiator: agent`** -- calls routed through `complete()` (compaction, title generation) or any `stream()` call continuing a tool-use loop

The task-local scope prevents the flag from leaking across concurrent requests on the same provider instance.

## Changes

- `crates/goose/src/providers/githubcopilot.rs` only -- no other files changed

## Tests

Two new unit tests:

- `is_agent_call_task_local_scoped` -- verifies the task-local is absent before the scope, visible inside, and absent again after (the isolation property `complete()` relies on)
- `tool_response_variant_matches` -- canary that confirms `MessageContent::ToolResponse` is the correct variant; fails if the enum is renamed or restructured


itscooper:
I've been running this patch locally for a couple of weeks (in Goose Desktop) and it seems to be working as expected for me.

DOsinga (approved):
Good fix for a real issue (#6915). The task-local approach is a clean way to thread the user/agent distinction without changing the Provider trait.

I merged main (resolved a conflict in the imports — `DEFAULT_PROVIDER_TIMEOUT_SECS` was added upstream) and made one cleanup commit:

- Removed `is_agent_call_task_local_scoped` test — it tests `tokio::task_local!` semantics, not this PR's logic
- Removed `tool_response_variant_matches` test — tautologically asserts the constructor returns its own variant
- Removed a redundant comment on the `complete()` override

PR #8810, opened 2026-04-24
fix: return 400 instead of panicking on invalid CSP header value

r0x0d:
## Summary

- User-controlled CSP strings from the request body (`body.csp`) could contain invalid HTTP header characters, causing `csp.parse().unwrap()` to panic the `serve_guest_html` request handler
- Replace the unwrap with a `match` that returns `400 Bad Request` with a descriptive message on invalid input
- Also replace the static `"strict-origin".parse().unwrap()` with `HeaderValue::from_static` which is validated at compile time and has zero runtime cost

## Changes

`crates/goose-server/src/routes/mcp_app_proxy.rs`
- `"strict-origin".parse().unwrap()` → `header::HeaderValue::from_static("strict-origin")`
- `csp.parse().unwrap()` → `match csp.parse::<header::HeaderValue>()` returning `400 Bad Request` on error

### DIFF
--- a/crates/goose-server/src/routes/mcp_app_proxy.rs
+++ b/crates/goose-server/src/routes/mcp_app_proxy.rs
@@ -255,10 +255,21 @@ async fn serve_guest_html(
             // no-referrer would cause 401s from SDK servers.
             headers.insert(
                 header::HeaderName::from_static("referrer-policy"),
-                "strict-origin".parse().unwrap(),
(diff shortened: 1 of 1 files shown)


DOsinga (approved):
Clean, minimal fix that prevents a panic on user-controlled input. The use of `HeaderValue::from_static` for the compile-time-known value and graceful error handling for the dynamic CSP value are both correct. LGTM.

PR #8811, opened 2026-04-24
chore: move insta to dev-dependencies

r0x0d:
## Summary

- `insta` is a snapshot testing library used exclusively in `#[cfg(test)]` blocks inside `crates/goose`
- It was incorrectly listed under `[dependencies]`, causing it to be compiled into the release binary
- Moving it to `[dev-dependencies]` ensures it is only compiled during testing

## Changes

`crates/goose/Cargo.toml`
- Removed `insta = "1.43.2"` from `[dependencies]`
- Added `insta = "1.43.2"` to `[dev-dependencies]`


DOsinga (approved):
Correct fix — `insta` is only used in `#[cfg(test)]` blocks and belongs in `[dev-dependencies]`. Merged `origin/main` and resolved the version conflict (bumped to `1.47.2` to match current main).

PR #8813, opened 2026-04-24
fix: gracefully handle corrupted permission config instead of panicking

r0x0d:
## Summary

- A corrupted or unreadable `permission.yaml` caused `PermissionManager::new()` to `panic!` at startup, crashing the entire process with no recovery path
- Replace the panic with structured `tracing::error!` logging and a fallback to an empty permission map, allowing the process to start normally
- Handles both parse errors (invalid YAML) and read errors (permissions, I/O) gracefully
- The error message instructs the user to fix or remove the file to suppress the warning

## Changes

`crates/goose/src/config/permission.rs`
- Replaced `panic!` on corrupted config with error logging + empty `HashMap` fallback
- Also handles the file read failure case (previously used `.expect()`)
- Updated `test_corrupted_permission_file_panics` → `test_corrupted_permission_file_falls_back_to_empty` to assert the new graceful behaviour

### DIFF
--- a/crates/goose/src/config/permission.rs
+++ b/crates/goose/src/config/permission.rs
@@ -45,19 +45,28 @@ impl PermissionManager {
     pub fn new(config_dir: PathBuf) -> Self {
         let permission_path = config_dir.join(PERMISSION_FILE);
         let permission_map = if permission_path.exists() {
-            let file_contents =
-                fs::read_to_string(&permission_path).expect("Failed to read permission.yaml");
-            serde_yaml::from_str(&file_contents).unwrap_or_else(|e| {
-                tracing::error!(
-                    "Failed to parse {}: {}. Refusing to start with corrupted permission config.",
-                    permission_path.display(),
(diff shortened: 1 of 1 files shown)


DOsinga:
Thanks for the PR! Closing this because the Codex review raised a real issue 17 days ago that hasn't been addressed: when `read_to_string` fails, the manager falls back to an empty map but `self.config_path` still points at the unreadable file — so later calls to `update_smart_approve_permission` will hit `fs::write(...).expect(...)` and panic at runtime. The fix trades a startup panic for a deferred one.

Additionally, silently starting with empty permissions on a corrupted config could be a security concern — previously-set restrictions would be dropped without the user knowing.

Please see our guidelines on addressing AI code review comments: https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#ai-code-reviews

Feel free to reopen once the Codex feedback is addressed and the deferred-panic path is handled.

PR #8814, opened 2026-04-24
fix: return error instead of panicking on llama backend init failure

r0x0d:
## Summary

- `InferenceRuntime::get_or_init()` would `panic!` if `LlamaBackend::init()` failed (e.g. missing GPU, unsupported hardware, bad native library), giving callers no recovery path and crashing the entire process
- Change the return type from `Arc<Self>` to `Result<Arc<Self>>` so the failure propagates naturally up the call stack
- Both callers already return `Result`, so the fix is a minimal `?` propagation at each call site

## Changes

`crates/goose/src/providers/local_inference.rs`
- `get_or_init()` return type changed to `Result<Arc<Self>>`
- `panic!` on backend init failure replaced with `Err(anyhow::anyhow!(...))`
- All `return runtime` / `runtime` paths wrapped in `Ok(...)`
- `from_env()` caller updated with `?`

`crates/goose-server/src/state.rs`
- `AppState::new()` caller updated with `?` (already returns `anyhow::Result`)

### DIFF
--- a/crates/goose-server/src/state.rs
+++ b/crates/goose-server/src/state.rs
@@ -48,7 +48,7 @@ impl AppState {
             gateway_manager,
             extension_loading_tasks: Arc::new(Mutex::new(HashMap::new())),
             #[cfg(feature = "local-inference")]
-            inference_runtime: InferenceRuntime::get_or_init(),
+            inference_runtime: InferenceRuntime::get_or_init()?,
             session_buses: Arc::new(Mutex::new(HashMap::new())),
         }))
     }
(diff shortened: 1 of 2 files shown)


DOsinga:
Hey @r0x0d — apologies for the slow reaction on this one! Your fix was spot-on, but in the meantime the same change landed via #8656 (and was further refactored in #9137), so this PR is now superseded. Thanks for taking the time to submit it, and sorry again for not getting to it sooner. 🙏

PR #8816, opened 2026-04-24
chore: introduce DEFAULT_PROVIDER_TIMEOUT_SECS constant for providers

r0x0d:
## Summary

- All 10+ provider files independently hardcoded `600` seconds as their HTTP timeout with no shared reference, creating drift risk when the value needs to change
- Define a single `DEFAULT_PROVIDER_TIMEOUT_SECS = 600` constant in `providers/base.rs` with a doc comment explaining the rationale (long-running inference can take several minutes)
- Replace all inline literals across every affected provider

## Affected files

- `providers/base.rs` — new constant defined here
- `providers/api_client.rs`
- `providers/databricks.rs`
- `providers/gcpvertexai.rs`
- `providers/gemini_oauth.rs`
- `providers/githubcopilot.rs`
- `providers/kimicode.rs`
- `providers/litellm.rs`
- `providers/ollama.rs`
- `providers/openai.rs`
- `providers/toolshim.rs`


DOsinga (approved):
Thanks for the contribution! The idea of centralizing the timeout constant is good.

I pushed a small fixup commit that removes the redundant local const aliases in `databricks.rs`, `gcpvertexai.rs`, and `gemini_oauth.rs` — those were just re-exporting the global constant under a different name, which adds indirection without benefit. Now those files use `DEFAULT_PROVIDER_TIMEOUT_SECS` directly.

Note: `OLLAMA_TIMEOUT` is kept as-is since it's `pub` and used as a named default throughout the Ollama provider's config/stream timeout logic.

PR #8817, opened 2026-04-24
refactor(logging): consolidate logging setup into shared helper in goose crate

r0x0d:
## Summary

- Extracts the duplicated file-appender / env-filter / layer-composition logic shared between `goose-cli` and `goose-server` into a new `build_logging_subscriber()` function in `crates/goose/src/logging.rs`.
- Introduces `LoggingConfig` struct with fields: `component`, `name`, `extra_directives`, and `console` (the only behavioural difference between CLI and server).
- Adds `tracing-appender` and `tracing-subscriber` (with `env-filter`, `fmt`, `json`, `time` features) to `goose`'s dependencies so the shared code compiles.
- Reduces duplication by ~80 lines across the two logging modules.
- All existing logging tests continue to pass.


DOsinga (approved):
Pushed a fix commit that addresses two issues:

1. **Codex review feedback**: `extra_directives` are now only applied in the fallback path (when `RUST_LOG` is not set). Previously they were unconditionally appended, which meant `RUST_LOG=error` couldn't fully suppress component-specific targets.

2. **Server log format preservation**: Added a `json` field to `LoggingConfig`. The server's file layer originally used plain text (`.with_file(true)`), not JSON. The shared code now respects this — CLI gets `json: true`, server gets `json: false`.

Also merged origin/main and resolved the Cargo.toml conflict.

PR #8837, opened 2026-04-24
fix(ci): prevent flaky smoke test timeouts from failing the build

michaelneale:
flaky providers (e.g. `nvidia/nemotron-3-nano-30b-a3b`) sometimes hang or respond slowly, hitting vitest's 60s `testTimeout`. Vitest timeout errors **abort the test externally** and are not catchable by try/catch. The flaky test wrapper in `test_providers_lib.ts` uses try/catch to suppress failures from flaky models, but this doesn't work for timeouts.

## Fix

1. **Add a 55s internal timeout to `runGoose()`** — kills the child process and rejects with a normal `Error` that the try/catch can handle
2. **Give flaky tests a 90s vitest timeout** — ensures the internal timeout (55s) always fires first, gets caught by the flaky handler, and the test passes as an allowed failure

### DIFF
--- a/ui/desktop/tests/integration/test_providers_lib.ts
+++ b/ui/desktop/tests/integration/test_providers_lib.ts
@@ -313,13 +313,19 @@ function registerTests(label: string, cases: TestCase[], fn: ProviderTestFn): vo
   }
 
   if (flaky.length > 0) {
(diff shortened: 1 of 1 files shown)


DOsinga (approved):
Not sure why we need the settled flag, but probably doesn't hurt

PR #8839, opened 2026-04-25
fix(oidc-proxy): enforce exp independently of MAX_TOKEN_AGE_SECONDS (#8832)

### LINKED ISSUE #8832: [Security] High: OIDC proxy MAX_TOKEN_AGE_SECONDS bypasses exp check (expired token replay)
## Severity
**High** (Confirmed) — deterministic authentication bypass when `MAX_TOKEN_AGE_SECONDS` is configured.

## CWE
- CWE-613 (Insufficient Session Expiration)
- CWE-287 (Improper Authentication)
- OWASP A07:2021

## Affected code
- `oidc-proxy/src/index.js:193-200`

```js
if (env.MAX_TOKEN_AGE_SECONDS && payload.iat) {
  const age = Date.now() / 1000 - payload.iat;
  if (age > parseInt(env.MAX_TOKEN_AGE_SECONDS, 10))
    return { valid: false, reason: "Token too old" };
} else if (!payload.exp || payload.exp < Date.now() / 1000) {  // skipped when MAX_TOKEN_AGE_SECONDS is set
  return { valid: false, reason: "Token expired" };
}
```

## Root cause
The `exp` validation is inside an `else if` that only runs when `MAX_TOKEN_AGE_SECONDS` is **not** configured. When the operator sets `MAX_TOKEN_AGE_SECONDS` (intending to add a stricter upper bound), the `exp` branch is never taken — so validly-signed but **expired** tokens are accepted until `iat + MAX_TOKEN_AGE_SECONDS`.

Example: GitHub OIDC default `exp = iat + 300`; operator sets `MAX_TOKEN_AGE_SECONDS = 3600` ("reject tokens older than an hour"). A token leaked at t=100s is accepted at t=400s — past `exp`.

## Reproduction steps
```bash
# $LEAKED_TOKEN issued at iat=T0, exp=T0+300. Now T0+400.
# Worker configured with MAX_TOKEN_AGE_SECONDS=3600.
curl -H "Authorization: Bearer $LEAKED_TOKEN" \
     https://proxy.example/api/...
# => 200 OK (exp check skipped entirely)
```

## Suggested fix
Split the two checks so both windows must pass (independent `if`s, not an `else if`):

```js
if (!payload.exp || payload.exp < Date.now() / 1000)
  return { valid: false, reason: "Token expired" };
if (env.MAX_TOKEN_AGE_SECONDS && payload.iat) {
  const age = Date.now() / 1000 - payload.iat;
  if (age > parseInt(env.MAX_TOKEN_AGE_SECONDS, 10))
    return { valid: false, reason: "Token too old" };
}
```

This is a one-line fix but closes a deterministic auth-bypass. Consider pairing with an integration test that asserts both gates apply simultaneously.

## References
- CWE-613, CWE-287
- OWASP A07:2021

_Filed from an out-of-band security audit against `main` @ `eb60770c8` (snapshot 2026-04-24)._

parasol-aser:
## Summary

Closes #8832 — `[Security] High: OIDC proxy MAX_TOKEN_AGE_SECONDS bypasses exp check (expired token replay)` (CWE-613, CWE-287, OWASP A07:2021).

`oidc-proxy/src/index.js` previously wired the `exp` check and the `iat`-based age check with `if … else if`, so when `MAX_TOKEN_AGE_SECONDS` was configured the `exp` branch was never executed. A signed-but-expired token could be replayed up to `iat + MAX_TOKEN_AGE_SECONDS`. Concrete attack from the issue: GitHub OIDC issues `exp = iat + 300` (5 min); operator sets `MAX_TOKEN_AGE_SECONDS = 3600` ("reject anything older than an hour"); a token leaked at t=100s replays at t=400s — past `exp` — and is accepted.

The fix splits the two checks into independent guards: `exp` is now enforced unconditionally, and `MAX_TOKEN_AGE_SECONDS` is a *stricter upper bound layered on top*, never a replacement. This matches the fix proposed in the issue body.

## Changes

- **`oidc-proxy/src/index.js`** — split the `if/else if` into two independent `if` blocks; `exp` checked first (cheap, IdP-authoritative), then the operator's age cap. Reason strings (`"Token expired"`, `"Token too old"`) preserved exactly so operator alerts and existing tests do not churn. Switched the age subtraction to `Math.floor(Date.now() / 1000) - payload.iat` so both operands are integers (cosmetic; behavior unchanged at second granularity).
- **`oidc-proxy/test/index.test.js`** — inverted the existing `"accepts recently-expired token within MAX_TOKEN_AGE_SECONDS"` test (which codified the buggy behavior) into `"rejects expired token even when MAX_TOKEN_AGE_SECONDS is set"`. Added a regression test asserting the age cap fires independently of `exp` (`iat` past cap, `exp` still valid → `"Token too old"`), and a defensive test confirming `exp` is enforced when the cap is unset.
- **`oidc-proxy/test/token-gates.test.js`** *(new)* — comprehensive dual-gate suite (32 cases): happy paths, `exp` gate unconditional, age-cap gate independence, both-fail ordering, the canonical issue-#8832 replay scenario, reason-string stability, and the PLAN.md §5 strategy matrix reproduced row-for-row.
- **`oidc-proxy/README.md`** — rewrote the `Token age vs expiry` section to describe the new dual-gate semantics (both gates must pass; `MAX_TOKEN_AGE_SECONDS` cannot extend a token past its `exp`). Also tightened the env-var table entry to make the layered relationship explicit.

## Breaking change for operators

Operators who relied on `MAX_TOKEN_AGE_SECONDS` to *extend* usable token lifetime past the IdP's `exp` (e.g. long-running CI jobs against GitHub OIDC's 5-min `exp`) will see those tokens rejected. The README now recommends refreshing the OIDC token rather than extending it via the age cap. The default `MAX_TOKEN_AGE_SECONDS = 1200` (20 min) in `wrangler.toml` is already stricter than the 5-min `exp`, so default deployments are unaffected.

## Out of scope

The following findings from the same audit are intentionally not addressed here (tracked separately):
- F-7: JWKS refetch amplification on unknown `kid`
- F-8: claims-validated-before-signature ordering
- F-9: DO `timestamps` not persisted

## Test plan

- [x] `npm test` in `oidc-proxy/` — **42/42 passing** (10 in `index.test.js`, 32 in the new `token-gates.test.js`).
- [x] Verified the canonical replay scenario from the issue (`exp = iat + 300`, replay at t=400s with `MAX_TOKEN_AGE_SECONDS=3600`) is now rejected as `"Token expired"`.
- [x] Verified age cap still fires independently when `exp` is comfortably valid (`iat` past cap → `"Token too old"`).
- [x] Verified happy path with cap configured still proxies a 200.
- [x] Verified missing `exp` is now rejected even with a fresh `iat` and cap set (deliberate tightening, called out in PLAN §4).
- [x] Verified reason strings (`"Token expired"`, `"Token too old"`) and 401 status are unchanged.


DOsinga (approved):
The core fix is correct and valuable — thanks for catching this. We stripped the 530-line `token-gates.test.js` file because it was AI-generated bloat (32 test cases where the 3 already in `index.test.js` cover the fix). We also merged main into the branch.

For future contributions: please don't send up AI-generated fixes for AI-found issues with this sort of test bloat. Strip it down to what it has to be before opening the PR. See [Quick responsible AI tips](https://github.com/block/goose/blob/main/CONTRIBUTING.md#quick-responsible-ai-tips).

DOsinga:
One thing: your commit is missing the DCO `Signed-off-by` line. Please amend it with `git commit --amend --signoff` and force-push so the DCO check passes.

parasol-aser:
Done — amended commit `a803adb` with `Signed-off-by` and force-pushed. Tree is unchanged from the prior tip; only my commit's metadata moved (also reset the author from an internal hostname to my real email so the trailer matches).

Heads-up: rewriting my commit's SHA meant your two follow-up commits got re-grafted onto the new base, so they have new SHAs too — content/author preserved.

PR #8843, opened 2026-04-25
fix(bedrock): handle ReasoningContent blocks gracefully

### LINKED ISSUE #8751: Goose bails if it receives a Bedrock ContentBlock of type reasoningContent
## Description ##

The function _from_bedrock_content_block()_ in **crates/goose/src/providers/formats/bedrock.rs** does not handle content blocks of type _bedrock::ContentBlock::ReasoningContent_ and bails instead. Unfortunately various models do return these blocks, making them unusable with Goose.

## To Reproduce ##

1. Configure Goose for use with Amazon Bedrock.
2. Select a model which returns reasoningContent blocks, for example openai.gpt-oss-safeguard-120b.
3. Send a prompt to the model.
4. Instead of the response, an error is displayed: "Unsupported content block type from Bedrock".

NOTE: The Anthropic models do not seem to return these reasoningContent blocks, so the problem is never seen with them. Returning reasoningContent blocks seems to be a very common feature of other models though, so it's definitely a problem.

## Expected Behavior ##

I would expect all content blocks listed in the API reference (https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ContentBlock.html) to be either:

1. Correctly handled
2. Justifiably ignored
3. Discarded with an informative warning that the content block type is not supported. The warning should state what the type of the content block is and not just that it is unsupported.

Since the model response can include multiple content blocks, I am not sure that it is correct to bail just because one of them is unsupported. The content of the others might still be supported and desired.

## Temporary Fix ##

I have a hacky fix which consists of amending the filtering on line 304 of crates/goose/src/providers/formats/bedrock.rs so that it filters out bedrock::ContentBlock::ReasoningContent as well. I'm happy to raise a pull request if desired but I don't think this is really proper solution.

fresh3nough:
## Summary

Bedrock's `ContentBlock` enum has a `ReasoningContent` variant that some models (e.g. `openai.gpt-oss-safeguard-120b` and other gpt-oss models on Bedrock) include in their responses. Goose's current `from_bedrock_content_block` falls into the `_ => bail!("Unsupported content block type from Bedrock")` arm for any block it doesn't explicitly handle, so the whole response is aborted as soon as a reasoning block appears.

This change:

- Maps `ContentBlock::ReasoningContent(ReasoningText)` to `MessageContent::Thinking` (preserving the optional signature) so reasoning text round-trips with the rest of the assistant message.
- Maps `ContentBlock::ReasoningContent(RedactedContent(blob))` to `MessageContent::RedactedThinking` (base64-encoded), matching how other providers surface encrypted reasoning.
- Updates `from_bedrock_content_block`'s signature to `Result<Option<MessageContent>>` and filters `None` values in `from_bedrock_message`.

Fixes aaif-goose/goose#8751.

## Tests

New unit tests in `crates/goose/src/providers/formats/bedrock.rs`:

- `test_from_bedrock_content_block_reasoning_text` – `ReasoningText` with text + signature → `Thinking`.
- `test_from_bedrock_content_block_reasoning_text_without_signature` – signature is optional in the SDK; produces `Thinking` with empty signature.
- `test_from_bedrock_content_block_reasoning_redacted_content` – `RedactedContent(Blob)` → `RedactedThinking` with base64 payload.
- `test_from_bedrock_message_includes_reasoning_content` – regression test for #8751: a mixed message (`ReasoningContent` + `Text`) now converts successfully and preserves both blocks instead of erroring out.

```
cargo test -p goose --lib providers::formats::bedrock
running 15 tests
...
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 1099 filtered out; finished in 0.01s
```

`cargo fmt` and `cargo clippy -p goose --all-targets -- -D warnings` both pass.

---

[Conversation](https://app.warp.dev/conversation/c7198a26-1801-46b0-b2be-cdc3665f89c6)


DOsinga (approved):
Fixes a real bug (#8751) — Bedrock reasoning models crash because `ReasoningContent` blocks weren't handled. The approach is correct: map them to/from `Thinking`/`RedactedThinking` so they round-trip properly.

Pushed a cleanup commit that:
- Simplifies `from_bedrock_content_block` back to `Result<MessageContent>` (the `Option` wrapper was dead — `Ok(None)` was never returned)
- Makes `from_bedrock_reasoning_content_block` bail on unknown variants instead of silently dropping them (consistent with the rest of the function)
- Removes the `filter_map` that was only needed for the `Option` wrapper
- Strips comments that restate what the code does

PR #8844, opened 2026-04-25
fix: convert quoted numeric config values to numbers if needed

philipphenkel:
## Summary

  Fix typed config reads so provider timeout values stored as strings can be used as numbers.

  Provider config values are currently stored in `config.yaml` as strings, which means entries like:

  ```yaml
  OLLAMA_TIMEOUT: '1200'
  ```
  failed to convert to u64 when provider code read them as numeric values. This now works for provider timeout configs such as OPENAI_TIMEOUT, OLLAMA_TIMEOUT, and
  LITELLM_TIMEOUT.

  ## Context

  A follow-up in #8437 reported that timeout values written with quotes in config.yaml were ignored unless the quotes were removed manually:
  https://github.com/aaif-goose/goose/issues/8437#issuecomment-4291079539

Problem exist in version 1.32.0

  ## Changes

  - Keep normal YAML deserialization as the first path.
  - If typed YAML deserialization fails for a string scalar, parse it with the same scalar parsing used for environment variables.
  - Preserve string behavior when callers request String.
  - Add tests for numeric YAML values, quoted numeric YAML values, and invalid quoted numeric values.

  ## Testing
  - cargo test -p goose providers::
  - cargo test -p goose config::base::tests

Ollama provider specific test. Set timeout to 6000 via goose desktop app. 
OLLAMA_TIMEOUT: '6000' is written to the config .yaml. It has quotes around it, causing it to be ignored before this fix. With this fix the timeout value is respected.

### Related Issues
Relates to #8437
[problem report](https://github.com/aaif-goose/goose/issues/8437#issuecomment-4318175576)

### DIFF
--- a/crates/goose/src/config/base.rs
+++ b/crates/goose/src/config/base.rs
@@ -732,10 +732,20 @@ impl Config {
         }
 
         let values = self.load()?;
-        values
+        let value = values
             .get(key)
-            .ok_or_else(|| ConfigError::NotFound(key.to_string()))
-            .and_then(|v| Ok(serde_yaml::from_value(v.clone())?))
+            .ok_or_else(|| ConfigError::NotFound(key.to_string()))?;
+
+        match serde_yaml::from_value(value.clone()) {
+            Ok(value) => Ok(value),
+            Err(yaml_err) => {
+                let Some(string_value) = value.as_str() else {
+                    return Err(yaml_err.into());
+                };
+                let parsed = Self::parse_env_value(string_value)?;
+                serde_json::from_value(parsed).map_err(|_| yaml_err.into())
+            }
+        }
     }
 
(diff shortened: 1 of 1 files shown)


DOsinga (approved):
Clean fix, good tests. Thanks!

One thing: your commit is missing a `Signed-off-by` line. Please amend with `git commit --amend --signoff` and force-push so the DCO requirement is satisfied.

philipphenkel:
> Clean fix, good tests. Thanks!
> 
> One thing: your commit is missing a `Signed-off-by` line. Please amend with `git commit --amend --signoff` and force-push so the DCO requirement is satisfied.

@DOsinga Thanks for the review! I’ve added the sign-off.

Issue #8672, opened 2026-04-20
Installation script incorrectly detects Windows OS when running from WSL2 inside a mounted drive


JheisonMB:
**Describe the bug**

The download_cli.sh script fails to correctly identify the operating system when executed from a Windows-mounted directory within WSL2. Despite uname -s reporting Linux, the script defaults to the Windows build (goose-x86_64-pc-windows-msvc.zip) and attempts to install it as a .exe in a Linux binary path. It also attempts to use Windows-specific configuration paths (AppData\Roaming).

---

**Screenshots**
<img width="944" height="421" alt="Image" src="https://github.com/user-attachments/assets/43e7fd0f-ac00-46bf-b353-82ba278f818f" />

sunilkumarvalmiki:
I will take this up, work on it, and submit a PR.

JheisonMB:
Resolved:
PR #8869

PR #8869, opened 2026-04-27
fix: correct WSL2 OS detection by removing PWD-based Windows override

JheisonMB:
## Summary
Fixes a bug in download_cli.sh where WSL2 was incorrectly detected as Windows when executed from a Windows-mounted directory (e.g., /mnt/c/...).
The script previously used $PWD to decide whether to install the Windows or Linux build of goose. In WSL2, uname -s correctly reports "Linux", but the script was overriding this based on the current working directory path. This caused:

- Download of goose-x86_64-pc-windows-msvc.zip instead of the Linux binary
- Installation of a .exe file into a Linux binary path ($HOME/.local/bin/)
- Attempts to use Windows-specific configuration paths (AppData\Roaming)

### Testing
- Verified bash -n download_cli.sh passes syntax check
- Manual review of all OS detection branches:
  - Native Windows (WINDIR, msys, cygwin): still detects as windows
  - WSL (from /mnt/c/... or Linux paths): now correctly detects as linux

### Related Issues
Relates to #8672 

After:   
image

image

### DIFF
--- a/download_cli.sh
+++ b/download_cli.sh
@@ -59,14 +59,8 @@ OUT_FILE="goose"
 if [[ "${WINDIR:-}" ]] || [[ "${windir:-}" ]] || [[ "$OSTYPE" == "msys" ]] || [[ "$OSTYPE" == "cygwin" ]]; then
     # Native Windows environments - use Windows user profile path
     DEFAULT_BIN_DIR="$USERPROFILE/goose"
-elif [[ -f "/proc/version" ]] && grep -q "Microsoft\|WSL" /proc/version 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


JheisonMB:
Hey @alexhancock @matt2e @jamadeo 👋
I found a bug and opened a PR with a fix, but it hasn’t been reviewed yet.
Would really appreciate it if you could take a look when you have time. Thanks!

DOsinga (approved):
Clean fix — WSL is always a Linux environment regardless of PWD. The simplification is correct and well-motivated. Thanks for the contribution!

PR #8870, opened 2026-04-27
fix(cli): emit cumulative token usage in stream-json/json complete event

namanvats-dev:
## Summary
The `complete` event in `goose run --output-format stream-json` (and `metadata.total_tokens` in `--output-format json`) reports `Session::total_tokens`, which is the **last turn's** context size, it gets overwritten every LLM round-trip and reset to the summary output count after compaction (see `update_session_metrics` in `crates/goose/src/agents/reply_parts.rs`).

Downstream tools (eval harnesses, billing dashboards) interpret `total_tokens` as the cumulative session cost, so they under-report by an order of magnitude on multi-turn runs. We hit this in our eval harness where a session that streamed thousands of chunks reported `total_tokens=17099`, actually just the final-turn context.

**Fix:** read `Session::accumulated_total_tokens` (the true running sum), with `total_tokens` as a fallback for legacy/empty sessions. Also surface `input_tokens` / `output_tokens` so consumers can split prompt vs. completion cost without re-deriving them.

**Backwards compatible:** `total_tokens` keeps its name and position; existing consumers see a *correct* number where they previously saw an under-count. The two new fields are `Option<i32>` with `skip_serializing_if = "Option::is_none"` — omitted when unset, so no parser breakage.

**Files touched:** `crates/goose-cli/src/session/mod.rs` (`JsonMetadata`, `StreamEvent::Complete`, both emit sites).

### Testing
- [x] `cargo check -p goose-cli` : clean (verified locally, ~3 min build)
- [ ] `cargo test -p goose-cli` : please run in CI
- [x] Manual: re-ran a multi-turn `goose run --output-format stream-json` against our eval harness and confirmed the emitted `total_tokens` now matches the cumulative provider-reported usage rather than the last-turn context size

### Related Issues
N/A : surfaced internally while debugging token under-reporting in an external eval harness consuming `--output-format stream-json`. Happy to file a tracking issue if preferred.

N/A : backend change, no UX surface.

### DIFF
--- a/crates/goose-cli/src/session/mod.rs
+++ b/crates/goose-cli/src/session/mod.rs
@@ -1281,15 +1281,9 @@ impl CliSession {
                 .await
             {
                 Ok(session) => JsonMetadata {
-                    total_tokens: session
-                        .accumulated_total_tokens
-                        .or(session.total_tokens),
-                    input_tokens: session
-                        .accumulated_input_tokens
-                        .or(session.input_tokens),
-                    output_tokens: session
-                        .accumulated_output_tokens
-                        .or(session.output_tokens),
+                    total_tokens: session.accumulated_total_tokens.or(session.total_tokens),
+                    input_tokens: session.accumulated_input_tokens.or(session.input_tokens),
+                    output_tokens: session.accumulated_output_tokens.or(session.output_tokens),
                     status: "completed".to_string(),
                 },
                 Err(_) => JsonMetadata {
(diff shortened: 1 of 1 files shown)


namanvats-dev:
Closes #8871

angiejones (changes requested):
please remove comments

namanvats-dev:
@angiejones removed the comment, can you please check once

namanvats:
@lifeizhou-ap @angiejones Can anyone please take a look at this ?

DOsinga (approved):
Clean, correct fix that aligns CLI token reporting with what the server already does. The `.or()` fallback for legacy sessions is the right approach.

One thing: your commits don't have `Signed-off-by` lines. If the repo enforces DCO, you'll need to amend them with `git commit --amend --signoff` (and interactive rebase for the earlier ones). Check if that's required before merging.

PR #8893, opened 2026-04-29
fix: cache ACP configs by path

matt2e:
Now that SACP handles requests concurrently, the previous approach of creating a new Config per request could have race conditions if multiple requests tried to change config.

## Summary
This fixes ACP config races by sharing a `Config` handle for each config path instead of creating a new config object per request. The default Goose config directory still uses `Config::global()`, while custom ACP config directories use a process-wide cache keyed by the normalized `config.yaml` path.

- Add `Config::for_config_dir(...)` and `ConfigHandle` to select either the default global config or a cached custom-path config.
- Store the selected config handle on `GooseAcpAgent` and reuse it for session setup, provider/model resolution, config CRUD, and secret requests.
- Update `AcpServer` creation to read Goose mode from the same path-aware config handle.
- Keep custom config directories isolated while serializing concurrent writes for the same path.
- Add ACP regression tests for default/global selection, custom path sharing, equivalent path normalization, custom-dir coexistence, shared writes, and concurrent upserts.

### DIFF
--- a/crates/goose/src/acp/server.rs
+++ b/crates/goose/src/acp/server.rs
@@ -7,11 +7,10 @@ use crate::agents::extension_manager::TRUSTED_TOOL_UPDATE_META_KEY;
 use crate::agents::mcp_client::{GooseMcpHostInfo, McpClientTrait};
 use crate::agents::platform_extensions::developer::DeveloperClient;
 use crate::agents::{Agent, AgentConfig, ExtensionConfig, GoosePlatform, SessionConfig};
-use crate::config::base::CONFIG_YAML_NAME;
 use crate::config::extensions::get_enabled_extensions_with_config;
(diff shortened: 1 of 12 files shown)


kalvinnchau:
AI comment:

This removes the agent-owned config path, so `GooseAcpAgent::new(..., config_dir, ...)` no longer guarantees config reads/writes use that directory. That already affects the ACP test fixture, and it would
affect any caller that wants an ACP agent with a custom config path.

If ACP should always use process-global config, should we remove `config_dir` from `GooseAcpAgent::new` or initialize/validate the global config there? Otherwise this constructor still looks like it accepts
a custom config dir even though config access now depends on prior global initialization.

kalvinnchau (approved):
lgtm!

PR #8900, opened 2026-04-29
fix: model picker stays usable during provider loading

matt2e:
https://github.com/user-attachments/assets/fbaacc25-81ec-474c-8006-7480e59e94bf

## UI Changes

- **No more "Loading" flash on startup** — the model picker button shows the current provider name (e.g. "Claude") or model name instead of briefly flashing "Loading" while providers refresh
- **Button stays clickable** — the picker is only disabled when loading AND no provider is already known, so users can interact with it during background refreshes
- **Smarter loading state in dropdown** — when models are loading but the current model name is known, shows it as a placeholder item with a spinner instead of a generic "Loading models…" message
- **Background refresh on open** — opening the picker triggers a silent inventory refresh so newly available models/providers appear without an app restart

## Implementation

1. **Reorder button label priority** (`AgentModelPicker.tsx`) — Display `triggerModelLabel ?? selectedAgentLabel` first; only fall back to "Loading" when neither is available. Conditionally disable the button only when `loading && !selectedAgentLabel`.

2. **Deduplicate startup RPC** (`useAppStartup.ts`, `acp.ts`, `acpApi.ts`) — Merge `loadProviders` and `loadProviderInventory` into a single `loadProvidersAndInventory` call. Extract `buildProviderListFromEntries()` so both `listProviders()` and the startup path can derive the provider list from the same inventory response, eliminating a redundant `_goose/providers/list` RPC.

3. **Extract shared background refresh** (`inventory.ts`) — Move the refresh+poll loop (refresh configured providers, poll with delays until all finish) into a reusable `backgroundRefreshInventory()` function. It merges results into the store without setting `loading=true`, so cached data stays visible during the refresh.

4. **Background refresh on picker open** (`useAgentModelPickerState.ts`) — Add a `handlePickerOpen` callback that triggers `backgroundRefreshInventory()` when the popover opens (guarded by a ref to prevent concurrent refreshes). This replaces the old `useEffect` in `AgentModelPicker` that called `getProviderInventory()` on every open and merged results synchronously.

5. **Thread `onPickerOpen` through the component tree** — Wire `handlePickerOpen` from the hook through `useResolvedAgentModelPicker` → `useChatSessionController` → `ChatView`/`HomeScreen` → `ChatInput` → `ChatInputToolbar` → `AgentModelPicker`'s `onOpen` prop, which fires on `onOpenChange(true)`.

### DIFF
--- a/ui/goose2/src/features/providers/api/inventory.ts
+++ b/ui/goose2/src/features/providers/api/inventory.ts
@@ -59,8 +59,14 @@ export async function backgroundRefreshInventory(
     .map((entry) => entry.providerId);
   if (configuredProviderIds.length === 0) return;
 
+  const refresh = await refreshProviderInventory(configuredProviderIds);
+  if (refresh.started.length === 0 && (refresh.skipped ?? []).length === 0) {
+    return;
+  }
+
   const { syncProviderInventory } = await import("./inventorySync");
   await syncProviderInventory(configuredProviderIds, {
+    initialRefresh: refresh,
     onEntries: (entries) => inventoryStore.mergeEntries(entries),
   });
 }
--- a/ui/goose2/src/features/providers/api/inventory.ts
+++ b/ui/goose2/src/features/providers/api/inventory.ts
@@ -33,8 +33,12 @@ export async function refreshProviderInventory(
 
 /**
  * Refresh configured provider inventories in the background, polling until
- * all providers finish refreshing. Does NOT set the store's `loading` flag,
- * so the UI keeps showing cached data during the refresh.
+ * all providers finish refreshing. If no entries are supplied, fetch and merge
+ * the current inventory snapshot first so the UI sees fresh cached data even
+ * when no refresh starts.
+ *
+ * Does NOT set the store's `loading` flag, so the UI keeps showing cached data
+ * during the refresh.
  */
 export async function backgroundRefreshInventory(
   inventoryStore: {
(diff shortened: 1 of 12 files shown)


ui/goose2/src/features/providers/api/inventory.ts
```diff
@@ -30,3 +30,43 @@ export async function refreshProviderInventory(
...
+ * all providers finish refreshing. Does NOT set the store's `loading` flag,
+ * so the UI keeps showing cached data during the refresh.
+ */
+export async function backgroundRefreshInventory(
+  inventoryStore: {
+    mergeEntries: (entries: ProviderInventoryEntryDto[]) => void;
+  },
+  initialEntries?: ProviderInventoryEntryDto[],
+): Promise<void> {
+  const entries =
```
kalvinnchau: AI:
The new path calls ui/goose2/src/features/providers/api/inventory.ts:45. When called from picker open, it fetches current entries here:
```
const entries = initialEntries?.length ? initialEntries : await getProviderInventory();
```
But it does not merge those fetched entries. It only merges later if a refresh actually starts:
```
const refresh = await refreshProviderInventory(configuredProviderIds);
if (refresh.started.length === 0) return;

inventoryStore.mergeEntries(await getProviderInventory(refresh.started));
```
So these cases now fetch fresh inventory and then throw it away

- no configured providers
- configured providers that do not support refresh
- providers skipped because they are already refreshing
- refresh request returns started: [] for any reason

baxen (approved):

PR #8906, opened 2026-04-29
fix(provider_registry): case-insensitive model name lookup for context_limit

### LINKED ISSUE #8752: Case mismatch in custom provider model name breaks `context_limit` lookup
### Bug

When a custom provider is created through the UI/CLI, the model name is saved **as-is** (preserving the original case, e.g. `GLM-5.1`). However, when goose resolves the model's `context_limit` at runtime, it performs a **strict case-sensitive comparison** against `GOOSE_MODEL` from `config.yaml`, which is stored in **lowercase**. This causes the lookup to fail, silently falling back to the default `128_000` context limit.

### Root cause

Two locations that disagree on normalization:

1. **`create_custom_provider` / `update_custom_provider`** ([`declarative_providers.rs`](https://github.com/block/goose/blob/main/crates/goose/src/config/declarative_providers.rs)) — saves the model name without normalization:
   ```rust
   // Line ~217 (create) and ~266 (update)
   .map(|name| ModelInfo::new(name, 128000))
   ```

2. **`normalize_model_config`** ([`provider_registry.rs`](https://github.com/block/goose/blob/main/crates/goose/src/providers/provider_registry.rs)) — performs a strict equality check:
   ```rust
   if let Some(info) = self.metadata.known_models
       .iter()
       .find(|m| m.name == model.model_name && m.context_limit > 0)
   {
       model.context_limit = Some(info.context_limit);
   }
   ```
   `model.model_name` comes from `ModelConfig::new()` → `GOOSE_MODEL` in `config.yaml`, which is lowercased. But `m.name` in `known_models` retains the original case from the JSON config.

### Impact

Custom provider models with uppercase letters in their name (e.g. `GLM-5.1`, `Qwen3-Coder-480B`) will always use the default 128k context limit, ignoring the `context_limit` value set in their JSON config file.

### Suggested fix

Add case-insensitive comparison in `normalize_model_config`:

```rust
.find(|m| m.name.to_lowercase() == model.model_name.to_lowercase() && m.context_limit > 0)
```

Or alternatively, normalize the model name to lowercase in `create_custom_provider` / `update_custom_provider` when writing the JSON config.

### Workaround

Manually edit the custom provider JSON file (`~/.config/goose/custom_providers/<provider>.json`) to use the exact same case as `GOOSE_MODEL` in `config.yaml` (typically lowercase).

### Environment

- goose version: 1.31.1
- OS: Debian 12

bzqzheng:
## Problem

Custom provider model names with case mismatches (e.g. `GLM-5.1` in the provider JSON config vs `glm-5.1` in `GOOSE_MODEL`) silently failed the `known_models` `context_limit` lookup in `normalize_model_config` and fell back to the default 128k.

Fixes #8752.

## Change

Use `eq_ignore_ascii_case` instead of strict equality when matching `model.model_name` against entries in `metadata.known_models`.

This is the smallest possible fix: it only affects the lookup, does not mutate stored model names, and does not change what gets sent to provider APIs.

## Verification

- Added 3 unit tests covering:
  - Case mismatch (uppercase in registry, lowercase in model config) → context_limit resolved correctly
  - Exact case match → no regression
  - Unknown model → falls back gracefully (context_limit stays None)
- All 3 tests pass.

## Checklist

- [x] DCO sign-off
- [x] Tests added
- [x] No regression for native providers (Anthropic, OpenAI) where case is already lowercase


DOsinga (approved):
Good fix — the case-insensitive lookup is the right call here.

I removed the tests: they were only asserting that `eq_ignore_ascii_case` works as documented (exact match still matches, non-match still doesn't). The case-mismatch test was the only one that tested the actual bug, and even that is really just testing a standard library function at this point. The one-liner speaks for itself.

One process note: please don't open multiple PRs at the same time — submit one, wait for it to land, then open the next. See [the contributing guide](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#pull-requests).

PR #8910, opened 2026-04-29
fix(cli): report cumulative total_tokens in stream-json/json output

### LINKED ISSUE #8871: # stream-json/json `complete.total_tokens` reports last-turn context, not cumulative session usage
**Describe the bug**

`goose run --output-format stream-json` ends with `{"type":"complete","total_tokens":N}` (and `--output-format json` exposes `metadata.total_tokens`). Both currently emit `Session::total_tokens`, which is the *last LLM call's* context size — overwritten on every turn and reset to the summary output count after compaction. The intended cumulative session usage lives in `Session::accumulated_total_tokens` (see [`update_session_metrics`](https://github.com/aaif-goose/goose/blob/main/crates/goose/src/agents/reply_parts.rs#L482-L509)).

Downstream tools (eval harnesses, billing dashboards) interpret `total_tokens` as the cumulative session cost, so they under-report by an order of magnitude on multi-turn runs.

Concrete example from our harbor evals: a session with ~1100 streamed chunks and several tool calls reported `complete.total_tokens=17099`, while the cumulative provider-reported usage across all turns was many times higher.

---

**To Reproduce**

1. Run a multi-turn goose session with stream-json output:
goose run --recipe my-recipe.yaml --output-format stream-json 2>&1 | tee out.txt

2. Inspect the final event in `out.txt`: `{"type":"complete","total_tokens":N}`.
3. Compare `N` against `accumulated_total_tokens` from the same session:
goose session list --format json --limit 1

(or against upstream provider-reported usage).
4. Observe: `N` matches `total_tokens` (last-turn only) instead of `accumulated_total_tokens` (cumulative).

---

**Expected behavior**

`complete.total_tokens` (and `metadata.total_tokens` in `--output-format json`) should report the **cumulative** session usage : `Session::accumulated_total_tokens` so downstream tooling can rely on it for billing and eval reporting without re-deriving from the SQLite session DB.

Bonus: surface `input_tokens` / `output_tokens` alongside, so consumers don't need to estimate the prompt/completion split.

---

**Screenshots**

N/A — backend / output-format issue, no UI surface.

---

- **OS & Arch:** Linux x86_64 (Docker container; bug is platform-independent — lives in CLI session-event emit code, reproduces anywhere)
- **Interface:** CLI
- **Version:** v1.32.0 (reproduces on `stable` as well)
- **Extensions enabled:** developer (builtin), todo (platform), summon (platform)
- **Provider & Model:** openrouter / minimax-m2.5 (provider-agnostic — bug is in session-event emit, not provider accounting)

---

**Additional context**

Root cause: [`crates/goose-cli/src/session/mod.rs:1289`](https://github.com/aaif-goose/goose/blob/main/crates/goose-cli/src/session/mod.rs#L1289) reads `s.total_tokens` instead of `s.accumulated_total_tokens` for the stream-json `complete` event. Same issue at line 1268 for `--output-format json`.

bzqzheng:
## Problem

The CLI's "--json" and "--stream-json" output modes were reading `session.total_tokens`, which stores only the **last turn's** token usage. This caused `complete.total_tokens` to reset to the current turn's count on every chunk, rather than reporting the running session total (fixes #8871).

## Root Cause

- `update_session_metrics` in `crates/goose/src/agents/reply_parts.rs` correctly computes `accumulated_total_tokens` by adding usage across turns.
- However, `total_tokens` is set to the **current turn's** usage.
- The CLI output code in `crates/goose-cli/src/session/mod.rs` reads `session.total_tokens` instead of `session.accumulated_total_tokens`.

## Fix

Read `session.accumulated_total_tokens` in:
- `JsonMetadata` construction (`--json` mode)
- `StreamEvent::Complete` emission (`--stream-json` mode)
- `get_total_token_usage()` for consistency

## Test

Added `test_accumulated_total_tokens_across_multiple_turns` which:
1. Creates a mock provider reporting 15 tokens per turn
2. Runs 3 reply turns
3. Asserts `accumulated_total_tokens == 45` (cumulative)
4. Asserts `total_tokens == 15` (last turn only)

## Verification

```bash
cargo test --package goose test_accumulated_total_tokens_across_multiple_turns
cargo test --package goose-cli
```


DOsinga (approved):
Merged main, resolved conflicts, cleaned up and approved.

Changes made during cleanup:
- Removed unused `ProviderDef` impl and `call_count` field from the mock provider
- Extracted `run_turn` helper to reduce test boilerplate
- Reduced test from 3 turns to 2 (sufficient to prove accumulation)
- Removed assert messages that just restate the assertion
- Removed `#[cfg(test)]` on inner module (already inside a test module)
- Stripped comment on `get_total_token_usage` that restated the function name

Note: main already fixed the json/stream-json output paths (using `.or()` fallback). The remaining unique fix here is `get_total_token_usage` using `accumulated_total_tokens`. That function is currently unused but the fix is correct.

PR #8911, opened 2026-04-29
goose2 distribution bundling

jamadeo:


crates/goose/src/config/base.rs
```diff
@@ -170,6 +176,7 @@ impl Default for Config {
         if let Some(defaults) = bundled_defaults_path() {
```
lifeizhou-ap: We can remove this now.  The defaults.yaml was just introduced for security (prompt injection configurations) before, we have removed this file in the goose core. 

However, I guess the `default.yaml` might be used as the bundle default in this PR. I feel it is a bit overkill, as we already have system level, additional_config and user level. probably these three are already good enough?

ui/goose2/src-tauri/src/services/distro_bundle.rs
```diff
@@ -0,0 +1,165 @@
...
+use std::collections::HashMap;
+use std::env;
+use std::path::{Path, PathBuf};
+use tauri::{AppHandle, Manager};
+
+const DISTRO_DIR_NAME: &str = "distro";
+const DISTRO_JSON_NAME: &str = "distro.json";
+const DISTRO_CONFIG_NAME: &str = "config.yaml";
+const DISTRO_BIN_DIR_NAME: &str = "bin";
+const DISTRO_OVERRIDE_ENV: &str = "GOOSE_DISTRO_DIR";
```
lifeizhou-ap: Nit: Not sure we need this variable.  It seems it only used once.  we can either inline or reuse it in the other places using the string value. eg 

`command.env("GOOSE_DISTRO_DIR", &bundle.root_dir);`

lifeizhou-ap (approved):
looks good, thanks @jamadeo! I left 2 comments, please feel free to take or leave them.

crates/goose/src/config/base.rs
jamadeo: That's good to hear -- I'd only included it here for backwards compatibility. Will remove!

PR #8912, opened 2026-04-29
feat: make ollama host configurable in goose2

kalvinnchau:
**Category:** fix
**User Impact:** Users can configure Ollama with a custom host or port from goose2 instead of having it appear connected by default.
**Problem:** goose2 treated Ollama as connected because the backend default host counted as configured, even when users had not saved any Ollama settings. That made it unclear how to use a non-default Ollama host or port.
**Solution:** Ollama now reports configured status only when `OLLAMA_HOST` is explicitly set, and goose2 exposes a host field pre-filled with the effective local default so users can save or edit it.

<details>

**crates/goose/src/providers/ollama.rs**
Overrides Ollama inventory configuration status so the runtime fallback host does not make the provider appear configured. Adds focused tests for unset, environment, and saved-config host states.

**ui/goose2/src/features/providers/providerCatalog.ts**
Changes Ollama from a local zero-config entry to a configurable provider with an `OLLAMA_HOST` field and default host value.

**ui/goose2/src/features/providers/providerCatalog.test.ts**
Covers the Ollama catalog entry so the host field and default remain present.

**ui/goose2/src/features/settings/ui/ModelProviderRow.test.tsx**
Adds coverage that provider field defaults are pre-filled and saved from the setup form.

**ui/goose2/src/features/settings/ui/modelProviderHelpers.tsx**
Uses provider field defaults when initializing setup drafts and no saved config value exists.

**ui/goose2/src/shared/types/providers.ts**
Adds an optional `defaultValue` property to provider field metadata.

</details>

### Reproduction Steps

1. Open goose2 settings and go to model providers.
2. Expand Ollama before setting `OLLAMA_HOST`.
3. Confirm it is not shown as already connected and that the Host field is pre-filled with `http://localhost:11434`.
4. Change the Host field to a custom Ollama endpoint or keep the default, then save.
5. Confirm Ollama is marked connected and model inventory refresh starts for the saved host.

### Screenshots/Demos

Not included. The UI change is a settings form field and can be verified with the reproduction steps above.

### Verification

- `cargo fmt`
- `pnpm exec biome format --write ...`
- `git diff --check`
- `git push -u origin HEAD` pre-push checks passed, including goose2 `pnpm test`, `pnpm check`, `pnpm build`, Tauri `cargo fmt --check`, `cargo check`, and `cargo clippy -- -D warnings`.

## before:
image

## after:
image

### DIFF
--- a/crates/goose/src/providers/ollama.rs
+++ b/crates/goose/src/providers/ollama.rs
@@ -122,6 +122,10 @@ fn apply_ollama_options(payload: &mut Value, model_config: &ModelConfig) {
     }
 }
 
+fn ollama_host_configured(config: &crate::config::Config) -> bool {
+    std::env::var("OLLAMA_HOST").is_ok() || config.get("OLLAMA_HOST", false).is_ok()
+}
+
 impl OllamaProvider {
     pub async fn from_env(model: ModelConfig) -> Result<Self> {
         let config = crate::config::Config::global();
@@ -260,6 +264,10 @@ impl ProviderDef for OllamaProvider {
 
     fn supports_inventory_refresh() -> bool {
         true
+    }
+
+    fn inventory_configured() -> bool {
+        ollama_host_configured(crate::config::Config::global())
     }
 
     fn inventory_identity() -> Result<InventoryIdentityInput> {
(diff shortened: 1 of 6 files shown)


crates/goose/src/providers/ollama.rs
```diff
@@ -122,6 +122,10 @@ fn apply_ollama_options(payload: &mut Value, model_config: &ModelConfig) {
     }
 }
 
+fn ollama_host_configured(config: &crate::config::Config) -> bool {
+    std::env::var("OLLAMA_HOST").is_ok() || config.get("OLLAMA_HOST", false).is_ok()
```
baxen: meganit: i think config has a func that gets from env var first so you don't need the separate check,but this works ofc

baxen (approved):
LGTM

crates/goose/src/providers/ollama.rs
kalvinnchau: ooh yeah good catch, updating

PR #8913, opened 2026-04-29
feat: support google model inventory refresh

kalvinnchau:
**Category:** fix
**User Impact:** Users setting up Google Gemini in goose2 now see model refresh and API key errors handled consistently with other providers.
**Problem:** Google Gemini credentials could be saved in goose2 without triggering the provider inventory refresh used by other model providers. Invalid Google API keys also produced a misleading missing models array error because the models response was parsed before checking the HTTP status.
**Solution:** Enable inventory refresh for the Google provider and include the relevant host/API key in its inventory identity. The Google model listing path now maps non-success responses before parsing the model list, so invalid keys surface the provider's actual error.

<details>

**crates/goose/src/providers/google.rs**
Adds Google provider support for ACP inventory refresh and identifies inventory entries by `GOOGLE_HOST` and `GOOGLE_API_KEY`. Updates model listing to handle non-success responses before parsing the `models` array, matching the behavior users expect from other providers.

</details>

### Reproduction Steps

1. Open goose2 provider settings and set up Google Gemini with an invalid API key.
2. Save the credentials and wait for the model refresh result.
3. Confirm the refresh warning shows the Google API key error instead of a missing `models` array message.
4. Set up Google Gemini with a valid API key and confirm model inventory refresh completes and Gemini models are listed.

### DIFF
--- a/crates/goose/src/providers/google.rs
+++ b/crates/goose/src/providers/google.rs
@@ -1,7 +1,7 @@
 use super::api_client::{ApiClient, AuthMethod};
 use super::base::MessageStream;
 use super::errors::ProviderError;
-use super::openai_compatible::handle_status;
+use super::openai_compatible::{handle_status, map_http_error_to_provider_error};
 use super::retry::ProviderRetry;
 use super::utils::RequestLog;
 use crate::conversation::message::Message;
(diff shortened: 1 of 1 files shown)


baxen (approved):
LGTM

Issue #8915, opened 2026-04-29
Bedrock prompt cache targets earliest messages; Anthropic recommends a trailing breakpoint for growing conversations


carl-auctane:
## Summary

When `BEDROCK_ENABLE_CACHING=true` and an Anthropic Claude model is in use, `BedrockProvider::converse()` places a cache breakpoint on the first three user/assistant messages. This doesn't match how Anthropic (and Bedrock, which implements the same semantics) actually look up cached prefixes, and in an agentic tool-use loop it causes fresh-token processing to grow linearly with turn count instead of staying bounded.

Introduced in #6710 (merged 2026-02-23).

## Where

`crates/goose/src/providers/bedrock.rs`, in `converse()` around lines 232-256:

```rust
let visible_messages: Vec<&Message> =
    messages.iter().filter(|m| m.is_agent_visible()).collect();

// Cache the earliest messages (not most recent) because prompt caching
// requires exact prefix matching — caching recent messages would shift
// positions each turn, causing misses.
const MESSAGE_CACHE_BUDGET: usize = 3;
let cache_count = if enable_caching {
    visible_messages.len().min(MESSAGE_CACHE_BUDGET)
} else {
    0
};

let mut request = self
    .client
    .converse()
    .set_system(Some(system_blocks))
    .model_id(model_name.to_string())
    .set_messages(Some(
        visible_messages
            .iter()
            .enumerate()
            .map(|(idx, m)| to_bedrock_message_with_caching(m, idx < cache_count))
            .collect::<Result<_>>()?,
    ));
```

The comment states the reasoning, quoted verbatim:

> Cache the earliest messages (not most recent) because prompt caching requires exact prefix matching — caching recent messages would shift positions each turn, causing misses.

## Why this is wrong

Anthropic's [prompt caching docs](https://platform.claude.com/docs/en/build-with-claude/prompt-caching) describe the lookup model:

> Cache writes happen only at your breakpoint. Marking a block with `cache_control` writes exactly one cache entry: a hash of the prefix ending at that block.

> Cache reads look backward for entries that prior requests wrote... It is looking for prior writes, not for stable content.

> The lookback window is 20 blocks.

Breakpoints are identified by the **hash of the prefix up to the block**, not by position index. "Shifting positions" is not a failure mode. The docs' key takeaway is explicit:

> Place `cache_control` on the last block whose prefix is identical across the requests you want to share a cache. In a growing conversation the final block works as long as each turn adds fewer than 20 blocks.

[...]

PR #8916, opened 2026-04-29
fix(bedrock): cache trailing message for stable prefix across agent turns

carl-auctane:
## Summary

When `BEDROCK_ENABLE_CACHING=true`, `BedrockProvider::converse()` currently places cache points on the first three visible messages. This doesn't match how Anthropic and Bedrock actually look up cached prefixes: cache entries are keyed by the hash of the prefix ending at the breakpoint, and reads walk backward up to 20 blocks looking for prior writes. With a breakpoint fixed to early messages, every turn reprocesses everything appended after position 3, which in an agent loop grows linearly with turn count.

This change places the cache point on the trailing message instead. On each new turn the lookback finds the breakpoint the previous turn wrote, so fresh processing is bounded to the content added since the last request. This matches the pattern Anthropic's prompt caching documentation recommends for growing conversations:

> Place `cache_control` on the last block whose prefix is identical across the requests you want to share a cache. In a growing conversation the final block works as long as each turn adds fewer than 20 blocks.

The system-prompt cache point is unchanged. The misleading comment about "caching recent messages would shift positions each turn" is replaced with an accurate description of the lookup model.

For a worked cost comparison over a 10-turn agent loop and links to the relevant Anthropic and Bedrock docs, see the linked issue.

### Testing

- `cargo fmt --all`
- `cargo check -p goose`
- `cargo clippy -p goose --all-targets -- -D warnings` (no warnings)
- `cargo test -p goose --lib providers::bedrock` (4 passed)
- `cargo test -p goose --lib providers::formats::bedrock` (11 passed)

No test changes were needed. The existing per-message helper tests in `providers::formats::bedrock` exercise `to_bedrock_message_with_caching` directly with `enable_caching=true` and are unaffected. The `test_caching_*` tests in `providers::bedrock` assert on `should_enable_caching()` returning the right boolean and are also unaffected.

### Related Issues

Relates to #8915


DOsinga (approved):
Nice fix — the analysis in #8915 is thorough and the change is correct. I pushed a small cleanup commit to simplify the index logic (saturating_sub instead of checked_sub + Option wrapping) and dropped the block comment since the issue already covers the rationale.

PR #8926, opened 2026-04-30
improve extensions management and usage context

morgmart:
🤖 **Category:** improvement
**User Impact:** Users get a dedicated Extensions view and clearer chat context about which extensions were actually used, while agent-enabled MCP extensions are now scoped to the current chat/session instead of silently changing future chats.

**Problem:** Extension management was split across settings, config defaults, chat-time extension usage, and agent-managed MCP enablement. The previous branch behavior blurred those boundaries: when the Extension Manager enabled an MCP extension from chat, it persisted that as a global config change for future chats. We also had a temporary special case that kept global toggles for three default-off Goose capabilities (`chatrecall`, `summarize`, and `code_execution`), which is no longer the right model.

**Solution:** This PR separates extension defaults from session state:

- The Extensions page manages configured extensions and global defaults for future chats.
- User-managed external extensions (`stdio` / `streamable_http`) can still be added, edited, and deleted from the Extensions view, but they no longer have global enable/disable toggles.
- Built-in/platform Goose capabilities no longer get the old special-case global toggles for `chatrecall`, `summarize`, and `code_execution`.
- When the agent uses the Extension Manager to enable or disable an extension, that change updates the live extension manager and persists only to the current session's `extension_data.enabled_extensions`.
- Resuming a session restores that session-specific extension set; starting a different chat falls back to the global defaults unless that chat has its own saved session extension state.
- The chat context panel now uses typed session extension status data to show only extensions actually used by tool requests in that conversation, including connection state and tool counts.

<details>
<summary>Notable implementation details</summary>

**Session-scoped agent extension management**

- `crates/goose/src/agents/platform_extensions/ext_manager.rs` now treats `manage_extensions` as a session operation.
- Enable looks up the configured extension, adds it to the live `ExtensionManager`, upserts it into the current session's saved extension list, and reports that it was enabled "for this session."
- Disable removes the extension from the current session's desired extension list and from the live manager when present. This also handles the case where an extension was saved on the session but failed to connect.
- The extension manager no longer writes agent-managed enable/disable actions back to global config.

**Session status API and generated types**

- `crates/goose-sdk/src/custom_requests.rs` adds typed DTOs for session extension status, including `ExtensionConfigDto`, `SessionExtensionStatusDto`, and `ExtensionConnectionStatusDto`.
- `crates/goose/src/acp/server/extensions.rs` returns typed status records instead of ad hoc JSON, redacts secret env/header values, and includes tool inventory for connected extensions.
- Generated ACP schema and TypeScript SDK files were regenerated for the new typed status response.

**Global config and settings behavior**

- `crates/goose/src/config/extensions.rs` resolves extension lookup through normalized config keys, so display names and config keys line up more reliably.
- `ui/goose2/src/features/extensions/ui/ExtensionItem.tsx` no longer renders global enable/disable toggles; user-managed external extensions remain configurable via the settings button.
- `ui/goose2/src/features/extensions/hooks/useExtensionsSettings.ts` preserves an existing extension's `enabled` value when editing its config.
- The previous special-case toggle list for `chatrecall`, `summarize`, and `code_execution` was removed.

**Chat extension usage**

- `ui/goose2/src/features/chat/ui/widgets/ExtensionsWidget.tsx` fetches `_goose/session/extensions/status`, merges in configured extension metadata for historical/unavailable usage, and displays only extensions used by tool requests in the current chat.
- The widget refreshes when a new tool owner appears, so newly enabled/used extensions do not stay stale.
- `ui/goose2/src/shared/api/acpNotificationHandler.ts` now reads tool identity from the actual `_meta.goose.mcpApp` metadata shape, keeps the older `_meta.goose.toolCall` path as a fallback, and attaches `toolName` / `extensionName` in both live and replay flows.

**Navigation and UI**

- Extensions now has a top-level sidebar entry and dedicated page shell.
- The settings modal no longer owns extension management.
- Extension rows, filters, usage helpers, and i18n copy were updated for the new management model.

</details>

### Reproduction Steps

1. Open Goose desktop and confirm the sidebar includes an Extensions navigation item.
2. Navigate to Extensions and verify configured extensions are searchable and grouped between apps/services and Goose capabilities.
3. Confirm user-managed `stdio` / `streamable_http` extensions can be configured, edited, and deleted, but do not show global enable/disable controls.
4. Confirm built-in/platform Goose capabilities, including `chatrecall`, `summarize`, and `code_execution`, do not get special global enable/disable controls.
5. From chat, ask Goose to use an extension that is not active in the current session.
6. Confirm the Extension Manager can enable it for that chat and that the response says it was enabled for the session.
7. Start or open another chat and confirm that agent-enabled extension does not become globally enabled there unless it is part of that chat's own saved session state.
8. Open a chat that has used extension-backed tools and check the context panel Details tab.
9. Confirm the Extensions used widget lists only extensions used in that chat, with connection status and tool counts.
10. Add, edit, disable, or delete a configurable external extension and confirm the list refreshes without reopening the page.

### Verification

- `cargo fmt`
- `just generate-acp-types`
- `pnpm test src/features/extensions/lib/__tests__/extensionUsage.test.ts src/features/extensions/lib/__tests__/extensionCategories.test.ts`
- `pnpm test src/features/extensions/ui/__tests__/ExtensionsSettings.test.tsx`
- `pnpm test src/shared/api/__tests__/acpNotificationHandler.test.ts`
- `pnpm typecheck`
- `git diff --check`
- Commit/pre-commit hooks passed their configured formatting, file-size, i18n, and typecheck steps. They emitted existing non-fatal warnings about broken local skill symlinks and platform/engine mismatches.

https://github.com/user-attachments/assets/49b26265-9ca1-47ef-8413-df73408036ad


ui/goose2/src/features/chat/ui/widgets/ExtensionsWidget.tsx
```diff
@@ -1,90 +1,131 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+  const messages = useChatStore(
+    (s) => s.messagesBySession[sessionId] ?? EMPTY_MESSAGES,
+  );
+
+  const fetchStatuses = useCallback(() => {
+    setIsLoading(true);
+    listSessionExtensionStatus(sessionId)
+      .then((all) => setExtensions(all))
+      .catch(() => setExtensions([]))
```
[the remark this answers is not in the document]
morgmart: Addressed in `3a426e383` by guarding the async session extension status refresh inside the effect. Stale requests now skip `setExtensions` / `setIsLoading` after the session or tool-request signature changes, so older responses cannot overwrite the current chat state. Verification: `pnpm typecheck`, targeted Biome check, and the full pre-push suite passed.

crates/goose/src/acp/server.rs
```diff
@@ -3153,6 +3178,101 @@ impl GooseAcpAgent {
...
+        for extension in connected_extensions {
+            let key = name_to_key(&extension.name());
+            if seen_keys.insert(key) {
+                extensions.push(extension);
+            }
+        }
+
+        let mut extensions_json = Vec::new();
+        for extension in extensions {
+            if is_hidden_extension(&extension.name()) {
```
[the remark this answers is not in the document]
morgmart: Thanks for the callout. We checked this against the existing Electron extensions behavior and are intentionally keeping hidden extension visibility as a backend-owned registry decision through `is_hidden_extension`. We are not adding a separate type gate in the ACP session-status path in this PR, so the status response stays consistent with the extension-management response and the existing backend behavior.

crates/goose/src/acp/server.rs
```diff
@@ -3045,7 +3068,10 @@ impl GooseAcpAgent {
 
     #[custom_method(GetExtensionsRequest)]
     async fn on_get_extensions(&self) -> Result<GetExtensionsResponse, sacp::Error> {
-        let extensions = crate::config::extensions::get_all_extensions();
+        let extensions = crate::config::extensions::get_all_extensions()
+            .into_iter()
+            .filter(|ext| !is_hidden_extension(&ext.config.name()))
```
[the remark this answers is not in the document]
morgmart: Thanks for flagging this. We checked the existing Electron path and it also treats hidden extension visibility as backend-owned behavior via `is_hidden_extension`, rather than adding UI/API-specific type filtering. For this PR we are going to keep trusting that backend registry as the source of truth and keep the ACP extension list aligned with that behavior. If we want to revisit name-collision semantics, that should be a follow-up backend/product decision rather than a separate special case here.

ui/goose2/src/features/sidebar/ui/SidebarRecentsSection.tsx
```diff
@@ -96,16 +96,16 @@ export function SidebarRecentsSection({
...
-            size="xs"
+            size="icon-xs"
             onClick={onNewChat}
             aria-label={t("actions.newChat")}
             title={t("actions.newChat")}
             className={cn(
-              "mr-1 h-6 flex-shrink-0 rounded-full bg-muted px-2 text-[11px] text-foreground opacity-0 transition-opacity duration-150 ease-out hover:bg-muted/80 hover:text-foreground",
-              "pointer-events-none group-hover/chats-header:pointer-events-auto group-hover/chats-header:opacity-100 focus-visible:pointer-events-auto focus-visible:opacity-100",
+              "mr-1 size-6 flex-shrink-0 rounded-md",
+              "invisible opacity-0 group-hover/chats-header:visible group-hover/chats-header:opacity-100 group-focus-within/chats-header:visible group-focus-within/chats-header:opacity-100",
```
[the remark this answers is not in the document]
morgmart: Thanks for flagging this. We agree this is a valid keyboard-accessibility issue, and the same pattern should be reviewed across the sidebar controls rather than patched only on this one Recents button. We are going to handle that as a focused follow-up PR so the sidebar hover/focus behavior can be made consistent across Recents, Projects, and chat row action menus.

crates/goose-sdk/src/custom_requests.rs
```diff
@@ -148,6 +148,21 @@ pub struct GetSessionExtensionsResponse {
...
+    response = GetSessionExtensionStatusResponse
+)]
+#[serde(rename_all = "camelCase")]
+pub struct GetSessionExtensionStatusRequest {
+    pub session_id: String,
+}
+
+#[derive(Debug, Default, Clone, Serialize, Deserialize, JsonSchema, JsonRpcResponse)]
+pub struct GetSessionExtensionStatusResponse {
+    pub extensions: Vec<serde_json::Value>,
```
kalvinnchau: imo we should make this a typed payload rather than just a `serde_json::Value`, so the generated types can use real types rather than a `Array<unknown>`

this would help clean up alot of the string/json parsing logic

crates/goose/src/config/extensions.rs
```diff
@@ -114,6 +114,32 @@ pub fn set_extension_enabled(key: &str, enabled: bool) {
     }
 }
 
+fn set_extension_enabled_by_name_in_map(
+    extensions: &mut IndexMap<String, ExtensionEntry>,
+    name: &str,
+    enabled: bool,
+) -> bool {
+    let Some(entry) = extensions
```
kalvinnchau: think we need to also normalize the extension name here:
>If the caller only has a name, resolve it once to a canonical config key using the same normalization path as ExtensionConfig::key() / name_to_key().

crates/goose/src/config/extensions.rs
```diff
@@ -114,6 +114,32 @@ pub fn set_extension_enabled(key: &str, enabled: bool) {
...
+        .values_mut()
+        .find(|entry| entry.config.name() == name)
+    else {
+        return false;
```
kalvinnchau: are we purposely treating the extension not found case as a false? or should this error?

crates/goose/src/acp/server.rs
```diff
@@ -3153,6 +3178,101 @@ impl GooseAcpAgent {
...
+                obj.insert(
+                    "config_key".to_string(),
+                    serde_json::Value::String(config_key),
+                );
+                obj.insert(
+                    "status".to_string(),
+                    serde_json::Value::String(if connected {
+                        "connected".to_string()
+                    } else {
+                        "failed".to_string()
```
kalvinnchau: related to my other comment about typed responses - the states here could be an ENUM

crates/goose-sdk/src/custom_requests.rs
morgmart: 🤖 Fixed in the latest branch tip: `_goose/session/extensions/status` now returns `Vec<SessionExtensionStatusDto>` backed by typed `ExtensionConfigDto` payloads instead of `Vec<serde_json::Value>`, so generated SDK types are concrete rather than `unknown[]`.

crates/goose/src/config/extensions.rs
morgmart: 🤖 Fixed in the latest branch tip: name lookup now resolves through `extension_key_for_name_in_map`, which normalizes with `name_to_key` and matches `ExtensionConfig::key()` instead of doing a raw name comparison.

crates/goose/src/config/extensions.rs
morgmart: 🤖 Fixed by removing this helper path. Config toggles now validate the supplied config key in the ACP handler and return `invalid_params` when it is missing; session-scoped extension-manager changes return `RESOURCE_NOT_FOUND` when the target is absent.

crates/goose/src/acp/server.rs
morgmart: 🤖 Fixed in the latest branch tip: session extension status now uses the typed `ExtensionConnectionStatusDto` enum (`connected`, `failed`, `available`, `unavailable`) and the server constructs enum values instead of stringly JSON.

ui/goose2/src/features/extensions/api/extensions.ts
```diff
@@ -1,29 +1,31 @@
...
 }
 
+export async function listSessionExtensions(
+  sessionId: string,
+): Promise<ExtensionConfig[]> {
+  const client = await getClient();
+  const response = await client.goose.GooseSessionExtensions({
+    sessionId,
+  });
+  return (response.extensions ?? []) as ExtensionConfig[];
```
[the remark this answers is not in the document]
morgmart: 🤖 Fixed in the latest branch tip: `listSessionExtensions` now calls `GooseSessionExtensionsStatus` and returns `SessionExtensionStatus[]`, so the widget receives real status and tool metadata from the backend.

ui/goose2/src/features/extensions/hooks/useExtensionsSettings.ts
```diff
@@ -0,0 +1,151 @@
...
+
+        if (
+          (isAdd || keyChanged) &&
+          extensions.some((extension) => extension.config_key === newKey)
+        ) {
+          toast.error(t("extensions.errors.nameConflict", { name }));
+          return;
+        }
+
+        await addExtension(name, config);
```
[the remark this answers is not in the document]
morgmart: 🤖 Fixed in the latest branch tip: the edit flow now passes `editingExtension?.enabled ?? true` into `addExtension`, so saving configuration changes preserves an existing disabled state instead of silently re-enabling the extension.

ui/goose2/src/shared/api/acpNotificationHandler.ts
```diff
@@ -52,6 +52,31 @@ const pendingUsageUpdates = new Map<
...
+
+function getToolCallIdentity(update: SessionUpdate): {
+  toolName?: string;
+  extensionName?: string;
+} {
+  if (!isRecord(update._meta)) {
+    return {};
+  }
+  const goose = update._meta.goose;
+  if (!isRecord(goose) || !isRecord(goose.toolCall)) {
```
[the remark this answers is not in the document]
morgmart: 🤖 Fixed in `9150ad196`: `getToolCallIdentity` now reads the actual `_meta.goose.mcpApp` metadata shape, keeps `_meta.goose.toolCall` as a fallback, and applies identity from `tool_call_update` in both live and replay flows. Added focused assertions in `acpNotificationHandler.test.ts` for `toolName` and `extensionName`.

PR #8929, opened 2026-04-30
fix(providers): refresh GCP metadata server token on expiration

froody:
The GCP metadata server provides a token that expires, but the original implementation of DefaultAccount in GcpAuth was caching the response of the metadata server directly in the credentials structure instead of fetching a new token when needed. This effectively created an infinite-lived default credential until Goose restarted, at which point it'd read the metadata server again. I have modified AdcCredentials::DefaultAccount to store the base URL instead of the one-time token, and get_default_access_token to actually re-fetch the token using the URL so the standard caching mechanism works properly and automatically refreshes it on expiration.

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:

### DIFF
--- a/crates/goose/src/providers/gcpauth.rs
+++ b/crates/goose/src/providers/gcpauth.rs
@@ -58,7 +58,8 @@ enum AdcCredentials {
     /// Credentials for a service account
     ServiceAccount(ServiceAccountCredentials),
     /// Credentials for the GCP native default account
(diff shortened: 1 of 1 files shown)


jessburnett (approved):
Looks great @froody. Thank you for your contribution

DOsinga (approved):

PR #8935, opened 2026-04-30
fix: keep renamed skills open in detail view

kalvinnchau:
**Category:** fix
**User Impact:** Users stay on the renamed skill's detail page after saving changes instead of being returned to the skills list.
**Problem:** Renaming a skill changed its ID, so the skills view could no longer match the previous active skill and fell back to the list page. This made a successful rename feel like it lost the user's place.
**Solution:** The editor now reports the saved skill back to the skills view, which refreshes the list and selects the saved skill ID when it is present.

<details>

**ui/goose2/src/features/skills/ui/SkillEditor.tsx**
Renamed the save callback to reflect both create and edit flows, and returns the saved skill from updates so the parent can keep the correct detail view selected.

**ui/goose2/src/features/skills/ui/SkillsDialogs.tsx**
Threaded the renamed save callback through the dialog wrapper so create and edit flows share the same post-save handling.

**ui/goose2/src/features/skills/ui/SkillsView.tsx**
Refreshes skills after a save and keeps the saved skill active when it appears in the refreshed list.

**ui/goose2/src/features/skills/ui/__tests__/SkillEditor.test.tsx**
Updated callback naming expectations to match the editor's broader save behavior.

**ui/goose2/src/features/skills/ui/__tests__/SkillsView.test.tsx**
Added coverage for renaming a skill and staying on the renamed skill's detail page.

</details>

### Reproduction Steps

1. Open the skills view and select an existing skill.
2. Click Edit, change the skill name, and save the changes.
3. Confirm the renamed skill remains open in the detail view and the editor closes.
4. Run the skills view tests to confirm the rename flow remains covered.

### DIFF
--- a/ui/goose2/src/features/skills/ui/SkillEditor.tsx
+++ b/ui/goose2/src/features/skills/ui/SkillEditor.tsx
@@ -11,21 +11,26 @@ import {
   DialogHeader,
   DialogTitle,
 } from "@/shared/ui/dialog";
-import { createSkill, updateSkill, type EditingSkill } from "../api/skills";
+import {
+  createSkill,
+  updateSkill,
+  type EditingSkill,
+  type SkillInfo,
+} from "../api/skills";
 import { formatSkillName, isValidSkillName } from "../lib/skillsHelpers";
 import { getRenamedSkillFileLocation } from "../lib/skillsPath";
 
 interface SkillEditorProps {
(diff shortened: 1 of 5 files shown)


morgmart (approved):
🤖 Reviewed locally and it looks good. Checks passed and I did not find any blocking issues.

PR #8941, opened 2026-05-01
fix(acp): seed provider handoff history

matt2e:
There were two situations where the agent would not see the conversation history:
- Continuing a thread with an ACP provider after relaunching the app (we configure ACP providers to not store session history)
- Switching a thread over to an ACP provider


baxen (approved):
This is a super clean approach, and i tested it out it works nicely

PR #8943, opened 2026-05-01
fix: slow UI when trying to start a new project session

matt2e:
Previously it took a very long time for the UI for new sessions to show up:

https://github.com/user-attachments/assets/7efe0cdd-0223-46be-9480-5e6d06b1d3c1

Fixes the delay after clicking New Session in a project by creating a local chat immediately, then binding it to ACP with the project working directory and model. Centralizes the binding flow in the chat session store and preserves local draft sessions during ACP refreshes.

Verification: pre-push checks passed (Biome, file-size/i18n checks, typecheck, build, Vitest, cargo fmt/check/clippy).


matt2e:
Converted to draft... app seems snappier now so im not sure this fix is still relevant. Will check later

PR #8947, opened 2026-05-01
feat(acp): add GOOSE_DISABLE_TOOL_CALL_SUMMARY to opt out of per-tool-call summaries

ocervinka:
## Motivation

The ACP server fires an auxiliary `provider.complete_fast` call for every tool
  invocation. Its only purpose is to upgrade the synchronous `fallback_title`
  (tool name + truncated args) into a model-generated short phrase shown in the
  ACP UI's tool-call list.

  For two real-world cases this auxiliary call is the difference between a run
  that finishes and one that gets throttled mid-session:

  1. **Code Assist OAuth** has a tight per-second cap. A single user turn that
     makes 4 tool calls fires 4 extra `complete_fast` requests on the same
     bucket, plus the existing session-naming call (when not opted out), and
     trips the 429.
  2. **Gemini API free tier** allots ~250 requests/day for Flash. Per-tool-call
     summaries triple or quadruple the request count of any non-trivial agent
     loop and exhaust the quota long before the day ends.

  Headless ACP integrations that don't surface the title at all still pay the
  cost.

  ## Change

  Add `GOOSE_DISABLE_TOOL_CALL_SUMMARY` (env var / `config.yaml` key, bool,
  default `false`). Behavior unchanged when unset. When set, the synchronous
  fallback title remains visible to the client and the `complete_fast` round
  trip is skipped.

  This mirrors the existing `GOOSE_DISABLE_SESSION_NAMING` precedent line-for-line:
  - registered via the same `config_value!` macro entry
  - gated at the call site with the same `Config::global().get_*().unwrap_or(false)` shape
  - documented in the same env-var reference table
  - defaults preserve current behavior

  ## Diff size

  3 files, +17 lines, no deletions. No public API changes.

  ## Test plan

  Manual:
  - With the var unset, every tool call still receives a `ToolCallUpdate` whose
    title is the model-generated phrase (existing behavior).
  - With `GOOSE_DISABLE_TOOL_CALL_SUMMARY=true`, only the initial `ToolCall`
    notification is sent (with the synchronous fallback title); no follow-up
    `ToolCallUpdate` for the title; no `provider.complete_fast` call appears in
    provider logs.

  No unit tests added — the existing `GOOSE_DISABLE_SESSION_NAMING` opt-out
  ships without one and follows the same pattern. Happy to add a test if a
  reviewer prefers; would likely involve a mock `Provider` asserting that
  `complete_fast` is not invoked when the flag is set.

  ## Checklist

  - [x] `cargo fmt` (formatting matches surrounding code)
  - [ ] `cargo clippy --all-targets -- -D warnings` (will run via CI)
  - [ ] `cargo test` (will run via CI)
  - [x] DCO sign-off
  - [x] Conventional Commits commit message


DOsinga (approved):
Clean, minimal change that follows the existing `GOOSE_DISABLE_SESSION_NAMING` pattern. I pushed a small cleanup commit removing the redundant inline comment (the config key name is self-documenting) and the example block in the env-var docs. LGTM.

PR #8949, opened 2026-05-01
install plugins

jamadeo:
Some minimal scaffolding for plugin install support from git. Only supports gemini format and skills so far.

To expand this further, we should look into adding:
 - open-plugins and claude marketplace formats
 - GUI install feature
 - updating


michaelneale (approved):
I think ok - I did think that goose-cli crate may go away but that is more the TUI aspect of it, not the install I guess?

jamadeo:
> I think ok - I did think that goose-cli crate may go away but that is more the TUI aspect of it, not the install I guess?

yeah, the next bit for this would be to add an ACP+ handler that can do the install

Issue #8180, opened 2026-03-28
connect to remote LM Studio


erlangparasu:
**Please explain the motivation behind the feature request.**
connect to remote LM Studio instance (current version only support local instance)

**Describe the solution you'd like**
connect to remote LM Studio

**Describe alternatives you've considered**
.

**Additional context**
.

- [x] I have verified this does not duplicate an existing feature request

cetindogu:
+1 


Lm studio local asks: "Provider LM Studio requires OPENAI_HOST, please enter a value"

monroewilliams:
+1
- The LM Studio provider should have a configurable base URL, not be hardcoded to `http://localhost:1234/v1/chat/completions`.  I'd like to be able to point it at another local machine on my network, without having to run a local copy of LM Studio on localhost.
- It should also be able to fetch the model list from the endpoint, using either the OpenAI-compatible `/v1/models` endpoint, or the richer `/api/v1/models` LM Studio-specific endpoint. (The former is probably sufficient, it will give the full list of model identifiers. The latter provides friendly names, architecture, format, and other additional data.)
- Related: when configuring the LM Studio provider through the desktop Goose UI, it should not require the user to provide a list of models, since it can fetch them from the endpoint.

monroewilliams:
Looking into this just a bit further, it appears that using the llama_swap provider works perfectly with LM Studio.
With an LM Studio server running on http://othermachine.local:1234 and this in config.yaml:

```
GOOSE_PROVIDER: llama_swap
LLAMA_SWAP_HOST: http://othermachine.local:1234
```

it will correctly fetch the list of models and everything will work.

The UI in the desktop app doesn't seem to allow for this to be set up, so fixing that would be ideal, but at least there's a way to make it work.

Changing `goose/crates/goose/src/providers/declarative/lmstudio.json` to exactly mirror `llama_swap.json` in the same directory would probably be a good start. Something like:

```json
{
  "name": "lmstudio",
  "engine": "openai",
  "display_name": "LM Studio",
  "description": "Run local models with LM Studio",
  "api_key_env": "",
  "base_url": "${LMSTUDIO_HOST}/v1/chat/completions",
  "env_vars": [
    {
      "name": "LMSTUDIO_HOST",
      "required": false,
      "secret": false,
      "primary": true,
      "default": "http://localhost:1234",
      "description": "Base URL of the LMStudio server (default: http://localhost:1234)"
    }
  ],
  "dynamic_models": true,
  "models": [],
[...]

PR #8973, opened 2026-05-03
Improvements to LM Studio declarative provider

monroewilliams:
## Summary
This changes lmstudio.json to better align with the capabilities of the LM Studio server.
I used llama_swap.json as a template. The change to the lmstudio provider:
- gives the ability to set the host url via `LMSTUDIO_HOST`, with a default of `http://localhost:1234`
- sets `dynamic_models` to `true`, since the LM Studio provides an OpenAI-compliant `models` endpoint 

### Testing
I've manually tested this by pointing it at LM Studio instances running on both localhost and on a remote machine. 
It allows for dynamic querying of the installed models through both the GUI app and `goose configure` when the LM Studio provider is selected.

### Related Issues
Relates to #8180
Discussion: LINK (if any)

Before:  

After:

### DIFF
--- a/crates/goose/src/providers/declarative/lmstudio.json
+++ b/crates/goose/src/providers/declarative/lmstudio.json
@@ -4,7 +4,18 @@
   "display_name": "LM Studio",
   "description": "Run local models with LM Studio",
   "api_key_env": "",
-  "base_url": "http://localhost:1234/v1/chat/completions",
+  "base_url": "${LMSTUDIO_HOST}/v1/chat/completions",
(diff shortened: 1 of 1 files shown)


monroewilliams:
@Abhijay007 , I see you're assigned to #8180.  Does this seem reasonable?

Abhijay007:
> @Abhijay007 , I see you're assigned to #8180. Does this seem reasonable?

Hi @monroewilliams thanks for taking this up will review it soon

jh-block (approved):

PR #8978, opened 2026-05-03
fix: unscheduling a recipe should not delete them

angiejones:
Unscheduling a recipe also deletes the underlying recipe file, making the workflow unrecoverable.

This fix only removes the schedule.

Added a test as well.

### DIFF
--- a/crates/goose-cli/src/commands/schedule.rs
+++ b/crates/goose-cli/src/commands/schedule.rs
@@ -180,10 +180,10 @@ pub async fn handle_schedule_remove(schedule_id: String) -> Result<()> {
         .await
         .context("Failed to initialize scheduler")?;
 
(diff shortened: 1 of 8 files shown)


michaelneale (approved):
I think makes sense - seems odd that it would also delete recipe (maybe that was by design but is surprising which I assume is why you want to change this, so makes sense to change behavior to be less surprising!)

lifeizhou-ap (approved):
Thank you @angiejones for the fix and making the behaviour consistent!

angiejones:
> I think makes sense - seems odd that it would also delete recipe (maybe that was by design but is surprising which I assume is why you want to change this, so makes sense to change behavior to be less surprising!)

yeah, bit me twice 😭 figured I should put in a fix

PR #8989, opened 2026-05-04
fix(extension-manager): require extension_name on read_resource

### LINKED ISSUE #8988: read_resource: optional extension_name causes silent probe-all across extensions
**Describe the bug**

The `read_resource` tool the Extension Manager exposes to the model treats `extension_name` as optional, and when the model omits it, the handler iterates every resource-capable extension, swallows every error, and returns the first match. This produces silent failures on URI collisions and silent loss of distinguishable errors (auth, transport, not-found, server-side). The tool description actively encourages the model to take this path.

A related but separate bug: the `list_resources` JSON schema declares `extension_name`, but the handler reads `params.get("extension")` — model calls passing `extension_name` are silently ignored and fall into the same fan-out.

---

**To Reproduce**

1. Connect goose to two MCP servers that both expose a resource at the same URI (e.g. two filesystem-style servers both exposing `file:///some/path`, or two skill-style servers both exposing `skill://example/SKILL.md`).
2. Start a session and ask the model to read that URI.
3. Observe that the model calls `read_resource` with `uri` only (the tool description encourages this), and goose silently returns whichever extension happens to be hit first by the iteration order, with no indication that there was ambiguity.

To reproduce the related `list_resources` mismatch:

1. Connect goose to multiple resource-capable extensions.
2. Have the model call `list_resources` with `{"extension_name": "<known>"}`.
3. Observe that the result is the multi-extension fan-out, not the filtered per-extension list. The `extension_name` argument was silently ignored because the handler reads `extension`.

---

**Expected behavior**

- `read_resource` should require `extension_name`. The model already has the (extension, uri) pair available from `list_resources`, whose output at `extension_manager.rs:1482` is formatted `extension - name, uri: (uri)`.
- When the extension is unknown, surface the existing `"Extension '{}' not found. Here are the available extensions: …"` error (already at `extension_manager.rs:1394-1402`).
- The fan-out loop and `Err(_) => continue` should be removed entirely; the in-code TODO about URI collisions becomes moot.
- `list_resources`'s handler should read `extension_name` to match its declared schema. The field stays optional (list-all is the natural default for discovery), and only the parameter name needs to be reconciled.

---

**Screenshots**

N/A

---

- **OS & Arch:** Not platform-specific (model-facing tool contract)
- **Interface:** Both UI and CLI affected (Extension Manager is shared)
- **Version:** `goose` crate `1.33.0` (current `main`, branched at `45d8bf81d`)
- **Extensions enabled:** Reproducible with any combination of two or more resource-capable extensions
- **Provider & Model:** Provider-agnostic; reproduces with any model that uses the Extension Manager's `read_resource` / `list_resources` tools

---

**Additional context**

### Proposed fix

1. Make `ReadResourceParams.extension_name` required.
2. Rewrite the `read_resource` tool description to direct the model to call `list_resources` first when ownership is unknown.
3. Remove the fan-out fallback in `read_resource_tool`; rely on the existing `read_resource()` helper's error path.
4. Fix the `list_resources` param-name mismatch (`extension` → `extension_name`).

### Tradeoff considered

If the model has a bare URI from outside (chat paste, memory recall, etc.) it can't call `read_resource` directly — it has to call `list_resources` first to find the owning extension. That's one extra round-trip in a rare (?) case, in exchange for retiring the silent probe path in the common case.

olaservo:
## Summary

Makes `extension_name` required on the model-facing `read_resource` tool and drops the "probe every extension and swallow errors" fallback. Also fixes a separate bug where `list_resources`'s schema declares `extension_name` but the handler reads `extension`, silently ignoring the model's filter argument.

Two files touched:

- `crates/goose/src/agents/platform_extensions/ext_manager.rs` — `ReadResourceParams.extension_name: String` (was `Option<String>`); rewritten tool description directs the model to call `list_resources` first when ownership is unknown.
- `crates/goose/src/agents/extension_manager.rs` — `read_resource_tool` simplified from ~80 lines to ~20 (removes the fan-out loop, error swallowing, and the in-code TODO about URI collisions); `list_resources` now reads `params.get("extension_name")` to match its schema.

The existing `read_resource()` helper already returns a clear `"Extension '{}' not found. Here are the available extensions: …"` error, so error reporting actually improves — distinguishable failures (auth, transport, server-side) are no longer swallowed.

### Testing

- `cargo check -p goose` — clean.
- `cargo test -p goose --lib` — 1182 of 1189 pass. The 7 failures all reproduce on `main` (Windows-only path-separator and snapshot drift, none in code paths this PR touches).
- `cargo clippy -p goose --lib -- -D warnings` — clean for lib changes.
- `cargo fmt` — applied.
- Manual: not run end-to-end against a live LLM yet.

### Related Issues

Fixes #8988

N/A — model-facing tool contract change, no UI involved.

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -1304,76 +1304,19 @@ impl ExtensionManager {
         cancellation_token: CancellationToken,
     ) -> Result<Vec<Content>, ErrorData> {
         let uri = require_str_parameter(&params, "uri")?;
+        let extension_name = require_str_parameter(&params, "extension_name")?;
 
-        let extension_name = params.get("extension_name").and_then(|v| v.as_str());
+        let read_result = self
+            .read_resource(session_id, uri, extension_name, cancellation_token)
+            .await?;
 
-        // If extension name is provided, we can just look it up
-        if let Some(ext_name) = extension_name {
-            let read_result = self
-                .read_resource(session_id, uri, ext_name, cancellation_token.clone())
-                .await?;
-
-            let mut result = Vec::new();
-            for content in read_result.contents {
-                if let ResourceContents::TextResourceContents { text, .. } = content {
-                    let content_str = format!("{}\n\n{}", uri, text);
-                    result.push(Content::text(content_str));
-                }
-            }
-            return Ok(result);
-        }
-
-        // If extension name is not provided, we need to search for the resource across all extensions
-        // Loop through each extension and try to read the resource, don't raise an error if the resource is not found
-        // TODO: do we want to find if a provided uri is in multiple extensions?
-        // currently it will return the first match and skip any others
-        let extension_names: Vec<String> = self
-            .extensions
-            .lock()
-            .await
-            .iter()
-            .filter(|(_name, ext)| ext.supports_resources())
-            .map(|(name, _)| name.clone())
-            .collect();
-
-        for extension_name in extension_names {
-            let read_result = self
-                .read_resource(session_id, uri, &extension_name, cancellation_token.clone())
-                .await;
-            match read_result {
-                Ok(read_result) => {
-                    let mut result = Vec::new();
-                    for content in read_result.contents {
-                        if let ResourceContents::TextResourceContents { text, .. } = content {
-                            let content_str = format!("{}\n\n{}", uri, text);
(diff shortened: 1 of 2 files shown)


alexhancock (approved):
Tested this and verified

Thank you @olaservo !

PR #8999, opened 2026-05-04
fix: 8531 - elicitation fixes

### LINKED ISSUE #8531: User elicitation does not function correctly with custom MCP server
**Describe the bug**

User elicitation is not working when using a custom MCP server based on FastMCP v3.2.0.

The functions are simple and taken directly from FastMCP documentation: https://gofastmcp.com/servers/elicitation

Full code of the server.py:
```
from fastmcp import FastMCP, Context

mcp = FastMCP("Test MCP")
    
@mcp.tool
async def approve_action(ctx: Context) -> str:
    """Example tool that elicits approval from the user."""
    result = await ctx.elicit("Approve this action?", response_type=None)

    if result.action == "accept":
        return "Hello, World!"
    else:
        raise ValueError("Action rejected")

@mcp.tool()
async def set_priority(ctx: Context) -> str:
    """Example tool that elicits a priority level from the user."""
    result = await ctx.elicit(
        "What priority level?",
        response_type=["low", "medium", "high"],
    )

    if result.action == "accept":
        return f"Priority set to: {result.data}"

if __name__ == "__main__":
    mcp.run()
```

**Example 1: Action approval**
No buttons appear in the UI so user cannot accept or reject a request and has to wait until it times out:
Image

**Example 2: Setting a value**
In this case values and button do appear:
Image

**Issue 1:**
The first value in the list is not selected by default. When clicking "Submit", the error "This field is required" is shown:
Image

To select "low", the user must first choose another value (e.g., "medium") and then switch back to "low".

**Issue 2:**
After selecting the value, a "Failed to Load Session" error appears, and it is no longer possible to reopen the chat:
Image

- **OS & Arch:** Microsoft Windows 11 Home v10.0.26200
- **Interface:** UI
- **Version:** v1.30.0
- **Extensions enabled:** Apps, Developer, Extension Manager, Test MCP
- **Provider & Model:** LM Studio – qwen3.5-9b

alexhancock:
Fixes #8531 

Fixes three elicitation bugs surfaced by FastMCP servers (#8531): elicitations with no schema (`response_type=None`) now render an Accept button instead of a dead form, and required enum fields default to their first option so users aren't blocked by a spurious "required" error. Also fixes the "Failed to Load Session" crash by routing elicitation responses through the agent's short-circuit path instead of opening a second SSE stream that collides with the in-flight tool call.

Demo:

Screenshot 2026-05-04 at 3 27 34 PM

### DIFF
--- a/crates/goose-cli/src/session/elicitation.rs
+++ b/crates/goose-cli/src/session/elicitation.rs
@@ -11,9 +11,25 @@ pub fn collect_elicitation_input(
         println!("\n{}", style(message).cyan());
     }
 
-    let properties = match schema.get("properties").and_then(|p| p.as_object()) {
-        Some(props) => props,
-        None => return Ok(Some(HashMap::new())),
+    let properties = schema.get("properties").and_then(|p| p.as_object());
+
+    // Schema-less (or empty-schema) elicitations are pure approval prompts —
+    // offer an explicit Y/N confirmation instead of silently auto-accepting.
+    let properties = match properties {
+        Some(props) if !props.is_empty() => props,
+        _ => {
+            let prompt = if message.is_empty() {
+                "Approve this action?"
+            } else {
+                "Approve?"
+            };
+            return match cliclack::confirm(prompt).initial_value(true).interact() {
+                Ok(true) => Ok(Some(HashMap::new())),
+                Ok(false) => Ok(None),
+                Err(e) if e.kind() == io::ErrorKind::Interrupted => Ok(None),
+                Err(e) => Err(e),
+            };
+        }
     };
 
     let required: Vec<&str> = schema
--- a/crates/goose-server/src/routes/session_events.rs
+++ b/crates/goose-server/src/routes/session_events.rs
@@ -313,6 +313,38 @@ pub async fn session_reply(
         }
     }
 
+    let user_message = request.user_message;
+    let override_conversation = request.override_conversation;
+
+    // An elicitation response unblocks an in-flight tool call that is already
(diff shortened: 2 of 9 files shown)


jh-block:
The Codex comment seems legitimate.

Also, is decline/cancel out of scope? if the schema has no fields the desktop UI [only renders](https://github.com/aaif-goose/goose/blob/fe16fc120255fcec47c15311fc28780d4891b2fa/ui/desktop/src/components/ElicitationRequest.tsx#L169-L173) 'Accept', and the MCP client [maps all responses](https://github.com/aaif-goose/goose/blob/fe16fc120255fcec47c15311fc28780d4891b2fa/crates/goose/src/agents/mcp_client.rs#L389-L394) to `ElicitationAction::Accept`. The CLI has a [negative path in the UI](https://github.com/aaif-goose/goose/blob/fe16fc120255fcec47c15311fc28780d4891b2fa/crates/goose-cli/src/session/elicitation.rs#L26-L29) but it doesn't send the negative action, it just cancels locally.

alexhancock:
@jh-block Good callouts. I see the issues, and will push a rev tomorrow morning.

alexhancock:
@jh-block I fixed the codex one, and goose suggested we keep this scoped to fixing https://github.com/aaif-goose/goose/issues/8531 and filed #9024 to track the other two response actions

jh-block (approved):

teknovizier:
I believe the UI should also include a reject button as there is no explicit way to decline the request. The user must wait for the request to time out instead.

PR #9004, opened 2026-05-04
support file-backed agent and skill editing

morgmart:
**Category:** improvement
**User Impact:** Users can view, edit, duplicate, import, and share file-backed agents and skills from the Goose desktop app without losing context while pages refresh.
**Problem:** Agents and skills were drifting between editable app objects and files on disk, which made file-backed agents feel incomplete and made the Skills page visually reload when users navigated back to it. The related menus and dialogs also needed enough shared structure to support the larger editing surface without awkward overflow or stuck sibling menus.
**Solution:** This PR makes agents and skills first-class file-backed objects in Goose2: Markdown-backed agents can be created, imported, edited, copied, and saved; skills get matching detail/share/duplicate flows; and the Skills view keeps cached content visible while refreshing. It also tightens shared dialog/dropdown primitives and normalizes page-entry motion across Agents, Skills, and Extensions.

<details>

**crates/goose-sdk/src/custom_requests.rs**
Updates the custom request shape used by the app/backend contract so agent file metadata can flow through the UI correctly.

**crates/goose/src/sources.rs**
Adjusts source handling for the new file-backed agent/skill behavior and removes older assumptions that no longer fit Markdown-backed records.

**ui/goose2/src-tauri/src/commands/agents.rs**
Adds Markdown agent import/parsing support and wires agent commands into the file-backed persona service.

**ui/goose2/src-tauri/src/commands/system.rs**
Adds system file actions used by the UI, including copying files to the clipboard and saving file copies.

**ui/goose2/src-tauri/src/commands/system_tests.rs**
Adds coverage for the new system file command behavior and path validation.

**ui/goose2/src-tauri/src/lib.rs**
Registers the new/updated Tauri commands with the Goose2 app.

**ui/goose2/src-tauri/src/services/personas.rs**
Moves persona persistence toward Markdown-backed agents, including migration, loading, update, import, and file-path handling.

**ui/goose2/src-tauri/src/services/personas_tests.rs**
Adds service coverage for Markdown persona paths, migration behavior, and local avatar reference handling.

**ui/goose2/src-tauri/src/types/agents.rs**
Extends agent/persona types with file metadata needed by the frontend and command layer.

**ui/goose2/src-tauri/src/types/builtin_personas.rs**
Updates built-in persona definitions to fit the new file-backed type model.

**ui/goose2/src/app/AppShell.tsx**
Passes agent start-chat behavior through the app shell so agent detail actions can create chats.

**ui/goose2/src/app/ui/AppShellContent.tsx**
Connects the Agents page route to the new start-chat affordance.

**ui/goose2/src/features/agents/lib/personaImport.ts**
Changes agent import validation from JSON exports to Markdown agent files.

**ui/goose2/src/features/agents/lib/personaPresentation.ts**
Simplifies persona presentation helpers around display initials now that source/read-only handling lives elsewhere.

**ui/goose2/src/features/agents/stores/__tests__/agentStore.test.ts**
Removes store tests for source buckets that no longer exist in the new model.

**ui/goose2/src/features/agents/stores/agentStore.ts**
Removes built-in/custom persona selector helpers that were tied to the old source model.

**ui/goose2/src/features/agents/ui/AgentConfig.tsx**
Simplifies persona labels in agent configuration so source labels do not leak into the picker.

**ui/goose2/src/features/agents/ui/AgentDetailPage.tsx**
Adds the agent detail view with edit, start-chat, share, duplicate, and delete actions for file-backed agents.

**ui/goose2/src/features/agents/ui/AgentsView.tsx**
Reworks the Agents page around list/detail navigation, Markdown import, file actions, duplicate/delete handling, and start-chat integration.

**ui/goose2/src/features/agents/ui/PersonaCard.tsx**
Adds card actions for file-backed agents and removes the extra per-card intro animation so page transitions feel consistent.

**ui/goose2/src/features/agents/ui/PersonaDetails.tsx**
Updates persona detail rendering to align with the new detail-page and file-backed metadata model.

**ui/goose2/src/features/agents/ui/PersonaEditor.tsx**
Updates the editor to create and update file-backed personas, including optional provider/model handling.

**ui/goose2/src/features/agents/ui/PersonaGallery.tsx**
Updates the gallery to support the new card actions and empty/loading states.

**ui/goose2/src/features/agents/ui/__tests__/AgentDetailPage.test.tsx**
Adds coverage for agent detail primary actions and share-menu file actions.

**ui/goose2/src/features/agents/ui/__tests__/PersonaCard.test.tsx**
Updates card tests for the new actions and presentation behavior.

**ui/goose2/src/features/chat/lib/newChat.test.ts**
Adds coverage for starting chats from an agent/persona selection.

**ui/goose2/src/features/chat/lib/newChat.ts**
Updates new-chat creation to respect the selected agent/persona path.

**ui/goose2/src/features/chat/ui/widgets/WorkspaceCreateDialog.tsx**
Adopts the shared dialog body/width structure so larger forms scroll correctly.

**ui/goose2/src/features/extensions/ui/ExtensionModal.tsx**
Adopts the shared dialog width/body changes for consistent modal behavior.

**ui/goose2/src/features/projects/ui/CreateProjectDialog.tsx**
Adopts the shared dialog body/width structure for project creation.

**ui/goose2/src/features/providers/ui/CustomProviderDialog.tsx**
Adopts the shared dialog width/body changes for provider configuration.

**ui/goose2/src/features/skills/api/skills.ts**
Updates skill import/export expectations for canonical Markdown skill files.

**ui/goose2/src/features/skills/hooks/useSkillImportExport.ts**
Adds shared skill file copy/save/import behavior used by the list and detail views.

**ui/goose2/src/features/skills/lib/skillsHelpers.ts**
Removes helper behavior that moved into the view/editor flow.

**ui/goose2/src/features/skills/ui/SkillDetailPage.tsx**
Adds file actions, duplicate/delete/start-chat affordances, and updated detail presentation for skills.

**ui/goose2/src/features/skills/ui/SkillEditor.tsx**
Updates the skill editor for Markdown-backed skill editing and consistent dialog scrolling.

**ui/goose2/src/features/skills/ui/SkillsListSections.tsx**
Adds per-skill row actions and removes the inner accordion fade so collapsing sections do not ghost content before height changes.

**ui/goose2/src/features/skills/ui/SkillsView.tsx**
Adds duplicate/detail/file-action flows and caches loaded skills so returning to the Skills page shows content immediately while refreshing.

**ui/goose2/src/features/skills/ui/__tests__/SkillsView.test.tsx**
Updates Skills page tests for row actions, detail navigation, refresh behavior, and the new local cache.

**ui/goose2/src/shared/api/__tests__/agents.test.ts**
Updates agent API tests for the new file-backed command shape.

**ui/goose2/src/shared/api/agents.ts**
Updates frontend agent API helpers to match the new backend contract.

**ui/goose2/src/shared/api/system.ts**
Adds frontend wrappers for file copy/save system commands.

**ui/goose2/src/shared/i18n/locales/en/agents.json**
Adds English copy for agent detail, Markdown import, file actions, and updated menu labels.

**ui/goose2/src/shared/i18n/locales/en/skills.json**
Adds English copy for skill detail/file actions and updated import/export labels.

**ui/goose2/src/shared/i18n/locales/es/agents.json**
Adds Spanish strings corresponding to the updated agent flows.

**ui/goose2/src/shared/i18n/locales/es/skills.json**
Adds Spanish strings corresponding to the updated skill flows.

**ui/goose2/src/shared/types/agents.ts**
Extends shared agent/persona types with file metadata used across the UI.

**ui/goose2/src/shared/ui/badge.tsx**
Adds a badge variant needed by the updated detail/list surfaces.

**ui/goose2/src/shared/ui/dialog.tsx**
Adds dialog body support and width variants so complex dialogs can keep fixed headers/footers with scrollable content.

**ui/goose2/src/shared/ui/dropdown-menu.test.tsx**
Adds tests for sibling dropdown behavior and submenu icon treatment.

**ui/goose2/src/shared/ui/dropdown-menu.tsx**
Updates dropdown defaults and submenu item styling so adjacent menus close predictably and submenu rows match regular menu rows.

**ui/goose2/src/test/setup.ts**
Updates test setup for the new browser/API behavior exercised by this branch.

**ui/goose2/tests/e2e/fixtures/tauri-mock.ts**
Updates the Tauri test fixture for the new commands and file-backed agent/skill data.

</details>

### Reproduction Steps

1. Open Goose2 and navigate to Agents.
2. Import or select a file-backed Markdown agent, then confirm the detail page shows Edit, Start a chat, Share, More, duplicate, and delete actions.
3. Use the Share menu to copy the agent file or export a copy.
4. Navigate to Skills, open a skill detail page, edit or duplicate it, and confirm returning to Skills keeps the existing list visible while it refreshes.
5. Click quickly between Agents, Skills, and Extensions and confirm the pages use a consistent page-entry feel without extra card or accordion intro animations.

### Screenshots/Demos

Not attached. This PR is best reviewed by exercising the Agents and Skills flows above because the changes are primarily interaction and file-backed behavior.

### Verification

- `pnpm test src/features/skills/ui/__tests__/SkillsView.test.tsx`
- `pnpm typecheck`
- Pre-push hook passed: `pnpm check`, `pnpm build`, `pnpm test` (109 files, 642 tests), `cargo fmt --check`, `cargo check`, and `cargo clippy -- -D warnings`

Screenshot 2026-05-04 at 2 59 50 PM

### 🤖 Review follow-up summary

**Addressed from review**
- Validated share/copy paths and avoided raw path interpolation for clipboard commands.
- Hid Copy file on Linux while keeping Save a copy available.
- Added provider fallback when persona provider lookup misses.
- Cleared stale active agent when a persona has no agent mapping.
- Froze PersonaEditor state while open so refreshes do not wipe edits.
- Preserved unknown Markdown frontmatter on persona updates.
- Added stale-skills inline banner with Retry.
- Added provider/model null round-trip coverage.
- Renamed export response field from `json` to `contents`.
- Allowed plugin and nested skill files in share validation.
- Preserved project scope when duplicating project skills.

**Intentionally held for follow-up**
- Legacy JSON import compatibility: not needed before Goose2 launch.
- Atomic writes/partial-file hardening: defer to backend/storage refactor.
- Persona filename/id rename behavior: keep stable identity for now; revisit with storage model.
- Persona mutex/file-I/O restructuring: defer to backend/storage concurrency work.


crates/goose/src/sources.rs
```diff
@@ -180,80 +179,27 @@ pub fn export_source(source_type: SourceType, path: &str) -> Result<(String, Str
...
-    let description = value
-        .get("description")
-        .and_then(|v| v.as_str())
-        .ok_or_else(|| Error::invalid_params().data("Missing or invalid \"description\" field"))?
-        .to_string();
-    if description.is_empty() {
+    let (metadata, content): (SkillFrontmatter, String) = parse_frontmatter(data)
+        .map_err(|e| Error::invalid_params().data(format!("Invalid skill frontmatter: {e}")))?
+        .ok_or_else(|| Error::invalid_params().data("Missing skill frontmatter"))?;
+
```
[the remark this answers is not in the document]
morgmart: the app hasn't launched yet, so backwards compatibility not needed

ui/goose2/src-tauri/src/commands/agents.rs
```diff
@@ -111,153 +111,73 @@ pub fn read_import_persona_file(source_path: String) -> Result<ImportFileReadRes
...
     store: State<'_, PersonaStore>,
     file_bytes: Vec<u8>,
     file_name: String,
 ) -> Result<Vec<Persona>, String> {
-    // Validate file extension
-    if !file_name.ends_with(".persona.json") && !file_name.ends_with(".json") {
-        return Err("Unsupported file type. Expected a .persona.json or .json file.".to_string());
+    if !file_name.to_lowercase().ends_with(".md") {
+        return Err("Unsupported file type. Expected a .md file.".to_string());
     }
```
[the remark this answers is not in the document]
morgmart: the app hasn't launched yet, so no need for backwards compatibility

ui/goose2/src-tauri/src/services/personas.rs
```diff
@@ -2,31 +2,37 @@ use crate::types::agents::{
...
 
 pub struct PersonaStore {
     personas: Mutex<Vec<Persona>>,
-    store_path: PathBuf,
 }
 
 /// YAML frontmatter fields parsed from markdown persona files.
-#[derive(serde::Deserialize)]
+#[derive(Clone, Default, Deserialize, Serialize)]
 struct MarkdownFrontmatter {
```
kalvinnchau, line 16: This update path does not preserve the original Markdown frontmatter shape. `MarkdownFrontmatter` only models `name`, `description`, `avatar`, `provider`, and `model`, and `persona_to_frontmatter` writes `description: None`, so editing a file-backed agent through the UI silently drops unknown YAML keys and can also drop `description`.

For file-backed agents, round-trip safety is part of the contract. Please preserve unknown frontmatter fields, for example with a flattened `IndexMap<String, serde_yaml::Value>`, and carry the existing `description` through unless the user explicitly changes/removes it.

---
*ai generated*

ui/goose2/src/features/agents/ui/PersonaEditor.tsx
```diff
@@ -164,13 +153,21 @@ export function PersonaEditor({
       e.preventDefault();
       if (!isValid || isReadOnly) return;
 
-      const data: CreatePersonaRequest | UpdatePersonaRequest = {
-        displayName: displayName.trim(),
-        avatar: avatar ?? undefined,
-        systemPrompt: systemPrompt.trim(),
```
kalvinnchau, line 170: The editor still reinitializes form state from the selected persona object while the dialog is open, and this edit path now makes file-backed personas mutable from the UI. That makes the form vulnerable to losing in-progress edits if the selected persona object is replaced while the dialog is open.

Please freeze the initial edit snapshot when the dialog opens, key the reset effect on `isOpen` + `persona?.id` rather than the whole object, and consider dirty-state protection for external refreshes.

---
*ai generated*

ui/goose2/src-tauri/src/commands/system.rs
```diff
@@ -78,6 +80,126 @@ pub async fn save_exported_session_file(
     Ok(Some(path.to_string_lossy().into_owned()))
 }
```
kalvinnchau, line 82: `copy_file_to_clipboard` and `save_file_copy` accept renderer-supplied paths and only validate that the path is a file. That turns these commands into generic arbitrary-file copy/read primitives from JS, even though the UI only exposes them for agent/skill files. `copy_file_to_clipboard` shells out to osascript/powershell with a JS-supplied path. Escape is fragile (CR/LF on macOS; only ' doubled on Windows). `save_file_copy` accepts any readable path, should reject outside allowed roots.

Please canonicalize the source path and restrict it to expected roots, such as `~/.agents/**`, `~/.goose/**`, or project `.agents/**` paths. The platform shell-outs should also avoid interpolating raw paths into command strings where possible; prefer native APIs, literal-path APIs, or encoded command arguments.

---
*ai generated*

ui/goose2/src-tauri/src/services/personas_tests.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    assert!(PersonaStore::is_local_avatar_referenced(
+        "shared.png",
+        &personas
+    ));
+    assert!(!PersonaStore::is_local_avatar_referenced(
+        "missing.png",
+        &personas
+    ));
+}
```
kalvinnchau, line 61: The nullable provider/model deserialization is tested at the DTO layer, but the end-to-end file-backed behavior is not. The important regression case is: update an existing Markdown persona with `provider: null` / `model: null`, write the file, read it back, and confirm the YAML keys are cleared and the parsed persona has `None`.

Please add a `personas_tests.rs` round-trip test through the store/update path so the optional field persistence fix is covered at the file boundary.

---
*ai generated*

ui/goose2/src-tauri/src/services/personas.rs
```diff
@@ -186,17 +294,155 @@ impl PersonaStore {
...
+    fn update_markdown_persona_file(
+        id: &str,
+        req: &UpdatePersonaRequest,
+    ) -> Result<Persona, String> {
+        let path = Self::markdown_persona_path(id)?;
+        let content =
+            std::fs::read_to_string(&path).map_err(|e| format!("Failed to read file: {}", e))?;
+        let (yaml_str, current_body) = Self::split_markdown_persona(&content)?;
+
+        let original_frontmatter: MarkdownFrontmatter = serde_yaml::from_str(yaml_str)
```
kalvinnchau, line 417: Renaming a Markdown persona updates the `name:` frontmatter but leaves the filename, `source_path`, and id derived from the old slug. After a rename, the visible name and persisted identity diverge, and subsequent operations still key off the stale filename.

Please either make this stable-id behavior explicit in the product/API contract, or rename the file atomically to a unique slug for the new name and return the updated id/path to the frontend so the store can resync.

---
*ai generated*

ui/goose2/src-tauri/src/services/personas.rs
```diff
@@ -256,22 +486,45 @@ impl PersonaStore {
...
             avatar: req.avatar,
             system_prompt: req.system_prompt,
             provider: req.provider,
             model: req.model,
             is_builtin: false,
-            is_from_disk: false,
+            is_from_disk: true,
+            source_path: None,
             created_at: now.clone(),
             updated_at: now,
```
kalvinnchau, line 511: The create/import paths use "pick unique path, then write" and direct `fs::write`. Two concurrent creates can race on the same path, and a crash during write can leave a truncated Markdown file that disappears on the next parse. The skill import path has a similar partial-create issue where the directory can be created before `SKILL.md` is written.

Please use atomic creation/writes: reserve paths with `OpenOptions::create_new`, write to a temp file, then rename into place. For skill imports, clean up the directory if writing `SKILL.md` fails.

---
*ai generated*

ui/goose2/src-tauri/src/services/personas.rs
```diff
@@ -282,11 +535,10 @@ impl PersonaStore {
             .find(|p| p.id == id)
             .ok_or_else(|| format!("Persona '{}' not found", id))?;
```
kalvinnchau, line 547: `update` and `delete` hold the `personas` mutex while doing blocking filesystem work. A slow or blocked disk operation can prevent unrelated `list_personas`/refresh operations from reading the in-memory list.

Please snapshot the needed persona data under the lock, drop the guard while doing file I/O, then reacquire the lock to commit the updated list. An `RwLock` would also fit this read-heavy store better.

---
*ai generated*

crates/goose-sdk/src/custom_requests.rs
```diff
@@ -755,11 +755,12 @@ pub struct ExportSourceRequest {
 #[derive(Debug, Default, Clone, Serialize, Deserialize, JsonSchema, JsonRpcResponse)]
 #[serde(rename_all = "camelCase")]
 pub struct ExportSourceResponse {
+    /// Canonical source contents. The field remains `json` for wire compatibility.
```
kalvinnchau: `ExportSourceResponse.json` now carries canonical Markdown, not JSON. The comment explains this as wire compatibility, but the generated frontend already aliases the field to `contents`, and keeping the wire field named `json` makes the SDK contract misleading.

Please either rename the response field to `contents` and regenerate the SDK, or keep backward compatibility explicitly by adding a new `contents` field while deprecating `json`.

---
*ai generated*

ui/goose2/src/features/skills/ui/SkillsView.tsx
```diff
@@ -58,22 +108,23 @@ export function SkillsView({ onStartChatWithSkill }: SkillsViewProps) {
...
     try {
-      const projectDirs = projects.flatMap((project) => project.workingDirs);
       const result = await listSkills(projectDirs);
       if (loadRequestIdRef.current !== requestId) {
         return [];
       }
       const nextSkills = withInferredSkillCategories(
         hydrateProjectNames(result, projects),
       );
+      cachedProjectDirsKey = requestProjectDirsKey;
```
kalvinnchau, line 166: The cached skills behavior keeps old data visible after refresh failure, but there is no persistent stale/error indicator. A dismissed toast leaves the page looking healthy while users may be acting on stale skill data, especially after project changes or mutations.

Please track refresh error state separately and render a small inline stale-data banner with retry. Mutation flows should also avoid showing success if the follow-up refresh fails or should update local cache directly.

---
*ai generated*

ui/goose2/src-tauri/src/commands/system.rs
morgmart: 🤖 Addressed this one. The file copy/save commands now canonicalize the renderer-supplied path and only allow shareable agent/skill files instead of any readable file. I also changed the macOS/Windows clipboard shell-outs to pass the path as a command argument rather than interpolating it into the script string, and the UI now hides “Copy file” on Linux while leaving “Export a copy” available.

ui/goose2/src/app/AppShell.tsx
```diff
@@ -264,18 +265,36 @@ export function AppShell({ children }: { children?: React.ReactNode }) {
...
+              provider.label
+                .toLowerCase()
+                .includes(persona.provider?.toLowerCase() ?? ""),
+          )
+        : undefined;
       const providerId =
-        project?.preferredProvider ?? agentStore.selectedProvider ?? "goose";
+        personaProvider?.id ??
+        persona?.provider ??
+        project?.preferredProvider ??
```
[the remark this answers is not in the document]
morgmart: 🤖 Addressed this one. Persona provider values are now resolved through a shared matcher and only used when they map to an installed provider id/label. If the imported/edited persona has stale or differently formatted provider text that does not resolve, chat creation falls back to the normal project/default/global provider path instead of passing the raw string into session preparation. I also reused the same helper in the in-chat persona switch flow and added focused unit coverage.

ui/goose2/src/features/agents/ui/PersonaEditor.tsx
morgmart: 🤖 Addressed this one. The persona editor now initializes from a snapshot only when the dialog opens, the mode changes, or the selected persona id changes. Background refreshes that replace the same persona object no longer re-seed the form and wipe in-progress edits. I also added focused component coverage for same-persona refresh, switching personas, and close/reopen reset behavior.

ui/goose2/src-tauri/src/services/personas.rs
morgmart: 🤖 Addressed the unknown-frontmatter part of this. `MarkdownFrontmatter` now carries extra YAML fields via a flattened map, so when Goose rewrites a Markdown-backed agent it updates the fields the UI owns while preserving custom metadata keys it does not understand. I also added focused coverage showing custom fields like tags/owner/experimental survive serialization while Goose-controlled fields can still change.

ui/goose2/src/features/skills/ui/SkillsView.tsx
morgmart: 🤖 Addressed this. The skills view now keeps cached skills visible but tracks refresh failures separately and renders a persistent inline stale-data banner with Retry. A successful retry refreshes the list and clears the banner. I also made import/delete/duplicate/save avoid showing success if the follow-up refresh fails, so users are not told the page is current when it is actually showing last-loaded data. Added focused SkillsView coverage for the stale banner and retry behavior.

ui/goose2/src-tauri/src/services/personas_tests.rs
morgmart: 🤖 Addressed this. I added a file-boundary round-trip test that starts from an existing Markdown-backed persona with `provider` and `model`, clears both through the store update path, verifies the rewritten Markdown no longer contains those YAML keys, then parses the file back and confirms both values load as `None`. I also adjusted the file-backed update path to use the persona’s actual `source_path` when available so the test exercises the real update flow safely against a temp file.

ui/goose2/src-tauri/src/services/personas.rs
morgmart: 🤖 We are intentionally deferring this one for now. The risk is real, but there is likely backend/storage refactoring coming soon, and we do not want to pre-empt that architecture with a partial atomic-write implementation in this PR. We are keeping this as a known hardening follow-up rather than blocking the current file-backed agent/skill editing work on it.

crates/goose-sdk/src/custom_requests.rs
morgmart: 🤖 Addressed this. Since this app has not launched yet, I renamed the export response field from `json` to `contents` instead of preserving the misleading wire contract. The server now returns `{ contents, filename }`, the ACP schema and generated TypeScript SDK were regenerated, the Goose2 skills export path reads `response.contents`, and the local e2e mock was updated to match. I committed the full set of outstanding review-feedback fixes in `3a77265f9`.

ui/goose2/src-tauri/src/services/personas.rs
morgmart: 🤖 We are intentionally deferring this one for now. The concern is valid hardening, but it belongs with the upcoming backend/storage refactor rather than as a narrow mutex restructuring in this PR. Changing the lock/file-I/O boundary safely also means deciding the store’s concurrency and race semantics, so we do not want to pre-empt that architecture while this PR is focused on file-backed agent/skill editing behavior.

ui/goose2/src-tauri/src/services/personas.rs
morgmart: 🤖 We are intentionally keeping display name and file identity separate for this PR. The editable name is the user-facing label, while the filename-derived id/path remains stable so existing sessions and references do not churn on rename. Renaming the backing file would require deciding storage identity, reference migration, and atomic rename semantics, which is better handled with the upcoming storage/backend refactor rather than patched narrowly here.

ui/goose2/src/app/AppShell.tsx
```diff
@@ -351,6 +373,24 @@ export function AppShell({ children }: { children?: React.ReactNode }) {
...
   );
 
+  const handleStartChatWithPersona = useCallback(
+    (persona: Persona) => {
+      const matchingAgent = useAgentStore
+        .getState()
+        .agents.find((agent) => agent.personaId === persona.id);
+      if (matchingAgent) {
+        useAgentStore.getState().setActiveAgent(matchingAgent.id);
+      }
```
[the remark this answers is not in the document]
morgmart: 🤖 Addressed in 3712c9feb. When a persona has no matching agent mapping, Goose now explicitly clears the active agent instead of leaving the previous agent selected. This is applied both when starting a chat from a persona and when switching personas inside an existing chat, with focused coverage for the in-chat persona switch path.

ui/goose2/src-tauri/src/commands/system.rs
```diff
@@ -78,6 +80,169 @@ pub async fn save_exported_session_file(
...
+            || has_sequence_with_remaining(path, &[".goose", "agents"], 1))
+}
+
+fn is_shareable_skill_file(path: &Path) -> bool {
+    path.file_name().and_then(|name| name.to_str()) == Some("SKILL.md")
+        && (has_sequence_with_remaining(path, &[".agents", "skills"], 2)
+            || has_sequence_with_remaining(path, &[".goose", "skills"], 2)
+            || has_sequence_with_remaining(path, &[".claude", "skills"], 2)
+            || has_sequence_with_remaining(path, &[".config", "agents", "skills"], 2)
+            || has_sequence_with_remaining(path, &["goose", "skills"], 2))
```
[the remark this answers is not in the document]
morgmart: 🤖 Addressed in ec61ffbe5. The share validator now accepts plugin-installed skill files in the Goose plugin data directory shape, so plugin skills that appear in the Skills UI can be copied/saved through the same share flow instead of failing as “not shareable.” I also added Rust coverage for plugin skill paths. This commit also fixes the Linux CI lint failure by avoiding the unused `source` binding on platforms where file clipboard copy is unsupported.

ui/goose2/src/features/skills/ui/SkillsView.tsx
```diff
@@ -172,9 +274,25 @@ export function SkillsView({ onStartChatWithSkill }: SkillsViewProps) {
...
-  }, []);
+  const handleDuplicate = useCallback(
+    async (skill: SkillInfo) => {
+      const duplicateName = getDuplicateSkillName(
+        skill.name,
+        new Set(skills.map((currentSkill) => currentSkill.name)),
+      );
+
+      try {
+        await createSkill(duplicateName, skill.description, skill.instructions);
```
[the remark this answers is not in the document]
morgmart: 🤖 Addressed in ec61ffbe5. Duplicating a project-scoped skill now passes the originating project working directory into source creation, so the duplicate stays in the same project section instead of being created under Personal and disappearing from the user’s current context. I added focused SkillsView coverage for this path.

ui/goose2/src-tauri/src/commands/system.rs
```diff
@@ -78,6 +80,182 @@ pub async fn save_exported_session_file(
...
+    let components = path_components(path);
+    components
+        .windows(sequence.len())
+        .enumerate()
+        .any(|(index, window)| {
+            window
+                .iter()
+                .map(String::as_str)
+                .eq(sequence.iter().copied())
+                && components.len() == index + sequence.len() + remaining
```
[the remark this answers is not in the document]
morgmart: 🤖 Addressed in a8a3de577. The share validator now mirrors skill discovery more closely: skill roots can contain nested directories, so paths like `.agents/skills/team/review/SKILL.md` and plugin skill equivalents are accepted instead of only allowing exactly one skill folder. I added Rust coverage for both nested user skills and nested plugin skills.

DOsinga (changes requested):
this looks like a lot of churn for just making the agent file backed. I think we should break this up in 

* a PR that actually moves this to the sources CRUD api
* unify the various client actions we do based on this (copy/save/duplicate/delete handling)
* cleans up the existing Persona & Agent duplication
* then make sure the UI works fluid (which I think was the original issue)

ui/goose2/src-tauri/src/services/personas.rs
```diff
@@ -43,58 +64,161 @@ impl PersonaStore {
...
+                continue;
+            }
+            if let Err(err) = std::fs::write(&path, Self::persona_to_markdown(&persona)) {
+                warn!("Failed to seed agent {:?}: {}", path, err);
+            }
+        }
+        let _ = std::fs::write(Self::seed_marker_path(), chrono::Utc::now().to_rfc3339());
+    }
+
+    fn migrate_legacy_markdown_agents() {
```
jamadeo, line 112: I don't think we should have any treatment of "legacy" formats that were at some point part of goose2. We should prefer breaking compatibility for ourselves up until the first release

PR #9028, opened 2026-05-05
docs: improve goose2 AGENTS.md

alexhancock:
## Summary
Fixing a couple issues in the AGENTS.md I set up two weeks ago for goose2

1. There was already an Architecture & File Structure section that was likely stomping on the further down Architecture system from being respected
2. We never said anything to the agent about not just building things entirely in the frontend TS, which we don't generally want because then it's hard for any other client to leverage them

### Testing
"what are the architectural issues in the implementation here?"

### Related Issues
N/A

N/A

### DIFF
--- a/ui/goose2/AGENTS.md
+++ b/ui/goose2/AGENTS.md
@@ -39,8 +39,11 @@ Treat this repo as partially Hermit-managed. Do not assume `just`, `pnpm`, `node
 React UI  ──►  features/<feature>/api/  ──►  @aaif/goose-sdk (TS)  ──►  goose-acp  (WebSocket, ACP)  ──►  goose (core)
 ```
 
+**YOU MUST TREAT THE CLIENT as a THIN CLIENT****
+
 - The Tauri shell spawns a long-lived `goose serve` process and exposes its WebSocket URL via the `get_goose_serve_url` Tauri command. That is essentially the only Tauri command the frontend needs for backend work — it is how the renderer discovers the ACP endpoint.
 - The frontend opens a WebSocket to `goose serve` and talks to it using `@aaif/goose-sdk` (published from `ui/sdk/`). The SDK is generated from the ACP custom-method definitions in `crates/goose-sdk/src/custom_requests.rs`, so every backend method has a typed TypeScript client method.
+- The goose2 TypeScript code should only be UI
 - `goose-acp` (`crates/goose-acp/src/server.rs`) is the server side of the WebSocket. It implements handlers for the custom ACP methods and calls into the `goose` core crate to do the actual work (providers, config, sessions, dictation, etc.).
 - `goose` is the pure domain crate. It knows nothing about Tauri or WebSockets — it just exposes Rust APIs that `goose-acp` handlers invoke.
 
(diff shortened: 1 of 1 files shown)


baxen (approved):
LGTM!

ui/goose2/AGENTS.md
```diff
@@ -29,7 +29,37 @@ Treat this repo as partially Hermit-managed. Do not assume `just`, `pnpm`, `node
...
+
+**This is the pattern you must follow when adding any new backend-touching feature.** When you are vibecoding in this app, it is very tempting to reach for `invoke()` or add an HTTP fetch — don't. The rule is: if a feature needs to talk to `goose` core, it goes through the SDK → ACP → goose chain above.
+
+### Don't build features entirely in the frontend
+
+If a feature involves data, persistence, secrets, provider config, sessions, filesystem access, network calls to external services, or anything else that could plausibly be reused by the CLI or another goose surface, **the logic belongs in the `goose` core crate, exposed via a typed ACP method.** Do not:
+
+- Stand up a feature whose business logic lives in a Zustand store, a React hook, or a `features/<feature>/api/` adapter that calls `localStorage`, `fetch`, or filesystem APIs directly.
+- Reach for `invoke()` to add a new Tauri command that proxies into `goose` — add an ACP custom method instead.
+- Persist domain state to `localStorage` when goose core already has (or should have) a config/state surface for it. `localStorage` is for **frontend-only** UI state (theme mode, sidebar collapse, last-viewed tab, etc.).
```
DOsinga: I would go further - do not use localStorage. we should really have a client side settings module that re-uses the existing settings from goose 1.0 as much as possible

ui/goose2/AGENTS.md
```diff
@@ -29,7 +29,37 @@ Treat this repo as partially Hermit-managed. Do not assume `just`, `pnpm`, `node
...
+- Duplicate types between TS and Rust by hand — let the SDK generation produce them from `crates/goose-sdk/src/custom_requests.rs`.
```
DOsinga: add to really really use those types and not create shadow types of those if that seems convenient. Agents love to do that, oh I need an isSaved field, I'll just add it

ui/goose2/AGENTS.md
```diff
@@ -38,8 +68,8 @@ src/
     <feature>/
       ui/        — React components (required)
       hooks/     — Custom React hooks for feature logic (when needed)
-      stores/    — Zustand state management (when feature needs shared state)
-      api/       — Backend API integration (when feature calls backend)
+      stores/    — Zustand state management (frontend-only UI state; not a substitute for goose core)
+      api/       — Thin wrappers around GooseClient SDK calls (the ONLY place ACP is touched)
       types.ts   — Feature-specific type definitions (when needed)
   shared/
     types/       — Canonical shared type definitions (single source of truth)
```
DOsinga: this should go. it is not the single source of truth, it is a shadow set of types

ui/goose2/AGENTS.md
```diff
@@ -51,29 +81,32 @@ src/
...
+| **Tabs**             | `ui/` + `types.ts`               | tabs                            | No backend (frontend-only UI state)            |
+
+If you're tempted to build a "Full-featured" or "Data-driven" feature without a corresponding `api/` module that calls a typed ACP method, stop and add the ACP method first. See "The canonical example" below.
 
 ### Import Rules for Features
 
 - Shared types live in `src/shared/types/` — this is the single source of truth for cross-feature types.
 - There should be NO root-level `src/stores/` or `src/types/` directories.
 - Feature stores use feature-relative imports (e.g., `../stores/featureStore`).
 - Cross-feature imports use `@/features/*/stores/` or `@/shared/types/`.
```
DOsinga, line 107: idem

ui/goose2/AGENTS.md
```diff
@@ -29,7 +29,37 @@ Treat this repo as partially Hermit-managed. Do not assume `just`, `pnpm`, `node
 - `just ci` is the main local verification gate.
 - `just clean` removes Rust build artifacts, `dist`, and `node_modules`, so `just setup` is required again before `just dev`.
 
-## Architecture & File Structure
+## Architecture
+
+### The frontend → ACP → goose core path
+
+**All frontend ↔ backend communication in goose2 flows through a single path:**
```
DOsinga, line 36: add something like YOU MUST TREAT THE CLIENT as a THIN CLIENT

DOsinga (approved):

PR #9038, opened 2026-05-05
Add Linux Vulkan support for local inference

jh-block:
- Add an opt-in `vulkan` feature for `goose`, `goose-cli`, and `goose-server` to enable llama.cpp Vulkan support on Linux.
- Enable Vulkan in Linux CI, bundle, and native packaging paths while leaving macOS Metal and Windows unchanged.
- Add Linux build/runtime dependency notes for the Vulkan loader and `glslc`, and log discovered non-CPU llama.cpp backend devices at startup for validation.

### DIFF
--- a/ui/goose2/src-tauri/tauri.conf.json
+++ b/ui/goose2/src-tauri/tauri.conf.json
@@ -50,6 +50,14 @@
       "../distro": "distro"
     },
     "externalBin": ["../../../target/release/goose"],
+    "linux": {
+      "deb": {
+        "depends": ["libvulkan1"]
+      },
+      "rpm": {
+        "depends": ["vulkan-loader"]
+      }
+    },
     "macOS": {
       "entitlements": "entitlements.plist",
       "hardenedRuntime": true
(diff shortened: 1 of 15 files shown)


ui/goose2/src-tauri/tauri.conf.json
```diff
@@ -50,6 +50,14 @@
       "../distro": "distro"
     },
     "externalBin": ["../../../target/release/goose"],
+    "linux": {
+      "deb": {
+        "depends": ["libvulkan1"]
```
alexhancock, line 55: does this also need `libvulkan-dev` and/or `glslc`?

alexhancock (approved):

ui/goose2/src-tauri/tauri.conf.json
jh-block: if I understand the [meaning of this key](https://v2.tauri.app/distribute/debian/#custom-files) correctly, I think it's runtime dependencies for the .deb when installed, so it should be fine without those build deps

PR #9052, opened 2026-05-06
refactor(goose2): remove attachment preamble

alexhancock:
## Summary
Remove the preamble for images, and make it so for files/dirs it just does the same thing goose1 desktop used to, which is to suffix an absolute path to the files and let the agent read it. This version has one improvement where it doesn't include it as plaintext and renders a nice bubble instead.

### Testing
Existing tests + manual test in goose2 desktop

### Related Issues
N/A

Screenshot 2026-05-06 at 10 15 48 AM
Screenshot 2026-05-06 at 10 40 19 AM


DOsinga (approved):
nice improvement. the tests we are touching here are showing they are not very useful seeing that we need to change them for something not that related. I think we could drop them

Issue #9005, opened 2026-05-04
Poor contrast for some inline code in GUI light theme


jramnani:
There is an edge case in the UI light theme where code snippets rendered within bolded text result in light gray text on a white background, which is very hard to read at a glance.

### Root Cause

Two issues combine to produce nearly invisible text:

1. **Direct-child CSS selector fails for nested inline code.** The accent color override uses `p > code.bg-inline-code` and `li > code.bg-inline-code` (direct-child combinator). But when inline code appears inside `<strong>`, `<em>`, or other inline formatting elements, the `<code>` is no longer a direct child of `<p>` — so it falls back to the base color.

2. **The base `.bg-inline-code` color has poor contrast on white.** The fallback color is `--color-neutral-200` (`#cbd1d6`) which was designed for dark backgrounds. On white, it produces a **1.54:1** contrast ratio — far below the WCAG 2.1 AA minimum of 4.5:1.

### Measured Contrast Ratios

| Element | Color | Background | Contrast Ratio | WCAG AA (4.5:1) |
|---------|-------|------------|---------------|-----------------|
| Base inline code (`.bg-inline-code`) | `#cbd1d6` | `#ffffff` | **1.54:1** | ❌ FAILS |
| Accent inline code (in `p > code`) | `#ff4f00` | `#ffffff` | **3.30:1** | ❌ FAILS |

WCAG 2.1 Level AA requires a minimum contrast ratio of **4.5:1** for normal-sized text.

---

**To Reproduce**
Send this markdown to Goose and view the response in light theme:

```markdown
1. **Add `DB::enableQueryLog()` to the service provider** — but then `UserController` and `OrderController` would also log queries, which adds overhead.

2. **Override `getResults()` in the cached repository** — this works cleanly without affecting other repositories.
```

The key trigger is inline code (`backticks`) **inside bold text** (`**...**`) within a list item. This produces HTML like:

```html
<li>
  <p>
    <strong>
      Add <code class="bg-inline-code">DB::enableQueryLog()</code> to the service provider
    </strong>
    — but then...
  </p>
</li>
```

Since `<code>` is inside `<strong>`, the selector `p > code.bg-inline-code` doesn't match, and the near-invisible base grey is used instead of the orange accent.

---

**Expected behavior**

The text is styled in a way that is easier for the user to read.

---

**Screenshots**

Here is an example that illustrates the poor contrast. For example, the text, `DB::enableQueryLog()`

[...]

PR #9058, opened 2026-05-06
fix(a11y): improve inline code contrast in light theme

jramnani:
## Summary

Inline code text rendered as nearly invisible light grey (#cbd1d6) on white background in the light theme. Contrast ratio was 1.54:1 — far below the WCAG 2.1 AA minimum of 4.5:1.

Root cause:
1. The base .bg-inline-code class used --color-neutral-200 (#cbd1d6), a color designed for dark backgrounds, with no light/dark theme differentiation.
2. The accent color override used parent-specific selectors (p > code, li > code) which failed in many common contexts:
   - Inline code inside `<strong>` (bold text)
   - Inline code inside `<em>` (italic text)
   - Inline code inside `<td>/<th>` (table cells)
   - Inline code inside `<a>` (links)
   - Any other nesting beyond direct child of `<p>` or `<li>`

Fix:
- Apply the accessible accent color directly on the base .bg-inline-code class so it works in ALL contexts (tables, headings, bold, italic, links, etc.)
- Light theme: #c2410c (4.78:1 contrast ✅ AA) with subtle #f4f6f7 background.
- Dark theme: #ff7235 (4.51:1 contrast ✅ AA) with #32353b background.
- Removed the parent-specific selectors entirely — no longer needed since the base class now carries the correct color everywhere.

### Testing

I've been manually testing this on macOS arm64.

### Related Issues
Relates to #9005 
Discussion: N/A

Before:  

Image

After:   

Image

### DIFF
--- a/ui/desktop/src/styles/main.css
+++ b/ui/desktop/src/styles/main.css
@@ -183,6 +183,8 @@
   --text-inverse: var(--color-white);
   --text-danger: var(--color-red-200);
 
+  --color-inline-code: #c2410c;
+
   /* Search highlighting */
   --highlight-color: rgba(255, 213, 0, 0.5);
   --highlight-current: rgba(252, 213, 3, 0.6);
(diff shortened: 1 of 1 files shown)


DOsinga (approved):
Clean accessibility fix — thanks!

PR #9063, opened 2026-05-06
plugins: add open plugins (just skills for now)

jamadeo:


DOsinga (approved):
Looks good. Merged main, fixed the manifestless naming bug, removed the dbg! macro, hardened copy_dir_all against symlink traversal, and integrated the hooks format detection from main.

PR #9064, opened 2026-05-06
fix: TUI proper binary resolution

alexhancock:
## Summary

Fixes two problems

1. When someone doesn't have goose at all, the TUI doesn't find the appropriate binary package in its `node_module` tree and it is unable to run `goose acp`. This is fixed by adding the `ui/goose-binary` packages as optional deps at the level of `ui/text` as well as `ui/sdk`.
2. Removes `bin` definitions in the `ui/goose-binary` npm `package.json` files which can activate if the CWD is in a location where an executable at that relative path exists. In this scenario you currently see the goose-cli experience when running `npx @aaif/goose`

### Testing
Will test published package

### Related Issues
N/A

N/A

### DIFF
--- a/ui/goose-binary/goose-binary-darwin-arm64/package.json
+++ b/ui/goose-binary/goose-binary-darwin-arm64/package.json
@@ -13,6 +13,9 @@
     "agent",
     "binary"
   ],
+  "scripts": {
+    "postinstall": "node -e \"require('fs').chmodSync(require('path').join(__dirname,'bin','goose'), 0o755)\""
+  },
   "os": [
     "darwin"
   ],
--- a/ui/goose-binary/goose-binary-darwin-arm64/package.json
+++ b/ui/goose-binary/goose-binary-darwin-arm64/package.json
@@ -7,9 +7,6 @@
     "type": "git",
     "url": "git+https://github.com/aaif-goose/goose.git"
   },
-  "bin": {
-    "goose": "bin/goose"
-  },
   "keywords": [
     "goose",
     "ai",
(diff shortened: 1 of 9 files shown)


ui/goose-binary/goose-binary-darwin-arm64/package.json
```diff
@@ -7,9 +7,6 @@
     "type": "git",
     "url": "git+https://github.com/aaif-goose/goose.git"
   },
-  "bin": {
-    "goose": "bin/goose"
-  },
```
jamadeo, line 12: these were here because without them, the npm packaging would remove the executable bit from the file

alexhancock: Interesting. I think it's causing the `npx @aaif/goose` experience to launch the old cli though - depending on CWD somehow. Example:

```
🐚 8:59:27 ~/Development/goose alexhancock/goose-tui-binary-resolution $ npx @aaif/goose

    __( O)>  ● new session · databricks databricks-claude-opus-4-7
   \____)    20260507_1 · /Users/alexhancock/Development/goose
     L L     goose is ready
  ╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌ 0% 0/128k
🪿 Enter to send · Ctrl+J newline
```

I may be wrong about root cause, but this is what goose thought it was. Trying to figure it out and fix it.

jamadeo: huh! that is unexpected...

alexhancock:
Closing in favor of something simpler

PR #9069, opened 2026-05-07
feat: add /goal command for agent self-evaluation before finishing

michaelneale:
I found using goose just ... bare, I wanted to have it loop a little more when not around - giving it a goal to check on, a target. 

-------

Add a /goal command that lets users set a goal the agent must verify before exiting its reply loop. When the agent would normally finish (text response, no tool calls), it instead injects a nudge message asking itself to verify the goal is met. If not met, it continues working. Capped at 3 goal-check nudges to prevent loops.

Usage:
  /goal ensure all tests pass and code compiles
  /goal             (show current goal)
  /goal off         (clear the goal)


jamadeo (approved):
I like the simplicity of this. I agree with @DOsinga 's ideas mentioned in discord but also it's hard to imagine keeping it this simple while fitting it in with the other bits.

One other thought here is that this has a lot in common with the "final output" tool, and maybe we could do something to tidy that up.

michaelneale:
@jamadeo yeah interesting. FWIW been using it for a few days and the simplicity does seem a good balance of just working.

DOsinga:
Hey @michaelneale — this PR has merge conflicts with main. Could you rebase/merge to resolve them? Happy to help get this landed once it's clean.

PR #9070, opened 2026-05-07
More improvements to local inference providers

monroewilliams:
## Summary
Add nonempty api_key_env configuration to lmstudio and llama_swap declarative providers.
Add a new declarative provider for [oMLX](https://omlx.ai), similar to the above.

These providers all set requires_auth=false since auth is optional for local inference, but due to logic elsewhere in the code, openai-engine based providers' api keys could not be configured through the UI and would not be sent to the endpoint unless requires_auth was set to true.

To fix this, also correct the logic in several places so that declarative providers based on the openai engine which define a nonempty api_key_env and have requires_auth=false allow the key to be set through the UI and will send the key to the endpoint.

Changes:
- omlx.json: New provider with api_key_env set to OMLX_API_KEY
- lmstudio.json: Set api_key_env to LMSTUDIO_API_KEY
- llama_swap.json: Set api_key_env to LLAMA_SWAP_API_KEY
- declarative_providers.rs: Update llama_swap api_key_env test assertion
- provider_registry.rs: Expose api_key_env config in UI regardless of requires_auth setting
- openai.rs: Read API key when configured, regardless of requires_auth. For the requires_auth=false case:
    - handle NotFound as not an error (the user didn't provide the optional api key)
    - warn on keyring errors instead of failing (the api key isn't needed, so this isn't fatal)
- inventory/mod.rs: Include api_key_env in identity hash regardless of the value of requires_auth so model cache invalidates when the optional key changes.

### Testing
I've tested this manually against the oMLX provider with it set to require an api key for authentication.

### Related Issues
N/A

Before:  
587483722-185f4ed4-34e8-4ff7-84fd-520270c22b10

After:   
587478770-0189abba-d143-4194-b64c-29055fe888de

### DIFF
--- a/.plan/per_model_context_limit_ui.md
+++ b/.plan/per_model_context_limit_ui.md
@@ -1,0 +1,159 @@
+# Per-Model Context Limit UI in Context Popover
+
+## Goal
+
+Let users override the context limit for dynamic models directly from the context usage popover (near the "Compact now" button). The override applies to the current session and persists across future sessions for that provider+model combination.
+
+## Background
+
+- Declarative providers (lmstudio, llama_swap, omlx) now have a `context_limit` field in their JSON config as a provider-level default (e.g., 1024K).
+- Users may need per-model overrides that differ from the provider default.
+- The local inference provider already has a `ModelSettingsPanel` with a `context_size` field — the declarative provider UI should offer a lighter-weight equivalent in the popover.
+
+---
+
+## Backend (Rust)
+
+### 1. ACP custom methods
+
+**File:** `crates/goose-sdk/src/custom_requests.rs`
+
+Add three new request/response structs:
+
+```
+_get goose/model/context_limit/set    — Store a per-model context limit override
+_get goose/model/context_limit/get    — Read the stored override for a provider+model
+_get goose/model/context_limit/reset  — Remove the override (revert to provider/default)
+```
+
+Request struct for set/get:
+```rust
+pub struct ModelContextLimitRequest {
+    pub provider: String,
+    pub model: String,
+}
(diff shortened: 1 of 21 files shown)


monroewilliams:
@Abhijay007 , here's a follow-up to my earlier LM Studio provider PR. 

It's a bit more involved than the last one, since it had to fix some logic assumptions in the code (tying api_key_env to requires_auth=true) that didn't hold for optional api keys.

Thanks!

DOsinga:
Thanks for the contribution! We're closing this because it bundles too many unrelated changes for a single PR — especially at this stage. Per our [contributing guide](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#getting-started), please split this up into focused PRs.

For example, we'd be happy to take the new oMLX declarative provider as a standalone PR. The optional API key fix could be another. The per-model context limit override system is a larger feature that would benefit from an [issue or discussion](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#discussions-issues-and-prs) first.

A couple of other things:

- The `.plan/` file doesn't belong in a PR — that's local planning material.
- There are [unaddressed codex review comments](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#ai-code-reviews) (including two P1s) that need to be resolved before we can review.

Please don't take this personally — the work looks thoughtful, it just needs to be broken up. Looking forward to smaller follow-ups!

monroewilliams:
All fair requests. I've been working on various usability fixes on a personal branch, and should have closed the PR to keep the churn to myself. I'll split things into smaller pieces for actual PRs.

PR #9078, opened 2026-05-07
refactor: remove threads layer, use sessions directly for ACP

DOsinga:
The threads abstraction (threads + thread_messages tables) duplicated what sessions + messages already provide. The messages table has user_visible/agent_visible flags that handle the separation between what the user sees and what the agent works with.

Backend changes:
- Deletes thread_manager.rs entirely (-947 lines)
- Rewrites all ACP handlers to use session_manager directly
- on_new_session: creates a Session (type Acp) directly
- on_load_session: loads session with messages, replays user_visible ones
- on_list_sessions: lists User + Scheduled sessions from sessions table
- on_prompt: agent persists messages internally, no duplicate writes
- on_fork_session: uses session_manager.copy_session
- Session handlers (delete/export/import/rename/archive): all direct
- Adds archived_at and project_id columns to sessions (migration 12)
- Drops threads and thread_messages tables in migration 12

Frontend changes:
- Removes personaId from backend round-trip (pure UI concern)
- Renames session-level personaId to agentId on ChatSession
- agentId is stored in the Zustand chatSessionStore only
- Removes personaId from ACP API calls (newSession, prompt, sendMessage)

This makes all sessions visible in goose2 regardless of how they were created (CLI, old Electron app, scheduled runs).

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:


crates/goose/src/acp/server.rs
```diff
@@ -3327,24 +3134,15 @@ impl GooseAcpAgent {
...
-                let cwd = t
-                    .working_dir
-                    .as_deref()
-                    .map(std::path::PathBuf::from)
-                    .unwrap_or_default();
-                let meta = thread_session_meta(&t);
-                SessionInfo::new(SessionId::new(t.id), cwd)
-                    .title(t.name)
-                    .updated_at(t.updated_at.to_rfc3339())
+            .filter(|s| s.message_count > 0 || s.archived_at.is_some())
```
jh-block: Is it deliberate that archived sessions are included? (The old behaviour excluded archived ones.)

crates/goose/src/acp/server.rs
```diff
@@ -1797,30 +1776,9 @@ impl GooseAcpAgent {
...
-                            );
-                        }
-                    } else {
-                        warn!(
-                            "tool call summary: missing message_id for {request_id} — title will not survive reload",
-                        );
-                    }
-                }
+                // Tool call title generated; displayed in UI via notification.
+                // Persistence to messages table for reload is not yet implemented.
```
jh-block: This will cause the tool titles to disappear on reload I think; if punting it is deliberate we might as well remove the earlier LLM call as well?

crates/goose/src/session/session_manager.rs
```diff
@@ -1075,6 +1056,40 @@ impl SessionStorage {
             11 => {
                 crate::providers::inventory::create_tables_in_tx(tx).await?;
             }
+            12 => {
```
jh-block, line 1063: This will cause a few problems for people who already ran with threads, in particular the old ACP session ID was the thread ID so sessions might be unreachable since they have no corresponding row in the sessions table. do we care?

crates/goose/src/session/session_manager.rs
```diff
@@ -384,21 +407,16 @@ impl SessionManager {
...
-                    let thread = thread_mgr
-                        .update_thread(thread_id, Some(name), Some(false), None)
-                        .await?;
-                    return Ok(Some(SessionNameUpdate { thread }));
-                }
-            }
-            Ok(None)
-        } else {
-            Ok(None)
+            let session = self.get_session(id, false).await?;
```
jh-block, line 401: I think this will emit even if the name didn't change, does this code path ever get called with the same name? Will emit spurious notifications if so.

crates/goose/src/acp/server.rs
```diff
@@ -3356,28 +3154,34 @@ impl GooseAcpAgent {
...
-        let source_thread_id = &*args.session_id.0;
+        let source_session_id = &*args.session_id.0;
 
-        // Fork the thread (copies metadata + messages).
-        let new_thread = self
-            .thread_manager
-            .fork_thread(source_thread_id)
+        let new_session = self
+            .session_manager
+            .copy_session(source_session_id, "Fork".to_string())
```
jh-block, line 3196: copy_session doesn't copy the project_id which used to be carried via thread metadata.

crates/goose/src/acp/server.rs
```diff
@@ -152,7 +152,7 @@ async fn ensure_refresh_identity_current(
 /// - **Thread** (`threads` table) = the ACP session. The `sessionId` that ACP clients
```
jh-block: I think this is stale

crates/goose/src/acp/server.rs
```diff
@@ -248,51 +244,40 @@ fn sid_short(id: &str) -> String {
...
-        );
-    }
-    if let Some(ref model_id) = thread.metadata.model_id {
-        meta.insert(
-            "modelId".to_string(),
-            serde_json::Value::String(model_id.clone()),
-        );
+    if let Some(ref provider) = session.provider_name {
+        meta.entry("providerId".to_string())
+            .or_insert_with(|| serde_json::Value::String(provider.clone()));
```
jh-block: This (and modelId below) can just be insert() calls

jh-block (approved):

crates/goose/src/acp/server.rs
DOsinga: thanks

crates/goose/src/acp/server.rs
DOsinga: yes

crates/goose/src/session/session_manager.rs
DOsinga: I think it is fine yeah

crates/goose/src/acp/server.rs
DOsinga: good catch!

crates/goose/src/session/session_manager.rs
```diff
@@ -379,26 +402,24 @@ impl SessionManager {
...
+            return Ok(Some(SessionNameUpdate {
+                session_id: id.to_string(),
+                name,
+                updated_at: session.updated_at,
+                message_count: session.message_count,
+                user_set_name: session.user_set_name,
```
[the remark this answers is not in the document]
DOsinga: yes it does

crates/goose/src/session/session_manager.rs
```diff
@@ -1075,6 +1049,40 @@ impl SessionStorage {
...
+                    sqlx::query("ALTER TABLE sessions ADD COLUMN project_id TEXT")
+                        .execute(&mut **tx)
+                        .await?;
+                }
+
+                // Drop thread tables (no longer used).
+                sqlx::query("DROP TABLE IF EXISTS thread_messages")
+                    .execute(&mut **tx)
+                    .await?;
+                sqlx::query("DROP TABLE IF EXISTS threads")
```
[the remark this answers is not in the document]
DOsinga: that's alright

crates/goose/src/session/session_manager.rs
DOsinga: I suppose it could happen, but seems quite unlikely, probably not worth protecting against one spurious notification every so often. good find though

PR #9079, opened 2026-05-07
fix(ui): preserve user-set session name for recipe-based chats

### LINKED ISSUE #8865: Chats started from recipes cannot be renamed in left side bar
The left sidebar shows the chats as rows. When a chat is started from a recipe, it shows a chef's hat and the recipe name instead of the chat name. This cannot be changed and makes it very difficult to know what is what.

---

**To Reproduce**
1. Use a recipe
2. Use the same recipe again and again. Now I have three chats with the recipe name.
3. Try to rename one of the recipe chats in the left sidebar via double click and by adding the word "BROKEN" and hitting enter. Nothing happens
4. Go into "Chat" -> "Show All". The chat title seems to have been renamed, but it is not reflected in the left sidebar.

---

**Expected behavior**
The left sidebar should show the chat title instead of the recipe name.

---

**Screenshots**

Image

Image

---

- **OS & Arch:** Ubuntu 24.04 x86-64
- **Interface:** UI
- **Version:** 1.32.0
- **Extensions enabled:** 
- **Provider & Model:** custom

williams145:
## Summary

`getSessionDisplayName` in `useNavigationSessions.ts` returned `recipe.title` unconditionally for any session associated with a recipe, so renaming a recipe-based chat in the left sidebar appeared to do nothing — the new name was persisted on the backend (and visible in **Chat → Show All**) but the sidebar kept showing the recipe title.

This PR fixes the priority order to honour the existing `user_set_name` flag, which the backend already toggles in `update_session_name` (see `crates/goose/src/session/session_manager.rs`). The frontend just needed to read it.

```diff
 export function getSessionDisplayName(session: Session): string {
+  if (session.user_set_name) {
+    return session.name;
+  }
   if (session.recipe?.title) {
     return session.recipe.title;
   }
   if (shouldShowNewChatTitle(session)) {
     return 'New Chat';
   }
   return session.name;
 }
```

Fixes #8865

## Test plan

- [x] Added unit tests in `ui/desktop/src/__tests__/sessions.test.ts` under `describe('getSessionDisplayName (fix for #8865)')`:
  - recipe + user-renamed → returns user name (the bug)
  - recipe + not renamed → returns recipe title (regression guard)
  - empty non-recipe → returns `'New Chat'`
  - non-recipe + user-renamed → returns user name
- [ ] Manual verification in the desktop UI:
  1. Start a chat from a recipe.
  2. Double-click the entry in the left sidebar and rename it (e.g. add "RENAMED").
  3. Confirm the sidebar entry now shows the renamed title and updates immediately.
  4. Confirm **Chat → Show All** still reflects the same name (no regression).
  5. Start a fresh recipe chat with no rename — confirm the sidebar still shows the recipe title and the chef-hat icon, as before.


DOsinga (approved):
The fix is correct and minimal —  takes priority over `recipe.title` in `getSessionDisplayName`, which is the right behavior.

I pushed two commits:
1. A merge of main to bring the branch up to date (this also reverted an unrelated deletion of the post-delete session re-fetch logic in `handleSessionDeleted` that was in the original commit)
2. Removal of two redundant tests that were asserting pre-existing behavior unrelated to the #8865 fix

**DCO**: Your original commit is missing the `Signed-off-by` line required by the DCO check. Please amend it:
```
git commit --amend -s
git push --force-with-lease
```

PR #9084, opened 2026-05-07
Agents crud

jamadeo:
This PR moves the goose2 agents (f.k.a. "personas") to use the ACP+ "sources" construct. This adds handling for the Agent variant of a Source: discovery on the filesystem and CRUD operations.

This also supports the existing goose2 built-in personas by allowing a caller to set additional discovery paths using the `--agent-root` CLI argument. It moves the hard-coded personas "ralph", "scout", and "solo" into markdown files that are packaged into the tauri app and discovered using the usual filesystem discovery.

The UI bits of this are very much vibe coded, as are some of the sources changes, though I did iterate manually on those a bit.


crates/goose-cli/src/cli.rs
```diff
@@ -791,6 +792,14 @@ enum Command {
...
         )]
         builtins: Vec<String>,
+
+        #[arg(
+            long = "agent-root",
+            value_name = "PATH",
+            help = "Add a read-only directory containing bundled agent markdown files",
+            action = clap::ArgAction::Append
+        )]
+        agent_roots: Vec<std::path::PathBuf>,
```
DOsinga: this seems unexpected

crates/goose/src/sources.rs
```diff
@@ -488,6 +875,15 @@ pub fn list_sources(
     source_type: Option<SourceType>,
     project_dir: Option<&str>,
     include_project_sources: bool,
+) -> Result<Vec<SourceEntry>, Error> {
+    list_sources_with_roots(source_type, project_dir, include_project_sources, &[])
+}
+
+pub fn list_sources_with_roots(
```
DOsinga, line 857: we now have the api every where with and without sources. the without sources is only called from tests(?) -- suggests to just simplify to list_sources and just have the tests where needed pass in an empty list

crates/goose/src/sources.rs
```diff
@@ -601,6 +1008,27 @@ pub fn export_source(source_type: SourceType, path: &str) -> Result<(String, Str
...
+                is_global_agent_file(&file_path) || !writable,
+                writable,
+            )?;
+            let export = serde_json::json!({
+                "version": 1,
+                "type": "agent",
+                "name": source.name,
+                "description": source.description,
+                "content": source.content,
+            });
```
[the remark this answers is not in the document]
jamadeo: this is going to be rewritten

crates/goose/src/sources.rs
```diff
@@ -290,6 +313,322 @@ fn builtin_skill_entry(mut source: SourceEntry) -> SourceEntry {
...
+        trimmed
+            .chars()
+            .take(64)
+            .collect::<String>()
+            .trim_end_matches('-')
+            .to_string()
+    }
+}
+
+fn parse_agent_frontmatter(raw: &str) -> Result<(MarkdownSourceFrontmatter, String), Error> {
```
DOsinga, line 380: here and elsewhere, this is getting all a bit more duplicated than needed maybe. can do in a separate PR of course, but something where we have a parse the type of source around could reduce dupliation

crates/goose/src/sources.rs
```diff
@@ -290,6 +313,322 @@ fn builtin_skill_entry(mut source: SourceEntry) -> SourceEntry {
...
+        });
+    }
+
+    dirs.push(SourceRoot {
+        path: Paths::agents_dir(),
+        writable: true,
+    });
+    if let Some(home) = dirs::home_dir() {
+        dirs.push(SourceRoot {
+            path: home.join(".agents").join("agents"),
```
DOsinga, line 506: this pattern happens 3 times - should extract into something data

crates/goose/src/sources.rs
```diff
@@ -71,37 +74,43 @@ fn project_file_path(slug: &str) -> PathBuf {
...
+    let mut md = format!("---\n{yaml}---\n");
+    if !content.is_empty() {
+        md.push('\n');
+        md.push_str(content);
+        md.push('\n');
+    }
+    Ok(md)
+}
+
+fn build_source_markdown(
```
DOsinga, line 77: this is better, but we still have the build_source_md -- maybe delete that one

DOsinga (approved):

crates/goose/src/sources.rs
jamadeo: Good point!

crates/goose/src/sources.rs
jamadeo: yeah that was used only here; done

PR #9087, opened 2026-05-07
fix: insert tool pair summaries at chronological position in conversation

DOsinga:
The in memory version of tool pair summaries would get out whack. this changes the behavior to only doing tool pair summarization once and leaving the in memory version of it unchanged


michaelneale:
I think I am hitting this.

michaelneale (approved):
i think I may have been bitten hard by this... hrm

PR #9092, opened 2026-05-07
poc: add Terminal-Bench 2.0 runner

Raptors65:
## Summary

This is a POC for adding a way to run Terminal-Bench 2.0 with Goose. There's some slop here but working on cleaning it up :)

With this POC, the goal is to do the following (open to discussing whether this is the *right* goal to aim for):
make it easy for anyone (can be an end user, developer, or some CI workflow) to run Terminal-Bench 2.0 (and in the future, other benchmarks) with Goose; it should basically be as easy as running `goose bench run terminal-bench`

Terminal-Bench uses Harbor to run tasks ([docs](https://www.harborframework.com/docs/tutorials/running-terminal-bench)). Harbor lets you run Terminal-Bench within a Docker container locally, and provides an interface for plugging custom agents into the benchmark. goose actually has a built-in Harbor adapter; the issue is that the built-in implementation pulls goose from GitHub's releases, which is not what we want here (we want to run the goose that the user has installed locally).

So the overall approach here is that we have a custom Python adapter (in `goose_external.py`) that runs the already installed version of goose instead of downloading it, and we have a custom MCP server that runs *outside* the Harbor container with the `developer` tools (shell, write, edit, tree) and forwards tool calls into the container using `docker exec` (the reasoning for why I didn't just use `--container` is below).

For the reasoning behind the overall approach, some notes:
- my initial idea for this involved installing goose *into* the task container. The issue with this is that if the user is not on a Linux machine, this requires a build step. I had something for this working locally, but it took ~15-20 min to build, and of course is not possible to do for an end user who doesn't have the actual source code
- there's a `--container` flag to `goose run` that allows you to run extensions inside a container. the issue is that as currently built, this actually still requires goose to be installed inside the container, because it starts extensions as `goose mcp <extension-name>`. This is the reasoning for the bench-specific MCP server in `bench.rs`. As an alternative, we could have proper support for having tools run inside a container without actually requiring goose to be installed inside the container, but I haven't looked into that in this POC. again, the issue with needing to install goose into the container is that it requires a build step if you're not on Linux, and you can't build if you don't have the source code

also just a note: this is currently tied to the CLI, but we might want to make it work with the desktop app / other frontends in the future; I think the overall approach here would still work, we'd just need to move some of the logic from the CLI into the goose core or into some other shared crate

### Testing

something like the following should work, replacing `goose` with the path to the goose binary you want to use (or using `cargo run --package goose-cli --bin goose --`)

```bash
goose bench run terminal-bench \
  --model 'PUT_MODEL_HERE' \
  --task fix-git \
  --trials 1 \
  --concurrency 1 \
  --max-turns 20
```

ran this successfully locally with Qwen3.6-27B on some tasks

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:


Raptors65:
* **#9092** <a href="https://app.graphite.com/github/pr/aaif-goose/goose/9092?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/aaif-goose/goose/9092?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

PR #9093, opened 2026-05-07
feat: Hooks

alexhancock:
## Summary
Hooks support https://open-plugins.com/agent-builders/components/hooks

### Testing
Manually tested with desktop via `just run-ui`

### Related Issues
N/A

Screenshot 2026-05-08 at 10 39 11 AM

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -171,6 +171,7 @@ pub struct Agent {
 
     pub(super) retry_manager: RetryManager,
     pub(super) tool_inspection_manager: ToolInspectionManager,
+    pub(super) hook_manager: crate::hooks::HookManager,
     container: Mutex<Option<Container>>,
 }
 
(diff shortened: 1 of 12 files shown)


michaelneale:
that seems nicer/crisper/cleaner than I thought! pretty cool

crates/goose/src/agents/agent.rs
```diff
@@ -613,22 +667,52 @@ impl Agent {
...
+                    .with_tool(
+                        tool_call.name.to_string(),
+                        tool_call
+                            .arguments
+                            .as_ref()
+                            .map(|a| serde_json::Value::Object(a.clone())),
+                    )
+                    .with_working_dir(session.working_dir.to_string_lossy().to_string());
+            self.hook_manager
+                .emit(crate::hooks::HookEvent::PreToolUse, ctx)
```
jamadeo, line 685: PreToolUse would allow blocking tool execution, right? Is that something we'd come back to?

Looks like open-plugins itself doesn't specify, but I believe in Claude Code this is designed such that you could  e.g. have a hook deny a tool call

jamadeo (approved):

crates/goose/src/agents/agent.rs
alexhancock: Interesting! Going to merge to take advantage of the green ci but will make a note to revisit in a future change

PR #9114, opened 2026-05-08
feat(cli): add `goose review` local code review command

joahg:
## Summary

Adds `goose review`, a local code review CLI inspired by Amp's [Code Review > Checks](https://ampcode.com/manual#code-review). Compatible with the same `.agents/checks/*.md` frontmatter shape so existing checks port over without modification.

## What it does

`goose review [<RANGE>]` collects the diff for the working tree (or an explicit range like `main...HEAD`), discovers project + global review configuration, fans out a per-file main correctness pass plus N parallel check subagents, and emits findings as JSON-per-line carrying a `check` field for end-to-end attribution.

```
goose review
goose review main...HEAD
goose review --prompt my-prompt.md --model claude-sonnet-4
goose review --override-model gpt-5 --turn-limit 40
goose review --dry-run        # show the prompt that would be sent
```

## Discovery

Subagent reviewers live in Markdown files with YAML frontmatter under any `.agents/checks/` directory:

```markdown
---
name: perf            # required; must match filename for repo-local checks
description: Flag perf regressions          # optional
model: claude-sonnet-4                      # optional, see precedence below
turn-limit: 40                              # optional, default 25
tools: [Read, Grep]                         # optional, default = inherit full toolset
severity-default: high                      # optional
---
Look for N+1 queries.
```

All frontmatter fields except `name` are optional. The `tools` and `severity-default` fields are optional and backwards compatible — checks without them continue to work unchanged.

Search order, with closer scopes shadowing same-named checks:

1. `~/.config/goose/checks/`, `~/.config/agents/checks/` (global)
2. `<repo>/.agents/checks/`
3. `<repo>/<scope>/.agents/checks/` for every directory above a touched file

Global directories are loaded leniently: parse errors are warned and skipped, `README.md` is always skipped, and `name:` is allowed to differ from the filename for cross-tool compatibility.

## REVIEW.md → virtual `repo-rules` checks

`**/.agents/REVIEW.md` files are auto-derived into virtual checks named `repo-rules` (root) or `repo-rules:<scope>` (sub-tree), so their findings carry their own provenance:

```jsonl
{"severity": "critical", "path": "api/handlers/users.py", "line_start": 12, "line_end": 12, "summary": "...", "check": "repo-rules"}
```

## CLI flags

| Flag | Description |
|------|-------------|
| `--prompt <FILE>` | Replace the embedded default base prompt |
| `--model <MODEL>` | Default model for the main agent and any check without its own `model:` |
| `--override-model <MODEL>` | Force every check to use this model, ignoring per-check declarations |
| `--turn-limit <N>` | Default per-check turn limit (default `25`) |
| `--provider <PROVIDER>` | Provider for the main review agent |
| `--dry-run` | Print the assembled prompt and discovered checks instead of running |
| `-q`, `--quiet` | Suppress non-result output |
| `--no-orchestrate` | Disable the Rust-driven parallel orchestrator and fall back to the legacy single-prompt path that asks the main agent to delegate checks |

### Amp-compatible flags

The CLI also accepts the same flag surface as `amp review`, so existing review wrappers can switch tools by swapping the binary name:

| Flag | Description |
|------|-------------|
| `-i`, `--instructions <TEXT>` | Free-form reviewer instructions (PR intent, "this is a refactor — flag any behavior change", etc.). Threaded through to the main agent **and** every check subprocess so checks see the same context. |
| `-f`, `--files <FILE>...` | Restrict the diff (and therefore the review) to specific files. |
| `-c`, `--check-filter <NAME>...` | Run only the named checks; skip the rest. |
| `-s`, `--check-scope <DIR>` | Discover `.agents/checks/*.md` from this directory instead of the repo root. |
| `--checks-only` | Skip the main correctness pass; only run check subagents. |
| `--summary-only` | Print `git diff --stat` and exit; do not call the agent. |
| `--severity <LEVEL>` | Severity floor for displayed findings. Default `medium` matches Amp's CLI behavior of hiding `low` from review output. Suppressed findings are counted on stderr. |

## Precedence

- **Model:** `--override-model` > per-check frontmatter `model:` > `--model` > agent default
- **Turn limit:** per-check frontmatter `turn-limit:` > `--turn-limit` > `25`
- **Checks (same name):** closest scope (deepest sub-tree) > project root > globals

## Default prompt

The embedded default prompt does two passes before (or alongside) delegating to checks:

- **Correctness pass.** Walk every changed function looking for silent error paths (missing-row → silent zero), off-by-one and boundary errors, unhandled error returns, concurrency hazards, resource-lifecycle bugs, input-validation gaps, leaked state across requests, and logic that contradicts comments / docstrings / function names.
- **Code-quality pass.** Call out bugs and hackiness (suspicious workarounds, copy-pasted blocks that drifted), unnecessary code (dead branches, redundant null checks), too much shared mutable state, and abstraction fit *in both directions* — flag both unnecessary indirection (factories/wrappers/adapters with one caller) **and** missing abstractions (the same five-line block repeated, or hard-coded values that belong behind a name). Each finding must cite a concrete location and recommend exactly one action only when it improves the current code.

Findings from this pass are tagged with `"check": "main"`.

### Per-check prompt

Each check subprocess receives a strict, tool-free prompt that (a) requires `{"findings": [...]}` JSON output, (b) embeds the check body and any `--instructions` text, and (c) explicitly restricts findings to lines beginning with `+` in the diff and forbids reporting on unchanged/pre-existing context. Without that restriction the model routinely flags surrounding context lines that just happen to appear in the hunk.

Override the entire base prompt with `--prompt my-prompt.md`.

## Tests

```bash
cargo test -p goose-cli --lib commands::review
# 46 passed
cargo test -p goose --lib session::session_manager
# 11 passed (including test_concurrent_session_creation)
```

Coverage includes frontmatter parsing (full, minimal, invalid, extended with `severity-default` + `tools`), filename/name validation (parse keeps declared name; strict validation only enforced for repo-local checks), default-stem behavior, model + turn-limit + tools precedence, scope walking from touched files, closer-scope shadowing, virtual `repo-rules` checks synthesized from REVIEW.md at every scope, the assembled prompt layout (including the `tools` and `severity_default` columns and the attribution instruction), the orchestrator (strict per-check prompt, JSON parsing with code-fence stripping and chatter tolerance, model-resolution precedence including the CLI `--provider`+`--model` override), the per-file diff splitter (multi-file, single-file, empty input, renames using post-image path), the strict main-pass prompt (file-pinned, JSON-only, optional reviewer instructions block), reviewer-instructions threading into per-check prompts, the `--check-filter` / `--instructions` helpers, the per-check guardrails that restrict findings to `+` lines, and `Severity` ordering / lenient parsing / strict CLI parsing.

## Out of scope (follow-ups)

- Wiring `goose review` into the goose 2.0 UI (the CLI is the priority for this PR).
- Posting findings as GitHub PR review comments.
- Reusing one in-process LLM client across all checks (the current
  orchestrator spawns a `goose run` subprocess per check + per touched file;
  subprocess startup is ~50ms × N and can be eliminated by replacing the
  subprocess with a direct provider call).

## Performance + recall

Both running on **Gemini 3.1 Pro** (the same model Amp's review subagent uses on `ampcode.com/models`).

### Real PR (squareup/agents#2978, +1500-ish lines across 8 files)

| Tool | Wall clock | Findings |
|------|------------|---------------|
| `amp review` | 528s | 5 substantive correctness findings |
| `goose review` (orchestrated) | **~90s** | 2 substantive correctness findings + others as the diff evolved |

Direct `goose review` probe (bypassing the wrapper) on the same head emitted 4 findings in 26s, two of which matched Amp's findings exactly (`humanInt(-N)` formatting bug, strict branch-equality false negative) plus a `SessionID`-overwrite bug Amp missed. The remaining gap is a model-quality / non-determinism gap that lands well within the same neighborhood as Amp on Gemini 3.1 Pro.

### Synthetic fixtures (small files, planted bugs)

| Fixture | Tool | Median wall clock | Findings |
|---------|------|-------------------|----------|
| Single buggy file (5 planted bugs) | `amp review` | 45s | 5 |
| | `goose review` (orchestrated) | **17s** | 20 (all 5 checks attributed) |
| Multi-file (9 planted bug classes) | `amp review` | 50s | 30 |
| | `goose review` (orchestrated) | **20s** | 22-30, all 6 checks attributed |

### What changed under the hood

1. **Rust-driven parallel orchestrator for both checks and the main pass** (the big win).

   The original review path relied on the LLM to (a) walk the entire diff in one in-process `session.headless()` call and (b) decide whether to dispatch each check as a real subagent. Both are non-deterministic and were the dominant source of variance — and on long real-world diffs (1000+ lines) the in-process main pass reliably short-circuited with `[]` after ~30s instead of doing the work.

   The orchestrator now fans out from Rust:

   - **Per-check parallelism.** One `goose run -i - --no-session --no-profile --quiet` subprocess per discovered check, capped at `MAX_WORKERS = 4` via a Tokio semaphore, with a 5-minute per-check timeout. Each check receives a strict, tool-free prompt requiring `{"findings": [...]}` JSON output, and findings are attributed to the originating `check` in Rust so attribution no longer depends on the model.
   - **Per-file main pass.** The main correctness pass is split per touched file and dispatched as N parallel subprocesses with the same JSON contract. File-by-file context keeps the model from bailing on long diffs, and findings are tagged with `"check": "main"` so the orchestrator's emitter can attribute them like any other check.
   - **Failure isolation.** A per-check or per-file failure (subprocess error, timeout, malformed JSON) emits a warning on stderr and contributes zero findings; one broken file or check never fails the whole review.
   - **Wall clock.** Both fan-outs run concurrently via `tokio::join!`, so total wall clock is bounded by `max(slowest_main_file, slowest_check)` instead of scaling with diff size or check count.

   Add `--no-orchestrate` to fall back to the legacy single-prompt path that asks the main agent to dispatch checks via `delegate(... async: true ...)`.

2. **Idempotent SQLite schema bootstrap.** `SessionManager::create_schema` previously had a TOCTOU race (`SELECT EXISTS('schema_version')` → `CREATE TABLE schema_version`) that surfaced as `Could not create session: ... table schema_version already exists` whenever 4+ subprocesses started simultaneously on a fresh container — exactly the workload the orchestrator generates. The path now runs under `BEGIN IMMEDIATE` with `IF NOT EXISTS` on every DDL statement and `INSERT OR IGNORE` on the bootstrap version row, so concurrent first-run startup is safe.

3. **Trimmed the agent's toolset.** Skip the user's configured extension profile and load only `developer` + `summon`. On a typical setup that drops the tool list from ~35 functions (github, blockcell, MCPs, etc.) to 6, cutting the per-request prompt by ~50%.

4. **Stronger correctness prompt** that walks the diff for the bug classes subagents miss (silent error paths, off-by-one, dropped error returns, concurrency hazards, resource lifecycle, input validation, leaked state, comment/code contradictions).

---

## Update: Codex review feedback addressed (commit 47989e3)

Codex flagged 16 issues across 8 reviews; this revision fixes all 16.

**P1 correctness/security:**
- `discover.rs`: repo-root checks now properly override same-named global checks (both used `scope_priority("") = 1`; globals demoted to priority 0).
- `handler.rs`: synthesize `new file` diff chunks for untracked files in default-range mode so brand-new files actually get reviewed (`git diff HEAD` silently dropped them).
- `default_review_prompt.md`: stop telling the agent to pass per-check `tools` as `delegate.extensions` — that filter matches extension names (`developer`, `summon`), not tool names (`Read`, `Grep`), so the subagent ended up tool-less.
- `orchestrator.rs`: forward per-check `turn-limit` to spawned `goose run` subprocesses via `--max-turns` (was parsed and dropped on the floor).
- `handler.rs`: `--checks-only --no-orchestrate` no longer no-ops; falls through to the orchestrator's check runner so checks actually execute.
- `orchestrator.rs`: `kill_on_drop(true)` on the spawned `Command` so timed-out children get SIGKILLed instead of leaking and racking up tokens.
- `orchestrator.rs`: `--provider` alone (without `--model`) now drops the per-check `model:`. A Claude-pinned per-check model would 404 against Google's API; only the explicit-CLI model is safe to inherit across providers.
- `orchestrator.rs`: `parse_diff_header_path` handles git's quoted-header form (e.g. `"a/dir/\303\251.txt" "b/dir/..."`) so non-ASCII or spaced paths aren't silently dropped from the per-file main pass. Also paired with `-c core.quotePath=off` on the source side.

**P2 robustness/UX:**
- `check.rs`: tolerate CRLF line endings in the frontmatter terminator.
- `handler.rs`: pass `-c core.quotePath=off` to all `git` invocations so non-ASCII paths return as clean UTF-8 (fixes scope discovery on quoted paths).
- `handler.rs`: `--check-scope` rebases touched paths to the discovery root so candidate scope walking doesn't double-prefix `<scope>/api/...`.
- `handler.rs`: `--severity` is parsed up front and applied to `--checks-only --no-orchestrate` output too.
- `handler.rs`: `--quiet` honored before `print_discovered_summary` (no `discovered N check(s)` chatter to stderr).
- `handler.rs`: reviewer instructions only prepended to `base_prompt` for the legacy single-prompt path; orchestrated mode injects per-subprocess to avoid duplication on every per-file main-pass call.
- `discover.rs`: synthesized `REVIEW.md` checks now use the priority-aware recorder so a user-authored check named `repo-rules.md` isn't silently overwritten.

Adds 9 new tests; full `commands::review` suite is 56/56. `cargo fmt --all -- --check` and `cargo clippy -p goose-cli --all-targets -- -D warnings` clean.

---

🤖 Sent by Joah's AI agent — tracking ticket: AGNTOPS-23

---

## Update: address @lifeizhou-ap review feedback — checks are `SourceType::Agent`, not a new variant (commit 767a718)

Per the suggestion to use `agent` as the surface for the review sub-agents:

- Dropped the `SourceType::Check` enum variant entirely.
- `Check::to_source_entry` now returns `SourceType::Agent` with `properties["kind"] = "check"` so clients can differentiate review checks from `.agents/agents/*.md` agents while sharing one source type.
- `list_sources(SourceType::Agent)` now walks `.agents/checks/` in addition to the existing agent directories, so checks flow through the same listing pipeline as agents.
- The on-disk path stays at `.agents/checks/` for Amp portability — the `goose::checks::{discover, Check}` discovery layer (scope walking, REVIEW.md → virtual `repo-rules`, closer-scope shadowing) is unchanged, and the review orchestrator continues to consume `Check` directly.

This keeps the source schema aligned with the planned ACP "start agent in new session or as sub-agent" RPC so review checks can be a first consumer once that lands, without blocking on it.

Adds 1 new test covering check → Agent listing. Full suite: `cargo test -p goose --lib` 1467/1467, `cargo test -p goose-cli --lib` 236/236, `cargo clippy -p goose -p goose-sdk -p goose-cli --all-targets -- -D warnings` clean.


alexhancock (approved):
I'd like to eventually see this land in a plugin as our support for all the components of plugins gets more mature, but this seems really useful and a high value for the code review use-case so stamping to go into goose-cli!

We can revisit alternative implementations as plugin support gets better.

DOsinga:
Hey @joahg — the **Check Rust Code Format** check is failing. Could you run `cargo fmt` and push the result? That should get this green. Thanks!

DOsinga:
ok, sorry for taking so long

I think you should add this as a new type to crates/goose/src/sources.rs and then reuse the discover and parsing stuff we have going on there. how does that sound?

DOsinga:
happy to discuss in person what that would look like. sources.rs is pretty new and can do with some refactoring itself

joahg:
🤖 Sent by Joah's AI agent:

Done in 71805be — moved `Check` parsing and discovery into a new `goose::checks` module that lives next to `sources` / `skills` and reuses `sources::parse_frontmatter` for YAML.

- Added `SourceType::Check` to the SDK so checks can flow through the same listing pipeline as skills and projects (`Check::to_source_entry` stuffs per-check tunables — `model`, `turn-limit`, `tools`, `severity-default`, `scope_dir` — into `SourceEntry.properties`).
- Wired `SourceType::Check` into `sources::list_sources` so e.g. the workbench/UI can list discovered checks for a repo without depending on goose-cli.
- Deleted `goose-cli/src/commands/review/{check,discover}.rs`; `handler.rs`, `orchestrator.rs`, and `prompt.rs` now import from `goose::checks`.

Net: -719 / +33 in `goose-cli` plus the new module in `goose`. 19 `checks` tests + 37 `commands::review` tests pass; `cargo fmt` and `cargo clippy --all-targets -- -D warnings` clean.

lifeizhou-ap:
Hi @joahg, 

Thanks for the contributions!  Just wondering instead of creating a specific source type, shall we use `agent` as a specific agent to do the code review.  We are going to implement this goose acp protocol to start the agent in a new session or as a sub agent soon.   cc @DOsinga

joahg:
🤖 Sent by Joah's AI agent:

Thanks @lifeizhou-ap! Done in 767a718 — dropped `SourceType::Check`, and checks now surface as `SourceType::Agent` entries (with `properties["kind"] = "check"` so the UI can differentiate). `list_sources(SourceType::Agent)` now walks `.agents/checks/` alongside the existing agent dirs, so they flow through the same listing pipeline.

Kept the on-disk path at `.agents/checks/` for Amp portability, and the `goose::checks::{discover, Check}` discovery layer (scope walking, REVIEW.md → virtual `repo-rules`, closer-scope shadowing) plus the orchestrator are unchanged.

This lines the source schema up with the upcoming ACP "start agent in new session or as sub-agent" RPC so review checks can be a first consumer once that lands — without blocking on it. Happy to also migrate the orchestrator off subprocesses onto the ACP path as a follow-up once the RPC is ready; just give me a heads up.

PR #9123, opened 2026-05-09
fix: omit max_tokens for OpenAI-compatible requests when unset

### LINKED ISSUE #9007: Default max_output_tokens of 4096 is unreasonably low
**Describe the bug**

When a user installs Goose and uses one of the local providers (llama_swap, lmstudio, etc.) to point it at any model that's not in a known list, the "max_tokens" value defaults to 4096.  This short limit is sent as a parameter to the OpenAI API, and if the endpoint honors it, nontrivial responses are truncated and the user won't understand why -- it just seems like Goose barely works at all.

The user can work around this by setting `GOOSE_MAX_TOKENS` to a reasonable value in ~/config/goose/config.yaml if they realize what's going on, but the chances of the average user figuring out why it doesn't work are very low. The user experience is just that the model keeps stopping in the middle of trying to do things, with no indication of why.

---

**To Reproduce**
1. Install a local inference server (I've been using oMLX) and download some random models
2. Set up a provider that points to it (`llama_swap` should work with the proper host url)
3. Pick a model and start a chat
4. Run some prompts that generate large outputs (I was using a Qwen-3.6 variant to explore the Ollama codebase)

---

**Expected behavior**
The output should be complete, and the model should be able to complete complex tasks

---

**Actual behavior**
The model keeps stopping mid-tool-use and can't complete complex tasks

---

**Screenshots**
N/A

---

- **OS & Arch:** Mac OS X 26.4.1 on Apple Silicon
- **Interface:** UI
- **Version:** latest (1.33.1 / tip of master from github)
- **Extensions enabled:** default set, but doesn't seem to matter
- **Provider & Model:** llama_swap with any downloade Qwen variant (I'm using Qwen3.6-35B-A3B-MLX-VL-oQ8-FP16)

---

**Additional context**
N/A

SarthakB11:
## Summary

For local OpenAI-compatible providers (`llama_swap`, `lmstudio`) the OpenAI request builder always sets `max_tokens` / `max_completion_tokens` to whatever `ModelConfig::max_output_tokens()` returns. When the user-supplied model has no canonical record and `GOOSE_MAX_TOKENS` is unset, that value is the legacy 4096 fallback, and local servers honor it and truncate non-trivial responses mid-tool-use. The user has no way to find this without discovering `GOOSE_MAX_TOKENS`.

This change skips emitting the field in `providers/formats/openai.rs::create_request` when `model_config.max_tokens` is `None`, letting the server pick its own ceiling. Behavior is unchanged when `GOOSE_MAX_TOKENS` is set or the model has a canonical record (which populates `max_tokens` via `with_canonical_limits`).

Only the shared OpenAI request builder is changed. The Ollama provider also routes through this `create_request` (re-exported at `formats/ollama.rs:24-26`); for unknown models with `GOOSE_MAX_TOKENS` unset, `apply_ollama_options` will simply leave `options.num_predict` unset, letting the Ollama server pick its own default — same fix, applied for free. Other format builders (anthropic, databricks, snowflake, openai_responses) still call `max_output_tokens()` and so still send the 4096 fallback; those APIs require `max_tokens`, so they're intentionally untouched.

### Testing

New unit test `test_create_request_omits_max_tokens_when_unset` asserts that an unknown model with `max_tokens: None` produces a payload containing neither `max_tokens` nor `max_completion_tokens`. Existing tests still pass.

```
cargo test -p goose --lib --no-default-features --features rustls-tls \
  providers::formats::openai providers::ollama
```

### Related Issues

Fixes #9007

### DIFF
--- a/crates/goose/src/providers/formats/openai.rs
+++ b/crates/goose/src/providers/formats/openai.rs
@@ -1029,15 +1029,22 @@ pub fn create_request(
         }
     }
 
-    let key = if is_reasoning_model {
-        "max_completion_tokens"
-    } else {
-        "max_tokens"
-    };
-    payload
-        .as_object_mut()
-        .unwrap()
-        .insert(key.to_string(), json!(model_config.max_output_tokens()));
+    // Only emit max_tokens / max_completion_tokens when the user (via
+    // GOOSE_MAX_TOKENS) or a canonical model record has supplied a value.
+    // For unknown models on OpenAI-compatible endpoints (e.g. llama_swap,
+    // lmstudio) sending the historic 4096 default truncates non-trivial
+    // responses; omitting the field lets the server use its own max.
+    if let Some(max_tokens) = model_config.max_tokens {
+        let key = if is_reasoning_model {
+            "max_completion_tokens"
+        } else {
+            "max_tokens"
+        };
+        payload
+            .as_object_mut()
+            .unwrap()
+            .insert(key.to_string(), json!(max_tokens));
+    }
 
     if for_streaming {
         payload["stream"] = json!(true);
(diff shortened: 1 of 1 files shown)


DOsinga (approved):
Clean, well-scoped fix that solves a real pain point for local provider users. The conditional omission of max_tokens when unset is the right approach — lets the server decide its own limits rather than imposing an arbitrary 4096 cap. Test follows existing patterns. LGTM.

monroewilliams:
Thanks for fixing this! :)

PR #9133, opened 2026-05-11
fix: deduplicate tool_result blocks to prevent LLM API 400 errors

allanm-ops:
## Problem

When using Goose with the Databricks provider (or direct Anthropic), long sessions with many tool calls can fail with:

```
HTTP 400: each tool_use must have a single result. Found multiple tool_result blocks with id: <id>
```

The Anthropic API requires **exactly one `tool_result` per `tool_use`**, but Goose can produce duplicate `tool_result` blocks for the same `tool_use` ID.

## Root Cause

Duplicates enter the conversation history through several paths:

1. **`merge_consecutive_messages()`** — combines two consecutive user/tool messages that both contain a `ToolResponse` for the same ID into a single message with duplicate responses
2. **Session compaction** — can reconstruct the conversation with duplicates during context window management
3. **Session restore from SQLite** — can produce duplicate entries, especially after interrupted tool calls

The existing `fix_tool_calling()` processor handles **orphaned** tool responses (no matching request) but does not deduplicate responses that share the same `tool_use` ID.

Neither the Databricks nor Anthropic message formatters (`format_messages()`) check for duplicates — they blindly emit a `tool`/`tool_result` block for every `ToolResponse` in the message content.

Note: Other Goose frontends (kgoose backend, goosed-slackbot, sales-bridge) already have their own dedup logic for this exact issue. The goosed core was the missing piece.

### 1. New `deduplicate_tool_responses()` processor (primary fix)

Added to the `fix_messages` pipeline, running **after** `merge_consecutive_messages` (which is the main source of duplicates). For each user message, it:
- Walks content items in reverse order
- Tracks seen tool response IDs in a `HashSet`
- Removes earlier duplicates, keeping the **last** response for each ID (matching kgoose's dedup strategy)

### 2. Defense-in-depth in formatters

Added `HashSet`-based dedup in both `databricks.rs` and `anthropic.rs` `format_messages()` functions, so even if a duplicate somehow survives the conversation pipeline, it won't be sent to the API.

## Tests

Three new tests in `conversation/mod.rs`:

| Test | Scenario |
|------|----------|
| `test_duplicate_tool_responses_are_deduplicated` | Two `ToolResponse` items with the same ID in a single message — verifies the last one is kept |
| `test_duplicate_tool_responses_across_merged_messages` | Two consecutive user messages with the same tool response ID — verifies merge + dedup work together |
| `test_no_false_positive_dedup_for_different_tool_ids` | Two different tool IDs in the same message — verifies they are NOT deduplicated |

## Files Changed

- `crates/goose/src/conversation/mod.rs` — new `deduplicate_tool_responses()` processor + 3 tests
- `crates/goose/src/providers/formats/databricks.rs` — defense-in-depth dedup in `format_messages()`
- `crates/goose/src/providers/formats/anthropic.rs` — defense-in-depth dedup in `format_messages()`

### DIFF
--- a/crates/goose/src/conversation/mod.rs
+++ b/crates/goose/src/conversation/mod.rs
@@ -208,6 +208,7 @@ fn fix_messages(messages: Vec<Message>) -> (Vec<Message>, Vec<String>) {
         fix_empty_tool_results,
         fix_tool_calling,
         merge_consecutive_messages,
+        deduplicate_tool_responses,
         fix_lead_trail,
         populate_if_empty,
     ]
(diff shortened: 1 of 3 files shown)


DOsinga:
Thanks for the contribution! However, I'm closing this PR for a few reasons:

1. **No prior issue or discussion** — Per our [contributing guidelines](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#discussions-issues-and-prs), feature-level changes should start with an issue or discussion so we can align on the approach before code is written.

2. **Unaddressed Codex review comment** — The P1 comment about pipeline ordering hasn't been addressed. Per our [AI Code Reviews policy](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#ai-code-reviews), review comments need to be resolved or explained.

3. **A question about the premise** — `fix_tool_calling()` already runs *before* your new processor in the pipeline and handles orphaned tool responses (those without a matching request). If two consecutive messages carry a `ToolResponse` for the same ID, the first one satisfies the pending request and the second becomes an orphan — which `fix_tool_calling` would strip. So in what scenario does a duplicate actually survive to reach the formatter? Could you describe the concrete reproduction steps or session state where you hit the 400 error? Without that it's hard to evaluate whether the fix is solving a real problem or papering over something that's already handled.

If you can open an issue describing the exact reproduction (ideally with a minimal session transcript showing the duplicate), we'd be happy to discuss the right fix. A smaller, targeted PR with a clear test that fails without the fix would be the way to go.

Thanks for your interest in improving goose!

PR #9137, opened 2026-05-11
refactor local inference around backends

jh-block:
**Category:** improvement
**User Impact:** Users should see the same local inference behavior while goose gains a cleaner foundation for future local runtimes.
**Problem:** The local inference provider was tightly coupled to llama.cpp implementation details, making it hard to add another in-process backend such as MLX without mixing runtime-specific code into the provider. This also made model caching, generation, and memory reporting assume one backend globally.
**Solution:** This refactors local inference around a `LocalInferenceBackend` trait and moves llama.cpp-specific loading, generation, memory detection, and tool/vision handling into a dedicated backend module. The provider now routes through backend-neutral request and loaded-model abstractions while preserving the existing llama.cpp behavior as the default backend.

<details>

**crates/goose/src/providers/local_inference.rs**
Refactors the top-level provider into backend orchestration: it owns backend registration, backend-aware model cache keys, request setup, and stream plumbing while delegating runtime-specific work to a backend implementation.

**crates/goose/src/providers/local_inference/backend.rs**
Adds the backend-neutral traits and generation request type that future local inference runtimes can implement without changing provider flow.

**crates/goose/src/providers/local_inference/llamacpp/mod.rs**
Adds the llama.cpp backend implementation, preserving current model loading, device memory reporting, multimodal setup, and native/emulated generation routing behind the new trait.

**crates/goose/src/providers/local_inference/llamacpp/inference_emulated_tools.rs**
Moves the emulated tool generation path under the llama.cpp backend and updates it to use the backend context instead of the provider runtime.

**crates/goose/src/providers/local_inference/llamacpp/inference_engine.rs**
Moves the llama.cpp generation engine under the backend module and routes context/memory operations through `LlamaCppBackend`.

**crates/goose/src/providers/local_inference/llamacpp/inference_native_tools.rs**
Moves native tool-call generation under the llama.cpp backend and updates context creation calls to use the backend abstraction.

</details>

### Reproduction Steps

1. Build or run goose with local inference enabled.
2. Open the local inference settings or use the local provider with an existing downloaded GGUF model.
3. Confirm the model still loads through the existing llama.cpp path and streams responses as before.
4. Inspect `crates/goose/src/providers/local_inference/backend.rs` and `llamacpp/mod.rs` to see the new backend boundary for future MLX support.

### Validation

- `cargo fmt`
- `cargo fmt --check`
- `cargo check -p goose --features local-inference`
- `cargo test -p goose local_inference --features local-inference`

### DIFF
--- a/crates/goose/src/providers/local_inference.rs
+++ b/crates/goose/src/providers/local_inference.rs
@@ -1,18 +1,9 @@
+mod backend;
 pub mod hf_models;
-mod inference_emulated_tools;
-mod inference_engine;
-mod inference_native_tools;
+mod llamacpp;
 pub mod local_model_registry;
 pub(crate) mod multimodal;
 mod tool_parsing;
-
-use inference_emulated_tools::{
-    build_emulator_tool_description, generate_with_emulated_tools, load_tiny_model_prompt,
-};
-use inference_engine::GenerationContext;
-use inference_engine::LoadedModel;
-use inference_native_tools::generate_with_native_tools;
-use tool_parsing::compact_tools_json;
 
 use crate::config::ExtensionConfig;
 use crate::conversation::message::{Message, MessageContent};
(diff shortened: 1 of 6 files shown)


michaelneale (approved):
nice refactor too

PR #9147, opened 2026-05-11
protocol cleanup

jamadeo:
Clean up the ACP+ protocol a bit. Add a version tag to the custom methods and improve some types.


crates/goose-sdk/src/custom_requests.rs
```diff
@@ -881,25 +881,33 @@ impl SourceEntry {
     }
 }
 
+/// Target scope for creating or importing sources.
+#[derive(Debug, Default, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(tag = "scope", rename_all = "camelCase")]
```
lifeizhou-ap, line 886: nice!

lifeizhou-ap (approved):

alexhancock (approved):
It looks good. I do think we should have a planned timeline on which we expect to progress things to stable, but that can come at another time.

jamadeo:
> It looks good. I do think we should have a planned timeline on which we expect to progress things to stable, but that can come at another time.

Yeah, 100% agree. If anything I think this encourages us to stabilize -- it makes the api uglier! But right now we have a consumer of ACP+ that isn't part of this repo, so we want to make sure these are able to break compatibility.

PR #9151, opened 2026-05-12
fix(openai): apply request_params to outgoing API payload

### LINKED ISSUE #9083: `extra_body` settings not applied to OpenAI-compatible provider requests
## Title
`extra_body` in recipe settings is parsed but not merged into API request for OpenAI-compatible providers

## Description

When setting `extra_body` in recipe `settings` for OpenAI-compatible providers (e.g., Mistral), the configuration is correctly parsed into `request_params` on `ModelConfig`, but it is **not merged into the actual request payload** sent to the API.

### Impact
This prevents users from passing provider-specific parameters like:
- Mistral's `reasoning_effort` ("high", "none") for adjustable reasoning models
- Other OpenAI-compatible providers' custom request parameters

### Reproduction Steps

1. Create a recipe with `extra_body` in settings:
```yaml
settings:
  goose_provider: mistral
  goose_model: mistral-medium-3-5
  extra_body:
    reasoning_effort: "high"
```

2. Check the LLM request logs (`llm_request.0.jsonl`)
3. Observe `model_config.request_params` contains `{"reasoning_effort":"high"}`
4. Observe the actual `input` payload sent to the API does **not** include `reasoning_effort`

### Current Behavior
- `extra_body` in recipe YAML is parsed ✓
- `request_params` field on `ModelConfig` is populated ✓
- `request_params` is merged into the API request payload ✗

### Expected Behavior
`request_params` should be merged into the request payload before sending to the API, similar to how OpenCode's `options` field works.

### Root Cause
In `crates/goose/src/providers/formats/openai.rs`, the `create_request` function builds the payload but never iterates over or merges `request_params` from `ModelConfig`. The field exists but is unused.

### Proposed Fix
After building the `payload` object in `create_request`, merge `request_params`:

```rust
// After payload is constructed, before returning
if let Some(params) = &model_config.request_params {
    if let Some(obj) = payload.as_object_mut() {
        for (key, value) in params {
            obj.insert(key.clone(), value.clone());
        }
    }
}
```

### Example Use Case
Enable Mistral's native reasoning mode:
```yaml
settings:
  goose_provider: mistral
  goose_model: mistral-medium-3-5
  extra_body:
    reasoning_effort: "high"
```

### Environment
- Goose Desktop (v1.33.1+, supports `extra_body`)
- OS: Windows 11
- Provider: Mistral AI

### Related
- OpenCode implementation: `options` field merged into API requests
- Issue #9041: DeepSeek reasoning_content handling (similar parameter passing issue)

TurboWindX:
## Problem
  `request_params` is defined in `ModelConfig` and documented as supporting
  provider-specific parameters (e.g. `thinking`, `extra_body` fields), but
  `create_request()` in `formats/openai.rs` never reads it. Any params set
  in config are silently discarded before the request is sent.

  This breaks use cases like disabling DeepSeek V4 thinking mode:
  ```yaml
  request_params:
    thinking:
      type: disabled
  The payload goes out without the field, causing a 400: reasoning_content must be passed back error on subsequent
  turns.

  Fix

  Merge request_params into the payload just before return in
  create_request(). Added a test covering both the merge and field
  override behaviour.

  Testing

  cargo test -p goose --lib -- test_create_request_merges_request_params
  test result: ok. 1 passed

  Fixes #9083
  Related to #904


DOsinga (approved):
LGTM — the fix correctly merges `request_params` into the OpenAI format payload, matching the existing pattern in `databricks.rs`. Dropped the test and merged main.

One thing: your original commit is missing the DCO sign-off. Please amend it with `git commit --amend -s` and force-push so the DCO check passes.

PR #9152, opened 2026-05-12
fix: gate tool-pair summarization to context windows <= 64K

michaelneale:
I don't find tool summarization is a good idea - at most it should be for only when context is really constrained. There is compaction and tool calls are already limited. This has repeatedly been a problem when you have a longer session with a large model and many tool calls, making goose really harmfully bad. 

This will gate it to context windows <= 64K where it might still help. No behavior change for small-context models. Users can still disable entirely with GOOSE_TOOL_PAIR_SUMMARIZATION=false.


jh-block:
Can we just turn this off completely for now, we can do some testing with tiny models and turn it on there **if** it helps and doesn't make them perform even worse.

michaelneale:
@jh-block yeah - I was kind of assuming it would help for small ones but perhaps not (I expect modest models - probably won't make it to 100+ tool cools at this context size to see the pain here, but I agree).

michaelneale:
@jh-block PTAL now

PR #9191, opened 2026-05-13
fix: persist accumulated cost in session DB to survive reload

### LINKED ISSUE #6141: Accumulated pricing after model changes does not persist across session reload
**Describe the bug**

There is some client side code to handle this but if I were to use:

cheap model
- talk to goose a bunch
expensive model
- talk to goose a bunch

reload session

It has no recollection of what was part of cheap vs expensive model.

I think my ideal fix would be to attach some metadata to the session around what provider/model a session started with and when.switches occur. Something simpler could be to just store the accumulated cost, not just the accumulated tokens.

DOsinga:
## Summary

Fixes #6141

When users switch models mid-session, the cost tracking was computed client-side in an ephemeral React hook (`useCostTracking`). On session reload (page refresh), this in-memory state was lost and cost was incorrectly recalculated using the current model's pricing against all accumulated tokens.

## Fix

Compute cost at the time tokens are generated (when we know which model produced them), persist it in a new `accumulated_cost` column on the sessions table, and expose it via `TokenState` to the frontend.

**Backend:**
- Added `accumulated_cost` (f64) column to sessions table (schema v13)
- In `update_session_metrics`, compute the cost for each usage chunk using the canonical model pricing for the model that generated those tokens, and accumulate it
- Exposed `accumulated_cost` in `TokenState` SSE events

**Frontend:**
- `CostTracker` now uses the backend-persisted `accumulatedCost` directly
- Removed the ephemeral `useCostTracking` hook that lost state on reload
- Falls back to client-side calculation when backend cost is not yet available

### Type of Change
- [x] Bug fix

### Testing
- All Rust tests pass (59 session tests)
- All desktop UI tests pass (346 tests)
- TypeScript compiles cleanly
- Clippy clean


ui/desktop/src/components/bottom_menu/CostTracker.tsx
```diff
@@ -138,9 +111,7 @@ export function CostTracker({
 
     const inputCost = (inputTokens * (costInfo.input_token_cost || 0)) / 1_000_000;
     const outputCost = (outputTokens * (costInfo.output_token_cost || 0)) / 1_000_000;
-    const total = inputCost + outputCost;
-
-    return total;
+    return inputCost + outputCost;
```
jamadeo: shouldn't we move this part to goosed too? and get rid of all calculation on the client?

jamadeo (approved):

PR #9193, opened 2026-05-13
refactor(goose2): narrow project property reads instead of unchecked casts

enilsen16:
## Summary

`features/projects/api/projects.ts` was using unchecked `as` casts to pull fields off the project shape returned by ACP. This replaces those with narrow per-property reads that return safe fallbacks instead of casting the entire response object.

- Replaces `project as Project` casts with direct property reads (`project?.name ?? ""`, etc.)
- Removes the local `Project` type alias that was shadowing the SDK type
- No behaviour change; types are stricter

## Test plan
- [ ] Existing project functionality works (open project, session context, etc.)

---
_Reopened from #9169 — rebased on current main._

### DIFF
--- a/ui/goose2/src/features/projects/api/projects.ts
+++ b/ui/goose2/src/features/projects/api/projects.ts
@@ -29,22 +29,37 @@ interface SourceEntry {
   properties: Record<string, unknown>;
 }
 
+function asString(v: unknown): string | null {
+  return typeof v === "string" ? v : null;
+}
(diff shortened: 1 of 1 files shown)


jamadeo:
Thank you @enilsen16 , but this particular implementation of the app is going away. Fore more information please see https://github.com/aaif-goose/goose/discussions/9367 on the future plans for revamping the desktop app.

Issue #8864, opened 2026-04-27
Canceling recipe parameter form spawns new extra window instead of closing the modal overlay


mxmlnkn:
**Describe the bug**

Some actions spawn new windows instead of updating the current one.

---

**To Reproduce**
1. Add a recipe with arguments
2. Run that recipe
3. On the "Recipe Parameters" overlay, press Cancel, then press "Start New Chat"

---

**Expected behavior**
The modal dialog overlay "Cancel Recipe Setup" should disappear (left side in the screenshot).  Instead, a new window pops up (see the right side in the screenshot).

---

**Screenshots**

<img width="1676" height="717" alt="Image" src="https://github.com/user-attachments/assets/7939079f-7327-467f-a31e-74f7ebfd780b" />

---

**Please provide the following information**
- **OS & Arch:** Ubuntu 24.04 x86-64
- **Interface:** UI
- **Version:** 1.32.0
- **Extensions enabled:** 
- **Provider & Model:** OpenAI compatible custom provider

--

DOsinga:
Thanks for the clear report! This is a confirmed bug — in `ParameterInputModal.tsx`, the "Start New Chat" cancel option calls `createChatWindow()` + `hideWindow()` instead of just dismissing the modal and resetting the current window to a fresh chat state.

The fix should be straightforward: replace the new-window creation in `handleCancelOption('new-chat')` with `onClose()` (or equivalent state reset), so the modal closes and the current window continues as a normal chat.

Snoozing this for a couple of weeks — if anyone wants to pick this up, a PR would be welcome!

PR #9194, opened 2026-05-13
feat(analyze): add Elixir support and clarify tool description

enilsen16:
## Summary

Adds Elixir to the `analyze` platform extension alongside the existing languages (Python, JS/TS, Ruby, etc.).

- Language detection via `.ex`/`.exs` extensions
- `fn_kinds`/`class_kinds` set to `&["call"]` — Tree-sitter sees both `def`/`defp` and `defmodule` as `call` nodes in Elixir's grammar
- `elixir_def_name`: extracts function names from `def`/`defp` call nodes
- `elixir_module_def_name`: extracts module names from `defmodule`/`defprotocol`/`defimpl`
- `find_enclosing_class` walks past Elixir function `call` nodes so it returns the enclosing module, not the function itself
- Module attributes (`@moduledoc`, `@doc`, etc.) excluded from call capture
- `defimpl` included in class extraction so protocol implementations are tracked

Review comments from #8998 / #9159 have been addressed.

## Test plan
- [ ] `analyze` correctly attributes Elixir function calls and module membership
- [ ] `MyMod.foo` parenting test passes (no `foo.foo` self-parenting regression)

---
_Reopened from #9159 — rebased on current main, fixes squashed into single commit._

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4908,6 +4908,7 @@ dependencies = [
  "tracing-opentelemetry",
  "tracing-subscriber",
  "tree-sitter",
+ "tree-sitter-elixir",
  "tree-sitter-go",
  "tree-sitter-java",
  "tree-sitter-javascript",
@@ -11917,6 +11918,16 @@ dependencies = [
  "regex-syntax",
  "serde_json",
  "streaming-iterator",
+ "tree-sitter-language",
+]
+
+[[package]]
+name = "tree-sitter-elixir"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "66dd064a762ed95bfc29857fa3cb7403bb1e5cb88112de0f6341b7e47284ba40"
+dependencies = [
+ "cc",
  "tree-sitter-language",
 ]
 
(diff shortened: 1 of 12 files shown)


DOsinga:
Hey @enilsen16 — first off, apologies it took us so long to get around to looking at this. Thanks for your patience.

Having gone through it, the main thing is that this PR bundles three pretty distinct changes together:

1. **Elixir support for `analyze`** — the headline feature, and it looks well thought through.
2. **An agent loop-breaker** (`tool_monitor.rs` / `tool_inspection.rs`, repeated-tool-error detection + recovery hints).
3. **A desktop fix** for cancelling the recipe parameter form (#8864 / `ParameterInputModal.tsx`).

Could you split these into separate PRs? It'd make each one much easier to review and land. A couple of notes:

- For the **Elixir/analyze** PR (which I'd suggest leading with), there are two unaddressed [codex](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#ai-code-reviews) comments worth resolving — in particular the `defimpl` one. The description mentions `defimpl` is included in class extraction, but the query currently only matches `defmodule`/`defprotocol`.
- For the **agent loop-breaker** piece: heads up that we're currently in the middle of reworking the agent loop, so that one might be best held back for now to avoid churn — happy to point you at where things land once it settles.

Going to close this for now, but please reopen / open the split PRs whenever you're ready. Really appreciate the contribution!

PR #9197, opened 2026-05-13
Add typed config schema with ACP custom methods

wpfleger96:
## Summary

Adds a comprehensive typed config schema (`GooseConfigSchema`) covering all `config_value!` keys plus structured blocks (`extensions`, `slash_commands`, `experiments`, `providers`), exposed via ACP custom methods `_goose/unstable/config/read` and `_goose/unstable/config/write`.

- `GooseConfigSchema` derives `schemars::JsonSchema` for standalone `config.schema.json` generation, with `from_config()` / `apply_to_config()` methods that read/write through the `active_provider` resolution chain for `GOOSE_PROVIDER`/`GOOSE_MODEL`
- Parallel DTO types (`ConfigSchemaDto`, `ExtensionConfigDto`, etc.) in `goose-sdk` bridge to `GooseConfigSchema` via serde, following the established cross-crate pattern (e.g., `ProviderInventoryEntryDto`)
- TypeScript SDK gets `GooseConfigRead()` / `GooseConfigWrite()` on `GooseExtClient` with Zod validators via the ACP codegen pipeline
- Sparse patch semantics on write: only non-`None` fields are persisted; `null` and missing are both no-ops (documented on `ConfigWriteRequest` and `apply_to_config`)
- `from_config` returns the resolved config view (env vars override file values); documented as intentional since sparse-patch writes prevent accidental persistence of env-only overrides
- CI step (`just check-config-schema`) ensures `config.schema.json` stays in sync with Rust types
- Serde round-trip drift test covers flat scalar fields plus nested `extensions`, `providers`, and `experiments` blocks

### Security & data consistency

- `Envs` struct uses a custom `Deserialize` impl that routes through `Envs::new()`, enforcing the 31-key `DISALLOWED_KEYS` blocklist (`PATH`, `LD_PRELOAD`, `DYLD_INSERT_LIBRARIES`, etc.) on all deserialization paths
- Extension writes merge caller-supplied entries with existing hidden platform extensions instead of replacing the whole map, preventing a read-modify-write cycle from permanently deleting hidden extension state
- `active_provider` and legacy `GOOSE_PROVIDER` flat key are both cleared when a providers-map write excludes the currently-active provider, preventing dangling references through all fallback paths in `get_active_provider()`

### Type safety

- All timeout/limit/retry `u64` fields changed to `u32` (values are well within range), fixing a TypeScript mismatch where Zod produced `bigint` but the TS type said `number`
- `GOOSE_MODE` uses a `schemars(schema_with)` annotation to generate a proper `'auto' | 'approve' | 'smart_approve' | 'chat'` union type in TypeScript instead of plain `string`


jamadeo:
Hey @wpfleger96 I am all for stronger typing, but in general we are moving away from the HTTP API in favor of what ACP: https://github.com/aaif-goose/goose/discussions/9173

We could explore adding stronger types around config within that protocol.

wpfleger96:
hey @jamadeo thanks for the direction here, I definitely want to stay aligned with the direction toward ACP! 

the core of this PR is the `GooseConfigSchema` struct and the `from_config()`/`apply_to_config()` logic, which are transport-agnostic and the REST endpoints were thin wrappers. I'm thinking I could rework this to drop the REST endpoints entirely and expose the typed config through ACP as `_goose/config/read` and `_goose/config/write` custom methods, following the same `_goose/*` extension pattern used by preferences, providers, etc.

is this more aligned with what you're thinking?

PR #9198, opened 2026-05-13
Prompt injection mitigation: update pattern-based detection

dorien-koelemeijer:
## Summary
This PR aims to decrease false positives in pattern-based detection, especially in `rm -rf` commands. Whenever the command injection classifier is not available (i.e. not enabled in settings or WARP disabled), there is an automatic fallback to pattern-based detection. The command injection classifier has reasonably high accuracy, but the pattern-based detection is harder to get right. 

### Changes
* Removed rm_rf_system: Matched any rm -rf targeting system directory names (bin, etc, usr, var, tmp, opt, etc.). In practice this fired on build cleanup (rm -rf build), temp file deletion (rm -f /tmp/file.json), and Homebrew management (rm -rf /opt/homebrew/...). The rm_rf_root_bare pattern still catches the truly catastrophic rm -rf / and rm -rf /*.
* Removed alternative_shell_invocation: Matched any bash -c or sh -c containing ;, |, or &. This is normal shell usage (build scripts, CI commands, conditional checks). The ML classifier distinguishes actual injection from legitimate shell invocations.
* Simplified log_manipulation: Removed > /dev/null from the pattern. Stderr suppression (2>/dev/null) is standard in virtually every shell command. Kept the actual log tampering patterns: truncate.*log, rm /var/log/, echo > /var/log. Also fixed the rm flag regex from (-[rf]\s+)* to (-[rRfF]+\s+)* to correctly match rm -rf /var/log/... (aligned with rm_rf_root_bare).

### Estimated impact
Eliminates ~257 of ~310 weekly pattern-based flags (~83%) while all true positives remain covered by:
* The ML classifier (primary path)
* rm_rf_root_bare for rm -rf /
* curl_bash_execution for curl | bash
All other 20+ patterns unchanged (reverse shells, SUID, credential access, data exfil, etc.) 

### Testing
Unit tests.

### DIFF
--- a/crates/goose/src/security/patterns.rs
+++ b/crates/goose/src/security/patterns.rs
@@ -55,7 +55,7 @@ pub const THREAT_PATTERNS: &[ThreatPattern] = &[
     },
     ThreatPattern {
         name: "rm_rf_home_or_root",
-        pattern: r"rm(\s+(--[a-zA-Z][a-zA-Z\-]*|--|-[a-zA-Z]+))+\s+['\x22]?(~|\$HOME|\$\{HOME\}|/home|/root)/?(\*)?['\x22]?(\s|[;&|]|$)",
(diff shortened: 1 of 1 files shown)


lifeizhou-ap:
Hi @dorien-koelemeijer,

Looks good! Would you please address the above codex comments before merge the PR?

Thanks!

lifeizhou-ap (approved):

michaelneale (approved):

shellz-n-stuff (approved):

PR #9199, opened 2026-05-14
feat(acp): paginate session list

kalvinnchau:
## summary

implements acp pagination for `session/list`, per https://agentclientprotocol.com/protocol/session-list#pagination

- uses a fixed backend page size of 50
- returns `nextCursor` when more sessions are available
- accepts `cursor` from the previous response
- honors `cwd` as an exact absolute-path filter
- rejects malformed cursors, relative `cwd`, and cursor/cwd mismatches with `invalid_params`

## changes
https://github.com/aaif-goose/goose/pull/9199/changes#diff-fc56f9761e5df2e2d8fdb2e729313a03b0ae5832d0155b665a8f21af5bc36d36R333-R338

cc @alexhancock this does change the behavior of `session/list` from `?` to 
```rust
match agent.on_list_sessions(req).await {
      Ok(response) => responder.respond(response)?,
      Err(e) => responder.respond_with_error(e)?,
  }
```

i'm unsure if this might break the existing TUI, i can't find the exact usage in `ui/text/*`

## notes

this is a behavioral change for acp clients: `session/list {}` now returns the first page instead of every visible session.
clients must follow `nextCursor` to retrieve all sessions.

the cursor stores both `updated_at` and `session_id` because goose timestamps can collide at second precision. this avoids
skipping or duplicating sessions when multiple rows share the same timestamp.

### DIFF
--- a/crates/goose/src/acp/server.rs
+++ b/crates/goose/src/acp/server.rs
@@ -21,7 +21,7 @@ use crate::providers::inventory::{
     InventoryIdentity, ProviderInventoryEntry, ProviderInventoryService, RefreshJobPlan,
     RefreshPlan, RefreshSkipReason,
 };
-use crate::session::session_manager::SessionType;
+use crate::session::session_manager::{SessionListCursor, SessionType};
 use crate::session::{EnabledExtensionsState, Session, SessionManager};
 use crate::source_roots::SourceRoot;
 use crate::utils::sanitize_unicode_tags;
@@ -51,6 +51,7 @@ use agent_client_protocol::{
     Responder,
 };
 use anyhow::Result;
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine as _};
 use fs_err as fs;
 use futures::future::{BoxFuture, Either};
 use futures::stream::{self, StreamExt};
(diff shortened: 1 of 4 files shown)


DOsinga:
thanks -- this PR is the right idea but I think we need to clean up a few things here before we merge:

`list_nonempty_sessions_by_types_paged` re-implements the column list, JOIN, GROUP BY, and bind logic of `list_sessions_by_types`. Refactor to one query function with optional filters/cursor/limit, otherwise this is going to be a continuous maintenance burden.

`insert_session_for_list` looks like a method that is only there for testing, but that in effect means we are testing what that method does, not what happens in the wild.

moving `session/list` from `?` to `respond_with_error` is a behavior change for existing clients. Mention it in the PR description and cc @alexhancock to see if this breaks the existing TUI library. 

I would also hash the filters into the cursor to make this more future proof when we want to add more to the filter -- like the session type.

kalvinnchau:
thanks for the review @DOsinga !

* refactored session listing through a shared query helper
* removed the `insert_session_for_list` test helper, and force tests through `sessionManager`
* added a cursor filter hash so cursors are tied to the filter list

crates/goose/src/acp/server.rs
```diff
@@ -253,6 +259,94 @@ fn sid_short(id: &str) -> String {
     id.chars().take(8).collect()
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+struct SessionListCursorToken {
```
alexhancock, line 263: We should probably define these in `crates/goose-sdk` so the client gets things typegened for it on the TS side as well

alexhancock (approved):

crates/goose/src/acp/server.rs
kalvinnchau: chatted with @alexhancock - this is the internal server cursor token, which we send back up to the client as an opaque token string, so we don't need/want the clients to use this

PR #9204, opened 2026-05-14
fix: build Linux binaries in Ubuntu 22.04 container for glibc 2.35 compat

### LINKED ISSUE #9203: v1.34.0 requires libc6 2.38
**Describe the bug**

I'm running Ubuntu 22.04 LTS, which provides libc6 2.35. After updating goose to v1.34.0, it no longer starts:

```
./goose: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by ./goose)
./goose: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by ./goose)
```

Downgrading to v1.33.1 fixes the issue. I see that https://github.com/aaif-goose/goose/pull/9075 changed the build to use Ubuntu 24.04 runners. That's unfortunate given that 22.04 is still supported for at least another year or so. I think this is also going to break on Debian 12 and RHEL 8 and 9, all of which are still supported.

---

**To Reproduce**
1. Install v1.34.0 on Ubuntu 22.04 and try to run it.
---

**Expected behavior**
It runs.

---

- **OS & Arch:** Ubuntu 22.04 x86_64
- **Interface:** CLI
- **Version:** v1.34.0

omehegan:
## Summary

PR #9075 switched Linux builds from cross-compiled (with older glibc) to native `ubuntu-24.04` runners, producing binaries that require glibc >= 2.38. This broke compatibility with Ubuntu 22.04 and potentially some other distros that are still supported.

Users on these distros see:
```
./goose: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by ./goose)
./goose: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by ./goose)
```

## Fix

Build the Linux Rust binaries inside an `ubuntu:22.04` Docker container on the existing `ubuntu-24.04` runners. This preserves the native arm64 runner support added in #9075 (no `ubuntu-22.04-arm` hosted runners exist) while producing binaries compatible with glibc 2.35+.

Fixes #9203.

### Changes

- **`build-cli.yml`**: Split Linux builds into a separate `build-cli-linux` job that builds inside an `ubuntu:22.04` container. macOS/Windows jobs are unchanged. Artifact names are preserved so downstream workflows (`release.yml`, `canary.yml`, `pr-comment-build-cli.yml`) work without modification.
- **`bundle-desktop-linux.yml`**: Build the `goosed` binary inside an `ubuntu:22.04` container, then package with Electron on the host `ubuntu-24.04` runner as before. The Electron binary itself is built on Ubuntu 22.04 by the Electron project and is already compatible.
- `glslc` (needed for Vulkan shader compilation) is installed via the LunarG Vulkan SDK repo since it is not in Ubuntu 22.04 base repos.

### Not in scope

`bundle-goose2.yml` has the same `ubuntu-24.04` issue for its Linux job, but that is a separate v2 pipeline — left as a follow-up.

### DIFF
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -66,7 +66,7 @@ jobs:
             ubuntu:22.04 bash -c '
               set -euo pipefail
 
-              # Install build dependencies
+              # Install build dependencies as root
               apt-get update
               DEBIAN_FRONTEND=noninteractive apt-get install -y \
                 build-essential \
@@ -78,7 +78,8 @@ jobs:
                 curl \
                 ca-certificates \
                 wget \
-                gnupg
+                gnupg \
+                sudo
 
               # Install glslc from LunarG Vulkan SDK repo (not available in Ubuntu 22.04 base repos)
               wget -qO - https://packages.lunarg.com/lunarg-signing-key-pub.asc | apt-key add -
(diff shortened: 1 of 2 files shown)


omehegan:
.build-cli

omehegan:
After iterating on the container-based build approach, it's become clear that the glibc compatibility issue is deeper than just the runner choice.

PR #9075 switched Linux builds from `cross`-compiled (older glibc) to native `ubuntu-24.04` runners (glibc 2.39), breaking several LTS distros (Ubuntu 22.04, Debian 12, RHEL 8/9, Amazon Linux 2023, etc.).

Building inside an `ubuntu:22.04` container hits a second problem — the Vulkan headers on Ubuntu 22.04 (SDK 1.3.204) are too old for `llama-cpp-sys-2` v0.1.146, which requires types from ~1.3.275+ (`PipelineRobustnessCreateInfoEXT`, `CooperativeMatrixFeaturesKHR`, `LayerSettingEXT`, etc.). This forces bind-mounting newer Vulkan headers from the host into the container, plus bind-mounting `glslc` (since the LunarG SDK repo is amd64-only, breaking arm64 builds).

In short, this got messy quickly, and I decided to pull the plug.

PR #9206, opened 2026-05-14
Surface resolved Databricks model metadata

jh-block:
## Summary

- add inference metadata to assistant messages so providers can report the resolved model used for a response
- opt Databricks into resolved-model reporting using the model returned in provider usage
- show the latest resolved model in the desktop model dropdown and inline model-change notifications between messages

## Scope

Only Databricks for now, but the mechanism is intentionally general and support can be added to any other providers that report the resolved model (e.g. gpt-5.5 -> gpt-5.5-2026-04-23) in their responses.

## Screenshots

When clicking on the model name in the bottom bar:
image

Disclosure shown inline in the chat when the resolved model changes (either because the user changed the model, or because an alias like 'goose' was updated server-side to point to a different model):
image


DOsinga (approved):
Solid work. The feature threads resolved model metadata (provider, requested model, resolved model) from the Databricks provider through the agent, streaming pipeline, and into the desktop UI — showing the resolved model in the bottom bar and inline model-change disclosures when the resolved model changes mid-conversation.

I pushed a small fix to use `intl.formatMessage` for the two hardcoded UI strings ('Resolved model' in ModelsBottomBar and 'Model changed: ...' in ProgressiveMessageList) and added the corresponding entries to the en.json and zh-CN.json i18n message files.

All checks pass: cargo build, clippy, cargo fmt, tsc --noEmit, pnpm test (347/347).

PR #9209, opened 2026-05-14
feat: support GOOSE_OAUTH_CALLBACK_PORT for stable OAuth redirect_uri

### LINKED ISSUE #6268: Support stable localhost port for oauth2 flow redirect_uri from IDP
We have the same issue as the one described [in this ticket](https://github.com/block/goose/issues/5107): we are using an internal OAuth2 Identity Provider, and we have to configure all the various authorized oauth2 callback uris.
For security reasons, the jocker ('*') is forbidden in the callback uris, so we must set fixed port number for the callback.

Goose seems to have random port for the callbacks, so we are blocked with an error "400 - Invalid redirect_uri" after the authentication.

Would it be possible to configure a fixed port for the return from the identity provider?

It is the case with Copilot, either from VS Code or from IntelliJ IDEA, or for the mcp inspector.

We study the possibility to have a jocker for the port, for locahost uris. It may not be possible for us.

**Additional context**

=> there is another ticket that is mark as closed, the original poster didn't update it after the answer from your team

DOsinga:
## Summary

Allow users to set a fixed localhost port for the OAuth callback server via the `GOOSE_OAUTH_CALLBACK_PORT` environment variable. When set, both OAuth flows (MCP server auth and Databricks provider auth) bind to that port instead of a random ephemeral one.

This unblocks enterprise IdPs that require exact `redirect_uri` matching and forbid wildcard ports in their callback URI whitelist.

### Usage

```bash
export GOOSE_OAUTH_CALLBACK_PORT=8456
```

Then configure your IdP to allow `http://127.0.0.1:8456/oauth_callback` (or `http://localhost:8456` for the Databricks flow) as a redirect URI.

### Changes

- `crates/goose/src/oauth/mod.rs`: Read `GOOSE_OAUTH_CALLBACK_PORT` env var when binding the MCP OAuth callback server
- `crates/goose/src/providers/oauth.rs`: Same for the Databricks OAuth flow (respects explicit redirect_url port first, then env var, then random)

Fixes #6268

### DIFF
--- a/crates/goose/src/oauth/mod.rs
+++ b/crates/goose/src/oauth/mod.rs
@@ -68,7 +68,11 @@ pub async fn oauth_flow(
         .route("/oauth_callback", get(handler))
         .with_state(app_state);
 
-    let addr = SocketAddr::from(([127, 0, 0, 1], 0));
+    let port: u16 = std::env::var("GOOSE_OAUTH_CALLBACK_PORT")
+        .ok()
+        .and_then(|p| p.parse().ok())
+        .unwrap_or(0);
+    let addr = SocketAddr::from(([127, 0, 0, 1], port));
     let listener = tokio::net::TcpListener::bind(addr).await?;
     let used_addr = listener.local_addr()?;
     tokio::spawn(async move {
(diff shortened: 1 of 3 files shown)


alexhancock (commented):
is port the only thing worth making configurable? path? host?

i imagine there may be some security considerations but thought I would ask

alexhancock (approved):

PR #9210, opened 2026-05-14
feat: add Atomic Chat as declarative OpenAI-compatible provider

yanalialiuk:
## Summary

Adds Atomic Chat as a declarative OpenAI-compatible provider: point Goose at the app’s local API ([Atomic-Chat](https://github.com/AtomicBot-ai/Atomic-Chat), default http://localhost:1337/v1) so chat uses the same MLX / llama.cpp models as the desktop client.

## Included changes

- `crates/goose/src/providers/declarative/atomic_chat.json` — provider config (`atomic_chat`), configurable `ATOMIC_CHAT_HOST`, dynamic models from `/v1/models`.
- `documentation/docs/getting-started/providers.md` — table entry and setup (Desktop + CLI).
- `crates/goose/src/config/declarative_providers.rs` — `test_atomic_chat_json_deserializes`.

## Configuration
| Goose / env | Role |
|-------------|------|
| `ATOMIC_CHAT_HOST` | Origin of the Atomic Chat API (default `http://localhost:1337`). |
| Declarative `base_url` | Resolves to `{host}/v1/chat/completions` — same shape as other OpenAI-compatible local providers. |

No API key is wired in this declarative profile; if the local API requires auth, use a **custom OpenAI-compatible provider** and store credentials there.

## Testing
- `cargo test -p goose test_atomic_chat_json_deserializes`
- Manual: start Atomic Chat, load a model, run `goose configure` → **Atomic Chat** → pick a model from `/v1/models`.

### DIFF
--- a/crates/goose/src/config/declarative_providers.rs
+++ b/crates/goose/src/config/declarative_providers.rs
@@ -880,6 +880,37 @@ mod tests {
     }
 
     #[test]
+    fn test_atomic_chat_json_deserializes() {
+        let json = include_str!("../providers/declarative/atomic_chat.json");
+        let config: DeclarativeProviderConfig =
+            serde_json::from_str(json).expect("atomic_chat.json should parse");
+        assert_eq!(config.name, "atomic_chat");
+        assert_eq!(config.display_name, "Atomic Chat");
+        assert!(matches!(config.engine, ProviderEngine::OpenAI));
+        assert_eq!(config.api_key_env, "");
+        assert!(!config.requires_auth);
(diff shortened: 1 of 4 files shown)


alexhancock (approved):

alexhancock:
Verified this works by downloading Atomic Chat, running it with gemma4 and connecting from goose

@yanalialiuk I did notice after selecting the provider, it didn't set a default model. There may be config you can modify to make it smooth, but I will get the provider in for now

PR #9211, opened 2026-05-14
Build non-vulkan linux variants using ubuntu 22.04

### LINKED ISSUE #9203: v1.34.0 requires libc6 2.38

### LINKED ISSUE #9205: v1.34.0 published Linux binary requires libvulkan.so.1 at startup — not flagged in release notes
## Summary

Starting in v1.34.0, the prebuilt Linux tarball
(`goose-x86_64-unknown-linux-gnu.tar.bz2` / aarch64 equivalent, served via the
`stable` `download_cli.sh`) fails to start on systems without the Vulkan
loader installed:

    goose: error while loading shared libraries: libvulkan.so.1:
    cannot open shared object file: No such file or directory

This affects users who run inference exclusively through cloud providers
(e.g. AWS Bedrock) and never touch the local-inference code path — the
binary still links against `libvulkan.so.1` and the dynamic linker rejects
it before `main` runs.

## Why this is surprising

#9038 introduces `vulkan` as an **opt-in Cargo feature** for `goose`,
`goose-cli`, and `goose-server`. The PR description and v1.34.0 release
notes describe it that way ("Linux Vulkan support for local inference"
under ✨ Features). However the same PR also enables the feature in the
Linux CI/bundle/packaging paths, so the **published** Linux artifact has
it compiled in. From a downstream user's perspective the feature is not
opt-in: upgrading from v1.33.x to v1.34.0 silently adds a new runtime
dependency on the Vulkan loader.

The v1.34.0 release notes have no Breaking Changes section and don't
mention the new runtime dependency. `BUILDING_LINUX.md` lists `libvulkan1`
/ `libvulkan-dev` but that doc is framed as build-from-source guidance,
not runtime requirements for tarball users.

## Repro

Minimal Debian/Ubuntu image without the Vulkan loader (e.g.
\`node:22-bookworm\`, \`debian:bookworm-slim\`):

    curl -fsSL https://github.com/<org>/goose/releases/download/stable/download_cli.sh \
      | CONFIGURE=false bash
    /root/.local/bin/goose --version
    # goose: error while loading shared libraries: libvulkan.so.1: ...

Workaround: \`apt-get install libvulkan1\`.

## Suggestions (pick one)

- Document the new runtime dep in the v1.34.0 release notes and CHANGELOG
  as a breaking change for existing Linux users.
- Publish the Linux tarball **without** \`vulkan\` by default and offer a
  separate \`-vulkan\` artifact, mirroring the existing \`-cuda\` split for
  Windows.
- Statically link the Vulkan loader, or \`dlopen\` it lazily so the binary
  only fails when local inference is actually used.

Happy to PR any of these if there's a preferred direction.

jamadeo:
Fixes #9203
Fixes #9205 

Adds non-vulkan linux builds using ubuntu 22.04. This might not put us exactly where we want to be, but should get things working again for users who need this.

### DIFF
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -4,7 +4,7 @@
 #  - canary.yml
 #
 # Platform Build Strategy:
-#  - Linux: Uses native Ubuntu runners, including arm64 hosted runners
+#  - Linux: Uses native Ubuntu 22.04 runners to keep glibc compatibility with Ubuntu 22.04 LTS
 #  - macOS: Uses native macOS runners for each architecture
 #  - Windows: Uses Windows runner with native MSVC build
 on:
@@ -34,12 +34,12 @@ jobs:
           - platform: linux
             architecture: x86_64
             target-suffix: unknown-linux-gnu
-            build-on: ubuntu-24.04
+            build-on: ubuntu-22.04
             variant: standard
           - platform: linux
             architecture: aarch64
             target-suffix: unknown-linux-gnu
-            build-on: ubuntu-24.04-arm
+            build-on: ubuntu-22.04-arm
             variant: standard
           - platform: macos
             architecture: x86_64
@@ -93,7 +93,7 @@ jobs:
         if: matrix.platform != 'windows'
         uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
         with:
-          key: ${{ matrix.architecture }}-${{ matrix.target-suffix }}-native-macos-deployment-target-12
+          key: ${{ matrix.architecture }}-${{ matrix.target-suffix }}-${{ matrix.build-on }}-native-macos-deployment-target-12
 
       - name: Cache Cargo artifacts (Windows)
         if: matrix.platform == 'windows'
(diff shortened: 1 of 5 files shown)


jamadeo:
@omehegan could you test a non-vulkan build from here?

https://github.com/aaif-goose/goose/actions/runs/25873673204

jh-block (approved):

omehegan:
@jamadeo that works!

PR #9217, opened 2026-05-14
Dynamically refresh skill instructions each turn

### LINKED ISSUE #6382: Feature Request: Hot Reload Skills

According to Goose's documentation, "When a session starts, goose adds any skills that it discovers to its instructions." This suggests that skills are loaded at session start, and modifications to skills would require restarting the session to take effect.

Claude Code 2.1.0 (released January 7, 2026) introduced automatic skill hot-reload where skills created or modified in `~/.claude/skills` or `.claude/skills` are immediately available without restarting the session (see [[Issue #14314](https://github.com/anthropics/claude-code/issues/14314)](https://github.com/anthropics/claude-code/issues/14314)).

Goose implemented skills support in [[PR #5760](https://github.com/block/goose/pull/5760)](https://github.com/block/goose/pull/5760). If Goose requires session restarts for skill changes, this creates friction during skill development and iteration. When refining a skill's instructions, testing different approaches, or debugging skill behavior, users would need to:
- Exit the current session
- Restart Goose  
- Re-establish context and conversation state
- Test the modified skill

This workflow becomes particularly painful when:
- **Iterating on complex skills**: Multi-step workflows like deployment procedures require multiple refinement cycles
- **Debugging skill behavior**: Small tweaks to instructions need immediate testing
- **Developing new skills**: Initial skill creation involves substantial trial-and-error

Hot reloading would unlock:
- **Rapid skill iteration**: Test changes immediately without losing conversation context
- **Better debugging experience**: See how instruction changes affect behavior in real-time
- **Improved developer productivity**: Eliminate session restart overhead
- **Faster skill development**: Reduce the feedback loop from minutes to seconds
- **Feature parity**: Match the developer experience that Claude Code users already enjoy

Implement automatic hot-reload functionality for skills similar to Claude Code 2.1.0's approach. When skill files are created or modified in any of the skill directories (global or project-level), those changes should be immediately available in active Goose sessions without requiring a restart.

The reloaded skills should be immediately available to Goose, and users should receive some indication that a skill was reloaded (to avoid confusion about whether changes took effect).

1. **Manual reload command**: Add a `/reload-skills` CLI command
   - Pros: Simpler implementation, explicit user control
   - Cons: Still requires manual action, doesn't match Claude Code's seamless experience

2. **Automatic reload on next message**: Detect changes but only reload when user sends next message
   - Pros: Avoids interrupting mid-task, simpler than real-time watching
   - Cons: Not truly "hot" reload, still creates a delay in feedback loop

3. **File watcher with opt-in**: Require users to enable hot-reload explicitly
   - Pros: More conservative rollout, users choose when to use it
   - Cons: Extra configuration step, many users won't discover the feature

4. **Periodic polling**: Check for skill changes every 5-10 seconds
   - Pros: Simpler than file system watchers
   - Cons: Introduces delay, less efficient than event-based watching

**Additional context**

From Claude Code's CHANGELOG.md (version 2.1.0):
> Added automatic skill hot-reload - skills created or modified in `~/.claude/skills` or `.claude/skills` are now immediately available without restarting the session

This feature is particularly valuable for Goose users who:
- Maintain organization-specific skills in version control
- Develop custom workflows for their teams
- Use project-specific skills (`./.goose/skills/`) that change frequently
- Create skills as part of their development process

DOsinga:
## Summary

Skills listed in the system prompt were computed once at session start and cached in `server_info`. Adding or modifying `SKILL.md` files mid-session had no effect on the system prompt until restart.

## Changes

- **`McpClientTrait`**: Add `get_instructions()` with a default implementation that reads from `get_info()`. This is a non-breaking addition — all existing implementors get the default.
- **`SkillsClient`**: Override `get_instructions()` to re-discover skills from disk on every call. The constructor no longer bakes instructions into `InitializeResult`.
- **`Extension::get_instructions()`**: Delegate to `self.client.get_instructions()` instead of reading from the cached `server_info` snapshot.

The system prompt is rebuilt each turn, and `get_extensions_info()` calls `Extension::get_instructions()` each time — so newly added, modified, or removed skills are now reflected immediately without session restart.

This is safe for non-skill platform extensions and external MCP servers: they use the default `get_instructions()` which reads from their static `get_info()`, preserving existing behavior.

Closes #6382

### DIFF
--- a/crates/goose/src/skills/client.rs
+++ b/crates/goose/src/skills/client.rs
@@ -19,6 +19,26 @@ pub struct SkillsClient {
     working_dir: PathBuf,
 }
 
+fn build_skill_instructions(working_dir: &Path) -> String {
+    let sources = discover_skills(Some(working_dir));
+    let mut skills: Vec<&SourceEntry> = sources
+        .iter()
+        .filter(|s| s.source_type == SourceType::Skill || s.source_type == SourceType::BuiltinSkill)
+        .collect();
+    skills.sort_by(|a, b| (&a.name, &a.path).cmp(&(&b.name, &b.path)));
+
+    let mut instructions = String::new();
+    if !skills.is_empty() {
+        instructions.push_str(
+            "\n\nYou have these skills at your disposal, when it is clear they can help you solve a problem or you are asked to use them:",
+        );
+        for skill in &skills {
+            instructions.push_str(&format!("\n• {} - {}", skill.name, skill.description));
+        }
+    }
+    instructions
+}
+
 impl SkillsClient {
     pub fn new(context: PlatformExtensionContext) -> anyhow::Result<Self> {
         let working_dir = context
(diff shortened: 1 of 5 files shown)


crates/goose/src/skills/client.rs
```diff
@@ -252,6 +250,15 @@ impl McpClientTrait for SkillsClient {
         Some(&self.info)
     }
 
+    fn get_instructions(&self) -> Option<String> {
+        let instructions = build_skill_instructions(&self.working_dir);
```
alexhancock: nit: i would inline the impl of `build_skill_instructions` here as it's only called once

alexhancock (approved):

PR #9224, opened 2026-05-14
Fix OTLP trace flush on exit with configurable shutdown timeout

### LINKED ISSUE #7844: Goose: Flush Traces on `run` Exit
We are integrating goose cli into our workflow which involves invoking goose in a headless mode - with `goose run` commands. The problem is if the goose command is fase, there are not traces output (and/or you are at risk of "loosing" the last portion of the trace) because the goose command exits as soon as it is done without flushing the traces out. 

This means that we don't get traces, or we have to add artificial delays to allow for the background trace flush to complete.

I would like goose to automatically wait to flush traces, up to a configurable limit before exiting completely. The limit should ideally be configurable through an ENV or other arg. Default behavior can remain same as far as I am concerned. 

Based on the discord conversation with the AI bot the only other haacky solution is to sleep for 6 seconds to ensure it goes out. Another option I had thought about was having otel traces come to my calling app at which point I can detect I got traces and cancel, but that is overly complex. 

**Additional context**
[here is the conversation on discord - starts with confusion out langfuse conenctions, but then I ask about (https://discord.com/channels/1287729918100246654/1481717712681046199)](https://discord.com/channels/1287729918100246654/1481717712681046199/1481721658204094594)

DOsinga:
## Summary

Previously, `shutdown_otlp()` used the default 5s timeout via `provider.shutdown()` and silently discarded errors. Fast `goose run` invocations could lose traces because the arbitrary 100ms pre-sleep was insufficient and errors were invisible.

## Changes

- Read shutdown timeout from config key `otel_shutdown_timeout_ms` (env `OTEL_SHUTDOWN_TIMEOUT_MS`), defaulting to 5000ms
- Use `shutdown_with_timeout()` so users can tune the flush window
- Log warnings on shutdown errors instead of discarding them
- Remove the arbitrary 100ms pre-sleep (shutdown itself handles flushing)

## Configuration

Set via goose config file:
```yaml
otel_shutdown_timeout_ms: 10000
```

Or environment variable:
```
OTEL_SHUTDOWN_TIMEOUT_MS=10000
```

Default is 5000ms (matching the OpenTelemetry SDK default).

Fixes #7844

### DIFF
--- a/crates/goose-cli/src/main.rs
+++ b/crates/goose-cli/src/main.rs
@@ -10,7 +10,6 @@ async fn run() -> Result<()> {
 
     #[cfg(feature = "otel")]
     if goose::otel::otlp::is_otlp_initialized() {
-        tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
         goose::otel::otlp::shutdown_otlp();
     }
 
--- a/crates/goose-server/src/commands/agent.rs
+++ b/crates/goose-server/src/commands/agent.rs
@@ -137,7 +137,6 @@ pub async fn run() -> Result<()> {
 
     #[cfg(feature = "otel")]
     if goose::otel::otlp::is_otlp_initialized() {
-        tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
         goose::otel::otlp::shutdown_otlp();
     }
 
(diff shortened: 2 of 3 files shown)


DOsinga:
Superseded by #9228

PR #9229, opened 2026-05-14
feat(acp): pass session cwd param to acp providers

kalvinnchau:
## summary
implements acp cwd handling for session setup, per https://agentclientprotocol.com/protocol/session-setup

- validates session cwd inputs are absolute paths
- applies validation to session/new, session/load, session/fork, and _goose/update_working_dir
- uses the session/load request cwd for provider and extension setup
- passes session cwd through to acp subprocess providers
- keeps non-acp providers on the default cwd-agnostic provider path
- adds coverage to ensure loaded sessions pass the load cwd to the provider factory

noticed that even when passing the `cwd` it wouldn't be in the directory i sent and found that this was just using `std::env::current_dir().unwrap_or_else(|_| PathBuf::from("."))`

### notes

acp subprocess providers must override `from_env_with_working_dir` to preserve session cwd, i intentionally kept the provider trait shape cwd-agnostic for non-acp providers so all non-acp providers didn't have to deal with an unused `working_dir` param

acp subprocess providers now share a small current-working-directory fallback helper for non-session construction paths. the provider trait default remains cwd-agnostic for non-subprocess providers

### DIFF
--- a/crates/goose/src/acp/server.rs
+++ b/crates/goose/src/acp/server.rs
@@ -61,7 +61,7 @@ use rmcp::model::{
 use serde::Deserialize;
 use std::collections::{HashMap, HashSet};
 use std::panic::AssertUnwindSafe;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use strum::{EnumMessage, VariantNames};
 use tokio::sync::{Mutex, OnceCell};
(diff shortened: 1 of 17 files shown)


michaelneale (approved):
looks good, keeps the way it used to work otherwise, and seems to be the way the spec goes. nice.

PR #9232, opened 2026-05-14
Include request URL in provider error messages

### LINKED ISSUE #7453: When a 404 error happens, show the URL
It is so frustrating to get an error when configuring Goose that just says a 404 error happened.  Like this one:

Image

Every time you give a 404 error, show the URL that caused the 404 error.  This makes troubleshooting actually possible.

Madness

**Additional context**
No other context.

DOsinga:
When a provider request fails (404, auth errors, server errors, etc.), the error message now includes the URL that was requested. This makes it much easier to troubleshoot configuration issues — especially 404s where the user needs to know which endpoint was hit.

The change threads the request URL through `map_http_error_to_provider_error()` and `handle_status()` in the central HTTP error handling path, as well as all direct callers across provider implementations (Anthropic, Databricks, GCP Vertex AI, Google, Snowflake, Tetrate, Venice) and the Google-compat error handler in utils.rs.

Fixes #7453


jh-block (approved):
lgtm but you have conflicted with yourself

crates/goose/src/providers/tetrate.rs
```diff
@@ -76,15 +76,15 @@ impl TetrateProvider {
...
+    fn error_from_tetrate_error_payload(payload: Value, url: &str) -> ProviderError {
         let code = payload
             .get("error")
             .and_then(|e| e.get("code"))
             .and_then(|c| c.as_u64())
             .unwrap_or(500) as u16;
         let status = reqwest::StatusCode::from_u16(code)
             .unwrap_or(reqwest::StatusCode::INTERNAL_SERVER_ERROR);
-        Self::enrich_credits_error(map_http_error_to_provider_error(status, Some(payload)))
+        Self::enrich_credits_error(map_http_error_to_provider_error(status, Some(payload), url))
```
jh-block, line 87: Did it forget to sanitize here? Or is it that tetrate URLs don't have secrets in them.

I wonder whether having the URL be part of the error type (outside of the string) and having it be sanitised in the Display/Debug impl might be safer than needing to remember to sanitise it before constructing the error?

jh-block (approved):
lgtm with one question, and you conflicted with yourself

crates/goose/src/providers/tetrate.rs
DOsinga: yeah, that's probably better. going to merge it though to get ahead

PR #9238, opened 2026-05-15
feat: slash commands (built-in, skill, recipe) in acp server

lifeizhou-ap:
## Summary

Wires `builtin`, `skill`, and `recipe` slash commands into the ACP `available_commands` update so ACP clients get a unified list, and restructures the slash-command code into a proper module with first-class skill and recipe handlers (incl. argument parsing).

### ACP: surface all slash commands
- ACP `available_commands_update` now returns **builtins + recipes + skills** in one list, with name-collision precedence `builtin > recipe > skill`.
- UI: `MentionPopover` now inserts `/skill-name` for skill items, matching builtin/recipe behavior (previously inserted prose `"Use the X skill to "`).

### New `slash_commands/` module
Replaces the old single-file `crates/goose/src/slash_commands.rs` with a module that splits recipe, skill, into their own files and merges them into a unified list for ACP.

### Skill slash commands
- Skills now resolvable as `/skill-name [args]`.
- Skills can declare argument placeholders in their content and have them filled in at invocation time: a raw `$ARGUMENTS` blob, positional tokens (`$1`, `$ARGUMENTS[0]`), or named args (`$component`, `$from`) tied to the skill's declared argument names. Skills with no placeholders still receive the raw args appended so the model can see them.
- **Argument metadata location**: currently read from `metadata.argument-hint` and `metadata.arguments` (per the agentskills.io frontmatter spec). Open question whether to promote these to top-level frontmatter to match Claude Code — happy to discuss.
- `argument-hint` from frontmatter is surfaced as the ACP `input_hint`.
- Refactored: skill context rendering is now shared between the `load_skill` MCP tool and the slash-command path. The `load_skill` tool also accepts an optional `args` field.

### Recipe slash commands
- Multi-parameter support: required params are positional, optional params are `--flag value`. Example: `/migrate "Button Group" old-lib --to new-lib`.
- Single-required-param recipes still accept the full args string greedily (so `/weather Melbourne weather` works without quoting).
- Input hint rendered as `<required1> <required2> [--optional <optional>]` and shown in the ACP command list.
- Removed the old "only 1 param supported" error path and the now-unused `build_recipe_from_template_with_positional_params` helper.
- **Storage**: recipe→command mappings still live in the global config under `slash_commands`. Open question whether to move these to a per-project / discoverable location like skills — happy to discuss.

### Misc cleanup
- Moved `split_quoted` from `goose-cli` to shared `goose::utils::split_command_args`; reused by stdio extension parsing, recipe args, and skill args. Now also tolerates apostrophes inside unquoted words (`O'Reilly`, `don't`).
- `handle_skills_command` (`/skills` listing) now delegates to `skill_slash_command::format_installed_skills`.

### Testing
Unit Test
Manual testing in Desktop App
Created a simple client locally to test integrate with acp server

## Open questions
1. Should skill `argument-hint` / `arguments` live at the top level of frontmatter (Claude Code style) instead of under `metadata:`?
2. Should recipe→slash-command mappings move out of the global config (e.g. project-discoverable recipes)?
3. Do we need a custom api to get the available slash commands?  Currently we have notification in new_session, load_session, fork_session with the available slash command.  However, when user add a skills, recipe slash command, or change working directory, there is no other ways for client to get the latest slash commands.  We can add notification in those individual triggers, but it is hard to manage. Instead, having a custom api to return those commands seems be easier for the client.

## Next
- migrate ui/desktop to use these acp protocols
- agent slash command
- plugin skills slash command such as `plugin_name: skill_name"
- Handle SystemNotification Messages in acp as some of the slash command has/should generate these kind of messages


crates/goose/src/acp/server.rs
```diff
@@ -2895,10 +2836,7 @@ impl GooseAcpAgent {
                 SessionUpdate::UsageUpdate(usage_update),
             ))?;
         }
-
-        self.send_available_commands_update(cx, &args.session_id)
-            .await?;
-
+        Self::send_available_commands_update(cx, &args.session_id, &args.cwd)?;
```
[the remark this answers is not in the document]
lifeizhou-ap: I think `args.cwd` is intentional here. `on_load_session` updates the persisted session working_dir before publishing commands:

```
  self.session_manager
      .update(&session_id)
      .working_dir(args.cwd.clone())
      .apply()
      .await?;
```

Slash-skill execution later reloads the session from `session_manager`, so it should resolve skills from that same updated cwd.

crates/goose/src/skills/mod.rs
```diff
@@ -96,6 +99,69 @@ pub(crate) fn validate_skill_name(name: &str) -> Result<(), Error> {
...
+                let rel_str = relative.to_string_lossy().replace('\\', "/");
+                output.push_str(&format!(
+                    "- {} → load_skill(name: \"{}/{}\")\n",
+                    rel_str, skill.name, rel_str
+                ));
+            }
+        }
+    }
+
+    output.push_str("\n---\nThis knowledge is now available in your context.");
```
jamadeo: I know this is pre-existing but I don't really see why we need to say to the model "This is now in your context"

crates/goose/src/agents/agent.rs
```diff
@@ -1282,7 +1282,9 @@ impl Agent {
         if message_text.trim().starts_with('/') {
             let command = message_text.split_whitespace().next();
             if let Some(cmd) = command {
-                if crate::slash_commands::get_recipe_for_command(cmd).is_some() {
+                if crate::slash_commands::recipe_slash_command::get_recipe_for_command(cmd)
+                    .is_some()
+                {
```
jamadeo: also pre-existing, but feels like this tracking should be inside Agent.execute_command after we parse

jamadeo (approved):
Really nice. Only comments are things I noticed are pre-existing.

For recipe slash commands, I also wonder why we put them in the user config instead of the recipe itself as a `slash-command` field

crates/goose/src/agents/agent.rs
alexhancock: agree!

crates/goose/src/skills/client.rs
```diff
@@ -60,6 +87,7 @@ impl McpClientTrait for SkillsClient {
              load it first to get the detailed instructions.\n\n\
              Examples:\n\
              - load_skill(name: \"gdrive\") → Loads the gdrive skill instructions\n\
+             - load_skill(name: \"my-skill\", args: \"the arguments for the skill\") → Loads a skill with arguments\n\
```
alexhancock, line 68: Should we have some structure to the args instead of a string and the REGEX and replace logic in the frontend?

It seems to me like if we had that structure - we could build a nice argument entry experience on the frontend

alexhancock (approved):

lifeizhou-ap:
> Really nice. Only comments are things I noticed are pre-existing.
> 
> For recipe slash commands, I also wonder why we put them in the user config instead of the recipe itself as a `slash-command` field

Yes, will think about in the "@agent/recipe" work

crates/goose/src/skills/mod.rs
lifeizhou-ap: I was wondering about this too! I'll remove it and check there are any difference

crates/goose/src/skills/client.rs
lifeizhou-ap: Good point on the UX! The reason we matched this shape is portability. Tt mirrors Claude Code and the agentskills.io spec, so skills move between tools without rewriting.

One option: keep the string as the source of truth (required, works everywhere), and let skills optionally declare structured arg metadata that the frontend uses when present. Authors who only care about portability do nothing extra; goose-native skills can opt in for the richer entry experience.

crates/goose/src/slash_commands/recipe_slash_command.rs
```diff
@@ -0,0 +1,628 @@
...
+        }
+    }
+
+    let mut result = Vec::new();
+    if required.len() == 1 && !positionals.is_empty() {
+        result.push((required[0].key.clone(), positionals.join(" ")));
+    } else {
+        for (idx, value) in positionals.into_iter().enumerate() {
+            if idx >= required.len() {
+                return Err(anyhow!("Unexpected positional argument: {}", value));
```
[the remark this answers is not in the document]
lifeizhou-ap: logic/rule for recipe params in slash command: required params take positionals, optional params use --<key> flags. We deliberately removed the args special-case

PR #9240, opened 2026-05-15
Add Linux musl CLI builds

### LINKED ISSUE #9072: Installer downloads glibc Linux binary on Alpine/musl and fails with missing symbols
### Summary

The official CLI installer does not distinguish musl-based Linux distributions from glibc-based Linux distributions. On Alpine Linux, it downloads `goose-x86_64-unknown-linux-gnu.tar.bz2`, installs it, then immediately fails during `goose configure` with missing glibc symbols.

### Repro

```bash
curl -fsSL https://github.com/aaif-goose/goose/releases/download/stable/download_cli.sh | bash
```

### Environment

OS: Alpine Linux v3.21.3
Runtime: musl libc 1.2.5
Kernel: WSL2
Arch: x86_64
OSTYPE: linux-musl

Installer output:
```
  WINDIR: <not set>
  OSTYPE: linux-musl
  uname -s: Linux
  uname -m: x86_64
  Detected OS: linux with ARCH x86_64
  Downloading stable release: goose-x86_64-unknown-linux-gnu.tar.bz2...
  Moving goose to /home/wintercounter/.local/bin/goose

  Configuring goose

  Error relocating /home/wintercounter/.local/bin/goose: fcntl64: symbol not found
  Error relocating /home/wintercounter/.local/bin/goose: __res_init: symbol not found
```

### What I found

The downloaded binary is glibc-targeted:

ELF 64-bit LSB pie executable, x86-64, dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2

gcompat was installed, but it was not sufficient. Running ldd showed more missing glibc symbols such as:

__vsnprintf_chk
__snprintf_chk
__memcpy_chk
backtrace
__sched_cpualloc
fcntl64
__res_init
gnu_get_libc_version

The release assets for stable / v1.33.1 appear to include:

goose-x86_64-unknown-linux-gnu.tar.bz2
goose-aarch64-unknown-linux-gnu.tar.bz2

but no x86_64-unknown-linux-musl build.

### Solution

Publish musl-compatible CLI release artifacts, e.g. goose-x86_64-unknown-linux-musl.tar.bz2, and have the installer select them on musl systems.

### Workaround

I was able to build a working Alpine/musl binary from source for v1.33.1 with:

```
apk add build-base rust cargo openssl-dev pkgconf clang musl-dev bzip2-dev bzip2-static

cargo build --release -p goose-cli --bin goose \
  --no-default-features \
  --features rustls-tls,aws-providers,telemetry,otel
```

The resulting binary works on Alpine:

goose --version -> 1.33.1
file goose -> static-pie linked musl binary

### Extra Note

The full default feature build was not straightforward on Alpine/musl:

- local-inference pulled in llama-cpp / bindgen and failed around libclang dynamic loading.
- code-mode pulled in v8-goose, which attempted to download a prebuilt x86_64-unknown-linux-musl V8 archive that returned 404.

So a musl release may need adjusted feature selection or build handling.

jh-block:
Build reduced-feature musl CLI artifacts for Linux x86_64 and aarch64, and teach the installer and self-update command to select musl archives on musl-based systems.

Fixes #9072

Includes a commit to fix building without local-inference (needed for the musl build currently)

### DIFF
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -6,6 +6,7 @@
 # Platform Build Strategy:
 #  - Linux standard: Uses native Ubuntu 22.04 runners to keep glibc compatibility with Ubuntu 22.04 LTS
 #  - Linux Vulkan: Uses native Ubuntu 24.04 runners for newer Vulkan headers/tooling
+#  - Linux musl: Uses native Ubuntu 22.04 runners with reduced features for musl compatibility
 #  - macOS: Uses native macOS runners for each architecture
 #  - Windows: Uses Windows runner with native MSVC build
 on:
@@ -42,6 +43,16 @@ jobs:
             target-suffix: unknown-linux-gnu
             build-on: ubuntu-22.04-arm
             variant: standard
+          - platform: linux
+            architecture: x86_64
+            target-suffix: unknown-linux-musl
+            build-on: ubuntu-22.04
+            variant: musl
+          - platform: linux
+            architecture: aarch64
+            target-suffix: unknown-linux-musl
+            build-on: ubuntu-22.04-arm
+            variant: musl
           - platform: linux
             architecture: x86_64
             target-suffix: unknown-linux-gnu
@@ -104,6 +115,10 @@ jobs:
               glslc
           fi
 
+          if [ "${{ matrix.variant }}" = "musl" ]; then
+            sudo apt-get install -y musl-tools
+          fi
+
       - name: Cache Cargo artifacts (Linux/macOS)
         if: matrix.platform != 'windows'
         uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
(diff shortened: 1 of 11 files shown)


.github/workflows/build-cli.yml
```diff
@@ -133,7 +148,14 @@ jobs:
           if [ "${{ matrix.variant }}" = "vulkan" ]; then
             FEATURE_ARGS=(--features vulkan)
           fi
-          cargo build --release --target ${TARGET} -p goose-cli "${FEATURE_ARGS[@]}"
+
+          if [ "${{ matrix.variant }}" = "musl" ]; then
+            cargo build --release --target ${TARGET} -p goose-cli --bin goose \
+              --no-default-features \
+              --features rustls-tls,aws-providers,telemetry,otel
```
jamadeo: would this be easier to maintain if defined in Cargo.toml as a `default-musl` feature that enables these?

jamadeo (approved):

download_cli.sh
```diff
@@ -141,6 +141,28 @@ case "$ARCH" in
...
+detect_linux_musl() {
+  if [[ "$OSTYPE" == "linux-musl"* ]]; then
+    return 0
+  fi
+
+  if command -v ldd >/dev/null 2>&1 && ldd --version 2>&1 | grep -qi musl; then
+    return 0
+  fi
+
+  compgen -G "/lib/ld-musl-*.so.*" >/dev/null
```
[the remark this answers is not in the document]
jh-block: lol, I questioned codex about this when it was implementing it and it insisted it would be fine

.github/workflows/build-cli.yml
jh-block: Yeah that seems reasonable, especially as I'd like if we could eventually support more features on musl. I'll do it

PR #9242, opened 2026-05-15
Add unified thinking effort control across all providers

jh-block:
Replaces fragmented provider-specific thinking controls with a single unified Thinking Effort setting across supported providers.

Supersedes #7628 with the following changes:

- Preserved legacy model suffix migration while keeping predefined model limits intact.
- Applied unified effort to ChatGPT Codex's existing reasoning payload.
- Honored explicit off for Codex-backed providers by omitting reasoning configuration.
- Removed the duplicate ChatGPT Codex reasoning prompt from CLI configuration.

Also updated some providers to support reasoning where they did not (or did not fully) before:
- OpenRouter (if the canonical metadata says the model supports reasoning)
- Databricks (follows the serving-endpoint chain to find the underlying model and checks if it supports reasoning)


kyledef:
Hey @jh-block 👋 — I was independently investigating how thinking/reasoning parameters flow across provider types in Goose and was about to start work on exactly this gap. Really glad to see this PR land first!

kyledef:
Great work on this — the `ThinkingEffort` enum and unified `GOOSE_THINKING_EFFORT` knob are exactly the right abstraction layer.

One gap worth flagging: **local providers aren't wired up yet**.

- **Ollama** (`formats/ollama.rs`) supports `"think": true` in `/api/chat` for models like `qwq` and `deepseek-r1`, but the new `ThinkingEffort` enum isn't connected to it.
- **Local/llama.cpp** (`local_inference.rs`) already has its own `enable_thinking: bool` in `ModelSettings`, but it's a separate boolean that doesn't plug into `ThinkingEffort`. The mapping would be straightforward: `Off` → `false`, anything else → `true`.

Both of these are listed as first-class providers on the [goose docs providers page](https://goose-docs.ai/docs/getting-started/providers/), so community users running local models would benefit from the same unified knob.

The infrastructure is now in place — `model_config.thinking_effort()` is the right read path — so these could either be added here or as small follow-on PRs that build cleanly on this one. Happy to contribute the Ollama and local wiring if that's helpful.

jh-block:
@kyledef 

Yeah, many providers are not wired up yet. But the scope of this PR is already quite large so I think follow-up PRs for other providers is the way. Definitely would love to get a PR for Ollama / local inference.

crates/goose-server/src/routes/agent.rs
```diff
@@ -48,6 +48,7 @@ pub struct UpdateProviderRequest {
     model: Option<String>,
     session_id: String,
     context_limit: Option<usize>,
+    reasoning: Option<bool>,
```
DOsinga: can we drop the Option here? it's going to default to probably false, so might as well be explicit about it

crates/goose-server/src/routes/config_management.rs
```diff
@@ -392,14 +392,60 @@ pub async fn get_provider_models(
...
+    ),
+    request_body = ProviderModelInfoQuery,
+    responses(
+        (status = 200, description = "Model metadata fetched successfully", body = ModelInfo),
+        (status = 400, description = "Unknown provider, provider not configured, or authentication error"),
+        (status = 429, description = "Rate limit exceeded"),
+        (status = 500, description = "Internal server error")
+    )
+)]
+pub async fn get_provider_model_info(
```
DOsinga, line 458: I think we should put some effort in looking at whether we can avoid having this. we already have routes to get the model info from the providers. I realize this ends up doing a live check for databricks(?) but can we do that deeper in the stack and avoid the client side second useeffect?

crates/goose/src/acp/server.rs
```diff
@@ -3308,11 +3308,14 @@ impl GooseAcpAgent {
             current_model
         };
         let model = model_name.unwrap_or(&default_model);
```
DOsinga, line 3342: Only the explicit-override `ModelConfig` site is updated with `with_merged_request_params` while the two other construction sites (`resolve_provider_and_model` and the `switch_model` path) aren't — is that intentional because thinking effort is meant to flow through the global `GOOSE_THINKING_EFFORT` config, and if so does that mean per-session/request thinking effort overrides get dropped on model switch?

crates/goose/src/model.rs
```diff
@@ -1,12 +1,51 @@
...
 use serde_json::Value;
 use std::collections::HashMap;
+use std::fmt;
+use std::str::FromStr;
 use thiserror::Error;
 use utoipa::ToSchema;
 
 pub const DEFAULT_CONTEXT_LIMIT: usize = 128_000;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
```
DOsinga: consider exporting this through the API so we have this typed in the frontend

DOsinga (approved):

PR #9260, opened 2026-05-15
feat(tui): diff viewer

alexhancock:
## Summary
I often find myself wanting to see the diff and then iterate with goose, so this PR:

* Adds a diff viewer to the TUI UX
* Adds a tiny local slash commands system to the TUI so we can trigger it with `/diff`. Can see longer term about integrating with https://github.com/aaif-goose/goose/pull/9238 as that settles 

### Testing
Local usage (see demo)

### Related Issues
N/A

https://github.com/user-attachments/assets/8e97ec51-d1f5-41ec-84b8-fcbb91dd8374

### DIFF
--- a/ui/text/src/components/DiffViewer.tsx
+++ b/ui/text/src/components/DiffViewer.tsx
@@ -209,5 +209,3 @@ export function DiffViewer({
     </Box>
   );
 }
-
-export default DiffViewer;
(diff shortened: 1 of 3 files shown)


DOsinga (approved):
Looks good — nice feature! I pushed one fix: the `git diff` call was using `--color=always`, which wraps lines in ANSI escape codes and breaks the `classifyLine()` prefix detection in `DiffViewer`. Changed to `--no-color` since the component does its own coloring. Also removed the redundant default export.

hi from goose!

PR #9276, opened 2026-05-16
feat(providers): support OpenRouter request parameters

### LINKED ISSUE #8646: Document how to configure generic openrouter API parameters
I'm using the goose CLI and mostly with the openrouter.ai provider.

In the docs I don't seem to find how I can configure my openrouter model with the options they provide e.g. for
anthropic claude opus 4.7 there's [this documentation](https://openrouter.ai/docs/guides/evaluate-and-optimize/model-migrations/claude-4-7#new-xhigh-effort-level):

```json
{
"model": "anthropic/claude-4.7-opus",
"verbosity": "xhigh"
}
```

And in my `~/.config/goose/config.yaml` I have:

```
GOOSE_PROVIDER: openrouter
GOOSE_MODEL: anthropic/claude-4.7-opus
OPENROUTER_HOST: https://openrouter.ai
```

But, how would I configure e.g. `verbosity`. Or similarly [`reasoning-tokens`](https://openrouter.ai/docs/guides/best-practices/reasoning-tokens#controlling-reasoning-tokens) 

```json
{
  "model": "your-model",
  "messages": [],
  "reasoning": {
    // One of the following (not both):
    "effort": "high", // Can be "xhigh", "high", "medium", "low", "minimal" or "none" (OpenAI-style)
    "max_tokens": 2000, // Specific token limit (Anthropic-style)

    // Optional: Default is false. All models support this.
    "exclude": false, // Set to true to exclude reasoning tokens from response

    // Or enable reasoning with the default parameters:
    "enabled": true // Default: inferred from `effort` or `max_tokens`
  }
}
```

More Reading:

* https://openrouter.ai/docs/api/reference/overview
* https://openrouter.ai/docs/api/reference/parameters

How to play with [`plugins`](https://openrouter.ai/docs/api/reference/overview#plugins)?

```json
{
  "plugins": [
    { "id": "web" },
    { "id": "response-healing" }
  ]
}

```

sunilkumarvalmiki:
Resolves #8646.

## Summary
- Add optional `OPENROUTER_PARAMETERS` provider config.
- Accept both YAML object values and JSON string values for advanced OpenRouter request fields.
- Merge custom OpenRouter parameters without overwriting Goose-managed request fields.
- Document YAML and JSON string examples for `verbosity`, `reasoning`, and `plugins`.

## Testing
- `cargo fmt --all --check`
- `cargo check -p goose --no-default-features --features rustls-tls,local-inference --lib`
- `CARGO_BUILD_JOBS=1 cargo test -p goose --no-default-features --features rustls-tls,local-inference --lib openrouter_parameters`


DOsinga (approved):

DOsinga (changes requested):
Eh, actually can you rework this to not duplicate the existing `request_params` infrastructure? `ModelConfig.request_params` already gets merged into the OpenAI-compatible payload by `create_request()` in `formats/openai.rs` — so the simpler approach is to parse `OPENROUTER_PARAMETERS` in `from_env()` and feed it into `model_config.request_params` there. That way you don't need the `apply_openrouter_parameters` function, the extra field on the provider struct, or the second merge path with subtly different collision semantics (`or_insert` vs the existing `insert`-with-reserved-key-check). Most of the tests can go too since they're just testing serde_json basics.

sunilkumarvalmiki:
Reworked this to use the existing `ModelConfig.request_params` flow in `OpenRouterProvider::from_env()` instead of a provider-specific payload merge path. The provider field and `apply_openrouter_parameters` helper are removed, and the tests now cover parsing plus propagation into both the primary and fast model configs.

Local checks run:
- `cargo fmt --all --check`
- `CARGO_BUILD_JOBS=1 cargo test -p goose --no-default-features --features rustls-tls,local-inference --lib openrouter_parameters`
- `CARGO_BUILD_JOBS=1 cargo check -p goose --no-default-features --features rustls-tls,local-inference --lib`

sunilkumarvalmiki:
@DOsinga The requested rework is pushed and CI is green now. Could you take another look when you have a chance?

sunilkumarvalmiki:
Update: the branch is ready for another look.

What changed:
- Resolved the current `main` conflict in `documentation/docs/getting-started/providers.md`.
- Kept the OpenRouter parameter docs and the upstream Perplexity provider row.

Validation:
- `cargo fmt --all --check`
- `CARGO_BUILD_JOBS=1 cargo test -p goose --no-default-features --features rustls-tls,local-inference --lib openrouter_parameters -- --nocapture`
- `CARGO_BUILD_JOBS=1 cargo check -p goose --no-default-features --features rustls-tls,local-inference --lib`
- GitHub CI is green on `c34d53b87`.

Branch hygiene:
- One open PR from me in this repo.
- Changed files against current `main`: `crates/goose/src/providers/openrouter.rs`, `documentation/docs/getting-started/providers.md`.
- Conflicts resolved.

@DOsinga Could you please check and merge

DOsinga (approved):
Thanks for the rework — this is exactly what I had in mind, reusing the existing `request_params` infrastructure rather than duplicating it.

I pushed a small fix on top to address the codex P2: `apply_reasoning_config` was unconditionally overwriting the merged `reasoning` object, so a user-supplied `reasoning` in `OPENROUTER_PARAMETERS` would get clobbered when a thinking effort was also configured. It now respects an existing `reasoning` value and only synthesises one when none was provided, with a test covering that case.

LGTM, approving.

DOsinga:
Good catch — pushed a follow-up that removes the stale top-level `reasoning_effort` before returning when a user-supplied `reasoning` object is present, so we no longer send the shorthand and the full object together. Test updated to assert `reasoning_effort` is dropped.

DOsinga:
eh, over to you @sunilkumarvalmiki -- can you make codex happy?

DOsinga:
can you have another look @sunilkumarvalmiki ? also merge conflicts

PR #9278, opened 2026-05-16
fix(desktop): ScheduleModal error message styling

steveofficer:
## Summary
The ScheduleModal had the error messages styled with danger border, background and text. This meant that the messages were not able to be seen due to the red-on-red.

This has been addressed by removing the danger background style so that only the text is styled.

Additionally, the schedule name and source fields have been marked up in a similar way to the CreateRecipe modal. Required fields have an explicit asterix `*` to give a visual indicator that they are required.

### Testing
This has been manually tested.

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

**No visual indicator**
image

**Unreadable error text**
image

After:   

**Visual indicator added to labels**
image

**Error message styling fixed**
image


DOsinga (approved):
thanks for this!

PR #9280, opened 2026-05-16
fix(build): reduce release link-phase memory on macOS

### LINKED ISSUE #9258: just run-ui: release binary build spikes memory near the end
## What I'm seeing

Running `just run-ui` from a clean state locally on macOS, memory usage stays manageable through most of the Rust release compile, but spikes hard near the end of the build (around the linking / final optimization stages) and nearly caps out my MacBook Pro's memory.

Is anyone else hitting this? Wondering if it's expected at this point in the build, or if there's something we could tune (codegen-units, lto settings, fewer parallel jobs, etc.) to make a fresh `just run-ui` run friendlier on machines with less headroom.

## Environment

- macOS, Apple Silicon
- Fresh clone / clean target directory
- Command: `just run-ui` (which runs `just release-binary` then `pnpm run start-gui`)

colesmcintosh:
## Summary
- Adds `[profile.release]` with `strip = "debuginfo"` and `split-debuginfo = "unpacked"` so the linker (and dsymutil on macOS) has less to process at the end of a release build.
- Targets the late-stage memory spike reported on Apple Silicon during `just run-ui` / `just release-binary`, with no impact on runtime behavior.

Closes #9258

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -96,6 +96,10 @@ tree-sitter-rust = "0.24"
 tree-sitter-swift = "0.7"
 tree-sitter-typescript = "0.23"
 
+[profile.release]
+strip = "debuginfo"
+split-debuginfo = "unpacked"
+
 [patch.crates-io]
 v8 = { path = "vendor/v8" }
 cudaforge = { git = "https://github.com/jbg/cudaforge", rev = "e7c1967340e40673db98dc9e17da0f04834a456f" }


colesmcintosh:
Closing — profiled the change and it doesn't address the actual cause. The memory peak (~4.7 GB total, ~3.4 GB single rustc) occurs during LLVM codegen of large crates, not at link time. `strip` / `split-debuginfo` only affect post-link steps, and `dsymutil` doesn't even run under defaults because `debug = false` is the release default (no `.dSYM` produced in baseline). Will revisit with a targeted fix (likely a separate dev profile or `-j` knob for `just run-ui`).

PR #8920, opened 2026-04-29
feat(providers): add Perplexity as a supported model provider

jliounis:
## Summary

Adds **Perplexity** as a supported model provider in `crates/goose/src/providers/`, mirroring the layout of the existing xAI provider (`xai.rs`).

Perplexity's chat completions API is OpenAI-compatible (`https://api.perplexity.ai`), so the provider is a thin `ProviderDef` that returns an `OpenAiCompatibleProvider` configured with bearer-token auth. Its chat models ship with built-in real-time web search grounding, which is a natural fit for a developer agent that frequently needs to ground answers in current information.

## What's added

- **`crates/goose/src/providers/perplexity.rs`** (new) — `PerplexityProvider`:
  - Base URL: `https://api.perplexity.ai` (configurable via `PERPLEXITY_HOST`)
  - Auth: `PERPLEXITY_API_KEY` (canonical) → `PPLX_API_KEY` (alias used by Perplexity's SDKs) → error
  - Default model: `sonar-pro`. Known models: `sonar`, `sonar-pro`, `sonar-reasoning`, `sonar-reasoning-pro` (users can override `GOOSE_MODEL` to point at any other model the API accepts)
  - Setup steps wired up for the onboarding wizard.
- **`crates/goose/src/providers/mod.rs`** — register `pub mod perplexity;`
- **`crates/goose/src/providers/init.rs`** — register `PerplexityProvider` in the registry (default-off, matching xAI/Venice)
- **`crates/goose/tests/providers.rs`** — `test_perplexity_provider` smoke test entry following the same `ProviderTestConfig::with_llm_provider` shape as `test_xai_provider`
- **`documentation/docs/getting-started/providers.md`** — Perplexity row added to the provider matrix

## Tests

5 unit tests in `crates/goose/src/providers/perplexity.rs` (alongside provider source, matching the convention used by `venice.rs`):

1. `test_metadata_structure` — checks name, display_name, default_model, model_doc_link, and both `ConfigKey`s (`PERPLEXITY_API_KEY` required+secret+primary, `PERPLEXITY_HOST` with default).
2. `test_known_models_non_empty` — default model is included in the known list.
3. `test_setup_steps_present` — onboarding steps reference `PERPLEXITY_API_KEY`.
4. `test_default_model_is_known`.
5. `test_doc_url_points_to_perplexity_docs`.

[...]

PR #9284, opened 2026-05-17
feat(providers): add Perplexity as a declarative provider

jliounis:
## Summary

- Add Perplexity as a declarative OpenAI-compatible provider.
- Include the Sonar models from the maintainer-provided template with `PERPLEXITY_API_KEY`, streaming support, model docs, and setup steps.
- Add Perplexity to the supported providers documentation.

## Context

This follows up on #8920, which was closed because the requested direction was to use goose's JSON-based declarative provider system instead of adding a Rust provider module. The provider implementation delta here is JSON-only, so there is no Rust module or provider-specific Rust test surface to maintain.

## Validation

- `jq empty crates/goose/src/providers/declarative/perplexity.json`
- `git diff --check`
- Attempted `cargo test -p goose declarative_providers`; this Buildkite image could not complete it because Hermit expects `/bin/ln`, the AGI Cargo wrapper points at an internal registry, and an isolated crates.io Cargo run later failed on missing `c++` for `llama-cpp-sys-2`.

Signed-off-by: James Liounis <james.liounis@perplexity.ai>

---
🤖 _Generated by [PSI](https://www.notion.so/perplexity-ai/HowTo-PSI-Perplexity-Super-Intelligence-29c6172b229b80d3a8ece89a6cfb6ae6)_
🧠 _Agent used: `codex`_


jamadeo:
dupe of #9324 ? thanks for the contribution!

PR #9293, opened 2026-05-18
docs: custom agents

angiejones:
documenting new Custom Agents feature. might want to wait to merge with v2 desktop.


The-Best-Codes (approved):

michaelneale (approved):
can merge when it is ready and makes sense

PR #9298, opened 2026-05-18
fix(desktop): handle resume deep links

### LINKED ISSUE #9187: Feature Request: Session Resume Deep Link (goose://resume/<session-id>) for Workspace Save/Restore
### Problem

When you close Goose (e.g., to restart for permission changes, updates, or end of day), there's no way to programmatically reopen the specific set of sessions you had active. You're left scrolling through the sidebar history trying to remember which conversations were in progress.

This is especially painful for power users who keep multiple Goose windows open simultaneously — each focused on a different workstream (e.g., one for a data analysis, one for a code review, one for a product spec).

### Proposed Solution

Add a `goose://resume/<session-id>` deep link that:

1. Opens a new Goose window (or reuses an existing one)
2. Navigates directly to the specified session, resuming it as if the user had clicked it in the sidebar

This is the missing piece that would enable **workspace save/restore** — saving a named set of session IDs and reopening them all at once after a restart.

### Current State

- Sessions are already persisted in SQLite (`~/.local/share/goose/sessions/sessions.db`) with stable IDs
- The desktop app already supports `goose://` protocol handling for `bot`, `recipe`, `extension`, and `sessions` (shared) hostnames
- The React router already supports `?resumeSessionId=<id>` for internal navigation
- The `second-instance` handler already processes `goose://` URLs from the command line

So the implementation should be straightforward — add a `resume` hostname handler that creates a new `BrowserWindow` navigated to `/pair?resumeSessionId=<id>`.

### Use Cases

1. **Daily workspace restore** — Close Goose at EOD, reopen all your active conversations the next morning
2. **Restart recovery** — Grant a new permission (e.g., Full Disk Access), restart Goose, restore your windows
3. **Context switching** — Save workspaces by project ("upmarket-card", "oncall", "data-analysis") and switch between them
4. **Scripting/automation** — Shell scripts or tools that open specific sessions programmatically

### Workaround

I built a CLI wrapper (`goose-workspace`) that saves/restores session ID sets and opens N Goose windows, but the restore step requires the user to manually click each session in the sidebar since there's no deep link to target a specific session.

### Additional Context

The implementation in `main.ts` would look roughly like:

```typescript
} else if (parsedUrl.hostname === 'resume') {
  const sessionId = parsedUrl.pathname.replace(/^\//, '');
  if (sessionId) {
    await createChat(app, {
      dir: openDir || undefined,
      resumeSessionId: sessionId,
    });
  }
}
```

With a corresponding update to `createChat` to accept and pass through `resumeSessionId` as a URL parameter.

McGluut:
Fixes #9187.

## Summary
- add `goose://resume/<session-id>` handling alongside the existing `new-session` protocol flow
- pass the extracted session id through `createChat({ resumeSessionId })` so the existing `/pair?resumeSessionId=...` path can resume the session
- cover the Windows/Linux second-instance flow, startup protocol handling, and macOS `open-url` flow

## Validation
- `pnpm run typecheck`
- `pnpm exec prettier --check src/main.ts`
- `pnpm exec eslint src/main.ts --max-warnings 0 --no-warn-ignored`
- `git diff --check`


DOsinga (approved):
Clean, well-scoped feature that adds `goose://resume/<session-id>` deep link support, matching the existing patterns for `new-session` and `bot`/`recipe` handlers. Codex review comments were all addressed in follow-up commits.

I pushed one small fix: made `parsedUrl` a required parameter in `handleProtocolUrl` since both call sites already have it available — no reason for it to be optional. Also merged origin/main (no conflicts).

McGluut:
Thanks for the review and for pushing the cleanup. Making parsedUrl required is a good simplification since both callers already have it; I'll carry that forward as the right pattern for protocol handlers here. Appreciate you merging main as well.

Abhijay007:
@McGluut can you please look into merge conflicts

McGluut:
Updated the branch with current main to clear the merge conflicts.

PR #9303, opened 2026-05-18
fix(litellm): use context limit from /model/info for custom models

### LINKED ISSUE #8835: Context limit from Provider API is discarded, causing fallback to 128k for local models
**Describe the bug**

In `crates/goose/src/providers/litellm.rs` (and potentially `openai.rs`), Goose fetches supported models from the proxy's API and successfully parses the `max_input_tokens` / `context_length`. However, it explicitly discards this context length when returning the list of supported models from the provider. 

This causes Goose's `ModelConfig` to fall back to checking its internal static `canonical_models.json`. When it fails to find a match for a custom local model (e.g., `local/unsloth/Qwen3-8B-Q4_K_M.gguf`), it blindly defaults to its hardcoded `DEFAULT_CONTEXT_LIMIT` of 128,000. 

As a result:
1. Goose can feed too many tokens to local models with smaller context windows (e.g., 8k tokens for Qwen), causing them to crash or throw errors during long sessions unless the user manually overrides `GOOSE_CONTEXT_LIMIT`.
2. Features that depend on the percentage of context used, such as automatic context compaction, will not work correctly because the perceived context limit (128k) is drastically different from the actual limit.

---

**To Reproduce**
1. Start a local LiteLLM proxy (or any OpenAI-compatible server) that returns custom local models where `max_input_tokens` is smaller than 128k.
2. Configure Goose to use the `litellm` provider.
3. Start Goose without setting `GOOSE_CONTEXT_LIMIT`.
4. Check the parsed context limit for the model (or try a long session).
5. See error: Goose assumes a 128k context window and either overflows the model or fails to trigger context compaction at the correct threshold.

---

**Expected behavior**
Goose should retain and respect the `context_length` (or `max_input_tokens`) provided by the `/model/info` (or `/v1/models`) endpoint of the provider, rather than discarding it and defaulting to 128,000.

---

**Screenshots**
N/A

---

- **OS & Arch:** Ubuntu 24.04.4 x86_64
- **Interface:** CLI
- **Version:** Latest (main branch)
- **Extensions enabled:** Developer, Tom, Skills, Todo, Summon, Apps, Extension Manager
- **Provider & Model:** LiteLLM / local .gguf models

---

**Additional context**
The issue is clearly visible in the `fetch_supported_models` implementation. 

In `crates/goose/src/providers/litellm.rs` (around line 111):
```rust
let context_length = model_info["max_input_tokens"].as_u64().unwrap_or(128000) as usize;
let mut model_info_obj = ModelInfo::new(model_name, context_length);
model_info_obj.supports_cache_control = supports_cache_control;
models.push(model_info_obj);

// ... Later in fetch_supported_models:
let models = self.fetch_models().await.map_err(|e| { /* ... */ })?;

// BUG: Goose correctly fetched and parsed it, but throws it away here!
Ok(models.into_iter().map(|m| m.name).collect())
```
Because the `Provider::fetch_supported_models` trait signature expects `Result<Vec<String>, ProviderError>`, the rich metadata (including the context limit) is lost in transit.

DOsinga:
When a model isn't in the canonical database (e.g. local/custom models behind a LiteLLM proxy), `context_limit` was left as `None` and silently defaulted to 128k. This caused context overflow on smaller models and broke context compaction.

Cache the `/model/info` response in a `OnceCell`, populated lazily on first use (`supports_cache_control` during `stream()`, or `fetch_supported_models`). `get_model_config()` reads the cache synchronously — no blocking at init, no extra HTTP calls, and the context limit is available before any compaction decision since the first `stream()` call populates the cache.

Also fixes the pre-existing issue where `supports_cache_control` and `fetch_supported_models` each made separate HTTP calls to `/model/info` on every invocation.

Fixes #8835

### DIFF
--- a/crates/goose/src/providers/litellm.rs
+++ b/crates/goose/src/providers/litellm.rs
@@ -30,6 +30,8 @@ pub struct LiteLLMProvider {
     model: ModelConfig,
     #[serde(skip)]
     name: String,
+    #[serde(skip)]
+    cached_model_info: tokio::sync::OnceCell<Vec<ModelInfo>>,
 }
 
 impl LiteLLMProvider {
@@ -77,10 +79,18 @@ impl LiteLLMProvider {
             base_path,
             model,
             name: LITELLM_PROVIDER_NAME.to_string(),
+            cached_model_info: tokio::sync::OnceCell::new(),
         })
     }
 
-    async fn fetch_models(&self) -> Result<Vec<ModelInfo>, ProviderError> {
+    async fn get_or_fetch_models(&self) -> Result<&[ModelInfo], ProviderError> {
+        self.cached_model_info
+            .get_or_try_init(|| self.fetch_models_from_api())
+            .await
+            .map(|v| v.as_slice())
+    }
+
+    async fn fetch_models_from_api(&self) -> Result<Vec<ModelInfo>, ProviderError> {
         let response = self
             .api_client
             .request(None, "model/info")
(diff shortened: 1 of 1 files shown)


DOsinga:
Addressing Codex review feedback:

**P1 (blocking init):** Already addressed in the previous push — replaced the eager `fetch_models().await` with a lazy `OnceCell` that populates on first use during `stream()`. Zero blocking at init.

**P1 (get_model_config not triggering fetch):** By design — `get_model_config()` is a sync trait method so it can't initiate an async fetch. The cache is populated by `supports_cache_control()` during the first `stream()` call on turn 1. On that first turn the conversation is too small to trigger compaction, so the 128k default is harmless. From turn 2 onward the real limit is available. Added a code comment explaining this.

**P2 (zero context limits):** Fixed — now skips non-positive context limits from upstream metadata.

alexhancock (approved):

PR #9304, opened 2026-05-18
feat(hooks): PreToolUse denial

### LINKED ISSUE #9068: HITL and lifecycle hooks
There is nothing about HITL. lifecycle hooks and approval mechanishm on goose

Different hooks connected to different events

### LINKED ISSUE #9277: extensions: would a detection-rule hook at tool-call boundary be in scope?
Hi maintainers,

Opening an issue first to validate appetite before any code work.

## Context I checked

Goose already ships a plugins + hooks framework (`crates/goose/src/hooks`, `crates/goose/src/plugins`) following the Open Plugins spec, with `PreToolUse`, `PostToolUse`, `BeforeShellExecution`, etc. So the surface I had in mind exists. The question is whether a contribution that uses it for tool-call-boundary detection would be in scope.

## Question

Would maintainers accept a reference plugin that subscribes to `PreToolUse` and `PostToolUse`, runs a detection-rule scan over the tool input and output, and short-circuits (non-zero exit per the hooks spec) when a configured severity threshold is hit?

The intended contribution:

- One self-contained plugin under the existing discovery path, shipping a `hooks/hooks.json` with `PreToolUse` and `PostToolUse` entries
- A `type: "command"` script that wraps ATR (Agent Threat Rules, MIT-licensed open ruleset, 425 rules across 10 categories including prompt injection, tool poisoning, credential exfiltration)
- Pluggable detector — the script interface accepts any callable that returns a verdict, so users can swap ATR for another ruleset
- Configuration through standard plugin config; default severity threshold is conservative
- Zero changes to goose core, purely additive

## Why this might be useful

Tool-call boundary scanning is a recurring ask for production agent deployments. The same pattern is running in production at Microsoft Agent Governance Toolkit (PRs #908 and #1277, merged 2026-04) and Cisco AI Defense skill-scanner (PRs #79 and #99, merged 2026-04).

## What I would draft if you say yes

- The plugin directory matching goose's discovery conventions
- A small integration test exercising one PreToolUse rule and one PostToolUse rule
- Docs page under `documentation/docs/` covering install and the severity-threshold config
- Reference implementation using ATR; the wrapper is small enough that swapping detectors is trivial

Rule pack repo: https://github.com/Agent-Threat-Rule/agent-threat-rules

## If you say no

No PR will follow.

Thanks for considering.

alexhancock:
## Summary
* Make it so hooks can deny tool calls.
* Runs hooks in order and stops at the first explicit deny (exit code 2 or {"decision":"block",...} on stdout). Any other failure — spawn errors, timeouts, other non-zero exits — is logged and treated as Allow, so a misbehaving hook can't block.

### Testing
Local CLI and Desktop test

### Related Issues
https://github.com/aaif-goose/goose/issues/8742
Fixes #9068

Screenshot 2026-05-18 at 10 45 46 AM

Fixes #9277

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -916,8 +916,11 @@ impl Agent {
                 return (
                     request_id,
                     Err(ErrorData::new(
-                        ErrorCode::INVALID_REQUEST,
-                        format!("Tool call blocked by plugin `{plugin}`: {reason}"),
+                        ErrorCode::INTERNAL_ERROR,
+                        format!(
+                            "Tool call denied by policy hook `{plugin}`: {reason}. \
+                             Do not retry; this is a policy denial, not a transient failure."
+                        ),
                         None,
                     )),
                 );
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -908,9 +908,20 @@ impl Agent {
                             .map(|a| serde_json::Value::Object(a.clone())),
                     )
                     .with_working_dir(session.working_dir.to_string_lossy().to_string());
-            self.hook_manager
-                .emit(crate::hooks::HookEvent::PreToolUse, ctx)
-                .await;
+            if let crate::hooks::HookDecision::Deny { reason, plugin } = self
+                .hook_manager
+                .emit_blocking(crate::hooks::HookEvent::PreToolUse, ctx)
+                .await
+            {
+                return (
+                    request_id,
+                    Err(ErrorData::new(
+                        ErrorCode::INVALID_REQUEST,
+                        format!("Tool call blocked by plugin `{plugin}`: {reason}"),
+                        None,
+                    )),
+                );
+            }
         }
 
         let tool_input_for_extended = tool_call
--- a/crates/goose/src/hooks/mod.rs
+++ b/crates/goose/src/hooks/mod.rs
@@ -208,6 +208,13 @@ impl HookContext {
     }
 }
 
+/// Outcome of a blocking hook dispatch.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum HookDecision {
+    Allow,
+    Deny { reason: String, plugin: String },
+}
+
 /// Loads and executes plugin hooks.
 #[derive(Debug, Default, Clone)]
 pub struct HookManager {
@@ -297,9 +304,20 @@ impl HookManager {
                     command = %command,
                     "Running plugin hook",
                 );
-                if let Err(err) =
-                    run_command_hook(command, &rule.plugin_root, &payload, *timeout).await
-                {
+                let res = run_command_hook(command, &rule.plugin_root, &payload, *timeout)
+                    .await
+                    .and_then(|o| {
+                        if o.status.success() {
+                            Ok(())
+                        } else {
+                            anyhow::bail!(
+                                "hook `{command}` exited with {:?}: {}",
+                                o.status.code(),
+                                String::from_utf8_lossy(&o.stderr).trim()
+                            )
+                        }
+                    });
+                if let Err(err) = res {
                     warn!(
                         plugin = %rule.plugin_name,
                         event = %event,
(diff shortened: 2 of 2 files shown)


DOsinga:
A couple of notes on the cleanup commit (8c81a5b) so future-me remembers why, and one bigger question I'd like us to make a call on before this lands.

## Why the cleanup

Three things, all small:

1. **Dropped the `/// Outcome of a blocking hook dispatch.` doc comment on `HookDecision`.** The type is named `HookDecision` with variants `Allow` and `Deny { reason, plugin }`. The comment restated the signature — no information added. Per `AGENTS.md`: "Never add comments that restate what code does."

2. **Dropped the `if rules.is_empty() { return Allow; }` short-circuit in `emit_blocking`.** It's dead code — `HookManager::from_plugins` never inserts an empty `Vec` into `rules`, and the `for rule in rules` loop below falls through to `Allow` correctly anyway. (Same redundant check exists in `emit`; left it alone there to keep this PR's scope tight, but we should clean that up too eventually.)

3. **Removed `emit_blocking_denies_on_exit_code_2_and_fails_open_otherwise`.** This is the one worth explaining. It was an implementation-coupled change-detector test: it asserts the function returns the enum variant the function was written to return, not that the user-facing behavior works. Concretely — if someone removed the early-return in `dispatch_tool_call` so the tool ran *anyway* after a deny, this test still passes. That's the bug we'd actually care about catching, and it doesn't catch it.

   The two scenarios crammed into one tuple-loop also obscured what's being tested and produced worse failure messages than two focused tests would.

   What we *should* have here is an end-to-end test that installs a denying plugin, dispatches a tool through the agent, and asserts (a) the tool didn't run, and (b) the model got a sensible error. Plus a few unit tests on `deny_reason` directly covering the JSON-on-stdout path and the `DEFAULT` fallback (both currently uncovered). I'd rather land those as a follow-up than leave the existing test in as a green-CI placeholder.

## Bigger question: align `HookContext` field names with Claude Code?

While reviewing this PR I noticed our `HookContext` JSON payload diverges from Claude Code's hook payload on field names with no obvious justification. The Open Plugins spec doesn't pin payload field names down, so we're not violating any spec — but in practice Claude Code is the de facto reference implementation and Open Plugins examples follow its conventions.

Diff summary:

| Concept | Claude Code | Goose today |
|---|---|---|
| Event name | `hook_event_name` | `event` |
| Working directory | `cwd` | `working_dir` |
| User prompt text | `prompt` | `message` |
| Tool result | `tool_response` | `tool_output` (declared, not actually populated — separate bug) |
| Session ID | `session_id` | `session_id` ✓ |
| Tool name | `tool_name` | `tool_name` ✓ |
| Tool input | `tool_input` | `tool_input` ✓ |

The naming divergences look incidental, not principled — there's no design doc or commit message I could find that explains them. Concrete cost: a plugin author can't write one hook script that works against both agents, even for something trivial like "log the tool name", because the event-name field is `event` here and `hook_event_name` there.

Worth noting: PR #9304's denial protocol (`exit 2` / `{"decision":"block","reason":"..."}` on stdout) **does** follow Claude Code's convention. So we've already established a precedent of borrowing their hook conventions when useful. The inconsistency is that we copied their *output* protocol but invented our own *input* schema.

Example, concretely — our `PreToolUse` payload today:

```json
{
  "event": "PreToolUse",
  "session_id": "abc123",
  "matcher_context": "developer__shell",
  "tool_name": "developer__shell",
  "tool_input": { "command": "ls" },
  "working_dir": "/Users/douwe/proj/goose"
}
```

Claude-Code-aligned version (same data, names a plugin author already knows):

```json
{
  "hook_event_name": "PreToolUse",
  "session_id": "abc123",
  "matcher_context": "developer__shell",
  "tool_name": "developer__shell",
  "tool_input": { "command": "ls" },
  "cwd": "/Users/douwe/proj/goose"
}
```

`matcher_context` stays as a goose-only convenience addition.

**The call I think we need to make now, before this PR makes plugin authors start writing real scripts against the deny protocol:** do we want to converge with Claude Code's field names, or deliberately diverge? I lean strongly toward converge — there's no flexibility we'd be giving up, and the ecosystem benefit (existing Claude Code hook scripts work as-is, examples from the community port over) is real. The blast radius right now is small because the API isn't documented externally yet; in three months it won't be.

Either way, the answer should be deliberate. If we converge, the rename is a small follow-up PR. If we diverge, we should document *why* so the next reviewer doesn't ask the same question.

Happy to take this one if there's agreement on the direction.

crates/goose/src/agents/agent.rs
```diff
@@ -908,9 +908,20 @@ impl Agent {
...
+            if let crate::hooks::HookDecision::Deny { reason, plugin } = self
+                .hook_manager
+                .emit_blocking(crate::hooks::HookEvent::PreToolUse, ctx)
+                .await
+            {
+                return (
+                    request_id,
+                    Err(ErrorData::new(
+                        ErrorCode::INVALID_REQUEST,
+                        format!("Tool call blocked by plugin `{plugin}`: {reason}"),
```
DOsinga: One more thing worth raising before merge — separate from the cleanup commit, this is about behavior.

**The `INVALID_REQUEST` error code here may cause the model to retry the denied tool call.**

Semantically, `INVALID_REQUEST` means "your request was malformed" — the natural response from a well-behaved model is to *correct* and retry. That's the opposite of what we want for a policy denial. A model that just got told "you can't run `rm -rf /`" should accept the denial, not infer "I phrased the request wrong, let me try again."

It's worse than just "wrong error code semantics." The message we hand back includes the literal string `"Tool call blocked by plugin"`, which a sufficiently clever model might read as an invitation to *work around* the plugin — rename the tool call, split the command into pieces, try a different tool that does the same thing. We'd be teaching the model that policy denials are obstacles to route around, not boundaries to respect.

Two ways out:

1. **Return a successful `ToolCallResult` with `is_error: true`** instead of an MCP `Err`. The denial then surfaces to the model as a tool *result* ("I called the tool, it told me no"), which is its actual mental model — and tool-result errors don't trigger the retry path that MCP request errors do. Message could be along the lines of *"This tool call was blocked by plugin `<name>`: <reason>. Do not retry; this is a policy denial, not a transient failure."*

2. **At minimum, pick a less retry-friendly MCP error code.** I don't think MCP has a perfect one, but `INTERNAL_ERROR` is less misleading than `INVALID_REQUEST` — it doesn't suggest "fix your request and try again."

I'd lean strongly toward (1) — it's the more honest representation of what happened and removes the retry ambiguity entirely. Either way, I'd want a test that confirms a denied call doesn't trigger a retry storm, since that's the actual user-facing failure mode we're trying to prevent.

(Related, and probably a follow-up PR rather than this one: the deny protocol today only applies to `PreToolUse`. The more useful policy primitives — `BeforeShellExecution` with command-string matchers, `BeforeReadFile` with path matchers — are still fire-and-forget. A plugin author can scope a deny rule to `developer__shell` but not to `rm -rf *`. Worth noting in the PR description so we don't accidentally ship the impression that command-content denial works.)

DOsinga (approved):

Issue #9153, opened 2026-05-12
Severe memory leak in Docker container when MCP tool returns very large responses (~3 million characters)


Surendhar-N-D:
#### Steps to Reproduce

1. Run `goose-server` (`goosed`) inside a **Docker container**.
2. Connect to it from **Goose Desktop** as an external Goose server.
3. Use an MCP tool that returns a very large response — approximately **3,000,000 characters** (~3MB of text).
4. Call the tool once or multiple times.
5. Monitor memory usage inside the container using `docker stats`, `htop`, or `top`.

#### Expected Behavior
Memory should increase temporarily while processing the large tool response, then **drop back** to normal baseline levels after the response is handled and the chat turn is completed (this works correctly on native macOS and Windows).

#### Actual Behavior
- When the MCP tool returns ~**3 million characters**, RAM spikes sharply.
- Memory **does not decrease** after the tool call finishes.
- Repeated calls cause continuous memory growth, easily exceeding **10GB**.
- Memory only goes down after restarting the entire Docker container.

This memory leak / spike issue occurs **only inside Docker**. On native macOS and Windows, memory rises during the large tool call but eventually returns to normal.

#### Environment
- **Deployment**: `goosed` running in Docker container
- **Client**: Goose Desktop (external server mode)
- **Host OS**: Mac

<img width="762" height="299" alt="Image" src="https://github.com/user-attachments/assets/ccb16a37-8552-48be-813d-4a0933b4eb2e" />

DOsinga:
Thanks for the detailed report. We do have a large-response handler that spills tool output >200K chars to a temp file, but in Docker the default `/tmp` is often backed by tmpfs (RAM), which would explain why memory grows and never reclaims — the 'spilled' data is still in memory.

As a **workaround**, try mounting a real volume at the spill directory:

```
docker run -v /host/path:/tmp/goose_mcp_responses ...
```

We're considering a fix that either uses a non-tmpfs spill path inside containers or cleans up temp files after the agent is done with them. If anyone else is hitting this, please react or comment so we can prioritize.

Surendhar-N-D:
Thanks @DOsinga, I tried the workaround but the memory still isn't reclaimed, and I think the evidence points away from the `/tmp` tmpfs being the root cause. 

<img width="487" height="164" alt="Image" src="https://github.com/user-attachments/assets/9a97367e-06ba-4ec3-9b37-0fc6031882b9" />
[...]

PR #9308, opened 2026-05-18
fix: keep large MCP response files off tmpfs

bzqzheng:
## Summary
When an MCP tool response exceeds the large-response threshold, goose writes the full text to a spill file and returns the file path to the model. That path previously lived under the process temp directory (`/tmp/goose_mcp_responses` on Unix-like systems). In Docker, `/tmp` is often tmpfs-backed, so large responses can still consume container memory even after goose has moved them out of the model context.

This keeps the existing large-response behavior, but moves the spill files to goose's data directory and keeps the path private:
- write large MCP response files under `Paths::in_data_dir("mcp_responses")` instead of the OS temp directory
- ensure the spill directory is a real directory with `0700` permissions on Unix
- reject a symlink or non-directory at that path before writing
- create persisted response files via `tempfile`, which gives the file a randomized name and `0600` permissions
- avoid counting the same large text twice while deciding whether to spill it

### Testing
- `cargo fmt --package goose`
- `cargo test -p goose agents::large_response_handler::tests -- --nocapture`
- `cargo clippy -p goose --all-targets -- -D warnings`
- `git diff --check`

### Related Issues
Related to #9153. This addresses the tmpfs-backed `/tmp` spill path described there by moving large MCP response files into goose's data directory. It does not add lifecycle cleanup for old spill files.

Before: N/A

After: N/A

### DIFF
--- a/crates/goose/src/agents/large_response_handler.rs
+++ b/crates/goose/src/agents/large_response_handler.rs
@@ -1,7 +1,11 @@
+use crate::config::paths::Paths;
 use chrono::Utc;
 use rmcp::model::{CallToolResult, Content, ErrorData};
-use std::fs::File;
-use std::io::Write;
+use std::io::{self, Write};
+use std::path::{Path, PathBuf};
+
+#[cfg(unix)]
+use std::os::unix::fs::{DirBuilderExt, PermissionsExt};
 
 const LARGE_TEXT_THRESHOLD: usize = 200_000;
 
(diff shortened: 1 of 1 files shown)


DOsinga:
Thanks for the PR — the Docker tmpfs issue in #9153 is real, but we're not sure moving spill files to `data_dir` is the right fix.

The current code writes to `/tmp`, which on real Linux systems gets cleaned up periodically by `systemd-tmpfiles` / `tmpwatch` (typically files older than 10 days), and on macOS gets cleaned on reboot. These spill files are temporary by nature — they're written once, read by the agent, and never needed again.

Moving them to `~/.local/share/goose/mcp_responses/` means nothing ever cleans them up. With tool calls that can generate 3MB+ per spill file, this will grow without bound in a user data directory. That trades the Docker RAM issue for unbounded disk growth everywhere.

Could you think about an approach that keeps the temporary nature of these files? For example:
- Clean up spill files after they've been consumed
- Clean up old spill files on startup
- Keep `/tmp` but document that Docker users should mount a real filesystem there

We'd want to see a cleanup story before moving these into a persistent directory.

DOsinga:
Going to close this since no answer to the concern. it's a genuine issue though, so if you come up with a better idea, let me know.

PR #9314, opened 2026-05-18
fix: preserve thinking content for provider context

### LINKED ISSUE #9291: preserve_thinking_context ineffective for OpenAI-compatible providers
## Problem

The `preserve_thinking_context` flag has no effect for OpenAI-compatible providers (including `omlx`, `lmstudio`, `llama_swap`, etc.). Thinking content from model responses is never sent back to the model in subsequent turns, regardless of the flag's value.

## Root Cause

In `crates/goose/src/agents/reply_parts.rs:264`, the `stream_response_from_provider()` function calls `agent_visible_content()` on all messages before sending them to the provider:

```rust
let filtered_messages: Vec<Message> = messages
    .iter()
    .filter(|m| m.is_agent_visible())
    .map(|m| m.agent_visible_content())  // <-- strips Thinking content
    .collect();
```

The `agent_visible_content()` method calls `filter_for_audience(Role::Assistant)`, which unconditionally filters out `MessageContent::Thinking` and `MessageContent::RedactedThinking` (line 380 of `crates/goose/src/conversation/message.rs`):

```rust
MessageContent::Thinking(_) | MessageContent::RedactedThinking(_) => None,
```

Since thinking content is stripped before reaching the provider's format module, `format_messages_with_options()` never sees any `Thinking` content, so `reasoning_text` is always empty and the `reasoning_content` field is never emitted.

## Expected Behavior

When `preserve_thinking_context` is `true` (the default for OpenAI-engine providers), thinking content should be included in the messages sent to the provider so that models can receive their own reasoning on subsequent turns.

## Fix

In `crates/goose/src/conversation/message.rs`, change `filter_for_audience()` to preserve `Thinking` content when the audience is `Role::Assistant`:

```rust
MessageContent::Thinking(_) | MessageContent::RedactedThinking(_) => {
    if audience == Role::Assistant {
        Some(self.clone())
    } else {
        None
    }
}
```

This preserves thinking for the provider (agent) while still filtering it out for user-facing contexts (e.g., UI display, session export).

## Impact

- Anthropic provider: Not affected (uses different message flow)
- OpenAI-compatible providers: Affected (all use the `agent_visible_content()` filter)

asim48-ctrl:
## Summary
- Preserve `Thinking` and `RedactedThinking` blocks when filtering messages for the assistant/provider audience.
- Keep thinking blocks hidden from the user-facing audience filter.
- Add focused unit coverage for provider-visible thinking and user-hidden thinking.

Fixes #9291.

## Validation
- `git diff --check`

Not run: full Rust test suite/build; kept this to a lightweight source-only patch.

### DIFF
--- a/crates/goose/src/conversation/message.rs
+++ b/crates/goose/src/conversation/message.rs
@@ -377,7 +377,13 @@ impl MessageContent {
                     metadata: res.metadata.clone(),
                 }))
             }
-            MessageContent::Thinking(_) | MessageContent::RedactedThinking(_) => None,
+            MessageContent::Thinking(_) | MessageContent::RedactedThinking(_) => {
+                if audience == Role::Assistant {
+                    Some(self.clone())
+                } else {
+                    None
+                }
+            }
             _ => Some(self.clone()),
         }
     }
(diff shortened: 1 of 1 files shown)


DOsinga (approved):
Clean, correct fix for #9291. Merged main, removed the trivial `test_user_audience_filters_thinking_content` test, and pushed.

Please make sure your commits have DCO sign-off (`git commit -s`) — the original commit is missing it.

PR #9318, opened 2026-05-19
Remove vendored Windows binaries

jh-block:
Fetch pinned uv binaries during Windows packaging instead of keeping the binaries in the repository.

Switch local Windows release helpers to the MSVC target and drop references to the old GNU runtime DLL bundle.


DOsinga (approved):
Clean removal of vendored Windows binaries in favor of build-time download with SHA-256 verification. MSVC alignment with CI is correct — the GNU DLLs are no longer needed. Codex comment about cross-platform builds is not actionable since CI already runs natively on windows-latest with MSVC.

PR #9320, opened 2026-05-19
fix: send empty object instead of null for anthropic tool input

### LINKED ISSUE #9287: Error using a parameterless MCP tool with Anthropic provider
Here is an analysis of the issue from Claude Code:

# Bug: Anthropic provider serializes parameterless tool_use `input` as `null`, causing 400 from Anthropic API

## Summary

When Goose's Anthropic provider re-serializes conversation history containing a `tool_use` block for a parameterless MCP tool (a tool whose JSON Schema declares no properties), the `input` field is emitted as JSON `null` instead of `{}`. The Anthropic Messages API rejects the request with:

```
400 Bad request: messages.<N>.content.<M>.tool_use.input: Input should be an object.
```

This looks like the same bug that was fixed for Bedrock in #8121, but the equivalent fix was not applied to the Anthropic provider.

## Environment

- Goose: latest release (Desktop, Windows; CLI behavior expected to match)
- Provider: Anthropic
- Trigger: any MCP server that exposes at least one tool with an empty input schema (`{"type": "object", "properties": {}}` or equivalently no `properties` key)

## Steps to Reproduce

1. Configure an MCP extension whose tool list includes a parameterless tool, e.g. a schema like:
   ```json
   {
     "name": "list_things",
     "description": "...",
     "inputSchema": { "type": "object" }
   }
   ```
2. Start a chat with the Anthropic provider selected.
3. Send a prompt that causes the model to call the parameterless tool (or call any other tool first, then this one).
4. After the tool completes successfully, send any follow-up user message.
5. Goose constructs the next request to Anthropic and the API returns:
   ```
   400 Bad request: messages.<N>.content.<M>.tool_use.input: Input should be an object.
   ```

## Root Cause

In `crates/goose/src/providers/formats/anthropic.rs`, the tool_use payload is built directly from `tool_call.arguments`, which is `Option<serde_json::Map<String, Value>>`. When `arguments` is `None`, `serde_json::json!` serializes it as JSON `null`:

`crates/goose/src/providers/formats/anthropic.rs:174` (and the parallel `FrontendToolRequest` path around line 256):

```rust
content.push(json!({
    TYPE_FIELD: TOOL_USE_TYPE,
    ID_FIELD: tool_request.id,
    NAME_FIELD: tool_call.name,
    INPUT_FIELD: tool_call.arguments   // <-- None -> JSON null
}));
```

`arguments` ends up as `None` on several paths, including:

- Session reload from disk: `tool_result_serde.rs:38` maps `Value::Null` -> `arguments = None`.
- Conversation history that originated from a different provider where the field was stored as `null`.
- Provider switching.
- Any construction path that uses `CallToolRequestParams::new(name)` without `.with_arguments(...)`.

The **decode** path already handles this correctly:

- Streaming decoder (`anthropic.rs:829`) coerces empty/missing input to `json!({})`.
- Non-streaming `response_to_message` (`anthropic.rs:386`) receives `{}` from Anthropic and wraps it via `with_arguments`.

So the very first round trip works fine. The failure manifests on the **next** request after the tool_use block has been stored and is being replayed in history. That matches what users report: "the tool runs successfully, then the next message fails."

## Identical Bug, Already Fixed for Bedrock

- Issue: #8050 (Bedrock provider sends `null` `toolUse.input` for no-argument tool calls)
- Fix:   #8121 (patches `formats/bedrock.rs`)

The Anthropic provider was not touched in #8121. A search of the repo turns up no equivalent fix for `formats/anthropic.rs`.

## Suggested Fix

Mirror the helper pattern from #8121. At `formats/anthropic.rs:174` and the parallel `FrontendToolRequest` path:

```rust
let input_val = tool_call.arguments
    .clone()
    .map(serde_json::Value::Object)
    .unwrap_or_else(|| serde_json::Value::Object(Default::default()));

content.push(json!({
    TYPE_FIELD: TOOL_USE_TYPE,
    ID_FIELD: tool_request.id,
    NAME_FIELD: tool_call.name,
    INPUT_FIELD: input_val,
}));
```

Or, since two providers now need the same coercion, factor it out into a shared helper (e.g., `args_to_json_value` in a common module) and call it from both `formats/anthropic.rs` and `formats/bedrock.rs`.

## Workarounds (for users hitting this today)

None of these are great:

1. Avoid MCP servers with parameterless tools.
2. Wrap or patch the MCP server to add a dummy optional property to the tool's input schema, so the model is steered into emitting a non-empty input.
3. Start a new chat session before each follow-up that would replay the affected tool_use block (the first call succeeds; the failure is on replay).

Setting the affected tool to "Never allow" in Tool Permissions does **not** fix this — the `tool_use` block from Anthropic still enters conversation history and is re-serialized on the next turn.

## Impact

Any user who pairs the Anthropic provider with an MCP server that exposes one or more parameterless tools will hit this on the message immediately following the first call to such a tool. Several widely-used MCP servers (file system, environment introspection, "list X" style endpoints) commonly declare parameterless tools.

DOsinga:
When an MCP tool has no parameters, `tool_call.arguments` is `None`, which serialized to JSON `null`. The Anthropic API rejects `null` for the `input` field, requiring an empty object `{}` instead.

This is the same fix applied to Bedrock in #8121 but was missing from the Anthropic provider.

Fixes #9287

### DIFF
--- a/crates/goose-cli/src/session/input.rs
+++ b/crates/goose-cli/src/session/input.rs
@@ -149,9 +149,7 @@ pub fn get_input(
         rustyline::EventHandler::Conditional(Box::new(CtrlCHandler::new(completion_cache))),
     );
 
-    let prompt = get_input_prompt_string();
-
-    let input = match editor.readline(&prompt) {
+    let input = match editor.readline("> ") {
         Ok(text) => text,
         Err(e) => match e {
             rustyline::error::ReadlineError::Interrupted => return Ok(InputResult::Exit),
@@ -385,24 +383,6 @@ fn parse_plan_command(input: String) -> Option<InputResult> {
     };
 
     Some(InputResult::Plan(options))
-}
-
-fn get_input_prompt_string() -> String {
-    if is_vte_with_broken_emoji_width() {
-        return "> ".to_string();
-    }
-    "🪿 ".to_string()
-}
-
-/// VTE < 0.70 renders 🪿 as 1 cell while unicode-width counts 2, causing cursor offset.
-fn is_vte_with_broken_emoji_width() -> bool {
-    let Ok(vte_version) = std::env::var("VTE_VERSION") else {
-        return false;
-    };
-    let Ok(version) = vte_version.parse::<u32>() else {
-        return true;
-    };
-    version < 7000
 }
 
 fn print_help() {
--- a/crates/goose/src/providers/formats/anthropic.rs
+++ b/crates/goose/src/providers/formats/anthropic.rs
@@ -171,7 +171,7 @@ fn format_messages_with_options(
                                 TYPE_FIELD: TOOL_USE_TYPE,
                                 ID_FIELD: tool_request.id,
                                 NAME_FIELD: tool_call.name,
-                                INPUT_FIELD: tool_call.arguments
+                                INPUT_FIELD: tool_call.arguments.clone().unwrap_or_default()
                             }));
                         }
                         Err(_tool_error) => {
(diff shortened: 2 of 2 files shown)


jamadeo:
fixed in #9355

PR #9325, opened 2026-05-19
Feat/summon subagent instructions

DOsinga:
## Summary

Make the summon extension handle the @agent stuff. will also call now if you just mention the name or if it just seems fitting. So "hey @pirate, where is the treasure" should work now.

### DIFF
--- a/crates/goose/src/agents/platform_extensions/summon.rs
+++ b/crates/goose/src/agents/platform_extensions/summon.rs
@@ -35,12 +35,8 @@ use tracing::{info, warn};
 
 pub static EXTENSION_NAME: &str = "summon";
 
-/// Max characters when rendering a subagent's description in a list shown
-/// to the model (system-prompt blurb, `load` discovery output, etc.).
 const SUBAGENT_DESCRIPTION_BUDGET: usize = 160;
 
-/// Max characters for a human-readable label of a running or completed
-/// task. Used for progress display and bookkeeping, not sent to the model.
 const TASK_LABEL_BUDGET: usize = 60;
 
 fn kind_plural(kind: SourceType) -> &'static str {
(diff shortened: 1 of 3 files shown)


pirate:
Arrr matey, it worked, you have summoned me from a great slumber. 
The treasure be buried in a far off land marked by the big red image 🏴‍☠️ 
Good luck on ye journey.

crates/goose/src/agents/platform_extensions/summon.rs
```diff
@@ -291,6 +291,70 @@ pub fn discover_filesystem_sources(working_dir: &Path) -> Vec<SourceEntry> {
...
+    let mut current_kind: Option<SourceType> = None;
+    for s in &subagents {
+        if current_kind != Some(s.source_type) {
+            out.push_str(&format!("\n{}:", kind_plural(s.source_type)));
+            current_kind = Some(s.source_type);
+        }
+        out.push_str(&format!(
+            "\n• {} — {}",
+            s.name,
+            truncate(&s.description, 200)
```
jamadeo: 200 here seems a tad arbitrary? though I'm not able to find a better limit

jamadeo (approved):
nice.

As for the extension itself, do you think it should continue to be
1) called "summon"
2) (en|dis)abled from the extension page?

it is not that clear to me that toggling a box called "summon" under "extensions" would affect goose's ability to use agents I've defined

DOsinga:
> nice.
> 
> As for the extension itself, do you think it should continue to be
> 
> 1. called "summon"
> 2. (en|dis)abled from the extension page?
> 
> it is not that clear to me that toggling a box called "summon" under "extensions" would affect goose's ability to use agents I've defined

I think we should probably always enable it unless there is a setting like DISSABLE_SUBAGENTS. or maybe we should have some sort of features thing?

alexhancock:
> it is not that clear to me that toggling a box called "summon" under "extensions" would affect goose's ability to use agents I've defined

This is my main feedback too. I think we should make one extension which does sub-agents & agents and just call it "sub-agents"

crates/goose/src/agents/platform_extensions/summon.rs
DOsinga: yeah good point. also it is using a locally defined truncate instead of the utility we have. introduced two constants to do it instead

PR #9326, opened 2026-05-19
CLI to list skills with token counts

jamadeo:
A simple utility to see which skills goose discovers.

### DIFF
--- a/crates/goose-cli/src/commands/skills.rs
+++ b/crates/goose-cli/src/commands/skills.rs
@@ -2,7 +2,8 @@ use anyhow::Result;
 use console::{measure_text_width, Term};
 use goose::skills::list_installed_skills;
 
-const SEPARATOR: &str = " - ";
+const SEPARATOR: &str = " | ";
+const MIN_DESCRIPTION_WIDTH: usize = 4;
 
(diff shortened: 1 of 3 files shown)


crates/goose-cli/src/commands/skills.rs
```diff
@@ -0,0 +1,228 @@
...
+    let mut skills = list_installed_skills(Some(&cwd));
+    skills.sort_by(|a, b| a.name.cmp(&b.name));
+
+    let rows = skills
+        .iter()
+        .map(|skill| SkillRow {
+            name: skill.name.clone(),
+            description: description_preview(&skill.description),
+            description_tokens: token_counter.count_tokens(&skill.description).to_string(),
+            content_tokens: token_counter.count_tokens(&skill.content).to_string(),
```
DOsinga, line 45: hmm, this is not going to be very accurate, is it?

DOsinga (approved):

crates/goose-cli/src/commands/skills.rs
jamadeo: No, but is there a better way? This doesn't need high accuracy.

PR #9329, opened 2026-05-19
Build summon instructions per turn

DOsinga:
Move the subagent listing out of SummonClient::new (and off InitializeResult::instructions) and into a get_instructions() override on the McpClientTrait impl.

Skills already does this, and the system prompt pipeline already calls get_instructions() fresh on every reply turn via prepare_tools_and_prompt -> get_extensions_info. With summon now overriding the trait method, an agent or recipe added to .agents/ or .goose/ mid-session shows up on the next turn, instead of being frozen at session start.


lifeizhou-ap (approved):
Nice! 

The codex comments seem valid.  Can either not using cached source in load/delegate, or update the cached after get the instructions

PR #9331, opened 2026-05-19
chore: update canonical model registry

baxen:
## Summary
- Regenerate canonical model registry from models.dev
- Update provider metadata snapshot
- Includes newly available Gemini entries such as `google/gemini-3.5-flash`

## Validation
- `source bin/activate-hermit && just build-canonical-models`
- `cargo fmt`
- `jq empty crates/goose/src/providers/canonical/data/canonical_models.json`
- `jq empty crates/goose/src/providers/canonical/data/provider_metadata.json`

### DIFF
--- a/crates/goose/src/model.rs
+++ b/crates/goose/src/model.rs
@@ -502,7 +502,7 @@ mod tests {
                 ("GOOSE_CONTEXT_LIMIT", None::<&str>),
             ]);
             let config =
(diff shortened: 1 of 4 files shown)


jamadeo (approved):

DOsinga (approved):
I think this is supposed to run when we do a new release automatically but no harm in merging it before. also we should do a release

PR #9338, opened 2026-05-20
WIP: fixes to save session bloat and confusion

michaelneale:
See [BRAINROT.md](./BRAINROT.md).

some things tackled: 
 - Caps fat tool results — every tool_response text block limited to 16 KB; marker tells the model how to fetch more
 - Dedupes re-reads — same file read twice with same args → second copy becomes a pointer, not a duplicate dump
 - Caps code_execution bundle fields — one fat field in { a, b, c } no longer drowns out the small useful ones (8 KB per field)
 - Pushes the model toward narrow reads — execute_typescript desc now tells it to use line/limit/head/grep instead of dumping whole files
 - Real read tool with a 2000-line default cap — no more shell({command:"cat <bigfile>"}) to get a structured read
 - Structured compaction summaries — explicit Goal / Decisions / Next Step sections + <read-files>/<modified-files> blocks instead of freeform <analysis> prose
 - tweaked post compaction prompts — model is told it's working from a summary and to admit gaps, not to hide that summarization happened (kills the confabulation cue)


crates/goose/src/agents/reply_parts.rs
```diff
@@ -264,6 +264,11 @@ impl Agent {
             .map(|m| m.agent_visible_content())
             .collect();
 
+        // Replace duplicate tool-result text blocks with short pointers
+        // (P1 in BRAINROT.md). The session DB and on-screen view keep the
```
jh-block: These "P1 in BRAINROT.md" etc comments should be elided. Overall maybe there's a few too many comments in general

crates/goose/src/conversation/dedup.rs
```diff
@@ -0,0 +1,316 @@
+//! Pre-send deduplication of repeated tool-result text blocks.
+//!
+//! When the same `(tool_name, arguments)` pair returns the same text body
+//! twice within a single conversation (e.g. the agent re-reads `plan.md`
+//! three times), the second and subsequent copies are replaced with a short
+//! pointer back to the first occurrence. The session DB and the on-screen
+//! view are untouched — only the conversation snapshot sent to the LLM is
+//! deduped.
+//!
+//! See `BRAINROT.md` (P1) for rationale.
```
jh-block: I think we probably should leave BRAINROT.md out of the repo and remove these references to it. It will itself rot, otherwise

crates/goose/src/conversation/message.rs
```diff
@@ -597,6 +597,80 @@ impl MessageContent {
...
+    while end > 0 && !text.is_char_boundary(end) {
+        end -= 1;
+    }
+    let elided = text.len() - end;
+    let mut out = String::with_capacity(end + 220);
+    // `end` was just walked back to a char boundary above, so this slice is safe.
+    out.push_str(text.get(..end).unwrap_or(""));
+    out.push_str(&format!(
+        "\n\n[goose: truncated {elided} bytes of tool output (cap={cap} bytes, set {env}=0 to disable, or pass a larger value). \
+Re-call the tool with narrower parameters (e.g. line/limit, grep, head, or a more specific path) to inspect the rest.]",
```
jh-block: What if the tool is not idempotent? The model may not safely be able to re-call it. It might be better to shunt the full result to a file and provide the filename in case the model wants to inspect other parts of the output beyond the truncated part.

crates/goose/src/agents/platform_extensions/code_execution.rs
```diff
@@ -28,6 +28,108 @@ use tokio_util::sync::CancellationToken;
...
+        serde_json::Value::String(s) => {
+            if s.len() > cap {
+                let mut end = cap;
+                while end > 0 && !s.is_char_boundary(end) {
+                    end -= 1;
+                }
+                let elided = s.len() - end;
+                let truncated = s.get(..end).unwrap_or("");
+                let marker = format!(
+                    "{truncated}\n[goose: truncated {elided} bytes from this field (per-field cap={cap} bytes, set {env}=0 to disable). Re-run with a narrower selection (e.g. line/limit, head, grep) or return only the slice you need.]",
```
jh-block: (as with the text truncation) What if the tool is not idempotent? The model may not safely be able to re-run it. It might be better to shunt the full result to a file and provide the filename, and suggest that jq could be used to fetch the truncated leaf field values.

Also, it might make sense to prepend or append a single "Some fields were truncated, [advice]" notification to the whole JSON payload, and then just put something short like "[truncated]" in the truncated fields.

I think saying "set ENV_VAR=0 to disable" to the model is not really helpful since the model can't set env vars? a tool parameter might be better?

crates/goose/src/agents/reply_parts.rs
michaelneale: yeah this is a speed run to try, yeah will edit it down if it works how I hope

crates/goose/src/conversation/dedup.rs
michaelneale: yeah it is only while WIP, sometimes I find it helpful when the session lets me down to reflect back

crates/goose/src/conversation/message.rs
michaelneale: hrm yeah - really should only be for read ones I think?

crates/goose/src/agents/platform_extensions/code_execution.rs
michaelneale: I'm also hoping to not need anything specific in code mode here - so might be able to drop this.

FredApps:
The **Smoke Tests (Code Execution)** failure is a flaky/infra timeout — vitest integration test hit the time limit with no assertion error. Not caused by this PR. Re-run the job.

DOsinga:
Closing this out as it's gone stale (no activity in a while). No worries though — please feel free to reopen it when it's current again, or open a fresh PR if that's easier. Thanks for the work here! 🙏

PR #9343, opened 2026-05-20
Add goose://resume session deep link

### LINKED ISSUE #9187: Feature Request: Session Resume Deep Link (goose://resume/<session-id>) for Workspace Save/Restore

impetus82:
Summary:
- add a goose://resume/<session-id> handler alongside new-session
- create a chat window with resumeSessionId populated from the deep link path
- cover second-instance, startup handleProtocolUrl, and macOS open-url dispatch paths
- ignore malformed or empty resume links without crashing

Verification:
- git diff --check
- full desktop typecheck not run locally because this clone does not have the Goose workspace dependencies installed

Fixes #9187

No payment expected upfront. If this PR is useful and gets merged, an optional tip/donation is appreciated so I can keep doing small maintainer-unblocking fixes.

### DIFF
--- a/ui/desktop/src/main.ts
+++ b/ui/desktop/src/main.ts
@@ -420,6 +420,15 @@ if (process.platform !== 'darwin') {
           return;
         }
 
+        if (parsedUrl.hostname === 'resume') {
+          app.whenReady().then(async () => {
+            const recentDirs = loadRecentDirs();
+            const openDir = recentDirs.length > 0 ? recentDirs[0] : null;
+            await createResumeChatWindow(parsedUrl, openDir || undefined);
+          });
+          return;
+        }
+
         // For non-bot URLs, continue with normal handling
         handleProtocolUrl(protocolUrl);
       }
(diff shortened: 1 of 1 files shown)


impetus82:
Addressed the P1 review feedback in commit 160cbf6.

The macOS open-url resume path now sets openUrlHandledLaunch only when createResumeChatWindow succeeds, so invalid or empty goose://resume URLs can fall back to normal window creation instead of leaving the app with no visible window.

Verification: git diff --check.

jamadeo (approved):
Looks good, thank you for the contribution!

PR #9347, opened 2026-05-20
check for responses API support in databricks

jamadeo:
Checks the databricks endpoint information for responses API support instead of inferring it from the endpoint name.


FredApps:
The **Smoke Tests (Code Execution)** failure is a flaky/infra timeout — vitest integration test hit the 4-minute wall with no assertion error. Not caused by this PR. Re-run the job.

DOsinga:
Pushed a merge with `main` plus a follow-up fix for the codex P1.

**codex P1 (responses routing regression):** the endpoint metadata fetch is best-effort (`resolve_endpoint_info_cached(...).await.ok()`), so when it fails `supports_responses_api` was always `false`, which would route responses-only OpenAI models (gpt-5 / o-series) to `/invocations` and break them. Previously the name-based check covered this case. I restored the name-based detection as a *fallback* that's only consulted when metadata is unavailable, keeping metadata authoritative when present (new `uses_responses_api` helper + tests `responses_routing_prefers_metadata_over_model_name` and `responses_routing_falls_back_to_model_name_without_metadata`).

Also resolved the merge conflict against `main`'s move of `extract_reasoning_effort` into `goose_providers::formats::openai`.

One thing for you @jamadeo: the original commit `7f35ef427` isn't signed off, so the DCO check is failing. Could you rebase/amend that commit with a `Signed-off-by` line (`git commit --amend -s` then force-push) so we can merge?

DOsinga:
Good catch on the P2. `pending_config` describes the target of an in-progress/failed update rather than the live queryable config, so routing on it could send live requests to `/serving-endpoints/responses` for an endpoint whose active config is still chat-only. Pushed de5a981b7 to base responses routing on the active `config` only. Added a test (`endpoint_metadata_ignores_pending_config_for_responses_routing`) that asserts a chat-active endpoint with a pending responses config is *not* routed to the responses endpoint. (Note: `pending_config` is still used for upstream model-name/context resolution, which is unaffected by this concern.)

DOsinga (approved):

PR #9351, opened 2026-05-21
Fix: deleting a recently created session and then creating a new session shows the deleted session's content

monroewilliams:
## Summary
fix(ui): clear session cache on delete to prevent recycled ID showing stale data

Session IDs are date-based (YYYYMMDD_N) and can be recycled when a session is deleted and a new one is created. The resultsCache in useChatStream.ts cached session data by ID, so a recycled ID would return the old session's cached messages for the new session.

Also remove deleted sessions from activeSessions so the sidebar updates.

### Testing
Manual test:
- create a new session, and run some prompts so it's not empty
- delete that session
- create another new session
- before change: the new session shows the content from the deleted session
- after change: the new session is empty

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:


jamadeo:
Looks like a reasonable thing to do, but I think reusing a session ID here is probably not something we should be doing. Thoughts @DOsinga ?

DOsinga (approved):
Thanks for the fix! This is a clean, well-scoped solution to a real bug — recycled date-based session IDs returning stale cached data. The new `clearSessionCache` call on delete and the `activeSessions` cleanup in `App.tsx` follow the existing event-handler patterns nicely. All CI checks pass and DCO is signed off. Approving.

PR #9352, opened 2026-05-21
feat(providers): add NEAR AI Cloud provider

PierreLeGuen:
## Summary
Adds NEAR AI Cloud as a fixed declarative OpenAI-compatible provider so users can configure `NEARAI_API_KEY` and use `https://cloud-api.near.ai/v1`.

### Implementation Notes
- Adds `nearai` provider metadata with dynamic model discovery and TEE-oriented fallback models.
- Reuses the existing OpenAI-compatible provider path with NEAR-specific request normalization for chat completions.
- Documents NEAR AI Cloud in the supported provider list and adds focused provider tests.

### Testing
- `cargo fmt`
- `cargo test -p goose nearai`
- `cargo test -p goose providers::openai::tests::sanitize`
- `cargo test -p goose providers::openai::tests::responses_api_routing_uses_model_family_unless_path_forces_chat`

### Related Issues
None.


jamadeo (approved):
Looks good. I think on our side we can do a better job of expressing the ways in which providers might need to change the shape of the API structure. But that can happen outside this scope.

PR #9357, opened 2026-05-21
fix(agents): serialize per-session agent creation to stop duplicate MCP init

### LINKED ISSUE #9031: Race condition while connecting to MCP servers & elicitation errors
**Describe the bug**

### Issue 1
When Goose Desktop is spawning up MCP servers, there is a weird race condition happening. From the screen recording, we can see that it actually spawns the same MCP server three times and quickly abandoned two of them. It does not hurt any performance but it's a buggy behavior.

https://github.com/user-attachments/assets/40db9b82-11fe-469b-9383-9785679b41d4
 
### Issue 2
After users respond to the elicitation, the result is not successfully delivered to the MCP server, and it causes Goose Desktop hanging forever.

I didn't encounter the same failed to load session problem in https://github.com/aaif-goose/goose/issues/8531 but the other parts are same. 

This bug seems to be fixed by https://github.com/aaif-goose/goose/pull/8999. From the demo picture, the LLM knows the response sent by users and it's still responsive. I would like to test it myself but I can not successfully build the dev version of Goose Desktop - https://github.com/aaif-goose/goose/issues/8757

#### Elicitation not delivered
https://github.com/user-attachments/assets/baf0666f-92b1-4b4d-a4cb-ddcf83ffa248

#### Goose Desktop hanging after elicitation

https://github.com/user-attachments/assets/09588581-5fa8-4386-b3e7-ba9fd5aae716

#### Expected behavior from goose cli

https://github.com/user-attachments/assets/ae8df36a-0a73-40c7-bcf8-fa42222561cb

---

**To Reproduce**

#### Issue 1
1. Host a MCP server with http transport that logs every `initialize` request.
2. We can see from the log that Goose Desktop send the `initialize` request three times.

#### Issue 2
1. Connect to a MCP server that has tools sending elicitation request to MCP clients.
2. Submit the elicitation and ask for the result you submit to see if the LLM really get the response.
---

**Expected behavior**
#### Issue 1
Goose Desktop should only connect each MCP server once.

#### Issue 2
MCP server should receive the response of the elicitation after users submit the response from Goose Desktop. The LLM should still be able to respond to users input after an elicitation request.

---

- **OS & Arch:** macOS 15.7.5 M3 chip
- **Interface:** UI
- **Version:** 1.33.1
- **Extensions enabled:** A dummy MCP server for testing
- **Provider & Model:** Google-gemini-3.1-pro-preview, Google-gemini-2.5-pro

fresh3nough:
## Summary

`AgentManager::get_or_create_agent` in `crates/goose/src/execution/manager.rs`
had a TOCTOU race: the slow path released the `sessions` `RwLock` between the
initial cache check and the expensive create+restore+load work, so concurrent
callers for the same brand-new session id (typically `start_agent`'s
background extension-loading task and a frontend `resume_agent` racing through
`get_agent_for_route`) could each construct their own `Agent` and
independently connect to every MCP server.

The existing double-check after the slow work hid the symptom from callers
(only one `Arc<Agent>` won the cache slot), but the discarded `Agent`
instances had already finished `load_extensions_from_session`, which is
where the duplicated MCP `initialize` traffic originated. Users observed
their MCP extensions being initialized 2-3x at startup.

## Change

Introduce a per-session creation `Mutex` on `AgentManager`:

- After the fast cache check misses, acquire the per-session lock
  (`Arc<Mutex<()>>` looked up from a `HashMap<String, _>` keyed by
  session id).
- Re-check the `sessions` cache under the lock; if another caller already
  populated it while we were waiting, return that `Arc<Agent>`.
- Only then run the expensive `Agent::with_config` /
  `restore_provider_from_session` / `load_extensions_from_session` pipeline.
- The `HashMap` entry is cleaned up in `remove_session`. Callers still
  holding a clone of the `Arc<Mutex<()>>` keep the mutex alive until they
  finish — so the cleanup is race-free.

Different sessions remain fully independent; the only synchronization is
between callers competing for the same session id. The fix does not hold
the broad `sessions` `RwLock` across the network I/O of
`load_extensions_from_session`, so unrelated sessions are not blocked by
slow MCP startup.

## Testing

The bug is invisible to callers that only check the returned `Arc<Agent>`
(every caller still gets the cached `Arc`), so I verified the fix locally
with a temporary `#[cfg(test)]` counter inserted on the slow path:

- Pre-fix (per-session lock disabled, `tokio::task::yield_now()` to
  schedule the racers): 20 concurrent `get_or_create_agent` calls for the
  same new session id produced **20** counter increments, exactly
  reproducing the duplicate-creation behavior described in #9031.
- Post-fix: the same 20 concurrent calls produced **1** counter
  increment, confirming the per-session mutex serializes creation.

The instrumentation and the temporary test were removed before this
commit (race-condition timing tests are flaky in CI), but verification was
done end-to-end against this exact change. The remaining 12 AgentManager
unit tests still pass:

- `cargo test -p goose --lib execution::manager` — 12/12 pass.
- `cargo test -p goose --lib` — 1619/1619 pass, no regressions.
- `cargo clippy --all-targets -- -D warnings` — clean.
- `cargo fmt` — clean.

## Related Issues

Fixes #9031

### DIFF
--- a/crates/goose/src/execution/manager.rs
+++ b/crates/goose/src/execution/manager.rs
@@ -10,7 +10,7 @@ use lru::LruCache;
 use std::collections::HashMap;
 use std::num::NonZeroUsize;
 use std::sync::Arc;
-use tokio::sync::{OnceCell, RwLock};
+use tokio::sync::{Mutex, OnceCell, RwLock};
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, info};
 
@@ -25,6 +25,14 @@ pub struct AgentManager {
     default_provider: Arc<RwLock<Option<Arc<dyn crate::providers::base::Provider>>>>,
     default_mode: GooseMode,
     cancel_tokens: Arc<RwLock<HashMap<String, CancellationToken>>>,
+    /// Per-session creation locks.  When `get_or_create_agent` misses the
+    /// `sessions` cache it acquires the per-session lock before doing the
+    /// expensive work (provider restore, MCP extension initialization) so
+    /// concurrent callers for the same session never race into doing the
+    /// work twice.  Entries are inserted on demand and removed when the
+    /// session is removed; the underlying `Arc<Mutex<()>>` stays alive as
+    /// long as any caller still holds it, even after removal.
+    creation_locks: Arc<Mutex<HashMap<String, Arc<Mutex<()>>>>>,
 }
 
 impl AgentManager {
@@ -46,6 +54,7 @@ impl AgentManager {
             default_provider: Arc::new(RwLock::new(None)),
             default_mode,
             cancel_tokens: Arc::new(RwLock::new(HashMap::new())),
+            creation_locks: Arc::new(Mutex::new(HashMap::new())),
         };
 
         Ok(manager)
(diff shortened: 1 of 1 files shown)


crates/goose/src/execution/manager.rs
```diff
@@ -136,21 +199,53 @@ impl AgentManager {
...
+    /// first to clone the inner `Arc`.
+    ///
+    /// If a waiter is still in flight (strong_count > 1) we leave the
+    /// entry in place so the in-flight callers continue to serialize
+    /// through the same lock; a later removal or eviction will sweep it.
+    async fn prune_creation_lock(&self, session_id: &str) {
+        let mut locks = self.creation_locks.lock().await;
+        let in_use = locks
+            .get(session_id)
+            .is_some_and(|lock| Arc::strong_count(lock) > 1);
```
jamadeo, line 246: Could the clean-up/pruning be a bit less manual if you instead have a `HashMap<String, Weak<Mutex<()>>>`? Then the mutexes are dropped when no in-flight references exist. You'd still have map entries that you'd want to prune, but they'd point to nothing instead of unused Mutexes.

jamadeo (approved):
Nice, good fix. Left one comment about an idea to maybe improve it a little bit. LMK what you think

crates/goose/src/execution/manager.rs
fresh3nough: I think its cleaner and a great idea, i will implement it and push :)

fresh3nough:
@jamadeo should be good now

PR #9360, opened 2026-05-22
Improve dependency hygiene

jh-block:
* Disabled default features on dependencies across the board. Added back what we actually use.
* Expressed true minimum dependency versions (verified with `cargo +nightly update -Z direct-minimal-versions && cargo check`)
* Updated a few crates that caused trouble with the above.
* Verified build and run with all dependencies at the minimal and maximal compatible versions.

* Switched AWS to use smithy-transport-reqwest which delegates HTTP requests to reqwest which we already use. This avoids a second HTTP client being compiled in, and also avoids the problem that AWS's client uses rustls unconditionally, bringing us one step closer to the native-tls feature actually doing what it promises.

* Put nostr behind a feature flag

* Put self-update functionality (and thus the sigstore-verify dep) behind a feature flag in goose-cli

With all of this, a minimal goose-cli (native TLS, no optional features) clocks in at 99MB, down from 223MB.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,6 @@ members = [
     # Mainly for cargo-machete to not error out during inspection.
     "vendor/v8",
 ]
-exclude = ["ui/goose2/src-tauri"]
 resolver = "2"
 
 [workspace.package]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ bytes = { version = "1.10.1", default-features = false, features = ["std"] }
 candle-core = { version = "0.10", default-features = false }
 candle-nn = { version = "0.10", default-features = false }
 chrono = { version = "0.4.44", default-features = false, features = ["serde", "std"] }
-clap = { version = "4.1.14", default-features = false, features = ["derive", "std"] }
+clap = { version = "4.1.14", default-features = false, features = ["derive", "std", "help", "suggestions", "usage", "color", "error-context"] }
 dirs = { version = "5", default-features = false }
 dotenvy = { version = "0.15.7", default-features = false }
 env-lock = { version = "1", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,72 +25,72 @@ rmcp = { version = "1.4", default-features = false, features = ["schemars", "aut
 agent-client-protocol-schema = { version = "0.12", default-features = false, features = ["unstable"] }
 agent-client-protocol = { version = "0.11", default-features = false }
 arboard = { version = "3", default-features = false }
-anyhow = { version = "1.0.102", default-features = false }
+anyhow = { version = "1.0.102", default-features = false, features = ["std"] }
 async-stream = { version = "0.3.6", default-features = false }
 async-trait = { version = "0.1.89", default-features = false }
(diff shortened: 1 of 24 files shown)


crates/goose/src/session/mod.rs
```diff
@@ -2,6 +2,7 @@ mod chat_history_search;
 mod diagnostics;
 pub mod extension_data;
 mod legacy;
+#[cfg(feature = "nostr")]
```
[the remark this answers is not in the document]
jamadeo: good catch I think? need more flags in `crates/goose-server/src/routes/session.rs`?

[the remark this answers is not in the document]
jh-block: yes good catch indeed. fixed.

r0x0d:
I love this. It will help us a lot with getting goose updated to the newest version across Fedora/RHEL. I was really struggling with sigstore-verify, as it does not honor completely the native-tls feature they have :cry:.

crates/goose/Cargo.toml
```diff
@@ -9,15 +9,7 @@ repository.workspace = true
 description.workspace = true
 
 [features]
-default = [
```
r0x0d, line 12: I think we may need at least `rustls` as a default feature here and in the other places where it might be used

jh-block: This crate `goose` is the internal library crate; my idea was that binary crates that depend on it should be responsible for setting the defaults

crates/goose/Cargo.toml
```diff
@@ -141,43 +136,46 @@ sqlx = { version = "0.8", default-features = false, features = [
...
-byteorder = { version = "1.5.0", optional = true }
-tokenizers = { version = "0.21.0", default-features = false, features = ["onig"], optional = true }
-symphonia = { version = "0.5", features = ["all"], optional = true }
-rubato = { version = "0.16", optional = true }
+candle-core = { workspace = true, optional = true }
+candle-nn = { workspace = true, optional = true }
+candle-transformers = { version = "0.10", default-features = false, optional = true }
+byteorder = { version = "1.5", default-features = false, features = ["std"], optional = true }
+tokenizers = { version = "0.21", default-features = false, features = ["onig"], optional = true }
+symphonia = { version = "0.5", default-features = false, features = ["aac", "adpcm", "alac", "isomp4", "mkv", "mp3", "pcm", "vorbis", "wav"], optional = true }
```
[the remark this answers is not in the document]
jh-block: the server accepting `audio/webm;codecs=opus` seems to be a bug; symphonia doesn't have opus support on main either

crates/goose/Cargo.toml
jh-block: (I have since modified `goose` so that it can build successfully with no features enabled at all)

Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = [
     # Mainly for cargo-machete to not error out during inspection.
     "vendor/v8",
 ]
+exclude = ["ui/goose2/src-tauri"]
```
jamadeo: #9368 merged, so you shouldn't need this

jamadeo (approved):

PR #9366, opened 2026-05-22
Fix desktop chat search session limiting

angiejones:
## Summary

Searching previous sessions from Desktop was only returning chats from like the last day or so. This is because it was using the same search behavior as ChatRecall. That search is designed for agents looking through past messages, so its limit applied to matching **messages**. In the Desktop UI, that meant one recent chat with several matching messages could use up the entire result limit, hiding older chats that also matched the search.

This fixes Desktop chat history search so it returns a better spread of matching previous chats, instead of only showing the most recent few chats when a search term appears a lot.

This change gives Desktop chat search its own search path where the limit applies to matching chats/sessions instead of individual messages.

## What changed

- Desktop chat history search now returns distinct matching chats.
- Recent chats with many matching messages no longer crowd out older matching chats.
- ChatRecall behavior is preserved.

## Why separate Desktop search from ChatRecall?

Desktop search and ChatRecall have different jobs:

- **Desktop search** should quickly show a useful list of previous chats that match what the user typed.
- **ChatRecall** should help the agent search through past messages for context and details.

Because those goals are different, sharing the exact same search behavior caused Desktop search to behave poorly. Separating them keeps ChatRecall’s existing behavior intact while making Desktop search work the way users expect.

## Testing

Ran:

```bash
cargo test -p goose search_chat --lib
cargo test -p goose session_manager::tests --lib
```

Also added tests to make sure:
  - ChatRecall keeps its existing message-based search behavior.
  - Desktop search limits results by chats/sessions.
  - Desktop search filters still work with dates, excluded sessions, and session types.

### DIFF
--- a/crates/goose-server/src/routes/session.rs
+++ b/crates/goose-server/src/routes/session.rs
@@ -685,9 +685,9 @@ async fn search_sessions(
         .and_then(|s| chrono::DateTime::parse_from_rfc3339(&s).ok())
         .map(|dt| dt.with_timezone(&chrono::Utc));
 
-    let search_results = state
+    let sessions = state
         .session_manager()
-        .search_chat_history(
+        .search_chat_sessions(
             query,
             Some(limit),
             after_date,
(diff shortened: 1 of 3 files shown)


michaelneale (approved):
Makes sense, never really intended chatrecall to be universal search

PR #9385, opened 2026-05-22
feat: tui command on goose-cli

alexhancock:
## Summary
* Adds a rust based launcher for the new tui
* In `goose-cli`: `goose tui` will now launch https://www.npmjs.com/package/@aaif/goose
* It supports running the local version if running goose-cli locally or the deployed version if local is not present
* This is one step closer to turning the default interactive experience for terminal to the TUI

### Testing
Local usage

### Related Issues
https://github.com/aaif-goose/goose/issues/6642

`cargo run --bin goose -- tui`

Screenshot 2026-05-22 at 3 15 27 PM

### DIFF
--- a/crates/goose-cli/src/commands/tui.rs
+++ b/crates/goose-cli/src/commands/tui.rs
@@ -1,0 +1,99 @@
+use anyhow::{anyhow, Context, Result};
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+const TUI_NPM_SPEC_ENV: &str = "GOOSE_TUI_NPM_SPEC";
+const TUI_REL_PATH: &str = "ui/text/dist/tui.js";
+const DEFAULT_NPM_SPEC: &str = "@aaif/goose@latest";
+const NPM_BIN_NAME: &str = "goose-tui";
+
+enum TuiSource {
+    LocalScript(PathBuf),
+    Npx(String),
+}
+
+fn find_local_script() -> Option<PathBuf> {
+    let exe = std::env::current_exe().ok()?;
+    let exe_dir = exe.parent().unwrap_or_else(|| Path::new("."));
+
+    let mut dir = Some(exe_dir.to_path_buf());
+    for _ in 0..6 {
+        if let Some(d) = dir.clone() {
+            let candidate = d.join(TUI_REL_PATH);
+            if candidate.is_file() {
+                return Some(candidate);
+            }
+            dir = d.parent().map(Path::to_path_buf);
+        }
+    }
+
+    if let Ok(cwd) = std::env::current_dir() {
+        let candidate = cwd.join(TUI_REL_PATH);
+        if candidate.is_file() {
+            return Some(candidate);
+        }
+    }
+
+    None
+}
+
+fn resolve_source() -> TuiSource {
+    if let Some(script) = find_local_script() {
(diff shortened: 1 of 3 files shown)


crates/goose-cli/src/commands/tui.rs
```diff
@@ -0,0 +1,107 @@
...
+const DEFAULT_NPM_SPEC: &str = "@aaif/goose@latest";
+const NPM_BIN_NAME: &str = "goose-tui";
+
+enum TuiSource {
+    LocalScript(PathBuf),
+    Npx(String),
+}
+
+fn find_local_script() -> Option<PathBuf> {
+    if let Ok(p) = std::env::var(TUI_SCRIPT_ENV) {
```
DOsinga: should we get this from the settings instead? seems like something you might want to set in config.yaml

crates/goose-cli/src/commands/tui.rs
```diff
@@ -0,0 +1,107 @@
...
+        let path = PathBuf::from(p);
+        if path.is_file() {
+            return Some(path);
+        }
```
DOsinga: I think we should probably error out here if we can't find the file -- the user set it, it doesn't exist, trying to find it elsewhere seems wrong? die early, die often

crates/goose-cli/src/commands/tui.rs
```diff
@@ -0,0 +1,107 @@
...
+            cmd.arg("--yes")
+                .arg("--package")
+                .arg(spec)
+                .arg("--")
+                .arg(NPM_BIN_NAME)
+                .args(args);
+            Ok(cmd)
+        }
+    }
+}
```
DOsinga, line 66: so this doesn't actually run the current goose binary as the source of ACP, right?

crates/goose-cli/src/commands/tui.rs
```diff
@@ -0,0 +1,107 @@
...
+    let mut dir = Some(exe_dir.to_path_buf());
+    for _ in 0..6 {
+        if let Some(d) = dir.clone() {
+            let candidate = d.join(TUI_REL_PATH);
+            if candidate.is_file() {
+                return Some(candidate);
+            }
+            dir = d.parent().map(Path::to_path_buf);
+        }
+    }
```
DOsinga, line 28: what is the use case here? if I am developing the TUI, am I likely to use this path?

DOsinga (approved):

crates/goose-cli/src/commands/tui.rs
alexhancock: The TUI does. So we have some psuedo-cycles here... different commands but:

goose tui -> runs TUI -> uses `goose acp` (or connects to server via HTTP/WS)

crates/goose-cli/src/commands/tui.rs
alexhancock: Yeah I thought it could be helpful that when running `cargo run --bin goose -- tui` it also runs the tui from source

crates/goose-cli/src/commands/tui.rs
alexhancock: removed

crates/goose-cli/src/commands/tui.rs
alexhancock: removed

PR #9387, opened 2026-05-22
fix(providers): detect image paths with spaces

### LINKED ISSUE #9576: Local vision: image paths containing spaces are silently not detected (detect_image_path)
## Summary
Attaching a local image by putting its path in the prompt silently does nothing when the path contains spaces. macOS screenshot filenames always contain spaces (e.g. `~/Desktop/Screenshot 2026-06-02 at 1.23.45 PM.png`), so local vision appears to "not work" — no error is surfaced, the image just isn't attached to the request.

## Cause
`detect_image_path` in `crates/goose/src/providers/utils.rs` split candidate paths on whitespace, so any path was truncated at its first space and never matched a real file on disk.

## Expected
A pasted absolute path to an existing image — including paths with spaces and the backslash-escaped characters Finder inserts on drag-and-drop — is detected and attached as a multimodal image part.

## Fix
Addressed by #9387, which backward-scans from each image extension and validates candidates against the filesystem with Unicode-aware boundary checks. Further hardening is tracked in #9398 (more shell escapes) and #9399 (Unicode/emoji separators); atomic-paste UX in #9396.

jlmalone:
Closes #9576

## Summary
`detect_image_path` dropped any image path containing a space — most importantly macOS screenshots like `Screenshot 2026-05-04 at 10.32.57 PM.png`. Local vision through OpenAI-format providers (LM Studio, Ollama, etc.) then silently received no image, and the model would reply that it couldn't see the file. The function has since moved out of `crates/goose/src/providers/utils.rs` into the new `goose-providers` crate (`crates/goose-providers/src/images.rs`) and was rewritten to split on whitespace there, which reintroduced the same blind spot — this PR re-applies and hardens the fix in its new home.

## Change (`crates/goose-providers/src/images.rs`)
Replace the whitespace split with an extension-anchored scan:

- For each `.png` / `.jpg` / `.jpeg` occurrence (case-insensitive, allocation-free via `find_ascii_ci` so the returned byte index stays valid for slicing), walk backward over `/`-rooted candidates and return one that is an existing image file (`is_absolute` + `is_file` + magic-byte `is_image_file`).
- **Boundary guards** so the heuristic doesn't over-match:
  - The extension must *terminate* the candidate (next char is `/`, whitespace, a quote, or end of input) — so `/tmp/shot.png.backup` isn't truncated to `/tmp/shot.png`.
  - The leading `/` must follow a whitespace/quote boundary — so a local path isn't carved out of a URL like `https://host/some/local/path.png`.
  - Quote characters terminate a candidate, so quoted spaced paths (`"/tmp/Screen Shot.png"`) match.
- **Spaced filenames:** keep the first referenced path, but allow a longer match anchored at the same start to extend it (an earlier extension can be a prefix of the real, longer filename, e.g. `Screen Shot.png edited.jpg`).
- The backward scan is bounded by `MAX_PATH_LEN`, keeping the function linear even on extension-heavy prompts.

## Credit
The `main` merge, the URL/longer-extension boundary hardening, and the `cargo fmt` fix were pushed to this branch by @DOsinga — thank you.

## Test plan (`images.rs` unit tests, all passing)
- [x] `test_detect_image_path` — existing relative / fake / nonexistent behavior preserved
- [x] `test_detect_image_path_with_spaces` — absolute and quoted paths containing spaces
- [x] `test_detect_image_path_ignores_urls_and_longer_extensions` — URL suffix and `.png.backup` not mis-extracted
- [x] `test_detect_image_path_ignores_extension_flood` — bounded scan on extension-heavy input
- [x] `test_load_image_file`
- [x] `cargo fmt` clean; `cargo clippy -p goose-providers --all-targets` and the `images` tests pass locally.

### DIFF
--- a/crates/goose-providers/src/images.rs
+++ b/crates/goose-providers/src/images.rs
@@ -37,6 +37,7 @@ pub fn detect_image_path(text: &str) -> Option<&str> {
     const EXTENSIONS: [&str; 3] = [".png", ".jpg", ".jpeg"];
     const MAX_PATH_LEN: usize = 4096;
 
+    let mut best: Option<&str> = None;
     let mut from = 0;
     while from < text.len() {
         let Some(end) = EXTENSIONS
@@ -71,14 +72,20 @@ pub fn detect_image_path(text: &str) -> Option<&str> {
                     };
                     let path = Path::new(candidate);
                     if path.is_absolute() && path.is_file() && is_image_file(path) {
-                        return Some(candidate);
+                        // A whitespace-terminated extension is ambiguous: the
+                        // filename may continue to a later extension, so keep
+                        // the longest existing match rather than the first.
+                        if best.is_none_or(|b| candidate.len() > b.len()) {
+                            best = Some(candidate);
+                        }
+                        break;
                     }
                 }
             }
         }
         from = end;
     }
-    None
+    best
 }
 
 /// Case-insensitive ASCII substring search returning a byte index into
@@ -232,6 +239,16 @@ mod tests {
         // unquoted path.
         let text = format!("here {}\" trailing", png_path_str);
         assert_eq!(detect_image_path(&text), Some(png_path_str));
+
+        // When a spaced filename contains an earlier image extension, prefer
+        // the longer existing candidate over the embedded prefix.
+        let edited = temp_dir.path().join("Screen Shot.png edited.jpg");
+        std::fs::write(&edited, png_data).unwrap();
+        let edited_str = edited.to_str().unwrap();
+        let prefix = temp_dir.path().join("Screen Shot.png");
+        std::fs::write(&prefix, png_data).unwrap();
+        let text = format!("look at {}", edited_str);
+        assert_eq!(detect_image_path(&text), Some(edited_str));
     }
 
     #[test]
--- a/crates/goose-providers/src/images.rs
+++ b/crates/goose-providers/src/images.rs
@@ -37,7 +37,7 @@ pub fn detect_image_path(text: &str) -> Option<&str> {
     const EXTENSIONS: [&str; 3] = [".png", ".jpg", ".jpeg"];
     const MAX_PATH_LEN: usize = 4096;
 
-    let mut best: Option<&str> = None;
+    let mut best: Option<(usize, &str)> = None;
     let mut from = 0;
     while from < text.len() {
         let Some(end) = EXTENSIONS
@@ -72,11 +72,16 @@ pub fn detect_image_path(text: &str) -> Option<&str> {
                     };
                     let path = Path::new(candidate);
                     if path.is_absolute() && path.is_file() && is_image_file(path) {
-                        // A whitespace-terminated extension is ambiguous: the
-                        // filename may continue to a later extension, so keep
-                        // the longest existing match rather than the first.
-                        if best.is_none_or(|b| candidate.len() > b.len()) {
-                            best = Some(candidate);
+                        // Keep the first referenced path, but allow a longer
+                        // match anchored at the same start to extend it (a
+                        // whitespace-terminated extension may be a prefix of a
+                        // spaced filename ending in a later extension).
+                        match best {
+                            Some((best_start, _)) if start == best_start => {
+                                best = Some((start, candidate));
+                            }
+                            None => best = Some((start, candidate)),
+                            Some(_) => {}
                         }
                         break;
                     }
@@ -85,7 +90,7 @@ pub fn detect_image_path(text: &str) -> Option<&str> {
         }
         from = end;
     }
-    best
+    best.map(|(_, candidate)| candidate)
 }
 
 /// Case-insensitive ASCII substring search returning a byte index into
@@ -249,6 +254,19 @@ mod tests {
         std::fs::write(&prefix, png_data).unwrap();
         let text = format!("look at {}", edited_str);
         assert_eq!(detect_image_path(&text), Some(edited_str));
+
+        // With multiple distinct images, the first referenced one wins even if
+        // a later one has a longer path.
+        let a = temp_dir.path().join("a.png");
+        std::fs::write(&a, png_data).unwrap();
+        let longer = temp_dir.path().join("much-longer.png");
+        std::fs::write(&longer, png_data).unwrap();
+        let text = format!(
+            "compare {} with {}",
+            a.to_str().unwrap(),
+            longer.to_str().unwrap()
+        );
+        assert_eq!(detect_image_path(&text), Some(a.to_str().unwrap()));
     }
 
     #[test]
--- a/crates/goose-providers/src/images.rs
+++ b/crates/goose-providers/src/images.rs
@@ -44,7 +44,8 @@ pub fn convert_image(image: &ImageContent, image_format: &ImageFormat) -> Value 
 /// The extension must terminate the candidate (so `/tmp/foo.png.backup` is not
 /// truncated to `/tmp/foo.png`) and the leading `/` must follow a whitespace or
 /// quote boundary (so a `://` in a URL like `https://host/x.png` is not mistaken
-/// for an absolute path).
+/// for an absolute path). A path may be wrapped in matching quotes
+/// (`"/tmp/Screen Shot.png"`), in which case the closing quote terminates it.
 pub fn detect_image_path(text: &str) -> Option<&str> {
     const EXTENSIONS: [&str; 3] = [".png", ".jpg", ".jpeg"];
     const MAX_PATH_LEN: usize = 4096;
@@ -59,10 +60,9 @@ pub fn detect_image_path(text: &str) -> Option<&str> {
             break;
         };
 
-        let terminated = text
-            .get(end..)
-            .and_then(|rest| rest.chars().next())
-            .is_none_or(|c| c == '/' || c.is_whitespace());
+        let terminator = text.get(end..).and_then(|rest| rest.chars().next());
+        let terminated =
+            terminator.is_none_or(|c| c == '/' || c.is_whitespace() || c == '"' || c == '\'');
 
         if terminated {
             let mut floor = end.saturating_sub(MAX_PATH_LEN);
@@ -234,6 +234,17 @@ mod tests {
         let upper_str = upper.to_str().unwrap();
         let text = format!("see {}", upper_str);
         assert_eq!(detect_image_path(&text), Some(upper_str));
+
+        // Quoted path with spaces: the closing quote terminates the candidate.
+        let text = format!("describe \"{}\" please", png_path_str);
+        assert_eq!(detect_image_path(&text), Some(png_path_str));
+        let text = format!("describe '{}'", png_path_str);
+        assert_eq!(detect_image_path(&text), Some(png_path_str));
+
+        // A stray closing quote in prose must not act as a terminator for an
+        // unquoted path.
+        let text = format!("here {}\" trailing", png_path_str);
+        assert_eq!(detect_image_path(&text), Some(png_path_str));
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


jlmalone:
Addressed Codex's P2 in 1f6ff98 — replaced `suffix.to_lowercase().find(ext)` with an ASCII case-insensitive byte-level search (`find_ascii_ci`). Extensions are pure ASCII, so we no longer allocate a lowercased copy per iteration *and* the returned index is guaranteed to map directly into the original `text`, avoiding the `İ → i\u{307}` byte-length-shift hazard. Added tests for both the Turkish-İ prefix case and an uppercase `.PNG` filename.

jlmalone:
Addressed Codex's P2 in 6ebb907 — the loop now picks the earliest extension match across .png/.jpg/.jpeg at each step instead of exhausting one extension before trying the next, restoring "first path in the prompt wins" semantics. Added a test with an earlier .jpg followed by a later .png that asserts the .jpg is returned.

jlmalone:
Addressed Codex's P1 in 27a1118 — the extension match now has to terminate the filename. The byte right after `.png`/`.jpg`/`.jpeg` must not be `[a-zA-Z0-9_-]`, and a literal `.` is only accepted when followed by a non-alphanumeric. That keeps `see /tmp/foo.png.` (trailing sentence period) working while rejecting compound names like `/tmp/photo.png.backup`. Tests cover both behaviors.

jlmalone:
Addressed both Codex comments in 005be54:

- **P1 (URL extraction):** Added `extends_path_left`. The candidate's start position must sit at a real boundary — the byte immediately before must not be alphanumeric or any of `/ \\ . - _ ~ :`, and non-ASCII is treated as continuation. This blocks pulling `/home/alice/photo.png` out of `https://example.com/home/alice/photo.png`.
- **P2 (Unicode suffix):** Extended the right-side `continues_filename` rule to flag any non-ASCII byte after the extension (and after a trailing `.`). `/tmp/foto.pngé` no longer truncates to a sibling `/tmp/foto.png`.

Two tests added covering both cases.

jlmalone:
Addressed Codex's P1 in bc64948 — added `/` and `\\` to the continues_filename set so `/tmp/photo.png/thumbnail` no longer truncates to `/tmp/photo.png`. Test added.

jlmalone:
Addressed both Codex P2s in 61e0255 — switched both the post-extension and pre-candidate guards from byte-level to char-level classification.

The new continuation rule: alphanumeric (any script) / `_ - / \\` / `.` followed by alphanumeric → continues a filename; everything else (Unicode whitespace, punctuation, emoji) is a terminator. `extends_path_left` likewise uses `char_indices().next_back()` so the previous *char* (not byte) is classified, restoring `split_whitespace` semantics for NBSP / ideographic space / etc.

Tests added for ideographic-space prefix and NBSP suffix.

jlmalone:
Opened #9396 with a longer-term proposal for an atomic-paste span model (the `[Image 1]` UX you see in Claude Code / Cursor) that would eventually let us retire the heuristic in this PR altogether. Independent of #9387 landing — the heuristic still wants to be safe in the meantime.

jlmalone:
Addressed all three P1s in 1f4110f:

- **P1.1 (boundary controls/marks):** Replaced the deny-list with `is_path_separator`, an explicit allow-list of whitespace + the punctuation that visually separates words. Zero-width spaces, combining marks, bidi/format controls, emoji, and other symbols now glue rather than terminate, so payloads like `/Users/me/private.png\u{200B}.backup` and `https://example.com\u{200B}/Users/me/private.png` no longer extract sibling files. Tests added for ZWSP / combining mark / ZWSP-in-URL.
- **P1.2 (O(n²)):** Single forward byte scan (`match_image_extension`) replaces the per-extension re-scan, the backward walk is bounded to `MAX_PATH_LEN = 4096` (above Linux PATH_MAX), and `clean_path` now returns `Cow<'_, str>` so the common case borrows instead of allocating. Regression test runs detect_image_path against a 1 MB prompt and asserts sub-second.
- **P1.3 (TOCTOU):** Extracted `is_image_bytes(&[u8])` and reworked `load_image_file` to read the file exactly once and validate magic bytes from the same buffer it base64-encodes. No more separate check-and-read opens for a symlink to be swapped between. Added a non-image-with-.png-name test that exercises the buffer-validation path.

Tests for P2 partials added: path at byte 0, file:// URI false negative, backslash-escaped and quoted paths through `load_image_file`. Markdown/HTML embedding behavior is unchanged from upstream (the heuristic still matches src="/abs/path.png" if the file exists); that's an inherited property of the substring approach and is the motivation for #9396. Windows backslash-vs-escaped-space edge cases are out of scope for this PR but tracked.

jlmalone:
FYI — this finding was reviewed against `61e0255`, one commit behind. The boundary rewrite in `1f4110f` already replaced the deny-list with `is_path_separator` (whitespace + a small explicit punctuation allow-list); `@`, `~`, `+`, `%`, etc. are treated as continuation by default, so `/tmp/photo.png@2x` no longer truncates. Added explicit regression tests for `@2x`, `~`, and `+thumb` in 903602e to nail this down.

jlmalone:
Addressed in dfc3240. The previous code did iterate backwards (shortest-suffix first), but the `extends_path_left` guard already rejected any candidate whose left side wasn't a real path-separator — so a wrapped suffix like `/tmp/photo.png` inside `/workspace/x/tmp/photo.png` would be skipped because its prev char is alphanumeric. Same final result, but the algorithm's intent was not obvious from the loop direction.

Switched to forward iteration. The first valid candidate is now also the longest, so 'longest valid path wins' is encoded in the loop itself rather than relying on the boundary check to reject sub-suffixes. Added a test that builds a nested wrapper path and asserts the full path round-trips through detect_image_path.

jlmalone:
External re-review came back clean — declared ship-safe. Three follow-up notes; one was actionable and just fixed, two tracked separately:

- ✅ Issue #9396 comment had `see #4 below` which GitHub auto-linked to repo #4. Edited to `see item 4 below`.
- 🔁 `clean_path` only unescapes `\\ `, so drag-drop paths containing other shell-escaped metacharacters (`\\(`, `\\&`, etc.) silently miss. Filed as #9398.
- 🔁 The heuristic still attaches standalone local paths embedded in free-form user text. Tracked by #9396 (atomic-paste UX).

jlmalone:
Addressed the new round in 12b916f, plus filed follow-ups:

- **P1 (load TOCTOU + memory-DOS):** `load_image_file` now does `File::open` once, sniffs the first 8 bytes from that handle, then `read_to_end` from the same handle. A post-open symlink swap can't substitute the file (we're bound to the descriptor we validated), and a multi-GB non-image is rejected after 8 bytes instead of after a full read.
- **P2 perf bound:** Added `MAX_FS_PROBES = 256` to bound total `is_file`/`is_image_file` syscalls per call. A hostile prompt with thousands of `/.../foo.png` patterns now degrades to ~1 ms of stat work instead of seconds.
- **P2 conservative `is_path_separator`** (emoji / Unicode sentence punctuation glue when they probably shouldn't) — filed as #9399. The conservative rule shipped here is the right default for the security-vs-UX tradeoff; broadening it needs a curated set of code-point ranges that doesn't reopen the zero-width / combining-mark holes.
- **Semantic false positives in Markdown / HTML / code-fences** — already noted as a heuristic limitation, the structural fix is #9396.

jlmalone:
Addressed in e68367f — dropped the 1s wall-clock assertion from the huge-prompt test. The functional check (correct result on a 1 MB prefix) is kept; the structural bounds (MAX_PATH_LEN, MAX_FS_PROBES) are what keep the algorithm linear. No more flaky-on-slow-CI risk.

jlmalone:
Addressed in 23b8f5c. The per-match cap I added in 12b916f could still drop a valid path when one match's backward walk exhausted its budget. Removed the explicit syscall cap entirely; `MAX_PATH_LEN = 4096` already bounds the per-match backward walk, and since only `/`-prefixed candidates trigger `is_file` the syscall count is structurally bounded without a counter.

Added a regression test: 100 fake `.png` paths followed by the real one — the function must keep scanning past every failure and return the real path. Both the original global cap and the per-match cap variants would have failed this test.

jlmalone:
This is ready for a maintainer review when someone has a moment 🙏

- Fixes `detect_image_path` so local vision works with macOS screenshot paths that contain spaces, and hardens the heuristic against the edge cases Codex flagged (Unicode-aware boundaries, TOCTOU-safe filesystem probes, earliest-path-wins ordering).
- Now linked to #9576 (the originating bug). Follow-up hardening is tracked in #9398 and #9399, and atomic-paste UX in #9396.
- Rebased on the latest `main` (no conflicts), all CI green, every commit DCO-signed, and Codex's P2s are addressed.

cc @jamadeo — would appreciate a look when you have time. Thank you!

jlmalone:
Friendly bump 🙏 This PR has been green and has worked through every Codex round since early June — a single-file fix to `detect_image_path` (closes #9576): on macOS, screenshot paths contain spaces, which silently broke local vision. Just pushed one more perf hardening (`ff26ac5`) so a prompt full of non-path `*.png` tokens can't trigger a per-token backward scan.

Would really appreciate a maintainer review when there's a spare cycle — happy to squash the hardening commits if that helps the merge. cc @aaif-goose/goose-maintainers

jlmalone:
Rebased onto current main. The original change targeted `detect_image_path` in `crates/goose/src/providers/utils.rs`, but main has since extracted that function into the new `goose-providers` crate (`crates/goose-providers/src/images.rs`) and rewritten it to split on whitespace — which still drops absolute paths containing spaces (e.g. macOS screenshots). This version re-applies the fix there: it anchors on each image-extension occurrence and walks back over `/`-rooted starts (bounded to avoid quadratic scanning), returns the longest existing image file, and now matches extensions case-insensitively. Existing behavior (relative/fake/nonexistent paths) is preserved; added spaces + extension-flood regression tests.

DOsinga:
Thanks for this — the core fix (detecting absolute paths with spaces) is a real improvement. I pushed a follow-up commit to your branch (`d7912b8`) along with a merge of `main`:

- **Fixed two boundary regressions.** The simplified backward-scan that was re-applied after the move to `goose-providers` dropped the boundary guards from your earlier iterations, so a couple of the cases codex had flagged came back:
  - A path that is a suffix of a URL (e.g. `https://host/some/local/path.png`) was extracted via the `://` separator and would attach a local file if one happened to exist at that path.
  - A backup file like `/tmp/shot.png.backup` was truncated to `/tmp/shot.png`.

  The fix requires the extension to terminate the candidate (next char is `/`, whitespace, or end) and the leading `/` to follow a whitespace/quote boundary. Added regression tests for both.
- **`cargo fmt`** — the format check was failing in CI; it's clean now.

One thing to tidy up on your side: the **PR description is now stale** — it still describes a `clean_path` helper and quote/escape unescaping in `load_image_file`, but none of that is in the current diff. Could you update the description to match what the code actually does?

`cargo clippy -p goose-providers --all-targets` and the `images` tests pass locally. (There's an unrelated pre-existing `string_slice` clippy error in `json.rs` that's already on `main`.)

jlmalone:
Thanks @DOsinga — really appreciate you taking it the rest of the way. The two boundary regressions were exactly the cases my rebase onto `goose-providers` had flattened, and your termination + leading-`/` boundary guards are the right call; the longest-existing-candidate ordering is cleaner than what I had.

Done on my side: I've updated the PR description to match the current code — dropped the stale `clean_path` / `load_image_file` unescaping section (none of that survived the rebase) and documented the actual extension-anchored scan + boundary guards, with credit to you for the hardening and fmt. Re-ran locally: `cargo fmt` clean, `cargo clippy -p goose-providers --all-targets` clean, and all 5 `images` tests pass. Ready whenever you are.

DOsinga:
great. I'm going to merge this

FWIW we're probably moving away from this and let the model handle this direclty using a tool call. more agentic!

DOsinga (approved):

PR #9389, opened 2026-05-23
fix: serialize agent creation per session

### LINKED ISSUE #9031: Race condition while connecting to MCP servers & elicitation errors

Chris79OG:
## Summary

- add a per-session creation mutex in `AgentManager`
- keep the existing LRU sessions lock short-lived while serializing only same-session creation
- re-check the sessions cache after acquiring the per-session lock so concurrent `start_agent` / `resume_agent` callers reuse the first created agent

Fixes #9031.

## Verification

- `cargo fmt -- crates/goose/src/execution/manager.rs`
- `git diff --check`
- Attempted `cargo test -p goose execution::manager::tests::test_concurrent_session_creation_race_condition`, but this Windows host cannot link Rust build scripts because MSVC cannot find `msvcrt.lib` (`LNK1104`).

### DIFF
--- a/crates/goose/src/execution/manager.rs
+++ b/crates/goose/src/execution/manager.rs
@@ -10,7 +10,7 @@ use lru::LruCache;
 use std::collections::HashMap;
 use std::num::NonZeroUsize;
 use std::sync::Arc;
-use tokio::sync::{OnceCell, RwLock};
+use tokio::sync::{Mutex, OnceCell, RwLock};
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, info};
 
@@ -25,6 +25,7 @@ pub struct AgentManager {
     default_provider: Arc<RwLock<Option<Arc<dyn crate::providers::base::Provider>>>>,
     default_mode: GooseMode,
     cancel_tokens: Arc<RwLock<HashMap<String, CancellationToken>>>,
+    session_creation_locks: Arc<RwLock<HashMap<String, Arc<Mutex<()>>>>>,
 }
 
 impl AgentManager {
@@ -46,6 +47,7 @@ impl AgentManager {
             default_provider: Arc::new(RwLock::new(None)),
             default_mode,
             cancel_tokens: Arc::new(RwLock::new(HashMap::new())),
+            session_creation_locks: Arc::new(RwLock::new(HashMap::new())),
         };
 
         Ok(manager)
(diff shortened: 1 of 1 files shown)


jamadeo:
Thank you @Chris79OG ! Looks like this was fixed in #9357 which came in just slightly before.

PR #9403, opened 2026-05-24
fix: show error for unsupported image formats

### LINKED ISSUE #9359: Cannot attach HEIF format images
**Describe the bug**

When a HEIF image is attached, the preview keeps loading infinitely and image is not attached to chat. I also saw that while that preview is loading, I cannot open a new chat

---

**To Reproduce**
1. Start new chart
2. Attach any HEIF image
---

**Screenshots**

https://github.com/user-attachments/assets/56fb4493-33ed-4935-a338-3edbd1852db1

---

- **OS & Arch:** MacOS 26.4.1
- **Interface:** UI
- **Version:** 1.34.1
- **Provider & Model:** Ollama – gemma-4

---

**Additional context**

[diagnostics_20260522_3.zip](https://github.com/user-attachments/files/28145240/diagnostics_20260522_3.zip)

awaissaeed530:
## Summary
Display an error message when user uploads unsupported image format

### Related Issues
Fixes #9359

### DIFF
--- a/ui/desktop/src/components/ChatInput.tsx
+++ b/ui/desktop/src/components/ChatInput.tsx
@@ -1,46 +1,42 @@
-import { AppEvents } from '../constants/events';
-import React, { useRef, useState, useEffect, useMemo, useCallback } from 'react';
-import { ArrowUp, Bug, ScrollText } from 'lucide-react';
-import { Tooltip, TooltipContent, TooltipTrigger } from './ui/Tooltip';
-import { Button } from './ui/button';
-import type { View } from '../utils/navigationUtils';
-import Stop from './ui/Stop';
-import { Attach, Close, Microphone } from './icons';
-import { ChatState } from '../types/chatState';
 import debounce from 'lodash/debounce';
-import { LocalMessageStorage } from '../utils/localMessageStorage';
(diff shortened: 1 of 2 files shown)


awaissaeed530:
I have added this toast error for now

Screenshot 2026-05-30 at 14 09 32

DOsinga:
Thanks for taking a crack at this, @awaissaeed530! After reviewing, we're going to close this one.

The underlying problem in #9359 is that HEIC/HEIF can't be decoded by Chromium's `` element at all, so `compressImageDataUrl` rejects and — because that rejection isn't caught in the paste/attach handlers — the preview stays stuck in its loading state. Blocking `.heic`/`.heif` by extension hides the symptom for that specific case, but the underlying "spinner hangs forever on any undecodable image" bug remains, and it doesn't add the HEIC support that the issue is really asking for.

Properly fixing this means (a) wrapping the `compressImageDataUrl` call in a try/catch so *any* undecodable image fails gracefully, and (b) actually converting HEIC → JPEG/PNG (e.g. in the main process) before it reaches the canvas. That's a larger, separate piece of work, so we'd rather track it on the issue than carry a partial draft.

Really appreciate the contribution and hope you'll pick up another one!

PR #9415, opened 2026-05-25
fix(ci): build linux x86_64 standard inside manylinux_2_28 for glibc 2.28+ compat

### LINKED ISSUE #9414: Linux x86_64 install fails on glibc < 2.34 (Ubuntu 18.04 ESM / 20.04 / RHEL 8 / Debian 10)
### Upfront framing

Yes, the host is Zorin 15.3 (Ubuntu 18.04 base, glibc 2.28). That's an old base — I'm not asking goose to commit to indefinite support for ancient distros.

I'm filing this because it was my **literal first interaction with the product**. I followed the install command straight from the README:

```
curl -fsSL https://github.com/aaif-goose/goose/releases/download/stable/download_cli.sh | bash
```

The script ran to completion, reported success, then installed a binary that immediately fails to execute. No warning, no exit code, no breadcrumb. From a first-touch UX standpoint that's a real bug regardless of how old the host is. The root cause is also fixable at near-zero ongoing cost using infrastructure that's already standard in the Python ecosystem.

### What happens

The `goose configure` step that the install script invokes (and any later `goose` invocation) fails:

```
/home/me/.local/bin/goose: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' not found
/home/me/.local/bin/goose: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.30' not found
/home/me/.local/bin/goose: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `CXXABI_1.3.13' not found
/home/me/.local/bin/goose: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found
/home/me/.local/bin/goose: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found
/home/me/.local/bin/goose: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.30' not found
/home/me/.local/bin/goose: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found
/home/me/.local/bin/goose: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found
/home/me/.local/bin/goose: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found
```

goose install failure on glibc 2.28 (Zorin 15.3 / Ubuntu 18.04 base)

### Root cause

The `goose-x86_64-unknown-linux-gnu.tar.bz2` artifact is built on `ubuntu-22.04` runners (`.github/workflows/build-cli.yml:39`). The header comment at line 7 documents this is intentional:

> Linux standard: Uses native Ubuntu 22.04 runners to keep glibc compatibility with Ubuntu 22.04 LTS

Ubuntu 22.04 ships glibc 2.35 / libstdc++ 12. Rust dynamically links against whichever glibc is on the build host, so the resulting binary carries `GLIBC_2.34`, `GLIBCXX_3.4.30`, `CXXABI_1.3.13` as required minimums.

### Affected systems (still in support as of 2026)

| Distro | glibc | Status |
|---|---|---|
| Ubuntu 18.04 ESM | 2.27 | Extended security maintenance |
| Ubuntu 20.04 LTS | 2.31 | ESM (standard support ended Apr 2025) |
| Zorin OS 15.x | 2.28 | Active community use |
| RHEL 7 | 2.17 | ESM until 2028 |
| RHEL 8 | 2.28 | Maintenance until 2029 |
| Debian 10 | 2.28 | LTS / ELTS |
| Amazon Linux 2 | 2.26 | Maintenance until 2026-06 |

### Proposed fix

Build the `x86_64-unknown-linux-gnu` standard artifact inside the `quay.io/pypa/manylinux_2_28_x86_64` container instead of natively on `ubuntu-22.04`. This is exactly the approach `rust-lang/rust-analyzer` adopted in [PR #19586](https://github.com/rust-lang/rust-analyzer/pull/19586) after migrating away from `cargo-zigbuild` (zig doesn't support PGO, and for projects with C++ deps built via cmake — goose has `llama-cpp-2` — containerised builds are the more reliable approach per the [canonical reference](https://kobzol.github.io/rust/ci/2021/05/07/building-rust-binaries-in-ci-that-work-with-older-glibc.html)).

Outcome:
- Binary requires glibc ≥ 2.28 (covers every distro in the table above except RHEL 7 / Amazon Linux 2).
- **All default features retained**: `local-inference`, `system-keyring`, `aws-providers`, `telemetry`, `otel`, `rustls-tls`, `code-mode`. No artifact split, no feature drop.
- Other matrix entries (musl, vulkan, aarch64, macOS, Windows) unchanged.
- One additional CI tool (the manylinux image) — pinned by digest.

Locally validated: built `goose-cli` inside `quay.io/pypa/manylinux_2_28_x86_64` on a glibc-2.28 host and confirmed the resulting binary runs natively with all default features.

### Environment

- **OS & Arch:** Zorin OS 15.3 x86_64 (Ubuntu 18.04 base, glibc 2.28, libstdc++ 6.0.25)
- **Interface:** CLI (binary fails to load before any UI/server path is reachable)
- **Version:** v1.35.0 stable (artifact `goose-x86_64-unknown-linux-gnu.tar.bz2` from `download_cli.sh`)
- **Extensions enabled:** N/A — binary fails to load, `goose configure` never runs
- **Provider & Model:** N/A — same reason

88plug:
## Summary

> **Upfront:** the reporting host is Zorin 15.3 (Ubuntu 18.04 base, glibc 2.28). Yes, that's old. The reason this is worth fixing is that it was a **literal first-touch interaction** — `curl … | bash` from the README, exactly as documented — and the script ran to completion while installing a binary that won't load. That's a first-user-experience bug independent of host age.

Fresh install on Zorin 15:
```
curl -fsSL https://github.com/aaif-goose/goose/releases/download/stable/download_cli.sh | bash
```
The script reports success. The binary it installs cannot be loaded:
```
goose: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found
goose: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.30' not found
goose: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `CXXABI_1.3.13' not found
...
```

The standard linux artifact is built on `ubuntu-22.04` runners (`.github/workflows/build-cli.yml:39`), which pins its glibc floor at 2.34. Systems older than that — Ubuntu 18.04 ESM / 20.04, RHEL 8, Debian 10, Zorin 15 — get a broken install with no warning.

Fixes #9414.

## The fix

Build the `x86_64-unknown-linux-gnu` standard artifact inside `quay.io/pypa/manylinux_2_28_x86_64` (pinned by digest). All other matrix entries are unchanged.

Result:
- Binary now requires glibc ≥ 2.28 / libstdc++ shipping with AlmaLinux 8.
- **All default features retained**: `local-inference`, `system-keyring`, `aws-providers`, `telemetry`, `otel`, `rustls-tls`, `code-mode`. No artifact split, no feature drop, no musl fallback.
- Other matrix entries (musl, vulkan, aarch64, macOS, Windows) untouched.

The diff (`+50/-6` lines in `.github/workflows/build-cli.yml`):
1. `container:` field on the linux-x86_64-standard matrix entry (digest-pinned).
2. Job-level `container: ${{ matrix.container }}` — entries without a container value run on the host runner as before.
3. New "Install Linux build dependencies (manylinux container)" step using `dnf` (openssl-devel, dbus-devel, libxcb-devel, cmake, perl-core, clang, git, tar).
4. Existing "Install Linux build dependencies" step renamed to "(host runner)" and gated on `matrix.container == ''`.
5. New "Build CLI (manylinux container)" step that bootstraps rustup directly (hermit's tool cache is host-runner-scoped) and reads the channel from `rust-toolchain.toml` (with a non-empty guard).
6. Existing "Build CLI (Linux/macOS)" step gated on `matrix.container == ''`.
7. Cache key extended with `${{ matrix.container || 'native' }}` so container and host builds don't collide.
8. Package step conditionally skips `source ./bin/activate-hermit` when running inside the container.

## Why manylinux instead of cargo-zigbuild

`rust-analyzer` originally used `cargo-zigbuild` for old-glibc builds and migrated to `quay.io/pypa/manylinux_2_28_x86_64` in [rust-lang/rust-analyzer#19586](https://github.com/rust-lang/rust-analyzer/pull/19586). Two reasons applied directly here:

1. **C++ deps via cmake.** `goose` depends on `llama-cpp-2` (`crates/goose/Cargo.toml:196`) which builds `llama.cpp` via cmake. Kobzol's [canonical CI guide](https://kobzol.github.io/rust/ci/2021/05/07/building-rust-binaries-in-ci-that-work-with-older-glibc.html) specifically recommends containerised builds for projects with C++ deps and reports `cross`-style cross-compilation as unreliable for them.
2. **PGO future-compatibility.** zigbuild doesn't support PGO. If goose ever wants to enable PGO (rust-analyzer got a 15% perf win from it), we'd have to migrate off zigbuild anyway.

## Local validation

Built `goose-cli` inside `quay.io/pypa/manylinux_2_28_x86_64@sha256:441c35…` on this glibc-2.28 host:

| Check | Result |
|---|---|
| Build completes with default features (incl. `local-inference` via llama.cpp) | ✅ 282 MB binary |
| Required `GLIBC_*` symbols | All ≤ **2.28** |
| Required `GLIBCXX_*` / `CXXABI_*` symbols | All ≤ **3.4.25** / **1.3.11** |
| `goose --version` on glibc-2.28 host | `1.35.0` |
| `goose --help` shows `local-models` (proves `local-inference` linked in) | ✅ |
| New binary on Ubuntu 24.04 (glibc 2.39) with `libgomp1` installed | `1.35.0` |
| Current stable binary on the same Ubuntu 24.04 + `libgomp1` | `1.35.0` (identical behaviour — `libgomp1` is a pre-existing runtime dep, not introduced by this PR) |

The new binary is a **strict superset** of compatible systems: it runs everywhere the current binary does, plus glibc 2.28-2.33 systems that previously broke.

Build script and full log are reproducible from any glibc-2.28+ host with Docker.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(...)"` parses cleanly.
- [x] `actionlint` adds no new warnings (3 remaining `SC2086` info-lines are pre-existing on `$GITHUB_ENV` echoes, untouched).
- [x] Local manylinux container build on glibc-2.28 host: binary runs natively.
- [x] Cross-distro smoke: new binary + current stable binary both run on Ubuntu 22.04 / 24.04 / Debian 12 when `libgomp1` is present (identical behaviour).
- [ ] CI green on this PR via `.build-cli` PR command.
- [ ] Confirm musl / vulkan / aarch64 / macOS / Windows artifacts unchanged in build matrix output.

## What this doesn't fix

- Users on glibc < 2.28 (RHEL 7 / CentOS 7 / Amazon Linux 2) — they remain on the existing `GOOSE_LINUX_VARIANT=musl` path. A separate, smaller change to `download_cli.sh` could auto-route those users to the musl artifact; happy to send that as a follow-up if maintainers want it.
- aarch64 systems with old glibc — same situation. A manylinux-aarch64 container would address it but adds CI matrix surface; deferring unless asked.

### DIFF
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -4,7 +4,8 @@
 #  - canary.yml
 #
 # Platform Build Strategy:
-#  - Linux standard: Uses native Ubuntu 22.04 runners to keep glibc compatibility with Ubuntu 22.04 LTS
+#  - Linux x86_64 standard: Builds inside manylinux_2_28 container for glibc 2.28+ compat
+#  - Linux aarch64 standard: Uses native Ubuntu 22.04 ARM runners (glibc 2.35)
 #  - Linux Vulkan: Uses native Ubuntu 24.04 runners for newer Vulkan headers/tooling
 #  - Linux musl: Uses native Ubuntu 22.04 runners with reduced features for musl compatibility
 #  - macOS: Uses native macOS runners for each architecture
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -4,8 +4,7 @@
 #  - canary.yml
 #
 # Platform Build Strategy:
-#  - Linux x86_64 standard: Builds inside manylinux_2_28 container for glibc 2.28+ compat
-#  - Linux aarch64 standard: Uses native Ubuntu 22.04 ARM runners (glibc 2.35)
+#  - Linux standard (x86_64 + aarch64): Builds inside manylinux_2_28 container for glibc 2.28+ compat
 #  - Linux Vulkan: Uses native Ubuntu 24.04 runners for newer Vulkan headers/tooling
 #  - Linux musl: Uses native Ubuntu 22.04 runners with reduced features for musl compatibility
 #  - macOS: Uses native macOS runners for each architecture
@@ -46,6 +45,8 @@ jobs:
             architecture: aarch64
             target-suffix: unknown-linux-gnu
             build-on: ubuntu-22.04-arm
+            # Pinned by digest for reproducible builds; bump explicitly when newer manylinux_2_28 images ship.
+            container: quay.io/pypa/manylinux_2_28_aarch64@sha256:8b5f2b4e8c072ae5aefeb659f22c03e1ff46e6a82f154b6c904b106c87e65ff7
             variant: standard
           - platform: linux
             architecture: x86_64
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -27,6 +27,7 @@ jobs:
   build-cli:
     name: Build CLI
     runs-on: ${{ matrix.build-on }}
+    container: ${{ matrix.container }}
     env:
       MACOSX_DEPLOYMENT_TARGET: "12.0"
     strategy:
@@ -37,6 +38,8 @@ jobs:
             architecture: x86_64
             target-suffix: unknown-linux-gnu
             build-on: ubuntu-22.04
+            # Pinned by digest for reproducible builds; bump explicitly when newer manylinux_2_28 images ship.
+            container: quay.io/pypa/manylinux_2_28_x86_64@sha256:441c35fdc6ee809ff9260894f8468ab4fea8c15dc880f8700a3f81b7922c1cda
             variant: standard
           - platform: linux
             architecture: aarch64
@@ -97,8 +100,8 @@ jobs:
           sed -i.bak 's/^version = ".*"/version = "'${{ inputs.version }}'"/' Cargo.toml
           rm -f Cargo.toml.bak
 
-      - name: Install Linux build dependencies
-        if: matrix.platform == 'linux'
+      - name: Install Linux build dependencies (host runner)
+        if: matrix.platform == 'linux' && matrix.container == ''
         run: |
           sudo apt-get update
           sudo apt-get install -y \
@@ -119,11 +122,28 @@ jobs:
             sudo apt-get install -y musl-tools
           fi
 
+      - name: Install Linux build dependencies (manylinux container)
+        if: matrix.platform == 'linux' && matrix.container != ''
+        run: |
+          # perl-core provides FindBin, File::Compare, etc. that openssl-sys's
+          # vendored openssl build needs; in AlmaLinux 8 these aren't standalone packages.
+          # clang provides libclang.so for bindgen (used by llama-cpp-sys-2).
+          # Defensive: avoid actions/checkout falling back to a tarball download if base image changes.
+          dnf install -y --setopt=install_weak_deps=False \
+            openssl-devel \
+            dbus-devel \
+            libxcb-devel \
+            cmake \
+            perl-core \
+            clang \
+            git \
+            tar
+
       - name: Cache Cargo artifacts (Linux/macOS)
         if: matrix.platform != 'windows'
         uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
         with:
-          key: ${{ matrix.architecture }}-${{ matrix.target-suffix }}-${{ matrix.build-on }}-native-macos-deployment-target-12
+          key: ${{ matrix.architecture }}-${{ matrix.target-suffix }}-${{ matrix.build-on }}-${{ matrix.container || 'native' }}-macos-deployment-target-12
 
       - name: Cache Cargo artifacts (Windows)
         if: matrix.platform == 'windows'
@@ -131,8 +151,8 @@ jobs:
         with:
           key: windows-msvc-cli-${{ matrix.variant }}
 
-      - name: Build CLI (Linux/macOS)
-        if: matrix.platform != 'windows'
+      - name: Build CLI (Linux/macOS host runner)
+        if: matrix.platform != 'windows' && matrix.container == ''
         env:
           RUST_LOG: debug
           RUST_BACKTRACE: 1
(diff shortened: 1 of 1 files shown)


.github/workflows/build-cli.yml
alexhancock: @88plug can you update this comment?

```suggestion
#  - Linux x86_64 standard: Builds inside manylinux_2_28 container for glibc 2.28+ compat
#  - Linux aarch64 standard: Uses native Ubuntu 22.04 ARM runners (glibc 2.35)
```

alexhancock (commented):
cc @jamadeo and @jh-block to take a look as well.

looks good in general to me, just want to make sure we don't break anything for other linux distro installers

jamadeo:
Yeah this seems fine to me. As for the install script, if we detect an unsupported glibc version, we could download the `musl` build.

88plug:
Thanks for the quick look, @alexhancock and @jamadeo.

**@alexhancock — on the "other linux distros" concern:** this PR only modifies the `linux-x86_64-standard` matrix entry's build path. For the musl / vulkan / aarch64 / macOS / Windows entries:

- No `container:` field is added.
- The new install-deps and build steps are guarded by `if: matrix.container != ''` and never run for them.
- The existing install-deps and build steps are guarded by `if: matrix.container == ''` and run exactly as before.
- The only shared change is the build cache key, which now ends in `${{ matrix.container || 'native' }}` — for entries without a container that's literally `native`. One-time cache miss on the first build after merge, then steady-state cache hits resume.

So the behavioural diff for every non-x86_64-standard entry is: nothing, except the first build paying for a cache refresh.

**@jamadeo — agreed, that's the right safety net for the systems below the new floor** (RHEL 7 / Amazon Linux 2 / Alpine — and Alpine in particular is what got reported in #9072). I have the change ready locally on a separate branch (~22 lines in `download_cli.sh`: parse `ldd --version`, route to the musl artifact when glibc < 2.34, log a one-line warning that the musl path doesn't include `local-inference` or `system-keyring` so users aren't surprised by the feature delta). Per CONTRIBUTING.md ("don't open many PRs in quick succession") I'd rather send it as a small follow-up once this one lands — happy to open it whenever you give the nod.

jh-block:
I think we should do this for aarch64 too if there's an easy path -- it shouldn't expand the matrix if we just switch the existing aarch64 build to use older glibc as well? It seems better to have one glibc floor with two arches than different floors for each arch.

88plug:
Good call — bundled. Added a one-line `container:` field to the aarch64 standard entry (same `manylinux_2_28_aarch64` digest from the same `2026.05.22-1` release as the x86_64 pin), and collapsed the header comment back to one line since both arches now share the same floor.

Validated on our fork before pushing: ✅ in 21m06s on a free `ubuntu-22.04-arm` runner ([run](https://github.com/88plug/goose/actions/runs/26470864658)). Symbol floors mirror x86_64 exactly: GLIBC ≤ **2.28**, GLIBCXX ≤ **3.4.25**, CXXABI ≤ **1.3.11**. `goose --version` runs.

@alexhancock — the header comment goes from 2 lines back to 1 since the split-by-arch is no longer needed; same intent as your suggestion, just collapsed.

alexhancock (approved):

PR #9425, opened 2026-05-26
Add Cloudflare Workers AI as a native provider

jezweb:
## Add Cloudflare Workers AI as a native provider

This PR adds [Cloudflare Workers AI](https://developers.cloudflare.com/workers-ai/) as a built-in provider, joining the 30+ native LLM providers already in Goose.

### Why

Workers AI offers low-cost, low-latency inference for many open-source models (gpt-oss-20b/120b, Llama 4 Scout, Nemotron, Kimi K2.6, Gemma 4, GLM-4.7, Qwen3, Mistral) on Cloudflare's global edge network. It's a natural fit for teams already deployed on Cloudflare, and pricing is competitive (gpt-oss-20b at ~\$0.20/M tokens).

### What

A new \`cloudflare\` provider that hits Workers AI's OpenAI-compatible chat completion endpoint:

\`\`\`
https://api.cloudflare.com/client/v4/accounts/{ACCOUNT_ID}/ai/v1/chat/completions
\`\`\`

It implements the standard \`Provider\` + \`ProviderDef\` traits using the same OpenAI-compatible request/streaming helpers used by other native providers (e.g. NanoGPT, Avian), so tool calling, streaming, and prompt formats work out of the box.

### Configuration

Set three values via \`goose configure\` (or env vars):

| Key | Required | Notes |
|---|---|---|
| \`CLOUDFLARE_API_TOKEN\` | Yes (secret) | Workers AI API token from My Profile → API Tokens |
| \`CLOUDFLARE_ACCOUNT_ID\` | Yes | Account ID from the dashboard sidebar |
| \`CLOUDFLARE_HOST\` | No | Defaults to \`https://api.cloudflare.com\`, override for testing or alternate regions |

### Supported models

Curated \`KNOWN_MODELS\` list of validated tool-calling Workers AI models:

- \`@cf/openai/gpt-oss-20b\` (default — fast, strong on structured JSON)
- \`@cf/openai/gpt-oss-120b\`
- \`@cf/nvidia/nemotron-3-120b-a12b\`
- \`@cf/meta/llama-4-scout-17b-16e-instruct\`
- \`@cf/moonshotai/kimi-k2.6\`, \`@cf/moonshotai/kimi-k2.5\`
- \`@cf/qwen/qwen3-30b-a3b-fp8\`
- \`@cf/google/gemma-4-26b-a4b-it\`
- \`@cf/zai-org/glm-4.7-flash\`
- \`@cf/mistralai/mistral-small-3.1-24b-instruct\`

Users can pass any \`@cf/...\` model ID — \`fetch_supported_models\` hits Cloudflare's catalog API (\`/ai/models/search\`) and filters to \`task: \"Text Generation\"\` so the model picker shows the live list.

The canonical model registry already includes \`cloudflare-workers-ai\` entries shipped with Goose. This PR adds the \`cloudflare → cloudflare-workers-ai\` provider-name mapping so those entries resolve correctly when users pick Cloudflare models.

### Test coverage

Six unit tests in \`crates/goose/src/providers/cloudflare.rs\` using \`wiremock\`:

- \`test_metadata_shape\` — required/secret/primary flags on config keys
- \`test_known_models_use_cf_prefix\` — guards against typos in the model list
- \`test_build_client_trims_trailing_slash_and_whitespace\` — common configure-flow footgun
- \`test_chat_completion_url_path\` — verifies the request hits \`/client/v4/accounts/{id}/ai/v1/chat/completions\` with \`Bearer\` auth
- \`test_fetch_supported_models_filters_text_generation\` — model list endpoint filters non-text models
- \`test_fetch_supported_models_surfaces_api_errors\` — Cloudflare API errors propagate as \`ProviderError::RequestFailed\`

\`cargo check -p goose\` clean; \`cargo test --lib -p goose providers::cloudflare\` — 6 passed. Existing init/canonical tests continue to pass.

### Deferred (future PRs)

- **Embeddings.** Workers AI offers \`@cf/baai/bge-*\` embedding models; they need a separate call shape (not OpenAI-compatible) so left for a follow-up.
- **Image generation.** FLUX models on Workers AI use multipart requests and are out of scope for a chat provider.
- **AI Gateway routing.** Workers AI can also be reached via Cloudflare AI Gateway URLs. Users can already point \`CLOUDFLARE_HOST\` at a gateway hostname; no code change required, but worth documenting if there's interest.

---

Co-authored by Jezweb / Office Town (https://jezweb.com.au) — this provider also ships preconfigured in the upcoming \"Office Town Desktop\" Custom Distribution of Goose for AI agent fleets.

### DIFF
--- a/crates/goose/src/providers/canonical/name_builder.rs
+++ b/crates/goose/src/providers/canonical/name_builder.rs
@@ -50,6 +50,7 @@ pub fn map_provider_name(provider: &str) -> &str {
         "zhipu" => "zhipuai",
         "novita" => "novita-ai",
         "opencode_go" => "opencode-go",
+        "cloudflare" => "cloudflare-workers-ai",
         _ => provider,
     }
 }
--- a/crates/goose/src/providers/cloudflare.rs
+++ b/crates/goose/src/providers/cloudflare.rs
@@ -132,67 +132,103 @@ impl Provider for CloudflareProvider {
     /// Endpoint: `/client/v4/accounts/{account_id}/ai/models/search`. We filter to
     /// `task: "Text Generation"` models — embedding, image, audio, and classification
     /// models won't work through the chat completion path.
+    ///
+    /// Iterates pages using Cloudflare's `result_info.total_pages` metadata so
+    /// accounts with more than 100 matching models don't get a truncated list.
+    /// Capped at MAX_PAGES (50 = 5000 models) as a safety bound.
     async fn fetch_supported_models(&self) -> Result<Vec<String>, ProviderError> {
-        // The OpenAI-compat base URL ends in `/ai/v1`. Models live at `/ai/models/search`,
-        // one level up. Resolve relative to the base URL by going up a segment.
-        let response = self
-            .api_client
-            .request(None, "../models/search?per_page=100")
-            .response_get()
-            .await
-            .map_err(|e| {
+        const MAX_PAGES: u32 = 50;
+        let mut models: Vec<String> = Vec::new();
+        let mut page: u32 = 1;
+
+        loop {
+            // The OpenAI-compat base URL ends in `/ai/v1`. Models live at
+            // `/ai/models/search`, one level up. Resolve relative to the base URL
+            // by going up a segment.
+            let path = format!("../models/search?per_page=100&page={}", page);
+            let response = self
+                .api_client
+                .request(None, &path)
+                .response_get()
+                .await
+                .map_err(|e| {
+                    ProviderError::RequestFailed(format!(
+                        "Failed to fetch models from Cloudflare API (page {}): {}",
+                        page, e
+                    ))
+                })?;
+
+            let json: serde_json::Value = response.json().await.map_err(|e| {
                 ProviderError::RequestFailed(format!(
-                    "Failed to fetch models from Cloudflare API: {}",
-                    e
+                    "Failed to parse Cloudflare models response as JSON (page {}): {}",
+                    page, e
                 ))
             })?;
 
-        let json: serde_json::Value = response.json().await.map_err(|e| {
-            ProviderError::RequestFailed(format!(
-                "Failed to parse Cloudflare models response as JSON: {}",
-                e
-            ))
-        })?;
-
-        if let Some(errors) = json.get("errors").and_then(|v| v.as_array()) {
-            if !errors.is_empty() {
-                let msg = errors
-                    .iter()
-                    .filter_map(|e| e.get("message").and_then(|v| v.as_str()))
-                    .collect::<Vec<_>>()
-                    .join("; ");
-                return Err(ProviderError::RequestFailed(format!(
-                    "Cloudflare API returned errors: {}",
-                    msg
-                )));
+            if let Some(errors) = json.get("errors").and_then(|v| v.as_array()) {
+                if !errors.is_empty() {
+                    let msg = errors
+                        .iter()
+                        .filter_map(|e| e.get("message").and_then(|v| v.as_str()))
+                        .collect::<Vec<_>>()
+                        .join("; ");
+                    return Err(ProviderError::RequestFailed(format!(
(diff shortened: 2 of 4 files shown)


jezweb:
Thanks for the catch on the Codex review.

Pushed 653b1cdde to paginate the model catalog properly. `fetch_supported_models` now walks `result_info.total_pages`:

- per-page=100 + explicit `page=N` query param
- stops on reaching `total_pages` (defaults to 1 if absent — preserves single-page test behaviour)
- defensive early-exit on empty page
- `MAX_PAGES = 50` safety cap (5,000 models)
- final `sort()` + `dedup()` to guard against cross-page duplicates
- error messages now include the page number that failed

New test `test_fetch_supported_models_paginates_across_pages` mocks 3 pages with mixed task types and asserts the merged + filtered output. Local: `cargo test --lib -p goose providers::cloudflare` → 7 passed (was 6).

DOsinga:
Thanks for putting this together, and for being thorough with pagination and tests! 🙏

I'd like to suggest a different (and much smaller) approach before we take this further.

Cloudflare Workers AI exposes an OpenAI-compatible endpoint, and goose already has a **declarative provider** system precisely for this shape — no Rust needed. The whole of this PR (the new `CloudflareProvider` class, the hand-rolled `Provider` impl, the `name_builder` change, and the `init.rs`/`mod.rs` registration) collapses into a single ~30-line JSON file in `crates/goose/src/providers/declarative/`.

Everything Cloudflare needs is already supported:

- **Account ID in the URL via `${VAR}` templating** — see [`tanzu.json`](https://github.com/aaif-goose/goose/blob/main/crates/goose/src/providers/declarative/tanzu.json), which does `"base_url": "${TANZU_AI_ENDPOINT}/openai/v1/chat/completions"` with a required, non-secret `env_vars` entry. Cloudflare would use `https://api.cloudflare.com/client/v4/accounts/${CLOUDFLARE_ACCOUNT_ID}/ai/v1/chat/completions`.
- **`engine: "openai"`** gives you the full OpenAI-compatible streaming/chat-completions plumbing for free.
- **`dynamic_models: true`** makes `fetch_supported_models` hit the standard `/v1/models` endpoint, falling back to the static `models` list.
- **`catalog_provider_id: "cloudflare-workers-ai"`** reuses the existing canonical registry entry, so you don't need the `name_builder` mapping at all.

A sketch:

```json
{
  "name": "cloudflare",
  "engine": "openai",
  "display_name": "Cloudflare Workers AI",
  "description": "Run open models on Cloudflare's global network via Workers AI",
  "api_key_env": "CLOUDFLARE_API_TOKEN",
  "base_url": "https://api.cloudflare.com/client/v4/accounts/${CLOUDFLARE_ACCOUNT_ID}/ai/v1/chat/completions",
  "env_vars": [
    { "name": "CLOUDFLARE_ACCOUNT_ID", "required": true, "secret": false, "description": "Your Cloudflare account ID" }
  ],
  "catalog_provider_id": "cloudflare-workers-ai",
  "dynamic_models": true,
  "models": [ /* a few real default models as fallback */ ],
  "supports_streaming": true,
  "model_doc_link": "https://developers.cloudflare.com/workers-ai/models/"
}
```

The one genuinely Cloudflare-specific thing in this PR was the custom model listing against `/ai/models/search` (with the `task` filter). Cloudflare's OpenAI-compat endpoint also exposes `/v1/models`, so `dynamic_models: true` should cover it; if it doesn't, the static `models` list is a fine fallback. **If you do find you need extra functionality the declarative system can't express, please add it to the declarative framework itself** (so every provider benefits) rather than as a one-off custom provider class.

I'm going to close this in favour of that approach — please open a fresh PR with the declarative JSON config and we'll get it in. Note that goose isn't yet on Cloudflare's [supported integrations list](https://developers.cloudflare.com/workers-ai/), so it's worth a quick sanity check that the OpenAI-compat endpoint behaves as expected end-to-end. Thanks again!

PR #9431, opened 2026-05-27
Make compaction fit within bounded context

### LINKED ISSUE #8642: Context limit reached + compaction failure on sessions that haven't exceeded context
Describe the bug

When using Goose, I receive the following error mid-session:

Context limit reached. Compacting to continue conversation...
Ran into this error trying to compact: Failed to compact: context limit exceeded even after removing all tool responses.
Please try again or create a new session
The issue is that this occurs on sessions that have not actually exhausted the context window. The error also appears to affect brand new sessions.

Expected behavior
Compaction should succeed, or at minimum the context indicator should accurately reflect actual token usage before triggering compaction.

Actual behavior

Goose triggers auto-compaction, but the compaction call itself fails because there's no headroom left

The error fires even when the context indicator shows low usage (reported by other users as 9–14% in issue #3606)

Starting a new session does not always resolve the issue

Related issues

#3606 — Context window indicator inconsistency (UI shows low usage but error fires)

#4635 — Compaction no longer resets token count after update

#6588 — Manual compact does not update token count correctly

#4986 — Context length exceeded while context is much smaller than max

Environment

Goose version: [your version here]

OS: [your OS here]

Model/provider: [e.g. Claude, GPT-4o]

Additional context
The root issue appears to be a token counter bug or miscalculation — Goose is falsely reporting the context as full, causing compaction to fire prematurely. When compaction fires too late (or at the wrong threshold), the summarization call also exceeds the limit, creating an unrecoverable state within the session.

LanceTMason:
## Summary
- Clamps auto-compaction to preserve at least 10% context headroom even when `GOOSE_AUTO_COMPACT_THRESHOLD` is configured higher.
- Bounds the conversation history serialized into the compaction prompt, keeping recent context first and progressively shrinking non-tool history after context-length failures.
- Adds focused tests for bounded compaction input, recent-message retention, and the auto-compaction headroom clamp.

Fixes #8642.

## Test plan
- `RUSTUP_HOME="$PWD/.rustup-local" CARGO_HOME="$PWD/.cargo-local" PATH="$PWD/.cargo-local/bin:$PATH" cargo fmt`
- `RUSTUP_HOME="$PWD/.rustup-local" CARGO_HOME="$PWD/.cargo-local" PATH="$PWD/.cargo-local/bin:$PATH" cargo test -p goose context_mgmt::tests`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/crates/goose/src/context_mgmt/mod.rs
+++ b/crates/goose/src/context_mgmt/mod.rs
@@ -334,7 +334,7 @@ fn truncate_to_token_budget(
             mid -= 1;
         }
 
-        let candidate = &text[..mid];
+        let candidate = text.get(..mid).unwrap_or_default();
         if token_counter.count_tokens(candidate) <= content_budget {
             best = mid;
             low = mid.saturating_add(1);
@@ -348,7 +348,7 @@ fn truncate_to_token_budget(
     if best == 0 {
         String::new()
     } else {
-        format!("{}{}", &text[..best], notice)
+        format!("{}{}", text.get(..best).unwrap_or_default(), notice)
     }
 }
 
@@ -860,12 +860,12 @@ mod tests {
 
         let mut messages = Vec::new();
         for i in 0..20 {
-            messages.push(Message::user().with_text(&format!(
+            messages.push(Message::user().with_text(format!(
                 "large non-tool user message {} {}",
                 i,
                 "word ".repeat(1000)
             )));
-            messages.push(Message::assistant().with_text(&format!(
+            messages.push(Message::assistant().with_text(format!(
                 "large non-tool assistant message {} {}",
                 i,
                 "reply ".repeat(1000)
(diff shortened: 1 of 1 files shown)


DOsinga:
Thanks for digging into this and for filing the linked issue. After reviewing the change, we're going to close this for now.

The core concern is that this is a mitigation built on speculation rather than a fix with a clear root cause. The linked issue #8642 itself speculates that the underlying problem is a token-counter miscalculation, and this PR works around that by bounding/truncating the compaction prompt rather than addressing what's actually going wrong. We'd want to understand the real failure before layering on retry-with-shrinking-budget logic in core compaction.

There are also two codex review comments that haven't really been addressed: the multibyte-truncation comment was only superficially touched (the panic was avoided but the loop-progress concern wasn't explained), and the separator-token-budget comment is unaddressed. Per our contribution guide we expect codex comments to be resolved or answered: https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#ai-code-reviews

Please feel free to reopen (or open a fresh PR) if you can pin down what's actually being fixed here — ideally with a reproduction showing the real cause — and address the review comments. Appreciate the effort!

PR #9437, opened 2026-05-27
feat: propagate elicitation decline and cancel actions end-to-end

### LINKED ISSUE #9024: Elicitation: support decline and cancel actions end-to-end
## Summary

Goose's elicitation flow currently only supports the `accept` action defined in the [MCP elicitation spec](https://modelcontextprotocol.io/specification/2025-03-26/client/elicitation). The `decline` and `cancel` actions are not representable end-to-end, so MCP servers can't distinguish a user who explicitly rejected a request from one who submitted empty data, and CLI/UI "no" paths only cancel locally without informing the server.

## Current gaps

- **Wire format** — `ActionRequiredData::ElicitationResponse` ([crates/goose/src/conversation/message.rs](https://github.com/aaif-goose/goose/blob/main/crates/goose/src/conversation/message.rs#L163-L166)) carries only `{ id, user_data }`; there is no `action` field.
- **MCP client** — [`mcp_client.rs:389-394`](https://github.com/aaif-goose/goose/blob/main/crates/goose/src/agents/mcp_client.rs#L389-L394) unconditionally maps every response to `ElicitationAction::Accept`.
- **ActionRequiredManager** — `submit_response` / `request_and_wait` only carry `Value` user data, with no action variant.
- **Desktop UI** — [`ElicitationRequest.tsx:169-173`](https://github.com/aaif-goose/goose/blob/main/ui/desktop/src/components/ElicitationRequest.tsx#L169-L173) only renders an Accept button on schema-less elicitations; there is no Decline or Cancel button anywhere.
- **CLI** — [`elicitation.rs:26-29`](https://github.com/aaif-goose/goose/blob/main/crates/goose-cli/src/session/elicitation.rs#L26-L29) returns `Ok(None)` on a "no" answer, which cancels locally but never sends a decline/cancel action back to the MCP server.

## Proposed work

1. Extend `ActionRequiredData::ElicitationResponse` with an `action: ElicitationAction` field (or split into `Accept { user_data } | Decline | Cancel` variants).
2. Update `ActionRequiredManager` to carry action through the oneshot.
3. Update `mcp_client.rs` to emit the user's actual action instead of hardcoding `Accept`.
4. Add Decline / Cancel buttons to the desktop `ElicitationRequest` component (for both schema-ful and schema-less flows).
5. Update the CLI to send `decline`/`cancel` over the wire when the user rejects.
6. Regenerate the OpenAPI spec and ui/desktop API client.
7. Tests covering all three actions.

## Context

Surfaced during review of #8999, which scoped to FastMCP-related bugs (no-schema rendering, required-enum defaults, SSE-collision crash) and intentionally deferred decline/cancel as a separate spec-compliance effort.

mvanhorn:
## Summary

Propagates elicitation `decline` and `cancel` actions end-to-end through the agent stack, so user-driven declines/cancellations during MCP elicitation surface to the calling tool instead of getting swallowed at the boundary.

## Why this matters

Issue #9024 reported that when a user declined or cancelled an elicitation prompt from a tool, the CLI handled the UI gracefully but the agent kept thinking the elicitation was still pending. The MCP client never received the negative response, so the calling tool either timed out or hung. This PR threads the negative actions through every layer that previously dropped them.

## Changes

- `crates/goose-cli/src/session/elicitation.rs`: emit `Decline` / `Cancel` action variants from the CLI prompt loop
- `crates/goose-cli/src/session/mod.rs` + `export.rs`: thread the action through session state and export paths
- `crates/goose/src/action_required_manager.rs`: store the decline/cancel verdict against the pending elicitation id
- `crates/goose/src/agents/agent.rs`: surface the verdict on the next tool poll
- `crates/goose/src/agents/mcp_client.rs`: forward the verdict to the MCP client as an `ElicitationResult::Decline` / `::Cancel` payload
- `crates/goose/src/conversation/message.rs`: serialize the verdict into the conversation transcript

## Testing

Cargo check passes for `goose` and `goose-cli` crates. `cargo fmt --check` clean. Manual elicitation flow tested in `goose-cli`.

Fixes #9024


mvanhorn:
Regenerated the TypeScript schema in f08dbb3 - added `action?: string` to the ElicitationResponse type in `types.gen.ts` so it matches `openapi.json` (which already had the field from the Rust side). That should clear the Check Generated Schemas job.

DOsinga (approved):
Approving. The change cleanly threads the elicitation `ElicitationAction` (accept/decline/cancel) end-to-end from the CLI prompt through `ActionRequiredManager` and into the MCP client, fixing the case where declined/cancelled elicitations previously left the calling tool hanging (#9024).

The codex feedback about regenerating the OpenAPI artifacts has been addressed, and the added tests are meaningful (the manager round-trip and the serde backward-compat default).

I merged latest `main` into the branch and resolved the resulting conflict — `main` moved the conversation models into the new `goose-providers` crate and added an ACP elicitation respond handler (#9560), so I updated that handler's two call sites to pass the new `action` argument (defaulting to `Accept`, preserving existing ACP behavior). Build, clippy, fmt, the two new unit tests, and the OpenAPI/TS generation all pass.

Note: extending the ACP path and the desktop UI to surface Decline/Cancel buttons (item #4 in #9024) is out of scope here and can be a follow-up.

mvanhorn:
Thanks @DOsinga for merging the elicitation fix. Propagating decline and cancel end-to-end means user choices actually reach the agent stack.

mvanhorn:
Appreciate the merge, @DOsinga. Propagating elicitation decline and cancel end-to-end means those user actions reach the calling tool now instead of being swallowed.

PR #9439, opened 2026-05-27
fix: preserve unparseable extension entries during config refresh

### LINKED ISSUE #8970: Silent Rollback of Manually Edited config.yaml on Restart Breaks Custom MCP Extensions
**Describe the bug**
windows 10 22H2 , python 3.10.11 ,nodejs 22.16.0, goose 1.33.1
Bug Description: The UI state is seriously out of sync with config.yaml. Manually edited changes in the YAML are silently rolled back on restart, preventing custom MCP extensions from persisting.
Environment: The config directory has been physically redirected to the D: drive using mklink, and a custom LMStudio provider is in use.
Core Pain Point: The configuration self‑check mechanism on software startup is too aggressive, directly wiping out any non‑built‑in extensions.

---

**To Reproduce**

---

**Expected behavior**

---

**Screenshots**

---

- **OS & Arch:** [e.g. Ubuntu 22.04 x86]
- **Interface:** [UI / CLI]
- **Version:** [e.g. v1.0.2]
- **Provider & Model:** [e.g. Google – gemini-1.5-pro]

---

**Additional context**

mvanhorn:
## Summary

Preserves unparseable extension entries during config refresh so a single malformed entry no longer wipes its siblings from the on-disk config when a user enables/disables/removes a different extension.

## Why this matters

Issue #8970 reported that toggling one extension via the UI silently dropped a different, unrelated extension entry whose YAML was malformed. The old `set_param(EXTENSIONS_CONFIG_KEY, &extensions)` serialized the parsed map back to disk, which by construction excluded any entry that failed `serde_yaml::from_value::<ExtensionEntry>`. Result: the user lost extension config they did not even touch in that session.

## Changes

- `parse_extensions_map` extracted from `get_extensions_map_with_config`; it logs the offending key as `info!` (not `warn!`) and skips the entry instead of dropping it from the raw mapping
- New `with_raw_extensions_mapping` wrapper performs mutations against the raw `serde_yaml::Mapping`, then writes only the keys it actually touched -- so untouched-but-unparseable siblings round-trip unchanged
- `save_extensions_map` replaced by callers that go through the wrapper
- Six new tests covering: clean-sibling preservation, unparseable-sibling preservation across set_extension_enabled, set_extension add, remove_extension, single-log-per-malformed-key, and existing platform-filter behavior

## Testing

`cargo test -p goose --lib config::extensions::` -- 6/6 pass.

Fixes #8970


michaelneale (approved):
this looks good, thanks! I updated it, should merge in when green. this could be annoying I agree. seems well tested.

mvanhorn:
Much appreciated, @michaelneale. Preserving unparseable extension entries through config refresh prevents silent data loss.

PR #9447, opened 2026-05-27
fix(providers): tolerate missing Responses output

### LINKED ISSUE #9446: chatgpt_codex gpt-5.5 stream decode error when response.completed omits output
## Bug Description

When using the `chatgpt_codex` provider with `gpt-5.5`, Goose can abort a successful streamed response with a parser error because the terminal `response.completed` event omits `response.output`.

The stream appears otherwise valid: earlier chunks contain output data, and the final `response.completed` event has `status: "completed"` plus usage metadata. However Goose treats the missing `output` field as a fatal deserialization error.

## Error

```text
Request failed: Stream decode error: Failed to parse Responses stream event: missing field `output`
```

## Environment

- Goose version: `1.35.0`
- Provider: `chatgpt_codex`
- Model: `gpt-5.5`
- Interface: Desktop app / server logs
- OS observed: Windows; also reproduced/observed with a separate Goose 1.35.0 install on Linux/Zorin

## Representative failing event

The failing packet shape in the server log is:

```json
{
  "type": "response.completed",
  "response": {
    "id": "...",
    "status": "completed",
    "model": "gpt-5.5",
    "usage": {
      "input_tokens": 18406,
      "output_tokens": 7,
      "total_tokens": 18413
    }
  }
}
```

Important detail: `response.output` is absent, not empty.

In local logs I found multiple identical failures for `gpt-5.5`, all with the same missing `output` field on `response.completed`.

## Expected behavior

Goose should tolerate `response.completed.response.output` being absent and continue using the output accumulated from prior streamed events, or at minimum treat absent output as `[]` for the terminal metadata event.

## Actual behavior

Goose aborts the whole response with a stream decode error before the session can use the already-streamed content.

## Likely cause

In `crates/goose/src/providers/formats/openai_responses.rs`, `ResponseMetadata.output` appears to be required:

```rust
pub output: Vec<ResponseOutputItemInfo>,
```

Because it is not optional/defaulted, serde rejects a `response.completed` event that does not include `output`.

## Suggested fix

Make the terminal response metadata tolerant of omitted output, e.g.:

```rust
#[serde(default)]
pub output: Vec<ResponseOutputItemInfo>,
```

or otherwise make the field optional/defaulted wherever the `response.completed.response` object is deserialized.

A regression test could cover a `response.completed` event with no `output` field and assert that the parser does not fail.

## Related prior issues

This looks like the same class of strict Responses-stream parsing issue as earlier ChatGPT Codex keepalive failures, but with a different event shape:

- #7374
- #7464

Those were about unknown `keepalive` events; this issue is specifically about a valid/terminal `response.completed` event missing `response.output`.

serejaris:
## Summary

Fixes #9446.

`chatgpt_codex` / Responses streaming can receive a terminal `response.completed` event that includes usage metadata but omits `response.output`. Goose currently treats that as a fatal serde error before it can use text already accumulated from earlier stream events.

This makes `ResponseMetadata.output` default to an empty list when omitted and adds a regression that streams text first, then completes with no `response.output` field.

### Testing

- `source bin/activate-hermit && cargo test -p goose test_responses_completed -- --nocapture`
  - `31 passed; 0 failed` for `providers::formats::openai_responses::tests`
- `source bin/activate-hermit && cargo fmt --check`

Observed existing unrelated warning during tests: `method clear_env is never used` in `crates/goose/tests/providers.rs`.

### Related Issues

Fixes #9446
Discussion: n/a

Before: n/a

After: n/a

### DIFF
--- a/crates/goose/src/providers/formats/openai_responses.rs
+++ b/crates/goose/src/providers/formats/openai_responses.rs
@@ -273,6 +273,7 @@ pub struct ResponseMetadata {
     pub created_at: i64,
     pub status: String,
     pub model: String,
+    #[serde(default)]
     pub output: Vec<ResponseOutputItemInfo>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub usage: Option<ResponseUsage>,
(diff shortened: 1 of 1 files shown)


serejaris:
Small CI note: the red `changes` job failed before running the path filters because GitHub returned `diff temporarily unavailable due to heavy server load` from the PR files API. I tried `gh run rerun 26510596648 --failed`, but this actor cannot rerun that workflow.

I also ran one extra local check after opening the PR:

```
source bin/activate-hermit && cargo clippy -p goose --lib -- -D warnings
```

That passed. A broader `cargo clippy -p goose --all-targets -- -D warnings` currently fails on the pre-existing `clear_env` dead-code warning in `crates/goose/tests/providers.rs`, which is unrelated to this patch and also appears as a warning during the targeted test run.

jamadeo:
Thank you for the fix! Fixed in #9449 (which did come in slightly after -- sorry about that)

PR #9452, opened 2026-05-27
fix(local-inference): use media marker parts for vision prompts

jh-block:
## Summary
- Convert local multimodal placeholder text into llama.cpp `media_marker` content parts before applying chat templates.
- Preserve marker ordering while removing separator newlines adjacent to media markers so rendered prompts match llama.cpp expectations more closely.
- Add focused tests for string and content-part marker conversion.

### Testing
- `cargo test -p goose --features local-inference local_inference::tests`

### Related Issues
None.

N/A


DOsinga (approved):
Approving. Verified against the vendored llama.cpp source (`common/chat.cpp`): llama.cpp explicitly recognises the `media_marker` content-part type and deliberately omits newlines immediately before/after a media marker when concatenating typed content. This PR makes goose's rendered vision prompt match that behaviour by splitting the placeholder text into proper `media_marker` parts and stripping the adjacent separator newlines, which aligns the image embeddings with the marker position.

Code is clean (no restating comments, no over-defensive code, no double-prefixed errors), the change is correctly gated to only run when media is present, and the added tests exercise the real splitting/ordering/newline behaviour rather than trivial implementation details. CI is green and the commit is DCO-signed. LGTM.

DOsinga (approved):

PR #9457, opened 2026-05-28
feat(observability): trace provider/model at every reply hand-off

matt2e:
## Summary

Add structured tracing across the reply pipeline so we can correlate which provider/model is requested vs. actually used on every message, end-to-end per session.

- Always populate `Message.metadata.inference` in `reply_internal` so each emitted message carries provider + requested model (with `resolved_model: None` now a legitimate value when `fetch_model_info()` can't canonicalize it).
- Add `gen_ai.provider` and `gen_ai.request.model` fields to the existing `tracing::instrument` on `stream_response_from_provider` (the wrapper around `Provider::stream`; instrumenting the trait method itself isn't possible).
- Add `tracing::info!` at the four reply hand-off points, all tagged with `session.id`:
  - `/reply` entry after the session is loaded (`goose-server/src/routes/reply.rs`)
  - `restore_provider_from_session`, with `provider_source` / `model_source` = `session` | `config_default`
  - `update_provider`, with `source = "explicit_update"`
  - Just before `provider.stream(...)` in `reply_parts.rs`
  - `MessageEvent::Message` emit, with the message's inference metadata
- ACP path (`GooseAcpAgent::on_prompt`) gets symmetric entry + emit logs. The actual LLM call site is shared via `Agent::reply`, so the agent-level changes cover both `/reply` and ACP.

Together these allow tracing one log chain per session and pinpointing where the requested model and the actually-used model diverge.

## Test plan

- [x] `cargo build -p goose -p goose-server`
- [x] `cargo clippy --all-targets -p goose -p goose-server -- -D warnings`
- [x] `cargo test -p goose --lib conversation::`
- [ ] Manual: trigger a reply, confirm `session.id`, `provider`, `requested_model`, and `resolved_model` appear in logs at each hand-off

### DIFF
--- a/crates/goose-server/src/routes/reply.rs
+++ b/crates/goose-server/src/routes/reply.rs
@@ -283,6 +283,13 @@ pub async fn reply(
             }
         };
 
+        tracing::info!(
+            session.id = %session_id,
+            session.provider = ?session.provider_name,
+            session.model = ?session.model_config.as_ref().map(|m| &m.model_name),
+            "reply: incoming request",
+        );
+
         let session_config = SessionConfig {
             id: session_id.clone(),
             schedule_id: session.schedule_id.clone(),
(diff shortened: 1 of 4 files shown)


DOsinga:
Hey @matt2e — triaging open PRs. This one's still a draft and has picked up merge conflicts with main over the last few weeks. The change itself looks good (nice end-to-end provider/model tracing). Do you still want to land it? If so, a rebase + marking it ready for review would get it moving. Otherwise we can close it out for now. Snoozing for a few days.

PR #9462, opened 2026-05-28
fix(agent): ensure tool request timestamp precedes tool response timestamp

### LINKED ISSUE #9461: Tool request messages get later timestamps than tool responses, breaking Claude API message ordering
**Describe the bug**

```
```

---

**To Reproduce**

```bash
python3 -c "
import sqlite3, json
for r in rows: print(r)
"
```

---

**Expected behavior**

---

**Screenshots**

```
```

---

---

**Additional context**

```rust
}
```

msteve123:
## Summary

- Tool response placeholders are created in `agent.rs` before tools execute, so they capture an earlier `Utc::now()` timestamp than the tool request message built afterward
- The session DB retrieves messages with `ORDER BY created_timestamp, id` — when timestamps differ the `id` tiebreaker is irrelevant, so responses sorted before requests
- The Claude API then rejects the conversation with `400 unexpected tool_use_id in tool_result blocks`
- Fix: after building the request message, cap its timestamp to the response timestamp when it is greater; equal timestamps rely on the DB `id` tiebreaker (request is inserted first, so it always sorts first)

Fixes #9461

## Test plan

- [ ] Run a goose session that invokes at least one tool and verify no `400` errors from the Claude API
- [ ] Inspect the session DB to confirm no `user` rows have a lower `created_timestamp` than the preceding `assistant` row for the same tool pair

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -1982,10 +1982,18 @@ impl Agent {
                                                 request.metadata.as_ref(),
                                                 request.tool_meta.clone(),
                                             );
-                                        messages_to_add.push(request_msg);
                                         let final_response = request_to_response_map
                                             .remove(&request.id)
                                             .unwrap_or_else(|| Message::user().with_generated_id());
+                                        // Ensure the tool request timestamp precedes the tool response
+                                        // timestamp. The response placeholder is created before tools
+                                        // execute, so its timestamp can be earlier than the request
+                                        // message built here — which causes the DB to return them out
+                                        // of order and triggers a Claude API 400 error.
+                                        if request_msg.created > final_response.created {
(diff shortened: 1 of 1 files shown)


DOsinga (approved):
Thanks for the careful fix, @msteve123! 🙏 This is exactly the kind of subtle, well-diagnosed bug fix we love to see — you traced the seconds-granularity timestamp collision through the DB `ORDER BY created_timestamp, id` all the way to the Claude `400 unexpected tool_use_id` error, and the clamp-to-response approach (letting the `id` tiebreaker take over) is clean and minimal. Nicely done.

By the way, are you on our Discord (https://discord.gg/goose-oss)? We can always use solid fixes like this, and it'd be great to have you around the community. Approving and merging now. 🚀

PR #9464, opened 2026-05-28
[RFC] feat(goose-embed): embedding SDK for the agent loop

vincenzopalazzo:
## Summary

- New crate `crates/goose-embed/` exposes a builder API to construct and drive goose's agent loop from a plain Rust program -- no `goose` binary subprocess, no ACP over stdio
- `Goose::builder()` wraps `Agent::with_config` directly so the global `Config` is not touched. `SessionManager` / `PermissionManager` singletons are still used internally for now
- `PermissionDecider` async trait for headless tool-confirmation decisions, with `AutoApprove` and `DenyAll` built-ins
- `Recipe` first-class via the prelude -- `.recipe(r)` applies extensions, response schema, and (if not set explicitly) provider/model from `settings`
- Two examples (`hello_agent`, `with_recipe`) and one smoke test on builder validation

## Why

External Rust programs that want to embed goose today either spawn `goose acp` as a subprocess (works, but a lot of ceremony for a Rust-to-Rust call) or depend on `crates/goose` and reach into it (no stable surface, singletons everywhere). This RFC adds a thin facade so you can `cargo add goose-embed`, build a `Goose`, and drive the loop.

This is the **PoC**. I am not minimizing deps in this PR -- the crate inherits the full `goose` dep tree. The actual dep work lands in a follow-up `goose-core` carve-out. I want feedback on the surface before doing that so it has a contract to lift to. See `docs/brainstorms/2026-05-28-goose-embed-rust-sdk.md` for the longer story.

## Caveats

- Elicitations are not handled. If the agent requests structured input via MCP, it stalls with a warning. Richer decider API to come.
- No ACP server export. If you want to re-expose your embedded agent over ACP, use `goose acp` directly. A `goose-acp-server` bridge could follow once `goose-core` exists.
- Rust only. Other languages stay on ACP via the existing `goose-sdk` crates.

## Questions for reviewers

- Name -- `goose-embed` vs `goose-runtime` vs `goose-rs-sdk`. IDK, no strong preference.
- `PermissionDecider` is a `Send + Sync` async trait that hides the in-stream confirmation routing. Shape look right, or would you prefer a callback with `oneshot` channels?

## Test plan

- [x] `cargo build --workspace`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test -p goose-embed`
- [x] Remote CI green
- [ ] Manual: `GOOSE_EMBED_PROVIDER=anthropic GOOSE_EMBED_MODEL=claude-sonnet-4 cargo run -p goose-embed --example hello_agent`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4640,6 +4640,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "goose-embed"
+version = "1.36.0"
+dependencies = [
+ "anyhow",
+ "async-stream",
+ "async-trait",
+ "futures",
+ "goose",
+ "serde_yaml",
+ "tokio",
+ "tracing",
+]
+
+[[package]]
 name = "goose-mcp"
 version = "1.36.0"
 dependencies = [
--- a/crates/goose-embed/Cargo.toml
+++ b/crates/goose-embed/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "goose-embed"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+authors.workspace = true
+license.workspace = true
+repository.workspace = true
+description = "Embeddable Rust SDK for the goose AI agent. Build your own Rust program that runs goose's agent loop with custom providers, extensions, recipes, and skills."
+
+[lints]
+workspace = true
+
+[dependencies]
+goose = { path = "../goose" }
+anyhow = { workspace = true }
+async-stream = { workspace = true }
+async-trait = { workspace = true }
+futures = { workspace = true }
+tokio = { workspace = true }
+tracing = { workspace = true }
+serde_yaml = { workspace = true }
+
+[dev-dependencies]
+tokio = { workspace = true }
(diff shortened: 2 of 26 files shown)


DOsinga:
Thanks for putting this together, Vincenzo — the goal of being able to embed and drive the agent loop in-process is one we very much share, and it's clearly thoughtful work.

The thing is, we already have a concrete plan for how we want to expose the agent loop as an API, and it's a different shape than the `goose-core` trait carve-out this PR builds toward. We're unrolling the agent loop into a re-entrant, turn-at-a-time state machine where the conversation *is* the state and each turn is a plain function call — that's what gives us a clean, composable, embeddable surface without a parallel trait layer over the existing internals. The full write-up is here: https://github.com/aaif-goose/goose/discussions/9944

That unrolling is how we plan to offer goose up as an API, so rather than land this speculative scaffolding I'd love for you to get active over in that discussion (or in Discord) — your perspective on the embedding use case would be really valuable in shaping it. Going to close this for now, but please dive into #9944.

PR #9465, opened 2026-05-28
Bench marking

DOsinga:
## Summary

update the benchmarking with some subcommands and try it on some combinations


alexhancock (approved):
I would look at this one to verify if there is an issue

https://github.com/aaif-goose/goose/pull/9465/changes#r3343111441

PR #9471, opened 2026-05-28
Support MCP extensions in open plugins

jamadeo:
- discover MCP extensions in [open plugins](https://open-plugins.com/)
- add entries for discovered plugins to config.yaml
- allow disabling discovered configs in config.yaml


jamadeo:
There's a UX question around how plugins should be enabled/disabled. Ideally we'd have this work similar to how we do extensions, where you have global defaults plus session-level options. But unlike extensions, you have filesystem discovery of plugins. The way I've done it in this PR is: if goose discovers a plugin in one of your plugins paths, it will check if there's a matching entry in your config and if that is set to "disabled", it won't load it. Otherwise, it treats the plugin as "enabled".

DOsinga:
Pushed a commit addressing the two Codex review comments:

1. **Inline `mcpServers` form** — `is_inline_config` now detects the standard single-nested inline form (`"mcpServers": {"server-name": {"command": ...}}`) by treating any object that isn't a path-config (i.e. not made up solely of `paths`/`exclusive` keys) as a server map, rather than requiring a redundant nested `mcpServers` key. Previously the documented inline form was silently skipped. Updated the tests to use the single-nested form and added a test that an empty inline server command is rejected.

2. **CLI recipe gating** — the CLI session builder now gates plugin MCP loading on `recipe_extensions.is_none()` instead of `recipe.is_none()`, matching the server (`recipe_extensions.is_none()`) and scheduler (`recipe.extensions.is_none()`) paths. Previously `goose run --recipe foo.yaml` with a recipe that omits `extensions` would silently miss plugin MCP tools, unlike the equivalent server/scheduled session.

Also added `cwd: None` to the ACP stdio extension conversion that landed on main after this branch was created (needed to compile after merging main).

DOsinga (approved):

PR #9473, opened 2026-05-28
feat(server): paginate messages in GET /sessions/{session_id}

### LINKED ISSUE #9358: Pagination for messages in GET /sessions/{session_id}

GET /sessions/{session_id} returns the full Session object including the entire conversation history inline. For long-running sessions this could be a very large payload, even after compaction.

This makes it impractical to use the endpoint for displaying session history incrementally, and puts unnecessary pressure on memory and network for clients that only need recent messages.

Optional pagination query params on GET /sessions/{session_id} to allow fetching messages in chunks e.g. ?limit=50&cursor=<token> or ?limit=50&before=<message_id>. 

When omitted the endpoint would behave exactly as today, keeping it fully backwards-compatible.

None

**Additional context**

A few open questions:
  - Should pagination live on the existing endpoint or a dedicated /messages sub-resource?
  - What's the right cursor anchor — message timestamp, index, or an opaque ID?

Happy to put together a PR if the approach sounds reasonable.

- I have verified this does not duplicate an existing feature request

SarthakB11:
## Summary

Adds optional `limit` and `offset` query params to `GET /sessions/{session_id}` so clients can fetch session history in chunks instead of receiving the full conversation inline (#9358).

- `limit`: maximum messages to return, counted back from the most recent.
- `offset`: messages to skip from the most recent before applying `limit`.
- Both optional. When neither is set the response is byte-for-byte unchanged.
- `message_count` on `Session` still reflects the full conversation length so clients know when to stop paging.

Param shape follows the existing `SessionsQuery` pattern in `crates/goose-server/src/routes/schedule.rs` (`utoipa::IntoParams`, camelCase via `#[serde(rename_all = "camelCase")]`). Picked `offset` over a cursor token because the issue left the cursor anchor open and `offset` keeps the change minimal without locking in an opaque token format. A cursor variant can layer on later without breaking this one.

Slicing happens at the route layer over the `Conversation` returned by `SessionManager::get_session`, via a pure `paginate_messages` helper, so no core-crate changes. The helper clamps oversized `limit`/`offset` and preserves chronological order.

`ui/desktop/openapi.json` regenerated via `just generate-openapi`; the only delta is the two new query params on this route.

### Testing

- `cargo test -p goose-server --lib session::tests`: 4/4 pass (full history, limit-only, offset paging, oversized clamp).
- `cargo fmt`, `cargo clippy -p goose-server --all-targets -- -D warnings`: clean.

### Related Issues

Closes #9358


DOsinga:
Merged `main` into the branch to clear the conflict in `crates/goose-server/src/routes/session.rs` (the only real conflict was the import block — kept both `goose::agents::ExtensionConfig` and the new `goose::conversation::Conversation` imports). Verified: `cargo build -p goose-server`, `cargo test -p goose-server --lib session` (all pagination tests pass), and `cargo clippy -p goose-server --all-targets -- -D warnings` are all clean. The generated OpenAPI schema and TS client are already in sync — re-running the generator produced no diff. The branch is now up to date with main.

SarthakB11:
The goose-server REST crate was removed in #10224 when goose moved to ACP, so this endpoint no longer exists. Closing this. I opened #10365 to add the same pagination on the ACP load-session path.

PR #9474, opened 2026-05-28
Import sesssions

DOsinga:
## Summary

Allow importing sessions from codex, pi & claude


michaelneale (approved):
I have always wanted this. For bonus - you could enhance the recall platform extension to let it find and import sessions from any of these - ie you just ask goose "can you find X from claude" and it could import it (I guess then you want to start a new session... hrm). 

But nice!

FredApps:
Only failing check is `cargo fmt`. Run `cargo fmt --package goose` (or `cargo fmt --all`) and push the result.

PR #9479, opened 2026-05-29
fix(agent): stamp tool request with originating response timestamp

### LINKED ISSUE #9461: Tool request messages get later timestamps than tool responses, breaking Claude API message ordering

jpatel3:
## Summary

Fixes #9461.

Tool request messages (`assistant`/`tool_use`) were being assigned a `created` timestamp **after** tool execution, while their paired tool response messages (`user`/`tool_result`) were stamped **before** execution. When a tool ran long enough to cross a one-second boundary, the request ended up with a *later* timestamp than its response.

The session DB retrieves messages with `ORDER BY created_timestamp, id` (`crates/goose/src/session/session_manager.rs`). The `id` auto-increment tiebreaker only matters when timestamps are equal — so an out-of-order pair causes the `tool_result` to sort ahead of its `tool_use`, and the Claude API rejects the conversation with:

```
```

## Fix

In the tool-request construction loop in `crates/goose/src/agents/agent.rs`, carry the originating provider response's `created` timestamp onto the tool-request message. This mirrors how thinking/reasoning content is already stamped (`thinking_msg` uses `response.created`). The request and response now share a timestamp, so the insertion-order `id` tiebreaker preserves the correct request-before-response ordering.

## Test

Added `test_tool_request_not_stamped_after_response` in `crates/goose/tests/agent.rs`. A mock provider emits a tool request stamped well in the past; the test drives the agent through one tool turn and asserts the persisted tool-request keeps that timestamp and is never stamped after its response. Verified the test fails without the fix (the request gets the wall-clock time) and passes with it.

```
cargo test -p goose --features rustls-tls --test agent max_turns_tests
```

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -1981,6 +1981,15 @@ impl Agent {
                                                 request.metadata.as_ref(),
                                                 request.tool_meta.clone(),
                                             );
+
+                                        // Stamp the tool request with the originating response's
+                                        // timestamp. Tool execution can cross a one-second boundary,
+                                        // and the response placeholder was created before execution
+                                        // began; without this the request would sort after its
+                                        // response in the session DB (ORDER BY created_timestamp, id)
+                                        // and the Claude API would reject the out-of-order tool_result.
+                                        request_msg.created = response.created;
+
                                         messages_to_add.push(request_msg);
                                         let final_response = request_to_response_map
                                             .remove(&request.id)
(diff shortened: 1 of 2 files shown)


DOsinga:
Thanks for digging into this and the clear write-up in #9461! This turned out to be a popular bug — the same fix landed in #9462, which clamps the tool request timestamp to its response so the session ordering stays correct. Since that's now merged on `main`, I'm going to close this as a duplicate. Really appreciate the contribution, and hope you'll send more our way! 🙏

PR #9480, opened 2026-05-29
fix(agent): keep tool request before its response in session history

### LINKED ISSUE #9461: Tool request messages get later timestamps than tool responses, breaking Claude API message ordering

MukundaKatta:
## Summary

The response placeholder for a tool call is inserted into `request_to_response_map` before the tool runs, so it gets a `created` timestamp from before execution. The matching request message is built after the tool finishes, so it gets a later timestamp. Session history is ordered by `created`, so the tool request can sort after its own response, or tie and order nondeterministically by id. That breaks request-before-response ordering when the conversation is reloaded.

This retrieves the response first and backdates the request when it is not already earlier, so the request always sorts before its response. The logic lives in a small `ensure_request_precedes_response` helper.

### Testing

Added two unit tests for the helper: equal and later request timestamps get backdated, and an already-earlier request is left alone. `cargo test -p goose --lib ensure_request_precedes_response` passes and `cargo fmt --check` is clean.

### Related Issues

Fixes #9461

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -99,6 +99,19 @@ fn extract_string_arg(input: &Value, keys: &[&str]) -> Option<String> {
     None
 }
 
+/// Ensure a tool request message sorts before its response.
+///
+/// The response placeholder is created before tool execution begins, so it can
+/// carry an earlier (or equal) `created` timestamp than the request message,
+/// which is built afterward. Session history is ordered by `created`, so without
+/// this the tool request would sort after its own response. Backdate the request
+/// when needed so request-before-response ordering is preserved.
+fn ensure_request_precedes_response(request: &mut Message, response: &Message) {
+    if request.created >= response.created {
+        request.created = response.created.saturating_sub(1);
+    }
+}
+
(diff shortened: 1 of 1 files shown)


DOsinga:
Thanks so much for digging into this and sending a fix — really appreciate it! 🙏

Sorry, somebody beat you to it on this one: the same ordering bug (#9461) was already being tackled in a few earlier PRs, including #9462 (from the person who reported the issue) and a maintainer attempt in #9739. To avoid duplicate effort I'm going to close this one in favour of those.

One thing worth flagging for next time: this approach backdates the request to `response.created - 1`, which can push the tool request *before* the preceding user message when they land in the same second (timestamps are second-grained), re-introducing an ordering problem on reload. The other PRs align the request to the *same* second as the response and lean on the insertion-order id tiebreaker, which sidesteps that.

Please don't let this discourage you — would love to see more contributions! 🚀

PR #9482, opened 2026-05-29
fix(desktop): activate hermit on macOS so STDIO extensions resolve node

### LINKED ISSUE #8046: [macOS] STDIO extensions fail: "chown: command not found" and "env: node: No such file or directory" in Hermit bootstrap
**Describe the bug**

STDIO-type extensions that use `npx mcp-remote` fail to initialize on macOS (Apple Silicon) due to two issues in the Hermit bootstrap and Node.js PATH resolution:

1. **`chown: command not found` (exit 127)** — The Hermit install script (`install-<sha>.sh`) calls `chown` without a full path. On macOS, `chown` lives in `/usr/sbin/`, which is not included in the PATH inherited by GUI-launched apps (Electron). The bootstrap script at `~/.config/goose/mcp-hermit/bin/hermit` does not add `/usr/sbin` to PATH before invoking the install script.

2. **`env: node: No such file or directory`** — Even after Hermit successfully bootstraps and installs Node.js, the `npx` shim fails because `node` is not in PATH at runtime. Root causes:
   - The generated `hermit.hcl` contains a hardcoded `env.PATH` that overrides Hermit's automatic PATH management, preventing it from adding the Node.js binary directory.
   - Hermit sets `NPM_CONFIG_PREFIX` to `~/.config/goose/mcp-hermit/.hermit/node/`, and adds `.hermit/node/bin/` to PATH — but this directory does not contain actual `node`/`npx`/`npm` binaries. The real binaries are in `~/.config/goose/mcp-hermit/cache/pkg/node-<version>/bin/`.

**Note:** The `node-setup-common.sh` script in `Goose.app/Contents/Resources/bin/` only runs `hermit install node` and does NOT activate the Hermit environment on macOS (activation is Linux-only per the conditional on line ~113). This means `npx` is invoked via the Hermit shim (`bin/npx` → symlink to `bin/hermit`), which relies on `hermit exec` to set up the environment — but the hardcoded PATH in `hermit.hcl` breaks this mechanism.

---

**To Reproduce**

1. Install Goose Desktop on macOS (Apple Silicon)
2. Add a new STDIO extension with command: `npx mcp-remote https://block.gitmcp.io/goose/`
3. Save and observe "Failed to add extension" error
4. Check `/tmp/mcp.log` for detailed error output

First run shows:
```
Bootstrapping .../cache/pkg/hermit@stable/hermit from https://github.com/cashapp/hermit/releases/download/stable
/var/folders/.../tmp.XXXXXXXX: line 67: chown: command not found
An error occurred. Exiting with status 127.
```

After manually fixing the `chown` PATH issue, subsequent runs show:
```
Node setup (common) completed successfully.
Executing 'npx' command with arguments: mcp-remote https://block.gitmcp.io/goose/
env: node: No such file or directory
Failed to execute 'npx' with arguments: mcp-remote https://block.gitmcp.io/goose/
```

---

**Expected behavior**

The extension should initialize successfully. Hermit bootstrap should complete without `chown` errors, and `npx` should be able to find `node` in PATH when executing MCP extensions.

---

**Screenshots**

*(Extension panel showing 6/7 loaded, "Goose Docs" failed with red indicator)*

---

- **OS & Arch:** macOS 25.3.0 aarch64 (Apple Silicon)
- **Interface:** UI
- **Version:** 1.28.0
- **Extensions enabled:** code_execution, tom, summon, todo, Extension Manager, analyze, goose Docs
- **Provider & Model:** openrouter – moonshotai/kimi-k2-thinking

---

**Additional context**

### Environment details

- The GUI app's PATH (via `launchctl getenv PATH`) was: `/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin` — missing `/usr/sbin` and `/sbin`
- macOS system `/etc/paths` correctly includes `/usr/sbin` and `/sbin`, but these are not inherited by Electron apps in all configurations

### Workarounds applied

Three manual fixes were required to get the extension working:

**Fix 1** — Added `/usr/sbin` and `/sbin` to PATH in `~/.config/goose/mcp-hermit/bin/hermit`:
```bash
# After "set -eo pipefail", added:
export PATH="/usr/sbin:/sbin:${PATH}"
```

**Fix 2** — Removed hardcoded PATH override from `~/.config/goose/mcp-hermit/bin/hermit.hcl`:
```hcl
# Changed from:
env = {
  "PATH": "/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin",
}
# To:
env = {
}
```

**Fix 3** — Created symlinks in `~/.config/goose/mcp-hermit/.hermit/node/bin/` pointing to the actual Node.js binaries in `~/.config/goose/mcp-hermit/cache/pkg/node-<version>/bin/`:
```bash
ln -sf .../cache/pkg/node-24.14.0/bin/node  .../hermit/node/bin/node
ln -sf .../cache/pkg/node-24.14.0/bin/npm   .../hermit/node/bin/npm
ln -sf .../cache/pkg/node-24.14.0/bin/npx   .../hermit/node/bin/npx
ln -sf .../cache/pkg/node-24.14.0/bin/corepack .../hermit/node/bin/corepack
```

Pluviobyte:
## Summary

STDIO extensions that shell out to `npx` (e.g. `npx mcp-remote ...`) fail to initialize in the **macOS desktop app** because of two PATH problems in the shared Hermit bootstrap shim `ui/desktop/src/bin/node-setup-common.sh`:

1. **`chown: command not found` (exit 127).** Hermit's bootstrap calls `chown`, which lives in `/usr/sbin` on macOS. GUI-launched (Electron) apps inherit a minimal PATH — the reporter's was `/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin`, with no `/usr/sbin` or `/sbin` — so the bootstrap aborts.
2. **`env: node: No such file or directory`.** `activate-hermit` is sourced on **Linux only** (the guard was added in #5372 to "preserve existing behavior" on macOS). With that shim path now broken on macOS, the hermit-managed `node`/`npx` never get onto PATH at runtime.

This applies exactly the fix outlined by @DOsinga in #8046:

- Prepend `/usr/sbin:/sbin` to PATH early in the script (harmless on Linux, where these are normally already present).
- Remove the Linux-only guard so `bin/activate-hermit` is sourced on **every** platform — the same activation Linux already relies on.

The wrappers `bin/node` and `bin/npx` both `source` this common script, so both inherit the fix. Net change is +13/-5 in one file.

### Testing
- `bash -n` passes on `node-setup-common.sh`, `node`, and `npx`.
- I do **not** have a macOS machine to verify the runtime path end-to-end. As with the original Linux fix (#5372, whose author also requested macOS testers), I'd appreciate a **macOS (Apple Silicon)** user confirming the fix: build the desktop UI (`just run-ui`), then disable/re-enable an `npx`-based STDIO extension (e.g. `npx mcp-remote https://block.gitmcp.io/goose/`) and check `/tmp/mcp.log` — it should now reach `Node setup (common) completed successfully.` and the extension should load.

### Related Issues
Fixes #8046

N/A — build-time shim change only.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/ui/desktop/src/bin/node-setup-common.sh
+++ b/ui/desktop/src/bin/node-setup-common.sh
@@ -23,6 +23,12 @@ trap 'log "An error occurred. Exiting with status $?."' ERR
 
 log "Starting node setup (common)."
 
+# GUI-launched macOS apps inherit a minimal PATH that omits the system sbin
+# directories, so Hermit's bootstrap cannot find tools like `chown` (which lives
+# in /usr/sbin on macOS) and aborts with exit 127. Ensure they are reachable;
+# this is harmless on Linux, where these paths are typically already present.
+export PATH="/usr/sbin:/sbin:${PATH}"
+
 if [ -n "${GOOSE_PATH_ROOT:-}" ]; then
     RESOLVED_GOOSE_CONFIG_DIR="${GOOSE_PATH_ROOT}/config"
 elif [ -n "${GOOSE_CONFIG_DIR:-}" ]; then
@@ -108,11 +114,13 @@ else
     log "Hermit environment already initialized. Skipping init."
 fi
 
-# Activate the environment with output redirected to log
-if [[ "$(uname -s)" == "Linux" ]]; then
-    log "Activating hermit environment."
-    { . "bin/activate-hermit"; } >> "${LOG_FILE}" 2>&1
-fi
+# Activate the environment with output redirected to log.
+# Activation must run on every platform: macOS GUI apps otherwise never get the
+# hermit-managed node/npx onto PATH, so STDIO extensions fail with
+# "env: node: No such file or directory". The Linux-only guard was introduced in
+# #5372 to "preserve existing behavior" on macOS, but that path is now broken.
+log "Activating hermit environment."
+{ . "bin/activate-hermit"; } >> "${LOG_FILE}" 2>&1
 
 # Install Node.js using hermit
 log "Installing Node.js with hermit."


jamadeo (approved):

jamadeo:
This worked for me on macos. Thank you @Pluviobyte

PR #9484, opened 2026-05-29
fix(model): raise default max_output_tokens floor for Claude 4+ (adds opus-4-7/4-8)

michaelneale:
Update canonical models and improve default floor for claude family. 

opus 4.8 is out.

### DIFF
--- a/crates/goose/src/model.rs
+++ b/crates/goose/src/model.rs
@@ -451,17 +451,23 @@ impl ModelConfig {
         if !lower.contains("claude") {
             return false;
         }
-        // Match `claude-<family>-<major>` where major >= 4. Covers ids like
-        // `claude-opus-4-8`, `claude-sonnet-4-5`, `anthropic/claude-opus-4.6`,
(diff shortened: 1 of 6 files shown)


DOsinga:
Quick question on the `max_output_tokens()` fallback: rather than special-casing `is_claude_4_or_newer()` to return 32k while everything else still falls back to 4096, why not just bump the global default floor to 32k?

The reasoning:
- Older models that genuinely cap below 32k should already be in the canonical table (the source of truth), so they'd return their real `max_tokens` and never hit the fallback at all.
- These days there are very few models that can't do 32k output to begin with, so 4096 is a far more dangerous default than 32k — the whole motivation here is that 4096 silently truncates tool calls mid-stream.
- A flat 32k default would let us drop the Claude-specific regex (`is_claude_4_or_newer`, which has to handle both family-first and version-first id orderings) and its tests, reducing the maintenance surface.

Is there a known class of models that (a) are missing from the canonical table and (b) actually break or over-bill if asked for 32k? If not, a flat 32k default seems simpler and safer than the per-family floor.

PR #9519, opened 2026-05-30
feat(summon): add peek mode for async background tasks

jeffa-block:
## Summary
Adds `load(source: "task_id", peek: true)` — a non-blocking way to check the progress of a running async delegate without waiting for completion or cancelling it.

## Motivation
Currently, the only ways to interact with a running async task are:
- `load(task_id)` — blocks up to 5 minutes waiting for completion
- `load(task_id, cancel: true)` — kills the task

There's no way to ask "how's it going?" without committing to one of those outcomes. The infrastructure for status reporting already exists (`notification_buffer`, `last_activity`, `turns` counter) but isn't exposed to the model.

## Changes
- Add `peek: bool` parameter to the `load` tool JSON schema
- Parse `peek` at the call site alongside `cancel`
- When `peek: true` and task is running: return status (description, elapsed time, turns taken, idle duration, buffered tool call count) without blocking or consuming the task
- When `peek: true` and task is already completed: return the full result (same as normal load)
- 3 new tests: peek running task, peek nonexistent task, peek completed task

## Testing
- `cargo check` clean
- `cargo clippy` clean
- `cargo fmt` clean
- Task remains in `background_tasks` after peek (non-destructive read)


DOsinga:
Pushed a fix for the codex review comment about peek being destructive at the completion boundary.

Previously the completed-task branch in `handle_load_task_result` unconditionally did `completed.remove(task_id)`, so if a task finished just before a peek, the peek would consume the result and a subsequent peek/load would return "not found". Now peek reads completed results in place (leaving them retrievable), while a normal (non-peek) load still consumes them as before. Added an assertion to `test_peek_completed_task_returns_result` to lock in the non-destructive behaviour.

Also merged latest `main`.

DOsinga (approved):

PR #9524, opened 2026-05-30
feat(recipes): add diagnose-systemd-service recipe

JuanMarchetto:
## Summary
Adds a recipe, **Diagnose a Failed systemd Service**, that walks goose through diagnosing a failing systemd unit using the builtin `developer` extension.

## Why
A very common Linux task. The recipe is strictly **read-only**: it inspects `systemctl status`, `journalctl`, and `systemctl cat`, identifies the root cause with evidence from the logs, and presents a fix for the user to review and run themselves. It never restarts, enables, or modifies a unit on its own, and it shows (but never executes) any `sudo` command.

## Testing
- YAML parses and includes all required fields (version, title, description, instructions, prompt, extensions, activities).
- Unique filename; takes one `service` parameter.
- Read-only / no destructive actions.

`documentation/src/pages/recipes/data/recipes/diagnose-systemd-service.yaml`

### DIFF
--- a/documentation/src/pages/recipes/data/recipes/diagnose-systemd-service.yaml
+++ b/documentation/src/pages/recipes/data/recipes/diagnose-systemd-service.yaml
@@ -1,0 +1,46 @@
+version: "1.0.0"
+title: "Diagnose a Failed systemd Service"
+author:
+  contact: JuanMarchetto
+description: "Investigate why a systemd service is failing and suggest a fix — strictly read-only, never restarts or changes anything itself."
+instructions: |
+  You are a Linux systems expert helping diagnose a failed or misbehaving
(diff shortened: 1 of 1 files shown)


DOsinga:
Thanks for the contribution, @JuanMarchetto! This is a nicely-written, read-only recipe and the safety guardrails are appreciated.

That said, we're currently not accepting new recipes into the catalog while we work on a more scalable solution for managing the recipes/extensions registry — the same reason we've paused on adding new MCP servers. We don't want to keep growing the in-repo catalog ad hoc in the meantime.

I'm going to close this for now. We really appreciate the effort and hope you'll contribute again once we have the new registry approach in place.

PR #9530, opened 2026-05-31
openai: read meta.n_ctx from /v1/models to fix context limit for local servers

### LINKED ISSUE #9529: Enhancement: read context limit from meta.n_ctx in llama.cpp/Ollama model responses to fix auto-compaction for local servers
## Summary

When using Goose with a local llama.cpp or Ollama server, auto-compaction never fires and the session eventually crashes with a context length error. The root cause is that Goose's OpenAI-compatible provider doesn't read the context limit from local server API responses — it falls back to `DEFAULT_CONTEXT_LIMIT = 128_000` (hardcoded in `crates/goose/src/model.rs`), while ignoring any context information in the API response entirely. Local servers like llama.cpp expose the actual runtime context window via a non-standard `meta.n_ctx` field in the `/v1/models` response, but Goose ignores it since it's not part of the OpenAI spec.

## Environment

- Goose version: 1.36.0
- Provider: openai (pointing to local llama.cpp server)
- Model: Qwen3.6-35B-A3B (GGUF, llama.cpp b9186)
- OS: Windows 11

## What the API Returns

llama.cpp's `/v1/models` response includes a `meta` field:

```json
{
  "meta": {
    "n_ctx": 32768,
    "n_ctx_train": 262144
  }
}
```

`n_ctx` is the actual allocated context window (set via the `-c` flag). `n_ctx_train` is the model's theoretical training context. Goose ignores both and uses the 128K default, so `check_if_compaction_needed` computes usage ratio against 128K while the server hard-errors at 32K.

## Proposed Enhancement

In the OpenAI-compatible provider's model list parser, check for `meta.n_ctx` in the response and use it to set `context_limit` on `ModelConfig` when present. This would make auto-compaction work correctly for llama.cpp and Ollama servers without any manual config.

## Workaround

Adding `GOOSE_CONTEXT_LIMIT: 32768` to `config.yaml` fixes it — the explicit override works correctly and auto-compaction triggers as expected at 80% of the real context window.

## Why This Matters

Local servers are the only case where context limits *really* matter for compaction (cloud providers have huge windows and usually manage context themselves), and they're also the only case where the API actually tells you the real limit. The fix would be a small addition to the model list response parser.

CreamyG31337:
## Summary

Fixes #9529 (related: #8835).

When Goose is pointed at a local llama.cpp or Ollama server, auto-compaction never fires correctly because the context limit falls back to `DEFAULT_CONTEXT_LIMIT = 128_000` — even though the server reports the actual window via a non-standard `meta.n_ctx` field in the `/v1/models` response.

**Changes:**

- Add `fetch_n_ctx_from_api(&self, model_name: &str) -> Option<usize>` to `OpenAiProvider` — hits the models endpoint, finds the entry matching the active model, and returns `meta.n_ctx` if present.
- In `from_env()`, after construction, if `model.context_limit` is still `None` (no `GOOSE_CONTEXT_LIMIT` config and no canonical registry match), call `fetch_n_ctx_from_api` and set `context_limit` from the result.
- Fails silently — real OpenAI servers don't include `meta`, so this is a no-op for cloud usage.

**Example response from llama.cpp:**
```json
{
  "meta": {
    "n_ctx": 32768,
    "n_ctx_train": 262144
  }
}
```
With this change, `context_limit` is set to `32768` instead of `128000`, so auto-compaction fires at the correct threshold.

## Test plan

- [ ] `cargo check -p goose` passes
- [ ] Pointing Goose at a llama.cpp server (without `GOOSE_CONTEXT_LIMIT` set) shows the correct context window size
- [ ] Pointing Goose at real OpenAI shows no change in behavior (field absent, falls back to canonical/default as before)
- [ ] Existing tests in `openai.rs` pass

### DIFF
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -634,9 +634,9 @@ impl OpenAiProvider {
         Ok(models)
     }
 
-    /// Read `meta.n_ctx` from the `/v1/models` response for the given model name.
-    /// llama.cpp and Ollama include this non-standard field with the actual allocated
-    /// context window size. Returns `None` if the field is absent (e.g. real OpenAI).
+    /// llama.cpp and Ollama expose the actual allocated context window in the
+    /// non-standard `meta.n_ctx` field of `/v1/models`. Returns `None` when absent
+    /// (e.g. real OpenAI).
     async fn fetch_n_ctx_from_api(&self, model_name: &str) -> Option<usize> {
         let models_path =
             Self::map_base_path(&self.base_path, "models", OPEN_AI_DEFAULT_MODELS_PATH);
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -1488,17 +1488,6 @@ mod tests {
     }
 
     #[test]
-    fn parse_n_ctx_matches_model_by_id() {
-        let body = json!({
-            "data": [
-                { "id": "other-model", "meta": { "n_ctx": 4096 } },
-                { "id": "my-model", "meta": { "n_ctx": 8192 } }
-            ]
-        });
-        assert_eq!(parse_n_ctx_from_models(&body, "my-model"), Some(8192));
-    }
-
-    #[test]
     fn parse_n_ctx_falls_back_to_sole_entry_when_id_differs() {
         let body = json!({
             "data": [
@@ -1517,11 +1506,5 @@ mod tests {
             ]
         });
         assert_eq!(parse_n_ctx_from_models(&body, "model-c"), None);
-    }
-
-    #[test]
-    fn parse_n_ctx_absent_when_meta_missing() {
-        let body = json!({ "data": [ { "id": "my-model" } ] });
-        assert_eq!(parse_n_ctx_from_models(&body, "my-model"), None);
     }
 }
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -264,7 +264,7 @@ impl OpenAiProvider {
             api_client = api_client.with_headers(header_map)?;
         }
 
-        let provider = Self {
+        let mut provider = Self {
             api_client,
             base_path,
             organization,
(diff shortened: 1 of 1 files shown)


CreamyG31337:
Addressed both Codex P2 items in 5e01592:

**P2-1 (alias matching known OpenAI model):** Changed the guard from `context_limit.is_none()` to `!is_openai`. For any non-api.openai.com host the server-reported `meta.n_ctx` now wins regardless of what the canonical registry pre-populated. Hosts without `meta.n_ctx` (Azure, OpenRouter, etc.) return `None` and their existing limit is unchanged.

**P2-2 (single-model server without `--alias`):** After the id-match loop, if `data.len() == 1` we fall back to that sole entry's `n_ctx`. Covers the common llama.cpp case where the id is the model file path rather than the client-facing name.

CreamyG31337:
Addressed the P2 in 0311b38.

You're right — switching the guard to `!is_openai` made it stomp an explicit `GOOSE_CONTEXT_LIMIT`, which is ironic since that's the exact workaround #9529 documents. Since `provider_registry::normalize_model_config` runs `with_canonical_limits` before `from_env`, an explicit override and a canonical default are indistinguishable on the `ModelConfig` by the time the provider sees it.

Fix: re-check `GOOSE_CONTEXT_LIMIT` via `config.get_param` the same way `ModelConfig::new_base` does, and skip the `meta.n_ctx` override when it's explicitly set (`> 0`). Canonical/known-model alias defaults — which don't come from that key — are still overridden as intended.

Resulting precedence for non-OpenAI hosts: `GOOSE_CONTEXT_LIMIT` > server `meta.n_ctx` > canonical/known-models default > `DEFAULT_CONTEXT_LIMIT`.

CreamyG31337:
Good catch on the ACP path — fixed in 57cd91d, but by stepping back rather than adding a fourth special case.

The ACP finding exposed the real issue: there are several explicit-override channels — `GOOSE_CONTEXT_LIMIT`, ACP/server per-session `with_context_limit()`, and `GOOSE_PREDEFINED_MODELS` — and by provider-construction time **all of them, plus a canonical/known-models default, are indistinguishable** on the `ModelConfig` (every one is just `context_limit = Some(x)`). Chasing each override channel with its own guard is a losing game.

So I've reverted the guard to `context_limit.is_none()`, which mirrors how `with_canonical_limits` fills derived limits and preserves **every** explicit override at once — env, ACP, and predefined all keep working.

**On the earlier P2-1 (override canonical alias defaults):** I'm intentionally *not* doing that anymore, and I'd argue it shouldn't be done at this layer. Overriding a canonical default is indistinguishable from overriding a deliberate per-session/ACP limit, so honoring P2-1 necessarily reintroduces this ACP regression. The niche case it covered — deliberately aliasing a local model to a known OpenAI name like `gpt-3.5-turbo` — is still fully addressable with `GOOSE_CONTEXT_LIMIT`, the documented escape hatch. Fully resolving P2-1 *and* preserving all overrides would require threading provenance (explicit vs. registry-derived) through `ModelConfig` + `normalize_model_config`, which is a larger change I'm happy to do as a follow-up if you'd prefer it.

Net precedence for local servers now: any explicit limit (env / ACP / predefined / canonical) > server `meta.n_ctx` > `DEFAULT_CONTEXT_LIMIT` (128k).

CreamyG31337:
Good catch, and a real one — fixed in 7281663.

The probe reused the provider's `ApiClient`, whose timeout is `OPENAI_TIMEOUT` (default `DEFAULT_PROVIDER_TIMEOUT_SECS` = 600s), so a reachable-but-hanging `/v1/models` could have blocked provider construction — and thus Goose startup / model switching — for up to 10 minutes. Wrapped the probe in a 5s `tokio::time::timeout`; on timeout or any error it silently falls back to the previous behavior, so a slow or missing models endpoint can't stall construction.

FWIW this revision now compiles clean locally (`cargo check -p goose`, only the two pre-existing unrelated warnings) and I'm running the `providers::openai` unit tests.

FredApps:
The only failing check is `cargo fmt`. All tests/lint/MSRV pass. Just run `cargo fmt --package goose` and push — the long line in `from_env()` is almost certainly what rustfmt wants to wrap:

```rust
tokio::time::timeout(N_CTX_PROBE_TIMEOUT, provider.fetch_n_ctx_from_api(&model_name))
```

CreamyG31337:
Thanks! Done in `1c23a58` — `cargo fmt --package goose` wrapped exactly that `from_env()` line as you predicted. `cargo fmt --package goose --check` is now clean locally. Should be green on the next CI run.

DOsinga:
Thanks for the fix and for filing the linked issue (#9529) — this is a genuinely useful bug for local OpenAI-compatible servers, and nice work responding to the Codex feedback.

I pushed a small cleanup commit to your branch (`603c9833`):
- Condensed the long rationale comment in `from_env` and dropped the ownership shuffle (made `provider` `mut` instead).
- Extracted the `meta.n_ctx` parsing into a pure `parse_n_ctx_from_models` function so the matching/fallback logic is unit-testable without an HTTP mock.
- Added four tests covering: match-by-id, single-entry fallback when the id differs (the llama.cpp file-path case), no fallback with multiple unmatched entries, and missing `meta`.

One thing left before this can merge: **the DCO check is failing** because none of your commits carry a `Signed-off-by` line. Could you sign off your existing commits? The simplest way:
```
git rebase HEAD~6 --signoff   # or however many commits are yours
git push --force-with-lease
```
(Note that will re-push over my cleanup commit too — feel free to re-sign all of them, or cherry-pick mine back; it's already signed.) Once DCO is green this looks good to me.

DOsinga (approved):

DOsinga (approved):
Thanks for the careful work here, and especially for engaging so thoroughly with the codex review — addressing each point with clear reasoning (including knowing when *not* to chase an override path). This is a clean, well-scoped fix that reads `meta.n_ctx` from `/v1/models` for local OpenAI-compatible servers, which fixes context sizing without affecting cloud OpenAI. Approving.

PR #9531, opened 2026-05-31
fix: use llama.cpp context metadata for OpenAI-compatible models

he-yufeng:
﻿## Summary
- read `meta.n_ctx` from OpenAI-compatible `/models` entries and surface it as `ModelInfo.context_limit`
- keep the existing supported-model ID API behavior by deriving IDs from the parsed model metadata
- preserve recommended-model filtering without dropping the provider-reported context limit

## Why
Local llama.cpp-compatible servers can expose their runtime context window in `/models` as `meta.n_ctx`. Goose currently only keeps the model IDs from that response, so custom OpenAI-compatible local providers fall back to the default 128k context limit. That makes compaction/token-limit UI decisions too optimistic for servers running with smaller contexts.

## To verify
- `cargo fmt --package goose`
- `cargo test -p goose models_api --lib`
- `cargo test -p goose openai_compatible --lib`

### DIFF
--- a/crates/goose/src/providers/openai_compatible.rs
+++ b/crates/goose/src/providers/openai_compatible.rs
@@ -13,7 +13,8 @@ use tokio_util::codec::{FramedRead, LinesCodec};
 use tokio_util::io::StreamReader;
 
 use super::api_client::ApiClient;
-use super::base::{stream_from_single_message, MessageStream, Provider};
+use super::base::{stream_from_single_message, MessageStream, ModelInfo, Provider};
+use super::canonical::{map_to_canonical_model, CanonicalModelRegistry};
 use super::retry::ProviderRetry;
 use super::utils::RequestLog;
 use crate::conversation::message::Message;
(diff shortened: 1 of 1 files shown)


FredApps:
Definitely have this problem, it does not read the context limit that is present in the API. I checked the code over, this is a legit fix. Recommend approval.

he-yufeng:
Rebased this branch onto current `aaif-goose:main` and resolved the provider conflict with the newer upstream streaming usage changes.

Validation:

```text
cargo fmt --check

cargo test -p goose providers::openai_compatible --lib
# 13 passed; 0 failed

git diff --check
```

The test run emits existing warnings outside this PR's touched provider file (`goose-mcp::subprocess`, `extension_manager`, `config::base`, and platform extension tests); I left those untouched.

he-yufeng:
Rebased this onto current `origin/main` and resolved the provider import conflict from the recent `goose_providers` split, while keeping the llama.cpp `meta.n_ctx` context-limit handling intact.

Validated locally:

- `cargo fmt --check`
- `cargo test -p goose providers::openai_compatible --lib` (13 passed)
- `git diff --check origin/main..HEAD`

CI is running again on the rebased head.

DOsinga:
Thanks so much for taking the time to put this together, @he-yufeng! 🙏

It turns out this overlaps heavily with #9530, which tackles the same underlying issue (#9529) — surfacing llama.cpp's context metadata from `/v1/models` so auto-compaction kicks in correctly for local servers. Since #9530 is already actively being iterated on (and is authored by the person who reported the issue), we're going to consolidate the effort there to avoid duplicate work.

One thing worth flagging for future reference: this PR patches `OpenAiCompatibleProvider`, but local llama.cpp / LM Studio / Ollama-style servers actually flow through `OpenAiProvider` (via `from_custom_config` in the declarative provider registration), so the new parsing path wouldn't be reached for those providers in practice. #9530 targets `OpenAiProvider` directly, which is why we're going with that one. There were also a couple of unaddressed codex review comments here (see https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#ai-code-reviews).

Going to close this as a duplicate of #9530, but please don't be discouraged — we'd love to have you contribute again, and jumping in on #9530's review thread is a great place to keep the momentum going. Thanks again! 🪿

Issue #9369, opened 2026-05-22
Session name is not updated when edited in certain cases (Linux)


LFd3v:
**Describe the bug**

💡 Before filing, please check common issues:  
https://goose-docs.ai/docs/troubleshooting  

📦 To help us debug faster, attach your **diagnostics zip** if possible.  
👉 How to capture it: https://goose-docs.ai/docs/troubleshooting/diagnostics-and-reporting/

The session name is not updates when it is edited in certain circumstances, like when a new chat is created.

---

**To Reproduce**
1. Start Goose Desktop
2. Click o +Start New Chat
3. Click on Show All
4. Select the New Chat that was recently created either by clicking on it in the top of the chat list or in Chat History (left most box)
5. Click on Show All again
6. Click on Edit Session Name (pencil icon) and change the session name
7. A notification is shown saying the session name was updated but New Chat is still visible.

---

**Expected behavior**
Session name should be updated every where in the app once it is saved.

---

**Screenshots**

<img width="1398" height="872" alt="Image" src="https://github.com/user-attachments/assets/9a83fb4e-c154-413f-a94a-dfa57c21d65f" />

---

**Please provide the following information**
- **OS & Arch:** ArchLinux/Garuda GNOME
- **Interface:** Desktop
- **Version:** v1.35.0 (also happens in v1.34.1)
- **Extensions enabled:** tom, apps, summon, skills, analyze, todo, Extension Manager, developer
- **Provider & Model:** Local Inference - unsloth/Qwen3.5-2B-GGUF:Q4_K_XL

---

**Additional context**
If a chat is selected first before creating a new chat, then the session name is updated in the chat list but not in the Chat History panel. Maybe this glitch is caused by an empty chat, because the session name is updated once a message is sent in the new chat that was renamed, or in Chat History after navigation to some other place (like opening the chat) and going back to Chat History (Show All).

DOsinga:
Fixed by #9543, merged on June 16, 2026. The fix updates `user_set_name` and synchronizes rename events across the sidebar and Chat History, including the reported case of renaming an empty “New Chat” session. Thanks for the detailed report!

PR #9543, opened 2026-05-31
fix: refresh session name in sidebar after rename

nuthalapativarun:
## Summary

After renaming a session that has no messages (message_count = 0), the sidebar would still display New Chat because the local state update after rename only updated the name field, not user_set_name. The getSessionDisplayName function checks user_set_name first; when it is false and message_count is 0, shouldShowNewChatTitle returns true and overrides the name with New Chat.

This fix:
- Sets user_set_name: true alongside name when handling the SESSION_RENAMED event in useNavigationSessions (sidebar list update)
- Sets user_set_name: true in the optimistic state update in SessionListView after a successful rename
- Fires SESSION_RENAMED from the NavigationPanel inline rename so other consumers are also notified

### Testing
- Manual: create a new chat (message_count = 0), rename it via Edit Session Name, and verify the new name appears immediately in both the sidebar and the chat history list
- Existing UI tests pass

### Related Issues
Relates to #9369

Before: Session name in sidebar remains New Chat after rename when no messages exist
After: Session name in sidebar updates immediately after rename

### DIFF
--- a/ui/desktop/src/components/Layout/NavigationPanel.tsx
+++ b/ui/desktop/src/components/Layout/NavigationPanel.tsx
@@ -105,6 +105,11 @@ const SessionRow: React.FC<SessionRowProps> = ({ session, active, status, onClic
             path: { session_id: session.id },
             body: { name: newName },
           });
+          window.dispatchEvent(
+            new CustomEvent(AppEvents.SESSION_RENAMED, {
+              detail: { sessionId: session.id, newName },
+            })
+          );
           onRenamed();
         }}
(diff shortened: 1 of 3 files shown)


DOsinga (approved):
Thanks for the fix — this is a clean, well-scoped fix tied to a real issue (#9369), and you correctly addressed codex's P2 by only setting `user_set_name` on genuinely user-initiated renames (via the `userInitiated` flag) while leaving the automatic title-refresh path in `useChatStream` untouched. Verified locally that recipe titles are preserved on auto-refresh. LGTM!

Issue #9365, opened 2026-05-22
Start new chat shows the most recent deleted chat contents (Linux)


LFd3v:
**Describe the bug**

💡 Before filing, please check common issues:  
https://goose-docs.ai/docs/troubleshooting  

📦 To help us debug faster, attach your **diagnostics zip** if possible.  
👉 How to capture it: https://goose-docs.ai/docs/troubleshooting/diagnostics-and-reporting/

When a new chat is started in Goose Desktop after deleting the most recent chat, the new chat shows the content of the deleted one.

---

**To Reproduce**
1. Start Goose Desktop
2. Select the most recent chat (top of the list)
3. In the Chat list, click on Show All
4. Delete the most recent chat (click on the red trash icon)
5. Confirm clicking on Delete Session
6. Click on +Start New Chat
7. New Chat is created with the contents from the chat deleted in step 4.

---

**Expected behavior**
A new chat should always be empty, without the contents of deleted chats.

---

**Screenshots**
N/A

---

**Please provide the following information**
- **OS & Arch:** ArchLinux/Garuda GNOME
- **Interface:** Desktop
- **Version:** v1.35.0 (also happens in v1.34.1)
- **Extensions enabled:** tom, apps, summon, skills, analyze, todo, Extension Manager, developer
- **Provider & Model:** Local Inference - unsloth/Qwen3.5-2B-GGUF:Q4_K_XL

---

**Additional context**
After the app is restarted (closed and relaunched) the new chat does not show the content from the deleted one.

lifeizhou-ap:
HI @LFd3v,

I think the issue has been fixed in https://github.com/aaif-goose/goose/pull/9351.  Could you please check whether this issue still exists with the latest goose version? Thank you!

LFd3v:
@lifeizhou-ap The UI changed a bit, but it seems to be working fine for me (v1.45.0). Thank you for taking time to reply. Closing.

PR #9544, opened 2026-05-31
fix: clear session state after delete so new chat starts empty

nuthalapativarun:
## Summary
When the user deletes the currently-active session and then creates a new chat, the new chat window could show the deleted session's content. This happened because:

1. The deleted session remained in `activeSessions` — keeping its `BaseChat` mounted and visible if the user navigated back to that session's URL.
2. The `chat` context state (`chatContext.chat.sessionId`) still held the deleted session's ID, causing `handleNavClick('/pair')` to navigate back to the deleted session URL.

The fix adds a `SESSION_DELETED` listener in `AppInner` that:
- Removes the deleted session from `activeSessions`, unmounting its `BaseChat`
- Resets the `chat` context state if it belonged to the deleted session, so navigation to `/pair` no longer falls back to the deleted session

### Testing
- Manual: delete the active session from the sessions list, create a new chat, verify it starts empty
- Existing UI tests pass (tests cannot be run in current environment due to missing node_modules, but no test logic was changed)

### Related Issues
Relates to #9365

Before: New chat after deleting the active session shows content from the deleted session  
After: New chat after delete starts with an empty conversation

### DIFF
--- a/ui/desktop/src/App.tsx
+++ b/ui/desktop/src/App.tsx
@@ -410,6 +410,24 @@ export function AppInner() {
     };
   }, []);
 
+  useEffect(() => {
+    const handleSessionDeleted = (event: Event) => {
+      const { sessionId } = (event as CustomEvent<{ sessionId: string }>).detail;
+      setActiveSessions((prev) => prev.filter((s) => s.sessionId !== sessionId));
+      setChat((prev) => {
+        if (prev.sessionId === sessionId) {
+          return { sessionId: '', name: DEFAULT_CHAT_TITLE, messages: [], recipe: null };
+        }
+        return prev;
+      });
+    };
+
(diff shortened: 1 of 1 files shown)


nuthalapativarun:
Bumping this — happy to address any feedback or rebase if needed.

DOsinga:
Thanks for the fix, @nuthalapativarun! The change itself looks clean and well-scoped.

However, you currently have a large number of PRs open at once, several opened in quick succession. Per our [contributing guidelines on pull requests](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#pull-requests), we ask contributors not to open many PRs at the same time, but instead to submit them in order of preference and wait for earlier ones to land before opening more. This keeps the review queue manageable and helps you build a track record.

I'm going to close this one for now. Please feel free to reopen/resubmit it once one or two of your earlier PRs have landed. Thanks for understanding!

PR #9560, opened 2026-06-02
feat: steering messages with ACP

michaelneale:
Adds an ACP steering method that queues prompt input into the active Goose run after validating 

similar in logic to what things like codex do - and an acp extension to support it.

cc @alexhancock (only as the acp maven!) am I on the right track?


michaelneale:
This can now tell you a message is a steering one or not from the metadata: 

On the ACP client, read the boolean off the chunk's meta — `update.sessionUpdate === 'agent_message_chunk' && chunk._meta?.goose?.steer === true` marks the echoed steer message; chunks before it are pre-steer assistant text, chunks after it are post-steer — no text matching, and `_meta.goose.messageId` / `activeRunId` disambiguate multiple steers.

lifeizhou-ap:
Hi @michaelneale 

Really like this steering feature! 

I have a couple of thoughts on the architecture:

- It would be great to keep the ACP server as a thin pass-through layer and let goose core own the steering logic. Right now the ACP server also holds run state (`active_run_id` alongside the agent's `pending_steers`). If the agent owned the queue end-to-end, like enqueuing the message when `session/steer` is called, and picking it up in the loop at the right time, the  ACP server wouldn't need to track runs, and the client wouldn't even need to maintain a run id. @DOsinga is working on the agent loop refactoring and may be able to consider in this steer use case in the state management too.

- It would also be nice to notify the client about the steer message's status (e.g. pending/queued, picked up), so the UI can reflect whether it's still pending.

michaelneale:
@lifeizhou-ap not sure I follow on 1 - the logic is in agent not relay already, not sure what you mean? the only code in acp is the run id stuff (which is fiddly, but not the steering specifically, will check with douwe)

will look at 2, this is long overdue in general (as all other agent harnesses have something like this, I just like codex the most).

michaelneale:
@lifeizhou-ap addressed the second point - on first one, I think for now want to keep more code out of agent loop until it settles maybe?

DOsinga (approved):
would be nice to build this into. the current desktop too. but we can follow up for that. I love this feature

PR #9568, opened 2026-06-02
Pick the last canonical model

DOsinga:
## Summary

@baxen noticed that we were changing output limits. we didn't pick the right canonical.


baxen (approved):
nice

PR #9570, opened 2026-06-02
Complexity router

DOsinga:
## Summary

Very much a work in progress but uses the complexity router to decide which model to pick (fast, slow)

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -214,6 +214,7 @@ pub struct Agent {
     container: Mutex<Option<Container>>,
     goal: Mutex<Option<String>>,
     grind: Mutex<Option<String>>,
+    complexity_router: Option<Arc<super::complexity_router::ComplexityModel>>,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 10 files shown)


crates/goose/src/agents/agent.rs
```diff
@@ -1641,17 +1589,46 @@ impl Agent {
...
+        let base_model_config = provider.get_model_config();
+        let requested_model = base_model_config.model_name.clone();
+
+        let route_decision = self
+            .complexity_router
+            .as_ref()
+            .and_then(|router| super::complexity_router::route(router.as_ref(), &conversation));
+
+        // Decide once per `reply_internal` call which model config to use.
+        // If the complexity router says "use_fast" and the provider has a
```
michaelneale, line 1601: I guess we will have dimensions like fast, cheap etc (not sure if we end up grouping, today often fast is cheap but not always: can think of them as different dimensions to solve for)

DOsinga: the only reason to go with fast here is that we already support that in our infra

michaelneale:
@DOsinga I think maybe we should pick this up next week or week after and pick an approach to add as a crate (and let it be extended, somehow. although most of the interesting routers I have seen out there are pretty heavy and run as proxies)

PR #9574, opened 2026-06-03
Unrolled agent loop

DOsinga:
## Summary

Unrolls the agent loop into a re-entrant state machine. Each call runs from the persisted conversation to the next applied step or client yield; no coroutine state is held between steps.

The new path remains behind `GOOSE_STATE_MACHINE` while it establishes parity with the legacy reply loop.

## Design

- `StateMachine::step` evaluates ordered steps until one returns `Applied`; `NotApplicable` passes control onward.
- `StepResult` contains explicit effects. `StateMachine::apply` persists them through `SessionManager`, while `StateMachine::run` is the convenience method that drives both until a yield.
- `Operation` is the common protocol for commands, tools, context management, hooks, and inference. Operations can also contribute tools, prompt parts, and MOIM parts to inference.
- Inference is an operation with one extra `applies` gate. It receives the collected inference input without knowing which operations produced it.
- The conversation is execution state. Operation notes are stored on messages under the operation name, so a pipeline can be discarded and rebuilt after any applied step.
- Goose constructs its standard pipeline beside `Agent::reply`; callers can construct a different ordered set of steps without depending on `Agent`.

## Operations

- **Commands:** slash-command dispatch, bang shell, doctor, and skills.
- **Turn and context:** steering, max turns, context compaction, and old tool-pair compaction.
- **Tools:** approval, cooperative execution, and unknown-tool recovery.
- **Recipes:** recipe tools/final output and retry checks.
- **Completion:** stop hooks and terminal error handling.
- **Inference:** provider streaming, prompt/tool/MOIM composition, usage, cancellation, and inference hooks.

## Testing

The state-machine suite builds the full standard test pipeline rather than testing operations in isolation.

- `DummyApi` emulates the provider HTTP API; the real OpenAI provider and API client still parse and stream its responses.
- A stateful calculator MCP extension provides deterministic tool execution, failures, concurrency, cancellation, elicitation, and audience filtering.
- Tests script provider behavior with `api.on(...)`, run user turns, and assert against the persisted session conversation.
- Lifecycle scenarios cover tools, compaction, hooks, prompts and skills, recipes and scheduling, steering, provider behavior, reconstruction, and session isolation.
- Reconstruction tests discard and rebuild the pipeline after each applied step, proving that progress comes from persisted state rather than operation instances.


jamadeo (approved):

michaelneale:
I think get this in. Only thing flagged which I noted of interest "pending_tool_requests scans persisted messages and treats request with no response as work to do" - ie if things break after a tool does some work (say move a file) and then it never persists, and then agent loop comes back one day it may behave differently to now, but I don't think relaly an issue in practice. 

Otherwise - lets get this in when is clean.

michaelneale (approved):

PR #9584, opened 2026-06-03
fix: show resolved skill supporting file paths

### LINKED ISSUE #9558: scripts relative paths in SKILL.md don't work — shell CWD is workspace, not skill directory
Hey! Running into an issue with agent skills that use executables in scripts.

The [Agent Skills spec](https://agentskills.io/specification) says skills should reference scripts using relative paths from the skill root, e.g.:

```
Run the extraction script:
scripts/my-tool.exe
```

But when the agent calls the `shell` tool with `scripts/my-tool.exe`, it executes with the **session's working directory** (the user's workspace) as CWD — not the skill's directory. So the file is never found.

The skill directory path does show up in the `load_skill` response when there are supporting files, but the LLM doesn't reliably use it to rewrite relative paths before calling `shell`.

The spec treats relative-path script references as a first-class feature, so I'd expect Goose to handle the path resolution somewhere rather than putting the burden on skill authors to work around it.

Thanks for the great project!

he-yufeng:
## Summary
- show the resolved filesystem path for each loaded skill supporting file
- make it explicit that relative paths in a skill resolve from the skill directory, not the session working directory
- add unit coverage for script-style supporting files

Fixes #9558

## Verification
- `cargo fmt --check`
- `git diff --check`
- `CARGO_INCREMENTAL=0 cargo test -p goose --lib loaded_skill_context --jobs 1 -- --nocapture`

### DIFF
--- a/crates/goose/src/skills/mod.rs
+++ b/crates/goose/src/skills/mod.rs
@@ -109,15 +109,20 @@ fn loaded_skill_context(skill: &SourceEntry, content: &str) -> String {
     if !skill.supporting_files.is_empty() {
         let skill_dir = Path::new(&skill.path);
         output.push_str(&format!(
-            "\n## Supporting Files\n\nSkill directory: {}\n\n",
+            "\n## Supporting Files\n\nSkill directory: {}\n\n\
+             Relative paths in this skill resolve from the skill directory. \
+             The shell tool runs in the session working directory, so use the \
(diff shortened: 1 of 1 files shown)


DOsinga (approved):

DOsinga:
thanks for doing this! would love it if you joined the discord too if you have other ideas on how to make this all work better

PR #9586, opened 2026-06-03
Use LRU cache for token counting

jh-block:
## Summary

Replace the token counter's per-instance `DashMap<u64, usize>` cache with the existing `LruCache<String, usize>` behind a mutex.

## Rationale

The token counter was the only direct `goose` crate user of `dashmap`, but its cache is per `TokenCounter` instance and production callers generally create a counter for one local counting operation. A concurrent hash map is more machinery than this cache needs, and the previous eviction path was not atomic. It also keyed entries by a `u64` text hash, which could return an incorrect count on collision.

Using `LruCache` keeps the cache bounded with deterministic eviction, preserves the existing `&self` counting API, avoids hash-key correctness issues by keying on the original text, and removes the direct `dashmap` and `ahash` dependencies from `goose`.

## Validation

- `cargo fmt`
- `cargo check -p goose`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2669,20 +2669,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "dashmap"
-version = "6.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e6361d5c062261c78a176addb82d4c821ae42bed6089de0e12603cd25de2059c"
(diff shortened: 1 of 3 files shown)


DOsinga (approved):
too be honest we could just kill that whole caching thing. it is something that an agent thought very important a year ago. we now barely use it and the caller can just take care of it.

PR #9588, opened 2026-06-03
create goose-providers crate with canonical models, conversation and other types

jamadeo:
Phase 1 of creating a new provider crate. Makes the crate itself and moves a few key types into it -- no provider implementation moves yet.


crates/goose-providers/src/base.rs
```diff
@@ -683,27 +661,16 @@ impl ProviderUsage {
...
-        crate::providers::usage_estimator::ensure_usage_tokens(
-            self,
-            system_prompt,
-            request_messages,
-            response,
-            tools,
-        )
-        .await
-        .map_err(|e| ProviderError::ExecutionError(format!("Failed to ensure usage tokens: {}", e)))
+        Ok(())
```
alexhancock: Why does this just do `Ok(())` now?

crates/goose-providers/src/config.rs
```diff
@@ -0,0 +1,72 @@
...
+        maybe_secret: &[&str],
+    ) -> Result<HashMap<String, String>, ProviderConfigError>;
+
+    fn set_param_value(&self, key: &str, value: Value) -> Result<(), ProviderConfigError>;
+    fn set_secret_value(&self, key: &str, value: Value) -> Result<(), ProviderConfigError>;
+    fn delete_secret(&self, key: &str) -> Result<(), ProviderConfigError>;
+    fn invalidate_secrets_cache(&self);
+}
+
+pub trait ProviderConfigExt {
```
alexhancock: These traits feel similar. I am missing the need for both of them.

crates/goose/src/bin/build_canonical_models.rs
```diff
@@ -320,7 +321,7 @@ impl MappingReport {
 
 fn data_file_path(filename: &str) -> PathBuf {
     PathBuf::from(env!("CARGO_MANIFEST_DIR"))
-        .join("src/providers/canonical/data")
+        .join("../goose-providers/src/canonical/data")
```
alexhancock, line 324: should look if we can use https://docs.rs/embuild/latest/embuild/cargo/fn.workspace_dir.html

crates/goose/src/providers/mode.rs
```diff
@@ -0,0 +1,69 @@
...
+    ($($provider:ty),* $(,)?) => {
+        $(
+            #[async_trait]
+            impl GooseProvider for $provider {}
+        )*
+    };
+}
+
+impl_default_goose_provider!(
+    AnthropicProvider,
```
alexhancock: Is there a way to automate this so we don't need to update it as new ones are added? May not be frequent, but I could still seeing it tripping us up.

alexhancock (approved):
A great move for provider logic to isolate it from the rest of goose

crates/goose-providers/src/base.rs
jamadeo: Yeah that's some model laziness. What we should do here is remove it from ProviderUsage completely, since it's used in only one place. Interestingly, the original PR that added this used it in .reply(), but it no longer seems to be there...

crates/goose-providers/src/model.rs
```diff
@@ -83,6 +85,23 @@ pub enum ConfigError {
...
+
+impl ProviderConfigReader {
+    fn get_param<T: DeserializeOwned>(&self, key: &str) -> Result<T, ProviderConfigError> {
+        get_provider_param(key)
+    }
+}
+
+fn get_provider_param<T: DeserializeOwned>(key: &str) -> Result<T, ProviderConfigError> {
+    let value = std::env::var(key.to_uppercase())
+        .map_err(|_| ProviderConfigError::NotFound(key.to_string()))?;
```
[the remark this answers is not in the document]
jamadeo: grr, yeah, there is more to do here.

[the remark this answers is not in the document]
jamadeo: Originally I'd thought we'd be able to move the Provider interface while making minimal changes to the API to allow injecting config. But, as seen here especially with ModelConfig, the dependency very much bleeds into the API itself, not just the implementation.

So my new plan is to keep the Provider trait in-place and start on a new one in goose-providers. We can then move provider implementation behind this new interface and delegate to that within the existing goose providers.

PR #9590, opened 2026-06-03
feat: add configurable legal disclaimer support for providers

madaosik:
## Feature purpose
Goose is increasingly adopted by organizations with legal and compliance requirements around AI-generated content. Today, any distributor needing to display legal disclaimers (e.g., warnings about AI accuracy, data privacy notices, or usage terms) must maintain source patches against the CLI output code — patches that break on every upstream update.

This change makes disclaimer messages a first-class, optional configuration in the existing custom provider JSON format. Distributors can define a startup disclaimer (shown once at session start) and a per-response disclaimer (shown after each AI response) without modifying any source code. When no disclaimers are configured, behavior is unchanged, preserving full backward compatibility.

## Implementation summary
Add optional `startup_disclaimer` and `response_disclaimer` fields to `DeclarativeProviderConfig`, allowing distributors to configure legal disclaimer messages through custom provider JSON configs without patching the source.

- `startup_disclaimer`: shown once after the session greeting
- `response_disclaimer`: shown after each agent response

When no disclaimers are configured, no messages are displayed, preserving full backward compatibility.

## Sample config for startup and response disclaimers
```
{
    "name": "custom_disclaimer",
    "engine": "openai",
    "display_name": "custom_disclaimer_demo",
    "description": "Required AI usage disclaimers",
    "api_key_env": "",
    "base_url": "http://127.0.0.1:7080",
    "base_path": "v1/chat/completions",
    "models": [
        {
            "name": "demo_model,
            "context_limit": 128000
        }
    ],
    "headers": {},
    "timeout_seconds": null,
    "supports_streaming": true,
    "requires_auth": true,
    "startup_disclaimer": "IMPORTANT: This tool uses AI technology.\nAI-generated output may be inaccurate or incomplete.\nDo not input personal or confidential data.",
    "response_disclaimer": "Always review AI-generated content prior to use."
}
```

## Example usage screenshot
goose_screenshot

### Testing
The feature has been tested manually.

### DIFF
--- a/crates/goose-cli/src/scenario_tests/scenario_runner.rs
+++ b/crates/goose-cli/src/scenario_tests/scenario_runner.rs
@@ -262,6 +262,7 @@ where
         None,
         None,
         "text".to_string(),
+        None,
     )
     .await;
 
--- a/crates/goose-cli/src/session/builder.rs
+++ b/crates/goose-cli/src/session/builder.rs
@@ -5,7 +5,7 @@ use super::CliSession;
 use console::style;
 use goose::agents::{Agent, Container, ExtensionError};
 use goose::config::resolve_extensions_for_new_session;
-use goose::config::{Config, ExtensionConfig, GooseMode};
+use goose::config::{get_provider_disclaimers, Config, ExtensionConfig, GooseMode};
 use goose::providers::create;
 use goose::recipe::Recipe;
 use goose::session::session_manager::SessionType;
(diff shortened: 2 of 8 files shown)


DOsinga:
Thanks for the contribution, @madaosik, and for the really clear write-up! This is a thoughtful idea and I can see why it'd be useful for distributors with compliance needs.

I'm going to close this one for now, but please don't read that as a no. A couple of reasons:

- There's an unaddressed [codex review comment](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#ai-code-reviews) flagging that the response disclaimer is skipped on the first turn of `goose run --interactive <prompt>`, which has been open for a couple of weeks.
- More importantly, if we want disclaimers as a first-class feature, it probably shouldn't live only in the CLI — it should work in the desktop too, and we'd want to think about the right config home for it.

That design conversation is best had before code. Would you mind opening a discussion on our [Discord](https://discord.gg/block-opensource) about how we should land this feature? If we align on the approach there, we'd be very happy to pick the implementation back up. Thanks again!

PR #9592, opened 2026-06-03
Add openapi feature flag

samdoran:
## Summary
Add a new `openapi` feature flag to allow packagers to build without the `openapi` module and the `generate_schema` binary, reducing package size.

### Testing
Tested building with and without the feature flag enabled.
```
> cargo build -p goose-server --no-default-features --features rustls-tls
```

Verified that the `generate_scheam` binary was not compiled.

### Related Issues
None

N/A

### DIFF
--- a/crates/goose-server/Cargo.toml
+++ b/crates/goose-server/Cargo.toml
@@ -19,9 +19,11 @@ default = [
     "telemetry",
     "nostr",
     "otel",
+    "openapi",
     "rustls-tls",
     "system-keyring",
 ]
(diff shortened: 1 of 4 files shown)


jh-block:
Wouldn't it be enough to just specify the binary you want to build?

```
cargo build -p goose-server --bin goosed
```

samdoran:
After looking at our build process and talking this over with some folks, your suggestion to just specify the binary is a much better approach. Thank you!

PR #9597, opened 2026-06-04
fix(deepseek): Add V4 support with reasoning_content gating to assistant messages

### LINKED ISSUE #9333: DeepSeek V4 update breaks provider — unable to use V4 models
Current issues with deepseek.json:

Model names still use V3 (deepseek-chat / deepseek-reasoner)
Context limit is set to 128K, while V4 actually supports 1M
Backend does not handle reasoning_content — multi-turn conversations always get a 400 error on the second message

hanthor:
Fixes #9333.

## Summary

- **Add V4 model names**: `deepseek-v4-flash` and `deepseek-v4-pro` with 1M token context windows, matching the current DeepSeek API
- **Fix context limits**: `deepseek-chat` and `deepseek-reasoner` (now deprecated aliases for `deepseek-v4-flash`) updated from 128K → 1M per the DeepSeek docs
- **Fix multi-turn 400 errors**: DeepSeek V4 requires every assistant message in the history to include a `reasoning_content` field, even when empty. Added a `requires_reasoning_content_on_all_messages` flag to `DeclarativeProviderConfig` and `OpenAiFormatOptions` that ensures this field is always present. Modelled after pi-coding-agent's `requiresReasoningContentOnAssistantMessages` compat flag.

The existing `preserve_thinking_context` path already omits empty `reasoning_content` to avoid breaking Kimi (which rejects empty strings); the new flag is distinct so neither provider is affected by the other's constraint.

## Test plan

- [ ] Configure goose with `DEEPSEEK_API_KEY` and select `deepseek-v4-flash`
- [ ] Run a multi-turn conversation — confirm no 400 error on the second message
- [ ] Confirm `deepseek-v4-pro` is listed as a model option
- [ ] Existing unit tests pass: `cargo test -p goose`

### DIFF
--- a/crates/goose-server/src/routes/utils.rs
+++ b/crates/goose-server/src/routes/utils.rs
@@ -309,6 +309,7 @@ mod tests {
                 setup_steps: vec![],
                 fast_model: None,
                 preserves_thinking: false,
+                requires_reasoning_content_on_all_messages: false,
             },
             is_editable: false,
         };
--- a/crates/goose/src/config/declarative_providers.rs
+++ b/crates/goose/src/config/declarative_providers.rs
@@ -105,6 +105,9 @@ pub struct DeclarativeProviderConfig {
     pub fast_model: Option<String>,
     #[serde(default)]
     pub preserves_thinking: bool,
+    /// Always emit `reasoning_content` on assistant messages (even `""`); required by DeepSeek V4.
+    #[serde(default)]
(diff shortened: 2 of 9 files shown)


hanthor:
Testes this quite a bit today and it seems to work well

hanthor:
fixed elsewhere I believe I was using an older version

PR #9601, opened 2026-06-04
fix(desktop): align session history scrollbar with window edge

bdeenyy:
Move horizontal padding inside ScrollArea so the scrollbar sits at the panel edge instead of overlapping session cards.

### Related Issues
Relates to #ISSUE_ID  
Discussion: LINK (if any)

Before:  

After:


DOsinga (approved):
Clean, minimal CSS fix — moving `px-8` from the wrapper outside the `ScrollArea` to the inner content div lets the scrollbar reach the panel edge while the cards keep their padding. Does exactly what it says. Thanks for the contribution! 🎉

PR #9607, opened 2026-06-04
Image read tool

DOsinga:
## Summary

Image read tool including cropping for zoom

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4581,6 +4581,7 @@ dependencies = [
  "icu_calendar",
  "icu_locale",
  "ignore",
+ "image 0.24.9",
  "include_dir",
  "indexmap 2.14.0",
  "indoc",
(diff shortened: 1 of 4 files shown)


michaelneale (approved):
surprising we got this far with this missing!

@DOsinga could this be a platform/default tool, separate to developer? as it is generally applicable wouldn't you say?

alexhancock (approved):

DOsinga:
@michaelneale 

> @DOsinga could this be a platform/default tool, separate to developer? as it is generally applicable wouldn't you say?

yeah, we should talk about what we want default on. I put it back into developer because that is where it was, but I agree, it might not be the right bit. I would also like a tool to read pdfs, word documents, excel and what have you. and then maybe also something to read html from the internet while converting it to markdown.

I see this got merged, so eh, here we are

Issue #9610, opened 2026-06-04
Set request params per model settings in recipe


kjanosz:
**Please explain the motivation behind the feature request.**
Right now one can only customize [temperature](https://github.com/aaif-goose/goose/blob/main/crates/goose/src/recipe/mod.rs#L107) for model responsible for driving recipe. There is no way to set however `reasoning_effort` or other (like `top_k` or `top_p` for open weights models) params. There is benefit in tweaking those per-recipe instead of going with global `GOOSE_PREDEFINED_MODELS`, as various recipes might benefit from different model tuning.

**Describe the solution you'd like**
Ideally model settings in recipe accept `request_params`.
From what I see that could be plugged in goose-clii [there](https://github.com/aaif-goose/goose/blob/main/crates/goose-cli/src/session/builder.rs#L232-L292).

**Describe alternatives you've considered**
One can workaround current limitation with having proxy between goose and provider(s) to provide model aliases so that via `GOOSE_PREDEFINED_MODELS` (or proxy alone) one could map those tweaks uniquely. That is however cumbersome and as far as understand will not work with ACP providers.

- [x] I have verified this does not duplicate an existing feature request

alexhancock:
@lifeizhou-ap Adding you on this one - do you have a thought or opinion on how to achieve this?

alexhancock:
Looking for an owner for this one

PR #9611, opened 2026-06-05
Recipe request params

kjanosz:
## Summary
This PR adds per recipe request_params along with their propagation.
Please note that slash commands are not affected - I do not see other parameters, like temperature propagated too, so decided to leave it, as I assume that was the intention. 

### Testing
Two unit tests added

### Related Issues
Relates to #9610


DOsinga:
Thanks for the PR, @kjanosz! The underlying idea — letting recipes carry provider-specific `request_params` — is reasonable, but I'm going to close this for now.

The main blocker is that the [Codex review](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#ai-code-reviews) left three unaddressed comments, and per our contributing guidelines we ask that those be resolved (or explicitly explained) before a PR moves forward. A couple of them are real correctness gaps rather than nits:

- The server/desktop recipe path (`goose-server/src/routes/agent.rs`) only propagates `goose_provider`/`goose_model` from recipe settings, so `request_params` is silently ignored when a recipe is launched from the desktop app.
- The slash-command recipe path also drops the field.
- The OpenAPI schema / generated TS types weren't regenerated for the new `Settings.request_params` field (`just generate-openapi`).

There's also an unrelated `flake.nix` change bundled in, and the branch is ~58 commits behind `main` so the generated files conflict.

Since this adds a new feature without a prior discussion, could you come hang out in our [Discord](https://discord.gg/block-opensource) and make the pitch for why we need recipe-level request params? If there's appetite for it we'll happily revisit and reopen. See [Discussions, Issues, and PRs](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#discussions-issues-and-prs) for more context.

Issue #9523, opened 2026-05-30
[BUG] Cmd+T / Ctrl+T keyboard shortcut for new chat is not working


AungMyoKyaw:
**Describe the bug**
The keyboard shortcut Cmd+T (macOS) / Ctrl+T (Windows/Linux) for starting a new chat in Goose Desktop is not working. Users have to manually click the "New Chat" button every time they want to start a new session.

**To Reproduce**
1. Open Goose Desktop
2. Start a conversation
3. Press Cmd+T (macOS) or Ctrl+T (Windows/Linux)
4. Nothing happens — the app does not navigate to the Home screen to start a new chat

**Expected behavior**
Pressing Cmd+T / Ctrl+T should take the user to the Home tab, allowing them to start a new chat session (same behavior as clicking the "New Chat" button).

**Environment**
- Goose Desktop (all platforms affected)
- Previously implemented in PR #4541 / Issue #4535

**Additional context**
- This feature was previously added via PR #4541 ("Added CMD+T keyboard shortcut that takes you to the Home tab")
- The original feature request was Issue #4535
- The shortcut appears to have regressed / stopped working in recent builds
- The only workaround is to manually click the "New Chat" button

michaelneale:
fixed with #9659

PR #4541, opened 2025-09-05
Added CMD+T keyboard shortcut that takes you to the Home tab

GuiSim:
## Pull Request Description
A new CMD/CTRL + T keyboard shortcut was added to Goose. This takes you to the Home page, allowing you to start a new chat.

Addresses https://github.com/block/goose/issues/4535

PR #9614, opened 2026-06-05
fix(desktop): restore Cmd/Ctrl+T new chat shortcut

sanidhyasin:
## Summary
The **New Chat** keyboard shortcut (`Cmd+T` on macOS / `Ctrl+T` on Windows/Linux) does nothing — users must click the "New Chat" button manually.

The shortcut chain is:

`CmdOrCtrl+T` menu accelerator → `webContents.send('new-chat')` (`main.ts`) → `handleNewChat` (`App.tsx`)

`handleNewChat` only dispatched a `trigger-new-chat` `CustomEvent`, but **nothing in the codebase listens for it** (`AppEvents.TRIGGER_NEW_CHAT` was referenced only at its definition and dispatch sites), so pressing the shortcut had no effect. This regressed the behavior originally added in #4541.

### Fix
Make `handleNewChat` navigate to the Home tab (`/`), exactly mirroring what the "New Chat" navigation item does (`handleNavClick('/')` → `navigate('/')`). Also removes the now-unused `TRIGGER_NEW_CHAT` event constant.

### Testing
- Added a regression test in `App.test.tsx` asserting the `new-chat` handler navigates to `/`.
- Did not run the desktop test suite locally (heavy Electron toolchain, no committed JS lockfile); relying on CI.

### Related Issues
Relates to #9523


DOsinga (approved):
Thanks @sanidhyasin! The actual Cmd/Ctrl+T fix (`handleNewChat` calling `navigate('/')`) already landed on main separately, so after merging main this PR narrowed down to just removing the now-unused `TRIGGER_NEW_CHAT` constant. I pushed a small commit dropping the leftover comment on `handleNewChat`, so the PR is now a clean, risk-free cleanup. Approving.

Issue #8757, opened 2026-04-22
[BUG] `just run-ui` build the electron app with empty screen


Jazzcort:
**Describe the bug**

I tried to build the desktop app using `just run-ui` to play with some changes, but the app just displays an empty screen. I followed the instructions from CONTRIBUTING.md. 

---

**To Reproduce**
1. Clone the project
2. Build the electron app from the source

---

**Expected behavior**
The app built in debug mode from the source should work the same as the release built

---

**Screenshots**

https://github.com/user-attachments/assets/b8cbb538-6849-442e-beee-0c258e587622

---

**Please provide the following information**
- **OS & Arch:** macOS arm64
- **Interface:** UI
- **Version:** main branch

DOsinga:
Thanks for reporting this! A blank screen in the Electron app typically means the goosed backend process failed to start. The UI connects to it on launch and if it's not running, you get an empty window.

Could you check the server logs? You can find them at:
- **macOS**: `~/Library/Application Support/goose/logs/`
- Or look at the terminal output from `just run-ui` for any errors during the binary build or startup.

It's often a permissions issue — e.g. the built `goosed` binary at `ui/desktop/src/bin/goosed` not being executable, or a firewall blocking the local port. The logs should tell you exactly what went wrong.

Also, make sure you activated Hermit (`source bin/activate-hermit`) before running the build so you get the right Node/pnpm/Rust versions.

Jazzcort:
@DOsinga Thanks for following up on this! Here is the log history. I did activated Hermit before running `just run-ui`. I didn't see anything that's clearly indicate an error. Let me know if you need more info to debug this. I don't know what has been changed but I could run the developing mode without this issue back in Feb.
```text
[2026-04-22 12:28:44.177] [info]  Registering update IPC handlers...
[2026-04-22 12:28:44.216] [info]  Starting goosed from: /Users/chihlee/goose/ui/desktop/src/bin/goosed on port 50462 in dir /Users/chihlee
[2026-04-22 12:28:44.218] [info]  Spawn options: {
  "env": {
    "NVM_INC": "/Users/chihlee/.nvm/versions/node/v24.14.1/include/node",
    "TERM_PROGRAM": "WezTerm",
    "NODE": "/Users/chihlee/Library/pnpm/.tools/pnpm-exe/10.33.1/pnpm",
    "INIT_CWD": "/Users/chihlee/goose/ui/desktop",
    "NVM_CD_FLAGS": "-q",
[...]

PR #9631, opened 2026-06-05
fix(ui): dedupe React to prevent multiple instances in renderer

Jazzcort:
Adds resolve aliases to pin react and react-dom to the desktop app's node_modules, preventing duplicate React instances that cause hooks and context errors.

## Summary
I've not been able to get a successful build in developer mode with `just run-ui` for about two months -> #8757
And I finally found the fix for me to successfully build it in developer mode. Btw, I'm also wondering that why it doesn't seem to happen to many contributors here. Is it just my machine (macOS 15.7.7 M3 pro) or just me. 😢 

Anyway, I found out that there were actually two React instances been created during developer mode (Not sure about the production mode but since all the release builds are working this bug might just happen to the developer build). 

```text
ui/                          ← workspace root
  ├── node_modules/
  │   ├── react/               ← Copy A
  │   └── react-intl/          ← installed here, resolves React from ../react
  └── desktop/
      └── node_modules/
          └── react/           ← Copy B (the app's React)
```
React hooks (like useRef) store state on an internal shared object. When react-intl calls useRef, it uses Copy A of React, but the app's component tree was created by Copy B. Copy A's internal state is null because no component tree was ever mounted with it — hence the error Cannot read properties of null (reading 'useRef'). This causes a blank screen when I launch my developer build. After the patch in `vite.renderer.config.mts`, it ensures there is only one React instance at runtime and all the `import 'react'` point to `ui/desktop/node_modules/react` which solves the problem!

I also checkout out both `http://localhost:5173/node_modules/.vite/deps/react.js` and `http://localhost:5173/node_modules/.vite/deps/react-intl.js` and we can see before this patch they refer to different chunks for react.

#### react-intl.js
react-intl-broken

#### react.js
react-broken

#### react-intl.js
react-intl-fixed

#### react.js
react-fixed

Here is the screen recording to show that the patch works.

https://github.com/user-attachments/assets/a4263a22-1f37-4a6a-a635-bb19909e95d4

### Testing
Run the developer build with `just run-ui` or `pnpm run start-gui` from `ui/desktop` if you already had the binary in `ui/desktop/src/bin` and everything should work like a charm instead of a blank screen. 😁 

### Related Issues
Relates to #8757

### DIFF
--- a/ui/desktop/vite.renderer.config.mts
+++ b/ui/desktop/vite.renderer.config.mts
@@ -1,5 +1,6 @@
 import { defineConfig } from 'vite';
 import tailwindcss from '@tailwindcss/vite';
+import { resolve } from 'path';
 
 // https://vitejs.dev/config
 export default defineConfig({
@@ -8,6 +9,14 @@ export default defineConfig({
   },
 
   plugins: [tailwindcss()],
+
+  resolve: {
+    dedupe: ['react', 'react-dom'],
+    alias: {
+      react: resolve(__dirname, 'node_modules/react'),
+      'react-dom': resolve(__dirname, 'node_modules/react-dom'),
+    },
+  },
 
   build: {
     target: 'esnext'


DOsinga (approved):
Thanks @Jazzcort! This is the idiomatic Vite fix for the duplicate-React situation in the workspace — `dedupe` plus the aliases force a single React instance in the renderer, which resolves the `useRef` null error and blank screen from #8757. Small, focused, dev-renderer-only change. LGTM 🚀

PR #9633, opened 2026-06-05
move formats/openai.rs into goose-providers crate, along with several dependencies

jamadeo:
This PR moves formats/openai.rs into the new goose-providers crate. Where it depended on other parts of the goose crate, those have also been moved (if they're provider-related utilities) or lightly refactored (ModelConfig -> ModelConfigParams)


crates/goose/src/agents/platform_extensions/developer/shell.rs
```diff
@@ -1049,4 +1051,30 @@ mod tests {
             "should capture output after background cmd"
         );
     }
+
+    #[test]
+    fn test_shell_tool_schema_is_valid_openai() {
```
DOsinga: why does this test live here? it seems more related to openai than to the developer tools?

crates/goose-providers/src/thinking.rs
```diff
@@ -0,0 +1,590 @@
...
+    think_depth: usize,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum ThinkTag {
+    Open,
+    Close,
+    // `<think/>` is XML-legal but carries no reasoning payload. Treat it as a
+    // no-op so we don't flip `inside_think` forever and swallow the rest of
+    // the stream into the thinking bucket.
```
DOsinga: this comment sounds like the LLM trying to explain why it made a mistake before

crates/goose-server/src/routes/errors.rs
```diff
@@ -5,7 +5,7 @@ use axum::{
 };
 use goose::config::ConfigError;
 use goose::model::ConfigError as ModelConfigError;
-use goose::providers::errors::ProviderError;
+use goose_providers::errors::ProviderError;
 use serde::Serialize;
 use utoipa::ToSchema;
```
DOsinga, line 10: not for now, but we should think about this dependency and maybe make it a featured one for callers that will want this?

crates/goose-providers/src/conversation/token_usage.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default, Copy)]
+pub struct Usage {
+    pub input_tokens: Option<i32>,
+    pub output_tokens: Option<i32>,
+    pub total_tokens: Option<i32>,
+    pub cache_read_input_tokens: Option<i32>,
+    pub cache_write_input_tokens: Option<i32>,
+}
```
DOsinga, line 33: again not for now, but this might be worth revisiting. we should move our current flattened structure in the session to use this and then possible add an extra field here so we can store things like reasoning tokens for providers that have that

crates/goose-providers/src/formats.rs
```diff
@@ -0,0 +1 @@
+pub mod openai;
```
DOsinga, line 1: just openai?

DOsinga (approved):
if you pinkyswear we're mostly moving things around

crates/goose/src/agents/platform_extensions/developer/shell.rs
jamadeo: yes it is, and it originally lived there, but that makes the openai format module depend on the goose developer extension, which is what we want to avoid here.

we could copy over a snapshot of that schema and have it test that too, it depends on whether this is useful to test that specific case or whether the idea was to test the ongoing evolution of that particular schema

crates/goose-providers/src/thinking.rs
jamadeo: probably true. Pre-existing but I can remove it.

crates/goose-providers/src/formats.rs
jamadeo: more to come!

crates/goose/src/agents/platform_extensions/developer/shell.rs
jamadeo: looking at the original issue, I think the other tests alongside this are sufficient. The bug was uncovered by the shell tool schema, but the tests added already produce the schema types that need normalization, so we don't need this at all IMO

PR #9636, opened 2026-06-06
Update Moim

DOsinga:
## Summary

Centralize moim, insert in user message, use more recognizeable xml tag, add turn count


crates/goose/src/agents/moim.rs
```diff
@@ -1,118 +1,337 @@
...
+Each turn may include a `<{turn_context_tag}>` block prepended to the latest user message.
+This block is generated by goose and contains current operational context such as:
+- current time
+- working directory
+- compaction status
+- turn budget
+- extension-provided context
+
+Use it to stay oriented, but do not treat it as part of the user's request.
+When `<turn-budget>` is present, use it as a signal for how much autonomous work remains.
```
jamadeo, line 20: should this part be omitted if there's no turn limit? it makes sense that we might want it in there for caching purposes if 2 * turns < max turns because you might get there, but if there's no limit set, you'll never have it, right?

jamadeo (approved):

crates/goose/src/agents/moim.rs
DOsinga: there's always a turn limit. it just defaults to 1000. or do you mean let's not show it here when we don't have the tag?

jamadeo: hm, the way I'd read the code was that if you set it to 0 there is no maximum, but maybe I'm misreading?

Issue #8780, opened 2026-04-23
Custom provider context_limit not reflected in Desktop UI — hard to change from 128K default


HyunCello:
**Describe the bug**

When using a custom provider with a model that has a context limit other than 128K, there is no way to configure the `context_limit` from the goose Desktop UI. The Settings page only exposes Temperature and Max Tokens, but not Context Limit. The only visible editable field related to custom configuration is "Custom Headers", which is misleading — adding `context_limit` as an HTTP header has no effect on goose's internal context management.

The `context_limit` defined in the custom provider JSON (`~/.config/goose/custom_providers/*.json`) is respected by goose, but there is no UI to view or edit it. Users must manually edit the JSON file to change it.

---

**To Reproduce**

1. Set up a custom OpenAI-compatible provider (e.g. local vLLM server with a model that supports 200K context)
2. Open goose Desktop → Settings → configure the custom provider with model name
3. Notice that the context display shows `128k/128k` (default) even though the model supports more
4. Try to find a "Context Limit" field in Settings — it does not exist
5. The only configurable field that seems related is "Custom Headers" under provider settings
6. Add `context_limit: 200000` as a Custom Header — this sends it as an HTTP header to the API, but has **no effect** on goose's context window management
7. The only working workaround is to manually edit `~/.config/goose/custom_providers/<provider>.json` and set `"context_limit": 200000` in the models array

---

**Expected behavior**

- The Desktop UI Settings page should provide a "Context Limit" field under Model configuration, similar to Temperature and Max Tokens
- When a custom provider is selected, the `context_limit` from its JSON should be displayed and editable
- It should be clear that "Custom Headers" is for HTTP request headers only, not for goose configuration settings

---

**Screenshots**

N/A

---

**Please provide the following information**

- **OS & Arch:** Ubuntu 24.04 x86_64
- **Interface:** UI (goose Desktop)
- **Version:** v1.32.0
- **Extensions enabled:** Developer, Apps, Analyze, Summon, Todo, Top Of Mind, Extension Manager, Skills
- **Provider & Model:** Custom (custom_aeirobot) – GLM-5.1-FP8

---

**Additional context**

The `context_limit` field in the custom provider JSON works correctly once set manually:

```json
{
  "name": "custom_aeirobot",
  "models": [
    {
      "name": "GLM-5.1-FP8",
      "context_limit": 200000
    }
  ]
}
```
However, this requires the user to:

1. Know that this JSON file exists
2. Know the correct field name and location
[...]

PR #9639, opened 2026-06-06
fix: correctly map ollama_cloud to canonical provider and update max_…

murphysean:
## Summary
This PR fixes an issue where `ollama_cloud` providers were not correctly mapped to their canonical model names, causing them to fall back to the default context limit (128K) instead of the model's specific capacity.

**Note on the `max_tokens` logic:**
I initially attempted to change the `max_tokens` filter from `<` to `<=` to ensure canonical output limits were applied. However, it was discovered that for models where `output_limit == context_limit`, setting `max_tokens` to the full window results in `Prompt Tokens + Max Tokens > Total Context`, which causes API providers to reject the request. The strict `<` filter is necessary as a safety guard to let the provider dynamically calculate the available output space.

### Testing
**Manual Testing:**
- Launched the CLI using `GOOSE_MODEL=gemma4:31b goose`.
- Verified that the status bar correctly displays the expanded context window (e.g., `0/262k`) instead of the 128k default.
- Confirmed that requests to the model do not trigger context window overflow errors.

### Related Issues
Relates to #8780

### DIFF
--- a/crates/goose-providers/src/canonical/name_builder.rs
+++ b/crates/goose-providers/src/canonical/name_builder.rs
@@ -49,6 +49,7 @@ pub fn map_provider_name(provider: &str) -> &str {
         "zhipu" => "zhipuai",
         "novita" => "novita-ai",
         "opencode_go" => "opencode-go",
+        "ollama_cloud" => "ollama-cloud",
         _ => provider,
     }
 }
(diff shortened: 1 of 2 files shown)


michaelneale:
taking a look.

michaelneale (approved):
thanks!

Issue #9634, opened 2026-06-05
Add a supply-chain inspector (malicious package installs) to the security pipeline


AyoubTadlaoui:
### The gap

Goose's tool-inspection pipeline catches prompt/command injection, data egress, and adversarial patterns, but nothing checks the package an install command is pulling in. If the agent runs `npm install <pkg>` and `<pkg>` is a typosquat or a known-bad name, it just runs it. That's one of the more common ways an autonomous coding agent gets compromised.

### Where it fits

`ToolInspector` (`crates/goose/src/tool_inspection.rs`) is the extension point, and the inspectors get registered in `agents/agent.rs`:

```rust
add_inspector(Box::new(SecurityInspector::new()));
add_inspector(Box::new(EgressInspector::new()));
add_inspector(Box::new(AdversaryInspector::new(provider.clone())));
```

`EgressInspector` already matches shell tool calls (`"shell" | "bash" | "execute_command" | "run_command" | "terminal"`), so inspecting shell commands is established precedent.

### Proposal

A fourth inspector, `SupplyChainInspector`, that:

1. Matches the same shell tool calls.
2. Parses package install and runner commands (npm/yarn/pnpm/bun install/add, npx/bunx/pnpm dlx/yarn dlx/bun x) and pulls out the target package name(s).
3. Runs a deterministic, offline check, starting with typosquat detection against a bundled list of popular names, and returns `RequireApproval` with a reason when a name looks like a typosquat.

It's deterministic and needs no network, so it adds no latency or new failure modes, and it would sit behind its own `SECURITY_*` flag like the existing scanners.

### Scope

v1 is just the parser plus typosquat detection, to keep it small and reviewable. An optional OSV malicious-package lookup for confirmed-malware blocking is a sensible follow-up, but it needs a network call so I'd keep it out of the first PR.

### Questions

- Default on, or off behind a flag like the prompt scanner?
- Is that tool-name set the right match surface, or is there a canonical list I should reference?
- Where should a bundled popular-name list live?

I already have the command parser and typosquat logic as pure tested functions (from a Rust npm-security tool I maintain), so I can port them to fit the inspector trait, with tests. Happy to implement end to end if it's something you'd welcome.

jamadeo:
Thank you for the suggestion @AyoubTadlaoui. This is something you could do with a PreToolUse hook, and we don't plan on further expanding any built-in security tools in core goose, I'll close this this as not planned. It sounds like something that could be useful for some users, but should be implemented separately.

PR #9642, opened 2026-06-06
Add a supply-chain inspector for typosquat package installs

AyoubTadlaoui:
## What

Adds a `SupplyChainInspector` to the tool-inspection security pipeline. The existing inspectors look at *how* a command behaves (injection, egress); this one looks at *what package it installs*.

When the agent runs a shell command that installs or fetch-executes an npm package, the inspector extracts the target package name(s) and checks them against a deterministic, offline typosquat heuristic: a name one edit away from a well-known package (`lodahs` for `lodash`, `exprcss` for `express`) is a classic supply-chain attack. A match returns `RequireApproval` so the user can confirm before installing.

Covers `npm`/`yarn`/`pnpm`/`bun` `install`/`add` and the `npx`/`bunx`/`pnpm dlx`/`yarn dlx`/`bun x` runners. For a runner, only the executed package is checked, not its arguments (`npx create-react-app my-app` checks `create-react-app`, not `my-app`).

## Design

- A new `ToolInspector` impl, registered alongside the security and egress inspectors in `agent.rs`, matching the same shell tool set the egress inspector uses.
- Deterministic and offline: no network call, no latency, and **no new dependencies** (the edit distance is implemented inline). It complements the pattern/ML scanners rather than overlapping them.
- Returns `RequireApproval` (asks the user), never a hard `Deny`, since a single edit is a strong but not certain signal.

## Tests

10 unit tests covering the command parser, the typosquat heuristic, and the inspector end to end: a typosquat install asks for approval, a legitimate install passes, non-shell tools and non-install commands are ignored, and runner typosquats are caught. `cargo test`, `cargo fmt --check`, and `cargo clippy` are all clean.

## Open questions (from #9634)

- Default on, or gated behind a `SECURITY_*` flag like the prompt scanner? It is currently always-on but approval-only, so it is non-intrusive. Easy to move behind a flag if you prefer opt-in.
- Is the `["shell","bash","execute_command","run_command","terminal"]` tool set the right match surface, or is there a canonical list?
- Where should the popular-name list live longer term?

Implements #9634. An OSV malicious-package lookup for confirmed-malware blocking is a natural follow-up, kept out of this PR to stay deterministic and dependency-free.

### DIFF
--- a/crates/goose/src/agents/agent.rs
+++ b/crates/goose/src/agents/agent.rs
@@ -49,6 +49,7 @@ use crate::scheduler_trait::SchedulerTrait;
 use crate::security::adversary_inspector::AdversaryInspector;
 use crate::security::egress_inspector::EgressInspector;
 use crate::security::security_inspector::SecurityInspector;
+use crate::security::supply_chain_inspector::SupplyChainInspector;
 use crate::session::extension_data::{EnabledExtensionsState, ExtensionState};
 use crate::session::{Session, SessionManager, SessionNameUpdate};
 use crate::tool_inspection::ToolInspectionManager;
@@ -540,6 +541,9 @@ impl Agent {
         // Add security inspector (highest priority - runs first)
         tool_inspection_manager.add_inspector(Box::new(SecurityInspector::new()));
         tool_inspection_manager.add_inspector(Box::new(EgressInspector::new()));
+
+        // Add supply-chain inspector (flags likely-typosquat package installs)
+        tool_inspection_manager.add_inspector(Box::new(SupplyChainInspector::new()));
 
         // Add adversary inspector (LLM-based review, enabled by ~/.config/goose/adversary.md)
         tool_inspection_manager.add_inspector(Box::new(AdversaryInspector::new(provider.clone())));
--- a/crates/goose/src/security/mod.rs
+++ b/crates/goose/src/security/mod.rs
@@ -4,6 +4,7 @@ pub mod egress_inspector;
 pub mod patterns;
 pub mod scanner;
 pub mod security_inspector;
+pub mod supply_chain_inspector;
 
 use crate::config::Config;
 use crate::conversation::message::{Message, ToolRequest};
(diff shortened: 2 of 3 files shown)


AyoubTadlaoui:
quick status update: CI is green here, and the Codex review feedback was addressed in 67b8c6b (handles the wrapper-operand evasion case and adds tests for it, suite passing). no rush at all, but this is ready for a human look whenever a maintainer has the bandwidth. happy to adjust anything if it doesn't fit how you'd like supply-chain checks structured.

DOsinga:
Thanks for this — and genuinely, the code here is thoughtful: the parser handles real evasion cases, the tests check behavior rather than implementation, and keeping it offline/dependency-free was the right call.

That said, I'm going to close this one. A few reasons:

- This is a large (~940 line) single-feature PR as a first contribution. We ask first-time contributors to start smaller so we can build a working relationship before taking on big features — see https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#getting-started

- The feature landed without prior maintainer alignment. The tracking issue (#9634) was opened a day before the PR and never got a maintainer response, and it raises real open design questions (default-on vs. flag-gated, where the popular-name list should live, what the canonical shell tool set is) that we'd want to settle *before* implementation. See https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#discussions-issues-and-prs

- There are also unaddressed Codex review comments — specifically the `npm exec -w` runner-options case and the `pkg@npm:typosquat` alias-target case both describe ways a typosquat slips past the check. See https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#ai-code-reviews

None of this is a knock on the quality of the work. If you'd like to pursue this, the best path is to open a focused discussion on the design first, and consider breaking it into smaller pieces. Thanks again for taking the time.

AyoubTadlaoui:
@DOsinga  Thanks, that’s helpful context.

  Before I open a separate design discussion, I want to make sure I’m asking the right questions rather than guessing at the maintainer preference.
  From your note, I’m reading the design decisions as:
  - whether this belongs in goose at all, and if so whether it should start default-off / flag-gated
  - where the popular-package name list should live and how it should be maintained
  - which package-manager command forms are in scope for a first version
  - how runner options like npm exec -w and scoped alias targets like pkg@npm:typosquat should be modeled

  My current strawman would be: start flag-gated, keep the package list small and local to the inspector until there’s agreement on maintenance, support only the canonical npm install/exec forms first, and explicitly defer broader package-manager coverage.

 Is that the right shape for a design discussion, or is there a different direction maintainers would prefer me to start from?

PR #9649, opened 2026-06-06
fix(providers/openai): preserve custom API path in base_url derivation

SjB:
Previously, custom API paths like `/zen` or `/zen/go` were incorrectly handled when deriving the base path, causing them to be stripped instead of preserved. Now the URL correctly appends `/v1/chat/completions` to custom paths while still handling existing `/v1` and `/v1/chat/completions` suffixes.

## Summary
if we have a base url not terminating in 'v1' to make sure to append the original base path and the 'v1/chat/completions' to the url

### Testing
I have added 3 test to validate the changes


DOsinga:
Thanks for the fix! I removed the redundant comment flagged by codex and merged the latest `main` into the branch. Tests pass and clippy is clean for the touched file. This looks good to go.

DOsinga (approved):

DOsinga:
Addressed the P2: `derive_base_path` now detects any `/vN` version segment (not just `/v1`) and appends only `chat/completions` instead of an extra `/v1/chat/completions`. So Zhipu's default `/api/paas/v4` now correctly resolves to `/api/paas/v4/chat/completions`. Added regression tests (`derive_base_path_preserves_non_v1_version_prefix` and `derive_base_path_does_not_treat_v_word_as_version`).

PR #9650, opened 2026-06-06
fix(summon): give GOOSE_SUBAGENT_PROVIDER/MODEL priority over LLM-injected params

### LINKED ISSUE #9644: GOOSE_SUBAGENT_PROVIDER is ignored when orchestrator LLM injects its own provider into delegate() calls
**Describe the bug**

When a recipe's instructions omit `provider`/`model` from `delegate()` blocks, the orchestrator LLM infers and injects its own session's provider/model into the tool call. This causes `params.provider` to be non-None, winning at priority 1 in `resolve_provider` and silently bypassing `GOOSE_SUBAGENT_PROVIDER`. All subagents run on the orchestrator's provider/model regardless of the env var.

Traced to `build_recipe_from_agent` (`summon.rs:1332`): `goose_provider: params.provider.clone()` propagates the confabulated value directly into the agent recipe settings.

---

**To Reproduce**

1. Set `GOOSE_SUBAGENT_PROVIDER=aws_bedrock` and `GOOSE_SUBAGENT_MODEL=global.anthropic.claude-haiku-4-5-20251001-v1:0` in shell environment.
2. Run a recipe whose instructions contain `delegate()` JSON blocks with no `provider` or `model` fields.
3. Observe all subagents spawn on the orchestrator's provider instead.
4. Confirm: `SELECT provider_name, model_config_json FROM sessions WHERE session_type = 'SubAgent'` shows orchestrator provider on every row.

---

**Expected behavior**

When `delegate()` omits `provider`/`model`, resolution falls through to `GOOSE_SUBAGENT_PROVIDER`/`GOOSE_SUBAGENT_MODEL`. The env var should be respected.

---

- **OS & Arch:** macOS arm64
- **Interface:** CLI
- **Version:** 1.37.0
- **Extensions enabled:** summon
- **Provider & Model:** `gcp_vertex_ai` / `claude-sonnet-4-6@default` (orchestrator); intended subagent `aws_bedrock` / `global.anthropic.claude-haiku-4-5-20251001-v1:0`

---

**Additional context**

`GOOSE_SUBAGENT_MAX_TURNS` does not have this problem because it uses `std::env::var` directly (`summon.rs:1485`) rather than routing through `params`. The provider/model path goes through `params` first, making it susceptible to LLM field injection.

nuthalapativarun:
## Summary

When the orchestrator LLM injects its own provider and model into a `delegate()` call, those values can mask the user's subagent environment overrides.

This change gives real environment variables priority while keeping persisted config-file defaults below explicit delegate and recipe settings:

```
GOOSE_SUBAGENT_PROVIDER environment > params.provider > recipe.settings > persisted config > session
GOOSE_SUBAGENT_MODEL environment    > params.model    > recipe.settings > persisted config
```

Model values are used only when their provider context matches the selected provider. This keeps explicit models when `GOOSE_SUBAGENT_PROVIDER` names the same provider, discards models paired with a different provider, and falls back to the selected provider's registry default. Dynamic providers without a static default now return a clear error asking for `GOOSE_SUBAGENT_MODEL`.

The branch is merged with current `main`.

## Verification

- `cargo test -p goose test_resolve_model_config` — 8 passed
- `cargo fmt --all`

## Related issue

Closes #9644


clouatre (approved):
The fix is correct. Both `or_else` chains now check `GOOSE_SUBAGENT_PROVIDER` and `GOOSE_SUBAGENT_MODEL` before `params.*`, which is the right priority order.

One minor observation, no action required: `build_recipe_from_agent` still copies `params.provider` into `recipe.settings.goose_provider` (line 1332). This is harmless -- `resolve_provider` now reads the env var first and `params.provider` second, so `recipe.settings` is never reached when both hold the same confabulated value. But the slot is misleading since it looks like an authoritative recipe setting. Worth a follow-up cleanup.

nuthalapativarun:
Good catch — agreed it's dead weight now. `resolve_provider` reads the env var first and `params.provider` second, so `recipe.settings.goose_provider` is never actually consulted in this path anymore. I'll handle that cleanup as a quick follow-up separate from this fix to keep this PR scoped to the priority-ordering bug.

nuthalapativarun:
Hi maintainers, just a friendly bump on this one — it's been a week without a formal review. @clouatre gave informal approval with a minor non-blocking note (tracked as a separate follow-up), so I believe this is otherwise ready. Would appreciate a maintainer look when you have a chance. Thanks!

nuthalapativarun:
Gentle bump on this — happy to make any further adjustments if there's outstanding feedback beyond what's already been addressed.

nuthalapativarun:
Hi team — following up on PR #9650 (GOOSE_SUBAGENT_PROVIDER/MODEL priority fix). @clouatre left an approval with a minor non-blocking note. CI is green. This is the 2nd bump — first was 2026-06-23. Happy to address any feedback from the core maintainers. Thanks!

DOsinga (approved):
Reviewed the final provider/model precedence logic and regression coverage. Environment overrides win, same-provider explicit models are preserved, cross-provider models are rejected, and dynamic providers without defaults fail clearly. Focused tests and local clippy pass.

PR #9651, opened 2026-06-06
fix(ui): purge empty phantom sessions on startup

### LINKED ISSUE #9608: Desktop app creates empty 'New Chat' sessions on every reconnection attempt when goosed is unreachable
## Bug Description

The Goose desktop app (Electron) creates a new empty session in the database every time the renderer reloads while the `goosed` backend is unreachable. This results in dozens or hundreds of phantom "New Chat" sessions accumulating over time.

## Root Cause

When the connection between the Electron renderer and `goosed` is lost (e.g., after sleep/wake, goosed restart, or network interruption):

1. The renderer periodically reloads (fires "React ready event")
2. Each reload initializes the React app, which eagerly creates a new session record in the SQLite database
3. The app then attempts to connect to goosed and fails with `Error: Not connected` (specifically in `notification()` → `sendHostContextChange()` → `setHostContext()`)
4. Since no messages are ever sent, the session remains empty with the default name "New Chat"
5. The app retries every 5-10 minutes, creating a new empty session each time

## Evidence from logs

```
[2026-06-03 17:02:52.824] [info]  React ready event received
[2026-06-03 17:02:52.840] [info]  from renderer: [UNHANDLED REJECTION] Error: Not connected
[2026-06-03 17:07:59.530] [info]  React ready event received
[2026-06-03 17:07:59.558] [info]  from renderer: [UNHANDLED REJECTION] Error: Not connected
[2026-06-03 17:12:18.634] [info]  React ready event received
[2026-06-03 17:12:18.653] [info]  from renderer: [UNHANDLED REJECTION] Error: Not connected
```

This pattern repeated **151 times** in the log, correlating with **143 empty sessions** found in the database (the difference being legitimate window opens).

## Additional observations

- Sessions always appear in **pairs** at the same timestamp (two `setHostContext` notifications fire per reload)
- The issue began when a `goosed` server died and was not restarted (e.g., over a weekend), but the desktop app remained open
- All ghost sessions have `working_dir = ~/` and zero messages
- The `goose session remove --regex` command does not match on session names (only IDs), making cleanup difficult

## Expected Behavior

- The app should **not** create a session record in the database until a connection to goosed is successfully established AND the user sends their first message
- Alternatively, empty sessions (0 messages) should be cleaned up automatically on app startup or after a timeout
- The reconnection loop should not trigger full React re-initialization that creates new sessions

## Workaround

Users can manually clean up with:
```sql
sqlite3 ~/.local/share/goose/sessions/sessions.db "
DELETE FROM sessions 
WHERE name = 'New Chat' 
AND id NOT IN (SELECT DISTINCT session_id FROM messages);
"
```

## Environment

- Goose v1.33.1-block (macOS desktop app)
- macOS with Electron renderer
- Issue has been occurring since at least April 21, 2026

nuthalapativarun:
## Summary

When the Electron renderer reloads while `goosed` is unreachable (e.g. after the backend dies overnight), each reload can create an empty session record — `message_count=0`, no recipe, `user_set_name` not set. Over time these accumulate as phantom \"New Chat\" entries with no user value. One reporter found 143 phantom sessions after a weekend with a dead backend.

On `App` mount, silently delete all sessions that satisfy the empty-chat criteria (`message_count === 0 && !user_set_name && !recipe`). The deletions are fire-and-forget so they never block startup or surface errors to the user.

**Before:** 143 phantom \"New Chat\" sessions accumulate and clutter chat history.  
**After:** Phantom sessions are swept on every startup.

### Testing

- Manual: confirmed the filter correctly identifies sessions with `message_count=0`, no `user_set_name`, and no `recipe`
- No existing test coverage for this path; the fix is a single fire-and-forget effect with no new state

### Related Issues

Closes #9608


michaelneale (approved):
yeah I think those codex comments are a bit of nonsense, so approving this. nice one.

DOsinga:
> yeah I think those codex comments are a bit of nonsense, so approving this. nice one.

I'm not sure. they seem reasonable to me. did we test this code with a recipe from a deeplink? new session?

@nuthalapativarun

michaelneale:
@nuthalapativarun might need to merge in main into this and resolve the desktop lint issues if you get a minute (I think that is failing on the merge queue)

nuthalapativarun:
Gentle ping — this was approved by @michaelneale. Happy to address any remaining concerns before merge.

nuthalapativarun:
Friendly ping — this is still approved and mergeable whenever you get a chance. Thanks for your patience!

PR #9654, opened 2026-06-07
feat(server): cache provider model lists with a short TTL

jezweb:
## What

`GET /config/providers/{name}/models` creates the provider and fetches its `/models` endpoint **over the network on every call**. The desktop model picker queries this for **each configured provider** every time it opens, so with several providers wired the picker can take many seconds to populate.

This adds a small in-process cache: each provider's fetched model list is cached for 5 minutes, so repeat picker opens are instant.

## Behaviour

- First fetch per provider hits the network as before and is cached.
- Subsequent calls within 5 min return the cached list (no network round-trip).
- Entries are **invalidated per provider** in `cleanup_provider_cache` — which is already called when a provider's config changes — so adding or changing an API key still triggers a fresh fetch immediately.

## Why

Reported by a user with ~8 providers connected: the picker took ~6s to load every time because each provider's `/models` is fetched fresh and one slow/large provider (e.g. OpenRouter's big catalogue) gates the whole list. Model catalogues change rarely, so a short TTL is a safe, large win.

## Scope

One file (`crates/goose-server/src/routes/config_management.rs`): a `std::sync::OnceLock` + `Mutex<HashMap>` cache (no new dependency), a read/populate in `get_provider_models`, and an invalidation line in `cleanup_provider_cache`. `cargo check -p goose-server` passes.

### DIFF
--- a/crates/goose-server/src/routes/config_management.rs
+++ b/crates/goose-server/src/routes/config_management.rs
@@ -907,6 +907,22 @@ pub async fn providers() -> Result<Json<Vec<ProviderDetails>>, ErrorResponse> {
     Ok(Json(providers_response))
 }
 
+// Per-provider cache of fetched model lists. Listing models hits each provider's
+// `/models` endpoint over the network, so with several providers configured the
+// model picker can take many seconds on every open. Cache each provider's list
+// for a short TTL so repeat opens are instant. Entries are invalidated per
+// provider by `cleanup_provider_cache` (called when a provider's config changes),
+// so a freshly-added API key still triggers a refetch.
+const MODELS_CACHE_TTL: std::time::Duration = std::time::Duration::from_secs(300);
+
+type ModelsCache =
+    std::sync::Mutex<std::collections::HashMap<String, (std::time::Instant, Vec<ModelInfo>)>>;
+
+fn provider_models_cache() -> &'static ModelsCache {
+    static CACHE: std::sync::OnceLock<ModelsCache> = std::sync::OnceLock::new();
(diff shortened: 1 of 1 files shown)


DOsinga:
Thanks for this @jezweb, and sorry for the slow response.

We're going to close this one, but the reasoning is more about timing and placement than the idea itself.

The core problem is that the cache lives at the HTTP route level, which is the wrong layer. `get_provider_models` builds a fresh provider on every request, and the natural home for any model-list caching is the providers module in the `goose` crate, keyed by provider, with invalidation tied to the existing `cleanup_provider` hook so that all callers (the ACP path, secret deletion, the cleanup route) get correct invalidation for free.

There's also a correctness issue that Codex flagged: the desktop save flow saves the new config and then calls `getProviderModels` *as the validation step*, rolling back if it fails. With a TTL cache that doesn't get invalidated on `upsert_config`, a freshly-changed API key can validate against a stale cached list and silently pass. Moving the cache down doesn't fix that on its own — the validation call would need to bypass the cache.

More importantly, we're in the middle of reworking how providers are structured, so anything built around the current provider/route layout is likely to get churned. Rather than land this and redo it, we'd rather wait for that to settle and then think holistically about what's reasonable to cache and where.

cc @jamadeo — once the provider rework lands it'd be worth revisiting model-list caching as part of that, at the provider level.

@jezweb if you're not already, please come hang out in our Discord (https://discord.gg/block-opensource) — it's the best place to chat through ideas like this before investing build time. We'd genuinely welcome a revisit once the provider work settles.

Issue #9646, opened 2026-06-06
"GOOSE_TOOLING_MODEL config is ignored — Gemini hardcoded for session naming"


Nihalreddy0226:
## Describe the bug

Even when `GOOSE_TOOLING_MODEL: openrouter/owl-alpha` is explicitly set 
in config.yaml, Goose still makes a secondary call to Gemini 2.5 Flash 
(Google Vertex) for every session — ignoring the config entirely.

## To Reproduce

1. Set provider to OpenRouter with model `openrouter/owl-alpha`
2. Add `GOOSE_TOOLING_MODEL: openrouter/owl-alpha` to config.yaml
3. Add `GOOSE_MODE: chat` to config.yaml
4. Fully restart Goose
5. Send any message e.g. "Hello!"
6. Check OpenRouter logs

## Expected behavior

Only `openrouter/owl-alpha` should be called. 
`GOOSE_TOOLING_MODEL` config should be respected.

## Actual behavior

Two API calls fire simultaneously every session:
- Owl Alpha → ~4,100 tokens (main response) → $0.00
- Gemini 2.5 Flash → ~117 tokens (unknown purpose) → $0.000189

## Config used

```yaml
GOOSE_MODE: chat
GOOSE_TOOLING_MODEL: openrouter/owl-alpha
GOOSE_MODEL: openrouter/owl-alpha
GOOSE_PROVIDER: openrouter
```

## Environment
- Goose version: 0.68.0
- OS: Windows 11
- Provider: OpenRouter

DOsinga:
This has been fixed. The supported configuration key for the model used by auxiliary operations such as session-title generation is `GOOSE_FAST_MODEL`, not `GOOSE_TOOLING_MODEL` (see #9295 and #9296). Current Goose reads `GOOSE_FAST_MODEL` through the unified configuration path, so it can be set in `config.yaml`, for example:

```yaml
GOOSE_FAST_MODEL: "your-preferred-model"
```

If you want to avoid the additional session-title request entirely, set:

```yaml
GOOSE_DISABLE_SESSION_NAMING: true
```

The unified config behavior landed as part of #9953, superseding the closed implementation attempt in #9655.

PR #9655, opened 2026-06-07
fix(model): read GOOSE_FAST_MODEL from config.yaml

nuthalapativarun:
## Summary

`ModelConfig::with_fast` only checked the raw `GOOSE_FAST_MODEL` environment variable via `std::env::var`, so users who set it in `config.yaml` (the way `GOOSE_CONTEXT_LIMIT` and other settings already support) had it silently ignored — the provider's hardcoded default fast model (e.g. `google/gemini-2.5-flash` for OpenRouter) was used instead for auxiliary calls like session naming.

This routes the lookup through `Config::global().get_param()`, which checks both the environment variable and `config.yaml`, matching the precedent already established for `GOOSE_CONTEXT_LIMIT` (see the comment near `ModelConfig::new_base`).

Also clarifies the environment variables guide to mention that `GOOSE_FAST_MODEL` can be set in `config.yaml`.

### Testing
Added `test_with_fast_uses_goose_fast_model_override` and `test_with_fast_falls_back_to_provider_default` to `crates/goose/src/model.rs`, mirroring the existing `GOOSE_MAX_TOKENS` env-var tests. Ran `cargo test -p goose --lib model::` — all 42 tests pass. `cargo clippy -p goose --lib --tests -- -D warnings` reports no issues for the changed files.

### Related Issues
Relates to #9646

### DIFF
--- a/crates/goose/src/model.rs
+++ b/crates/goose/src/model.rs
@@ -372,7 +372,11 @@ impl ModelConfig {
         fast_model_name: &str,
         provider_name: &str,
     ) -> Result<Self, ConfigError> {
-        let name = std::env::var("GOOSE_FAST_MODEL")
+        // Using Config::global().get_param() reads from both environment
+        // variables and config.yaml, so users can set `GOOSE_FAST_MODEL: ...`
+        // in config.yaml instead of exporting an env var. See #9646.
+        let name = crate::config::Config::global()
+            .get_param::<String>("GOOSE_FAST_MODEL")
             .ok()
             .map(|v| v.trim().to_string())
             .filter(|v| !v.is_empty())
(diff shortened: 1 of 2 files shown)


nuthalapativarun:
Gentle ping — happy to address any feedback if there are concerns.

DOsinga:
Thanks for the fix, and apologies for the slow response.

A couple of things going on here:

1. You currently have a large number of PRs open simultaneously (14 open, none merged yet). As a first-time contributor we'd ask you to pick your single highest-priority change and let it land before opening more — see https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#pull-requests. Opening many PRs in quick succession makes it hard for us to build the trust signal that helps us review confidently.

2. More importantly for this specific change: we're in the middle of reworking how providers are configured, and part of that is moving all of the environment-variable / settings handling out of the per-model code (`ModelConfig`) into a unified configuration layer. So fixes that add another individual env-var/config lookup in `model.rs` are likely to conflict with that work rather than build on it.

Closing for now — please feel free to reopen or resubmit once the provider rework has landed if it's still relevant. Thanks again!

PR #9656, opened 2026-06-07
fix(providers): read context limit from local server's meta.n_ctx

nuthalapativarun:
## Summary

When pointing the OpenAI-compatible provider at a local llama.cpp or Ollama server, the model isn't in the canonical model registry, so `context_limit` stays unset and falls back to `DEFAULT_CONTEXT_LIMIT` (128K). This causes `check_if_compaction_needed` to compute usage against 128K while the server hard-errors at its actual (often much smaller) context window — auto-compaction never fires in time and sessions crash with a context length error.

llama.cpp and Ollama expose the real, currently-allocated context window via a non-standard `meta.n_ctx` field on `/v1/models`, which isn't part of the OpenAI spec and so was previously ignored entirely.

This adds `OpenAiProvider::fetch_context_limit_from_api`, which queries the models endpoint and reads `meta.n_ctx` for the configured model. It's only consulted when no context limit is otherwise configured (env var, `config.yaml`, or canonical registry), so behavior for known cloud models is unchanged.

### Testing
Added `fetch_context_limit_from_api_reads_meta_n_ctx` and `fetch_context_limit_from_api_returns_none_without_meta` using `wiremock`, following the existing test patterns in `providers::openai::tests`. Ran `cargo test -p goose --lib providers::openai::` — all 36 tests pass. `cargo clippy -p goose --lib --tests -- -D warnings` reports no issues for the changed file.

### Related Issues
Relates to #9529

### DIFF
--- a/crates/goose/src/providers/openai.rs
+++ b/crates/goose/src/providers/openai.rs
@@ -264,6 +264,21 @@ impl OpenAiProvider {
             api_client = api_client.with_headers(header_map)?;
         }
 
+        // Local servers (llama.cpp, Ollama) aren't in the canonical model
+        // registry, so context_limit is still unset at this point. Ask the
+        // server for its actual runtime context window rather than falling
+        // back to DEFAULT_CONTEXT_LIMIT. See #9529.
+        let model = if model.context_limit.is_none() {
+            match Self::fetch_context_limit_from_api(&api_client, &base_path, &model.model_name)
+                .await
+            {
+                Some(n_ctx) if n_ctx > 0 => model.with_context_limit(Some(n_ctx)),
+                _ => model,
+            }
+        } else {
+            model
+        };
+
         Ok(Self {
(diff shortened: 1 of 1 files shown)


nuthalapativarun:
Gentle ping — any chance this could get a review? Happy to address any feedback.

DOsinga:
Thank you so much for this — it's a clean, well-tested fix for a real problem (and nicely targeted at the right provider path). The trouble is that several contributors converged on this exact same `meta.n_ctx` fix at once, so we had to pick one to land to avoid duplicate work. We're going with #9530, which was opened by the author of the originating issue (#9529) and has already been through review.

This isn't a reflection on your code — it's genuinely good, and we really appreciate you following the existing test patterns with wiremock. Please don't let this discourage you; we'd love to see more contributions from you. Closing as a duplicate of #9530.

PR #9664, opened 2026-06-07
fix(local-inference): recover tool calls emitted as JSON content

macgenerate:
## Problem

Several local GGUF models — Qwen2.5-Coder among them — emit tool calls as a JSON object in their **content** stream (commonly ```json-fenced or wrapped in `<tool_call>` tags) rather than through the native OpenAI-compatible `tool_calls` channel that `generate_with_native_tools` reads. The streaming parser accumulates zero structured calls, so no `ToolRequest` is emitted and the call surfaces to the user as plain assistant text — the agent never executes it.

Reproducible with the developer extension loaded (`goose acp --with-builtin developer`): a Qwen2.5-Coder-7B-Instruct turn asked to write a file produces a correct call —

```json
{"name": "write", "arguments": {"path": "/tmp/spike.txt", "content": "hello"}}
```

— right tool name, right args, but nothing runs.

## Fix

When the native channel yields no tool calls, scan the raw generated text for `{"name","arguments"}` objects and convert each into a `ToolRequest`. A brace-matching scanner respects string literals and escapes (so `}` inside a string value doesn't end the object early) and ignores surrounding prose, code fences, and tags. `arguments` may be a JSON object or a JSON-encoded string; `parameters` is accepted as an alias.

The fallback is **guarded on the native channel being empty**, so models that already emit structured `tool_calls` are completely unaffected.

## Testing

10 new unit tests cover fenced / bare / `<tool_call>`-tagged input, prose-wrapped JSON, string-encoded arguments, the `parameters` alias, multiple calls in one message, braces inside string values, and non-tool JSON (ignored). All 18 tests in the module pass (`cargo test -p goose --features local-inference inference_native_tools`).

Verified end-to-end on macOS/Metal with a local Qwen2.5-Coder-7B GGUF: the `write` tool now executes and the turn terminates cleanly. No change to the native grammar path.

### DIFF
--- a/crates/goose/src/providers/local_inference/llamacpp/inference_native_tools.rs
+++ b/crates/goose/src/providers/local_inference/llamacpp/inference_native_tools.rs
@@ -194,7 +194,17 @@ pub(super) fn generate_with_native_tools(
     //   1. Extract thinking and attach it to per-tool-request messages
     //   2. Enable merge_split_tool_call_messages to reconstruct the standard
     //      OpenAI format (one assistant msg with N tool_calls, then N tool results)
-    let tool_call_contents = extract_oai_tool_call_contents(&accumulated_tool_calls);
+    let mut tool_call_contents = extract_oai_tool_call_contents(&accumulated_tool_calls);
+    // Fallback: several local GGUF models (e.g. Qwen2.5-Coder) emit tool calls as
+    // JSON in the *content* stream rather than through the native `tool_calls`
+    // channel, so the streaming parser accumulates zero structured calls and the
+    // call leaks to the user as plain assistant text — the agent never executes
+    // it. Recover those by scanning the raw generated text for `{"name",
+    // "arguments"}` objects. Guarded on the native channel being empty so
+    // well-behaved models are never double-parsed.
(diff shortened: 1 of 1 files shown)


DOsinga:
Thanks for digging into the local-inference tool-call recovery — this is a real problem worth solving.

A couple of things though. This is a fairly large first contribution, and codex left two P2 review comments (the recovered JSON still being streamed as assistant text before the fallback runs, and malformed string `arguments` producing a zero-arg call) that have been open for over a week without a response. Per our [AI Code Reviews](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#ai-code-reviews) guidance we expect those to be addressed (either fixed or with a one-line reason why they're not issues) before we can take a PR forward.

The first codex point in particular is important — as written the raw tool-call JSON would both leak into the UI/history *and* get executed, which undercuts the goal of the change. The brace-scanner approach may also not be the cleanest way to solve this; it'd be good to align on the design first.

So I'm going to close this for now. Please come join us on [Discord](https://discord.gg/goose-oss) so we can discuss the underlying issue and the right approach, and we'd love for you to try again from there.

/cc @jh-block

PR #9667, opened 2026-06-08
fix(desktop): re-apply per-session model when switching to a cached chat

jezweb:
## Problem

In Goose Desktop the active model "sticks" to whatever was last selected when you switch **back** to a chat you already opened this session, instead of restoring that chat's own model.

It's intermittent in a specific way:
- **First open of a chat** (not cached) → full `resumeAgent({ load_model_and_extensions: true })` → `restore_provider_from_session` → correct per-chat model ✅
- **Return to an already-opened (cached) chat** → `resultsCache` hit in `useChatStream` returns early **without** re-applying the agent's provider → model sticks to the last selected ❌

## Repro
1. Open chat A, set model **X** (in-chat switcher).
2. Open chat B, set model **Y**.
3. Switch back to A → A is still on **Y**, not X.

(Opening A fresh after a restart works, because that's a clean resume, not a cache hit — which is why it looks intermittent.)

## Root cause

The backend already persists and restores per-session models: `agent.rs` `update_provider()` persists `model_config` to the session, and `restore_provider_from_session()` re-applies it on resume. The gap is purely the Desktop cache path: the session-load effect in `ui/desktop/src/hooks/useChatStream.ts` early-returns on a `resultsCache` hit and never re-applies the live agent's provider for the chat you switched to.

## Fix

On the cache path, re-apply the cached session's persisted provider/model via the existing `updateAgentProvider` API. Falls back to the global default when the session has no `model_config` (behaviour unchanged for those). Frontend-only, one file.

## Testing

- `tsc --noEmit` passes (no errors in the changed file).
- ⚠️ **Not yet runtime-tested on a built app** — would appreciate verification against the 3-step repro above. Keeping this as a **draft** until confirmed.

### DIFF
--- a/ui/desktop/src/hooks/useChatStream.ts
+++ b/ui/desktop/src/hooks/useChatStream.ts
@@ -12,6 +12,7 @@ import {
   sessionCancel,
   sessionReply,
   TokenState,
+  updateAgentProvider,
   updateFromSession,
   updateSessionUserRecipeValues,
   listApps,
@@ -748,6 +749,24 @@ export function useChatStream({
           },
         },
       });
+      // Re-apply this session's persisted provider/model to the live agent.
+      // The non-cached path does this via resumeAgent({ load_model_and_extensions: true }).
+      // Without it, switching back to an already-visited (cached) chat leaves the agent on
+      // the previously-active model, so the per-session model "sticks" to the last selected.
+      const cachedModelConfig = cached.session?.model_config;
+      const cachedProvider = cached.session?.provider_name;
+      if (cachedModelConfig?.model_name && cachedProvider) {
+        void updateAgentProvider({
+          body: {
+            session_id: sessionId,
+            provider: cachedProvider,
+            model: cachedModelConfig.model_name,
+            context_limit: cachedModelConfig.context_limit,
+          },
+        }).catch((error) => {
+          console.error('Failed to re-apply session provider on cached switch:', error);
+        });
+      }
       window.dispatchEvent(
         new CustomEvent(AppEvents.SESSION_EXTENSIONS_LOADED, { detail: { sessionId } })
       );


jezweb:
### Why this is an agent bug (not just display), and distinct from #7402

I want to pre-empt the natural "isn't this the already-fixed display bug?" question. #7402 ("Switching model in one tab incorrectly updates model display in all tabs", closed/completed) was **display-only** — it explicitly noted *each session continues to use its original model*. This PR addresses a different, deeper path: the **agent actually uses the wrong model** when you switch back to a *cached* chat.

The mechanism, from the backend:

- The agent holds a **single shared provider** — `agents/agent.rs:198 → provider: SharedProvider` (not one provider per session).
- **`reply()` always uses that single provider** (`self.provider()` at agent.rs:1483 / 1588 / 1679 …), there's no per-session provider lookup at reply time.
- `update_provider(provider, session_id)` *replaces* that single shared provider **and** persists the model to session metadata.
- So the model a reply runs on = whatever `self.provider` was last set to, which only changes on a `resume` / explicit model switch.

The Desktop side: `useChatStream`'s session-load effect early-returns on a `resultsCache` hit (switching back to an already-opened chat) and **never calls `resumeAgent`/`updateAgentProvider`**, so the shared provider is never re-applied for the chat you switched to.

Sequence:
1. Open A → resume → shared provider = **X**
2. Open B → resume → shared provider = **Y**
3. Switch back to A → **cache hit, no resume** → shared provider still **Y** → next message in A runs on **Y**

Likely origin: `resultsCache` is a perf optimisation added after #7402's display fix, and it reintroduced an agent-model bug by skipping the re-apply on the cache path.

This PR re-applies `updateAgentProvider` with the cached session's persisted `provider_name` + `model_config` on the cache path, restoring the correct per-chat model (falls back to the global default when the session has no `model_config`).

Analysis is static; a runtime repro (A=model X, B=model Y, switch back to A, send a message, confirm which model the agent uses) would be the final confirmation — happy to iterate.

DOsinga:
Closing as stale — this draft has had no activity since it was opened (~23 days). Please feel free to reopen if you'd like to continue the work. Thanks!

PR #9669, opened 2026-06-08
fix(local-inference): size context to the turn, not the model's full limit

macgenerate:
## Problem

On local inference through the llama.cpp Metal backend, a multi-turn session can fail mid-stream with:

```
Prefill decode failed: Decode Error -3: unknown
```

and the underlying log shows the macOS GPU watchdog aborting the command buffer:

```
kIOGPUCommandBufferCallbackErrorImpactingInteractivity
command buffer 0 failed with status 5
failed to decode, ret = -3
```

Once this fires, the backend is poisoned for the rest of the session — every subsequent decode also returns `-3`.

## Root cause

`effective_context_size()` returned the model's *full advertised* context limit (e.g. 128k for Qwen2.5) on every turn. Allocating a KV/compute graph that large builds a Metal command buffer big enough that, as the conversation grows, macOS's GPU watchdog ("Impacting Interactivity") kills it — surfacing as `ret = -3`. It looks intermittent because it only trips once the running context crosses the threshold, but it's deterministic per conversation size, not flaky.

## Fix

Size the per-turn context to what the turn actually needs — `prompt_token_count + generation headroom` — and keep the configured limit (explicit `context_size` / memory cap) as the **upper bound**, not the default allocation. This keeps the compute graph proportional to real usage and well under the watchdog ceiling.

```rust
let headroom = settings.max_output_tokens.unwrap_or(8192).max(2048);
prompt_token_count.saturating_add(headroom).min(limit)
```

One file, `inference_engine.rs`, +14/-5.

## Testing

- A multi-turn growing conversation that reliably hit `Decode Error -3` (around the 4th turn) now completes cleanly through many turns, no `-3`.
- `cargo check -p goose` green.

Metal-specific in symptom, but sizing context to the turn rather than the model max is a reasonable default for any backend.

### DIFF
--- a/crates/goose/src/providers/local_inference/llamacpp/inference_engine.rs
+++ b/crates/goose/src/providers/local_inference/llamacpp/inference_engine.rs
@@ -282,16 +282,25 @@ pub(super) fn effective_context_size(
     memory_max_ctx: Option<usize>,
 ) -> usize {
     let limit = context_cap(settings, context_limit, n_ctx_train, memory_max_ctx);
-    let min_generation_headroom = 512;
-    if prompt_token_count + min_generation_headroom > limit {
+
+    // Size the context to what THIS turn actually needs (prompt + generation
+    // headroom), not the model's full advertised limit. Allocating the full limit
+    // (e.g. 128k for Qwen) builds a Metal compute graph large enough that macOS
+    // aborts the GPU command buffer ("Impacting Interactivity" -> command buffer
+    // status 5 -> "failed to decode, ret = -3"), which then poisons the backend for
+    // the rest of the session. `limit` (which honors an explicit `context_size`
+    // setting and the memory cap) stays the upper bound.
+    let headroom = settings.max_output_tokens.unwrap_or(8192).max(2048);
+    let sized = prompt_token_count.saturating_add(headroom).min(limit);
+
+    if prompt_token_count + 512 > limit {
         tracing::warn!(
-            "Prompt ({} tokens) + minimum headroom ({}) exceeds context limit ({})",
+            "Prompt ({} tokens) + minimum headroom (512) exceeds context limit ({})",
             prompt_token_count,
-            min_generation_headroom,
             limit,
         );
     }
-    limit
+    sized
 }
 
 pub(super) fn build_context_params(


kimnamu (commented):
Thanks for the careful root-cause writeup — pinning `ret = -3` to the macOS GPU watchdog aborting an oversized Metal command buffer (and noting it poisons the backend for the rest of the session) is genuinely useful, and sizing the context to the turn is a sensible default. I'm not a maintainer, just another goose/local-inference user; I built the branch locally (`--features local-inference`, cargo 1.92) and ran the unit tests. A few things:

**Confirmed safe across turns.** I traced the callers — `create_and_prefill_context` creates a fresh `LlamaContext` sized to `effective_ctx` every turn, and `effective_context_size` is recomputed per turn from the current `prompt_token_count`, so growing conversations still get a proportionally larger context. Good.

1. *(the main gap)* All 9 existing tests still pass, but only by coincidence: they all use `limit ≤ 4096`, and with the default `max_output_tokens = None` the headroom is `8192`, so `min(prompt + 8192, limit)` always collapses to `limit`. None of them exercises the new sub-limit sizing — and `test_effective_context_size_uses_full_limit` now returns the full limit only because headroom happens to exceed it. Adding a test under a large limit pins the actual fix (these pass on the branch, and fail on `main`):
   ```rust
   #[test]
   fn test_effective_context_size_sized_to_turn_under_large_limit() {
       // Regression for the GPU-watchdog crash: under a large advertised limit
       // (e.g. 128k) we allocate prompt + headroom, not the full limit.
       // default max_output_tokens = None -> headroom = 8192.
       assert_eq!(
           effective_context_size(100, &default_settings(), 131072, 131072, None),
           100 + 8192,
       );
   }

   #[test]
   fn test_effective_context_size_headroom_floor() {
       let mut settings = default_settings();
       settings.max_output_tokens = Some(100); // below the 2048 floor
       assert_eq!(
           effective_context_size(100, &settings, 131072, 131072, None),
           100 + 2048,
       );
   }

   #[test]
   fn test_effective_context_size_honors_max_output_tokens() {
       let mut settings = default_settings();
       settings.max_output_tokens = Some(40000);
       assert_eq!(
           effective_context_size(100, &settings, 131072, 131072, None),
           100 + 40000,
       );
   }
   ```
   (I verified these compile and pass on your branch; please double-check the `default_settings()` / signature names against the file — I'm transcribing from the local checkout.)

2. *(behavioral, non-blocking)* This narrows generation headroom from "up to `limit - prompt`" to a fixed `headroom` (8192 by default). On a large local model that previously could keep generating well past 8k tokens of context, output now gets capped earlier unless the user sets `max_output_tokens`. That's a reasonable default, but worth a one-line mention in the PR body so a maintainer doesn't read it as an unintended cap.

3. *(nit)* The `tracing::warn!` still says `minimum headroom (512)` and triggers on `prompt + 512 > limit`, but the actual sizing headroom is now `8192`/`max_output_tokens`. The 512 threshold and the real headroom have diverged — consider warning against `headroom` (or noting 512 is just the "won't fit at all" floor) so the log matches the allocation.

Tight, well-motivated fix — adding the sub-limit test is the main thing I'd want before it lands. (Prepared with help from an AI agent (Claude Code); I built the branch and verified each point myself.)

DOsinga:
Thanks for digging into this — the GPU watchdog `Decode Error -3` is a real bug and sizing the context to the turn is a sensible direction.

I'm going to close this for now though. The main blocker is the unaddressed codex/review point: when `max_output_tokens` is unset (the default), the old code let generation use all remaining context, but with this change `effective_ctx` becomes `prompt + 8192`, which collapses the generation headroom to 8192 and will surface as `ContextLengthExceeded` for longer responses that previously worked. That regression, plus the now-stale "minimum headroom (512)" log and the lack of a test exercising the new sub-limit path, all need addressing — see https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#ai-code-reviews.

The underlying fix has merit, so please feel free to reopen once the headroom behavior is reworked (e.g. allocate the full remaining context for output while capping the compute graph some other way).

/cc @jh-block in case you want to pick this up later — the watchdog crash is worth fixing.

PR #5563, opened 2025-11-04
disable RPM build-ID generation to prevent package conflicts

DOsinga:
Fixes #5432

I haven't tested this though, this is what the LLMs said we should do

PR #9671, opened 2026-06-08
fix(packaging): disable RPM build-id links

### LINKED ISSUE #9485: RPM build-id file conflict with opencode on Fedora 43 (v1.36.0)
## Bug description
Installing Goose-1.36.0-1.x86_64.rpm on Fedora 43 fails with:
```
file /usr/lib/.build-id/f2/1efa6a9c0979ef4eea2b0d873506f78579c48c from install of Goose-1.36.0-1.x86_64 conflicts with file from package opencode-1.15.11-1.x86_64
```

## Root cause
Both Goose and opencode ship a binary with the same ELF build-id hash. The offending path is under /usr/lib/.build-id/f2/, which comes from Goose's bundled tools (uvx/node/npx etc.).

## Steps to reproduce
1. Install opencode: `sudo dnf install opencode`
2. Install Goose RPM: `sudo dnf install ./Goose-1.36.0-1.x86_64.rpm`
3. Observe rpm transaction failure

## Workaround
`sudo rpm -i --replacefiles ./Goose-1.36.0-1.x86_64.rpm`

## Related issues
- #5432 - original report (conflict with Lens)
- #6609 - duplicate (conflict with Slack)
- PR #5563 - fix shipped in v1.13.0 but appears to have regressed or was incomplete

yibo365:
## Summary

- configure `rpmbuild` with `%_build_id_links none` through `~/.rpmmacros` in the Linux packaging job
- remove the unsupported `fpm` option from the Electron Forge RPM configuration
- fail the packaging job if a generated RPM still contains `/usr/lib/.build-id` entries

The current RPM maker uses `electron-installer-redhat`, which invokes `rpmbuild` directly. As a result, the existing `fpm` option is ignored. The v1.37.0 release RPM still contains the exact build-id path reported in the issue.

### Testing

- `actionlint .github/workflows/bundle-desktop-linux.yml`
- parsed the workflow with Ruby YAML
- `git diff --check`
- inspected `Goose-1.37.0-1.x86_64.rpm` and confirmed it contains `/usr/lib/.build-id/f2/1efa6a9c0979ef4eea2b0d873506f78579c48c`
- reproduced the RPM behavior in an Ubuntu 22.04 container using a minimal ELF binary:
  - default `%_build_id_links=compat` produced `/usr/lib/.build-id/...` entries
  - `%_build_id_links none` through `/root/.rpmmacros` produced an RPM with no build-id links
  - the same path-check logic used by the workflow accepted the fixed RPM

### Related Issues

Fixes #9485

N/A

### DIFF
--- a/.github/workflows/bundle-desktop-linux.yml
+++ b/.github/workflows/bundle-desktop-linux.yml
@@ -104,6 +104,11 @@ jobs:
             sudo apt-get install -y libasound2
           fi
 
+      - name: Configure RPM packaging
+        run: |
+          echo '%_build_id_links none' > ~/.rpmmacros
+          test "$(rpm --eval '%{_build_id_links}')" = "none"
+
       - name: Setup Flatpak Runtimes
         run: |
           flatpak remote-add --if-not-exists --user flathub https://dl.flathub.org/repo/flathub.flatpakrepo
@@ -180,6 +185,14 @@ jobs:
               mv "$package" "${base}-vulkan.${extension}"
             done
           fi
+
+          for package in out/make/rpm/x64/*.rpm; do
+            package_contents="$(rpm -qlp "$package")"
+            if grep -qE '^/usr/lib/\.build-id(/|$)' <<< "$package_contents"; then
+              echo "::error file=$package::RPM contains build-id links"
+              exit 1
+            fi
+          done
 
           echo "Build completed. Checking output..."
           ls -la out/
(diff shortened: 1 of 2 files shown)


DOsinga (approved):
Thanks for tracking this down! The analysis is spot on — `@electron-forge/maker-rpm` uses `electron-installer-redhat`, which invokes `rpmbuild` directly and ignores the `fpm` option from the previous attempt (#5563), so setting `%_build_id_links none` via `~/.rpmmacros` is the right mechanism. Nice touch adding the CI guard to fail if a `/usr/lib/.build-id` entry ever slips back in. Approving.

PR #9678, opened 2026-06-08
View json

DOsinga:
## Summary

Show json for a conversation directly


ui/desktop/src/i18n/messages/zh-CN.json
```diff
@@ -3746,6 +3746,36 @@
   "securityToggle.warpNotice": {
     "defaultMessage": "连接到 WARP 时，命令注入检测效果最佳（访问分类服务需要 WARP）。"
   },
+  "sessionActionsHeader.close": {
```
jamadeo, line 3749: These (and in the other files) look like they're the same as the default ones

jamadeo (approved):

PR #9679, opened 2026-06-08
docs: add RunAPI OpenAI-compatible example

runapi-builder:
Adds RunAPI to the custom OpenAI endpoint examples in the provider setup docs.

### DIFF
--- a/documentation/docs/getting-started/providers.md
+++ b/documentation/docs/getting-started/providers.md
@@ -308,6 +308,15 @@ Need to connect to multiple OpenAI-compatible endpoints? [Configure custom provi
     OPENAI_PROJECT=ml-serving
     ```
   </TabItem>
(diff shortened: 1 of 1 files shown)


DOsinga:
Thanks for the PR! We're going to pass on adding RunAPI to the provider docs here — we don't want our docs to become a directory of every OpenAI-compatible vendor, and this kind of vendor self-listing is a long-term curation/maintenance cost that's best decided with a prior discussion/issue (see https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#discussions-issues-and-prs).

That said, if you'd like RunAPI to be a first-class option in goose, a much more valuable contribution would be to add it as a declarative provider rather than a docs example. That makes it discoverable and configurable for users directly, instead of being buried in a markdown tab. Happy to point you in the right direction if you want to go that route.

PR #9680, opened 2026-06-08
feat: add Gladia speech-to-text dictation provider

jqueguiner:
## What

Adds [Gladia](https://gladia.io) as a voice dictation provider, alongside the existing OpenAI, Groq, ElevenLabs and local Whisper options.

## Why

Gladia is a multilingual speech-to-text API. This wires it into the existing dictation subsystem so users can transcribe voice input with their Gladia key.

## How

Gladia's pre-recorded API is **asynchronous** and doesn't fit the shared single-request `transcribe_with_provider` path (which expects one multipart POST returning `{ "text": ... }`). It gets a dedicated `transcribe_gladia` implementing the upload → create job → poll flow:

1. `POST /v2/upload` (multipart `audio`) → `audio_url`
2. `POST /v2/pre-recorded` `{ audio_url, model }` → job `id` (model: `solaria-1`)
3. `GET /v2/pre-recorded/{id}` until `status == "done"`, then read `result.transcription.full_transcript`

Auth uses the `x-gladia-key` header with its own key (`GLADIA_API_KEY`), mirroring the ElevenLabs key-based pattern. The desktop settings UI is data-driven from `/dictation/config`, so the provider and its API-key field appear automatically — no UI code change.

### Touch points
- `DictationProvider` enum, `PROVIDERS` registry, auth arm + `transcribe_gladia` (`crates/goose/src/dictation/providers.rs`)
- REST route dispatch (`crates/goose-server/src/routes/dictation.rs`)
- ACP transcribe + model-config handlers (`crates/goose/src/acp/server/dictation.rs`)
- Regenerated OpenAPI schema + frontend types

## Tests

Unit tests for transcript parsing, poll-status interpretation, request payload (model included), and provider registration. `cargo fmt`, `cargo clippy`, and `cargo test -p goose` all pass.

## Manual verification

Tested end-to-end in the desktop app: configured a real `GLADIA_API_KEY`, spoke into the mic, and got a correct live transcription via the Gladia async pipeline.

### DIFF
--- a/crates/goose-server/src/routes/dictation.rs
+++ b/crates/goose-server/src/routes/dictation.rs
@@ -9,11 +9,7 @@ use axum::{
 #[cfg(feature = "local-inference")]
 use axum::{extract::Path, routing::delete};
 use base64::{engine::general_purpose::STANDARD as BASE64, Engine};
-#[cfg(feature = "local-inference")]
-use goose::dictation::providers::transcribe_local;
-use goose::dictation::providers::{
-    all_providers, is_configured, transcribe_with_provider, DictationProvider,
-};
+use goose::dictation::providers::{all_providers, is_configured, transcribe, DictationProvider};
 #[cfg(feature = "local-inference")]
 use goose::dictation::whisper;
 #[cfg(feature = "local-inference")]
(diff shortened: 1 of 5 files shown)


DOsinga:
Thanks for the PR, and for the clean implementation — the dispatch consolidation into a single `transcribe()` entry point is tidy.

That said, I'm going to close this one. A few reasons:

- This adds Gladia as a fourth *batch* cloud speech-to-text provider, using their `v2/pre-recorded` upload-and-poll flow (poll every 1s, up to ~2 min). Functionally it sits in the same category as the OpenAI/Groq/ElevenLabs providers we already have, and we also ship local Whisper which runs offline with no API key. So it adds maintenance surface (new provider, config key, registry entry, UI enum) for fairly marginal additional value.
- The thing we'd actually be interested in here is a **realtime/streaming** dictation provider (partial transcripts, low latency) — which Gladia does support via their WebSocket API, but that's not what this implements, and it would need a proper design discussion first since it touches the dispatch shape.
- Per our [contributing guidelines](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#discussions-issues-and-prs), a new third-party integration like this is best started as a discussion/issue so we can agree on the direction before the code — and as a first contribution, smaller, scoped changes (or opening one PR and waiting for it to land before the next) work best per the [PR guidelines](https://github.com/aaif-goose/goose/blob/main/CONTRIBUTING.md#pull-requests).

If you'd like to pursue realtime/streaming dictation, please open a discussion and we can figure out the right design together — that'd be a genuinely valuable contribution.

PR #9694, opened 2026-06-09
fix: clear rejected OAuth credentials after refresh

### LINKED ISSUE #9673: MCP OAuth: post-refresh 401 loops on refresh instead of falling back to interactive re-auth
### Summary

When an MCP `initialize` returns `401 invalid_token` and Goose has a stored **refresh** token, Goose silently `POST /token` refreshes and retries. If the **refreshed** token is *still* rejected (e.g. `error_description="Invalid audience"`), Goose loops on refresh and surfaces `failed to initialize MCP client … Auth required` — **without ever triggering interactive re-authorization**.

Contrast: a token that is simply expired/invalid works fine — refresh *fails* → creds cleared → fresh `/authorize` browser flow. The dead-end is specifically when refresh **succeeds** but yields a token that's still rejected on use.

### Root cause

`crates/goose/src/oauth/mod.rs` — `oauth_flow` returns immediately on successful refresh, without verifying the token actually works:

```rust
if auth_manager.initialize_from_store().await? {
    match auth_manager.refresh_token().await {
        Ok(_)  => { return Ok(auth_manager); }   // returns refreshed token, unverified
        Err(e) => { warn!("...clearing stored credentials and falling back to browser auth"); }
    }
    credential_store.clear().await...;            // clear only on refresh Err
}
```

`crates/goose/src/agents/extension_manager.rs` (proactive path) then `return`s `connect_with_auth(refreshed_manager, …)` directly. If that `connect`/`initialize` 401s (refreshed token rejected), the error is returned — creds are never cleared (refresh "succeeded"), no interactive auth is attempted. The next session repeats it → loop.

### Fix

On a **post-refresh 401** (refreshed token rejected at `connect`/`initialize`), discard the grant (`credential_store.clear()`) and start interactive authorization (the browser `/authorize` flow), rather than returning the error. Concretely: in the proactive `connect_with_auth(refreshed_manager)` path, if the result is an auth-required error, clear creds and fall through to the full `oauth_flow` browser path.

### Secondary improvements

- Removing an extension should purge its stored OAuth creds (`oauth_creds_<name>`) so a re-add starts clean.
- A **"Reconnect / clear credentials"** action on an extension would let users escape this dead-end without hand-editing secrets.

### Repro

Use a server that rejects a token's audience (an audience/resource-indicator mismatch). Authenticate once (stores a refresh token), then put it in a state where the refreshed access token's audience is rejected → Goose loops on "Auth required" and never re-opens the browser, even though discarding the grant + re-authing would succeed.

qybaihe:
Fixes #9673

## Summary
- Detect OAuth auth failures from both typed `AuthRequired` errors and wrapped HTTP 401 transport errors.
- Clear rejected stored OAuth credentials after a refresh attempt still fails authentication.
- Fall back to the existing browser OAuth path so the user can re-authorize instead of looping on stale credentials.

## Tests
- `cargo fmt --check`
- `cargo test -p goose test_post_refresh_auth_failure_clears_credentials`
- `cargo test -p goose oauth_fallback`

### DIFF
--- a/crates/goose/src/agents/extension_manager.rs
+++ b/crates/goose/src/agents/extension_manager.rs
@@ -387,11 +387,11 @@ async fn child_process_client(
 }
 
 /// Retry with OAuth for typed auth challenges and wrapped bare HTTP 401 responses.
-fn should_attempt_oauth_fallback(res: &Result<McpClient, ClientInitializeError>) -> bool {
-    let Err(ClientInitializeError::TransportError {
+fn is_oauth_auth_failure(err: &ClientInitializeError) -> bool {
+    let ClientInitializeError::TransportError {
         error: DynamicTransportError { error, .. },
         ..
-    }) = res
+    } = err
     else {
         return false;
     };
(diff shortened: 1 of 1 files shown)


DOsinga (approved):
Thanks for the clean, well-scoped fix — and for linking it to #9673. I checked out the branch, merged main (no conflicts), and verified locally:

- `cargo build -p goose` ✅
- New test `test_post_refresh_auth_failure_clears_credentials` and the existing oauth/custom-header tests all pass ✅
- `cargo clippy -p goose --all-targets` clean (the only warning is a pre-existing unrelated dead-code warning in tests/providers.rs)
- `cargo fmt --check` clean ✅

The control flow is right: on a post-refresh auth failure we now clear the rejected stored credentials and fall through to the unauthenticated connect, which 401s and triggers the browser `oauth_flow` so the user can re-authorize instead of looping on stale credentials. Nice reuse of `is_oauth_auth_failure` rather than duplicating the matcher, and the test follows the existing patterns in the file. LGTM 🚀

PR #9712, opened 2026-06-10
pin goose-sdk package in tui

lifeizhou-ap:
## Summary
Since TUI are not in active development now, we pin the goose-sdk package so that it won't be impacted by the ACP schema changes

### Testing
Local testing

### DIFF
--- a/ui/pnpm-lock.yaml
+++ b/ui/pnpm-lock.yaml
@@ -404,8 +404,8 @@ importers:
   text:
     dependencies:
       '@aaif/goose-sdk':
-        specifier: workspace:*
-        version: link:../sdk
+        specifier: 0.20.2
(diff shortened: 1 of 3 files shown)


jamadeo (approved):

jamadeo:
We can always bump it manually too if we make changes.

@lifeizhou-ap does this also effectively pin the binary version? if we don't pin that as well you could get something incompatible I think?

lifeizhou-ap:
> We can always bump it manually too if we make changes.
> 
> @lifeizhou-ap does this also effectively pin the binary version? if we don't pin that as well you could get something incompatible I think?

Thank you @jamadeo for the review! 


Yes, the binary's pinned too. The SDK declares the @aaif/goose-binary-* packages as workspace:*, which publishes as an exact pin, so SDK 0.20.2 can only pull binary 0.20.2

PR #9713, opened 2026-06-10
feat(security): unified OTLP logging schema for cross-tool detection

dorien-koelemeijer:
Applies a unified logging schema (security.*, session.*, tool.*, network.*) to Goose's security subsystem, aligning it with the same schema implemented in Sandpit and BuilderBot/Blox.

## Why
Security telemetry across our AI agent tools (Goose, Sandpit, BuilderBot) used inconsistent field names and structures, making cross-tool Snowflake queries and DART detections impractical. This PR standardises Goose's security log attributes so a single query can surface findings across all tools.

### Security log attributes (Layer 1 base + Layer 2 category):
- security/mod.rs: finding + passed events emit security.event_type, security.action, security.confidence,  security.threat_type, etc.
- security/egress_inspector.rs: egress events emit security.threat_type = "data_exfiltration" + network.* attributes
- security/adversary_inspector.rs: ALLOW/BLOCK events use security.* namespace
- security/scanner.rs: analysis complete uses scanner.* namespace
- tool_inspection.rs: inspection result applied uses unified naming
- agents/tool_execution.rs: user decision uses unified naming

### Session context (pivot points for correlation):
- Added session.user, session.host, session.agent_type to OTLP span allowlist - these propagate to all log records within a reply_stream span
- session.user and session.host cached via OnceLock (computed once per process)

### OTel resource enrichment:
- Added host.name (via gethostname) and user.name to the OTel resource

### Test cleanup:
- Simplified create_resource tests to assert contains rather than exact equality - less brittle when resource attributes change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4572,6 +4572,7 @@ dependencies = [
  "fs-err 3.3.0",
  "fs2",
  "futures",
+ "gethostname",
  "goose-acp-macros",
  "goose-mcp",
  "goose-providers",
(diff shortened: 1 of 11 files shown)


shellz-n-stuff (approved):
🔥 /10 LGTM

michaelneale (approved):
LGTM - only possible concern is gethostname - want to make sure works on windows etc.

PR #9716, opened 2026-06-10
feat(azure): support Entra ID bearer token auth via AZURE_OPENAI_AD_TOKEN

### LINKED ISSUE #9548: Support Azure OpenAI Bearer Token Authentication (AZURE_OPENAI_AD_TOKEN)
**Motivation:**

The Azure OpenAI provider currently requires Goose to obtain its own credentials via Azure CLI, Service Principal credentials, or API keys.

In enterprise environments, especially hosted VS Code/code-server deployments, it is often necessary to expose only a short-lived Azure OpenAI access token to the runtime while keeping Azure credentials (client ID, client secret, tenant ID) secured in a backend service.

Currently Goose appears to ignore a provided Azure OpenAI bearer token and attempts to acquire its own token, which prevents this deployment pattern.

**Description:**

Support direct Azure OpenAI bearer token authentication via an environment variable, for example:

```AZURE_OPENAI_AD_TOKEN=<access_token>```

When this variable is present, Goose should:

* Use the token directly as:

```Authorization: Bearer <token>```

* Skip Azure CLI authentication.
* Skip Service Principal credential flows.
* Skip token acquisition.

**Alternatives Considered:**

* Exposing Azure Service Principal credentials to the runtime (not acceptable).
* Building a custom proxy for azure models (additional infrastructure and complexity).
* Maintaining a custom Goose fork.

**Additional context:**

I verified that Azure OpenAI works correctly using a bearer token obtained via:

```az account get-access-token --resource https://cognitiveservices.azure.com```

which goose does internally using azure cli.

Supporting bearer-token authentication would enable secure enterprise deployments where only short-lived access tokens are exposed to the user runtime.

* I have verified this does not duplicate an existing feature request

hammadxcm:
## Summary

Fixes #9548

In enterprise deployments (e.g. hosted VS Code / code-server), only a short-lived Azure OpenAI access token is exposed to the runtime, while Azure credentials (client ID, secret, tenant) stay in a backend service. The Azure OpenAI provider previously supported only API keys or acquiring its own token through the Azure CLI credential chain, so this deployment pattern wasn't possible.

## Changes

- Add a `BearerToken` variant to `AzureCredentials`, populated from a new optional secret config key `AZURE_OPENAI_AD_TOKEN` (works from env var or config.yaml, same as the API key).
- When set, the token is sent directly as `Authorization: Bearer <token>` and Azure CLI / token acquisition is skipped entirely — no caching, no subprocess.
- Precedence: `AZURE_OPENAI_AD_TOKEN` > `AZURE_OPENAI_API_KEY` > Azure credential chain (the AD token is short-lived and explicitly injected, so setting it signals intent). Documented in the provider docs.
- Registered the key in provider metadata (optional, secret) and the curated catalog so the Desktop setup UI shows it as a masked optional field.
- Updated `documentation/docs/getting-started/providers.md` (provider table + the Azure OpenAI auth section, now listing all three methods and their precedence).

Behavior is unchanged when the new variable is unset (empty values are filtered, matching the `AWS_BEARER_TOKEN_BEDROCK` pattern).

## Testing

- New unit tests in `azureauth.rs`: credential selection precedence (token > key > default chain) and `get_token()` returning the bearer token without invoking the Azure CLI.
- New unit tests in `azure.rs`: auth header is `Authorization: Bearer <token>` for the bearer path and `api-key` for the API key path.
- `cargo fmt`, `cargo clippy --all-targets -- -D warnings` (with `aws-providers`), and `cargo test -p goose providers::azure` (7/7 passing) all clean.

Manual verification of the token flow matches the issue author's setup: a token from `az account get-access-token --resource https://cognitiveservices.azure.com` placed in `AZURE_OPENAI_AD_TOKEN` authenticates directly.

### DIFF
--- a/crates/goose-providers/src/canonical/catalog.rs
+++ b/crates/goose-providers/src/canonical/catalog.rs
@@ -693,6 +693,12 @@ const SETUP_METADATA: &[CuratedSetupMetadata] = &[
                 placeholder: Some("Paste your API key"),
                 default_value: None,
             },
+            CuratedFieldMetadata {
+                key: "AZURE_OPENAI_AD_TOKEN",
+                label: "Entra ID Token",
+                placeholder: Some("Optional: short-lived Microsoft Entra access token"),
+                default_value: None,
+            },
         ],
     },
     CuratedSetupMetadata {
--- a/crates/goose/src/providers/azure.rs
+++ b/crates/goose/src/providers/azure.rs
@@ -66,7 +66,7 @@ impl ProviderDef for AzureProvider {
                 ConfigKey::new("AZURE_OPENAI_DEPLOYMENT_NAME", true, false, None, true),
                 ConfigKey::new("AZURE_OPENAI_API_VERSION", false, false, None, false),
                 ConfigKey::new("AZURE_OPENAI_API_KEY", false, true, Some(""), true),
-                ConfigKey::new("AZURE_OPENAI_AD_TOKEN", false, true, Some(""), false),
+                ConfigKey::new("AZURE_OPENAI_AD_TOKEN", false, true, Some(""), true),
             ],
         )
     }
--- a/crates/goose/src/providers/azure.rs
+++ b/crates/goose/src/providers/azure.rs
@@ -41,7 +41,8 @@ impl AuthProvider for AzureAuthProvider {
             super::azureauth::AzureCredentials::ApiKey(_) => {
                 Ok(("api-key".to_string(), auth_token.token_value))
             }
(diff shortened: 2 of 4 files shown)


DOsinga (approved):
Thanks for the contribution! This cleanly adds Entra ID bearer token auth via `AZURE_OPENAI_AD_TOKEN`, follows the existing credential pattern, has meaningful tests, and addresses both codex comments. I pushed a small follow-up renaming the docs section heading to cover all three auth methods. LGTM.

